PR #345, opened 2016-10-12
Nirspec msa conf

nden:
Nirspec MSA WCS with new configuration file.
Also the code was refactored so that the same code base is used for the ifu and fs mode.
The only interface difference is the way to get a WCS object for a specific slit now takes only the name of the slit, e.g. 

```
nirspec.nrs_wcs_set_input(input_model, 'S200A1') # fixed slits
nirspec.nrs_wcs_set_input(input_model, 0) # ifu slice number
nirspec.nrs_wcs_set_input(input_model, 58) # msa slitlet id
```

### DIFF
--- a/jwst/assign_wcs/assign_wcs.py
+++ b/jwst/assign_wcs/assign_wcs.py
@@ -35,6 +35,9 @@ def load_wcs(input_model, reference_files={}):
         output_model.meta.cal_step.assign_wcs = 'SKIPPED'
     else:
         wcs = WCS(pipeline)
+        #if output_model.meta.exposure.type.lower() in ['nrs_fixedslit', 'nrs_msaspec']:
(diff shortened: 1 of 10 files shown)


nden:
@brechmos-stsci I think this is now ready to be merged if you want to review it. I added two more members to the `Slit` structure - the number of shutters, "nshutters", needed by the pathloss correction and a "shutter_id" - computed (unique per quadrant) shutter id which is a key in the MSA transform.

brechmos-stsci:
I read through it and it all seems fine.

I like the error :-)
```
       models = [m1, m2]
       s0 = models.Slit("s0", 1, 2, 3, 4, 5, 6, 7)
```

nden:
This PR is stuck on the OSX tests but has passed on my personal Travis so merging.

PR #347, opened 2016-10-13
Changes towards MIRI LRS to go to sky coordinates

brechmos-stsci:
Under @nden 's direction, this will run on both slitless prism exposure and fixed slit files (at least the two that I have:

```
jw00025001001_01107_00002_MIRIMAGE_rate.fits - slitlessprism exposure
jw00035001001_01101_00001_MIRIMAGE_rate.fits - fixed slit
```

### DIFF
--- a/jwst/assign_wcs/miri.py
+++ b/jwst/assign_wcs/miri.py
@@ -74,53 +74,128 @@ def imaging_distortion(input_model, reference_files):
     ref_file: filter_offset.asdf - (1)
     ref_file: distortion.asdf -(2,3,4)
     """
+
+    # Load in the distortion file.
     distortion = AsdfFile.open(reference_files['distortion']).tree['model']
-    filter_offset = AsdfFile.open(reference_files['filteroffset']).tree[input_model.meta.instrument.filter]
-    full_distortion = models.Shift(filter_offset['row_offset']) & models.Shift(
-        filter_offset['column_offset']) | distortion
+
+    # ToDo: This will likely have to change in the future, but the "filteroffset" file we have
+    # ToDo: currently does not contain that key.
+    filter_offset = None
+    if input_model.meta.instrument.filter in  AsdfFile.open(reference_files['filteroffset']).tree:
+        filter_offset = AsdfFile.open(reference_files['filteroffset']).tree[input_model.meta.instrument.filter]
+        full_distortion = models.Shift(filter_offset['row_offset']) & models.Shift(
+            filter_offset['column_offset']) | distortion
+    else:
+        full_distortion = distortion
+
     full_distortion = full_distortion.rename('distortion')
+
     return full_distortion
 
+
+def is_number(s):
+
+    # Check, first, to make sure it is not None
+    if not s:
+        return False
+
+    # Now, check to see if it is a number
+    try:
+        float(s)
+        return True
+    except ValueError:
+        return False
 
 def lrs(input_model, reference_files):
     """
     Create the WCS pipeline for a MIRI fixed slit observation.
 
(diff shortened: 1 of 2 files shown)


brechmos-stsci:
@nden We should go over this before it is committed to be sure it is correct.

jwst/assign_wcs/miri.py
```diff
@@ -74,53 +74,128 @@ def imaging_distortion(input_model, reference_files):
...
+    except ValueError:
+        return False
+
 def lrs(input_model, reference_files):
     """
     Create the WCS pipeline for a MIRI fixed slit observation.
 
-    reference_files = {"specwcs": 'MIRI_FM_MIRIMAGE_P750L_DISTORTION_04.02.00.fits'}
+    reference_files = {
+        "specwcs": 'MIRI_FM_MIRIMAGE_P750L_DISTORTION_04.02.00.fits'
```
hbushouse: You should be aware that there was an error in the header keyword values of the original version of this reference file as delivered by the MIRI IDT in CDP-4. In particular, IMXSLTL=60, which is the x-axis zeropoint of the slitless trace, is wrong (by a lot). The correct value is 34. The other values for IMX and IMY were also updated (by small amounts) in the more recent CDP-6 delivery of this ref file. The latest file can be found at /grp/jwst/wit/miri/CDP-6/CALSPEC2_LRS/distortion_wavelength_calibration/DISTORTION/referencefile/MIRI_FM_MIRIMAGE_P750L_DISTORTION_06.02.00.fits

jwst/assign_wcs/miri.py
```diff
@@ -74,53 +74,128 @@ def imaging_distortion(input_model, reference_files):
...
+
+    ra_t2d = models.Tabular2D(lookup_table=ra_full)
+    dec_t2d = models.Tabular2D(lookup_table=dec_full)
+
+    # Create the model transforms.
     lrs_wav_model = jwmodels.LRSWavelength(lrsdata, zero_point)
-    ref.close()
+
+    # Incorporate the small rotation
     angle = np.arctan(0.00421924)
```
hbushouse: Shouldn't the rotation angle come from the reference file, rather than being hardwired in the code? Granted, it's unlikely to change (much) in the future, other than having the whole instrument get nocked around by launch, but still seems more appropriate to have it come from the ref file.

brechmos-stsci: Good point. I'll look into this.

jwst/assign_wcs/miri.py
brechmos-stsci: I changed the file listed in the comment, the correct file was being used in the `assign_wcs.cfg` configuration file:
```
$ cat assign_wcs.cfg
name = "assign_wcs"
class = "jwst.assign_wcs.AssignWcsStep"
override_specwcs="/Users/crjones/Documents/JWST/MIRI-LRS/MIRI_FM_MIRIMAGE_P750L_DISTORTION_06.02.00.fits"
override_distortion="/Users/crjones/Documents/JWST/MIRI-LRS/jwst_miri_distortion_0001.asdf"
override_filteroffset="/Users/crjones/Documents/JWST/MIRI-LRS/jwst_miri_filter_offset_0001.asdf"
```

jwst/assign_wcs/miri.py
nden: I agree the angle should be in the reference file. Right now it comes from the report which accompanies the old reference file. I have not looked at the new report. 
I am not sure this rotation should be applied though. It was my interpretation of the report at the time that it should be applied but this was never confirmed/denied by the team.
We should look at the report that came with the latest reference file and at a recent response from the IDT (which I;ll forward separately).

PR #366, opened 2016-10-20
Adding actual sky equalization capability for 2D models

mcara:
Previously ``cube_skymatch`` could only compute background polynomials that, when subtracted from model data, would produce "matched" backgrounds. This PR adds the capability to subtract background corrections from 2D model data.


mcara:
I just want to note that this addition has not been tested since I still have not received any test images (2D and resampled cubes) with valid WCS. I have moderate confidence in this code but I think any bugs revealed when running on real images should be trivial and easily fixable.

mcara:
I just want to note that this addition has not been tested since I still have not received any test images (2D and resampled cubes) with valid WCS. I have moderate confidence in this code but I think any bugs revealed when running on real images should be trivial and easily fixable.

PR #368, opened 2016-10-20
fix wcs of a resampled cube

nden:
fix bugs reported by @jemorrison


nden:
unrelated install error, so merging

PR #369, opened 2016-10-20
Add source information to slits

### LINKED ISSUE #212: extract_2d must include source information in the slits
As this [comment](https://github.com/STScI-JWST/jwst/pull/207#issuecomment-245620576) clarifies `extract_2d` should use the MSA configuration file to include some (all?) source information in the slits.

nden:
Closes #212


jwst/extract_2d/extract_2d.py
```diff
@@ -17,54 +18,62 @@
...
+            output_model.slits[nslit].name = str(slit.name)
+            output_model.slits[nslit].xstart = xlo + 1
+            output_model.slits[nslit].xsize = xhi - xlo
+            output_model.slits[nslit].ystart = ylo + 1
+            output_model.slits[nslit].ysize = yhi - ylo
+            if exp_type.lower() == 'nrs_msaspec':
+                output_model.slits[nslit].source_id = int(slit.source_id)
+                output_model.slits[nslit].source_name = slit.source_name
+                output_model.slits[nslit].source_alias = slit.source_alias
+                output_model.slits[nslit].catalog_id  = int(slit.catalog_id)
```
hbushouse: We need to check with Lisa Sherbert to confirm the data types for a few of these entries, like source_id and catalog_id. The correct data type may not be the same as the data type in the few example MSA config files we got many months ago. We used to think the catalog_id would be a simple number, but I believe it's actually going to be a string giving the name of the catalog from which the sources come.

nden:
Merging this. If we get clarifications on the format, will create a new PR.

PR #385, opened 2016-10-26
Add correct L2C naming to calwebb_image3

stsci-hack:
These changes should (based on my testing) correctly implement the Level 2C naming convention for L3 processing.  It also runs FITSBLENDER on the generated L3 products to create the output extension necessary for the CAOM processing by the archive.


hbushouse:
If this has been tested to be working correctly, let's go ahead and merge it so that it gets into the next B7 release.

PR #387, opened 2016-10-27
Determine cube type

jemorrison:
save celts as degrees
added weighting option  to speed up MIRI cube building
removed print statements
handle units from assign_wcs correctly


hbushouse:
Travis failures were for wcs and crds tests (i.e. not this code) and only occurred on Linux systems, so merging this now because we need it.

Issue #379, opened 2016-10-26
Change MSA slit name for flat_field


philhodge:
The `flat_field` step assumes that the MSA slit name is a string containing two integers within brackets.  The format is now a string containing a single integer, so `flat_field` should use that integer as the identifier and get the quadrant separately.

PR #390, opened 2016-10-28
Changed MSA slit name syntax, see #379

philhodge:
@nden I think this is straightforward, but if you have time could you take a quick look at it?


jwst/flatfield/flat_field.py
```diff
@@ -405,8 +397,21 @@ def do_NIRSpec_flat_field(output_model,
...
         log.debug("Processing slit %s", slit.name)
+        slit_nt = None
+        if exposure_type == "NRS_MSASPEC":
+            # Find this slit in the list of open slits.
+            for j in range(len(slits)):
+                if str(slits[j].name) == slit.name:
+                    slit_nt = slits[j]
+                    break
+            if slit_nt is None:
+                log.error("Couldn't find slit %s in list of open slits; "
```
nden: Is this actually the list of extracted slits. If so the message may be a bit misleading.

nden:
@philhodge Looks good to me. I left a couple of minor comments. The test failure looks like aTravis problem so :+1: to merge.

jwst/flatfield/flat_field.py
```diff
@@ -405,8 +397,21 @@ def do_NIRSpec_flat_field(output_model,
         interpolated_flats = None
 
     any_updated = False
+    if exposure_type == "NRS_MSASPEC":
+        slits = nirspec.get_open_slits(output_model)
     for (k, slit) in enumerate(output_model.slits):
         log.debug("Processing slit %s", slit.name)
```
philhodge: Most of the examples in the documentation for logger, even for Python 3, use the old % syntax with a comma-separated list of arguments to the log.debug function (and log.info, etc).  The doc says this is for backward compatibility, but I read somewhere that this style is used because it allows logger to use any formatter.  Note that this isn't the same as this kind of formatting:  "abc%s" % "def"; instead, it's %s (%d, etc) in the string, but the values are passed as arguments.

jwst/flatfield/flat_field.py
philhodge: `slits` is the output of `nirspec.get_open_slits(output_model)`.  I need more information than is present in the `slit` object, which was gotten by `for slit in output_model.slits` (not exactly, but that's the idea).

PR #391, opened 2016-10-28
Baseline version of velocity aberration script

hbushouse:
Committing initial version of script to compute velocity aberration scale factor and populate the appropriate FITS keyword, to be applied by SDP to level-1b products. This version uses the RA_REF and DEC_REF values in the image as the "target" location. This could be changed to something else, like TARG_RA/TARG_DEC, if desired. It obviously assumes that RA_REF and DEC_REF have already been populated.

This script just wraps the routine "vel_scale.py" that @philhodge wrote.

### DIFF
--- a/scripts/set_velocity_aberration.py
+++ b/scripts/set_velocity_aberration.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python
+
+# Copyright (C) 2010-2011 Association of Universities for Research in Astronomy (AURA)
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+
+#     1. Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+
+#     2. Redistributions in binary form must reproduce the above
+#       copyright notice, this list of conditions and the following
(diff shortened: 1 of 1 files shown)


nden:
I guess this needs to be added to the WCS. 
Build 7.1?

hbushouse:
The keywords that it populates, DVA_RA, DVA_DEC, VA_SCALE, have always been in the schema and headers, they just weren't being populated. And there's no plan, at least for now, to actually apply them to anything, but rather just have them in the header as info. Although if I remember correctly, I think the various incarnations of [astro/multi/py]drizzle did at least apply the scale factor when computing/building the WCS for an image? I guess that would happen within the resample step? But yes, anything along those lines is definitely B7.1 (not B7).

philhodge:
Are DVA_RA and DVA_DEC useful? Those values are the actual offsets in RA 
and Dec, not differential offsets.  The differential offsets between the 
guide stars and the target will be much smaller.

Phil

nden:
The velocity aberration is just a scaling factor, a correction on the WCS. `VA_SCALE` is the only thing the WCS needs, I don't know if the other keywords have some informational value.

PR #398, opened 2016-10-31
MSACONFL --> MSAMETFL

### LINKED ISSUE #388: Change MSACONFL to MSAMETFL
DMS has decided that calling the MSA data file that we get from SDP the "configuration" file has too much potential to cause confusion with the form of the MSA shutter state file that is uploaded to the telescope and also known as the "configuration" file. Hence they have decided to call the version of the file that we use within DMS and cal pipeline the MSA "metadata" file.

This means changing the keyword name MSACONFL (MSA configuration file) to MSAMETFL (MSA metadata file) in the level-1b FITS products that we use as input to the cal pipeline. Hence the core data model schema needs to be updated for this keyword name change.

And of course as soon as we merge the change to the core schema we'll all need to change the name of that keyword in any test data files we're using.

nden:
Closes #388

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -322,7 +322,7 @@ properties:
           msa_configuration_file:
             title: MSA configuration file name
             type: string
-            fits_keyword: MSACONFL
+            fits_keyword: MSAMETFL
           msa_metadata_id:
             title: MSA metadata ID
             type: integer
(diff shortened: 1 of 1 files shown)


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -322,7 +322,7 @@ properties:
           msa_configuration_file:
             title: MSA configuration file name
```
hbushouse: A minor nit, but let's also change the comment portion of the FITS keyword record to "MSA metadata file name". That's what's now in the keyword dictionary and is consistent with the way this file is now being called.

PR #401, opened 2016-10-31
Add SUBARRAY values SUBSTRIP80 and MASKBSWB for PR 85469

jaytmiller:
These SUBARRAY values were not in the core schema making it appear that SDP was not populating the FITS header when in fact the cal s/w is probably suppressing the values as invalid.

This is a very simple 2 value change to the core schema,  accidentally committed to my master branch.   The comments are mostly irrelevant,  catch up pulls to STScI-JWST master.   Let me know if the comments are a significant problem.


hbushouse:
:+1:

PR #402, opened 2016-10-31
Mod extract 1d

philhodge:
@hbushouse or @nden can you take a look at this?


hbushouse:
All the rest looks OK to me, at least I don't see any obvious issues. Real test will be in running it.

philhodge:
Thanks, I'll clean this up a bit. FLUX will be computed if there's a 
RELSENS array in the input.

Issue #187, opened 2016-08-25
Flat-field step doesn't handle ref file DQ flags properly


hbushouse:
The Build 6 testers in INS have discovered that the DQ flags from a flat-field ref file are not being handled properly within the flat-field step. This is due to the fact that the flat-field ref file - at least in some cases - is being loaded into a MultiSlitModel and this data model does NOT call the routine to perform dynamic DQ flag remapping. So the ref file DQ values are being left as is, instead of being translated via the settings in the DQ_DEF table included in the ref file.

A couple of possible solutions include:
1) switching the flat-field step back to loading ref files using a FlatModel, now that we don't have NIRSpec flats in the form of MultiSlitModel's anymore. The FlatModel data model does apply the dynamic DQ remapping.

2) Adding dynamic DQ remapping to all data models (such as MultiSlitModel), at least when a DQ_DEF table is available.

hbushouse:
Fixed in #404

Issue #250, opened 2016-09-16
Flat-field looping over integrations


hbushouse:
The flat-field step is setup to automatically loop over all integrations in an input data model, if it contains multiple integrations (i.e. if it's a CubeModel instead of a simple ImageModel). Based on the log messages, however, it looks as if the retrieval of the proper flat-field data (i.e. extracting a subarray, if necessary) is being done inside that loop, such that it's retrieving the flat data many, many times. The retrieval and setup of the flat data should only be done once. Only the application of the flat data needs to be contained with the loop over integrations.

It doesn't actually hurt anything to leave it as is, but it's just horribly inefficient.

hbushouse:
Fixed in #404

PR #404, opened 2016-10-31
Rework non-NIRSpec flat-field algorithms

hbushouse:
These updates address several issues, including #187. Much of the entire structure of handling and applying flat-field to non-NIRSpec exposures has been updated to remove a lot of the old unnecessary infrastructure that was put in place over the past 1-2 years to handle special modes of NIRSpec data. These are no longer needed, because there is a completely separate routine now for all NIRSpec data.

Specifically, non-NIRSpec flat ref files will never be in the form of CubeModel's or MultiSlitModel's - they will always be simple image-like models and hence the flat ref file can be loaded into a FlatModel, which applies the dynamic DQ mapping to the ref file data (thus fixing #187). Not having to worry about Flats in the form of MultiSlitModel's allows for the removal of a lot of old infrastructure for populating slit-related metadata and making a simple flat ImageModel look like a MultiSlitModel.

Also rewrote the apply_flat_field function so that the looping over multiple integrations in the input science data is done there, rather than in a higher-level routine. The old approach was only applying the flat data to the science data SCI array, leaving out application to the ERR array and also not propagating the DQ values from the ref file to the science data.

### DIFF
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -65,19 +65,19 @@ def do_correction(input_model, flat_model,
         if flat_suffix is not None:
             log.warning("The flat_suffix parameter is not implemented"
                         " for this mode; will be ignored.")
-        do_slit_flat_field(output_model, flat_model)
+        do_flat_field(output_model, flat_model)
         interpolated_flats = None
 
     return (output_model, interpolated_flats)
 
 #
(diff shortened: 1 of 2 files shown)


hbushouse:
Also addresses issue #250, by no longer looping over integrations at the higher-level, but rather only at the low level when the flat data are actually be applied to the science data.

PR #410, opened 2016-11-01
Multidriz model

stsci-hack:
This change adds the definition for a datamodel that supports multiple DrizProduct models in a manner similar to the MultiSlitModel datamodel.


jwst/datamodels/multiprod.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import absolute_import, unicode_literals, division, print_function
+
+from . import model_base
+from .image import ImageModel
+
+
+__all__ = ['MultiSlitModel']
```
hbushouse: Forgot to update this line.

jwst/datamodels/multiprod.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Parameters
+    ----------
+    init : any
+        Any of the initializers supported by `~jwst.datamodels.DataModel`.
+    """
+    schema_url = "multiproduct.schema.yaml"
+
+    def __init__(self, init=None, **kwargs):
+        if isinstance(init, ImageModel):
```
hbushouse: Shouldn't this be "DrizProductModel"? An ImageModel doesn't have the WHT and CON attributes that are referenced later.

stsci-hack:
This PR should now be ready for review and inclusion into calwebb_spec2 updates by @jdavies-st after cleaning up the problems identified by Howard.

hbushouse:
👍

PR #413, opened 2016-11-01
Minor edits for ami docs, added descriptions of all NIRSpec flat ref …

dmggh:
…file types


hbushouse:
👍 

It would be good to have @philhodge look over the writeups for the new NIRSpec flat ref files.

PR #422, opened 2016-11-04
Update schema for necessary SIAF keywords

hbushouse:
Two more values are needed from the SIAF to compute telescope pointing info. Those will be stored in two new header keywords VPARITY and V3I_YANG. Added these to the core schema. DP will be populating them from the SIAF entries, along with V2_REF and V3_REF.

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -826,7 +826,7 @@ properties:
           vparity:
             title: V2-V3 coordinate frame parity
             type: integer
-            default: 1.0
+            default: 1
             fits_keyword: VPARITY
           v3yangle:
             title: V3 position angle (deg)
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -813,16 +813,6 @@ properties:
             title: default spectral order
             type: number
             fits_keyword: SPORDER
-          ra_ref:
-            title: Right ascension of the reference point (deg)
-            type: number
-            default: 0.0
-            fits_keyword: RA_REF
-          dec_ref:
-            title: Declination of the reference point (deg)
-            type: number
-            default: 0.0
-            fits_keyword: DEC_REF
           v2_ref:
             title: Telescope v2 coordinate of the reference point (arcsec)
             type: number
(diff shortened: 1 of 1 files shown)


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -833,6 +823,26 @@ properties:
...
+          vparity:
+            title: V2-V3 coordinate frame parity
+            type: integer
+            default: 1.0
+            fits_keyword: VPARITY
+          v3yangle:
+            title: V3 position angle (deg)
+            type: number
+            default: 0.0
+            fits_keyword: V3I_YANG
```
nden: Are we not using `PA_V3`?

philhodge:
On 11/04/2016 09:29 AM, Nadia Dencheva wrote:
> +          vparity:
> +            title: V2-V3 coordinate frame parity
> +            type: integer
> +            default: 1.0
> +            fits_keyword: VPARITY

The default is shown as a float.

Just out of idle curiosity, what are the allowed values, -1 and +1?  Is 
the parity fixed for all instruments for JWST?

jwst/datamodels/schemas/core.schema.yaml
hbushouse: PA_V3 is one of the keyword values calculated and populated by Perry's pointing script. So it's an output rather than an input.

hbushouse:
@philhodge Yes, the allowed values are -1 and +1. I would assume the parity would be fixed for a given instrument, but don't know if it's the same for all instruments. And good catch with the default value being specified as float (my bad).

jwst/datamodels/schemas/core.schema.yaml
hbushouse: Yes, it is. I'm getting better, more accurate descriptions from the TEL folks. Stay tuned for further updates.

hbushouse:
Third time's the charm?  Updated the VPARITY and V3I_YANG comments based on accurate descriptions from Colin Cox and Matt Lallo.

hbushouse:
Long description for parity, which will be included in the keyword dictionary:

The V2/V3 coordinate system appears on the sky with the V2 axis to the left and V3 up. Thus the angle from V2 to V3 is clockwise. The Ideal coordinate system has the x axis to the right and y up. Thus the angle from x to y is anti-clockwise. The VPARITY value indicates the relative parity between these two coordinate systems. In the situation described above the parity is -1, because the V2-to-V3 and x-to-y angles are opposite from one another. OSS coordinates has x to the left and y up, hence in that case the parity is +1 (the same).

nden:
:+1:

jdavies-st:
@hbushouse Is the V3I_YANG keyword the angle of an axis of the instrument (detector) relative to the V2/V3 frame at V2_REF and V3_REF?  I.e. is it the orientation of the detector in the focal plane?  It seem to be that way as V3I_YANG is close to zero for many of the instruments, except for NIRSpec, which has its orientation offset from the others by ~45 degrees.  If so, it would be very useful for determining the orientation of the detector on the sky, and subsequently slit angle and dispersion orientation on the sky too.

PR #427, opened 2016-11-07
Fix bugs related to extending schema for cube sky matching

mcara:
This should fix most of the bugs related to expanding schemas in cube skymatching. One major issue left is that input cube's ``meta.filename`` contains cube's actual file name instead of input 2D image file name that was used to create the cube. A follow-up PR will be needed once there is an agreement on how to store these 2D file names.


mcara:
Ah, I forgot to mention that this PR also implements a workaround to a bug in ``astropy.modeling`` that is affecting ``gwcs``.

Issue #431, opened 2016-11-08
Update core schema for new FGS subarray names


hbushouse:
The FGS team has informed us that there a lot of subarray names that aren't currently in our list of allowed values in the core schema. The list of new names is:

SUB128LLCORNER
SUB128DIAGONAL
SUB128CENTER
SUB32LLCORNER
SUB32DIAGONAL
SUB32CENTER
SUB8LLCORNER
SUB8DIAGONAL
SUB8CENTER

hbushouse:
Also see INS JWST pipeline JIRA ticket JP-87: https://jira.stsci.edu/browse/JP-87

hbushouse:
Fixed by #433

PR #433, opened 2016-11-09
Updated FGS subarray names

hbushouse:
Added new FGS subarray names to core schema, at the request of the FGS instrument team.


hbushouse:
Fixes #431

PR #435, opened 2016-11-09
Update exp type

nden:
An attempt to update the list of exp_types in assign_wcs.
@hbushouse Could you take a look and suggest additions, corrections or if any of these modes should go through assign_wcs.
I want to get these correct before updating CRDS rules.

### DIFF
--- a/jwst/assign_wcs/miri.py
+++ b/jwst/assign_wcs/miri.py
@@ -313,5 +313,7 @@ exp_type2transform = {'mir_image': imaging,
                       'mir_mrs': ifu,
                       'mir_flatmrs': not_implemented_mode,
                       'mir_flatimage': not_implemented_mode,
-                      'mir_dark': not_implemented_mode
+                      'mir_flat-mrs': not_implemented_mode,
+                      'mir_flat-image': not_implemented_mode,
+                      'mir_dark': not_implemented_mode,
                       }
--- a/jwst/assign_wcs/miri.py
+++ b/jwst/assign_wcs/miri.py
@@ -310,5 +310,8 @@ exp_type2transform = {'mir_image': imaging,
                       'mir_coroncal': imaging,
                       'mir_lrs-fixedslit': lrs,
                       'mir_lrs-slitless': lrs,
-                      'mir_mrs': ifu
+                      'mir_mrs': ifu,
+                      'mir_flatmrs': not_implemented_mode,
+                      'mir_flatimage': not_implemented_mode,
+                      'mir_dark': not_implemented_mode
                       }
--- a/jwst/assign_wcs/miri.py
+++ b/jwst/assign_wcs/miri.py
@@ -12,7 +12,7 @@ import gwcs.coordinate_frames as cf
 from gwcs import selector
 from . import pointing
 from ..transforms import models as jwmodels
-
+from .util import not_implemented_mode
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
--- a/jwst/assign_wcs/nircam.py
+++ b/jwst/assign_wcs/nircam.py
@@ -48,14 +48,15 @@ def imaging_distortion(input_model, reference_files):
 
 
 exp_type2transform = {'nrc_image': imaging,
-                      'nrc_slitless': not_implemented_mode, #WFSS mode
+                      'nrc_grism': not_implemented_mode,
                       'nrc_tacq': imaging,#       ?? distortion
                       'nrc_taconfirm': imaging,
                       'nrc_coron': not_implemented_mode,#    ?? distortion
                       'nrc_focus': imaging,#       ?? distortion
-                      'nrc_tss': not_implemented_mode,# custom SOSS like mode TBC
                       'nrc_tsimage': imaging,# custom soss like mode (TBC
                       'nrc_tsgrism': not_implemented_mode,
                       'nrc_led': not_implemented_mode,#    ?? WFSS mode
-                      'nrc_dark': not_implemented_mode
+                      'nrc_dark': not_implemented_mode,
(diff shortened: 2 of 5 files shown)


jwst/assign_wcs/miri.py
```diff
@@ -310,5 +310,10 @@ def alpha_beta2XanYan(input_model, reference_files):
                       'mir_coroncal': imaging,
                       'mir_lrs-fixedslit': lrs,
                       'mir_lrs-slitless': lrs,
-                      'mir_mrs': ifu
+                      'mir_mrs': ifu,
+                      'mir_flatmrs': not_implemented_mode,
+                      'mir_flatimage': not_implemented_mode,
+                      'mir_flat-mrs': not_implemented_mode,
+                      'mir_flat-image': not_implemented_mode,
```
hbushouse: mir_flat-mrs and mir_flat-image are obsolete (superseded by mir_flatmrs and mir_flatimage), but I guess it doesn't hurt to leave them in there for backwards compatibility. Otherwise this list is good and complete. And the assignments look good too - there's no reason to have WCS coverage for flats and darks.

jwst/assign_wcs/nircam.py
```diff
@@ -48,11 +48,15 @@ def imaging_distortion(input_model, reference_files):
...
+                      'nrc_tacq': imaging,
+                      'nrc_taconfirm': imaging,
+                      'nrc_coron': not_implemented_mode,
+                      'nrc_focus': imaging,
+                      'nrc_tsimage': imaging,
+                      'nrc_tsgrism': not_implemented_mode,
+                      'nrc_led': not_implemented_mode,
+                      'nrc_dark': not_implemented_mode,
+                      'nrc_flat': not_implemented_mode,
+                      'nrc_wfsc': not_implemented_mode,
```
hbushouse: The only outlier I see here is nrc_wfsc. I don't have that in any of my lists of NIRCam exp_types. I think wavefront sensing and control (WFSC) exposure are only identified via another flag in their meta data, not by exp_type (probably have regular NRC_IMAGE). For the assignments, all the imaging entires look good. I'm not sure if NRC_CORON would also use imaging. Appropriate to have flat, led, and dark not implemented. Eventually NRC_GRISM and NRC_TSGRISM will be implemented, but they haven't been yet.

jwst/assign_wcs/niriss.py
```diff
@@ -138,7 +138,12 @@ def imaging_distortion(input_model, reference_files):
...
-                      'nis_wfss': not_implemented_mode,#        ?? WFSS spec
+                      'nis_wfss':
                       'nis_soss': niriss_soss,
-                      'nis_ami': not_implemented_mode#        ?? imaging
+                      'nis_ami': not_implemented_mode,
+                      'nis_tacq': imaging,
+                      'nis_taconfirm': imaging,
+                      'nis_focus': imaging,
+                      'nis_dark': not_implemented_mode,
+                      'nis_lamp': not_implemented_mode,
```
hbushouse: This list is complete. nis_wfss doesn't have an entry. It's not implemented yet, but will be in B7.1. I'm guessing that nis_ami would be assigned to imaging, but we should confirm with the NIRISS team.

jwst/assign_wcs/nirspec.py
```diff
@@ -1224,8 +1224,9 @@ def nrs_ifu_wcs(input_model):
...
                       'nrs_mimf': imaging,
                       'nrs_bota': imaging,
-                      'nrs_autoflat': not_implemented_mode, #TBD
-                      'nrs_autowave': not_implemented_mode, #TBD
+                      'nrs_autoflat': not_implemented_mode,
+                      'nrs_autowave': not_implemented_mode,
                       'nrs_lamp': slits_wcs,
-                      'nrs_brightobj': slits_wcs
+                      'nrs_brightobj': slits_wcs,
+                      'nrs_dark': not_implemented_mode,
```
hbushouse: This list appears complete. I would think that NRS_AUTOWAVE could probably go through WCS assignment, but the problem is that it's used for slits and ifu, so which one would you assign it to in this list?

hbushouse: Where's the list for FGS? The valid exp_types for it are:

FGS_IMAGE
FGS_FOCUS (same as imaging)
FGS_SKYFLAT (same as imaging)
FGS_INTFLAT (not_implemented)
FGS_DARK (not_implemented)

jwst/assign_wcs/nircam.py
nden: `NRC_WFSC` is also in the core schema. Should it be removed?

jwst/assign_wcs/niriss.py
nden: oops, will add it.

jwst/assign_wcs/nirspec.py
nden: I got that already in the repo. Will correct fgs_intflat.

Thanks!

hbushouse: I suppose, theoretically, INTFLAT could also be considered the same as imaging, as we're doing for SKYFLAT, but for several of the other instruments we've relegated all the flats to the not_implemented category. I think that most of the time the level-2b pipeline won't be run on flats and hence they won't be going through assign_wcs, but if someone were to ever try to run a flat through assign_wcs it could be useful to have it work (same as imaging).

jwst/assign_wcs/nircam.py
hbushouse: Confirmed with Daryl Swade that NRC_WFSC is NOT a legal exp_type value. So it should at least be removed here and would be good to remove from core schema as well.

PR #444, opened 2016-11-10
Refactored step_from_cmdline() in stpipe

jaytmiller:
to support determining the reference types that would be requested by a dataset with a given parameter set if it was calibrated.   The refactored functions don't require a dataset FITS file in order to instantiate a Step corresponding to a pipeline .cfg file.


jaytmiller:
The most recent Travis run corresponds to a docstring change.   The preceding run passed so if this gets cancelled due to time constraints it should probably be graded as "passing" and merged during some non-critical window.

Issue #371, opened 2016-10-21
S_REFPIX status for subarray processing


hbushouse:
A user reported that when they ran the level-2a pipeline on a subarray expsoure the refpix step produced log info saying that the step does not support processing of subarrays and aborted/skipped the step. But the output file still showed S_REFPIX='COMPLETE'. The refpix step needs to be updated to set S_REFPIX='SKIPPED' when the step is skipped for subarrays.

Also, it might be useful to have the step abort for subarrays earlier in the step flow. As of now it prints a lot of information to the log (parameter settings, etc.) before finally saying that it's aborting because of the subarray. That could be misleading in that at first glance it looks like the step did run.

hbushouse:
For now I've assigned this to B7.1, because it's not a show stopper, but if we happen to have time to get it fixed in time for inclusion in B7, that would be fine.

hbushouse:
Fixed by #449

PR #449, opened 2016-11-11
Updated documentation to remove reference to steps subtracting and ad…

stscirij:
…ding

back the first group for NIR data.

Step now sets S_REFPIX to 'SKIPPED' for non-IRS2 subarray data.


hbushouse:
Fixes #371

PR #454, opened 2016-11-15
Fix lrs

nden:
The changes in this PR:

- V2_REF, V3_REF, ROLL_REF, RA_REF, DEC_REF have default values and will always be present in the header.
- Removed the filter-offset correction from the spatial distortion.
- Change the name of the final output systems to be "world"
- Offset the input to the spatial transforms to account for the subarray.


nden:
Travis sems stuck again on OSX but all tests on this PR passed on my personal travis account.

PR #456, opened 2016-11-15
Resample step updated to report status as COMPLETE when finished and to

stsci-hack:
apply sky value (computed by skymatch) when resampling with drizzle.


hbushouse:
👍

Issue #438, opened 2016-11-10
Jump step crashing on ngroups=1 data


hbushouse:
The jump step is crashing on a CV3 test data set that says ngroups=4 in the header, but the actual dimensions of the data array only have 1 group of data. The step is already setup to skip processing if ngroups<=2, but it's using the value of the ngroups header keyword to determine that, which in this case causes the failure (because it doesn't match the actual data dimensions). Some of the other steps, like dark_current, key off of the actual data dimensions to get this kind of info. Jump should be updated to do the same thing, to make it more robust against failures of this type.

philhodge:
I think it should fail in this case.  How do you know which is correct, 
the keyword or the data?  It's certainly possible that the keyword is 
right and the data have been truncated.

hbushouse:
That's a good point. It would perhaps be best to have any steps that depend on knowing the value of either ngroups or nints to do a comparison of the keyword values vs the data dimensions and issue an error when they don't match. That would illuminate the existence of the problem and eliminate any other processing issues downstream.

hbushouse:
Fixed by #458

Issue #445, opened 2016-11-11
Bug in jump step y-intercept


hbushouse:
A portion of the jump.py module that is only used when executing the y-intercept method has a reference to the data model meta attribute "group_integration_time". The name of that attribute changed a long time ago in the core schema to just "group_time". The jump.py code needs to be updated to match.

hbushouse:
Fixed by #458

PR #458, opened 2016-11-16
Jump step bug fixes

hbushouse:
Updated the jump step to address two bugs, as documented in #438 and #445.

Changed the reference to "group_integration_time" to "group_time", which is the proper name as currently defined in the core schema.

Added checks to see if the ngroups keyword value is consistent with the actual data array dimensions. If not, throw an error.

Added similar checks for the values of nints and ngroups to the dq_init step, not because it relies on them, but because it's usually the first step in any pipeline processing and hence any problems with the data dimensions will be caught early.


hbushouse:
Fixes #438. Fixes #435.

PR #460, opened 2016-11-16
resample_spec for calwebb_spec2

jdavies-st:
First implementation of the resample_spec step.  Output wcs is still a bit broken.


hbushouse:
Merging because we need to get this in and all but one OSX test has completed successfully and the remaining one is just taking forever because of overall Travis OSX slowdowns.

PR #464, opened 2016-11-16
fix processing of nirspec mode NRS_BRIGHTOBJ

### LINKED ISSUE #463: Assign_wcs issue with slit_wcs compatible exp_types
As part of the recent assign_wcs overhaul for missing exp_types,  some new types were equated with slit_wcs.  There's top level logic that is still shunting those types into the "unsupported" exception since the types aren't listed along with NRS_FIXEDSLIT in get_open_slits().

nden:
Closes #463 

`NRS_BRIGHTOBJ` mode is a time series mode resttricted to `S1600A1`.

Also added processing of `NRS_LAMP` with all fixed slits opened. This has to be confirmed by the ins team and possibly changed.

### DIFF
--- a/jwst/assign_wcs/nirspec.py
+++ b/jwst/assign_wcs/nirspec.py
@@ -270,15 +270,29 @@ def get_open_slits(input_model):
 
 def get_open_fixed_slits(input_model):
     slits = []
-    slits.append(Slit('S200A1', 0, 0, 0, -.5, .5, 5))
-    slits.append(Slit('S200A2', 1, 0, 0, -.5, .5, 5))
-    slits.append(Slit('S400A1', 2, 0, 0, -.5, .5, 5))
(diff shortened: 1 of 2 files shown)


hbushouse:
I think this routine might need some additional updates to handle "normal" NRS_FIXEDSLIT exposures that are taken using a subarray readout that only encompasses one of the slits. For example, if SUBARRAY=S200A1, then that's the only slit that we want to (or can) extract. In which case the list of open slits needs to be restricted to just that one. Otherwise, if the WCS object is setup to contain information for all 5 slits, it may cause extract_2d to try to extract regions that are off the image.

hbushouse:
Or this that already taken care of down in the get_open_fixed_slits function?

nden:
@hbushouse I added code to handle single slits. Will merge after tests finish.

hbushouse:
👍

nden:
passed on Travis

PR #469, opened 2016-11-17
Include metadata in MultiProduct model

hbushouse:
Updated the datamodels/schemas/multiproduct.schema.yaml file to contain the same list of meta data definitions as the MultiSlitModel, so that those values can then be copied over from a MultiSlitModel to a MultiProductModel.


hbushouse:
All Travis Linux tests have passed. Merging.

PR #479, opened 2016-11-21
JP-101, JP-100

nden:
Added checks and log messages for missing keywords.


nden:
Verified that the data sets failing in DMS now pass.

PR #481, opened 2016-11-21
Photom bug fix and wavelength updates

hbushouse:
Two updates to the photom step code.

1) Treat NRS_BRIGHTOBJ exposures the same as NRS_FIXEDSLIT. Previously it was defaulting to treating them like IFU and MSA exposures, which uses a slightly different photom ref table format.

2) Check the wavelength array read from the photom reference file and convert the values from meters to microns, when necessary. Eventually I want to find a way to then set the TUNIT keyword for the wavelength column in the output RELSENS extension, but can't seem to do that right now when creating the RELSENS data table from scratch.

### DIFF
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -93,7 +93,7 @@ class DataSet(object):
 
         # Handle fixed-slit exposures separately
         if (isinstance(self.input, datamodels.MultiSlitModel) and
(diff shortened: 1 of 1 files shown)


hbushouse:
Fixes JP-106 https://jira.stsci.edu/browse/JP-106

PR #485, opened 2016-11-28
fix undefined variable

nden:
Depending on the compination of filter and grating a Nirspec IFU observation may fall completely outside the detector in which case the domain is also outside the detector. The flat fielding step should skip these kind of observations. 
The assign_wcs step should probably also do something about it so that the flat field step may be skipped completely even before starting.


nden:
@philhodge Could you check if the changes make sense.

philhodge:
On 11/28/2016 12:33 PM, Nadia Dencheva wrote:
>
> @philhodge <https://github.com/philhodge> Could you check if the 
> changes make sense.
>

Yes, your changes look fine to me.  Thanks!

PR #490, opened 2016-11-29
Fall off detector

nden:
This PR implements a check that an open slit/slice is projected on one of the NIRSpec detectors. If it does not fall within a detector it is removed from the list of open slits. If there's no open slit that falls on the detector the assign_wcs step is set to "SKIPPED". 
This works by running the models from the MSA to the detector for each open slit and examining the domain. If the domain shows the slit does not fall on the detector, the slit is removed.

Extract2d was changed to get the list of open slits from one of the transforms in the WCS pipeline ( a cludge but will work for build 7). Since flat field and resample work with outputs fromextract_2d, they don't need to be modified. They also don;t need to check the domain now (flat field may be modified to not check the domain). For IFU observations this also works because either all slices fall on the detector or none (in which case assign_wcs is set to "SKIPPED" altogether).

@hbushouse @philhodge @jdavies-st

### DIFF
--- a/jwst/assign_wcs/assign_wcs.py
+++ b/jwst/assign_wcs/assign_wcs.py
@@ -32,13 +32,13 @@ def load_wcs(input_model, reference_files={}):
     # Make the copy after the WCS pipeline is created in order to pass updates to the model.
     if pipeline is None:
         input_model.meta.cal_step.assign_wcs = 'SKIPPED'
-        log.info("SKIPPED assign_wcs")
+        log.warning("SKIPPED assign_wcs")
         return input_model
     else:
         output_model = input_model.copy()
         wcs = WCS(pipeline)
         output_model.meta.wcs = wcs
         output_model.meta.cal_step.assign_wcs = 'COMPLETE'
-        log.info("COMPLETE assign_wcs")
+        log.info("COMPLETED assign_wcs")
     return output_model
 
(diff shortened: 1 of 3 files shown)


nden:
Need to write many tests for this - essentially create a table of possible configurations and write a test for each one.

jwst/assign_wcs/assign_wcs.py
```diff
@@ -30,14 +30,15 @@ def load_wcs(input_model, reference_files={}):
     pipeline = mod.create_pipeline(input_model, reference_files)
     # Initialize the output model as a copy of the input
     # Make the copy after the WCS pipeline is created in order to pass updates to the model.
-    output_model = input_model.copy()
     if pipeline is None:
-        output_model.meta.cal_step.assign_wcs = 'SKIPPED'
+        input_model.meta.cal_step.assign_wcs = 'SKIPPED'
+        log.info("SKIPPED assign_wcs")
```
hbushouse: Might be good to turn this into a warning, as opposed to just info.

jwst/assign_wcs/assign_wcs.py
```diff
@@ -30,14 +30,15 @@ def load_wcs(input_model, reference_files={}):
...
+        return input_model
     else:
+        output_model = input_model.copy()
         wcs = WCS(pipeline)
-        #if output_model.meta.exposure.type.lower() in ['nrs_fixedslit', 'nrs_msaspec']:
-            #record_open_slits(wcs)
         output_model.meta.wcs = wcs
         output_model.meta.cal_step.assign_wcs = 'COMPLETE'
+        log.info("COMPLETE assign_wcs")
```
hbushouse: Minor nit, but log messaged should be "COMPLETED" (past tense).

nden:
One conda build error on OSX, everything else passing.

PR #494, opened 2016-12-01
Update FCSRLPOS keyword data type

hbushouse:
Level-1 data processing is failing because the values coming from the database for the FCRLPOS keyword are in the form of a string, but the keyword is currently defined as integer type. So updated the type to string, in order to allow schema validation to succeed.


hbushouse:
https://jira.stsci.edu/browse/JWSTKD-40

hbushouse:
Most Travis tests have passed. Can't wait any longer for rest to finish.

PR #501, opened 2016-12-02
Updated pointing quaternion data

stscieisenhamer:
NOTE: This set of data matches that provided by W.Kinzel and
is correct V1/V2/PA_V3 calculation.


stscieisenhamer:
Oh travis, you are so cute. enough success. merging.

PR #503, opened 2016-12-06
Full implementation of MIRI LRS resampling

jdavies-st:
Sample dataset jw00035001001_01101_00001_MIRIMAGE_rate.fits now runs through calwebb_spec2 to the end without crashing.  Still seeing a bug in he pixmap mapping which is passed to cdrizzle.tdriz().


jdavies-st:
The current version of the code still produces the wrong output wcs in the resampled frame.  The projection of the slit on the sky is rotated ~46 deg relative to the distorted wcs of the input frame.  This still needs to be addressed.

PR #529, opened 2016-12-12
Update superbias to check ref file subarray

hbushouse:
Updated the superbias step to do a more thorough check of computed slice indexes when trying to extract a subarray from the reference file data. There were instances of bad ref files or bad SUBSTRTn values leading to attempts to slice outside the boundaries of the reference data array.


hbushouse:
Merging inspite of Travis failures (they're all astropy/fits related).

PR #530, opened 2016-12-12
Revert calspec2 for MIRI LRS fixedslit

hbushouse:
Reverted calwebb_spec2 to NOT call resample_spec for MIRI LRS fixed-slit exposures, because output is still currently garbage or blank. By skipping the step, the unresampled data are fed downstream to extract_1d, so that it can produce reasonable output for the extracted spectrum.


hbushouse:
Can't wait for Travis. Tested it on my Linux and OSX systems. Merging now.

PR #535, opened 2016-12-13
removed a few print statements

jemorrison:


hbushouse:
So now that we're back to a state with no checking for exposures that don't have data on them, what's to prevent errors from occurring when we send in an exposure that doesn't have slices in it? Or is there a check upstream, like in assign_wcs, that stops it before getting to cube_build?

hbushouse:

PR #536, opened 2016-12-13
Superbias patch and version updates

hbushouse:
Updated the superbias step to allow it to handle the current batch of NIRCam superbias ref files that don't have SUBSTRTn and SUBSIZEn keywords in them. If the ref data are full-frame, use default values for these keywords.

Also updated the version numbers of a few steps that've been updated in rc6.


hbushouse:

PR #537, opened 2016-12-13
WCS documentation

brechmos-stsci:
This PR should not be merged as yet, it is a placeholder for some work in conjunction with @nden .


hbushouse:
Speaking of not merging - we're getting ready to put together a patch release for B7rc6, containing only those fixes absolutely necessary to fix rc6, with no other new material. So everyone please hold on merging anything that isn't absolutely necessary to fix rc6. We don't want to introduce any more new errors.

Issue #429, opened 2016-11-08
flat_field should not fail with NIRSpec image data


philhodge:
For NIRSpec, the flat_field step will work with spectroscopic data but not with image data.  There are currently no reference files for NIRSpec image data, and the reference file formats currently in use are for wavelength-dependent flats.  The flat_field step should be modified so that it does not fail when given NIRSpec image data; instead, a warning should be logged, and the calibration step keyword should be set to "SKIPPED".

philhodge:
A new issue will be filed for changes to actually do flat fielding for NIRSpec image data.

PR #538, opened 2016-12-13
Put back the changes for issue #429.

philhodge:
This checks for the case that for NIRSpec data, EXP_TYPE is not one of the supported types, and if it isn't supported the calibration switch will be set to SKIPPED and a copy of the input returned.  This is primarily to trap NRS_IMAGE or similar types.

### DIFF
--- a/jwst/flatfield/flat_field_step.py
+++ b/jwst/flatfield/flat_field_step.py
@@ -63,6 +63,16 @@ class FlatFieldStep(Step):
 
         # Retrieve the reference file name or names
         if is_NIRSpec:
+            exp_type = input_model.meta.exposure.type
+            if exp_type not in ["NRS_BRIGHTOBJ", "NRS_FIXEDSLIT",
+                                "NRS_IFU", "NRS_MSASPEC"]:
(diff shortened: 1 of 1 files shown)


jwst/flatfield/flat_field_step.py
```diff
@@ -63,6 +63,16 @@ def process(self, input):
 
         # Retrieve the reference file name or names
         if is_NIRSpec:
+            exp_type = input_model.meta.exposure.type
+            if exp_type not in ["NRS_BRIGHTOBJ", "NRS_FIXEDSLIT",
+                                "NRS_IFU", "NRS_MSASPEC"]:
```
nden: Does "NRS_LAMP" need to be included in this list?

philhodge: There is currently no flat-field reference file for "NRS_LAMP", so if we included that in the list of allowed exposure types, the step would fail at the point of trying to get the reference file.  We don't want the step to fail, we just want to set S_FLAT to "SKIPPED" and return the input.

hbushouse:

Issue #473, opened 2016-11-17
Need TUNIT keywords for table extensions


hbushouse:
The photom reference tables contain a column of wavelengths, which are in turn copied into the RELSENS extensions that get added to science products by the photom step. The tables in the photom ref files and the RELSENS extension should have TUNIT keywords so that the code can determine the units for the wavelengths (as opposed to guessing based on the values).

bernie-simon:
This problem was fixed way back with pull request #540

PR #540, opened 2016-12-14
Update TUNIT when column units are changed

bernie-simon:
It was reported that the TUNIT keyword was not being updated when a table was opened using the datamodels package and a column's units was changed. The problem was in fits_support._fits_array_loader. That code will convert the datatypes of the table columns when they do not agree with the schema. That conversion involves a copy and the copy loses the listeners which update the header keywords when a change is made to a column's attributes. Code was added to copy the listeners from the original data to the copy of the data so that column units are
updated.


bernie-simon:
This change partially addreses issue #473 "Need TUNIT keywords for table extensions." raised by @hbushouse  There may need to be another fix addressing reading the TUNIT keyword through the datamodels interface, I'm not sure yet.

Issue #543, opened 2016-12-16
IFU Cube_Build do not produce valid cube for NIRSPEC NRS2 


jemorrison:
When determining the size of the final 3D spectral cube on the sky, the code loops over every slice on the detector and checks that the x,y values bounds on the detector are > 0 . However x = 0 is a valid value of a pixel in a slice since the indexing starts at 0 and not 1.  Most of the Filter/Gratting combinations for NRS1 have slices that start a x> 1, but for NRS2 all the slices start at x = 0. 
Thus for NRS2 the cube_build code thinks there are no valid slices and returns a slice of 0 (for the wavelength dimension).

hbushouse:
Fixed by #557.

PR #557, opened 2016-12-21
Nirspec noifudata

jemorrison:
Updated Cube Build Documentation
Updated Straylight Documenation
Fixed an error with NIRSPEC Issue#543. The x,y bounds of the slice boundaries are indexed starting at 0. The code was assuming the indexing started at 1. For NIRSPEC 2 the x boundary always starts at 0 so the code thought no valid data was falling on the detector. 
Also updated the weighting function to be more inline with what David Law was doing. An input parameter on the power of the distance between point cloud member and spaxel center was added.


hbushouse:
Fixes #543.

PR #560, opened 2016-12-30
Prevent srctype step from returning closed model

hbushouse:
When running the srctype step standalone, it's possible in some circumstances for the input model to get modified in place, without a new model being created for output. The use of a "with datamodels.open" block was causing the updated input_model to be closed by the time it was returned, resulting in an error when stpipe tries to save the returned model to disk. Removing the "with" block and just opening the input model directly allows it to stay open for the return in those circumstances where it's been modified in-place.


hbushouse:
Also update core schema to include S_SRCTYP keyword for recording the status of the source type step.

PR #581, opened 2017-01-10
Two Point Difference Rewrite

mwregan2:
This routine uses pytest to run a suite of test for
twopoint_differene.py. They focus on simple test that deal with problem
areas associated with small numbers of groups, saturation, and the bias
to the median introduced by a cosmic ray.


jwst/jump/Tests/test_twopoint_difference.py
```diff
@@ -0,0 +1,431 @@
+
+from twopoint_difference import find_CRs
```
hbushouse: In the jwst package we need imports to be relative, which means this needs to be:

from ..twopoint_difference import find_CRs

to get the import to succeed. Right now it's failing in the Travis tests.

mwregan2:
Howard,
  Well, I forgot to make the change. I had it in the same directory as two point_difference.
  Now, I’ve got some combination of not knowing the right way to import and having a messed up environment, I can’t even get the tests to run. I’ll work on it some more.
Sorry
Mike

jwst/jump/twopoint_difference.py
```diff
@@ -9,31 +9,30 @@
...
 This is MUCH faster than doing all the work on a pixel-by-pixel basis.
-'''
+"""
 
 import logging
 import numpy as np
 
-from ..datamodels import dqflags
-from .. import datamodels
+import dqflags
```
hbushouse: For use in the package structure this will need to be:

from ..datamodels import dqflags

so that the dqflags module is imported from the correct location within the package.

jwst/jump/twopoint_difference.py
```diff
@@ -53,18 +52,22 @@ def find_CRs(data, gdq, read_noise, rej_threshold, nframes):
...
-        # don't cause trouble in later calculations
-        nans = np.where(np.isnan(med_diffs))
-        med_diffs[nans] = 0.
-        first_diffs[nans] = 0.
-
         nans = np.where(np.isnan(first_diffs))
-        first_diffs[nans] = 0.
+        first_diffs[nans] = 100000.
+
+        print(repr(first_diffs[100,100,:]))
```
hbushouse: All print statements will need to either be removed for use within the pipeline environment or converted to logging calls. If it's useful to keep the statements for debugging work, then they can be converted to calls to "log.debug" so that they appear only when run with the logging level set to debug by the user.

mwregan2:
Howard,
  I don’t like asking you to debug my problems but I’m really confused about the packaging. I now get the error message:
../__init__.py:1: in <module>
    from .version import *
E   ImportError: No module named 'jwst.version'
I get this for any package that I try to run pytest in. I feel like I’m missing something in my environment.

  Have you seen this before?
Thanks

Mike

jhunkeler:
@mwregan2 In your jwst source directory run `python setup.py build` at least once to generate the missing `.version` module.

hbushouse:
This new version of the twopoint_difference module and the new unit tests module have been merged into jwst master in #748 (avoids the tedium of resolving conflicts on this branch). Closing this PR. Thanks Mike for your first direct contribution to the JWST pipeline!

PR #591, opened 2017-01-13
Removed statements which erroneously set some nan values to 0 during …

dmggh:
…jump detection. This fixes the problem noted by Mike Swam during testing in November 2016, in which the execution time of the jump step was nearly 2 hours, and resulted in certain reads (including saturated pixels) being erroneously flagged as CRs. This update fixes these problems (decreasing the execution for that dataset to less than 1 minute). @hbushouse


hbushouse:
👍

Issue #610, opened 2017-01-18
Task: Flat-fielding for NIRSpec imaging exposures


hbushouse:
The new NIRSpec flat-fielding scheme implemented in Build 7 did not include any methods (or reference files) for correcting imaging-mode exposures, like target-acqs. This capability needs to be implemented for Build 7.1.

philhodge:
This can be done as follows:  In flat_field_step.py, add an ``IMAGING_MODES`` list to specify all known values of EXP_TYPE that indicate NIRSpec imaging data rather than spectrographic (see ``exp_type2transform`` in assign_wcs/nirspec.py).  Replace the ``is_NIRSpec`` flag with ``is_NRS_spectrographic``.  Set the latter to True for NIRSpec spectrographic data and False otherwise.  For NIRSpec imaging data, the reftype will then be flat, rather than fflat, sflat, and dflat.  In flat_field.py, call ``do_NIRSpec_flat_field`` only for NIRSpec spectrographic data; otherwise, call ``do_flat_field``.

hbushouse:
Fixed by #634.

PR #628, opened 2017-01-20
Handle NIRSpec imaging data.

philhodge:
For NIRSpec imaging data (identified by a new IMAGING_MODES list), reference
file type flat will be used, rather than fflat, sflat, and dflat.  Such data
will then be processed the same as non-NIRSpec data.  See issue #610.

### DIFF
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -34,7 +34,7 @@ def do_correction(input_model, flat_model,
 
     flat_model: JWST data model
         Data model containing flat-field for all instruments other than
-        NIRSpec.
+        NIRSpec spectrographic data.
 
     f_flat_model: NirspecFlatModel or NirspecQuadFlatModel object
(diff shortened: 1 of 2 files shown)


philhodge:
Do I need to put back the check for unsupported modes (nrs_autoflat, nrs_autowave, nrs_lamp, nrs_dark), or will flat_field never be called for any of these modes?

hbushouse:
A dark will never be used as input to either calwebb_spec2 or the flat-field step, at least not in the automated/operational pipeline, but of course that doesn't stop a user from doing it on their own. We also don't know at this time just what processing needs to be applied to the nrs_autoflat, nrs_autowave, and nrs_lamp types. I believe they will be dispersed, and hence could conceivably be input to calwebb_spec2, but we just don't know how to handle them yet. And even if Nadia has implemented checks upstream in assign_wcs to abort processing for these types, that again doesn't prevent a user from manually sticking one into the flat-field step by itself. So I would say keep as many checks in as many places as possible to try to make them "idiot proof".

PR #637, opened 2017-01-24
Initial implementation of new step EngDBLogStep

### LINKED ISSUE #600: Task: DMS-458 (retrieve EngDB parameters)
DMS Level 4 requirements 256 and 458 state that the calibration pipeline must access the Engineering Database (EngDB) and retrieve values of engineering mnemonics from it, to use during calibration processing. Given that the cal pipeline currently does not need any engineering parameters to perform calibration, but waiving, removing, or putting a lien on a Level 4 requirement requires and wastes so much time and effort, we will therefore create a new calibration step ("dms458") that will do nothing but retrieve some parameter value from the EngDB and print its value in a log message. This cal step will not be used in any of our current pipelines. It will simply exist for the purpose of I&T to run it from the command line and verify requirements 256 and 458.

stscieisenhamer:
Resolves #600

### DIFF
--- a/jwst/pipeline/engdblog.cfg
+++ b/jwst/pipeline/engdblog.cfg
@@ -1,0 +1,2 @@
+name = "engdblog"
+class = "jwst.engdblog.EngDBLogStep"
--- a/jwst/pipeline/engdblog.cfg
+++ b/jwst/pipeline/engdblog.cfg
@@ -1,2 +1,6 @@
 name = "engdblog"
 class = "jwst.engdblog.EngDBLogStep"
+
+stime = "2016-01-01"
+etime = "2016-01-31"
+verbosity = "initial"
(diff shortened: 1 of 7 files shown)


hbushouse:
Should also create a engdblog.cfg file to go with the step, which gets stored in the "pipeline" directory of the tree along with all the other .cfg files.

jwst/pipeline/engdblog.cfg
```diff
@@ -0,0 +1,2 @@
+name = "engdblog"
+class = "jwst.engdblog.EngDBLogStep"
```
hbushouse: The default .cfg file in the repo should also include the list of optional arguments (the items in the "spec" list in the step module) set to their default values. That way a user can easily modify the argument settings if they want. For example: https://github.com/STScI-JWST/jwst/blob/master/jwst/pipeline/ramp_fit.cfg

stscieisenhamer: That's what I thought should be, but I had looked through a number of them, and found that many did not have anything, so I went with less. Will put them in.

Also, wasn't sure what stspec would give, but since that is [failing at the moment](#649), could not find out. I'm actually going to take that up momentarily.

PR #638, opened 2017-01-24
Change clobber to overwrite; Astropy 1.3 depreciation

jdavies-st:
The keyword `clobber` has been depreciated in Astropy 1.3 for io.fits.  The keyword `overwrite` is now preferred.  This PR updates all instances in the jwst package.

### DIFF
--- a/jwst/assign_wcs/tools/nirspec/compute_world_coordinates.py
+++ b/jwst/assign_wcs/tools/nirspec/compute_world_coordinates.py
@@ -84,7 +84,7 @@ def ifu_coords(fname, output=None):
     else:
         root = model.meta.filename.split('_')
         output = "".join([root[0], '_world_coordinates', '.fits'])
-    hdulist.writeto(output, clobber=True)
+    hdulist.writeto(output, overwrite=True)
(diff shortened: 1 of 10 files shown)


hbushouse:
Got an email from Steven Beard (MIRI IDT) pointing out the need to do this. He'll need to coordinate a similar change in the MIRI team's code base, so we should hold off just a bit on merging this until we get confirmation from him that it's an OK time to do so.

jdavies-st:
Ok, I will hold off on merge.  That said, the **kwargs for datamodels are still passed through to the underlying astropy.io.fits methods.  So if their code uses `clobber` when writing out a datamodel, then it will still get passed to fits.write() and they will get the same astropy depreciation warning.  The changes above just guarantee that our code is compliant in datamodels and stpipe (and all the tests).

hbushouse:
Just got a message from Steven that says we can go ahead and merge on our side, at which point he'll make the necessary update on his side. In the meantime, his users are using a "static" version of the code, so it won't affect them during the short amount of time that the two sets of development code are out of synch.

PR #642, opened 2017-01-25
Python 3 compatibility regarding exception message

philhodge:


hbushouse:
Is this change backwards compatible with python 2.7, while we continue to use it?

philhodge:
Yes.  I tried this test using Python 2.7.5:

    x = 37.
    try:
        y = x / 0.
    except ZeroDivisionError as e:
        if "division" in str(e):
            print("yes, division by zero")
        else:
            print("no, something else")
   
    yes, division by zero

philhodge:
By the way, there's another problem with `stpipe`:

      File "/user/hodge/jwst/git/jwst/jwst/stpipe/cmdline.py", line 324, in step_from_cmdline
        if debug_on_exception:
    NameError: name 'debug_on_exception' is not defined

`debug_on_exception` is defined in `just_the_step_from_cmdline` but not in `step_from_cmdline`.  This is unrelated to the change I just made.

hbushouse:
I've been getting that NameError recently too, whenever I run a step or pipeline from the command line and it encounters some kind of error (e.g. input file can't be found). This seems to have appeared only recently.

hbushouse:
The only Travis failure was for the PEP8 style check, so I say go ahead and merge. 👍

PR #643, opened 2017-01-25
Remove yield tests

stscieisenhamer:


stscieisenhamer:
Besides the for-some-reason-now-required doc test failure, a "job running too long" failure was the only culprit which, given that all other platforms succeeded and the failed job's platform is nearly identical to the developer's platform, this be a-merged.

PR #651, opened 2017-01-26
Allow Step.print_configspec to operate under Python3

stscieisenhamer:


nden:
I wonder if we should update to the latest configobj. v5 is out now

https://pypi.python.org/pypi/configobj

Although I haven't looked at the differences with v 4.

stscieisenhamer:
@nden commit 3c98eac includes the update, though clearly there are some minor edits that need to occur, and I believe I missed a file. Will be repairing that.

About changes in v5: according to the changelog and doing diff's, a refactor has occurred, but API-wise all appears about the same.

stscieisenhamer:
@nden Is there any reason there is a local copy of configobj? The only jwst-specific modification I could tell with the original install was the use of extern.astropy.six instead of just six.  If using this as just an external dependency works, is that ok? Or is there some other reason?

hbushouse:
Unfortunately Mike D. is probably the only person who can really answer the question as to why there's a local copy in use.

nden:
@stscieisenhamer I don't know why it was included locally (we've done that for teal as well), perhaps to guard against unexpected changes and to have some control over when it changes. It may not be a bad thing to have it locally and update it only when we have tested new versions.

stscieisenhamer:
And that is the opening for the "include or use dependencies" flamewar!

And despite the fact that that discussion should occur and would be quite fun, the actual fix is simple enough, such that what I'm going to do is redact this PR and start a new one with just the fix and I'll open another issue for the 'update configobj'.  I had only attempted the update hoping that was the fix.  But the simple fix works for the embedded version, so we'll take the "if it ain't broke, don't fix it" approach for now.

Issue #658, opened 2017-01-30
NameError: name 'debug_on_exception' i


nden:
I now have a reproducible case of this exception. 
I don't have time to look at it right now but if someone wants to the files are in 

/user/dencheva/debug-on-exception

To reproduce, run

```
strun assign_wcs.cfg filename
```
Another comment on the issue is [here](https://github.com/STScI-JWST/jwst/pull/642#issuecomment-275127432)

philhodge:
Function `just_the_step_from_cmdline` assigns a value to variable `debug_on_exception` but doesn't use it.  Function `step_from_cmdline` uses `debug_on_exception`.  A simple way to fix this would be to modify `just_the_step_from_cmdline` to return `debug_on_exception`, in addition to the other three items that it returns, since `step_from_cmdline` calls `just_the_step_from_cmdline`.  (Note that `steps_to_reftypes_from_config` also calls `just_the_step_from_cmdline`.)

jaytmiller:
I will look it over,  this is probably code I added.

hbushouse:
According to github the last time jwst.datamodels.cmdline.py was modified was in mid-November 2016. So why is this issue just rearing it's head now?

jaytmiller:
It's related to debugging;  the failure occurs when there's an exception that propagates all the way to the top level.   There's a flag to branch into pdb and I missed it when I refactored the top level,  so the flag is undefined if/when an exception occurs.   Phil's fix seems correct,  I will do a pull request momentarily.

hbushouse:
Fixed by #663

PR #663, opened 2017-01-30
Fix debug_on_exception in cmdline.py

jaytmiller:
Bug fix for top level exception handling that resulted from factoring out just_the_step_from_cmdline() and not accounting for the debug_on_exception local variable in the original step_from_cmdline() function.


nden:
Closes #658

Issue #623, opened 2017-01-19
Task: Data products for B7.1


hbushouse:
## Definition
This task consists of several main pieces of work:

- Define any new types of data products that will come out of the B7.1 cal pipeline
- Create data models for the new types of products
- Create examples of all new types of products from B7.1, where "new" is defined as a type that the rest of DMS has not yet seen (even if we may have been creating them for a while already)

The last item can be done by either doing a "one off" creation of a static example data file for each product type or by rigging up skeletons of the pipeline modules (e.g. calwebb_tso3) that will be creating those products such that they create dummy (i.e. zero-filled) output files of the correct type when the skeleton pipeline module is executed.

The rest of the downstream DMS subsystems would like to have examples of all types of B7.1 data products by mid-2017 (June-July) in order to give them time to finish their B7.1 development work.

## Resources
- [Science Data Products](https://confluence.stsci.edu/pages/viewpage.action?spaceKey=SSB&title=Science+Data+Products)

stscieisenhamer:
# Implementation
## In Development
- [ ] calwebb_coron3
- [ ] calwebb_tso3

## Future
- [ ] calwebb_spec3
- [ ] calwebb_wfss

stscieisenhamer:
The collecting of products will be in the directory
```
/grp/jwst/ssb/test_build7.1/examples_for_dms
```

hbushouse:
All data product formats, contents, and file name syntax have been defined and documented at https://confluence.stsci.edu/display/SSB/Science+Data+Products. Example files were created through the use of either actual pipeline processing or generation of fake data products, using a set of python notebooks developed by @stscieisenhamer @stsci-hack and @hbushouse. The notebooks and resulting example data products were placed in the directory tree at /grp/jwst/ssb/test_build7.1/examples_for_dms/ and made available to the other DMS subsystems for their use in testing archiving, etc.

PR #666, opened 2017-01-30
Add model names to the output product listing

stscieisenhamer:
WIP #623

### DIFF
--- a/docs/pipeline/Makefile
+++ b/docs/pipeline/Makefile
@@ -151,3 +151,7 @@ doctest:
 	$(SPHINXBUILD) -b doctest $(ALLSPHINXOPTS) $(BUILDDIR)/doctest
 	@echo "Testing of doctests in the sources finished, look at the " \
 	      "results in $(BUILDDIR)/doctest/output.txt."
+
+livehtml:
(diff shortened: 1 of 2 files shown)


stscieisenhamer:
@hbushouse Very minor thing, but started adding what the datamodels used are to the pipelines document. DIdn't find anyplace where this equivalency was done and thought this would be the place for it.

stscieisenhamer:
Error is a build timeout. Will merge when discussion is complete.

hbushouse:
This is a good idea. Very useful info for folks who want to work with the data using our models. In addition to adding the class names of the models in the text, there should also be some sort of brief introductory remark at the beginning indicating what those names are that appear in parentheses for each product.

hbushouse:
One thing that will make this listing a bit messy is that there isn't always a one-to-one correspondence between a product type and data model. For example, _cal products can be in the form of an ImageModel for both imaging and simple spectra (like MIRI LRS), while other spectral _cal's use a MultiSlitModel.

stscieisenhamer:
I presumed that was the case; exactly why such documentation would be good.

stscieisenhamer:
All errors have to do with a CRDS issue unrelated to any changes. Merging.

PR #676, opened 2017-02-01
wcs_from_footprints

nden:
The original version of this function assumed the WCS scheme is similar to the one on HST imagers.
Since this is not valid now and currently JWST does not use FITS WCS keywords, it has be modified.

The input is now a list of models. The scaling model uses CDELTn values and the rotation uses ROLL_REF.
@jdavies-st @stsci-hack This may require further tweaks. Please review.

### DIFF
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -15,6 +15,8 @@ from astropy.modeling import models as astmodels
 from gwcs import WCS
 from gwcs.wcstools import wcs_from_fiducial
 
+from . import pointing
+
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
@@ -56,8 +58,8 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, domain=None):
     If ``refmodel`` is None, the first WCS object in the list is considered
     a reference. The output coordinate frame and projection (for celestial frames)
     is taken from ``refmodel``.
-    If ``transform`` is not suplied, a compound transform is made using
-    CDELTs and rotation by ROLL_REF.
+    If ``transform`` is not suplied, a compound transform is created using
+    CDELTs and PC.
     If ``domain`` is not supplied, the domain of the new WCS is computed
     from the domains of all input WCSs.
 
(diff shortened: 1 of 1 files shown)


jwst/assign_wcs/util.py
```diff
@@ -45,67 +45,65 @@ def _reproject(x, y):
...
+    scales = astmodels.Scale(refmodel.meta.wcsinfo.cdelt1) & \
+        astmodels.Scale(refmodel.meta.wcsinfo.cdelt2)
+    trans.append(scales)
+
     if trans:
         tr = functools.reduce(lambda x, y: x | y, trans)
     else:
         tr = None
-    out_frame = refwcs.output_frame
+    rotation = astmodels.Rotation2D._compute_matrix(np.deg2rad(refmodel.meta.wcsinfo.roll_ref))
```
jdavies-st: I think `meta.wcsinfo.v3yangle` and `meta.wcsinfo.vparity` need to be used to compute this rotation angle.  `roll_ref` only gives the roll angle of V3 from celestial north, whereas we need to also take into account the angle between the y axis of the science frame and V3 which is given by `v3yangle`.  `vparity` is important here to as it gives the flip across the y axis if necessary.

Also, `rotation` here is a np.array object, not an astropy.modeling model.  It probably needs to be made into and `AffineTransformation2D` model or similar.

jwst/assign_wcs/util.py
```diff
@@ -45,67 +45,65 @@ def _reproject(x, y):
...
+    trans.append(rotation)
+    out_frame = refmodel.meta.wcs.output_frame
     wnew = wcs_from_fiducial(fiducial, coordinate_frame=out_frame,
```
jdavies-st: `tr` needs to include the rotation defined above.

jwst/assign_wcs/util.py
nden: @jdavies-st  I have modified the rotation to use the PC matrix instead of ROLL_REF. The PC matrix should account for the parity and the YIDLAngle. Let me know if that's sufficient.

nden:
@jdavies-st A new version of this function.
The transforms is constructed from the `PC` matrix of the reference image, followed by an average scale. After the domain is computed the crpix are defined as the center of the domain.

jdavies-st:
Thanks @nden.  Yes, useful!  Doing some checks with my code.  Can you wait to merge?  I have to make a change to the call to `wcs_from_footprints()` in resample as the api has changed from passing a list of wcs objects to passing a list of datamodels.  Would be nice to make the changes at the same time so we don't break `calwebb_image3`.

Issue #632, opened 2017-01-20
GLS option fails with bitwise comparison between int32 and uint32


philhodge:
In function ``gls_ramp_fit`` in ramp_fit.py, the ``temp_dq`` array is created with data type ``np.int32``, but a bitwise comparison is done with array ``dq_int``, which is ``np.uint32``.  ``temp_dq`` should be ``np.uint32`` as well.

hbushouse:
Fixed by #680.

PR #680, opened 2017-02-02
Change data type from int32 to uint32.

philhodge:
Temporary array dq_int has data type uint32, while temp_dq has data type
int32, and the code does a bitwise OR between dq_int and temp_dq.  This
used to work, but more recently (e.g. using Python 3.5.2 and numpy 1.11.3)
it fails because of the difference in data type.  Function gls_ramp_fit has
been modified to create temp_dq as data type uint32.  See #632.


philhodge:
The travis tests have not even started, and they're not going to test anything that I haven't already checked, so I'm going to merge.

Issue #669, opened 2017-01-31
Update input read noise to be in units of DN


hbushouse:
The PWG made the decision a while back to provide read noise reference files in units of DN, rather than electrons, so that whenever the instrument teams feel they need to make a change to detector gain values, it will not be necessary to compute and deliver new read noise ref files as well. See https://confluence.stsci.edu/display/JWSTPWG/JWST+Calibration+Reference+Files%3A+File+Formats+for+the+Build+6+Pipeline#JWSTCalibrationReferenceFiles:FileFormatsfortheBuild6Pipeline-ReadNoise

We need to update the jump and ramp_fit steps to multiply the input read noise image by the gain reference image, to convert the read noise data to electrons for use in the steps (the algorithms are already expecting units of electrons). We also need to update the jump and ramp_fit step docs to indicate the new expectation of units for the read noise ref file.

hbushouse:
Fixed by #681, #688, and #692.

PR #681, opened 2017-02-03
Updating jump code and documentation to use input read noise in units…

dmggh:
… of DN, in partial support of issue #669. During development I noticed that some gain values had invalid values (0, negative, or NaN), so have also updated code to flag those pixels as DO_NOT_USE and NO_GAIN_VALUE in the PIXELDQ array. Similar updates will also be made for ramp_fitting. @hbushouse


hbushouse:
The Travis failures all seem to be associated with one or more particular CRDS tests, so I think you can go ahead and merge this.

PR #685, opened 2017-02-07
Change error types that are caught when loading associations

bernie-simon:
The code in the generic datamodels.open function in datamodels/util.py did not catch IOError. Instead, it caught ValueError, which are not raised on the section of code it called. So the code was changed to catch IOError instead of Value Error.
    
Similarly, the code in from_asn in datamodels/container.py caught ValueError instead of IOError, which is thrown by the code it called. So the code was changed to catch IOError instead of ValueError.


bernie-simon:
All tests have passed, so I am merging this change to datamodels.open()

PR #697, opened 2017-02-13
[WIP] Core schema refactor

nden:
Warning: This involves some pain ...
If this PR gets accepted the pain comes from the fact that we may need to update some data files. Hopefully in most cases it will be sufficient to delete the ASDF extension and open the file with the model but in some cases we need to rerun the pipeline.

This came up in the process of writing WCS reference file schemas.

Currently all data models and all reference files inherit from the same `core.schema.yaml` schema.
This means that reference file models have entries which don't make sense for reference files, like `meta.wcsinfo`, `meta.pointing`, etc. 

In this PR `core.schema.yaml` contains only entries which are common for the data and reference files. The core schema for data models is called `core_data.schema.yaml` and the core schema for reference files is still `referencefile.schema.yaml`. Both inherit from `core.schema.yaml`.

Some side issues surfaced and need to be resolved before this is merged:

[] core_data.schema.yaml has a section `meta.ref_file` and referencefile.schema.yaml has a section `meta.reffile`. These have different purpose but close names and are confusing. I don;t see why we need the extra section in the reference file schema. We can have the entries directly in `meta`. The entries are `type`, `pedigree`, `description`, `useafter`, `author`. **Edited**: Any changes should be cleared with Todd.

[] The `image.schema.yaml` contains a section about level3 products. There's a separate `le3_product.schema` which looks like a duplicate. Consider removing the entries in the image schema.

[] Some reference files do not inherit from `referencefile.schema`.

[] Schemas don't have description. I added a title and description to the core and image schemas as examples. If all schemas are updated with this then the documentation may create a page with all schemas and an explanation of their applicability.

[] **Edited**: Should coordinate with @bernie-simon re syncing with SDP keyword dictionary.

WCS reference files and models will be in a different PR.

@STScI-JWST/maintainers Comments on this approach, the pain involved ... ?


nden:
As mentioned above, the core_data schema has a section `ref_file` which lists the names of the reference files used in the pipeline, while the reference file schema has a section `reffile` which lists a few keywords required to be present in a reference file. Is this confusing? Should we rename one of the sections and if so which one? An alternative is to remove the `reffile` section in the reference file schema and move the keywords directly into meta. For clarity the keywords are `reftype`, `pedigree`, `author`, `description`, `useafter`. Any optinions on this? (I don't mind removing the `reffile` section and recording them directly under `meta`.)

philhodge:
On 02/17/2017 03:55 PM, Nadia Dencheva wrote:
> (I don't mind removing the |reffile| section and recording them 
> directly under |meta|.)

That makes sense to me.  Is there any disadvantage?  Might we need more 
structure later?

bernie-simon:
I understand that the similarity between the names can cause confusion, but renaming reffile or moving the keywords in may affect Todd's code and should be cleared  with him.

jdavies-st:
I agree with Phil that it makes sense to remove the meta.reffile.* nodes and record them directly under meta.

And of course Bernie is right.  @jaytmiller How would this this affect your code?

hbushouse:
The main purpose of the ref_file section in the core schema is simply to create a subsection in the FITS headers that contains all the keywords pertaining to reference files and has a subsection label "Reference file information", just like all the other subsections that are defined to organize the header a bit. So I want that subsectioning to stay, although I'm open to a change in the data model name. The "reffile" section in the referencefile schema, on the other hand, doesn't really seem necessary to me and hence I also agree that it could be removed, as long as it's done in coordination with Todd and CRDS, because I'm sure he's using those datamodel names in the CRDS code. I think I'm probably the one who originally created the referencefile schema some years ago, but I honestly can't remember why all the individual keyword entries were placed into a "reffile" grouping like that.

We might also need to inform folks in MESA in the various instrument teams who probably have python code they're using to create reference files and populate those keywords. Their scripts would need to be updated too.

jaytmiller:
The affect to CRDS code,   AFAIK,  will be very minor.    From a CRDS 
perspective the main thing that makes it hackish is that it will require 
a code change,  not just an rmap change.   This part is not a big deal.

nden:
superseded by #715

Issue #611, opened 2017-01-18
Task: Implement MIRI MRS flux calibration


hbushouse:
All spectroscopic instrument modes (so far) define a wavelength-dependent spectral response function that's stored in a PHOTOM reference file and loaded by the photom pipeline step into the RELSENS table extension of the science product. The RELSENS data are then used during the extract_1d step to convert observed count rates into absolute fluxes. Flux calibration for the MIRI MRS mode, on the other hand, is being provided as a 2D array of flux conversion factors that are to be applied pixel-by-pixel to the 2D stage of MRS slice images, to convert them from count rates to fluxes.

So at minimum we need to define a new PHOTOM reference file format for the MRS 2D flux array and modify the PHOTOM step to attach this array to the science product, rather than the traditional RELSENS table extension.

We currently believe that the flux-cal information will be applied to the science data during the cube_build process, such that the output cubes from that step will be in units of physical fluxes.

hbushouse:
Copies of the new MRS photom/flux-cal reference files and documentation are available in:
file:///grp/jwst/wit/miri/CDP-6/CALSPEC2_MRS/absolute_flux_calibration/PHOTOM_Jan2017/

There is one ref file per sub-band (short, medium, long) for each detector, for a total of 6 reference files.

hbushouse:
Discussion of this approach with both the MIRI MRS and NIRSpec IFU reps has led to the realization that the NIRSpec IFU group is expecting a similar approach for their calibration, which was not known until now. The NIRSpec IFU mode will still use a photom reference file that defines a simple 1D spectral response function as a function of wavelength, but instead of just attaching that 1D RSRF to the science data as is currently done, they expect it to be used to create, on the fly, a full 2D array that gives the response of each pixel, based on its wavelength. This 2D response array then gets attached to the science product, as is the case with MIRI MRS, and is also applied to the 2D unrectified data on its way into the cube_build routine, so that in both cases cube_build is working with calibrated data that have the effects of varying pixel sizes removed. A separate issue will be filed for the NIRSpec IFU work.

philhodge:
If I understand correctly, this means that the extract_1d step should not even attempt to apply flux calibration for any IFU data.  The net and background columns will be in units of flux density, and the net will be the same as the flux.

hbushouse:
That's my understanding as well. The cubes coming out of cube_build should already be in calibrated units, for both instruments.

hbushouse:
[...]

PR #698, opened 2017-02-13
Add handling of BUNIT keyword in data extension headers

hbushouse:
Adding BUNIT keywords in the extension headers of science data models that use "data (SCI)" and "err (ERR)" arrays. Also updated the MIRI MRS portion of the photom step to update the BUNIT values after applying the new flux calibration scheme to the data (see #611 and #694).

The BUNIT keyword definitions can't belong to the core schema, because that's currently inherited by all data models, even those that don't use "SCI" and "ERR" extensions in their products. Including them in the core causes empty SCI and ERR extensions to be created in those FITS products, just so the BUNIT keyword value can be recorded. So I've put the definitions, for now, in a separate schema file ("bunit.schema.yaml"), which is then imported into the half-dozen or so science data model schemas in which they're needed.

This is a work in progress and is subject to revision, especially with regard to #697. Comments or suggestions about any aspect of this are welcome.

### DIFF
--- a/jwst/datamodels/schemas/bunit.schema.yaml
+++ b/jwst/datamodels/schemas/bunit.schema.yaml
@@ -1,0 +1,17 @@
+allOf:
+- type: object
+  properties:
+    meta:
+      type: object
+      properties:
+        bunit_data:
+          title: physical units of the array values
+          type: string
+          fits_hdu: SCI
+          fits_keyword: BUNIT
+        bunit_err:
+          title: physical units of the array values
+          type: string
+          fits_hdu: ERR
+          fits_keyword: BUNIT
+$schema: http://stsci.edu/schemas/fits-schema/fits-schema
--- a/jwst/datamodels/schemas/cube.schema.yaml
+++ b/jwst/datamodels/schemas/cube.schema.yaml
@@ -1,5 +1,6 @@
 allOf:
 - $ref: core.schema.yaml
+- $ref: bunit.schema.yaml
 - type: object
   properties:
     data:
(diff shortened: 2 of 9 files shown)


nden:
This looks good to me. When you merge it I'll pick up the changes in #697.

hbushouse:
I like the idea of having the model/keyword definitions in a separate schema file, so that it's easy to include in any data model schema that needs them, but these are specifically tied to keywords that appear in SCI and ERR extensions. There may be some types of products, such as resampled products, that use extensions with different names that might also need BUNIT keywords, in which case you can't just include bunit.schema.yaml in those models. So I'm wondering if it might be more obvious to the user/developer if these definitions were in fact copied (repetitively) into each model schema that needs them. At least that way it's obvious when looking at the model schema exactly what you're getting. And you can easily then define, say, one BUNIT keyword for a SCI extension and another for an extension of some other name.

nden:
Can we use enumerate? I haven't tried it, the datamodels may need a change to make it work.

```
bunit_data:
  fits_hdu: SCI
  enumerate: [SCI, DRZ]
  fits_keyword: BUNIT
```

hbushouse:
Will that work in the way intended, as a list of possible fits_hdu values? Usually when an enum is specified in a keyword definition it specifies the values that the keyword can have.

bernie-simon:
The schema is used for both reading from the fits file and writing back to it. So it needs to be one to one: one schema entry maps to one fits keyword, which possibly can be in another hdu than the primary hdu. So you would need something like a hierarchy, for example,  meta.science.bunit and meta.error.bunit.

hbushouse:
Merging for now, just to get BUNIT keywords into the headers that need them. Could modify the approach in the future.

PR #699, opened 2017-02-15
Add NullHandler to loggers

bernie-simon:
This suppresses a warning message about no handler found when running tests.


nden:
@bernie-simon Could you  post the Warning message and where it's coming from.

hbushouse:
Are we referring to situations like this?

In [2]: input=datamodels.open('test_lrs1_cal.fits')
No handlers could be found for logger "jwst.datamodels.util"

bernie-simon:
Yes, that's the message I was referring to.

Issue #677, opened 2017-02-01
Photom uses order=1 for NIRISS imaging modes


hbushouse:
The photom step is currently hardwired to assume order=1 when a NIRISS input data model is in the form of an ImageModel, which was intended for use with SOSS exposures that have not gone through the extract_2d step and hence are still in the form of a simple ImageModel. Imaging mode NIRISS exposures also use an ImageModel and hence use this same branch of code. The NIRISS photom reference file has order=0 for all filter+pupil combinations that result in direct images. Hence the photom step does not find a matching row in the ref table because it's looking for order=1.

The solution to this is to not use order as a matching criterion for NIRISS imaging-mode exposures.

hbushouse:
Fixed by #708

PR #708, opened 2017-02-22
Don't use order for NIRISS imaging-mode photom

hbushouse:
The NIRISS function in the photom step has been updated to fix the issue described in #677. Briefly, the function now only uses Order as a photom table row selector for the NIRISS spectral modes and does not use it for NIRISS imaging mode exposures.

The photom step docs have been updated to account for this change, as well as updates to the photom step handling for MIRI MRS mode.


hbushouse:
Fixes #677

Issue #710, opened 2017-02-23
extract_1d does not recognised upper case in src_type keyword


PaddyKavanagh:
The src_type step assigns the default value for each instrument, which for MIRI MRS is 'EXTENDED'. However, when extract_1d checks this keyword, it does a string comparison against lower case keywords, which results in 'EXTENDED' not being recognised and the src_type keyword is reset to 'point'.

philhodge:
Thank you!  I'll fix this.

philhodge:
It looks as if the keyword is 'SRCTYPE', rather than 'SRC_TYPE', based on core.schema.yaml.

hbushouse:
Fixed by #711

PR #711, opened 2017-02-23
Keyword was being compared with lower-case value; see issue #710.

philhodge:
For IFU data, keyword SRCTYPE is read to determine whether the target
is a point source or extended.  The keyword values were assumed to be
either "point" or "extended"; they are, but upper case.

For an extended source with the IFU, background subtraction is not
supposed to be done.  The code now explicitly turns off background
subtraction for IFU data.


philhodge:
The failure looks like a time-out error.

PR #713, opened 2017-02-23
Update documentation for Step.call() in stpipe.

jdavies-st:
@hbushouse Is this reasonable?


jwst/stpipe/step.py
```diff
@@ -424,13 +424,19 @@ def resolve_file_name(self, file_name):
...
 
         To set configuration parameters, pass a `config_file` path or
         keyword arguments.
 
         Any positional `*args` will be passed along to the step's
         `process` method.
+
+        Note: this method creates a new instance of `Step` with the given
+        `*args` and `**kwargs`.  If not used with a `config_file` or specific
+        `*args` and `**kwargs`, it would be better to use the `run` method,
```
hbushouse: It's probably worthwhile to explain why it would be better (because if you create an instance in python and set some of its params, it'll get wiped out by the new instance that's created when using .call) - or whatever the appropriate terminology is.

PR #715, opened 2017-02-26
separate reference file schema from core data schema

nden:
Should be merged after spacetelescope/asdf#239 because this PR changes the order in which the input is checked for fits or asdf in model_base.py.

This is alternative to #697 . 
In #697  `core.schema` contains only the entries common to observational data models and reference file models and `core_data.schema` and `referencefile.schema` inherit from it.

This PR completely separates the core schemas for data models and reference files. `referencefile.yaml` is still the basic schema for all reference files but it does not include `core.schema`.
`referencefile.yaml` uses the `pattern` tag for `exp_type`, `subarray` and other fields which may have more than one value in a reference file. The FITS keyowrds in these cases are `P_{x}`, where `{x}` is an abriviation of the corresponding keyword in the data model. This allow to nmatch and validate more than one value for each of these keywords. Note, that this is used simply for documenting a reference file and not for CRDS selection.

@hbushouse


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -4,13 +4,13 @@ properties:
...
-        #- $ref: http://stsci.edu/schemas/asdf/time/time-1.0.0
-        #- title: Date this file was created (UTC)
-        #  fits_keyword: DATE
-        title: Date this file was created (UTC)
-        type: string
-        fits_keyword: DATE
+        allOf:
+        - $ref: http://stsci.edu/schemas/asdf/time/time-1.0.0
+        - title: Date this file was created (UTC)
+          fits_keyword: DATE
```
nden: Note, the type of `DATE` is changed to `time` and is not a `string` any more.

jwst/datamodels/schemas/referencefile.schema.yaml
```diff
@@ -1,32 +1,203 @@
...
+            fits_keyword: DESCRIP
+          author:
+            title: Author of the reference file
+            type: string
+            fits_keyword: AUTHOR
+          useafter:
+            title: Use after date of the reference file
+            type: string
+            fits_keyword: USEAFTER
+        required: [author, pedigree, description, useafter]
```
nden: Setting these to required to make sure reference file creators don't forget them. However, if current files in CRDS don't have them this may cause them to fail (not necessearily a bad thing).

jaytmiller: The CRDS certifier already checks these so this is redundant and probably too strict.  While the file creators may indeed forget them (unless there are warnings),  these days CRDS won't allow them to be missing when files are submitted.   This was not always the case because historically many files came in "bad" with respect to trivia and I did not make it an issue because we could barely get essential matching parameters correct.  The only thing operationally critical is USEAFTER, and even that was not critical when we started out and even now is only critical when references are submitted and added to the rmaps.   It's an OK idea to issue warnings for these,  but I suspect failing them is bad if it makes "working" references start failing at runtime.

nden: OK, thanks  @jaytmiller. I'll remove `required`.

jwst/datamodels/schemas/referencefile.schema.yaml
```diff
@@ -1,32 +1,200 @@
...
+            title: Type of data in the exposure
+            type: string
+            pattern: "|\
+              ^FGS_(?=IMAGE|DARK|FOCUS|INTFLAT|SKYFLAT)|\
+              ^MIR_(?=IMAGE|TACQ|LYOT|4QPM|LRS-FIXEDSLIT|LRS-SLITLESS|MRS|DARK|FLAT-IMAGE|FLATIMAGE|FLAT-MRS|FLATMRS|CORONCAL)|\
+              ^NIS_(?=AMI|DARK|FOCUS|IMAGE|LAMP|SOSS|TACQ|TACONFIRM|WFSS)|\
+              ^NRC_(?=IMAGE|GRISM|TACQ|CORON|FOCUS|DARK|FLAT|LED|WFSC|TACONFIRM|TSIMAGE|TSGRISM)|\
+              ^NRS_(?=AUTOFLAT|AUTOWAVE|BOTA|BRIGHTOBJ|CONFIRM|DARK|FIXEDSLIT|FOCUS|IFU|IMAGE|LAMP|MIMF|MSASPEC|TACONFIRM|TACQ|TASLIT)|\
+              ANY|N//A"
+            fits_keyword: P_EXPTYP
```
hbushouse: Does this imply that all existing ref files in CRDS will need to have their headers updated to rename the EXP_TYPE keyword to P_EXPTYP?

jwst/datamodels/schemas/referencefile.schema.yaml
```diff
@@ -1,32 +1,200 @@
...
+              SUB128CENTER|SUB16|SUB160|SUB160P|SUB1A|SUB1B|\
+              SUB2048|SUB256|SUB32|SUB32LLCORNER|SUB32DIAGONAL|SUB32CENTER|\
+              SUB320|SUB400P|SUB512|SUB64|SUB640|\
+              SUB64P|SUB80|SUB8LLCORNER|SUB8DIAGONAL|SUB8CENTER|SUB96|\
+              SUBAMPCAL|SUBFP1A|SUBFP1B|SUBGRISM128|\
+              SUBGRISM256|SUBGRISM64|SUBSTRIP256|SUBSTRIPE256|SUBSTRIP96|\
+              SUBPRISM|SUBSTRIP80|SUBTASOSS|SUBTAAMI|\
+              WFSS128C|WFSS128R|WFSS64C|WFSS64R|N/A
+              "
+            fits_keyword: P_SUBARR
```
hbushouse: Same question as above for EXP_TYPE keyword name ...

philhodge:
Reference files are not necessarily exposures, even reference files that 
contain images.  And even if they were derived from exposures of the 
same type (e.g. ordinary flats or darks), what's wrong with EXP_TYPE as 
the keyword?

bernie-simon:
> On Mar 9, 2017, at 6:01 AM, Howard Bushouse <notifications@github.com> wrote:
> 
> Do we know what the effect will be of breaking up existing blocks of keywords (e.g. instrument, exposure, ...) into separate schema in terms of the organization of the header in FITS files that are created from a data model, i.e. will the keywords defined in the different schemas show up in different places in the header, such that keywords that used to be all together in an "instrument" block will now be split apart in some way?

If you update an existing header the keywords will be written back to their original location. If you create a new fits header, the keywords will be written in the order they are encountered when validating the schema. That is, all the keywords in one schema will be written together.

jwst/datamodels/schemas/referencefile.schema.yaml
nden: `EXP_TYPE` in a reference file is used only for documenting the file. Unlike an exposure where `EXP_TYPE` has only one value, the same reference file can be used with several type of observations. How do we document that? The core data schema which uses `enum` allows for only one value while we need to match a "pattern" of values. After talking this over with @jaytmiller I am proposing to replace `EXP_TYPE` with `P_EXPTYP` in future files.
We don't have to change existing reference files. None of the fields here is `required` .

hbushouse: Instead of having P_EXPTYP or P_SUBARR **_replace_** the "normal" EXPTYPE and SUBARRAY keywords in the ref files, would it be feasible to have both of them appear? The "normal" version would serve as documentation of where the data came from (how they were obtained), while the new form would be used as the selector in CRDS.

nden:
@hbushouse I suspect the order of the keywords depends on the order the schemas are included in the model. But I need to test that.

jwst/datamodels/schemas/referencefile.schema.yaml
nden: This comment is to document what @hbushouse and I discussed which is - although EXP_TYPE and P_EXPTYP are not used for selection EXP_TYPE can be used to describe the type of observation used to create the reference file.

jwst/datamodels/schemas/dark.schema.yaml
```diff
@@ -1,25 +1,30 @@
...
-      ndim: 2
-      datatype: uint32
+      ndim: 4
+      datatype: int16
     err:
       title: Error array
       fits_hdu: ERR
       default: 0.0
-      ndim: 3
+      ndim: 4
```
hbushouse: The darks for all near-IR instruments use 3D SCI and ERR arrays, and a 2D DQ array, as was listed in the previous version of the dark.schema. It's only MIRI darks, which have their own dedicated dark model, that use 4D arrays.

jwst/datamodels/schemas/dark.schema.yaml
```diff
@@ -1,25 +1,30 @@
 title: Dark current data model
 allOf:
 - $ref: referencefile.schema.yaml
+- $ref: instrument_name.schema.yaml
+- $ref: detector.schema.yaml
+- $ref: filter_core.schema.yaml
```
hbushouse: Is it going to be confusing for developers to have to learn which individual keyword schema files need to use the _core suffix in their names? And does this mean it's an identical copy of what's in the science core.schema, as opposed to the way it's defined for use in reference files?

jwst/datamodels/schemas/dark.schema.yaml
nden: Ah, good catch. I was testing a fix to a memory problem in asdf and forgot to reset this. Will push the fix with the next update.

jwst/datamodels/schemas/dark.schema.yaml
nden: There was already a `filter` schema which is for the throughput HDU. Maybe that one should be renamed?

hbushouse: That's a possibility. The existing "filter" schema is for a set of reference files that contain filter throughput curves, which are used in the NIRISS AMI level-3 processing. We could potentially rename that schema to something like "throughput" or "filter_throughput", so that your schema for the filter keyword could be just "filter".

Alternatively, what do folks think about naming the mini-schemas for individual keywords with something like a "_keyword" suffix, such as "filter_keyword.schema.yaml"? It requires more typing, but at the same time it does make it obvious that these schemas are dedicated to just defining a keyword and not an entire data model.

philhodge: If there are going to be schemas for individual keywords, I would prefer longer and clearer names (such as filter_keyword), because these are in a category of their own, in some sense, and a common suffix would help us to find these schemas.

hbushouse: If we want to go that route, would it be beneficial to make "keyword" a prefix, as opposed to suffix, such as "keyword_filter.schema.yaml" and "keyword_exptype.schema.yaml", just so they all show up together in directory listings? There are some cases of data models where we've tried to do something like, for example having all of the various flavors of photom data model schemas start with the word "photom", so that they're easy to find and automatically get grouped together.

nden: Using a "keyword" prefix sounds reasonable. I'll make the changes.

nden:
This is ready for final review.

jwst/datamodels/schemas/dark.schema.yaml
```diff
@@ -1,30 +1,28 @@
...
       default: 0.0
-      ndim: 4
+      ndim: 3
       datatype: float32
     dq:
       title: 2-D data quality array for all planes
       fits_hdu: DQ
       default: 0
-      ndim: 4
+      ndim: 3
```
hbushouse: I think the original schema had the DQ array defined as 2D. Only the SCI and ERR are 3D.

jwst/datamodels/schemas/psubarray.schema.yaml
```diff
@@ -0,0 +1,22 @@
+type: object
```
hbushouse: Should the name of this file be "keyword_psubarray.schema.yaml"?

jwst/datamodels/schemas/dark.schema.yaml
```diff
@@ -13,8 +16,8 @@ allOf:
       title: 2-D data quality array for all planes
       fits_hdu: DQ
       default: 0
-      ndim: 2
-      datatype: uint32
+      ndim: 3
+      datatype: int16
```
hbushouse: Both of these lines should revert to their original values of 2 and uint32, respectively.

jwst/datamodels/schemas/referencefile.schema.yaml
```diff
@@ -1,32 +1,73 @@
...
+          name:
+            title: Instrument used to acquire the data
+            type: string
+            enum: [NIRCAM, NIRSPEC, MIRI, TFI, FGS, NIRISS, ANY, N/A]
+            fits_keyword: INSTRUME
+          detector:
+            title: Name of detector used to acquire the data
+            type: string
+            enum: [NRCA1, NRCA2, NRCA3, NRCA4, NRCALONG, NRCB1, NRCB2, NRCB3, NRCB4,
+              NRCBLONG, NRS1, NRS2, ANY, MIRIMAGE, MIRIFULONG, MIRIFUSHORT, NIRISS,
```
hbushouse: I vote for having "NIRISS" removed from the list and no longer treat it as deprecated. The few test data files laying around that might have this value will just need to be updated accordingly.

hbushouse:
Collisions between the filter keyword schema file and the "filter" reference file data model schema file are no longer an issue, because we've decided to call the keyword schema file "keyword_filter.schema.yaml", but I'm wondering if this would be an opportune time to just go ahead and change the name of the "filter" data model schema anyway. The REFTYPE for this reference file is "THROUGHPUT", not "FILTER", and hence it would probably make more sense to have the data model and schema files for it called "throughput.py" and "throughput.schema.yaml", respectively. That would also likely entail a change to the data model class name (ThroughputModel instead of FilterModel), which also means a minor code change in one or two places.

jwst/datamodels/schemas/dark.schema.yaml
nden: Good catch!

jwst/datamodels/schemas/psubarray.schema.yaml
nden: Yes, it should.  I had it originally as the whole section but realized we only need the name of the subarray and not the other attributes.

jwst/datamodels/schemas/dark.schema.yaml
nden: Agreed.

jwst/datamodels/schemas/referencefile.schema.yaml
nden: "NIRISS" removed.

hbushouse: Of course it suddenly occurs to me that this is the schema that will be used to validate reference files (not science files) and hence if there are any existing reference files in CRDS that have DETECTOR='NIRISS', we're going to start seeing validation errors. We should have @jaytmiller check to see if there are any such ref files in the currently active set.

jaytmiller: Famous last words,  I don't see any active references with DETECTOR=NIRISS.

hbushouse: Well, once Nadia merges this, we should find out the hard way whether you're right. ;-)

jwst/datamodels/schemas/dark.schema.yaml
```diff
@@ -1,6 +1,9 @@
 title: Dark current data model
 allOf:
 - $ref: referencefile.schema.yaml
+- $ref: keyword_filter.schema.yaml
+- $ref: keyword_exptype.schema.yaml
```
hbushouse: Wait a sec. I just now dawns on me: why are we including FILTER and EXPTYPE keyword definitions in the dark data model schema? Darks are selected on DETECTOR, SUBARRAY, and sometimes READPATT. They are never selected based on either FILTER or EXPTYPE. So we don't need FILTER or EXPTYPE keywords in darks and we DO need READPATT (at least in some of them). I assume DETECTOR is already included in the standard list contained in the new referencefile.schema.yaml.

nden: I don't know what needs to be done for darks. I looked at one dark file and tried to include keywords that I saw. But mostly the purpose was to test the new schema on FITS files. This and all other schemas need to be reworked by step developers who know what keywords are actually needed.

nden: Should we have a sprint next week and work out the details for all schemas at once?

hbushouse: That might be the best approach. We can have you and/or I present the basics of what needs to be done (i.e. teach everyone about using the new keyword_xxxx.schema.yaml files) and then let them dive into updating an assigned set of schemas.

hbushouse: So I say go ahead and merge this fairly soon, so that the basics (i.e. the new referencefile.schema.yaml and a few of the keyword_xxxx.schema.yaml files) are in there and available for everyone to start building on.

hbushouse:
Merged for Nadia, because she appears to have been banned from merging PR's on github.

Issue #718, opened 2017-03-01
For IFU data, check BUNIT to see whether flux correction was done


philhodge:
For data other than IFU, the `extract_1d` step can convert count rate to flux density.  For IFU data, however, the data should have been flux calibrated (via the `photom` step) at an earlier stage.  `extract_1d` should check whether the data have in fact been converted to flux density.  If so, the extracted values should be written to the `flux` column and `net` should be set to zero; if not, the extracted values should be written to the `net` column and `flux` should be set to zero.

If the data input to `extract_1d` have been flux calibrated, the BUNIT keyword should have been written to the header, and its value should contain some variation of jansky (the unit).  If the value of bunit contains "Jy" or "jansky" or "Jansky" (yes, I know jansky shouldn't be capitalized, but be lenient), take that to mean the data have been flux calibrated.

PR #719, opened 2017-03-01
Check BUNIT for IFU data.  #718

### LINKED ISSUE #648: extract_1d should not do flux calibration for IFU
The `extract_1d` step applies flux calibration to the count rate values in the net column (using the RELSENS extension of the input file) in order to populate the flux column.  For both MIRI and NIRSpec IFU cubes, however, the data values will already be in flux units; see #611 and #647.  The `extract_1d` step should be modified to not attempt to apply flux calibration to any IFU data; instead, the net column should be copied directly to the flux column.

philhodge:
For IFU data (only), keyword BUNIT is checked to see whether flux
calibration has been done.  If it has, the extracted data will be
written to the flux column; if not, the extracted data will be written
to the net column.

This also closes #648.

### DIFF
--- a/jwst/extract_1d/ifu.py
+++ b/jwst/extract_1d/ifu.py
@@ -60,20 +60,32 @@ def ifu_extract1d(input_model, refname, source_type):
         log.critical('Missing extraction parameters.')
         raise ValueError('Missing extraction parameters.')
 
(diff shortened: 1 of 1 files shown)


philhodge:
The Travis test failure was "The job exceeded the maximum time limit for jobs, and has been terminated."

PR #726, opened 2017-03-13
Add schema editor to datamodels

bernie-simon:
The schema editor compares a downloaded copy of keyword database to
the schema files under the datamodels directory and ouputs modified
schema files that agree with the keyword database. The schema editor
is mmeant to be run as a command line script and the comments at the
top of schema_editor.py give example scripts on how to do that, none
longer than five lines. The user can set the options that control the
schema editor either on the command line, from a file, interactively,
or when the schema editor is initialized.


bernie-simon:
Time limit errors only, so I am merging.

Issue #639, opened 2017-01-25
Level2 Associations


stscieisenhamer:
Finalize Level2 associations.  Issues include
- [x] [Level2 association file naming scheme](https://aeon.stsci.edu/ssb/trac/jwst/ticket/203)
- [x] How the workflow will produce/use them
- [x] How the Level2b module is using them
    See implementation in comments.
- [x] Complete ToDo below

stscieisenhamer:
# Resources
- [Background types discussion](https://confluence.stsci.edu/pages/viewpage.action?pageId=58786600#Level3FileNames(AssociationProducts)-Feb28th,2017-Associationsweeklymeeting(Backgrounds,Testdata))

stscieisenhamer:
In discussions with @hbushouse, the level2 associations should be an allowed input for the calweb_spec/image2 module. The modules will loop through all the members, processing each member, using any extra attributes, such as background or imprint, for processing purposes.

From the SDP standpoint. the generator will generate a Level 2 association for each observation. The association will contain all the exposures of that observation. For each exposure that needs extra information, that information will be added to the member's attributes, such as background or imprint.

hbushouse:
A couple of subtleties here:

1) imprint and dedicated background exposures will also appear as members with exptype=science, but with no additional attributes such as 'bkgexps' or 'imprint', so that they will also get processed through all of the steps of the pipeline like a normal exposure, producing a calibrated product.

2) collections of exposures from something like a dither or mosaic pattern that do NOT use any mechanisms such as nodded/dedicated backgrounds or imprint will still be built into a level-2b ASN file, and the level-2b pipeline will simply calibrate all of them one by one.

stscieisenhamer:
@hbushouse So, there are two level2 modules: `spec2` and `image2`. Do we produce two level2 associations: one for each? Or is there/can there be a wrapper Level2b pipeline, `calwebb_level2` that wraps/decides which sub-pipeline to run per exposure? This way the looping through the association is done by the wrapper, and the individual `spec2|image2` routines do not have to.

Methinks the workflow, and even for user processing, having the wrapper pipeline would be much easier.

sosey:
How does this connect into the WFSS pipeline which needs the association with the grism and imaging observations to do spec2 processing of the grisms?

stscieisenhamer:
## Naming proposal
Level 3 associations are named as follows:
```
[...]

PR #733, opened 2017-03-20
Update Level2 schema to match current pipeline usage

stscieisenhamer:
WIP #639


stscieisenhamer:
@hbushouse This should take care of the current `load_asn` issues with level2 associations.

PR #740, opened 2017-03-22
Updates to skymatch docs

hbushouse:
A few minor updates to skymatch docs, including fixing the file name of "skimage.rst" to "skyimage.rst". This is the first round of updates needed for skymatch docs so that they include more user-level descriptive information, as is done for other pipeline step docs.


hbushouse:
Relates to #741

PR #744, opened 2017-03-22
ModelContainer grouping

jdavies-st:
This PR makes grouping more robust.

- No more unexpected grouping based on filename
- Grouping is strictly done based on `meta` keywords of input datamodels, i.e. for what we expect from NIRCam data
- Warnings are produced if the `meta` keywords are not populated
- Several new unit tests now test ModelContainer generally and grouping specifically

### DIFF
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -247,8 +247,8 @@ class ModelContainer(model_base.DataModel):
                                 model_attrs[8].lower()]))
                 model.meta.group_id = group_id
             except:
-                w = 'One or more datamodels is missing correct metadata. ' + \
-                    'Grouping will not be correct.'
+                w = '`{}` is missing'.format(model.meta.filename) + \
+                    ' metadata. Grouping by exposure may not be correct.'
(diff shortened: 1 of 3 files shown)


hbushouse:
Is the grouping now only done for NIRCam data (hard to tell from browsing the code) or is it still attempted for all data? In other words, do testers using data from other instruments such as MIRI need to worry about having all the grouping-related meta data populated properly?

jdavies-st:
The code is agnostic as to instrument. It will group properly for NIRCam.  For the other instruments, if the metadata is there, it will make a single group (single detector) and work as it did before.  If the metadata is not there regardless of instrument, the code gives a warning and produces a single group.  The warning.warn() message can be treated how we chose in the pipeline, but it will get passed on to the user just like deprecation warnings do.

jdavies-st:
Also, if the metadata needed for grouping is not present, it produces a single group (i.e. it doesn't group), sets the group name to `None` and issues a warning.  The warning encourages people to sort out their metadata.  But there should be no unintended grouping now.

So the metadata has to be populated for grouping to work, and if its not, no grouping and a warning.

Issue #631, opened 2017-01-20
Update calwebb_spec2 to use proper _x1dints output product name


hbushouse:
The 1D extracted spectra produced by the extract_1d step at the end of the calwebb_spec2 pipeline can be one of two flavors. If the input is a product containing data averaged over all integrations for the exposure (e.g. a _rate file), then a single extracted spectrum (per source) will be produced by extract_1d and the resulting output product type is _x1d. But if the input contains per-integration data (e.g. a _rateints file), extracted spectra will be produced for each integration, in which case the output product should use the _x1dints file type suffix. Currently both cases use the _x1d designation.

hbushouse:
Fixed by #745

PR #745, opened 2017-03-23
calspec2 updates

hbushouse:
Several updates to the calwebb_spec2 pipeline module, including:

 - handling members that do not have "bkgexps" or "imprint" attributes
 - saving extracted spectra from multi-integration exposures as "x1dints" files
 - ability to handle a single data model as input
 - general updates to the input handling and asn template construction

The input routine "Lvl2Input" is now capable of handling

 - an ASN file name
 - a single exposure file name
 - a single exposure data model


hbushouse:
Fixes #631

PR #748, opened 2017-03-23
New twopoint_difference routine from M. Regan

hbushouse:
In PR #581 Mike Regan provided us with an updated and optimized version of the twopoint_different routine used in the jump step. The version included here has had just slight modifications for working within the jwst package environment and just a bit of cosmetic changes. The routine has been tested and compared against the previous version of twopoint_difference and has been found to not only run 2-3x times faster than the previous version, but also properly handles all corner cases of things like saturated groups. It also uses the approach of the original creators of the algorithm in terms of not including the largest outlier of all groups (regardless of whether or not it's above the rejection threshold) when computing the median of group differences.

### DIFF
--- a/jwst/jump/tests/test_twopoint_difference.py
+++ b/jwst/jump/tests/test_twopoint_difference.py
@@ -424,6 +424,6 @@ def setup_cube(ngroups,readnoise=10):
     return data, gdq, nframes, read_noise, rej_threshold
 
 
-if __name__ == '__main__':
-    pytest.main(['-x', 'test_find_CRs_mwr.py'])
+#if __name__ == '__main__':
+#    pytest.main(['-x', 'test_find_CRs_mwr.py'])
 #   # pytest.main(['test_find_CRs_mwr.py'])
--- a/jwst/jump/tests/test_twopoint_difference.py
+++ b/jwst/jump/tests/test_twopoint_difference.py
@@ -1,0 +1,429 @@
+from ..twopoint_difference import find_CRs
+import numpy as np
+import pytest
+from ...datamodels import dqflags
+
+#@pytest.mark.skip(reason="testing skipping")
+def test_noCRs_NoFlux():
+    ngroups=5
+    data, gdq, nframes, read_noise, rej_threshold = setup_cube(ngroups)
+    assert(0 == np.max(find_CRs(data,gdq,read_noise,rej_threshold,nframes))) # no CR found
+
+#@pytest.mark.skip
+def test_5grps_cr3_NoFlux():
+    ngroups=5
+    data, gdq, nframes, read_noise, rej_threshold = setup_cube(ngroups)
+
+    data[0, 0:2, 100, 100] = 10.0
+    data[0, 2:5, 100, 100] = 1000
+    medianDiff = find_CRs(data, gdq, read_noise, rej_threshold, nframes)
+    assert(4 == np.max(gdq)) #a CR was found
+    assert(2 == np.argmax(gdq[0,:,100,100])) #find the CR in the expected group
+
+#@pytest.mark.skip("testing skipping")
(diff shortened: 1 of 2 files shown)


hbushouse:
Does anyone have any idea as to why the test module "test_twopoint_difference.py" is consistently failing in Travis when run on Linux under py2.7, but succeeds on all other platform+OS combinations? @jhunkeler @stscieisenhamer @jdavies-st @bernie-simon

stscieisenhamer:
hmmmm, just form inspection, test_twopoint_difference.py has the ending lines

if __name__ == '__main__':
    pytest.main(['-x', 'test_find_CRs_mwr.py'])
#   # pytest.main(['test_find_CRs_mwr.py'])

No other test, except for the package setup, has this. Its possible its getting triggered somehow and is linux 2.7 incompatible?  I'd remove that just to see.

hbushouse:
Commented out those lines at the end of the test routine and it's STILL failing on just that one platform (Linux + py2.7).

stscieisenhamer:
I have tried this on my on ubuntu setup, using python 2.7, and it works fine. It appears to be a Travis-specific issue. Besides the brute-force approach, disabling all tests except one and going from there, I am not up on my Travis-fu to help with that.  Otherwise, I would not hold up the merge for this.

hbushouse:
I now see that your #749 is having the same issue: Travis is killing the process in the middle of one of the exp_to_source tests on the Linux+py27 platform. So now I'm wondering if it's an issue with that system.

stscieisenhamer:
*facepalm*

hbushouse:
facepalm-meme-03

stscieisenhamer:
I have reproduced the error. When running
```
pytest jwst/jump/tests/test_twopoint_difference.py
````
it works fine.
But with
```
python setup.py test
```
under my ubuntu, I do get:
```
jwst/exp_to_source/tests/test_exp_to_source.py ...
jwst/exp_to_source/tests/test_main.py ...
jwst/jump/tests/test_twopoint_difference.py FFFKilled
```
Testing further...

hbushouse:
Runs fine under "python setup.py test" on my 'tute RedHat linux system with py2.7, as well as pytest.

stscieisenhamer:
Hmmm, what is the memory on you linux system? I'm going to guess its more the 2GB?  My linux is at 2GB + 2GB swap, and that is the problem, which exhibits the same behavior as the Travis builds.

@jhunkeler How much memory do the Travis vm's get? I'm guessing its 2GB?

hbushouse:
Seriously?! I've got 16GB on my Mac laptop and 32GB on my Linux desktop. No wonder it works for me. The google queries that I did earlier had a lot of references to a 3GB limit on Travis systems.

stscieisenhamer:
And why the Python3 ones don't fail: The memory requirements under Python3 are half that of the Python2 runs. But, its close: Whereas the Python2 exhibits memory leak behavior which the Python3 does not, the twopoint test does consume my 2+2 memory...it just barely makes it.

So, we need memory. The twopoint is creating some largish arrays that may not be necessary for testing, so that will alleviate the issue. exp_to_source could use a smaller test data set.

However, the big story is that neither of these will alleviate the Python2.7 memory leak issue necessarily. The addition of more tests to the suite will overcome the decrease in memory usage of any particular test soon.

hbushouse:
Yes, I noticed that the twopoint tests are creating zero-filled 2048x2048xngroups arrays and then only setting non-zero values for one pixel. So presumably the same tests could be accomplished using much smaller arrays. I'll look into that.

stscieisenhamer:
Just for kicks, I re-ran the tests with the reduced twopoint test incorporated, and all went fine. However, note still that the suite takes ~3GB to run. If this what the Travis limit is at, we're gonna have a bad time in the near future.

PR #749, opened 2017-03-23
Refactor Level2 Associations

stscieisenhamer:
With decisions made concerning pipeline and SDP process, the first in a series of formalizations for Level2 associations has begun.


stscieisenhamer:
Failure has been determined to be due to VM/Test memory issue. Since unrelated, merging.

PR #753, opened 2017-03-24
Make Datamodels a subclass of astropy.NDData

bernie-simon:
Erik Tollerud requested this modification.The motivation is that it will be easier to develop software that works across different telescopes if the data format they use all adhere to a common api,
which is defined in the NDDataBase class. No, it's not a database, it's the base of the NDData class. This abstract class defines six properties that its subclasses must implement. These have been added to datamodels.model_base.py. The meta property required the most work, as it is a flat dictionary in NDData and has a hierachical structure in datamodels. The code which creates a facade for the meta propery to implement the behavior NDData expects has been added to
datamodel.properties.py. None of the existing interfaces to datamodels have been modified, all code that uses datamodels should work as before.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -17,6 +17,7 @@ from astropy.extern import six
 from astropy.io import fits
 from astropy.time import Time
 from astropy.wcs import WCS
+from astropy.nddata import nddata_base
 
 from asdf import AsdfFile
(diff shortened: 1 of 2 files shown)


jdavies-st:
Looks good Bernie!  It passes our unit tests, but perhaps there's more testing that could be done on the astropy NDData side?  Maybe @eteq can have a look?

PR #755, opened 2017-03-27
calspec2: add exptype to asn template

hbushouse:
This fixes a bug that I inadvertently introduced in #747, in which calspec2 checks the value of the "exptype" attribute in the association. For single file/model input, the exptype attribute of the asn was not defined. It is now, and is set to a default of "science."


hbushouse:
Travis failures not related to this PR. Merging.

PR #758, opened 2017-03-28
Updated test for tuple lists

stscieisenhamer:


stscieisenhamer:
linux+py2 memory issue; merging

PR #761, opened 2017-03-29
enable ami mode

### LINKED ISSUE #737: Enable assign_wcs for NIRISS AMI
As far as we know right now exposures taken in NIRISS AMI mode should use the same WCS distortion (and other) transforms as direct imaging. The current distortion reference file in CRDS is already enabled in the rmap to be used with EXP_TYPE='NIS_AMI'. So all that's needed in assign_wcs to allow it to compute a WCS for AMI exposures is to enable the setting in niriss.py:

exp_type2transform = {'nis_image': imaging,
                      'nis_wfss': not_implemented_mode,
                      'nis_soss': niriss_soss,
                      'nis_ami': not_implemented_mode,
                      'nis_tacq': imaging,
                      'nis_taconfirm': imaging,
                      'nis_focus': imaging,

so that nis_ami is set to "imaging".

nden:
Fixes #737

### DIFF
--- a/jwst/assign_wcs/niriss.py
+++ b/jwst/assign_wcs/niriss.py
@@ -141,7 +141,7 @@ def imaging_distortion(input_model, reference_files):
 exp_type2transform = {'nis_image': imaging,
                       'nis_wfss': not_implemented_mode,
                       'nis_soss': niriss_soss,
-                      'nis_ami': not_implemented_mode,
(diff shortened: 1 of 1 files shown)


hbushouse:
The only Travis failure is due to the now ubiquitous Linux+Py2.7 memory limit issue.

PR #764, opened 2017-03-30
Calwebb_ami3 updates for product naming

hbushouse:
Updated calwebb_ami3 pipeline to use new product suffixes for output files ("lg" -> "ami"), and use a new entry in the ASN table called "psf_name" for constructing the file name of the averaged PSF fringe data, so that the file name is based on the PSF target name rather than the science target name.


hbushouse:
Travis failures as expected and nothing to do with this PR. Merging.

Issue #765, opened 2017-03-30
NIRISS imaging distortion model needs updating


jdavies-st:
The current NIRISS imaging distortion is broken.

1) The distortion reffile currently in CRDS `jwst_niriss_distortion_0006.asdf` only implements a part of the distortion and scale correction, translating pixel coords to arcsec.  There is no account of the reference pixel in aperture nor the location of that reference pixel in v2v3, nor the parity or the v3yangle.

So while it does have the correct units, it needs significant fixing to work.  These are essentially the same issues that the NIRCam distortion reffile had, and the solutions for fixing them are likely similar.  The current reffile was created by Michael Wolfe based on data from Andre Martel.  It might pay for them to have a chat with Bryan Hilbert on how to go about fixes.

2) On our end, we currently assume the distortion comes in units of arcmin. So we need to remove that scaling factor in our NIRISS `assign_wcs` code.  This is an easy fix and will at least get the (wrong) distortion above into the correct units.

nden:
True. The NIRISS team is aware of these issues and has agreed to rework the files. Same goes for FGS.

nden:
Just to add to this - because of other work the NIRISS team won't be able to provide the new files soon, the deadline for this is early September. Also all instrument teams have agreed to provide reference files wih units of arcsec in V2, V3.

jdavies-st:
Thanks Nadia!  I'll modify the NIRISS imaging for the changes then, which will also solve the issue of wrong units being produced by our code and the current NIRISS imaging distortion reffile.  This will help with testing on our end.

Currently, the FGS distortion files only correct for distortion.  They go from distorted pixels to undistorted pixels, so Science to Ideal frames in SIAF terms.  They contain no scaling to arcsec, no offsets to account for crval_n_ or v2/v3_ref, no rotation to account for v3yangle and no parity.

jdavies-st:
2. above is fixed in #768

PR #768, opened 2017-03-31
Fix assign_wcs for expected units (arcsec) for NIRISS imaging distortion

jdavies-st:
Current and future WCS distortion reffiles for NIRISS imaging mode produce units of arcsec.  We had treated them as units in arcmin before.

### DIFF
--- a/jwst/assign_wcs/niriss.py
+++ b/jwst/assign_wcs/niriss.py
@@ -133,9 +133,7 @@ def imaging(input_model, reference_files):
 
 def imaging_distortion(input_model, reference_files):
     distortion = AsdfFile.open(reference_files['distortion']).tree['model']
-    # Convert to arcsec
-    transform = distortion | Scale(1/60) & Scale(1/60)
-    return transform
+    return distortion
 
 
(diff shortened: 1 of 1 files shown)


jwst/assign_wcs/niriss.py
```diff
@@ -133,8 +133,8 @@ def imaging(input_model, reference_files):
 
 def imaging_distortion(input_model, reference_files):
     distortion = AsdfFile.open(reference_files['distortion']).tree['model']
-    # Convert to arcsec
-    transform = distortion | Scale(1/60) & Scale(1/60)
+    # Convert to deg.  Output of distortion model is in arcsec.
+    transform = distortion | Scale(1/3600) & Scale(1/3600)
```
nden: Eventually (after all instruments have correct reference file) the conversion to degrees should be moved to the V2V3 to sky transform if possible.

jdavies-st: That makes a lot of sense, as that will remove some code redundancy redundancy.

jdavies-st:
Fixes #765

PR #771, opened 2017-04-03
Read ra/dec from TARG_RA/TARG_DEC instead of PROP

### LINKED ISSUE #767: set_bary_helio_times uses wrong keywords for target RA and Dec
The set_bary_helio_times script uses the target RA and Dec from the header as one of the inputs to its calculations. It is currently getting those values from the "PROP_RA" and "PROP_DEC" keywords, which to the uninitiated seem like they are the RA and Dec for the target as specified in the proposal. They are in fact the **_PROPer motion_** of the target in RA and Dec. The correct keywords to use as input are "TARG_RA" and "TARG_DEC".

stscieisenhamer:
Resolves #767


stscieisenhamer:
Besides the too-small-vm issue (next on list), all is good. merging.

PR #772, opened 2017-04-03
Change error handling in datamodels

bernie-simon:
Error handling is now controlled by two keyword parameters in the __init__ method, error_if_invalid and pass_invalid_values. Both are false by default. Error_if_invalid is new.

If error_if_invalid is true, datamodels throws a ValidationError if a fits file does not validate against the schema. It raises a ValidationWarning if error_if_invalid is false. To convert the warning message to a log file message, call "logging.captureWarnings(True)". Datamodels will delay throwing an error when reading a fits file until all the data is checked so that a mesaage is printed for all problems with validation. Howerever, an error will be generated whenever a value is modified after reading, as the code has no way of knowing how many changes will be made.

If pass_invalid_values is true, datamodels will permit changes to the files which do not validate. If it is false, invalid changes will be rolled back.  An error will be generated even if pass_invalid_values is true if error_if_invalid is also true, the keywords are independent of one another.

The default case, where both keyword parameters are false, is most likely what you want. Invalid changes generate a warning mmessage and the change is rolled back.

### DIFF
--- a/jwst/datamodels/fits_support.py
+++ b/jwst/datamodels/fits_support.py
@@ -529,7 +529,7 @@ def _load_from_schema(hdulist, schema, tree, error_if_invalid,
     if error_if_invalid and invalid_keywords:
         msgfmt = "fits keywords are not valid: {0}"
         msg = msgfmt.format(','.join(list(invalid_keywords)))
-        raise properties.ValidationError(msg)
+        raise jsonschema.ValidationError(msg)
     return known_keywords, known_datas
 
 
--- a/jwst/datamodels/fits_support.py
+++ b/jwst/datamodels/fits_support.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, division, unicode_literals, print_functi
 import datetime
 import os
 import re
+import warnings
 
 import numpy as np
 
@@ -232,7 +233,7 @@ def _fits_comment_section_handler(validator, properties, instance, schema):
 
 def _fits_element_writer(validator, fits_keyword, instance, schema):
     if schema.get('type', 'object') == 'array':
-        raise ValueError("'fits_keyword' not valid with type of 'array'")
+        raise ValueError("'fits_keyword' is not valid with type of 'array'")
 
     hdu_name = _get_hdu_name(schema)
     index = getattr(validator, 'sequence_index', None)
@@ -467,8 +468,8 @@ def _schema_has_fits_hdu(schema):
     return has_fits_hdu[0]
 
 
-def _load_from_schema(hdulist, schema, tree, validate=True,
-                      pass_invalid_values=False):
+def _load_from_schema(hdulist, schema, tree, warn_if_invalid,
+                      pass_invalid_values):
     known_keywords = {}
     known_datas = set()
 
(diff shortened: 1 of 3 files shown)


hbushouse:
Remind me - is there anything like an environment variable that can be set to determine whether or not schema validation results in either a warning or error, or must it be set as an argument within code?

bernie-simon:
I keep changing my mind on this, but I think there will soon be one keyword parameter, pass_invalid_values, and one environment variable, PASS_INVALID_VALUES. If it is False (falsey in the case of the environment variable) the code will generate an error if the model does not validate against its schema. In the new code it tries to delay throwing the error until the whole image is read so you get maximum information. If pass_invalid_values is true (truthy),  changes which invalidate the schema are allowed and a warning is generated (jwst.datamodels.properties.ValidationWarning).

sosey:
Just a side question - is there a reason to have environment variables instead of default configuration files that live with the code?

philhodge:
This validation is a royal pain.  I REALLY want a way to ignore these 
"errors".  If it matters that S_REGION be a string instead of a float 
(to pick an example from this morning), then I'll get an exception if 
and only if some part of the step code actually needs the value, and in 
that case it will be a meaningful exception.

hbushouse:
My motivation for wanting an environment variable is to make it easy to change the behavior in any given circumstance. Like Phil said, a lot of times you don't need or want the strict checking, because the vast majority of the time it's throwing an error due to things that the pipeline, or at least a given part of it, doesn't even use or care about. Since the error throwing was turned back on a week or so ago I've spent a LOT of time updating keyword values in all of the input files to our regression tests, even though none of those keyword values are used anywhere in the pipeline. Same goes for a developer using a random input image for testing some step or pipeline. There needs to be an easy way to turn it off.

A config file that lives with the code is one approach, but that then requires editing of that config file in the environment in which things are being run. In the case of the regression tests, for example, the code base is rebuilt every night and hence if the default config file setting is in the opposite sense of what we want for regression tests, there'd have to be a hook in the reg test setup somewhere to flip that switch every time the tests run. Similarly for testers/users outside of SSB (e.g. INS). It's hard enough to explain to them how to run things - having to edit another config file only compounds the opacity.

So personally I like the environment variable approach. That's easy to set in the regression test (we already set one that points to the desired CRDS context) and personal developer environments (IMHO).

bernie-simon:
If you set pass_invalid_values to True (or PASS_INVALID_VALUES to 1) you will get warning messages that contain the name of the fits keyword when I have it. It then will be the responsibility of your code to ignore errors when they are unimportant and raise them when the are,

bernie-simon:
Here is how error handling in datamodels stands now. The initial read and all changes are validated against the schema. Each error detected throws a warning of type jwst.datamodels.properties.ValidationWarning. The warning message also contains the text "not valid" in case you want to match on that instead. On opening a file with datamodels you can set the keyword parameter pass_invalid_values. If it is False (the default), an error is raised in addition to the warnings. The type of this error is jsonschema.ValidationError. If pass_invalid_values is True, no error is raised. **But** the data which failed to validate is not read into the datamodel, or the change is rolled back. This is because the integrity of the validation depends on there being no more than one invalid value in the datamodel. If there is more than one, the validation can produce false positives. I understand that this can be a problem if the schema used for validation is out of date. On your own machine, at least, you have control of this by editing the schemas in the directory under datamodels. It is also possible to pass a custom schema to datamodels when opening a file and when I find the time I will document this.  The environment variable PASS_INVALID_VALUES has the same effect as the keyword parameter pass_invalid_values and overrides it if both are set. It takes an integer value, 0 is false and any other value (typically 1) is true.

PR #776, opened 2017-04-04
transforms: domain to bounding_box

nden:
Switch `jwst.transforms` to using transforms schema v 1.1


nden:
Hm, the tests are set to run with the `jwst` package, and there`s no way currently to run them with the master branch of `asdf` and `gwcs`. Instead of waiting for those two dependencies to get into th e`jwst` package and then test this PR (while everything failing because jwst/master does not work with the new code) I'll merge this one now so it gets with the other packages in the nightly build. If there are problems will resolve them tomorrow in a new PR.

PR #777, opened 2017-04-04
This update fixes the problem reported by Mike Wolfe (ticket 9952). T…

dmggh:
…his occured in OLS processing of the optional output for a very noisy dataset using a low CR rejection threshold, which resulted in many alleged cosmic rays detected (some in consecutive reads). The problem occurred in the ramp fitting code that calculates the maximum number of semiramps. This code update fixes the problem, and makes this calculation more robust.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit_step.py
+++ b/jwst/ramp_fitting/ramp_fit_step.py
@@ -19,15 +19,12 @@ class RampFitStep (Step):
 
     spec = """
         int_name = string(default='')
-        save_opt = boolean(default=False) # Save optional output
+        save_opt = boolean(default=False) # Save optional output (Boolean)
         opt_name = string(default='')
-        algorithm = option('OLS', 'GLS', default='OLS') # 'OLS' or 'GLS'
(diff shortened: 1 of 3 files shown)


jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -19,16 +19,13 @@ class RampFitStep (Step):
 
     spec = """
         int_name = string(default='')
-        save_opt = boolean(default=False) # Save optional output
+        save_opt = boolean(default=False) # Save optional output (Boolean)
```
hbushouse: Was just reminded after trying to run this code for the first time that at least on my systems I always get an argument parsing error when there's a second set of parentheses used in argument spec:

ValueError: Config parameter u'steps/ramp_fit/save_opt': the value "False) # Save optional output (Boolean" is of the wrong type.

Getting rid of the parentheses around "Boolean" makes the parser happy. Of course the actual spec for the argument already indicates that the type is boolean, so it's kind of redundant to have it in the comment portion too. This parsing problem is why it was removed from the previous version of the code.

PR #778, opened 2017-04-05
Update jump step default rejection threshold

### LINKED ISSUE #706: Jump step default rejection threshold
The jump_step python module has the default value for the rejection_threshold parameter set to 4, while the value listed in the default jump.cfg file is 5. The cfg file should be updated to match the code module (for all parameter values).

hbushouse:
Updated the default value of the rejection_threshold parameter in the jump.cfg file from 5.0 to 4.0, so that it matches the default value set in the jump_step.py module. Fixes #706.


hbushouse:
The usual memory error limit in Travis-CI, so merging.

PR #779, opened 2017-04-05
Update NIRCam photom table to include order number

### LINKED ISSUE #760: Add Order column to NIRCam photom table in data model
The NIRCam grism modes will use flux sensitivity/response vectors that are order-dependent, just like NIRISS SOSS and grism modes. So we need to add an entry to the NIRCam photom table for an integer column "ORDER". We also need to increase the default length of the Wavelength and Relresponse arrays, which are currently set to only 50. Waiting for info from Bryan Hilbert as to what a good max value should be for their length (probably on the order of 1000-1500).

hbushouse:
Updated the NIRCam photom table schema to include a column for spectral order number (for the WFSS modes) and also increased the length of the wavelength and relresponse arrays to 3000 (Bryan Hilbert has constructed sensitivity curves for the grisms that are ~2k in length).

Also added column info to the photom data model for each instrument-specific photom table, so that it shows up in the data model docs.

Fixes #760.


hbushouse:
Everything looks OK now (except for the 1 expected Travis-CI error due to memory), but will wait to merge until Bryan Hilbert has a chance to update and redeliver the existing NIRCam photom ref files in the new format, in order to avoid failures in regression tests and anyone else using the dev version of the pipeline.

hbushouse:
The updated NIRCam photom files that Bryan created have the columns defined in a different order than what's in our nircam_photom.schema.yaml file, which if I remember correctly was not a problem in the past (the correct FITS table columns, based on the TTYPEn names, would get loaded into the datamodel table columns of the same name). But now it is a problem, which was causing the wrong data values to get loaded into the datamodel table columns. So I rearranged the order of the column definitions in the schema to match the actual FITS tables. Now the new photom files (delivered in CRDS_CONTEXT 317) work correctly.

philhodge:
That sounds like a problem with the datamodels code.  IMHO it should use 
the column names (case insensitive), not column number.

Phil

hbushouse:
Agreed. I know that in the past we had situations where a new type of column was needed and that new column was defined in our schema somewhere in the middle of the list of existing columns. But meanwhile the new column of data was tacked onto the end of the actual BINTABLE in the FITS file in CRDS and everything seemed to work OK. That doesn't seem to be the case anymore, for some reason.

hbushouse:
Just the usual expected Travis failure due to memory limit. Merging.

Issue #754, opened 2017-03-25
NIRSPEC WCS: remove hard-coded XanYan--> V2V3 transform


nden:
The Xan,Yan to V2V3 transform is hard-coded in the WCS pipeline because transforms in earlier model deliveries went to XanYan instead of V2V3.
This is corrected in the latest delivery of the NIRSPEC model and the transform should be removed from the WCS pipeline.

nden:
Fixed in #782

PR #782, opened 2017-04-05
Remove hardcoded XYan to V2V3 transform

nden:
NIRSpec CDP3 delivery includes the transform from XanYan to V2V3 in the reference files now.
The V2V3 values need to be in arcsec. I need to confirm this is the case before merging this PR.


nden:
Fixes #754

PR #794, opened 2017-04-07
Add BKGLEVEL to the core schema.

mcara:
This addresses https://github.com/STScI-JWST/jwst/issues/788#issuecomment-292516180 and https://github.com/STScI-JWST/jwst/issues/788#issuecomment-292569709

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -1556,4 +1556,8 @@ properties:
         title: Computed/Matched background level
         type: float
         fits_keyword: BKGLEVEL
+      bkgsub:
+        title: Has background level been subtracted from data?
+        type: boolean
+        fits_keyword: BKGSUB
 $schema: http://stsci.edu/schemas/fits-schema/fits-schema
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -1552,12 +1552,46 @@ properties:
             title: Coronagraphic PSF Subtraction
             type: string
             fits_keyword: S_KLIP
-      bkglevel:
-        title: Computed/Matched background level
-        type: float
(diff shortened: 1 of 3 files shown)


mcara:
CC: @hbushouse

hbushouse:
I just noticed/realized that we have a lev3_product schema file that already contains a bunch of keyword definitions related to skymatch (see the "skymatch" section of the schema). We should just go ahead and put this in that schema instead, that way all of the skymatch-related keywords are in one place, both in our schemas and in the organization of the resulting FITS header. Of course right now it looks like that schema is only included by the DrizProductModel. You probably need it go into the individual ImageModels (or CubeModels) that skymatch is working with, right?

mcara:
@hbushouse My second commit adds ``BKGSUB`` as discussed offline.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1552,4 +1552,12 @@ properties:
             title: Coronagraphic PSF Subtraction
             type: string
             fits_keyword: S_KLIP
+      bkglevel:
```
hbushouse: I would suggest putting these 2 keywords into their own sub-section of the meta data, so that their corresponding FITS keywords also appear under a separate section heading in the FITS header, e.g.
`    skymatch:
        title: Skymatch background information
        type: object
        properties:
          bkglevel:
            title: Computed/matched background level
            type: float
            fits_keyword: BKGLEVEL
          bkgsub:
            title: Background level subtracted from data?
            type: boolean
            fits_keyword: BKGSUB`

Of course that means you'll also then need to update the way you reference them in the skymatch_step.py code, to include the extra layer of meta hierarchy, i.e. "image.meta.skymatch.bkglevel" and "image.meta.skymatch.bkgsub".

mcara:
@hbushouse Please review latest changes that provide a "unified" background section for multiple types of images.

PR #795, opened 2017-04-07
Allow (i.e., do not overwrite) custom group IDs in models

mcara:
Proposed changes will allow users to set their own custom ``group_id``s. I found this to be very useful for testing and debugging.


jdavies-st:
Hmm, looks like a test failed.  Let me look closer at this.

nden:
@mcara This broke master. Can you take a look? May be revert it?

PR #801, opened 2017-04-11
Fixes the pedestal calculation in which the dataset's NFRAMES and GRO…

dmggh:
…UPGAP values were not correctly used in calculating the effective integration time; corrects a bug reported by Bryan Hilbert - Ramp-fitting with grouped data.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -191,8 +191,11 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
 
     # Calculate effective integration time (once EFFINTIM has been populated
     #   and accessible, will use that instead), and other keywords that will
-    #   needed if the pedestal calculation is requested
-    effintim, nframes, groupgap= utils.get_efftim_ped(model)
+    #   needed if the pedestal calculation is requested. Note 'nframes'
+    #   is the number of given by the NFRAMES keyword, and is the number of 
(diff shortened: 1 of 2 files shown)


hbushouse:
Based on further conversation at the basecamp page referenced above, it appears that the generalized expression for the effective integration time of group 1 is:

`t_group1 = ((nframes+1)/2 + dropframes1) * tframe`

where tframe is the frame time in secs.

Hence if tframe were available in utils.py, and the slope were in units of DN/sec, the expression in line 441 could be:

`ped = ff_all - slope_int[num_int, : :] * ((nframes+1)/2 + dropframes1)*tframe`

where the term that's being multiplied by the slope is simply the actual integration time of group 1. However, I believe that the the slope value at that point in the code is still in units of DN/group (correct?) and the tframe value is also not available, hence line 441 would need to be:

`ped = ff_all - slope_int[num_int, : :] * (((nframes+1)/2 + dropframes1) / (nframes+groupgap))`

where now the term being multiplied by the slope is the ratio of the group1 integration time to the integration time of all other groups. So the slope value is effectively being rescaled to account for the fact that the integration time of group1 is potentially not the same as the remaining groups.

This obviously requires knowing dropframes1, which we don't currently have in the product headers (we only have groupgap = dropframes2). So I would suggest that for now the dropframes1 argument is included in this calculation, which means it'll need to be passed down from get_efftim_ped along with nframes and groupgap, but simply hardwire it's value to zero in get_efftim_ped until it becomes available in the meta data.

PR #803, opened 2017-04-12
Add astropy registry support to datamodels and write support to nddata interface

bernie-simon:
The astropy registry interface provides a generic way to read and write different astronomical images and datamodels has new functions to support it, read, write, and identify. Also, setters have been to the getters in the nddata interface. Most of the code to implement the astropy registry and nddata nterface has been moved into a new file, ndmodel.py. There is still some glue code in __init__.py, model_base.py, and properties.py. DataModel now subclasses a new class, NDModel, a mixin that implements the nddata interface.


nden:
One test failed consistently through all configurations. Perhaps worth investigating?

But where is this headed at this time? More specifically a model now has two types of WCS, one of which is incorrect because the NDData interface does not support gwcs yet.

bernie-simon:
I looked at the failing tests and they fail because of an argument list mismatch. Here is a typical case:

> test_modelcontainer_iteration() takes exactly 1 argument (0 given)

James will have to look at these tests and fix them. Or perhaps has already has and the fix has not been merged yet.

The code I wrote merely provides an alternative interface to datamodels that conforms to the api in astropy.io.registry. If you look at the source you will see it's all simple glue code that wraps our existing datamodels interface. I simply called the wcs interface methods currently in datamodels, the new code is no more correct or incorrect than the existing code. If there is a better way to access the wcs, updating the current datamodels methods should fix it and the new astropy.io.registry api.

nden:
I'm not saying the code is wrong. I was asking for clarification on the general direction because it is confusing to have, for example

```
print(model.wcs)
Number of WCS axes: 2
CTYPE : ''  ''  
CRVAL : 0.0  0.0  
CRPIX : 0.0  0.0  
PC1_1 PC1_2  : 1.0  0.0  
PC2_1 PC2_2  : 0.0  1.0  
CDELT : 1.0  1.0  
NAXIS : 0  0
```
and
```
print(model.meta.wcs)
  From       Transform     
-------- ------------------
detector 3-order SOSS Model
   world               None

```
They are different and one of them is incorrect.

bernie-simon:
With your help I would want to revise the incorrect method to call the correct method and perhaps mark it as deprecated.

nden:
Isn't it better to have a full gwcs suport in nddata? I always thought that's the general goal.

jdavies-st:
I've fixed the offending test in PR #810 

I think it would be good to not merge though until all tests pass.

jdavies-st:
Also, it might be worth reverting this PR until we get the WCS stuff sorted out, especially since we are doing a DMS delivery tomorrow, and it would be best to avoid confusion.

Perhaps a discussion amongst the JWST developers and astropy developers on how to best do this would be useful?  Perhaps that involves gwcs support and asdf support in the IO registry on the astropy.nddata side instead of on the jwst side as @nden mentions above?  Not sure.

hbushouse:
I agree that we might not want to merge this until after we've given DMS a drop of the master in the next day or so, just to avoid any unnecessary unpleasantries.

bernie-simon:
The nddata interface is mostly a collection of getters and setters, one of which is for the wcs. I merely hooked up the getter and setter in model_base.py to the nddata interface by making it a property on the ndmodel mixin. The interface to nddata is just as broken or working as the datmodels code, because it's nothing more than a thin facade over it and there's no more reason to hold off on it than to retire datamodels. When we come up with a solution to the wcs problem that makes everyone happy we can apply it to datamodels and the nddata interface will automatically get it.

PR #811, opened 2017-04-13
add subarray transform

nden:
This PR adds support for subarrays for all instruments. Before this only Nirspec supported subarray mode.
In addition this fixes a problem with the order of inputs to the NIRISS SOSS wavelength solution mode
reported by the NIRISS team and debugged by @philhodge .


hbushouse:
I think Nadia is away this evening and I'd like to get this tested in advance of our hoped for dev delivery to DMS tomorrow, so I'm merging.

PR #812, opened 2017-04-13
skip failing test

nden:
This skips the always failing tests in exp_to_source when the tests are run under python 2.7. In this way at least the rest o fthe tests get run under this configuration.


hbushouse:
I think Nadia is away this evening, so I'm going to merge (in her absence).

PR #816, opened 2017-04-14
correct size of extracted slit

nden:
Since the limits of the `bounding_box` are now inclusive the size of the extracted slit needs to be increased by 1.

### DIFF
--- a/jwst/extract_2d/extract_2d.py
+++ b/jwst/extract_2d/extract_2d.py
@@ -42,8 +42,6 @@ def extract2d(input_model, which_subarray=None):
 
         for slit in open_slits:
             slit_wcs = nirspec.nrs_wcs_set_input(input_model, slit.name)
-                      str(slit_wcs.bounding_box[0]),
-                      str(slit_wcs.bounding_box[1]))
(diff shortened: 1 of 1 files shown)


jwst/extract_2d/extract_2d.py
```diff
@@ -16,7 +16,7 @@
 
 
 log = logging.getLogger(__name__)
-log.setLevel(logging.DEBUG)
+log.setLevel(logging.INFO)
```
hbushouse: Actually I believe we need to leave them set to DEBUG level. It was discovered years ago, by Mike D. and myself, that for some reason this setting does not set the default level that will be used whenever the module is executed, but rather behaves more like an upper-limit as to what level of messages are passed through. So if you set it to INFO, debug level messages will not appear, even when the user requests them. If you set it to DEBUG, you only actually see debug level messages when the user requests them. That's why I set all of these values in the various step modules to DEBUG several years ago.

nden:
@hbushouse @jdavies-st Just to be clear. This isn't fixng the error of validating xstart/ystart. I think it should be fixed in the validation code.

nden:
@hbushouse reverted back to DEBUG level and merging

PR #828, opened 2017-04-17
fix units in computation of roll_ref

nden:
Similar code is in assign_wcs. The model there uses the correct units. This was missed when the codde was copied here.


nden:
Updated tests, new values are consistent with results from Colin's code.

jdavies-st:
Ran this code on all the NIRCam CV3 test data.  Gives results that look correct - no 180 degree flip between PA_V3 and ROLL_REF.  *thumbs up*

nden:
This is passing on my personal Travis so merging.

Issue #647, opened 2017-01-26
Task: Update NIRSpec IFU flux calibration


hbushouse:
As discussed in #611, the MIRI MRS flux calibration is going to be using 2D reference data in the photom file, as opposed to the other spectroscopic modes where a 1D function of response vs. wavelength is used. The NIRSpec IFU team says that their data also need to use a 2D response array that's divided into the science data in order to calibrate it. The NIRSpec IFU mode, however, will continue to use the 1D spectral response function (RSRF) that's stored in the photom reference file, but instead of simply attaching that 1D function to the science data as is currently done, the photom step needs to use that information to construct, on the fly, a 2D response array that gets attached to the science data. The 2D array is computed by evaluating the 1D RSRF at the wavelength of each pixel in the 2D science data.

The attached 2D response array will then be used in the subsequent cube_build step to calibrate the 2D science data before it is used to build a cube.

hbushouse:
In addition to constructing a 2D spectral response function from the 1D RSRF in the photom ref file, it is also necessary to divide by pixel area, in order to place the science data in units of surface brightness, for use in cube_build. The IDT delivered 1 example pixel area reference file in CDP3:

file:///grp/jwst/wit4/nirspec/CDP3/05_Other_Calibrations/5.4_Pixel_Area_Map/jwst-nirspec-clear-prism.parf.fits

This ref file is for the filter/grating combination CLEAR+PRISM. There will be one pixel area ref file for each of the available filter/grating combinations (the IDT needs to deliver the rest). Each ref file contains 3 table extensions, the first giving pixel areas for the fixed slits, the second giving pixel areas for the MOS shutters, and the 3rd giving pixel areas for the IFU slices.

The NIRSpec IFU portion of the photom step must load the pixel area reference data for the IFU slices from this type of reference file and construct a 2D pixel area map, where the pixel area within a given IFU slice is a constant.

The 2D spectral response array created by photom will then be formed as the combination of:

  scalar flux conversion factor (photmjsr) * 2D response / 2D pixel area map

The photmjsr value is retrieved, as usual, from the row in the photom ref table that matches the filter/grating combination for the exposure.

hbushouse:
All updates to the photom step, as well as support for the new NirspecAreaModel data model for the spectroscopic-mode pixel area reference files are contained in #832. Waiting for the new spectroscopic pixel area reference files to be implemented in CRDS before closing, in case there are last minute tweaks needed for the NirspecAreaModel.

[...]

PR #832, opened 2017-04-19
WIP: Update photom for NIRSpec IFU mode

hbushouse:
Initial round of updates in support of #647. This includes updates to the photom step to compute and apply a 2D spectral response array for the NIRSpec IFU mode. Because this operation also requires the use of pixel areas, a data model and schema have also been created for NIRSpec pixel area reference files, which haven't been used before.

The NirspecAreaModel, for the pixel area reference file, assumes the ref files contain 3 table extensions, one for SLIT mode, one for MOS, and one for IFU. The tables in each extension have slightly different formats. The SLIT and IFU tables are pretty simple, just giving pixel area for each fixed-slit or IFU slice. The MOS table gives pixel area for each shutter, based on the MSA quadrant and x/y indexes of each shutter within the quadrant.

The update to photom first loads the scalar flux conversion factor (photmjsr) and 1D relative spectral response function (RSRF), as usual, from the row of the photom ref table that matches the filter+grating of the exposure. A 2D wavelength array is computed using methods stolen from the NIRSpec IFU portion of the flat_field step. A 2D pixel area map is computed at the same time, storing the appropriate pixel area value from the area ref file in the region of each IFU slice. The 1D RSRF is then evaluated at the wavelength of each pixel in the entire 2D image, resulting in a full-frame 2D sensitivity array. That 2D array is multiplied by photmjsr and divided by the 2D pixel area map, resulting in a final 2D sensitivity array that when divided into the science data will result in pixels having units of surface brightness (mJy/arcsec^2). The 2D sensitivity array is applied to the science data and err arrays, and is also stored in the RELSENS2D extension of the science product.


jwst/photom/photom.py
```diff
@@ -453,6 +513,98 @@ def calc_fgs(self, ftab):
...
+                xstart = 0.
+            if ystart < -0.5:
+                log.debug("ystart from WCS bounding_box was %g" % ystart)
+                ystart = 0.
+            if xstop > 2047.5:
+                log.debug("xstop from WCS bounding_box was %g" % xstop)
+                xstop = 2047.
+            if ystop > 2047.5:
+                log.debug("ystop from WCS bounding_box was %g" % ystop)
+                ystop = 2047.
```
nden: I believe the WCS is also checking for off the edge values when computing the bounding box. This check is probably not necessary.

jwst/photom/photom.py
```diff
@@ -453,6 +513,98 @@ def calc_fgs(self, ftab):
...
+
+            # Convert these to integers, and add one to the upper limits,
+            # because we want to use these as slice limits.
+            xstart = int(math.ceil(xstart))
+            xstop = int(math.floor(xstop)) + 1
+            ystart = int(math.ceil(ystart))
+            ystop = int(math.floor(ystop)) + 1
```
nden: This should be OK but I've been thinking I should talk with you about how to deal with the edges of pixels consistently.

jwst/photom/photom.py
```diff
@@ -453,6 +513,98 @@ def calc_fgs(self, ftab):
...
+
+            # Construct array indexes for pixels in this slice
+            dx = xstop - xstart
+            dy = ystop - ystart
+            ind = np.indices((dy, dx))
+            x = ind[1] + xstart
+            y = ind[0] + ystart
```
nden: The indices can be constructed with a call to 
`x, y = gwcs.wcstools.grid_from_bounding_box(bounding_box, step=(1,1), center=True)`

jwst/photom/photom.py
```diff
@@ -453,6 +513,98 @@ def calc_fgs(self, ftab):
...
+
+            # Get the world coords for all pixels in this slice
+            coords = ifu_wcs(x, y)
+
+            # Pull out the wavelengths only
+            wl = coords[2]
+            nan_flag = np.isnan(wl)
+            good_flag = np.logical_not(nan_flag)
+            if wl[good_flag].max() < MICRONS_100:
+                log.info("Wavelengths in WCS table appear to be in meters")
```
nden: Early on the pipeline produced wavelengths in meters as this was how the models were made. Now wavelengths are in microns. Not sure if we need to keep checking this.

jwst/photom/photom.py
```diff
@@ -347,14 +407,14 @@ def calc_miri(self, ftab):
                                                 dqflags.pixel['NON_SCIENCE'])
 
             # Divide the science data and err by the conversion factors
-            self.input.data /= ftab.data*ftab.pixsiz
-            self.input.err /= ftab.data*ftab.pixsiz
+            self.input.data /= ftab.data * ftab.pixsiz
+            self.input.err /= ftab.data * ftab.pixsiz
```
philhodge: Just a suggestion, but since you compute ftab.data * ftab.pixsiz a few lines below, maybe you could move that line farther up and then divide data and err by self.input.relsens2d, instead of duplicating the arithmetic.

philhodge:
This looks good to me.

hbushouse:
Updates to the nirspec_area.schema.yaml in support of #861

jwst/photom/photom.py
hbushouse: Removed.

jwst/photom/photom.py
hbushouse: Leaving as-is for now.

hbushouse: Actually, removing entirely, because by using "grid_from_bounding_box" (below), there's no need anymore for any xstart, ystart, xstop, ystop values.

jwst/photom/photom.py
hbushouse: Thanks, that really simplifies the code!

jwst/photom/photom.py
hbushouse: Leaving for now, just in case.

jwst/photom/photom.py
hbushouse: Done

hbushouse:
Based on testing, this work now appears to be complete. Just waiting for the new spectroscopic pixel area reference files to get installed in CRDS before merging the code, so that nothing breaks in the meantime.

hbushouse:
New NIRSpec spectroscopic area reference files are available starting with CRDS_CONTEXT jwst_0333.pmap. That makes it possible to finally merge the new code and allow it to get the ref files it needs.

PR #821, opened 2017-04-14
Complete Level2 refactor + Background Candidate/Chop mode

stscieisenhamer:
*DO NOT MERGE*
- [ ] Waiting for software delivery. Putting this out for testing at least.
- [ ] Need to implement #825 

This PR represents the refactor to handle Level2 associations. Also, the Background Candidate/CHOP associations are handled.

The level2 schema has changed, hopefully final change. However, once merged, anyone with level2 associations will need to update.

Issue #750, opened 2017-03-23
Update calwebb_image2 to take Level2 association as input


stscieisenhamer:

hbushouse:
It would be good to do this in a way that makes use of commonalities between multiple pipelines, like calwebb_spec2. In other words, refactor things like the "Lvl2Input" function in calwebb_spec2 to make it a generic routine that could live within a common library somewhere (perhaps either within the associations or pipelines sub-packages?) and be used by both calwebb_image2 and calwebb_spec2. It needs to have the capability to handle multiple input use cases, including

1) an ASN file name
2) a single exposure file name
3) a single exposure data model

hbushouse:
This work could also include #687.

stscieisenhamer:
# Implementation

- [ ] Tests
    - [ ] use the TEST_BIGDATA environment as is done on the pandokia tests.
    - [ ] modify the spec2 tests also.
- [ ] `calwebb_image2.py`
    Modify as `calwebb_spec2.py`
- [ ] docs
    - [ ] image2
    - [ ] spec2

stscieisenhamer:
Not sure why closed; must have had a wayward issue number somewhere.

PR #833, opened 2017-04-19
Update calwebb_spec2 to handle general Level2 input

stscieisenhamer:
Implementation is done through the `Level2Association` class that allows a variety of inputs as specified in #750 

This PR supercedes PR #821. As with the previous PR, since this involves a fundamental change in how `calwebb_spec2` runs and in the basic format of Level2 associations, merging will only occur after review and independent confirmation


stscieisenhamer:
@hbushouse A couple of items.

First: If none of the Pandokia test suite WRT `calwebb_spec2` use associations, then this PR could be merged. Otherwise, I'll troll through them and fix the associations before merging. Any pointers to such tests would be greatly appreciated.

Second: Concerning conversion of level2 associations that folks may be using, there is a module, `jwst.associations.lib.lv2_v1_to_v2` which should make converting much easier.

Anyone else who may have a vested interest in this should add themselves to the reviewers.

hbushouse:
There are a few calwebb_spec2 tests in the reg tests. They would be in the jwcalibdev /data1/jwst_rt/general/pipelines/ area. Not all of them use ASN's as input (they use a single file). But I'm sure there is at least 1 test that does use an ASN (example of MIRI LRS slit background nodding). Just look for any _asn.json files over in the /data4/jwst_test_data/ subdirectories that would be used as input.

docs/associations/source/reference/level2_asn_technical.rst
```diff
@@ -0,0 +1,151 @@
...
+
+Level 2 Associations: Technical Specifications
+==============================================
+
+Logical Structure
+-----------------
+
+Independent of the actual format, all Level 3 associations have the
+following structure. Again, the structure is defined and enforced by
+the Level 3 schema
```
hbushouse: I think you mean Level **2** in lines 9 and 11?

docs/associations/source/reference/level2_asn_technical.rst
```diff
@@ -0,0 +1,151 @@
...
+Example Association
+-------------------
+
+The following example will be used to explain the contents of an association::
+
+    {
+        "asn_rule": "Asn_Lv2Image",
+        "asn_pool": "jw82600_001_20160304T145416_pool",
+        "program": "82600",
+        "asn_type": "image2",
```
hbushouse: Not a really big deal, but since the example files reference "lrs", which are spectra, it might be more appropriate to label the asn_type of this example as "spec2", rather than "image2."

docs/associations/source/reference/level2_asn_technical.rst
```diff
@@ -0,0 +1,151 @@
...
+        "products": [
+            {
+                "name": "test_lrs1_rate",
```
hbushouse: Why do the input member names and output product names both have the "_rate" suffix? Wouldn't only inputs have this suffix and then the pipeline will determine a new suffix for the products?

docs/associations/source/reference/level2_asn_technical.rst
```diff
@@ -0,0 +1,151 @@
...
+
+expname *required*
+  The exposure file name
+
+exptype *required*
+  Type of information represented by the exposure. Possible values are
+
+  * `SCIENCE`
+  * `BACKGROUND`
+  * `LEAKCAL`
```
hbushouse: The step that subtracts the "leakcal" exposures is known as the "imprint" subtraction step and the effect has already been referred to - at least within calibration circles - as "imprint." I think the "leakcal" term is terminology from the APT/PPS folks? As far as I can tell the NIRSpec folks have always called it "imprint."

jwst/associations/lib/asn_schema_jw_level2b.json
```diff
@@ -3,7 +3,7 @@
     "title": "Association: JWST DMS Level2a->Level2b",
     "description": "The data structure that, within the JWST DMS processing system, defines how to group level2a exposures to produce Level2b data products.",
     "type": "object",
-    "required": ["asn_type", "asn_rule", "asn_pool", "members"],
+    "required": ["asn_type", "asn_rule", "asn_pool", "products"],
```
hbushouse: Why is asn_rule required? What do we use it for?

jwst/pipeline/calwebb_spec2.py
```diff
@@ -43,179 +48,226 @@ class Spec2Pipeline(Pipeline):
...
+        if len(science) != 1:
+            log.warn(
+                'Wrong number of science files found in {}'.format(
+                    exp_product['name']
+                )
+            )
+        science = science[0]
+
+        self.log.info('Working on input %s ...', science)
+        if isinstance(science, datamodels.DataModel):
```
hbushouse: Is this if-block necessary? I believe this functionality is already built into datamodels.open, such that it tests to see if the input is a string or data model. If it's a string, it loads it into a data model, else if it's already a data model, it simply returns without doing anything.

jwst/pipeline/calwebb_spec2.py
```diff
@@ -43,179 +48,226 @@ class Spec2Pipeline(Pipeline):
...
+            if self.save_bsub:
+                if isinstance(input, datamodels.CubeModel):
+                    self.save_model(input, "bsubints")
+                else:
+                    self.save_model(input, "bsub")
+
+        # Apply NIRSpec MSA imprint subtraction
+        # Technically there should be just one.
+        # We'll just get the first one found
+        imprint = members_by_type['IMPRINT']
```
hbushouse: Yeah, see here we're using the label "IMPRINT", not "LEAKCAL". So the earlier instance of LEAKCAL in the asn properties should be changed to IMPRINT.

jwst/pipeline/tests/test_calwebb_spec2.py
```diff
@@ -0,0 +1,63 @@
...
+
+def abspath(filepath):
+    """Get the absolute file path"""
+    return path.abspath(path.expanduser(path.expandvars(filepath)))
+
+
+DATAPATH = abspath(
+    '$DEVDIR/testdata/jwst_data/dev/build7.1/spec2_test'
+)
+EXPFILE = 'jw00035001001_01101_00001_mirimage_rate.fits'
```
hbushouse: You can't use a MIRI image exposure to test calspec2. Unless of course this happens to be an LRS spectrum, which is taken on the imager detector. What's the exp_type?

stscieisenhamer:
Hmmm, not finding any asn files in the test suite.

hbushouse:
Looks like there's only 1 test*.py module that references an asn file as input:

/data1/jwst_rt/general/pipelines/test_miri_lrs_bkgnod.py

which uses /data4/jwst_test_data/pipelines/lrs_bkgnod_asn.json

stscieisenhamer:
Ahhh, lookn' in the wrong place.

jwst/pipeline/level2association.py
```diff
@@ -0,0 +1,100 @@
+"""Open a Level2 association given various inputs"""
```
hbushouse: Would it perhaps make more sense to have this module live within /associations? It's currently used by pipelines that live here, but theoretically it could be used by anyone, anywhere.

jwst/pipeline/calwebb_spec2.py
```diff
@@ -43,179 +48,226 @@ class Spec2Pipeline(Pipeline):
...
-                input.meta.asn.table_name = input_table.filename
-            else:
-                input.meta.asn.pool_name = ' '
-                input.meta.asn.table_name = ' '
+        asn = Level2Association.open(input)
+
+        # Each exposure is a product in the association.
+        # Process each exposure.
+        for product in asn['products']:
+            log.info('Processing exposure {}'.format(product['name']))
```
hbushouse: Nomenclature question: We are technically processing individual exposures at level-2b, rather than a collection of exposures, but the name that's going to appear in this log line will be the name of the output product that's in the process of being created, so shouldn't we perhaps change the message to read "Processing **product** ..."

jwst/pipeline/calwebb_spec2.py
```diff
@@ -43,179 +48,226 @@ class Spec2Pipeline(Pipeline):
...
+            log.info(
+                'Saved extracted spectrum to %s' % x1d_output.meta.filename
+            )
+
+        input.close()
+        x1d_output.close()
+
+        # That's all folks
+        log.info(
+            'Finished processing exposure {}'.format(exp_product['name'])
```
hbushouse: Same comment as earlier: perhaps this should say "Finished processing **product** ..."

jwst/pipeline/calwebb_spec2.py
```diff
@@ -43,179 +48,226 @@ class Spec2Pipeline(Pipeline):
...
+        )
         return
```
hbushouse: This old Lvl2Input function is no longer needed and should be deep-sixed.

jwst/pipeline/level2association.py
stscieisenhamer: Because there is a touch of datamodel dependency in there, it seemed more appropriate for it to be more closely tied to the pipeline/model code. But, considering this is all JWST, the presumption that datamodels are baked into everything can be a given, allowing this to be considered a generic function.

docs/associations/source/reference/level2_asn_technical.rst
stscieisenhamer: 👍

docs/associations/source/reference/level2_asn_technical.rst
stscieisenhamer: 👍

docs/associations/source/reference/level2_asn_technical.rst
stscieisenhamer: 👍

docs/associations/source/reference/level2_asn_technical.rst
stscieisenhamer: Imprint it is. 👍

jwst/associations/lib/asn_schema_jw_level2b.json
stscieisenhamer: 👍

jwst/pipeline/calwebb_spec2.py
stscieisenhamer: @bernie-simon @jdavies-st Is this true? Is re-opening a datamodel a null operation?

jwst/pipeline/calwebb_spec2.py
stscieisenhamer: 👍

jwst/pipeline/calwebb_spec2.py
stscieisenhamer: 👍

jwst/pipeline/tests/test_calwebb_spec2.py
stscieisenhamer: fascinating, exp_type is MIR_LRS-FIXEDSLIT. not sure where i got this example from. It runs without error; _cal and _x1d is created.

jwst/pipeline/level2association.py
stscieisenhamer: So, I did everything: Genericized the whole thing and put it back into associations. Will be evident in the push that is about to happen...once the @#$@#$@ tests run.

jwst/pipeline/tests/test_calwebb_spec2.py
hbushouse: That's appropriate. The LRS spectra fall on the imager detector, so the file name will have the field "mirimage", because that's the name of the detector (like nrc1a, nrcblong, guider1). Coronagraphic data also fall on the imager detector. So the down side to this is that you can't tell just by glancing at the file name whether you've got an LRS fixed-slit, LRS slitless, coronagraphic, or imaging mode exposure. You have to look inside at exp_type.

jwst/pipeline/level2association.py
hbushouse: Fascinating. Is "@#$@#$@" a new kind of command that automatically gets fed to Travis CI?

jwst/pipeline/tests/test_calwebb_spec2.py
stscieisenhamer: Call shenanigans because, well, aren't all 2d spectra falling on image detectors really? But understood.

hbushouse: No, not all. MIRI MRS spectra use their own dedicated detectors (IFUSHORT, IFULONG). NIRISS and NIRCam, on the other hand, project both direct imaging and grism spectra onto the same detectors (like MIRI imaging and LRS spectra).

stscieisenhamer:
Closing due to some further optimizations done, and now the documentation restructure. New PR will be created to include both.

PR #835, opened 2017-04-20
Fix PC matrix computation

jdavies-st:
For cases where quaternions are not found in DB:

- PA_V3 is zero, but compute ROLL_REF correctly for the detector at hand
- Use the correct ROLL_REF and V3I_YANG to compute the angle for
  use in PC matrix

Correct the parity of the PC matrix

Make log messages a bit more informative.

### DIFF
--- a/scripts/set_telescope_pointing.py
+++ b/scripts/set_telescope_pointing.py
@@ -257,40 +257,43 @@ def add_wcs(filename):
     except ValueError as exception:
         logger.warning(
             'Cannot retrieve telescope pointing.'
-            '\nUsing proposal values.'
-            '\nFailure {}'.format(exception)
+            '\n{}'
+            '\nUsing TARG_RA, TARG_DEC and PA_V3=0 '
+            'to set pointing.'.format(exception)
         )
         ra = pheader['TARG_RA']
         dec = pheader['TARG_DEC']
         roll = 0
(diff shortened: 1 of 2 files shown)


scripts/set_telescope_pointing.py
```diff
@@ -257,40 +257,43 @@ def add_wcs(filename):
...
-    v1_ra_deg, v1_dec_deg, v3_pa_deg = vinfo
-    # update header
     pheader['RA_V1'] = v1_ra_deg
     pheader['DEC_V1'] = v1_dec_deg
     pheader['PA_V3'] = v3_pa_deg
     pheader['CRVAL1'] = crval1
     pheader['CRVAL2'] = crval2
-    pheader['PC1_1'] = np.cos(pa_aper_deg * D2R)
-    pheader['PC1_2'] = -np.sin(pa_aper_deg * D2R)
+    pheader['PC1_1'] = -np.cos(pa_aper_deg * D2R)
```
hbushouse: So is this the fix that will finally make E point left, when N is up, instead of right?!

jdavies-st: Yes, now the PC matrix (if read in to an astropy.wcs object) will give roughly the same ra/dec results as the more accurate gwcs object produced by `assign_wcs`.

Issue #840, opened 2017-04-22
Ramp fitting step: disable unweighted OLS and GLS options


hbushouse:
The JCCWG has decided that the unweighted OLS fitting option and the GLS fitting option should be disabled for B7.1, because the optimally weighted OLS is the preferred approach over unweighted, and the GLS is currently computationally expensive and untested. Hence optimally weighted OLS should become the default method. See https://jira.stsci.edu/browse/JP-136.

This can perhaps be accomplished by simply disabling the optional arguments in ramp_fit_step.py that allow selection of those options.

hbushouse:
Fixed by #842, #873, and possibly others as well. Closing.

PR #842, opened 2017-04-26
Disables unweighted OLS and GLS options in ramp fitting. The only all…

dmggh:
…owed algorithm is ols , and the only allowed weighting is optimal. See ticket #840.


hbushouse:
Oh, and docs will also need to be updated to comment out or remove info concerning these options.

Issue #839, opened 2017-04-22
Jump step: disable y-intercept method


hbushouse:
The JCCWG has decided that the y-intercept method currently available in the jump step should be disabled or somehow made unavailable to users in B7.1, because it is computationally very expensive, has not been thoroughly tested by INS, and will likely be replaced in the future with a different algorithm.

See https://jira.stsci.edu/browse/JP-137

This can perhaps be accomplished by simply disabling the optional argument in jump_step.py that allows the y-intercept method to be turned on.

hbushouse:
Fixed by #846 and #844. Closing.

PR #844, opened 2017-04-26
Disables the y-intercept method in the jump step. The only allowed me…

dmggh:
…thod is the Two-Point Difference Method. Removed the associated parameters do_intercept and yint_threshold. Updated docs. See ticket #839.


jwst/jump/jump_step.py
```diff
@@ -8,17 +8,21 @@
...
-        yint_threshold = float(default=1.0,min=0) # y-intercept signal threshold
     """
 
+    # Prior to 04/26/17, the following were also in the spec above:
+    #    do_yintercept = boolean(default=False) # do y-intercept method?
+    #    yint_threshold = float(default=1.0,min=0) # y-intercept signal threshold
+    # As of 04/26/17, do_yintercept is not an option. Only the 2-point 
+    #   difference method is allowed for Build 7.1.
+    do_yintercept = False  # do_intercept is no longer an option     
+    yint_threshold = 0.0   # placeholder in case algorithm is re-enabled later
```
hbushouse: Very minor nit: the default (and hence hardwired) value for yint_threshold should be 1.0, not 0.0.

Issue #831, opened 2017-04-18
Task: Implement group scale correction step


hbushouse:
As documented by the JCCWG at https://confluence.stsci.edu/display/JWSTCC/Vanilla+Group+Scale+Correction, some exposures can be taken using a value of NFRAMES that is not a power of 2, in which case the division that's done on-board when multiple frames are averaged into a group - which always uses a power of 2 divisor - results in the wrong scaling for the raw pixel values in the entire exposure.

A new correction step, "group_scale", needs to be implemented and placed at the head end of the calwebb_sloper pipeline, in order to rescale the raw pixel values to their proper values. The rescaling is accomplished by multiplying the pixel data array by FRMDIVSR/NFRAMES.

Because this is being applied at the very start of processing, no operations are necessary on the ERR or various dq arrays.

hbushouse:
Closed by #845

PR #845, opened 2017-04-27
Add new group_scale step

hbushouse:
This implements the new group_scale step as specified in #831. It includes the step modules, step config file, and step docs. It also updates the calwebb_sloper and calwebb_dark pipelines, and their config files, to include this new step at the very beginning of processing. Docs for calwebb_sloper and calwebb_dark are also updated.

The group_scale step has logic in it to detect when NFRAMES = FRMDIVSR, which means the data were scaled correctly on-board, and hence doesn't need correcting. In this case the step is skipped and the input is returned unchanged. Hence the calwebb_sloper and calwebb_dark pipelines can call the step for all exposures and only those actually needing the correction will get the rescaling applied.

It also includes updates to the core schema to include the FRMDIVSR keyword (see https://jira.stsci.edu/browse/JWSTKD-14), as well as DRPFRMS1 and DRPFRMS3 (see https://jira.stsci.edu/browse/JWSTKD-47).


jwst/group_scale/group_scale.py
```diff
@@ -0,0 +1,56 @@
...
+    -------
+    output_model: data model object
+        rescaled science data
+
+    """
+
+    # Get the meta data values that we need
+    nframes = input_model.meta.exposure.nframes
+    frame_divisor = input_model.meta.exposure.frame_divisor
+    if (nframes is None) or (frame_divisor is None):
```
philhodge: If frame_divisor is None (but nframes is not), one option would be to check whether nframes is a power of two.  If it is, it's likely (or at least possible) that no correction is needed.  If nframes is not a power of two, you could log an error message saying that a correction couldn't be applied but really needs to be.  You could even SWAG it and set frame_divisor to the power of two that is greater than nframes, but of course print a message saying that you guessed at the value of the missing keyword.

hbushouse: The "nframes is a power of 2" check is done up in the group_scale_step function, but was not replicated here. I sort of figured that this lower-level function would only be called directly by folks who know what they're doing and hence the meta data would be complete and accurate at that point, and if it isn't just punt without attempting any heroic measures. But you never know ...

hbushouse:
Fixes #831

PR #847, opened 2017-04-27
Update main.rst

dmggh:
Updated rampfitting docs to reflect code changes in tickets #840, #842, #843

### DIFF
--- a/docs/ramp_fitting/source/main.rst
+++ b/docs/ramp_fitting/source/main.rst
@@ -12,7 +12,7 @@ data cube in a second output data product.  A third, optional output product
 is also available and is produced only when the step parameter 'save_opt' is
 True (the default is False).  The output values will be in units of counts (DN)
 or counts per second.  Following a description of the fitting algorithm, these
-three type of output files are detailed below.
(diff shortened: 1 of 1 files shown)


docs/ramp_fitting/source/main.rst
```diff
@@ -12,7 +12,7 @@ data cube in a second output data product.  A third, optional output product
 is also available and is produced only when the step parameter 'save_opt' is
 True (the default is False).  The output values will be in units of counts (DN)
```
hbushouse: In what case(s) would the output values ever be in units of counts? I would think that even for pixels that only have 1 valid group, you then divide that value by the group time to get a slope (counts per second).

hbushouse: Also, further up at line 8 there's a reference to the primary output file being "RampFit". I would change that to "rate" or "_rate". And then mention that the output file for multiple integrations is "rateints".

sosey:
@dmggh This file will probably get deleted as soon as the doc restructuring pull request gets accepted. Keep it on hand so we can update the same file in the new tree?

PR #848, opened 2017-04-27
Update main.rst

dmggh:
Update includes text on naming of the final output file: _RampFit if ramp fit is run separately, _rate if ramp fit is run within a pipeline.

### DIFF
--- a/docs/ramp_fitting/source/main.rst
+++ b/docs/ramp_fitting/source/main.rst
@@ -5,14 +5,14 @@ This step determines the mean count rate for each pixel by performing a linear
 fit to the data in the input (jump) file.  The fit is done using "ordinary
 least squares" (the "generalized least squares" is no longer an option).
 The fit is performed independently for each pixel.  There are up to three
-output files. The primary output file (RampFit, giving the slope at each
-pixel) is always produced.  If the input exposure contains more than one
-integration, the resulting slope images from each integration are stored as a
-data cube in a second output data product.  A third, optional output product
-is also available and is produced only when the step parameter 'save_opt' is
-True (the default is False).  The output values will be in units of counts (DN)
-or counts per second.  Following a description of the fitting algorithm, these
-three type of output files are detailed below. 
+output files. The primary output file, giving the slope at each pixel, is 
+always produced.  If the input exposure contains more than one integration, 
+the resulting slope images from each integration are stored as a data cube 
+in a second output data product.  A third, optional output product is also 
+available and is produced only when the step parameter 'save_opt' is True 
+(the default is False).  The output values will be in units of counts per 
+second.  Following a description of the fitting algorithm, these three type of 
+output files are detailed below 
 
 
 The count rate for each pixel is determined by a linear fit to the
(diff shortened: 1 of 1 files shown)


docs/ramp_fitting/source/main.rst
```diff
@@ -37,7 +37,11 @@ a given pixel, the final slope is determined as a weighted average from all
 intervals and is written to a file as the primary output product.  In this
 output product, the 4-D GROUPDQ from all integrations is compressed into 2-D,
 which is then merged (using a bitwise OR) with the input 2-D PIXELDQ to create
-the output PIXELDQ.
+the output PIXELDQ.  If the ramp fitting step is run by itself, the output file 
+name will have the suffix '_RampFit'; if the ramp fitting step is run as the
```
jdavies-st: One slight wrinkle here is that if one runs a step like so:

`strun jwst.rampfit.RampFitStep myfile_uncal.fits`

the output will have the `_RampFitStep` suffix.  If instead you run it using an input config file:

`strun rampfit.cfg myfile_uncal.fits`

it will have a `_RampFit` suffix, i.e. a suffix based off whatever the `name` keyword is set to in the .cfg file.  For other pipeline steps, many of our default config files have inconsistent naming. Some are camel case, some are not.  Might be a good idea to decide on a standard and stick with it.  Maybe lower case?

All this is controlled in step.py in the `__init__()` and `from_config_file()` methods.

hbushouse: Way back, in the primeval days of Chris Hanley and Robert J., camel case was adopted as the default for step class names, due to the fact that most of them consist of multiple words and so the camel case makes the wording obvious.

nden: :+1: for keeping close to PEP8 - camel case for class names, lower case with underscores for methods and functions. As for achieving consistency we can have a `Model.name` attribute used in the naming of the products. We can implement this in our spare time.

jdavies-st: Here's the first line from each of our default step and pipeline `*.cfg` files:
```
name = "alignrefs"
name = "ami_analyze"
name = "ami_average"
name = "ami_normalize"
name = "assign_wcs"
name = "Ami3Pipeline"
name = "Coron3Pipeline"
name = "DarkPipeline"
name = "Image2Pipeline"
name = "Image3Pipeline"
name = "SloperPipeline"
name = "Spec2Pipeline"
name = "combine_1d"
name = "cube_build"
name = "dark_current" 
name = "dq_init"
name = "emission" 
name = "engdblog"
name = "extract_1d"
name = "extract_2d" 
name = "flat_field" 
name = "fringe" 
name = "hlsp"
name = "imprint" 
name = "ipc" 
name = "jump"
name = "klip"
name = "lastframe" 
name = "linearity"
name = "outlier_detection" 
name = "pathloss" 
name = "Persistence"
name = "photom"
name = "RampFit"
name = "refpix"
name = "resample"
name = "reset" 
name = "rscd" 
name = "saturation"
name = "skymatch"
name = "source_catalog"
name = "srctype" 
name = "stackrefs"
name = "straylight"
name = "subtract_images"
name = "superbias" 
name = "tweakreg"
name = "tweakreg_catalog"
name = "WfsCombine"
```

Agree that it would be good to make them consistent (and lowercase?) at some point.

PR #854, opened 2017-05-04
Prevents build failure if sphinx is not installed

jhunkeler:
A person trying to build the jwst source code without `sphinx` installed will see this:

```bash
$ python setup.py install
  File "setup.py", line 11, in <module>
    import sphinx
ImportError: No module named sphinx
```

This change prevents `sphinx` from becoming a hard dependency at compile-time. If the user runs `build_sphinx` and `sphinx` is not installed, the script dies "gracefully".

```bash
$ python setup.py build_sphinx
running build_sphinx
!
! Sphinx is not installed!
!
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,34 +1,50 @@
+from __future__ import print_function
+
 import os
 import subprocess
 import sys
-from setuptools import setup, find_packages, Extension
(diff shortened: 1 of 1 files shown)


sosey:
arg, good catch

PR #856, opened 2017-05-04
Modifying miri.py for new MRS structure

drlaw1558:
Revises miri.py to work with the newly-created MRS reference files; abstracts XAN/YAN into the reference file instead of requiring handling in miri.py.  Changes some function names for better clarity.  Updates imager bounding boxes.

### DIFF
--- a/jwst/assign_wcs/miri.py
+++ b/jwst/assign_wcs/miri.py
@@ -69,9 +69,9 @@ def imaging(input_model, reference_files):
     try:
         bb = distortion.bounding_box
     except NotImplementedError:
-        shape = input_model.data.shape
+        shape = 1032,1024
         # Note: Since bounding_box is attached to the model here it's in reverse order.
-        distortion.bounding_box = ((-0.5, shape[0] - 0.5), (3.5, shape[1] - 0.5))
+        distortion.bounding_box = ((-0.5, shape[1] - 0.5), (3.5, shape[0] - 4.5))
 
     # Create the pipeline
     pipeline = [(detector, distortion),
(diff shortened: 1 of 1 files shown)


jwst/assign_wcs/miri.py
```diff
@@ -69,9 +69,9 @@ def imaging(input_model, reference_files):
     try:
         bb = distortion.bounding_box
     except NotImplementedError:
-        shape = input_model.data.shape
+        shape = 1032,1024
```
nden: @drlaw1558 This change here made me think about how `bounding_box` works with subarrays. Hard coding shape to the full frame size means that with subarrays the bounding_box will be larger than the image. As an example, for a subarray with xstart=1000, xsize=30, the image `x` coordinates are from 0 to 30. Input of x= 100 is within the bounding_box, however in full frame coordinates `x_full=1032` and so is outside the full frame.

If we keep the bounding_box to the image shape, this problem is solved. However, in this case, for a subarray and for inputs outside the subarray size, the inputs will be considered outside the bounding_box when in fact the distortion is defined there because it's defined over the full frame. But may be this is not a problem and we want to keep the bounding_box to the image size?

drlaw1558: Hm, I see the complexity.  If we have a static shape=1032,1024 then the subarrays will not return errors even when a specified point is outside both the subarray and the full frame as per your example.  If we have a dynamic shape though, then the bounding box given also seems like it will be wrong for subarrays.  I.e., it will say that everything within 4 pixels of the left or right edge of the subarray doesn't map to the sky, which is untrue.  Is that right?  Some subarrays include reference pixels, and some do not.  It seems like we want the call to bounding_box to always be evaluated in full frame- does this make sense?

PR #858, opened 2017-05-04
Pass thru jsonschema error message after validation error

bernie-simon:
Datamodels validates changes against the schema. However, the change itself may not be the actual error if the model is already invalid. Passing thru the message generated by jsonschema ensures the
error message is not misleading.


nden:
I wonder if it's relatively easy to come up with a test for this.

PR #862, opened 2017-05-05
Fix bugs in Image3Pipeline

jdavies-st:
- Prevent flag_cr from writing out the updated dq array to the
  input FITS files
- Turn off blendheaders for outlier_detection step
- Turn off 2c names in Image3Pipeline


hbushouse:
Do these changes still in fact create separate level-2c files for each input level-2b file and update the DQ arrays (based on outlier_detection results) in only the new level-2c files? Unlike HST-land, we do not want the original inputs updated by outlier_detection. The updates need to be in a new (level-2c) product.

jdavies-st:
Hi Howard,

I temporarily turned off 2c generation for this commit in order to fix a bug where the flag_cr function was updating the input `_cal.fits` files in place.  I will turn the 2c stuff back on in a new commit before merging this PR.

Issue #861, opened 2017-05-05
Update all reference file schema to separate out the core


hbushouse:
The referencefile.schema.yaml file is being updated in #715 to remove the inheritance of core.schema.yaml. In its place, there will be a "core" set of keywords defined in the new version of referencefile.schema.yaml that are used by all reference files, which will be inherited by all of the individual reference file schemas. In addition, additional keywords needed only for certain reference files will be inherited, one-by-one, in the schema for those particular reference files.

In order to avoid having multiple definitions of keywords that are used in several reference files, such as FILTER, EXP_TYPE, SUBARRAY, etc., there will be a "mini-schema" for each of those individual keywords, which will then be inherited by any reference file schema that needs that keyword. That way the definition for a given keyword occurs in only one place. These mini-schema files for individual keywords will use a file name syntax of "keyword_<name>.schema.yaml", e.g. "keyword_filter.schema.yaml" for the FILTER keyword.

A schema for something like the dark reference file, which needs the keywords DETECTOR, READPATT, SUBARRAY, and USEAFTER for CRDS selectors, as well as keywords NFRAMES, NGROUPS, and GROUPGAP for use in the pipeline step, would look like this:

```
title: Dark current data model
allOf:
- $ref: referencefile.schema.yaml
- $ref: keyword_readpatt.schema.yaml
- $ref: keyword_nframes.schema.yaml
- $ref: keyword_ngroups.schema.yaml
- $ref: keyword_groupgap.schema.yaml
- $ref: subarray.schema.yaml
- type: object
  properties:
    data:
      title: Dark current array
      fits_hdu: SCI
      ...
```
In this example, the keywords INSTRUME, DETECTOR, and USEAFTER are defined in referencefile.schema.yaml, SUBARRAY is defined (along with other subarray characteristics) in subarray.schema.yaml, and the 4 remaining keywords that are needed are defined in their individual `keyword_xxxxx.schema.yaml` files, all of which are inherited by the dark data model.

hbushouse:
Initial assignments for ref file schemas that need updating. Will shuffle around once we see how many are assigned to each person, who doesn't have any, or who isn't available for the sprint.

- [ ] cubeflat.schema.yaml  Phil
- [ ] dark.schema.yaml  Nadia
- [ ] drizpars.schema.yaml  James
- [ ] fgs_photom.schema.yaml  Howard
- [ ] flat.schema.yaml  Phil
- [ ] fringe.schema.yaml  Jane
[...]

PR #865, opened 2017-05-08
add more keyword schemas [skip ci]

nden:
Issue #861

### DIFF
--- a/jwst/datamodels/schemas/keyword_nframes.schema.yaml
+++ b/jwst/datamodels/schemas/keyword_nframes.schema.yaml
@@ -1,0 +1,12 @@
+type: object
+properties:
+  meta:
+    type: object
+    properties:
+      exposure:
+        type: object
+        properties:
+          nframes:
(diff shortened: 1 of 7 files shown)


jwst/datamodels/schemas/keyword_nframes.schema.yaml
```diff
@@ -0,0 +1,12 @@
...
+properties:
+  meta:
+    type: object
+    properties:
+      exposure:
+        type: object
+        properties:
+          nframes:
+            title: Number of frames per group
+            type: number
```
nden: @hbushouse Should `nframes` be of type `integer` ?

hbushouse: The current datamodels infrastructure will allow "number" to be either integer or float, but given that this keyword value should always be a whole number, I would say we could explicitly declare it to be integer (that's how it's written in FITS headers).

PR #866, opened 2017-05-08
datamodels.schemas update for linearity and superbias 

sosey:
Updates for the linearity and suberbias reference file schema
No changes were necessary for the saturation schema

### DIFF
--- a/jwst/datamodels/schemas/linearity.schema.yaml
+++ b/jwst/datamodels/schemas/linearity.schema.yaml
@@ -1,6 +1,9 @@
 title: Linearity correction data model
 allOf:
 - $ref: referencefile.schema.yaml
+- $ref: subarray.schema.yaml
+- $ref: keyword_band.schema.yaml
+- $ref: keyword_filter.schema.yaml
 - type: object
   properties:
     coeffs:
--- a/jwst/datamodels/schemas/saturation.schema.yaml
+++ b/jwst/datamodels/schemas/saturation.schema.yaml
@@ -1,6 +1,7 @@
 title: Saturation checking data
 allOf:
 - $ref: referencefile.schema.yaml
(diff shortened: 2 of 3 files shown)


hbushouse:
Hmmm, at first it surprised me that the saturation ref file didn't need to have subarray-related keywords, but then a check of the saturation step code confirmed that it does one of the old-style "crude" checks to see if the science data and reference data match one another by simply comparing the shapes of their data arrays. And when the shapes don't match, it only uses the subarray corner and size info from the science data and effectively assumes that the ref image is full-frame, without explicitly verifying that it is. So technically the saturation reference file schema does not need to include any of the subarray keywords in order to conform to how things work now.

But the saturation ref files really should include those keywords, if for nothing else than at least documentation (via the FASTAXIS and SLOWAXIS keywords) that the data have been rotated to the DMS orientation on the sky, as well as to allow us to update the saturation step someday to make its on-the-fly subarray extraction more robust. So I guess the question is, do we want to make those updates later or do it now and get it over with? Personally, because it's such an easy change to add the "subarray.schema.yaml" entry to the new saturation.schema.yaml file, I'd say go ahead and do it now, so that it's there when we want to enhance the cal step later. That will also help to guarantee that ref file creators include those keywords in new ref file deliveries.

sosey:
This last commit adds subarrays to the saturation reference file. Note that these keywords are not used by CRDS for selection right now.

hbushouse:
👍

sosey:
Tested the updated schemas on a sample file from each type. The NIRISS superbias I tried issues warnings for the grating keywords which are strings in the reference file but numbers in the schema. Nadia says this is fine - merging now.

hbushouse:
In partial fulfillment of #861

PR #867, opened 2017-05-08
Update photom reference file schema [skip ci]

hbushouse:
Updated the various photom reference file schemas to use the new (non-core) approach. Also added a couple of new keyword_xxxxx schemas that I needed: one for the PHOTMJSR keyword and another for the 2 PIXAR_SR and PIXAR_A2 keywords. 

For the latter, I included the definitions for both keywords in a single schema and hence if we stick to the agreed upon convention for naming these files, it should be "pixelarea.schema.yaml". But because we already have a schema by that name, for the pixel area reference file, I named it "keyword_pixelarea.schema.yaml", even though it contains 2 keyword definitions. We could potentially get around this by renaming the existing pixelarea.schema.yaml (and associated pixearea.py module) to just "area.schema.yaml" (and "area.py").

### DIFF
--- a/jwst/datamodels/schemas/fgs_photom.schema.yaml
+++ b/jwst/datamodels/schemas/fgs_photom.schema.yaml
@@ -1,6 +1,7 @@
 title: FGS photometric flux conversion data model
 allOf:
 - $ref: photom.schema.yaml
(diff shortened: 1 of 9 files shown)


hbushouse:
These are updates in support of #861

PR #868, opened 2017-05-08
Update flat field reference file schemas [skip ci]

philhodge:
The subarray schema was added to all three.  In the nirspec_flat and
nirspec_quad_flat schemas, core.schema.yaml was replaced with
referencefile.schema.yaml.  See issue #861.

### DIFF
--- a/jwst/datamodels/schemas/flat.schema.yaml
+++ b/jwst/datamodels/schemas/flat.schema.yaml
@@ -1,5 +1,6 @@
 allOf:
 - $ref: referencefile.schema.yaml
+- $ref: subarray.schema.yaml
 - type: object
   properties:
     data:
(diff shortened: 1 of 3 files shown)


hbushouse:
I assume that cubeflat.schema.yaml does not need updating, because it's no longer in use (replaced by newer nirspec flat schemas)?

philhodge:
It was for the flat_field step, and that step no longer uses 
CubeFlatModel.  I don't see anything else that uses it either.  Should I 
delete it, or just leave it for the time being?

hbushouse:
Let's go ahead and delete it, to remove possibility of confusion. It can always be resurrected from github archives if it ever turns out to be useful.

PR #869, opened 2017-05-09
Update reffile schema for drizpars and outlierpars

jdavies-st:
The other instrument-specific drizpars and outlierpars reffiles do not need to be updated.  This is in support of #861.

### DIFF
--- a/jwst/datamodels/schemas/drizpars.schema.yaml
+++ b/jwst/datamodels/schemas/drizpars.schema.yaml
@@ -1,10 +1,10 @@
 title: Default Drizzle parameters data model
 allOf:
-- $ref: core.schema.yaml
+- $ref: referencefile.schema.yaml
 - type: object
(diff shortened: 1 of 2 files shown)


hbushouse:
I see that drizpars is already defined in CRDS and is selected only by INSTRUMENT, so just having the basic referencefile.schema.yaml inherited covers that. I assume the cal code doesn't need any other meta data from the drizpars files? Do we know yet what selection criteria will be used for outlierpars? If that's also just INSTRUMENT, then we should be fine.

jdavies-st:
No other metadata is needed for drizpars or outlierpars in the pipeline steps right now.  I suspect outlierpars would be instrument-specific, in that the MIRI detector might have different noise model parameters than the NIRCam detectors.  But we haven't figured these out yet.

hbushouse:
👍

PR #871, opened 2017-05-09
Fix overrides of distortion file units

bernie-simon:
The DistortionModel type in datamodels is implemented in distortion.py.  It allows the user to override the units of the distortion model through the optional parameters input_units and output_units. The checks which determine whether these optional parameters are set was incorrect and the result was both units were set to None. Fixing the checks resolved the problem.


bernie-simon:
Some extraneous changes got into this request, so I am pulling it.

PR #878, opened 2017-05-09
Update throughput schema per ticket #861.

dmggh:


jwst/datamodels/schemas/throughput.schema.yaml
```diff
@@ -1,6 +1,6 @@
 title: Filter throughput data model
 allOf:
-- $ref: core.schema.yaml
```
hbushouse: The core schema should be replaced by referencefile.schema.yaml

hbushouse: Also, there was a typo in the mods applied to throughput.py in #715. Line 16 needs to have "ThroughModel" updated to "ThroughputModel". Please include that update here too.

hbushouse:
It appears that throughput.schema.yaml is still missing a reference to "referencefile.schema.yaml". Is that going to be included in another PR?

dmggh:
Yes.

PR #879, opened 2017-05-09
Updated pixelarea schema per ticket #861

dmggh:


jwst/datamodels/schemas/pixelarea.schema.yaml
```diff
@@ -1,5 +1,13 @@
 allOf:
 - $ref: referencefile.schema.yaml
+- $ref: nirspec_area.schema.yaml
```
hbushouse: nirspec_area.schema.yaml is the schema for a completely separate flavor of pixel area reference file that is unique to NIRSpec spectroscopic modes and hence does not belong here amongst the pixel area ref file for imaging modes.

jwst/datamodels/schemas/pixelarea.schema.yaml
```diff
@@ -1,5 +1,13 @@
...
+- $ref: keyword_band.schema.yaml
+- $ref: keyword_exptype.schema.yaml
+- $ref: keyword_filter.schema.yaml
+- $ref: keyword_grating.schema.yaml
+- $ref: keyword_pixelarea.schema.yaml
+- $ref: keyword_pupil.schema.yaml
+
```
hbushouse: CRDS selectors for the AREA ref file consist of Detector and Useafter, which are included in referencefile.schema.yaml, as well as Filter, Pupil, and Exptype. Two keywords from the AREA ref file header are used in the photom step: PIXAR_SR and PIXAR_A2. Hence the list of necessary individual keyword schema files should be:

keyword_filter.schema.yaml
keyword_pupil.schema.yaml
keyword_exptype.schema.yaml
keyword_pixelarea.schema.yaml

Band and grating are not needed for the pixel area ref file.

jwst/datamodels/schemas/throughput.schema.yaml
```diff
@@ -1,6 +1,6 @@
 title: Filter throughput data model
 allOf:
-- $ref: core.schema.yaml
+- $ref: keyword_filter.schema.yaml
```
hbushouse: This was already included and commented on in #878

PR #881, opened 2017-05-10
Schema updates for issue #861

stscirij:
pathloss and mask schema updates

### DIFF
--- a/jwst/datamodels/schemas/mask.schema.yaml
+++ b/jwst/datamodels/schemas/mask.schema.yaml
@@ -1,6 +1,7 @@
 allOf:
 - $ref: referencefile.schema.yaml
 - $ref: subarray.schema.yaml
+- $ref: keyword_preadpatt.schema.yaml
 - type: object
   properties:
     dq:
--- a/jwst/datamodels/schemas/mask.schema.yaml
+++ b/jwst/datamodels/schemas/mask.schema.yaml
@@ -1,5 +1,6 @@
 allOf:
 - $ref: referencefile.schema.yaml
+- $ref: subarray.schema.yaml
 - type: object
   properties:
     dq:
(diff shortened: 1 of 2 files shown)


jwst/datamodels/schemas/mask.schema.yaml
```diff
@@ -1,5 +1,7 @@
 allOf:
 - $ref: referencefile.schema.yaml
+- $ref: subarray.schema.yaml
+- $ref: keyword_preadpatt.schema.yaml
```
hbushouse: MASK files use DETECTOR, SUBARRAY, and READPATT (in an or-like fashion) as CRDS selectors, so this looks good. I'm guessing the dq_init step code does not use any other keywords from the reference file.

jwst/datamodels/schemas/pathloss.schema.yaml
```diff
@@ -1,6 +1,7 @@
 title: Pathloss correction data model
 allOf:
 - $ref: referencefile.schema.yaml
+- $ref: keyword_pexptype.schema.yaml
```
hbushouse: PATHLOSS ref files use only EXP_TYPE, in an or-like fashion, as a CRDS selector, so this looks good. Are any other keywords from the ref file used in the pathloss step code?

stscirij: No

jwst/datamodels/schemas/mask.schema.yaml
hbushouse: 👍

jwst/datamodels/schemas/pathloss.schema.yaml
hbushouse: 👍

PR #888, opened 2017-05-11
Schemas and models for WCS reference files

### LINKED ISSUE #263: write schemas for WCS reference files
A schema for the base reference file will include all common fields, mostly keywords required by the redcat team. Each WCS reference type needs its own schema which will reference the base schema and add fields specific for the reftype. This needs to be done in collaboration with the INS teams as they will be the ones preparing reference files in the future.

nden:
Initial implementation of all WCS reference files schemas and models.
All but the Nirspec reference schemas were tested. Merging this so the MIRI team can start updating the reference files. Nirspec changes, if needed, will go in a separate PR.


nden:
Although not visible because of the squashed commits, tests have passed.

PR #891, opened 2017-05-12
fix indexing errors

nden:
We should be catching this kind of errors with Travis. I added a test option with numpy 1.12.


nden:
Another error due to numpy 1.12 surfaced. It's in datamodels. @bernie-simon Could you look at it?

nden:
Merging this as the failure is not related to these changes.

PR #894, opened 2017-05-15
Add exptype keyword to pathloss ref file schema

hbushouse:
The pathloss step code accesses the value of the EXP_TYPE keyword in the pathloss reference file returned from CRDS, so that keyword needs to be defined in the pathloss ref file schema.


hbushouse:
Travis failure is unrelated. Merging.

PR #895, opened 2017-05-15
Add new PsfMask data model for coronagraphic level-3 processing

hbushouse:
The align_refs step in the calwebb_coron3 pipeline uses a PSFMASK reference file. This PR includes a new PsfMask data model module and an associated schema. The reference file needs to use the CORONMSK keyword, so a new keyword_coronmsk schema has also been created. The PSFMASK reference files contain a SCI extension, containing a 2-D floating-point array for the mask data.


hbushouse:
Unrelated numpy error in Travis. Merging.

PR #896, opened 2017-05-16
Fixed a bug where the source was not reconstructed correctly when var…

jemorrison:
…ying size of cube


nden:
The Travis failure is a known and unrelated problem in datamodels. I think this can be merged.

Issue #687, opened 2017-02-08
Task: Add resample step calwebb_image2 pipeline


hbushouse:
A resampled image (_i2d) product should be produced at the end of calwebb_image2 pipeline processing, in addition to the existing calibrated, but unresampled (_cal), product, as is done in the calwebb_spec2 pipeline for spectroscopic exposures. The resample step should be called at the very end of the calwebb_image2 pipeline, using the single calibrated exposure as input.

Note that not all imaging-mode exposures should have a resampled product produced. In particular, images from Wave-Front Sensing and Control (WFS&C) visits should NOT be resampled (there's no need or use for a resampled product). These exposures can be identified by the value of WFSVISIT keyword. The resample step should be skipped for these exposures.

This is a low-priority task for B7.1 and will only be implemented as time allows.

hbushouse:
In addition to not producing a resampled product for WFS&C images, there are other types of exposures that should also NOT produce a resampled product, including:

1) Any time the input to calimage2 is a _rateints product, which are 3D cubes of countrate images and are used for time-series imaging and coronagraphy.

2) NIRISS AMI exposures, identified by EXP_TYPE='NIS_AMI'.

hbushouse:
Fixed by #1419.

PR #897, opened 2017-05-17
Make consistent pipeline/step I/O prototype

stscieisenhamer:
Possible solution to the various consistency and saving issues.  The stpipe infrastructure has been updated and calwebb_spec2 modified. Other pipelines have not been "repaired" pending review.


jwst/pipeline/calwebb_image2.py
```diff
@@ -1,21 +1,24 @@
...
+from ..stpipe import Pipeline
 
 # calwebb IMAGE2 step imports
 from ..assign_wcs import assign_wcs_step
 from ..flatfield import flat_field_step
 from ..photom import photom_step
 
 
-__version__ = "2.2"
+__version__ = "2.3"
```
hbushouse: I think this is a sufficiently major revision (supporting ASN) to deserve a full version increment, i.e. 3.0

jwst/pipeline/calwebb_image2.py
```diff
@@ -37,48 +40,68 @@ def process(self, input):
...
+        input.meta.asn.table_name = asn_file
 
-                # Remove the step name that stpipe appended to the file name,
-                # as well as the original suffix on the input file name,
-                # and create a new name with the appropriate output suffix
-                root = root[:root.rfind('_')]
-                self.output_file = root[:root.rfind('_') + 1] + suffix + ext
+        # work on slope images
+        input = self.assign_wcs(input)
+        input = self.flat_field(input)
```
hbushouse: Should go ahead and take the opportunity to add a call to background_subtract (after assign_wcs) if there are background image(s) defined in the ASN, just as is done in calwebb_spec2. This something the JCCWG has requested for future expansion. Probably won't encounter any instances using background any time soon, but at least the code is ready to support it.

jwst/pipeline/calwebb_image2.py
```diff
@@ -37,48 +40,68 @@ def process(self, input):
...
+        input = self.photom(input)
```
hbushouse: Should go ahead and add call to resample step to produce individual i2d product, as requested in #687. Would need to be inserted after saving the current "input" model to cal product and then add a call to save the resample output to a "i2d" product.

jwst/stpipe/pipeline.py
```diff
@@ -33,19 +33,30 @@
...
     """
     A Pipeline is a way of combining a number of steps together.
     """
+
+    # Configuration
+    spec = """
+    output_basename = string(default=None) # Output base name
+    output_ext = string(default=".fits") # Output extension
+    suffix = string(default=None) # Suffix for output file name
+    output_use_model = boolean(default=False)
```
jdavies-st: What does this `output_use_model` config option do?  Would be good to have a comment documenting it.

jwst/stpipe/pipeline.py
```diff
@@ -33,19 +33,30 @@
...
 class Pipeline(Step):
     """
     A Pipeline is a way of combining a number of steps together.
     """
+
+    # Configuration
+    spec = """
+    output_basename = string(default=None) # Output base name
+    output_ext = string(default=".fits") # Output extension
+    suffix = string(default=None) # Suffix for output file name
```
jdavies-st: I like the addition of `suffix`!

jwst/stpipe/step.py
```diff
@@ -661,8 +698,107 @@ def save_model(self, model, name, *args, **kwargs):
...
+    def make_output_path(step, data, basepath=None, suffix=None):
+        """Make up a path based on data and user specification
+
+        Parameters
+        ----------
+        step: Step
+            The step which produced the data
+
+        data: obj
+            Unused by this routine
```
jdavies-st: Might want to updated this docstring.  Looks like `data` is used for `meta.filename`.

jwst/pipeline/calwebb_image2.py
stscieisenhamer: Documented in #901 Would rather this and the comment below be in separate PR's: This one has the potential of causing much headache and would rather not clutter up with actual science changes.

jwst/pipeline/calwebb_image2.cfg
```diff
@@ -1,5 +1,7 @@
 name = "Image2Pipeline"
 class = "jwst.pipeline.Image2Pipeline"
+output_use_model = True
+suffix = "cal"
```
hbushouse: So how does this work when the input is a rateints product and hence the output needs to be calints (not cal)? Also, now that we're adding the resample step to the end of calwebb_image2, the last product generated is an i2d. Where/how does that get specified?

stscieisenhamer: Basically, this will work as in `calwebb_spec2` then. That would be the example for anything more complicated that just a step.

stscieisenhamer:
OK. Everything with this should be a go. I'll plan an doing the merge Sunday evening, so that Monday afternoon can be spent fixing it all once the regression tests have run

hbushouse:
Still planning to merge today yet, so that we can do BDA tomorrow afternoon?

stscieisenhamer:
Done, hopefully in time for regression...

PR #899, opened 2017-05-17
calwebb_coron3: create fake output products

hbushouse:
Added temporary stubs to the calwebb_coron3 pipeline module to create fake outlier_detection products and resample products, until those steps are working for coronagraphic data. This will allow the pipeline to at least create all output product types as examples. Set the outlier_detection and resample steps to skip in the default calwebb_coron3.cfg.

Also updated the stack_psfs and align_psfs steps to set the step completion status in the metadata of their output models.


hbushouse:
Travis failure is from unrelated numpy issue. Merging.

PR #902, opened 2017-05-19
Update all doc index files

hbushouse:
Minor cosmetic updates to lots of doc index files and a few corrections here and there.


hbushouse:
Travis failures appear to be an issue with the OSX servers, unrelated to this PR (all Linux tests pass). Merging.

PR #903, opened 2017-05-19
Save arrays

nden:
I am proposing removing some of previous changes to `properties._cast`.
The problem I am trying to solve is to serialize numpy arrays directly to asdf files. Note that currently all numpy array serialization in datamodels is to FITS files which calls specialized code.
Serializing ndarrays works with asdf but not with datamodels.

Looking at the code in `properties._cast` removed in this PR, I think it was added to solve an issue with saving Time objects to FITS files using the core schema. I believe there was a bug/typo in the core schema which caused the Time object problems. So here I've also commented out and changed the `date` field in the core schema to use `anyOf` instead of `allOf`.

This is all motivate by work on #888 which is currently stalled until the problem with numpy arrays is solved.

### DIFF
--- a/jwst/datamodels/properties.py
+++ b/jwst/datamodels/properties.py
@@ -25,7 +25,6 @@ log.addHandler(logging.NullHandler())
 
 __all__ = ['ObjectNode', 'ListNode']
 
-
 def _cast(val, schema):
     val = _unmake_node(val)
     if val is not None:
@@ -43,11 +42,6 @@ def _cast(val, schema):
             raise ValueError(
                 "Array has wrong number of dimensions.  Expected <= {0}, got {1}".format(
                     schema['max_ndim'], len(val.shape)))
-        #tag = schema.get('tag')
-        #if tag is not None:
-            #val = tagged.tag_object(tag, val)
-        #if isinstance(val, np.generic) and np.isscalar(val):
-            #val = np.asscalar(val)
 
     return val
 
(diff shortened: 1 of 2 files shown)


nden:
Failure is numpy 1.12 related.

bernie-simon:
The reason for tagging in general is that yaml (and asdf) is a serialization of a data structure and the code needs to know the type of objects embedded in the structure so that it can recreate them when reading back from disk. The tagging is also used to validate an object against the schema. The tag attribute acts as a check of the object type.

bernie-simon:
The tagging verifies that when you set a datamodels attribute to a complex object, like a date, that the type of the complex object agrees with the type that the schema expects. If there's no tag field in the schema, the check won't be performed, but if there is, the code needs to run through the tagging mechanism in asdf.

nden:
@bernie-simon Agreed. However, complex objects already have tags in their schemas. For example

```
date:
  $ref: http://stsci.edu/schemas/asdf/time/time-1.0.0
```

contains the tag for Time objects.
I think only primitive python types, like lists, etc..., which normally don't have tags, need to be tagged. The custom types, like `Time` and `ndarray` are already tagged by definition. At least this is my understanding. In any case the code removed here breaks functionality. 
Can you provide an example which fails without this code?

nden:
I am going to merge this and let the nightly tests run. It can always be reverted if needed.

PR #904, opened 2017-05-19
Set wcs to None when naxis is zero in header

bernie-simon:
ipython shows a warning message when tab is used to auto-complete the list
of attributes of a datamodel. The warning message comes from the wcs code and
says:

The WCS transformation has more axes (2) than the image it is
associated with (0)

The problem is that the code is trying to create a wcs when naxis in
the primary heade is zero. I have modified get_fits_wcs so that it
sets the wcs to None when naxis is zero. This eliminates the warning
message.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -776,7 +776,12 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
         hdu = fits_support.get_hdu(ff._hdulist, hdu_name)
         header = hdu.header
 
-        return WCS(header, key=key, relax=True, fix=True)
+        naxis = header.get('NAXIS', 0)
+        if naxis == 0:
+            wcs = None
+        else:
+            wcs = WCS(header, key=key, relax=True, fix=True)
+        return wcs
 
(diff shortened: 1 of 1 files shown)


nden:
Perhaps we should hold off on this - for two reasons. The warning message is actually correct, not harmful, and the WCS object is created correctly. It just warns users that there's no data associated with this header. We know this is currently the case because the WCS is in the primary header. If the goal is to simply get rid of the warning, then it should be caught.

The bigger issue is what should actually be done here. For some models the GWCS object is the only correct one and the FITS WCS should not be created. For others the FITS WCS is good.
It's still unclear to me what is the specific reason for inheriting from NDModel at this moment. Until all these questions are answered I think we should not put additional code.

jdavies-st:
Ok, I agree with @nden.  I will just have to live with my tab completion giving warnings for the time being until the larger issue (where to put WCS objects, and of what type) is dealt with for JWST.  Thanks Bernie & Nadia.

Further, once we get the wcs keywords out of the [0] fits header and into the [sci] header, then the error message will go away, if the fits wcs object is created.  So the problem will solve itself.

bernie-simon:
Closing this pull request pending a better and more permanent solution to the missing wcs problem.

Issue #907, opened 2017-05-22
Dark step needs to allow for nframes>1 in dark reference file


hbushouse:
Julia Duval (INS) reports:

I am processing a batch of NIRISS WFSS simulated data through the pipeline, and am finding a weird behavior with the dark current subtraction. The science data has 22 groups (1 integration, 4 frames per group). The reference file in CRDS has 25 groups. Yet, I get the following warning by the pipeline that the reference file has less groups than the science and therefore the subtraction is not done. Can you explain this?

2017-05-22 14:23:51,784 - stpipe.dark_current - INFO - DarkCurrentStep instance created.
2017-05-22 14:24:09,432 - stpipe.dark_current - INFO - Prefetch for DARK reference file is '/grp/crds/cache/references/jwst/jwst_niriss_dark_0044.fits'.
2017-05-22 14:24:09,478 - stpipe.dark_current - INFO - Step dark_current running with args ('wfss_long_ramps/niriss_uncal_F115W_gr150c_position4_nis_out_dq_sat_superbias_refpix_oec_linearity.fits',).
2017-05-22 14:24:10,264 - stpipe.dark_current - INFO - Using DARK reference file /grp/crds/cache/references/jwst/jwst_niriss_dark_0044.fits
2017-05-22 14:24:10,443 - stpipe.dark_current - INFO - Science data nints=1, ngroups=22, nframes=4, groupgap=0
2017-05-22 14:24:10,443 - stpipe.dark_current - INFO - Dark data nints=1, ngroups=25, nframes=4, groupgap=0
2017-05-22 14:24:10,443 - stpipe.dark_current - WARNING - There are more groups in the science data than in the dark data.
2017-05-22 14:24:10,443 - stpipe.dark_current - WARNING - Input will be returned without subtracting dark current.

This is due to the fact that when the dark_current step checks to see if it has enough dark data to apply to the science data it is assuming that nframes=1 in the dark data, without actually looking at the value of nframes in the dark data (it reports it in a log message, but then ignores it). So in this case the step thinks it needs 22*4=88 groups of dark data in order to do on-the-fly frame-averaging of the dark data to match the science data, but it only has 25 groups of dark data.

The check in dark_sub.py needs to be enhanced to recognize when nframes and groupgap are the same in the science and dark data, in which case the dark data can be applied as-is, with no need for on-the-fly frame averaging.

hbushouse:
Fixed by #910

PR #910, opened 2017-05-23
Fix dark_current nframes problem #907

hbushouse:
This update to the dark_current step fixes the problem identified in #907, by computing the full amount of data available in the dark reference file via its values of ngroups, nframes, and groupgap, with no assumption that nframes=1 and groupgap=0 for dark reference files.


hbushouse:
Travis failures expected and unrelated. Merging.

PR #918, opened 2017-05-24
Outlier refactor

jdavies-st:
1. Many changes to outlier_detection_step and outlier_detection:

- add `save_intermediate_files` to user parameters for the step
- replace `stsci.ImageStats` with `astropy.stats.sigma_clipped_stats`
- replace `stsci.image.numcombine` with `stsci.image.median`
- remove `nsigma, hthresh, lthresh, ctegrow` step parameters
- units of readnoise in `flag_cr` is now DN, not electrons

2. Use built-in `Step` logging in `Image3Pipeline`

3. Fix bug in GWCSBlot where pixmap was mapping in the wrong direction.
    The function `resample_utils.calc_gwcs_pixmap` was being passed the
    input and output `gwcs.WCS` objects in the wrong order, producing
    the wrong mapping that `drizzle.cdrizzle.tblot` was expecting.

4. Fix bug where resample pixmap was different shape from  input sci


jdavies-st:
CI fail is unrelated to this PR.  So merging.

Issue #928, opened 2017-05-25
Add CRDS URI to cache path translation


jaytmiller:
Robert requested an interface to translate the CRDS URI's recorded in dataset headers:

crds://jwst_miri_flat_0177.fits

into absolute file paths in the current/local CRDS cache:

/grp/crds/cache/references/jwst/jwst_miri_flat_0177.fits

jaytmiller:
I added a method reference_uri_to_cache_path() to jwst.step.Step.
I added the related function reference_uri_to_cache_path() to jwst.crds_client.

PR #929, opened 2017-05-25
Crds reference uri translation

jaytmiller:
Pull for issue #928.   Adds translation from crds:// to CRDS cache paths for reference files.


jaytmiller:
Investigating the Travis failure,  it looks unrelated to CRDS changes and isolated to a particular test run.   These CRDS changes are standalone relatively shallow function and method additions,  they're not being called yet.

PR #932, opened 2017-05-26
Updates to calspec2 steps logging

hbushouse:
Updated the log messages in several calspec2 steps, mainly for the case of processing a NIRSpec MSA exposure. When there are many slits to process, it's nice for the user to see some progress messages, rather than just waiting 10-20 minutes to see that the step has completed. Also changed some info messages to debug, because they really aren't needed on a regular basis.


hbushouse:
Only Travis failure is unrelated numpy issue. Merging.

Issue #892, opened 2017-05-12
test failure with numpy 1.12


nden:
The following test in datamodels fails with numpy 1.12. @bernie-simon 

```
================================== FAILURES ===================================

______________________________ test_replace_table ______________________________

    def test_replace_table():
        from astropy.io import fits
        schema_narrow = {
            "allOf": [
                mschema.load_schema(
                    os.path.join(os.path.dirname(__file__),
                                 "../schemas/core.schema.yaml"),
                    resolve_references=True),
                {
                    "type": "object",
                    "properties": {
                        "data": {
                            "title": "relative sensitivity table",
                            "fits_hdu": "RELSENS",
                            "datatype": [
                                {"name": "TYPE", "datatype": ["ascii", 16]},
                                {"name": "T_OFFSET", "datatype": "float32"},
                                {"name": "DECAY_PEAK", "datatype": "float32"},
                                {"name": "DECAY_FREQ", "datatype": "float32"},
                                {"name": "TAU", "datatype": "float32"}
                            ]
                        }
                    }
                }
            ]
        }
        schema_wide = {
            "allOf": [
                mschema.load_schema(
                    os.path.join(os.path.dirname(__file__),
                                 "../schemas/core.schema.yaml"),
                    resolve_references=True),
                {
                    "type": "object",
                    "properties": {
                        "data": {
                            "title": "relative sensitivity table",
                            "fits_hdu": "RELSENS",
                            "datatype": [
                                {"name": "TYPE", "datatype": ["ascii", 16]},
                                {"name": "T_OFFSET", "datatype": "float64"},
                                {"name": "DECAY_PEAK", "datatype": "float64"},
                                {"name": "DECAY_FREQ", "datatype": "float64"},
                                {"name": "TAU", "datatype": "float64"}
                            ]
                        }
                    }
                }
            ]
        }
        x = np.array([("string", 1., 2., 3., 4.)],
                     dtype=[(str('TYPE'), str('S16')),
                            (str('T_OFFSET'), np.float32),
                            (str('DECAY_PEAK'), np.float32),
                            (str('DECAY_FREQ'), np.float32),
                            (str('TAU'), np.float32)])
        m = DataModel(schema=schema_narrow)
        m.data = x
        m.to_fits(TMP_FITS, overwrite=True)
        with fits.open(TMP_FITS) as hdulist:
>           assert repr(list(x)) == repr(list(np.asarray(hdulist[1].data)))
[...]

PR #933, opened 2017-05-26
Fix fits test that fails, but shouldn't

bernie-simon:
A test in jwst.datamodels was reporting a failure. The values reported were ('string', 1., 2. 3., 4.) and ('string', 1.0, 2.0, 3.0, 4.0). The test should be a success, but was a failure because of how the two records were compared. I wrote a  small function that compares the two records element by element and now the test passes.


nden:
Closes #892

Issue #601, opened 2017-01-13
Task: NIRSpec wavecorr step


hbushouse:
This is a continuation of Trac 373 https://aeon.stsci.edu/ssb/trac/jwst/ticket/373. A wavelength zeropoint correction needs to be implemented in calwebb_spec2, either as part of the initial assign_wcs process or as a separate step, that will correct wavelength assignments for sources not centered in a fixed slit or MSA slitlet.

Documentation is available at ​https://confluence.stsci.edu/download/attachments/61442267/ESA-JWST%E2%80%93SCI-NRS-TN-2016-003_wavelength_zero_point_1.pdf?version=1&modificationDate=1462212841000&api=v2

nden:
As I find out the details I'll document them here. 
The source x,y position in the shutter is given in the msa configuration file, columns `estimated_source_in_shutter_x` and `estimated_source_in_shutter_y`. These are read in by assign_wcs and saved in each `Slit` as `Slit.source_xpos` and `Slit.source_ypos`. The source position is in a coordinate system associated with each shutter whose origin is the upper left corner of the shutter. It looks like (unconfirmed yet) that positive x is to the right and positive y is downwards. To convert to the coordinate frame associated with the slit, where 0,0 is in the center of the slit, need to subtract -.05 in both directions.

hbushouse:
Don't subtract -.05, but rather subtract 0.5

hbushouse:
For fixed-slit exposures we don't have estimated source position in the slit being given to us from upstream, as is the case with MOS observations that have this information in the MSA metadata file. For fixed-slit exposures we will rely on the x/yoffset information for each exposure that is passed from PPS. It gives the relative offset of each exposure in a dither pattern relative to the center of the pattern, which for the fixed-slits is centered on the particular slit.

The exposure offsets will be recorded in the new header keywords XOFFSET and YOFFSET (see https://github.com/STScI-JWST/jwst/pull/1112). The keywords give offsets from the center of a slit in units of arcsec. Those values need to be transformed into fractional slit units for use in the wavelength zero-point correction, e.g. if the offset is -1.0 arcsec and the slit size is 1.5 arcsec, the fractional slit offset is -0.67.

nden:
@hbushouse In the example above, since (0,0) is the center of the slit, isn't the fractional offset -0.33?

hbushouse:
[...]

PR #935, opened 2017-05-28
change coord system for source position

nden:
#601 describes the coordinate frame used for the source position in the slit for Nirspec MSA and FS observations. The origin of the frame is the upper left corner of the pixel with positive x to the right and positive y downwards. The position needs to be converted to the frame the Nirspec instrument model uses where the origin is in the center of the slit , +y is up and +x is right. This PR makes the conversion of the cooridnates when the Slit values are read in from the msa configuration file. This means that the wavelength zero-point correction and the path loss correction may use the values directly.

@hbushouse @stscirij 

Another related question (which was briefly discussed) is if this needs to be further transformed because of the DMS vs native coordinates. I am pretty sure the source position values are in native coordinates. The wavelength zero-point table is in native coordinates as well. So I think that at least for the WZPC this does not matter.

### DIFF
--- a/jwst/assign_wcs/nirspec.py
+++ b/jwst/assign_wcs/nirspec.py
@@ -445,6 +445,17 @@ def get_open_msa_slits(msa_file, msa_metadata_id):
                 for s in msa_source if s['source_id'] == source_id][0]
             # Create the output list of tuples that contain the required
             # data for further computations
+            """
+            Convert source positions from PPS to Model coordinate frame.
+            The source x,y position in the shutter is given in the msa configuration file,
+            columns "estimated_source_in_shutter_x" and "estimated_source_in_shutter_y".
+            The source position is in a coordinate system associated with each shutter whose
+            origin is the upper left corner of the shutter, positive x is to the right
+            and positive y is downwards. To convert to the coordinate frame associated with the
+            slit, where (0, 0) is in the center of the slit, we subtract 0.5 in both directions.
+            """
+            source_xpos = source_xpos - 0.5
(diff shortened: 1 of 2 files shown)


nden:
@hbushouse @stscirij Do you agree with this change?

When reading the msa configuration file we save the source position source_x, source_y in the Model coordinate frame instead of the raw values in the msa configuration file. In this way the zero-point and the pathloss corrections don't have to do this.

hbushouse:
Looks good. Only comment would be that you might want to add a little more commentary in the actual code where the conversion is done to indicate the rationale behind the subtraction of 0.5 (i.e. the info above about the origins of the 2 coord frames). Just for future reference.

jwst/assign_wcs/nirspec.py
```diff
@@ -445,6 +445,17 @@ def get_open_msa_slits(msa_file, msa_metadata_id):
...
             # Create the output list of tuples that contain the required
             # data for further computations
+            """
+            Convert source positions from PPS to Model coordinate frame.
+            The source x,y position in the shutter is given in the msa configuration file,
+            columns "estimated_source_in_shutter_x" and "estimated_source_in_shutter_y".
+            The source position is in a coordinate system associated with each shutter whose
+            origin is the upper left corner of the shutter, positive x is to the right
+            and positive y is downwards. To convert to the coordinate frame associated with the
+            slit, where (0, 0) is in the center of the slit, we subtract 0.5 in both directions.
```
hbushouse: 👍

PR #937, opened 2017-05-30
Add filter and pupil keywords to FlatModel schema

hbushouse:
Bryan Hilbert noticed that the FILTER and PUPIL keywords were not be written to flat-field reference files he was creating via the updated FlatModel data model. Added definitions for those keywords to the FlatModel schema, because they are used as flat selectors for MIRI, NIRCam, and NIRISS.


nden:
Superseded by #936

PR #945, opened 2017-06-01
fix reftype

nden:
write validation code


nden:
One failing test is a CRDs connection issue.

PR #958, opened 2017-06-07
Make ModelContainer.save consistent with DataModel.save

jdavies-st:
The `save` method gets passed a filename.  This gets ignored for ModelContainer, and `meta.filename` is used to write out the file. This change is needed as `stpipe.Step.save_model` method needs the same call signature for all DataModels that are returned from a step.

Also, add a filter to suppress "Card is too long" warnings from `astropoy.io.fits` when writing out a DataModel.  These warning messages are caused by our schemas containing descriptions that are too long for the FITS standard in some cases, and in others by keyword value strings that are very long.  All `DataModel.save` operations produce this warning, so let's suppress it.

### DIFF
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -186,12 +186,18 @@ class ModelContainer(model_base.DataModel):
         self.meta.asn_rule = str(asn_data['asn_rule'])
 
 
-    def save(self, path=None, *args, **kwargs):
+    def save(self, path_not_used, path=None, *args, **kwargs):
         """
         Write out models in container to FITS or ASDF.
 
         Parameters
         ----------
+        path_not_used : string
+            This first argument is ignored in this implementation of the
+            save method.  It is used by pipeline steps to save individual
+            files, but that is not applicable here.  Instead, we use the path
+            arg below and read the filename output from `meta.filename` in each
+            file in the container.
 
         path : string
             Directory to write out files.  Defaults to current working dir.
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -9,6 +9,7 @@ import datetime
 import inspect
 import os
 import sys
+import warnings
 
 import numpy as np
 import jsonschema
@@ -362,7 +363,9 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
 
         with fits_support.to_fits(self._instance, self._schema,
                                   extensions=self._extensions) as ff:
-            ff.write_to(init, *args, **kwargs)
+            with warnings.catch_warnings():
+                warnings.filterwarnings('ignore', message='Card is too long')
+                ff.write_to(init, *args, **kwargs)
 
     @property
     def shape(self):


nden:
@jdavies-st Do we know what keywords are causing the warning message? I think we should fix these while it's not too late, if it's possible, instead of catching the warning.

hbushouse:
There could be some keyword definitions where the comment string is too long and hence causes a warning every time, and hence could/should be shortened in the schema. But others occur when the keyword string value happens to be very long, such as the path to a reference file that was used, which then causes the comment field to fall off the end of the record. Those can't be predicted in advance or fixed in the schema.

jdavies-st:
As Howard suggests, there are many keywords causing this warning, and even with short descriptions, the (keyword, value, description) tuple can have a very long string value, which causes the description to get truncated to fit in the 80 character limit required by FITS.

On the flip side, the keyword descriptions are generally taken from the JWST Keyword Dictionary, though some of them have been already shortened significantly.  The good thing is that these are stored in the schema, so get written fresh each time we create an asdf-in-fits file.

One thing I'm not clear on, is if the schema is stored in the ASDF extension.  Certainly when these files are read in via `datamodels`, the resulting object in memory has an `_schema` attribute which contains the full schema, including the descriptions.  Is this schema stored in the ASDF extension of the fits file when it is written?

nden:
My concern was we may miss a real issue, e.g. a keyword being more than 8 characters, as not all keywords are in the keyword dictionary (it has happened before). But it looks like the reference file path is a real use case. Is the entire path saved if it's longer than 80 characters or is it truncated?

hbushouse:
This warning can be triggered even if the keyword value string is only 20-40 chars, because in that case it shoves the comment field 20-40 chars "to the right", which then causes the entire keyword+value+comment string to be longer than 80 chars, which is when we get the warning (at least I believe the warning is based on the entire keyword+value+comment being > 80).

philhodge:
It appears that astropy.io.fits handles a long keyword name and a long 
keyword value (string) reasonably:

 >>> hdr['longkwtest'] = 37.19
WARNING: VerifyWarning: Keyword name 'longkwtest' is greater than 8 
characters or contains characters not allowed by the FITS standard; a 
HIERARCH card will be created. [astropy.io.fits.card]

 >>> hdr['longpath'] = 
'dsfjkeinciieiii28dfkwekddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd'

Note that there was no warning for the long keyword value, and:

HIERARCH longkwtest = 37.19
LONGPATH= 
'dsfjkeinciieiii28dfkwekdddddddddddddddddddddddddddddddddddddddddddd&'
CONTINUE 
'ddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd&'
CONTINUE  'dddddddddddddddddddddddddd'

jdavies-st:
Yes, I think the most common throwing of the "Card is too long" warning is that the keyword value is a long string or float, but not long enough to cause a CONTINUE line to be made.

hbushouse:
Huh. So maybe the warning is coming from just the comment field being too long? In which case we can easily fix once and for all by modifying the schema.

philhodge:
@hbushouse , I thought that's what you've been saying all along, that the comment can get pushed to the right by a long value (this would only happen for a string value, because floats are right-justified at column 30), and if the comment is pushed far enough it will be truncated.  We can only fix that in the schema by eliminating the comment, if the string value can be arbitrarily long.

Issue #650, opened 2017-01-26
Interpolation Range Error


stscieisenhamer:
From [Discrepancy Report #003862](https://www.ess.stsci.edu/prsystem/servlet/prbrowse/dr.3862)
```
While processing we noticed a interpolation range error in the Level 2b step
for the files and error message is listed below.

Files:
jw95115001001_02105_00001_nrs1_rate.fits
jw95115001001_02104_00001_nrs1_rate.fits
jw95115001001_02106_00001_nrs1_rate.fits
jw95115001001_02102_00001_nrs1_rate.fits

2017-01-24 00:41:54,923 - stpipe.Spec2Pipeline.resample_spec - INFO -
Converting MultiSlit to ModelContainer Traceback (most recent call last):
File
"/ifs/int/jwstd/pipeline/build7/pkgs/miniconda3/envs/jwstb7rc7/lib/python2.7/sit
e-packages/jwst-0.7.0rc7-py2.7-linux-x86_64.egg/EGG-INFO/scripts/strun", line
14, in step = Step.from_cmdline(sys.argv[1:]) File
"/ifs/int/jwstd/pipeline/build7/pkgs/miniconda3/envs/jwstb7rc7/lib/python2.7/sit
e-packages/jwst-0.7.0rc7-py2.7-linux-x86_64.egg/jwst/stpipe/step.py", line 128,
in from_cmdline return cmdline.step_from_cmdline(args) File
"/ifs/int/jwstd/pipeline/build7/pkgs/miniconda3/envs/jwstb7rc7/lib/python2.7/sit
e-packages/jwst-0.7.0rc7-py2.7-linux-x86_64.egg/jwst/stpipe/cmdline.py", line
275, in step_from_cmdline step.run(*positional) File
"/ifs/int/jwstd/pipeline/build7/pkgs/miniconda3/envs/jwstb7rc7/lib/python2.7/sit
e-packages/jwst-0.7.0rc7-py2.7-linux-x86_64.egg/jwst/stpipe/step.py", line 343,
in run result = self.process(*args) File
"/ifs/int/jwstd/pipeline/build7/pkgs/miniconda3/envs/jwstb7rc7/lib/python2.7/sit
e-packages/jwst-0.7.0rc7-py2.7-linux-x86_64.egg/jwst/pipeline/calwebb_spec2.py",
 line 166, in process resamp = self.resample_spec(input) File
[...]

PR #962, opened 2017-06-08
Fix for #650, interpolation in building output frame in Spec2Pipeline [skip CI]

jdavies-st:
This fixes the failure at `interpolate`, but it doesn't solve the larger issue of pixels being thrown to the wind and not landing in the output frame when we resample.


jdavies-st:
Resolves #650

PR #966, opened 2017-06-09
Update links and doc strings in various docs

hbushouse:
Changed hard url links in the introductory user's guide to use cross-references instead. Also updated the doc strings in a few modules to fix formatting issues. 

Also reordered the list of model classes in datamodels/__init__.py so that the automatic api builder puts them in a logical (basically alphabetical) order in the resulting docs. It would be great if everyone would continue to pay attention to and think about this ordering when making future updates/additions to the list of data model classes.


hbushouse:
Travis failures are unrelated. Merging.

PR #970, opened 2017-06-12
base framework for calwebb_spec3

stscieisenhamer:


jwst/pipeline/calwebb_spec3.cfg
```diff
@@ -0,0 +1,18 @@
+name = "Spec3Pipeline"
+class = "jwst.pipeline.Spec3Pipeline"
+suffix = 'x1d'
+save_results = True
+
+    [steps]
+      [[skymatch]]
+        suffix = 'skymatch'
+      [[outlier_detection]]
+        suffix = 'crj'
```
hbushouse: Looks like the AA is recommending the suffix "cr**f**" (cosmic ray **flagged**) for this type of product.

jdavies-st: Horray!  Does that mean we can change the output product names for the image3 pipeline too from the currently-very-awkward `_cal-a3001.fits` to `_crf.fits`?

hbushouse: Indeed.

hbushouse: Oh wait - I take that back. We can only remove the asn_id suffix from file names that are source-based, because the source-based convention already includes asn_id in it. For exposure-based level-2c files, which is the normal output of outlier_detection in calimage3, we do need to keep the asn_id suffix. But there is talk about moving it to **before** the _crf suffix.

jdavies-st: Ok, that would make sense.  `_a3001_crf.fits` is much nicer than what we have currently.

Issue #976, opened 2017-06-12
source_catalog ValueError comparing micro_Jy


jdavies-st:
@larrybradley 

Seeing a problem on [line 155](https://github.com/STScI-JWST/jwst/blob/master/jwst/source_catalog/source_catalog.py#L155) in `make_source_catalog`.  In the previous block, if the `else` clause gets executed, then the comparison on line 155 becomes a comparison between an array and 0.0.  The traceback:
```
Traceback (most recent call last):
  File "/Users/jdavies/dev/jwst/scripts/strun", line 14, in <module>
    step = Step.from_cmdline(sys.argv[1:])
  File "/Users/jdavies/dev/jwst/jwst/stpipe/step.py", line 142, in from_cmdline
    return cmdline.step_from_cmdline(args)
  File "/Users/jdavies/dev/jwst/jwst/stpipe/cmdline.py", line 320, in step_from_cmdline
    step.run(*positional)
  File "/Users/jdavies/dev/jwst/jwst/stpipe/step.py", line 356, in run
    result = self.process(*args)
  File "/Users/jdavies/dev/jwst/jwst/pipeline/calwebb_image3.py", line 97, in process
    out_catalog = self.source_catalog(output)
  File "/Users/jdavies/dev/jwst/jwst/stpipe/step.py", line 356, in run
    result = self.process(*args)
  File "/Users/jdavies/dev/jwst/jwst/source_catalog/source_catalog_step.py", line 42, in process
    npixels, deblend=deblend)
  File "/Users/jdavies/dev/jwst/jwst/source_catalog/source_catalog.py", line 155, in make_source_catalog
    if micro_Jy > 0.0:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

The MIRI imaging test data that is throwing this error has the following `meta.photometry` block:
```
{'conversion_megajanskys': 2.449465751647949,
 'conversion_microjanskys': 57.57329629700661,
 'pixelarea_arcsecsq': 0.0121,
 'pixelarea_steradians': 2.84403609523084e-13}
```
Not sure how this block gets populated.  Other imaging test data I have from NIRCam doesn't have it and doesn't throw this error.

hbushouse:
[...]

PR #981, opened 2017-06-14
Fix flux calculation in source catalog

larrybradley:
Fixes the flux comparison bug reported #976.  The other issue reported in #976 will require updated NIRCam reference files from the instrument team.

### DIFF
--- a/jwst/source_catalog/source_catalog.py
+++ b/jwst/source_catalog/source_catalog.py
@@ -81,9 +81,9 @@ def make_source_catalog(model, kernel_fwhm, kernel_xsize, kernel_ysize,
     """
 
     if not isinstance(model, DrizProductModel):
-        raise ValueError('The input model must be an DrizProductModel.')
+        raise ValueError('The input model must be a DrizProductModel.')
 
-    # Remove until model.wht contains an IVM map
+    # Use this when model.wht contains an IVM map
     # Calculate "background-only" error assuming the weight image is an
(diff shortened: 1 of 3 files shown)


jwst/source_catalog/source_catalog.py
```diff
@@ -144,24 +144,28 @@ def make_source_catalog(model, kernel_fwhm, kernel_xsize, kernel_ysize,
     catalog['orientation_sky'] = ((270. - rot +
                                    catalog['orientation'].value) * u.deg)
 
-    # define AB mag and AB mag error
+    # define flux in microJanskys
+    nsources = len(catalog)
     pixelarea = model.meta.photometry.pixelarea_arcsecsq
     if pixelarea is None:
-        micro_Jy = 0.0
+        micro_Jy = np.full(nsources, np.nan)
```
philhodge: If you don't specify the data type, will this give you the data type that you expect?

larrybradley: It returns the data type of the fill value.  In this case it's `np.nan`, which are always float (which is desired/expected).

philhodge:
np.nan gives you float64 rather than float32; that's the distinction I 
was referring to.  I'm all for double precision when it's possible to 
use it, but sometimes you do want single precision.

larrybradley:
Yep, `float64` is what I intended.  Thanks.

larrybradley:
`travis-ci` failures are unrelated...merging.

Issue #465, opened 2016-11-16
WCS keywords for MultiProduct files


hbushouse:
Most JWST data products currently contain the basic FITS WCS keywords in the primary header, because they usually apply to all image extensions in the file. This starts to break down already as soon as we create MultiSlit products, where each slit instance (SCI, ERR, DQ tuple) is likely to have unique WCS parameters. This is currently overcome by the fact that the WCS object in the ASDF extension has the appropriate WCS information for each slit and that's what the cal pipeline steps rely on. So to a certain extent the "stale" WCS keywords in the primary header are ignored. But when we create resampled versions of MultiSlit products, which are now known as MultiProduct models, the WCS for each of those products can now be represented by a simple set of FITS WCS keywords, because the data are now rectified and linearized. Hence we should at that point move the simple FITS WCS keywords to the header of each SCI extension and populate them with accurate values from the WCS object.

hbushouse:
As opposed to moving the WCS-related keywords from the primary header to extension headers in the midst of cal pipeline processing, we have decided to have all WCS-related keywords located in the extension headers to begin with, starting at the level-1b product stage (produced by SDP). Updates to the Keyword Dictionary schemas will be necessary to indicate the new home for the WCS keywords.

hbushouse:
Closed by #982 and #1105.

PR #982, opened 2017-06-14
Move WCS keywords to SCI extension

hbushouse:
Initial work to satisfy #465. This moves the couple dozen standard FITS WCS keywords from the primary header to the SCI extension header by updating their definitions in the core schema. Has been tested on a FITS file in which I manually (astropy.io.fits) moved the keywords from the primary to the SCI header and then loaded the modified FITS file into a datamodel. The model.meta.wcsinfo block appeared to be populated correctly and all keywords again appeared in the SCI header when the model was saved to a new file.

The only minor annoyance at this point is simply cosmetic: the section header "WCS parameters" now only appears in the SCI header, which means the rest of the keywords from this block that are left behind in the primary (V2_REF, V3_REF, etc.) now appear as a run-on with keywords from the previous header section, with no section header of their own. This could be fixed by defining an additional block in the schema for these remaining primary header keywords, but that would also mean having to change any references to their meta data values in code modules, because they'd appear in a new meta block.


hbushouse:
Superseded by more recent PR's. Closing.

PR #983, opened 2017-06-15
Updates to photom to address #976

hbushouse:
Updates to the save_area_info function in photom, so that the nominal pixel area values stored in keywords in the pixel area reference file take priority over the same keyword values in the photom reference file. Instead of not copying any pixel area values to the output (science data model) when the two sets of pixel area values don't agree, just issue a warning about the discrepancy and still copy the values from the pixel area reference file, so that there are values to work with in later steps, such as source_catalog.

Addresses #976 and related to #981.


hbushouse:
Unrelated Travis failures. Merging.

PR #984, opened 2017-06-15
Update remaining reference file schemas to use new approach

hbushouse:
Update more reference file schemas in support of #861. These are the reference image, cube, and quad models. I don't think they're in use by any code modules anywhere, but they are used in the datamodels.open utilities to try to open reference files in a generic way.


hbushouse:
Unrelated Travis failures. Merging.

PR #985, opened 2017-06-15
Explicitly set backwards compatible behavior of AsdfFile.open ...

drdavella:
AsdfFile.open recently changed its implementation to automatically
process FITS files that contain ASDF extensions. However, the way that
this function was being used here expected a ValueError for all input
other than pure ASDF files. This change updates the call to
AsdfFile.open to explicitly request the old behavior. However, in the
long term this should be refactored to no longer require such behavior.


drdavella:
This PR requires the changes in https://github.com/spacetelescope/asdf/pull/250 to be merged first.

PR #989, opened 2017-06-20
New generalized and optimized MIRI MRS cube background matching algorithm

mcara:
This PR adds new and improved (compared to ``cube_skymatch``) background matching algorithm. This algorithm computes parameters of the background polynomials (of arbitrary degree - but don't go crazy 😄) by minimizing the total sum of errors. The core algorithm is located in the package ``wiimatch`` which can also handle arbitrary data dimensionality. The new step is called ``mrs_imatch``.

CC: @hbushouse


hbushouse:
Only Travis failure was due to a processing time limit. Merging.

PR #990, opened 2017-06-21
fixed error when running single = true

jemorrison:
Fixed an error when running cube_build single=true. No longer is it assumed that the ModelContainer has association info. Also fixed a error in writing the FLUX_EXT, ERR_EXT and WMAP_EXT names to the header. 
@mcara 
@hbushouse

### DIFF
--- a/jwst/cube_build/cube_build.py
+++ b/jwst/cube_build/cube_build.py
@@ -181,8 +181,9 @@ class CubeData(object):
         cube_build_io_util.check_cube_type(self)
 
         self.output_name = cube_build_io_util.update_output_name(self)
-        log.info('Output Name %s',self.output_name)
-        log.info('Output Base %s ', self.output_name_base)
+        if not self.single:
+            log.info('Output Name %s',self.output_name)
+#            log.info('Output Base %s ', self.output_name_base)
 #________________________________________________________________________________
 # InstrumentDefaults is an  dictionary that holds default parameters for
 # difference instruments and for each band
@@ -827,10 +828,8 @@ class CubeData(object):
         err_cube = np.zeros((naxis3, naxis2, naxis1))
 
         IFUCube = datamodels.IFUCubeModel(data=data, dq=dq_cube, err=err_cube, weightmap=idata)
-
-
-
         IFUCube.update(self.input_models[j])
+
         IFUCube.meta.filename = self.output_name
         if self.single:
             with datamodels.open(self.input_models[j]) as input:
@@ -838,7 +837,6 @@ class CubeData(object):
                 # back to model container - do we want to define
                 # a new KEYWORD - filename_org ?
                 #IFUCube.meta.filename = input.meta.filename
-
 
                 filename = self.input_filenames[j]
                 indx = filename.rfind('.fits')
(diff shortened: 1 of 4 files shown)


jwst/cube_build/cube_build.py
```diff
@@ -874,10 +872,10 @@ def setup_IFUCube(self,j):
 
         IFUCube.meta.wcsinfo.wcsaxes = 3
 
-        IFUCube.flux_extension = 'SCI'
-        IFUCube.error_extension = 'ERR'
+        IFUCube.meta.flux_extension = 'SCI'
+        IFUCube.meta.error_extension = 'ERR'
         IFUCube.meta.dq_extension = 'DQ'
-        IFUCube.meta.weightmap = 'WMAP'
+        IFUCube.meta.weightmap_extension = 'WMAP'
```
hbushouse: The IFUCubeModel data model schema doesn't have a "weightmap_extension" attribute defined. The only ones that David Law asked us to define, to keep in sync with other IFU projects, were flux, error, and dq. So you don't need line 878.

jwst/cube_build/cube_build.py
```diff
@@ -874,10 +872,10 @@ def setup_IFUCube(self,j):
...
         IFUCube.meta.data_model_type = 'IFUCubeModel'
```
hbushouse: data_model_type is now defined in the core schema and automatically populated by the data model library when a model is created. So you don't need to set the value here (it can't hurt, it just isn't necessary). Also, you should remove the definition of data_model_type from the ifucube.schema.yaml (in datamodels/schemas/), because it's now already defined in core.schema.yaml which is inherited by ifucube.schema.yaml.

jwst/datamodels/schemas/ifucube.schema.yaml
```diff
@@ -18,6 +18,10 @@ allOf:
           title: EXTNAME of the extension containing data quality mask
           type: string
           fits_keyword: MASKEXT
+        weightmap_extension:
+          title: EXTNAME of the extension containing weight map
+          type: string
+          fits_keyword: WMAPEXT
```
hbushouse: Did someone ask to have WMAPEXT added as another keyword? The FLUXEXT, ERREXT, and MASKEXT were asked for by David Law a couple years ago, because that's what some other observatories are using in their IFU data, and hence we want ours to be consistent so that data analysis tools can find things. Is anyone else using WMAPEXT?

jemorrison:
A few questions:
I added the weight map extension because all the other data products had 
extensions in the header.  The data was being
written to the final IFUCube product it just was not listed in the 
header. A few weeks ago
Mihai email me that he wanted the WeightMap data and was concerned it 
was not part of the output products. I thought maybe
since it was not listed in the Primary header with an extension name 
that is was confusing if it existed or not.
Mihai needs this data  in this background tool. So I am a little 
confused if I should take out the IFUCube.meta.weightmap_extension or not

hbushouse:
The meta attributes model.meta.flux_extension, .error_extension, etc. don't map to or have anything to do with the presence of actual data array extensions that contain data. They simply map to header keywords that give the EXTNAME's of the extensions that contain the flux, error, and dq data. The IFU community apparently started using these keywords (FLUXEXT, ERREXT, MASKEXT) in order to provide data analysis tools with a pointer as to where to find the flux, error, and mask (dq) data in the files. We don't use them for any other data products than IFU cubes. The weight map extension containing weight map data will still appear in the cube product regardless of whether this keyword (meta data) entry appears. They have no direct effect on one another. So you can remove the population of model.meta.weightmap_extension (which isn't defined in the schema anyway) without losing the actual WMAP data extension in the cube product (FITS file).

jemorrison:
I did add it to the schema. But I can take it out. It seems more 
complete to keep it - but it sound like
it may just add confusion so I will take it out.

Jane

hbushouse:
To my knowledge the rest of the IFU community has only agreed to using FLUXEXT, ERREXT, and MASKEXT, to point to the extensions containing flux, errors, and DQ mask. No pointer to anything like a weight map. So I'd leave out this keyword (but still of course put the actual weight map in the product).

PR #994, opened 2017-06-21
Augment crds exceptions

jaytmiller:
This patch adds automatic debug information (input parameter dump) in the event of a CRDS exception including bad file results.


jaytmiller:
I'm withdrawing this for now,  it still needs more work.

PR #995, opened 2017-06-21
Remove duplicate crds console logging

jaytmiller:
This patch removes the default CRDS console logging handler and instead relies on the handler installed by stpipe.   This prevents duplicates of the same log message,  one starting with "CRDS" and one in the normal stpipe log format.


jaytmiller:
Withdrawing this too in favor of one unified "improve CRDS exception handling and misconfiguration management" TBD pull.

PR #997, opened 2017-06-22
Allow script to execute if a GROUP extension does not exist

### LINKED ISSUE #996: set_bary_helio_times modified to operate on data that does not have GROUP (GS data)
M.Swam requests:

> We find we need a small adjustment in the SSB set_bary_helio_times.py tool, so that it is able to run on Guide Star Data product files.  Currently the tool expects a GROUP extension (which science files have) and goes through the times in that table to do conversions, but GS Data doesn’t have such an extension it it’s product files.  Can the tool be altered to just ignore a missing GROUP extension, so it would work for both science data and guide star data files? 

H.Bushouse further specifies:

> When there's no GROUP table extension, the routine would only compute the time values for the start, midpoint, and end of the exposure, which are stored in primary header keywords.

stscieisenhamer:
Resolves #996

### DIFF
--- a/scripts/set_bary_helio_times.py
+++ b/scripts/set_bary_helio_times.py
@@ -37,15 +37,21 @@ for each group by adding the necessary columns to that table.
 '''
 
 from __future__ import print_function, division
-from jwst import timeconversion
-import sys
-import astropy.io.fits as fits
+
 import argparse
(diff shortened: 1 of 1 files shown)


hbushouse:
👍

PR #1000, opened 2017-06-23
Hide MultiExposure in a SourceContainerModel

stscieisenhamer:
This will allow, in theory, the use of MultiExposureModels everywhere a ModelContainer is expected. Also, in theory, when saving a SourceContainerModel, the actual MultiExposureModel will be the saved model.

In theory...


jwst/pipeline/calwebb_spec3.py
```diff
@@ -2,7 +2,7 @@
 
 from ..stpipe import Pipeline
 from .. import datamodels
-from ..exp_to_source import exp_to_source
+from ..exp_to_source import multislit_to_container
 
 # step imports
 from ..skymatch import skymatch_step
```
hbushouse: Mihai's MRS background matching step is now in place, but bears a different name. You'll need:

from ..mrs_imatch import mrs_imatch_step

with similar updates to the list of step defs.

stscieisenhamer: So `mrs_imath_step` replaces `skymatch_step`, correct?

stscieisenhamer: And, should the user-level step name change? If so, what should it be?

hbushouse: Yes, mrs_imatch_step replaces skymatch_step. You can call it whatever you like for use within the calspec3 module (i.e. the mapping within the step_defs). Not exactly sure what you mean by the user-level step name. If you mean the name that would be used to run the step individually, yes they would need to use either mrs_imatch or MRSIMatchStep. Unless we decide we want to rename it to something a little simpler.

stscieisenhamer: Sorry, but yes, I was referring to the `--steps.?????` specification the user sees. So, it looks like it should be `--steps.mir_imatch` and no longer `--steps.sky_match`

hbushouse: That's `--steps.mrs_imatch` (not `mir_imatch`), but yes, other than the typo, that's it.

PR #1004, opened 2017-06-26
Change code to close hdulist in datamodels.open

bernie-simon:
Datamodels.open opens a fits file as a hdulist in order to examine the
header andshape and determine which class to use to open the
file. Since it opens the hdulist, it should also close it. Previously
I had the open function close the hdulist, which was wrong. (What was
I thinking?) The code now appends the hdulist to the list of files to
close when closing the datamodel.


hbushouse:
👍   This certainly seems to fix the error that the regression tests were throwing. Let's see how they all do with the new build overnight.

Issue #615, opened 2017-01-18
Task: Outlier detection for calwebb_coron3


hbushouse:
The outlier detection step must be called within the calwebb_coron3 pipeline, as it is already done within calwebb_image3. The coronagraphic data are imaging data, so there should be relatively little work needed to adapt it for use. The only real difference is that the input data will be in the form of 3D data models, consisting of stacks of 2D images (multiple integrations), hence some updates will be needed just to handle the different organization of the data.

hbushouse:
This is a continuation of Trac ticket 363: https://aeon.stsci.edu/ssb/trac/jwst/ticket/363

hbushouse:
Closed by #1010 #1018 and #1174.

Issue #619, opened 2017-01-19
Task: Outlier detection for calwebb_tso3 pipeline


hbushouse:
Outlier detection is needed within the calwebb_tso3 pipeline, because many time-series observations will consist of integrations with very small numbers of groups, which will make it difficult for the preceding jump step to detect outliers within the integrations (ramps). The existing outlier_detection step should require little, if any, modification to work with time-series data, but small changes may be needed in order to simply handle the different format of the data for time-series exposures.

hbushouse:
Closed by #1117 #1174 and #1200.

Issue #616, opened 2017-01-18
Task: calwebb_tso3 pipeline module


hbushouse:
We need to create a calwebb_tso3 pipeline module to apply level-3 processing steps to time-series mode data. The list of steps and information about them is available on the JCCWG confluence site at https://confluence.stsci.edu/display/JWSTCC/CALTSO3

This is a continuation of Trac ticket 367: https://aeon.stsci.edu/ssb/trac/jwst/ticket/367

hbushouse:
Closed by #1117 #1174 and #1200.

PR #1010, opened 2017-06-27
Add outlier_detection to coron3 pipeline step

stsci-hack:
This work updates the outlier_detection step to implement a new parameter, resample_data.  This parameter controls whether resampling is done in outlier_detection or not.  The value of this parameter was then hard-coded in the calwebb_coron3 pipeline step to 'False' since resampling is not needed for coronographic data.  This change in logic should NOT change how outlier_detection works for any other step that calls it since all the original code doing the resampling has simply moved inside if-else blocks.  

The calwebb_coron3 step was also updated to create and then to write out the L2C product generated from outlier_detection.  

These changes were only tested interactively under Python to insure that when the correct inputs are provided, the updated outlier_detection code would work on coronographic data.  However, calwebb_coron3 step as a whole still needs CRDS updates for PSFMASK reference file before it can be tested end-to-end as a pipeline step.


jwst/pipeline/calwebb_coron3.py
```diff
@@ -123,13 +123,15 @@ def process(self, input):
...
-            # TEMPORAY HACK UNTIL OUTLIER_DETECTION IS VIABLE
-            # Create a dummy level-2c output product
-            log.warning('Creating fake outlier_detection results until step is available')
+            # Create a level-2c output product
+            log.debug('Creating and saving Level-2C result')
             lev2c_name = mk_filename(self.output_dir, target_file, 'calints-'+asn['asn_id'])
             lev2c_model = psf_sub.copy()
+            # Update/replace L2B product DQ array with L2C results
+            for i in range(len(target_models)):
+                psf_sub.dq[i] = target_models[i].dq
```
hbushouse: In line 131 lev2c_model is created as a copy of psf_sub. So shouldn't it be lev2c_model.dq that gets updated from target_models.dq?

hbushouse:
Applies to #615

hbushouse:
@stscieisenhamer When you have a chance we could use your help in cleaning up and/or streamlining the data flow and saving of products through calwebb_coron3, the same way you've been doing for calwebb_spec3. Of course it actually works as-is, it just isn't "pretty."

jwst/pipeline/calwebb_coron3.py
stsci-hack: Good catch, as you are correct that lev2c_model should be updated.  I will straighten this out shortly...

hbushouse:
@jdavies-st Any comments or suggestions on this before Warren merges?

hbushouse:
Fixes #615. Updates to outlier_detection also fulfill #619 and will be used by #616.

PR #1018, opened 2017-06-30
Min crds only

stsci-hack:
The outlier_detection STEP code understands how many unique observations it is working with.  However, the usual pipeline operations with CRDS blindly has CRDS run bestref functionality for each input, which for coronographic or TSO data can involve hundreds of inputs(slices) all from the same observation.  

This logic in outlier_detection_step will reduce the number of CRDS calls to 1 for those inputs which are derived from a 3D cube of data (e.g., TSO data).  This logic defaults to the previous code for cases where more than 1 reference file is specified for the stack of inputs.  

**This code is also expected to be temporary** and should be removed after CRDS gets updated to recognize and efficiently handle such duplicate calls. This PR should be evaluated by @stscieisenhamer  and/or @hbushouse  to determine whether this should be used at all.


jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -96,11 +95,18 @@ def _build_reffile_container(self, reftype):
         reffile_to_model = {'gain': datamodels.GainModel,
             'readnoise': datamodels.ReadnoiseModel}
 
-        reffiles = [self.get_reference_file(im, reftype) for im in self.input_models]
+        if reftype in self.input_models[0].meta.ref_file._instance:
+            reffile_names = [tm.meta.ref_file._instance[reftype]['name'] for tm in self.input_models]
+            if len(set(reffile_names)) == 1:
+               reffiles = [self.get_reference_file(self.input_models[0],reftype)]*len(reffile_names)
+            else:
+                reffiles = [self.get_reference_file(im, reftype) for im in self.input_models]
```
hbushouse: So it looks like the call to get_reference_file, which invokes CRDS, is still being made even when the ref file name was already available from the meta data in the input model(s)? From what I can tell, the only increase in efficiency is in the case where all the input models had the same ref file name in the meta data, in which case only 1 call is made to get_reference_files (and hence CRDS), rather than calling it for every input.

Why not just use the ref file name(s) retrieved from tm.meta.ref_file._instance[reftype]['name'] in line 99 and forget about calling get_reference_file altogether? Only resort to calling get_reference_file when the names were not available in the meta data (i.e. as is done in line 105).

All get_reference_file is doing is returning the file **_name_** for you, which you already have in the meta data. It doesn't actually get (open and load) the reference file. You have to do that using a datamodels.open() call of some sort (usually invoking the appropriate data model type, such as datamodels.GainModel() or datamodels.ReadNoiseModel()).

stsci-hack: I looked at doing that, but the filename stored in the metadata comes wrapped in the URL for CRDS which needs to be interpreted before it can be used.  The current implementation results in a filename that can be used directly, unlike the value in the metadata.

nden: I believe @jaytmiller  has implemented a function which expands the name saved in `meta.ref_file.name` . Perhaps he or @stscirij can confirm this.

stscirij:
a = Step()
local_filename = a.reference_uri_to_cache_path("crds://blah.blah.blah")
it's a method on any Step object

jaytmiller:
Nadia and Robert are correct about the Step method for converting the CRDS URI into a cache path so I was thinking this was a dead pull request.

Revisiting today,  I think there's an issue with using Step sub-classes on model containers.   I had this sense when associations were added,  there might need to be something like a "ContainerStep",  if only to sort out how reference files work given multiple model inputs.

Steps come with the assumption of a reference pre-fetch and with recording reference file results...  so that needs to be expanded somehow to handle N-model-input.

The assumption when you say:

there is an assumption that there will be two reference files required to execute that Step,  period.    They're declared so they can be automatically pre-fetched and used.    In theory they will be recorded in every result model.   Each result model has room for only one reference value per type.

So backing up:  how should these references be pre-fetched?   how should the results be recorded?

hbushouse:
Hmmm, this is going to be non-trivial in the general case of doing a datamodels.open() on an input that's an ASN file name, which then results in a ModelContainer being returned. The reason for this is that there often can be cases where the multiple models contain data from different detectors (e.g. the 8 detectors that make up the short-wavelength module of NIRCam), and some reference files are detector-specific. So you can't just make the simplifying assumption that it's good enough to do the pre-fetch based on, say, just the first model in the container and apply that result to all of them. It would need to be done for each model individually.

We haven't run into any problems with this yet, because at least until recently the level-3 pipelines - where ASN's are always used as input - weren't using any reference files. This, however, is starting to change. It's also now the case that some of the level-2b pipelines use ASN's as input and they use reference files that are detector-specific, such as flats. So I'm actually kind of surprised that no one has yet noticed a problem with this. Perhaps it's because any issues with the pre-fetch are sort of ignored and then when a given step, like flat_field, is executed, it gets executed on each input, one at a time, so that when the step does its call to "get_reference_file", it gets the correct reference file for that particular (individual) data model, does its work, and then gets called again for the next data model in the container. So in that case it all works out OK.

jaytmiller:
I think in general this is probably "working" because (a) the pre-fetch is currently disabled for associations and (b) nobody is looking at the record of reference files used in the container-related output products...  so the Step "doing the wrong thing" is just overlooked.

I understand about using a single model for common parameters not working.   Fixing the pre-fetch does not look hard however;   iterating over the ModelContainer and recursively pre-fetching looks sufficient,  there are no other Step side effects I see.

A long standing inconsistency of Step is that get_reference_file() accepts a model parameter in addition to the reference type...  which is not consistent with the pre-fetch that assumes a single input model from the command line or preceding Step is in use,  and not consistent with reference file recording code that assumes there is only one reference of each type,  not one per model and type.

Warren's _build_reffile_container() seems like a good start for generic ContainerStep code that supports **defining and using** reference files per-model.    How they are recorded in outputs is another area where ContainerStep would define an abstract interface that sub-classes might customize and implement.

jaytmiller:
Correction:  Step currently disables outputting assigned references for associations.

hbushouse:
I would say to go ahead and merge this for now, so that we can speed up the caltso3 pipeline (it's a real dog right now, making hundreds of calls to bestref), at least as a workaround. If a better solution comes up in the future, perhaps within CRDS itself, we'll cross that bridge when it gets built (if you build it ...)

jdavies-st:
This PR has broken `outlier_detection` in `Image3Pipeline`.  The regression test `test_nrc_image3_1.py` has been broken since this was merged, and I've seen the same issue with some of my local test data I'm using to try to fix some other issues.

What seems to be going on is that CRDS is not queried at all now in `outler_detection`, and instead, whatever happens to be stored in the `meta.ref_file` dictionary is being used.  But of course this can't work, as one needs to at least query CRDS once to correctly populate that dictionary.

My suggested solution would be to revert this.  But I don't know how the TSO pipeline depends on this.

If the TSO pipeline needs to specially handle how it pulls reffiles, then it should do it in `outlier_detection_tso_step` I think.

Thoughts @stsci-hack?

jaytmiller:
I thought the concept behind using meta.ref_file was that it was already populated with usable reference  files,  and that those were to be used as an optimization to avoid calling CRDS.  The hang up for that was that they used crds:// notation as an abstract cache path,  so Warren had to call a method to translate it to an absolute cache path before using them.

What exactly is wrong with the references?  Wrong path?  Undefined?  Nowhere to be found?  Wrong file?

jdavies-st:
Wrong file.  One needs to first populate `meta.ref_file` with something correct first, which means calling CRDS at least once.  But this is not done in this version of the code.

Or does `meta.ref_file` get populated when the prefetch happens?

jdavies-st:
To be clearer, generally speaking for the two reftypes that `outlier_detection` uses `['gain', 'readnoise']`, the `meta.ref_file` entries for these get updated in the level 1 processing.  But the assumption that one can just use those same reffiles for the level 3 processing is not a good one, as lots of our test data is built from either simulations, or specially made at sometime in the past from CV3 test data.

So the input `_cal.fits` files that come into the level 3 pipeline may not have the proper CRDS reffiles populated in `meta.ref_files`.  Best to check to be sure, and not short-circuit the standard pipeline procedure, especially when it is not necessary in `Image3Pipeline`, which is where `outlier_detection` is currently being used.

nden:
This may be irrelevant for this issue, but if any WCS ref files are used, it's probably best to use the ones that were recorded in meta.ref_file since this will keep results consistent.

jemorrison:
Ok I am beginning to think that an error I am having is related to this 
topic.
I am trying to run the step mrs_imatch. Now the mrs_imatch step calls 
the cube_build_step.
The cube_build_step calls crds to get the needed reference file.
When cube_build is called from mrs_imatch  I get an error stems from 
trying to find the reference file 'cubepar'

BUT when I insert some dummy code before I call the mrs_imatch step that 
first
calls the cube_build code. The first call the cube_build has no error 
and correct cubepar file is found.
Then when mrs_imatch is run after first calling cube_build the original 
problem I had with
  mrs_imatch having and error goes away too.

Should  I get looking at the meta.ref_file entries to figure out what 
the HECK is going on.

I am very puzzled by this behavior.

jaytmiller:
I think we need Warren to weigh in,  the strategy of using meta.ref_file may be unworkable.

I reviewed the step.py code and unless I missed something meta.ref_file is updated after processing,  effectively after both prefetch and runtime  fetch.

hbushouse:
The idea was that entries like meta.ref_file.gain.name and meta.ref_file.readnoise.name would have been populated already in the input files when they were run through an earlier pipeline stage, like calwebb_sloper. They were not relying on them being populated in near real time via a prefetch in the current execution.

In the future we should be able to drop all of this nonsense by having steps like outlier_detection use what will (hopefully) be reliable values in the input ERR arrays, rather than needing to use ref files like gain and readnoise to recompute noise estimates on the fly, hence they won't need the ref files anymore.

jaytmiller:
NOTE to Todd:  long,  maybe obvious and redundant.

Because one of the root processes involved here is exploding a cube into many files (with different names but common matching parameters),  the goto optimization of hashing on a model or filename and returning prior results cannot work because the files are all different.    If you make 300 *different* function calls,  CRDS will do 300 different computations; something has to be done to hint to CRDS that they're effectively the same.    

That being the case,  rather than setting up a handshake so that CRDS can tell 300 different files are effectively the same (say __hash__()),   it'd be better just to call get_reference_file() once on the cube or some representative file;  that leverages knowledge only the Step has.

If it's not clear:  I believe what is slow about get_reference_file() is the act of opening the dataset model and reading out the matching parameters.   Explicit garbage collection is also surprisingly slow.   Since get_reference_file() is called for every type in a Step,  there may still be room for optimization in crds_client.py so that it only constructs the matching parameter dictionary for the same filename/model once.   That however will not help for outlier_detection as far as I can see,  it would only help Steps with small numbers of input models (say 1) and large numbers of reference types (say  >= 5-10).   The optimization idea is to use the parameters read from a single model several times while reading it only once.   Since it doesn't apply to 100's of different models without hints,  it may not be worth it.

CRDS internal file i/o,  i.e. loading the rules and type information,  is already minimized and I believe would take no more than 5 seconds even on slow Isilon.   Hence,  I don't think further optimization opportunities lie inside the CRDS package side of the fence until timing gets into the sub-second range.  Opportunities are more likely in crds_client.py or Step sub-classes.

hbushouse:
I tend to agree that this logic that relies on the values of the R_GAIN and R_READNO keywords in the inputs should probably be confined to TSO data processing, where you're just trying to avoid having to call the CRDS "bestref" hundreds or thousands of times for the same result. If we could assume that the ref files used during calwebb_sloper processing (which populated those keywords) were still valid we could just use them all the time, but even that logic would need to be made more robust, such as the case of the files being used in the regression tests right now that made use of personal ref files for the earlier processing (and stored in the R_GAIN and R_READNO keywords), which aren't available for use in our calimage3 or caltso3 reg test runs.

jdavies-st:
The issues I raised here have been fixed in #1174.  Thanks @stsci-hack!

PR #1019, opened 2017-06-30
Use a shallow copy when opening an image from another image

bernie-simon:
Datamodels can be opened in two ways: first by calling the generic open function and second by creating a new DataModel object. Several different kinds of input can be used to open a model. One is another model. When a model is used to open another model the code makes a copy of the previous model. Previously it made a deep copy. Now it makes a shallow copy, meaning that only the top level attributes of the original image are copied. This is faster and uses less memory, particularly in the case of the pipeline code, where the pipeline opens the input file and the input file is re-opened by each step. The copy method of DataModel still does a deep copy, which agrees with how
the copy method is used by the code. The revised code sets an internal attribute, _iscopy, to True if the model is a shallow copy. It checks this attribute when closing the model. If _iscopy is True, the close method is a no-op. This prevents the same model from being closed twice.

Current code in jwst should not need any modification, but beware of the case when the original model is closed before its copy is finished being used.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -113,9 +113,12 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
         # Determine what kind of input we have (init) and execute the
         # proper code to intiailize the model
         self._files_to_close = []
+        self._iscopy = False
+    
         is_array = False
         is_shape = False
         shape = None
+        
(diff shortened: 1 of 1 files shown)


mcara:
This should have referenced directly the original issue: https://github.com/STScI-JWST/jwst/issues/284

PR #1021, opened 2017-07-03
fix asdf links [skip ci]

nden:


nden:
This only fixes the link to asdf documentation. No need to run tests.

PR #1024, opened 2017-07-06
use reference file models in nirspec

nden:


nden:
Merging this to start using the new reference files just delivered to CRDS. Let's see if there's smoke coming out of RT.

PR #1026, opened 2017-07-07
Make wcs attribute a pointer to GWCS object for nddata interface

jdavies-st:
- This resolves my concerns in PR #803 that there's two WCS objects that are different in our datamodels since the subclassing of `NDData`.  Now the `NDData` interface points to the GWCS object if it exists.  Also, this PR leaves open the possibility that we may want to use astropy.wcs objects in rectified datamodels in the future.
- Minor change to make the `my_attribute` non-public.

### DIFF
--- a/jwst/datamodels/ndmodel.py
+++ b/jwst/datamodels/ndmodel.py
@@ -11,6 +11,8 @@ import collections
 from astropy.extern import six
 from astropy.units import Quantity
 from astropy.nddata import nddata_base
+from astropy import wcs
+import gwcs
 
 from . import util
 from . import filetype
@@ -105,7 +107,7 @@ def write(data, path, *args, **kwargs):
 #---------------------------------------
 
 class NDModel(nddata_base.NDDataBase):
-    def my_attribute(self, attr):
+    def _my_attribute(self, attr):
         """
         Test if attribute is part of the NDData interface
         """
@@ -162,15 +164,30 @@ class NDModel(nddata_base.NDDataBase):
     def wcs(self):
         """
         Read the world coordinate system (WCS) for the dataset.
+
+        Get the GWCS object or if it hasn't been set, generate an astropy WCS
+        object from the header info.
         """
-        return self.get_fits_wcs()
+        try:
+            val = self.meta.wcs
+        except AttributeError:
+            val = None
+        if val:
+            return val
+        else:
+            return self.get_fits_wcs()
 
     @wcs.setter
     def wcs(self, value):
         """
         Write the world coordinate system (WCS) to the dataset.
+
+        Write an astropy WCS or GWCS object as required.
         """
-        return self.set_fits_wcs(value)
+        if isinstance(value, wcs.WCS):
+            self.set_fits_wcs(value)
+        elif isinstance(value, gwcs.WCS):
+            self.meta.wcs = value
 
     @property
     def meta(self):
--- a/jwst/datamodels/ndmodel.py
+++ b/jwst/datamodels/ndmodel.py
@@ -169,12 +169,8 @@ class NDModel(nddata_base.NDDataBase):
         object from the header info.
         """
         try:
-            val = self.meta.wcs
+            return self.meta.wcs
         except AttributeError:
-            val = None
-        if val:
-            return val
-        else:
             return self.get_fits_wcs()
 
     @wcs.setter
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -394,14 +394,14 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
                 return None
         return self._shape
 
-    def my_attribute(self, attr):
+    def _my_attribute(self, attr):
         properties = frozenset(("shape", "history", "_extra_fits", "schema"))
         return attr in properties
 
     def __setattr__(self, attr, value):
-        if self.my_attribute(attr):
+        if self._my_attribute(attr):
             object.__setattr__(self, attr, value)
-        elif ndmodel.NDModel.my_attribute(self, attr):
+        elif ndmodel.NDModel._my_attribute(self, attr):
             ndmodel.NDModel.__setattr__(self, attr, value)
         else:
             properties.ObjectNode.__setattr__(self, attr, value)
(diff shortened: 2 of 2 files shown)


jwst/datamodels/ndmodel.py
```diff
@@ -162,15 +164,30 @@ def unit(self, value):
...
         """
-        return self.get_fits_wcs()
+        try:
+            val = self.meta.wcs
+        except AttributeError:
+            val = None
+        if val:
+            return val
+        else:
+            return self.get_fits_wcs()
```
nden: This could perhaps go under the exception block?

nden:
I still think having the NDdata interface in datamodels at this stage is premature because the two objects have different interface. This can only work consistently if the common interface (yet to be finished) is used. Now it just creates confusion since a user needs to know if assign_wcs run or not and how to call the two different objects.

jdavies-st:
Would it be better to not include the interface to `astropy.wcs` at all then?  Just don't do anything with the FITS WCS?

nden:
@hbushouse Should we merge this? 
@jdavies-st Can you rebase when you get a chance?

I wonder if we should write a warning out if `model.meta.wcs` is not used.

jdavies-st:
Wow, this PR is a moldy oldie.  Thanks for dragging it out of the cellar!  =)

I do think it would be good to merge some or all of this at some point.  Mostly to clean up the interface.

What sort of warning under what circumstances are you suggesting @nden?

nden:
I was thinking that if `assign_wcs` was not run `model.wcs` will be created and will probably be incorrect.
This will be confusing for later steps, I think even with an warning.

It seems to me the only way this isn't going to be confusing is to have only one wcs object. Since the NDData format was imposed on the DataModel, perhaps we should move Model.meta.wcs to Model.wcs and be done with this.

hbushouse:
Given the fact that, as far as I know, the only purpose of the NDData format being included here is because the Data Analysis folks wanted to use NDData objects in their tools, I really don't know what's a good way to do this, because I don't know how they want to use it. So perhaps someone who has knowledge of the DA tools can chime in and give some insight.

jdavies-st:
Agree on both counts @nden and @hbushouse.  I don't think we should merge this PR.  Rather, coming up with a better, forward-looking solution would be best.  I like the idea of moving the `wcs` object out of `meta`.

Also, it might be OK to just duck-type to NDData instead of inheriting from it.  I don't know if that will solve any further issues.  We really need to know what the Data Analysis folks want to do with it.

stscicrawford:
I'm not sure how it is being used in the DATB tools, but in `ccdproc`, we did use the nddata.wcs.   We had a function that removed it from the header as it was read in from FITS: 
https://github.com/astropy/ccdproc/blob/8ddcb82e3d1060f00497ef07aea4a9a29ed6a8ea/ccdproc/ccddata.py#L618
And then added it back as it was read out to FITS: 
https://github.com/astropy/ccdproc/blob/8ddcb82e3d1060f00497ef07aea4a9a29ed6a8ea/ccdproc/ccddata.py#L354

This does mean our 'data model' for CCDs does not strictly follow the FITS model (with WCS in headers), but it does make the rest of the coding clear about where WCS is suppose to be accessed.

bernie-simon:
> Also, it might be OK to just duck-type to NDData instead of inheriting from it

You need to consider the possibility that a coder may want to check the type of an object to see if it's NDData. The type of an object is part of the object's api.

nden:
I don't how this is used in DAT either. I imagine that the NDData object at some point will start using the implementation of the WCS APE in astropy which will provide a uniform interface to astropy.wcs and gwcs.wcs. It's possible that the pipeline can switch to using this interface too. In this case it'd be good to move the wcs object out of model.meta. But this needs more discussion and evaluation of possible impact on the pipeline.

larrybradley:
This isn't being used yet in any DA tools, but I think the point was that the JWST datamodels could be used to load data from FITS, instead of the DA tools needing to write custom loaders.  The DA tools could then use the `NDData` interface.

This is related to the discussion in #1805.

CC: @eteq

eteq:
@jdavies-st @nden - @mcara and I have encountered a need for this being consistent (for image reg & comb DATs). Can we maybe just get this PR in ASAP and get the "proper" fix of cleaning it all up out of `meta` as a follow-on.

That is, I think it's OK for the `.wcs` to be a gwcs - it's been an understanding from the start of `nddata` that the `.wcs` keyword's API is sort of "undefined" until APE14 is resolved, so we're not doing anything bad by making it be a gwcs.

sosey:
This may confuse the science team testing, they are expecting `model.wcs` to return the fits wcs and `model.meta.wcs` to return the gwcs object, it should be made really clear to them, especially if they are looking at intermediate pipeline products in a DAT tool, *which* representation is being returned so that even their code an be updated. It's already been a source of confusion.

eteq:
@sosey - Ah, thanks for pointing that out.  That actually means *definitely* this should get resolved ASAP: if the science team is using the FITS WCS they are not using the intended end-user WCS, so they are testing the wrong thing...

Is there some previous discussion about this (outside this PR) that can be referenced beyond the NDData-related docs/discussions?

eteq:
(one solution if they *need* the FITS one would be to create a `_fitswcs` attribute... kinda hacky but better than hiding from folks what will happen.

jdavies-st:
Nobody should be using `model.wcs` at all in datamodels.  This is not a documented attribute.  `model.meta.wcs` is the correct one.

Eventually in the future this may switch, but science teams should not be using `model.wcs` for anything.

sosey:
cross referencing https://github.com/spacetelescope/jwst/issues/2262 is related

stscieisenhamer:
@jdavies-st Is this PR relevant still? Should this be closed?

jdavies-st:
Well, the issue is still relevant.  We still have an object at `model.wcs` that points to a wrong `astropy.wcs.WCS` object instead of to the correct GWCS object at `model.meta.wcs`.

This particular PR probably is not the correct solution as it is.  My thoughts are that 2 things should probably happen:

1. Remove all the `astropy.NDData` subclassing in `datamodels`.  There are no unit tests for it, adds no functionality, and needlessly complicates the code.

2. Move the GWCS object to `model.wcs`, so that at least it duck-types `astropy.NDData`.  `model.meta` should be metadata, which the GWCS object certainly is not.  I believe @nden may have opinions about this as well.

Can we do this before `jwst 1.0` is released?

stscieisenhamer:
How about an issue be made of the points above and close this PR?

hbushouse:
Wasn't the `astropy.NDData` subclassing added in order for the data analysis folks to do something they needed in their code?

jdavies-st:
In theory that was the idea, but it never actually happened.  They don't use `jwst.datamodels` at all as far as I'm aware.  They just read in the FITS file using `astropy.io.fits` which is used to instantiate their internal representations of the data.

jdavies-st:
Closing this PR, as I don't think this is the correct solution to this issue.

Issue https://github.com/spacetelescope/jwst/issues/5380 captures what ought to be done.

PR #1027, opened 2017-07-07
Change the model container iterator to open one model at a time

bernie-simon:
The iterator that is part of the model container has been rewritten so that at each iteration the current model is opened and the previous model is closed. To support this, the code in model container has been changed so that models are not opened when the container is created and random access checks if the model is open and opens it if it is not. The purpose of this change is to reduce the memory footprint of code that uses model containers.

### DIFF
--- a/jwst/cube_build/cube_build_io_util.py
+++ b/jwst/cube_build/cube_build_io_util.py
@@ -489,7 +489,8 @@ class IFUCubeASN(object):
             with datamodels.ModelContainer(input) as input_model:
                 self.output_name =input_model.meta.asn_table.products[0].name
 
-            for model in input:
+            for i in range(len(input)):
+                model = input[i]
                 self.input_models.append(model)
                 self.filenames.append(model.meta.filename)
 #            print('number of models',len(self.filenames))
--- a/jwst/cube_build/data_types.py
+++ b/jwst/cube_build/data_types.py
@@ -70,7 +70,8 @@ class DataTypes(object):
                 with datamodels.ModelContainer(input) as input_model:
                     self.output_name =input_model.meta.asn_table.products[0].name
 
-            for model in input:
+            for i in range(len(input)):
+                model = input[i]
                 self.input_models.append(model)
                 self.filenames.append(model.meta.filename)
 #            print('number of models',len(self.filenames))
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -71,10 +71,7 @@ class ModelContainer(model_base.DataModel):
         if init is None:
             self._models = []
         elif isinstance(init, list):
-            for item in init:
-                if not isinstance(item, model_base.DataModel):
-                    raise ValueError('list must contain only DataModels')
-            self._models = init
+            self._models = init[:]
         elif isinstance(init, self.__class__):
             instance = copy.deepcopy(init._instance)
             self._schema = init._schema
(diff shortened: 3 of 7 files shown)


jdavies-st:
If the DataModels are being opened and closed every time you're iterating through the ModelContainer, then will this slow things down dramatically?  We often iterate through a container several times in a step or pipeline.  Of course maybe we should not be iterating several times, but for some steps, it is definitely needed to do it more than once.

Could we run some tests to see how much this PR decreases memory usage and how much it slows down existing pipelines before this is merged?

bernie-simon:
Since @jdavies-st has raised the reasonable suspicion that opening and closing the datamodels in the container iterator may slow things down, here are some timings, based on the jwst test code. The test cases marked (master) are the old code and the test cases marked (jit_container) have the iterator which opens and closes each datamodel. As you can see, the new code doesn't slow the test cases down.

(master) time pytest --runslow test_calwebb_image2.py

real	0m47.977s
user	0m43.628s
sys	0m2.486s

(jit_container) time pytest --runslow test_calwebb_image2.py

real	0m46.258s
user	0m42.888s
sys	0m2.284s

(master) time pytest --runslow test_calwebb_spec2.py

real	1m14.337s
user	1m10.503s
sys	0m2.918s

(jit_container) pytest --runslow test_calwebb_spec2.py

real	1m14.339s
user	1m9.504s
sys	0m3.205s

(master) time pytest --runslow test_calwebb_spec3_mir_lrs_fixedslit.py

real	0m8.432s
user	0m6.636s
sys	0m0.485s

(jit_container) time pytest --runslow test_calwebb_spec3_mir_lrs_fixedslit.py

real	0m4.333s
user	0m3.779s
sys	0m0.426s

(master) time pytest --runslow test_calwebb_spec3_mir_mrs.py

real	1m35.022s
user	1m32.112s
sys	0m1.641s

(jit_container) time pytest --runslow test_calwebb_spec3_mir_mrs.py

real	1m40.752s
user	1m42.772s
sys	0m1.720s

(master) time pytest --runslow test_calwebb_spec3_nrs_msa.py

real	1m39.475s
user	1m35.552s
sys	0m0.912s

(jit_container) time pytest --runslow test_calwebb_spec3_nrs_msa.py

real	0m47.348s
user	0m46.003s
sys	0m0.799s

hbushouse:
Does anyone know how many of these unit tests involve the use of multiple models stored in a ModelContainer, as opposed to a single ImageModel (or whatever) going through the various pipelines?

stscieisenhamer:
none of them open too many models; i know there are a couple cases where there are 3 or 4 models that get dealt with. I had already warned that these were that large in the sense of # of models, but that was the best readily had at the time. There is definite need to come up with a larger set.

bernie-simon:
If anyone has a good test case, one that iterates over a large number of datamodels in an association, I would appreciate getting it.

jdavies-st:
I can come up with a test case using simulated NIRCam data.

bernie-simon:
Here are some more timings. Two of the test cases had multiple models in their association files. In both cases I modified the files by duplicating the two files twenty times, making an association with 40 files. The intent was to see how large associations are handled by the current code (master) and the proposed change, which opens and closes each model in an association when iterating over the association (jit_container). As you can see, there is a significant speedup rather than the feared slowdown from the proposed change.. I would welcome other test cases, but for now it looks like the proposed change is a win.

(40 master) time pytest --runslow test_calwebb_spec3_mir_lrs_fixedslit.py

real	0m54.273s
user	0m48.701s
sys	0m4.312s

(40 jit_container) time pytest --runslow test_calwebb_spec3_mir_lrs_fixedslit.py

real	0m4.451s
user	0m3.971s
sys	0m0.385s

(40 master) time pytest --runslow test_calwebb_spec3_nrs_msa.py

real	75m45.455s
user	75m27.455s
sys	0m13.201s

(40 jit_container) time pytest --runslow test_calwebb_spec3_nrs_msa.py

real	46m39.533s
user	46m31.338s
sys	0m6.928s

jdavies-st:
Comparison using our regression test data suite for the Image3Pipeline (the subpix simulated 47 Tuc data from Hilbert):
```
(master)> time strun jwst.pipeline.Image3Pipeline mosaic_long_asn.json --steps.tweakreg.skip=True --steps.tweakreg_catalog.skip=True --steps.source_catalog.skip=True

real	2m17.847s
user	2m24.505s
sys	0m25.337s

(jit_container)>> time strun jwst.pipeline.Image3Pipeline mosaic_long_asn.json --steps.tweakreg.skip=True --steps.tweakreg_catalog.skip=True --steps.source_catalog.skip=True

real	2m15.933s
user	2m23.706s
sys	0m26.834s
```
So no difference.

jdavies-st:
Doing the same tests but on 12 datamodels in a modelcontainer from a very similar dataset to above, we have:
```
(master)> time strun jwst.pipeline.Image3Pipeline mosaic_short_asn.json --steps.tweakreg.skip=True --steps.tweakreg_catalog.skip=True --steps.source_catalog.skip=True

real	8m30.829s
user	9m33.644s
sys	1m40.721s

(jit_container)>> time strun jwst.pipeline.Image3Pipeline mosaic_short_asn.json --steps.tweakreg.skip=True --steps.tweakreg_catalog.skip=True --steps.source_catalog.skip=True

real	8m39.187s
user	9m36.996s
sys	1m42.170s
```
Again, virtually identical times.

PR #1029, opened 2017-07-07
Get file type from extension if file does not exist

bernie-simon:
This is not needed for how the function is currently used. But ignoring this case in the past has caused me problems, so it's best to handle the case of a non-existent file now when I remember it.

### DIFF
--- a/jwst/datamodels/filetype.py
+++ b/jwst/datamodels/filetype.py
@@ -16,11 +16,18 @@ def check(init):
     -------
     file_type: a string with the file type ("asdf", "asn", or "fits")
     """
-    
+
     if isinstance(init, six.string_types):
-        fd = open(init, "rb")
-        magic = fd.read(5)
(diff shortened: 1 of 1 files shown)


jwst/datamodels/filetype.py
```diff
@@ -16,11 +16,18 @@ def check(init):
...
-        magic = fd.read(5)
-        fd.close()
+        if os.path.exists(init):
+            fd = open(init, "rb")
+            magic = fd.read(5)
+            fd.close()
+        else:
+            filename, file_extension = os.path.splitext(init)
+            file_type = file_extension[1:]
+            if file_type not in ("asdf", "fits", "asn"):
```
hbushouse: Will the file name _**extension**_ ever have a value of "asn"? Currently our ASN files have "asn" as the last suffix of the root name, but an extension of "json".

PR #1038, opened 2017-07-12
CRDS Cache locking

jaytmiller:
Added CRDS cache file-based locking (based on the lockfile package) to the CRDS interface layer in the cal code to support associations multi-processing.   Since associations launch multiple concurrent processes,  it poses a problem for the shared CRDS cache resource.   This is one solution.

This should typically support remote users unhappy with VPN who can explicitly configure CRDS something like this:

```
CRDS_PATH=$HOME/crds_cache_ops
CRDS_SERVER_URL=https://jwst-crds.stsci.edu
```

When the cal code is run,  CRDS will dynamically update their personal cache with needed references based on their particular datasets.

The cal code integration is essentially two lines of incredibly dangerous code:

```
from crds.core import crds_cache_locking

with crds_cache_locking.get_cache_lock():
    ... do getreferences() ...
```

There are multiple conditions in CRDS that determine when locking is really used:

1. lockfile package must be installed and importable
2. the lock must be successully created
3. the CRDS cache must be physically writable
4. CRDS_USE_LOCKING must be undefined or 1
5. CRDS_READONLY_CACHE must be undefined or 0

otherwise,  cache syncing and locking are moot.

This code is not currently used for HST since there has been no demand for it and how well it performs in the real world should be demonstrated first.  In principle it could also apply to the CRDS sync tool,  or HST's CRDS bestrefs,  or to a CRDS cache rewrite in general.   This code should not apply to pipeline installations or local users who use the default JWST CRDS configuration.

`CRDS_LOCK_PATH 
`

can be used to define the location of file locks:  defaults to /tmp.   It should be noted that the existence of the lock file directory is itself a concurrency issue,  so it should be created or otherwise available before calibration code is run.

`$ crds sync --clear-locks 
`

can be used to remove orphan locks that are blocking processing.

Locking requires installation of lockfile and CRDS-7.1.4 or later.


jaytmiller:
Joe H. just merged the CRDS 7.1.4 release update into Astroconda,  so this should be ready to go.

It occurs to me that this is going to fail hard for users without 7.1.4.   

Should I modify this to handle a crds_cache_locking import failure gracefully?   

That would enable the old non-locking behavior for people who have not updated to 7.1.4.

jaytmiller:
Based on a discussion with Robert,  I tweaked the locking integration to handle an import failure of the new CRDS locking module with a warning and reversion to the old "unprotected" behavior which will still work for non-association cases.

PR #1045, opened 2017-07-14
Prototype code for FGS pipeline, only partially tested. The 3 calibra…

dmggh:
…tion steps in this pipeline are dq_init, flat_field, and guider_cds (which is the modified version of ramp fitting).  Two new datamodel classes are defined - the first is for the input and output of dq_init (and the input of guider_cds), and the second is for the output of guider_cds.

### DIFF
--- a/jwst/datamodels/fgs.py
+++ b/jwst/datamodels/fgs.py
@@ -1,0 +1,71 @@
+from __future__ import absolute_import, unicode_literals, division, print_function
+
+from . import model_base
+
+
+__all__ = ['FGSModel']
+
+
+class FGSModel(model_base.DataModel):
+    """
+    A data model for input for the FGS pipeline.
+
+    Parameters
+    ----------
+    init : any
+        Any of the initializers supported by `~jwst.datamodels.DataModel`.
+
+    data : numpy array
+        The science data. 4-D
+
+    dq : numpy array
+        The data quality array. 2-D.
+
+    plan_star : table
+        The planned reference star table
+
+    flight_star : table
+        The flight reference star table
+
+    pointing : table
+        The pointing table
+
+    centroid : table
+        The centroid packet table
+
+    track_sub : table
+        The track subarray table
+
+    """
+    schema_url = "fgs.schema.yaml"
+
+    def __init__(self, init=None, data=None, dq=None, 
+                 plan_star_table=None, flight_star_table=None, 
+                 pointing_table=None, centroid_table=None, 
+                 track_sub_table=None, **kwargs):
+        super(FGSModel, self).__init__(init=init, **kwargs)
+
+        if data is not None:
+            self.data = data
+
+        if dq is not None:
+            self.dq = dq
+
+        if plan_star_table is not None:
+            self.plan_star_table = plan_star_table
+
+        if flight_star_table is not None:
+            self.flight_star_table = flight_star_table
+
+        if pointing_table is not None:
+            self.pointing_table = pointing_table
+
+        if centroid_table is not None:
+            self.centroid_table = centroid_table
+
+        if track_sub_table is not None:
+            self.track_sub_table = track_sub_table
+
+        # Implicitly create arrays
+        self.dq = self.dq
+
--- a/jwst/datamodels/schemas/fgs.schema.yaml
+++ b/jwst/datamodels/schemas/fgs.schema.yaml
@@ -1,0 +1,132 @@
+1allOf:
+- $ref: core.schema.yaml
+- $ref: bunit.schema.yaml
+- type: object
+  properties:
+    data:
+      title: The output science data
+      fits_hdu: SCI
+      default: 0.0
+      ndim: 4
+      datatype: float32
+    dq:
+      title: Data quality array
+      fits_hdu: DQ
+      default: 0
+      ndim: 2
+      datatype: uint32
+    planned_star_table: 
+      title: Planned reference star table
+      fits_hdu: PLANNED_STARS
+      datatype:
+      - name: guide_star_order
+        datatype: int32
+      - name: reference_star_id
+        datatype: string
+      - name: ra
+        datatype: float64
+      - name: dec
+        datatype: float64
+      - name: id_x
+        datatype: float64
+      - name: id_y
+        datatype: float64
+      - name: fgs_mag
+        datatype: float64
+      - name: fgs_mag_uncert
+        datatype: float64
+      - name: count_rate
+        datatype: float64
+      - name: count_rate_uncert
+        datatype: float64
+    flight_star_table: 
+      title: Planned reference star table
+      fits_hdu: FLIGHT_STARS
+      datatype:
+      - name: reference_star_id
+        datatype: string
+      - name: id_x
+        datatype: float64
+      - name: id_y
+        datatype: float64
+      - name: count_rate
+        datatype: float64
+    pointing_table:
+      title: Pointing table
+      fits_hdu: POINTING
+      datatype:
+      - name: time
+        datatype: float64
+      - name: jitter
+        datatype: float64
+      - name: delta_ddc_ra
+        datatype: float64
+      - name: delta_ddc_dec
+        datatype: float64
+      - name: delta_aperture_pa
+        datatype: float64
+      - name: delta_v1_ra
+        datatype: float64
+      - name: delta_v1_dec
+        datatype: float64
+      - name: delta_v3_pa
+        datatype: float64
+      - name: delta_j1_ra
+        datatype: float64
+      - name: delta_j1_dec
+        datatype: float64
+      - name: delta_j3_pa
+        datatype: float64
+    centroid_table: 
+      title: Centroid packet table
+      fits_hdu: CENTROID_PACKET
+      datatype:
+      - name: observatory_time   
+        datatype: string
+      - name: centroid_time   
+        datatype: string
+      - name: guide_star_position_x
+        datatype: float64
+      - name: guide_star_position_y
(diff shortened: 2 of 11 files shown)


jwst/datamodels/schemas/fgs.schema.yaml
```diff
@@ -0,0 +1,132 @@
+1allOf:
```
hbushouse: The character "1" in the first column appears to be a typo. Should not be there.

jwst/datamodels/schemas/fgs.schema.yaml
```diff
@@ -0,0 +1,132 @@
...
+      - name: fgs_mag
+        datatype: float64
+      - name: fgs_mag_uncert
+        datatype: float64
+      - name: count_rate
+        datatype: float64
+      - name: count_rate_uncert
+        datatype: float64
+    flight_star_table: 
+      title: Planned reference star table
```
hbushouse: Looks like the title of this one should not be "Planned", same as the one above, but should be something like "Flight".

jwst/datamodels/schemas/guider_cds.schema.yaml
```diff
@@ -0,0 +1,132 @@
+1allOf:
```
hbushouse: The character "1" in the first column here also needs to be removed.

jwst/datamodels/schemas/guider_cds.schema.yaml
```diff
@@ -0,0 +1,132 @@
...
+      - name: fgs_mag
+        datatype: float64
+      - name: fgs_mag_uncert
+        datatype: float64
+      - name: count_rate
+        datatype: float64
+      - name: count_rate_uncert
+        datatype: float64
+    flight_star_table: 
+      title: Planned reference star table
```
hbushouse: Should be "Flight", not "Planned"

jwst/guider_cds/guider_cds_step.py
```diff
@@ -0,0 +1,32 @@
...
+log.setLevel(logging.DEBUG)
+
+class GuiderCdsStep (Step):
+
+    """
+    This step .....
+    """
+
+    spec = """
+        int_name = string(default='')
```
hbushouse: What's int_name intended to be used for?

jwst/guider_cds/guider_cds.py
```diff
@@ -0,0 +1,133 @@
...
+        readpattern ...number of groups per integration
+
+
+    """ 
+    instrume = model.meta.instrument.name
+    frame_time = model.meta.exposure.frame_time
+    ngroups = model.meta.exposure.ngroups
+
+    effinttm = model.meta.exposure.integration_time
+    readpatt = model.meta.exposure.readpatt
```
hbushouse: This could now be upgraded to use the EXP_TYPE keyword (model.meta.exposure.type), since we've convinced the SDP folks to include the EXP_TYPE keyword in guider products.

jwst/guider_cds/guider_cds.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Parameters
+    ----------
+    model: data model
+        input data model, assumed to be of type FGSModel
+    """
+    # get needed sizes and shapes
+    imshape, n_int, effinttm, readpatt = get_dataset_info(model)
+
+    if readpatt == 'ID': # force output to have single slice
```
hbushouse: This could now use the EXP_TYPE keyword. I'm not sure how consistent or useful the readpatt values are guider products.

jwst/guider_cds/guider_cds.py
```diff
@@ -0,0 +1,133 @@
...
+        new_model = datamodels.GuiderCdsModel((1,)+imshape)
+    else:
+        new_model = datamodels.GuiderCdsModel()
+
+    new_model.dq = model.pixeldq
```
hbushouse: I think the right-hand side of this needs to be model.dq, now that the guider-specific data models are being used, which have a .dq array and not a .pixeldq.

jwst/guider_cds/guider_cds.py
```diff
@@ -0,0 +1,133 @@
...
+    slope_int_cube = np.zeros(( n_int,) + imshape, dtype = np.float64)
```
hbushouse: I don't think this needs to be float64 data type. float32 should be fine, like the rest of data arrays in other data models.

jwst/guider_cds/guider_cds.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # loop over data integrations
+    for num_int in range(0, n_int):
+        data_sect = model.data[num_int, :,:,:]
+
+        if readpatt == 'FINEGUIDE':
```
hbushouse: Switch all instances of readpatt to exptype.

jwst/guider_cds/guider_cds.py
```diff
@@ -0,0 +1,133 @@
...
+            grp_last = data_sect[1, :, :]
+            grp_first = data_sect[0, :, :]
+            slope_int_cube[ num_int,:,: ] = grp_last - grp_first
+        
+    if readpatt == 'ID':
+        new_model.data[0,:,:] = np.minimum( diff_int1, diff_int0 )/effinttm
+
+    else:  # FINEGUIDE, ACQ1, ACQ2, or TRACK
+        new_model.data = slope_int_cube/effinttm
+
```
hbushouse: This all looks like the proper computations for the various modes. 👍

jwst/guider_cds/__init__.py
```diff
@@ -0,0 +1,4 @@
+from __future__ import absolute_import
+from .guider_cds_step import GuiderCdsStep
+
+__version__ = '0.7.0'
```
hbushouse: We should mark this as 0.7.1

jwst/guider_cds/setup.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
```
hbushouse: setup.py files are no longer used or needed for step packages. Just the `__init__.py`. So this should be removed.

jwst/pipeline/calwebb_guider.cfg
```diff
@@ -0,0 +1,10 @@
+name = "FGSPipeline"
+class = "jwst.pipeline.FGSPipeline"
```
hbushouse: I think we should try not to use the generic name "FGS" too much, because the FGS instrument can also be used in science mode. Because all of this code is specific to handling guide-mode data, I suggest more explicit names like "GuiderPipeline" and "jwst.pipeline.GuiderPipeline", which is also consistent with the "calwebb_guider" name.

jwst/pipeline/calwebb_guider.py
```diff
@@ -0,0 +1,49 @@
...
+from ..stpipe import Pipeline
+from .. import datamodels
+import os
+
+# step imports
+from ..dq_init import dq_init_step
+from ..flatfield import flat_field_step
+from ..guider_cds import guider_cds_step
+
+__version__ = "7.0.0"
```
hbushouse: set to 0.7.1

jwst/pipeline/calwebb_guider.py
```diff
@@ -0,0 +1,49 @@
...
+
+# Define logging
+import logging
+log = logging.getLogger()
+log.setLevel(logging.DEBUG)
+
+class GuiderPipeline(Pipeline):
```
hbushouse: Ah, here you are using "GuiderPipeline." Good. So the .cfg file just needs to be updated to be consistent.

jwst/pipeline/calwebb_guider.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    GuiderPipeline: For FGS observations, apply all calibration 
+    steps to raw JWST ramps to produce a 3-D slope product. 
+    Included steps are: dq_init, flat_field, and guider_cds.
+    """
+
+    spec = """
+        save_calibrated_ramp = boolean(default=False)
```
hbushouse: We shouldn't need this option for guide mode data. It can be removed.

jwst/pipeline/calwebb_guider.py
```diff
@@ -0,0 +1,49 @@
...
+                 'guider_cds': guider_cds_step.GuiderCdsStep,
+                 }
+
+    # start the actual processing
+    def process(self, input):
+
+        log.info('Starting calwebb_guider ...')
+
+        # open the input
+        input = datamodels.open(input)
```
hbushouse: I wonder if we need to explicitly open as an FGS/Guider data model, rather than using the generic .open() and relying on it to figure out what kind of data model to load it into. It might not be very good at recognizing this unique case.

jwst/datamodels/fgs.py
```diff
@@ -0,0 +1,71 @@
+from __future__ import absolute_import, unicode_literals, division, print_function
+
+from . import model_base
+
+
+__all__ = ['FGSModel']
+
+
+class FGSModel(model_base.DataModel):
```
hbushouse: For the 2 "raw" and "calibrated" versions of the data models, I would suggest trying not to use the somewhat generic "FGS" name, because the FGS can be used in science mode as well. These data models are specific to guide-mode data coming out of the FGS, so I think we should make that explicit in the data model names. And because the 2 data models are nearly identical to one another (the only difference is the shape of the SCI or "data" array), I would also suggest a common naming scheme for them, such as "GuiderRawModel" (or "GuiderRampModel", similar to our existing RampModel for raw science data) and "GuiderCalModel".

PR #1046, opened 2017-07-14
Add utility convert_to_uncal

stscieisenhamer:
[skip ci]

This utility takes a 'cal' based file and creates the equivalent
level1b file from it. Not very sophisticated; more of an example.

### DIFF
--- a/jwst/lib/convert_to_uncal.py
+++ b/jwst/lib/convert_to_uncal.py
@@ -1,0 +1,90 @@
+from astropy.io import fits
+from jwst import datamodels
+import numpy as np
+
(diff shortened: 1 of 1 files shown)


hbushouse:
As @stscieisenhamer said, this is strictly an example and is completely tied to performing the proper operations for one particular exposure. I also discovered after the fact that the output file still has both an ERR and ASDF extension. The ASDF extension is expected due to the fact that the output model contains a meta tree, but I'm not sure why the ERR extension gets carried over from the input (the DQ extension doesn't). So to make the output look exactly like a level-1b product, with just a SCI and GROUP table extension, you need to add something like:
```
f = fits.open(output_file, mode='update')
del f[2]
del f[-1]
f.close
```
at the end.

PR #1052, opened 2017-07-18
Do not include _i2d in source catalog filename

### LINKED ISSUE #1051: source_catalog output product names from calimage3
Running the calimage3 pipeline on an ASN that specifies a product name of "jw99999-a3001_t001_nircam_f410m-clear" results in an i2d output product named "jw99999-a3001_t001_nircam_f410m-clear_i2d.fits" and a source catalog product named "jw99999-a3001_t001_nircam_f410m-clear_i2d_cat.ecsv". I don't think the source catalog product file name should contain both "i2d" and "cat" product type suffixes. It should only be "cat".

This is either a problem in calimage3 or the source_catalog step.

larrybradley:
This fixes #1051 **assuming** the input `DrizProductModel` filename always ends in `_i2d.fits`.  Please let me know if that's not the case and I will generalize this.

This PR also includes a small fix to the masked AB mag assignment.

### DIFF
--- a/jwst/source_catalog/source_catalog.py
+++ b/jwst/source_catalog/source_catalog.py
@@ -1,10 +1,33 @@
+import re
+from os.path import split, splitext, join
+
 import numpy as np
 from astropy.convolution import Gaussian2DKernel
 from astropy.stats import sigma_clipped_stats, gaussian_fwhm_to_sigma
 from astropy.table import QTable
 import astropy.units as u
 import photutils
+
 from ..datamodels import DrizProductModel
+
+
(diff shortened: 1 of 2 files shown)


hbushouse:
I can't imagine why, but I suppose it's theoretically possible that someone could try running source_catalog on an image that has NOT been through the resample step and hence the product type suffix wouldn't necessarily always be _i2d. So I guess it's perhaps safer to just replace the last suffix (i.e. anything following the last "_" in the root name), whatever it is, with "_cat".

larrybradley:
Test failures are unrelated.

PR #1065, opened 2017-07-19
071817 a

dmggh:
Revised initial FGS pipeline code -13 files. (Plus an extraneous documentation file from an earlier commit which was unintentionally included, which I will clean up later)


jwst/datamodels/schemas/guider_cds.schema.yaml
```diff
@@ -0,0 +1,132 @@
+allOf:
```
hbushouse: I don't see the "guider_cds.schema.yaml" file used in any of the data models. Only guider_fgs_raw.schema.yaml and guider_fgs_cal.schema.yaml.

jwst/datamodels/schemas/guider_fgs_cal.schema.yaml
```diff
@@ -0,0 +1,132 @@
+allOf:
```
hbushouse: Do we need "fgs" in these names? Why not just guider_cal.schema.yaml and guider_raw.schema.yaml?

jwst/datamodels/schemas/guiderfgs.schema.yaml
```diff
@@ -0,0 +1,131 @@
+allOf:
```
hbushouse: Why do we need this one, when we've got the "raw" and "cal" versions?

jwst/guider_cds/guider_cds.py
```diff
@@ -0,0 +1,130 @@
...
+
+    exp_type: string
+        exposure type
+    """ 
+    instrume = model.meta.instrument.name
+    frame_time = model.meta.exposure.frame_time
+    ngroups = model.meta.exposure.ngroups
+
+    effinttm = model.meta.exposure.integration_time
+    exp_type = model.meta.exposure.exposure_time
```
hbushouse: The right-hand side should be model.meta.exposure.**type** (not exposure_time).

jwst/pipeline/calwebb_guider.py
```diff
@@ -0,0 +1,45 @@
...
+    # start the actual processing
+    def process(self, input):
+
+        log.info('Starting calwebb_guider ...')
+
+        # open the input
+        input = datamodels.GuiderRawModel(input)   # change  to 'input/Raw' type-  071817 02:50
+
+        input = self.dq_init(input)  
+        input = self.flat_field(input)
```
hbushouse: It would be easiest to call flat_field after the guider_cds step, just as we do in normal calwebb_sloper, calimage2 processing flow, so that the flat_field step only has to work on the countrate images, rather than being applied to each of the raw groups. Speaking of which, it'll be interesting to see if flat_field is able to handle the new data models as input, or whether it will also need updating, like dq_init.

nden:
@dmggh This isn't merging into master but a separate newly created branch. Was that the intention?

hbushouse:
Yeah, I was kind of wondering about that too, including the cryptic PR title of "071817 a" as opposed to some English prose describing (briefly) the contents and purpose of the PR. No idea what the answers are. @dmggh ?

dmggh:
The intent was to merge into master. I will name future titles with something more descriptive.

PR #1067, opened 2017-07-19
Fix association reference file prefetch

jaytmiller:
I refactored the stpipe reference file pre-fetching to enable association (ModelContainer) pre-fetches and to better handle model opening, closing, and garbage collection.  Hopefully it minimizes the pipeline prefetch overhead by sticking to a single open model and garbage collecting only at the top level.  (step.py, pipeline.py).  

Refactored Step methods into focused and nested goals (open a model if needed and garbage collect,  handle containers,  handle overrides and result checking, etc).  I also removed some of the "retreat logic" like "if self.skip: return" by using it to avoid calling the method instead.

I removed the seemingly unused get_reference_file_model() method.  list_reference_files(),  which does not handle associations and containers,  should likewise be considered for removal.

Simplified some code in crds_client.py,  a couple pylint tweaks,  and fixed some bad comments.

### DIFF
--- a/jwst/stpipe/pipeline.py
+++ b/jwst/stpipe/pipeline.py
@@ -129,7 +129,7 @@ class Pipeline(Step):
         Precache all of the expected reference files in this Pipeline
         and all of its constituent Steps process method is called.
 
-        input_file:  a filename string, model, or container model.
+        input_file:  a filename string, model, or model container.
 
         """
         from .. import datamodels
--- a/jwst/stpipe/pipeline.py
+++ b/jwst/stpipe/pipeline.py
@@ -131,8 +131,6 @@ class Pipeline(Step):
         """
         from .. import datamodels
         gc.collect()
-        if self._is_association_file(input_file):
-            return
         try:
             with datamodels.open(input_file) as model:
                 pass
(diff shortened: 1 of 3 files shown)


jwst/stpipe/pipeline.py
```diff
@@ -128,23 +128,22 @@ def _precache_reference_files(self, input_file):
         """
         Precache all of the expected reference files in this Pipeline
         and all of its constituent Steps process method is called.
+
+        input_file:  a filename string, model, or container model.
```
jdavies-st: `input_file:  a filename string, model, or model container`

nden:
👍 on removing `list_reference_files`. It was added as a workaround to a problem which was later resolved by having CRDS return `N/A`.
The nirspec tests are failing because the code and reference files are non in sync.

jaytmiller:
Thanks for all the reviews,  I will correct the comment James noticed and drop list_reference_files() in my next pull request.

@nden is it time to update the default CRDS context to the value (349) used last night for regression tests?   Howard mentioned that might be affecting the Travis outcomes and you seem to be referring to that as well.

nden:
@jaytmiller  I'm looking now at failures in the nightly tests - still can't decide whether they are real problems. I need some more time.

jaytmiller:
Is there any reason not to merge this now?

stscieisenhamer:
👍

PR #1068, opened 2017-07-20
Revision of prototype code for the FGS pipeline, including step code …

dmggh:
…and new datamodels code. The 3 calibration steps in this pipeline are the existing dq_init and flat_field, and the new (analog of ramp fitting) guider_cds step.

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -1598,7 +1598,7 @@ properties:
             type: string
             fits_keyword: S_SRCCAT
           guider_cds:
-            title: Guider CDS Correction
+            title: Guider CDS Calculation
             type: string
             fits_keyword: S_GUICDS
       background:
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -1597,6 +1597,10 @@ properties:
             title: Source Photometry and Morphology
             type: string
             fits_keyword: S_SRCCAT
+          guider_cds:
+            title: Guider CDS Correction
+            type: string
+            fits_keyword: S_GUICDS
       background:
         title: Background information
         type: object
--- a/jwst/guider_cds/guider_cds.py
+++ b/jwst/guider_cds/guider_cds.py
@@ -1,0 +1,130 @@
+#! /usr/bin/env python
+# 
+#  guider_cds.py
+
+from __future__ import division
+import time
+import numpy as np
+import logging
+
+from .. import datamodels
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+def guider_cds(model):
+    """
+    Extended Summary
+    ----------------
+    Calculate the count rate for each pixel in all data cube sections and all
+    integrations.
+
+    For each integration in a given FGS input dataset whose mode is ACQ1, 
+    ACQ2, or TRACK, the count rate is the last group minus the first group, 
+    divided by the effective integration time.  If the mode is ID, the last 
+    group minus the first group is calculated for both integrations; the count 
+    rate is then given by the minimum of these two values for each pixel, 
+    divided by the effective integration time.  For the FINEGUIDE mode, the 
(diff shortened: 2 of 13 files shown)


dmggh:
I just noticed I inadvertently removed a carriage return in jwst/datamodels/__init__.py

dmggh:
I just noticed I inadvertently removed a carriage return in datamodels/__init__.py

dmggh:
Also a couple types in datamodels/guiderraw.py and guidercal.py - missing colons.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1597,6 +1597,10 @@ properties:
             title: Source Photometry and Morphology
             type: string
             fits_keyword: S_SRCCAT
+          guider_cds:
+            title: Guider CDS Correction
```
hbushouse: Minor nit, but I would call this "Guider CDS **Calculation**", as opposed to Correction, because it isn't really correcting for anything, just computing differences.

jwst/guider_cds/guider_cds.py
```diff
@@ -0,0 +1,130 @@
...
+
+    exp_type: string
+        exposure type
+    """ 
+    instrume = model.meta.instrument.name
+    frame_time = model.meta.exposure.frame_time
+    ngroups = model.meta.exposure.ngroups
+
+    effinttm = model.meta.exposure.integration_time
+    exp_type = model.meta.exposure.exposure.type
```
hbushouse: Just noticed when trying to run this step that there's an error in this line: too many instances of "exposure". Should be "model.meta.exposure.type"

jwst/guider_cds/guider_cds.py
```diff
@@ -0,0 +1,130 @@
...
+    nreads = model.data.shape[1]
+    asize2 = model.data.shape[2]
+    asize1 = model.data.shape[3]
+
+    npix = asize2 * asize1  # number of pixels in 2D array
+    imshape = (asize2, asize1)
+
+    log.info('Instrume: %s' % (instrume))
+    log.info('Number of integrations: %d' % (n_int))
+    log.info('Number of reads: %d' % (nreads))
```
hbushouse: nreads, being derived from model.data.shape[1], is the same as (or at least should always be the same as) ngroups. So we probably don't need to report both of them. Plus the fact that "nreads" could be misinterpreted as referring to the number of times a pixel is sampled when doing each readout.

jwst/guider_cds/guider_cds.py
```diff
@@ -0,0 +1,130 @@
...
+            if num_int == 1:
+                diff_int1 = grp_last - grp_first
+
+        else:  # ACQ1, ACQ2, or TRACK
+            grp_last = data_sect[1, :, :]
+            grp_first = data_sect[0, :, :]
+            slope_int_cube[ num_int,:,: ] = grp_last - grp_first
+        
+    if exp_type[:6] == 'FGS_ID':
+        new_model.data[0,:,:] = np.minimum( diff_int1, diff_int0 )/effinttm
```
hbushouse: It appears that effinttm is the total duration of all groups within an integration, which in the case of 2 groups means that effinttm = 2 * group_time. But because we're taking the difference of grp_last - grp_first, that means the real amount of signal collection time represented by this difference image is just the time between groups 1 and 2, which is the time of one group. It's not the total time since the start of the exposure. So we should be dividing by the group_time here (and in line 74), rather than the effinttm.

jwst/guider_cds/guider_cds.py
```diff
@@ -0,0 +1,130 @@
...
+    asize2 = model.data.shape[2]
+    asize1 = model.data.shape[3]
+
+    npix = asize2 * asize1  # number of pixels in 2D array
+    imshape = (asize2, asize1)
+
+    log.info('Instrume: %s' % (instrume))
+    log.info('Number of integrations: %d' % (n_int))
+    log.info('Number of reads: %d' % (nreads))
+    log.info('Frame time: %d' % (frame_time))
```
hbushouse: Frame time should be printed as a float, not an integer.

PR #1069, opened 2017-07-20
Add NOUTPUTS keyword to core schema

hbushouse:
NIRCam is adding a new keyword that indicates the number of detector outputs used when reading out an exposure. The default is 4 outputs, but for some subarrays the user can choose to use just 1 amplifier output.

Implementing as documented in [JWSTKD-28](https://jira.stsci.edu/browse/JWSTKD-28)


hbushouse:
All CI failures are unrelated. Merging.

PR #1070, opened 2017-07-20
Data only schema

nden:
This PR refactors a bit the base DataModel to allow writing out files containing only data and no metadata. The changes to DataModel are simple - check if `meta` exists before trying to access it.

Several new schemas are introduced:
- The `pathlosscorr` schema includes the data extensions that the pathloss correction step writes out.
- The `slitdata` schema includes only the data extensions and other attributes (`xstart`, `ystart`, etc) typical for the items in `MultiSlitModel.slits`.
- The `slit` schema is a union of the `slitdata` schema and the `core` schema.
- The `ifuimage` schema combines the modified `image` schema with the new `pathlosscorr`
  schema.

Changes to existing schemas:
- The pathloss correction extensions are removed from the `image` schema. 
  **Note** that for this not to break existing functionality there needs to be an IFUModel and schema to deal with pathloss correction in IFU data.
- The `multislit` schema now includes the `core` schema and an array of `slitdata` refs.

Other changes:
- `ramp_fitting` now returns an `IFUImageModel` as first output in case EXT_TYPE is either `NRS_IFU` or `MIR_MRS`.
- steps which use `ImageModel` for IFU data were changed to use `IFUImageModel` instead.
   The affected steps are: ramp_fitting, cube_build, straylight, flat_field, extract_1d,
    fringe, photom

This is done as a first step in refactoring `extract_2d` to improve pereformance.  The speedup is enormous. The second step would be to implement non-coordinate inputs in `gwcs` so that the `MultiSlitModel` can keep only one copy of the WCS object.


jwst/datamodels/schemas/ifuimage.schema.yaml
```diff
@@ -0,0 +1,41 @@
...
+      datatype: uint32
+    err:
+      title: Error array
+      fits_hdu: ERR
+      default: 0.0
+      datatype: float32
+    zeroframe:
+      title: Zeroframe array
+      fits_hdu: ZEROFRAME
+      default: 0.0
```
hbushouse: I wonder if we should remove the default values for all arrays that are optional or don't appear in all datasets, so that the extension is not written out to the product when it's not needed for a given product. This would include zeroframe, area, relsens2d, and all of the ones in pathloss.schema.

nden: Agreed and done for som eof them. The pathloss correction schema does not have defaults and yet these are the ones that show up. I wonder if it's a bug in models for 1D arrays.

hbushouse:
I just added a regression test on jwcalibdev that runs calspec2 on a NIRSpec IFU exposure, which results in pathloss extensions in the _cal product. So we should let this run at least once, to verify that the current code is stable, before we attempt to switch any calspec2 steps over to using the new IFUImageModel.

nden:
Coming back to this PR after so many months and finding, of course, it doesn't work any more.

@hbushouse Can we assume that a `MultiSlit` model contains slits coming from a single observation and so that all slits have the same `meta` data, except for the wcs object? Or is this model also used for storing slits from multiple observations and so can have different meta data?

jwst/datamodels/ifuimage.py
```diff
@@ -0,0 +1,77 @@
+from __future__ import absolute_import, unicode_literals, division, print_function
+
+from . import model_base
+
+
+__all__ = ['ImageModel']
```
hbushouse: Shouldn't this now be 'IFUImageModel'?

hbushouse:
I think that assumption about a `MultiSlit` model is valid or at least workable. In calspec3, when exp_to_source is called to rearrange and collect data for a given source from multiple `MultiSlit` models, it uses the `MultiExposure` model as the output that contains data from multiple exposures. But internally the pipeline is still using data from the `MultiSlit` model that's simply been repackaged into a `ModelContainer` (hence a list of `MultiSlit` models).

Would need @stscieisenhamer to confirm.

dmggh:
As I understand the effect of this update on ramp fitting, it would apply to only the primary output (which is an ImageModel) of ramp fitting, as it will be needed later in the pipeline.  For the other ramp fitting output products, it appears that the additional extensions are already in the integration-specific CubeModel output (which seems reasonable), but not in the optional RampFitOutputModel output (which also seems reasonable, unless there are segment-specific aspects of the new extensions).

jwst/datamodels/ifuimage.py
```diff
@@ -0,0 +1,93 @@
...
+        if data is not None:
+            self.data = data
+
+        if dq is not None:
+            self.dq = dq
+
+        if err is not None:
+            self.err = err
+
+        if relsens is not None:
```
hbushouse: A check of the photom step indicates that the RELSENS extension is NOT populated for either NIRSpec IFU or MIRI MRS exposures. They only use the RELSENS2D extension. Hence RELSENS should not be needed in IFUImageModel.

jwst/datamodels/schemas/ifuimage.schema.yaml
```diff
@@ -0,0 +1,51 @@
...
+      datatype: float32
+    zeroframe:
+      title: Zeroframe array
+      fits_hdu: ZEROFRAME
+      datatype: float32
+    area:
+      title: Pixel area map array
+      fits_hdu: AREA
+      datatype: float32
+    relsens:
```
hbushouse: As mentioned above, RELSENS can be removed from IFUImageModel.

jwst/datamodels/schemas/image.schema.yaml
```diff
@@ -37,46 +37,18 @@ allOf:
       fits_hdu: RELSENS2D
       default: 1.0
       datatype: float32
-    wavelength:
```
hbushouse: You're sure we don't ever need the wavelength array with ImageModel? I'm thinking of simple spectroscopic cases like NIRISS SOSS and MIRI LRS, which use ImageModel. Do we only ever populate a WAVELENGTH array for NIRSpec fixed-slit and MSA data? If so, then I guess we don't need wavelength in ImageModel, because the NIRSpec cases use MultiSlitModel.

jwst/datamodels/schemas/slitdata.schema.yaml
```diff
@@ -0,0 +1,141 @@
...
+      fits_keyword: SRCXPOS
+      fits_hdu: SCI
+    source_ypos:
+      title: Source position in slit (y-axis)
+      type: number
+      default: 0.0
+      fits_keyword: SRCYPOS
+      fits_hdu: SCI
+    shutter_state:
+      title: All (open and close) shutters in a slit
```
hbushouse: Change "close" to "closed"

jwst/datamodels/schemas/slitdata.schema.yaml
```diff
@@ -0,0 +1,141 @@
...
+      fits_hdu: WAVELENGTH
+      ndim: 2
+      datatype: float32
+    barshadow:
+      title: Bar shadow correction
+      fits_hdu: BARSHADOW
+      ndim: 2
+      default: 1.0
+      datatype: float32
+    bunit_data:
```
hbushouse: Could you just import bunit.schema.yaml here instead of defining them again?

jwst/datamodels/slit.py
```diff
@@ -0,0 +1,163 @@
...
+    dq : numpy array
+        The data quality array.
+
+    err : numpy array
+        The error array.
+
+    relsens : numpy array
+        The relative sensitivity table.
+
+    relsens2d: numpy array
```
hbushouse: relsens2d is not included (not needed) in SlitModel, so remove it from the list here.

jwst/datamodels/slit.py
```diff
@@ -0,0 +1,163 @@
...
+
+        if data is not None:
+            self.data = data
+
+        if dq is not None:
+            self.dq = dq
+
+        if err is not None:
+            self.err = err
+
```
hbushouse: What about area and relsens? Do they need to be in this list?

jwst/datamodels/slit.py
```diff
@@ -0,0 +1,163 @@
...
+    dq : numpy array
+        The data quality array.
+
+    err : numpy array
+        The error array.
+
+    relsens : numpy array
+        The relative sensitivity table.
+
+    relsens2d: numpy array
```
hbushouse: Remove relsens2d

jwst/extract_1d/extract.py
```diff
@@ -1257,8 +1257,9 @@ def do_extract1d(input_model, refname, smoothing_length, bkg_order):
             # For this case, we'll call get_spectral_order to get the order.
             spectral_order_list = ["not set yet"]
 
-        if isinstance(input_model, datamodels.ImageModel) or \
-           isinstance(input_model, datamodels.DrizProductModel):
+        if isinstance(input_model, (datamodels.ImageModel,
+                                    datamodels.DrizProductModel,
+                                    datamodels.IFUImageModel)):
```
hbushouse: I don't think extract_1d should ever be fed an IFUImageModel as input. IFUImageModel is a simple 2D image of the individual slices. Extract_1d can only work on an IFUCubeModel.

jwst/extract_1d/extract_1d_step.py
```diff
@@ -31,6 +31,9 @@ def process(self, input):
         elif isinstance(input_model, datamodels.ImageModel):
             # It's a single 2-D image
             self.log.debug('Input is an ImageModel')
+        elif isinstance(input_model, datamodels.IFUImageModel):
```
hbushouse: Not needed, as mentioned above.

jwst/flatfield/flat_field_step.py
```diff
@@ -46,6 +46,8 @@ def process(self, input):
             self.log.debug('Input is a CubeModel')
         elif isinstance(input_model, datamodels.ImageModel):
             self.log.debug('Input is an ImageModel')
+        elif isinstance(input_model, datamodels.IFUImageModel):
```
hbushouse: There must be a simpler, more elegant way to do this whole if-else block. Rather than testing for individual instances of a particular datamodel type, there must be a way to just print what the data model type is from some attribute value somewhere within the datamodel.

jwst/fringe/fringe_step.py
```diff
@@ -13,7 +13,7 @@ class FringeStep(Step):
     reference_file_types = ['fringe']
 
     def process(self, input):
-        with datamodels.ImageModel(input) as input_model:
+        with datamodels.open(input) as input_model:
```
hbushouse: Theoretically the fringe step can only ever be applied to an input in the form of an IFUImageModel, so should we hardwire the open to use that particular data model, e.g. datamodels.IFUImageModel? I know that the generic .open() function should automatically use the model type set in DATAMODL, but what if someone feeds in a generic, simulated image or something that doesn't have the DATAMODL keyword. We want to know right up front if the input isn't compatible with an IFUImageModel.

jwst/photom/photom_step.py
```diff
@@ -26,6 +26,8 @@ def process(self, input_file):
             self.log.debug('Input is a CubeModel for a multiple integ file.')
         elif isinstance(dm, datamodels.ImageModel):  # standard product: 2D array
             self.log.debug('Input is an ImageModel.')
+        elif isinstance(dm, datamodels.IFUImageModel):  # standard product: 2D array
```
hbushouse: Same comment as earlier about trying to find a simpler way to just print out the datamodel type.

jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -89,5 +89,7 @@ def process(self, input):
         if int_model is not None:
             int_model.meta.cal_step.ramp_fit = 'COMPLETE'
 
-        return out_model, int_model
+        if input_model.meta.exposure.type in ('NRS_IFU', 'MIR_MRS'):
+            out_model = datamodels.IFUImageModel(out_model)
```
hbushouse: If out_model was originally created within ramp_fit as ImageModel, does this just magically convert it to an IFUImageModel? Cool (didn't know you could do that).

jwst/straylight/straylight_step.py
```diff
@@ -14,7 +14,7 @@ class StraylightStep (Step):
     def process(self, input):
 
         # Open the input data model
-        with datamodels.ImageModel(input) as input_model:
+        with datamodels.IFUImageModel(input) as input_model:
```
hbushouse: See here we're specifically loading into an IFUImageModel, as opposed to using the generic datamodels.open(). I think we should do the same in fringe_step.py

hbushouse:
Top comment says:
```
steps which use ImageModel for IFU data were changed to use IFUImageModel instead.
The affected steps are: ramp_fitting, cube_build, straylight, flat_field, extract_2d, extract_1d,
fringe, photom
```
extract_2d was in fact changed, but not for changing ImageModel to IFUImageModel. IFU data don't go through extract_2d. And IFUImageModel also doesn't go through extract_1d. Only IFUCubeModel goes through extract_1d.

jwst/datamodels/schemas/image.schema.yaml
nden: Yes, as of now,  we only populate it for NIRSPEC spectroscopic data.

jwst/datamodels/slit.py
nden: Well, these are created by the photom step but it doesn't hurt to add them here.

jwst/ramp_fitting/ramp_fit_step.py
nden: Yes, it does. You can't do the opposite though.

jwst/straylight/straylight_step.py
nden: 👍

nden:
@hbushouse Thanks for the review! All comments were addressed - code and issue description.

jwst/datamodels/schemas/slitdata.schema.yaml
```diff
@@ -0,0 +1,137 @@
...
+      fits_hdu: WAVELENGTH
+      ndim: 2
+      datatype: float32
+    barshadow:
+      title: Bar shadow correction
+      fits_hdu: BARSHADOW
+      ndim: 2
+      default: 1.0
+      datatype: float32
+    bunit_err:
```
hbushouse: You forgot to remove bunit_err, which is now already included in bunit.schema.yaml

jwst/flatfield/flat_field_step.py
```diff
@@ -41,15 +41,7 @@ def process(self, input):
...
-        if isinstance(input_model, datamodels.CubeModel):
-            # multi-integration dataset
-            self.log.debug('Input is a CubeModel')
-        elif isinstance(input_model, datamodels.ImageModel):
-            self.log.debug('Input is an ImageModel')
-        elif isinstance(input_model, datamodels.MultiSlitModel):
-            self.log.debug('Input is a MultiSlitModel')
-        elif isinstance(input_model, datamodels.GuiderCalModel):
-            self.log.debug('Input is a GuiderCalModel')
+        self.log.debug("Input is {}".format(input_model.__class__.__name__))
```
hbushouse: 👍

jwst/datamodels/schemas/slitdata.schema.yaml
nden: I didn't realize it was included in `bunit`. Nice.

nden:
I am merging this PR in order to deal with possible test failures tomorrow.

Issue #589, opened 2017-01-12
set_telescope_pointing.py as a module


jdavies-st:
It would be nice to have the functions in

jwst/scripts/set_telescope_pointing.py

available as a module so that they could be imported into other python code.  Currently they are just available in the set_telescope_pointing.py script for use by DMS.

Currently, Bryan Hilbert in the NIRCam instrument team is creating simulated NIRCAM data and is using the computations in this script to set the meta.wcsinfo block to the correct values given input target and SIAF data provided by the simulator.  I suspect other instrument teams may want this functionality too.  The current workaround for importing the functions from the installed script in the astroconda distribution is ugly.

My suggestion would be to make a jwst.pointing module with all the functions contained within and just have the set_telescope_pointing.py script import this module.

@perrygreenfield @stscieisenhamer what do you think?

stscieisenhamer:
👍 absolutely

PR #1071, opened 2017-07-20
Update set_telescope_pointing to log more info + workflow diagram updates

stscieisenhamer:
ya, two completely separate things got mashed together here.


stscieisenhamer:
Resolves #589

PR #1073, opened 2017-07-20
fix nirspec with new ref files

nden:
fix problems discovered by nightly tests [skip ci]


nden:
There's no sense in running Travis tests for this.

Issue #991, opened 2017-06-21
Create unique product type names for different flavors of source/photometry catalogs.


hbushouse:
The various cal pipeline modules currently produce a different flavors of source and/or photometry catalogs, which are stored in ecsv format. The calimage3 source_catalog step produces a source catalog from the combined science image. The caltso3 pipeline will produce a photometry catalog for imaging TSO modes and a "white light" photometry catalog for spectroscopic TSO modes. Each of these different types of catalogs can contain different sets of meta data in their headers (in addition to the portion of the header that simply defines the data columns).

Instrument scientists would like to have the meta data from these catalogs stored in the archive catalog, so that they can be used in searches. In order for the archive catalog ingest to do this for the different sets of meta data that will appear in each of the different catalog types, they will need to have each different catalog type identified using a unique product type file name suffix, so that there's a one-to-one correspondence between the product type suffix and the list of meta data values that need to be ingested into the archive catalog.

Right now all flavors of these catalogs use the "_cat" suffix in their file names. We need to define and implement unique suffixes for each of the 3 different types of catalogs.

hbushouse:
How about the following for new product type names:

1) Source_catalog step output remains _cat (it is a catalog after all)
2) tso_photometry step output becomes _phot (it's just photometry; no catalog info)
3) white_light step output becomes _whtlt (there's no limit to length, but don't want folks to have to type out long strings like "_whitelight")

Thoughts/comments?: @stscieisenhamer @larrybradley @hack

hbushouse:
Hello? (tap, tap, tap) Is this thing on?  Any other comments yay or nay on the proposed new product type names? Otherwise, hearing no objections, I will make it so.

larrybradley:
I updated `tso_photometry` here: https://github.com/STScI-JWST/jwst/pull/1076.

PR #1076, opened 2017-07-21
Rename tso_photometry product name to '_phot'

larrybradley:
Partially addresses #991.

### DIFF
--- a/jwst/tso_photometry/tso_photometry_step.py
+++ b/jwst/tso_photometry/tso_photometry_step.py
@@ -2,6 +2,7 @@
 
 from ..stpipe import Step, cmdline
 from ..datamodels import CubeModel
+from ..source_catalog.source_catalog import _replace_suffix_ext
 from .tso_photometry import tso_aperture_photometry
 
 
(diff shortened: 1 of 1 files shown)


larrybradley:
@hbushouse This step is also currently *appending* `_phot` to the existing name.  Should `_phot` replace the previous product name, and if so, what is that name?

hbushouse:
The input to the calwebb_tso3 pipeline will be a _calints.fits product, which is a 3D cube model. The first step of the pipeline will be outlier_detection, which will result in an updated version of the input model that has additional DQ flags set. This updated model will be saved as a new product that will use a _crfints (CR-Flagged) product type extension. But if the pipeline flow is constructed in such a way that the updated model is simply passed (in memory) to the tso_photometry step, then the meta data for that model would still contain a filename attribute that uses the _calints suffix. So I would assume _calints for now and if we need to we can always update to _crfints. And yes, the suffix should be _replaced_, rather than just appending _phot.

larrybradley:
In that case I'll make `_phot` replace both `_calints` and `_crfints` product names.

hbushouse:
👍

PR #1087, opened 2017-07-26
Save results from each association product as produced

stscieisenhamer:
Previously, the pipeline retained results from each product in an
association, saving all only at the end of processing. This was both
memory-inefficient and, if the process crashed, would cause loss of
results.


stscieisenhamer:
Failures unrelated. Merging.

PR #1090, opened 2017-07-31
Fix bugs in resample_spec

jdavies-st:
- take out units in `astropy.modeling.models` instances
- compensate for `SourceModelContainer` replacing `ModelContainer`. I.e. when iterating through `SourceModelContainer`, one does not get datamodel instances, instead one gets nodes.

### DIFF
--- a/jwst/pipeline/calwebb_image2.py
+++ b/jwst/pipeline/calwebb_image2.py
@@ -17,20 +17,19 @@ __version__ = "3.0"
 
 class Image2Pipeline(Pipeline):
     """
-
-    CalWebbImage2: Processes JWST imaging-mode slope images from
-                   Level-2a to Level-2b.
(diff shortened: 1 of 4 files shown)


nden:
@jdavies-st Note that models should be working with units now, if you need them. The bug that caused the reg test to fail was fixed in astropy. We just haven't had a current astropy/master installed in the nightly tests.

jdavies-st:
Thanks Nadia. Yeah, my problem was that I was using units in the code already for some of the models in the compound model transforms, but not consistently.  The old behavior of `astropy.modeling.models` was just using the values from the coeffs and ignore the units.

Now that `units` are implemented in `'astropy.modeling.models` in astropy 2.0, the checking is much stricter.  Units must propagate.

The fix here just gets rid of all units in my compound models.

PR #1031, opened 2017-07-09
[WIP] refactor extract_2d 

nden:
This PR creates a `slit` schema and model. This new schema references the image schema and adds slit specific extensions - `wavelength` and all the pathloss extension. I am assuming the pathloss extensions were added to the `image` schema simply because there was no other specific schema to use (@stscirij?).  This PR removes them from the `image` schema but if they need to be I can add them back.

Also now extract_2d writes out a `wavelength` extension for use by other steps. The wavelength zero-point correction will be added in a separate PR. The main point of this refactoring is to improve performance. 
Every slit is created and validated separately  and then the list of slits is assigned to MultiSlitModel onnce.

todo:
[ ] documentation
[ ] testing

PR #1091, opened 2017-07-31
Nirspec wavelength zero-point correction

nden:
First cut at nirspec zero-point correction. As agreed on it is part of the extract_2d step. `ImageModel` and its schema were modified to include a `wavelength` array. This will eventually be changed after #1070 and #1031 or variants of these are merged.

Currently this only works on MSA data. Still need to 

[x ] add code for fixed slit  
[x ] make sure this runs only on point source
[ ] deliver reference files
[x ] docs


hbushouse:
Work done in support of #601

nden:
Travis on STScI-JWST works at snail speed. Linux tests are passing here and Mac tests passed on my private Travis. Merging to make space for the other PRs.

PR #1094, opened 2017-08-02
Fix assign_wcs test_nirspec_imaging

jdavies-st:
Fixes a small bug in `assign_wcs/nirspec.py` and updates `DATE-OBS` in the generated data so that the tests use the latest CV3 NIRSpec reffiles.

### DIFF
--- a/jwst/assign_wcs/nirspec.py
+++ b/jwst/assign_wcs/nirspec.py
@@ -72,7 +72,7 @@ def imaging(input_model, reference_files):
     dircos2unitless = DirCos2Unitless(name='directional_cosines2unitless')
 
     col_model = CollimatorModel(reference_files['collimator'])
-    col = col.model
+    col = col_model.model
     col_model.close()
 
     # Get the default spectral order and wavelength range and record them in the model.
--- a/jwst/assign_wcs/tests/test_nirspec.py
+++ b/jwst/assign_wcs/tests/test_nirspec.py
@@ -57,7 +57,7 @@ def create_hdul():
     phdu.header['instrume'] = 'NIRSPEC'
     phdu.header['detector'] = 'NRS1'
     phdu.header['time-obs'] = '8:59:37'
-    phdu.header['date-obs'] = '2014-09-05'
+    phdu.header['date-obs'] = '2016-09-05'
 
     for item in wcs_kw.items():
         phdu.header[item[0]] = item[1]


jwst/assign_wcs/nirspec.py
```diff
@@ -72,7 +72,7 @@ def imaging(input_model, reference_files):
     dircos2unitless = DirCos2Unitless(name='directional_cosines2unitless')
 
     col_model = CollimatorModel(reference_files['collimator'])
-    col = col.model
+    col = col_model.model
```
nden: Good catch @jdavies-st ! May be this will fix other problems I see with CV3 data.

jwst/assign_wcs/tests/test_nirspec.py
```diff
@@ -57,7 +57,7 @@ def create_hdul():
     phdu.header['instrume'] = 'NIRSPEC'
     phdu.header['detector'] = 'NRS1'
     phdu.header['time-obs'] = '8:59:37'
-    phdu.header['date-obs'] = '2014-09-05'
+    phdu.header['date-obs'] = '2016-09-05'
```
nden: What is the current default context?

jdavies-st: @jaytmiller are we at `CRDS_CONTEXT ≥ 349` now?

jdavies-st: Regardless, I think the `DATE-OBS` here should be later, as even when pmaps ≥ 349 become default, the earlier files will be selected, as the later newer contexts will have USEAFTER dates for the CV3 data, i.e. ~Nov 2015.

jwst/assign_wcs/nirspec.py
jdavies-st: I don't think it will fix the remaining 3 tests in `assign_wcs` that are failing, but perhaps those tests need to be re-written, as they currently test against Build 6 versions of the reffiles.  From what I can see, `test_correct_tilt`, `test_nirspec_fs_esa` and `test_nirspec_ifu_against_esa` seem to compare to old, non-CV3 reffiles?

Should these tests be disabled or is it possible to rewrite them?  Or perhaps disabled temporarily?

jwst/assign_wcs/tests/test_nirspec.py
hbushouse: But isn't 2016-09-05 already later than ~Nov. 2015 and hence will always get CV3-era ref files selected, because the ref files have USEAFTER < 2016.

jdavies-st: Yes, that's the hope!  Right now the `assign_wcs` code these tests are testing only works on CV3 files.  Therefore we need to make these tests use CV3 reffiles, and thus have an appropriate USEAFTER date so CRDS selects them.

nden: I don't expect these tests to pass because the data used for comparison is pre-CV3. As soon as we see that they fail and don't error (using CV3 reference files) they can be x-failed. The comparison data will be updated later in a separate PR.
The `correct_tilt` test failure indicates pre-CV3 reference files are used.

nden: Actually the test `correct_tilt` has a dict as a `disperser` so I take back my comments above. Apparently CV3 ref files are used. 
@jdavies-st Could you disable now the nirspec tests and merge this PR. Thanks for the work!

PR #1095, opened 2017-08-02
Updates to level 2 pipelines an use of output_file

stscieisenhamer:
Re-enabled use of `output_file` when specifying individual files to the Level 2 pipelines.


stscieisenhamer:
unrelated failures, merging

PR #1097, opened 2017-08-03
 Updated dq_init code to handle FGS modes for the FGS pipeline. For d…

dmggh:
…atasets having EXP_TYPE beginning with FGS, input GuiderRaw model has a DQ array, unlike this step's default Ramp Model having PIXELDQ and GROUPDQ extnesions.

### DIFF
--- a/jwst/dq_init/dq_init_step.py
+++ b/jwst/dq_init/dq_init_step.py
@@ -4,6 +4,7 @@ from ..stpipe import Step
 from .. import datamodels
 from . import dq_initialization
 
+
 class DQInitStep(Step):
     """
 
@@ -14,11 +15,13 @@ class DQInitStep(Step):
 
     reference_file_types = ['mask']
 
-    def process(self, input): 
-
+    def process(self, input):
+        # Determine Model type from EXP_TYPE of RampModel
         input_model = datamodels.open( input )
-        exp_type_pref = input_model.meta.exposure.type[:3]
-        if exp_type_pref == 'FGS':  # reopen as a GuiderRawModel 
+
+        # Set flag for guider mode operations
+        if input_model.meta.exposure.type in dq_initialization.guider_list: # Reopen as GuiderRawModel
+            input_model.close()
             input_model = datamodels.GuiderRawModel( input )
 
         # Check for consistency between keyword values and data shape
@@ -57,7 +60,7 @@ class DQInitStep(Step):
         mask_model = datamodels.MaskModel(self.mask_filename)
 
         # Apply the step
-        result = dq_initialization.correct_model(input_model, mask_model, exp_type_pref)
+        result = dq_initialization.correct_model(input_model, mask_model)
 
         # Close the reference file
         mask_model.close()
--- a/jwst/dq_init/dq_init_step.py
+++ b/jwst/dq_init/dq_init_step.py
@@ -14,48 +14,52 @@ class DQInitStep(Step):
 
     reference_file_types = ['mask']
 
-    def process(self, input):
+    def process(self, input): 
 
-        with datamodels.open(input) as input_model:
+        input_model = datamodels.open( input )
+        exp_type_pref = input_model.meta.exposure.type[:3]
+        if exp_type_pref == 'FGS':  # reopen as a GuiderRawModel 
+            input_model = datamodels.GuiderRawModel( input )
 
-            # Check for consistency between keyword values and data shape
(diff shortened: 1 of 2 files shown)


jwst/dq_init/dq_init_step.py
```diff
@@ -14,48 +14,52 @@ class DQInitStep(Step):
...
-            # Apply the step
-            result = dq_initialization.correct_model(input_model, mask_model)
-
-            # Close the reference file
-            mask_model.close()
+    def process(self, input): 
+
+        input_model = datamodels.open( input )
+        exp_type_pref = input_model.meta.exposure.type[:3]
+        if exp_type_pref == 'FGS':  # reopen as a GuiderRawModel
```
hbushouse: This logic will need to be expanded, because the normal (science) modes of FGS also start with the 3 letters "FGS", such as FGS_IMAGE, FGS_DARK, and FGS_FOCUS. The only ones that are exclusive to guiding operations are FGS_ID-IMAGE, FGS_ID-STACK, FGS_ACQ1, FGS_ACQ2, FGS_TRACK, and FGS_FINEGUIDE. Those values could be put into a list and then just test to see if exp_type_pref has a value in that list.

jwst/dq_init/dq_init_step.py
```diff
@@ -14,48 +14,52 @@ class DQInitStep(Step):
...
-            result = dq_initialization.correct_model(input_model, mask_model)
-
-            # Close the reference file
-            mask_model.close()
+    def process(self, input): 
+
+        input_model = datamodels.open( input )
+        exp_type_pref = input_model.meta.exposure.type[:3]
+        if exp_type_pref == 'FGS':  # reopen as a GuiderRawModel 
+            input_model = datamodels.GuiderRawModel( input )
```
hbushouse: It might be good to first do a "input_model.close()" on the existing instance before reopening as a new model.

jwst/dq_init/dq_initialization.py
```diff
@@ -15,52 +15,65 @@
...
     """DQ Initialize a JWST Model"""
-
-    output_model = do_dqinit(input_model, mask_model)
+    output_model = do_dqinit(input_model, mask_model, exp_type_pref)
 
     return output_model
 
-def do_dqinit(input_model, mask_model):
+
+def do_dqinit(input_model, mask_model, exp_type_pref):
```
hbushouse: Because we can't do a simple check on the first 3 characters of exp_type_pref, I wonder if it would be just as simple to not bother sending that value down to all lower-level routines to use as a check and instead just have all of the checks based on the type of the input_model, e.g. if isinstance(input_model, datamodels.GuiderRawModel). That way we don't have add the additional argument to all of the function calls.

jwst/dq_init/dq_initialization.py
```diff
@@ -15,52 +15,65 @@
...
-        mask_array = get_mask_subarray(mask_model, output_model)
+        mask_array = get_mask_subarray(mask_model, output_model, exp_type_pref)
     else:
         mask_array = mask_model.dq
-    #
-    # Bitwise-OR the data pixeldq with the reference file mask
-    dq = np.bitwise_or(input_model.pixeldq, mask_array)
-    output_model.pixeldq = dq
+
+    if exp_type_pref == 'FGS':
```
hbushouse: Need to fix this logic (as above)

jwst/dq_init/dq_init_step.py
philhodge: > ... and then just test to see if exp_type_pref has a value in that list.

Not `exp_type_pref`; instead, compare input_model.meta.exposure.type itself with that list.

dmggh:
It was also my initial idea to just check the type of the input_model. However, datamodels.open( input ) currently always opens as a type of RampModel, so checking the type can not be used to check if the input is a GuiderRawModel. The keyword EXP_TYPE should always be populated, so its value from an opened RampModel can be used to determine if the dataset should be reopened as a GuiderRawModel.

hbushouse:
I agree that at least on the initial datamodels.open() call you can't trust it to use the right data model, but once you've done the check and reopened as GuiderRawModel, then it'll stay that way throughout the rest of the lower-level routines and hence should be safe to check the model instance. But the way you've got it now is also pretty much just as efficient, since you're checking against the model meta directly, rather than passing down the exp_type to all the other routines. So that's fine.

PR #1101, opened 2017-08-08
update Image2Pipeline to handle integrations output

stscieisenhamer:


jwst/pipeline/calwebb_image2.py
```diff
@@ -51,8 +50,12 @@ def process(self, input):
 
             # Save result
             suffix = 'cal'
+            if isinstance(input, datamodels.CubeModel):
```
hbushouse: This logic isn't working, because `input` is still just a file name at this point - it hasn't been loaded into a data model. If you change this check to use `result` it works fine. calwebb_spec2 may have the same problem.

stscieisenhamer: good catch. this bug is in spec2 also. No regression for ints?  will have to make one.

hbushouse: Yes, there is a reg test for ints (test_image2pipeline_2b). That's how I caught it. It was the 1 failing test for the past couple days.

PR #1102, opened 2017-08-08
Document dark_output config parameter

### LINKED ISSUE #1017: dark_output is not documented in the spec
When running:
```
$ strun dark_current.cfg -h
```
The line about dark_output is blank.

Should also add logging out when dark_output is requested but no averaging occurs, so there is nothing to put out.

stscieisenhamer:
Changes:
- Document flag
- If specified, will write out either the averaged dark or
  the dark model from the reference file.

Resolves #1017

### DIFF
--- a/jwst/dark_current/dark_current_step.py
+++ b/jwst/dark_current/dark_current_step.py
@@ -10,7 +10,7 @@ class DarkCurrentStep(Step):
     """
 
     spec = """
-        dark_output = output_file(default = None)
+        dark_output = output_file(default = None) # Dark model or averaged dark subtracted
     """
 
     reference_file_types = ['dark']
@@ -40,8 +40,9 @@ class DarkCurrentStep(Step):
                 dark_model = datamodels.DarkModel(self.dark_name)
 
             # Do the dark correction
-            result = dark_sub.do_correction(input_model, dark_model,
-                                                    self.dark_output)
+            result = dark_sub.do_correction(
+                input_model, dark_model, self.dark_output
+            )
             dark_model.close()
 
 
--- a/jwst/dark_current/dark_sub.py
+++ b/jwst/dark_current/dark_sub.py
@@ -52,18 +52,24 @@ def do_correction(input_model, dark_model, dark_output=None):
     drk_nframes = dark_model.meta.exposure.nframes
     drk_groupgap = dark_model.meta.exposure.groupgap
 
-    log.info('Science data nints=%d, ngroups=%d, nframes=%d, groupgap=%d',
-              sci_nints, sci_ngroups, sci_nframes, sci_groupgap)
-    log.info('Dark data nints=%d, ngroups=%d, nframes=%d, groupgap=%d',
-              drk_nints, drk_ngroups, drk_nframes, drk_groupgap)
+    log.info(
+        'Science data nints=%d, ngroups=%d, nframes=%d, groupgap=%d',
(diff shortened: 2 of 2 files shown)


jwst/dark_current/dark_sub.py
```diff
@@ -87,6 +87,13 @@ def do_correction(input_model, dark_model, dark_output=None):
         # They match, so we can subtract the dark ref file data directly
         output_model = subtract_dark(input_model, dark_model)
 
+        # If the user requested to have the dark file saved,
+        # save the reference model as this file. This will
+        # ensure consistency fro the user's standpoint
```
hbushouse: "fro" should be "from"

jwst/dark_current/dark_sub.py
```diff
@@ -87,6 +87,13 @@ def do_correction(input_model, dark_model, dark_output=None):
...
+        if dark_output is not None:
+            log.info('Writing dark reference model to %s', dark_output)
```
hbushouse: Your typical Joe User might not fully appreciate the meaning of "reference model". Perhaps something a little simpler, like "dark current data"? We have to be careful with the phrasing so that we don't give the impression that it's just a copy of the reference file - it is in some cases, when no averaging is done, but the whole point of having this option is to get a copy of it when it's not the same as the reference file (i.e. when it's been averaged). So I think something like "dark current data" is generic enough to not imply it's just another copy of the reference file.

jwst/dark_current/dark_current_step.py
```diff
@@ -10,7 +10,7 @@ class DarkCurrentStep(Step):
     """
 
     spec = """
-        dark_output = output_file(default = None)
+        dark_output = output_file(default = None) # Dark model or averaged dark subtracted
```
hbushouse: This should be OK, but just as a warning for the future, we have to be careful with what gets included in a comment field for the spec params. In the past we've noticed that including anything in parentheses in the comment, e.g. "# Switch for blee (Boolean)" causes the arg parser to think that's part of the definition for the param and immediately gets confused and chokes.

jwst/dark_current/dark_sub.py
stscieisenhamer: In this case, it is the reference model. If the dark is calculated, its done in the else part at line 113.

stscieisenhamer:
@hbushouse are we good?

jwst/dark_current/dark_sub.py
```diff
@@ -244,14 +261,14 @@ def average_MIRIdark_frames(input_dark, nints, ngroups, nframes, groupgap):
...
+            # If there's only 1 frame per group, just copy the dark frames
             if nframes == 1:
                 log.debug('copy dark frame %d', start)
                 avg_dark.data[it, group] = input_dark.data[it, start]
                 avg_dark.err[it, group] = input_dark.err[it, start]
 
-        # Otherwise average nframes into a new group: take the mean of
-        # the SCI arrays and the quadratic sum of the ERR arrays.
+            # Otherwise average nframes into a new group: take the mean of
+            # the SCI self.assertRaises(Exception, fun)rays and the quadratic sum of the ERR arrays.
```
hbushouse: what's the self.assertRaises() doing in here?

stscieisenhamer: omg editor being a tad over-aggressive on autocompletion. fixed. 👍

PR #1104, opened 2017-08-09
WIP: Initial NIRCAM and NIRISS assign_wcs grism pipelines

sosey:
Initial code for the NIRCAM and NIRISS assign_wcs grism pipelines. NIRCAM is currently working as expected, except the final writeout of the datamodel in the step call which needs creation of an additional model schema. I'm still making updates to the NIRISS code to get it inline with recent changes to the models. 

TODO:
- [x] still need to create schema and tags for the selector models in transforms so they save to the output model
  - [x] Nircam
  - [x] Niriss
- [x] fix up formatting in reference files to send to todd
  - [x] Nircam
  - [x] Niriss
- [x] convert the coefficients style specwcs reference to have models instead
  - [x] Nircam
  - [x] Niriss
- [x] clean up the grism object function for extract2d to use
- [x] need to remove the catalog passing in favor of saving the catalog filename to the meta (requires update to the core datamodel schema) so that the extract2d step has access to it
- write module tests (currently have a giant jupyter notebook with use examples)


jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -57,7 +57,17 @@ def process(self, input):
             for reftype in self.reference_file_types:
                 reffile = self.get_reference_file(input_model, reftype)
                 reference_file_names[reftype] = reffile if reffile else ""
-            result = load_wcs(input_model, reference_file_names)
+
+            # need to get the imaging mode distorion reference file
+            # is there a better way to do this? Can we have crds assign?
+            exptype_trans = {"NRC_GRISM": "NRC_IMAGE", "NIS_WFSS": "NIS_IMAGE"}
```
nden: The Nircam rmap will need to be modified to provide a distortion file for this mode.

jwst/assign_wcs/nircam.py
```diff
@@ -1,35 +1,49 @@
...
+        pipeline = exp_type2transform[exp_type](input_model, reference_files)
 
     return pipeline
 
 
 def imaging(input_model, reference_files):
-    """
-    The NIRCAM imaging pipeline includes 3 coordinate frames -
-    detector, focal plane and sky
+    """ The NIRCAM imaging pipeline
```
nden: Could you please keep the formatting of docstrings. This makes it hard to see the real changes (and I like docstrings to start on a new line ;-)).

jwst/assign_wcs/nircam.py
```diff
@@ -54,14 +70,271 @@ def imaging_distortion(input_model, reference_files):
...
+    # translate the x,y detector-in to x,y detector out coordinates
+    # Get the disperser parameters which are defined as a model for each
+    # spectral order
+    log.info("Collecting specwcs information")
+    with NIRCAMGrismModel(reference_files['specwcs']) as f:
+        xcoeff = f.xcoeff
+        ycoeff = f.ycoeff
+        lcoeff = f.lcoeff
+        orders = [j for j, k in enumerate(f.orders)]  # ordered [str names]
+        wrange_selector = f.wrange_selector
```
nden: Does it make sense to have the wavelengthrange information in a separate ref file? We already use `wavelengthrange` reftype for MIRI and Nirspec.

jwst/assign_wcs/nircam.py
```diff
@@ -54,14 +70,271 @@ def imaging_distortion(input_model, reference_files):
...
+
+    # could add wrange input here to track user wavelength input validity
+    nbtrace = NIRCAMGrismModelSelector(orders,
+                                       xcoeff=xcoeff,
+                                       ycoeff=ycoeff,
+                                       lcoeff=lcoeff,
+                                       name='grism2image')
+
+    # (x,y,l,spec_order)->through grism trace polynomials -> x, y
+    det2det = (Identity(4)) | nbtrace
```
nden: What is the purpose of the `Identity` model here? `nbtrace` takes 4 inputs, `Identity`  seems unnecessary.

nden:
I left only a few comments. Could you not change the formatting of existing code. It's really hard to follow the changes because so much of this is just formatting. I'd like to keep the existing code formatting as it is. Part of this I think is caused by your IDE configuration to use less than 100 char line length.

Is there a need for two separate schemas for niriss and nircam? They seem identical although there are subtly space changes in different places which will make them fail. If you choose one I will comment on it.  If both are necessary for some reason I'll look at them.

jwst/assign_wcs/assign_wcs_step.py
sosey: Yup,  working on that with Todd and the science team :)

jwst/assign_wcs/nircam.py
sosey: apologies, I think that happened awhile ago and my editor marks through pep8/linter stuff. Has JWST decided on a particular  formatting for docstrings? I see them a few different ways in the code, but I'll be wary of messing with functions I know I'm not touching :-)  We should probably decide on a consistent way to go at some point since it makes a difference for the sphinx docs

jwst/assign_wcs/nircam.py
sosey: Possibly, I noticed a bug in my model logic this afternoon, this may change when I figure out how to deal with it. Lemme ping you again with the next push?

jwst/assign_wcs/nircam.py
sosey: hmm, they need to be matched with the grism + module + filter + spectral order. The main use right now is for creating the bounding boxes per object+order, since that's moved to extract2D, maybe a separate file makes sense? If we want to add wavelength limitations for inputs to the models then I'd need to pull from somewherel. I guess we could make one that crds matches against grism and filter which contains the wavelength range for each spectral order. This would be for nircam, niriss is using one wavelength range for each grism+filter combination and not specifying anything different by order.

sosey:
>Is there a need for two separate schemas for niriss and nircam?

Are you talking about the reference file schemas or the transform model schemas? I feel like it's schemas all the way down ;-)

The datamodel schemas for the reference files have different required members because the teams have chosen to make the reference files differently; niriss has one for each filter and nircam has one for each grism+module. 

The transform models are different too, though the higher selector models do look very similar. The biggest difference at that level is having to supply a filter wheel offset for niriss. Lower down in the forward and backward transforms they are pretty different. 

It's possible the reference file schemas could merge if we take out the wavelength information and put that in the waverange reference file you suggested. I need to look at that tomorrow with fresher eyes.

jwst/assign_wcs/nircam.py
sosey: you're right, Identity removed :+1:

sosey:
I reworked the NIRCAM models to make them simpler. NIRCAM also now saves the full transform through the grism to the specwcs reference file. I added and tested schemas and tags for the models.  It now runs entirely through the assign_wcs step and writes the output model without error. I also checked the grism object code that reads the source catalog and uses the input_model wcs object to make a list of grism_objects which contain bounding boxes for each order and that appears to be working as well. 

Now on to do the same thigns for Niriss....

sosey:
When the nircam team updates CRDS to match the NRC_IMAGE distortion file with NRC_GRISM we can remove this hack: [fakeout NRC_IMAGE to get distortion file](https://github.com/sosey/jwst/blob/810c579c57b8c6f2a6eb5417855729cbe9244599/jwst/assign_wcs/assign_wcs_step.py#L64)

sosey:
The latest commit has working NIRISS transforms, using the specwcs files with models saved to them. 
I had a filter mismatch on validation for one of the files, with F158M, in `datamodels.properties`:

@nden @bernie-simon 
```ValidationError: In None 'F158M' is not one of ['ANY', 'CLEAR', 'F070LP', 'F070W', 'F090W', 'F1000W', 'F100LP', 'F1065C', 'F110W', 'F1130W', 'F1140C', 'F115W', 'F1280W', 'F140M', 'F140X', 'F1500W', 'F150W', 'F150W2', 'F1550C', 'F170LP', 'F1800W', 'F182M', 'F187N', 'F200W', 'F2100W', 'F210M', 'F212N', 'F227W', 'F2300C', 'F250M', 'F2550W', 'F2550WR', 'F277W', 'F290LP', 'F300M', 'F322W2', 'F335M', 'F356W', 'F360M', 'F380M', 'F410M', 'F430M', 'F444W', 'F460M', 'F480M', 'F560W', 'F770W', 'FLENS', 'FND', 'FNDP750L', 'GR150C', 'GR150R', 'OPAQUE', 'P750L', 'WL3', 'WLP4', 'N/A']```

Need to grab some of the new simulated images available to test out the NIRISS assign_wcs step next.

**Comment by sosey** [comment]
@hbushouse @stscieisenhamer  instead of adding a new catalog entry to the core schema, it seemed better to update the `source_catalog` name that was already there for the grism images being processed, added that at the spec2 level: [source_catalog update for grism images](https://github.com/sosey/jwst/blob/121235613269c11ef7cc710f4988dab2d1273005/jwst/pipeline/calwebb_spec2.py#L141)

**Comment by nden** [comment]
@sosey `F158M` is defined under `pupil` in the core schema.

**Comment by sosey** [comment]
@nden shoot, you're right, I have some of my pupil+filter designations reversed, doh!

**Comment by sosey** [comment]
@nden @hbushouse: Nadia suggested putting the wavelength information that's currently in the grism specwcs files for each instrument into separate waverange reference files. If we do this, we'll need another schema to write this new file, but I think we can get away with using the same one for both instruments. NIRCAM has their grism reference information separated by filter designation and NIRCAM does not. This means that the wavelength range file  will need two items, one for the filter name, and one for the wavelength range for that filter. The waverange information here is used for creating the bounding boxes for each object+order. Each order+filter has a difference min and max wavelength to limit the bounding box.  Since we moved the bounding boxes for objects to extrac2d, I don't need to change anything in the assign_wcs code, but I need to remake the reference files I currently have and update the schemas to remove the wavelength information. 

For niriss the specwcs file is being matched with filter and pupil in crds, but not for nircam. If we move the wavelength range information out to a new file, and include filter names, order information, and min/max wavelength for both instruments in their respective new waverange file, this file would need to match on exp_type and filter(grism)  for both instruments. Then extract2d could pull the filter+range information from the waverange reference file to make the boxes, it will get the wcs model from the input_data wcs which takes x,y,lam,order for nircam and x,y,lam,order,theta for niriss.

File: jwst/assign_wcs/niriss.py [hunk 5]
```diff
@@ -164,8 +171,151 @@ def imaging_distortion(input_model, reference_files):
...
+    # which can translate ra,dec to image frame reference pixels
+    # it also needs to be part of the grism image wcs pipeline to
+    # go from detector to world coordinates. However, the grism image
+    # will be effectively translating pixel->world coordinates in a
+    # manner that gives you the originating pixels ra and dec, not the
+    # pure ra/dec on the sky from the pointing wcs.
+
+    # use the imaging_distortion reference file here
+    img_reference = deepcopy(reference_files)
+    img_reference['distortion'] = reference_files['imaging_distortion']
```
nden: Isn't this just `reference_files['distortion']` instead of `reference_files['imaging_distortion']` or was this set somewhere else and I missed it?

jwst/transforms/models.py
```diff
@@ -941,3 +1023,538 @@ def _toindex(value):
...
+    specified as input using the input delta pix for the specified order
+
+    Notes:
+    ------
+    The evaluation here is linear currently because higher orders have not yet been
+    defined for NIRCAM (NIRCAM polynomials currently do not have any field
+    dependence)
+    """
+    standard_broadcasting = False
+    seperable = False
```
nden: `separable` here and elsewhere

jwst/transforms/tags/jwst_models.py
```diff
@@ -248,8 +323,14 @@ class SnellType(TransformType):
...
-        return Snell(node['prism_angle'], node['kcoef'], node['lcoef'], node['tcoef'],
-                     node['ref_temp'], node['ref_pressure'], node['temp'], node['pressure'])
+        return Snell(node['prism_angle'],
+                     node['kcoef'],
+                     node['lcoef'],
+                     node['tcoef'],
+                     node['ref_temp'],
+                     node['ref_pressure'],
+                     node['temp'],
+                     node['pressure'])
```
nden: Please put this back in the previous formatting.
I don't want to be a pain but accepting this means pain in the future because we have different settings and will have to constantly resolve conflicts. A better solution may be to have the grism stuff in a separate module. I'll leave it up to you but there's really no need to change classes that do not belong to this PR.

jwst/assign_wcs/niriss.py
sosey: yes, it should just be distortion. I missed changing this when the teams agreed to redeliver the imaging distortion refs

jwst/transforms/tags/jwst_models.py
sosey: sorry  - I thought I had reformatted this one too, will do.

sosey:
merging this pr ..the one test failure looks like a timeout and the previous test run all passed (the only change was space formatting on the nirspec file)

PR #1105, opened 2017-08-09
tool to move WCS keywords from Primary to SCI extension

nden:
This is the initial commit tested on one file only.
Needs more testing and some documentation.

It moves the following keywords:
```
['PA_V3', 'V2_REF', 'V3_REF', 'PA_APER', 'VPARITY',
'RA_V1', 'DEC_V1', 'ROLL_REF', 'RA_REF', 'DEC_REF',
'PCi_j', 'CRVAL', 'CRPIX', CDELT', 'CUNIT', 'CTYPE', 'RADESYS'
```

We should consider moveing the velocity aberration keywords as well. We don't currently correct for this but I imagine we will (similar to HST).

### DIFF
--- a/scripts/move_wcs.py
+++ b/scripts/move_wcs.py
@@ -47,7 +47,10 @@ def _collect_wcs_keywords(f):
 
     # Remove kw added by wcslib from the new header
     for kw in wcslib_kw_to_remove:
-        del new_hdr[kw]
+        try:
+            del new_hdr[kw]
+        except KeyError:
+            continue
 
     # Add STScI specific kw to new_hdr
     for kw in stsci_wcs_kw:
--- a/scripts/move_wcs.py
+++ b/scripts/move_wcs.py
@@ -4,14 +4,7 @@ from astropy import wcs
 from astropy.wcs import InvalidTransformError
 import six
 
-"""
-More Kw to consider 
-radesys - I think it should be moved - "OTHER" in HST
-velocity aberrration related kw - move them
-  DVA_RA< DVA_DEC, VA_SCALE
 
-TARG_RA, TARG_DEC
-"""
 stsci_wcs_kw = ['PA_V3', 'V2_REF', 'V3_REF', 'PA_APER', 'VPARITY',
                 'RA_V1', 'DEC_V1', 'ROLL_REF', 'RA_REF', 'DEC_REF']
 
@@ -52,6 +45,7 @@ def _collect_wcs_keywords(f):
             new_hdr[kw] = f[0].header[kw]
         except KeyError:
             pass
+    new_hdr = add_default_keywords(new_hdr)
     return new_hdr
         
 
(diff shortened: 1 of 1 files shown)


hbushouse:
When we first discussed this a month or so ago we had agreed on moving only the FITS standard WCS keywords, like CRPIX, CRVAL, CDELT, PC, etc., but leaving the JWST-specific keywords like V2_REF, V3_REF, RA_REF, DEC_REF, ROLL_REF, etc. in the primary header. Have you now had second thoughts about splitting them up and instead moving them all to the SCI header? What happens when we get to MultiSlit type products that have multiple SCI extensions? They would theoretically all need their own set of FITS WCS keywords, because they'll all have unique values, but they would all still have the same values for things like RA/DEC_REF, V2/V3_REF, etc. Are we just going to have duplicate copies of them in all the SCI headers for completeness?

scripts/move_wcs.py
```diff
@@ -0,0 +1,97 @@
...
+
+"""
+More Kw to consider 
+radesys - I think it should be moved - "OTHER" in HST
+velocity aberrration related kw - move them
+  DVA_RA< DVA_DEC, VA_SCALE
+
+TARG_RA, TARG_DEC
+"""
+stsci_wcs_kw = ['PA_V3', 'V2_REF', 'V3_REF', 'PA_APER', 'VPARITY',
```
hbushouse: What about WCSAXES and V3I_YANG? Or is WCSAXES handled automatically along with the other FITS standard WCS keywords?

hbushouse:
If we do want to have all of the JWST-specific keywords like V2/V3_REF and RA/DEC/ROLL_REF moved, we need to tell SDP ASAP, because so far they've only moved the FITS standard WCS keywords in their code that builds level-1b products. We would also need to update our core schema to move those too.

bernie-simon:
I'm working on a change to datamodels to support getting the wcsinfo metadata out of either the primary header or science extension. I've modified the core schema to include both wcsinfo and wcsinfo_primary, as @hbushouse suggested. Wcsinfo points to the science extension, wcsinfo_primary to the primary header. If after reading a fits file wcsinfo_primary is populated, but not wcsinfo, the code moves the former to the latter. However, it only does this if the fits file has one, and only one, science extension. These changes are all coded :-) but none of them work yet :-(.

scripts/move_wcs.py
nden: Yes, `WCSAXES` is a FITS standard keyword.

nden:
Hm, I guess I have memory leaks. Although #465 says `all WCS related keywords` . 

Can you clarify why in a `MultiSlitModel` each slit would have different FITS WCS keywords but the same STScI specific WCS kw? `CRVAL` is the same as `RA/DEC_REF`, right?

hbushouse:
In a MultiSlitModel for something like a NIRSpec MOS exposure, the full exposure had a single telescope pointing on the sky and hence one set of values for things like V2/V3_REF, RA/DEC/ROLL_REF, etc., but once you've extracted each 2D slit into its own mini-image I would assume that things like CRPIX, CRVAL, CDELT, etc. would get reset to reflect the location of each slit on the sky (i.e. they'd each have CRPIX set to be somewhere inside the slit image boundary, CRVAL would reflect the unique RA/Dec of each slit on the sky, etc.).

philhodge:
In a `MultiSlitModel`, if CRVALi stay the same as for the original image (i.e. the one from which the slices were extracted), CRPIXi would be different from in the original, but all other FITS keywords could be the same.

nden:
We adjust the transform for each slit to go to full frame coordinates first, instead of modifying `CRPIX` which is also currently in the primary header only.

hbushouse:
I have no problem with moving all of the keywords, both FITS standard and JWST-specific, into the SCI header. It does make some sense to keep them all together. We just need to decide very quickly one way or the other, so that SDP can also update their stuff accordingly, because I had previously filed a keyword dictionary ticket to have them move only the FITS standard keywords.

philhodge:
For FITS keywords, it makes sense to change only the CRPIXi keywords.  Consider the case of tangent-plane projection.  The reference point should be the point where the focal surface is tangent to the plane of the detector.  CRVALi are the right ascension and declination at that tangent point.  When you take a subset of the full image, the tangent-plane projection will only be correct if you keep CRVALi the same and change CRPIXi according to the location of the subset within the full image.

nden:
@philhodge Agreed. That's what's done now. Except instead of changing CRPIX we prepend a `Shift` transform to the entire transform and use the original `CRPIX` values.

philhodge:
But how do you record the Shift values in FITS keywords?

nden:
@hbushouse Is there a use case when we may have different pointings? What is the model that rearranges the files to be source based instead of observation based?

nden:
@philhodge Yes, that's `CRPIX` - we just don't use them for non-resampled data. The offset corresponding to `CRPIX` is included in the distortion files.

nden:
If there's no use case where we may have different pointings in different SCI extensions, then I'll modify this script to only move the FITS keywords.

hbushouse:
There is a use case for NIRSpec MOS and FS modes where we have several exposures taken at different dither positions, or even the data for a given source spread across the 2 detectors, which will get reshuffled into a single product at the start of calspec3, such that we now have multiple instances of data cutouts for a given source, each of which will have different pointings from their original exposures and hence will have different RA/DEC/ROLL_REF values, different RA/DEC_V1 values, etc.

At first I had been thinking of handling this in a "fitsblender" like way of somehow including multiple instances of those keywords, but I'm guessing they should remain tied to the rest of their FITS standard companion keywords and hence it would be cleanest and most direct to just have them all grouped together in the SCI headers in the first place.

nden:
What is the name of the model used in the output product?

hbushouse:
The data model that used to be used for the rearranged products (the one that has data for one source from multiple exposures) is MultiExposureModel, but that may have changed with some rework that @stscieisenhamer did on calspec3 a while back.

stscieisenhamer:
Yes, the model the is the result of rearranging exposures to sources is MultiExposureModel.

scripts/move_wcs.py
```diff
@@ -0,0 +1,97 @@
...
+    
+def _collect_wcs_keywords(f):
+    # Convert the WCS to a header
+    try:
+        w = wcs.WCS(f[0].header)
+    except InvalidTransformError:
+        if f[0].header['cunit3'] and f[0].header['cunit3'] == 'micron':
+            f[0].header['cunit3'] = 'um'
+            w = wcs.WCS(f[0].header)
+    new_hdr = w.to_header()
```
hbushouse: The to_header() method does not include elements of the PCn_m matrix if they are equal to the unit matrix, which many of our test data have. Hence the PCn_m keywords do not appear in the SCI header, nor are they removed from the primary header (because only the keywords in new_hdr are removed).

hbushouse:
Why not go ahead and merge it, rather than just closing, so that the script is available in the package.

Issue #1103, opened 2017-08-09
Task: NIRSpec gain rescaling step


hbushouse:
Beginning around April 2017 the NIRSpec commanding was changed to obtain all subarray readouts using a commanded gain of 2, rather than the standard gain 1. The NIRSpec team will be delivering new reference files, where necessary, to use for subarrays that are based on gain=2, including new subarray gain reference files.

Once all detector-level calibrations have been applied in calwebb_sloper, using the gain-specific reference files, the team wants to have the countrate images output from ramp_fit rescaled to account for the different gain setting (i.e. convert the DN/sec values to a new scale as if they had been taken with gain=1). That way all remaining calibrations applied in subsequent pipeline steps, such as flux calibration, can be applied in a gain-independent fashion.

So a new step needs to be created and inserted into the calwebb_sloper pipeline, immediately following ramp_fit, to apply the rescaling operation for subarray exposures taken with gain=2. The subarray gain reference files will contain a keyword called "GAINFACT" that will provide the scaling factor that needs to be applied to the science data to convert them from the gain=2 scale to the gain=1 scale.

hbushouse:
Things that need to be done:

- [ ] Update the gain reference file model schema to include the GAINFACT keyword in its meta data
- [ ] Create the gain_scale step
- [ ] Update the core data model schema to include the GAINFACT keyword, which will be populated by the gain_scale step, and add a S_GANSCL keyword for the new calibration step
- [ ] Update the calwebb_sloper pipeline to call the gain_scale step

hbushouse:
Abandoning the sub-task of updating the outlier_detection step to back out the gain scaling when computing its noise model, in favor of a more general upgrade to outlier detection to have it use the ERR values propagated from upstream in the pipeline. Closing.

PR #1106, opened 2017-08-09
Baseline version of gain_scale step

hbushouse:
Initial version of new and updated modules for the new gain_scale step specified in #1103. Needs testing yet. Relies on the presence of the GAINFACT keyword in the gain reference files. If the keyword isn't present, the step is skipped. Updated the gain reference file schema to include the GAINFACT keyword, and also updated the core schema to include both GAINFACT (which gets populated by the gain_scale step) and the step status indicator S_GANSCL.


hbushouse:
Latest commit includes 

1. A few minor updates to the gain_scale step itself.

2. Implements a couple of first attempts at generic library routines to use in any step for extracting subarrays from reference file data to match science data. These are in module lib/reffile_utils.py. The get_subarray_data function only extracts and returns a subarray out of the "data" array of a reference file. Functions to extract subarrays from all reference file arrays and return a whole new reference file model would also be useful.

3. Updates the jump and ramp_fit steps to use the new reffile utilities to extract subarrays from the gain and readnoise reference files, when necessary.

4. Updates the calwebb_sloper pipeline to call the gain_scale step at the very end on the _rate product model before it is saved to disk.

4. Updates the ramp_fit step to apply gain_scale to the multi-integration (_rateints) product that it creates. I don't like this arrangement - having the ramp_fit step apply gain_scale to the _rateints product, while the calwebb_sloper pipeline applies gain_scale to the _rate product. I'd like to eventually change the ramp_fit step to return both the _rateints and _rate models to the pipeline and let the pipeline apply gain_scale to both of them and then save both products as well.

hbushouse:
The updates to the jump and ramp_fit steps fixes #1108.

Issue #597, opened 2017-01-13
Task: Calibrate FGS guide mode data


hbushouse:
DMS Level 4 requirement 692 specifies the calibration of FGS guide mode data.

FGS guide mode data are acquired and processed differently than regular imaging exposures. The guide mode data have several different flavors:

 - Identification (ID)
 - Acquisition (ACQ1, ACQ2)
 - Track (TRACK)
 - Fine Guide (FINEGUIDE)

Each mode involves reading the FGS detector in either full-frame or subarray modes, where each exposure can consist of many integrations (NINTS>>1), with each integration containing a ramp made up of only 2 reads (NGROUPS=2). SDP will produce level-1b products for each exposure containing the raw detector readouts. The cal pipeline will need to produce level-2 products that have had only minimal processing applied to them, including forming simple differences of the two groups in each ramp, and applying bad pixel masks and flat-fielding.

Daryl Swade is producing an FGS Guide Star Pipeline document specifying the detailed characteristics of each guiding mode and the required processing to be applied.

hbushouse:
Daryl has setup a confluence page tracking the discussion of everything involved in the FGS guide star pipeline. The page is at: https://confluence.stsci.edu/pages/viewpage.action?spaceKey=jwstdms&title=JWST+DMS+Guide+Star+Data+Products+and+Processing

There's a link at the top of the page to the latest version of the design document.

Fortunately for us, the majority of the new work involved in creating the whole pipeline is in the realm of SDP for creating the level-1 products and their contents. Level-2 products, which will be produced by the SSB cal pipeline, do not require a great deal of additional processing, much of which is already in place for calibration of science mode images.

dmggh:
Thanks. I had just asked Daryl and he also pointed me to that page, and said he updates the document almost daily

hbushouse:
The 3 steps that need to be applied to turn a level-1 product into a level-2 product are:

1) Apply bad pixel mask. The existing dq_init step can be used for this. All FGS guide modes use subarray readouts and the dq_init step is already setup to extract matching subarray regions from full-frame MASK reference files.

[...]

PR #1110, opened 2017-08-11
Updates for the FGS guiding data calibration

hbushouse:
Modified the flat-field step to handle Guide mode data models. Misc other little updates to the guider pipeline, adding its import to the pipeline package, adding a .cfg file for the guider_cds step, etc.

As far as I can tell this really does complete the work on the guider calibration pipeline. All steps now work on all modes of guiding data. All we're missing is mask reference files in CRDS that match the size of the ID mode stacked images.


hbushouse:
Fixes #597.

PR #1111, opened 2017-08-11
Update pointing algorithms

### LINKED ISSUE #1107: update pointing algorithms
Northrup Grumman is updating the algorithms for the quaternion to sky coords algorithm. Changes include:

- 2.5 updated steering mirror transformation to new algorithm
- 2.3 Change reading of the J-frame mnemonics. If the mnemonics are all zero, use a predefined set.

If time permits, re-run the original Kinzel test data to track differences.

stscieisenhamer:
Resolves #1107. Specific changes:

- Refactor
- Allow default for missing J-Frame mnemonics
- Update algorithm for the FSM correction


stscieisenhamer:
@jdavies-st The siaf routine I was talking about is `calc_aperture_wcs`, where the reference point is (0, 0).  Not sure this is particularly a SIAF issue.

jwst/lib/set_telescope_pointing.py
```diff
@@ -1,231 +1,222 @@
...
+    header['DEC_V1'] = vinfo.dec
+    header['PA_V3'] = vinfo.pa
+
+    # Update Aperture pointing
+    header['CRVAL1'] = wcsinfo.ra
+    header['CRVAL2'] = wcsinfo.dec
+    header['PC1_1'] = -np.cos(wcsinfo.pa * D2R)
+    header['PC1_2'] = np.sin(wcsinfo.pa * D2R)
+    header['PC2_1'] = np.sin(wcsinfo.pa * D2R)
+    header['PC2_2'] = np.cos(wcsinfo.pa * D2R)
```
hbushouse: wcsinfo.pa is being computed and then in turn used to compute the PCn_m values, but should we also be storing its value directly in the PA_APER keyword? I don't think anyone (SDP or this script) is populating that keyword yet.

hbushouse: Based on the long description of PA_APER in the keyword dictionary, it sure sounds like this is the quantity that's expected to be in that keyword:

Position angle of aperture used with target (deg). The PRD definition of an aperture includes a reference position in the V2-V3 frame. The position angle of the aperture reference position is calculated from the position angle of the V3 axis and PRD defined translations from V2-V3 to aperture reference frames.

stscieisenhamer: 👍 I was just looking to see if it is so populated. The identified code would be what produces this value and can make this happen.

jdavies-st: I believe PA_APER is the position angle at the aperture center of V3.  The PC matrix needs to also take into account V3I_YANG, which is the orientation of the detector relative to V3.  `vinfo.pa` takes this into account, yes?

nden: Are `PA_APER` and `ROLL_REF` the same angle?

hbushouse: That's a good question. According the keyword dictionary PA_APER is defined as:

Position angle of aperture used with target (deg). The PRD definition of an aperture includes a reference position in the V2-V3 frame. The position angle of the aperture reference position is calculated from the position angle of the V3 axis and PRD defined translations from V2-V3 to aperture reference frames.

So the first part makes it sound like it's the position angle at the reference point of the aperture, in which case I would think it's the same as ROLL_REF (as opposed to PA_V3, which is the roll/pa at V1, rather than at the aperture reference point). The rest I'm not sure about.

nden: I'll ping Colin.

hbushouse: The text at this web site (http://www.stsci.edu/hst/acs/documents/handbooks/cycle19/c07_obstechniques09.html) makes it sound like PA_APER is definitely at the aperture, not at V1, but I'm not sure about how V3I_YANG folds into that.

nden: From Colin:
```
These are not the same. ROLL_REF is measured from North to V3.
PA_APER is from North to the direction of the aperture Y axis.
So PA_APER = ROLL_REF + V3SciYAngle
```

stscieisenhamer: So, there are two code bits to consider: The SIAF transformation is being done by the `m_v2siaf` matrix, calculated in the `calc_v2siaf_matrix` routine. Then, using that matrix, the wcs, ra/dec/pa, is calculated in `calc_aperture_wcs`.  The question is, does the matrix + calculation based on that matrix account for the above? Note that V3IDLYANG is involved in the matrix, and if this is the V3SciYAngle, then what is coming out of `wcsinfo.pa` is PA_APER. No?

jdavies-st: In the old `set_telescope_pointing.py` script, `pa_aper_deg` was the same as the ROLL_REF fits keyword.  I believe the difference between PA_APER and ROLL_REF keywords is PA_APER is just PA_V3 translated to the aperture center, but ROLL_REF also takes into account V3I_YANG, i.e. the angle of the y-axis of the detector in the V2/V3 plane.

nden: Here's some more info from Colin in addition to the piece above:
```
The V3SciYAngle and V3IdlYAngle can differ by 180 degrees, but that only happens
for some OSS apertures. 
```
In `calc_aperture_wcs` the angle `vy_dec` is the Y-axis angle of the aperture in the Ideal system. The calculation also uses `vparity` which means the angle is now always identical to V3SciAngle, right? So I think I agree with @stscieisenhamer - this gives us `pa_aper` which accounts for the YIdlAngle.

`Roll_ref` uses `PA_V3` and does not take into account the `V3IdlAngle`.

Also from Colin: 
```
For the PC matrix, yes I think you need PA_APER which accounts for the telescope position.
```
We already do this. So I think this is good.
👍

jdavies-st: Thanks for the clarification @nden (and Colin).  Looks good to me.

I agree with Howard then, that PA_APER should be populated by the pointing script, as this is a very useful quantity for level 3 processing.  That way it doesn't need to be always backed out of the PC matrix.

hbushouse:
This comment is really more relevant to the work of #1123 and #982, but will affect this PR as well. The changes in #1123 moved all keywords located in the meta.wcsinfo block to the SCI header, but left the keywords in the meta.pointing block, which contains RA_V1, DEC_V1, and PA_V3, in the primary header. These 3 keywords are also populated by the pointing script and are directly related to other keywords like RA_REF, DEC_REF, and ROLL_REF and hence IMHO they should all be kept together. At the very least this means modifying the core schema to also designate these last 3 keywords as being in the SCI header.

But I would like to propose an even further modification. Because RA_V1, DEC_V1, and PA_V3 are very much related to RA_REF, DEC_REF, and ROLL_REF, why shouldn't they be in the same meta data block? We could accomplish this by removing the "pointing" block and just putting them all in the "wcsinfo" block, but I'm thinking it might be more logical to only have standard FITS WCS type keywords in the "wcsinfo" block and move everything like V2_REF, V3_REF, VPARITY, V3YANGLE, RA_REF, DEC_REF, ROLL_REF, RA_V1, DEC_V1, and PA_V3 all together in the "pointing" block. As far as users are concerned it's only the standard FITS WCS keywords that set the WCS for the image, while all of these other keywords give some form of the telescope pointing. Making this change would of course require changes to any code (e.g. assign_wcs) that works with these meta data so that it looks in the right meta block for them.

Granted a lot of this is just cosmetics of how the keywords are arranged in the FITS header and don't matter to functionality and would also mean additional work of updating meta references in some code (which is additional work we don't necessarily need heaped on us right now), but at the very least we have to get the RA_V1, DEC_V1, and PA_V3 keywords into the SCI header, because for MultiExposure products, there will be unique values of these keywords for every SCI extension.

hbushouse:
Here's what can be added in at an appropriate point near the end of this script, once it has calculated and updated the standard FITS WCS keyword info, in order to give the archive folks a quick-n-dirty version of the S_REGION keyword to play with in our next pre-delivery. It will need to be updated in the future to use the proper scheme from APT that employs aperture vertices from the SIAF.

```
import astropy.wcs as wcs

w = wcs.WCS(pheader, naxis=2)
footprint = w.calc_footprint(center=True)
pheader['s_region'] = 'POLYGON  ICRS' + \
    str(footprint[0][0]) + ' ' + str(footprint[0][1]) + ' ' + \
    str(footprint[1][0]) + ' ' + str(footprint[1][1]) + ' ' + \
    str(footprint[2][0]) + ' ' + str(footprint[2][1]) + ' ' + \
    str(footprint[3][0]) + ' ' + str(footprint[3][1]) + ' ' + \
    str(footprint[0][0]) + ' ' + str(footprint[0][1])
```

Of course that code assumes you've got a fits header object to use as input to wcs.WCS(), which we don't have anymore (now that the script uses a data model). I'm sure you can figure out a way to make it work!

stscieisenhamer:
@hbushouse added the info, please review last commit and take into account the related test in `test_set_telescope.py`

jwst/lib/set_telescope_pointing.py
```diff
@@ -196,6 +198,50 @@ def update_wcs(model, default_pa_v3=0., **kwargs):
...
+
+    # Get the WCS info from the model
+    wcsinfo = model.meta.wcsinfo
+
+    # Using astropy.wcs, the most fullproof method
+    # of getting the wcs is to put everything in a FITS
+    # header and initialize from that.
+    header = Header(wcsinfo.instance)
+    wcs = WCS(header)
+    footprint = wcs.calc_footprint(undistort=True, center=True)
```
hbushouse: I'm not sure what effect, if any, the undistort option will have here, since we don't have any info in the wcs at this point that provides any kind of distortion solution. I'm guessing this only works if there's something like SIP polynomials specified in the wcs keywords.

stscieisenhamer: It was suggested by Brian, so I left it in for consistency.  If it may be detrimental, I'll remove, but if the concept is, as stated in another issue/pr, to keep this updated all the way through, this routine could be the central one to use, maybe.

hbushouse: As long as it doesn't cause a problem when executed on level-1b headers, it's fine to leave it in. I'm just concerned that it might not be able to find whatever it needs to do the undistort and cause an error in the process. If the routine is robust enough to move on without an error when it can't find distortion info, then we're good.

stscieisenhamer: well, the current set of tests don't of SIP, and it seems to work, so, will leave in until otherwise noted. However, it is probably good to trap exceptions here anyways since it is not a critical item to the script running.

PR #1112, opened 2017-08-11
add ideal to v2v3 transform

nden:
Implements transforms between the telescope Ideal and V2,V3 coordinate systems.
Also adds XOFFSET, YOFFSET to the core schema.


jwst/transforms/models.py
```diff
@@ -813,6 +813,122 @@ def __call__(self, v2, v3):
...
+            Parity.
+
+        Returns
+        -------
+        v2, v3 : ndarray-like
+            Coordinates in the (V2, V3) telescope system [in deg].
+
+        """
+        v2ref = np.deg2rad(v2ref / 3600)
+        v3ref = np.deg2rad(v3ref / 3600)
```
hbushouse: Up above in Parameters you indicate that v2ref and v3ref are in degrees, yet here it looks like you're assuming the input values will be in arcsec and are being converted to degrees.

jwst/transforms/models.py
```diff
@@ -813,6 +813,122 @@ def __call__(self, v2, v3):
...
+        v3idlyangle = np.deg2rad(v3idlyangle)
+        xidl = np.deg2rad(xidl)
+        yidl = np.deg2rad(yidl)
```
hbushouse: For the case that we need this transform (converting the XIdl, YIdl values from PPS) the inputs will be in units of arcsec. Here you're assuming they're in degrees. Will the inputs be converted from arcsec to degrees before being sent to this function?

jwst/transforms/models.py
```diff
@@ -813,6 +813,122 @@ def __call__(self, v2, v3):
...
+        vparity : int
+            Parity.
+
+        Returns
+        -------
+        xidl, yidl : ndarray-like
+            Coordinates in the Ideal telescope system [in deg]. 
+
+        """
+        v2ref = np.deg2rad(v2ref / 3600)
```
hbushouse: Same comment regarding degrees vs. arcsec

nden:
Thanks @hbushouse. Units fixed and tests passing now - had to update a couple test files.

hbushouse:
:1st_place_medal:

PR #1113, opened 2017-08-11
A few more updates to FGS guider calibration

hbushouse:
Updates to dq_init and flat_field to handle the case of reference files not matching the weird size of the ID mode images that have the strips stacked into a larger than 2048x2048 shape. Eventually there should be reference files in CRDS to match the size of this mode, but for now these checks will cause the dq_init and flat_field steps to be skipped.

Also minor updates to core schema and some pep8 cleanup.


hbushouse:
In support of #597.

PR #1114, opened 2017-08-11
Modify datamodels to read and write wcs from/to SCI extension

bernie-simon:
The pipeline poobahs have decreed that the fits keywords describing the wcs should be in the science extension instead the primary header. To support this, the core schema has been updated with two sets of wcs keywords: wcsinfo, which now points to the SCI extension, and wcsinfo_primary, which points to the primary header. The code which reads fits headers in fits_support.py has been modified to copy wcs metadata from wcsinfo_primary to wcsinfo, so by the time the user sees the metadata the wcs fields are under meta.wcsinfo. This copying is only done if a fits file has a single SCI extension, none or more than one leaves the wcs metadata under wcsinfo_primary. The get_fits_wcs and set_fits_wcs methods now default to the SCI extension. This shouldn't matter, as no code seems to use them.
This change will break existing tests. First, because any code that writes a model will now write the wcs keywords to the SCI extension and not to the primary header. Second, because datamodels now reads the wcs in the SCI extension, the keywords there must validate. For those reasons the tests and test data for datamodels was changed.

This example script shows how simply reading and writing a datamodel can copy the wcs keywords from the primary header to the SCI extension:

    import os
    import sys
    from jwst import datamodels
    
    def move_wcs(filename):
        backup = filename + "~"
        os.rename(filename, backup)
        try:
            im = datamodels.open(backup)
            im.save(filename)
            im.close()
        except:
            os.rename(backup, filename)
            print("Couldn't update " + filename)
        else:
            os.remove(backup)
    
    if __name__ == "__main__":
        if len(sys.argv) > 1:
            for filename in sys.argv[1:]:
                move_wcs(filename)


nden:
👎  on having `wcsinfo_primary`. It's too confusing. I'd rather have everyone fix their data now than put stuff like this in the code that magically changes the data.
And how do we decide when to remove `wcsinfo_primary` and the related code from the pipeline? Surely we are not going to keep it there forever.

hbushouse:
It makes me feel bad that Bernie went to all this work to make it happen, but I'm beginning to agree with Nadia (and others) that it's probably better in the long run to just make a clean break and force people to update their files, rather than allowing them to get away with bad files for some indeterminate amount of time.

bernie-simon:
The bi-weekly paycheck takes some of the sting out of the disappointment. Even if you don't want datamodels to copy from the primary header to the science extension, half of the change still needs to go through. The core schema needs to have wcsinfo point at the SCI extension and get_fits_wcs and set_fits_wcs need to default to it as well. If this "tough love" version of the fix is installed, expect some breakage in the tests.

jdavies-st:
Agree with Nadia and Howard.

Howard's PR #982 (not merged yet) already updates the core schema.  As for the methods `get_fits_wcs` and `set_fits_wcs`, these are leftovers from the pipeline infrastructure before we switched from `astropy.wcs` to `gwcs`. They are not currently being used for anything.

We should have a general discussion elsewhere what to do with these methods.  This is related to the `astropy.nddata` and `datamodels` subclassing.

Finally, the issue of using FITS standard WCS in the resampled data products has been mentioned - `gwcs` may not be necessary as the resampled products have a simple linear tangent plane.  Whether this can work for spectral modes is another question.  If we decide to do this, then the 2 methods above will need to stay and be updated appropriately as Bernie has done.

bernie-simon:
The methods get_fit_wcs and set_fit_wcs are needed for compatibility with NDData. Test changes are needed to prevent errors when wcs is moved to SCI extension.

larrybradley:
I'm using `get_fits_wcs` in the source catalog step.

philhodge:
Shouldn't `get_fits_wcs` have a keyword argument `hdu_ver`, with default value 1?

jdavies-st:
@larrybradley, currently we do not update the FITS header keywords in `meta.wcsinfo` with the new resampled output frame, as we are using the `gwcs` object to define the output frame.  Which means that currently the source catalogs have the wrong coordinates.  I need to fix this, but we need to decide on the standard way we'd like to do that.  I.e. a standard way of representing a linear FITS WCS in `gwcs`.

bernie-simon:
It hasn't been there, but the keyword hdu_ver can certainly be added.

larrybradley:
I'm using `get_fits_wcs` because `photutils` currently only works with astropy `WCS` objects (i.e. FITS WCS) -- and FITS WCS should be fine for a rectified image (assuming it agrees with the gwcs).  Unfortunately `gwcs` and astropy `WCS` objects do not have a common interface, so supporting both in a given package will require lots of conditional branching code.

@nden Are there any plans to "unify" the wcs interfaces or should I just start adding a bunch of conditionals in `photutils` to support both?

nden:
@larrybradley Yes, there will be a unified interface for the two WCS objects. Just haven't had time to work much on this.

bernie-simon:
I'm closing this pull request and will open a new request that reflects the reduced scope of the requested change.

PR #1117, opened 2017-08-11
Add caltso3 step

stsci-hack:
Adding the calwebb_tso3 pipeline step to the JWST pipeline.  This includes revisions to the tso_photometry and white_light packages in order to allow the caltso3 pipeline to collate and write out the expected products.  This code was successfully used to manually process both NIRCAM (imaging) and NIRISS (spectroscopic) TSO data, and should support ASNs with multiple TSO exposures.  

This update also requires PR #1010 in order to successfully process the TSO data without resampling.  

A version of outlier_detection, outlier_detection_scaled, has also been included for validation with TSO data to enable the detection code to scale properly with the source flux and hopefully produce results which do not clip variable sources.  It is not used by default, but can be turned on with the use of a step parameter.


jwst/pipeline/calwebb_tso3.py
```diff
@@ -0,0 +1,143 @@
...
+from ..outlier_detection import outlier_detection_tso_step
+from ..outlier_detection import outlier_detection_scaled_step
+from ..tso_photometry import tso_photometry_step
+from ..extract_1d import extract_1d_step
+from ..white_light import white_light_step
+
+__version__ = "0.7.8"
+
+
+class TSO3Pipeline(Pipeline):
```
hbushouse: Extremely minor nit: Do we have a consensus on camel-case naming of classes, in terms of what we do with acronyms? Some folks, including myself, only capitalize the first letter of acronyms, e.g. Ami3Pipeline, just so that it's always only the first letter of a word that's capitalized. Is there a convention for this?

jwst/pipeline/calwebb_tso3.cfg
```diff
@@ -0,0 +1,14 @@
+name = "Image3Pipeline"
+class = "jwst.pipeline.Image3Pipeline"
```
hbushouse: Your going to need to rename these lines to TSO3Pipeline.

jwst/pipeline/calwebb_tso3.py
```diff
@@ -0,0 +1,143 @@
...
+
+    Included steps are:
+        outlier_detection
+        tso_photometry
+        extract_1d
+        white_light
+    """
+
+    spec = """
+        suffix = string(default='i2d')
```
hbushouse: I'm not sure we need or want to have a single "standard" suffix defined for the output from this pipeline. Unlike pipelines like calspec2 or image2 where you know you're always going to at least get a _cal product out the backend, with others possible, there isn't one single product that will always be produced by this pipeline, because it's handling both images and spectra. The only one you should always get are the level-2c (crf) products from outlier_detection, but I think those are saved by outlier_detection itself? From there on, you might get an imaging photometry catalog (_phot product) or you might get an _x1d and _whtlt product.

jwst/pipeline/calwebb_tso3.py
```diff
@@ -0,0 +1,143 @@
...
+        Run the TSO3Pipeline
+
+        Parameters
+        ----------
+        input: Level3 Association, or ModelContainer
+            The exposures to process
+        """
+
+        self.log.info('Starting calwebb_tso3 ...')
+        print("scale_detection set to : {}".format(self.scale_detection))
```
hbushouse: Aaaack! NO PRINT STATEMENTS. Use the logging facility.

jwst/pipeline/calwebb_tso3.py
```diff
@@ -0,0 +1,143 @@
...
+        input_asn = datamodels.open(input)
+        product_name = input_asn.meta.asn_table.products[0].name
+
+        # Setup output creation
+        #self.output_basename = product['name']
+        self.output_basename = product_name 
+        
+        # Input may consist of multiple exposures, so loop over each of them
+        for cube in input_asn:
+            # Convert CubeModel into ModelContainer of 2-D DataModels
```
hbushouse: The comment should say "3-D DataModels"

jwst/pipeline/calwebb_tso3.py
```diff
@@ -0,0 +1,143 @@
...
+                        err=cube.err[i], dq=cube.dq[i])
+                image.meta = cube.meta
+                input_models.append(image)
+            
+            if not self.scale_detection:
+                self.log.info("Performing outlier detection on input images...")
+                results = self.outlier_detection(input_models)
+            else:
+                self.log.info("Performing scaled outlier detection on input images...")
+                results = self.outlier_detection_scaled(input_models)
```
hbushouse: How is `results` different than `input_models`? If `input_models` is unchanged, while `results` is a copy of the input with the DQ arrays updated, then line 90 below should be copying from `results[i].dq`, not `input_models[i].dq.` If `input_models` has updated DQ flags, then what is `results` used for?

jwst/pipeline/calwebb_tso3.py
```diff
@@ -0,0 +1,143 @@
...
+        self.log.info("Writing Level 2c images with updated DQ arrays...")
+        suffix_2c = 'crfints'
+        for cube in input_asn:
+            self.output_basename = cube.meta.filename
+            self.save_model(cube, suffix=suffix_2c)
+
+        # Create final photometry results as a single output
+        # regardless of how many members their may be...
+        phot_result_list = []
+        if 'image' in results[0].meta.exposure.type.lower():
```
hbushouse: I think this works right now for NIRCam TSO imaging, where the EXP_TYPE = 'NRC_TSIMAGE', but other instruments will also (in the future) have the capability of doing TSO in imaging mode (MIRI, NIRISS) and we don't know what the EXP_TYPE values will be for those yet. So perhaps it's safer to do this by having a dictionary of EXP_TYPE values, e.g. image_exptypes = ['NRC_TSIMAGE'], and check to see if results[0].meta.exposure.type is in that list. Then the list can simply be updated in the future when we find out what the EXP_TYPE's will be for the other instruments.

jwst/pipeline/calwebb_tso3.py
```diff
@@ -0,0 +1,143 @@
...
+            # Create name for extracted photometry (Level 3) products
+            phot_tab_name = "{}_phot.ecsv".format(product_name)
+
+            # For each cube, extract photometry...
+            for cube in input_asn:
+                # Extract Photometry from imaging data
+                phot_result_list.append(self.tso_photometry(cube))
+        else:
+            # Working with spectroscopic TSO data...
+            # define output for x1d (level 2c/3) products
```
hbushouse: Technically all output products from here on are only level 3. The 2c's were already created above.

jwst/pipeline/calwebb_tso3.py
```diff
@@ -0,0 +1,143 @@
...
+                self.log.info("Extracting 1-D spectra...")
+                self.extract_1d.suffix = 'x1dints'
+                result = self.extract_1d(cube)
+                x1d_models.append(result)
+                #
+                # perform white-light photometry on 1d extracted data
+                self.log.info("Performing white-light photometry...")
+                phot_result_list.append(self.white_light(result))
+                
+            # Define suffix for stack of extracted 1d (level 2b) products
```
hbushouse: comment should say 'level 3', not 'level 2b'

jwst/pipeline/calwebb_tso3.py
```diff
@@ -0,0 +1,143 @@
...
+            #self.save_model(x1d_models,suffix="x1dints") # did not work!!!
```
hbushouse: Interesting. Better have @stscieisenhamer or @bernie-simon have a look at that.

jwst/pipeline/outlier_detection.cfg
```diff
@@ -13,3 +13,5 @@ snr = '4.0 3.0'
 scale = '0.5 0.4'
 backg = 0.0        
 save_intermediate_results = False
+resample_data = True
+good_bits = 4
```
hbushouse: 4 is the value used for a CR hit, so why is this value considered "good"?

hbushouse:
In support of #616 and #619.

jwst/pipeline/calwebb_tso3.py
stsci-hack: Debugging statement that slipped through... will remove!

jwst/pipeline/calwebb_tso3.py
stsci-hack: The comment is actually correct.  This code is converting a 3D DataModel(DM) into a container (list) of 2D DataModels where each 2D DM represents a single slice from the cube so that they can be run through outlier_detection (which only works on stacks of 2D data).

jwst/pipeline/calwebb_tso3.py
stsci-hack: After experimenting for quite a while, I determined that it comes down to what value gets used to set the self.product_basename.  It is easy to set it to create a final product, but this method can not be used easily to generate output names based on each of the multiple input exposures.  Therefore, I found it simpler and more transparent to define the output names in the calling code then call the write method on the generated product.

stsci-hack: Actually, after playing around some more, this call to '.save_model' just did not write out anything. Fortunately, we have a workaround for the time being.

jwst/pipeline/outlier_detection.cfg
stsci-hack: One reason for this value would be to ignore any previously identified CRs so that this algorithm can identify them independently.  However, if someone (like the JWST CWG) has suggestions on what bits should be considered good, we can use that value instead.

jwst/pipeline/calwebb_tso3.py
stscieisenhamer: There are a couple of generic approaches in play here. This will work for now but will look more closely within a day.

jwst/pipeline/calwebb_tso3.py
larrybradley: PEP8 (https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles) says "Note: When using abbreviations in CapWords, capitalize all the letters of the abbreviation. Thus HTTPServerError is better than HttpServerError."

So that would mean `TSO3Pipeline` is preferred.  For the record, I also capitalized:  `TSOPhotometryStep`.

hbushouse: OK, so because:
1) foolish consistency is the hobgoblin of little minds, and
2) I have a little mind,
we will therefore use all caps for acronyms that appear in class names.

jwst/pipeline/calwebb_tso3.py
hbushouse: I knew that ...

jwst/pipeline/outlier_detection.cfg
hbushouse: I see, so good_bits is used to tell the routine to ignore certain conditions on input and in this case it's telling it to ignore anything that was flagged as a jump on a previous pass (so that it starts over from scratch).

jwst/pipeline/calwebb_tso3.py
stsci-hack: Discussing this with Howard, we realized that the code is trying to use the .save_model() method on a ModelContainer created within the Step.  This may be causing this code to not initialize everything as needed to work; the method simply runs without complaint and produces no file on disk.

hbushouse:
Can this be merged so that it's available in the pre-release delivery we're planning in the next few days?

stsci-hack:
I believe all the issues raised have been resolved.  Regression tests will be added to track this pipeline.

PR #1122, opened 2017-08-14
Rename white light suffix to 'whtlt'

### LINKED ISSUE #991: Create unique product type names for different flavors of source/photometry catalogs.

larrybradley:
This PR closes #991.

### DIFF
--- a/jwst/white_light/white_light_step.py
+++ b/jwst/white_light/white_light_step.py
@@ -2,6 +2,7 @@
 
 from ..stpipe import Step, cmdline
 from .. import datamodels
+from ..source_catalog.source_catalog import _replace_suffix_ext
 from .white_light import white_light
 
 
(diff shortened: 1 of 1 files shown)


jwst/white_light/white_light_step.py
```diff
@@ -2,6 +2,7 @@
 
 from ..stpipe import Step, cmdline
 from .. import datamodels
+from ..source_catalog.source_catalog import _replace_suffix_ext
```
hbushouse: I'm not a big fan of using semi-generic routines like this that come from another step, but it works for now. If we really need this kind of capability, we should put it into a lib routine. @stscieisenhamer may know if we've already got something like this somewhere in stpipe or datamodels.

larrybradley: I agree.  I'll submit a separate PR moving this function to `lib`.

PR #1123, opened 2017-08-14
Move wcsinfo to science extension

bernie-simon:
The core schema has been updated to read and write the wcs from the science extension instead of the primary header. Tests and test data have been updated to match. Default arguments of get_fits_wcs and set_fits_wcs have been updated. A hdu_ver argument has been added to get_fits_wcs.


jwst/datamodels/tests/test_wcs.py
```diff
@@ -49,7 +49,7 @@ def test_wcs():
 
     wcs1.wcs.crpix[0] = 42.0
 
-    dm2.set_fits_wcs(wcs1)
+    dm2.set_fits_wcs(wcs1, hdu_name="PRIMARY")
```
hbushouse: Should this be "SCI" instead of PRIMARY, since we're now putting them all into the SCI header?

PR #1130, opened 2017-08-16
Allow dash as a calibration suffix indicator

### LINKED ISSUE #1128: handle dash separated suffixes
@hbushouse writes:

> So I've discovered a problem with the FGS guide mode data sets when processing them through the new calwebb_guider pipeline. The problem is that the names of some of the input (raw) products look like:
>
> jw88600073001_gs-id_7_image-uncal.fits
> jw88600073001_gs-id_7_stacked-uncal.fits
>
>The problem here, in case you haven't already spotted it, is the fact that the "uncal" suffix is preceded by a dash rather than an underscore.
>
> In the calwebb_guider pipeline I set self.suffix='cal' and it succeeds in creating an output file name that has the "cal" suffix tacked on, but because of the dash in the existing file name it doesn't replace the "uncal" with "cal." Instead it just appends "cal", so that you end up with:
>
> jw88600073001_gs-id_7_image-uncal_cal.fits
> jw88600073001_gs-id_7_stacked-uncal_cal.fits
>
> I find this result surprising in that I would've expected the step infrastructure that does the file name creation to look for the last occurrence of underscore and replace everything following it, e.g.
>
> jw88600073001_gs-id_7_cal.fits
> jw88600073001_gs-id_7_cal.fits
>
> which is equally bad, because that would lead to file name degeneracy.
>
> Ideas as to why it's doing what it's doing and how we might be able to fix it so that we end up with names like:
>
> jw88600073001_gs-id_7_image-cal.fits
> jw88600073001_gs-id_7_stacked-cal.fits
>
>i.e. just the "uncal" has been replaced with "cal".

stscieisenhamer:
Arrrggghhhhh....argggghhhhh

arggghhhhhhh

and one more....arrrrgggghhhhh

Resolves #1128

### DIFF
--- a/jwst/stpipe/step.py
+++ b/jwst/stpipe/step.py
@@ -24,7 +24,8 @@ from .. import __version_commit__, __version__
 
 SUFFIX_LIST = ['rate', 'cal', 'uncal', 'i2d', 's2d', 's3d',
                'jump', 'ramp', 'x1d', 'x2d', 'x1dints', 'calints', 'rateints']
-REMOVE_SUFFIX = '^(.+?)(_(' + '|'.join(SUFFIX_LIST) + '))?$'
+REMOVE_SUFFIX = '^(?P<root>.+?)((?P<separator>_|-)(' \
+                + '|'.join(SUFFIX_LIST) + '))?$'
 
 
 class Step(object):
@@ -554,11 +555,11 @@ class Step(object):
         gc.collect()
 
     def _precache_reference_files_opened(self, model_or_container):
-        """Pre-fetches references for `model_or_container`.   
+        """Pre-fetches references for `model_or_container`.
 
         Handles recursive pre-fetches for any models inside a container,
         or just a single model.
-        
+
         Assumes model_or_container is an open model or container object,
         not a filename.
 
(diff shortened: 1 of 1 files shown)


hbushouse:
He must have died while writing the PR ...

jwst/stpipe/step.py
```diff
@@ -24,7 +24,8 @@
 
 SUFFIX_LIST = ['rate', 'cal', 'uncal', 'i2d', 's2d', 's3d',
                'jump', 'ramp', 'x1d', 'x2d', 'x1dints', 'calints', 'rateints']
```
hbushouse: There isn't an "x2d" suffix used anywhere. We should probably add "crf" and "crfints" for completeness.

jwst/stpipe/step.py
```diff
@@ -24,7 +24,8 @@
...
-REMOVE_SUFFIX = '^(.+?)(_(' + '|'.join(SUFFIX_LIST) + '))?$'
+REMOVE_SUFFIX = '^(?P<root>.+?)((?P<separator>_|-)(' \
+                + '|'.join(SUFFIX_LIST) + '))?$'
```
hbushouse: Are you really expecting me to verify that's the correct expression?  Aaaaaaaargh. I see an underscore and a dash separated by an or-bar buried in there, so I'm going to guess that'll do it.

PR #1131, opened 2017-08-17
Move remaining pointing keywords to SCI header

hbushouse:
Move remaining V1 axis pointing keywords RA_V1, DEC_V1, PA_V3 from primary to SCI header. Now all pointing related keywords are in the SCI header.


hbushouse:
Failing tests are all due to some of these keywords now being in a different header than the FITS files used in the tests. Will update the test files. Merging.

PR #1139, opened 2017-08-22
Place wcsinfo metadata in its own schema

bernie-simon:
The wcsinfo metadata is now read from the SCI extension. However, some image types do not contain a SCI extension and others contain multiple SCI extensions. For these images wcsinfo is not relevant, so wcsinfo should not be in the core schema. It was moved out into its own schema and those schemas with a single SCI extension were modified to include it. One test needed to be modified so that it creates an ImageModel, which has wcsinfo, instead of a DataModel, which does not.


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -730,215 +730,6 @@ properties:
             type: number
             fits_keyword: PA_APER
             fits_hdu: SCI
```
hbushouse: We also have to move out the other few WCS-related keywords that are now assigned to the SCI header, which includes RADESYS, PA_APER, VA_SCALE, RA_V1, DEC_V1, PA_V3 (I think that's all of them). Basically just look for any keyword the has "fits_hdu: SCI", even if they aren't part of the wcsinfo block.

jwst/datamodels/schemas/guider_cal.schema.yaml
```diff
@@ -1,5 +1,6 @@
 allOf:
 - $ref: core.schema.yaml
+- $ref: wcsinfo.schema.yaml
 - $ref: bunit.schema.yaml
 - type: object
   properties:
```
hbushouse: Should also be added into the guider_raw.schema.yaml, which uses a SCI array.

jwst/datamodels/schemas/miri_ramp.schema.yaml
```diff
@@ -1,5 +1,6 @@
 allOf:
 - $ref: core.schema.yaml
+- $ref: wcsinfo.schema.yaml
 - $ref: bunit.schema.yaml
 - type: object
   properties:
```
hbushouse: I think it also needs to be added multiexposure.schema.yaml and multiproduct.schema.yaml, because they inherit core and use a SCI array, just like multislit.schema.yaml.

jwst/datamodels/schemas/core.schema.yaml
bernie-simon: What do you want to do in the case where one keyword in a group is written to the SCI extension, but another is not? For example, in velocity_aberration va_scale is written to the SCI extension, but ra_offset and dec_offset are not. Should all the properties under velocity_aberration be copied into wcsinfo.schema.yaml or only the one keyword, va_scale?

jwst/datamodels/schemas/miri_ramp.schema.yaml
bernie-simon: These two schemas have multiple SCI extensions, as indicated by the comments "An array of slits" and "An array of DrizProducts." If there is more than one SCI extension datamodels will only write to the first one. As I said in my original comment, I decided to only include wcsinfo when there is exactly one SCI extension. But if you feel differently, let me know.

jwst/datamodels/schemas/core.schema.yaml
hbushouse: For now just move the individual keyword from the block. That means there won't be a block header/comment for it in the SCI header, but that's OK.

jwst/datamodels/schemas/miri_ramp.schema.yaml
hbushouse: I'm not sure how these are supposed to work. I know that for models like MultiSlitModel, where each slit instance contains a SCI extension, it worked fine to have keywords assigned to the SCI header in the schema, in which case you get unique values of those keywords in each SCI extension in the FITS file. It looks like models like MultiExposure are built the same way, using a list of "data" attributes, each one of which maps to a SCI extension, and hence it should be possible to have the WCS keywords defined to be in the SCI extension and they should show up - with potentially unique values - in each SCI extension.

bernie-simon:
I've updated the pull request with a new version of the schemas that should resolve all the issues.

PR #1145, opened 2017-08-23
Fix min crds

stsci-hack:
Update outlier_detection step to fix a problem with code to optimally call CRDS to obtain requested reference files. This should fix problems identified in regression testing.


stsci-hack:
This PR has been superceded by #1174 and therefore will be closed to avoid any confusion.

PR #1146, opened 2017-08-23
Fix asdf version mismatch warnings

drdavella:
Recent updates to the ASDF library added more stringent warnings for mismatches between supported tag versions and tag versions found in ASDF files. In general, this is a good thing. However, it caused a lot of new warnings when running the built-in JWST test suite. This PR implements the changes that make the warnings go away.

### DIFF
--- a/jwst/exp_to_source/tests/data/jwst_nod1_cal.fits
+++ b/jwst/exp_to_source/tests/data/jwst_nod1_cal.fits
@@ -147,7 +147,7 @@ slits:
             forward:
             - !transform/shift-1.0.0 {offset: 0.0}
             - !transform/shift-1.0.0 {offset: 0.0}
-          - !transform/identity-1.0.0 {n_dims: 2}
+          - !transform/identity-1.1.0 {n_dims: 2}
           name: dms2sca
       - !wcs/step-1.0.0
         frame: !wcs/frame-1.1.0
@@ -192,7 +192,7 @@ slits:
                   - !transform/compose-1.1.0
                     forward:
                     - !transform/remap_axes-1.0.0
-                      inverse: !transform/identity-1.0.0 {n_dims: 2}
+                      inverse: !transform/identity-1.1.0 {n_dims: 2}
                       mapping: [0, 1, 0, 1]
                       name: input_mapping
                     - !transform/concatenate-1.1.0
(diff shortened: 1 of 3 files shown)


nden:
These are all changes to the data files. How were the files changed? (or I can run fitsdiff on them)
This is fine now for the test suite but in reality we will run into this problem and changing the data will not be feasible. We can suppress the warnings but I wonder if this will hide potential problems. OTH, if asdf guarantees backwards compatibility the warnings seem not that important.

drdavella:
The easiest way to see the changes is probably just to check out the branch and run `git diff` against `master`. Here's a fairly representative snippet from the full diff:
```
-        frame: !wcs/frame-1.0.0
+        frame: !wcs/frame-1.1.0
           axes_names: [x, y]
           name: detector
           unit: [!unit/unit-1.0.0 pixel, !unit/unit-1.0.0 pixel]
-        transform: !transform/compose-1.0.0
+        transform: !transform/compose-1.1.0
           domain:
           - !transform/domain-1.0.0 {includes_lower: true, lower: 0, upper: 1892}
           - !transform/domain-1.0.0 {includes_lower: true, lower: 0, upper: 57}
           forward:
-          - !transform/concatenate-1.0.0
+          - !transform/concatenate-1.1.0
             forward:
-            - !transform/shift-1.0.0 {offset: 0.0}
-            - !transform/shift-1.0.0 {offset: 0.0}
-          - !transform/identity-1.0.0 {n_dims: 2}
+            - !transform/shift-1.1.0 {offset: 0.0}
+            - !transform/shift-1.1.0 {offset: 0.0}
+          - !transform/identity-1.1.0 {n_dims: 2}
           name: dms2sca
       - !wcs/step-1.0.0
-        frame: !wcs/frame-1.0.0
+        frame: !wcs/frame-1.1.0
           axes_names: [x, y]
           name: sca
           unit: [!unit/unit-1.0.0 pixel, !unit/unit-1.0.0 pixel]
-        transform: !transform/compose-1.0.0
+        transform: !transform/compose-1.1.0
```

I'm inclined to think that enabling these warnings is very valuable for unit testing since it can expose latent issues (there were several within ASDF itself that only came to light when these warnings were first introduced). However, I understand that these warnings can probably be pretty annoying when working with real world data. I'll see if we can introduce some global configuration for warning levels.

nden:
Merging this PR as is, will have to think of a long-term solution.

drdavella:
I think the long term solution may be somewhere along the lines of https://github.com/spacetelescope/asdf/issues/313. **TL;DR** version mismatch warnings will be silenced by default.

PR #1155, opened 2017-08-29
Support skipping reference file prefetch, OutlierDetectionStep

jaytmiller:
This pull adds a class level flag to Step named `prefetch_references`. 

Subclasses of Step that wish to skip the reference file prefetch (pre-caching) should define prefetch_references=False.   Other classes don't need to define it,  the default will do.

Also defined prefetch_references=False for OutlierDetectionStep which was the original motivation for this feature and use case.

### DIFF
--- a/jwst/outlier_detection/outlier_detection_step.py
+++ b/jwst/outlier_detection/outlier_detection_step.py
@@ -36,6 +36,7 @@ class OutlierDetectionStep(Step):
         good_bits = integer(default=4)
     """
     reference_file_types = ['gain', 'readnoise']
+    prefetch_references = False
 
     def process(self, input):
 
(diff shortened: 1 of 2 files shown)


stsci-hack:
Running the regression tests manually with this PR implemented (in my own reproduction of the regression test environment) demonstrates that it does work as expected with no apparent side-effects or additional issues.  In fact, the run time for the test went from 759 seconds (over 12 minutes) to 149 seconds (~2.5 minutes).  
Based on this evaluation, I recommend that this PR be merged.

jdavies-st:
Looks great Todd.  Thanks!

stsci-hack:
Another pull request (#1174) can take advantage of this once it has been merged to dramatically cut the run-time of outlier_detection in the pipeline, and in the nightly regression tests.  Testing has not identified any side-effects of this change, so I am merging this into master.

Issue #1108, opened 2017-08-09
Jump and Ramp_fit steps don't handle subarray readnoise and gain ref files


hbushouse:
NIRSpec has begun delivering subarray-specific reference files for both readnoise and gain. The jump and ramp_fit steps are currently setup in a way that assumes those reference data arrays will always be full-frame and hence take a very simple approach to extracting a subarray from those reference arrays to match a science exposure subarray. These need to be updated to handle reference files that are themselves subarrays, in the way some of the other steps like dq_init and saturation handle subarray reference files.

hbushouse:
A typical error resulting from this problem is:

ERROR RUNNING STEP 'SloperPipeline':
    operands could not be broadcast together with shapes
    (2,12,256,2048) (0,2048) (2,12,256,2048)
Traceback (most recent call last):
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python2.7/site-packages/jwst-0.7.8rc1.dev307-py2.7-macosx-10.7-x86_64.egg/jwst/stpipe/cmdline.py", line 320, in step_from_cmdline
    step.run(*positional)
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python2.7/site-packages/jwst-0.7.8rc1.dev307-py2.7-macosx-10.7-x86_64.egg/jwst/stpipe/step.py", line 361, in run
    result = self.process(*args)
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python2.7/site-packages/jwst-0.7.8rc1.dev307-py2.7-macosx-10.7-x86_64.egg/jwst/pipeline/calwebb_sloper.py", line 106, in process
    input = self.jump(input)
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python2.7/site-packages/jwst-0.7.8rc1.dev307-py2.7-macosx-10.7-x86_64.egg/jwst/stpipe/step.py", line 361, in run
    result = self.process(*args)
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python2.7/site-packages/jwst-0.7.8rc1.dev307-py2.7-macosx-10.7-x86_64.egg/jwst/jump/jump_step.py", line 69, in process
    rej_thresh, do_yint, sig_thresh)
[...]

PR #1156, opened 2017-08-30
More subarray library function updates

hbushouse:
Fleshed out some more library functions for handling subarray extractions from reference file data. Also updated many (all?) of the steps that had been using private routines to use the new library functions. I believe they include enough error checking to handle all the bad reference files still lurking in CRDS.

All steps first call the function `ref_matches_sci` to see if the reference data match the science data and, if so, just use the reference data directly. This is more appropriate than testing whether the science data are full-frame or subarray (via `is_subarray`), because there are cases where science data can be subarray and yet still have matching (subarray) reference files and hence no extraction is needed. They have to match in more than just size. They must have an exact match in terms of the indexes of the pixels contained in the subarray.

The steps then call either the `get_subarray_data` or `get_subarray_model` function. `get_subarray_data` extracts a subarray from only the `data` attribute of the reference file model and returns that extracted data array. The `get_subarray_model` function extracts a subarray from all array attributes contained in the reference file model, creates a new subarray-based reference file model using those arrays, and returns the new model.

### DIFF
--- a/jwst/dq_init/dq_initialization.py
+++ b/jwst/dq_init/dq_initialization.py
@@ -11,6 +11,7 @@ import logging
 import numpy as np
 
 from .. import datamodels
+from ..lib import reffile_utils
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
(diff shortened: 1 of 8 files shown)


hbushouse:
Fixes #1108.

PR #1157, opened 2017-08-30
Updates for gain_scale, including docs

hbushouse:
Several more updates related to the new gain_scale step.

- Added docs for the gain_scale step
- Updated the pipeline/calwebb_sloper docs to include gain_scale and persistence
- Added a gain_scale.cfg file and added an entry for it to the calwebb_sloper.cfg file
- Modified the ramp_fit step to no longer save the rateints model on its own, but rather return it (along with the rate product model) to the caller. This also means the ramp_fit step no longer applies gain_scale to the rateints model.
- Modified the calwebb_sloper pipeline to receive the rateints model from ramp_fit, apply the gain_scale step to both the rate and rateints models (if both exist), and save the rateints model to an output file

This new arrangement, which has the ramp_fit step return both the rate and rateints models, is a much cleaner flow. Ramp_fit now only saves the optional products that it's capable of creating.

### DIFF
--- a/docs/jwst/gain_scale/arguments.rst
+++ b/docs/jwst/gain_scale/arguments.rst
@@ -1,0 +1,4 @@
+Arguments
+=========
+
+The gain_scale correction has no step-specific arguments.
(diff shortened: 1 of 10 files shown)


hbushouse:
Fixes all but one remaining item in #1103

Issue #1042, opened 2017-07-13
Copy slit-related keywords to extract_1d output


hbushouse:
The MultiSlitModels that serve as input to extract_1d contain slit-specific keywords as attributes of each slit instance (stored in the SCI extension header of FITS files). These meta attributes are currently defined in the multislit.schema.yaml file. Most of these meta attributes should be copied over to the MultiSpecModel created as output from extract_1d, in order to carry along useful info like source_id, etc. Right now the MultiSpecModel does not define these meta data. It should be enhanced to include them (or at least the ones that are relevant) and then have extract_1d populate them from the input model.

hbushouse:
The following are relevant entries from an email thread on this topic:

@nden:
PR #1031 creates a slit.schema.yaml. 
Can that be used for MultiSpecModel?
Does it need anything else?

@philhodge:
Four of the keywords in MultiSlitModel are SLTSTRT[12] and SLTSIZE[12], which are not relevant for 1-D extracted spectra because the extraction region can be smaller than those limits, and (because of Mihai's enhancements) the extraction region can have polynomial upper and lower boundaries.  The extraction region can even consist of multiple, disjoint regions.

For a MultiSpecModel, there should be info regarding the background region(s), if any.  Those can also have polynomial upper and lower boundaries.

@philhodge:
It seems to me that there are a few items in slit.schema.yaml that would not be relevant for the 1-D spectrum.  The wavelength is already in the MultiSpecModel table, but as a 1-D array.  I don't understand the distinction between wavelength_pointsource and wavelength_uniformsource; but in any case, if they were in the MultiSpecModel, I would expect to find them as table columns, not independent extensions.  (By the way, if these and the pathloss items are independent extensions in the FITS file, how do we associate them with the science data?  By matching name, or source_id, or something else?)  The xstart, xsize, ystart, and ysize I already mentioned as being not necessarily relevant for 1-D spectra.

hbushouse:
PR #1031 does indeed create a slit.schema.yaml, but it appears to include not only the meta data definitions, but also the data arrays normally used in a MultiSlitModel. For the MultiSpecModel we only need the meta data and perhaps not all of them. It has its own data structure, in the form of a table.

philhodge:
There are possible input models other than MultiSlitModel.  This is a complete list (at the current time):

* MultiSlitModel
* MultiProductModel
* ImageModel
* DrizProductModel
* CubeModel
* IFUCubeModel

[...]

PR #1161, opened 2017-08-31
Slit- and source-related keywords are copied to output.

philhodge:
Attributes such as the slit name, source name, and source location are now
copied from the input file (if it's a MultiSlitModel or MultiProductModel)
to the output MultiSpecModel file.  For other models (except CubeModel),
the slit name will be copied to the output, but other attributes are not
known or are not relevant, so they will not be assigned.  See issue #1042.


philhodge:
The one failed Travis test was a time-out.

PR #1162, opened 2017-08-31
Remove code that stores cube's WCS for background polynomials

mcara:
It seems that models that have other WCSes in their meta currently cannot be serialized (reported by @jemorrison @nden and @drdavella). Since saving cube's WCS in input image's meta was a feature designed for the future, we can safely remove it until issues with schemas and asdf are ironed out and/or until there is an express need to save this WCS.


mcara:
@nden Thanks!

PR #1164, opened 2017-08-31
Make apply_background_2d a public function

mcara:
I have renamed ``_apply_sky_2d`` to ``apply_background_2d`` and made it a public function. I have also added the following functionality to it:

1. It can take a single channel, a list of several channels, or None. ``None`` indicates that background should be applied for all channels.

2. ``subtract`` parameter can be used to indicate that background be subtracted from data or added back.

CC: @hbushouse @jemorrison

### DIFF
--- a/jwst/mrs_imatch/mrs_imatch_step.py
+++ b/jwst/mrs_imatch/mrs_imatch_step.py
@@ -10,6 +10,7 @@ JWST pipeline step for image intensity matching for MIRI images.
 from __future__ import (absolute_import, division, unicode_literals,
                         print_function)
 
+import six
 import numpy as np
 
 from .. stpipe import Step, cmdline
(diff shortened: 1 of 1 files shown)


jwst/mrs_imatch/mrs_imatch_step.py
```diff
@@ -163,7 +229,11 @@ def _apply_sky_2d(model2d, channel):
     bkg = np.polynomial.polynomial.polyval3d(r, d, l, c)
 
     # subtract background:
-    model2d.data[y, x] -= bkg
+    if subtract:
+        model2d.data[y, x] -= bkg
+        bkgmeta
```
hbushouse: Is there supposed to be more to this line, like modifying or using the value of bkgmeta?

mcara: @hbushouse You are right: this is a leftover from an earlier attempt to make this function set ``meta.background.subtracted`` - I gave up on this idea. Thank you for noticing this!

PR #1165, opened 2017-09-01
Update references to transform in yaml schemas

drdavella:
The references in these schemas should reflect the latest versions provided by ASDF.


nden:
Thanks @drdavella !

drdavella:
@nden there are still some data files with old references to `wcs` in them. They will show up with `git grep wcs-1.0.0`. I wasn't sure whether they were supposed to be changed or not, so I left them alone for now.

PR #1166, opened 2017-09-04
Ramp fit error prop

dmggh:
First implementation of error propagation in ramp fitting.


philhodge:
This looks good.  I think you should merge it.
Just a minor comment:  Function `calc_opt_fit` still returns `sig_slope` and `sig_intercept`, but these were calculated using the Fixsen weights, so they're not really useful.  This isn't a change to the code, but eventually (not now) these could be replaced or removed.

hbushouse:
Superseded by #1183. Closing.

PR #1167, opened 2017-09-05
Switched crds_client.reference_uri_to_cache_path() from AssertionError to CrdsError

jaytmiller:
Attempting to modify behavior with py.test and interpretation of assertions vs. exceptions.

### DIFF
--- a/jwst/stpipe/crds_client.py
+++ b/jwst/stpipe/crds_client.py
@@ -29,19 +29,17 @@
 """
 A client library for CRDS
 """
-import contextlib
-from os.path import dirname, join
 import re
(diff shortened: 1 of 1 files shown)


jaytmiller:
I also made a number of miscellaneous updates driven by pylint:

Reordered some imports
Removed unused imports
Added two docstrings
Changed CRDS imports to pass pylint

There are still a few pylint errors (about six) and warnings (bad var names) but nothing too confusing.   My apologies for the creep.

jdavies-st:
Thanks @jaytmiller.  Looks great!  I think this will help in testing.

Issue #312, opened 2016-09-30
Rule: MIRI Coronographic


stscieisenhamer:
Supersedes [TRAC issue 365](https://aeon.stsci.edu/ssb/trac/jwst/ticket/365)

stscieisenhamer:
# Implementation
# ToDo

# Constraints
**_TBD_**

stscieisenhamer:
# Resources
- [Problem Report #87294](https://www.ess.stsci.edu/prsystem/servlet/prbrowse/pr.87294): JWST: add PSF column (flag) to Asn Pool file to indicate exposures of this type

PR #1171, opened 2017-09-06
Implement Coronographic rules to create associations for calwebb_coron3

### LINKED ISSUE #311: Rule: NIRCam Coronographic
Supersedes [TRAC issue 365](https://aeon.stsci.edu/ssb/trac/jwst/ticket/365)

stscieisenhamer:
Resolves #311, #312


stscieisenhamer:
Also added AMI per issue #310

jwst/associations/lib/rules_level3.py
```diff
@@ -326,3 +326,110 @@ def __init__(self, *args, **kwargs):
...
+            },
+            'target': {
+                'value': None,
+                'inputs': ['targetid'],
+                'onlyif': lambda item: self.get_exposure_type(item) == 'science',
+                'force_reprocess': ProcessList.EXISTING,
+            }
+        })
+
+        # PSF is required
```
hbushouse: For AMI observations I believe that PSF exposures are not actually required, but are highly advantageous. The calwebb_ami3 pipeline has logic to handle the situation where there aren't any PSF exposures. It can still compute fringe params, etc. for the science target only. It's only coronagraphic that requires PSF exposures.

stscieisenhamer: Will handle!

PR #1173, opened 2017-09-06
Correct Level 2c filenames

### LINKED ISSUE #1159: Level-2c file names from calimage3 outlier_detection are incorrect
The latest version of the calimage3 and outlier_detection software is still producing level-2c files (the files that have their DQ arrays updated to contain the results of outlier_detection) with the old naming syntax, e.g.
```
2017-08-30 15:03:11,981 - stpipe.Image3Pipeline.outlier_detection - INFO - Step outlier_detection done
2017-08-30 15:03:11,982 - stpipe.Image3Pipeline - INFO - Writing Level 2c images with updated DQ arrays...
2017-08-30 15:03:11,983 - stpipe.Image3Pipeline - INFO - Step.save_model jw10002001001_01101_00006_nrcb2_cal-o001.fits
2017-08-30 15:03:13,037 - stpipe.Image3Pipeline - INFO - Step.save_model jw10002001001_01101_00006_nrcb4_cal-o001.fits
2017-08-30 15:03:13,873 - stpipe.Image3Pipeline - INFO - Step.save_model jw10002001001_01101_00006_nrcb1_cal-o001.fits
```
The latest agreed upon level-2c file name syntax is:

`jw<ppppp><ooo><vvv>_<gg><s><aa>_<eeeee>_<detector>_<AC_ID>_<suffix>.fits`

where `suffix` is `crf` if the level-2b file suffix was `cal`, and `crfints` if the level-2b file suffix was `calints`. So the order of appearance of the association candidate ID (AC_ID) and the suffix need to be swapped, and we need to use the new suffix.

Not sure if this is handled within calimage3 or in outlier_detection.

jdavies-st:
Make filename suffixes `<ASN_ID>_crf` or `<ASN_ID>_crfints` to indicate cosmic-ray flagged 2c files.

- `outlier_detection_step` now updates to COMPLETE when finished
- the pipelines only write out 2c files if `meta.cal_step.outlier_detection` is COMPLETE
- calwebb_coron3 now uses `stpipe` `self.log` instead of importing logging separately
- the pipelines now use `stpipe` `self.save_model` to write out datamodel results.  This method has its own logging, so extra logging in the pipeline is not needed.
- remove `mk_filename` and `mk_prodname` from calwebb_coron3, this being replaced by the built-in `stpipe` infrastructure methods `save_model` and `output_basename`.
- add some extra suffixes to `stpipe` SUFFIX_LIST

Resolves #1159

### DIFF
--- a/jwst/pipeline/calwebb_coron3.py
+++ b/jwst/pipeline/calwebb_coron3.py
@@ -16,10 +16,6 @@ from ..resample import resample_step
 
 __version__ = "0.7.1"
 
-# Define logging
-import logging
-log = logging.getLogger()
-log.setLevel(logging.DEBUG)
 
 class Coron3Pipeline(Pipeline):
     """
@@ -33,19 +29,21 @@ class Coron3Pipeline(Pipeline):
     """
 
     spec = """
+        suffix = string(default='i2d')
     """
 
     # Define aliases to steps
     step_defs = {'stack_refs': stack_refs_step.StackRefsStep,
                  'align_refs': align_refs_step.AlignRefsStep,
                  'klip': klip_step.KlipStep,
-                 'outlier_detection': outlier_detection_stack_step.OutlierDetectionStackStep,
+                 'outlier_detection':
+                     outlier_detection_stack_step.OutlierDetectionStackStep,
                  'resample': resample_step.ResampleStep
                  }
 
     def process(self, input):
 
-        log.info('Starting calwebb_coron3 ...')
+        self.log.info('Starting calwebb_coron3 ...')
 
         # Load the input association table
         with open(input, 'r') as input_fh:
@@ -60,13 +58,13 @@ class Coron3Pipeline(Pipeline):
 
         # Make sure we found some PSF and target members
         if len(psf_files) == 0:
-            log.error('No reference PSF members found in association table')
-            log.error('Calwebb_coron3 processing will be aborted')
+            self.log.error('No reference PSF members found in association table')
+            self.log.error('Calwebb_coron3 processing will be aborted')
             return
 
         if len(targ_files) == 0:
-            log.error('No science target members found in association table')
-            log.error('Calwebb_coron3 processing will be aborted')
+            self.log.error('No science target members found in association table')
+            self.log.error('Calwebb_coron3 processing will be aborted')
             return
 
         # Assemble all the input psf files into a single ModelContainer
@@ -76,15 +74,13 @@ class Coron3Pipeline(Pipeline):
             psf_models.append(psf_input)
             psf_input.close()
 
-        # Call the stack_refs step to stack all the PSF images into
-        # a single CubeModel
+        # Stack all the PSF images into a single CubeModel
         psf_stack = self.stack_refs(psf_models)
         psf_models.close()
 
         # Save the resulting PSF stack
-        output_file = mk_prodname(self.output_dir, prod['name'], 'psfstack')
-        log.info('Saving psfstack file %s', output_file)
-        psf_stack.save(output_file)
+        psf_stack.meta.filename = prod['name']
+        self.save_model(psf_stack, suffix='psfstack')
 
         # Call the sequence of steps align_refs, klip, and outlier_detection
         # once for each input target exposure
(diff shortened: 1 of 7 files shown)


jdavies-st:
Ok, looking at `calwebb_coron3`, it looks like the calints changes need to go in there, not `calwebb_image3`.  And it looks like the coron pipeline needs to take advantage of the @stscieisenhamer stpipe refactor for file I/O.

Ok, another commit coming.

jdavies-st:
A further question @hbushouse, is the 2c file saved for the coron pipeline supposed to be the SCIENCE or PSF files?   Currently it looks like [it is the PSF](https://github.com/STScI-JWST/jwst/blob/master/jwst/pipeline/calwebb_coron3.py#L132).

jdavies-st:
Ah, nevermind.  I see it is the PSF-subtracted SCIENCE frame that is saved.

jwst/pipeline/calwebb_image3.py
```diff
@@ -56,10 +56,17 @@ def process(self, input):
...
-        # Check if input is multiple exposures, as required by some steps
+        # Check if input is single or multiple exposures
         is_container = isinstance(input_models, datamodels.ModelContainer)
         if is_container and len(input_models.group_names) > 1:
 
+            # Set up Level 2c suffix to be used later
+            suffix_2c_end = 'crf'
+            if isinstance(input_models[0], datamodels.CubeModel):
+                suffix_2c_end = 'crfints'
+            asn_id = input_models.meta.asn_table.asn_id
```
hbushouse: Is the "asn_id" here the same as the association candidate ID that's passed along from PPS through the asn pool? They are entries like "o001" or "a3001".

jdavies-st: Yes, that's the one.

jdavies-st:
@stsci-hack @hbushouse I've made changes to this PR to fix the issues.  Can you have a look?  I've updated the description at the top of this PR too.

jwst/pipeline/calwebb_coron3.py
```diff
@@ -16,10 +16,6 @@
 
 __version__ = "0.7.1"
 
-# Define logging
-import logging
-log = logging.getLogger()
-log.setLevel(logging.DEBUG)
```
hbushouse: I wonder what the default Level is set to in the stpipe Step class. We always explicitly set it to DEBUG in the individual step and pipeline classes to make sure that the logger will pass DEBUG level messages when requested to do so via the verbose argument. If the default in the stpipe Step class is INFO, for example, then the pipelines and steps may not pass DEBUG level messages. The setLevel seems to act as an "upper limit" as to what can get passed.

jwst/pipeline/calwebb_coron3.py
```diff
@@ -76,44 +74,37 @@ def process(self, input):
...
-            filename = mk_filename(self.output_dir, target_file, 'psfalign')
-            log.info('Saving psfalign file %s', filename)
-            psf_aligned.save(filename)
+            self.save_model(psf_aligned, suffix='psfalign')
 
             # Call KLIP
-            log.debug('Calling klip for member %s', target_file)
-            #psf_sub, psf_fit = self.klip(target_file, psf_aligned)
+            self.log.debug('Calling klip for member %s', target_file)
             psf_sub = self.klip(target_file, psf_aligned)
```
hbushouse: Is klip returning 2 data models (psf_sub and psf_fit)? If so, what happens on this line when you're capturing only 1 returned object? Does the 2nd one just vanish into the ether?

jwst/pipeline/calwebb_coron3.py
```diff
@@ -76,44 +74,37 @@ def process(self, input):
...
             psf_aligned.close()
 
             # Save the psf subtraction results
-            filename = mk_filename(self.output_dir, target_file, 'psfsub')
-            log.info('Saving psfsub file %s', filename)
-            psf_sub.save(filename)
+            self.save_model(psf_sub, suffix='psfsub')
```
hbushouse: Does the psf_sub data model have the proper filename value in its meta data, equivalent to 'target_file'?

jwst/pipeline/calwebb_coron3.py
hbushouse: Never mind. I see that I in fact was the one to update the klip step some time ago so that it only returns a single data model and old line 105 was also commented out at that time. So new line 99 is correct.

jwst/pipeline/calwebb_coron3.py
jdavies-st: The default setup in `step.py` allows passing of DEBUG level messages.  I've had `Image3Pipeline` running like this for a long time.  The steps within that pipeline have the same setup.  The logging setup is inherited from the parent class.

jwst/pipeline/calwebb_coron3.py
jdavies-st: Yes, it gets it from `self.output_basename`.

jdavies-st: Correction.  `meta.filename` is populated with the filename of the input file in the association, but of course if you supply the suffix to `save_model`, the old suffix is automatically replaced with the new one.

jwst/pipeline/calwebb_coron3.py
```diff
@@ -76,44 +74,37 @@ def process(self, input):
...
-        # a single CubeModel
+        # Stack all the PSF images into a single CubeModel
         psf_stack = self.stack_refs(psf_models)
         psf_models.close()
 
         # Save the resulting PSF stack
-        output_file = mk_prodname(self.output_dir, prod['name'], 'psfstack')
-        log.info('Saving psfstack file %s', output_file)
-        psf_stack.save(output_file)
+        psf_stack.meta.filename = prod['name']
```
stsci-hack: The original code (`output_file = mk_prodname()`) defines an output filename based on `output_dir` to support writing out the product in another directory.  This seems to be missing from the change where meta.filename is simply set to the rootname without support for specifying an output directory.  Does the `.save_model()` method handle this now in a hidden manner or does this need to be more explicitly defined?  This same question applies for all remaining similar changes...

jwst/pipeline/calwebb_tso3.py
```diff
@@ -83,63 +79,62 @@ def process(self, input):
...
+
             # Working with spectroscopic TSO data...
             # define output for x1d (level 3) products
-            x1d_models = datamodels.MultiSpecModel()
-            x1d_models.update(input_asn)
-
-            # Create name for extracted white-light (Level 3) products
-            phot_tab_name = "{}_whtlt.ecsv".format(product_name)
+            x1d_result = datamodels.MultiSpecModel()
+            # TODO: check to make sure the following line is working
```
stsci-hack: We can probably remove this comment if it is working now...

jwst/pipeline/calwebb_coron3.py
jdavies-st: Yes, @stscieisenhamer refactor of `stpipe` put all the path control internally into the pipeline infrastructure.

PR #1174, opened 2017-09-06
Fix outlier step

stsci-hack:
Replace #1145 with this revised code which:
- reverts to using get_reference_file() CRDS call as default in outlier_detection
- replaces outlier_detection_tso_step with a more generically named outlier_detection_stack_step for use in CALTSO3 as well as CALCORON3 and any others which do not require (by default) resampling
- relies on #1155 to help make use of CRDS more efficient by skipping prefetch for outlier_detection with the assumption (which we know is not always correct) that the inputs have correctly specified reference file names/URI's
- retains logic to only make minimum number of calls to CRDS for all types of outlier_detection steps

This will affect development being done primarily by @jdavies-st.


jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -36,7 +36,8 @@ class OutlierDetectionStep(Step):
         good_bits = integer(default=4)
     """
     reference_file_types = ['gain', 'readnoise']
-
+    prefetch_references = False
```
jdavies-st: I think we want this to set `prefetch_references = True` for the general case of outlier detection, yes?

Issue #596, opened 2017-01-13
Task: Error propagation


hbushouse:
The JCCWG is close to agreement on a baseline scheme for tracking and propagating various error components through pipeline processing. Their work on the subject is being tracked at https://confluence.stsci.edu/display/JWSTCC/Optimal+Error+Initialization and discussion from a recent meeting on the subject is available at https://confluence.stsci.edu/display/JWSTCC/2017-01-10+Meeting+notes

Implementation of some form of error propagation is required by DMS Level 4 requirement 464.

hbushouse:
Documenting some email discussion, based on a conversation between Karl and Phil in July. He wants to support the following approach:

Store variances rather than error estimates.

At the very least keep read noise and photon (Poisson) noise.

The ramp_fit step should be modified to do the following:

(1) Fit the slope using ordinary least squares with optimal weighting.  [Note that this may actually be done multiple times for each pixel, once for each segment between jumps.  This is a significant complication.]  The weighting (i.e. the weight as a function of group number) depends only on the signal-to-noise ratio of the data.  The weighting function has an expression raised to a power, and that power is obtained from a table using the SNR.  Initially, use all available variance estimates (currently only read noise and photon noise) to compute the signal-to-noise ratio.  Do the fit; save the value of the slope. Also compute the formal variance (i.e. based on the input variances, not based on the scatter of the data about the fitted line), but this is only used for weighting the segments when averaging to get the slope for each integration and for weighting integrations to get the overall slope for the exposure.

[So far, I think this is what is currently done, but using only the read noise and photon noise.]

Do step (2) for each of the available variance estimates.

(2) Re-compute the slope using OLS and optimal weighting as in step (1), but when computing the signal-to-noise ratio use only the current variance estimate.  Ignore the fitted slope but compute and save the formal variance.

The variances (so far) are additive.  The full variance of the slope should be the sum of all these individual variances from iterating step (2); this "full variance" is the variance that we got in step (1) and discarded.

A pipeline or step that combines or further processes slope images should compute the variance of the output, and that pipeline or step would (with the modification outlined above) have a choice about which variances in (or with) the _rate files to use when computing the output variance.

Complications:

The read noise is no problem.  Note that it's a 2-D array, dependent only on the detector.  The photon variance array might be 3-D:  integrations, rows, columns.  Other variance arrays might be 4-D.  Storing these arrays in the _ramp file or _rate file would be pretty unweildy.  (Consider TSO data.)

[...]

PR #1183, opened 2017-09-08
Ramp_fitting error propagation

hbushouse:
New version of changes committed in #1166 in order to resolve conflicts that appeared (somewhat mysteriously) in that PR. This is a direct copy of the files committed in that PR, which adds computation of slope variances based on read noise and Poisson noise to the ramp_fit step.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -94,13 +94,14 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
                                 readnoise_model, gain_model)
         opt_model = None
     else:
-        new_model, int_model, opt_model = ols_ramp_fit(model,
+        new_model, int_model, opt_model, var_slope_r, var_slope_p,\
+        var_r_s_4d, var_p_s_4d = ols_ramp_fit(model,\
                                 buffsize, save_opt,
                                 readnoise_model, gain_model, weighting)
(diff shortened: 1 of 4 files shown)


hbushouse:
Addresses #596.

dmggh:
These differences appear correct, including the unrelated differences in main.rst, so I am in favor of a merge. However, I am unable to merge them right now, as my laptop VPN does not seem to recognize the wifi of my hotel (which does not have ethernet as an option). Also, although my phone (from which I'm sending this) allows me to log on to github, github will not allow a merge from my phone.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -1756,14 +1796,21 @@ def calc_opt_fit(nreads_wtd, sumxx, sumx, sumxy, sumy):
     sig_intercept = (sumxx / denominator)**0.5
     sig_slope = (nreads_wtd / denominator)**0.5
 
+    # variance of slope due to Poisson noise only, and
+    # variance of slope due to Read noise only
+    sig_slope_p = slope/(mask_2d.sum(axis=0) - 1.)
+    sig_slope_r = nreads_wtd/denominator
```
philhodge: `nreads_wtd` is the sum of the Fixsen weights (I think also divided by the read noise).  In this case, the variance computed from just read noise should be `(ngroups / readnoise^2) / denominator`.

Issue #1184, opened 2017-09-08
Enable resampling of coron3 output data


stsci-hack:
The `Coron3Pipeline` step includes outlier_detection to generate updated PSF-subtracted images.  These products are then expected (based on JWST CWG outline of CALCORON3) to be combined into a single final product, something which would be done with the resample step. The current default configuration file for the `Coron3Pipeline` skips the resample step.

stsci-hack:
Investigating how to turn on resampling for the `Coron3Pipeline` output product revealed that it can NOT be turned on until support for this observation type has been added to `assign_wcs`.  Adding this support will require obtaining updated WCS information for the observing mode from the instrument team.  

@hbushouse and @nden will need to report on when the update has been implemented in `assign_wcs` in order to finally enable the full range of planned initial processing of coronographic data in the pipeline.

hbushouse:
I believe that for MIRI the already delivered distortion reference file for the imager detector is applicable to the entire detector FOV, which includes the areas of the coronagraphic masks and Lyot stops. Is that sufficient for assign_wcs to work with MIRI coronagraphic exposures?

What's the situation with NIRCam? Do the distortion reference files already delivered for imaging modes not cover the areas/subarrays of coronagraphic exposures?

nden:
@stsci-hack @hbushouse `assign_wcs` for `NRC_CORON` mode was enabled in #1193. Reference files have been delivered though it may take a day to show up in CRDS.

hbushouse:
Fixed by #1185 and #1433

PR #1185, opened 2017-09-08
Update coron3

stsci-hack:
Minimal changes to Coron3Pipeline:
- enable outlier_detection by default in config file
- revised logic to correctly handle output when resampling of final product does get enabled

Supports #1184.
Resampling can not be turned on at this point until support for this mode has been added to `assign_wcs`.


jwst/pipeline/calwebb_coron3.py
```diff
@@ -142,18 +142,22 @@ def process(self, input):
...
-        result.meta.asn.pool_name = asn['asn_pool']
-        result.meta.asn.table_name = input
+            result.update(resample_input[0])
+            # The resample step blends headers already...
+            log.debug('Blending metadata for {}'.format(output_file))
+            blend.blendfitsdata(targ_files, result)
+            result.meta.asn.pool_name = asn['asn_pool']
+            result.meta.asn.table_name = input
+
         result.meta.cal_step.outlier_detection = 'COMPLETE'
```
hbushouse: Was this hardwired here because it was assumed that outlier_detection was being skipped? If so, now that outlier_detection has been enabled, the status should be set within outlier_detection and hence shouldn't be needed here anymore.

jwst/pipeline/calwebb_coron3.py
```diff
@@ -142,18 +142,22 @@ def process(self, input):
...
         result.meta.cal_step.resample = 'COMPLETE'
         result.meta.model_type = 'DrizProductModel'
```
hbushouse: Is this necessary? Does it get set when the model is created in line 151?

jwst/pipeline/calwebb_coron3.py
```diff
@@ -142,18 +142,22 @@ def process(self, input):
 
         # Call the resample step to combine all the psf-subtracted target images
         result = self.resample(resample_input)
+        output_file = mk_prodname(self.output_dir, prod['name'], 'i2d')
```
hbushouse: Someday this should be updated to use the library and/or datamodels functions that @stscieisenhamer implemented for use in other pipelines like calspec2 and calspec3, so that we can do away with standalone functions like mk_prodname. But that's not super important right now.

jwst/pipeline/calwebb_coron3.py
jdavies-st: It turns out `outlier_detection_step` does not update to "COMPLETE", so this is a change I've put into #1173 with an upcoming commit and merge.  We need `outlier_detection_step` to do this in order for the pipeline to know if it should write out the Level 2c product.  That logic is also coming in #1173.

jwst/pipeline/calwebb_coron3.py
jdavies-st: This is coming in #1173.  Changes already made.  Just need to be tested and pushed to that PR.

PR #1187, opened 2017-09-08
Change creation date code

bernie-simon:
Datamodels adds the creation date of a file into the metadata when saving it. The current code adds the date to the metadata and then updates the format. This assumes that saving to the metadata does not  change the format. It's safer to update the format of the date and then add it to the metadata.


philhodge:
These changes to the on_save method look good.

Note that lines 192 - 195 are very similar to the ones in on_save.

hbushouse:
👍

bernie-simon:
Thanks for noticing the same code is in __init__. Put the same change there.

philhodge:
Thanks for updating it, @bernie-simon .

Issue #1191, opened 2017-09-11
The output model for 1-D extracted spectra should be MultiSpecModel


philhodge:
The 1-D extracted spectra from the calwebb_tso3 pipeline should be concatenated to an output `MultiSpecModel`, rather than to a `ModelContainer`.  The critical difference is that when the data are written to a FITS file, only one such file should be written.

philhodge:
Closed by #1192.

PR #1192, opened 2017-09-11
The output model for 1-D extracted spectra should be MultiSpecModel

philhodge:
The 1-D extracted spectra from the calwebb_tso3 pipeline are now
concatenated to an output MultiSpecModel, rather than to a ModelContainer.
See GitHub issue #1191.
Also, trailing blanks were deleted.


philhodge:
One test timed out -- merging.

PR #1203, opened 2017-09-13
Update ami3

stsci-hack:
These updates implement blended headers for the Level 3 products generated by the `Ami3Pipeline`.  Specifically, the averaged target product with the `_amiavg` suffix, and the PSF normalized target product with the `_aminorm` suffix have been processed with `blendheaders`.  

This required an update to the `blendfitsdata()` function in `resample.blend` in order to correctly combine products which already have blended headers from previous processing steps.

### DIFF
--- a/jwst/pipeline/calwebb_ami3.py
+++ b/jwst/pipeline/calwebb_ami3.py
@@ -10,7 +10,7 @@ from .. import datamodels
 from ..ami import ami_analyze_step
 from ..ami import ami_average_step
 from ..ami import ami_normalize_step
-
+from ..resample import blend
 
 __version__ = "1.2"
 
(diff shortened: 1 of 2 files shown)


jwst/pipeline/calwebb_ami3.py
```diff
@@ -10,7 +10,7 @@
 from ..ami import ami_analyze_step
 from ..ami import ami_average_step
 from ..ami import ami_normalize_step
-
+from ..resample import blend
```
hbushouse: We really should put the blend stuff in its own package, or the lib area, in the future.

jwst/pipeline/calwebb_ami3.py
```diff
@@ -97,9 +97,13 @@ def process(self, input):
...
-                output_file = mk_prodname(self.output_dir, prod['psf_name'], 'amiavg')
-                self.log.info('Saving averaged PSF results to %s', output_file)
-                psf_avg.save(output_file)
+                psf_avg_output_file = mk_prodname(self.output_dir, prod['psf_name'], 'amiavg')
+                self.log.info('Saving averaged PSF results to %s', psf_avg_output_file)
+                # Perform blending of metadata for all inputs to this output file
+                self.log.info('Blending metadata for averaged target {}'.format(psf_avg_output_file))
+                blend.blendfitsdata(psf_files, psf_avg)
+
+                psf_avg.save(psf_avg_output_file)
```
hbushouse: So you're able now to run blend directly on the data model, before it's been saved to a FITS file? In the past the FITS file had to be created first and then run blend on it.

jdavies-st: This functionality was added several months ago by @stsci-hack

Issue #1211, opened 2017-09-14
Nirspec slitlets need to include shutter_state


nden:
A slitlet can be non-contiguous. In this case the closed shutters should also be included in the slitlet (though there won't be any data at that place). The closed shutters are currently included when the size of the slit is computed. However, we record only the number of open shutters, `nshutters`,  and the information about possible closed shutters within the slit is lost. `nirspec.get_msa_open_slits` should be modified to include the ~~`shutter_id` of all open shutters~~ state of all shutters within a slit. This will allow to identify if there are and how many  closed shutters are in a slit. This information is needed for the bar shadow correction. 
In this case `nshutters` is not necessary and should be removed.

_Edit:
The state is recorded in a string where "1" indicates an open shutter, "0" - a closed one and "x" - the main shutter. The allows the string to be recorded in a FITS header.__

nden:
@stscirij

nden:
Closed by #1214

PR #1214, opened 2017-09-15
add Slit.shutter_state

### LINKED ISSUE #1213: bug in ymin, ymax slit computation for all background shutters
For a slitlet with all background shutters we still need to compute a "main" shutter, in order to compute `ymin`, `ymax` and the `shutter_id` in the MSA metrology file. The main shutter is computed incorrectly in assign_wcs as the index of the shutter in the list of open shutters. Instead it should return the column of the shutter. This is only a bug when the entire slitlet is background.

nden:
Closes #1213, #1211


stscirij:
Looks good to me!

PR #1216, opened 2017-09-19
Add crds context test check

stscieisenhamer:
New testing helper decorator `require_crds_context` for tests the require a certain level of crds context.


stscieisenhamer:
Linux passed and just a test change, so merging.  osx will get there this week maybe?

PR #1218, opened 2017-09-19
Fix bug in Step where ModelContainer meta.ref_file wasn't handled

### LINKED ISSUE #1217: stpipe.Step writing to meta.ref_file with ModelContainer fails
@jemorrison reports that an stpipe Step fails when trying to write out a `ModelContainer`.  This step takes a `DataModel` as input and produces a `ModelContainer` with 2 `CubeModel` instances as output.  The failure:

```
  File "jwst/datamodels/properties.py", line 210, in __getattr__
    val = self._instance[attr]
KeyError: 'ref_file'

During handling of the above exception, another exception occurred:

  File "test_compare_methods.py", line 19, in <module>
    config_file=config)
  File "jwst/stpipe/step.py", line 501, in call
    return instance.run(*args)
  File "jwst/stpipe/step.py", line 398, in run
    if hasattr(result.meta.ref_file, ref_name):
  File "jwst/datamodels/properties.py", line 213, in __getattr__
    raise AttributeError("No attribute '{0}'".format(attr))
AttributeError: No attribute 'ref_file'
```

Basically, a ModelContainer does not have a `meta.ref_file` attribute, though it's members do.  The code here:

https://github.com/STScI-JWST/jwst/blob/master/jwst/stpipe/step.py#L394

checks to see if the input is a `ModelContainer`, not he output.  @jemorrison's corner case leads me to think that we should be checking if the output is a `ModelContainer`, not the input.

Further, it would be good to have `meta.ref_file` updated in a `ModelContainer`'s members, just as if a step produces more than one result, we return a tuple of results with each result's `meta.ref_file` info updated.

jdavies-st:
Fixes a bug in the Step corner case where input is a single `DataModel` but output is a `ModelContainer`.  Now each member of `ModelContainer` result has its `meta.ref_file` updated with the step-specific reference files when the step finishes.

Resolves #1217.

### DIFF
--- a/jwst/stpipe/step.py
+++ b/jwst/stpipe/step.py
@@ -386,12 +386,12 @@ class Step(object):
             result_return = result
 
             # Update meta information
-            if not isinstance(result, (list, tuple)):
+            if not isinstance(result, (list, tuple, datamodels.ModelContainer)):
                 results = [result]
             else:
                 results = result
 
-            if len(self._reference_files_used) and not self._is_container(args[0]):
+            if len(self._reference_files_used):
                 for result in results:
                     if isinstance(result, datamodels.DataModel):
                         for ref_name, filename in self._reference_files_used:


jaytmiller:
I looked this over and agreed with James that this is probably the right fix to add model container support to reference file output to models.   <thumbs up shit eating grin emoticon>

Issue #1221, opened 2017-09-19
Fix outlier_detection output of intermediate results


stsci-hack:
The outlier_detection step includes the option to write out the intermediate products.  Unfortunately, using that option results in the following (partial) traceback:

```
/user/hack/dev/release/stsci_python/pypi/jwst/git/jwst/jwst/outlier_detection/outlier_detection.py in do_detection(self)
    149             if save_intermediate_results:
    150                 log.info("Writing out BLOT images...")
--> 151                 blot_models.save()
    152         else:
    153             # Median image will serve as blot image

TypeError: save() takes at least 2 arguments (1 given)

```
This traceback came from trying to write out the blot image, with a similar error reported earlier when trying to write out the single drizzled images.  

This does NOT need to be fixed in order for the pipeline to pass I&T, since this option is never used in any pipeline processing.  However, it does help in debugging the pipeline on occasion.

stsci-hack:
This should be fully addressed by #1222 once it gets successfully reviewed and merged.

PR #1222, opened 2017-09-19
Define intermediate product outputs correctly

stsci-hack:
This set of changes defines correct output names (or at least full FITS filenames) for the intermediate products generated by the outlier_detection step.  It then writes them out to disk for as requested by the user.  

The use of `model.save()` was necessary since the `outlier_detection` class was not a sub-class of `Step`, and therefore did not have access to the `step.save_model()` method used elsewhere.  It also was not entirely appropriate, since these products are not officially supported/defined by the association naming rules anyway.  

This should address #1221.

### DIFF
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -119,10 +119,10 @@ class OutlierDetection(object):
             sdriz.do_drizzle(**pars)
             drizzled_models = sdriz.output_models
             for model in drizzled_models:
-                model.meta.filename += ".fits"
-            if save_intermediate_results:
-                log.info("Writing out resampled exposures...")
-                drizzled_models.save()
+                model.meta.filename += "_i2d.fits"
+                if save_intermediate_results:
+                    log.info("Writing out resampled exposures...")
+                    model.save(model.meta.filename)
         else:
             drizzled_models = self.input_models
             for i in range(len(self.input_models)):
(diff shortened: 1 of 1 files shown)


jwst/outlier_detection/outlier_detection.py
```diff
@@ -119,10 +119,10 @@ def do_detection(self):
             sdriz.do_drizzle(**pars)
             drizzled_models = sdriz.output_models
             for model in drizzled_models:
-                model.meta.filename += ".fits"
-            if save_intermediate_results:
-                log.info("Writing out resampled exposures...")
-                drizzled_models.save()
+                model.meta.filename += "_i2d.fits"
```
hbushouse: Is it safe to just append the new suffix, rather than do a replacement of the old suffix? Or do you know that meta.filename will only ever be a "bare" base name?

jdavies-st: The file naming in outlier detection needs to be refactored.  Right now, a bunch of naming decisions (and appropriate suffixes) get done in subroutines or subroutines of subroutines, and really, they should all be happening in `OutlierDetectionStep.process` using the standard `Step` methods.

It's a bit messy, and the refactor will take some effort, so I would suggest we do this when we have some spare time in the near future.

hbushouse: Agreed. If it ain't broke (right now), don't fix it (in B7.1). But be sure to keep an issue open so that we don't forget about it.

jdavies-st: I will open an issue then.

I've already refactored outlier_detection once.  Always fun to do it again.  Maybe the 2nd time we can do it with propagated error arrays.

hbushouse: 👍 for propagated error arrays

Issue #1199, opened 2017-09-13
Updated FITS HDU names in guide star data models


hbushouse:
The EXTNAME values that SDP is using for the various types of special table extensions in guide star data products are different that what's currently defined in our guider raw and cal data models. We need to update our data models to match the SDP values, otherwise those extensions are found when loading a file into a data model and hence don't get propagated to the output products.

hbushouse:
Closed by #1226.

PR #1226, opened 2017-09-21
Update handling of GS table extensions

hbushouse:
SDP is changing some of the definitions of the table extensions contained in FGS guider data products, so this updates our guider data models to stay in sync. It also updates updates the guider_cds step to copy over the table data from the input raw data model to the output cal data model, so that the table extensions get carried over to our level-2 products.


hbushouse:
Fixes #1199.

Issue #672, opened 2017-02-01
Task: NIRSpec pathloss bar shadow correction


hbushouse:
An additional component of the pathloss correction for NIRSpec MSA observations is the bar shadow correction, which accounts for the shadowing caused by the physical borders (bars) that separate one MSA shutter from another. The correction is a function of wavelength and the location of a pixel relative to the center of a shutter in the spatial direction.

This correction is cited on the JCCWG calspec2 pathloss correction confluence page at: https://confluence.stsci.edu/display/JWSTCC/Vanilla+Path-Loss+Correction

An initial set of reference files and a describing document were provided in NIRSpec CDP-3. The document is available at 

/grp/jwst/wit4/nirspec/CDP3/05_Other_Calibrations/5.0_Documents/ESA-JWST-SCI-NRS-TN-2016-016.pdf 

and the references files are available at 

/grp/jwst/wit4/nirspec/CDP3/05_Other_Calibrations/5.3_BarShadow/referenceFilesBS-20160401/

hbushouse:
Fixed by #1230 and #1270.

PR #1230, opened 2017-09-22
Barshadow

stscirij:
Initial delivery of barshadow code and BarshadowModel datamodel code
Still needs some work, but the datamodel should be good. Print statements will be removed when they aren't needed any more


hbushouse:
Addresses #672

jwst/barshadow/__init__.py
```diff
@@ -0,0 +1,4 @@
+from __future__ import absolute_import
+from .barshadow_step import BarShadowStep
+
+__version__ = '0.7.0'
```
hbushouse: Minor nit, but let's set this to 0.7.1.

jwst/barshadow/bar_shadow.py
```diff
@@ -0,0 +1,440 @@
...
+                    nonnan = np.where(~np.isnan(yrow))
+                    ymin = yrow[nonnan].min()
+                    ymax = yrow[nonnan].max()
+                    wmin = wcol[nonnan].min()
+                    wmax = wcol[nonnan].max()
+                    print(ymin, ymax, wmin, wmax, shutter_status)
+                    #   Interpolate the bar shadow correction for non-Nan pixels
+                    correction = interpolate(yrow, wcol, shadow)
+                    #   Divide the data by the correction
+                    slitlet.data = slitlet.data / correction
```
hbushouse: Are we supposed to actually be applying the correction to the data within the step or just attaching the correction values to the data model (to be applied at a later time)?

jwst/barshadow/bar_shadow.py
```diff
@@ -0,0 +1,440 @@
...
+                    # Add the correction array to the datamodel?
```
hbushouse: So the correction array is not yet added to the datamodel?

jwst/barshadow/barshadow_step.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    reference_file_types = ['barshadow']
+
+    def process(self, input):
+
+        # Open the input data model
+        with datamodels.open(input) as input_model:
+
+            # Get the name of the bar shadow reference file to use
```
hbushouse: If this only gets applied to NRS_MSASPEC mode, should we have a check up here at the very front that skips the step if the data are not from that mode? When run from the calspec2 pipeline the pipeline module can have the logic to only call the step for that exp_type, but we should also have a check in the step to prevent a user from applying just that step to non-MSA exposures.

jwst/barshadow/bar_shadow.py
```diff
@@ -0,0 +1,440 @@
...
+# of a set of slit[].nshutters in a vertical line
+#
+# Reference file information is a 1x1 ref file and a 1x3 ref file
+# Both the 1x1 and 1x3 ref files are 1001 pixel high and go from
+# -1 to 1 in their Y value WCS.  
+# 
+# 
+    exp_type = input_model.meta.exposure.type
+    log.info(exp_type)
+    #
```
hbushouse: Maybe add a check at this point that skips (returns) right away if exp_type is not NRS_MSASPEC.

jwst/barshadow/barshadow_step.py
```diff
@@ -0,0 +1,50 @@
...
+            instrument = input_model.meta.instrument.name
+            # Open the barshadow ref file data model
+            barshadow_model = datamodels.BarshadowModel(self.barshadow_name)
+
+            # Do the bar shadow correction     
+            result = bar_shadow.do_correction(input_model, barshadow_model)
+
+            barshadow_model.close()
+
+        return result
```
jdavies-st: You might want to add at the end of this step something like
```
if __name__ == '__main__':
    cmdline.step_script(BarShadowStep)
```
if this step can be called from the command line.

@hbushouse , should the cmdline be on all our steps?  I see it is on all our pipelines and many of our steps, but not all.

jwst/barshadow/bar_shadow.py
hbushouse: Or alternatively just move the first 3 lines of code below to be inside of the "if exp_type" block, so that we don't waste time doing things that aren't necessary.

jwst/barshadow/barshadow_step.py
```diff
@@ -0,0 +1,50 @@
+from ..stpipe import Step
```
jdavies-st: `from ..stpipe import Step, cmdline`

if the other changes are needed below.

jwst/datamodels/__init__.py
```diff
@@ -41,6 +41,7 @@
 from .model_base import DataModel
 from .amilg import AmiLgModel
 from .asn import AsnModel
+from .barshadow import BarshadowModel
```
jdavies-st: You should add `BarShadowModel` to `__all__` below as well.

hbushouse: And add it in the proper alphabetical order!

jwst/barshadow/barshadow_step.py
hbushouse: This was a practice started many years ago in the early days of step development, largely before stpipe/strun was in use and hence allowed for the step to be executed directly from the command-line. It's really of little use anymore and I'm not aware of anyone who makes any regular use of it, so it's not required at all (IMHO).

jwst/barshadow/barshadow_step.py
jdavies-st: Is there a `meta.cal_step.barshadow` keyword that can be set to "SKIP" or "COMPLETE"? in the step?

jwst/barshadow/barshadow_step.py
jdavies-st: Ah, ok.  Right now 29 of our steps use this `cmdline.step_script` call.  Maybe we should do some purging?

jwst/barshadow/bar_shadow.py
```diff
@@ -0,0 +1,440 @@
...
+                    ymax = yrow[nonnan].max()
+                    wmin = wcol[nonnan].min()
+                    wmax = wcol[nonnan].max()
+                    print(ymin, ymax, wmin, wmax, shutter_status)
+                    #   Interpolate the bar shadow correction for non-Nan pixels
+                    correction = interpolate(yrow, wcol, shadow)
+                    #   Divide the data by the correction
+                    slitlet.data = slitlet.data / correction
+                    # Add the correction array to the datamodel?
+        input_model.meta.cal_step.barshadow = 'COMPLETE'
```
hbushouse: datamodels core.schema.yaml will need to be updated to include "barshadow" in the cal_step list. And we should also add an R_BARSHA keyword for the reference file.

jwst/barshadow/bar_shadow.py
stscirij: The NIRSpec Technical Note ESA-JWST-SCI-NRS-TN-2016-016 says:
"The results can be stored as a new extension in the 2D irregular data product"
I'll take out the line that applies the correction and instead attach the correction to the MultiSpec model.

hbushouse: Just comment out the line for now that applies the correction, but definitely also add the correction array as a new extension in the data model. The JCCWG is still trying to reach consensus on whether or not to apply all the corrections as they're computed or apply them later. In either case they want the corrections saved.

stscirij: Added the correction array to the multispec data model.  I removed the line that applies the correction, it's pretty straightforward to put it back if/when needed

jwst/datamodels/__init__.py
```diff
@@ -104,7 +104,7 @@
 __all__ = [
     'open',
     'DataModel', 'AmiLgModel', 'AsnModel',
-    'CameraModel', 'CollimatorModel',
+    'BarshadowModel', 'CameraModel', 'CollimatorModel',
```
hbushouse: 👍  for alphabetical order!

jwst/barshadow/bar_shadow.py
```diff
@@ -0,0 +1,440 @@
...
+import numpy as np
+import logging
+from .. import datamodels
+from jwst.assign_wcs import nirspec
+from gwcs import wcstools
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+slit_spacing = 202.0
```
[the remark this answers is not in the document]
stscirij: Maybe better to just go with the ratio of 1.15, after all, that's what assign_wcs uses

jwst/barshadow/bar_shadow.py
```diff
@@ -0,0 +1,440 @@
...
+slit_height = 175.0
+SLITRATIO = slit_spacing/slit_height     # Approximate value until we have something better
+
+def do_correction(input_model, barshadow_model):
+    """
+    Short Summary
+    -------------
```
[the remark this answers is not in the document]
stscirij: OK

jwst/barshadow/bar_shadow.py
```diff
@@ -0,0 +1,440 @@
...
+    Execute all tasks for Bar Shadow Correction
+
+    Parameters
+    ----------
+    input_model: data model object
```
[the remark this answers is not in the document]
stscirij: Agreed

jwst/barshadow/bar_shadow.py
```diff
@@ -0,0 +1,440 @@
...
+                if array_column >= array.shape[1]:
+#                    print("Column out of range: %d %d %d" % (array_row, row, column))
+                    array_column = 0
+                if array_row < 0:
+#                    print("Row out of range: %d %d %d" % (array_row, row, column))
+                    array_row = 0
+                if array_column < 0:
+#                    print("Column out of range: %d %d %d" % (array_row, row, column))
+                    array_column = 0
+                correction[row, column] = array[array_row, array_column]
```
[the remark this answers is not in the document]
stscirij: It's not really interpolation, but the barshadow model is oversampled by a factor of 50 in Y so it seemed that interpolation was a bit of overkill.  But technically the report calls for interpolation, so I'll put that in.

jwst/barshadow/bar_shadow.py
```diff
@@ -0,0 +1,440 @@
...
+                log.info("Slitlet %d has zero length, correction skipped" % slitlet_number)
+                #
+                # Put an array of ones in a correction extension
+                slitlet.barshadow = np.ones(slitlet.data.shape)
+        else:
+            log.info("Bar shadow correction skipped for slitlet %d (source not uniform)" % slitlet_number)
+            #
+            # Put an array of ones in a correction extension
+            slitlet.barshadow = np.ones(slitlet.data.shape)
+    return input_model.copy()
```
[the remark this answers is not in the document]
stscirij: Probably not, it was force of habit...

jwst/barshadow/bar_shadow.py
```diff
@@ -0,0 +1,440 @@
...
+                xslit, yslit, wavelength = det2slit(x, y)
+                #   Scale the yslit values by the ratio of slit spacing to slit height
+                yslit = yslit * SLITRATIO
+                #   Convert the Y and wavelength to a pixel location
+                #   in the  bar shadow array
+                index_of_fiducial = shutter_status.find('x')
+                index_of_fiducial_in_array = 501 + index_of_fiducial*500
+                yrow = index_of_fiducial_in_array - yslit*500.0
+                wcol = (wavelength - w0)/wave_increment
+                nonnan = np.where(~np.isnan(yrow))
```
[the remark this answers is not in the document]
stscirij: It was part of some debugging code that didn't get removed with the rest

hbushouse:
If/when you think this is ready for prime time, update the calwebb_spec2 pipeline to call this step right after the pathloss step, but only for EXP_TYPE='NRS_MSASPEC' of course. Are there ref files in CRDS yet to support this?

stscirij:
Not quite - I am in the process of delivering the reference files.  Once they are in there, I'll add the step to the pipeline and we'll see what happens...

PR #1232, opened 2017-09-23
NRC_TSGRISM mode initial implementation for assign_wcs

sosey:
assign_wcs for NRC_TSGRISM mode is implemented. **Before merging** I need some more information on the source location in the subarrays. Is the source pixel location saved in the meta information somewhere, or is it a known value for each subarray? @hbushouse ?


sosey:
After discussing at the standup, I'm going to leave the source pixel location open-ended right now. See this issue for future implementation details see #1235

PR #1236, opened 2017-09-25
Modify registry initialization so it is only invoked once

bernie-simon:
The __init__ method for datamodels can be call more than once. But  the astropy.io.registry code throws an error if it is. So code was added to check if the __init__ method has been called previously and only do the registry initialization if it has not been called  previously.


jaytmiller:
Why would datamodels.__init__ be run more than once,  the first time it is imported anywhere?   Aren't modules cached once they've been imported one time?  (And if it is not cached,  and is being re-evaluated,  then wont the side effects you intend to thwart the re-registration be missing?)  I am just theorizing but I don't see why this would work.

bernie-simon:
If loaded through an import statement, it would only be loaded once. However, you can force a load by specifically calling the import library and that's what the script that threw the error did.

jdavies-st:
Btw, this reloading is of __init__.py for the datamodels module that is being run anytime datamodels is imported.  Because we have a bunch of stuff in there to remap the module namespace to make datamodels.open() work, that code gets rerun when imported.

PR #1239, opened 2017-09-26
Single band

jemorrison:
@hbushouse 
@stscieisenhamer 
Finished looping over cube_build to make a set of default IFU Cubes. In the spec2 pipeline we want to make only 1 single IFUCube so we need to use --ouput_type = 'multi'
In the spec3 pipeline a set of IFUCube are returned from cube_build_step. These IFUCubemodels are stored in a ModelContainer. 
Cube_build creates all the filenames for each IFUCube.

### DIFF
--- a/jwst/cube_build/blot_cube_build.py
+++ b/jwst/cube_build/blot_cube_build.py
@@ -63,7 +63,7 @@ class CubeBlot(object):
         self.cdelt2 = median_model.meta.wcsinfo.cdelt2*3600.0
         self.cdelt3 = median_model.meta.wcsinfo.cdelt3*3600.0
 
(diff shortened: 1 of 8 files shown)


jemorrison:
IFUCube.meta.cal_step.cube_build = 'COMPLETE' is set in ifu_cube.py

But now I am not sure where it should go. Currently it is in process the creates an IFUCube Model.

So it is set for each ifu cube model, if I put it in Final_IFUCube.meta.cal_step.cube_build then
that is for the Model Container. If I  add the COMPETE to the model container will it go in and
add COMPLETE to the meta of each IFUCube ?

jdavies-st:
No, 'COMPLETE' needs to be set for each model.  And yes, I see it now in ifu_cube.py.  Sorry, had missed it earlier.

PR #1243, opened 2017-09-27
Fix input_models var name bug

jdavies-st:
PR #1231 changed a variable name to one that was already in use elsewhere, inadvertently braking the writing out of Level 2c files.  This fixes that bug.


jdavies-st:
This incorrect behavior is caused by the default of `ModelContainer` opening to be `persist=False`.   This is addressed in #1246.

Otherwise OK to merge?

jdavies-st:
Fix to `ModelContainer` in #1246 is merged.  Ok to merge this @stsci-hack?

PR #1248, opened 2017-09-27
Resample spec fixes

jdavies-st:
A grab-bag of fixes, mostly to the `resample` modules:
- PEP8 fixes
- Remove `with()` block at beginning of `ResampleSpecStep.process()`
- Make `input_models` not a class attribute
- Make resample_step result var name consistent with other steps
- Write `meta.cal_step.resample` 'COMPLETE' correctly
- Simplify single input to `ResampleStep`
- Simplify single input to `ResampleSpecStep`

### DIFF
--- a/jwst/pipeline/resample.cfg
+++ b/jwst/pipeline/resample.cfg
@@ -8,4 +8,4 @@ kernel = 'square'
 fillval = 'INDEF'
 good_bits = 4
 blendheaders = True
-suffix = 'resamp'
+suffix = 'i2d'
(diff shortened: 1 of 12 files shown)


jwst/pipeline/resample.cfg
```diff
@@ -7,4 +7,5 @@ pixfrac = 1.0
 kernel = 'square'
 fillval = 'INDEF'
 good_bits = 4
-blendheaders = True
\ No newline at end of file
+blendheaders = True
+suffix = 'resamp'
```
hbushouse: Resampled products should use a suffix of 'i2d' or 's2d', depending on whether they're images or spectra.

jdavies-st: Ah, right.  this suggests that I should have a separate `resample_spec.cfg`, and then have specific suffixes appropriate to each.

PR #1249, opened 2017-09-28
For WFSS/GRISM data, do flat_field before extract_2d

philhodge:
calwebb_spec2.py has been modified to do the `flat_field` step prior to `extract_2d` for NRC_GRISM and NIS_WFSS, and `extract_2d` will actually be run for those types.


jwst/pipeline/calwebb_spec2.py
```diff
@@ -179,12 +179,24 @@ def process_exposure_product(
...
+        # It isn't really necessary to include 'NRC_TSGRISM' in this list,
+        # but it doesn't hurt, and it makes it clear that flat_field
+        # should be done before extract_2d for all WFSS/GRISM data.
+        if exp_type in ['NRC_GRISM', 'NIS_WFSS', 'NRC_TSGRISM']:
+            # Apply flat-field correction
+            input = self.flat_field(input)
 
-        # Apply flat-field correction
-        input = self.flat_field(input)
+            if exp_type in ['NRS_FIXEDSLIT', 'NRS_BRIGHTOBJ', 'NRS_MSASPEC',
```
sosey: I think the logic here should just be `if exp_type != NRC_TSGRISM`, the NRS modes have already been excluded by the previous if logic and won't make it here.

philhodge: Yes, that's much better; thanks.

PR #1250, opened 2017-09-28
Bots extract2d

### LINKED ISSUE #1081: Run extract_2d on Nirspec TSO data
From Howard:
For the sake of consistency and in order for WCS mapping to work properly, we should be applying extract_2d to single slit TSO exposures too, which means the extract_2d step needs to be updated to handle 3D data arrays (all of the data arrays will be 3D: SCI, ERR, and DQ), so that it produces a cutout from the "x" and "y" axes of the cubes (axes 2 and 1, respectively, in python), but leaves all of the data along the "z" axis (axis 0 in python) intact, so that we retain the data for all integrations. 

From James M:
Our TSO mode, BOTS, automatically selects the S1600A slit.  The other slits cannot be used.

nden:
Closes #1081 

Implements the TSO mode in NIRSpec - `EXP_TYPE  = BRIGHTOBJ`.

Refactored the extract_2d code to split it into functions and moved it to a module `nirspec.py`.
Extraction code for other instruments will go in other modules - all called from the lightweight top level `exract_2d.py`.


jwst/extract_2d/nirspec.py
```diff
@@ -0,0 +1,390 @@
...
+    supported_modes = ['NRS_FIXEDSLIT', 'NRS_MSASPEC', 'NRS_BRIGHTOBJ', 'NRS_LAMP']
+
+    exp_type = input_model.meta.exposure.type.upper()
+    log.info('EXP_TYPE is {0}'.format(exp_type))
+
+    wavecorr_supported_modes = ['NRS_FIXEDSLIT', 'NRS_MSASPEC', 'NRS_BRIGHTOBJ']
+    if exp_type in wavecorr_supported_modes:
+        if reffile and reffile.strip().upper() == 'N/A':
+            apply_wavecorr = False
+            log.info("WAVECORR reference file missing - skipping correction")
```
hbushouse: Might want to change that log message to a warning, rather than just info.

jwst/extract_2d/extract_2d.py
```diff
@@ -1,302 +1,16 @@
...
-        (x, y) positions in the ``msa_frame``.
 
-    Returns
-    -------
-    xpos : float
-        The fractional X coordinates within the slit.
-    """
-    num, xcenter, ycenter, xsize, ysize = msa_model.Q5.data[slit.shutter_id]
-    return (xposabs - xcenter) / (xsize/ 2.)
+    if exp_type in nrs_modes:
```
hbushouse: So what about using extract_2d for NRC_GRISM and NIS_WFSS? Is that part coming yet?

sosey: ...it's coming....it's coming....

PR #1252, opened 2017-09-28
PEP8 fixes

nden:


nden:
Too much new code is in to use this check.

PR #1259, opened 2017-09-29
Add .cfg for resample_spec; Fix ref_filename bug

jdavies-st:
- Add a `resample_spec.cfg` file for the spec version of the step, as it is has a different `suffix` than the `resample.cfg`
- Fix a bug in the reference file filename not getting passed to the resample routine


jdavies-st:
Merging since local unit testing on OS X pass, and Travis Linux tests both pass.  Travis OS X tests are ridiculously slow this past week.

PR #1260, opened 2017-09-29
Mark model containers as not persistent

bernie-simon:
The model containers in cube_skymatch do not need to remain open after a loop, so I added the persist=False flag.


hbushouse:
Has anyone like @jdavies-st or @jemorrison confirmed that it's OK to have these not be persistent?

jemorrison:
Looks good I do not think the cube_models needs to remain open

PR #1261, opened 2017-09-29
Skip persistence step for MIRI and NIRSpec

### LINKED ISSUE #1247: Skip persistence step for NIRSpec and MIRI for B7.1
The NIRSpec and MIRI instrument teams have requested that the persistence correction step be skipped for all data from those instruments, at least for B7.1, due to the lack of useful reference file information for MIRI and the NIRSpec team is simply not sure they want the correction applied at all (see https://jira.stsci.edu/browse/JP-180).

So the calwebb_sloper pipeline should be updated to skip that step for these 2 instruments.

hbushouse:
This updates the calwebb_detector1 pipeline to skip the persistence correction step for both MIRI and NIRSpec.

Fixes #1247.


philhodge:
Looks good to me.

PR #1266, opened 2017-10-02
allow additional properties in subarray

nden:
This allows `p_subarray` to be included in the `subarray` schema from the `core` schema, so that `datamodels.open(distortinon_file)` works.

@hbushouse We haven't updated the `core` schema to reference the individual schemas. So both the short schemas and their predecessors in the `core` schema exist. Should we update the `core` schema to use the smaller schemas in one of the 7.1 updates or in 7.2 ?


hbushouse:
I would wait till B7.2 to update the core to use the smaller schemas. Let's not introduce any further risks this late in the B7.1 cycle.

PR #1268, opened 2017-10-02
updated extract2d to use reference file dictionary in step

sosey:
This updates the extrac2d step to pass a dictionary of reference file names to the lower calls. 
The update passes for grisms, @nden should check I did the correct thing for nirspec

### DIFF
--- a/jwst/extract_2d/extract_2d.py
+++ b/jwst/extract_2d/extract_2d.py
@@ -13,18 +13,24 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 
-def extract2d(input_model, which_subarray=None, apply_wavecorr=False, reffile="", grism_objects=[]):
+def extract2d(input_model, which_subarray=None, apply_wavecorr=False, reference_files={}, grism_objects=[]):
+
+    if not any(reference_files.values()):
+        log.critical("extract2d needs reference files to make extractions, none were passed")
+        raise ValueError("extract2d needs reference files to make extractions, none were passed")
+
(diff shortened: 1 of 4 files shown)


jwst/extract_2d/extract_2d.py
```diff
@@ -13,18 +13,24 @@
 log.setLevel(logging.DEBUG)
 
 
-def extract2d(input_model, which_subarray=None, apply_wavecorr=False, reffile="", grism_objects=[]):
+def extract2d(input_model, which_subarray=None, apply_wavecorr=False, reference_files={}, grism_objects=[]):
+
+    if not any(reference_files.values()):
+        log.critical("extract2d needs reference files to make extractions, none were passed")
```
hbushouse: Is this true (that extract_2d needs reference files) for all cases in which extract_2d is used, including, for example, NIRSpec fixed-slit and MOS? What reference files does it use?

sosey: grisms always needs the wavelengthrange reference file.  nirspec is using the wavecorr file for NRS_FIXEDSLIT, NRS_MSASPEC, NRS_BRIGHTOBJ. If we want to avoid doing checks we need to move the exp_type checking logic up to the step level

sosey: okay - just removing the warning, that should make NRS_LAMP go through with no reference files and the other classes can check reference files for themselves :)

hbushouse: 👍  Just didn't realize that all the modes needed some kind of ref file. Now I know.

PR #1269, opened 2017-10-02
calspec2 should always create a single IFU cube

### LINKED ISSUE #1265: Update calspec2 to always build a single IFU cube
When cube_build is called for a single input exposure, which is always the case when it's called from the calspec2 pipeline, it should always return a single IFU cube even when the input data span multiple wavelength bands. Cube_build is capable of doing this by setting the step parameter "output_type=multi". The calspec2 pipeline needs to be updated to always call cube_build in this way.

hbushouse:
Updated the calspec2 pipeline to set the cube_build output_type argument to "multi", so that it always produces a single IFU cube. Fixes #1265.

### DIFF
--- a/jwst/pipeline/calwebb_spec2.py
+++ b/jwst/pipeline/calwebb_spec2.py
@@ -241,7 +241,10 @@ class Spec2Pipeline(Pipeline):
 
         elif exp_type in ['MIR_MRS', 'NRS_IFU']:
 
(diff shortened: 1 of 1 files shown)


hbushouse:
All Linux tests passed; iOS tests don't seem to have any intention of starting anytime soon. Merging.

PR #1278, opened 2017-10-04
Compute the footprint of the aperture on the sky

### LINKED ISSUE #580: Populate S_REGION keyword in calibrated products
The S_REGION keyword is intended to store the spatial footprint of an imaging observation using the VO standard STCS representation. It can be easily computed using the astropy.wcs calc_footprint method on the WCS of the image. Brian McLean suggests the following settings to make the result consistent with the way CAOM has been computing footprints for HST and other MAST products:

footprint = w.calc_footprint(undistort=True, center=True)

The resulting footprint information, which consists of the indexes of the 4 corners of the image on the sky, should then be formatted into a string value as follows:

stcs = 'POLYGON ' + str(footprint[0][0]) + ' ' + str(footprint[0][1]) + ' '  + str(footprint[1][0]) + ' ' + str(footprint[1][1]) + ' ' + str(footprint[2][0]) + ' ' + str(footprint[2][1]) + ' ' + str(footprint[3][0]) + ' ' + str(footprint[3][1])

nden:
Close #580.

Implements a footprint calculation in products at various levels and populates `S_REGION`.
The first time the string is set is by `set_telescope_pointing`. It uses the PRD database to get the vertices of the aperture in the SIAF file. The keyword is updated by `assign_wcs` and `extract_2d`. Are there other steps that need to update it?
@sosey can you check the grism code?


hbushouse:
The resample and resample_spec steps should also update this, because they can combine data that results in a larger footprint. Of course because the data have been resampled at that point, it should be pretty easy to recompute the footprint (just get the RA/Dec of each corner pixel?).

jwst/assign_wcs/util.py
```diff
@@ -471,3 +471,25 @@ def get_num_msa_open_shutters(shutter_state):
     if 'x' in shutter_state:
         num += 1
     return num
+
+
+def update_s_region(model):
```
hbushouse: If this function is being called at the time of assign_wcs, when for example IFU images are still in their full 2D form containing many slices, will the wcs.footprint() method compute a reasonable result? Can it handle complex WCS objects like that?

hbushouse: Ideally, for something like IFU data, it should compute the footprint of the physical IFU aperture on the sky.

jwst/lib/set_telescope_pointing.py
```diff
@@ -207,45 +212,72 @@ def update_wcs(model, default_pa_v3=0., **kwargs):
...
-            footprint[0][0], footprint[0][1]
-        )
-    )
-    wcsinfo.s_region = s_region
+        "POLYGON ICRS "
+        " {0} {1}"
+        " {2} {3}"
+        " {4} {5}"
+        " {6} {7}"
+        " {0} {1}".format(*footprint.flatten()))
```
hbushouse: See question above as to whether this order of arguments goes counter-clockwise around the perimeter of the footprint.

jwst/assign_wcs/util.py
nden: Good catch! It will work for the MIRI IFU but not for Nirspec (IFU, FS, or MSA).

jwst/assign_wcs/util.py
```diff
@@ -471,3 +471,25 @@ def get_num_msa_open_shutters(shutter_state):
...
+        shape = model.data.shape
+        bbox = ((0, shape[1]), (0, shape[0]))
+    footprint = model.meta.wcs.footprint(bbox, center=True).T
+    s_region = (
+        "POLYGON ICRS "
+        " {0} {1}"
+        " {2} {3}"
+        " {4} {5}"
+        " {6} {7}"
+        " {0} {1}".format(*footprint.flatten()))
```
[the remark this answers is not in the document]
nden: I think the order is correct here but I'll double check.

jwst/lib/set_telescope_pointing.py
nden: Pretty sure it's in the correct order here.

jwst/assign_wcs/util.py
nden: I excluded the complex NRS types for now, until the WCS is modified to run in a simplified way.

nden:
The current test failure is due to lack of URI support when opening a database connection in python 2. Should I make it work on python 2 as well or is that not a consideration any more because we are dropping support for python 2?

hbushouse:
Good question, regarding python 2. We still have CI tests running under both py2 and py3 and it would be a "nice to have" if we could make sure that at least B7.1 continues to work as best it can under py2, just to give folks in INS, instrument teams, etc. a bit of a transition period. But for something like this, which only runs in the Ops environment (because it needs db access), I don't think we need to worry about py2 support. Ops is switching to py3 exclusively for B7.1 and beyond. So is there a way to mark these specific tests as "expected/allowed" failures, just so they aren't annoying?

jhunkeler:
@nden @hbushouse

Thoughts:
Dropping 2.7 support entirely and disabling relevant 2.7 testing in Travis would be easier. If the pipeline requires >=3.5 then the end-user's environment should match what works, or we might end up wasting time tracking down 2.7-specific bug reports that are actually non-issues in 3.x.

jwst/lib/set_telescope_pointing.py
```diff
@@ -104,18 +109,20 @@ def add_wcs(filename, default_pa_v3=0., **kwargs):
         Keyword arguments used by matrix calculation routines
     """
     logger.info('Updating WCS info for file {}'.format(filename))
-    model = dm_open(filename)
+    model = DataModel(filename)
+    model = _extend_schema(model, 'wcsinfo.schema.yaml')
```
hbushouse: 👍

Issue #1277, opened 2017-10-04
Convert reference data from float to complex, if necessary


philhodge:
The reference data for reference pixel correction for NIRSpec IRS2 data are 1-D arrays of complex numbers.  The NIRSpec team has delivered reference files (tables) with the data declared as float, i.e. interleaved real and imaginary parts.  The refpix step should check for this case and convert the data on-the-fly to complex.

hbushouse:
Check to see if new ref files are selected in latest CRDS context. If so, merge.

hbushouse:
Fixed in #1279

PR #1279, opened 2017-10-04
Convert reference data from float to complex.

philhodge:
The schema for the IRS2 reference file was changed to specify that the
columns are float32 rather than complex64.  irs2_subtract_reference.py
was modified to convert these values to complex.
See GitHub issue #1277 and JIRA ticket JP-181.

### DIFF
--- a/jwst/datamodels/irs2.py
+++ b/jwst/datamodels/irs2.py
@@ -16,8 +16,10 @@ class IRS2Model(model_base.DataModel):
         Any of the initializers supported by `~jwst.datamodels.DataModel`.
 
     irs2_table : numpy array
-        A table with 8 columns and 1458176 (2048 * 712) rows; all values
-        are complex.  There are four columns for ALPHA and four for BETA.
+        A table with 8 columns and 2916352 (2048 * 712 * 2) rows.  All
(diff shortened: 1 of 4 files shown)


hbushouse:
Should we wait to merge this until the new format ref files get installed in CRDS? Otherwise it won't be able to read the existing ref files that store things as complex.

philhodge:
The new format reference files already are in CRDS; that was the problem.  The new ones have float data that contain alternating real and imaginary parts, while the original ones have complex data.  The code in this PR will work with the new reference files and will not work with the original reference files.  I couldn't make it work with either format of reference file without having two separate IRS2Models, and I didn't think it was worth the bother.  The NIRSpec team must have known that the original reference files had complex data, but they delivered their new ones with float data, so I'm assuming they will continue to deliver float data in the future.

hbushouse:
CRDS context 409 contains an updated NIRSpec refpix rmap that now includes the new refpix ref files and removes the old ones. So as soon as testing of this code is satisfied it can be merged.

philhodge:
The Linux tests have passed, but the Mac tests haven't even started yet (after three hours).  Merging.

PR #1282, opened 2017-10-05
Add sky bounding box properties to source catalog

larrybradley:
These properties will be used by the wide-field slitless spectroscopy pipeline.

CC: @sosey, @hbushouse


hbushouse:
👍

sosey:
w00t! I'll go test it out :)

Issue #1285, opened 2017-10-05
mrs_imatch getting zero-length cube container from CubeBuild


hbushouse:
When attempting to run an ASN of 4 MIRI MRS exposures through the calspec3 pipeline, the mrs_imatch step is failing with the error:
"The number of generated cube models does not match the number of input 2D images."

This occurs in line 265 of the mrs_imatch_step.py module, in function _match_models:

    cbs = CubeBuildStep()
    cbs.channel = str(channel)
    cbs.band = 'ALL'
    cbs.single = True
    cube_models = cbs.process(models)
    if len(cube_models) != len(models):
        raise RuntimeError("The number of generated cube models does not match the number of input 2D images.")

This is due to the fact that `cube_models` has length 0, while len(models) in this case is 4. The processing log up to that point shows that cube_build actually did build 4 cubes, but for some reason the length of the `cube_models` ModelContainer that's returned is 0.

I verified this interactively. If I simply call CubeBuildStep on a single MRS exposure, using all default values, it builds 2 cubes - one for each channel's worth of data in the exposure - and the returned ModelContainer has length 2. Fine so far. But when I do the same thing using the special param settings above, to have CubeBuild construct individual cubes on a common output frame, the CubeBuild step again seems to go through the motions of creating the desired number of cubes, but the ModelContainer that's returned has length 0:

In [5]: cbs=CubeBuildStep()
2017-10-05 14:26:14,232 - stpipe.CubeBuildStep - INFO - CubeBuildStep instance created.
In [6]: cbs.channel='1'
In [7]: cbs.band='ALL'
In [8]: cbs.single=True
In [9]: cube_models=cbs.process('MIRM103-Q0-SHORT_495_cal.fits')
2017-10-05 14:27:04,283 - stpipe.CubeBuildStep - INFO - Starting IFU Cube Building Step
2017-10-05 14:27:04,283 - stpipe.CubeBuildStep - INFO - Input interpolation: pointcloud
2017-10-05 14:27:04,283 - stpipe.CubeBuildStep - INFO - Coordinate system to use: ra-dec
2017-10-05 14:27:04,283 - stpipe.CubeBuildStep - INFO - Weighting method for point cloud: msm
2017-10-05 14:27:04,283 - stpipe.CubeBuildStep - INFO - Power Weighting distance : 2.000000
2017-10-05 14:27:15,559 - stpipe - INFO - The desired cubes covers the MIRI Channels: ['1']
[...]

PR #1289, opened 2017-10-06
Blendheaders

jemorrison:
This fixes issue #1285 
Also updated the way the out put file name is created, added an option 'pipeline' if 'pipeline =2' then uses calspec2 rules. No need to use the output_type='multi' in calspec2. If pipeline=2 then internally output_type= 'multi'

### DIFF
--- a/jwst/cube_build/cube_build_step.py
+++ b/jwst/cube_build/cube_build_step.py
@@ -27,7 +27,6 @@ class CubeBuildStep (Step):
          band = option('SHORT','MEDIUM','LONG','ALL','short','medium','long','all',default='ALL')
          grating   = option('PRISIM','G140M','G140H','G235M','G235H',G395M','G395H','ALL','all',default='ALL')
          filter   = option('CLEAR','F100LP','F070LP','F170LP','F290LP','ALL','all',default='ALL')
-         pipeline = integer(2,3,default=3)
          scale1 = float(default=0.0)
          scale2 = float(default=0.0)
          scalew = float(default=0.0)
@@ -119,8 +118,6 @@ class CubeBuildStep (Step):
         if self.single :
             self.output_type = 'single'
 
-        if self.pipeline == 2:
-            self.output_type = 'multi'
 #________________________________________________________________________________
     # read input parameters - Channel, Band (Subchannel), Grating, Filter
 #________________________________________________________________________________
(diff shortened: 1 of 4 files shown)


jwst/cube_build/cube_build_step.py
```diff
@@ -27,6 +27,7 @@ class CubeBuildStep (Step):
          band = option('SHORT','MEDIUM','LONG','ALL','short','medium','long','all',default='ALL')
          grating   = option('PRISIM','G140M','G140H','G235M','G235H',G395M','G395H','ALL','all',default='ALL')
          filter   = option('CLEAR','F100LP','F070LP','F170LP','F290LP','ALL','all',default='ALL')
+         pipeline = integer(2,3,default=3)
```
hbushouse: I'm not sure I really see the advantage in having the `pipeline` argument instead of just using the `output_type` setting when calling cube_build from the calspec2 pipeline. Either way calspec2 has to use and set a particular cube_build argument (either pipeline or output_type), so why is it preferable to set `pipeline`? IMHO, setting `output_type=multi` is a more direct and obvious way of indicating the desired behavior, rather than setting `pipeline=2`. It's not obvious without reading the code or the documentation exactly what's going to happen when you set `pipeline=2`, whereas `output_type=multi` tells you right away what behavior you're going to get.

jemorrison:
The reason I added a pipeline option is only when Pipeline = 2 does it 
form the output name differently.
The output_type = 'multi' is valid for building cubes in a calspec3 type 
situation.
When people run it on their own and want to combine band they will use 
output_type = 'multi'
I think people are going to want to build multi type cubes routinely 
from an assocation. If you can think of something else
- I could change an option in output_type = to be something that 
indicates it is for calspec2 pipeline but
I was not sure if that was confusing.

When I opened this branch I was going to try and make the blendheaders 
work when cube_build
returns a ModelContainer of IFU cubes. I could not really get it work so 
I just delayed that work.
I need to get David Law and others from NIRSPEC to tell me what they 
want in Blendheaders anyway.
I was going to meet with them and James and Warren next week.

Jane

jemorrison:
Removed the 'pipeline' options. Instead if output_type = 'multi' and there is only a single exposure then
it is assumed this is the calspec2 pipeline and the output file follows the rules of calspec2.

PR #1299, opened 2017-10-11
Support background subtraction for WFSS/GRISM data

philhodge:
The code is included to do background subtraction for WFSS/GRISM data, but it won't actually be called at the present time because we don't have any reference files.

### DIFF
--- a/jwst/background/background_sub.py
+++ b/jwst/background/background_sub.py
@@ -119,6 +119,9 @@ def subtract_wfss_bkg(input_model, bkg_ref):
     input_model: JWST data model
         input target exposure data model
 
+    bkg_ref: reference file model
+        master background file for WFSS/GRISM
+
     Returns
     -------
     result: JWST data model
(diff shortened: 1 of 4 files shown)


jwst/background/background_sub.py
```diff
@@ -91,3 +109,191 @@ def average_background(bkg_list):
...
+    mask = np.isnan(model.data)
+    if mask.sum(dtype=np.intp) == 0:
+        return model
+    else:
+        temp = model.copy()
+        temp.data[mask] = fill_value
+        return temp
+
+
+def mask_from_source_cat(shape, source_catalog):
```
sosey: The mask for grism mode cannot be made directly from the source catalog. The source catalog *only* contains information about the objects in the direct image. We must use the list of GrismObjects which contain the translated bounding boxes which should be used to avoid areas in the sky background reference image.

jwst/background/background_sub.py
```diff
@@ -91,3 +109,191 @@ def average_background(bkg_list):
...
+
+    Returns
+    -------
+    bkg_mask: ndarray
+        Boolean mask:  True for background, False for pixels that are
+        inside at least one of the source regions defined in the source
+        catalog.
+    """
+
+    # This was mostly copied from get_object_info() in assign_wcs/util.py
```
sosey: You can use create_grism_bbox directly here, that function calls get_object_info and checks that the required catalog fields are present. It returns a list of GrismObjects containing information about all the sources as well as the appropriate bounding boxes around the trace for each spectral order for each object.

PR #1305, opened 2017-10-12
Remove extra_fits when readung and writing model file from disk

bernie-simon:
The extra_fits area in a datamodel is intended to hold the information from a file that is not contained in the model schema. It is temporary and should not be persisted to disk in the asdf extension
when saving a model file. This change removes the extra_fits section of a tree after loading the asdf extension from a file and before writing the asdf extension to a file. In addition, it checks to make
sure that the asdf extension is not saved in the extra_fits area.

### DIFF
--- a/jwst/datamodels/fits_support.py
+++ b/jwst/datamodels/fits_support.py
@@ -370,6 +370,10 @@ def _save_extra_fits(hdulist, tree):
                     continue
                 hdu.header.append((key, val, comment), end=True)
 
+    # Remove extra_fits so it is not written to the asdf extension
+    if 'extra_fits' in tree:
+        del tree['extra_fits']
+
 
 def _save_history(hdulist, tree):
     history = tree.get('history', [])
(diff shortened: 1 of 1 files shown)


hbushouse:
How will this affect a situation where there are extra FITS keywords (not defined in the data model) in an input file that gets loaded into a data model? Will they still round-trip to a subsequent output file?

bernie-simon:
The contents of extra_fits comes from header keywords and extensions that are not defined in the schema. These are still written back to their original locations when the file is saved, so everything round trips. The difference is that keywords in the schema are also read from and saved to the asdf extension.  After this change keywords and extension not in the schema will no longer be read from and written to the asdf extension. 

The reason for the change is that, as @philhodge  noticed, if extra_fits extensions are read from the asdf extension they can be duplicated in the fits file when it is saved. This happens if the file is opened with two different schemas, throwing some of the data into extra_fits.

jdavies-st:
Yes, the side effect of this old bug was that the FITS ASDF extension contained a copy of...

...the ASDF extension.  Like nesting Russian dolls.

This change is very welcome, as now there's not a copy of the whole ASDF extension which gets loaded into the `extra_fits` attribute and then gets written back out to the ASDF extension when the model is written.

Now it works as intended - FITS keywords not in the datamodel schema are dumped to the `extra_fits` attribute.  And then these are round tripped back to the FITS file as keywords when it is written out.

PR #1306, opened 2017-10-13
Update NIRSpec IRS2 keyword attribute names

hbushouse:
Updated 3 modules in the ipc, refpix, and saturation steps to use new meta attribute names for the NIRSpec IRS2 readout keywords NRS_NORM and NRS_REF, in keeping with recent changes/additions to the core schema for these keywords.


philhodge:
That looks fine.  Thanks.

PR #1314, opened 2017-10-16
Fix WCS crpix values in cube_build

drlaw1558:
IFU cube WCS is saving 0-indexed crpix locations instead of 1-indexed FITS standard, so cube WCS is off by 1 pixel from correct values.  I've added this back into the metadata values, and verified performance against an idealized input array.

### DIFF
--- a/jwst/cube_build/ifu_cube.py
+++ b/jwst/cube_build/ifu_cube.py
@@ -709,9 +709,9 @@ class IFUCubeData(object):
         IFUCube.meta.wcsinfo.crval1 = self.Crval1
         IFUCube.meta.wcsinfo.crval2 = self.Crval2
         IFUCube.meta.wcsinfo.crval3 = self.Crval3
-        IFUCube.meta.wcsinfo.crpix1 = self.Crpix1
-        IFUCube.meta.wcsinfo.crpix2 = self.Crpix2
-        IFUCube.meta.wcsinfo.crpix3 = self.Crpix3
+        IFUCube.meta.wcsinfo.crpix1 = self.Crpix1+1
+        IFUCube.meta.wcsinfo.crpix2 = self.Crpix2+1
+        IFUCube.meta.wcsinfo.crpix3 = self.Crpix3+1
         IFUCube.meta.wcsinfo.cdelt1 = self.Cdelt1/3600.0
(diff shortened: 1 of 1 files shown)


hbushouse:
Noticed a couple of odd things about the new CRPIXn values in the regression test output files:

Extension HDU 1:
   Headers contain differences:
     Keyword CRPIX1   has different values:
        a> 26
        b> 25
     Keyword CRPIX2   has different values:
        a> 24
        b> 23
     Keyword CRPIX3   has different values:
        a> 2.0
        b> 1.0

First, why are CRPIX1 and CRPIX2 integer values? I would expect them to be float (although I note that they were integer even before this change).

Second, are you deliberately setting the reference pixel of the wavelength axis (CRPIX3) to the second pixel? Seems like it would be more natural to have it at the first pixel (1.0), which it was before.

jemorrison:
In #1314 David Law noticed that crpix1 & 2 should of had 1 added to them .
But it seems that we incorrectly also added 1 to crpix3. 

In cube build we are setting the CRPIX values to integer values. This is based on the first pixel has a center of 1,1,1.  But really those should be floats.

jemorrison:
I looked at the IFUCube written to disk and the CRPIX1 and CRPIX2 are floats.
So how are you probing these values ?

hbushouse:
The output above was from fitsdiff report output in the regression test. When I look at some of the s3d output files directly using fits.getheader() they also appear as integers:

fits.getheader('jw10001001001_01101_00001_mirifushort_s3d.fits',ext=1)

        WCS parameters                                                          
                                                                                
WCSAXES =                    3 / number of World Coordinate System axes         
CRPIX1  =                   23 / axis 1 coordinate of the reference pixel       
CRPIX2  =                   22 / axis 2 coordinate of the reference pixel       
CRPIX3  =                  2.0 / axis 3 coordinate of the reference pixel       
CRVAL1  =    5.458482612960622 / RA at the reference pixel (deg)                

This file was just created a couple hours ago from the reg test run based on last night's build.

jemorrison:
Ok I thought I was looking at a cube I created before I changed the code but I think that was not the case. The PR I just put in should fix  the CPRIX3 values. I moved where 1 was  being added to to CRPix1 and 2 so it did not add one to those values when the cubes are being made in the alpha-beta coordinate system.
I also cast an int to float when assigning crpix1 and 2 to values.

PR #1315, opened 2017-10-16
Fix for jump detection when there are saturated groups

mwregan2:
The program was using a wrong index to find the largest non-saturated difference. I added test to recreate the original problem that now work with the fix.
Also, the calculation of the median was not correct since it assumed all pixels in the integration had the same number of non-saturated groups.
An expected to fail test was added for a future feature to flag pixels adjacent to detected jumps.


hbushouse:
Related to #1225

PR #1320, opened 2017-10-17
disable osx tests

nden:
Travis has continuous problems with OSX testing. Here's a [blog post](https://blog.travis-ci.com/2017-09-22-macos-update) about this.
Since the code is mostly python, I suggest we disable the testing on osx for now. Any objections?

Should we disable python 27 as well and include python 3.6? (Python 3.6 is probably used in the nightly tests.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: c
 
 os:
     - linux
-    - osx
 
 sudo: false
 


hbushouse:
The py2.7 tests don't take terribly long to run and even though we're not officially supporting it anymore I'm just curious to continue to know - for at least a little while - whether things still work in it or not. So let's maybe keep them going a little longer yet (perhaps at least until B7.1 is completely finalized and officially shipped off to I&T). It would be useful perhaps at that time to then add py3.6 into the mix.

stscieisenhamer:
sooo we should still not willfully put 3-only stuff in yet? not that someone may have already done so; just asking...

nden:
I  put 3-only stuff in the set_pointing script and will merge it soon. It's run only in DMS and they run 3 so it should be fine. If someone in INS uses 2 they may (or may not) have a problem, depending on what they run. But then it's time they upgrade. Perhaps after tis is shipped to I&T we should remove 2.

PR #1321, opened 2017-10-17
Create a MultiSlitModel when one slit is extracted.

nden:
Create a `MultiSlitModel` product with a single slit as long as exp_type is not `NRS_BRIGHTOBJ`.
`NRS_BRIGHTOBJ` creates a `CubeModel` as output. Still need to run some tests before merging.


hbushouse:
I tested calwebb_spec2 on a couple of NIRSpec fixed-slit exposures that use a subarray containing only 1 slit and it now works fine. Even makes it through resample_spec without dying!

hbushouse:
calwebb_spec2 still dies in flat_field, however, when processing a NRS_BRIGHTOBJ exposure, because the output from extract_2d is a CubeModel, which the NIRSpec fixed-slit branch of flat_field doesn't know how to handle.

Issue #1312, opened 2017-10-16
Task: MIRI first-frame correction step


karllark:
The MIRI first frame is currently not used by anyone on the MIRI team in analyzing data.  There is some kind of transient seen and no correction algorithm is currently available. 

This step should flag the 1st frame as bad and later steps should not use it.  This is similar to the 1st pass at the last frame correction.
Later, there may be a correction algorithm.

karllark:
I will work on adding this step this week based on the existing last-frame correction step.
Cannot assign milestones, but someone who can should add this for B7.2.

hbushouse:
At what location in the MIRI data flow in calwebb_detector1 should this step be added?

karllark:
Right before the last-frame correction.

jemorrison:
I just looked at the calwebb_detector1.py code and the I think the order of the steps is wrong
it should be dq_init, saturation,reject first frame, reject last frame,  [ipc], linearity, rscd, dark current, refpix.
The RSCD correction uses the last frame from the last integration for a correction. Eventually we hope to have a last frame correction which will need to be applied before the RSCD correction.  It really does not matter if we reject last frame and then first frame or vice verse we just want to occurring  as one of the first steps in the pipeline. 
The RSCD correction in the JWST pipeline is an old algorithm I need to update it to the new algorithm that James Colbert has given me. It could be that the old algorithm is using the last frame- let me check

karllark:
@jemorrison: your ordering corresponds to what is the on the CALDETECTOR1 confluence page.  
With one exception, the IPC is in a different place (before first-frame correction).  Not sure this is important at this point as we are not doing IPC.  Where IPC fits in the pipeline will need to be discussed if/when we decide to use it - at least this is my thinking.

hbushouse:
Fixed by #1327.

PR #1327, opened 2017-10-19
MIRI firstframe correction step

karllark:
This code provides the MIRI firstframe correction step.  This is a MIRI only step.  Addresses issue #1312.

The step is straightforward and flags the 1st group in an integration with the 'DO_NOT_USE' flag.
This is a very similar step to the MIRI lastframe correction and was directly based on the code for that step.

I have included tests for this step in the tests subdirectory.  There are two steps.  One to test that the flag was set for the 1st group for an integration with 5 groups.  And one to test that the flag is not set for an integration with a single group.

This is my first time adding a step.  Please provide feedback on what more is needed to fully implement this step.  Additional tests?


docs/jwst/firstframe/description.rst
```diff
@@ -0,0 +1,8 @@
+
+Description
+===========
+
+The first frame correction step flags the first group as bad (the GROUP data
```
hbushouse: Two suggestions. Mention right off the bat that this is a MIRI-only correction, e.g. "The MIRI first frame correction ...". Second, wouldn't hurt to be explicit and say "first group in every integration".

docs/jwst/firstframe/description.rst
```diff
@@ -0,0 +1,8 @@
...
+quality flags for the first group in all integrations are reset to DO_NOT_USE),
```
hbushouse: Technically the DQ flags are not "reset" to DO_NOT_USE, but rather the DO_NOT_USE flag value is added to or propagated into any existing flag values.

docs/jwst/package_index.rst
```diff
@@ -21,6 +21,7 @@ Package Index
    extract_2d/index.rst
    fits_generator/index.rst
    flatfield/index.rst
+   firstframe/index.rst
```
hbushouse: OK, this is hopelessly picky, but to maintain strict alphabetical listing, firstframe should appear between extract_2d and fits_generator.

jwst/firstframe/__init__.py
```diff
@@ -0,0 +1,4 @@
+from __future__ import absolute_import
+from .firstframe_step import FirstFrameStep
+
+__version__ = '0.7.0'
```
hbushouse: Let's go ahead and set this to 0.7.2, because it will have its first official release in build 7.2

jwst/firstframe/firstframe_step.py
```diff
@@ -0,0 +1,29 @@
+from ..stpipe import Step
+from .. import datamodels
+from . import firstframe_sub
+
+
+class FirstFrameStep(Step):
+    """
+    FirstFrameStep: This is a MIRI specific task.  If the number of groups
+    is greater than 1, the GROUP data quality flags for the first group will
+    be set to DO_NOT_USE.
```
hbushouse: Again, let's phrase it as "added to", rather than "set to".

jwst/firstframe/firstframe_step.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    def process(self, input):
+
+        # Open the input data model
+        with datamodels.open(input) as input_model:
+
+            # check the data is MIRI data
+            detector = input_model.meta.instrument.detector
```
hbushouse: Couldn't hurt to make sure we're dealing with uppercase for the purpose of the comparison in line 20, so add the "upper" method here:
```
 detector = input_model.meta.instrument.detector.upper()
```

jwst/firstframe/firstframe_step.py
```diff
@@ -0,0 +1,29 @@
...
+            if detector[:3] == 'MIR':
+                # Do the firstframe correction subtraction
+                result = firstframe_sub.do_correction(input_model)
+            else:
+                self.log.warning('First Frame Correction is only for MIRI data')
+                self.log.warning('First frame step will be skipped')
+                result = input_model.copy()
+                result.meta.cal_step.firstframe = 'SKIPPED'
```
hbushouse: We'll need to add a new entry to datamodels/schemas/core.schema.yaml for this new attribute "firstframe" in the cal_step hierarchy, so that this attribute is defined.

jwst/firstframe/firstframe_sub.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # Update the step status, and if ngroups > 1, set all of the GROUPDQ in
+    # the first group to 'DO_NOT_USE'
+    if sci_ngroups > 1:
+        output.groupdq[:, 0, :, :] = \
+            np.bitwise_or(output.groupdq[:,0,:,:], dqflags.group['DO_NOT_USE'])
+        log.debug("FirstFrame Sub: resetting GROUPDQ in first frame to DO_NOT_USE")
+        output.meta.cal_step.firstframe = 'COMPLETE'
+    else:   # too few groups
+        log.warning("FirstFrame Sub: too few groups, skipping step")
```
hbushouse: Minor thing, but might want to refer to it as "FirstFrame Corr", because it isn't a subtraction operation.

docs/jwst/firstframe/description.rst
karllark: Done.

docs/jwst/firstframe/description.rst
karllark: Done.

docs/jwst/package_index.rst
karllark: Done.

jwst/firstframe/__init__.py
karllark: Done.

jwst/firstframe/firstframe_step.py
karllark: Done.

jwst/firstframe/firstframe_step.py
karllark: Done.

jwst/firstframe/firstframe_step.py
karllark: Done.  Please check that I did it correctly.

jwst/firstframe/firstframe_sub.py
karllark: Done.

karllark:
Many of these comments apply to the lastframe correction as well, given that I used that as my template and it is also a MIRI only step.  I could easily make similar changes to that step and add similar tests.  I'd be happy to do this as long as there are no objections.

hbushouse:
By the way, **_DO NOT MERGE_** this code for at least a few days, so that we make sure it doesn't inadvertently conflict with anything in the b7.1rc4 patch release we're planning to make at the end of this week (or very beginning of next week).

karllark:
No worries, I can't do the merge anyway - no privilege to do so.  I assume your comment is directed at others.  :-)

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1160,6 +1160,14 @@ properties:
                 title: Filter Offset reference file name
                 type: string
                 fits_keyword: R_FILOFF
+          firstframe:
```
hbushouse: Because this step doesn't use a reference file we don't need the new entries in the schema to define a R_FRSTFR keyword. We only need the S_FRSTFR keyword for the completion status. So the new lines 1163-1170 can be removed.

jwst/datamodels/schemas/firstframe.schema.yaml
```diff
@@ -0,0 +1,26 @@
+title: First Frame Correction data model
```
hbushouse: Because the first-frame step doesn't use a reference file there's no need to define a reference file schema. This new file (firstframe.schema.yaml) can be removed.

jwst/datamodels/schemas/core.schema.yaml
karllark: It is correct that this step does not use a reference file as it is currently written.  But it will need one when we get a true correction algorithm and not just flagging.  Should I still delete these lines?

jwst/datamodels/schemas/firstframe.schema.yaml
karllark: Same answer/question as above.

jwst/datamodels/schemas/core.schema.yaml
hbushouse: Need @jemorrison to answer that.

jemorrison: There is a desire one day to have a real correction for the first frame. This may happen after build 7.2 (certainly not for 7.2). I do not know of anyone on MIRI team activity working on a correction algorithm - it is just a hope.  So this would probably fall in the "place holder" status and I know you guys don't really want "place holders" in the pipeline. So I think at this time I would remove the reference to the first frame reference files.

karllark:
Requested changes made.

hbushouse:
Looks good now. Merging.

PR #1334, opened 2017-10-19
Update to support WFSS/GRISM data

philhodge:
Added code to get the spectral order number (needed for GRISM data, at least), and to include it in the call to the wcs function when appropriate.  The spectral order number is now saved in the output file.

Expanded the allowed extraction limits by a half pixel at each endpoint of the extraction width, if the width is the full width (or height) of the input 2-D image.  The previous limits stopped at the centers of the first and last pixels.


sosey:
@philhodge  i'm submitting a PR later today that removes the need to specify the order, you just need to specify the x,y in the forward direction from grism_detector -> sky

sosey:
The spectral order number is saved in the MultiSlit models meta.wcsinfo.spectral_order during the extract_2d execution

philhodge:
I need the spectral order for other reasons, for NIS_SOSS and to save the value in the output header.  But thanks for the heads-up, because I'll need to change the code that calls the wcs function.

PR #1316, opened 2017-10-17
Update Step data saving infrastructure

stscieisenhamer:
# Issue
Refactor a number of components in both `Step`s and `DataModel` related to file saving. Issues resolved are:

# Implementation
- [x] Resolves #1307 
- [x] Fix saving multiple results/`ModelContainer` results in the `Step` architecture. In particular, during `Spec3Pipeline` processing, properly write out multiple results for the `extract_1d` step when `cube_build` produces more than one result.
- [x] Resolves #1322 
- [x] `caldetector1` is not replacing known suffix `rate` when `output_file` is specified.
- [x] Remove `output_basename` and make everything use `output_file`
- [x] Add `fits` as a default extension since associations don't put one on.  Or maybe they should.
- [x] `Step.py::427`: remove `output_file` from the logic
- [x] Reconsider `Step._make_result_id`
- [x] template usage
    - [x]  `format_template`
        - add ability to handle lists
            _not necessary at the moment_
        - [x] add parameter to drop unused replacement strings.
    - [ ] update association generation
    - [x] update pipelines
        - [x] #1598 
        - [x] #1478
- [x] walk through all steps/pipelines for consistency
    - [x] `calwebb_tso3`
        - [x] `ModelContainer` and things simply appending
        - [x] `output_dir` not being honored. wtf
- [x] pandokia regression
    - [x] update local repo
    - [x] merge local repo back
    - [x] merge this PR
    - [x] ???

PR #1338, opened 2017-10-19
Fix step.run itiration bug when saving

stscieisenhamer:
A previous commit expanded the results list
to treat incoming `ModelContainer`s as an iterable,
resulting in filenames being overwritten by the indexed version.
The list is recreated forcing a `ModelContainer` to be witin the list
as opposed to representing the list.

This is actually a component of PR #1316, but has warranted its own PR.  Should really have been so in the first place.

### DIFF
--- a/jwst/stpipe/step.py
+++ b/jwst/stpipe/step.py
@@ -337,7 +337,7 @@ class Step(object):
         orig_log = log.delegator.log
         log.delegator.log = self.log
 
-        result = None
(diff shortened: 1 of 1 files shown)


stscieisenhamer:
@jemorrison this pr will fix your issue with the different output names depending on how you run. won't fix the bigger issues, but should get you back to before the "appending the index" issue arose.

PR #1339, opened 2017-10-19
Updated schema used by ramp fitting so that extra_fits is not needed …

dmggh:
…to output the variances of the slope.


jwst/datamodels/schemas/image.schema.yaml
```diff
@@ -43,6 +43,18 @@ allOf:
       ndim: 2
       default: 0.0
       datatype: float32
+    var_p2d:
+      title: variance due to poisson noise 
+      fits_hdu: VAR_P2D
```
hbushouse: I would suggest using the same EXTNAME "VAR_POISSON" as in the cube model.

jwst/datamodels/schemas/image.schema.yaml
```diff
@@ -43,6 +43,18 @@ allOf:
...
+      ndim: 2
+      default: 0.0
+      datatype: float32
+    var_r2d:
+      title: variance due to read noise 
+      fits_hdu: VAR_R2D
```
hbushouse: Same here. Why not use "VAR_RNOISE" as in the cube model.

jwst/datamodels/schemas/rampfitoutput.schema.yaml
```diff
@@ -14,6 +14,18 @@ allOf:
       default: 0.0
       ndim: 4
       datatype: float32
+    var_p:
+      title: Variance due to poisson noise for segment-specific slope
+      fits_hdu: VAR_P
```
hbushouse: Might as well spell it out to make it more obvious: "VAR_POISSON"

jwst/datamodels/schemas/rampfitoutput.schema.yaml
```diff
@@ -14,6 +14,18 @@ allOf:
...
+      default: 0.0
+      ndim: 4
+      datatype: float32
+    var_r:
+      title: Variance due to read noise for segment-specific slope
+      fits_hdu: VAR_R
```
hbushouse: VAR_RNOISE

jwst/datamodels/image.py
```diff
@@ -33,7 +33,8 @@ class ImageModel(model_base.DataModel):
     schema_url = "image.schema.yaml"
 
     def __init__(self, init=None, data=None, dq=None, err=None, relsens=None,
-                 relsens2d=None, zeroframe=None, area=None, wavelength=None, **kwargs):
+                 relsens2d=None, zeroframe=None, area=None, wavelength=None, 
+                 var_p2d, var_r2d, **kwargs):
```
hbushouse: Should these be set to var_p2d=None and var_r2d=None, like the rest?

hbushouse:
In support of #596.

PR #1345, opened 2017-10-20
Rename dunderdel to reset

### LINKED ISSUE #1272: ExtractModel defines __del__: is this necessary?
`__del__` methods can be detrimental to the garbage collection algorithm, in particular in the removal of circular references. There is no commentary in the code, so wondering if anyone would know a reason for the this method on this model. This is the only model, and in fact the only code in the jwst package, that defines `__del__`.

stscieisenhamer:
Resolves #1272

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -1009,7 +1009,7 @@ class ExtractModel(object):
         return (ra, dec, wavelength, net, background)
 
 
-    def __del__(self):
+    def reset(self):
         self.dispaxis = None
         self.xstart = None
         self.xstop = None
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -1009,25 +1009,6 @@ class ExtractModel(object):
         return (ra, dec, wavelength, net, background)
 
 
-    def reset(self):
-        self.dispaxis = None
-        self.xstart = None
-        self.xstop = None
-        self.ystart = None
-        self.ystop = None
-        self.extract_width = None
-        self.independent_var = None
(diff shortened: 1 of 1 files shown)


stscieisenhamer:
In that function, `extract_one_slit`, the object `extract_model` is completely local and the object should go out-of-scope on exit.  For all practical purposes, the `del` is redundant; it is simply removing the binding, decreasing the reference count to the object. This will happen anyways on function exit. Then, since the reference count is, presumably, zero, it will be collected after function exit.

When the object itself is collected, all its attributes will also be unbound, reducing their references counts. Again, if their references go to zero, they'll be collected also.  `reset` may make the collection happen earlier, but not significantly since the the object itself is going out-of-scope immediately thereafter.

If there is a need to force collection, the function would want to do
```
del extract_model
gc.collect()
```

The only use of `reset` that I could see is if one wanted to keep the `ExtractModel` object but "clear it out" for further usage, i.e. to actually reset the object itself.  But otherwise, in the world of memory management, it serves no particular use.

That's if I got the interpretation right.

philhodge:
Yes, that's my understanding, too.  I added `__del__` (and called it) because I didn't completely believe that it would actually be deleted when it went out of scope.

Why don't you just delete the method and the `del extract_model`?  There's no need for a `reset` method that is never called.

stscieisenhamer:
👍 And, as recent events have demonstrated, there is a good chance that some of the children objects may not go out-of-scope. However, also from recent events, nothing can be done about that except at the the level of those objects; if someone else has a hold of them, no amount of `del`/`gc.collect` is going to fix it.

Remove all the del's!

PR #1359, opened 2017-10-24
Update core schema for new subarray values

hbushouse:
JWST Keyword dictionary ticket 111 (https://jira.stsci.edu/browse/JWSTKD-111) provided the latest set of allowed SUBARRAY values for each instrument and also requested that the enum list be segregated by instrument, for ease of future maintenance. Those updates are implemented here.


nden:
Should `p_subarray` and  `subaray.schema.aml` be updated too?

hbushouse:
Yep, good point. Forgot about those. Make it so.

PR #1367, opened 2017-10-24
default value of S_REGION

nden:
Set the default to " " in the core schema. Updated the order of vertices reported in assign_wcs.


nden:
This passed on my Travis account, so I'm merging it.

Issue #1294, opened 2017-10-10
Calspec3 outlier detection needs to use resample_spec


hbushouse:
The calspec3 pipeline calls outlier_detection, which in turn is using the resample step to attempt to resample the input "images", but because these data are spectra, it should be calling resample_spec instead. This only applies to slit-like spectra, not IFU data. For IFU data outlier_detection needs to use cube_build to do resampling to a common output frame (as already documented in #1286).

stsci-hack:
The work done in #1369 allowed outlier_detection to call resample_spec for spectroscopic data, while using outlier_detection logic specific to spectroscopic data where necessary.

Issue #1286, opened 2017-10-05
Update calspec3 to call IFU-specific outlier detection step


stsci-hack:
It was reported that "when I run calspec3 on an ASN of MIRI MRS exposures...the outlier_detection step still seem to be trying to call the imaging resample routine when trying to create resampled arrays".  This resulted in the code raising an Exception as shown in the traceback shown below.  

A version of outlier-detection, therefore, needs to be developed to support this mode of data, so that calspec3 can be updated to call a working version of this step.

```
2017-10-05 14:56:11,337 - stpipe.Spec3Pipeline.outlier_detection - INFO - Performing outlier detection on 4 inputs
2017-10-05 14:56:31,132 - stpipe.Spec3Pipeline.outlier_detection - WARNING - /user/bushouse/anaconda2/envs/myjwst/lib/python2.7/site-packages/gwcs/selector.py:433: RuntimeWarning: invalid value encountered in less_equal
  temp = np.where(np.logical_or(temp <= val_range[0],

2017-10-05 14:56:31,134 - stpipe.Spec3Pipeline.outlier_detection - WARNING - /user/bushouse/anaconda2/envs/myjwst/lib/python2.7/site-packages/gwcs/selector.py:434: RuntimeWarning: invalid value encountered in greater_equal
  temp >= val_range[1]),

----------------------------------------------------------------------
ERROR RUNNING STEP 'Spec3Pipeline':
    'WCS' object has no attribute 'wcsinfo'
Traceback (most recent call last):
  File "/user/bushouse/anaconda2/envs/myjwst/lib/python2.7/site-packages/jwst-0.7.8rc3.dev10-py2.7-linux-x86_64.egg/jwst/stpipe/cmdline.py", line 320, in step_from_cmdline
    step.run(*positional)
  File "/user/bushouse/anaconda2/envs/myjwst/lib/python2.7/site-packages/jwst-0.7.8rc3.dev10-py2.7-linux-x86_64.egg/jwst/stpipe/step.py", line 370, in run
    result = self.process(*args)
  File "/user/bushouse/anaconda2/envs/myjwst/lib/python2.7/site-packages/jwst-0.7.8rc3.dev10-py2.7-linux-x86_64.egg/jwst/pipeline/calwebb_spec3.py", line 125, in process
    result = self.outlier_detection(result)
[...]

Issue #612, opened 2017-01-18
Task: Outlier detection for calwebb_spec3


hbushouse:
The calwebb_spec3 pipeline needs to include outlier detection, just as we do for calwebb_image3. This task is a placeholder for any work that might need to be done to the outlier_detection step itself to get it to work with the spectroscopic data that will be fed to it from calwebb_spec3. Some of these data have different formats than the simple 2D imaging and hence minor updates might be needed to handle those formats.

hbushouse:
This is a carryover from Trac 356 https://aeon.stsci.edu/ssb/trac/jwst/ticket/356

hbushouse:
Moving to B7.2 milestone. Updates necessary for slit-like spectra (calling resample_spec routines to accomplish outlier detection) are tracked in #1294, and updates for IFU data (calling cube_build functions) are tracked in #1286.

hbushouse:
Updates for redoing the noise model computation and use are tracked in https://github.com/STScI-JWST/jwst/issues/1759

hbushouse:
@jdavies-st What's the status of this? @hack reported in #1294 that updates had been made to call `resample_spec`, but that was before `resample_spec` was rewritten. Does it need further updates to make use of the new version, or would the existing implementation work (for 2-D spectra).

jdavies-st:
I have not worked on this at all.  I would think the new version of `resample_spec` has slotted in and should work fine here.  Though I do note that there is a known issue with MIRI FS in #2473.

But again, until we have test data and something in our regression tests, it's hard to say whether it works or not.  Do we have level 3 MIRI FS and level 3 NIRSpec simulated datasets?

I may find out later this week at the hack day.

PR #1369, opened 2017-10-25
Unify outlier step

stsci-hack:
Major update to outlier_detection step to provide a unified interface for all observation modes (exptypes).  This update also implements, for the first time, a version of outlier_detection for IFU cube data, enabling #1286 to be completed.

This update did not (intentionally) change the API for the outlier_detection step as used by any of the level 3 pipelines; in fact, the whole point was to try to maintain a single, unchanging interface for this step regardless of the exptype. 

As long as resample_spec works, this PR should (partially?) address #1294 and #612.

### DIFF
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -19,8 +19,6 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 CRBIT = np.uint32(datamodels.dqflags.pixel['JUMP_DET'])
-DEFAULT_SUFFIX = 'i2d'
-
 
 class OutlierDetection(object):
     """
@@ -43,9 +41,10 @@ class OutlierDetection(object):
       6. Updates input data model DQ arrays with mask of detected outliers.
 
     """
+    DEFAULT_SUFFIX = 'i2d'
 
     def __init__(self, input_models, 
-                    reffiles=None,  
+                    reffiles=None,
                     **pars):
         """
         Parameters
@@ -64,20 +63,14 @@ class OutlierDetection(object):
         """
         self.input_models = input_models
         self.reffiles = reffiles
-        
+
         self.num_groups = len(self.input_models.group_names)
-
-        # Parse any user-provided filename suffix for resampled products
-        self.resample_suffix = '_outlier_{}.fits'.format(pars.get('resample_suffix' , DEFAULT_SUFFIX))
-        if 'resample_suffix' in pars: del pars['resample_suffix']
-
-        self.exptype = self.input_models[0].meta.exposure.type
-                
+        self.exptype = input_models[0].meta.exposure.type 
+        
         self.outlierpars = {}
         if 'outlierpars' in reffiles:
             self._get_outlier_pars()
         self.outlierpars.update(pars)
-
 
     def _get_outlier_pars(self):
         """ Extract outlier detection parameters from reference file
@@ -118,16 +111,27 @@ class OutlierDetection(object):
         for kw in list(self.outlierpars.keys()):
             self.outlierpars[kw] = ref_model['outlierpars_table.{0}'.format(kw)]
 
+    def build_suffix(self, **pars):
+        """ Class-specific method for defining the resample_suffix attribute
+            using a suffix specific to the sub-class.
+        """
+        # Parse any user-provided filename suffix for resampled products
+        self.resample_suffix = '_outlier_{}.fits'.format(pars.get('resample_suffix' , self.DEFAULT_SUFFIX))
+        if 'resample_suffix' in pars: del pars['resample_suffix']
+        log.info("Defined output product suffix as: {}".format(self.resample_suffix))
+        
     def do_detection(self):
         """Flag outlier pixels in DQ of input images
         """
+        self.build_suffix(**self.outlierpars)
+        
         pars = self.outlierpars
         save_intermediate_results = pars['save_intermediate_results']
         if pars['resample_data'] is True:
             # Start by creating resampled/mosaic images for each group of exposures
             sdriz = resample.ResampleData(self.input_models, single=True,
                 blendheaders=False, **pars)
-            sdriz.do_drizzle()
+            sdriz.do_drizzle(**pars)
             drizzled_models = sdriz.output_models
             for model in drizzled_models:
                 model.meta.filename += self.resample_suffix
@@ -174,7 +178,6 @@ class OutlierDetection(object):
         # clean-up (just to be explicit about being finished with these results)
         del median_model, blot_models
 
-
     def create_median(self,resampled_models):
         """Create a median image from the singly resampled images.
 
@@ -189,9 +192,9 @@ class OutlierDetection(object):
         resampled_sci = [i.data for i in resampled_models]
         resampled_wht = [i.wht for i in resampled_models]
 
-        nlow = self.pars.get('nlow', 0)
-        nhigh = self.pars.get('nhigh', 0)
-        maskpt = self.pars.get('maskpt', 0.7)
+        nlow = self.outlierpars.get('nlow', 0)
+        nhigh = self.outlierpars.get('nhigh', 0)
+        maskpt = self.outlierpars.get('maskpt', 0.7)
 
         badmasks = []
         for w in resampled_wht:
@@ -213,8 +216,8 @@ class OutlierDetection(object):
     def blot_median(self, median_model):
         """Blot resampled median image back to the detector images
         """
-        interp = self.pars.get('interp', 'poly5')
-        sinscl = self.pars.get('sinscl', 1.0)
+        interp = self.outlierpars.get('interp', 'poly5')
+        sinscl = self.outlierpars.get('sinscl', 1.0)
 
         # Initialize container for output blot images
         blot_models = datamodels.ModelContainer()
@@ -267,7 +270,8 @@ class OutlierDetection(object):
         gain_models = self.reffiles['gain']
         rn_models = self.reffiles['readnoise']
 
-        for image, blot, gain, rn in zip(self.input_models, blot_models,  gain_models, rn_models):
+        for image, blot, gain, rn in zip(self.input_models, blot_models, gain_models,
+            rn_models):
             flag_cr(image, blot, gain, rn, **self.outlierpars)
 
 
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -1,10 +1,9 @@
+"""Primary code for performing outlier detection on JWST observations."""
+
 from __future__ import (division, print_function, unicode_literals,
-    absolute_import)
+                        absolute_import)
 
-
-import time
 import numpy as np
-from collections import OrderedDict
 
 from stsci.image import median
 from stsci.tools import bitmask
@@ -20,8 +19,10 @@ log.setLevel(logging.DEBUG)
 
 CRBIT = np.uint32(datamodels.dqflags.pixel['JUMP_DET'])
 
+
 class OutlierDetection(object):
-    """
+    """Main class for performing outlier detection.
+
     This is the controlling routine for the outlier detection process.
     It loads and sets the various input data and parameters needed by
     the various functions and then controls the operation of this process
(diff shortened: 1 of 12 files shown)


hbushouse:
FYI, some of us are in the midst of fixing a bug in rc5 that was delivered yesterday, which is going to require a new build sometime today or tomorrow, so in the meantime let's hold off on merging new things, just to keep the build process simple.

jwst/outlier_detection/outlier_detection.py
```diff
@@ -107,19 +111,30 @@ def _get_outlier_pars(self):
...
     def do_detection(self):
         """Flag outlier pixels in DQ of input images
         """
+        self.build_suffix(**self.outlierpars)
+        
         pars = self.outlierpars
         save_intermediate_results = pars['save_intermediate_results']
         if pars['resample_data'] is True:
             # Start by creating resampled/mosaic images for each group of exposures
             sdriz = resample.ResampleData(self.input_models, single=True,
```
hbushouse: Is this the version of ResampleData that currently only works on images? If so, will this need to be updated at some point to call a version that resamples spectroscopic "images", including both slit-like and IFU spectra, for use in the calwebb_spec3 pipeline?

jwst/outlier_detection/outlier_detection.py
```diff
@@ -107,19 +111,30 @@ def _get_outlier_pars(self):
...
                 blendheaders=False, **pars)
-            sdriz.do_drizzle(**pars)
+            sdriz.do_drizzle()
             drizzled_models = sdriz.output_models
             for model in drizzled_models:
-                model.meta.filename += "_i2d.fits"
+                model.meta.filename += self.resample_suffix
```
hbushouse: This assumes that the value of model.meta.filename only contains a root name, with no existing product suffix or "fits" extension. Is that a valid assumption?

jwst/outlier_detection/outlier_detection.py
```diff
@@ -158,108 +173,106 @@ def do_detection(self):
 
         # Perform outlier detection using statistical comparisons between
         # each original input image and its blotted version of the median image
-        detect_outliers(self.input_models, blot_models,
-            self.reffiles, **self.outlierpars)
+        self.detect_outliers(blot_models)
```
hbushouse: If `resample_data` was not true (e.g. when you're processing coronagraphic or TSO data), will `blot_models` contain the unresampled data?

jwst/outlier_detection/outlier_detection.py
```diff
@@ -145,7 +160,7 @@ def do_detection(self):
         if pars['resample_data'] is True:
```
hbushouse: ACK! You're going to get a PEP8 evil eye from @pllim on this!

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -0,0 +1,174 @@
...
+            ModelContainer containing drizzled grouped input images
+
+        reffiles : dict of `jwst.datamodels.DataModel`
+            Dictionary of datamodels.  Keys are reffile_types.
+                        
+            
+        """
+        OutlierDetection.__init__(self, input_models, reffiles=reffiles, **pars)
+        # NOTE:  Need to confirm that this attribute accurately reports the
+        #        channel 'names' for both types of IFU data; MIRI and NRS
```
hbushouse: I'm guessing meta.instrument.channel is going to be None for NIRSpec IFU exposures. NIRSpec doesn't have a concept of channels.

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def do_detection(self):
+        """Flag outlier pixels in DQ of input images
+        """
+        self.build_suffix(**self.outlierpars)
+        save_intermediate_results = self.outlierpars['save_intermediate_results']
+
+        # start by creating copies of the input data to place the separate
+        # data in after blotting the median-combined cubes for each channel
+        self.blot_models = self.input_models.copy()
```
hbushouse: The number of input models could get to be pretty large in real-life observations. Is it necessary to keep a 2nd copy of all of them that are active at once? Or can they be created and deleted one at a time as they're used?

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -0,0 +1,174 @@
...
+        for channel in range(len(self.channels)):
+            ch = self.channels[channel]
+            
+            cubestep = CubeBuildStep(config_file=cube_build_config,
+                                        channel=ch,
+                                        single='true')
+            single_IFUCube_result = cubestep.process(self.input_models)
+
+            for model in single_IFUCube_result:
+                model.meta.filename += self.resample_suffix
```
hbushouse: Same comment as before regarding appending a suffix to an existing filename.

jwst/outlier_detection/outlier_detection_scaled.py
```diff
@@ -45,6 +44,7 @@ class OutlierDetectionScaled(object):
       5. Updates input data model DQ arrays with mask of detected outliers.
 
     """
+    DEFAULT_SUFFIX = 'i2d'
```
hbushouse: Because OutlierDetectionScaled is only used for TSO exposures, where no resampling is done, is this suffix definition even necessary?

jwst/outlier_detection/outlier_detection_spec.py
```diff
@@ -0,0 +1,124 @@
...
+    def do_detection(self):
+        """Flag outlier pixels in DQ of input images
+        """
+        self.build_suffix(**self.outlierpars)
+
+        pars = self.outlierpars
+        save_intermediate_results = pars['save_intermediate_results']
+        if pars['resample_data'] is True:
+            # Start by creating resampled/mosaic images for each group of exposures
+            sdriz = resample_spec.ResampleSpecData(self.input_models, single=True,
```
hbushouse: Ah hah! Here's the resample_spec version. Ignore previous comment!

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -4,6 +4,25 @@
...
+
+# Categorize all supported versions of outlier_detection
+outlier_registry = {'imaging' : outlier_detection.OutlierDetection,
+                    'scaled' : outlier_detection_scaled.OutlierDetectionScaled,
+                    'ifu' : outlier_detection_ifu.OutlierDetectionIFU,
+                    'slitspec' : outlier_detection_spec.OutlierDetectionSpec
+                    }
+
+# Categorize all supported modes
+IMAGE_MODES = ['NRC_IMAGE', 'MIR_IMAGE', 'NRS_IMAGE', 'NIS_IMAGE', 'FGS_IMAGE']
```
hbushouse: I don't think NRS_IMAGE exposures will ever make it to level-3 processing (just used as singletons for target-acq and confirmation images), but it can't hurt to include it in the list, just in case.

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -70,9 +93,45 @@ def process(self, input):
...
+                pars['resample_data'] = False # force resampling off...
+                detection_step = outlier_registry['imaging']
+                pars['resample_suffix'] = 'i2d'
+            elif exptype in TSO_IMAGE_MODES+CORON_IMAGE_MODES:
+                # algorithm selected for TSO data (no resampling)
+                pars['resample_data'] = False # force resampling off...
+                detection_step = outlier_registry['imaging']
+                pars['resample_suffix'] = 's2d'                
+            elif exptype in TSO_IMAGE_MODES and self.scale_detection:
+                # selected scaled algorithm for TSO data
```
hbushouse: TSO images will never have any dithering, hence there's no need or desire for resampling. Also, will this `elif` clause ever get reached? If the exposures are TSO_IMAGE_MODES, won't the previous `elif` clause have already gotten triggered and hence there's never a chance of reaching this one, regardless of the value of `self.scale_detection`?

jwst/outlier_detection/outlier_detection.py
pllim: `flake8` is the one who does the actual complaining. Don't shoot the messenger! :sweat_smile:

jwst/outlier_detection/outlier_detection.py
stsci-hack: This assumption has been made based on currently available test data replicating what is believed to be produced by the pipeline.  A more general function/operation could be applied here if it was deemed an invalid assumption.

jwst/outlier_detection/outlier_detection.py
stsci-hack: The previous code already accommodates the case where resampling is done (default imaging case) and when it is NOT done (as needed for TSO or coron data).  It will actually contain the unresampled data as copies of the median model.

jwst/outlier_detection/outlier_detection_ifu.py
stsci-hack: That is what happens when test data for that mode is not available.  This can be easily supported, though.

jwst/outlier_detection/outlier_detection_ifu.py
stsci-hack: We would be able to work with one blot model at a time for outlier_detection if all the data from the input images went into the same output resampled product.  Unfortunately, MIRI data gets split into 2 products that need to be recombined in order to perform the detection on the input image.  This forces us to keep a separate copy in memory.

jwst/outlier_detection/outlier_detection_ifu.py
stsci-hack: Again, if this assumption is not valid, a more general operation can be defined.

jwst/outlier_detection/outlier_detection_scaled.py
stsci-hack: Since it is the same as used for the default imaging mode, probably not.

jwst/outlier_detection/outlier_detection_step.py
stsci-hack: I will have to go over this logic again given your comments and see whether this needs to be cleaned up or revised.

jwst/outlier_detection/outlier_detection.py
stsci-hack: Time to improve my code compliance... ;-)

jwst/outlier_detection/outlier_detection.py
```diff
@@ -107,19 +111,30 @@ def _get_outlier_pars(self):
...
             self.outlierpars[kw] = ref_model['outlierpars_table.{0}'.format(kw)]
 
+    def build_suffix(self, **pars):
+        """ Class-specific method for defining the resample_suffix attribute
+            using a suffix specific to the sub-class.
+        """
+        # Parse any user-provided filename suffix for resampled products
+        self.resample_suffix = '_outlier_{}.fits'.format(pars.get('resample_suffix' , self.DEFAULT_SUFFIX))
+        if 'resample_suffix' in pars: del pars['resample_suffix']
+        log.info("Defined output product suffix as: {}".format(self.resample_suffix))
```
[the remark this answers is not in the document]
stsci-hack: This can be reset to DEBUG output only since it does not seem useful or helpful for the user.

jwst/outlier_detection/outlier_detection_step.py
stsci-hack: This allows other packages to get the full categorization of exptypes, while also covering all possible options for currently implemented modes.

stsci-hack:
This PR was tested using regression tests for calimage3 and caltso3, as well as interactive testing using test MIRI IFU data provided by Jane.  This code is not validated to run for 2-D spectroscopic modes as needed by calspec3 (to be done for Build 7.2).

hbushouse:
Given that it hasn't been validated yet for slit-like spectra, shall we deliver the B7.1 calwebb_spec3.cfg file with outlier_detection hardwired to be skipped? Of course that would also skip it for IFU exposures too.

hbushouse:
Alternatively, DMS actually has a couple spec3 ASN's that they're using in their B7.1 testing, all of which use slit-like data (MIR_LRS-FIXEDSLIT and NRS_MSASPEC), so we could just let them test it for us. The b7.1rc6 that they're currently running is throwing errors in outlier_detection due to the resample step looking for model.meta.wcs, which for some reason doesn't exist.

PR #1371, opened 2017-10-25
Optimize reference fetch

jaytmiller:
This is a response to Jeb's performance issues when attempting to multi-process 5 or so MIRI associations concurrently.   I believe any remaining CRDS performance issues are in data model i/o or garbage collection.  This addresses that.   This changes only the crds_client glue code.

**The primary optimization is to cache the header parameters CRDS extracts from incoming data models and uses for bestrefs matching.**   Since the cache is keyed by filenames,  this scheme can be defeated by similar files with different names (pipeline progression?) or by datasets that change in ways that affect bestrefs matching (only original header counts...)   Nevertheless,  potentially it can reduce CRDS data models related file i/o by 3:1,  maybe 4:1.   This does not reduce Step file i/o.   This does not eliminate redundant but fast calls to CRDS bestrefs.  It just caches parameters.

**A secondary optimization was to remove two calls to gc.collect() for every bestrefs call.**  memory_profiler.profile() showed zero growth,  I've seen gc be surprisingly expensive,  seconds.

**Removes handling for older versions of CRDS that don't have crds_cache_locking module.**

**Additional caching to avoid repeat equivalent calls to crds.getreferences(),  strategically good even though those are currently cheap for most CRDS configurations.**

### DIFF
--- a/jwst/stpipe/crds_client.py
+++ b/jwst/stpipe/crds_client.py
@@ -69,18 +69,21 @@ def get_multiple_reference_paths(dataset_model, reference_file_types):
     key = (filename, tuple(reference_file_types))
     
     try:
-        refpaths = _BESTREFS_CACHE[key]
+
+        refpaths = { reftype:_BESTREFS_CACHE[(filename, reftype)]
+                     for reftype in reference_file_types }
         log.verbose("Using cached bestrefs for", repr(filename), "with types", repr(reference_file_types))
+        
     except KeyError:
 
         data_dict = _get_data_dict(filename, dataset_model)
 
         # Cache prefetch-like results
-        _BESTREFS_CACHE[key] = refpaths = _get_refpaths(data_dict, tuple(reference_file_types))
+        refpaths = _get_refpaths(data_dict, tuple(reference_file_types))
 
         # Cache results for each individual reftype, as-in get_reference_file().
         for reftype, path in refpaths.items():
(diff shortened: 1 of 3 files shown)


jaytmiller:
I think this does achieve at least a 2:1 improvement in file i/o and CRDS performance just within a single Step like assign_wcs.  For a pipeline,  it may be 3:1 since the pipeline does a prefetch for all steps at once,  then launches each Step which also does a prefetch and normal fetch for each type.

jaytmiller:
I did this pull request now (and it passed Travis too fast,  what's up?) but I'm really just suggesting this if there's a B7.1 lull or maybe for B7.2.   It might be useful for Jeb in early B7.2 DEV until/if a better rewrite of stpipe is done in general.   Another approach is to keep the prefetch results and/or pass them down from Pipeline to Step.  This seemed easier to do though,  the crds_client code is less intricate and with fewer concerns.

jaytmiller:
I made another pass today and was able to simplify more code and add caching for the "theoretically fast" calls to the CRDS core library in addition to the previous data models header caching.  I also eliminated some duplicate pre-fetch code in pipeline.py.   Potentially rewriting Pipeline and Step to just optimally call crds_client to start with would be simpler and better,  you don't have to cache calls you don't make;  but this is very close to equivalent now,  and in a sense the rewrite would just move caching from crds_client to Pipeline and Step.   This may now be good enough for a few years to forever.

PR #1372, opened 2017-10-25
More fixes to get set_telescope_pointing to work

hbushouse:
The changes in #1370 resolved the issue of having two SCI extensions appear in files updated by set_telescope_pointing, but it still resulted in an output file that had the DATAMODL keyword set to 'DataModel'. This causes failures in the `calwebb_detector1` pipeline (and steps), because they all use the generic `datamodels.open()` function to load their inputs. It was loading them into a DataModel, which doesn't work. They need to load into a RampModel.

These changes have the effect of removing the DATAMODL keyword from the files produced by set_telescope_pointing, so that the subsequent calls to `datamodels.open()` rely on file structure info to correctly determine that they should use RampModel.

This also allowed for a change to set_telescope_pointing to have it use the appropriate `Level1bModel` for input, instead of `DataModel`. `Level1bModel` already has the wcsinfo.schema included and hence there's no need to extend the schema to include the WCS keywords.

A change was also made to the Level1bModel itself, such that it no longer automatically creates a ZEROFRAME array/extension if there isn't one in the input. This is not needed or desired. If there's no ZEROFRAME in the input, don't create one (just like for REFOUT).

Finally, because of the change to fits_support in #1370, there's no longer a guarantee that the SCI extension will always be the first extension HDU in a FITS file, so the set_velocity_abberation script has been updated to explicitly use the SCI extension (rather than extension 1).

I have tested these changes on several level-1b files provided by Mike Swam and the updated _uncal.fits files come out correctly and they in turn process through calwebb_detector1 correctly.

### DIFF
--- a/jwst/datamodels/level1b.py
+++ b/jwst/datamodels/level1b.py
@@ -56,8 +56,8 @@ class Level1bModel(model_base.DataModel):
         #
         # TODO: Hacky. Need solution which involves schema
         # specification and embedded in DataModel.
-        if 'zeroframe' not in self.instance and \
-           'data' in self.instance and \
-           len(self.data.shape) == 4:
-            nints, ngroups, ny, nx = self.data.shape
-            self.zeroframe = np.zeros((nints, ny, nx))
+        #if 'zeroframe' not in self.instance and \
+        #   'data' in self.instance and \
+        #   len(self.data.shape) == 4:
+        #    nints, ngroups, ny, nx = self.data.shape
+        #    self.zeroframe = np.zeros((nints, ny, nx))
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -284,7 +284,8 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
         current_date = Time(datetime.datetime.now())
         current_date.format = 'isot'
         self.meta.date = current_date.value
-        self.meta.model_type = self.__class__.__name__
+        if self.meta.model_type is not None:
+            self.meta.model_type = self.__class__.__name__
 
     def save(self, path, *args, **kwargs):
         """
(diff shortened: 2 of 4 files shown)


hbushouse:
Of course eventually (when time permits) the appropriate solution to this would be to allow the level-1b files to have `DATAMODL="Level1bModel"` in their headers, because they are in fact level-1b models, and modify the calwebb_detector1 pipeline and all its steps to explicitly open their inputs as RampModel. They all must have that kind of model as input and hence they should explicitly load them that way to begin with. If someone tries to use an input that doesn't conform to RampModel, it should throw an error, as is appropriate.

hbushouse:
Interesting idea. That might be just as appropriate.

Sent from my Verizon Wireless 4G LTE DROID


Bernie Simon <notifications@github.com> wrote:


@bernie-simon commented on this pull request.

I would rather check to see if the class name is "DataModel" and not write meta.model_type if that is the case. There is no point that I can see in setting DATAMODL to "DataModel", it's the generic case.

hbushouse:
Yep, failure is on test of a no zeroframe situation, which now fails because we're no longer creating a zeroframe when one doesn't exist in the input. So I'm going to merge, so that we can get another build+test run kicked off.

bernie-simon:
Okay, I think best when staring at the ceiling in the wee hours of the morning, so here are some thoughts on datamodels.  You can solve the ordering problem with writing out extensions by moving _save_extra_fits after _save_form_schema, as before, and modifying _save_extra_fits to call _get_hdu or _make_new_hdu. (Not _get_or_make_hdu. I know, it's confusing.) 

Meta.model_type is only going to be set correctly if it's opened with the correct class initializer, not with the generic open in datamodels.util. So datamodels.util.open should not be setting model_type, that is, it should unset it if it was not originally set and on_save should respect that. The only code that uses meta.model_type is calwebb_spec3. It's unclear to me what the effect of not having it set would be.

For now I have no plans to change datamodels but will come back to it later.

PR #1376, opened 2017-10-26
Update WFS&C checks to use `VISITYPE` instead of `WFSVISIT`

### LINKED ISSUE #1340: Rule: WFS&C
A couple of changes are coming with regard to WFS&C exposures and associations:

- [remove WFSVISIT keyword from all headers](https://jira.stsci.edu/browse/JWSTKD-114)
- [replace asn pool column WFSVISIT with VISITYPE](https://jira.stsci.edu/browse/JSDP-287)

Should look for an APT v25.4 for the correct information.

stscieisenhamer:
Resolves #1340 

Changes:
- Use `VISITYPE` for WFS&C determination
- Expand check on values for `VISITYPE`
- Update `AssociationPool` to force `null` on undefined/masked columns
- Update the test `mega_pool.csv` for all the new test cases created recently.


pllim:
The time out happens at `jwst/associations/tests/test_level3_product_names.py ...` which points to `test_level35_names(pool_file)` (more or less). Your diff suggests that you did not modify the test, so can you think of any changes that would adversely affect this test in only Python 3?

For example, any inconsistency of `bytes`/`str` handling between PY2 and PY3 here? If you do `from __future__ import unicode_literals`, do you start seeing problem in Python 2 as well? I am just guessing here, so if this doesn't make sense, feel free to ignore.

stscieisenhamer:
@pllim explain how it is a "timeout"?  Usually there is a very clear message about timeouts; this just dies?  And yes, note that the 2.7 test runs to completion.

And also note that, if there is a way of seeing the history of builds,  the kill was happening at test_io. The only change here is to remove all tests from the file test_io. Now the kill is happening at the next test file, which happens to be test_level3_product_names. This behavior is indicative of some sort of timeout, but its a timeout of the whole run and not of the usual "no output seen" timeout.  So, is this typical of a timeout and what is this timeout based on? Because I have had many a timeout of the flavor of "no output seen", but never this.

stscieisenhamer:
It does seem this "just die off" is some type of timeout. Add `runslow` to a number of tests and will migrate everything to pandokia.

PR #1377, opened 2017-10-26
EXP: Update .travis.yml

pllim:
I might or might not be procrastinating...


.travis.yml
```diff
@@ -41,25 +41,19 @@ matrix:
...
 
         # build sphinx documentation with warnings
         - os: linux
           env: PYTHON_VERSION=3.5
-               SETUP_CMD='build_sphinx'
+               SETUP_CMD='build_docs'
                CONDA_DEPENDENCIES=$CONDA_JWST_DEPENDENCIES
                PIP_DEPENDENCIES='sphinx_rtd_theme sphinx-automodapi'
 
-    allow_failures:
```
hbushouse: Does this mean that PEP8 failures will no longer be classified as allowed and hence Travis will report that all tests have failed?

pllim: Yes, unless it is one of the many ignored codes as stated in `setup.cfg` under `[flake8]` section. This is for the best because in my recent PEP8 campaign over at Ginga, I found several actual bugs hiding in the `flake8` warnings (e.g., undefined variables).

docs/jwst/introduction.rst
```diff
@@ -26,7 +26,7 @@ and :ref:`stage2-spectroscopic-flow`.
 Stage 3 processing consists of routines that combine the data from
 multiple exposures, for all observing modes.
 Details of stage 3 imaging processing are at
-:ref:`stage3-imaging-flow`.
+:ref:`stage3-imaging-flow-blahblah`.
```
hbushouse: Why is this reference being changed to include "blahblah"? I don't think that's a reference to an actual doc anywhere.

pllim: This is just to see if `build_docs` (or `build_sphinx`) is doing what it is supposed to do, i.e., to catch any Sphinx warnings. This is because I don't see you have the `-w` option as `astropy` has, so not sure if that is oversight or not, as you don't use `astropy-helpers`. I expect this commit to cause `build_docs` to fail. If it does not, you have a problem. This particular change should not be merged.

pllim: https://travis-ci.org/STScI-JWST/jwst/jobs/293305788 did not fail despite "544 warnings" stated in its log. That is a bit concerning.

.travis.yml
pllim: By the way, there is quite a lot of undefined variables...

pllim:
These need to be fixed. I don't think they should be ignored:
https://travis-ci.org/STScI-JWST/jwst/jobs/293404067

philhodge:
Some of these are real bugs.

But what about this (there were many of these):
jwst/ramp_fitting/ramp_fit.py:1509:33: E712 comparison to True should be 'if cond is True:' or 'if cond:'

The first recommendation (if cond is True:) is a direct contradiction to PEP 8, which says:
Yes:   if greeting:
No:    if greeting == True:
Worse: if greeting is True:

jdavies-st:
From [PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations):

"Comparisons to singletons like None should always be done with `is` or `is not`, never the equality operators."

nden:
@philhodge `E712 comparison to True should be 'if cond is True:' or 'if cond:'` are bugs too, since we want to compare values not identity of objects and `is` compares identity.

jdavies-st:
Some discussion of this [here](https://stackoverflow.com/questions/4050335/strange-pep8-recommandation-on-comparing-boolean-values-to-true-or-false).

pllim:
Fun fact, when `flake8` is used instead of `pycodestyle` (in the same allowed failure mode with no `ignore=`), there are 8604 warnings (was 1987).

nden:
I think this can be closed.

PR #1386, opened 2017-10-27
Made GWATILT robust against non-float input.  Added code so that prop…

stscirij:
…osal id from .prop file is either truncated or padded to 5 characters

Also made subarray name 'GENERIC' if the parameters don't match those of known subarrays


jwst/fits_generator/create_dms_data.py
```diff
@@ -111,7 +111,10 @@ def get_subarray_name(instrument, colstart, colstop, rowstart, rowstop):
         if rowstart == subarray[2] and rowstop == subarray[3] and colstart == subarray[0] \
                 and colstop == subarray[1]:
             return subarray[4]
-    return 'UNKNOWN'
+    #
+    # If we can't match the subarray parameters we return 'GENERIC', which is
+    # a legal subarray name
+    return 'GENERIC'
```
hbushouse: Yes, it's a legal value, but technically only for reference files, never for science products. But what else can you do. It'll probably still cause errors when running the pipeline and calling bestref to find ref files, just like UNKNOWN. But at least it passes schema validation.

jwst/fits_generator/templates/nirspec_configuration.gen.inc
```diff
@@ -4,7 +4,7 @@ FILTER    = input('FWA_POS') or 'CLEAR'   /  name of the filter element used
 GRATING   = input('GWA_POS') or 'N/A'  /     name of the grating element used
 GWA_XTIL  = input('GWA_XTIL') or 0.0  /       grating x tilt
 GWA_YTIL  = input('GWA_YTIL') or 0.0  /       grating y tilt
-GWA_TILT  = input('GWA_TTIL') or 0.0 /    GWA temperature ave [K]
+GWA_TILT  = float(input('GWA_TTIL')) or 0.0 /    GWA temperature ave [K]
```
hbushouse: Yeah, the NIRSpec ground test images from FITSWriter just love to encode all their numerical values into a string. If I remember correctly they do it with all of the GWA* keyword values, so it might not hurt to make them all robust.

jwst/fits_generator/create_dms_data.py
stscirij: The schema validation was what I was aiming for

jdavies-st:
So since this functionality has been implemented in a different PR, #2494, can we close this PR?

stscirij:
These changes were implemented in #1946 and #2494, closing

Issue #63, opened 2016-06-14
Update BUNIT in all output products


hbushouse:
SDP is including the keyword BUNIT in the SCI extension header of the level-1b products that serve as input to the cal pipeline. They have its value set to 'DN' in the level-1b products. We don't currently have BUNIT in our datamodels schema and hence that keyword simply gets passed along untouched (presumably via the extra_fits attribute of our data models) to all the output files created by the cal pipeline. But the value of 'DN' is no longer correct for those output products and can be misleading. So we need to add that keyword to our core schema and make sure it gets updated when necessary.

Normally this would be an easy thing to implement, but given that the keyword resides in the SCI extension header, rather than the primary header, I'm not exactly sure how we can implement or specify it in our core schema. Is there a way to designate the fits_hdu to which a given keyword belongs?

bernie-simon:
On Jun 14, 2016, at 1:47 PM, Howard Bushouse <notifications@github.com> wrote:

>  Is there a way to designate the fits_hdu to which a given keyword belongs?

Yes, you just add a fits_hdu tag to the schema, as in this example from multislit.schema.yaml:

          source_id:
            title: Source ID
            type: integer
            default: 0
            fits_keyword: SOURCEID
            fits_hdu: SCI

hbushouse:
#1389 includes updates to ramp_fit to set BUNIT to 'DN/s' in _rate and _rateints products. Products stay in those units until they are converted to calibrated fluxes. This happens, if at all, in different places for different modes of data.

Images: Currently not converted to calibrated fluxes, so all we need to do is verify that BUNIT='DN/s' is being carried along to all subsequent data products, like the _cal and _calints products coming out of `calwebb_image2` and the _i2d products coming out of the `resample` step.

Spectra - IFU: They get converted to calibrated fluxes by the `photom` step in `calwebb_spec2`, which does update BUNIT appropriately in the _cal products. The `cube_build` step appears to be propagating these values to the _s3d products, so this mode appears to be complete.

Spectra - non-IFU: It appears that the `resample_spec` step is NOT propagating the BUNIT keyword at all to _s2d products, so this needs to be fixed. These data get converted to flux calibrated units during the `extract_1d` step, so that step should be setting appropriate units for each column of the `MultiSpecModel` (_x1d) product tables that it creates.

hbushouse:
Fixed by #1633.

PR #1389, opened 2017-10-30
Update calwebb_detector1 and steps to use RampModel input

hbushouse:
These changes update the `calwebb_detector1` pipeline and (nearly) all of its steps to explicitly open the input data as a `RampModel`, rather than using the generic `datamodels.open()` function. All steps are setup to work only on a `RampModel` in that they either need to use data spread across multiple groups or integrations, or rely on data arrays like `groupdq` and `pixeldq`, all of which are only present in a `RampModel`. There is no way, therefore, that any of these steps could ever be applied to any kind of simpler model, like `ImageModel` or `CubeModel`, hence there's no reason to allow for any kind of input other than `RampModel`.

The only step in the `calwebb_detector1` pipeline that does not work on 4-D `RampModel` data is the final step `gain_scale`, which is applied after the `ramp_fit` step, and hence works on an `ImageModel` or `CubeModel`. So it still uses the generic `datamodels.open()` function on its input.

Note that while we have an explicit `MIRIRampModel` defined in datamodels, the only difference between it and the generic `RampModel` is the presence of the extra data array `REFOUT` (contains values from the MIRI detector reference output). Currently the `REFOUT` extension is not used in any calibration pipeline steps and is not passed along to any of the `calwebb_detector1` output products, so there's no harm in not loading it from the input file to this pipeline. We could, if desired, update the `RampModel` definition to include the `REFOUT` extension, but in a way that it would be optional, in which case it would get loaded only when it's actually present in the input FITS file. At that point the dedicated `MIRIRampModel` would no longer be needed.

These changes address the fundamental issue encountered in #1372, in which the DATAMODL keyword in level-1b files created by the `set_telescope_pointing` pointing script are set to `Level1bModel`, which when opened with the generic `datamodels.open()` function was causing the data to be loaded into a `Level1bModel`, which doesn't work with the `calwebb_detector1` pipeline (must be a `RampModel`). Now that the pipeline and steps have been fixed to explicitly load the data into the proper type of model, the tricks employed in `set_telescope_pointing` and `datamodels.model_base` to remove the DATAMODL keyword could be backed out to allow the level-1b files to have `DATAMODL="Level1bModel"` in their headers.

One other minor updated included here is to set the value of the BUNIT keyword in the `ramp_fit` output products to be `DN/s`, which partially fulfills #63.

### DIFF
--- a/jwst/dark_current/dark_current_step.py
+++ b/jwst/dark_current/dark_current_step.py
@@ -18,7 +18,7 @@ class DarkCurrentStep(Step):
     def process(self, input):
 
         # Open the input data model
-        with datamodels.open(input) as input_model:
+        with datamodels.RampModel(input) as input_model:
 
             # Get the name of the dark reference file to use
             self.dark_name = self.get_reference_file(input_model, 'dark')
--- a/jwst/dq_init/dq_init_step.py
+++ b/jwst/dq_init/dq_init_step.py
@@ -17,7 +17,7 @@ class DQInitStep(Step):
 
     def process(self, input):
         # Determine Model type from EXP_TYPE of RampModel
-        input_model = datamodels.open(input)
+        input_model = datamodels.RampModel(input)
 
         # Set flag for guider mode operations
         if input_model.meta.exposure.type in dq_initialization.guider_list: # Reopen as GuiderRawModel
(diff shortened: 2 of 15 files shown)


hbushouse:
Everything's passing in my tests, so I'm merging.

PR #1395, opened 2017-10-30
ENH: Update Numpy, Python, and use astropy dev for PY3

pllim:
UPDATE: Now that Howard concluded the timeout was Travis' fault. I updated this PR to use Numpy "stable" (this is set by `ci-helpers` to latest Numpy release), astropy dev for PY3 (same as pandokia), and Python 3.6 (why not?). And also `flake8` like #1377 but still as allowed failure.

Python 3.6 change log: https://docs.python.org/3/whatsnew/3.6.html

Numpy 1.13 change log: https://docs.scipy.org/doc/numpy-1.13.0/release.html

~~Similar to #1391 but with some Travis CI voodoo.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,13 +23,15 @@ env:
         - PIP_DEPENDENCIES=''
         - CRDS_SERVER_URL='https://jwst-crds.stsci.edu'
         - CRDS_PATH='/tmp/crds_cache'
-        - NUMPY_VERSION=1.12
+        - PYTHON_VERSION=3.6
+        - NUMPY_VERSION=stable
+        - ASTROPY_VERSION=development
 
     matrix:
-        - PYTHON_VERSION=2.7 SETUP_CMD='install'
-        - PYTHON_VERSION=2.7 SETUP_CMD='test'
-        - PYTHON_VERSION=3.5 SETUP_CMD='install'
-        - PYTHON_VERSION=3.5 SETUP_CMD='test'
+        - SETUP_CMD='install'
+        - SETUP_CMD='test'
+        - PYTHON_VERSION=2.7 SETUP_CMD='install' ASTROPY_VERSION=stable
+        - PYTHON_VERSION=2.7 SETUP_CMD='test' ASTROPY_VERSION=stable
 
 matrix:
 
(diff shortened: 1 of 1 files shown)


pllim:
@stscieisenhamer , https://travis-ci.org/STScI-JWST/jwst/builds/295038965 did not time out...

pllim:
#1396 times out but not here. Why??? I'll undo one change at a time from here and track it down.

pllim:
I can't get it to time out. Argh...

hbushouse:
Unfortunately it appears that the py3.5 Travis tests are now failing again! So perhaps we do still need some tweaking to the py3.5 configuration?

pllim:
Why not just upgrade to Python 3.6 on Travis and get it over with? I can't get it to timeout... 😄 

You still test with Python 3.5 on pandokia, so indirectly, both Python 3.5 and 3.6 will be covered (more or less).

pllim:
FWIW when I played with the configs last night (logs below), the culprit seems to be Python 3.5 somehow. I cannot explain why though.

https://travis-ci.org/STScI-JWST/jwst/builds/295122334

https://travis-ci.org/STScI-JWST/jwst/builds/295123609

hbushouse:
I guess it can't hurt to at least try python3.6 and see what happens. Would be nice to get tests to pass under some version of python3. But I'm not sure about numpy version issue. @nden do you have an idea as to whether we should/need to keep in pinned to 1.12 or can we move to "stable"?

nden:
This looks reasonable to me. 👍 on merging it.

.travis.yml
```diff
@@ -23,41 +23,39 @@ env:
         - PIP_DEPENDENCIES=''
         - CRDS_SERVER_URL='https://jwst-crds.stsci.edu'
         - CRDS_PATH='/tmp/crds_cache'
-        - NUMPY_VERSION=1.12
+        - PYTHON_VERSION=3.6
```
pllim: @hbushouse , I made Python 3.6 the default Python version unless stated otherwise here.

nden:
python 3.6 is now a global default. 
IIRC there was a build problem with np 1.13 in the past which appears to be resolved now.

nden:
Still interesting why it's failing on py 3.5 though.

hbushouse:
OK. Go for it!

pllim:
@drdavella said something was funky with XML parsing in Python 3.5 (my memory is fuzzy, so maybe he can clarify). Incidentally, I notice that you don't use `from__future__ import unicode_literals` consistently. But these two statements might or might not be relevant.

tl;dr ¯\_(ツ)_/¯

(I think I have procrastinated enough over here, I'll go back to my day job now...)

nden:
I found this issue: 
https://github.com/travis-ci/travis-ci/issues/8172

May be it is a Travis issue after all, and not python.

pllim:
I can see the shiny green merge button but I usually don't merge my own PR if I can help it. 😅

pllim:
@nden, rebase your PR and find out...

nden:
@hbushouse was right, py 3 tests have disappeared

pllim:
What do you mean? Can you point me to a build URL?

nden:
https://travis-ci.org/STScI-JWST/jwst

pllim:
The first two jobs are Python 3.6. The Python version is not listed because it is global now. If you look at the log after completion, you will see that it used Python 3.6.x.

nden:
Oh right, I missed that.

pllim:
You can also click on a job and then the "View config" tab but it is less obvious that way (at least to me).

nden:
Travis recommends using `sudo: required` for jobs which take many resources. I tried this with the previous version of `.travis.yml` - it failed again but gave an error message:
```
jwst/associations/tests/test_level3_product_names.py ...

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```
Does this test really not produce output for 10 min?

nden:
Here's the log if anyone's interested:
https://travis-ci.org/nden/jwst/jobs/295328846

sosey:
perhaps it's time to play with a local travis container to debug this? I started down this path yesterday but realized I was just procrastinating my other work...if anyone else wants to give it a go there are decent instructions here, I managed to install everything into the container, but hadn't played with the test yet:  https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image

pllim:
Re: jwst/associations/tests/test_level3_product_names.py

@stscieisenhamer reported that it timed out on a test, so he disabled that test, then it timed out on a different test...

stscieisenhamer:
@nden that particular test normally does not run that long, so that is something new.  The issue we've actually been having is something killing the job/destroying the container, which itself is almost certainly not a timeout issue, though we've been using that term. It is possibly a overuse-of-resource issue, but nothing is telling us what that is. The travis issue you pointed out actually seems to be the related problem. 

However, now its working? At least the latest log show this.

PR #1396, opened 2017-10-30
fix order of data points in s_region

nden:
This fixes the order of RA, DEC in the `S_REGION` keyword as computed by assign_wcs and extract_2d. The order is correct in set_telescope_pointing.

Also the region is not closed any more (as requested by the archive).


hbushouse:
Travis CI with py3.5 is running!

pllim:
4 failures is a step from time out!

PR #1399, opened 2017-10-31
Backout out pr# 1376 just to see if it works

stscieisenhamer:


pllim:
Python 3.6 passed!

nden:
It was failing with py 3.5, right? I believe DMS is running 3.5.

pllim:
OPS... Sorry for changing the YML (too soon?). You can set the global Python version here back to 3.5 in `.travis.yml` if you want to test that. You still won't see that version reported in the build summary but trust me.

hbushouse:
Yes, I believe DMS is running py3.5, but so are our regression tests on jwcalibdev, and neither of them have been showing any kind of errors with any processing. Everyone's local Travis test environments are also not showing any errors. So it does really appear to be something that's confined to the py3.5 tests on the Travis servers.

nden:
I actually can't run the tests locally, started a fresh conda env  and still no luck. probably just me though.

pllim:
> neither of them have been showing any kind of errors with any processing

Ehem... It is not 100% pass on pandokia...

pllim:
Note on pandokia: When a test times out, sometimes it won't show up as missing nor failed on the report. It is simply *not there*. Always check if the *total* tests is the number you expected. If you don't believe me, ask Michele.

hbushouse:
@pllim I am well aware of the tests that are throwing errors or failing on Pandokia (I review them at least 2-3 times EVERY DAY). The two persistent errors are due to errors in the test modules themselves that have yet to be fixed by the responsible developers and have nothing to do with any of the issues we've been seeing here. And the frequent failures are by design. When we make a change in the code, we expect and want to see resulting changes in the outputs. That's how we verify that the code changes did what we expected and did not do anything unexpected. What would be the point of having regression tests that never showed any changes? We might as well not have them run at all.

PR #1402, opened 2017-10-31
Add Python 3.5 back but as allowed failure

pllim:
Follow up of #1395.

There seems to be concern that Python 3.5 is no longer tested. However, Python 3.5 mysteriously times out on Travis CI. So this PR adds it back but as an allowed failure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,9 @@ matrix:
 
     include:
 
-        - os: linux
+        - os: osx
           env: PYTHON_VERSION=3.5 SETUP_CMD='test'
+          compiler: clang
 
         # PEP8 check with flake8 (only once, i.e. "os: linux")
         - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,9 @@ matrix:
 
     include:
 
+        - os: linux
+          env: PYTHON_VERSION=3.5 SETUP_CMD='test'
+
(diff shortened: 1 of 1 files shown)


pllim:
Python 3.5 is still timing out, by the way.

pllim:
Thoughts: If it is really a bug that only manifests in Python 3.5 and not Travis Linux issue, then theoretically switching that job to Mac should pass? However, introducing Mac build should be avoided if possible. Do you want me to change the OS to Mac for a test run?

stscieisenhamer:
FYI: There is definitely a performance difference between 3.5 and 3.6. In general, the association tests run ~50% faster in 3.6 on my osx. However, for any of the tests currently not skipped, the 3.5 runs don't take more than 3-4 minutes. So, if this is a time resource issue, it is an issue for the Travis containers that is being exacerbated by the fact the 3.5 is a bit slow. How 3.5 compares to previous pythons is another question.

pllim:
@stscieisenhamer might be right... it took 30 minutes on a Mac build -- https://travis-ci.org/STScI-JWST/jwst/jobs/295401945 . For comparison, it took about 15 mins in both Python 2.7 and Python 3.6. Whatever you have in the pipeline somehow tripped some performance issue in Python 3.5.

Shall we close without merge this then?

pllim:
Knowing this... Maybe it is time to ask DMS to upgrade to Python 3.6.

PR #1405, opened 2017-11-01
Systemic updates for version numbers and others

hbushouse:
Updated the __version__ number everywhere it currently appears in all steps and pipelines to get them ready for the final B7.1 delivery. Most instances were updated to a value of "0.7.1". In some cases the existing number was already ahead of that value, e.g. "0.7.3" or "0.7.8", in which case the last value was preserved by adding "1" as the "patch" version number (in the system major.minor.patch) and moving the existing value to a 4th digit, e.g. "0.7.1.3" or "0.7.1.8". Once we make our first public release (i.e. beyond B7.2) I plan to change all of these to start out fresh at "1.0.0", after which we can continue to use the scheme in the way it was intended.

Also removed the ability to call via __main__ from the many step classes that still had that in place. I sure hope no one is trying to use those calls anywhere.

Finally, a few step classes were updated to close their input data models before returning.


jwst/assign_wcs/__init__.py
```diff
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 from .assign_wcs_step import AssignWcsStep
 
-__version__ = '0.7.4'
+__version__ = '0.7.1.4'
```
nden: This doesn't look right, a typo?

hbushouse: As mentioned in the description, I took cases where the existing version number - in this case 0.7.4 - was already ahead of 0.7.1 and tried to preserve it some fashion by moving the 3rd digit to a 4th digit, hence 0.7.4 becomes 0.7.1.4.

If folks don't want to use 4 digits where necessary (or think it'll actually cause problems), I can just set them all to 0.7.1 regardless of what their latest value had been.

nden: Sorry, I didn't see the initial description.
Perhpas it's time to start using semantic versioning and a scheme that's recognized in general. How about setting all versions to `0.8.0` and incresing the numbers in the future following rules for major.minor.patch. It has the benfit of communicating the significance of a release while tying the version to internal builds means nothing to most users. 
Perhaps software tools can take advantage of the semantic versioning as well.

hbushouse: What do others think of Nadia's suggestion to just set all versions to 0.8.0 for B7.1, which could then go to 0.9.0 for B7.2, and eventually 1.0.0 for the first public release.

stscieisenhamer: i don't believe the software versions for anyone else in dms corresponds to the "build identifier" anyways.

dmggh: Seems reasonable to me.

Issue #1407, opened 2017-11-01
Update documentation for extract_1d and flat_field


philhodge:
The documentation for `extract_1d` should mention support for GRISM data.  There are several other supported models that are not listed.

The documentation for `flat_field` should mention NIRSpec imaging data and BRIGHTOBJ data.

hbushouse:
Fixed by #1408.

PR #1408, opened 2017-11-01
Update documentation for extract_1d and flat_field

philhodge:
Support for GRISM data in `extract_1d` was mentioned.  The `flat_field` step can process NIRSpec imaging data and BRIGHTOBJ data.


hbushouse:
Fixes #1407.

PR #1413, opened 2017-11-02
Preserve slit order in exp_to_source

### LINKED ISSUE #1390: MultiProductModel FITS files have extensions in random order after resample_spec
`resample_spec` takes in `MultiSlitModel` files and generates `MultiProductModel` files.  The order of the slits in the resulting FITS files is not retained.

jdavies-st:
Resolves #1390


hbushouse:
Is it OK to merge this, so that it'll get into the 6pm build and 7pm regression test run?

jdavies-st:
I'm fine with merging.  @stscieisenhamer?

PR #1417, opened 2017-11-03
Add ctype and cunit for the spectral axis

nden:
This adds CTYPE3 and CUNIT3 values if `WCSAXES==3`. 
It may be moved to SDP processing for build 7.2.


nden:
@hbushouse I set the milestone to 7.1 It's up to you if you want to include it in the next delivery. 
the call is wrapped in a try ... except block because I'm extra paranoid not to break anything at this point.

hbushouse:
Feeling brave today (what could possibly go wrong?). Merging.

Issue #1381, opened 2017-10-26
outlier_detection intermediate product units are anomalous


jdavies-st:
The intermediate products from `outlier_detection` that are produced when the arg `save_intermediate_files=True` show that the units in the output image are way off.
```
In [1]: from jwst import datamodels

In [2]: cal = datamodels.open('nrca5_47Tuc_primary_dither3_newpos_cal.fits')

In [3]: blot = datamodels.open('nrca5_47Tuc_primary_dither3_newpos_blot.fits')

In [6]: resamp = datamodels.open('jw99999001001_01105_1_nircam_long_outlier_i2d.fits')

In [7]: median = datamodels.open('nrca5_47Tuc_median.fits')

In [8]: cal.data.sum()
Out[8]: 9311076.0

In [9]: blot.data.sum()
Out[9]: 1.5816919e+16

In [10]: resamp.data.sum()
Out[10]: 1.6033523e+16

In [11]: median.data.sum()
Out[11]: 1.7541661e+16


```
This doesn't seem to be a problem in the normal `resample` process and seems to be confined to `outlier_detection`.  Obviously this means that when we do the comparison between the blotted median and each input image, nothing is getting flagged.

PR #1419, opened 2017-11-03
initial version with resample

stsci-hack:
This enables resampling in CALIMAGE2 and now requires a VALID WCS in order for the resampling to work.

### DIFF
--- a/jwst/pipeline/calwebb_image2.py
+++ b/jwst/pipeline/calwebb_image2.py
@@ -128,11 +128,12 @@ class Image2Pipeline(Pipeline):
         input = self.flat_field(input)
         input = self.photom(input)
 
+        
         # Resample individual exposures
         result = self.resample(input)
-
-        # write out resampled exposure
-        self.save_model(result, suffix='i2d')
+        if result:
+            # write out resampled exposure
+            self.save_model(result, suffix='i2d')
         
         # That's all folks
         self.log.info(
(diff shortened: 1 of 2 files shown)


jwst/pipeline/calwebb_image2.py
```diff
@@ -10,6 +10,7 @@
 from ..assign_wcs import assign_wcs_step
 from ..flatfield import flat_field_step
 from ..photom import photom_step
+from ..resample import resample_step
 
 
 __version__ = "3.0"
```
hbushouse: This version number is probably in conflict with what's in the repo now (0.8.0).

jwst/pipeline/calwebb_image2.py
```diff
@@ -126,6 +128,12 @@ def process_exposure_product(
         input = self.flat_field(input)
         input = self.photom(input)
 
+        # Resample individual exposures
+        result = self.resample(input)
+
+        # write out resampled exposure
+        self.save_model(result, suffix='i2d')
```
hbushouse: Might need to put this in an IF-block and only call it if `result` exists. In the event a user skips the resample step, it won't exist.

hbushouse:
Fixes #687.

jwst/pipeline/calwebb_image2.py
stsci-hack: Latest commit addresses this last comment.

jdavies-st:
Simulated NIRCam data ran through the `calwebb_image2` pipeline without problem, producing a nice rectified `i2d` resampled output.  I did notice that that flux values are much too high, but I think this is related to issue #1381, which can be tracked there.  This is OK to merge.

PR #1422, opened 2017-11-03
Turn off resampling slitless modes in spec2pipeline

jdavies-st:


sosey:
lgtm - I'm going to merge this so that I can run my final test for grisms....

PR #1436, opened 2017-11-09
TSO bug and bad pool issues

stscieisenhamer:
Changes
- Current pool has bad values for `TSOVISIT` so also base TSO rules off of EXP_TYPE
- Bug fix where pool values are different than expected which caused no Level2b associations to be created


stscieisenhamer:
Test failure related to CRDS and an attempted access to the internet. @jaytmiller ?

jaytmiller:
We concluded this was proably a transient outage on the JWST OPS server required for this testing in the Travis env setup,  maybe the nightly reboot or something similar.

PR #1443, opened 2017-11-09
Update tests; no functional changes required

stscieisenhamer:


stscieisenhamer:
@jaytmiller doing the same thing.

stscieisenhamer:
Failure unrelated to code. Merging.

jaytmiller:
I just did a pingdom test and it looked OK.  So outside connectivity is at least intermittent.  But I also noticed in the warnings that Travis expects us to declare access to outside websites (CRDS)... and we've never done that.  If there's some technical or security reason for that declaration,  it may have come to a head.

jaytmiller:
This:

E   crds.core.exceptions.ServiceError: CRDS jsonrpc failure 'get_server_info' <urlopen error An attempt was made to connect to the internet by a test that was not marked `remote_data`. The requested host was: jwst-crds.stsci.edu>


may be new cleverness at Travis.   

That looks like Travis hacking the CRDS urlopen somehow.   CRDS has no proper timeout for these so the URL open either waits indefinitely, defaults to something,  or has been hacked by Travis to add a timeout.

stscieisenhamer:
@drdavella Been suggest to ping you on this: Does the travis failure for this PR have anything to do with the astropy remote_data system?  If so, how?

drdavella:
Yes, this appears to be related to `remote_data`. Earlier this week there were changes to Astropy's test infrastructure, but it should not have changed this behavior. Should I be able to reproduce this test locally by running the unit tests?

drdavella:
Actually, I know what the issue is. I can provide a quick workaround and you can let me know if it works. I'll get started on a longer term fix now.

PR #1445, opened 2017-11-09
Remove astropy.test.helper.pytest

jaytmiller:
Lim said we should drop this,  believe it is general advice not critical to any specific problem or B7.1 release.


pllim:
I think @drdavella said that it doesn't matter... `¯\_(ツ)_/¯`

PR #1446, opened 2017-11-09
TST: Testing against astropy-dev is now allowed to fail

pllim:
This should hopefully fix the failures. xref astropy/ci-helpers#255

c/c @stscieisenhamer @jaytmiller @drdavella

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,13 +25,13 @@ env:
         - CRDS_PATH='/tmp/crds_cache'
         - PYTHON_VERSION=3.6
         - NUMPY_VERSION=stable
-        - ASTROPY_VERSION=development
+        - ASTROPY_VERSION=stable
 
     matrix:
         - SETUP_CMD='install'
         - SETUP_CMD='test'
-        - PYTHON_VERSION=2.7 SETUP_CMD='install' ASTROPY_VERSION=stable
-        - PYTHON_VERSION=2.7 SETUP_CMD='test' ASTROPY_VERSION=stable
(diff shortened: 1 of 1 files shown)


pllim:
A simple pip uninstall did not work. I knew that was too easy!

pllim:
Okay I think the hack is working. The build that matters passed and I cancelled the rest. Do you want me to squash and run full test suite, which only makes sense if you plan to merge this?

stscieisenhamer:
Merge the uninstall? Until the permanent fix is available, this looks good to me.

pllim:
Okie dokie. I'll squash when I can get in front of a computer again.

pllim:
@stscieisenhamer , change of plans... I updated https://github.com/STScI-JWST/jwst/pull/1446#issuecomment-343310804 to explain.

nden:
@pllim But then we have to remember to enable it again as we _do_ test against master. Isn't the fix ready now, as in there's a PR - astropy/pytest-remotedata#20 ?

drdavella:
@nden the fix is ready. I just want to wait for tests to pass and then I can make a new release of `pytest-remotedata`.

This does seem like a good catalyst for the discussion about github workflows and testing against development vs stable releases.

pllim:
You're all correct. This PR can be:

(a) A fallback if Dan got distracted by other work and can't do a release soon.
(b) If you really, really want your tests to pass in the next hour (or so) before the release is made.

pllim:
OK @drdavella fixed `remotedata` upstream, so this should be unnecessary now.

Issue #1449, opened 2017-11-09
Resampled (i2d) product from calwebb_image2 has nearly empty header


hbushouse:
The resampled products for individual exposures coming out of the calwebb_image2 pipeline have nearly empty primary headers. Only the dozen or so keywords containing a record of the drizzle params, and a few others calculated by the resample step (e.g. EXPSTART, EXPEND) are present. I imagine this is due to the fact that the output DrizProductModel is created from scratch and a .update is never done on it to copy the meta data from the input file.

The resample code appears to have the param `blendheaders` hardwired to True, so I guess that means it's being called on the output DrizProductModel, but what does it do when there's only 1 input image? In the case of a single input, either `blendheaders` should be turned off and you just call output.update(input) instead, or modify blendheaders to do that when it's given only 1 input.

PR #1451, opened 2017-11-10
Update meta data in calimage2 resampled product

hbushouse:
Quick fix for #1449, by calling .update on the resampled result model before saving it in the calwebb_image2 pipeline. A real fix should be done within the resample step itself in the future.


hbushouse:
Checks passed the first time. Merging so it gets into overnight build.

PR #1453, opened 2017-11-10
Populate metadata from _cal to _i2d in image2 pipeline

### LINKED ISSUE #1449: Resampled (i2d) product from calwebb_image2 has nearly empty header

jdavies-st:
Resolves #1449

### DIFF
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -71,6 +71,9 @@ class ResampleData(object):
         self.output_wcs = resample_utils.make_output_wcs(self.input_models)
         log.debug('Output mosaic size: {}'.format(self.output_wcs.data_size))
         self.blank_output = datamodels.DrizProductModel(self.output_wcs.data_size)
+
+        # update meta data and wcs
+        self.blank_output.update(input_models[1])
         self.blank_output.meta.wcs = self.output_wcs
 
         self.output_models = datamodels.ModelContainer()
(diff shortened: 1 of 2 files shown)


jwst/resample/resample.py
```diff
@@ -71,6 +71,9 @@ def __init__(self, input_models, output=None, ref_filename=None, **pars):
         self.output_wcs = resample_utils.make_output_wcs(self.input_models)
         log.debug('Output mosaic size: {}'.format(self.output_wcs.data_size))
         self.blank_output = datamodels.DrizProductModel(self.output_wcs.data_size)
+
+        # update meta data and wcs
+        self.blank_output.update(input_models[1])
```
hbushouse: Just curious, why index 1 in input_models, rather than 0?

jdavies-st: Doh!  *brainfart*  That's a bug.  I should be tagging you to review all my PRs.

hbushouse: You actually forgot that python is zero-indexed?! Shame, shame, shame ...

jdavies-st: I know.  I should probably stop working right now.

PR #1463, opened 2017-11-13
Add WCS to median_model

stsci-hack:
This PR corrects the failure seen in the calimage3 regression test where the WCS was not defined for the median image.  
The problem turned out to be caused by relying on update to add the WCS from the input model to the newly created updated datamodel.  However, since the WCS is not defined in the schema for an ImageModel (or any model for that matter), the update operation does not see the WCS attribute and therefore does not copy it over.  This behavior makes perfect sense for model.update() and should simply be documented prominently.

### DIFF
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -203,6 +203,7 @@ class OutlierDetection(object):
         median_model = datamodels.ImageModel(
                                         init=drizzled_models[0].data.shape)
         median_model.update(drizzled_models[0])
+        median_model.meta.wcs = drizzled_models[0].meta.wcs
         base_filename = self.input_models[0].meta.filename
         median_model.meta.filename = '_'.join(
                         base_filename.split('_')[:2] + ['median.fits'])


jdavies-st:
Since this was a bug in 7.1, would it be good to merge this PR into the 0.8.x branch as well?

nden:
@jdavies-st If we have to make a bug fix release then we'll create a patch branch from 0.8.x where this (and other fixes) can be merged. Then that branch will be eventually merged with 0.8.x and stable as a release.
If we are not making a bug-fix release, this should stay on master only until we release master with the next 0.9 release.

jdavies-st:
Ok, cool.  That makes sense.

Is the `bug` tag good enough to identify this as something that should go into that patch if it is eventually needed?  Or will we really have to have a close look at all commits to figure out if they should go in or not?

Hopefully we don't have to do a patch release.  =)

nden:
It may be good to have a specific label, e.g. `7.1bug`.

nden:
Actually, even better is `0.8bug` or simply `0.8.1`, something that references the software version, not the build version.

hbushouse:
Correct. Only if we find out that we have to fix something in the official B7.1 and redeliver to I&T will we merge things into that branch. Otherwise everything else continues to go into the development (master) branch, even if it was a known (to us) issue in B7.1.

PR #1482, opened 2017-11-16
Fix outlier_registry keyword

### LINKED ISSUE #1479: Outlier detection registry naming problem
[Line 121](https://github.com/STScI-JWST/jwst/blob/master/jwst/outlier_detection/outlier_detection_step.py#L121) of outlier_detection_step.py makes reference to `outlier_registry['slit']`, which is inconsistent with the definitions in [Line 17](https://github.com/STScI-JWST/jwst/blob/master/jwst/outlier_detection/outlier_detection_step.py#L17), which uses the name `slitspec`.

jdavies-st:
Resolves #1479

### DIFF
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -44,7 +44,7 @@ class OutlierDetection(object):
 
     """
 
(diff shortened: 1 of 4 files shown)


hbushouse:
👍

PR #1485, opened 2017-11-16
Add build_schema2fits_dict() method for datamodels

jdavies-st:
A new method `build_schema2fits_dict` added.  It does the mirror image of the `build_fits_dict`.  It creates a dictionary that maps flattened metadata attributes to a tuple containing their FITS keyword and FITS HDU location (if any).

### DIFF
--- a/jwst/datamodels/schema.py
+++ b/jwst/datamodels/schema.py
@@ -68,6 +68,37 @@ def build_fits_dict(schema):
     return results
 
 
+def build_schema2fits_dict(schema):
+    """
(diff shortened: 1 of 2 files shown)


jdavies-st:
@sosey The first commit the little dict utility I mentioned for figuring out which metadata values map to which FITS headers (PRIMARY or SCI).

@nden The 2nd commit gives an example of its use in defining the wcsinfo keywords to move in old data from PRIMARY to SCI header, based on what the schemas says their locations should be.  Is this reasonable?  Do we want it to automatically figure this out from the schema?

PR #1486, opened 2017-11-16
Add schema to support tangent plane WCS corrections

mcara:
CC: @nden Please review this PR.

### DIFF
--- a/jwst/transforms/schemas/tpcorr-0.7.0.yaml
+++ b/jwst/transforms/schemas/tpcorr-0.7.0.yaml
@@ -1,0 +1,46 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/yaml-schema/draft-01"
+id: "http://stsci.edu/schemas/jwst_pipeline/tpcorr-0.7.0"
+tag: "tag:stsci.edu:jwst_pipeline/tpcorr-0.7.0"
+title: >
+  Tangent Plane Correction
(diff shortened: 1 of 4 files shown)


jwst/transforms/schemas/tpcorr-0.7.0.yaml
```diff
@@ -0,0 +1,46 @@
...
+          Roll angle.
+        type: number
+      matrix:
+        description: |
+          Generalized rotation matrix.
+        type: array
+        items:
+          type: array
+          items:
+            type: number
```
nden: This limits the schema to a list of lists. If you know that you are never going to pass a numpy array then it's fine. Or yo ucan pass a ndarray but make sure that you convert it to a list of lists. 

Alternatively you can us `anyOf` and add ndarray as an alternative.

In any case add a test to tags/tests which writes the model to a file and then reads it back in. There are examples in the test file.

mcara: @nden Thank you! I could not find any examples of how to do this but I changed the code `TPCorrType` to perform the conversion after/before serialization.

mcara:
@nden I think I got the idea of `anyOf`. Many thanks!

mcara:
@jdavies-st I think now you can try running the entire pipeline - hopefully serialization will work as expected. Thanks!

PR #1489, opened 2017-11-16
Account for recent changes to asdf and astropy

drdavella:
We just moved astropy-specific tag implementations from asdf to the astropy core repository. This required a few minor changes to gwcs and to the JWST pipeline. This PR should account for those changes, assuming that the latest versions of asdf, astropy, and gwcs are pulled in by the travis build.

These changes exposed a minor issue with `JWSTExtension`, which has been addressed in this PR as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,9 @@ env:
     matrix:
         - SETUP_CMD='install'
         - SETUP_CMD='test'
-        - PYTHON_VERSION=2.7 SETUP_CMD='install' ASTROPY_VERSION=stable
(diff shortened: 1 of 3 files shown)


drdavella:
Also important to note: the latest development version of asdf now has a hard requirement on astropy-3.0 (meaning the latest development version for now). The development version of astropy is no longer compatible with Py2, so it was necessary to disable the Py2 travis tests for this PR.

If it's important to continue building against Py27, it will be necessary to build against astropy-2.0.x and asdf-1.3.x.

hbushouse:
It is _**not**_ important or necessary to build/test against py27 anymore.

PR #1491, opened 2017-11-17
define a JWSTTransformType, fix typos, add tests

nden:
Transforms used in the pipeline should inherit now from JWSTTransformType. They don't need to be added to the JWSTExtension any more as this is done automatically.

### DIFF
--- a/jwst/transforms/jwst_types.py
+++ b/jwst/transforms/jwst_types.py
@@ -7,7 +7,6 @@ All types are added automatically to ``_jwst_types`` and the JWSTExtension.
 """
 import six
 
-from asdf.asdftypes import ExtensionTypeMeta
 from astropy.io.misc.asdf.tags.transform.basic import TransformType
 from astropy.io.misc.asdf.types import AstropyTypeMeta
 
@@ -25,7 +24,7 @@ class JWSTTypeMeta(AstropyTypeMeta):
     def __new__(mcls, name, bases, attrs):
         cls = super(JWSTTypeMeta, mcls).__new__(mcls, name, bases, attrs)
         # Classes using this metaclass are automatically added to the list of
-        # astropy extensions
+        # jwst types and JWSTExtensions.types.
         if cls.organization == 'stsci.edu' and cls.standard == 'jwst_pipeline':
             _jwst_types.add(cls)
 
--- a/jwst/transforms/jwst_types.py
+++ b/jwst/transforms/jwst_types.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+
+"""
+Defines a ``JWSTTransformType`` used by jwst pipeine transform types.
+All types are added automatically to ``_jwst_types`` and the JWSTExtension.
+
+"""
+import six
+
+from asdf.asdftypes import ExtensionTypeMeta
+from astropy.io.misc.asdf.tags.transform.basic import TransformType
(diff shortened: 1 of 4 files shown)


jwst/transforms/jwst_types.py
```diff
@@ -0,0 +1,42 @@
...
+from astropy.io.misc.asdf.tags.transform.basic import TransformType
+from astropy.io.misc.asdf.types import AstropyTypeMeta
+
+__all__ = ['JWSTTransformType']
+
+
+_jwst_types = set()
+
+
+class JWSTTypeMeta(AstropyTypeMeta):
```
drdavella: I think this should inherit `ExtensionTypeMeta`, not `AstropyTypeMeta`

jwst/transforms/jwst_types.py
```diff
@@ -0,0 +1,42 @@
...
+"""
+Defines a ``JWSTTransformType`` used by jwst pipeine transform types.
+All types are added automatically to ``_jwst_types`` and the JWSTExtension.
+
+"""
+import six
+
+from asdf.asdftypes import ExtensionTypeMeta
+from astropy.io.misc.asdf.tags.transform.basic import TransformType
+from astropy.io.misc.asdf.types import AstropyTypeMeta
```
drdavella: This import shouldn't be required: see comment below.

jwst/transforms/tags/jwst_models.py
```diff
@@ -4,6 +4,7 @@
 from numpy.testing import assert_array_equal
 from asdf import yamlutil
 from astropy.io.misc.asdf.tags.transform.basic import TransformType
```
drdavella: This import is no longer required.

jwst/transforms/jwst_types.py
```diff
@@ -0,0 +1,42 @@
...
+
+class JWSTTypeMeta(AstropyTypeMeta):
+    """
+    Keeps track of `JWSTType` subclasses that are created so that they can
+    be stored automatically by astropy extensions for ASDF.
+    """
+    def __new__(mcls, name, bases, attrs):
+        cls = super(JWSTTypeMeta, mcls).__new__(mcls, name, bases, attrs)
+        # Classes using this metaclass are automatically added to the list of
+        # astropy extensions
```
drdavella: Should be JWST extensions.

jwst/transforms/jwst_types.py
```diff
@@ -0,0 +1,42 @@
...
+        if cls.organization == 'stsci.edu' and cls.standard == 'jwst_pipeline':
```
drdavella: This `if` statement is probably not necessary in this case. It is useful in the astropy extension because Astropy may theoretically implement tags from multiple standards (e.g. from the ASDF standard and from Astropy itself). There's no harm in leaving it, though.

jwst/transforms/jwst_types.py
nden: I get an error if it inherits from `ExtensionTypeMeta`.
```
E   TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

nden: Basically because I need to inherit from `TransformType`.

drdavella: That's a little bit unfortunate, but yes you're right.

Issue #1063, opened 2017-07-19
default values for WCS keywords


nden:
The core schema defines defaults for some WCS keywords and for some it doesn't. For example, CRVAL3=0.0 while CUNIT3 has no default value. Because of the way models work when writing a file, keywords are populated from the defaults in the schema if they have no assigned other values. So CRVAL3, for example, is written out to image headers which is clearly wrong.

There are two options to mitigate this.
- Remove defaults from all 3-indexed WCS keywords.
  This makes other WCS code more clunky because now it has to check if the value is None and set it to the default (in case there's really a 3rd axis).
- Create a dedicated Spectral2D model, remove the 3-rd axis keywords from the core schema and add them in this Spectral2D model schema with their defaults. This Spectral2D model can be the SlitModel some of us have talked about recently.

TBD ...

PR #1499, opened 2017-11-20
remove defaults

nden:
Crashes are better than silent failing.


nden:
Also #1063 needs to be addressed.

nden:
Just to add a bit more context here. The WCS keywords were moved from the Primary header to the SCI extension header. Keywords which correspond to the FITS WCS standard keywords had default values, specified in the schema, and this ensured they were written to the SCI header. This was handy for a while because initially test data files had no pointing information. Now, that we are testing the entire pipeline, we want to catch errors as early as possible. Some old data files still have their WCS keywords in the primary header. By setting defaults in the wcsinfo schema we are using the default values in the pipeline instead of the intended WCS keywords from the Primary header. While the problem is really just with the data, by removing the default values we are aiming for the pipeline to crash instead of running with wrong data.  

This change will have the following effects:

In some cases, some of the WCS keywords will likely not appear in the SCI header at all, because there are no default values. This means we may start getting pipeline crash reports because of missing keywords.

SDP does not populate the third axis keywords, only the spatial axes. The spectral keywords are populated in assign_wcs and there may be cases when spatial keywords are missing but the spectral ones are there.

jdavies-st:
I would be happy to see this merged.

We're starting to get simulated data from different instrument groups that have various problems with the header keywords, and it would be nice to have `assign_wcs` fail on these files so that it is easier to diagnose the problems with the input data.

PR #1500, opened 2017-11-20
Modify the update method to not copy certain protected keywords

bernie-simon:
Several metadata keywords: date, model_type, and filename are set when a model is opened and saved and should not be overwritten whan one model is updated from another. This code change checks keywords against this list and doesn't update them if they are on it.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -289,6 +289,11 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
         current_date.format = 'isot'
         self.meta.date = current_date.value
 
+        if not hasattr(self.meta, 'model_type'):
+            klass = self.__class__.__name__
+            if klass != 'DataModel':
+                self.meta.model_type = klass
+
     def save(self, path, *args, **kwargs):
         """
         Save to either a FITS or ASDF file, depending on the path.
@@ -627,7 +632,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
         ----------
         d : model or dictionary-like object
             The model to copy the metadata elements from. Can also be a
-            dictionary or dictionary of dictionaries or lists.
+            dictionary or dictionary of dictionaries or lists. 
         only: only update the named hdu from extra_fits, e.g.
             only='PRIMARY'. Can either be a list of hdu names
             or a single string. If left blank, update all the hdus.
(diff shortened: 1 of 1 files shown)


stscieisenhamer:
@bernie-simon Was there a reason that `filename` is one of the protected keywords? This is the reason for the regression test failures involving `make_output_path` and `ModelContainer.save`.

Note: the regression failures will be fixed in an upcoming PR, #1316, for other reasons.

However, the semantic has been to use meta.filename as a basis to construct, or know from where, a model had come from. Was there a discussion that has decided otherwise on this?

bernie-simon:
> On Nov 24, 2017, at 10:15 PM, Jonathan Eisenhamer <notifications@github.com> wrote:
> 
> @bernie-simon Was there a reason that filename is one of the protected keywords? This is the reason for the regression test failures involving make_output_path and ModelContainer.save.
> 
> Note: the regression failures will be fixed in an upcoming PR, #1316, for other reasons.
> 
> However, the semantic has been to use meta.filename as a basis to construct, or know from where, a model had come from. Was there a discussion that has decided otherwise on this?

I set the list of protected keywords to those that are set in model_base.py in __init__() and on_save(). They are meta.filename, meta.date, and meta.model_type. I assumed that it would be a mistake to pull the filename from the file used in update(). Sorry if your code was relying on it.

stscieisenhamer:
Thanks! Question: does this affect when copies of `DataModel`s are being made?

philhodge:
Why shouldn't `meta.filename` be copied?  I use the `update` method when creating a model for an optional output file, and I want the optional file to have the same name as the normal output file but with a different suffix.  For example:

        interpolated_flats = datamodels.MultiSlitModel()
        interpolated_flats.update(output_model, only="PRIMARY")

nden:
I also use `update` assuming the `filename` keyword stays the same. Isn't this the most common use case? What would it be set to if not copied?

bernie-simon:
> On Nov 27, 2017, at 9:35 AM, Nadia Dencheva <notifications@github.com> wrote:
> 
> I also use update assuming the filename keyword stays the same. Isn't this the most common use case? What would it be set to if not copied?

The change in the code only relates to the update method. If a model is copied, the result is a deep copy of the model. If a new model is created using another model as an argument, you get a shallow copy. 

When I made the change, I considered the case:

foo_model = DataModel("foo.fits")
print(foo_model.meta.filename)
# foo.fits
bar_model = DataModel("bar.fits")
print(bar_model.meta.filename)
# bar.fits
foo_model.update(bar_model)
print(foo_model.meta.filename)
# foo.fits, but used to be bar.fits

If a new model is created without an argument or from a shape or array, it will not have a filename until it is saved and under the new code will not get one from the update method.

hbushouse:
I had also assumed that model.update() would copy the filename value, so that a subsequent call to anything like model.save_model() would have something to work with to construct the proper output filename for the updated model. If filename is only populated during model.__init__, what happens when you create a new model from scratch (in memory)? There's no existing file name to give it. Hence the usefullness of having a subsequent call to model.update() populate at least something for the file name, which will then presumably get updated or modified later when the new model is actually saved to a file for the first time.

PR #1517, opened 2017-11-22
Correct? default suffixes assigned to TSO output products

stsci-hack:
The default suffix defined for outlier detection intermediate products appears to have been mis-assigned previously.  This PR resets the default so that 's2d' is used for spectroscopic data and 'i2d' for imaging, as opposed to the exact opposite assignment delivered in B7.1.


hbushouse:
It's correct that if you were to resample imaging data the suffix for the resampled product would be i2d, and s2d (or s3d if it's IFU) for resampled spectroscopic data, but the outlier_detection step should never be resampling the input data (for either imaging or spectroscopic) if the exposures are from one of the TSO modes. Outlier detection on those data is always done by simply forming a median directly from the stack of individual 2-D input images (one per integration), without any resampling. So theoretically it doesn't really matter what you set the intermediate resampled suffix to, because there should never be any produced.

PR #1529, opened 2017-11-24
Update caltso3 to pass along filenames

stsci-hack:
This PR addresses problems with CALTSO3 not passing along the filename to outlier_detection, since apparently filename is an attribute which is NOT updated with the .update() method for datamodels.  

Testing indicates it will resolve the regression test errors showing up as of 24 Nov.


hbushouse:
@stsci-hack Given that all calwebb_tso3 regression tests have been passing for a while now, with outlier_detection activated, was the original problem here fixed via some other change somewhere, like in the model.update() method?  If so, then I guess this PR could be closed.

Issue #1374, opened 2017-10-26
Use wavelength array in preference to wcs function.


philhodge:
The extract_1d and flat_field steps compute the wavelength at each pixel by calling the wcs function.  If the wavelength attribute exists and is populated, that should be used for the wavelengths instead.  Since the default value for the wavelength array is 0, "is populated" means the values are not all 0.

philhodge:
Closed by PRs #1542 and #1546.

PR #1546, opened 2017-11-30
Use wavelength array in preference to wcs function.

philhodge:
For NIRSpec spectroscopic data, the flat_field step computes the wavelength
at each pixel by calling the WCS function (except for BRIGHTOBJ data, which
already get the wavelength from the wavelength attribute).  If the wavelength
attribute exists and is populated (i.e. not all zeros), that should be used
for the wavelengths instead.  See issue #1374.

### DIFF
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -304,10 +304,17 @@ def do_NIRSpec_flat_field(output_model,
         got_wcs = hasattr(slit.meta, "wcs") and slit.meta.wcs is not None
 
         # Get the wavelength at each pixel in the extracted slit data.
-        wl = slit.wavelength                    # a 2-D array
+        # If the wavelength attribute exists and is populated, use it
+        # in preference to the wavelengths returned by the wcs function.
         got_wl_attribute = True
-        # There must be either a wavelength array or a meta.wcs.
-        if wl.min() == 0. and wl.max() == 0.:
+        try:
+            wl = slit.wavelength                # a 2-D array
+        except AttributeError:
+            got_wl_attribute = False
+        # The default value is 0, so all 0 values means that the
+        # wavelength attribute was not populated.  We need either a
+        # wavelength array or a meta.wcs.
+        if not got_wl_attribute or wl.min() == 0. and wl.max() == 0.:
             got_wl_attribute = False
             log.warning("The wavelength array for slit %s has not "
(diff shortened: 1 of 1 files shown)


jwst/flatfield/flat_field.py
```diff
@@ -301,34 +301,47 @@ def do_NIRSpec_flat_field(output_model,
...
-        # Make sure there is a WCS.
-        if not hasattr(slit.meta, "wcs") or slit.meta.wcs is None:
-            log.error("Slit %s does not have a 'wcs' attribute.", slit.name)
-            if output_model.meta.cal_step.assign_wcs == 'COMPLETE':
-                raise RuntimeError("WCS was not found, but it isn't clear "
-                                   "why not.")
+        got_wcs = hasattr(slit.meta, "wcs") and slit.meta.wcs is not None
+
+        # Get the wavelength at each pixel in the extracted slit data.
+        wl = slit.wavelength                    # a 2-D array
```
nden: Would this raise an AtrttibuteError in slitless mode, since there's no wavelength array written out? Should the code check if the `wavelength` attribute is available? Or is the `wavelength` attribute always available because of the schema?

jwst/flatfield/flat_field.py
```diff
@@ -301,34 +301,47 @@ def do_NIRSpec_flat_field(output_model,
...
+        got_wl_attribute = True
+        # There must be either a wavelength array or a meta.wcs.
+        if wl.min() == 0. and wl.max() == 0.:
```
nden: wl.min() and wl.max() will return np.nan for Nirspec data since there are NaNs in the arrays. Is `np.nanmin()` better?

jwst/flatfield/flat_field.py
philhodge: This section would only be executed for `MultiSlit` data, so the attribute should always exist because of the schema, but I'll add a check anyway.  It's better not to rely too much on assumptions.

jwst/flatfield/flat_field.py
philhodge: This should be OK.  If I really wanted the min or max, I would need to exclude NaNs, but all I need to know is that the values are not zero.  If there are NaNs in the array, it has at least been populated with something.

nden: Ah, I see. It makes sense.

Issue #1520, opened 2017-11-22
Add support for aligning "groups" of images in tweakreg


mcara:
As mentioned in https://github.com/STScI-JWST/jwst/pull/1460 support for aligning "groups" of images in tweakreg has been dropped. This support needs to be implemented.

CC: @hbushouse

mcara:
Solved via #1622

Issue #1050, opened 2017-07-17
tweakreg work for B7.1


hbushouse:
Several changes/upgrades are needed for tweakreg.

1) Upgrade the infrastructure to use the JWST-style WCS/gwcs objects

2) @jdavies-st and I think it would be reasonable to combine the tweakreg_catalog and tweakreg steps into a single step. This could be accomplished by simply writing a high-level tweakreg step that calls the existing tweakreg_catalog and tweakreg steps. Can't imagine it ever being necessary or useful in the pipeline environment to run one without the other. This would also involve an update to the calimage3 pipeline module to only call the one new combined step.

3) tweakreg should delete the catalogs produced by tweakreg_catalog when it's all done doing its work. There are no plans to archive the source catalogs for the individual exposures, hence they should be cleaned out of the directories.

larrybradley:
:+1: to combining the steps.

mcara:
Most definitely tweakreg step cannot run without catalogs. However, can anyone ever want to create catalogs without aligning the images? I was under impression that this was the initial thought... If it is not true any longer than definitely it makes sense to combine the two steps.

hbushouse:
Can anyone think of any use cases where someone might want to just create the catalogs for each image? Could the source_catalog step be used for that (on one image at a time) or is it a bit different than tweakreg_catalog?

sosey:
I can imagine you might to use different catalog parameters to create each of those catalogs (source vs tweakreg) since they have a different purpose - i.e. find me all the objects vs find me all the objects which are appropriate for calculating offsets? Are the input parameter sets the same to both tweakreg_catalog and source_catalog?

larrybradley:
They are fundamentally different catalogs.  `tweakreg_catalog` searches for point-like sources (i.e. sources similar to an input kernel) to be used for *alignment*.   `source_catalog` catalogs all sources (point-like and extended) that are above some given threshold (e.g. a S/N threshold above the background).  The source catalog also contains a lot more measurements (incl. morphological) than the tweakreg catalog.

`tweakreg_catalog` is essentially `DAOFind` and `source_catalog` is like `SExtractor`.

jdavies-st:
If a user wants to run the step without actually doing the `tweakreg` part and just make catalogs, then there could be a flag for that.  But it does make sense to combine them from my view.  If one needs to control photutils params for these catalogs, they're available in the step `spec`.

[...]

PR #1556, opened 2017-12-02
Combine tweakreg_catalog with tweakreg

jdavies-st:
This is a stab at combining `tweakreg_catalog` and `tweakreg` into a single step.  This is needed as one cannot perform or skip one without the other.  For historic reasons, they were separate, but this is no longer needed or wanted.  Here we:

- remove the `tweakreg_catalog` step
- merge `tweakreg_catalog` functionality into `tweakreg`
- keep `tweakreg` step turned off by default in the `calwebb_image3` config until #1520 is addressed

This addresses part 2 of #1050


jwst/tweakreg/tweakreg_step.py
```diff
@@ -136,20 +179,9 @@ def _imodel2wcsim(self, image_model):
                         'v2_ref': refang['v2_ref'] / 3600.0,
                         'v3_ref': refang['v3_ref'] / 3600.0},
             catalog=catalog,
-            name=_get_filename(image_model),
+            name=image_model.meta.filename,
```
mcara: The reason for `_get_filename` was to reduce the length of the file name so that log messages are not excessively long. I am fine with keeping full path if that is what you really want.

jdavies-st: Thanks Mihai.  The reason I removed this is that I think the filename complete path will eventually be handled by stpipe, not individual steps.  Only the filename (not the absolute directory) will get written to this keyword, so we shouldn't have to do any special handling at the step level.  Is this a correct assumption @stscieisenhamer?

PR #1568, opened 2017-12-05
Revert "remove defaults"

nden:
Reverts STScI-JWST/jwst#1499
It broke master.


nden:
passed in my travis account

PR #1573, opened 2017-12-07
Fix default value for sovler parameter to match_lsq

stscieisenhamer:
The default given for parameter `solver` was `LU`.
However, according to comments and code, only `RLU` and `PINV`
are supported.

Though it seems to be just a typo and the intent was for `RLU`
to be default, I went with `PINV` because then the code would
actually run for MRS test data.

### DIFF
--- a/jwst/wiimatch/match.py
+++ b/jwst/wiimatch/match.py
@@ -26,7 +26,7 @@ SUPPORTED_SOLVERS = ['RLU', 'PINV']
 
 def match_lsq(images, masks=None, sigmas=None, degree=0,
               center=None, image2world=None, center_cs='image',
-              ext_return=False, solver='LU'):
+              ext_return=False, solver='PINV'):
     """
     Compute coefficients of (multivariate) polynomials that once subtracted
     from input images would provide image intensity matching in the least
--- a/jwst/wiimatch/match.py
+++ b/jwst/wiimatch/match.py
@@ -26,7 +26,7 @@ SUPPORTED_SOLVERS = ['RLU', 'PINV']
 
 def match_lsq(images, masks=None, sigmas=None, degree=0,
               center=None, image2world=None, center_cs='image',
-              ext_return=False, solver='PINV'):
+              ext_return=False, solver='RLU'):
     """
     Compute coefficients of (multivariate) polynomials that once subtracted
     from input images would provide image intensity matching in the least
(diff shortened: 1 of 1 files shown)


mcara:
What kind of problems do you have when running MRS data using `'RLU'`? Testing by Katherine Murray shows that `'PINV'` (previously the only method) can produce really bad results. My investigation showed that under some circumstances the method is unstable. `'RLU'` seems to be more stable. Even though `'PINV'` may "run for MRS data" - the results may not be as good as `'RLU'` which is intended to replace `'PINV'` for all practical purposes.

mcara:
I did test the code and RLU ('LU' at that time) worked just fine using 2D and 3D data (in fact producing identical data to PINV when that one was stable and the results expected in simulated data). So I would be surprised if the code does not work on MRS data.

The mistake that I did was to forget renaming LU->RLU in the default parameter to `match_lsq()` in the last moment (I thought it was important to add 'R' to indicate that a reduced, smaller, system is being solved).

stscieisenhamer:
I had presumed the error was based on old data, but looking now in more detail, it is a code error. The error is:
```
       # solve the system:
        tol = np.finfo(images[0].dtype).eps**(2.0/3.0)
        if solver == 'RLU':
>           bkg_poly_coef = rlu_solve(a, b, nimages, tol)
E           TypeError: rlu_solve() takes 3 positional arguments but 4 were given

/Users/eisenham/Documents/ssbdev/jwst/jwst/wiimatch/match.py:272: TypeError
```
And looking at the signature for `rlu_solve`:
```
def rlu_solve(matrix, free_term, nimages):
    """
    Computes solution of a "reduced" system of linear equations
```
If the fix is easy, let me know and I'll put it in this PR. If not, I'll just fix the default value and merge and let you deal with the above.

mcara:
Yep, it's easy: just remove `tol` argument from the `rlu_solver`. Thanks!

mcara:
... and while you are at it - move `tol = ...` line to the `else` branch as the `'RLU'` branch does not need it.

stscieisenhamer:
Now I'm getting
```
        getrf, = get_lapack_funcs(('getrf',), (a1,))
>       lu, piv, info = getrf(a1, overwrite_a=overwrite_a)
E       ValueError: failed to create intent(cache|hide)|optional array-- must have defined dimensions but got (0,)

/Users/eisenham/anaconda3/envs/pipesaving/lib/python3.6/site-packages/scipy/linalg/decomp_lu.py:65: ValueError
```

Shall I continue merge and you can deal with this separately?

mcara:
Yes, merge it. Do you have these data somewhere? I was not getting this error with my own simulated data and so think I need the data to reproduce this error.

mcara:
Make a commit with `'RLU'` and `tol` stuff and I will approve the changes. Then merge.

stscieisenhamer:
Will do and I'll send you the data. This could be an "old data" issue.

stscieisenhamer:
@mcara ok, changes in. re-review and i'll merge

mcara:
Looks good!

mcara:
@stscieisenhamer The error reported in https://github.com/STScI-JWST/jwst/pull/1573#issuecomment-349994607 is because you are trying to "match" the background of a single image (to what???) ...

https://github.com/STScI-JWST/jwst/pull/1581 should fix this.

PR #1575, opened 2017-12-08
Replace nightly doc build with RTD

### LINKED ISSUE #198: Move doc build to RTD?
Why not?

pllim:
Fix #198

c/c @rendinam @jhunkeler

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 JWST Calibration Pipeline
 =========================
+[![Documentation Status](https://readthedocs.org/projects/jwst-pipeline/badge/?version=latest)](http://jwst-pipeline.readthedocs.io/en/latest/?badge=latest)
 [![Build Status](https://travis-ci.org/STScI-JWST/jwst.svg?branch=master)](https://travis-ci.org/STScI-JWST/jwst)
 [![Powered by STScI Badge](https://img.shields.io/badge/powered%20by-STScI-blue.svg?colorA=707170&colorB=3e8ddd&style=flat)](http://www.stsci.edu)
 [![Powered by Astropy Badge](http://img.shields.io/badge/powered%20by-AstroPy-orange.svg?style=flat)](http://www.astropy.org/)
@@ -44,6 +45,6 @@ We strive to provide a welcoming community to all of our users by abiding to the
 Using
 -----
 
-Documentation (built nightly) is available here:
+Documentation (latest off `master`) is available here:
 
-http://jwst-pipeline.readthedocs.io/en/latest/
+https://jwst-pipeline.readthedocs.io/en/latest/


README.md
```diff
@@ -26,6 +26,4 @@ For developers:
 Using
 -----
 
-Documentation (built nightly) is available here:
-
-http://ssb.stsci.edu/doc/jwst_dev/
+[![Documentation Status](https://readthedocs.org/projects/jwst-pipeline/badge/?version=latest)](http://jwst-pipeline.readthedocs.io/en/latest/?badge=latest)
```
hbushouse: There are a lot of people at STScI and around the world who've been told for the past couple years to find our docs at http://ssb.stsci.edu/doc/jwst_dev/, so this change is going to cause a lot of bookmarks and links to break, and also cause a flood of sudden help calls. Is there some way to keep both URL's active for a while to allow for a transition period? I don't want jwst_dev to suddenly just be gone or obsolete suddenly as of today. A change like this really needs to be coordinated before implementing.

pllim: Of course, that is understandable. It is easy to have http://ssb.stsci.edu/doc/jwst_dev/ re-direct to RTD URL as we have full control of `ssb.stsci.edu`. Alternately, we can provide both links but then we have to explain how one is nightly (which might or might not be slightly outdated compared to `master`) and one is on-the-fly (give or take 30 mins to build, if no error).

hbushouse:
Was this change to README.md incorporated into another PR, because I see the result of the change on the front page, but this PR hasn't been merged yet? So can this PR be closed?

@pllim @jhunkeler @rendinam

pllim:
@hbushouse , the README change was unrelated to this. This one replaces the link *at the bottom*. I'll rebase shortly.

pllim:
Rebased.

jhunkeler:
Hey @pllim shouldn't we put this badge at the top alongside the others?

nden:
Was the removal of the text deliberate? I kind of like to have the link visible. It would not occur to me to click on the badge in order to get to the docs. Also I think it's useful to clarify at the moment that this is the nightly build of the docs.

pllim:
@jhunkeler -- Ideally, yes. However, if people are used to looking for it at the bottom of README already, they would be confused that it is not there anymore. If there are no such people, I would be glad to move it up to the top.

@nden -- Strictly speaking, RTD builds on every commit to `master` now, so it is not "nightly" anymore. I could write out the link in addition to the badge, but I wouldn't say it is "nightly".

pllim:
Also, if you really want the doc to be easily discoverable on this repo, you could add http://jwst-pipeline.readthedocs.io/en/latest/ at the top next to the repo description. I could have done it for you but looks like *someone* removed my edit access. 😉

nden:
@pllim You are right. I still think it's useful to clarify that the docs reflect the code in github and not what is currently in testing. After all this code is "not public" yet. I also think it's best to have the link visible instead of hidden in the badge. Then the status badge can go up where the other badges are.

pllim:
@nden , I believe I have addressed the comments.

PR #1594, opened 2017-12-14
Update calwebb_tso3.cfg to match pipeline

hbushouse:
The calwebb_tso3 pipeline was updated in the final B7.1 delivery to no longer use separate outlier_detection_scaled and outlier_detection_stacked steps, but the default .cfg file was not updated to reflect this change. The calwebb_tso3 pipeline then fails, when executed using the default .cfg file, because the step outlier_detection_scaled is no longer defined in the pipeline.

This update brings the default calwebb_tso3.cfg file in synch with the latest calwebb_tso3.py pipeline module.

This error wasn't discovered via the calwebb_tso3 regression tests, because the tests were still using an old version of the calwebb_tso3.cfg file that didn't have this problem. Mike Swam discovered the problem during DMS testing of the B7.1 delivery.

### DIFF
--- a/jwst/pipeline/calwebb_tso3.cfg
+++ b/jwst/pipeline/calwebb_tso3.cfg
@@ -5,9 +5,7 @@ scale_detection = False
 
     [steps]
       [[outlier_detection]]
-        config_file = outlier_detection_stack.cfg
-      [[outlier_detection_scaled]]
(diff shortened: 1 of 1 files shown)


hbushouse:
I've also updated the tso3 regression tests on jwcalibdev to use the proper calwebb_tso3.cfg files.

Issue #1582, opened 2017-12-11
Ramp Fitting with only two groups in semi-ramp


mwregan2:
While testing jump detection, I found that in a 100 group integration all the CRs found in the 3rd and 98th group yielded widely varying slopes. These CRs in these two groups cause the formation of a 2-group semi-ramp. Given that the slope and uncertainty have to be calculated differently for a 2-group semi-ramp, it's clear that something is not working correctly in this case.

dmggh:
Hi Mike,


There have been semi-recent updates to that part of the ramp fitting code. Please copy your jump step output to somewhere accessible and I will take a look.


Thanks,

Dave Grumm

________________________________

mwregan2:
Dave,
    I’ve attached the code. I added some print statements to help me understand what it is doing. I just pulled the latest version from Github on Thursday.
  Karl and I have been talking about what we think is the correct way to weight the semi-ramps. We should all probably talk about it.
Mike

dmggh:
Mike,


Hmm, I don't see an attachment.  Anyway, if you can copy your ramp fit input (jump output) to central storage I will take a look at it.


-Dave

________________________________

mwregan2:
Dave,
  I put it in /grp/jwst/wit/odl/jump_output.fits
Mike
Ps There must be something with the github reply system that prevents attachements.

dmggh:
Mike,


Ok, thanks. Are there particular pixels that you have noticed this problem ?


-Dave

________________________________

mwregan2:
Dave,
  Look at any pixel with a cr flagged in group 3 (index =2). They all get slopes that are either very high or very low. The inherent slope is zero but there is read noise. So, the slopes are not exactly zero.
Mike

mwregan2:
Dave,
  The problem is that the two-group semi-ramp has the wrong weight applied. So, when it is combined with the longer data is dominates. The two group semi-ramp is using the MIN_ERR value which 1e-10. The standard deviation of the error is sqrt(2)* read_noise. The Variance is this squared. So, starting at line 1714 inside of fit_double_read change 
From
variance_s[ pixel_ff ] = MIN_ERR
sig_slope_s[ pixel_ff ] = MIN_ERR
sig_intercept_s[ pixel_ff ] = MIN_ERR

to
variance_s[ pixel_ff ] =  2.0 * rn_sect.flatten()[pixel_ff] * rn_sect.flatten()[pixel_ff]
[...]

PR #1596, opened 2017-12-14
Corrected bug which gave artificially high weight to semi-ramps have …

dmggh:
…exactly 2 good reads.


hbushouse:
Fixes #1582

hbushouse:
Fixes #1655.

PR #1600, opened 2017-12-15
Modify datamodels test to not check for ASDF tree in ImageHDU

jdavies-st:
The storage of the ASDF tree in either an `ImageHDU` or another type of HDU is an ASDF implementation detail, and should not be tested here.  This allows for the proposed change in [ASDF PR 412](https://github.com/spacetelescope/asdf/pull/412) to put the ASDF tree in a `NonstandardExtHDU` instead of an `ImageHDU`.


jdavies-st:
One further change in this PR is adding the `.eggs` directory in the repository to be ignored when running test.  There's a corner case if you first run `pythons setup.py test` without having the dependencies installed, it will install them in the `.eggs/` subdir.  Then if you later try to run pytest in the top level directory, it will try to run the tests inside the `.eggs/` subdir, i.e. the test of the package dependencies.

PR #1606, opened 2017-12-19
Tweaks to cube building documentation

drlaw1558:
Jane- I've made some suggested modifications here regarding the overall cube build process documentation.  I haven't really touched the details of the cubepar reference files though since these are slated to be updated soon anyway.


jemorrison:
David. Thanks for going through the documentation and updating it.

PR #1607, opened 2017-12-19
Adding populate_meta to WCS RegionsModel

drlaw1558:
The MIRI MRS code to create CRDS reference files is crashing using the latest version of the pipeline.  On further investigation, this seems to be because there is no 'populate_meta' function for the RegionsModel MIRI-only reference file within the datamodel.  Since this function exists for all other reference file types I've added it for the Regions files as well, which seems to fix the problem so that MRS reference files can be created again.


nden:
Thanks @drlaw1558 !

PR #1610, opened 2017-12-20
Fix msaflagopen for missing v2v3 ref file

hbushouse:
Quick fix to get regression tests running, while @stscirij works on a permanent fix. This handles the fact that the v2v3 reference file was recently retired from use, but the step was still trying to find a name for it, which is now None.

### DIFF
--- a/jwst/msaflagopen/msaflagopen_step.py
+++ b/jwst/msaflagopen/msaflagopen_step.py
@@ -52,7 +52,8 @@ def create_reference_filename_dictionary(input_model):
     reffiles['specwcs'] = input_model.meta.ref_file.filteroffset.name
     reffiles['regions'] = input_model.meta.ref_file.regions.name
     reffiles['wavelengthrange'] = input_model.meta.ref_file.wavelengthrange.name
-    reffiles['v2v3'] = input_model.meta.ref_file.v2v3.name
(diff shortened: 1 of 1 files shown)


hbushouse:
Merging so it gets in this evening's build and reg test cycle.

jwst/msaflagopen/msaflagopen_step.py
```diff
@@ -52,7 +52,8 @@ def create_reference_filename_dictionary(input_model):
     reffiles['specwcs'] = input_model.meta.ref_file.filteroffset.name
     reffiles['regions'] = input_model.meta.ref_file.regions.name
     reffiles['wavelengthrange'] = input_model.meta.ref_file.wavelengthrange.name
-    reffiles['v2v3'] = input_model.meta.ref_file.v2v3.name
+    if input_model.meta.ref_file.v2v3.name is not None:
+        reffiles['v2v3'] = input_model.meta.ref_file.v2v3.name
```
philhodge: Does there need to be a default value for `reffiles['v2v3']`?

stscirij:
I have a fix for this, I just can't get a working version of the jwst package to try it out :(

hbushouse:
Doesn't appear to need a default value, because it works now without having it defined at all (I ran 3 reg tests with it).

PR #1622, opened 2017-12-21
Add support for tweaking *groups* of images that "move" together

mcara:
This PR adds support for "tweaking" the WCS of groups of images: sets of images that must be adjusted together as one large image. This PR addresses https://github.com/STScI-JWST/jwst/issues/1520 and also adds a couple of bug fixes.

CC @hbushouse


mcara:
@hbushouse If something needs to be done to turn on/enable tweakreg in the pipeline - please do that.

Issue #1392, opened 2017-10-30
Remove all references to python 2


nden:
This will be done possibly in a sprint. The ticket here is to collect comments and some obvious things we need to look for. 

- Remove all `from __future__ import ...`
- Remove six
- Remove any direct references to python versions
- Classes don't need to inherit from `object` any more
- Remove `unicode_literals`, if there are any

subpackages
- [x] ami
- [x] assign_wcs
- [x] associations
- [x] background
- [x] barshadow
- [x] combine_1d
- [x] coron
- [x] csv_tools
- [x] cube_build
- [x] cube_skymatch
- [x] dark_current
- [x] datamodels
- [x] dq_init
- [x] emission
- [x] engdblog
- [x] exp_to_source
- [x] extract_1d
- [x] extract_2d
- [x] firstframe
- [x] fits_generator
- [x] flatfield
- [x] fringe
- [x] gain_scale
- [x] group_scale
- [x] guider_cds
- [x] imprint
- [x] ipc
- [x] jump
- [x] jwpsf
- [x] lastframe
- [x] lib
- [x] linearity
- [x] model_blender
- [x] mrs_imatch
- [x] msaflagopen
- [x] outlier_detection
- [x] pathloss
- [x] persistence
- [x] photom
- [x] pipeline
- [x] ramp_fitting
- [x] refpix
- [x] resample
- [x] reset
- [x] rscd
- [x] saturation
- [x] skymatch
- [x] source_catalog
- [x] srctype
- [x] stpipe
- [x] straylight
- [x] superbias
- [x] tests
- [x] timeconversion
- [x] transforms
- [x] tso_photometry
- [x] tweakreg
- [x] tweakreg_catalog
- [x] wfs_combine
- [x] white_light
- [x] wiimatch

nden:
@STScI-JWST/maintainers Here's a list of all subdirectories. When you update a step please check it out so we all have an idea how this is progressing.
(Although I think it's probably best to do it approximately at the same time.)

nden:
@STScI-JWST/maintainers Please don't forget to update the table in this issue. To facilitate finding it, there's a link now to it under `projects`.

hbushouse:
@nden There are still a couple of references to "six" in ``transforms/jwst_types.py``.

nden:
@hbushouse Yeah, let's keep these for now. I'll do something about them in a separate PR.

stscirij:
[...]

PR #1632, opened 2017-12-27
Remove py2 dependencies and add step completion status

hbushouse:
Updated the mrs_imatch step to set the step completion status in the input images, add a log message when applying the background subtraction, and remove python 2 dependencies.


hbushouse:
Addresses #1392

PR #1634, opened 2017-12-28
guider_cds updates for BUNIT and python2 dependencies

hbushouse:
Removed python2 dependencies and added a line to update the value of BUNIT in the output product, to reflect the fact that the output data are in count rates (DN/s).


hbushouse:
Partially addresses #1392.

hbushouse:
Also addresses #63.

PR #1639, opened 2018-01-02
fix nirspec slit validation code

### LINKED ISSUE #1638: slit validation code does not use correct sporder and wavelength range
The NIRSPEC WCS pipeline uses `nirspec.get_open_slits()`  to determine which slits are open AND are projected on the current detector. For this it runs part of the model - from slit to detector using the wavelength range and spectral order for the current instrument configuration. The spectral order and the wavelengh range are determined from the `wavelengthrange` reference file based on the grating, filter and lamp state. If the configuration is missing then the full NIRSPEC wavelength range is used.

When the pipeline runs the spectral order and wavelength range used in calculating the WCS are stored in `model.meta.wcsinfo.sporder`, `model.meta.wcsinfo.waverange_start` and `model.meta.wcsinfo.waverange_end` and the stored values are used elsewhere in the code instead of recomputing them every time.  However, this was not done when the vaidation code was run and this resulted in rejecting some valid open slits because an incorrect spectral order was passed to the grating equation. The example reported was grating: G395M, lamp: ARGON, filter: OPAQUE. 
As a note this only happened in the validation code, the WCS pipeline is working correctly in this respect.

nden:
Close #1638

### DIFF
--- a/jwst/assign_wcs/nirspec.py
+++ b/jwst/assign_wcs/nirspec.py
@@ -223,6 +223,7 @@ def slitlets_wcs(input_model, reference_files, open_slits_id):
     sporder, wrange = get_spectral_order_wrange(input_model, reference_files['wavelengthrange'])
     input_model.meta.wcsinfo.waverange_start = wrange[0]
     input_model.meta.wcsinfo.waverange_end = wrange[1]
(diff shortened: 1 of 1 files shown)


nden:
Unrelated test failure, merging.

PR #1646, opened 2018-01-05
rename nirspec subarrays

### LINKED ISSUE #1613: Task: Update NIRSpec fixed-slit subarray names to match PPS DB
The cal pipeline data models and CRDS reference files currently assume the following subarray values associated with NIRSpec fixed-slit readouts:

ALLSLITS, S200A1, S200A2, S200B1, S400A1, and SUB2048 (the latter is for the S1600A1 slit)

while PPS has a constraint to use the names:

ALLSLITS, SUBS200A1, SUBS200A2, SUBS200B1, SUBS400A1, and SUB2048 (edited)

(see http://www-int.stsci.edu/dsd/cns/database/r2d2/PPSDB79/nirspec_exposure_specification.html)

The core schema needs to be updated to allow for the SUBxxxxx versions of the subarray names, and check for any uses of these names in the actual cal code. This discrepancy was discovered during the last stages of B7.1 subsystem testing, hence the fix was left to B7.1.x.

nden:
resolves #1613


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -611,10 +611,10 @@ properties:
               # Obsolete NIRISS
               SUBSTRIP80,
               # NIRSpec
-              ALLSLITS, S200A1, S200A2, S200B1, S400A1, SUB1024A, SUB1024B,
+              ALLSLITS, SUB200A1, SUB200A2, SUB200B1, SUB400A1, SUB1024A, SUB1024B,
               SUB512, SUB32, SUB512S, SUB2048,
               # Obsolete NIRSpec
-              FULL_IRS2,
+              FULL_IRS2, S200A1, S200A2, S200B1, S400A1,
```
hbushouse: I see that you're leaving the old values in place, at least for now. That's good for backwards compatibility for old test files.

hbushouse:
#1613 mentions that updates will also be needed in extract_2d. Is that still the case?

nden:
@hbushouse Re extract_2d - this still needs to be decided. Currently assign_wcs sand extract_2d use the old slit names for `Slit.name`: "S200A1", .... I've always read this as "Slit200A1" not as "Subarray200A1". I kept these names in assign_wcs for now though I'm not opposed to switching to "SUB200A1", .... It's mostly an interface issue, what we think is more intuitive to users. Or may be how the general NIRSPEC documentation refers to these slits.

The only place this matters in the code is in extract_2d when a subarray name is passed to the "which_subarray" input argument. Currently `which_subarray` assumes the slit names, not the subarray names. If this is how we want it to work then we don't need to change the code but perhaps the argument should be renamed to `which_slit` or just `slit_name`. 
The alternative would be to use the subarray names everywhere: "SUBA2001", ...
Any thoughts on this?

nden:
@stscirij Can you confirm the changes to fits_generator are OK?

philhodge:
Will the `MultiSlitModel` be renamed to `MultiSubarrayModel`?  Will `slit` in that model be renamed?  And the corresponding FITS keywords?

hbushouse:
I don't see the need to rename `MultiSlitModel`. It will still be used to contain multiple instances of the data from individual slits, as opposed to data from subarrays. There can be multiple slit instances within a given subarray (e.g. the ALLSLITS subarray).

jwst/assign_wcs/nirspec.py
```diff
@@ -304,15 +304,15 @@ def get_open_fixed_slits(input_model):
...
+    elif subarray == "SUB200A2":
         slits.append(s2a2)
-    elif subarray == "S400A1":
+    elif subarray == "SUB400A1":
         slits.append(s4a1)
-    elif subarray == "S1600A1":
+    elif subarray == "SUB2048":
         slits.append(s16a1)
-    elif subarray == "S200B1":
+    elif subarray == "SUB200B1":
```
hbushouse: Need to change all of these to include another `S` following the substring `SUB`, i.e. `SUB**S**200A1`, `SUB**S**200A2`, etc., except for `SUB2048` (that stays as is).

hbushouse:
Here's the latest info on what SDP will be doing with the population of SUBARRAY and FXD_SLIT keywords.

The SUBARRAY values will be changing to those listed above, i.e. FULL, ALLSLITS, SUB512, SUB1024A, SUB1024B, SUB2048, SUBS200A1, SUBS200A2, SUBS200B1, and SUBS400A1, when the NIRSpec Fixed Slit Spectroscopy observing template is in use. The correspondence between subarray values and slits is noted below.

The FXD_SLIT keyword values will be: S200A1, S200A2, S200B1, S400A1, and S1600A1.

In the APT, when a user selects one of those FXD_SLIT values, the SUBARRAY value defaults to:

S200A1 -> SUBS200A1
S200A2 -> SUBS200A2
S200B1 -> SUBS200B1
S400A1 -> SUBS400A1
S1600A1 -> SUB2048

However, the user is also given the option of selecting either the FULL or ALLSLITS subarray along with each of those slits. In addition, when the S1600A1 slit is selected, the user can also choose the SUB512, SUB1024A, or SUB1024B subarrays (as well as the aforementioned SUB2048, FULL, and ALLSLITS).

Hence even though a single FXD_SLIT value will be present, the exposure can contain data from multiple slits (e.g. when the user selects the ALLSLITS subarray). The NIRSpec team wants the pipeline to always extract and process the data from all available slits, even when only 1 slit has been selected as the primary. Even if no sources are present in the other slits, the data could be used as background.

Therefore we should continue to have the pipeline processing key off of the SUBARRAY value, rather than the FXD_SLIT value, when determining which slits to process within a given exposure. For example, if FXD_SLIT='S200B1', but SUBARRAY='ALLSLITS', we should attempt to process all 5 slits (or at least have assign_wcs check to see if they fall within the image boundaries). For each slit that is present within the subarray, the SLIT name assigned to the extracted region should correspond to one of the 5 "official" values of S200A1, S200A2, S200B1, S400A1, and S1600A1. Don't use the SUBARRAY value to populate the slit name.

nden:
@hbushouse I think I addressed all your comments. Also I renamed the argument to exatrct_2d: `which_subarray` --> `slit_name` . Let me know if you disagree with the change.

docs/jwst/extract_2d/main.rst
```diff
@@ -32,9 +32,9 @@ NIRCAM GRISM and NIRISS WFSS observations.
 
 For NIRSPEC:
 
-If the step parameter ``which_subarray`` is left unspecified, the default behavior is
+If the step parameter ``slit_name`` is left unspecified, the default behavior is
 to extract all slits which fall within a detector. Only one slit may be extracted by
```
hbushouse: Maybe instead of saying "all slits which fall within a detector" it would be more correct - for all situations involving full-frame and subarray exposures - to say something like "all slits that fall within the limits of the image." And how much of the slit/spectrum trace has to fall within the image to be included in the list of slits to process? Does the entire trace (i.e. covering all wavelengths) have to be inside the image, or is there some minimum portion that must fall inside the image, or what? I'm thinking in particular about the SUB512 and SUB1024 subarrays that can be used with the S1600A1 slit. Those subarrays will likely truncate the trace (at one end or the other, if not both), while leaving some part of it available in the image.

docs/jwst/extract_2d/main.rst
```diff
@@ -32,9 +32,9 @@ NIRCAM GRISM and NIRISS WFSS observations.
...
-specifying the slit name with the ``which_subarray`` argument, using one of the following
+specifying the slit name with the ``slit_name`` argument, using one of the following
 accepted names: ``S1600A1``, ``S200A1``, ``S200A2``, ``S200B1``, ``S400A1`` or ``S200B1``
```
hbushouse: S200B1 appears twice in that list. There should only be 5.

docs/jwst/extract_2d/main.rst
```diff
@@ -58,19 +58,19 @@ corresponding to the FITS keywords ``SLTNAME``, ``SLTSTRT1``, ``SLTSIZE1``,
...
-``jwst.transforms.models``. 
+``jwst.transforms.models``.
 
 
 Step Arguments
 ==============
-The extract_2d step has two optional arguments for NIRSPEC observations: 
+The extract_2d step has two optional arguments for NIRSPEC observations:
 
 * ``--which_subarray``: name (string value) of a specific slit region to
```
hbushouse: Need to change this one too, from which_subarray to slit_name.

docs/jwst/extract_2d/main.rst
nden: If any part of the trace projects on the detector it will be extracted as a separate slit.

But it's not sufficient for a slit to be in the image to be extracted. It has to be projected within the detector limits as well. Part of the model is run on every open slit to determine if it projects on the detector or not. This is especially true for the high-resolution gratings and NRS2. Or am I  mixing terminology?

docs/jwst/extract_2d/main.rst
nden: done, thanks.

nden:
@hbushouse  I think this is ready to merge. Any other comments?

jwst/assign_wcs/nirspec.py
```diff
@@ -304,15 +304,15 @@ def get_open_fixed_slits(input_model):
...
+    if subarray == "SUBS200A1":
         slits.append(s2a1)
-    elif subarray == "S200A2":
+    elif subarray == "SUBS200A2":
         slits.append(s2a2)
-    elif subarray == "S400A1":
+    elif subarray == "SUBS400A1":
         slits.append(s4a1)
-    elif subarray == "S1600A1":
+    elif subarray == "SUB2048":
```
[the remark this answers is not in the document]
hbushouse: Yes. The SUB512, SUB1024A, SUB1024B, and SUB2048 subarrays only cover the portion of the detector occupied by the S1600A1 trace, so whenever those subarrays are used we should assume that the only slit of interest is the S1600A1. According to the instrument docs, the SUB512, SUB1024A, and SUB1024B subarrays will typically be used only for NRS_BRIGHTOBJ exposures with the S1600A1, but those subarrays are available in the APT when using the regular Fixed Slit Spectroscopy observing template, so you know that someone will use them for regular NRS_FIXEDSLIT mode too.

Docs are at https://jwst-docs.stsci.edu/display/JTI/NIRSpec+Fixed+Slits+Spectroscopy and
https://jwst-docs.stsci.edu/display/JTI/NIRSpec+Bright+Object+Time-Series+Spectroscopy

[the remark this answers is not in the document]
nden: Done. In addition I added SUB512S as it was also listed as possible subarray in BOTS mode.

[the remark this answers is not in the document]
hbushouse: 👍

PR #1648, opened 2018-01-09
Update tweakreg documentation and minor API changes

mcara:
Update documentation for the ``tweakreg`` step.  This addresses PR https://github.com/STScI-JWST/jwst/issues/1523


docs/jwst/tweakreg/index.rst
```diff
@@ -1,13 +1,25 @@
...
+**Also See:**
+
+.. toctree::
+   :maxdepth: 2
+
+   imalign
+   wcsimage
+   linearfit
+   matchutils
+   tweakreg_catalog
```
hbushouse: I don't think tweakreg_catalog and tweakreg_step exist anymore. They've been combined into the single tweakreg, which is what these docs are about. Or do they still exist as lower-level functions/modules?

jwst/tweakreg/__init__.py
```diff
@@ -7,8 +7,8 @@
 __docformat__ = 'restructuredtext'
 
 __taskname__ = 'tweakreg'
-__version__ = '0.8.1'
-__vdate__ = '14-November-2017'
+__version__ = '0.8.2'
```
hbushouse: We should bump this up to 0.9.1, to keep in synch with the current development work that will be released as 0.9.x

jwst/tweakreg/matchutils.py
```diff
@@ -1,3 +1,10 @@
+"""
+A module that provides algorithms for initial estimation of shifts
+based on 2D histograms.
+
+:License: `<http://www.stsci.edu/resources/software_hardware/pyraf/LICENSE>`_
```
hbushouse: Is this the appropriate license reference to use? This isn't part of pyraf.

docs/jwst/tweakreg/index.rst
mcara: @hbushouse I see both files in https://github.com/STScI-JWST/jwst/tree/master/jwst/tweakreg The new `tweakreg` step essentially combines the functionality of the two steps (old `tweakreg` and `tweakreg_catalog` steps). Probably `tweakreg_catalog.py` holds what was not moved to the common `tweakreg_step.py`. Maybe `tweakreg_catalog.py` should be renamed to simply `catalog.py`

jwst/tweakreg/__init__.py
mcara: OK

jwst/tweakreg/matchutils.py
mcara: @hbushouse Oh, this is used almost everywhere in my contributed code - I carried it over from HST module "model". I will see how I can make this to point to the `LICENSE` file in the root directory.

mcara:
@hbushouse Thanks!

PR #1660, opened 2018-01-12
Add docs for guider pipeline

hbushouse:
Added a description of the calwebb_guider pipeline to the pipeline doc, and added docs for the guider_cds step.


dmggh:
These docs look good to me.

Issue #1661, opened 2018-01-12
Limit reference prefetch to Pipeline level,  remove CRDS header caching


jaytmiller:
Data models header caching was added to the crds_client module to optimize data model header reads for more complex pipeline cases with heavy file loads,  avoiding multiple dataset reads by caching parameters keyed using the model filename.

Nadia directly encountered the limitation of that approach,  the header for a particular filename is only read once so it cannot change easily during interactive sessions,  particularly for people unaware of the optimization.  Nadia debugged it,  most may not.   So I now believe CRDS header caching is too dangerous,  it will give incorrect results to some users doing (interactive) work.

Jonathan had a better design shift for optimizing the prefetch.  Noticing that the calls to get_reference_file() always immediately follow the prefetch within individual Step functions,  he suggested moving the Step version of the prefetch entirely to the Pipeline level where it is most applicable and also most efficient. 

A minor(?) downside with my implementation is that Steps no longer prefetch,  only the Pipeline.  Because Steps no longer prefetch,  single Step runs will not detect all reference file errors at the outset,  only type-by-type as the Step runs and calls get_reference_file().    Pipelines will still have the property that all reference availability errors and caching for the pipeline are detected before Steps start running.

hbushouse:
Fixed by #1662

PR #1662, opened 2018-01-12
Limit reference prefetch to Pipeline level, remove CRDS header caching

jaytmiller:
1. Refactor Pipeline and Step to move prefetch to Pipeline only

Moved reference file precaching from Step to Pipeline since precaching in Step
is of limited use and adds an extra CRDS operation and potential datamodel read
for each Step.  This leaves the Pipeline level prefetch and individual calls to
get_reference_file() within each Step.   The Pipeline level prefetch is essentially 
one operation for all Steps and continues to serve the purpose of broadly checking 
for errors and and caching references at the start of a Pipeline.

2. Disable the (dangerous) caching in the crds_client module.

Commented out active areas of CRDS client header + results caching effectively
disabling caching in crds_client w/o fully removing the code or refactoring
much.


jwst/assign_wcs/tests/test_nirspec.py
```diff
@@ -366,6 +366,7 @@ def test_shutter_state(open_shutters, main_shutter, result):
     assert shutter_state == result
 
 
+@pytest.mark.xfail(reason="Corrected reference after CRDS prefetch change producing wrong slit count.")
```
nden: 👍

jaytmiller: Thanks.   I will note that I cannot merge this,  so if there is no objection,  someone please merge this in now.   (Fingers crossed,  it should at least avoid future incorrect assignments like this one.)

nden: You should be able to merge now.

jaytmiller: Nadia added me to maintainers,  by the power vested in me as CRDS-dude,  I'm merging,  fingers crossed,  unless someone else does first.

hbushouse:
Fixes #1661

Issue #1668, opened 2018-01-18
Incorrect calculation of the median 1st difference is twopointdifference.py


mwregan2:
When calculating the median 1st difference twopointdifference.py uses the absolute value of the 1st differences. This is incorrect because it leads to a positive bias in the low to zero flux case. When the flux is zero the median 1st difference should be zero. Instead it's around 1 sigma. This leads to false detection of jumps.
  I have coded the fix and tested it. I'll be submitting a pull request.

hbushouse:
Fixed by #1670.

PR #1670, opened 2018-01-18
First diff fix

mwregan2:
This is the fix for the problem of using only positive first differences to calculate the median first difference. I also added a new test case that tests that the fix works.  #1668


jwst/jump/twopoint_difference.py
```diff
@@ -67,7 +67,7 @@ def find_CRs(data, gdq, read_noise, rej_threshold, nframes):
         number_sat_groups = (sat_groups * 1).sum(axis=2)
         ndiffs = ngroups - 1
         sort_index = np.argsort(positive_first_diffs)
```
hbushouse: Is it OK to still be using `positive_first_diffs` here, rather than `first_diffs`?

hbushouse:
Addresses #1668

mwregan2:
Howard,
  Yes, this one needs to use positive first differences to find large negative jumps. We want a list of differences that are in absolute magnitude order to flag the biggest changes.
Mike

PR #1680, opened 2018-01-22
install schemas

nden:
@jhunkeler The only way I can get `transform/schemas/*` installed from the new location is to add
`install_package_data` to `setup`. Does this look right?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,6 +73,25 @@ PACKAGE_DATA = {
 }
 
 
+def get_transforms_data():
+    # Installs the schema files in jwst/transforms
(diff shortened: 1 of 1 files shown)


nden:
This, apparently,  is not the best solution as this flags means all data files are installed including files like `*~`

nden:
Used a custom function to install the schemas in jwst/transforms, in order to avoid the additional junk installed by `install_package_data`. The reason the setuptools mechanism does not work is that the path contains a `.` in the directory name - `stsci.edu`.

nden:
Merging so it can be picked up by the nightly build.

PR #1689, opened 2018-01-26
Add model blender

stsci-hack:
This PR adds a new package to the jwst environment: model_blender. 

This package enables blending of input metadata to update the metadata in an output model without file I/O or use of FITS.  
 
Code for implementing the model_blender package serves as the core of this PR.  In addition, an initial sanity unit-test has also been included, along with updates to core schemas to add initial blending rules. 

A separate PR will be submitted to show how this code will be used with the resample step to replace the use of HST's fitsblender package.


hbushouse:
Looks good. Couple questions about how the rules are used.

1) If a given keyword has the "blend_table" attribute set, but not "blend_rule", does that mean it'll only show up in the table extension of the blended product and not in the header of the product? Or is their a default rule set so that everything shows up with at least some kind of value in the product header?

2) I assume that the rules "first" and "last" just pick off the value that occurs first or last in the input list of models. I could see where this might not be robust enough for keywords containing something like a time stamp, in the event that the list of models happens to be in a semi-random order (i.e. the first model in the list is not actually the exposure that occurred first in time). So we might need to be careful with the use of those particular rules. For things like a time stamp we'd want to use a rule that's more specific, like min and max.

PR #1698, opened 2018-01-31
Modify error handling for schema validation errors

bernie-simon:
This update modifies the interface for creating new datamodels and the generic open function. The new interface has two optional parameters, pass_invalid_values and strict_validation, which control the handling of schema errors. There are also two environment variables PASS_INVALID_VALUES and STRICT_VALIDATION, which if set, override the values of the parameters. The default values of both parameters are False. When the metadata values are read from a FITS file or modified in a script, the value is checked against the schema for that model. If the value does not agree with the schem an exception is thrown if strict_validation is set to True or a warning is issued if strict_validation is set to False. If the value does not match the schema, the metadata is set to that value if pass_invalid_values is True or it is set to None if pass_invalid_values is False.

Please note that the default behavior of datamodels differs from previous behavior. Schema errors now generate warnings instead of exceptions.

Because the error handling is more complex, the code has been centralized into a new function in util.py, validate_schema. All schema checking calls this code directly or indirectly.

Where practical, the schema checking code in properties.py has been modified to only check the value being modified. Previously it would check the entire metadata tree. This change should lead to a significant speed improvement when many metadata items are set, deleted, or appended to a list.

Several tests have been modified to correctly test the modified code. One new test has been added.

### DIFF
--- a/jwst/datamodels/fits_support.py
+++ b/jwst/datamodels/fits_support.py
@@ -438,32 +438,11 @@ def _schema_has_fits_hdu(schema):
     return has_fits_hdu[0]
 
 
-def _load_from_schema(hdulist, schema, tree, pass_invalid_values):
+def _load_from_schema(hdulist, schema, tree, context):
     known_keywords = {}
     known_datas = set()
     invalid_values = set()
     missing_values = set()
-
-    def build_errmsg():
-        # Prefix filename to error message where it can be found
-        errmsg = ""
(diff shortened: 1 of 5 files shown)


nden:
This was merged with a failing test.  The same failutre is in master and in the new PRs.

https://travis-ci.org/STScI-JWST/jwst/jobs/336194851

I can't tell at a glance what caused the failure though.

PR #1700, opened 2018-02-02
fix problems created by #1070

nden:
Make sure the wcs object is stored with the slit after extract_2d.
Fix copy and paste error.


nden:
@hbushouse After this "passes" I will merge it as it will take care of most errors in the nightly tests.

PR #1701, opened 2018-02-02
Additional log info for jump and photom

hbushouse:
Added a log message to jump/twopoint_difference to indicate loop over integrations and updated a log message in photom to include spectral order number along with the slit.


hbushouse:
So what do we do about the failing Travis nirspec test? It looks like it might've been caused by the changes to datamodels in #1698.
@nden @bernie-simon

bernie-simon:
I've looked at the problem a bit. It's the resurfacing of an old problem. Validation fails when setting complex objects. This first came to light when validating dates. I put a fix in and then took it out again, both times for good reasons. I'll have another look at this problem next week.

hbushouse:
Merging despite unrelated Travis error.

Issue #1702, opened 2018-02-02
change in behavior


nden:
#1698 introduce a change in behaviour. Here's a simple example which reproduces the problem:

```
from jwst import datamodels
from astropy.modeling import models
d=datamodels.DistortionModel()
d.model=models.Polynomial1D(1)
d.save('dist.asdf')

dist = datamodels.DistortionModel('dist.asdf')
dist.model(1)
TypeError: 'ObjectNode' object is not callable

```

hbushouse:
There's no way #1701 did this - it only contained a couple log message updates. I think you mean #1698 which included datamodels changes.

nden:
Yes, fixed the PR number. Hoping the example can help debug this.

bernie-simon:
> #1698 introduce a change in behaviour. Here's a simple example which reproduces the problem:

from jwst import datamodels
from astropy.modeling import models
d=datamodels.DistortionModel()
d.model=models.Polynomial1D(1)
d.save('dist.asdf')

dist = datamodels.DistortionModel('dist.asdf')
dist.model(1)
TypeError: 'ObjectNode' object is not callable

I believe the fix for metadata validation resolved this issue. At least, I could not reproduce the problem using the example code above. Please check to see if the problem still exists and if it doesn't, I will close the call.

nden:
Yes, this was resolved by the metadata validation changes.

bernie-simon:
I'm closing the call. Fixed by metadata validation change.

PR #1705, opened 2018-02-02
only temporary fix 

nden:
Attempt a temporary fix for #1702.


bernie-simon:
I'm in the process of making a similar fix.

PR #1707, opened 2018-02-03
allow 3d arrays in slits

nden:
Fixes problems with #1070.


hbushouse:
I assume this is to allow for handling NIRSpec BrightObj mode, which is TSO and hence 3D?

nden:
Yes, that's right. I forgot about this mode but a nightly test reminded me.

hbushouse:
👍  for regression tests!

PR #1708, opened 2018-02-05
Code changes for AMI3 and CORON3 to remove fitsblender

stsci-hack:
This set of changes removes the dependency on fitsblender from resample, calwebb_ami3, and calwebb_coron3 code.  It includes a removal of the resample.blend module which served as an interface from JWST to the HST fitsblender package.  

NOTE: The revision to resample also includes a change that explicitly limits the re-defintion of meta.wcsinfo attributes from the GWCS object to only ImageModel instances.  This change should not cause any problems, but merely serves as an example of what code calling model_blender will need to manage after it has finished blending the models metadata.


jwst/pipeline/calwebb_ami3.py
```diff
@@ -10,14 +10,14 @@
...
-from ..resample import blend
+from ..model_blender import blendmeta
 
 __version__ = '0.8.0'
 
 # Define logging
 import logging
 log = logging.getLogger()
-log.setLevel(logging.DEBUG)
+log.setLevel(logging.INFO)
```
hbushouse: Actually it's desired to keep this set at DEBUG level. We've discovered over the years that this setting acts as sort of a "floor" for the lowest level of messages that can/will be passed up by the routine. If it's set to INFO and the user requests DEBUG level, the DEBUG messages will NOT appear. If you set it here to DEBUG, the default level returned is INFO, but when the user requests DEBUG they'll actually show up.

PR #1712, opened 2018-02-06
fix last known error from reg tests

nden:
re #1070

### DIFF
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -265,7 +265,7 @@ def do_NIRSpec_flat_field(output_model,
     # check for that case.
     if not hasattr(output_model, "slits"):
         if exposure_type == "NRS_IFU":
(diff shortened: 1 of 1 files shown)


jwst/flatfield/flat_field.py
```diff
@@ -265,7 +265,7 @@ def do_NIRSpec_flat_field(output_model,
     # check for that case.
     if not hasattr(output_model, "slits"):
         if exposure_type == "NRS_IFU":
-            if not isinstance(output_model, datamodels.ImageModel):
+            if not isinstance(output_model, datamodels.IFUImageModel):
                 log.error("NIRSpec IFU data is not an ImageModel; "
```
hbushouse: Log messages here and line 271 should update `ImageModel` to `IFUImageModel`

nden: Done. Thanks!

PR #1718, opened 2018-02-07
Validate individual item changed instead of whole tree when setting metadata

bernie-simon:
Validating an individual item instead of the whole schema saves time when updating a metadata item. I backed out a previous change to do this because it failed on complex values. This new change validates a simplified version of the value the item value, as it should.


hbushouse:
Is this PR now obsolete due to the changes in #1747 or do we still need these changes too?

bernie-simon:
The change was made, partly backed out when it caused a regression, and then fixed and completed in #1747.

hbushouse:
In that case, closing.

PR #1723, opened 2018-02-07
gracefully skip extract_2d if assign_wcs was skipped

### LINKED ISSUE #1713: calwebb_spec2 problems when there aren't any shutters/slits on the detector
In https://jira.stsci.edu/browse/JP-193 a tester reported a crash of the calwebb_spec2 pipeline when testing on a NIRSpec MSA exposure that used an MSA meta data file with only 1 open shutter, which happened to be a shutter listed in the failed open/closed shutter reference file as stuck closed. Assign_wcs correctly warned that there weren't any shutters that projected onto the detector, which means no wcs info was created. The extract_2d step then crashed when it couldn't find the wcs object.

Something somewhere should be made more robust to handle situations like this (a similar one is when no fixed slits project onto the NRS2 detector), such that when assign_wcs sees that there aren't any spectra to process the whole pipeline stops and shuts down gracefully at that point already.

Details are in the basecamp messages at:
https://stsci-ins.basecamphq.com/projects/11477312-jwst-pipeline/posts/104252645/comments

nden:
Resolves #1713

### DIFF
--- a/jwst/assign_wcs/assign_wcs.py
+++ b/jwst/assign_wcs/assign_wcs.py
@@ -37,7 +37,7 @@ def load_wcs(input_model, reference_files={}):
     # Make the copy after the WCS pipeline is created in order to pass updates to the model.
     if pipeline is None:
         input_model.meta.cal_step.assign_wcs = 'SKIPPED'
-        log.warning("SKIPPED assign_wcs")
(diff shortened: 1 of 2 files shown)


nden:
@hbushouse The JIRA ticket can be closed too.

Issue #1715, opened 2018-02-07
skymatch error handling


hbushouse:
A B7.1 tester encountered a situation where overly aggressive jump detection resulted in flagging so many pixels that there were apparently no good/useable pixels left when skymatch attempted to measure the signal in overlap regions, leading to a crash in numpy linear algebra routines. See https://stsci-ins.basecamphq.com/projects/11477312-jwst-pipeline/posts/104257921/comments#379445833

If possible (without great effort) it would be good to try to catch such a situation (no pixels in overlap region) in skymatch and have it issue a warning that it can't compute sky levels and return the input images unaltered, all without crashing.

This issue is also being tracked in https://jira.stsci.edu/browse/JP-224

hbushouse:
Fixed by #1724.

PR #1724, opened 2018-02-08
Make skymatch quietly report lack of data

mcara:
This PR should address https://github.com/STScI-JWST/jwst/issues/1715 in that now `skymatch` will no longer throw an exception when either images do not intersect at all or when there are no valid (=non-flagged) data in the overlap regions. Instead, `skymatch` will print a warning message and will set sky values to `0.0`.

### DIFF
--- a/jwst/pipeline/skymatch.cfg
+++ b/jwst/pipeline/skymatch.cfg
@@ -5,6 +5,7 @@ skymethod = 'global+match'
 match_down =True
 subtract = False
 skystat = 'mode'
+dqbits = 0
 nclip = 5
 lsigma = 4.0
 usigma = 4.0
--- a/jwst/skymatch/skymatch.py
+++ b/jwst/skymatch/skymatch.py
@@ -543,7 +543,14 @@ def _find_optimum_sky_deltas(images, apply_sky=True):
                 invalid[j] = False
                 ieq += 1
 
-    rank = np.linalg.matrix_rank(K, 1.0e-12)
+    try:
+        rank = np.linalg.matrix_rank(K, 1.0e-12)
+    except np.linalg.LinAlgError:
+        log.warning("Unable to compute sky: No valid data in common "
+                    "image areas")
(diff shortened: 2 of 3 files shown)


hbushouse:
I have another suggestion that would help alleviate the problem that skymatch was having with a couple of NIRISS test images. The images in question had some NaN pixels, but fortunately those pixels also had non-zero DQ values. Running skymatch with the default dqbits=None causes the statistics routines called by _overlap_matrix to throw an error, due to the NaN's. However, if you set dqbits = '0', which will cause the routines to only consider pixels with DQ=0 as good (an exclude all others), it succeeds.

So we should consider 2 things:
- a way for some of the lower-level routines to handle NaN's
- setting the default dqbits to something other than None (e.g. '0')

mcara:
@hbushouse But then images with almost all pixels flagged as it was done in initial tests would have resulted in code crashing due to lack of data (this PR fixes this) 😄 

In my opinion, the code should crash when "valid" (according to the user) data contain invalid data. For example, if someone has an image with NaNs then they should create a mask and pass it to `skymatch` which is not overly complicated:

```python
mask = ~np.isfinite(image)
```

mcara:
That is, having the code check for NaNs defeats the purpose of `dqmask`. However, if you insist I will be happy to implement it in the **`step`**.

mcara:
@hbushouse If I change default parameter value in the step (in the docstring for `spec`), do I also need to change it in a "configuration file"? If so, where do these files reside?

mcara:
Changed default `dqbits` to 0. Let me know if you would like to make code mask `NaN`s and/or `inf` before proceeding.

mcara:
@hbushouse Changed default `dqbits` to 0. Let me know if you would like to make code mask `NaN`s and/or `inf` before proceeding.

mcara:
I just went ahead and implemented automatic filtering of non-finite data.

hbushouse:
Fixes #1715

PR #1727, opened 2018-02-08
fix s_region

nden:
#1722 was not quite right. `WCS.footprint` is a (3, 3) array for spectral observations. Also there was a numerical typo - we are looking for negative value of RA, not values < 360.

### DIFF
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -548,13 +548,15 @@ def update_s_region(model):
     if bbox is None:
         bbox = _bbox_from_shape(model)
 
-    #footprint = model.meta.wcs.footprint(bbox, center=True).T
-    ra, dec = model.meta.wcs.footprint(bbox, center=True)
-    negative_ind = ra < 360
-    if negative_ind.any():
-        ra[negative_ind] = 360 + ra[negative_ind]
+    # footprint is an array of shape (2, 2) or (3, 3)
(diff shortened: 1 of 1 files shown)


jwst/assign_wcs/util.py
```diff
@@ -548,13 +548,15 @@ def _bbox_from_shape(model):
...
+    footprint = model.meta.wcs.footprint(bbox, center=True)
+
+    # Make sure RA values are all positive
+    negative_ind = footprint[0] < 0
     if negative_ind.any():
-        ra[negative_ind] = 360 + ra[negative_ind]
+        footprint[0][negative_ind] = 360 + footprint[0][negative_ind]
 
-    footprint = np.array([ra, dec]).T
+    footprint = footprint.T
```
hbushouse: So if we're dealing with a spectroscopic case where footprint is a 3x3 array, are we going to get tuples of RA, Dec, lambda values written out to the S_REGION string, as opposed to just RA, Dec?

hbushouse: Or at least update the way the S_REGION string is populated so that we only pick out values of RA and Dec from the contents of `footprint`?

nden: We are ignoring the spectral part because at least for now it creates NaNs. For spectral observations s_region="". I've pushed a commit to make this explicit.

PR #1734, opened 2018-02-09
some steps need to work with SlitModel instead of ImageModel

nden:
more #1070 related changes

Because the items in `MultiSlitModel.slits`  are now instances of `SlitModel` and not `ImageModel` some models need to check for `SlitModel` as input. 
In addition there are some minor changes in other packages.

This is the result of debugging a problem with resample_spec_step when run on NRS Fs data. The error as seen in the reg tests is:

```
/data4/iraf_conda/bldtmp/pF3k/p/envs/dev/lib/python3.5/site-packages/jwst-0.9.1a0.dev108-py3.5-linux-x86_64.egg/jwst/resample/resample_spec_step.py", line 61, in process
    output_product.products[-1].bunit_data = input_models[0].meta.bunit_data
  File "/data4/iraf_conda/bldtmp/pF3k/p/envs/dev/lib/python3.5/site-packages/jwst-0.9.1a0.dev108-py3.5-linux-x86_64.egg/jwst/datamodels/properties.py", line 274, in __getitem__
    return _make_node(self._instance[i], schema, self._ctx)
```

However, this does not resolve the above problem. I think some of the schemas used by resample_spec_step may need to be updated to include the `slit` schema instead of individual items from it. @stscieisenhamer may know more about these schemas and how they are used, specifically the `product` schemas.

### DIFF
--- a/jwst/datamodels/slit.py
+++ b/jwst/datamodels/slit.py
@@ -33,9 +33,9 @@ class SlitDataModel(model_base.DataModel):
 
     def __init__(self, init=None, data=None, dq=None, err=None,
                  wavelength=None, var_poisson=None, var_rnoise=None,
-                 relsens=None, area=None, 
+                 relsens=None, area=None,
                  wavelength_pointsource=None, pathloss_pointsource=None,
-                 wavelength_uniformsource=None, pathloss_uniformsource=None, 
+                 wavelength_uniformsource=None, pathloss_uniformsource=None,
                  **kwargs):
         if isinstance(init, (SlitModel, ImageModel)):
             super(SlitDataModel, self).__init__(init=None, **kwargs)
@@ -141,6 +141,7 @@ class SlitModel(model_base.DataModel):
             self.err = init.err
             self.relsens = init.relsens
             self.area = init.area
+            self.wavelength = init.wavelength
             if hasattr(init, 'var_poisson'):
                 self.var_poisson = init.var_poisson
             if hasattr(init, 'var_rnoise'):
@@ -170,7 +171,7 @@ class SlitModel(model_base.DataModel):
 
         if var_rnoise is not None:
             self.var_rnoise = var_rnoise
-        
+
         if dq is not None:
             self.dq = dq
         if err is not None:
--- a/jwst/datamodels/slit.py
+++ b/jwst/datamodels/slit.py
@@ -141,7 +141,8 @@ class SlitModel(model_base.DataModel):
             self.err = init.err
             self.relsens = init.relsens
             self.area = init.area
-            self.wavelength = init.wavelength
+            if hasattr(init, 'wavelength'):
+                self.wavelength = init.wavelength
             if hasattr(init, 'var_poisson'):
                 self.var_poisson = init.var_poisson
             if hasattr(init, 'var_rnoise'):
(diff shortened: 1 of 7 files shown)


jwst/datamodels/slit.py
```diff
@@ -155,10 +170,23 @@ def __init__(self, init=None, data=None, dq=None, err=None,
...
-
+        
+        if dq is not None:
+            self.dq = dq
+        if err is not None:
+            self.err = err
+        if bunit_data is not None:
+            self.meta.bunit_data = bunit_data
+        if bunit_err is not None:
+            self.meta.bunit_err = bunit_err
```
hbushouse: Since when is it necessary to populate meta attributes in datamodel class?

nden: It's not necessary in general. It only covers the case when a model is initialized not from a file but passing in the (items, values) specified in the schema. 
When the data arrays are extracted, a SlitDataModel is initialized by passing the arrays to the initializer. The meta data don't matter in this case as they are stored in MulitSlitModel.meta. But when it's written to a fits file the extensions need to have this kw in the header. That's why it's added to the initializer. This is all used internally in extract_2d.

nden: In the future, my goal is to completely strip all `meta` items from `SlitDataModel`, including the wcs and keep everything in MultiSlitModel.meta. Then when a SlitModel is created from a SlitDataModel the meta and the wcs will be populated. But let's get first this working again.

jwst/extract_1d/extract.py
```diff
@@ -1316,8 +1316,9 @@ def do_extract1d(input_model, refname, smoothing_length, bkg_order):
                     spec.name = slitname
                 output_model.spec.append(spec)
 
-        elif isinstance(input_model, datamodels.CubeModel):
+        elif isinstance(input_model, (datamodels.CubeModel, datamodels.SlitModel)):
```
hbushouse: So do the NIRSpec BrightObj exposures now come through here as a SlitModel, instead of CubeModel? If so, I guess that implies that SlitModel can handle 3D data?

nden: Yes, NRS_BRIGTOBJ comes out of extarct_2d as a SlitModel. SlitModel acts as enhanced version of either ImageModel or CubeModel. The issue with the test is solved in the SlitModel class. When the input to a  SlitModel is a SlitModel (instead of a file) it was not copying the wavelength array and that forced it to created a wavelength array with the sahpe of the data which is 3D. For NRS_BRIGHTOBJ the dat ais 3D but the wavelength array is 2D.

nden: Oh, this also means that the output of calweb_spec2 is a SlitModel in this case.
Does this have any implications for the spec3 pipeline?

hbushouse: Possibly. We'll obviously need to check that. calwebb_spec3 might also need updates to account for the fact that input IFU files are no longer an ImageModel, but IFUImageModel. As I looked at the changes in this PR it also made me wonder if the Photom step is now going to have issues because of the NIRSpec BrightObj data coming through as SlitModel, instead of CubeModel. I guess the reg tests will let us know.

nden: The photom step was also adjusted in this PR.

PR #1737, opened 2018-02-09
Cube wavetab

jemorrison:
Improved how the ave min, and max right ascension of the set of images to create the IFU cube from is determined.

### DIFF
--- a/jwst/datamodels/schemas/ifucube.schema.yaml
+++ b/jwst/datamodels/schemas/ifucube.schema.yaml
@@ -35,6 +35,10 @@ allOf:
           title: Weighting power for Modified Shepard Method
           type: number
           fits_keyword: WPOWER
+        channel:
+          title: MIRI MRS channels represented in the Cube
+          type: string
+          fits_keyword: CHANNEL
 
 - type: object
   properties:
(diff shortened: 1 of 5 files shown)


jwst/datamodels/schemas/ifucube.schema.yaml
```diff
@@ -35,6 +35,10 @@ allOf:
           title: Weighting power for Modified Shepard Method
           type: number
           fits_keyword: WPOWER
+        channel:
```
hbushouse: Can't the channel attribute (and corresponding keyword) that's already defined in core.schema.yaml be used for this, rather than duplicating it again here? The attribute is meta.instrument.channel. I notice that the current keyword comment field for that is "NIRCam channel: long or short", but that should be updated anyway, because the CHANNEL keyword is already used for both NIRCam and MIRI. We could update the comment to something more generic, like "Instrument channel."

jemorrison:
First of all the channel comment should be changed - Instrument channel sounds good to me.
I put the channel in ifucube.schema.yaml so as a test for what I am going to do with blenders.
In the core schema channel can have values of 12 or 34, but this channel - which should probably
be renames to blend_channel or something - contains all the channels that are in the IFU Cube.
I have not used this variable yet - it was just me starting to code up how to use the blend header information. I will take it out until I think about this some more.

Issue #1738, opened 2018-02-09
Allow calspec2 grism processing to use SCATFILE value as input


hbushouse:
Currently, the calwebb_spec2 pipeline assumes that WFSS (grism) exposures will be processed using an ASN file as input to the pipeline, where that ASN file contains an entry specifying the source catalog file to be used in steps like extract_2d. After getting the name of the source catalog file from the ASN, the calwebb_spec2 pipeline stores that file name in model.meta.source_catalog, for use by the other steps. That results in the SCATFILE keyword in output products being populated with the name of catalog file.

It has been suggested (see https://jira.stsci.edu/browse/JP-190) that it would be a convenience for users if the calwebb_spec2 pipeline could be run using just the grism exposure as input, rather than an ASN file, in which case the name of the source catalog would be retrieved from the SCATFILE keyword in the input image (the user would have to set the keyword value).

This has no use outside of DMS operations, and hence is relatively low priority for implementation. Will implement in B7.2 if time allows.

hbushouse:
Fixed by #1740

PR #1740, opened 2018-02-12
Allow catalog to be passed to pipeline from datamodel instead of association

sosey:
addresses #1738.

This updates the WFSS section of the Spec2 pipeline to look for the name of the source catalog only if the datamodel is not populated with a value for  datamodel.meta.source_catalog.filename, which is read from the SCATFILE FITS keyword. 

Do the science teams have a preference? What if an association is given that has a catalog name AND the datamodel also specifies a catalog name, which should take precedence?

### DIFF
--- a/jwst/pipeline/calwebb_spec2.py
+++ b/jwst/pipeline/calwebb_spec2.py
@@ -142,7 +142,12 @@ class Spec2Pipeline(Pipeline):
         # a grism image, if so get the catalog
         # name from the asn and record it to the meta
         if exp_type in WFSS_TYPES:
-            input.meta.source_catalog.filename = members_by_type['sourcecat'][0]
+            if input.meta.source_catalog.filename is None:
+                try:
+                    input.meta.source_catalog.filename = members_by_type['sourcecat'][0]
+                except IndexError:
+                    raise IndexError("No source catalog specified in association or datamodel")
+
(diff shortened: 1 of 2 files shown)


hbushouse:
In terms of precedence, I would always give priority to a catalog specified in the ASN, in which case the header keyword then gets updated to record that value as info.

jwst/pipeline/calwebb_spec2.py
```diff
@@ -142,7 +142,12 @@ def process_exposure_product(
         # a grism image, if so get the catalog
         # name from the asn and record it to the meta
         if exp_type in WFSS_TYPES:
-            input.meta.source_catalog.filename = members_by_type['sourcecat'][0]
+            if input.meta.source_catalog.filename is None:
```
hbushouse: This logic gives precedence to the keyword value. I think we should have the value in the ASN file take precedence, since that was the original design. So look for that first and then if it doesn't exist, try to get it from the keyword. As soon as a user runs the pipeline once on a given exposure, the keyword will then be populated and hence upon subsequent executions (if they rerun it) it was always default to using the old keyword value. I think it should come from the ASN, when available.

sosey: okay, I will reverse the logic an make the asn take precedence.

PR #1741, opened 2018-02-12
Straylight2

jemorrison:
added straylight removal by using Modified Shepard Method to  find the distances of the pixels in the slice gaps. 
This method is set to be used by default, but the other straylight method was kept and can be used
by --method=Nearest. We can probably remove the old method (nearest) at a later date after INS has tested this method. For now I left it.

### DIFF
--- a/jwst/cube_build/cube_build_wcs_util.py
+++ b/jwst/cube_build/cube_build_wcs_util.py
@@ -14,6 +14,8 @@ from .. import datamodels
 from ..assign_wcs import nirspec
 from . import coord
 from gwcs import wcstools
+from astropy.stats import circmean
+from astropy import units as u 
 
 import logging
 log = logging.getLogger(__name__)
@@ -115,7 +117,6 @@ def setup_wcs(self):
                                                        this_a,
                                                        self.instrument_info)
                     amin, amax, bmin, bmax, lmin, lmax = ch_footprint
-
 
 # If a dither offset list exists then apply the dither offsets (offsets in arc seconds)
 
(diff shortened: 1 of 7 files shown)


jemorrison:
I removed 'channel' from IFUCube model schema. This will all be handled by model_blender

jwst/straylight/straylight.py
```diff
@@ -214,12 +185,139 @@ def correct_MRS(input_model, straylight_model):
...
+    #plt.show()
+
+    # The regions mask has values of 0 and nonzero, values of 0 present pixel
+    # in a slice gap and non zero values are science pixels.  The straylight task uses data
+    # in-between the slices (also called slice gaps) of the MRS data to correct 
+    # the science data in the slices. 
+
+    # Create output as a copy of the input science data model
+    output = input_model.copy() # this is used in algorithm to
+    output_data = input_model.data.copy()
```
hbushouse: Why is `output_data` needed as a separate array? Can't you just use `output.data`? They both contain the same array.

jwst/straylight/straylight_step.py
```diff
@@ -3,45 +3,73 @@
...
-
+from ..datamodels import RegionsModel
 
 class StraylightStep (Step):
     """
     StraylightStep: Performs straylight correction image using a Mask file.
     """
-    reference_file_types = ['straymask']
+
+    spec = """
```
hbushouse: These new arguments also need to be added to the default `straylight.cfg` file in the pipeline directory.

hbushouse:
This PR also contains updates to cube_build from #1737

jwst/cube_build/cube_build_wcs_util.py
```diff
@@ -582,6 +648,82 @@ def set_geometryAB(self, footprint):
 
 
 
+#_______________________________________________________________________
+# from a set of ra values find the average ra.
+# This is tricky because of the convergence of hour angles
+# this method is taken from the SPITZER SSC MOSAICER tool
+def average_ra(ravalues):
+
```
crawfordsm: I was just taking a quick look through this, but did you consider using [astropy.units.Angle](http://docs.astropy.org/en/stable/coordinates/angles.html) and [astropy.stats.circmean](http://docs.astropy.org/en/stable/stats/circ.html)?

jemorrison: I was trying to find something in astro py that did something like this. I thought it must exist but
I could not find it. I will look at those two routines in astropy.

jemorrison:
Now using astropy.stats.circmean to find the average right ascension

jwst/straylight/straylight_step.py
jemorrison: Do we put all the default values in the cfg files  ?
If so then I need to update the cube_build on too.

hbushouse: Yes. See, for example, https://github.com/STScI-JWST/jwst/blob/master/jwst/pipeline/skymatch.cfg

hbushouse:
Given that the straylight algorithm has been replaced I assume this means the documentation needs to be replaced/updated too. I don't see documentation updates in this PR. That needs to be done yet.

PR #1747, opened 2018-02-14
Verify individual attributes when getting and setting

bernie-simon:
This update corrects and completes the change to validating attributes when they are set. It creates a new node, validates the node, and if the validation does not fail, it sets the attribute. This update
corrects the previous fix by calling custom_tree_to_tagged_tree on the node's data prior to validation. This function converts asdf types into lists and dictionaries which can be validated by the json/yaml parser.

This update also revises the code in resample_spec which builds the metadata structure appended to the datamodel. Because validation is an all or nothing operation, it is important when building the metadata structure to check the result of getattr for None. That indicates the value was not found in the input. Blindly copying the value from input to output without checking would result in spurious values being copied.

Why not call hasattr before getattr instead of checking the value of getattr for None? In datamodels hasattr indicates the value is either in the model or in the schema. This was not changed as a good bit of code expects that hasattr works this way. So to check if a value is in the data, check the return value of getattr for None. To check if the value is in the schema, call hasattr. Resample_spec has been revised to follow this guideline.

### DIFF
--- a/jwst/datamodels/properties.py
+++ b/jwst/datamodels/properties.py
@@ -97,7 +97,7 @@ def _make_default(attr, schema, ctx):
         return {}
     elif schema.get('type') == 'array':
         return []
-    return copy.deepcopy(schema.get('default'))
+    return None
 
 
 def _make_node(instance, schema, ctx):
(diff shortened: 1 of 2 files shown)


hbushouse:
I've verified that this at least fixes the problem enough that the resample_spec step no longer crashes when processing a NIRSpec fixed-slit exposure containing multiple slits, but a detailed comparison of the old and new _s2d product coming out of resample_spec shows:
- the SCI header keywords SRCXPOS and SRCYPOS are no longer in the s2d product
- none of the RELSENS extensions got carried over into the s2d product
These discrepancies may be related to updates that are needed in the DrizProductModel or MultiProductModel in order to work properly with the new SlitModel.

PR #1758, opened 2018-02-20
New noise model

stsci-hack:
These changes remove the dependence of the readnoise and gain reference files from the outlier_detection step.  The use of these reference files are replaced with use of the ERR arrays from each of the input files.  This version includes conversion of the ERR arrays to remove NaNs since the current computations still results in NaNs in the ERR array. 

The weighting of the outlier_detection step can eventually be changed to use 'error' instead of 'exptime' to better account (on a pixel-by-pixel basis) for the readout characteristics of each exposure.  However, that will wait until the computation of the error arrays does not include NaNs.  

This PR leaves the reffile interface for the outlier_detection step to support any use of outlier-specific reference files for parameter settings (if that ever gets implemented).

### DIFF
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -472,14 +472,6 @@ def flag_cr(sci_image, blot_image, **pars):
     np.bitwise_or(sci_image.dq, np.invert(cr_mask) * CRBIT, sci_image.dq)
 
 
-def build_mask(dqarr, bitvalue):
-    """Build a bit-mask from an input DQ array and a bitvalue flag."""
-    bitvalue = bitmask.interpret_bit_flags(bitvalue)
-    if bitvalue is None:
-        return (np.ones(dqarr.shape, dtype=np.uint32))
-    return np.logical_not(np.bitwise_and(dqarr, ~bitvalue)).astype(np.uint32)
-
-
 def abs_deriv(array):
     """Take the absolute derivate of a numpy array."""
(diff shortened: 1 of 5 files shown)


jwst/outlier_detection/outlier_detection.py
```diff
@@ -392,39 +375,7 @@ def flag_cr(sci_image, blot_image, gain_image, readnoise_image, **pars):
     blot_data = blot_image.data * exptime
     blot_deriv = abs_deriv(blot_data)
 
-    # This mask can take into account any crbits values
-    # specified by the user to be ignored.
-    # dq_mask = build_mask(sci_image.dq, CRBIT)
```
hbushouse: I see that the build_mask function is still included in the code, but it isn't used anywhere. Will it be in the future?

jwst/outlier_detection/outlier_detection_scaled_step.py
```diff
@@ -33,8 +33,6 @@ class OutlierDetectionScaledStep(Step):
         save_intermediate_results = boolean(default=False)
         good_bits = integer(default=4)
     """
-    reference_file_types = ['gain', 'readnoise']
```
hbushouse: Yay! No more fetching dozens of reference files! 👍

hbushouse:
Addresses #1759. Partially addresses #612.

jwst/outlier_detection/outlier_detection.py
stsci-hack: That function has now been removed.  We have another copy of it in the resample package if needed by any JWST code.

Issue #1379, opened 2017-10-26
outlier_detection bitmask and dq array handling needs to be updated


jdavies-st:
Update `outlier_detection` to use the `datamodels.dqflags.pixel` dictionary of named flags instead of raw bit values.

So instead of the code explicitly referring to bit 4, instead refer to `dqflags.pixel['JUMP_DET']` where needed.

stsci-hack:
No changes can be implemented for the use of hard-coded dqflag values in the outlier_detection as those values are used in defining the default parameter values for the step which requires hard-coded values, not variables.  

However, the resample package relied on default dqflag values as default parameter values and those instances should be corrected.

PR #1760, opened 2018-02-21
Remove hard-coded dqflag use from resample

stsci-hack:
This addresses #1379 .  See comments in issue regarding the scope of these changes. 

It also includes removal of yet another extraneous import of a stsci package; namely, stsci.tools.bitmask.

### DIFF
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -8,6 +8,8 @@ from . import gwcs_drizzle
 from . import resample_utils
 from ..model_blender import blendmeta
 
+CRBIT = np.uint32(datamodels.dqflags.pixel['JUMP_DET'])
+
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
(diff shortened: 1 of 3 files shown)


jwst/resample/resample.py
```diff
@@ -36,7 +38,7 @@ class ResampleData:
     drizpars = {'single': False,
                 'kernel': 'square',
                 'pixfrac': 1.0,
-                'good_bits': 4,
+                'good_bits': CRBIT,
```
hbushouse: Remind me why it is we want resample to treat pixels flagged as CR as a good pixel (instead of rejecting them)?

stsci-hack: No one else has provided any other value to use as a default.  This can certainly be updated to better match what the instrument teams feel would be best.  We just need a new value from someone...

hbushouse: Since this is being run after outlier_detection, which inserts JUMP_DET flags, we definitely don't want to treat those as good. Looking over the list of predefined flag conditions I don't really see any that would be totally harmless or OK to include when combining data, so I would set the default to zero (i.e. reject anything with DQ > 0). That way we also automatically get rid of reference pixels from mosaics, assuming of course that the reference pixels were flagged upstream with `REFERENCE_PIXEL`.

Issue #925, opened 2017-05-25
Outlier detection background subtraction needs fixing


jdavies-st:
Background subtraction is not currently handled correctly in `outlier_detection`.  The `skymatch` step in `Image3Pipeline` computes the background for each input detector exposure and writes them to metadata.  Currently, `outlier_detection` resamples those images without subtracting off the background first.  In cases where background varies significantly between exposures, the subsequent blotted median will have a pedestal offset compared the the detector exposures.

Instead, the background should be subtracted before the resampling in `outlier_detection`, and the background value needs to be taken into account for the noise model in `flag_cr`.

The final resampling in `Image3Pipeline` should be done after sky background has been subtracted.  Perhaps this can be an argument to resample?  Or it can be done in the pipeline itself.

hbushouse:
#1761 is labeled as "initial corrections". What additional work, if any, still needs to be done for B7.3?

hbushouse:
See https://jira.stsci.edu/browse/JP-477

hbushouse:
Closing, because it's now superseded and duplicated by #2930.

PR #1761, opened 2018-02-21
Initial corrections for background subtraction

stsci-hack:
This set of changes includes:
- updates to resample to correctly apply computed background values
- an update to outlier_detection to recognize when background values are available for use
- an update to tweakreg default parameters to make allow tweakreg to work on a wider range of datasets (including simulated NIRCAM associated images that include uncal files)

 This should mostly address issue #925 .

### DIFF
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -361,13 +361,10 @@ def flag_cr(sci_image, blot_image, **pars):
     snr1, snr2 = [float(val) for val in pars.get('snr', '5.0 4.0').split()]
     scl1, scl2 = [float(val) for val in pars.get('scale', '1.2 0.7').split()]
 
-    if sci_image.meta.background.subtracted:
+    if not sci_image.meta.background.subtracted:
+        # Include background back into blotted image for comparison
         subtracted_background = sci_image.meta.background.level
         log.debug("Subtracted background: {}".format(subtracted_background))
-    else:
-        subtracted_background = backg
-        log.debug("No subtracted background found. "
-                  "Using default value from outlierpars: {}".format(backg))
 
     exptime = sci_image.meta.exposure.exposure_time
 
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -190,8 +190,8 @@ class ResampleData:
                 exposure_times['end'].append(img.meta.exposure.end_time)
 
                 # apply sky subtraction
-                if 'skybg' in img.meta._instance:
-                    img.data -= img.meta.skybg
+                if not img.meta.background.subtracted:
+                    img.data -= img.meta.background.level
 
                 outwcs_pscale = output_model.meta.wcsinfo.cdelt1
                 wcslin_pscale = img.meta.wcsinfo.cdelt1
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -190,8 +190,9 @@ class ResampleData:
                 exposure_times['end'].append(img.meta.exposure.end_time)
 
                 # apply sky subtraction
-                if not img.meta.background.subtracted:
-                    img.data -= img.meta.background.level
+                blevel = img.meta.background.level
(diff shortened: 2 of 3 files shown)


jwst/outlier_detection/outlier_detection.py
```diff
@@ -361,13 +361,10 @@ def flag_cr(sci_image, blot_image, **pars):
     snr1, snr2 = [float(val) for val in pars.get('snr', '5.0 4.0').split()]
     scl1, scl2 = [float(val) for val in pars.get('scale', '1.2 0.7').split()]
 
-    if sci_image.meta.background.subtracted:
+    if not sci_image.meta.background.subtracted:
```
hbushouse: I don't understand why the logic has been flipped here. If the background has NOT already been subtracted, then why is it being reported as subtracted and the (unsubtracted) level passed along for use elsewhere?

jwst/resample/resample.py
```diff
@@ -190,8 +190,8 @@ def do_drizzle(self):
                 exposure_times['end'].append(img.meta.exposure.end_time)
 
                 # apply sky subtraction
-                if 'skybg' in img.meta._instance:
-                    img.data -= img.meta.skybg
+                if not img.meta.background.subtracted:
+                    img.data -= img.meta.background.level
```
hbushouse: Do we need a check here to make sure that img.meta.background.level actually exists and/or is populated with a value? E.g. what if the skymatch step was not run upstream? If so, this attribute could be None and hence the arithmetic operator would not be too keen about that.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -51,7 +51,7 @@ class TweakRegStep(Step):
 
         # Object matching parameters:
         minobj = integer(default=15) # Minimum number of objects acceptable for matching
-        searchrad = float(default=1.0) # The search radius in arcsec for a match
+        searchrad = float(default=10.0) # The search radius in arcsec for a match
```
hbushouse: I know we needed a value of 10" in order to get tweakreg to work on the simulated data that had bogus RA/Dec values, but I'm not sure we want to make that the default. Could such a large search radius ever cause harm when running it on data that have semi-accurate pointing info to begin with? I would think the default should be somewhere on the order of what we think the pointing uncertainty will be (with some buffer) and then let folks use non-default (larger) values when necessary.

stsci-hack: Fortunately, a large default value has no real bearing on run-time or accuracy of the final result unless, by some rare theoretical possibility, an extremely crowded field confuses the algorithm with a many spurious matches at various deltas (an occurance I have never seen or heard of, BTW).  Besides, for HST, I have seen delta's larger than 10 arcseconds (some have been determined to be over 60).  Granted, JWST should be more accurate, but this default should allow the pipeline to run on everything with, IMO, minimal/no risk of problems.  It can be reduced to original value if requested.

jwst/outlier_detection/outlier_detection.py
stsci-hack: The background level does NOT get subtracted by skymatch from the original input _cal.fits files.  The value only gets applied when resampling the image.  Thus, when outlier_detection gets run, it needs to convert a background-subtracted-median-blotted image with the original non-background-subtracted values in the input _cal file.  If a change was made or the user performed background-subtraction on the _cal.fits file before running outlier_detection, then we would not need to get the background value to amend the background-subtracted-blot image.   I agree it seems backwards on the face of it, but I believe it insures the right values are used for the detection of bad pixels/cosmic-rays.

jwst/resample/resample.py
stsci-hack: It probably would be a good idea to add such a check.  I will look into it immediately.

jwst/outlier_detection/outlier_detection.py
hbushouse: 👍

jwst/tweakreg/tweakreg_step.py
hbushouse: 👍

Issue #1142, opened 2017-08-23
LG fit orientation is incorrect in ami_analyze step


hbushouse:
The AMI working group testing of Build 7 revealed that the orientation of the LG fit that's applied in the ami_analyze step is in the native detector readout orientation and needs to be updated to be in the DMS orientation, so that it's consistent with the images coming through the pipeline (see https://jira.stsci.edu/browse/JP-176).

Awaiting input from the AMIWG as to how to make this update.

hbushouse:
High priority for B7.2, because it's a bug in a previously delivered level 4 requirement (NIRISS AMI processing in B7).

hbushouse:
Fixed by #1764, modulo any issues that the AMI WG might find once they've tested the updates. But basic development work has been completed, so closing this.

PR #1764, opened 2018-02-23
Implementing a bug fix for the LG fit orientation; the fit will now b…

dmggh:
…e in the DMS orientation. This version is what Anand et al call LG+. Removed functions and variables that are no longer used - although some have been left in for possible use by future updates. Also numerous PEP8-related changes, and improved documentation.

### DIFF
--- a/jwst/ami/NRM_Model.py
+++ b/jwst/ami/NRM_Model.py
@@ -1,6 +1,13 @@
 #
 # A module for conveniently manipulating an 'NRM object' using the
 # Lacour-Greenbaum algorithm. First written by Alexandra Greenbaum in 2014.
+#
+# This module:
+#   Defines mask geometry and detector-scale parameters
+#   Simulates PSF (broadband or monochromatic)
+#   Builds a fringe model - either by user definition, or automated to data
+#   Fits model to data by least squares
+#
 # Algorithm documented in:  Greenbaum, A. Z., Pueyo, L. P.,
 # Sivaramakrishnan, A., and Lacour, S. ; Astrophysical Journal (submitted) 2014.
 # Developed with NASA APRA (AS, AZG), NSF GRFP (AZG), NASA Sagan (LP), and
@@ -11,13 +18,15 @@
 # - updated May 2013 to include hexagonal envelope
 # - updated (hard refactored) Oct-Nov 2014 Anand S.
 
-from __future__ import absolute_import, division
+from __future__ import (absolute_import, division)
 
 import logging
 import numpy as np
+
 from . import leastsqnrm as leastsqnrm
 from . import analyticnrm2
 from . import utils
+from . import hexee
 from . import NRM_consts
 
 log = logging.getLogger(__name__)
@@ -26,9 +35,9 @@ log.setLevel(logging.DEBUG)
 
 class NRM_Model():
 
-    def __init__(self, mask=None, holeshape="circ",
-                  pixscale=leastsqnrm.mas2rad(64), rotate=False,
-                  over=None, scallist=None, rotlist_deg=None):
+    def __init__(self, mask=None, holeshape="circ", pixscale=hexee.mas2rad(65),
+                 rotate=False, over=1, flip=False, pixweight=None, scallist=None,
+                 rotlist_deg=None, phi="perfect"):
         """
         Short Summary
         -------------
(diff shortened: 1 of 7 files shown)


jwst/ami/NRM_Model.py
```diff
@@ -11,13 +18,15 @@
 # - updated May 2013 to include hexagonal envelope
 # - updated (hard refactored) Oct-Nov 2014 Anand S.
 
-from __future__ import absolute_import, division
+from __future__ import (absolute_import, division)
```
hbushouse: Do we still need this now that we're no longer supporting py2.7? Are there places in the remaining code that aren't yet updated to py3.5?

jwst/ami/NRM_Model.py
```diff
@@ -43,73 +52,99 @@ def __init__(self, mask=None, holeshape="circ",
...
+            self.d = 0.80
+            self.D = 6.5
+        else:
+            try:
+                log.debug('mask.ctrs:%s', mask.ctrs)
+            except AttributeError:
+                raise AttributeError("mask must be either 'jwst' \
+                                                      or NRM_mask_geometry object")
 
         log.info('NRM_Model: ctrs flipped in init for CV1, CV2')
```
hbushouse: This should probably be a debug level message, rather than info. Don't need to see it during every normal processing run.

jwst/ami/NRM_Model.py
```diff
@@ -160,48 +202,53 @@ def simulate(self, fov=None, bandpass=None, pixel=None, rotate=False,
...
-                        self.fov_sim, self.over, self.rotctrs, self.d, l, \
-                        self.phi, centering=self.centering, \
-                        shape=self.holeshape)
+        if hasattr(self.bandpass, '__iter__'):
+            log.debug("------Simulating Polychromatic------")
+            self.psf_over = np.zeros((self.over*self.fov_sim,
+                                      self.over*self.fov_sim))
+            for w,l in self.bandpass: # w: weight, l: lambda (wavelength)
+                self.psf_over += w*analyticnrm2.PSF(self.pixel_sim, \
+                            self.fov_sim, self.over, self.rotctrs, self.d, l, \
```
hbushouse: Don't need the \ at the ends of these lines. The implied continuation within the parentheses already takes care of the continuation after the commas.

jwst/ami/NRM_Model.py
```diff
@@ -160,48 +202,53 @@ def simulate(self, fov=None, bandpass=None, pixel=None, rotate=False,
...
+
+        # The monochromatic case if bandpass input is a single wavelength
+        else:
+            self.lam = bandpass
 
-        log.debug("BINNING UP TO PIXEL SCALE")
+            log.debug("Calculating Oversampled PSF")
+            self.psf_over = analyticnrm2.PSF(self.pixel_sim, self.fov_sim,
+                            self.over, self.rotctrs, self.d, self.lam, \
+                            self.phi, centering=centering, \
```
hbushouse: Don't need \

jwst/ami/NRM_Model.py
```diff
@@ -160,48 +202,53 @@ def simulate(self, fov=None, bandpass=None, pixel=None, rotate=False,
...
         self.psf = utils.rebin(self.psf_over, (self.over, self.over))
-        self.psf = self.psf[::-1, :]
 
         return self.psf
 
 
-    def make_model(self, fov=None, bandpass=None, over=False,
-                    centering='PIXELCENTERED', pixscale=None, rotate=False):
+    def make_model(self, fov=None, bandpass=None, over=False, \
+                    centering='PIXELCENTERED', pixweight=None, pixscale=None, \
```
hbushouse: Don't need \

jwst/ami/ami_analyze.py
```diff
@@ -1,17 +1,18 @@
-from __future__ import absolute_import, division
-
 #
 #  Module for applying the LG algorithm to an AMI exposure
 #
 
+from __future__ import (absolute_import, division)
```
hbushouse: Do we still need these for py2.7 compatibility?

hbushouse:
Fixes #1142

dmggh:
Yes, we got some data from them which I have used to verify that their code and ours give the same results.  I will include it in a new regression test.

PR #1767, opened 2018-02-26
Straylight na ns

jemorrison:
The straylight algorithm uses the gap pixels (pixels between the slices) to determine a straylight correction. This pull request removes checking if these gap pixels have the DO_NOT_USE dq_flag set.
This flag will often be set for the slice pixels - so we need to not use that flags as to whether they should be used in this algorithm


jemorrison:
@hbushouse could you check this travis failure - I am not sure if this is a known issue and it is ok to merge or if it is a real error.

hbushouse:
Looks like it's due to the CRDS server being down. I haven't been able to get to it directly (https://jwst-crds.stsci.edu) all morning. Let's wait a bit to see if it comes up and allows the tests to run.

Issue #1619, opened 2017-12-21
Task: tso_photometry updates for B7.2


hbushouse:
The tso_photometry task will need some updates to finalize it for B7.2. This includes settling the way the source location (aperture center) is determined for each different type of TSO imaging data, the aperture size(s) to be used, whether or not to include background aperture(s), etc. Once SDP implements time stamps for each integration in the GROUP table extension carried along with the data products, it should also be updated to use those time stamps, rather than computing the time of each integration on its own (#1617).

hbushouse:
JCCWG page for TSO photometry (https://outerspace.stsci.edu/display/JWSTCC/Vanilla+TSO+Photometry+Extraction) now includes information from the presentation on [2016-09-29](https://outerspace.stsci.edu/display/JWSTCC/2016-09-20+Meeting+notes) that lists the following photometry aperture radii info, at least for NIRCam TS imaging mode:
```
weak-lens: source = 50 pix, sky = 60-70 pix
focused: source = 3 pix, sky = 4-5 pix
```
where a `weak-lens` exposure is identified by a FILTER or PUPIL value of WL3, WLP4, WLM8, WLP8 and all others are assumed to be `focused`.

hbushouse:
Subarrays allowed for NIRCam TS imaging are:
```
FULL
SUB64P
SUB160P
SUB400P
```

hbushouse:
The MIRI imaging template in APT now allows for selecting the "time series observation" requirement, which allows for subarrays:
```
FULL
BRIGHTSKY
SUB64
SUB128
SUB256
```
But there's no information in the MIRI JDox pages yet about characteristics of the imaging TSO mode.

hbushouse:
This is high priority for B7.2, because it is needed to fulfill level 4 requirement DMS-784.

larrybradley:
Outstanding issues:

- [ ] Get MIRI aperture definitions
- [ ] Move instrument-dependent aperture definitions to a reference file
- [ ] Handle case when NIRCam SUB64 (64x64) subarray is used with WLP8 (need direction from TSO WG)

hbushouse:
The new integration times table extension format and contents is described in https://jira.stsci.edu/browse/JSDP-387

skendrew:
[...]

PR #1769, opened 2018-02-27
Use CRPIX values as the target location for TSO imaging

larrybradley:
Addresses part of #1619.  Still need to get MIRI aperture parameters.

### DIFF
--- a/jwst/tso_photometry/tso_photometry_step.py
+++ b/jwst/tso_photometry/tso_photometry_step.py
@@ -24,20 +24,19 @@ class TSOPhotometryStep(Step):
 
     def process(self, input):
         with CubeModel(input) as model:
-            # TODO:  need information about the actual source position in
-            # TSO imaging mode (for all subarrays).
-            # Meanwhile, this is a placeholder representing the geometric
-            # center of the image.
-            nint, ny, nx = model.data.shape
-            xcenter = (ny - 1) / 2.
-            ycenter = (ny - 1) / 2.
+            xcenter = model.meta.wcsinfo.crpix1 - 1    # 1-based origin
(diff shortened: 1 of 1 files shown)


jwst/tso_photometry/tso_photometry_step.py
```diff
@@ -24,20 +24,19 @@ class TSOPhotometryStep(Step):
...
-            xcenter = (ny - 1) / 2.
-            ycenter = (ny - 1) / 2.
+            xcenter = model.meta.wcsinfo.crpix1 - 1    # 1-based origin
+            ycenter = model.meta.wcsinfo.crpix2 - 1    # 1-based origin
 
+            # these aperture definitions are for NIRCam
+            # TODO:  need MIRI aperture definitions
             # all radii are in pixel units
             if model.meta.instrument.pupil == 'WLP8':
+                # weak-lens exposure
```
hbushouse: We should probably consider moving these to a reference file at some point, rather than hard-coded.

jwst/tso_photometry/tso_photometry_step.py
```diff
@@ -24,20 +24,19 @@ class TSOPhotometryStep(Step):
...
                 radius = 50
```
hbushouse: Be aware that SUB64 (64x64 image) is an allowed subarray size for NIRCam TSImage and the APT does NOT stop the user from selecting the WLP8 with that subarray, in which case these radii are going to fall outside the limits of the image. The TSO WG is currently coming to consensus on what should be done in that case (e.g. use the entire image, with no background region, use a much smaller aperture that only covers the PSF core, ...)

larrybradley:
The Travis-CI error is unrelated.

Issue #1771, opened 2018-02-28
outlier detection step fails when sky subtraction not done


stsci-hack:
Nightly regression testing uncovered a use-case that was not covered when updating how outlier_detection uses background values from the sky subtraction step (submitted as PR #1761 ).  

Last night's reg tests showed the coron3 and tso3 pipelines dying in outlier_detection on the line:
`
387:  ta = np.sqrt(np.abs(blot_data + subtracted_background) + err_data ** 2)
`
apparently due to the fact that 'subtracted_background' is None.

PR #1772, opened 2018-02-28
Fix problem with no background values

stsci-hack:
This should address the problem outlined in issue #1771 .


hbushouse:
Travis failure is unrelated issue with sphinx docs. Merging.

jdavies-st:
Resolves #1771

Issue #1732, opened 2018-02-09
NIRSpec slit positions and wavelength assignments


hbushouse:
The NIRSpec team has compared wavelength assignments and extracted slit positions computed by the IDT pipeline and instrument model to what the cal pipeline (assign_wcs and extract_2d) are computing. There are small differences showing up in both slit positions and wavelength assignments, at least for some modes.

This issue serves as a high-level tracker for all work related to fixing these differences.

INS is tracking in multiple tickets: https://jira.stsci.edu/browse/JP-159, https://jira.stsci.edu/browse/JP-238, and https://jira.stsci.edu/browse/JP-241

Latest results are in notebooks [here](https://github.com/nden/nrs_testing).

pllim:
Just curious, what does :fire: label mean? >_>

hbushouse:
The issue is so hot, it's on fire!

nden:
The latest results show agreement within machine precision for fixed slits, MOS data and IFU with gratings. IFU with prism results shows some differences at short wavelengths. Worked on the prism is tracked in #1998.

nden:
This issue was fixed in #1773, #1831 and #1959

PR #1773, opened 2018-02-28
use 1-based cooridnates with the NRS WCS

nden:
The Nirspec WCS expects 1-based coordinates.

Related to #1732.


nden:
unrelated doc build error

Issue #1763, opened 2018-02-22
Update core schema for PATTTYPE allowed values


hbushouse:
Mike Swam reports that processing of a WFSC exposure from OTIS testing yielded:
```
Setting pointing for /ifs/archive/dev/jwst/info/mswam/auto/test_mswam_20180222_02/regr_10006/pipe/sdp/fits/jw00328001001_02105_00001_nrcalong_uncal.fits 
/ifs/pipeline/dms/rhel7/pkgs/miniconda3/envs/jwstdp-0.9.1-py35/lib/python3.5/site-packages/jwst-0.9.1-py3.5-linux-x86_64.egg/jwst/datamodels/util.py:431: ValidationWarning: 'WFSC' is not one of ['CYCLING', 'EXTENDED_TARGET', 'FULL', 'FULL-TIGHT', 'GAUSSIAN', 'INTRAMODULE', 'INTRASCA', 'NONE', 'POINT_SOURCE', 'REULEAUX']
```
So we need to add `WFSC` as an allowed value for PATTTYPE in the core schema.

hbushouse:
Fixed by #1774.

PR #1774, opened 2018-03-01
Update schemas for GRATING and SUBARRAY names

### LINKED ISSUE #1766: Update core schema for allowed GRATING values
DMS CV3 test data has revealed that NIRSpec dark (NRS_DARK) exposures can have null PPS table entries for the grating setting, because the grating is irrelevant and not commanded for darks. This leads to SDP populating the GRATING keyword in level-1b products with the string value 'NULL'. Our core schema doesn't allow for that value and hence throws a validation warning or error. Because this is a common and expected situation for darks, we should add 'NULL' to the list of allowed GRATING values.

hbushouse:
Updated the core schema to include 'NULL' as an allowed GRATING name (fixes #1766), and also include new FGS guide star mode SUBARRAY names.

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -391,7 +391,7 @@ properties:
           grating:
             title: Name of the grating element used
             type: string
-            enum: [G140M, G235M, G395M, G140H, G235H, G395H, PRISM, MIRROR, N/A, ANY]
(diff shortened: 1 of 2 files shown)


hbushouse:
Additional commit to fix #1763, which adds "WFSC" as an allowed value to the PATTTYPE keyword.

hbushouse:
Unrelated doc build error in Travis. Merging.

PR #1775, opened 2018-03-01
Change gain_scale warning to info

### LINKED ISSUE #1657: Modify gain_scale step warnings
https://jira.stsci.edu/browse/JP-207 reports that INS would like the gain_scale step to NOT issue a warning when running it on an exposure from an instrument that doesn't need the step applied.

hbushouse:
Changed the log level from warning to info in the gain_scale step, when it can't find a gain factor keyword in the gain reference file. This happens when processing data from instruments/modes that use the nominal gain and so it isn't really a problem that the keyword is missing for those.

Fixes #1657.

### DIFF
--- a/jwst/gain_scale/gain_scale_step.py
+++ b/jwst/gain_scale/gain_scale_step.py
@@ -26,8 +26,8 @@ class GainScaleStep(Step):
 
                 # Try to read the GAINFACT keyword value
                 if gain_model.meta.gain_factor is None:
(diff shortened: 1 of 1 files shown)


hbushouse:
Unrelated Travis error. Merging.

KevinVolkSTScI:
Changing from a warning to an information message is sufficient, in that it does not flag a possible problem.  So this should be OK and the ticket can be closed.

PR #1786, opened 2018-03-06
updated the nircam specwcs grism mode

sosey:
while making example notebooks for the team I realized the orders member was missing from the nircam wfss specwcs datamodel schema...that's been added and it validates correctly now.


sosey:
This should have no effect on the reference files already in CRDS as they contain the expected orders member

Issue #1647, opened 2018-01-08
calspec3 outlier detection on 1 MIRI LRS fixed-slit exposure


hbushouse:
A calspec3 ASN in the DMS test suite is currently failing in the outlier_detection step. The ASN is:

/ifs/int/jwstc/store/doggett/tests/SICDIL/cstring/JWSTDP-2018_1-171222/2018-01-03-224524/sdp/asn_creation/cal/level3/jw80600-o013_20180103t225333_spec3_001_asn.json

(on the DMS C-string VM, which is accessible as server "owlc-2"), which contains a single MIRI LRS fixed-slit exposure:

`{
    "asn_pool": "jw80600_20180103T225333_pool",
    "asn_rule": "candidate_Asn_MIRI_LRS_FIXEDSLIT",
    "constraints": "Constraints:\n    instrument: {'miri'}\n    is_not_tso: set()\n    asn_candidate: {\"\\\\(\\\\'o013\\\\'\\\\,\\\\ \\\\'observation\\\\'\\\\)\"}\n    opt_elem: {'p750l'}\n    exp_type: {'mir_lrs\\\\-fixedslit'}\n    subarray: {'full'}\n    program: {'80600'}\n    target: {'1'}",
    "target": "t001",
    "asn_id": "o013",
    "products": [
        {
            "members": [
                {
                    "expname": "jw80600013001_02101_00001_mirimage_cal.fits",
                    "asn_candidate": "('o013', 'observation')",
                    "exptype": "science",
                    "exposerr": "null"
                }
            ],
            "name": "jw80600-o013_t001_miri_p750l"
        }
    ],
    "degraded_status": "No known degraded exposures in association.",
    "code_version": "0.9.4",
    "program": "80600",
    "asn_type": "spec3",
    "version_id": "20180103T225333"
}`

The error is:

2018-01-04 08:50:04,721 - stpipe.Spec3Pipeline - INFO - Step Spec3Pipeline running with args ('/ifs/int/jwstc/store/doggett/tests/SICDIL/cstring/JWSTDP-2018_1-171222/2018-01-03-224524/sdp/asn_creation/cal/leve
l3/jw80600-o013_20180103t225333_spec3_001_asn.json',).
2018-01-04 08:50:04,721 - stpipe.Spec3Pipeline - INFO - Starting calwebb_spec3 ...
[...]

PR #1787, opened 2018-03-06
Update logic to correctly flag as skipped

stsci-hack:
This PR should address #1647 , at least based on using this updated code to successfully process the dataset specified in that issue.  

The outlier_detection step, with this change, will no longer run when only provided with a single input dataset.  Instead, it will return a copy of the input and flag the cal_step.outlier_detection as "SKIPPED" in the returned data's metadata.


hbushouse:
Fixes #1647.

Issue #1731, opened 2018-02-09
Enhance outlier_detection documentation


hbushouse:
Currently the sphinx docs for the outlier_detection step mainly just contain reference type material for the interfaces and arguments for the various functions. It also needs to include a high-level description of how it works and what it does, that is geared to the user level (not developer level). A good overall description should be added once all of the B7.2 updates to the code have been implemented, which needs to cover the different ways the data are handled for different observing modes (or input data type), such as regular imaging vs. TSO data cubes vs. coronagraphy data cubes.

hbushouse:
INS tracking in https://jira.stsci.edu/browse/JP-166

hbushouse:
Fixed by #1791

PR #1791, opened 2018-03-07
Update outlier docs

stsci-hack:
Fairly extensive update to the outlier_detection step documentation to try to make it a little more user-friendly (read that: human readable as opposed to strictly describing the API).  

This should address many of the points raised in #1731 .

### DIFF
--- a/docs/jwst/outlier_detection/index.rst
+++ b/docs/jwst/outlier_detection/index.rst
@@ -2,6 +2,52 @@
 Outlier Detection
 =================
 
+Processing multiple datasets together allows for the identification of bad pixels
+or cosmic-rays that remain in each of the input images, many times at levels which
+were ambiguous for detection during ramp fitting.  The outlier detection step 
+implements the following algorithm to identify and flag any remaining cosmic-rays or
+other artifacts left over from previous calibrations:
+
+  - build a stack of input data
+  
+    - all inputs will need to have the same WCS in order
+    - if needed, each input will be resampled to a common output WCS
+  
+  - create a median image from the stack of input data
+
+    - this median operation will ignore any input pixels which have a weight 
+      which is too low (<70% max weight)
+  
+  - create "blotted" data from the median image to exactly match each original
+    input dataset
+  
+  - perform a statistical comparison (pixel-by-pixel) between the median,blotted
+    data with the original input data to look for pixels with values greater 
+    that some sigma based on the noise model
+
+    - the noise model used relies on the error array computed by previous 
+      calibration steps based on the readnoise and calibration errors
+  
+  - flag the DQ array for the input data for any pixel (or affected neighboring
+    pixels) identified as a statistical outlier
+
+The outlier detection step serves as a single interface to apply this general 
+process to any JWST data, with specific variations of this algorithm for each 
+type of data.  Sub-classes of the outlier detection algorithm have been developed
+specifically for 
+
+  - Imaging data
+  - IFU spectroscopic data
+  - TSO data
+  - coronagraphic data
+  - spectroscopic data
+  
+This allows the outlier_detection step to be tuned to the variations in each type 
+of JWST data.
+           
+Outlier Detection Code API
+===========================
+
 .. toctree::
    :maxdepth: 2
 
--- a/docs/jwst/outlier_detection/index.rst
+++ b/docs/jwst/outlier_detection/index.rst
@@ -10,7 +10,10 @@ other artifacts left over from previous calibrations:
 
   - build a stack of input data
   
-    - all inputs will need to have the same WCS in order
+    - all inputs will need to have the same WCS since outlier detection assumes 
+      the same flux for each point on the sky, and variations from one image to
+      the next would indicate a problem with the detector during readout of that
+      pixel
     - if needed, each input will be resampled to a common output WCS
   
   - create a median image from the stack of input data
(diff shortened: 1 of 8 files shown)


docs/jwst/outlier_detection/outlier_detection_ifu.rst
```diff
@@ -1,23 +1,29 @@
...
+ Specifically, this routine performs the following operations (modified from the
+ :ref:`outlier-detection-image`):
 
  * Extract parameter settings from input model and merge them with any user-provided values
 
- * Resamples all input images into :py:class:`~jwst.datamodels.IFUCubeModel` observations.
+   - the same set of parameters available to :ref:`outlier-detection-imaging`
+     also applies to this code
+
+ * Resample all input :py:class:`~jwst.datamodels.ImageModel` images into
```
hbushouse: As of a couple weeks ago the datamodels got reworked such that the 2D form of IFU images are now an IFUImageModel, instead of a plain ImageModel. So the inputs will now be IFUImageModel.

docs/jwst/outlier_detection/outlier_detection_step.rst
```diff
@@ -1,7 +1,7 @@
-.. _outlier_B7_design_:
+.. _outlier_B7_design:
```
hbushouse: Not sure we want to use a tag that refers to a specific DMS build

docs/jwst/outlier_detection/outlier_detection_step.rst
```diff
@@ -28,19 +28,14 @@ input data:
...
-  - **Time-Series Observations(TSO)**: both imaging and spectroscopic modes, will use :py:class:`~jwst.outlier_detection.outlier_detection.OutlierDetection` with resampling turned off
-  - **NIRSpec and MIRI IFU observations**: use :py:class:`~jwst.outlier_detection.outlier_detection_ifu.OutlierDetectionIFU`
-  - **Long-slit spectroscopic observations**: use :py:class:`~jwst.outlier_detection.outlier_detection_spec.OutlierDetectionSpec`
-
-* Read in reference files used by outlier detection, currently:
-
-  - **readnoise**: based on :py:class:`~jwst.datamodels.ReadnoiseModel`
-  - **gain**: based on :py:class:`~jwst.datamodels.GainModel`
+  - **Images**: like those taken with NIRCam, will use :py:class:`~jwst.outlier_detection.outlier_detection.OutlierDetection` as described in :ref:`outlier-detection-imaging`
+  - **Coronagraphic observations**: use :py:class:`~jwst.outlier_detection.outlier_detection.OutlierDetection` with resampling turned off in :ref:`outlier-detection-imaging`
```
hbushouse: add the words "as described" between "turned off" and "in"

docs/jwst/outlier_detection/outlier_detection_step.rst
```diff
@@ -28,19 +28,14 @@ input data:
...
+  - **Time-Series Observations(TSO)**: both imaging and spectroscopic modes, will use :py:class:`~jwst.outlier_detection.outlier_detection.OutlierDetection` with resampling turned off in :ref:`outlier-detection-imaging`
```
hbushouse: add the words "as described" between "turned off" and "in"

hbushouse:
Hard to tell what all is described by just looking at the diffs, but I'm thinking it would be good to include some kind of really generic paragraph somewhere near the top/front that tells folks that outliers are found by creating a median from multiple inputs and then comparing each input against that median, and then the details for each type of data differ in the way the median is created and what constitutes multiple input images (e.g. for regular imaging the inputs are all separate 2D images that get combined to form a median and then comparison, while for coron and TSO the individual inputs are 3D data cubes where each plane of the cube is used to form the median and do comparison, etc.).

stsci-hack:
This latest commit should start to address your comments.  Additional documentation still needs to be added to more document specific variations in how the code works (as raised in your comments).

stsci-hack:
Further documentation of outlier detection for spectroscopic modes will need to wait until the implementation of support for those modes has been finalized and will be provided through a separate PR.

docs/jwst/outlier_detection/index.rst
```diff
@@ -2,6 +2,52 @@
...
 
+Processing multiple datasets together allows for the identification of bad pixels
+or cosmic-rays that remain in each of the input images, many times at levels which
+were ambiguous for detection during ramp fitting.  The outlier detection step 
+implements the following algorithm to identify and flag any remaining cosmic-rays or
+other artifacts left over from previous calibrations:
+
+  - build a stack of input data
+  
+    - all inputs will need to have the same WCS in order
```
hbushouse: Not sure what the phraseology "the same WCS in order" means? Or did a typo happen?

docs/jwst/outlier_detection/index.rst
```diff
@@ -2,6 +2,52 @@
...
+  - create a median image from the stack of input data
+
+    - this median operation will ignore any input pixels which have a weight 
+      which is too low (<70% max weight)
+  
+  - create "blotted" data from the median image to exactly match each original
+    input dataset
+  
+  - perform a statistical comparison (pixel-by-pixel) between the median,blotted
+    data with the original input data to look for pixels with values greater
```
hbushouse: "greater" implies we're only looking for and flagging positive outliers. Is that true or does it also check for negative outliers?

stscicrawford:
@stsci-hack When you have a chance, can you either update #1731 with what further work is needed or close that issue and create a new one for the spectroscopic updates that will be required?

PR #1797, opened 2018-03-09
Add TUNIT to table-based products

hbushouse:
#1754 provided an update that allows column attributes, like units, to be set for table-based data models that have been created from scratch within software. This takes advantage of that ability by setting column units values for the spec_table's created in the extract_1d step, which are stored in x1d products.

It also includes an update to the photom step to set the units value for wavelength column in the RELSENS tables that are attached to science products.

Those are all the FITS table-based science products that I can think of that should set column units. Can anyone think of any others?  The source catalog, TSO photometry, and TSO white-light tables are all ASDF-based and already specify column units.

### DIFF
--- a/jwst/extract_1d/ifu.py
+++ b/jwst/extract_1d/ifu.py
@@ -81,9 +81,11 @@ def ifu_extract1d(input_model, refname, source_type):
         flux = net.copy()
         net[:] = 0.
         log.info("Data have been flux calibrated; setting net to 0.")
+        data_units = 'mJy'
     else:
         flux = np.zeros_like(net)
         log.info("Data have NOT been flux calibrated; setting flux to 0.")
+        data_units = 'mJy'
 
     fl_error = np.ones_like(net)
     nerror = np.ones_like(net)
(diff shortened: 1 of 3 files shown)


hbushouse:
Fixes #473.

jwst/extract_1d/ifu.py
```diff
@@ -81,9 +81,11 @@ def ifu_extract1d(input_model, refname, source_type):
         flux = net.copy()
         net[:] = 0.
         log.info("Data have been flux calibrated; setting net to 0.")
+        data_units = 'mJy'
     else:
         flux = np.zeros_like(net)
         log.info("Data have NOT been flux calibrated; setting flux to 0.")
+        data_units = 'mJy'
```
philhodge: Shouldn't the units be 'DN/s' for this case (not flux calibrated)?

hbushouse: Of course. Copy-n-paste error. Good eye!

philhodge:
The changes look good to me.

hbushouse:
@philhodge The 2D and 3D inputs, after they've had flux calibration applied, are in units of surface brightness, such as mJy/arcsec^2 for IFU data. Is there any kind of normalization that's done to the extracted spectrum to account for the fact that the values at each wavelength have now been summed over some area and hence it's correct that we're claiming they're now in units of mJy?

philhodge:
@hbushouse  Not that I know of; it certainly isn't done in `extract_1d`.  Maybe @jemorrison did this in `cube_build`.  I presume you mean multiplying by the solid angle of a pixel?  Could that vary from pixel to pixel, due to distortion in the projection onto the detector?  Or would that variation be taken out by the `flat_field` step?

PR #1798, opened 2018-03-09
Initial model_blender sphinx docs

stsci-hack:
Initial cut at adding user documentation for the model_blender package.  Adding a label to some datamodels docs made it easy to reference the datamodels documentation from within the model_blender (or any JWST) package.  I also needed to update the package index for JWST pipeline to include the model blender docs.

### DIFF
--- a/docs/jwst/model_blender/blender_rules.rst
+++ b/docs/jwst/model_blender/blender_rules.rst
@@ -1,0 +1,38 @@
+.. _blender_rules:
+
+Model Blender Rules
+====================
+
+Blending models relies on rules to define how to evaluate all the input values 
+for a model attribute in order to determine the final output value. These rules
+then get specified in the model schema for each attribute.  
+
+The rules get interpreted and applied as list or array operations which work on 
+the set of input values for each attribute.  The full set of pre-defined rules
+includes:
+
+.. code-block:: python
+
+  # translation dictionary for function entries from rules files
+  blender_funcs = {'first': first,
+                 'last': last,
+                 'float_one': float_one,
+                 'int_one': int_one,
+                 'zero': zero,
+                 'multi': multi,
+                 'multi?': multi1,
+                 'mean': np.mean,
+                 'sum': np.sum,
+                 'max': np.max,
+                 'min': np.min,
+                 'stddev': np.std}
+
+The rules which should be referenced in the model schema definition are the 
+keys defined for `jwst.model_blender.blender_rules.blender_funcs` listed 
+above.  This definition illustrates how several rules are simply interfaces for 
+numpy array operations, while others are defined internally to `model_blender`.  
+
+.. toctree::
+   :maxdepth: 2
+
+.. automodapi:: jwst.model_blender.blendrules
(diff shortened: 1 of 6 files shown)


docs/jwst/model_blender/blender_rules.rst
```diff
@@ -0,0 +1,38 @@
+.. _blender_rules:
+
+Model Blender Rules
+====================
+
+Blending models relies on rules to define how to evaluate all the input values 
+for a model attribute in order to determine the final output value. These rules
+then get specified in the model schema for each attribute.  
+
+The rules get interpreted and applied as list or array operations which work on
```
hbushouse: operations **that** work on

docs/jwst/model_blender/blender_rules.rst
```diff
@@ -0,0 +1,38 @@
...
+                 'zero': zero,
+                 'multi': multi,
+                 'multi?': multi1,
+                 'mean': np.mean,
+                 'sum': np.sum,
+                 'max': np.max,
+                 'min': np.min,
+                 'stddev': np.std}
+
+The rules which should be referenced in the model schema definition are the
```
hbushouse: The rules **that** should be ...

docs/jwst/model_blender/main.rst
```diff
@@ -0,0 +1,105 @@
...
+  - updating the output model's metadata with the new values
+  - generating the output table with one row for each input model's values
+  
+  
+Using model_blender
+===================
+The model blender package requires
+
+  - all the input models (or FITS files) be available
+  - the output product has already been generated
```
hbushouse: Does the output product have to already exist in the form of a FITS file, or simply a data model in memory? I believe it's the latter, right?

docs/jwst/model_blender/main.rst
```diff
@@ -0,0 +1,105 @@
...
+The simplest way to run model blender only requires calling a single interface:
+
+.. code-block:: python
+  
+  from jwst.model_blender import blendmeta
+  blendmeta.blendmodels(product, inputs=input_list)
+  
+where 
+
+  - `product`: the datamodel(or FITS filename) for the already combined product
```
hbushouse: Missing a whitespace between "datamodel" and "("

docs/jwst/model_blender/main.rst
```diff
@@ -0,0 +1,105 @@
...
+
+  - minimum
+  - maximum
+  - first
+  - last
+  - mean
+  - zero
+  
+These can then be used to customize the output value for any given attribute
+should the default value provided by default with the schema installed with the
```
hbushouse: too many instances of "default" in this phrase

docs/jwst/model_blender/main.rst
stsci-hack: Actually, the code has been written to support both options for both inputs and output products.  As you can see in the commit, the documentation states that working directly with the datamodels is preferred and points out the I/O efficiencies as the primary advantage.  However, anyone wanting to run this code on data already processed and written out to disk will still be able to blend their data as a stand-alone, file-based operation.

PR #1799, opened 2018-03-11
Move the docs to the release stsci_rtd_theme instead of the local hack

sosey:
This moves the documentation to use the released stsci_rtd_theme instead of the local hack that was in place. I've tested the same thing on the stsci-package-template and locally, so this should work :crossed_fingers:


hbushouse:
Is this ready to merge?

sosey:
should be, just left the tests to run and didn't get back to it :)

Issue #1810, opened 2018-03-15
Data models support for GAINFACT / meta.exposure.gain_factor


jaytmiller:
There have been requests on Jira CRDS-120 and CRDS-175 for additional checking of the gain factor used in the gain_scale step.   There's a desire to document it in general for:

    Dark???           W
    Bias???            W
    Saturation         W
    linearity             W
    Gain                  R
    Readnoise        W

where it is really only functionally required for the gain_scale step.  

It is currently a scalar value in the continuous range between 1.0 and 10.0 which is checked independently in CRDS.  W corresponds to a CRDS warning when absent.  R (required) corresponds to a CRDS error when absent.

To properly support the creation of reference files,  three things need to happen:

1.  GAINFACT needs to be added to additional reference file schema.
2.  The reference file DM structure needs to change from meta.gain_factor to meta.exposure.gain_factor to mirror the core schema used for datasets.
3.  The gain_scale step needs to be tweaked to the new DM structure.

This primarily makes the CAL code data models responsible for *naming* the parameter and mapping it onto FITS.  It remains optional in CAL and unconstrained.

hbushouse:
Just to be explicit, the reason for adding the GAINFACT keyword definition to reference file data models that don't actually require it (dark, superbias, linearity, ...) is because ref files creators are being encouraged to include that keyword in those ref files as information/documentation and if they use our datamodels to create the ref files (which we're also trying to encourage) they need to have the keyword defined in the datamodel in order to get it to appear in the FITS file.

hbushouse:
Fixed by #1811.

PR #1811, opened 2018-03-15
Update reference file gain_factor schema,  #1810

jaytmiller:
Schema and step change to add additional support for GAINFACT and to change the data models path for reference files from meta.gain_factor to meta.exposure.gain_factor.  See issue #1810.

### DIFF
--- a/jwst/datamodels/schemas/keyword_gainfact.schema.yaml
+++ b/jwst/datamodels/schemas/keyword_gainfact.schema.yaml
@@ -7,7 +7,7 @@ properties:
             type: object
             properties:
                   gain_factor:
-                  title: Gain scale factor
-                  type: number
-                  fits_keyword: GAINFACT
-
+                         type: number
+                         properties:
+                            title: Gain scale factor
+                            fits_keyword: GAINFACT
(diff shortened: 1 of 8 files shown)


jaytmiller:
Nice catch,  thanks!

jaytmiller:
Since CRDS use cases and tests don't instantiate GainModel,  there was an undetected issue with the gainfact keyword YAML structure.   I was able to interactively instantiate a GainModel() so hopefully this fixes the schema issue.

jwst/datamodels/schemas/keyword_gainfact.schema.yaml
```diff
@@ -7,7 +7,7 @@ properties:
             type: object
             properties:
                   gain_factor:
-                  title: Gain scale factor
-                  type: number
-                  fits_keyword: GAINFACT
-
+                         type: number
+                         properties:
```
hbushouse: You shouldn't need this additional "properties" entry. Just make sure the 3 properties "title", "type", and "fits_keyword" are all indented (at the same level) under "gain_factor", e.g.
```
  gain_factor:
    title: Gain scale factor
    type: number
    fits_keyword: GAINFACT
```

PR #1817, opened 2018-03-16
Cubepars

jemorrison:
The MIRI and NIRSPEC teams (via Ben Sargent) have delivered new IFU cube par reference files.
This PR reads in the new format of the reference files and makes the needed changes to the code to
use these values.
Also the final WCS of the IFUCube, PC1_1 = -1 (instead of 1)

### DIFF
--- a/jwst/cube_build/cube_build_io_util.py
+++ b/jwst/cube_build/cube_build_io_util.py
@@ -69,46 +69,117 @@ def read_cubepars(self, instrument_info):
     if self.instrument == 'MIRI':
         ptab = datamodels.MiriIFUCubeParsModel(self.par_filename)
         number_bands = len(self.all_channel)
-
         # pull out the channels and subcahnnels that cover the data making up the cube
         for i in range(number_bands):
             this_channel = self.all_channel[i]
-            compare_channel = 'CH'+this_channel
+            #compare_channel = 'CH'+this_channel
             this_sub = self.all_subchannel[i]
-                # find the table entries for this combination
+            # find the table entries for this combination
             for tabdata in ptab.ifucubepars_table:
                 table_channel = tabdata['channel']
                 table_band = tabdata['band']
-                table_plt_scale = tabdata['PLTSCALE']
-                table_wresol = tabdata['WRESOL']
-                table_sroi = tabdata['SROI']
-                table_wroi = tabdata['WROI']
+                table_spaxelsize = tabdata['SPAXELSIZE']
+                table_spectralstep = tabdata['SPECTRALSTEP']
+                table_wavemin = tabdata['WAVEMIN']
+                table_wavemax = tabdata['WAVEMAX']
                 #match on this_channel and this_sub
-                if(compare_channel == table_channel and this_sub == table_band):
-                    instrument_info.SetSpatialScale(table_plt_scale,this_channel,this_sub)
-                    instrument_info.SetWaveRes(table_wresol,this_channel,this_sub)
-                    instrument_info.SetWaveROI(table_wroi,this_channel,this_sub)
-                    instrument_info.SetSpatialROI(table_sroi,this_channel,this_sub)
+                if(this_channel == table_channel and this_sub == table_band):
+                    instrument_info.SetSpatialSize(table_spaxelsize,this_channel,this_sub)
+                    instrument_info.SetSpectralStep(table_spectralstep,this_channel,this_sub)
+                    instrument_info.SetWaveMin(table_wavemin,this_channel,this_sub)
+                    instrument_info.SetWaveMax(table_wavemax,this_channel,this_sub)
 
+            for tabdata in ptab.ifucubepars_msn_table:
+                table_channel = tabdata['channel']
+                table_band = tabdata['band']
+                table_sroi = tabdata['ROISPATIAL']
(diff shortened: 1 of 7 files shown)


jemorrison:
With this PR we can update the CRDS_CONTEXT to use the new cube par reference files
We probably want to time these so they part of the pipeline at the same time.

jaytmiller:
I can't/won't comment on the technical aspects of your format changes unless you ask  me to attempt it.  

 Strategically,  I have these CRDS comments:

1.  We use setting the CRDS_CONTEXT env var to enable developers to preview references before releasing them for broad use as the default.   So whenever your new code is deployed,  and the references check out,  Howard will typically bless the situation and I will adjust the default on the CRDS server... which is then distributed to those who sync.

2.  Once the pipeline code is updated,  are these new references *required*?   If that's the case,  we should do the context update in lock step and possibly with an announcement of some kind.

3. We should consider what happens with the old references,  if they're to be removed from subsequent CRDS rules,  or if they need to remain available for older versions of code or test cases or are just flat out still needed.

jaytmiller:
I should make the additional comment that most people don't configure CRDS and use the default onsite CRDS cache at /grp/crds/cache.   That cache is automatically synced to the CRDS server so it will automatically deliver the new default context to people who use /grp/crds/cache without knowing or thinking about it.

jwst/cube_build/cube_build_io_util.py
```diff
@@ -69,46 +69,117 @@ def read_cubepars(self, instrument_info):
...
+            table_wave = tabdata['WAVELENGTH']
+            table_sroi = tabdata['ROISPATIAL']
+            table_wroi = tabdata['ROISPECTRAL']
+            table_power = tabdata['POWER']
+            table_softrad = tabdata['SOFTRAD']
+            instrument_info.SetMultiChannelTable(table_wave,table_sroi,
+                                              table_wroi,table_power,
+                                              table_softrad)
+
+        print('Done reading cubepar reference file')
```
hbushouse: Shouldn't use `print` statements. Always use some flavor of `log.xxxx` (info, debug, warning, ...)

jwst/cube_build/cube_build_io_util.py
```diff
@@ -69,46 +69,117 @@ def read_cubepars(self, instrument_info):
...
+        for tabdata in ptab.ifucubepars_high_wavetable:
+            table_wave = tabdata['WAVELENGTH']
+            table_sroi = tabdata['ROISPATIAL']
+            table_wroi = tabdata['ROISPECTRAL']
+            table_power = tabdata['POWER']
+            table_softrad = tabdata['SOFTRAD']
+            instrument_info.SetHighTable(table_wave,table_sroi,
+                                          table_wroi,table_power,
+                                          table_softrad)
+        print('Done reading cubepar reference file')
```
hbushouse: Use `log` instead of `print`

jwst/cube_build/ifu_cube.py
```diff
@@ -447,10 +444,13 @@ def determine_roi_size(self):
 
             for i in range(number_gratings):
                 this_gwa = self.list_par1[i]
-                wroi = self.instrument_info.GetWaveRoi(this_gwa)
+                this_filter = self.list_par2[i]
+                print('Grating and Filter',this_gwa,this_filter)
```
hbushouse: Use `log` instead of `print`

jwst/cube_build/ifu_cube.py
```diff
@@ -714,7 +714,7 @@ def setup_IFUCube(self,j):
         IFUCube.meta.wcsinfo.ctype3 = 'WAVE'
         IFUCube.meta.wcsinfo.cunit3 = 'um'
         IFUCube.meta.wcsinfo.wcsaxes = 3
-        IFUCube.meta.wcsinfo.pc1_1 = 1
+        IFUCube.meta.wcsinfo.pc1_1 = -1
```
hbushouse: The values of exactly 0 and +-1 for the PC matrix terms suggest that the data are oriented with North up. Are the cube data in fact always rotated so that North is up? If not, the PC matrix terms need to reflect the orientation on the sky.

hbushouse:
@jaytmiller I'm guessing that the new code requires the new cubepars ref files. I don't know if old versions of the code can read (the old portions) of the new ref files. Hence I'm not sure about the synchronizing of updating the default context with the merge of this PR. Users would only need the new ref files once they update their code environment. So it might be safest for a short period to NOT update the default context while we let users know that a new code version is available and if they install it they'll also need to (manually) use the new context. After everyone has had a chance to update their code, we can then update the default context.

The old ref files should stay available for use with old versions of the code (i.e. do not completely replace the old ref files with the new ones).

jwst/cube_build/ifu_cube.py
jemorrison: On a few IFUcubes (in the Northern and Southern hemisphere)  I did the simple check about looking at the same column of data  in the IFUCube and verifying that ra did not change and then the same row in the IFUcube to verify the dec did not change.  In the current implementation North is always up. I will add to an issue that in the future we probably want to do what resample for images does and have an option to have the IFUCube rotated on the sky from North more in alignment with the actual footprint.

Issue #1814, opened 2018-03-16
outlier_detection fails in calwebb_spec3 on ifu data


hbushouse:
Running the calwebb_spec3 pipeline on an association of MIRI MRS (IFU) exposures results in the following error near the end of the outlier_detection step:
```
2018-03-15 18:08:10,833 - stpipe.Spec3Pipeline.outlier_detection - INFO - Time Create Single IFUcube  = 48.5.s
2018-03-15 18:08:10,930 - stpipe.CubeBuildStep - INFO - Number of IFUCube models returned from building single IFUCubes 3 
----------------------------------------------------------------------
ERROR RUNNING STEP 'Spec3Pipeline':
    No attribute 'rfind'
Traceback (most recent call last):
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python3.5/site-packages/jwst-0.9.2a0.dev82-py3.5-macosx-10.7-x86_64.egg/jwst/datamodels/properties.py", line 204, in __getattr__
    val = self._instance[attr]
KeyError: 'rfind'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python3.5/site-packages/jwst-0.9.2a0.dev82-py3.5-macosx-10.7-x86_64.egg/jwst/stpipe/cmdline.py", line 297, in step_from_cmdline
    step.run(*positional)
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python3.5/site-packages/jwst-0.9.2a0.dev82-py3.5-macosx-10.7-x86_64.egg/jwst/stpipe/step.py", line 405, in run
    step_result = self.process(*args)
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python3.5/site-packages/jwst-0.9.2a0.dev82-py3.5-macosx-10.7-x86_64.egg/jwst/pipeline/calwebb_spec3.py", line 120, in process
    result = self.outlier_detection(result)
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python3.5/site-packages/jwst-0.9.2a0.dev82-py3.5-macosx-10.7-x86_64.egg/jwst/stpipe/step.py", line 405, in run
    step_result = self.process(*args)
[...]

PR #1818, opened 2018-03-16
Fix output path naming

stsci-hack:
This is my interpretation of how to call the 'make_output_path' method for outlier_detection_ifu based on the usage from other modules in the package.  At the very least, this avoids the original error reported in Issue #1814  based on my interactive testing with the original problem dataset.

### DIFF
--- a/jwst/outlier_detection/outlier_detection_ifu.py
+++ b/jwst/outlier_detection/outlier_detection_ifu.py
@@ -108,7 +108,8 @@ class OutlierDetectionIFU(OutlierDetection):
 
             for model in single_IFUCube_result:
                 model.meta.filename = self.make_output_path(
-                    model, suffix=self.resample_suffix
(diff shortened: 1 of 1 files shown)


jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -119,7 +120,8 @@ def do_detection(self):
                             init=single_IFUCube_result[0].data.shape)
             median_model.meta = single_IFUCube_result[0].meta
             median_model.meta.filename = self.make_output_path(
-                self.input_models[0], suffix='ch{}_media'.format(ch)
+                basepath=self.input_models[0].meta.filename, 
+                suffix='ch{}_media'.format(ch)
```
hbushouse: Was that suffix string supposed to be 'ch{}_media**n**'?

Issue #1704, opened 2018-02-02
The guider pipeline may need to accept 3-D input


philhodge:
The input to `GuiderPipeline` is normally a 4-D array, an uncalibrated ramp image.  Shannon Osborne tried running this with the first two steps set to be skipped, correctly passing a 3-D slope image as input, in order to run the third step, flat_field.  The pipeline failed because currently the input file is explicitly opened as a `GuiderRawModel`, which must be 4-D.  The guider pipeline should be modified to accept a 3-D slope image if the first two steps are to be skipped.

hbushouse:
See https://stsci-ins.basecamphq.com/projects/11477312-jwst-pipeline/posts/104610817/comments for the original reporting of the problem.

INS is tracking in https://jira.stsci.edu/browse/JP-212

hbushouse:
Fixed by #1821

Issue #1612, opened 2017-12-21
Task: FGS Guider pipeline updates


hbushouse:
Make all necessary updates to the FGS guide star pipeline, including the data models and calibration steps, to be compatible with SDP B7.1.1 updates to the level-1b guide star data products. This will include any updates to the schema definitions of the various table extensions that are carried along in GS products, as well as any updates that might be necessary to fix issues discovered from testing of soon-to-be delivered MASK and FLAT reference files for the ID-STACK mode.

hbushouse:
A B7.1 tester reported in https://stsci-ins.basecamphq.com/projects/11477312-jwst-pipeline/posts/104547619/comments that products coming out of the guider_pipeline do not have an ERR extension. This was actually by design in the guider-specific data models, based on no ERR extension being specified in the guider pipeline design document. We should rethink whether we want to include an ERR extension, at least for consistency with other science products.

hbushouse:
This is a high priority task for DMS Build 7.2, in order to satisfy level 4 requirements DMS-189 and DMS-692.

hbushouse:
So far all testing has revealed only 1 minor change needed to the guider data models in order to match the SDP uncal files, which is the format of the `reference_star_id` column in the `Planned Reference Stars` table extension. The SDP level-1b files have the column format as A12, while our data models had it as A10.

The latest FGS MASK and FLAT reference files in CRDS that were tailored for use with ID `stacked` mode exposures are being selected and applied properly.

hbushouse:
Fixed by #1821, at least until any other issues are discovered via OTB testing.

PR #1821, opened 2018-03-21
Updates to calwebb_guider processing

hbushouse:
First set of updates to guide star processing for #1612. The two main changes are the addition of an ERR array to the data models and a minor update to the column format of one of the table extensions (in order to match what SDP is generating). The guider_cds step has been updated to create an ERR array in the output model. Note that nothing is done yet with the ERR array at any stage of guider processing. It's just there for future expansion and to make it consistent with the SCI, DQ, ERR contents of science products.

Many other changes were made just for general cleanup and to get pep8/flake8 off my back. All of the changes to dq_init are for this purpose only, with no actual changes to the processing.

### DIFF
--- a/jwst/datamodels/guidercal.py
+++ b/jwst/datamodels/guidercal.py
@@ -19,6 +19,9 @@ class GuiderCalModel( model_base.DataModel):
     dq: numpy array
         The data quality array. 2-D
 
+    err: numpy array
+        The error array. 3-D
+
     plan_star_table: table
         The planned reference star table
 
@@ -37,7 +40,7 @@ class GuiderCalModel( model_base.DataModel):
 
     schema_url = "guider_cal.schema.yaml"
 
-    def __init__(self, init=None, data=None, dq=None,
+    def __init__(self, init=None, data=None, dq=None, err=None,
                  plan_star_table=None, flight_star_table=None,
                  pointing_table=None, centroid_table=None,
                  track_sub_table=None, **kwargs):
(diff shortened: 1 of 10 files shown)


dmggh:
In guider_cal.schema.yaml, isn't 'ndim: 3' needed in :   

err:
      title: Error array
      fits_hdu: ERR
      default: 0.0
      datatype: float32

and in guider_raw.schema.yaml, shouldn't there be 'ndim: 4' ?

hbushouse:
You're right that the ERR array is 4D in the raw and 3D in the cal, and it wouldn't hurt to explicitly specify that in the schema, but it's not required in order to get them to come out that way. In the raw model the ERR array will always be created by the data model infrastructure to have the same dimensions as the SCI array (which is 4D), and for the cal model we explicitly create the ERR array in the guider_cds step using the same dimensions as the SCI array (which is 3D).

hbushouse:
Latest commit fixes #1704 by loading the input into a GuiderCalModel if the dq_init and guider_cds steps are skipped. It makes the very implicit assumption that the input has been run through those steps previously, without actually verifying that. So it could be more robust, but good enough for now to handle this relatively obscure corner case (IMHO).

dmggh:
This looks good to me. It might be helpful to also issue a log info statement that those steps are being skipped.

hbushouse:
The steps themselves will send that log message. The skip operation is a bit unusual in that the step still gets called, but then immediately issues a message saying that it's being skipped and returns the input unchanged.

dmggh:
Yes, I realize that the individual steps would each give that message, which should be apparent to the user. I was just suggesting an additional statement making it blatantly (over-) obvious to the casual user what was happening, in case that was not what they were expecting.

hbushouse:
Latest commit applies the flat to the ERR array, now that we've implemented an ERR array in the guide star data models.

hbushouse:
These are all of the changes needed in the near term, so merging now.

PR #1832, opened 2018-03-26
WIP: enabled master background subtraction for team to test wfss

sosey:
The only real change right now is that I uncommented the lines in the step that avoided the correction for WFSS modes.  I emailed @bhilbert4 to check it out with his test reference files and I'll make updates to this branch as needed.

see: https://jira.stsci.edu/browse/CRDS-181

### DIFF
--- a/jwst/background/background_step.py
+++ b/jwst/background/background_step.py
@@ -14,8 +14,7 @@ class BackgroundStep(Step):
     """
 
     # These reference files are only used for WFSS/GRISM data.
-    # xxx can't be used yet:
-    # xxx reference_file_types = ["wfssbkg", "wavelengthrange"]
+    reference_file_types = ["wfssbkg", "wavelengthrange"]
 
     def process(self, input, bkg_list):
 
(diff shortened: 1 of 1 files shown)


sosey:
@hbushouse I don't think they are needed, the logic that catches WFSS modes looks for the reference files specifically and the other modes have a list of background exposure filenames passed to the step in `bkg_list`, they aren't using delivered reference files. 

 @bhilbert4, since there are no delivered files, it's probably easiest to call this function (https://github.com/STScI-JWST/jwst/pull/1832/files#diff-3bf7670f3a429c3d11221aec7a7315a8R55) directly with the two filenames as you are testing...

sosey:
@hbushouse apologies, I totally overlooked the fact that the base class from `stpipe` used the `reference_file_types` to override `get_reference_files`.... fixing :-)

sosey:
@hbushouse sounds good :smile:

sosey:
closing this because the updates were merged in https://github.com/spacetelescope/jwst/pull/1918

PR #1836, opened 2018-03-27
Fix resample output frame offset

### LINKED ISSUE #1816: wcs_from_footprints miscalculating output frame
Testing of the Image3 Pipeline has uncovered an apparent problem with the computation of the WCS for the output frame.  For observations which are dithered/mosaiced (generally offset from one another), the output data seems to be offset from center by the same amount as the offset between the observations.  Fortunately, this does not seem to be affected by any corrections applied by the 'tweakreg' step.

Test data has been made available on central store in:
    /user/hack/data/jwst/Image3Test/

The test data consists of 4 frames at one pointing (*1101_A[1-4]) and 4 frames at another pointing (*1105_A[1-4]).  This test data includes a sample output which demonstrates this problem.  The test can be run using the included .cfg file and _asn.json file which performs a basic resampling to create the mosaic without outlier_detection or alignment.  

The problem was originally reported under the 0.9.0 environment, but it was reproduced using the latest dev version of the code.  The script included in this test directory, reset_gwcs.py, was used to reset the GWCS ASDF object in the input files to allow them to run under the dev environment.

jdavies-st:
Resolves #1816.

### DIFF
--- a/jwst/model_blender/blendmeta.py
+++ b/jwst/model_blender/blendmeta.py
@@ -91,41 +91,21 @@ def blendmodels(product, inputs=None, output=None, verbose=False):
         output_model = datamodels.open(product)
     else:
         output_model = product
-    """
-        # NOTE 17-Jan-2017:
-        #  Effort needs to be made to insure that new blended values conform
-        #  to the definitions of the attribute as provided by the schema.
-        #  This will address #1650 for the jwst package.
-        #    https://github.com/STScI-JWST/jwst/issues/1650
-        #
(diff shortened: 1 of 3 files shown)


jdavies-st:
Going to merge this now so it gets into the nightly build and allows us to fix up any issues with the overnight regression test data tomorrow before the DMS drop.

jdavies-st:
This PR also resolved a discrepancy between the `spec` and sample `.cfg` file for `resample_spec`.  Both now contain the `blendheaders` keyword.

PR #1839, opened 2018-03-28
Implement more robust TSO checking

### LINKED ISSUE #1815: calspec2 not using correct product suffixes for NIRSpec BrightObj (TSO) exposures
There are hooks in the calwebb_spec2 pipeline that set the correct output product suffix for models that contain multiple integrations using:
```
if isinstance(input, datamodels.CubeModel):
```
which assumes that all multiple-integration models use CubeModel. A while back the processing flow for NIRSpec BrightObj mode exposures (which is a TSO mode and hence the data contain multiple integrations) was modified to send it through the `extract_2d` step, which means the output model type is no longer CubeModel, but rather something like MultiSlitModel (?). Ever since that change, the output files for NIRSpec BrightObj exposures have used the regular `_cal` and `_x1d` suffixes, instead of `_calints` and `_x1dints` (which finally explains why subsequent changes to calspec2 step code have not resulted in regression test failures for this mode, because the reg test isn't using the right input files).

The logic in calspec2 to determine when to use `calints` and `x1dints` needs to be updated to include this type of exposure.

stscieisenhamer:
Resolves #1815

### DIFF
--- a/jwst/lib/pipe_utils.py
+++ b/jwst/lib/pipe_utils.py
@@ -34,7 +34,4 @@ def is_tso(model):
     except AttributeError:
         pass
 
-    # Check on DataModel types
-    is_tso = is_tso or isinstance(model, TSO_MODEL_TYPES)
-
     return is_tso
--- a/jwst/lib/pipe_utils.py
+++ b/jwst/lib/pipe_utils.py
@@ -1,0 +1,40 @@
+"""Pipeline utilities objects"""
+
+from ..associations.lib.dms_base import TSO_EXP_TYPES
+from ..datamodels import CubeModel
+
+# Model types that typically represent TSO's
+TSO_MODEL_TYPES = (CubeModel,)
+
+
+def is_tso(model):
+    """Is data Time Series Observation data?
+
+    Parameters
+    ----------
+    model: datamodels.DataModel
+        Data to check
+
+    Returns
+    -------
+    is_tso: bool
+       `True` if the model represents TSO data
+    """
+    is_tso = False
+
+    # Check on exposure types
+    try:
+        is_tso = model.meta.exposure.type.lower() in TSO_EXP_TYPES
+    except AttributeError:
+        pass
+
+    # Check on JWST-specific TSOVISIT flag
+    try:
+        is_tso = is_tso or model.meta.observation.tsovisit
+    except AttributeError:
+        pass
+
+    # Check on DataModel types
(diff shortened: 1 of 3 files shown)


jwst/lib/pipe_utils.py
```diff
@@ -0,0 +1,40 @@
...
+        pass
+
+    # Check on JWST-specific TSOVISIT flag
+    try:
+        is_tso = is_tso or model.meta.observation.tsovisit
+    except AttributeError:
+        pass
+
+    # Check on DataModel types
+    is_tso = is_tso or isinstance(model, TSO_MODEL_TYPES)
```
hbushouse: Need to be careful with this one. `TSO_MODEL_TYPES` contains `CubeModel` and while most TSO data may be represented using `CubeModel`, not all `CubeModel` instances are TSO. Coronagraphy exposures, for example, are stored as `CubeModel`, because we need the per-integration data. I think it's safer to stick with defining TSO based on only the meta attributes (EXP_TYPE, TSOVISIT) and not the datamodel instance.

hbushouse: Of course if you only ever execute this check within a pipeline like `calwebb_spec2`, then you're safe, because coronagraphy data are image-like and hence go through `calwebb_image2`.

jwst/pipeline/calwebb_spec2.py
```diff
@@ -86,10 +87,7 @@ def process(self, input):
             )
 
             # Save result
-            suffix = 'cal'
-            if isinstance(result, datamodels.CubeModel):
-                suffix = 'calints'
-            self.save_model(result, suffix)
+            self.save_model(result)
```
hbushouse: This looks like you're changing the interface to `save_model`, but I don't see an update to `save_model` in here. Will `save_model` work without supplying the `suffix` argument? Maybe you meant to change the call to just use the regular `save` function?

jwst/lib/pipe_utils.py
stscieisenhamer: 👍 on removing the `DataModel` check. Main reason for the review.

jwst/pipeline/calwebb_spec2.py
stscieisenhamer: Nope. This is intended interface usage.  When no suffix is supplied, the `Step.suffix` is used.  The rest of the removed code is actually duplicate of what is done in the `process_exposure_product` method.

PR #1855, opened 2018-04-01
Setup Fgs pointing

### LINKED ISSUE #1765: FGS guide star data WCS computation for level-1b
# Issue
SDP has asked for our support in providing the algorithms necessary for computing basic WCS info and populating FITS WCS header keywords for level-1b FGS guide star products, similar to the algorithms we provided in the set_telescope_pointing script for application to science exposures.

This is a high-priority task for DMS Build 7.2, in order to satisfy level 4 requirements DMS-189 and DMS-692.

# Resources
- SDP either is already or will be populating the following keywords, based on sources like the SIAF, OSF, and PPSDB:

    - CRPIX1/2
    - CRVAL1/2
    - CDELT1/2 (from SIAF)
    - V2_REF, V3_REF, VPARITY, V3I_YANG (from SIAF)
    - RA_V1, DEC_V1, PA_V3 (from OSF)

# Implementation
- [x] bump version
- [x] tests
- [ ] SCSB code, presumably in a new branch of the set_telescope_pointing script, will compute and populated the PC matrix keywords, based on PA_V3 and VPARITY.
- [ ] The keywords RA_REF, DEC_REF, and ROLL_REF will not appear in guide star product headers.
- [ ] tests
- [ ] bump version

stscieisenhamer:
Resolves #1765


jwst/lib/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '0.8.0'
+__version__ = '0.8.1'
```
hbushouse: Technically this will eventually be part of 0.9.x

jwst/lib/set_telescope_pointing.py
```diff
@@ -148,6 +144,77 @@ def update_wcs(model, default_pa_v3=0., siaf_path=None, **kwargs):
         Keyword arguments used by matrix calculation routines.
     """
 
+    # If the type of exposure is not FGS, then attempt to get pointing
+    # from telemetry.
+    if model.meta.exposure.type.lower().startswith('fgs'):
```
hbushouse: Can't just use a simple test based on `fgs` being the first 3 characters, because science mode exposures from the FGS also start with those characters (e.g. FGS_IMAGE). So you need a specific list of guide-mode EXP_TYPE's, which is FGS_ID-IMAGE, FGS_ID-STACK, FGS_ACQ1, FGS_ACQ2, FGS_TRACK, and FGS_FINEGUIDE.

jwst/lib/tests/test_fgs_pointing.py
```diff
@@ -0,0 +1,72 @@
...
+            'cdelt2': 1.94607944444444e-05,
+            'crpix1': 1024.5,
+            'crpix2': 1024.5,
+            'crval1': 0,
+            'crval2': 0,
+            'ctype1': 'RA---TAN',
+            'ctype2': 'DEC--TAN',
+            'cunit1': 'deg',
+            'cunit2': 'deg',
+            'dec_ref': 0.0,
```
hbushouse: I believe we had decided to not include the RA_REF, DEC_REF, and ROLL_REF keyword values for guide mode exposures, so you don't really need these in the list (unless something in an interface somewhere depends on them being there).

jwst/lib/set_telescope_pointing.py
```diff
@@ -148,6 +144,77 @@ def update_wcs(model, default_pa_v3=0., siaf_path=None, **kwargs):
...
+        update_wcs_from_telem(
+            model, default_pa_v3=default_pa_v3, siaf_path=siaf_path, **kwargs
+        )
+
+
+def update_wcs_from_header(model, default_pa_v3=0., siaf_path=None, **kwargs):
+    """ Update WCS pointing from header information
+
+    For FGS-like observations, nearly all information is already populated
+    except for the CD matrix. This simply calculates that matrix.
```
nden: Could we include a more specific comment here - to facilitate future readings. These are FGS GUIDER observations.
We populate only the FITS WCS keywords.

jwst/lib/set_telescope_pointing.py
```diff
@@ -148,6 +144,77 @@ def update_wcs(model, default_pa_v3=0., siaf_path=None, **kwargs):
...
+        The model to update.
+    default_pa_v3 : float
+        If pointing information cannot be retrieved,
+        use this as the V3 position angle.
+    siaf_path : str
+        Unused and present only for API compatibility.
+
+    kwargs: dict
+        Unused and present only for API compatibility.
+    """
```
nden: May be just drop the unused arguments, especially `siaf_path` from this function. Unless we need to include a footprint calculation for GUIDER observations, @hbushouse ?

jwst/lib/set_telescope_pointing.py
```diff
@@ -148,6 +144,77 @@ def update_wcs(model, default_pa_v3=0., siaf_path=None, **kwargs):
...
+    try:
+        pa = model.meta.wcsinfo.pa_v3
+    except AttributeError:
+        pa = default_pa_v3
+    pa_rad = pa * D2R
+
+    model.meta.wcsinfo.pc1_1 = -np.cos(pa_rad)
+    model.meta.wcsinfo.pc1_2 = np.sin(pa_rad)
+    model.meta.wcsinfo.pc2_1 = np.sin(pa_rad)
+    model.meta.wcsinfo.pc2_2 = np.cos(pa_rad)
```
nden: Is this the correct angle rotation? Usually the `-` sign goes with one of the `sin` functions depending on the direction.

jwst/lib/set_telescope_pointing.py
```diff
@@ -148,6 +144,77 @@ def update_wcs(model, default_pa_v3=0., siaf_path=None, **kwargs):
...
+
+    kwargs: dict
+        Unused and present only for API compatibility.
+    """
+
+    logger.info('Updating WCS from headers.')
+
+    # Get position angle
+    try:
+        pa = model.meta.wcsinfo.pa_v3
```
nden: The `PA_V3` angle in `model.meta.wcsinfo.pa_v3` is computed in ` calc_wcs`. It's not in the SIAF. 
So this will not have the correct value. Also, perhaps it returns `None` instead of raising an `AttributeError` since it's in the core schema.

jwst/lib/set_telescope_pointing.py
hbushouse: You're right that there's no need to compute a footprint (S_REGION value) for guider mode exposures.

jwst/lib/set_telescope_pointing.py
hbushouse: For guider mode exposures, SDP will be getting values from the OSF (Observatory Status File) event messages for populating the RA_V1, DEC_V1, and PA_V3 keywords, rather than the usual procedure for science exposures where our scripts (like calc_wcs) compute those values. So we should already have a value of PA_V3 to work with in the input product.

jwst/lib/set_telescope_pointing.py
hbushouse: That's exactly the same as the way it's currently computed for science exposures in the `update_wcs` function. So maybe they're both wrong?

jwst/lib/set_telescope_pointing.py
nden: Mention this in the comments for those that forget.

jwst/lib/tests/test_fgs_pointing.py
stscieisenhamer: Sure, just copied from another test file, but will remove.

jwst/lib/set_telescope_pointing.py
stscieisenhamer: This came from further past in the history, but will go over the docs to make sure.

stscieisenhamer: However, the question I have here is whether `vparity` should play a part here. For science, `vparity` is incorporated into the full transformation path. However, since we're short-circuiting here, I've got a sneaky suspicion that it needs to be included. I'll go through the transform logic, but was hoping the answer was already known 😄

stscieisenhamer: Also, about the science version of the transform: Have we not already had confirmation through archive/COAM with their preview code? I presume they are using the FITS WCS for that work?

nden: An example of using the parity is here

https://github.com/spacetelescope/stwcs/blob/master/stwcs/updatewcs/makewcs.py#L185

hbushouse: As far as confirmation from CAOM is concerned, it's not obvious to me that we've ever yet had any consistent and/or reliable WCS info go all the way through the whole system to allow that kind of confirmation. Most of the test data end up with a rotation of zero and hence the PC matrix is just 1's and 0's, so there's been no confirmation of whether the rotation implied by the PC is correct. The most that's been verified is that when we set PC1_1 = -1, PC2_2 = 1 (and the cross terms to 0) images display properly with North up and East to the left.

jwst/lib/set_telescope_pointing.py
stscieisenhamer: 👍

jwst/lib/set_telescope_pointing.py
stscieisenhamer: Discussed with Colin Cox and he says that VIdlParity should be included. He notes that for guiding, the OSS apertures will be used where VIdlParity=1. Regardless, added in VIdlParity inclusion.

jdavies-st: I'll add that VPARITY for all instruments, including FGS, is -1.  This means the sense of rotation of the V2/V3 frame is the opposite of that of the science X/Y frames, i.e. positive rotation of one means a negative rotation in the other.

stscieisenhamer:
@nden @jdavies-st The CD/PC matrix calculation should now be explicit and consistent. Correct is another matter.

jdavies-st:
So it turns out that the block of code for computing the PC matrix from position angle was correct if you assume that all instruments have VPARITY = -1, which they currently do.  But it would be good to make this explicit.  This also means that the code for doing it for GUIDER mode needs to be modified, or use the same function as for science data.

PA_V3 is the angle from north towards east of the V3 axis of the telescope. The direction of north is at V1.  So PA_V3 increases as V3 moves counter-clockwise.  And of course V2/V3 has the same parity as RA/DEC.  Both are right-handed systems.

So now, as @nden mentions, The normal equations for computing the PC matrix are
```
pc1_1, pc1_2 = np.cos(theta), np.sin(theta)
pc2_1, pc2_2 = -np.sin(theta), np.cos(theta)
```
but this assumes `theta` is an angle between the Y axis of the detector and celestial north in the direction of y towards x, i.e. `theta` increases as as Y moves clockwise.

So in order to compute the PC matrix using PA_V3, we need to account for VPARITY in the computation.  VPARITY is a flip of the X axis across the Y axis.  So we multiply the X terms by VPARITY
```
pc1_1, pc1_2 = vparity * np.cos(theta), np.sin(theta)
pc2_1, pc2_2 = -vparity * np.sin(theta), np.cos(theta)
```
And in the case where VPARITY=-1, this simplifies to what was in the code:
```
    model.meta.wcsinfo.pc1_1 = -np.cos(pa_rad)
    model.meta.wcsinfo.pc1_2 = np.sin(pa_rad)
    model.meta.wcsinfo.pc2_1 = np.sin(pa_rad)
    model.meta.wcsinfo.pc2_2 = np.cos(pa_rad)
```
But I agree we should explicitly use VPARITY in the computation of the PC matrix and not hard-code it as we currently do, assuming it is always -1.

jwst/lib/set_telescope_pointing.py
```diff
@@ -148,6 +155,100 @@ def update_wcs(model, default_pa_v3=0., siaf_path=None, **kwargs):
...
+            model, default_pa_v3=default_pa_v3, siaf_path=siaf_path, **kwargs
+        )
+
+
+def update_wcs_from_fgs_guiding(model, default_pa_v3=0.0, default_vparity=1):
+    """ Update WCS pointing from header information
+
+    For Fine Guidance guidine observations, nearly everything
+    in the `wcsinfo` meta information is already populated,
+    except for the CD matrix. This function updates the CD
```
hbushouse: We're updating the PC matrix, not CD.

jwst/lib/set_telescope_pointing.py
```diff
@@ -148,6 +155,100 @@ def update_wcs(model, default_pa_v3=0., siaf_path=None, **kwargs):
...
+    else:
+        update_wcs_from_telem(
+            model, default_pa_v3=default_pa_v3, siaf_path=siaf_path, **kwargs
+        )
+
+
+def update_wcs_from_fgs_guiding(model, default_pa_v3=0.0, default_vparity=1):
+    """ Update WCS pointing from header information
+
+    For Fine Guidance guidine observations, nearly everything
```
hbushouse: typo: "guidine" should be "guiding"

jwst/lib/set_telescope_pointing.py
```diff
@@ -992,4 +1104,45 @@ def _add_axis_3(model):
...
+    angle: float in radians
+        The angle to create the matrix
+
+    vparity: {1, -1}
+        The x-axis parity, usually taken from
+        the JWST SIAF parameter VIdlParity
+
+    Returns
+    -------
+    matrix: [cd1_1, cd1_2, cd2_1, cd2_2]
```
hbushouse: While it doesn't affect the outcome of the function, we really shouldn't be referring to these values as "cd". In the FITS standard WCS the "CD" matrix elements contain both rotation and scale. What we're computing is just rotation, which is designated as the "PC" matrix.

jwst/lib/set_telescope_pointing.py
```diff
@@ -148,6 +155,100 @@ def update_wcs(model, default_pa_v3=0., siaf_path=None, **kwargs):
...
+    Parameters
+    ----------
+    model : `~jwst.datamodels.DataModel`
+        The model to update.
+
+    default_pa_v3 : float
+        If pointing information cannot be retrieved,
+        use this as the V3 position angle.
+
+    default_vparity: {-1, 1}
```
nden: I'm not sure how sphinx will parse this as it is supposed to be a type. Explain the possible values in the definition. 
Also use double quotes as in ``1`` , otherwise sphinx will attempt to link to it.

jwst/lib/set_telescope_pointing.py
```diff
@@ -992,4 +1104,45 @@ def _add_axis_3(model):
...
+
+def calc_rotation_matrix(angle, vparity=1):
+    """ Calculate the rotation matrix
+
+    Parameters
+    ----------
+    angle: float in radians
+        The angle to create the matrix
+
+    vparity: {1, -1}
```
nden: Same as above - use type here, not values.

jwst/lib/set_telescope_pointing.py
```diff
@@ -992,4 +1104,45 @@ def _add_axis_3(model):
...
+        cd1_1 = vparity * cos(angle)
+        cd1_2 = sin(angle)
+        cd2_1 = -1 * vparity * sin(angle)
+        cd2_2 = cos(angle)
+    """
+
+    cd1_1 = vparity * cos(angle)
+    cd1_2 = sin(angle)
+    cd2_1 = -1 * vparity * sin(angle)
+    cd2_2 = cos(angle)
```
nden: BTW, it's somewhat faster to use `vparity * -cos(angle)`.
It's not part of this PR but since we are looking at these functions now,  it's ~10x faster to use `math.cos` instead of `np.cos`. Since there's no reason here to use the `numpy` trig functions we should switch to using `math`.

stscieisenhamer:
Review comments addressed.

Issue #1858, opened 2018-04-03
Replace deprecated sphere package


stsci-hack:
The image alignment code used by the outlier_detection step still relies on the original spherical geometry package developed and originally used in the HST pipeline; namely, stsci.sphere.  However, it has since been replaced and subsequently deprecated.  The use of this package by the JWST code should be replaced with calls to the new spherical_geometry package instead.

pllim:
Note: This is making RTD fail for the past 3 days. Some intermediate quick-fix might be needed to get RTD to build again. FYI.

jdavies-st:
It looks like outlier_detection doesn't depend on `stsci.sphere` but `tweakreg` and `skymatch` do.  Thoughts @mcara?

mcara:
Link to the opened PR: https://github.com/STScI-JWST/jwst/pull/1859

stsci-hack:
The astroconda dev specification will be modified using ([astroconda dev PR #126](https://github.com/astroconda/astroconda-dev/pull/126) to pick up spherical geometry instead of stsci.sphere, and, as I understand it, allow the RTD build to succeed.

pllim:
I am not convinced that #1859 fixed everything though. RTD and Travis failures were not addressed in that PR, which has since been merged.

pllim:
p.s. I'll give RTD a few minutes to build to see if we need to explicitly list the new dependency on that end. https://readthedocs.org/projects/jwst-pipeline/builds/6986317/

pllim:
Update: RTD failed
```
Running Sphinx v1.7.2
loading translations [en]... done
making output directory...
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://docs.scipy.org/doc/numpy/objects.inv...
loading intersphinx inventory from https://docs.scipy.org/doc/scipy/reference/objects.inv...
loading intersphinx inventory from http://matplotlib.org/objects.inv...
intersphinx inventory has moved: http://matplotlib.org/objects.inv -> https://matplotlib.org/objects.inv
[...]

PR #1859, opened 2018-04-03
Initial replacement of code and docstrings

stsci-hack:
All references to 'stsci.sphere' that could be found within the JWST package should have been replaced, with this PR, with references to the newer spherical_geometry package.  This includes the code as well as the docs.

### DIFF
--- a/jwst/skymatch/skyimage.py
+++ b/jwst/skymatch/skyimage.py
@@ -1,4 +1,4 @@
-"""
+spherical_geometry"""
 The ``skyimage`` module contains algorithms that are used by
 ``skymatch`` to manage all of the information for footprints (image outlines)
 on the sky as well as perform useful operations on these outlines such as
@@ -16,7 +16,7 @@ from __future__ import (absolute_import, division, unicode_literals,
 import numpy as np
 
 # THIRD-PARTY
-from stsci.sphere.polygon import SphericalPolygon
+from spherical_geometry.polygon import SphericalPolygon
 
 #LOCAL
 from . skystatistics import SkyStats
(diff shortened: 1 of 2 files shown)


pllim:
Note: Fix #1858 

Does RTD requirements need updating to pick up this new dependency, @jhunkeler or @rendinam ?

jhunkeler:
The RTD environment file should be updated to include the new dependency, yes. The old sphere stuff was installed via drizzlepac as an implicit requirement and it didn’t work out too well for us.

pllim:
By the way, Travis failed for this PR. Error seems related. FYI.

jwst/skymatch/skyimage.py
```diff
@@ -1,4 +1,4 @@
-"""
+spherical_geometry"""
```
mcara: @stsci-hack Undescore begins "italics" but it never closes it. You should use double back-ticks for package name.

mcara: @stsci-hack Actually, disregard my previous comment - simply remove `spherical_geometry` before the docstring. What is it doing there?

stsci-hack: This change is fine...

hbushouse:
According to @jhunkeler in #1864 the old package named `spherical_geometry` is being removed and replaced by `spherical-geometry`. So do these changes to import need to be updated to use the new version of the package name?

jhunkeler:
The package `spherical-geometry` provides the python package `spherical_geometry`. So this is fine.

Note: `install_requires = ['spherical_geometry']` is fine. `pip` and `setuptools` will still resolve it correctly on PyPi

PR #1860, opened 2018-04-03
slit PPS to slit DMS coordinate conversion

nden:
Fix an issue with converting between PPS and DMS shutter coordinate systems.
The source position in a shutter is reported in the MSA configuration file in a coordinate system associated with the shutter with origin (0, 0) in the upper left corner (PPS system). `x` increases to the right. `y` increases downwards. The DMS uses a shutter coordinate system with origin  (0, 0) in the center of the shutter, with positive `x` to the right and positive `y` upwards.

### DIFF
--- a/jwst/assign_wcs/nirspec.py
+++ b/jwst/assign_wcs/nirspec.py
@@ -457,11 +457,10 @@ def get_open_msa_slits(msa_file, msa_metadata_id):
             columns "estimated_source_in_shutter_x" and "estimated_source_in_shutter_y".
             The source position is in a coordinate system associated with each shutter whose
             origin is the upper left corner of the shutter, positive x is to the right
-            and positive y is downwards. To convert to the coordinate frame associated with the
-            slit, where (0, 0) is in the center of the slit, we subtract 0.5 in both directions.
+            and positive y is downwards.
             """
             source_xpos = source_xpos - 0.5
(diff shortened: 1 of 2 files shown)


jwst/assign_wcs/nirspec.py
```diff
@@ -457,11 +457,10 @@ def get_open_msa_slits(msa_file, msa_metadata_id):
             columns "estimated_source_in_shutter_x" and "estimated_source_in_shutter_y".
             The source position is in a coordinate system associated with each shutter whose
             origin is the upper left corner of the shutter, positive x is to the right
-            and positive y is downwards. To convert to the coordinate frame associated with the
-            slit, where (0, 0) is in the center of the slit, we subtract 0.5 in both directions.
+            and positive y is downwards.
             """
             source_xpos = source_xpos - 0.5
-            source_ypos = source_ypos - 0.5
+            source_ypos = -source_ypos + 0.5
```
stscirij: This looks right:

(0,0) in PPS coordinates (upper left corner of aperture) becomes (-0.5, 0.5)
(0.5, 0.5) in PPS coordinates (center of aperture) becomes (0.0, 0.0)

hbushouse:
And a source in the bottom-right corner would be at PPS coords of (+1,+1), which according to the math above translates to (+0.5, -0.5) in DMS, which also looks right.

PR #1863, opened 2018-04-03
Remove extraneous spherical_geometry occurence

mcara:
Fixes error introduced in https://github.com/STScI-JWST/jwst/issues/1858

CC: @stsci-hack @pllim @hbushouse

### DIFF
--- a/jwst/skymatch/skyimage.py
+++ b/jwst/skymatch/skyimage.py
@@ -1,4 +1,4 @@
-spherical_geometry"""
+"""
 The ``skyimage`` module contains algorithms that are used by
 ``skymatch`` to manage all of the information for footprints (image outlines)
 on the sky as well as perform useful operations on these outlines such as


hbushouse:
**_NO MERGING_** until Travis completes successfully.

pllim:
p.s. @hbushouse -- There is a repo setting to disable merge unless CI passes. FYI. However, the downside of this setting is that any PR with `[skip ci]` in commit (e.g., a simple comment line typo fix) will also be forbidden.

mcara:
I have asked @rendinam to notify us when astroconda was rebuild with the new `spherical_geometry` package so that @hbushouse could restart the failing test.

rendinam:
Bit of chicken-and-egg here. The package cannot be built successfully without incorporating this correction. Is it possible to merge this one-line fix so that the package can be rebuilt?

mcara:
@rendinam I would say yes but I am afraid of @hbushouse

pllim:
Or... Publish `spherical_geometry` in PyPI and use `pip install`? :grimacing:

PR #1864, opened 2018-04-03
Add explicit dependencies to travis.yml

jhunkeler:
Relying on the `jwst` package for dependency resolution is not a viable solution in certain situations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,22 +10,42 @@ sudo: false
 addons:
     apt:
         packages:
-            - graphviz
             - texlive-latex-extra
             - dvipng
 
 env:
     global:
-        - MAIN_CMD='python setup.py'
-        - CONDA_CHANNELS='http://ssb.stsci.edu/conda-dev'
-        - CONDA_DEPENDENCIES='pytest jwst sphinx'
-        - CONDA_JWST_DEPENDENCIES='pytest jwst sphinx'
-        - PIP_DEPENDENCIES=''
+        - CONDA_CHANNELS='http://ssb.stsci.edu/astroconda-dev'
+        - CONDA_DEPENDENCIES="asdf
+                              astropy
+                              crds
+                              dask
+                              drizzle
+                              fitsblender
+                              flake8
+                              gwcs
+                              jsonschema
+                              jplephem
+                              matplotlib
+                              namedlist
+                              numpy
+                              photutils
+                              scipy
+                              six
+                              spherical-geometry
+                              stsci.image
+                              stsci.imagestats
+                              stsci.stimage
+                              stsci.tools
+                              stwcs
+                              verhawk"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -10,16 +10,36 @@ sudo: false
...
+                              gwcs \
+                              jsonschema \
+                              jplephem \
+                              matplotlib \
+                              namedlist \
+                              numpy \
+                              photutils \
+                              scipy \
+                              six \
+                              spherical-geometry \
```
hbushouse: Is it `spherical-geometry` or `spherical_geometry`? (underscore vs dash)

jhunkeler:
The conda package is named `spherical-geometry`. There was older package named `spherical_geometry` that @rendinam is removing it from the channel.

.travis.yml
```diff
@@ -10,22 +10,42 @@ sudo: false
...
+                              numpy
+                              photutils
+                              scipy
+                              six
+                              spherical-geometry
+                              stsci.image
+                              stsci.imagestats
+                              stsci.stimage
+                              stsci.tools
+                              stwcs
```
nden: I thought we removed this dependency :confused:

jhunkeler: Which dependency?

nden: Sorry I meant stwcs.

jhunkeler:
This is ready to squash/merge unless someone finds a problem.

.travis.yml
jhunkeler: It's gone now :)

.travis.yml
```diff
@@ -10,22 +10,41 @@ sudo: false
...
             - dvipng
 
 env:
     global:
-        - MAIN_CMD='python setup.py'
-        - CONDA_CHANNELS='http://ssb.stsci.edu/conda-dev'
-        - CONDA_DEPENDENCIES='pytest jwst sphinx'
-        - CONDA_JWST_DEPENDENCIES='pytest jwst sphinx'
-        - PIP_DEPENDENCIES=''
+        - CONDA_CHANNELS='http://ssb.stsci.edu/astroconda-dev'
```
hbushouse: Why the switch from conda-dev to astroconda-dev? Is conda-dev going away?

jhunkeler: From an administrative point of view we found it easier to write scripts around `astroconda-dev` than `conda-dev`. The original `conda-dev` name isn't going anywhere for historical reasons so you're free to keep using it.

```sh
channel_public="http://ssb.stsci.edu/astroconda"
channel_dev="${channel_public}-dev"
```

jhunkeler:
Resolves #1858

hbushouse:
Going to try merging, despite current issue with ci-helpers in Travis. Cross your fingers and watch for smoke.

PR #1865, opened 2018-04-03
Install readme

stscicrawford:
An update to the installation instructions to the README to point to using conda to install and maintain the software and its dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,30 +17,26 @@ Installing
 To install the latest stable version of the library, we recommend using [conda](https://conda.io/docs/index.html) and
 the [astroconda](https://astroconda.readthedocs.io) channel:
 
-    % conda config --add channels http://ssb.stsci.edu/astroconda-dev
-    % conda install jwst
+    conda config --add channels http://ssb.stsci.edu/astroconda-dev
+    conda install jwst
 
 To install the development version of the repository, we recommend using the [astroconda](https://astroconda.readthedocs.io) channel
 to install the dependencies, and then installing from the github repository:
 
-    % conda config --add channels http://ssb.stsci.edu/astroconda-dev
-
-    % conda install jwst
-
-    % git clone https://github.com/STScI-JWST/jwst.git
-
-    % cd jwst
-
-    % python setup.py install
+    conda config --add channels http://ssb.stsci.edu/astroconda-dev
+    conda install jwst
+    git clone https://github.com/STScI-JWST/jwst.git
+    cd jwst
+    python setup.py install
 
     or
 
-    % python setup.py develop
+    python setup.py develop
 
 Once installed, the software can be updated to the lastest development version by running the following command inside the `jwst` 
 repository:
 
-    % git pull origin master
+    git pull origin master
 
 
 Contributing Code, Documentation or Feedback
--- a/README.md
+++ b/README.md
@@ -13,17 +13,19 @@ Beginning with the next version (0.9.0), **JWST requires Python 3.5 and above**.
 
 Installing
 ----------
-Releases are made on the **stable** branch. To install the latest release:
 
-    % git clone https://github.com/STScI-JWST/jwst.git
+To install the latest stable version of the library, we recommend using [conda](https://conda.io/docs/index.html) and
+the [astroconda](https://astroconda.readthedocs.io) channel:
 
-    % cd jwst
+    % conda config --add channels http://ssb.stsci.edu/astroconda-dev
+    % conda install jwst
 
-    % git checkout stable
+To install the development version of the repository, we recommend using the [astroconda](https://astroconda.readthedocs.io) channel
+to install the dependencies, and then installing from the github repository:
 
-    % python setup.py install
+    % conda config --add channels http://ssb.stsci.edu/astroconda-dev
 
-The main development is on the **master** branch. To install the development version:
+    % conda install jwst
 
     % git clone https://github.com/STScI-JWST/jwst.git
 
@@ -34,6 +36,11 @@ The main development is on the **master** branch. To install the development ver
     or
 
     % python setup.py develop
+
+Once installed, the software can be updated to the lastest development version by running the following command inside the `jwst` 
+repository:
+
+    % git pull upstream master
 
 
 Contributing Code, Documentation or Feedback
--- a/README.md
+++ b/README.md
@@ -9,32 +9,39 @@ JWST Calibration Pipeline
 
 Note
 ----
-Beginning with  version 0.9.0, **JWST requires Python 3.5 and above**.
+Beginning with version 0.9.0, **JWST requires Python 3.5 and above**.
 
 Installing
 ----------
 
-To install the latest stable version of the library, we recommend using [conda](https://conda.io/docs/index.html) and
-the [astroconda](https://astroconda.readthedocs.io) channel:
+To install the latest stable version of the library, we recommend using [conda](https://conda.io/docs/index.html)
+and the [astroconda](https://astroconda.readthedocs.io) channel to install the
+software and using an conda environemnt for the installation:
 
     conda config --add channels http://ssb.stsci.edu/astroconda-dev
+    conda create -n jwst python=3
+    source activate jwst
     conda install jwst
 
-To install the development version of the repository, we recommend using the [astroconda](https://astroconda.readthedocs.io) channel
+To install the development version of the repository, we recommend created a new
+nvironment, using the [astroconda](https://astroconda.readthedocs.io) channel
 to install the dependencies, and then installing from the github repository:
 
     conda config --add channels http://ssb.stsci.edu/astroconda-dev
+    conda create -n jwst python=3
+    source activate jwst
     conda install jwst
     git clone https://github.com/STScI-JWST/jwst.git
     cd jwst
     python setup.py develop
 
-If you will not need to update the code, then you can replace the last step with the following command:
+If you will not need to update the code, then you can replace the last step with
+the following command:
 
     python setup.py install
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,27 +13,30 @@ Beginning with the next version (0.9.0), **JWST requires Python 3.5 and above**.
...
-    % cd jwst
+    conda config --add channels http://ssb.stsci.edu/astroconda-dev
+    conda install jwst
 
-    % git checkout stable
+To install the development version of the repository, we recommend using the [astroconda](https://astroconda.readthedocs.io) channel
+to install the dependencies, and then installing from the github repository:
 
-    % python setup.py install
+    conda config --add channels http://ssb.stsci.edu/astroconda-dev
```
jdavies-st: I recall adding dev to the channels was a no-no back when we first set all this up.  Is that still the case @jhunkeler?

README.md
```diff
@@ -13,27 +13,30 @@ Beginning with the next version (0.9.0), **JWST requires Python 3.5 and above**.
```
hbushouse: Now that we're past version 0.9.0, let's change the wording of line 12 to "Beginning with version 0.9.0, ..."

README.md
```diff
@@ -13,27 +13,30 @@ Beginning with the next version (0.9.0), **JWST requires Python 3.5 and above**.
...
 ----------
-Releases are made on the **stable** branch. To install the latest release:
 
-    % git clone https://github.com/STScI-JWST/jwst.git
+To install the latest stable version of the library, we recommend using [conda](https://conda.io/docs/index.html) and
+the [astroconda](https://astroconda.readthedocs.io) channel:
 
-    % cd jwst
+    conda config --add channels http://ssb.stsci.edu/astroconda-dev
+    conda install jwst
```
hbushouse: Is jwst available from the astroconda-dev channel? To date we've been using conda-dev.

README.md
```diff
@@ -13,27 +13,30 @@ Beginning with the next version (0.9.0), **JWST requires Python 3.5 and above**.
...
 
-The main development is on the **master** branch. To install the development version:
+If you will not need to update the code, then you can replace the last step with the following command:
 
-    % git clone https://github.com/STScI-JWST/jwst.git
+    python setup.py install
 
-    % cd jwst
+Once installed, the software can be updated to the lastest development version by running the following command inside the `jwst` 
+repository:
```
hbushouse: Is it really correct to refer to a user's checked-out/cloned version of the repo as "the jwst repository"? When I read that, I think of the actual repository in github.

stscicrawford:
@hbushouse @jdavies-st @jhunkeler  I'm happy to change the channel that should be used, if someone ones to point me to the right one.   That was the first one that I found and that worked.  

I expect the installation instructions to be evolving as we have different releases, but the goal here was to update them for someone starting fresh with our current recommended installation instructions

README.md
jdavies-st: A bunch of this info is in the **Conda Dev and You** document on Confluence, but I agree it the basics of how to install and update should be here in the repository.

README.md
```diff
@@ -9,31 +9,41 @@ JWST Calibration Pipeline
...
 ----------
-Releases are made on the **stable** branch. To install the latest release:
 
-    % git clone https://github.com/STScI-JWST/jwst.git
+To install the latest stable version of the library, we recommend using [conda](https://conda.io/docs/index.html)
+and the [astroconda](https://astroconda.readthedocs.io) channel to install the
+software and using an conda environemnt for the installation:
 
-    % cd jwst
+    conda config --add channels http://ssb.stsci.edu/astroconda-dev
```
jhunkeler: **_Danger_**
Adding the `astroconda-dev` channel to your general conda configuration will make your `astroconda` channel experience very unpleasant. Development packages will supersede stable packages from that point forward.

README.md
```diff
@@ -9,31 +9,41 @@ JWST Calibration Pipeline
...
+    conda create -n jwst python=3
```
jhunkeler: The only "sane" way to install development packages without breaking your entire conda ecosystem is:
```
$ conda create -n jwst --override-channels -c http://ssb.stsci.edu/astroconda-dev -c defaults jwst
```

_NOTE:_ Installing `jwst` from the development channel like this will always yield the latest `master` build. In order to install a `stable` package, you need to supply the exact version: `jwst=9.0.0.dev0`.

_NOTE_: Because the package, channel, and dependencies are marching forward, this might not work as intended anyway. You *could* point people to the "stable" development snapshots under https://github.com/astroconda/astroconda-releases instead.

README.md
```diff
@@ -9,31 +9,41 @@ JWST Calibration Pipeline
...
+    git clone https://github.com/STScI-JWST/jwst.git
+    cd jwst
+    python setup.py develop
 
-The main development is on the **master** branch. To install the development version:
+If you will not need to update the code, then you can replace the last step with
+the following command:
 
-    % git clone https://github.com/STScI-JWST/jwst.git
+    python setup.py install
```
jhunkeler: The pipeline's dependency list is gigantic and requires a few non-python packages in order to run. Here's a way to install those dependencies using `--only-deps` (as of conda 4.4.10):

```
$ conda create -n jwst_devel --only-deps --override-channels -c http://ssb.stsci.edu/astroconda-dev -c defaults jwst
```

README.md
```diff
@@ -9,31 +9,41 @@ JWST Calibration Pipeline
...
+to install the dependencies, and then installing from the github repository:
 
-    % python setup.py install
+    conda config --add channels http://ssb.stsci.edu/astroconda-dev
+    conda create -n jwst python=3
+    source activate jwst
+    conda install jwst
+    git clone https://github.com/STScI-JWST/jwst.git
+    cd jwst
+    python setup.py develop
```
jhunkeler: The same thing applies to this block^^

README.md
```diff
@@ -9,31 +9,42 @@ JWST Calibration Pipeline
...
+Beginning with version 0.9.0, **JWST requires Python 3.5 and above**.
 
 Installing
 ----------
-Releases are made on the **stable** branch. To install the latest release:
 
-    % git clone https://github.com/STScI-JWST/jwst.git
+To install the latest released version of the library, we recommend using [conda](https://conda.io/docs/index.html)
+and the [astroconda](https://astroconda.readthedocs.io) channel to install the
+software and using a conda environemnt for the installation:
```
jhunkeler: typo: `environemnt` :)

README.md
```diff
@@ -9,31 +9,42 @@ JWST Calibration Pipeline
...
 
-    % cd jwst
+    conda create -n jwst --override-channels -c http://ssb.stsci.edu/astroconda-dev -c defaults jwst
+    source activate jwst
 
-    % git checkout stable
+This will install the latest released version of the package, and within the
+`jwst` environment, the pipeline can be run from the command line or imported
+into python programs.   In order to install a specific version, the exact 
+version can be supplied, e.g. jwst=9.0.0.dev0.
```
jhunkeler: This should have been `jwst=0.9.0.dev0`; my fault! Although maybe it doesn't matter because this is only an example string.

README.md
```diff
@@ -9,31 +9,42 @@ JWST Calibration Pipeline
...
+    source activate jwst
 
-    % git checkout stable
+This will install the latest released version of the package, and within the
+`jwst` environment, the pipeline can be run from the command line or imported
+into python programs.   In order to install a specific version, the exact 
+version can be supplied, e.g. jwst=0.9.0. 
 
-    % python setup.py install
+To install the development version of the repository, we recommend created a new
```
hbushouse: "created" should be "creating"

README.md
```diff
@@ -9,31 +9,42 @@ JWST Calibration Pipeline
...
+version can be supplied, e.g. jwst=0.9.0. 
 
-    % python setup.py install
+To install the development version of the repository, we recommend creating a new
+environment, using the [astroconda](https://astroconda.readthedocs.io) channel
+to install the dependencies, and then installing from the github repository:
 
-The main development is on the **master** branch. To install the development version:
+    conda create -n jwst_devel --only-deps --override-channels -c http://ssb.stsci.edu/astroconda-dev -c defaults jwst
+    source activate jwst
```
jdavies-st: This should probably be `source activate jwst_devel`, as that's the name given above.

README.md
```diff
@@ -9,31 +9,42 @@ JWST Calibration Pipeline
...
+    git clone https://github.com/STScI-JWST/jwst.git
```
jdavies-st: If the user is going to contribute code, they should fork the repo and then clone their fork locally.  But maybe we don't want to get into that here.  Keeping it simple here is good.

README.md
```diff
@@ -9,31 +9,42 @@ JWST Calibration Pipeline
...
 ----
-Beginning with the next version (0.9.0), **JWST requires Python 3.5 and above**.
+Beginning with version 0.9.0, **JWST requires Python 3.5 and above**.
 
 Installing
 ----------
-Releases are made on the **stable** branch. To install the latest release:
 
-    % git clone https://github.com/STScI-JWST/jwst.git
+To install the latest released version of the library, we recommend using [conda](https://conda.io/docs/index.html)
```
jdavies-st: Perhaps not call it "latest released version" and instead call it "previous night's build" or something similar.  Releases can actually be pulled from the normal `astroconda` channel and don't need this dev stuff.

README.md
stscicrawford: Thanks good catch

README.md
stscicrawford: This is the latest 'drop' for the purposes of supplying a release to DMS, so to keep it simple, I'll leave it as is.

PR #1869, opened 2018-04-04
Add exp_type vs pipeline info

hbushouse:
Adding more detailed information to the pipeline docs that shows which pipeline modules get applied to each EXP_TYPE value. Similar to, but more complete than, the info in JDox (https://jwst-docs.stsci.edu/display/JDAT/Modes+of+Observing)


hbushouse:
Merging despite unrelated Travis error involving ci-helpers.

PR #1877, opened 2018-04-05
Modify code that clears meta.model_type

bernie-simon:
The generic open function in datamodels does not set the meta.model_type attribute because the class it uses to open an image is only a best guess. Previously the code cleared the model_type by
setting it to None. This is not desirable, because if the metadata is re-validated, a value of None with throw a warning or error. This revision deletes the model_type attribute instead. To keep datamodels
from setting the value again, the code in on_save which sets model_type when not found was removed. It was redundant in any case.


hbushouse:
Are the Travis failures related to this change or a very recent change in asdf? Either way, the unit tests need fixing before we can merge this.

PR #1883, opened 2018-04-06
Pathloss 2d

### LINKED ISSUE #1152: Pathloss step needs to compute and store 2-D corrections
The current implementation of the pathloss step in calspec2 computes and stores 1-D correction vectors that are a function of wavelength, which was done under the assumption that the corrections wouldn't be applied until after a 1-D spectrum had been extracted. These corrections are one of several corrections/calibrations that will now need to be applied already to the 2-D spectral data (e.g. especially for IFU exposures) and hence the step needs to be updated to produce full 2-D correction arrays for each image or slit instance.

stscirij:
Added code to output the 2-d pathloss correction and modified the multislit datamodel to include them.  Fixes #1152


jwst/datamodels/schemas/pathlosscorr.schema.yaml
```diff
@@ -1,5 +1,10 @@
 type: object
 properties:
+  pathloss_pointsource2d:
```
hbushouse: Why create a new 2D version of the arrays in the schema, as opposed to just converting the existing ones from 1D to 2D?

Also, because the new correction factors will presumably be sampled onto exactly the same 2D grid as the science arrays, they will then also have the same wavelength assignments per pixel, so can we get rid of the wavelength arrays from the schema?

jwst/datamodels/schemas/pathlosscorr.schema.yaml
```diff
@@ -10,6 +15,11 @@ properties:
     fits_hdu: WAVELENGTH_POINTSOURCE
     ndim: 1
     datatype: float32
+  pathloss_uniformsource2d:
+    title: 2-d array for pathloss (uniform source)
+    fits_hdu: PATHLOSS_UNIFORMSOURCE2D
```
hbushouse: For the ease of folks interfacing to these specific extensions, is there a way we could trim down the length of their names, while still retaining clarity? For example, `PATHLOSS_POINT` and `PATHLOSS_UNIFORM`?

jwst/datamodels/schemas/pathlosscorr.schema.yaml
stscirij: I'm keeping the 1-d pathloss correction arrays, the wavelength arrays refer to the 1-d correction.  The 2-d pathloss correction is already interpolated using the wavelength array, so no 2-d array is needed.

jwst/datamodels/schemas/pathlosscorr.schema.yaml
stscirij: Still need the 2d part as the 1-d is also there.  Not sure it's worth taking out the SOURCE part, it is what the POINT and UNIFORM adjectives describe...

PR #1884, opened 2018-04-06
Make suffix removal more robust.

### LINKED ISSUE #1870: suffix removal not robust enough
# Issue
What should have been "known" suffixes are still not getting removed.

# Implementation
- [x] bump version
- [x] tests
- [x] code
    - [x] Add suffix generation code
    - [x] Update `remove_suffix` to use the above results.
- [x] tests
- [x] bump version

stscieisenhamer:
Resolves #1870 
Developed a more complete list and a utility to rebuild that list.

Later, the list generation could be done dynamically, such as on package loading, if deemed too much a burden on maintenance. unit testing will currently provide the check on whether updates are necessary.


jwst/stpipe/suffix.py
```diff
@@ -0,0 +1,386 @@
...
+logger.addHandler(logging.NullHandler())
+
+# Suffixes that are hard-coded or otherwise
+# have to exist. Used by `find_suffixes` to
+# add to the result it produces.
+SUFFIXES_TO_ADD = [
+    'cal', 'calints', 'crf', 'crfints',
+    'dark',
+    'i2d',
+    'jump',
```
hbushouse: I'm not aware of `jump` being one of the regular product types. If someone saves the corrected results from the calwebb_detector1 pipeline, they get a `ramp` product, which is the state of the data immediately following the `jump` step. But it's called `ramp`, not `jump`.

jwst/stpipe/suffix.py
```diff
@@ -0,0 +1,386 @@
...
+
+# Configure logging
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+
+# Suffixes that are hard-coded or otherwise
+# have to exist. Used by `find_suffixes` to
+# add to the result it produces.
+SUFFIXES_TO_ADD = [
+    'cal', 'calints', 'crf', 'crfints',
```
hbushouse: I don't see the standard AMI products in the list, which include `ami`, `amiavg`, and `aminorm`.

jwst/stpipe/suffix.py
```diff
@@ -0,0 +1,386 @@
...
+    'dark',
+    'i2d',
+    'jump',
+    'psfalign', 'psfstack', 'psfsub',
+    'ramp', 'rate', 'rateints',
+    's2d', 's3d',
+    'uncal',
+    'wfscmb',
+    'x1d', 'x1dints',
+]
```
hbushouse: Should perhaps also add the various flavors of photometry and source catalog product types, including `cat`, `phot`, and `whtlt`.

stscieisenhamer: and i did a more thorough search and found a couple more.

jwst/stpipe/suffix.py
stscieisenhamer: However, note that at some point, a step is called `jump` and hence there is a chance of having that as a suffix.  Regardless, that is found in the searching for step names and should not be in the explicit list.

stscieisenhamer:
@hbushouse pushed suffix changes as suggested.

PR #1885, opened 2018-04-06
WIP: add nightly pandokia tests

nden:
TODO:

- [ ] as a start switch the nightly pandokia tests to use github 
- [ ] choose tests to be run on jenkins
- [ ] setup jenkins tests
- [ ] setup the nightly tests to run using jenkins on jwcalibdev

Will need help from @jhunkeler or @rendinam


stscicrawford:
Sorry for not clarifying before,  my primary intention was to ask to have the tests migrated from svn to a github repository, and a small update to pandokia to run the tests from the new github repository rather than from the svn repository.     For right now, I think having these as their own github repository would work best as there still needs to be some planning and discussion with how the new CI testing with Jenkins will work (and there are other priorities in the path right now).   As the format might change, I would hold off on trying to integrate these directly into the main repository (unless @pllim @jhunkeler @rendinam disagree).

jwst/tests_nightly/general/fgs/test_guider_pipeline_1.py
```diff
@@ -0,0 +1,40 @@
...
+    href = pf.open( n_ref )
+    newh = pf.HDUList([h['primary'],h['sci'],h['dq']])
+    newhref = pf.HDUList([href['primary'],href['sci'],href['dq']])
+    result = pf.diff.FITSDiff(newh,
+                              newhref,
+                              ignore_keywords = ['DATE','CAL_VER','CAL_VCS','CRDS_VER','CRDS_CTX'],
+                              rtol = 0.000001
+    )
+
+    print (' Fitsdiff comparison between the calibrated product file - a:', n_cr )
```
pllim: Although for `pytest` reports, `print` functions are unnecessary. Everything you need to know when a test fails can be moved to the assertion error message. Such a change is trivial.

jhunkeler:
@nden @stscicrawford @pllim 

Just recording a few TODO items I thought about while working on this today:

- [ ] Add `import pytest` to each `test_*.py` script
- [ ] Get rid of the `try`/`except` blocks wrapping every assert call. (This was a workaround for a bug in pandokia -- it can go away)
- [ ] Replace `TEST_BIGDATA+'/long/path/to/data.ext'` with a function or write a data retrieval class to give us some uniformity/compatibility across tests.
- [ ] Make **_heavy_** use of pytest's `tmpdir` to avoid output cluttering up the project root

pllim:
I agree with all the suggestions above. :+1:

pllim:
This PR is integrated to, and hence, superseded by #1888. FYI.

Issue #1360, opened 2017-10-24
Update MIRI RSCD correction


jemorrison:
The MIRI RSCD corrections needs to be updated to be in line with the recommended MIRI team algorithm,

jemorrison:
Add saturation checking and modification to the Counts in the last frame in the last integration if the data is saturated. A different algorithm is used to find the scale factor for the RSCD correction

jemorrison:
@stscicrawford

hbushouse:
Since merging https://github.com/STScI-JWST/jwst/pull/1886 all MIRI regression tests that execute the RSCD step are throwing an error, because the updated code is getting old reference files. Is there an ETA on when the new form of the RSCD ref files will be delivered to CRDS?

stscicrawford:
@hbushouse  please see https://jira.stsci.edu/browse/CRDS-190 and similar Jira tickets.  We can raise this as an issue to discuss this afternoon to check on the progress of it.

PR #1886, opened 2018-04-06
Updated RSCD correction and docs

jemorrison:
Updated RSCD correction algorithm. I updated some of the documentation. 
This update requires a new reference file. Misty Cracraft is going to get them into CRDS but
this might take a little time. So we need to be careful when this is merged.

### DIFF
--- a/docs/jwst/rscd/reference_files.rst
+++ b/docs/jwst/rscd/reference_files.rst
@@ -6,6 +6,14 @@ The correction to be added to the input data has the form::
     corrected data = input data data + dn_accumulated * scale * exp(-T / tau)
 
 where, dn_accumulated is the DN level that was accumulated for the pixel from the previous integration. 
+In case where the dn_accumulated does not saturation the scale factor is determined as follows:
+       :math:`scale = b{1}* [Counts{2}^b{2} * [1/exp(Counts{2}/b{3}) -1]`
+    where :math:'b{1} = ascale * (illum_zpt + illum_slope*N + illum2* N^2` (N is the number of groups per integration)
+    The :math:`Counts{2}` = Final DN in last frame in last integration - Crossover Point
+If previous integration saturates, the the scale factor is modified in the following manner:
+   :math:` scale_\text{sat} = slope * Counts{3} + sat_text{mzp}`, where :math:`Counts{3}' = extrapolated counts past 
+   saturation * sat_scale and :math: 'slope = sat_zp + sat_slope * N + sat_2*N^2 + evenrow_corrections'
+
 
 CRDS Selection Criteria
 -----------------------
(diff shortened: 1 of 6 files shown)


docs/jwst/rscd/reference_files.rst
```diff
@@ -1,34 +1,55 @@
...
 -----------------------
 RSCD reference files are selected on the basis of INSTRUME and DETECTOR
-values for the input science data set.  The referene file for each detector is a table of values based on
+values for the input science data set.  The reference file for each detector is a table of values based on
 READPATT (FAST, SLOW) , SUBARRAY (FULL or one the various subarray types) , and ROWS type (even or odd row).
-The correction values, tau and scale, are read in separately for even and odd rows for the readout pattern and  
+The fourtenn correction values are read in separately for even and odd rows for the readout pattern and  
 if it is for the full array or one of the imager subarrays. The table actually contains the parameters for a 
-double-exponential function, but currently  only the single exponential values are used. 
+the non-saturating data and saturating data all in one row.
```
hbushouse: seems to be a discontinuity between the end of line 26 and beginning of line 27

jwst/rscd/rscd_step.py
```diff
@@ -32,6 +31,18 @@ def process(self, input):
                     result.meta.cal_step.rscd = 'SKIPPED'
                     return result
 
+
+                # Check that data has the minimum number of groups/int
+                sci_ngroups = input_model.data.shape[1]     # number of groups
+                min_number = 4
+                if sci_ngroups < min_number :
+                    self.log.warning('Input file does not contain enough groups '
+                                     ' for RSCD correctionto be applied ')
```
hbushouse: don't need whitespace at the end of line 39 and at the beginning of line 40 (can remove one of them)

jwst/rscd/rscd_step.py
```diff
@@ -32,6 +31,18 @@ def process(self, input):
...
+                    self.log.warning('RSCD step will be skipped')
+                    result = input_model.copy()
```
hbushouse: In cases where a step is being skipped and hence the data aren't being modified in any way other than perhaps setting the cal_step meta attribute, I suggest NOT making a copy and just returning the input in place. I.e. just set input_model.meta.cal_step.rscd and return input_model. No point in making a copy of an unchanged data model.

jwst/rscd/rscd_sub.py
```diff
@@ -135,58 +250,136 @@ def get_rscd_parameters(input_model, rscd_model):
...
+    saturated = np.full((nrows,ncols),False)
+    is_ref = saturated.copy()
+
+    diff = dn_lastframe2 - dn_lastframe3
+    dn_lastframe23 = dn_lastframe2 + diff
+    # check if pixel is saturated
+    saturated_flag = datamodels.dqflags.group['SATURATED']
+    ref_flag = datamodels.dqflags.pixel['REFERENCE_PIXEL']
+    
     for j in range(nrows):
```
hbushouse: Someday it would be nice to find a way to do this without the looping over pixels, which can be really slow in python. Especially because it's being done for every integration. But if it works as-is, without being ridiculously slow, let's leave it for the future.

hbushouse:
Fixes #1360

PR #1887, opened 2018-04-10
update to stray-light algorithm

### LINKED ISSUE #1873: straylight documentation
In https://jira.stsci.edu/browse/JP-243:

The current documentation for the straylight correction needs to be updated with the method provided by the MIRI team with CDP5 delivery

@hbushouse @jemorrison

jemorrison:
Updated new straylight algorithm. This closes #1873

### DIFF
--- a/docs/jwst/straylight/arguments.rst
+++ b/docs/jwst/straylight/arguments.rst
@@ -1,4 +1,43 @@
 Step Arguments
 ==============
 
-The straylight correction has no step-specific arguments.
+There are two possible algorithms  to use for the stray-light correction step. The first one is more 
+simplistic and uses a row-row interpolation of the gap pixels to determine the
+stray-light correction. The second algorithm uses a 2-D approach by using a Modified
+Shepard's Method to interpolate the light in the gap pixels. The default algorithm 
+is to use the second method. The first method was  kept for comparison to  the second
+method and to confirm the second method works well in the pipeline.
+
+The argument which sets which algorithm to use is
+ * ``--method [string]``
+
+The default Modified Shepard's Method has the value ModShepard. To set the step to use
+the simplistic row-row interpolate use Nearest.
+
+There are two arguments if the Modified Shepard's Method is being used. These are
+
+* ``--roi [integer]``
+
+This parameter sets the 'radius of influence' to select the gap pixels to be used
+in the correction. The default value is set to 50 pixels. 
+
+*``power [integer]`` 
+
+This parameter is the power :math:`k` in the Modified Shepard's Method weighting
+ equation.
+
+The usage of both parameters are shown in the description of the 
+Modified Shepard’s Method distance weighting equation:
+
+.. math::
+
+   s = \frac{ \sum_{i=1}^n p_i w_i}{\sum_{i=1}^n w_i}
+
+where,
+.. math::
+
+   w_i =\frac{ max(0,R-d_i)} {R d_i}^ k
+
+The radius of influence :math: 'R' and the exponent :math: 'k' are variables that 
+can be adjusted to the actual problem. The default values for these parameters are
+:math:`R = 50` pixels and :math:`k = 1`.
--- a/docs/jwst/straylight/arguments.rst
+++ b/docs/jwst/straylight/arguments.rst
@@ -9,10 +9,11 @@ is to use the second method. The first method was  kept for comparison to  the s
 method and may be removed in a future version. 
 
 The argument which sets which algorithm to use is
- * ``--method [string]``
 
-The default Modified Shepard's Method has the value ModShepard. To set the step to use
-the simplistic row-row interpolate use Nearest.
+* ``--method [string]``
+
+The default Modified Shepard's Method has the value 'ModShepard'. To set the step to use
+the simplistic row-row interpolate use 'Nearest'.
 
 There are two arguments if the Modified Shepard's Method is being used. These are
 
@@ -21,10 +22,10 @@ There are two arguments if the Modified Shepard's Method is being used. These ar
 This parameter sets the 'radius of influence' to select the gap pixels to be used
 in the correction. The default value is set to 50 pixels. 
 
-*``power [integer]`` 
+* ``--power [integer]`` 
 
 This parameter is the power :math:`k` in the Modified Shepard's Method weighting
- equation. The default value is set to 1. 
+equation. The default value is set to 1. 
 
 The usage of both parameters are shown in the description of the 
 Modified Shepard’s Method distance weighting equation:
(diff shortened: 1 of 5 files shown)


docs/jwst/straylight/arguments.rst
```diff
@@ -1,4 +1,43 @@
...
+There are two arguments if the Modified Shepard's Method is being used. These are
+
+* ``--roi [integer]``
+
+This parameter sets the 'radius of influence' to select the gap pixels to be used
+in the correction. The default value is set to 50 pixels. 
+
+*``power [integer]`` 
+
+This parameter is the power :math:`k` in the Modified Shepard's Method weighting
```
hbushouse: should mention what the default value is

docs/jwst/straylight/main.rst
```diff
@@ -19,20 +15,57 @@ in the MRS slice
...
+from about 1 % in Channel 1A to undetectable low levels long-ward of Channel 2B.
+Two components of the stray-light have been observed, a smooth and a structured
+distribution. 
 
 Algorithm
 ---------
+The basic idea of the stray-light removal algorithm is to deal with the 
+smooth component of the stray-light only. Due to the extended nature of the
+stray-light we use the detected signal in the slice gaps, where nominally no photons
+should it the detectors, and assume that all detected light is the stray-light.
```
hbushouse: I guess "it" should be "**h**it"?

docs/jwst/straylight/main.rst
```diff
@@ -19,20 +15,57 @@ in the MRS slice
...
+There are two possible algorithms in the stray-light step. The first algorithm is
+a more simplistic approach by dealing with the stray-light estimation row-by-row
+and interpolating the gap flux linearly. An intermediate stray-light map is 
+generated row-by-row and then this map is further smoothed to remove row-by-row
+variations. This algorithm uses a stray-light mask reference file that contains
+1s for gap pixels and 0s for science pixels.  
+
+Given the extended nature of the smooth component of the MRS stray-light, it
+is obvious that a row-by-row handling of the stray-light could be replaced
+by a two-dimensional approach such that not additional smoothing is required.
```
hbushouse: "such that not additional" should be "such that **no** additional"

docs/jwst/straylight/main.rst
```diff
@@ -19,20 +15,57 @@ in the MRS slice
...
+For the second algorithm we improve the technique by using the Modified Shepard's
+Method to interpolate the gap fluxes two dimensionally. 
+
+The second algorithm uses the MRS distortion reference file. The first extension
+of this file contains the "Slice_Number" and consists of a detector pixel map 
+indicating the slice numbers or in case of a non-science pixel, the value zero. 
+Consequently, the gap pixels can be easily found by masking all pixels that are 
+non-science.The assign_wcs step takes the "Slice_Number" information and stores it in
```
hbushouse: missing a white space between sentences

docs/jwst/straylight/main.rst
```diff
@@ -19,20 +15,57 @@ in the MRS slice
...
+a regions file.  This second algorithm takes each slice pixel and determines the
```
hbushouse: where is the regions file? Is it really a file (on disk) that's generated by assign_wcs or is it just an additional data structure of some kind added to the data model?

docs/jwst/straylight/reference.rst
```diff
@@ -1,20 +1,25 @@
 Reference File
 ==============
-The MIRI MRS straylight correction step uses a straylight mask. There
+The default MIRI MRS stray-light correction step uses Regions reference file. This
```
hbushouse: "uses **THE** Regions reference file"

docs/jwst/straylight/reference.rst
```diff
@@ -1,20 +1,25 @@
...
+Regions files is the first extension in the MIRI Distortion file and has  been read
```
hbushouse: Oh wait, so it's not a separate reference file, but is actually just information in the first extension of the Distortion reference file? In that case, you should not say that this step uses the Regions reference file, but instead it uses the Distortion reference file.

docs/jwst/straylight/reference.rst
```diff
@@ -1,20 +1,25 @@
...
+in and pulled into a separate reference type by the assign_wcs step.
```
hbushouse: How can the assign_wcs step be creating new reference files on the fly while the pipeline is running? This whole section on Regions vs Distortion vs assign_wcs creating things is confusing, at least to me.

docs/jwst/straylight/main.rst
```diff
@@ -68,4 +63,14 @@ where,
 
 The radius of influence :math: 'R' and the exponent :math: 'k' are variables that 
 can be adjusted to the actual problem. The default values for these parameters are
-:math:`R = 50` pixels and :math:`k = 1`.
\ No newline at end of file
+:math:`R = 50` pixels and :math:`k = 1`.
+
+
+The second algorithm uses a 'regions' reference file type that is stored in
```
hbushouse: We really can't be referring to the regions information as a reference file type. A reference file is a physical file that's stored in and retrieved from CRDS and gets fetched when the step runs. To me it sounds like the regions information may originally come from one of the official CRDS reference files (the distortion ref file, the actual regions ref file, ?) that gets loaded during the execution of the assign_wcs step and assign_wcs is then storing that information as part of the rest of the gwcs object that it creates (and stores in the ASDF extension)? So the info is available later for this step to use. If that's the case, then that's how it should be explained in the docs. So while the straylight step may not fetch and load the regions reference file directly, it is dependent upon having previously run the assign_wcs step to do that loading, so the info is available for the straylight step.

docs/jwst/straylight/main.rst
```diff
@@ -68,4 +63,14 @@ where,
...
+the ASDF extension of the input model. The 'regions' information is a 1032 by 1024
+mask containing the 'slice number' of the pixel or in the case of  non-science
+gap pixels, the value zero. The assign_wcs step creates the 'regions' information
+by reading in the MIRI MRS distortion file. The first extension
```
hbushouse: Ah, OK, so it sounds like my assumption was correct. Here's the info that verifies that. I just think the flow would be better if the origin of the info were described first, and then describe its format and how it's used. E.g. something along the lines of "The straylight step uses information from the `regions` reference file that was previously loaded into the meta data of the input image by the assign_wcs step. Hence running the assign_wcs step is a prerequisite to the straylight step. The regions information is a 1032 x 1024 ...."

jemorrison:
closes #1873

jemorrison:
@nden I added you to review this code because I now read in the slice regions from the meta data that
assign_wcs fills in - instead of the "regions" reference file.

nden:
@jemorrison The call to get the slices array is correct.

PR #1888, opened 2018-04-10
Regtest POC

### LINKED ISSUE #1208: Move pandokia test scripts back here
This would be needed for integration with Jenkins. Obviously, you also have to rewrite it so it does not rely on pandokia anymore. I recommend you add some kind of pytest fixture to them (akin to what you did to the "slow tests") so they won't run on Travis but will run on Jenkins.

c/c @jhunkeler 

Supersede #674

jhunkeler:
Based on #1885 

- Controlled by an internally facing Jenkins instance.
  - https://boyle.stsci.edu:8081/job/RT/job/JWST/job/regtest-poc
  - Click "Test Result Analyzer" on the left navigation bar
  - OR select a job in the "Build History" then click  "Test result" on the left navigation bar
- Executes tests on the same hardware as pandokia.
- Mildly exciting.

EDIT: Fix #1208 , close #1885


nden:
@hbushouse Those tests were changed to pass the non-default parameters directly to the step or set them (in the case of a pipeline). Here's an [example](https://github.com/STScI-JWST/jwst/pull/1888/commits/3a25617786d34911f222d5b13e6574e387060057#diff-69863ad1a27948124a9b585408d69a63R14).

pllim:
Re: squash and merge -- If you use the squash and merge option, Nadia might lose contribution credit?

pllim:
As requested, here are comments from @hbushouse on Apr 12 regarding the failures (some might have since been fixed):

All of the missing configuration file errors are due to the fact that the aforementioned configuration files were not copied over from the svn repo to the github repo. Nadia made the conscious decision to no longer use them, but that requires updating most/all of the test modules to remove the references to them.

Some of the CRDS errors that say something like "jwst_miri_cubepar_0001.fits is marked as scientifically bad" indicates that an old CRDS context is in use that's in turn trying to use old reference files. Those reference files were replaced in context 0438 and later. How is CRDS_CONTEXT being set (if at all) in the new Jenkins test environment? If it's being set manually somewhere, it's obviously pegged to an old value and needs to be updated (something like 0445).

The more recent results that still show vast numbers of CRDS errors again suggests to me that some horribly outdated version of the CRDS context is in use (e.g. all the errors involving "wavecorr" as an unknown reference file type). Or it's using some really old instance of the CRDS server or something. Todd is the CRDS person.

All of the errors like:
```
E       FileNotFoundError: [Errno 2] No such file or directory: 'flatfield1_output_flat_field.fits'
```

indicate that it can't find the expected output file produced by the test. In the old environment the output files went to the working directory, where the test module itself is located and run from. Are the output files going somewhere else now? If so, all the test modules have to be updated to look for them in the right place.

jhunkeler:
Do you think the "co-authors" feature still works with squashed commits? https://help.github.com/articles/creating-a-commit-with-multiple-authors/

pllim:
If you want to be adventurous, you can squash them such that this PR still has multiple authors. You just can't use the convenient "squash and merge" button. GitHub can show "x committed with y" but git log only shows the primary AFAIK.

jwst/tests_nightly/general/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+import pytest
+import re
+import requests
+
+
+RE_URL = re.compile('\w+://\S+')
+
+
+def pytest_addoption(parser):
+    parser.addoption('--bigdata', action='store_true',
```
pllim: As discussed offline, it might be a good idea to just have one `conftest` for all tests, regardless of Travis or Jenkins.

jwst/tests_nightly/general/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+@pytest.fixture
+def _bigdata():
+    """ Return path to large data sets
+
+    Note: Support for URLs added for future integrations
+    """
+    origins = [
+        os.environ.get('TEST_BIGDATA', ''),
```
pllim: Can't we just do `path = os.environ.get('TEST_BIGDATA', '/data4/jwst_test_data')` and just test for it without the following loop?

jwst/tests_nightly/general/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+
+    for path in origins:
+        if os.path.exists(path) or check_url(path):
+            return path
+
+    raise BigdataError('Data files are not available.')
+
+
+@pytest.fixture(scope='function')
+def _jail(tmpdir):
```
pllim: For future maintenance by someone else, a docstring here would be nice.

jwst/tests_nightly/general/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+    path = str(tmpdir)
+    os.chdir(path)
```
pllim: `os.chdir(tmpdir.strpath)` should work as well.

jwst/tests_nightly/general/fgs/test_guider_pipeline.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def test_guider_pipeline1(_bigdata):
+    """
+
+    Regression test of calwebb_guider pipeline performed on ID-image data.
+
+    """
+
+    GuiderPipeline.call(_bigdata+'/fgs/test_guiderpipeline/jw88600073001_gs-id_7_image-uncal.fits',
```
pllim: If there is any chance you would want to test under Windows, using `os.path.join` to string together (pun!) path name might be safer. But that can be future work in separate PR, if needed.

jwst/tests_nightly/general/fgs/test_guider_pipeline.py
```diff
@@ -0,0 +1,89 @@
...
+                        output_file='jw88600073001_gs-id_7_image-cal.fits')
+
+    # Compare calibrated ramp product
+    n_cr = 'jw88600073001_gs-id_7_image-cal.fits'
+    h = pf.open( n_cr )
```
pllim: Nitpick: Would prefer `with pf.open(n_cr) as h` context manager. But this can probably be addressed in a future sprint or something.

In addition, the `fitsdiff` logic seems repetitive across the board. Perhaps that whole logic block can be turned into an utility function in the future to minimize code duplication.

jwst/tests_nightly/general/fgs/test_guider_pipeline.py
```diff
@@ -0,0 +1,89 @@
...
+    pytest.mark.usefixtures('_jail'),
+    pytest.mark.skipif(not pytest.config.getoption('bigdata'),
+                       reason='requires --bigdata')
+]
+
+
+def test_guider_pipeline1(_bigdata):
+    """
+
+    Regression test of calwebb_guider pipeline performed on ID-image data.
```
pllim: Future work: In the spirit of documenting tests, as brought up by @mdlpstsci , developers should have more details on what this test is supposed to accomplish.

jwst/tests_nightly/general/framework/test_self.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+pytestmark = [
+    pytest.mark.usefixtures('_jail'),
+    pytest.mark.skipif(not pytest.config.getoption('bigdata'),
+                       reason='requires --bigdata')
+]
+
+
+def test_bigdata(_bigdata):
```
pllim: Please document these tests in this PR, as they actually test the new infrastructure.

jwst/tests_nightly/general/fgs/test_guider_pipeline.py
```diff
@@ -0,0 +1,89 @@
+import os
+import pytest
+from astropy.io import fits as pf
+from jwst.pipeline.calwebb_guider import GuiderPipeline
+
+
+pytestmark = [
+    pytest.mark.usefixtures('_jail'),
```
pllim: Such usage of `_jail` can be good or bad. For instance, if you want to add a new test in this file that does not want that functionality, it would be impossible. It is also hard to read for someone that is not familiar with `pytest`. I understand that this is necessary now because of the requirement to not change ported tests too much. However, I recommend a rewrite going forward, so things are more explicit.

For example, in spacetelescope/hstcal#71, you can explicitly mark a base class however you like and then have tests inherit from the base class as needed. However, that is not the only solution available.

jwst/tests_nightly/general/miri/test_bias_drift2.py
```diff
@@ -0,0 +1,41 @@
...
+def test_refpix_miri2(_bigdata):
+    """
+
+    Regression test of refpix step performed on MIRI data.
+
+    """
+    output_file_base, output_file = add_suffix('refpix2_output.fits', 'refpix')
+
+    try:
+        os.remove(output_file)
```
pllim: I would argue that this is no longer necessary given that `pytest` runs each test in its own `tmpdir` sub-directory.

jwst/tests_nightly/general/miri/test_miri_fixed_wcs.py
```diff
@@ -0,0 +1,44 @@
...
+
+    input_file = os.path.join(_bigdata, 'miri', 'test_wcs', 'fixed', 'jw00035001001_01101_00001_mirimage_rate.fits')
+    ref_file = os.path.join(_bigdata, 'miri', 'test_wcs', 'fixed', 'jw00035001001_01101_00001_mirimage_assign_wcs.fits')
+
+    AssignWcsStep.call(input_file,
+                       output_file=output_file_base
+                       )
+    im = ImageModel(output_file)
+    imref = ImageModel(ref_file)
+    y, x = np.mgrid[:1031, :1024]
```
pllim: Magic numbers!

jwst/tests_nightly/general/miri/test_miri_ifu_wcs.py
```diff
@@ -0,0 +1,72 @@
+import os
+import pytest
+
+import numpy as  np
+from numpy.testing import utils
```
pllim: This won't work with Numpy 1.15+. Replace all `from numpy.testing import utils` with simply `from numpy.testing import your_assert_function`. Same deal with `from numpy.testing import decorators`, if you use it.

jwst/tests_nightly/general/miri/test_miri_slitless_wcs.py
```diff
@@ -0,0 +1,44 @@
...
+
+    input_file = os.path.join(_bigdata, 'miri', 'test_wcs', 'slitless', 'jw80600012001_02101_00003_mirimage_rateints.fits')
+    ref_file = os.path.join(_bigdata, 'miri', 'test_wcs', 'slitless', 'jw80600012001_02101_00003_mirimage_assign_wcs.fits')
+
+    AssignWcsStep.call(input_file,
+                       output_file=output_file_base
+                       )
+    im = CubeModel(output_file)
+    imref = CubeModel(ref_file)
+    x, y = np.mgrid[:1031, :1024]
```
pllim: Repetitive magic numbers!

jwst/tests_nightly/general/miri/test_ramp_fit2.py
```diff
@@ -0,0 +1,75 @@
...
+                       reason='requires --bigdata')
+]
+
+def test_ramp_fit_miri2(_bigdata):
+    """
+
+    Regression test of ramp_fit step performed on MIRI data.
+
+    """
+    output_file_base, output_files = add_suffix('rampfit2_output.fits', 'rampfit', list(range(2)))
```
pllim: `list(range(2))` seems like an overkill when all you really want is `[0, 1]`.

jwst/tests_nightly/general/nircam/#todo#
```diff
@@ -0,0 +1,13 @@
+strun assign_wcs.cfg wfs_in_1.fits
```
pllim: Um... Should we be checking in `#todo#`? This does not seem legit. Is this a legit test file, or just some temporary file that no one cleaned up?

jwst/tests_nightly/general/nircam/out.wfs
```diff
@@ -0,0 +1,203 @@
+
+
+START
```
pllim: Why are we checking in old `pandokia` log?

jwst/tests_nightly/general/niriss/test_ramp_fit.py
```diff
@@ -0,0 +1,63 @@
...
+]
+
+
+def test_ramp_fit_niriss(_bigdata):
+    """
+
+    Regression test of ramp_fit step performed on NIRISS data.
+
+    """
+    output_file_base, output_files = add_suffix('rampfit_output.fits', 'rampfit', list(range(1)))
```
pllim: Isn't `list(range(1))` just `[0]`?

jwst/tests_nightly/general/nirspec/test_extract2d_msa_8slits.py
```diff
@@ -0,0 +1,66 @@
...
+    except:
+        raise OSError("Could not copy inputs files")
+
+    Extract2dStep.call('jw00038001001_01101_00001_NRS1_assign_wcs.fits',
+                       output_file='extract2d2_output.fits'
+                       )
+    h = fits.open('extract2d2_output.fits')
+    href = fits.open(_bigdata+'/nirspec/test_extract_2d/msa/jw00038001001_01101_00001_NRS1_assign_wcs_extract_2d.fits')
+    '''
+    newh = fits.HDUList([h['primary'],h['sci',1],h['err',1],h['dq',1],
```
pllim: This is a question for the person who wrote that test originally -- Why the commented code block? If it is not used going forward, it is less confusing to simply delete it.

jwst/tests_nightly/general/nirspec/test_nrs1_fs_wcs.py
```diff
@@ -0,0 +1,55 @@
...
+
+    input_file = os.path.join(_bigdata, 'nirspec', 'test_wcs', 'nrs1-fs', 'jw00023001001_01101_00001_NRS1_ramp_fit.fits')
+    ref_file = os.path.join(_bigdata, 'nirspec', 'test_wcs', 'nrs1-fs', 'jw00023001001_01101_00001_NRS1_assign_wcs.fits')
+
+    AssignWcsStep.call(input_file,
+                       output_file=output_file_base
+                       )
+    im = ImageModel(output_file)
+    imref = ImageModel(ref_file)
+    #ystart = im.meta.subarray.ystart
```
pllim: Again, please consider deleting dead code.

jwst/tests_nightly/general/nirspec/test_nrs1_ifu_wcs.py
```diff
@@ -0,0 +1,53 @@
...
+
+    """
+    output_file_base, output_file = add_suffix('nrs1_ifu_wcs_output.fits', 'assignwcsstep')
+
+    try:
+        os.remove(output_file)
+    except:
+        pass
+
+    input_file = os.path.join(_bigdata, 'nirspec', 'test_wcs', 'nrs1-ifu', 'jw00011001001_01120_00001_NRS1_rate.fits')
```
pllim: Here's a nice example of `os.path.join` usage I mentioned above. Kudos to this one. :clap:

jwst/tests_nightly/general/pipelines/test_mrs_spec3.py
```diff
@@ -0,0 +1,98 @@
...
+from jwst.pipeline.calwebb_spec3 import Spec3Pipeline
+
+pytestmark = [
+    pytest.mark.usefixtures('_jail'),
+    pytest.mark.skipif(not pytest.config.getoption('bigdata'),
+                       reason='requires --bigdata')
+]
+
+
+@pytest.mark.skipif(not pytest.config.getoption('--runslow'),
```
pllim: JWST devs might want to check why this one test takes over 4 hours, and if it cannot be scaled down to perform faster.

jwst/tests_nightly/general/pipelines/test_nrc_image3_1.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Regression test of calwebb_image3 pipeline on NIRCam
+    simulated long-wave data.
+
+    """
+    subdir = os.path.join(_bigdata, 'pipelines', 'nircam_calimage3')
+    ignore_kws = ['DATE', 'CAL_VER', 'CAL_VCS', 'CRDS_VER', 'CRDS_CTX']
+
+    try:
+        os.remove("nrca5_47Tuc_subpix_dither1_newpos_a3001_crf.fits")
```
pllim: I am guessing these are output files?

jwst/tests_nightly/general/pipelines/test_nrc_tso3_1.py
```diff
@@ -0,0 +1,195 @@
+import os
+import pytest
+from astropy.io import fits as pf
+from jwst.pipeline.calwebb_tso3 import Tso3Pipeline
+# DISABLED
+# import pandokia.helpers.filecomp as filecomp
```
pllim: Delete this! Exterminate!

jwst/tests_nightly/general/pipelines/test_nrc_tso3_1.py
```diff
@@ -0,0 +1,195 @@
...
+    Default imaging mode outlier_detection will be tested here.
+    """
+    testname = "test_tso3_pipeline1"
+    # Define where this test data resides in testing tree
+    subdir = 'pipelines/nircam_caltso3/'  # Can NOT start with path separator
+
+    # You need to have a tda dict for:
+    #  - recording information to make FlagOK work
+    #  - recording parameters to the task as attributes
+    global tda
```
pllim: Maybe future work, but the tests here need some clean up to remove unnecessary `pandokia` artifacts.

jwst/tests_nightly/general/pipelines/test_nrc_tso3_1.py
```diff
@@ -0,0 +1,195 @@
...
+    result = perform_FITS_comparison(fname, refname, extn_list=extn_list)
+    assert result.identical, result.report()
+
+    # compare the output files - use this exact command
+    # DISABLED
+    # filecomp.compare_files(output, (__file__, testname), tda=tda,)
+
+
+# Utility function to simplify FITS comparisons
+def perform_FITS_comparison(filename, refname, **pars):
```
pllim: Ah, looks like you already have that `fitsdiff` helper I mentioned earlier in this review. So, I guess, the work now is simply to get all the other tests to use this, as applicable.

JenkinsfileRT
```diff
@@ -0,0 +1,74 @@
...
+                  photutils \
+                  scipy \
+                  six \
+                  spherical-geometry \
+                  stsci.image \
+                  stsci.imagestats \
+                  stsci.stimage \
+                  stsci.tools \
+                  verhawk"
+def CONDA_CHANNEL = "http://ssb.stsci.edu/conda-dev"
```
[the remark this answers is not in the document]
jhunkeler: There's really nothing stopping anyone from using the public channel, rather than dev, then filling in the blanks with `pip install -q [git+https://github.com/some/repo.git#egg=name]` for each bleeding edge package they require.

Unfortunately time will be lost compiling the dependencies every single time this runs.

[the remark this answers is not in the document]
pllim: > each bleeding edge package they require

Hopefully not that much? Usually it is just Astropy, right? I mean, you can't possibly deliver dev of everything to DMS, can you? I just thought that testing what you actually want to deliver might be more useful.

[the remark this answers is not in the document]
jhunkeler: DMS receives `master` for almost every dependency of the pipeline right now. That's going to change eventually, however at the moment you can't run the pipeline with the stable/public releases of most packages. `jwst` inevitably requires  _[unreleased Astropy feature here]_ used by _[development code in another package there]_ so distributing that dependency to the public is pretty difficult. 

I'd like to figure out exactly which dependencies are doing this. ASDF tracks changes made upstream in Astropy very closely, and maybe the same can be said for WCS-related packages... As for others I'm not 100% sure. (< This sounds like it needs to be in its own issue.)

jwst/tests_nightly/general/conftest.py
jhunkeler: Agreed. I'll likely consolidate this into the main `conftest` tonight or tomorrow.

jwst/tests_nightly/general/conftest.py
jhunkeler: Yeah, it won't hurt to set the default path within `.get`. As for the list and loop -- The list will eventually contain a URL pointing back to Artifactory. The `check_url` function determines whether the remote resource is available or not. Another (unwritten) function in the test framework uses `_bigdata` to copy or download the data to the local directory.

In other words this should allow you to transparently run the test suites on your desktop at work, or directly on the test system. Let's revisit this when I'm done with the second half of it. I must have gotten sidetracked and never finished.

jwst/tests_nightly/general/conftest.py
jhunkeler: Thanks, I'll do that.

jwst/tests_nightly/general/fgs/test_guider_pipeline.py
jhunkeler: All of these `thing + "stuff"` concatenations are going away, don't worry :smile: 
Note: I didn't bother changing any of these because I anticipated replacing all of them with something more robust.

jwst/tests_nightly/general/fgs/test_guider_pipeline.py
jhunkeler: Agreed. I originally intended to replace these redundant `fitsdiff` calls with a generic wrapper, but I was told not to change the tests (too much) so I left them alone.

(This will probably need its own issue)

jwst/tests_nightly/general/framework/test_self.py
jhunkeler: Will do

jwst/tests_nightly/general/fgs/test_guider_pipeline.py
jhunkeler: I definitely agree with you. If/when tests are consolidated to single files and further broken down by test classes, I'd like to see something like `_jail` used as-needed, rather than forced.

Just something to keep in mind though... If you *don't* jail your test, or use `tmpdir` on your own, the directory where `pytest` was executed will become a minefield of input and output data. If those files are not in `.gitignore` there's a greater chance you will accidentally commit the data to the repository (this is especially true for anyone that relies on `git commit -a`).

jwst/tests_nightly/general/nircam/#todo#
jhunkeler: I'll delete it.

jwst/tests_nightly/general/nircam/out.wfs
jhunkeler: I'll delete this too

jwst/tests_nightly/general/fgs/test_guider_pipeline.py
jhunkeler: Nitpick+1: Is there a good reason to continue importing this as `pf` (pyfits)?
Why not `from astropy.io import fits` then `with fits.open(...) as h`?

jwst/tests_nightly/general/pipelines/test_nrc_image3_1.py
hbushouse: Yes, many of the tests remove old versions of their output files before running. That way you know you're starting cleanly. There are occasions where a change in behavior causes a step or test to put its output in a file of a different name, so if the old one is still there, it gets used when doing the comparison and you don't know that something's changed.

jwst/tests_nightly/general/fgs/test_guider_pipeline.py
pllim: Yes, that is a risk. We can certainly defer this to a separate PR.

jwst/tests_nightly/general/pipelines/test_nrc_image3_1.py
pllim: OK, but as stated above, this is no longer necessary with `pytest + tmpdir` usage.

pllim:
OK, Joe. I moved most of my comments that is out of scope here over to #1913

pllim:
p.s. Would be nice to have a link to the cron/nightly builds on README on this PR or after this is merged.

jhunkeler:
@pllim URL added to main PR body.

jhunkeler:
@jdavies-st Here's what #1939 produced...

:tada: 

chart

screenshot_2018-04-24_21-09-31

@hbushouse, @stsci-hack indicated that in order for `general.pipelines` tests to pass he needs to copy reference files and/or data from the old SVN directories into the `TEST_BIGDATA` tree. I'm assuming the `+3` failures introduced during test run `#39`  may be the result of some cleanup on his part. 

Note1: #1939 changed the `CRDS_CONTEXT` from `445` to `452`. One or more new failures could be related to this, rather than what Warren is doing.

Note2: If the data actually changed in `TEST_BIGDATA` then I suspect pandokia could show failures tomorrow. I'll have to check on that when I wake up.

jdavies-st:
Horray! 🤓 

Changing `CRDS_CONTEXT` is necessary for this version of the tests to be doing the same thing as what's running on Pandokia now.

I do suspect that there's a number of SVN commits since Apr 6 when Nadia first forked that repo that will have to be merged back into this branch.  By hand.  I'm looking at you @hbushouse.  I suspect this might resolve the issues `general.pipelines`, but obviously we need to check.

jhunkeler:
**MARK** - Rebased history against latest `jwst/master` (as of `e8e13d97ba982ad1f8b9b936685add8c9f9609c4`)

jhunkeler:
An eighteen hour run time is a bit too long...

build_42_too_long

build_42_too_long_detail

chart 1

hbushouse:
That's interesting. It only takes about 3-4 hours for the Pandokia set to run.

hbushouse:
Some of the processing logs show an hour or more elapsed in between fetching of various reference files at the beginning of pipeline runs, so I suspect the ridiculously long test times are due to either first time caching of reference files and/or problems in network or server connectivity to CRDS.

jdavies-st:
Yeah, the Jenkins tests were competing with resources that were running the Pandokia tests Friday evening.  I think Pandokia won.

jdavies-st:
So build 43 happened.  But since the last commit on this `jwst/regtest-poc` branch had a message including `[ci skip]` so that it wouldn't kick off the Travis CI tests *and* I rebase and merged, then Jenkins decided that it too could skip all the tests for this branch.

screen shot 2018-04-30 at 2 46 49 pm

That seems like bug.  Or maybe a feature?  Maybe I should have used a Merge Pull Request on Github instead?  I will do another change and try Merge Pull Request to see if that fixes things.

jdavies-st:
Yesterday's build 45 had 4 more tests pass compared to build 42.

unknown

The tests are all running successfully, but there are still 17 failures when comparing to the truth data.  Since everything is currently passing on the Pandokia side now with the same truth data, any testing failures are a problem with the test code or configurations themselves here on the Jenkins side, or the environment on the Jenkins side.  The failing tests are:

- [ ] miri.test_miri_ifu_wcs.test_miri_ifu_wcs
- [ ] miri.test_rscd2.test_rscd_miri2
- [ ] miri.test_sloperpipeline_1.test_detector1pipeline1
- [ ] miri.test_sloperpipeline_2.test_detector1pipeline2
- [ ] nircam.test_sloperpipeline_3.test_detector1pipeline3
- [ ] nircam.test_wfs_combine_step_refine2.test_wfs_combine
- [ ] niriss.test_ami_analyze.test_ami_analyze
- [ ] niriss.test_ami_pipeline.test_ami_pipeline
- [ ] niriss.test_photom.test_photom_niriss
- [ ] pipelines.test_coron3_1.test_coron3_pipeline1
- [ ] pipelines.test_fgs_sloper_1.test_fgs_detector1_1
- [ ] pipelines.test_nirspec_dark.test_nirspec_dark_pipeline
- [ ] pipelines.test_nrc_image3_1.test_image3_pipeline1
- [ ] pipelines.test_nrc_tso3_1.test_tso3_pipeline2
- [ ] pipelines.test_nrs_ifu_spec2.test_nrs_ifu_spec2
- [ ] test_nrs_msa_spec2.test_nrs_msa_spec2
- [ ] test_nrs_msa_spec2b.test_nrs_msa_spec2b

Well check these off as they get repaired.

pllim:
Is it too controversial to suggest that we stop running pandokia for this pipeline for, say, 24 hours, and let Jenkins run without competing for resources? Is it also that pandokia is doing funky things in the "big data" directory, causing these failures when Jenkins runs concurrently? (It seems a bit excessive that you run pandokia twice a day.)

hbushouse:
I don't want to cut running pandokia until we have one-to-one correspondence with Jenkins results. Remember that pandokia still essentially serves as our CI testing for PR's that get merged. Hence we need more frequent results than just once a day. The tests only take ~3 hours to run on pandokia, so even with running it twice a day, that still leaves 18 hours available to Jenkins runs.

jhunkeler:
MARK - Rebased history against latest jwst/master (as of `5ee053342ce79fc14dc6727b365c6c6c51e9ffe2`)

This commit matches today's pandokia build, so from now on, if you rebase this branch make sure it's based on a version that has passed all tests on pandokia.

`conda search -c http://ssb.stsci.edu/conda-dev jwst | tail -n1` < to get the current hash

jhunkeler:
**MARK** - Rebased history against 0400 build hash of `astroconda-dev/jwst` (`84b84a9537b370350bbf240d373da02bb47133a0`)

jdavies-st:
So it turns out that our 12 to 18 hour build problem was caused by trying to download reference files to a local cache within the Docker container.  The CRDS server has been unable to serve up files intermittently, in some cases making the test wait an hour for the file.

In any case, if we instead just use the CRDS cache that already is defined in the `crds_client` code and located on central store, then the only thing the CRDS servers need to return is the name of the file location in the cache.  And this is much quicker.  So now our runtimes are 2h40m.  Much more reasonable.

jdavies-st:
Build 49 completed in 2h34m, had 106 passes and 9 failures.

unknown

No new failures since Build 48.

Need to investigate the differences between Pandokia using numpy 1.13 and Jenkins using the latest numpy 1.14.3.

jdavies-st:
The following tests failures may be caused by differences between the numpy version in Pandokia (old) vs Jenkins (current).  The comparison files are generated by Pandokia using numpy 1.13.  Jenkins builds used the current version, numpy 1.14.2.

- [ ] niriss.test_ami_analyze.test_ami_analyze
- [ ] niriss.test_ami_pipeline.test_ami_pipeline
- [ ] pipelines.test_coron3_1.test_coron3_pipeline1
- [ ] pipelines.test_nrc_image3_1.test_image3_pipeline1
- [ ] pipelines.test_nrs_fs_multi_spec2_1.test_nrs_fs_multi_spec2_1 **(Passed in build 45)**

The following test is being skipped on Pandokia.  It has a fundamental error in the input data.

- [ ] miri.test_miri_ifu_wcs.test_miri_ifu_wcs

The following test is failing on both the Pandokia and Jenkins side due to a bug in our pipeline code.

- [ ] pipelines.test_nrs_ifu_spec2.test_nrs_ifu_spec2

The following test is failing due to a change in keyword reffile name.  Can be OK'ified.

- [ ] pipelines.test_nrs_msa_spec2.test_nrs_msa_spec2

The following test is still failing because of output file naming.  Hmmm.

- [ ] pipelines.test_fgs_sloper_1.test_fgs_detector1_1

stsci-hack:
Running the 9 failed tests under Numpy 1.13 provides the following comparisons with the results generated by Pandokia (also running Numpy 1.13):

-  **jwst.tests_nightly.general.pipelines.test_nrs_fs_multi_spec2_1.test_nrs_fs_multi_spec2_1**:  
    Identical results were obtained showing the exact same differences reported by Pandokia (where this test also failed).  The start of reported differences from the pytest run are shown for comparison with the results reported by Pandokia at [https://ssb.stsci.edu/pandokia/pandokia.cgi?query=detail&key_id=146834284](https://ssb.stsci.edu/pandokia/pandokia.cgi?query=detail&key_id=146834284)
```
E         Extension HDU 1:
E         
E            Data contains differences:
E              Column FLUX data differs in row 3:
E                 a> 0.33294644951820374
E                 b> 0.33298653364181519
E              Column FLUX data differs in row 4:
E                 a> 0.844227135181427
E                 b> 0.8442726731300354
E              Column FLUX data differs in row 5:
E                 a> 6.8564972877502441
E                 b> 6.8567428588867188
E              Column FLUX data differs in row 6:
E                 a> 1.9554159641265869
E                 b> 1.9552460908889771
E              Column FLUX data differs in row 7:
E                 a> 1.2003463506698608
E                 b> 1.2003757953643799
E              Column FLUX data differs in row 8:
E                 a> 8.3223638534545898
E                 b> 8.3230571746826172
E              Column FLUX data differs in row 9:
E                 a> 3.1123654842376709
E                 b> 3.112147331237793
E              Column FLUX data differs in row 10:
E                 a> 1.9586137533187866
E                 b> 1.9589259624481201
E              Column FLUX data differs in row 11:
E                 a> 5.9800047874450684
E                 b> 5.9802947044372559
E              Column FLUX data differs in row 12:
E                 a> 1.642103910446167
E                 b> 1.6425613164901733
E              ...3499 additional difference(s) found.
E              ...
E              3509 different table data element(s) found (21.42% different).
{Remainder of log removed for brevity}
```
-  **jwst.tests_nightly.general.miri.test_miri_ifu_wcs.test_miri_ifu_wcs**:  
    Still comes out with the same problem seen in previous pytest runs; namely, 
```
        AttributeError: No attribute 'wcs'
```
- **jwst.tests_nightly.general.niriss.test_ami_analyze.test_ami_analyze**: 
    Generates IDENTICAL results as Pandokia as evidenced by the following from the pytest log:
```
==================== 1 passed, 2 warnings in 11.65 seconds =====================
```

- **jwst.tests_nightly.general.niriss.test_ami_pipeline.test_ami_pipeline**:
    Only differences from Pandokia results stem from different path for ASNTABLE, as seen in the pytest log:
```
E         Primary HDU:
E         
E              Keyword ASNTABLE has different values:
E                 a> /internal/1/data/jwstdev/rtjenkins/niriss/test_ami_pipeline/test_lg1_asn.json
E                  ? -----------           ^^^^ ^^^^^^^
E                 b> /data4/jwst_test_data/niriss/test_ami_pipeline/test_lg1_asn.json
E                  ?      +     ++++++ ^ ^
E              Keyword ASNTABLE has different comments:
E                 a> Name of the ASN table
E                 b> N
```
- **jwst.tests_nightly.general.pipelines.test_coron3_1.test_coron3_pipeline1**:
    Only differences from Pandokia stem (again) from differences in paths for input files, in this case the PSF Mask file, as seen in the pytest log:
```
E         Primary HDU:
E         
E              Keyword R_PSFMAS has different values:
E                 a> /internal/1/data/jwstdev/rtjenkins/nircam/test_coron3/jwst_nircam_psfmask_somb.fits
E                  ? -----------           ^^^^ ^^^^^^^
E                 b> /data4/jwst_test_data/nircam/test_coron3/jwst_nircam_psfmask_somb.fits
E                  ?      +     ++++++ ^ ^
E         
```

- **jwst.tests_nightly.general.pipelines.test_nrc_image3_1.test_image3_pipeline1**:
    Inconsequential differences in the generated blendheader table, as seen in the pytest log:
```
E         Extension HDU 4:
E         
E              Keyword NAXIS1   has different values:
E                 a> 3814
E                 b> 3835
E              Keyword TFORM8   has different values:
E                 a> 3A
E                 b> 23A
E                  ? +
E              Keyword TFORM9   has different values:
E                 a> 7A
E                 b> 8A
E         
```
An additional difference was seen in the path for the ASNTABLE.

- **jwst.tests_nightly.general.pipelines.test_nrs_ifu_spec2.test_nrs_ifu_spec2**:
  We reproduce the exact same ERROR reported by Pandokia; namely (from the pytest log):
```
E           UnboundLocalError: local variable 'slit' referenced before assignment
```

- **jwst.tests_nightly.general.pipelines.test_nrs_msa_spec2.test_nrs_msa_spec2**:
    The only difference from Pandokia seen under pytest are the paths for the output file as reported in the header, as seen in the pytest log:
```
E         Primary HDU:
E         
E              Keyword FILENAME has different values:
E                 a> F170LP-G235M_MOS_observation-6-c0e0_001_DN_NRS1_mod_x1d.fits
E                  ? ^   ^^ ^   ^ ^^^                        ^^ ^^^
E                 b> f170lp-g235m_mos_observation-6-c0e0_001_dn_nrs1_mod_x1d.fits
E                  ? ^   ^^ ^   ^ ^^^                        ^^ ^^^
```

Based on these reports, **I would support flagging these tests as IDENTICAL** in behavior and results to what we see under Pandokia, and therefore, recognize the pytest tests as identical and available for replacing the Pandokia-based immediately.  The only lien against this comes from the recent failures seen under pytest for the  **jwst.tests_nightly.general.pipelines.test_fgs_sloper_1.test_fgs_detector1_1** test, which is already being investigated for corrective action.

jdavies-st:
So the latests test shows that we are on par with Pandokia now.

unknown

The only differences are the numpy 1.13 vs 1.14 differences as @stsci-hack documented above.  We will file issues for these, but they are not show-stoppers.  I think we can merge to `jwst/master`.

Issue #1900, opened 2018-04-16
Log association exposure being processed


jaytmiller:
One of the recent SIC/DIL errors was related to processing this association:

/ifs/int/jwstc/store/doggett/tests/SIC/DIL/cstring/JWSTDP-2018_1-180323/2018-03-28-210018/sdp/asn_creation/cal/level2/jw82600-o018_20180328t210625_image2_002_asn.json

More details:  https://jira.stsci.edu/browse/CRDS-191

Q:  what logging exists to identify the level-2 exposure .fits being processed in an association?  Would that logging be generally applicable to debugging or is CRDS the only application?

jdavies-st:
It seems like some logging in the `stpipe.pipeline` code might make this possible?  The private method

https://github.com/STScI-JWST/jwst/blob/22a880b8981b06f4c14bd37ad7a0430991b8907d/jwst/stpipe/pipeline.py#L174

seems to do the work for pipelines.  Would we want this logging in steps as well?  Right now, the `Step` class is only aware of containers/associations in writing out results, not reading them in.

jdavies-st:
@stscieisenhamer any thoughts?

stscieisenhamer:
Not having looked very closely at how reference files are retrieved, there seems to be two situations:

1. Precaching, done by `Pipeline`'s
2. Actual reference file retrieval.

For the easy case, case 2., the call to retrieve reference files should be the one logging for what data the reference files are being retrieved for. I'm guessing this is being passed a `DataModel`, so the log should report on what `DataModel.meta.filename` is. This can be done in either the jwst code, or the crds code.  Presuming crds is a bit more general, where sometimes the data is not jwst, I would stick in in the jwst side.

For case 1., how are associations being handled anyways? Presumably either the first member is being examined, or all the members are being examined. Regardless, at some point, a call is being made to fetch the reference files, or at least the reference file names. Whatever this call is, the `DataModel.meta.filename` can be reported, again on the jwst side of the call.

jaytmiller:
As it turns out,  I think the best place to do it is probably (1) where the exposure .fits filename is first resolved and the model is opened.   It's just direct and cleaner.

[...]

PR #1907, opened 2018-04-18
Added INFO log message to pipeline prefetch to report exposure filenames

jaytmiller:
Implementation resolving #1900.

### DIFF
--- a/jwst/stpipe/pipeline.py
+++ b/jwst/stpipe/pipeline.py
@@ -206,6 +206,8 @@ class Pipeline(Step):
 
         fetch_types = sorted(set(self.reference_file_types) - set(ovr_refs.keys()))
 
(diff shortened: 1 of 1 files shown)


jwst/stpipe/pipeline.py
```diff
@@ -206,6 +206,8 @@ def _precache_references_impl(self, model):
 
         fetch_types = sorted(set(self.reference_file_types) - set(ovr_refs.keys()))
 
+        self.log.info("Prefetching reference files for dataset: " + repr(model.meta.filename) +
```
hbushouse: At this point in the calling sequence can `model` ever be a `ModelContainer`, as opposed to a single data model? If so, then I don't think `model.meta.filename` is defined or at most might be the name of an ASN file. Or are `ModelContainer`s broken down into individual models before this is called?

PR #1911, opened 2018-04-18
Nirspec WCS 

nden:
This can be merged after the new OTE and IFUPOST files are delivered to CRDS.

Changes in this branch were made during the nirspec sprint. Relevant commits are referenced in the issues.


nden:
I will merge this now to allow the reference files to be tested against the pipeline. The `ifupost` reference file has a new schema and the CRDS certifier fails without it.

PR #1923, opened 2018-04-20
Fix pathloss 1-d wavelength vectors

### LINKED ISSUE #759: Pathloss wavelength vectors error
James Muzerolle reports (https://stsci-ins.basecamphq.com/projects/11477312-jwst-pipeline/posts/101800996/comments):

There is a mistake in the point and uniform source wavelength vectors added to the output file in the pathloss correction step.  The first wavelength value in each should be 0.6 microns, not 0.365 (and the last point should be 5.3).  Looks like this is just an offset problem when reading in the CRPIX/CRVAL keywords from the reference file headers?

stscirij:
Fixes #759


stscicrawford:
Just double checking, is the off-by-one related at all to the NIRSPEC index issue?  If it is, was this fixed with @nden 's updates or will this have to be updated separately?   

Otherwise, it might be good to add a comment in the code about this in case the models and WCS end up changing in the future.

hbushouse:
@stscirij OK to merge?

PR #1924, opened 2018-04-20
Put output frame shape values in correct numpy order

### LINKED ISSUE #1875: Resample output frame is computed to be rotated by 90 deg
Some simulated data from the NIRCam team show that for a certain rotation, the output frame of resampled data was computed to be the correct size to fit the 2 NIRCam long detectors in the output frame, but the frame itself was rotated 90 degrees from what it should have been.

See figure below:

m51-2mass-09

The figure shows two output resample frames, but with the 2 files in reverse order in the association.  These have been wcs matched in the figure.  Also shown are the point sources from the catalog that have been simulated.

jdavies-st:
The shape of the output frame is a numpy shape in (Y, X) order.  The bounding box of the GWCS object is in (X, Y) order.  We were were swapping the order of axes twice, which produced a transpose of the output frame size.

Resolves #1875

### DIFF
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -117,7 +117,7 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
         transform = []
         wcsinfo = pointing.wcsinfo_from_model(refmodel)
         sky_axes, spec, other = gwutils.get_axes(wcsinfo)
-        rotation = astmodels.AffineTransformation2D(np.array(wcsinfo['PC']))
+        rotation = astmodels.AffineTransformation2D(wcsinfo['PC'])
         transform.append(rotation)
         if sky_axes:
             cdelt1, cdelt2 = wcsinfo['CDELT'][sky_axes]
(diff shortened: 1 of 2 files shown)


jwst/assign_wcs/util.py
```diff
@@ -117,7 +117,7 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
         transform = []
         wcsinfo = pointing.wcsinfo_from_model(refmodel)
         sky_axes, spec, other = gwutils.get_axes(wcsinfo)
-        rotation = astmodels.AffineTransformation2D(np.array(wcsinfo['PC']))
+        rotation = astmodels.AffineTransformation2D(wcsinfo['PC'])
```
hbushouse: Is this the line that actually fixes the swapping of the axis order? I assume the previous application of np.array() was doing the swapping?

jdavies-st: No, that line is just an unnecessary casting of the wcsinfo into a numpy array.  But the API of `pointing.wcsinfo_from_model` already returns this as a numpy array, so no need to have it in there twice.

The actual repair is the deletion of line 56 in resample_utils.py that did an unnecessary swapping of the order.  The order had already been swapped.  I clarified this behavior in the docstrings on the called functions.

Issue #1681, opened 2018-01-22
MIRI FirstFrame and LastFrame minimum number of groups


karllark:
Currently the first and last frame steps only flag groups if there are at least 2 groups.  Now that we have both corrections, this can result in both groups in a 2 group integration being flagged as bad.  This is not what should happen.

The best solution would be to reorder the steps to have the LastFrame run before the FirstFrame.  And have the LastFrame check that there are at least 3 good groups and the FirstFrame that there are least 3 good groups.  This will have the effect of running the LastFrame step for 3 group integrations *but* not the FirstFrame step.  The LastFrame step is more important, especially as most 3 group integrations will be for bright sources.  In summary:

run LastFrame before FirstFrame
Have both steps check the number of good groups, not just number of groups as currently written.

2 groups: Neither LastFrame or FirstFrame run.
3 groups: LastFrame run, but not FirstFrame.
4+ groups: LastFrame and FirstFrame run.

I could add this to the pull request #1327.  Or this could be a separate pull request depending on what makes the most sense.

jemorrison:
Why is the last frame rejection more important than the first frame rejection ?  In the first integration for bright data the frame that is more incorrect is the last frame, however I think the opposite is true for the
second integration where the first frame is affected by the additional reset switch charge decay.
I am not following why we would want to switch the order ? 

As it is currently written what happens if there are 3 groups in the integration and the first and last are rejected - so there is only one valid group. Does the pipeline fail ? How does it give a slope for 1 point ?
@mwregan2 
@karllark 
@hbushouse

hbushouse:
The ramp_fit step is already capable of handling cases where you only have 1 group to work with. It'll simply use the value of that group and divide it by the group integration time to produce a slope value.

mwregan2:
That one-group method doesn’t work for MIRI since there is no super bias subtraction for MIRI.

hbushouse:
Ah, right. I don't think anyone thought of that until now. So that suggests anything in ramp_fit always needs a minimum of 2 groups when working with MIRI data. For 2 groups, it just computes (group2 - group1) / group_time.

dmggh:
I'll add that to the list of ramp_fitting updates.

________________________________

karllark:
[...]

Issue #1714, opened 2018-02-07
Ramp_fit optional output file has zeros for some y-intercept values


hbushouse:
A B7.1 tester reported finding some zero values in the y-intercept results in the ramp_fit optional fit results file. See https://stsci-ins.basecamphq.com/projects/11477312-jwst-pipeline/posts/104607184/comments#379448456

This may get resolved by other ramp_fit rework that's currently underway. Will need to check once those changes are finished.

hbushouse:
Fixed by #1925

Issue #1687, opened 2018-01-25
Incorrect Slope Combination in Ramp Fitting


mwregan2:
In ramp fitting when slopes are averaged either between segments or between integrations. The total variance is used. This variance is the sum of the Poisson and what is referred to as the read noise variance. This is incorrect and leads to a bias in the combined slope. This is because including the Poisson noise is wrong. When multiple slopes are averaged the Poisson noise will always be less for smaller slopes. Thus, the average is biased low. I have first hand experience with this since I made the same mistake in the NICMOS pipeline. Reporting this total variance as the uncertainty of the fit also has serious negative consequences for any image combination steps that follow.
Also, what is referred to in the code as "sig_slope_r" is not well named. This is the estimated variance of the fitted slope. It includes components from both read noise and Poisson noise. This is the value that is used in the HST pipelines to weight the segments. It does have the problem that it is poorly determined when the number of samples in the fit is small. Even so, for now it is the best value to use for weighting slopes when calculating their weighted mean. Karl Gordon and I are discussing a better estimate of the slope variance that we could use in the future.

hbushouse:
Fixed by #1925

Issue #1686, opened 2018-01-24
Incorrect calculation of the variance of the slope due to Poisson Noise


mwregan2:
In ramp_fit.calc_opt_fit the variance of the slope due to Poisson noise is calculated as:
` 
 sig_slope_p = slope/(mask_2d.sum(axis=0) - 1.)
`
Since at this point in the code the slope is in units of DN/frame. The output of this is in units of DN. The actual variance in DN requires the use of the gain in units of e-/DN. If we express the variance in DN, then we need to divide the right hand side of the equation by the gain.

  Also, the variable name should be changed to match what it is. It should be:
`variance_slope_poisson`
It is not a measurement of the standard deviation. It is the variance.

hbushouse:
INS tracking in https://jira.stsci.edu/browse/JP-236

hbushouse:
Fixed by #1925

PR #1925, opened 2018-04-23
Update ramp_fit slope variances

dmggh:
Updates to the ramp_fit computations for segment-specific and integration-specific variances of the slope due to either Poisson noise, readnoise, or both.

2. Updated computations of integration-specific and overall slopes, and weights.

3. Added code to use data from a good initial group in a ramp only if there are not good data in later groups.

4. Added check to ensure that MIRI dataset has a minimum of 2 groups.

5. Updated deprecated logging format

6. Removed obsolete code and variables.

7. Updated code resolves these github issues: 1681, 1683, 1686, 1687, 1714

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -1,5 +1,5 @@
 #! /usr/bin/env python
-# 
+#
 #  ramp_fit.py - calculate weighted mean of slope, based on Massimo
 #                Robberto's "On the Optimal Strategy to fit MULTIACCUM
 #                ramps in the presence of cosmic rays."
@@ -10,11 +10,12 @@
 #                determined from the input data quality arrays.
 #
 # Note:
-# In this module, comments on the 'first read','second read', etc are 1-based.
+# In this module, comments on the 'first group','second group', etc are
+#    1-based, unless noted otherwise.
 
 import time
-import numpy as np
 import logging
+import numpy as np
 
 from .. import datamodels
 from ..datamodels import dqflags
@@ -26,7 +27,6 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 BUFSIZE = 1024 * 30000  # 30Mb cache size for data section
-MIN_ERR = 1.E-10 # minimum allowed value for error cube
 
 # Replace zero or negative variances with this.
 LARGE_VARIANCE = 1.e8
@@ -40,9 +40,9 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
     Calculate the count rate for each pixel in all data cube sections and all
     integrations, equal to the slope for all sections (intervals between
     cosmic rays) of the pixel's ramp divided by the effective integration time.
-    If the weighting parameter is set to 'optim', the optimal weighting from a
-    pere by Fixsen (ref. TBA) will be used in the fitting; otherwise the
-    fitting will be unweighted.
+    The weighting parameter must currently be set to 'optim', to use the optimal
+    weighting (paper by Fixsen, ref. TBA) will be used in the fitting; this is
+    currently the only supported weighting scheme.   
 
     Parameters
     ----------
@@ -66,8 +66,8 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
         'GLS' specifies that generalized least squares should be used.
 
     weighting: string
-        'unweighted' specifies that no weighting should be used (default)
-        'optimal' specifies that optimal weighting should be used
+        'optimal' specifies that optimal weighting should be used;
+         currently the only weighting supported.
 
     Returns
     -------
@@ -108,7 +108,7 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
 
 
 def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
-                  weighting):
+                 weighting):
     """
     Extended Summary
     ----------------
@@ -136,7 +136,8 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
 
     weighting: string
         'unweighted' specifies that no weighting should be used
-        'optimal' specifies that optimal weighting should be used
+        'optimal' specifies that optimal weighting should be used (no longer
+            supported
 
     Returns
     -------
@@ -152,21 +153,21 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
         DM object containing optional OLS-specific ramp fitting data for the
         exposure; this will be None if save_opt is False
     """
+
     tstart = time.time()
- 
+
     # get needed sizes and shapes
-    nreads, npix, imshape, cubeshape, n_int, instrume, frame_time, ngroups = \
-        utils.get_dataset_info(model)
+    nreads, npix, imshape, cubeshape, n_int, instrume, frame_time, ngroups, \
+        group_time = utils.get_dataset_info(model)
 
     # Save original shapes for writing to log file, as these may change for MIRI
     orig_nreads = nreads
     orig_cubeshape = cubeshape
 
-    # For MIRI datasets having >1 reads, if all final reads are flagged
+    # For MIRI datasets having >1 groups, if all final groups are flagged
     # as DO_NOT_USE, resize the input model arrays to exclude the final group.
     if (instrume == 'MIRI' and nreads > 1):
         last_gdq = model.groupdq[:,-1,:,:]
-        gdq_shape = last_gdq.shape
 
         if np.all(np.bitwise_and( last_gdq, dqflags.group['DO_NOT_USE'] )):
             model.data = model.data[:,:-1,:,:]
@@ -175,23 +176,25 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
             nreads -= 1
             ngroups -= 1
             cubeshape = (nreads,)+imshape
-            log.info('MIRI dataset has all final reads flagged as DO_NOT_USE.')
+            log.info('MIRI dataset has all final groups flagged as DO_NOT_USE.')
+
+        # Next block is to satisfy github issue 1681:
+        # "MIRI FirstFrame and LastFrame minimum number of groups #1681"
+        if (ngroups < 2):
+            log.error('MIRI datasets require at least 2 groups/integration')
+            log.error('(NGROUPS), so will not process this dataset.')
+            return None, None, None
 
     if (ngroups == 1):
-        log.warn('Dataset has NGROUPS=1, so count rates for each integration')
-        log.warn('will be calculated as the value of that 1 group divided by')
-        log.warn('the group exposure time.')
+        log.warning('Dataset has NGROUPS=1, so count rates for each integration')
+        log.warning('will be calculated as the value of that 1 group divided by')
+        log.warning('the group exposure time.')
 
-    slopes = np.zeros(imshape, dtype=np.float32)
-
-    slope_wtd = np.zeros(imshape, dtype=np.float64)
-    m_sum_2d = np.zeros(imshape, dtype=np.float64)
-    var_sum_2d = np.zeros(imshape, dtype=np.float64)
+    med_slopes = np.zeros(imshape, dtype=np.float32)
+    slope_wtd = np.zeros(imshape, dtype=np.float32)
 
     # For multiple-integration datasets, will output integration-specific
     #    results to separate file named <basename> + '_integ.fits'
-    slope_int, err_int, dq_int, m_by_var_int, inv_var_int, \
-            var_p_int, var_r_int = utils.alloc_int(n_int, imshape)
 
     # Calculate effective integration time (once EFFINTIM has been populated
     #   and accessible, will use that instead), and other keywords that will
@@ -203,18 +206,17 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
 
     # Get GROUP DQ and ERR arrays from input file
     gdq_cube = model.groupdq
-    err_cube = model.err
 
     # get max number of segments fit in all integrations
     max_seg = calc_num_seg(gdq_cube, n_int)
     f_max_seg = 0  # final number to use, usually overwritten by actual value
 
-    opt_res = utils.OptRes(n_int, imshape, max_seg, nreads)
+    (slope_int, dq_int, var_p3, var_r3, median_diffs_2d, var_p4, var_r4,
+        var_both4, var_both3, inv_var_p4, inv_var_r4, inv_var_both4,
+        s_inv_var_p3, s_inv_var_r3, s_inv_var_both3, segs_4, num_seg_per_int) =\
+        utils.alloc_arrays(n_int, imshape, max_seg)
 
-    # Allocate for the various types of variances of the_slope:
-    # Poisson noise only (per segment); Read noise only, per segment
-    var_p_s_4d = np.zeros((n_int,)+(max_seg,)+imshape,dtype=np.float32)
-    var_r_s_4d = np.zeros((n_int,)+(max_seg,)+imshape,dtype=np.float32)
+    opt_res = utils.OptRes(n_int, imshape, max_seg, nreads)
 
     # Get Pixel DQ array from input file. The incoming RampModel has uint8
     #   PIXELDQ, but ramp fitting will update this array here by flagging
@@ -229,16 +231,21 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
 
     # Get readnoise array for calculation of variance of noiseless ramps, and
     #   gain array in case optimal weighting is to be done
-    readnoise_2d, gain_2d = utils.get_ref_subs(model, readnoise_model, 
+    readnoise_2d, gain_2d = utils.get_ref_subs(model, readnoise_model,
                             gain_model)
 
     # Flag any bad pixels in the gain
     pixeldq = utils.reset_bad_gain( pixeldq, gain_2d )
 
-    # loop over data integrations
+    # In this 'First Pass' over the data, loop over integrations and data
+    #   sections to calculate the estimated median slopes, which will be used
+    #   to calculate the variances. This is the same method to estimate slopes
+    #   as is done in the jump detection step, except here CR-affected and
+    #   saturated groups have already been flagged.
+    # Loop over data integrations:
     for num_int in range(0, n_int):
 
-        # loop over data sections
+        # Loop over data sections
         for rlo in range(0, cubeshape[1], nrows):
             rhi = rlo + nrows
 
@@ -247,53 +254,71 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
 
             data_sect = model.get_section('data')[num_int, :, rlo:rhi, :]
 
-            # skip data section if it is all NaNs
+            # Skip data section if it is all NaNs
             if  np.all(np.isnan( data_sect)):
                 log.error('Current data section is all nans, so not processing the section.')
-                break   
+                break
 
-            # Poisson noise only; read noise only
-            var_p_2d = np.zeros((max_seg, (rhi - rlo) * imshape[-1]),
-                                dtype=np.float32)
-            var_r_2d = np.zeros((max_seg, (rhi - rlo) * imshape[-1]),
-                                dtype=np.float32)
-
-            # first frame section for 1st read of current integration
+            # first frame section for 1st group of current integration
             ff_sect = model.get_section('data')[ num_int, 0, rlo:rhi, :].\
                 astype(np.float32)
-            # get appropriate sections
+
+            # Get appropriate sections
             gdq_sect = gdq_cube[num_int, :, rlo:rhi, :]
             rn_sect = readnoise_2d[rlo:rhi, :]
             gain_sect = gain_2d[rlo:rhi, :]
 
-            t_err_cube, t_dq_cube, m_by_var, inv_var, opt_res, f_max_seg = \
-                 calc_slope(data_sect, gdq_sect, frame_time, opt_res,
-                     rn_sect, gain_sect, max_seg, ngroups, weighting, f_max_seg,
-                     var_p_2d, var_r_2d)
+            # Reset all saturated groups in the input data array to NaN
+            where_sat = np.where( np.bitwise_and(gdq_sect,
+                                  dqflags.group['SATURATED']) != 0)
+            data_sect[ where_sat ] = np.NaN
 
-            err_cube[num_int, :, rlo:rhi, :] += t_err_cube
+            # Compute the first differences of all groups
+            first_diffs_sect = np.diff(data_sect, axis=0)
+
+            # If the dataset has only 1 group/integ, assume the 'previous group'
+            #   is all zeros, so just use data as the difference
+            if (first_diffs_sect.shape[0] == 0):
+                first_diffs_sect = data_sect[0,:,:]
+                first_diffs_sect = first_diffs_sect[np.newaxis,:]
+
+            # Similarly, for datasets having >1 group/integ and having
+            #  single-group segments, just use the data as the difference
+            wh_nan = np.where( np.isnan( first_diffs_sect[0,:,:]) )
+            if (len(wh_nan[0]) > 0):
+                first_diffs_sect[0,:,:][ wh_nan ] = data_sect[0,:,:][ wh_nan ]
+
+            # Mask all the first differences that are affected by a CR, starting 
+            #  at group 1.  The purpose of starting at index 1 is to shift all the
+            #  indexes down by 1, so they line up with the indexes in first_diffs.
+            i_group, i_yy, i_xx, = np.where( np.bitwise_and(gdq_sect[:,:,1:],
+                                             dqflags.group['JUMP_DET']) != 0)
+            first_diffs_sect[ i_group-1, i_yy, i_xx ] = np.NaN
+
+            # All first differences affected by saturation and CRs have been set
+            #  to NaN, so compute the median of all non-NaN first differences.
+            median_diffs_2d[ rlo:rhi, : ] += np.nanmedian(first_diffs_sect, axis=0 )
+
+            # Calculate the slope of each segment
+            t_dq_cube, inv_var, opt_res, f_max_seg, num_seg = \
+                 calc_slope(data_sect, gdq_sect, frame_time, opt_res, rn_sect,
+                            gain_sect, max_seg, ngroups, weighting, f_max_seg,
+                            group_time)
+
+            # Populate 3D num_seg { integ, y, x } with 2D num_seg for this data 
+            #  section (y,x) and integration (num_int)
+            sect_shape = data_sect.shape[-2:]
+            num_seg_per_int[num_int, rlo:rhi, :] = num_seg.reshape(sect_shape)
+
             gdq_cube[num_int, :, rlo:rhi, :] = t_dq_cube
 
-            # Compress 4D->2D dq arrays for saturated and jump-detected pixels
             pixeldq_sect = pixeldq[rlo:rhi, :].copy()
             dq_int[num_int, rlo:rhi, :] = \
                   dq_compress_sect(t_dq_cube, pixeldq_sect).copy()
 
-            sect_shape = data_sect.shape[-2:]
-            m_sum_2d[rlo:rhi, :] += m_by_var.reshape(sect_shape)
-            var_sum_2d[rlo:rhi, :] += inv_var.reshape(sect_shape)
-
             # Loop over the segments and copy the reshaped 2D segment-specific
             #  results for the current data section to the 4D output arrays.
             opt_res.reshape_res(num_int, rlo, rhi, sect_shape, ff_sect)
-   
-            # Populate the segment- and instrument-specific variances for the 
-            #  recent section results
-            for ii_seg in range(0, max_seg):
-                var_p_s_4d[num_int, ii_seg, rlo:rhi, :] = \
-                     var_p_2d[ii_seg, :].reshape(sect_shape)
-                var_r_s_4d[num_int, ii_seg, rlo:rhi, :] = \
-                     var_r_2d[ii_seg, :].reshape(sect_shape)
 
             # Calculate difference between each slice and the previous slice
             #   as approximation to cosmic ray amplitude for those pixels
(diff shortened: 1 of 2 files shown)


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -175,23 +176,25 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+    if (ngroups == 1):
+        log.warning('Dataset has NGROUPS=1, so count rates for each integration')
+        log.warning('will be calculated as the value of that 1 group divided by')
+        log.warning('the group exposure time.')
 
-    slope_wtd = np.zeros(imshape, dtype=np.float64)
-    m_sum_2d = np.zeros(imshape, dtype=np.float64)
-    var_sum_2d = np.zeros(imshape, dtype=np.float64)
+    med_slopes = np.zeros(imshape, dtype=np.float32)
+    slope_wtd = np.zeros(imshape, dtype=np.float32)
```
philhodge: It doesn't look as if `slope_wtd` is ever used.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -247,53 +254,71 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
             if  np.all(np.isnan( data_sect)):
                 log.error('Current data section is all nans, so not processing the section.')
-                break   
-
-            # Poisson noise only; read noise only
-            var_p_2d = np.zeros((max_seg, (rhi - rlo) * imshape[-1]),
-                                dtype=np.float32)
-            var_r_2d = np.zeros((max_seg, (rhi - rlo) * imshape[-1]),
-                                dtype=np.float32)
+                break
```
philhodge: It looks to me as if this would exit the loop over data sections, i.e. there would be no further processing of the data.  Isn't `continue` what you want, rather than `break`?

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -247,53 +254,71 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+            where_sat = np.where( np.bitwise_and(gdq_sect,
+                                  dqflags.group['SATURATED']) != 0)
+            data_sect[ where_sat ] = np.NaN
+
+            # Compute the first differences of all groups
+            first_diffs_sect = np.diff(data_sect, axis=0)
+
+            # If the dataset has only 1 group/integ, assume the 'previous group'
+            #   is all zeros, so just use data as the difference
+            if (first_diffs_sect.shape[0] == 0):
```
philhodge: Can't you just use `data_sect` for this case, i.e.:

    first_diffs_sect = data_sect.copy()

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -247,53 +254,71 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+            # Similarly, for datasets having >1 group/integ and having
+            #  single-group segments, just use the data as the difference
+            wh_nan = np.where( np.isnan( first_diffs_sect[0,:,:]) )
+            if (len(wh_nan[0]) > 0):
+                first_diffs_sect[0,:,:][ wh_nan ] = data_sect[0,:,:][ wh_nan ]
+
+            # Mask all the first differences that are affected by a CR, starting 
+            #  at group 1.  The purpose of starting at index 1 is to shift all the
+            #  indexes down by 1, so they line up with the indexes in first_diffs.
+            i_group, i_yy, i_xx, = np.where( np.bitwise_and(gdq_sect[:,:,1:],
```
philhodge: Shouldn't the slice be `gdq_sect[1:, :, :]`?  And if so, won't that be a problem in the next line for data with only one group when you index with `i_group-1`?

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -303,92 +328,237 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+    log.debug('The execution time in seconds: %f', tstop - tstart)
 
     # Create new model...
     new_model = datamodels.ImageModel(data=c_rates.astype(np.float32),
-            dq=final_pixeldq.astype(np.uint32), 
-            var_poisson=var_p_2d_all.astype(np.float32),
-            var_rnoise=var_r_2d_all.astype(np.float32), err=err_2d_all.copy())
+            dq=final_pixeldq.astype(np.uint32),
+            var_poisson=var_p2.astype(np.float32),
+            var_rnoise=var_r2.astype(np.float32), err=err_2d.copy())
```
hbushouse: Is the copy really needed for err_2d, or would a simple view of the array work, as is done for the other arrays?

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -303,92 +328,237 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
                        data_diff * (dq_cr != 0)
- 
-            m_by_var_int[num_int, rlo:rhi, :] = m_by_var.reshape(sect_shape)
-            inv_var_int[num_int, rlo:rhi, :] = inv_var.reshape(sect_shape)
 
-        slope_int[num_int, :, :] = \
-            utils.calc_slope_int(slope_int, m_by_var_int, inv_var_int, num_int)
+    # Compute the final 2D array of differences.
+    median_diffs_2d /= n_int
+    med_slopes = median_diffs_2d.copy()
```
philhodge: `median_diffs_2d` isn't used again, so it could be deleted to save space.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -657,13 +827,11 @@ def gls_ramp_fit(model,
 
     # Create new model...
     if n_int > 1:
-        new_model = datamodels.ImageModel(data=slopes.astype(np.float32),
-                                      dq=final_pixeldq,
-                                      err=gls_err.astype(np.float32))
+        new_model = datamodels.ImageModel(data=med_slopes.astype(np.float32),
```
hbushouse: Assuming that this is the data model that will end up as the `_rate.fits` product, where are the `med_slopes` values coming from? I thought it looked like earlier that these were the slopes computed from the simple process of taking the first-differences of all the groups. But what we really want here are the slopes computed from the "real" fitting process, and for n_int>1 the result of taking the average of those fits over all integrations.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -733,13 +901,13 @@ def dq_compress_sect(gdq_sect, pixeldq_sect):
...
-    flagged as a jump detection in the 4D GROUPDQ array, find the corresponding
-    image locations, and set the COSMIC_BEFORE flag in those locations in the
-    PIXELDQ array
+    Get ramp locations where the data has been flagged as saturated in the 4D
+    GROUPDQ array for the current data section, find the corresponding image
+    locations, and set the SATURATED flag in those locations in the PIXELDQ
+    array. Similarly, get the ramp locations where the data has been flagged as
+    a jump detection in the 4D GROUPDQ array, find the corresponding image 
+    locations, and set the COSMIC_BEFORE flag in those locations in the PIXELDQ
+    array.
```
hbushouse: If saturation and CR flags are copied over from `groupdq` to `pixeldq` doesn't that then mark all of the data for the pixel as bad/unusable? Only the particular groups that are flagged as bad are bad, not all the data for the pixel. Or maybe `pixeldq` is being used in some special way in this function?

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -1198,32 +1368,27 @@ def fit_next_segment(start, end_st, end_heads, pixel_done, data_sect, mask_2d,
...
-    # CASE E) - interval too short to fit normally (only 2 good read)
-    #    at end of array, NGROUPS>1, but exclude NGROUPS==2 datasets
+            return 1, num_seg
+
+
+    # CASE E) - interval too short to fit normally (only 2 good groups)
+    #    At end of array, NGROUPS>1, but exclude NGROUPS==2 datasets
     #    as they are covered in CASE D
-    #    - set start to -1 to designate all fitting done
+    #    - set start to -1 to designate all fitting doned
```
hbushouse: "doned"? 😀

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -229,16 +231,21 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
-    readnoise_2d, gain_2d = utils.get_ref_subs(model, readnoise_model, 
+    readnoise_2d, gain_2d = utils.get_ref_subs(model, readnoise_model,
                             gain_model)
 
     # Flag any bad pixels in the gain
     pixeldq = utils.reset_bad_gain( pixeldq, gain_2d )
 
-    # loop over data integrations
+    # In this 'First Pass' over the data, loop over integrations and data
+    #   sections to calculate the estimated median slopes, which will be used
```
philhodge: I think it's worth mentioning here that the actual slopes are also computed.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -303,92 +328,237 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+        var_r3[num_int, :, :] = 1./ s_inv_var_r3[num_int, :, :]
+        var_r3 = utils.rem_nan_inf(var_r3)
+
+        var_both4[ num_int,:,:,:] = var_r4[num_int,:,:,:] + var_p4[num_int,:,:,:]
+        inv_var_both4[num_int, :, :, :] = 1./var_both4[num_int, :, :, :]
+        inv_var_both4 = utils.rem_nan_inf(inv_var_both4)
+
+        # Want to retain values in the 4D arrays only for the segments that each
+        #   pixel has, so will zero out values for the higher indices. Creating
+        #   and manipulating intermediate arrays (views, such as var_p4_int)
```
philhodge: It looks as if the close parenthesis should be moved to the next line.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -303,92 +328,237 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+    s_slope_by_var2 = s_slope_by_var3.sum(axis=0) # sum over integrations
+
+    s_inv_var_both2 = s_inv_var_both3.sum(axis=0)
+
+    # Compute the 'dataset-averaged' slope
+    slope_dataset2 = s_slope_by_var2/s_inv_var_both2
+    slope_dataset2 = utils.rem_nan_inf(slope_dataset2)
+
+    # Compute the integration-specific slope
+    the_num = (slope_seg * inv_var_both4).sum(axis = 1) # sum over all segments
```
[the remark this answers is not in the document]
hbushouse: I don't recall exactly what we decided during discussions a month or so ago, but I think it does make some sense to NOT include Poisson noise in the weighting, because that would have a tendency for a bias towards lower slope values (which have lower Poisson noise and hence higher weight).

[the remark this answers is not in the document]
philhodge: @hbushouse If the median slope over the entire exposure is used for computing the Poisson noise, that should eliminate the bias toward lower slopes, shouldn't it (because the same slope is used for every segment)?  I thought the bias that was of concern had to do with the lengths of the segments, though I don't understand why that would introduce a bias.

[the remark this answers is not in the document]
hbushouse: Yes, I believe you're right and I think that's why and where we wanted to use the median slopes (from first differences) - it was for use in estimating the Poisson noise. As opposed to using the actual fitted slopes, which vary from segment to segment and integration to integration. So in that case it should be OK to include variances from both Poisson and read noise in the weighting for computing averaged slopes?

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -1053,28 +1209,32 @@ def fit_next_segment(start, end_st, end_heads, pixel_done, data_sect, mask_2d,
...
         'optimal' specifies that optimal weighting should be used
 
-    total_mask_sum: int, 1D arrays
-        initial number of good reads along ramp, used to distinguish between
-        cosmic rays and saturated pixels
-
     f_max_seg: int
         actual maximum number of segments within a ramp, updated here based on
-        fitting ramps in the currret data section; later used when truncating 
+        fitting ramps in the currret data section; later used when truncating
```
philhodge: typo:  currret --> current
Also 10 lines further down.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -1086,72 +1246,84 @@ def fit_next_segment(start, end_st, end_heads, pixel_done, data_sect, mask_2d,
...
 
-    # CASE A) Long enough (semiramp has >2 reads), at end of ramp
+    # CASE A) Long enough (semiramp has >2 groups), at end of ramp
     #    - set start to -1 to designate all fitting done
     #    - remove current end from end stack
     #    - set number of ends to 0
     #    - add slopes and variances to running sums
-    mask_sum = mask_2d.sum(axis=0)
-    wh_check = np.where((l_interval>2) & (end_locs==nreads - 1) & (mask_sum>0))
+    wh_check = np.where((l_interval>2) & (end_locs==nreads - 1) & (~pixel_done))
```
philhodge: Is it safe to use `&` for logical `and` with boolean arrays?  I can see that it works, but I wonder whether it's a throwback to when True and False were just aliases for 1 and 0.

jwst/ramp_fitting/utils.py
```diff
@@ -855,11 +902,15 @@ def get_dataset_info(model):
...
+        Time increment between groups, in seconds.
     """
 
     instrume = model.meta.instrument.name
     frame_time = model.meta.exposure.frame_time
     ngroups = model.meta.exposure.ngroups
+    group_time = model.meta.exposure.group_time
 
     n_int = model.data.shape[0]
     nreads = model.data.shape[1]
```
philhodge: `nreads` should be the same as `ngroups`.  If they're not the same, it would be better to log a warning and override the value of `ngroups` with `nreads`, which is more likely to be correct, since it's based on the image shape.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -136,7 +136,8 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
 
     weighting: string
         'unweighted' specifies that no weighting should be used
-        'optimal' specifies that optimal weighting should be used
+        'optimal' specifies that optimal weighting should be used (no longer
```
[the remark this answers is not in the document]
dmggh: You are correct - updating.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -175,23 +176,25 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
-    slopes = np.zeros(imshape, dtype=np.float32)
+    if (ngroups == 1):
+        log.warning('Dataset has NGROUPS=1, so count rates for each integration')
+        log.warning('will be calculated as the value of that 1 group divided by')
+        log.warning('the group exposure time.')
 
-    slope_wtd = np.zeros(imshape, dtype=np.float64)
-    m_sum_2d = np.zeros(imshape, dtype=np.float64)
-    var_sum_2d = np.zeros(imshape, dtype=np.float64)
+    med_slopes = np.zeros(imshape, dtype=np.float32)
```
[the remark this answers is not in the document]
dmggh: You are correct - updating.

jwst/ramp_fitting/ramp_fit.py
dmggh: You are correct - updating.

jwst/ramp_fitting/ramp_fit.py
dmggh: Yes, switched to continue.

jwst/ramp_fitting/ramp_fit.py
dmggh: Yes.

jwst/ramp_fitting/ramp_fit.py
dmggh: Yes, and fixing this uncovered a semi-related bug (now fixed).

jwst/ramp_fitting/ramp_fit.py
dmggh: Right.

jwst/ramp_fitting/ramp_fit.py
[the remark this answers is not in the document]
dmggh: Yes.   Back in July of 2017, before there was any calculation of the variances (so those extensions did not yet exist in the output), I ran a niriss test of 2k x 2k, with 100 groups per integration, and 1 integration.  The execution time then was 98 seconds. Now, running with the same dataset, and the variance calculations, the execution time is 122 seconds, so about 24% longer.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -175,23 +176,25 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
 
-    if (ngroups == 1):
-        log.warn('Dataset has NGROUPS=1, so count rates for each integration')
-        log.warn('will be calculated as the value of that 1 group divided by')
-        log.warn('the group exposure time.')
+        # Next block is to satisfy github issue 1681:
+        # "MIRI FirstFrame and LastFrame minimum number of groups #1681"
+        if (ngroups < 2):
+            log.error('MIRI datasets require at least 2 groups/integration')
+            log.error('(NGROUPS), so will not process this dataset.')
```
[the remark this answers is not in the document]
dmggh: Yes, it does that.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -175,23 +176,25 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+        log.warning('the group exposure time.')
 
-    slope_wtd = np.zeros(imshape, dtype=np.float64)
-    m_sum_2d = np.zeros(imshape, dtype=np.float64)
-    var_sum_2d = np.zeros(imshape, dtype=np.float64)
+    med_slopes = np.zeros(imshape, dtype=np.float32)
+    slope_wtd = np.zeros(imshape, dtype=np.float32)
 
     # For multiple-integration datasets, will output integration-specific
     #    results to separate file named <basename> + '_integ.fits'
```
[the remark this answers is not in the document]
dmggh: Yes, comment fixed.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -229,16 +231,21 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
 
     # Get readnoise array for calculation of variance of noiseless ramps, and
```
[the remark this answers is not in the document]
dmggh: Comment updated.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -247,53 +254,71 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+                                             dqflags.group['JUMP_DET']) != 0)
+            first_diffs_sect[ i_group-1, i_yy, i_xx ] = np.NaN
+
+            # All first differences affected by saturation and CRs have been set
+            #  to NaN, so compute the median of all non-NaN first differences.
+            median_diffs_2d[ rlo:rhi, : ] += np.nanmedian(first_diffs_sect, axis=0 )
+
+            # Calculate the slope of each segment
+            t_dq_cube, inv_var, opt_res, f_max_seg, num_seg = \
+                 calc_slope(data_sect, gdq_sect, frame_time, opt_res, rn_sect,
```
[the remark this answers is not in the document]
dmggh: The overall slope uses segment-specific slopes, and uses the median of the first differences only in the variances. From an email I sent you and Phil on 03/13/18, an email I set Karl Gordon the following day, and subsequent discussions, it was agreed that the slope for the entire dataset is a sum over all segments in all integrations:

slope_dataset = sum_over_segs_integs(slope/var)/ sum_over_segs_integs(1/var)

... where the 'slope' is the slope of the segment from the fit using the Fixsen weights, and 'var' is the variance of the slope of the segment due to both readnoise and poisson noise (as calculated using the median differences).


Similarly, the integration-specific slopes are sums over all segments in the integration:

slope_int = sum_over_segs(slope/var)/ sum_over_segs(1/var)

... where, here again, 'slope' is the fit (using the Fixsen weights) for the segment and 'var' is the total variance as calculated from the estimated median slopes.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -303,92 +328,237 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
             opt_res.cr_mag_seg[num_int, :, rlo:rhi, :] = \
                        data_diff * (dq_cr != 0)
- 
-            m_by_var_int[num_int, rlo:rhi, :] = m_by_var.reshape(sect_shape)
-            inv_var_int[num_int, rlo:rhi, :] = inv_var.reshape(sect_shape)
 
-        slope_int[num_int, :, :] = \
-            utils.calc_slope_int(slope_int, m_by_var_int, inv_var_int, num_int)
+    # Compute the final 2D array of differences.
+    median_diffs_2d /= n_int
```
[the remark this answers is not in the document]
dmggh: Due to a somewhat less than clear choice of a variable name on my part, median_diffs_2d is actually a cumulative sum over integrations.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -303,92 +328,237 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+    #   sections to calculate the variances of the slope using the estimated
+    #   median slopes from the 'First Pass'. These variances are due to Poisson
+    #   noise only, read noise only, and the combination of Poisson noise and
+    #   read noise. The integration-specific variances are 3D arrays, and
+    #   the segment-specific variances are 4D arrays . The naming convention for
+    #   the arrays:
+    #     'var': a variance
+    #     'p3': intermediate 3D array for variance due to Poisson noise
+    #     'r4': intermediate 4D array for variance due to read noise
+    #     'both4': intermediate 4D array for ocmbined variance due to both
```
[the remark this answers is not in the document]
dmggh: Fixed.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -303,92 +328,237 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+    var_both3[ var_both3 < 0.] = 0.
+
+    # Now that the segment-specific and integration-specific variances have
+    #   been calculated, the segment-specific, integration-specific, and
+    #   overall slopes will be calculated. The integration-specific slope is
+    #   calculated as a weighted average of the segments in the integration:
+    #     slope_int = sum_over_segs(slope_seg/var+seg)/ sum_over_segs(1/var_seg)
+    #  The overall slope is calculated as a weighted average of the segments in
+    #     all integrations:
+    #     slope = sum_over_integs_and_segs(slope_seg/var+seg)/
```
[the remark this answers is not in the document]
dmggh: Right (off by 1 key)

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -303,92 +328,237 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+    # Loop over data integrations to calculate integration-specific pedestal
+    for num_int in range(0, n_int):
+        opt_res.ped_int[ num_int, :, : ] = \
             utils.calc_pedestal(num_int, slope_int, opt_res.firstf_int,\
             gdq_cube, nframes, groupgap, dropframes1)
 
-        var_p_int[num_int, :, :], var_r_int[num_int, :, :] = \
-            utils.calc_vars_int( slope_int, readnoise_2d, gdq_cube, num_int, \
-            var_p_int, var_r_int)
+    # Compute 2D 'error' for primary output
```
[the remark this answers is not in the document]
dmggh: Updated.

jwst/ramp_fitting/ramp_fit.py
dmggh: No, copy() is not needed - fixed.

jwst/ramp_fitting/ramp_fit.py
dmggh: This pixeldq is not used to flag groups for any computations. It is only used for output in the integration-specific output, and for each ramp has these values:

 0 if all reads in the ramp are good
 2 ("SATURATED") if there are >0 SAT reads and no CRs in the ramp
 4 ("JUMP_DET") if there are >0 CRs and no SATs in the ramp
 6 if there >0 SAT reads and >0 CRs in the ramp

I've updated the comment accordingly.

jwst/ramp_fitting/ramp_fit.py
dmggh: Perhaps... it would then be overdone.

jwst/ramp_fitting/ramp_fit.py
dmggh: Agreed, updated.

jwst/ramp_fitting/ramp_fit.py
dmggh: Right, updated.

jwst/ramp_fitting/ramp_fit.py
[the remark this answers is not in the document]
dmggh: I agree. It is my recollection of our earlier discussion that that was the reason that we agreed to use both Poisson and readnoise.

jwst/ramp_fitting/ramp_fit.py
dmggh: Fixed.

jwst/ramp_fitting/ramp_fit.py
dmggh: Yes, and this seems to be the recommended method. Using np.logical_and() also works.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -447,8 +617,8 @@ def gls_ramp_fit(model,
             utils.get_more_info(model)
     if n_int > 1:
         # `slopes` will be used for accumulating the sum of weighted slopes.
-        slopes = np.zeros(imshape, dtype=np.float64)
-        sum_weight = np.zeros(imshape, dtype=np.float64)
+        med_slopes = np.zeros(imshape, dtype=np.float32)
```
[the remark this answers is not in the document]
dmggh: Okay.

jwst/ramp_fitting/ramp_fit.py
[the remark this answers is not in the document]
dmggh: Okay.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -925,21 +1089,20 @@ def calc_slope(data_sect, gdq_sect, frame_time, opt_res, rn_sect, gain_sect,
 
     # Section of GROUPDQ dq section, excluding bad dq values in mask
     gdq_sect_r = np.reshape(gdq_sect, (nreads, npix))
-
     mask_2d[gdq_sect_r != 0] = False  # saturated or CR-affected
+    mask_2d_init = mask_2d.copy() # initial flags for entire ramp
 
     wh_f = np.where(mask_2d == False)
```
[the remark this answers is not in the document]
dmggh: Good point, updated it.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -1086,72 +1246,84 @@ def fit_next_segment(start, end_st, end_heads, pixel_done, data_sect, mask_2d,
...
+            c_mask_2d_init = mask_2d_init.copy()
+
+            # create array: 0...nreads-1 in a column for each pixel
+            arr_ind_all = np.array( [np.arange(nreads),] *
+                                    c_mask_2d_init.shape[1]).transpose()
+            wh_c_start_all = np.zeros( c_mask_2d_init.shape[1], dtype=np.uint8)
+            wh_c_start_all[ g_pix ]= start[ g_pix]
+
+            # set to False all groups before start group
+            c_mask_2d_init[ arr_ind_all < wh_c_start_all ] = 0
```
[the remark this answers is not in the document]
dmggh: Yes, updated.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -1164,28 +1336,26 @@ def fit_next_segment(start, end_st, end_heads, pixel_done, data_sect, mask_2d,
         end_st[end_heads[all_pix] - 1, all_pix] = 0
         end_heads[all_pix] = 0
         pixel_done[all_pix] = True
-        g_pix = all_pix[variance[all_pix] > 0.]
 
-        # Ignore all pixels having no good reads, meaning the single read is bad
-        g_pix = g_pix[ total_mask_sum != 0]
+        wh_check = np.where((mask_2d_init[0, :] == 1) & (ramp_mask_sum == 1))
```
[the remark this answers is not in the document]
dmggh: Right, updated it here and elsewhere.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -1337,81 +1489,73 @@ def fit_next_segment(start, end_st, end_heads, pixel_done, data_sect, mask_2d,
...
-               (mask_sum > 0))
+    # CASE H) - the segment has a good 0th group and a bad 1st group. For the
+    #  data from the 0th good group of this segment to possibly be used as a
+    #  slope, that group must necessarily be the 0th group of the entire ramp.
+    #  It is possible to have a single 'good' group segment after the 0th group
+    #  of the ramp; in that case the 0th group and the 1st group would both have
+    #  to be CRs, and the data of the 0th group would not be included as a slope.
+    #  For a good 0th group in a ramp followed by a bad 1st group there must be
+    #  good groups later in the segment because if there were not, the segment
+    #  would already have be classified as CASE H. In this situation, since
```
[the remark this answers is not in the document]
dmggh: Yes.

jwst/ramp_fitting/utils.py
```diff
@@ -336,197 +335,235 @@ def alloc_int(n_int, imshape):
...
+    segs_beg_3_m1 = segs_beg_3.copy()
+    segs_beg_3_m1[wh_pos_3] -= 1
+    segs_beg_3_m1[ segs_beg_3_m1 < 1 ] = 1
+
+    # For a segment, the variance due to Poisson noise
+    # = slope/(tgroup * gain * (ngroups-1)),
+    #   where slope is the estimated median slope, tgroup is the group time,
+    #   and ngroups is the number of groups in the segment (minimum = 2).
+    #   Here the denominator of this quantity will be computed, which will be
+    #   later multiplied it by the estimated median slope.
```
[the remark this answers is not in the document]
dmggh: Right - thanks.

jwst/ramp_fitting/utils.py
```diff
@@ -336,197 +335,235 @@ def alloc_int(n_int, imshape):
...
+    # For a segment, the variance due to readnoise noise
+    # = 12 * readnoise**2 /(ngroups_seg**3. - ngroups_seg)/( tgroup **2.)
+    num_r3 = 12. * (rn_sect/group_time)**2.  # always >0
+    # Reshape for every group, every pixel in section
+    num_r3 = np.dstack( [num_r3] * max_seg )
+    num_r3 = np.transpose( num_r3, (2, 0, 1))
+
+    # Denominator den_r3 = 1./(segs_beg_3 **3. -segs_beg_3). The minimum number
+    #   of allowed groups is 2, which will apply if there is actually only 1
+    #   group; in this case den_r3 = 1/6. For longer segments, this value is
```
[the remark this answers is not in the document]
dmggh: Yes, updated the comment.

jwst/ramp_fitting/utils.py
dmggh: Good point. Updated the code and comments to reflect that.

jwst/ramp_fitting/ramp_fit.py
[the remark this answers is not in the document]
dmggh: As for your performance question : back in July of 2017, before there was any calculation of the variances (so those extensions did not yet exist in the output), I ran a niriss test of 2k x 2k, with 100 groups per integration, and 1 integration.  The execution time then was 98 seconds. Now, running with the same dataset, and the variance calculations, the execution time is 122 seconds, so about 24% longer

hbushouse:
Fixes #1681 #1683 #1686 #1687 #1714

PR #1929, opened 2018-04-23
Rscd error

jemorrison:
Added a check that the correct entry in the reference file is found. If it is not found then throw an exception. I also updated the Imager RSCD reference file and sent it to Misty who will get it in CRDS, then this step will work for SUBPRISM data (the reference file assumed it was SLITLESSPRISM)


hbushouse:
Updated RSCD reference file with appropriate DMS subarray value "SUBPRISM" was included in CRDS context 452, which essentially makes this code change unnecessary, but we'll keep it in just in case.

PR #1933, opened 2018-04-24
Fix history validation error

bernie-simon:
Datamodels tests were throwing a validation error when writing new history records. It's not clear why we started seeing the errors, perhaps an update to asdf. But the cause is that there were two
different data structures to hold the history records and only one of them passed validation. The solution is to convert the invalid structure to the valid one before serializing the asdf file to its
extension.

### DIFF
--- a/jwst/datamodels/fits_support.py
+++ b/jwst/datamodels/fits_support.py
@@ -377,6 +377,7 @@ def _save_history(hdulist, tree):
         history = tree['history']
     else:
         history = tree['history'].get('entries', [])
+        tree['history'] = history
 
     for i in range(len(history)):
         # There is no guarantee the user has added proper HistoryEntry records


jwst/datamodels/fits_support.py
```diff
@@ -377,6 +377,7 @@ def _save_history(hdulist, tree):
         history = tree['history']
     else:
         history = tree['history'].get('entries', [])
+        tree['history'] = history
```
nden: `AsdfFile.tree.history` is now a dictionalry, not a list. Why is it necessary to assign a list to `.tree`? 
Is there an example that failed before this change?

bernie-simon:
The asdf core schema lists it as a tree. It was failing validation

description: |
  This schema contains the top-level attributes for every ASDF file.

tag: "tag:stsci.edu:asdf/core/asdf-1.0.0"
type: object
properties:
 
  history:
    description: |
      A log of transformations that have happened to the file.  May
      include such things as data collection, data calibration
      pipelines, data analysis etc.
    type: array
    items:
      $ref: "history_entry-1.0.0"

bernie-simon:
I mean, it lists it as an array.

bernie-simon:
OK, I've removed this change,

PR #1939, opened 2018-04-24
Fix MIRI regtests

jdavies-st:
We do 2 things here:

- update the CRDS context so it matches what is currently happening on the Pandokia side
- Build the output names correctly so the tests can pass.

There's still a couple tests that are not passing due to them being broken (or evaporating) on the SVN/pandokia side, but these issues are unrelated to output names, so we'll fix these later.

### DIFF
--- a/jwst/tests_nightly/general/miri/test_linearity3.py
+++ b/jwst/tests_nightly/general/miri/test_linearity3.py
@@ -28,7 +28,7 @@ def test_linearity_miri3(_bigdata):
 
     LinearityStep.call(_bigdata+'/miri/test_linearity/jw00001001001_01109_00001_MIRIMAGE_dark_current.fits',
                        override_linearity=_bigdata+'/miri/test_linearity/lin_nan_flag_miri.fits',
-                       output_file=output_file_base
+                       output_file=output_file_base, name='linearity'
                        )
     h = pf.open(output_file)
     href = pf.open(_bigdata+'/miri/test_linearity/jw00001001001_01109_00001_MIRIMAGE_linearity.fits')
@@ -37,7 +37,7 @@ def test_linearity_miri3(_bigdata):
     newhref = pf.HDUList([href['primary'],href['sci'],href['err'],href['pixeldq'],href['groupdq']])
     result = pf.diff.FITSDiff(newh,
                               newhref,
-                              ignore_keywords = ['DATE','CAL_VER','CAL_VCS','CRDS_VER','CRDS_CTX'],
+                              ignore_keywords = ['DATE','CAL_VER','CAL_VCS','CRDS_VER','CRDS_CTX', 'R_LINEAR'],
                               rtol = 0.00001
     )
     assert result.identical, result.report()
--- a/jwst/tests_nightly/general/miri/test_sloperpipeline_1.py
+++ b/jwst/tests_nightly/general/miri/test_sloperpipeline_1.py
@@ -25,9 +25,11 @@ def test_detector1pipeline1(_bigdata):
     step.refpix.odd_even_rows = True
     step.jump.rejection_threshold = 250.0
     step.ramp_fit.save_opt = False
+    step.output_file='jw00001001001_01101_00001_MIRIMAGE'
+    step.suffix='rate'
 
-    step.run(_bigdata+'/miri/test_sloperpipeline/jw00001001001_01101_00001_MIRIMAGE_uncal.fits',
-             output_file='jw00001001001_01101_00001_MIRIMAGE_rate.fits')
+    step.run(_bigdata+'/miri/test_sloperpipeline/jw00001001001_01101_00001_MIRIMAGE_uncal.fits'
+             )
 
     # Compare calibrated ramp product
     n_cr = 'jw00001001001_01101_00001_MIRIMAGE_ramp.fits'
--- a/jwst/tests_nightly/general/miri/test_sloperpipeline_2.py
+++ b/jwst/tests_nightly/general/miri/test_sloperpipeline_2.py
@@ -24,9 +24,11 @@ def test_detector1pipeline2(_bigdata):
     step.refpix.odd_even_rows = True
     step.jump.rejection_threshold = 250.0
     step.ramp_fit.save_opt = False
+    step.output_file='jw80600012001_02101_00003_mirimage'
+    step.suffix='rate'
 
-    step.run(_bigdata+'/miri/test_sloperpipeline/jw80600012001_02101_00003_mirimage_uncal.fits',
-             output_file='jw80600012001_02101_00003_mirimage_rate.fits')
+    step.run(_bigdata+'/miri/test_sloperpipeline/jw80600012001_02101_00003_mirimage_uncal.fits'
+             )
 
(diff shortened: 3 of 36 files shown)


nden:
These look good to me but I don't know much about the naming rules and workings.

jwst/tests_nightly/general/miri/test_linearity3.py
```diff
@@ -37,7 +37,7 @@ def test_linearity_miri3(_bigdata):
     newhref = pf.HDUList([href['primary'],href['sci'],href['err'],href['pixeldq'],href['groupdq']])
     result = pf.diff.FITSDiff(newh,
                               newhref,
-                              ignore_keywords = ['DATE','CAL_VER','CAL_VCS','CRDS_VER','CRDS_CTX'],
+                              ignore_keywords = ['DATE','CAL_VER','CAL_VCS','CRDS_VER','CRDS_CTX', 'R_LINEAR'],
```
hbushouse: Why would you want to ignore the name of the linearity reference file that was used? Comparison against that keyword value will tell you when a new/different reference file has been used. That way you can confirm changes to CRDS and also then know why the image pixel values have changed.

jwst/tests_nightly/general/miri/test_sloperpipeline_1.py
```diff
@@ -25,9 +25,11 @@ def test_detector1pipeline1(_bigdata):
     step.refpix.odd_even_rows = True
     step.jump.rejection_threshold = 250.0
     step.ramp_fit.save_opt = False
+    step.output_file='jw00001001001_01101_00001_MIRIMAGE'
+    step.suffix='rate'
```
hbushouse: You probably don't even need these two lines (28-29), because they're setting the output file name to exactly the same name that the pipeline infrastructure would use on its own. But it certainly doesn't hurt to explicitly force it.

jwst/tests_nightly/general/miri/test_sloperpipeline_2.py
```diff
@@ -24,9 +24,11 @@ def test_detector1pipeline2(_bigdata):
     step.refpix.odd_even_rows = True
     step.jump.rejection_threshold = 250.0
     step.ramp_fit.save_opt = False
+    step.output_file='jw80600012001_02101_00003_mirimage'
+    step.suffix='rate'
```
hbushouse: Same comment as above.

jwst/tests_nightly/general/miri/test_srctype2.py
```diff
@@ -27,7 +27,7 @@ def test_srctype2(_bigdata):
 
 
     SourceTypeStep.call(_bigdata+'/miri/test_srctype/jw80600012001_02101_00003_mirimage_flat_field.fits',
-                        output_file=output_file_base
+                        output_file=output_file_base, suffix='sourcetypestep'
```
hbushouse: Curious as to why you're including "step" in this suffix, when all the other tests don't.

jwst/tests_nightly/general/miri/test_linearity3.py
jdavies-st: The problem is that the way the test is written is that this is machine-dependent.  It fails if we allow the comparison:
```
E              Keyword R_LINEAR has different values:
E                 a> /Users/jdavies/jwst_test_data/miri/test_linearity/lin_nan_flag_miri.fits
E                  ?  -------  ^^^^
E                 b> /data4/jwst_test_data/miri/test_linearity/lin_nan_flag_miri.fits
E                  ?    ^^^
E              Keyword R_LINEAR has different comments:
E                 a> Linearity reference file name
E                 b> L
```
This is a general problem with absolute paths in our testing environment.  We need to figure out a way to get away from this.  Any ideas?

jwst/tests_nightly/general/miri/test_srctype2.py
jdavies-st: Because that's what the previous test used for the output file name.  I didn't want to change things too much.  But for consistency sake, perhaps this would be good thing to change in an upcoming PR.

jwst/tests_nightly/general/miri/test_linearity3.py
philhodge: For reasons of privacy and security, I think only the file name should be included in the metadata for the file.  The directory path should be omitted.  That would also help with the problem of strings that are too long for the archive.

hbushouse: Oh, I see. I didn't realize/remember that some tests were using custom reference files, as opposed to the latest/greatest from CRDS. Of course you still don't really need to skip the comparison of the keyword value; just update the comparison file so that it has the keyword value with the new path.

@philhodge I see the point about privacy, in terms of paths that are personal directories, but there's also the flip side of a user doing custom work outside of the pipeline wanting to have complete, explicit history of what was used in their processing. In the Ops environment, which are the only files that will ever go into the archive catalog, the path is the standard "crds://", so there's no privacy issue and it's also a short string.

jwst/tests_nightly/general/miri/test_srctype2.py
hbushouse: 👍

jdavies-st:
Just leaving a note here for my future self and others to record how these tests can be run locally for testing.

Copy over the regtest data from jwcalibdev to a local disk and point TEST_BIGDATA to it.
```
scp -r myusername@jwcalibdev:/data4/jwst_test_data/ ~/
export TEST_BIGDATA=~/jwst_test_data/
```
Create an output directory as a temp space for pytest, otherwise it will use /tmp, if you're into that.  We can delete this later, so we name it something obvious.
```
mkdir ~/delete_me
```
Now run all the tests.  We now need the `--bigdata` flag in order to run these tests, in addition to having TEST_BIGDATA pointing to the base of the test data tree.
```
cd ~/dev/jwst/jwst/tests_nightly/miri
pytest --basetemp=~/delete_me --bigdata .
```

jdavies-st:
Finally, these are the tests that are still broken in `tests_nightly/miri`:

`test_cube_build.py`
`test_miri_ifu_wcs.py` _<-- results for this test seem to have evaporated on the pandokia side_ @nden?
`test_rscd2.py` _<-- rscd issues_
`test_sloperpipeline_1.py`_<-- rscd issues_
`test_sloperpipeline_2.py`_<-- rscd issues_

hbushouse:
`test_rscd2.py`, `test_sloperpipeline_1.py`, and `test_sloperpipeline_2.py` are now working thanks to the updated RSCD reference file included in CRDS context 452. They may still be showing differences (due to new RSCD correction), but not errors.

jwst/tests_nightly/general/miri/test_linearity3.py
jdavies-st: I will back out this change in a subsequent PR @hbushouse.  And we can deal with this issue in the pipeline code.

nden:
The input file for `test_miri_ifu_wcs.py` could not open with a data model because of validation errors. I updated it in /data4 though I'm not sure this is what's causing the test to disappear. We'll see if the test runs tomorrow.

Issue #1895, opened 2018-04-12
IFU cube WCS parity is backwards


drlaw1558:
With how we define xi,eta when building cubes (cube_build.coord.py):
radiff = ra*deg2rad - ra0;
xi = np.cos(decr)*np.sin(radiff)/h;
then it means that RA increases in the direction of greater X pixel coordinate.  The pc1_1 parity of -1 in cube_build.ifu_cube.py incorrectly describes the WCS of the data in the wrong orientation (suggesting that RA is along -X instead).  

I think this can be fixed either by making pc1_1 parity +1 in cube_build.ifu_cube.setup_IFUCube, or by setting xi=-xi in cube_build.coord.radec2std and std2radec.

jemorrison:
@hbushouse 
Howard do you have a preference on how to do this. I can just set PC1_1 back to 1.
Will that confuse folks that are used to it being -1 with imager data ?

hbushouse:
At the very least we obviously need to make sure the WCS matches the actual data. Given that the cubes are being constructed such that RA increases along the +x axis, that means PC1_1 = +1.0. However, I'll leave it up to the MIRI team to decide if that's really how they want the cubes built. When someone doing analysis does something like taking an x/y slice of the cube (i.e. a spatial slice at a given wavelength) and displaying it, will it cause confusion for them to be looking at a RA/Dec system that is backwards from normal imaging (where RA increases with -x)?

drlaw1558:
RA increases with -x for the MaNGA IFU data, and if that's also the standard for HST/JWST imaging then we should follow that standard.  I think that should be possible to achieve by simply introducing - signs on xi in radec2std and std2radec, although I don't know whether that will cause any issues for how the cube boundaries/sizes are defined.  Results looked ok in a quick test that I did.

hbushouse:
Fixed by #1941

PR #1941, opened 2018-04-25
Cube build parity

jemorrison:
Fix the IFU cubes parity that I messed up a few months ago.
PC1_1 = -1 is still used but the naxis1 dimension is constructed opposite what is was. 
The naxis1 dimension now increase in the opposite direction as ra does. 

For regression testing we need to establish a new reference IFU cube.

### DIFF
--- a/jwst/cube_build/coord.py
+++ b/jwst/cube_build/coord.py
@@ -7,77 +7,11 @@ import math
 
 
 #_______________________________________________________________________
-def V2V32RADEC(ra_ref,dec_ref,roll_ref,v2_ref,v3_ref,v2, v3):
-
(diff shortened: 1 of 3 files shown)


jemorrison:
addresses #1895

drlaw1558:
Doesn't there need to be a xi=-xi added into std2radec as well in order to keep the forward and backward transforms consistent?

jemorrison:
Yes good check David. I made that change and will make a pull request soon. It only breaks things when we go backwards which would be for blotting - outlier detection.

hbushouse:
Sorry for jumping the gun on merging last night. Glad @drlaw1558 caught that.

PR #1943, opened 2018-04-25
Regtest poc

stsci-hack:
This only updates the tests under 'general/pipelines' so that they will run to completion and perform comparisons on all products from each test.

### DIFF
--- a/jwst/tests_nightly/general/pipelines/test_coron3_1.py
+++ b/jwst/tests_nightly/general/pipelines/test_coron3_1.py
@@ -54,8 +54,7 @@ def test_coron3_pipeline1(_bigdata):
     newh = pf.HDUList([h['primary'], h['sci'], h['err'], h['dq']])
     newhref = pf.HDUList([href['primary'], href['sci'],
                          href['err'], href['dq']])
-    kws_to_ignore = ['DATE', 'CAL_VER', 'CAL_VCS', 'CRDS_VER', 'CRDS_CTX', 
-                     'R_PSFMAS']
+    kws_to_ignore = ['DATE', 'CAL_VER', 'CAL_VCS', 'CRDS_VER', 'CRDS_CTX']
 
     result = pf.diff.FITSDiff(newh,
                               newhref,
--- a/jwst/tests_nightly/general/pipelines/test_coron3_1.py
+++ b/jwst/tests_nightly/general/pipelines/test_coron3_1.py
@@ -76,7 +76,7 @@ def test_coron3_pipeline1(_bigdata):
     result = pf.diff.FITSDiff(newh,
                               newhref,
                               ignore_keywords=kws_to_ignore,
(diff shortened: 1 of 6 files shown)


stsci-hack:
There may be another test which needs to be updated in order to run.  Look for another commit to address that test before merging.

jwst/tests_nightly/general/pipelines/test_coron3_1.py
```diff
@@ -30,7 +54,8 @@ def test_coron3_pipeline1(_bigdata):
     newh = pf.HDUList([h['primary'], h['sci'], h['err'], h['dq']])
     newhref = pf.HDUList([href['primary'], href['sci'],
                          href['err'], href['dq']])
-    kws_to_ignore = ['DATE', 'CAL_VER', 'CAL_VCS', 'CRDS_VER', 'CRDS_CTX']
+    kws_to_ignore = ['DATE', 'CAL_VER', 'CAL_VCS', 'CRDS_VER', 'CRDS_CTX', 
+                     'R_PSFMAS']
```
hbushouse: I still don't like this idea of ignoring the reference file name keywords. When a new ref file gets installed in CRDS and the test starts showing differences in the results, you won't know why (because it won't tell you that the ref file changed). You'll have to do more detailed sleuthing on your own to figure out why things suddenly changed. Once you get a clean run and the only difference is the name of ref file, just update the comparison file so that it has that new ref file name.

jwst/tests_nightly/general/pipelines/test_image2pipeline_1.py
```diff
@@ -20,7 +20,7 @@ def test_image2pipeline1(_bigdata):
                         )
 
     h = pf.open('jw00001001001_01101_00001_mirimage_cal.fits')
-    href = pf.open(_bigdata+'/miri/test_image2pipeline/jw00001001001_01101_00001_MIRIMAGE_uncal_MiriSloperPipeline_Image2Pipeline.fits')
+    href = pf.open(_bigdata+'/miri/test_image2pipeline/jw00001001001_01101_00001_mirimage_cal_ref.fits')
```
hbushouse: Yeah, I recently (finally) cleaned up the name of that comparison file over on the pandokia side.

jwst/tests_nightly/general/pipelines/test_coron3_1.py
stsci-hack: The best solution would be to not include the absolute path when recording the name of this specific filename in the metadata/keyword in the first place.  This would actually require a change to the code itself which is beyond the scope of these changes.  Until then, I can back out this change.

stsci-hack:
As a side note:  the input ASN file for the nircam_calimage3 test has been updated in our input data area to remove absolute paths from the member filenames.  This will allow these tests to verify (in one known case) whether the code will actually find the input members when run in the test directory's configuration.

PR #1945, opened 2018-04-25
Updates to nirspec regtests

stsci-hack:
This only updates the tests under 'general/nirspec' so that they should run to completion and perform comparisons on all products from each test.

### DIFF
--- a/jwst/tests_nightly/general/nirspec/test_extract2d_msa_8slits.py
+++ b/jwst/tests_nightly/general/nirspec/test_extract2d_msa_8slits.py
@@ -32,13 +32,13 @@ def test_extract2d_nrs_msa(_bigdata):
         # This is needed because we can't run the step with the
         # MSA configuration file not in the current dir.
         # When this is fixed, the copy can be removed.
-        for f in input_file:
+        for f in input_files:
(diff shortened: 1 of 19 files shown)


jwst/tests_nightly/general/nirspec/test_extract2d_msa_8slits.py
```diff
@@ -32,13 +32,13 @@ def test_extract2d_nrs_msa(_bigdata):
         # This is needed because we can't run the step with the
         # MSA configuration file not in the current dir.
         # When this is fixed, the copy can be removed.
-        for f in input_file:
+        for f in input_files:
```
jdavies-st: Nice catch!  Is this also one that wasn't showing up in pandokia as an error?

hbushouse: This test doesn't exist at all on the pandokia side. Where did it come from? Did someone add it as a new test just recently only within Jenkins?

nden: This test should be removed. IIRC I tried adding it unsuccessfully to svn some time ago and probably included it up by mistake here.

PR #1947, opened 2018-04-26
Reorganize IFU-specific keywords

hbushouse:
As part of a larger effort to reorganize and clean-up some of the keyword schemas, this updates the IFUCubeModel schema to place all of the cube-specific keyword definitions within a new meta subsection called "ifu". Updates to cube_build modules are also made to point to the new meta tree locations of these keywords.


jemorrison:
Looks good to me.

PR #1948, opened 2018-04-26
fix extract_2d

nden:
The structure of the WCS pipeline was changed in recent updates - `wcs.get_transform("slit_frame", "msa_frame")` returns now a `Slit2Msa` model which should not be indexed.


nden:
There are no extract_2d Travis tests. Merging so it gets in the nightly build and runs on the regression tests.

PR #1950, opened 2018-04-26
Fix NIRISS regtests

jdavies-st:
Also do the following:

- update CRDS_CONTEXT to jwst_0454.pmap
- remove a bad NIRSpec test


jwst/tests_nightly/general/niriss/test_niriss_sloper.py
```diff
@@ -24,13 +24,14 @@ def test_niriss_detector1(_bigdata):
     step.refpix.side_gain=1.0
     step.jump.rejection_threshold = 250.0
     step.ramp_fit.save_opt = False
+    step.output_file = 'jw00034001001_01101_00001_NIRISS_ramp.fits'
+    step.suffix = 'ramp'
```
hbushouse: Because this is calling the whole calwebb_detector1 pipeline, there isn't a single output product type and hence I don't see how you can specify a suffix. At the very least the suffix value should probably be set to the product type for the main pipeline output product, which is "rate". You may also get "ramp" and "rateints" products.

hbushouse: There should be comparisons with 2 or 3 different output products: ramp, rate, and (possibly) rateints.

PR #1955, opened 2018-04-26
add model blender

jemorrison:
Added model blender to cube_build.
I still need to add functions to override some of the values (for example, Channel)

@hbushouse 
@stsci-hack

### DIFF
--- a/jwst/cube_build/ifu_cube.py
+++ b/jwst/cube_build/ifu_cube.py
@@ -684,7 +684,7 @@ class IFUCubeData(object):
         IFUCube.meta.filename = self.output_name
         
         IFUCubeData.blend_output_metadata(self,IFUCube)
-
+#______________________________________________________________________
         if self.output_type == 'single':
             with datamodels.open(self.input_models[j]) as input:
                 # define the cubename for each single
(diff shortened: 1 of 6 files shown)


jwst/cube_build/ifu_cube.py
```diff
@@ -685,14 +680,13 @@ def setup_IFUCube(self,j):
         err_cube = np.zeros((naxis3, naxis2, naxis1))
 
         IFUCube = datamodels.IFUCubeModel(data=data, dq=dq_cube, err=err_cube, weightmap=idata)
-
         IFUCube.update(self.input_models[j])
-
         IFUCube.meta.filename = self.output_name
+        
+        IFUCubeData.blend_output_metadata(self,IFUCube)
```
hbushouse: I see `IFUCube` referenced in preceding lines of the code, but I don't see `IFUCubeData`, at least not in the diffs displayed here. Does `IFUCubeData` exist somewhere?

jemorrison:
added GRATING,FILTER and BAND = 'MULTIPLE' is there are more than 2 values
The user should look the HDRTAB for values

jwst/cube_build/ifu_cube.py
```diff
@@ -685,28 +681,53 @@ def setup_IFUCube(self,j):
...
+# fill in Channel, Band for MIRI
+        if self.instrument == 'MIRI':
+            # fill in Channel output meta data
+            num_ch = len(self.list_par1)
+            IFUCube.meta.instrument.channel = self.list_par1[0]
+            num_ch = len(self.list_par1)
+            for j in range (1, num_ch):
+                IFUCube.meta.instrument.channel =  IFUCube.meta.instrument.channel + str(self.list_par1[j])
+
+            # fill in Band output meta  data
```
hbushouse: I've confirmed by looking at the model_blenders rules that the "multi" rule will do exactly what we want here (and exactly what you're doing in this block of code. The "multi" rule determines the number of unique values in the input list and if there's only 1 unique value, it sets the output keyword to that value. If there are multiple unique values, it sets the output keyword value to "MULTIPLE". So you can remove this block of code for the BAND keyword and add "blend_rule: multi" to the BAND keyword definition in core.schema.yaml (see below).

jwst/cube_build/ifu_cube.py
```diff
@@ -685,28 +681,53 @@ def setup_IFUCube(self,j):
...
+            num_band = len(set(self.list_par2))
+            if num_band == 1:  
+                IFUCube.meta.instrument.band = self.list_par2[0]
+            else:
+                IFUCube.meta.instrument.band = 'MULTIPLE'
+#______________________________________________________________________
+# fill in Grating, Filter for  NIRSPEC
```
hbushouse: Same comment here as for BAND: can remove this code for GRATING and FILTER and just add "blend_rule: multi" to these keywords in core.schema.yaml.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -371,7 +371,7 @@ properties:
               F212N, F227W, F2300C, F250M, F2550W, F2550WR, F277W, F290LP, F300M,
               F322W2, F335M, F356W, F360M, F380M, F410M, F430M, F444W,
               F460M, F480M, F560W, F770W, FLENS, FND, FNDP750L, GR150C,
-              GR150R, OPAQUE, P750L, WL3, WLP4, N/A]
+              GR150R, OPAQUE, P750L, WL3, WLP4, N/A, MULTIPLE]
             fits_keyword: FILTER
             blend_table: True
```
hbushouse: Add "blend_rule: multi" here, after "blend_table".

hbushouse: And do the same for GRATING and BAND below.

hbushouse:
@jemorrison Do you want to try to get this into B7.1.3 yet? If so, we need to finalize the changes and get it merged within the next 24 hours, so that we have to time for regression testing and possible iterations on bugs, before we have to create a build on Friday.

jemorrison:
I do not plan any more changes on this PR for build 7.1.3 (unless a typo or bug is found)

hbushouse:
Merging to get into the build that will be used for tonight's regression test run.

PR #1956, opened 2018-04-27
CubeModel bbox should be 2D

nden:
For a stack of images when the WCS bounding box is retrieved from the shape of the data it should still have length 2 because the input coordinate frame has 2 axes.

### DIFF
--- a/jwst/assign_wcs/fgs.py
+++ b/jwst/assign_wcs/fgs.py
@@ -69,10 +69,20 @@ def imaging_distortion(input_model, reference_files):
         bb = transform.bounding_box
     except NotImplementedError:
         shape = input_model.data.shape
-        # Note: Since bounding_box is attached to the model here it's in reverse order.
-        bb = ((-0.5, shape[0] - 0.5),
-              (-0.5, shape[1] - 0.5))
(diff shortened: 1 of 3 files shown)


jwst/assign_wcs/fgs.py
```diff
@@ -69,10 +69,20 @@ def imaging_distortion(input_model, reference_files):
...
     except NotImplementedError:
         shape = input_model.data.shape
-        # Note: Since bounding_box is attached to the model here it's in reverse order.
-        bb = ((-0.5, shape[0] - 0.5),
-              (-0.5, shape[1] - 0.5))
-    transform.bounding_box = bb
+        # Note: Since bounding_box is attached to the model here                                                                 
+        # it's in reverse order.                                                                                                 
+        # If the input is a CubeModel, need to set bbox to the                                                                   
+        # size of 2nd and 3rd dimension.
```
hbushouse: Might want to elaborate a bit on the explanation in the comment to indicate why this is the case, e.g. because a CubeModel is always treated as a stack (in dimension 1) of 2D images, as opposed to actual 3D data.

PR #1958, opened 2018-04-27
Merge recent SVN changes to nightly tests into repo

jdavies-st:


jdavies-st:
Tested this locally.  Will merge and run Jenkins build to see where we are.  Please do comment here if there's any problems with this PR.

PR #1959, opened 2018-04-28
Nirspec bug fixes

nden:
This requires a bugfix in astropy and should be merged after astropy/astropy#7411 is merged.

There are three substantial changes in this PR:

- The computation of slit-y positions was done through fitting a line. This is now replaced 
   by linear interpolation.
- Similarly a relationship `lambda(n)`  was established by fitting a line. Now this is replaced with interpolation.
- The equation computing the refraction index from angles was changed to match the one in the  IDT team code.

This improves the prism wavelength computation by several orders of magnitude.


nden:
related to #1732

PR #1970, opened 2018-05-01
Fix regression tests for MIRI, NIRCam and NIRISS

jdavies-st:


jwst/tests_nightly/general/niriss/test_photom.py
```diff
@@ -14,22 +14,16 @@
...
     suffix = 'photom'
     output_file_base, output_file = add_suffix('photom1_output.fits', suffix)
-
-    try:
-        os.remove(output_file)
-    except:
-        pass
-
-
+    override_photom = _bigdata+'/niriss/test_photom/jwst_niriss_photom_b7a.fits'
```
hbushouse: We should go back and investigate some time why it is that this test is using an override for the photom ref file. I suspect that the test was created a long time ago, before suitable/complete/correct photom ref files were in CRDS for all instruments, so an override was used. But now all instruments and modes should have valid photom ref files in CRDS and hence the test should be using the standard one.

Issue #168, opened 2016-08-18
Documentation (DUH!)


stscieisenhamer:
This actually means producing docs that are not automatically created from code.

stscieisenhamer:
= Implementation
- [ ] Rule concept
- [ ] Association concept
- [ ] generate concept

hbushouse:
INS folks would like additional documentation on how to use command-line tools to generate ASN tables (see https://jira.stsci.edu/browse/JP-299).

PR #1972, opened 2018-05-01
Update association documentation for asn_from_list [ci skip]

stscieisenhamer:
WIP #168 
[WIP JP-299](https://jira.stsci.edu/browse/JP-299)

### DIFF
--- a/docs/jwst/associations/asn_from_list.rst
+++ b/docs/jwst/associations/asn_from_list.rst
@@ -83,7 +83,7 @@ To create a Level3 association, use the following command:
 
 The `-o` option defines the name of the association file to create.
 
-The `--product-name` will set the `name` field which the Level3
+The `--product-name` will set the `name` field that the Level3
 calibration code will use as the output name. For the example, the
 output files created by `calwebb_image3`, or other Level3 pipelines,
 will all begin with **l3_results**.
@@ -94,8 +94,30 @@ association, with the presumption that all files will be combined.
 For coronagraphic or AMI processing, set the `exptype` of the
 exposures that are the PSF reference exposures to `psf`.  If the
 PSF files are not in the `members` list, edit the association and add
-them as members.
-  
+them as members. An example product with a psf exposure would look
+like::
+
+  "products": [
+      {
+          "name": "jw99999-o001_t14_nircam_f182m-mask210r",
+          "members": [
+              {
+                  "expname": "jw99999001001_011001_00001_nircam_cal.fits",
+                  "exptype": "science"
+              },
+              {
+                  "expname": "jw99999001001_011001_00002_nircam_cal.fits",
+                  "exptype": "science"
+              },
+              {
+                  "expname": "jw99999001001_011001_00003_nircam_cal.fits",
+                  "exptype": "psf"
+              }
+          ]
+      }
+  ]
+
+
 API
 ---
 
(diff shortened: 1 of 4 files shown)


docs/jwst/associations/asn_from_list.rst
```diff
@@ -0,0 +1,108 @@
...
+
+To create a Level3 association, use the following command:
+
+.. code-block:: shell
+
+   asn_from_list -o l3_asn.json --product-name l3_results *.fits
+
+The `-o` option defines the name of the association file to create.
+
+The `--product-name` will set the `name` field which the Level3
```
hbushouse: For proper human parsing, change "... field which the Level3" to either "field **that** Level3" or "field**,** which the Level3"

docs/jwst/associations/asn_from_list.rst
```diff
@@ -0,0 +1,108 @@
...
+will all begin with **l3_results**.
+
+The list of files will all become `science` members of the
+association, with the presumption that all files will be combined.
+
+For coronagraphic or AMI processing, set the `exptype` of the
+exposures that are the PSF reference exposures to `psf`.  If the
+PSF files are not in the `members` list, edit the association and add
+them as members.
+
```
hbushouse: How about an example of at least a simple level3 ASN (as was included for the level2).

stscieisenhamer:
Question on overall documentation: jdocs seems to have taken the route of calling the different levels of calibration "stages". Should our documentation start converting to that also?

sosey:
I dunno, given that our code is tied to the requirements that are all specified with "level" I find that more confusing. Though HST, JWST, and WFIRST all have slightly different definitions of level x processing, so it's all somewhat confusing anyways :woman_shrugging:

hbushouse:
The "stages" terminology was invented for user-facing docs, so that it would correspond to the numbers that appear at the end of the pipeline module names, i.e. stage 1 is calwebb_detector1, stage 2 is calwebb_xxxx2, and stage 3 is calwebb_xxxx3. Folks were constantly getting confused by all the level-1a, level-1b, level-2a, level-2b, level-2c, level-3 terminology.

The use of "levels" in requirements is irrelevant to users, because users will never see or know anything about our pipeline requirements.

sosey:
right, but have we established who exactly the audience for our docs are? Are the docs we are writing for the pipeline mainly meant for the scientist users or the developers? If the teams are writing jdocs, I would guess our docs are more for developers trying to understand the code? (fwiw I don't really care about the naming, but I am interested in the distinction about how the docs are written and who they are meant for)

hbushouse:
Internal folks (developers, INS scientists, INS RIA's, etc.) are expected to know and understand both terminologies. That's why they're paid the big money ;-)  Docs like these, which explain how to use our software, are obviously intended for users. This is true for the docs that explain the inner workings of each pipeline step. We're not targeting developers who need to know detailed info about API's, argument lists, data types (or at least not only them). The main targets are users in INS, so that they can verify that we've written it to match the algorithm specs, and users in the greater external community. The JDocs only present high-level, cursory info. They all link to our docs for the details.

sosey:
So our sphinx jwst docs are basically replacing the hst data handbooks...if that's the case I think  they need lots more flushing out and we should get the teams involved in updating the docs themselves to make sure they are understandable to the wider audience, they could do this as they test the code itself to make sure it's compliant with their specs :smile:

stscieisenhamer:
<troll> No requirement for documentation I believe </troll>
Regardless, jwst docs are user-facing, jdocs are user-facing, so seems like consistency there would be a good thing.

This documentation business sounds like a grand opportunity for a calibration group-wide sprint...guessing post 7.2 delivery...

PR #1982, opened 2018-05-03
Strip abspath from ASNTABLE keyword value

### LINKED ISSUE #1940: Strip path from ASNTABLE keyword value
DMS has encountered issues with saving the value of the ASNTABLE keyword in the archive catalog, due to the fact that the full directory path is included with the file name, which in turn overfills the archive catalog field for this keyword (see https://www.ess.stsci.edu/prsystem/servlet/prbrowse/dr.4566). They would like to have the directory path removed from the value, as is apparently already done with the value of ASNPOOL keyword.

jdavies-st:
Resolves #1940.

### DIFF
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -197,13 +197,14 @@ class ModelContainer(model_base.DataModel):
 
         filepath = op.abspath(op.expanduser(op.expandvars(filepath)))
         basedir = op.dirname(filepath)
+        filename = op.basename(filepath)
         try:
             with open(filepath) as asn_file:
                 asn_data = load_asn(asn_file)
         except AssociationNotValidError:
             raise IOError("Cannot read ASN file.")
 
-        # make a list of all the input FITS files
+        # make a list of all the input files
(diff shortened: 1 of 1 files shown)


jwst/datamodels/container.py
```diff
@@ -214,15 +215,9 @@ def from_asn(self, filepath, **kwargs):
             self.meta.asn_table._instance, asn_data
         )
 
-        # populate the output metadata with the output file from the ASN file
-        # Should remove the following lines eventually
-        self.meta.resample.output = str(asn_data['products'][0]['name'])
-        self.meta.table_name = str(filepath)
-        self.meta.pool_name = str(asn_data['asn_pool'])
-        self.meta.targname = str(asn_data['target'])
```
hbushouse: I assume that the removal of targname, program, asn_type, and asn_rule is part of the effort to clean out rogue keyword definitions?

jdavies-st: Yes.  These ones were easy as they were not being used anywhere.  Will clean up the schemas in a separate PR.

Issue #1985, opened 2018-05-04
Change NRC_GRISM to NRC_WFSS


hbushouse:
SDP is working on updates that will change the EXP_TYPE value 'NRC_GRISM' that is used in all front-end systems for the NIRCam grism (Wide-Field Slitless Spectroscopy) mode to 'NRC_WFSS', for the sake of consistency with the 'NIS_WFSS' EXP_TYPE used for NIRISS WFSS mode. All references to 'NRC_GRISM' in Cal pipeline code and data model schemas should be updated to include the value 'NRC_WFSS' (i.e. we want Cal code to accept **both** values, for backwards compatibility with existing data products and test files).

hbushouse:
A quick check through the code base shows that `NRC_GRISM` appears in:

- [ ] assign_wcs/assign_wcs.py
- [ ] assign_wcs/nircam.py
- [ ] background/background_step.py
- [ ] datamodels/wcs_ref_models.py
- [ ] extract_1d/extract.py
- [ ] extract_2d/extract_2d.py
- [ ] outlier_detection/outlier_detection_step.py
- [ ] photom/photom.py
- [ ] pipeline/calwebb_spec2.py
- [ ] datamodels/schemas/core.schema.yaml
- [ ] datamodels/schemas/keyword_exptype.schema.yaml
- [ ] datamodels/schemas/keyword_pexptype.schema.yaml

hbushouse:
extract_1d/extract.py was updated in #1987

stscieisenhamer:
Added the general `associations` to the above checklist.

Also question: Is there a change to `NRC_TSGRISM` or does this stay the same?

hbushouse:
`NRC_TSGRISM` is staying the same, because it's consistent with the name of the observing template "Time Series Grism".

sosey:
see https://github.com/STScI-JWST/jwst/pull/2037

Issue #1881, opened 2018-04-06
Task: Update datamodels and steps for new integration times table


hbushouse:
SDP is adding a new table extension to all level-1b products that will contain time stamps for each integration within an exposure (see https://jira.stsci.edu/browse/JSDP-387). While primarily intended for use with Time Series exposures, the table will be added to all exposures and is expected to be carried along through all stages of cal pipeline processing.

In order to propagate the table through all stages of pipeline processing we will need to update all relevant science data product data models (e.g. RampModel, ImageModel, SlitModel, etc.) to define this table extension. We will also need to update all calibration steps that create their output(s) as a new datamodel type so that they explicitly copy the integration times table over from the input model to the output model. This includes steps like `ramp_fit`, `extract_2d`, `cube_build`, `resample`, and `resample_spec`.

Propagating this table through the various pipeline stages is vital in order to make it available to level-3 TSO steps that need it, like `tso_photometry` (#1619), `extract_1d` (#1620), and `white_light` (#1621).

This is high priority for Build 7.2, in order to satisfy the TSO processing requirement.

hbushouse:
Chris Heller indicates that each exposure segment will contain integration time table entries that cover at least the integrations contained in that segment, and often a few more at either the beginning or end of the table (because the way science data is downlinked in packets is not always broken on boundaries that match the start or end of an integration). So any pipeline task that needs to use the time stamps will first need to query the INTSTART and INTEND keywords to determine which integrations are contained in the segment being worked on and then use the corresponding entries in the integration times table.

philhodge:
I'm closing this for the time being, but it may be necessary to re-open later if keywords or column names need to be changed.

hbushouse:
Given that the DMSWG has decided that the INT_TIMES table does NOT need to be propagated beyond level-1b products for exposures taken in non-TSO modes, we don't need to update nearly as many of the data models or steps to propagate it. We only need to propagate it for TSO exposures. So that means we only need to propagate it through the following steps:

- ramp_fit: Propagate INT_TIMES table to the ``rateints`` product. The corresponding data model is ``CubeModel``.
- calwebb_image2: For imaging mode TSO, all steps in ``calwebb_image2`` pass or copy the input ``CubeModel`` as a ``CubeModel``, without creating a new model type for the output, so the output ``CubeModel`` that is saved to the ``calints`` product should naturally contain the INT_TIMES table from the input ``rateints`` product. Hence no steps or data models need to be updated.
[...]

PR #1987, opened 2018-05-04
The extract_1d step should copy data from the integration times table

philhodge:
Function `populate_time_keywords` was added to the `extract_1d` step.  This copies the start, middle, and end times for each integration from entries in the INT_TIMES table in the input file to header keywords in the output file.

Various schema files were modified to include the INT_TIMES table or the new keywords in the `MultiSpecModel`.  core.schema.yaml was also modified to include the start and end integration numbers for segmented data, which are needed to find the relevant rows in the INT_TIMES table.

See issue #1881.

The extract_1d step was also updated to treat EXP_TYPE = "NRC_WFSS" the same as "NRC_GRISM".  See issue #1985.

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -591,6 +591,16 @@ properties:
             type: integer
             fits_keyword: EXSEGTOT
             blend_table: True
+          integration_start:
+            title: Integration number at start of segment
+            type: integer
+            fits_keyword: INTSTART
+            blend_table: True
+          integration_end:
+            title: Integration number at end of segment
+            type: integer
+            fits_keyword: INTEND
+            blend_table: True
           noutputs:
             title: Number of detector outputs used
             type: integer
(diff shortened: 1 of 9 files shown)


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -591,6 +591,16 @@ properties:
...
+          integration_start:
+            title: Integration number at start of segment
+            type: integer
+            fits_keyword: INTSTART
+            blend_table: True
+          integration_end:
+            title: Integration number at end of segment
+            type: integer
+            fits_keyword: INTEND
+            blend_table: True
```
hbushouse: Whoops, looks like we've both made changes to add these keywords (see #1989). They're very close to being identical changes, but I've used titles in #1989 that match the specs for what's being added to the dictionary (https://jira.stsci.edu/browse/JWSTKD-216).

philhodge: Yes, your changes should override mine.

PR #1993, opened 2018-05-07
The extract_1d step should copy data from the integration times table

philhodge:
The file core.schema.yaml was updated to agree with the changes in PR #1989.


jwst/extract_1d/extract_1d_step.py
```diff
@@ -58,7 +58,8 @@ def process(self, input):
                 self.log.debug("Input contains %d items", len(input_model))
                 result = datamodels.ModelContainer()
                 for model in input_model:
-                    if model.meta.exposure.type in ['NIS_WFSS', 'NRC_GRISM']:
+                    if (model.meta.exposure.type in
+                                ['NIS_WFSS', 'NRC_GRISM', 'NRC_WFSS']):
```
hbushouse: Might be nice to put this list into a macro (is that the correct term for python?) such as `WFSS_EXPTYPES`, so that it only has to be defined in one place, which then makes future maintenance and updates easier.

Issue #1969, opened 2018-05-01
Reference file overrides need a URI when recorded in the header keyword


jdavies-st:
When a Step is run with a reference file override, the full path of that reference file is stored in the header keyword of the output file. This makes the output file not very portable.  Would it be good to have some sort of $OREF equivalent for the JWST pipeline?

This has come up in testing where the "truth" comparison data is produced using an override, and then a test is run and fitsdiff run to compare to the truth data.  If the truth data was produced on a different system from the test data, then the full path to the override reffile will be different.

hbushouse:
The concept of an environment variable like `$oref` only works when reference files are stored in a standard place, e.g. /grp/crds/jwst/. When overrides are used, the ref file is no longer in a standard place and hence an environment variable to represent the path is not feasible. In the past when I've encountered the situation described above, once the test result has been shown to be identical to the "truth" file in every way other than the path to the ref file, then I just update the "truth" file so that it now contains the new path and does not show any more differences in the future.

jdavies-st:
Could we at least strip the absolute path from the reffile override?  That way at least if the reference file lives with the input data it can be portable?

hbushouse:
That seems reasonable. If the path is not the standard crds:// (or whatever the exact syntax is), then strip it before it gets stored in the relevant R_xxxxx keyword. I think that may happen in the CRDS server code somewhere and hence the change would need to be done on that side. @jaytmiller ?

jdavies-st:
I like that.

I think that functionality could be done on the crds_client.py in `stpipe`.

jdavies-st:
And then we can just update the documentation to indicate, "if you are using a reffile override, it should live with the input data so that it is portable."

jaytmiller:
That's the correct syntax.   I think all of the relevant crds:// handling occurs in stpipe.crds_client and stpipe.step,  it's really a JWST-specific syntax.

The relevant code is probably here in stpipe:

    def get_reference_file(self, input_file, reference_file_type):
        """
        Get a reference file from CRDS.

        If the configuration file or commandline parameters override the
        reference file, it will be automatically used when calling this
        function.

        Parameters
        ----------
        input_file : jwst.datamodels.ModelBase instance
            A model of the input file.  Metadata on this input file
            will be used by the CRDS "bestref" algorithm to obtain a
            reference file.

        reference_file_type : string
            The type of reference file to retrieve.  For example, to
            retrieve a flat field reference file, this would be 'flat'.

[...]

PR #1995, opened 2018-05-07
Strip abspath from reffiles stored in R_* header keywords for override

jdavies-st:
Now the overridden reference file stored in the keyword will be

`my_override_reffile.fits`

instead of

`/path/to/my_override_reffile.fits`

This improves portability of testing.


jdavies-st:
Resolves #1969.

PR #2006, opened 2018-05-08
Pathloss 2d

stscirij:
Bug fix for NRS_IFU data type. Also saves the computed 2-D IFU wavelength array in the output IFUImageModel data model.

### DIFF
--- a/jwst/pathloss/path_loss.py
+++ b/jwst/pathloss/path_loss.py
@@ -99,7 +99,7 @@ def calculate_pathloss_vector(pathloss_refdata, pathloss_wcs, xcenter, ycenter):
         crval1 = pathloss_wcs.crval1
         cdelt1 = pathloss_wcs.cdelt1
         for i in np.arange(wavesize):
-            wavelength[i] = crval1 +(float(i) - crpix1)*cdelt1
+            wavelength[i] = crval1 +(float(i+1) - crpix1)*cdelt1
         return wavelength, pathloss_refdata
     #
     # pointsource.data is 3-d, so we have to extract a wavelength vector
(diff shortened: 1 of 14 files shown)


jwst/pathloss/path_loss.py
```diff
@@ -93,7 +99,7 @@ def calculate_pathloss_vector(pathloss_refdata, pathloss_wcs, xcenter, ycenter):
         crval1 = pathloss_wcs.crval1
         cdelt1 = pathloss_wcs.cdelt1
         for i in np.arange(wavesize):
-            wavelength[i] = crval1 +(float(i+1) - crpix1)*cdelt1
+            wavelength[i] = crval1 +(float(i) - crpix1)*cdelt1
```
hbushouse: Since CRPIX is 1-indexed doesn't that also then require 1-indexing of the pixel numbers? So why the reversal here from 1-indexed to 0-indexed?

jwst/pathloss/path_loss.py
```diff
@@ -103,7 +109,7 @@ def calculate_pathloss_vector(pathloss_refdata, pathloss_wcs, xcenter, ycenter):
         crval3 = pathloss_wcs.crval3
         cdelt3 = pathloss_wcs.cdelt3
         for i in np.arange(wavesize):
-            wavelength[i] = crval3 +(float(i+1) - crpix3)*cdelt3
+            wavelength[i] = crval3 +(float(i) - crpix3)*cdelt3
```
hbushouse: Same question as above.

hbushouse:
Fixes a bug introduced by #1883 that prevented the NIRSpec IFU branch of the correction from executing.

stscirij:
Fixes #1152

stscirij:
Fixes #1152

Issue #2005, opened 2018-05-08
StrayLightStep method missing


stscieisenhamer:
# Issue

The following occurs:
```
>>> from jwst.straylight import StraylightStep
>>> sls = StraylightStep()
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
<ipython-input-2-933888786a79> in <module>()
----> 1 sls = StraylightStep()

~/Documents/ssbdev/jwst/jwst/stpipe/step.py in __init__(self, name, parent, config_file, _validate_kwds, **kws)
    278             spec = self.load_spec_file()
    279             kws = config_parser.config_from_dict(
--> 280                 kws, spec, root_dir=dirname(config_file or ''))
    281
    282         if name is None:

~/Documents/ssbdev/jwst/jwst/stpipe/config_parser.py in config_from_dict(d, spec, root_dir)
    194
    195     if spec:
--> 196         validate(config, spec, root_dir=root_dir)
    197     else:
    198         config.walk(string_to_python_type)

~/Documents/ssbdev/jwst/jwst/stpipe/config_parser.py in validate(config, spec, section, validator, root_dir)
    255
    256         if len(messages):
--> 257             raise ValidationError('\n'.join(messages))
    258     finally:
    259         config.main.configspec = orig_configspec

ValidationError: Config parameter 'method': missing
```

If the comment part of the `method` definition in `StraylightStep.spec` is removed, the error goes away. Implication is that the line is too long for the `method` definition.

hbushouse:
I assume this is based on the recently updated straylight code in #1887?

stscieisenhamer:
current master as of the filing of this issue.

hbushouse:
I agree that even if it did work to have such a long comment as:
```
#Algorithm to use to determine straylight correction, default is to use the Modified Shepard Method for weighting (ModShepard). The other option (Nearest) uses the nearest gap pixels in a science pixels row to find the correction.
```
this is not the appropriate place in the code to include such details. Those should go into the code itself, in the form of comments. I would propose modifying the comments on all of the spec entries to just:
```
         method = option('Nearest', 'ModShepard', default='ModShepard') # Algorithm method
         roi = float(default = 50.0) # Region of interest
         power = float(default = 1.0) # Power of weighting function
```
There's no need to include the default value in the comment; it's already right there in the same line in the definition of the argument itself.

jemorrison:
Ok I will make the change

[...]

PR #2008, opened 2018-05-08
fixed method comment

jemorrison:
fixed comments for method,roi and power


hbushouse:
Merging so that it gets into the overnight build.

hbushouse:
Fixes #2005

PR #2009, opened 2018-05-09
Make output_use_model default for CubeBuildStep

### LINKED ISSUE #1978: Output files have wrong name for Cube_build
# Issue
When running cube_build and providing an output name with --output_file 
the resulting MIRI IFUCubes have numbers inserted in the names instead of
the channel number.
For example:
```
strun cube_build.cfg MIRM103-Q0-SHORT_495.fits --output_file=test
```
produces `test_0_s3d.fits` and `test_1_s3d.fits`.
But it should produce `test_ch1_s3d.fits` and `test_ch2_s3d.fits` 

The files names are being set by the step class function I believe. 
@stscieisenhamer 

# Implementation
- [x] bump version
- [x] documentation
- [x] tests
- [x] code
- [x] tests
- [x] documentation
- [x] bump version

stscieisenhamer:
Resolves #1978 

Other changes:
- Update an NRC test to check for the correct files.

### DIFF
--- a/jwst/cube_build/__init__.py
+++ b/jwst/cube_build/__init__.py
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 from .cube_build_step import CubeBuildStep
 
-__version__ = '0.8.0'
+__version__ = '0.9.0'
--- a/jwst/cube_build/cube_build_step.py
+++ b/jwst/cube_build/cube_build_step.py
@@ -45,6 +45,7 @@ SHORT,MEDIUM,LONG, or ALL
          single = boolean(default=false) # Internal pipeline option used by mrs_imatch and outlier detection
          output_type = option('band','channel','grating','multi',default='band') # Type of output cube to create. Options = band,channel,grating, multi
          search_output_file = boolean(default=false)
+         output_use_model = boolean(default=true) # Use filenames in the output models
        """
     reference_file_types = ['cubepar','resol']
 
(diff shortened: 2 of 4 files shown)


jwst/cube_build/__init__.py
```diff
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 from .cube_build_step import CubeBuildStep
 
-__version__ = '0.8.0'
+__version__ = '0.9.0'
```
hbushouse: Might as well go ahead and set that to '0.9.3' in anticipation of the B7.1.3 delivery this week.

jwst/cube_build/cube_build_step.py
```diff
@@ -45,6 +45,7 @@ class CubeBuildStep (Step):
          single = boolean(default=false) # Internal pipeline option used by mrs_imatch and outlier detection
          output_type = option('band','channel','grating','multi',default='band') # Type of output cube to create. Options = band,channel,grating, multi
```
hbushouse: I guess this comment line isn't quite long enough to cause trouble with parsing the params?

hbushouse:
Because this fix only applies to off-line processing where a user specifies one or more non-default parameters (in this case `--output_file`), it isn't important to get this to DMS in a hurry, but if you feel it's ready and OK to be included in B7.1.3, we might as well do so.

stscieisenhamer:
This should be good to go for 7.1.3. If there is time to go through the full regression, I can go ahead and merge.

PR #2010, opened 2018-05-09
Fix straylight typo in log message

hbushouse:
Change '62f' format to '6.2f' (62 was a little excessive for values that are only in the hundreds!)


jemorrison:
Looks good - good catch.

Issue #1837, opened 2018-03-28
ami_analyze takes forever to run on full-frame exposures


hbushouse:
While DMS was testing `calwebb_ami3` on some data (from who knows where) they ran across a situation where the pipeline ran for more than 24 hours before someone finally just killed it. Turns out it was trying to process one or more full-frame AMI exposures, which I've confirmed take a really long time to process through the `ami_analyze` step. All of the test data we've used during development (provided by the AMI WG) has been much smaller 19x19 or 80x80 subarrays. I've confirmed that when I take an 80x80 subarray out of the middle of the original full-frame exposure, it then only takes ~7 minutes to process through `ami_analyze`.

I've queried the AMI WG as to whether we should always extract an 80x80 subarray from full-frame data when sending it through `ami_analyze`. It's doubtful that there would ever be useful source information outside of that region and hence shouldn't affect the processing. Waiting for the AMI WG to respond.

An example test exposure that causes the problem is jw00308001001_02102_00001_nis_cal.fits

hbushouse:
The AMI WG agrees that it's OK and appropriate to extract an 80x80 subarray from full-frame exposures (equivalent to the user using the SUB80 subarray in the first place) before sending the data into the ami_analyze fringe computation algorithms.

Extraction of the subarray could probably be implemented in the ami_analyze.apply_LG function somewhere before the image is sent to the jwnrm.fit_image() function.

hbushouse:
The AMI observing template in APT allows a user to select a SUBARRAY of "FULL" or "SUB80". So our extraction from full-frame exposures should match the location and size of the SUB80 subarray. The latest info we have for the SUB80 characteristics is:
```
SUBARRAY= 'SUB80   '           / Subarray used                                  
SUBSTRT1=                 1045 / Starting pixel in axis 1 direction             
SUBSTRT2=                    1 / Starting pixel in axis 2 direction             
SUBSIZE1=                   80 / Number of pixels in axis 1 direction           
SUBSIZE2=                   80 / Number of pixels in axis 2 direction  
```
which is based on the most recent dark and superbias ref files in CRDS that use the SUB80 subarray.

hbushouse:
Fixed by #2017

PR #2012, opened 2018-05-09
Update ami_analyze to always extract an 80x80 subarray from full-fram…

dmggh:
…e data. This is equivalent to the user using the SUB80 subarray in the first place, so the subarray matches the location and size of the SUB80 subarray. Updated docs. This addresses github issue #1837.


docs/jwst/ami/main.rst
```diff
@@ -71,7 +71,12 @@ The ``ami_analyze`` step applies the Lacour-Greenbaum (LG) image plane
 modeling algorithm to a NIRISS AMI image.
 The routine computes a number of parameters, including a model fit (and
 residuals) to the image, fringe amplitudes and phases, and closure phases
-and amplitudes.
+and amplitudes.  For a full-frame exposure, instead of using the FULL subarray,
```
hbushouse: I would say "instead of using the FULL **image** ..."

docs/jwst/ami/main.rst
```diff
@@ -71,7 +71,12 @@ The ``ami_analyze`` step applies the Lacour-Greenbaum (LG) image plane
...
+a subarray will be extracted from its center to keep the execution time acceptable.
```
hbushouse: Is the subarray actually at (or near) the center of the full image? If not, I would remove the words "from its center".

jwst/ami/ami_average.py
```diff
@@ -1,8 +1,7 @@
-from __future__ import division
-
 #
 #  Module for averaging LG results for a set of AMI exposures
 #
+from __future__ import division
```
hbushouse: Now that we're only support py3.x environments, we can remove all the `from __future__ import ...` statements, right?

jwst/ami/ami_normalize.py
```diff
@@ -1,12 +1,10 @@
-from __future__ import division
-
 #
 #  Module for normalizing the LG results for a science target by
 #  the LG results for a reference target
 #
+from __future__ import division
```
hbushouse: Remove?

dmggh:
Unfathomable messages from github when I try to resolve this conflict, so will close this PR and try again,

PR #2019, opened 2018-05-10
Fix some model_blender issues in cube_build

hbushouse:
Several fixes related to the recent implementation of model_blender in cube_build:
 - Updated the cube_build step itself to only call model_blender when there are multiple inputs. It still leaves in place the error associated with input file names not being in the working directory (@hack is working on a parallel fix for that).
 - Changed the comment string on the `CHANNEL` keyword back to its original value, just to avoid a whole lot of regression tests showing differences. It needs to be updated some time in the future, but not right now.
 - Added `MULTIPLE` as an allowed value to the keyword-specific schemas for `FILTER`, `BAND`, and `GRATING`, as was done recently for those keywords in the core schema. So now they're in sync again.

### DIFF
--- a/jwst/cube_build/ifu_cube.py
+++ b/jwst/cube_build/ifu_cube.py
@@ -684,7 +684,12 @@ class IFUCubeData(object):
         IFUCube.update(self.input_models[j])
         IFUCube.meta.filename = self.output_name
         
-        self.blend_output_metadata(IFUCube)
+        # Call model_blender if there are multiple inputs
+        if len(self.input_models) > 1:
+            saved_model_type = IFUCube.meta.model_type
+            self.blend_output_metadata(IFUCube)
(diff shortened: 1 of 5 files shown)


hbushouse:
Latest commit completely comments out the call to blend_output_models in cube_build for now, because there are still other lower-level problems with blending the meta data from the `mrs_imatch` step (the array-based polynomial_info items) that can't be fixed in time for our initial B7.1.3 delivery tomorrow. All this means is that the output _s3d.fits products from cube_build won't have blended metadata in a HDRTAB table extension, which isn't a big deal (at least for this delivery).

stsci-hack:
Subsequent testing using the nightly regression test test_mrs_spec3 confirms that this change will still allow the pipelines for Cube data to run successfully.

PR #2027, opened 2018-05-14
Remove default extensions from model_base, add them to setup file

bernie-simon:
Asdf has changed the way it handles asdf extensions. It now loads them using the entry_points attribute to the setup function. I added the BaseExtension class to the setup entry_points and removed the default extension list handling in model_base, as it is no longer needed.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -32,9 +32,6 @@ from jwst.transforms.jwextension import JWSTExtension
 from gwcs.extension import GWCSExtension
 
 
-jwst_extensions = [GWCSExtension(), JWSTExtension(), BaseExtension()]
-
-
 class DataModel(properties.ObjectNode, ndmodel.NDModel):
     """
     Base class of all of the data models.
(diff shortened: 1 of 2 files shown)


drdavella:
I think it's still worth asking whether the `BaseExtension` is needed at all.

jwst/datamodels/model_base.py
```diff
@@ -80,10 +77,6 @@ def __init__(self, init=None, schema=None, extensions=None,
             an excption. If false, they will generate a warning.
         """
         # Set the extensions
-        if extensions is None:
-            extensions = jwst_extensions[:]
-        else:
-            extensions.extend(jwst_extensions)
         self._extensions = extensions
```
drdavella: Based on a quick review of the places where this class is used, it seems like it's no longer necessary to carry along this `extensions` argument at all. That would mean it could be removed from all of the calls to `AsdfFile` below as well. Are there any cases where a different set of extensions other than those provided by the entry points will be needed?

Issue #1169, opened 2017-09-06
Define write() method for ModelContainer


jdavies-st:
Right now `write` is defined as an alias for `save` in datamodels.  `ModelContainer` subclasses from these, redefines `save`, but does not redefine `write` as an alias for its own version of `save`. It ought to.

bernie-simon:
This issue has been resolved by pull request #2032

PR #2032, opened 2018-05-15
Redefine write method to allow override

bernie-simon:
Datamodels defines the read and write methods for astropy registry compatibility. This update changes the definition of write from an assignment to a method call so that the save method can be overridden by subclasses of datamodels, such as ModelContainer.


bernie-simon:
This change resolves issue #1169. ModelContainer does not need a write method, as the code now calls the object's save method when write is invoked.

PR #2037, opened 2018-05-16
Change NRC_GRISM to NRC_WFSS

sosey:
This applied the request changes for #1985, including update of the documentation.


sosey:
GAH! Since there's no real data is that being done for CRDS reasons?

hbushouse:
The CRDS updates are actively being worked in https://jira.stsci.edu/browse/CRDS-214. They can't deliver the updated rmaps until the value "NRC_WFSS" is acknowledged as an allowed value in our datamodel schemas. So let's go ahead and merge these updates as-is, without the changes for backwards compatibility (since we really don't need it for CRDS).

Are their any other updates/commits to be done before we merge? (everything looked good to me)

sosey:
These are all the updates I found...feel free to hit the merge when you're ready

PR #2038, opened 2018-05-17
Update Jenkins CRDS_CONTEXT to 460

hbushouse:
Updating regression tests to latest CRDS context.


jaytmiller:
Is jwcalibdev history now?

stscicrawford:
Two questions:

1) To simplify the process, could you just make this change by editing the file on the github interface? (Yes I know this is bad for a number of reasons, but this could be temporary solution until a better solution is found)

2) Is there a reason this is being set and it isn't just using the current context?

hbushouse:
The reason it's hardwired for the reg tests is because we usually want to do testing of the changes included in a new context before that context is made the default, i.e. save regular users (who use the default) from suffering the pain of errors. Once the reg tests have shown that everything with a new context is OK, it then becomes the default.

Issue #2034, opened 2018-05-16
NIRSpec fixed slit data fails in extract_1d when slit.meta.exposure.type is checked


jdavies-st:
PR #2033 seems to have introduced a bug, which was caught in the nightly regression tests.

https://boyle.stsci.edu:8081/job/RT/job/JWST/31/testReport/jwst.tests_nightly.general.nirspec/test_extract_1d/test_extract1d_nirspec/

The line that [checks `slit.meta.exposure.type`](https://github.com/STScI-JWST/jwst/blob/43157dfb27e22c8533f4743ee19bb4077eab5a05/jwst/extract_1d/extract.py#L310) fails for NIRSpec slits that only have a GWCS object in the slit `meta`.  The datamodel itself contains `meta.exposure.type`, but the slits do not.

jdavies-st:
Hmm, interesting.  I wonder if that's a problem with the old input data on our nightly regression tests?  That said, I don't see `meta.exposure.type` in the schema for a slit in a MultiSlitModel, which relies on `slitdata.schema.yaml` to define what's in each slit.
```
In [12]: a = datamodels.open('nirspec/test_extract_1d/jw00023001001_01101_00001_NRS1_cal.fits')

In [13]: a
Out[13]: <MultiSlitModel from jw00023001001_01101_00001_NRS1_cal.fits>

In [14]: a.meta.exposure.type
Out[14]: 'NRS_FIXEDSLIT'

In [15]: s0 = a.slits[0]

In [16]: s0.meta.exposure.type
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
~/dev/jwst/jwst/datamodels/properties.py in __getattr__(self, attr)
    203         try:
--> 204             val = self._instance[attr]
    205         except KeyError:

KeyError: 'exposure'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
<ipython-input-16-f6d1c6b2551f> in <module>()
----> 1 s0.meta.exposure.type

~/dev/jwst/jwst/datamodels/properties.py in __getattr__(self, attr)
    205         except KeyError:
    206             if schema == {}:
--> 207                 raise AttributeError("No attribute '{0}'".format(attr))
    208             val = _make_default(attr, schema, self._ctx)
    209             self._instance[attr] = val

AttributeError: No attribute 'exposure'
```

Issue #2031, opened 2018-05-15
In extract_1d, regard wavelength = 0 as invalid


philhodge:
The `find_dispaxis` function uses the wavelengths to determine which is the dispersion axis.  The wavelengths may be taken from the wavelength attribute (extension in the input file), or they may be computed by using the wcs function.  Wavelengths are not necessarily defined over the entire data range, and `find_dispaxis` currently makes the assumption that if the wavelength for a pixel is not defined (i.e. if one or more reference files for the wcs do not cover that pixel), the value of the wavelength will be NaN.  That assumption is not always correct; sometimes the wavelength is 0.  `find_dispaxis` should be modified to treat both 0 and NaN as invalid.

PR #2041, opened 2018-05-17
In find_dispaxis, regard wavelength = 0 as invalid

philhodge:
The `input_model` was added to the calling sequence of `find_dispaxis`, and some information (e.g. exposure type) is gotten from `input_model` rather than from `slit`.  In the section for computing wavelengths from the wcs function, grism data are now supported.

See issue #2031.


jdavies-st:
Resolves #2034.

PR #2045, opened 2018-05-17
Fix NIRISS WFSS spec2 regression test

jdavies-st:
- Removed the `runslow` marker from the test, as @hbushouse has pruned the input file so that the test runs in a few minutes.  This now enables this test to run on Jenkins.
- Extended the work of #1982 which didn't catch all (or even most) of the instances of `meta.asn.table_name` being populated with a full path in our pipelines.  This test uses the spec2 pipeline, which had this bug.

### DIFF
--- a/jwst/pipeline/calwebb_image3.py
+++ b/jwst/pipeline/calwebb_image3.py
@@ -92,7 +92,7 @@ class Image3Pipeline(Pipeline):
 
         try:
             result.meta.asn.pool_name = input_models.meta.asn_table.asn_pool
-            result.meta.asn.table_name = input
+            result.meta.asn.table_name = os.path.basename(input)
             result.meta.filename = input_models.meta.asn_table.products[0].name
         except:
             pass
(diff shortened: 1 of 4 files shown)


jdavies-st:
In addition to fixing the ASNTABLE bug in this PR, I am also updating the regression test comparison data for the following tests, where the only difference now is the ASNTABLE keyword:
- pipelines.test_nrc_image3_1.test_image3_pipeline1
- pipelines.test_nrc_tso3_2.test_tso3_pipeline3
```
cp -u test_image3_pipeline1-current/mosaic_long_i2d.fits /data4/jwst_test_data/pipelines/nircam_calimage3/mosaic_long_i2d_ref.fits
cp -u test_tso3_pipeline3-current/jw87600-a3001_t1_niriss_clear-gr700xd_x1dints.fits /data4/jwst_test_data/pipelines/niriss_caltso3/jw87600-a3001_t1_niriss_clear-gr700xd_x1dints_ref.fits
```

PR #2046, opened 2018-05-18
Updated docs to reflect code updates for PR #1925 'Update ramp_fit sl…

dmggh:
…ope variances'.

### DIFF
--- a/docs/jwst/ramp_fitting/main.rst
+++ b/docs/jwst/ramp_fitting/main.rst
@@ -16,34 +16,36 @@ detailed below.
 
 
 The count rate for each pixel is determined by a linear fit to the
-cosmic-ray-free ramp intervals for each pixel. CR-free intervals are derived
-using the 4-D GROUPDQ array of the input data set, under the assumption that
-the jump step will have already flagged CR's. Ramp intervals are also terminated
-where saturation flags are found.  Ramp intervals that are noiseless, or have
-no signal, or contain only 2 reads will initially have fits with variance = 0,
-preventing their slopes from contributing to the weighted slopes.  In these
-cases, the variance will be recalculated as the poisson noise of the ramp added
-in quadrature to the pixel-specific read noise, ensuring that all variance
-values are positive.  If the input dataset has only a single group in each
-integration, the count rate for all unsaturated pixels in that integration will
-be calculated to be the value of the science data in that group divided by the
-exposure time.  If the input dataset has only two groups per integration, the
-count rate for all unsaturated pixels in each integration will be calculated
-from the 2 valid values of the science data.  If any input dataset contains
-ramps saturated in their second read, the count rates for those pixels in that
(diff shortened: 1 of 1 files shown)


docs/jwst/ramp_fitting/main.rst
```diff
@@ -16,34 +16,36 @@ detailed below.
...
+GROUPDQ array of the input data set, under the assumption that the jump step
+will have already flagged CR's. Ramp intervals are also terminated where
+saturation flags are found. Pixels are processed simultaneously in blocks 
+using the array-based functionality of numpy.  The size of the block depends
+on the image size and the number of groups.
+
+
+If the input dataset has only a single group in each integration, the count rate
+for all unsaturated pixels in that integration will be calculated to be the
+value of the science data in that group divided by the exposure time.  If the
```
philhodge: The slope should be the value in that group divided by the integration time (or the group time), not the exposure time.

docs/jwst/ramp_fitting/main.rst
```diff
@@ -16,34 +16,36 @@ detailed below.
...
+input dataset has only two groups per integration, the count rate for all
+unsaturated pixels in each integration will be calculated from the 2 valid
+values of the science data.  If any input dataset contains ramps saturated in
+their second read, the count rates for those pixels in that integration will be
+calculated to be the value of the science data in that group divided by the
```
philhodge: "In that group" is misleading, since you just referred to the second group, the saturated one.  The slope is the value in the first group divided by the group time (but see below).  Anyone reading this would know exactly what you meant, however.

On the other hand, the slope is definitely not the value in the first group divided by the exposure time (or even by the integration time).  The integration time could include many groups.  I think you should divide by the group time.  I'm uncertain as to whether the number of dropped frames (DRPFRMS1) at the beginning of an integration should be included, but as far as I can tell that value is zero for every mode except target acq.

docs/jwst/ramp_fitting/main.rst
```diff
@@ -84,29 +86,110 @@ read with a non-zero magnitude. The order of the cosmic rays within the ramp is
...
+Segment-specific computations:
+------------------------------
+
+The slope of each segment is calculated using the least-squares method with 
+optimal weighting. The variance of the slope of the segment due to read noise is: 
+
+.. math::  
+   var^R_{s} = \frac{12 \ R^2 }{ (ngroups_{s}^3 - ngroups_{s})(tgroup^2) } \,,
+
+\noindent where $R$ is the noise in the difference between 2 frames,
```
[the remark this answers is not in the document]
philhodge: Shouldn't R be the value read from the readnoise reference file divided by the square root of 2?

dmggh:
I notice 2 problems with the equations after the build: 2 equations were not rendered (which I think can be fixed by adding a blank line before each), and variables intended to be italicized in the text are not.

PR #2050, opened 2018-05-18
Update regression context to jwst_0461.pmap

jaytmiller:
For NIRSPEC SFLAT NRS_AUTOWAVE handling.


jaytmiller:
I also updated pdk_environment.

jaytmiller:
OK,  just a minute,  Travis is still checking...

jaytmiller:
Once Travis is done,  Jenkins and Pandokia can do a little follow up and we can breathe easier.

Issue #1915, opened 2018-04-19
support for compressed files in datamodels


nden:
It may be useful to support compressed files in datamodels although it is a somewhat tricky issue.
FITS has support for compression and ASDF currently supports compression of the binary arrays.
What would be most useful initially is to be able to work with compressed FITS files.

bernie-simon:
Update #2051 adds support for compressed image extensions to datamodels.

PR #2051, opened 2018-05-18
Add support for compressed image extensions to datamodels

bernie-simon:
This fix allows programs to read compressed image extensions from fits files and update them. The main change has been to how the code determines the type of an extension. It no longer assumes that anything that's not a binary table is a regular image extension. The code that reads and writes compressed images is layered on top of the astropy fits library code.

This fix addresses issue #1915

### DIFF
--- a/jwst/datamodels/fits_support.py
+++ b/jwst/datamodels/fits_support.py
@@ -99,27 +99,18 @@ def _get_hdu_name(schema):
     return hdu_name
 
 
-def _get_hdu_type(hdu_name, value):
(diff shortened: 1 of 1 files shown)


jdavies-st:
Might be a good idea to add a simple unit test (or tests) to generate a datamodel, write it out in compressed format, and then read it in, verifying you get the same thing back.

bernie-simon:
You can't write a compressed extension from scratch, the interface doesn't support it. You can read existing compressed files and update them.

Issue #2049, opened 2018-05-18
MIRI RSCD exception for no match in reference table


hbushouse:
DMS test exposure jw80600013001_02101_00001_mirimage_uncal.fits is failing in the `rscd` step, because it can't find a matching entry in the RSCD reference table. The science exposure uses SUBARRAY='FULL' and READPATT='FASTGRPAVG', but the only READPATT entries in  jwst_miri_rscd_0009.fits for SUBARRAY='FULL' are 'FAST' and 'SLOW'. The RSCD step then raises an exception for this.

If possible and appropriate, the RSCD reference file should be updated to include an entry for READPATT='FASTGRPAVG' (will leave it to the MIRI team to decide if correction params can be implemented for this mode). If it's possible to update the RSCD reference file, a separate CRDS ticket should be filed for that.

For the `rscd` step in the Cal pipeline, we should at least change the way this situation is handled. It's never a good idea to raise an exception, because that means the entire process shuts down without the ability of the Ops error logging/handling to close things off in a neat and tidy way. The handling of this in the `rscd` step should be to issue a log warning (or error), and then skip the remaining processing of the step, and return a copy of the input model to the caller, as is done in other steps that have to abort mid-stream.

hbushouse:
Fixed by #2052

PR #2052, opened 2018-05-18
fixed error on FASTGRPAVE data

jemorrison:
Fixed problem in #2049

### DIFF
--- a/jwst/rscd/rscd_sub.py
+++ b/jwst/rscd/rscd_sub.py
@@ -217,8 +217,8 @@ def get_rscd_parameters(input_model, rscd_model):
         log.warning('No matching row found in RSCD reference table for')
         log.warning('READPATT=%s, SUBARRAY=%s, Row type=EVEN',
                     readpatt, subarray)
-        raise NoTableEntry('No match for READPATT or SUBARRAY in RSCD ref table')
-        return None, None
+        log.warning('RSCD correction will be skipped, only 1 Group need at least 2')
+        return None
 
     # Load the params from the matching table row
     param = {}
@@ -249,8 +249,8 @@ def get_rscd_parameters(input_model, rscd_model):
         log.warning('No matching row found in RSCD reference table for')
         log.warning('READPATT=%s, SUBARRAY=%s, Row type=ODD',
                     readpatt, subarray)
-        raise NoTableEntry('No match for READPATT or SUBARRAY in RSCD ref table')
-        return None, None
+
+        return None
 
     # Load the params from the matching table row
     param['odd']['tau'] = tau_table[index2]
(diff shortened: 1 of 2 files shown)


jwst/rscd/rscd_sub.py
```diff
@@ -217,8 +217,8 @@ def get_rscd_parameters(input_model, rscd_model):
         log.warning('No matching row found in RSCD reference table for')
         log.warning('READPATT=%s, SUBARRAY=%s, Row type=EVEN',
                     readpatt, subarray)
-        raise NoTableEntry('No match for READPATT or SUBARRAY in RSCD ref table')
-        return None, None
+        log.warning('RSCD correction will be skipped, only 1 Group need at least 2')
```
hbushouse: The last phrase here "only 1 Group need at least 2" is not relevant to this situation. The appropriate reason was already stated in the warning message in line 217. So this message can be trimmed to just "RSCD step will be skipped."

jwst/rscd/rscd_sub.py
```diff
@@ -249,8 +249,8 @@ def get_rscd_parameters(input_model, rscd_model):
         log.warning('No matching row found in RSCD reference table for')
         log.warning('READPATT=%s, SUBARRAY=%s, Row type=ODD',
                     readpatt, subarray)
-        raise NoTableEntry('No match for READPATT or SUBARRAY in RSCD ref table')
-        return None, None
+
```
hbushouse: Should add the same warning here as in line 220, "RSCD step will be skipped", before returning.

jwst/rscd/rscd_sub.py
jemorrison: it is not rscd_sub that returns NONE but a function in rscd_sub (get_rscd_parameters) that returns NON to rscd_sub. If none is returned from get_rscd_parameters to rscd_sub
then a warning message is sent that the table value could not be found, output.meta.cal_step.rscd is set to skipped and output is returned (output is set to input_model.copy() in the beginning of rscd_sub.

jemorrison: I took out that extra (incorrect warning message) in the most recent pr

hbushouse:
Testing using a version of the code that removes line 220 I get the following result for the dataset in question:
```
2018-05-22 10:30:15,816 - stpipe.Detector1Pipeline.rscd - INFO - Step rscd running with args (<RampModel(1, 20, 1024, 1032) from jw80600013001_02101_00001_mirimage_uncal.fits>,).
2018-05-22 10:30:15,852 - stpipe.Detector1Pipeline.rscd - INFO - Using RSCD reference file /grp/crds/cache/references/jwst/jwst_miri_rscd_0009.fits
2018-05-22 10:30:16,053 - stpipe.Detector1Pipeline.rscd - WARNING - No matching row found in RSCD reference table for
2018-05-22 10:30:16,053 - stpipe.Detector1Pipeline.rscd - WARNING - READPATT=FASTGRPAVG, SUBARRAY=FULL, Row type=EVEN
2018-05-22 10:30:16,053 - stpipe.Detector1Pipeline.rscd - WARNING - RSCD correction will be skipped
2018-05-22 10:30:16,057 - stpipe.Detector1Pipeline.rscd - INFO - Step rscd done
```
so that looks good.

PR #2053, opened 2018-05-20
Reorder multislit extensions for NIRISS WFSS spec2 test

jdavies-st:
Reverting changes I'd made to this test, as it seems the spec2 pipeline doesn't preserve slit order.  This fix sets the slit order for the test.


jdavies-st:
Merging this so that it gets into tonight's regression tests.

Issue #940, opened 2017-06-01
running assign_wcs with inputs not  in current directory


nden:
assign_wcs does not find the msa configuration file when the input data is not in the current directory.
I think we can assume the msa configuration file is always in the directory with the data. assign_wcs needs to be modified to use the full path when reading the configuration  file

PR #2073, opened 2018-05-22
Handle MSA metadata file only in assign_wcs step using Step.input_dir

jdavies-st:
* Utilize Step.input_dir to find the MSA metadata file
* Move all reading of the MSA metadata file to assign_wcs step
* Store MSA info needed in flat_field step in each slit instead of rereading MSA file
* Update the MultiSlitModel schema to store `quadrant, xcen, ycen` in each slit

### DIFF
--- a/jwst/assign_wcs/assign_wcs_step.py
+++ b/jwst/assign_wcs/assign_wcs_step.py
@@ -1,4 +1,5 @@
 #! /usr/bin/env python
+from os.path import join, basename
 from ..stpipe import Step, cmdline
 from .. import datamodels
 import logging
@@ -59,7 +60,12 @@ class AssignWcsStep(Step):
                 reffile = self.get_reference_file(input_model, reftype)
                 reference_file_names[reftype] = reffile if reffile else ""
 
+            # Get the MSA metadata file if needed and add to reffiles
+            msa_metadata_file = input_model.meta.instrument.msa_metadata_file
+            if msa_metadata_file:
+                msa_metadata_file = join(self.input_dir, basename(msa_metadata_file))
+                reference_file_names['msametafile'] = msa_metadata_file
+
             result = load_wcs(input_model, reference_file_names)
 
         return result
-
--- a/jwst/assign_wcs/assign_wcs_step.py
+++ b/jwst/assign_wcs/assign_wcs_step.py
@@ -1,5 +1,4 @@
 #! /usr/bin/env python
-from os.path import join, basename
 from ..stpipe import Step, cmdline
 from .. import datamodels
 import logging
@@ -62,8 +61,8 @@ class AssignWcsStep(Step):
 
             # Get the MSA metadata file if needed and add to reffiles
             msa_metadata_file = input_model.meta.instrument.msa_metadata_file
-            if msa_metadata_file:
-                msa_metadata_file = join(self.input_dir, basename(msa_metadata_file))
+            if msa_metadata_file is not None:
+                msa_metadata_file = self.make_input_path(msa_metadata_file)
                 reference_file_names['msametafile'] = msa_metadata_file
 
             result = load_wcs(input_model, reference_file_names)
--- a/jwst/assign_wcs/nirspec.py
+++ b/jwst/assign_wcs/nirspec.py
@@ -292,7 +292,7 @@ def slitlets_wcs(input_model, reference_files, open_slits_id):
 def get_open_slits(input_model, reference_files=None):
     exp_type = input_model.meta.exposure.type.lower()
     if exp_type == "nrs_msaspec":
-        msa_metadata_file, msa_metadata_id = get_msa_metadata(input_model)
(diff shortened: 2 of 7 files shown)


jdavies-st:
Since `quadrant`, `xcen` and `ycen` for each slit are used downstream by flat_field, it makes sense to put them in each slit of the `MultiSlitModel` produced by extract_2d.  @hbushouse, I've added them to schema, but have not added FITS keywords for these, as I did not think they were needed.  Is that OK?

Is there a use case for having them in the FITS header?

nden:
LGTM. Closes #940

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -59,7 +60,12 @@ def process(self, input):
                 reffile = self.get_reference_file(input_model, reftype)
                 reference_file_names[reftype] = reffile if reffile else ""
 
+            # Get the MSA metadata file if needed and add to reffiles
+            msa_metadata_file = input_model.meta.instrument.msa_metadata_file
+            if msa_metadata_file:
+                msa_metadata_file = join(self.input_dir, basename(msa_metadata_file))
```
stscieisenhamer: At points like these, the intent is to use `Step.make_input_path`, to encapsulate the appending and any other `Step` dependent things may come into play when creating paths.

jdavies-st: Ah, ok!  That can be fixed.

jwst/flatfield/flat_field.py
```diff
@@ -278,29 +278,12 @@ def do_NIRSpec_flat_field(output_model,
...
     any_updated = False
-    if exposure_type == "NRS_MSASPEC":
-        # This is not the same as output_model.slits; `slits` will be
-        # used a few lines farther down.
-        slits = nirspec.get_open_slits(output_model)
+
     for (k, slit) in enumerate(output_model.slits):
         log.info("Processing slit %s", slit.name)
-        slit_nt = None
+        slit_nt = slit
```
philhodge: The changes to flat_field.py look fine.

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -59,7 +60,12 @@ def process(self, input):
                 reffile = self.get_reference_file(input_model, reftype)
                 reference_file_names[reftype] = reffile if reffile else ""
 
+            # Get the MSA metadata file if needed and add to reffiles
+            msa_metadata_file = input_model.meta.instrument.msa_metadata_file
+            if msa_metadata_file:
```
nden: Should it be checking for `None` here?

jwst/assign_wcs/nirspec.py
```diff
@@ -338,16 +338,17 @@ def get_open_fixed_slits(input_model):
...
     """
     Get the MSA metadata file (MSAMTFL) and the msa metadata id (MSAMETID).
 
     """
-    msa_config = input_model.meta.instrument.msa_metadata_file
-    if msa_config is None:
-        message = "MSA metadata file is not available (keyword MSAMETFL)."
+    try:
+        msa_config = reference_files['msametafile']
+    except KeyError:
```
nden: Thinking about this some more, here the fall back should be getting the file name from the model and assuming it is run from the current directory (essentially duplicating the current behaviour). If this fails then raise the error. The reason for this is that there are (limited) cases when this function is called directly from code outside assign_wcs and the call may bypass assign_wcs_step. 

Another option would be to somehow pass `input_dir` downstream.

jdavies-st: Makes sense.  I will update the PR.

I avoided passing `Step.input_dir` downstream, as that would have changed several function calls.

jwst/datamodels/schemas/slitdata.schema.yaml
```diff
@@ -80,6 +80,18 @@ allOf:
       default: ""
       fits_keyword: SHUTSTA
       fits_hdu: SCI
+    quadrant:
+      title: MSA quadrant for slit
+      type: integer
+      fits_hdu: SCI
```
hbushouse: Because these don't have a `fits_keyword` attribute, do they need to have `fits_hdu` defined?

jwst/assign_wcs/assign_wcs_step.py
jdavies-st: You mean it should be
```if msa_metadata_file is not None:```
?

nden: May be. It's in the core schema and if it's not set (e.g. fixed slits) you get `None`, right?
Better try that before changing it.

jdavies-st: Since it is defined in the schema, if it is not populated, the line above just returns None.  `is not None` is probably the better syntax here.  I've changed it.

Issue #2076, opened 2018-05-23
RSCD step not making a correction  to science data


jemorrison:
Misty Cracraft reported that the rscd step was not correcting the science data for integrations > 1.
J Morrison determined it was because the step was incorrectly flagging the science pixels are reference pixels and skipping the correction for those pixels. PR coming with a fix
@hbushouse

hbushouse:
Fixed by #2077.

PR #2077, opened 2018-05-23
Fix of rscd step, science pixels flagged as ref pixels so step was sk…

jemorrison:
The science pixels were incorrectly flagged as reference pixels. This step does not correct reference pixels so the science pixels were not corrected for the rscd effect.
This PR closes issue #2076


hbushouse:
Merging to get into tonight's regression tests.

PR #2082, opened 2018-05-23
Remove absolute file paths from ASNTABLE keyword and assocation tables

jdavies-st:
- strip abspath from ASNTABLE keyword in calwebb_ami3, calwebb_coron3
- use Step.load_as_level3_asn in calwebb_coron3
- use Step.load_as_level2_asn in wfs_combine_step

### DIFF
--- a/jwst/wfs_combine/wfs_combine_step.py
+++ b/jwst/wfs_combine/wfs_combine_step.py
@@ -18,7 +18,7 @@ class WfsCombineStep(Step):
 
     def process(self, input_table):
 
-        asn_table = json.load(open(input_table, 'r'))
+        asn_table = self.load_as_level2_asn(input_table)
         num_sets = len(asn_table['products'])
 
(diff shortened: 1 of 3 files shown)


jwst/wfs_combine/wfs_combine_step.py
```diff
@@ -18,7 +18,7 @@ class WfsCombineStep(Step):
 
     def process(self, input_table):
 
-        asn_table = json.load(open(input_table, 'r'))
+        asn_table = self.load_as_level2_asn(input_table)
```
hbushouse: The wfs_combine_step is a level-3 step that combines multiple (usually a pair of) dithered WFS&C images, so the ASN file should be composed as a level3 ASN, not level2.

jdavies-st: Thanks for catching that @hbushouse.  The regression test did pass, but perhaps it is a special case.  Will change this in new PR.

PR #2091, opened 2018-05-24
Switched CRDS_CONTEXT in JenkinsfileRT to jwst-edit

jaytmiller:
The jwst-edit context tracks the last ReDCaT delivery,  currently jwst_0462.pmap.   

Also updated regression context in 

jwcalibdev:/data1/jwst_rt/pdk_environment 

to jwst-edit.

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -7,7 +7,7 @@ def test_env = [
     "PATH=./miniconda/bin:/bin:/usr/bin",
     "TEST_BIGDATA=/data4/jwst_test_data",
     "CRDS_SERVER_URL=https://jwst-crds.stsci.edu",
-    "CRDS_CONTEXT=jwst_0461.pmap",
+    "CRDS_CONTEXT=jwst-edit",
 ]
 
 //def python_version = ['3.5', '3.6']


jaytmiller:
Might be best to let Pandokia take the lead for the jwst-edit switch the first night.   If that works,  merging this should be OK as well.

hbushouse:
It looks like the edit to the Pandokia pdk_environment file didn't get implemented in time to make it effective for last night's reg test run, so let's see what happens during this morning's run. If that checks out OK, we'll then go ahead and merge this change to the Jenkins side.

jaytmiller:
That's something we didn't consider and "blind" context switches in testing does seem sub-par.

James and I talked and some possible "solutions" came up (and I'm adding now):

1.  Wait and see how bad a problem fuzz on the context is.   If it's a real problem,  just revert to manual updates.   Todd can do pull requests and/or merging.

2. Set up a "CRDS bot" which automatically does a pull request to define a new explicit Jenkins context whenever ReDCaT deliveries occur.  (Poll say every 10 minutes). 

3. Add some kind of Jenkins test logging based on CRDS tools to make the active test context and transitions clearer. 
  
4. Enhance CRDS tools and/or displays to give easier representations of context history from the perspective of active and past test contexts.

Considerations for the chosen solution:

1.  Using jwst-edit removes Todd from the update loop and reduces the chances of introducing unnecessary end-to-end delays due to not tracking ReDCaT closely enough.

2. There are 100's of contexts generated so modest efforts can probably be justified if this is indeed an issue and no simpler solution can be found.

PR #2099, opened 2018-05-29
update Readme with new release to DMS [skip ci]

nden:


stscicrawford:
Thanks @nden 

Can you add 0.9.3 -- which was 'First release candidate for Build 7.1.3'?

I think 0.9.4 would be better described as 'Second release candidate for Build 7.1.3'

Also, can you move this whole section above the Unit Testing section?

stscicrawford:
LGTM!  Thanks @nden

PR #2100, opened 2018-05-29
Reference file  specifications part I

stscicrawford:
This is the first pull request to move the reference file specifications back into the JWST pipeline documentation from http://jwst-reffiles.stsci.edu/index.html.   These updates are minor and take the following forms:
* format changes 
* adding additional text that appeared in the jwst-reffiles especially referring to files using gain
* fixing the latex in the RCSD reference file

@hbushouse @rizeladiaz @jemorrison @dmggh

### DIFF
--- a/docs/jwst/ramp_fitting/reference_files.rst
+++ b/docs/jwst/ramp_fitting/reference_files.rst
@@ -31,8 +31,8 @@ with ``EXTNAME=SCI``, which contains a 2-D floating-point array of gain values
 READNOISE Reference Files
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 The read noise reference file is a FITS file with a single IMAGE extension,
-with ``EXTNAME=SCI``, which contains a 2-D floating-point array of read noise values
-per pixel. The units of the read noise should be electrons and should be the
+with ``EXTNAME=SCI``, which contains a 2-D floating-point array of read noise
+values per pixel. The units of the read noise should be DN and should be the
 CDS (Correlated Double Sampling) read noise, i.e. the effective noise between
 any pair of non-destructive detector reads. The REFTYPE value is ``READNOISE``.
 
--- a/docs/jwst/ramp_fitting/reference_files.rst
+++ b/docs/jwst/ramp_fitting/reference_files.rst
@@ -6,29 +6,33 @@ electrons, and convert the results of ramp fitting back to DN.
 The read noise values are used as part of the noise estimate for
 each pixel. Both are necessary for proper computation of noise estimates.
 
-
-CRDS Selection Criteria
------------------------
-The readnoise reference file is selected by instrument, detector and, where
-necessary, subarray.
-
-The gain reference file is selected based on instrument, detector and,
+GAIN Reference Files
+^^^^^^^^^^^^^^^^^^^^
+The GAIN reference file is selected based on instrument, detector and,
 where necessary, subarray.
 
+READNOISE Reference Files
+^^^^^^^^^^^^^^^^^^^^^^^^^
+The READNOISE reference file is selected by instrument, detector and, where
+necessary, subarray.
 
-Gain Reference File Format
---------------------------
+
+Reference File Formats
+----------------------
+
+GAIN Reference Files
+^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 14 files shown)


docs/jwst/ramp_fitting/reference_files.rst
```diff
@@ -6,29 +6,33 @@ electrons, and convert the results of ramp fitting back to DN.
...
---------------------------------
+READNOISE Reference Files
+^^^^^^^^^^^^^^^^^^^^^^^^^
 The read noise reference file is a FITS file with a single IMAGE extension,
-with ``EXTNAME=SCI``, which contains a 2-D floating-point array of read noise 
-values per pixel. The units of the read noise should be DN and should be the 
+with ``EXTNAME=SCI``, which contains a 2-D floating-point array of read noise values
+per pixel. The units of the read noise should be electrons and should be the
 CDS (Correlated Double Sampling) read noise, i.e. the effective noise between
 any pair of non-destructive detector reads. The REFTYPE value is ``READNOISE``.
```
dmggh: Currently, READNOISE reference files are in units of DN; which is what the ramp fitting code expects - the code temporarily converts to electrons for noise calculations. Some older reference files are in electrons, and should have been overridden with correct units.

rizeladiaz: The information about the Gain reference files format should be kept.

stscicrawford:
Since I've already updated the files to match @dmggh suggestion, I'm moving the discussion here to the main comments just to clarify with @rizeladiaz comment.   

The pipeline is currently expecting the files to be in DN.   Should they instead be in electrons?  If so, we should create a new issue so everything can be updated to a consistent units.

stscicrawford:
Turns out everything is okay and this can be merged

hbushouse:
I think you mean **READNOISE** ref files should be in units of DN, not GAIN files. GAIN files are e/DN.

docs/jwst/refpix/reference_files.rst
```diff
@@ -1,6 +1,22 @@
...
+
+CRDS Selection Criteria
+-----------------------
+Refpix reference files are selected by DETECTOR and READPATT.
+
+
+Reference File Format
+---------------------
+
+A single IRS2 extension provides the complex coefficients for the correction,
```
hbushouse: Perhaps be a bit more explicit in indicating that the name of the extension, or the value of the EXTNAME keyword, is "IRS2". Also, I notice that the actual ref files in CRDS have a "DQ" extension too. Are those necessary and/or used? If so, they should be documented here too.

docs/jwst/rscd/reference_files.rst
```diff
@@ -1,21 +1,35 @@
...
+Reference File Types
+====================
+
+The RSCD correction step uses an RSCD reference file. 
+
+
+Description
+-----------
+
+This correction is only applied to integration > 1.
```
hbushouse: Shouldn't this kind of info be in the docs for the step itself, rather than the description of the ref file?

docs/jwst/rscd/reference_files.rst
```diff
@@ -43,7 +57,7 @@ It uses ``EXTNAME=RSCD`` and contains seventeen columns:
 * ILLUM_ZP: float
 * ILLUM_SLOPE: float
 * ILLUM2: float
-* PARAM3: :math:`b{3}' in equation
+* PARAM3: :math:`b{3}` in equation
 * CROSSOPT: float, crossover point
 * SAT_ZP: float
 * SAT_SLOPE: float
```
hbushouse: Are there short (1-3 word) descriptions available for columns like ILLUM_ZP, ILLUM_SLOPE, ILLUM2, SAT_ZP, SAT_SLOPE?

docs/jwst/straylight/reference_files.rst
```diff
@@ -1,5 +1,5 @@
-Reference File
-==============
+Reference File Types
+====================
 The default algorithm  in the MIRI MRS stray-light correction step uses information contained
 in the  meta data of the input image which maps each pixels to a slice or the  region between the
 slices, also known as the slice gaps. This information was previously loaded from a reference file into the meta data by the assign_wcs
```
hbushouse: "... loaded from a reference file into ..."  Can we say which reference file?

PR #2103, opened 2018-05-30
Ignore requests to skip test suite execution for regression tests.

rendinam:


hbushouse:
Why?

rendinam:
The regression tests are initiated on a schedule, so it makes sense that they should always run, even if the latest commit message happens to contain a `[skip ci]`/`[ci skip]` directive that was used to skip the per-push CI tests, as per conversation with @jdavies-st .

jdavies-st:
Thanks @rendinam!

I should have filed an issue.  I guess I filed in the hallway in conversation.

The problem was if you had merged a commit that, say, corrected a spelling mistake in the docs and had `[ci skip]` in the commit message, and if that was the last commit made before a Jenkins build began for the daily/nightly testing, then it would just skip the whole build.  Not good, when really the intention was to only skip the per-PR Travis CI builds.

I had run into this in testing when I was triggering builds by hand.  Now that we have them scheduled, it hasn't happened yet, but it could.  This prevents it and allows the regression tests to always run.

hbushouse:
👍

jdavies-st:
So this seems to have broken our Jenkins builds.

https://boyle.stsci.edu:8081/job/RT/job/JWST/61/console
```
> git rev-list 004ab489b16a6839e12840b454b68c1d20f03c41 # timeout=10
[Pipeline] End of Pipeline
java.lang.NoSuchMethodError: No such DSL method 'scm_checkout' found among steps [ansiColor, ...
```

rendinam:
The failure was due to the Jenkins internal configuration pointing to an incorrect version of the shared library that hosts the `scm_checkout` function. The configuration now references the correct version of the shared library.

PR #2108, opened 2018-05-31
step.input_dir from step.process()

nden:
`Step.input_dir` is not set when the step is run in an interactive session using `Step.process(input)`.
This PR fixes the immediate issue by calling `_set_input_dir` within `assign_wcs.process`.

Consider making `step._set_input_dir` public.

### DIFF
--- a/jwst/assign_wcs/assign_wcs_step.py
+++ b/jwst/assign_wcs/assign_wcs_step.py
@@ -1,4 +1,5 @@
 #! /usr/bin/env python
+import os.path
 from ..stpipe import Step
 from .. import datamodels
 import logging
@@ -65,7 +66,8 @@ class AssignWcsStep(Step):
             # Get the MSA metadata file if needed and add to reffiles
             msa_metadata_file = input_model.meta.instrument.msa_metadata_file
             if msa_metadata_file is not None:
-                msa_metadata_file = self.make_input_path(msa_metadata_file)
+                input_dir = os.path.split(self.make_input_path(input))[0]
+                msa_metadata_file = os.path.join(input_dir, msa_metadata_file)
                 reference_file_names['msametafile'] = msa_metadata_file
 
             result = load_wcs(input_model, reference_file_names)
--- a/jwst/assign_wcs/assign_wcs_step.py
+++ b/jwst/assign_wcs/assign_wcs_step.py
@@ -66,8 +66,8 @@ class AssignWcsStep(Step):
             # Get the MSA metadata file if needed and add to reffiles
             msa_metadata_file = input_model.meta.instrument.msa_metadata_file
             if msa_metadata_file is not None:
-                input_dir = os.path.split(self.make_input_path(input))[0]
-                msa_metadata_file = os.path.join(input_dir, msa_metadata_file)
+                self._set_input_dir(input))
+                msa_metadata_file = self.make_input_path(msa_metadata_file)
                 reference_file_names['msametafile'] = msa_metadata_file
 
             result = load_wcs(input_model, reference_file_names)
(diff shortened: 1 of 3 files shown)


jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -65,7 +66,8 @@ def process(self, input):
             # Get the MSA metadata file if needed and add to reffiles
             msa_metadata_file = input_model.meta.instrument.msa_metadata_file
             if msa_metadata_file is not None:
-                msa_metadata_file = self.make_input_path(msa_metadata_file)
+                input_dir = os.path.split(self.make_input_path(input))[0]
+                msa_metadata_file = os.path.join(input_dir, msa_metadata_file)
```
jdavies-st: Shouldn't this change be unnecessary if the code in `Step` is modified, as you've done below?  I believe the purpose of `Step.input_dir` was to avoid having to do this `os.path` stuff within each step's `process` method.

Further, I suspect we should be calling `process` directly.  Instead, we should be using `run` or `call`, as both of those eventually call `process` after doing the necessary setup of the step, which includes setting `Step.input_dir`.

nden:
@jdavies-st I just saw your comment ... after I made other changes. You may be right about running `step.process`. We should certainly not advertise it. However, I do use it occasionally when convenient and this fixes it. The code now also checks that step.input_dir is not None before attempting to use it.
Could we also consider making `step._set_input_dir` public?

I don't see the harm in this change (the new changes). However, if this is considered inappropriate I can live without it.

jdavies-st:
Agree that if we do advertise `Step.process` in our documentation, we should fix that.  Perhaps this discussion is part of the larger issue of what is the correct way to call a step or pipeline - `run`, `call`, `process`, etc.  I think this needs to be sorted out so that users (and developers) are less confused, but this topic is best discussed outside this PR.

I like the current changes, except I would change the name of the new test, as it now does not test `process`.

👍

hbushouse:
I agree that it can't hurt to leave these changes in place, but we should encourage the user (from JIRA ticket https://jira.stsci.edu/browse/JP-322) to use step.run() rather than step.process().

hbushouse:
Hmmm, looking at the original ticket again, it appears that they were using step.call(), not step.process():
```
stp = AssignWcsStep()
result = stp.call(step_input_file)
```

jwst/stpipe/step.py
```diff
@@ -935,7 +935,7 @@ def make_input_path(self, file_path):
         full_path = file_path
         if isinstance(file_path, str):
             original_path, file_name = split(file_path)
-            if not len(original_path):
+            if not len(original_path) and self.input_dir is not None:
```
stscieisenhamer: Inadvertently, I believe #2157 handles this. I would recommend waiting for that pr to merge, or merge it explicitly here, and ensure that it does the job.

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -65,6 +66,8 @@ def process(self, input):
             # Get the MSA metadata file if needed and add to reffiles
             msa_metadata_file = input_model.meta.instrument.msa_metadata_file
             if msa_metadata_file is not None:
+                if isinstance(input, str):
```
stscieisenhamer: Second @jdavies-st  in that, though not apparently stated anywhere, it is heavily implied that `process` should not be called directly. The real practical matter is that many other `process` definitions heavily rely on all sorts of parameters and setups. If one starts down the path of re-implementing setup steps when one is explicitly bypassing the setup, I fear for the future.

That being said, the `Step` API should allow one to easily do the step setup and `process` explicitly. At the moment, this is not the case: `Step.call` is actually close, but `Step.run` definitely has too much logic baked-in and could stand for a refactor.

So, for a simple issue as this with current architecture, the change is fine.

nden: Given that #2157 now implements a setter for step.input_dir, I think this is all I need.

PR #2110, opened 2018-05-31
skip assign_wcs if the MSAMETFL keyword is not populated correctly

### LINKED ISSUE #2107: Make opening of MSAMETFL more robust in assign_wcs
An update was recently made to the `asn_generator` rules so that level-2b ASN's are not created if an NRS_MSASPEC exposure does not have an `MSAMETFL` (MSA metadata file) specified, which means level-2b processing won't get triggered within the Ops environment.

But it's still possible for someone to try to manually execute `calwebb_spec2` on an NRS_MSASPEC exposure that doesn't have `MSAMETFL` defined or is set to "N/A". This causes an error in the `assign_wcs.nirspec.get_msa_metadata` function. The logic in that function should be "beefed up" to check for conditions like "MSAMETFL = 'N/A'", issue an error message to the log, and shutdown cleanly, rather than crashing.

nden:
Fixes #2107

### DIFF
--- a/jwst/assign_wcs/assign_wcs_step.py
+++ b/jwst/assign_wcs/assign_wcs_step.py
@@ -63,10 +63,16 @@ class AssignWcsStep(Step):
                 reference_file_names[reftype] = reffile if reffile else ""
 
             # Get the MSA metadata file if needed and add to reffiles
-            msa_metadata_file = input_model.meta.instrument.msa_metadata_file
-            if msa_metadata_file is not None:
-                msa_metadata_file = self.make_input_path(msa_metadata_file)
-                reference_file_names['msametafile'] = msa_metadata_file
+            if input_model.meta.exposure.type == "NRS_MSASPEC":
+                msa_metadata_file = input_model.meta.instrument.msa_metadata_file
+                if msa_metadata_file is not None and msa_metadata_file.strip() not in ["", "N/A"]:
+                    msa_metadata_file = self.make_input_path(msa_metadata_file)
+                    reference_file_names['msametafile'] = msa_metadata_file
+                else:
+                    log.info("MSA metadata file (MSAMETFL) is required for NRS_MSASPEC exposures.")
+                    input_model.meta.cal_step.assign_wcs = 'SKIPPED'
+                    log.warning("assign_wcs: SKIPPED")
+                    return input_model
 
             result = load_wcs(input_model, reference_file_names)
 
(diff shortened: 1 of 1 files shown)


jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -63,10 +63,16 @@ def process(self, input):
...
-            if msa_metadata_file is not None:
-                msa_metadata_file = self.make_input_path(msa_metadata_file)
-                reference_file_names['msametafile'] = msa_metadata_file
+            if input_model.meta.exposure.type == "NRS_MSASPEC":
+                msa_metadata_file = input_model.meta.instrument.msa_metadata_file
+                if msa_metadata_file is not None and msa_metadata_file.strip() not in ["", "N/A"]:
+                    msa_metadata_file = self.make_input_path(msa_metadata_file)
+                    reference_file_names['msametafile'] = msa_metadata_file
+                else:
+                    log.info("MSA metadata file (MSAMETFL) is required for NRS_MSASPEC exposures.")
```
hbushouse: Might want to bump this up to warning level, if not error.

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -63,10 +63,16 @@ def process(self, input):
...
+                    input_model.meta.cal_step.assign_wcs = 'SKIPPED'
+                    log.warning("assign_wcs: SKIPPED")
+                    return input_model
```
hbushouse: If this occurs, then none of the rest of the `calwebb_spec2` steps can proceed either. Is there already something implemented in `calwebb_spec2` that can detect that `assign_wcs` has been skipped and issue another warning (or error) and shutdown cleanly?

stscieisenhamer: Yep: https://github.com/STScI-JWST/jwst/blob/master/jwst/pipeline/calwebb_spec2.py#L169

jwst/assign_wcs/assign_wcs_step.py
nden: done - it's log.error now

hbushouse:
It occurs to me - after the fact of course - that we might also want imbed https://github.com/STScI-JWST/jwst/blob/master/jwst/assign_wcs/nirspec.py#L431 in a `try` block and again shutdown cleanly with an error message if the file can't be opened (e.g. doesn't exist, is in another directory, etc.).

nden:
I'll do that, though to get there one has to call the function directly, in which case they better be giving a name. But it's good to have this check.

nden:
Fixed in #2126

PR #2112, opened 2018-05-31
Update configobj as a bundled dependency for stpipe

### LINKED ISSUE #653: consider upgrading configobj
Continuing the discussion in PR #651, the questions of whether and how to upgrade configobj has been brought up. Currently, the embedded version works, but newer versions do exist that have been explicitly upgraded for Python3.  This issue is just to mark this discussion for future consideration.

jdavies-st:
We've had `configobj` version 4 embedded in `stpipe` for a very long time.  Since we are now python 3 only, and `configobj` is being actively developed for python 3.6 and above here:

https://github.com/DiffSK/configobj

it makes good sense to remove our internal version and use the open source version which is undergoing active development and bug fixing.  See discussion on PR #651.

This PR removes our internal version and adds the open source version as a dependency to this package.

Merging this should be coordinated with updates to our astroconda and astroconda-dev build recipes.  I believe `configobj` should be added to those recipes before this is merged, @jhunkeler?

Resolves #653.


jhunkeler:
I believe the only thing left is to change the documentation:
```
docs/jwst/stpipe/devel_step.rst:210:<http://www.voidspace.org.uk/python/configobj.html#validation>`_.
docs/jwst/stpipe/user_pipeline.rst:29:<http://www.voidspace.org.uk/python/configobj.html>`_ library.
docs/jwst/stpipe/user_step.rst:24:<http://www.voidspace.org.uk/python/configobj.html>`_ library to parse
```
The voidspace.org.uk site points to `ConfigObj-4`. The new docs are here:
https://configobj.readthedocs.io/

nden:
I just want to caution that it's harder to maintain an external dependence than to distribute it with the code.
`configobj` would probably be also on user systems and may be picked up from there so we need to make sure we have a version we can work with. We also need very good test coverage on all our use cases to make sure that changes upstream are caught and addressed quickly. And we may need to have to support several versions of configobj if there's ever a backwards incompatible change in the package.
I look at this as another dependence similar to astropy except that we have even less control over it.

jdavies-st:
I look at this as a dependency similar to numpy.  We have no control over it, except for the fact that it is open source.  But we completely depend on it.

If we really did want to keep it internal, then we could at least update the internal one to the latest version 5.  But it would require us to remain vigilante to bugs and bug fixes.  And given our past record...

Or better yet (if we're concerned about an uncontrolled dependency), in our builds we could peg the version at the last "fully tested" version, as we have done in the past with numpy.  At least this would remind us every time we build that we've got a pegged dependency.

Test coverage for parts of this package are quite low.  I don't know what the coverage for `stpipe` is.  I can run `pytest --cov` today and find out.  I'll include all the `--runslow` and `--bigdata` tests.

jdavies-st:
The coverage of the unit tests (as currently run per-PR on Travis CI) are the following:
```
jwst/stpipe/__init__.py                                           8      0   100%
jwst/stpipe/cmdline.py                                          149     30    80%
jwst/stpipe/config_parser.py                                    147     30    80%
jwst/stpipe/configobj/__init__.py                                 0      0   100%
jwst/stpipe/configobj/configobj.py                             1106    537    51%
jwst/stpipe/configobj/validate.py                               281     95    66%
jwst/stpipe/crds_client.py                                       58     15    74%
jwst/stpipe/function_wrapper.py                                   8      3    62%
jwst/stpipe/hooks.py                                             25     12    52%
jwst/stpipe/linear_pipeline.py                                   65      8    88%
jwst/stpipe/log.py                                              129      6    95%
jwst/stpipe/pipeline.py                                          92     36    61%
jwst/stpipe/step.py                                             360     60    83%
jwst/stpipe/subproc.py                                           35     29    17%
jwst/stpipe/suffix.py                                            86      9    90%
jwst/stpipe/utilities.py                                         36      2    94%
```
While some of our `stpipe` coverage is decent, our coverage of the internal version of `configobj` is 51% and 66%.

jdavies-st:
Another option would be just to use the version of `configobj` bundled with `astropy`.  It can be imported this way
```
from astropy.extern.configobj import configobj, validate
```
That way we don't add a new dependency, we retain the "pegging" to a particular version, and it's more likely that the astropy project is going to keep the bundled version updated with the latest security and bug fixes.

Does this seem like a better solution?

jdavies-st:
On the flip side, the version of `configobj` bundled with `astropy` has been essentially forked from the open source version being developed. The astropy-bundled version has had all the python 2 stuff removed from it.

https://github.com/astropy/astropy/commits/master/astropy/extern/configobj/configobj.py

So we could go with the `astropy` version (it works fine; our tests pass) which avoids `six` (though we depend on it indirectly elsewhere), or we can go with the publicly-developed version and add yet another dependency to our already very large list of dependencies for `jwst`.

jdavies-st:
Considering input above, the conservative solution is to update our bundled version to the latest version of `configobj`.  I've done this by copying over that version which is currently used by `astropy`.  This version bundled with astropy is essentially a fork where the python 2 support has been removed, which suits us fine.  This also means the `six` dependency has been removed, which is also nice.

We can bundle this as an external dependency just as we have up to this point, but put it in a more explicit location so it is clear it is a bundled dependency.  This will hopefully make maintenance easier if needed in the future.
```
jwst/extern/configobj/configobj.py
jwst/extern/configobj/validate.py
```
This final commit does this.  Everybody happy?

PR #2113, opened 2018-06-01
Divided median differences by the group time in the calculation of th…

dmggh:
…e estimated count rates. Included division of sqrt(2) for the readnoise values from the reference file, because the ref file values are for the noise in a CDS pair of reads but will be applied to individual reads.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -340,6 +340,8 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
     median_diffs_2d /= n_int
     med_rates = median_diffs_2d/group_time
 
+    del median_diffs_2d
+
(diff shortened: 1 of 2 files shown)


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -336,11 +336,9 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
             opt_res.cr_mag_seg[num_int, :, rlo:rhi, :] = \
                        data_diff * (dq_cr != 0)
 
-    # Compute the final 2D array of differences.
+    # Compute the final 2D array of differences; create rate array
     median_diffs_2d /= n_int
-    med_slopes = median_diffs_2d.copy()
-
-    del median_diffs_2d
```
hbushouse: Do you still need the median_diffs_2d array later, or could you delete it, as had been done before?

PR #2119, opened 2018-06-01
Remove test_nrs_fs_multi_spec2_3 regression test

### LINKED ISSUE #2117: Resolve identical regression tests
The following two tests are identical:
```
jwst/tests_nightly/general/pipelines/test_nrs_fs_multi_spec2_3.py
jwst/tests_nightly/general/pipelines/test_nrs_fs_single_spec2.py
```
They take in the same input and produce the same output.
```
> diff test_nrs_fs_multi_spec2_3.py test_nrs_fs_single_spec2.py 
12c12
< def test_nrs_fs_multi_spec2_3(_bigdata):
---
> def test_nrs_fs_single_spec2(_bigdata):
16c16
<     using the ALLSLITS subarray and detector NRS2.
---
>     that uses a single-slit subarray (S200B1).
```

We should remove one of them or modify one to do what we think it should be doing.

jdavies-st:
Removing this as it is a duplicate of `test_nrs_fs_single_spec2.py`.

Resolves #2117.


jdavies-st:
I will close this and modify the issue.

PR #2137, opened 2018-06-07
Purge unused dependencies

jhunkeler:
This prevents `drizzlepac` and `fitsblender` from getting used by RTD, Travis, and Jenkins.

Here's a PR to do the same thing to our `astroconda-dev` recipe:
https://github.com/astroconda/astroconda-dev/pull/137

### DIFF
--- a/.rtd-environment.yml
+++ b/.rtd-environment.yml
@@ -1,6 +1,6 @@
 channels:
   - http://ssb.stsci.edu/astroconda-dev
-  
+
 dependencies:
   - python>=3
   - setuptools
@@ -9,7 +9,6 @@ dependencies:
   - crds
   - dask
   - drizzle
-  - drizzlepac
   - gwcs
   - jsonschema
   - jplephem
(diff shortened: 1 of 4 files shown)


jdavies-st:
We should be able to merge this now since 0.9.5 was released to DMS yesterday.  @hbushouse?

stscicrawford:
There might have been some new errors in the build.   Let's wait for the okay before merging

hbushouse:
In fact there was a new error in the build (see #2139). But we don't need the changes included in #2130 for the DMS B7.1.3 delivery, and DMS test runs executed so far have not shown any other new errors. So we're probably going to just create a new build from the state of the repo immediately before #2130, in which case changes merged after that shouldn't have any effect (theoretically).

jdavies-st:
Why do we depend on `dask`?  Isn't that for multiprocessing?  It is not imported anywhere in `jwst`.

Also, can we merge this?

jdavies-st:
Let's remove `dask` later when we can first figure out why it was put in in the first place.

PR #2097, opened 2018-05-28
update SourceCatalogStep for standard Step I/O API

stscieisenhamer:
Resolves #1994

Issue #1951, opened 2018-04-26
default output file name not set when not running from command line with save_results


stscieisenhamer:
# Issue
`save_results` allows output to occur from `Step`/`Pipeline` when not run from the command line, as desired. However, when run from the command line, `save_results` is set and the `output_file` is taken from the primary input.

But, when running from the `call`, if `save_results = True`, there is no default input file name, so the behavior is different. The default output names are all the step-generated ones. 

Desired behavior in this case, is to attempt to use the primary input, as with the command line. However, since the input may also be an arbitrary object, at least use the top-level pipeline/step name for the output file template for all the subsequent steps.

# Implementation
## To Do
- [x] documentation
- [x] tests
- [x] code
    - [x] Centralize setup of `input_dir` and `input_file`
- [x] tests
- [x] documentation

PR #2146, opened 2018-06-08
fix for save_catalog bug introduced in #2097

sosey:
This updates the configspec for `source_catalog` to set save_results to true, the default in `stpipe` is false and this step has no `output_file` spec. There may be a better way to do this, @stscieisenhamer should verify this fix.

I also added the datamodel filename as the the root for the output. 

Testing this locally with my wfss data give results I would expect.

### DIFF
--- a/jwst/source_catalog/source_catalog_step.py
+++ b/jwst/source_catalog/source_catalog_step.py
@@ -25,6 +25,7 @@ class SourceCatalogStep(Step):
         deblend = boolean(default=False)      # deblend sources?
         output_ext = string(default='.ecsv')  # Default type of output
         suffix = string(default='cat')        # Default suffix for output files
+        save_results = boolean(default=True)  # Save the output catalog
     """
 
     def process(self, input):
@@ -48,7 +49,7 @@ class SourceCatalogStep(Step):
             self.log.info('Detected {0} sources'.format(len(catalog)))
 
             if self.save_results:
-                cat_filepath = self.make_output_path()
+                cat_filepath = self.make_output_path(basepath=model.meta.filename)
                 catalog.write(
                     cat_filepath, format='ascii.ecsv', overwrite=True
                 )


larrybradley:
LGTM, thanks!

jdavies-st:
One reason it is not being returned in memory is that back when this was originally written, I believe there was no way to return something that was not a datamodel and have `stpipe` handle it properly.  This might be fixed now, but generally, astropy tables and datamodels do not have the same `write` or `save` method API.  So they're not even duck-type-able.  Maybe we should fix this.

stscieisenhamer:
Presuming that the scenario is running the step on its own in a script, which is the situation described by #1951, I have [created a branch](https://github.com/stscieisenhamer/jwst/tree/i1951_defoutputname) that address the issue. @sosey please try that out and let me know.

sosey:
Friday, I was calling the step directly like this using defaults:
```
dpm=DrizProductModel(direct_file)
sc=source_catalog_step.SourceCatalogStep()
sc.process(dpm)
```
with that call, no output file is saved, no catalog is returned

I hadn't thought then to check the defaults setting on the object call, that's not cemented in my brain yet: 
```
sc=source_catalog_step.SourceCatalogStep(save_results=True)
sc.process(dpm)

2018-06-11 10:52:00,345 - stpipe.SourceCatalogStep - INFO - Detected 49 sources
2018-06-11 10:52:00,396 - stpipe.SourceCatalogStep - INFO - Wrote source catalog: step_SourceCatalogStep_cat.ecsv
```
We'd like the output file root from `source_catalog`, by default, to contain the input filename

I can also set the output_filename there:
```
sc=source_catalog_step.SourceCatalogStep(save_results=True, output_file=dpm.meta.filename)
sc.process(dpm)
2018-06-11 10:56:11,970 - stpipe.SourceCatalogStep - INFO - Detected 49 sources
2018-06-11 10:56:12,017 - stpipe.SourceCatalogStep - INFO - Wrote source catalog: niriss_F200W_image_position1_cat.ecsv
```

I did this example on your branch. I haven't had a chance to look at the pipeline section, but if it's calling source catalog in the same way then I think we are okay. Next,  I'll go back and check all the different call methods and make sure everything still as I expect on output.

sosey:
the source catalog is returned from the downline call to `make_source_catalog()` so anyone wanting to work with the catalog object directly (and saving or further editing it)  can get that by using `from jwst.source_catalog.source_catalog import make_source_catalog`

closing this PR. thank you all for the discussion!

PR #2147, opened 2018-06-08
Wrap validation assertions with test for strict_validation

bernie-simon:
Reference file models have a validate method which contain assertions about its properties. This update wraps the assertions in a try block. If the exception is thrown, it tests strict_validation. If it is true, the error is re-thrown. If it is false, the exception is converted to a warning. This update also adds a call to the super-class validate method to a few cases where it was not being called by a reference file model.

### DIFF
--- a/jwst/datamodels/reference.py
+++ b/jwst/datamodels/reference.py
@@ -1,5 +1,6 @@
+import warnings
 from .model_base import DataModel
-
+from .validate import ValidationWarning
 
 __all__ = ['ReferenceFileModel']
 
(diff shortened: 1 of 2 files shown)


nden:
I think this looks good for fixing the immediate problem. However, I need to fix the wcs schemas first. You are not seeing errors here but if you try to validate some of the wcs reference files with their schemas you will see the errors. This PR **cannot be merged** until the other issue is fixed.

nden:
@jaytmiller Todd, is crds is using `strict_validation=False` when opening models? If so, let's merge this as a fix to the wcs schemas is still not ready.

jaytmiller:
CRDS is using default datamodels.open() parameters.  If strict_validation is the default,  then "yes",  which potentially could be an issue.   If not,  eventually CRDS will use it somewhere during file ingest,   but there will be breathing room to adapt.    I haven't done any testing with this,  if it's an issue at all,  it will be that validation errors blind the rest of CRDS certify;  I'd expect CRDS to trap the error and it should reject the file correctly,  but the diagnostics may be poor.

nden:
`strict_validation=False` by default.

jaytmiller:
Great,  so I think that puts CRDS in the not-yet-benefiting camp but will give me a chance to kick the tires and turn it on in the right places.   Thanks!

PR #2153, opened 2018-06-12
Revert "NRS: fix input units to meters when filter=OPAQUE"

nden:
Reverts STScI-JWST/jwst#2134


nden:
There's a better fix to this problem (coming later). This one introduces a problem with computing the bounding_box in the case when the pipeline stops at the MSA. The wavelength range as currently recorded in the header is in units of meters while this assumes the inputs are always in microns.

PR #2162, opened 2018-06-15
WIP: fix NRS prism model

nden:
DO NOT MERGE:

The team uses a different wavelength range than what's in the reference file.
Results improve if we use that wavelength range in the pipeline.

use `scipy.interp1d` instead of `scipy.interpnd`


nden:
superseded by #2163

Issue #1188, opened 2017-09-08
exposures that get Level 2b processing


stscieisenhamer:
Prompted by [this comment](https://github.com/STScI-JWST/jwst/issues/735#issuecomment-292226126), this issue documents exactly which exposures will go through any type of Level 2b processing, such as Spec2Pipeline or Image2Pipeline.  Please review, update, and if updated, leave a comment that one has done so. Questions in following comments.

EXP_TYPE | Association Exposure Type | Specials | Level2b?
---------- | --------------------------- | --------- | ---------
NRC_IMAGE | science | n/a | YES
NRC_GRISM/NRC_WFSS | science | n/a | YES
NRC_TACQ | target_acquisition | n/a | YES
NRC_TACONFIRM | target_acquisition | n/a | YES
NRC_CORON | science | n/a | YES
NRC_CORON | psf | PSF | YES
NRC_TSIMAGE | science | n/a | YES
NRC_TSGRISM | science | n/a | YES
NRC_FOCUS | science | n/a/ | YES
NRC_DARK | dark | n/a | NO
NRC_FLAT | flat | n/a | NO
NRC_LED | science | n/a | NO
 | | |
MIR_IMAGE | science | n/a | YES
MIR_TACQ | target_acquisition | n/a | YES
MIR_LYOT | science | n/a | YES
MIR_LYOT | psf | PSF | YES
MIR_4QPM | science | n/a | YES
MIR_4QPM | psf | PSF | YES
MIR_LRS-FIXEDSLIT | science | n/a | YES
MIR_LRS-FIXEDSLIT | background | BACKGROUND | YES
MIR_LRS-SLITLESS | science | n/a | YES
MIR_LRS-SLITLESS | background | BACKGROUND | YES
MIR_MRS | science | n/a | YES
MIR_MRS | background | BACKGROUND | YES
MIR_DARKIMG | dark | n/a | NO
MIR_DARKMRS | dark | n/a | NO
MIR_FLATIMAGE | flat | n/a | NO
MIR_FLATIMAGE-EXT | flat | n/a | NO
MIR_FLATMRS | flat | n/a | NO
MIR_FLATMRS-EXT | flat | n/a | NO
MIR_CORONCAL | science | n/a | YES
 | | | 
NRS_WATA | target_acquisition | n/a | YES
NRS_MSATA | target_acquisition | n/a | YES
NRS_TACONFIRM | target_acquisition | n/a | YES
[...]

PR #2166, opened 2018-06-18
process NRS_AUTOFLAT as a MOS observation

### LINKED ISSUE #1984: NRS_AUTOFLAT processing
#1188 [comment](https://github.com/STScI-JWST/jwst/issues/1188#issuecomment-386470346) indicates that NRS_AUTOFLAT should be processed through assign_wcs and extract_2d as an MSA exposure.

nden:
Needs rmaps update to include NRS_AUTOFLAT in order to pick the reference files.
Code was tested local reference files.
**Alternatively**, we can merge this and let the INS team use local reference files during their tests.

The INS team indicates that NRS_AUTOFLAT observations should be processed by assign_wcs and extract_2d as MOS observations.

Fixes #1984.

### DIFF
--- a/jwst/assign_wcs/assign_wcs.py
+++ b/jwst/assign_wcs/assign_wcs.py
@@ -50,7 +50,7 @@ def load_wcs(input_model, reference_files={}):
         output_model.meta.cal_step.assign_wcs = 'COMPLETE'
         exclude_types = ['NRS_FIXEDSLIT', 'NRS_BRIGHTOBJ', 'NRS_IFU',
                          'NRS_MSASPEC', 'NRS_LAMP', 'MIR_MRS', 'NIS_SOSS',
-                         'NRC_WFSS', 'NRC_TSGRISM', 'NIS_WFSS']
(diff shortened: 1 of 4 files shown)


nden:
I suggest we merge this. The INS team can use reference files locally to run it. If they feel adding the EXP_TYPE to crds, they can do this without changes in the cal code.

hbushouse:
I notice that in #1188 there's a comment saying the NIRSpec team is OK with only applying level-2a processing to AUTOFLAT exposures, at least for now. @stscieisenhamer Is the level-2b ASN processing setup to skip AUTOFLAT's? If so, then the Ops environment would adhere to the rule of not applying level-2b to AUTOFLAT, and then merging this PR would at least give the team the ability to process them themselves if they want to, which sounds like an acceptable solution.

stscieisenhamer:
AUTOFLAT does not get Level-2b processing. Merging of this PR will have no affect on Ops.  See [comment](https://github.com/STScI-JWST/jwst/issues/1188#issuecomment-388250551).

👍

PR #2168, opened 2018-06-19
Ensure calwebb_spec2 fails when `assign_wcs` fails or is skipped

### LINKED ISSUE #2151: calspec2 bogus product when MSAMETFL is not available
DMS testing of 0.9.6 has revealed that when ``calspec2`` is run on a NIRSpec MSA exposure and the MSAMETFL isn't specified or can't be opened, the way in which ``calspec2`` aborts processing results in a product being saved that has a file name of the form "jwxxxxxxx_nrs1_spec2pipeline.fits". We should look for a way to not return anything to `stpipe` in this case, so that no output product gets created at all.

stscieisenhamer:
Ensure that `calwebb_spec2` raises an exception for a failed or skipped `assign_wcs` step.

Resolves #2151

### DIFF
--- a/jwst/pipeline/calwebb_spec2.py
+++ b/jwst/pipeline/calwebb_spec2.py
@@ -39,7 +39,8 @@ class Spec2Pipeline(Pipeline):
     """
 
     spec = """
-        save_bsub = boolean(default=False) # Save background-subracted science
+        save_bsub = boolean(default=False)        # Save background-subracted science
+        fail_on_exception = boolean(default=True) # Fail if any product fails.
     """
 
     # Define aliases to steps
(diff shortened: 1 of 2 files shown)


jwst/pipeline/calwebb_spec2.py
```diff
@@ -77,20 +78,34 @@ def process(self, input):
...
+                    product,
+                    asn['asn_pool'],
+                    asn.filename
+                )
+            except Exception as exception:
+                exceptions.append((product['name'], exception))
+            else:
+                if result is not None:
+                    self.save_model(result)
+                    self.closeout(to_close=[result])
```
hbushouse: I've never seen this ``closeout`` method before. What does it do?

stscieisenhamer: Hmmm, that was leftover from the memory leak issues. It probably does nothing useful, but will investigate to ensure that.

jwst/pipeline/calwebb_spec2.py
```diff
@@ -39,7 +39,8 @@ class Spec2Pipeline(Pipeline):
     """
 
     spec = """
-        save_bsub = boolean(default=False) # Save background-subracted science
+        save_bsub = boolean(default=False)        # Save background-subracted science
+        fail_on_exception = boolean(default=True) # Fail if any product fails.
```
jdavies-st: You may want to add this `fail_on_exception` spec member to the default .cfg file for this pipeline.  =)

jdavies-st: Otherwise looks good.

PR #2170, opened 2018-06-19
raise an error if msa metafile is missing

nden:
as discussed in #2151


nden:
Yes, rebased earlier.

PR #2175, opened 2018-06-20
Round S_REGION RA/Dec values to 9 decimal places

### LINKED ISSUE #2165: Round S_REGION keyword corner values to significant digits
The S_REGION keyword records the sky coordinates of the corners of an image.  This keyword is used by the archive to be able to show a footprint in sky coords.  Currently we record those at machine precision.  We should instead record only significant digits here, say to 100th of a pixel precision.

Currently, writing them out at machine precision makes diff'ing the keyword very difficult, as changes the last digit of one or more corners cause the diff to fail.  This is made worse by the fact that all these machine-precision floats are recorded as a FITS keyword string.

jdavies-st:
Resolves #2165

### DIFF
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -572,10 +572,10 @@ def update_s_region(model):
 
     s_region = (
         "POLYGON ICRS "
-        " {0} {1}"
-        " {2} {3}"
-        " {4} {5}"
-        " {6} {7}".format(*footprint.flatten()))
+        " {0:.9f} {1:.9f}"
+        " {2:.9f} {3:.9f}"
+        " {4:.9f} {5:.9f}"
+        " {6:.9f} {7:.9f}".format(*footprint.flatten()))
     if "nan" in s_region:
         # do not update s_region if there are NaNs.
         log.info("There are NaNs in s_region")


hbushouse:
Is this the only place within the package that the value of S_REGION is composed?

jdavies-st:
Yes, this is the only place.  This function is called in a few different locations though - `assign_wcs` and `resample`.  I.e., any time a new WCS or output frame is created.

jwst/assign_wcs/util.py
```diff
@@ -572,10 +572,10 @@ def _bbox_from_shape(model):
 
     s_region = (
         "POLYGON ICRS "
-        " {0} {1}"
-        " {2} {3}"
-        " {4} {5}"
-        " {6} {7}".format(*footprint.flatten()))
+        " {0:.9f} {1:.9f}"
```
philhodge: This looks fine, but as an alternative you could have used the `round` function, e.g.

    round(math.pi, 9)
    3.141592654

jdavies-st: `round()` is built into the string formatting sublanguage.  I had initially implemented it with a call to `np.round()` but then realized it was not necessary.  A nice discovery!

Issue #1251, opened 2017-09-28
Remove the schema from ModelContainer


jdavies-st:
`ModelContainer` currently has schema entries that correspond to some items found in Level 3 association tables.  These should be removed, and any code that relies on them should switch to using the `ModelContainer.asn_table` attribute to gather this information.

jdavies-st:
After discussion with @nden today, it's clear that `ModelContainer` should not have a schema, as it does not correspond to a FITS file and thus does not need the schema mapping for serialization.  As long as its members have schemas, the members can be serialized each to its own file.

Updating this issue to remove the schema completely.  Any metadata that `ModelContainer` needs to carry along with it can be stored in instance attributes or properties.

nden:
@stscijgbot-jp

stscijgbot-jp:
_This issue is tracked on JIRA as [JP-2040](https://jira.stsci.edu/browse/JP-2040)._

PR #2181, opened 2018-06-21
ModelContainer, Level 3 schema cleanup; resample output metadata updates

jdavies-st:
- Remove unnecessary, unused metadata from `ModelContainer` schema
- Remove unnecessary, unused metadata from level 3 product schema
- move population of association metadata up into the `ResampleStep` classes
- update S_REGION correctly for `ResampleSpecStep` outputs

### DIFF
--- a/jwst/datamodels/schemas/container.schema.yaml
+++ b/jwst/datamodels/schemas/container.schema.yaml
@@ -4,10 +4,6 @@ properties:
     title: Level 3 Schema Metadata
     type: object
     properties:
-      date:
-        title: Date this file was created (UTC)
-        type: string
-        fits_keyword: DATE
       table_name:
         title: Name of ASN table 
         type: string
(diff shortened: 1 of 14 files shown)


jdavies-st:
NOTE: Because we are now populating `meta.asn.pool_name` and `meta.asn.table_name` in the output of `ResampleSpecStep`, this puts a new "Association information" section into the primary headers of the output `_s2d` files.  This will cause a number of our overnight regression tests to fail.

This is a reminder to me to fix the comparison data for those when this is merged.

jdavies-st:
Addresses issue #1251.  Though we do not completely remove the `ModelContainer` schema here.  We retain the association information needed to produce output names.  It's possible to move these calls to access the `asn_table` stored within the `ModelContainer`, but these are not serialized.  Perhaps at a future date.

jwst/datamodels/schemas/lev3_prod.schema.yaml
```diff
@@ -4,238 +4,35 @@ properties:
...
-              fits_keyword: DWTSCL
-            resample_bits:
-              title: Bit values to consider good when interpreting input DQ arrays
-              type: integer
-              fits_keyword: RESBITS
             weight_type:
               title: Type of drizzle weighting to use in resampling input
               type: string
               enum: ['exptime','error']
-              fits_keyword: RESWHT
```
hbushouse: Why remove the keyword definition for this? Are you assuming it's not something of interest to the user and hence doesn't need to be documented in a keyword?

hbushouse: Well for now let's leave the keyword out, until we actually implement IVM weighting. We can put it back in then.

jwst/datamodels/schemas/container.schema.yaml
```diff
@@ -16,32 +11,14 @@ properties:
         title: Name of ASN pool used to generate this table
         type: string
         fits_keyword: ASNPOOL
```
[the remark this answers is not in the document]
hbushouse: For now let's just remove the ``fits_keyword`` definitions for ASNTAB and ASNPOOL, so that they're only defined in core.schema.yaml. You can keep the meta data objects for use within ``ModelContainer``'s.

PR #2186, opened 2018-06-24
use tagged tree in validaiton of custom tags

nden:
This allows custom tags, like transforms, to be validated. With this fix the WCS  reference files validate against their schemas.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,5 +7,6 @@ recursive-include jwst *.txt
 recursive-include jwst *.cfg
 recursive-include jwst *.csv
 recursive-include jwst *.yaml
+recursive-include jwst *.asdf
 prune relic/relic/__pycache__
 prune relic/.git
--- a/conftest.py
+++ b/conftest.py
@@ -14,9 +14,9 @@ try:
 except (NameError, KeyError):
     pass
 
-pytest_plugins = [
-    'asdf.tests.schema_tester'
-]
+#pytest_plugins = [
+    #'asdf.tests.schema_tester'
+#]
 
 # Add option to run slow tests
 def pytest_addoption(parser):
--- a/jwst/datamodels/tests/data/collimator_fake.asdf
+++ b/jwst/datamodels/tests/data/collimator_fake.asdf
@@ -1,0 +1,40 @@
+#ASDF 1.0.0
+#ASDF_STANDARD 1.2.0
+%YAML 1.1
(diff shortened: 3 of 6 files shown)


nden:
I don't understand why the new test is failing on Travis and passing locally.

nden:
Digging a little bit into the test logs to figure out why the two new tests pass on Jenkins and locally but fail on Travis. From the logs, the reported versions of jwst used in testing are different:

Travis:  Normalizing '0.9.7a.dev48+g47993893' to '0.9.7a0.dev48+g47993893'

Jenkins:  jwst==0.9.7a0.dev47+gdbaad5a4

The Jenkins SHA corresponds to the SHA of the latest commit in this PR. I am not sure how Travis came up with `g47993893` and what revision this is. What am I missing? Perhaps @jhunkeler or @rendinam are familiar with this?

pllim:
https://docs.travis-ci.com/user/pull-requests/ -- looks like Travis uses a different strategy from Jenkins and it is not something we can change.

pllim:
@nden , here is the astropy helpers magic that pulls in Astropy test runner for packages using helpers:

https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/setup_helpers.py#L288

https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/setup_helpers.py#L149

And here is how the magic is applied in package template:

https://github.com/astropy/package-template/blob/cookiecutter/%7B%7B%20cookiecutter.package_name%20%7D%7D/setup.py#L96

None of these explain what is going on here though because it looks like `jwst` rolled its own test runner that is independent of Astropy.

hbushouse:
Will the implementation of these updates allow for #2147 to be merged?

nden:
@hbushouse Yes, this is the idea.

nden:
Summary of the problem with Travis tests:

It appears that the [asdf plugin](https://github.com/spacetelescope/jwst/blob/master/conftest.py#L17) causes the command `python setup.py test` run by Travis to fail the two new tests, basically ignoring the change made in this PR. How exactly the plugin changes the environment needs to be further investigated. For now, I am commenting out the plugin to allow this PR to be merged. The plugin tests the examples in the transforms schemas. The problem does not exist if the tests are run by pytest directly. In this case the schemas are tested as well.

nden:
Issue #2205 was created as a remonder to look into the plugin issue.

PR #2191, opened 2018-06-26
Remove deprecated photutils.properties_table

jdavies-st:
The `properties_table()` function is deprecated in `photutils`.  The `SourceCatalog.to_table()` method is used instead.


larrybradley:
LGTM, thanks!  The test failure is unrelated.

PR #2193, opened 2018-06-27
Strip folder path from table and pool name keywords

### LINKED ISSUE #2192: ensure meta.asn.table_name contains only filename
# Issue
`calwebb_image2` populates `meta.asn.table_name` with the full path of the association table. Only the basename portion should be used.

stscieisenhamer:
Resolves #2192

### DIFF
--- a/jwst/pipeline/calwebb_image2.py
+++ b/jwst/pipeline/calwebb_image2.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 from collections import defaultdict
+import os.path as op
 
 from .. import datamodels
 from ..associations.load_as_asn import LoadAsLevel2Asn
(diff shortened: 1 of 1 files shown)


jwst/pipeline/calwebb_image2.py
```diff
@@ -108,8 +109,8 @@ def process_exposure_product(
             input = datamodels.open(science)
 
         # Record ASN pool and table names in output
-        input.meta.asn.pool_name = pool_name
-        input.meta.asn.table_name = asn_file
+        input.meta.asn.pool_name = op.basename(pool_name)
```
jdavies-st: I think `pool_name` is always pulled from inside the association, so it never has a path associated with it.  Is that correct?  If it does have a path, then we should use `basename()` on all the other places in the pipeline we populate `pool_name`.

PR #2203, opened 2018-06-28
Updates to entire package for py2-to-3 conversion

hbushouse:
Lots and lots of updates for #1392. Removed all "future" imports, removed "object" from all class definitions, and removed some instances of "six." There's still a lot of use of "six" in ``fits_generator``, which I'm going to let @stscirij deal with.


stscirij:
This is probably a good time to move to the most recent version of the pyparsing package on github (https://github.com/pyparsing/pyparsing) rather than using our version frozen from the time Mike D was working on the fits generator.

PR #2206, opened 2018-06-29
Add usage examples to blendmeta docs [ci skip]

stsci-hack:
Mostly minor clean-up of model_blender docs, but it does add a usable example of how to run model_blender with in-memory models.


hbushouse:
@stsci-hack OK to merge or did you want to make additional updates?

Issue #1652, opened 2018-01-10
Task: Update the refpix step to handle subarray exposures


hbushouse:
The JCCWG has decided on an approach to use for reference pixel correction when an exposure is taken in subarray mode, at least for the NIR instruments. Put simply, the approach boils down to using whatever reference pixels are available in the subarray and applying the same processing as for full-frame exposures. Information from the JCCWG is available at:

https://confluence.stsci.edu/display/JWSTCC/Vanilla+Reference+Pixel+Correction

Note that as of right now, there appears to be some discussion going on yet as to the exact implementation of the algorithm.

hbushouse:
INS is tracking in https://jira.stsci.edu/browse/JP-264

hbushouse:
Fixed by #2207

PR #2207, opened 2018-06-29
Refpix subarrays

stscirij:
Addresses #1652

### DIFF
--- a/jwst/refpix/reference_pixels.py
+++ b/jwst/refpix/reference_pixels.py
@@ -26,7 +26,17 @@
 #       these smoothed and scaled values should be subtracted from every pixel
 #       in the corresponding row.
 
-from __future__ import division
+#
+#  Subarray processing added 7/2018
+#
+#  For NIR exposures, calculate the clipped means of odd and even columns
+#  in detector coordinates.  Subtract the odd mean from the odd columns, and
+#  the even mean from the even columns.  If there are no reference pixels in the
+#  subarray, omit the refpix step.
+#
+#  For MIRI subarray exposures, omit the refpix step.
+# 
+
 import numpy as np
 from scipy import stats
 import logging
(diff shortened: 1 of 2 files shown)


jwst/refpix/reference_pixels.py
```diff
@@ -181,6 +184,56 @@ def sigma_clip(self, data, dq, low=3.0, high=3.0):
...
+        """
+        Transfer the reference-pixel corrected data arrays from the
+        dataset to the input model
+
+        Parameters:
+        -----------
+
+        input_model: JWST datamodel
+
+        The datamodel the data arrays are to be transferred to
```
hbushouse: "The datamodel to which the data arrays are to be transferred"  (I find it ironic that an Amurican is correcting a Brit's grammar ;-) )

jwst/refpix/reference_pixels.py
```diff
@@ -181,6 +184,56 @@ def sigma_clip(self, data, dq, low=3.0, high=3.0):
...
+        embedded_subarray: nddata array
+        
+        The subarray embedded in the full-frame array
+        
+        """
+        rowstart = self.ystart - 1
+        rowstop = rowstart + self.ysize
+        colstart = self.xstart - 1
+        colstop = colstart + self.xsize
+        embedded_subarray = self.data[:, :, rowstart:rowstop, colstart:colstop]
```
hbushouse: Is it ever possible to have the calculation of these slice indexes end up being outside the bounds of the actual array and hence we should implement some kind of checking to make sure that doesn't happen?

jwst/refpix/reference_pixels.py
```diff
@@ -695,6 +755,61 @@ def do_corrections(self):
...
+        refdq = dqflags.pixel['REFERENCE_PIXEL']
+        (nints, ngroups, nrows, ncols) = self.data.shape
+        for integration in range(nints):
+            for group in range(ngroups):
+                #
+                # Get the reference values from the top and bottom reference
+                # pixels
+                #
+                thisgroup = self.data[integration, group].copy()
+                refpixindices = np.where(np.bitwise_and(self.pixeldq, refdq) == refdq)
```
hbushouse: Where is it in the upstream processing that we're assuming the ``REFERENCE_PIXEL`` DQ flags got added to the ``pixeldq`` array? Are they supposed to come from the MASK reference file that gets loaded by the ``dq_init`` step? I don't recall if all MASK ref files have yet been updated to include those flags. Are the checks below sufficient to keep things from blowing up if those flags haven't been set?

stscirij: Yes, I'm assuming they get added to the pixeldq array from the dq_init step.  If the flags haven't been set, the step should be skipped.

jwst/refpix/reference_pixels.py
stscirij: It's possible, but that would mean that data processing would have had to have introduced the error.  I'll put the check in, it's only a few extra lines.

jwst/refpix/reference_pixels.py
```diff
@@ -194,7 +216,7 @@ def transfer_to_model(self, input_model):
 
         input_model: JWST datamodel
 
-        The datamodel the data arrays are to be transferred to
+        The datamodel to which the data arrays are to be transferred
```
hbushouse: 👍

jwst/refpix/reference_pixels.py
```diff
@@ -1327,6 +1365,16 @@ def do_left_right_correction(self, group, refvalues):
...
 
     def do_corrections(self):
+        if self.is_subarray:
+            self.do_subarray_corrections()
+        else:
+            self.do_fullframe_corrections()
+
+    def do_subarray_corrections(self):
+        log.warning("Refpix correction skipped for MIRI subarray")
+        return
```
hbushouse: Does this in turn trigger the setting of the step completion status to "SKIPPED"?

stscirij: Actually, this case is handled higher up now, so this line should never be needed.  The completion status will indeed be set to skipped for MIRI subarrays.

hbushouse: 👍

PR #2215, opened 2018-07-02
Add MIRI apertures for TSO photometry

larrybradley:
The MIRI aperture definitions were provided by the instrument team in #1619.


larrybradley:
Yes, I still need to update this to handle the NIRCam WLP8 case.  It will require a workaround for the background annulus aperture.

larrybradley:
This change was implemented in #2254 using a reference file.

PR #2222, opened 2018-07-06
Fix update_fits_wcs() to work on DrizProductModels

jdavies-st:
Bug found by @gbrammer where the `update_fits_wcs()` method was only doing the updates on `ImageModel`, but of course the output of `ResampleStep` is a `DrizProductModel`. So this code was never applied and the FITS WCS values were never being updated.  Instead the old values from the first image in the input stack were being retained.

The result was the `SourceCatalogStep` which uses the FITS WCS to write out RA/DEC values was writing them out using the wrong WCS.  Correct `xcentroid, ycentroid` values, but wrong sky coordinates for the sources.

@sosey This should help the grism pipeline.

### DIFF
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -213,18 +213,19 @@ class ResampleData:
             self.output_models.append(output_model)
 
     def update_fits_wcs(self, model):
-        """Update FITS WCS keywords of the resampled image."""
-        if isinstance(model, datamodels.ImageModel):
-            transform = model.meta.wcs.forward_transform
-            model.meta.wcsinfo.crpix1 = -transform[0].offset.value + 1
-            model.meta.wcsinfo.crpix2 = -transform[1].offset.value + 1
-            model.meta.wcsinfo.cdelt1 = transform[3].factor.value
-            model.meta.wcsinfo.cdelt2 = transform[4].factor.value
-            model.meta.wcsinfo.ra_ref = transform[6].lon.value
-            model.meta.wcsinfo.dec_ref = transform[6].lat.value
-            model.meta.wcsinfo.crval1 = model.meta.wcsinfo.ra_ref
-            model.meta.wcsinfo.crval2 = model.meta.wcsinfo.dec_ref
-            model.meta.wcsinfo.pc1_1 = transform[2].matrix.value[0][0]
-            model.meta.wcsinfo.pc1_2 = transform[2].matrix.value[0][1]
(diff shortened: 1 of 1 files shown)


jdavies-st:
Jenkins failure is unrelated.

gbrammer:
Great.  Thanks, @jdavies-st!

jwst/resample/resample.py
```diff
@@ -213,18 +213,19 @@ def do_drizzle(self):
...
+        Update FITS WCS keywords of the resampled image.
+        """
+        transform = model.meta.wcs.forward_transform
+        model.meta.wcsinfo.crpix1 = -transform[0].offset.value + 1
+        model.meta.wcsinfo.crpix2 = -transform[1].offset.value + 1
+        model.meta.wcsinfo.cdelt1 = transform[3].factor.value
+        model.meta.wcsinfo.cdelt2 = transform[4].factor.value
+        model.meta.wcsinfo.ra_ref = transform[6].lon.value
+        model.meta.wcsinfo.dec_ref = transform[6].lat.value
+        model.meta.wcsinfo.crval1 = model.meta.wcsinfo.ra_ref
```
hbushouse: Just curious: Why are ``ra_ref`` and ``dec_ref`` being set to new values? Originally they correspond to the sky coords at the location of ``v2_ref`` and ``v3_ref``. Is that still the case here after resetting them?

jdavies-st: Good question.  Basically these keywords are the same CRVAL1 and CRVAL2.  We probably don't need to update them as we are updating the FITS standard CRVAL* keywords.  But it would be odd if they are inconsistent.  There is a v2v3 frame in the output transform, and this RA_REF and DEC_REF do correspond to it.

@nden thoughts?

nden: Do we need to keep these keywords in the header of a resampled image at all (V2/V3_REF, RA/DEC_REF and ROLL_REF)? They don't represent the WCS of a resampled image correctly. If we keep only the FITS WCS keywords we can still create a GWCS object from them if we need to.

hbushouse: I don't think we need them. They are vital to the original (unresampled) data, but once we've got a resampled image that is completely and accurately described by just the standard (FITS) WCS keywords, we might as well get rid of those. It's possible that at least some of them had different values in the multiple input images (v2/v3 would presumably stay the same, but RA/DEC_REF would change within a dither/mosaic pattern) and hence there's no single value that's appropriate anyway. The individual values will be documented in the HDRTAB table by ``model_blender``.

PR #2240, opened 2018-07-13
Re-enable and fix errors with ASDF schema tester

### LINKED ISSUE #2205: enable asdf plugin jwst.transforms
In #2186 two new tets failed on Travis and passed on jenkins. The command to ruin tests on Travis is:
`python setup.py test`, while on Jenkins it is 
`"pytest -r s  -v  --basetemp=./test_results  --junitxml=results.xml"`. It's unclear for now how the plugin changes the testing environment but the problem does not exist if pytest is run directly. More details are in #2186. The plugin was disable in #2186. This issue is to remind us to enable it again.
Perhaps @drdavella will have some insights into the problem when he comes back.

drdavella:
This should resolve #2205 and allow the ASDF schema tester to run successfully under both `pytest` and `python setup.py test`.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -14,9 +14,9 @@ try:
 except (NameError, KeyError):
     pass
 
-#pytest_plugins = [
-    #'asdf.tests.schema_tester'
-#]
+pytest_plugins = [
+    'asdf.tests.schema_tester'
+]
 
 # Add option to run slow tests
 def pytest_addoption(parser):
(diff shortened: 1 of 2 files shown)


nden:
Hm, This was not obvious to me at all.
:+1:

drdavella:
It actually wasn't obvious to me either. After poking around for awhile I eventually remembered similar issues in other packages that were due to the order in which extension types were imported. Basically it turns out that all custom extension types need to be imported and registered as part of an ASDF extension before any other code is executed, or else there can be weird interactions.

I probably should add a comment to the ASDF documentation about this kind of thing...

Issue #1273, opened 2017-10-02
update core schema to reference individual schemas


nden:

melanieclarke:
@emolter - do you know if there is still work that needs to be done to refactor the core schema? If so, this ticket should be moved to stdatamodels.  If not, we should close it.

nden:
This was created a long time ago and was meant to update the core schema to point to individual `keyword_*` schemas, to avoid updates in two places. I'm OK with closing it as it seems everyone is used to the current state.

melanieclarke:
Okay, thanks @nden.  I will close.

PR #2241, opened 2018-07-13
Enable tests for datamodel schemas and fix errors

drdavella:
This enables the ASDF schema tester for schemas found under `jwst/datamodels/schemas`. It also fixes a number of apparent errors that were exposed by these tests.

Since I'm lacking some context about how the datamodels work, someone should check to make sure the changes I made are in fact valid fixes.

This PR is based on #2240, which should be merged first.

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -764,35 +764,47 @@ properties:
           name:
             title: Subarray used
             type: string
-            enum: # Values grouped by instrument:
+            # Values grouped by instrument:
+            anyOf:
               # FGS
-             [8X8, 32X32, 128X128, 2048X64,
-              SUB128CENTER, SUB128DIAGONAL, SUB128LLCORNER, SUB32CENTER,
-              SUB32DIAGONAL, SUB32LLCORNER, SUB8CENTER, SUB8DIAGONAL, SUB8LLCORNER,
+              - enum:
+                 [8X8, 32X32, 128X128, 2048X64, SUB128CENTER, SUB128DIAGONAL,
+                  SUB128LLCORNER, SUB32CENTER, SUB32DIAGONAL, SUB32LLCORNER,
+                  SUB8CENTER, SUB8DIAGONAL, SUB8LLCORNER]
               # MIRI
-              BRIGHTSKY, MASK1065, MASK1140, MASK1550, MASKLYOT, SLITLESSPRISM,
-              SUB128, SUB256, SUB64, SUBPRISM,
+              - enum:
+                 [BRIGHTSKY, MASK1065, MASK1140, MASK1550, MASKLYOT,
+                  SLITLESSPRISM, SUB128, SUB256, SUB64, SUBPRISM]
               # NIRCam
-              SUB160, SUB160P, SUB320, SUB320A335R, SUB320A430R, SUB320ALWB,
-              SUB32TATS, SUB32TATSGRISM, SUB400P, SUB640, SUB640A210R, SUB640ASWB,
-              SUB64FP1A, SUB64FP1B, SUB64P, SUB8FP1A, SUB8FP1B, SUB96DHSPILA,
-              SUB96DHSPILB, SUBFSA210R, SUBFSA335R, SUBFSA430R, SUBFSALWB, SUBFSASWB,
-              SUBGRISM128, SUBGRISM256, SUBGRISM64, SUBNDA210R, SUBNDA335R,
-              SUBNDA430R, SUBNDALWBL, SUBNDALWBS, SUBNDASWBL, SUBNDASWBS,
+              - enum:
+                  [SUB160, SUB160P, SUB320, SUB320A335R, SUB320A430R,
+                   SUB320ALWB, SUB32TATS, SUB32TATSGRISM, SUB400P, SUB640,
+                   SUB640A210R, SUB640ASWB, SUB64FP1A, SUB64FP1B, SUB64P,
+                   SUB8FP1A, SUB8FP1B, SUB96DHSPILA, SUB96DHSPILB, SUBFSA210R,
+                   SUBFSA335R, SUBFSA430R, SUBFSALWB, SUBFSASWB, SUBGRISM128,
+                   SUBGRISM256, SUBGRISM64, SUBNDA210R, SUBNDA335R, SUBNDA430R,
+                   SUBNDALWBL, SUBNDALWBS, SUBNDASWBL, SUBNDASWBS]
               # Obsolete NIRCam
-              MASKA210R, MASKA430R, MASKBLWB, MASKBSWB,
+              - enum:
+                  [MASKA210R, MASKA430R, MASKBLWB, MASKBSWB]
               # NIRISS
-              SUB64, SUB80, SUB128, SUB256, SUBAMPCAL, SUBSTRIP96, SUBSTRIP256,
-              SUBTAAMI, SUBTASOSS, WFSS64C, WFSS64R, WFSS128C, WFSS128R,
+              - enum:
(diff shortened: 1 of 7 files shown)


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -782,7 +782,7 @@ properties:
               # Obsolete NIRCam
               MASKA210R, MASKA430R, MASKBLWB, MASKBSWB,
               # NIRISS
-              SUB64, SUB80, SUB128, SUB256, SUBAMPCAL, SUBSTRIP96, SUBSTRIP256,
+              SUB80, SUBAMPCAL, SUBSTRIP96, SUBSTRIP256,
```
drdavella: Since these are defined as part of an `enum`, it's not valid to have duplicate values here. Ditto on the change below.

jwst/datamodels/schemas/ifucubepars.schema.yaml
```diff
@@ -1,6 +1,6 @@
 title: Default IFU CUBE parameters data model
 allOf:
-- $ref:	referencefile.schema.yaml
+- $ref: referencefile.schema.yaml
```
drdavella: Tabs are not considered valid by the YAML parser. Maybe there should be some kind of git hook to check for this in new commits?

jwst/datamodels/schemas/core.schema.yaml
jdavies-st: Looks like NIRISS and MIRI have some of the same subarray names.  One of the disadvantages of having an `enum` that applies to all instruments, given the way we've organized it in the code here.  I suspect a `set()` of all these would create a valid enum, but I suspect we can't do that in YAML schemas.

philhodge: You could define a section with:
# MIRI, NIRISS
and list the common subarray names there.

jdavies-st: That's a good idea Phil.

jaytmiller: For CRDS I've tried to exploit this by-instrument breakdown and transcribe the enums to add extra checks to CRDS.   So,  where needed,  CRDS verifies (a) that SUBARRAY is definitely defined and (b) that it has instrument-specific values.   CRDS also uses these enums to check rmaps which can be directly edited by people.

Can I count on these by-instrument breakdowns to remain accurate?   Is there some other more definitive source than the CAL schema which breaks value enums down by instrument?

hbushouse: The JWST Keyword Dictionary (json) schemas have keywords like FILTER, PUPIL, READPATT, SUBARRAY, etc. broken down by instrument.

jdavies-st:
For clarification, does the ASDF standard require that enums be a `set` in the YAML schema, or is it possible to have duplicates in the schema and have this enum list or tuple made into a `set` when it is parsed?  I couldn't find documentation on this in either the YAML standard or in the ASDF standard.

If enums are always sets, is there a danger of casting them to a set in the parser in the ASDF code?

jwst/datamodels/schemas/ifucubepars.schema.yaml
jdavies-st: I would hope delinters would catch this, but requires one to actually use them.  One of our TravisCI tests actually runs flake8, but failures there do not prevent the code from being merged or show up as failures.  Maybe it should?

jdavies-st: Here's the flake8 test on TravisCI for the last commit on this PR.

https://travis-ci.org/spacetelescope/jwst/jobs/404451874

drdavella:
@jdavies-st this isn't related to YAML or ASDF directly but instead to JSON Schema. The only thing in question here is whether the schema itself is well-formed. Since the type is given as `enum`, no duplicate values will be tolerated in the schema when the schema itself is validated.

jdavies-st:
The only thing I can find is for JSON schema:

http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.1.2

> **6.1.2. enum**
> The value of this keyword MUST be an array. This array SHOULD have at least one element. Elements in the array SHOULD be unique.
> 
> An instance validates successfully against this keyword if its value is equal to one of the elements in this keyword's array value.

jdavies-st:
And I see now that is a constraint put in by the `jsonschema` implementation in python:
```
In [1]: import jsonschema

In [3]: jsonschema.validate('foo', {"enum": ['foo', 'bar']})

In [5]: jsonschema.validate('foo', {"enum": ['foo', 'bar', 'foo']})
SchemaError                               Traceback (most recent call last)
<ipython-input-5-d615aa762a06> in <module>()
----> 1 jsonschema.validate('foo', {"enum": ['foo', 'bar', 'foo']})

~/anaconda3/envs/nirspec_scsb/lib/python3.6/site-packages/jsonschema/validators.py in validate(instance, schema, cls, *args, **kwargs)
    538     if cls is None:
    539         cls = validator_for(schema)
--> 540     cls.check_schema(schema)
    541     cls(schema, *args, **kwargs).validate(instance)

~/anaconda3/envs/nirspec_scsb/lib/python3.6/site-packages/jsonschema/validators.py in check_schema(cls, schema)
     81         def check_schema(cls, schema):
     82             for error in cls(cls.META_SCHEMA).iter_errors(schema):
---> 83                 raise SchemaError.create_from(error)
     84 
     85         def iter_errors(self, instance, _schema=None):

SchemaError: ['foo', 'bar', 'foo'] has non-unique elements

Failed validating 'uniqueItems' in schema['properties']['enum']:
    {'minItems': 1, 'type': 'array', 'uniqueItems': True}

On instance['enum']:
    ['foo', 'bar', 'foo']
```

drdavella:
@jdavies-st as we discussed offline, it might be better to use unique enums for each instrument to contain the subarrays. I just pushed a commit to this effect.

drdavella:
I just rebased and squashed to clean up the history a little bit.

jdavies-st:
This can be merged I think.

@hbushouse or @nden, is there a reason that the subarray names are in two separate schemas instead of being defined just once in `subarray.schema.yaml` and then included in `core.schema.yaml` via
```
- $ref: subarray.schema.yaml
```
in the correct location?

Currently, all the schemas that reference the subarray schema are for reffiles:
```
$ grep -r subarray.schema.yaml
dark.schema.yaml:- $ref: subarray.schema.yaml
darkMIRI.schema.yaml:- $ref: subarray.schema.yaml
distortion.schema.yaml:- $ref: subarray.schema.yaml
distortion.schema.yaml:- $ref: keyword_psubarray.schema.yaml
extract1dimage.schema.yaml:- $ref: subarray.schema.yaml
flat.schema.yaml:- $ref: subarray.schema.yaml
gain.schema.yaml:- $ref: subarray.schema.yaml
linearity.schema.yaml:- $ref: subarray.schema.yaml
mask.schema.yaml:- $ref: subarray.schema.yaml
nirspec_flat.schema.yaml:- $ref: subarray.schema.yaml
nirspec_quad_flat.schema.yaml:- $ref: subarray.schema.yaml
persat.schema.yaml:- $ref: subarray.schema.yaml
psfmask.schema.yaml:- $ref: subarray.schema.yaml
readnoise.schema.yaml:- $ref: subarray.schema.yaml
saturation.schema.yaml:- $ref: subarray.schema.yaml
specwcs.schema.yaml:- $ref: subarray.schema.yaml
superbias.schema.yaml:- $ref: subarray.schema.yaml
trapdensity.schema.yaml:- $ref: subarray.schema.yaml
trapsfilled.schema.yaml:- $ref: subarray.schema.yaml
wfssbkg.schema.yaml:- $ref: subarray.schema.yaml
```

nden:
The idea was to have the core schema reference individual smaller schemas, like the subarray, so that these schemas can be updated only once. This hasn't been done completely yet, ticket #1273.

Issue #2243, opened 2018-07-16
resampled spectral s2d files have all zeros


jdavies-st:
Currently, `resample_spec` at the end of `calwebb_spec2` produces _s2d files with data arrays that are all zeros for the fixed slit modes.  Our overnight regression tests show this for the following tests:
```
test_nrs_fs_single_spec2
test_nrs_fs_multi_spec2_1
test_nrs_fs_multi_spec2_2
test_nrs_fs_multi_spec2_3
```

jdavies-st:
Resolved by #2245

Issue #2218, opened 2018-07-03
Simplify drizpars reffile schemas


jdavies-st:
Currently there are separate datamodels and schemas for drizpars for different instruments, plus a general one.
```
DrizParsModel
NircamDrizParsModel
MiriImgDrizParsModel
```
We only need the general one, as all the parameters for each instrument can go in the same format table.  Currently the NIRCam drizpars reffile uses `DrizParsModel`, as does the MIRI drizpars reffile.

Remove `NircamDrizParsModel` and `MiriImgDrizParsModel`.

jdavies-st:
Resolved by #2245

Issue #2059, opened 2018-05-21
NIRSpec FS exposure fails in resample_spec


hbushouse:
The data set jw95115001001_02102_00001_nrs1 fails in the `resample_spec` step with the traceback:
```
2018-05-21 16:03:17,003 - stpipe.Spec2Pipeline.photom - INFO - Step photom done
2018-05-21 16:03:17,056 - stpipe.Spec2Pipeline.resample_spec - INFO - Step resample_spec running with args (<MultiSlitModel from jw95115001001_02102_00001_nrs1_rate.fits>,).
2018-05-21 16:03:17,926 - stpipe.Spec2Pipeline.resample_spec - WARNING - /Users/bushouse/anaconda3/envs/myjwst/lib/python3.5/site-packages/jwst-0.9.4a0.dev36-py3.5-macosx-10.7-x86_64.egg/jwst/resample/resample_spec.py:169: RuntimeWarning: Mean of empty slice
  lookup_table = np.nanmean(grid_slit.wavelength, axis=0)

2018-05-21 16:03:18,863 - stpipe.Spec2Pipeline.resample_spec - INFO - Resampling slit S200A1 (31, 429)
2018-05-21 16:03:19,378 - stpipe.Spec2Pipeline.resample_spec - WARNING - /Users/bushouse/anaconda3/envs/myjwst/lib/python3.5/site-packages/scipy/interpolate/interpolate.py:2556: RuntimeWarning: invalid value encountered in greater
  if not np.all(np.diff(p) > 0.):

----------------------------------------------------------------------
ERROR RUNNING STEP 'Spec2Pipeline':
    The points in dimension 0 must be strictly ascending
Traceback (most recent call last):
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python3.5/site-packages/jwst-0.9.4a0.dev36-py3.5-macosx-10.7-x86_64.egg/jwst/stpipe/cmdline.py", line 297, in step_from_cmdline
    step.run(*positional)
  File "/Users/bushouse/anaconda3/envs/myjwst/lib/python3.5/site-packages/jwst-0.9.4a0.dev36-py3.5-macosx-10.7-x86_64.egg/jwst/stpipe/step.py", line 391, in run
    step_result = self.process(*args)
[...]

PR #2245, opened 2018-07-16
Resample refactor

jdavies-st:
Refactor of resample that does the following:

- move drizpars reference file parsing to the top level Step function
- remove the NIRCam-, NIRISS- and MIRI-specific drizpars datamodels and schemas.  They aren't used or needed.  The standard drizpars datamodel works for all instruments.
- remove the NIRCam- and MIRI-specific outlierpars datamodels and schemas.  They aren't used or needed.  The standard outlierpars datamodel works for all instruments.
- change `wht_type` to `weight_type` in resample
- subclass ResampleSpecStep from ResampleStep so they can share attributes and methods that are common.
- fix bugs in parameter passing between top-level ResampleStep and the underlying classes and functions that do that actual work.


jdavies-st:
Resolves #2243.

jdavies-st:
Resolves #2218.

jdavies-st:
Addresses the problems with `resample` in issue #2059, but that issue should remain open as the `pathloss` step now fails for that dataset.

jwst/resample/resample_spec_step.py
```diff
@@ -44,6 +33,13 @@ def process(self, input):
         for reftype in self.reference_file_types:
             ref_filename = self.get_reference_file(input_models[0], reftype)
 
+        if ref_filename != 'N/A':
+            self.log.info('Drizpars reference file: {}'.format(ref_filename))
+            kwargs = self.get_drizpars(ref_filename, input_models)
+        else:
+            # Deal with NIRSpec which currently has no default drizpars reffile
```
hbushouse: I assume this is a temporary workaround until someone creates a drizpars reffile for NIRSpec?

jwst/resample/resample_utils.py
```diff
@@ -171,21 +171,19 @@ def _reproject(x, y):
...
-    if wht_type.lower()[:3] == 'err':
+    if weight_type == 'err':
         err_model = np.nan_to_num(model.err)
         inwht = (exptime / err_model)**2 * dqmask
         log.debug("DEBUG weight mask: {} {}".format(type(inwht), np.sum(inwht)))
-    # elif wht_type == 'IVM':
+    # elif weight_type == 'ivm':
     #     _inwht = img.buildIVMmask(chip._chip,dqarr,pix_ratio)
-    elif wht_type.lower()[:3] == 'exp':
+    elif weight_type == 'exp':
```
hbushouse: Isn't the value that gets passed in for exposure time weighting "exptime", as opposed to "exp"? Or is that reset somewhere in the param parsing?

jdavies-st: Ah, thanks for catching that!  That's a bug.

jwst/resample/resample_spec_step.py
jdavies-st: Yes, it also is a general "if there is no reffile" workaround for any undefined modes that don't have a reffile.  NIRSpec is the one that I ran into that did not have a reffile, but perhaps there are others.  NIRCam, MIRI and NIRISS have reffiles right now.  So perhaps FGS needs one too?  Not sure they'll be resampling FGS science data.

jwst/resample/resample_utils.py
hbushouse: minion_what

PR #2247, opened 2018-07-16
RSCD update to array-based math

hbushouse:
Update the core slope-fitting function used in the RSCD step to use array-based calculations, instead of looping over pixels. This speeds up the calculations by about 20x.


hbushouse:
@jemorrison checked the new results against her existing test files and found good agreement. So the revised algorithm looks good. Merging.

PR #2254, opened 2018-07-17
Use a reference file for tso_photometry

philhodge:
The `tso_photometry` step now uses a reference file to obtain the radii of the source aperture and background annulus.  Documentation was added.
See issue #1619.


docs/jwst/tso_photometry/arguments.rst
```diff
@@ -0,0 +1,10 @@
+Step Arguments
+==============
+
+The tso_photometry step has one step-specific argument:
+
+*  ``--save_catalog``
+
+If ``save_catalog`` is set to True (the default is False),
+the output table of times and count rates will be written to an ecsv file
+with suffix "phot".
```
hbushouse: Perhaps highlight that this argument is really only needed when you're running the step standalone, because by default when it's run as part of the calwebb_tso3 pipeline the catalog will always be saved.

larrybradley:
Thanks, @philhodge!

PR #2259, opened 2018-07-19
Update Readme with CRDS configuration

nden:
Adds instructions on how to set up CRDS outside the STScI network.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,11 @@ pulling the latest version of `master` from the Github repository inside the `jw
     git pull origin master
     python setup.py develop
 
+To run the pipeline outside the STScI network, CRDS must be configured by setting two environment variables:
+
+    export CRDS_PATH=$HOME/crds_cache
+    export CRDS_SERVER_URL=https://jwst-crds.stsci.edu
+
 
 Documentation
 -------------


jaytmiller:
This looks good but there may be a few other complications:

1. User's should be forewarned that the first run CRDS will download required rules and references on demand and it may take a long time.   For better feedback on downloads  do:

$ export CRDS_VERBOSITY=50

2. Running multiple concurrent processes to synchronize the cache dynamically under pytest sounds **incredibly dicey**.   I've never done it,  **I'm in the process of trying it**.   Potentially this is also needed:

$ export CRDS_LOCKING_MODE=filelock

which uses a file based process locking mechanism to prevent concurrent updates to the same files.   I'm vaguely worried it too will not be sufficient because I wrote it in the context of strun etc,  not arbitrary parallelism.   

Your first cut looks good but lets slow down before merging it.

jaytmiller:
I tried this out on my laptop after rsync'ing bigdata.   I pummeled it with:

pytest -n 8 --bigdata .

and the default CRDS configuration using multiprocessing locks.

There was at least one CRDS concurrency issue and a number of failures (not sure how well this runs with /grp/crds/cache) but at the end of the run the CRDS cache looked fully correct with 16G of demand-based downloaded references.   

The one glitch I'm sure occurred was reading an instrument mapping (.imap),  apparently from a second process before the first had downloaded or fully downloaded it.   I'll need to do more analysis to determine how that was possible.

Thinking more I'm not sure either CRDS_LOCKING_MODE or CRDS_VERBOSITY is worth mentioning.  The former will really only help with multiple pytest runs in different windows and the latter is pretty much invisible due to pytest anyway.

Without more debug,  for I'd just recommend either (a) don't use -n 8 on your first "cache sync" run or (b) do one run with -n 8,  expect possible errors,  and repeat with a second run expecting it to be clean.   I will play around with this some more this weekend and next week.   I'm now set to do a download free run from home and see what happens.

jdavies-st:
Thanks for testing this out!  I usually run the unit tests locally at the top level with `-n 24`, but if I'm running the nightly regression tests locally, I don't do them in parallel, because several of the tests consume lots of memory.  Normally on Jenkins the regression tests take ~3 hours.  Did you notice a significant speedup running them with `-n 8`?

jaytmiller:
I think this is ready to merge.    

Summing up follow-on discussion and explained testing anomalies:

1. Since the nightlies aren't run multi-threaded anyway,   CRDS cache locking is moot.   I think pytest is using threading so the default CRDS multiprocessing cache locks do not work.   CRDS_LOCKING_MODE=filelock should work but I did not test for this.    Running pytests single threaded eliminated caching errors.

2. In a number of offsite tests I had universally successful dynamic cache syncs of around 16G required to run the nightlies.   So even the thread related errors were temporary.

Running directly from /grp/crds/cache with no downloads my OS-X laptop still had nightly regression errors,  but I think they're innocuous:

3. There were errors due to insufficient file handles.   datamodels file handle leaks are exposed under OS-X because it has a default of 256 available handles vs. the Linux default on jwcalibdev of 1024.  My tests were leaking around ~750-800 handles so open()'s > 256 failed.  See bash ulimit -n and Linux lsof -p for help debugging handles.  @jdavies-st wrote a related ticket against datamodels.

4. I encountered precision errors in the fitsdiffs.   After discussion with @jdavies-st we concluded that those are probably acceptable platform or conda stack induced differences.

Issue #1078, opened 2017-07-21
Background_sub should do sigma-clipping when computing average


hbushouse:
The background subtraction step used in the calspec2 (and eventually calimage2) pipeline to subtract dedicated background exposures from science target exposures currently produces a combined background image (when there are multiple background inputs) by computing a simple mean of the inputs. The JCCWG would like to also have sigma-clipping applied, in an attempt to remove any sources that might be present in some of the input images.

hbushouse:
@karllark @drlaw1558 The CALWG description indicates:
```
The combined background image will be constructed by averaging the data from all background exposures in detector space and will use a sigma-clipping algorithm to remove outliers.
```
Any thoughts/desires as to the exact details of this? Should we in fact use a straight average/mean as opposed to, say, median? Should the choice of mean vs median vary as a function of the number of images being combined? Should there be a single iteration of sigma-clipping or multiple? What should the default sigma value be?

philhodge:
Here's an alternative that looks good to me.

    limits = np.percentile(x, (lowlim, highlim))
    mask = np.logical_and(x >= limits[0], x <= limits[1])
    mean_value = x[mask].mean(dtype=np.float)

`lowlim` and `highlim` are in percent, e.g. 25 and 75 respectively.
Sigma clipping would do a poor job for an array such as this:

    [1, 3, 5, 7, 9, 20, 40, 60, 80, 91, 93, 95, 97, 99]

hbushouse:
Interesting. I guess it depends a lot on what kinds of variation we're expecting within a given pixel for a series of background images. If it's the case that the actual background signal does not vary all that much from image to image, such that we're mainly interested in clipping out sources, and those sources only show up in a given pixel in a small number (1 or 2) of images because of dithering, then I would think we'd usually be dealing with fairly normal looking (not in the official statistical sense of "normal") distribution of values containing a small number of outliers. E.g. something more like:
```
[2, 5, 42, 3, 4, 6, 3, 50]
```
But even in this type of case we might be dealing with some values that are the low-level wings of a source that's sitting a couple pixels away, and regular sigma clipping is not going to be very good at picking out those values that are elevated only a small amount. That's where something like a percentage clip would be useful (e.g. always clip out anything that's >80%).

philhodge:
[...]

PR #2270, opened 2018-07-20
Add sigma-clipping to background subtraction

hbushouse:
An initial attempt to add sigma-clipping to the process of creating an average background image in the background subtraction step, as requested in #1078. This uses the astropy.stats.sigma_clip function to do the clipping, with default values of sigma=3 and maxiters=None (which means to iterate until convergence is achieved). The sigma and maxiter arguments have been added as step arguments and the docs have been updated accordingly.

It's likely that we'll want to add some more fine tuning for handling situations with small numbers of background images (e.g. perhaps fewer iterations, different sigma, etc.).


hbushouse:
The latest commit contains lots of updates to the docs to fix formatting, add references/links, and add a lot more detailed info about the reference files. It also updates the list of available data models in the datamodels docs. Finally, it removes the `maxiters` argument from the `background.cfg` file, because ConfigObj doesn't seem to be able handle the case of an integer argument having a default of `None`.

Issue #2237, opened 2018-07-13
Replace and redefine __hasattr__


bernie-simon:
I'm making a small change to datamodels to remove what I always thought was a wart, but have resisted changing in the past because I didn't want to modify the interface. But I have searched the source code of jwst and found that \_\_hasattr\_\_ is never called,  so I thought it's better to fix it now before it's too late.

First, I am renaming \_\_hasattr\_\_ to hasattr. Perhaps it was named that on the belief that Python has such a double underscore method, but it does not. The hasattr function calls the \_\_getattr\_\_ double underscore method inside a try block.  And you should avoid using the hasattr function for that reason, because \_\_getattr\_\_ creates the attribute you are testing for if the attribute has a default. So using the hasattr function is like striking a match to see if it works.  So instead of using the hasattr function, you should use the hasattr method. That is, use

    model.hasattr('zeroframe')

and not

    hasattr(model, 'zeroframe')

Second, I am going to modify the test hasattr performs. Currently it checks to see if the attribute is a key in the dictionary or if it's present in the schema. In the typical case I don't see why you would want the second part of the test. You want to know instead if the attribute is currently present in the model.

Finally, to test for the presence of a metadata item, use the usual dotted notation

    model.meta.hasattr('telescope')

or

    model.meta.target.hasattr('catalog_name')

hbushouse:
Fixed by #2275

PR #2275, opened 2018-07-23
Rewrote the hasattr method in Datamodels

bernie-simon:
The `__hasattr__` method has been replaced by `hasattr`. It tests to see
if a node has a specified attribute, much like the hasattr function in
Python. However, the hasattr method does not create the attribute
while the hasatter function does. This is a peculiarity of the
implementation of Datamodels nodes and not a general issue for Python.

I added a new test for hasattr and went through the datamodels code to
replace the use of the hasattr function with the new method where that
was appropriate.


hbushouse:
Fixes #2237

PR #2284, opened 2018-07-25
Fix handling of default data array name

bernie-simon:
Datamodels normally store their most significant data in an array
named 'data'. Datamodels allows this to be changed in a subclass by
overriding the method get_primary_array_name. The base class returns
the name 'data'. I've modified it so that it now checks the schema and
returns the empty string if it is not found. I used the empty string
rather than None in case a user feeds the result directly into
getattr. A value of None will throw an error and both are falsey. The
new code prevents datamodels that contain tables from reporting that
they have a data array.

I've modified the shape method/property to test the result of
get_primary_array_name, which restores the behavior I accidentally
changed when adding the hasattr method.

I've modified the data getter/setter in NDModel to call
get_primary_array_name. This fixes a bug where the code assumed the
name was 'data'. NDModel is a mixin class that provides compatibility
with astropy's NDData.

I've added methods overriding get_primary_array_name to AMilgModel and
LinearityModel, since, according to the schema, these two classes,
along with MaskModel, have data arrays with names other than 'data'.

### DIFF
--- a/jwst/datamodels/amilg.py
+++ b/jwst/datamodels/amilg.py
@@ -43,3 +43,12 @@ class AmiLgModel(model_base.DataModel):
 
         if solns_table is not None:
             self.solns_table = solns_table
+
+    def get_primary_array_name(self):
+        """
+        Returns the name "primary" array for this model, which
+        controls the size of other arrays that are implicitly created.
+        This is intended to be overridden in the subclasses if the
+        primary array's name is not "data".
+        """
+        return 'fit_image'
--- a/jwst/datamodels/linearity.py
+++ b/jwst/datamodels/linearity.py
@@ -42,3 +42,12 @@ class LinearityModel(ReferenceFileModel):
 
         # Implicitly create arrays
         self.dq = self.dq
+
+    def get_primary_array_name(self):
+        """
+        Returns the name "primary" array for this model, which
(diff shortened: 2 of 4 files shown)


stscicrawford:
Would it be possible to fix this the other way?  ie to have those two types of files still using .data for their primary data holder?

bernie-simon:
The method get_primary_array_name was written to handle the case where the name differs. The mask model already utilizes it and sets the primary array to "dq." It would be confusing to rename it for the mask model and I assumed whoever wrote the other two models had equally good reasons for not using "data".

stscicrawford:
Thanks for the explanation.   Should `NDModel` have an abstract`get_primary_array_name` function?   As it is, without that function, anything that is based on it would fail if it didn't implement it (which isn't currently a problem, but would be in the future).

bernie-simon:
The purpose of the NDModel code is to implement astropy's NDData interface, so that astropy code written to that interface can access datamodels. The get_primary_array_name is meant to be implemented by persons writing new datamodels subclasses. It's not something persons writing astronomical software would be concerned about, unless they were adding a subclass. Once this fix is installed the NDModel interface will call get_primary_array, so that it accesses the primary array in a transparent manner.

Issue #1756, opened 2018-02-19
Extract specific orders for WFSS data


sosey:
See also https://jira.stsci.edu/browse/JP-204

From Kevin Volk:

> For NIRISS WFSS we want the first order spectra extracted for all filters, and also the second order spectrum extracted for the F090W filter only. The second order spectra for the other filters F115W to F200W are quite faint and assumed to be not worth extracting or calibrating. Similarly, the order -1 and order 3 spectra are generally rather faint and there was no plan to extract and calibrate these orders routinely in the pipeline, although for bright sources one might want to extract these orders, and also for bright sources we need to know where these orders are to model spectral overlap. For a really bright source some of the ground test data shows orders 4 and higher for some cases where were really blasting the array with a very bright point source, but these grow progressively fainter and for science exposures it is unlikely that something that bright will ever be in the field of view. The “extra” orders would not be processed in the pipeline.

There are a few ways we could go about implementing this, a combination of methods might prove the most flexible:

**1. add information on orders to be extracted to the `specwcs` reference file, this reference file is instrument specific.** 

This reference file is also specific to the grism and the pupil/filter, so a new data member could effectively just be a list of orders to extract by default. 

> `instrument:` {detector: NIS, filter: GR150R, name: NIRISS, pupil: F200W}`

The orders that are available (have defined polynomials) in the reference file itself are listed under the `orders` member:

> `orders: [1, 0, 2, 3, -1]`

This method would require updates to the class code to pull in what was specified in the reference file

**2.  Add parameter to the processing class for assign_wcs that specifies what orders and wavelengths should be used.**

- `assign_wcs` takes in the `specwcs` and `wavelengthrange` reference files
- `extract_2d` currently only looks for the `wavelengthrange` reference file

`extract_2d` requires the WCS object already exists, the WCS object already contains information on the orders that are available, this could be modified at the start of `assign_wcs` to represent only the requested orders, not all available orders.

-- At the pipeline level -- 

**3. Add parameters to the cfgspec file for the step that specifies what orders per filter should be extracted.**

We would need one that accepts a dictionary keyed by filter, or a list of lists, or for the caller to only provide data (single files or an asn of files) that only pertain to one filter+grism combination.
I'm not sure I see more that string, bool, int, and list as available datatypes in the configobj specification for the cfg files in stpipe (@stscieisenhamer?) 

This method, when considered by itself, would require a change to the input step parameters so that the information could be passed down from the step level. 

[...]

PR #2286, opened 2018-07-25
Implements NRC_TSGRISM with set source location and extraction options

### LINKED ISSUE #1235: NRC_TSGRISM source object location
The TSGRISM mode was implemented in assign_wcs using the the same general models as NRC_GRISM. However, this mode will always place the source object in the same detector position for each exposure. Since we know where the source should be, if we add this information to the meta information in the data model, we can update the gwcs models to include this information for the users.  #1232 includes the general implementation.

### LINKED ISSUE #1710: Implement NIRCAM TSO grism extraction from CubeModel in extract 2D
In TSO mode, the data arrays are 3D, where each plane of the cube corresponds to the countrate image for a given integration within the exposure. I think we've already enabled the various calwebb_spec2 steps to handle this form of data (a "CubeModel", as opposed to a more normal 2D "ImageModel"). So the thing to ensure is that extract_2d does an extraction from the input 3D cube, producing an output 3D cube that is simply smaller in the first two (x, y) dimensions, but still has all of the axis 3 planes. The extract_1d step then knows to produce a 1D spectrum from each plane of the 3D cube that it's given.

sosey:
***DO NOT MERGE THIS PR UNTIL AFTER SNAPSHOT FOR 0.10 IS TAKEN***
closes #1710,  closes #1235
related to #1756

This PR accomplishes several things:

* it reworks` assign_wcs` for the `NRC_TSGRISM` mode to always use the values for `CRPIX1` and `CRPIX2` as the source location. The new way to call the WCS object is:  `wcs(x, y, order)` which prints ra, dec, wavelength, order, where ra,dec is `CRVAL1, CRVAL2`.  The inverse transform is `wcs(wavelength, order)` which prints out x, y, order  using the same ra, dec values. 

* it implements `extract_2d` for the first time for this mode. Since there is a single source location for all exposures, no catalog of sources are needed. Hence, this does not use the same methods that the WFSS modes use to do extraction and is much simpler, and much faster. The input model is a `jwst.datamodels.CubeModel`, and depending on the number of orders to be extracted for the filter in process, the output will either be a `jwst.datamodel.SlitModel` or a `jwst.datamodel.MultiSlitModel` with 3D data members.  Where, each slit is valid for 1 order, but may contain n integrations. The output wcs is also altered here to take into account that we now know the order of the data it's attached to, so the calling structure is again simplified to: `wcs(x, y)` which prints wavelength, order... but I just realized I could remove the order and go straight from (x,y) <-> (wavelength) at this point....another commit might be added tomorrow

* extract_orders.... this is also a new feature, first meant for use with the WFSS modes, but since I have the updated `wavelengthrange` reference file all made and I was writing most of this step from scratch for TSO I added those updates in this PR as well. For `NRC_TSGRISM` if nothing is passed for the parameter `extract_orders` in `extract_tso_object()` then it will use the ones specified in the reference file parameter of the same name. Currently, for this mode explicitly, `extract_2d` has the default set to only extract the first order for filters used with this mode.

*reviewers* should note that the reference file update changes the calls for the WFSS mode as well, those files will be updated in a corollary PR that I will merge locally with this branch before submitting. We should coordinate the acceptance of both these PRs with each other before delivery of the reference file to CRDS so that CRDS can validate the new datamodels with the appropriate code base.

### DIFF
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -298,7 +298,7 @@ def not_implemented_mode(input_model, ref):
     return None
 
 
-def get_object_info(catalog_name=''):
+def get_object_info(catalog_name=None):
     """Return a list of SkyObjects from the direct image
 
     the source_catalog step are read into a list of  SkyObjects
@@ -318,6 +318,8 @@ def get_object_info(catalog_name=''):
     -----
 
     """
+    if catalog_name is None:
+        raise TypeError("Expected name of the catalog file")
     objects = []
     catalog = QTable.read(catalog_name, format='ascii.ecsv')
 
@@ -359,7 +361,8 @@ def get_object_info(catalog_name=''):
     return objects
 
 
-def create_grism_bbox(input_model, reference_files, mmag_extract=99.0):
+def create_grism_bbox(input_model, reference_files,
+                      mmag_extract=99.0):
     """Create bounding boxes for each object in the catalog
 
     The sky coordinates in the catalog image are first related
@@ -379,7 +382,6 @@ def create_grism_bbox(input_model, reference_files, mmag_extract=99.0):
     mmag_extract : float
         The faintest magnitude to extract from the catalog
 
-
     Returns
     -------
     A list of GrismObject(s) for every source in the catalog
@@ -388,17 +390,17 @@ def create_grism_bbox(input_model, reference_files, mmag_extract=99.0):
 
     Notes
     -----
-    The wavelengthrange reference file is used to govern the extent of the bounding box
-    for each object. The name of the catalog has been stored in the input models meta
-    information under the source_catalog key.
+    The wavelengthrange reference file is used to govern the extent of the
+    bounding box for each object. The name of the catalog has been stored
+    in the input models meta information under the source_catalog key.
 
-    It's left to the calling routine to cut the bounding boxes at the extent of the
-    detector (for example, extract 2d would only extract the on-detector portion of
-    the bounding box)
+    It's left to the calling routine to cut the bounding boxes at the extent
+    of the detector (for example, extract 2d would only extract the on-detector
+    portion of the bounding box)
 
-    Bounding box dispersion direction is dependent on the filter and module for NIRCAM
-    and changes for GRISMR, but is consistent for GRISMC,
-    see https://jwst-docs.stsci.edu/display/JTI/NIRCam+Wide+Field+Slitless+Spectroscopy
+    Bounding box dispersion direction is dependent on the filter and module for
+    NIRCAM and changes for GRISMR, but is consistent for GRISMC, see
+    https://jwst-docs.stsci.edu/display/JTI/NIRCam+Wide+Field+Slitless+Spectroscopy
 
     NIRISS only has one detector, but GRISMC disperses along rows and GRISMR disperses
     along columns.
@@ -433,9 +435,6 @@ def create_grism_bbox(input_model, reference_files, mmag_extract=99.0):
     xsize = input_model.meta.subarray.xsize
     ysize = input_model.meta.subarray.ysize
 
-    # extract the catalog objects
-    if input_model.meta.source_catalog.filename is None:
-        raise ValueError("No source catalog listed in datamodel")
     skyobject_list = get_object_info(input_model.meta.source_catalog.filename)
 
     # get the imaging transform to record the center of the object in the image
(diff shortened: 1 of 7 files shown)


sosey:
@drdavella if you have the time, the test is failing on schema validation, but it validates for me locally. Can you give a look and see what I might have missed? This PR includes an updated schema for this file https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/schemas/wavelengthrange.schema.yaml

jwst/assign_wcs/util.py
```diff
@@ -227,7 +227,7 @@ def is_fits(input):
         isfits = True in [input.endswith(l) for l in names]
 
     # if input is a fits file determine what kind of fits it is
-    #waiver fits len(shape) == 3
+    # waiver fits len(shape) == 3
```
hbushouse: Aaack! I don't think we need to worry about wavered FITS format anymore ...

jwst/extract_2d/extract_2d.py
```diff
@@ -27,29 +28,49 @@ def extract2d(input_model, slit_name=None, apply_wavecorr=False, reference_files
         Reference files.
     grism_objects : list
         A list of grism objects.
+    extract_height: int
+        Cross-dispersion extraction height to use for time seriesgrisms.
```
hbushouse: missing a white space in "seriesgrisms"

jwst/extract_2d/grisms.py
```diff
@@ -4,20 +4,213 @@
...
+        extract_orders = ref_extract_orders
+
+    available_orders = [x[1] for x in extract_orders if x[0] == input_model.meta.instrument.filter].pop()
+    if len(available_orders) > 1:
+        log.info("Extracting into a MultiSlitModel")
+        output_model = datamodels.MultiSlitModel()
+        output_model.update(input_model)
+    else:
+        log.info("Extracting into a SlitModel")
+        output_model = datamodels.SlitModel()
```
hbushouse: Is this the way it's done for NRS_BRIGHTOBJ mode too - the output is a SlitModel, as opposed to a MultiSlitModel that contains just 1 Slit instance? I just want to make sure we're consistent, so that extract_1d is able to handle both NRS_BRIGHTOBJ and NRC_TSGRISM the same way.

philhodge: `extract_1d` tests for CubeModel (the old model for NRS_BRIGHTOBJ) and SlitModel.  So NRC_TSGRISM should be handled by the same section, yes.

jwst/assign_wcs/util.py
sosey: :rofl:

jwst/extract_2d/grisms.py
sosey: that was my understanding. Hold off on reviews because after todays discussion, almost all of this code is changing. The only output will be a SlitModel with a single data cube.

sosey:
The current branch contains the most flexible TSO code. It only extracts 64 pixels in the cross-dispersion direction, but it adds a wcs bounding box for the first order, which makes that the only valid region in the image for the wcs. The full array is extracted in the dispersion direction. It also only creates one `SlitModel` cube, for the filters that NRC_TSGRISM uses. All of these things can be overriden by users.  Still working on the wavelengthrange reference file, so don't merge this PR without letting me know.

jwst/assign_wcs/util.py
sosey: Just FYI, I'm not changing this part of the code right now, it was copied in for use in some other part of the pipeline (though I don't think the waivered fits logic is being used)

sosey:
This should be ready to go, the reference files are created and checked. I ended up splitting to two wavelengthrange reference files, one for WFSS and one for TSGRISM in order to accommodate the different default orders that each team wants extracted. The updated reference files and delivery text are explained in https://jira.stsci.edu/browse/CRDS-232.  I'll add documentation updates in a separate PR.

sosey:
uh, a few people have reviewed, but I need someone to hit the official okay so we can merge this

stscicrawford:
There is still a small typo that @hbushouse raised about 'seriesgrism'.    

Can you also add a change log entry for this? 

This will also need both unit and regression tests added or updated.   I don't think that should hold up the pull request, but can you open an issue with a list of the tests that should be added?  

Thanks @sosey

hbushouse:
Looks like that typo that I mentioned was fixed in a later commit. So I approve.

stscicrawford:
I've approved as well, so merge when you are ready @sosey

PR #2289, opened 2018-07-26
Ramp fit fixes for ramps that are all NaNs

dmggh:
Corrected handling of ramps whose first differences are all NaNs (such as ramps with all groups saturated), segregated code to calculate quantities needed for the optional output product only if the save_opt is selected, updated variance calculations so that their values in the optional output will be 0 for non-existing segments, added code to remove segments having a single group if the ramp has other segments.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -29,12 +29,9 @@ log.setLevel(logging.DEBUG)
 
 BUFSIZE = 1024 * 30000  # 30Mb cache size for data section
 
-# Replace zero or negative variances with this.
-LARGE_VARIANCE = 1.e8
-
 
 def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
-              algorithm, weighting):
+             algorithm, weighting):
     """
     Extended Summary
     ----------------
@@ -43,7 +40,7 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
     cosmic rays) of the pixel's ramp divided by the effective integration time.
     The weighting parameter must currently be set to 'optim', to use the optimal
     weighting (paper by Fixsen, ref. TBA) will be used in the fitting; this is
-    currently the only supported weighting scheme.   
+    currently the only supported weighting scheme.
 
     Parameters
     ----------
@@ -136,7 +133,7 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
         gain for all pixels
 
     weighting: string
-        'optimal' specifies that optimal weighting should be used; currently 
+        'optimal' specifies that optimal weighting should be used; currently
         the only weighting supported.
 
     Returns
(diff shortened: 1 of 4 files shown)


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -2132,7 +2126,7 @@ def fit_2_group(slope_s, intercept_s, variance_s, sig_intercept_s,
     # length) in this function, but are being included here to explicitly
     # cover all possibilities for pixels in datasets with ngroups=2. Will
     # later consider refactoring.
-    wh_sat1 = np.where((mask_2d[:, :].sum(axis=0) == 1) & mask_2d[0, :] == True)
+    wh_sat1 = np.where((mask_2d[:, :].sum(axis=0) == 1) & mask_2d[0, :] is True)
```
philhodge: According to PEP8, "stuff is True" is even worse than "stuff == True".

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -2379,7 +2373,7 @@ def calc_opt_sums(rn_sect, gain_sect, data_masked, mask_2d, xvalues, good_pix):
         data_masked[:, wh_m2d_f] = np.roll(data_masked[:, wh_m2d_f], -1, axis=0)
         c_mask_2d[:, wh_m2d_f] = np.roll(c_mask_2d[:, wh_m2d_f], -1, axis=0)
         xvalues[:, wh_m2d_f] = np.roll(xvalues[:, wh_m2d_f], -1, axis=0)
-        wh_m2d_f = (c_mask_2d[0, :] == False)
+        wh_m2d_f = (c_mask_2d[0, :] == False) # '==' instead of 'is' for sum()
```
philhodge: PEP8 again.  You can use `np.logical_not(c_mask_2d[0, :])`.

jwst/ramp_fitting/utils.py
```diff
@@ -401,39 +451,32 @@ def alloc_arrays(n_int, imshape, max_seg):
...
-    var_p3 = np.zeros((n_int,) + imshape, dtype=np.float32)
-    var_r3 = var_p3.copy() * 0.
-    var_both3 = var_r3.copy() * 0.
+    # Initialize variances so that non-existing ramps and segments will have
+    #   negligible contributions
+    # Integration-specific:
+    var_p3 = np.zeros((n_int,) + imshape, dtype=np.float32) + LARGE_VARIANCE
+    var_r3 = var_p3.copy()
+    var_both3 = var_p3.copy()
     s_inv_var_p3 = var_p3.copy() * 0.
```
philhodge: Instead of copying and multiplying, you could use `s_inv_var_p3 = np.zeros_like(var_p3)`.  I doubt that that would make any noticeable difference, but in principle it should be quicker and use less temporary memory.

jwst/ramp_fitting/utils.py
```diff
@@ -591,8 +633,8 @@ def calc_pedestal(num_int, slope_int, firstf_int, dq_cube, nframes, groupgap,
     firstf_int: float, 3D array
         integration-specific first frame array
 
-    dq_cube: int, 4D array
-        hypercube of DQ array
+    dq_first: int, 3D array
```
philhodge: This array is only 2D.

hbushouse:
Changes contained here caused many of the regression tests that run the `ramp_fit` step to take hours to complete. One or two isolated comparisons showed an increase in processing time of factors of 6-7 (e.g. 5 minutes to 35 minutes). Something has obviously caused a big decrease in efficiency. The changes were reverted in #2301 in order to restore normal processing capabilities.

PR #2296, opened 2018-07-27
Fix timeconverson for change to ephemeris fields

### LINKED ISSUE #2238: timeconversion needs update for change to ephemeris fields
The jwst ephemeris used in the ``timeconversion`` pkg has had a new table column added to it by SDP in order to fix another issue, which is now causing an error, because we blindly load all entries from the whole table. Mike Swam has developed a potential solution, which they've tested in the SDP dev environment:
```
We came up with this work-around in our JWST-DEV copy of SCSB’s timeconversion
pkg (in the __init__.py file):
 
    conn = pymssql.connect(server=eserver, database=edb)
    cur = conn.cursor()
    #cur.execute('select * from predictephemeris')
    #TEMP HACK by SDP to use new table format
    cur.execute('select ephem_time,jwst_x,jwst_y,jwst_z,jwst_dx,jwst_dy,jwst_dz from predictephemeris')
 
By using an explicit set of fields in the DB select statement, we avoid picking up the new 
time-string field that was added to that table to help with another issue, and restore the 
behavior that the SCSB code was expecting under the old table field layout via the 
“select *” call.  I’m sending this in case you want to use it, as is, to deliver back to us in a 
future release, or use it as a reference in developing your own updates to deal with the 
change in table format.
```

hbushouse:
Fix the problem in `timeconversion` that was caused by a recent addition of a new field to the ephemeris by retrieving only the fields needed. Not sure if the jwst velocity fields are actually needed in the script, but they were retrieved before, so keeping them for now.

Fixes #2238.

### DIFF
--- a/jwst/timeconversion/__init__.py
+++ b/jwst/timeconversion/__init__.py
@@ -238,7 +238,7 @@ def get_jwst_ephemeris():
     )
     conn = pymssql.connect(server=eserver, database=edb)
     cur = conn.cursor()
-    cur.execute('select * from predictephemeris')
+    cur.execute('select ephem_time,jwst_x,jwst_y,jwst_z,jwst_dx,jwst_dy,jwst_dz from predictephemeris')
     etab = np.array(cur.fetchall())
     return etab
 


perrygreenfield:
Just to clarify, the problem occurred because the previous code retrieved all columns and then converted them to an array, spoiled since they put in a new column thus screwing up the assumed meaning of the previous array?

philhodge:
It looks to me as if the new field could only be a problem if it was inserted within what was the first four columns, i.e. time, x, y, z.

jwst/timeconversion/__init__.py
```diff
@@ -238,7 +238,7 @@ def get_jwst_ephemeris():
     )
     conn = pymssql.connect(server=eserver, database=edb)
     cur = conn.cursor()
-    cur.execute('select * from predictephemeris')
+    cur.execute('select ephem_time,jwst_x,jwst_y,jwst_z,jwst_dx,jwst_dy,jwst_dz from predictephemeris')
```
philhodge: The velocities are not needed, but it shouldn't hurt to retrieve them.

hbushouse:
@perrygreenfield That's correct. Apparently the presence of the addition column of data in the returned array caused problems somewhere further down within the `timeconversion` calculations (probably trying to use specific array indices or something, which are now different?).

perrygreenfield:
My screwup for not making it invariant to additions.

PR #2298, opened 2018-07-27
Added annotation to embedded ASDF exception message

jaytmiller:
This enables CRDS certify to issue an error message like this:

     Validation error : JWST Data Models: ERROR loading embedded ASDF: newline not found

instead of a less helpful message like this:

     Validation error : JWST Data Models: newline not found

Certify options for trapping this robustly are pretty limited, this particular
pull was inspired by a corrupt .fits with bad/missing ASDF extension which
bubbles up to CRDS as "ValueError('newline not found')".

IMHO the datamodels are probably the right place to add explanation since
they're doing the funky embedded ASDF thing.   I guess an alternative would
be to improve the ASDF embed functions.

### DIFF
--- a/jwst/datamodels/fits_support.py
+++ b/jwst/datamodels/fits_support.py
@@ -537,7 +537,10 @@ def _load_history(hdulist, tree):
 
 
 def from_fits(hdulist, schema, extensions, context):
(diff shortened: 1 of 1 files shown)


jaytmiller:
NOT urgent,  hold for after end-of-July.

Issue #735, opened 2017-03-21
Level2: Implement background


stscieisenhamer:
# Issue
Two of three cases of backgrounds need be handled:
- MSA-specific backgrounds: No association needed
- Nodding
   These are multiple exposures that have the science target(s) moving a small distance within the available slit(s). The multiple exposures will be subtracted from one another in order to remove background signal. This can be applied to several observing modes, including MIRI LRS fixed-slit, NIRSpec fixed-slit, and NIRSpec MSA (and rumor has it from Jeff Valenti that it might also be used for NIRSpec IFU observations of smallish sources). These need association information, because they involve the use of multiple exposures, and hence we need to know which exposures to group together into an association. Because all of the exposures contain the science target(s), none of them is designated as being a background exposure. They are all science exposures, containing spectra of the science target(s).
   - Spectroscopic cases only
    - All exposures are listed as science, none are labeled as background
    - As far as we know we have all the information we need to distinguish source from background, but that may change once we get real examples out of the OTB simulations.
- Chopping = Off-target backgrounds
   These are multiple exposures that have the science target in some of them and regions of blank sky in others. These are the flavor being discussed in the recent email thread. The regions of blank sky will be designated in the APT by creating virtual targets for them and taking observations of them. The APT user will also indicate which of these background targets should be used with each science target. Associations are needed to tell us which exposures (science and background) get grouped together and which of them in the resulting group are the science exposures and which are the background exposures.

stscieisenhamer:
# Resources
- [Problem Report #87164](https://www.ess.stsci.edu/prsystem/servlet/prbrowse/pr.87164): JWST: add level-1b FITS keyword for BKGDTARG (background = T/F) for use in Asn Pool

stscieisenhamer:
# Implementation
- [ ] Mockup Pool with
    - BACKGROUND column where non-Null flags the exposure as background
    - association candidate BACKGROUND with one observation as background
- [ ] Rules
    - [ ] Nodding _block_
    - [ ] Chopping

stscieisenhamer:
Question: Besides DETECTOR, what else has to match between the science and background exposures? Sub-array size, groups, integrations, anything else? Or not?

philhodge:
Filter and/or grating should match.  The subarray size (but not 
location) might need to match between science and background exposures, 
but maybe not.  I can imagine computing a single value for the 
background by averaging (or sigma-clipped average, or median) over an 
[...]

PR #2307, opened 2018-07-31
Implement Rule for Level 2 NIRSpec Fixed-slit background

### LINKED ISSUE #2127: Rule: NRS Fixed Slit Level2 background nod subtraction
# Issue
Implement Level2 associations for background subtraction from nodding for NIRSpec Fixed Slit.

# Resources
- #735 
- #2121
- [NIRSpec FS Dither and Nod Patterns](https://jwst-docs.stsci.edu/display/JTI/NIRSpec+FS+Dither+and+Nod+Patterns)

# Implementation
All exposures that satisfy the constraints below are nods of each other. The background for each exposure is calculated from all the other exposures.

## Constraints
- optical path (standard)
- `obs_num`
- `visit`
- `patttype`
- `patt_num`
- `subpxnum`

## To Do
- [ ] bump version
- [ ] documentation
- [ ] tests
- [ ] code
- [ ] tests
- [ ] documentation
- [ ] bump version

stscieisenhamer:
Implement NIRSpec Fixed-slit Science Level2 background gathering as per the specifications defined in [JP-350](https://jira.stsci.edu/browse/JP-350).

Also implemented is a small code change but significant architecture change in the `Association.check_and_set_constraints` method. The associations constraints are cached before the check and set. If the match fails, the cached version is restored.

Resolves #2127 
WIP #735


jwst/associations/__init__.py
```diff
@@ -1,5 +1,4 @@
-
-__version__ = '0.9.20'
+__version__ = '0.9.22'
```
hbushouse: We just completed release 0.10.0, so this should be bumped up accordingly (0.10.1?)

stscieisenhamer: 👍 

About the documentation. If you are referring to [line 365](https://github.com/spacetelescope/jwst/pull/2307/files#diff-45c66a113fcc54865dc17ccf6190a464R365), where in the `finalize` statement, it refers to MSA specifically, yes, they are handled the same. I'll update the comment to be clearer. Otherwise, I'm not sure what you are referring to. There are two rules because the relevant keywords are different.

PR #2315, opened 2018-08-01
Revamped refpix to be more efficient

stsci-hack:
This version of refpix.reference_pixels completely revises the memory usage of the algorithm, while also resulting in improved run-times (from 3m16s to 23s for 250,12,256,2048 NRCALONG simulated observation).

### DIFF
--- a/jwst/refpix/reference_pixels.py
+++ b/jwst/refpix/reference_pixels.py
@@ -825,6 +825,7 @@ class NIRDataset(Dataset):
         #
         refdq = dqflags.pixel['REFERENCE_PIXEL']
 
+        self.DMS_to_detector(0,0)
         # Determined refpix indices to use on each group
         refpixindices = np.where(np.bitwise_and(self.pixeldq, refdq) == refdq)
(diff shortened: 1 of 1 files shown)


jwst/refpix/reference_pixels.py
```diff
@@ -788,279 +823,300 @@ def do_subarray_corrections(self):
         #
         #  First transform to detector coordinates
         #
-        self.DMS_to_detector()
         refdq = dqflags.pixel['REFERENCE_PIXEL']
```
stscirij: need to add
self.DMS_to_detector(0,0)
to make sure the data and dq groups are shifted/rotated to DMS coordinates

stsci-hack: Addressed with commit stsci-hack/jwst@47dd1d9b839fff2d14806580c18a5a1ed3d72da4.

stsci-hack: This should now be addressed properly with 86be1ed.

stscirij: This looks good now

Issue #1508, opened 2017-11-21
Determine and simplify jwst dependencies


nden:
The JWST dependencies are not listed anywhere and sometimes it's a matter of trial and error to determine everything that's needed for the pipeline to run. This issue makes an attempt to determine the dependencies, try to reduce their number and eventually document  them.
I put inline comments/questions. Feel free to add more.

Installation dependencies:
- `sphinx`
  Is this really necessary from the user point of view? Isn't it sufficient to document the command to build the documentation if they wanted to?
    - _jde_: Agreed from a normal user-viewpoint. However, it would be nice if when one does the `python setup.py build_sphinx`, that the dependencies get installed, much as they do with `python setup.py test`.
- ~`namedlist`~ #3065
 Could this be replaced with `namedtuple`? What is the functionality that requires install time third party dependence?
    - _jde_ Yes, the functionality can be reproduced internally. Big question though is what are the standards for using/not using third-party? Discuss among yourselves...

Testing dependencies:
- ~`backports.tmpfile`~ #1510, #1585
    - _jde_ Now that we're python3 only, [this can go](https://github.com/STScI-JWST/jwst/issues/930).
- `pytest`
- ~`pytest_catchlog`~ #1585
- `requests_mock`

Running calspec2:
- `astropy`
- `asdf`
- `gwcs`
- `jsonschema`
- ~`nose`~ #1585
- `drizzle`
- ~`fitsblender`~ #1689 , #1690 , #1708, #1955
  An old version of fitsblender is on pypi, needs to be updated as the one there uses pyfits.- 
  (spacetelescope/fitsblender#15)
- `stsci.tools` #3145 - the only module from `stsci.tools` is `bitmask`. It may be worth copying it here to avoid the dependency. But `stsci.image` still depends on `stsci.tools`.
- ~`d2to1`~ (required by `stsci.tools`) - spacetelescope/stsci.tools#55
- ~`stsci.distutils`~  (required by `stsci.tools`)  - spacetelescope/stsci.tools#55
- ~`stwcs`~ (imported by `stsci.tools.fileutil.OpenImage` via `fitsblender`):scream: - #1689 , #1690 , #1708 
- `stsci.image`
- ~`stsci.convolve`~ (required by `stsci.image`) 
  Isn't the `astropy.convolution.convolve` function good enough?
- `photutils`
- `stsci.imagestats`
  It may be possible to replace use of this with `astropy.stats.sigma_clipped_stats`...
[...]

PR #2326, opened 2018-08-03
Implement package dependencies

### LINKED ISSUE #1861: Updating install_requires
I was trying to install the pipeline from a fresh environment and there were a number of packages that were not included in the setup.py install_requires.

jhunkeler:
Changes:
* Populate `install_requires=` with meaningful dependencies
  * Settled on a mixture of inline building development via `requirements` file
  * and using `>=` specifier under `install_requires`

~* Use `~=` version specifier rather than `>=`~
~* `~=1.1` is equivalent to `>=1.1, ==1.*`~

Known problems:
* Non-Python dependencies cannot be handled gracefully. (Out of scope of this PR)
* `jwst` cannot be installed without `numpy` present beforehand
* `crds` cannot be installed without `astropy` and `numpy` present beforehand (use of `requires=` instead of `install_requires=` in `setup.py` under older releases is the cause of this behavior)
* `pytest` is a run-time dependency of `jwst` because it loads `asdf`'s `pytest` plugin as it attempts to verify JSON schemas
* `pytest-remotedata` package causes `pytest` installation failure. (Removed from dependency list)

Resolves #1861


jhunkeler:
See: #1508

jdavies-st:
Looks good.  I did test of this by creating a clean conda environment and running setup in this repository:
```
$ conda create --name pr2326 python=3.7
$ conda activate pr2326
(pr2326) $ python setup.py install
Unable to import "numpy".
Please install "numpy" and try again.
(pr2326) $ conda install numpy
(pr2326) $ python setup.py install
/Users/jdavies/anaconda3/envs/pr2326/lib/python3.7/distutils/dist.py:274: UserWarning: Unknown distribution option: 'long_desctiption'
  warnings.warn(msg)
/Users/jdavies/anaconda3/envs/pr2326/lib/python3.7/site-packages/setuptools/dist.py:398: UserWarning: Normalizing '0.10.1a.dev95' to '0.10.1a0.dev95'
  normalized_version,
running install
    <snip>
Processing dependencies for jwst==0.10.1a0.dev95
Searching for verhawk~=0.0
Reading https://pypi.org/simple/verhawk/
    <snip>
Installed /Users/jdavies/anaconda3/envs/pr2326/lib/python3.7/site-packages/verhawk-0.0.3-py3.7.egg
Searching for pytest-remotedata
Reading https://pypi.org/simple/pytest-remotedata/
    <snip>
Installed /Users/jdavies/anaconda3/envs/pr2326/lib/python3.7/site-packages/pytest_remotedata-0.3.0-py3.7.egg
Searching for pytest
Downloading https://files.pythonhosted.org/packages/99/c5/49bee2b4123c9738760ff9915970973434d8c6ad1759faa40e20bc8a2a8e/pytest-remotedata-0.3.0.tar.gz#sha256=cb0a6e22ba07ff870edeb17d5d38c495e8f37e2db741cd15e723fb8e609909eb
Best match: pytest remotedata-0.3.0
    <snip>
Installed /Users/jdavies/anaconda3/envs/pr2326/lib/python3.7/site-packages/pytest_remotedata-0.3.0-py3.7.egg
error: The 'pytest' distribution was not found and is required by jwst
```
It only installed `jwst`, `verhawk`, `pytest-remotedata` and got stuck on `pytest`.  So then I decided to run the tests:
```
$ python setup.py test
```
And then all the dependencies got installed, but there was an error and the tests didn't run.  See attached output file.

[pr2326.txt](https://github.com/spacetelescope/jwst/files/2294920/pr2326.txt)

I then decided to just run pytest at the top level, and all the tests ran and passed.

So it looks good?!  Qualified 👍

nden:
#2382 removes `six` .

jhunkeler:
@jdavies-st The `ImportError` was due to how `crds` uses `requires` instead of `install_requires`. I guess @jaytmiller hasn't made a new pypi release yet. 

Workaround if you install `astropy` and `numpy` _before_ installing `jwst` the problem will go away.

`pip or conda install astropy numpy` then `python setup.py install` or `python setup.py test`

jhunkeler:
@jdavies-st The `pytest` installation failure was caused by `pytest-remotedata`. There's something wrong with it... I've removed it from the requirements list since nothing is using it anyway.

jdavies-st:
Ok!  The install now runs to completion:
```
rey> conda create --name pr2326 python=3.7 numpy astropy
Solving environment: done

## Package Plan ##

  environment location: /Users/jdavies/anaconda3/envs/pr2326

  added / updated specs: 
    - astropy
    - numpy
    - python=3.7


The following NEW packages will be INSTALLED:

    astropy:            3.0.4-py37h1de35cc_0   
    atomicwrites:       1.2.1-py37_0           
    attrs:              18.2.0-py37h28b3542_0  
    blas:               1.0-mkl                
    ca-certificates:    2018.03.07-0           
    certifi:            2018.8.24-py37_1       
    intel-openmp:       2018.0.3-0             
    libcxx:             4.0.1-h579ed51_0       
    libcxxabi:          4.0.1-hebd6815_0       
    libedit:            3.1.20170329-hb402a30_2
    libffi:             3.2.1-h475c297_4       
    libgfortran:        3.0.1-h93005f0_2       
    mkl:                2018.0.3-1             
    mkl_fft:            1.0.4-py37h5d10147_1   
    mkl_random:         1.0.1-py37h5d10147_1   
    more-itertools:     4.3.0-py37_0           
    ncurses:            6.1-h0a44026_0         
    numpy:              1.15.1-py37h6a91979_0  
    numpy-base:         1.15.1-py37h8a80b8c_0  
    openssl:            1.0.2p-h1de35cc_0      
    pip:                10.0.1-py37_0          
    pluggy:             0.7.1-py37h28b3542_0   
    psutil:             5.4.7-py37h1de35cc_0   
    py:                 1.6.0-py37_0           
    pytest:             3.7.4-py37_0           
    pytest-arraydiff:   0.2-py37h39e3cac_0     
    pytest-astropy:     0.4.0-py37_0           
    pytest-doctestplus: 0.1.3-py37_0           
    pytest-openfiles:   0.3.0-py37_0           
    pytest-remotedata:  0.3.0-py37_0           
    python:             3.7.0-hc167b69_0       
    readline:           7.0-h1de35cc_5         
    setuptools:         40.2.0-py37_0          
    six:                1.11.0-py37_1          
    sqlite:             3.24.0-ha441bb4_0      
    tk:                 8.6.8-ha441bb4_0       
    wheel:              0.31.1-py37_0          
    xz:                 5.2.4-h1de35cc_4       
    zlib:               1.2.11-hf3cbc9b_2      

Proceed ([y]/n)? 

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
# To activate this environment, use
#
#     $ conda activate pr2326
#
# To deactivate an active environment, use
#
#     $ conda deactivate

rey> conda activate pr2326
(pr2326) rey> pwd
/Users/jdavies/dev/jwst
(pr2326) rey> python setup.py install
/Users/jdavies/anaconda3/envs/pr2326/lib/python3.7/site-packages/setuptools/dist.py:398: UserWarning: Normalizing '0.10.1a.dev250' to '0.10.1a0.dev250'
  normalized_version,
running install
running bdist_egg
running egg_info
writing jwst.egg-info/PKG-INFO

...

Using /Users/jdavies/anaconda3/envs/pr2326/lib/python3.7/site-packages
Finished processing dependencies for jwst==0.10.1a0.dev250
(pr2326) rey>
```
But running the `jwst` unit tests sees 3 tests fail:
```
(pr2326) rey> pytest -n 8 .
==================================================================== test session starts ====================================================================
platform darwin -- Python 3.7.0, pytest-3.7.4, py-1.6.0, pluggy-0.7.1
plugins: xdist-1.23.0, remotedata-0.3.0, openfiles-0.3.0, forked-0.2, doctestplus-0.1.3, arraydiff-0.2, requests-mock-1.5.2
gw0 [1016] / gw1 [1016] / gw2 [1016] / gw3 [1016] / gw4 [1016] / gw5 [1016] / gw6 [1016] / gw7 [1016]
scheduling tests via LoadScheduling
........................................................s...s...s...s......................x.s......sssssss...........................ssssss......... [ 14%]
.............s....................FF.F........s.x.sssssss...............................................................s............................ [ 29%]
..................................................s.........sssssss............................................s..................................... [ 43%]
.s..s................................................................................................................................................ [ 58%]
...........................................................ssssssssssssssssssssssssssssssssssssssssssss.sssssssssssss................................ [ 73%]
...............................sssssss...x.......................sssssss......................................................x...................... [ 87%]
...........................s..sssssss...............................................sssssss..............................x
========================================================================= FAILURES ==========================================================================
_________________________________________________________________ test_historylist_methods __________________________________________________________________
[gw2] darwin -- Python 3.7.0 /Users/jdavies/anaconda3/envs/pr2326/bin/python

    def test_historylist_methods():
        m = DataModel()
>       h1 = m.history

jwst/datamodels/tests/test_history.py:32: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <DataModel>, attr = 'history'

    def __getattr__(self, attr):
        from . import ndmodel
    
        if attr.startswith('_'):
            raise AttributeError('No attribute {0}'.format(attr))
    
        schema = _get_schema_for_property(self._schema, attr)
        try:
            val = self._instance[attr]
        except KeyError:
            if schema == {}:
>               raise AttributeError("No attribute '{0}'".format(attr))
E               AttributeError: No attribute 'history'

jwst/datamodels/properties.py:206: AttributeError
______________________________________________________________ test_history_from_model_to_fits ______________________________________________________________
[gw2] darwin -- Python 3.7.0 /Users/jdavies/anaconda3/envs/pr2326/bin/python

    def test_history_from_model_to_fits():
        m = DataModel()
        m.history = [HistoryEntry({
            'description': 'First entry',
>           'time': Time(datetime.datetime.now())})]

jwst/datamodels/tests/test_history.py:68: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
jwst/datamodels/model_base.py:572: in __setattr__
    object.__setattr__(self, attr, value)
jwst/datamodels/model_base.py:940: in history
    entries = self.history
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <DataModel>, attr = 'history'

    def __getattr__(self, attr):
        from . import ndmodel
    
        if attr.startswith('_'):
            raise AttributeError('No attribute {0}'.format(attr))
    
        schema = _get_schema_for_property(self._schema, attr)
        try:
            val = self._instance[attr]
        except KeyError:
            if schema == {}:
>               raise AttributeError("No attribute '{0}'".format(attr))
E               AttributeError: No attribute 'history'

jwst/datamodels/properties.py:206: AttributeError
__________________________________________________________________ test_history_from_fits ___________________________________________________________________
[gw2] darwin -- Python 3.7.0 /Users/jdavies/anaconda3/envs/pr2326/bin/python

    def test_history_from_fits():
        header = fits.Header()
        header['HISTORY'] = "First entry"
        header['HISTORY'] = "Second entry"
        fits.writeto(TMP_FITS, np.array([]), header, overwrite=True)
    
        m = DataModel(TMP_FITS)
>       assert m.history == [{'description': 'First entry'},
                             {'description': 'Second entry'}]
E       AssertionError: assert <jwst.datamod...t 0x1158f64a8> == [{'description...econd entry'}]

jwst/datamodels/tests/test_history.py:101: AssertionError
======================================== 3 failed, 883 passed, 125 skipped, 5 xfailed, 82 warnings in 213.90 seconds ========================================
```
The failure is due to the fact that code in `jwst` depends on features in `asdf` that have not made it into an official release yet.  Easy Install pulls `asdf 2.0.2` which is the latest release, but the features the `jwst` pipeline currently need are in `asdf 2.1.devXXX`.  This highlights the need to coordinate updating these dependencies when they occur.

This may involve on `jwst/master` putting in a placeholder dependency to `git+git://github.com/spacetelescope/asdf` with the relevant commit, so that it is clear when we do a release that this be converted to whatever `asdf` release includes that commit, or if it doesn't yet, waiting until it does before we do a `jwst` release.

I don't think this is a showstopper for this PR.  I think it can be merged with the caveat that we need to start keeping track of these dependencies, mostly for `asdf` and `gwcs`.  Probably `astropy` too.

My only request would be that the message requiring numpy be installed at the beginning be amended with astropy as well, at least until crds issue is sorted out.

jhunkeler:
The `crds` installation problem is more involved than I thought. The latest version uses `install_requires` now, but `setup.py` imports `crds` locally to read the value of `__version__`. That's fine, except `crds/__init__.py` imports parts of the package that also require `astropy`.

I was able to circumvent the error locally by checking whether `distutils.core` was imported or not. I'm not sure if this is an acceptable solution though.

`crds/__init__.py`:

```python
# [...]
__version__ = "7.2.6"   # XXXX  see also ../setup.cfg
__rationale__ = "JWST DMS 7.2,  end-of-August release"

if 'distutils.core' in sys.modules:
    print("CRDS partially initialized...")
else:
    from .core import config   # module

    from .core.rmap import get_cached_mapping, asmapping
    from .core.config import locate_mapping, locate_file
    # [...]
```

cc: @jaytmiller

jdavies-st:
Btw, why does `jwst` require `numpy` to be installed beforehand?  That seems like an odd requirement just for installation.

sosey:
It's in the setup to help with the tweakreg extension, we could probably code that by hand

stscicrawford:
I think it is using np_include in the seutp.py here:
https://github.com/spacetelescope/jwst/blob/1a8c6b627546d9e29cd3e08c87e16873a76a74d3/setup.py#L172

which is what is requiring it

sosey:
then again, I'm not sure we can remove it from the extension build...but also, we use numpy all over the place in the package, what would be the specific purpose to run setup without numpy and then make the users install it to use the package?

jdavies-st:
Ah, I see now.

The purpose of this PR is to be able to download the `jwst` repository and do `python setup.py install` and get it installed with all its runtime dependencies.  It will install `numpy`, as that is a dependency.  But it actually needs `numpy` before you even try installing.  That's the issue.  It currently also needs `astropy` because of an issue in `crds`.  Instead of runtime dependencies, they are installation dependencies.

Anyway, would be nice to get it sorted.

That said, I think the PR in its current state is worth merging, with the above caveats and the `asdf` release version issue.

drdavella:
The requirement on ASDF can/should now be updated to `~=2.1`.

drdavella:
Re the `numpy` dependency within `setup.py`: this is a problem that packages need to solve when they provide extensions that have a build dependency on `numpy`. The way `astropy` currently solves this is by implementing a custom `distutils` extension that *waves hands* ensures that `numpy` is installed before the extensions are built.

The way going forward is to use `pypackage.toml` (see [PEP518](https://www.python.org/dev/peps/pep-0518/)). This allows packages to express dependencies of the `setup.py` file itself, and ensures they will be installed before `setup.py` is executed. However, this requires `pip>=10`, and I'm not sure we can guarantee adoption by all users yet (although `conda` looks like it now provides `pip-10.0.1` by default).

The current workaround is okay, since if anyone tries to `pip install jwst` without `numpy` in their environment, they'll get a relatively helpful error message. But it's not ideal. If we really wanted, we could use `astropy_helpers` to solve this problem in the interim, but I'm a little hesitant to recommend it. We could also implement a custom `distutils` extension here to solve this problem. But ultimately I think we want to move in the direction of using `pyproject.toml`.

jdavies-st:
Thanks @drdavella.  We should look into that, perhaps separately from this PR.

Now that ASDF 2.1 is released, I ran the installation again, as done above.  All the unit tests pass.  So I would recommend merging this with an update to CHANGES.rst.

It might also be a good idea to add this build to the CI test matrix at some point.

jaytmiller:
@jhunkeler  Sorry Snake,  I didn't see this due to a bad mail filter.

I modified CRDS setup.py to get rid of the crds.__version__ dependency and avoid importing crds.   I think that should prevent the heavier CRDS imports,  hope it helps.

jdavies-st:
For the CRDS fix to take hold, will we have to wait for a new release of CRDS @jaytmiller?

jaytmiller:
Not sure.   In theory the CRDS import fix is on CRDS master now and hence on conda-dev starting today.   It is not in astroconda-contrib.

jdavies-st:
A quick test of this PR as I'd done above indicates that the CRDS issue discussed above is now fixed.  The `install_requires` versions should be:
```
'asdf~=2.1',
'crds~=7.2.7',
```
I've also set the PR against `0.13.0_new_features` in case we'd like to merge this.  Since Joe is out, perhaps I'll make a new PR with these new changes, and also remove `six`?  I think this can be merged.

On a related note, it would be nice if we could manage these dependencies in one place.  As we can see just in this PR, it's very easy to miss that we have the `six` dependency in `.travis.yml` file but not in `setup.py`.   Any suggestions for centralizing dependencies would be great.

jdavies-st:
I suspect changing `crds~=7.2.7` to `crds>=7.2.7` will fix the Jenkins failure.  Not sure why.

jdavies-st:
This looks great!  I like the parallel test on ssbjenkins testing this `pip` installation machinery.

setup.py
```diff
@@ -167,13 +173,33 @@ def run_tests(self):
...
+                  define_macros=[('NUMPY', '1')]),
     ],
     install_requires=[
-        'namedlist'
+        'asdf>=2.1',
+        'astropy>=3.0',
+        'crds>=7.2.7',
+        'drizzle>=1.12',
+        'gwcs>=0.9',
+        'jsonschema>=2.6',
```
jdavies-st: This needs to be changed to `>=2.3,<=2.6` to correspond with what `asdf` needs.

setup.py
```diff
@@ -167,13 +173,33 @@ def run_tests(self):
...
+        'photutils>=0.4',
+        'pytest',
+        'verhawk>=0.0',
     ],
+    extras_require={
+        'ephem': ['pymssql>=2.1', 'jplephem>=2.8'],
+    },
     tests_require=[
         'pytest',
         'requests_mock',
```
jdavies-st: tests require `ci-watson` and `pytest-doctestplus`.

requirements-dev.txt
```diff
@@ -0,0 +1,4 @@
+git+https://github.com/astropy/astropy.git#egg=astropy
+git+https://github.com/spacetelescope/asdf.git#egg=asdf
+git+https://github.com/spacetelescope/gwcs.git#egg=gwcs
+pytest==3.8.2
```
jdavies-st: This `pytest==3.8.2` can be removed.  The normal `pytest` dependency in `setup.py` is fine for dev.  The `3.8.2` peg was there to avoid `3.9.x` which was completely broken.

setup.py
```diff
@@ -73,7 +78,19 @@ def run(self):
         '*.asdf'
     ]
 }
-
+DOCS_REQUIRE = [
+    'matplotlib',
+    'sphinxcontrib-programoutput',
+    'sphinx-automodapi',
+    'stsci-rtd-theme',
+    'sphinx-rtd-theme'
```
jdavies-st: Do we also need `sphinx` in `DOCS_REQUIRE`?

setup.py
```diff
@@ -167,18 +185,36 @@ def run_tests(self):
...
+                  include_dirs=[np_include()],
+                  define_macros=[('NUMPY', '1')]),
     ],
     install_requires=[
-        'namedlist'
-    ],
-    tests_require=[
-        'pytest',
+        'asdf>=2.1',
+        'astropy>=3.0',
```
jdavies-st: Let's bump `astropy>=3.1`, as that has now been released (and is required) since you first started this PR.  We'll have to bump `asdf` too once the next release is made.

Jenkinsfile
```diff
@@ -32,47 +31,65 @@ def conda_packages = [
...
+            def name = "pip_py${python_ver}np${numpy_ver}ap${astropy_ver}"
+            bc = new BuildConfig()
+            bc.nodetype = 'linux'
+            bc.env_vars = test_env
+            bc.name = name
+            bc.conda_packages = ["python=${python_ver}"]
+            bc.build_cmds = [
+                "pip install ${pip_install_args} numpy==${matrix_numpy[0]}",
+                "pip install ${pip_install_args} ${pip_packages_tests}",
+                "pip install ${pip_install_args} -r requirements-dev.txt .",
```
jdavies-st: I think we need to change this line to
```
pip install ${pip_install_args} -r requirements-dev.txt .[test]
```
?

jdavies-st: And we probably don't need `pip_packages_tests` and `pip_packages_docs` defined or used anymore.

jdavies-st:
Looks great @jhunkeler.  This is a big improvement and paves the way released versions of upstream dependencies.

Issue #1744, opened 2018-02-14
Update sky footprint (S_REGION) for resampled spectroscopic products


hbushouse:
Right now the S_REGION keyword is first populated by the set_telescope_pointing script during level-1b processing, which computes the footprint based on the JWST aperture vertices listed in the SIAF. For imaging exposures, S_REGION is then updated by the assign_wcs step based on the full WCS specification for the exposure, and again by the resample step when it creates a drizzled/combined image.

But spectroscopic exposures do not have their S_REGION value updated at all during either assign_wcs or resampling. Hence the footprint in the calibrated products can often still be set to the full area of the detector on the sky, rather than something more useful like the actual slit or entrance aperture (IFU) projection on the sky. The archive (CAOM in particular) really needs to have the actual slit/aperture footprint computed, at least for our final, resampled products.

So at the very least we should try to recompute a realistic value for S_REGION in:
- [x] resample_spec results (s2d) for individual slits
- [x] cube_build results (s3d) for the NIRSpec and MIRI IFU apertures

hbushouse:
A problem in the current population of S_REGION at level-1b is due to the fact that the SIAF doesn't have complete info for all instrument apertures, especially some of the IFU and slit apertures. For example, MIRI MRS exposures on the short-wave detector may use an original PPS aperture of 'MIRIFU_CHANNEL1A', but because some of the SIAF info for that aperture is missing (like X/Y reference point and X/Y image scale), SDP resets the APERNAME to something generic like 'MIRIM_FULL', which does have a complete set of info in the SIAF. But that in turn means we're getting incorrect values for items like V2_REF, V3_REF, and all the aperture vertices. This in turn means that not only is our computed pointing (RA_REF, DEC_REF) incorrect (it's giving pointing for the Imaging detector instead of the MRS detector), but the S_REGION is also using values appropriate for the footprint of the Imaging detector, instead of the MRS detector.

For example, CV3 test dataset jw93135311001_02101_00001_mirifushort_uncal.fits (available at /grp/jwst/ins/mary/b7.1rc9_full/jw93135/) has S_REGION =
```
S_REGION= 'POLYGON ICRS  315.4334256011184 68.13928143846911 &'                 
CONTINUE  '315.3486633938344 68.14172066715514 315.35470198410417 &'            
CONTINUE  '68.17294015103279 315.43961112297023 68.1707868402061&'              
CONTINUE  '' / spatial extent of the observation                        
```
[...]

PR #2327, opened 2018-08-03
spectral footprint

nden:
This is part one of #1744.

There are two functions now - `update_s_region_imaging` and `update_s_region_spectral`.

The flow is:
- `set_telescope_pointing` populates `S_REGION` using the vertices of the APERZTURE in the SIAF file.
   For this to work we need the correct APERTURE in the level 1B headers (populated by SDP).
- `assign_wcs` updates `S_REGION` for imaging types.
- `extract_2d updates the keyword for each slit
- `resample` updates fo rimaging and spectral (@jdavies-st)
- `cube_build` updates it for cubes (@jemorrison please look at the changes)

### DIFF
--- a/jwst/assign_wcs/assign_wcs.py
+++ b/jwst/assign_wcs/assign_wcs.py
@@ -1,7 +1,14 @@
 import logging
 import importlib
 from gwcs.wcs import WCS
-from.util import update_s_region
+from .util import update_s_region_keyword, update_s_region_imaging
+from ..associations.lib.dms_base import (TSO_EXP_TYPES, ACQ_EXP_TYPES,
+                                         IMAGE2_SCIENCE_EXP_TYPES,
+                                         IMAGE2_NONSCIENCE_EXP_TYPES)
+
+IMAGING_TYPES = set(tuple(ACQ_EXP_TYPES) + tuple(IMAGE2_SCIENCE_EXP_TYPES)
+                    + tuple(IMAGE2_NONSCIENCE_EXP_TYPES))
+
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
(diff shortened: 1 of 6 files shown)


jemorrison:
The cube_build update of S_REGION seems fine to me.

jwst/assign_wcs/assign_wcs.py
```diff
@@ -50,18 +57,29 @@ def load_wcs(input_model, reference_files={}):
...
-            if orig_s_region != output_model.meta.wcsinfo.s_region.strip():
-                log.info("assign_wcs updated S_REGION to {0}".format(output_model.meta.wcsinfo.s_region))
-        else:
-            log.info("assign_wcs did not update S_REGION for type {0}".format(output_model.meta.exposure.type))
-        log.info("COMPLETED assign_wcs")
+        if output_model.meta.exposure.type.lower() not in exclude_types:
+            if output_model.meta.exposure.type.lower() in IMAGING_TYPES:
+                try:
+                    update_s_region_imaging(output_model)
+                except:
```
jdavies-st: What kinda of error are you expecting here?  ValueError?  Generic Exception?  If so, it might be good to catch it:
```
except Exception as error:
    log.error("Unable to update S_REGION for type  %s: %s", output_model.meta.exposure.type, error)
```
Similar with the bare `except` blocks below.

jemorrison:
I am not sure what sregion = Final_IFUCube.meta.wcs.footprint(kind="spatial") is doing.
That line is not in my version of the code. It must of been added by someone. 

Final_IFUCube = datamodels.ModelContainter()
It contains all the IFUCubes that are created with a single run of cube_build.
So if say 12 IFUCubes are to be created (one for each MIRI band)
then after each one has been created 
Final_IFUCube.append(result)
where result is a single IFU Cube. 

Is this failing in regression test for when cube_build is being called to support outlier detection
or imatch_mrs ? Final_IFUCube  is still a ModelContainer but a container for Single IFU Cubes.
If you can point me to the test I can try and figure out what is failing.

jdavies-st:
I think it's a case of an ambiguously-named variable.  `Final_IFUCube` should really be called `cube_container` or something similar, as it is a container of `IFUCubeModel` instances.

jemorrison:
ok I will up date that in my code but it should still not give an error in regression testing. That part
does not make sense.

jdavies-st:
This PR added the S_REGION lines to cube_build.  But containers do not contain `meta.wcs`.  So it fails.  It ought to be something like:
```
for model in Final_IFUCube:
    s_region = model.meta.wcs.footprint(kind="spatial")
    model.meta.wcsinfo.s_region = s_region
    log.info("Updated S_REGION to {}".format(s_region))
```
but the code would be much clearer if Final_IFUCube was renamed to something indicating it is not actually a cube but a container of cubes.  I think that's were the confusion entered.

There was a similar problem in the update to my resample code in this PR.  I didn't catch it either.  I have fixed it in #2348.

jemorrison:
How about we just get this PR merged. I have many changes to the cube_build step that I am working
on - trying to increase speed  and just general pep8 clean up. I will fix the name of the Final_IFUCube
in my changes - otherwize I think I might get really confused trying to merge these changes with what
I have already done.

Issue #159, opened 2016-08-15
naming of source-based files


stscieisenhamer:
How will the source based data be named? Possible sources:
- MultiSlitModel.slit.name
- MultiSlitModel.slit.slitlet_id
- MultiSlitModel.slit.source_id
- Same combination involving filenames.

stscieisenhamer:
@hbushouse Thoughts?

hbushouse:
The exp_to_source task will be run as essentially the first step of the calspec3 pipeline, at least for some observing modes, i.e. the ones that obtain spectra of multiple sources simultaneously (which means NIRCam and NIRISS WFSS, NIRSpec MOS, and NIRSpec fixed-slit). The input to the calspec3 pipeline will be a level-3 ASN table, which will list all the input level-2b exposure-based products. Because the ASN generator can't know at the time it runs what the various output sources will be and hence it can't specify the output source-based file name(s) in the ASN table, either the calspec3 pipeline module or the exp_to_source task itself will need to construct the source-based output file names. The ASN generator could at least specify a template source-based file name in the ASN table, such as:

"jw12345_o0001_<targ_id/source_id>_nirspec_g140m_<prod_type>.fits"

and then leave it to either calspec3 or exp_to_source to fill in both <targ_id/source_id> and <prod_type> as each source-based file is created. For WFSS and MOS modes, we should have an actual source_id available for each source from the WFSS source catalog or the MSA meta data table, respectively. Those names could easily be stored in the SOURCEID keyword for each slit instance that was created upstream in calspec2 during the extract_2d step, and hence that's where exp_to_source can get them. The NIRSpec fixed-slit case is not as straightforward, because we won't have individual source id's specified upstream for each slit. I think the best that can be done for that mode is to use the generic target_id from the exposures (which should be the same for all input exposures) and perhaps append the slit name (S200A1, S200B1, etc.) as an additional part of either the target_id field or the optical element field of the output source-based file name.

stscieisenhamer:
Resolved by #2330

PR #2330, opened 2018-08-07
Handle both integer and string source names

### LINKED ISSUE #2311: NIRSpec FSS source id handling
# Issue
For NIRSpec Fixed-slit, conversion from exposure to source raises the following:
```
2018-07-31 15:41:18,296 - stpipe.Spec3Pipeline - INFO - Convert from exposure-based to source-based data.
----------------------------------------------------------------------
ERROR RUNNING STEP 'Spec3Pipeline':
    invalid literal for int() with base 10: 'S400A1'
  File "/Users/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/jwst-0.10.1a0-py3.6-macosx-10.7-x86_64.egg/jwst/stpipe/cmdline.py", line 297, in step_from_cmdline
    step.run(*positional)
  File "/Users/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/jwst-0.10.1a0-py3.6-macosx-10.7-x86_64.egg/jwst/stpipe/step.py", line 400, in run
  File "/Users/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/jwst-0.10.1a0-py3.6-macosx-10.7-x86_64.egg/jwst/pipeline/calwebb_spec3.py", line 106, in process
    output_file, source_id=int(source_id)
ValueError: invalid literal for int() with base 10: 'S400A1'

----------------------------------------------------------------------
  File "/Users/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/jwst-0.10.1a0-py3.6-macosx-10.7-x86_64.egg/EGG-INFO/scripts/strun", line 14, in <module>
  File "/Users/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/jwst-0.10.1a0-py3.6-macosx-10.7-x86_64.egg/jwst/stpipe/step.py", line 162, in from_cmdline
  File "/Users/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/jwst-0.10.1a0-py3.6-macosx-10.7-x86_64.egg/jwst/stpipe/cmdline.py", line 297, in step_from_cmdline
    step.run(*positional)
  File "/Users/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/jwst-0.10.1a0-py3.6-macosx-10.7-x86_64.egg/jwst/stpipe/step.py", line 400, in run
  File "/Users/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/jwst-0.10.1a0-py3.6-macosx-10.7-x86_64.egg/jwst/pipeline/calwebb_spec3.py", line 106, in process
    output_file, source_id=int(source_id)
ValueError: invalid literal for int() with base 10: 'S400A1'
```

# Resources
- #159 
- Example data set `/grp/jwst/ssb/eisenhamer/testing_jscott/20180731/level3_sourceid`

stscieisenhamer:
Resolves #2311, #159


stscicrawford:
Looking at your commit messages, I imagine that you have had some problems with rebasing so I don't want to pile on, but it also looks like the diff is including a number of files that it might not need to be part of this pull request.   Any chance you'd have to clean up the PR again or can you at least point out the updates to be reviewed?

stscieisenhamer:
wow. didn't realize it went that far south...just a sec...

stscieisenhamer:
OK. Cleaned. Pending successful completion, the files for review are as follows, basically all the code files. Note, all the json files and tests that use them are slated to be moved to the regression set of tests, and the files moved to artifactory.

Code files:
- association/lib/format_template.py
- association/lib/rules_level3.py
- association/lib/rules_level3_base.py
- pipeline/calwebb_spec3.py

jwst/associations/lib/rules_level3.py
```diff
@@ -112,14 +114,194 @@ def __init__(self, *args, **kwargs):
             ),
             Constraint_Optical_Path(),
             Constraint_Target(),
+            DMSAttrConstraint(
+                name='exp_type',
+                sources=['exp_type'],
+                value=(
+                    'mir_lrs-fixedslit'
+                    '|mir_lrs_slitless'
```
hbushouse: Should ``nis_soss`` be included in this list, because it's an equivalent mode to ``mir_lrs_slitless`` (slitless, but single target always located at the same place)?

jwst/associations/lib/rules_level3.py
```diff
@@ -112,14 +114,194 @@ def __init__(self, *args, **kwargs):
...
+            ),
+            Constraint_Optical_Path(),
+            Constraint_Target(),
+            Constraint(
+                [
+                    DMSAttrConstraint(
+                        name='exp_type',
+                        sources=['exp_type'],
+                        value=(
+                            'nrc_grism'
```
hbushouse: ``nrc_grism`` has been relabeled to ``nrc_wfss``, which is already in the list, so ``nrc_grism`` could be removed. Also, ``nrc_tsgrism`` likely belongs in the other list, because it's a single target slitless mode.

jwst/associations/lib/rules_level3.py
stscieisenhamer: More specifically, `mir_lrs-slitless` does not need to be here, and the reasons `nis_soss` and `nrc_tsgrism` are not here, is that these are both "always time series", and handled by the `Asn_TSO` rule.

hbushouse: Got it. However, at some time in the future when a TSO option/switch is added for all observing modes, it will be possible for things like ``nis_soss`` to be treated as non-TSO, so additional logic may be needed then.

stscieisenhamer: So, the plan is to allow "always TSO" to also be switchable? I had understood that even when the option is there, that the "always TSO" one would not be allowed. Or should we just "CYA" this and plan for both?

Note: currently APT 26.0.3 allows addition of TSO special requirement, but it does not allow one to "take it away".

rizeladiaz: Currently, nis_soss observations can be non-TSO. Should be treated as  TSO only when the flag says it is TSO. If the "Time Series Observation" is NOT present in the special requirements then it means it is a non-TSO. I believe that in one of the test cases I created for one of the OTB or EMTB tests I removed this flag. Though I don't remember if it got executed. So I agree with Howard that additional logic is needed.

hbushouse: Clicking through APT 26.0.3 shows that TSO status for various modes is:
 - ``mir_image``: TSO default Off, but can be turned On
 - ``mir_lrs-slitless``: TOS default On, but can be turned Off
 - ``mir_lrs-fixedslit``: TSO not allowed
 - ``mir_mrs``:  TSO not allowed (but supposedly will be in future)
 - ``nrc_image``: TSO not allowed (use ``nrc_tsimage`` instead)
 - ``nrc_tsimage``: TSO required (cannot be turned Off)
 - ``nrc_tsgrism``: TSO required (cannot be turned Off)
 - ``nrc_wfss``: TSO not allowed (use ``nrc_tsgrism`` instead)
 - ``nrs_fixedslit``: TSO not allowed (use ``nrs_brightobj`` instead)
 - ``nrs_ifu``: TSO not allowed (but supposedly will be in future)
 - ``nrs_msaspec``: TSO not allowed
 - ``nrs_brightobj``: TSO required (cannot be turned Off)
 - ``nis_image``: TSO not allowed (but supposedly will be in future)
 - ``nis_wfss``: TSO not allowed (use ``nrc_tsgrism`` instead)
 - ``nis_soss``: TSO default On, but can be turned Off
 - ``nis_ami``: TSO not allowed

So the permanent TSO modes are ``nrc_tsimage``, ``nrc_tsgrism``, and ``nrs_brightobj``, while ``mir_image``, ``mir_lrs-slitless``, and ``nis_soss`` can be toggled on/off.

stscieisenhamer:
@hbushouse @rizeladiaz I believe the issues have been addressed. Re-review please.

jdavies-st:
The test for this is in
```
jwst/tests_nightly/general/pipelines/test_nirspec_fs_spec3.py::test_save_source_only
```
but the input test data is missing from `jwcalibdev:/data/jwst_test_data/`, hencing causing the nightly regression test to fail:

https://boyle.stsci.edu:8081/job/RT/job/JWST/236/testReport/jwst.tests_nightly.general.pipelines/test_nirspec_fs_spec3/test_save_source_only/

PR #2333, opened 2018-08-07
Fitsgen dotfiles

### LINKED ISSUE #2317: Create_data crashing
This is for tracking the JIRA issue https://jira.stsci.edu/browse/JP-316

stscirij:
Hopefully this will fix #2317

### DIFF
--- a/jwst/fits_generator/create_data.py
+++ b/jwst/fits_generator/create_data.py
@@ -25,7 +25,7 @@ def get_proposals(base_directory='.'):
     proposal_list = []
     for dirname, subdirlist, filelist in os.walk(base_directory):
         for name in filelist:
-            if name.startswith('_'):
+            if name.startswith('.'):
                 pass
             else:
                 if name.endswith('.prop'):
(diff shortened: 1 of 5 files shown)


jwst/fits_generator/create_data.py
```diff
@@ -25,8 +25,11 @@ def get_proposals(base_directory='.'):
     proposal_list = []
     for dirname, subdirlist, filelist in os.walk(base_directory):
         for name in filelist:
-            if name.endswith('.prop'):
-                proposal_list.append((dirname, name))
+            if name.startswith('.'):
```
stscicrawford: I think this could be compacted down by using `not`:  

So either:
```
if not name.startswith('.'):
   if name.endswith('.prop'):
```
or 
```
if name.ednswith('.prop') and not name.startswith('.'):
```

jwst/fits_generator/create_data.py
```diff
@@ -25,8 +25,11 @@ def get_proposals(base_directory='.'):
...
+                pass
```
stscicrawford: Or if you prefer to stick with this format, `continue` should be used here instead of `pass` as you do just want to ignore these files

stscicrawford:
There currently isn't any unit testing in this package.  Could you add a unit test for `get_proposals` testing that files starting with `.` are ignore?  

Thanks for updating this quickly!

Issue #2273, opened 2018-07-20
MIRI testing for tweakreg step


stscijgbot:
Issue [JP-346](https://jira.stsci.edu/browse/JP-346) was created by Misty Cracraft:

MIRI starting to test tweakreg as part of calwebb_image3.

stscijgbot:
Comment by Misty Cracraft: I just tried a first test of tweakreg and it failed to find any matches. I ran the calwebb_image3 pipeline on a .json file with four images I had run through calwebb_image2 (and successfully tested through source_catalog without running tweakreg). The json file is det_dithered_5stars.json, and the four cal files *5stars_cal.fits are in /user/cracraft/temp.

When the pipeline was running tweakreg, it claimed to find too many stars (I only put 5 in each image), and then it couldn't make any matches, saying it needed at least three points to find the transformation. If these files previously aligned fine without tweakreg to give me a good source_catalog output, I'm assuming the data doesn't actually need tweakreg, but I also don't think it should crash on the data. So, I don't know what the problem is here, the data, something missing in the headers, or something to do with the pipeline itself. I did change the minobj parameter in the .cfg file to be 5, since I only have 5 sources in the image.

 

2018-07-16 15:47:14,498 - stpipe.Image3Pipeline.tweakreg - INFO - Step tweakreg running with args (<ModelContainer>,).
2018-07-16 15:47:14,502 - stpipe.Image3Pipeline.tweakreg - WARNING - /Users/cracraft/anaconda3/envs/jwstdev/lib/python3.5/site-packages/photutils/detection/core.py:97: AstropyDeprecationWarning: "iters" was deprecated in version 3.1 and will be removed in a future version. Use argument "maxiters" instead.
 iters=sigclip_iters)

2018-07-16 15:47:16,704 - stpipe.Image3Pipeline.tweakreg - INFO - Detected 99 sources in det_image_1_MIRIMAGE_F770Wexp1_5stars_cal.fits.
2018-07-16 15:47:16,705 - stpipe.Image3Pipeline.tweakreg - WARNING - /Users/cracraft/anaconda3/envs/jwstdev/lib/python3.5/site-packages/photutils/detection/core.py:97: AstropyDeprecationWarning: "iters" was deprecated in version 3.1 and will be removed in a future version. Use argument "maxiters" instead.
 iters=sigclip_iters)

[...]

PR #2342, opened 2018-08-08
Adjust TPCorr to work with V2, V3 angles in arcsec instead of degrees

mcara:
This PR adresses issue https://github.com/spacetelescope/jwst/issues/2273 by making `TPCorr` work with `V2` and `V3` angles provided in arcseconds instead of the original degrees.

### DIFF
--- a/jwst/transforms/tpcorr.py
+++ b/jwst/transforms/tpcorr.py
@@ -127,7 +127,7 @@ class TPCorr(Model):
     @staticmethod
     def cartesian2spherical(x, y, z):
         """
-        Convert cartesian coordinates to spherical coordinates (in deg).
+        Convert cartesian coordinates to spherical coordinates (in acrsec).
 
         """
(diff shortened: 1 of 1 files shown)


nden:
The comments of the two modified methods say `degrees` when it should be `arcsec` now. 
Alternatively the unit conversion can be done in `evaluate`.
Do we have a regression test for this? If not, is it possible to add one?

mcara:
@nden I fixed docstrings. There is no unit test for TPCorr but I plan to have one as part of the more general data analysis tools package.

nden:
@mcara I was thinking of a regression test so we catch up problems like these earlier.

mcara:
Unfortunately I did not manage to do a regression test (only half-way through). I will finish it after 09/06/2018. Meanwhile, if @hbushouse need testing for `tweakreg` to continue, please consider merging this PR without regression test for now.

PR #2340, opened 2018-08-08
pep8 setup

nden:
Addresses #2329.
This is a (conservative) attempt to define some initial PEP8 rules and is up for discussion. The PEP8 codes enabled here (these will make a build fail) are:
```
E101: indentation contains mixed spaces and tabs
E111: indentation is not a multiple of four
E112: expected an indented block
E113: unexpected indentation
E401: multiple imports on one line
E402: module level import not at top of file
E711: comparison to None should be ‘if cond is None:’

--max-line-length=110

F*: all flake8 codes
Some of the failures reported here are
F401: something is imported but never used
F841: variable is assigned but never used
F821: undefined variable
F811: redefinition of unused variable
```
Definition of all flake8 codes is [here](http://flake8.pycqa.org/en/latest/user/error-codes.html).

If something should be removed from this list please comment below.
Once these are resolved (~1200 messages) we can add more codes.

These involve the following steps:
* [x] ami @jdavies-st
* [x] assign_wcs  - ~@nden~ @sosey to look at two remaining issues
* [ ] associations
* [x] background @jdavies-st
* [x] barshadow  @stscirij 
* [x] combine_1d @philhodge
* [x] coron @jdavies-st
* [x] cube_build @jemorrison
* [x] cube_skymatch @nden
* [x] datamodels @jdavies-st
* [x] darkcurrent @jdavies-st
* [x] dq_init  @stscirij 
* [x] emission @jdavies-st
* [x] engdblog @nden
* [ ] exp_to_source @jdavies-st
* [x] extract_1d @philhodge
* [x] extract_2d @nden
* [x] first_frame @jemorrison
* [ ] fits_generator @stscirij 
* [x] flatfield @philhodge
* [x] fringe (@stscicrawford)
* [x] gain_scale @hbushouse
* [x] group_scale @hbushouse
* [x] guider_cds @hbushouse
* [x] imprint @hbushouse
* [x] ipc @philhodge
* [x] jump @jdavies-st
* [ ] jwpsf @jdavies-st
* [x] last_frame @jemorrison
* [ ] lib @jdavies-st
* [x] linearity @hbushouse
* [x] model_blender  @stsci-hack
* [x] mrs_imatch @nden
* [x] msa_flagopen  @stscirij 
* [x] outlier_detection @jdavies-st
* [x] pathloss     @stscirij 
* [x] persistence @philhodge
* [x] photom @hbushouse
* [x] pipeline @hbushouse
* [ ] ramp_fitting @dmggh
[...]

PR #2346, opened 2018-08-10
fix PEP8 issues in assign_wcs, enable an IFU test again

nden:
#2340 

This actually fixed some real problems discovered by flake8!


jwst/assign_wcs/nircam.py
```diff
@@ -139,7 +139,6 @@ def tsgrism(input_model, reference_files):
         dispx = f.dispx
         dispy = f.dispy
         invdispx = f.invdispx
-        invdispy = f.invdispy
```
nden: @sosey Is it OK to remove this?

sosey: no. it is necessary.

sosey: It's used for the column grism model in WFSS

nden: Oh, I see. But that  is a different function. This one here is `tsgrism` so perhaps can be removed?

sosey: ah - when I clicked it was taking me to the new file without the line, and only the wfss part...I had missed the grayed out text suggesting it was in tsgrism in the  master...yah, you can remove it. I've already removed it my PR for this mode as well.

PR #2347, opened 2018-08-10
add missing fgs image types to footprint calculation

nden:
These were inadvertently dropped in #2327. Addnig them back.


nden:
@stscieisenhamer I'm guessing there's a reason the FGS types are not in associations. This should be pretty straightforward to review.

PR #2349, opened 2018-08-10
Add 2D pathloss and wavelength attributes to SlitModel

jdavies-st:
I believe we had discussed this @nden and @stscirij , and I don't recall the issue that prompted it.  I found these code changes in a `git stash`.

### DIFF
--- a/jwst/datamodels/slit.py
+++ b/jwst/datamodels/slit.py
@@ -36,6 +36,8 @@ class SlitDataModel(model_base.DataModel):
                  relsens=None, area=None,
                  wavelength_pointsource=None, pathloss_pointsource=None,
                  wavelength_uniformsource=None, pathloss_uniformsource=None,
+                 wavelength_pointsource2d=None, pathloss_pointsource2d=None,
+                 wavelength_uniformsource2d=None, pathloss_uniformsource2d=None,
                  **kwargs):
         if isinstance(init, (SlitModel, ImageModel)):
             super(SlitDataModel, self).__init__(init=None, **kwargs)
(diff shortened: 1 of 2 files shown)


jwst/datamodels/slit.py
```diff
@@ -36,6 +36,8 @@ def __init__(self, init=None, data=None, dq=None, err=None,
                  relsens=None, area=None,
                  wavelength_pointsource=None, pathloss_pointsource=None,
                  wavelength_uniformsource=None, pathloss_uniformsource=None,
+                 wavelength_pointsource2d=None, pathloss_pointsource2d=None,
```
stscirij: There aren't actually wavelength_pointsource2d or wavelength_uniformsource2d attributes, the regular wavelength attribute is the corresponding wavelength values for each of the pathloss2d attributes.

jwst/datamodels/slit.py
```diff
@@ -86,17 +88,34 @@ def __init__(self, init=None, data=None, dq=None, err=None,
...
             self.pathloss_uniform_source = pathloss_uniformsource
 
+        if pathloss_pointsource2d is not None:
+            self.pathloss_pointsource2d = pathloss_pointsource2d
+
+        if pathloss_pointsource2d is not None:
+            self.pathloss_pointsource2d = pathloss_pointsource2d
+
+        if wavelength_uniformsource2d is not None:
+            self.wavelength_uniformsource2d = wavelength_uniformsource2d
```
stscirij: Not needed

jwst/datamodels/slit.py
```diff
@@ -135,6 +154,8 @@ def __init__(self, init=None, data=None, dq=None, err=None,
                  int_times=None, barshadow=None,
                  wavelength_pointsource=None, pathloss_pointsource=None,
                  wavelength_uniformsource=None, pathloss_uniformsource=None,
+                 wavelength_pointsource2d=None, pathloss_pointsource2d=None,
+                 wavelength_uniformsource2d=None, pathloss_uniformsource2d=None,
```
stscirij: Again, these 2 aren't needed

jwst/datamodels/slit.py
```diff
@@ -184,20 +205,39 @@ def __init__(self, init=None, data=None, dq=None, err=None,
...
         if wavelength_uniformsource is not None:
             self.wavelength_uniformsource = wavelength_uniformsource
+
         if pathloss_uniformsource is not None:
             self.pathloss_uniform_source = pathloss_uniformsource
+
+        if pathloss_pointsource2d is not None:
+            self.pathloss_pointsource2d = pathloss_pointsource2d
+
+        if pathloss_pointsource2d is not None:
```
stscirij: The second pathloss_pointsource2d was probably supposed to be wavelength_pointsource2d, either way, it isn't needed

jdavies-st:
Thanks Robert!  I'm closing this PR then.

jdavies-st:
I'm adding here that often the attributes for numpy arrays are added to the `foo.py` datamodel in order for it to be instantiated with a passed numpy array.  I think this was an API feature envisioned long ago by the datamodel designers, but I don't think it is used much or at all now.

Would be good to revisit this when we do some infrastructure maintenance, as keeping all the datamodel APIs updated with their schemas involves a lot of boilerplate and a lot of work.  Automating this via the schema might be a better solution.

jdavies-st:
I will file a separate issue for us to think about how we want to make this aspect of datamodels easier to maintain.

PR #2350, opened 2018-08-11
Wavetab

jemorrison:
This pr adds the functionality to the datamodels to have WCS-TAB extension tables. 
In the IFUCubeModel a WAVE-TAB (which is a WCS-TAB) was added and will store the
wavelengths of IFU when a non-linear wavelength sampling is used. 
The pipeline can now write and read these types of datamodels. 

One thing  that seemed a little odd is the WAVE-TAB written to the fits file seemed to be in reverse
order (it started with the largest indices - longest wavelengths first). I was not sure it that is "standard" .  It is read in correctly by the pipeline so maybe it is not an issue.

### DIFF
--- a/jwst/datamodels/schemas/ifucube.schema.yaml
+++ b/jwst/datamodels/schemas/ifucube.schema.yaml
@@ -30,6 +30,10 @@ allOf:
               title: EXTNAME of the extension containing data quality mask
               type: string
               fits_keyword: MASKEXT
+            wave_extension:
+              title: EXTNAME of the extension containing wavelength slice data
+              type: string
+              fits_keyword: WAVEEXT
             roi_spatial:
               title: "[arcsec] Size of ROI in spatial dimension"
               type: number
@@ -71,5 +75,11 @@ allOf:
       fits_hdu: WMAP
       default: 0.0
       ndim: 3
+      datatype: float32
+    wavetable:
+      title: Wavelength value for slices
+      fits_hdu: WCS-TAB
+      default: 0.0
+      ndim: 1
       datatype: float32
 $schema: http://stsci.edu/schemas/fits-schema/fits-schema
--- a/jwst/datamodels/schemas/ifucube.schema.yaml
+++ b/jwst/datamodels/schemas/ifucube.schema.yaml
@@ -30,10 +30,6 @@ allOf:
               title: EXTNAME of the extension containing data quality mask
               type: string
               fits_keyword: MASKEXT
-            wave_extension:
-              title: EXTNAME of the extension containing wavelength slice data
-              type: string
-              fits_keyword: WAVEEXT
             roi_spatial:
               title: "[arcsec] Size of ROI in spatial dimension"
               type: number
@@ -78,8 +74,8 @@ allOf:
       datatype: float32
     wavetable:
       title: Wavelength value for slices
-      fits_hdu: WCS-TAB
-      default: 0.0
-      ndim: 1
-      datatype: float64
+      fits_hdu: WCS-TABLE
+      datatype:
+      - name: wavelength
+        datatype: float64
 $schema: http://stsci.edu/schemas/fits-schema/fits-schema
--- a/jwst/datamodels/schemas/ifucube.schema.yaml
+++ b/jwst/datamodels/schemas/ifucube.schema.yaml
@@ -81,5 +81,5 @@ allOf:
       fits_hdu: WCS-TAB
       default: 0.0
       ndim: 1
-      datatype: float32
+      datatype: float64
 $schema: http://stsci.edu/schemas/fits-schema/fits-schema
--- a/jwst/datamodels/schemas/wcsinfo.schema.yaml
+++ b/jwst/datamodels/schemas/wcsinfo.schema.yaml
@@ -211,6 +211,16 @@ properties:
             fits_keyword: PC3_3
             fits_hdu: SCI
             blend_table: True
+          ps3_0:
+            title: table extnesion name
+            type: string
+            fits_keyword: PS3_0
+            fits_hdu: SCI
+          ps3_1:
+            title: column name for the coordinate array
+            type: string
+            fits_keyword: PS3_1
+            fits_hdu: SCI
           s_region:
             title: spatial extent of the observation
             type: string
--- a/jwst/datamodels/schemas/wcsinfo.schema.yaml
+++ b/jwst/datamodels/schemas/wcsinfo.schema.yaml
@@ -212,12 +212,12 @@ properties:
             fits_hdu: SCI
             blend_table: True
           ps3_0:
-            title: table extnesion name
+            title: Coordinate table extension name
             type: string
             fits_keyword: PS3_0
             fits_hdu: SCI
           ps3_1:
-            title: column name for the coordinate array
+            title: Coordinate table column name
             type: string
             fits_keyword: PS3_1
             fits_hdu: SCI
--- a/jwst/assign_wcs/pointing.py
+++ b/jwst/assign_wcs/pointing.py
@@ -120,7 +120,7 @@ def wcsinfo_from_model(input_model):
     return wcsinfo
 
 
-def fitswcs_transform_from_model(wcsinfo,wavetab=None):
+def fitswcs_transform_from_model(wcsinfo, wavetab=None):
     """
     Create a WCS object using from datamodel.meta.wcsinfo.
     Transforms assume 0-based coordinates.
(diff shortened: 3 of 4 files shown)


jemorrison:
I don't know if this is of interest but I put the test routine I wrote, input data and output data on
/grp/jwst/ssb/morrison/WAVE_TAB_EXAMPLE

jwst/datamodels/schemas/ifucube.schema.yaml
```diff
@@ -30,6 +30,10 @@ allOf:
               title: EXTNAME of the extension containing data quality mask
               type: string
               fits_keyword: MASKEXT
+            wave_extension:
```
jdavies-st: Is there a reason to have a FITS keyword pointing to the FITS extension?  Setting
```
CTYPE3  = 'WAVE-TAB'           / Wavelength from lookup table
PS3_0   = 'WCS-TAB'            / WAVE-TAB extension name
PS3_1   = 'wavelength_column_name'         / Column name of lookup table
```
Should take care of where this FITS extension lives, right?  CTYPE3 tells you there's a WAVE-TAB, PS3_0 tells you where to find it in the FITS file, and PS3_1 indicates the column to find the wavelengths.

jwst/datamodels/schemas/ifucube.schema.yaml
```diff
@@ -72,4 +76,10 @@ allOf:
       default: 0.0
       ndim: 3
       datatype: float32
+    wavetable:
+      title: Wavelength value for slices
+      fits_hdu: WCS-TAB
```
jdavies-st: It might be useful to name this something more specific than WCS-TAB.  Maybe WAVETAB?  Indicate that is a tabular wavelength array.  One could also have tabular RA and DEC arrays.  The mapping of `wavetable <--> WAVETAB` is also more clear then.

jemorrison: I believe the WCS-TAB is the fits standard. It what what the Spectral Fits paper used for the general case. WAVE-TAB is one the of WCS-TAB

jwst/datamodels/schemas/ifucube.schema.yaml
jemorrison: As I understand it this is the fits standard recommended by the Spectral Fits standard paper.

hbushouse: The CTYPEn, PSi_0, and PSi_1 keywords as described by @jdavies-st are what's required by the FITS WCS standard. There's no need for an additional ad hoc keyword like "WAVEEXT". The FITS standard doesn't care about a keyword of that name. The CTYPEn, PSi_0, and PSi_1 keywords should be handled by or bundled with the rest of the WCS-related keywords contained in the header of the IFU cube HDU. As such I would expect them to be handled/created by the underlying gwcs/wcslib functionality.

jwst/datamodels/schemas/ifucube.schema.yaml
hbushouse: The EXTNAME value of "WCS-TAB" is not required by the FITS WCS standard for the xxxx-TAB algorithm. The only statement regarding the EXTNAME of the table HDU is "We recommend strongly that the value chosen for EXTNAME always begin with the four letters WCS-. If this is done, generic programs will recognize the table as part of the WCS portion of the data and will be less likely to separate the table from the rest of the WCS."

So in this case I agree with @jdavies-st that while EXTNAME='WCS-TAB' is perfectly acceptable given the FITS standard recommendations, I would suggest we change it to something a little more distinct from the "WAVE-TAB" value used for the CTYPEn keyword in the data HDU, just to avoid confusion when referring to their respective names. Perhaps something like "WCS-TABLE", as used in the example shown in the FITS WCS paper in Table 11.

jwst/datamodels/schemas/ifucube.schema.yaml
```diff
@@ -72,4 +76,10 @@ allOf:
...
+      default: 0.0
+      ndim: 1
+      datatype: float64
```
hbushouse: These definitions for the ``wavetable`` object look like they describe a single float64 data value. But this FITS HDU will be a BINTABLE, containing 1 row and at least 1 column containing a 1-D vector of wavelength values. So don't we need definitions for the column name, column data type, etc? (as is done with other data models, like PHOTOM, that include a BINTABLE object)

jwst/datamodels/schemas/wcsinfo.schema.yaml
```diff
@@ -211,6 +211,16 @@ properties:
             fits_keyword: PC3_3
             fits_hdu: SCI
             blend_table: True
+          ps3_0:
+            title: table extnesion name
```
hbushouse: I suggest making the keyword comment "Coord table extension name"

jwst/datamodels/schemas/wcsinfo.schema.yaml
```diff
@@ -211,6 +211,16 @@ properties:
...
+            type: string
+            fits_keyword: PS3_0
+            fits_hdu: SCI
+          ps3_1:
+            title: column name for the coordinate array
```
hbushouse: I suggest "Coord table column name"

jwst/datamodels/schemas/ifucube.schema.yaml
jemorrison: I again do not understand if the fits community " recommends strongly that the value chosen for EXTNAME always begin with the four letters WCS-. " why should we have it start with WAVE ?

hbushouse: We aren't recommending starting with "WAVE". We're saying to use the recommended value of EXTNAME in the wavelength table HDU that starts with the recommended prefix "WCS-". We're just saying that the remaining suffix doesn't necessarily have to be "TAB". It could be anything we want and is not limited to the 8 character length of keyword names (this is a keyword **value**, not a name). So we could call it something like "WCS-ifunonlinearwavelengthtable" if we wanted to (although that would be very bad ...). We're just saying that an EXTNAME value of "WCS-TAB" is so similar to the required CTYPEn value of "WAVE-TAB", that there could be confusion when referring to one or the other. So make the EXTNAME value a little more distinct, such as "WCS-TABLE".

jwst/datamodels/schemas/ifucube.schema.yaml
jemorrison: Question. I used the extract1d.schema.yaml as an example because it seemed the closest thing to what I wanted. If it would be better to put it in a table ok. I looked at the photom.schema.yaml
and phometry.schema.yalm and I don't see a BINTABLE object. Maybe I am misreading your comment.
Do I just have to do this is ifucube.schema.yaml
..........
.........
      wavetable:
           title: Wavelength value of IFUCube slices
           fits_hdu: WCS_TABLE
           datatype:
           - name: wavelength
             datatype: float64

jwst/datamodels/schemas/ifucube.schema.yaml
hbushouse: All of the keywords FLUXEXT, ERREXT, ERRTYPE, and MASKEXT were included here because they conform to conventions used in the data products for other IFU instruments (e.g. groundbased IFU's). There's no requirement for them and the only product that includes them is the IFU cubes (we don't put them in any other products). If other IFU groups include a keyword giving the name of the wavelength table, then it's fine with me to include it (but I doubt that's the case, because it seems no one else uses wavelength tables). It's not needed in order to satisfy any FITS standard requirements.

jemorrison: Thanks that makes sense. I got it.

jwst/datamodels/schemas/ifucube.schema.yaml
hbushouse: The extract1d.schema.yaml does in fact include definitions for table column names and data types, e.g.:
```
title: Extracted count rate table
fits_hdu: EXTRACT1D
datatype:
- name: column
  datatype: float64
- name: wavelength
  datatype: float64
- name: background
  datatype: float64
- name: countrate
  datatype: float64
$schema: http://stsci.edu/schemas/fits-schema/fits-schema
```
So following that schema we would define a table containing a single float64 column using:
```
title: Wavelength values
fits_hdu: WCS-TABLE
datatype:
- name: wavelength
  datatype: float64
$schema: http://stsci.edu/schemas/fits-schema/fits-schema
```
The ``ifucube.py`` module also needs updating to include a new entry for the additional table.

Issue #2352, opened 2018-08-13
Remove version from individual files


stscicrawford:
The line referring to the `__version__` of each step or each file can be removed. 

Currently it looks like there are 119 instances of it: 
https://github.com/spacetelescope/jwst/search?q=__version__&unscoped_q=__version__

hbushouse:
All instances of step or other modules setting their own version numbers have now been removed, by the various PR's referenced above.

PR #2355, opened 2018-08-13
Remove version and PEP8 fixes

larrybradley:
This PR:

* Removes version from `tso_photometry` and `source_catalog` steps (#2352)
* PEP8 fixes in both steps (part of #2340)

### DIFF
--- a/jwst/source_catalog/__init__.py
+++ b/jwst/source_catalog/__init__.py
@@ -1,1 +1,3 @@
 from .source_catalog_step import SourceCatalogStep
+
+__all__ = ['SourceCatalogStep']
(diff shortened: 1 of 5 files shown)


stscicrawford:
Actually looks like there is still two left: 
jwst/source_catalog/__init__.py:1:1: F401 '.source_catalog_step.SourceCatalogStep' imported but unused

You will need to add the `__all__ = ['SourceCatalogStep']` step to the `__init__.py` file and the same thing for the TSO_photometry step.

larrybradley:
@stscicrawford Done, thanks!

larrybradley:
@stscicrawford Fixed, thanks!  (my first change didn't seem right :) ).

PR #2356, opened 2018-08-13
Update code to compute primary array name

bernie-simon:
Datamodels has a method named get_primary_array_name that returns the name of the primary data array. Its purpose is to allow the name to be overridden by a subclass, something only a few subclasses do. This update sets the name to the empty string when a datamodel does not have a primary data array. It also updates ndmodel so that it calls get_primary_array_name instead of assuming the name is 'data', as it did before. It also overrides the method in a few subclasses where the schema indicated another array was the primary array.

### DIFF
--- a/jwst/datamodels/irs2.py
+++ b/jwst/datamodels/irs2.py
@@ -6,7 +6,7 @@ __all__ = ['IRS2Model']
 
 class IRS2Model(model_base.DataModel):
     """
-    A data model for the IRS2 refpix reference file.
+    A data model for the IRS2 reference file.
 
     Parameters
     ----------
--- a/jwst/datamodels/irs2.py
+++ b/jwst/datamodels/irs2.py
@@ -6,7 +6,7 @@ __all__ = ['IRS2Model']
 
 class IRS2Model(model_base.DataModel):
     """
-    A data model for the IRS2 refpix rreference file.
+    A data model for the IRS2 refpix reference file.
 
     Parameters
     ----------
--- a/jwst/datamodels/irs2.py
+++ b/jwst/datamodels/irs2.py
@@ -6,7 +6,7 @@ __all__ = ['IRS2Model']
 
 class IRS2Model(model_base.DataModel):
     """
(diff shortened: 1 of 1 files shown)


stscicrawford:
Same pull request as   #2284 but seems to be a weird effect that the changes are not showing up for @bernie-simon and not when doing the pull request.   (The update does appear to already be in the main code, but not in Bernie's -- this seems weird to me).   Any ideas @nden or @stscieisenhamer before merging ?

stscieisenhamer:
Seeing that #2284 has been merged, then any common changes between this pr and that one will not show up, since that is in master and is no longer different.  Is that what you are seeing?

bernie-simon:
> On Aug 14, 2018, at 10:04 AM, Jonathan Eisenhamer <notifications@github.com> wrote:
> 
> Seeing that #2284 has been merged, then any common changes between this pr and that one will not show up, since that is in master and is no longer different. Is that what you are seeing?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

It should have absolutely no affect on this problem.

jdavies-st:
This is a duplicate of #2284.  This was a result of this branch `bernie-simon:new_primary_array` tracking local `master` which had not been rebased from `jwst/master` (i.e. `upstream/master`) with the changes from #2284.  The solution was to locally:
```
git fetch upstream
git checkout new_primary_array
git rebase upstream/master
git push -f origin
```
And now there is no diff between this branch and `jwst/master`.

Closing.

The best practice is when making a new branch for a new feature, always start with:
```
git fetch upstream
git checkout -b my-new-branch upstream/master
```
so that the new branch tracks `upstream/master` and `git diff` shows changes in the branch relative to `upstream/master`.

To see what branch your local branch is tracking (i.e. the comparison branch, the branch your branch has branched from and is showing diffs relative to) is to use
```
git branch -avv
```
This shows all your local branches, what they are currently tracking, and whether they are behind or ahead of that tracking branch.

Issue #1742, opened 2018-02-13
Correct wavelengths for JWST velocity


hbushouse:
The JCCWG has requested the addition of wavelength corrections for the effective velocity of JWST relative to the barycenter for any given exposure. There are 3 components to implementing this:

- [x] Add the keyword VELOSYS to the wcsinfo schema (corresponding KWD ticket is https://jira.stsci.edu/browse/JWSTKD-193) [#2406]
- [x] Update one of the pointing/velocity/time scripts applied during level-1b processing to compute the appropriate value for VELOSYS and populate it [#2359]
- [x] Update the assign_wcs step to compute the wavelength correction and add it as another transform in the WCS object for spectroscopic exposures [#2406]

hbushouse:
[CALWG_doppler_lambda_correction.pdf](https://github.com/STScI-JWST/jwst/files/1721876/CALWG_doppler_lambda_correction.pdf)

hbushouse:
https://jira.stsci.edu/browse/JP-245

nden:
For EXPTYPE `NRS_FIXEDSLIT` and `NRS_MSASPEC` the extract_2d step computes a wavelength zero-point correction and applies it to the computed wavelengths. The zero-point correction also needs to be corrected for the JWST velocity.

hbushouse:
This is a high priority enhancement for B7.2, in order to produce properly calibrated data products.

stscicrawford:
Also related: https://jira.stsci.edu/browse/JP-244

philhodge:
Here are a few comments/questions.
VELOSYS is a FITS keyword for spectral coordinates, so it belongs in an extension header.  I suppose that would be the SCI header for level 1b data, but should VELOSYS also be added to the `MultiSpecModel` and the `SlitDataModel`?
If the INT_TIMES table is populated, the change in (UTC - TDB) from one integration to the next will be proportional to the radial velocity.  Even if there's just one integration, it should be easy to modify the `utc_to_bary` module (for populating the TDB columns in the INT_TIMES table) to have it also compute the radial velocity.
Should there be a header keyword that indicates that the wavelengths have been corrected for radial velocity?  Or can we assume that if S_WCS is 'COMPLETE' that the wavelengths will have been corrected?
This is to be a barycentric velocity correction rather than heliocentric?  I don't think the difference will be detectable, but we do have to call it something, and @hbushouse  did say "relative to the barycenter."

hbushouse:
[...]

PR #2359, opened 2018-08-13
Update utc_to_tdb to also update VELOSYS.

philhodge:
The `utc_to_tdb` module in jwst.lib has been modified to (optionally) compute the radial velocity (m / s) of JWST with respect to the solar-system barycenter, and to assign that value to keyword VELOSYS in the SCI header of the specified FITS file.


jwst/lib/utc_to_tdb.py
```diff
@@ -121,15 +189,12 @@ def find_hdu(fd, extname):
...
         hdr = fd[i].header
         if "EXTNAME" in hdr and hdr["EXTNAME"] == extname_uc:
-            if hdunum is not None:
-                fd.close()
-                raise RuntimeError("There are at least two HDUs with "
-                                   "EXTNAME = {}; there should only be one."
-                                   .format(extname_uc))
-            hdunum = i 
+            hdunum = i
+            break
```
nden: With this `break` added here it looks like this function returns the index of he first extension with EXTNAME found in the file and it doesn't raise an error if there are more than one extensions with this EXTNAME. If so I wonder if
it's not faster to use `fd.index_of(EXTNAME)` instead of this function.

nden:
Part of #1742

jwst/lib/utc_to_tdb.py
philhodge: Yes, I need to take another look at this function.  Originally it was only for finding the INT_TIMES extension, and there shouldn't be more than one.  Now it's also used for finding a SCI extension, and it should not be any problem if there's more than one, though I would be a little surprising.

jwst/lib/utc_to_tdb.py
```diff
@@ -24,77 +24,145 @@
...
+        If True, the radial velocity of JWST will be computed at the
+        middle of the exposure, and the value in meters / second will
+        be assigned to keyword VELOSYS in the first SCI extension.
+
+    delta_t: float
+        This is used only if `update_velosys` is True.  The TDB times
+        corresponding to two UTC times (that are centered on the middle
+        of the exposure and separated by `delta_t` seconds) will be
+        computed, and the radial velocity will be computed from the
+        difference between the two TDB times.
```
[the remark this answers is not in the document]
philhodge: I tried to explain it in the comments, but maybe they could be written more clearly.  A better way to get the velocity would be to simply compute it directly in the `timeconversion` module, but I didn't want to change the calling sequence (to add an EXPMID value) or the number of items that are returned.

Issue #2282, opened 2018-07-24
datamodels uses up file handles, not respecting context managers


jdavies-st:
Datamodels that are opened and then closed using a context manager don't release their file handles.

An example:
```
In [1]: pwd
Out[1]: '/user/jdavies/dms_test_data_probems_mswam'

In [2]: from jwst import datamodels

In [3]: from glob import glob

In [4]: files = glob('*_rate.fits')

In [5]: len(files)
Out[5]: 14

In [6]: for file in files:
   ...:     with datamodels.open(file) as model:
   ...:         pass
   ...:     !lsof -c py | grep fits | wc -l
   ...:     
       7
       9
      11
      13
      15
      17
      19
      21
      23
      25
      27
      29
      31
      33

In [7]: for file in files:
   ...:     with datamodels.open(file) as model:
   ...:         pass
   ...:     !lsof -c py | grep fits | wc -l
   ...:     
      34
      35
      36
      37
      38
      39
      40
      41
      42
      43
      44
      45
      46
      47

In [8]: for file in files:
   ...:     with datamodels.open(file) as model:
   ...:         pass
   ...:     !lsof -c py | grep fits | wc -l
   ...:     
      48
      49
      50
      51
      52
      53
      54
      55
      56
      57
      58
      59
      60
      61
```
And it doesn't seem to be an `astropy.io.fits` issue.  I.e. `fits.open` respects the context manager and closes any open file handles.
```
In [9]: from astropy.io import fits

In [10]: for file in files:
    ...:     with fits.open(file) as hdulist:
    ...:         pass
    ...:     !lsof -c py | grep fits | wc -l
    ...:     
      61
      61
      61
      61
      61
      61
      61
      61
      61
      61
      61
      61
      61
      61
```

drdavella:
Yikes. Could it be an ASDF issue? (🤞it's not...)

jdavies-st:
This is an issue if one doesn't use `datamodels.open()` and instead specifies the datamodal type to be used:
```
In [4]: for file in files:
   ...:     with datamodels.ImageModel(file) as model:
   ...:         pass
   ...:     print(len(proc.open_files()))
   ...:     
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
```

drdavella:
Where could I find a set of files to use for testing whether this is a problem in ASDF?

jdavies-st:
[...]

PR #2360, opened 2018-08-13
Remove dangling pointers to data from datamodels

bernie-simon:
An issue was filed (#2282) that datamodels was not closing file
handles after opening them. After some investigation, it looks like
there are two problems. The first problem is that a fits file is
opened, normally the data in the file is accessed as a memory mapped
array. When the fits file is closed, the handle to the memory map is
omly closed if there is a single reference to the memory map. Normally
this condition is not met. Instead, the memory containing the fits
descriptor is repleased, and some time after this the memory is
garbage collected. When this happens, the file handle tothe memory
mapped array is closed. This causes a problem when a reference to the
fits file descriptor is not released. In that case, the file handle is
not closed and if one is looping over a set of fits files, one file
handle will be leaked for each file.

This fix contains a partial fix for the problem in that it closes some
extra references to the memory mapped data. However, since I have not
located them all, the leak is not yet fixed. The first change was to
add a new method to model_base.py, _drop_arrays, which walks the data
structure and deletes all numpy arrays. The second change is in
from_fits_hdu in fits_support.py, where instead of creating a new copy
of a numpy array before casting it to a new type, it over writes the
original array.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -17,6 +17,7 @@ from astropy.wcs import WCS
 
 from asdf import AsdfFile
 from asdf import yamlutil
+from asdf import treeutil
 from asdf import schema as asdf_schema
 from asdf import extension as asdf_extension
 
@@ -32,7 +33,6 @@ from .history import HistoryList
 from .extension import BaseExtension
 from jwst.transforms.jwextension import JWSTExtension
 from gwcs.extension import GWCSExtension
-
 
 class DataModel(properties.ObjectNode, ndmodel.NDModel):
     """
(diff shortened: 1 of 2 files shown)


jwst/datamodels/model_base.py
```diff
@@ -253,12 +253,29 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
+    def _drop_arrays(self):
+        def _drop_array(d, p):
```
stscieisenhamer: Why the extra argument for a locally defined function that is not used? Suggest removal.

PR #2363, opened 2018-08-14
PEP8 cleanup of datamodels

jdavies-st:
Clean up `flake8` problems pointed out in PR #2340 for datamodels.

- remove lots of unused imports
- remove python 2 code
- remove unused variables; don't use assignments if not needed
- fix bug in `model_base.py::set_fits_wcs()`
- fix bug in `ResolutionModel`
- fix bug in `tests/test_fits.py::test_from_fits()`
- remove two commented-out tests `test_from_fits_to_yaml()` and `test_from_fits_to_json()` that test no-longer-supported functionality
- fix bug in `tests/test_models.py::test_multislit_metadata()` and rename a test that was using the same name
- enable commented-out test `tests/test_schema.py::test_date2()`
- clarify the asserts in `tests/test_schema.py::test_dictionary_like()`
- re-enable and rename `tests/test_schema.py::test_multislit_append_string()`
- remove `util.py::fits_header_name()` which was for python 2 compatibility
- fix bug in `IFUSlicerModel`

Still there are 4 unresolved `flake8` problems:
```
$ flake8 --count --select=F, E101, E111, E112, E113, E401, E402, E711, E722 --max-line-length=110 .
./__init__.py:68:1: F403 'from .wcs_ref_models import *' used; unable to detect undefined names
./__init__.py:122:5: F405 '_defined_models' may be undefined, or defined from star imports: .wcs_ref_models
./schema_editor.py:754:17: F841 local variable 'current_value' is assigned to but never used
./schema_editor.py:1396:29: F841 local variable 'model_field' is assigned to but never used
4
```
The `__init__.py` issues I don't think are problems.  It is because we use `__all__` in `wcs_ref_models`.  @nden, is this OK, or should we be explicit in the import?

The `schema_editor.py` issues are I think two small bugs.  @bernie-simon, could you take a look at those?

### DIFF
--- a/jwst/datamodels/__init__.py
+++ b/jwst/datamodels/__init__.py
@@ -65,7 +65,11 @@ from .trapdensity import TrapDensityModel
 from .trappars import TrapParsModel
 from .trapsfilled import TrapsFilledModel
 from .tsophot import TsoPhotModel
-from .wcs_ref_models import *
+from .wcs_ref_models import (DistortionModel, DistortionMRSModel, SpecwcsModel,
+    RegionsModel, WavelengthrangeModel, CameraModel, CollimatorModel, OTEModel,
+    FOREModel, FPAModel, IFUPostModel, IFUFOREModel, IFUSlicerModel, MSAModel,
+    FilteroffsetModel, DisperserModel, NIRCAMGrismModel, NIRISSGrismModel,
+    WaveCorrModel)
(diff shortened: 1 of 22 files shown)


nden:
@jdavies-st In the `__init__`, could you add an explicit import from `wcs_ref_models`. Just to get rid of F403 and F405. Otherwise looks good. Thanks.

PR #2365, opened 2018-08-14
Updated test for firstframe step.

cracraft:
Added 2 tests, swapped x and y and updated for PEP 8.

### DIFF
--- a/jwst/firstframe/tests/test_firstframe.py
+++ b/jwst/firstframe/tests/test_firstframe.py
@@ -17,7 +17,7 @@ def test_firstframe_set_groupdq():
     ysize = 1024
 
     # create the data and groupdq arrays
-    csize = (1, ngroups, xsize, ysize)
+    csize = (1, ngroups, ysize, xsize)
     data = np.full(csize, 1.0)
     groupdq = np.zeros(csize, dtype=int)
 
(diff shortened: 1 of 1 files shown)


stscicrawford:
Thanks @cracraft !   We will give it a review.  

Can you add a comment here for why you swapped x and y?

cracraft:
I don't think x and y were used in the proper order to define the data cube in the original code. It should have been defined with (nint, ngroups, ysize, xsize), but was originally defined (nint, ngroups, xsize, ysize). It doesn't really matter for this test, but I had to swap the order to use it for tests in which the array size matters, so I wanted to make sure it was correct here as well.

stscicrawford:
@cracraft This looks pretty good to me and I've approved it.   I'm also going to ask @jemorrison  to take a look over it, but thanks for the contribution!

jemorrison:
The switch in the ysize and xsize is correct.

I tested the code and it seems to work correctly.

I actually played around with your code and increased NGROUPS to 3. I think it should still
not touch the first frame if NGROUPS < 4.  If there are only 3 groups then after first frame and
last frame rejection only one frame is left so the ramp fitting can not work. 
So when I modified NGROUPS to 3 - the step was skipped [which I believe is correct]
If NGROUPS = 4 then the step ran as it should. 

Maybe this is overkill Misty - but should be put in a check for NGROUPS=3 ?

cracraft:
I have no problem with putting in a test for 3 frames. I actually have a script that tests both first and last frame (runs one step and then the other) for files of 5 frames down to single frames. Do you want me to go ahead and add in the 3 frame test? It's a simple copy/paste matter. I can do that and re-upload it, or someone else can add it in, whichever is easier. I agree, 3 frames should not run the step, 4 frames and up should flag the first frame.

hbushouse:
The ngroups checking is consistent with #1681, as implemented in #2040, where ``firstframe`` is skipped if ngroups <= 3, and ``lastframe`` is skipped if ngroups < 3.

Issue #2357, opened 2018-08-13
NIRSpec assign_wcs problem with simulated MOS data


stscijgbot:
Issue [JP-359](https://jira.stsci.edu/browse/JP-359) was created by James Muzerolle:

When processing a simulated MOS exposure, the pipeline is not finding the correct location of the 2D spectra (they are off by hundreds of pixels in both dispersion and cross-dispersion directions).  We've tested this successfully using ground test data, so it's not clear why the simulations are having a problem.  It could be that the header is missing important keywords; the simulated data are count rate images, so do not get run through caldetector1, and we prepare them by running a custom script that is supposed to supply any missing keywords.  The file used as input into calwebb_spec2, along with its associated MSA metafile, can be found here:  /grp/jwst/wit4/nirspec/Muzerolle/sim_example/

stscijgbot:
Comment by Nadia Dencheva: There's a fix implemented in the pipelinenow. It will be available in conda-dev tomorrow (Aug 16). Internally, in assign_wcs,  there was a mismatch of slit names and transforms. This is now corrected but I wonder if it fixes the issue completely. So please test it and let me know. This bug slipped in previous testing because we only tested one slitlet. And it doesn't apply to other observing modes.

stscijgbot:
Comment by James Muzerolle: I just tried reprocessing with the dev version (0.10.1a.dev79+gaf19b1ba), I get the exact same result.

It's strange because we have previously tested a different ground test exposure and gotten good results.  That has a "longslit" MSA configuration where every shutter in the same row is opened; we processed it by treating each open shutter as a separate slitlet, so I would not have expected any inconsistency with the simulated data, which has multiple slitlets at different locations across the MSA.  For reference, you can find the ground test exposure results here: /grp/jwst/wit4/nirspec_vault/pipe_testing_tool/data2test_tool_development/MOS_G395M.

stscijgbot:
Comment by Nadia Dencheva: I looked at the MOS_G395M test. It worked "by accident" - The slitlets are in 3rd and 4th quadrant and the metadata file listed them in order - first slitlets in 3rd quadrant , then those in 4th. The current case is more general and quadrants are mixed. it's a good test case.

0.10.1a.dev79+gaf19b1b is the correct version and I see changes in the bounding_box for each slitlet. I also see that they are now matched correctly with their respective transforms. How are you testing it? If you are using extract_2d for the test you should run assign_wcs on the file before that.

[...]

PR #2366, opened 2018-08-14
Fix slits ordering

nden:
Addresses #2357.

This fixes a fairly serious bug in assign_wcs where the matchinf of slits and models in some transforms was not correct. Models are created for each slit in a given quadrant. This order was not preserved in the final output and the original ordering of the slits was kept. This has not shown in testing sofar because the only tests we have are for one open slitlet.

### DIFF
--- a/jwst/assign_wcs/nirspec.py
+++ b/jwst/assign_wcs/nirspec.py
@@ -678,6 +678,7 @@ def slit_to_msa(open_slits, msafile):
     """
     msa = MSAModel(msafile)
     models = []
+    slits = []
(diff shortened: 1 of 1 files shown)


nden:
I am not sure this will fix the issue reported in #2357 completely but I think this should be merged to help the team with testing.

stscicrawford:
If this needs to be merged for testing purposes, then I approve.  Once the fault is identified, it would be good to add a test once #2357 is closed.

nden:
I'll add a test before #2357 is closed.

PR #2371, opened 2018-08-15
fix ordering of cube footprint

nden:
Another fix to the cube footprint. The ordering of the corners was not correct. I ran the regression tests locally now and this is hopefully the last fix. :crossed_fingers: 

Just to clarify we need to write out the footrpint as [RA, DEC, RA, DEC ...] of the corners. They were written as [RA, RA..., DEC, DEC,...]. This PR provides the correct order.

### DIFF
--- a/jwst/cube_build/cube_build_step.py
+++ b/jwst/cube_build/cube_build_step.py
@@ -294,8 +294,7 @@ SHORT,MEDIUM,LONG, or ALL
                 self.spaxel_debug.close()
         for cube in Final_IFUCube:
             footprint = cube.meta.wcs.footprint(axis_type="spatial")
-            update_s_region_keyword(cube, footprint.T)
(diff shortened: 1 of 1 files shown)


jemorrison:
I took these changes and added them to a branch of cube_build that I am working on.
I then ran it over some data and following is printed to the screen.
Is this what you expect ?
I get Update S_REGION to POLYGON ICRS  0.000534738 -0.001707255 0.000534738 -0.000298922 359.999270849 -0.000298922 359.999270849 -0.001707255

nden:
@jemorrison Thanks! I think this looks as expected.

hbushouse:
The output provided above by @jemorrison certainly looks like it's in the correct [(RA, Dec), (RA, Dec), ...] order, at least for a cube that's centered at RA=0. I think we can accept the test results as approval for the PR.

Issue #2329, opened 2018-08-06
pep8 allowed exceptions


stscicrawford:
Update the list of allowed exceptions for pep8 and flake

nden:
Resolved in #2542. A few remaining steps should be fixed - #2552

PR #2372, opened 2018-08-15
Fix PEP8 issues in model_blender

stsci-hack:
Minor cleanup of model_blender step to meet PEP8 standards outlined in #2329 in support of #2340.


nden:
@stsci-hack Some of the issues in #2340 were missed. The latest log is [here]( https://travis-ci.org/spacetelescope/jwst/jobs/416818547#L1795)

PR #2376, opened 2018-08-15
remove version, fix flake8

nden:
#2340 transforms, assign_wcs


jwst/transforms/models.py
```diff
@@ -47,7 +49,7 @@ class GrismObject(namedtuple('GrismObject', ("sid",
                                              "partial_order",
                                              "waverange",
                                              "sky_bbox_ll",
-                                             "sky_bbox_lr",
+                  # Licensed under a 3-clause BSD style license - see LICENSE.rst                           "sky_bbox_lr",
```
jdavies-st: This looks like a mistake?

PR #2377, opened 2018-08-15
Pep8 changes for straylight step

jemorrison:


nden:
A few more issues to address here in order to resolve #2340

```
straylight_step.py:3:1: F401 '..stpipe.cmdline' imported but unused
jwst/straylight/straylight_step.py:6:1: F401 '..datamodels.RegionsModel' imported but unused
jwst/straylight/__init__.py:1:1: F401 '.straylight_step.StraylightStep' imported but unused
``
For the last one you can add `__all__ in  __init__`

jemorrison:
I forgot to check the _init__.py file
In this file there is a __version__ should I take that out ?

nden:
Yes, take out version strings.

PR #2378, opened 2018-08-15
Pep8 changes for lastframe step

jemorrison:


jwst/lastframe/lastframe_sub.py
```diff
@@ -40,7 +38,7 @@ def do_correction(input_model):
     # the final group to 'DO_NOT_USE'
     if sci_ngroups > 2:
         output.groupdq[:, -1, :, :] = \
-            np.bitwise_or(output.groupdq[:,-1,:,:], dqflags.group['DO_NOT_USE'])
+            np.bitwise_or(output.groupdq[:, -1, :, :], dqflags.group['DO_NOT_USE'])
```
jdavies-st: I think this line was fine before.  Slicing is a special case were you don't need space for readability.  If you're going to add space, it's good to have it symmetric.  See https://legacy.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

jdavies-st:
Here's an example of `__all__` from resample's `__init__.py`:
```
from .resample_step import ResampleStep
from .resample_spec_step import ResampleSpecStep

__all__ = ['ResampleStep', 'ResampleSpecStep']
```

Issue #2375, opened 2018-08-15
What is the purpose of flatten_combiners?


drdavella:
The function `jwst.datamodels.schema.flatten_combiners` appears to be causing some failures for specific cases of schema validation via `asdf` and `jsonschema`. More specifically, it causes failures when attempting to use `anyOf` or `oneOf` to specify multiple possible types for a node in the schema when one of the possible types is itself a complex type (such as `array` or `object`).

Before I dig deeper and try to understand if and/or how the function can be fixed, I'm trying to find out why it's necessary at all. Why is it necessary to modify schemas that are used by `asdf` for validation instead of passing them as-is?

FWIW, eliminating all use of `flatten_combiners` fixes the bug I am trying to fix, but I have not yet run the `jwst` test suite with this modification.

bernie-simon:
I believe the point of flattening the schema is to make it easier to traverse. Combiners like anyOf or oneOf are non-physical and we would like the schema to only represent physical quantities, mostly the contents of the FITS file.

drdavella:
Okay, then I'll do some more investigation into this issue. The problem is that in certain cases the original intent of the schema is not being preserved by `flatten_combiners`, which means that some well-formed schemas fail to accept valid files. This might just be an edge case in `flatten_combiners` which needs to be addressed, although I still question the logic of blindly removing valid JSON Schema keywords from schemas.

Alternatively, maybe certain schema constructions that are legal according to the JSON Schema standard are simply not acceptable in the context of JWST data models? But that seems unsatisfying. (cc @sosey)

In any case, it's clear that `flatten_combiners` has some non-trivial effect since removing its use causes a variety of failures in the test suite.

jdavies-st:
Looking at the code, it seems very likely that there are some edge cases.

https://github.com/spacetelescope/jwst/blob/c2f253ebf256cf218bc7a2b0e19151dd577b5fd2/jwst/datamodels/schema.py#L215-L220

But yeah, converting allOf and anyOf to lists seems to be its function, though you say it is not doing it correctly for complex object types?

drdavella:
I'm beginning to think that this is actually a fairly serious bug that affects the validation behavior of multiple schemas in `datamodels`. The tl;dr is that anywhere that `anyOf` is used in a schema, the corresponding field will not be properly validated. The consequence is that when files contain invalid values in these fields, it will not be detected.

[...]

PR #2380, opened 2018-08-16
WIP: Remove use of flatten_combiners

drdavella:
This attempts to address the issues raised in #2375. Currently the tests will not pass, but the PR is intended to shed light on any steps that need to be taken in order to eliminate the destructive modification of schemas by the use of `flatten_combiners`.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -101,7 +101,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
                                              resolver=file_resolver,
                                              resolve_references=True)
 
-        self._schema = mschema.flatten_combiners(schema)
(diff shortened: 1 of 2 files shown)


nden:
The `allOf` seems not to work without `flatten_combiners`, e.g. [here](https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwst/detail/PR-2380/1/pipeline#step-86-log-435)

drdavella:
@nden that's very interesting: it suggests that `flatten_combiners` is used to combine top-level properties from various schemas. Maybe it happens at other levels as well.

It seems like the same result could be achieved without resorting to custom manipulation of schema files, but it would involve some refactoring of schemas.

Issue #2379, opened 2018-08-16
Reference file spec update issues


jaytmiller:
I propose opening a blanket issue for Q&A for the updated reference file spec.

jaytmiller:
Generic Keywords

During the spec update the section on boiler plate keywords either disappeared or became harder to find.  Keywords like AUTHOR, USEAFTER, DATE, etc.   Some sections (AMI, TSO_PHOT) do define them but there are advantages to defining them generically as well if that section was indeed dropped.

jaytmiller:
ASSIGN_WCS  - DISTORTION
NIRCAM is matching on both EXP_TYPE, DETECTOR, CHANNEL, PUPIL and FILTER.   PUPIL looks dominant relative to FILTER.   FILTER appears to be unused.
NIRISS is matching only on EXP_TYPE.

jaytmiller:
ASSIGN_WCS WAVELENGTHRANGE
NIRCAM is matching on EXP_TYPE, PUPIL, MODULE.

jaytmiller:
DQ ARRAY BIT DEFINITIONS
Thus far CRDS has done no checking of DQ_DEF bit definitions other than the extension is present and has the required table and columns.   However,  I noticed in the spec for background subtraction that data DQ and reference DQ are combined with a bitwise or operation.   The implications of the bitwise or is that DQ_DEF values either need to be universally consistent (and hence,  why define them file-by-file?) or more complex combination code is needed to normalize the data and reference DQ arrays to the same bit definitions prior to or'ing.

hbushouse:
Bit definitions contained in the DQ_DEF table extension are automatically remapped onto the standard DQ bit definitions when reference files are loaded into a data model, by calls to the ``dynamic_mask`` function that are included in each reference file data model. Hence they can have unique mappings in each ref file.

jaytmiller:
BARSHADOW
Intro typos.

jaytmiller:
CUBE BUILDING CUBEPAR
retype is named cubepar,  not cubepars, correct spec.

To define checks in CRDS,  spec should literally name extensions, columns, and column type.   (Literal means e.g. don't say GRATING when the file uses DISPERSER.)

MIRI has 3 table extensions with 6 columns. column names are not identical.
NIRSPEC has 2 table extensions with 6 columns with differing names,  3 table extensions with 5 columns with identical names.

[...]

PR #2385, opened 2018-08-16
First updates to the reference files documentation

rizeladiaz:
I am trying to add some missing documentation that was part of the jwst-reffiles.stsci.edu document but was not transferred here. More additions will come later.

### DIFF
--- a/docs/jwst/introduction.rst
+++ b/docs/jwst/introduction.rst
@@ -37,8 +37,18 @@ of each are available at
 The remainder of this document discusses pipeline configuration files and
 gives examples of running pipelines as a whole or in individual steps.
 
-CRDS Reference Files
-====================
+Reference Files
+===============
+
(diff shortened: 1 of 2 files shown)


stscicrawford:
This does start to take care of some the issues raised in #2379

stscicrawford:
Thanks for making this update @rizeladiaz !   These were some of the files I hadn't moved over but reading them through, it does look very useful to have them part of these files

hbushouse:
The file /docs/jwst/reference-file-formats-documentation.rst included in this commit is not showing up anywhere in the as-built docs, because it's not included in any index file anywhere. It should be added to either the /docs/index.rst file, if you want it to appear at the top level along with the introduction and package index, or the /docs/jwst/package_index.rst file, so that it shows up along with all the other step/pipeline level docs. And while we're at it, /docs/jwst/package_index.rst should be updated to remove the old and now obsolete file "ref_file_doc_DMS_653/index.rst", which was compiled a couple years ago just to act as a deliverable to satisfy the level 4 requirement DMS-653. We can keep the actual doc file - just don't include it in the package index. It has old and conflicting information relative what's in the rest of the current docs.

Issue #2268, opened 2018-07-20
datamodels should have a .info() method


jdavies-st:
It is currently difficult to introspect an open `datamodel`.  In fact it is much easier to find out what is in a JWST FITS file by just running the `fitsinfo` at the command line:
```
> fitsinfo nrs1_fs_s2d.fits 
Filename: nrs1_fs_s2d.fits
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU     299   ()      
  1  SCI           1 ImageHDU        40   (1382, 31)   float32   
  2  WHT           1 ImageHDU         9   (1382, 31)   float32   
  3  CON           1 ImageHDU         9   (1382, 31)   int32   
  4  RELSENS       1 BinTableHDU     15   101R x 2C   [D, D]   
  5  SCI           2 ImageHDU        40   (1465, 31)   float32   
  6  WHT           2 ImageHDU         9   (1465, 31)   float32   
  7  CON           2 ImageHDU         9   (1465, 31)   int32   
  8  RELSENS       2 BinTableHDU     15   101R x 2C   [D, D]   
  9  SCI           3 ImageHDU        40   (1451, 36)   float32   
 10  WHT           3 ImageHDU         9   (1451, 36)   float32   
 11  CON           3 ImageHDU         9   (1451, 36)   int32   
 12  RELSENS       3 BinTableHDU     15   101R x 2C   [D, D]   
 13  SCI           4 ImageHDU        40   (1427, 15)   float32   
 14  WHT           4 ImageHDU         9   (1427, 15)   float32   
 15  CON           4 ImageHDU         9   (1427, 15)   int32   
 16  RELSENS       4 BinTableHDU     15   101R x 2C   [D, D]   
 17  ASDF          1 BinTableHDU     11   394824R x 1C   [B]   
```
The `astropy.io.fits` library has an `info()` method for introspecting an `HDUList`.  We should have something similar in datamodels to give a layout of the datamodel - its arrays, tables, if it's a container of slits (such as `MultiSlitModel`) info about each slit, etc.

This `info()` method would make it much easier for users to find what they need in our datamodels.

bernie-simon:
The info method has been added, so I am closing this issue.

jdavies-st:
[...]

PR #2393, opened 2018-08-20
Datamodels has a new info method, similar to astropy fits [#2268]

bernie-simon:
This update adds a new info method that prints the attribute name, size and datatype of each array and the size, column names and datatypes of each table in the model. It also prints the filename, date created, and model typee when these are present.


jwst/datamodels/model_base.py
```diff
@@ -306,6 +307,73 @@ def validate(self):
...
+                        meta_info.append(attribute + ': ' + value)
+            return meta_info
+
+        def get_shape_info(instance):
+            if hasattr(instance, '_coldefs'):
+                nrows = instance.shape[0]
+                ncols = len(instance._coldefs)
+                shape_info = "{}R x {}C".format(nrows, ncols)
+            else:
+                # Display shape in fits order (reversed)
```
hbushouse: I appreciate the desire for consistency with fits.info (and catfits), but I hope the ordering doesn't cause confusion with folks who are used to dealing with data model shapes in python order. I guess we'll find out.

hbushouse:
Fixes #2268

PR #2394, opened 2018-08-20
Fix problems in schema editor uncovered by flake8

bernie-simon:
Flake8 pointed out some unused variables that were the result of coding errors. The changes in this update fixes these problems, as well as a bug in matching the integer and int datatypes/


jdavies-st:
In support of #2340.

Issue #2248, opened 2018-07-16
Improve datamodels.open() for missing DATAMODL


jaytmiller:
When it is defined,  the DATAMODL keyword provides an excellent method of choosing an appropriate model class to construct for any given FITS file.

However,  a sizable number of legacy references and/or data do not define DATAMODL.  In these cases,  particularly for reference files,  the guess made by the datamodels may be poor.

For those instances where datamodels.open() is guessing,  **consider issuing a warning**.   

If the guesses are reasonable for science data,  it may be necessary to differentiate reference files (using REFTYPE?) to avoid spurious warnings.

It may be useful to hint at better practices for opening files in the warning message.   Something like:

WARNING:   DATAMODL keyword not present.  Guessing datamodels class [class.__name__] for [filename].  If you are creating or opening a reference file, consult [documentation link] to help open an explicit datamodels class corresponding to your file.

An alternative approach to adding a warning would be to build in a "better guesser" which can functionally replace the DATAMODL keyword.

stscicrawford:
cc @mfixstsci

bernie-simon:
The warning added to the generic open satisfies the original request, so I am closing this issue.

PR #2401, opened 2018-08-21
Add warning message to datamodels.open when model_type not found [#2248]

bernie-simon:
The generic open function in datamodels first uses the metadata parameter model_type to decide what class to instantiate a model as. If it cannot find it, datamodels falls back on other information to guess the type. If it has to guess, it now issues a warning message with the name of the file (if found) and the class has guessed:
    
    NoTypeWarning: model_type not found. Opening flat.fits as a ImageModel


hbushouse:
I think this is a significant enough change in behavior to warrant an entry in the change log.

drdavella:
This PR was merged even though tests were failing?

bernie-simon:
The failed test had nothing to do with the change. The cause of the failing test was that one test was supposed to throw an error because of a schema error, but the code threw a warning instead. That's the correct behavior, the code is supposed to only throw an error if you specifically tell it to (strict_validation=True). So the test needs to be fixed, in another branch,

Issue #2400, opened 2018-08-21
Final_IFUCube not used?


stscieisenhamer:
There is a reference to Final_IFUCube that does not seem set anywhere else:

https://github.com/spacetelescope/jwst/blob/b7284ab11d582e69adb54282daada66c511092ef/jwst/cube_build/cube_build_step.py#L282

hbushouse:
Related to #2397

stscieisenhamer:
Resolved by #2402

Issue #2397, opened 2018-08-21
cube_build regression test errors


hbushouse:
A couple of regression tests are failing in the ``cube_build`` step:
https://boyle.stsci.edu:8081/job/RT/job/JWST/231/testReport/junit/

Specifically, test_mrs_spec3.test_spec3_pipeline1 is failing with:
```
    # If single = True: map each file to output grid and return single mapped file
    #to output grid
    # this option is used for background matching and outlier rejection
            if self.single:
                self.output_file = None
                cube_container = thiscube.build_ifucube_single()
                self.log.info("Number of Single IFUCube models returned %i ",
>                             len(Final_IFUCube))
E               NameError: name 'Final_IFUCube' is not defined

../../jwst/cube_build/cube_build_step.py:282: NameError
```

test_nrs_ifu_spec2.test_nrs_ifu_spec2 is failing with:
```
        if has_exceptions and self.fail_on_exception:
            raise RuntimeError(
>               'One or more products failed to process. Failing calibration.'
            )
E           RuntimeError: One or more products failed to process. Failing calibration.

../../jwst/pipeline/calwebb_spec2.py:102: RuntimeError

2018-08-20 20:21:02,704 - stpipe.Spec2Pipeline.cube_build - INFO - Output Name jw95175001001_02104_00001_nrs1_s3d.fits
2018-08-20 20:21:10,051 - stpipe.Spec2Pipeline.cube_build - INFO - Mapping each NIRSPEC slice to sky, this takes a while for NIRSPEC data
Traceback (most recent call last):
  File "/data1/jenkins/workspace/RT/JWST/jwst/pipeline/calwebb_spec2.py", line 90, in process
    asn.filename
  File "/data1/jenkins/workspace/RT/JWST/jwst/pipeline/calwebb_spec2.py", line 280, in process_exposure_product
    cube = self.cube_build(input)
  File "/data1/jenkins/workspace/RT/JWST/jwst/stpipe/step.py", line 400, in run
    step_result = self.process(*args)
  File "/data1/jenkins/workspace/RT/JWST/jwst/cube_build/cube_build_step.py", line 286, in process
    result = thiscube.build_ifucube()
  File "/data1/jenkins/workspace/RT/JWST/jwst/cube_build/ifu_cube.py", line 242, in build_ifucube
    self.map_detector_to_spaxel(this_par1, this_par2, self.spaxel)
[...]

PR #2402, opened 2018-08-21
Final ifu cube error

jemorrison:
Fixed an incorrect log.info statement 
The determination of x,y for NIRSPEC data was incorrectly commented out in the last pr. Fixed t.


jemorrison:
this address #2397 and #2400

PR #2407, opened 2018-08-22
Fix bug in flatten_combiners

drdavella:
This is a proposed fix for the issue described in #2375. It is a bit rough at the moment, and will need to be updated to properly account for `oneOf` as well. It also could use some more thorough tests.

Also, with this fix, it seems like the name `flatten_combiners` is no longer appropriate. The real purpose of this function seems to be to merge external schema file references into a single schema. So maybe `merge_schemas` would be more appropriate.

cc @sosey

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -102,7 +102,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
                                              resolver=file_resolver,
                                              resolve_references=True)
 
-        self._schema = mschema.flatten_combiners(schema)
+        self._schema = mschema.merge_property_trees(schema)
 
         # Provide the object as context to other classes and functions
         self._ctx = self
@@ -585,7 +585,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
         new_schema : schema tree
         """
         schema = {'allOf': [self._schema, new_schema]}
-        self._schema = mschema.flatten_combiners(schema)
+        self._schema = mschema.merge_property_trees(schema)
         self.validate()
         return self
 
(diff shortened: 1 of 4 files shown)


drdavella:
Also I'd like to mention that the amount of time spent on diagnosing and fixing this issue is not accurately represented by the diff 😬.

drdavella:
The failures coming from `jwst/datamodels/tests/test_schema.py` are on `master` as well. It seems like it must have been from a very recent change?

drdavella:
@jdavies-st I believe it should, but a test should be added. I think I need to add `oneOf` to the change in `blendrules.py` as well. I don't believe any schemas are currently using `oneOf`, though.

jdavies-st:
Right, no schemas are currently using `oneOf`.  We've generally been using `enum` in our schemas.  Would have to go through and see if `oneOf` would be more concise in some of those cases, but that's a separate issue.

drdavella:
Any thoughts on renaming this function? Maybe `resolve_and_flatten_schema`? It's verbose but descriptive...

jdavies-st:
The parameterized test looks good to me!  I would be happy with a more descriptive function name as well, and yours is very clear, though perhaps a cleanup of the docstring to indicate what it actually does as well?

sosey:
no schemas are currently using oneOf.. but it makes sense that the new wavelengthrange schema  use it, since datamembers should only belong to either the slit or wfss definition but not either. Using  anyOf might allow more backwards compatibility with the old code, but the old code should be pulled with the CRDS context it was valid for, which is probably the recommended way to go?

I think your new name is fine, the only other option I could think of is `explode_schema` :smiling_imp:

drdavella:
I actually settled on `merge_property_trees` and (hopefully) improved the docstring. Any feedback on the name and documentation is welcome.

jdavies-st:
Thanks @drdavella!  Docstring is *much* better.

Feel free to merge.

PR #2408, opened 2018-08-22
Fix failing schema tests caused by #2401

drdavella:


hbushouse:
So now that #2410 has been merged, is this PR no longer needed?

bernie-simon:
Yes, both fix the same error, but IMHO #2410 fixes it better.

PR #2411, opened 2018-08-22
Add a calimage3 regression test that includes tweakreg

hbushouse:
Adds another calimage3 regression test that has ``tweakreg`` turned on. The test data use a 6-point dither pattern (3 primary, plus 2 sub-dithers at each primary), but all have the same WCS pointing. ``tweakreg`` is used to successfully recover the actual pointing offsets in the simulated images, using a larger than normal search radius. The data are simulated as a NIRCam GOODS-S field. Input and comparison test data files have been copied to the appropriate test directory on jwcalibdev.


hbushouse:
Jenkins failure is due to bug fixed by #2410

hbushouse:
That's your job to fix.

PR #2415, opened 2018-08-23
Second pass at pep8 standards

jemorrison:
This is a second pass at making the code follow the pep8 standards.  I changed all variable names so they did not have capital letters in their names. In addition @nden pointed out from the last review that I was passing 'self' to an outside routine not part of the class.  My original intention was to not have such long class file with so many routines - I found it a bit hard to keep up. But passing 'self' to outside a class routine is not standard and caused confusion. There is still one (or two)  case (s)  this is still happening - cube_cloud.py but that is time consuming part of the code that I am working on re-writing as soon as I get the code all pep8 standard cleaned up. I pulled in some routines from cube_build_wcs_util.py and made them part of ifu_cube.py which defines the class IFUCubeData.

### DIFF
--- a/jwst/cube_build/cube_build_step.py
+++ b/jwst/cube_build/cube_build_step.py
@@ -6,7 +6,8 @@ from . import ifu_cube
 from . import data_types
 from ..assign_wcs.util import update_s_region_keyword
 
-__all__ = ["CubeBuildStep", "read_user_input"]
+
+__all__ = ["CubeBuildStep"]
 
 class CubeBuildStep (Step):
     """
@@ -18,9 +19,8 @@ class CubeBuildStep (Step):
 
     spec = """
          channel = option('1','2','3','4','ALL','all',default='ALL') # Options: 1,2,3,4,or All
-         band = option('SHORT','MEDIUM','LONG','ALL','short','medium','long','all',default='ALL') # Options: \
-SHORT,MEDIUM,LONG, or ALL
-         grating   = option('PRISM','G140M','G140H','G235M','G235H',G395M','G395H','ALL','all',default='ALL')  # Options: PRISM,G140M,G140H,G235M,G235H,G395M,G395H, or ALL
+         band = option('SHORT','MEDIUM','LONG','ALL','short','medium','long','all',default='ALL') # Options: SHORT,MEDIUM,LONG, or ALL
+         grating   = option('PRISM','G140M','G140H','G235M','G235H',G395M','G395H','ALL','all',default='ALL') # Options: PRISM,G140M,G140H,G235M,G235H,G395M,G395H, or ALL
          filter   = option('CLEAR','F100LP','F070LP','F170LP','F290LP','ALL','all',default='ALL') # Options: CLEAR,F100LP,F070LP,F170LP,F290LP, or ALL
          scale1 = float(default=0.0) # cube sample size to use for axis 1, arc seconds
          scale2 = float(default=0.0) # cube sample size to use for axis 2, arc seconds
@@ -131,7 +131,7 @@ SHORT,MEDIUM,LONG, or ALL
 
         self.pars_input['filter'] = []   # input parameter
         self.pars_input['grating'] = []  # input parameter
-        read_user_input(self)  # see if options channel, band,grating filter are set
+        self.read_user_input()  # see if options channel, band,grating filter are set
                                # is they are then self.output_type = 'user'
                                # if they are filling par_input with values
 #________________________________________________________________________________
@@ -193,7 +193,7 @@ SHORT,MEDIUM,LONG, or ALL
             'offset_list': self.offset_list}
 
 # shove the input parameters in to pars_cube to pull out ifu_cube.py
-# these parameters are related to the IFUCube
+# these parameters are related to the building a single ifucube_model
         pars_cube = {
             'scale1': self.scale1,
             'scale2': self.scale2,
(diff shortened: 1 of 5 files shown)


jwst/cube_build/cube_build_step.py
```diff
@@ -18,9 +19,8 @@ class CubeBuildStep (Step):
 
     spec = """
          channel = option('1','2','3','4','ALL','all',default='ALL') # Options: 1,2,3,4,or All
```
jdavies-st: Is there a reason to list all capitalizations of each spec?  Shouldn't it just allow a single one and make it consistent?  I.e. lowercase?

That said, I notice that our current error message is not very good in helping users. I.e. if a wrong option is specified, we currently do not let the user know what the correct option is.  That should be improved I think.  I will file an issue.
```
In [1]: from jwst.cube_build import CubeBuildStep

In [2]: step = CubeBuildStep(band='foo')
ValidationError                           Traceback (most recent call last)
<ipython-input-3-39016978c84c> in <module>()
----> 1 step = CubeBuildStep(band='foo')

~/dev/jwst/jwst/stpipe/step.py in __init__(self, name, parent, config_file, _validate_kwds, **kws)
    286             spec = self.load_spec_file()
    287             kws = config_parser.config_from_dict(
--> 288                 kws, spec, root_dir=dirname(config_file or ''))
    289 
    290         if name is None:

~/dev/jwst/jwst/stpipe/config_parser.py in config_from_dict(d, spec, root_dir)
    194 
    195     if spec:
--> 196         validate(config, spec, root_dir=root_dir)
    197     else:
    198         config.walk(string_to_python_type)

~/dev/jwst/jwst/stpipe/config_parser.py in validate(config, spec, section, validator, root_dir)
    255 
    256         if len(messages):
--> 257             raise ValidationError('\n'.join(messages))
    258     finally:
    259         config.main.configspec = orig_configspec

ValidationError: Config parameter 'band': the value "foo" is unacceptable.
```

jwst/cube_build/cube_build_wcs_util.py
```diff
@@ -16,241 +12,9 @@
...
-        w_scale = self.scalew
-
-    scale = [a_scale, b_scale, w_scale]
-    return scale
-#_______________________________________________________________________
-
 
 #********************************************************************************
-def find_footprint_MIRI(self, input, this_channel, instrument_info):
+def find_footprint_MIRI(input, this_channel, instrument_info,coord_system):
```
jdavies-st: need a space before `coord_system`.

jwst/cube_build/cube_build_wcs_util.py
```diff
@@ -312,7 +76,7 @@ def find_footprint_MIRI(self, input, this_channel, instrument_info):
     return a_min, a_max, b_min, b_max, lambda_min, lambda_max
 
 #********************************************************************************
-def find_footprint_NIRSPEC(self, input, flag_data):
+def find_footprint_NIRSPEC(input, flag_data,coord_system):
```
jdavies-st: again, space before `coord_system`.

jemorrison:
is it standard to make them lower case for parameter options ?
I have assumed if it is lower case then I convert it to upper case and use upper case in dictionaries. 
It lower case is standard then I will adjust my dictionary

jdavies-st:
Btw, I'm not sure why the jenkins build has failed.  It looks to be unrelated to this PR.

jdavies-st:
I think lowercase is standard for all the steps and for python in general, but I haven't surveyed all our steps.  @hbushouse?

As for the `spec` options though, you should pick one that users can use.  What you use internally is up to you, though again, lowercase is standard.  Uppercase in python is usually reserved for module-level variables, i.e. variables that are defined within a module that are outside of any function or class definitions.  Obviously for FITS keywords and keyword values that are uppercase, one has to deal with those as they are.

jdavies-st:
Anyway, I don't think these are showstoppers.  Feel free to sort out those `spec` options at some future date.  Just a comment here as I noticed it here.

jemorrison:
I changed all the parameter names to lower case - thought I might as well do it now -
and now I get this error - is ALL the default some place. 

:\raceback (most recent call last):
  File "/Users/morrison/git/jwst/scripts/strun", line 14, in <module>
  File "/Users/morrison/git/jwst/jwst/stpipe/step.py", line 162, in from_cmdline
  File "/Users/morrison/git/jwst/jwst/stpipe/cmdline.py", line 289, in step_from_cmdline
    just_the_step_from_cmdline(args, cls)
  File "/Users/morrison/git/jwst/jwst/stpipe/cmdline.py", line 250, in just_the_step_from_cmdline
    raise ValueError(str(e))
ValueError: Config parameter 'band': the value "ALL" is unacceptable.
Config parameter 'filter': the value "ALL" is unacceptable.
Config parameter 'grating': the value "ALL" is unacceptable.

nden:
@jemorrison These are all good corrections. Can you please make sure the things listed in the Travis log for  #2340 are fixed. This will allow merging that PR and it will be easier to follow after that. The latest log is [here](https://travis-ci.org/spacetelescope/jwst/jobs/419374436). Note that there are also things listed there for rscd, reset and first_frame. I think those ar the most important issues that need to be fixed first. Feel free to merge when you think it's ready.

jdavies-st:
@jemorrison , you may have to edit the default config file for cube_build in `jwst/pipeline/cube_build.cfg`, as those are still set to `ALL`.

As for `ra-dec`, `world` may be better, but perhaps @nden has an opinion?

Looks good!

Issue #2395, opened 2018-08-20
resample: points in dimension 0 must be strictly ascending


stscieisenhamer:
# Issue
In working with recent SDP data on another issue, #2338, in trying to reproduce with current master gives ultimately the following error. The data has been placed in
```
/grp/jwst/ssb/eisenhamer/i2338
```
In hopes the error could have been mitigated, a full run from `detector1` on up is documented in the notebook `process.ipynb`. Note that this is a new error since milestone 0.10.

Full log:
```
2018-08-20 15:42:41,786 - stpipe.Spec3Pipeline - INFO - Spec3Pipeline instance created.
2018-08-20 15:42:41,788 - stpipe.Spec3Pipeline.mrs_imatch - INFO - MRSIMatchStep instance created.
2018-08-20 15:42:41,790 - stpipe.Spec3Pipeline.outlier_detection - INFO - OutlierDetectionStep instance created.
2018-08-20 15:42:41,791 - stpipe.Spec3Pipeline.resample_spec - INFO - ResampleSpecStep instance created.
2018-08-20 15:42:41,794 - stpipe.Spec3Pipeline.cube_build - INFO - CubeBuildStep instance created.
2018-08-20 15:42:41,796 - stpipe.Spec3Pipeline.extract_1d - INFO - Extract1dStep instance created.
2018-08-20 15:42:41,797 - stpipe - INFO - Hostname: inanna2.stsci.edu
2018-08-20 15:42:41,798 - stpipe - INFO - OS: Darwin
2018-08-20 15:42:41,843 - stpipe.Spec3Pipeline - INFO - Step Spec3Pipeline running with args ('output/jw98010-o001_20180806t184118_spec3_001_asn.json',).
2018-08-20 15:42:42,051 - stpipe.Spec3Pipeline - INFO - Prefetching reference files for dataset: 'jw98010001001_02104_00001_mirimage_cal.fits' reftypes = ['cubepar', 'drizpars', 'extract1d', 'resol']
2018-08-20 15:42:42,055 - stpipe.Spec3Pipeline - INFO - Prefetch for CUBEPAR reference file is 'N/A'.
2018-08-20 15:42:42,056 - stpipe.Spec3Pipeline - INFO - Prefetch for DRIZPARS reference file is '/Users/eisenham/Downloads/crds-cache/references/jwst/miri/jwst_miri_drizpars_0001.fits'.
[...]

PR #2422, opened 2018-08-24
Fix bug in resample_spec wavelength reverse lookup table

jdavies-st:
When `resample_spec` creates an output frame, the wavelengths are done via tabular interpolation.  Both the forward transform (pixels to wavelength) and reverse transform (wavelength to pixels) need to be defined.  This fixes a bug in the reverse where for MIRI LRS fixed slit data, the wavelength is descending as a function of pixel index.  For `scipy.interpolate`, which is the underlying interpolation method used in the tabular models, the interpolation points need to be strictly ascending.

Resolves issue #2395.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -160,6 +160,13 @@ refpix
 resample
 --------
 
+- Fixed spectral resampling so the 2D output for MIRI LRS and NIRSpec MSA
+  has the correct orientation and a dispersion that matches the input, i.e.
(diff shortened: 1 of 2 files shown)


jdavies-st:
If we re-enable resampling for MIRI LRS FIXED SLIT in the spec2 pipeline, then this will be covered by the nightly regression tests.  Currently, it is not covered.

Added an entry to the changelog.

PR #2424, opened 2018-08-24
Enable resample_spec on MIR_LRS-FIXEDSLIT data in spec2 pipeline

jdavies-st:
The spec2 pipeline has not had `resample_spec` run on MIRI fixed slit data for almost 2 years.  Now that `resample_spec` can be reliably run on MIRI data, we can now re-enable it, giving us test coverage for this mode as well.

Cleaned up the spec2 pipeline a bit to remove two unnecessary datamodel `copy()` operations and changed some variable names for clarity.

A potential bug was found in the following operations:
```
input.close()

return input
```
Nothing was done with the returned datamodel, but if `process()` did try to do something with it, it would (I believe) fail.  So I have removed the `close()` line.

### DIFF
--- a/jwst/pipeline/calwebb_spec2.py
+++ b/jwst/pipeline/calwebb_spec2.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 from collections import defaultdict
 import os.path as op
 import traceback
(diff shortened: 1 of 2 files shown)


jwst/pipeline/calwebb_spec2.py
```diff
@@ -277,29 +273,22 @@ def process_exposure_product(
...
+            self.save_model(rectified_result[0], 's3d')
 
         # Extract a 1D spectrum from the 2D/3D data
-        self.extract_1d.suffix = 'x1d'
         if tso_mode:
             self.extract_1d.suffix = 'x1dints'
-        x1d_output = self.extract_1d(x1d_input)
+        else:
+            self.extract_1d.suffix = 'x1d'
+        x1d_result = self.extract_1d(rectified_result)
```
hbushouse: I don't see a path to ``extract_1d`` for data sets that don't get rectified. Right now the only exp_type's that produce a "rectified_result" are NIRSpec fixed-slit and MSA, MIRI LRS fixed-slit, MIRI MRS, and NIRSpec IFU. This leaves out exposure types like MIR_LRS-SLITLESS, NRC_WFSS, NIS_WFSS, NIS_SOSS, and NRS_BRIGHTOBJ, all of which don't get resampled, but do need to go through ``extract_1d`` (in unresampled form).

jdavies-st:
Thanks for catching that @hbushouse.  I think we're fixed now.

jdavies-st:
Could you do one verification of this change?

Run the spec2 pipeline with `resample` and/or `cube_build` turned off and verify that the output **_cal** file written out is actually a cal file?  I want to make sure that since the explicit `model.copy()` was removed, that it's not modifying in place within the steps once it passes the _cal onto `extract_1d`.  I'm pretty sure since `resample`, `cube_build` and `extract_1d` all return models of different types, that it cannot modify in the input in place. I.e. any copying necessary should already be taking place within the step if needed.

I didn't get to test this aspect of that final commit while on vacation.  No VPN.

Issue #2341, opened 2018-08-08
NIRSpec: calwebb_spec2 crashing with ALLSLITS file


stscijgbot:
Issue [JP-357](https://jira.stsci.edu/browse/JP-357) was created by Maria Pena-Guerrero:

Hello,

I am running a Fixed Slit ALLSLITS file through calwebb_spec2 and the pipeline crashes at the flat_field step. It seems that the file is read as having a model of type CubeModel, which is wrong for an EXPT_TYPE=NRS_FIXEDSLIT. I wonder if this actually an error with calwebb_detector1, because the output should not be a multiple integration file for the ALLSLITS case.

I placed the initial file I used for caldetector1, as well as the final output that I use for calwebb_spec2 at:

/grp/jwst/wit4/nirspec/penaguerrero/allslits_data

The files are called FS235M_f170lp_NRS1_uncal.fits and FS235M_f170lp_gain_scale_NRS1.fits.

Here is the full traceback:

strun calwebb_spec2.cfg /Users/pena/Documents/PyCharmProjects/nirspec/pipeline/build7.1/part2/FS_ALLSLITS/G235M_F170LP/491_results/gain_scale_NRS1.fits 

2018-08-08 12:28:16,723 - stpipe.Spec2Pipeline - INFO - Spec2Pipeline instance created.

2018-08-08 12:28:16,724 - stpipe.Spec2Pipeline.srctype - INFO - SourceTypeStep instance created.

2018-08-08 12:28:16,725 - stpipe.Spec2Pipeline.pathloss - INFO - PathLossStep instance created.

2018-08-08 12:28:16,726 - stpipe.Spec2Pipeline.imprint_subtract - INFO - ImprintStep instance created.

2018-08-08 12:28:16,727 - stpipe.Spec2Pipeline.bkg_subtract - INFO - BackgroundStep instance created.

2018-08-08 12:28:16,728 - stpipe.Spec2Pipeline.barshadow - INFO - BarShadowStep instance created.

2018-08-08 12:28:16,729 - stpipe.Spec2Pipeline.straylight - INFO - StraylightStep instance created.

2018-08-08 12:28:16,730 - stpipe.Spec2Pipeline.flat_field - INFO - FlatFieldStep instance created.

2018-08-08 12:28:16,731 - stpipe.Spec2Pipeline.photom - INFO - PhotomStep instance created.

[...]

PR #2429, opened 2018-08-27
Explicitly define suffixes for the gain_scale output

stscieisenhamer:
Ready-to-test #2341

The actual issue in #2341 arose from confusion over what output was being produced when the `gain_scale_step` was requested to save its data. Since the step is called twice, once on the integrated ramp, and once on the non-integrated ramp ("ints"), but without any other file specification, caused the second call to use the same suffix as the first, overwriting the first results.

`calwebb_detector1` is modified to explicitly defines two separate suffixes for the two calls, ensuring that the second does not overwrite the first.

### DIFF
--- a/jwst/pipeline/calwebb_detector1.py
+++ b/jwst/pipeline/calwebb_detector1.py
@@ -132,11 +132,13 @@ class Detector1Pipeline(Pipeline):
             input, ints_model = self.ramp_fit(input)
 
         # apply the gain_scale step to the exposure-level product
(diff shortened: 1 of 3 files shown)


stscieisenhamer:
Though the gain_scale is the "final" result, the reasons for ensuring that the naming works correctly for this particular step are:
- In the future, this step may not be the final step. Yes, that is probably not going to happen, but is still a possibility.
- If one has specifically requested that gain_scale save its results, it really should do so with correct naming.

Issue #2003, opened 2018-05-08
Fix and/or remove BaseExtension in datamodels


drdavella:
The `BaseExtension` class provided by the `datamodels` subpackage does not appear to be used at all. Can it be removed?

If it is actually useful, then it should be registered as a `setuptools` entry point in the same way that the extension for `transforms` is registered:
https://github.com/STScI-JWST/jwst/blob/0a442bb603075ebb4774fb3a7ec2e30360947851/setup.py#L141

In either case, it is no longer necessary to explicitly pass other extensions such as `GWCSExtension` and `JWSTExtension` to calls to `asdf.open` and `asdf.AsdfFile` since they are now automatically detected as plugins by ASDF.

PR #2430, opened 2018-08-28
Removed BaseExtension class, it was not being used

bernie-simon:
And I removed the mention of it in other files.


stscicrawford:
Closes #2003

Issue #2174, opened 2018-06-20
Rule: Level 2: NIRSpec internal lamp


stscieisenhamer:
# Issue
NIRSpec lamp exposures are to get modified Level2 processing. As such, a separate association type will be required to distinguish from normal Level2 processing. Lamp exposures only need to be distinguished by FILTER == OPAQUE. This can occur for any EXP_TYPE.

# Resources
- #1782 
- #1188 

# Implementation
Create a new association type, `spec2-nrslamp`.

## Constraints
- FILTER == OPAQUE

## To Do
- [x] Inform SDP of new type and how to handle.
- [x] documentation
- [x] tests
    - pool: sdp/jw96215_20180602T170215
- [x] code
    - [x] Re-implement rule
    - [x] Disallow `filter == 'opaque'` in `Constraint_Mode`
    - [x] Remove duplicate `program` constraint between `Constraint_Base` and `Constraint_Mode`.
- [x] tests
- [x] documentation

stscieisenhamer:
With filing of [jsdp-695](https://jira.stsci.edu/browse/JSDP-695), this issue is resolved.

PR #2449, opened 2018-09-04
Rename association type `nrslamp-spec2` for consistency.

stscieisenhamer:
More for #2174, decided to rename the association type to match the configuration file naming.


docs/jwst/associations/jwst_conventions.rst
```diff
@@ -60,4 +60,4 @@ The current association types are:
   * `tso3`: Intended for `calwebb_tso3` processing
   * `image2`: Intended for `calwebb_image2` processing
   * `spec2`: Intended for `calwebb_spec2` processing
-  * `spec2-nrslamp`: Intended for `calwebb_spec2_nrslamp` processing
+  * `nrslamp-spec2`: Intended for `calwebb_nrslamp_spec2` processing
```
hbushouse: A really minor nit, but do the other asn type names also use hyphens, as opposed to underscores? The config file names use underscores.

stscieisenhamer: I believe the file naming conventions for association names require that this be a hyphen since this all goes in the `type` field of the name:

```
jw<ProgramID>-<AC_ID>_<GT-DATETIME>_<type>_<mmm>_asn.json
```

hbushouse: Ah, good point. 👍

PR #2456, opened 2018-09-04
Don't resample _calints data in Spec2Pipeline

### LINKED ISSUE #2455: Resampling multi-integration products
Since #2424 was merged, which turns on calls to `resample_spec` in the `calwebb_spec2` pipeline for MIRI LRS fixed-slit exposures, the regression test "test_miri_lrs_slit_1b" has been throwing an error. The reason is that the input for the test is a multi-integration data set (_rateints product), which uses a `CubeModel` data model for the 3-D data and the `resample_spec` step doesn't know what to do with a `CubeModel`. This issue raises a couple different questions.

1) Most of the time we only encounter 3-D spectral data for TSO-mode exposures and TSO data don't get resampled. So the simplest approach to fixing the current issue would be to just not call `resample_spec` when someone's running 3-D data through `calwebb_spec2`. Technically, however, the MIRI LRS fixed-slit mode is not a TSO mode and in fact the APT doesn't even allow a user to mark it as TSO (only MIRI LRS slitless can be marked as TSO). But it is reasonable to assume that a user may at some point intentionally want to process the multi-integration form of MIRI LRS fixed-slit data (or some other mode, like NIRSpec fixed-slit) in order to investigate the per-integration data. So in that case we should NOT simply disable resampling for any 3-D spectral data. `resample_spec` should instead be enhanced to know how to handle a `CubeModel`, resampling each individual "slice" of the 3-D cube and return some kind of 3-D resampled product.

2) Recent changes to stpipe and ASN infrastructure that determine product naming based on whether an exposure is a TSO type may be too restrictive. For example, in the case of the regression test mentioned above, the MIRI LRS fixed-slit data would be regarded as NOT TSO and hence the determination of the output product name would want to make it a _cal product, instead of the proper _calints product for 3-D data. The infrastructure for determining "ints" product types (for both imaging and spectral modes) should not simply rely on whether or not a given exposure was taken in a TSO mode, but should instead rely on the form of the data, i.e. is it a 2-D `ImageModel` or a 3-D `CubeModel`? If the former, the output is a _cal product, and if the latter, the output is a _calints product.

jdavies-st:
The `resample` step only works on 2D images, not 3D cubes.  So multiple integration cubes should not be resampled in our Spec2Pipeline.  Adds a check for this to the pipeline and step.

Fixes #2455


jwst/pipeline/calwebb_spec2.py
```diff
@@ -273,7 +273,7 @@ def process_exposure_product(
         # Produce a resampled product, either via resample_spec for
         # "regular" spectra or cube_build for IFU data. No resampled
         # product is produced for time-series modes.
-        if exp_type in ['NRS_FIXEDSLIT', 'NRS_MSASPEC', 'MIR_LRS-FIXEDSLIT']:
+        if exp_type in ['NRS_FIXEDSLIT', 'NRS_MSASPEC', 'MIR_LRS-FIXEDSLIT'] and not tso_mode:
```
hbushouse: I'm not sure `tso_mode` will be set for instances like the multi-integration version of a MIR_LRS-FIXEDSLIT product, because technically it's not a TSO mode. So it might be better to explicitly check for the data model type, i.e. skip resample if a `CubeModel` is being processed.

jwst/resample/resample_spec_step.py
```diff
@@ -45,6 +46,9 @@ def process(self, input):
         if isinstance(input_models[0], MultiSlitModel):
             # result is a MultiProductModel
             result = self._process_multislit(input_models)
+        elif is_tso(input_models[0]):
```
hbushouse: Same comment here: you're equating an observation mode with a data model type, which isn't necessarily consistent. The multi-integration version of a MIR_LRS-FIXEDSLIT exposure will not be TSO, but will be a `CubeModel`. It's the `CubeModel` that can't be handled by `resample_spec`.

jwst/pipeline/calwebb_spec2.py
stscieisenhamer: FYI:, the `is_tso` check has a check for `CubeModel` there. Some previous issue discussed this, with anytime a `CubeModel` appears as input, it is TSO.   May not be sound, but is the current standard.

hbushouse: OK, so if that's the case, then mechanically this should work as desired. But it won't be obvious to someone just taking the code at face value.

stscieisenhamer: Agreed on not clear. Not clear to me if this is a documentation issue, or something else. As said, `is_tso` defines that anything `CubeModel` is time series. Which may be correct. But, as @hbushouse said, it may be more appropriate to make the case that both are checked for, since here being a `CubeModel` has other implications. Just because the check may be happening twice is a non-issue. But, if at some point, if `CubeBuild` is not always pinned to TSO, the explicit check will still be necessary.

philhodge: `CubeModel` isn't actually pinned to TSO.  Wouldn't it be better for `resample` to check the dimension of the data array?  That's the issue, isn't it, that `resample` won't process data unless it's 2D?

jdavies-st: So the consensus is that `is_tso()` is not appropriate for checking if something is 3D CubeModel or similar, despite it doing the check?

The `is_tso()` check further up in the code is actually the only thing making sure that a _rateints file run through this pipeline will actually produce a _calints output.  I used it for consistency's sake.  If it is not appropriate, then it should probably be changed further up in the code as well?  Thoughts?

jdavies-st: Also, `CubeModel` is always multi-integration, yes?  And whether a multi-integration cube is used for TSO purposes or some other (something like WFC3 DASH mode), shouldn't really matter to this pipeline or the steps in it.  The processing is the same.

sosey: It seems like in general, that `resample` and `CubeBuild` should define what they can accept, and return an error that's passed up to calwebb_spec2 which is then handled appropriately? Unless this is an expensive operation? That way the extra checking and decision making pipeline code gets called for the outliers, and the lower code is additionaly beefed up in the case it's called directly.

jdavies-st: Agree that returning useful errors from steps would be be the best.

As you mentioned in discussion before, it would be nice to have a `JwstError` or `StpipeError` class that could be subclassed with useful boilerplate error messages about completion or failure of steps, invalid input to steps, etc.  Since every step handles success or failure in its own way currently, I feel that is more of a infrastructure upgrade issue, not something unique to `resample`.  So if we update how the steps report success or failure, the pipelines calling them need to be modified as well to handle those properly.  So we need consistent reporting.  Currently, only `Spec2Pipeline` does any sort of error handling in terms of success or failure, so that it can exit gracefully.

Agree though, that this should be on our infrastructure to-do list.  👍

sosey: If it doesn't make sense to change the flow now, it may make the current code more readable to simply condense the `tso_mode` checks. i.e. set the suffix and make the correct calls in one `if is_tso(input):` block. That way everything explicitly done for tso mode is structured together

hbushouse: Regarding `CubeModel` vs. multi-integration, as far as I know, right now the only type of product `CubeModel` is used for is when you have 2-D data from multiple integrations, resulting in a 3-D stack. Other forms of pure 3-D data, such as IFU cubes, use their own unique data models, like `IFUCubeModel`. So right now there is a one-to-one correspondence between multi-integration and `CubeModel`, at least when you're past the ramp_fit stage (otherwise it's still 4-D) and you're before the spectral extraction stage (then you get into models like `MultiSpecModel`).

stscieisenhamer: And not sure how this factors in, but there will be (#2457 as soon as review is done), TSO-specific Level2 processing.

jdavies-st: Excellent!  Then we can punt on this issue because eventually TSO level 2 processing will have a separate pipeline?  We should resolve `is_tso()` at that point, and resolve it consistently.

hbushouse: No, the TSO-specific level-2b processing won't involve a separate pipeline module, but it will use a separate config file ("calwebb_tso_spec2.cfg"), which by default has skip = True for the `resample_spec` step. So it might accomplish the same thing, at least when the pipeline is executed using the cfg file (won't help if someone just calls it directly in python without the cfg).

But again, this will only address exposures explicitly tagged as being a TSO mode and hence invoke (in the default DMS processing environment) the `calwebb_tso_spec2` config. The original problematic data set is not TSO and hence nothing would change up front for that case (would still run the regular `calwebb_spec2` pipeline and hence the `is_tso()` hook is still needed to skip `resample_spec`.

jdavies-st: I see.  So any further change required to this PR then?

stscicrawford: Just followed up in conversation with @jdavies-st.  As @philhodge and @hbushouse suggested, this should be a dimensionality check rather than relying on checking for `tso` which is not very explicit on what it is checking for.   Once that is updated, this can be merged.   A unit test should be added to check for that, and I'll open a separate issue that the `CubeModel` should be more descriptive.

jdavies-st:
I've pulled out usage of `is_tso()` and instead just checked for 2D image dimensionality, as @philhodge and @stscicrawford suggest.  I also added a unit test to resample to verify this check works.

PR #2458, opened 2018-09-05
Set association code version to the jwst package version

### LINKED ISSUE #2447: replace local _version with package-wide one
# Resources
- #2352

stscieisenhamer:
Resolves #2447 
WIP #2352


stscieisenhamer:
@bernie-simon @jdavies-st `datamodels.make_headers` is failing now because of the `from .. import __version__`. However, `stpipe.step` does the exact same thing and does not error. Before diving in, any thoughts?

stscieisenhamer:
Nevermind. Understand the potential bug. Has to do with interaction of `make_headers.get_package` and getting the package of an object defined in an `__init__`.

PR #2460, opened 2018-09-05
enable extract_2d for NRC_TSGRISM in pipeline [#2549]

### LINKED ISSUE #2459: Allow extract_2d in calspec2 for NRC_TSGRISM?
https://github.com/spacetelescope/jwst/blob/af3347fec1b0dc3164d4160e0714f7aa6e358439/jwst/pipeline/calwebb_spec2.py#L230

I just noticed this looking at code in another PR. Since https://github.com/spacetelescope/jwst/pull/2286 allows extract2d for this mode, should we remove this if block? This is probably my oversight to begin with.... @hbushouse

sosey:
closes #2459

I'm going to open a separate issue for the `grism_objects` discussed there as I see another bug with call flow for `extract_orders`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -163,6 +163,8 @@ pipeline
 
 - Implement specific exit status for "no science on detector" [#2336]
 
+- Enabled `extract_2d` for NRC_TSGRISM [#2460]
+
 ramp_fitting
 ------------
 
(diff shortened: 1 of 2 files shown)


sosey:
do you want us referencing issue or PR numbers in the changelog

stscicrawford:
Yes -- makes it easier to track things.   Thanks for doing so!

sosey:
@stscicrawford  uh...yes PR...or...yes issue....or both?  I would tend to reference the PR in the changelog and the issue in the PR...but I don't remember hearing if there was a preference here

hbushouse:
Looks like the reference of choice for the log so far is PR number (as opposed to Issue number)

stscicrawford:
PR seems sensible as that adds the change.  The issue should be referenced in the PR.

PR #2464, opened 2018-09-05
Improved memory management: implemented a more judicious choice of me…

dmggh:
…mory allocation, deleted no longer used arrays, corrected function that removes all segments having only a single group if there are other segments in the ramp. Segregated code needed for optional output to only execute if requested. Correct handling of groups in which all pixels have insufficient data for a first difference.  Correct handling of data whose initial group is saturated. Ensured that variances of non-existing segments are not populating with erroneous values in the optional output. Correct handling of ramps whose single good segment is a single group.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -24,17 +24,15 @@ from ..lib import pipe_utils
 from . import gls_fit           # used only if algorithm is "GLS"
 from . import utils
 
+
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 BUFSIZE = 1024 * 30000  # 30Mb cache size for data section
 
-# Replace zero or negative variances with this.
-LARGE_VARIANCE = 1.e8
-
 
 def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
-              algorithm, weighting):
+             algorithm, weighting):
     """
     Extended Summary
     ----------------
@@ -43,7 +41,7 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
     cosmic rays) of the pixel's ramp divided by the effective integration time.
     The weighting parameter must currently be set to 'optim', to use the optimal
     weighting (paper by Fixsen, ref. TBA) will be used in the fitting; this is
-    currently the only supported weighting scheme.   
+    currently the only supported weighting scheme.
 
     Parameters
     ----------
@@ -136,7 +134,7 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
         gain for all pixels
 
     weighting: string
-        'optimal' specifies that optimal weighting should be used; currently 
+        'optimal' specifies that optimal weighting should be used; currently
         the only weighting supported.
 
     Returns
@@ -155,7 +153,7 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
     """
     tstart = time.time()
 
-    # get needed sizes and shapes
+    # Get needed sizes and shapes
     nreads, npix, imshape, cubeshape, n_int, instrume, frame_time, ngroups, \
         group_time = utils.get_dataset_info(model)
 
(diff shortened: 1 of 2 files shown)


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -383,87 +417,77 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+        #   variance is proportional to the median estimated slope) to
+        #   outrageously large values so that they will have negligible
+        #   contributions.
+        var_p4[num_int,:,:,:] *= ( segs_4[num_int,:,:,:] > 0)
+        var_p4[var_p4 <= 0.] = utils.LARGE_VARIANCE
+        var_r4[num_int,:,:,:] *= ( segs_4[num_int,:,:,:] > 0)
+        var_r4[var_r4 <= 0.] = utils.LARGE_VARIANCE
+
+        s_inv_var_p3[num_int, :, :] = (1./var_p4[num_int, :, :, :]).sum(axis=0)
+        var_p3[num_int, :, :] = 1./ (1./var_p4[num_int, :, :, :]).sum(axis=0)
```
philhodge: Is there a reason for recomputing the sum?  It looks as if you could use `s_inv_var_p3`.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -383,87 +417,77 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+        s_inv_var_r3[num_int, :, :] = (1./var_r4[num_int, :, :, :]).sum(axis=0)
+        var_r3[num_int, :, :] = 1./ (1./var_r4[num_int, :, :, :]).sum(axis=0)
```
philhodge: Similar comment as for line 435.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -383,87 +417,77 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
                             var_p4_int.shape[1]*var_p4_int.shape[2]))
-        var_r4_int2 = var_r4_int.reshape(( var_r4_int.shape[0],
-                            var_r4_int.shape[1]*var_r4_int.shape[2]))
-        inv_var_both4_int2 = inv_var_both4_int.reshape(
-                           ( inv_var_both4_int.shape[0],
-                             inv_var_both4_int.shape[1] *
-                             inv_var_both4_int.shape[2]))
 
         num_seg_int = num_seg_per_int[num_int,:, :] # number of segs per pixel
         num_pix = num_seg_int.shape[0] * num_seg_int.shape[1]
```
philhodge: `num_pix` is never used, so you could delete this line.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -474,69 +498,84 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+
+    del s_inv_var_both2, s_slope_by_var2, s_slope_by_var3, slope_by_var4
+    del s_inv_var_both3
+
+    #  Replace nans in slope_dataset2 with 0 (for non-existing segments)
+    slope_dataset2[np.isnan(slope_dataset2)] = 0.
 
     # Compute the integration-specific slope
-    the_num = (slope_seg * inv_var_both4).sum(axis = 1) # sum over all segments
+    the_num = (opt_res.slope_seg.copy() * inv_var_both4).sum(axis = 1)
```
philhodge: I don't think you need to copy `opt_res.slope_seg`, since it's used in an expression rather than being assigned directly.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -1559,10 +1628,9 @@ def fit_next_segment(start, end_st, end_heads, pixel_done, data_sect, mask_2d,
         end_heads[end_heads < 0.] = 0.
 
     return f_max_seg, num_seg
-    # end of fit_next_segment()
 
 
-def fit_lines(data, mask_2d, rn_sect, gain_sect, ngroups, weighting, group_time):
+def fit_lines(data, mask_2d, rn_sect, gain_sect, ngroups, weighting):
```
philhodge: The doc string still defines `group_time`, but it has been removed from the calling sequence.

jwst/ramp_fitting/utils.py
```diff
@@ -1089,17 +1130,128 @@ def get_ref_subs(model, readnoise_model, gain_model, nframes):
...
+
+    slope_int: float, 3D array
+        Cube of integration-specific slopes; for ramps that are saturated in
+        the initial group, this variance has been reset to a huge value to
+        minimize the ramps contribution.
+    """
+    where_all_groups_sat = np.where( sat_0th_group_int != 0)
+    num_of_sats = len(where_all_groups_sat[0])
+
+    if num_of_sats > 0:
```
philhodge: It doesn't look as if this test is needed.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -474,69 +498,84 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+    if var_r4 is not None:
+        del var_r4
+
+    if inv_var is not None:
+        del inv_var
+
+    if pixeldq is not None:  
+        del pixeldq
 
     # For multiple-integration datasets, will output integration-specific
```
stsci-hack: Initialization of 'int_times' is not assured at this point.  In fact, interactive testing using the niriss/test_ramp_fit nightly test triggered an UnboundLocalError Exception at this point due to an uninitialized variable 'int_times'.  Adding the line 'int_times = None' prior to the 'if n_int' block will resolve this problem.

Issue #2445, opened 2018-08-31
move "non-regression" regressions into tests_nightly


stscieisenhamer:
A number of tests were developed during the period of transition for the nightly regression tests. As such, these tests were included in the unit tests of the package so that they were documented, accessible, and runnable locally.

Now that the transition phase is complete, time to move them into the `tests_nightly` folder and properly setup the fixtures/data locations. The desired effects of 1) tests will get run regularly and 2) the test data will be moved out of the repository, will be then achieved.

jdavies-st:
One thing to add here is that the test these tests and those in `tests_nightly` use the TESTS_BIGDATA environment variable in a slightly different way:

- `--runslow` with the TESTS_BIGDATA env variable correctly set enable the ones in the unit test area to run.
- `--bigdata` with the TESTS_BIGDATA env variable correctly set enable the regression tests to run properly

We may want there to be a difference between `--runslow` and `--bigdata`, as one could conceivably have slow tests that don't require TESTS_BIGDATA, and vice versa.  In fact, this is already the case in the unit tests I believe, where some `--runslow` ones don't need TEST_BIGDATA.

jdavies-st:
A further characteristic of the regression tests currently in the unit tests of the package is that if one has TEST_BIGDATA defined, one cannot turn off running those tests that might use TEST_BIGDATA. I would prefer that the `--bigdata` flag be used to run these and not have them automatically run if I have TEST_BIGDATA env variable defined.

The current workaround of having to `unset TEST_BIGDATA` if you don't want those tests run, and then `export TEST_BIGDATA=/my/local/test/data` when you do is a bit cumbersome.  Better to have the running of them triggered from the command line switch.

PR #2465, opened 2018-09-05
Add back in fixture mk_tmp_dirs

stscieisenhamer:
At some point in PEP8 fixing, a fixture was removed from a number of tests. They are added back in, so that once #2445 is dealt with, there is consistency during the process.


jdavies-st:
I"ll add that this `mk_tmp_dirs` fixture might either go away or perhaps get merged with the `_jail` fixture once #2445 is resolved?

Issue #1720, opened 2018-02-07
Background step can't run via command-line arguments


hbushouse:
As documented at https://stsci-ins.basecamphq.com/projects/11477312-jwst-pipeline/posts/104239169/comments#379399911 the background step in calwebb_spec2 can't be run from the command-line using simple file name arguments for the input science and background images. Right now it only works as part of the calwebb_spec2 pipeline when using an ASN file as input, which specifies one or more background exposures to pass as a list to the background step.

Would be convenient if the step could be made to run from the command line using file names as the arguments.

This is also being tracked in https://jira.stsci.edu/browse/JP-196

hbushouse:
Low priority for B7.2. Falls into the category of "user convenience" features.

jdavies-st:
This should be able to be accomplished by adding a custom parameter type to the argparse class.  I'd looked into this about a year ago and already have it on my to-do list.  I'll take a crack at this.

stscijgbot-jp:
_This issue is tracked on JIRA as [JP-196](https://jira.stsci.edu/browse/JP-196)._

stscijgbot-jp:
_Comment by [Rosa Diaz](https://jira.stsci.edu/secure/ViewProfile.jspa?name=rdiaz) on [JIRA](https://jira.stsci.edu/browse/JP-196?focusedCommentId=34244#comment-34244):_

### [ **Phil Hodge** Mon, 11 Dec at 5:15pm ](https://stsci-ins.basecamphq.com/projects/11477312-jwst-pipeline/posts/104239169/comments#comment_377005828)
Jonathan just explained to me that if you're running the calwebb_spec2 pipeline, you need to give it an association file.  A member with 'exptype': 'science' will be your science exposure (just one, in this case), and the member with type 'exptype': 'background' (again, just one in this case) will be your "background" exposure.
If you run the background step by itself on the command line (strun background.cfg ...), that won't work, because the arguments are strings unless the step defines arguments to be something else (such as a list), and the step doesn't define any arguments (i.e. the `spec` variable is empty).
 
### [ **James Davies** Mon, 11 Dec at 5:44pm ](https://stsci-ins.basecamphq.com/projects/11477312-jwst-pipeline/posts/104239169/comments#comment_377007608)
Yeah, background step not running from the command line is a known issue and I believe is a limitation of argparse at the moment, a library we depend on for the command line interface. Would be nice to fix this.

That said, a quick API check for the step in ipython:
[...]

PR #2467, opened 2018-09-05
fix for command line extract_2d optional grism inputs [#2466]

### LINKED ISSUE #2466: extract_2d_step optional inputs need type update for grism modes
The types in the spec for grism mode variable inputs are wrong, update to be consistent and make sure that using them produces the correct error.

`extract_height =  int(default=None)` --> `extract_height =  integer(default=None)`

This is a harder one, when `extract_orders` is used from the command-line, there's an error in grisms because it's expecting nested lists
```
available_orders = [x[1] for x in extract_orders if x[0] == input_model.meta.instrument.filter].pop()
IndexError: pop from empty list
```
I think because when I try and set the cfg override to 
```
name = "extract_2d" 
class = "jwst.extract_2d.Extract2dStep"
extract_orders=[['F444W', [2]]]
```
stpipes configobj parser morphs it to 
`["[['F444W'", '[2]]]']`

:/  Not that I'm shocked, I was just hoping it had more devious smarts. The updated schema for the `wavelengthrange` model in grism modes has a tuple of (filter, [orders]) that read in as nested lists: 
`[['F277W', [1]], ['F322W2', [1]], ['F356W', [1]], ['F444W', [1]]]`

This doesn't affect the default pipeline for `nrc_tsgrism` which just uses the reference file.

But I'm not sure there's a way to specify a complex input like this in the cfg or the command line without editing the regex inside configobj.validate.  It's possible that this optional input could be regardless of filter, such that the user would specify the list of orders they want, and those would be used regardless of filter for the input model.  I'll reference a PR for that shortly that includes the example syntax.

sosey:
closes #2466 
* updates the int to integer type spec for extract_height
* fixes type specification for user-override extract_orders

For the `--extract_orders`, this allows the user to specify a list of orders to extract that will be used instead of the default in the reference file. The reference file specifies the orders to be extracted by filter. The override will extract the specified orders for any filter. 

This is currently an option in `NRC_TSGRISM`, the team only wants one order extracted, the reference file specifies the first order, if the user would instead like to extract the second order, the `extract_2d.cfg` file can be updated as such:
```
name = "extract_2d" 
class = "jwst.extract_2d.Extract2dStep"
extract_orders = 2,
```
and order 2 will be extracted for the input model that's specified.  The extraction code will return an error if more than one order is specified (such as `extract_orders = 1,2,`) because it only outputs a `SlitModel` not a `MultiSlitModel`.
```
strun extract_2d_orders.cfg ../tso_wcs_assigned_test.fits

2018-09-05 18:19:35,154 - stpipe.extract_2d - INFO - Extract2dStep instance created.
2018-09-05 18:19:35,179 - stpipe.extract_2d - INFO - Step extract_2d running with args ('../tso_wcs_assigned_test.fits',).
2018-09-05 18:19:39,096 - stpipe.extract_2d - INFO - EXP_TYPE is NRC_TSGRISM
[2]
2018-09-05 18:19:39,184 - stpipe.extract_2d - INFO - xmin, xmax: 1948.0147270747234 2048  ymin, ymax: 0 63.0
2018-09-05 18:19:39,575 - stpipe.extract_2d - INFO - WCS made explicit for order: 2
2018-09-05 18:19:39,575 - stpipe.extract_2d - INFO - Trace extents are: (xmin:1948, ymin:0), (xmax:2048, ymax:63)
2018-09-05 18:19:39,692 - stpipe.extract_2d - INFO - Finished extractions
2018-09-05 18:19:40,975 - stpipe.extract_2d - INFO - Saved model in tso_wcs_assigned_test_extract_2d.fits
2018-09-05 18:19:40,975 - stpipe.extract_2d - INFO - Step extract_2d done
```
These are the default results:
```
(jwstdev) bash-3.2$ strun extract_2d.cfg ../tso_wcs_assigned_test.fits 
2018-09-05 18:30:54,643 - stpipe.extract_2d - INFO - Extract2dStep instance created.
2018-09-05 18:30:54,673 - stpipe.extract_2d - INFO - Step extract_2d running with args ('../tso_wcs_assigned_test.fits',).
2018-09-05 18:30:58,698 - stpipe.extract_2d - INFO - EXP_TYPE is NRC_TSGRISM
2018-09-05 18:30:58,719 - stpipe.extract_2d - INFO - Using default order extraction from reference file
2018-09-05 18:30:58,791 - stpipe.extract_2d - INFO - xmin, xmax: 647.9255139717636 1845.7302759243282  ymin, ymax: 0 63.0
2018-09-05 18:30:59,211 - stpipe.extract_2d - INFO - WCS made explicit for order: 1
2018-09-05 18:30:59,211 - stpipe.extract_2d - INFO - Trace extents are: (xmin:647, ymin:0), (xmax:1845, ymax:63)
2018-09-05 18:30:59,326 - stpipe.extract_2d - INFO - Finished extractions
2018-09-05 18:31:00,760 - stpipe.extract_2d - INFO - Saved model in tso_wcs_assigned_test_extract_2d.fits
2018-09-05 18:31:00,760 - stpipe.extract_2d - INFO - Step extract_2d done
```
However, I'm not sure how this would work purely on the command line because these return type errors:
```
571  strun extract_2d.cfg ../tso_wcs_assigned_test.fits --extract_orders=2,
572  strun extract_2d.cfg ../tso_wcs_assigned_test.fits --extract_orders=2
573  strun extract_2d.cfg ../tso_wcs_assigned_test.fits --extract_orders="2,"
574  strun extract_2d.cfg ../tso_wcs_assigned_test.fits --extract_orders=[2,]
575  strun extract_2d.cfg ../tso_wcs_assigned_test.fits --extract_orders=[2]
576  strun extract_2d.cfg ../tso_wcs_assigned_test.fits --extract_orders=2
```
@nden @stscieisenhamer  is there a way you know to specify list optionals, what have I missed?

~~This also only affects this specific entry, anyone can pass `extract_2d.extract2()` the filter specified list and everything works.~~ on second thought, yah, this makes all input for the optional `extract_orders` be a list of ints, which is different than the reference file container, but I think that's okay here since the function only takes one input_model, and the reference file is meant to cover all available options.

### DIFF
--- a/jwst/extract_2d/grisms.py
+++ b/jwst/extract_2d/grisms.py
@@ -89,10 +89,15 @@ def extract_tso_object(input_model,
     if extract_orders is None:
         log.info("Using default order extraction from reference file")
         extract_orders = ref_extract_orders
+        available_orders = [x[1] for x in extract_orders if x[0] == input_model.meta.instrument.filter].pop()
     else:
-        raise NotImplementedError("Multiple order extraction for TSO not currently implemented")
+        if not isinstance(extract_orders, list):
+            raise TypeError('Expected extract_orders to be a list of integers')
+        print(extract_orders)
+        available_orders = extract_orders
 
-    available_orders = [x[1] for x in extract_orders if x[0] == input_model.meta.instrument.filter].pop()
+    if len(available_orders) > 1:
+        raise NotImplementedError("Multiple order extraction for TSO not currently implemented")
 
     # team currently wants only a slit model
     # if len(available_orders) > 1:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -89,6 +89,7 @@ extract_2d
 ----------
 - NRC_TSGRISM implemented with set source location and extraction options [#1710, #1235]
 - Fixed step calling error for unreferenced attribute [#2463]
+- Fixed type specification for optional grism mode inputs [#2467]
 
 firstframe
 ----------
--- a/jwst/extract_2d/extract_2d_step.py
+++ b/jwst/extract_2d/extract_2d_step.py
@@ -16,9 +16,9 @@ class Extract2dStep(Step):
     spec = """
         slit_name = string(default=None)
         apply_wavecorr = boolean(default=True)
-        extract_orders = list(default=None)
-        extract_height =  int(default=None)
-        grism_objects = list(default=None)
+        extract_orders = int_list(default=None)  # list of orders to extract
+        extract_height =  integer(default=None)  # extraction height in pixels
+        grism_objects = list(default=None)  # list of grism objects to use
     """
 
     reference_file_types = ['wavecorr', 'wavelengthrange']


hbushouse:
The problem of not being able to give a list as an argument value has been plaguing us for years (same problem with the background step and the param that allows a list of background image/exposure names - it doesn't work from command line). I'm hoping someone can find a cure. Might require digging into stpipe and/or the way it uses configobj for param parsing.

jwst/extract_2d/grisms.py
```diff
@@ -89,10 +89,15 @@ def extract_tso_object(input_model,
...
     else:
-        raise NotImplementedError("Multiple order extraction for TSO not currently implemented")
+        if not isinstance(extract_orders, list):
+            raise TypeError('Expected extract_orders to be a list of integers')
+        print(extract_orders)
+        available_orders = extract_orders
 
-    available_orders = [x[1] for x in extract_orders if x[0] == input_model.meta.instrument.filter].pop()
+    if len(available_orders) > 1:
+        raise NotImplementedError("Multiple order extraction for TSO not currently implemented")
```
hbushouse: Just curious as to why we can't do multiple orders for TSO mode? Is it a problem with not knowing where the other orders are located or what?

stscieisenhamer:
with regards to the list issue: is there an issue? if not can, someone file it?

jwst/extract_2d/grisms.py
sosey: The team explicitly does not want anything but the first order subtracted, so the output from the step is a `SlitModel` only, not a `MultiSlitModel`... This is reinforced in the reference file. However, I left the information there for the second order, and wanted to make it possible for someone to choose to try and extract the second order if they wanted, it just needs to create two different output files.  I can add in creation of a `MultiSlitModel` if in the future the team wants to extract multiple orders at the same time, but that's not something they've asked for and I'm working hard at delivering requirements w/o too many unasked for bells and whistles ;)

jdavies-st:
The argparse list issue is #1720.

PR #2468, opened 2018-09-06
A new script to add or remove filenames from an association file

bernie-simon:
This update adds a new command line script, supporting code in the associations directory, and tests to add or remove filenames from an association file. Only json format association files are supported.

### DIFF
--- a/jwst/associations/asn_edit.py
+++ b/jwst/associations/asn_edit.py
@@ -196,8 +196,9 @@ def _lookup(asn, filename, ignore_suffix=False):
     found = []
     path = _path(filename)
     basename = op.basename(path)
+    (root, ext) = op.splitext(basename)
     if ignore_suffix:
-        search_key, separator = suffix.remove_suffix(basename)
+        search_key, separator = suffix.remove_suffix(root)
     else:
         search_key = basename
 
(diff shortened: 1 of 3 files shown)


scripts/asn_edit
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python
+
+import argparse
+from ..associations import asn_edit
```
stscieisenhamer: scripts require absolute import paths, since they are run outside of the package context.

jwst/associations/asn_edit.py
```diff
@@ -0,0 +1,237 @@
...
+        search_key, separator = suffix.remove_suffix(basename)
+    else:
+        search_key = basename
+
+    for i, product in enumerate(asn['products']):
+        for j, member in  enumerate(product['members']):
+            expname = member.get('expname')
+            if expname:
+                if ignore_suffix:
+                    expname, separator = suffix.remove_suffix(expname)
```
stscieisenhamer: Need to do `remove_suffix(splitext(basename(expname))[0])[0]`

jwst/associations/asn_edit.py
```diff
@@ -0,0 +1,237 @@
...
+# Internal functions
+
+def _lookup(asn, filename, ignore_suffix=False):
+    """
+    Look up the locations a file is found in an association
+    """
+
+    found = []
+    path = _path(filename)
+    basename = op.basename(path)
```
stscieisenhamer: Need to get just the root: `splitext(basename(path))[0].  `remove_suffix` does not expect an extension such a `.fits`.

PR #2472, opened 2018-09-06
Improve speed in cube_build

jemorrison:
This PR removed the class spaxel and replaced it with numpy arrays to improve the speed of cube_build. An unnecessary arrays of providing the ra and dec offset was removed (this also
removed user options to provide these values).
Some redundant code was removed and a single function map_detector_to_outputframe was added.


jemorrison:
The routine cube_cloud_quick.py was an attempt to speed up the code by reducing which spaxels a point needed to search to determine if it fell in the roi. It is not faster than cube_cloud.py. I left cube_cloud_quick.py in the pr it is not called but maybe in the future I can think of a faster way to do what I am trying.

stscicrawford:
Can you mention in the CHANGELOG a summary of the performance improvement that James measured?   A rough estimate of it would be the type of information that I think INS would be interested in.

PR #2481, opened 2018-09-07
Update Level2 product naming to use remove_suffix

### LINKED ISSUE #2324: too much suffix being removed
# Issue
In re-working regression `test_nirisssoss2_pipeline.py`, noted that the reworked version was taking input name
```
jw00034001001_01101_00001_NIRISS_rate_ref.fits
```
and saving results in names such as
```
jw00034001001_01101_00001_NIRISS_calints.fits
```
when the name should have been
```
jw00034001001_01101_00001_NIRISS_rate_ref_calints.fits
```
since `ref` is not a standard suffix. Hand-testing `remove_suffix` shows correct behavior.

stscieisenhamer:
Resolves #2324 

# Note
Reviewers: The one code file is `associations.lib.rules_level2_base.py`


stscieisenhamer:
@bernie-simon 👍 will add to #2482.

Issue #2487, opened 2018-09-10
flat_field doesn't always find the dispersion direction correctly


philhodge:
In the `flat_field` step for NIRSpec data, the dispersion direction is determined by comparing wavelengths in the wavelength array in the input file, to see along which image axis is the wavelength increasing more quickly.  This is needed in order to average the "fast variation" flat (flat field as a function of wavelength, from a table in the reference file) over each pixel in the flat field image.  Using the wrong dispersion direction would probably not usually result in a large error in the computed (on-the-fly) flat.

The main reason why the dispersion direction was not always computed correctly is that there are (often) NaNs in the wavelength array in the input file, and those NaNs get replaced early in `flat_field` by "harmless" values that turned out not to be so harmless.  Depending on the type of data, values of -1000 and 1 were used as replacements for NaN.  Also, the wavelengths were compared only for three pixels near the center of the 2-D cutout, and in some cases those pixels were on the edge or entirely outside the bounding box.  The replacement value should be the same for all types of data, and the function for determining the dispersion direction must ignore pixels with this replacement value.

stscicrawford:
Closed with #2492

PR #2492, opened 2018-09-11
Improve the code for finding the dispersion direction

philhodge:
Function `find_dispaxis` was added.  This determines the dispersion direction by comparing the rate at which wavelength changes from pixel to pixel, in the horizontal and vertical directions.  All pixels with non-NaN and non-zero values of wavelength are now used, rather than just three points near the middle of the image.

The function that combines the flat-field images (slowly varying with wavelength) with the tabular flat field (varying more quickly with wavelength) replaces the NaN or zero wavelength values with an average wavelength, to reduce interpolation problems near the edges of the bounding box.

See issue #2487.

### DIFF
--- a/jwst/flatfield/tests/test_find_dispaxis.py
+++ b/jwst/flatfield/tests/test_find_dispaxis.py
@@ -7,28 +7,24 @@ from jwst.flatfield import flat_field
 
 def test_find_dispaxis():
 
-    # Create a dummy array of wavelengths.  The array is mostly zero,
-    # but the non-zero values are clearly increasing more rapidly
-    # in the horizontal direction than in the vertical direction.
-    wl = np.zeros((30, 200), dtype=np.float32)
-    wl[16, :] = np.arange(200, dtype=np.float32)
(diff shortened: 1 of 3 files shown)


jwst/flatfield/tests/test_find_dispaxis.py
```diff
@@ -0,0 +1,34 @@
...
+    wl[16, :] = np.arange(200, dtype=np.float32)
+    wl[17, :] = np.arange(200, dtype=np.float32) + 0.01
+    wl[18, :] = np.arange(200, dtype=np.float32) + 0.02
+    wl[19, :] = np.arange(200, dtype=np.float32) + 0.03
+
+    dispaxis = flat_field.find_dispaxis(wl)
+    assert dispaxis == 1
+
+    # This is the code that was in combine_fast_slow and that find_dispaxis
+    # replaced.
```
stscicrawford: I think this can be removed as, unless  I'm missing something, it isn't testing the code itself and is replaced.

stscicrawford:
It would be good to add a vertical test as well as the horizontal test.   Apologies for not specifying that previously.

PR #2494, opened 2018-09-11
Sanitize GWA_TTIL keyword for NIRSpec

### LINKED ISSUE #2493: GWA_TILT keyword not automatically sanitized
Issue [JP-368](https://jira.stsci.edu/browse/JP-368) was created by Graham Kanarek:

The create_data script sanitizes the GWA_XTIL and GWA_YTIL keywords – see create_dms_data.py lines 370, 371, 427, & 428. However the same is not done for the GWA_TILT keyword, so when the raw file has a string instead of a float, the datamodel throws a warning. The GWA_TILT keyword should also be automatically sanitized.

gkanarek:
The GWA_TTIL keyword should be sanitized along with the GWA_XTIL and GWA_YTIL keywords, so that the datamodel doesn't throw a validation error.

Resolves #2493.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -73,6 +73,8 @@ firstframe
 fits_generator
 --------------
 
+- NIRSpec data now automatically sanitizes the GWA_TILT keyword. [#2494]
+
 
 flatfield
 ---------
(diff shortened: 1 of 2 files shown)


jdavies-st:
Similar to un-merged PR #1386.

stscicrawford:
@gkanarek Can you also add an entry in the Change log?   After that, we can merge this.

gkanarek:
@stscicrawford done!

stscicrawford:
Thanks @gkanarek !

PR #2502, opened 2018-09-12
Initialize data arrays and tables from function args in model_base

### LINKED ISSUE #2351: datamodel subclasses should not require so much boilerplate to define the API
Currently, most datamodel subclasses have lots of boilerplate in the `__init__` method, expanding the API to list all the possible attributes than can be specified on creation.  These are not required in order for the datamodel to work, but they do make the API flexible.

So if I'm making an `ImageModel` and want to specifiy a wavelength extension, I can do this:
```
from jwst import datamodels
import numpy as np

data = np.random.random((100, 100))
wavelength = np.arange(data.shape[0]) * 8.5e-5 + 3.2

im = datamodels.ImageModel(data=data, wavelength=wavelength)
```
This requires that any `kwargs` that can possibly be passed to that datamodel need to be listed in the datamodel's `__init__` function.  This is done like:

https://github.com/spacetelescope/jwst/blob/cc51287cc48575cb3f008e44dc7e9fb2900a90ba/jwst/datamodels/image.py#L48-L75

But this is a lot of boilerplate, especially when the exact same information is in the schema for `ImageModel`:

https://github.com/spacetelescope/jwst/blob/cc51287cc48575cb3f008e44dc7e9fb2900a90ba/jwst/datamodels/schemas/image.schema.yaml#L7-L52

As you can see, for `wavelength`, it's actually not in the schema.  Which points to the problems of having to update boilerplate in a couple different places.

I suspect this can be simplified.  Either:

1. Have the list of properties in one location, i.e. the schema.  And have simple code in the datamodels base class generate these attributes when reading the schema, **or**

2. Have them listed in both places, but have there be automated checks to make sure that if it's in the `foo.schema.yaml` schema, then it's in the `foo.py` code and _vice versa_.

Or perhaps there are other ideas?  Would be interested in your thoughts.

Either way, we need to make the code more maintainable.  As it is, it is very easy for us developers to make mistakes in implementation of a new datamodel or in modifying an old one.

bernie-simon:
The classes in datamodels contain a lot of code to initialize arrays
from function arguments. This update centralizes this code in
DataModel, the base class of the pther classes. It reads the arguments
from **kwargs, sees if the arguments are mentioned in the schema, and
if so, uses the value of the argument to initialize an array or table.
The intialization of the arrays in each of the subclasses has been
removed. This has no effect on the code that calls the DataModel
subclasses, but any programmer adding a new subclass should examine
the code first. Eliminating the initialization code has allowed many
of the subclasses to eliminate their \_\_init\_\_ function, they
instead use the function in their base class.
    
There is another kind of initialization code in the DataModel
subclasses that could not be eliminated. That is the code that
initializes arrays to their default values when they are otherwise not
present. These are the lines that look like:
    
    self.dq = self.dq
    
These lines could not be eliminated because there is no indication in
the schema files which arrays should be initialized and which should
not. However, the most common case is initializing the dq and err
arrays. The code to do this is in the ImageModel and
ReferenceImageModel classes and when subclasses initialized just these
two arrays, they have been modified to have one of these two classes
as their base class.

Resolves #2351.


jdavies-st:
Cool Bernie!  Looks great.  I'll run the regression tests on this branch, as I think they have better coverage for some of these models than our unit tests.

Also, the CHANGES.rst should indicate [#2502] not [#2351].

jwst/datamodels/cube.py
```diff
@@ -44,43 +44,3 @@ class CubeModel(model_base.DataModel):
...
-            self.wavelength = wavelength
-
-        if var_poisson is not None:
-            self.var_poisson = var_poisson
-
-        if var_rnoise is not None:
-            self.var_rnoise = var_rnoise
-   
-        # Implicitly create arrays
-        self.dq = self.dq
```
hbushouse: So how or where is this functionality now handled, so that the DQ and ERR arrays get "inflated" even when they aren't present in the incoming data?

jdavies-st: Looks like `CubeModel` is now subclasssed from `ImageModel`, which means it gets the array inflation goodies in `ImageModel`, but of course has its own schema, so dimensions of the attributes are not inherited from `ImageModel`

jdavies-st:
So I'm seeing several failures in the regression test suite related to this PR.  Here's the first:
```
(jwst_dev) rey> pytest -s --bigdata --basetemp=/Users/jdavies/delete_me tests_nightly/general/miri/test_extract1d2.py
==================================================================== test session starts ====================================================================
platform darwin -- Python 3.6.6, pytest-3.7.4, py-1.6.0, pluggy-0.7.1
plugins: requests-mock-1.5.0, xdist-1.23.0, remotedata-0.3.0, openfiles-0.3.0, forked-0.2, doctestplus-0.1.3, cov-2.6.0, arraydiff-0.2
collected 1 item                                                                                                                                            

miri/test_extract1d2.py 
2018-09-14 11:15:46,850 - stpipe.Extract1dStep - INFO - Extract1dStep instance created.
2018-09-14 11:15:46,882 - stpipe.Extract1dStep - INFO - Step Extract1dStep running with args ('/Users/jdavies/jwst_test_data//miri/test_extract1d/jw80600012001_02101_00003_mirimage_photom.fits',).
2018-09-14 11:15:47,367 - stpipe.Extract1dStep - INFO - Using EXTRACT1D reference file /grp/crds/cache/references/jwst/jwst_miri_extract1d_0004.json
2018-09-14 11:15:47,396 - stpipe.Extract1dStep - WARNING - spectral_order is None; using 1
2018-09-14 11:15:47,409 - stpipe.Extract1dStep - INFO - Total number of dither points = None; assuming no nod/dither offset
F

========================================================================= FAILURES ==========================================================================
___________________________________________________________________ test_extract1d_miri2 ____________________________________________________________________

_bigdata = '/Users/jdavies/jwst_test_data/'

    def test_extract1d_miri2(_bigdata):
        """
    
        Regression test of extract_1d step performed on MIRI LRS slitless data.
    
        """
        suffix = 'extract_1d'
        output_file_base, output_file = add_suffix('extract1d2_output.fits', suffix)
    
        try:
            os.remove(output_file)
        except:
            pass
    
    
    
        Extract1dStep.call(_bigdata+'/miri/test_extract1d/jw80600012001_02101_00003_mirimage_photom.fits',
                           smoothing_length=0,
>                          output_file=output_file_base, suffix=suffix
                           )

../../dev/jwst/jwst/tests_nightly/general/miri/test_extract1d2.py:32: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../dev/jwst/jwst/stpipe/step.py:531: in call
    return instance.run(*args)
../../dev/jwst/jwst/stpipe/step.py:400: in run
../../dev/jwst/jwst/extract_1d/extract_1d_step.py:117: in process
    self.log_increment)
../../dev/jwst/jwst/extract_1d/extract.py:2179: in do_extract1d
    prev_offset, True, extract_params)
../../dev/jwst/jwst/extract_1d/extract.py:2709: in extract_one_slit
    extract_model.extract(data, wl_array, verbose)
../../dev/jwst/jwst/extract_1d/extract.py:1578: in extract
    image = np.transpose(data, (1, 0))
../../anaconda3/envs/jwst_dev/lib/python3.6/site-packages/numpy/core/fromnumeric.py:598: in transpose
    return _wrapfunc(a, 'transpose', axes)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = array([[[ 0.        ,  0.        ,  0.        , ..., -0.370115  ,
          0.21816392,  0.27713132],
        [ 0.    ...
        [ 0.        ,  0.        ,  0.        , ...,  0.19403386,
          0.39696485, -0.07921567]]], dtype=float32)
method = 'transpose', args = ((1, 0),), kwds = {}

    def _wrapfunc(obj, method, *args, **kwds):
        try:
>           return getattr(obj, method)(*args, **kwds)
E           ValueError: axes don't match array

../../anaconda3/envs/jwst_dev/lib/python3.6/site-packages/numpy/core/fromnumeric.py:51: ValueError
```

jwst/datamodels/cube.py
jdavies-st: Though I see that there isn't a line in the `__init__` method that has
```
super(CubeModel, self).__init__(init=init, **kwargs)
```

jdavies-st: In fact there isn't an `__init__` method.  Should there be?  `ImageModel` which subclasses `DataModel` has one.  I would think we would need one here too?

jdavies-st:
This latest commit 9ac8444 improves things!  We've gone from 9 test failures down to 4 now.  The tests currently failing:
```
jwst/tests_nightly/general/miri/test_saturation2.py
jwst/tests_nightly/general/miri/test_sloperpipeline_2.py
jwst/tests_nightly/general/nirspec/test_detector1.py
jwst/tests_nightly/general/pipelines/test_miri_lrs_slit_1b.py
```
And it looks like the same reason as above in the saturation step.

jdavies-st:
When you resolve the CHANGES.rst conflict, just be sure to update this PR [#2502] as the basis for these changes, not the issue as currently listed.

PR #2510, opened 2018-09-14
Make tweakreg source detection threshold higher

### LINKED ISSUE #2489: Raise default source detection threshold for tweakreg
Currently it is SNR = 3.0, which is much too low and causes many spurious detections in simulated data from both NIRCam and MIRI.  It needs to be much higher and robust.  Say 10?  @larrybradley

jdavies-st:
The low source detection threshold SNR of 3.0 was causing spurious detections in `tweakreg`.  Many spurious sources would be detected and an incorrect alignment was computed.  Better to err on the side of caution and have a high detection threshold of 10.  If there are too few sources found at this higher detection threshold, then the step is skipped, which I believe is a desirable default for pipeline use.

Also made consistent between the .cfg and default spec the parameters
```
enforce_user_order = True
kernel_fwhm = 2.5
searchrad = 1.0
```
Could reviewer(s) confirm that these are reasonable defaults?  The point is to have them be consistent between the default config and the default spec.

Finally, `tweakreg` has been off in our default configuration for a long time.  Time turn turn it back on.

Resolves #2489.

### DIFF
--- a/jwst/pipeline/calwebb_image3.cfg
+++ b/jwst/pipeline/calwebb_image3.cfg
@@ -4,7 +4,6 @@ class = "jwst.pipeline.Image3Pipeline"
     [steps]
       [[tweakreg]]
         config_file = tweakreg.cfg
-        skip = True
       [[skymatch]]
         config_file = skymatch.cfg
       [[outlier_detection]]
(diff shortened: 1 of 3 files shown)


bhilbert4:
Looks good to me too. (Just a little bit late)

Issue #2323, opened 2018-08-02
DOC: RTD is choking again


pllim:
https://readthedocs.org/projects/jwst-pipeline/builds/7575838/

FYI

pllim:
Again. For the past 5 days. https://readthedocs.org/projects/jwst-pipeline/builds/7735066/

jdavies-st:
The build log is completely unhelpful.  Hmm.

pllim:
Did you add any new dependencies? Looks familiar.

pllim:
I feel like it is unproductive for RTD to spam me when this fails, as I cannot do anything about it. Please name a tribute to take charge of documentation from your team. Thanks.

drdavella:
Use the heimlich!

pllim:
Dan, did you just volunteer? What is your RTD username? 😄

jdavies-st:
Does anyone in the `jwst` project know anything about RTD?  How are we supposed to debug if there's no log of the build accessible?

@nden @sosey any insight here?

stscicrawford:
@pllim -- For the time being, make me admin and then we will find someone responsible from there

sosey:
I think those with package rights can see the build file. If you have an RTD account we (the people with the spacetelescope user access) can add you to the package

pllim:
I can't remove myself. Someone please remove me at https://readthedocs.org/dashboard/jwst-pipeline/users/ Thanks.

stscicrawford:
Closing this as Read the Docs is now working, but @rendinam is looking at why the SSB channel and Read the Docs are not working with each other.

PR #2516, opened 2018-09-15
fix for the RTD build

stscicrawford:
Conda on RTD seems to be having a problem with accessing the SSB astroconda channel.  This removes this for the purposes of building the docs but should be looked at again.


stscicrawford:
cc @rendinam

stscicrawford:
Address #2323 but does not fix the underlying problem with conda

PR #2523, opened 2018-09-18
Update README.md

jdavies-st:
There was some confusion from users at the image alignment hack day.  This clarifies the 2 ways one can set up the `jwst` package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ Beginning with version 0.9.0, **JWST requires Python 3.5 or above**.
 Installing
 ----------
 
+### A release ###
+
 To install a particular released version of the package, and all dependencies, we recommend using
 [conda](https://conda.io/docs/index.html) and a spec file that lists the exact versions of all packages to be installed.
 To create a new environment, use:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,6 +14,8 @@ Beginning with version 0.9.0, **JWST requires Python 3.5 or above**.
 Installing
 ----------
 
+### A release ###
+
```
stscicrawford: Installing a release version'

README.md
```diff
@@ -33,6 +35,8 @@ To update to the latest nightly build:
 
     conda update -n jwst --override-channels -c http://ssb.stsci.edu/astroconda-dev -c defaults --all
 
+### Latest dev ###
+
```
stscicrawford: Installing the latest development version

README.md
```diff
@@ -50,6 +54,8 @@ pulling the latest version of `master` from the Github repository inside the `jw
     git pull origin master
     python setup.py develop
 
+### CRDS Setup ###
+
```
stscicrawford: It would be useful to add: 'Within the STScI network, the default configuration of CRDS can be used to run the pipeline.'

sosey: oh yah, sorry, I forgot crds does keep a local cache, I always use a local cache.... but maybe we should point to the crds docs here: https://jwst-crds.stsci.edu/static/users_guide/environment.html

jdavies-st: I added the "default config" info.  I don't want to point to CRDS docs here, as I want to keep this as short and simple as possible.  In fact, the README is already too long.  =)

PR #2530, opened 2018-09-20
Added short/medium/long combinations to BAND

### LINKED ISSUE #2398: Add BAND values for MRS crossed dichroic modes
The MIRI MRS dichroic wheels (DGAA and DGAB) can be commanded to different sub-channel positions in a single exposure, such that a single BAND value of SHORT, MEDIUM, or LONG is not correct. We need to add allowed values to the data model schema for the crossed dichroic conditions of "SHORT-MEDIUM", "SHORT-LONG", "MEDIUM-SHORT", "MEDIUM-LONG", "LONG-SHORT", and "LONG-MEDIUM".

The MIRI IDT will be delivering reference files in their next CDP that use these values, hence we must get them in place so that the reference files do not fail schema validation.

stscirij:
(datamodel.meta.instrument.band) enumeration list
Fixes #2398

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -403,7 +403,8 @@ properties:
           band:
             title: MRS wavelength band
             type: string
-            enum: [SHORT, MEDIUM, LONG, N/A, ANY, MULTIPLE]
+            enum: [SHORT, MEDIUM, LONG, SHORT-MEDIUM, SHORT-LONG, MEDIUM-SHORT,
+              MEDIUM-LONG, LONG-SHORT, LONG-MEDIUM, N/A, ANY, MULTIPLE]
             fits_keyword: BAND
             blend_table: True
             blend_rule: multi


nden:
What is the context of this PR and how should these be used? 
`assign_wcs` assumes the band is the same for both MRS channels. Is this assumption wrong?

stscirij:
I just implemented what was in Issue #2398 - it only specifies that the datamodel schema needs to accommodate different values for different channels.  Based on that issue, it looks like the assumption that the values are the same is now incorrect.  Interestingly, the fits generator has been able to create files with BAND values of this type for a while.  Internally, it means that, for example, if the BAND value is MEDIUM-SHORT, it means that the DGAA_POS setting is MEDIUM and the DGAB_POS setting is SHORT.  I think this means that, for example, if the setting is MEDIUM-SHORT for a MIRIFUSHORT exposure, then the MEDIUM wavelength range is selected for the first IFU (i.e. channel 1 for MIRIFUSHORT data) and SHORT is selected for channel 2.  Since channels 1 and 4 share a grating wheel, as do 2 and 3, then the corresponding MIRIFULONG settings will be SHORT-MEDIUM.

stscirij:
@jemorrison can you confirm what I said?

hbushouse:
To my knowledge this crossed dichroic combination can only occur in engineering/test exposures - not regular science. It's not obvious to me how far the team would like to take level-2b processing for such exposures (i.e. do we need to compute an accurate WCS). We'll need team input on that. But the fact that this results in mixed sub-bands for the 2 channels that fall on a given detector (as @stscirij pointed out) does certainly complicate things.

jemorrison:
As I understand the cross dichroic combinations this will be for 
engineering tests. I believe the MIRI team needs data taken
like this to determine the spectral leak correction.  There many be 
other uses for this cross dichroic combintations in engineering
test that will be taken during commissioning. But there is no plan (to 
my knowledge) to use it for science data.

hbushouse:
It can't hurt anything in the current operations of the various cal pipeline routines to add these additional enum values, which will bring our schema in synch with the JWKD. Figuring out how to update applicable steps to handle the cross-dichroic situation can be handled later as a separate issue. I say we merge.

Issue #2266, opened 2018-07-20
Container datamodels should be indexable and iterable


jdavies-st:
Right now the datamodels that act like containers for multiple slits:
```
MultiSlitModel
MultiSpecModel
MultiProductModel
MultiExtract1dImageModel (for reference files)
MultiExposureModel
```
have multiple slits within them, but the datamodel itself is not iterable or indexable (@nden has made `MultiSlitModel` indexable currently).  Each would be much simpler to use if the model itself was indexable and iterable.

Users should not have remember or look up in documentation what the actual indexable attribute is in order to find the slits - the model itself should be indexable.

So for MultiSlitModel, instead of
```
slit = model.slits[2]

for slit in model.slits:
    print(slit.name)
```
we would prefer
```
slit = model[2]

for slit in model:
    print(slit.name)
```
Users thus won't have to remember the name of the attribute that contains the list of slits.  All of the above models name that attribute differently:
```
MultiSlitModel.slits
MultiSpecModel.spec
MultiProductModel.products
MultiExtract1dImageModel.images
MultiExposureModel.exposures
```
And when one does iterate or index the model, it should return a new datamodel, not an `ObjectNode`.  Currently, `MultiSlitModel` does this for indexing, but not iterating.  All the others return `ObjectNode`s currently.

For these datamodels, it will require defining `__getitem__` and `__iter__` methods at a minimum I believe.

philhodge:
Making it possible to iterate over the model itself would be a big improvement.  I would like to still be able to access the slit (or spec, or whatever) object as well, however.  One reason is that there are models that have more than one iterable.  The NirspecFlatModel, for example, contains two tables, an array of wavelengths and a "fast variation" table.

bernie-simon:
Are there circumstances where the code can't be expected to know which metadata attribute to iterate over?

jdavies-st:
The developers know that `MultiSlitModel`'s iterable is `slits`, but it is not immediately obvious to users, or even other developers.  It would be nice to either:

1) iterate over the model itself so one doesn't need to know what the iterable attribute is

or 

2) have an iterable attribute name that is common to all these container-type datamodels.

Both of these could be accomplished with a mix-in probably.  Maybe the 2nd option might be cleaner?

[...]

PR #2535, opened 2018-09-21
Datamodels containing submodels have a standard iteration procedure [#2266]

bernie-simon:
Some datamodels are collections of submodels. The most obvious example is ModelContainer, but there are also several Multi-Models (MultiExposureModel, MultiSlitModel, and others.) This update provides a consistent way to access the submodels. A new property, collection, has been added to DataModel. It returns a list of nodes if the schema has a single top level array. Otherwise, it raises an AttributeError. Each node represents a submodel. Because it is a list, it can be looped over or accessed by index. The supporting method to_model() has been added to ObjectNode. It converts a node to a datamodel, but only if the node contains a meta.model_type attribute. If it does not, it raises a ValueError. The method of the same name in ListNode converts the entire array to a list of models.

The corresponding methods collection and to_model() in ModelContainer allow the same code to loop over the models in a model container. The following code examples will work for both:

```
model = dmopen(filename)
node_list = model.collection
for node in node_list:
    submodel = node.to_model()
    print(submodel.info())
submodels = model.collection.to_model()
for submodel in submodels:
    print(submodel.info())
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,7 @@ assign_wcs
 
 - The bounding box for Nirspec WCS objects was modified to include the
   edges of the pixels. [#2491]
-  
+
 - Updated assign_wcs to compute the sky footprint of MIRI MRS and Nirspec
   IFU observations. [#2472]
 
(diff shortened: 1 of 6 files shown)


jdavies-st:
@nden I changed this to be PR against `0.13.0_new_features`.

jdavies-st:
Closing as the implementation here is incomplete, and I'd prefer to see this done via a mix-in.

PR #2539, opened 2018-09-24
fix flake8 problems

nden:
Ref #2340

### DIFF
--- a/jwst/cube_skymatch/cube_skymatch_step.py
+++ b/jwst/cube_skymatch/cube_skymatch_step.py
@@ -4,13 +4,11 @@ JWST pipeline step for sky matching.
 
 :Authors: Mihai Cara
 
-:License: :doc:`../LICENSE`
-
 """
 
 import numpy as np
 
-from ..stpipe import Step, cmdline
+from ..stpipe import Step
 from .. import datamodels
 
 try:
@@ -169,10 +167,10 @@ class CubeSkyMatchStep(Step):
             del m.meta.background
 
     def _set_cube_bkg_meta(self, model, sky_cube):
-        np = len(meta.background.polynomial_info)
-        if np > 1 or (np == 1 and meta.background.polynomial_info[0].channel):
+        np = len(model.meta.background.polynomial_info)
+        if np > 1 or (np == 1 and model.meta.background.polynomial_info[0].channel):
             raise ValueError("Cube's 'polynomial_info' must be empty or "
-                             "containt at most one element with a 'channel' "
+                             "contain at most one element with a 'channel' "
                              "property set to None.")
 
         pinfo = {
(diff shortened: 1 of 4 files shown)


jwst/cube_skymatch/cube_skymatch_step.py
```diff
@@ -169,10 +167,10 @@ def _reset_background(self, models):
             del m.meta.background
 
     def _set_cube_bkg_meta(self, model, sky_cube):
-        np = len(meta.background.polynomial_info)
-        if np > 1 or (np == 1 and meta.background.polynomial_info[0].channel):
+        np = len(model.meta.background.polynomial_info)
```
nden: I am assuming the intention here (and elsewhere) was to reference `model.meta` as `meta` is not defined.

jwst/cube_skymatch/cube_skymatch_step.py
```diff
@@ -186,12 +184,12 @@ def _set_cube_bkg_meta(self, model, sky_cube):
...
         if np == 0:
-            meta.background.polynomial_info.append(pinfo)
+            model.meta.background.polynomial_info.append(pinfo)
         else:
-            meta.background.polynomial_info[0] = pinfo
+            model.meta.background.polynomial_info[0] = pinfo
 
     def _set_model2d_bkg_meta(self, model3d, model2d, channel):
-        _extend_2d_schema(model2d)
+        # _extend_2d_schema(model2d)
```
nden: Not sure commenting out `extend_2d_schema` is correct but it is not defined anywhere. Should it be imported from somewhere instead?

jwst/cube_skymatch/skycube.py
```diff
@@ -417,6 +417,6 @@ def combine_with_other(self, other):
         w1 = self.cube_weight * self.weights[m]
         w2 = other.cube_weight * other.weights[m]
         tw = w1 + w2
-        d = (w1 * self.data[m] + w2 * other.skysub_data[m]) / tw
+        # d = (w1 * self.data[m] + w2 * other.skysub_data[m]) / tw
```
nden: `d` is not used. If it is not needed it should be removed from code.

jwst/cube_skymatch/skymatch.py
```diff
@@ -74,12 +66,6 @@ def match(skycubes, subtract=False):
     # Time it
     runtime_begin = datetime.now()
 
-    log.info(" ")
-    log.info("***** {:s}.{:s}() started on {}"
-             .format(__name__, function_name, runtime_begin))
-    log.info("      Version {} ({})".format(__version__, __vdate__))
-    log.info(" ")
```
nden: This kind of logging info is done automatically for each step from the base class `Step`. If more logging is needed it should not use `__vdate__ ` and `__version__`.

nden:
@mcara Please see inline questions.

jwst/cube_skymatch/cube_skymatch_step.py
mcara: Probably not. I was using it before things got hardcoded in the schema, see https://github.com/spacetelescope/jwst/pull/794

jwst/cube_skymatch/skycube.py
mcara: Agree, it seems that `self._data[m] = (w1 * self.data[m] + w2 * other.skysub_data[m]) / tw` is doing the same thing. I don't remember, but maybe I was doing something like self._data[m] = d` in the past (or when developing/debugging).

jwst/cube_skymatch/cube_skymatch_step.py
mcara: It's a good assumption.

PR #2541, opened 2018-09-24
Fix extract_2d cutout size bug

### LINKED ISSUE #2517: NRS slit xsize (and ysize?) are 1 pixel too large
Since the change to the bounding box in #2491, we've seen the following in our regression tests for NIRSpec:
```
E         Extension HDU 1:
E         
E              Keyword SLTSIZE1 has different values:
E                 a> 2049
E                  ?    ^
E                 b> 2048
E                  ?    ^
E         
```
I think the `slit.xsize` (FITS keyword SLTSIZE1) is perhaps being computed incorrectly?  Before the fix it was correctly being computed as 2048 pixels.  Now it is 2049, a physical impossibility I believe.  Perhaps here?

https://github.com/spacetelescope/jwst/blob/96105d1cd737f798b0a5d90a5ab968396b687a23/jwst/extract_2d/nirspec.py#L160

Perhaps `xstart`, `ystart` and `ysize` need to modified as well?

jdavies-st:
All the NIRSpec MSA and FS regression tests were failing because of this bug.

Resolves #2517.

### DIFF
--- a/jwst/extract_2d/nirspec.py
+++ b/jwst/extract_2d/nirspec.py
@@ -146,7 +146,7 @@ def set_slit_attributes(output_model, slit, xlo, xhi, ylo, yhi):
 
     Parameters
     ----------
-    output_model : `~jwst.datamodels.multislit.MultiSlitModel`
+    output_model : `~jwst.datamodels.SlitModel`
         The output model representing a slit.
     slit : namedtuple
         A `~jwst.transforms.models.Slit` object representing a slit.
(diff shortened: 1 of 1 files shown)


jwst/extract_2d/nirspec.py
```diff
@@ -146,20 +146,19 @@ def set_slit_attributes(output_model, slit, xlo, xhi, ylo, yhi):
...
-    output_model : `~jwst.datamodels.multislit.MultiSlitModel`
+    output_model : `~jwst.datamodels.SlitModel`
         The output model representing a slit.
     slit : namedtuple
         A `~jwst.transforms.models.Slit` object representing a slit.
     xlo, xhi, ylo, yhi : float
         Indices into the data array where extraction should be done.
         These are converted to "pixel indices" - the center of a pixel.
     """
-    xlo_ind, xhi_ind, ylo_ind, yhi_ind = _toindex((xlo, xhi, ylo, yhi)).astype(np.int16)
```
jdavies-st: This call to `_toindex` is superfluous as the inputs to this function have already been run through `_toindex` in the `offset_wcs()` function.

jwst/extract_2d/nirspec.py
```diff
@@ -221,24 +220,22 @@ def extract_slit(input_model, slit, exp_type):
...
-        ext_var_poisson = input_model.var_poisson[ylo: yhi + 1, xlo: xhi + 1].copy()
+    ndim = len(input_model.data.shape)
+    if ndim == 2:
+        slit_slice = np.s_[ylo: yhi, xlo: xhi]
+        ext_data = input_model.data[slit_slice].copy()
+        ext_err = input_model.err[slit_slice].copy()
+        ext_dq = input_model.dq[slit_slice].copy()
+        ext_var_rnoise = input_model.var_rnoise[slit_slice].copy()
+        ext_var_poisson = input_model.var_poisson[slit_slice].copy()
         int_times = None
```
jdavies-st: Since `xlo, xhi, ylo, yhi` are now generated from bounding boxes of the form `((-0.5, 2047.5))`, the result from `_toindex` is `((0, 2048))`.  So no need to add 1 to these anymore.

PR #2542, opened 2018-09-24
flake8 issues fixed

nden:
Ref #2340
These were all trivial so I fixed them in one PR.

After this PR is merged what's left in #2340 is `associations` , `jwpsf`, `ramp_fitting`, `fits_generator` and `exp_to_source`. I'll open a separate issue for them (for now they are ignored in setup.cfg). I'd like to merge this PR followed by #2340 so that Travis starts running the flake8 test. Otherwise every new PR has the potential to introduce new flake8 problems in steps that were already fixed.


nden:
The last commit here shows that we constantly introduce new flake8 issues so I will merge this PR after tests finish. This PR supersedes #2340 and addresses #2329.
It also skips the checks for `associations`, `exp_to_source`, `fits_generator`, `lib`, `ramp_fitting`  and `jwpsf`, for which  #2552 was filed.

PR #2545, opened 2018-09-25
Attempt to fix ResourceWarning in datamodel unit test

### LINKED ISSUE #2544: master failing on jenkins
https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwst/detail/PR-2542/1/pipeline#step-94-log-1081

bernie-simon:
The unit test `datamodels/tests/test_models.py::test_open_warning` was failing on `ssbjenkins` unit test build because when testing for a warning message about DATAMODL not being found, it gets a ResourceWarning about a file being left open instead. I believe the error is caused by a fits file being left open, because the code which schedules the fits file to be closed is after the warning message. When a warning filter is set to convert the warning to an error, this code is never executed. So I have moved the code that schedules the fits file to be closed before the warning message and modified the test so that is done outside the with block that opens the datamodel. This should make sure the fits file is closed and get rid of the ResourceWarning.

Resolves #2544.


nden:
Resolves #2544

jdavies-st:
Looks like we're still failing.  😕 

https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwst/detail/PR-2545/1/pipeline/#step-94-log-1081

PR #2589, opened 2018-09-27
Add validate_required_fields to DataModel

bernie-simon:
An issue was raised during a meeting that datamodels was not correctly
validating schema items with the fits_required attribute. I did some
checking and found that fits_required is used in validation when
reading a fits file or deleting a model value. But it was not used in
validation when writing a model. The reason is that when writing the
code traverses the model, not the schema, so fields that are missing
in the model are not detected. I've added a new method,
validate_required_fields, to model_base.py that traverses the schema
and checks all fields where fits_required is true. This method is
called in save() before writing the file. It honors the usual flags,
pass_invalid_values and strict_validation.
   
I also added two tests to test_models.py to check the handling of
fits_required. Fits_required is not currently used in any of the
schemas, so for the present the new method has no effect.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -335,6 +335,27 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
                               self._pass_invalid_values,
                               self._strict_validation)
 
+    def validate_required_fields(self):
+        """
+        Walk the schema and make sure all required fields are
+        in the model
+        """
+        def callback(schema, path, combiner, ctx, recurse):
+            if 'fits_required' not in schema:
+                return
+
+            node = ctx
+            for attr in path:
+                node = getattr(node, attr)
+                if node is None:
+                    break
+
+            validate.value_change(path, node, schema,
+                                  ctx._pass_invalid_values,
+                                  ctx._strict_validation)
+
+        mschema.walk_schema(self._schema, callback, ctx=self)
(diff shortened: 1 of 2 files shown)


jwst/datamodels/model_base.py
```diff
@@ -335,6 +335,27 @@ def validate(self):
...
+    def validate_required_fields(self):
+        """
+        Walk the schema and make sure all required fields are
+        in the model
+        """
+        def callback(schema, path, combiner, ctx, recurse):
+            if 'fits_required' not in schema:
+                return
+
+            node = ctx
```
stscieisenhamer: Curiosity question: After this loop, there are only three possible outcomes: node is ctx, node is None, or node is the last attribute. Any other attributes in path are ignored. Is this intended?

bernie-simon: The path is what you get after you split the schema field you are looking for, For example, 'meta.program.title' becomes ['meta', 'program', 'title']. The reason for the test for None is that one of the intermediates may be missing. In this example, there may be 'meta', but no 'meta.program', in which case there is also no 'meta.program.title'. Checking for None prevents the error of trying to find 'title' as the attribute of None and also returns the correct result from the loop, which is None.

stscieisenhamer: A comment to this effect may be in order, otherwise 👍

PR #2548, opened 2018-09-25
Undo temporary workaround implemented in 2404

mcara:
This PR undoes the temporary workaround implemented in https://github.com/spacetelescope/jwst/pull/2404 per https://github.com/spacetelescope/gwcs/pull/174#issuecomment-424439826

PR #2590, opened 2018-09-27
add error messages to check_wcs_structure

nden:
Attempt to fix #2548. This PR only adds error messages to the `_check_wcs_structure` method in order to facilitate debugging. I also fixed one "bug" - the way to get the names of all frames in the WCS pipeline is `wcs.available_frames`. The way it was done a list of frame objects was returned.

When I run the regression test now I get a different error. It is looking for a frame called `v2v3corr`. I thought at this point in the code `v2v3corr` should not be in the pipeline but I am not very familiar with the code. @mcara
```
Traceback (most recent call last):
  File "scripts/strun", line 26, in <module>
    step = Step.from_cmdline(sys.argv[1:])
  File "/internal/1/astropy/jwst/jwst/stpipe/step.py", line 162, in from_cmdline
    return cmdline.step_from_cmdline(args)
  File "/internal/1/astropy/jwst/jwst/stpipe/cmdline.py", line 297, in step_from_cmdline
    step.run(*positional)
  File "/internal/1/astropy/jwst/jwst/stpipe/step.py", line 400, in run
    step_result = self.process(*args)
  File "/internal/1/astropy/jwst/jwst/tweakreg/tweakreg_step.py", line 114, in process
    wcsimlist = list(map(self._imodel2wcsim, g))
  File "/internal/1/astropy/jwst/jwst/tweakreg/tweakreg_step.py", line 170, in _imodel2wcsim
    meta={'image_model': image_model}
  File "/internal/1/astropy/jwst/jwst/tweakreg/wcsimage.py", line 360, in __init__
    self.set_wcs(wcs, ref_angles)
  File "/internal/1/astropy/jwst/jwst/tweakreg/wcsimage.py", line 411, in set_wcs
    dec_ref=ref_angles['dec_ref']
  File "/internal/1/astropy/jwst/jwst/tweakreg/wcsimage.py", line 65, in __init__
    valid, message =  self._check_wcs_structure(wcs)
  File "/internal/1/astropy/jwst/jwst/tweakreg/wcsimage.py", line 244, in _check_wcs_structure
    idx_v2v3corr = frms.index('v2v3corr')
ValueError: 'v2v3corr' is not in list
```

PR #2594, opened 2018-09-27
Use wcs.available_frames in tweakreg

jdavies-st:
Resolves the issue in #2548.

Regression tests now pass with this fix.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -212,6 +212,8 @@ tso_photometry
 tweakreg
 --------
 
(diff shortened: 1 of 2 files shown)


nden:
Ah, I see. I added a couple more changes to the other PR.
I prefer to have the error messages included as this is the second time I struggle with a tweakreg error with the cryptic "WCS structure error".

jdavies-st:
Ah, ok.  No problem.  I'll close this PR and test yours.

jdavies-st:
#2590 resolves this in a better way, so closing this PR.

Issue #2592, opened 2018-09-27
Add DATE-BEG keyword to core schema


hbushouse:
JWST Keyword Dictionary ticket https://jira.stsci.edu/browse/JWSTKD-87 added the new keyword DATE-BEG. Need to add this to our yaml schema to stay in synch with the JWKD.

DATE-BEG:
 fits_comment: Date-time start of data acquisition
 data type: string (or optionally "datetime"), formatted as a ISO-8601 string containing both date and time
 data model name: observation.date_beg

Note that in keeping with the FITS time standards for this keyword, the value contains a complete datetime string, as opposed to the similar existing keywords DATE-END and TIME-END, which contain date and time separately.

hbushouse:
Fixed by #2595

Issue #2591, opened 2018-09-27
Add PWFSEET and NWFSEST keywords to core schema


hbushouse:
The JWST Keyword Dictionary ticket https://jira.stsci.edu/browse/JWSTKD-60 added the keywords PWFSEET and NWFSEST, so we need to add them to our data model schema too. Information is as follows:

PWFSEET:
 fits_comment: Previous WFS exposure end time
 data_type: float, in MJD format and units of UTC
 data model name: pwfseet

NWFSEST:
 fits_comment: Next WFS exposure start time
 data_type: float, in MJD format and units of UTC
 data model name: nwfsest

hbushouse:
Fixed by #2595

PR #2529, opened 2018-09-19
Translation of all pipelines tests 

stsci-hack:
The changes here redefine all the tests from 'general/pipelines/' to work with Artifactory using ci_watson while moving these tests from 'general/pipelines/' into the instrument-specific test directory appropriate for the test.  

Nearly all of these newly defined tests have been run interactively during development to insure that they return the same results as the currently defined nightly tests.

**WARNING:  This PR not only adds ci_watson as a dependency for the package, but requires the use of changes submitted for ci_watson in [PR #14](https://github.com/spacetelescope/ci_watson/pull/14).  This PR also requires availability of data as organized for Artifactory (whether that is a new local directory or on Artifactory itself, depending on when the data gets moved).**

PR #2595, opened 2018-09-27
Keyword additions and changes

hbushouse:
Add new keywords requested in #2591 and #2592.

Also update many keyword title/comment fields to conform to the FITS standard recommendation that units appearing in keyword comments should be enclosed in **square brackets** at the **beginning** of the comment field, and use IAU-recommended units strings. This will cause LOTS of regression test failures (sorry about that).

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -75,6 +75,16 @@ properties:
         type: string
         fits_keyword: HGA_STOP
         blend_table: True
+      pwfseet:
+        title: Previous WFS exposure end time
+        type: number
+        fits_keyword: PWFSEET
+        blend_table: True
+      nwfsest:
+        title: Next WFS exposure start time
+        type: number
+        fits_keyword: NWFSEST
+        blend_table: True
       asn:
         title: Association information
         type: object
@@ -139,6 +149,14 @@ properties:
             title: "[hh:mm:ss.sss] UTC time at start of exposure"
             type: string
             fits_keyword: TIME-OBS
+            blend_table: True
+            blend_rule: first
+          date_beg:
+            anyOf:
+              - $ref: http://stsci.edu/schemas/asdf/time/time-1.0.0
+              - type: string
+            title: "Date-time start of data acquisition"
+            fits_keyword: DATE-BEG
             blend_table: True
             blend_rule: first
           date_end:
@@ -363,7 +381,17 @@ properties:
           channel:
             title: 'NIRCam channel: long or short'
             type: string
-            enum: [LONG, SHORT, '12', '34', '1', '2', '3', '4', ANY, N/A, '1234', '123', '234']
+            # Values grouped by instrument:
+            anyOf:
+              # MIRI
+              - enum:
+                 ['1', '2', '3', '4', '12', '34', '123', '234', '1234']
+              # NIRCam
+              - enum:
+                  [LONG, SHORT]
+              # All
+              - enum:
+                  [ANY, N/A]
             fits_keyword: CHANNEL
             blend_table: True
           filter:
@@ -495,7 +523,7 @@ properties:
             fits_keyword: GWA_PYAV
             blend_table: True
           gwa_tilt:
-            title: GWA TILT (avg/calib) temperature [K]
+            title: "[K] GWA TILT (avg/calib) temperature"
             type: number
             fits_keyword: GWA_TILT
             blend_table: True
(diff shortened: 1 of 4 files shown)


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -141,6 +151,14 @@ properties:
             fits_keyword: TIME-OBS
             blend_table: True
             blend_rule: first
+          date_beg:
+            anyOf:
+              - $ref: http://stsci.edu/schemas/asdf/time/time-1.0.0
```
philhodge: Did you intend to use the "$ref" for date_end as well?

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -658,28 +686,28 @@ properties:
...
             blend_table: True
           frame_time:
-            title: Time between frames (sec)
+            title: "[s] Time between frames"
             type: number
             fits_keyword: TFRAME
             blend_table: True
           group_time:
-            title: Time between groups (sec)
+            title: "[s] Time between groups"
```
philhodge: It looks as if there are trailing blanks at the end of the group_time title.  Was that intentional?

jwst/datamodels/schemas/core.schema.yaml
hbushouse: I guess we could use that for DATE-END as well, because it is also in the format of a datetime string (like the DATE keyword, which also uses the time schema reference).

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -75,6 +75,16 @@ properties:
...
+      pwfseet:
+        title: Previous WFS exposure end time
+        type: number
+        fits_keyword: PWFSEET
+        blend_table: True
+      nwfsest:
+        title: Next WFS exposure start time
+        type: number
+        fits_keyword: NWFSEST
+        blend_table: True
```
jdavies-st: Since there's no limit on characters for names in the meta tree, would it make sense for `meta.pwfseet` and `meta.nwfsest` to be a bit more descriptive?  `prev_wfs_exp_end_time`?  or similar?

jwst/datamodels/schemas/image.schema.yaml
```diff
@@ -53,7 +53,6 @@ allOf:
 - type: object
   properties:
     meta:
-      title: Level 3 Schema Metadata
```
jdavies-st: 🎉

jwst/datamodels/schemas/core.schema.yaml
philhodge: Yes!

hbushouse: I agree, but unfortunately SDP already added these to the Keyword Dictionary and they chose the short (same as the keyword) names, so we have to match. Someday when we're all done with the pipeline and we have nothing to do, I intend to go back and have SDP+CAL do a thorough scrub of the KWD to clean up and improve lots of things like this. Fortunately right now we have no use for these keywords in the Cal pipeline, so we don't need to worry about putting the cryptic names in our code.

jdavies-st: Wait, does SDP use our datamodel schemas?  In the JWST Keyword Dictionary, there's no reference to what we call it in the pipeline, at least available through the web interface.

I thought they only cared about what the FITS keyword was named.

hbushouse: No, they don't use our data model schemas or even use the data model names that are in the KWD (at least not yet), but there's a comparator that gets run on two sets of schemas relatively frequently that points out all the differences, which then means someone (i.e. me) has to take time to investigate and explain them all the time. So for now it's just easier to keep them matched up. I have on several occasions had the KWD get changed to match us. But the week before a delivery deadline is not the time to make one of those requests.

jdavies-st: So the comparison is between their schemas and our schemas.  But if we're only comparing the FITS keyword, are we not free to make the name in `meta` that maps to the FITS keyword something a bit more descriptive?  We do that everywhere in our schemas.

hbushouse: The KWD schemas have data model names mapped to keyword names, just like our schemas, because their schemas were created as a copy of ours years ago (or at least they used ours as a template). They add many more attributes to each keyword than we do, for things like sql_datatype, long description, units, example values, calculation to derive the keyword value, etc.

jdavies-st:
So this PR broke **all** of our regression tests.

https://boyle.stsci.edu:8081/job/RT/job/JWST/test_results_analyzer/

Can we back it out?

jdavies-st:
The issue is that we're in the middle of making big changes to the regression tests (see #2529).  Reorganizing the input and truth data and getting onto Artifactory.  And if we have OK-ify all of these, then it will pretty much require us to do all that reorg again.

jdavies-st:
Or perhaps the copying of file from `/data4` on jwcalibdev is scripted already?  @stsci-hack ?

hbushouse:
The keyword updates in this PR need to get into our B7.2 delivery, so I guess it could be backed out temporarily until we make our delivery on 10/10.

jdavies-st:
I see.  If we can OK-ify these today and make sure we keep track which files have been updated in `/data4`, then it will help in getting these updated on Artifactory as well.  Currently there's a parallel `/data4` with a slightly different structure that is getting mapped into Artifactory, but we haven't switched over to it yet because the software side is still in a bit of flux.

PR #2598, opened 2018-09-27
Nonlinear wavelength for cube_build

jemorrison:
This PR build can create cubes with a nonlinear wavelength dimension. The wavelength bins to use in
the IFU cube are read in from the CUBEPAR reference file.  These types of IFU cubes are created when the input data is from different bands (ie. Ch 1 Short and Ch 1 Medium or for NIRSPEC different filters for the same grating). 
This PR also uses the wavelength ranges given in the CUBEPAR reference file to build cube from
a single band rather than determining the maximum and minimum values from the data. This PR will result in the regression tests failing because the min and max wavelength of the ifucube are slightly different. 
When an ifucube is made from more than 1 band then additional table extensions in the CUBEPARS are used to define the wavelength sampling and the roi to use for each wavelength region. These additional binary table extensions are now being read in with this PR.  One of the requirements if detector pixel is matched to a cube spaxel  is if  falls within a roi in the wavelength dimension set by spectral roi in the wavelength table. If building non-linear wavelength cubes then this
roi spectral varies with wavelength. Each detector pixel has its own associated spectral roi. Adding this feature has slowed down the matching of the detector pixels to the ifucube. 
If an ifucube has a non-linear wavelength dimension then the wavelength planes of the IFU are written to a wavelength array that is contained in the additional FITS extension WCS-TABLE. To make this output file adhere to the FITS standard the TDIM2 keyword in the WCS-TABLE needs to be populated with a string containing the size of the wavelength array. This has not been done in this PR (the keyword is missing when the output FITS file is written).
11 files where changed in this PR with ~1000 lines of code changes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,7 +42,7 @@ csv_tools
 
 cube_build
 ----------
-This version supports creating IFU Cubes with non-linear wavelength dimension.
+This version supports creating IFU Cubes with non-linear wavelength dimension. [#2598]
 
 
 cube_skymatch
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,24 +44,6 @@ cube_build
 ----------
 This version supports creating IFU Cubes with non-linear wavelength dimension.
 
-IFU cubes made from multiple band may have a non-linear wavelength dimension. 
-The wavelength ranges, spectral size, spectral roi size to use to build the
-IFU Cube  are read in from the CUBEPAR reference files for each band. If the 
-spectral size is not the same for all the bands then the IFU cubes will have
-a nonlinear wavelength dimension. The wavelength to use for the output IFU 
-are found in additional table extensions in the CUBEPAR reference file. 
-In addition, the spectral region of influence can also vary as a function
-of wavelength for these type IFU cubes. The spectral region of influence
-is also found in the additional tables in the CUBEPAR reference file.
-The output of these non-linear IFU cubes contain the wavelength planes of
-the IFU cube in the extension table 'WCS-TABLE' 
-
-This version also uses the minimum and maximum wavelengths defined in the 
-first extension of the CUBEPAR reference file to set the wavelength range
-of the IFU Cube for a linear wavelength dimension. Before the wavelength range
-was calculated from the data. The values found in the CUBEPAR reference file
-clips the wavelength ranges so that the boundaries of the IFU cube do not
-contain sparse data.  
 
 cube_skymatch
 -------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,8 +42,26 @@ csv_tools
 
 cube_build
 ----------
+This version supports creating IFU Cubes with non-linear wavelength dimension.
 
+IFU cubes made from multiple band may have a non-linear wavelength dimension. 
+The wavelength ranges, spectral size, spectral roi size to use to build the
+IFU Cube  are read in from the CUBEPAR reference files for each band. If the 
+spectral size is not the same for all the bands then the IFU cubes will have
+a nonlinear wavelength dimension. The wavelength to use for the output IFU 
+are found in additional table extensions in the CUBEPAR reference file. 
+In addition, the spectral region of influence can also vary as a function
+of wavelength for these type IFU cubes. The spectral region of influence
+is also found in the additional tables in the CUBEPAR reference file.
+The output of these non-linear IFU cubes contain the wavelength planes of
+the IFU cube in the extension table 'WCS-TABLE' 
 
+This version also uses the minimum and maximum wavelengths defined in the 
+first extension of the CUBEPAR reference file to set the wavelength range
+of the IFU Cube for a linear wavelength dimension. Before the wavelength range
+was calculated from the data. The values found in the CUBEPAR reference file
+clips the wavelength ranges so that the boundaries of the IFU cube do not
+contain sparse data.  
 
 cube_skymatch
 -------------
--- a/jwst/cube_build/blot_cube_build.py
+++ b/jwst/cube_build/blot_cube_build.py
@@ -214,7 +214,8 @@ class CubeBlot():
 
             if self.instrument == 'MIRI':
                 valid3 = np.isfinite(lam_det)
-                good_data = valid3 & pixel_mask
+                good_data1 = valid3 & pixel_mask
+                good_data = np.where(good_data1)
             elif self.instrument == 'NIRSPEC':
                 good_data = np.where(flag_det == 1)
 
@@ -222,7 +223,6 @@ class CubeBlot():
             ra_blot = ra_det[good_data]
             dec_blot = dec_det[good_data]
             wave_blot = lam_det[good_data]
-            
             crval1 = model.meta.wcsinfo.crval1
             crval2 = model.meta.wcsinfo.crval2
 
--- a/jwst/cube_build/cube_build_step.py
+++ b/jwst/cube_build/cube_build_step.py
@@ -95,7 +95,6 @@ class CubeBuildStep (Step):
         # 2. world
         self.interpolation = 'pointcloud' # true for self.weighting  = 'msm' or 'miripsf'
 
-
         # if the weighting is area then interpolation is area
         # if the weighting is area then interpolation is area. Weighting of area or interpolation
         # of area is only for single band data.
@@ -294,13 +293,13 @@ class CubeBuildStep (Step):
             else:
                 result = thiscube.build_ifucube()
                 cube_container.append(result)
-
             if self.debug_pixel == 1:
                 self.spaxel_debug.close()
         for cube in cube_container:
             footprint = cube.meta.wcs.footprint(axis_type="spatial")
             update_s_region_keyword(cube, footprint)
 
+        
         return cube_container
 
 #********************************************************************************
@@ -329,10 +328,6 @@ class CubeBuildStep (Step):
         """
         valid_channel = ['1', '2', '3', '4', 'all']
         valid_subchannel = ['short', 'medium', 'long', 'all']
-#        valid_fwa = ['f070lp', 'f100lp', 'f100lp', 'g170lp',
-#                    'g170lp', 'f290lp', 'f290lp', 'clear', 'all']
-#        valid_gwa = ['g140m', 'g140h', 'G140M', 'g140h', 'g235m', 'g235h',
-#                    'g395m', 'g395h', 'prism', 'all']
 
         valid_fwa = ['f070lp', 'f100lp', 'g170lp',
                     'g170lp', 'f290lp', 'clear', 'all']
--- a/jwst/cube_build/cube_build_step.py
+++ b/jwst/cube_build/cube_build_step.py
@@ -273,7 +273,7 @@ class CubeBuildStep (Step):
 # if linear wavelength (single band) single values for rois, roiw, weight_power, softrad
 # if not linear wavelength (multi bands) an array based on wavelength for:
 #  rois, roiw, weight_power, softrad
-
+ 
             thiscube.determine_cube_parameters()
 
             thiscube.setup_ifucube_wcs()
@@ -298,6 +298,7 @@ class CubeBuildStep (Step):
         for cube in cube_container:
             footprint = cube.meta.wcs.footprint(axis_type="spatial")
             update_s_region_keyword(cube, footprint)
+
 
         
         return cube_container
--- a/jwst/cube_build/cube_build_step.py
+++ b/jwst/cube_build/cube_build_step.py
@@ -95,7 +95,11 @@ class CubeBuildStep (Step):
         # 2. world
         self.interpolation = 'pointcloud' # true for self.weighting  = 'msm' or 'miripsf'
 
+
         # if the weighting is area then interpolation is area
+        # if the weighting is area then interpolation is area. Weighting of area or interpolation
+        # of area is only for single band data.
+
         if self.weighting == 'area':
             self.interpolation = 'area'
             self.coord_system = 'alpha-beta'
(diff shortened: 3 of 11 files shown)


jdavies-st:
How do I run `cube_build` to be able to get the non-linear wavelengths to test this PR?  Is there a change to the interface to `CubeBuildStep` to allow specifying non-linear wavelengths, or does it just do it be default now?

jwst/cube_build/ifu_cube.py
```diff
@@ -728,46 +559,54 @@ def build_ifucube(self):
...
                                                       flux,
-                                                      coord1, coord2, wave)
+                                                      coord1, coord2, wave,
+                                                      rois_pixel, roiw_pixel, weight_pixel,
+                                                      softrad_pixel)
 
 
                         t1 = time.time()
-                        log.info("Time match point cloud to ifucube = %.1f.s" % (t1 - t0,))
+                        log.info("Time to match file to ifucube = %.1f.s" % (t1 - t0,))
```
hbushouse: Eventually we probably don't want this kind of message to be at the regular "info" level, but perhaps "debug" instead. We really don't need timing info printed out in all processing logs for normal runs.

jwst/cube_build/ifu_cube.py
```diff
@@ -908,41 +750,275 @@ def determine_roi_size(self):
...
+            wave_roi = roiw[0]
+            self.soft_rad = softrad[0]
+            weight_power = power[0]
+        else:
+            self.linear_wavelength = False
+            if self.instrument == 'MIRI':
+                table = self.instrument_info.Get_multichannel_table()
+                table_wavelength, table_sroi, table_wroi, table_power, table_softrad = table
+
+# getting MIRI Table Values
```
hbushouse: The next lines of code are dealing with NIRSpec, but the comment says MIRI.

jwst/cube_build/ifu_cube.py
```diff
@@ -908,41 +750,275 @@ def determine_roi_size(self):
...
+    # done looping over files determine final size of cube
+
+        final_a_min = min(a_min)
+        final_a_max = max(a_max)
+        final_b_min = min(b_min)
+        final_b_max = max(b_max)
+        final_lambda_min = min(lambda_min)
+        final_lambda_max = max(lambda_max)
+
+        print('wavelengths', final_lambda_min, final_lambda_max, self.wavemin, self.wavemax)
```
hbushouse: Shouldn't have bare print statements. Change to "log.debug()" or remove.

jwst/cube_build/ifu_cube.py
jemorrison: Ok this was a request from the MIRI team so I stuck it in. They wanted to keep track of what ws going on because it was taking so long. I think after some more efficiency improvements I can take this out.

jemorrison:
To run cube_build and produce non-linear wavelength dimension then first the input data needs to contain more than 1 band of data. For MIRI a single exposure always contains more than 1 band of data (because there are 2 channels per exposure). For NIRSPEC the input association table needs to contain exposure from more than one grating/filter. In addition, the option --output_type='multi'
needs to be used to create an IFUcube from multiple bands. If the data from the multiple bands has a different spectral size defined in the cubepar reference table then the output IFUcube will have non-linear wavelength dimension.

jwst/cube_build/blot_cube_build.py
```diff
@@ -214,15 +214,15 @@ def blot_images(self):
 
             if self.instrument == 'MIRI':
                 valid3 = np.isfinite(lam_det)
-                good_data = valid3 & pixel_mask
+                good_data1 = valid3 & pixel_mask
+                good_data = np.where(good_data1)
```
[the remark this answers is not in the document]
jdavies-st: Instead of computing `valid` for every slice and trying to replace NaNs for every slice, just go ahead and dump everything from `ra` into `ra_det` (and the others), and then do a single call at the end, replacing NaNs with zeros:
```
notvalid = np.isnan(lam_det)
lam_det[notvalid] = 0
ra_det[notvalid] = 0
dec_det[notvalid] = 0
flag_det[lam_det != 0.] = 1
```
or something similar.

Btw, `np.zeros` produces floats by default, so `flag_det` above has a `dtype='float64'` by default.  If you want it to be an int, you need
```
flag_det = np.zeros((2048, 2048), dtype=np.int)
```

jwst/cube_build/cube_build_step.py
```diff
@@ -239,7 +242,8 @@ def process(self, input):
 # or (grating,filter)
 
         num_cubes, cube_pars = cubeinfo.number_cubes()
-        if not self.single: self.log.info('Number of ifucubes produced by a this run %i', num_cubes)
+        if not self.single: self.log.info('Number of ifucubes produced by a this run %i',
+                                          num_cubes)
```
jdavies-st: Move `self.log.info` to a new line after the `if` statement.

jwst/cube_build/cube_cloud.py
```diff
@@ -60,53 +62,70 @@ def match_det2cube_msm(naxis1, naxis2, naxis3,
...
+#               ilow = ilow + 1
+#            elif wave[ipt] > zcoord[-1]: 
+#               ihigh = ihigh + 1
+#            else:
+#                imatch = imatch + 1
+#                print(' no z match found ',wave[ipt],roiw_pixel[ipt])
+#                print(zcoord[naxis3-11:naxis3])
+#                diff = abs(zcoord[naxis3-11:naxis3] - wave[ipt])
+#                print(diff)
+#                exit()
```
jdavies-st: Remove these commented-out lines (and the ones below).

jwst/cube_build/instrument_defaults.py
```diff
@@ -494,47 +494,26 @@ def SetSpatialSize(self, value, parameter1, parameter2=None):
             self.Info[parameter1][parameter2]['bscale'] = value
 
 
-    def SetSpectralStep(self, value, parameter1, parameter2=None):
-        if parameter2 is None: # data is NIRSPEC paramters do not vary with filter
-            self.Info[parameter1]['wscale'] = value
-        else:
-            self.Info[parameter1][parameter2]['wscale'] = value
+    def SetSpectralStep(self, value, parameter1, parameter2):
+        self.Info[parameter1][parameter2]['wscale'] = value
```
jdavies-st: `self.Info` should be `self.info` as it is a dict, not a class.  Not necessary for this PR, but just an observation.

jdavies-st: Similarly, these `SetSpectralStep` functions should be `set_spectral_step`, as they are methods, not classes.  Again, not necessary change for this PR, but to get things more [PEP8 compliant](https://www.python.org/dev/peps/pep-0008/?#naming-conventions).

jwst/cube_build/blot_cube_build.py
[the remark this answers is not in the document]
jemorrison: I would rather not make such a change now with the build 7.2 due - the latest tomorrow. It took me a long time to get this figured out for NIRSPEC because I had to do things slice by slice. If I make a change to it now it takes a while to do all the testing needed. I will put it on the list of first things to do after 7.2.

jwst/cube_build/cube_cloud.py
jemorrison: The reason I left them in is because I am waiting for a new and improved MIRI cubepars reference file. The current reference file for long wavelengths is too fine in the wavelength dimension or the roi in the spectral dimension is too small. There are detector values that do not match with any cube spaxel. Ben Sargent makes the cubepars reference files and is waiting for an improved MIRI simulator to make work out what the values in the reference file should be. 
I left this code in because I can check to make sure the new reference file is working better. 
Once we have this sorted out I can see if I want to make this a debug option.

I did not want to do that at this time - but I could if you prefer ?

jwst/cube_build/instrument_defaults.py
jemorrison: ok I will add that to the list of things to change after Build 7.2

jemorrison:
I still have not gotten the wavelength dimension to be written out using the wcsinfo.schema.yaml
file. If you could take a look at that file and see if I have wavedim correct in that file.

jemorrison:
@jdavies-st 
If you want some data to test with I have put some on
/grp/jwst/ssb/morrison/ForJames 
I am putting MIRI and NIRSPEC data

jemorrison:
@jdavies-st Ok I just have miri data on that location. I forgot I do not have any multiple band data for the same area on the sky for NIRSPEC. I just have dithered data with the same grating and filter.

jwst/cube_build/cube_build_step.py
```diff
@@ -320,10 +329,6 @@ def read_user_input(self):
         """
         valid_channel = ['1', '2', '3', '4', 'all']
         valid_subchannel = ['short', 'medium', 'long', 'all']
-#        valid_fwa = ['f070lp', 'f100lp', 'f100lp', 'g170lp',
-#                    'g170lp', 'f290lp', 'f290lp', 'clear', 'all']
-#        valid_gwa = ['g140m', 'g140h', 'G140M', 'g140h', 'g235m', 'g235h',
-#                    'g395m', 'g395h', 'prism', 'all']
 
         valid_fwa = ['f070lp', 'f100lp', 'g170lp',
                     'g170lp', 'f290lp', 'clear', 'all']
```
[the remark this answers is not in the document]
jemorrison: fixed this

jemorrison:
@jdavies-st @hbushouse 
I believe except for changing the blot_cube code (which I will do for Build 7.3) all the requested changes have been made. I think it is ready to merge

CHANGES.rst
```diff
@@ -42,7 +42,7 @@ csv_tools
 
 cube_build
 ----------
-
+This version supports creating IFU Cubes with non-linear wavelength dimension.
```
hbushouse: Add the PR number to the end of the message.

jwst/cube_build/cube_build_step.py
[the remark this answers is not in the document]
hbushouse: But now that I look at it again, shouldn't "g170lp" be "**f**170lp" (it's a filter name, right? Not a grating name)

jdavies-st:
The non-linear wavelengths in the WAVE-TAB extension still don't work though, right?  I've been looking at this last night and this morning.

jwst/cube_build/cube_build_step.py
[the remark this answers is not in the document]
jemorrison: YES it should!

jemorrison:
The non-linear wavelength table is written and if you read it with datamodels the correct wavelengths are given. What is not working is writing a simple keyword to the WCS-TABLE extension header. This will make the FITS files fully standard.  I have downloaded the newest ds9 beta version 8r5 and testing
reading in the files after I add the TDIM2 keyword

jemorrison:
I have confirmed that if I take the output ifucube and write the TDIM2 keyword that contains the size of the wavelength table as a string to the WCS-TABLE header then the new beta version of DS9 correctly
reads in the wavelengths.  I will keep looking into why this is not being written. Do we have a list of allowed keyword some place and only those keywords get written to the output fits file ?

jdavies-st:
Ok, I was not able to figure it out either.  The schema really mangles any numpy array that is passed in.

That said, I suspect that the TDIMn keyword does not have to be added by hand.  It gets added automatically to Bintable HDUs just like the TFORM keywords.

I do notice that the schema indicates the output WAVE-TAB has 2 columns labeled `nelem` and `wavelength`.  I was under the impression that it was only to have one column?

Anyway, I'm approving these changes.  Could you edit the inital PR comment at the top to indicate what else has changed in the ~1000 lines of code changes across 11 files?

Ready to squash/merge.

philhodge:
I don't see (just looking at the changes) where you assign a value to wavedim.

jemorrison:
It is ifu_cube.py
Right now I hard code the value to '(1,10420)' it is set to the keyword: TDIM2
This is because I have to set the shape of wavetable in the schema. I set it to the maximum number of
wavelengths that is in the reference file - 10420. It is not ideal and hopefully we can improve it.
Right now I run cube_build and then I have a separate piece of personal code that adds TDIM2
to the header of the WCS-TABLE. When I do that then the new beta version of ds9 understands 
the wavelength dimension. So I am assuming that is the FITS standard format.

philhodge:
I thought you had a 1-D array in each cell.  If that's the case, TDIM2 should be '(10420)'.  It's not the dimension of the column; it's the dimension of the array in each cell.

jemorrison:
The table has two columns. The first column is the number of valid wavelengths (out of the 10420) and
the second column is the wavelength table - it has a shape of 1,10420
I can try making it 10420 and see what happens but the shape of table in the cell is 1,10420. That was one of the reasons I had so much trouble making this wavelength table

philhodge:
OK, I see what you mean; the array of wavelengths has to be that shape because it's the entire column, not just a cell (even though there's only one row).  But TDIM2 should be '(10420)'.  Here's an example, using astropy.io.fits:

    import numpy as np
    from astropy.io import fits
    fd = fits.HDUList(fits.PrimaryHDU())
    nelem = np.zeros(1, dtype=np.int32) + 10
    values = np.arange(10, dtype=np.float32).reshape(1, 10)
    col1 = fits.Column(name='nelem', format='J', array=nelem)
    col2 = fits.Column(name='wavelength', format='10E', dim='(10)', array=values)
    coldefs = fits.ColDefs([col1, col2])
    hdu = fits.BinTableHDU.from_columns(coldefs)
    fd.append(hdu)
    fd.writeto('example.fits')
    fd.close()

    catfits.py example.fits[1] yes
    XTENSION= 'BINTABLE'           / binary table extension
    BITPIX  =                    8 / array data type
    NAXIS   =                    2 / number of array dimensions
    NAXIS1  =                   44 / length of dimension 1
    NAXIS2  =                    1 / length of dimension 2
    PCOUNT  =                    0 / number of group parameters
    GCOUNT  =                    1 / number of groups
    TFIELDS =                    2 / number of table fields
    TTYPE1  = 'nelem   '
    TFORM1  = 'J       '
    TTYPE2  = 'wavelength'
    TFORM2  = '10E     '
    TDIM2   = '(10)    '
    END

Note TDIM2.  fitsverify does not complain about anything with this file.  If ds9 requires '(1, 10420)', that's a bug in ds9.  The FITS document describes TDIMn, and the string does not include the number of rows.

jemorrison:
Well I can not even get TDIM2 = '10420' to be written to the header. So If I can get over that hurdle then I can get back to what the value should be. 
Our datamodels does not write it out even when I force it to be a certain value. I think I have to wait until Nadia is back to figure this one out.

jemorrison:
Phil if you read the page 760 of the "Representation of spectral coordinates in FITS"
that discusses the form of TDIMn: " The dimension of the coordinate array will be given in the FITS table header by the keyword TDIMn set to'(M,K1,K2,..KM'), where n is the column containing the coordinate array. Note in particular that if M=1 the coordinate array is formally two-dimensional though the first axis is degenerate. The repeat count in the TFORMn keywword is the product of M and ll the KM."  So I think it is acceptable the TDIM2 = '(1,10420)' that is the allowed and it is the actual shape of the data.

Issue #1215, opened 2017-09-15
silence asdf warnings


nden:
ASDF introduced two new flags with defaults:
`ignore_version_mismatch=True, ignore_unrecognized_tag=False`. 
These flags should be added to calls to `AsdfFile` to provide more flexible control over warnings coming from, e.g., reference files.

bernie-simon:
@nden has a good point. I will also need to modify from_asdf to take the new aguments. After that works, the change can be merged.

PR #2600, opened 2018-09-28
Support asdf flags when creating a new datamodel [#1215]

bernie-simon:
Asdf supports two flags, ignore_version_mismatch and ignore_unrecognized_flag, when creating a new asdf file. Part of creating a new datamodel is opening an asdf file, either natively or as embedded in a FITS extension. This update passes these two flags from the DataModel argument list to the call that opens or creates an asdf file. The result is better control of the warning messages produced by asdf.

### DIFF
--- a/jwst/datamodels/fits_support.py
+++ b/jwst/datamodels/fits_support.py
@@ -531,9 +531,10 @@ def _load_history(hdulist, tree):
         history['entries'].append(HistoryEntry({'description': entry}))
 
 
-def from_fits(hdulist, schema, extensions, context):
+def from_fits(hdulist, schema, extensions, context, **kwargs):
     try:
-        ff = fits_embed.AsdfInFits.open(hdulist, extensions=extensions)
+        ##ff = fits_embed.AsdfInFits.open(hdulist, extensions=extensions, **kwargs)
+        ff = from_fits_asdf(hdulist, extensions=extensions, **kwargs)
     except Exception as exc:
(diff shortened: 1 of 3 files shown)


jwst/datamodels/fits_support.py
```diff
@@ -544,6 +544,17 @@ def from_fits(hdulist, schema, extensions, context):
 
     return ff
 
+def from_fits_asdf(hdulist, extensions=None,
+                  ignore_version_mismatch=True,
+                  ignore_unrecognized_tag=False,
+                  **kwargs):
+    kwargs = None
```
nden: flake8 caught this is not used.

jwst/datamodels/model_base.py
```diff
@@ -481,6 +488,28 @@ def save(self, path, dir_path=None, *args, **kwargs):
...
+    @staticmethod
+    def open_asdf(init=None, extensions=None,
+                  ignore_version_mismatch=True,
+                  ignore_unrecognized_tag=False,
+                  **kwargs):
+        """
+        Open an asdf object from a filename or create a new asdf object
+        """
+        # Drop arguments not explicitly mentioned in the arg list
+        kwargs = None
```
nden: This assignment is not used.

nden:
It's great to see this done.
Should the classmethod `from_asdf` be updated to use the flags?

nden:
@bernie-simon Could you rebase this.

PR #2629, opened 2018-09-28
Use frames instead of str in GWCS pipeline and code reformat WCS validation code

mcara:
This PR addresses two minor issues:

1. Cleans up the code from the https://github.com/spacetelescope/jwst/pull/2590 fix

2. Defines the frame for the V2V3 correction transformation using the same frame as used in the `WCS` pipeline for the `'v2v3'` transformation (but renames it). This way the pipeline will use frame objects instead of string names and "re-using" `'v2v3'` frame will preserve the same frame type as it used to be in the original (uncorrected/not "tweaked") WCS pipeline.

### DIFF
--- a/jwst/tweakreg/wcsimage.py
+++ b/jwst/tweakreg/wcsimage.py
@@ -183,7 +183,9 @@ class ImageWCS():
             pipeline = deepcopy(self._wcs.pipeline)
             pf, pt = pipeline[idx_v2v3]
             pipeline[idx_v2v3] = (pf, deepcopy(self._tpcorr))
-            pipeline.insert(idx_v2v3 + 1, ('v2v3corr', pt))
+            frm_v2v3corr = deepcopy(pf)
+            frm_v2v3corr.name = 'v2v3corr'
+            pipeline.insert(idx_v2v3 + 1, (frm_v2v3corr, pt))
             self._wcs = gwcs.WCS(pipeline, name=self._owcs.name)
             self._v23name = 'v2v3corr'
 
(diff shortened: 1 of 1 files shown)


jwst/tweakreg/wcsimage.py
```diff
@@ -209,51 +211,41 @@ def set_correction(self, matrix=[[1, 0], [0, 1]], shift=[0, 0]):
...
-            valid = False
-            return valid, message
+            return False, "There are fewer than 3 frames in the WCS pipeline."
 
         if frms.count(frms[0]) > 1 or frms.count(frms[-1]) > 1:
-            valid = False
-            message = "One or more frames in the WCS pipeline are not unique."
-            return valid, message
+            return (False, "First and last frames in the WCS pipeline must "
+                    "be unique.")
```
nden: BTW, all frame names in the pipeline **must** be unique.

mcara: @nden Is this guaranteed? If yes, then I can remove this `IF`. However, when using the standard example from the `gWCS` docs modified to have two identical frames, it does not raise an exception: `wcsobj = wcs.WCS([('det', det2sky), ('det', det2sky), ('sky', None)])` (but the code does crash when attempting to evaluate forward transform.

nden:
@mcara At some point it was. I need to check if this is still the case and fix it if not. I'll get back to you.

hbushouse:
Is this ready to merge? Do we want or need it in 0.12.0 or B7.2?

mcara:
@hbushouse I think it is ready. The only issue that could be "improved" would be that if https://github.com/spacetelescope/gwcs/issues/178 were implemented then [this check](https://github.com/spacetelescope/jwst/pull/2629/files#diff-3549795b42f783f2a253e58d71ee8f3aR223) could be removed. However, its presence does not hurt the code in any way (well, except from performance point of view).

PR #2644, opened 2018-10-02
Make finding dispersion axis robust in resample

### LINKED ISSUE #2499: resample_spec problems/questions for NIRSpec MOS
Issue [JP-369](https://jira.stsci.edu/browse/JP-369) was created by James Muzerolle:

When processing NIRSpec MOS data, the products from the resample_spec step have a few peculiarities.  Some of the individual 2d spectra end up with all zeroes in the s2d output, while others show very blocky resampling with significant gaps.  The attached figure shows an example of both cases from the same exposure, which is a simulated MOS scene.  For the latter case, the problem might be fixed by using a finer grid of output points in the cross-dispersion direction (all spectra in this particular example only have 4 or 5 output pixels in that direction).  Is there a way to specify a different sampling than the default?

 

This exposure was processed using pipeline version 0.10.1a.dev79+gaf19b1ba.  The input file, calspec2 products, and MSA metafile can be found in /grp/jwst/wit4/nirspec/Muzerolle/sim_example/.

jdavies-st:
The dispersion axis is now being found correctly when the spectrum trace is short relative to the size of the cross-dispersion.  An improved algorithm is borrowed from @philhodge's `extract_1d` code.

Resolves #2499.


jwst/resample/resample_spec.py
```diff
@@ -275,3 +271,17 @@ def do_drizzle(self, **pars):
...
+
+def find_dispersion_axis(wavelength_array):
+    """
+    Find the dispersion axis (0-indexed) of the given 2D wavelength array
+    """
+    diffx = wavelength_array[:, 1:] - wavelength_array[:, 0:-1]
+    diffy = wavelength_array[1:, :] - wavelength_array[0:-1, :]
+    dwlx = np.abs(np.nanmean(diffx))
+    dwly = np.abs(np.nanmean(diffy))
+    if dwlx > dwly:
```
philhodge: If `dwlx` is equal to `dwly`, this function will return None.  That's probably OK, because it "shouldn't happen," and it would be caught when attempting to evaluate `spectral_axis ^ 1`; the drawback is that the error message would be somewhat obscure.

jdavies-st: Thanks.  Yeah, I was trying to figure out how to deal with that situation.  It would be an amazing confluence of events if that happened!  Perhaps I should catch that case and issue an error message.  `raise LotteryErrror('You win!')`

jdavies-st: Anyway, I'll add a `RunTimeError`.  😄

PR #2646, opened 2018-10-02
Fix deprecation warnings in tests and other cleanup

jdavies-st:
This fixes the following warnings:
- Test fixtures being called as functions is deprecated.
```
lib/python3.7/site-packages/_pytest/fixtures.py:795: RemovedInPytest4Warning: Fixture "full_pool_rules" called directly. Fixtures are not meant to be called directly, are created automatically when test functions request them as parameters. See https://docs.pytest.org/en/latest/fixture.html for more information.
```
- The regex below should use a raw `r'string'` string instead for python 3.  But we don't actually use the `@require_crds_context` marker in any of our tests.  And if we do, we should just the function in `ci_watson` for this, which is identical and doesn't suffer from the DeprecationWarning above.  So this is just removed completely.
```
jwst/jwst/tests/helpers.py:51: DeprecationWarning: invalid escape sequence \d
  match = re.match('jwst_(\d\d\d\d)\.pmap', current_context_string)
```
- Our `configobj`, which we copy from `astropy.extern` to `jwst.extern` has the deprecation warning below.  This is updated to match what is in `astropy.extern` now.
```
jwst/jwst/extern/configobj/configobj.py:597: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  elif isinstance(value, collections.Mapping) and not unrepr:
```

Also, remove the `@not_under_travis` test decorator, as it is redundant with `@runslow`.  And we don't use it.

And finally, move `confest.py` down a level in the package:
```
jwst/confest.py -> jwst/jwst/conftest.py
```
so that it gets packaged in the installation.


jwst/tests/helpers.py
```diff
@@ -14,50 +10,21 @@ def abspath(filepath):
 
 
 # Decorator to indicate slow test
+#  - should use @pytest.mark.slow from ci_watson plugin instead
```
stscieisenhamer: Why not remove uses of this?

jwst/tests/helpers.py
```diff
@@ -14,50 +10,21 @@ def abspath(filepath):
...
 runslow = pytest.mark.skipif(
     not pytest.config.getoption("--slow"),
     reason="need --slow option to run"
 )
 
 
 # Decorator to indicate TEST_BIGDATA required
+#  - should use @pytest.mark.bigdata from ci_watson plugin instead
```
stscieisenhamer: Why not remove uses of this?

jwst/tests/helpers.py
jdavies-st: A can do that!  Another commit coming.

PR #2648, opened 2018-10-02
Truncate long schema validation error message

bernie-simon:
Datamodels can produce very long schema validation warnings or errors. This update truncates them to 2000 characters if they are longer, which is the size of an old fashioned terminal screen (80x25).


bernie-simon:
Too much stuff in pull request. will have to fix.

PR #2657, opened 2018-10-03
Truncate long schema validation errors

bernie-simon:
This update truncates schema validation warnings or errors at 2000 characters, which is the size of a terminal screen (80x25). Schema validation can create very large eror messages and these obscure the real problems with regression test runs.


CHANGES.rst
```diff
@@ -60,6 +60,8 @@ datamodels
 
 - Initialize arrays and tables from function args in model_base [#2502]
 
+- Truncate long schema validation errors to 2000 characters
```
hbushouse: Include the PR number

PR #2669, opened 2018-10-05
Update MIRI detector1 step order

### LINKED ISSUE #1454: calwebb_detector1 re-ordering for MIRI
The current ordering for callwebb_detector1 for MIRI is not in agreement with the agreed ordering via the JCalWG confluence pages.
The correct ordering of the MIRI specific portion should be:
dq_init
saturation
ipc
first-frame
last-frame
llinearity
rscd
dark_current
refpix

hbushouse:
Updated the order of MIRI steps in level-2a processing to be consistent with JCCWG specs, which boils down to moving `firstframe` and `lastframe` to be ahead of `linearity` and `rscd`. Also updated the `calwebb_dark` pipeline to include the `firstframe` step for MIRI.

Fixes #1454.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -173,6 +173,8 @@ pipeline
 
 - Added new Image2Pipeline configuration calwebb_wfs-image2.cfg for WFS&C processing [#2599]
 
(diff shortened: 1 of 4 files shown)


stscicrawford:
LGTM -- thanks for remember the change log!

Issue #2674, opened 2018-10-08
Remove multi-product capability of Image2Pipeline and Spec2Pipeline


stscieisenhamer:
# Issue
Remove multi-product capability of Image2Pipeline and Spec2Pipeline. Also ensure that the pipelines return the "final" results of the pipeline programmatically.

# Resources
- [JP-440](https://jira.stsci.edu/browse/JP-440)
- #1096
- #1694

stscieisenhamer:
# Implementation

- [ ] documentation
- [ ] tests
- [ ] code
- [ ] tests
- [ ] documentation
- [ ] changelog

stscieisenhamer:
Closing this issue in favor of #1096 and #1694.

PR #2676, opened 2018-10-08
Spec2Pipeline and Image2Pipeline returns results

stscieisenhamer:
A regression was reported, [JP-440](https://jira.stsci.edu/browse/JP-440), where `Image2Pipeline` no longer returned its primary, or `cal` result, when called programmatically.

This occurred when `Image2Pipeline`, along with `Spec2Pipeline`, was modified to handle associations with multiple products; returning results was removed.

Initial consideration was to remove multi-product handling, #2674. However, this would cause other inconsistencies in interface, usage, etc. In particular, multiple product handling is in use and works as expected when all results are written to files.

So, this PR takes the more direct tack: Before, `Image2Pipeline`, for an input, returned an output. Hence, now that `Image2Pipeline` handles multiple inputs, it should return a list of results.

`Spec2Pipeline` never returned anything. However, it has been modified to mimic `Image2Pipeline` behavior.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -181,6 +181,9 @@ pipeline
 
 - Updated the order of MIRI steps in calwebb_detector1 and calwebb_dark. [#2669]
 
+- Update Image2Pipeline and Spec2Pipeline to properly return results. [#2676]
+
+
 ramp_fitting
 ------------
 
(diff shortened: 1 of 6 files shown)


stscieisenhamer:
Note: PR created to have the unit tests at least run and to get reviewers a chance to dwell on the tack taken. Approval of the solution and updating of documentation need to occur before merging.

stscicrawford:
LGTM, but it does need the Change log updated appropriately

PR #2678, opened 2018-10-08
Remove inheritance from object

stscirij:
References #1392


nden:
I'm not sure we need to do the changes in pyparsing since it's a third party package.

PR #2688, opened 2018-10-10
Add missing reference file keywords to reffile schemas

### LINKED ISSUE #2684: Reference file schemas need the P_* keyword schemas added
- [x] `SuperBiasModel` needs keyword_preadpatt
- [x] `NirspecFlatModel` needs keyword_pexptype

Reported by @gkanarek of the NIRSpec team.  These keywords are required for CRDS validation, but are not in the schemas.  Since they are not in the schemas, if the reference file models above are used to generate them and they are populated, they don't get written out to the FITS file.

The workaround was to manually add the values to the FITS file afterwards.  But if they are required, they should be in the schema.

jdavies-st:
- Adds P_EXP_TY to NIRSpec Flat reffile
- Adds P_READPA to SuperBias reffile
- cleans up the `self.validate()` method so the warning is useful and the error message is not emitted twice.

Resolves #2684.

### DIFF
--- a/jwst/datamodels/reference.py
+++ b/jwst/datamodels/reference.py
@@ -1,4 +1,7 @@
 import warnings
+import sys
+import traceback
+
 from .model_base import DataModel
 from .dynamicdq import dynamic_mask
 from .validate import ValidationWarning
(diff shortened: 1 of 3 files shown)


jdavies-st:
I've added the extra keywords.

Agree to leave the larger issue of updating the rest of the schemas to a separate issue and PR(s).

stscicrawford:
Does this need to go into the patch?   If so, please add the `0.12.1 patch` label.   It would probably be good to close the related issue once this is merged and open a bigger issue to revisit in 7.3.

jdavies-st:
I don't think this needs to go into `0.12.1`, as this code is only used by the instrument teams in creating reffiles.  They tend to use the latest dev of `jwst` anyway.

Will definitely open a larger issue on related discussion in the issue for 7.3.

nden:
Re: validating the required ref file keywords - we have complaints that the code checks each keyword individually and has to be run several times before all keywords are cleared. Although the usual mode of operation is probably running this from scripts, it may still be good to check all required keywords and report once on the their status.

nden:
Agreed. I thought about it now reading this PR. I'll open a separate issue for this.

bernie-simon:
"we have complaints that the code checks each keyword individually and has to be run several times before all keywords are cleared"

I looked at doing this, but it's really difficult given the way the code is structured.

PR #2689, opened 2018-10-10
Update saturation and background docs

### LINKED ISSUE #2066: Update saturation step docs
INS testers have pointed out in https://jira.stsci.edu/browse/JP-144 that the saturation step docs do not explicitly mention the fact that all groups beyond the point at which saturation sets in are flagged as saturated in the GROUPDQ array. Right now the wording could be construed to indicate that only the group in which saturation occurs is flagged.

hbushouse:
Updated saturation step docs to more clearly explain which and how pixels are flagged in certain groups or in the pixeldq array. Made a few minor changes to the background step docs in the process, in order to maintain consistency (was using background as a template).

Fixes #2066.


stscicrawford:
LGTM -- and thanks for the update of the reference files as well.

nden:
@hbushouse Does this need to go in the next patch delivery?

hbushouse:
Yes, we should include any and all doc updates in the next patch.

Issue #2675, opened 2018-10-08
Reorganize reference file documentation in JWST pipeline


stscicrawford:
There are several different links for documentation in different places.  Here are some things that should be improved:

- [x] Linking together the high level pages
- [x] Removing the old ref_file_doc_DMS_65
- [x] look at ways to handle the linkage between steps and reference files
- [x] Assure that all reference file pages have the same format using including the standard keywords, format for the files, and include selection criteria

sosey:
related: #2536

jaytmiller:
CRDS probably has this in it's repro pseudo reference,  which may be a start:

```
steps_to_reftypes:
  assign_wcs: [camera, collimator, disperser, distortion, filteroffset, fore, fpa,
    ifufore, ifupost, ifuslicer, msa, ote, regions, specwcs, wavelengthrange]
  barshadow: [barshadow]
  bkg_subtract: [wavelengthrange, wfssbkg]
  cube_build: [cubepar, resol]
  dark_current: [dark]
  dq_init: [mask]
  extract_1d: [extract1d]
  extract_2d: [wavecorr, wavelengthrange]
  firstframe: []
  flat_field: [dflat, fflat, flat, sflat]
  fringe: [fringe]
  gain_scale: [gain]
  group_scale: []
  guider_cds: []
  imprint_subtract: []
  ipc: [ipc]
  jump: [gain, readnoise]
  lastframe: []
  linearity: [linearity]
  msa_flagging: [msaoper]
  pathloss: [pathloss]
  persistence: [persat, trapdensity, trappars]
  photom: [area, photom]
  ramp_fit: [gain, readnoise]
  refpix: [refpix]
  resample: [drizpars]
  resample_spec: [drizpars]
  rscd: [rscd]
  saturation: [saturation]
  srctype: []
  straylight: [straymask]
  superbias: [superbias]

```

hbushouse:
The old ref_file_doc_DMS_653 pages were removed in #2690.

hbushouse:
As of #2882 the docs for the reference files for each step have been updated to use a common format, include links to standard keywords and selection keywords, etc.

What remains to be done is to now create a high-level reference file "map" that includes pointers from steps to reffiles, and from reffiles to steps, using the now common format docs for each reffile.

hbushouse:
#2936 contains updates that provide hyper links in the general reference file doc page to all individual reffile docs, including tables that give both REFTYPE -> step and step -> REFTYPE linkages.

PR #2690, opened 2018-10-11
Reference file doc updates

hbushouse:
First round of updates for #2675. Removed the old ref_file_doc_DMS_653 (and subsidiary files) and began updates to individual step docs to unify their descriptions of reference files. Added PSFMASK docs to coron.

There will be lots more to come.


stscicrawford:
I'd suggest doing multiple PR's for the updates that will make it easier to review and merge, but I haven't merged this just in case you would prefer to do it as one big PR.

hbushouse:
@stscicrawford Now that I've completed my in-process updates I'll do separate PR's from now on.

hbushouse:
@dmggh Please verify AMI updates.
@rij Please verify barshadow updates.
@jemorrison Please verify cube_build ref file updates.

hbushouse:
@jemorrison While making the updates to the `cube_build` ref files, I noticed that there's a discrepancy between the names of the columns in the actual MIRI RESOL reference file (jwst_miri_resol_0003.fits) and what's defined in miri_resolution.schema.yaml for the RESOLVING_POWER table. In particular, the data model schema has:
```
       - name: band
       - name: center
       - name: a_low
       - name: b_low
       - name: c_low
       - name: a_high
       - name: b_high
       - name: c_high
       - name: a_ave
       - name: b_ave
       - name: c_ave
```
while the actual reference file has column names:
```
SUB_BAND
R_CENTRE
R_A_LOW
R_B_LOW
R_C_LOW
R_A_HIGH
R_B_HIGH
R_C_HIGH
R_A_AVG
R_B_AVG
R_C_AVG
```
How is the ref file able to be loaded into the data model without throwing errors?

jemorrison:
the code cube_build_io_util.read_resolution_file reads in the MIRI resolution file. Maybe I am not doing something standard with the datamodels reading reference file  but I search on the column names defined in the reference file and fill in the parameters I need 
for example
ptab = datamodels.MiriResolutionModel(resol_filename)
table_alpha_cutoff = ptab.psf_fwhm_alpha_table['A_CUTOFF']

I have just found a error in the code when using miripsf - but it is related to the changes I made to support  non-linear wavelength dimension and I will fix it now with a different PR

hbushouse:
@jemorrison You're right that when loading jwst_miri_resol_0003.fits into a MiriResolutionModel, the resolving_power_table attributes match what's actually in the ref file, i.e. names like SUB_BAND, R_CENTRE, and R_A_LOW. But those names are different than what's specified in the data model schema for that table. So I'm just really surprised that it works. 

@bernie-simon Do the table column names in a file override the names specified in the schema?

Either way, I'd feel better if the schema were updated to match the actual reference file, for consistency and to prevent questions like this thread.

PR #2692, opened 2018-10-11
Update CI to test against unmerged ASDF branch

drdavella:
**DO NOT MERGE**

The only purpose of this PR is to test the pipeline against a proposed change to ASDF.  It updates the CI scripts to install ASDF from a topic branch on GitHub.

The proposed change is that arrays below a certain threshold size (50 elements, to be specific) will automatically be inlined rather than stored in binary blocks. While the threshold can be customized, the change makes this new behavior the default. See https://github.com/spacetelescope/asdf/pull/557 for more details.

I'm testing this here since it has the potential to cause comparisons to old reference files to fail. We'll see how the test goes and move forward from there.

Also, feel free to voice and questions/concerns here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,7 @@ addons:
 env:
     global:
         - CONDA_CHANNELS='http://ssb.stsci.edu/astroconda-dev'
-        - CONDA_DEPENDENCIES="asdf
-                              astropy
+        - CONDA_DEPENDENCIES="astropy
                               crds
                               dask
                               drizzle
@@ -42,7 +41,7 @@ env:
         - CRDS_PATH='/tmp/crds_cache'
         - MAIN_CMD='python setup.py'
         - MC_URL=https://repo.continuum.io/miniconda
-        - PIP_DEPENDENCIES=''
+        - PIP_DEPENDENCIES='git+git://github.com/drdavella/asdf@inline-threshold'
         - PYTHON_VERSION=3.6
 
     matrix:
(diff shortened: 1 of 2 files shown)


jdavies-st:
The regression tests are run in `JenkinsfileRT` that are not run in `Jenkinsfile`, and these exercise the reference files a bit more.  If you make the same changes to `JenkinsfileRT`, we can kick of a custom Jenkins instance to run against it.

nden:
It's good to have some performance comparisons. In the past I've tested this with different sizes written inline and decided the performance hit was too much so I sacrificed the readability. A good test case would be a nirspec observation or its WCS object only because the WCS includes many polynomial models and the coefficients are written out as arrays.

nden:
I'm mostly concerned with reading/writing files. So I did a test - writing and reading the WCS of one nirspec slit. 

With the proposed PR:

timeit f.write_to('s0.asdf')                                                                                            
173 ms ± 498 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

timeit AsdfFile.open('s0.asdf')                                                                                         
705 ms ± 2.12 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

With master:

timeit f.write_to('s0_master.asdf')                                                                                      
107 ms ± 747 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

timeit AsdfFile.open('s0_master.asdf')                                                                                   
637 ms ± 2.57 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

nden:
I'd say we probably don't want to use this in the pipeline and should set the threshold to `0` (which would be equivalent to the old behavior, right?

drdavella:
Yes, that would be equivalent. If that's the case, I can just close this PR and then open a new one here once the one in ASDF is merged.

nden:
Sounds good. As long as the `threshold` argument is available in the datamodels we can use it for smaller files.

PR #2693, opened 2018-10-11
Update ifucube schema to correctly write tdim keyword

bernie-simon:
The value of the shape field under the datatype field of the schema must be the reverse of the value of the tdim keyword that is needed in the header. This update does that and deletes the unnecessary wavedim keyword from the wcsinfo schema.


stscicrawford:
Does this need to go in as a patch?

jemorrison:
I don't think this works. Did you try and run it on some data ?  I made the two changes you did -
removing the wavedim from wcsinfo and reversing the order of the shape of wavetable to
[10420,1] and I get an error that in ifu_cube.py line 1279 that - could not broadcast input arrays
from shape (1,10420) into shape (10420,1)

If we remove wavedim from the wcsinfino then we have to take out in ifu_cube.py setting
this value ifucube_model.meta.wcsinfo.wavedim

philhodge:
I didn't just delete wavedim from wcsinfo, I moved it to ifucube.schema.yaml right after the `wavetable` entry.

jemorrison:
Is that in this PR I am not seeing it

hbushouse:
If this actually helps to solve the problem, yes it would be good to include in a patch. But this has to be tested from within cube_build itself to verify that everything comes out right. I'm pretty sure I tried this approach myself when playing around with the wavetab a few days ago and flipping the dimensions ends up with the array being written incorrectly in the FITS table extension (at least it did in my experiment).

stscicrawford:
Closing this in favor of #2719

PR #2702, opened 2018-10-15
Set default inline threshold for ASDF to 0

drdavella:
ASDF just introduced new functionality that causes arrays below a particular threshold size to be automatically stored inline rather than in internal blocks (see https://github.com/spacetelescope/asdf/pull/557).

This has potential performance issues for the pipeline, and it also may affect comparisons with existing reference files. This PR updates the pipeline to make use of the new `inline_threshold` parameter introduced by ASDF to ensure that all arrays continue to be stored in internal blocks for now.

**EDIT**: some initial discussion of this took place in #2692.

### DIFF
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -67,6 +67,7 @@ class ModelContainer(model_base.DataModel):
 
     def __init__(self, init=None, persist=True, **kwargs):
 
+        inline_threshold = kwargs.pop('inline_threshold', 0)
         super(ModelContainer, self).__init__(init=None, **kwargs)
         self._persist = persist
 
@@ -79,7 +80,8 @@ class ModelContainer(model_base.DataModel):
             instance = copy.deepcopy(init._instance)
             self._schema = init._schema
             self._shape = init._shape
-            self._asdf = AsdfFile(instance, extensions=self._extensions)
+            self._asdf = AsdfFile(instance, extensions=self._extensions,
+                                  inline_threshold=inline_threshold)
             self._instance = instance
             self._ctx = self
             self.__class__ = init.__class__
@@ -155,7 +157,7 @@ class ModelContainer(model_base.DataModel):
         self._open_model(index)
         return self._models.pop(index)
 
-    def copy(self, memo=None):
+    def copy(self, memo=None, inline_threshold=0):
         """
         Returns a deep copy of the models in this model container.
         """
@@ -164,7 +166,8 @@ class ModelContainer(model_base.DataModel):
                                 pass_invalid_values=self._pass_invalid_values,
                                 strict_validation=self._strict_validation)
         instance = copy.deepcopy(self._instance, memo=memo)
-        result._asdf = AsdfFile(instance, extensions=self._extensions)
+        result._asdf = AsdfFile(instance, extensions=self._extensions,
+                                inline_threshold=inline_threshold)
         result._instance = instance
         result._iscopy = self._iscopy
         result._schema = result._schema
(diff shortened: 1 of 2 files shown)


jdavies-st:
It looks like there's something funky going on in `ModelContainer`.  Can have a look.

nden:
The asdf PR was just merged and we are not testing against the asdf repo but against conda-dev.
@drdavella Can you temporarily modify  .travis to point to the asdf repo for the test?
If it passes I suggest we merge this in order to have consistent code in conda-dev tomorrow.

jdavies-st:
Ah, I see.  Yes, also modify `Jenkinsfile` in addition to `.travis.yaml`.  You'll have to `pip install git+...` for asdf in order for the tests to run  I believe.

drdavella:
Just to clarify and add some thoughts from offline discussion with @stscicrawford: this change is completely compatible with the standard itself. No changes were made to the standard, and files created both before and after this change are readable by any ASDF implementation that conforms to the standard.

The difference is an implementation-specific change that decides the default behavior for whether a data array should be stored as an internal block or as an inline array. Again, both cases are compliant with the standard as-is. However, prior to this change, **all** arrays were stored in internal blocks by default, whereas with this change, numerical arrays with 50 or fewer items will be stored as inline arrays by default. Even prior to this change, users were free to override the storage type of any data block, and that remains true now. The initial suggestion for this can be found in https://github.com/spacetelescope/asdf/issues/556.

As @nden mentioned, it is probably best for the pipeline to stick with the original behavior for performance reasons. Also, while the data contents of the file are exactly the same in either case, a raw diff of files containing data before and after this change will fail. However, raw diffs of ASDF files can be problematic for other reasons as well.

Currently the burden is on users to set the default threshold to 0 if they want to preserve the previous behavior. As @stscicrawford pointed out, this change is backwards-incompatible for users who expect the old behavior, even if files in both cases remain compliant with the standard itself. So it's possible that ASDF will need to set the default threshold to 0 until the next major release in order to follow the principle of least surprise.

cc @perrygreenfield

jdavies-st:
A note, we currently don't do any raw diffs of ASDF files here in JWST-land.  Raw asdf files are only used for reference files.  All our diffs are of the `AsdfInFits` FITS files, and this change doesn't change those if I understand correctly.

drdavella:
@jdavies-st since the contents of the ASDF file are placed inside the FITS file, diffing the FITS files themselves will still turn up any differences that result from this change.

However, your comment made me realize that there might be a nicer solution to this, at least for the JWST pipeline case, which is to simply make this change not apply to ASDF-in-FITS files. In that case, this PR could be closed.

drdavella:
If https://github.com/spacetelescope/asdf/pull/561 makes sense and seems reasonable, this PR can be closed.

nden:
I really think this is only a jwst issue and in fact only for some instruments/modes. For those modes we don't want to write inline arrays for pure asdf files either because the WCS reference files are pure asdf files. So I think we should go and merge this PR. I am also against spacetelescope/asdf#516 as i think it's an unnecessary restriction.

stscicrawford:
I think that does appear reasonable.  Are you still going to set the default to zero?   Otherwise, this will still need to be updated in the pipeline for performance issues.

perrygreenfield:
Is there any specific information on performance impact? I'm curious about how small arrays would have a large impact.

nden:
@perrygreenfield 

https://github.com/spacetelescope/jwst/pull/2692#issuecomment-429046366

nden:
@drdavella It'd be useful if you can modify `.travis.yaml` to point to the asdf repo for this PR.

drdavella:
@nden I am having no luck with this. I think even though it's being installed from `pip` first, it gets overridden by the version from astroconda. Can we wait to rerun this tomorrow?

nden:
Yeah, let's wait. If you get a chance please restore the original `.travis`. Thanks for all the efforts!

drdavella:
This is still not picking up the right version of ASDF 🤦‍♂️

PR #2706, opened 2018-10-16
Add filtering/selection by peakmax and brightest

mcara:
This PR adds two new parameters to `tweakreg_step` star object detection algorithm to allow source selection/filtering by:

1. `peakmax` - maximum peak pixel value for selecting stars.
2. `brightest` - number of brightest objects to keep.

The rationale for these parameters is that `peakmax` could allow automatic exclusion of sources that reach a known saturation level of the detector and `brightest` would keep top `brightest` stars thus eliminating the faintest objects that could be false detections.

CC: @hbushouse @larrybradley

### DIFF
--- a/jwst/tweakreg/tweakreg_catalog.py
+++ b/jwst/tweakreg/tweakreg_catalog.py
@@ -4,7 +4,8 @@ from ..datamodels import ImageModel
 
 
 def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
-                          sharphi=1.0, roundlo=-1.0, roundhi=1.0):
+                          sharphi=1.0, roundlo=-1.0, roundhi=1.0,
+                          brightest=None):
     """
     Create a catalog of point-line sources to be used for image
     alignment in tweakreg.
@@ -37,6 +38,21 @@ def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
     roundhi : float, optional
         The upper bound on roundess for object detection.
 
+    brightest : int, None, optional
+        Number of brightest objects to keep after sorting the full object list.
+        If ``brightest`` is set to `None`, all objects will be selected.
+
+    peakmax : float, None, optional
+        Maximum peak pixel value in an object. Only objects whose peak pixel
+        values are *strictly smaller* than ``peakmax`` will be selected.
+        This may be used to exclude saturated sources. By default, when
+        ``peakmax`` is set to `None`, all objects will be selected.
+
+        .. warning::
+            `DAOStarFinder` automatically excludes objects whose peak
+            pixel values are negative. Therefore, setting ``peakmax`` to a
+            non-positive value would result in exclusion of all objects.
+
     Returns
     -------
     catalog : `~astropy.Table`
@@ -52,7 +68,8 @@ def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
 
     daofind = DAOStarFinder(fwhm=kernel_fwhm, threshold=threshold,
                             sharplo=sharplo, sharphi=sharphi, roundlo=roundlo,
-                            roundhi=roundhi)
+                            roundhi=roundhi, brightest=brightest,
+                            peakmax=peakmax)
     sources = daofind(model.data)
 
     columns = ['id', 'xcentroid', 'ycentroid', 'flux']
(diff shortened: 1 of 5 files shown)


jwst/tweakreg/tweakreg_catalog.py
```diff
@@ -4,7 +4,8 @@
 
 
 def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
-                          sharphi=1.0, roundlo=-1.0, roundhi=1.0):
+                          sharphi=1.0, roundlo=-1.0, roundhi=1.0,
+                          brightest=None):
```
hbushouse: Need to add peakmax to the argument list here, so that it's defined for use in line 72.

jwst/tweakreg/tweakreg_catalog.py
```diff
@@ -52,7 +68,8 @@ def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
 
     daofind = DAOStarFinder(fwhm=kernel_fwhm, threshold=threshold,
                             sharplo=sharplo, sharphi=sharphi, roundlo=roundlo,
-                            roundhi=roundhi)
+                            roundhi=roundhi, brightest=brightest,
+                            peakmax=peakmax)
```
hbushouse: The value "peakmax" is currently undefined at this point in the code

jwst/tweakreg/tweakreg_step.py
```diff
@@ -27,9 +27,11 @@ class TweakRegStep(Step):
...
         # Source finding parameters:
         save_catalogs = boolean(default=False) # Write out catalogs?
-        catalog_format = string(default='ecsv')   # Catalog output file format
-        kernel_fwhm = float(default=2.5)    # Gaussian kernel FWHM in pixels
-        snr_threshold = float(default=10.0)  # SNR threshold above the bkg
+        catalog_format = string(default='ecsv') # Catalog output file format
+        kernel_fwhm = float(default=2.5) # Gaussian kernel FWHM in pixels
+        snr_threshold = float(default=10.0) # SNR threshold above the bkg
+        brightest = int(default=None) # Keep top ``brightest`` objects
+        peakmax = float(default=None) # Filter out objects with pixel values >= ``pixmax``
```
hbushouse: "pixmax" should be "peakmax"

jwst/tweakreg/tweakreg_catalog.py
mcara: Thank you!

jwst/tweakreg/tweakreg_step.py
```diff
@@ -27,9 +27,11 @@ class TweakRegStep(Step):
...
         # Source finding parameters:
         save_catalogs = boolean(default=False) # Write out catalogs?
-        catalog_format = string(default='ecsv')   # Catalog output file format
-        kernel_fwhm = float(default=2.5)    # Gaussian kernel FWHM in pixels
-        snr_threshold = float(default=10.0)  # SNR threshold above the bkg
+        catalog_format = string(default='ecsv') # Catalog output file format
+        kernel_fwhm = float(default=2.5) # Gaussian kernel FWHM in pixels
+        snr_threshold = float(default=10.0) # SNR threshold above the bkg
+        brightest = int(default=None) # Keep top ``brightest`` objects
+        peakmax = float(default=None) # Filter out objects with pixel values >= ``peakmax``
```
hbushouse: The /pipeline/tweakreg.cfg file should also be updated to include these parameter changes (and to make sure they're consistent with one another).

mcara:
@hbushouse I think I have addressed the issues pointed out by you. Thanks!

I also want to ask if you would like to limit number of objects used for alignment by setting the default value of `brightest` to something like `50`, or `100`, etc. Or, will this be handled differently in the pipeline in an instrument-specific manner? I think even 50 objects should be enough for good image alignment (unless all these sources are clustered in a single corner of the image). All of this is to make object detection less susceptible to "fake objects" (i.e., hot pixels, random structure in the noise, etc.)

mcara:
@hbushouse I keep forgetting about `tweakreg.cfg`...

mcara:
@hbushouse Also, IMO, a minimum of 3 sources are necessary for image alignment using 'general' (shifts, rotations, scale, skew) fit. Should I lower `minobj` to 3? In fact we can lower it to 1 because I think the algorithm will automatically downgrade `fitgeom` to the appropriate level (i.e., from 'general' to 'rscale' to 'shifts').

mcara:
@nden Done with @larrybradley help

mcara:
@hbushouse  astropy/photutils#750
https://github.com/astropy/photutils/pull/750 was merged

hbushouse:
@stscieisenhamer @mcara Any ideas as to why lib/tests/test_suffix.py is throwing an error in the Travis tests? Seems to be a mismatch in "tweakreg" suffix. How would these updates be causing that?

stscieisenhamer:
Not exactly sure how the current changes affect this. I'm guessing there is a change upstream in the 0.13.0_new_features that actually did this.

Regardless, all that is needed is to remove `tweakreg` from the list of calculated suffixes in `lib/suffix.py`.

PR #2709, opened 2018-10-16
Update MIRI dark and flat EXP_TYPEs

### LINKED ISSUE #2694: Update EXP_TYPE allowed values for MIRI external flats
https://jira.stsci.edu/browse/JSDP-751 documents changes being made in SDP, such that they will be setting the EXP_TYPE keyword to values of "MIR_IMAGE-EXTFLAT" and "MIR_MRS-EXTFLAT" when external flats are obtained. These are new EXP_TYPE values and hence they need to be added to the list of allowed values in our keyword schemas.

hbushouse:
Updated allowed EXP_TYPE values in all applicable schema for the new MIRI dark and flat types. Fixes #2694.


hbushouse:
Latest commit adds the value "MIR_FLATALL" to the enum list, because SDP may not get their updates into B7.2 that will translate this value into "MIR_FLATIMAGE" and "MIR_FLATMRS" for individual exposures, so we may see the "MIR_FLATALL" value in level-1b products.

jaytmiller:
Does the MIR_FLATALL change require CRDS rmap support?

hbushouse:
@jaytmiller Yes, NIS_EXTCAL was missing from pexptype. Last commit added it. And yes, eventually we want to use includes to go from the individual "keyword_xxxx.schema.yaml" files to "core.schema.yaml" (there's been a ticket open for a while for that). I've been doing the p<keyword> schemas manually by just doing a copy-n-paste from the yaml and then editing the syntax to change commas to |, etc. It's a pain, but it is a way to make sure they're both the same.

hbushouse:
@jaytmiller Regarding "MIR_FLATALL", theoretically no CRDS rmaps should be needed, because flats don't go through level-2b processing and the only ref file types that currently depend on EXP_TYPE are confined to level-2b processing. So there won't be any need for CRDS coverage in regular Ops. And eventually we won't see any values of either "MIR_DARKALL" or "MIR_FLATALL", once SDP implements their updates to change those values to detector-specific values of "MIR_FLATIMAGE", "MIR_FLATMRS", etc. (we won't see any "ALL" values anymore after that change).

Issue #2681, opened 2018-10-09
revise all docstrings


nden:
As part of the 7.2 documentation update make sure code has docstrings, they are correct and they are displayed correctly. Docstrings must be present in all public classes and functions, and in complex private functions/methods. 

[An example module](https://gist.github.com/nden/d8a7617424ecf9d67aaf4f485d0f561f)

Some useful links for reference:

[Numpydoc docstring convention](https://numpydoc.readthedocs.io/en/latest/format.html)
[STScI's Python style guide](https://github.com/spacetelescope/style-guides/blob/master/guides/python.md)

To build the documentation locally:

`python setup.py build_sphinx`

I had to pip install `sphinx_automodapi`, all other packages install with conda.

stscicrawford:
List of subpackages
- [ ] ami
- [ ] assign_wcs
- [ ] associations
- [ ] background
- [ ] barshadow
- [ ] combine_1d
- [ ] coron
- [ ] csv_tools
- [ ] cube_build
- [ ] cube_skymatch
- [ ] dark_current
- [ ] datamodels
- [ ] dq_init
- [ ] emission
- [ ] engdblog
- [ ] exp_to_source
- [ ] extract_1d
- [ ] extract_2d
- [ ] firstframe
- [ ] fits_generator
- [ ] flatfield
- [ ] fringe
- [ ] gain_scale
- [ ] group_scale
- [ ] guider_cds
- [ ] imprint
- [ ] ipc
- [ ] jump
- [ ] jwpsf
- [ ] lastframe
- [ ] lib
- [ ] linearity
- [ ] model_blender
- [ ] mrs_imatch
- [ ] msaflagopen
- [ ] outlier_detection
- [ ] pathloss
- [ ] persistence
- [ ] photom
- [ ] pipeline
- [ ] ramp_fitting
- [ ] refpix
- [ ] resample
- [ ] reset
- [ ] rscd
- [ ] saturation
- [ ] skymatch
- [ ] source_catalog
- [ ] srctype
- [ ] stpipe
- [ ] straylight
- [ ] superbias
- [ ] tests
- [ ] timeconversion
- [ ] transforms
- [x] tso_photometry
- [ ] tweakreg
- [ ] tweakreg_catalog
- [ ] wfs_combine
- [ ] white_light
- [ ] wiimatch

philhodge:
If a function only returns None (and maybe not even explicitly), does there need to be a Returns section?

nden:
[...]

PR #2715, opened 2018-10-17
Added docstrings to dq_init step

stscirij:
See #2681

### DIFF
--- a/jwst/dq_init/dq_initialization.py
+++ b/jwst/dq_init/dq_initialization.py
@@ -20,7 +20,7 @@ def correct_model(input_model, mask_model):
     Parameters
     ----------
     input_model : input JWST datamodel
-        The jwst datamodel to the corrected
+        The jwst datamodel to be corrected
 
     mask_model : mask datamodel
         The mask model to use in the correction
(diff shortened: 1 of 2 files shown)


jwst/dq_init/dq_initialization.py
```diff
@@ -23,14 +15,43 @@
 
 
 def correct_model(input_model, mask_model):
-    """DQ Initialize a JWST Model"""
+    """Perform the dq_init step on a JWST datamodel
+
+    Parameters
+    ----------
+    input_model : input JWST datamodel
+        The jwst datamodel to the corrected
```
philhodge: This looks like a typo:  "to the corrected" probably should be "to be corrected".

jwst/dq_init/dq_initialization.py
```diff
@@ -23,14 +15,43 @@
...
 
 
 def do_dqinit(input_model, mask_model):
-    """Do the DQ initialization"""
+    """Perform the dq_init step on a JWST datamodel
+
+    Parameters
+    ----------
+    input_model : input JWST datamodel
+        The jwst datamodel to the corrected
```
philhodge: Same comment as above, "to the corrected".

PR #2719, opened 2018-10-17
Test wave

jemorrison:
Correctly writing TDIM2 keyword to WCS_TABLE header.
Fixed two errors when weighting-'miripsf.'
1. fixed  incorrectly filling in the wave resolution value
2. fixed writing ifucube.meta.ifu.weighting when weighting set my command line 

This replaces #2693 which had the order of the wavetable in the wrong order

### DIFF
--- a/jwst/cube_build/ifu_cube.py
+++ b/jwst/cube_build/ifu_cube.py
@@ -1263,7 +1263,6 @@ class IFUCubeData():
             # for now we need to pad wavelength to fix datamodel size
             num = len(wave)
             nn = 10420 # This number needs to match the shape of the
-            nn = 1782
             # wavetable['wavelength'] value in the ifucube.schema.
             # In the future it would be good to remove that we have to
             # set the size of this value in the schema.
@@ -1340,9 +1339,8 @@ class IFUCubeData():
             ifucube_model.meta.wcsinfo.crval3 = 1.0
             ifucube_model.meta.wcsinfo.crpix3 = 1.0
             ifucube_model.meta.wcsinfo.cdelt3 = None
-#            ifucube_model.wavedim = '(1,10420)'
-            ifucube_model.wavedim = '(1,1782)'
-            #print('wrote',ifucube_model.meta.wcsinfo.wavedim)
+            ifucube_model.wavedim = '(1,10420)'
+
 
         ifucube_model.meta.wcsinfo.ctype1 = 'RA---TAN'
         ifucube_model.meta.wcsinfo.ctype2 = 'DEC--TAN'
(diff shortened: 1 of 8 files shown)


jwst/cube_build/ifu_cube.py
```diff
@@ -1367,7 +1367,8 @@ def setup_ifucube(self, j):
         ifucube_model.meta.ifu.dq_extension = 'DQ'
         ifucube_model.meta.ifu.roi_spatial = self.rois
         ifucube_model.meta.ifu.roi_wave = self.roiw
```
jdavies-st: Since `rois` and `roiw` are both parameters that can be set from the `strun` command line, they also will need to be coerced to a string, i.e.:
```suggestion
        ifucube_model.meta.ifu.roi_spatial = str(self.rois)
        ifucube_model.meta.ifu.roi_wave = str(self.roiw)
```

jwst/cube_build/ifu_cube.py
```diff
@@ -1422,6 +1423,17 @@ def setup_ifucube(self, j):
...
+# remove any axis4 keywords left over from other steps. These keywords interfere with
+# DS9 in understanding the wavelength planes
+# This did not work - but I left it because maybe it is something similar
+#        ifucube_model.meta.wcsinfo.ctype4 = None
+#        ifucube_model.meta.wcsinfo.crval4 = None
+#        ifucube_model.meta.wcsinfo.crpix4 = None
+#        ifucube_model.meta.wcsinfo.cdelt4 = None
+#        ifucube_model.meta.wcsinfo.cunit4 = None
+
+# set WCS information
```
jdavies-st: You can use `del` here:
```
try:
    del ifucube_model.meta.wcsinfo.ctype4
except AttributeError:
    pass
```

jdavies-st: But actually setting to `None` should work.  Just tried this out.  Hmm.

philhodge: It's not an attribute, so you can't assign None to it, and you can't del it.  It's just a keyword in the file.

jdavies-st: Ah, I see.  None of these are in our schemas.  So they're actually all being stored in `ifucube_model._extra_fits`.  Delete them there if you want them gone, or delete all of `_extra_fits`.

jdavies-st: You can see what is stored in `extra_fits` via:
```
In [90]: model.extra_fits.instance                                                                                                                           
Out[90]: 
{'PRIMARY': {'header': [['FOO', 'bar', '']]},
 'SCI': {'header': [['FOO', 'bar', '']]}}
```
And delete stuff individually:
```
In [91]: model.extra_fits.SCI.header[0]                                                                                                                      
Out[90]: ['FOO', 'bar', '']

In [92]: del model.extra_fits.SCI.header[0]
```

jemorrison: del ifu_cubemodel.extra.fits.SCI,header[0] did not delete the AXIS4 keywords
I am not sure if this is correct place to delete these keywords. They should not be in
the *rate files and they are. I think they need to be removed in the CALDETECTOR1 pipeline.

PR #2720, opened 2018-10-17
Reffile docs3

hbushouse:
Updated reference file docs for the `dq_init` and `extract_2d` steps. In partial fulfillment of #2675.


hbushouse:
FYI: I deleted the "fgs_mask_format.rst" and "miri_mask_format.rst" files from the dq_init docs, because they weren't being used and aren't necessary (all the MASK formats are the same and are described in the updates docs).

hbushouse:
@nden Could you look over the extract_2d updates too, since you're an extract-2d ninja?

docs/jwst/extract_2d/reference_files.rst
```diff
@@ -1,26 +1,91 @@
...
+It can be turned off by specifying ``apply_wavecorr=False`` when running the step.
 
 .. include:: wavecorr_selection.rst
 
-WAVELENGTHRANGE Reference
--------------------------
+.. include:: ../includes/standard_keywords.rst
+
+Type Specific Keywords for WAVECORR
++++++++++++++++++++++++++++++++++++++
```
nden: This line looks longer than needed. Might mess up formatting.

docs/jwst/extract_2d/reference_files.rst
```diff
@@ -1,26 +1,91 @@
...
+
+=========  ==============================
+Keyword    Data Model Name
+=========  ==============================
+EXP_TYPE   model.meta.exposure.type
+=========  ==============================
+
+Reference File Format
++++++++++++++++++++++
+WAVELENGTHRANGE reference files are ASDF format, with the format and contents
```
nden: Should this be "in ASDF format"?

docs/jwst/extract_2d/reference_files.rst
```diff
@@ -1,26 +1,91 @@
...
+specified by the `WavelengthragneModel` data model schema.
```
nden: typo: Wavelengthra**gn**eModel

BTW, single back ticks will attempt to link the type. If this doesn't work it needs to be specified as `~jwst.datamodels.WavelengthrangeModel` so that it links correctly.

docs/jwst/extract_2d/main.rst
```diff
@@ -1,4 +1,3 @@
-
 Description
 ============
```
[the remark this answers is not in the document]
hbushouse: Section underlines can be any length longer than the text, just not shorter than the text.

docs/jwst/extract_2d/reference_files.rst
hbushouse: I don't think it hurts to have them longer than the text, but I can always fix to make sure.

docs/jwst/extract_2d/reference_files.rst
hbushouse: Ooh, I was wondering if it was somehow possible to include a link to each data model. Now I know how. Thanks!

Issue #2729, opened 2018-10-19
enable docstring testing


nden:
We are currently not testing the docstrings. Running `python setup.py test` lists the number of tests as
```
w/0 docstring testing : 1372
w/ docstring testing: 1438
```
There are quite a few doctests with examples and I can see the formatting of some of them is incorrect so it's good to run these tests.

When I enable docstring testing locally I get 3 errors which prevent the tests from running:

This one may be related to #2552
```
Error collecting jwst/associations/tests/data/rules_level3_coron.py ______________________
data/rules_level3_coron.py:23: in <module>
    AsnMixin_OpticalPath,
E   NameError: name 'AsnMixin_OpticalPath' is not defined

```
This looks like perhaps a file is missing. I'm not sure what the purpose of the `convert_to_uncal.py` script is and whether it's still needed but  t's brittle for automatic testing.

```
__ ERROR collecting jwst/lib/convert_to_uncal.py_________________________
../../lib/convert_to_uncal.py:9: in <module>
    input = datamodels.ImageModel(input_file)
../../datamodels/image.py:24: in __init__
    super(ImageModel, self).__init__(init=init, **kwargs)
../../datamodels/model_base.py:160: in __init__
    hdulist = fits.open(init)
/user/dencheva/anaconda3/envs/asdf-test-order/lib/python3.7/site-packages/astropy/io/fits/hdu/hdulist.py:151: in fitsopen
    lazy_load_hdus, **kwargs)
/user/dencheva/anaconda3/envs/asdf-test-order/lib/python3.7/site-packages/astropy/io/fits/hdu/hdulist.py:390: in fromfile
    lazy_load_hdus=lazy_load_hdus, **kwargs)
/user/dencheva/anaconda3/envs/asdf-test-order/lib/python3.7/site-packages/astropy/io/fits/hdu/hdulist.py:1039: in _readfrom
    fileobj = _File(fileobj, mode=mode, memmap=memmap, cache=cache)
/user/dencheva/anaconda3/envs/asdf-test-order/lib/python3.7/site-packages/astropy/utils/decorators.py:488: in wrapper
    return function(*args, **kwargs)
/user/dencheva/anaconda3/envs/asdf-test-order/lib/python3.7/site-packages/astropy/io/fits/file.py:177: in __init__
    self._open_filename(fileobj, mode, overwrite)
[...]

PR #2730, opened 2018-10-19
enable docstring testing

nden:
#2729


nden:
I think something else may be pulling it in as it seems to be running in this PR.
https://travis-ci.org/spacetelescope/jwst/jobs/443724499#L1133

jdavies-st:
Yes, a bit confusing!

nden:
@stscieisenhamer Could you look at these:

- Is this [failure](https://travis-ci.org/spacetelescope/jwst/jobs/443862465#L1563) expected? What do I need to do to fix it? Is setting the expected return value to `None` instead of `[]` correct? [here's the doctest](https://github.com/spacetelescope/jwst/blob/master/jwst/associations/lib/constraint.py#L181)

- [this docsrting](https://github.com/spacetelescope/jwst/blob/master/jwst/associations/lib/format_template.py#L103) raises [this error](https://travis-ci.org/spacetelescope/jwst/jobs/443862465#L1588). Shoul dthe example be changed and how?

@jdavies-st Can you look at this

- [failure](https://travis-ci.org/spacetelescope/jwst/jobs/443862465#L1588) : Is it because the file is missing or is it a datamodels issue?

@mcara How should these docstrings be fixed?

- [failure](https://travis-ci.org/spacetelescope/jwst/jobs/443862465#L1660), [doctest](https://github.com/spacetelescope/jwst/blob/master/jwst/resample/bitmask.py#L45)

- [failure](https://travis-ci.org/spacetelescope/jwst/jobs/443862465#L1704): [doctest](https://github.com/spacetelescope/jwst/blob/master/jwst/wiimatch/lsq_optimizer.py#L378)

- [failure](https://travis-ci.org/spacetelescope/jwst/jobs/443862465#L1718): [doctest](https://github.com/spacetelescope/jwst/blob/master/jwst/wiimatch/lsq_optimizer.py#L378) In this one the output is multiplied by -1.

stscieisenhamer:
Concerning the first error: I will need to check on that; it should work but the doctest is probably out-of-date.

However, the second error is because a file is attempting to be accessed.  Real issue is that this was never really meant to be a "doctest", but just an example.  I guess examples should not look like doctests? What is the standard here?

nden:
@stscieisenhamer Doctest can be disabled for individual examples with 
```
# doctest: +SKIP
```
But just for my understanding - is the `template` variable the file you are talking about? Further up `template` is defined as 
```
template = 'name="{name}" value="{value}"'
```
and this seems to work just fine for all preceding examples. Why is it failing for this one?
(Shows I know nothing about the associations code.)

mcara:
@nden `jwst.bitmask` should be removed. Almost all modules that depend on it have been switched to `stsci.tools.bitmask`. I'll make a PR.

nden:
@stscieisenhamer Thanks! I'll make the changes in this PR to minimize rebasing.

nden:
Not sure why but github is not picking up my last two commits. I'll close this one and open another PR in attempt to resolve this.

setup.cfg
```diff
@@ -23,8 +23,9 @@ exclude = jwst/extern,docs,jwst/associations,jwst/jwpsf,jwst/ramp_fitting, jwst/
 [tool:pytest]
 minversion = 3
 addopts = --ignore=build
-norecursedirs = .eggs build docs/_build relic
+norecursedirs = .eggs build docs/_build relic jwst/timeconversion jwst/lib jwst/extern jwst/pipeline/tests
```
jdavies-st: Is there a reason `jwst/pipeline/tests` is added here?  There's actually a bunch of tests in there that are run using `pytest`.

nden:
Ah, good catch. I had set it temporarily to get around an error and forgot to remove it. Something was failing, can't remember what was it. Will remove it shortly.

nden:
This is down to 3 errors. 2 of them look like real problems. @stscieisenhamer Can you take a look at the failures in `constraints` and `registry`.  

- I don't see `Registry.finalize`. Is the example wrong? Should it be removed/corrected?
- In the constraints example - `sources` is a list so it looks like the error is legit or the example is wrong.  Please advise.

The failure in blendmodels - the example is attempting to open an `asn` file. @bernie-simon Is datamodels.filecheck supposed to work on asn files? If so, perhaps this case should raise a `OSError('File not found')` error.

jdavies-st:
The failure in `blendmodels` is that the asn file does not exist.  Should we put one there?  In the end, that line is not actually testing `modelblender`, but testing `ModelContainer` and `datamodels.open()`.

stscieisenhamer:
About error in registry.py: Yes, the api has changed. It should read

```
final_asns = registry.callback.reduce('finalize', associations)
```

However, that should also be doctest skipped, since it will not operate as the previous one is also skipped.

Checking on second issue...

stscieisenhamer:
For constraint.py, that whole section `Examples` need be replaced:
```
    Examples
    --------

    Create a constraint where the attribute `attr` of an object
    matches the value `my_value`:

    >>> c = SimpleConstraint(value='my_value')
    >>> print(c)
    SimpleConstraint({'name': None, 'value': 'my_value'})

    To check a constraint, call `check_and_set`. A successful match
    will return a tuple of `True` and a reprocess list.
    >>> item = 'my_value'
    >>> c.check_and_set(item)
    (True, [])

    If it doesn't match, `False` will be returned.
    >>> bad_item = 'not_my_value'
    >>> c.check_and_set(bad_item)
    (False, [])

    A `SimpleConstraint` can also be initialized by a `dict`
    of the relevant parameters:
    >>> init = {'value': 'my_value'}
    >>> c = SimpleConstraint(init)
    >>> print(c)
    SimpleConstraint({'name': None, 'value': 'my_value'})

    If the value to check is `None`, the `SimpleConstraint` will
    succesfully match whatever object given. However, a new `SimpleConstraint`
    will be returned where the `value` is now set to whatever the attribute
    was of the object.
    >>> c = SimpleConstraint(value=None)
    >>> matched, reprocess = c.check_and_set(item)
    >>> print(c)
    SimpleConstraint({'name': None, 'value': 'my_value'})

    This behavior can be overriden by the `force_unique` paramter:
    >>> c = SimpleConstraint(value=None, sources=['attr'], force_unique=False)
    >>> matched, reprocess = c.check_and_set(item)
    >>> print(c)
    SimpleConstraint({'name': None, 'value': None})
```

nden:
@jdavies-st I am going to mark those doctests as `SKIP`. I think the `filecheck` code should probably raise a `OSError`, not `ValueError` in this case but that's for a different PR.

stscieisenhamer:
Wow. It would help if I ran the doctest on the right file. Replace that line with
```
    >>> c = SimpleConstraint(value=None, force_unique=False)
```
basically removing the `sources` parameter.

jwst/associations/tests/data/rules_level3_coron.py
```diff
@@ -17,11 +17,15 @@
 # --------------------------------
 # Start of the User-level rules
 # --------------------------------
-
+'''
```
stscieisenhamer: *facepalm*. this will be fixed in #2750

stscieisenhamer: Actually something happened here. The file did not start out with this double definition. The `class` line of the first part is correct with the inheritance. The second `class` line should not be there.

jwst/datamodels/container.py
```diff
@@ -39,12 +39,14 @@ class ModelContainer(model_base.DataModel):
...
-    persist: boolean. If True, do not close model after opening it
+    persist : bool
+        If True, do not close model after opening it.
+
 
     Examples
     --------
-    >>> container = datamodels.ModelContainer('example_asn.json')
-    >>> for dm in container:
+    >>> container = ModelContainer('example_asn.json') # doctest: +SKIP
```
jdavies-st: Since none of these examples work, maybe a blanket
```
__doctest_skip__ = ['ModelContainer']
```
might be better?

jwst/associations/tests/data/rules_level3_coron.py
nden: Ah, I see. I removed the inheritance because
```
data/rules_level3_coron.py:23: in <module>
    AsnMixin_OpticalPath,
E   NameError: name 'AsnMixin_OpticalPath' is not defined
data/rules_level3_coron.py:23: in <module>
    AsnMixin_Base
E   NameError: name 'AsnMixin_Base' is not defined
```
And I don't see them defined anywhere. Any suggestions?

stscieisenhamer: After full investigation, this file, `rules_level3_coron.py`, is not used anymore. At this point it is best to remove it permanently. Would explain all the wonky.

nden: file removed!

nden:
@stscieisenhamer `rules_level3_coron.py` was removed and tests pass again. Please review when you get a chance.

PR #2733, opened 2018-10-19
Revert "Add relresperror column to MIRI photom schema"

hbushouse:
Reverts spacetelescope/jwst#2721. Backing this out for now, because it's causing errors with all the current MIRI PHOTOM ref files in CRDS, which don't have the extra column in the table. Don't need it for 0.12.1. Will reinstate later after B7.2 delivery and we figure out how to get the existing MIRI PHOTOM ref files updated (they may get updated naturally through the team's upcoming CDP-7 delivery).


stscicrawford:
I agree this shouldn't go into 0.12.1.  But this raises a good question on how to handle an update like this and an update needed by the team.   This should probably be discussed in a separate issue

hbushouse:
Had we not been in the midst of a release cycle it wouldn't have been a problem to put it into the dev version of the code, except of course for all the reg test failures. But at least it wouldn't have caused problems in Ops, like it will now if we had left it in the upcoming 0.12.1 delivery.

PR #2734, opened 2018-10-19
Update extract_1d reffile docs

hbushouse:
Updates to the reference file info in the extract_1d step docs, making it more compatible with the updated formats for other reffile docs. Also includes minor formatting updates to a bunch of previously merged docs.

@hodge You only need to look at the extract_1d updates for review. Note that the only major change I made was to switch the ref file that's included as an example at the end. I thought the NIRISS ref file contained more interesting examples of things like background coeffs.

Another piece of #2675.


docs/jwst/extract_1d/reference_files.rst
```diff
@@ -189,44 +209,56 @@ value.  The polynomial will then be evaluated at each pixel within the
...
-        "id": "S1600A1",
+        "id": "FULL",
+        "region_type": "target",
+        "bkg_coeff": [[2014.5],[2043.5]],
+        "xstart":    4,
+        "xstop":  2044,
+        "ystart": 1792,
+        "ystop":  1972,
+        "dispaxis": 1,
+        "extract_width": 201
```
philhodge: This is the value that's in the file, but I'm not sure that this is what was intended.  The extraction width as specified by ystart and ystop is 181 pixels.  Specifying extract_width = 201 overrides that range (but uses the Y center location implied by ystart and ystop, 1882).  It might be less confusing to the user to have "extract_width": 181

PR #2737, opened 2018-10-19
fixed error in writting self.roiw to header

jemorrison:
When the wavelength dimension  is non-linear the roiw value is actually a table of values and not
a single value. When the wavelength dimension is non-linear self.roiw = None and writing
this value to the header as float(self.roiw) resulted in an error. 
This value is needed in the header when running in 'single' mode to
support background matching and outlier detector (blotting). Single mode will always
produce linear wavelength dimension ifucubes which have a single value in roiw.
This PR populates self.roiw with the mean of the roiw array when the wavelength dimension is non-linear.


jdavies-st:
I'm merging now so that tonight's regression test uses this new code.

PR #2740, opened 2018-10-20
Fix wiimatch-related doctest failures

mcara:
This should fix failing tests in `jwst.wiimatch` reported in https://github.com/spacetelescope/jwst/pull/2730#issuecomment-431498822


nden:
@mcara This appears  to fix the issues in wiimatch. How abou tthe first one in resample/bitmask? Is it sufficient to simply replace `interpret_bit_flags` with `interpret_bits_values`? If so I can do the change on my branch or you can do it here.

mcara:
@nden I forgot about `bitmask`. Back to work...

nden:
@mcara I rebased #2730 and got these changes but the doctests still fail. The numbers are different. The latest log is [here](https://travis-ci.org/spacetelescope/jwst/jobs/444685900). Can you take a look?

mcara:
@nden I can't make this work in doctests: it seems it is not capable of comparing approximate numbers. One solution would be to skip these tests here. Another would be to modify examples to use `np.allclose()` or to add a line like `np.set_printoptions(suppress=True)`. This would make the output more predictable (though not totally). I am against modifying examples by introducing operations that users would not normally do. So, IMO the only option for `wiimatch` would be to skip these tests.

mcara:
@nden If you use `pytest-doctestplus` then maybe it would be possible to do do something like this: http://docs.astropy.org/en/stable/development/testguide.html#handling-float-output ?

nden:
@mcara Sorry, I misread the numbers, didn't realize they are floating point differences.

mcara:
@nden Let's keep fingers crossed that https://github.com/spacetelescope/jwst/commit/8d56014 will solve the issue.

PR #2741, opened 2018-10-22
Switch from internal jwst bitmask to the stsci.tools version

mcara:
This PR should address `bitmask` doctest failures reported in https://github.com/spacetelescope/jwst/pull/2730#issuecomment-431498822

CC: @nden


jdavies-st:
I think it should go in 0.12.1 if #2730 is going in, as it resolves the `bitmask`-related issues there.

Issue #2497, opened 2018-09-12
Consolidate documentation of GAIN reference type


jaytmiller:
The current RTD CAL spec documents GAIN partially under gain_step and partially under ramp fitting.    Should we consolidate that to one common spec included for both steps/processes?   If so I'd be happy to fold it in with my other documentation pull request,  or file a separate one.

rizeladiaz:
Yes, these two files need to be consolidated.

jaytmiller:
OK,  I got it.  Since we're openly discussing it now,  I will fold it into my other "make it so" pull request,  and we will call this the review for that tiny piece.

hbushouse:
Perhaps in the process of fixing this particular issue with the GAIN docs, we should take the opportunity to reorganize the whole document, such that instead of having sections based on calibration steps - which then lists the ref files they use - we have it organized by ref file and within each ref file section mention which step(s) it's used by?

Of course one down side of this is that the dozens of WCS-related ref files would then end up scattered around the whole document, instead of being all grouped together in a single section under the `assign_wcs` step.

jaytmiller:
I like that reorg idea,  but I can't support that for 7.2 without delaying or dropping other work so we might want a different ticket and longer term goal.

I'd go one step further though:  the Step sections and the reference file sections would ideally have bidirectional links.  It's useful to be able to clearly index both Steps and reference file types,  and starting from either,  get to the other.

philhodge:
I like the idea of organizing by reference file, but it would help to have a table with a list of all steps, and for each step a list of all reference file types that are or could be used by the step.

jaytmiller:
I think that table is a very good idea,  probably clearer and easier to execute than inlining bidirectional links.   Steps and types could link to one coherent table if desired,  or not.

PR #2742, opened 2018-10-22
Update fringe and gain_scale reffile docs

hbushouse:
Main updates here are to the reference file descriptions for the fringe and gain_scale steps. But because we're now up to 3 steps that use the gain reference file, I've also rearranged how the gain and readnoise reffile docs are handled. There's now a single copy of the description of each of those reference files, which is included in the gain_scale, jump, and ramp_fit docs, as appropriate.

Also a few minor tweaks to previously updated step reffile docs are included.


hbushouse:
Addresses #2675 and #2497.

hbushouse:
Reviewers: Just concentrate on Fringe and Gain_Scale updates.

hbushouse:
@dmggh If they look OK, could you post an approval, so that I can merge? Thanks.

Issue #2746, opened 2018-10-22
Out of domain for acos


stscieisenhamer:
# Issue

From WFR2-OTB-20181004 processing run doggett/tests/wfr2/cstring/JWSTDP-2018_2-181017, running `calwebb_image3.cfg` gives the following:

<details><summary>Show log...</summary>
<p>

```
$ strun cfgs/calwebb_image3.cfg jw00759-o014_20181019t220201_image3_002_asn.json
2018-10-22 16:04:41,127 - stpipe.Image3Pipeline - INFO - Image3Pipeline instance created.
2018-10-22 16:04:41,129 - stpipe.Image3Pipeline.tweakreg - INFO - TweakRegStep instance created.
2018-10-22 16:04:41,130 - stpipe.Image3Pipeline.skymatch - INFO - SkyMatchStep instance created.
2018-10-22 16:04:41,132 - stpipe.Image3Pipeline.outlier_detection - INFO - OutlierDetectionStep instance created.
2018-10-22 16:04:41,133 - stpipe.Image3Pipeline.resample - INFO - ResampleStep instance created.
2018-10-22 16:04:41,134 - stpipe.Image3Pipeline.source_catalog - INFO - SourceCatalogStep instance created.
2018-10-22 16:04:41,134 - stpipe - INFO - Hostname: iljwdmscdp1.stsci.edu
2018-10-22 16:04:41,134 - stpipe - INFO - OS: Linux
2018-10-22 16:04:41,163 - stpipe.Image3Pipeline - INFO - Step Image3Pipeline running with args ('jw00759-o014_20181019t220201_image3_002_asn.json',).
2018-10-22 16:04:42,630 - stpipe.Image3Pipeline - INFO - Prefetching reference files for dataset: 'jw00759014001_02105_00001_nrcalong_cal.fits' reftypes = ['drizpars']
2018-10-22 16:04:43,057 - stpipe.Image3Pipeline - INFO - Prefetch for DRIZPARS reference file is '/grp/crds/cache/references/jwst/jwst_nircam_drizpars_0001.fits'.
2018-10-22 16:04:44,345 - stpipe.Image3Pipeline - INFO - Prefetching reference files for dataset: 'jw00759014002_02105_00001_nrcalong_cal.fits' reftypes = ['drizpars']
[...]

PR #2751, opened 2018-10-22
Fix a typo in branching by number of points

mcara:
This is an obvious typo that needs to be fixed. I am not sure though whether this is going to fix https://github.com/spacetelescope/jwst/issues/2746 by itself (though it might, depending on detected points).

CC: @stscieisenhamer


stscieisenhamer:
Nope, sorry, doesn't address #2746. Just tried it. 😞

PR #2753, opened 2018-10-23
Peg pytest to version 3.8.2 due to tmpdir bug in 3.9.1

jdavies-st:
This fix will allow our regression tests to run on Jenkins.

There seems to be a bug in how the `tmpdir` fixture works in `pytest 3.9` and above.


jdavies-st:
Btw, this seems to be tracked here, though I can't be sure this is the same error.

https://github.com/pytest-dev/pytest/issues/4181

drdavella:
(╯°□°）╯︵ ┻━┻

Issue #2747, opened 2018-10-22
RESOL reference file column names don't match the schema


hbushouse:
There's a discrepancy between the names of the columns in the actual MIRI RESOL reference file (jwst_miri_resol_0003.fits) and what's defined in miri_resolution.schema.yaml for the RESOLVING_POWER table. In particular, the data model schema has:
```
       - name: band
       - name: center
       - name: a_low
       - name: b_low
       - name: c_low
       - name: a_high
       - name: b_high
       - name: c_high
       - name: a_ave
       - name: b_ave
       - name: c_ave
```
while the actual reference file has column names:
```
SUB_BAND
R_CENTRE
R_A_LOW
R_B_LOW
R_C_LOW
R_A_HIGH
R_B_HIGH
R_C_HIGH
R_A_AVG
R_B_AVG
R_C_AVG
```

When loading this ref file into a `MiriResolutionModel` the names of columns that appear in the data model are those from the reference file. No errors or warnings are generated concerning the mismatch in column names.

jemorrison:
I can change the MiriResolutionModel to match what is in the reference file.
I think I am not using the datamodels when reading in this reference file in a standard way in
in cube_build because I read in the actual reference file column names (with capital letters)
in the code and I think ignore what is in the datamodels.

hbushouse:
It's not necessary to change the cubepar reference file schema to put the table column names in upper case. The column names are stored in TTYPEn keywords in the extension header and the FITS standard says "String comparisons with the values of TTYPEn keywords should not be case sensitive." So the lowercase names in the schema will still be a match to the uppercase names in the FITS file.

hbushouse:
Yes, you are using the datamodels when loading the RESOL reference file in `cube_build`:
```
ptab = datamodels.MiriResolutionModel(resol_filename)
```
But then the subsequent code accesses attributes within `ptab` by the names in the FITS file, as opposed to the names in the schema.

jemorrison:
So how is this different than how the miri_resolution data is stored ?

hbushouse:
Sorry, how is _what_ different than the miri_resolution data? The change involved in this PR is for the miri_resolution data.

jemorrison:
Why does it matter that the miri_resolution columns for the schema  had to be CAPITAL names to match the reference file and the cubepars columns for the schema do not have to be CAPITAL names even though the reference file has CAPITAL names.

hbushouse:
Fixed by #2757

PR #2757, opened 2018-10-23
updated MIRI resolution schema

jemorrison:
Updated schema for miri resolution reference file.
May also need to update cubepars reference file schemas to make the capital column names in reference file to the schema names. 
Addresses #2747


hbushouse:
Oh, need to add an entry to the CHANGELOG for this.

jemorrison:
oh I got it now. Thanks

PR #2760, opened 2018-10-23
 a) create an integration-specific model only if the dataset has  mor…

dmggh:
…e than 1 integration, and b) have code check for DO_NOT_USE keywords for the first and last frames for MIRI.


hbushouse:
There's no change log entry. I guess someone can add that when doing another PR.

Issue #2745, opened 2018-10-22
Missing extraction parameters


stscieisenhamer:
#  Issue

From WFR2-OTB-20181004 processing run doggett/tests/wfr2/cstring/JWSTDP-2018_2-181017, running `calwebb_nrslamp-spec2.cfg`, gives the following:

<details><summary>Show log...</summary>
<p>

```
$ strun /ifs/int/jwstc/pipeline/pkgs/miniconda3/envs/jwstdp-0.12.0-py36/lib/python3.6/site-packages/jwst-0.12.0-py3.6-linux-x86_64.egg/jwst/pipeline//calwebb_nrslamp-spec2.cfg /ifs/int/jwstc/store/doggett/tests/wfr2/cstring/JWSTDP-2018_2-181017/2018-10-19-215606/sdp/asn_creation/cal/level2/jw00791-o004_20181019t221608_nrslamp-spec2_002_asn.json
2018-10-22 15:53:21,247 - stpipe.Spec2NRSLamp - INFO - Spec2Pipeline instance created.
2018-10-22 15:53:21,248 - stpipe.Spec2NRSLamp.bkg_subtract - INFO - BackgroundStep instance created.
2018-10-22 15:53:21,249 - stpipe.Spec2NRSLamp.assign_wcs - INFO - AssignWcsStep instance created.
2018-10-22 15:53:21,250 - stpipe.Spec2NRSLamp.imprint_subtract - INFO - ImprintStep instance created.
2018-10-22 15:53:21,251 - stpipe.Spec2NRSLamp.msa_flagging - INFO - MSAFlagOpenStep instance created.
2018-10-22 15:53:21,252 - stpipe.Spec2NRSLamp.extract_2d - INFO - Extract2dStep instance created.
2018-10-22 15:53:21,253 - stpipe.Spec2NRSLamp.flat_field - INFO - FlatFieldStep instance created.
2018-10-22 15:53:21,254 - stpipe.Spec2NRSLamp.srctype - INFO - SourceTypeStep instance created.
2018-10-22 15:53:21,255 - stpipe.Spec2NRSLamp.straylight - INFO - StraylightStep instance created.
2018-10-22 15:53:21,256 - stpipe.Spec2NRSLamp.fringe - INFO - FringeStep instance created.
2018-10-22 15:53:21,257 - stpipe.Spec2NRSLamp.pathloss - INFO - PathLossStep instance created.
[...]

PR #2761, opened 2018-10-23
NRS grating=MIRROR should trigger the imaging pipeline

nden:
assign_wcs should run the imaging WCS pipeline when grating=MIRROR. The specific case in DMS is
```
EXP_TYPE=NRS_LAMP
GRATING=MIRRROR
```
This was processed as a spectral WCS based on EXP_TYPE only.

Related to #2745.

### DIFF
--- a/jwst/assign_wcs/nirspec.py
+++ b/jwst/assign_wcs/nirspec.py
@@ -49,8 +49,11 @@ def create_pipeline(input_model, reference_files):
     reference_files : dict
         {reftype: reference_file_name} mapping.
     """
-    exp_type = input_model.meta.exposure.type.lower()
-    pipeline = exp_type2transform[exp_type](input_model, reference_files)
+    if input_model.meta.grating.lower() == "mirror":
+        pipeline = imaging(input_model, reference_files)
+    else:
+        exp_type = input_model.meta.exposure.type.lower()
+        pipeline = exp_type2transform[exp_type](input_model, reference_files)
     if pipeline:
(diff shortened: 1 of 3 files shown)


nden:
Are we using the `0.12.0` Changelog section for this patch or should we create a new section for `0.12.1`?

hbushouse:
I've been continuing to use the 0.12.0 section, because eventually that'll all be relabeled as the B7.2 section.

jwst/assign_wcs/nirspec.py
```diff
@@ -49,8 +49,11 @@ def create_pipeline(input_model, reference_files):
...
-    exp_type = input_model.meta.exposure.type.lower()
-    pipeline = exp_type2transform[exp_type](input_model, reference_files)
+    if input_model.meta.instrument.grating.lower() == "mirror":
+        pipeline = imaging(input_model, reference_files)
+    else:
+        exp_type = input_model.meta.exposure.type.lower()
+        pipeline = exp_type2transform[exp_type](input_model, reference_files)
     if pipeline:
         log.info("Created a NIRSPEC {0} pipeline with references {1}".format(
                 exp_type, reference_files))
```
hbushouse: If the new line 52 is True, then line 55 won't be executed and hence "exp_type" won't yet be defined in line 59 (caught by Travis tests - not me).

hbushouse:
Don't forget about updating the change log.

nden:
@hbushouse Done.

PR #2767, opened 2018-10-24
Fix units for Read Noise Variance in ramp_fit

### LINKED ISSUE #2756: Incorrect Units for Read Noise Variance in Ramp Fit
The units of the Read Noise variance that are output from the ramp_fit step are in (electrons/s)**2 the correct units are (DN/s)**2. That matches the units of the Poisson variance. 
  I have a fix for the issue and a test that shows the error and will be submitting a PR.

mwregan2:
This PR has the changes that fix the Issue and initial tests for ramp_fitting.

Resolves #2756.


hbushouse:
There's a unit test that needs fixing, in `test_ramp_fit.py`. It threw an import error in the Travis tests.

jwst/ramp_fitting/tests/test_ramp_fit.py
```diff
@@ -0,0 +1,168 @@
+-import pytest
```
hbushouse: Looks like the leading "-" at the beginning of this line is a typo and causing Travis to crash.

jwst/ramp_fitting/tests/test_ramp_fit.py
```diff
@@ -0,0 +1,168 @@
...
+def setup_inputs(ngroups=10, readnoise=10, nints=1,
+                 nrows=1032, ncols=1024, nframes=1, grouptime=1.0,gain=1, deltatime=1):
+        
+        print('readnoise', readnoise)
+        print('gain',gain)
+        times = np.array(list(range(ngroups)),dtype=np.float64) * deltatime
+        gain = np.ones(shape=(nrows, ncols), dtype=np.float64) * gain
+        err = np.ones(shape=(nints, ngroups, nrows, ncols), dtype=np.float64)
+        data = np.zeros(shape=(nints, ngroups, nrows, ncols), dtype=np.float64)
+        pixdq = np.zeros(shape=(nrows, ncols), dtype= np.float64)
```
philhodge: pixdq ought to be an integer type, e.g. np.uint32.

.gitignore
```diff
@@ -19,3 +19,6 @@ relic/
 *wpu
 .pytest_cache
 results.xml
+*json
```
jdavies-st: Is there a reason these are added to `.gitignore`?  I don't think we want to to do this, as actually do have JSON and FITS files that are currently part of the repo.

If you're trying to make sure temporary outputs of tests don't get into the repo, then use the pytest fixture `tmpdir` with the test.

jwst/ramp_fitting/tests/test_ramp_fit.py
```diff
@@ -0,0 +1,168 @@
...
+    model1.data[0, 1, 500, 500] = 30.0
+    model1.data[0, 2, 500, 500] = 50.0
+    model1.data[0, 3, 500, 500] = 70.0
+    model1.data[0, 4, 500, 500] = 90.0
+    model1.data[0, 5, 500, 500] = 110.0
+    model1.data[0, 6, 500, 500] = 130.0
+    model1.data[0, 7, 500, 500] = 150.0
+    model1.data[0, 8, 500, 500] = 170.0
+    model1.data[0, 9, 500, 500] = 190.0
+    slopes = ramp_fit(model1, 64000, True, rnModel, gain, 'GLS', 'optimal')
```
philhodge: The GLS option isn't supported.  I don't have any objection to this test; I'm just a little surprised that it still works.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -464,7 +464,7 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
         del var_p4_int
         del var_p4_int2
 
-    del gain_2d
+#    del gain_2d  we still need this to convert back to DN
```
jdavies-st: This line can be deleted.

jwst/ramp_fitting/tests/test_ramp_fit.py
```diff
@@ -0,0 +1,168 @@
+-import pytest
+import numpy as np
+
+from jwst.ramp_fitting.ramp_fit import ramp_fit
+from jwst.datamodels import dqflags
+from jwst.datamodels import MIRIRampModel
+from jwst.datamodels import GainModel, ReadnoiseModel
+from nose.tools import assert_almost_equals
+
+#@pytest.mark.skip(reason="not using now")
```
jdavies-st: All these commented-out `@pytest.mark.skip` lines can be deleted as well.

mwregan2:
I have the commented out pytest skip lines in there to make the incremental addition of new tests run faster. There is no reason to run a large set of tests that pass multiple times when I’m working on just one step. I think I should change to skipif as a better way to control things rather than commenting out the line.
Mike

mwregan2:
This fix is not correct. The full fix is coming.

PR #2770, opened 2018-10-24
Check table column names against schema

bernie-simon:
This update checks table column names against the column names in the schema datatype and throws a ValueError if they don't match. Matching is case insensitive. Previously the code only checked if the number of columns matched. There's no option to change the error into a warning as columns are accessed by name.


jwst/datamodels/util.py
```diff
@@ -319,15 +319,18 @@ def gentle_asarray(a, dtype):
         elif in_dtype.fields is not None and out_dtype.fields is not None:
             if in_dtype == out_dtype:
                 return a
-            if len(in_dtype) != len(out_dtype):
+            in_names = {n.upper() for n in in_dtype.names}
+            out_names = {n.upper() for n in out_dtype.names}
+            if in_names == out_names:
+                # Change the dtype name to match the fits record names
+                # as the mismatch causes case insensitive access to fail
+                out_dtype.names = in_dtype.names
```
philhodge: I don't see what this has to do with case insensitive column access.  This just replaces the column names.  The new names will still have a definite case for each letter in the name.

hbushouse: Is `in_names` the column names from the FITS file and `out_names` the names from the schema? If so, then I guess line 327 forces the names from the FITS file to be the ones that will show up in the data model attributes, right? I'm honesty not sure which direction we should be coercing things. After just a little bit of thought, it seems to me like the names in the schema ought to have priority and hence recast the FITS file names to match the schema (as opposed to the other way around).

bernie-simon: The coercion in line 327 is not part of this fix. It is from a previous fix to a problem where you could not access a table column if the case of the column in the schema differed from the case of the column in the table. Numpy asanyarray will coerce the column names to the names in the schema, so line 327 sets the names in the schema to the names in the table. so reads are performed correctly. Doing the reverse of the assignment in line 327 has no effect, because asanyarray will change the names to match the schema.

PR #2776, opened 2018-10-25
Update Jenkinsfile to rely on CI utils library

jdavies-st:
Let's see if Docker lets us do this.


jdavies-st:
The original purpose of this PR was implement `pytest-xdist` functionality to our per-PR CI tests.  This paves the way for that, but doesn't actually implement that.

PR #2782, opened 2018-10-26
Updated lastframe unit test to test NIR data

cracraft:
Added tests to test whether the full step properly rejects NIR data and runs for MIR data, resulting in 100% code coverage for lastframe step.


jdavies-st:
@nden, should we be adding anything to CHANGES.md for PRs like this that just add to the unit tests?

jdavies-st:
Test failure on Jenkins is not really a test failure.

PR #2786, opened 2018-10-26
Add a function to generate documentation from schema files

bernie-simon:
This update adds the function build_docstring to schema.py. It takes a datamodels class and extracts field values from its schema for each schema item associated with a fits hdu. It then inserts the field values into a format string, which is passed as the second argument. This update also includes a new script, schemadoc, which calls it.


jwst/datamodels/schema.py
```diff
@@ -295,3 +298,97 @@ def asdf_file_resolver(uri):
...
+    Build a docstring for the specified DataModel class from its schema.
+
+
+    Parameters
+    ----------
+
+    klass : Python class
+        A class instance of a datamodel
+    template: str
+        A string format template to be applied to each schema item
```
stscieisenhamer: Add a description for what is returned.

nden:
This shows a conflict in datamodels.

hbushouse:
Shall we revert until it gets fixed?

PR #2787, opened 2018-10-26
Update doc strings for cube_build

jemorrison:
Add the doc string to all modules. Also did a general clean up of comments in the code to make it
follow python rules.

### DIFF
--- a/docs/jwst/cube_build/index.rst
+++ b/docs/jwst/cube_build/index.rst
@@ -10,4 +10,12 @@ Cube Building
    examples.rst
    reference_files.rst
    
-.. automodapi:: jwst.cube_build
+.. automodapi:: jwst.cube_build.cube_build_step
+.. automodapi:: jwst.cube_build.data_types
+.. automodapi:: jwst.cube_build.cube_build
+.. automodapi:: jwst.cube_build.instrument_defaults
+.. automodapi:: jwst.cube_build.cube_build_io_util
+.. automodapi:: jwst.cube_build.cube_build_wcs_util
+.. automodapi:: jwst.cube_build.ifu_cube
+.. automodapi:: jwst.cube_build.cube_cloud
+.. automodapi:: jwst.cube_build.cube_overlap
--- a/docs/jwst/cube_build/index.rst
+++ b/docs/jwst/cube_build/index.rst
@@ -11,11 +11,3 @@ Cube Building
    reference_files.rst
    
 .. automodapi:: jwst.cube_build.cube_build_step
-.. automodapi:: jwst.cube_build.data_types
-.. automodapi:: jwst.cube_build.cube_build
-.. automodapi:: jwst.cube_build.instrument_defaults
-.. automodapi:: jwst.cube_build.cube_build_io_util
-.. automodapi:: jwst.cube_build.cube_build_wcs_util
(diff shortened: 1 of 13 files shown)


docs/jwst/cube_build/index.rst
```diff
@@ -10,4 +10,12 @@ Cube Building
    examples.rst
    reference_files.rst
    
-.. automodapi:: jwst.cube_build
+.. automodapi:: jwst.cube_build.cube_build_step
+.. automodapi:: jwst.cube_build.data_types
```
jdavies-st: Do we want all these lower level functions in the main docs?  I would suspect all we need is `cube_build_step` for the docs.  Or do we expect users to need to call these lower level functions?

jwst/cube_build/coord.py
```diff
@@ -1,19 +1,19 @@
...
-    CRVAL1,CRVAL2 and the tangent point
+    """ Compute the tangent projection coordinates (xi,eta) from ra,dec
+    using crval1 and crval2 (the tangent point).
 
-    Input Values
+    Parameters
     ____________
-    CRVAL1: ra value of tangent point
-    CRVAL2: dec value of tangent point
+    crval1: ra value of tangent point
```
jdavies-st: Here, it should indicate the type, so
```
crval1 : float
    RA value of tangent point
```
Sae for `crval2`.

jwst/cube_build/cube_build_wcs_util.py
```diff
@@ -146,17 +142,19 @@ def find_footprint_NIRSPEC(input, flag_data, coord_system):
...
-    Test for 0/360 wrapping in ra values. If exists it makes it difficult to determine
+    """Test for 0/360 wrapping in ra values.
+
+    If exists it makes it difficult to determine
     ra range of IFU cube. So put them all on "one side" of 0/360 border
 
     Input
     -----
-    ravalues a numpy array of ra values
+    ravalues : a numpy array
```
jdavies-st:
```
ravalues : numpy.ndarray
    RA values
```

docs/jwst/cube_build/index.rst
jemorrison: I was not sure if every module had to be listed in index.rst or just the main code.
I will make it just cube_build_step

jemorrison:
@jdavies-st I am not sure why there was a problem with the jenkins testing.

jwst/cube_build/coord.py
```diff
@@ -62,11 +66,13 @@ def std2radec(crval1, crval2, xi, eta):
     Parameters
     ____________
     crval1 : float
-      ra value of tangent point
+      RA value of tangent point
     crval2 : float
-      dec value of tangent point
-    xi, eta : float, float
-      rectangular coordinates of tangent plane projected ra,dec
+      RA value of tangent point
```
jdavies-st: `DEC value` here I believe.

jdavies-st:
Oh, it's a problem with the `Jenkinsfile` not being pinned to `numpy 1.15.2` in your branch.  Not a problem.

jdavies-st:
Jenkins failure is unrelated to this PR.  A rebase to `jwst/master` and a force push should fix it.

PR #2791, opened 2018-10-26
restore tests

nden:
Removes `convert_to_uncal.py` to allow tests in jwst/lib to run. If this module is still needed we need to put the missing data file somewhere.

There are no tests in `timeconversion`.


jdavies-st:
And it looks like all the failures are happening because of failing `configobj` doctests.  So feel free to add those back into the ignore list, and then we can at least run the `jwst/lib` tests.

What does `convert_to_uncal.py` do?  @hbushouse ?

nden:
@stscieisenhamer I see a failure in `jwst.lib` which looks like an API change.
Do you know how it should be resolved?

https://travis-ci.org/spacetelescope/jwst/jobs/447844411#L1464

nden:
Oh, I see. The formatting was broken. Thanks.

PR #2794, opened 2018-10-27
Remove dask dependency

jdavies-st:
`dask` is not a dependency of `jwst`, yet it is listed as one in our CI scripts.  As such, it brings in a bunch of its own dependencies and probably makes our CI builds take longer.  Any ideas why it was added @jhunkeler?

This PR removes it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ env:
         - CONDA_DEPENDENCIES="asdf
                               astropy
                               crds
-                              dask
                               drizzle
                               flake8
                               gwcs
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,7 +14,6 @@ def conda_packages = [
     "asdf",
     "astropy",
     "crds",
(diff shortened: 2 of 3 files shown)


jhunkeler:
Last I remember this was a hacky workaround. `scikit-image` requires `dask` but the dependency wasn't satisfied by the package in the `defaults` channel, so under certain circumstances scripts would throw an `ImportError`.

I _think_ `scikit-image` was getting pulled in by `scipy`.

jdavies-st:
It looks like `scikit-learn` developers made `dask` a requirement a year ago.

https://github.com/scikit-image/scikit-image/commit/044382773594925868eb0b5fe4be2203e6b15b70#diff-f7b7288f81db184bb127adf1389dafc4R8

So I suspect we can remove it.

jdavies-st:
RT run off this PR is here.

https://boyle.stsci.edu:8081/job/RT/job/JWST-jdavies/43/

Issue #2716, opened 2018-10-17
tweakreg: KeyError: 'id'


stscieisenhamer:
# Issue
Running `calwebb_image3` on some EMTB data, the following occurs

<details><summary>Show log...</summary>
<p>

```
$ strun cfgs/calwebb_image3.cfg jw00601-o012_image3_001_asn.json
2018-10-17 13:18:04,033 - stpipe.Image3Pipeline - INFO - Image3Pipeline instance created.
2018-10-17 13:18:04,035 - stpipe.Image3Pipeline.tweakreg - INFO - TweakRegStep instance created.
2018-10-17 13:18:04,036 - stpipe.Image3Pipeline.skymatch - INFO - SkyMatchStep instance created.
2018-10-17 13:18:04,038 - stpipe.Image3Pipeline.outlier_detection - INFO - OutlierDetectionStep instance created.
2018-10-17 13:18:04,039 - stpipe.Image3Pipeline.resample - INFO - ResampleStep instance created.
2018-10-17 13:18:04,040 - stpipe.Image3Pipeline.source_catalog - INFO - SourceCatalogStep instance created.
2018-10-17 13:18:04,041 - stpipe - INFO - Hostname: inanna2.local.stsci.edu
2018-10-17 13:18:04,041 - stpipe - INFO - OS: Darwin
2018-10-17 13:18:04,082 - stpipe.Image3Pipeline - INFO - Step Image3Pipeline running with args ('jw00601-o012_image3_001_asn.json',).
2018-10-17 13:18:04,719 - stpipe.Image3Pipeline - WARNING - /Users/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/asdf/block.py:708: RuntimeWarning: invalid value encountered in greater
  if (array > 2**52).any():

2018-10-17 13:18:04,773 - stpipe.Image3Pipeline - INFO - Prefetching reference files for dataset: 'jw00601012001_02102_00004_nrcblong_cal.fits' reftypes = ['drizpars']
2018-10-17 13:18:05,171 - stpipe.Image3Pipeline - INFO - Prefetch for DRIZPARS reference file is '/Users/eisenham/Downloads/crds-cache/references/jwst/nircam/jwst_nircam_drizpars_0001.fits'.
2018-10-17 13:18:05,590 - stpipe.Image3Pipeline - INFO - Prefetching reference files for dataset: 'jw00601012001_02102_00005_nrcblong_cal.fits' reftypes = ['drizpars']
[...]

PR #2796, opened 2018-10-28
Make the code more robust with bad catalogs

mcara:
This PR should address issue https://github.com/spacetelescope/jwst/issues/2716. However, that specific test will fail next in `skymatch` because there are not enough data in overlapping area (probably bad WCS or all pixels flagged in DQ). If you want that one to succeed as well, I'll make another PR.

CC: @stscieisenhamer @hbushouse @stscicrawford

### DIFF
--- a/jwst/tweakreg/tweakreg_catalog.py
+++ b/jwst/tweakreg/tweakreg_catalog.py
@@ -61,7 +61,7 @@ def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
     if len(sources) > 0:
         catalog = sources[columns]
     else:
-        catalog = Table(names=columns, dtype=(np.int_, np.float_, np.float_, np.float_))
-        catalog.add_row([1, 0.5, 0.5, 100.])
+        catalog = Table(names=columns, dtype=(np.int_, np.float_, np.float_,
+                                              np.float_))
 
     return catalog
--- a/jwst/tweakreg/tweakreg_catalog.py
+++ b/jwst/tweakreg/tweakreg_catalog.py
@@ -1,4 +1,4 @@
-from astropy.table import Table, Column
+from astropy.table import Table
 import numpy as np
 from photutils import detect_threshold, DAOStarFinder
 
--- a/jwst/tweakreg/tweakreg_catalog.py
+++ b/jwst/tweakreg/tweakreg_catalog.py
@@ -1,3 +1,5 @@
+from astropy.table import Table, Column
+import numpy as np
 from photutils import detect_threshold, DAOStarFinder
 
 from ..datamodels import ImageModel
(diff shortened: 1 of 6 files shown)


jwst/tweakreg/tweakreg_catalog.py
```diff
@@ -56,6 +58,10 @@ def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
     sources = daofind(model.data)
 
     columns = ['id', 'xcentroid', 'ycentroid', 'flux']
-    catalog = sources[columns]
+    if len(sources) > 0:
+        catalog = sources[columns]
+    else:
+        catalog = Table(names=columns, dtype=(np.int_, np.float_, np.float_, np.float_))
+        catalog.add_row([1, 0.5, 0.5, 100.])
```
larrybradley: Why are you adding a dummy source here?

larrybradley: Why not just return an empty table and check for its length when you need to use it?

mcara: @larrybradley Because I was too tired to remove debugging code... Thanks for looking at this with fresh eyes :smile: I'll remove it.

mcara: > Why not just return an empty table and check for its length when you need to use it?

Because saving empty catalogs would produce empty files (see https://github.com/spacetelescope/jwst/blob/d7daa6c9cc60fc9870add2950ca06a424056db28/jwst/tweakreg/tweakreg_step.py#L84). Catalog files with headers but no sources look much better (IMO) and less worrisome than completely empty files, if we even can save empty tables - I get a crash:

```python
ValueError: Inconsistent data column lengths: set()
```

larrybradley: Why do you need to save an empty table?  Does your rely on a saved catalog (not in memory)?

mcara: No, it does not, but the pipeline can save catalogs to files. As an user it would be weird to see that catalogs have been saved only for some images but not for others. Having a file with a header (but no sources) is much more informative (instead of being puzzling).

larrybradley: Can't we just issue a warning?

mcara: We could if that's fine with @hbushouse . However, it would still be puzzling not to find all files and searching through the pipeline log to figure out why some catalogs have not been written out seems like a less obvious solution.

hbushouse: For the purpose of running in a pipeline, where tweakreg is just 1 step and other steps could continue to work without it, I'd like to see just a warning issued and the step return with a "skipped" status. The remaining `calwebb_image3` steps, like `skymatch`, `outlier_detection`, and `resample` can still be executed using the original (untweaked) WCS info.

hbushouse: In that case I think it makes sense to just issue a warning for the particular catalog that had no sources, and move on with the rest of processing (without skipping the entire step). And let the logic that's already there later on figure out if it really has enough catalog data to work with.

mcara: @hbushouse That's being done in https://github.com/spacetelescope/jwst/blob/1316c3f6a1ccf767698dca30bbdf9e1a86112ef5/jwst/tweakreg/tweakreg_step.py#L71

larrybradley: @mcara For zero detected sources, I would change that from an INFO message to a warning.

mcara: @larrybradley Done

PR #2797, opened 2018-10-29
Fix logic for computing catalog footprints

mcara:
This PR addresses the issue https://github.com/spacetelescope/jwst/issues/2746. However, it will only work (tests will pass *tweakreg* stage) only if https://github.com/spacetelescope/jwst/pull/2796 is merged as well.

Unfortunately this PR makes algorithm a little inefficient when image catalogs contain 1 or two stars. Optimization can be addressed later if need be. For now, IMO, this should be robust enough to work.

CC: @hbushouse @stscieisenhamer @stscicrawford

### DIFF
--- a/jwst/tweakreg/wcsimage.py
+++ b/jwst/tweakreg/wcsimage.py
@@ -345,7 +345,7 @@ class WCSImageCatalog():
             stored (for convenience) within `WCSImageCatalog` object.
 
         """
-        self._shape = shape
+        self.imshape = shape
         self._name = name
         self._catalog = None
         self._bb_radec = None
@@ -438,6 +438,9 @@ class WCSImageCatalog():
 
     @imshape.setter
     def imshape(self, shape):
+        if not all(npix > 0 for npix in shape):
+            raise ValueError("Null image: Image dimension must be positive.")
+
         self._shape = shape
 
     @property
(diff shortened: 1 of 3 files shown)


jwst/tweakreg/wcsimage.py
```diff
@@ -655,43 +658,19 @@ def _calc_cat_convex_hull(self):
...
-            vx /= norm
-            vy /= norm
-
-            xv = [x[0] + vx, x[1] + vx, x[1] + vx, x[0] - vx, x[0] + vx]
-            yv = [y[0] + vy, y[1] + vy, y[1] + vy, x[0] - vx, x[0] + vx]
-
-            ra, dec = self.tanp_to_world(xv, yv)
-
-        else:
+        if len(x) > 2:
```
hbushouse: So what happens when len(x) <= 2 and this code block doesn't get executed?

jwst/tweakreg/wcsimage.py
```diff
@@ -1296,9 +1275,14 @@ def __init__(self, catalog, name=None):
         name : str, None, optional
             Name of the reference catalog.
 
+        footprint_tol : float, optional
+            Matching tolerance in arcsec. This is used to estimate catalog's
```
hbushouse: The `imalign` doc string says tolerance is in units of pixels. Is this an inconsistency or is this a different tolerance?

mcara: @hbushouse It is inconsistency (pixels were used in the old tweakreg). `spec` however, was documented correctly: https://github.com/spacetelescope/jwst/blob/d7daa6c9cc60fc9870add2950ca06a424056db28/jwst/tweakreg/tweakreg_step.py#L45

jwst/tweakreg/wcsimage.py
mcara: When `len(x) <= 2` footprint (already) calculated based on image size/shape is used. This will lead to poor estimates of "overlapping" but it should work.

hbushouse: 👍

jwst/tweakreg/wcsimage.py
hbushouse: 👍

PR #2799, opened 2018-10-29
Fix x2d bounding box sizes and object wcs information for wfss

### LINKED ISSUE #2762: WFSS bounding boxes in extract_2d
For the WFSS modes, when `extract_2d` is run, it makes a cutout around each objects dispersion for each order. The size of the cutout is found by taking the corners of the minimum bounding box that is stored in the source catalog. Those corners are used along with the min and max wavelengths for the filter in use to calculate the extraction box. The dispersed trace is inside that box.

Currently, the bounding box is set to the size of the cutout. However, there's a bug somewhere which is causing the bounding box attached to each slit to be the same when it's returned from the MultiSlitObject. The grism extraction method in `extract_2d` is reporting creating the correct bounding box, but the returned object either ignores it, or always returns the bounding box for the same slit in the MultiSlit object. 

Additionally, I wonder if we should be making the bounding box actually be contained to the visible trace? This would mean adding the calculations for those pixels in the dispersion direction. There's only the size of the catalog bounding box to constrain the cross dispersion direction.

sosey:
Fixes #2762 , where the bounding box covers the entire cutout. I'll make another issue for setting the bounding box to just the size of the trace inside the cutout.

The wcs returning the same info for every slit is also fixed, it was my own stupid stupid oversight, thanks @nden

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -156,6 +156,8 @@ extract_2d
 
 - Added unit tests for grism modes [#2649]
 
+- Bounding box sizes in extracted WFSS exposures now correctly cover entire extraction [#2799]
+
 firstframe
 ----------
 


sosey:
I also added a nicer fail message in the MultiSlitModel when a slit is asked for that doesn't exist.

Issue #2420, opened 2018-08-24
Apply JWST velocity correction to WFSS modes


nden:
Similarly to #2406 append the velocity correction at the end of the wavelength computation for these modes.

stscicrawford:
This should include NIRISS SOSS mode as well.

nden:
NIRISS SOSS is done already. Only the modes that @sosey implemented need to be updated.

sosey:
Has this been discussed with @NorPirzkal or the niriss team?  Wavelength computation for WFSS modes already includes some measure of distortion, and fitted from the observed trace location I believe, I'm not sure that the proper place to put this correction is directly after traversal of the trace polynomials or directly on the wavelength entered by the user (and just before it's returned in the forward direction?

stscicrawford:
This was a request from Kevin Volk on https://jira.stsci.edu/browse/JP-245

nden:
Fixed in #2801

PR #2801, opened 2018-10-29
Added velocity correction to WFSS and Nircam TSO

sosey:
Implements #2420 for WFSS modes and NIRCAM TSGRISM

The velocity correction gets called before going into the trace polynomials.
See also discussion here: https://jira.stsci.edu/browse/JP-245

@stscicrawford @nden not sure what milestone to use

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ ami
 
 assign_wcs
 ----------
+ - Added velocity correction model to the WFSS and TSGRISM wcs pipelines [#2801]
 
 associations
 ------------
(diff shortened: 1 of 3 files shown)


hbushouse:
It probably wouldn't do any harm to try to include in 0.12.1, but last minute new features always make me nervous. I think we should probably hold off for now. If we do a 0.12.2, it could go in there, or we just wait for 0.13.0. It isn't something that DMS needs.

sosey:
I made up a wcsinfo.velosys number to test with the update and the results seemed reasonable (very small changes on the wavelength value) in both instruments. I'm good with whenever the team wants to push this.

Issue #2804, opened 2018-10-30
Bug in tweakreg/imalign


hbushouse:
#2796 introduced a bug in imalign.py, which was caught by the nightly regression tests:
```
            if len(group.catalog) < minobj:
                for i in group:
                    log.warning("Not enough sources to align image {:s}. "
                                "This image will not be aligned."
                                .format(i.meta['image_model'].meta.filename))
                    i.meta['image_model'].meta.cal_step.tweakreg = "SKIPPED"
                    i.meta['skip_alignment'] = True
                skipped_imcat.append(group)
    
            else:
>               i.meta['skip_alignment'] = False
E               UnboundLocalError: local variable 'i' referenced before assignment

../../jwst/tweakreg/imalign.py:198: UnboundLocalError
```
In the new code, when len(group.catalog) is >= minobj, the variable "i" is not set when executing the "else" half of the if/else block.

mcara:
Yeah... I'll fix it in a moment. Sorry

mcara:
@hbushouse This should be fixed via https://github.com/spacetelescope/jwst/pull/2805 Please merge that PR at your earliest convenience.

hbushouse:
Fixed by #2805

PR #2805, opened 2018-10-30
Fix burning bug (undefined i) in tweakreg

mcara:
Fixes the bug in https://github.com/spacetelescope/jwst/issues/2804.


hbushouse:
By running these updates on the NIRCam `calwebb_image3` regression test, I've confirmed that this fixes the problem. It now succeeds.

PR #2807, opened 2018-10-30
WIP - Dqinittesting unit test problems with dq_init.

cracraft:
I'm attempting to write unit tests for dq_init that do not rely on calling a reference file, modifying it, writing it back out to a file and reading the file name into the full DQInitStep in order to avoid all of the file reading/writing. So I attempted to create a data model for MaskModel (in setup_models.py) and use the model for the mask as input to the function do_dqinit rather than the full step. When that kept failing to properly map the dq flags and giving me the following error: 
        # assert that the pixels read back in match the mapping from ref data to science data
>       assert(dqdata[100, 100] == dqflags.pixel['DEAD'])
E       assert 2 == 1024

test_dq_init.py:62: AssertionError
----------------------------- Captured stdout call -----------------------------
(0, 0, 0, 0)
------------------------------ Captured log call -------------------------------
reffile_utils.py           115 DEBUG     ref xstart=1, xsize=1032, ystart=1, ysize=1024
reffile_utils.py           115 DEBUG     ref xstart=1, xsize=1032, ystart=1, ysize=1024
reffile_utils.py           158 DEBUG     sci xstart=1, xsize=1032, ystart=1, ysize=1024
reffile_utils.py           158 DEBUG     sci xstart=1, xsize=1032, ystart=1, ysize=1024

I decided to read in the actual reference file (that worked fine when testing the full step) as a model, and feeding that model into the do_dqinit function. That failed with the same error, which tells me that the full step is initializing something that allows this to work that the function is not. Is there a way for me to test this strictly with models, or do I have to read the reference files in to get the mapping to correctly read the dq_def and know that 'DEAD' pixel in the mask should map to 'DEAD' pixel in the output?

Also, how do I create multiple branches and commit just the file I want without it uploading all of the recent commits? Firstframe and lastframe have nothing to do with this commit.

### DIFF
--- a/jwst/dq_init/tests/setup_models.py
+++ b/jwst/dq_init/tests/setup_models.py
@@ -1,0 +1,86 @@
+from jwst.datamodels import MIRIRampModel
+from jwst.datamodels import ImageModel
+from jwst.datamodels import MaskModel
+import numpy as np
+
+
+def make_rawramp(nints,ngroups,ysize,xsize):
+    # create the data and groupdq arrays
+    csize = (nints, ngroups, ysize, xsize)
+    data = np.full(csize, 1.0)
+    # pixeldq = np.zeros((ysize,xsize), dtype=int)
+
+    # create a JWST datamodel for MIRI data
+    dm_ramp = MIRIRampModel(data=data)
+    if (ysize == 1024) and (xsize == 1032):
+        dm_ramp.meta.subarray.xstart = 1
+        dm_ramp.meta.subarray.xsize = xsize
+        dm_ramp.meta.subarray.ystart = 1
+        dm_ramp.meta.subarray.ysize = ysize
+    else:
+        # how to set up for subarray... assume
+        # specific subarray values or allow
+        # user to specify?
+        dm_ramp.meta.subarray.xstart = 1
+        dm_ramp.meta.subarray.xsize = xsize
+        dm_ramp.meta.subarray.ystart = 1 # 467 for MASK1550
+        dm_ramp.meta.subarray.ysize = ysize
+
+    return dm_ramp
+
+
+def make_rampmodel(nints, ngroups, ysize, xsize):
+    # create the data and groupdq arrays
+    csize = (nints, ngroups, ysize, xsize)
+    data = np.full(csize, 1.0)
+    pixeldq = np.zeros((ysize, xsize), dtype=int)
+    groupdq = np.zeros(csize, dtype=int)
+
+    # create a JWST datamodel for MIRI data
+    dm_ramp = MIRIRampModel(data=data, pixeldq=pixeldq, groupdq=groupdq)
+
+    dm_ramp.meta.instrument.name = 'MIRI'
+    dm_ramp.meta.observation.date = '2018-01-01'
+    dm_ramp.meta.observation.time = '00:00:00'
+    dm_ramp.meta.subarray.xstart = 1
+    dm_ramp.meta.subarray.xsize = xsize
+    dm_ramp.meta.subarray.ystart = 1
+    dm_ramp.meta.subarray.ysize = ysize
+
+    return dm_ramp
+
+
+def make_imagemodel(ysize, xsize):
+    # create the data array
+    csize = (ysize, xsize)
+    data = np.full(csize, 1.0)
+    err = np.zeros(csize)
+    dq = np.zeros(csize, dtype=int)
+    im = ImageModel(data=data, err=err, dq=dq)
+
+    return im
+
+
+def make_maskmodel(ysize, xsize):
+    # create a mask model for the dq_init step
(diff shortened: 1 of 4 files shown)


jdavies-st:
Thanks for the PR!  I can see what is going on here and might have a couple solutions.

The issue is that `MaskModel` (and the other models that use `dq_def` to map an input dq array to the standard bit values) is that this mapping only occurs once when the model is opened or instantiated.  The mapping happens in `__init__()`, and the original array as was contained in the FITS file is not preserved in memory anywhere.  Once the FITS file is read in (or the `numpy` array is passed to `MaskModel()`), it disappears.  There's no continuous mapping of these bit flags such as `'BAD'` to the original bit value in the FITS file.  It's one-way.

The design for that is because we didn't think we'd ever need to go the other way, and it also conserves memory and computation.

On the flip side, it means you'll have to test it by fiddling with a fits file or numpy array first, then passing it to `MaskModel()`.

It's possible we can change the behavior by making `dq` for `MaskModel` a  class property with a proper getter and setter, which would allow round-tripping from the internally-stored array of the reffile and the version that is presented in `MaskModel.dq`.

If this makes creation of new versions of this reffile easier for the instrument teams using our datamodels, then that would be a good motivation for that.

More next week.

mfixstsci:
Hey @jdavies-st! Thanks for getting around to this and am glad to see you were able to diagnose the issue! I like the idea of the `dq` array being an attribute of the `MaskModel` class that way you can access it when necessary. What do you think @cracraft?

cracraft:
I think that sounds like a great idea if it will allow us to access the dq_def that we need to properly test the mapping.

cracraft:
Hi. The workaround for this issue was the following: In my reference file creation script, set up the arrays and the dq_def table, but don't define the data model until back in the main script for execution.  Then once the model is created, it can be read into the do_dqinit function, though still not the full step.

Example code shown here:

def make_maskmodel(ysize, xsize):
    # create a mask model for the dq_init step
    csize = (ysize, xsize)
    dq = np.zeros(csize, dtype=int)
    # how do we define a dq_def extension?
    mask = MaskModel()

    dqdef = [(0, 1, 'DO_NOT_USE', 'Bad Pixel do not use'),
             (1, 2, 'DEAD', 'Dead Pixel'),
             (2, 4, 'HOT', 'Hot pixel'),
             (3, 8, 'UNRELIABLE_SLOPE', 'Large slope variance'),
             (4, 16, 'RC', 'RC pixel'),
             (5, 32, 'REFERENCE_PIXEL', 'Reference Pixel')]

    dq_def = np.array((dqdef), dtype=mask.dq_def.dtype)

    return dq, dq_def


def test_dq_im():
    # Check that PIXELDQ is initialized with the information from the reference file.
    # test that a flagged value in the reference file flags the PIXELDQ array
    # read in reference mask file

    # size of integration
    nints = 1
    ngroups = 5
    xsize = 1032
    ysize = 1024

    # create raw input data for step
    dm_ramp = make_rawramp(nints, ngroups, ysize, xsize)

    # create a MaskModel for the dq input mask

    dq, dq_def = make_maskmodel(ysize, xsize)

    # edit reference file with known bad pixel values

    dq[100, 100] = 2   # Dead pixel
    dq[200, 100] = 4   # Hot pixel
    dq[300, 100] = 8   # Unreliable_slope
    dq[400, 100] = 16  # RC
    dq[500, 100] = 1   # Do_not_use
    dq[100, 200] = 3   # Dead pixel + do not use
    dq[200, 200] = 5   # Hot pixel + do not use
    dq[300, 200] = 9   # Unreliable slope + do not use
    dq[400, 200] = 17  # RC + do not use

    # write mask model
    ref_data = MaskModel(dq=dq, dq_def=dq_def)
    ref_data.meta.instrument.name = 'MIRI'
    ref_data.meta.subarray.xstart = 1
    ref_data.meta.subarray.xsize = xsize
    ref_data.meta.subarray.ystart = 1
    ref_data.meta.subarray.ysize = ysize

    # run do_dqinit

    outfile = do_dqinit(dm_ramp, ref_data)
    print(outfile.err.shape)
    dqdata = outfile.pixeldq

    # assert that the pixels read back in match the mapping from ref data to science data
    assert(dqdata[100, 100] == dqflags.pixel['DEAD'])
    assert(dqdata[200, 100] == dqflags.pixel['HOT'])
    assert(dqdata[300, 100] == dqflags.pixel['UNRELIABLE_SLOPE'])
    assert(dqdata[400, 100] == dqflags.pixel['RC'])
    assert(dqdata[500, 100] == dqflags.pixel['DO_NOT_USE'])
    assert(dqdata[100, 200] == 1025)
    assert(dqdata[200, 200] == 2049)
    assert(dqdata[300, 200] == 16777217)
    assert(dqdata[400, 200] == 16385)

PR #2814, opened 2018-10-31
Remove soon-to-be-deprecated usage of AsdfFile.open

### LINKED ISSUE #2795: AsdfFile.open will be deprecated
`AsdfFile.open` will be deprecated in asdf v2.2.0 (no due date) spacetelescope/asdf#579. jwst will need the corresponding changes after asdf 2.2.0 is released.

drdavella:
This will take effect once https://github.com/spacetelescope/asdf/pull/579 is merged, but there's no harm in making this change now.

This closes #2795.

### DIFF
--- a/docs/jwst/assign_wcs/asdf-howto.rst
+++ b/docs/jwst/assign_wcs/asdf-howto.rst
@@ -117,17 +117,18 @@ Save a transform to an ASDF file
 `asdf <http://asdf.readthedocs.io/en/latest/>`__ is used to read and write reference files in
 `ASDF <http://asdf-standard.readthedocs.org/en/latest/>`__ format. Once the model is create using the rules in the above section, it needs to be assigned to the ASDF tree.
 
+>>> import asdf
 >>> from asdf import AsdfFile
 >>> f = AsdfFile()
 >>> f.tree['model'] = model
 >>> f.write_to('reffile.asdf')
 
-The ``write_to`` command validates the file and writes it to disk. It will catch any errors due to
-inconsistent inputs/outputs or invalid parameters.
+The ``write_to`` command validates the file and writes it to disk. It will
+catch any errors due to inconsistent inputs/outputs or invalid parameters.
 
-To test the file, it can be read in again using the ``AsdfFile.open()`` method:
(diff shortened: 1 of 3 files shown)


drdavella:
Seems like the Jenkins job is hung...

nden:
The `wcs` tags were removed from asdf as they are now (and have been for quite a while) in gwcs. Something must be using an old tag or it's a bug. I'll investigate.

drdavella:
I think the problem is files in `jwst/exp_to_source/tests/data` that contain references to `!wcs/...`. We just removed the old WCS tags and schemas from ASDF, so these tags won't resolve anymore.

I suspect that a simple search and replace of these for the corresponding tag name in `gwcs` should fix the problem. For example, `!wcs/step-1.0.0` should become `tag:stsci.edu:gwcs/step-1.0.0`.

drdavella:
@jdavies-st the short answer is yes. Of course, you could grab an older version of `asdf`, but that's not ideal, is it?

@nden we made this change expecting that there would be no residual uses of ASDF WCS, but maybe we should go with a deprecation period instead.

We're in a tricky place here because if we don't try to be diligent about cleaning these things up, ASDF is going to become a giant crapball of spaghetti code that mostly supports legacy uses. We should handle this better going forward, but I don't think we were expecting this particular change to have any  impact.

Unfortunately, making this change has exposed some other issues that will need to be addressed (see #2849).

Considering the hour, I think the best course of action is probably to sit down with a tall scotch and forget all about of this for a little while.

stscieisenhamer:
hold it...wait...i'm in the wrong conversations....

philhodge:
Is it (theoretically) possible to edit the ASDF extension to change the tags, as an alternative to rebuilding the extension?

drdavella:
@philhodge if I understand correctly, it's probably possible in theory, but I'm not sure we want to support this for the general case since it will probably be fairly difficult to maintain. It would probably be easier just to write a one-off script to convert files like these, but at that point it might just be easier to regenerate them.

drdavella:
Looks like we're good now!

jdavies-st:
Thanks @drdavella!

Issue #2815, opened 2018-10-31
Use tmpdir fixture in tests that create new files


drdavella:
Recently after running the pipeline tests locally, I noticed that there were two new files created:

```
jwst/associations/tests/science.fits
jwst/associations/tests/flat.fits
```

I'm not exactly sure where these originated from (I don't see either of these filenames anywhere in the tests in `jwst/associations/tests`), but we should be using the `tmpdir` fixture from `pytest` in order to make sure that we don't leave any test artifacts in user directories.

jdavies-st:
I've seen this same problem.  I'm not sure where they are being produced.  Possibly in

https://github.com/spacetelescope/jwst/blob/master/jwst/stpipe/tests/data/mkdata.py

This showed up after our doctest PR was merged.  I think this script needs to disappear, or it needs to be wrapped in a function.

It auto generates the two FITS files that are in that same directory.  But those FITS files are also in the repo.  A bunch of the tests in `jwst/stpipe/tests` rely on these FITS files.

jdavies-st:
And I have no idea why they show up in the `associations` subdir.

jdavies-st:
Yeah, these are definitely being produced when doctest parses that `mkdata.py` script.  If I run just the `associations` tests, no remnant files.  If I run just the `stpipe` tests, I get the remnant files, but in the top level directory, not in `associations`.

@hbushouse, do we need this script?  And if we do, can we wrap it in a function at least?  Or make it a pytest fixture?

drdavella:
If that's all it is, then that script should just be wrapped in an `if __name__ == '__main__'`.

hbushouse:
It sure looks like `mkdata.py` was kept there just to be able to regenerate the "science.fits" and "flat.fits" data files that are used in several of the unit tests. I don't see the harm in doing the necessary wrapping to fix it.

hbushouse:
On the other hand, if those 2 FITS files are also in the repo, then we don't need the script around anymore to regenerate them.

jdavies-st:
Turning it into a function works.  This will solve the immediate problem.

jdavies-st:
Resolved by #2818.

PR #2818, opened 2018-10-31
Make mkdata.py a function instead of a script

jdavies-st:
Addresses #2815.


jwst/stpipe/tests/data/mkdata.py
```diff
@@ -1,18 +1,18 @@
-#!/usr/bin/env python
-import numpy
-from astropy.io import fits as pyfits
+import numpy as np
+from astropy.io import fits as fits
```
hbushouse: Trivial nitpick, but why bother using an "as <name>" when <name> is the same as the package? Why not just leave it as "from astropy.io import fits"? The use in the subsequent code is identical.

jdavies-st: Doh!  I missed that.  Will change.

jdavies-st: Also taking Dan's suggestion and just wrapping the code in an `if` statement, so it can still be run as a script if needed.

Issue #2790, opened 2018-10-26
Incorrect Units for Read Noise Variance for Ramp Segments


stscijgbot:
Issue [JP-446](https://jira.stsci.edu/browse/JP-446) was created by Michael Regan:

Incorrect Units for Read Noise Variance for Ramp Segments

PR #2820, opened 2018-10-31
Fix for Issue #2756 and #2790 Wrong Units for Readnoise Variance

mwregan2:
This Pull request fixes issues #2756 and #2790. The correct fix is changes the units at the segment level. The first fix was incorrect when it changed the units at the integration level.
This also uploads the first version of the unit tests.


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -416,6 +416,7 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
             segs_4[num_int, :, rlo:rhi, :] = segs_beg_3
             var_p4[num_int, :, rlo:rhi, :] = den_p3 * med_rates[ rlo:rhi, :]
             var_r4[num_int, :, rlo:rhi, :] = num_r3 * den_r3
+            var_r4[num_int,:,rlo:rhi,:] = var_r4[num_int, :, rlo:rhi,:] /gain_2d[rlo:rhi,:]**2
```
philhodge: This is fine, but you could use gain_sect**2 instead of gain_2d[rlo:rhi,:]**2.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -464,7 +465,7 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
         del var_p4_int
         del var_p4_int2
 
-    del gain_2d
+        del gain_2d
```
philhodge: This won't work.  By shifting this to the right, you have put it inside the loop over integrations, but it was assigned outside the loop over integrations.

jwst/ramp_fitting/ramp_fit.py
hbushouse: Minor, but for the sake of not inflating the code base, couldn't the division by gain be incorporated into the computation of `var_r4` in line 418?

mwregan2: I have committed a change with to do the unit conversion on one line and add a comment.

jwst/ramp_fitting/ramp_fit.py
mwregan2: I changed it back to the original indentation.

hbushouse:
The CI problems are due to build issues with astropy needing a version of numpy that appears to be incompatible with something else (like sphinx or conda). It started about 24 hours ago with an astropy update. "Top men" are working on it ...

jwst/ramp_fitting/tests/test_ramp_fit.py
```diff
@@ -0,0 +1,629 @@
...
+from jwst.datamodels import MIRIRampModel
+from jwst.datamodels import GainModel, ReadnoiseModel
+
+
+#@pytest.mark.skip(reason="not using now")
+def test_nocrs_noflux():
+        #all pixel values are zero. So slope should be zero
+    model1, gdq, rnModel, pixdq, err, gain = setup_inputs(ngroups=5)
+    slopes = ramp_fit(model1, 60000, False, rnModel, gain, 'OLS', 'optimal')
+    print(slopes[0].data)
```
jdavies-st: I would suggest removing the `print()` statements in these tests, as they are not necessary for the tests to run as CI unit tests.  It looks like they were there for debugging purposes in setting up your `assert` statements?  I think they can be removed now.

jwst/ramp_fitting/gls_fit.py
```diff
@@ -576,7 +576,7 @@ def gls_fit(ramp_data, input_var_data,
         pixel, in electrons per second.
 
     readnoise: 1-D ndarray, length nz.
-        The read noise in electrons at each detector pixel.
+        The read noise in electrons at each deector pixel.
```
hbushouse: Was the creation of a typo intentional here?

mwregan2: I removed the typo

jwst/ramp_fitting/tests/test_ramp_fit.py
mwregan2: I have removed the print and pytest.mark.skip lines.

mwregan2:
I think I have resolved all of the comments and the merge conflict. Let me know if there is something I still need to do.

jdavies-st:
Thanks @mwregan2!  Looks really good from my end.

mwregan2:
I cleaned up the extraneous white space issues. I think this is now ready.

PR #2822, opened 2018-10-31
Updated doc strings for persistence.  See issue #2681.

philhodge:


jwst/ipc/x_irs2.py
```diff
@@ -104,8 +126,21 @@ def is_irs2(input_model):
...
+    ----------
+    input_model : JWST data model, or an ndarray
+        Either the input science data model or the data array from the
+        input data model.
+
+    Returns
+    -------
+    tuple of int
+        The shape of the input science data array.
+    """
```
stscirij: Should probably document n and r here

PR #2829, opened 2018-11-01
Fix tweakreg spec issues

### LINKED ISSUE #2828: Tweakreg `spec` problems
When I try to run `TweakRegStep` from the default config file via `strun tweakreg.cfg`, it fails configspec checking of the spec:
```
E               jwst.stpipe.config_parser.ValidationError: Config parameter 'steps/tweakreg/brightest': the check "int" is unknown.
E               Config parameter 'steps/tweakreg/peakmax': the value "None" is of the wrong type.
```
The problems are
```
        brightest = int(default=None) # Keep top ``brightest`` objects
        peakmax = float(default=None) # Filter out objects with pixel values >= ``peakmax``
```
`int` is not valid.  It should be `integer`.

But for both of these, `default=None` is a special case in `ConfigObj`. This is a known limitation that I believe @hbushouse ran into before.  Basically you can have that there, but you cannot set an actual config file to have
```
peakmax = None
```
Because that's a string that does not pass validation (`peakmax` needs `float`).  For anything in `spec` that has a default of `None`, one must leave it out in the `.cfg` file, unless you are setting it to a non-default value, in which it must be `integer` or `float` as above.

Read more here
https://configobj.readthedocs.io/en/latest/configobj.html#mentioning-default-values

and here
https://github.com/DiffSK/configobj/issues/111

jdavies-st:
This fixes a bug in `0.12.1` release.  Not sure if it is needed as a `0.12.2` patch because it is for level 3 processing.

This wasn't picked up in our regression tests because we don't run them via a `.cfg` file.  I've modified two of our regression tests to instead use `.cfg` files to run the pipeline so we can pick this up.  Since DMS/SDP uses `.cfg` files to run our pipeline, that's a good reason to make sure we're doing the same in regression tests.

Resolves #2828.


jwst/pipeline/tweakreg.cfg
```diff
@@ -5,8 +5,6 @@ save_catalogs = False
 catalog_format = 'ecsv'
 kernel_fwhm = 2.5
 snr_threshold = 10.
-brightest = None
```
mcara: @jdavies-st Make this a 100 default... This should keep top 100 sources for alignment and 100 should be enough to get a good alignment. This will help eliminate low SNR or fake sources and thus make the algorithm more robust against low SNR detection thresholds.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -30,7 +30,7 @@ class TweakRegStep(Step):
         catalog_format = string(default='ecsv') # Catalog output file format
         kernel_fwhm = float(default=2.5) # Gaussian kernel FWHM in pixels
         snr_threshold = float(default=10.0) # SNR threshold above the bkg
-        brightest = int(default=None) # Keep top ``brightest`` objects
+        brightest = integer(default=None) # Keep top ``brightest`` objects
```
mcara: Same here: `default=100`

jdavies-st:
CI failures unrelated.  Tests pass locally. Merging.

nden:
Moving this to `0.13`.

PR #2830, opened 2018-11-01
Don't use numpy 1.15.3 in testing

jdavies-st:
It seems masked arrays are broken in some way in `numpy 1.15.3`.  Seeing if this will fix our CI tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -48,7 +48,7 @@ def pip_packages_tests = "requests_mock ci_watson"
 dist = new BuildConfig()
 dist.nodetype = 'linux'
 dist.name = 'dist'
-dist.conda_packages = ["numpy=${matrix_numpy[0]}"]
+dist.conda_packages = ['numpy=${matrix_numpy[0]}']
 dist.build_cmds = [
     "python setup.py sdist",
     "python setup.py bdist_egg",
@@ -62,7 +62,7 @@ docs = new BuildConfig()
 docs.nodetype = 'linux'
 docs.name = 'docs'
 docs.conda_channels = ['http://ssb.stsci.edu/astroconda-dev']
-docs.conda_packages = conda_packages + conda_packages_docs + ["numpy=${matrix_numpy[0]}"]
+docs.conda_packages = conda_packages + conda_packages_docs + ['numpy=${matrix_numpy[0]}']
 docs.build_cmds = [
     "pip install -q ${pip_packages_docs}",
     "python setup.py build_sphinx"
@@ -80,7 +80,7 @@ for (python_ver in matrix_python) {
             bc.env_vars = test_env
             bc.name = name
             bc.conda_channels = ['http://ssb.stsci.edu/astroconda-dev']
-            bc.conda_packages = conda_packages + ["python=${python_ver}"] + ["numpy=${numpy_ver}"]
+            bc.conda_packages = conda_packages + ['python=${python_ver}'] + ['numpy=${numpy_ver}']
             bc.build_cmds = [
                 "pip install -q ${pip_packages_tests}",
                 "python setup.py install"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 if (utils.scm_checkout()) return
 
 matrix_python = ['3.6']
-matrix_numpy = ['1.14']
+matrix_numpy = ['1.15.2']
 matrix_astropy = ['4']
 matrix = []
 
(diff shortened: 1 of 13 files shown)


nden:
Hm, so this worked for Travis but not Jenkins?

Jenkinsfile
```diff
@@ -72,7 +72,7 @@ dist = new BuildConfig()
 dist.nodetype = 'linux'
 dist.name = 'dist'
 dist.build_cmds = [
-    "${CONDA_INST} numpy",
+    "${CONDA_INST} numpy=1.15.2",
```
nden: Is it necessary to specify a version here given that it is specified in the requirements above?
Also, isn't numpy pinned to 1.14 at the beginning of the file?
I don't understand enough of this, perhaps @mrendina can help.

jdavies-st: This is for the building of just the `jwst` package, which is the only thing done here.  It's actually probably not necessary to specify numpy version here, as I doubt `1.15.3`  breaks the build process.

nden:
Is Travis showing a real error due to a recent merge?

jdavies-st:
Ok, well this seems to have fixed the build issues.  But there's actually a failure in our tests now:
```
=================================== FAILURES ===================================
____________________________ test_suffix_existence _____________________________
    def test_suffix_existence():
        """Generate current suffix list and compare"""
        calculated_suffixes = s.find_suffixes()
        found_suffixes = s.combine_suffixes(
            to_add=(calculated_suffixes, s.SUFFIXES_TO_ADD)
            to_remove=(s.SUFFIXES_TO_DISCARD, )
        )
>       assert set(found_suffixes) == set(s.KNOW_SUFFIXES)
E       AssertionError: assert {'alignrefs',...average', ...} == {'alignrefs', ...average', ...}
E         Extra items in the left set:
E         'tweakreg'
../../lib/tests/test_suffix.py:15: AssertionError
```
Any ideas @stscieisenhamer ?

stscieisenhamer:
hmmm, thought that had been fixed, but apparently #2829 re-broke. Will PR that shortly. Or if you want to fix in this PR, add `tweakreg` to `_calculated_suffixes` found [here](https://github.com/spacetelescope/jwst/blob/2e0d07e311b10f3ac6d04cbdfe8f202644685b66/jwst/lib/suffix.py#L63).

nden:
@jdavies-st Let's merge a version of this PR with the change to Travis only, so that we can check tests.

jdavies-st:
I'll fix the `tweakreg` suffix in this PR @stscieisenhamer.

One more commit I hope.

rendinam:
It looks like the conda version in use in Travis is 4.4.10, whereas the most recent conda installer provides 4.5.11. Travis is obviously able to install numpy when the version clamps are applied as shown in this PR, but the most recent conda (4.5.11) stalls forever on both OSX and Linux when installing the package version specification `numpy>=1.13,!=1.15.3` in an otherwise bare python 3.5 or python 3.6 environment.
There appear to be recent complaints of similar behavior elsewhere: https://github.com/conda/conda/issues/5219
An explicit single version specification (=1.15.2) is honored, so that's probably the best bet for now.

stscicrawford:
As you included the tweakreg fix in this, can you also include the flake8 fix?

nden:
🎉 Tests passing, except for a flake8 issue

jdavies-st:
@stscieisenhamer, I see that the #2706 CI test failure of `test_suffix_existence` was a bit of a canary for that PR.  That PR broke instantiating `tweakreg`, which was caught by the test:

https://travis-ci.org/spacetelescope/jwst/jobs/442822570#L1541

We should keep this in mind when the suffix test fails again.

jdavies-st:
Since github is having issues right now,

https://status.github.com/messages

what were once passing builds are no longer passing builds on Jenkins.  I.e. when Jenkins requests this branch from Github, it's sending out an outdated one.

Everything is on fire this week.  I'm going to go take a nap and see if this sorts itself out.

FWIW, the tests originally passed on Jenkins.  Now they are not.  No changes.  I'll submit one more commit after Github sorts out its issues, and that should build successfully on Jenkins.

jdavies-st:
Don't merge this yet.  I would like to switch the Jenkins CI builds back from numpy 1.14 to numpy 1.15.2.  But I don't want to do that until Github gets caught up on its backlog.  Once it does, I will push that change, and then the Jenkins build should pass.  Currrently, no Jenkins builds are going pass regardless of what we do, as long as Github has a backlog.

Jenkinsfile
```diff
@@ -1,7 +1,7 @@
 if (utils.scm_checkout()) return
 
 matrix_python = ['3.6']
-matrix_numpy = ['1.15.2']
+matrix_numpy = ['1.14']
```
jdavies-st: If you do want to merge this PR, make one more change after the GIthub issues blow over.  Change this back to `1.15.2`.  The tests should pass, and we can all have a nice weekend.

nden:
🎉 🍺

Issue #2817, opened 2018-10-31
IndexError: Compound model CompoundModel56 does not have a component named crval


stscieisenhamer:
# Issue
Running on WFR2 data on a fine phasing proposal, jw00711, using `wfs_combine` on a pair, gives the following:

<details><summary>Show log...</summary>
<p>

```
$ strun cfgs/wfs_combine.cfg jw00711_20181027T043250_pool-o002_wfs-image3_001_asn.json
2018-10-31 11:51:28,011 - stpipe.WfsCombine - INFO - WfsCombineStep instance created.
2018-10-31 11:51:28,038 - stpipe.WfsCombine - INFO - Step WfsCombine running with args ('jw00711_20181027T043250_pool-o002_wfs-image3_001_asn.json',).
2018-10-31 11:51:28,060 - stpipe.WfsCombine - INFO - Using input table: jw00711_20181027T043250_pool-o002_wfs-image3_001_asn.json
2018-10-31 11:51:28,061 - stpipe.WfsCombine - INFO - The number of pairs of input files: 1
2018-10-31 11:51:40,669 - stpipe.WfsCombine - INFO - File 1 to combine: jw00711002001_02105_00002_nrca4_cal.fits
2018-10-31 11:51:40,669 - stpipe.WfsCombine - INFO - File 2 to combine: jw00711002001_02105_00001_nrca4_cal.fits
2018-10-31 11:51:40,669 - stpipe.WfsCombine - INFO - Output file: jw00711-o002_t001_nircam_wlp4-nrca4_{suffix}-05
2018-10-31 11:51:40,669 - stpipe.WfsCombine - INFO - do_refine: False
----------------------------------------------------------------------
ERROR RUNNING STEP 'WfsCombineStep':
    Compound model CompoundModel56 does not have a component named
    crval
Traceback (most recent call last):
  File "/Users/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/astropy-3.2.dev23179-py3.6-macosx-10.7-x86_64.egg/astropy/modeling/core.py", line 2705, in get_index_from_name
    return cls.submodel_names.index(name)
ValueError: tuple.index(x): x not in tuple

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #2831, opened 2018-11-01
Update WFS&C associations and rename wfs_combine to match workflow convention

### LINKED ISSUE #2672: Update WFS&C rules with actual example pools
# Issue
With the Wavefront Rehearsals,, there are solid examples now of coarse phasing. Update to the `Asn_WFSCMB` rule with how the dither pairs are actually delineated.

# Resources
- Proposal jw00747
- Proposal jw00711
- WFR 2
- #1340
- #1615

stscieisenhamer:
Resolves #2672 

Note to reviewers: code files are:
- jwst/associations/lib/rules_level3.py
- jwst/associations/lib/rules_level3_base.py
- jwst/pipeline/calwebb_wfs-image3.cfg
- jwst/stpipe/step.py
- jwst/wfs_combine/wfs_combine_step.py


dmggh:
Does this have any bearing on the WCS issue in #2817 ?

jwst/wfs_combine/wfs_combine_step.py
```diff
@@ -37,7 +37,7 @@ def process(self, input_table):
             output_model = wfs.do_all()
             output_model.meta.cal_step.wfs_combine = 'COMPLETE'
             self.save_model(
-                output_model, suffix='wfscmb', output_file=outfile
+                output_model, suffix='wfscmb', output_file=outfile, format=False
```
hbushouse: Does the step module need to be explicitly saving the output model, as opposed to just returning the model and letting stpipe handle it?

stscieisenhamer:
Woah, some type of nasty build issue failure here. @rendinam

nden:
@stscieisenhamer This is a known issue. Numpy 1.5.3 has a bug in masked arrays and astropy is pinned not to use it. Unfortunately it is the default version in conda now. Numpy 1.5.4 will soon become available in conda and this will be solved in a natural way. #2830 attempts to fix the np version in testing but for some reason only works in Travis.

nden:

PR #2834, opened 2018-11-02
Fixed bug in which the number of segments returned exceeds the number…

dmggh:
… of groups, which had occurred for a MIRI dataset in which the first or last group was flagged as  DO_NOT_USE and also flagged as a jump. This addresses failure discovered during MIRI LEVEL_2A testing under v0.12.1 for imaging & IFUs.


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -2225,8 +2225,8 @@ def calc_num_seg(gdq, n_int):
 
     Return:
     -------
-    int(max_cr) +1 : int
-        maxmimum number of segments; n CRS implies n+1 segments
+    max_num_seg : int
+        maxmimum number of segments which can not exceed thenumber of groups
```
philhodge: Typo:  "thenumber" --> "the number".

stscicrawford:
Thanks @dmggh.  This will probably have to wait on the tests being fixed and re-run once they are.  

Can you add one of the datasets that this was failing on to the regression tests?

Can you also add this to the 0.12.1 section to the change log along with the details from #2760?

dmggh:
@stscicrawford : Yes, I will add an example of such a (formerly) failing test to the regression tests, and I will update the changelog.

stscicrawford:
Failure was a random conda failure on building docs.   Merging.

nden:

PR #2840, opened 2018-11-05
Use pytest-xdist to parallelize our ssbjenkins CI builds

jdavies-st:
Trying out this test 🎈


jdavies-st:
So using `pytest-xdist` with `n=4` seems to allow the unit tests to run in 319 sec (vs 684 sec without).

Of course the reduction in overall build time is not as dramatic, as the environment build takes the same amount of time.  Total time goes from 18 min down to 12 min.

jdavies-st:
Well, I think given the load on `ssbjenkins` currently, this PR is probably premature.  Closing.

PR #2841, opened 2018-11-05
Updated to fix IndexError: Compound model CompoundModel56 does not ha…

dmggh:
…ve a component named crval (Issue 2817)

### DIFF
--- a/jwst/wfs_combine/wfs_combine.py
+++ b/jwst/wfs_combine/wfs_combine.py
@@ -261,7 +261,7 @@ class DataSet():
         xcen = int(self.input_1.data.shape[1]/2)
         ycen = int(self.input_1.data.shape[0]/2)
 
-        radec = tr1(ycen, xcen)
+        radec = tr1(xcen, ycen)
         pixels = tr2(radec[0], radec[1])
         off_x = pixels[0] - xcen
         off_y = pixels[1] - ycen
--- a/jwst/wfs_combine/wfs_combine.py
+++ b/jwst/wfs_combine/wfs_combine.py
@@ -252,31 +252,19 @@ class DataSet():
         off_y: integer
             difference (#2 -#1) in pointing in pixels in the y-direction
         """
+        wcs1 = self.input_1.meta.wcs
+        wcs2 = self.input_2.meta.wcs
+        tr1 = wcs1.get_transform('detector', 'world')
+        tr2 = wcs2.get_transform('world', 'detector')
+ 
+        # Get coords of center pixel
+        xcen = int(self.input_1.data.shape[1]/2)
+        ycen = int(self.input_1.data.shape[0]/2)
 
-        #crval1 = self.input_1.meta.wcs.crval
-        #crval2 = self.input_2.meta.wcs.crval
-        #crpix2_in_1 = self.input_1.meta.wcs.invert(*crval2)
-        #crpix = self.input_1.meta.wcs.crpix
-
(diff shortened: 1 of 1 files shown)


jwst/wfs_combine/wfs_combine.py
```diff
@@ -252,31 +252,19 @@ def get_wcs_offsets(self):
...
+        wcs1 = self.input_1.meta.wcs
+        wcs2 = self.input_2.meta.wcs
+        tr1 = wcs1.get_transform('detector', 'world')
+        tr2 = wcs2.get_transform('world', 'detector')
+ 
+        # Get coords of center pixel
+        xcen = int(self.input_1.data.shape[1]/2)
+        ycen = int(self.input_1.data.shape[0]/2)
+
+        radec = tr1(ycen, xcen)
```
philhodge: The arguments are reversed, i.e. it should be:

    radec = tr1(xcen, ycen)

jwst/wfs_combine/wfs_combine.py
```diff
@@ -252,31 +252,19 @@ def get_wcs_offsets(self):
...
-
-        log.info('CRVAL of Image 1 in WCS of #1 has pixel coords:%s %s',
-                 crval1[0], crval1[1])
-        log.info('CRVAL of Image 2 in WCS of #2 has pixel coords:%s %s',
-                 crval2[0], crval2[1])
-        log.info('Image 2 has these nominal offsets relative to image 1\
-                 (x,y): %s %s', off_x, off_y)
+        wcs1 = self.input_1.meta.wcs
+        wcs2 = self.input_2.meta.wcs
+        tr1 = wcs1.get_transform('detector', 'world')
```
nden: This is equivalent to calling just calling `wcs` as a function, e.g. `wcs(xcen, ycen)`.

jwst/wfs_combine/wfs_combine.py
```diff
@@ -252,31 +252,19 @@ def get_wcs_offsets(self):
...
+        tr2 = wcs2.get_transform('world', 'detector')
```
nden: This is equivalent to calling `wcs2.invert(ra, dec)`.

nden:
To clarify my clarification :wink: - just calling the WCS is a bit more general because it does not rely on coordinate frames `world` and `detector` being there.

jwst/wfs_combine/wfs_combine.py
philhodge: That's what I thought.  I suggested explicitly specifying frames because when I tried just calling wcs I saw something that implied it was mapping to 'v2v3' instead of to 'world' (it was a warning message when I called wcs with too few arguments).  If I understand correctly, we really need the mapping from 'detector' to 'world', though perhaps it would work to map between 'v2v3' and 'world' (but definitely not from 'detector' to 'v2v3').

nden: That makes sense. If `world` is not in the WCS, then it will fail. Perhaps this is what needs to happen.

stscieisenhamer: FYI: This is what happens with the outdated regression test data. 👍

nden: There was no pointing information at the time. I think this kind of data will fail in the assign_wcs step if it doesn't have the pointing info. The problem with the test is it had a stale WCS from years ago.

nden:

PR #2849, opened 2018-11-08
Replace tag references to ASDF WCS with GWCS in test data files

drdavella:
I believe this is necessary, but it does not appear to be sufficient, to get tests to pass. More investigation is required.

This appears to be due to the removal of WCS tags from ASDF in https://github.com/spacetelescope/asdf/pull/593, but we should have been using GWCS instead of ASDF for WCS tags for a while now, so this seems like an oversight.

### DIFF
--- a/jwst/exp_to_source/tests/data/jwst_nod1_cal.fits
+++ b/jwst/exp_to_source/tests/data/jwst_nod1_cal.fits
@@ -130,11 +130,11 @@ slits:
   meta:
     date: '2017-04-04T12:28:33.674'
     model_type: ImageModel
-    wcs: !wcs/wcs-1.0.0
+    wcs: !<tag:stsci.edu:gwcs/wcs-1.0.0>
       name: ''
       steps:
-      - !wcs/step-1.0.0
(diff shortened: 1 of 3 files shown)


jdavies-st:
Yeah, these files were probably generated over 2 years ago and just need a GWCS object update.

These appear to be `MultiSlitModel` FITS files.  I suspect it would b easiest to regenerate them from the `_rate.fits` files by just running them through the `spec2` pipeline.  @stscieisenhamer, any idea where the `_rate.fits` files for these are?  There's a jupyter notebook in those directories that take the original 2k^2 files and downsize them to 10x10, but we really need the original rate file to update the ASDF.

drdavella:
If that's all it is, then I'm not going to worry about reverting the change we made to ASDF. We'll just consider it an oversight.

In any case, I'm a bit confused about these failures, but it's probably not worth tracking down. It will be interesting to see the diff when the new files are available.

drdavella:
Closing since it sounds like the right solution is just to regenerate the data files.

PR #2850, opened 2018-11-08
Fix test_exp_to_source test data

jdavies-st:
WCS tags were removed from ASDF in https://github.com/spacetelescope/asdf/pull/593, and they have lived in the `gwcs` package now for a while.  Now they need to be referenced as such.

Test input files here are regenerated using a recent version of `gwcs` and `asdf`.  Now the tags point to GWCS instead of ASDF WCS.

This should fix the current failures in our CI testing.


jdavies-st:
This is another crack at #2849.

jwst/exp_to_source/tests/data/jwst_nod1_cal.fits
```diff
@@ -199,29 +183,65 @@ slits:
                       forward:
                       - !transform/polynomial-1.1.0
                         coefficients: !core/ndarray-1.0.0
-                          source: 1
+                          data:
```
drdavella: @nden this and similar below makes it seem like we might have missed an `inline_threshold` somewhere? I can look into it.

nden: Hm, yes.

PR #2851, opened 2018-11-08
Support asdf flags when creating a new datamodel

### LINKED ISSUE #1215: silence asdf warnings

bernie-simon:
Asdf supports two flags, ignore_version_mismatch and gnore_unrecognized_flag, when creating a new asdf file. Part of creating a new datamodel is opening an asdf file, either natively or as embedded in a FITS extension. This update passes these two flags from the DataModel argument list to the call that opens or creates an asdf file. The result is better control of the warning messages produced by asdf.

Resolves #1215.

### DIFF
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -82,7 +82,10 @@ class ModelContainer(model_base.DataModel):
             instance = copy.deepcopy(init._instance)
             self._schema = init._schema
             self._shape = init._shape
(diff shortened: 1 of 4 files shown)


hbushouse:
Needs a change log entry.

PR #2852, opened 2018-11-09
Define pytest_plugins in top level conftest.py

jdavies-st:
Having pytest plugins anywhere except the top level `conftest.py` has been deprecated since `pytest 3.5`.  We've been seeing this warning for a while:
```
conftest.py:0: RemovedInPytest4Warning: Defining pytest_plugins in a non-top-level conftest is deprecated, because it affects the entire directory tree in a non-explicit way.
Please move it to the top level conftest file instead.
```
This PR resolves it.

We currently use the `asdf` schema tester pytest plugin to make sure our datamodel schemas are valid YAML.

I'm also adding the `scripts/` subdir to our pytest ignore directory, as in order to parse the `set_bary_helio_times.py` script, the environment variable JPL_EPHEMERIS needs to be set, and one gets a pytest collection error when one doesn't.  We started seeing this collection error when we started doctesting.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,8 @@
+"""Project default for pytest"""
+from astropy.tests.helper import enable_deprecations_as_exceptions
+
+enable_deprecations_as_exceptions()
+
+pytest_plugins = [
+    'asdf.tests.schema_tester'
+]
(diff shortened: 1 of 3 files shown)


jdavies-st:
This needs to be reviewed and merged before we can unpin `pytest`, i.e. before #2853.

PR #2856, opened 2018-11-09
Associations docstring/document update & cleanup

stscieisenhamer:
WIP #2681

PR #2862, opened 2018-11-14
Update of association docstring-part deux

stscieisenhamer:
WIP #2681 plus types found in review of #2856 but not applied there.


jwst/associations/lib/rules_level3.py
```diff
@@ -65,11 +73,16 @@ def _init_hook(self, item):
 
 @RegistryMarker.rule
 class Asn_WFSCMB(AsnMixin_Science):
-    """Wavefront Sensing association
+    """Level 3 Wavefront Control & Sensing Association
 
     For coarse and fine phasing, dither pairs need
```
philhodge: "need to be associated"?

jwst/associations/lib/acid.py
```diff
@@ -17,7 +17,7 @@ class ACID():
 
     Parameters
     ----------
-    in: str or 2-tuple
+    in : str or 2-tuple
```
philhodge: This is probably OK, since it's just documentation, but `in` is a reserved word in Python.

jwst/associations/registry.py
```diff
@@ -139,28 +139,28 @@ def match(self, item, version_id=None, allow=None, ignore=None):
 
         Parameters
         ----------
-        item: dict
+        item : dict
             A item, like from a Pool, to find assocations for.
```
philhodge: typo:  "A item" --> "An item"

jwst/associations/lib/acid.py
stscieisenhamer: That should have been `input`. Fixed.

stscieisenhamer:
system-level travis failure. mergine.

PR #2863, opened 2018-11-14
Add test to open a read-only-permissions file in datamodels

jdavies-st:
Make sure datamodels can open read-only files, specifically reffiles from CRDS, which can be from a read-only cache.


jdavies-st:
Superceded by #2897.

PR #2864, opened 2018-11-14
Add tests for dq_init step

### LINKED ISSUE #2561: MIRI team to review and agree with unit-testing plan.
Issue [JP-378](https://jira.stsci.edu/browse/JP-378) was created by Macarena Garcia Marin:

MIRI team to review and agree with unit-testing plan.

cracraft:
Adding unit tests for dq_init step. Created by @cracraft and @shanosborne.

Closes #2807.

Resolves #2561

### DIFF
--- a/jwst/dq_init/tests/test_dq_init.py
+++ b/jwst/dq_init/tests/test_dq_init.py
@@ -1,12 +1,8 @@
 import pytest
-from astropy.io import fits
 from jwst.dq_init import DQInitStep
 from jwst.dq_init.dq_initialization import do_dqinit
-from jwst.datamodels import MIRIRampModel, MaskModel
-from jwst.datamodels import GuiderRawModel, RampModel
-from jwst.datamodels import ImageModel, dqflags
+from jwst.datamodels import MIRIRampModel, MaskModel, GuiderRawModel, RampModel, dqflags
 import numpy as np
-import numpy.testing as nptest
 
 
 # Set parameters for multiple runs of data
@@ -20,11 +16,11 @@ ids = ["GuiderRawModel-Stack", "GuiderRawModel-Image", "RampModel", "MIRIRampMod
 
 @pytest.mark.parametrize(args, test_data, ids=ids)
 def test_dq_im(xstart, ystart, xsize, ysize, nints, ngroups, instrument, exp_type):
-    # Check that PIXELDQ is initialized with the information from the reference file.
-    # test that a flagged value in the reference file flags the PIXELDQ array
+    """ Check that PIXELDQ is initialized with the information from the reference file.
+    test that a flagged value in the reference file flags the PIXELDQ array"""
 
     # create raw input data for step
-    dm_ramp = make_rawramp(instrument, nints, ngroups, ysize, xsize, xstart, ystart, exp_type)
+    dm_ramp = make_rawramp(instrument, nints, ngroups, ysize, xsize, ystart, xstart, exp_type)
 
     # create a MaskModel for the dq input mask
     dq, dq_def = make_maskmodel(ysize, xsize)
@@ -50,7 +46,6 @@ def test_dq_im(xstart, ystart, xsize, ysize, nints, ngroups, instrument, exp_typ
 
     # run do_dqinit
     outfile = do_dqinit(dm_ramp, ref_data)
-    print(outfile.err.shape)
 
     if instrument == "FGS":
         dqdata = outfile.dq
@@ -70,7 +65,8 @@ def test_dq_im(xstart, ystart, xsize, ysize, nints, ngroups, instrument, exp_typ
 
 
 def test_groupdq():
-    # Check that GROUPDQ extension is added to the data and all values are initialized to zero.
+    """Check that GROUPDQ extension is added to the data and all values are initialized to zero."""
+
     # size of integration
     instrument = 'MIRI'
     nints = 1
@@ -81,11 +77,11 @@ def test_groupdq():
     ystart = 1
 
     # create raw input data for step
-    dm_ramp = make_rawramp(instrument, nints, ngroups, ysize, xsize, xstart, ystart)
+    dm_ramp = make_rawramp(instrument, nints, ngroups, ysize, xsize, ystart, xstart)
 
     # create a MaskModel for the dq input mask
-
     dq, dq_def = make_maskmodel(ysize, xsize)
+
     # write mask model
     ref_data = MaskModel(dq=dq, dq_def=dq_def)
     ref_data.meta.instrument.name = instrument
@@ -108,7 +104,8 @@ def test_groupdq():
 
 
 def test_err():
-    # Check that a 4-D ERR array is initialized and all values are zero.
+    """Check that a 4-D ERR array is initialized and all values are zero."""
+
     # size of integration
     instrument = 'MIRI'
     nints = 1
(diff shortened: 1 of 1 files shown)


jwst/dq_init/tests/test_dq_init.py
```diff
@@ -0,0 +1,340 @@
...
+test_data = [(1, 1, 2304, 2048, 2, 2, 'FGS', 'FGS_ID-STACK'),
+             (1, 1, 2048, 2048, 2, 2, 'FGS', 'FGS_ID-IMAGE'),
+             (1, 1, 2048, 2048, 2, 2, 'NIRCAM', 'NRC_IMAGE'),
+             (1, 1, 1032, 1024, 1, 5, 'MIRI', 'MIR_IMAGE')]
+ids = ["GuiderRawModel-Stack", "GuiderRawModel-Image", "RampModel", "MIRIRampModel"]
+
+
+@pytest.mark.parametrize(args, test_data, ids=ids)
+def test_dq_im(xstart, ystart, xsize, ysize, nints, ngroups, instrument, exp_type):
+    # Check that PIXELDQ is initialized with the information from the reference file.
```
jdavies-st: Could you convert these to docstrings for the tests.  Something like
```
"""Check that flagged value in the reference file flags the DQ array in science image"""
```
or similar.  A one-liner is good.  But the triple-quotes are preferred `#`.

jwst/dq_init/tests/test_dq_init.py
```diff
@@ -0,0 +1,340 @@
+import pytest
+from astropy.io import fits
+from jwst.dq_init import DQInitStep
+from jwst.dq_init.dq_initialization import do_dqinit
+from jwst.datamodels import MIRIRampModel, MaskModel
+from jwst.datamodels import GuiderRawModel, RampModel
+from jwst.datamodels import ImageModel, dqflags
+import numpy as np
+import numpy.testing as nptest
```
hbushouse: flake8 says that `nptest` is also unused and hence can be removed.

jwst/dq_init/tests/test_dq_init.py
```diff
@@ -0,0 +1,340 @@
...
+    outfile = DQInitStep.call(dm_ramp)
+
+    # test that a pixeldq frame has been initlialized
+    if instrument == "FGS":
+        assert outfile.dq.ndim == 2
+    else:
+        assert outfile.pixeldq.ndim == 2  # a 2-d pixeldq frame exists
+
+
+def make_rawramp(instrument, nints, ngroups, ysize, xsize, xstart, ystart, exp_type=None):
```
hbushouse: Seems a little confusing to me to have the arguments in the order of "ysize, xsize" and then "xstart, ystart", but as long as you know that and populate them consistently within the tests, it'll work OK.

jdavies-st:
Other than the comments above, the tests look good!  Thanks.  @hbushouse, could you have a look too?

Also, how do we want to milestone these unit tests from the instrument teams?

jwst/dq_init/tests/test_dq_init.py
```diff
@@ -0,0 +1,340 @@
...
+    ref_data = MaskModel(dq=dq, dq_def=dq_def)
+    ref_data.meta.instrument.name = instrument
+    ref_data.meta.subarray.xstart = xstart
+    ref_data.meta.subarray.xsize = xsize
+    ref_data.meta.subarray.ystart = ystart
+    ref_data.meta.subarray.ysize = ysize
+
+    # run do_dqinit
+    outfile = do_dqinit(dm_ramp, ref_data)
+    print(outfile.err.shape)
```
hbushouse: Probably want to remove the print statements for use in the CI environment.

jwst/dq_init/tests/test_dq_init.py
```diff
@@ -0,0 +1,340 @@
...
+    groupdq = outfile.groupdq
+
+    np.testing.assert_array_equal(np.full((1, ngroups, ysize, xsize),
+                                          0,
+                                          dtype=int),
+                                  groupdq,
+                                  err_msg='groupdq not initialized to zero')
+
+
+def test_err():
```
hbushouse: I'm not 100% sure this test will succeed, because it's using data models as input, rather than disk files. The `dq_init` step checks for the existence of `pixeldq` and `groupdq` arrays and will create them if they aren't in the data model, but it does not do that for the `err` array. The only place/way the `err` will be initialized if it doesn't exist in the input is when a disk file is loaded into a data model. The check is done at the time the data model is created. So I'm not sure this will work if you're starting with an existing data model (but I could be wrong).

hbushouse:
@jdavies-st Milestones can just be set to the next planned release, which as of right now is 0.13.0.

cracraft:
Is there a way for me to add people to watch this request like I can add watchers in JIRA?

hbushouse:
@cracraft The most obvious way I know of to add "watchers" is to simply list their github names in a comment like this (i.e. @name).

cracraft:
@mfixstsci @shanosborne @aliciacanipe  So you can see what's going on with this request.

jwst/dq_init/tests/test_dq_init.py
cracraft: Do you want just the description comments as triple quotes, or all comments?

jwst/dq_init/tests/test_dq_init.py
cracraft: That came about because of two programmers working on the same code.

jwst/dq_init/tests/test_dq_init.py
cracraft: This test completes fine on my computer. It does create a 4D ERR array, though with dimensions of [0, 0, 0, 0] rather than matching the dimensions of the data.

jwst/dq_init/tests/test_dq_init.py
```diff
@@ -0,0 +1,340 @@
+import pytest
+from astropy.io import fits
+from jwst.dq_init import DQInitStep
+from jwst.dq_init.dq_initialization import do_dqinit
+from jwst.datamodels import MIRIRampModel, MaskModel
+from jwst.datamodels import GuiderRawModel, RampModel
+from jwst.datamodels import ImageModel, dqflags
```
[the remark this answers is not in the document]
cracraft: done

jwst/dq_init/tests/test_dq_init.py
```diff
@@ -0,0 +1,340 @@
+import pytest
+from astropy.io import fits
```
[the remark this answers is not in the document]
cracraft: done

jwst/dq_init/tests/test_dq_init.py
cracraft: done

jwst/dq_init/tests/test_dq_init.py
cracraft: The order for parameters in make_rawramp have been updated to be y first and then x.

jwst/dq_init/tests/test_dq_init.py
hbushouse: 👍

hbushouse: OK, so apparently the creation of `err` and all flavors of `dq` arrays is done when the data model instance is created, rather than when it's populated from a file. That's good.

jwst/dq_init/tests/test_dq_init.py
jdavies-st: No comments are fin as is.  Just the docstring after `def foo()`.  These are special in python and are used to generate documentation.  Just for consistency's sake.  =)

jdavies-st:
Thanks @cracraft!  All systems go.  Merging.

PR #2866, opened 2018-11-14
Fix circular import in schema.py

bernie-simon:
The addition of build_docstring added a circular import to datamodels. I fixed it in the usual way by moving the import of model_base into the function.


bernie-simon:
Closing this pull request because extraneous changes leaked in here.

PR #2872, opened 2018-11-15
Update ModelContainer to accept already open associations

bernie-simon:
Previously ModelContainer only accepted filenames. It's been modified  to work if its argument is an open association. The generic open function in util.py has also been modified to accept an open association.

 Associations are plain dictionaries. The new fumction is_association looks for the fields asn_pool and asn_id and if it finds them, says the dictionary is an association.


stscieisenhamer:
Closing due to changes needed and original owner no longer present.

PR #2879, opened 2018-11-16
Filter out reffile keyword validation warnings out of dq_init tests

jdavies-st:
Since a `MaskModel` reffile is generated on the fly in 2 of the `dq_init` tests, and it's required keywords such as
```
meta.description
meta.telescope
meta.reftype
meta.author
meta.pedigree
meta.useafter
```
are not populated, we get validation warnings.  These keywords are there for CRDS purposes and are not needed in the test, so let's just filter out these warnings.

### DIFF
--- a/jwst/dq_init/tests/test_dq_init.py
+++ b/jwst/dq_init/tests/test_dq_init.py
@@ -1,7 +1,9 @@
+import warnings
 import pytest
 from jwst.dq_init import DQInitStep
 from jwst.dq_init.dq_initialization import do_dqinit
 from jwst.datamodels import MIRIRampModel, MaskModel, GuiderRawModel, RampModel, dqflags
+from jwst.datamodels.validate import ValidationWarning
 import numpy as np
 
 
@@ -129,6 +131,9 @@ def test_err():
     ref_data.meta.subarray.ystart = ystart
     ref_data.meta.subarray.ysize = ysize
 
+    # Filter out validation warnings from ref_data
+    warnings.filterwarnings("ignore", category=ValidationWarning)
+
     # run correction step
     outfile = do_dqinit(dm_ramp, ref_data)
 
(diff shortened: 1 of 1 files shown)


cracraft:
Would you prefer that we add the keywords into the model? Either one works for me. Can you tell me what command to use to find out what keywords are required for each model?

jdavies-st:
No, adding the keywords would be an extra ~16 lines of code.  This is easier. And as I said, they're only needed for CRDS, and these models are not intended for that use.

That said, it would be nice to have required keywords discoverable.  Right now, the `validate()` method in `MaskModel` lists the required fields for all reffiles.
```

In [2]: mask = datamodels.MaskModel()                                                                                                                                

In [3]: mask.validate??                                                                                                                                              
Signature: mask.validate()
Source:   
    def validate(self):
        """
        Convenience function to be run when files are created.
        Checks that required reference file keywords are set.
        """
        try:
            assert self.meta.description is not None
            assert self.meta.telescope == 'JWST'
            assert self.meta.reftype is not None
            assert self.meta.author is not None
            assert self.meta.pedigree is not None
            assert self.meta.useafter is not None
            assert self.meta.instrument.name is not None
        except AssertionError:
            if self._strict_validation:
                raise
            else:
                tb = sys.exc_info()[-1]
                tb_info = traceback.extract_tb(tb)
                text = tb_info[-1][-1]
                warnings.warn(text, ValidationWarning)

        super(ReferenceFileModel, self).validate()
File:      ~/dev/jwst/jwst/datamodels/reference.py
Type:      method
```
And you can call `validate()` and see what is missing.  Then fix it, and call again, to see what else is missing.
```
In [4]: mask.validate()                                                                                                                                              
/Users/jdavies/dev/jwst/jwst/datamodels/reference.py:48: ValidationWarning: assert self.meta.description is not None
  warnings.warn(text, ValidationWarning)
```
Clunky for now, I know.

You can also turn on `strict_validation` so you get errors:
```
In [9]: mask = datamodels.MaskModel(strict_validation=True)                                                                                                          

In [10]: mask.validate()                                                                                                                                             
AssertionError                            Traceback (most recent call last)
<ipython-input-10-9ade1670369c> in <module>
----> 1 mask.validate()

~/dev/jwst/jwst/datamodels/reference.py in validate(self)
     32         """
     33         try:
---> 34             assert self.meta.description is not None
     35             assert self.meta.telescope == 'JWST'
     36             assert self.meta.reftype is not None

AssertionError: 
```
Anyway, I would like to make the required keywords more easily discoverable.

PR #2880, opened 2018-11-17
Update ref file docs for straylight and superbias

hbushouse:
Update the reference file docs for the `straylight` and `superbias` steps, for #2675. Also contains a minor fix for the previous updates to `saturation` reffile docs. Reviewers: feel free to only review the docs for your steps.

### DIFF
--- a/docs/jwst/superbias/reference_files.rst
+++ b/docs/jwst/superbias/reference_files.rst
@@ -1,23 +1,58 @@
-Reference File Types
-====================
-The superbias subtraction step uses a SUPERBIAS reference file.
+Reference Files
+===============
+The ``superbias`` subtraction step uses a SUPERBIAS reference file.
 
-.. include:: ../includes/standard_keywords.rst
+SUPERBIAS Reference File
+------------------------
+
+:REFTYPE: SUPERBIAS
+:Data model: `~jwst.datamodels.SuperBiasModel`
+
+The SUPERBIAS reference file contains a 2-D image of the detector bias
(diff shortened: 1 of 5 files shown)


docs/jwst/superbias/reference_files.rst
```diff
@@ -1,23 +1,58 @@
...
-SUPERBIAS Reference File Format
--------------------------------
-Superbias reference files are FITS files with 3 IMAGE extensions and 1 BINTABLE
-extension. The FITS primary data array is assumed to be empty. The 
-characteristics of the three image extension are as follows:
+.. include:: ../includes/standard_keywords.rst
+
+Type Specific Keywords for SUPERBIAS
+++++++++++++++++++++++++++++++++++++
+In addition to the standard reference file keywords listed above,
```
dmggh: All of the superbias updates look good to me. Extremely minor comment: the 'standard reference file keywords' are in the rst file in the /includes directory, so would it be clearer to have 'the standard reference file keywords included above...' ?

hbushouse: When you're reading the built version of the docs (html, pdf, ...) it looks as if the standard keywords are listed directly within the document being read. It's only under the hood that we use the include mechanism for efficiency. So I think the wording is OK as is.

hbushouse:
Closing and reopening just to retrigger jenkins tests.

hbushouse:
Jenkins tests seem to be hung. Travis passed OK. Merging.

PR #2891, opened 2018-11-21
update docs conf to conform to pygments change

sosey:
small change to the docs to conform to template change for `stsci_rtd_theme` and pygments. See also https://github.com/spacetelescope/stsci_rtd_theme/pull/10 and accept this after that PR is merged. I'll push a release for the theme then as well.


sosey:
@hbushouse bueller...

PR #2897, opened 2018-11-26
Remove soon-to-be-deprecated usage of AsdfFile.open

jdavies-st:
This re-implements #2814 and adds 2 unit tests to make sure read-only files can be read by `datamodels`.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -15,6 +15,7 @@ from astropy.io import fits
 from astropy.time import Time
 from astropy.wcs import WCS
 
+import asdf
 from asdf import AsdfFile
 from asdf import yamlutil
 from asdf import schema as asdf_schema
@@ -119,15 +120,15 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
         shape = None
 
         if init is None:
-            asdf = self.open_asdf(init=None, extensions=self._extensions,
+            asdffile = self.open_asdf(init=None, extensions=self._extensions,
                                   **kwargs)
 
         elif isinstance(init, dict):
-            asdf = self.open_asdf(init=init, extensions=self._extensions,
+            asdffile = self.open_asdf(init=init, extensions=self._extensions,
                                   **kwargs)
 
         elif isinstance(init, np.ndarray):
-            asdf = self.open_asdf(init=None, extensions=self._extensions,
+            asdffile = self.open_asdf(init=None, extensions=self._extensions,
                                   **kwargs)
 
             shape = init.shape
@@ -140,7 +141,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
 
             shape = init
             is_shape = True
-            asdf = self.open_asdf(init=None, extensions=self._extensions,
+            asdffile = self.open_asdf(init=None, extensions=self._extensions,
                                   **kwargs)
 
         elif isinstance(init, DataModel):
@@ -150,10 +151,10 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
             return
 
         elif isinstance(init, AsdfFile):
-            asdf = init
+            asdffile = init
 
         elif isinstance(init, fits.HDUList):
-            asdf = fits_support.from_fits(init, self._schema,
+            asdffile = fits_support.from_fits(init, self._schema,
                                           self._extensions, self._ctx)
 
         elif isinstance(init, (str, bytes)):
@@ -163,7 +164,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
 
             if file_type == "fits":
                 hdulist = fits.open(init)
-                asdf = fits_support.from_fits(hdulist,
+                asdffile = fits_support.from_fits(hdulist,
                                               self._schema,
                                               self._extensions,
                                               self._ctx,
@@ -171,7 +172,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
                 self._files_to_close.append(hdulist)
 
             elif file_type == "asdf":
-                asdf = self.open_asdf(init=init, extensions=self._extensions,
+                asdffile = self.open_asdf(init=init, extensions=self._extensions,
                                       **kwargs)
 
             else:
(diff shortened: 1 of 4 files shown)


jdavies-st:
@jaytmiller if you could use this branch in your `crds` unit tests, that would be a nice test as well.  =)

jwst/datamodels/model_base.py
```diff
@@ -140,7 +141,7 @@ def __init__(self, init=None, schema=None, extensions=None,
 
             shape = init
             is_shape = True
-            asdf = self.open_asdf(init=None, extensions=self._extensions,
+            asdffile = self.open_asdf(init=None, extensions=self._extensions,
                                   **kwargs)
 
         elif isinstance(init, DataModel):
```
jdavies-st: I'm not pleased that `asdffile` goes undefined in this block.  Yet our tests pass.  I still feel it is a bug.

drdavella: Maybe we should set `asdffile = None` above just for consistency/sanity? FWIW this seems like something that would be picked up by a code style check.

jdavies-st: That works.

Your code style checker must be better than mine!  😂

drdavella: Er, maybe I meant a linter more correctly. But I'm not sure if anyone is using `pylint` around here?

jdavies-st: I now see that the `return` means the lower code below does not get run.  🤦‍♂️ So yes, just style, not a bug.

jwst/datamodels/tests/test_open.py
```diff
@@ -96,6 +98,48 @@ def test_open_reference_files():
...
+        model.meta.instrument.channel = 'SHORT'
+        model.save(tmpfile)
+
+    os.chmod(tmpfile, 0o440)
+    assert os.access(tmpfile, os.W_OK) == False
+
+    with datamodels.open(tmpfile) as model:
+        assert model.meta.telescope == 'JWST'
+
+    # The following *should* fail with a permission error, but it does not.
```
drdavella: It's not clear what you mean by this comment. Is this a problem with data models? If so, I would recommend adding the correct assertion here (using `pytest.raises`) and then marking this test with `xfail`. This will make it easier to test this issue once the proper fix is applied. (Same applies for the test below).

drdavella: Actually, I think it makes sense to just move the `save` part of it to a separate test and mark that one as `xfail` so that these are allowed to pass.

jdavies-st: Thanks.  Yes, I would like to resolve this functionality before merging.  Should it or should it not fail.  I will have to trace through the code and figure out what is going on.

jwst/datamodels/model_base.py
jdavies-st: We've been using `flake8`, but that doesn't catch this.

jwst/datamodels/tests/test_open.py
jdavies-st: Ok.  These `save` parts of the test overwrite the file because a new `AsdfInFits` class instance is created in order to write it out.  Using `asdf.open` vs `AsdfFile.open` on the original read-only file does not affect this functionality in `datamodels`.  So I should just remove these asserts for now I think.

We can always revisit this "feature" in a separate issue or PR.

jdavies-st: For `datamodels`, the `save` method always uses `overwrite=True`, so if it can it will overwrite it, even if it is read-only.  In the case where the user doesn't have permissions to overwrite a read-only file, we get
```

In [2]: distortion_file = '/grp/crds/cache/references/jwst/jwst_nircam_distortion_0066.asdf'

In [3]: with datamodels.open(distortion_file) as model:
    ...:     print(model._asdf._mode) 
    ...:     model.save(distortion_file) 
    ...:                                                                                                                                                
r
PermissionError                           Traceback (most recent call last)
<ipython-input-11-b8b03ca9da2d> in <module>
      1 with datamodels.open(distortion_file) as model:
      2     print(model._asdf._mode)
----> 3     model.save(distortion_file)
      4 

~/dev/jwst/jwst/datamodels/model_base.py in save(self, path, dir_path, *args, **kwargs)
    513             self.to_fits(output_path, *args, **kwargs)
    514         elif ext == '.asdf':
--> 515             self.to_asdf(output_path, *args, **kwargs)
    516         else:
    517             raise ValueError("unknown filetype {0}".format(ext))

~/dev/jwst/jwst/datamodels/model_base.py in to_asdf(self, init, *args, **kwargs)
    580         asdffile = self.open_asdf(self._instance, extensions=self._extensions,
    581                               **kwargs)
--> 582         asdffile.write_to(init, *args, **kwargs)
    583 
    584     @classmethod

~/miniconda3/envs/jwst_dev/lib/python3.7/site-packages/asdf/asdf.py in write_to(self, fd, all_array_storage, all_array_compression, auto_inline, pad_blocks, include_block_index, version)
   1031 
   1032 
-> 1033         with generic_io.get_file(fd, mode='w') as fd:
   1034             # TODO: This is not ideal: we really should pass the URI through
   1035             # explicitly to wherever it is required instead of making it an

~/miniconda3/envs/jwst_dev/lib/python3.7/site-packages/asdf/generic_io.py in get_file(init, mode, uri, close)
   1187             realpath = url2pathname(parsed.path)
   1188             if mode == 'w':
-> 1189                 fd = atomicfile.atomic_open(realpath, realmode)
   1190             else:
   1191                 fd = open(realpath, realmode)

~/miniconda3/envs/jwst_dev/lib/python3.7/site-packages/asdf/extern/atomicfile.py in atomic_open(filename, mode)
    137     f = tempfile.NamedTemporaryFile(mode, prefix='.___atomic_write',
    138                                     dir=os.path.dirname(filename),
--> 139                                     delete=False)
    140     return _AtomicWFile(f, f.name, filename)

~/miniconda3/envs/jwst_dev/lib/python3.7/tempfile.py in NamedTemporaryFile(mode, buffering, encoding, newline, suffix, prefix, dir, delete)
    545         flags |= _os.O_TEMPORARY
    546 
--> 547     (fd, name) = _mkstemp_inner(dir, prefix, suffix, flags, output_type)
    548     try:
    549         file = _io.open(fd, mode, buffering=buffering,

~/miniconda3/envs/jwst_dev/lib/python3.7/tempfile.py in _mkstemp_inner(dir, pre, suf, flags, output_type)
    256         file = _os.path.join(dir, pre + name + suf)
    257         try:
--> 258             fd = _os.open(file, flags, 0o600)
    259         except FileExistsError:
    260             continue    # try again

PermissionError: [Errno 13] Permission denied: '/grp/crds/cache/references/jwst/.___atomic_writexq917gcy'
```

Issue #1658, opened 2018-01-11
Quaternion handling with incomplete quaternion data


stscieisenhamer:
# Issue
Better handling of quaternion calculation when some/all of the quaternion information is not with the exposure period.

# Resources
- [DR #4622](https://www.ess.stsci.edu/prsystem/servlet/prbrowse/dr.4622)

stscieisenhamer:
As per the DR, current approach is to ensure that the engineering db is self-consistent. Until the DR itself is resolved, no code changes are desired.

stscieisenhamer:
Closing this because, as per the DR, this is not `set_telescope_pointing` responsibility.

stscieisenhamer:
Though not responsibility, the error situation needs to be handled more gracefully.  In particular, 
https://github.com/spacetelescope/jwst/blob/200642764bff554b0f5f06234bed47a980d52e67/jwst/lib/set_telescope_pointing.py#L947-L950 
should be in a `try` block, with the exception being reported, break out of the loop, and continue.

hbushouse:
@stscirij I'm guessing this was fixed as a by-product of #5540. Can you confirm?

stscirij:
> @stscirij I'm guessing this was fixed as a by-product of #5540. Can you confirm?

I believe so, although the link to the original issue in the PR system is now stale so it's hard to tell what the exact issue raised was

hbushouse:
The part of the code referenced above doesn't even exist anymore and is now part of larger sections that check for zero-valued entries and raise an error when found. So I'm going to assume this has been handled and if anything bad shows up in the future, a new ticket can be created. Closing.

PR #2900, opened 2018-11-27
Allow interpolation and non-strict time matching of engineering parameters

### LINKED ISSUE #2886: Implement interpolation & averaging of engineering pointing info
# Issue
There are three matrices that come from the engineering telemetry that define how the spacecraft is pointing. These arrive with different cadences as follows:

- Quaternions (ZATTEST*): ~ 1/minute
- FSM correction (ZADUCMDX/Y): ~5/sec
- FGS J-frame matrix (ZRFGS2Jxx): ~5/sec

Being slow at a slow cadence, the quaternion will not have data for exposures less than a minute. Conversely, the other two will produce much data.  Algorithms will need be identified on how to both interpolate and average/select values.

Current implementation chooses the first values appearing within the exposure. Failures from OTB/EMTB data retrieving quaternions have been seen because no quaternion appear in the exposure time.

stscieisenhamer:
Resolves #2886.


stscieisenhamer:
Resolves #1658

PR #2905, opened 2018-11-28
Add unit tests for saturation step

### LINKED ISSUE #2569: NIRCam team to review and agree with unit-testing plan.
Issue [JP-386](https://jira.stsci.edu/browse/JP-386) was created by Macarena Garcia Marin:

NIRCam team to review and agree with unit-testing plan.

### LINKED ISSUE #2571: MIRI team to review and agree with unit-testing plan.
Issue [JP-384](https://jira.stsci.edu/browse/JP-384) was created by Macarena Garcia Marin:

MIRI team to review and agree with unit-testing plan.

aliciacanipe:
@cracraft @mfixstsci I am adding unit tests for the saturation step of the pipeline. 

Resolves #2571
Resolves #2569


jwst/saturation/tests/test_saturation.py
```diff
@@ -0,0 +1,344 @@
...
+        nints = 1
+
+        # create the data and groupdq arrays
+        csize = (nints, ngroups, nrows, ncols)
+        data = np.full(csize, 1.0)
+        pixeldq = np.zeros((nrows, ncols), dtype=int)
+        groupdq = np.zeros(csize, dtype=int)
+
+        # create a JWST datamodel for MIRI data
+        data_model = MIRIRampModel(data=data, pixeldq=pixeldq, groupdq=groupdq)
```
jdavies-st: Same as my comment above for NIRCam.  The `data`, `pixeldq` and `groupdq` automatically get generated (filled with zeros of the correct `dtype`) when the constructor is passed a shape tuple.
```
data_model = MIRIRampModel(csize)
data_model.data += 1
```

jwst/saturation/tests/test_saturation.py
```diff
@@ -0,0 +1,344 @@
...
+
+    # Add ramp values up to the saturation limit
+    data.data[0, 0, 500, 500] = 0
+    data.data[0, 1, 500, 500] = 20000
+    data.data[0, 2, 500, 500] = 40000
+    data.data[0, 3, 500, 500] = 60000   # Signal reaches saturation limit
+    data.data[0, 4, 500, 500] = 62000
+
+    # Get reference file
+    reffile = "/grp/crds/cache/references/jwst/jwst_nircam_saturation_0064.fits"
```
jdavies-st: This line is causing this test to fail.  None of our CI testing infrastructure has access to central store, so instead of setting `override_saturation`, perhaps one could just specify the required header keywords in `data` so that the 0064 reffile gets selected.

The current rmap for NIRCam saturation reffile

https://jwst-crds.stsci.edu

Indicates that the following keywords are required:
```
data.meta.instrument.detector = 'NRCA1'
data.meta.observation.date = '2017-10-01'
data.meta.observation.time = '00:00:00'
```
and it will get selected.  I don't know if this is too fragile, but it will work.  One may want to first assert that the reffile is the correct one you're expecting too.
```
assert output.meta.ref_file.saturation.name == 'crds://jwst_nircam_saturation_0064.fits'
```

jdavies-st:
I've been holding off merging this until we can figure out if we want to have these new unit tests depend not only on CRDS, but on specific reffiles from CRDS getting selected. Perhaps @hbushouse has a strong opinion?

I do recall in conversation with @cracraft that the preferable method would be to mock up the reference file and then be able to pass that in memory as a `saturation_override` arg.  `stpipe.Step` doesn't currently accept this, but an easy way to get around this is to use pytest's `monkeypatch` fixture.

https://docs.pytest.org/en/latest/monkeypatch.html

Basically this allows one to redefine the return of a function, method, attribute, etc, just for the current test.  Think of it as a short circuit.  In this case, we can short-circuit two functions in the `SaturationStep.call` method.  One calls CRDS, gets the reffile name, finds it or downloads it locally, and returns the path.  We can just short-circuit that and have it return `None` instead.  The other opens that path as a `SaturationModel`, but we want to use our mock saturation model instead, so again, we short-circuit that function too and make the call to it return our mocked `SaturationModel`.

So here's some example code that works:
```
def test_full_step(setup_nrc_cube, monkeypatch):
     '''Test full run of the SaturationStep.'''

    # Create inputs, data, and saturation maps
    ngroups = 5
    nrows = 2048
    ncols = 2048
    satvalue = 60000

    ...

    satmap.data[500, 500] = satvalue

    # monkeypatch with the mock saturation reffile
    monkeypatch.setattr('jwst.stpipe.Step.get_reference_file', lambda x,y,z: None)
    monkeypatch.setattr('jwst.datamodels.SaturationModel', lambda x: satmap)

     # Run the pipeline
     output = SaturationStep.call(data)
```
Because both are functions, the `monkeypatch` call needs to use `lambda` with the same number of args the function calls, but again, we're just returning the exact items we need for this test.

Does this seem like a reasonable workaround that does not rely on CRDS to pick the right reffile for us?

aliciacanipe:
Misty and I decided it was best to keep the tests as simple as possible and remove the reference file dependence altogether for superbias and saturation. The latest commit has the simplified test for the full saturation step.

Issue #2898, opened 2018-11-27
Improve support for WFSS data.


philhodge:
EXP_TYPE = NRC_TSGRISM needs to be added to the list of WFSS exposure types in `extract_1d`.  For this mode, the wavelength attribute is not populated by `extract_2d`.  When the output of `extract_2d` is then opened by `extract_1d`, because the model is a `SlitDataModel` object, the wavelength attribute is 3-D (`extract_1d` expects 2-D), and the wavelength values are uninitialized.  The wavelength section of slitdata.schema.yaml should specify "ndim: 2" and "default: 0.0".

hbushouse:
Fixed by #2911

PR #2911, opened 2018-11-30
Improve support for WFSS data.

philhodge:
'NRC_TSGRISM' was added to `WFSS_EXPTYPES`, the list of exposure types that are WFSS/GRISM types.

The `SlitDataModel` schema was modified to specify that the wavelength attribute should be two-dimensional, and that the default value is zero.

See issue #2898.

### DIFF
--- a/jwst/datamodels/schemas/slitdata.schema.yaml
+++ b/jwst/datamodels/schemas/slitdata.schema.yaml
@@ -107,6 +107,8 @@ allOf:
     wavelength:
       title: Wavelength array, corrected for zero-point
       fits_hdu: WAVELENGTH
+      ndim: 2
+      default: 0.0
       datatype: float32
     barshadow:
       title: Bar shadow correction
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -55,6 +55,9 @@ exp_to_source
 
 extract_1d
 ----------
+- Updated to recognize NRC_TSGRISM as WFSS data.  SlitDataModel schema now
+  specifies that the wavelength attribute should be 2-D, with a default
+  value of 0. [#2911]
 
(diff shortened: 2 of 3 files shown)


jwst/datamodels/schemas/slitdata.schema.yaml
```diff
@@ -107,6 +107,8 @@ allOf:
     wavelength:
       title: Wavelength array, corrected for zero-point
       fits_hdu: WAVELENGTH
+      ndim: 2
```
hbushouse: Do we know whether this may cause any problems downstream in any steps that might be assuming that 3D data will also have a 3D wavelength array? Granted, for TSO modes, the wavelengths in each plane of the 3D stack would be identical, but at some point we'll just need to make sure downstream code isn't trying to access the 3rd dimension of "wavelength".

philhodge: I don't know for sure.  I can't find email about this right now, but my understanding was that we adopted a policy that the wavelength array (attribute) would always be 2-D, regardless of the shape of the SCI data.  For TSO data, a 3-D wavelength array could be large.  The `SlitDataModel` was created relatively recently; it's the one that did not specify ndim and default.  The `CubeModel` schema does already specify "ndim: 2", and "default: 0.0".  Currently, `extract_2d` does not create the wavelength attribute for NRC_TSGRISM data, and `extract_1d` got a 3-D wavelength attribute containing uninitialized memory, so if any step relied on those wavelengths, that step would be using garbage.

nden: I think @philhodge is right and not having `ndim` in the schema was an oversight.

Issue #2915, opened 2018-12-03
NirspecIFUCubeParsModel schema is inconsistent with delivered reffile


jdavies-st:
The delivered reffile uses a column name `disperser`, whereas the schema indicates this should be `grating`.  Example below:
```
In [39]: reffile = datamodels.NirspecIFUCubeParsModel('/grp/crds/cache/references/jwst/jwst_nirspec_cubepar_0002.fits')                                 
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-39-4a74f3692902> in <module>
----> 1 reffile = datamodels.NirspecIFUCubeParsModel('/grp/crds/cache/references/jwst/jwst_nirspec_cubepar_0002.fits')

~/dev/jwst/jwst/datamodels/reference.py in __init__(self, init, **kwargs)
     22 
     23     def __init__(self, init=None, **kwargs):
---> 24         super(ReferenceFileModel, self).__init__(init=init, **kwargs)
     25         self._no_asdf_extension = True
     26         self.meta.telescope = "JWST"

~/dev/jwst/jwst/datamodels/model_base.py in __init__(self, init, schema, extensions, pass_invalid_values, strict_validation, **kwargs)
    170                                               self._extensions,
    171                                               self._ctx,
--> 172                                               **kwargs)
    173                 self._files_to_close.append(hdulist)
    174 

~/dev/jwst/jwst/datamodels/fits_support.py in from_fits(hdulist, schema, extensions, context, **kwargs)
    540 
    541     known_keywords, known_datas = _load_from_schema(hdulist, schema,
--> 542                                                     ff.tree, context)
    543     _load_extra_fits(hdulist, known_keywords, known_datas, ff.tree)
    544     _load_history(hdulist, ff.tree)

~/dev/jwst/jwst/datamodels/fits_support.py in _load_from_schema(hdulist, schema, tree, context)
    487                 return True
    488 
--> 489     mschema.walk_schema(schema, callback)
    490     return known_keywords, known_datas
    491 

~/dev/jwst/jwst/datamodels/schema.py in walk_schema(schema, callback, ctx)
    210                 recurse(items, path + ['items'], combiner, ctx)
    211 
--> 212     recurse(schema, [], None, ctx)
    213 
    214 

~/dev/jwst/jwst/datamodels/schema.py in recurse(schema, path, combiner, ctx)
    200         if schema.get('type') == 'object':
    201             for key, val in schema.get('properties', {}).items():
[...]

PR #2916, opened 2018-12-03
fix nirspec cubepars schema

jemorrison:
Updated the nirspec_cubepars schema to have the correct name associated with DISPERSER.
The code for reading in the reference file already was updated with DISPERSER.


jemorrison:
address #2915

PR #2936, opened 2018-12-12
Tables of reference file docs

hbushouse:
First round of major reorganization to reference file docs.

1) All ref file docs have been pulled out the step docs and have been made standalone docs (one for each reffile type) that now reside in /docs/jwst/references_general/, including any "include" files, such as the ones that describe the CRDS selectors for each reffile type.

2) The docs for each step now load the appropriate reffile docs via "include"s

3) The list of reference file types vs steps in the general reference file doc has been updated to contain 2 tables: one that gives REFTYPEs vs. step and one that gives step vs. REFTYPEs.

4) Still in progress: active links from the tables mentioned above to the docs for each pipeline step.


hbushouse:
Most recent commit restructures all the docs for reference files used by `assign_wcs`. Previously they had been described in one monolithic document included with the `assign_wcs` step itself. That document has been split apart into separate docs describing each reffile, in a format similar to all other ref files (e.g. include indications of REFTYPE, link to datamodel, table of keyword selectors, etc.). All the actual content describing file formats, etc. has been copied from the original.

This also allows for the tables in the general reference file doc to now contain links to **all** reference file types (the circle is now complete ...).

hbushouse:
This completes the majority of the work for #2675.

hbushouse:
Reviewers: It's probably easiest to download all of these changes and build the docs, and then do the review from the actual sphinx docs, rather than trying to figure out what's going on by looking directly at all the changes in the .rst files.

jaytmiller:
I've barely started  but I noticed "explain_rmaps.rst~" in references_general.   Should be deleted I think.   (Nice matrix!)

hbushouse:
I don't see the "explain_rmaps.rst~" file. Just the normal "explain_rmaps.rst".

jaytmiller:
Sorry,  you're right,  I was tricked by my "git ignore" for ~ files.   It's a local left over from when I wrote explain_rmaps.rst and not in the repo.

jaytmiller:
I noticed that I generated selection criteria for ASDF in FITS notation.  I probably did it on purpose for brevity but it does seem a little off.   Not sure what "fixed" is though.  Datamodels names are pretty verbose,  particularly when you string 3-4 together.   (If there was a "better",  I could adjust and regenerate.)

hbushouse:
If you're referring to the lists of FITS keyword names that are listed as selectors for all the WCS ref files (e.g. CAMERA, DISPERSER, FPA, MSA, ...), I generated those tables from the selection info that was provided in text mode in the `assign_wcs` docs at https://jwst-pipeline.readthedocs.io/en/latest/jwst/assign_wcs/reference_files.html. The wording there suggests that the matching is done based on the values in the input science data set, in which case the FITS keyword names are correct (because the input science data sets are always FITS). It's the "keyword" names in the ASDF reference files that are different. But the selection isn't really done based on those values - it's done via the science FITS file keyword names matching to entries in the rmaps. So perhaps using the FITS style names is still appropriate for the ASDF ref files?

jaytmiller:
I noticed GAIN is not requiring BUNIT because it's not a matching parameter although units are mentioned.   Certainly for some reftypes people get these wrong so calling out BUNIT somehow might be good,  maybe just inline with e/DN.   From the description it may be unlikely,  but possibly for other reftypes.

jaytmiller:
Type specific keywords for SATURATION may be wrong.

hbushouse:
> Type specific keywords for SATURATION may be wrong.

According to CRDS rmaps, they're correct. All instruments are INSTRUME, DETECTOR, and DATE/TIME, except for NIRSpec, which adds SUBARRAY.

hbushouse:
> I noticed GAIN is not requiring BUNIT because it's not a matching parameter although units are mentioned. Certainly for some reftypes people get these wrong so calling out BUNIT somehow might be good, maybe just inline with e/DN. From the description it may be unlikely, but possibly for other reftypes: READNOISE. (I checked and don't see BUNIT in the schema. Was it dropped?)

Very few gain reffiles currently in CRDS have the BUNIT keyword, and only a slightly larger fraction of readnoise files have it. So that suggests we aren't requiring it, but it certainly is encouraged. I can at least add the encouragement to the docs.

I don't recall if we ever had it in the gain or readnoise reffile schemas. We should probably add it, so it's at least available to reffile creators.

hbushouse:
Filed #2943 to add BUNIT to the GAIN and READNOISE schemas.

hbushouse:
Latest commit adds BUNIT keyword info to the GAIN and READNOISE reffile docs and also adds a link to the general reffile info page from the main introduction.

hbushouse:
> > Type specific keywords for SATURATION may be wrong.
> 
> According to CRDS rmaps, they're correct. All instruments are INSTRUME, DETECTOR, and DATE/TIME, except for NIRSpec, which adds SUBARRAY.
> 
> > For this I'm referring to _type specific_,  not _selection_.   Selection are correct as you say.  Type specific runs through the unique selection parameters and defines them father down.  It's defining DETECTOR, EXP_TYPE, FILTER, PUPIL,  not the ones you mention above.   I think it needs to keep DETECTOR, add SUBARRAY, and drop EXP_TYPE, FILTER, PUPIL.   (Type specific is where the data models translations are defined.)

Ah, yes, I see now. That table doesn't need FILTER and PUPIL, but needs to add SUBARRAY. Will fix in next commit. Good eye!

nden:
I am just starting now to look at this so it will take me a while.
The front page has no mention of JWST. It only says `User Manual`. 
Can we mention it's about JWST? 

Also after building the docs I see the number of warnings had increased roughly twice, to 500.
But that can be fixed in a different PR.

hbushouse:
> I am just starting now to look at this so it will take me a while.
> The front page has no mention of JWST. It only says `User Manual`.
> Can we mention it's about JWST?
> 
> Also after building the docs I see the number of warnings had increased roughly twice, to 500.
> But that can be fixed in a different PR.

Yes, there are a lot of warnings now because all of the <reftype>_reffile.rst are used in a "standalone" fashion and aren't listed directly in any index.rst or other file, so we get a lot of warnings of the type:
```
/Users/bushouse/JWST/softw/jwst/docs/jwst/references_general/camera_reffile.rst: WARNING: document isn't included in any toctree
/Users/bushouse/JWST/softw/jwst/docs/jwst/references_general/collimator_reffile.rst: WARNING: document isn't included in any toctree
/Users/bushouse/JWST/softw/jwst/docs/jwst/references_general/disperser_reffile.rst: WARNING: document isn't included in any toctree
...
```
I also get a lot of warnings from my links that are used to point to other files, e.g.
```
docs/jwst/references_general/straymask_reffile.rst:4: WARNING: duplicate label straymask_reffile, other instance in /Users/bushouse/JWST/softw/jwst/docs/jwst/references_general/straymask_reffile.rst
docs/jwst/references_general/straymask_selection.rst:4: WARNING: duplicate label straymask_selectors, other instance in /Users/bushouse/JWST/softw/jwst/docs/jwst/references_general/straymask_selection.rst
docs/jwst/references_general/superbias_reffile.rst:4: WARNING: duplicate label superbias_reffile, other instance in /Users/bushouse/JWST/softw/jwst/docs/jwst/references_general/superbias_reffile.rst
```
But the warnings point to the same instance of a label in the same file and claim it's a duplicate. So I don't know what's going on there. Can someone verify that I'm using external (i.e. from one .rst file to another) links correctly?

nden:
> Yes, there are a lot of warnings now because all of the _reffile.rst are used in a "standalone" fashion and aren't listed directly in any index.rst or other file, so ...

I think adding `:orphan:` at the top of those pages will fix this. I'll push to your branch some changes shortly.

nden:
I suggest you merge this and I'll submit another PR for the warnings. I fixed ~ half of them but am still working on the others. If you merge this we'll have a baseline to compare to.

hbushouse:
At the advice and consent of @nden I'm merging ...

Issue #2893, opened 2018-11-22
coron3 pipeline gives unstable results


jdavies-st:
Our regression tests show that the pipeline results can vary on the order of a few percent depending on upstream dependency changes, i.e. in `mkl`, `blas` and `numpy`.  Here's an example:

https://boyle.stsci.edu:8081/job/RT/job/JWST/55/testReport/junit/jwst.tests_nightly.general.nircam.test_coron3_1/TestCoron3Pipeline/test_coron3_1/

No code in `jwst` changed, yet most pixels in the output images changed by a few percent.

We should investigate this. Perhaps there's some machine precision problems?  Are we using double precision where needed?  `datamodels` stores and uses 32-bit single-precision floats by default.

This is not the first time we've seen this in `coron3`, though it is not the only pipeline either (`ami` suffers from this as well).

The failure of the RT above were due to a change in the `mkl` library use for on-chip acceleration.  The new versions:
```
2018-11-19T16:03:35 : main/osx-64 : mkl-include-2019.1-144.tar.bz2          
2018-11-19T16:04:03 : main/osx-64 : mkl-2019.1-144.tar.bz2                  
2018-11-19T16:08:11 : main/osx-64 : mkl-devel-2019.1-144.tar.bz2            
```
before:
```
[jwst/3.6]     mkl:                2019.1-144                                                                                              
```
test failure above:
```
[jwst/3.6]     mkl:                2018.0.3-1                                                                                                   
```

Issue #1988, opened 2018-05-04
Numpy 1.13 vs 1.14 differences in numerical results


jdavies-st:
Our nightly regression tests show differences in numerical results between numpy 1.13 and numpy 1.14.  This issue is to track those and determine if they're important or not.  We discuss these at length in https://github.com/STScI-JWST/jwst/pull/1888#issuecomment-386433739.

The nightly regression tests affected are:

- [x] niriss.test_ami_analyze.test_ami_analyze
- [x] niriss.test_ami_pipeline.test_ami_pipeline
- [x] pipelines.test_coron3_1.test_coron3_pipeline1
- [ ] pipelines.test_nrc_image3_1.test_image3_pipeline1

We need to

1. Figure out which common numpy function or computation(s) are causing this.
2. Reproduce with a simple test case.
3. Figure out if this is a numpy issue or an Anaconda issue.
4. Decide if we need to fix it, or if we're OK with the differences.

jdavies-st:
Interesting discussion of these issue here

https://sinews.siam.org/Details-Page/reproducible-blas-make-addition-associative-again

ht: @pllim

pllim:
At some point, maybe this 1.13 vs 1.14 won't be relevant anymore because 1.15.3 is out. :grimacing:

jdavies-st:
Yeah, maybe this issue should be a catch-all for BLAS differences between different versions and builds of `numpy`.  😬

PR #2946, opened 2018-12-14
Use double-precision in AMI analyze and pipeline

jdavies-st:
Also:
-  PEP8 cleanup of the AMI class and module names
- close datamodels in the ami3 pipeline

Addresses #1988 and #2893.

### DIFF
--- a/jwst/ami/ami_analyze.py
+++ b/jwst/ami/ami_analyze.py
@@ -7,7 +7,7 @@ import warnings
 import numpy as np
 from .. import datamodels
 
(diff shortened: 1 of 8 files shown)


jdavies-st:
To be clear the important change is here:
https://github.com/jdavies-st/jwst/blob/5b52bca9ddfc775c39e77ab9f3528115e554581c/jwst/ami/ami_analyze.py#L77

jdavies-st:
I have also updated the RT truth files.

PR #2953, opened 2018-12-17
Add linearity unit tests

mfixstsci:
Contributing the unit tests for the linearity correction from @cracraft

### DIFF
--- a/jwst/linearity/tests/test_linearity.py
+++ b/jwst/linearity/tests/test_linearity.py
@@ -6,7 +6,6 @@ Authors:
 from jwst.linearity import LinearityStep
 from jwst.linearity.linearity import do_correction as lincorr
 from jwst.datamodels import dqflags, LinearityModel, MIRIRampModel
-from astropy.io import fits
 import numpy as np
 
 
--- a/jwst/linearity/tests/test_linearity.py
+++ b/jwst/linearity/tests/test_linearity.py
@@ -1,0 +1,596 @@
+"""Unit tests for the non-linearity correction.
+
+Authors:
+    M. Cracraft
+"""
+from jwst.linearity import LinearityStep
+from jwst.linearity.linearity import do_correction as lincorr
+from jwst.datamodels import dqflags, LinearityModel, MIRIRampModel
+from astropy.io import fits
+import numpy as np
+
+
+def test_coeff_dq():
+    """test linearity algorithm with random data ramp (does algorithm match expected algorithm)
+    also test a variety of dq flags and expected output """
+
+    # size of integration
+    nints = 1
+    ngroups = 160
+    xsize = 1032
+    ysize = 1024
+
+    # create the data and groupdq arrays
+    # create a JWST datamodel for MIRI data
+    im = make_rampmodel(nints, ngroups, ysize, xsize)
+    im.meta.instrument.detector = 'MIRIMAGE'
+
+    # Create reference file
+    numcoeffs = 5
+
+    ref_model = LinearityModel()
+    ref_model.data = np.zeros(shape=(numcoeffs, ysize, xsize), dtype=np.float32)
+    ref_model.dq = np.zeros((ysize, xsize), dtype=int)
+
+    ref_model.meta.instrument.name = 'MIRI'
+    ref_model.meta.instrument.detector = 'MIRIMAGE'
+    ref_model.meta.subarray.xstart = 1
+    ref_model.meta.subarray.ystart = 1
+    ref_model.meta.subarray.xsize = xsize
+    ref_model.meta.subarray.ysize = ysize
+
+    # Set coefficient values in reference file to check the algorithm
+    # Equation is DNcorr = L0 + L1*DN(i) + L2*DN(i)^2 + L3*DN(i)^3 + L4*DN(i)^4
+    # DN(i) = signal in pixel, Ln = coefficient from ref file
+    # L0 = 0 for all pixels for CDP6
+    L0 = 0
+    L1 = 0.85
+    L2 = 4.62E-6
+    L3 = -6.16E-11
+    L4 = 7.23E-16
+
+    coeffs = np.asfarray([0.0e+00,  0.85, 4.62e-06,  -6.16e-11, 7.23e-16])
+
+    ref_model.data[:, 300, 500] = coeffs
+    print(ref_model.data[:, 300, 500])
+
+    # check behavior with NaN coefficients: should not alter pixel values
+    coeffs2 = np.asfarray([L0, np.nan, L2, L3, L4])
+    ref_model.data[:, 200, 500] = coeffs2
+    im.data[0, 50, 200, 500] = 500.0
+
+    tgroup = 2.775
+
+    # set pixel values (DN) for specific pixels up the ramp
+    im.data[0, :, 300, 500] = np.arange(ngroups) * 100 * tgroup
+
+    scival = 40000.0
+    im.data[0, 45, 300, 500] = scival  # to check linearity multiplication is done correctly
+    im.data[0, 30, 350, 360] = 35  # pixel to check that dq=2 meant no correction was applied
+
+    # check if dq flags in pixeldq are correctly populated in output
+    im.pixeldq[50, 40] = 1
+    im.pixeldq[50, 41] = 2
+    im.pixeldq[50, 42] = 1024
+    im.pixeldq[50, 43] = 2048
+
+    # set dq flags in DQ of reference file
+    ref_model.dq[350, 350] = 1  # DO_NOT_USE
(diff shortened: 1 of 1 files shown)


jdavies-st:
Looks like all the unit tests where one tries to access a particular file on the central store CRDS cache fail.  Unit tests can't do this as our CI tools do not have (and never will have) access to central store.  Best to use a mock reference file.  See discussion in #2905.

jwst/linearity/tests/test_linearity.py
```diff
@@ -0,0 +1,596 @@
...
+    L0 = 0
+    L1 = 0.85
+    L2 = 4.62E-6
+    L3 = -6.16E-11
+    L4 = 7.23E-16
+
+    coeffs = np.asfarray([0.0e+00,  0.85, 4.62e-06,  -6.16e-11, 7.23e-16])
+
+    ref_model.data[:, 300, 500] = coeffs
+    print(ref_model.data[:, 300, 500])
```
hbushouse: Probably don't want to leave a print statement in here for the CI test runs.

jwst/linearity/tests/test_linearity.py
```diff
@@ -0,0 +1,596 @@
...
+    ref_model.dq[350, 360] = dqflags.pixel['NO_LIN_CORR']  # NO_LIN_CORR
+    ref_model.dq[300, 500] = 0  # good pixel
+
+    # run through Linearity pipeline
+    outfile = lincorr(im, ref_model)
+
+    # check that multiplication of polynomial was done correctly for specified pixel
+    outval = L0+(L1*scival)+(L2*scival**2)+(L3*scival**3)+(L4*scival**4)
+
+    print(outfile.data[0, 45, 300, 500])
```
hbushouse: Same comment as above regarding print statements.

jwst/linearity/tests/test_linearity.py
```diff
@@ -0,0 +1,596 @@
...
+    assert(np.isclose(outfile.data[0, 45, 300, 500], outval, rtol=0.1))
```
hbushouse: I think rtol=0.1 represents a relative tolerance of 10%, which is pretty big (if I'm correct). We should get matches down to at least ~5 significant digits (at 6 or 7 you could start running into machine precision issues, for single precision floats). So this could likely be changed to something on the order of rtol=0.00001.

jwst/linearity/tests/test_linearity.py
```diff
@@ -0,0 +1,596 @@
...
+    # Create reference file
+    dq = np.zeros((1024, 1032), dtype=int)
+    numcoeffs = 3
+
+    # set PIXELDQ to 'NO_LIN_CORR'
+    dq[500, 500] = dqflags.pixel['NO_LIN_CORR']
+    dq[550, 550] = dqflags.pixel['DO_NOT_USE']
+    dq[560, 550] = dqflags.pixel['HOT']
+    dq[550, 560] = dqflags.pixel['DEAD']
+    dq[500, 300] = 2049  # Hot and DO_NOT_USE
```
hbushouse: Instead of having to know the desired values (2048 and 1 in this case), you could potentially just use:
```
dq[500, 300] = np.bitwise_or(dqflags.pixel['HOT'], dqflags.pixel['DO_NOT_USE'])
```
to do the math for you. You've already got numpy imported as "np", so it should be ready to use.

jwst/linearity/tests/test_linearity.py
```diff
@@ -0,0 +1,596 @@
...
+    ref_model.meta.subarray.ysize = 1024
+
+    # run through linearity correction
+    outfile = lincorr(im, ref_model)
+
+    assert(outfile.pixeldq[500, 500] == dqflags.pixel['NO_LIN_CORR'])
+    assert(outfile.pixeldq[550, 550] == dqflags.pixel['DO_NOT_USE'])
+    assert(outfile.pixeldq[560, 550] == dqflags.pixel['HOT'])
+    assert(outfile.pixeldq[550, 560] == dqflags.pixel['DEAD'])
+    assert(outfile.pixeldq[500, 300] == 2049)
```
hbushouse: Same comment as at line 188.

jwst/linearity/tests/test_linearity.py
```diff
@@ -0,0 +1,596 @@
...
+
+    # set filter
+    filter = 'F2100W'
+
+    # run pipeline
+    outfile = LinearityStep.call(im)
+
+    # get linearity reference file from header
+    linfile = outfile.meta.ref_file.linearity.name
+    print('Linfile', linfile)
```
hbushouse: More print statements (and lines 304, 311, 312, ...)

jwst/linearity/tests/test_linearity.py
```diff
@@ -0,0 +1,596 @@
...
+
+    # parse linfile name to discard crds://
+    file = linfile.split("/")[2]
+    print('split file', file)
+
+    # read in reference file
+    filepath = '/grp/crds/jwst/references/jwst/'+file
```
hbushouse: @jdavies-st has already pointed out that CI tests don't have access to central store (/grp).

jwst/linearity/tests/test_linearity.py
cracraft: agreed

jwst/linearity/tests/test_linearity.py
cracraft: agreed

jwst/linearity/tests/test_linearity.py
cracraft: I wasn't sure of the syntax for combining flags in the tests, so agreed to change it.

jwst/linearity/tests/test_linearity.py
cracraft: can remove print statements

jwst/linearity/tests/test_linearity.py
cracraft: This brings up a question. I agree that checking that the correct reference file is being called isn't necessarily a unit test, but it is an important test to run. If we remove these tests from unit testing, is there a place within scsb code that it can be run, or is this something that the teams will have to run on their own?

jdavies-st: Well, the "correct reference file" will depend on CRDS selection criteria, and this is controlled by the reference files that the INS teams deliver to redcat/CRDS.  This is really a CRDS issue I think.  I don't think our cal code should be checking this.  It relies on CRDS to recommend the best reffile for a given dataset.

Perhaps @hbushouse or @jaytmiller have something further to add?

But in these unit tests, you're certainly free to mock up a reference file and use `linearity_override`.  In that case, the test does not depend on reference file deliveries that might change the reference file over time.

jaytmiller:
One comment I should make is that a more portable way of interpreting crds:// is to call:
```
from jwst.stpipe import crds_client
actual_file_path = crds_client.reference_uri_to_cache_path(recorded_uri)
```
There is a related step.reference_uri_to_cache_path(recorded_uri) which is preferred when you have access to a Step or Pipeline instance. 

reference_uri_to_cache_path() will still work if e.g. CRDS_PATH is set to find files in a local cache.    

It doesn't guarantee that the reference is already in the CRDS cache,  but it does interpret crds:// appropriately given env var settings like CRDS_PATH.

One last point to make is that CRDS can be set up to dynamically sync files to a local cache.   When the sync overheads are not too high relative to everything else you're doing,  this can serve as a work around to "we don't have access to /grp/crds/cache."

jwst/linearity/tests/test_linearity.py
mfixstsci: Removed in 095ff52

jwst/linearity/tests/test_linearity.py
mfixstsci: Updated in f98cc1e

jwst/linearity/tests/test_linearity.py
mfixstsci: updated in f98cc1e

mfixstsci:
After talking with @cracraft and @aliciacanipe we decided to remove the tests that used the hardcoded CRDS path in commit f98cc1e and that we would figure out a more robust solution in the future.

jdavies-st:
Looks like one more fix is needed.  If you look at the Travis CI build details above:

https://travis-ci.org/spacetelescope/jwst/jobs/476548264#L1041

So there's an unnecessary import.  And it looks like the Jenkins check is a fluke.  A commit fixing the import will kick off both test runs, and both should pass.

Otherwise looks good!

cracraft:
The astropy.io import was to import fits which was used in the reference file checking tests that were removed, so yes, that import is no longer needed. We'll probably have the same unused import for the dark correction tests.

mfixstsci:
Removed the import in c29328d

Issue #2744, opened 2018-10-22
Add CRDS context to README


stscicrawford:
In our release table, we should also add the delivered CRDS context

hbushouse:
Fixed by #2969

Issue #2176, opened 2018-06-20
Update the help


pllim:
Tyler Desjardins mentions that we should consider moving emails from `help[at]stsci.edu` to point to the web portal where possible and appropriate. For HST (or any non-JWST), it is https://hsthelp.stsci.edu . For JWST, it is https://jwsthelp.stsci.edu . Please update info in `setup.py`, `setup.cfg`, documentation, etc as appropriate. 

Please close this issue if it is irrelevant to your repository. This is an automated issue. *If this is opened in error, please let pllim know!*

xref spacetelescope/hstcal#317

hbushouse:
Closing for jwst.

PR #2969, opened 2018-12-19
Update help and CRDS_CONTEXT info

hbushouse:
In support of #2176 the `author_email` entries have been removed from `setup` and instead the web portal for jwst help has been added to both the README and the introductory doc.

CRDS_CONTEXT numbers have been added to the DMS Build Info table in the README, to address #2744.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Software vs DMS build version map
 
 | jwst tag | DMS build | CRDS_CONTEXT |   Date     |          Notes                           |
 | -------- | --------- | ------------ | ---------- | -----------------------------------------|
-|  0.12.2  | B7.2rc3   | 0495         | 11/07/2018 | Final release candidate for Build 7.2    |
+|  0.12.2  | B7.2      | 0495         | 11/07/2018 | Final release candidate for Build 7.2    |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -71,7 +71,7 @@ https://jwst-pipeline.readthedocs.io/en/latest/
 
 One can clone this repository and build the documentation with
 
-    pip install sphinx_rtd_theme stsci_rtd_theme sphinx_automodapi
+    pip install sphinx_rtd_theme stsci_rtd_theme sphinx_automodapi sphinxcontrib_programoutput
```
pllim: I don't see `sphinxcontrib_programoutput` directive used anywhere in this PR. I guess you know best, although I am concerned that usage of this extension would impede any doctest that you might have.

jaytmiller:
B7.2 final **should be corrected** to 495.
B7.1.3 final as 468 is correct.
B7.1 **should be corrected** to 422 for all .z releases.
B7.0 final is correct at 303.

Since CRDS_CONTEXT has only been formally tracked on SDN boundaries (across 2-3 strings anyway) I'd suggest adding a * to the intermediate releases and saying "estimated".

README.md
hbushouse: @nden added it in #2955 and I found out the hard way that it's now required in order to `build_sphinx`.

Issue #2670, opened 2018-10-05
Open functions that allow association files should also allow association objects


stscieisenhamer:
# Issue
Methods/functions that open association files should also take association objects. When passed an object, the return should just be that object.

bernie-simon:
There are several issues here. First, if you pass an association object to datamodels.open, it must return a datamodel, specifically a ModelContainer, just as it would if you passed it an association filename. Second, the association object returned from load_asn is a plain dict. The only way to know it's an association is to check for certain fields, like asn_id and asn_pool. Third, the filenames contained in association objects do not contain a directory. When passed a filename, the directory is copied from the directory of the filename. If there is no filename, there is nothing I can do except assume the files are in the current directory.

ccossou:
About your third comment, I think relative path works in the association. I remember trying and I think it worked. 

For me the solution is for load_asn or another function to return a ModelContainer directly. This way, when the pipeline does 
`datamodels.open(input)` it will simply return input, since it's already the correct format. 

One of the problems for me is that association formats are not a datamodel structure, and are currently not compatible with ModelContainer.

jdavies-st:
Associations should not have absolute or relative paths.  Only the filename.  If there are paths, then it is not portable.  I.e. it it tied to the directory structure of where it is generated.

Running:
```
container = datamodels.open('my_association_asn.json')
```
will return a `ModelContainer` with all the files in the association as iterables.

jdavies-st:
I think it would be fairly straightforward to modify `ModelContainer` to open the datamodels within an association object rather than an association file as it currently does, if that is what is desired.

Though it seems the main issue is an easy way to write out a generated association.  This should be a one-liner `.write()` or `save()` method.

Finally, if one just wants to load a bunch of files in a `ModelContainer`, one can do:
```
from jwst import datamodels
from glob import glob
```
container = datamodels.ModelContainer(glob('*.fits'))
```
This will only give you the list of files though, and won't populate information about the association pool or product name, things which are only available from the association JSON file.

ccossou:
About relative path, since files can be in sub-folders because they are different exposures, this did not shock me, but you might be right. 

The issue here is not really about writing an association. It's about using an association object as input for the pipeline. Right now I MUST create a .json file to read the association from it.

For instance I'd like to do:
`from jwst.associations.asn_from_list import asn_from_list
from jwst.associations.lib.rules_level3_base import DMS_Level3_Base
[...]

Issue #2892, opened 2018-11-22
datamodels should accept a pathlib.Path object as init


jdavies-st:
Currently datamodels can be given a string of a path to a FITS or ASDF file to open via `datamodels.open()` or directly passing the string to a `datamodels` subclass.  One should also be able to pass a `pathlib.Path` object (introduced in `python 3.4`).  These objects are becoming more widely used in python, specifically in `pytest`, and we should support their use.

PR #2971, opened 2018-12-19
Update ModelContainer to accept already open associations

stscieisenhamer:
Redo of #2872.

Previously ModelContainer only accepted filenames. It's been modified to work if its argument is an open association. The generic open function in util.py has also been modified to accept an open association.

Associations are plain dictionaries. The new fumction is_association looks for the fields asn_pool and asn_id and if it finds them, says the dictionary is an association.


stscieisenhamer:
Resolves #2670

jwst/datamodels/container.py
```diff
@@ -87,13 +90,16 @@ def __init__(self, init=None, persist=True, **kwargs):
...
             self.__class__ = init.__class__
             self._models = init._models
+        elif is_association(init):
+            self.from_asn(init)
         elif isinstance(init, str):
             try:
-                self.from_asn(init, **kwargs)
+                init_from_asn = self.read_asn(init)
             except (IOError):
                 raise IOError('Cannot open files.')
```
jdavies-st: Instead of triggering off the `IOError` and then re-emitting another `IOError` could we just raise the existing `IOError` along with its error message?  It would avoid having then "While handling error X another error Y was..." messages.  Same below.  I know these are not your changes, but it would clean up the errors a bit.

jwst/lib/file_utils.py
```diff
@@ -0,0 +1,20 @@
+"""File utility functions"""
+from contextlib import contextmanager
+import os
+
+
+@contextmanager
+def pushdir(directory):
```
jdavies-st: This looks like you've re-implemented `pathlib.Path` context manager?  Perhaps it would be too early for us to start using `pathlib.Path` objects in this library, though I have added it as an issue.  #2892.

jdavies-st: Yeah, feel free to ignore it for now.  I feel like getting into `pathlib` stuff in our code is a larger problem, so we can address it in some future PR.

PR #2972, opened 2018-12-19
Add back NRC_GRISM as allowed EXP_TYPE

### LINKED ISSUE #2902: Add back NRC_GRISM as an allowed EXP_TYPE value in schemas
As noted in https://jira.stsci.edu/browse/JWSTKD-287 the NRC_GRISM EXP_TYPE is still used when data are taken using the APT Engineering Mode template and one of the grisms is in the optical path. So both NRC_GRISM and NRC_WFSS are allowed EXP_TYPE values for NIRCam, but all science mode exposures will use NRC_WFSS. We need to add NRC_GRISM back to the schema enum lists.

hbushouse:
Added NRC_GRISM back to the enum lists for EXP_TYPE where it appears in all schemas. Fixes #2902.


nden:
Could you add it also to `assign_wcs.nircam.exp_type2transform`?
Does it need to be added to somewhere in associations?

hbushouse:
> Could you add it also to `assign_wcs.nircam.exp_type2transform`?
> Does it need to be added to somewhere in associations?

The problem with exposures that will still have EXP_TYPE='NRC_GRISM' is that they will only occur when the APT NIRCam Engineering Imaging template is used and there happens to be one of the grisms in the beam for the long-wave channel. Hence they aren't regular science data and certainly can't be processed in the normal way for real NRC_WFSS observations, because they won't have any associated direct images and hence no source catalog to use as input.

So I think we're going to have to treat these like other calibration/engineering exposures, which means we process only through `calwebb_detector1` and that's it (no level-2b processing). We can't try to apply `calwebb_spec2` at all, because we wouldn't have the info necessary to run `assign_wcs`.

So EXP_TYPE='NRC_GRISM' should probably be given a separate entry in the table in #1188 (separate from NRC_WFSS) to indicate that it does NOT get level-2b processing.

In that case, is it necessary to make any updates to any parts of `assign_wcs` or other areas?

hbushouse:
@stscieisenhamer see comment above about modifying #1188.

nden:
@hbushouse We can add  `'nrc_grism': not_implemented_mode,` to assign_wcs, to avoid running if someone attempts to do so outside the pipeline.

hbushouse:
> @hbushouse We can add `'nrc_grism': not_implemented_mode,` to assign_wcs, to avoid running if someone attempts to do so outside the pipeline.

OK, I'll go ahead and at that to this PR.

Issue #2821, opened 2018-10-31
Generalized Least Square fitting updates


karllark:
Generalized Least Square (GLS) fitting is an optimal pipeline update that would be useful for both better fitting of ramps using a more complete noise model (beyond  just photon and read noise) and for better cosmic ray jump detection (CRFIT) in the read noise dominated/influenced regime.  Seems like a good idea to setup tests as a start, especially as there already is GLS code in the current pipeline.

This issue is just to say that I've started doing this on my fork of the pipeline.  Comments?  Suggestions?

karllark:
Definitely for the future.  I see this as a long lead item for the optimal pipeline and have been thinking about it.  This is the start of working to define what is needed via CI tests.  And maybe some work to speed up the algorithm (TBD).

stscicrawford:
@karllark I think this is already a duplicate of this ticket: https://jira.stsci.edu/browse/JP-267

But I'll leave connect these two tickets so that work is tracked, but discussion should probably be on the Jira ticket.

karllark:
That JIRA ticket is for the whole thing, both GLS for fitting to enable better uncertainty use *and* using the CRFIT method for sensitive cosmic ray detection.

Is there a github issue for that JIRA ticket?  Or only the JIRA ticket?

philhodge:
Are there cosmic-ray jumps in your test image?

dmggh:
Karl -

It would be interesting to see how it performs on a large dataset having cosmic rays. If you need one I can supply it.

-Dave

________________________________

karllark:
I agree.  I am slowly working my way to doing such a test.  I can easily generate such myself, so no worries.  I am working on this slowly, so don't expect progress quickly.  :-)

karllark:
Results from my weekly 2 hour session on this topic (plus a bit more time this morning).

Worked through the gls_fit function.  Simplified the code a little bit, removed the use of residuals from creation of the ramp covariance matrix (not needed actually), and made a few variable changes to make it easier to understand. 

Simple timing shows that GLS takes longer for longer ramps.  This is expected given that the covariance matrix scales as (# groups)^2.   Such simple testing shows that GLS scales worse with # groups than OLS.

GLS does scales better with adding a ramp jump given that this only adds 1 parameter to the solution (i.e., slightly larger matrices) resulting is not many more calculations. 

[...]

PR #2975, opened 2018-12-19
Fixing Generalized Least Square ramp fitting

karllark:
GLS ramp fitting now works.  The same tests as for OLS fitting are done for GLS.

Main updates are to remove the use of fit residuals from the covariance matrix (not needed), remove the "extra_terms" associated with solving for non-linearities as part of the fit (not requested by the JCalWG and unlikely to be), andfix the number of iterations of the GLS algorithm to 1.  Iterations are not needed if the noise model is close enough, and it should be as long as the slope can be estimated approximately.  Currently, the 1st difference is used to estimate the slope for the noise model.  The average of the good differences should be used to be more robust (e.g., ramps jumps already detected, so this should be good enough).

The appropriate results are likely not returned, especially given the need for the uncertainties returned have the appropriate combined, photon noise alone, and read noise alone.

One caveat: GLS fails when a single group integration is passed.  The tests for this are marked as expecting to fail.

Code in the ramp_fitting/test directory to test the timing of the OLS and GLS solutions.  Maybe should delete it?  Not sure as speed is the issue with GLS (and OLS to some extent), so maybe useful.

Addresses issue #2821.


karllark:
While not needed for baseline, would be useful to have working GLS code in the pipeline.  I would like to investigate more sensitive ramp jump detection (aka CRFIT that Anderson and I worked on a few years ago, to be combined intelligently with the existing 2pt difference method).   As GLS fitting (and CRFIT ramp jump detection) is a request for the optimal pipeline, would be easiest to do my investigation using pipeline code.  And most efficient for the future -> just deliver CRFIT algorithm as pipeline code with tests!

jwst/ramp_fitting/gls_fit.py
```diff
@@ -692,52 +681,44 @@ def gls_fit(ramp_data, input_var_data,
 
     # Divide by sqrt(2) to convert the readnoise from CDS to single readout.
     rn3d = readnoise.reshape((nz, 1, 1)) * SINGLE_READOUT_RN_FACTOR
```
philhodge: The factor of 1 / sqrt(2) has already been applied by function utils.get_ref_subs.

jwst/ramp_fitting/gls_fit.py
```diff
@@ -692,52 +681,44 @@ def gls_fit(ramp_data, input_var_data,
...
-    cov += (I * (rn3d**2 / M))
+    ramp_cov += (I * (rn3d**2 / M))
```
philhodge: The factor 1 / sqrt(nframes) has already been applied by utils.get_ref_subs.

jwst/ramp_fitting/gls_fit.py
```diff
@@ -692,52 +681,44 @@ def gls_fit(ramp_data, input_var_data,
...
-                          (g3d * M)**2 / 12.)
-        cov += delta_diag
-        del delta_diag
-
-    # This is the solution:  (xT @ weight @ x)^-1 @ [xT @ weight @ y]
-    # where @ means matrix multiplication.
+    # The resulting fit parameters are
+    #  (xT @ ramp_cov^-1 @ x)^-1 * [xT @ ramp_cov^-1 @ y]
+    #  = [y-intercept, slope, cr_amplitude_1, cr_amplitude_2, ...]
+    # where * means matrix multiplication.
```
philhodge: This is OK, but it was OK before, too.  That is, @ does actually mean matrix multiplication in numpy.

jwst/ramp_fitting/gls_fit.py
```diff
@@ -692,52 +681,44 @@ def gls_fit(ramp_data, input_var_data,
...
+    # shape of `ramp_invcov` is (nz, ngroups, ngroups)
     I = I.reshape((1, ngroups, ngroups))
-    weight = la.solve(cov, I)                   # inverse of cov
+    ramp_invcov = la.solve(ramp_cov, I)
+
     del I
 
-    # temp1 = xT @ weight
+    # temp1 = xT @ ramp_invcov
+    # np.einsum use is equivalent to matrix mulitplication
```
philhodge: typo:  mulitplication --> multiplication

dmggh:
@karllark , a few questions/commsnts -
1) As you mentioned, it seems that this has not been tested in 'full step' mode, where ramp_fit_step.py is used. For pipeline use, that file would also have to be modified (currently the 'algorithm' is hard-coded to be 'ols'), as would the configuration file.

2) Has this been tested on data containing cosmic rays? I assume it hasn't, based on an earlier message of yours.

3) Why were many of the tests in test_ramp_fit.py removed ?

4) I am currently working on some additional regression tests and unit tests, including some that test for every type of ramp  (containing multiple cosmic rays, and saturated groups; 1-group integrations; 2-group integrations) that I could imagine, and checks the variances calculations of these multisegment ramps. It seems that these tests can not be used to test your GLS code.

karllark:
@dmggh: thanks for the comments.  Here are my responses to your points.

1) This is correct - ols ramp fitting is hard coded for the pipeline.  This is fine and correct at this point given gls fitting is an improvement for the optimal pipeline.  And gls is slower in most cases that ols.  It might be good to have some way to use gls instead of ols even now, but that could be somewhat risky as the gls code has not been extensively tested yet.  I am fine with waiting till the gls code is better tested (e.g., are the uncertainties correct?).

2) The gls code has been tested to fit correctly when there are CRs marked and in the ramp (near the end of the test_ramp_fit.py file).  These were already there for testing the ols method and I just extended the tests to also check the gls method. 

3) Not tests were deleted.  I did reorganize the file a bit to put the two tests that fail for gls (but pass for ols) at the top.  And the rest of the tests are now in a class (TestMethods) to have both the ols and gls methods tested (being in the class means they all have access to the parameter 'method' without having to define it for every test).  I've added a comment to explain the use of a class in the test file (learned this capability for this work).

4) It is not clear to me why the tests you are working on should not also work for the gls code (modulo point 1).  At this point, the same slopes, y-intercepts, and uncertainties (variances) should be calculated by both methods (or at least *really* close) as the same noise model is used for both.  If not, then there is a problem that needs to be fixed.

jwst/ramp_fitting/gls_fit.py
karllark: division by M removed

jwst/ramp_fitting/gls_fit.py
karllark: Removed multiplication by SINGLE_READOUT_RN_FACTOR

jwst/ramp_fitting/gls_fit.py
karllark: Fixed

jwst/ramp_fitting/gls_fit.py
karllark: You are very correct.  I had put that in before I read that @ is matrix multiplication in numpy.  Thought I'd take all of my "fixes" in this respect out, but clearly missed this one.  I've changed it back to @.

karllark:
Anymore updates needed?  (Modulo the conflict in ramp_fit.py).

karllark:
Merge me!  :-P

Issue #1604, opened 2017-12-18
Move the bounding box to the ref file for MIRI Imager


nden:
The bounding box for the MIRI imager should be attached to the distortion transform in the reference file.
The assign_wcs step will change the bounding box for subarrays but otherwise will use the one that comes with the distortion transform. This can be done as part of the next reference file delivery to CRDS.

jdavies-st:
It seems this work is done.

Using the current latest MIRI distortion reffile from CRDS:
```
In [9]: a = datamodels.DistortionModel('jwst_miri_distortion_0023.asdf')

In [10]: a
Out[10]: <DistortionModel from jwst_miri_distortion_0023.asdf>

In [11]: a.model
Out[11]: <CompoundModel137(offset_0=-4., c0_2=-12.78750038, c1_2=0.025, c0_3=-12.78750038, c1_3=0.025, c0_0_5=0.00000647, c1_0_5=-0.00020063, c2_0_5=0.00002433, c3_0_5=0.00000219, c4_0_5=-0.00000025, c0_1_5=2.82207821, c0_2_5=-0.00002873, c0_3_5=0.00000244, c0_4_5=0.00000016, c1_1_5=-0.00167748, c1_2_5=0.00000198, c1_3_5=0.00000181, c2_1_5=-0.00003313, c2_2_5=-0.00000019, c3_1_5=0.00000226, c0_0_6=0.00027793, c1_0_6=2.84373271, c2_0_6=0.00011154, c3_0_6=-0.00011824, c4_0_6=0.00000331, c0_1_6=0.00001385, c0_2_6=-0.00083572, c0_3_6=-0.00000011, c0_4_6=0.00000049, c1_1_6=-0.00001816, c1_2_6=-0.00006579, c1_3_6=0.00000018, c2_1_6=0.00000107, c2_2_6=0.0000033, c3_1_6=-0.00000026, c0_0_8=-1.54000068, c1_0_8=-0.02608358, c0_1_8=-0.00202991, c0_0_9=-7.22800207, c1_0_9=0.00202991, c0_1_9=-0.02608359, factor_13=-1., offset_14=-7.8, factor_15=60., factor_16=60.)>

In [12]: a.model.bounding_box
Out[12]: ((-0.5, 1023.5), (3.5, 1027.5))
```

jdavies-st:
Also tracked by [JP-232](https://jira.stsci.edu/browse/JP-232).

[...]

PR #2980, opened 2018-12-19
refactor subarray transform

nden:
Currently `assign_wcs.util.subarray_transform` returns a `Shift(xstart) & Shift(ystart)` transform  even for full frame observations (`Shift(0) & Shift(0)`. This transform is prepended to the WCS forward transform, e.g. the distortion. In the case of full frame observations this becomes a problem when the bounding box is attached to the transform in the reference file because then it is overwritten. In addition there are two unnecessary transforms prepended to the pipeline. 

This PR changes the logic to:
- if `subarray.xstart` and `subarray.ystart` are not `None` or `1`
  - prepend a transofrm (shift(xtsrat) & Shft(ystart)
  - change the bounding box to ((0, xstart), (0, ystart))
- if one of `xstart` or `ystart` is not None or `1` (e.g. `ystart`
  - prepend Identity(1) & Shift(ystart)
  - change the bounding box to be ((0, xstart), (0, ystart))
- if both are None or `1`:
  - do not prepend transform
  - do not change the bounding box

Related to #1604

### DIFF
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -21,7 +21,7 @@ from gwcs import utils as gwutils
 from . import pointing
 from ..lib.catalog_utils import SkyObject
 from ..transforms.models import GrismObject
-from ..datamodels import WavelengthrangeModel, DataModel
+from ..datamodels import WavelengthrangeModel, DataModel, ImageModel, CubeModel
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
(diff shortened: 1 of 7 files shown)


nden:
ping @sosey

jwst/assign_wcs/util.py
```diff
@@ -275,18 +275,39 @@ def is_fits(input):
...
 
+    Parameters
+    ----------
+    input_model : `~jwst.datamodels.DataModel`
+        Data model.
+
+    Returns
+    -------
+    subarray2full : `~astropy.modeling.core.Model` or ``None``
+        Returns a (combination of ) ``Shift`` models is a subarray is used.
```
sosey: typo: is --> if

jwst/assign_wcs/util.py
```diff
@@ -275,18 +275,39 @@ def is_fits(input):
...
+    # These quanities are 1-based
     xstart = input_model.meta.subarray.xstart
     ystart = input_model.meta.subarray.ystart
-    if xstart is None:
-        xstart = 1
-    if ystart is None:
-        ystart = 1
-    subarray2full = astmodels.Shift(xstart - 1) & astmodels.Shift(ystart - 1)
-    return subarray2full
+
```
sosey: This section is because in the data, xstart and ystart are 1-based, and if they are 1 then it's a full frame. The shift of -1 puts them in zero-based ast reference that the rest of the pipeline is using..  ....Making sure I understand the logic flow correctly?

nden: Correct. And now if both are None or 1 in FITS, then a transform is not returned.
If only one is present a Shift is returned for this one. The other transform is `Identity` instead of `Shift(0)` (because it's a little faster).

PR #2983, opened 2018-12-20
Use float64 in coron3 pipeline

### LINKED ISSUE #2893: coron3 pipeline gives unstable results

jdavies-st:
This is a small change to use double-precision floats where they might be needed in key areas of `Coron3Pipeline`.

This also updates the regression test for this module (which is currently failing).

Hopefully the changes here will make the code more robust to upstream changes in numpy, and make the test more reliable.

Seems to resolve #2893.

### DIFF
--- a/jwst/coron/imageregistration.py
+++ b/jwst/coron/imageregistration.py
@@ -222,12 +222,17 @@ def align_models(reference, target, mask):
         # Compute the shifts of the PSF ("target") images relative to
         # the science ("reference") image in this integration, and apply
         # the shifts to the PSF images
-        d, shifts = align_array(reference.data[k], target.data, mask.data)
+        d, shifts = align_array(
+            reference.data[k].astype(np.float64),
+            target.data.astype(np.float64),
+            mask.data)
         output_model.data[k] = d
 
         # Apply the same shifts to the PSF error arrays, if they exist
         if target.err is not None:
-            output_model.err[k] = fourier_imshift(target.err, -shifts)
+            output_model.err[k] = fourier_imshift(
+                target.err.astype(np.float64),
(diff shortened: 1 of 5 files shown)


jdavies-st:
Truth files in our regression test have now been updated.  Let's see how stable this is moving forward.

jdavies-st:
Address #1988.

PR #2984, opened 2018-12-20
Don't rename failed test outputs; segregate truth files in test dir

jdavies-st:
This changes the behavior of our regression tests by turning off renaming the test output to be the new truth file (which currently overwrites the actual truth file).  This also adds a segregated `truth` directory inside the test output dir.

This will make comparing the test output with the truth much easier, as both will be located in the test output directory for failed tests.

Also cleans up the test log to give abspaths to output and truth files, so they can easily be found in the output traceback.

### DIFF
--- a/jwst/tests/compare_outputs.py
+++ b/jwst/tests/compare_outputs.py
@@ -1,7 +1,6 @@
 import copy
 from datetime import datetime
 import os
-import shutil
 from difflib import unified_diff
 from io import StringIO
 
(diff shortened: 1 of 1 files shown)


jdavies-st:
Thanks Dave.  2 and 3 are addressed with this PR.  Full paths and verbiage for `Failed comparison: ` now are there.  As for 1, that's actually a bug I just fixed in `astropy.io.fits` https://github.com/astropy/astropy/pull/8295, so hopefully that will show up in our comparisons once we stop making custom `HDUList` objects, we we can do for all the tests now.  We do that for some already.  I'll make this change on our side a separate PR.

PR #2985, opened 2018-12-20
Initialize PyArrayObject in carrutils.c

### LINKED ISSUE #2982: PyArray_NewSimple does not initialize returned arrays
@jdavies-st reported random/intermittent crashes in `tweakreg` regression tests which appear to happen after https://github.com/spacetelescope/jwst/pull/2948

Upon further investigation, I believe the issue could be due to the fact that the `PyArray_SimpleNew` does not return an _initialized_ array (unlike the `PyArray_FromDims` which it has replaced). Therefore a separate array initialization must be performed, i.e., using [`PyArray_FILLWBYTE(new_array, 0)`](https://docs.scipy.org/doc/numpy-1.13.0/reference/c-api.array.html#c.PyArray_FILLWBYTE).

CC: @stsci-hack This could be the cause of `tweakwcs` intermittently producing different results because the same change `PyArray_FromDims`  -> `PyArray_SimpleNew` was done in that package - see https://github.com/spacetelescope/tweakwcs/pull/30

jdavies-st:
This should fix the problems seen in the following regression test:

https://boyle.stsci.edu:8081/job/RT/job/JWST/lastCompletedBuild/testReport/jwst.tests_nightly.general.nircam.test_nrc_image3_1/TestImage3Pipeline1/test_image3_pipeline2/

Fixes #2982.

### DIFF
--- a/src/tweakreg/carrutils.c
+++ b/src/tweakreg/carrutils.c
@@ -83,6 +83,7 @@ arrxyzero(PyObject *obj, PyObject *args)
   dimensions[0] = (integer_t)(searchrad*2) + 1;
   dimensions[1] = (integer_t)(searchrad*2) + 1;
   ozpmat = (PyArrayObject *)PyArray_SimpleNew(2, dimensions, NPY_DOUBLE);
+  PyArray_FILLWBYTE(ozpmat, 0);
   if (!ozpmat) {
     goto _exit;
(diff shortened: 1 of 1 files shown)


jdavies-st:
Note: regression test above may need to have its truth file updated.

PR #2986, opened 2018-12-20
collect_pipeline_cfgs now won't import all pipeline steps

### LINKED ISSUE #2976: collect_pipeline_configs imports drizzle
Is there a reason the collect_pipeline_configs script needs to import the drizzle packages?  I didn't have it installed in my jwst environment, but wonder if it's truley needed to load the configs....I think it's getting called on the import of the jwst package, which is used to find where the config files are located.

Wondering if we can use something more like this to avoid the actual import of jwst

> import importlib
> jwst = importlib.find_loader("jwst")
> jwst.path

jdavies-st:
Resolves #2976

### DIFF
--- a/jwst/pipeline/collect_pipeline_cfgs.py
+++ b/jwst/pipeline/collect_pipeline_cfgs.py
@@ -2,17 +2,14 @@
 
 import os
 import shutil
+from glob import glob
+from importlib.util import find_spec
 
 
 def collect_pipeline_cfgs(dst='./'):
-    # find installed pipeline *.cfg files
-    import jwst.pipeline as p
+    """Copy step and pipeline .cfg files to destination"""
+    os.makedirs(dst, exist_ok=True)
 
-    if not os.path.exists(dst):
-        os.makedirs(dst)
-
-    cfg_dir = p.__path__[0]
-    for i in os.listdir(cfg_dir):
-        if i.endswith('.cfg'):
-            cfg = os.path.join(cfg_dir, i)
+    cfg_dir = find_spec('jwst.pipeline').submodule_search_locations[0]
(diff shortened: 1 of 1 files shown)


jdavies-st:
I guess this technically still imports `jwst`, but it doesn't import `jwst.pipeline`, which is where the problem lay earlier.  Is this OK?

sosey:
What you could do it search on "jwst" and we know there's a pipeline dir, in this case jwst won't be imported, it looks on the path for the jwst package, if it's not there it returns None.  You could then append the pipeline dir

jwst/pipeline/collect_pipeline_cfgs.py
```diff
@@ -2,17 +2,14 @@
...
+    os.makedirs(dst, exist_ok=True)
 
-    if not os.path.exists(dst):
-        os.makedirs(dst)
-
-    cfg_dir = p.__path__[0]
-    for i in os.listdir(cfg_dir):
-        if i.endswith('.cfg'):
-            cfg = os.path.join(cfg_dir, i)
+    cfg_dir = find_spec('jwst.pipeline').submodule_search_locations[0]
```
sosey: so  this would become
`cfg_dir = find_spec('jwst').submodule_search_locations[0] + '/pipeline'`

sosey: this would have to be updated if we moved the location of pipeline, but chances are slim? What do you think?

jdavies-st: A nice side effect of this is that `collect_pipeline_cfgs` is now very fast.  🏎

PR #2988, opened 2018-12-20
Move update of meta info to the MultiSlitModel instead of the SlitModels it contains

### LINKED ISSUE #2978: Output extract_2d multislit datamodel should be instantiated from input model
The multislit datamodel is empty when instantiated in grisms.extract_grism_objects:
https://github.com/spacetelescope/jwst/blob/3c5068be81929818c2652d4f7d3ae8fe8292ba51/jwst/extract_2d/grisms.py#L304
When the resulting model is returned from extract_2d it's missing meta information (individual `slit` models do have meta information). This becomes a problem later on when the multislit model is passed to the photometry step and CRDS tries to fetch the appropriate reference file:
`crds.core.exceptions.CrdsError: Missing 'META.INSTRUMENT.NAME' keyword in header for determining instrument.`
It's easily fixed by instantiating the multislit model from the input image datamodel.

sosey:
This fixes #2978 and closes #2979 in favor of using the update method.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -61,6 +61,8 @@ extract_1d
 
 extract_2d
 ----------
+- Moved the update of meta information to the MultiSlitModel instead of the
+  SlitModels that compose it. [#2988]
 
 firstframe
 ----------
--- a/jwst/extract_2d/grisms.py
+++ b/jwst/extract_2d/grisms.py
@@ -116,7 +116,7 @@ def extract_tso_object(input_model,
         lmin, lmax = range_select.pop()
 
         # create the order bounding box
-        source_xpos = input_model.meta.wcsinfo.crpix1 - 1  # remove fits 
+        source_xpos = input_model.meta.wcsinfo.crpix1 - 1  # remove fits
         source_ypos = input_model.meta.wcsinfo.crpix2 - 1  # remove fits
         transform = input_model.meta.wcs.get_transform('full_detector', 'grism_detector')
         xmin, ymin, _ = transform(source_xpos,
@@ -302,6 +302,7 @@ def extract_grism_objects(input_model,
 
     log.info("Extracting grism objects into MultiSlitModel")
     output_model = datamodels.MultiSlitModel()
+    output_model.update(input_model)
 
     # One WCS model can be used to govern all the extractions
     # and in fact the model transforms rely on the full frame
@@ -376,7 +377,6 @@ def extract_grism_objects(input_model,
                 subwcs.set_transform('grism_detector', 'detector', tr)
 
                 new_slit = datamodels.SlitModel(data=ext_data, err=ext_err, dq=ext_dq)
-                new_slit.update(input_model)
                 new_slit.meta.wcsinfo.spectral_order = order
                 new_slit.meta.wcs = subwcs
 


sosey:
Yes, I was surprised our tests didn't catch  this, but we only have a single niriss exposure in DMS to work with, nothing for nircam that I know of.

I had talked with @nden  about this, and she said that the adopted approach was to store all the meta information in MultiSlitModel, and just the additional attributes that each slit needs in the SlitModels that compose it. The update method is preferred so that the wcs information is not copied.

jdavies-st:
Our current WFSS regression tests is xfailed.

https://github.com/spacetelescope/jwst/blob/1789741e647bba4ed6c2146f695acc80af27af71/jwst/tests_nightly/general/niriss/test_nis_wfss_spec2.py#L14-L35

hbushouse:
But having the test set to xfail just means that we expect failures due to differences in the results, right? Not an outright error/crash situation, due to something like CRDS bestref failing because it can't get its hands on model.meta.instrument.name, which is what this PR is supposed to be fixing.

jdavies-st:
Well, if this PR fixes the pipeline and the test can now run without crashing, then yeah, let's get rid of the `xfail`.

hbushouse:
We claimed to have delivered processing of WFSS data in B7.2, up through the completion of `calwebb_spec2`. Have we now determined that's not actually true? Did we never have any data running all the way through without error? I know @sosey was iterating back and forth with the teams to get simulated data from them that was consistent with and worked in our pipeline. Did we never actually get anything that works?

sosey:
just for reference, this is the test image we have for the niriss grism, with a big stretch, with a blue dot where the catalog says a source should be, and a red box where the extract_2d says the first order spectra should be located. It doesn't matter too much at this point if there's no actual source flux since the catalog was made by hand just to say there was an object located somewhere. We could move the center of the object so that we can get all the orders falling on the detector (the code current defaults to only extracting order 1 in this filter, but we can add tests to fudge that)

screen shot 2018-12-21 at 2 49 59 pm

screen shot 2018-12-21 at 3 13 59 pm


a quick check of the extraction shows reasonable results that are on the detector:
screen shot 2018-12-21 at 3 02 16 pm

screen shot 2018-12-21 at 3 14 30 pm

and Spec2Pipeline completes without Error, just warnings: 

> WARNING - Keyword CDP_BAD_BACKGROUND does not correspond to an existing DQ mnemonic, so will be ignored
> WARNING - Keyword CDP_INTERNAL_FLAT does not correspond to an existing DQ mnemonic, so will be ignored
> 2018-12-21 15:01:02,087 - stpipe.Spec2Pipeline.srctype - WARNING - EXP_TYPE NIS_WFSS not applicable to this operation
2018-12-21 15:01:02,087 - stpipe.Spec2Pipeline.srctype - WARNING - Step will be skipped
> 2018-12-21 15:01:02,615 - stpipe.Spec2Pipeline.extract_1d - WARNING - Can't determine dispaxis from the WCS.
2018-12-21 15:01:02,616 - stpipe.Spec2Pipeline.extract_1d - WARNING - The dispersion direction couldn't be determined, so skipping ...
2018-12-21 15:01:02,617 - stpipe.Spec2Pipeline.extract_1d - INFO - Step extract_1d done
2018-12-21 15:01:02,617 - stpipe.Spec2Pipeline - INFO - Finished processing product jw87600017001_02101_00002_nis
2018-12-21 15:01:02,617 - stpipe.Spec2Pipeline - INFO - Ending calwebb_spec2
2018-12-21 15:01:02,953 - stpipe.Spec2Pipeline - INFO - Saved model in jw87600017001_02101_00002_nis_cal.fits
2018-12-21 15:01:02,953 - stpipe.Spec2Pipeline - INFO - Step Spec2Pipeline done

The disp axis error is still there, it's possible that the x1d code is looking for a different number of input parameters that it's getting from x2d grisms, I'll look into that and submit a separate PR if needed, but I think we are good to merge this and turn off the xfail since x1d exits clean if it can't find the dispersion direction. I'll submit the xfail pr separately.

sosey:
@hbushouse just for completeness, here's the sim detections with the current code, merge this PR if you're satisfied

screen shot 2018-12-21 at 5 27 29 pm
screen shot 2018-12-21 at 5 29 42 pm

PR #2992, opened 2018-12-21
Update README.md with RT test running info

jdavies-st:
- ci_watson is now included in the dependencies that get installed
   with the `jwst` conda package
 - It is more reliable to point TEST_BIGDATA to our artifactory
   server, as the state of `/grp/jwst/ssb/jwst_test_data` may or may
   not be in sync.  Currently.
 - Add info about running specific regression tests and using `--basetemp`

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ Unit Tests
 
 Unit tests can be run via `pytest`.  All tests need a the `ci_watson` pytest plugin to run.
 
-    pip install requests_mock ci_watson
+    pip install requests_mock
     pytest jwst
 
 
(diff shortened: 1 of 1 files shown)


stscicrawford:
Are there unit tests that rely on ci_watson?   It's okay if that is currently the case, but I think longer term, we do want the unit tests to be able to be run anywhere.

jdavies-st:
Currently the whole test infrastructure relies on `ci_watson`.  I would like that not to be the case too.  Perhaps @pllim has some thoughts about appropriate setup for this?

Basically, if you use the `ci_watson` @pytest.mark.bigdata` and `@pytest.mark.slow`, the discovery process will find those tests regardless if they will be run, so one needs these defined somewhere.  Currently, they are only defined in `ci_watson`.

Currently, no unit tests actually use the `ci_watson` fixtures or functionality, but that could change, as we would like to get a lot of the test data for `associations` out of the repository and into Artifactory (or Box).  Same with some the tests in `jwst.pipeline`.  Currently, these tests are marked for `--bigdata` and `--slow`, but we don't currently run them in either per-PR basis or nightly.

stscicrawford:
Ah in that case, depending on ci_watson is fine.  I was afraid there was a dependency on data in artifactory that was causing that, and that is really what we should avoid so that anyone can run the unit tests. 

If there isn't one already, can you open an issue on refactoring the association tests?

pllim:
What is the concern of having `ci-watson` as dependency? It is lightweight. The original idea was to have common CI functionality being in a library that different packages at STScI can use.

pllim:
> so that anyone can run the unit tests

Provided that "anyone" has the bandwidth to download the whole test suite of data to local machine, it is as simple as defining `TEST_BIGDATA` to point to your local "installation" of the data.

stscicrawford:
@pllim My concern was the dependency on ci_watson indicated a need for data in artifactory for the unit tests.  That does not appear to be the case and so my concern was mistaken.

jdavies-st:
Well, it turns out getting `ci_watson` as a dependency in the astroconda-dev `jwst` package is a bit of a problem, so I'm just putting it back as a `pip install`.

Yeah, `ci_watson` is lightweight, but I would prefer it be installed automatically as a hard dependency of the package like `pytest` currently is.  Right now, that is not done.

Also our astroconda package for `ci_watson` is called `ci-watson`.
```
$ conda install ci_watson
Solving environment: failed

PackagesNotFoundError: The following packages are not available from current channels:
```
But yes, `ci_watson` is currently required even to run the unit tests.  Is this OK by everyone?

pllim:
`pip install` (especially useful for pulling dev version) or `conda install ci-watson -c astroconda`

pllim:
p.s. I believe it is a dash, not underscore.

PR #2999, opened 2018-12-26
Move regression from unit test area and use the ci-watson framework

### LINKED ISSUE #2445: move "non-regression" regressions into tests_nightly

stscieisenhamer:
Move various tests that are currently in the unit test area out to the regression area and convert to ci-watson and artifactory.

Resolves #2445.


jdavies-st:
I kicked off a Jenkins RT build against this PR branch:

https://boyle.stsci.edu:8081/job/RT/job/jdavies-dev/69

Let's see how it goes.

jdavies-st:
From the regression test run on this PR:
```
[jwst/3.6] ====== 4 failed, 152 passed, 16 xfailed, 412 warnings in 17134.31 seconds ======
```
which compared to the current tests on master:
```
[jwst/3.6] ====== 4 failed, 119 passed, 1 xfailed, 301 warnings in 12260.85 seconds =======
```
Looks good!  15 of the new tests added are set to `xfail`.  Is this expected?

Our test runtime increased from 3.4 hours to 4.7 hours.  That's a 40% increase in runtime while adding 40% more tests.  172 up from 124.

jdavies-st:
Yes, I see that the tests in `jwst.tests_nightly.general.pipelines.test_nis_soss_tso3` take up 1 hour, 2 full runs of the `tso3` pipeline.  Our existing `tso3` test `jwst.tests_nightly.general.niriss.test_tso3` also takes the exact same amount of time and uses the exact same dataset as input.  I think we have some duplication here.  Would be good to reduce these 3 tests down to 1 if possible.

I also notice that you organized these tests into a new subdir `tests_nightly/general/pipelines/` instead of into `tests_nightly/general/<INSTRUMENT>/` as previously.  I believe we had decided not to use a `pipelines` subdir back when we did the reorg, right?  If we want to revisit that decision, that's fine, but it would be good that we do this deliberately rather than accidentally.  =)

stscieisenhamer:
Went through the tests and determine that nearly all are duplicated. The one nirspec that wasn't clearly duplicated has been moved to the `nirspec` folder.

jwst/tests/base_classes.py
```diff
@@ -163,3 +213,48 @@ def raw_from_asn(asn_file):
...
+    glob: str
+        The file name match criterion
+
+    Returns
+    -------
+    url_paths: [str[, ...]]
+        Full URLS that match the glob criterion
+    """
+    path = requests.get(url).text
+    soup = BeautifulSoup(path, 'html.parser')
```
jdavies-st: Would be nice if we can use Artifactory's search API here instead of scraping our Artifactory web server.  I'm investigating and testing.

stscieisenhamer: you did it. resolved 👍

jdavies-st: Commit added to use Artifactory's search API.

Issue #2951, opened 2018-12-17
Update MIRI LRS WCS transforms to include inverse functions


stscijgbot:
Issue [JP-484](https://jira.stsci.edu/browse/JP-484) was created by Howard Bushouse:

The MIRI LRS WCS transforms do not currently include inverse functions. There is also an updated reference file in the MIRI CDP-7 that needs to be incorporated.

hbushouse:
Roll-up for several other existing issues related to MIRI LRS WCS:
#268 
#1604 
#1809
#2825

stscijgbot:
Comment by Howard Bushouse: See [#268|https://github.com/spacetelescope/jwst/issues/268] and [#2825|https://github.com/spacetelescope/jwst/issues/2825]

stscijgbot:
Comment by Steven Crawford: Assigning ticket to [~dlaw] as we will be waiting on delivery of updated transforms for further work on this.    

[~dlaw] please update ticket if you won't be able to deliver an update by 11 February 2019 with an updated estimate of delivery time so we can plan for when to complete this work.

stscijgbot:
Comment by David Law: I'm planning to submit a PR for this the week of Feb 4

stscijgbot:
Comment by David Law: Modifications now submitted via https://github.com/spacetelescope/jwst/pull/3106

hbushouse:
Reopened and closed Jira ticket to fix tracking, which in turn reopened this issue.

stscijgbot:
Comment by Howard Bushouse: Fixed by [https://github.com/spacetelescope/jwst/pull/3106] and [https://github.com/spacetelescope/jwst/pull/3324|https://github.com/spacetelescope/jwst/pull/3106]

Issue #2825, opened 2018-11-01
update MIRI LRS WCS spatial transform


nden:
The WCS transform for the MIRI LRS includes a spatial transform from detector to world using the Imager distortion. However the WCS pipeline does not include an intermediate frame at V2V3. Instead it goes directly to sky. 
An intermediate V2V3 frame is needed in order to get extract_1d to use the same algorithm as other steps.

Note: The spatial transform includes the distortion models but not the filter offset correction.

skendrew:
See also https://github.com/spacetelescope/jwst/issues/268 - does this help?

nden:
An intermediate frame was added  in #2894.
The remaining work includes composing an inverse transform. The transform is a `TabularModel` and an inverse model can be easily added in the 1D case. This can be done initially in the pipeline but should be eventually moved to modeling.

@hbushouse

hbushouse:
Now being tracked in JIRA in https://jira.stsci.edu/browse/JP-484

nden:
Resolved in #3001, #2951 and #3324

PR #3001, opened 2018-12-28
rework spatial transform for the LRS modes

nden:
The MIRI LRS spatial transform is computed for one row,  `x, zero_point_y`, and replicated for all other rows. It was represented as a `Tabular2D` model. This PR changes the representation to use analytical models. It uses the distortion model but passes `zero_point_y` for all y values in the forward transform, e.g.
```
forward = Identity(1) & Const1D(zero_point_y) | distortion
``` 
It assigns a similar inverse which uses `Const1D(zero_point_v3)` as second input.

Note: This assigns an inverse to the spatial transform but not to the spectral. @drlaw1558 has an upcoming PR which changes the representation of the wavelength solution from a lookup table to an analytical model and introduces an inverse of the spectral part of the WCS.

Anyone want to review this?

### DIFF
--- a/jwst/assign_wcs/assign_wcs.py
+++ b/jwst/assign_wcs/assign_wcs.py
@@ -66,11 +66,12 @@ def load_wcs(input_model, reference_files={}):
         exclude_types = ['nrc_wfss', 'nrc_tsgrism', 'nis_wfss',
                          'nrs_fixedslit', 'nrs_msaspec',
                          'nrs_autowave', 'nrs_autoflat', 'nrs_lamp',
-                         'nrs_brightobj', 'mir_lrs-fixedslit', 'mir_lrs-slitless',
-                         'nis_soss']
+                         'nrs_brightobj', 'nis_soss']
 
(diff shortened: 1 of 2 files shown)


nden:
Fixes parts of #2825 and #2951

nden:
```
MD5MismatchError: Conda detected a mismatch between the expected content and downloaded content
for url 'http://ssb.stsci.edu/astroconda-dev/linux-64/astropy-4.0.0.dev11880 ga671d49-py36_0.tar.bz2'.
  download saved to: /opt/conda/pkgs/astropy-4.0.0.dev11880+ga671d49-py36_0.tar.bz2
  expected md5 sum: c911b3c001699b13bc1711ae236f8790
  actual md5 sum: ba9411565cabb4a47acbb915ade5d512

```

This is the error on Jenkins. The same error was on Travis. Posting it here because it's the first time I see an error like this.

Is there a way to restart Jenkins tests without privilege? Closing and opening the PR did not do it.

PR #3005, opened 2018-12-31
Reverse order of RELSENS arrays if the wavelength is not increasing

### LINKED ISSUE #3004: extract_1d needs to check for decreasing wavelengths in RELSENS
The `extract_1d` step interpolates within the RELSENS table using `np.interp` without checking whether the wavelengths in the RELSENS table are increasing.  `np.interp` requires that its second argument (the wavelength column from the RELSENS table) be increasing.  The `interpolate_response` function in extract.py should be modified to check the order of the RELSENS wavelengths, and to reverse the order of wavelength and response if wavelength is not increasing.

philhodge:
Function `interpolate_response` was modified.  Resolves #3004.

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -2421,6 +2421,13 @@ def interpolate_response(wavelength, relsens, verbose):
     if bad:
         raise ValueError("Found NaNs in RELSENS table.")
 
+    # np.interp requires that wl_relsens be increasing.
+    if wl_relsens[-1] < wl_relsens[0]:
+        if verbose:
+            log.warning("The wavelength column in RELSENS was decreasing.")
+        wl_relsens = wl_relsens[::-1].copy()
+        resp_relsens = resp_relsens[::-1].copy()
+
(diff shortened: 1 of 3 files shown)


jwst/extract_1d/extract.py
```diff
@@ -2421,6 +2421,13 @@ def interpolate_response(wavelength, relsens, verbose):
     if bad:
         raise ValueError("Found NaNs in RELSENS table.")
 
+    # np.interp requires that wl_relsens be increasing.
+    if wl_relsens[-1] < wl_relsens[0]:
+        if verbose:
+            log.warning("The wavelength column in RELSENS was decreasing.")
+        wl_relsens = wl_relsens[::-1].copy()
+        resp_relsens = resp_relsens[::-1].copy()
```
nden: I don't think you need to make a copy here. Reversing a list or an array already makes a copy.

nden:
@philhodge One minor comment, other than that looks good. Feel free to merge.

Issue #3013, opened 2019-01-04
Make bounding boxes from shape clearer to understand and implement


sosey:
This started from a discussion with @philhodge about the bounding boxes coming out of the wfss pipeline....but after discussion with @nden it seems to be unclear in general.

When creating bounding boxes:
* attached to a wcs object, the tuple should be a range in (x, y) order
* attached to a transform (for example a wcs frame transform), the tuple should be a range in (y, x) order

There are three functions in `assign_wcs.util` to help with this, but they are misleading:
* https://github.com/spacetelescope/jwst/blob/master/jwst/assign_wcs/util.py#L665
* https://github.com/spacetelescope/jwst/blob/master/jwst/assign_wcs/util.py#L683
* https://github.com/spacetelescope/jwst/blob/master/jwst/assign_wcs/util.py#L713

We need to update the functions to be clearer what their input and use should be

nden:
Resolved by #3014

PR #3014, opened 2019-01-04
make bounding_box clearer

sosey:
This PR creates a new function in `assign_wcs.util` called `bounding_box_for_transform` which returns the correctly arranged tuple for attaching to a transform.

It also renames the current `assign_wcs.util.bounding_box_from_shape` to `bounding_box_for_wcs`, which still returns the tuple in (x,y) order.

The names made sense to me, but this is up for discussion.  see #3013

### DIFF
--- a/jwst/extract_2d/grisms.py
+++ b/jwst/extract_2d/grisms.py
@@ -191,7 +191,7 @@ def extract_tso_object(input_model,
             output_model.err = ext_err
             output_model.dq = ext_dq
             output_model.meta.wcs = subwcs
-            output_model.meta.wcs.bounding_box = util.bounding_box_from_shape(ext_data.shape)
+            output_model.meta.wcs.bounding_box = util.bounding_box_for_wcs(ext_data.shape)
             output_model.meta.wcs.crpix2 = 34  # update for the move, vals are the same
             output_model.meta.wcsinfo.spectral_order = order
             output_model.name = str('TSO object')
@@ -373,7 +373,7 @@ def extract_grism_objects(input_model,
                 ext_err = input_model.err[ymin: ymax + 1, xmin: xmax + 1].copy()
                 ext_dq = input_model.dq[ymin: ymax + 1, xmin: xmax + 1].copy()
 
-                tr.bounding_box = util.bounding_box_from_shape(ext_data.shape)
+                tr.bounding_box = util.bounding_box_for_transform(ext_data.shape)
                 subwcs.set_transform('grism_detector', 'detector', tr)
 
                 new_slit = datamodels.SlitModel(data=ext_data, err=ext_err, dq=ext_dq)
--- a/jwst/extract_2d/grisms.py
+++ b/jwst/extract_2d/grisms.py
@@ -191,7 +191,7 @@ def extract_tso_object(input_model,
             output_model.err = ext_err
             output_model.dq = ext_dq
             output_model.meta.wcs = subwcs
-            output_model.meta.wcs.bounding_box = util.bounding_box_for_wcs(ext_data.shape)
+            output_model.meta.wcs.bounding_box = util.wcs_bbox_from_shape(ext_data.shape)
             output_model.meta.wcs.crpix2 = 34  # update for the move, vals are the same
             output_model.meta.wcsinfo.spectral_order = order
             output_model.name = str('TSO object')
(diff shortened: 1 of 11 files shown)


sosey:
> Would bounding_box_from_shape_for_wcs() be too long?

I'm not opposed to it, but it does start to creep on the line width limits if you're using it practice. We should wait for @nden to tally her preference for naming and implementation.

mcara:
This is so confusing...

mcara:
What about APE 14?

mcara:
Please remind in #3015 when this PR is merged. I can switch to this new function later.

jwst/extract_2d/grisms.py
```diff
@@ -373,7 +373,7 @@ def extract_grism_objects(input_model,
                 ext_err = input_model.err[ymin: ymax + 1, xmin: xmax + 1].copy()
                 ext_dq = input_model.dq[ymin: ymax + 1, xmin: xmax + 1].copy()
 
-                tr.bounding_box = util.bounding_box_from_shape(ext_data.shape)
+                tr.bounding_box = util.bounding_box_for_transform(ext_data.shape)
```
nden: Is `input_model` here always an `ImageModel`? Can it be a `CubeModel` (for tso observations??

sosey: No, there's a separate function just for TSO extractions because they dont require a catalog
https://github.com/spacetelescope/jwst/blob/master/jwst/extract_2d/grisms.py#L18

nden:
@sosey @jdavies-st How about these names:
```
transform_bbox_from_shape()
was_bbox_from_shape()
transform_bbox_from_datamodel()
```

sosey:
did you mean `wcs_bbox_from_shape`? If so, I'm okay with those names too.

mcara:
How about a single function `to_bbox(shape)`?

nden:
@mcara The issue is that the order of the bounding box on the WCS object is (x, y) while on a transform it is (y, x). We are trying to come up with names which show what object it is supposed to be applied to. In addition the input to this function can be a shape or a data model.

mcara:
IMO, it should be clearly documented what `to_bbox(shape)` would do. Figuring out when to add `[::-1]` to the result cannot be a difficult exercise. For me, it is more difficult to read documentation for six functions (and actually even to know that I should not stop reading at the first function and that there are five more functions that I need to look at).

nden:
@mcara In principle I am not opposed to what you are saying. However, we started with one function and it was problematic so this is an attempt to make it clearer by having two functions.

jdavies-st:
So this PR seems to have caused a regression test failure:

https://boyle.stsci.edu:8081/job/RT/job/JWST/98/testReport/junit/jwst.tests_nightly.general.niriss.test_nis_wfss_spec2/TestSpec2Pipeline/test_nis_wfss_spec2/

Any idea what is going on?

sosey:
quick guess, looks like old data in the ref comparison? The SLITID should exist...but also, there was this PR https://github.com/spacetelescope/jwst/pull/2988 where we changed the meta population to be just in the primary, with slit and wcs in the extensions.

mcara:
I have accidentally "overheard" discussions in this PR... and I looked at the code (I know I should have done it before). With regard to [transform_bbox_from_datamodel](https://github.com/sosey/jwst/blob/8fbaabd02e4964a8a2966e48b81f9bba633ac737/jwst/assign_wcs/util.py#L723). It is not clear why there is a check of instance at all: first of all `shape = input_model.data.shape` already assumes a specific structure and after that we should be using plain duck typing:

```python
def transform_bbox_from_datamodel(input_model):
    return ((-0.5, shape[-2] - 0.5), (-0.5, shape[-1] - 0.5))
```
Why all that complication and performance hit?

philhodge:
For the x1d file, the latest output (Jan 15, 20:29) has an extracted 1-D spectrum with length 9, but the "truth" file contains only a primary HDU and an ASDF extension.  The latest output looks correct:
print(wl)
[2.3        2.3        2.3        2.3        2.29822985 1.5586398  0.81904974 0.75       0.75      ]
net.mean()
4.1189449859990015
flux.mean()
2.5873096610898467

jdavies-st:
Thanks Phil.  So perhaps this means that before `extract_1d` was not finding a spectrum to extract?  Is that an expected outcome of this PR, that now it does?

philhodge:
Changing the name of the function for getting the bounding box should not have had any effect, unless the wrong function was called previously.  I do know that at one time (as recently as Jan 14, I think) the wcs `bounding_box` had its elements reversed, and that could easily have the effect that `extract_1d` would fail to find the dispersion direction.
I think the x1d truth file must have been taken as the output of `extract_1d` for a case where `extract_1d` was not able to determine the dispersion direction.  In that case, the extraction is skipped, which would explain why the file had no EXTRACT1D extension at all, since there was only one slit in the input _cal.fits file.  I merged #3032 this morning, and that should help `extract_1d` find the dispersion direction more reliably.

jdavies-st:
Mihai, I like your suggestion.  I believe originally one of the functions did exactly that (or perhaps an unmerged PR of mine).  Do a PR please!  ;-)

sosey:
I believe the wrong function was being called in the WFSS code, and me having to change it was what sparked the conversation about bounding_box

PR #3015, opened 2019-01-04
Rely on bounding_box when available

### LINKED ISSUE #3009: calwebb_image3.tweakreg to better handle runaway processing
Issue [JP-498](https://jira.stsci.edu/browse/JP-498) was created by Jonathan Eisenhamer:

# Issue

During SDP runs, proposal jw00752, observation 8, a MIRI mosaic, while being processed by `calwebb_image3`, was running for over a week. Previous runs had noted 24+ hour processing before being manually killed off.

The data was provided by the OTB and is known to be bad. However, it was been requested that the pipeline be able to detect and handle bad data/runaway processing.

### LINKED ISSUE #3012: Make tweakwcs use WCS's bounding box instead of image size
This is the real issue that needs to be fixed instead of fixing "runaway processing" - https://github.com/spacetelescope/jwst/issues/3009

I do not know whether to classify this as a bug or an enhancement. The issue is that tweakreg attempts to compute the bounding box of an image based on the image size/shape. For unknown to me reasons (yet), MIRI's WCS returns `nan` for a coordinate such as `(-0.5, -0.5)` (corner of the first pixel).

CC: @stscieisenhamer @jdavies-st @nden

mcara:
Resolves #3009. Resolves #3012 

CC: @stscieisenhamer

### DIFF
--- a/jwst/tweakreg/wcsimage.py
+++ b/jwst/tweakreg/wcsimage.py
@@ -605,7 +605,7 @@ class WCSImageCatalog():
             ninty = 2
         else:
             nintx = max(2, int(np.ceil((hx - lx) / stepsize)))
-            ninty = max(2, int(np.ceil((ny - ly) / stepsize)))
+            ninty = max(2, int(np.ceil((hy - ly) / stepsize)))
 
         xs = np.linspace(lx, hx, nintx, dtype=np.float)
         ys = np.linspace(ly, hy, ninty, dtype=np.float)[1:-1]
--- a/jwst/tweakreg/wcsimage.py
+++ b/jwst/tweakreg/wcsimage.py
@@ -592,17 +592,23 @@ class WCSImageCatalog():
         if self.wcs is None:
             return
 
-        ny, nx = self.imshape
+        if self.wcs.bounding_box is None:
+            lx = -0.5
+            ly = -0.5
+            hx = self.imshape[1] - 0.5
+            hy = self.imshape[0] - 0.5
+        else:
+            ((lx, hx), (ly, hy)) = self.wcs.bounding_box
 
         if stepsize is None:
             nintx = 2
             ninty = 2
         else:
-            nintx = max(2, int(np.ceil((nx + 1.0) / stepsize)))
-            ninty = max(2, int(np.ceil((ny + 1.0) / stepsize)))
+            nintx = max(2, int(np.ceil((hx - lx) / stepsize)))
+            ninty = max(2, int(np.ceil((ny - ly) / stepsize)))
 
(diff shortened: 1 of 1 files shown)


jwst/tweakreg/wcsimage.py
```diff
@@ -592,17 +592,23 @@ def _calc_chip_bounding_polygon(self, stepsize=None):
         if self.wcs is None:
             return
 
-        ny, nx = self.imshape
+        if self.wcs.bounding_box is None:
+            lx = -0.5
```
jdavies-st: You should use the existing function that does this in
```
from jwst.assign_wcs.util import bounding_box_from_shape

self.wcs.bounding_box = bounding_box_from_shape(self.imshape)
```

jdavies-st: Do note that you are in a race with @sosey, as she has a PR that will change the name of this function in #3014.  So maybe wait to for her to merge before this is done.

sosey: muahahahahaha

mcara: 😈

jwst/tweakreg/wcsimage.py
```diff
@@ -592,17 +592,23 @@ def _calc_chip_bounding_polygon(self, stepsize=None):
...
+            ((lx, hx), (ly, hy)) = self.wcs.bounding_box
 
         if stepsize is None:
             nintx = 2
             ninty = 2
         else:
-            nintx = max(2, int(np.ceil((nx + 1.0) / stepsize)))
-            ninty = max(2, int(np.ceil((ny + 1.0) / stepsize)))
+            nintx = max(2, int(np.ceil((hx - lx) / stepsize)))
+            ninty = max(2, int(np.ceil((ny - ly) / stepsize)))
```
jdavies-st: This should be `hy - ly`.

mcara:
@sosey It looks like you are a reviewer for this PR. Do you approve it?

sosey:
how did that happen? uh, lemme look?

jwst/tweakreg/wcsimage.py
sosey: so, as James mentions above, are you going to change this section to use the bounding_box_xxx function or do you want to merge this and then change it?

mcara: @sosey I am not familiar with `assign_wcs` but based on https://github.com/spacetelescope/jwst/pull/3014/files#diff-0173fbd4fd4061722ffb9856bf96c4e9R194 I have trouble seeing how is anything affecting `bounding_box` used here. It looks like it is still going to be in place... right?

PR #3016, opened 2019-01-04
More robust handling of ramps flagged as DO_NOT_USE

dmggh:
For ramps having pixel data quality flagged as DO_NOT_USE, the step now outputs the variances in the output as 0. The associated slopes and other quantities had been (and still are) calculated correctly in all 3 output products - this update is essentially cosmetic as it is only the variances that appear in the output arrays that have changed.


dmggh:
With update of the change log soon.

PR #3021, opened 2019-01-08
WIP Multi processing jump step

mwregan2:
This is a version of the jump step that uses Python Multiprocessing to spin up process that work on the jump detection in parallel. I've also added a couple of new test routines and some improvements in the comments and variable names.
  I measure a factor of 2.5 improvement in the clock time on my Mac Pro workstation and a factor of 4.5 running on a 24 core Linux server. 
  I have a multi-processing version of the ramp_fitting step that shows even larger improvements but I want to get people's opinions on this before I proceed with that.


sosey:
I have a peanut gallery question ... has SDP given us direction on how they'd like us to deal with multiprocessing as they run the jwst pipeline? For HST, I had to explicitly code a switch that allowed them to run on a single core so that they could have OWL drive the workload. Since the multiprocessing package is doing process splitting not shared memory threading this could impact their load balancing. I do see the option to set it to 1 core here, just wondering if that's SDPs intended usecase still?

mwregan2:
Megan,
  That’s one of my questions. I added the single core/thread option with the idea that this would be the default when running the production pipeline. I kind of see the ability to use multiple cores as something users would do when they want to rerun the pipeline and have full control over the computing resources. I guess the means that I should have set the default to be ‘one’ instead of ‘half’.
Mike

sosey:
Hey Mike, yup, if that is what SDP wants, then setting the default to 1 core would probably be a good idea.

mwregan2:
I'm confused why the Travis run of the unit tests for the  jump step are still failing. I've updated the code and they all run clean for me. Is it possible to see the output of the CI tests?

jdavies-st:
Yeah, we definitely don't want multiprocessing in SDP operations.  So multiprocessing turned off is what we want for normal pipeline operations here at ST.  @hbushouse can you confirm?

The other issue is that `numpy` already makes extensive use of multiple cores via the MKL library (if you are using the version availabe via `conda`.  This is probably why you are not seeing linear scaling of speed with number of cores, as there's already parallel processing going on in the array linear algebra even when not run in multi-processing mode.  Of course this depends on how `numpy` is built.

Finally, TravisCI only has 2 cores available if I recall correctly.  This should not be a problem if the default is to have MP turned off.  Perhaps testing the case with 2 cores would be possible in the unit tests.

hbushouse:
> Yeah, agree with Meg. We don't want multiprocessing in SDP operations. So multiprocessing turned off is what we want for normal pipeline operations here at ST. @hbushouse can you confirm?
> 

Can't confirm or deny at this time, although I suspect that DMS does NOT want multiprocessing in Ops. I'll raise it at the next DMS Design meeting.

jdavies-st:
I've restarted the TravisCI build to see if it stalls at the same point.

mwregan2:
It really looks like TravisCI is running with the initial version of this pull request rather than the one that includes commit bbffefe. The unit tests that are failing are the ones that are not directly related to multiprocessing but failed because I changed the calling parameters to twopoint_difference. I fixed this in the updated commit.

mwregan2:
I agree that the lack of fully linear scaling is expected because of the built in multi-threading in Numpy. When running the routine asks for the number of available cores. If TravisCI only has two available, it should adjust to that. 
  The latest commit changes the default number of requested cores to be 'one'.

jdavies-st:
I believe the Travis CI tests are failing here because they are running out of memory.  In the new step unit tests, a single science image takes up multiple GB because it is full frame and has 100 groups.  And there are at least 2 copies open at one time, plus the reffile arrays.

These tests will run much faster (and will be better for CI systems like Travis) if the test datasets are smaller, i.e. not full-frame.

mwregan2:
James,
  A 100 frame exposure for one of the NIR instruments uses 1.7 GB for MIRI it would only be 400 MB. It doesn’t seem to me that either of these should be a problem given that we have to process many of these exposures in parallel in production.
Mike

jdavies-st:
Don't forget the `err`, `pixeldq` and `groupdq` arrays.  A `RampModel` for NIR data with 100 groups takes up ~3.7GB when instantiated, as these arrays get instantiated as well.  If one has two open at once, as in this code, that pretty much maxes out the memory on a service like Travis CI where the workers generally have 8GB of memory.  Leaves no room for any other arrays you may need open in a temporary block.

You can see the effect here:

https://travis-ci.org/jdavies-st/jwst/jobs/529478961#L828

mwregan2:
Steve,

 If SDP is running in single thread does that mean the NumPy is single threaded too? That would really slow things down.

Mike

PR #3023, opened 2019-01-09
Add dark current tests

mfixstsci:


cracraft:
There are a set of tests written for MIRI that check that the correct subarray reference file is called. This is similar to the problem we encountered with the linearity tests in that those tests try to access files in CRDS, which cannot be done here. Until we come up with a solution for this type of test, we should remove the tests that check for the correct subarray reference file.

mfixstsci:
@cracraft should we remove them temporarily until we find a solution?

cracraft:
@mfixstsci I think that's our only option right now. We already know they will fail if they can't access the reference files. This is part of the issue we need to solve.

jaytmiller:
Depending on the properties of your tests,  there may be an existing CRDS API or one which can be easily added which would dynamically fetch and store references you need on the local test file system.    Would the time for downloading the files you need to run your tests be acceptable for your use case?   Is there enough local storage to hold the files you need?

cracraft:
I don't think the download time would be that long, but I haven't tested that. The purpose of the tests are simply to make sure that the proper subarray reference file was called, so it reads in a keyword from the header. That's the only thing the test needs the actual reference file for, to make sure the header subarray keyword matches the data subarray keyword.

jaytmiller:
If a Step or Pipeline actually ran as part of the test the references may have been cached locally automatically someplace other than /grp/crds/jwst.

If that's true,   I think e.g. this code:
```
    darkfile = outfile.meta.ref_file.dark.name
    dfile = darkfile.split("/")[2]        # parse darkfile name to discard crds://
    filepath = '/grp/crds/jwst/references/jwst/' + dfile
```
just changes to this:
```
   darkfile = outfile.meta.ref_file.dark.name
   from jwst.stpipe import crds_client
   filepath = crds_client.reference_uri_to_cache_path(darkfile)
```
and the refererence_uri_to_cache_path() function figures out what crds:// means in the context of the current environment.    Step class/subclass instances also have a similar method.

I'm not advocating for or against this check.     But that's how this might possibly be solved.

jwst/dark_current/tests/test_dark_sub.py
```diff
@@ -0,0 +1,517 @@
...
+
+def test_frame_averaging(setup_nrc_cube):
+    '''Check that if nframes>1 or groupgap>0, then the pipeline reconstructs
+       the dark reference file to match the frame averaging and groupgap
+       settings of the exposure.'''
+
+    # Values to build the fake data arrays
+    ngroups = 5
+    nrows = 2048
+    ncols = 2048
```
jdavies-st: Currently this is the test that takes 70% of the time in running these 8 tests, so it would be good to get this to use a bit less memory and go faster.

I think these sizes can be reduced dramatically so as to use less memory and speed the tests up.  Really just a 200x200 array with a few groups should be ok, right?  And instead of accessing pixel location `500,500` below, access `100,100`.

jwst/dark_current/tests/test_dark_sub.py
```diff
@@ -0,0 +1,517 @@
...
+    assert outfile.data[0, 1, 500, 500] == pytest.approx(1.45)
+    assert outfile.data[0, 2, 500, 500] == pytest.approx(2.05)
+    assert outfile.data[0, 3, 500, 500] == pytest.approx(2.65)
+
+    # check that the error array is not modified.
+    np.testing.assert_array_equal(outfile.err[:, :], 0,
+                                  err_msg='error array should remain 0 ')
+
+
+def make_rampmodel(nints, ngroups, ysize, xsize):
```
jdavies-st: It would be nice if these `make_rampmodel` and `make_darkmodel` functions used a consistent fixture format like that of `setup_nrc_cube` below.

jwst/dark_current/tests/test_dark_sub.py
```diff
@@ -0,0 +1,517 @@
...
+    # create raw input data for step
+    dm_ramp = make_rampmodel(nints, ngroups, ysize, xsize)
+    dm_ramp.meta.exposure.nframes = 1
+    dm_ramp.meta.exposure.groupgap = 0
+
+    # populate data array of science cube
+    for i in range(0, ngroups-1):
+        dm_ramp.data[0, i, :, :] = i
+
+    refgroups = 50
```
hbushouse: As above, for this type of test it's not necessary to have realistic data array dimensions, and hence xsize, ysize, ngroups, and refgroups could be reduced, if these tests end up being a burden on the overall CI unit test runs. (Remember, every time anyone does a commit, all of these tests are going to run).

jwst/dark_current/tests/test_dark_sub.py
```diff
@@ -0,0 +1,517 @@
...
+
+    # test that the output data file is equal to the difference found when subtracting ref file from sci file
+
+    np.testing.assert_array_equal(outfile.data[0, :, :, :], diff,
+                                  err_msg='dark file should be subtracted from sci file ')
+    np.testing.assert_array_equal(outfile.data[1, :, :, :], diff_int2,
+                                  err_msg='dark file should be subtracted from sci file ')
+
+
+def test_dark_skipped():
```
hbushouse: Isn't this effectively the same as `test_more_sci_frames` above? In which case they could be merged into a single test (just add the test of `darkstatus` to the other test).

jwst/dark_current/tests/test_dark_sub.py
```diff
@@ -0,0 +1,517 @@
...
+    outfile = darkcorr(dm_ramp, dark)
+
+    # get dark correction status from header
+    darkstatus = outfile.meta.cal_step.dark_sub
+    print('Dark status', darkstatus)
+
+    assert darkstatus == 'SKIPPED'
+
+
+def test_frame_avg():
```
hbushouse: Isn't this testing the same things as `test_frame_averaging` above?

jwst/dark_current/tests/test_dark_sub.py
```diff
@@ -0,0 +1,517 @@
...
+
+    # populate data array of reference file
+    for i in range(0, refgroups - 1):
+        dark.data[0, i, :, :] = i * 0.1
+
+    # apply correction
+    outfile = darkcorr(dm_ramp, dark)
+
+    # dark frames should be averaged in groups of 4 frames
+    # this will result in average values of 0.15, 0.55, .095. 1.35
```
hbushouse: The values in the comment here don't seem to match the values in lines 410-413. Unless `pytest.approx` provides a **very** rough approximation ;-).

jwst/dark_current/tests/test_dark_sub.py
```diff
@@ -0,0 +1,517 @@
...
+
+
+def test_more_sci_frames():
+    '''Check that data is unchanged if there are more frames in the science data is than in the reference file'''
+
+    # size of integration
+    nints = 1
+    ngroups = 30
+    xsize = 1032
+    ysize = 1024
```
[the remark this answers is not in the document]
cracraft: I have a question about this. The dark reference files are subarray specific. If we start just putting in random array sizes, won't this cause a problem? Are you saying we should be doing all these tests on subarrays instead?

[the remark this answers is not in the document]
hbushouse: It's true that if you were using actual dark reference files in these tests you'd need to worry about whether they're subarray or full-frame. But since all of the tests create both the science data and reference data arrays on the fly, all that matters is that the dimensions match one another (at least the 2 spatial dimensions). The low-level routines that do the actual data manipulation have no knowledge of (and don't care about) whether the data came from full-frame or subarray.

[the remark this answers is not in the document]
cracraft: Ok, you're right. I forgot that we set these up to create the dark data models as well.

jwst/dark_current/tests/test_dark_sub.py
cracraft: I'd like more flexibility in how I set up the sizes of the science and dark data arrays, so want to keep them as separate functions.

PR #3032, opened 2019-01-14
Check for constant wavelengths

### LINKED ISSUE #3031: In find_dispaxis in extract_1d, check for constant wavelengths
In the function for determining which is the dispersion direction, if there is no wavelength attribute, the wavelengths will be computed using the `WCS`.  To avoid evaluating the function more often than is necessary, the wavelengths are initially computed at only three pixels near the middle of the bounding box.  If any of the wavelengths is zero or NaN, the wavelengths will be computed over the entire bounding box in order to find all the valid values.  That test (zero or NaN) needs to be expanded to include the possibility that the differences between wavelengths in the horizontal and vertical directions are the same (e.g. the wavelengths could be constant).

philhodge:
Function find_dispaxis in extract.py was modified to also test whether the difference in wavelengths in the horizontal direction was exactly equal to the difference in wavelengths in the vertical direction.  Closes #3031.

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -458,10 +458,16 @@ def find_dispaxis(input_model, slit, spectral_order, extract_params):
             wl_10 = stuff[2]
         dwlx = wl_10 - wl_00
         dwly = wl_01 - wl_00
-        if (np.isnan(dwlx) or np.isnan(dwly) or
+        if (np.isnan(dwlx) or np.isnan(dwly) or dwlx == dwly or
             wl_00 == 0 or wl_01 == 0 or wl_10 == 0):
-                log.warning("wavelength from WCS is NaN or 0 "
-                            "within the bounding box")
+                if dwlx == dwly:
+                    log.warning("One-pixel offset gives dwlx = {}, dwly = {}"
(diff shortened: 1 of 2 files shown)


jwst/extract_1d/extract.py
```diff
@@ -458,10 +458,16 @@ def find_dispaxis(input_model, slit, spectral_order, extract_params):
             wl_10 = stuff[2]
         dwlx = wl_10 - wl_00
         dwly = wl_01 - wl_00
-        if (np.isnan(dwlx) or np.isnan(dwly) or
+        if (np.isnan(dwlx) or np.isnan(dwly) or dwlx == dwly or
             wl_00 == 0 or wl_01 == 0 or wl_10 == 0):
-                log.warning("wavelength from WCS is NaN or 0 "
-                            "within the bounding box")
+                if dwlx == dwly:
```
sosey: Does this also trigger a full bounding box computation like the nan/zero case?

philhodge: Yes.  This is already included above, in line 461.  I added this separate test just so I could log a different message from the nan/zero case.

sosey: got it.

PR #3034, opened 2019-01-15
Remove commented code block related to pandokia

jhunkeler:
:godmode:

PR #3035, opened 2019-01-15
Remove subpackage version information

jhunkeler:
Ref: https://github.com/spacetelescope/jwst/pull/3034#pullrequestreview-192754357


jwst/cal_ver_steps.py
```diff
@@ -43,7 +42,7 @@ def __init__(self, author, **pars):
         """
         # Initialize output reference file information based on user input
         descrip = pars.get('pars', "JWST calibration processing step version reference file")
-        history = pars.get('history', "Created by cal_ver_steps version {}".format(__version__))
+        history = pars.get('history', "Created by jwst version {}".format(__version__))
```
jhunkeler: How much does this change matter? Will this break something elsewhere?

hbushouse: No we aren't using it anymore, at least to my knowledge. It was used in the past to create one giant list of the versions of all the sub-packages, which was then stored in a file. I don't know if @jaytmiller is still using that file anywhere in CRDS, but we don't use it for the CAL pipeline anymore.

jaytmiller:
Yeah,  actually,  we were not able to ditch the CALVER related requirements this summer although we tried because it seems silly.   

This script is used by CAL to generate a reference file defining CAL versions,  however CAL wants to.  

The reference file is then managed by CRDS and used to support pass-through web interface from CRDS to the AUI...  to meet requirements of other systems.

Fishing around,  I found these requirements in a 2017 version of the JWST requirements:

8.1.7.2	Header Keywords for Software Versions
Header keywords shall reflect the software version of each calibration step used to process the data file. [DMS-594]

8.1.3.2	Monitor Software Versions
The Reprocessing System shall monitor changes in data processing software. [DMS-582]

I don't immediately see an AUI requirement but I'm sure something exists or they would never have bothered with this service.

@stsci-hack was leading this effort (among many things) at the time and may recall more.  

Since we couldn't ditch it this summer,  I think our intent was to automate something which meets requirements and then wait it out.

jhunkeler:
Hmm...
The only module reporting a version from/to anywhere was `cal_ver_steps`. The rest of the modules defined it and never used it. If there's some need for backwards compatibility then each `step/__init__.py` could execute `from ..version import __version__`, so `cal_ver_steps` would continue recording values like nothing ever happened.

Everybody wins.

hbushouse:
That line in `cal_ver_steps` only populates a single history record at the top of the reference file that it's creating, in order to track the origin of the reference file. The real work of compiling the list of sub-package version numbers in the body of the reference file occurs later in the script. So the modification to line 45 just changes the formatting and content of that history record, that's all. So it's fine.

jdavies-st:
Cool.  Yeah, I would say let's merge this as-is and deal with step versions separately.

PR #3036, opened 2019-01-15
Fix our RTD documentation builds

jdavies-st:
So this fixes the build.  Basically, peg `python` and `numpy` in the conda env, and then do all other dependencies via `pip`.

The build results of this PR are here

https://readthedocs.org/projects/jwsttest/builds/8396900/

### DIFF
--- a/.rtd-environment.yml
+++ b/.rtd-environment.yml
@@ -2,8 +2,8 @@ name: jwst
 dependencies:
   - "python>=3.5"
   - setuptools
-  - numpy
   - pip:
+    - numpy
     - sphinx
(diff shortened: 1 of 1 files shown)


jhunkeler:
I can't remember if these docs require `graphviz`, but if you're having trouble getting module dependency graphs to render, you'll want to add it as requirement under `conda:`. Looks good to me otherwise.

jdavies-st:
@hbushouse, do the built docs here

https://jwsttest.readthedocs.io/en/rtd/

look as you expect them to?  I haven't looked at these for a while, so my eyes are probably not the best checks.

PR #3038, opened 2019-01-16
Update README for B7.2.1 patch release

hbushouse:
Updated the jwst pkg vs. DMS build table to include info for the B7.2.1 (0.12.3) release.


stscicrawford:
Just want to confirm -- in the end, context 500 went into the pipeline?

hbushouse:
As far as I know, yes. Mike briefly tried 0495 when we were having problems, but that didn't make any difference.

PR #3040, opened 2019-01-17
Simplify transform_bbox_from_datamodel

mcara:
This PR is a follow-up to https://github.com/spacetelescope/jwst/pull/3014#issuecomment-454909182 and simplifies the code for `transform_bbox_from_datamodel`.

However, I would suggest a further simplification: combine three functions into a single one:
```python
def to_bbox(a, order='yx'):
    if order not in ('xy', 'yx'):
        raise ValueError("Unsupported 'order' value")
    try:
        a = a.data.shape
    except AttributeError:
        pass

    bbox = ((-0.5, a[-2] - 0.5), (-0.5, a[-1] - 0.5))

    return bbox if order == 'yx' else bbox[::-1]
```
The `try-except` block above can be replaced with an `if isinstance()`.

### DIFF
--- a/jwst/assign_wcs/niriss.py
+++ b/jwst/assign_wcs/niriss.py
@@ -249,7 +249,7 @@ def imaging_distortion(input_model, reference_files):
         # Check if the model has a bounding box.
         distortion.bounding_box
     except NotImplementedError:
-        distortion.bounding_box = transform_bbox_from_shape(input_model)
+        distortion.bounding_box = transform_bbox_from_shape(input_model.data.shape)
 
     dist.close()
     return distortion
--- a/jwst/assign_wcs/niriss.py
+++ b/jwst/assign_wcs/niriss.py
@@ -8,7 +8,7 @@ import gwcs.coordinate_frames as cf
 from gwcs import wcs
 
 from .util import (not_implemented_mode, subarray_transform,
-                   velocity_correction, bounding_box_from_subarray, transform_bbox_from_datamodel)
+                   velocity_correction, bounding_box_from_subarray, transform_bbox_from_shape)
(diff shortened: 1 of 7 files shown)


nden:
This looks fine.

:-1: on merging the three functions in one. As I said in a previous comment we started with one function and it was confusing. It's easy to forget that the order is different.

:+1: on merging `transform_bbox_from_shape` and `transform_bbox_from_datamodel` in one function.

mcara:
Do you want a deprecation warning for `transform_bbox_from_shape` and  `transform_bbox_from_datamodel`? How to call the new function? ``transform_bbox()`?

nden:
I don't think we need a deprecation warning. I would call it `transform_bbox_from_shape`. The only difference with the previous function is this now can be applied to any model. However, AFAIK the data in all models hsa a shape consistent with the new function. If anyone can think of a case when this isn't true, speak up.

jwst/assign_wcs/niriss.py
```diff
@@ -249,7 +249,7 @@ def imaging_distortion(input_model, reference_files):
         # Check if the model has a bounding box.
         distortion.bounding_box
     except NotImplementedError:
-        distortion.bounding_box = transform_bbox_from_datamodel(input_model)
+        distortion.bounding_box = transform_bbox_from_shape(input_model)
```
nden: This now needs the `shape`, right?

mcara: I must have missed it.

mcara: @nden Check out latest change.

nden:
Thanks @mcara

PR #3057, opened 2019-01-24
Temporarily suppress benign arithmetic warnings in ramp fitting

dmggh:
Temporarily suppress benign arithmetic warnings (divide by zero or operating on a NaN) in statements where they are expected to occur, then re-enable as soon afterwards as appropriate.   Calculations from these statements that result in these runtime warnings are already correctly handled in the code, as the results are later reset to the appropriate values. These updates are to reduce the amount of clutter in the log file.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -16,6 +16,7 @@
 import time
 import logging
 import numpy as np
+import warnings
 
 from .. import datamodels
 from ..datamodels import dqflags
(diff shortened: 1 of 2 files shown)


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -443,10 +444,19 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
                                        group_time, max_seg )
 
             segs_4[num_int, :, rlo:rhi, :] = segs_beg_3
+
+            # Suppress harmless arithmetic warnings for now
+            warnings.filterwarnings("ignore", ".*invalid value.*",
+                RuntimeWarning)
+            warnings.filterwarnings("ignore", ".*divide by zero.*",
+                RuntimeWarning)
```
jdavies-st: Another way to do this would be to use `numpy.errstate` context manager.  For example:
```
with np.errstate(invalid='ignore', divide='ignore'):
    np.sqrt(-1)
    np.array([1, 1]) / np.array([1, 0])
```
Once you exit the block, the error state returns to the default.

jdavies-st: I.e. doing it this way will ask `numpy` to not even emit the warnings.

dmggh: I had considered that but decided I preferred to avoid the extra indentation.

PR #3059, opened 2019-01-24
Run doctestsplus items in a tmpdir

### LINKED ISSUE #3053: doctests are leaving behind output files in the jwst repo
Since the merging of #3003, when one runs the unit tests locally, the following files get left behind:
```
	jwst/associations/tests/new_distortion.asdf
	jwst/associations/tests/niriss_photom_0001.fits
	jwst/associations/tests/reffile.asdf
```
These are produced by doctests.  Doctests should not be writing out files to the local repo.  Two solutions:

1) Use a _tmpdir_ fixture that [autodetects for doctests](https://stackoverflow.com/questions/46962007/how-to-automatically-change-to-pytest-temporary-directory-for-all-doctests).

2) Comment out any `write()` or `save()` methods in the documentation.

Preferences?

The offending docs:

https://github.com/spacetelescope/jwst/blame/master/docs/jwst/assign_wcs/asdf-howto.rst#L120

https://github.com/spacetelescope/jwst/blame/master/docs/jwst/assign_wcs/asdf-howto.rst#L133

https://github.com/spacetelescope/jwst/blame/master/docs/jwst/references_general/photom_reffile.rst#L254

jdavies-st:
Resolves #3053 by running each `doctestplus` item in its own _tmpdir_.

Perhaps this is overkill?  Maybe we should just comment out those offending lines in the documentation?  Or maybe we shouldn't be running the documentation of some of these items as unit tests?  Opinions?  @nden

### DIFF
--- a/docs/conftest.py
+++ b/docs/conftest.py
@@ -1,0 +1,13 @@
+import pytest
+
+@pytest.fixture(autouse=True)
+def _docdir(request):
+
+    # Trigger ONLY for doctestplus.
+    doctest_plugin = request.config.pluginmanager.getplugin("doctestplus")
+    if isinstance(request.node, doctest_plugin._doctest_textfile_item_cls):
+        tmpdir = request.getfixturevalue('tmpdir')
+        with tmpdir.as_cwd():
+            yield
+    else:
+        yield


nden:
I wonder what the options are out there of cleaning after running a sphinx build.

jdavies-st:
I will scope this to `'session'`.  Thanks for the suggestion @stscieisenhamer.

As for cleanup after a sphinx build, I've often wondered that myself.  Currently we stuff the build into a directory that's in `.gitignore`, which is probably the easiest way to do it.  The problem with the doctest outputs littering the repo, is that they end up where ever you happen to be running the tests from within the repo (or in our case, `associations` because of a "bad" test in there).

I think either removing the lines that write out files (or commenting them out) or this solution is probably our 2 choices right now.

jdavies-st:
So because this fixture is `autouse` and needs to be used each time on each test to see if it uses the `doctestplus` plugin, it cannot be scoped to `session`.  If I scope it to `session`, it only runs on the first doctest and then packs its bags for Aruba.  Bummer.  That said, it does leave behind a bunch of temporary directories that are mostly empty.  Not a big overhead.  I say we merge.

PR #3062, opened 2019-01-25
Remove pytest as an install dependency

jdavies-st:
To remove `pytest` as an install dependency, we
 - move `full_pool_rules` pytest fixture from `helpers` to a new `conftest.py` in `associations`
 - move unit tests out of `associations` modules and into `test_registry.py`
 - update some minimum versions of dependencies


jdavies-st:
I will wait to merge this until it is rebased to the changes from #2999.

PR #3063, opened 2019-01-25
correct bounding box for subarray in MIRI LRS observations

nden:
The bounding box of MIRI LRS observations is determined by the wavelength solution in the `specwcs` reference file, in full frame coordinates. This adjust the bounding box when the observation is a subarray.

There's also a minor correction for the NIRISS SOSS mode from the center to the edge of the pixel.

### DIFF
--- a/jwst/assign_wcs/miri.py
+++ b/jwst/assign_wcs/miri.py
@@ -184,8 +184,13 @@ def lrs(input_model, reference_files):
     y0 = lrsdata[:, 4]
     x1 = lrsdata[:, 5]
 
-    bb = ((x0.min() - 0.5 + zero_point[0], x1.max() + 0.5 + zero_point[0]),
-          (y0.min() - 0.5 + zero_point[1], y0.max() + 0.5 + zero_point[1]))
+    subarray_xstart = input_model.meta.subarray.xstart - 1
+    subarray_ystart = input_model.meta.subarray.ystart - 1
+
+    # The bounding box is computed from the data in the wavelength solution
+    # and corrected for subarray.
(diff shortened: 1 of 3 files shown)


nden:
Just to clarify: The initial problem was that the bounding box was in full frame coordinates while it's expected to be in "image" coordinates. Fixing it revealed that the input `x` and `y` coordinates were corrected for the subarray origin only when passed to the spatial transform. The last commit ensures that the subarray corrected inputs are passed  to the spatial and spectral transform. The output wavelengths for slitless modes now are in the range 3.5 - 14.7 microns.

jwst/assign_wcs/niriss.py
```diff
@@ -167,8 +167,8 @@ def niriss_soss(input_model, reference_files):
         cm_order2 = subarray2full | cm_order2
         cm_order3 = subarray2full | cm_order3
 
-        bbox = ((0, input_model.meta.subarray.ysize),
-                (0, input_model.meta.subarray.xsize))
+        bbox = ((-0.5, input_model.meta.subarray.ysize - 0.5),
+                (-0.5, input_model.meta.subarray.xsize - 0.5))
```
jdavies-st: Nice catch.

jwst/assign_wcs/miri.py
```diff
@@ -157,8 +157,8 @@ def lrs(input_model, reference_files):
     with DistortionModel(reference_files['distortion']) as dist:
         distortion = dist.model
 
-    if subarray2full is not None:
-        distortion = subarray2full | distortion
+    #if subarray2full is not None:
+    #    distortion = subarray2full | distortion
```
jdavies-st: These commented-out lines can be removed.

PR #3067, opened 2019-01-29
Add dark current unit tests

jdavies-st:
This is rebased version of #3023 with 2 extra commits implementing some memory reduction to see if the TravisCI builds can pass.

Closes #3023.


jdavies-st:
Same problem in TravisCI as the other PR.  Hmm.

cracraft:
I'm making edits to this code based on earlier comments. I've noticed that the version I have (grabbed right before lunch), doesn't have James' most recent changes. Will that be a problem?

jdavies-st:
No problem Misty.  I'm just trying to debug the issues this PR is having on Travis.  It passes when we run it locally on our machine, but I think it these tests are running out of memory on the Travis builds.  I've noticed a few places we can save on memory here and there, and have made some progress.

Feel free to hold off on your own edits if you want until we can see how mine look once we can get the Travis builds to pass.  If that's OK for you.

jdavies-st:
Oh, and to answer your comment, the reason my edits are not showing up on your branch is that I'm working on my own branch right now (and this PR is tied to my branch).  I just wanted to be able to make lots of edits and iterate with Travis without messing up your branch.

cracraft:
Well, I had already edited the code and was ready to commit the changes, but I see that a lot of the changes you're making are to the same things I changed. Do you want me to wait and see what you do, then compare my new version to yours? If so, you'll have to talk me through pulling the correct version of the code to update and putting it back so you can get them.

jdavies-st:
You're welcome to push to your branch, but I think the Travis build will still fail and we won't know why.  In this branch, I've changed the Travis configuration so I can see a more verbose output.  Hopefully this will help us debug the memory issues.

The test `test_frame_averaging` is the one that is still failing.  I think it is because the fixture objects never go out of scope, and it iterates over all readpatts in the test, instead of parametrizing the test with the readpatts, so you get a generated test for each readpatt.  But I'm not positive.

cracraft:
For whatever reason, test_frame_averaging is also failing on my system. It's Alicia's test, so I wasn't making any edits to that one. (Just to be clear, the MIRI version of the test, test_frame_avg is the one I didn't want the values edited on.)

jdavies-st:
Got it.  Thanks @cracraft.

I suspect it is one of these irritating `conda` problems where they recompile `numpy` to use multiple cores for array operations, but of course if you do this, machine precision can mean that results are not reproducible.  Addition and multiplication is no longer associative.  😱 

So perhaps setting a lower relative tolerance on that test might get it to pass for you, using `np.testing.assert_allclose()`.

jdavies-st:
Ah ha!  Reducing the number of groups in the NRC dark model from 100 down to 10 was the key.  Since the dark model is always 2048x2048, this is a huge savings in compute time and memory.

Now that test runs 10 times faster.  🎉

jdavies-st:
Looks like the tests have finally passed.  :tada:

There are lots of commits here, but the main issue with these nice, new unit tests was that they were using too many resources.  They ran fine on our personal machines, but the free tier of TravisCI has limited resources in the build containers, so they were failing with `OutOfMemory` errors.  The takeaways:

 - The 4D multi-read, multi-group cubes for JWST are big.  2Kx2K imagers combined with other dimensions of 50 or more are probably too big four unit tests running on a CI system.  Added that the `dq` and `err` arrays, it can make for very large datasets.
 - Given the size of the data, dq and error arrays for these, we need to be careful in the tests not to make unnecessary copies of them if we can avoid it.

The solution was to drastically cut the number of groups and the overall size of the arrays.  A cube with 5 groups test the code the same as a cube with 50 groups.  Since we are not testing the top-level `Step`, but testing the lower-level functionality, it is fine to use smaller arrays, as long as they sufficiently test the algorithm.

So feel free to write the tests with the standard NIRCam and MIRI detector sizes, make sure they pass locally.  Then look to see how you can optimize them for CI by reducing to a minium the integrations, groups or even detector size, if allowable.

The other major change I made was to clean up the configuration of the NRC `test_frame_averaging`.  Some of its config was coming from the read pattern global variables, some internally, and the fixture it was calling was also getting configured in this hybrid way.  Made it difficult to debug.  And the test was actually running 9 tests, one for each readpatt, in a loop.  Because the fixture was being called in a loop, I think we had 9 copies of the fixture sitting around in memory, as the fixture is scoped to the function.  The solution was to parametrize the test into 9 tests, 1 for each readpatt, and the fixture is created and evaporates with each test.  And the configuration for the fixture is now funneled from the parameters through the test.  No more need for the global variables, and just a simple function that generates the parameters, one for each readpatt.  Running the tests now gives:
```
$pytest -v jwst/dark_current/tests/test_dark_sub.py
======================================== test session starts ========================================
platform darwin -- Python 3.7.0, pytest-4.1.1, py-1.7.0, pluggy-0.8.1 -- /Users/jdavies/miniconda3/envs/jwst_dev/bin/python
cachedir: .pytest_cache
plugins: requests-mock-1.5.2, xdist-1.26.0, remotedata-0.3.1, openfiles-0.3.2, forked-0.2, doctestplus-0.2.0, cov-2.6.1, arraydiff-0.3, ci-watson-0.2.dev13+g93c80610
collected 16 items                                                                                  

jwst/dark_current/tests/test_dark_sub.py::test_frame_averaging[DEEP8-3-8-12-20-20] PASSED     [  6%]
jwst/dark_current/tests/test_dark_sub.py::test_frame_averaging[DEEP2-3-2-18-20-20] PASSED     [ 12%]
jwst/dark_current/tests/test_dark_sub.py::test_frame_averaging[MEDIUM8-3-8-2-20-20] PASSED    [ 18%]
jwst/dark_current/tests/test_dark_sub.py::test_frame_averaging[MEDIUM2-3-2-8-20-20] PASSED    [ 25%]
jwst/dark_current/tests/test_dark_sub.py::test_frame_averaging[SHALLOW4-3-4-1-20-20] PASSED   [ 31%]
jwst/dark_current/tests/test_dark_sub.py::test_frame_averaging[SHALLOW2-3-2-3-20-20] PASSED   [ 37%]
jwst/dark_current/tests/test_dark_sub.py::test_frame_averaging[BRIGHT2-3-2-0-20-20] PASSED    [ 43%]
jwst/dark_current/tests/test_dark_sub.py::test_frame_averaging[BRIGHT1-3-1-1-20-20] PASSED    [ 50%]
jwst/dark_current/tests/test_dark_sub.py::test_frame_averaging[RAPID-3-1-0-20-20] PASSED      [ 56%]
jwst/dark_current/tests/test_dark_sub.py::test_more_sci_frames PASSED                         [ 62%]
jwst/dark_current/tests/test_dark_sub.py::test_sub_by_frame PASSED                            [ 68%]
jwst/dark_current/tests/test_dark_sub.py::test_nan PASSED                                     [ 75%]
jwst/dark_current/tests/test_dark_sub.py::test_dq_combine PASSED                              [ 81%]
jwst/dark_current/tests/test_dark_sub.py::test_2_int PASSED                                   [ 87%]
jwst/dark_current/tests/test_dark_sub.py::test_dark_skipped PASSED                            [ 93%]
jwst/dark_current/tests/test_dark_sub.py::test_frame_avg PASSED                               [100%]

===================================== 16 passed in 7.55 seconds =====================================

```
Before, the were taking ~31 seconds to pass on my personal machine.  This new optimized version with smaller array sizes takes ~7 seconds.  Same coverage and functionality.

Anyway, I would be happy to get reviews from @cracraft @mfixstsci and @aliciacanipe to make sure I did not mess up anything important.

A note: while debugging these `dark_current` tests, one of the `linearity` tests was also failing on TravisCI due to memory issues.  I did some changes to that set of tests to reduce the number of redundant 4-D arrays held in memory at any point.  When a datamodel is instantiated from a shape, it automatically creates `data`, `err` and `dq` arrays of the correct size filled with zeros, so this saves on having to create those numpy arrays first and then pass them to the datamodel to get copied into the datamodel.  Probably reduces memory usage by a factor of 3 or so.

jdavies-st:
@aliciacanipe, no worries.  We have a lot of code like this in the actual `jwst` pipeline itself that could easily be optimized.  First crack is to always get it working.  One is usually forced to do optimization when one runs into problems.  =)

@cracraft In the case of converting your functions to fixtures, it doesn't do much for us except make the fixture dependency explicit in each test that uses them, i.e. 
```
def test_more_sci_frames(make_rampmodel, make_darkmodel):
    ...
```
and makes it consistent with NRC fixture.

Generally though, fixtures are more flexible than functions for `pytest`.  You can scope a fixture to be run once per function, class, module or session.  Which means instead of running for every function that calls them, it yields something for the first, then the same thing (in memory) for every subsequent function, class or module that uses them.  Like test functions with `pytest`, you can also parametrize a fixture.  So they're really useful, as they can break up complex tests into simple tests and often speed things up dramatically and reduce the size of the codebase.

You can image a test that ran a level 2 or level 3 pipeline and then did some asserts with some of the pipeline outputs.  It would not be very scalable, and if one assert failed for one of the outputs (and the others didn't), the whole test would fail.  Meanwhile, if you wrote a fixture that ran the pipeline and provide the outputs to a few tests, each test could test something different in the outputs, and if one failed it would be very easy to see which failed.  The tagline for `pytest` fixtures is "explicit, modular, scalable", and I think that's pretty accurate.

I do like that a test's dependency on a fixture is explicit in the test definition.  This makes it easy to see which tests depend on which fixtures without code diving.  For a module with 40 tests, that's really useful.

Looks like these changes are OK.  I will merge, which will close the sister PR.

PR #3068, opened 2019-01-30
make y-slit range an input parameter

### LINKED ISSUE #2819: NIRSpec MOS default slit_y limits need to be extended
Issue [JP-448](https://jira.stsci.edu/browse/JP-448) was created by James Muzerolle:

Currently, the assign_wcs step calculates the WCS parameters for each spectrum within slit_y limits of -0.55 to 0.55.  It turns out this results in many MOS spectra not having real WCS info (e.g., wavelength) over the full area subtended by a typical source spectrum.  The problem is most acute for observations of a point source that is placed close to the edges of a shutter/slitlet in the cross-dispersion direction.  Among other things, this causes significant jumps in the flat field-corrected data because that correction defaults to 1 (no correction) where there is no wavelength information.  For now, I would recommend a generous range, say -1 to 1.

A related question - is there a way to manually specify the slit_y limits?  That would be useful from a testing perspective to assess what the optimal range should be.

nden:
Resolves #2819

### DIFF
--- a/jwst/assign_wcs/assign_wcs_step.py
+++ b/jwst/assign_wcs/assign_wcs_step.py
@@ -41,9 +41,9 @@ class AssignWcsStep(Step):
     """
 
     spec = """
-        slit_y_low = float(default=-.55)
-        slit_y_high = float(default=.55)
-        # The y-range of a slit, (0, 0) is the center.
+        slit_y_low = float(default=-.55)  # The lower edge of a slit, (0, 0) is the center.
+        slit_y_high = float(default=.55)  # The upper edge of a slit, (0, 0) is the center.
+
     """
 
     reference_file_types = ['distortion', 'filteroffset', 'specwcs', 'regions',
@@ -79,7 +79,7 @@ class AssignWcsStep(Step):
                     message = "MSA metadata file (MSAMETFL) is required for NRS_MSASPEC exposures."
                     log.error(message)
                     raise MissingMSAFileError(message)
-            slit_y_range = self.slit_y_low, self.slit_y_high
+            slit_y_range = [self.slit_y_low, self.slit_y_high]
             result = load_wcs(input_model, reference_file_names, slit_y_range)
 
         return result
--- a/jwst/assign_wcs/assign_wcs_step.py
+++ b/jwst/assign_wcs/assign_wcs_step.py
@@ -41,6 +41,9 @@ class AssignWcsStep(Step):
     """
 
     spec = """
+        slit_y_low = float(default=-.55)
+        slit_y_high = float(default=.55)
+        # The y-range of a slit, (0, 0) is the center.
     """
 
     reference_file_types = ['distortion', 'filteroffset', 'specwcs', 'regions',
@@ -48,7 +51,7 @@ class AssignWcsStep(Step):
                             'fore', 'fpa', 'msa', 'ote', 'ifupost',
                             'ifufore', 'ifuslicer']
 
-    def process(self, input):
+    def process(self, input, *args, **kwargs):
         reference_file_names = {}
         with datamodels.open(input) as input_model:
             # If input type is not supported, log warning, set to 'skipped', exit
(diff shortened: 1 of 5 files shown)


jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -41,14 +41,17 @@ class AssignWcsStep(Step):
     """
 
     spec = """
+        slit_y_low = float(default=-.55)
+        slit_y_high = float(default=.55)
+        # The y-range of a slit, (0, 0) is the center.
```
jdavies-st: Would the spec be clearer or more concise if
```
slit_y_range = float_list(default=[-.55, .55]) # The y-range of a slit, (0, 0) is the center
```
Does the `float_list` ConfigObj type work with our command line usage?  Not sure.

Regardless, the `# comment` should got on the same line as the each spec entry so they get picked up in the help.

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -76,7 +79,7 @@ def process(self, input):
                     message = "MSA metadata file (MSAMETFL) is required for NRS_MSASPEC exposures."
                     log.error(message)
                     raise MissingMSAFileError(message)
-
-            result = load_wcs(input_model, reference_file_names)
+            slit_y_range = self.slit_y_low, self.slit_y_high
```
philhodge: You refer to this later as a list.  If it matters that it should be a list rather than a tuple, you should enclose the terms on the right hand side in brackets.

jwst/assign_wcs/nirspec.py
```diff
@@ -805,7 +824,8 @@ def gwa_to_ifuslit(slits, input_model, disperser, reference_files):
     return Gwa2Slit(slits, slit_models)
 
 
-def gwa_to_slit(open_slits, input_model, disperser, reference_files):
+def gwa_to_slit(open_slits, input_model, disperser,
+                reference_files, slit_y_range=[-.55, .55]):
```
philhodge: It doesn't look as if you passed `slit_y_range` to this function where it is called.

jwst/assign_wcs/nirspec.py
```diff
@@ -1348,7 +1368,7 @@ def get_slit_location_model(slitdata):
     return model
 
 
-def gwa_to_ymsa(msa2gwa_model, lam_cen=None, slit=None):
+def gwa_to_ymsa(msa2gwa_model, lam_cen=None, slit=None, slit_y_range=None):
```
philhodge: Is this a safe default value for `slit_y_range`?

jwst/assign_wcs/tests/test_nirspec.py
```diff
@@ -274,7 +274,7 @@ def test_msa_configuration_normal():
     # Test 1: Reasonably normal as well
     msa_meta_id = 12
     msaconfl = get_file_path('msa_configuration.fits')
-    slitlet_info = nirspec.get_open_msa_slits(msaconfl, msa_meta_id)
+    slitlet_info = nirspec.get_open_msa_slits(msaconfl, msa_meta_id, slit_y_range=[-.5, .5])
```
philhodge: Do you want to include `slit_y_range` in the call to `get_open_msa_slits` at line 291 as well?

ddavis-stsci:
Looks good to me.

jwst/assign_wcs/assign_wcs_step.py
nden: `float_list` does not work. It should probably be fixed in a separate PR. Comment moved.

jwst/assign_wcs/assign_wcs_step.py
nden: Any iterable is fine at this point. I enclosed it in brackets for clarity.

jwst/assign_wcs/nirspec.py
nden: `slit_y_range` is used only in IFU mode and it should always be populated. I can't make it a position argument because it's only for IFU (as `slit` is for MOS and FS only). But I want it to fail if the value was not passed.

nden:
Wow, parentheses in an inline comment  to `step.spec` broke configobj validation badly.

sosey:
ooo...more consideration of seeing if we can replace configobj with asdf?

PR #3077, opened 2019-02-01
Ramp fitting unit tests for ramp cases.

dmggh:


jwst/ramp_fitting/tests/test_ramp_fit_cases.py
```diff
@@ -0,0 +1,541 @@
...
+#  results for each of the individual segments in a ramp.  Within each test is a
+#  description of classification ('CASE') within the code of all of the segments
+#  for the pixel based on the ramp's GROUPDQ, and the resulting segments and
+#  their SCI values (these are mostly for my reference).
+#
+
+def test_pix_0():
+    """
+    CASE A: segment has >2 groups, at end of ramp.
+    SCI seg is [15., 25., 35., 54., 55., 65., 75., 94., 95., 105.](A)
```
hbushouse: It may not matter to the actual test, but it appears that you're intending to simulate a ramp that has accumulating signal of 10 DN per group, with groups 4 and 8 having anomalous values relative to that steady increase of 10 DN per group. If groups 4 and 8 are supposed to represent something like CR hits (they have 19 DN of signal, instead of the normal 10, for an excess of 9), all of the groups following those simulated hits should continue to carry along the extra 9 DN of signal from the hits, i.e. the sequence would be [15, 25, 35, 54, 64, 74, 84, 103, 113, 123], so that the last 2 groups have a total excess signal of 18 DN relative to normal. Or are CR hits in groups 4 and 8 not what you're intending to simulate here?

jwst/ramp_fitting/tests/test_ramp_fit_cases.py
```diff
@@ -0,0 +1,541 @@
...
+    """
+
+    ngroups, nints, nrows, ncols, deltatime, gain, readnoise = set_scalars()
+    RampMod, RnMod, GMod, pixdq, groupdq, err = create_mod_arrays( ngroups,
+                           nints, nrows, ncols, deltatime, gain, readnoise )
+
+    # Populate pixel-specific SCI and GROUPDQ arrays
+    RampMod.data[0,:,0,0] = np.array([ 15., 25., 35., 54., 55.,
+                                    65., 75., 94., 95., 105.], dtype=np.float32)
+    RampMod.groupdq[0,:,0,0] = np.array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
```
hbushouse: I see that there aren't any non-zero DQ flags set for the ramp, so I guess you're NOT simulating CR hits, which means you can ignore my comment above.

jwst/ramp_fitting/tests/test_ramp_fit_cases.py
```diff
@@ -0,0 +1,541 @@
...
+def create_mod_arrays(ngroups, nints, nrows, ncols, deltatime, gain, readnoise):
+    """
+    For an input datacube (arbitrarily chosen to be MIRI), create arrays having
+    the specified dimensions for the pixel DQ, the group DQ, and the
+    ERR extensions, and create datamodels for the ramp, readnoise, and gain.
+    """ 
+
+    gain = np.ones(shape=(nrows, ncols), dtype=np.float64) * gain
+    err = np.zeros(shape=(nints, ngroups, nrows, ncols), dtype=np.float64)
+    data = np.zeros(shape=(nints, ngroups, nrows, ncols), dtype=np.float64)
```
hbushouse: When we use the ramp_fit routines in the pipeline step we aren't actually passing it float64 data, are we? I thought the data model .data and .err arrays were only float32. Will that bias the test results at all, by not testing in exactly the same way things are used in the actual step?

jwst/ramp_fitting/tests/test_ramp_fit_cases.py
dmggh: Ok. This is just a test for the DQ values, so the actual dummy SCI values up the ramp don't really matter.

Issue #3082, opened 2019-02-05
create the master background step as a NOOP


nden:
Initially identified tasks - 3 sp

- [x] Write the step module and the cfg file
- [x] specify a user parameter for the user supplied 1d file
- [x] input is a ModelContainer, (or AT, single file)  
- [x] the step status should be updated - FITS kw S_MSBSUB
- [x] the step should save the output
- [x] ~~Add the step result suffix to the standard locations; set suffix in step~~ Move this into the future
- [x] Record the file name of the 1D bckg file used in the FITS header  
- [x] File keyword dict tickets
    - [x] keyword `S_MSBSUB` is step completion status of Master background step
    - [x] keyword `MSTRBKGD` is User-supplied master background file

jdavies-st:
A clarification on recording in the SCIENCE FITS header the 1D background file name used - this can only be done if
```
MasterBackgroundStep.call(input, user_background="my_amazing_background.fits")
```
or if the computed one from the association is written out to file
```
MasterBackgroundStep.call(input, save_background=True)
```
Neither of these will be the default for this step when running in `Spec3Pipeline`.  So do we still want to record this?  And just not populate it in the default mode?

hbushouse:
Yes, I would populate the keyword whenever `user_background` is not None.

jdavies-st:
JWST Keyword Dictionary requests filed:

https://jira.stsci.edu/projects/JWSTKD/issues/JWSTKD-300
https://jira.stsci.edu/projects/JWSTKD/issues/JWSTKD-301

PR #3090, opened 2019-02-05
Master background step

jdavies-st:
Implements the basic step scaffolding for `MasterBackgroundStep`.

Left undone:
 - pick a default suffix for this step
 - how to handle the various inputs and feed them to an underlying looping function
 - how to handle input background spectrum vs computing one from the inputs
 - flesh out the unit test to do something useful

Addresses some of #3082.


jwst/master_background/master_background_step.py
```diff
@@ -0,0 +1,86 @@
...
+    spec = """
+        user_background = str(default=None) # Path to user-supplied master background
+        save_background = boolean(default=False) # Save computed master background
+    """
+
+
+    def process(self, input):
+
+        """
+        Compute a master background from N > 1 dedicated background exposures
```
philhodge: Why must there be more than one background exposure?

jwst/master_background/master_background_step.py
```diff
@@ -0,0 +1,86 @@
...
+        and subtract it from the target science exposure(s)
+
+        Parameters
+        ----------
+        input : `~jwst.datamodels.ImageModel`, `~jwst.datamodels.IFUImageModel`, `~jwst.datamodels.ModelContainer`, association
+            Input target data model(s) to which master background subtraction is
+            to be applied
+
+        user_background: string path or `~jwst.datamodels.MultiSpecModel`
```
philhodge: Or None?

jwst/master_background/master_background_step.py
```diff
@@ -0,0 +1,86 @@
...
+            # Handle associations, or input ModelContainers
+            elif isinstance(input_data, datamodels.ModelContainer):
+                pass
+
+            # Handle MIRI LRS
+            elif isinstance(input_data, datamodels.ImageModel):
+                pass
+
+            # Handle MIRI MRS and NIRSpec IFU
+            elif isinstance(input_data, datamodels.IFUImageModel):
```
philhodge: Could IFUImageModel be handled the same as ImageModel?

jwst/master_background/master_background_step.py
hbushouse: I would leave this a bit more generic, such as "Compute a master background spectrum", because it's not always the case that the MSB will come from dedicated background exposures. For NIRSpec MSA, for example, it'll come from background slitlets in the same exposure.

jwst/master_background/master_background_step.py
```diff
@@ -0,0 +1,86 @@
...
+                pass
+
+            else:
+                result = input_data
+                self.log.warning(
+                    "Input %s cannot be handled.  Doing nothing.", input
```
philhodge: Maybe "type(input)" instead of just input?

jwst/pipeline/master_background.cfg
```diff
@@ -0,0 +1,5 @@
+name = "master_background" 
+class = "jwst.background.MasterBackgroundStep"
```
philhodge: Should the directory be master_background rather than just background?  (There already is a background subdirectory of jwst.)

jwst/master_background/master_background_step.py
hbushouse: Should change the message to saying that the step is being skipped, and then set `result.meta.cal_step.master_back_sub = 'SKIPPED'`.

jwst/master_background/master_background_step.py
```diff
@@ -0,0 +1,86 @@
...
+                    )
+
+            # Check if user has supplied a master background spectrum.
+            if self.user_background is None:
+                pass
+            else:
+                pass
+
+            # Save the computed background if requested by user
+            if self.save_background:
```
hbushouse: Should we make this and the if-block starting in line 66 mutually exclusive? i.e. if the user has supplied the master background spectrum, what's the point in saving it?

jwst/master_background/master_background_step.py
jdavies-st: Ah, so there could be just one dedicated background exposure in an association?  Hardly a "master" background then, but I suppose we should handle it, right?

jwst/master_background/master_background_step.py
jdavies-st: I think an `IFUImageModel` will have to have each slice iterated over?  But my understanding is that the slices are still embedded in the 2D `ImageModel`, right?  Maybe @jemorrison can clarify?

jwst/pipeline/master_background.cfg
hbushouse: The code files in the PR are in jwst.master_background, so yes this should be updated.

jdavies-st: Yes! That is a bug.  I will write a unit test for this, as my current tests did not catch this bug.

jwst/master_background/master_background_step.py
jdavies-st: True!  See my question in #3082 also.

jwst/master_background/master_background_step.py
hbushouse: Correct.

jwst/master_background/master_background_step.py
```diff
@@ -0,0 +1,86 @@
...
+    MasterBackgroundStep:  Compute and subtract master background from spectra
+    """
+
+    spec = """
+        user_background = str(default=None) # Path to user-supplied master background
+        save_background = boolean(default=False) # Save computed master background
+    """
+
+
+    def process(self, input):
```
nden: Should `user_background` and `save_background` be arguments to `process`?

jwst/master_background/master_background_step.py
nden: Yes, an `IFUImageModel` is a 2D IFU image.

nden:
What happened to `jwst/lib/suffix.py`? Is it just reordering?

jwst/master_background/master_background_step.py
hbushouse: There are a few differences in the content of `ImageModel` vs. `IFUImageModel`. For example, `ImageModel` includes things like the pathloss correction arrays, but `IFUImageModel` does not (because IFU data don't get that correction). So while the master background subtraction step may operate on both of them in the same way, it's still important to carry through the original input model type to the output that's created, because the output is intended to be an updated copy of the input (and hence we don't want to accidentally lose some extra data arrays along the way).

jdavies-st: The main difference is that the `IFUImageModel` has ~30 cutouts and WCS objects that need to be iterated over, whereas the MIRI LRS `ImageModel` has 1.  Is the way these are accessed different as well?  I.e. one has to tell the WCS object which slice to grab for the `IFUImageModel`?

nden: The IFU modes do not go through extract_2d, so no cutouts. Unless you meant something else.

philhodge: For NIRSpec there is a list of wcs objects, one for each IFU stripe.  Each stripe has a bounding_box that specifies where that stripe is on the full 2-D image.  I don't know how MIRI handles this.

hbushouse: There is code in the NIRSpec portion of the flat-field step, as well as the NIRSpec IFU and MIRI MRS portions of the photom step, that already do the operation of building a 2-D wavelength "image" for those types of exposures, because the corrections/calibrations done in those steps are wavelength-dependent. So those can be used as examples. And don't we also carry along a 2-D wavelength array in the data products for spectroscopic modes now, or is that only for slit-like modes and not IFU? If the 2-D wavelength array is already in the product/model, you're all set.

jdavies-st:
Yes, just reordering and adding the new suffix for this step.  There's a function that generates this list.  When it was originally generated, perhaps it was generated with a different version of `python`?  I know `dict` are ordered differently in 3.5 vs 3.6 and later.

jwst/master_background/master_background_step.py
jdavies-st: They are not for any of our other steps.  I was also looking at this.  The arguments get passed via `call` or `run` whose args are generated on-the-fly from the `configobj` parsing of the config file or `spec`.  I hate this, as it is very non-standard python, but it is currently what we do.

There's probably nothing wrong with allowing calls to process, but all the args to `call` are class attributes, so they are accessible within `process` via `self.arg`.

jdavies-st: Btw, I also dislike that `process` is a public method.  Since it doesn't take args, and we don't generally want users to use it instead of `call` or `run`, it probably should not be public.  `_process` would be better.

It also annoys me that steps do not have an `__init__` method.  Many of our steps set up class attributes within `process` that are not setup via the `Step.__init__` superclass.  An `__init__` method would be natural location to put these, especially if there are other methods other than `process` within the step were these attributes are needed.

jwst/master_background/master_background_step.py
jdavies-st: We really need a class method for writing out all this `'SKIPPED'` business in our steps.  I'm filing an issue for that.

jdavies-st:
@hbushouse, I think I've addressed your and everyone else's comments with the latest commit.

Issue #3099, opened 2019-02-07
Issue with make_header output file


jotaylor:
While using make_header I found that something is wrong in the RELSENS2D (8th) extension. I created the file like so:
`make_header NIRISS NIS_WFSS 1 test_wfss1_new.fits`

I tried to see what the structure of the file was like, but encountered this error:
```
> fits.info("test_wfss1_new.fits")                                                                                                  
WARNING: The following header keyword is invalid or follows an unrecognized non-standard convention:
 [astropy.io.fits.card]
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-58-06a5b374c598> in <module>
----> 1 pf.info("test_wfss1_new.fits")

~/anaconda3/envs/jwst_dev/lib/python3.6/site-packages/astropy/io/fits/convenience.py in info(filename, output, **kwargs)
    730     f = fitsopen(filename, mode=mode, **kwargs)
    731     try:
--> 732         ret = f.info(output=output)
    733     finally:
    734         if closed:

~/anaconda3/envs/jwst_dev/lib/python3.6/site-packages/astropy/io/fits/hdu/hdulist.py in info(self, output)
    992         default = ('', '', '', 0, (), '', '')
    993         for idx, hdu in enumerate(self):
--> 994             summary = hdu._summary()
    995             if len(summary) < len(default):
    996                 summary += default[len(summary):]

AttributeError: '_ValidHDU' object has no attribute '_summary'

2019-02-07 15:44:51,126 - stpipe - WARNING - The following header keyword is invalid or follows an unrecognized non-standard convention:
```

Seems like the problem begins in the 8th extension. 

```
> fits.getheader("test_wfss1_new.fits", 8)                                                                                           
WARNING: The following header keyword is invalid or follows an unrecognized non-standard convention:
 [astropy.io.fits.card]
```
It then prints a LOT of blank lines. Like.... thousands, millions, billions- I don't know how many because I didn't have enough time to scroll through the terminal. After the 10^? blank lines it finally prints this:
```
XTENSION= 'IMAGE   '           / Image extension                                
BITPIX  =                  -32 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                 2048                                                  
NAXIS2  =                 2048                                                  
PCOUNT  =                    0 / number of parameters                           
GCOUNT  =                    1 / number of groups                               
EXTNAME = 'RELSENS2D'          / extension name                                 
[...]

PR #3100, opened 2019-02-07
remove relsens_2d from image schema

nden:
`relsens_2d` is applied only to IFU data.


jdavies-st:
Might relate to #3099.

jdavies-st:
This PR might also break some regression tests, FYI.  But no worries.

PR #3101, opened 2019-02-07
Record user-supplied master background in MSTRBKGD keyword

jdavies-st:


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -2132,4 +2132,8 @@ properties:
                 channel:
                   title: Channel ID
                   type: string
+      master_background:
+        title: User-supplied master background file
```
nden: Should the user be able to pass alternatively a model, when called from a python session or a script?

hbushouse: That would be really convenient and has been requested in general for things like overriding reference files (right now override only takes a file name). But when that's done, you don't have a file name to save in this keyword. So then what? Don't populated the keyword?

nden: If the model was initialized with a file name, model.meta.filename will be populated. If it was initialized with an HDUList, model.meta.filename wwas populated by HDUList.fileinfo.
There's still some possibility that model.meta.filename will be None, in which case we can record the class name of the model?

philhodge: I would suggest populating the keyword with an empty string.  I think it's confusing to have a keyword disappear, and blank seems like an appropriate value to me when there is no file name.

jwst/master_background/master_background_step.py
```diff
@@ -73,19 +78,23 @@ def process(self, input):
...
+                # Return input as dummy result for now
+                result = input_data
             else:
-                pass
+                # Return input as dummy result for now
+                result = input_data
+
+                # Record name of user-supplied master background spectrum
+                try:
+                    result.meta.background.master_background = basename(self.user_background)
```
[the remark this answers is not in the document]
jdavies-st: Doh!  My test didn't catch this.  Weird.

jwst/datamodels/schemas/core.schema.yaml
jdavies-st: Agree @nden that passing a file path or an open datamodel would be ideal.  Currently the use of the standard types provided with `configobj` does not allow this (i.e. you could pass only a `string`).

I can create a custom type called `datamodel` which would be anything that `datamodels.open()` can handle.  This would also solve the issue for override reffiles that @hbushouse mentions above too.  I looked at this validation code in `stpipe` yesterday, and it is pretty simple.  I can make this change I think in a separate PR.

hbushouse: If necessary (i.e. time becomes critical), this kind of enhancement can be left to a later sprint.

jwst/master_background/master_background_step.py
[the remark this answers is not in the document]
jdavies-st: The test didn't catch this because `ModelContainer` contains `meta`, and will happily accept whatever you put into it, instead of rejecting it in the `try/except` block and putting the result in each model of the `ModelContainer`.

jwst/datamodels/schemas/core.schema.yaml
jdavies-st: How about we leave this as it is for now.  I'll work on `datamodel` configspec validator separately, and if successful, we can easily change this to use it.

Issue #3083, opened 2019-02-05
reproject a 1-D background spectrum to 2-D 


nden:
Write a (to be named) function which reprojects a 1-D background spectrum into the 2-D space of source data. - 13 sp

[ ] Use the wavelengths from the 2D cutout and the 1D flux(wav) relation to get interpolated background flux values.

jdavies-st:
Resolved by #3103.

Issue #2308, opened 2018-07-31
TASK: Implementation of Master background subtraction


stscicrawford:
This is to implement the level 3 required master background subtraction. 
The https://jira.stsci.edu/browse/JP-257

Issues should be opened for the planning, implementation, and testing specifications required for this step

PR #3103, opened 2019-02-08
Function to convert 1-D spectrum to 2-D

philhodge:
Function `expand_to_2d` was added, to expand a 1-D background file to 2-D.  A utility function `create_background` was also added; this takes 1-D arrays of wavelength and background flux, and copies those values to a table in the appropriate format for a 1-D master background spectrum.
See issue #2308.

### DIFF
--- a/jwst/master_background/create_master_bkg.py
+++ b/jwst/master_background/create_master_bkg.py
@@ -2,18 +2,15 @@ import numpy as np
 
 from .. import datamodels
 
-def create_background(m_bkg_spec, wavelength, flux):
+def create_background(wavelength, flux):
     """Create a 1-D spectrum table as a MultiSpecModel.
 
     This is the syntax for accessing the data in the columns:
-    wavelength = output_model.spec_table[0]['wavelength'])
-    background = output_model.spec_table[0]['flux'])
+    wavelength = output_model.spec[0].spec_table['wavelength']
+    background = output_model.spec[0].spec_table['flux']
 
     Parameters
     ----------
-    m_bkg_spec : string
-        The name of the file to create.
-
     wavelength : 1-D ndarray
         Array of wavelengths, in nanometers.
 
@@ -26,7 +23,7 @@ def create_background(m_bkg_spec, wavelength, flux):
         A data model containing the 1-D background spectrum.  This can be
         written to disk by calling:
 
-        output_model.save(m_bkg_spec)
+        output_model.save(<filename>)
     """
 
     wl_shape = wavelength.shape
--- a/jwst/master_background/create_master_bkg.py
+++ b/jwst/master_background/create_master_bkg.py
@@ -24,7 +24,7 @@ def create_background(wavelength, flux):
 
     Returns
     -------
-    output_model : datamodels.MultiSpecModel, or None
+    output_model : `~jwst.datamodels.MultiSlitModel`, or None
         A data model containing the 1-D background spectrum.  This can be
         written to disk by calling:
 
--- a/jwst/master_background/create_master_bkg.py
+++ b/jwst/master_background/create_master_bkg.py
@@ -1,6 +1,11 @@
+import logging
+
 import numpy as np
 
 from .. import datamodels
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
 
 def create_background(wavelength, flux):
     """Create a 1-D spectrum table as a MultiSpecModel.
@@ -12,7 +17,7 @@ def create_background(wavelength, flux):
     Parameters
     ----------
     wavelength : 1-D ndarray
-        Array of wavelengths, in nanometers.
+        Array of wavelengths, in micrometers.
 
     flux : 1-D ndarray
         Array of background fluxes.
@@ -31,19 +36,20 @@ def create_background(wavelength, flux):
     bad = False
     if len(wl_shape) > 1:
         bad = True
-        print("ERROR:  The wavelength array has shape {}; expected "
+        log.error("The wavelength array has shape {}; expected "
               "a 1-D array".format(wl_shape))
     if len(flux_shape) > 1:
         bad = True
-        print("ERROR:  The background flux array has shape {}; expected "
+        log.error("The background flux array has shape {}; expected "
               "a 1-D array".format(flux_shape))
     if bad:
         return None
 
     if wl_shape[0] != flux_shape[0]:
-        print("ERROR:  wavelength array has length {},".format(wl_shape[0]))
-        print("but background flux array has length {}.".format(flux_shape[0]))
-        print("The arrays must be the same size.")
+        log.error("wavelength array has length {}, "
+                  "but background flux array has length {}."
+                  .format(wl_shape[0], flux_shape[0]))
+        log.error("The arrays must be the same size.")
         return None
 
     # Create arrays for columns that we won't need.
@@ -52,18 +58,14 @@ def create_background(wavelength, flux):
 
     output_model = datamodels.MultiSpecModel()
 
-    # This is temporary, just to give us spec.spec_table.dtype
-    spec = datamodels.SpecModel()
+    spec_dtype = datamodels.SpecModel().spec_table.dtype
 
     # xxx Include one more argument at the end, after column NPIXELS has
     # xxx been added to the x1d table.  The new argument should be float64
     # xxx and with values of 1 (i.e. not dummy).
     otab = np.array(list(zip(wavelength, flux,
                              dummy, dq, dummy, dummy, dummy, dummy)),
-                    dtype=spec.spec_table.dtype)
-
-    # We're done with the temporary one.
-    spec.close()
+                    dtype=spec_dtype)
 
     spec = datamodels.SpecModel(spec_table=otab)
     output_model.spec.append(spec)
--- a/jwst/master_background/create_master_bkg.py
+++ b/jwst/master_background/create_master_bkg.py
@@ -1,0 +1,74 @@
+import numpy as np
+
+from .. import datamodels
+
+def create_background(m_bkg_spec, wavelength, flux):
+    """Create a 1-D spectrum table as a MultiSpecModel.
+
+    This is the syntax for accessing the data in the columns:
+    wavelength = output_model.spec_table[0]['wavelength'])
+    background = output_model.spec_table[0]['flux'])
+
+    Parameters
+    ----------
+    m_bkg_spec : string
+        The name of the file to create.
+
+    wavelength : 1-D ndarray
+        Array of wavelengths, in nanometers.
+
+    flux : 1-D ndarray
+        Array of background fluxes.
+
+    Returns
+    -------
+    output_model : datamodels.MultiSpecModel, or None
+        A data model containing the 1-D background spectrum.  This can be
+        written to disk by calling:
+
+        output_model.save(m_bkg_spec)
+    """
+
+    wl_shape = wavelength.shape
+    flux_shape = flux.shape
+    bad = False
+    if len(wl_shape) > 1:
+        bad = True
+        print("ERROR:  The wavelength array has shape {}; expected "
+              "a 1-D array".format(wl_shape))
+    if len(flux_shape) > 1:
+        bad = True
+        print("ERROR:  The background flux array has shape {}; expected "
+              "a 1-D array".format(flux_shape))
+    if bad:
+        return None
+
+    if wl_shape[0] != flux_shape[0]:
+        print("ERROR:  wavelength array has length {},".format(wl_shape[0]))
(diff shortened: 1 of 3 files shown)


hbushouse:
Fixes #3083

jwst/master_background/expand_to_2d.py
```diff
@@ -0,0 +1,314 @@
...
+    # bkg_flux will be a 2-D array, because wl_array is 2-D.
+    bkg_flux = np.interp(wl_array, tab_wavelength, tab_background,
+                         left=0., right=0.)
+
+    background.data[:] = bkg_flux.copy()
+
+    return background
+
+
+def bkg_for_IFUimage(input, tab_wavelength, tab_background):
```
jdavies-st: To appease the PEP8 gods, should be `bkg_for_ifuimage` or similar, since this is a function.

jwst/master_background/create_master_bkg.py
```diff
@@ -0,0 +1,74 @@
...
+
+    if wl_shape[0] != flux_shape[0]:
+        print("ERROR:  wavelength array has length {},".format(wl_shape[0]))
+        print("but background flux array has length {}.".format(flux_shape[0]))
+        print("The arrays must be the same size.")
+        return None
+
+    # Create arrays for columns that we won't need.
+    dummy = np.zeros(wl_shape[0], dtype=np.float64)
+    dq = np.zeros(wl_shape[0], dtype=np.int32)
```
hbushouse: I think we normally use uint32 for DQ values, since we don't need negatives.

jwst/master_background/expand_to_2d.py
```diff
@@ -0,0 +1,314 @@
...
+    Returns
+    -------
+    background : JWST data model
+        A copy of `input` but with the data replaced by the background,
+        "expanded" from 1-D to 2-D.
+    """
+
+    with datamodels.open(m_bkg_spec) as bkg:
+        if len(bkg.spec) > 1:
+            print("Warning:  The input 1-D spectrum contains multiple spectra")
```
hbushouse: Should use log.warning() instead of print, at least once this becomes part of the "official" step package.

jwst/master_background/expand_to_2d.py
```diff
@@ -0,0 +1,314 @@
...
+            xstart = max(xstart, 0)
+            ystart = max(ystart, 0)
+            xstop = min(xstop, 2048)
+            ystop = min(ystop, 2048)
+
+            shape = (ystop - ystart, xstop - xstart)
+            grid = np.indices(shape, dtype=np.float64)
+            grid[0] += ystart
+            grid[1] += xstart
+            wl_array = ifu_wcs(grid[1], grid[0])[2]
```
jdavies-st: The block from lines 240-260 can probably be replaced by
```

        wl_array = grid_from_bounding_box(ifu_wcs.bounding_box)
```
though you should verify it is giving the same result as your code.  This is the standard library function for generating a grid of points from a WCS with a `bounding_box` defined.

jwst/master_background/expand_to_2d.py
```diff
@@ -0,0 +1,314 @@
...
+    if (wl_array is None or len(wl_array) == 0 or
+        np.nanmin(wl_array) == 0. and np.nanmax(wl_array) == 0.):
+            got_wavelength = False
+            wl_array = None
+
+    if hasattr(model, "wcs") and not got_wavelength:
+        wcs = model.wcs.forward_transform
+        shape = model.data.shape
+        grid = np.indices(shape[-2:], dtype=np.float64)
+        wl_array = wcs(grid[1], grid[0])[2]
```
jdavies-st: again, using `grid_from_bounding_box` might be the better thing to use here.

jwst/master_background/expand_to_2d.py
```diff
@@ -0,0 +1,314 @@
...
+
+    # We're going to use np.interp, so tab_wavelength must be strictly
+    # increasing.
+    if not np.all(np.diff(tab_wavelength) > 0):
+        index = np.argsort(tab_wavelength)
+        tab_wavelength = tab_wavelength[index].copy()
+        tab_background = tab_background[index].copy()
+
+    # Handle associations, or input ModelContainers
+    if isinstance(input, datamodels.ModelContainer):
```
hbushouse: I wonder if we should let some higher-level routine, i.e. between the step class and this routine, take care of the handling of different forms of input science/source data, in which case this routine could assume it's always given a single data model instance and even a single source instance (e.g. if the input to the step is a MultiSlitModel, only pass one Slit instance at a time to this function).

jwst/master_background/expand_to_2d.py
```diff
@@ -0,0 +1,314 @@
...
+    -------
+    background : JWST IFUImageModel
+        A copy of `input` but with the data replaced by the background,
+        "expanded" from 1-D to 2-D.
+    """
+
+    background = input.copy()
+    background.data[:, :] = 0.
+
+    if input.meta.instrument.name == "NIRSPEC":
```
hbushouse: Might want to add .upper() to the end of instrument.name, just to make sure the value is upper-case.

jwst/master_background/expand_to_2d.py
jdavies-st: Just checked it.  Your code and `grid_from_bounding_box` produce identical results for the IFUImage data.

jwst/master_background/expand_to_2d.py
philhodge: It does pass flake8; however, using lower case does seem reasonable.  I'll change it to bkg_for_ifu_image.

jwst/master_background/expand_to_2d.py
jdavies-st: I think this *is* the intermediate function we were thinking of, as the functions this calls (i.e. `bkg_for_multislit`) actually do the work of making the background.

jwst/master_background/expand_to_2d.py
jdavies-st: Cool.  Yeah, `flake8` doesn't do anything about function/class naming conventions in PEP8.  It ignores it, probably because there's legacy code in `python` itself that is not PEP8 compliant.  If you want that functionality in `flake8`, you can install `pep8-naming`, which will extend `flake8` to check for that. It's essentially a plugin.

jwst/master_background/expand_to_2d.py
philhodge: I want the wavelengths over the full size of the data array, regardless of `bounding_box`.  Wavelengths outside the bounding box should be NaN, and those values will be set to -1 via the check for `np.nan`, and the flux will therefore be set to 0 by the interpolation routine.  I have comments to that effect, but they're only in one function (`bkg_for_multislit`).

jwst/master_background/create_master_bkg.py
```diff
@@ -0,0 +1,74 @@
...
+    wavelength = output_model.spec_table[0]['wavelength'])
+    background = output_model.spec_table[0]['flux'])
+
+    Parameters
+    ----------
+    m_bkg_spec : string
+        The name of the file to create.
+
+    wavelength : 1-D ndarray
+        Array of wavelengths, in nanometers.
```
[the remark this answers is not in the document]
philhodge: That's a typo; I meant micrometers.  Thanks for catching that.

jwst/master_background/create_master_bkg.py
philhodge: It's currently int32 in spec.schema.yaml, but I can change that.  I'm going to change that file anyway to add a new column.

jwst/master_background/expand_to_2d.py
jdavies-st: Ah, right. My bad.

jwst/master_background/expand_to_2d.py
philhodge: Yes, I forgot to change these.

jwst/master_background/expand_to_2d.py
jdavies-st: This means that some of the code in the top-level `MasterBackgroundStep` class can be torn out, and we can just do the subtraction of the background in the `process()` method.

philhodge: Some of them separate cleanly, but not all.

jdavies-st: I can do that in a separate PR that pulls this functionality into the step.

jwst/master_background/create_master_bkg.py
```diff
@@ -0,0 +1,74 @@
...
+        return None
+
+    # Create arrays for columns that we won't need.
+    dummy = np.zeros(wl_shape[0], dtype=np.float64)
+    dq = np.zeros(wl_shape[0], dtype=np.int32)
+
+    output_model = datamodels.MultiSpecModel()
+
+    # This is temporary, just to give us spec.spec_table.dtype
+    spec = datamodels.SpecModel()
```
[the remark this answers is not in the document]
philhodge: Good suggestion; thanks.

jwst/master_background/expand_to_2d.py
```diff
@@ -0,0 +1,314 @@
...
+            # Convert to integers, and add one to the upper limits,
+            # because we want to use these as slice limits.
+            xstart = math.floor(xstart)
+            xstop = math.ceil(xstop) + 1
+            ystart = math.floor(ystart)
+            ystop = math.ceil(ystop) + 1
+            xstart = max(xstart, 0)
+            ystart = max(ystart, 0)
+            xstop = min(xstop, 2048)
+            ystop = min(ystop, 2048)
```
nden: Lines 240 - 254 can be substituted with 
```
from gwcs.utils import _toindex

xstart, xstop, ystart, ystop = _toindex(ifu_wcs.bounding_box)
```
This ensures that xstart/ystart are >-0 and xend/yend <=2048

jwst/master_background/expand_to_2d.py
```diff
@@ -0,0 +1,314 @@
...
+
+            shape = (ystop - ystart, xstop - xstart)
+            grid = np.indices(shape, dtype=np.float64)
+            grid[0] += ystart
+            grid[1] += xstart
```
nden: lines 257-259 are equivalent to

```

x, y = grid_from_bounding_box(ifu_wcs.bounding_box)
```
The return is already in x, y (Cartisian) order.

jwst/master_background/expand_to_2d.py
```diff
@@ -0,0 +1,314 @@
...
+        wl_array = model.wavelength.copy()
+        got_wavelength = True                   # may be reset below
+    else:
+        wl_array = None
+    if (wl_array is None or len(wl_array) == 0 or
+        np.nanmin(wl_array) == 0. and np.nanmax(wl_array) == 0.):
+            got_wavelength = False
+            wl_array = None
+
+    if hasattr(model, "wcs") and not got_wavelength:
```
nden: "wcs" is an attribute of model.meta. `model.wcs` points to the FITS WCS which we don't want to use.
  This needs to be fixed but is a separate issue.

jwst/master_background/expand_to_2d.py
nden: Is there a reason not to cal lthe `wcs` object directly, like model.meta.wcs(x, y)? Just wondering why you are calling the forward_transform?

jwst/master_background/expand_to_2d.py
```diff
@@ -0,0 +1,314 @@
...
+        The input science data.
+
+    m_bkg_spec : string or JWST data model
+        Either the name of a file containing a 1-D background spectrum,
+        or a data model containing such a spectrum.
+
+    Returns
+    -------
+    background : JWST data model
+        A copy of `input` but with the data replaced by the background,
```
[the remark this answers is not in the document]
philhodge: The numpy documentation on writing docstrings says to use single backticks for variables.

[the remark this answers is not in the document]
nden: OK, let's see how it renders online.

jwst/master_background/expand_to_2d.py
philhodge: Thanks!  I wonder whether I made this mistake in other steps, too.

jwst/master_background/expand_to_2d.py
philhodge: I don't know why I thought that was preferable.  I'll delete that.

jwst/master_background/create_master_bkg.py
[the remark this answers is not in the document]
philhodge: The extract1d table has wavelengths in microns, and I actually intended to write "micrometers" in the docstring.  Should I change the units to meters?  Both for the extract1d table and the input to this function?  I thought JWST users would expect microns.

PR #3106, opened 2019-02-08
Update MIRI LRS distortion model for CDP-7

### LINKED ISSUE #2951: Update MIRI LRS WCS transforms to include inverse functions

drlaw1558:
This PR contains modifications to the MIRI LRS distortion model to enable forward/backwards transforms and incorporates subarray bounding boxes in the transform.  Additional changes have been made to the MIRI imaging comments to indicate that the CDP-7 delivery no longer uses the XAN/YAN coordinate frame.

Updated MIRI LRS and imager distortion reference files are being delivered to CRDS for CDP-7 as well, but the new files should not be required for the code updated here to work.

Resolves #2951, #2825

### DIFF
--- a/jwst/assign_wcs/miri.py
+++ b/jwst/assign_wcs/miri.py
@@ -1,13 +1,17 @@
+# Nadia
 import os.path
 import logging
 import numpy as np
+from numpy import matlib as mb
 from astropy.modeling import models
 from astropy import coordinates as coord
 from astropy import units as u
 from astropy.io import fits
 
+from scipy.interpolate import UnivariateSpline
 import gwcs.coordinate_frames as cf
 from gwcs import selector
+from gwcs.utils import _toindex
 from . import pointing
 from ..transforms import models as jwmodels
 from .util import (not_implemented_mode, subarray_transform,
@@ -100,9 +104,8 @@ def imaging_distortion(input_model, reference_files):
        sign to that delivered by the IT) (uses the "filteroffset" ref file)
     2. Apply MI (uses "distortion" ref file)
     3. Apply Ai and BI matrices (uses "distortion" ref file)
-    4. Apply the TI matrix (this gives Xan/Yan coordinates) (uses "distortion" ref file)
-    5. Aply the XanYan --> V2V3 transform (uses "distortion" ref file)
-    6. Apply V2V3 --> sky transform
+    4. Apply the TI matrix (this gives V2/V3 coordinates) (uses "distortion" ref file)
+    5. Apply V2V3 --> sky transform
 
     """
     # Read in the distortion.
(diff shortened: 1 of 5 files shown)


jwst/assign_wcs/miri.py
```diff
@@ -134,102 +136,212 @@ def imaging_distortion(input_model, reference_files):
...
-    distortion = distortion | rotation
-
-    # Load and process the reference data.
+    # Combine models to create subarray to v2v3 distortion
+    if subarray2full is not None:
+        subarray_dist = subarray2full | distortion
+        subarray_dist.inverse = distortion.inverse | subarray2full.inverse
+    else:
+        subarray_dist = distortion
+        subarray_dist.inverst = distortion.inverse
```
nden:
```suggestion
        subarray_dist.inverse = distortion.inverse
```

jwst/assign_wcs/miri.py
```diff
@@ -134,102 +136,212 @@ def imaging_distortion(input_model, reference_files):
...
+                  (np.floor(y2.min()+ zero_point[1]) - 0.5, np.ceil(y0.max()+ zero_point[1]) + 0.5))
+        bb = ((bb_sub[0][0]+subarray_xstart,bb_sub[0][1]+subarray_xstart),
+              (bb_sub[1][0]+subarray_ystart,bb_sub[1][1]+subarray_ystart))
+
+    # Find the ROW of the zero point
+    row_zero_point = zero_point[1]
+    # Make a vector of x,y locations for every pixel in the reference row
+    yrow, xrow = np.mgrid[row_zero_point:row_zero_point + 1, 0:input_model.data.shape[1]]
+    # And compute the v2,v3 coordinates of pixels in this reference row
+    v2v3refrow = np.array(subarray_dist(xrow, yrow))[:, 0, :]
```
nden: I am assuming the zero point for slitless mode is in subarray coordinates (which it seems to be) or this won't be correct.

jwst/assign_wcs/miri.py
```diff
@@ -134,102 +136,212 @@ def imaging_distortion(input_model, reference_files):
...
+    # Therefore fit with a spline for the ycenrot->wavelength transform
+    # Reverse vectors so that yinv is increasing (needed for spline fitting function)
+    yrev=ycenrot[::-1]
+    wrev=wavetab[::-1]
+    # Spline fit with enforced smoothness
+    spl=UnivariateSpline(yrev,wrev,s=0.002)
+    # Evaluate the fit at the rotated-y reference points
+    wavereference=spl(yrev)
+    # wavereference now contains the wavelengths corresponding to regularly-sampled ycenrot, create the model
+    wavemodel = models.Tabular1D(lookup_table=wavereference, points=yrev,name='waveref',bounds_error=False, fill_value=np.nan)
```
nden: At some point would it be desirable to drop the `Tabular1D` model and use the spline directly? It will require writing the serialization code for the spline.

nden: Although I ran a quick test and I got slightly different results when using the spline instead of the Tabular model (I am guessing because of the interpolation) but that may not be significant.

jwst/assign_wcs/miri.py
```diff
@@ -134,102 +136,212 @@ def imaging_distortion(input_model, reference_files):
...
+    dettotel.inverse = xform1 | xysubtoxyrot.inverse
 
-    # Now the actual pipeline.
-    pipeline = [(detector, det_to_v2v3),
-                (v2v3, v23_to_world),
-                (world, None)
-                ]
-    return pipeline
+    # Bounding box is the subarray bounding box, because we're assuming subarray coordinates passed in
+    dettotel.bounding_box = bb_sub[::-1]
```
nden: :+1:

jwst/assign_wcs/miri.py
```diff
@@ -134,102 +136,212 @@ def imaging_distortion(input_model, reference_files):
...
+        
+    # If in slitless mode, define the bounding box X locations using the subarray x boundaries
+    # and the y locations using the corner locations in the CDP reference file.  Make sure to
+    # omit the 4 reference pixels on the left edge of slitless subarray.
+    if input_model.meta.exposure.type.lower() == 'mir_lrs-slitless':
+        bb_sub = ((input_model.meta.subarray.xstart-1+4-0.5,input_model.meta.subarray.xsize-1+0.5),
+                  (np.floor(y2.min()+ zero_point[1]) - 0.5, np.ceil(y0.max()+ zero_point[1]) + 0.5))
+        bb = ((bb_sub[0][0]+subarray_xstart,bb_sub[0][1]+subarray_xstart),
+              (bb_sub[1][0]+subarray_ystart,bb_sub[1][1]+subarray_ystart))
+
```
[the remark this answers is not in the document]
drlaw1558: - The zero point specified in the FITS headers is in full-frame coordinates for both slit and slitless.  However, after reading values from the reference file line 213 immediately converts it to subarray coordinates for slitless observations.  Thus, yes, as used in this code the zero point is in 'data' coordinates.
- Yes, fixed slit is always full frame
- Yes, slitless is always subarray

jwst/assign_wcs/miri.py
drlaw1558: Agreed this is a typo- please go ahead and fix in whatever manner is easiest.

jwst/assign_wcs/miri.py
drlaw1558: Right; as above the zeropoint is converted to subarray coordinates when first read from the reference file.

jwst/assign_wcs/miri.py
drlaw1558: Sure, this could be a long-term thing to do.  I think I used Tabular1D since I wasn't sure how to do the interpolation to arbitrary locations directly from the spline model, but doing it directly would probably be preferable.  I imagine that any small differences probably result from the Tabular1D linear interpolation between nearest points just approximating to the actual spline fit.

Issue #3107, opened 2019-02-08
Add a new column to the output table from extract_1d


philhodge:
A new NPIXELS column should be added to the 1-D extracted spectrum table, giving the number of pixels that were added together to get the net count rate.  The values will be float, because the lower and upper limits of the extraction region can include fractions of pixels.  The values can vary from pixel to pixel in the dispersion direction, and it can be difficult to determine this value unambiguously from the reference file (if any).  This is needed for the master spectroscopic background subtraction step, issue #2308.

hbushouse:
Fixed by #3108

PR #3108, opened 2019-02-11
Add new NPIXELS column to output extract1d tables.

philhodge:
The number of pixels that were added together to get the net count rate is now saved in a new NPIXELS column.  This involved changes to `extract_1d`, the schemas for `MultiSpecModel` and `SpecModel`, and for a utility function in `master_background`.

The data type of the DQ column has been changed from `int32` to `uint32`.

The functions in extract1d.py that do the actual spectral extraction (except when a reference image is used) now compute and save the results as `float64` instead of `float32`.


jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -21,7 +21,7 @@ allOf:
             - name: ERROR
               datatype: float64
             - name: DQ
-              datatype: int32
+              datatype: uint32
```
hbushouse: I wonder if this will break any existing truth files we have in the reg tests. Data models might throw a validation error now when it tries to load them. Oh well, it has to be done.

jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -30,6 +30,8 @@ allOf:
               datatype: float64
             - name: BERROR
               datatype: float64
+            - name: NPIXELS
+              datatype: float64
```
hbushouse: Does NPIXELS need to be a float? Are there cases where partial pixels are used?

jwst/extract_1d/extract.py
```diff
@@ -2110,7 +2114,10 @@ def extract(self, data, wl_array, verbose):
             The background count rate that was subtracted from the total
             source count rate to get `net`.
 
-        dq : ndarray, 1-D, int32
+        npixels : ndarray, 1-D, float32
```
hbushouse: data type should be listed as float64, right?

jwst/extract_1d/extract.py
```diff
@@ -3254,6 +3272,9 @@ def nans_at_endpoints(wavelength, net, background, dq, verbose):
     background : ndarray
         Array of background values that were subtracted to get `net`.
 
+    npixels : ndarray, float32
```
hbushouse: data type should be listed as float64

jwst/datamodels/schemas/multispec.schema.yaml
philhodge: Yes, but so will adding the NPIXELS column.  That's why I delayed making the change to DQ until also adding NPIXELS, i.e. break them only once.

jwst/datamodels/schemas/multispec.schema.yaml
philhodge: Yes, NPIXELS does need to be a float, because it is often the case that only a fraction of the first or last pixel in the extraction region will be included.

jwst/extract_1d/extract.py
philhodge: Yes, thank you!  I missed these two comments; I'll fix them.

jwst/datamodels/schemas/multispec.schema.yaml
hbushouse: Good point. Hadn't even thought about the extra column.

philhodge:
I'll merge this now so it can be used for testing the master spectroscopic background.  If further changes are needed, I'll create a new PR.

hbushouse:
Fixes #3107

PR #3109, opened 2019-02-12
Master background regression tests

### LINKED ISSUE #3204: Finalize regression tests for master_background manual mode
Issue [JP-566](https://jira.stsci.edu/browse/JP-566) was created by Nadia Dencheva:

Related to github PR #3109

jemorrison:
Closes #3204 

Here is the regression test for NIRSPEC FS. Once I get comments on this I will create a similar test for
NIRSPEC IFU and MIRI LRS. The routine is in test_nirspec_steps_single.py; the test
is test_nirspec_masterbackground_fs_user1d

This test has 3 tests. The first two are used to test the quality of how well the test is working. 
I see them as helping to define truth image. We might want to pull out the first two test later on
and just use the truth image as the sole regression test. 

I have a few figures for test 1.
It takes the output from the MasterBackbround Subtraction step and runs extract1d on the data.
Since this is FS data there are really 4 1-D spectra.
These four 1-D spectra are to compared to the four  1-d spectra produced by running extract 1d on the sci data with no background signal added.
The test loops over each 1-D spectrum and compares the two 1-D spectra.
There will be differences but they should be small. The plot below shows the 1-D spectra and the 1-D user provided spectra
compare_1d_1

When these images are differenced and the resulting data is plotted there are sometimes spikes.
I have been using this difference to test if the tests passes. But I am not sure what to do about the spikes.
diff_1d_1

### DIFF
--- a/jwst/tests_nightly/general/miri/test_miri_steps_single.py
+++ b/jwst/tests_nightly/general/miri/test_miri_steps_single.py
@@ -316,24 +316,25 @@ class TestMIRIMasterBackground_LRS(BaseJWSTTest):
         result = MasterBackgroundStep.call(input_file,
                                            user_background=input_1d_bkg_file,
                                            save_results=True)
-        #_____________________________________________________________________
+        # _____________________________________________________________________
         # Test 1
         # Run extract1D on the master background subtracted data (result)  and
         # the science data with not background added
 
-        # run 1-D extract on results from MasterBackground step                                        
+        # run 1-D extract on results from MasterBackground step
         result_1d = Extract1dStep.call(result, save_results=True)
 
-        # run 1-D extract on original science data without background                                  
+        # run 1-D extract on original science data without background
         input_sci_cal_file = self.get_data(*self.test_dir,
                                             'miri_lrs_sci_cal.fits')
-        # find the 1D extraction of this file                                                          
-        # find Extract1dStep on sci data to use the same version of                                    
-        # this rountine run on both files  rather than running it off line                             
-        # and having the 1-D extracted science file stored as input to step                            
+        # find the 1D extraction of this file
+        # find Extract1dStep on sci data to use the same version of
+        # this rountine run on both files  rather than running it off line
+        # and having the 1-D extracted science file stored as input to step
 
         sci_cal_1d = Extract1dStep.call(input_sci_cal_file, save_results=True)
 
         # Compare the MultiSpec 1-D data
-        atol = 100.
-        rtol = 0.05
+        atol = 100.  # set high for testing
+        rtol = 0.05  # set hight for testing
+        assert_allclose(result_1d,sci_cal_1d,atol=atol,rtol=rtol)
--- a/jwst/tests_nightly/general/miri/test_miri_steps_single.py
+++ b/jwst/tests_nightly/general/miri/test_miri_steps_single.py
@@ -9,13 +9,13 @@ from jwst import datamodels
 from jwst.datamodels import ImageModel, RegionsModel, CubeModel
 from jwst.stpipe import crds_client
 from jwst.lib.set_telescope_pointing import add_wcs
-
 from jwst.tests.base_classes import BaseJWSTTest
 
 from jwst.assign_wcs import AssignWcsStep
 from jwst.cube_build import CubeBuildStep
 from jwst.linearity import LinearityStep
 from jwst.ramp_fitting import RampFitStep
+from jwst.master_background import MasterBackgroundStep
 
 
 @pytest.mark.bigdata
@@ -290,33 +290,33 @@ class TestMIRISetPointing(BaseJWSTTest):
 
 
 @pytest.mark.bigdata
-class TestMIRIMBGMRS(BaseJWSTTest):
+class TestMIRIMBGLRS(BaseJWSTTest):
     input_loc = 'miri'
-    ref_loc = ['test_masterbackground','lrs', 'truth']
-    test_dir = ['test_masterbackground','lrs']
+    ref_loc = ['test_masterbackground', 'lrs', 'truth']
+    test_dir = ['test_masterbackground', 'lrs']
 
     rtol = 0.000001
 
     def test_miri_masterbackground_lrs_user1d(self):
         """
-        Regression test of masterbackgound subtraction with mrs, with user provided 1-D background
+        Regression test of masterbackgound subtraction with lrs, with user provided 1-D background
         """
 
-        # input file has the background added 
+        # input file has the background added
         # Copy original version of file to test file, which will get overwritten by test
         input_file = self.get_data(*self.test_dir,
                                    'miri_lrs_sci_bkg_cal.fits',
                                    docopy=True  # always produce local copy
                                    )
 
-        #user provided 1-D background
-        input_1dbgd_file = self.get_data(*self.test_dir,
+        # user provided 1-D background
+        input_1d_bkg_file = self.get_data(*self.test_dir,
                                          'miri_lrs_bkg_x1d.fits',
                                           docopy=True)
 
-
-        #result = MasterBackGroundStep.call(input_file,user_background=input_1d_bkg_file,
-#                                           save_results=True)
-        #Test up tests
-        # waiting to flush out test_nirspec_steps_single.py, test_nirspec_masterbackground_fs_userid
-
+        result = MasterBackgroundStep.call(input_file,
+                                           user_background=input_1d_bkg_file,
+                                           save_results=True)
+        # Test up tests
+        # waiting to flush out test_nirspec_steps_single.py,
+        # test_nirspec_masterbackground_fs_userid
--- a/jwst/tests_nightly/general/miri/test_miri_steps_single.py
+++ b/jwst/tests_nightly/general/miri/test_miri_steps_single.py
@@ -312,7 +312,8 @@ class TestMIRIMasterBackground_LRS(BaseJWSTTest):
         input_1d_bkg_model = datamodels.open(input_1d_bkg_file)
 
         result = MasterBackgroundStep.call(input_file,
-                                           user_background=input_1d_bkg_file)
+                                           user_background=input_1d_bkg_file,
+                                           save_results=True)
         # _____________________________________________________________________
         # Test 1
         # Run extract1D on the master background subtracted data (result)  and
@@ -345,7 +346,7 @@ class TestMIRIMasterBackground_LRS(BaseJWSTTest):
         if min_user_wave > min_wave: min_wave = min_user_wave
         if max_user_wave < max_wave: max_wave = max_user_wave
 
-        # Compare the data 
+        # Compare the data
         sub_spec = sci_cal_1d_data - result_1d_data
         valid = np.where(np.logical_and(sci_wave > min_wave, sci_wave < max_wave))
         sub_spec = sub_spec[valid]
@@ -384,12 +385,11 @@ class TestMIRIMasterBackground_LRS(BaseJWSTTest):
         # Test 3 Compare background subtracted science data (results)
         #  to a truth file.
 
-        ref_file = self.get_data(*self.ref_loc,
+        truth_file = self.get_data(*self.ref_loc,
                                   'miri_lrs_sci+bkg_masterbackgroundstep.fits')
 
         result_file = result.meta.filename
-        result.save(result_file)
-        outputs = [(result_file, ref_file)]
+        outputs = [(result_file, truth_file)]
         self.compare_outputs(outputs)
         result.close()
         input_sci.close()
--- a/jwst/tests_nightly/general/miri/test_miri_steps_single.py
+++ b/jwst/tests_nightly/general/miri/test_miri_steps_single.py
@@ -309,14 +309,14 @@ class TestMIRIMasterBackground_LRS(BaseJWSTTest):
         # user provided 1-D background
         input_1d_bkg_file = self.get_data(*self.test_dir,
                                          'miri_lrs_bkg_x1d.fits')
+        input_1d_bkg_model = datamodels.open(input_1d_bkg_file)
 
         result = MasterBackgroundStep.call(input_file,
-                                           user_background=input_1d_bkg_file,
-                                           save_results=True)
+                                           user_background=input_1d_bkg_file)
         # _____________________________________________________________________
         # Test 1
         # Run extract1D on the master background subtracted data (result)  and
-        # the science data with not background added
+        # the science data with no background added
 
         # run 1-D extract on results from MasterBackground step
         result_1d = Extract1dStep.call(result)
@@ -325,19 +325,34 @@ class TestMIRIMasterBackground_LRS(BaseJWSTTest):
         input_sci_cal_file = self.get_data(*self.test_dir,
                                             'miri_lrs_sci_cal.fits')
         # find the 1D extraction of this file
-        # find Extract1dStep on sci data to use the same version of
-        # this rountine run on both files  rather than running it off line
-        # and having the 1-D extracted science file stored as input to step
-
         sci_cal_1d = Extract1dStep.call(input_sci_cal_file)
 
-        # Compare the MultiSpec 1-D data
-        atol = 0.0001
-        rtol = 0.000001
-        result_1d_data = result_1d.spec[0].spec_table['flux']
         sci_cal_1d_data = sci_cal_1d.spec[0].spec_table['flux']
+        result_1d_data = result_1d.spec[0].spec_table['flux']
+        sci_wave = sci_cal_1d.spec[0].spec_table['wavelength']
+        # find the valid wavelenth of the user provided spectrun
+        user_wave = input_1d_bkg_model.spec[0].spec_table['wavelength']
+        user_flux = input_1d_bkg_model.spec[0].spec_table['flux']
+        user_wave_valid = np.where(user_flux > 0)
+        min_user_wave = np.amin(user_wave[user_wave_valid])
+        max_user_wave = np.amax(user_wave[user_wave_valid])
+        input_1d_bkg_model.close()
 
-        assert_allclose(result_1d_data, sci_cal_1d_data, atol=atol, rtol=rtol)
+        # find the waverange covered by both user and science                                          
+        sci_wave_valid = np.where(sci_cal_1d_data > 0)
+        min_wave = np.amin(sci_wave[sci_wave_valid])
+        max_wave = np.amax(sci_wave[sci_wave_valid])
+        if min_user_wave > min_wave: min_wave = min_user_wave
+        if max_user_wave < max_wave: max_wave = max_user_wave
+
+        # Compare the data 
+        sub_spec = sci_cal_1d_data - result_1d_data
+        valid = np.where(np.logical_and(sci_wave > min_wave, sci_wave < max_wave))
+        sub_spec = sub_spec[valid]
+        mean_sub = np.absolute(np.nanmean(sub_spec))
+        atol = 0.00005
+        rtol = 0.000001
+        assert_allclose(mean_sub, 0, atol=atol, rtol=rtol)
         # _____________________________________________________________________
         # Test 2
         # Compare result (background subtracted image) to science image with no
@@ -356,10 +371,11 @@ class TestMIRIMasterBackground_LRS(BaseJWSTTest):
         sci_mean = np.nanmean(sci_lrs_region)
         sci_std = np.nanstd(sci_lrs_region)
         upper = sci_mean + sci_std*5.0
-        mask_clean = sci_lrs_region < upper
+        lower = sci_mean - sci_std*5.0
+        mask_clean = np.logical_and(sci_lrs_region < upper, sci_lrs_region > lower)
 
         sub = result_lrs_region - sci_lrs_region
-        mean_sub = np.mean(sub[mask_clean])
+        mean_sub = np.absolute(np.mean(sub[mask_clean]))
 
         atol = 0.5
         rtol = 0.001
@@ -376,3 +392,4 @@ class TestMIRIMasterBackground_LRS(BaseJWSTTest):
         outputs = [(result_file, ref_file)]
         self.compare_outputs(outputs)
         result.close()
+        input_sci.close()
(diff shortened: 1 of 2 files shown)


jemorrison:
I have a few more plots. I took the NIRSPEC fixed slit data and ran it through a python code that basically does what the regression test is doing but it allows me more options to print information since it my copy.
So I have plotted the first FS. Images:
1.  Science+Background,
2. Science image (no background added),
3. third result from subtracting background
4. result - science image 
5. smoothing (result-science image).
The values of the smoothed (result - science image) are 40 - 60 (DN)
I think that seems ok - what do others think ? Right now one of the tests of the regression test is
to do some statistics on the smoothed image - either I need to sigma clip out the high/hot pixels before I smooth or somehow avoid these pixels in the statistics. The are in the original science data.
sci bkg
sci
sci-bkg
sub
sub_smo

hbushouse:
When you say the values of the smoothed (result - science image) are 40-60 DN, do you mean the average is on the order of 40-60 DN, or there are individual non-zero values that peak around 40-60, or what? Theoretically the average should be around zero, with some amount of dispersion around zero. The dispersion would depend on the level of noise that was built into the artificial background signal that was added to the science image.

jemorrison:
So if you look at the second to last image in the list of images above that is the background subtracted image - science image. There are ~5 rows on the top an bottom of the image (orange band) these values are 0. This image is a zoomed up region. 
For this image there are 60,808 pixels, 45,040 of them are nonzero
The mean of the nonzero values is 17.07 and standard dev 136 and median -2.7 
For the smoothed image the results of nonzero elements is similar 48428
Mean of non-zero 15.87 +/- 131.68 and a median of -0.45

First question - why is there is band feature of top and bottom rows exactly 0.
Second you will notice some large values in all the images, I looked in detail at one of these 749,14 (ds9). The input to masterbackground (sci + bkg) is 377,389,200, after step, 377,388,930 - difference of 270 (seems reasonable).
But in the test I compare the output from the step to the science image with no background added
and in this latter data this pixel has a value of 375,619,500. So the difference of this pixel  with the output from the step is 1769440 - REALLY big.  So I want to filter these kinds of pixels out before I smooth the image. So should I just search for large outliers in the science data with no background add
and eliminate these pixels in the subtracted image of results from step - science image.
I would still catch outlier pixels in the result image if they were not outliers in the original science image- which we want to look ok.
Ok ?
Suggestions ?

jdavies-st:
The bands on top and bottom in the extracted 2D cutout are areas not imaged by the slit.  Recall the spectrum is projected onto the detector at an angle and with a bit of curvature, so these are just areas that have no flux from the sky.

I wouldn't smooth image 4 above, but instead just compute statistics on it in the area of the bounding box were the wcs is defined, i.e. does not produce NaNs.  You can use `grid_from_bounding_box` to get these indices and filter out the NaNs.  And then do some sigma-clipped statistics.

hbushouse:
In the first set of spectral plots above, the 2nd plot supposedly includes the difference of the master background step output (which should be close to the original sci with no background) and the original sci with no background (plotted using the solid line). Shouldn't this plot be near zero? It's ranging from values of ~1200 to ~2500. That suggests that the background subtracted by the step is nowhere close to what was inserted into the original sci data.

jwst/tests_nightly/general/miri/test_miri_steps_single.py
```diff
@@ -283,3 +282,97 @@ def test_miri_setpointing(self):
...
+        # find the LRS region
+        bb = result.meta.wcs.bounding_box
+        x, y = grid_from_bounding_box(bb)
+        result_lrs_region = result.data[y.astype(int), x.astype(int)]
+        sci_lrs_region = input_sci.data[y.astype(int), x.astype(int)]
+
+        # do a 5 sigma clip on the science image
+        sci_mean = np.nanmean(sci_lrs_region)
+        sci_std = np.nanmean(sci_lrs_region)
+        upper = sci_mean + sci_std*3.0
```
philhodge: The comment (three lines up) says "5 sigma clip".
Also, do you want to clip on the low side as well?

jwst/tests_nightly/general/miri/test_miri_steps_single.py
```diff
@@ -283,3 +282,97 @@ def test_miri_setpointing(self):
...
+
+        sub = result_lrs_region - sci_lrs_region
+        mean_sub = np.mean(sub[mask_clean])
+        # median_sub = np.mean(sub[mask_clean])
+
+        atol = 2.0
+        assert_allclose(mean_sub, 0, atol=atol)
+
+# ______________________________________________________________________
+        # Test 3 Compare background sutracted science data (results)
```
philhodge: typo:  "sutracted"

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -168,3 +171,107 @@ def test_nrs_fs_single_spec2(self):
...
+        input_1dbkg_file = self.get_data(*self.test_dir,
+                                          'nrs1_bkg_x1d_user.fits',
+                                          docopy=True)
+
+        result = MasterBackgroundStep.call(input_file,
+                                           user_background=input_1dbkg_file,
+                                           save_results=True)
+
+        # _________________________________________________________________________
+        # Test 1 compare 4 FS 1D extracted spectra from sciene data with
```
philhodge: typo:  "sciene"

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -168,3 +171,107 @@ def test_nrs_fs_single_spec2(self):
...
+
+        # run 1-D extract on original science data without background
+        input_sci_cal_file = self.get_data(*self.test_dir,
+                                            'nrs1_sci_cal.fits')
+        # find the 1D extraction of this file
+        # find Extract1dStep on sci data to use the same version of
+        # this rountine run on both files  rather than running it off line
+        # and having the 1-D extracted science file stored as input to step
+
+        sci_cal_1d = Extract1dStep.call(input_sci_cal_file, save_results=True)
```
philhodge: I don't think you need to specify save_results=True, since you're using the object that is returned.

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -168,3 +171,107 @@ def test_nrs_fs_single_spec2(self):
...
+
+        # Compare the FS  1D extracted data. These types of data are
+        #  MultiSpec Models
+        # the tolerance on this compares could be the max of the user 1-D
```
philhodge: "compares" --> "comparison"?

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -168,3 +171,107 @@ def test_nrs_fs_single_spec2(self):
...
+            slit_sci = input_sci_model.slits[i].data
+            slit_result = result.slits[i].data
+
+            # check for outliers in the science image - need a method that iterates
+            # do not check for outliers on result - some of those  outliers might be
+            # something we need to flag
+            sci_mean = np.nanmean(slit_sci)
+            sci_std = np.nanstd(slit_sci)
+            upper = sci_mean + sci_std*3.0
+            mask_clean = slit_sci < upper
```
philhodge: Do you want to clip low values as well?

jwst/tests_nightly/general/miri/test_miri_steps_single.py
```diff
@@ -283,3 +282,97 @@ def test_miri_setpointing(self):
...
+
+        # run 1-D extract on original science data without background
+        input_sci_cal_file = self.get_data(*self.test_dir,
+                                            'miri_lrs_sci_cal.fits')
+        # find the 1D extraction of this file
+        # find Extract1dStep on sci data to use the same version of
+        # this rountine run on both files  rather than running it off line
+        # and having the 1-D extracted science file stored as input to step
+
+        sci_cal_1d = Extract1dStep.call(input_sci_cal_file, save_results=True)
```
philhodge: I don't think you need to specify save_results=True.

jwst/tests_nightly/general/miri/test_miri_steps_single.py
```diff
@@ -283,3 +282,97 @@ def test_miri_setpointing(self):
...
+        result = MasterBackgroundStep.call(input_file,
+                                           user_background=input_1d_bkg_file,
+                                           save_results=True)
+        # _____________________________________________________________________
+        # Test 1
+        # Run extract1D on the master background subtracted data (result)  and
+        # the science data with not background added
+
+        # run 1-D extract on results from MasterBackground step
+        result_1d = Extract1dStep.call(result, save_results=True)
```
philhodge: You can leave out save_results=True.

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
jemorrison: fixed

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
jemorrison: agree

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
jemorrison: I removed this code. I don't think I want to use max of user 1-D to set a tolerance for
the assert. I don't know what to use yet though

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
jemorrison: Ok I was not sure - I was just thinking of things like cosmic rays that could mess things up.
Should I also clip low vales

jemorrison:
The results of comparing the 1-D extracted spectrum of the masterbackground subtraction result and science file look good. I am still struggling with how to set the tolerance for when the test should fail.
The plot of the two 1-D spectra look nice but there are a few points that have  large differences.
So these few point make the test fail. For example the max difference between the spectra is 9.12 and max relative difference is 12. But the vast majority the points have a difference ~0.002 and a relative difference of 1e-5.

nrs_ifu_1d_extract_compare

jemorrison:
The results for the NRS-FS are not as good as IFU.  The comparison of the first FS between the background subtracted science slit  and the science slit has a noticeable shift in the two spectra.
compare_1d_1

When differencing the slits the mean difference is ~250. 
Here is the image of background subtracted - science for the first slit.
Does that seem ok to folks ?
nrs-fs_slit1_difference

jwst/tests_nightly/general/miri/test_miri_steps_single.py
jemorrison: fixed

jwst/tests_nightly/general/miri/test_miri_steps_single.py
jemorrison: fixed

jwst/tests_nightly/general/miri/test_miri_steps_single.py
jemorrison: fixed

jwst/tests_nightly/general/miri/test_miri_steps_single.py
jemorrison: changed to 5 in code

jwst/tests_nightly/general/miri/test_miri_steps_single.py
```diff
@@ -283,3 +282,97 @@ def test_miri_setpointing(self):
...
+
+        # find the LRS region
+        bb = result.meta.wcs.bounding_box
+        x, y = grid_from_bounding_box(bb)
+        result_lrs_region = result.data[y.astype(int), x.astype(int)]
+        sci_lrs_region = input_sci.data[y.astype(int), x.astype(int)]
+
+        # do a 5 sigma clip on the science image
+        sci_mean = np.nanmean(sci_lrs_region)
+        sci_std = np.nanmean(sci_lrs_region)
```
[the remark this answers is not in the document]
jemorrison: fixed

jemorrison:
The results for the MIR-LRS look the best.
The comparison of the extracted spectra is right on.
lrs_1d_extract_compare

The strange extra features in the master background subtracted image are gone.
lrs-background_subtracted

hbushouse:
I wouldn't worry too much about how to compute proper statistics (and clipping outliers) in the differences between the original science (no background added) and master background subtracted results, because we're just using that kind of comparison initially as a sanity check to satisfy ourselves that the MSB step is doing the right thing. Once we're convinced that its results are correct, we should then take the output files it produces and make them "truth" files, and update the regression tests themselves to no longer compare against the original science image, but instead just compare the step output against the truth files. That comparison should produce an exact match (no outliers or noise).

jemorrison:
I think James wanted to start having the regression tests do more than just look a truth file. More tests that  so real  analysis on how well the test is working. I am not sure what I have for these test is what we will want as more than just checking the truth file - but it was my first stab at it.

jemorrison:
I found a big  problem with the NRS-fixed slit. The user background spectra is 0.
So in the test I posted it makes perfect sense now. Nothing was subtracted in the master background.
I forgot to check what the new files looked like yesterday. I have a little program that plots the user background spectra - and I did not run it. I just did and BAM the spectrum is 0.
Nadia - I just grabbed the file again to make sure I grabbed the most current one
but nrs1_bkg_x1d.fits is all zero for FLUX and  NET. I read the first slit in that file and make it the user input to the step. 
It must have something do with turning off photom. 
This is teaching me a need a really good set of testing/plotting tools that I run on the data before
and after I run the step. I am slowly building these up.

jemorrison:
The NRS-Mos 1-D extracted spectra do not look good. 
Here is the 1-D user background for reference. 
user_bkg

Here is the comparison for the first slit - is there a unit issue ? 
compare_1d_1

Here is the comparison for the second slit
compare_1d_2

hbushouse:
Visually inspect the 2-D s2d images. Sometimes the resampling can create sudden jumps, depending on the nature of the input spectra.

jwst/tests_nightly/general/miri/test_miri_steps_single.py
```diff
@@ -287,3 +286,110 @@ def test_miri_setpointing(self):
...
+        input_file = self.get_data(*self.test_dir,
+                                   'miri_lrs_sci+bkg_cal.fits')
+
+        # user provided 1-D background
+        input_1d_bkg_file = self.get_data(*self.test_dir,
+                                         'miri_lrs_bkg_x1d.fits')
+        input_1d_bkg_model = datamodels.open(input_1d_bkg_file)
+
+        result = MasterBackgroundStep.call(input_file,
+                                           user_background=input_1d_bkg_file)
```
jdavies-st: This should be
```
result = MasterBackgroundStep.call(input_file, user_background=input_1d_bkg_file, save_results=True)
```

jwst/tests_nightly/general/miri/test_miri_steps_single.py
```diff
@@ -287,3 +286,110 @@ def test_miri_setpointing(self):
...
+        assert_allclose(mean_sub, 0, atol=atol, rtol=rtol)
+# ______________________________________________________________________
+        # Test 3 Compare background subtracted science data (results)
+        #  to a truth file.
+
+        ref_file = self.get_data(*self.ref_loc,
+                                  'miri_lrs_sci+bkg_masterbackgroundstep.fits')
+
+        result_file = result.meta.filename
+        result.save(result_file)
```
jdavies-st: The above 2 lines are actually overwriting the input as `result.meta.filename` still contains the input filename.  Instead, use the `save_results=True` arg to the step as mentioned further up, the call to `save()` above.

Also, could you rename `ref_file` to `truth_file` throughout these tests?  It makes it clear that these are truth files, not CRDS reference files.

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -168,3 +172,358 @@ def test_nrs_fs_single_spec2(self):
...
+        """
+        # input file has 2-D background image added to it
+        input_file = self.get_data(*self.test_dir,
+                                    'nrs_sci+bkg_cal.fits')
+        # user provided 1-D background was created from the 2-D background image
+        input_1dbkg_file = self.get_data(*self.test_dir,
+                                          'nrs_bkg_user_clean_x1d.fits')
+
+        result = MasterBackgroundStep.call(input_file,
+                                           user_background=input_1dbkg_file)
```
jdavies-st: `save_results=True`

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -168,3 +172,358 @@ def test_nrs_fs_single_spec2(self):
...
+            sub_valid = sub[valid]
+            mean_sub = np.mean(sub_valid[mask_clean])
+            atol = 0.5
+            assert_allclose(np.absolute(mean_sub), 0, atol=atol)
+        input_sci.close()
+        # ______________________________________________________________________
+        # Test 3 Compare background sutracted science data (results)
+        #  to a truth file. This data is MultiSlit data
+        result_file = result.meta.filename
+        result.save(result_file)
```
jdavies-st: Can remove the `save()` call here.

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -168,3 +172,358 @@ def test_nrs_fs_single_spec2(self):
...
+        """
+        # input file has 2-D background image added to it
+
+        input_file = self.get_data(*self.test_dir,
+                                    'prism_sci_bkg_cal.fits')
+        # user provide 1-D background was created from the 2-D background image
+        input_1dbkg_file = self.get_data(*self.test_dir,
+                                          'prism_bkg_x1d.fits')
+
+        result = MasterBackgroundStep.call(input_file,
```
jdavies-st: `save_results=True`

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -168,3 +172,358 @@ def test_nrs_fs_single_spec2(self):
...
+        # run 1-D extract on original science data without background
+        input_sci_cal_file = self.get_data(*self.test_dir,
+                                            'prism_sci_cal.fits')
+        # find the 1D extraction of this file
+        # find Extract1dStep on sci data to use the same version of
+        # this rountine run on both files  rather than running it off line
+        # and having the 1-D extracted science file stored as input to step
+
+        sci_s3d = CubeBuildStep.call(input_sci_cal_file)
+        sci_1d = Extract1dStep.call(sci_s3d, subtract_background=False)
```
jdavies-st: Is there a reason that `CubeBuildStep` and `Extract1dStep` are called here instead of just storing the built, extracted file on Artifactory? It will really speed up the test if we don't have to do this on the fly every time.

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -168,3 +172,358 @@ def test_nrs_fs_single_spec2(self):
...
+            sub_mean = np.absolute(np.nanmean(sub[mask_clean]))
+            atol = 2.0
+            assert_allclose(sub_mean, 0, atol=atol)
+        # ______________________________________________________________________
+        # Test 3 Compare background sutracted science data (results)
+        #  to a truth file. This data is MultiSlit data
+
+        input_sci_model.close()
+        result_file = result.meta.filename
+        result.save(result_file)
```
jdavies-st: Not needed.

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -168,3 +172,358 @@ def test_nrs_fs_single_spec2(self):
...
+        """
+        # input file has 2-D background image added to it
+        input_file = self.get_data(*self.test_dir,
+                                    'nrs_mos_sci+bkg_cal.fits')
+        # user provide 1-D background was created from the 2-D background image
+        input_1dbkg_file = self.get_data(*self.test_dir,
+                                          'nrs_mos_bkg_x1d.fits')
+
+        result = MasterBackgroundStep.call(input_file,
+                                           user_background=input_1dbkg_file)
```
jdavies-st: `save_results=True`

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -168,3 +172,358 @@ def test_nrs_fs_single_spec2(self):
...
+            sub_valid = sub[valid]
+            mean_sub = np.mean(sub_valid[mask_clean])
+            atol = 0.1
+            assert_allclose(np.absolute(mean_sub), 0, atol=atol)
+        # ______________________________________________________________________
+        # Test 3 Compare background sutracted science data (results)
+        #  to a truth file. This data is MultiSlit data
+
+        result_file = result.meta.filename
+        result.save(result_file)
```
jdavies-st: Not needed.

jdavies-st:
I should add that I uploaded the data to `bytesalad` and ran the tests locally.
```
rey> pytest --bigdata --basetemp=/Users/jdavies/scratch jwst/tests_nightly -k masterbackground
========================================================================== test session starts ==========================================================================
platform darwin -- Python 3.6.8, pytest-4.3.0, py-1.8.0, pluggy-0.8.1
plugins: requests-mock-1.5.2, remotedata-0.3.1, openfiles-0.3.2, doctestplus-0.2.0, arraydiff-0.3, ci-watson-0.2
collected 245 items / 241 deselected / 4 selected                                                                                                                       

jwst/tests_nightly/general/miri/test_miri_steps_single.py .                                                                                                       [ 25%]
jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py ...                                                                                               [100%]

============================================================== 4 passed, 241 deselected in 589.77 seconds ===============================================================
```

jemorrison:
I made the suggestions you gave me.
On the reason I call cube_build and then extract 1-D on the science comparison image is when we make changes to cube_build or extract 1-D I did not want it to break this test.
Since we are comparing the 1-D extracted spectra I thought it would be more of an apples to apples test to create it on the fly rather than use a stored one that could differ from the masterbackground 1-D
because of changes to cube-build or 1-d code changes.

But your call - if are the regression testing guru so if you think to compare the master background 1-D to a stored science 1-D I will make that change.

PR #3111, opened 2019-02-12
Update stack_refs to populate output meta data

### LINKED ISSUE #2945: PSFSTACK products have empty headers
Issue [JP-483](https://jira.stsci.edu/browse/JP-483) was created by Howard Bushouse:

`calwebb_coron3` testing by DMS has revealed that _psfstack products are being created with only minimal header info (i.e. just the FITS required keywords), which in turn causes problems for archive ingest when they try to read keywords like ASNTABLE from the header.

This is due to the fact the `stack_refs` creates its output model with just the necessary data arrays and does not copy or inherit any meta info:
https://github.com/spacetelescope/jwst/blob/master/jwst/coron/stack_refs.py#L51

This should be updated to do at least something simple like a model.update() using the first input exposure in the list.

hbushouse:
Updated the stack_refs routine to update the output data model with meta data from the first input model.

Fixes #2945.

### DIFF
--- a/jwst/coron/stack_refs.py
+++ b/jwst/coron/stack_refs.py
@@ -12,6 +12,7 @@ import logging
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
+
 def make_cube(input_models):
     """
     make_cube: Stack all of the integrations from multiple PSF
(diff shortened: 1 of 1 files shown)


hbushouse:
Tested on jw10005 data from DMS test cache. Have confirmed that primary, SCI, and ERR extension headers in the output _psfstack products are now fully populated.

PR #3066, opened 2019-01-28
populate WCS keywords in Level1 files

nden:
This is based and should be merged after #3065.

Close #2932 and #2262.

TODO:

[X] Test with the SIAF database

[] Edit  [this page](https://outerspace.stsci.edu/pages/viewpage.action?spaceKey=JWSTPWG&title=Required+Header+Keywords+for+Test+Exposures) to include the changes in this PR and pass it to the INS teams

PR #3115, opened 2019-02-13
fix a couple of typos

nden:
~#3066 changed the behaviour of set_telescope_pointing for Guider exposures. Previously the WCS keywords were populated by SDP and it was agreed that this will continue to be the case. #3066 changed that and populated WCS keywords for GUIDER exposures. This PR restores the original behaviour where we rely on SDP to populate the keywords.~

It was decided that `set_telescope_pointing` should  populate guider modes as well.

This PR fixes a couple of typos.


nden:
I restored `TYPES_TO_UPDATE` now to work on guider and science imaging modes.

PR #3120, opened 2019-02-14
Rename master background keyword MSTRBKGD > MASTERBG

jdavies-st:
The keyword for recording a user background file cannot be pronounced or remembered.  This is clearer.

Also made the attribute names in the datamodel clearer.

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -2133,7 +2133,7 @@ properties:
                   title: Channel ID
                   type: string
       master_background_file:
-        title: User-supplied master background file
+        title: Master background file used
         type: string
         fits_keyword: MASTERBG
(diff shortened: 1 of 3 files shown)


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1979,8 +1979,8 @@ properties:
             type: string
             fits_keyword: S_LINEAR
             blend_table: True
-          master_back_sub:
-            title: Master Background subtraction
+          master_background:
```
hbushouse: This change now requires a change to https://jira.stsci.edu/browse/JWSTKD-300. The data model name needs to be updated in the "Additional Fields" tab of the form.

jdavies-st: I updated this in the JWSTKD ticket.

PR #3126, opened 2019-02-15
Move MASTERBG keyword to background section of schema

jdavies-st:
Change
```
meta.master_background_file
```
to
```
meta.background.master_background_file
```
in the core schema, as the `meta.background` block already exists and is a natural place to put this property.


jdavies-st:
Also, this will make our schema consistent with the JWSTKD request to add this keyword.

https://jira.stsci.edu/projects/JWSTKD/issues/JWSTKD-301

PR #3157, opened 2019-02-19
add a boolean  'subtract_background' parameter 

### LINKED ISSUE #3142: extract_1d should accept a bool parameter "subtract_background"
Issue [JP-532](https://jira.stsci.edu/browse/JP-532) was created by Nadia Dencheva:

The background subtraction in extract_1d is controlled by a parameter in the extract_1d reference file. The step should accept a command line parameter to allow the background subtraction to be easily turned off, e.g. from the master background subtraction step.

nden:
This PR adds a parameter to extract_1d which allows a user to turn on and off the local background subtraction.
```
subtract_background : bool or None
        A flag which indicates whether the background should be subtracted.
        If None, the value in the extract_1d reference file will be used.
        If not None, this parameter overrides the value in the
        extract_1d reference file.
```
Close #3142


nden:
As far as I can tell the changes in this PR should not caused this failure

https://travis-ci.org/nden/jwst/jobs/495656270#L1406

@jdavies Any ideas?

jwst/pipeline/extract_1d.cfg
```diff
@@ -3,3 +3,4 @@ class = "jwst.extract_1d.Extract1dStep"
 
 log_increment = 50
 smoothing_length = 0
+subtract_background = None
```
jdavies-st: This might be the offending item causing the test failure.  `None` is special with ConfigObj.  If you pass it `None` in a parameter, it will read it as a string.  So our example configs should not have anything assigned to `None` in them.

jwst/extract_1d/ifu.py
```diff
@@ -50,7 +56,8 @@ def ifu_extract1d(input_model, refname, source_type):
         slitname = "ANY"
 
     extract_params = ifu_extract_parameters(refname, slitname, source_type)
-
+    if subtract_background is not None:
+        extract_params['subtract_background'] = subtract_background
```
philhodge: For IFU data, there already is a `subtract_background` key in `extract_params`, so I think the user should get an info message if the reference file specified that background subtraction should be done but the user specified that it should be turned off, or vice versa.  Also, background subtraction is currently only done for a point source, so if the user specified background subtraction for a target that is an extended, perhaps we really should subtract a background (i.e. maybe the target is only a little extended.

jwst/extract_1d/extract1d.py
```diff
@@ -85,6 +90,14 @@ def extract1d(image, lambdas, disp_range,
         For each column, this is the number of pixels that were added
         together to get `countrate`.
     """
+    if not subtract_background:
```
philhodge: This function can be called in a loop over planes of a 3-D array, e.g. TSO data, which could result in a very large number of info messages.  For the case that the reference is a JSON file, I would suggest checking the `subtract_background` flag in extract.py, in the `__init__` for `ExtractModel`, in this section:

    # Coefficients for source (i.e. target) and background limits and
    # corresponding polynomial functions.
    self.src_coeff = copy.deepcopy(src_coeff)
    self.bkg_coeff = copy.deepcopy(bkg_coeff)

If `bkg_coeff` was specified in the reference file, but the user said to not do background subtraction, just set `self.bkg_coeff` to `None` instead of copying `bkg_coeff`.  Or if the user said that background subtraction should be done but `bkg_coeff` was not defined in the reference file, you could log a warning to say that it can't be done because there's no information as to where to extract the background (as you did here).

jwst/pipeline/extract_1d.cfg
nden: You are right that None can only be used as a default value in the spec. But I would have expected a ConfigObj validation error, not this bizarre failure.

jdavies-st: I would have as well.  If you remove that line, does it make the error go away for that test?  I recall seeing something like this when I added the `MasterBackgroundStep` and made the same mistake of putting an optional parameter set to `None` in the example .cfg file.

jdavies-st: Hmm, but maybe there's something a little more subtle going on.  I believe @stscieisenhamer originally wrote this suffix test.  It looks like it is getting this "new" suffix from the example .cfg file `name` entry.  But it should have been getting it before.  I don't know why it was not getting before and now it is.  Hmm.

jwst/extract_1d/ifu.py
nden: @philhodge If background subtraction for an extended source is enabled how should `radius`, `inner_bkg` and `outer_bkg` be defined? Should they also be added as parameters?

philhodge: My understanding is that background subtraction should not be enabled for an extended source.  It would be possible to do so, however, using a reference image.  But parameters `radius`, `inner_bkg`, and `outer_bkg` should not be defined for IFU data.

But that's not what we were talking about; we want to be able to extract the background of an IFU image and ignore the source(s).  That is, the background is what we want as the result, i.e. to populate the FLUX column.  We can do that (or will be able to) by using a reference image.  After creating a segmentation map to isolate the sources, the non-source regions are the background.  So we would create an integer mask that is set to 0 for every pixel that has significant source flux, and set to 1 for all other pixels.  Using that as a reference image, extract_1d would give us the background flux for an IFU cube.

jwst/pipeline/extract_1d.cfg
jdavies-st: Yeah, there's definitely some subtle interaction going on.  If I you remove the
```
subtract_background = None
```
line from the config file, then the tests pass.  Hmm.  Would be nice to understand that.

stscieisenhamer:
About the travis error: That is a canary test to ensure that anything that could produce a suffix has been accounted for. The error is that the suffix "extract_1d" has vanished, implying that somehow the step `Extract1dStep` is no longer considered a Step. How the test works is basically going through and instantiating anything that is a Step. My guess is that the instantiation of `Extract1dStep` is failing. A bad ConfigObj specification would cause that.

nden:
Thanks @stscieisenhamer This now makes more sense.

stscieisenhamer:
What I would wonder about is why this does not fail on jenkins. I believe this is a "unit" test and should be running on all CI.

jdavies-st:
@stscieisenhamer, it does fail on Jenkins.  But right now
```
pytest -r s --basetemp=test_results --junitxml=results.xml || true
```
means that a single failure does not give us a big red `X`.

PR #3166, opened 2019-02-21
Add pytest.ini file to select correct junit test report family 

rendinam:
The Jenkins test reporting plugin requires a correctly constructed schema of the report produced by pytest. This `pytest.ini` directive provides that.
Honored as of pytest 2.4.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+junit_family=xunit2
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,0 @@
-[pytest]
-junit_family=xunit2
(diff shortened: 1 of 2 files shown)


jdavies-st:
Horray!

Could it pull this from the `[tool:pytest]` section of `setup.cfg`?  Or does it need to by in a `pytest.ini` file?

rendinam:
It should work via a `[tool:pytest]` section of `setup.cfg`. Would you prefer that?

jdavies-st:
I think so.  Currently that's where we have our `pytest` setup.  Would be nice to keep it all in one place.

This is going to fix that `pytest foo || true` Jenkins stuff, right?  If so, fantastic.

rendinam:
Yes. It should correct the errant job `SUCCESS` status when one or more tests fail.

Issue #3163, opened 2019-02-21
Modify combine_1d for combining background data


philhodge:
For the case that the FLUX column of the input extract1d tables is actually background, e.g. when combining background data to make a master spectroscopic background spectrum, the `combine_1d` step must scale the values read from the FLUX column by dividing by the NPIXELS column.  This is because NPIXELS can vary from row to row within a table, and it can differ from one table to another; the background values have to be put on the same scale (e.g. background level per pixel) before they can be combined.

Since the `combine_1d` step should treat background data differently from target data, the step must be able to distinguish between these two types of input.  A new command-line parameter could be added for this purpose.  The parameter could be `background`, a boolean with default value False; True would mean that all the input extract1d tables contain background rather than target data.

jdavies-st:
A boolean `background` would work well.  Also, this parameter will have to be switched by the input association.  The `exptype` of the association member can have `"background"` as a type, so this could also inform the function that it should treat it as background.

hbushouse:
We should keep in mind that upcoming changes to the level-2b processing flow, and the `photom` step in particular, will be applying flux conversion to the pixel values, where the conversion factors will be scaled in such a way that when `extract_1d` is run using an agreed upon standard aperture size (for a given instrument mode) the resulting 1-D spectrum will be in flux units scaled to an infinite aperture. So only in cases where we use an aperture height in `extract_1d` that is different than the standard (which we may do for extracting background) do we need to worry about doing any rescaling of the resulting 1-D flux values. That should likely be added into `extract_1d` itself, such that everything that comes out will always have the correct infinite aperture flux? That may make a `background` option in `combine_1d` unnecessary.

But then our routine that interpolates 1-D background spectra back to 2-D image space will also need to know how to rescale properly to a per pixel background flux (e.g. it may need to know what the standard aperture size - number of pixels - is for a given instrument mode).

philhodge:
[...]

PR #3180, opened 2019-02-22
Modify combine_1d for combining background data

philhodge:
For the case that the FLUX column of the input extract1d tables is actually background, the combine_1d step now scales the values read from the FLUX column by dividing by the NPIXELS column.  Also in this case, NPIXELS is included as a factor in the weight (if there's more data, the weight should be higher).

A new command-line parameter `background` was added, to allow specifying that the data are actually background rather than target flux.

The default for weighting the input tables has been changed from the integration time to the exposure time; see parameter `exptime_key`.

Closes issue #3163.


jwst/combine_1d/combine1d.py
```diff
@@ -596,22 +619,32 @@ def check_exptime(exptime_key):
...
+    background : bool, default=False
+        If the flux data are actually background rather than a target
+        spectrum, `background` should be set to True.  In this case, the
+        values read from the flux column of each input spectrum will be
+        divided by the npixels column (if that column exists).  This is
+        to convert the values to background per pixel.
     """
 
+    log.debug("Using exptime_key = {}.".format(exptime_key))
+    log.debug("The FLUX data will be treated as background data.")
```
hbushouse: Shouldn't this message be part of an if-block that depends on the value of `background`? Right now it looks like it's always printed (whenever debug is enabled).

Issue #3143, opened 2019-02-18
Implement improved automated testing


stscijgbot:
Issue [JP-531](https://jira.stsci.edu/browse/JP-531) was created by Nadia Dencheva:

Improve the jwst pipeline test matrices to include automatic testing with dev versions of dependencies and public versions of released code.

jdavies-st:
- [ ] Make the per-PR builds that use the `pip install` of `jwst` use released versions of the dependencies `astropy`, `asdf` and `gwcs`.  Currently these 3 dependencies are installed from master on github.
 - [ ] Figure out which elements of a dev vs released dependencies should be run on which services, i.e. Travis CI or Jenkins.
 - [ ] Move Jenkins RT builds to use released dependencies.

PR #3183, opened 2019-02-25
Switch regression tests to build with publicly-released dependencies

jdavies-st:
- use `pip install -e .[test]` to install `jwst` and its released dependencies listed in `setup.py`
- use `numpy` from PyPi instead of from Anaconda, i..e one without `mkl`
- because no `mkl`, use `pytest-xdist` to parallelize our builds (from 5.5 hours down to ~2 hours)

Addresses #3143

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -38,15 +38,14 @@ data_config.match_prefix = '(.*)_result' // .json is appended automatically
 
 bc = new BuildConfig()
 bc.nodetype = 'jwst'
+bc.name = "released dependencies"
 bc.env_vars = test_env
 bc.conda_packages = ["python=${python_version}"]
 bc.build_cmds = [
     "pip install ${pip_install_args} numpy~=${numpy_version}",
-    "pip install ${pip_install_args} develop",
+    "python setup.py ${pip_install_args} develop",
     "pip install ${pip_install_args} -e .[test]",
     "pip install pytest-xdist",
-    "pip freeze",
-    "conda list",
 ]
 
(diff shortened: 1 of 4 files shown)


jdavies-st:
The Jenkins builds from this PR for testing purposes are here

https://boyle.stsci.edu:8081/job/RT/job/jdavies-dev/

Note that our current setup using `conda` and `astroconda-dev` to install everything takes ~5.5 hours to build and run tests.  The initial implementation of this using `pip` took ~8.5 hours to run, largely because of use the version of `numpy`.  `pip` intalls the version built by the Numpy folks without on-chip parallelization.  Anaconda's versions of `numpy` uses Intel's `mkl` library to parallelize linear algebra functions on chip cores.

To regain some of the speed with the switch to `pip`, we can parallelize by farming out tests to the 32 cores available on jwcalibdev.  This gets the runtime on this PR down to ~1h45m.

jdavies-st:
Looking at the test builds on Jenkins above, one can see that there are some failing tests due to a few pixels being different between output and truth files.  This is because of a different version of `numpy`.  I'm not too worried about this, though it would be nice to make our code more robust where needed, i.e. use `float64`.

The real problem is the following test

https://boyle.stsci.edu:8081/job/RT/job/jdavies-dev/84/testReport/junit/jwst.tests_nightly.general.nircam.test_nrc_image3_1/TestImage3Pipeline1/___test_image3_pipeline2/

which reports that `tweakreg` cannot be run because `chelp` C extension cannot be found.  Odd.  @jhunkeler, any idea what is going on here?

PR #3185, opened 2019-02-25
EXP: See if ci skip directive prevents merging

pllim:
This is an experiment for @nden for my claim that the merge button will be grayed out if there is a `[ci skip]` directive in the commit and the repo requires passing test before merge.


pllim:
Ah ha! So I think we were disagreeing because I was referring to a behavior seen with Travis CI but you don't use Travis CI anymore here, it seems. So we were both correct, in a way. :stuck_out_tongue:

pllim:
Either that, or Travis itself changed behavior, as I do see it kicking off for another PR here. :thinking:

pllim:
p.s. If I see what I thought I used to see again in another repo, I'll try to remember to ping you.

PR #3186, opened 2019-02-25
add change log entry and log message

nden:
Follow up on #3157 - local changes I had did not make it into the previous PR.

### DIFF
--- a/jwst/extract_1d/ifu.py
+++ b/jwst/extract_1d/ifu.py
@@ -56,8 +56,14 @@ def ifu_extract1d(input_model, refname, source_type, subtract_background):
         slitname = "ANY"
 
     extract_params = ifu_extract_parameters(refname, slitname, source_type)
+    sub_bkg = extract_params['subtract_background']
     if subtract_background is not None:
         extract_params['subtract_background'] = subtract_background
+    if subtract_background and not sub_bkg:
+        log.info("Background subtraction was enabled - performing it.")
+    elif sub_bkg and not subtract_background:
+        log.info("Background subtraction was disabled - skipping it.")
+
     if extract_params:
         (ra, dec, wavelength, net, background, npixels, dq) = extract_ifu(
                         input_model, source_type, extract_params)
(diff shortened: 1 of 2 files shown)


jwst/extract_1d/ifu.py
```diff
@@ -56,8 +56,14 @@ def ifu_extract1d(input_model, refname, source_type, subtract_background):
         slitname = "ANY"
 
     extract_params = ifu_extract_parameters(refname, slitname, source_type)
+    sub_bkg = extract_params['subtract_background']
     if subtract_background is not None:
         extract_params['subtract_background'] = subtract_background
+    if subtract_background and not sub_bkg:
+        log.info("Background subtraction was enabled - performing it.")
+    elif sub_bkg and not subtract_background:
```
hbushouse: So this handles the case where subtract_background was enabled in the reference file, but explicitly disabled (set to False) in the command-line parameter, so the command-line param overrides and you skip the subtraction, right? As opposed to the case where the command-line param wasn't set at all.

jwst/extract_1d/ifu.py
```diff
@@ -56,8 +56,14 @@ def ifu_extract1d(input_model, refname, source_type, subtract_background):
         slitname = "ANY"
 
     extract_params = ifu_extract_parameters(refname, slitname, source_type)
+    sub_bkg = extract_params['subtract_background']
     if subtract_background is not None:
         extract_params['subtract_background'] = subtract_background
+    if subtract_background and not sub_bkg:
```
hbushouse: What if the command-line param subtract_background wasn't set at all, i.e. it's None? Will this test fail? Or is it the case that if the user doesn't set it, it still gets assigned a default value (False, I believe) by the step class, so that it's never None in this part of the code? In which case the test in line 60 isn't really necessary (if it can never be None).

jwst/extract_1d/ifu.py
nden: correct

jwst/extract_1d/ifu.py
nden: This should have been indented. Will fix it shortly.

nden:
Rebased several times to fix conflicts in the change log, merging now.

PR #3187, opened 2019-02-26
Refactor CI scripts to fix build issues

jdavies-st:
An attempt to fix our broken builds on Jenkins.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ env:
         - MC_URL=https://repo.continuum.io/miniconda
         - PIP_DEPENDENCIES='.[test]'
         - PYTHON_VERSION=3.6
-        - NUMPY_VERSION=1.15
+        - NUMPY_VERSION=1.16
 
     matrix:
         - SETUP_CMD='install'
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,12 +46,10 @@ matrix:
 
         # Run tests
         - env: TEST_COMMAND='python setup.py test'
-               CONDA_DEPENDENCIES="$CONDA_DEPENDENCIES"
                PIP_DEPENDENCIES='.[test]'
 
         # Build sphinx documentation with warnings
         - env: TEST_COMMAND='python setup.py build_sphinx -W'
-               CONDA_DEPENDENCIES="$CONDA_DEPENDENCIES"
                PIP_DEPENDENCIES='.[docs]'
 
         # PEP8 check
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,9 +80,9 @@ before_install:
   - bash $MC_INST -b -p $HOME/miniconda
   - export PATH=$HOME/miniconda/bin:$PATH
   - conda config --set always_yes yes --set changeps1 no
-  - conda config --set auto_update_conda false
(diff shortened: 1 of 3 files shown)


jdavies-st:
For sure.  I was just trying to do some debugging by doing 1 change at a time.  First change tests passed.  Added those back in now.

jhunkeler:
Are you having problems with the C extensions not being discoverable? I ran into that with `drizzlepac` and the only way I could get it to work was with `python setup.py develop`... for some reason `pip install -e` wasn't installing the extension(s) into the project root like it's supposed to.

Maybe there's a bug in the version of pip we're using. Hmm.

jdavies-st:
Yes, we've run into that issue here on our `ssbjenkins` per-PR build.  This is trying to fix that.  I also ran into it over at #3183.  Perhaps related issues, but seemingly slightly different.

jhunkeler:
https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwst/detail/PR-3187/2/pipeline/25#step-140-log-604

I've noticed this behavior is specific to `conda`-based environments, and happens regardless if their `numpy` or `mkl` packages are installed before `pip` installs the latest wheel. The circumstances behind what triggers this are a mystery to me. If I spawn an `astroconda/buildsys` docker container on my system and go through the build (`pip install -e .[test]`) then the C `chelp` extension is in the correct place and imports. The unit tests pass too using the 1.16 wheel from pypi as well.

On the CI system using the same docker image it crashes and burns.

_Hmm_

jdavies-st:
Well, this is frustrating.  In a conda environment that I created locally were this is fixed:
```
In [1]: import numpy as np 
   ...: import subprocess                                                                                                                                                

In [2]: try: 
   ...:     multiarray = np.core._multiarray_umath.__file__ 
   ...: except: 
   ...:     multiarray = np.core.multiarray.__file__ 
   ...:                                                                                                                                                                  

In [4]: print("Multiarray is:", multiarray) 
   ...: command = "otool -L {}".format(multiarray) 
   ...: print("LDD of multiarray is:") 
   ...: print(subprocess.getoutput(command))                                                                                                                             
Multiarray is: /Users/jdavies/miniconda3/envs/test16/lib/python3.6/site-packages/numpy/core/_multiarray_umath.cpython-36m-darwin.so
LDD of multiarray is:
/Users/jdavies/miniconda3/envs/test16/lib/python3.6/site-packages/numpy/core/_multiarray_umath.cpython-36m-darwin.so:
	@loader_path/../.dylibs/libopenblasp-r0.3.5.dev.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
```
FWIW this enivronment was created with
```
rey> conda create -n test16 python=3.6

rey> conda activate test16

(test16) rey> pip install --index-url https://bytesalad.stsci.edu/artifactory/api/pypi/datb-pypi-virtual/simple numpy==1.16.1

pip install --index-url https://bytesalad.stsci.edu/artifactory/api/pypi/datb-pypi-virtual/simple -e .[test]
```

PR #3188, opened 2019-02-26
Add unit tests for expand_to_2d

philhodge:
There are now unit tests for `expand_to_2d`.

expand_to_2d.py was also modified to accept the output of the `combine_1d` step as the user-supplied background spectrum.

combinedspec.schema.json was deleted, as it is obsolete.
See issue #2308.


jwst/master_background/expand_to_2d.py
```diff
@@ -29,14 +29,18 @@ def expand_to_2d(input, m_bkg_spec):
     """
 
     with datamodels.open(m_bkg_spec) as bkg:
-        if len(bkg.spec) > 1:
-            log.warning("The input 1-D spectrum contains multiple spectra")
-        tab_wavelength = bkg.spec[0].spec_table['wavelength'].copy()
+        if hasattr(bkg, 'spec'):                # MultiSpecModel
+            if len(bkg.spec) > 1:
+                log.warning("The input 1-D spectrum contains multiple spectra")
```
nden: Would it be useful to mention that only the first spec will be used?

CHANGES.rst
```diff
@@ -7,6 +7,9 @@ combine_1d
    error, and net by 1 / NPIXELS, and include NPIXELS in the weight;
    changed the default for ``exptime_key`` to "exposure_time". [#3180]
 
+ - Added unit tests for expand_to_2d.  Support CombinedSpecModel data
```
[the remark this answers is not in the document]
philhodge: Yes, of course; thanks.

jwst/master_background/expand_to_2d.py
```diff
@@ -29,14 +29,18 @@ def expand_to_2d(input, m_bkg_spec):
...
+            spec_table = bkg.spec[0].spec_table
+        else:                                   # CombinedSpecModel
+            spec_table = bkg.spec_table
+        tab_wavelength = spec_table['wavelength'].copy()
         try:
-            tab_npixels = bkg.spec[0].spec_table['npixels'].copy()
+            tab_npixels = spec_table['npixels'].copy()
         except KeyError:
             tab_npixels = np.ones_like(tab_wavelength)
```
nden: Is it true that the pipeline will always have this column populated with values > 1 and the column can be missing  only in  a user provided file? If so is it worth printing a warning to the user.

jwst/master_background/expand_to_2d.py
```diff
@@ -29,14 +29,18 @@ def expand_to_2d(input, m_bkg_spec):
...
-        tab_background = bkg.spec[0].spec_table['flux'] / tab_npixels
+        tab_background = spec_table['flux'] / tab_npixels
```
nden: I wonder if `SpecModel` should have some kind of data validation, making sure that the arrays are the same shape, mostly for the user supplied file case.
Some validation is done in  `create_background` but there's no option to give an array of npixels.

Issue #3189, opened 2019-02-26
Set flux to zero if wavelength is out of range


philhodge:
The `extract_1d` step uses the RELSENS extension to convert net count rate to flux.  For some modes, the wavelengths in the RELSENS table do not cover the full range of wavelengths for the science data.  The conversion factor is currently set to 1 for wavelength bins that are outside the range of the wavelengths in RELSENS, resulting in flux values for those pixels that are the same as the net.  The flux should instead be set to 0 for those pixels.

hbushouse:
fixed by #3190

PR #3190, opened 2019-02-27
Set flux to zero if wavelength is out of range

philhodge:
If the RELSENS information is used to convert NET to FLUX in `extract_1d`, the flux is now set to zero for each pixel for which the wavelength for that pixel is outside the range of the RELSENS table.

Closes issue #3189.


jdavies-st:
Though it looks like `test_interpolate_response()` needs to be updated.

philhodge:
> Though it looks like test_interpolate_response() needs to be updated.

Ah, thank you!

PR #3195, opened 2019-02-28
Remove conda from Travis CI build

jdavies-st:
- Rely on publicly-released dependencies
- Don't use `conda` in Travis CI build/test matrix
- New version of `flake8` founda  bug in the ami code

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 python:
-  - "pypy3.6"
+  - "3.6"
 sudo: false
 
 # The apt packages below are needed for sphinx builds
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 dist: xenial
 language: python
-python: 3.6.8
+python:
+  - 3.6.8
 sudo: false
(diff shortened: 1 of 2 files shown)


jhunkeler:
Quick question: Did you want to test 3.6 and 3.7, or just 3.6? According to the docs we can use 3.7 if you set `dist: xenial`.

https://docs.travis-ci.com/user/languages/python/#specifying-python-versions

jdavies-st:
Currently DMS is using 3.6.  But yeah, I would be happy to move to 3.7 as long as everyone is on board.  We can discuss.  Currently, I'm just going to keep it at 3.6.

jdavies-st:
So the other option for the Travis CI tests would be to go the route we're currently going in Jenkins builds - use `conda` to install `python`, but then install everything else via `pip` or `setup.py install` - a hybrid approach.  But of course we've been having problems with that approach, as documented in #3187.

Given the Travis YAML interface, not using `conda` in our Travis CI script makes it much simpler to implement.  Happy to discuss.

PR #3197, opened 2019-03-01
Add exit code to set_bary_helio_times script

### LINKED ISSUE #3196: set_bary_helio script needs an exit code
Issue [JP-563](https://jira.stsci.edu/browse/JP-563) was created by Howard Bushouse:

[~mswam] reports:

We noticed in some of the recent processing runs using SCSB 0.13.0 that occasionally set_bary_helio.py will fail with a DB connection problem, but we still get back a zero exitCode from the script: 

Some AUTO failures were seen around 2/18/2019 where the SCSB set_bary_helio.py script was failing with a DB error, yet the wrapper script was getting back exitCode=0, and continuing on as if successful. This later caused an AUTO failure when the baseline comparison of bary-helio columns in a FITS table extension came out with -1 (defaulted) values:

set_bary_helio_times.py /ifs/archive/dev/jwst/store/dmsauto/Nightly_tests/test_dmsauto_20190218_01/regr_10003/pipe/sdp/fits/jw82600025001_02101_00003_nrcblong_uncal.fits 
INFO:root:Starting set_bary_helio_times task 
INFO:jwst.timeconversion:Ephemeris connect info: eserver=DWJWDMSDB1 edb=jwdpmetrics_auto2 
INFO:root:Calculating GROUP times 
INFO:jwst.timeconversion:Ephemeris connect info: eserver=DWJWDMSDB1 edb=jwdpmetrics_auto2 
WARNING:root:Error in calculating times. Filling with invalid dates. 
Error is "(20017, b'DB-Lib error message 20017, severity 9:\nUnexpected EOF from the server (DWJWDMSDB1:1433)\nDB-Lib error message 20002, severity 9:\nAdaptive Server connection failed (DWJWDMSDB1:1433)\n')" 
INFO:root:Completed set_bary_helio_times task 
set savestatus = 0

Can you change the set_bary_helio.py code to return a non-zero value if such a DB error occurs? (or just let the exception “leak out” to the caller?)  Any non-zero exit will do.

We already have automatic retries built-in to our wrapper script that calls set_bary_helio.py and would try to run the script again, up to 5 times after a short delay, to see if the DB failure clears, but currently we can’t detect that there was a problem, so we don’t know to try a re-run.  I suspect there is some kind of temporary DB connection issue (contention between tasks) that causes once-in-a-while failures, and we added the retry code a good while ago to catch and try and correct those cases.

As is, we don’t get later failures (calibration is OK), but the data products are missing some updated keywords and table content (our baseline diffs picked that up).

hbushouse:
This adds a call to "sys.exit(1)" if the script fails to be able to compute the converted times. If everything goes OK, it exits with no explicit code, which is assumed to default to 0. Python experts, please comment on whether plain "exit" or "sys.exit" is preferred.

Also includes minor updates to other scripts in the package to remove future import statements (should've been taken out long ago, now that we're requiring Python 3.x for the jwst package), as well as random pep8 updates to other scripts.

Fixes #3196

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,8 +5,8 @@ combine_1d
 ----------
 
 - Added parameter ``background``; for background data, scale the flux,
-   error, and net by 1 / NPIXELS, and include NPIXELS in the weight;
-   changed the default for ``exptime_key`` to "exposure_time". [#3180]
+  error, and net by 1 / NPIXELS, and include NPIXELS in the weight;
+  changed the default for ``exptime_key`` to "exposure_time". [#3180]
 
 extract_1d
 ----------
@@ -24,6 +24,11 @@ master_background
 
  - Added unit tests for expand_to_2d.  Support CombinedSpecModel data
    for the 1-D user-supplied background spectrum. [#3188]
+
+set_bary_helio_times
+--------------------
+
+ - Added exit code of 1 when unable to compute converted times. [#3197]
 
 set_telescope_pointing
 ----------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,7 +28,7 @@ master_background
 set_bary_helio_times
 --------------------
 
- - Added exit code of 1 when unable to compute converted times. [#3197]
+ - Raise an exception when unable to compute converted times. [#3197]
 
 set_telescope_pointing
 ----------------------
--- a/scripts/move_wcs.py
+++ b/scripts/move_wcs.py
@@ -36,7 +36,7 @@ def _collect_wcs_keywords(f):
     # Get keywords to go in SCI header from the datamodels schema
     dm = datamodels.open(f, pass_invalid_values=True)
     stsci_wcs_kw = []
-    for k,v in datamodels.schema.build_schema2fits_dict(dm.meta._schema).items():
+    for k, v in datamodels.schema.build_schema2fits_dict(dm.meta._schema).items():
         if v[1] == 'SCI':
             if v[0] != 'WCSAXES':
                 stsci_wcs_kw.append(v[0])
@@ -96,14 +96,14 @@ def add_default_keywords(new_hdr):
     wcsaxes = new_hdr['WCSAXES']
     if wcsaxes == 3:
         default_pc = {"PC1_1": 1, "PC1_2": 0, "PC1_3": 0,
-                      "PC2_1": 0 , "PC2_2": 1, "PC2_3": 0,
+                      "PC2_1": 0, "PC2_2": 1, "PC2_3": 0,
                       "PC3_1": 0, "PC3_2": 0, "PC3_3": 1
                      }
         default_cunit = {"CUNIT1": "deg", "CUNIT2": "deg", "CUNIT3": "um"}
         default_ctype = {"CTYPE1": "RA---TAN", "CTYPE2": "DEC--TAN", "CTYPE3": "WAVE"}
     elif wcsaxes == 2:
         default_pc = {"PC1_1": 1, "PC1_2": 0,
-                      "PC2_1": 0 , "PC2_2": 1,
+                      "PC2_1": 0, "PC2_2": 1,
                      }
         default_cunit = {"CUNIT1": "deg", "CUNIT2": "deg"}
         default_ctype = {"CTYPE1": "RA---TAN", "CTYPE2": "DEC--TAN"}
@@ -136,5 +136,3 @@ if __name__ == '__main__':
             files = glob.glob(files)
 
     move_wcs(files)
-
-
--- a/scripts/set_bary_helio_times.py
+++ b/scripts/set_bary_helio_times.py
@@ -36,7 +36,6 @@ times. It also updates the extension that contains the times
 for each group by adding the necessary columns to that table.
 '''
 
-import sys
 import argparse
 import logging
 
@@ -89,7 +88,6 @@ def set_bary_helio_times(filename, jwstpos=None):
     pheader['HELIDELT'] = (hstrtime - starttime) * 86400.
 
     # Now modify the table
-    success = False
     try:
         tabhdu = hdul['GROUP']
     except KeyError:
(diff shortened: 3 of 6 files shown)


scripts/set_bary_helio_times.py
```diff
@@ -129,6 +129,8 @@ def set_bary_helio_times(filename, jwstpos=None):
             hcol = fits.Column(
                 name='helio_end_time', format='D', unit='MJD', array=htimes
             )
+            success = True
```
philhodge: I would suggest changing `logging.warning` to `logging.error` (line 114).

hbushouse:
> Presuming that this condition is a true "hard error", one that should prevent all further processing, then all looks good.
> 
> And yes, `sys.exit` is the appropriate call.

If you mean no further processing within just this script, yes. SDP wants their workflow manager to trap the exit condition and try again. And if repeated attempts fail, then they give up and say this dataset can't be processed. My concern, based on comments on the web, is that if SDP is calling our script from within another (higher level) python script, is our use of "sys.exit(1)" here going to cause an exit from ALL scripts, thereby not even allowing their error trapper to react?

philhodge:
Why not re-raise the exception?

stscieisenhamer:
Since the original comment from SDP involved "system exit status", and knowing how they run everything, `sys.exit` is appropriate: SDP is calling the script from a bash script, so our python script is running in its own process.  This is how the condition of "no science data" from calwebb_spec2 is handled.

stscieisenhamer:
FYI: [Difference between exit() and sys.exit() in Python](https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python)

stscieisenhamer:
However, @philhodge is right. Ultimately, if the error is a hard error, then simply raising/not catching the exception is the more appropriate thing to do.  Exceptions will cause non-zero system exits.

hbushouse:
It's debatable as to whether this particular exception should be treated as a hard error. When this occurs the script can't really do what it's supposed to be doing, but it's currently setup to handle this as a "soft" error, in that it fills the table entries with -1 and saves the updated file, so that at least some kind of result is produced (perhaps for the purpose of testing). Remember that this is another situation of us trying to develop and test a script in environments that don't necessarily have access to the DB's or ephemeris that it needs - hence the treatment as a soft error, just so it completes.

jdavies-st:
What type of `exception` is this line
https://github.com/spacetelescope/jwst/blob/f552d2cf380d7beb0bad77854828cf28441569cc/scripts/set_bary_helio_times.py#L113
supposed to catch?  I think it would be good to use a specific exception here, not just a bare exception.

scripts/move_wcs.py
```diff
@@ -96,14 +96,14 @@ def add_default_keywords(new_hdr):
     wcsaxes = new_hdr['WCSAXES']
     if wcsaxes == 3:
         default_pc = {"PC1_1": 1, "PC1_2": 0, "PC1_3": 0,
-                      "PC2_1": 0 , "PC2_2": 1, "PC2_3": 0,
+                      "PC2_1": 0, "PC2_2": 1, "PC2_3": 0,
```
nden: We don't have data with 3 wcsaxes any more. This script should be revised, in another PR.

hbushouse:
Blame indicates that the whole section within which line 113 is contained was added by @stscieisenhamer a couple years ago to force the script to always update the table entries with something, even if an error occurred in doing the time conversion. I suspect this was mainly meant to trap errors in accessing the JWST and/or JPL ephemeris.

nden:
If DMS has the database available I agree with @philhodge we should raise the error instead of catching it.
We are catching everything now, including potentially errors in our code.

nden:
targcoord = astropy.coordinates.SkyCoord(
        ra=pheader['TARG_RA'],
        dec=pheader['TARG_DEC'],
        frame='fk5',
        unit=(u.hourangle, u.deg)


Lines 65 -70: Is the frame really `fk5`?

stscieisenhamer:
probably because, at the time, it was considered a non-hard error; i knew there was some reason for asking that in the first place.  So, the obvious big question: at the end of the day, what does the science want?

hbushouse:
I believe @perrygreenfield is the original author, who can hopefully answer that.

jdavies-st:
Agree with Phil and Nadia.  Exception can be raised in both `set_bary_helio_times()` and `main()`, with a `try/except` block in the latter.

perrygreenfield:
As a more general comment, I tend to agree with raising an exception unless SDP demands an exit.

As to what was wanted science-wise, I'm not sure I recall correctly. My memory seems to recall that one of the drivers was to deal with time series observations (at the time Nikole Lewis was the main driver of that). Is she still around? ext doesn't show her any more.

hbushouse:
OK, everyone have a look at the latest commit. Did I do it right? Please suggest improvements.

hbushouse:
Note that this also removes the code that set all the values to -1 when an exception occurs. So this really just aborts mid-stream.

scripts/set_bary_helio_times.py
```diff
@@ -111,25 +109,17 @@ def set_bary_helio_times(filename, jwstpos=None):
...
-            bcol = fits.Column(
-                name='bary_end_time', format='D', unit='MJD', array=nulls
-            )
-            hcol = fits.Column(
-                name='helio_end_time', format='D', unit='MJD', array=nulls
-            )
+            # Cleanup before exiting
+            hdul.flush()
+            hdul.close()
+            raise Exception('Error in calculating times.') from exception
```
jdavies-st: Better here would be something like:
```
        except Exception as exception:
            # Cleanup before exiting
            hdul.flush()
            hdul.close()
            raise exception
```
That way you are raising the exception that actually caused the failure, not a new, no-name exception.

hbushouse: True, that I'm raising another exception, but the use of "from exception" does also cause the original exception (with its error message) to be "emitted", followed by the new one. So that way you get both the original error (whatever it was, like db access error, for example), plus a statement from set_bary_helio_times itself giving you a line number in the traceback. (Hey, I learned something about python 3 exceptions today!)

jdavies-st: Ah yes, getting the line in `set_bary_helio_times` is useful in the traceback.  🥇

CHANGES.rst
```diff
@@ -25,6 +25,11 @@ master_background
  - Added unit tests for expand_to_2d.  Support CombinedSpecModel data
    for the 1-D user-supplied background spectrum. [#3188]
 
+set_bary_helio_times
+--------------------
+
+ - Added exit code of 1 when unable to compute converted times. [#3197]
```
nden: Change this to raise an exception?

PR #3200, opened 2019-03-01
Pull test data inputs_root and results_root values from pytest config.

rendinam:
This standardizes the location for Artifactory/local data root locations for test inputs and results. It will facilitate in a general manner a request for incremental environment snapshot artifacts to be published at the conclusion of selected regression test runs.

The pytest config items `inputs_root` and `results_root` are supported by `ci-watson` starting in version `0.3`, released recently.

### DIFF
--- a/jwst/tests/base_classes.py
+++ b/jwst/tests/base_classes.py
@@ -40,9 +40,12 @@ class BaseJWSTTest:
     ignore_hdus = ['ASDF']
     ignore_keywords = ['DATE', 'CAL_VER', 'CAL_VCS', 'CRDS_VER', 'CRDS_CTX', 'FILENAME']
 
-    input_repo = 'jwst-pipeline'
-    results_root = 'jwst-pipeline-results'
     env = 'dev'
+
+    @pytest.fixture(autouse=True)
+    def config_access(self, pytestconfig):
+        self.inputs_root = pytestconfig.getini('inputs_root')[0]
+        self.results_root = pytestconfig.getini('results_root')[0]
 
     @pytest.fixture(autouse=True)
     def auto_toggle_docopy(self):
@@ -54,7 +57,7 @@ class BaseJWSTTest:
 
     @property
     def repo_path(self):
-        return [self.input_repo, self.env, self.input_loc]
+        return [self.inputs_root, self.env, self.input_loc]
 
     def get_data(self, *pathargs, docopy=True):
         """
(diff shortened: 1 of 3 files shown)


pllim:
To be clear, I guess you can pin `ci-watson` in Jenkins files but given CI is passing, I guess you don't have to.

jwst/tests/base_classes.py
```diff
@@ -42,6 +42,11 @@ class BaseJWSTTest:
 
     env = 'dev'
 
+    @pytest.fixture(autouse=True)
+    def setup_class(self, pytestconfig):
+        self.inputs_root = pytestconfig.getini('inputs_root')[0]
+        self.results_root = pytestconfig.getini('results_root')[0]
```
jdavies-st: I think this method should probably not be called `setup_class()`, as that is not what is doing the work here.  You've defined this an autouse fixture, so it is not doing what traditional `setup_class()` and `teardown_class()` classmethods are doing.  Those are always decorated `@classmethod` and cannot take a fixture (`pytestconfig` in this case) as an argument.  See

https://docs.pytest.org/en/latest/xunit_setup.html

https://stackoverflow.com/questions/28288385/pytest-how-to-pass-an-argument-to-setup-class

pllim: I asked Matt to use that. Matt said you two tried several things and it didn't work, so I suggested doing it this way that I know will work, as I use it myself for other projects. As long as it works, I am not too worried about tradition here. It can be used as an instance method too.

pllim:
The diff looks wrong. I asked Matt to rebase again.

jwst/tests/base_classes.py
jdavies-st: What also works:
```
    @pytest.fixture(autouse=True)
    def matt_is_brilliant(self, pytestconfig):
        self.inputs_root = pytestconfig.getini('inputs_root')[0]
        self.results_root = pytestconfig.getini('results_root')[0]
```
The `@pytest.fixture(autouse=True)` is the thing that is making it work, not the class method name.  I think putting `setup_class()` there makes it confusing, as it makes you think its using the old way of setup and teardown (that is not compatible with `pytest` fixtures).  Better to be clear.

pllim: But when and where to call this `config_access`? `setup_class` is automatically invoked at, well, class setup time.

jdavies-st: The decorator `@pytest.fixture(autouse=True)` is what invokes it.  I.e. it gets invoked for every test function within the class.  Without that decorator, it is not a fixture, and if it is not a fixture, it cannot accept `pytestconfig` as an arg.  We have a similar `autouse` fixture in the base class:

https://github.com/spacetelescope/jwst/blob/02e4e9085f021f562dd9bf7ad56cce980dbcca83/jwst/tests/base_classes.py#L50-L56

pllim: Ah, TIL. But why not have only one `toggle_my_fixtures` method and get everything set there?

jdavies-st: Good idea!

PR #3201, opened 2019-03-01
SpecModel.wcs

### LINKED ISSUE #3207: Improve SpecModel.wcs
Issue [JP-572](https://jira.stsci.edu/browse/JP-572) was created by Nadia Dencheva:

SpecModel.wcs should take advantage of improvements in gwcs:
 * input_frame should have 1 axis
 * inverse should be assigned to the WCS and possibly used instead of OutputCombinedModel.to_pixel.

nden:
The `input_frame` of SpectralModel was changed to be 1D (this was recently enabled in gwcs).
An inverse model was added in case when the wavelength is strictly monotonic.

The tests in combine_1d are still stub and here for the purpose of discussion.
Closes #3207


jwst/combine_1d/tests/test_combine1d.py
```diff
@@ -0,0 +1,40 @@
...
+    bkg = create_master_bkg.create_background(wave, flux)
+    inp = combine1d.InputSpectrumModel(bkg, bkg.spec[0],
+                                       exptime_key='unit_weight',
+                                       background=False)
+    outm = combine1d.OutputSpectrumModel()
+    outm.assign_wavelengths([inp])
+    wave = [.3, 6.8, 2.2, .1]
+    res_to_pixel = outm.to_pixel(wave)
+    res_wcs = outm.wcs.invert(wave, wave, wave)
+    assert_allclose(res_to_pixel, res_wcs)
```
nden: The `to_pixel` test compares the results of `OutputSpectrumModel.to_pixel()` with `OutputSpectrumModel.wcs.invert()`. How should inputs outside the wavelength range be handled? Right now, `wcs.invert` will return `np.nan` for those while `to_pixel` extrapolates.

nden: I set the model in spec_wcs to allow extrapolation and now the test comparison with `to_pixel` passes.

jwst/combine_1d/tests/test_combine1d.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
+from numpy.testing import assert_allclose
+
+from ... master_background import create_master_bkg
+from ...combine_1d import combine1d
```
hbushouse: Is there a need for consistency as to whether there's a whitespace after "..." in a relative import like this?

sosey: I would have thought it was required to not have a space, i.e. `from ...master_background` is explicitly preferred. I'm curious now though for myself, did we specify somewhere if the jwst package would use all relative explicit, or absolute imports?

hbushouse: We didn't officially specify it anywhere, but it was agreed that we'd use relative explicit, as opposed to things like ``from jwst.master_background``, in order to avoid issues due to folks possibly having their own "jwst" defined in their environment (either intentionally or not). At least that's what I recall from 5 years ago.

philhodge: Here is an example from PEP 328 (I left out the comments because they're confusing):

    from . import D
    from .. import E                
    from ..F import G

I've seen other examples that are consistent with this.  That is, use a space if ".." (or whatever) just identifies a directory from which you want to import a module (D or E in the above examples).  Do not use a space between the last dot if what follows is a module name in the indicated directory (module F in .., and module F contains G).

nden: This file will be removed before the PR is merged. The goal is to replace `OutputSpectrumModel.to_pixel()` with a call to the `OutputSpectrumModel.wcs.inverse`. This test is meant to show that the two functions give the same results. 

otherwise :+1: for consistency in whatever form.

jwst/extract_1d/spec_wcs.py
```diff
@@ -32,20 +32,34 @@ def create_spectral_wcs(ra, dec, wavelength):
...
+    if all(np.diff(wavelength) > 0):
+           tab.inverse = Mapping((2,)) | Tabular1D(points=wavelength,
+                                                   lookup_table=pixel,
+                                                   bounds_error=False,
+                                                   fill_value=None)
+    elif all(np.diff(wavelength) < 0):
+           tab.inverse = Mapping((2,)) | Tabular1D(points=wavelength[::-1],
+                                                   lookup_table=pixel[::-1],
+                                                   bounds_error=False,
+                                                   fill_value=None)
```
philhodge: Should there be an `else` to raise an exception?  Or is this caught somewhere else?

nden: The exception is raised in modeling. I added a warning here explaining why inverse is not set.

nden:
The Jenkins tests are failing (probably because of the schema issues resolved elsewhere).
I think this is ready to merge but I'll leave it to someone else to merge.

Issue #3122, opened 2019-02-14
Allow passing a reference dictionary to extract_1d


philhodge:
The top-level function `do_extract1d` in extract.py should just read the reference file into a dictionary and pass that dictionary to a function that is almost the same as the current `do_extract1d`.  The function to process IFU data also needs to accept a dictionary for the reference file information instead of a file name.

PR #3202, opened 2019-03-01
Option to invoke extract_1d by calling do_extract1d

philhodge:
The `extract_1d` step can be run by calling `extract.do_extract1d` and passing a dictionary containing the reference file information, rather than a reference file name.  This will work for a single model but not for a `ModelContainer`.

Closes issue #3122.


nden:
The changes look good and to the point.

Some general comments - it looks like the code would benefit in terms of readability if `ExtractBase.__init__` defines the attributes as parameters. Then the two subclasses can just call the `__init__` with the arguments passed. This will eliminate all the code repetition in the `__init__` functions of the subclasses. This should be a separate ticket for the next sprint.

PR #3219, opened 2019-03-05
Fix .travis.yml

jdavies-st:
My PR #3195 inadvertently broke our Travis CI builds.  This fixes it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ dist: xenial
 language: python
 python:
   - 3.6.8
-  - 3.7.2
 sudo: false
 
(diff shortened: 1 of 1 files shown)


jdavies-st:
@nden ready for final review.

jdavies-st:
`numpy 1.15.4` seems to be broken with respect to being able to install our C extensions.  I don't know why.  Since that's the case, and to be consistent with what we are doing on the Jenkins side, I would suggest moving to `1.16`.  Meanwhile, it would be good to file an issue with `numpy` or `pip` (or `conda`) to find out why this failure is happening.

PR #3220, opened 2019-03-05
Changes to the combine_1d step

### LINKED ISSUE #3208: Provide a direct interface to the combine_1d step
Issue [JP-571](https://jira.stsci.edu/browse/JP-571) was created by Nadia Dencheva:

The master_background step needs to call directly a function in the combine_1d step and bypass the step interface.

### LINKED ISSUE #3210: combine_1d needs to accept ModelContainer as input
Issue [JP-569](https://jira.stsci.edu/browse/JP-569) was created by Nadia Dencheva:

The combine_1d step and the function within the step that does the combination (do_combine?) should be able to accept ModelContainer objects as input.

philhodge:
The combine_1d step can be invoked by calling `combine1d.do_correction`.  The first argument is now an open `jwst.datamodels` model, rather than the name of the input file or association.  That model can be either a `ModelContainer` or a `MultiSpecModel`.

Closes #3208.
Closes #3210.


jwst/combine_1d/combine1d.py
```diff
@@ -619,64 +618,53 @@ def check_exptime(exptime_key):
     return exptime_key
 
 
-def do_correction(asn_file, exptime_key, interpolation, background=False):
+def do_correction(input_model, exptime_key, background=False):
```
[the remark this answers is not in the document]
philhodge: Thanks; I'll change this.

jwst/combine_1d/combine1d.py
```diff
@@ -619,64 +618,53 @@ def check_exptime(exptime_key):
...
+                                ms, in_spec, exptime_key, background))
+    else:
+        for in_spec in input_model.spec:
+            input_spectra.append(InputSpectrumModel(
+                                input_model, in_spec, exptime_key, background))
 
     output_spec = OutputSpectrumModel()
     output_spec.assign_wavelengths(input_spectra)
-    output_spec.accumulate_sums(input_spectra, interpolation)
+    output_spec.accumulate_sums(input_spectra, "nearest")
```
nden: Technically the interpolation method used in `combine_1d` is `linear`, not `nearest neighbour`.

jwst/combine_1d/combine_1d_step.py
```diff
@@ -17,17 +18,17 @@ class Combine1dStep(Step):
     spec = """
     # integration_time or exposure_time.
     exptime_key = string(default="exposure_time")
-    # Interpolation between pixels.
+    # Interpolation between pixels.  This is currently not used.
     interpolation = string(default="nearest")
```
nden: The comment about interpolation type applies here as well.

jwst/combine_1d/combine1d.py
```diff
@@ -619,64 +618,53 @@ def check_exptime(exptime_key):
     return exptime_key
 
 
-def do_correction(asn_file, exptime_key, interpolation, background=False):
+def do_combine(input_model, exptime_key, background=False):
```
nden: Sorry to be picky but this function is general enough that it can probably be used on its own by users. In which case the name `do_combine` is too general. Can we make it more specific, e.g. `combine_spectra_1d`. I'm open to other descriptive names.

jwst/combine_1d/combine1d.py
philhodge: Linear interpolation is used in `to_pixel`, yes, but the `interpolation` parameter was passed to `accumulate_sums` instead, where it referred to taking the nearest pixel in the output file.  I left this parameter in `spec` for the time being to avoid problems with user's copies of combine_1d.cfg that would include this parameter, but I'll remove the parameter completely if you like.  Most references to `interpolation` in the code have been removed.

jwst/combine_1d/combine1d.py
philhodge: I changed the function name to `combine_1d_spectra`.

jwst/combine_1d/combine_1d_step.py
```diff
@@ -17,17 +18,17 @@ class Combine1dStep(Step):
     spec = """
     # integration_time or exposure_time.
     exptime_key = string(default="exposure_time")
-    # Interpolation between pixels.
-    interpolation = string(default="nearest")
+    # This is currently not used.
```
jdavies-st: The `spec` ConfigSpec block needs to have the comment on the same line as the configuration parameter.  Like this:
```
param = string(default="foo") # Comment string
```
otherwise, the comment string doesn't get pulled into the configspec.  Currently, this step's `ConfigSpec` looks like:
```
In [12]: pprint(step.load_spec_file())                                                                                                                                   
{'background': 'boolean(default=False)',
 'exptime_key': 'string(default="exposure_time")',
 'input_dir': 'string(default=None)        # Input directory',
 'interpolation': 'string(default="nearest")',
 'output_dir': 'string(default=None)        # Directory path for output files',
 'output_ext': "string(default='.fits')     # Default type of output",
 'output_file': 'output_file(default=None)   # File to save output to.',
 'output_use_index': 'boolean(default=True)       # Append index.',
 'output_use_model': 'boolean(default=False)      # When saving use '
                     '`DataModel.meta.filename`',
 'post_hooks': 'string_list(default=list())',
 'pre_hooks': 'string_list(default=list())',
 'save_results': 'boolean(default=False)      # Force save results',
 'search_output_file': 'boolean(default=True)       # Use outputfile define in '
                       'parent step',
 'skip': 'boolean(default=False)      # Skip this step',
 'suffix': 'string(default=None)        # Default suffix for output files'}
```

jwst/combine_1d/combine_1d_step.py
```diff
@@ -17,17 +18,17 @@ class Combine1dStep(Step):
...
+    interpolation = string(default="")
```
jdavies-st: Would it make sense to set the default for `interpolation` to `None`?  If so, then it should be removed from the `combine_1d.cfg` file too.

Issue #3179, opened 2019-02-22
JWST datamodels does not handle FITS BINTABLE unsigned int32 correctly


philhodge:
For the case that a column in a FITS BINTABLE is specified this way in the datamodels/schemas/:

      - name: DQ
        datatype: uint32

as in spec.schema.yaml, it appears that datamodels adds the TZEROi value twice.  (For uint32, TFORMi = 'J', and TZEROi = 2147483648.)  Here is an example, using a table written by the extract_1d step.  The header of the first extension of nrs1_bkg2_extract_1d.fits contains the following keywords:

    XTENSION= 'BINTABLE'           / binary table extension
    BITPIX  =                    8 / array data type
    NAXIS   =                    2 / number of array dimensions
    NAXIS1  =                   68 / length of dimension 1
    NAXIS2  =                 1382 / length of dimension 2
    PCOUNT  =                    0 / number of group parameters
    GCOUNT  =                    1 / number of groups
    TFIELDS =                    9 / number of table fields
    EXTNAME = 'EXTRACT1D'          / extension name
    EXTVER  =                    1 / extension value
    TTYPE1  = 'WAVELENGTH'
    TFORM1  = 'D       '
    TUNIT1  = 'um      '
    TTYPE2  = 'FLUX    '
    TFORM2  = 'D       '
    TUNIT2  = 'mJy     '
    TTYPE3  = 'ERROR   '
    TFORM3  = 'D       '
    TUNIT3  = 'mJy     '
    TTYPE4  = 'DQ      '
    TFORM4  = 'J       '
    TZERO4  =           2147483648
    TTYPE5  = 'NET     '
    TFORM5  = 'D       '
    TUNIT5  = 'DN/s    '
    TTYPE6  = 'NERROR  '
    TFORM6  = 'D       '
    TUNIT6  = 'DN/s    '
    TTYPE7  = 'BACKGROUND'
    TFORM7  = 'D       '
    TUNIT7  = 'DN/s    '
    TTYPE8  = 'BERROR  '
    TFORM8  = 'D       '
    TUNIT8  = 'DN/s    '
    TTYPE9  = 'NPIXELS '
    TFORM9  = 'D       '
    SLTNAME = 'S200A1  '           / Name of the slit
    SLITID  =                    0 / Slitlet ID
    SOURCEID=                    0 / Source ID
    SRCXPOS =                  0.0 / Source position in slit (x-axis)
[...]

PR #3225, opened 2019-03-06
Change VALUE datatype in dq_def schema to uint

bhilbert4:
This PR makes a simple change to the datatype of the VALUE in  jwst.datamodels.schemas.dq_def.schema.yaml

Currently this is set to a datatype of `int`. But I believe it should be `uint`. 

I found this while attempting to populate the `dq_def` attribute in a MaskModel instance with the official list of DQ flags in `datamodels.dqflags.pixel`. In this case, the final entry in my dq_def list is the tuple: (31, 2147483648, 'REFERENCE_PIXEL', ''). But when I then set that as the dq_def attribute and print the attribute, the flag value is suddenly negative, and tuple is printed as: (31, -2147483648, 'REFERENCE_PIXEL', ''). I think this is due to the integer datatype.

Here's what I was doing when I found the problem:
```
    from jwst.datamodels import MaskModel, dqflags
    definitions = []
    standard_defs = dqflags.pixel
    for bitname in standard_defs:
        bitvalue = standard_defs[bitname]
        if bitvalue != 0:
            bitnumber = np.uint8(np.log(bitvalue)/np.log(2))
        else:
            bitnumber = 0
        newrow = (bitnumber, bitvalue, bitname, '')
        definitions.append(newrow)

    m = MaskModel()
    m.dq_def = definitions

    print(definitions)
    print(m.dq_def)
```

### DIFF
--- a/jwst/datamodels/schemas/dq_def.schema.yaml
+++ b/jwst/datamodels/schemas/dq_def.schema.yaml
@@ -4,7 +4,7 @@ datatype:
 - name: BIT
   datatype: int32
 - name: VALUE
-  datatype: int32
+  datatype: uint32
 - name: NAME
   datatype: [ascii, 40]
 - name: DESCRIPTION


jdavies-st:
Ooh, interesting!  Is this related to or resolve #3179?  @philhodge ?

philhodge:
@jdavies-st Yes, it's entirely possible that the data type was set to int32 because uint32 wouldn't work.  But if it works with the change that @bhilbert4 made, that will be very interesting!

hbushouse:
If this doesn't work, @bhilbert4 could workaround it by not using the defined value for REFERENCE_PIXEL (which uses bit 32), but instead use a lower bit. For example, if the MASK file that's being created only has 4 or 5 different DQ conditions in it, then just use the first (lowest) 4 or 5 bits. That's the whole purpose of having the DQ_DEF table in the first place. People can create DQ extensions that only use a few bits, but then the values will automatically get remapped to the official settings when the file is loaded into a data model.

hbushouse:
Now that 0.13.1 has been tagged and branched, I'm going to merge this and watch for smoke.

PR #3234, opened 2019-03-08
improve the Travis test matrix

### LINKED ISSUE #3209: Improve the Travi stesting matrix
Issue [JP-570](https://jira.stsci.edu/browse/JP-570) was created by Nadia Dencheva:

Add to the Travis tests matrices to:

 
 * test with python 3.7
 * cron jobs to test with dev versions of dependencies

nden:
Closes #3209 .

This implements testing with

- python 3.7
- with dev dependences of astropy, asdf and gwcs

The dev dependencies fail because of a change in asdf. The change in asdf actually corrected how schemas are tested so the fix is still probably in datamodels. 
@drdavella


jdavies-st:
The failed checks are because the Jenkins CI tests are using `asdf` from yesterday (via astroconda-dev).  The Travis CI are using today's, and they pass.  So merging.

PR #3240, opened 2019-03-08
Update README for 0.13.1 delivery [skip ci]

hbushouse:
Update the DMS deliveries table in the README to include 0.13.1.


jdavies-st:
Shall we update this with `0.13.2`?

hbushouse:
> Shall we update this with `0.13.2`?

Yes, I wasn't planning to go back and include it with 0.13.1. It'll get included when we deliver 0.13.2.

PR #3241, opened 2019-03-08
Pin CRDS_CONTEXT=jwst_0500.pmap

jdavies-st:
Temporarily pin the CRDS_CONTEXT for our regression tests until #3106 is merged.  This is to avoid using the new MIRI LRS reference files before the code handling them is merged.


jdavies-st:
Again, the failures on Jenkins CI astroconda-dev matrix element are expected as it is using yesterday's `asdf`. Merging.

PR #3243, opened 2019-03-08
Publish conda env listing to Artifactory upon failure-free test run.

rendinam:
Per request from DMS, this JenkinsfileRT change will publish the output of `conda list --explicit` from within the testing environment to the `jwst-pipeline-results` Artifactory repository after each RT run without failures.

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -1,6 +1,10 @@
 // [skip ci] and [ci skip] have no effect here.
 if (utils.scm_checkout(['skip_disable':true])) return
 
+jobconfig = new JobConfig()
+jobconfig.enable_env_publication = true
+jobconfig.publish_env_on_success_only = true
+
 def test_env = [
     "HOME=./",
     "TEST_BIGDATA=https://bytesalad.stsci.edu/artifactory",
@@ -66,4 +70,4 @@ bc.test_cmds = ["printenv | sort",
 
 bc.test_configs = [data_config]
 
-utils.run([bc])
+utils.run([jobconfig, bc])


jdavies-st:
Looks good.

Once concern I have is that if anything is pip installed, `conda list --explicit` will not pick it up.  Only `conda list` will.  Any ideas how to handle this?

rendinam:
That's an interesting point. DMS is equipped to receive explicit conda environment specification files currently, which is how the official deliveries are made. This PR is to honor their request, as I understand it, to have the ability to test as much as possible ahead of each of those official deliveries using the same procedures as would be employed for an actual delivery. This seems like a case where some coordination between JWST development and DMS will be required, as #3183 would change the definition of the deliverable.

jdavies-st:
Right.  Exactly.

But we really do have a need to use released dependencies.  I.e. we need to stop using `astroconda-dev`.  Perhaps a recipe for release versions of `jwst` in `astroconda`?  We would need to make sure that it's released upstream dependencies are also available.  Currently it seems `drizzle` is the only one that seems to be behind several versions relative to PyPi.

Of course that does not solve our issue of wanting to have dependencies tracked in the `jwst` repo itself so that it is a proper Python package.

jdavies-st:
Failure in Jenkins CI is unrelated.  Merging.

jdavies-st:
In the future we can figure out how to deal with `pip` installed dependencies.  It will require coordination with DMS and with the Jenkins CI utils

https://github.com/spacetelescope/jenkins_shared_ci_utils/blob/cf8bb87be6b64700ac67cfa02af922f76786c290/vars/utils.groovy#L282-L300

PR #3244, opened 2019-03-09
Update to handle new pytest configurations

stscieisenhamer:
Resolves failing regressions test caused by not taking into account recent pytest configuration parameters.

### DIFF
--- a/jwst/tests_nightly/general/associations/test_sdp_pools.py
+++ b/jwst/tests_nightly/general/associations/test_sdp_pools.py
@@ -1,7 +1,6 @@
 """Test using SDP-generated pools
 """
 from collections import Counter
-import configparser
 from pathlib import Path
 import re
 
@@ -52,7 +51,10 @@ class AssociationBase(BaseJWSTTest):
         return self._truth_paths
 
 ASN_BASE = AssociationBase()
-POOL_PATHS = ASN_BASE.pool_paths
+try:
+    POOL_PATHS = ASN_BASE.pool_paths
+except Exception as e:
+    POOL_PATHS = e
 
 
 # #####
@@ -61,9 +63,13 @@ POOL_PATHS = ASN_BASE.pool_paths
 class TestSDPPools(AssociationBase):
     """Test createion of association from SDP-created pools"""
 
+    @pytest.mark.skipif(
+        isinstance(POOL_PATHS, Exception),
+        reason='Test pool files not available. Reason={}'.format(POOL_PATHS)
+    )
     @pytest.mark.parametrize(
         'pool_path',
-        POOL_PATHS
+        [] if isinstance(POOL_PATHS, Exception) else POOL_PATHS
     )
     def test_against_standard(self, pool_path):
         """Compare a generated association against a standard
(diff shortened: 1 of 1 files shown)


jwst/tests_nightly/general/associations/test_sdp_pools.py
```diff
@@ -33,27 +39,28 @@ class AssociationBase(BaseJWSTTest):
...
             self._truth_paths = self.data_glob(*self.ref_loc, glob='*.json')
         return self._truth_paths
 
-asn_base = AssociationBase()
-try:
-    POOL_PATHS = asn_base.pool_paths
-except Exception:
-    POOL_PATHS = ['test will be skipped']
+ASN_BASE = AssociationBase()
+POOL_PATHS = ASN_BASE.pool_paths
```
jdavies-st: This line is causing the problem.  I think you still need the original `try/except` block around this.  Though this seems like a hacky way to do it.  Basically the way you have it set up bypasses the `@pytest.mark.bigdata` skipping mechanism on tests and test classes.  Though I'm not sure there's another way to do it.  Hmm.

jwst/tests_nightly/general/associations/test_sdp_pools.py
```diff
@@ -1,10 +1,12 @@
 """Test using SDP-generated pools
 """
 from collections import Counter
+import configparser
```
jdavies-st: This import isn't needed anymore.

jwst/tests_nightly/general/associations/test_sdp_pools.py
stscieisenhamer: Oh, you are going to like this...

jwst/tests_nightly/general/associations/test_sdp_pools.py
```diff
@@ -24,6 +25,10 @@
 # file lists for the test parametrization.
 # #############################################################
 class AssociationBase(BaseJWSTTest):
+    """Set testing environment"""
+
+    inputs_root = pytest.config.getini('inputs_root')[0]
+    results_root = pytest.config.getini('results_root')[0]
```
jdavies-st: I'm just highlighting this use method as I didn't notice it before @stscieisenhamer mentioned it to me.  This is an alternative way of accessing this `ini` info without having to use a fixture.  The fixture just provides the same methods as the `pytest.config` object.

Using it this way is required if one wants access to `inputs_root` during test collection.  Pytest fixtures are only available during test execution, and so cannot provide a value of `inputs_root` during test collection.  @rendinam 

So maybe we should update documentation in `ci_watson` to show that
```
def test_foo(pytestconfig):
    inputs_root = pytestconfig.getini('inputs_root')[0]
```
works for tests that need `inputs_root` during test execution.  But say a fixture or a test class setup need this information.  Then it needs to be retrieved with
```
    inputs_root = pytest.config.getini('inputs_root')[0]
```

Thanks @stscieisenhamer for discovering this little bit of undocumented `pytest` trivia.

rendinam: This is excellent. I tried something similar early on since I thought it _should_ absolutely be possible to get at pytest's config items once pytest was imported. But I don't think I attempted it with the `getini()` method. Thanks @stscieisenhamer ! I'll see that the `ci-watson` docs get updated with an example of each type of access.

jdavies-st: If you just do `import pytest` there is no `pytest.config` object.  Instead, it is created and is available only when `pytest` is collecting and executing tests.  Fun.

But yeah, this does open up new possibilities.

Issue #3235, opened 2019-03-08
Ramp fitting needs to be able to reject multiple first frames for MIRI


stscijgbot:
Issue [JP-583](https://jira.stsci.edu/browse/JP-583) was created by Howard Bushouse:

The MIRI team has requested the ability to have the ramp_fit team reject multiple first frames in each integration, due to data anomalies associated with the detector resets. Currently ramp_fit is setup to look for and ignore only the very first frame, based on the GROUPDQ flag "DO_NOT_USE".

hbushouse:
Fixed by #3249

PR #3249, opened 2019-03-12
Ignore contiguous leading groups in which all pixels have been flagged DO_NOT_USE.

dmggh:
Handles the MIRI case where multiple first groups/frames need to be excluded due to reset effects in the data.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit_step.py
+++ b/jwst/ramp_fitting/ramp_fit_step.py
@@ -86,11 +86,10 @@ class RampFitStep (Step):
 
         if out_model is not None:
             out_model.meta.cal_step.ramp_fit = 'COMPLETE'
+            if input_model.meta.exposure.type in ('NRS_IFU', 'MIR_MRS'):
+                out_model = datamodels.IFUImageModel(out_model)
 
         if int_model is not None:
             int_model.meta.cal_step.ramp_fit = 'COMPLETE'
-
-        if input_model.meta.exposure.type in ('NRS_IFU', 'MIR_MRS'):
-            out_model = datamodels.IFUImageModel(out_model)
 
         return out_model, int_model
--- a/jwst/ramp_fitting/ramp_fit_step.py
+++ b/jwst/ramp_fitting/ramp_fit_step.py
@@ -84,7 +84,9 @@ class RampFitStep (Step):
                 gls_opt_model, 'fitoptgls', output_file=self.opt_name
             )
 
-        out_model.meta.cal_step.ramp_fit = 'COMPLETE'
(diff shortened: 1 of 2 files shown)


dmggh:
See github #3225 and JIRA JP-583.

dmggh:
1) Regarding this question - this calculation is already done correctly. I spoke to Phil about this, and I think the 3 of us should get together for a brief discussion.

2) I will update this PR to cover the case in which all groups have all pixels as DO_NOT_USE.

hbushouse:
So with the latest change to return None when all groups are flagged as bad, what happens when `ramp_fit` returns control back up to a pipeline? Does the pipeline handle the returned None gracefully?

jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -84,7 +84,9 @@ def process(self, input):
                 gls_opt_model, 'fitoptgls', output_file=self.opt_name
             )
 
-        out_model.meta.cal_step.ramp_fit = 'COMPLETE'
+        if out_model is not None:
+            out_model.meta.cal_step.ramp_fit = 'COMPLETE'
+
```
hbushouse: `out_model` is also referenced in line 94. Does that need updating too, to check for None?

dmggh: 1) " Does the pipeline handle the returned None gracefully?"
Currently the pipeline does not handle this case gracefully. In the calwebb_detector1 pipeline (the only pipeline that has a step beyond ramp_fit) , the output of the ramp_fit step is intended to be the input model to the gain_scale step, which is the final step of the pipeline.  Like other steps, gain_scale will try to open the ramp_fit output without checking for None:
"with datamodels.open(input) as input_model:". 
This does *not* fail if 'input' is None.

The gain_scale step will then attempt to get the gain_factor from the input model; if that is not successful, the step tries getting gain_factor from the gain reference file (which itself is derived from the input model) . This last try is currently fatal if 'input' is None:

ERROR RUNNING STEP 'Detector1Pipeline':
    Missing 'META.INSTRUMENT.NAME' keyword in header for determining
    instrument.
  File "/internal/1/sc2/git/jwst/jwst/stpipe/cmdline.py", line 297, in step_from_cmdline
    step.run(*positional)

For the other pipelines having ramp_fit as a step, it is the final step in the pipeline. So I think there could be 2 approaches  to a fix - either modify ramp fit so that if its output is None pipeline execution is halted, or modify the stpipe machinery to check for this situation.

2) Yes - the out_model code should be checked also.

hbushouse:
Fixes #3235

Issue #3135, opened 2019-02-17
Modifications to the master background step control flow


stscijgbot:
Issue [JP-526](https://jira.stsci.edu/browse/JP-526) was created by Nadia Dencheva:

The following modifications to the master background step control flow were identified as needed to support the IFU modes.

- [x]  Issue a Warning if the calwebb_spec2 background step was run and skip the master background subtraction step.
- [x]  For NRS modes, if the position of the GWA was not repeated (check GWA_XTIL and GWA_YTIL keywords) skip the calwebb_spec2 background step and run the master background subtraction step.
- [x] Add progress info to the logging of the step [comment](https://github.com/spacetelescope/jwst/issues/3135#issuecomment-467975641)

hbushouse:
It would be nice to try to add some kind of logging to master_background step that would give progress information as it's looping over multiple instances of science data within the input product (e.g. the multiple slits in a NIRSpec FS exposure and the many multiple slits in a NIRSpec MOS exposure). Just so the user has an idea of what's going on, instead of just being told it started and it finished.

stscijgbot:
Comment by Howard Bushouse: [~muzerol] and [~cdeoliveira]: What sort of tolerance should we set on the values of GWA_XTIL and GWA_YTIL in order to consider them the same across multiple exposures? The values that we get contain a fair number of (supposedly) significant digits, e.g. 0.31953862, 0.35233405, 0.337818235. Are they quantized into any kind of discrete states?

This is for the purpose of knowing when to have the calwebb_spec2 background step NOT do a direct image-to-image subtraction of NIRSpec exposures.

stscijgbot:
Comment by Catarina Alves de Oliveira: [~bushouse] What is happening on the instrument, is that if the GWA has not been commanded to move, the values of the sensor are not even read, so that first number is not changed. So the criterion to consider the GWA as not moved, is that those numbers remain exactly the same. Does that help?

stscijgbot:
Comment by Howard Bushouse: OK, so at least to start with we will test for an exact match in values. We can always add some tolerance later, if it's discovered to be necessary.

stscijgbot:
Comment by Jane Morrison: For clarification, if the GWA_XTIL and GWA_YTIL values are exactly the same then do the background step and skip the masterbackground. If they are different then skip the background step and do the masterbackground step. Should the control for which step should be run - done in the individual steps
or in the calwebb_spec2.py pipeline code ?

[...]

PR #3252, opened 2019-03-12
Updates to calspec2 background step to check NIRSPEC GWA Tilt values

### LINKED ISSUE #1149: calspec2 background subtraction needs to check GWA tilt
The direct image-to-image background subtraction method implemented in calspec2, which uses either dithered/nodded exposures of the science source or exposures obtained on a dedicated background target, needs to have a check added to it the verifies the exposures to be used as background have the same NIRSpec GWA tilt values as the science exposures. Exclude any background exposures that don't have the same tilt and if there aren't any useable background exposures left, then skip the background subtraction step.

jemorrison:
This is for #3135 it it covers the checking the calspec2 background step when NIRSPEC data is run.
The GWA_XTILT and GWA_YTILT have to match. 
I did put in a tolerance - of 1e-15. I might need to change that part. 

Fixes #1149.

### DIFF
--- a/jwst/background/background_step.py
+++ b/jwst/background/background_step.py
@@ -3,7 +3,7 @@
 from ..stpipe import Step
 from .. import datamodels
 from . import background_sub
-
+import numpy as np
 
 __all__ = ["BackgroundStep"]
 
@@ -58,13 +58,38 @@ class BackgroundStep(Step):
                 result = background_sub.subtract_wfss_bkg(
                                 input_model, bkg_name, wlrange_name)
                 result.meta.cal_step.back_sub = 'COMPLETE'
-
             else:
+                # check if input data is NRS_IFU
+                tolerance = 1.0e-15
+                result = input_model.copy()
+                do_sub = True
+                if input_model.meta.exposure.type in ["NRS_IFU"]:
+                    # check if GWA_XTILT & GWA_YTILT values of source
+                    # background are the same. If not skip step
+                    input_xtilt = input_model.meta.instrument.gwa_xtilt
+                    input_ytilt = input_model.meta.instrument.gwa_ytilt
+                    for bkg_file in bkg_list:
+                        bkg_model = datamodels.ImageModel(bkg_file)
+                        bkg_xtilt = bkg_model.meta.instrument.gwa_xtilt
+                        bkg_ytilt = bkg_model.meta.instrument.gwa_ytilt
+                        xdiff = np.absolute(bkg_xtilt - input_xtilt)
+                        ydiff = np.absolute(bkg_ytilt - input_ytilt)
+                        bkg_model.close()
+                        if xdiff > tolerance or ydiff > tolerance:
+                            do_sub = False
+                            break
                 # Do the background subtraction
-                result = background_sub.background_sub(input_model,
-                                                       bkg_list,
-                                                       self.sigma,
+                if do_sub:
+                    result = background_sub.background_sub(input_model,
+                                                           bkg_list,
+                                                           self.sigma,
                                                        self.maxiters)
-                result.meta.cal_step.back_sub = 'COMPLETE'
+                    result.meta.cal_step.back_sub = 'COMPLETE'
+                else:
+                    print('skip')
+                    result.meta.cal_step.back_sub = 'SKIPPED'
+                    self.log.warning('Skipping background subtraction')
+                    self.log.warning('GWA_XTILT and GWA_YTILT source values '
+                                     'are not the same as bkg values')
 
         return result
--- a/jwst/background/background_step.py
+++ b/jwst/background/background_step.py
@@ -3,7 +3,7 @@
 from ..stpipe import Step
 from .. import datamodels
 from . import background_sub
-
+import numpy as np
 __all__ = ["BackgroundStep"]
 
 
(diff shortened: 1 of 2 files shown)


hbushouse:
Fixes #1149.

jwst/background/background_step.py
```diff
@@ -58,13 +58,38 @@ def process(self, input, bkg_list):
                 result = background_sub.subtract_wfss_bkg(
                                 input_model, bkg_name, wlrange_name)
                 result.meta.cal_step.back_sub = 'COMPLETE'
-
             else:
+                # check if input data is NRS_IFU
+                tolerance = 1.0e-15
```
hbushouse: Looking at the GWA* values in many of the NIRSpec datasets I have laying around it appears that these keyword values typically have between 6-9 significant digits and when they vary from one exposure to another they vary at around the level of 1.e-3 or 1.e-4. So I would set our starting tolerance somewhere around 1.0e-7 or 1.0e-8.

jwst/background/background_step.py
```diff
@@ -58,13 +58,38 @@ def process(self, input, bkg_list):
...
+                result = input_model.copy()
```
hbushouse: This copy operation could be moved to the "else" portion of the if-block in which the step is actually called, i.e. down at line 89 or so. That way the copy is only produced if/when we really need it. With it here, the copy gets performed before we even know if we'll need it and hence can get left laying around unused (if we don't need it).

jwst/background/background_step.py
```diff
@@ -58,13 +58,38 @@ def process(self, input, bkg_list):
...
+                do_sub = True
+                if input_model.meta.exposure.type in ["NRS_IFU"]:
+                    # check if GWA_XTILT & GWA_YTILT values of source
+                    # background are the same. If not skip step
+                    input_xtilt = input_model.meta.instrument.gwa_xtilt
+                    input_ytilt = input_model.meta.instrument.gwa_ytilt
+                    for bkg_file in bkg_list:
+                        bkg_model = datamodels.ImageModel(bkg_file)
```
hbushouse: Probably should use the generic `datamodels.open()` here, because the bkg files can use different models, depending on which instrument mode they use. NIRSpec and MIRI IFU, in particular, would be in the form of IFUImageModel, not ImageModel.

jwst/background/background_step.py
```diff
@@ -58,13 +58,38 @@ def process(self, input, bkg_list):
...
-                                                       self.sigma,
+                if do_sub:
+                    result = background_sub.background_sub(input_model,
+                                                           bkg_list,
+                                                           self.sigma,
                                                        self.maxiters)
-                result.meta.cal_step.back_sub = 'COMPLETE'
+                    result.meta.cal_step.back_sub = 'COMPLETE'
+                else:
+                    print('skip')
```
hbushouse: Don't forget to remove the print statement before merging.

jwst/background/background_step.py
```diff
@@ -58,13 +58,38 @@ def process(self, input, bkg_list):
...
+                    result.meta.cal_step.back_sub = 'SKIPPED'
+                    self.log.warning('Skipping background subtraction')
+                    self.log.warning('GWA_XTILT and GWA_YTILT source values '
```
philhodge: Did you intend for these to be the actual keyword names?  The keywords lack the final "T".

jwst/background/background_step.py
```diff
@@ -58,13 +58,38 @@ def process(self, input, bkg_list):
                 result = background_sub.subtract_wfss_bkg(
                                 input_model, bkg_name, wlrange_name)
                 result.meta.cal_step.back_sub = 'COMPLETE'
-
             else:
+                # check if input data is NRS_IFU
+                tolerance = 1.0e-15
+                result = input_model.copy()
+                do_sub = True
+                if input_model.meta.exposure.type in ["NRS_IFU"]:
```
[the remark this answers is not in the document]
philhodge: Is it necessary to check for particular modes, or would it be sufficient to just check for instrument NIRSPEC?

jwst/background/background_step.py
```diff
@@ -58,13 +58,38 @@ def process(self, input, bkg_list):
...
+                    # check if GWA_XTILT & GWA_YTILT values of source
+                    # background are the same. If not skip step
+                    input_xtilt = input_model.meta.instrument.gwa_xtilt
+                    input_ytilt = input_model.meta.instrument.gwa_ytilt
+                    for bkg_file in bkg_list:
+                        bkg_model = datamodels.ImageModel(bkg_file)
+                        bkg_xtilt = bkg_model.meta.instrument.gwa_xtilt
+                        bkg_ytilt = bkg_model.meta.instrument.gwa_ytilt
+                        xdiff = np.absolute(bkg_xtilt - input_xtilt)
```
philhodge: The tilts are scalars, so you can use Python's `abs` function, and you wouldn't need to import `numpy`.

jemorrison:
made changes based on reviewers comments

jwst/background/background_step.py
```diff
@@ -58,13 +57,37 @@ def process(self, input, bkg_list):
...
+                # check if input data is NRS_IFU
+                tolerance = 1.0e-8
+                do_sub = True
+                if input_model.meta.instrument.name in ["NIRSPEC"]:
+                    # check if GWA_XTIL & GWA_YTIL values of source
+                    # background are the same. If not skip step
+                    input_xtilt = input_model.meta.instrument.gwa_xtilt
+                    input_ytilt = input_model.meta.instrument.gwa_ytilt
+                    for bkg_file in bkg_list:
+                        bkg_model = datamodels.open(bkg_file)
```
jdavies-st: it would be better to use the the `with` context manager here so that you don't have to call `close()` a few lines down.
```
with datamodels.open(bkg_file) as bkg_model:
    bkg_xtilt = bkg_model.meta.instrument.gwa_xtilt
    bkg_ytilt = bkg_model.meta.instrument.gwa_ytilt
```

jwst/background/background_step.py
```diff
@@ -58,13 +57,37 @@ def process(self, input, bkg_list):
...
+                        bkg_xtilt = bkg_model.meta.instrument.gwa_xtilt
+                        bkg_ytilt = bkg_model.meta.instrument.gwa_ytilt
+                        xdiff = abs(bkg_xtilt - input_xtilt)
+                        ydiff = abs(bkg_ytilt - input_ytilt)
+                        bkg_model.close()
+                        if xdiff > tolerance or ydiff > tolerance:
```
jdavies-st: Here you've reimplemented a slightly different version of `np.allclose`.  You might just want to use it instead.
```
if np.allclose((input_xtilt, input_ytilt), (bkg_xtilt, bkg_ytilt), atol=1e-8, rtol=0):
```
The parameters `atol` and `rtol` are important for different sized numbers, so I'm not sure what those should be.  Numbers very close to zero need special consideration.  Do we know what range these GWA tilt numbers are?

jdavies-st:
Don't forget to edit CHANGES.rst

jemorrison:
@jdavies-st 
When I change looping over the backgrounds from 'for bkg_file in bkg_list' 
with 'with datamodels.open(bkg_list)'
my 'break' if they do not match does not work any more because it is not in a loop.

I added the break because I did not want to waste time openning models that did not
need to be openned. With your suggestion of using 'with datamodels" maybe
it is much more efficient and I can just check every file and worry about breaking out.

Could you confirm ?

jemorrison:
@jdavies-st nevermind last comment. I misunderstood what you suggested I think I have it correct now

PR #3253, opened 2019-03-13
Use publicly released dependencies for regression tests

### LINKED ISSUE #3143: Implement improved automated testing

jdavies-st:
- Continue using `conda` in our Jenkins RT builds, but avoid using the astroconda-dev channel for the dependencies.  Use astroconda and default channels for released versions of dependencies.  This has the advantage of still allowing a spec file via `conda list --explicit` at the end of the run to be generated.

- Use Anaconda's `numpy nomkl` packages, which allow more efficient parallelization of the test run.  Should be similar to normal `numpy` available via `pip install` which doesn't have on-chip parallelization enabled.  This should be beneficial for DMS in their runs on HT Condor as well.

- The kicker: instead of having our dependencies in 3 places

    - setup.py
    - astroconda-dev recipe for `jwst`
    - Jenkinsfile/JenkinsfileRT
    
    We now only have them in two places.  This PR enables the Jenkinsfile to use the astroconda-dev recipe's dependencies, so it does not have to list them itself.

Closes #3183 

Closes #3256 

Resolves #3143

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,33 +1,10 @@
 if (utils.scm_checkout()) return
 
-python_ver = '3.6'
-pip_numpy_ver = '1.16'
-conda_numpy_ver = '1.15'
+python_version = '3.6'
 
-def test_env = [
+test_env = [
     "CRDS_SERVER_URL=https://jwst-crds.stsci.edu",
     "CRDS_PATH=./crds_cache",
-]
-
-// Define conda packages needed for build bc2.  Some come from astroconda-dev
-def conda_packages = [
-    "asdf",
-    "astropy",
-    "crds",
-    "drizzle",
-    "flake8",
-    "gwcs",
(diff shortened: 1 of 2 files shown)


jdavies-st:
Regression test run using this branch can be found

https://boyle.stsci.edu:8081/job/RT/job/jdavies-dev/

Job 102.  Check out the run time compared to the current 5.5 hours runtime.

jdavies-st:
Thanks Matt.  I just figured that out myself, as I saw the output spec file not having a `.txt` appended.  Would it be better not to have spaces in the `buildConfig.name` string?

Issue #3121, opened 2019-02-14
Allow using an extract_1d reference image for IFU


philhodge:
The `extract_1d` step currently has the option to use a reference image to specify source and background regions (see issue #1777), except for the case that the input is IFU data.  The step should be modified to allow using a reference image for IFU data.

PR #3258, opened 2019-03-13
Use reference image for IFU data

### LINKED ISSUE #3213: extract_1d IFU mode should be able to work with reference images
Issue [JP-574](https://jira.stsci.edu/browse/JP-574) was created by Nadia Dencheva:

The master_background step needs changes in the extract_1d step. For IFU mode the extract_1d step/function needs to work with reference image specifying regions of source and background.

philhodge:
The `extract_1d` step can now use a reference image to define the source and optional background regions for IFU data.  The format of this reference image is the same as has been supported for other modes.
Resolves #3213.
Closes #3213.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,11 @@ datamodels
   http://stsci.edu/schemas/fits-schema/ to map to the correct location
   in the ``jwst`` package. [#3239]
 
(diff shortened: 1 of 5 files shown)


hbushouse:
Also fixes #3121

PR #3263, opened 2019-03-14
Add control flow logic for master background step

### LINKED ISSUE #3135: Modifications to the master background step control flow

jemorrison:
Modified `MasterBackgroundStep` to be skipped if `BackgroundStep` was already run on the input science data.  A new `force_subtract` parameter is added to the step to override this logic.

Resolves #3135

### DIFF
--- a/jwst/master_background/master_background_step.py
+++ b/jwst/master_background/master_background_step.py
@@ -1,5 +1,5 @@
 from os.path import basename
-
+import logging
 from ..stpipe import Step
 from .. import datamodels
 
@@ -17,8 +17,8 @@ class MasterBackgroundStep(Step):
     spec = """
         user_background = string(default=None) # Path to user-supplied master background
         save_background = boolean(default=False) # Save computed master background
+        subtract_background = boolean(default=None) #control flag
     """
-
 
     def process(self, input):
         """
@@ -36,6 +36,11 @@ class MasterBackgroundStep(Step):
 
         save_background : bool, optional
             Save master background.
+
+        subtract_background : bool, optional
+            A flag which indicates whether the background should be subtracted
+            If None, the logic in the step determines if backgrouned is subtracted
+            If not None, the parameter overrides logic in step
 
         Returns
         -------
@@ -73,7 +78,25 @@ class MasterBackgroundStep(Step):
                 self.record_step_status(result, 'master_background', success=False)
 
                 return result
+            # Check if subtract_background is set to False -> skip step
+            if self.subtract_background is False:
+                self.log.warning(
+                    "Not subtracting masterbackground, subtract_background set to False")
+                result = input_data.copy()
+                self.record_step_status(result, 'master_background', success=False)
+                return result
+            # Check if subtract_background is None but the background was
+            # subtracted in calspec2 background step  -> skip step
+            if self.subtract_background is None and \
+                input_data.meta.cal_step.back_sub == 'COMPLETE':
+                self.log.warning(
+                    "Not subtracting master background, background was subtracted in calspec2")
 
+                result = input_data.copy()
+                self.record_step_status(result, 'master_background', success=False)
+                return result
+
+            # various tests have passed and now we want to subtract the master background
             # Check if user has supplied a master background spectrum.
             if self.user_background is None:
                 # TODO: 1. compute master background from asn, 2. subtract it
@@ -94,8 +117,8 @@ class MasterBackgroundStep(Step):
             if self.save_background and self.user_background is None:
                 # self.save_model(background, suffix='masterbg', asn_id=asn_id)
                 pass
-            
-            self.record_step_status(result, 'master_background')
+
+            self.record_step_status(result, 'master_background', success=True)
 
         return result
 
@@ -118,11 +141,16 @@ def subtract_2d_background(source, background):
     `~jwst.datamodels.DataModel`
         Background subtracted from source.
     """
+
+    log = logging.getLogger(__name__)
+    log.setLevel(logging.DEBUG)
+
     def _subtract_2d_background(model, background):
         result = model.copy()
         # Handle individual NIRSpec FS, NIRSpec MOS
         if isinstance(model, datamodels.MultiSlitModel):
-            for slit, slitbg in zip(result.slits, background.slits):
+            for i, (slit, slitbg) in enumerate(zip(result.slits, background.slits)):
+                log.info('Master background subtraction being done for slit %i', i)
                 slit.data -= slitbg.data
 
         # Handle MIRI LRS, MIRI MRS and NIRSpec IFU
--- a/jwst/master_background/master_background_step.py
+++ b/jwst/master_background/master_background_step.py
@@ -37,10 +37,11 @@ class MasterBackgroundStep(Step):
             Save master background.
 
         force_subtract : bool, optional
-            Optinoal user-supplied flag which subtracts the master background overriding the checks
+            Optional user-supplied flag which subtracts the master background overriding the checks
             on whether the background in calspec2 has already been subtracted.
             Default is set to false. The step logic determines if the master background should be
-            subtracted. If set to true then the step logic is by passed and the master background is subtrcted
+            subtracted. If set to true then the step logic is bypassed and the master background is
+            subtracted.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


jwst/master_background/master_background_step.py
```diff
@@ -17,9 +17,9 @@ class MasterBackgroundStep(Step):
     spec = """
         user_background = string(default=None) # Path to user-supplied master background
         save_background = boolean(default=False) # Save computed master background
+        subtract_background = boolean(default=None) #control flag
```
hbushouse: Ideally the comment here should provide a user with a hint as to what the param does, so I would change the wording from "control flag" to something like "Subtract master background".

jwst/master_background/master_background_step.py
```diff
@@ -73,7 +78,25 @@ def process(self, input):
                 self.record_step_status(result, 'master_background', success=False)
 
                 return result
+            # Check if subtract_background is set to False -> skip step
+            if self.subtract_background is False:
```
hbushouse: I think you first need to check whether `self.subtract_background` is None and if it isn't, then you can check for False. If it's None I think the current line will fail (because it doesn't exist).

jwst/master_background/master_background_step.py
```diff
@@ -73,7 +78,25 @@ def process(self, input):
...
+                self.log.warning(
+                    "Not subtracting masterbackground, subtract_background set to False")
+                result = input_data.copy()
+                self.record_step_status(result, 'master_background', success=False)
+                return result
+            # Check if subtract_background is None but the background was
+            # subtracted in calspec2 background step  -> skip step
+            if self.subtract_background is None and \
+                input_data.meta.cal_step.back_sub == 'COMPLETE':
+                self.log.warning(
```
hbushouse: I'm not sure if we want this to be a warning. The normal/default flow will likely be to have calspec2 step applied and hence this will be a fairly standard case (to have `master_background` get skipped). When users see a warning message they tend to immediately think that something's wrong.

jwst/master_background/master_background_step.py
```diff
@@ -118,11 +141,16 @@ def subtract_2d_background(source, background):
...
+    log = logging.getLogger(__name__)
+    log.setLevel(logging.DEBUG)
+
     def _subtract_2d_background(model, background):
         result = model.copy()
         # Handle individual NIRSpec FS, NIRSpec MOS
         if isinstance(model, datamodels.MultiSlitModel):
-            for slit, slitbg in zip(result.slits, background.slits):
+            for i, (slit, slitbg) in enumerate(zip(result.slits, background.slits)):
+                log.info('Master background subtraction being done for slit %i', i)
```
hbushouse: I like the info this message provides, but I'm not sure this is where we want it in the flow. It's the place where the actual bkg signal is being subtracted from the (multiple) source data, but that's a really quick process of just doing the subtraction. What takes more time is the part that isn't really written yet and will eventually flesh out the If-else block that starts up at line 101. The process of loading or computing all the individual 1-D bkg spectra, combining them into the master 1-D bkg, and then expanding that into the 2-D space of all the input source data instances is what will take some real time and hence that's where it would be nice to have the logging take place. Unfortunately, someone was clever in the way functions like `expand_to_2d` and `subtract_2d_background` were written, in that they're only called once and internally they do all the looping over the multiple source inputs. So the log messages almost need to be imbedded down at that level.

jwst/master_background/master_background_step.py
philhodge: To elaborate a little on what @hbushouse wrote, this line should be:

    if self.subtract_background is not None and not self.subtract_background:

The point is that you're not supposed to check boolean variables by `if x is True:` or `if x == True:` (or False); you're supposed to use this syntax:  `if x:` or `if not x`.  This is a PEP 8 issue.

jwst/master_background/master_background_step.py
philhodge: Since there can be a lot of slits, you might want to log info only every 20 slits or so.  This is done in `extract_1d` in extract.py around line 2923 (see `log_increment`).  It's messy, though.  You could just check whether `(i + 1) % 20 == 0`, and if so, then log a message.

jwst/master_background/master_background_step.py
```diff
@@ -17,9 +16,9 @@ class MasterBackgroundStep(Step):
     spec = """
         user_background = string(default=None) # Path to user-supplied master background
         save_background = boolean(default=False) # Save computed master background
+        subtract_background = boolean(default=None) #Subtract master background
```
jdavies-st: space after the `#`

jwst/master_background/master_background_step.py
```diff
@@ -37,6 +36,11 @@ def process(self, input):
         save_background : bool, optional
             Save master background.
 
+        subtract_background : bool, optional
+            A flag which indicates whether the background should be subtracted
+            If None, the logic in the step determines if backgrouned is subtracted
```
jdavies-st: `backgrouned` -> `background`

jwst/master_background/master_background_step.py
```diff
@@ -73,7 +77,25 @@ def process(self, input):
...
+            if self.subtract_background is not None and not self.subtract_background:
+                self.log.warning(
+                    "Not subtracting masterbackground, subtract_background set to False")
+                result = input_data.copy()
+                self.record_step_status(result, 'master_background', success=False)
+                return result
+            # Check if subtract_background is None but the background was
+            # subtracted in calspec2 background step  -> skip step
+            if self.subtract_background is None and \
+                input_data.meta.cal_step.back_sub == 'COMPLETE':
```
jdavies-st: This will fail for a `ModelContainer`, as it has no `meta.cal_step.back_sub` attribute.  It's member models do.

This suggests that a `get_step_status()` method might be useful.

jwst/master_background/master_background_step.py
jdavies-st: Also, how will this interact with `skip`, which is built-in to the step?

hbushouse: Hmmm, @jdavies-st that's a good question. The normal way a user requests skipping is via the step class "skip" param. There's no equivalent for them to force a step to execute (just the absence of step.skip). In the `extract_1d` step it made sense to have a separate `subtract_background` param, because the step is still doing other work. But in this case we're trying to turn the entire step on or off. Perhaps we need to rethink this strategy a bit.

jwst/master_background/master_background_step.py
hbushouse: A quick-n-dirty approach would be to test for `ModelContainer` and if it is, then only check the value in the first model. If it isn't `ModelContainer`, then use the existing test above.

jdavies-st: Btw, this is why the unit tests below are failing.  I like Howards quick-n-dirty approach, but as we discussed this morning, it would be good to make sure all models pass the above check, not just the first one.

jemorrison:
I added a 'force_subtract' parameter and if the data is a model container it loops through the data.
I probably have too much log information to user, but I was trying to be clear.
Check and see if this is the logic of how to control the master background

jwst/master_background/master_background_step.py
```diff
@@ -37,6 +36,12 @@ def process(self, input):
         save_background : bool, optional
             Save master background.
 
+        force_subtract : bool, optional
+            Optinoal user-supplied flag which subtracts the master background overriding the checks
```
hbushouse: typo: "Optinoal" needs to be "Optional"

jwst/master_background/master_background_step.py
```diff
@@ -37,6 +36,12 @@ def process(self, input):
...
+            on whether the background in calspec2 has already been subtracted.
+            Default is set to false. The step logic determines if the master background should be
+            subtracted. If set to true then the step logic is by passed and the master background is subtrcted
```
hbushouse: typo: "by passed" should be "bypassed"

jwst/master_background/master_background_step.py
```diff
@@ -73,7 +78,53 @@ def process(self, input):
...
+                    isub = 0
+                    for indata in input_data:
+                        if indata.meta.cal_step.back_sub == 'COMPLETE':
+                            do_sub = False
+                            isub += 1
+
+                    if not do_sub and isub == len(input_data):
+                        self.log.info(
+                            "Not subtracting master background, background was subtracted in calspec2")
+                        self.log.info("To force the master background to be subtracted from this data,"
```
philhodge: This is a really minor comment, but you could add a space after the comma.  The same comment applies to lines 110 and 118 as well.

PR #3264, opened 2019-03-15
Revert "Merge pull request #1027 from bernie-simon/jit_container"

jdavies-st:
This reverts the complexity in `ModelContainer` introduced in #1027.

 - It added complexity to the code with no speed or memory usage benefit.  See discussion in #1027.  The data arrays in opened `DataModels` are memory mapped, so this is expected.
 - It made determining the state of `ModelContainer._models` difficult.  Instead of it always being a list of `DataModels`, sometimes it was a list of strings of FITS files, but that depended on the operations done to the `ModelContainer`, i.e. if one had tried to read one of the models.  One could end up in a state were `_models` was a mixed list of strings and open `DataModels`.  Not good.
 - Removes the `persist` arg to `ModelContainer`.  When `True`, it just delayed when the model was actually opened and could lead to the half-open list state described above.  When `False`, it didn't allow iterating through items in `meta`, which was always an important use case.

Regression test build on this branch is here:

https://boyle.stsci.edu:8081/job/RT/job/jdavies-dev/103/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,9 @@ datamodels
   http://stsci.edu/schemas/fits-schema/ to map to the correct location
   in the ``jwst`` package. [#3239]
 
+- Change ``ModelContainer`` to load and instantiate datamodels from an
+  association on init.  This reverts #1027. [#3264]
+
 master_background
 -----------------
 
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -42,10 +42,6 @@ class ModelContainer(model_base.DataModel):
         - None: initializes an empty `ModelContainer` instance, to which
           DataModels can be added via the ``append()`` method.
 
-    persist : bool
-        If True, do not close model after opening it.
-
-
     Examples
     --------
     >>> container = ModelContainer('example_asn.json')
@@ -70,16 +66,17 @@ class ModelContainer(model_base.DataModel):
     # does not describe the data contents of the container.
     schema_url = "container.schema.yaml"
 
-    def __init__(self, init=None, persist=True, **kwargs):
+    def __init__(self, init=None, **kwargs):
 
         super(ModelContainer, self).__init__(init=None, **kwargs)
-        self._persist = persist
 
         if init is None:
             self._models = []
         elif isinstance(init, list):
-            self._validate_model(init)
-            self._models = init[:]
+            for item in init:
+                if not isinstance(item, model_base.DataModel):
+                    raise ValueError('list must contain only DataModels')
+            self._models = init
         elif isinstance(init, self.__class__):
             instance = copy.deepcopy(init._instance)
             self._schema = init._schema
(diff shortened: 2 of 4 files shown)


hbushouse:
The discussion in #1027 presented tests of speed, which seemed to show that the new approach wasn't any slower than the old and in some cases was faster (surprising), but I don't see any evidence of memory tests. We're starting to get some simulated examples of associations containing 50+ dithered images. Cases like that could be a real problem if all the models are open simultaneously. If memory mapping is doing what it's supposed to be doing, it shouldn't be a problem, but when we did performance testing years ago already on `calwebb_detector1` for large raw data sets that were running out of memory, it wasn't at all obvious that memory mapping was actually working (the entire raw science data set, along with open reference files, all seemed to be in memory at once). No one could ever figure out why it didn't seem to be working (perhaps something to do with the way our low-level code interfaces to astropy.io.fits or whatever it's using to do the actual file mapping).

jdavies-st:
I ran memory profiler on a small script that simply loads an association of 36 NIRCam `_cal` images.  Each is 96MB on disk.

First under current master (with the complexity):
```
rey> python -m memory_profiler mem.py 
Filename: mem.py

Line #    Mem usage    Increment   Line Contents
================================================
     5     86.4 MiB     86.4 MiB   @profile
     6                             def my_func():
     7     92.7 MiB      6.2 MiB       with datamodels.open('association-f150w-modulea.json') as c:
     8   1292.8 MiB     37.0 MiB           for im in c:
     9   1292.8 MiB      0.0 MiB               filename = im.meta.filename
    10   1292.8 MiB      0.0 MiB           for im in c:
    11   1292.8 MiB      0.0 MiB               shape = im.data.shape
    12   1292.8 MiB      0.0 MiB       del c
    13    392.4 MiB      0.0 MiB       gc.collect()
    14    392.4 MiB      0.0 MiB       print("done")
```

And now the reverted branch:
```
rey> python -m memory_profiler mem.py 
Filename: mem.py

Line #    Mem usage    Increment   Line Contents
================================================
     5     86.4 MiB     86.4 MiB   @profile
     6                             def my_func():
     7   1292.6 MiB   1206.2 MiB       with datamodels.open('association-f150w-modulea.json') as c:
     8   1292.6 MiB      0.0 MiB           for im in c:
     9   1292.6 MiB      0.0 MiB               filename = im.meta.filename
    10   1292.6 MiB      0.0 MiB           for im in c:
    11   1292.6 MiB      0.0 MiB               shape = im.data.shape
    12   1292.6 MiB      0.0 MiB       del c
    13    392.2 MiB      0.0 MiB       gc.collect()
    14    392.2 MiB      0.0 MiB       print("done")
```
As you can see, these 36 images use up ~1.3GB of memory.  The only difference is when that memory gets used.

jdavies-st:
Agree @stscieisenhamer that memory management is a broader issue and may require rethinking or refactoring parts of the level 3 pipelines.

jdavies-st:
@bushouse, your concern about whether memory mapping is working as we hope is legit, though unrelated to the changes in this PR.  I just set `memmap=False` in the codebase (2 lines), and for the same association as above, we get:
```
rey> python -m memory_profiler mem.py 
Filename: mem.py

Line #    Mem usage    Increment   Line Contents
================================================
     5     86.5 MiB     86.5 MiB   @profile
     6                             def my_func():
     7   1608.9 MiB   1522.4 MiB       with datamodels.open('association-f150w-modulea.json') as c:
     8   1608.9 MiB      0.0 MiB           for im in c:
     9   1608.9 MiB      0.0 MiB               filename = im.meta.filename
    10   1608.9 MiB      0.0 MiB           for im in c:
    11   1608.9 MiB      0.0 MiB               shape = im.data.shape
    12   1608.9 MiB      0.0 MiB       del c
    13    477.9 MiB      0.0 MiB       gc.collect()
    14    477.9 MiB      0.0 MiB       print("done")
```
So it does use more memory as expected, but not dramatically more.  Another way to put it is that `memmap=True` is not saving us a lot of memory.  Part of that may come down to data munging, i.e. the `dtype` in the datamodel may only match what is stored in the FITS file on read and write, but not when open in memory.  Possibly some other causes.  But again, this is a separate issue worth looking at in the future.

Issue #3226, opened 2019-03-06
Schema updates for new keywords


hbushouse:
A few new keywords are being added to the keyword dictionary. We need to make corresponding additions to our datamodels schema.

OBSFOLDR: https://jira.stsci.edu/browse/JWSTKD-296: Add to primary header in "Observation identifiers" section, datatype = string, title = "Name of the APT observation folder", data model name = "observation_folder".

MIRNGRPS, MIRNFRMS: https://jira.stsci.edu/browse/JWSTKD-289: Add to primary header in "Exposure parameters" section, datatype = integer, titles = "MIRI flight system number of groups" and "MIRI flight system number of frames", data model names = "miri_fsw_ngroups" and "miri_fsw_nframes".

hbushouse:
Also need to modify characteristics of DATE-END keyword to be consistent with DATE-BEG: https://jira.stsci.edu/browse/JWSTKD-303. The type is changing from date to date/time, but our schema already has it as plain "string", so that can remain the same. Change the comment (title) field to "Date-time end of data acquisition".

hbushouse:
GS Pointing table extension header keywords: https://jira.stsci.edu/browse/JWSTKD-298

Add the keywords JITTRAVG, JITTRMAX, and JITTRMIN to the Pointing table header and remove the keyword JITTERMS (wherever it may appear).

hbushouse:
https://jira.stsci.edu/browse/JWSTKD-249: Change Guide Star-related "DATA_END" keyword name to "DATAEND" (just remove the underscore).

hbushouse:
Remaining fixes included in #3745.

PR #3266, opened 2019-03-15
Keyword updates

hbushouse:
Partially addresses #3226. Turns out that the changes requested to the DATE-END keyword need more investigation and consensus from the Keyword Dictionary review board before implementing. All other changes are implemented here, including some misc rearrangement of the keyword order in our schemas to match what SDP produces, and Guide Star keywords that were missing from a couple of the table extension headers.


jdavies-st:
It seems that the round-tripping of the comment section for the new keywords that map to the FITS `BinTableHDU` HDUs doesn't work.  I.e. the comment is not preserved.

hbushouse:
The latest commit adds allowed values to the PATTTYPE keyword that are needed as part of the `master_background` subtraction step development work.

I'd like to get all of these merged and will then file a separate issue to deal with the problem of keyword comments not showing up in the headers of BinTable extensions.

PR #3268, opened 2019-03-18
Improve subarray slice error messages

### LINKED ISSUE #3160: Print raw SUBARRAY bounding box values in failure diagnostics
This suggestion is about improving the utility which checks SUBARRAY bounding boxes and extracts SUBARRAY regions from reference file data.   

Nutshell:  (a) report the raw data SUBSTRT1/2 and SUBSIZE1/2 values when there's an error and (b)  name the reference file slice coordinates XSTART0/YSTART0 in the diagnostics to emphasize that they're shifted by 1 relative to the reference file and data bounding boxes defined by keywords.

These should be trivial  changes to make,  modifying a log message and exposing an existing debug log message.

For context,  there is a function jwst.lib.reffile_utils.get_subarray_data() with this signature:

```
def get_subarray_data(sci_model, ref_model):
    ...
    return ref_model.data[ystart:ystop, xstart:xstop]
```
1.  The role of the function is to combine data and reference SUBARRAY pixel coordinates (SUBSTRT1/2 and SUBSIZE1/2) and chop out a region of the reference file data corresponding to the SUBARRAY.   These keywords are also referred to by the datamodels as XSTART/YSTART and XSIZE/YSIZE in the data models;  they are 1-indexed values.

2. The diagnostics also dub the reference file 2D slice derived from the boxes in (1) as XSTART/YSTART,  but they are 0-based Pythonic index values.   It might be useful to name them XSTART0/YSTART0 in the diagnostics to make the change in index form clear,  i.e. this XSTART is not like that XSTART.

3. When the utility function detects invalid SUBARRAY arithmetic,  the current diagnostics don't report the raw bounding box defined by the data,  only the derived 0-based reference file data slice coordinates.  I think it would be helpful to report the raw SUBARRAY bounding box coordinates of both references and data.  That has the advantage that (a) they're both 1-based indices and (b) it's possible to say "look here, *this* is wrong or *that* is wrong" rather than doing inverse arithmetic based on the derived reference file region.

hbushouse:
Updated the reference file utility routines that extract subarrays from reference files so that when errors are encountered due to bad subarray keyword values (e.g. SUBSTRTn, SUBSIZEn) in either the science or reference file, the resulting error messages are more explicit about giving all the 1-indexed SUBSTRTn and SUBSIZEn values for both the science and reference file, and list the computed array slice indexes as python 0-indexed values.

Fixes #3160

### DIFF
--- a/jwst/lib/reffile_utils.py
+++ b/jwst/lib/reffile_utils.py
@@ -79,8 +79,7 @@ def ref_matches_sci(sci_model, ref_model):
         # If the ref file is full-frame, set the missing params to
         # default values
         if ref_model.meta.instrument.name.upper() == 'MIRI':
-            if (ref_model.data.shape[-1] == 1032 and
-                ref_model.data.shape[-2] == 1024):
+            if (ref_model.data.shape[-1] == 1032 and ref_model.data.shape[-2] == 1024):
                 log.warning('Missing subarray corner/size keywords in reference file')
                 log.warning('Setting them to full-frame default values')
                 xstart_ref = 1
@@ -95,8 +94,7 @@ def ref_matches_sci(sci_model, ref_model):
                 log.error('Missing subarray corner/size keywords in reference file')
                 raise ValueError("Can't determine ref file subarray properties")
         else:
-            if (ref_model.data.shape[-1] == 2048 and
-                ref_model.data.shape[-2] == 2048):
+            if (ref_model.data.shape[-1] == 2048 and ref_model.data.shape[-2] == 2048):
                 log.warning('Missing subarray corner/size keywords in reference file')
                 log.warning('Setting them to full-frame default values')
                 xstart_ref = 1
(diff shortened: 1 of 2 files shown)


jwst/lib/reffile_utils.py
```diff
@@ -337,27 +323,26 @@ def get_subarray_model(sci_model, ref_model):
...
-                  'incompatible with size of reference data array')
-        log.error('xstart=%d, xstop=%d, ystart=%d, ystop=%d',
-                   xstart, xstop, ystart, ystop)
-        log.error('Reference xsize=%d, ysize=%d',
-                   ref_model.meta.subarray.xsize,
-                   ref_model.meta.subarray.ysize)
+        xstop > ref_model.meta.subarray.xsize or ystop > ref_model.meta.subarray.ysize):
+        log.error('Computed reference file slice indexes are incompatible with size of reference data array')
+        log.error('Science: SUBSTRT1=%d, SUBSTRT2=%d, SUBSIZE1=%d, SUBSIZE2=%d',
+                   xstart_sci+1, ystart_sci+1, xsize_sci, ysize_sci)
```
philhodge: xstart_sci and ystart_sci are already one-based.

jwst/lib/reffile_utils.py
```diff
@@ -337,27 +323,26 @@ def get_subarray_model(sci_model, ref_model):
...
+        log.error('Reference: SUBSTRT1=%d, SUBSTRT2=%d, SUBSIZE1=%d, SUBSIZE2=%d',
+                   xstart_ref+1, ystart_ref+1, xsize_ref, ysize_ref)
```
philhodge: xstart_ref and ystart_ref are already one-based.

jwst/lib/reffile_utils.py
```diff
@@ -292,15 +278,15 @@ def get_subarray_data(sci_model, ref_model):
...
     # the reference file data array
     if (xstart < 0 or ystart < 0 or
-        xstop > ref_model.meta.subarray.xsize or
-        ystop > ref_model.meta.subarray.ysize):
+        xstop > ref_model.meta.subarray.xsize or ystop > ref_model.meta.subarray.ysize):
         log.error('Computed reference file slice indexes are ' +
                   'incompatible with size of reference data array')
-        log.error('xstart=%d, xstop=%d, ystart=%d, ystop=%d',
+        log.error('Science: SUBSTRT1=%d, SUBSTRT2=%d, SUBSIZE1=%d, SUBSIZE2=%d',
+                   xstart_sci+1, ystart_sci+1, xsize_sci, ysize_sci)
```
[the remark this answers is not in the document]
hbushouse: Ah, you're right. The `xstart_sci` and `ystart_sci` values come directly from model.meta.subarray.xstart/ystart and hence are already 1-indexed. The slice index values `xstart` and `ystart` reported later are the ones that are 0-indexed. Good catch!

jwst/lib/reffile_utils.py
```diff
@@ -292,15 +278,15 @@ def get_subarray_data(sci_model, ref_model):
...
+        log.error('Reference: SUBSTRT1=%d, SUBSTRT2=%d, SUBSIZE1=%d, SUBSIZE2=%d',
+                   xstart_ref+1, ystart_ref+1, xsize_ref, ysize_ref)
+        log.error('Slice indexes: xstart=%d, xstop=%d, ystart=%d, ystop=%d',
```
jdavies-st: It might be useful for developers to see the slice in slice notation:
```
log.error('Slice: [%d:%d, %d:%d', ystart, ystop, xstart, xstop)
```

PR #3275, opened 2019-03-19
Split input in master_background into source and background

### LINKED ISSUE #3211: Split input in master_background into source and background
Issue [JP-568](https://jira.stsci.edu/browse/JP-568) was created by Nadia Dencheva:

Currently the spec3 pipeline opens an association table as a ModelContainer.

For the case of the master_background subtraction step the association table will contain source and background exposures. Consider what is the best way and place (master_background step or spec3 pipeline) those two types to be split into two ModelContainers.

jdavies-st:
In addition to implementing the basic master background creation in `MasterBackgroundStep`, it also allows a `ModelContainer` to be initialized by a list of of files, i.e.
```
from glob import glob
from jwst import datamodels

files = glob('*_x1d.fits')
container = datamodels.open(files)
```

Which is useful for testing and general use.

Resolves #3211.

Depends on #3274 being merged first.

To do:

- [x] add regression test for MIRI IFU association with dedicated background exposures.

### DIFF
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -4,6 +4,7 @@ import os.path as op
 import warnings
 
 from asdf import AsdfFile
+from astropy.io import fits
 
 from ..associations import (
     AssociationNotValidError,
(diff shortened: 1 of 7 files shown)


jdavies-st:
The latest commit allows the whole step to run on MIRI IFU data.

Input science data with background:
Screen Shot 2019-03-20 at 2 09 26 PM

master background-subtracted science data:
Screen Shot 2019-03-20 at 2 09 37 PM

Still need to clean up the unit tests and finish the regression test for the MIRI IFU association.

jemorrison:
I am trying to run the same data through master background to better understand what is going on at the edges. Did you create an association with det_image_seq1_MIRIFULONG_34exp1_cal.fits as the source and det_image_seq1_MIRIFULONG_34exp1_bkg_cal.fits as the background ?

hbushouse:
The test data are in /grp/jwst/ssb/chartreuse/master_background_data/miri-mrs/dedicated-background/ and the ASN file is "miri_mrs_mbkg_0304_spec3_asn.json". This is a shortened ASN that only contains 3 on-source exposures and 4 bkg exposures, rather than the full 20 x 20 set.

PR #3279, opened 2019-03-21
Shorten test names so result dirs don't overwrite on Artifactory

jdavies-st:
Tests in
```
jwst/tests_nightly/general/nirspec/test_spec2pipelines.py
```
were producing the same output directory names
```
test_nrs_fs_multi_spec2_test_n0
```
when run using `pytest-xdist` and so when uploaded to Artifactory, one would overwrite the other.

This PR shortens names so that each test output dir gets a unique name.  Also, one of the IFU tests was being labeled as a NIRSpec FS test.  Fixes that too.  Shortens test IDs elsewhere too to remove the word "test" redundancy in the output directories.

### DIFF
--- a/jwst/tests_nightly/general/miri/test_miri_steps.py
+++ b/jwst/tests_nightly/general/miri/test_miri_steps.py
@@ -38,7 +38,7 @@ class TestMIRISteps(BaseJWSTTestSteps):
                                      side_gain=1.0),
                       output_truth='jw00001001001_01101_00001_MIRIMAGE_bias_drift.fits',
                       output_hdus=[],
-                      id='test_refpix_miri'
+                      id='refpix_miri'
                      ),
                 # test_refpix_miri2: refpix step performed on MIRI data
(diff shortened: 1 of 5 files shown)


jwst/tests_nightly/general/miri/test_miri_steps.py
```diff
@@ -58,7 +58,7 @@ class TestMIRISteps(BaseJWSTTestSteps):
                       step_pars=dict(),
                       output_truth='jw00001001001_01101_00001_MIRIMAGE_dark_current.fits',
                       output_hdus=[],
-                      id='test_dark_current_miri'
+                      id='dark_current_miri'
```
hbushouse: Could shorten these even further, if necessary, to just "dark_miri"

jdavies-st: True!  Though the step is called `jwst.dark_current.DarkCurrentStep`.  I think the above is short enough to not run into any of these problems anymore.

PR #3281, opened 2019-03-23
Remove C code; Improve 2dhist pre-match shift computation

### LINKED ISSUE #3267: use2dhist option in tweakreg does not work correctly
NOTE: this is a copy of https://github.com/spacetelescope/drizzlepac/issues/278 that affects the algorithm in `jwst.tweakreg`.

I have identified for the very least three issues with `use2dhist` option in `tweakreg`:

~1. In the histogram, `X-` and `Y-` axes are swapped leading to wrong estimates for the initial shifts when shifts along axes are different;~

2. Because the code converts floating point coordinates to "bin coordinates" using integer conversion (vs. rounding), the corner of a bin has coordinates `(0, 0)` instead of `(-0.5, -0.5)` leading to a half pixel shift in the returned estimate. This shift is not corrected for (as far as I could see) the code that follows.

3. The code zeroes all bins that are below `nsigma*sigma` level before finding the peak in the histogram and computing center of mass. First of all, a peak is a peak and so unless peak is below `nsigma*sigma` level this will have no effect on finding the peak. However, zeroed bins do affect the results of the center-of-mass computations: bin values are always non-negative and therefore zeroing bins below a chosen noise level likely will affect the quality of the center of mass. IMO, it would be better to leave noisy bins "as is".

Poor initial "pre-fitting" can lead to poor match results.

mcara:
This PR fixes the issues described in https://github.com/spacetelescope/jwst/issues/3267 As a side-effect, this PR also gets rid of the C code in `carrutils.c` => no more C code in the pipeline => @jhunkeler : please review if `setup.py` needs any further updates/simplifications (these`setup.py` updates, if any, do not need to happen in this PR and can be made at a later time).

This PR mirrors https://github.com/spacetelescope/tweakwcs/pull/69

Fixes #3267


jdavies-st:
You'll want to remove `jwst/tweakreg/tests/test_interface.py` in this PR as well.  I put that in there to make sure your C code got installed and was at least importable.  That's the reason the tests are failing now.

CHANGES.rst
```diff
@@ -377,6 +377,8 @@ tweakreg
 
 - Bug fix: Use integer division in Python 3. [#3072]
 
+- Bug fix: Improved 2D Histogram (pre-match shift) algorithm in Python. [#3281]
```
jdavies-st: Can you move this up to the 0.13.2 section?  It should be at the top of the file instead of down here.

mcara: @jdavies-st Thanks for noticing this! Done.

Issue #3287, opened 2019-03-25
Error in tweakreg when processing MIRI images


hbushouse:
Recent DMS testing has resulted in the following error during `tweakreg` processing of a set of MIRI images, which I suspect is due to the fact that the detected sources are all bogus, because the input images contain nothing but noise.
```
2019-03-25 14:24:02,721 - stpipe.Image3Pipeline - INFO - Starting calwebb_image3 ...
2019-03-25 14:24:04,052 - stpipe.Image3Pipeline - INFO - Aligning input images...
2019-03-25 14:24:04,099 - stpipe.Image3Pipeline.tweakreg - INFO - Step tweakreg running with args (<ModelContainer>,).
2019-03-25 14:24:04,556 - stpipe.Image3Pipeline.tweakreg - INFO - Detected 5 sources in jw80600006001_02101_00001_mirimage_cal.fits.
2019-03-25 14:24:05,179 - stpipe.Image3Pipeline.tweakreg - INFO - Detected 48 sources in jw80600006001_02102_00001_mirimage_cal.fits.
2019-03-25 14:24:05,625 - stpipe.Image3Pipeline.tweakreg - INFO - Detected 13 sources in jw80600006001_02103_00001_mirimage_cal.fits.
2019-03-25 14:24:06,245 - stpipe.Image3Pipeline.tweakreg - INFO - Detected 39 sources in jw80600006001_02104_00001_mirimage_cal.fits.
2019-03-25 14:24:06,516 - stpipe.Image3Pipeline.tweakreg - WARNING - /Users/bushouse/anaconda3/envs/myjwst/lib/python3.6/site-packages/spherical_geometry/polygon.py:594: RuntimeWarning: invalid value encountered in triple_product
  orient = great_circle_arc.triple_product(A-B, C-B, B)

2019-03-25 14:24:06,517 - stpipe.Image3Pipeline.tweakreg - WARNING - /Users/bushouse/anaconda3/envs/myjwst/lib/python3.6/site-packages/spherical_geometry/polygon.py:133: RuntimeWarning: invalid value encountered in triple_product
  orient = great_circle_arc.triple_product(A-B, C-B, B)

----------------------------------------------------------------------
ERROR RUNNING STEP 'Image3Pipeline':
    Out of domain for acos
Traceback (most recent call last):
[...]

PR #3292, opened 2019-03-28
Fix a failure in tweakreg when catalogs contain sources outside WCS domain

mcara:
This PR fixes the crash described in https://github.com/spacetelescope/jwst/issues/3287 by adding code to weed out sources that lie outside of the domain of definition of data model's WCS.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -67,6 +67,9 @@ tweakreg
 - Fixed a bug in handling situations when no useable sources are
   detected in any of the input images. [#3286]
 
+- Enhanced source catalog extraction algorithm to filer out sources outside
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -67,6 +67,9 @@ tweakreg
 - Fixed a bug in handling situations when no useable sources are
   detected in any of the input images. [#3286]
 
+- Enhanced source catalog extraction algorithm to filer out sources outside
```
hbushouse: Typo: "filer" should be "filter"

mcara: 😢

mcara: Done.

hbushouse:
Fixes #3287

PR #3296, opened 2019-03-28
Add master_background step to calspec3 pipeline

### LINKED ISSUE #3288: Add master_background step to calwebb_spec3 pipeline
Issue [JP-599](https://jira.stsci.edu/browse/JP-599) was created by Howard Bushouse:

The calwebb_spec3 pipeline module needs to be updated to call the master_background step when background members are present in the input spec3 ASN.

hbushouse:
This is a first-cut, crude attempt to add the `master_background` subtraction step to the `calwebb_spec3` pipeline. For sake of simplicity I've put it before the point in the flow where `MultiSlitModel` data get rearranged from exposure-based models to source-based models. We may need to revisit this when attempting to apply to NIRSpec MOS observations.

I noticed that at least the `cube_build` step needs access to the model.meta.asn_table meta info, which apparently is not preserved when using `split_container` or in the `ModelContainer` coming back from `master_background`, so I've got a line in there to copy it manually to the models passed to subsequent steps.

Fixes #3288

### DIFF
--- a/jwst/pipeline/calwebb_spec3.py
+++ b/jwst/pipeline/calwebb_spec3.py
@@ -76,11 +76,10 @@ class Spec3Pipeline(Pipeline):
         self.output_file = output_file
 
         # Split the inputs into science and background
-        background_data = None
         science_data, background_data = split_container(input_models)
 
         # If background data are present, call the master background step
-        if background_data is not None:
+        if len(background_data) > 0:
             source_models = self.master_background(input_models)
             source_models.meta.asn_table = input_models.meta.asn_table
         else:
--- a/jwst/pipeline/calwebb_spec3.py
+++ b/jwst/pipeline/calwebb_spec3.py
@@ -81,6 +81,8 @@ class Spec3Pipeline(Pipeline):
         # If background data are present, call the master background step
         if len(background_data) > 0:
             source_models = self.master_background(input_models)
+            if self.master_background.skip:
+                source_models = science_data
             source_models.meta.asn_table = input_models.meta.asn_table
         else:
             source_models = science_data
--- a/jwst/pipeline/calwebb_spec3.py
+++ b/jwst/pipeline/calwebb_spec3.py
@@ -3,11 +3,13 @@
 from .. import datamodels
 from ..associations.lib.rules_level3_base import format_product
 from ..exp_to_source import multislit_to_container
+from ..master_background.master_background_step import split_container
 from ..stpipe import Pipeline
 
 # step imports
 from ..cube_build import cube_build_step
 from ..extract_1d import extract_1d_step
+from ..master_background import master_background_step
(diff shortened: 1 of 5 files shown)


hbushouse:
@jdavies: FYI, testing `background_data is None` upon return from `split_container` does not work as desired, because both returns are initialized to an empty `datamodels.ModelContainer()`, so it's never None. You either have to test `len(background_data) > 0` or modify `split_container` to have it return None when there aren't any background members.

jwst/pipeline/calwebb_spec3.py
```diff
@@ -72,6 +75,16 @@ def process(self, input):
         output_file = input_models.meta.asn_table.products[0].name
         self.output_file = output_file
 
+        # Split the inputs into science and background
+        science_data, background_data = split_container(input_models)
+
+        # If background data are present, call the master background step
+        if len(background_data) > 0:
+            source_models = self.master_background(input_models)
+            source_models.meta.asn_table = input_models.meta.asn_table
```
philhodge: Do you want to copy the metadata from then entire `input_models`, or just from `science_data`?

hbushouse: The problem that I was trying to solve is the fact that the `split_container` function, which is used here and within `master_background`, seemed to be returning models that did not have `asn_table` copied over, so it's not in either `science_data` or `source_models`, hence the only place left to get it from is the original `input_models`. Definitely needs further investigation. But you're right that for the purpose of the rest of the pipeline we only need or want the `asn_table` info pertaining to just the science images (not the backgrounds).

hbushouse:
Merging. Let's see what blows up in the regression tests ...

jwst/pipeline/calwebb_spec3.py
```diff
@@ -72,6 +75,18 @@ def process(self, input):
         output_file = input_models.meta.asn_table.products[0].name
         self.output_file = output_file
 
+        # Split the inputs into science and background
+        science_data, background_data = split_container(input_models)
```
jdavies-st: This `split_container` function is memory intensive for a large association.  And it's necessary unless the background step is skipped.  I would move this splitting down to `if self.master_background.skip:`.

hbushouse: Yeah, I was thinking about that too, noticing how many times the models are opened now when turning on debug mode. Another alternative approach is to just loop through the asn_table info looking for background members, and break as soon as you find one. That completely avoids all the opening of the models themselves.

jdavies-st: Turns out is is also file handle intensive.  For our full 20+20 MIRI MRS dataset, I run out of file handles.

hbushouse: OK, so I guess that's a pretty good reason to make the change, right?

PR #3309, opened 2019-04-01
Fix issues with calspec3 use of split_container

hbushouse:
Fix some issues uncovered after merging PR #3296. The updates eliminate the use of ``split_container`` in ``calwebb_spec3`` if ``master_background`` is called, so that it only gets used once and hence avoids file handle and memory issues. Instead of calling ``split_container`` it simply looks at the ASN member info to see if there are any background members present.

Also updated the ``split_container`` function to use the ``lower()`` method on all string comparisons. This problem was revealed by one of the regression tests that used upper-case "SCIENCE" in an ASN file.

### DIFF
--- a/jwst/pipeline/calwebb_spec3.py
+++ b/jwst/pipeline/calwebb_spec3.py
@@ -83,7 +83,7 @@ class Spec3Pipeline(Pipeline):
             members_by_type[member['exptype'].lower()].append(member['expname'])
 
         # If background data are present, call the master background step
-        if len(members_by_type['background']) > 0:
+        if members_by_type['background']:
             source_models = self.master_background(input_models)
             source_models.meta.asn_table = input_models.meta.asn_table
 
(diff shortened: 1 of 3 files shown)


jwst/pipeline/calwebb_spec3.py
```diff
@@ -75,17 +76,26 @@ def process(self, input):
...
-        science_data, background_data = split_container(input_models)
+        # Find all the member types in the product
+        members_by_type = defaultdict(list)
+        product = input_models.meta.asn_table.products[0].instance
+        for member in product['members']:
+            members_by_type[member['exptype'].lower()].append(member['expname'])
 
         # If background data are present, call the master background step
-        if len(background_data) > 0:
+        if len(members_by_type['background']) > 0:
```
jdavies-st:
```
if members_by_type['background']:
```
is the pythonic, "truthy" check.  If the list has no length, it evaluates to `False`.

hbushouse: Done

Issue #1143, opened 2017-08-23
Can srctype step be moved earlier in calspec2 pipeline?


hbushouse:
The NIRSpec wavecorr step ( #601 and #1091) needs to know whether a given source (in MOS and fixed-slit exposures) is point or extended, because no correction is applied to the wavelength array for extended sources. Right now the srctype step, which sets keywords to indicate whether each source is to be treated as point or extended, is located following the extract_2d and flat-field steps, so that it can work on individual slit instances that have been produced by extract_2d. This information is needed in order for the wavecorr step to run, hence srctype must be run before wavecorr. 

Is this possible? Where will the wavecorr step be located in the calspec2 flow? @nden

hbushouse:
Right now extract_2d is duplicating some of the logic in the srctype step to determine whether sources in NIRSpec MSA and fixed-slit exposures are point or extended (and hence whether or not to apply the wavelength correction), so it works for now but the design is not optimal. Reassigning this to B7.2 to tackle it properly then.

hbushouse:
This was all reworked when the `calwebb_spec2` pipeline flow was updated to handle NRS MOS master background correction (B7.6 or B7.7), so I'm closing this.

PR #3310, opened 2019-04-01
Update srctype rules

### LINKED ISSUE #3294: SRCTYPE should be EXTENDED for dedicated background exposures
Issue [JP-601](https://jira.stsci.edu/browse/JP-601) was created by Howard Bushouse:

The source_type step in the calspec2 pipeline should be setting the type for dedicated background exposures to EXTENDED, if it isn't already. This will allow the extract_1d step to make good choices about how to do extractions.

hbushouse:
Updated the srctype step to apply more logic in certain situations. The new hierarchy is:
 - If it's a background target, set SRCTYPE = 'EXTENDED' no matter what (ignore any user input)
 - If it's a nodded exposure, set SRCTYPE = 'POINT' no matter what (ignore any user input)
 - If neither of those situations apply, use the user input, if it exists
 - If all else fails, apply specified defaults

The first 2 allow for correct functioning of down-stream steps, such as `extract_1d`, so that they do the proper thing when deciding how to do their processing.

Note that all of this logic only applies to observing modes that have a single pre-defined target/source, such as MIRI LRS, MIRI MRS, NIRSpec fixed-slit, and NIRSpec IFU. NIRSpec MOS still uses the stellarity values supplied in the MSA meta data file.

Fixes #3294

### DIFF
--- a/jwst/srctype/srctype.py
+++ b/jwst/srctype/srctype.py
@@ -3,32 +3,59 @@ import logging
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
+
 def set_source_type(input_model):
     """
+    Set the source_type, based on APT input or default values.
+
+    Parameters
+    ----------
+    input_model : `~jwst.datamodels.ImageModel`, `~jwst.datamodels.IFUImageModel`, or `~jwst.datamodels.MultiSlitModel`
+        The data model to be processed.
+
+    Returns
+    -------
+    input_model :
+        The updated model that has been processed.
     """
 
     # Get the exposure type of the input model
-    try:
-        exptype = input_model.meta.exposure.type
-        log.info('Input EXP_TYPE is %s' % exptype)
-    except:
-        log.error('Failed to access EXP_TYPE value in input')
-        log.error('Step will be skipped')
+    exptype = input_model.meta.exposure.type
+    if exptype is None:
+        log.warning('EXP_TYPE value not found in input')
+        log.warning('Step will be skipped')
         return None
+    else:
+        log.info('Input EXP_TYPE is %s' % exptype)
 
     # For exposure types that use a single source, get the user-supplied
     # source type from the selection they provided in the APT
     if exptype in ['MIR_LRS-FIXEDSLIT', 'MIR_LRS-SLITLESS', 'MIR_MRS',
-                   'NIS_SOSS', 'NRS_FIXEDSLIT', 'NRS_BRIGHTOBJ', 'NRS_IFU']:
+                   'NRC_TSGRISM', 'NIS_SOSS', 'NRS_FIXEDSLIT',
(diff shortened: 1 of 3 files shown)


jwst/srctype/srctype.py
```diff
@@ -3,32 +3,59 @@
...
+            src_type = 'EXTENDED'
+            log.info('Exposure is a background target; setting SRCTYPE = %s' % src_type)
+
+        elif (patttype is not None) and ('NOD' in patttype):
+
+            # Set all nodded exposures to POINT source type
+            src_type = 'POINT'
+            log.info('Exposure is nodded; setting SRCTYPE = %s' % src_type)
+
+        elif (user_type is not None) and (user_type in ['POINT', 'EXTENDED']):
```
jdavies-st:
```
elif user_type in ['POINT', 'EXTENDED']:
```
is simpler

jwst/srctype/srctype.py
```diff
@@ -3,32 +3,59 @@
...
+    Set the source_type, based on APT input or default values.
+
+    Parameters
+    ----------
+    input_model : `~jwst.datamodels.ImageModel`, `~jwst.datamodels.IFUImageModel`, or `~jwst.datamodels.MultiSlitModel`
+        The data model to be processed.
+
+    Returns
+    -------
+    input_model :
```
nden: This should have a return type.

hbushouse:
@nden The one complication with having this run before `extract_2d` is the fact that for NIRSpec MOS exposures it needs to set SRCTYPE for each slit instance. Do the slit instances exist before `extract_2d`?

jwst/srctype/srctype.py
```diff
@@ -3,32 +3,59 @@
...
+        log.warning('Step will be skipped')
         return None
+    else:
+        log.info('Input EXP_TYPE is %s' % exptype)
 
     # For exposure types that use a single source, get the user-supplied
     # source type from the selection they provided in the APT
     if exptype in ['MIR_LRS-FIXEDSLIT', 'MIR_LRS-SLITLESS', 'MIR_MRS',
-                   'NIS_SOSS', 'NRS_FIXEDSLIT', 'NRS_BRIGHTOBJ', 'NRS_IFU']:
+                   'NRC_TSGRISM', 'NIS_SOSS', 'NRS_FIXEDSLIT',
```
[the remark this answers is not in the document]
hbushouse: Alphabetical by full instrument name, in which case NRC (NIRCam) comes before NIS (NIRISS), and NRS (NIRSpec) is last. Doesn't work for the 3-char abbreviations.

nden:
@hbushouse The slit instances exist as part of the WCS which means the WCS object will have to be modified (if we decide to do this).

hbushouse:
> `extract_2d` determines the source type in the same way. If this step ran before extract_2d, then extract_2d could just look up `source_type`.

See #1143

PR #3311, opened 2019-04-02
Handle nodded data in master background step

### LINKED ISSUE #3283: Update master_background to use BACKGROUND spectra when necessary
Issue [JP-598](https://jira.stsci.edu/browse/JP-598) was created by Howard Bushouse:

The extract_1d step will extract and store spectra for both the source and background regions in nodded IFU and MIRI LRS exposures. The master_background step needs to have logic added that will identify input x1d products as nodded exposures and in that case have the combine_1d step combine the spectra contained in the x1d "BACKGROUND" table column, rather than the usual "FLUX" column.

jdavies-st:
The logic in this step is starting to look like spaghetti code.  I tried to make it a bit more modular.  Any suggestions on cleaning it up would be welcome.

Resolves #3283.

### DIFF
--- a/jwst/master_background/master_background_step.py
+++ b/jwst/master_background/master_background_step.py
@@ -102,6 +102,17 @@ class MasterBackgroundStep(Step):
                     input_data, background_data = split_container(input_data)
                     asn_id = input_data.meta.asn_table.asn_id
 
+                    # Check if the background members are nodded x1d extractions
+                    try:
+                        if 'NOD' in input_data[0].meta.dither.primary_type:
+                            for model in background_data:
+                                model = copy_background_to_flux(model)
+                    except TypeError:
+                        self.log.warning(
+                            "Cannot determine dither type. "
+                            "Assuming it is not nodded."
+                            )
+
                     master_background = combine_1d_spectra(
                         background_data,
                         exptime_key='exposure_time',
@@ -182,6 +193,17 @@ class MasterBackgroundStep(Step):
                         "run again and set force_subtract = True.")
 
         return do_sub
+
+
+def copy_background_to_flux(spectrum):
+    """Copy the background column to the flux column in a MultiSpecModel"""
+    result = spectrum.copy()
+    for spec in result.spec:
+        spec.spec_table['FLUX'] = spec.spec_table['BACKGROUND']
+        # Zero out the background column for safety
+        spec.spec_table['BACKGROUND'][:] = 0.0
+
+    return result
 
 
 def split_container(container):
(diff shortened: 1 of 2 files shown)


jwst/master_background/master_background_step.py
```diff
@@ -46,129 +47,164 @@ def process(self, input):
...
-
-            # Handle individual NIRSpec FS, NIRSpec MOS
-            if isinstance(input_data, datamodels.MultiSlitModel):
-                pass
 
-            # Handle associations, or input ModelContainers
-            elif isinstance(input_data, datamodels.ModelContainer):
-                input_data, background_data = split_container(input_data)
-                asn_id = input_data.meta.asn_table.asn_id
+            # Make the input data availabled to self
```
hbushouse: typo: lose the "d" at the end of "available**d**"

jwst/master_background/master_background_step.py
```diff
@@ -46,129 +47,164 @@ def process(self, input):
...
-            # Use user-supplied master background and subtract it
-            else:
-                background_2d = expand_to_2d(input_data, self.user_background)
-                result = subtract_2d_background(input_data, background_2d)
 
-                # Record name of user-supplied master background spectrum
-                if isinstance(result, datamodels.ModelContainer):
-                    for model in result:
-                        model.meta.background.master_background_file = basename(self.user_background)
+                # Skip step for case where no container and now user background
```
hbushouse: typo: "now" should be "no"

jwst/master_background/master_background_step.py
```diff
@@ -46,129 +47,164 @@ def process(self, input):
...
+        Encapsulates logic that checks if background step has already been run
+        on the data, or if the user has selected to force_subtract regardless.
+
+        Returns
+        -------
+        do_sub : bool
+            If ``True``, do the subtraction
+        """
+        do_sub = True
+        input_data = self.input_data
```
philhodge: You could move this line to be within the `if not self.force_subtract:` test.

jwst/master_background/master_background_step.py
```diff
@@ -46,129 +47,164 @@ def process(self, input):
...
+                        "run again and set force_subtract = True.")
+
+        return do_sub
+
+
+def copy_background_to_flux(spectrum):
+    """Copy the background column to the flux column in a MultiSpecModel"""
+    result = spectrum.copy()
+    for spec in result.spec:
+        spec.spec_table['FLUX'] = spec.spec_table['BACKGROUND']
```
philhodge: It looks to me as if you need a `.copy()` here, to avoid clobbering flux when zeroing out background.

jdavies-st: Thanks!  Fixed.

PR #3314, opened 2019-04-02
Guide star keyword updates

### LINKED ISSUE #3312: Updates needed for guide star keywords
https://jira.stsci.edu/browse/JWSTKD-307 requests 2 changes to guide star keywords that need to be implemented in our data model schemas:

 1) Remove the GS_STATE keyword
 2) Change the datatype of the GUIDESTA keyword from float to string

hbushouse:
Removed keyword GS_STATE and changed data type of GUIDESTA from number to string in both guide-star data model schemas.

Fixes #3312


jdavies-st:
Will this cause and reg test failures?

hbushouse:
> Will this cause and reg test failures?

Probably at least a couple. These keywords only appear in guidestar data sets, not science products. Right now we only have 1 or 2 reg tests of guidestar data processing.

PR #3316, opened 2019-04-02
Update set_telescope_pointing for null SIAF cdelt values

### LINKED ISSUE #3282: Need to set default CDELTn values in set_telescope_pointing
Mike Swam reports:
```
During JWST SDP testing of the SCSB v0.13.1 release, we found we needed to adjust the 
set_telescope_pointing.py code one more time (you had already given me the conversion
lines to put cdelt1/2 in the proper units).   The code was still failing in at least one case,
if siaf.cdelt1/2 were not defined (see Jesse’s note below, I assume for aperture cases
where the values come back unfilled from the SIAF?).   I adjusted the code a little (shown
below), and it worked, but you should decide whether you’d prefer a different handling if
siaf.cdelt1/2 come back undefined:
 
Original (with your suggested units adjustment):
            model.meta.wcsinfo.cdelt1 = siaf.cdelt1 / 3600.
            model.meta.wcsinfo.cdelt2 = siaf.cdelt2 / 3600.
 
My change:
        model.meta.wcsinfo.cdelt1 = 1.0
        model.meta.wcsinfo.cdelt2 = 1.0
        if siaf.cdelt1:
            model.meta.wcsinfo.cdelt1 = siaf.cdelt1 / 3600.
        if siaf.cdelt2:
            model.meta.wcsinfo.cdelt2 = siaf.cdelt2 / 3600.
```
We should apply these modifications to `set_telescope_pointing`.

hbushouse:
Updated the set_telescope_pointing script to set default values for CDELTn and CRPIXn keywords, in case there aren't any entries in the SIAF that can be used for populating them.

Fixes #3282


jwst/lib/set_telescope_pointing.py
```diff
@@ -1422,10 +1422,18 @@ def populate_model_from_siaf(model, siaf):
...
+        model.meta.wcsinfo.cdelt1 = 1.0  # initialize to default
+        model.meta.wcsinfo.cdelt2 = 1.0  # initialize to default
+        if siaf.crpix1:
+            model.meta.wcsinfo.crpix1 = siaf.crpix1
+        if siaf.crpix2:
+            model.meta.wcsinfo.crpix2 = siaf.crpix2
+        if siaf.cdelt1:
+            model.meta.wcsinfo.cdelt1 = siaf.cdelt1 / 3600  # in deg
+        if siaf.cdelt2:
+            model.meta.wcsinfo.cdelt2 = siaf.cdelt2 / 3600  # in deg
```
nden: The SIAF tuple has default values of `None` for all members, used in case the siaf database does not have values. This allows to writ ehe above lines as
```
model.meta.wcsinfo.crpix1 = siaf.crpix1 or 0
model.meta.wcsinfo.crpix2 = siaf.crpix2 or 0
model.meta.wcsinfo.cdelt1 = siaf.cdelt1 / 3600  or 1# in deg
model.meta.wcsinfo.cdelt2 = siaf.cdelt2 / 3600 or 1 # in deg
```

nden:
I added a test locally and it fails because `allow_default=False` in `update_wcs` (I assume this is the way it's run in SDP). The code is written to raise a `ValueError` in case of missing SIAF values unless `allow_default` is set to `True`. Should the code be changed to not raise an error on missing SIAF values?

nden:
@hbushouse  Do you want me to add the test to this PR so we can see the results?

stscieisenhamer:
To give context: The `allow_default` arose because of conversations long ago with DMS that, in an ideal system, exposures should always have engineering pointing information. If not, that would be a hard failure. Of course, that currently is not the case, so the current state is to run with `allow_default = True`, to at least get data through the system.

This was specifically for checking on quaternion availability that would produce a V1 pointing. Whether the 'hard failure' relates to SIAF-related completeness is not clear. If the requirement of at least having a good V1 pointing is sufficient, it may be fine to have a sub-par/missing SIAF transformation, where the default SIAF is basically V1 + default scale/mirror scale.

hbushouse:
> @hbushouse Do you want me to add the test to this PR so we can see the results?

Yes, please.

hbushouse:
Shouldn't the if-block starting at https://github.com/spacetelescope/jwst/blob/219e318d56780079d7fdb2a8c2d9ba45849bd405/jwst/lib/set_telescope_pointing.py#L378 have already trapped this problem, with one or more of the SIAF values coming back with a value of None?

The nested if-block that depends on the value of ``allow_default`` may not be getting executed properly, because in the main `set_telescope_pointing` script there appears to be an inconsistency in the exact spelling of that argument. Note that here https://github.com/spacetelescope/jwst/blob/master/scripts/set_telescope_pointing.py#L67 it's referenced as ``allow-default``, while here https://github.com/spacetelescope/jwst/blob/master/scripts/set_telescope_pointing.py#L102 it's ``allow_defaut`` (hyphen in the first case, underscore in the second). The ``dry_run`` arg shows the same inconsistency. Is this a real problem or not?

stscieisenhamer:
Ah yes! And that should handle the logging needs also.

jwst/lib/set_telescope_pointing.py
hbushouse: The above syntax involving crpix1 and crpix2 works fine when they're None, but it doesn't work for the cdelts, because it first tries to evaluate ``siaf.cdelt1 / 3600`` and throws the same ``TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'``

jdavies-st:
Looks like the code in `set_telescope_pointing` is still raising a `ValueError`

https://travis-ci.org/spacetelescope/jwst/jobs/515693053#L1023

It should instead handle it.

hbushouse:
@nden is working on updates to rework our handling of defaults and null SIAF entries that should fix this. Stay tuned for subsequent commits.

hbushouse:
@stscieisenhamer Interestingly enough, tests that I just ran indicate that the parsing of ``--allow-default`` and ``--dry-run`` do work properly, despite the apparent mismatch in spelling (hyphen vs. underscore). Not sure how that works. Perhaps the python ``parser.parse_args()`` somehow does a conversion? The version specified on the command-line, i.e. ``--allow-default`` is the form that must be used, other wise you get an error. Meanwhile that value somehow shows up in ``args.allow_default`` (the form with the underscore). Weird.

So it appears that no updates are needed to the script to fix this apparent mismatch in argument spelling.

stscieisenhamer:
Mismatch meaning between the switch name and the actual variable name? Yes, `parse_args` [does a conversion](https://docs.python.org/dev/library/argparse.html#dest).  For code clarity, explicitly specifying the `dest` argument can be done.

If anything, the inconsistencies of the command-line arguments themselves should be fixed introduced by short-attention-span-theater development.

hbushouse:
"Any internal - characters will be converted to _ characters to make sure the string is a valid attribute name."  Bingo.

hbushouse:
So given that the script already has 2 other arguments that employ underscores in their command-line names - "--engdb_url" and "--transpose_j2fgs" - I would suggest that we file another issue to change the hyphens to underscores in "--allow-default" and "--dry-run", for the sake of consistency across all arguments and for eliminating confusion for future code maintenance.

This change will of course require suitable notification to SDP in case they're currently using either of those arguments.

jdavies-st:
Perhaps we should convert `--engdb_url` and `--transpose_j2fgs` to `--engdb-url` and `--transpose-j2fgs`.  Generally, args to command-line interfaces in Unix don't have underscores.  Don't know if that is a rule, but certainly in practice seems to be the case.

nden:
I agree with @jdavies-st . I don't think it's a rule but it's commonly accepted (probably because a mixture of hyphens and underscores is very annoying).

PR #3319, opened 2019-04-04
Update barshadow step to apply correction

### LINKED ISSUE #1963: Apply NIRSpec barshadow correction to science data
As documented in https://jira.stsci.edu/browse/JP-292 the barshadow step needs to be updated to apply the corrections to the science data, in addition to attaching the correction arrays.

hbushouse:
Updated the ``barshadow`` step to apply the correction array to all of the science data arrays. Also just did a bunch of little stylistic cleanup to make the code more readable (IMHO).

Fixes #1963

### DIFF
--- a/jwst/barshadow/bar_shadow.py
+++ b/jwst/barshadow/bar_shadow.py
@@ -11,71 +11,76 @@ log.setLevel(logging.DEBUG)
 
 SLITRATIO = 1.15     # Ratio of slit spacing to slit height
 
+
 def do_correction(input_model, barshadow_model):
     """Do the Bar Shadow Correction
 
     Parameters
     ----------
-    input_model: MultiSlitModel datamodel object
+    input_model : `~jwst.datamodels.MultiSlitModel`
         science data model to be corrected
 
-    barshadow_model: BarshadowModel datamodel object
-        bar shadow datamodel from reference file
+    barshadow_model : `~jwst.datamodels.BarshadowModel`
+        bar shadow data model from reference file
 
     Returns
     -------
-    output_model: MultiSlitModel datamodel object
-        Science datamodel with bar shadow extensions added
-
+    output_model : `~jwst.datamodels.MultiSlitModel`
+        Science data model with correction applied and barshadow extensions added
     """
-#
-# Input is a MultiSlitModel science data model 
-# A MultislitModel has a member .slits that behaves like
-# a list of Slits, each of which has several data arrays and
-# keyword metadata items associated 
-#
-# At this point we're going to have to assume that a Slit is composed
-# of a set of slit[].nshutters in a vertical line
-#
-# Reference file information is a 1x1 ref file and a 1x3 ref file
-# Both the 1x1 and 1x3 ref files are 1001 pixel high and go from
-# -1 to 1 in their Y value WCS.  
-# 
-# 
-    exp_type = input_model.meta.exposure.type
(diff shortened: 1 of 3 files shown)


hbushouse:
Also updated to not check the stellarity value to determine the source type, but use the setting that comes from the `source_type` step upstream.

hbushouse:
Question for reviewers: we want to rescale all of the uncertainty arrays by the same amount as SCI (.data), but is it appropriate to rescale ERR, VAR_POISSON, and VAR_READNOISE all the same way, since ERR is expressed as standard deviation, while VAR_xxx are variances?

jwst/barshadow/bar_shadow.py
```diff
@@ -11,94 +11,109 @@
...
+
     # Create the pieces that are put together to make the barshadow model
     shutter_elements = create_shutter_elements(barshadow_model)
     w0 = barshadow_model.crval1
     wave_increment = barshadow_model.cdelt1
     y_increment = barshadow_model.cdelt2
     shutter_height = 1.0 / y_increment
-    # For each slitlet
+
+    # Loop over all the slits in the input model
```
jdavies-st: It looks like `input_model` gets modified in place in this `do_correction()` function.  And the caller from the step also doesn't make a copy beforehand.  Which means the data is getting modified in place?

It might make sense to have this function make a copy first then do the correction.

hbushouse: Good point. Back when it was just attaching the correction array, having it modify the input in-place was not so bad, but now it should make a new copy, just like all other steps do,

jdavies-st:
> is it appropriate to rescale ERR, VAR_POISSON, and VAR_READNOISE all the same way, since ERR is expressed as standard deviation, while VAR_xxx are variances?

Should the variances be scaled by the correction factor squared?

philhodge:
The ERR array should be scaled the same way as SCI, but I agree with @jdavies-st that the variance should be scaled with the square of the correction factor.  I'm not sure whether VAR_READNOISE should be modified, however; I'll think about it.

philhodge:
I don't see anything in the computation of `var_rnoise` that depends on the illumination, so I would say that `var_rnoise` should not be modified by the barshadow correction.

stscirij:
I also agree that the variance should be scaled with the square of the conversion factor.

jwst/barshadow/bar_shadow.py
```diff
@@ -11,94 +11,109 @@
...
+
                 # Interpolate the bar shadow correction for non-Nan pixels
                 correction = interpolate(yrow, wcol, shadow)
-                # Add the correction array and variance to the datamodel
+
+                # Add the correction array to the datamodel
                 slitlet.barshadow = correction
+
+                # Apply the correction to the science and uncertainty arrays
+                slitlet.data /= correction
```
philhodge: What is it that sets the shape of `correction` to be the same as `slitlet.data`?  Is it the `bounding_box`?  If so, is it a reasonable assumption that they are the same size in each dimension?

hbushouse: The size of the correction array is set in the ``interpolate`` function, based on the size of the input array.

jwst/barshadow/bar_shadow.py
hbushouse: Latest commit now creates a copy.

PR #3322, opened 2019-04-05
Add BUNIT to gain and readnoise schemas

### LINKED ISSUE #2943: Add BUNIT keyword to GAIN and READNOISE reffile schemas
GAIN and READNOISE reference data are expected to be in specific units, namely e/DN and DN, respectively. Hence we want reffile creators to include the BUNIT keyword in GAIN and READNOISE reffiles, so that the units are documented. In order to allow file creators to set BUNIT, we need to add it to the data model schemas for those reffiles.

hbushouse:
Added the BUNIT keyword to the datamodel schemas for the GAIN and READNOISE reference files, via inclusion of the bunit.schema.yaml file in those reffile schemas. This makes the bunit_data meta attribute available in the reffile datamodels, for use by reffile creators.

When checking the GAIN and READNOISE docs to make sure they listed the BUNIT keyword (which they did) I noticed some other errors in the way those docs were being constructed, which is also fixed here. Some of this involved splitting the single doc file for readnoise into individual components, for consistency with all other reffile docs.

Fixes #2943


jwst/datamodels/schemas/gain.schema.yaml
```diff
@@ -2,6 +2,7 @@ allOf:
 - $ref: referencefile.schema.yaml
 - $ref: subarray.schema.yaml
 - $ref: keyword_gainfact.schema.yaml
+- $ref: bunit.schema.yaml
```
nden: I can't remember whether the order of the reference `$ref` has effect on the order of keywords in the FITS header.
And if so if anyone cares if bunit is at the end of a header.

hbushouse: BUNIT appears in the SCI extension header, while all the others appear in the primary. Hence ordering is not a concern.

Issue #2941, opened 2018-12-13
Some regression test outputs overwrite their inputs


jdavies-st:
The test
https://github.com/spacetelescope/jwst/blob/ca84f0a73341126e7df5b23cd8568711032da437/jwst/tests_nightly/general/niriss/test_niriss_steps_single.py#L151-L157
overwrites its input.  `result.meta.filename` has the input filename in it.

I suspect calling the pipeline with `save_results=True` is the solution here. 

There are several other regression tests that do this as well.

jdavies-st:
```
$ grep -r result.meta.filename jwst/tests_nightly
jwst/tests_nightly/general/niriss/test_niriss_steps_single.py:        output_file = result.meta.filename
jwst/tests_nightly/general/niriss/test_niriss_steps_single.py:        output_file = result.meta.filename
jwst/tests_nightly/general/miri/test_miri_steps_single.py:        output_file = result.meta.filename
jwst/tests_nightly/general/miri/test_miri_steps_single.py:        output_file = result.meta.filename
jwst/tests_nightly/general/miri/test_miri_steps_single.py:        output_file = result.meta.filename
jwst/tests_nightly/general/miri/test_miri_steps_single.py:        output_file = result.meta.filename
jwst/tests_nightly/general/miri/test_miri_steps_single.py:        output_file = result.meta.filename
jwst/tests_nightly/general/nircam/test_nircam_steps_single.py:        output_file = result.meta.filename
jwst/tests_nightly/general/nirspec/test_nrs_ifu_wcs.py:    im = ImageModel(result.meta.filename)
jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py:        output_file = result.meta.filename
jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py:        output_file = result.meta.filename
jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py:        im = ImageModel(result.meta.filename)
```
 - [ ] jwst/tests_nightly/general/niriss/test_niriss_steps_single.py
 - [ ] jwst/tests_nightly/general/miri/test_miri_steps_single.py
 - [ ] jwst/tests_nightly/general/nircam/test_nircam_steps_single.py
 - [ ] jwst/tests_nightly/general/nirspec/test_nrs_ifu_wcs.py
 - [ ] jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py

jdavies-st:
[...]

PR #3323, opened 2019-04-05
Apply pathloss correction to NIRSpec and NIRISS SOSS data

### LINKED ISSUE #1152: Pathloss step needs to compute and store 2-D corrections

jdavies-st:
- Applies the computed pathloss correction to the `data, err, var_poisson` members of the datamodel
- Applies the `pathloss_pointsource2D` array if `SRCTYPE == "POINT"`, `pathloss_uniformsource2D` if `SRCTYPE == "EXTENDED"` or `None`.
- No longer attaches the various `pathloss_xxx` extensions.  Only a single `pathloss` extension representing the 2D correction applied.

### Questions:

1) ~~For NIRSpec IFU, is the distinction between point source and uniform source meaningful?  I.e. for a point source in the field, some slits will contain it and some won't.  Is this taken care of in the pathloss reffile?~~ This is taken care of in the reffile.

2) ~~It seems like the NIRISS SOSS (non-TSO) version of this step only computes a point source correction.  Should we be checking to make sure SRCTYPE == "POINT" before even computing the correction, and then apply it?~~ Instead of keying off SRCTYPE, this keys off aperture name.  It only gets applied for `SUBSTRIP96` aperture.  More info [here](https://jwst-docs.stsci.edu/display/JTI/NIRISS+Single+Object+Slitless+Spectroscopy#NIRISSSingleObjectSlitlessSpectroscopy-SOSSsubarrays)

Resolves #1152

### DIFF
--- a/docs/jwst/pathloss/description.rst
+++ b/docs/jwst/pathloss/description.rst
@@ -36,7 +36,7 @@ It creates 2 pairs of 1-d arrays, a wavelength array (calculated from the WCS ap
 the index of the plane in the wavelength direction) and a pathloss array
 calculated by interpolating each plane of the pathloss cube at the position of
 the source (which is taken from datamodel).  There are pairs of these arrays for
-both pointsource and uniformsource data types.
+both point source and uniform source data types.
 
 For the uniform source pathloss calculation, there is no dependence on position
 in the aperture, so the array of pathlosses and calculated wavelengths are attached
--- a/jwst/datamodels/schemas/pathlosscorr.schema.yaml
+++ b/jwst/datamodels/schemas/pathlosscorr.schema.yaml
@@ -1,7 +1,7 @@
 type: object
 properties:
   pathloss:
-    title: 2D array for pathloss
+    title: Path loss correction
     fits_hdu: PATHLOSS
     ndim: 2
     datatype: float32
--- a/jwst/datamodels/schemas/pathlosscorr.schema.yaml
+++ b/jwst/datamodels/schemas/pathlosscorr.schema.yaml
@@ -1,34 +1,8 @@
 type: object
 properties:
-  pathloss_pointsource2d:
-    title: 2-d array for pathloss (point source)
-    fits_hdu: PATHLOSS_POINTSOURCE2D
+  pathloss:
+    title: 2D array for pathloss
+    fits_hdu: PATHLOSS
     ndim: 2
     datatype: float32
-  pathloss_pointsource:
-    title: pathloss array for point sources
-    fits_hdu: PATHLOSS_POINTSOURCE
-    ndim: 1
-    datatype: float32
-  wavelength_pointsource:
-    title: wavelength array for point sources
-    fits_hdu: WAVELENGTH_POINTSOURCE
-    ndim: 1
-    datatype: float32
-  pathloss_uniformsource2d:
-    title: 2-d array for pathloss (uniform source)
-    fits_hdu: PATHLOSS_UNIFORMSOURCE2D
-    ndim: 2
-    datatype: float32
-  pathloss_uniformsource:
-    title: pathloss_array for uniform sources
-    fits_hdu: PATHLOSS_UNIFORMSOURCE
-    ndim: 1
-    datatype: float32
-  wavelength_uniformsource:
-    title: wavelength array for uniform sources
-    fits_hdu: WAVELENGTH_UNIFORMSOURCE
-    ndim: 1
-    datatype: float32
-
 $schema: http://stsci.edu/schemas/fits-schema/fits-schema
--- a/jwst/pathloss/pathloss.py
+++ b/jwst/pathloss/pathloss.py
@@ -419,7 +419,6 @@ def do_correction(input_model, pathloss_model):
                                       dx * pathloss_array[refrow_index, ix + 1]
 
         pathloss_2d = np.broadcast_to(correction, input_model.data.shape)
-        print(pathloss_2d.mean())
         output_model.data /= pathloss_2d
         output_model.err /= pathloss_2d
         output_model.var_poisson /= pathloss_2d**2
--- a/jwst/pathloss/pathloss.py
+++ b/jwst/pathloss/pathloss.py
@@ -1,6 +1,5 @@
-#
-#  Module for calculating pathloss correction for science data sets
-#
+# Module for calculating pathloss correction for science data sets
+
 import math
 import numpy as np
 import logging
@@ -11,13 +10,11 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 
-# There are 30 slices in the NIRSPEC IFU, numbered from
-# 0 to 29
+# There are 30 slices in the NIRSPEC IFU, numbered from 0 to 29
 NIRSPEC_IFU_SLICES = np.arange(30)
 
 def get_center(exp_type, input):
-    """
-    Get the center of the target in the aperture.
+    """Get the center of the target in the aperture.
     (0.0, 0.0) is the aperture center.  Coordinates go
     from -0.5 to 0.5.
     """
@@ -76,36 +73,28 @@ def calculate_pathloss_vector(pathloss_refdata,
 
     Parameters:
     -----------
+    pathloss_refdata : numpy ndarray
+        The input pathloss data array
 
-    pathloss_refdata:     numpy ndarray
+    pathloss_wcs : wcs attribute from model
 
-    The input pathloss data array
+    xcenter : float
+        The x-center of the target (-0.5 to 0.5)
 
-    pathloss_wcs:      wcs attribute from model
-
-    xcenter: Float
-
-    The x-center of the target (-0.5 to 0.5)
-
-    ycenter: Float
-
-    The y-center of the target (-0.5 to 0.5)
+    ycenter : float
+        The y-center of the target (-0.5 to 0.5)
 
     Returns:
     --------
+    wavelength : numpy ndarray
+        The 1-d wavelength array
 
-    wavelength:   numpy ndarray
+    pathloss : numpy ndarray
+        The corresponding 1-d pathloss array
 
-    The 1-d wavelength array
-
-    pathloss:   numpy ndarray
-
-    The corresponding 1-d pathloss array
-
-    is_inside_slitlet:    Boolean
-
-    Returns True if the object source position is inside the slitlet,
-    otherwise returns False
+    is_inside_slitlet : bool
+        Returns True if the object source position is inside the slitlet,
+        otherwise returns False
 
     """
     is_inside_slitlet = True
@@ -174,15 +163,15 @@ def do_correction(input_model, pathloss_model):
 
     Parameters
     ----------
-    input_model: data model object
+    input_model : data model object
         science data to be corrected
 
-    pathloss_model: pathloss model object
+    pathloss_model : pathloss model object
         pathloss correction data
 
     Returns
     -------
-    output_model: data model object
+    output_model : data model object
         Science data with pathloss extensions added
 
     """
@@ -369,9 +358,10 @@ def do_correction(input_model, pathloss_model):
 
     elif exp_type == 'NIS_SOSS':
         """NIRISS SOSS pathloss correction is basically a correction for the
-        flux that falls outside the subarray aperture.  The correction depends
+        flux from the 2nd and 3rd order dispersion that falls outside the
+        subarray aperture.  The correction depends
         on the pupil wheel position and column number (or wavelength).  The
-        simpler option is to do the correction by column number, then the only
+        simple option is to do the correction by column number, then the only
         interpolation needed is a 1-d interpolation into the pupil wheel position
         dimension."""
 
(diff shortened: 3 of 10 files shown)


jdavies-st:
It might also be worth removing the `pathloss_xxx` members from the datamodel schemas, if we're decided that we definitely don't want to attach these corrections anymore.  That should be fine with existing data, as if they exist, they will just get shuffled off to `extra_fits`.

hbushouse:
> It might also be worth removing the `pathloss_xxx` members from the datamodel schemas, if we're decided that we definitely don't want to attach these corrections anymore. That should be fine with existing data, as if they exist, they will just get shuffled off to `extra_fits`.

See the last comments in https://jira.stsci.edu/browse/JP-240

docs/jwst/pathloss/description.rst
```diff
@@ -36,7 +36,7 @@ It creates 2 pairs of 1-d arrays, a wavelength array (calculated from the WCS ap
 the index of the plane in the wavelength direction) and a pathloss array
 calculated by interpolating each plane of the pathloss cube at the position of
 the source (which is taken from datamodel).  There are pairs of these arrays for
-both pointsource and uniformsource data types.
+both point source and uniform source data types.
```
hbushouse: I think the whole doc needs a review and updates, now that nothing is done with the 1-D corrections. For example, up in line 9 it says this is the correction "to apply to spectra when the 1-d extraction is performed." That's no longer true. It's being applied here and now, at the 2-D stage. And we also need to remove or reword the couple of mentions regarding MIRI corrections not yet being implemented: it's more appropriate to say that correction algorithms for MIRI MRS and LRS modes have not yet been developed (not implemented could be construed as indicating the algorithms exist, but we just haven't gotten around to putting them in yet.)

jwst/pathloss/pathloss.py
```diff
@@ -213,26 +212,31 @@ def do_correction(input_model, pathloss_model):
...
-                        # Create the 2-d pathloss arrays
+
                         wavelength_array = slit.wavelength
-                        pathloss_pointsource_2d = interpolate_onto_grid(wavelength_array,
-                                                                        wavelength_pointsource,
-                                                                        pathloss_pointsource_vector)
-                        pathloss_uniformsource_2d = interpolate_onto_grid(wavelength_array,
-                                                                          wavelength_uniformsource,
-                                                                          pathloss_uniform_vector)
-                        slit.pathloss_pointsource2d = pathloss_pointsource_2d
```
hbushouse: Should still save whichever version (point or uniform) of the correction is applied. So this should be moved down into the if-blocks starting at line 224.

jwst/pathloss/pathloss.py
```diff
@@ -270,37 +274,42 @@ def do_correction(input_model, pathloss_model):
...
-                    # Create the 2-d pathloss arrays
+
                     wavelength_array = slit.wavelength
-                    pathloss_pointsource_2d = interpolate_onto_grid(wavelength_array,
-                                                                    wavelength_pointsource,
-                                                                    pathloss_pointsource_vector)
-                    pathloss_uniformsource_2d = interpolate_onto_grid(wavelength_array,
-                                                                      wavelength_uniformsource,
-                                                                      pathloss_uniform_vector)
-                    slit.pathloss_pointsource2d = pathloss_pointsource_2d
```
hbushouse: Same comment as above (save the version that gets applied to the slit)

jwst/pathloss/pathloss.py
```diff
@@ -335,18 +340,29 @@ def do_correction(input_model, pathloss_model):
             ymax = int(y.max())
             ra, dec, wavelength = slice_wcs(x, y)
             wavelength_array[ymin:ymax+1, xmin:xmax+1] = wavelength
-        pathloss_pointsource_2d = interpolate_onto_grid(wavelength_array,
-                                                        wavelength_pointsource,
-                                                        pathloss_pointsource_vector)
-        pathloss_uniformsource_2d = interpolate_onto_grid(wavelength_array,
-                                                          wavelength_uniformsource,
-                                                          pathloss_uniform_vector)
-        input_model.pathloss_pointsource2d = pathloss_pointsource_2d
```
hbushouse: Save whichever version gets applied.

jwst/pathloss/pathloss.py
```diff
@@ -335,18 +340,29 @@ def do_correction(input_model, pathloss_model):
...
+                wavelength_uniformsource,
+                pathloss_uniform_vector)
+            output_model.data /= pathloss_uniformsource_2d
+            output_model.err /= pathloss_uniformsource_2d
+            output_model.var_poisson /= pathloss_uniformsource_2d**2
+
         # This might be useful to other steps
-        input_model.wavelength = wavelength_array
-        input_model.meta.cal_step.pathloss = 'COMPLETE'
+        output_model.wavelength = wavelength_array
```
hbushouse: Looks like this wavelength array is saved to the output only for NIRSpec IFU mode, correct? I don't think we create a wavelength array for MIRI MRS exposures, so that's inconsistent for the 2 IFU's. @jemorrison is this wavelength array that's in the NIRSpec IFU 2-D slice images ever used anywhere downstream?

jwst/datamodels/schemas/pathlosscorr.schema.yaml
```diff
@@ -5,30 +5,9 @@ properties:
     fits_hdu: PATHLOSS_POINTSOURCE2D
```
hbushouse: I would suggest removing the "2D" suffix, because we no longer need to identify it as such, and perhaps just shorten the whole name (for readability and typability) to just "PATHLOSS_POINT".

jwst/pathloss/pathloss.py
jdavies-st: Yeah, I don't think we need to save this wavelength array.  Unless it gets recomputed down the line and would be useful to have.  @jemorrison?  @nden?

jwst/datamodels/schemas/pathlosscorr.schema.yaml
```diff
@@ -5,30 +5,9 @@ properties:
...
-    ndim: 1
-    datatype: float32
-  wavelength_pointsource:
-    title: wavelength array for point sources
-    fits_hdu: WAVELENGTH_POINTSOURCE
-    ndim: 1
-    datatype: float32
   pathloss_uniformsource2d:
     title: 2-d array for pathloss (uniform source)
     fits_hdu: PATHLOSS_UNIFORMSOURCE2D
```
hbushouse: As above, I suggest changing this to just "PATHLOSS_UNIFORM"

jwst/pathloss/pathloss.py
```diff
@@ -213,26 +212,31 @@ def do_correction(input_model, pathloss_model):
...
+                            slit.err /= pathloss_pointsource_2d
+                            slit.var_poisson /= pathloss_pointsource_2d**2
+                        else:
+                            pathloss_uniformsource_2d = interpolate_onto_grid(
+                                wavelength_array,
+                                wavelength_uniformsource,
+                                pathloss_uniform_vector)
+                            slit.data /= pathloss_uniformsource_2d
+                            slit.err /= pathloss_uniformsource_2d
+                            slit.var_poisson /= pathloss_uniformsource_2d**2
```
hbushouse: Need to save either pathloss_pointsource_2d or pathloss_uniformsource_2d - whichever one was applied - to the slit instance, so it appears in the output product.

jwst/pathloss/pathloss.py
```diff
@@ -270,37 +274,42 @@ def do_correction(input_model, pathloss_model):
...
+                        slit.err /= pathloss_pointsource_2d
+                        slit.var_poisson /= pathloss_pointsource_2d**2
+                    else:
+                        pathloss_uniformsource_2d = interpolate_onto_grid(
+                            wavelength_array,
+                            wavelength_uniformsource,
+                            pathloss_uniform_vector)
+                        slit.data /= pathloss_uniformsource_2d
+                        slit.err /= pathloss_uniformsource_2d
+                        slit.var_poisson /= pathloss_uniformsource_2d**2
```
hbushouse: Need to save pathloss_pointsource_2d or pathloss_uniformsource_2d to the slit instance.

jwst/pathloss/pathloss.py
```diff
@@ -335,18 +340,29 @@ def do_correction(input_model, pathloss_model):
...
+            output_model.var_poisson /= pathloss_pointsource_2d**2
+        else:
+            pathloss_uniformsource_2d = interpolate_onto_grid(
+                wavelength_array,
+                wavelength_uniformsource,
+                pathloss_uniform_vector)
+            output_model.data /= pathloss_uniformsource_2d
+            output_model.err /= pathloss_uniformsource_2d
+            output_model.var_poisson /= pathloss_uniformsource_2d**2
+
```
hbushouse: Save to the output model whichever correction array gets applied.

jwst/datamodels/schemas/pathlosscorr.schema.yaml
jdavies-st: I would suggest we remove the `_point` and `_uniform` from the attached name.  These 2D corrections are of the same type so they should have the same extension name.  The pipeline processing and SRC_TYPE keyword tells which was done.  This means that anything that needs to use these downstream can just read the `pathloss` extension instead of having to figure out whether it should read the `pathloss_point` or `pathloss_uniform` extension.

hbushouse: That's a good point (the potential use downstream needing to know which name to use). I was thinking of keeping points/uniform in the names just as history or logging, but you've convinced me that keeping them generic is probably best.

jdavies-st:
I've fixed up everything from the comments.  I still need input on how to apply the correction for the NIRISS SOSS non-TSO data.  As far as I can tell, the correction is a 1D array that is not the same shape as the input data model.  Do we have any example data that could point the way forward?

Or I can figure that out in a separate PR.

jwst/datamodels/schemas/pathlosscorr.schema.yaml
```diff
@@ -1,34 +1,8 @@
 type: object
 properties:
-  pathloss_pointsource2d:
-    title: 2-d array for pathloss (point source)
-    fits_hdu: PATHLOSS_POINTSOURCE2D
+  pathloss:
+    title: 2D array for pathloss
```
hbushouse: Minor nit, but for consistency with title of barshadow extension, which is "Bar shadow correction", why don't we make this "Path loss correction".

jdavies-st:
Added regression test for `PathLossStep` using NIRISS SOSS data and updated the NIRSPEC IFU regression test to not go looking for `pathloss_foo` extensions.  Regression test truth data will still need to be updated after this is merged.

Also fixed the base class `BaseJWSTTestSteps` so it doesn't overwrite its inputs, which addresses some of #2941.  That bug affected the new regression test.

hbushouse:
I know this is already closed, but I just noticed that when the pathloss extension was added back to the ``ImageModel`` schema (for NIRISS SOSS data) it was given a default value of 1.0. This may cause that extension to show up in all ``ImageModel``s that get saved to disk, even for imaging modes that don't need or use it. That should be checked. Removing the line ``default: 1.0`` may prevent it from showing up when it isn't needed.

jdavies-st:
The `pathlosscorr.schema` is what got added to `image.schema`, not the `pathloss.schema`.  The later is for the pathloss reference file.  The `.pathloss` attribute attached when the step is successfully run has no default value.

hbushouse:
Sorry, I should've been more specific in my comment. Line 53 of ``image.schema.yaml`` https://github.com/spacetelescope/jwst/pull/3323/files#diff-e3b3f74a5303d3dea3d6067ee01ca53dR53 added a reference to include ``pathloss.schema.yaml``, which defines a 2-d array with an attribute name of ``pathloss`` and fits_hdu of ``PATHLOSS``.

However, line 34 of ``image.schema.yaml`` https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/schemas/image.schema.yaml#L34 already contained the definition of 2-d data array with attribute name ``pathloss_pointsource`` and fits_hdu of ``PATHLOSS_POINTSOURCE``. So now we've got 2 pathloss correction arrays defined in ``ImageModel``. It's this second one, ``pathloss_pointsource``, that has default=1.0 set for it, which I believe will cause it to appear in all output files regardless of whether or not we actually populate it in the code.

So at the very least we need to clean up the double definition of these 2 pathloss-related arrays.

jdavies-st:
It doesn't get created because there's no auto-access in the datamodel of `self.pathloss_pointsource` as there is for `self.data`, etc.  But it should be removed.  Will do so.  Thanks for the 🦅 👁 .

PR #3349, opened 2019-04-11
Fix asdf_schema_tester for dev version of asdf

jdavies-st:
Currently, our builds against the dev version of `asdf` are failing because the schema tester has become an `entry_points` plugin.  This fixes the `jwst` package so it can get the plugin the old or new way.


jdavies-st:
More info:  https://github.com/spacetelescope/asdf/pull/676

PR #3359, opened 2019-04-11
Update photom to apply calibration

hbushouse:
Updated the ``photom`` step to apply flux calibration to the data. The IFU cases were already applying the calibration, so only slit and slitless like modes needed updating. All of those load 1-D vectors of wavelength and response, which has to be interpolated to the 2-D wavelength space of the science data being worked on. Borrowed some techniques from ``master_background`` ``expand_to_2d`` to do this. Then apply the calibration to the data and err arrays.

Also removed the relsens and relsens2d arrays from the datamodels, because we're no longer attaching the calibration to the outputs.

This is a first attempt that will likely need some further work, especially once reviewers have a look. It's been tested on MIRI LRS fixed-slit and slitless, and NIRSpec fixed-slit data (which is operationally the same as NIRSpec MOS data), and appears to work as desired.

Fixes https://jira.stsci.edu/browse/JP-255


jwst/photom/photom.py
```diff
@@ -618,21 +613,19 @@ def photom_io(self, tabdata):
...
-        log.info('PHOTMJSR value: %g', conv_factor)
+        # Store the conversion factor in the meta data
+        log.info('PHOTMJSR value: %g', conversion)
         if isinstance(self.input, datamodels.MultiSlitModel):
             self.input.slits[self.slitnum].meta.photometry.conversion_megajanskys = \
-                conv_factor
+                conversion
             self.input.slits[self.slitnum].meta.photometry.conversion_microjanskys = \
-                23.50443 * conv_factor
+                23.50443 * conversion
```
ddavis-stsci: Would it be better to have numbers such as 23.50443 defined as a constant in the file?

jwst/photom/photom.py
```diff
@@ -643,25 +636,47 @@ def photom_io(self, tabdata):
...
+            # Combine the scalar and 2-D conversions
+            # NOTE: the 2-D conversion is divided into the data for now, until the
+            # instrument teams deliver multiplicative conversions in photom ref files
+            conversion = conversion / conv_2d
+
+        # Apply the conversion to the data and all uncertainty arrays
+        if isinstance(self.input, datamodels.MultiSlitModel):
+            self.input.slits[self.slitnum].data *= conversion
+            self.input.slits[self.slitnum].err *= conversion
+            self.input.slits[self.slitnum].var_poisson *= conversion
```
philhodge: Shouldn't the variances be multiplied by `conversion**2`?

jwst/photom/photom.py
```diff
@@ -643,25 +636,47 @@ def photom_io(self, tabdata):
...
+                wl_array = expand_to_2d.get_wavelengths(self.input.slits[self.slitnum])
             else:
-                otab = np.array(list(zip(waves, relresps)),
-                                dtype=self.input.relsens.dtype)
-                self.input.relsens = otab
-                self.input.relsens.columns['wavelength'].unit= wl_unit
+                wl_array = expand_to_2d.get_wavelengths(self.input)
+
+            wl_array[np.isnan(wl_array)] = -1.
+            conv_2d = np.interp(wl_array, waves, relresps, left=0., right=0.)
```
philhodge: Since you're going to divide by `conv_2d`, wouldn't it be better for `left` and `right` be assigned a value of 1?  Or do you want to catch and handle divide by zero?

jwst/photom/photom.py
stscicrawford: Or to use astropy units for the conversion?

jwst/datamodels/schemas/cube.schema.yaml
```diff
@@ -32,8 +32,6 @@ allOf:
       default: 0.0
       ndim: 2
       datatype: float32
-    relsens:
-      $ref: relsens.schema.yaml
```
philhodge: Should you delete the relsens.schema.yaml file?

jwst/photom/photom.py
hbushouse: I wasn't sure about that. This is essentially a units change that's being applied, not a correction of some sort.

jwst/datamodels/schemas/cube.schema.yaml
hbushouse: Done it latest commit.

jwst/photom/photom.py
hbushouse: You're right. Karl confirms it should be multiplied by the square of the conversion.

jwst/photom/photom.py
hbushouse: Astropy units used in latest commit.

jwst/photom/photom.py
hbushouse: Good suggestion. I hadn't even thought about that when I just grabbed similar lines of code from the ``master_background expand_to_2d`` routine. I've updated the limits to 1.

jwst/photom/photom.py
```diff
@@ -34,52 +36,53 @@ def __init__(self, model):
...
-        self.input = model
+        # Create a copy of the input model
+        self.input = model.copy()
 
         self.instrument = model.meta.instrument.name.upper()
         self.detector = model.meta.instrument.detector.upper()
         self.exptype = model.meta.exposure.type.upper()
         self.filter = None
-        if model.meta.instrument.filter is not None:
+        if model.meta.instrument.filter:
```
nden: Why this change? This implies `filter` can be `False` .

jwst/photom/photom.py
```diff
@@ -34,52 +36,53 @@ def __init__(self, model):
...
-        will also check to see if there is a wavelength-dependent response
-        table; if there is it will be written to the output model.
+        file, apply it to the data, and write the scalar conversion
+        factor to the output model. All NIRSpec modes use wavelength-dependent
+        flux calibration factors.
 
         Parameters
         ----------
-        ftab: fits HDUList
+        ftab: FITS HDUList
```
nden: `~astrpy.io.fits.HDUList`

nden: and needs space before the column

jwst/photom/photom.py
```diff
@@ -34,52 +36,53 @@ def __init__(self, model):
...
             HDUList for NIRSPEC photom reference file
 
         area_fname: string
```
nden: space before column and typs is `str`

jwst/photom/photom.py
```diff
@@ -618,50 +624,69 @@ def photom_io(self, tabdata):
...
+        conversion = tabdata['photmjsr']
 
-        # Store the conversion factors in the meta data
-        log.info('PHOTMJSR value: %g', conv_factor)
+        # Store the conversion factor in the meta data
+        log.info('PHOTMJSR value: %g', conversion)
         if isinstance(self.input, datamodels.MultiSlitModel):
             self.input.slits[self.slitnum].meta.photometry.conversion_megajanskys = \
-                conv_factor
+                conversion
```
nden: Does this work? Do slits have `meta` attribute?

jwst/photom/photom.py
```diff
@@ -618,50 +624,69 @@ def photom_io(self, tabdata):
...
+            conversion = conversion / conv_2d
+
+        # Apply the conversion to the data and all uncertainty arrays
+        if isinstance(self.input, datamodels.MultiSlitModel):
+            self.input.slits[self.slitnum].data *= conversion
+            self.input.slits[self.slitnum].err *= conversion
+            self.input.slits[self.slitnum].var_poisson *= conversion**2
+            self.input.slits[self.slitnum].var_rnoise *= conversion**2
+            self.input.slits[self.slitnum].meta.bunit_data = 'MJy/sr'
+            self.input.slits[self.slitnum].meta.bunit_err = 'MJy/sr'
```
nden: again, may be I forget but is the `meta` slit specific?

jwst/photom/photom.py
hbushouse: It's intended to just check for whether ``filter`` is populated with anything in the model, because some instrument modes don't use the filter keyword. I was recently told this was a more "pythonic" way to check for its existence, but I could easily be wrong.

nden: If it's not populated then it's `None`. The way to check for `None` is to use `is None`

jwst/photom/photom.py
hbushouse: Yes, the ``photometry.schema`` that's inherited by ``slitdata.schema`` has a meta object.

jwst/photom/photom.py
hbushouse: Yes, ``bunit.schema`` has a meta object.

jwst/photom/photom.py
hbushouse: OK, I've put back all the instances of ``is not None`` that were there originally.

jwst/photom/photom.py
hbushouse: In the process of updating these I noticed that these actually aren't an HDUList at all, but rather a jwst DataModel. So I've declared them all that way (and cleaned up the spacing of colons).

jwst/photom/photom.py
hbushouse: Got it.

jwst/photom/photom.py
```diff
@@ -454,21 +449,18 @@ def calc_nircam(self, ftab):
...
+        exist, they will be included in the calibration.
         For WFSS (grism) mode, the calibration information extracted from the
-        reference file is attached to each slit instance in the science data.
+        reference file is applied to each slit instance in the science data.
 
         Parameters
         ----------
-        ftab: fits HDUList
-            HDUList for NIRCAM reference file
+        ftab : ~jwst.datamodels.DataModel
```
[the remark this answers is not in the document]
hbushouse: Updated all instances to reference the specific datamodel used.

jwst/photom/photom.py
```diff
@@ -30,11 +32,12 @@ def __init__(self, model):
 
         Parameters
         ----------
-        model: data model object
+        model : ~jwst.datamodels.DataModel
```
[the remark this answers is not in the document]
hbushouse: Thanks. Updated everywhere.

hbushouse:
We have a clean set of regression tests, which can now be broken by all the differences this code will produce. Merging.

PR #3360, opened 2019-04-11
fix spatial transform in lrs

nden:
This PR expresses the LRS spatial transform from detector to V2V3 without using Tabular models. 
Previously the spatial transform was constructed as a Tabular model where essentially one row and one column of detector `x` and `y` coordinates passing through the zero point were used to compute the V2, V3 values and these were replicated in a Tabular2D model. The Tabular2D model was not initialized with  `points`, just the lookup table and so evaluating it returned NaNs because the default value for `Tabular2D.points` was used.

The change in this PR does not precompute that row in V2, V3. Instead it uses `identity(1)` to pass the `x` values and `Const1D(zero_point_y)` for the `y` values. These are passed to the distortion model.
This solves the issue in reg tests where the footprint is not saved any more (it was computed but there were NaNs in it and so it was ingored).

In addition the two versions of the obsolete now `lrs_wavelength` schema are removed.

@drlaw1558 : fyi


hbushouse:
Don't forget to include a changelog entry.

nden:
Tests passed. @drlaw1558 I am going to merge this for the sake of running the reg tests. Still feel free to comment if you see something wrong.

drlaw1558:
Looking through it now

drlaw1558:
Looks ok to me, and is passing a few quick tests that I tried.  It looks like the conda-dev context isn't picking up the new LRS distortion files in crds though?  I can get around this by manually setting the correct context, but at what point will this become default?

nden:
Yes, for now you need to set the context manually to 501. The goal was to clear the regtest failures before updating the default context. The context will be updated probably on Monday.
 Thanks for checking it!

PR #3366, opened 2019-04-15
Apply corrections to readnoise variance

### LINKED ISSUE #3365: Update pathloss and barshadow steps to apply corrections to VAR_RNOISE
Issue [JP-646](https://jira.stsci.edu/browse/JP-646) was created by Howard Bushouse:

The pathloss and barshadow steps were recently updated to apply their corrections to the SCI, ERR, and VAR_POISSON arrays. It should also be applied to VAR_RNOISE.

hbushouse:
Updated the ``pathloss`` and ``barshadow`` steps to apply their corrections to the readnoise variance arrays, same as Poisson variance.

Fixes #3365.

### DIFF
--- a/jwst/barshadow/bar_shadow.py
+++ b/jwst/barshadow/bar_shadow.py
@@ -99,9 +99,8 @@ def do_correction(input_model, barshadow_model):
                 slitlet.barshadow = correction
 
                 # Apply the correction by dividing into the science and uncertainty arrays:
-                #     var_poission is divided by correction**2, because it's variance,
-                #         while err is standard deviation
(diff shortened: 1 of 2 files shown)


jwst/barshadow/bar_shadow.py
```diff
@@ -105,6 +105,7 @@ def do_correction(input_model, barshadow_model):
                 slitlet.data /= correction
```
philhodge: The previous few lines are comments, and the one that refers to var_rnoise should be updated.

hbushouse: Good catch. I've updated it.

PR #3367, opened 2019-04-15
Update set_telescope_pointing to convert FSMCORR units

### LINKED ISSUE #3300: FSMCORR: Add conversion from arcseconds to radians to set_telescope_pointing
Issue [JP-604](https://jira.stsci.edu/browse/JP-604) was created by Jonathan Eisenhamer:

Add the new conversion, introduced in the engineering telemetry, to convert the FSMCORR mnemonics, SA_ZADUCMDX and SA_ZADUCMDY, from arcseconds to radians in the transformation equations used by the `set_telescope_pointing.py` script. The conversion is

FSMCORR * Degree-to-radian / 3600.

The suggested function to modify is [`jwst.lib.set_telescope_pointing.calc_sifov_fsm_delta_matrix`|https://github.com/spacetelescope/jwst/blob/4fbc26378b00626075b010bed72e39c62cbfe74f/jwst/lib/set_telescope_pointing.py#L855].

hbushouse:
Updated ``set_telescope_pointing`` to convert the input FSM correction values from units of arcsec to radians. Added the switch ``fsmcorr_units`` to allow for the possibility of processing older data that still had these values in radians to begin with.

Fixes #3300.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -103,6 +103,8 @@ set_telescope_pointing
 - Updates to prevent crashes when SIAF values needed for crpix or
   cdelt keywords are missing. [#3316]
 
+- Convert FSM correction values from arcsec to radians. [#3367]
+
 srctype
 -------
 
--- a/jwst/lib/set_telescope_pointing.py
+++ b/jwst/lib/set_telescope_pointing.py
@@ -573,7 +573,7 @@ def calc_wcs(pointing, siaf, **transform_kwargs):
     return (wcsinfo, vinfo)
 
 
-def calc_transforms(pointing, siaf, fsmcorr_version='latest', j2fgs_transpose=True):
+def calc_transforms(pointing, siaf, fsmcorr_version='latest', fsmcorr_units='arcsec', j2fgs_transpose=True):
     """Calculate transforms from pointing to SIAF
 
     Given the spacecraft pointing parameters and the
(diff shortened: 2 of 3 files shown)


hbushouse:
Closing and reopening to retrigger Jenkins CI.

stscieisenhamer:
Ummm, good question. So, doing a quick removal of the except from the test, the initial error seems to be at [this point](https://github.com/spacetelescope/jwst/blob/8f7ee1eee834852dc23b7f2f50e564e7fc7de394/jwst/lib/set_telescope_pointing.py#L264). If the `else` is taken, `siaf` is left undefined but passed to `update_wcs_from_telem`. A default siaf should be assigned here. Methinks the statement should be `siaf = SIAF()`.

hbushouse:
You're right (as usual). Right now it's throwing an error because ``siaf`` is undefined, whereas if we add the line ``siaf = SIAF()`` it then makes it to the point where ``set_telescope_pointing`` raises the expected ``ValueError``, which is then trapped as an exception in the test module.

However, that only solves the first part of the problem with that test module. It then goes on to execute a list of ``assert`` statements on attributes that don't exist and hence throws yet another error.

hbushouse:
Jenkins failure due to server reboot. Travis CI passed, so I'm merging.

PR #3326, opened 2019-04-08
Flag Dq for wavelength outside range

jemorrison:

PR #3369, opened 2019-04-16
Update NIRSpec FS master_background reg test

hbushouse:
The regression tests for the ``master_background`` step using NIRSpec fixed-slit data and a user-supplied bkg spectrum no longer work correctly, since the merging of #3326. This change sets DQ='DO_NOT_USE' flags in the output 2-D image from ``master_background``, which then causes the ``extract_1d`` step to ignore those pixels when producing 1-D spectra. This is fine and as it should be.

However, parts of the regression test were setup to compare these results against real-time extractions from the original 2-D image that had no background signal added to it. This image has no DQ flags set to DO_NOT_USE, and hence includes a different set of pixels in its output than what's done for the background-subtracted image. This leads to large enough differences in the extracted 1-D spectra to cause the test to fail.

The part of the test that does real-time extractions from the original image and performs comparisons has been removed. It's impossible to get a match when comparing apples and oranges. What remains is the comparison of a truth file only. The corresponding truth file has been updated for the new results that have DQ flags set.

### DIFF
--- a/jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
+++ b/jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
@@ -190,101 +190,17 @@ class TestNIRSpecMasterBackground_FS(BaseJWSTTest):
         # input file has 2-D background image added to it
         input_file = self.get_data(*self.test_dir,
                                     'nrs_sci+bkg_cal.fits')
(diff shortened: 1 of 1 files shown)


hbushouse:
Jenkins is having problems. All Travis CI tests passed. Merging.

PR #3370, opened 2019-04-16
Master background regression tests for MIRI MRS and LRS

jemorrison:
Added four regression tests for masterbackground subtractions.
1. mrs nodded
2. mrs dedicated background
3. lrs nodded
4. lrs dedicated background
These test are done. I might need to reproduce the data once all the changes to extract1d have been finished for the nodded and dedicated background cases. Then I will put the data onto byesalad.

@jdavies-st I did not remove the auto_toggle_docopy because I really do not understand what it is 
trying to do when you run the tests locally. It seems it should always fail. Instead I changed the way
get_data works. Please look at the logical for docopy. I think it is what we want - do exactly what
the user wants and if nothing is given (docopy = None) then default back to the way it worked before
and use self.docopy.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -132,8 +132,8 @@ srctype
 
 tests
 -----
-- Removed set do_copy = True for get_data [#3370] 
-- Removed auto_toggle_do_copy  [#3370]
+- in base_classes.py forced do_copy = True for def get_data [#3370] 
+- in base_classes.py removed auto_toggle_do_copy  [#3370]
 tweakreg
 --------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -130,10 +130,6 @@ srctype
 
 - Updated logic for background targets and nodded exposures. [#3310]
 
-tests
------
-- in base_classes.py forced do_copy = True for def get_data [#3370] 
-- in base_classes.py removed auto_toggle_do_copy  [#3370]
 tweakreg
 --------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -130,6 +130,10 @@ srctype
 
 - Updated logic for background targets and nodded exposures. [#3310]
 
+tests
+-----
+- Removed set do_copy = True for get_data [#3370] 
+- Removed auto_toggle_do_copy  [#3370]
 tweakreg
 --------
 
--- a/jwst/tests/base_classes.py
+++ b/jwst/tests/base_classes.py
@@ -59,16 +59,21 @@ class BaseJWSTTest:
     def repo_path(self):
         return [self.inputs_root, self.env, self.input_loc]
 
-    def get_data(self, *pathargs, docopy=True):
+    def get_data(self, *pathargs, docopy=None):
         """
         Download `filename` into working directory using
         `artifactory_helpers/get_bigdata()`.
         This will then return the full path to the local copy of the file.
         """
-        # If user has specified action for no_copy, apply it with
+        # If user has specified action for docopy use it, If none
         # default behavior being whatever was defined in the base class.
-        local_file = get_bigdata(*self.repo_path, *pathargs, docopy=self.docopy)
-
+        if docopy is None: 
+            local_file = get_bigdata(*self.repo_path, *pathargs, docopy=self.docopy)
+        elif docopy:
+            local_file = get_bigdata(*self.repo_path, *pathargs, docopy=True)
+        else:
+            local_file = get_bigdata(*self.repo_path, *pathargs, docopy=False)
+            
         return local_file
 
     def compare_outputs(self, outputs, raise_error=True, **kwargs):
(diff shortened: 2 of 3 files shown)


jwst/tests/base_classes.py
```diff
@@ -59,16 +59,21 @@ def auto_toggle_docopy(self):
...
+        # If user has specified action for docopy use it, If none
         # default behavior being whatever was defined in the base class.
-        local_file = get_bigdata(*self.repo_path, *pathargs, docopy=self.docopy)
-
+        if docopy is None: 
+            local_file = get_bigdata(*self.repo_path, *pathargs, docopy=self.docopy)
+        elif docopy:
+            local_file = get_bigdata(*self.repo_path, *pathargs, docopy=True)
+        else:
+            local_file = get_bigdata(*self.repo_path, *pathargs, docopy=False)
```
jdavies-st: Nice fix here, though I think we can be even a bit more ruthless.  I think we can just completely get rid of `self.docopy`.  I.e we can delete the fixture

https://github.com/spacetelescope/jwst/blob/0e282178154de761d09459ac8d94458e7ea14012/jwst/tests/base_classes.py#L50-L56

And have `get_data()` always use `docopy=True`.  Since we are hardcoding it in tests, there's no reason to have it be dynamically set in the test infrastructure.  Besides, I think the use case of `docopy=False` has disappeared.

jemorrison:
Ok I tried as a Bundle and I think I put it in the wrong directory.
You will see it in the dev directory. I was trying for one above that.

jemorrison:
@jdavies-st 
I changed the base testing class. For now I just commented out what  I thought we did not want.
Once you review and if you agree I can take the comments out to make it cleaner. 
Jane

jwst/tests/base_classes.py
```diff
@@ -65,10 +65,15 @@ def get_data(self, *pathargs, docopy=True):
...
         """
-        # If user has specified action for no_copy, apply it with
+        # If user has specified action for docopy use it, If none
         # default behavior being whatever was defined in the base class.
-        local_file = get_bigdata(*self.repo_path, *pathargs, docopy=self.docopy)
-
+#        if docopy is None: 
+#            local_file = get_bigdata(*self.repo_path, *pathargs, docopy=self.docopy)
+#        elif docopy:
+        local_file = get_bigdata(*self.repo_path, *pathargs, docopy=True)
```
jdavies-st: You want `docopy=docopy` here so it respects the API of `get_data()`.  Otherwise looks good.  Delete all the commented-out stuff.  👍

jwst/tests_nightly/general/miri/test_miri_steps_single.py
```diff
@@ -359,26 +359,182 @@ def test_miri_lrs_masterbg_user(self):
...
+    ref_loc = ['test_masterbackground', 'mrs', 'dedicated', 'truth']
+    test_dir = ['test_masterbackground', 'mrs', 'dedicated']
+
+    rtol = 0.000001
+
+    def test_miri_masterbg_mrs_dedicated(self):
+        """Run masterbackground step on MIRI MRS association"""
+        asn_file = self.get_data(*self.test_dir,
+                                  'miri_mrs_mbkg_0304_spec3_asn.json',
+                                  docopy=True)
```
jdavies-st: Since `docopy=True` always now, you don't need this in the call to `get_data`, nor in the subsequent 7 calls to the function.

CHANGES.rst
```diff
@@ -130,6 +130,10 @@ srctype
 
 - Updated logic for background targets and nodded exposures. [#3310]
 
+tests
+-----
+- Removed set do_copy = True for get_data [#3370]
```
hbushouse: I'm not sure we really need a change log entry for this, but if we do want it there, then the description needs to be expanded to indicate which test(s) we're talking about.

jwst/tests_nightly/general/miri/test_miri_steps_single.py
```diff
@@ -284,9 +284,7 @@ def test_miri_setpointing(self):
 
         # Copy original version of file to test file, which will get overwritten by test
         input_file = self.get_data(self.test_dir,
-                                    'jw80600010001_02101_00001_mirimage_uncal_orig.fits',
-                                    docopy=True  # always produce local copy
```
hbushouse: Any tests of ``set_telescope_pointing`` may need to explicitly make a copy, because they modify the input in place. Or is this OK, because we always start with a fresh copy of the original input file for each test run?

jdavies-st: `docopy` defaults to `True`, so it is OK.  And we always start with a fresh copy.

CHANGES.rst
jdavies-st: In the past, we have not added any entries to CHANGES.rst for test infrastructure.  I suspect we don't need to, as it is internal to the package and not user-facing.

jemorrison: I was not sure what to do about changing the base class in the tests directory.
It seems I should log this but I was not sure were. I make it clearer what I did - changed
the base class.

jdavies-st:
I've moved the input and truth files to the correct locations on `bytesalad`.  So these tests are ready to go.

PR #3371, opened 2019-04-16
Simplify MIRI LRS master background regression test

jdavies-st:
The test
```
jwst.tests_nightly.general.miri.test_miri_steps_single.TestMIRIMasterBackgroundLRS.test_miri_masterbackground_lrs_user1d
```
has been failing since #3326 was merged.

This PR removes the part of the test that compares the `extract1d` outputs, as the DQ array affects the wavelengths compared.

Related to #3369.  See comments in that PR for more detail.

### DIFF
--- a/jwst/tests_nightly/general/miri/test_miri_steps_single.py
+++ b/jwst/tests_nightly/general/miri/test_miri_steps_single.py
@@ -312,61 +312,16 @@ class TestMIRIMasterBackgroundLRS(BaseJWSTTest):
         """
 
         # input file has the background added
-        input_file = self.get_data(*self.test_dir,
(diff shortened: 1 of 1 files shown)


hbushouse:
Could potentially make the tests smarter by doing comparisons/differences on only the pixels that have not been flagged as "DO_NOT_USE" by ``master_background``. But that won't solve the issue of comparing 1-D extractions from images that do and do not have those flags set.

jemorrison:
If we first determined the pixels that are good pixel (not DO_NOT_USE) in the output files then we could  make the comparisons with the data that does not have the DQ flags set only using these "good pixels"

PR #3378, opened 2019-04-18
Fix bugs in updated photom

hbushouse:
Fix a couple ``photom`` bugs that were in #3359, and update the step docs to indicate that the step now relies on having WCS (wavelength) info available, hence ``assign_wcs`` is a prerequisite.

Also updated one of the NRS MOS ``master_background`` regression tests to only check truth file, with no on-the-fly processing and comparison.

### DIFF
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -187,7 +187,7 @@ class DataSet():
 
                         # Populate the photometry keywords
                         self.input.meta.photometry.conversion_megajanskys = \
-                            conv_factor.value
+                            conv_factor
                         self.input.meta.photometry.conversion_microjanskys = \
                             conv_factor * MJSR_TO_UJA2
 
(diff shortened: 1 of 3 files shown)


jwst/photom/photom.py
```diff
@@ -187,7 +187,7 @@ def calc_nirspec(self, ftab, area_fname):
 
                         # Populate the photometry keywords
                         self.input.meta.photometry.conversion_megajanskys = \
-                            conv_factor.value
+                            conv_factor
```
jdavies-st: Is `tabdata['photmjsr']` an astropy quantity?  And if so, does `conv_factor.value` need to be used 2 lines down as well?

jdavies-st: Doh!  I see it is being _removed_ here.  Nevermind.  Looks good.

hbushouse: No, ``tabdata['photmjsr']`` is just a reference to a column in the photom reference file table, which contains a simple scalar value. That instance of ``conv_factor.value`` snuck into the code while I was playing around with adding the use of astropy.units to do the conversion from MJy/sr to uJy/arcsec^2, which I temporarily called ``conv_factor``, until I realised there was already a variable by that name. I obviously just missed this one line when backing out the new use of that variable name.

hbushouse:
Travis CI jobs are hanging for some reason and never completing. I ran pytest of the whole repo in my branch and it completes fine. None of the changes here are in any unit tests anyway. Merging despite the lack of successful Travis runs.

PR #3384, opened 2019-04-19
 Propagate flatfield uncertainties into science error and variance arrays

### LINKED ISSUE #3307: Error propagation due to flat-field for non-NRS modes
Issue [JP-609](https://jira.stsci.edu/browse/JP-609) was created by Nadia Dencheva:

This ticket captures the work defined in JP-55 for non Nirspec modes.

jdavies-st:
This propagates flat field reference file uncertainties to the science data via VAR_FLAT and ERR in `FlatFieldStep` for all modes except NIRSpec spectral modes.

Algorithm is below:

- ur = raw slope
- ff = flat field
- s = slope = ur / ff
- sigma_s^2 = (s^2/ur^2)sigma_r^2 + (s^2/ur^2)sigma_p^2 + (s^2/ff^2)sigma_ff^2
- Hence, the four uncertainties tracked are:
    - (s^2/ur^2)sigma_r^2 = sigma_r^2/ff^2
    - (s^2/ur^2)sigma_p^2 = sigma_p^2/ff^2
    - (s^2/ff^2)sigma_ff^2
    - sigma_s^2

Resolves #3307.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -67,6 +67,12 @@ extract_1d
   the areas of the source aperture and background annulus are computed
   differently. [#3362]
 
+flatfield
+---------
+
+- Propagate uncertainty from flat field into science ERR array and new
+  VAR_FLAT array which holds the variance due to the flat field.  [#3384]
+
 master_background
 -----------------
 
--- a/jwst/datamodels/schemas/cube.schema.yaml
+++ b/jwst/datamodels/schemas/cube.schema.yaml
@@ -52,4 +52,10 @@ allOf:
       default: 0.0
       ndim: 3
       datatype: float32
+    var_flat:
+      title: Integration-specific variances due to flat field
+      fits_hdu: VAR_FLAT
+      default: 0.0
+      ndim: 3
+      datatype: float32
 $schema: http://stsci.edu/schemas/fits-schema/fits-schema
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -158,11 +158,13 @@ def apply_flat_field(science, flat):
     if reffile_utils.ref_matches_sci(science, flat):
         flat_data = flat.data
         flat_dq = flat.dq
+        flat_err = flat.err
     else:
         log.info("Extracting matching subarray from flat")
         sub_flat = reffile_utils.get_subarray_model(science, flat)
         flat_data = sub_flat.data.copy()
         flat_dq = sub_flat.dq.copy()
+        flat_err = sub_flat.err.copy()
         sub_flat.close()
 
     # Find pixels in the flat that have a value of NaN and set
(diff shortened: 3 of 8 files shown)


CHANGES.rst
```diff
@@ -67,6 +67,12 @@ extract_1d
   the areas of the source aperture and background annulus are computed
   differently. [#3362]
 
+flatfield
+---------
+
+- Propagate uncertainty from flat field into science ERR array and new
+  VAR_FLAT array which holds the variance due to the flat field.  [#3382]
```
hbushouse: Actual PR number is 3384.

jwst/flatfield/flat_field.py
```diff
@@ -187,17 +189,40 @@ def apply_flat_field(science, flat):
     # For CubeModel science data, apply flat to each integration
     if isinstance(science, datamodels.CubeModel):
         for integ in range(science.data.shape[0]):
```
hbushouse: Nothing to do with this particular PR, but is there a reason why the flat needs to be applied to 3-D science data by looping over each integration individually, as opposed to just letting numpy array arithmetic do the 3d-by-2d division for us in one fell swoop?

CHANGES.rst
jdavies-st: Doh!  I tried guessing.  :joy:

jwst/flatfield/flat_field.py
jdavies-st: I.e. allowing array broadcasting to do its work?  The flat is 2D and science is 3D.

Good question.  I can try it out.

jwst/flatfield/flat_field.py
```diff
@@ -187,17 +189,40 @@ def apply_flat_field(science, flat):
...
     # For 2D ImageModel science data, apply flat to entire arrays
     else:
         # Flatten data and error arrays
         science.data /= flat_data
-        science.err /= flat_data
+
+        # Update the variances using BASELINE algorithm
+        science.var_poisson /= flat_data**2
+        science.var_rnoise /= flat_data**2
+        science.var_flat = science.data**2 / flat_data**2 * flat_err**2
```
nden: Here and above could you compute `flat_data * *2` once, assign it to a variable and reuse it.

jwst/datamodels/schemas/cube.schema.yaml
```diff
@@ -52,4 +52,10 @@ allOf:
       default: 0.0
       ndim: 3
       datatype: float32
+    var_flat:
+      title: Integration-specific variances due to flat field
+      fits_hdu: VAR_FLAT
+      default: 0.0
```
nden: Is this going to create the array in all data?

And for my understanding - isn't this a 2D array for CubeModel (since this is not the ifu cube)?

jwst/flatfield/flat_field.py
jdavies-st: Turns out `numpy` array broadcasting takes care of this automatically.  I'll update (and simplify) the code.  Thanks for the suggestion.

jwst/datamodels/schemas/cube.schema.yaml
jdavies-st: No, it will only create the array when accessed.  We don't access it on `__init__` in the datamodel, so will only get created (if it doesn't exist already from `ramp_fit`) in the `flat_field` step itself.

`CubeModel` is the multi-integration cube, so we need variance for every integration, even for flat field, as it depends not just on the input 2D flat ERR but also on the 3D science data array.  See algorithm above.

jwst/flatfield/flat_field.py
```diff
@@ -184,23 +186,27 @@ def apply_flat_field(science, flat):
...
+    science.err = np.sqrt(
+        science.var_poisson +
+        science.var_rnoise +
+        science.var_flat
+        )
 
-        # Combine the science and flat DQ arrays
-        science.dq = np.bitwise_or(science.dq, flat_dq)
+    # Combine the science and flat DQ arrays
+    science.dq = np.bitwise_or(science.dq, flat_dq)
```
hbushouse: I think the original version used the loop over integrations for CubeModel because I wasn't sure if the ``np.bitwise_or()`` would handle broadcasting the 2-d ``flat_dq`` to a 3-d ``science.dq``. Will it?

jdavies-st: Yes, `np.bitwise_or()` followings array broadcasting rules just like addition, subtraction, etc.

jdavies-st:
Ignoring test failures from `asdf-2.4.0.dev56+g1d9a56c`.  Merging.

PR #3386, opened 2019-04-22
Update 3 schemas in datamodels to comply with ASDF standard

jdavies-st:
We were seeing some failures testing against the latest dev version of `asdf`, version `2.4.0.dev56+g1d9a56c` and above.  Example here:

https://travis-ci.org/spacetelescope/jwst/jobs/522289192

And it appears that changes that `asdf` pulled in from `asdf-standard` submodule caused the issue.

https://github.com/spacetelescope/asdf/pull/678/files

Not sure which `asdf-standard` changes caused the problem in `jwst.datamodels`.  This PR fixes it.  It looks like our schema fragments were too "fragmentary".

The upside is that our schemas now have less code duplication. and look a bit cleaner.


jdavies-st:
Similar changes were made in `asdf-standard`:

https://github.com/spacetelescope/asdf-standard/commit/f6b50c188944f8b72408f54534cf34b8957344fe

And this is probably the culprit:

https://github.com/spacetelescope/asdf-standard/commit/6cb49f42c3f578924f3ca19547f551346a9f6d4e

jwst/datamodels/schemas/dq_def.schema.yaml
```diff
@@ -1,12 +1,16 @@
...
-- name: BIT
-  datatype: int32
-- name: VALUE
-  datatype: uint32
-- name: NAME
-  datatype: [ascii, 40]
-- name: DESCRIPTION
-  datatype: [ascii, 80]
+allOf:
+- type: object
```
nden: `allOf` above is not needed.

jwst/datamodels/schemas/group.schema.yaml
```diff
@@ -1,30 +1,34 @@
...
-  datatype: int16
-- name: completion_code_number
-  datatype: int16
-- name: completion_code_text
-  datatype: [ascii, 36]
-- name: bary_end_time
-  datatype: float64
-- name: helio_end_time
-  datatype: float64
+allOf:
```
nden: remove `allOf`

jwst/datamodels/schemas/int_times.schema.yaml
```diff
@@ -1,18 +1,22 @@
...
-  datatype: float64
-- name: int_end_MJD_UTC
-  datatype: float64
-- name: int_start_BJD_TDB
-  datatype: float64
-- name: int_mid_BJD_TDB
-  datatype: float64
-- name: int_end_BJD_TDB
-  datatype: float64
+allOf:
```
nden: and here

Issue #3301, opened 2019-04-01
revise the use of photom, pathloss and bar shadow corrections in the pipeline


stscijgbot:
Issue [JP-605](https://jira.stsci.edu/browse/JP-605) was created by Nadia Dencheva:

The corrections computed in the pathloss and barshadow steps will be applied now in the respective steps. The flux calibration will also be done in the photom step. The way these calibrations are used in other steps in the pipeline currently needs to be reviewed and possibly revised.

stscijgbot:
Comment by Nadia Dencheva: One example is extract_1d where the "net" and "neterror" columns can be removed.

stscijgbot:
Comment by Howard Bushouse: Inspection of all the pipeline/steps code shows that the barshadow correction data are not currently used anywhere else and hence changes to the barshadow do not have any other consequences. As noted in JP-292, however, we may need to start making use of the barshadow correction array in the master_background step when processing NIRSpec MSA data.

stscijgbot:
Comment by Howard Bushouse: Inspection of all the pipeline/steps code shows no indication of any of the pathloss correction arrays (either 1-d or 2-d) being used anywhere outside of the pathloss step itself. So there doesn't appear to be any need for other updates as a consequence of applying the correction or removing/renaming the extensions from the output products.

stscijgbot:
Comment by Nadia Dencheva: Updates to extract_!d to remove any logic based on `relsens` and remove the `net` and `neterr` columns from SpecModel. Two new columns should be added - one for surface brightness and one for surface brightness error.

stscijgbot:
Comment by Howard Bushouse: ``extract_1d`` is not behaving correctly for slit-like inputs where the ``photom`` step has already applied the flux calibration to the 2-D data and the ``relsens`` extension is no longer present, because of logic like this:

[https://github.com/spacetelescope/jwst/blob/master/jwst/extract_1d/extract.py#L2710]

where it sets the ``flux`` column to all zeros and leaves the extracted data in the ``net`` column.

stscijgbot:
[...]

PR #3387, opened 2019-04-22
Update photom to handle NIRISS SOSS

hbushouse:
Update the ``photom`` step to use the spectral order number when calling the ``get_wavelengths`` function for NIRISS SOSS mode, because the SOSS mode WCS is order-dependent. Also update the ``get_wavelengths`` function to make use of the ``order`` parameter when building the WCS transform for SOSS exposures.

Addresses some of the photom-related issues in #3301.

### DIFF
--- a/jwst/master_background/expand_to_2d.py
+++ b/jwst/master_background/expand_to_2d.py
@@ -4,7 +4,8 @@ import numpy as np
 
 from gwcs.wcstools import grid_from_bounding_box
 from .. import datamodels
-from .. assign_wcs import nirspec               # for NIRSpec IFU data
+from .. assign_wcs import nirspec   # For NIRSpec IFU data
+from .. assign_wcs import niriss    # For NIRISS SOSS data
 from .. datamodels import dqflags
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


jwst/master_background/expand_to_2d.py
```diff
@@ -329,8 +334,13 @@ def get_wavelengths(model):
             got_wavelength = False
             wl_array = None
 
+    # If no existing wavelength array, compute one
     if hasattr(model.meta, "wcs") and not got_wavelength:
-        wcs = model.meta.wcs
+        if hasattr(model.meta, "exposure") and model.meta.exposure.type == "NIS_SOSS":
+            transform = niriss.niriss_soss_set_input(model, order)
```
philhodge: It looks to me as if you could assign the value returned by `niriss_soss_set_input` directly to `wcs`, rather than first to `transform` and then to `wcs`.

nden: Yes, `niriss.niriss_soss_set_input` returns a WCS object/

hbushouse: Done.

hbushouse:
Travis failures using astropy_dev are due to ASDF issues documented in #3386, which are unrelated. Merging.

PR #3388, opened 2019-04-22
change indexing of model parameters

nden:
Instead of indexing model.Parameter index the value of the parameter.
This removes ambiguity in the current code where indexing is of a Parameter instance returns one parameter in a model set. With this changes the code works with current astropy.modeling master and the future refactored modeling.

In addition the prepare_inputs/outputs calls were removed as they are not necessary in this case.

A bug in models.V23ToSky was fixed.
Similar to spacetelescope/tweakwcs#85

### DIFF
--- a/jwst/transforms/tpcorr.py
+++ b/jwst/transforms/tpcorr.py
@@ -253,7 +253,7 @@ class TPCorr(Model):
         # convert cartesian to spherical coordinates:
         v2c, v3c = self.cartesian2spherical(zcr, xcr, ycr)
 
-        return v2c, v3c
+        return v2c.reshape(v2.shape), v3c.reshape(v3.shape)
 
 
     @property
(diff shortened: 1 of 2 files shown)


jwst/transforms/tpcorr.py
```diff
@@ -254,8 +253,8 @@ def evaluate(self, v2, v3, v2ref, v3ref, roll, matrix, shift):
         # convert cartesian to spherical coordinates:
         v2c, v3c = self.cartesian2spherical(zcr, xcr, ycr)
 
-        return self.prepare_outputs(format_info, v2c.reshape(v2.shape),
-                                    v3c.reshape(v3.shape))
+        return v2c, v3c
```
mcara:
```suggestion
        return v2c.reshape(v2.shape), v3c.reshape(v3.shape)
```

nden:
@mcara Suggestion merged and tests pass. Could you review this again.

PR #3393, opened 2019-04-24
Master_background regression test for nodded NIRSpec IFU observation

jemorrison:


jemorrison:
I just added the unit test for the master background. I forgot to make a new branch so it got into
this existing pr.

jemorrison:
@jdavies-st I put the regression data on artifactory. But I think I still messed it up.
I tarred up the dev and below and deployed it to artifacatory.  I think I should have not included
the dev directory but below that. Any look on artifactory under dev there is another dev. 
Ugh.

jdavies-st:
If you tar up `dev`, then you need to put it in `jwst-pipeline`.  Feel free to delete what you put there now and try again.

jwst/master_background/tests/test_master_background.py
```diff
@@ -63,3 +63,46 @@ def test_master_background_userbg(_jail, user_background):
...
+        image,
+        config_file='config/master_background.cfg',
+        user_background=user_background,
+        )
+
+    assert result.meta.cal_step.master_background == 'SKIPPED'
+    assert type(image) is type(result)
+
+
+# Now force it
```
jdavies-st: This should be indented.

jwst/master_background/tests/test_master_background.py
```diff
@@ -63,3 +63,46 @@ def test_master_background_userbg(_jail, user_background):
...
+    result = MasterBackgroundStep.call(
```
jdavies-st: This should be a separate test from the first.  If the first fails, this won't run.  I would make `image` a fixture just like `user_background` and pass it to both tests.

jwst/tests_nightly/general/nirspec/test_nirspec_steps_single.py
```diff
@@ -345,3 +346,49 @@ def test_nirspec_mos_masterbg_user(self):
...
+            self.compare_outputs(outputs)
+
+
+        # test 2
+        # compare the master background combined file to truth file
+        master_combined_bkg_file = 'ifu_prism_source_off_fix_NRS1_o001_masterbg.fits'
+        truth_background = self.get_data(*self.ref_loc,
+                                          master_combined_bkg_file)
+        outputs = [(master_combined_bkg_file, truth_background)]
+        self.compare_outputs(outputs)
```
jdavies-st: `compare_outputs` can compare many outputs.  I would compare them all at once.  It should only be called once per test.  So use the loop above to generate the outputs to compare, append the computed master background file, then send them all to `compare_outputs`.

See coron pipeline for an example:

https://github.com/spacetelescope/jwst/blob/7e412b5b0138bf2fb5dca61e87ecedb205a2c58b/jwst/tests_nightly/general/nircam/test_coron3_1.py#L37-L62

jemorrison:
@jdavies 
Still having problems getting the regression data set up on artificatory.
I deployed the dev.tar into the jwst_pipeline directory 
I deployed it as a "bundle" should not be doing that ?

jemorrison:
@jdavies-st check the first unit test in test_master_background.py. Since I added a science fixture I had to remove a check that image and result where the same. If you think I should add image = datamodels.open(science_image) so that I can check that image and result are the same type I can add that.

jemorrison:
@nden What do I need to do about this jenkins failure ?

nden:
@jemorrison - rebase on master to fix this.

jdavies-st:
You got everything on Artifactory!  Excellent.  Ran the tests myself.  Everything passes.  👍

The change mentioned above would make the tests better though.

jemorrison:
@nden  I fixed the travis failure by a rebase and forced push

PR #3396, opened 2019-04-24
fix dimensions of array in variance schema 

nden:


jwst/datamodels/schemas/slitdata.schema.yaml
```diff
@@ -92,6 +92,7 @@ allOf:
     data:
       title: The science data
       fits_hdu: SCI
+      max_ndim: 3
```
hbushouse: I'm trying to think where it is that we ever have slit-like data that are 3-d. Is it NIRSpec BrightObj? We only use 3-d for coronagraphic imaging and TSO imaging and spectra, and we only use the slit model for data that've gone through ``extract_2d``. The only TSO spectral mode that I can think of that might go through ``extract_2d`` is NRS BrightObj.

philhodge: Yes, is was for NRS_BRIGHTOBJ data that extract_1d failed because of a conflict with the schema.

PR #3399, opened 2019-04-25
Updates for JP-525, list cal and x1d background files in jw0023-o002-…

ddavis-stsci:
…spec3-asn association file.


.gitignore
```diff
@@ -20,3 +20,7 @@ relic/
 .pytest_cache
 results.xml
 pip-wheel-metadata/
+.DS_Store
+.project
+.pydevproject
+.vscode
```
jdavies-st: You'll want to add `*code-workspace` to `.gitignore` and a blank line at the end.

And remove the `*.code-workspace` files included in this PR.  We don't want these committed to the repo.

jwst/associations/tests/test_standards.py
```diff
@@ -0,0 +1,139 @@
+"""Test against Level2 standard associations
```
stscieisenhamer: This test fails for me on your new standards file. The truth has 4 associations, while the test generated 8.

jwst/associations/lib/rules_level3_base.py
```diff
@@ -294,6 +299,56 @@ def make_member(self, item):
         }
         return member
 
+    def make_fixedslit_bkg(self):
```
stscieisenhamer: Hmmm, apparently I forgot to save my previous commentary. Let's try this again....

If I have this right, the logic of this routine is basically is that, for each science member, add it as a background member and rename to 'x1d'.  The idea is that all the science should also be considered backgrounds in the processing.

If that is right, then a suggested refactor: The current routine is rebuilding the association from scratch. All that is necessary is to append the new backgrounds to the current association.

And hitting the save button...

jwst/associations/lib/rules_level3.py
```diff
@@ -273,7 +290,17 @@ def __init__(self, *args, **kwargs):
         self.constraints = Constraint([
             Constraint_Target(),
             Constraint_IFU(),
-        ])
+            Constraint(
```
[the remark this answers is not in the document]
hbushouse: INS claims that eventually IFU modes will allow TSO (they don't currently). I have no idea how we'll process them, but that's for the future. Anything that's TSO will never have any dithers/nods. Just sit and stare at the target.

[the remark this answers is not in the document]
hbushouse: Non-TSO IFU observations CAN and do have nods.

jwst/associations/lib/rules_level3.py
```diff
@@ -273,7 +290,17 @@ def __init__(self, *args, **kwargs):
...
+                [
+                    Constraint_TSO(),
+                    DMSAttrConstraint(
+                        name='patttype',
+                        sources=['patttype'],
+                        value=['2_point|4_point_nod|along_slit_nod'],
```
hbushouse: The versions of these strings with underscores is what appears in PPS DB tables, but in our FITS keywords those underscores will be hyphens/dashes.

stscieisenhamer: So, question, when do they get converted? This is comparing to what appears in the pools. Will the pools get populated with the FITS-version or the PPS-version?

jwst/associations/tests/test_standards.py
ddavis-stsci: James wants me to wait until after the merge to update the truth files.

.gitignore
ddavis-stsci: Done

jwst/associations/lib/rules_level3_base.py.orig
```diff
@@ -0,0 +1,804 @@
+"""Base classes which define the Level3 Associations"""
```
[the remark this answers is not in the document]
ddavis-stsci: Done

scripts/launch.json
```diff
@@ -0,0 +1,42 @@
+{
```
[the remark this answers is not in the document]
ddavis-stsci: removed

jwst/associations/lib/rules_level3.py
```diff
@@ -5,6 +5,7 @@
 from jwst.associations.registry import RegistryMarker
 from jwst.associations.lib.dms_base import (ACQ_EXP_TYPES, Constraint_TSO)
 from jwst.associations.lib.rules_level3_base import *
+from jwst.associations.lib.rules_level2_base import Constraint_Single_Science
```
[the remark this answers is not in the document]
ddavis-stsci: Done

jwst/associations/lib/rules_level3.py
```diff
@@ -208,6 +210,21 @@ def __init__(self, *args, **kwargs):
...
+
+        Returns
+        -------
+        associations: [association[, ...]] or None
+            List of fully-qualified associations that this association
+            represents.
+            `None` if a complete association cannot be produced.
+
+        """
+        return self.make_fixedslit_bkg()
```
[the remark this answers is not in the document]
ddavis-stsci: Done

jwst/associations/lib/rules_level3_base.py
ddavis-stsci: refactored

ddavis-stsci:
Code update and cleanup.

jwst/associations/lib/rules_level3.py
hbushouse: I believe the pools are populated from the FITS keyword values, but we should confirm with SDP (i.e. Mike Swam).

jwst/associations/lib/rules_level3_base.py
stscieisenhamer: Nice! A few more comments below...

jwst/associations/lib/rules_level3_base.py
```diff
@@ -294,6 +299,50 @@ def make_member(self, item):
...
+
+            # Split out the science vs. non-science
+            # The non-science exposures will get attached
+            # to every resulting association.
+            science_exps = [
+                member
+                for member in members
+                if member['exptype'] == 'science'
+            ]
+            nonscience_exps = [
```
stscieisenhamer: Since the code is now only appending to the current member list, this section, I believe, is unnecessary. If anything, it may be doubling up the non-science members, see below.

jwst/associations/lib/rules_level3_base.py
```diff
@@ -294,6 +299,50 @@ def make_member(self, item):
...
+                member
+                for member in members
+                if member['exptype'] != 'science'
+            ]
+            # Create new members for each science exposure in the association,
+            # using the the base name + _x1d as background.
+            results = []
+            new_members = self.current_product['members']
```
stscieisenhamer: No longer necessary? You have `members` already for the current product, from line 305. Can you just append back to `members`?

jwst/associations/lib/rules_level3_base.py
```diff
@@ -294,6 +299,50 @@ def make_member(self, item):
...
+                bkg_name = remove_suffix(
+                    splitext(split(science_exp['expname'])[1])[0])[0]
+                bkg_name = bkg_name+'_x1d.fits'
+                now_background = copy.copy(science_exp)
+                now_background['expname'] = bkg_name
+                now_background['exptype'] = 'background'
+                # Add the background file to the association table
+                new_members.append(now_background)
+
+            new_members += nonscience_exps
```
stscieisenhamer: This appears to be doubling the non-science members since they are already apart of the new_members list.

jwst/associations/lib/rules_level3.py
ddavis-stsci: Mike S. confirms that with the delivery of b7.3 the FITS keyword and the asn pool entries will use dashes (-) and not underscores (_) in the PATTTYPE keyword. Should this be a new ticket?

jwst/associations/lib/rules_level3_base.py
ddavis-stsci: removed

jwst/associations/lib/rules_level3_base.py
ddavis-stsci: code now works directly on the members.

jwst/associations/lib/rules_level3_base.py
ddavis-stsci: removed.

jwst/associations/lib/rules_level3.py
```diff
@@ -193,6 +194,7 @@ def __init__(self, *args, **kwargs):
             ),
             Constraint_Optical_Path(),
             Constraint_Target(),
+
```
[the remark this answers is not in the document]
ddavis-stsci: Done

jwst/associations/lib/rules_level3.py
stscieisenhamer: If any new tests in this PR do not require PATTTYPE, then it is probably best as a separate issue, since it will affect a number of other tests. However, if any new tests are implemented in this PR that require PATTTYPE, it is probably best to handle in this PR because might as well get it right in the first place.

nden: @ddavis-stsci Was a ticket filed?

ddavis-stsci: I just filed the ticket JP-664

PR #3400, opened 2019-04-25
Add unit tests for group scaling

mfixstsci:
Here is a list of unit test for the group scale step of the pipeline. The tests asked for by the team are in this table https://outerspace.stsci.edu/pages/viewpage.action?spaceKey=JWSTPWG&title=Group+Scale and I also made sure to add some other tests just to get better code coverage.

### DIFF
--- a/jwst/group_scale/tests/test_group_scale.py
+++ b/jwst/group_scale/tests/test_group_scale.py
@@ -38,13 +38,9 @@ def test_nframes_not_equal_frame_divisor(make_rampmodel):
     # Assert that the step completed
     assert(output.meta.cal_step.group_scale == 'COMPLETE')
     
-    # Since the step multiplies data by scale = frame_divisor/nframes (4/2=2)
-    # and the data array is all ones, the output array should contain 
-    # all twos since the correction applied is data*=scale.
-
-    # This assertion doesn't verify output like explained above, 
+    # This assertion doesn't verify for correct output, 
     # it just checks that the correction ran and that the data array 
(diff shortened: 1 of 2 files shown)


jwst/group_scale/tests/test_group_scale.py
```diff
@@ -0,0 +1,144 @@
...
+# as defined in the confluence table for this step. Here we set the read pattern 
+# mode to 'NRSIRS2RAPID' to make sure it fails on purpose.
+#
+# @pytest.mark.xfail
+# def test_correction_only_for_irs2(make_rampmodel):
+#     """Make sure pipeline doesnt run correction on data that isnt IRS2
+#     """
+#     datmod = make_rampmodel(2, 2, 4, 2048, 2048)
+#     datmod.meta.readpatt = 'NRSIRS2RAPID'
+#     output = GroupScaleStep.call(datmod)
```
jdavies-st: I think you can uncomment the test since you've marked it as xfail.  Perhaps do
```
@pytest.mark.xfail(reason='Not implemented yet")
```
or something more descriptive.

mfixstsci: Hey @jdavies-st I have been looking at the source code some more and from what I am noticing the code is written to be flexible and work for any `instrument` regardless of the `readpatt`. This seems to be handled in the reference files or are assigned in the headers by CRDS (outside of the scope of the JWST testing group). So I think I will just go ahead and remove it!

PR #3402, opened 2019-04-25
Exclude GuiderCalModel from flatfield error propogation

jdavies-st:
Follow-on to #3384 to exclude `GuiderCalModel` from error propogation from `FlatFieldStep`.

The guider pipeline currently does not do anything with errors.  The `err` attribute is not populated and `var_rnoise` and `var_poisson` don't exist in the schema for `GuiderCalModel`.  So there's no way to properly add in the variance due to the flat field.  So we skip it for now.

Tested against the 3 guider pipeline regression tests.  And they all will pass with this PR.


hbushouse:
Something appears to be wrong with the Jenkins tests. Travis tests came out fine, so I'm merging.

PR #3404, opened 2019-04-25
resample_spec wcs

### LINKED ISSUE #3394: WCS transforms of resampled spectral data do not work correctly with negative and positive angles
Issue [JP-654](https://jira.stsci.edu/browse/JP-654) was created by Nadia Dencheva:

The WCS of s2d files can produce negative RA values. The inverse transform roundtrips with negative values but fails with 360 + RA. The reason is that the spatial models fit a Linear1D model between pixels and angles.

nden:
Resolves #3394

This PR solves an issue reported by @philhodge 
```
wcs.invert(-1, dec, lam) != wcs.invert(359, dec, lam)
```

### DIFF
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -5,7 +5,8 @@ import numpy as np
 
 from astropy import coordinates as coord
 from astropy import units as u
-from astropy.modeling.models import Mapping, Tabular1D, Linear1D, Pix2Sky_TAN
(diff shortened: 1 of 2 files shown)


jdavies-st:
A note that this will cause several regression tests to fail because of the above mentioned differences in `S_REGION` and a very small number of pixels in the output data that have changed because of the slight change in mapping.

PR #3405, opened 2019-04-25
unit test for background

jemorrison:
Unit test for calspec2 background subtraction - testing logic when the data is NISPEC. The GWA x,y values have to be the same between science image and background.

I was not sure if I had to run assign_wcs on the data. I had this in the code at first but when I took
it out it did not seem to matter that is was not run. If I need it for some reason  let me know. If I don't need it I can probably take out all the setting of the wcs values. 

I used the _jail  for the background image since the input to background step is a list of background
filenames - at least I think that is correct.

### DIFF
--- a/jwst/background/tests/test_background.py
+++ b/jwst/background/tests/test_background.py
@@ -37,12 +37,10 @@ def background(tmpdir_factory):
     return filename
 
 
-@pytest.fixture(scope='module')
-def science_image(tmpdir_factory):
+@pytest.fixture(scope='function')
+def science_image():
     """Generate science image"""
 
-    filename = tmpdir_factory.mktemp('background_input')
-    filename = str(filename.join('science_image.fits'))
     image = datamodels.IFUImageModel((10, 10))
     image.data[:,:] = 100
     image.meta.instrument.name = 'NIRSPEC'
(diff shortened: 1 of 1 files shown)


jemorrison:
@jdavies should I make the science image a fixture. For now I just left it defining it in each step
but I can make it a fixture if you think that is better

nden:
@jemorrison The Jenkins build error is unrelated - please rebase to get rid of it.

nden:
And yes, a science file fixture would simplify the code and decrease code repetition.

jemorrison:
added a science fixture

jdavies-st:
For the science image, it doesn't need to be created and saved.  It just needs to be generated on the fly in memory and passed to the test.  So no need to write it out and pass the path to the test function.  Instead, just have the fixture return the datamodel itself, and pass that datamodel to the Step.  This will run faster and have less overhead.  The science image can be scoped to "function" instead of "module" so you get a fresh one for each test.

jwst/background/tests/test_background.py
```diff
@@ -0,0 +1,145 @@
...
+    image.meta.instrument.gwa_xtilt = 0.0001
+    image.meta.instrument.gwa_ytilt = 0.0001
+    image.meta.instrument.gwa_tilt = 37.0610
+
+    image.save(filename)
+    return filename
+
+
+@pytest.fixture(scope='module')
+def science_image(tmpdir_factory):
```
jdavies-st:
```
@pytest.fixture(scope='function')
def science_image():
```

jwst/background/tests/test_background.py
```diff
@@ -0,0 +1,145 @@
...
+    image.meta.date = '2019-02-27T13:37:18.548'
+    image.meta.subarray.xstart = 1
+    image.meta.subarray.ystart = 1
+
+    image.meta.instrument.gwa_xtilt = 0.0001
+    image.meta.instrument.gwa_ytilt = 0.0001
+    image.meta.instrument.gwa_tilt = 37.0610
+
+    image.save(filename)
+    return filename
```
jdavies-st: ~~image.save(filename)~~
```
return image
```

jwst/background/tests/test_background.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def test_nirspec_gwa(_jail, background, science_image):
+    """Verify NIRSPEC GWA logic for in the science and background"""
+
+    image = datamodels.open(science_image)
```
jdavies-st: ~image = datamodels.open(science_image)~

jwst/background/tests/test_background.py
```diff
@@ -0,0 +1,145 @@
...
+    image.meta.instrument.gwa_ytilt = back_image.meta.instrument.gwa_ytilt
+    image.meta.instrument.gwa_tilt = back_image.meta.instrument.gwa_tilt
+
+    bkg = [background]
+    # Test Run with GWA values the same - confirm it runs
+    # And gives the predicted result
+
+    collect_pipeline_cfgs('./config')
+    result = BackgroundStep.call(
+        image, bkg,
```
jdavies-st:
```
result = BackgroundStep.call(
         science_image, bkg,
```

jdavies-st:
🎉

PR #3410, opened 2019-04-26
Specify pip<19.1 to avoid PEP 517 breakage

jdavies-st:
This is because of

https://github.com/pypa/pip/issues/6434

Closes #3409


jdavies-st:
Avoiding `pip 19.1` for now seems like the best bet until they sort it out.  Hopefully soon.

PR #3411, opened 2019-04-26
Don't use pip install -e in Jenkinsfiles

jdavies-st:


jdavies-st:
This didn't work. And is not desirable anyway.

PR #3417, opened 2019-04-29
Use 50% throughput MRS slice mask

### LINKED ISSUE #3375: Update MIRI MRS distortion solution to CDP-8b
Issue [JP-650](https://jira.stsci.edu/browse/JP-650) was created by David Law:

CDP-7 provided a critical update to the MIRI MRS distortion solution reference files that incorporates improvements in both the spatial and spectral dimensions.  The spatial update fixes a trace-curvature issue present in the CDP-6 solution, while the spectral update provides ground-test calibrations for the wavelength solution for additional channels (that were previously based on the instrument design model). 

Additional updates include conversion directly to the V2,V3 coordinate system (instead of XAN,YAN) and an update to the slice mask reference file format.  While the slice mask was previously a single 2d image for each detector configuration, it is now a 3d array of images each corresponding to a different throughput threshhold.  This format change fixes a previous issue in which 'stripes' of low-throughput data were visible at the edges of the MRS field of view in dithered exposures.

The new files provided in CDP-7 had a few issues that precluded them from being used for these updates, but sometime in April 2019 CDP-8beta editions should be available.  At that time there will need to be some modifications made to the MIRI assign_wcs code to handle the changes.  I'm assigning this work to myself, since it is closely related to modifications that I will need to make to produce the updated asdf reference files anyway.

drlaw1558:
This addresses https://jira.stsci.edu/browse/JP-650 and implements the update necessary to work with the new MRS distortion reference files.  Specifically, the REGIONS file to be delivered is a 3d array instead of a 2d array, and miri.py needs to select out the appropriate plane.

Resolves #3375


nden:
@drlaw1558 It'd be good to add some clarification on the structure of the new `REGIONS` reference file in `docs/jwst/references_general/regions_reffile.rst`. Where is the documentation on the new file structure and more specifically how many planes are in the file and what are they?

drlaw1558:
Ah- I was wondering if there was anything else that should be updated for the data model.  I'll take a look at that file and modify accordingly.
In brief though, there are 9 planes, each containing a mask at the 10%-90% throughput levels, with more extensive documentation at https://innerspace.stsci.edu/download/attachments/152259793/MIRI-TN-00005-KUL_iss1_MRSDetectorSliceMap.pdf

nden:
The schema for the model may need updating as well. I'll look into it tomorrow.

drlaw1558:
I've now updated the regions_reffile.rst with more information.

nden:
As expected the tests are "failing nicely" with the default CRDS_CONTEXT and passing with `EDIT`.

PR #3419, opened 2019-05-01
Add gain_scale unit tests

### LINKED ISSUE #3245: Detector 1: Gain Scale Unit Testing
Issue [JP-585](https://jira.stsci.edu/browse/JP-585) was created by Macarena Garcia Marin:

Ensure there is a set of unit-tests (to be used by SCSB) covering the NIRSpec Gain Scale pipeline step. This includes already exsting tests, newly created tests and test concepts that can be coded by SCSB.
 * Primary objective: provide a certain degree of code coverage (as close to 100% as possible).
 * Secondary objective: Identify and generate unit tests of possible edge cases.

Details provided in this confluence page
h4. [https://outerspace.stsci.edu/display/JWSTPWG/Gain+Scale]

mfixstsci:
Add unit tests to make sure the gain assigned to the header/meta extensions of the data models are applied to the data and error arrays properly.

The only portion of the code we couldn't reach with a test are when the reference files are required if the gain isn't found in the data model.

Closes #3245.

### DIFF
--- a/jwst/gain_scale/tests/test_gain_scale.py
+++ b/jwst/gain_scale/tests/test_gain_scale.py
@@ -11,23 +11,24 @@ import pytest
 def test_correction(make_rampmodel):
     """Make sure correct gain factor is applied
     """
-    datmod = make_rampmodel(2, 2048, 2048)
+    datmod = make_rampmodel(2, 50, 50)
     output = do_correction(datmod, 
                            gain_factor=datmod.meta.exposure.gain_factor)
 
     assert(output.meta.cal_step.gain_scale == 'COMPLETE')
-    assert np.all(output.err == datmod.err*datmod.meta.exposure.gain_factor)
+    assert np.all(output.err == datmod.err * datmod.meta.exposure.gain_factor)
+    assert np.all(output.data == datmod.data * datmod.meta.exposure.gain_factor)
 
 
 def test_step(make_rampmodel):
     """Make sure correct gain factor is applied at step level
     """
-    datmod = make_rampmodel(2, 2048, 2048)
+    datmod = make_rampmodel(2, 50, 50)
     output = GainScaleStep.call(datmod)
 
     assert(output.meta.cal_step.gain_scale == 'COMPLETE')
-    assert np.all(output.err == datmod.err*datmod.meta.exposure.gain_factor)
-    assert np.all(output.data == datmod.data*datmod.meta.exposure.gain_factor)
+    assert np.all(output.err == datmod.err * datmod.meta.exposure.gain_factor)
+    assert np.all(output.data == datmod.data * datmod.meta.exposure.gain_factor)
 
 
 @pytest.fixture(scope='function')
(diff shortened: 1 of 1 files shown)


jwst/gain_scale/tests/test_gain_scale.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_correction(make_rampmodel):
+    """Make sure correct gain factor is applied
+    """
+    datmod = make_rampmodel(2, 2048, 2048)
+    output = do_correction(datmod, 
+                           gain_factor=datmod.meta.exposure.gain_factor)
+
+    assert(output.meta.cal_step.gain_scale == 'COMPLETE')
+    assert np.all(output.err == datmod.err*datmod.meta.exposure.gain_factor)
```
hbushouse: Any reason not to test the output.data vs datmod.data too?

jwst/gain_scale/tests/test_gain_scale.py
```diff
@@ -0,0 +1,62 @@
...
+        csize = (nints, ngroups, ysize, xsize)
+        data = np.random.randint(low=1, high=50, size=csize)
+        err = np.random.randint(low=0.1, high=5.0, size=csize)
+
+        # create a JWST datamodel for NIRSPEC data
+        dm = RampModel(data=data, err=err)
+
+        dm.meta.instrument.name = 'NIRSPEC'
+        dm.meta.date = '2018-01-01'
+        dm.meta.description = 'Fake data'
```
hbushouse: Only reference file data models have meta attributes like description, reftype, author, pedigree, and useafter. They aren't defined (or needed) for science models like ``RampModel``. So lines 50-54 could be removed.

jwst/gain_scale/tests/test_gain_scale.py
```diff
@@ -0,0 +1,62 @@
...
+from jwst.datamodels import RampModel
+from jwst.gain_scale.gain_scale import do_correction
+from jwst.gain_scale import GainScaleStep
+import numpy as np
+import pytest
+
+def test_correction(make_rampmodel):
+    """Make sure correct gain factor is applied
+    """
+    datmod = make_rampmodel(2, 2048, 2048)
```
hbushouse: I have no idea how long this test takes to run, but it could be made faster by not using a full 2048x2048 data array. You don't really need to test that many pixels to confirm that the math is being applied properly. Besides, in real life, the gain_scale step only gets applied to subarray data anyway. I would think that something on the order of 50x50 or 200x200 would be sufficient.

jwst/gain_scale/tests/test_gain_scale.py
jdavies-st: Added

jwst/gain_scale/tests/test_gain_scale.py
jdavies-st: Done

jwst/gain_scale/tests/test_gain_scale.py
jdavies-st: Reduced size to 50x50.  Tests run very fast.  Now they run faster.

Issue #3414, opened 2019-04-29
Propagate flat-field variance in subsequent steps


stscijgbot:
Issue [JP-659](https://jira.stsci.edu/browse/JP-659) was created by Howard Bushouse:

Flat-field variances were added to the pipeline flow in [https://github.com/spacetelescope/jwst/pull/3384]. Pipeline steps that come after flat-fielding now need to be updated to apply their corrections to the flat-field variance, in addition to the existing Poisson and read noise variances. This includes the pathloss, barshadow, and photom steps.

hbushouse:
Fixed in #3427

PR #3427, opened 2019-05-03
Apply barshadow, pathloss, and photom correction to VAR_FLAT

philhodge:
The correction factors (squared) are applied to the VAR_POISSON and VAR_RNOISE variances in the `barshadow`, `pathloss`, and `photom` steps.  These correction factors (also squared) are now also applied to the VAR_FLAT variance array.

Note that this PR does not close issue #3414, because there are other steps that can follow `flat_field`.

### DIFF
--- a/jwst/barshadow/bar_shadow.py
+++ b/jwst/barshadow/bar_shadow.py
@@ -105,6 +105,8 @@ def do_correction(input_model, barshadow_model):
                 slitlet.err /= correction
                 slitlet.var_poisson /= correction**2
                 slitlet.var_rnoise /= correction**2
+                if slitlet.var_flat is not None and np.size(slitlet.var_flat) > 0:
+                    slitlet.var_flat /= correction**2
             else:
                 log.info("Slitlet %d has zero length, correction skipped" % slitlet_number)
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,8 @@ barshadow
 - Updated to apply the correction to the science data arrays, in addition
   to attaching as an extension. [#3319]
 
+- Updated to apply the square of the correction to VAR_FLAT [#3427]
+
 calwebb_spec3
 -------------
 
@@ -120,6 +122,8 @@ pathloss
 
 - Updated to apply the correction to the science data and err arrays. [#3323]
 
+- Updated to apply the square of the correction to VAR_FLAT [#3427]
+
 photom
 ------
 
@@ -128,6 +132,8 @@ photom
 
 - Updated to compute a wavelength array for NIRISS SOSS exposures using
   spectral order 1. [#3387]
+
+- Updated to apply the square of the correction to VAR_FLAT [#3427]
 
 reffile_utils
 -------------
--- a/docs/jwst/barshadow/description.rst
+++ b/docs/jwst/barshadow/description.rst
@@ -41,9 +41,10 @@ heights to shutter spacings, and then the Y and wavelength values are interpolat
 into the model to determine the correction for each pixel.
 
 Once the 2-D correction array for a slit has been computed, it is applied to the
-science (SCI), error (ERR), and Poisson variance (VAR_POISSON) data arrays of the slit.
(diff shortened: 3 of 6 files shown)


jwst/barshadow/bar_shadow.py
```diff
@@ -105,6 +105,8 @@ def do_correction(input_model, barshadow_model):
                 slitlet.err /= correction
                 slitlet.var_poisson /= correction**2
                 slitlet.var_rnoise /= correction**2
+                if slitlet.var_flat is not None and np.size(slitlet.var_flat) > 0:
```
jdavies-st: Neither of these checks on `slitlet.var_flat` will do anything because as soon as it is accessed, it becomes an array of zeros the same size as `slitlet.data`.  It will never be `None` and will always have a size.

I.e. the above line will always evalutate to `True`.  I don't know if this is desirable.  Perhaps this is due to the default value for `var_flat` in the schema?

This is a larger `datamodels` issue I think.

jdavies-st: Some is true for all the checks elsewhere.

jdavies-st:
```
In [17]: im = datamodels.SlitModel((2, 10))                                                                                                                              

In [18]: im._asdf.tree                                                                                                                                                   
Out[18]: {'meta': {'date': '2019-05-03T15:24:07.530', 'model_type': 'SlitModel'}}

In [19]: im.data is None                                                                                                                                                 
Out[19]: False

In [20]: im._asdf.tree                                                                                                                                                   
Out[20]: 
{'meta': {'date': '2019-05-03T15:24:07.530', 'model_type': 'SlitModel'},
 'data': array([[0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]], dtype=float32)}

In [21]: np.size(im.var_flat) > 0                                                                                                                                        
Out[21]: True

In [22]: im._asdf.tree                                                                                                                                                   
Out[22]: 
{'meta': {'date': '2019-05-03T15:24:07.530', 'model_type': 'SlitModel'},
 'data': array([[0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]], dtype=float32),
 'var_flat': array([[0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]], dtype=float32)}
```

philhodge: This is why I added those checks:

    dm = datamodels.open("jw87600017001_02101_00002_nis_extract_2d.fits")
    type(dm)
    <class 'jwst.datamodels.multislit.MultiSlitModel'>
    slit = dm.slits[0]
    slit.var_poisson.shape
    (0, 0, 0)
    slit.var_rnoise.shape
    (0, 0, 0)
    np.nanmean(slit.var_poisson)
    __main__:1: RuntimeWarning: Mean of empty slice
    nan

    catfits.py jw87600017001_02101_00002_nis_extract_2d.fits
    HDU#  FILENAME             EXTNAME    EXTVER DIMENS         BITPIX
    0     jw87600017001_02101_00002_nis_extract_2d.fits         8
    1       IMAGE              SCI        1      9x105          -32
    2       IMAGE              DQ         1      9x105          32
    3       IMAGE              ERR        1      9x105          -32
    4       BINTABLE           ASDF              1Fx1R          8

philhodge:
I would much rather remove the checks on the existence of the variance arrays, since they are messy and make the code a lot harder to read.  But the step did fail without the checks, when run with an input file that lacked VAR_POISSON and VAR_RNOISE.

jdavies-st:
Thanks for pointing that out Phil.  That's very different behavior when the slit is a direct attribute of a datamodel, i.e. `SlitModel().var_flat` versus being an attribute of the `.slits` list.  It looks like what `MultiSlitModel` is doing is a bug.

I'll file a separate issue about this.  Agree that the checks are needed and you can ignore my comments.

PR #3430, opened 2019-05-04
Remove python 2 exception .message references

jdavies-st:
Exceptions in Python 3 no longer have a `self.message` attribute.  Use `self.args` instead.

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -80,9 +80,7 @@ class Extract1dError(Exception):
 
 class InvalidSpectralOrderNumberError(Extract1dError):
     """The spectral order number was invalid or off the detector."""
-    def __init__(self, message=None):
-        super().__init__()
-        self.message = message
+    pass
 
 
 def load_ref_file(refname):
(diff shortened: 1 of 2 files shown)


jdavies-st:
```
In [13]: try: 
    ...:     raise Exception("The Message") 
    ...: except Exception as e: 
    ...:     ee = e 
    ...:                                                                                                                                                                 

In [14]: ee.message                                                                                                                                                      
AttributeError                            Traceback (most recent call last)
<ipython-input-14-346c734da2d9> in <module>
----> 1 ee.message

AttributeError: 'Exception' object has no attribute 'message'

In [15]: ee.args                                                                                                                                                         
Out[15]: ('The Message',)
```

jwst/extract_1d/extract.py
```diff
@@ -80,9 +80,7 @@ class Extract1dError(Exception):
 
 class InvalidSpectralOrderNumberError(Extract1dError):
     """The spectral order number was invalid or off the detector."""
-    def __init__(self, message=None):
-        super().__init__()
-        self.message = message
```
philhodge: Interesting!  Thanks for spotting that.

PR #3436, opened 2019-05-07
Fix flake8 checks and don't suppress Warnings

jdavies-st:
As a project, we've been suppressing our warnings for a long time, and our `flake8` config in `setup.cfg` has been broken for a long time.  This fixes both issues.

- Fix `flake8` config in `setup.cfg` to check the things we want checked
- Remove `flake8` expected failure build in Travis CI.  If we care about any more PEP8 codes, we should fix the problems and add the code to the `[flake8]` section of `setup.cfg`.  Now that our `flake8` section is done correctly, the project defaults are always applied just by invoking `flake8 jwst` at the project top level.
- Warnings are no longer suppressed.  Many are fixed.  Not all.
- Remove unnecessary `setuptools` imports in `setup.py` which used to be for building C extensions.  We haven't needed these since tweakreg C code was removed.
- Add a build with CRDS_CONTEXT="jwst-edit" to the build matrix on Travis CI with an allowed failure to catch any issues with new reffiles that our unit tests use before the make it into the default context.

For review purposes, look at `.travis.yml`, `setup.cfg` and `setup.py`.  All the other edited files are to fix our undetected Warning and PEP8 problems that have been accumulating over the past several months.  I tried to use my best judgement on those.

We still have 46 Warnings.  Many of these are benign, as they are from reffiles that were created with a dev version of `asdf` greater than the current released version.  These will go away with the next `asdf` release.

We should also set our max line width to a value lower than 130 eventually to make diffs easier to read.  And we should consider using [black](https://github.com/python/black) as an autoformatter in a git pre-commit hook so that we never have to think about this in the future again.  Most of the big projects including the core language are using it.  It minimizes diffs and makes review much easier.  I.e. never worry about formatting again.


jdavies-st:
I was "sprinting" with CPython (the python core library) yesterday, and running the tests took over an hour.  So I had some time on my hands while I was waiting.  =)

nden:
This needs conflict resolution now.

PR #3437, opened 2019-05-07
Make URL mapping of datamodels schemas consistent with other STScI packages

### LINKED ISSUE #3232: BaseExtension and JWSTExtension in datamodels are inconsistent
We have 2 ASDF extensions in the `jwst` package (they subclass `AsdfExtension`) that are provided via `entry_points` in our `setup.py`:

- `BaseExtension` : provides a mapping between a URL such as

        http://jwst.stsci.edu/schemas/image.schema.yaml

    and the location in this package of the datamodels schema

        jwst/datamodels/schemas/image.schema.yaml

    so that `asdf` and other code can find it.

- `JWSTExtension` : provides a mapping between a URL such as

        http://stsci.edu/schemas/jwst_pipeline/v23tosky-0.7.0

    and the location in this package of the `gwcs` transform schema

        jwst/transforms/schemas/stsci.edu/jwst_pipeline/v23tosky-0.7.0.yaml

    so that `asdf`, `gwcs` and other code can find it.  It also maps tags.

You'll notice immediately that the both the base URL and the syntax for the schemas is different for the two extensions:
```
http://jwst.stsci.edu/schemas/
http://stsci.edu/schemas/jwst_pipeline/
```
and
```
image.schema.yaml  -->  image.schema.yaml
v23tosky-0.7.0  ------>  v23tosky-0.7.0.yaml
```
These should be consistent.  But they are not.

We should make the base URL consistent.  Pick one.  It's just a name.  @hbushouse @nden @drdavella thoughts?

And we should make the datamodel schema not include `.yaml` in the URI.  So that means changing how `BaseExtension` operates.  See discussion here https://github.com/spacetelescope/jwst/pull/3227#discussion_r263530875

jdavies-st:
Make the datamodel `AsdfExtension.url_mapping` consistent with the same property in `asdf` and `JWSTExtension` in `jwst.transforms`.  This means that schemas are referenced without the `.yaml` extension and have a consistent URL.
- Datamodels schemas should now be referenced with

        http://stsci.edu/schemas/jwst_datamodel/image.schema

     instead of

        http://jwst.stsci.edu/schemas/image.schema.yaml

     to make it consistent with URLs for `asdf` and `jwst.transforms` tags.
- The datamodels `AsdfExtension` subclass is renamed internally from `BaseExtension` to `DataModelExtension`, so that it is more descriptive.  This has no effect on outside users of our package.
- Each datamodel class now references its schemas without the `.yaml` in the string.
- Use `DataModelExtension.url_mapping` in the internal schema tests

Resolves #3232.


nden:
Isn't the MIRI team using `BaseExtension`? Is the new name going to break their code?

jdavies-st:
That's a good question.  The new name for `BaseExtension` will not cause any problems to external users, and if they're only using our datamodels, then then that's not a problem either.

If they are referencing our schemas directly via a URI via `entry_points`, then the reference needs to be changed in their code. If they reference any of our schemas as
```
http://jwst.stsci.edu/schemas/image.schema.yaml
```
then they will have to change their references to be
```
http://stsci.edu/schemas/jwst_datamodel/image.schema
```
Who can we coordinate with at MIRI to make sure their code is ready for this change when they upgrade their version of `jwst` package?  I know in the past they've pinned their code to a released version of this package.

smbeard:
Yes, the MIRI software uses BaseExtension. A few months ago our code broke when this feature was removed. It was restored in this pull request:

[https://github.com/spacetelescope/jwst/pull/3227](https://github.com/spacetelescope/jwst/pull/3227)

However, as long as the functionality itself is not going away, we can incorporate the change by making some global changes in our code and schemas. Our code is currently pinned to an older JWST release, but if we know in advance this change is coming we can plan for it and coordinate when we upgrade.

jdavies-st:
Hi @smbeard, thanks for the input.  Yes, functionality will remain, but the URL will change as discussed above.

This will be a change that will appear in the `0.13.3` release sometime in the next month or so. If you pin your dependencies, then it will be straightforward to pin to the new version at the same time you change the URLs in your own code.

PR #3440, opened 2019-05-08
Add multi-processing option to jump step

jdavies-st:
This PR makes a few changes to @mwregan2's very nice #3021.

- Remove `maximum_cores='one'` and make `maximum_cores=None` as the default behavior that SDP expects.  I.e. if `maximum_cores` is not set, no multiprocessing pool is created. 
- Use smaller in the unit test fixtures (instead of full frame) to reduce memory usage and runtime, as is standard for a number of the other pipeline step unit tests where appropriate.  Runtime is reduced because of the `assert` statements in a `for` loop and scales by the number of pixels in the array.
- Parametrize the tests instead of the test fixtures for the multiprocessing tests.  Makes it clearer what is going on.
- Change test fixtures to use `GainModel` and `ReadnoiseModel` instead of custom FITS versions of these to reduce future maintenance.
- Add `psutil` and `pytest-faulthandler` as new package dependencies.

@mwregan2, feel free to look at my commits one-at-a-time below to see the changes.  Let me know what you think.  Hopefully these changes will get us closer to merging this.

Closes #3021.

### DIFF
--- a/jwst/jump/jump.py
+++ b/jwst/jump/jump.py
@@ -38,9 +38,9 @@ def detect_jumps (input_model, gain_model, readnoise_model,
     elif max_cores == 'quarter':
         numslices = np.int(np.floor(num_cores/4))
     elif max_cores == 'half':
-        numslices = np.int(np.floor(num_cores/2))
+        numslices = np.int(np.floor(num_cores/2)) - 1 # leave room for the main thread
     elif max_cores == 'all':
-        numslices = np.int(num_cores)
+        numslices = np.int(num_cores) - 1 # leave room for the main thread
     log.info("Creating %d processes for jump detection " % numslices)
     pool = multiprocessing.Pool(processes=numslices)
 
@@ -122,6 +122,8 @@ def detect_jumps (input_model, gain_model, readnoise_model,
                 gdq[:, :, k * yincrement:(k + 1) * yincrement, :] = resultslice[1]
             k += 1
 
+    pool.terminate()
+    pool.close()
     elapsed = time.time() - start
     log.debug('Elapsed time = %g sec' %elapsed)
 
--- a/jwst/jump/jump.py
+++ b/jwst/jump/jump.py
@@ -13,7 +13,7 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 def detect_jumps (input_model, gain_model, readnoise_model,
-                  rejection_threshold, do_yint, signal_threshold, max_cores):
+                  rejection_threshold, do_yint, signal_threshold, max_cores=None):
     """
     This is the high-level controlling routine for the jump detection process.
     It loads and sets the various input data and parameters needed by each of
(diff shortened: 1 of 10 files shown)


mwregan2:
James,
  These changes all look good to me. Thanks for improving the changes.
Mike

jwst/jump/jump.py
```diff
@@ -75,12 +89,43 @@ def detect_jumps (input_model, gain_model, readnoise_model,
...
+        slices.insert(i, (data[:, :, i * yincrement:(i + 1) * yincrement, :],
+                          gdq[:, :, i * yincrement:(i + 1) * yincrement, :],
+                          readnoise_2d[i * yincrement:(i + 1) * yincrement, :],
+                          rejection_threshold, nframes))
+    slices.insert(numslices - 1, (data[:, :, (numslices - 1) * yincrement:numy, :],  # last slice get the rest
+                                 gdq[:, :, (numslices - 1) * yincrement:numy, :],
+                                 readnoise_2d[(numslices - 1) * yincrement:numy, :],
+                                 rejection_threshold, nframes))
+    if yincrement * numslices != numy:  # last slice is a larger one
+        print("odd last slice ", yincrement, numy - (numslices - 1) * yincrement)
```
hbushouse: Shouldn't have a print statement in operational code. Also, I think that all of the boilerplate for creating and distributing slices to cores, etc. could use a little more commentary, for the sake of future developers.

jwst/jump/jump_step.py
```diff
@@ -15,6 +16,7 @@ class JumpStep(Step):
 
     spec = """
         rejection_threshold = float(default=4.0,min=0) # CR rejection threshold
+        maximum_cores = option('quarter', 'half', 'all', default=None) # max number of processes to create
```
hbushouse: Is flake8 going to complain about having only 1 space before the # to start an in-line comment? I think it likes 2.

jwst/jump/twopoint_difference.py
```diff
@@ -19,15 +19,23 @@
...
+def find_crs(indata):
 
     """
     Find CRs/Jumps in each integration within the input data array.
     The input data array is assumed to be in units of electrons, i.e. already
     multiplied by the gain. We also assume that the read noise is in units of
     electrons.
     """
-
+    if isinstance(indata, tuple):
```
hbushouse: What's the reason for treating the inputs as a single argument that's a tuple? Is that for the purpose of passing slices when using multiple cores?

jwst/jump/twopoint_difference.py
```diff
@@ -19,15 +19,23 @@
...
+        read_noise = indata[2]
+        rej_threshold = indata[3]
+        nframes = indata[4]
+        data = indata[0]
+        ingdq = indata[1]
```
hbushouse: Minor nit, but why not list these in the order in which they appear in the tuple (i.e. 0, 1, 2, 3, 4) for the sake of readability and following what's being passed where?

jwst/jump/jump_step.py
jdavies-st: I think this is OK since it is within a string.  I think `ConfigObj` doesn't care.

jwst/jump/twopoint_difference.py
jdavies-st: Fixed.  use `pool.starmap` instead of `pool.map` to get argument unpacking in the multiprocessing pool.

Issue #2656, opened 2018-10-03
Update MSA metadata file handling for new primary_source column


stscijgbot:
Issue [JP-436](https://jira.stsci.edu/browse/JP-436) was created by Howard Bushouse:

#2143 documented a problem uncovered a couple months ago in which APT/MPT updates started allowing for more than one catalog source to be assigned to a given MSA slitlet, which caused an error in the current assign_wcs software, because it could no longer determine which shutter in the slitlet should be considered the primary shutter. Discussions with APT/MPT developers have resulted in them agreeing to add a new column to the MSA meta data called "primary_source", which will have Y/N values to indicate which shutter contains the primary source within a given slitlet. This update will render #2143 moot, in that we no longer need to update the CAL software to report an error and skip processing for the slitlet, but rather use the primary_source column to process as normal.

hbushouse:
https://jira.stsci.edu/browse/JSDP-714 tracks the SDP work needed to copy this new info from the PPS MSA metadata table to a new column in the FITS MSA metadata file shutter table.

hbushouse:
PPS work to add this new data to the PPS MSA metdata table is being tracked in [JSOCINT-77](https://jira.stsci.edu/browse/JSOCINT-77) and is currently scheduled to be included PPS 14.6, which should be available mid-February 2019.

stscijgbot:
Comment by Howard Bushouse: Also see the original PR at [https://prsystem.stsci.edu/prsystem/servlet/prbrowse/pr.90689]

stscijgbot:
Comment by Howard Bushouse: An unrelated error that was seen in the jw10006 MSA metadata file produced by DMS back around March 2018 is the fact that there are many entries in the shutter table with SLITLET_ID=-1. This is due to a kludge put in place by SDP to handle entries in the PPS msa tables that have a NULL entry for slitled_id. The following discussion about this issue took place at the JWST S&OC Test Data Curation meeting on 31-May-2018 (see [https://innerspace.stsci.edu/display/jwstdms/2018+-+DMS+Test+Data+Curation+Meeting+Minutes#id-2018-DMSTestDataCurationMeetingMinutes-.14]):

There are NULL values for a number of slitlet_ids.
 * A NULL value in this field indicates that the shutter is not part of a slitlet grouping.
 * We only expect NULL values for slitlets that are composed of a single shutter, for background.
 * We would NOT expect to see NULL values for this dataset (because of the way it is set up).

[...]

PR #3448, opened 2019-05-09
 add dither_point_index and primary_source columns to msa meta file 

### LINKED ISSUE #2655: Update assign_wcs to use new NIRSpec MSA dither_point_index meta data
Issue [JP-435](https://jira.stsci.edu/browse/JP-435) was created by Howard Bushouse:

The APT/MPT handling of nodded NIRSpec MSA exposures now treats all nods as the same exposure, which in turn means they all get assigned the same msa_metadata_id value. This makes it impossible for CAL to know which set of shutter/source configuration info contained in the FITS MSA metadata file should be used when calibrating a given exposure. However, a new meta data value "dither_point_index" has been added to the PPS msametadata table that does have a unique value for each nodded exposure. So the assign_wcs step needs to be updated to match both msa_metadata_id AND dither_point_index to each exposure.

nden:
This PR implements

- Adds 2 new columns to the MSA meta file: `dither_point_index` and `primary_source`
  - Both `dither_point_index` and `metadata_id` are used to match rows into the MSA meta file. 
  - Raises an `MSAFileError` if `model.meta.dither.position_number` or 
`model.meta.instrument.msa_metadata_id` is None
  - Rows with `slitlet_id == -1` are ignored
- Adds a field `dither_point` to the `Slit` structure.
- Adds an attribute `dither_point` to each slit in a `MultiSlitModel` for MOS observations
- `MissingMSAFileError` was renamed to `MSAFileError`
- `MSAFileError` is raised if there are multiple sources in the `primary_source` column for a specific `metadata_id` and `dither_point`.
- The ``msaflagging`` step was changed to use the updated ``Slit`` structure.

Resolves #2655 #2656


CHANGES.rst
```diff
@@ -160,6 +181,13 @@ srctype
 
 - Updated logic for background targets and nodded exposures. [#3310]
 
+
+transforms
+----------
+
+- A field ``dither_point`` to the ``Slit`` structure. [#3448]
```
philhodge: This is missing something like "was added" before "to the".

jwst/assign_wcs/nirspec.py
```diff
@@ -444,7 +453,9 @@ def get_open_msa_slits(msa_file, msa_metadata_id, slit_y_range=[-.55, .55]):
         ('background', 'S1'),
         ('shutter_state', 'S6'),
         ('estimated_source_in_shutter_x', '>f4'),
-        ('estimated_source_in_shutter_y', '>f4')])
+        ('estimated_source_in_shutter_y', '>f4')
```
philhodge: This probably should have a trailing comma, though I realize it's just a doc string.

jwst/assign_wcs/nirspec.py
```diff
@@ -453,10 +464,14 @@ def get_open_msa_slits(msa_file, msa_metadata_id, slit_y_range=[-.55, .55]):
...
-        msa_file : str
-            MSA configuration file name, FITS keyword MSACONFL.
-        msa_metadata_id : int
-            The MSA meta id for the science file, FITS keyword MSAMETID.
+    msa_file : str
+        MSA configuration file name, FITS keyword ``MSAMETFL``.
+    msa_metadata_id : int
+        The MSA meta id for the science file, FITS keyword ``MSAMETID``.
+    dither_position : int
+        The index in the dither patern, FITS keyword ``PATT_NUM``.
```
philhodge: "patern" --> "pattern"

jwst/assign_wcs/nirspec.py
```diff
@@ -472,50 +487,62 @@ def get_open_msa_slits(msa_file, msa_metadata_id, slit_y_range=[-.55, .55]):
     # of the configuration file.
     try:
         msa_file = fits.open(msa_file)
-    except:
-        message = "Unable to open MSA FITS file (MSAMETFL) {0}".format(msa_file)
+    except FileNotFoundError:
+        message = "Mising MSA meta (MSAMETFL) file {}".format(msa_file)
```
philhodge: "Mising" --> "Missing"

jwst/assign_wcs/nirspec.py
```diff
@@ -472,50 +487,62 @@ def get_open_msa_slits(msa_file, msa_metadata_id, slit_y_range=[-.55, .55]):
...
 
-        # Count the number of backgrounds that have an 'N' (meaning main shutter)
-        # This needs to be 0 or 1 and we will have to deal with those differently
-        # See: https://github.com/STScI-JWST/jwst/commit/7588668b44b77486cdafb35f7e2eb2dcfa7d1b63#commitcomment-18987564
-
-        n_main_shutter = len([s for s in slitlets_sid if s['background'] == 'N'])
+        n_main_shutter = len([s for s in slitlets_sid if s['primary_source'] == 'Y'])
 
         # In the next part we need to calculate, find, determine 5 things:
         #    quadrant,  xcen, ycen,  ymin, max
```
philhodge: The last word should be "ymax".

CHANGES.rst
```diff
@@ -90,6 +99,11 @@ extract_1d
 - The keywords that describe the units for the FLUX and ERROR columns
   have been corrected; the units are now specified as "Jy". [#3447]
 
+exrtact_2d
```
hbushouse: typo: should be "extract_2d"

CHANGES.rst
```diff
@@ -90,6 +99,11 @@ extract_1d
...
+----------
+
+- An ``attribute dither_point`` was added to each slit in a ``MultiSlitModel``
```
hbushouse: typo: only "dither_point" needs to be quoted, not attribute (right?)

jwst/assign_wcs/nirspec.py
```diff
@@ -444,7 +453,9 @@ def get_open_msa_slits(msa_file, msa_metadata_id, slit_y_range=[-.55, .55]):
         ('background', 'S1'),
         ('shutter_state', 'S6'),
         ('estimated_source_in_shutter_x', '>f4'),
-        ('estimated_source_in_shutter_y', '>f4')])
+        ('estimated_source_in_shutter_y', '>f4'),
+        ('dither_point_index', '>i2'),
+        ('primary_source', 'S1')
 
     For example, something like:
         (12, 2, 4, 251, 22, 1, 'Y', 'OPEN', nan, nan),
```
hbushouse: Should probably expand the example here to include the 2 additional arguments on the end

jwst/assign_wcs/nirspec.py
```diff
@@ -453,10 +464,14 @@ def get_open_msa_slits(msa_file, msa_metadata_id, slit_y_range=[-.55, .55]):
 
     Parameters
     ----------
-        msa_file : str
-            MSA configuration file name, FITS keyword MSACONFL.
-        msa_metadata_id : int
-            The MSA meta id for the science file, FITS keyword MSAMETID.
+    msa_file : str
+        MSA configuration file name, FITS keyword ``MSAMETFL``.
```
hbushouse: We might want to not use the term "configuration" here, because there is an actual MSA configuration file that gets uploaded to the spacecraft for setting the shutter states, and that's a different (simpler) kind of file (just a list of open/closed shutter indexes). We should refer to the FITS file we use as the "MSA meta data file", because it contains a lot more info than just the MSA shutter configuration.

jwst/assign_wcs/nirspec.py
```diff
@@ -472,50 +487,62 @@ def get_open_msa_slits(msa_file, msa_metadata_id, slit_y_range=[-.55, .55]):
...
+        message = "problem reading MSA metafile (MSAMETFL) {0}".format(msa_file)
+        log.error(message)
+        raise MSAFileError(message)
 
     # Get the configuration header from teh _msa.fits file.  The EXTNAME should be 'SHUTTER_INFO'
     msa_conf = msa_file[('SHUTTER_INFO', 1)]
     msa_source = msa_file[("SOURCE_INFO", 1)].data
 
     # First we are going to filter the msa_file data on the msa_metadata_id
     # as that is all we are interested in for this function.
```
hbushouse: Update the comment here so that it's clear we're now filtering on both msa_metadata_id AND dither_point_index

jwst/assign_wcs/nirspec.py
```diff
@@ -472,50 +487,62 @@ def get_open_msa_slits(msa_file, msa_metadata_id, slit_y_range=[-.55, .55]):
...
 
     # First thing to do is to get the unique slitlet_ids
     slitlet_ids_unique = list(set([x['slitlet_id'] for x in msa_data]))
 
+    # SDP may assign a value of "-1" tp ``slitlet_id`` - these need to be ignored.
+    # JP-436
+    if -1 in slitlet_ids_unique:
+        slitlet_ids_unique.remove(-1)
+
+    margin = 0.05
```
hbushouse: What's ``margin`` used for? Might be worth an in-line comment.

jwst/assign_wcs/nirspec.py
```diff
@@ -539,15 +566,25 @@ def get_open_msa_slits(msa_file, msa_metadata_id, slit_y_range=[-.55, .55]):
 
         # Not allowed....
         else:
-            raise ValueError("MSA configuration file has more than 1 shutter with "
-                             "sources for metadata_id = {}".format(msa_metadata_id))
+            message = ("MSA configuration file has more than 1 shutter with "
+                       "sources for metadata_id = {}".format(msa_metadata_id))
+            log.info(message)
+            raise MSAFileError(message)
```
hbushouse: I wonder if we should just report this as an error/warning for the given slitlet in which it occurs and remove that slitlet from the list, but then proceed with the rest of processing (rather than terminating with an exception).

jwst/assign_wcs/nirspec.py
```diff
@@ -539,15 +566,25 @@ def get_open_msa_slits(msa_file, msa_metadata_id, slit_y_range=[-.55, .55]):
...
-        source_name, source_alias, stellarity = [
-            (s['source_name'], s['alias'], s['stellarity']) \
-            for s in msa_source if s['source_id'] == source_id][0]
+        try:
+            source_name, source_alias, stellarity = [
+                (s['source_name'], s['alias'], s['stellarity']) \
+                for s in msa_source if s['source_id'] == source_id][0]
+        except IndexError:
+            # all background shutters
+            log.info("All background shutters, slitlet_id is {}".format(slitlet_id))
```
hbushouse: Maybe instead of the Yoda-like backwards talking we could say "Slitlet_id n contains all background shutters"

jwst/assign_wcs/nirspec.py
nden: It was originally called "configuration", need to keep up with the name changes.

jwst/assign_wcs/nirspec.py
nden: I considered this but decided against since it will hide errors in APT or SDP as this should never happen. However, I'm happy to change it if you think that's what needs to happen.

hbushouse: We can leave it as-is for now, and if we get complaints, it can be changed in the future.

PR #3449, opened 2019-05-09
JP 677 dsd

ddavis-stsci:
Update the association code to use AsnPool column name “PATT_NUM”, instead of the current DITHPTIN name. The code change was tested using the pools jw00623_pool.csv, jw00626_pool.csv, and pool_028_mir_lrsfs_nods.csv and the results matches that of the current code.


jwst/associations/lib/rules_level2b.py
```diff
@@ -356,8 +356,8 @@ def __init__(self, *args, **kwargs):
                     Constraint(
                         [
                             DMSAttrConstraint(
-                                name='dithptin',
-                                sources=['dithptin'],
+                                name='',
```
hbushouse: The previous version had the string "dithptin" populating both "name" and "sources", but now "name" is blank. Is that correct and/or desired?

nden:
There are conflict snow in `.gitigonre`. @ddavis-stsci Can you resolve these?

jdavies-st:
I think all the editor-specific `.gitignore` entries should not live in the `jwst` project.  Instead, developers using these editors that leave behind lots of files should put these entries into their own `~/.gitignore_global`.  So feel free to peel out all those changes in this PR and put them in your global ones.

PR #3461, opened 2019-05-13
Mask non-imaging regions in tweakreg and source_catalog

### LINKED ISSUE #2062: tweakreg_catalog DQ rejection
INS B7 testing of the `tweakreg_catalog` step on MIRI images resulted in several issues of concern, which are documented in https://jira.stsci.edu/browse/JP-164.
```
During Build 7 testing the MIRI team found that the Tweakreg Catalog does not work for MIRI data.

This information is in the MIRI Test report:
"The low-level structure in MIRI images may be affecting the source threshold determination. 
SSB is still investigating the details here, but they did get it to work manually. 
Second, the sources in this image are extremely sharp. The DAOfind algorithm rejects stars based
on a roundness and sharpness criteria and their default values are rejecting MIRI PSF-like objects.
To fix this, SSB would need to add instrument-dependent parameters for this step in pipeline.
The roundness and sharpness parameters were not exposed to the user/pipeline in the initial
implementation, but by relaxing the sharpness SSB got good source detections ...
Finally, the initial implementation of tweakreg_catalog did not include the exclusion of pixels or
regions based on the DQ arrays. Currently the entire image is being used (i.e. not excluding the
left-hand third of the image). That will need to be fixed. The false sources on the detector images
should hopefully be (mostly) cleaned up by applying a DQ mask."
```

larrybradley:
This PR adds masks for non-imaging regions in the `tweakreg` and `source_catalog` steps based on the `NON_SCIENCE` DQ flag (e.g. MIRI images).

I updated the `photutils` version dependency because the `photutils.DAOStarFinder` `mask` keyword requires `photutils >= 0.6`.

Closes #2062 and #3287.

### DIFF
--- a/jwst/tweakreg/tweakreg_catalog.py
+++ b/jwst/tweakreg/tweakreg_catalog.py
@@ -2,7 +2,7 @@ from astropy.table import Table
 import numpy as np
 from photutils import detect_threshold, DAOStarFinder
 
-from ..datamodels import ImageModel
+from ..datamodels import dqflags, ImageModel
 
 
 def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
@@ -71,7 +71,11 @@ def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
                             sharplo=sharplo, sharphi=sharphi, roundlo=roundlo,
                             roundhi=roundhi, brightest=brightest,
                             peakmax=peakmax)
-    sources = daofind(model.data)
+
+    # Mask the non-imaging area (e.g. MIRI)
+    mask = (dqflags.pixel['NON_SCIENCE'] & model.dq).astype(np.bool)
+
+    sources = daofind(model.data, mask=mask)
 
     columns = ['id', 'xcentroid', 'ycentroid', 'flux']
     if sources:
--- a/setup.py
+++ b/setup.py
@@ -193,7 +193,7 @@ setup(
         'gwcs>=0.10',
         'jsonschema>=2.3,<=2.6',
         'numpy>=1.13',
-        'photutils>=0.4',
+        'photutils>=0.6',
         'scipy>=1.0',
         'spherical-geometry>=1.2',
         'stsci.image>=2.3',


jwst/source_catalog/source_catalog.py
```diff
@@ -84,6 +84,9 @@ def make_source_catalog(model, kernel_fwhm, kernel_xsize, kernel_ysize,
     if not isinstance(model, DrizProductModel):
         raise ValueError('The input model must be a DrizProductModel.')
 
+    # Mask the non-imaging area (e.g. MIRI)
+    mask = (model.dqflags.pixel['NON_SCIENCE'] & model.dq).astype(np.bool)
```
jdavies-st: There is no `dq` array in the `resample` output unfortunately.  And in fact, the resampled output should have already taken into account the `'NON_SCIENCE'` flagging and not resampled those pixels to the output.  I suspect that all the changes in this PR to `source_catalog` can be removed, and only those to `tweakreg` will be necessary.

hbushouse: The only time a DQ array would be present is if the image has not been resample (i.e. not a DrizProductModel). Could something like a weight or context value of 0 in the DrizProduct serve a similar purpose?

larrybradley:
Thanks, @jdavies-st.  I removed the mask from source catalog.

jwst/tweakreg/tweakreg_catalog.py
```diff
@@ -71,7 +71,11 @@ def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
                             sharplo=sharplo, sharphi=sharphi, roundlo=roundlo,
                             roundhi=roundhi, brightest=brightest,
                             peakmax=peakmax)
-    sources = daofind(model.data)
+
+    # Mask the non-imaging area (e.g. MIRI)
+    mask = (model.dqflags.pixel['NON_SCIENCE'] & model.dq).astype(np.bool)
```
jdavies-st: This is a bug.  The model itself does not have `dqflags`.  One can get the dict from `datamodels` though.
```
from jwst import datamodels

mask = (datamodels.dqflags.pixel['NON_SCIENCE'] & model.dq).astype(np.bool)
```

larrybradley: Fixed, thanks!

PR #3465, opened 2019-05-14
fix change log for 0.13.2

nden:


nden:
Thanks @hbushouse  & @philhodge

PR #3469, opened 2019-05-14
Update ASN rules to create standalone science ASN for dedicated bkg observations

ddavis-stsci:
Updated code to generate o003 spec3 associations with 12 members of the type
```
           "members": [
                {
                    "expname": "jw00623003001_03102_00002_mirifushort_cal.fits",
                    "exptype": "science",
                    "exposerr": "null",
                    "asn_candidate": "('o003', 'observation')"
```
running the unit tests show no unexpected errors.


hbushouse:
So this allows for the generation of an ASN for just the exposures from observation 3, which is an observation of a background target, right? So that it can be processed on its own as if it were a regular, standalone science observation, right? It looks like that's what's implied in the description, based on the fact that the resulting ASN contains 12 members, which are the _cal files from observation 3, all of which are labeled as exptype=science. So if that's what's happening, then it all looks good.

ddavis-stsci:
If I understand then that is correct. The o003 observations which are listed as BKGDTARG=T are
treated as science observations. The twelve files are placed in an association with the 
expname= ..._cal.fits and the exptype set to science.

jwst/associations/lib/dms_base.py
```diff
@@ -315,8 +319,9 @@ def get_exposure_type(self, item, default='science'):
                 except KeyError:
                     pass
                 else:
-                    result = special
-                    break
+                        if exp_type not in SPEC3_SCIENCE_EXP_TYPES:
```
stscieisenhamer: This will result in any exposure in SPEC3_SCIENCE_EXP_TYPES to never be considered as background. For rules that do want such exposures to be treated as intended, this code would prevent that.

jwst/associations/lib/rules_level3_base.py
```diff
@@ -802,13 +802,12 @@ def __init__(self, *args, **kwargs):
...
             work_over=ProcessList.EXISTING
         )
-
         # Put all constraints together.
         self.constraints = Constraint(
             [
                 Constraint_Base(),
                 DMSAttrConstraint(
-                    sources=['is_imprt', 'bkgdtarg'],
+                    sources=['is_imprt'],
```
stscieisenhamer: The effect of this change can be that backgrounds will end up as backgrounds for themselves. Suggested approach is to create another rule that processes such exposures separately.

hbushouse:
The desired effect here is that when a pool contains dedicated background exposures, which will get assigned an observation candidate id (e.g. "o003" in this case), as well as a background candidate id (e.g. "c1001" in this case), want we want the rules to produce is 2 spec3 ASN's:

One for the background candidate c1001, which will consist of both science (non-background) and background exposures.

One for the observation candidate o003, which will consist of just the background exposures assigned to candidate o003, but marked as "exptype=science" in the resulting spec3 ASN file.

We do NOT need an ASN for observation candidate o001, which contains only the science (non-background) exposures that also belong to candidate c1001. The c1001 (containing both science and background exposures) should take precedence over o001.

PR #3471, opened 2019-05-15
Fix pixeldq array transformation in refpix step

### LINKED ISSUE #3181: Bug in reference pixel correction transformation of PIXELDQ array
In do_fullframe_corrections and do_subarray_corrections of reference_pixels.py there are transformations of the detector coordinates from DMS to native and then back again. However the pixeldq array is not dealt with symmetrically by these two transformations because it is only changed by the first one. The result is that the pixeldq array is incorrect for every other group. I would think the best way to do the transformation is once per data cube, however it is coded as once per group. This needs to be checked and possibly corrected for every NIRDataset class. 
I am working with NIRISS data. For NIRISS, the revised code that works is:
class NIRISSDataset(NIRDataset):
    """For NIRISS data"""

    def DMS_to_detector(self, integration, group):
        #
        # NIRISS has a 180 degree rotation followed by a flip across the line
        # X=Y
        self.get_group(integration, group)
        self.group = np.swapaxes(self.group[::-1, ::-1], 0, 1)
        self.pixeldq = np.swapaxes(self.pixeldq[::-1, ::-1], 0, 1)

    def detector_to_DMS(self, integration, group):
        #
        # Just flip and rotate back
        self.group = np.swapaxes(self.group, 0, 1)[ ::-1, ::-1]
        self.pixeldq = np.swapaxes(self.pixeldq, 0, 1)[ ::-1, ::-1]
        self.restore_group(integration, group)

where in detector_to_DMS the line self.pixeldq = np.swapaxes(self.pixeldq, 0, 1)[ ::-1, ::-1] has been added to change the pixeldq array back to the original.

stscirij:
Fixes #3181

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -180,6 +180,12 @@ reffile_utils
 - Improved error messages when problems are encountered in extracting
   subarrays from reference files. [#3268]
 
+refpix
+------
+
+- Fixed a bug where the pixeldq array was being repeatedly transformed from
+  DMS to detector frame with each group processed [#3181]
+
 resample_spec
 -------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -185,12 +185,6 @@ reffile_utils
 - Improved error messages when problems are encountered in extracting
   subarrays from reference files. [#3268]
 
-refpix
-------
-
-- Fixed a bug where the pixeldq array was being repeatedly transformed from
-  DMS to detector frame with each group processed [#3471]
-
 resample_spec
 -------------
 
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -158,6 +158,12 @@ reffile_utils
 - Improved error messages when problems are encountered in extracting
   subarrays from reference files. [#3268]
 
+refpix
+------
+
+- Fixed a bug where the pixeldq array was being repeatedly transformed from
+  DMS to detector frame with each group processed [#3181]
```
nden: We are recording the PR number here, not the issue number.

nden: Ah, and rebase and move the entry to 0.13.3.

PR #3473, opened 2019-05-15
fix -1 px off for NRS2 coordinates

### LINKED ISSUE #2957: NIRSpec 1-pixel offset in assign_wcs for NRS2 detector data
Issue [JP-487](https://jira.stsci.edu/browse/JP-487) was created by James Muzerolle:

The WCS calculations for spectra on the NIRSpec NRS2 detector appear to have a pixel indexing offset.  I get excellent agreement with ESA trace WCS quantities for all modes if I shift the pipeline pixel location by +1 in both the x- and y-directions.  This shift is also apparent looking at the wavelength image in the extract_2d output for both MOS and FS exposures.  Note there is no such offset in any NRS1 spectra that have been tested.

nden:
Resolves #2957 

It's been a while since this ticket was created and I lost track of the test data.
It would be good to get another NRS2 dataset and the corresponding ESA traces to test this fix.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 0.13.3 (Unreleased)
 ===================
 
+- assign_wcs
+------------
+
+- Fix a one pixel off problem with the Nirspec NRS2 WCS transforms. [#3473]
+
+
 datamodels
 ----------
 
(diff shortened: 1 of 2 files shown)


nden:
Tests passing now

https://github.com/nden/nrs_testing/blob/master/msa/NRS2_MOS.ipynb

PR #3475, opened 2019-05-16
Update photom to handle pixels with no flux cal

hbushouse:
The photom step has been updated to handle pixels in spectroscopic mode exposures that have wavelengths outside the range of the wavelengths of the calibration reference data. Previously the calibration value for such pixels had been set to 1, by the numpy interpolation routine that's used to interpolate the 1-D reference data into the 2-D space of the exposure. That's been changed to set the conversion values to NaN, and the locations of the NaN's are given a calibration value of zero, so that the pixels end up with a value of zero in the output. That set of pixels is also flagged with DO_NOT_USE in the DQ array. Updated the docs to add this info.

Also fixed a minor bug in the ``save_area_info`` function where it computes the difference between the nominal pixel area values given in the PHOTOM and AREA reference files.

### DIFF
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -239,9 +239,8 @@ class DataSet():
                         self.input.err /= sens2d
                         self.input.var_poisson /= sens2d**2
                         self.input.var_rnoise /= sens2d**2
-                        if (self.input.var_flat is not None and
-                            np.size(self.input.var_flat) > 0):
-                                self.input.var_flat /= sens2d**2
+                        if (self.input.var_flat is not None and np.size(self.input.var_flat) > 0):
+                            self.input.var_flat /= sens2d**2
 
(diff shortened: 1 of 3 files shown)


jwst/photom/photom.py
```diff
@@ -239,9 +239,8 @@ def calc_nirspec(self, ftab, area_fname):
                         self.input.err /= sens2d
                         self.input.var_poisson /= sens2d**2
                         self.input.var_rnoise /= sens2d**2
-                        if (self.input.var_flat is not None and
-                            np.size(self.input.var_flat) > 0):
-                                self.input.var_flat /= sens2d**2
+                        if (self.input.var_flat is not None and np.size(self.input.var_flat) > 0):
```
philhodge: The outer parentheses are no longer needed.  I only included them so that the line could be broken without using a backslash.

PR #3489, opened 2019-05-17
Fix bug in latest photom update

hbushouse:
The updates in #3475 had a couple of flaws. First, image mode exposures failed, because the if-block containing the setting of ``no_cal`` does not get entered, so that it was undefined when used later. Second, the use of ``no_cal`` in setting flags in the DQ array only worked properly when the DQ array is 2-D (same shape as ``no_cal``), which means it crashed on exposures that use ``CubeModel`` where the DQ array is 3-D. The updates here address both of those problems. If any reviewers know of a more elegant (pythonic) way to do this, please make suggestions.

### DIFF
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -650,8 +650,9 @@ class DataSet():
         # Get the length of the relative response arrays in this row
         nelem = tabdata['nelem']
 
-        # If the relative response arrays have length > 0, load and include them in
-        # the flux conversion
+        # If the relative response arrays have length > 0,
+        # load and include them in the flux conversion
+        no_cal = None
         if nelem > 0:
             waves = tabdata['wavelength'][:nelem]
             relresps = tabdata['relresponse'][:nelem]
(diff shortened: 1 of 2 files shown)


hbushouse:
And in case anyone's wondering ... yes I've actually now tested the updates on the failing modes and they work.

jwst/photom/photom.py
```diff
@@ -700,7 +701,13 @@ def photom_io(self, tabdata, order=None):
                 slit.var_rnoise *= conversion**2
             if slit.var_flat is not None and np.size(slit.var_flat) > 0:
                 slit.var_flat *= conversion**2
-            slit.dq[no_cal] = np.bitwise_or(slit.dq[no_cal], dqflags.pixel['DO_NOT_USE'])
+            if no_cal is not None:
+                    if len(slit.dq.shape) == 3:
+                        slit.dq[:,no_cal] = np.bitwise_or(slit.dq[:,no_cal],
```
philhodge: I would suggest using this syntax, which will work for `dq.shape` of 2 or higher (assuming `no_cal` is 2-D):

    slit.dq[..., no_cal] = np.bitwise_or(slit.dq[..., no_cal], dqflags.pixel['DO_NOT_USE'])

jwst/photom/photom.py
```diff
@@ -712,7 +719,13 @@ def photom_io(self, tabdata, order=None):
                 self.input.var_rnoise *= conversion**2
             if self.input.var_flat is not None and np.size(self.input.var_flat) > 0:
                 self.input.var_flat *= conversion**2
-            self.input.dq[no_cal] = np.bitwise_or(self.input.dq[no_cal], dqflags.pixel['DO_NOT_USE'])
+            if no_cal is not None:
+                if len(self.input.dq.shape) == 3:
+                    self.input.dq[:,no_cal] = np.bitwise_or(self.input.dq[:,no_cal],
+                        dqflags.pixel['DO_NOT_USE'])
```
philhodge: This doesn't bother me, but if `flake8` complains, you could move the first argument down to align with this argument.

hbushouse:
@nden The lack of NRS **image** mode flux cal is intentional. The team has no intention of deriving flux cal for imaging. As to the rest of the code, yes it's been pieced together and added to over several years and good stand to be redesigned from scratch once all the new flux cal methods are in place.

jwst/photom/photom.py
hbushouse: Thanks for the tip. Updated in the last commit.

PR #3495, opened 2019-05-22
Fix find_crs() in jump step to ignore groups set as DO_NOT_USE

### LINKED ISSUE #3104: Jump step does not ignore the first frame for MIRI
Issue [JP-515](https://jira.stsci.edu/browse/JP-515) was created by Patrick Kavanaugh:

It appears that the jump step is not ignoring the first frame in a MIRI integration which it should if the firstframe step is run prior. If group 2 is the start of the MIRI ramp to be fit then it can't also have a jump flag. This can lead to the problem described in https://jira.stsci.edu/browse/JP-342.

Here is an example of a MIRI ramp with a jump flagged at group 2 despite firstframe step having been run:

!group2_jump.png|width=347,height=277!

jemorrison:
The MIRI ramps can have the first and last groups  set as DO_NOT_USE. These groups should be excluded from the jump detection step. This PR checks if the group dq contains DO_NOT_USE and does not use them in the jump detection.

Below are two plots (both containing 2 plots). This data is  simulated data provided by Patrick Kavanagh of MIRI data with 7 groups. The top plot is a pixel ramp of DN values vs group number and the bottom plot of each set is the corresponding group DQ flag. The main points to look at are red - uncalibrated data and green triangles the ramp data after all the step before the ramp_fit step.

The two plots below  are  plots  before the change in this PR - the first and last  groups are used in the jump detection causing the second groups to flagged as JUMP_DET (4)  and the last frame is JUMP_DET + DO_NOT_USE (5).

pixel_ramp_before

The two plots below are after the change in this PR. No groups are flagged as JUMP_DET, the first frame step correctly flags the first group as DO_NOT_USE and the last frame step correctly flags the last group as do not use. 
pixel_ramp_after

Resolves #3104


jdavies-st:
Could you add a small unit test to test out this new functionality?  You can model it after the existing unit tests for the step.  Maybe reproduce a simplified case as the one in your plots above?

Also CHANGES.rst.

jemorrison:
Yes I will add a unit test.

jwst/jump/tests/test_miri_first_lastframe.py
```diff
@@ -0,0 +1,82 @@
...
+    jump_group = result.groupdq[0,3,:,:]
+
+    do_not_use_array = np.zeros((ysize,xsize))
+    jump_array = np.zeros((ysize,xsize))
+    do_not_use_array[:,:] = dqflags.group['DO_NOT_USE']
+    jump_array[:,:] = dqflags.group['JUMP_DET']
+
+    assert np.array_equal(first_group, do_not_use_array)
+    assert np.array_equal(last_group, do_not_use_array)
+    assert np.array_equal(jump_group, jump_array)
```
jdavies-st: Since this PR only changes the `twopoint_difference.find_crs()` function, I would suggest putting the new test with the other 29 tests for this function already in the module `test_twopoint_difference.py`.  And instead of testing the whole step (with all the overhead of a step), just test the function, like the other tests.  And use the same fixture already defined.  I.e. something like:
```

def test_first_last_frame(setup_cube):
    ngroups = 7
    nframes = 1
    data, gdq, nframes, read_noise, rej_threshold = setup_cube(ngroups, readnoise=5)
    # if the first and last group are used in jump detection it would cause
    # jumps to be detected between group 0-1 and group 5-6
    data[0, 0, 100, 100] = 10000.0
    data[0, 1, 100, 100] = 30000.0
    data[0, 2, 100, 100] = 30020.0
    data[0, 3, 100, 100] = 40040.0  # jump
    data[0, 4, 100, 100] = 40060.0
    data[0, 5, 100, 100] = 40080.0
    data[0, 6, 100, 100] = 50000.0
    # set the first and last group to DO_NOT_USE
    gdq[0, 0, 100, 100] = dqflags.group['DO_NOT_USE']
    gdq[0, 6, 100, 100] = dqflags.group['DO_NOT_USE']
    find_crs(data, gdq, read_noise, rej_threshold, nframes)
    assert gdq[0, 0, 100, 100] == dqflags.group['DO_NOT_USE']
    assert gdq[0, 6, 100, 100] == dqflags.group['DO_NOT_USE']
    assert gdq[0, 3, 100, 100] == dqflags.group['JUMP_DET']
```

jwst/jump/tests/test_miri_first_lastframe.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # create the data and groupdq arrays
+    # create a JWST datamodel for MIRI data
+    im = MIRIRampModel((nints, ngroups, ysize, xsize))
+    im.meta.instrument.detector = 'MIRIMAGE'
+    im.meta.instrument.name = 'MIRI'
+    im.meta.subarray.name = 'FULL'
+    im.meta.subarray.xstart = 1
+    im.meta.subarray.ystart = 1
+    im.meta.subarray.xsize = 1032
```
hbushouse: Again, reduce size. Don't need a full frame to test a million pixels with identical values.

jwst/jump/tests/test_miri_first_lastframe.py
```diff
@@ -0,0 +1,82 @@
...
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+
+
+@pytest.fixture(scope='function')
+def science_image():
+    """Generate science image"""
+    # size of integration
+    nints = 1
+    ngroups = 7
+    xsize = 1032
```
hbushouse: Don't really need to be testing this many pixels that all have identical values (1 would be sufficient). I suggest reducing to something like xsize=25, ysize=25, at most.

jwst/jump/twopoint_difference.py
```diff
@@ -42,6 +42,12 @@ def find_crs(data, gdq, read_noise, rej_threshold, nframes):
     wh_sat = np.where(np.bitwise_and(gdq, dqflags.group['SATURATED']))
     data[wh_sat] = np.NaN
 
+    # Reset groups set to DO_NOT_USE  values in input data array to NaN, so they don't get
+    # used in any of the subsequent calculations. For MIRI data the first and last frame
+    # can be set to DO_NOT_USE>
```
hbushouse: Clean up typos in comment.

jwst/jump/tests/test_miri_first_lastframe.py
jemorrison: I have a question on the setup_cube. Does it not set up a 2048 X 2048 array of pixels.
Should be make the setup_cube be smaller if you don't want to run find_crs over all
the pixels but a subset ?

hbushouse: Until earlier today it did setup 2048x2048 images. Now they're only 204x204.

jemorrison: I re-did the test by putting it in the existing test_twopoint_difference.py
I have the code that James has above. I added readnoise = 25 (because MIRI has a read noise ~ 25 electrons) 
I ran the test and the jump that I inserted just as a secondarytest (between groups 3 and 4) is not detected. Hum ?  I printed out the read_noise array - it is 25, the rej_threshold is 3.
Should I push this over to see if you can see anything wrong ?

jemorrison: do I have the read noise in the wrong units ? I don't see the gain in this test

jemorrison: changed the read_noise to 5 still no jump detection

jemorrison:
Ok I pushed it over to see if @hbushouse or @jdavies-st can see why the jump is not being detected.
I will remove the comments I just them there for now.
This test passes with the assert removed looking for the JUMP_DET

PR #3497, opened 2019-05-22
Truncate negative index to zero

### LINKED ISSUE #3496: Slice with negative index in extract_1d
Issue [JP-704](https://jira.stsci.edu/browse/JP-704) was created by Philip Hodge:

If a nod/dither offset is applied, one or the other of the source extraction limits can then extend beyond the image or 2-D cutout. If one of the shifted limits is negative, it should be set to zero. There are places where the limits (rounded to int) are used as a slice for a numpy array (e.g. when computing wavelengths), and a negative value is misinterpreted as being relative to the upper limit of an array axis (like a Python list index), but in this case it should be truncated to zero.

philhodge:
The lower limit for an extraction region was negative, due to a nod/dither offset, and this is now reset to zero.

Closes #3496.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,11 +2,6 @@
 ===================
 
 
-combine_1d
-----------
-
-- Unit tests were added to combine_1d.  [#3490]
-
 ami
 ---
 
@@ -21,6 +16,11 @@ assign_wcs
 
 - Fix a one pixel off problem with the Nirspec NRS2 WCS transforms. [#3473]
 
+combine_1d
+----------
+
+- Unit tests were added to combine_1d.  [#3490]
+
 datamodels
 ----------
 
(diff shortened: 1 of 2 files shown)


philhodge:
Actually, `ystart` and `ystop` are inclusive limits, so if they're both zero or both `shape[0]`, the extraction height will be one pixel.  There is a check in the function for finding the location of the spectrum from the target coordinates that the "found" location is within the bounding box; if it isn't, no nod/dither offset will be applied.  But a user could still override the reference file with one that specifies an extraction region that's outside the image.  In that case there are checks, and some of the specified limits will be reset to either the image size or the bounding box limits, but `extract_1d` should probably do something more drastic instead, such as raising an exception or populating the output arrays with NaNs.  There are also checks for the case that the limits were specified by polynomial functions, but if both limits are outside the image, the extraction width will just be zero.

Issue #2552, opened 2018-09-26
Enable flake8 for some steps


nden:
#2542 enables flake8 test in Travis excluding `associations`, `exp_to_source`, `fits_generator`, `lib`, `ramp_fitting`  and `jwpsf`. This issues is a reminder to fix those steps and complete #2329

- [ ] associations
- [x] exp_to_source
- [ ] fits_generator
- [x] lib
- [x] ramp_fitting
- [x] jwpsf

hbushouse:
`jwpsf` has been removed from the repo.

hbushouse:
As of 16-Jul-2019 the only 2 steps still being excluded are `associations` and `fits_generator`.

jdavies-st:
Those are still being excluded.  `fit_generator` has real bugs, i.e. code that doesn't work.  And I have not tried touching `associations`.

hbushouse:
OK, thanks. I'm leaving this open and setting to a future milestone.

braingram:
Fixed in #8652

PR #3513, opened 2019-05-23
Placate flake8 messages for #2552

dmggh:


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -462,8 +462,7 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
             if rhi > cubeshape[1]:
                 rhi = cubeshape[1]
 
-            # data_sect and gdq_sect are: [ groups, y, x ]
-            data_sect = model.get_section('data')[num_int, :, rlo:rhi, :]
```
hbushouse: So is ``data_sect`` not needed anymore? Was flake8 complaining because it was unused?

dmggh: Yes, flake8 was complaining: " F841 local variable 'data_sect' is assigned to but never used."

jdavies-st: Odd why this wasn't getting picked up via our `flake8` checker in Travis CI?  I.e. it doesn't show up as a problem on `master` now.

jdavies-st: Oh, I see.  Because it is being ignored.

https://github.com/spacetelescope/jwst/blob/3619272747ef5a24eea898e4fd828868f6de6f0e/setup.cfg#L18-L28

If you want to do `flake8` cleanup, I would suggest removing the line above from `setup.cfg` and then dealing with the ~170 problems produced by that.  =)

jdavies-st: If you remove the `exclude` for `ramp_fitting` in `setup.cfg`, then that will keep it fixed into the future as well.

Eventually, once all our excludes are gone, we can put a pre-commit hook to do a `flake8` check.

Issue #3458, opened 2019-05-13
exp_to_source should key off "source_id"


stscijgbot:
Issue [JP-689](https://jira.stsci.edu/browse/JP-689) was created by Nadia Dencheva:

exp_to_source should key off "source_id"

stscijgbot:
Comment by Jane Morrison: exp_to_source is using the slitid instead of the sourceid. I will modify the code and write a unit test.

stscijgbot:
Comment by Howard Bushouse: This change could be problematic, however, for the case of slitlets that contain only background. The source_id for such slits is set to 0 (zero) and there can be multiple instances of them. This will lead to multiple sets of slit data with source_id=0, which we want to keep separate from one another, not lump together. We'll need to figure out a workaround for this situation.

nden:
Do we need  unique `source_id` within an exposure or within a visit? 
The current situation where `Slit.name` is used does not ensure unique names for each background slitlet either. If it's sufficient to have unique names only  within one `MultiSlitModel` then we can make this happen within assign_wcs. We can assign a running counter to `Slit.source_id` in the case of background slitlets.

hbushouse:
At the very least they have to be distinct within an exposure. They have to be unique within an exposure, so that if there are multiple dedicated bkg slits, we keep the data for each (e.g. the 2-D cutouts from extract_2d) distinct within the resulting level-2b output files (_cal, _s2d, _x1d). Those distinct background instances then get used in the calspec3 master_background step, combining the multiple bkg x1d products into a master background spectrum. Beyond that point in calspec3 it would be very difficult for us to continue to keep them distinct, because if there are multiple bkg slits contained in multiple exposures, how do we match them up across exposures? I think at that point we have no choice but to have calspec3 combine the 2-D cutouts for all bkg instances in all exposures into a single bkg 2-D image and then produce a single 1-D bkg spectrum using extract_1d.

stscijgbot:
Comment by Jane Morrison: Summary of discussion with Howard and Nadia:
 For MSA the background subtraction is done on an exposure basis - not combining slit background from different exposures. The plan is to set the source_id = 0 for background in assign_wcs. If exp_to_source then
keys off source_id rather than slit_id and organizes the data by source_id rather than slit_id then resample_spec should work as expected.

[...]

PR #3517, opened 2019-05-23
spec3 pipeline for WFSS modes

### LINKED ISSUE #3415: Wide-Field Slitless Spectroscopy level-3 processing
Issue [JP-660](https://jira.stsci.edu/browse/JP-660) was created by Howard Bushouse:

At this time we have not received final specs for level-3 combination of NIRCam and NIRISS Wide-Field Slitless Spectroscopy (WFSS) data. There are incomplete ideas for either combining data from multiple exposures of a given source at the 2-D image cutout stage or combining the 2-D data during the process of 1-D extraction. Until those specifications are complete we will take the fallback approach of simply using the existing "combine_1d" step to combine the x1d products for a given source that are produced at the end of calwebb_spec2 processing.

A special calwebb_spec3 flow will be required that will need to make use of the existing "exp_to_source" step, to reorganize the per-exposure data that come out of calwebb_spec2 processing into per-source data models. After which the multiple 1-D spectra for a given source need to be combined.

This will also required new level-3 ASN rules for WFSS modes, in order to create ASN files that consist of the proper input member list (e.g. x1d products).

At a high level, the ASN rules and calwebb_spec3 processing should mimic the processing that's done for NIRSpec MOS observations, which have the same kind of level-2b organization (multiple exposures where each exposure consists of data for multiple sources), and get reorganized and processed on a per-source basis within calwebb_spec3. The number of calwebb_spec3 processing steps for WFSS data will just be greatly reduced in comparison to MOS (e.g. no resampling, etc.)

nden:
Change the calwebb_spec3 pipeline to allow processing of WFSS modes.

Also fixed a bug in extract_1d where the shape of the input data was not passed correctly to the extraction code in the case of a single 2D slit.

This was tested on one file from the regression tests - NIS-WFSS. More testing necessary.

Resolves #3415.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,10 +37,17 @@ extract_1d
 ----------
 - An indexing bug was fixed. [#3497]
 
+
 jump
 ----
 
 - Add multiprocessing capability to JumpStep [#3440]
+
+extract_2d
+----------
+
+- A space was removed from the names of grism objects in a MultiSlit model. [#3517]
+
 
 master_background
 -----------------
@@ -60,6 +67,11 @@ photom
 
 - Updated to zero-out pixels outside the wavelength range of flux calibration
   and set DQ=DO_NOT_USE. [#3475, #3489]
+
+pipeline
+--------
+
+- calwebb_spec3 was changed to allow processing of WFSS modes. [#3517]
 
 refpix
 ------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -46,7 +46,7 @@ jump
 extract_2d
 ----------
 
-- Replaced a white space in the names of grism objects with an upderscore. [#3517]
+- Replaced a white space in the names of grism objects with an underscore. [#3517]
 
 
 master_background
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -46,7 +46,7 @@ jump
 extract_2d
 ----------
 
-- A space was removed from the names of grism objects in a MultiSlit model. [#3517]
+- Replaced a white space in the names of grism objects with an upderscore. [#3517]
 
 
 master_background
--- a/jwst/extract_2d/grisms.py
+++ b/jwst/extract_2d/grisms.py
@@ -383,7 +383,7 @@ def extract_grism_objects(input_model,
                 # set x/ystart values relative to the image (screen) frame.
                 # The overall subarray offset is recorded in model.meta.subarray.
                 # nslit = obj.sid - 1  # catalog id starts at zero
-                new_slit.name = "Object {0}".format(obj.sid)
+                new_slit.name = "Object_{0}".format(obj.sid)
                 new_slit.xstart = 1  # fits pixels
                 new_slit.xsize = ext_data.shape[1]
                 new_slit.ystart = 1  # fits pixels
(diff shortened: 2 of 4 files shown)


nden:
The problem in #3458 is also seen in the results of  testing this PR. The output name contains the string from `slit.name` instead of `slit.source_id` which is supposed to be integer. The name of the output of spec3 for NIRISS WFSS for a specific test is `jw87600017001_02101_00002_nis_sobject 10_x1d.fits`. Note that there's also a space in the name. This is fixed in extract_2d.

philhodge:
Should there also be a call to `combine_1d_spectra` for each source?

nden:
@philhodge Yes, good point, I missed that part of the ticket.
Is this unique to WFSS data?

hbushouse:
Yes, the need to call ``combine_1d`` is only for WFSS data.

CHANGES.rst
```diff
@@ -33,6 +33,11 @@ extract_1d
 ----------
 - An indexing bug was fixed. [#3497]
 
+extract_2d
+----------
+
+- A space was removed from the names of grism objects in a MultiSlit model. [#3517]
```
hbushouse: Technically the white space was **replaced** by an underscore.

jwst/extract_2d/grisms.py
```diff
@@ -383,7 +383,7 @@ def extract_grism_objects(input_model,
                 # set x/ystart values relative to the image (screen) frame.
                 # The overall subarray offset is recorded in model.meta.subarray.
                 # nslit = obj.sid - 1  # catalog id starts at zero
-                new_slit.name = "Object {0}".format(obj.sid)
+                new_slit.name = "Object_{0}".format(obj.sid)
```
hbushouse: I wonder if we should make the name syntax here consistent with what we're doing for NRS MSA data, i.e. something that's just an integer without "Object_"?

jwst/pipeline/calwebb_spec3.py
```diff
@@ -20,6 +20,7 @@
 # Group exposure types
 MULTISOURCE_MODELS = ['MultiSlitModel']
 IFU_EXPTYPES = ['MIR_MRS', 'NRS_IFU']
+SLITLESS_TYPES = ['NIS_WFSS', 'NRC_WFSS'] #, 'NRC_TSGRISM']
```
hbushouse: NRC_TSGRISM is already comment out so it has no effect on anything, but we really could just completely remove it, because time-series grism data should never go through this pipeline anyway. They go through ``calwebb_tso3``.

jwst/extract_2d/grisms.py
nden: I agree. 👍 for less typing

nden:
I addressed all comments.
There's no data to test this code on. There's one NIRISS data file but the data looks like noise and although extract_1d does not crash the wavelengths it produces make combine_1d crash.

CHANGES.rst
```diff
@@ -37,11 +37,18 @@ extract_1d
...
+
 jump
 ----
 
 - Add multiprocessing capability to JumpStep [#3440]
 
+extract_2d
+----------
+
+- Replaced a white space in the names of grism objects with an upderscore. [#3517]
```
hbushouse: Typo: "u**p**derscore"

nden:
The failure on Jenkins is unrelated and due to a change in `photutils`. it is being fixed in a separate PR.

PR #3520, opened 2019-05-24
Update ami_analyze rotation param default

### LINKED ISSUE #3510: Fix default param values for ami_analyze step
Issue [JP-716](https://jira.stsci.edu/browse/JP-716) was created by Howard Bushouse:

The ami_analyze.cfg file sets default param values of:

oversample = 3

rotation = 1.49

but the "spec" block in the ami_analyze_step.py code has

oversample = 3

rotation = 0.0

The discrepancy in default rotation values needs to be fixed. I've asked [~thatte] what an appropriate default should be.

hbushouse:
Update the ami_analyze.cfg file to set the default value of ``rotation`` to zero, which matches the default value in the step "spec" block.

Fixes #3510.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,6 @@
 0.13.3 (Unreleased)
 ===================
 
-
 ami
 ---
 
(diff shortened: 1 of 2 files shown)


dmggh:
Looks good to me.

PR #3539, opened 2019-05-28
Don't use pixels with wavelength = NaN

### LINKED ISSUE #3223: Flag or exclude spectroscopic pixels with wavelength=NaN
Issue [JP-582](https://jira.stsci.edu/browse/JP-582) was created by Howard Bushouse:

JP-565 uncovered an issue where the extract_1d step seems to be including pixels that don't have wavelength values defined (set to NaN). We need to confirm whether this is the case and, if so, find a sensible way to exclude them from the column-by-column sums.

philhodge:
The `master_background/expand_to_2d.get_wavelengths` function is now used to get the 2-D array of wavelengths from the input file, instead of explicitly checking for the existence of the wavelength attribute (an extension in the input file), and evaluating the WCS function if the attribute does not exist or is not populated.

The replace_bad_values function has been modified.  Instead of setting the science data to a fill value (zero) for pixels where the data value was NaN or the data quality flag included DO_NOT_USE, the science data are now set to NaN wherever the wavelength is NaN or the DQ flag includes DO_NOT_USE.

Closes #3223.

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -11,6 +11,7 @@ from ..datamodels import dqflags
 from ..assign_wcs import niriss         # for specifying spectral order number
 from ..assign_wcs.util import wcs_bbox_from_shape
 from ..lib import pipe_utils
+from ..master_background.expand_to_2d import get_wavelengths
 from . import extract1d
 from . import ifu
 from . import spec_wcs
@@ -1957,6 +1958,9 @@ class ExtractModel(ExtractBase):
 
         # If the wavelength attribute exists and is populated, use it
         # in preference to the wavelengths returned by the wcs function.
+        # But since we're now calling get_wavelengths from
+        # master_background.expand_to_2d, wl_array should be populated,
+        # and we should be able to remove some of this code.
         if wl_array is None or len(wl_array) == 0:
             got_wavelength = False
         else:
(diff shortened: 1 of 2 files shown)


jwst/extract_1d/extract.py
```diff
@@ -11,6 +11,7 @@
 from ..assign_wcs import niriss         # for specifying spectral order number
 from ..assign_wcs.util import wcs_bbox_from_shape
 from ..lib import pipe_utils
+from ..master_background.expand_to_2d import get_wavelengths
```
hbushouse: At some point (after we deliver B7.3) we really need to move ``get_wavelengths`` to ``lib``, rather than always having to import from ``master_background``. But this is fine for now.

jwst/extract_1d/extract.py
```diff
@@ -2089,10 +2087,12 @@ def extract(self, data, wl_array, verbose):
         n_nan = nan_mask.sum(dtype=np.intp)
         if n_nan > 0:
             if verbose:
-                log.warning("%d NaNs in wavelength array", n_nan)
-            temp_wl[nan_mask] = 0.01            # because NaNs cause problems
+                log.debug("%d NaNs in wavelength array", n_nan)
```
jdavies-st: Not needed for this PR, but it would be nice to get rid of `verbose` and just do logging, allowing (eventually) the user to set the level, in order to get the log level one wants.  But we need to fix `stpipe` logging first.

philhodge: This is a little obscure (maybe more than a little), but the reason for a `verbose` argument in some of the functions in `extract_1d` is to allow logging info for just the first integration of a multi-integration exposure.  For TSO data, in particular, there can be zillions of integrations, and we don't need to log the same info for every one.  So `verbose` is set to True for the first integration, and then it's reset to False.

jdavies-st: Ah, that makes sense.  Maybe we should _add_ `verbose` to the `ami` code.  😂

Issue #3097, opened 2019-02-07
caldetector1 needs to allow for MIRIRampModel as input


hbushouse:
The `calwebb_detector1` pipeline module currently loads its input explicitly into a `RampModel`, without checking to see if the input data are from MIRI and hence should be loaded into a `MIRIRampModel`. That means MIRI-unique elements, like the REFOUT data extension, are not loaded into the data model. Fortunately at this time none of the MIRI-unique elements is used anywhere in the pipeline, but they should be available.

`calwebb_detector1` should be updated to check for MIRI as input and load into a `MIRIRampModel` in those cases.

stscirij:
The calwebb_detector1 pipeline can easily be fixed to load a MIRIRampModel, but the first step in the pipeline explicitly opens its input as a RampModel, so every step after this has a RampModel.
The most straightforward way of handling this is to add the refout attribute to the RampModel schema, and eventually do away with the MIRIRampModel (it's hardly ever used in the cal code).

jdavies-st:
Is there a reason we couldn't add the MIRI-specific data extensions to the normal `RampModel` schema and get rid of `MIRIRampModel`?  Everything in the schema is optional, so as long as `RampModel` is a superset of everything needed by MIRI and the other instruments, it would be fine I think.

hbushouse:
That's the approach that @stscirij and I discussed yesterday. If we can simply add the REFOUT extension to the generic ``RampModel`` and not have dummy or zero-filled versions of it appear in non-MIRI FITS files, then that's definitely the way to go. Back when the custom ``MIRIRampModel`` was created, it didn't seem possible to have extensions treated as optional.

jdavies-st:
As long as `refout` is not accessed in `RampModel.__init__()` or anytime else in the processing of MIRI data, then it won't appear with dummy zero values.

stscirij:
Yes, I just made the change to my local copy of datamodels, and adding the refout extension to the RampModel schema even with a default value of 0.0 doesn't create a refout array for non-MIRI data until it's accessed directly.  i.e. it doesn't appear in the .info() method, but as soon as you do non_miri_model.refout it appears.
I think for now we should keep the MIRIRampModel but update datamodels.open so that MIRI data return a RampModel.  The only place I really see much use of the MIRIRampModel is in the tests.

jemorrison:
[...]

PR #3546, opened 2019-05-29
MIRI data to use RampModel instead of MIRIRampModel

stscirij:
Add REFOUT attribute to RampModel so that MIRI data can be opened as a RampModel and retain the REFOUT data.  Change datamodels.open so that data with a REFOUT extension are opened as a RampModel instead of a MIRIRampModel.
MIRIRampModel is still available.

### DIFF
--- a/jwst/datamodels/schemas/ramp.schema.yaml
+++ b/jwst/datamodels/schemas/ramp.schema.yaml
@@ -34,6 +34,12 @@ allOf:
       default: 0.0
       ndim: 3
       datatype: float32
+    refout:
+      title: Reference Output
+      fits_hdu: REFOUT
+      default: 0.0
+      ndim: 4
+      datatype: float32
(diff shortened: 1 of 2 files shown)


jemorrison:
Looks good. When it is merged I will test it on some MIRI data with refout.

hbushouse:
Fixes #3097

nden:
It looks like `MiriRampModel` is not used any more. If so we should remove it.

Does this change need a log entry? `MiriRampModel` may be used by other packages or the MIRI team.

hbushouse:
Let's not remove ``MiriRampModel`` until the MIRI team (and anyone else) who's using it has had a chance to update their code. We've got a monthly MIRI software telecon coming up next week, so we'll alert them about this.

dmggh:
Unit tests for ramp fitting currently use MIRIRampModel.

PR #3556, opened 2019-05-30
move get_wavelengths to lib

nden:
Resolves #3553

This PR copies the `get_wavelength` function verbatim to `lib/wcs_utils.py` in order to fix an import error and allow tests to run. It may need to be cleaned and generalized in a later PR.

Issue #3553, opened 2019-05-30
ImportError: combine_1d_spectra


jhunkeler:
Importing `jwst` fails with:
```python
[...]/jwst-0.13.3a0.dev106+g9fe29906-py3.6.egg/jwst/master_background/master_background_step.py", line 5, in <module>
    from ..combine_1d.combine1d import combine_1d_spectra
ImportError: cannot import name 'combine_1d_spectra'
```

ref: https://plwishmaster.stsci.edu:8081/job/AstroConda/view/dev-problems-latest/job/dev_OSX-10.13_py3.6_np1.14/job/jwst/119/console

jdavies-st:
Hmm, this looks like it might be a circular import issue.  Shall we defer the import to the function that needs it?

It might be better to move the problematic code to `jwst/lib` where both packages can access it?

PR #3561, opened 2019-05-31
Add test to make sure submodules are importable

jdavies-st:
Make sure error in #3553 does not happen again.  Putting the check that packages are importable in this repo, not just in the conda package recipe.

Companion to the actual fix #3556.

Also, since `steps.py` is just used as in the conda recipe as a smoke test, and since we are now doing it here as part of our unit tests, can we remove the module and that line in the conda recipe?

https://github.com/astroconda/astroconda-dev/blob/aaad924953d4e608afda22cba7b90f6e6de32a46/jwst/meta.yaml#L76

I would feel more comfortable doing the smoke test in the package instead of in the conda recipe.  @jhunkeler 

Also, it seems the only thing that imports `steps.py` is `cal_ver_steps.py`.  But it seems to do nothing with it (but has a `flake8` override).  And maybe we don't need `cal_ver_steps.py` either?  We are not using this at all in our deployments, are we @hbushouse?

### DIFF
--- a/jwst/steps.py
+++ b/jwst/steps.py
@@ -1,47 +1,10 @@
-from . import datamodels
-from . import ami
-from . import assign_wcs
-from . import associations
-from . import background
-from . import combine_1d
-from . import coron
-from . import csv_tools
-from . import cube_build
-from . import cube_skymatch
-from . import dark_current
-from . import dq_init
-from . import emission
-from . import extract_1d
-from . import extract_2d
-from . import fits_generator
-from . import flatfield
-from . import fringe
-from . import guider_cds
-from . import imprint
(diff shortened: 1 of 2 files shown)


hbushouse:
I don't think ``cal_ver_steps`` is being used anymore. When it was in use, it generated a file of version numbers per step, which was then ingested as a reference file in CRDS (I think). @jaytmiller can you confirm whether or not this is still needed?

jdavies-st:
We'll how about we merge this now, and I can remove the other modules in a separate PR if needed.  After `import jwst.steps` is removed from the astroconda-dev `jwst` recipe.

PR #3563, opened 2019-05-31
JP-741 Customize reffile prefetch

### LINKED ISSUE #3549: Level-3 pipelines need to skip TA images in their ASN files
Issue [JP-741](https://jira.stsci.edu/browse/JP-741) was created by Howard Bushouse:

[~mswam] reports that DMS processing has encountered instances where the level-3 pipelines are attempting to include Target Acq exposures that are listed in the input ASN files. Any non-science member needs to be excluded from processing.

Example from calwebb_coron3 processing:

/ifs/archive/test/jwst/test1/info/owl/logs/rspencer_jw00628-c1015_20190516t191939_coron3_023_1558047879.121911/ALOG_1558048006_level_3_jw00628-c1015_20190516t191939_coron3_023.err

CRDS - DEBUG -  Bestrefs header:

{'META.EXPOSURE.READPATT [READPATT]': 'FAST',

'META.EXPOSURE.TYPE [EXP_TYPE]': 'MIR_TACQ',

'META.INSTRUMENT.BAND [BAND]': 'UNDEFINED',

'META.INSTRUMENT.CHANNEL [CHANNEL]': 'UNDEFINED',

'META.INSTRUMENT.CORONAGRAPH [CORONMSK]': 'UNDEFINED',

'META.INSTRUMENT.DETECTOR [DETECTOR]': 'MIRIMAGE',

'META.INSTRUMENT.FILTER [FILTER]': 'FND',

'META.INSTRUMENT.NAME [INSTRUME]': 'MIRI',

'META.OBSERVATION.DATE [DATE-OBS]': '2019-04-02',

'META.OBSERVATION.TIME [TIME-OBS]': '09:53:35.788',

'META.SUBARRAY.NAME [SUBARRAY]': 'MASK1550',

'META.VISIT.TSOVISIT [TSOVISIT]': 'F',

'REFTYPE': 'UNDEFINED'}

CRDS - DEBUG -  Reference type 'drizpars' defined as 'jwst_miri_drizpars_0001.fits'

CRDS - ERROR -  Error determining best reference for 'psfmask'  =    parameter='META.INSTRUMENT.FILTER [FILTER]' value='FND' is not in ['F1065C', 'F1140C', 'F1550C', 'F2300C']

2019-05-16 23:06:54,306 - stpipe.Coron3Pipeline - ERROR -  Error determining best reference for 'psfmask'  =    parameter='META.INSTRUMENT.FILTER [FILTER]' value='FND' is not in ['F1065C', 'F1140C', 'F1550C', 'F2300C']

jaytmiller:
Some members of a coron3 association should not be processed / prefetched.

Modified step.py to expose a `.prefetch()` method which is called just prior to
`.process()` IF `.prefetch_references == True`.   This is the default.

If `.prefetch_references == False`,  `.prefetch()` can still be called explicitly
within `.process()` if needed,  as is done here for `calwebb_coron3`.

To improve customization, changed `Pipeline.reference_file_types`, once defined
in __init__() as an attribute, into a `@property` which is deferred until needed.

Resolves #3549.

### DIFF
--- a/jwst/pipeline/calwebb_coron3.py
+++ b/jwst/pipeline/calwebb_coron3.py
@@ -42,6 +42,8 @@ class Coron3Pipeline(Pipeline):
         'resample': resample_step.ResampleStep
     }
 
+    prefetch_references = False
+
     def process(self, input):
         """Primary method for performing pipeline."""
         self.log.info('Starting calwebb_coron3 ...')
@@ -72,6 +74,9 @@ class Coron3Pipeline(Pipeline):
             self.log.error(err_str1)
             self.log.error('Calwebb_coron3 processing will be aborted')
             return
+
+        for member in psf_files + targ_files:
+            self.prefetch(member)
 
         # Assemble all the input psf files into a single ModelContainer
         psf_models = datamodels.ModelContainer()
--- a/jwst/stpipe/pipeline.py
+++ b/jwst/stpipe/pipeline.py
@@ -69,12 +69,11 @@ class Pipeline(Step):
 
             setattr(self, key, new_step)
 
-        self.reference_file_types = self._collect_active_reftypes()
-
-    def _collect_active_reftypes(self):
+    @property
+    def reference_file_types(self):
         """Collect the list of all reftypes for child Steps that are not skipped.
-        Overridden reftypes are included but handled normally later by the Pipeline
-        version of the get_ref_override() method defined below.
+        Overridden reftypes are included but handled normally later by the
+        Pipeline version of the get_ref_override() method defined below.
         """
         return [reftype for step in self._unskipped_steps
                 for reftype in step.reference_file_types]
--- a/jwst/stpipe/step.py
+++ b/jwst/stpipe/step.py
@@ -363,14 +363,6 @@ class Step():
             self.set_primary_input(args[0])
 
         try:
-            # prefetch truly occurs at the Pipeline (or subclass) level.
-            if (
-                    len(args) and len(self.reference_file_types) and
-                    not self.skip and
-                    self.prefetch_references
-            ):
-                self._precache_references(args[0])
-
             # Default output file configuration
             if self.output_file is not None:
                 self.save_results = True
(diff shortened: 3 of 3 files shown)


stscieisenhamer:
> On the other hand, each particular pipeline module may have specific needs and uses of the different kinds of input members in the ASN (e.g. some pipelines may not do any processing of "background" members, while others may). So perhaps it is appropriate for each pipeline module to specify which types need prefetching.

See previous comment about an `asn_prefetch_types` attribute. Pipelines can then define what types to prefetch.

jaytmiller:
Any thoughts on where the generalized code acting on asn_prefetch_types should be implemented?   In Pipeline the prefetch is iterating over a ModelContainer or DataModel.

The thing about the original PR is that it finally provides a pattern for dealing with customizing the prefetch during process():  it solves the perennial "prefetch happens to early" problem.  What it lacks in generality it makes up for with simplicity.

jdavies-st:
Just to add my 2¢ here, the problem of loading an association and having it split into various exptypes (SCIENCE, PSF, BACKGROUND, etc) some of which cause prefetching and some of which do not:

- is more adjacent to `datamodels` than to `stpipe`.  I.e. this is a general problem that both pipelines and steps need information about the data they are being asked to operate on. Currently there's a bunch of custom ways this is done in various pipelines or steps (most recently in `master_background`) where the association is parsed and loaded into combined or separate `ModelContainer` objects for each exptype.  It would be nice to have a single general function or method that does this.  Perhaps a rewrite of `ModelContainer`?  Perhaps something else?  A property that exposes the elements in of a requested exptype as an iterable in the `ModelContainer`?  Not sure.

- Agree `asn_prefetch_types` would be useful once it is clear how the association `exptype` labels apply to `ModelContainer` instances containing each type.  Currently there is no API for this.  But there could be.

I think the solution in this PR is a good compromise until the more general infrastructure issues outlined above are handled.

hbushouse:
Because this fix isn't crucial for 0.13.3 (there's a parallel fix being implemented via the CRDS rmap for the particular problem case) I'd rather not stick it into 0.13.3 at the last second. Let's wait for our next release candidate (0.13.4).

PR #3567, opened 2019-06-03
JP-326 Improve error message for AMI _calints input

### LINKED ISSUE #2138: calwebb_ami3 crashing when input files have suffix _calints.fits
Issue [JP-326](https://jira.stsci.edu/browse/JP-326) was created by Deepashri Thatte:

calwebb_ami3 crashing when input files have suffix _calints.fits

jdavies-st:
Resolves #2138.

It would be nice to have a more generalized Error class for `stpipe` steps and pipelines, but will require going through step-by-step and pipeline-by-pipeline and coming up with a consistent way of emitting errors by steps and handling them (if possible) in pipelines.  I will file an issue for this.

### DIFF
--- a/jwst/ami/ami_analyze_step.py
+++ b/jwst/ami/ami_analyze_step.py
@@ -38,31 +38,35 @@ class AmiAnalyzeStep(Step):
         self.log.info('Initial rotation guess = %g deg', rotate)
 
         # Open the input data model
-        with datamodels.ImageModel(input) as input_model:
+        try:
+            with datamodels.ImageModel(input) as input_model:
 
-            # Get the name of the filter throughput reference file to use
-            self.filter_ref_name = self.get_reference_file(input_model,
-                                   'throughput')
-            self.log.info('Using filter throughput reference file %s',
-                           self.filter_ref_name)
+                # Get the name of the filter throughput reference file to use
+                throughput_reffile = self.get_reference_file(input_model,
+                                       'throughput')
+                self.log.info('Using filter throughput reference file %s',
+                               throughput_reffile)
 
-            # Check for a valid reference file
-            if self.filter_ref_name == 'N/A':
-                self.log.warning('No THROUGHPUT reference file found')
-                self.log.warning('AMI analyze step will be skipped')
-                result = input_model.copy()
-                result.meta.cal_step.ami_analyze = 'SKIPPED'
-                return result
+                # Check for a valid reference file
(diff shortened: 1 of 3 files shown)


jwst/ami/ami_analyze_step.py
```diff
@@ -38,31 +38,36 @@ def process(self, input):
...
+                self.log.info('Using filter throughput reference file %s',
+                               self.filter_ref_name)
+
+                # Check for a valid reference file
+                if self.filter_ref_name == 'N/A':
+                    self.log.warning('No THROUGHPUT reference file found')
+                    self.log.warning('AMI analyze step will be skipped')
+                    result = input_model.copy()
+                    result.meta.cal_step.ami_analyze = 'SKIPPED'
+                    return result
```
hbushouse: Here's an instance of the step skipping and returning a copy of the input. Does this get caught and handled appropriately up in the ``calwebb_ami3`` pipeline module? The rest of the pipeline is impossible to execute without output from this step.

jdavies-st: Hmm, good question.  It doesn't, as the pipeline has no clue that the step has been skipped.

jdavies-st: I will fix that in this PR.

jdavies-st:
Let's hold off on merging this until the regression tests are fixed.  And it looks like there may need to be some code changes to fix those.

jdavies-st:
And it appears the error here is more subtle.  If I try to open a 3D `CubeModel` file (_calints) from disk as an `ImageModel`, then I get a `ValueError`.
```
ValueError: Array has wrong number of dimensions.  Expected 2, got 3
```
But if I pass an already-open `CubeModel` in memory and reopen it as an `ImageModel`, it will happily open it as a 3D `ImageModel`.
```
In [2]: model = datamodels.CubeModel((25, 19, 19))

In [3]: im = datamodels.ImageModel(model)

In [4]: im
Out[4]: <ImageModel(25, 19, 19)>
```

hbushouse:
Rather than relying on datamodels to throw an error, how about just an explicit check on the shape of model.data once it's been opened?

jdavies-st:
I think that makes sense.  It will have to be both.

But this is a deeper problem in `datamodels`.

jdavies-st:
I've added some unit tests to test all the changes made in this PR.  So I think it is fine to merge.

PR #3572, opened 2019-06-03
Update MultiExposure model to include all slit arrays

### LINKED ISSUE #3521: exp_to_source results in models with only SCI, ERR, DQ
Issue [JP-721](https://jira.stsci.edu/browse/JP-721) was created by Howard Bushouse:

When running the calwebb_spec3 pipeline on NIRSpec MSA exposures the calls to exp_to_source.multislit_to_container, which is used to reorganize data from multiple exposures, the resulting data models contain only the SCI, ERR, and DQ arrays from the original inputs, excluding various other spectroscopic-specific arrays like WAVELENGTH, BARSHADOW, PATHLOSS, and all the new VAR_xxxxxx arrays. It looks like this is due to multiexposure.schema.yaml specifying only the SCI, ERR, and DQ arrays.

hbushouse:
Updated the MultiExposure model to include all possible Slit arrays, not just SCI, DQ, and ERR. The lack of arrays like WAVELENGTH in the output was causing problems in downstream steps in ``calwebb_spec3``.

Fixes #3521.

### DIFF
--- a/jwst/datamodels/multiexposure.py
+++ b/jwst/datamodels/multiexposure.py
@@ -62,7 +62,13 @@ class MultiExposureModel(DataModel):
             self.exposures[0].data = init.data
             self.exposures[0].dq = init.dq
             self.exposures[0].err = init.err
+            self.exposures[0].wavelength = init.wavelength
+            self.exposures[0].barshadow = init.barshadow
             self.exposures[0].area = init.area
+            self.exposures[0].var_poisson = init.var_poisson
+            self.exposures[0].var_rnoise = init.var_rnoise
+            self.exposures[0].var_flat = init.var_flat
+            self.exposures[0].pathloss = init.pathloss
             return
 
         super(MultiExposureModel, self).__init__(
--- a/jwst/datamodels/schemas/multiexposure.schema.yaml
+++ b/jwst/datamodels/schemas/multiexposure.schema.yaml
@@ -26,9 +26,27 @@ allOf:
               fits_hdu: ERR
               default: 0.0
               datatype: float32
+            wavelength:
+              fits_hdu: WAVELENGTH
+              datatype: float32
+            barshadow:
+              fits_hdu: BARSHADOW
+              datatype: float32
             area:
               fits_hdu: AREA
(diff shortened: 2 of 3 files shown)


jwst/datamodels/multiexposure.py
```diff
@@ -62,7 +62,13 @@ def __init__(self, init=None, **kwargs):
             self.exposures[0].data = init.data
             self.exposures[0].dq = init.dq
             self.exposures[0].err = init.err
+            self.exposures[0].wavelength = init.wavelength
+            self.exposures[0].barshadow = init.barshadow
             self.exposures[0].area = init.area
+            self.exposures[0].var_poisson = init.var_poisson
+            self.exposures[0].var_rnoise = init.var_rnoise
+            self.exposures[0].var_flat = init.var_flat
+            self.exposures[0].pathloss = init.pathloss
```
jdavies-st: What happens if the input `MultiSlitModel` or `SlitDataModel` does not have one of these data arrays?  The new `MultiExposureModel` now will, but with the default values populated.

hbushouse: The results of 1 test (so far) for inputs that do not have BARSHADOW or PATHLOSS arrays shows that they don't show up in the outputs. Only things that actually exist in the input get copied.

jdavies-st: Ah good!  Whew.

jwst/datamodels/schemas/multiexposure.schema.yaml
```diff
@@ -26,10 +26,28 @@ allOf:
...
+              datatype: float32
+            var_rnoise:
+              fits_hdu: VAR_RNOISE
+              datatype: float32
+            var_flat:
+              fits_hdu: VAR_FLAT
+              datatype: float32
+            pathloss:
+              fits_hdu: PATHLOSS
+              datatype: float32
```
jdavies-st: I would point to `variance.schema.yaml` here instead of listing them separately.  That said, in the variance schema, there are defaults defined.  Might want to double-check the behavior.

Issue #3516, opened 2019-05-23
Consider stopping nightly builds of conda-dev


stscijgbot:
Issue [JP-718](https://jira.stsci.edu/browse/JP-718) was created by Nadia Dencheva:

Creating nightly packages with conda-dev has become a big burden. Before we can stop building them we need to create a reliable way to provide access to the jwst pipeline code using pip and the repository. This involves providing instructions of
 * how to migrate from an existing conda-dev installation to a conda environment using pip installations of jwst
 * how to update on demand from the repository using pip (including jwst dependencies)
 * stop using conda-dev in our Jenkins testing and start using `pip`.

stscijgbot:
Comment by James Davies: Now conda-dev is not being used on any of our CI or regression testing.

stscijgbot:
Comment by Nadia Dencheva: conda-dev is not used in testing now. Instructions are being written.

PR #3580, opened 2019-06-04
JP-718 Remove astroconda-dev from our Jenkins CI/RT builds

jdavies-st:
Addresses part of #3516.


jdavies-st:
The regression test results on this  PR branch

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/116/

match those of master

https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/268/

So I think we are getting consistent results using `pip install` of all upstream dependencies.  I was especially concerned about the different builds of `numpy` between Pypi and Anaconda, but it seems to be good for now.

Jenkinsfile
```diff
@@ -16,14 +16,13 @@ def pip_install_args = "--index-url ${pip_index} --progress-bar=off"
...
+bc0.conda_ver = '4.6.14'
 bc0.conda_packages = [
     "python=${python_version}",
-    "pip<19.1"
 ]
 bc0.build_cmds = [
-    "pip install ${pip_install_args} numpy",
-    "pip wheel ${pip_install_args} .",
+    "pip install numpy",
+    "pip wheel .",
```
hbushouse: Just for my own education, what is "pip wheel"?

jdavies-st: A `wheel` is the modern binary python package distribution format. It replaced `egg` a while back.  `.whl`.  This makes sure that our package and its dependencies can be built into a wheel.  I.e. it makes sure that we can distribute the package to others.

jdavies-st: In your own checkout of the repo you can
```
python setup.py sdist bdist_wheel
```
and it will build the `sdist` (source distribution) and `bdist_wheel` (binary wheel distribution) in the `build` directory.  These are the things that get uploaded to pypi.org in oder to do a proper release when the day comes.

The command
```
pip wheel .
```
builds not only the wheel for `jwst`, but also the wheels for all the dependencies.

Did I get all that right @jhunkeler?

hbushouse: Hmmm, so if it used to be called "egg", why didn't they name the new one "chicken", instead of "wheel"?

jdavies-st: 🐍 🥚

jdavies-st:
Hmm, looks like there's a very old version of SQLite on our `python3.7` nodetype.  I'll remove it from our matrix for now.

Adding numpy to our `requirements-dev.txt` for dev builds.

PR #3581, opened 2019-06-04
Fix base url handling for access to artifactory

### LINKED ISSUE #3551: artifactory file globbing failing when running on jenkins
Issue [JP-743](https://jira.stsci.edu/browse/JP-743) was created by Jonathan Eisenhamer:

jenkins regression tests, using artifactory globbing, started not working/failing with jwst [pr#3443|https://github.com/spacetelescope/jwst/pull/3443/], which fixed previous artifactory globbing issues.

It appears to be only when running under jenkins.  The related jenkins build is [#232|https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/232/].

stscieisenhamer:
Resolves #3551

### DIFF
--- a/jwst/tests/base_classes.py
+++ b/jwst/tests/base_classes.py
@@ -113,7 +113,6 @@ class BaseJWSTTest:
         # Get full path and proceed depending on whether
         # is a local path or URL.
         root = get_bigdata_root()
-        logger.debug('root = %s', root)
         if op.exists(root):
             path = op.join(root, *self.repo_path)
             root_len = len(path) + 1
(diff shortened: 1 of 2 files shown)


jwst/tests/base_classes.py
```diff
@@ -20,6 +21,10 @@
     'BaseJWSTTest',
 ]
 
+# Configure logging
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+
```
jdavies-st: Why are we logging in test classes?

jwst/tests/base_classes.py
```diff
@@ -1,4 +1,5 @@
 from glob import glob as _sys_glob
+import logging
```
jdavies-st: We should probably remove logging from this module.

stscieisenhamer: Left in from debugging the issue

PR #3584, opened 2019-06-04
JP-689: exp_to_source should key off "source_id"

### LINKED ISSUE #3458: exp_to_source should key off "source_id"

nden:
For NRS Fixed slits the source_id is defined as a fixed integer, essentially mapping it to a slit.

For MOS data a counter is used to assign a unique `source_id` to a background slit. If the value of the counter happens to be one of the real source_id in the exposure, it is shifted outside the range of all source ids by adding the highest source_id.

An enum was added to the `model.meta.instrument.fixed_slit` schema based on the keyword dictionary.

Resolves #3458.


jwst/assign_wcs/nirspec.py
```diff
@@ -381,14 +381,15 @@ def get_open_fixed_slits(input_model, slit_y_range=[-.55, .55]):
...
-    s2a1 = Slit('S200A1', 0, 0, 0, 0, ylow, yhigh, 5)
-    s2a2 = Slit('S200A2', 1, 0, 0, 0, ylow, yhigh, 5)
-    s4a1 = Slit('S400A1', 2, 0, 0, 0, ylow, yhigh, 5)
-    s16a1 = Slit('S1600A1', 3, 0, 0, 0, ylow, yhigh, 5)
-    s2b1 = Slit('S200B1', 4, 0, 0, 0, ylow, yhigh, 5)
+    s2a1 = Slit('S200A1', 0, 0, 0, 0, ylow, yhigh, 5, 1)
+    s2a2 = Slit('S200A2', 1, 0, 0, 0, ylow, yhigh, 5, 2)
+    s4a1 = Slit('S400A1', 2, 0, 0, 0, ylow, yhigh, 5, 3)
+    s16a1 = Slit('S1600A1', 3, 0, 0, 0, ylow, yhigh, 5, 4)
+    s2b1 = Slit('S200B1', 4, 0, 0, 0, ylow, yhigh, 5, 5)
```
hbushouse: What is the extra argument that's been added at the end of each of these calls? Are you just using the integer values 1-5 as the source_id? As opposed to using the string versions (S200A1, etc.) of the slit names?

jwst/assign_wcs/nirspec.py
```diff
@@ -509,14 +537,30 @@ def get_open_msa_slits(msa_file, msa_metadata_id, dither_position,
     msa_data = [x for x in msa_conf.data if x['msa_metadata_id'] == msa_metadata_id \
                 and x['dither_point_index'] == dither_position]
 
+    # Get all source_ids for slitlets with sources.
+    # These should not be used when assigning source_id to background slitlets.
+    source_ids = set([x[5] for x in msa_conf.data if x['msa_metadata_id'] == msa_metadata_id \
+                      and x['dither_point_index'] == dither_position])
+    # All BKG shutters get a source_id value of 0.
```
hbushouse: By "BKG shutters" do you mean the individual shutters of a larger slit that don't happen to have the source located in them in this particular exposure (e.g. the top and bottom shutters of a 1x3 slit when the source is in the middle shutter)? As opposed to shutters that explicitly designated as background only.

jwst/assign_wcs/nirspec.py
```diff
@@ -574,7 +620,7 @@ def get_open_msa_slits(msa_file, msa_metadata_id, dither_position,
 
         # subtract 1 because shutter numbers in the MSA reference file are 1-based.
         shutter_id = xcen + (ycen - 1) * 365
-        source_id = slitlets_sid[0]['source_id']
+        #source_id = slitlets_sid[0]['source_id']
```
hbushouse: Any reason to comment this out instead of removing the line?

jwst/assign_wcs/nirspec.py
nden: Yes, I am assigning integer values as source_id. Just to be consistent with MOS data.

jwst/assign_wcs/nirspec.py
nden: I meant that's what BKG shutters are assigned in the msa meta file. I'll clarify the comment.

jwst/assign_wcs/nirspec.py
nden: Yes, this is handled differently now for source and bkg slitlets and this is done in the two specific code blocks.

PR #3587, opened 2019-06-05
start change log for 0.13.4

nden:


CHANGES.rst
```diff
@@ -1,3 +1,6 @@
+0.13.4 (Unreleased)
+===================
+
 0.13.3 (2019-05-04)
```
hbushouse: D'oh! Just noticed now that the 0.13.3 release date is wrong. Should be 2019-0**6**-04.

nden: Is it June already? I'll fix it.

nden:
It will be fixed on "stable" with the next delivery.

PR #3598, opened 2019-06-05
JP-651 Update MIRI MRS straylight step to use CDP-7 REGIONS files

### LINKED ISSUE #3377: Update MIRI MRS straylight routine to use the slice mask
Issue [JP-651](https://jira.stsci.edu/browse/JP-651) was created by David Law:

The MRS straylight routine currently has two options coded to know how to find which pixels on the detector to use to calculate the straylight.
 * Option 1: uses 'STRAYMASK' type reference files to define the pixel locations.  This approach is extremely outdated (5 years old) and uses CDP-3 files that are deprecated.
 * Option 2: uses the distortion solution attached by assign_wcs to determine pixels that are not identified as being in a given slice.

The first option should be retired, as should the reference file type.  These reference files are degenerate with the REGIONS reference file and are not in use going forward.

The second option is also undesirable given the new updates to the slice mask.  While things such as cube_build should use an 80% throughput level (or something similarly high) to build cubes only from pixels with good photometric performance, straylight should use a different threshold (10% or similar) to ensure that no faint flux from the primary light path is used to construct the scattered light model.  This biasing has been observed in simulated data when the straylight algorithm uses an inclusive cut level.

There are a couple of possibilities for implementation.  The straylight routine could be modified to point directly to the updated REGIONS file and pull out the appropriate threshold image.  Alternatively, assign_wcs could be rewritten in such a way that it provides the slice mappings at all thresholds rather than the one that is most appropriate for cube building later on.

Since the detailed implementation of this will depend on the changes to the MRS distortion solution and REGIONS reference files, it must be taken care of after https://jira.stsci.edu/browse/JP-650

drlaw1558:
This pull request addresses https://jira.stsci.edu/browse/JP-651 (update MRS straylight module to use the new REGIONS files delivered in CDP-7).

The majority of changes center on straylight_step.py and now read in the MRS slice map from the 20% threshold extension of the REGIONS file instead of the (deprecated) STRAYMASK reference file or the slice map attached by assign_wcs.

Both 'Nearest' and 'ModShepard' algorithms have been updated to use this new reference file and tested on simulated data.

Resolves #3377.


nden:
I will fix the failing sphinx. The `straymask` reference file description should be removed from the docs, right?

drlaw1558:
Thanks; yes, I tried to link to the REGIONS reference file description instead but I'm not sure of the policy on how .rst vs .inc references are being used.

nden:
Are the second and third paragraph in `straylight/reference_files.rst` still valid?
The code seems to always use the `regions` reference file. Should those two paragraphs be removed?

drlaw1558:
I thought I'd removed those two paragraphs already; that's correct, paragraphs 2 and 3 should be taken out.

drlaw1558:
Have now merged in changes from Nadia, and fixed a couple of log statements

PR #3606, opened 2019-06-06
Change PATTSIZE from number to string

### LINKED ISSUE #3593: Change PATTSIZE from float to string
Issue [JP-762](https://jira.stsci.edu/browse/JP-762) was created by Howard Bushouse:

The final outcome of cross-instrument team analysis of all dither-related keywords is that PATTSIZE should be a string value for all instruments, instead of a float. The datamodels schema(s) need to be updated appropriately.

See related discussions in JWSTKD-109 and JWSTKD-308.

This is affecting GSEG3 data processing and hence is high priority for fixing ASAP.

hbushouse:
Updated core schema to change data type of PATTSIZE keyword from number to string, and removed arcsec units designation from the keyword comment, because the values will not be numerical (they'll be SMALL, MEDIUM, LARGE).

Fixes #3593.


hbushouse:
Now that CI is passing, can I get a little love for this PR? ;-)

jdavies-st:
If they are SMALL, MEDIUM and LARGE, should that be an enum?  Or do we not want to be that strict?

hbushouse:
I thought briefly about adding an enum list, but given the way things are still in flux these days I decided it's safer, for now, to leave it out and just take whatever comes along.

PR #3607, opened 2019-06-06
Add the fxd_slit keyword as the third optical component

### LINKED ISSUE #2263: Rule: Level 3: NIRSpec Fixed Slit has third "optical path" element: FXD_SLT
# Issue
NIRSpec Fixed-slit, `NRS_FIXEDSLIT`, has the following "optical path" elements:

- `filter`
- `grating`
- `subarray`
- `fxd_slit`

Currently, `fxd_slit`, the slit in use, is not accounted for and needs to be a constraint.

stscieisenhamer:
Resolves #2263
WIP JP-750

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,12 +19,17 @@ ami
   throughput reference file cannot be found.  [#3567]
 
 assign_wcs
-------------
+----------
 
 - Fix a one pixel off problem with the NIRSpec NRS2 WCS transforms. [#3473]
 
 - Raise a ``ValueError`` if the FWCPOS keyword isn't found in input NIRISS
   WFSS images. [#3574]
+
+associations
+------------
+
+- Added the fxd_slit keyword as the third optical component [#3607]
 
 combine_1d
 ----------
@@ -417,8 +422,6 @@ associations
 - Fixed NIRISS WFSS catalog naming and implement NIRCam WFSS [#3515]
 
 - Fixed treating non-science as TSO [#3601]
-
-- Added the fxd_slit keyword as the third optical component [#3607]
 
 background
 ----------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -418,6 +418,8 @@ associations
 
 - Fixed treating non-science as TSO [#3601]
 
+- Added the fxd_slit keyword as the third optical component [#3607]
+
 background
 ----------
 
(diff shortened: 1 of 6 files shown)


CHANGES.rst
```diff
@@ -418,6 +418,8 @@ associations
 
 - Fixed treating non-science as TSO [#3601]
 
+- Added the fxd_slit keyword as the third optical component [#3607]
```
hbushouse: This new entry needs to be up at the top of this file in the new 0.13.4 section.

jwst/associations/lib/dms_base.py
```diff
@@ -556,27 +556,21 @@ def _get_opt_element(self):
...
-                )
+        # Retrieve all the optical elements
+        opt_elems = []
+        for opt_elem in ['opt_elem', 'opt_elem2', 'opt_elem3']:
+            try:
+                value = format_list(self.constraints[opt_elem].found_values)
+            except KeyError:
+                pass
+            else:
+                if value not in _EMPTY and value != 'clear':
```
hbushouse: So if an instrument like NIRCam or NIRISS, which have both filter and pupil wheels, has an exposure with PUPIL='CLEAR' (or FILTER='CLEAR'), which is a valid mode for science observations, the value "clear" won't appear as an optical element in the source-based file name? I would think we want that to appear, just like any other non-clear value. I know it doesn't add anything to the actual optical path, but it is documentation of the setup.

stscieisenhamer: If FXD_SLIT is where `source_id` is populated from, then yes, this is redundant.  The selection of what will be associated will still depend on FXD_SLIT: Only level2b files with the same FXD_SLIT will be placed into the level3 association.

stscieisenhamer: About the `clear`: That was a decision long, long ago, but looking over the latest documentation, it is present in names, so I'll file an issue to ensure that `clear` is no longer removed.

hbushouse:
OK, I withdraw my concern about the possibility of having the FXD_SLIT name appear in 2 fields of the source-based file names. A closer look at #3584 indicates that simple integers (1 through 5) are being used to set the source_id for NIRSpec fixed slits, so the string names will only appear as the 3rd optical element. Of course the FXD_SLIT part of the name will now be semi-redundant with the source_id field in the file name, but that's not a big deal.

stscieisenhamer:
@hbushouse, if there are no other issues, can one assume that the pr is approved?

hbushouse:
The only unresolved issue is that the change log entry needs to be moved to the top of the log (the 0.13.4 area).

jwst/associations/lib/dms_base.py
stscieisenhamer: As per later commentary, this is now no longer an issue. Resolving.

PR #3608, opened 2019-06-06
Turn on crds log times

jaytmiller:


jaytmiller:
This adds date/time to the log messages CRDS outputs while running for the CAL code,  similar to the formatting of CAL log messages.   For other CRDS use cases like file submissions and certify,  time is not added.

Assuming this is approved,  please let me do the merge.

jaytmiller:
This shouldn't add any extra messages;  it will just add time to messages coming from "CRDS" instead of stpipe;  those are most prominent as DEBUG messages during DMS pipeline runs.

PR #3542, opened 2019-05-29
JP-688 Additions to have MIRI MRS dedicated background targets listed in a science association

ddavis-stsci:
… in a science association

PR #3611, opened 2019-06-06
JP-287 Remove exp_type="NIS_EXTCAL" from associations

ddavis-stsci:
This update removes data files with exp_type="NIS_EXTCAL" from the association files. This was found using the jw00629_20190517t170931_pool.csv file from the SOR4B tests. The differences between the two directories show only the two association files listed as NIS_EXTCAL missing, the other files are the same, excluding the code version and list of constraints. 
```
diffs -I "code_version" -I "constraints" output/ output_baseline/
Only in output_baseline/: jw00629-o009_wfs-image2_001_asn.json
Only in output_baseline/: jw00629-o010_wfs-image2_001_asn.json
```
The unit tests show,
```
-- Docs: https://docs.pytest.org/en/latest/warnings.html
============= 132 passed, 2 xfailed, 1 warnings in 222.52 seconds ==============
```
The above pool file generated no level 3 associations.


CHANGES.rst
```diff
@@ -1,6 +1,12 @@
 0.13.4 (Unreleased)
 ===================
 
+associations
+---
+
+- MIRI MRS dedicated background exposures are now listed as science observations in
```
hbushouse: Do the changes in #3542 really only affect MIRI MRS dedicated bkg exposures, or all dedicated bkg exposures? The particular example it was tested on is MIRI MRS, but hopefully it works for all.

hbushouse: Oh, and you need a change log entry for this PR (#3611).

jwst/associations/lib/rules_level2b.py
```diff
@@ -932,7 +932,16 @@ def __init__(self, *args, **kwargs):
                 sources=['visitype'],
                 value='.+wfsc.+',
```
stscieisenhamer: The value here should not be '.+wfsc.+' but directly 'prime_wfsc_sensing_control'.

ddavis-stsci:
Maybe I'm just confused. My understanding was that only the 
PRIME_WFSC_SENSING_ONLY that have
exp_type = 'NIS_EXTCAL' should not be used to create associations. In 
the pool file from SDP all the
visit_types are PRIME_WFSC_SENSING_ONLY. The fix below removes all the 
PRIME_WFSC_SENSING_ONLY observations, not just 
jw00629-o009_wfs-image2_001_asn.json and 
jw00629-o010_wfs-image2_001_asn.json.
So we don't want any of the wfs-image2 associations created or is this 
an error when SDP generates the
pool file?

I've put the pool file that is causing the problem on
/home/ddavis/Data/associations/jw00629_20190517t170931_pool.csv
the unix side if you want to take a look.

hbushouse:
We don't want associations created for ANY exposure that has EXP_TYPE='NIS_EXTCAL', regardless of what kind of visit they came from, etc. The level-2b cal pipelines cannot handle exposures with this EXP_TYPE and hence processing needs to stop at the end of level-2a. So no ASN's at all for anything with EXP_TYPE='NIS_EXTCAL'. The confusion may have originated due to the fact that the particular example that was cited happened to be contained in a WFSC visit.

CHANGES.rst
ddavis-stsci: Yes, this is only MIRI_MRS dedicated bkg exposures, this is a list so I can add any other exposure types needed.

stscieisenhamer:
Then the only thing is the more restrictive check on the WFSC.

ddavis-stsci:
Asn_Lv2WFSC now selects on VISITYPE='prime_wfsc_sensing_control' to isolate exposures that should have WFSC processing

stscieisenhamer:
@ddavis-stsci you closed this PR without merging? Was that what you wanted?

ddavis-stsci:
No I did not mean to close this. Sigh.

hbushouse:
Can this be finalized in time for the upcoming 0.13.4 release (i.e. finished by the end of this week)?

ddavis-stsci:
I made the change that Jonathan requested and it has passed all the tests so I think it can be merged if Jonathan is OK with it.

PR #3612, opened 2019-06-06
revert code to not trim filter data.

dmggh:


dmggh:
This reverts an inadvertent change that was included in the previous update.

PR #3616, opened 2019-06-07
Remove enum list for SUBPXPAT keyword

### LINKED ISSUE #3617: Update schema enum for SUBPXPAT keyword
Issue [JP-772](https://jira.stsci.edu/browse/JP-772) was created by Howard Bushouse:

https://jira.stsci.edu/browse/JWSTDMS-18 reports that the datamodels schema validation is causing failures due to new values of the SUBPXPAT keyword being used in some exposures. We need to update our schema to allow validation to succeed.

hbushouse:
Removed the enum list in the core schema for the SUBPXPAT keyword. Many new values have been implemented in the front-ends recently and are still in flux. Hence the safest way to achieve validation for now is to have any enum list at all.

This addresses https://jira.stsci.edu/browse/JP-772

Resolves #3617.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ datamodels
 
 - Changed PATTSIZE keyword data type from float to string. [#3606]
 
+- Removed the enum list for the SUBPXPAT keyword to allow validation of any value. [#3616]
+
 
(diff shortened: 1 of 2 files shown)


hbushouse:
Fixes #3617

PR #3625, opened 2019-06-10
Change MRS throughput threshold for assign_wcs

drlaw1558:
Continued testing with simulated data and the new CDP-7 photometric reference files indicates that an 80% throughput threshold gives better performance for mosaiced data than the 50% threshold.  This pull request changes that default.

### DIFF
--- a/jwst/assign_wcs/miri.py
+++ b/jwst/assign_wcs/miri.py
@@ -422,8 +422,8 @@ def detector_to_abl(input_model, reference_files):
 
     with RegionsModel(reference_files['regions']) as f:
         allregions = f.regions.copy()
-        # Use the 50% throughput slice mask
-        regions=allregions[4,:,:]
+        # Use the 80% throughput slice mask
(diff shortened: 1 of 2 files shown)


jwst/assign_wcs/miri.py
```diff
@@ -422,8 +422,8 @@ def detector_to_abl(input_model, reference_files):
 
     with RegionsModel(reference_files['regions']) as f:
         allregions = f.regions.copy()
-        # Use the 50% throughput slice mask
-        regions=allregions[4,:,:]
+        # Use the 80% throughput slice mask
+        regions=allregions[7,:,:]
```
jdavies-st: The regression test that tests this will have to updated as well.

https://github.com/spacetelescope/jwst/blob/06d53043fb642142652c89931eef6237db3ba1a7/jwst/tests_nightly/general/miri/test_miri_steps_single.py#L174

drlaw1558: Done

Issue #3602, opened 2019-06-05
Associations are creating duplicate product names


stscijgbot:
Issue [JP-767](https://jira.stsci.edu/browse/JP-767) was created by Jonathan Eisenhamer:

Associations are being created with duplicate product names. This is failure.

Example test run:

{{pytest -ra --slow --bigdata jwst/tests_nightly/general/associations/test_sdp_pools.py -k test_specified_sdp_pool_dup_product_names --sdp-pool jw00625_20190603t233254_pool}}

stscieisenhamer:
Way too optimistic; issue not resolved.

stscijgbot:
Comment by Jonathan Eisenhamer: Way too optimistic on this one. Issue still exists.

stscieisenhamer:
Furthermore, this is actually an AMI3 issue: the same association is being created for each science/psf exposure.

PR #3629, opened 2019-06-10
Ensure that higher-order candidates only exist for Level2 associations

### LINKED ISSUE #3041: Level 2 associations and overwriting of products
# Issue
For Level2, there is a situation where different association candidates can be produced, resulting in the possibility that the '_cal' product will be replaced with a lesser association. 

The particular case is when there is a background association. In initial DP processing, there will not be an issue since the background candidate will always complete after the initial observation candidate. However, when reprocessing, both associations will exist, and if the observation candidate is processed after the background candidate, the final '_cal' file will be the bad, non-background-subtracted, result.

The generator-specific solution is to check for this case before writing out the final associations, and remove the "lesser" association candidate, when using the `--all-candidates` flag.

stscieisenhamer:
Resolves #3041


stscieisenhamer:
Also resolves #3602/JP-767

Issue #3623, opened 2019-06-10
Update WFSS slit names to use simple integer value


stscijgbot:
Issue [JP-776](https://jira.stsci.edu/browse/JP-776) was created by Howard Bushouse:

The "extract_grism_objects" function in extract_2d/grisms.py needs to be updated to use a simple integer value as the slit name, instead of the current "Object_\{obj.sid}", to be consistent with NIRSpec fixed-slit and MOS slit names. Basically just use the obj.sid value by itself, without prepending "Object_". This affects the format of the source-based file names for WFSS data.

sosey:
https://github.com/spacetelescope/jwst/blob/master/jwst/extract_2d/grisms.py#L386 should be the only change, I was using the slit name for printing information only

sosey:
Don't forget to update the unit test here: https://github.com/spacetelescope/jwst/blob/master/jwst/extract_2d/tests/test_grisms.py#L331

PR #3632, opened 2019-06-11
Update WFSS slit names to use simple integer value (#3623) (JP-776)

dmggh:
Will update unit tests.

### DIFF
--- a/jwst/extract_2d/tests/test_grisms.py
+++ b/jwst/extract_2d/tests/test_grisms.py
@@ -6,7 +6,7 @@ No images are needed here to check the location
 and size of bounding boxes.
 
 NIRCAM and NIRISS WFSS use the same code, testing with NIRCAM
-settings here for funcationality. In the future, full data
+settings here for functionality. In the future, full data
 regression tests will provide the truth between instruments.
 
 For the testing catalog:
@@ -348,6 +348,9 @@ def test_extract_wfss_object():
     assert len(outmodel.slits) == 3
     ids = [slit.source_id for slit in outmodel.slits]
     assert ids == [9, 19, 19]
+
+    names = [slit.name for slit in outmodel.slits]
+    assert names == ['9', '19', '19']
 
     # check that bounding boxes exist
     del outmodel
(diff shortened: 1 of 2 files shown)


hbushouse:
Fixes #3623

jwst/extract_2d/tests/test_grisms.py
```diff
@@ -349,5 +349,8 @@ def test_extract_wfss_object():
     ids = [slit.source_id for slit in outmodel.slits]
     assert ids == [9, 19, 19]
```
nden: Is this just bogus data or can there be more than one source with the same id?

dmggh: I don't know - these are from @sosey .

hbushouse: Could this be a case where more than one source is in a slitlet? E.g. in this case source 9 is in one of the 3 shutters and source 19 is in the other two? Or is that not what I'm seeing here?

hbushouse: Oh, never mind. I've got NRS MSA on my mind. We're talking about WFSS data here. I suppose it's possible to have 2 sources defined in the source catalog that are close enough to one another on the sky such that when a slit is defined for one of them, it happens to also include (be contaminated by) the spectrum of the nearby source.

nden: There are two cases:

- two sources in one slit - we should still get two different source_id
- one source in two slits (seemingly the above case) - Shouldn't we still see two different source_ids since these are different locations on the sky (even if it's an extended source)?

hbushouse: All depends on how the source detection routine works in the `source_catalog` step that runs on the direct image in a WFSS observation. I guess it's possible that with complicated extended sources it could catalog individual nearby pieces of the source with unique ID's.

hbushouse: The comments in the test say it's a test of extracting 2 sources, yet it seems to be expecting a list of 3 to be returned, so that seems inconsistent. It doesn't really have anything to do with this particular update to the code, so I vote to ignore it for now and move along with getting this change merged.

sosey: Yah, I think it's a mistake by me, I'll take a closer look and remind myself what was going on and submit a separate PR to fix?

PR #3635, opened 2019-06-11
Update model blender to return strings instead of astropy Time objects

### LINKED ISSUE #3621: calwebb_ami3 failing in blendmodels
Issue [JP-778](https://jira.stsci.edu/browse/JP-778) was created by Howard Bushouse:

After recent updates to blendmodels and ami_analyze (not sure if they're connected) the calwebb_ami3 regression test is failing when attempting to run blendmodels after completing the ami_normalize step.

E.g. [https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/274/testReport/jwst.tests_nightly.general.niriss.test_niriss_steps_single/TestAMIPipeline/_stable_deps__test_ami_pipeline/]

eslavich:
https://jira.stsci.edu/browse/JP-778

This PR introduces two additional blender rules, `mindatetime` and `maxdatetime`, and has all of the date/time rules return strings in the appropriate format instead of astropy Time objects.

This may be controversial, but I'm also dropping support for the asdf time schema in meta.date and meta.observation.date.  As far as I know there is only one permissible format, so I'm not sure why the schema needs to allow everything in asdf/time/time-1.0.0.  This is relevant because the model blender sees that meta.observation.date can be an object with its own nested fields, so it ignores the blend_* settings (like it does with any object).

Anyway, with these changes the test_ami_pipeline regression test no longer crashes -- it still doesn't pass, but that doesn't seem related.

Resolves #3621.

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -158,7 +158,9 @@ properties:
             title: "[yyyy-mm-dd] UTC date at end of exposure"
             fits_keyword: DATE-END
             blend_table: True
-            blend_rule: maxdate
+            # Despite what the title might lead you to believe, this keyword
+            # actually contains a full datetime string.
+            blend_rule: maxdatetime
           time_end:
             title: "[hh:mm:ss.sss] UTC time at end of exposure"
             type: string
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -4,10 +4,8 @@ properties:
     type: object
     properties:
       date:
-        anyOf:
-          - $ref: http://stsci.edu/schemas/asdf/time/time-1.0.0
-          - type: string
         title: "[yyyy-mm-ddThh:mm:ss.ss] UTC date file created"
+        type: string
         fits_keyword: DATE
         blend_table: True
       origin:
@@ -138,10 +136,8 @@ properties:
         type: object
         properties:
           date:
-            anyOf:
-              - $ref: http://stsci.edu/schemas/asdf/time/time-1.0.0
-              - type: string
             title: "[yyyy-mm-dd] UTC date at start of exposure"
+            type: string
             fits_keyword: DATE-OBS
             blend_table: True
             blend_rule: mindate
@@ -156,7 +152,7 @@ properties:
             title: "Date-time start of data acquisition"
             fits_keyword: DATE-BEG
             blend_table: True
-            blend_rule: mindate
+            blend_rule: mindatetime
           date_end:
             type: string
             title: "[yyyy-mm-dd] UTC date at end of exposure"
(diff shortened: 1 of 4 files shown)


eslavich:
@hbushouse I don't possess the ability to add reviewers, so here's a comment instead

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -156,7 +152,7 @@ properties:
             title: "Date-time start of data acquisition"
             fits_keyword: DATE-BEG
             blend_table: True
-            blend_rule: mindate
+            blend_rule: mindatetime
```
hbushouse: The rule mindatetime for DATE-BEG is correct, because it contains a full date+time string. DATE-END is also being populated by SDP as a full date+time string, despite the information in the keyword dictionary and the keyword comment field that suggests it is only a date. Hence the blend rule for DATE-END (line 161) needs to be changed to "maxdatetime".

eslavich: This has been remedied.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -156,13 +152,15 @@ properties:
...
           date_end:
             type: string
             title: "[yyyy-mm-dd] UTC date at end of exposure"
             fits_keyword: DATE-END
             blend_table: True
-            blend_rule: maxdate
+            # Despite what the title might lead you to believe, this keyword
+            # actually contains a full datetime string.
```
hbushouse: We'll need to fix the keyword title at some point, to keep in sync with what's now in the keyword dictionary (it's been updated there to reflect the fact that it's now a full date-time value), but let's hold off on doing that temporarily, because it'll cause MASSIVE numbers of failures in our regression tests that we don't want to deal with when getting ready for another release.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -4,10 +4,8 @@ properties:
     type: object
     properties:
       date:
-        anyOf:
-          - $ref: http://stsci.edu/schemas/asdf/time/time-1.0.0
-          - type: string
```
jdavies-st: @nden or @hbushouse, anyone remember why Time objects were added to the core schema?  These cannot be serialized to a FITS keyword.

hbushouse: Nope. That's a question for either Mike D. or Bernie. I guess they thought it would be an elegant way to do validation of the formatting for the date+time keywords, but never really worked as hoped.

jdavies-st: https://github.com/spacetelescope/jwst/commit/6621c6fba120b38b460c462f1c34281221e35ffd

nden: @jdavies Why not? The value of a `Time` object in `isot` format is the value of the FITS keyword.
But that's not something to be done before a release. I think it's fine to remove those from schema now.

Issue #3596, opened 2019-06-05
outlier_detection error for MIRI TSO imaging in calwebb_tso3


stscijgbot:
Issue [JP-765](https://jira.stsci.edu/browse/JP-765) was created by Howard Bushouse:

JWSTDMS-15 reports another "output array is read-only" error seen in the outlier_detection step when processing MIRI Image TSO exposures in the calwebb_tso3 pipeline.

stscijgbot:
Comment by Howard Bushouse: These errors occurred for NOps4 datasets and hence is medium-high priority for fixing.

stscijgbot:
Comment by Howard Bushouse: Example data files have been copied to /grp/jwst/ssb/chartreuse/dms_bugs/jwstdms-15/, including the input asn and calints files, as well as all necessary cfg files.

nden:
Fixed in #3638

jdavies-st:
The underlying problem here is caused by a bug in `numpy`.

https://github.com/numpy/numpy/issues/12283

The above workaround will fix this dataset, but it could crop up again if all the array elements are masked.  The better workaround probably means avoiding sending fully-masked arrays to `astropoy.stats.sigma_clipped_stats`.

PR #3638, opened 2019-06-11
Changed good_bits from 4 to 6 to exclude SATURATION and JUMP_DET flags (JP-765)

stscirij:
Pixels with just either of these flags set won't be excluded from weight calculation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,11 @@ exp_to_source
 
 - Removed the enum list for the SUBPXPAT keyword to allow validation of any value. [#3616]
 
+outlier_detection
+-----------------
+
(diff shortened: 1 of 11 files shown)


stscirij:
The good_bits flag is also used in resample and resample_spec - do we want to change it from 4 to 6 there too?

jdavies-st:
I think that makes sense to make the change in `resample` too.  It will suffer from the same problem.

stscicrawford:
Let's do it in one PR and then please add a change log to this one as well.  This addresses the problem raised in #3596 though a ticket should be created to look at potential other ways to address the issue as discussed today.

hbushouse:
https://jira.stsci.edu/browse/JP-787 has been filed to do the future work of redesigning the way flags are used in these steps.

PR #3646, opened 2019-06-12
JP-764 Do not use CRPIX in TSGRISM mode

### LINKED ISSUE #3594: calwebb_spec2 extract_2d failure for NRC_TSGRISM exposures
Issue [JP-764](https://jira.stsci.edu/browse/JP-764) was created by Howard Bushouse:

JWSTDMS-14 reports a failure in the calwebb_spec2 extract_2d step when processing a NIRCam TSGRISM exposure. It appears to be due to an attempted use of FITS WCS keywords like CRPIXn that are no longer included in the headers of spectroscopic exposures.

This affects processing of NOps4 datasets and hence is medium-high priority for fixing, either in B7.3 or a patch soon after B7.3 release.

nden:
For `EXP_TYPE=NRC_TSGRISM` read in CRPIX1/2 values from the SIAF and save them as `model.meta.wcsinfo.siaf_xref_sci` and `model.meta.wcsinfo.siaf_yref_sci` so they can be used as a source location by extract_2d.

Resolves #3594


jwst/extract_2d/grisms.py
```diff
@@ -116,8 +118,8 @@ def extract_tso_object(input_model,
         lmin, lmax = range_select.pop()
 
         # create the order bounding box
-        source_xpos = input_model.meta.wcsinfo.crpix1 - 1  # remove fits
-        source_ypos = input_model.meta.wcsinfo.crpix2 - 1  # remove fits
+        source_xpos = input_model.meta.wcsinfo.siaf_xref_sci - 1  # remove fits
+        source_ypos = input_model.meta.wcsinfo.siaf_xref_sci - 1  # remove fits
```
hbushouse: change "xref" to "**y**ref"

jwst/extract_2d/grisms.py
```diff
@@ -192,7 +194,7 @@ def extract_tso_object(input_model,
             output_model.dq = ext_dq
             output_model.meta.wcs = subwcs
             output_model.meta.wcs.bounding_box = util.wcs_bbox_from_shape(ext_data.shape)
-            output_model.meta.wcs.crpix2 = 34  # update for the move, vals are the same
+            output_model.meta.wcs.siaf_yref_sci = 34  # update for the move, vals are the same
```
hbushouse: Above siaf_yref_sci is part of the meta.wcs**info** attribute, but here it's part of meta.wcs

nden: Hm, good point. I'll investigate.

nden:
Correct, these are used only internally and won't show as FITS keywords.
However, their value is used to populate `slit.source_x(y)pos` which is saved as FITS keyword `SRCX(Y)POS`.

jwst/extract_2d/grisms.py
nden: fixed

jwst/extract_2d/grisms.py
nden: fixed - appears to be an old bug

jwst/lib/set_telescope_pointing.py
```diff
@@ -1452,6 +1452,9 @@ def populate_model_from_siaf(model, siaf):
         model.meta.wcsinfo.cdelt1 = siaf.cdelt1 / 3600  # in deg
         model.meta.wcsinfo.cdelt2 = siaf.cdelt2 / 3600  # in deg
         model.meta.coordinates.reference_frame = "ICRS"
+    elif model.meta.exposure.type.lower() == 'nrc_tsgrism':
+        model.meta.wcsinfo.siaf_xref_sci = siaf.crpix1
+        model.meta.wcsinfo.siaf_yref_sci = siaf.crpix2
```
jdavies-st: It might be worth adding both of these to the `wcsinfo.schema` so they are checked when read in and written out.  Just don't give it a `fits_keyword` or `fits_hdu` attribute.  So something like:
```
          siaf_xref_sci:
            title: Aperture X reference point in SCI frame from SIAF
            type: number
          siaf_yref_sci:
            title: Aperture Y reference point in SCI frame from SIAF
            type: number
```
?

It will make it clear that they're supposed to be there.

jwst/extract_2d/tests/test_grisms.py
```diff
@@ -150,6 +144,8 @@ def create_tso_wcsimage(filtername="F277W", subarray=False):
 
     hdul['sci'].data = np.ones((2, subsize, 2048))
     im = CubeModel(hdul)
+    im.meta.wcsinfo.siaf_xref_sci = 887.0
+    im.meta.wcsinfo.siaf_yref_sci = 35.0
```
jdavies-st: Is there a reason to hard-code these without a check on whether SDP has populated them?  Might be good to first check, and than have this as a fall-back, for data that hasn't gone through the new SDP running of `set_telescope_pointing` yet.

jwst/extract_2d/grisms.py
```diff
@@ -192,7 +194,8 @@ def extract_tso_object(input_model,
             output_model.dq = ext_dq
             output_model.meta.wcs = subwcs
             output_model.meta.wcs.bounding_box = util.wcs_bbox_from_shape(ext_data.shape)
-            output_model.meta.wcs.crpix2 = 34  # update for the move, vals are the same
+            output_model.meta.wcsinfo.siaf_yref_sci = 34  # update for the move, vals are the same
+            output_model.meta.wcsinfo.siaf_xref_sci = input_model.meta.wcsinfo.siaf_xref_sci
```
jdavies-st: Not sure I understand the hard-coding of one and not the other here.  Also the hard-coding is in zero-indexed, not 1-indexed coordinates.  This keyword is supposed to be FITS, i.e. 1-indexed?

nden: The hard-coding is explained in the comments. It looks like it is supposed to be what it is for reasons not related to indexing.
Note, that the goal is to reproduce the previous code here, not invent new algorithms.

hbushouse: The NIRCam TS folks wanted the extraction box moved a couple rows for the case where one of the subarrays is in use, such that the cutout is intentionally off-center a bit from the source location. That's why the hardcoded value for yref.

jdavies-st:
Test failures on Jenkins are unrelated.

jwst/extract_2d/tests/test_grisms.py
nden: What SDP? This test creates the data.

jdavies-st: Ah, nevermind.  I see this is in the test.

PR #3648, opened 2019-06-13
Confirm Completion Keyword Assignment for Flat Field Correction

### LINKED ISSUE #3532: Spec2: Flat field Unit Tests
Issue [JP-723](https://jira.stsci.edu/browse/JP-723) was created by Maria Pena-Guerrero:

The tests for the flat field test can be located at:

[https://github.com/spacetelescope/nirspec_pipe_testing_tool/blob/master/calwebb_spec2_pytests/F_flat_field/test_flat_field.py]

The completion test (line 261) can be migrated into a unit test, as well as all the reference file checks (line 291 onwards).

 

* NOTE * 

In the latest dev_version of the pipeline, the pipeline is crashing because we are still using the flat_suffix option as a string rather than a boolean. PTT needs to be modified to match the changes in the pipeline.

mfixstsci:
This PR is a part of the INS migration of unit tests to the pipeline. This test was a request by the NIRSPEC team to check the completion of the flat field step. I added an assertion to an already existing unit test that just confirms the keyword is value `COMPLETE`. If it's necessary, I can make a separate test.

Resolves #3532.

### DIFF
--- a/jwst/flatfield/tests/test_flatfield.py
+++ b/jwst/flatfield/tests/test_flatfield.py
@@ -53,7 +53,7 @@ def test_flatfield_step_interface(instrument, exptype):
 
     assert (result.data == data.data).all()
     assert result.var_flat.shape == shape
-
+    assert result.meta.cal_step.flat_field == 'COMPLETE'
 
 def exptypes():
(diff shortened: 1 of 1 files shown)


jdavies-st:
Test failure is unrelated to this PR.  Merging.

PR #3650, opened 2019-06-13
fix for JP-786

jemorrison:
If a photometric catalog cannot be created a log message is printed and cleanly finishes calwebb_tso3


stscicrawford:
Can you also delete line 40 'image_exptypes = ['NRC_TSIMAGE', 'MIR_IMAGE']' as that would seem not to be needed any longer. 

This also needs a changelog entry.

PR #3653, opened 2019-06-13
Synchronize datamodels core schema with JWST Keyword Dictionary

### LINKED ISSUE #3624: Synchronize datamodels core schema with JWST Keyword Dictionary
Issue [JP-779](https://jira.stsci.edu/browse/JP-779) was created by Howard Bushouse:

Before B7.3 gets locked down by both SDP and Cal, we should do a comparison of the "official" JWST Keyword Dictionary json schemas, because they've made a number of updates recently to begin incorporating requests from the instrument teams to fix up all dither-related keywords. This involves getting a dump of the latest json files used by SDP and running Bernie Simon's old comparison script that compares them to our yaml schemas and lists differences. Need to figure out who's got that code and how to run it. Also need to send a request to Kim Duprie to get a dump of the current set of json schemas used in the dictionary.

eslavich:
Resolves #3624 

This updates the datamodels schema files with new keywords that appeared in the 2019-06-11 dump of the dev keyword dictionary.  I didn't change any titles, but left notes showing where they need an update.  I have some questions about what I did but I'll comment directly on the relevant lines.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,33 @@ datamodels
 
 - Added enumeration of allowed values of ``FXD_SLIT`` to the core schema. [#3584]
 
+- Changed ``WHT_TYPE`` keyword to ``RESWHT``. [#3653]
+
+- Add missing pattern/enum values to keyword_pband, keyword_pfilter, keyword_channel [#3653]
+
+- New keywords [#3653]
+   - ``DSETSTRT``
+   - ``NUMDSETS``
+   - ``DITHDIRC``
+   - ``DITHOPFR``
+   - ``DITHPNTS``
+   - ``MRSPRCHN``
+   - ``NDITHPTS``
+   - ``DWTSCL``
+   - ``DOUTUN``
+   - ``RESBITS``
+   - ``DFVAL``
+   - ``DPIXFR``
+   - ``DKERN``
+   - ``SCIEXT``
+   - ``CONEXT``
+   - ``WHTEXT``
+
+extract_1d
+----------
+
+- Checks for input from a SourceModelContainer. [#3649]
+
 exp_to_source
 -------------
 
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -155,6 +155,7 @@ properties:
             blend_rule: mindatetime
           date_end:
             type: string
+            # 2019-06-12: Title has changed to "Date-time end of data acquisition" in keyword dictionary.
             title: "[yyyy-mm-dd] UTC date at end of exposure"
             fits_keyword: DATE-END
             blend_table: True
@@ -1008,6 +1009,54 @@ properties:
             type: number
             fits_keyword: YOFFSET
             blend_table: True
+          starting_set:
+            title: Starting 4-point set number
+            type: integer
+            # 2019-06-12: The keyword dictionary specifies an enum constraint for this field, but
+            # due to discrepancies in the dictionary JSON we're going to leave this unconstrained
+            # for now.
+            # enum: [1, 3, 5]
+            fits_keyword: DSETSTRT
+          number_of_sets:
+            title: Total number of 4-point sets
+            type: integer
+            # 2019-06-12: The keyword dictionary specifies an enum constraint for this field, but
+            # due to discrepancies in the dictionary JSON we're going to leave this unconstrained
+            # for now.
+            # enum: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+            fits_keyword: NUMDSETS
+          direction:
+            title: Direction of dither pattern offsets
+            type: string
+            # 2019-06-12: The keyword dictionary specifies an enum constraint for this field, but
+            # due to discrepancies in the dictionary JSON we're going to leave this unconstrained
+            # for now.
+            # enum: [NEGATIVE, POSITIVE]
+            fits_keyword: DITHDIRC
+          optimized_for:
+            title: Dither pattern optimization
+            type: string
+            # 2019-06-12: The keyword dictionary specifies an enum constraint for this field, but
+            # due to discrepancies in the dictionary JSON we're going to leave this unconstrained
+            # for now.
+            # enum: ["POINT-SOURCE", "EXTENDED-SOURCE"]
+            fits_keyword: DITHOPFR
+          sparse_points:
+            title: Sparse cycling dither points
+            type: string
+            fits_keyword: DITHPNTS
+          primary_channel:
+            title: MRS primary channel
+            type: string
+            # 2019-06-12: The keyword dictionary specifies an enum constraint for this field, but
+            # due to discrepancies in the dictionary JSON we're going to leave this unconstrained
+            # for now.
+            # enum: [ALL, CHANNEL1, CHANNEL2, CHANNEL3, CHANNEL4]
+            fits_keyword: MRSPRCHN
+          dither_points:
+            title: Total number of points in primary dither pattern
+            type: string
+            fits_keyword: NDITHPTS
       ephemeris:
         title: JWST ephemeris information
         type: object
--- a/jwst/datamodels/schemas/keyword_channel.schema.yaml
+++ b/jwst/datamodels/schemas/keyword_channel.schema.yaml
@@ -9,5 +9,5 @@ properties:
           channel:
             title: NIRCAM or MIRI channel name.
             type: string
-            enum: [LONG, SHORT, '12', '34', '1', '2', '3', '4', ANY, N/A]
+            enum: [LONG, SHORT, '1234', '123', '234', '12', '34', '1', '2', '3', '4', ANY, N/A]
             fits_keyword: CHANNEL
--- a/jwst/datamodels/schemas/keyword_pband.schema.yaml
+++ b/jwst/datamodels/schemas/keyword_pband.schema.yaml
@@ -9,6 +9,6 @@ properties:
           p_band:
             title: MRS wavelength band
             type: string
-            pattern: ^((ANY|SHORT|MEDIUM|LONG|SHORT-MEDIUM|SHORT-LONG|\
+            pattern: ^((ANY|MULTIPLE|SHORT|MEDIUM|LONG|SHORT-MEDIUM|SHORT-LONG|\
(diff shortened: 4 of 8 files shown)


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1010,6 +1011,39 @@ properties:
...
+            fits_keyword: DSETSTRT
+          number_of_sets:
+            title: Total number of 4-point sets
+            type: integer
+            enum: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+            fits_keyword: NUMDSETS
+          direction:
+            title: Direction of dither pattern offsets
+            type: string
+            enum: [NEGATIVE, POSITIVE]
```
eslavich: The keyword dictionary has an explicit `NULL` value in the enum list for this and other keywords.  Do we need that?  I assumed that it would be represented in by a None in Python.

jwst/datamodels/schemas/lev3_prod.schema.yaml
```diff
@@ -4,22 +4,62 @@ properties:
...
+            # 2019-06-12: Title has changed to "[s] Total exposure time for product" in keyword dictionary.
             title: Total exposure time for product
             type: number
             fits_keyword: TEXPTIME
           weight_type:
             title: Type of drizzle weighting to use in resampling input
             type: string
             enum: [exptime, error]
-            fits_keyword: WHT_TYPE
+            fits_keyword: RESWHT
```
eslavich: Is it safe to change fits_keyword here?

jwst/datamodels/schemas/lev3_prod.schema.yaml
```diff
@@ -4,22 +4,62 @@ properties:
...
+          drizzle_weight_scale:
```
eslavich: The schema editor tool put this and the following keywords into core.schema.yaml, but it seemed to me that they belonged here.  Let me know if I've got that wrong.

jwst/datamodels/schemas/keyword_pband.schema.yaml
```diff
@@ -9,6 +9,6 @@ properties:
           p_band:
             title: MRS wavelength band
             type: string
-            pattern: ^((ANY|SHORT|MEDIUM|LONG|SHORT-MEDIUM|SHORT-LONG|\
+            pattern: ^((ANY|MULTIPLE|SHORT|MEDIUM|LONG|SHORT-MEDIUM|SHORT-LONG|\
```
eslavich: This now matches the enum list in keyword_band.schema.yaml

jwst/datamodels/schemas/keyword_pfilter.schema.yaml
```diff
@@ -16,5 +16,5 @@ properties:
               F212N|F2300C|F250M|F2550W|F2550WR|F277W|F290LP|F300M|\
               F322W2|F335M|F356W|F360M|F380M|F410M|F430M|F444W|\
               F460M|F480M|F560W|F770W|FLENS|FND|GR150C|\
-              GR150R|OPAQUE|P750L|WLP4|N/A|ANY)\\s*\\|\\s*)+$"
+              GR150R|OPAQUE|P750L|WLP4|F070W|MULTIPLE|N/A|ANY)\\s*\\|\\s*)+$"
```
eslavich: This now matches the enum list in keyword_filter.schema.yaml

jwst/datamodels/schemas/lev3_prod.schema.yaml
nden: This looks correct. @jdavies-st This won't break resample because it's using the data model, right?

jwst/datamodels/schemas/core.schema.yaml
hbushouse: They probably have the NULL values because some exposures won't use these attributes and hence they won't have a value. We need to determine exactly what value SDP will populate each keyword with in those cases. It can't be the string "NULL" for keywords that have int or float data types.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1385,6 +1419,10 @@ properties:
             type: string
             fits_keyword: GS_SRCKM
             blend_table: True
+          gs_jitter_rms:
+            title: RMS jitter over the exposure (arcsec).
+            type: number
+            fits_keyword: JITTERMS
```
hbushouse: https://jira.stsci.edu/browse/JWSTKD-298 includes a request to remove JITTERMS from the keyword dictionary. Perhaps that hasn't been done yet and hence it's still showing up in the json files used to do this comparison?

jwst/datamodels/schemas/keyword_channel.schema.yaml
```diff
@@ -9,5 +9,5 @@ properties:
           channel:
             title: NIRCAM or MIRI channel name.
             type: string
-            enum: [LONG, SHORT, '12', '34', '1', '2', '3', '4', ANY, N/A]
+            enum: [LONG, SHORT, '1234', '123', '234', '12', '34', '1', '2', '3', '4', ANY, N/A]
```
hbushouse: If the enum list is being updated in this sub-schema, then it should be updated in the core schema too.

hbushouse: Or were those additional values already in the core schema?

jwst/datamodels/schemas/lev3_prod.schema.yaml
hbushouse: Check with @jdavies-st for this one.

jwst/datamodels/schemas/lev3_prod.schema.yaml
jdavies-st: Correct.  It will just a diff in the regression tests.

jwst/datamodels/schemas/core.schema.yaml
jdavies-st: I suspect the keyword will not get populated if it doesn't apply to the data?

jwst/datamodels/schemas/keyword_channel.schema.yaml
eslavich: Yup, these were already in the core schema, now all three match.

nden: They appear to be in the core schema already.

jwst/datamodels/schemas/core.schema.yaml
nden: Then let's not include this in the schema.

jwst/datamodels/schemas/keyword_channel.schema.yaml
nden: However, the same change needs to be made to `keyword_pchannel`.

jemorrison: I am probably missing something - but for the NIRCAM or MIRI channel name you also need MEDIUM (for MIRI).

nden: @jemorrison Isn't `MEDIUM` a `band` value? `channel` for MIRI is a number, right?

jwst/datamodels/schemas/core.schema.yaml
nden: Sent email asking for clarification.

jwst/datamodels/schemas/keyword_channel.schema.yaml
jemorrison: My bad we use them interchangeable on the team. You are right  for the pipeline it is a number.

CHANGES.rst
```diff
@@ -22,6 +22,29 @@ datamodels
...
+
+- New keywords [#3653]
+   - ``DSETSTRT``
+   - ``NUMDSETS``
+   - ``DITHDIRC``
+   - ``DITHOPFR``
+   - ``DITHPNTS``
+   - ``MRSPRCHN``
+   - ``NDITHPTS``
+   - ``JITTERMS``
```
hbushouse: Should remove JITTERMS (we don't want to add it)

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1010,6 +1011,43 @@ properties:
...
+            # 2019-06-12: Remove the NULL value once the keyword dictionary has been corrected.
+            enum: [NEGATIVE, POSITIVE, "NULL"]
+            fits_keyword: DITHDIRC
+          optimized_for:
+            title: Dither pattern optimization
+            type: string
+            # 2019-06-12: Remove the underscore versions of these values once the keyword
+            # dictionary has been corrected.
+            # 2019-06-12: Remove the NULL value once the keyword dictionary has been corrected.
+            enum: ["POINT-SOURCE", "EXTENDED-SOURCE", POINT_SOURCE, EXTENDED_SOURCE, "NULL"]
```
hbushouse: Theoretically SDP should always be populating with the hyphen versions of these values, not the underscore. But I guess it's good to have the coverage just in case.

hbushouse: As a general rule, if we're not 100% sure what values PPS or SDP are going to be passing along, it's safer for us to just not use an enum list at all. Especially for these new dither-related keywords, for which the front-end systems (APT and PPS) are still in flux and hence these values may change frequently or this may not even be the correct list to start with. We can always come back and add an enum list when things have settled down more.

jwst/datamodels/schemas/core.schema.yaml
eslavich: Removed

CHANGES.rst
eslavich: Removed

jwst/datamodels/schemas/keyword_channel.schema.yaml
eslavich: (These are added to `keyword_pchannel` in this PR)

jwst/datamodels/schemas/core.schema.yaml
eslavich: I replaced this and other questionable enum constraints with a note.

nden:
Need to resolve conflicts.

nden:
@eslavich Could you rebase this to resolve the conflicts.

PR #3654, opened 2019-06-13
Remove editable install from Jenkins CI conda build

jdavies-st:
Hopefully this fixes the problems we're seeing with `astropy` defining `pytest_plugins` in a non-top-level `conftest.py`.


jdavies-st:
This is outside of our control.  A Jenkins CI issue.  Closing.

PR #3664, opened 2019-06-14
JP-784 For grism data, compute wavelengths and copy variance arrays

philhodge:
For each slit, function `extract_grism_objects` in extract_2d/grisms.py now computes and stores an array of wavelengths, and it extracts and stores the variance arrays VAR_POISSON, VAR_RNOISE, and VAR_FLAT.


jwst/extract_2d/grisms.py
```diff
@@ -375,13 +378,39 @@ def extract_grism_objects(input_model,
...
+                                                var_flat=var_flat)
                 new_slit.meta.wcsinfo.spectral_order = order
                 new_slit.meta.wcs = subwcs
+                grid = grid_from_bounding_box(new_slit.meta.wcs.bounding_box)
+                shape = grid[0].shape
+                wl = np.empty(shape, dtype=np.float64)
+                transform = new_slit.meta.wcs.forward_transform
+                for j in range(shape[0]):
+                    for i in range(shape[1]):
+                        wl[j, i] = transform(grid[0][j, i], grid[1][j, i])[2]
```
sosey: might be useful to add a reminder comment here that the transform returns (ra, dec, wavelength, order) so that future developers understand the [2] extraction immediately.

jwst/extract_2d/grisms.py
```diff
@@ -375,13 +378,39 @@ def extract_grism_objects(input_model,
...
+                                                err=ext_err,
+                                                dq=ext_dq,
+                                                var_poisson=var_poisson,
+                                                var_rnoise=var_rnoise,
+                                                var_flat=var_flat)
                 new_slit.meta.wcsinfo.spectral_order = order
                 new_slit.meta.wcs = subwcs
+                grid = grid_from_bounding_box(new_slit.meta.wcs.bounding_box)
+                shape = grid[0].shape
+                wl = np.empty(shape, dtype=np.float64)
```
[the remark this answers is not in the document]
philhodge: I agree; it was silly to try to save a few milliseconds and then take several minutes to compute the wavelengths.

jwst/extract_2d/grisms.py
philhodge: I should have included a comment here, yes.  I was in too much of a hurry.

PR #3679, opened 2019-06-17
Set TEST_BIGDATA to central store

jdavies-st:


jdavies-st:
And what looks like a problem with the CRDS server

https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/284/

Issue #3655, opened 2019-06-13
DQ column has type float64 in combine_1d


stscijgbot:
Issue [JP-789](https://jira.stsci.edu/browse/JP-789) was created by Edward Slavich:

From [~hodge]:

 
extract_1d wrote DQ values of 0, and the data type of the column was
uint32.  The values in the physical disk file are -2^31 (I checked using
od).  When read via astropy.io.fits, the values are correctly seen as 0
because bzero for that column is +2^31. Normally when I open this file
with datamodels.open, I see values of 2^31 (i.e. there's a bug); it
looks as if something is adding bzero a second time.  However, when I
read the column in combine_1d, I see values of 2^32 (yes, 4294967296),
and the data type is float64.  If I take a zero-filled array of type
uint32 and add 4294967296, I get an array of type uint64.  So this
doesn't quite make sense, but I suspect that the array is being
converted to float64 due to something that has to do with the incorrect
handling of bzero.

stscijgbot:
Comment by Philip Hodge: When I ran combine_1d by itself with an association of two files as input, the DQ column had data type uint32 and the values were 2147483648, i.e. the "normal" incorrect value that I see when opening a FITS BINTABLE using jwst.datamodels.open(filename). It was when I ran calwebb_spec3 that DQ in the combine_1d step was float64 with values 4294967296.0.

stscijgbot:
Comment by Philip Hodge: See also GitHub issue #3179.

stscijgbot:
Comment by Nadia Dencheva: This appears to be a bug in astropy. A workaround has been implemented while the biug is worked on.

nden:
For reference the issue on astropy is astropy/astropy#8862

hbushouse:
Fixed by #3680

stscijgbot:
Comment by Edward Slavich: Re-opening this because the PR has been reverted in master, and will require a gwcs update to merge.

stscijgbot:
Comment by Howard Bushouse: Fixed in [https://github.com/spacetelescope/jwst/pull/3680]

PR #3680, opened 2019-06-17
Fix corruption of FITS tables with unsigned int columns on save and load

eslavich:
Addresses #3655 

The problem on save turned out to be less of an astropy bug and more of a missing feature -- the public API doesn't currently allow us to put a FITS_rec into "pseudo unsigned integer" mode.  We were already messing with astropy implementation details by setting `_colrefs` directly, so I don't think it's a big deal to also set `_uint`.

The problem on load is, I think, an astropy bug.  I filed an issue and added a workaround here, which is to overwrite the incorrect FITS_rec dtype.

### DIFF
--- a/jwst/combine_1d/combine1d.py
+++ b/jwst/combine_1d/combine1d.py
@@ -58,6 +58,7 @@ class InputSpectrumModel:
             self.surf_bright = np.zeros_like(self.flux)
             self.sb_error = np.zeros_like(self.flux)
             log.warning("There is no SURF_BRIGHT column in the input.")
+        self.dq = spec.spec_table.field("dq").copy()
         self.nelem = self.wavelength.shape[0]
         self.unit_weight = False        # may be reset below
         self.right_ascension = np.zeros_like(self.wavelength)
--- a/jwst/combine_1d/combine1d.py
+++ b/jwst/combine_1d/combine1d.py
@@ -58,11 +58,6 @@ class InputSpectrumModel:
             self.surf_bright = np.zeros_like(self.flux)
             self.sb_error = np.zeros_like(self.flux)
             log.warning("There is no SURF_BRIGHT column in the input.")
-        # xxx temporary self.dq = spec.spec_table.field("dq").copy()
-        # xxx This is a workaround which should be deleted after the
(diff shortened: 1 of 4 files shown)


jwst/combine_1d/combine1d.py
```diff
@@ -58,11 +58,6 @@ def __init__(self, ms, spec, exptime_key):
             self.surf_bright = np.zeros_like(self.flux)
             self.sb_error = np.zeros_like(self.flux)
             log.warning("There is no SURF_BRIGHT column in the input.")
-        # xxx temporary self.dq = spec.spec_table.field("dq").copy()
```
philhodge: Only the "# xxx temporary " part of this line should be deleted.  The actual assignment to self.dq needs to be done.

jwst/datamodels/properties.py
```diff
@@ -56,8 +56,12 @@ def _as_fitsrec(val):
         return val
     else:
         coldefs = fits.ColDefs(val)
+        uint = any(c._pseudo_unsigned_ints for c in coldefs)
         fits_rec = fits.FITS_rec(val)
         fits_rec._coldefs = coldefs
+        # FITS_rec needs to know if it should be operating in pseudo-unsigned-ints mode,
+        # otherwise it won't properly convert integer columns with TZEROn before saving.
+        fits_rec._uint = uint
```
philhodge: Is this used in astropy.io.fits?

jwst/combine_1d/combine1d.py
eslavich: Oh!  Oops, thanks.

jwst/datamodels/properties.py
eslavich: Yes, [here](https://github.com/astropy/astropy/blob/de3933688433e22765b1bdffa175fcd954d691c3/astropy/io/fits/fitsrec.py#L940).  The else... is where it was changing the dtype to float64.

jwst/combine_1d/combine1d.py
eslavich: (this has been fixed)

hbushouse:
Shall we go ahead and merge this?

PR #3687, opened 2019-06-18
start change log for 0.13.5

nden:


hbushouse:
Once we know we've made our final B7.3 delivery to DMS, we should probably update to 0.14.x, rather than continuing the 0.13.x series.

jdavies-st:
More about that.  If we move from `relic` to using `setuptools_scm` to set our package version, I believe we don't have to do post-release "alpha" tags so that the version generated from `git describe` maps to a dev version of a future release.  It increments from the release tag automatically.  I.e. we don't have to anticipate what our future release will be numbered.

So we would create the tag `0.13.4` on our master branch, and only that tag.  No `0.13.5a` tag.  If a new commit was then put on master, the version auto-generated would be something like `0.13.5.dev1+g9e985edb`.

Issue #3575, opened 2019-06-03
Replace jwst.tweakreg with tweakwcs


stscijgbot:
Issue [JP-756](https://jira.stsci.edu/browse/JP-756) was created by Nadia Dencheva:

Reduce code duplication.

stscijgbot:
Comment by Nadia Dencheva: Make sure tweakwcs does not bring dependence on stsci.imagestats.

stscijgbot:
Comment by Rosa Diaz: Is it just a name change or code change?

stscijgbot:
Comment by Nadia Dencheva: Tweakwcs was based on jwst.tweakreg and as such is mostly code duplicate. Tweakwcs is currently more general and already in use so bugs get reported and fixed there. It makes sense to start using it instead of fixing the same bugs in two places.

jdavies-st:
`tweakwcs` also has ~100% code test coverage.  `jwst.tweakreg` does not.

nden:
Fixed in #3689

stscijgbot:
Comment by Rosa Diaz: This ticket was closed; though  I have a couple of questions:

1) rather than closing it, shouldn't it be moved to ready to test and assign it to Alicia or Maca?

2) it does not indicate which version of the calibration pipeline it affects. This information is essential to identify if the DMSWG needs to test it.

jdavies-st:
2) This is in `dev`, so not released yet.

PR #3689, opened 2019-06-18
Switch to using tweakwcs for alignment algorithm

mcara:
This PR addresses the issue described in https://github.com/spacetelescope/jwst/issues/3575 Fundamentally, `tweakwcs` is used to perform WCS alignment. Currently `tweakwcs` has almost 100% unit test coverage. This PR minimized maintenance efforts.


jdavies-st:
Need to add `tweakwcs` to the `install_requires` in `setup.py`.

nden:
As far as I can see `tweakreg` is called in calimage3. I don't see a test for calimage3 for any of the instruments. Did I miss one? How do we know this runs?

mcara:
@jdavies-st Thanks! I will look into it.

mcara:
`tweakwcs` sets the "fit status" (`fit_info['status']`) to several possible values:

- `'SKIPPED'`: if an image WCS was not modified/aligned for example because it does not overlap with any other image;
- `'FAILED: reason'`: unable to align an image, i.e., because there are too few matches;
- `'REFERENCE'`: not aligned because it was used as a reference image to which all other images were aligned;
- `'SUCCESS'`: image was aligned successfully to a reference image/catalog.

In this PR I assumed that `S_TWKREG` should be set to `'COMPLETE'` only  when a WCS was modified => when `fit_info['status']` is `'SUCCESS'`. If an image was used as a reference image, then its status now will be `'SKIPPED'`.

What is the purpose of `S_TWKREG`? @jdavies-st @hbushouse I can modify this PR to set status to `'SUCCESS'` unless `fit_info['status']` is `FAILED`.

hbushouse:
The `S_TWKREG` is intended to indicate simply whether the step was run at all for a given image. Hence even if the fit failed and/or the WCS was not updated, `S_TWKREG` should still be set to "COMPLETE". It would only be set to "SKIPPED" if the step had to actually abort processing for some reason. We should consider adding a new keyword to document the "fit status" for each image. In fact the Cal WG recently filed a ticket requesting some kind of information just like this as to which images that contribute to a combined/drizzled product did or did not have useful information from which their WCS was updated.

jdavies-st:
Running the NIRCam regression test `image3_pipeline2` and looking at the resulting catalog of the drizzled product, the RA/Dec position differences between the old `tweakreg` and this PR is ~0.01 arcsec between the ~50 matched sources in the 2 catalogs.  Does this seem reasonable?

Almost all the pixel values are different, but I figured a test of position of sources would be good.  Would it be worthwhile checking photometry as well?

jdavies-st:
Ok, a quick check of the flux indicates the differences are on the order of ~2% between the old `tweakreg` and this PR for the matched sources.

mcara:
The differences in tests (see https://github.com/spacetelescope/jwst/pull/3689#issuecomment-506508563) turn out to be due to factors: a bug in this PR (fixed in the second commit) due to which `searchrad` was used for `separation` and due to changes how in how clipping works in `tweakwcs` (using RSME value instead of RMS for stopping criterion). I have also added functionality that makes catalog/image/group naming more convenient than before (at least a part of the original file name is used to identify catalogs).

jdavies-st:
Looks good.  Why are there 96 commits?  Perhaps clean up the branch so it just has your new commits?

setup.py
```diff
@@ -200,6 +199,7 @@ def run_tests(self):
         'stsci.image>=2.3',
         'stsci.imagestats>=1.4',
         'stsci.stimage>=0.2',
+        'tweakwcs>=0.5.1',
```
jdavies-st: `stsci.stimage` can be removed as a direct dependency as well.  It is no longer a dependency of `jwst`, but is now a dependency of `tweakwcs`.

Issue #3690, opened 2019-06-19
WFSC association check way too restrictive


stscijgbot:
Issue [JP-804](https://jira.stsci.edu/browse/JP-804) was created by Jonathan Eisenhamer:

The WFSC association was recently changed to be more restrictive on visit type. Have confirmed with the WFSC group that this was not correct and will need to revert.

stscijgbot:
Comment by Howard Bushouse: Fixed by [https://github.com/spacetelescope/jwst/pull/3691]

PR #3691, opened 2019-06-19
Loosen the WFSC constraints and only exclude NIS_EXTCAL

stscieisenhamer:
Revert an over-strict test on WFSC associations.


stscieisenhamer:
The travis failure was a time limit exceeded issue in a linearity test. Since jenkins passed, merging is happening.

stscieisenhamer:
Resolves #3690/JP-804

hbushouse:
Is there any benefit to running the regression test suite on this change before we make the 0.13.5 release? Or do the unit tests already provide sufficient ASN coverage?

Issue #2927, opened 2018-12-07
NIRSpec internal lamp processing specifications


stscijgbot:
Issue [JP-473](https://jira.stsci.edu/browse/JP-473) was created by James Muzerolle:

The attached file contains the NIRSpec team's recommendations on the processing flow for each type of internal lamp (both via the lamp template and AUTOCALS).

[^lamp_processing_forDMS.pdf]

stscijgbot:
Comment by Rosa Diaz: [~bushouse], we have set up some functionality for this already. Can you let me know which cases we already cover? We also need to doublecheck that the reference files functionality is in place too.

stscijgbot:
Comment by Howard Bushouse: Example data for testing this are contained in GSEG3:

jw00620-o005_20190529t185430_nrslamp-spec2_003 LEVEL_2B none

    /ifs/archive/test/jwst/test1/info/owl/logs/rspencer_jw00620-o005_20190529t185430_nrslamp-spec2_003_1559158108.311733/ALOG_1559158356_level_2b_jw00620-o005_20190529t185430_nrslamp-spec2_003.err

            raise ValueError('Missing extraction parameters.')

        ValueError: Missing extraction parameters.

            'One or more products failed to process. Failing calibration.'

 

It should be accessible from owld-2 or other dstring machines.

stscijgbot:
Comment by Rosa Diaz: I see errors. Does that mean this has not been implemented?

stscijgbot:
Comment by Howard Bushouse: That's correct, which is why the status of this ticket is still set to OPEN and the resolution is Unresolved. It's assigned to be worked in our current sprint, so we're hoping to get it into B7.3 yet.

stscijgbot:
Comment by Robert Jedrzejewski: [~muzerol] the LAMPMODE keyword can take the values BRIGHTOBJ, FIXEDSLIT, GRATING-ONLY, IFU, MSASPEC and NULL.  I understand that BRIGHTOBJ should be treated like FIXEDSLIT, but what should we do with GRATING-ONLY data?

stscijgbot:
Comment by James Muzerolle: GRATING_ONLY is a limited special case that is used to move the grating wheel without taking any exposures (no lamp is ever turned on).  You can ignore it for the purposes of processing lamp data.

stscijgbot:
Comment by Robert Jedrzejewski: OK, thanks

stscijgbot:
[...]

PR #3693, opened 2019-06-19
Remove the Lv2NRSLAMPImage rule

stscieisenhamer:
As per new information in JP-473/JP-282, the image processing of NRS_LAMP should be removed. This removes the association rule that would have setup for NRS_LAMP.

### DIFF
--- a/jwst/associations/lib/rules_level2b.py
+++ b/jwst/associations/lib/rules_level2b.py
@@ -412,51 +412,6 @@ class Asn_Lv2MIRLRSFixedSlitNod(
 
 
 @RegistryMarker.rule
-class Asn_Lv2NRSLAMPImage(
-        AsnMixin_Lv2Special,
-        DMSLevel2bBase
-):
-    """Level2b NIRSpec Image Lamp calibrations Association
-
-    Characteristics:
-        - Association type: ``nrslamp-image2``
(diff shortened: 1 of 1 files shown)


stscieisenhamer:
Note: Regression tests will fail due to reduced association counts requiring updates to tests and truth files. For expediency, these failures can be ignored until such truth updates are done.

stscieisenhamer:
Local unit tests have passed. Merging up now.

hbushouse:
Should've asked earlier: are you sure that this will completely eliminate any FILTER=OPAQUE + GRATING=MIRROR exposures from showing up in ANY kind of level-2 ASN, as opposed to just eliminating the creation of "nrslamp-image2" ASNs?

stscieisenhamer:
I believe so. This was the reason for the explicit rule; otherwise nothing would have happened.  I have run the regressions over the SDP & test pools that we have collected, and, besides the loss of nrslamp-image2, nothing else has changed.

stscieisenhamer:
Resolves [JWSTDMS-2](https://jira.stsci.edu/browse/JWSTDMS-2) with respect to the nrslamp-image2 associations.
WIP #2927/JP-473

hbushouse:
Fixes #3696

PR #3704, opened 2019-06-20
Add new unit tests for flat_field

### LINKED ISSUE #3702: Add new unit tests for flat_field
Issue [JP-807](https://jira.stsci.edu/browse/JP-807) was created by Philip Hodge:

A few more unit tests for the flat_field step would be useful.

philhodge:
New unit tests were added for `flat_field`, and some tests were split into multiple functions.

Closes #3702

### DIFF
--- a/jwst/flatfield/tests/test_g_average.py
+++ b/jwst/flatfield/tests/test_g_average.py
@@ -4,36 +4,10 @@ Test for flat_field.g_average
 import math
 
 import numpy as np
+from scipy.integrate import quad
+from astropy.modeling import polynomial
 
 from jwst.flatfield.flat_field import g_average
-
-def evaluate_polynomial(x, coeff):
-    """Evaluate a polynomial."""
-    n = len(coeff)
-    sum = np.zeros_like(x) + coeff[0]
-    for i in range(1, n):
-        sum = sum * x + coeff[i]
-    return sum
-
-
-def integrate_polynomial(lower, upper, coeff):
-    """Integrate a polynomial."""
-
-    # Compute the indefinite integral by modifying the coefficients,
-    # setting the constant term to 0.
-    n = len(coeff)
-    icoeff = np.zeros(n + 1, dtype=np.float64)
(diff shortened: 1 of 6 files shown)


jwst/flatfield/tests/test_g_average.py
```diff
@@ -0,0 +1,75 @@
...
+import numpy as np
+
+from jwst.flatfield.flat_field import g_average
+
+def evaluate_polynomial(x, coeff):
+    """Evaluate a polynomial."""
+    n = len(coeff)
+    sum = np.zeros_like(x) + coeff[0]
+    for i in range(1, n):
+        sum = sum * x + coeff[i]
```
hbushouse: Shouldn't the "*" and "+" operators be swapped in this expression, i.e. multiply x by coeff[i] and add the result to sum?

philhodge: No, but I do have the coefficients in an unusual order.  Here is an example of what I was thinking:

    sum = c[0]
    sum = c[0] * x + c[1]
    sum = (c[0] * x + c[1]) * x + c[2]
    sum = (c[0] * x**2 + c[1] * x + c[2]) * x + c[3]
           = c[0] * x**3 + c[1] * x**2 + c[2] * x + c[3]

I think this is consistent with the way I took the integral by modifying the coefficients and evaluating the polynomial, but I'll do some further checking.

hbushouse: Oh, I see. OK, I guess that makes sense when it's spelled out in detail like that.

nden:
It looks fine but I wonder why you did not use existing functions for polynomial evaluation and integration, e.g. he ones available in numpy. I think we should try to reuse as much code as possible. If nothing else it reduces the code review time.

nden:
Using modeling is fine, just want to mention that numpy also has polynomial functions now.

`np.polynomial.polynomial.polyval`, `np.polynomial.polynomial.polyint` and others

PR #3705, opened 2019-06-20
Changes necessary to support jsonschema 3.x

eslavich:
It's been [pointed out](https://github.com/spacetelescope/asdf/issues/530#issuecomment-503811670) that jsonschema 2.6 is incompatible with Python 3.8, so at some point we'll need to upgrade.  These changes make jwst compatible with both jsonschema 2.6 and 3.0.

### DIFF
--- a/jwst/associations/lib/asn_schema_jw_level2b.json
+++ b/jwst/associations/lib/asn_schema_jw_level2b.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "http://json-schema.org/schema#",
+    "$schema": "http://json-schema.org/draft-04/schema",
     "title": "Association: JWST DMS Level2a->Level2b",
     "description": "The data structure that, within the JWST DMS processing system, defines how to group level2a exposures to produce Level2b data products.",
     "type": "object",
--- a/jwst/datamodels/fits_support.py
+++ b/jwst/datamodels/fits_support.py
@@ -293,7 +293,7 @@ def _fits_item_recurse(validator, items, instance, schema):
 def _fits_type(validator, items, instance, schema):
     if instance in ('N/A', '#TODO', '', None):
         return
-    return validators._validators.type_draft4(validator, items, instance, schema)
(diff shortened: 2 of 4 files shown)


jwst/datamodels/fits_support.py
```diff
@@ -293,7 +293,7 @@ def _fits_item_recurse(validator, items, instance, schema):
 def _fits_type(validator, items, instance, schema):
     if instance in ('N/A', '#TODO', '', None):
         return
-    return validators._validators.type_draft4(validator, items, instance, schema)
```
eslavich: This method disappears in jsonschema 3.0, but luckily there is a way to access it through a public API that does not change between the versions.

jwst/associations/lib/asn_schema_jw_level2b.json
```diff
@@ -1,5 +1,5 @@
 {
-    "$schema": "http://json-schema.org/schema#",
```
eslavich: This causes jsonschema to use the latest draft, whatever that is, which is a deprecated behavior in jsonschema 3.0.  Also, the latest draft is now draft-07, whereas in jsonschema 2.6 it was draft-04 anyway.

jwst/associations/lib/asn_schema_jw_level2b.json
```diff
@@ -1,5 +1,5 @@
...
+    "$schema": "http://json-schema.org/draft-04/schema",
```
hbushouse: So we'll have to manually update this whenever we want to start using a new draft?

eslavich: Yes that's correct.  Incidentally, later versions aren't backwards compatible with draft-04, so we'll need to audit these files anyway.

jwst/associations/lib/asn_schema_jw_level2b.json
nden: ping @stscieisenhamer

nden:
don't forget the change log

nden:
@eslavich Could you fix the conflict?

PR #3708, opened 2019-06-20
Update README.md with correct DMS release info

jdavies-st:
Got ahead of myself.


nden:
>  Got ahead of myself.


may be not ...

nden:
We can probably skip 0.13.5 and go directly to 0.13.7. whether it's final ... remains to be seen.

PR #3716, opened 2019-06-21
Possible fix to spec_wcs being fooled to use quantities

nden:


nden:
Regression tests pass. The problem is in gwcs.utils.isnumeric. It needs to be enhanced to capture more dtypes.

jwst/combine_1d/combine1d.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, ms, spec, exptime_key):
             to use weight = 1.
         """
 
-        self.wavelength = spec.spec_table.field("wavelength").copy()
+        self.wavelength = spec.spec_table.field("wavelength").copy().astype(np.float64)
```
hbushouse: Is there something special about the wavelength column in the table that needs this fix in order to avoid the problem in gwcs, but meanwhile other float64 columns like flux and error (below) are not affected?

nden: gwcs naively checks for dtype of `float64`. The dtype of the column is `>f8` and so it's missed and considered "not numeric" which triggers gwcs to use quantities. This issue needs to be fixed in gwcs but it will take longer to do it there.

nden: Only the wavelength column is passed to gwcs as input.

PR #3719, opened 2019-06-21
Jp 681

ddavis-stsci:
Updates so that only nodded observations have x1d products listed as background  & if there is only one science member do not have it be its own background.

### DIFF
--- a/jwst/associations/lib/rules_level3.py
+++ b/jwst/associations/lib/rules_level3.py
@@ -207,7 +207,7 @@ class Asn_SpectralTarget(AsnMixin_Spectrum):
             Constraint(
                 [
                     DMSAttrConstraint(
-                        name='patttype',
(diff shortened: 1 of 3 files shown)


ddavis-stsci:
Note: The unit test test_level3_spectrographic.py is failing. I think that the new association rules are generating an additional level3 association and the test is only expecting two. 
`AssertionError: /Users/ddavis/src/JWST/scsb/jwst/jwst/associations/tests/data/pool_007_spec_miri.csv: n_asns not expected 2 3`

ddavis-stsci:
Program 00623  with MIRI LRS data still produces an association with 12 _cal products as science and 12 _x1d products listed as background.

nden:
Are these failures expected?

stscieisenhamer:
This one dropped from my radar. Will look today.

PR #3725, opened 2019-06-21
Skip really slow FGS detector1 regtest

hbushouse:
One `calwebb_detector1` regression test using an FGS image-mode exposure as input consistently takes ~3 hours to execute, which is 1-2 hours longer than all of the remaining regression tests combined. This test needs to be reworked or we need to find better data to use, so that our regression test runs are not hostage to this one test. I'm marking it to be skipped for now, until something can be done to fix or replace it.

### DIFF
--- a/jwst/tests_nightly/general/fgs/test_fgs_sloper_1.py
+++ b/jwst/tests_nightly/general/fgs/test_fgs_sloper_1.py
@@ -4,6 +4,7 @@ from jwst.pipeline.calwebb_detector1 import Detector1Pipeline
 from jwst.tests.base_classes import BaseJWSTTest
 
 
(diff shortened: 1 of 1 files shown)


hbushouse:
It only takes ~10 minutes to execute all of the `calwebb_detector1` steps up through `jump`. All of the remaining time is then spent in `ramp_fit`. The dataset is 2048x2048 (full-frame) x 10 groups x 5 integrations. The data appear to be really bad, because the test already has the `jump` rejection_threshold set to 250 sigma, in order to avoid excessive outlier flagging. The bad data are likely then also the cause of the very long fitting times in `ramp_fit`.

PR #3736, opened 2019-06-25
Fix corruption of FITS tables with unsigned int columns on save and load

eslavich:
This was originally merged as #3680, but reverted in #3717 when it broke a regression test.  This PR revives the original changes, with the addition of a fix for the busted test.

I confirmed that `test_miri_masterbg_mrs_dedicated` passes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@
 datamodels
 ----------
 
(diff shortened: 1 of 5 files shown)


hbushouse:
Jenkins failure appears to be unrelated.

PR #3747, opened 2019-06-26
Add a requirements-sdp.txt file and new Jenkinsfiles to run on weekends

jdavies-st:
- Run both unit and regression tests in our `JenkinsfileRT` runs.  This will eventually enable us to track joint test coverage.
- Add `requirements-sdp.txt` with pinned dependencies.  The purpose of this is to run our main regression tests from it eventually and to keep our dependencies stable during the development cycle for in-house SDP processing.
- Add `JenkinsfileRT-dev` for regression testing against dev dependencies on the weekends
- Add `JenkinsfileRT-sdp` for regression testing using `requirements-sdp.txt` as either our daily test run or another weekend test run.
- Remove `verhawk` as a package dependency.  It is not used.
- Use `python 3.7` in regression testing.

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -28,7 +28,7 @@ data_config.server_id = 'bytesalad'
 data_config.root = '${pytest_basetemp}'
 data_config.match_prefix = '(.*)_result' // .json is appended automatically
 
-// Build and test with python 3.6 and released dependencies from astroconda
+// Build and test
 bc = new BuildConfig()
 bc.nodetype = 'jwst'
 bc.name = 'stable'
--- a/JenkinsfileRT-sdp
+++ b/JenkinsfileRT-sdp
@@ -6,7 +6,7 @@ jobconfig.enable_env_publication = true
 jobconfig.publish_env_on_success_only = true
 
 // Define python version for conda
-python_version = '3.7'
+python_version = '3.6'
 
 // pip related setup
 def pip_index = "https://bytesalad.stsci.edu/artifactory/api/pypi/datb-pypi-virtual/simple"
@@ -28,7 +28,7 @@ data_config.server_id = 'bytesalad'
 data_config.root = '${pytest_basetemp}'
 data_config.match_prefix = '(.*)_result' // .json is appended automatically
 
-// Build and test with python 3.6 and released dependencies from astroconda
+// Build and test
 bc = new BuildConfig()
 bc.nodetype = 'jwst'
 bc.name = 'sdp'
--- a/JenkinsfileRT-sdp
+++ b/JenkinsfileRT-sdp
@@ -3,7 +3,7 @@ if (utils.scm_checkout(['skip_disable':true])) return
 
 jobconfig = new JobConfig()
 jobconfig.enable_env_publication = true
-jobconfig.publish_env_on_success_only = false
+jobconfig.publish_env_on_success_only = true
 
 // Define python version for conda
 python_version = '3.7'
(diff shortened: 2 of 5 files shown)


jdavies-st:
Will run some sample Jenkins builds using these new Jenkinsfiles and will record the results here.

JenkinsfileRT
```diff
@@ -25,26 +25,26 @@ pytest_basetemp = "test_outputs"
...
-bc.name = 'stable-deps'
+bc.name = 'stable'
 bc.env_vars = env_vars
-bc.conda_ver = '4.6.14'
+bc.conda_ver = '4.7.5'
 bc.conda_packages = [
     "python=${python_version}",
 ]
 bc.build_cmds = [
+    "pip install numpy",
```
hbushouse: I thought line 42 was supposed to install absolutely everything that's needed for the jwst package. I guess there's something unique about numpy and the order in which it gets installed?

JenkinsfileRT-sdp
```diff
@@ -0,0 +1,52 @@
+// [skip ci] and [ci skip] have no effect here.
+if (utils.scm_checkout(['skip_disable':true])) return
+
+jobconfig = new JobConfig()
+jobconfig.enable_env_publication = true
+jobconfig.publish_env_on_success_only = true
+
+// Define python version for conda
+python_version = '3.7'
```
hbushouse: I believe that as of today DMS is still using 3.6.x for all subsystems.

JenkinsfileRT-sdp
```diff
@@ -0,0 +1,52 @@
...
+// Set pytest basetemp output directory
+pytest_basetemp = "test_outputs"
+
+// Configure artifactory ingest
+data_config = new DataConfig()
+data_config.server_id = 'bytesalad'
+data_config.root = '${pytest_basetemp}'
+data_config.match_prefix = '(.*)_result' // .json is appended automatically
+
+// Build and test with python 3.6 and released dependencies from astroconda
```
hbushouse: But line 9 is setting python to 3.7, not 3.6

requirements-sdp.txt
```diff
@@ -0,0 +1,14 @@
+asdf==2.3.3
```
hbushouse: So does this entire file need to be reviewed and updated periodically when any versions change in the DMS environment?

stscicrawford: Yes, that would be the idea.  We will only move forward on versions in the pipeline when we decide to, rather than having conda introduce an update outside of our control.    This does mean that at the start of a build that this file should be reviewed and updated accordingly.

JenkinsfileRT
jdavies-st: It does install numpy, but it bootstraps it.  Because currently `spherical_geometry`, and a two of the `stsci.foo` packages need `numpy` but don't currently specify that in a way that works, they fail on the initial installation, but everything else gets installed, including `numpy`, then `pip` bootstraps back to those 3 failed installations and installs them again, successfully.  When new releases of these packages specifying `numpy` in a `pyproject.toml` file appear, this shouldn't be necessary.  This basically just speeds up the installation, maybe.

JenkinsfileRT-sdp
jdavies-st: Right.  We will want to switch to 3.7.x eventually when we deliver B7.4 though, right?  I guess we should discuss when we should make this switchover for our own development and testing, and on which test runs.  The dailies or just the weekend runs?

jdavies-st:
Test run using `JenkinsfileRT-sdp`

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/123/

requirements-sdp.txt
jdavies-st: We will be testing both the pinned dependencies in the `requirements` file and the rolling, released dependencies.  How often we want to do each may depend on which part of the dev cycle we're in.

It may be that we want to not use a `requirements` file now for the daily builds, but use one for the build on the weekends.  And then switch that up as we get a month or so out from when we start having to have our first rc1 ready for Build 7.4.  But we can discuss this further and see what gives us the most flexibility and stability.

PR #3756, opened 2019-07-01
A function to determine the dispersion direction has been written.

### LINKED ISSUE #3619: Write a function to obtain the dispersion direction
Issue [JP-774](https://jira.stsci.edu/browse/JP-774) was created by Philip Hodge:

For spectroscopic data, there should be a function to determine the dispersion direction, based on instrument mode.

philhodge:
Closes #3619


jwst/lib/dispaxis.py
```diff
@@ -0,0 +1,152 @@
...
+    int : The dispersion direction
+
+-       0  the dispersion direction is not meaningful or not defined
+-       1  the dispersion direction is horizontal ("sky" coordinates)
+-       2  the dispersion direction is vertical ("sky" coordinates)
+    """
+
+    exposure_type = exposure_type.upper()
+    grating = grating.upper()
+    filter = filter.upper()
```
nden: It's a minor thing but could you change the name of this variable since it's a reserved python keyword.

jwst/lib/dispaxis.py
```diff
@@ -0,0 +1,152 @@
...
+        # MIRI
+        "MIR_4QPM": 0,
+        "MIR_CORONCAL": 0,
+        "MIR_DARKALL": 0,
+        "MIR_DARKIMG": 0,
+        "MIR_DARKMRS": 0,
+        "MIR_FLATALL": 0,
+        "MIR_FLATIMAGE": 0,
+        "MIR_FLATIMAGE-EXT": 0,
+        "MIR_FLATMRS": 2,
```
nden: Should this be 0? It's not going through assign_wcs.

jwst/lib/dispaxis.py
```diff
@@ -0,0 +1,152 @@
...
+        "MIR_FLATMRS-EXT": 2,
```
nden: not going through assign_wcs either

jwst/lib/dispaxis.py
```diff
@@ -0,0 +1,152 @@
...
+        "NIS_SOSS": 1,
+        "NIS_TACQ": 0,
+        "NIS_TACONFIRM": 0,
+        "NIS_WFSS": (exposure_type, "ANY", filter, "ANY"),
+        # NIRCam
+        "NRC_CORON": 0,
+        "NRC_DARK": 0,
+        "NRC_FLAT": 0,
+        "NRC_FOCUS": 0,
+        "NRC_GRISM": (exposure_type, "ANY", "ANY", pupil),
```
nden: Is this a valid mode? It's marked as "not_implemented" in assign_wcs.

jwst/lib/dispaxis.py
```diff
@@ -0,0 +1,152 @@
...
+        "NRC_WFSS": (exposure_type, "ANY", "ANY", pupil),
+        "NRC_LED": 0,
+        "NRC_WFSC": 0,
+        "NRC_TACONFIRM": 0,
+        "NRC_TACQ": 0,
+        "NRC_TSGRISM": (exposure_type, "ANY", "ANY", pupil),
+        "NRC_TSIMAGE": 0,
+        # NIRSpec
+        "NRS_AUTOFLAT": 0,
+        "NRS_AUTOWAVE": 1,
```
nden: This is currently not processed by assign_wcs. Should it?

jwst/lib/dispaxis.py
hbushouse: Given that it's not going through assign_wcs probably implies that we don't need to cover this mode (it'll never go through any level-2b processing), but I think the setting is still accurate (i.e. the data in the flat will be dispersed along axis 2).

jwst/lib/dispaxis.py
hbushouse: It's now an obsolete EXP_TYPE value, having been replaced by NRC_WFSS. Doesn't hurt to keep it in the list for backwards compatibility.

jwst/lib/dispaxis.py
philhodge: Yes; thanks for pointing this out.

jwst/lib/dispaxis.py
philhodge: I doubt that this should be processed by assign_wcs.

jwst/lib/dispaxis.py
hbushouse: Is this a potential problem throughout all the cal pipeline code? I'm sure the variable name "filter" must be used other places, given what a common thing it is.

jwst/lib/tests/test_dispaxis.py
```diff
@@ -0,0 +1,85 @@
+"""
+Test for dispaxis
+"""
+
+from .. import dispaxis
+
+def test_dispaxis_1():
+    value = dispaxis.get_dispersion_direction("FGS_IMAGE", "junk", "junk",
+                                              "junk")
+    assert value == 0
```
hbushouse: Need to update the "0" to "None". Same for other instances that used to be zero.

PR #3765, opened 2019-07-01
Jp825 Ensure all target acqs are processed by Level 2

stscieisenhamer:
Resolves [JP-825](https://jira.stsci.edu/browse/JP-825)/#3748


jwst/associations/lib/diff.py
```diff
@@ -11,6 +11,47 @@
...
+class DuplicateProductError(DiffError):
+    """Duplicate products found"""
+
+
+class MemberMismatchError(DiffError):
+    """Membership does not match"""
+
+
+class TypeMismatchError(DiffError):
+    """Associaiton type mismatch"""
```
hbushouse: Typo in "Associa**it**on"

jwst/associations/lib/dms_base.py
```diff
@@ -43,30 +43,30 @@
...
-    'nrc_taconfirm':     'target_acquistion',
+    'nrc_tacq':          'target_acquisition',
+    'nrc_taconfirm':     'target_acquisition',
     'nrs_autoflat':      'autoflat',
     'nrs_autowave':      'autowave',
-    'nrs_confirm':       'target_acquistion',
+    'nrs_confirm':       'target_acquisition',
     'nrs_dark':          'dark',
     'nrs_focus':         'engineering',
     'nrs_image':         'engineering',
```
hbushouse: NIRSpec imaging exposures are supposed to receive normal ``calwebb_image2`` processing and hence need to be included as science members in a level-2b ASN. What effect will this value of "engineering" have on them?

jwst/associations/lib/dms_base.py
```diff
@@ -90,21 +90,13 @@
...
-    'mir_tacq',
     'nis_focus',
-    'nis_tacq',
-    'nis_taconfirm',
-    'nrc_tacq',
-    'nrc_taconfirm',
     'nrc_focus',
-    'nrs_confirm',
     'nrs_focus',
     'nrs_image',
```
hbushouse: EXP_TYPE's like "_focus", "_mimf", and "_image" should be processed as normal images, hence they should appear as science type members in a level-2b ASN for the purpose of ``calwebb_image2`` processing. The list here is labeled as "NONSCIENCE". Will they be treated as science types for the purpose of creating image2 type ASN's?

jwst/associations/lib/dms_base.py
stscieisenhamer: Part 1 of answer: Will need to dig up the conversations, but why NRS_IMAGE is marked `engineering` was because it would not be used for science. If that is not the case, this will be fixed.

Regardless of Part 1, the answer to the question is Part 2: See the next comment...

jwst/associations/lib/dms_base.py
stscieisenhamer: All EXP_TYPE's found in `IMAGE2_NONSCIENCE_EXP_TYPES` do get Level2b processing through the rule `jwst.associations.lib.rules2b.Asn_Lv2ImageNonScience`. However, these exposures will not get things like backgrounds or imprint applied to them as "science" exposures would.  They will always be processed on their own.

stscieisenhamer:
@hbushouse are the responses to your commentary OK? If so, please approve.

PR #3769, opened 2019-07-02
Set pytest caplog to WARNING or greater in JenkinsfileRT

jdavies-st:
Logs are too long when set to DEBUG level as is default in pytest.  Setting the level to WARNING will still print WARNING, ERROR and CRITICAL log messages in the automated runs, but not DEBUG or INFO log messages.

Should make the test results much easier to read, and also for tests that have very verbose DEBUG logging, make the XML results file a reasonable size so Jenkins can slurp it up.


jdavies-st:
TravisCI is very slow today.  Jenkins CI passes.  Merging.

PR #3800, opened 2019-07-10
Fix bug in okify script when fitsdiff fails to write version description

eslavich:
https://jira.stsci.edu/browse/JP-856

I noticed that one of the regression tests was not okified by the new script.  It turns out fitsdiff may omit the version description line that I used to break up the output by file, and when that happens the script was failing to query the user about one of the files.

This PR fixes that issue.  I tested it on last night's test output and it was able to recognize the busted file.

### DIFF
--- a/scripts/okify_regression_tests.py
+++ b/scripts/okify_regression_tests.py
@@ -16,9 +16,9 @@ import os
 from argparse import ArgumentParser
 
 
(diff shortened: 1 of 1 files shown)


eslavich:
As it stands, the script won't find anything except complaints from fitsdiff.  I can look into also handling the plain text diff.

eslavich:
@hbushouse the latest comment adds support for output from unified_diff.

PR #3803, opened 2019-07-11
JP-842 Update exp_to_source docs

### LINKED ISSUE #3773: Update exp_to_source docs to give more details for different modes
Issue [JP-842](https://jira.stsci.edu/browse/JP-842) was created by Howard Bushouse:

The docs for the exp_to_source step in calwebb_spec3 are pretty minimal right now. Need to update to include a lot more details.

hbushouse:
Updated the `exp_to_source` docs to give a lot more details as to how things work for various modes, and examples of reorganized file names. Also fixed some broken internal references in the `gain_scale` doc. Also added a little bit of logging info to the `exp_to_source` step itself (it's nice to be able to see what's happening, instead of just staring at a blank screen).

Fixes #3773

### DIFF
--- a/docs/jwst/data_products/file_naming.rst
+++ b/docs/jwst/data_products/file_naming.rst
@@ -1,5 +1,10 @@
-File Naming Scheme
-------------------
+File Naming Schemes
+-------------------
+
+.. _exp_file_names:
+
+Stage 0 - 2 Naming
+^^^^^^^^^^^^^^^^^^
 The FITS file naming scheme for Stage 0, 1, and 2 "exposure-based" products is::
 
  jw<ppppp><ooo><vvv>_<gg><s><aa>_<eeeee>(-<"seg"NNN>)_<detector>_<prodType>.fits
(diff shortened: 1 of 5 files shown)


docs/jwst/data_products/file_naming.rst
```diff
@@ -1,5 +1,10 @@
-File Naming Scheme
-------------------
+File Naming Schemes
+-------------------
+
+.. _exp_file_names:
```
philhodge: The underscores here (and for src_file_names) don't look like ASCII.  Is this OK?

hbushouse: I think it might just be a weird way that it's being displayed. They are in fact ordinary ASCII underscores in the raw file and the links work.

hbushouse: I think github might be interpreting the leading underscore to mean turn on underlining for the following words, so that's why it looks that way.

PR #3804, opened 2019-07-11
JP-349 DQ Flagging for IFU cubes

jemorrison:
The PR sets the DQ plane of the IFU cube. The values it sets are:

- 0: good data
- 1: DO_NOT_USE (Hole - should have data but does not)
- 1 + 512 (NON_SCIENCE): no overlapping data with IFU cube

The number of holes for each IFU cube is printed in the logging.
This PR also fixed an error in CRVAL3 (it was off by 1/2 a pixel).

Messages are now printed if data falls outside the min and max wavelength (which is set by the cubepars reference file).

Note: A future update for NIRSpec will use the min and max wavelength of data to set the IFU wavelength ranges.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -713,6 +713,8 @@ csv_tools
 cube_build
 ----------
 
+- Added dq flagging [#3804]
+
 cube_skymatch
 -------------
 
--- a/jwst/cube_build/ifu_cube.py
+++ b/jwst/cube_build/ifu_cube.py
@@ -45,6 +45,9 @@ class IFUCubeData():
         self.new_code = 0
         self.input_filenames = input_filenames
         self.pipeline = pipeline
+        
+        print('defining cube',list_par1,list_par2)
+
 
         dq_file = 'cube_spaxel_dq'+str(list_par1[0]) +'.results'
         self.spaxel_dq_file =  open(dq_file, 'w')
@@ -138,7 +141,8 @@ class IFUCubeData():
         for i in range(num1):
             this_a = self.list_par1[i]
             this_b = self.list_par2[i]
-            n = len(self.master_table.FileMap[self.instrument][this_a][this_b])
+            print(this_a,this_b)
+            n = len(self.master_table.FileMap[self.instrument][this_a][this_b]['file'])
             num_files = num_files + n
         self.num_files = num_files
 # do some basic checks on the cubes
@@ -533,11 +537,11 @@ class IFUCubeData():
         for i in range(number_bands):
             this_par1 = self.list_par1[i]
             this_par2 = self.list_par2[i]
-            nfiles = len(self.master_table.FileMap[self.instrument][this_par1][this_par2])
+            nfiles = len(self.master_table.FileMap[self.instrument][this_par1][this_par2]['file'])
 # ________________________________________________________________________________
 # loop over the files that cover the spectral range the cube is for
             for k in range(nfiles):
-                ifile = self.master_table.FileMap[self.instrument][this_par1][this_par2][k]
+                ifile = self.master_table.FileMap[self.instrument][this_par1][this_par2]['file'][k]
                 self.this_cube_filenames.append(ifile)
                 log.debug("Working on Band defined by: %s %s ", this_par1, this_par2)
 # --------------------------------------------------------------------------------
@@ -556,14 +560,24 @@ class IFUCubeData():
                     min_wave_band = np.amin(wave)
                     max_wave_band = np.amax(wave)
                     
-                    print('slice projection pars',this_par1,this_par2)
 
-                    self.map_sliceprojection_to_dqframe(this_par1,
-                                                        this_par2,
-                                                        ifile,
-                                                        min_wave_band,
-                                                        max_wave_band,
-                                                        k)
+                    # send footprint of file on sky -spatial and wavelength
+                    # set the spaxel_dq values, need the roiw and self.zcoord - defines wavelengths
+                    print('slice projection pars',this_par1,this_par2)
+                    footprint =  self.master_table.FileMap[self.instrument][this_par1][this_par2]['footprint'][k]
+                    print('footprint',footprint)
+
+
+                    self.map_sliceprojection_to_dqframe(footprint,
+                                                        min_wave_band, max_wave_band)
+
+                    
+#                    self.map_sliceprojection_to_dqframe(this_par1,
+#                                                        this_par2,
+#                                                        ifile,
+#                                                        min_wave_band,
+#                                                        max_wave_band,
+#                                                        k)
 
 
 
@@ -929,6 +943,9 @@ class IFUCubeData():
 
         parameter1 = self.list_par1
         parameter2 = self.list_par2
+
+        print('parameter 1',parameter1)
+        print('parameter 2',parameter2)
         a_min = []
         a_max = []
         b_min = []
@@ -943,7 +960,7 @@ class IFUCubeData():
             this_a = parameter1[i]
             this_b = parameter2[i]
             log.debug('Working on data from %s, %s', this_a, this_b)
-            n = len(self.master_table.FileMap[self.instrument][this_a][this_b])
+            n = len(self.master_table.FileMap[self.instrument][this_a][this_b]['file'])
             log.debug('number of files %d', n)
             for k in range(n):
                 amin = 0.0
@@ -953,7 +970,8 @@ class IFUCubeData():
                 lmin = 0.0
                 lmax = 0.0
 
-                ifile = self.master_table.FileMap[self.instrument][this_a][this_b][k]
+                ifile = self.master_table.FileMap[self.instrument][this_a][this_b]['file'][k]
+                print(ifile)
 # ______________________________________________________________________________
 # Open the input data model
 # Find the footprint of the image
@@ -990,6 +1008,8 @@ class IFUCubeData():
                     b_max.append(bmax)
                     lambda_min.append(lmin)
                     lambda_max.append(lmax)
+# store footprint for later need it for setting up DQ flags to help find holdes
+#                self.master_table.FileMap[self.instrument][this_a][this_b].append(ch_footprint)
 # ________________________________________________________________________________
     # done looping over files determine final size of cube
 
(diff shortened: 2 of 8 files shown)


jemorrison:
The current DQ flagging is done for each wavelength bin.  The FOV at each wavelength is mapped to the IFU space. The shape of the FOV varies only at the boundaries of the IFU wavelength bins, but it is not just 1 or 2 wavelength bins but 10s of wavelengths before  the FOV becomes constant across wavelength space.

jwst/cube_build/ifu_cube.py
```diff
@@ -315,13 +321,14 @@ def set_geometry(self, footprint):
             self.lambda_max = self.lambda_min + (self.naxis3) * self.cdelt3
 
             self.zcoord = np.zeros(self.naxis3)
-            self.crval3 = self.lambda_min
+            # CRPIX3 for FITS is 1 (center of first pixel)
+            # CRVAL3 then is lambda_min + self.cdelt3/ 2.0, which is also zcoord[0]
```
hbushouse: So does lambda_min correspond to the left edge of the first pixel, and hence you have to add cdelt3/2 to get it to the center of the pixel?

jwst/cube_build/ifu_cube.py
```diff
@@ -487,6 +494,7 @@ def build_ifucube(self):
         self.spaxel_flux = np.zeros(total_num)
         self.spaxel_weight = np.zeros(total_num)
         self.spaxel_iflux = np.zeros(total_num)
+        self.spaxel_dq = np.zeros((self.naxis3, self.naxis2 * self.naxis1), dtype=np.int32)
```
hbushouse: Should the dtype be **u**int32 (unsigned int32)? If it's signed, you lose the use of 1 bit and hence have a max of only 31 values.

jwst/cube_build/ifu_cube.py
```diff
@@ -538,14 +545,23 @@ def build_ifucube(self):
...
                                                                    subtract_background,
                                                                    ifile)
 
-                    coord1, coord2, wave, flux, rois_pixel, roiw_pixel, weight_pixel,\
+                    coord1, coord2, wave, flux, slice, rois_pixel, roiw_pixel, weight_pixel,\
                         softrad_pixel, alpha_det, beta_det = pixelresult
                     t1 = time.time()
                     log.info("Time to transform pixels to output frame = %.1f.s" % (t1 - t0,))
+
+# If setting the DQ plane of the IFU
```
hbushouse: pep8/flake8 won't like this comment starting way on the left edge. Should indent to match the code block.

jwst/cube_build/ifu_cube.py
```diff
@@ -538,14 +545,23 @@ def build_ifucube(self):
...
+                    if self.skip_dqflag:
+                        log.info("Skipping setting DQ flagging")
+                    else:
+                        t0 = time.time()
+                        roiw_ave = np.mean(roiw_pixel)
+                        self.map_fov_to_dqplane(this_par1, coord1, coord2, wave, roiw_ave, slice)
+                        t1 = time.time()
+                        log.info("Time to set initial dq values = %.1f.s" % (t1 - t0,))
```
hbushouse: This log message could perhaps be downgraded to "debug" status. It's not really necessary for all users to see it every time it's run.

jwst/cube_build/ifu_cube.py
```diff
@@ -1124,6 +1156,8 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
                 dec_det = np.zeros((2048, 2048))
                 lam_det = np.zeros((2048, 2048))
                 flag_det = np.zeros((2048, 2048))
+
+                slice_det = np.zeros((2048, 2048))
```
hbushouse: I know NIRSpec IFU (and MIRI MRS) exposures are always taken full-frame, but it always makes me a little nervous when I see things hardcoded like this. For the sake of keeping the code semi-generic it would be good to just use the actual input data array shape/dimensions in places like this (and similar above).

jwst/cube_build/ifu_cube.py
```diff
@@ -1204,17 +1262,374 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
+    def map_fov_to_dqplane(self, this_par1, coord1, coord2, wave, roiw_ave, slice):
+        """ Set an initial DQ flag for the IFU cube based on FOV of input data
+
+        Map the FOV of channel (MIRI) or slice (NIRSPEC) to the DQ plane
+        and set an initial DQ flagging. The process is different for MIRI and NIRSpec.
+        In the MIRI case all the slices map roughly to the same FOV across the
+        wavelength range covered by the IFU. This is not the case for NIRSpec the
+        30 different slices map to different FOV on the range of wavelengths.
+
+        Paramteter
```
hbushouse: Typo: "Parameters"

jwst/cube_build/ifu_cube.py
```diff
@@ -1204,17 +1262,374 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
             softrad_det, alpha_det, beta_det
 # ********************************************************************************
 
+    def map_fov_to_dqplane(self, this_par1, coord1, coord2, wave, roiw_ave, slice):
+        """ Set an initial DQ flag for the IFU cube based on FOV of input data
+
+        Map the FOV of channel (MIRI) or slice (NIRSPEC) to the DQ plane
+        and set an initial DQ flagging. The process is different for MIRI and NIRSpec.
+        In the MIRI case all the slices map roughly to the same FOV across the
+        wavelength range covered by the IFU. This is not the case for NIRSpec the
```
hbushouse: Need punctuation between "NIRSpec" and "the". At least a semi-colon, if not a full stop (aka period).

jwst/cube_build/ifu_cube.py
```diff
@@ -1204,17 +1262,374 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
+        roiw_ave: average spectral roi used to determine which wavelength bins
+            the input values would be mapped to
+        slice: integer slice value of input data (used in MIRI case to find
+            the points of the edge slices.)
+        """
+
+        # MIRI mapping:
+        # The FOV is roughtly the same for all the wavelenth ranges.
+        # The offset in the slices makes the calculation of the four corners
+        # of the FOV more comlicated. So we only use the two slices at
```
hbushouse: typo: "complicated"

jwst/cube_build/ifu_cube.py
```diff
@@ -1204,17 +1262,374 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
+        # the edges of the FOV to define the 4 corners.
+        # Note we can not use the wcs.footprint (because this footprint only
+        # consists of 4 values  ra min, ra max, dec min, dec max and we
+        # need 4 corners made of 8 different values.
+
+        if self.instrument == 'MIRI':
```
hbushouse: Is there a guarantee somewhere earlier that the value of self.instrument will always be uppercase? If not, should add `.upper()`

jwst/cube_build/ifu_cube.py
```diff
@@ -1204,17 +1262,374 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
+
+                iend = end_region
+
+                while iend > start_region:
+                    index_use = np.where((wave_distance < roiw_ave) & (slice == iend))
+                    if len(index_use[0]) > 0:
+                        coord2_end = coord2[index_use]
+                        coord1_end = coord1[index_use]
+                        break
+                    iend = iend - 1
```
hbushouse: Python loops (because of how slow they are) always make me wonder if the same operation can be done using clever array-based operations instead. But of course I don't understand enough of the details of exactly what's being done here to make any suggestions. It would probably take some experimentation.

jwst/cube_build/ifu_cube.py
```diff
@@ -1204,17 +1262,374 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
+
+        # NIRSpec Mapping:
+        # The FOV of each NIRSpec slice varies across the wavelength range.
+        # Each slice is mapped to each IFU wavelength plane
+        # The FOV of the slice is really just a line, so instead of using
+        # the routines the finds the overlap between a polygon and regular grid-
+        # which is used for MIRI - an algorithm that determines the spaxels that
+        # the slice line intersects is used instead.
+
+        elif self.instrument == 'NIRSPEC':
```
hbushouse: Same comment as above regarding the case of `self.instrument`

jwst/cube_build/ifu_cube.py
```diff
@@ -1204,17 +1262,374 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
+        dx = x2 - x1
+        dy = y2 - y1
+
+        # calculate error
+        error = int(dx/2.0)
+        ystep = -1
+        if y1 < y2:
+            ystep = 1
+
+        # iterate over grif to generate points between the start and end of line
```
hbushouse: What's "grif"?

jwst/cube_build/ifu_cube.py
```diff
@@ -111,7 +112,11 @@ def __init__(self,
         self.ycoord = None
         self.zcoord = None
 
-# ********************************************************************************
+        self.overlap_partial = 4  # intermediate flag
+        self.overlap_full  = 2    # intermediate flag
+        self.overlap_hole = dqflags.pixel['DO_NOT_USE']
+        self.overlap_no_coverage = dqflags.pixel['NON_SCIENCE']
```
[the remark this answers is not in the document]
jemorrison: This is just an initial intermediate flagging. Later on (in set_final_dq_flags) NON_SCIENCE and DO_NOT_USE get added together. .

jwst/cube_build/ifu_cube.py
jemorrison: yes

jwst/cube_build/ifu_cube.py
jemorrison: made change

jwst/cube_build/ifu_cube.py
jemorrison: ok

jwst/cube_build/ifu_cube.py
jemorrison: done

jwst/cube_build/ifu_cube.py
jemorrison: I do think this can be improved. I need the MIRI and NIRSpec team to do some experimentation and let me know about using the same dqflagging for all wavelengths in a band. Currently it is setting the dqflagging per wavelength bin. This might be overkill.

jwst/cube_build/ifu_cube.py
jemorrison: typo - grid

jwst/cube_build/ifu_cube.py
jemorrison: ok I added .upper() where it is read in  and setup - file_table.py

jwst/cube_build/ifu_cube.py
```diff
@@ -1153,31 +1187,55 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
-                         dqflags.pixel['DEAD'] + dqflags.pixel['HOT'] +
-                         dqflags.pixel['RC'] + dqflags.pixel['NONLINEAR'])
+                         dqflags.pixel['NON_SCIENCE'])
+
+            valid3 = np.bitwise_and((wave >= min_wave_tolerance),
+                                     (wave <= max_wave_tolerance))
             # find the location of all the values to reject in cube building
-            good_data = np.where((np.bitwise_and(dq_all, all_flags) == 0) & (valid2 == True))
+            good_data = np.where((np.bitwise_and(dq_all, all_flags) == 0) &
+                                 (valid2 == True) & (valid3 == True))
```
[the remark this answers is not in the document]
jemorrison: ok thanks fixed code

jwst/cube_build/ifu_cube.py
```diff
@@ -315,13 +321,14 @@ def set_geometry(self, footprint):
             self.lambda_max = self.lambda_min + (self.naxis3) * self.cdelt3
```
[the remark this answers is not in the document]
jemorrison: self.cdelt3 is set if the ifu cube has a linear wavelength dimension. It is set in setup_ifube_wcs
and is set to what has been read in from the cubepars reference file or set on the command line.
It is not set if the wavelength range is nonlinear - there is no self.cdelt3

jwst/cube_build/ifu_cube.py
```diff
@@ -1204,17 +1262,374 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
+            # for each of the 30 slices - find the projection of this slice
+            # onto each of the IFU wavelength planes.
+            for islice in range(30):
+                index_slice = np.where(slice == islice+1)
+
+                # find the smaller set of wavelengths to search over for this slice
+                wavemin = np.amin(wave[index_slice])
+                wavemax = np.amax(wave[index_slice])
+                iwavemin = np.absolute(np.array(wavemin - self.zcoord) / (self.cdelt3_normal/2.0))
+                iwavemax = np.absolute(np.array(wavemax - self.zcoord) / (self.cdelt3_normal/2.0))
```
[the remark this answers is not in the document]
jemorrison: I don't think it should be divided by 2. Good catch.

jwst/cube_build/ifu_cube.py
```diff
@@ -1153,31 +1187,55 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
-                    # done looping  - pull out valid values
+                    slice_det[yind, xind] = ii+1
+
+                # after looping over slices  - pull out valid values
                 valid_data = np.where(flag_det == 1)
                 y, x = valid_data
                 ra = ra_det[valid_data]
                 dec = dec_det[valid_data]
                 wave = lam_det[valid_data]
+                slice = slice_det[valid_data]
```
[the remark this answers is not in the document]
jemorrison: ok I replaced slice by slice_no

jwst/cube_build/ifu_cube.py
```diff
@@ -1204,17 +1262,374 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
+        coord2: eta coordinates of input data (~y coordinate in IFU space)
+        wave: wavelength of input data
+        roiw_ave: average spectral roi used to determine which wavelength bins
+            the input values would be mapped to
+        slice: integer slice value of input data (used in MIRI case to find
+            the points of the edge slices.)
+        """
+
+        # MIRI mapping:
+        # The FOV is roughtly the same for all the wavelenth ranges.
```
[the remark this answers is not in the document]
jemorrison: fixed

jwst/cube_build/ifu_cube.py
```diff
@@ -1204,17 +1262,374 @@ def map_detector_to_outputframe(self, this_par1, this_par2,
...
+        and the spaxel area. The values assigned are:
+        a. self.overlap_partial = overlap partial
+        b  self.overlap_full = overlap_full
+        bit_wise combination of these values is allowed to account for
+        dithered FOVs.
+
+        Parameter
+        ----------
+        xi_corner: xi coordinates of the 4 corners of the FOV on the cube plane
+        eta_corner: xi coordinates of the 4 corners of the FOV on the cube plane
```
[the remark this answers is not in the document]
jemorrison: fixed

CHANGES.rst
```diff
@@ -713,6 +713,8 @@ csv_tools
 cube_build
 ----------
 
+- Added dq flagging [#3804]
```
hbushouse: Is this truly the first and only DQ flagging that's been in `cube_build`? Or is it just the addition of special flagging for holes and non-slice regions? If the latter, then perhaps the log info should be more specific.

jemorrison: This is the first time the DQ flag has been sent in cube_building.

drlaw1558:
I've been testing this PR for some simulated MRS data; on the whole it looks like it's working great!  However, the hole-finding algorithm is still having some issues around the edge of the footprint.  I was surprised when it reported about 16,000 holes in a good cube, and it looks like they are all around the border (see attached example).  With 800 planes in the cube, that's about 20 per plane, which seems to match the image.

Maybe we can add another step when defining a hole, so that we say:
1) possible_holes = np.where((self.spaxel_dq == 0) & (self.spaxel_weight ==0))
2) location_holes = (all possible_holes that are *not* bordered by at least one spaxel with the NON_SCIENCE dq bit set)

There's surely some clever numpy way to do this, and it would mean that the reported number of holes was more useful.  It might also be useful for the log to instead report out the average number of holes per wavelength plane.  The case that we're trying to catch is mostly a bad set of cube-building parameters, which would affect many planes.  A log statement of << 1 would thus be 'good', while I'd expect bad parameters to give >> 1.  This would be easy to parse visually.

dq

jemorrison:
David point me to your data - the test data I have does not give so many potential holes.
I have 2472 in 800 planes for ch1 - short and 217 in 645 planes for ch2 short.

drlaw1558:
Replied offline.

PR #3806, opened 2019-07-11
Enable environment snapshot creation for macos

rendinam:
Adds a second `BuildConfig` assigned to an OSX host which simply installs the package along with its testing dependencies as is done on Linux so that an environment snapshot can be created which includes the OS-specific differences in the minimal conda environment that serves as a base.

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -28,25 +28,37 @@ data_config.server_id = 'bytesalad'
 data_config.root = '${PYTEST_BASETEMP}'
 data_config.match_prefix = '(.*)_result' // .json is appended automatically
 
+
 // Build and test with python 3.6 and released dependencies from astroconda
-bc = new BuildConfig()
-bc.nodetype = 'jwst'
-bc.name = 'stable-deps'
-bc.env_vars = env_vars
-bc.conda_ver = '4.6.14'
-bc.conda_packages = [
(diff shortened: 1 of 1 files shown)


jdavies-st:
Will this cause a build and test on a mac os machine then?

rendinam:
The macos `BuildConfig` added to accomplish this simply installs the package in order to seed the environment for dumping an equivalent YML file as the Linux environment. If tests are desired, test commands can be added to the `test_cmds` list and they will run on the internal host assigned to the `macos` nodetype.

JenkinsfileRT
```diff
@@ -28,25 +28,37 @@ data_config.server_id = 'bytesalad'
...
+
+
+# macos-specific buildconfig to cause the creation of counterparts to the linux
+# environment dumps. Packages in a mininmal conda environment differ by OS
+# which is why this is needed.
+bc1 = utils.copy(bc0)
+bc1.nodetype = 'macos'
+bc1.name = 'macos-stable-deps'
+bc1.build_cmds = ["pip install -e .[test]"]
+bc1.test_configs = []
```
jdavies-st: I think this needs to be
```
bc1.test_cmds = []
```

rendinam: Good catch. The copied `test_cmds` list is still in effect here.

PR #3807, opened 2019-07-12
flat_field needs to support different data models

philhodge:
When getting the value of the DISPAXIS keyword for NIRSpec spectroscopic data, the `flat_field` step used the syntax that would work for an `ImageModel` (or similar), but sometimes the input is a `MultiSlitModel`.  For the latter case, the keyword will be gotten from the first slit in the model.


jwst/flatfield/flat_field.py
```diff
@@ -222,7 +222,13 @@ def do_nirspec_flat_field(output_model, f_flat_model, s_flat_model, d_flat_model
     log.debug("Flat field correction for NIRSpec spectrographic data.")
 
     exposure_type = output_model.meta.exposure.type
-    dispaxis = output_model.meta.wcsinfo.dispersion_direction
+    try:
+        dispaxis = output_model.meta.wcsinfo.dispersion_direction
+    except AttributeError:
+        if len(output_model.slits) > 0:
```
hbushouse: What happens in the confluence of bad events such that `wcsinfo.dispersion_direction` doesn't exist and hence it enters this block, but `output_model.slits` also doesn't exist (i.e. it's not a `MultiSlitModel`)? Will this operation throw an error? Perhaps that's appropriate, but just curious.

hbushouse: Oh and don't forget to update the change log.

philhodge: I wondered whether I should also check for `products` in addition to `slits` (i.e. a `MultiProductModel`), as in `extract_1d`, but I don't think they would ever reach `flat_field`.  We have too many models that aren't interchangeable.

hbushouse: Right. `MultiProductModel` is only used as a container for multiple `DrizProductModel` instances, which means the data have been resampled and resampling is always **after** flat-fielding.

Issue #2585, opened 2018-09-27
Reconcile SHUTTER vs SHUTSTA keyword names in schemas


hbushouse:
We currently have 4 data model schemas that define a data model property called "shutter_state". They are:

1. `multiexposure.schema.yaml`
2. `multiproduct.schema.yaml`
3. `multispec.schema.yaml`
4. `slitdata.schema.yaml`

`multiexposure.schema.yaml` has this property mapped to the keyword name "SHUTTER", while the other 3 schemas all map it do the keyword "SHUTSTA". "SHUTSTA" is a keyword that's documented in the JWST Keyword Dictionary; "SHUTTER" is not.

The name "SHUTTER" may be a hold-over from old code or keyword design. The discrepancy needs to be fixed.

hbushouse:
As part of the solution to this, would it be possible to only define all of the shutter/slit related keywords in a single sub-schema that is in turn inherited by these other model schemas, for the sake of consistency and ease of maintenance?

hbushouse:
Fixed by #3809

PR #3809, opened 2019-07-15
Removed unused BACKGRND keyword from multiexposure schema

### LINKED ISSUE #2586: Is BACKGRND keyword needed/used in multiexposure.schema.yaml?
The data model schema `multiexposure.schema.yaml` defines the keyword "BACKGRND", which is mapped to the model property "background". I can't find any obvious instance of this being used anywhere in the cal pipeline code base. This keyword is not defined in the JWST Keyword Dictionary, so if we do need it for something, we need to add it to the JWKD.

To my knowledge the `multiexposure.schema.yaml` model is used for the output products from the `exp_to_source` step, which is only applied to data from spectroscopic modes in which we have multiple defined sources, such as NRS_MSASPEC, NRC_WFSS, and NIS_WFSS. The only obvious connection to a "background" status for these modes is the background indicator in the NIRSpec MSA meta data file that tells us which shutters contain background (or not). But because multiple shutters usually make up a single slit, it's not obvious to me how the background status could get transferred to a given slit or spec instance (because there would be multiple different values).

We need to figure out if this keyword definition is really needed or even possible to be used and if not, then remove it.

hbushouse:
Fixes #2586 

The keyword BACKGRND does not appear in any other schemas and is not used in any of the pipeline code, and hence is being removed as unnecessary.


hbushouse:
That latest commit addresses #2585 by changing the SHUTTER keyword to SHUTSTA and bringing all the slit-specific meta data into sync with what's currently defined in `slitdata.schema.yaml`.

nden:
`slitdata.schema.yaml` can be included as a reference. Also `varience.schema.yaml`. I think this would make the schema a bit more readable, may be.

hbushouse:
Latest commit vastly simplifies the schema by just including `slitdata.schema.yaml` as a reference, which already includes all of the items that had been listed explicitly, as well as references to `variance.schema.yaml`, `photometry.schema.yaml`, `wcsinfo.schema.yaml`, etc. The output source-based products now contain many more keywords copied over to their SCI extension headers than before, including the photometry and WCS keywords (which is desired).

PR #3816, opened 2019-07-16
Update outlier_detection for photutils>=0.7 non-indexable Apertures

jdavies-st:
This came up in testing the released `jwst 0.13.7` SDP-delivered conda spec file on our 0.13.7 Artifactory input and truth data snapshot.

The dev version of `photutils` no longer has indexable ApertureMasks if there is only one Aperture.

### DIFF
--- a/jwst/outlier_detection/outlier_detection_scaled.py
+++ b/jwst/outlier_detection/outlier_detection_scaled.py
@@ -84,7 +84,12 @@ class OutlierDetectionScaled(OutlierDetection):
             radius_outer = 5
 
         apertures = CircularAperture((xcenter, ycenter), r=radius)
-        aperture_mask = apertures.to_mask(method='center')[0]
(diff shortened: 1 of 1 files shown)


jwst/outlier_detection/outlier_detection_scaled.py
```diff
@@ -84,7 +84,12 @@ def do_detection(self):
             radius_outer = 5
 
         apertures = CircularAperture((xcenter, ycenter), r=radius)
-        aperture_mask = apertures.to_mask(method='center')[0]
+        # ApertureMask in photutils<=0.6 is indexable.  In >=0.6 it is scalar
```
hbushouse: Title of the PR says ">=0.7", while this says ">=0.6". So which is it?

jdavies-st: PR title is correct.  Will change the comment.

PR #3827, opened 2019-07-17
works with new modeling

nden:
The changes in this PR are necessary (except the one in resample) for the code to run with the new modeling.

- assign_wcs.nirspec:
 Deleted the inverse of a model `M1` which is to be assigned as inverse to a model `M` because modeling checks recursively for "correct" inverses and raises an error even if they are not needed.

- combine_1d:
  There's only one  input to the WCS transform and it defines the wavelength. Spatial coordinates are constants.

- extract_1d:
  `util.pixel_area` takes as input the shape of the data, constructs input arrays from it and evaluates the WCS. In the case of a `CubeModel` a shape of length 3 was passed in but the WCS takes 2 inputs.
  
- resample:
 Fixed an unclosed file `Warning`.

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -2769,7 +2769,12 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
             del npixels_temp
 
             # Convert to flux density (for a point source).
-            pixel_solid_angle = util.pixel_area(slit.meta.wcs, slit.data.shape)
+            wcs_shape = slit.data.shape
+            wcs = slit.meta.wcs
+            if len(wcs_shape) > 2:
+                # CubeModel
+                wcs_shape = wcs_shape[1:]
+            pixel_solid_angle = util.pixel_area(wcs, wcs_shape)
             if pixel_solid_angle is None:
                 log.warning("Pixel solid angle could not be determined")
                 pixel_solid_angle = 1.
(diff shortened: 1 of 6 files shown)


jwst/resample/resample_step.py
```diff
@@ -100,7 +100,8 @@ def get_drizpars(self, ref_filename, input_models):
         used a resample.cfg file or run ResampleStep using command line args,
         then these will overwerite the defaults pulled from the reference file.
         """
-        drizpars_table = datamodels.DrizParsModel(ref_filename).data
+        with datamodels.DrizParsModel(ref_filename) as drpt:
+            drizpars_table =drpt.data
```
hbushouse: Need a white space after the = sign.

jwst/extract_1d/extract.py
```diff
@@ -2971,8 +2981,13 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
                                                            sp_order)
                     else:
                         wcs = input_model.meta.wcs
+
+                    wcs_shape = input_model.data.shape
+                    if len(wcs_shape) > 2:
+                        # CubeModel
+                        wcs_shape = wcs_shape[1:]
                     pixel_solid_angle = util.pixel_area(wcs,
```
philhodge: Hmmm, I just noticed that this section for getting `pixel_solid_angle` is called for every integration.  The value should be the same for every integration, and I have no idea why I didn't limit this to being called only for `integ == 0`.  This is my mistake, and it may explain why some tests have been taking over a day to run.

Issue #3703, opened 2019-06-20
Add keyword OPMODE and datamodel.meta.instrument.lamp_mode


stscijgbot:
Issue [JP-808](https://jira.stsci.edu/browse/JP-808) was created by Howard Bushouse:

https://jira.stsci.edu/browse/JWSTKD-320 requests the addition of a new keyword "LAMPMODE" for NIRSpec internal lamp exposures, to document the Operating Mode (MOS, IFU, FS) selected by the user, which Cal needs in order to know how to process lamp exposures in calwebb_spec2.

It's a string keyword with possible values BRIGHTOBJ, FIXEDSLIT, GRATING-ONLY, IFU, MSASPEC, NULL. It should be added immediately following the LAMP keyword, with a datamodel name of "lamp_mode."

stscirij:
Name was changed to "OPMODE"

PR #3842, opened 2019-07-19
Add attribute meta.instrument.lamp_mode from keyword OPMODE

stscirij:
Addresses #3703

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -493,6 +493,11 @@ properties:
             type: string
             fits_keyword: LAMP
             blend_table: True
+          lamp_mode:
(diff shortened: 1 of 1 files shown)


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -493,6 +493,11 @@ properties:
             type: string
             fits_keyword: LAMP
             blend_table: True
+          lamp_mode:
+            title: Operating mode for NIRSPEC internal lamp exposures
```
hbushouse: For consistency with what's been implemented in JWSTKD-320 the title should be "Lamp operating mode".

stscirij: Fixed

Issue #3818, opened 2019-07-16
NIS_EXTCAL exposures still showing up in some ASN's


stscijgbot:
Issue [JP-865](https://jira.stsci.edu/browse/JP-865) was created by Howard Bushouse:

JP-287 (via [https://github.com/spacetelescope/jwst/pull/3611]) removed NIS_EXTCAL exposures from at least some ASN's, but JWSTDMS-31 reports that they are still showing up in some "wfs-image2" associations and are failing to process. It appears that this is due to the fact that some WFS&C type observations can have NIS_EXTCAL exposures contained within them and in this case it seems that the wfsc status is overriding the exclusion of EXP_TYPE=NIS_EXTCAL and allowing them to still appear in the ASN's.

The ASN rules need to be updated to never allow any EXP_TYPE=NIS_EXTCAL exposures appear in any level-2b or level-3 ASN's.

stscijgbot:
Comment by David Davis: Is there another example? Checking the files listed in JWSTDMS-31 I am not seeing an examples of exptype='nis_extcal' in the associations. I logged onto owld-2,

ssh owld-2
...
>uname -a
Linux tljwdmsdp2.stsci.edu 3.10.0-957.10.1.el7.x86_64 #1 SMP Thu Feb 7 07:12:53 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

Under the directory, /ifs/archive/test/jwst/store/doggett_test/tests/SOR4B/dtest1/JWSTDP-2019.1.0-205~ed639c06/2019-06-24-200028/sdp

I see, asn_creation/asn_tables/, and the issue is with jw00629-o009 and jw00629-o010.

Both files are of type wfs-image2

>grep asn_type o009/*.json
o009/jw00629-o009_20190624t200638_wfs-image2_001_asn.json: "asn_type": "wfs-image2",
o009/jw00629-o009_20190624t200638_wfs-image2_002_asn.json: "asn_type": "wfs-image2",

neither of these have exptype = 'nis_extcal' and the ones listed are science,

grep exptype o009/*.json
o009/jw00629-o009_20190624t200638_wfs-image2_001_asn.json: "exptype": "science",
o009/jw00629-o009_20190624t200638_wfs-image2_002_asn.json: "exptype": "science",

and I get the same for o010,

grep exptype o010/*.json
o010/jw00629-o010_20190624t200638_wfs-image2_001_asn.json: "exptype": "science",
[...]

PR #3844, opened 2019-07-22
JP-865 Updated association rules to exclude NIS_EXTCAL eposures

ddavis-stsci:
The association rules are updated to exclude NIS_EXTCAL exposures. In JWSTDMS-31 the example given is for jw00629 where observation number 9 & 10 are listed as exptype='NIS_EXTCAL'. The new rules exclude those files and is shown in the difference between the updated associations for jw00629_20190624t200638_pool.csv,

grep Only diffs_output_July22.log
121:Only in output: jw00629-o009_wfs-image2_001_asn.json
122:Only in output: jw00629-o009_wfs-image2_002_asn.json
123:Only in output: jw00629-o010_wfs-image2_001_asn.json
124:Only in output: jw00629-o010_wfs-image2_002_asn.json

All the other associations remain the same.

### DIFF
--- a/jwst/associations/lib/rules_level2_base.py
+++ b/jwst/associations/lib/rules_level2_base.py
@@ -897,7 +897,7 @@ class Constraint_Target(DMSAttrConstraint):
 
 class Constraint_ExtCal(Constraint):
     """Remove any nis_extcals from the associations, they
-       are NOT to be processed in the pipline!"""
+       are NOT to receive level-2b or level-3 processing!"""
 
     def __init__(self):
         super(Constraint_ExtCal, self).__init__(
--- a/jwst/associations/lib/rules_level2_base.py
+++ b/jwst/associations/lib/rules_level2_base.py
@@ -46,6 +46,7 @@ __all__ = [
     'AsnMixin_Lv2Special',
     'AsnMixin_Lv2Spectral',
     'Constraint_Base',
+    'Constraint_ExtCal',
     'Constraint_Image_Nonscience',
     'Constraint_Image_Science',
     'Constraint_Mode',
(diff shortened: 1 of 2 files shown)


jwst/associations/lib/rules_level2_base.py
```diff
@@ -894,6 +895,21 @@ def __init__(self):
             sources=['targetid'],
         )
 
+class Constraint_ExtCal(Constraint):
+    """Remove any nis_extcals from the associations, they
+       are NOT to be processed in the pipline!"""
```
hbushouse: Technically they *are* processed in the cal pipeline, but only through the first stage (level-2a) of calibration. It's only the later stages (2b and 3) that use ASN's as drivers, hence we don't want them showing up in any ASN's. So to be completely correct, you could change this comment to "... they are NOT to receive level-2b or level-3 processing".

jwst/associations/lib/rules_level2_base.py
```diff
@@ -894,6 +895,21 @@ def __init__(self):
...
+
+    def __init__(self):
+        super(Constraint_ExtCal, self).__init__(
+            [
+                DMSAttrConstraint(
+                    name='exp_type',
+                    sources=['exp_type'],
+                    value='nis_extcal'
```
hbushouse: Not sure what the EXP_TYPE values look like in the pool files, from which these entries are read. Are the values in the pool files always uppercase (e.g. "NIS_EXTCAL") and, if so, does that make a difference for this matching?

jwst/associations/lib/rules_level2_base.py
ddavis-stsci: The comments are updated.

jwst/associations/lib/rules_level2_base.py
ddavis-stsci: Yes, the pool files have all of these values in upper case. It does not make a difference since they are all converted to lower case in the association code. Since I was looking at (copying from) the pool file I used upper case as written.

hbushouse:
Fixes #3818

Issue #3777, opened 2019-07-03
Problems with TA's in spec3 ASN's


stscijgbot:
Issue [JP-843](https://jira.stsci.edu/browse/JP-843) was created by Howard Bushouse:

JWSTDMS-40 reports a couple of problems related to the inclusion of TA exposures in spec3 ASN's. One of which is that the rate product is listed as the input member, rather than the cal product. And the second is that the TA exposure is being included in the "exp_to_source" conversion and throws an error due to data format. The TA's should not be included in any pipeline operations.

stscijgbot:
Comment by David Davis: As of July 11 the issue of the product name seems to be resolved. Using the current scsb association code and the input pool file from the dev run of 2019-07-03 tests show that the target_acquisition entries are listed using the expected _cal product, 
{noformat}
"expname": "jw00626022001_02101_00001_nrs1_cal.fits",
 "exptype": "target_acquisition",
 "exposerr": "null",
 "asn_candidate": "('o022', 'observation')"\},
 {"expname": "jw00626022001_02101_00002_nrs1_cal.fits",
 "exptype": "target_acquisition",
 "exposerr": "null",
 "asn_candidate": "('o022', 'observation')"
{noformat}
and not the, incorrect _rate, product
{noformat}
"expname": "jw00626022001_02101_00001_nrs1_rate.fits",
 "exptype": "target_acquistion",
 "exposerr": "null",
 "asn_candidate": "('o022', 'observation')"\},{noformat}
as in the run with the 2019-07-03 pipeline code. The second part of the issue, not including target_acquistion in the pipeline processing is in progress.

stscijgbot:
Comment by David Davis: The current code excludes all entries in the association file when the exptype does not match the entries in the allowed_exptypes variable. If any filtering is done the code lists the files that pass the selection criteria, e.g.,

2019-07-24 13:20:39,122 - stpipe.Spec3Pipeline - INFO - Starting calwebb_spec3 ...
Filtering datasets based on allowed exptypes: ['science']
Files accepted for processing: jw00626022001_03104_00001_nrs1_cal.fits
Files accepted for processing: jw00626022001_03102_00003_nrs1_cal.fits
Files accepted for processing: jw00626022001_03108_00004_nrs1_cal.fits
Files accepted for processing: jw00626022001_03104_00004_nrs1_cal.fits
Files accepted for processing: jw00626022001_03102_00002_nrs1_cal.fits
[...]

PR #3857, opened 2019-07-24
JP-843 Updates to exclude TA's from calwebb_spec3 processing

ddavis-stsci:
Updated `datamodels` to allow for specification of an allowed list of ASN `exptypes` to keep when loading an ASN into a `ModelContainer`, and update `calwebb_spec3` to make use of this feature to keep TA's from processing.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,8 @@ assign_wcs
 associations
 ------------
 
+- Update to prevent target_acquisitions from processing in the spec3 pipeline [#3777]
+
 - Return filename with extensions based on file type [#2671]
 
 - Ensured that all target acqs are processed by Level 2 [#3765]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,8 +15,6 @@ assign_wcs
 associations
 ------------
 
-- Update to prevent target_acquisitions from processing in the spec3 pipeline [#3777]
-
 - Return filename with extensions based on file type [#2671]
 
 - Ensured that all target acqs are processed by Level 2 [#3765]
@@ -32,6 +30,8 @@ associations
 
 datamodels
 ----------
+
+- Update to prevent target_acquisitions from processing in the spec3 pipeline [#3777]
 
 - Use public API of jsonschema to ease upgrade to 3.x. [#3705]
 
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -3,6 +3,7 @@ from collections import OrderedDict
 import os.path as op
 import warnings
 import re
+import logging
 
 from asdf import AsdfFile
 from astropy.io import fits
@@ -19,6 +20,9 @@ __doctest_skip__ = ['ModelContainer']
 
 __all__ = ['ModelContainer']
 
+# Configure logging
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
 
 class ModelContainer(model_base.DataModel):
     """
@@ -65,15 +69,12 @@ class ModelContainer(model_base.DataModel):
     # does not describe the data contents of the container.
     schema_url = "container.schema"
 
-    def __init__(self, init=None, **kwargs):
+    def __init__(self, init=None, allowed_exptypes=None, **kwargs):
 
-        super(ModelContainer, self).__init__(init=None, **kwargs)
+        super(ModelContainer, self).__init__(init=None, allowed_exptypes=None  , **kwargs)
 
         self._models = []
-        self.allowed_exptypes = None
-        for key in kwargs.keys():
-            if key == 'allowed_exptypes':
-                self.allowed_exptypes = kwargs.get('allowed_exptypes')
+        self.allowed_exptypes = allowed_exptypes
 
         if init is None:
             # Don't populate the container with models
@@ -191,11 +192,11 @@ class ModelContainer(model_base.DataModel):
         # grab all the files
         if self.allowed_exptypes:
             infiles = []
-            print("Filtering datasets based on allowed exptypes:", self.allowed_exptypes)
+            logger.debug("Filtering datasets based on allowed exptypes:", self.allowed_exptypes)
             for member in asn_data['products'][0]['members']:
                 if any([x for x in self.allowed_exptypes if re.match(member['exptype'], x, re.IGNORECASE)]):
                         infiles.append(member['expname'])
-                        print("Files accepted for processing:", member['expname'])
+                        logger.debug("Files accepted for processing:", member['expname'])
         else:
             infiles = [member['expname'] for member
             in asn_data['products'][0]['members']]
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -2,6 +2,7 @@ import copy
 from collections import OrderedDict
 import os.path as op
 import warnings
+import re
 
 from asdf import AsdfFile
 from astropy.io import fits
@@ -69,6 +70,10 @@ class ModelContainer(model_base.DataModel):
         super(ModelContainer, self).__init__(init=None, **kwargs)
 
         self._models = []
+        self.allowed_exptypes = None
+        for key in kwargs.keys():
+            if key == 'allowed_exptypes':
+                self.allowed_exptypes = kwargs.get('allowed_exptypes')
 
         if init is None:
             # Don't populate the container with models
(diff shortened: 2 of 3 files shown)


jwst/datamodels/container.py
```diff
@@ -69,6 +70,10 @@ def __init__(self, init=None, **kwargs):
         super(ModelContainer, self).__init__(init=None, **kwargs)
 
         self._models = []
+        self.allowed_exptypes = None
```
stscieisenhamer: Style comment: Since `allowed_exptypes` is for `ModelContainer` only, it may be appropriate to make it an explicit keyword argument. This would help in both self-documentation and in not carrying too much in the `kwargs` list.

jwst/datamodels/container.py
```diff
@@ -181,9 +186,20 @@ def from_asn(self, asn_data, asn_file_path=None):
...
         """
-        # make a list of all the input files
-        infiles = [member['expname'] for member
-                   in asn_data['products'][0]['members']]
+        # match the allowed_exptypes to the exptype in the association and retain
+        # only those file that match, as a list, if allowed_exptypes is set to none
+        # grab all the files
+        if self.allowed_exptypes:
+            infiles = []
+            print("Filtering datasets based on allowed exptypes:", self.allowed_exptypes)
```
stscieisenhamer: Design: No `print` statements. These should be logging. I would guess at the `INFO` level.

jwst/pipeline/calwebb_spec3.py
```diff
@@ -61,12 +61,13 @@ def process(self, input):
             The exposure or association of exposures to process
         """
         self.log.info('Starting calwebb_spec3 ...')
+        self.allowed_exptypes = ['science']
```
stscieisenhamer: For spec3, I believe `BACKGROUND` should also be included. @hbushouse ?

CHANGES.rst
```diff
@@ -15,6 +15,8 @@ assign_wcs
 associations
 ------------
 
+- Update to prevent target_acquisitions from processing in the spec3 pipeline [#3777]
```
hbushouse: Minor nit, but these changes are really in `datamodels` and `calwebb_spec3`, as opposed to the `associations` infrastructure or rules, so I'd categorize them in a different section of the change log.

jwst/datamodels/container.py
hbushouse: Agree on logging vs. print, but I'm not sure we'd want this at the `INFO` level. We already get complaints from regular users about too much infrastructure and process flow control type info filling up the logs. So perhaps `DEBUG`.

jwst/datamodels/container.py
```diff
@@ -181,9 +186,20 @@ def from_asn(self, asn_data, asn_file_path=None):
...
+            for member in asn_data['products'][0]['members']:
+                if any([x for x in self.allowed_exptypes if re.match(member['exptype'], x, re.IGNORECASE)]):
+                        infiles.append(member['expname'])
+                        print("Files accepted for processing:", member['expname'])
```
hbushouse: logging, not print

jwst/pipeline/calwebb_spec3.py
hbushouse: Correct. We need to keep both `science` and `background` member types.

You know, to avoid a LOT of confusion, we really should've called this attribute `memtype` (*member* type), instead of `exptype`. Is it too late for a total redesign?!

jwst/datamodels/container.py
ddavis-stsci: Moved allowed_exptypes to a keyword argument.

jwst/datamodels/container.py
ddavis-stsci: Set output as logging a the debug level as per Howard's request.

jwst/pipeline/calwebb_spec3.py
ddavis-stsci: Allowed exposure types are now ['science','background']

CHANGES.rst
ddavis-stsci: Moved this comment to the datamodels section.

jwst/datamodels/container.py
ddavis-stsci: Output at the DEBUG level.

jwst/datamodels/container.py
```diff
@@ -64,11 +69,12 @@ class ModelContainer(model_base.DataModel):
     # does not describe the data contents of the container.
     schema_url = "container.schema"
 
-    def __init__(self, init=None, **kwargs):
+    def __init__(self, init=None, allowed_exptypes=None, **kwargs):
 
-        super(ModelContainer, self).__init__(init=None, **kwargs)
+        super(ModelContainer, self).__init__(init=None, allowed_exptypes=None  , **kwargs)
```
hbushouse: I assume pep8/flake8 won't like those whitespaces in "None  ,"

jdavies-st: Calls to `super()` can be simplified in python 3:
```
super().__init__(init=None, allowed_exptypes=None, **kwargs)
```

jwst/pipeline/calwebb_spec3.py
```diff
@@ -61,12 +61,13 @@ def process(self, input):
             The exposure or association of exposures to process
         """
         self.log.info('Starting calwebb_spec3 ...')
+        self.allowed_exptypes = ['science','background']
```
jdavies-st: This can probably be `allowed_exptypes` instead of `self.allowed_exptypes` as it is not being used outside of `process()` method.  Unless you'd like to make this attribute available on the `ModelContainer` itself for introspection, which I suppose could be useful.

jwst/datamodels/container.py
```diff
@@ -64,11 +69,12 @@ class ModelContainer(model_base.DataModel):
     # does not describe the data contents of the container.
     schema_url = "container.schema"
 
-    def __init__(self, init=None, **kwargs):
+    def __init__(self, init=None, allowed_exptypes=None, **kwargs):
```
jdavies-st: This extra `allowed_exptypes` parameter should be added to the "Parameters" docstring above.

jdavies-st: Do you think `exptype` instead of `allowed_exptypes` would be more succinct?  Or `asn_exptype` to be explicit that it is an association exptype and not a FITS keyword EXPTYPE?

jwst/datamodels/container.py
```diff
@@ -181,9 +187,20 @@ def from_asn(self, asn_data, asn_file_path=None):
...
-        infiles = [member['expname'] for member
-                   in asn_data['products'][0]['members']]
+        # match the allowed_exptypes to the exptype in the association and retain
+        # only those file that match, as a list, if allowed_exptypes is set to none
+        # grab all the files
+        if self.allowed_exptypes:
+            infiles = []
+            logger.debug("Filtering datasets based on allowed exptypes:", self.allowed_exptypes)
+            for member in asn_data['products'][0]['members']:
+                if any([x for x in self.allowed_exptypes if re.match(member['exptype'], x, re.IGNORECASE)]):
```
jdavies-st: It be good to have `allowed_exptypes` be a string or list of strings.  I.e. if a string is passed in, put it in a list, then use it.  Many of the use cases for it will be to pick out a single exptype, so the call
```
container = datamodels.open('input_asn.json', allowed_exptypes='science')
```
is better than
```
container = datamodels.open('input_asn.json', allowed_exptypes=['science'])
```
I think.

hbushouse: But in the case of `calwebb_spec3`, where we need to allow 2 exptypes - science and background - what would the syntax look like? If it's:
```
allowed_exptypes=['science', 'background']
```
then aren't we right back where we started?

jdavies-st: Yes.  My suggestion is to allow both in `ModelContainer`.  If passed a string, wrap it in a list.

jdavies-st: I.e the calling sequence should allow
```
container = datamodels.open('input_asn.json', allowed_exptypes='science')
container = datamodels.open('input_asn.json', allowed_exptypes=['science', 'background'])
```

stscieisenhamer: On this point, my normal reaction is to "just expect an iterable", don't allow bare `str`, and be done with it. However, because `str` is an iterable, results can be very bizarre, and it will be done often. So, handling the case internally is reasonable.

jwst/datamodels/container.py
ddavis-stsci: Done

jwst/datamodels/container.py
ddavis-stsci: I went with asn_exptype to be explicit.

jwst/datamodels/container.py
ddavis-stsci: Since this is an internal variable I left it as is.

hbushouse:
Fixes #3777

PR #3862, opened 2019-07-25
Fixed error for MIRI DQ flagging for wavelengths no data

jemorrison:


jemorrison:
For Cubes made from 2 MIRI channels but the same sub-channel (so data from the same exposures) there are a number of wavelength bins with no data. This PR checks if there is
data in the wavelength band before it tries to  set the DQ flags. 
This type of cube is created by the spec2 pipeline

jwst/cube_build/ifu_cube.py
```diff
@@ -1343,6 +1352,7 @@ def map_fov_to_dqplane(self, this_par1, coord1, coord2, wave, roiw_ave, slice_no
 
                 iend = end_region
 
+                coord1_end = None
                 while iend > start_region:
                     index_use = np.where((wave_distance < roiw_ave) & (slice_no == iend))
```
philhodge: Similar question (different variables) as just above, regarding decrementing `iend` in the `while` loop.

jwst/cube_build/ifu_cube.py
```diff
@@ -1351,28 +1361,21 @@ def map_fov_to_dqplane(self, this_par1, coord1, coord2, wave, roiw_ave, slice_no
...
-                coord1_total = np.concatenate((coord1_start, coord1_end), axis=0)
-                coord2_total = np.concatenate((coord2_start, coord2_end), axis=0)
-
-#            coord1_start = coord1[slice == start_region]
-#            coord2_start = coord2[slice == start_region]
-
-#            coord1_end = coord1[slice == end_region]
-#            coord2_end = coord2[slice == end_region]
+                if coord1_start is not None and coord1_end is not None:
+                    coord1_total = np.concatenate((coord1_start, coord1_end), axis=0)
```
philhodge: Should there be an `else:` clause, in case `coord1_start` or `coord1_end` is still None?  A lot of variables are assigned values here.  Or you could just raise an exception if either `coord1_start` or `coord1_end` is None.

hbushouse:
Will this update also fix the error seen in https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/333/testReport/jwst.tests_nightly.general.miri.test_mrs_spec3/TestSpec3Pipeline/_stable_deps__test_spec3_pipeline1/
which occurs in line 1974 of `ifu_cube.py` (input IFUCubeData doesn't have a `spaxel_dq` attribute)?

jwst/cube_build/ifu_cube.py
jemorrison: For wavelength planes that fall in the gap between channels and there is no data in these planes the coord1_start and coord1_end will be None

jwst/cube_build/ifu_cube.py
```diff
@@ -1332,7 +1332,16 @@ def map_fov_to_dqplane(self, this_par1, coord1, coord2, wave, roiw_ave, slice_no
...
+                # coord2_start = coord2[slice == start_region]
+                # coord1_end = coord1[slice == end_region]
+                # coord2_end = coord2[slice == end_region]
+                # wmin = imin[0]
+                # wmax = imax[0]
+
                 istart = start_region
+                coord1_start = None
                 while istart < end_region:
                     index_use = np.where((wave_distance < roiw_ave) & (slice_no == istart))
```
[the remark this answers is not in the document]
jemorrison: Good suggestion. I was worried about my extreme slices on the FOV not falling on a wavelength plane - then I was going to go the next slice - until I found one on the plane.
This is confusing and unnecessary. So I removed the while

jemorrison:
This PR resolves JP-894 and the error in  https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/333/testReport/jwst.tests_nightly.general.miri.test_mrs_spec3/TestSpec3Pipeline/_stable_deps__test_spec3_pipeline1/
I ran the refression test_mrs_spec3 and it ran with and error in the code - just a error that data was different which is expected.

Issue #3836, opened 2019-07-19
Pixel solid angle is the same for each integration


stscijgbot:
Issue [JP-878](https://jira.stsci.edu/browse/JP-878) was created by Philip Hodge:

The extract_1d step needs the solid angle of a pixel in order to convert surface brightness to flux density.  The solid angle is computed using the wcs function.  If the input data array is multi-integration data, the pixel solid angle is currently computed for each integration.  It should only be computed for the first integration.

hbushouse:
Fixed by #3863

stscijgbot:
Comment by Howard Bushouse: Fixed by https://github.com/spacetelescope/jwst/pull/3863

Issue #3784, opened 2019-07-05
Compute wavelengths for NIRCam TSO data


stscijgbot:
Issue [JP-845](https://jira.stsci.edu/browse/JP-845) was created by Philip Hodge:

The changes described in JP-784 (GitHub #3637) should also be done for NIRCam TSO data, i.e. function extract_tso_object in extract_2d/grisms.py.  The wavelength array should be computed and assigned to an attribute, and the variance arrays should be copied from input to output.

stscijgbot:
Comment by Rosa Diaz: Is this an improvement or something that was missed during previous development?

stscijgbot:
Comment by Philip Hodge: This is something that I missed when I modified a function in extract_2d/grisms.py.  The original ticket was JP-768, Bugs and enhancements for WFSS processing.  Part of this ticket was for computing and saving the wavelength at each pixel, and part was for copying the variance arrays from input to output.  I did this for one function in grisms.py, but I didn't notice that there was a second function that also needed the same modifications.

stscijgbot:
Comment by Howard Bushouse: Computation and attachment of wavelength arrays has been implemented in [https://github.com/spacetelescope/jwst/pull/3863].

Still need to also cutout and propagate the variance arrays for this mode.

hbushouse:
Fixed in #4504

stscijgbot:
Comment by Howard Bushouse: Fixed in [https://github.com/spacetelescope/jwst/pull/4504]

PR #3863, opened 2019-07-26
JP-878 Don't try to compute pixel solid angle for WFSS data

philhodge:
The `extract_1d` step was modified to only call `util.pixel_area()` for the first integration of multi-integration data.  Also, that function immediately returns `pixel_solid_angle = None` for WFSS data; for these data the WCS function returns constant values for right ascension and declination, so the angular size of a pixel cannot be determined by this function.

The extract_2d step was modified to compute and populate the `wavelength` attribute for NIRCam TSO data.  This partially addresses JP-845, GitHub #3784, Compute wavelengths for NIRCam TSO data, but only the wavelength attribute is populated, not the variance arrays.  To close this issue, the variance arrays need to be copied from input to output.

### DIFF
--- a/jwst/extract_2d/tests/test_grisms.py
+++ b/jwst/extract_2d/tests/test_grisms.py
@@ -41,6 +41,8 @@ data_path = os.path.split(os.path.abspath(data.__file__))[0]
 # Default wcs information
 # This is set for a standard nircam image just as an example
 # It does not test the validity of the absolute results
+# for create_tso_wcsimage, set the width of the output image to this value:
+NIRCAM_TSO_WIDTH = 10
 wcs_image_kw = {'wcsaxes': 2, 'ra_ref': 53.1490299775, 'dec_ref': -27.8168745624,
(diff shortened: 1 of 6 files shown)


jwst/extract_2d/tests/test_grisms.py
```diff
@@ -41,6 +41,8 @@
 # Default wcs information
 # This is set for a standard nircam image just as an example
 # It does not test the validity of the absolute results
+# for create_tso_wcsimage, set the width of the output image to this value:
+NIRCAM_TSO_WIDTH = 10
```
hbushouse: Is this being reduced from 2048 to 10 just to save time/memory in the unit test?

philhodge: Yes.  Either the Travis or Jenkins CI test (I forget which) failed because it couldn't finish in the allotted 10 minutes.

hbushouse:
Fixes #3836

PR #3871, opened 2019-07-31
[WIP]: Adding pathloss reference tests

mfixstsci:
Adding unit tests for the pathloss correction. Have roughly 50% coverage and am working my way through `do_correction` currently.

### DIFF
--- a/jwst/pathloss/tests/test_pathloss.py
+++ b/jwst/pathloss/tests/test_pathloss.py
@@ -1,0 +1,249 @@
+"""
+Unit tests for pathloss correction
+"""
+
+from jwst.datamodels import MultiSlitModel, PathlossModel, IFUImageModel
+from jwst.pathloss import PathLossStep
+from jwst.pathloss.pathloss import (get_center, 
+                                    get_aperture_from_model, 
+                                    calculate_pathloss_vector,
+                                    is_pointsource)
+from jwst.pathloss.pathloss import do_correction 
+import numpy as np
+import pytest
+
+
+# Begin get_center tests
+def test_skip_step(make_imagemodel):
+    """If no pathloss reference file is specified or given in the header,
+    make sure the step skips.
+    """
+
+    im = make_imagemodel(10, 10, 'NRS_AUTOWAVE')
+    result = PathLossStep.call(im)
+
+    assert(result.meta.cal_step.pathloss == 'SKIPPED')
+
+
+def test_get_center_ifu():
+    """get_center assumes IFU targets are centered @ (0.0, 0.0)
+    """
+
+    x_pos,y_pos = get_center("NRS_IFU", None)
+
+    assert(x_pos==y_pos==0.0)
+
+
+def test_get_center_attr_err():
+    """if no center provided for modes that are not IFU,
+    center is assigned to (0.0, 0.0)
+    """
+    datmod = MultiSlitModel()
+    x_pos, y_pos = get_center("NRS_MSASPEC", datmod)
+
+    assert(x_pos==y_pos==0.0)
+
+
+def test_get_center_exp_type():
+    """if exp_type is not in NRS, center is returned (0.0,0.0) 
+    """
+    datmod = MultiSlitModel()
+    x_pos, y_pos = get_center("NRC_IMAGE", datmod)
+
+    assert(x_pos==y_pos==0.0)
+
+
+def test_get_center_exptype():
+    """ If exptype is "NRS_MSASPEC" | "NRS_FIXEDSLIT" | "NRS_BRIGHTOBJ" and
+    source_xpos and source_ypos exist in datamod.slits, make sure it's returned
+    """
+    
+    datmod = MultiSlitModel()
+    datmod.slits.append({'source_xpos':1, 'source_ypos':2})
+    
+    x_pos, y_pos = get_center("NRS_FIXEDSLIT", datmod.slits[0])
+
+    assert(x_pos==1)
+    assert(y_pos==2)
+
+
+# Begin get_aperture_from_model tests
+def test_get_app_from_model_null():
+    """If exp_type isn't the NRS or NIS specific mode,
+    routine returns None
+    """
+
+    datmod = MultiSlitModel()
+    datmod.meta.exposure.type = 'NRC_IMAGE'
+
+    result = get_aperture_from_model(datmod, None)
+
+    assert(result==None)
+
+
+def test_get_aper_from_model_fixedslit():
+    """For a given exposures aperture, make sure the correct
+    aperture reference data is returned for fixedslit mode
+    """
+
+    datmod = PathlossModel()
+    datmod.apertures.append({'name':'S200A1'})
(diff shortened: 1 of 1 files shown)


mfixstsci:
I noticed that the exception tests I wrote for `do_correction` for the `NRS_MSASPEC` and `NRS_FIXEDSLIT` modes, the step still exits with `result.meta.cal_step.pathloss == 'COMPLETE'` where I would expect it to be `result.meta.cal_step.pathloss == 'SKIPPED'`. The correction isn't actually applied here so maybe this slipped through the cracks. I noticed this isn't a problem for the `NIS_SOSS` modes because anytime there is a problem with the datamodel, the correction is skipped. @stscirij do you know why this is? For example:

https://github.com/spacetelescope/jwst/blob/master/jwst/pathloss/pathloss.py#L188

Here if the slit size is less than 0 is takes you down to this line:

https://github.com/spacetelescope/jwst/blob/master/jwst/pathloss/pathloss.py#L243

Where it says the step is being skipped, but then 2-3 lines down the `meta.cal_step.pathloss` is assigned to `COMPLETE`

https://github.com/spacetelescope/jwst/blob/master/jwst/pathloss/pathloss.py#L246

Under the log entry should there be a `result.meta.cal_step.pathloss == 'SKIPPED'`?

jwst/pathloss/tests/test_pathloss.py
```diff
@@ -0,0 +1,332 @@
...
+    assert(result == datmod.apertures[0])
+
+
+def test_get_aper_from_model_msa():
+    """For a given exposures aperture, make sure the correct
+    aperture reference data is returned for MSA mode
+    """
+
+    datmod = PathlossModel()
+    datmod.apertures.append({'shutters':1})
```
philhodge: It looks to me as if the comment in pathloss.py is incorrect with regard to NRS_MSASPEC.  The comment says that `match` is the number of shutters, but it's the shutter number instead.  I would suggest using a number other than 1 for this test (and most other tests, for that matter), e.g. 5.  That would check whether `result` was the shutter number or the number of shutters.

jwst/pathloss/tests/test_pathloss.py
```diff
@@ -0,0 +1,332 @@
...
+
+    wavelength, pathloss, is_inside_slitlet = calculate_pathloss_vector(datmod.apertures[0].pointsource_data,
+                                                                        datmod.apertures[0].pointsource_wcs,
+                                                                        0.0, 0.0)
+
+    # Wavelength array is calculated with this: crval3 +(float(i+1) - crpix3)*cdelt3
+    # Where i is the iteration of np.arange(wavesize) which is the 1st dimension of the pointsource
+    # data array.
+    wavelength_comparison = np.array([1 + (float(i+1) - 1.0)*1 for i in np.arange(10)])
+    assert(np.all(wavelength==wavelength_comparison))
```
philhodge: For float data it might be better to use `np.allclose(wavelength, wavelength_comparison)` instead of `np.all(wavelength == wavelength_comparison)`.  `np.allclose` lets you specify absolute and/or relative tolerances for the comparison.

jwst/pathloss/tests/test_pathloss.py
```diff
@@ -0,0 +1,332 @@
...
+
+    # pathloss vector gets assigned at beginning of calculate_pathloss_vector and in this
+    # case, doesnt change (np.zeros(wavesize, dtype=np.float32))
+    pathloss_comparison = np.zeros(10, dtype=np.float32)
+    assert(np.all(pathloss==pathloss_comparison))
```
philhodge: Wouldn't `pathloss` be an array of ones?  In this case, `np.all` should be fine.

jwst/pathloss/tests/test_pathloss.py
```diff
@@ -0,0 +1,332 @@
...
+    point_source = 'not a point'
+    result = is_pointsource(point_source)
+    assert(result == False)
+
+# Begin do_correction exception testing
+def test_do_correction_msa_slit_size_eq_0():
+    """If slits have size 0, quit calibration.
+    """
+    datmod = MultiSlitModel()
+    datmod.slits.append({'data':np.empty((10,10,10))})
```
philhodge: `np.empty` doesn't give you an array with size 0; it returns an array with uninitialized data.

jwst/pathloss/tests/test_pathloss.py
mfixstsci: @philhodge according to what is in master right now the pathloss vector is initiated as an array pf zeros (https://github.com/spacetelescope/jwst/blob/master/jwst/pathloss/pathloss.py#L106) and then a few lines down, if the shape of the source data is 1 (https://github.com/spacetelescope/jwst/blob/master/jwst/pathloss/pathloss.py#L111), according to the comment the source is always inside the slit so there is no change to the pathloss vector.

jwst/pathloss/tests/test_pathloss.py
mfixstsci: @philhodge ahh right, I went ahead and replace the lines with `np.array([])`! Thanks

jwst/pathloss/tests/test_pathloss.py
```diff
@@ -0,0 +1,332 @@
...
+    """If exp_type isn't the NRS or NIS specific mode,
+    routine returns None
+    """
+
+    datmod = MultiSlitModel()
+    datmod.meta.exposure.type = 'NRC_IMAGE'
+
+    result = get_aperture_from_model(datmod, None)
+
+    assert(result==None)
```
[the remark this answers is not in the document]
mfixstsci: @philhodge thanks, will go ahead and make the change.

jwst/pathloss/tests/test_pathloss.py
mfixstsci: Good idea, I went ahead and made the change.

jwst/pathloss/tests/test_pathloss.py
mfixstsci: @philhodge it looks like it is the number of shutters. `nshutters` is returned from `jwst.assign_wcs.util.get_num_msa_open_shutters` and then passed to `get_aper_from_model` in the `do_correction` portion of the script. https://github.com/spacetelescope/jwst/blob/master/jwst/pathloss/pathloss.py#L194 So, assuming that the value returned from `get_num_msa_open_shutters` matches the value of `shutters` for the slit. Since I am unit testing `get_aper_from_model` I am hardcoding the `nshutters` value to make them match on purpose. Should I unit test a case with `jwst.assign_wcs.util.get_num_msa_open_shutters` or should we leave that to the regression test for `do_correction`?

philhodge: @mfixstsci I think this is fine as it is.

stscicrawford:
@mfixstsci can you fix the flake8 failures in this and also rebase it once you have done?   Then I think this would be ready to be merged

mfixstsci:
@stscicrawford okay, I rebased and finally got all of the `flake8` issues taken care of!

PR #3873, opened 2019-08-01
Recover and save step configuration parameters

stscieisenhamer:
Resolves JP-875/#3834
WIP JP-876/#3835
Add the API necessary to retrieve and save a `Step`s parameters as an ASDF object.

Main code is to recover the parameters, since the current architecture did not support this. The API adds two new attributes to `Step`:

- `Step.pars`: `dict` of the parameters
- `Step.pars_model`: `StepParsModel` of the configuration.

As a convenience, a new data model, `StepParsModel`, has also been implemented. Using the datamodel provides the convenience of the defining schema for validation. It should also help in the CRDS interface, which usually takes a `DateModel` as input for searching.

Finally, a new command-line option `--save-parameters <filename.asdf>` is implemented, allowing saving of the `Step`s parameters.

### DIFF
--- a/jwst/datamodels/__init__.py
+++ b/jwst/datamodels/__init__.py
@@ -58,6 +58,7 @@ from .saturation import SaturationModel
 from .slit import SlitModel, SlitDataModel
 from .source_container import SourceModelContainer
 from .spec import SpecModel
+from .steppars import StepParsModel
 from .straylight import StrayLightModel
 from .superbias import SuperBiasModel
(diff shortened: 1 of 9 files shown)


stscieisenhamer:
Closing due to the situation where github does not want to include the latest commit. No other solutions have been offered, even stackoverflow.

stscieisenhamer:
Replacement PR #3886

PR #3876, opened 2019-08-02
Jp 868: Background target exposures need to create their own science ASN

### LINKED ISSUE #3824: Background-designated observation did not create a Level 3 observation
Issue [JP-868](https://jira.stsci.edu/browse/JP-868) was created by Jonathan Eisenhamer:

From DMS test run complete_long_01 done on 28Jun19 as run161, proposal jw00626 observation 29, a background observation, did not make a Level3 association, whereas its companion science observation, 28, did.

The pool file is attached.

ddavis-stsci:
NIS_IFU bkg observations as level3 spec association. 
Fix unit tests where target acquisitions were left out of associations. test_level3_basics.py  expects at least one member to be a target acquisition.  

Fixes #3824.

### DIFF
--- a/jwst/associations/lib/rules_level3.py
+++ b/jwst/associations/lib/rules_level3.py
@@ -364,7 +364,8 @@ class Asn_Lv3SpecAux(AsnMixin_AuxData, AsnMixin_BkgScience):
                     DMSAttrConstraint(
                         name='allowed_bkgdtarg',
                         sources=['exp_type'],
-                        value=['mir_mrs','nrs_ifu'],)
+                        value=['mir_mrs','nrs_ifu','mir_lrs-fixedslit',
+                               'nrs_fixedslit'],)
                 ],
             reduce=Constraint.any
                     ),
--- a/jwst/associations/lib/rules_level3.py
+++ b/jwst/associations/lib/rules_level3.py
@@ -352,21 +352,21 @@ class Asn_Lv3SpecAux(AsnMixin_AuxData, AsnMixin_BkgScience):
             ),
             Constraint(
                 [
-                DMSAttrConstraint(
-                name='bkgdtarg',
-                sources=['bkgdtarg'],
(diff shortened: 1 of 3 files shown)


jwst/associations/lib/rules_level3.py
```diff
@@ -352,21 +352,21 @@ def __init__(self, *args, **kwargs):
...
-                DMSAttrConstraint(
-                name='mir_bkgdtarg',
-                sources=['exp_type'],
-                value=['mir_mrs'],)
-                    ],
-                    reduce=Constraint.any
+                    DMSAttrConstraint(
+                        name='allowed_bkgdtarg',
+                        sources=['exp_type'],
+                        value=['mir_mrs','nrs_ifu'],)
```
hbushouse: I wonder if this is too restrictive. There are other EXP_TYPE's that are allowed to have associated background targets, such as mir_lrs-fixedslit, nrs_fixedslit, and eventually the imaging modes too (e.g. nrc_image, mir_image). Does there need to be a constraint involving EXP_TYPE at all or can we just go with the `bkgdtarg` constraint above?

ddavis-stsci: Good idea and I like simplification. So I tried it. Unfortunately it breaks several spec3 associations e.g. jw00623-o002_spec3_001_asn.json.

hbushouse: So if removing this constraint doesn't work, then we should at least expand the list of exp_type values to also include "mir_lrs-fixedslit" and "nrs_fixedslit".

ddavis-stsci: Updated the allowed_exptypes to include  "mir_lrs-fixedslit" and "nrs_fixedslit". nrs_fixedslit is already in the test suite but I'll need to find an example of mir_lrs-fixedslit for testing.

PR #3883, opened 2019-08-05
JenkinsfileRT: Add unstable deps

### LINKED ISSUE #3880: Start running regression tests with astropy master
Issue [JP-906](https://jira.stsci.edu/browse/JP-906) was created by Nadia Dencheva:

jwst will need to switch to the new astropy.modeling package which is currently only on astropy master. Travis testing already runs with the dev versions of dependencies. We need to start running the regression test suite with dev versions as well.

jhunkeler:
Resolves #3880.

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -46,7 +46,7 @@ bc0.test_cmds = [
     "pytest -r sxf -n 15 --bigdata --slow \
     --basetemp=${pytest_basetemp}  --junit-xml=results.xml \
     --log-level=INFO \
-    jwst/tests_nightly/general"
+    jwst"
 ]
 bc0.test_configs = [data_config]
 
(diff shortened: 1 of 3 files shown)


jdavies-st:
If we take the tack of running these in a matrix on Jenkins, then we're going to have to dial down the `pytest-xdist` parallelization, i.e. change both test builds to `pytest -n 15` so they can share the total number of cores.

The other option is to run them at different times of the day using different JenkinsfileRTs, as in my PR #3747.  This has more overhead on the Jenkins setup and maintenance I feel.

So if the first option works, then let's do it.

nden:
We can start with these hashes in requirements-dev.txt

astropy: a8f496b26
photutils: 99f101be95af67f
gwcs: 3e2bc108ec3a66139

jdavies-st:
I kicked off an RT job to try out the new JenkinsfileRT in this branch

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/

Looks like the dev build failed, as it couldn't compile `numpy`.  Do we want `numpy` dev in this build?  Or would it be better to just use the just-released version for now?

nden:
:+1: for a public release of numpy

jhunkeler:
@jdavies-st @nden 
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST_UNSTABLE/

When you click `Build with Parameters` you'll see a big text area. You can populate it with `pip install` arguments

```
somethingA
"somethingB<1.2.3"
git+https://something.com/blah.git
-r a_requirements_file_somewhere_else.txt
--upgrade numpy
# and so on
```

nden:
New hash for modeling is d16e0ff5ff3ba780b. Could we rerun the regression tests with this configuration? If they pass then this is ready to merge.

PR #3892, opened 2019-08-06
Document usage of ASDF as a configuration file format.

stscieisenhamer:
Update the documentation to include ASDF as a configuration file format.

### DIFF
--- a/docs/jwst/introduction.rst
+++ b/docs/jwst/introduction.rst
@@ -433,6 +433,7 @@ logging level designations of ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``, and
 ``CRITICAL``. Only messages at or above the specified level
 will be displayed.
 
+.. _`Configuration Files`:
 
 Configuration Files
 ===================
--- a/docs/jwst/introduction.rst
+++ b/docs/jwst/introduction.rst
@@ -404,10 +404,6 @@ of the stpipe-log.cfg file is:
     handler = file:pipeline.log
     level = INFO
 
-which specifies that all log messages will be directed to a file called
-"pipeline.log" and messages at a severity level of INFO and above will be
-recorded.
-
 If there's no ``stpipe-log.cfg`` file in the working directory, which specifies
 how to handle process log information, the default is to display log messages
 to stdout. If you want log information saved to a file, you can specify the
@@ -433,6 +429,14 @@ logging level designations of ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``, and
 ``CRITICAL``. Only messages at or above the specified level
 will be displayed.
 
+.. note::
+
+   Setting up ``stpipe-log.cfg`` can lead to confusion, especially if it is
+   forgotten about. If one has not run a pipeline in awhile, and then sees no
+   logging information, most likely it is because ``stpipe-log.cfg`` is
+   present. Consider using a different name and specifying it explicitly on the
+   command line.
+
 .. _`Configuration Files`:
 
 Configuration Files
@@ -446,9 +450,8 @@ overridden either by the command line options, as previously described, and by a
 local configuration file.
 
 A configuration file should be used when there are parameters a user wishes to
-change from the default/CRDS version, but will not be changed on every run of
-the step. To create a configuration file with the default
-values, use the command: ::
+change from the default/CRDS version for a custom run of the step. To create a
+configuration file with the default values, use the command: ::
 
  strun <step.class> --save-parameters <filename.asdf>
 
--- a/docs/jwst/introduction.rst
+++ b/docs/jwst/introduction.rst
@@ -85,17 +85,13 @@ referencing their class names is done as follows:
   $ strun jwst.dq_init.DQInitStep jw00017001001_01101_00001_nrca1_uncal.fits
 
 When a pipeline or step is executed in this manner (i.e. by referencing the
-class name), it will be run using all default parameter values. The same thing
-can be accomplished by using the default configuration file corresponding to
-each:
-::
-
-  $ strun calwebb_detector1.cfg jw00017001001_01101_00001_nrca1_uncal.fits
-  $ strun dq_init.cfg jw00017001001_01101_00001_nrca1_uncal.fits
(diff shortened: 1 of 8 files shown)


docs/jwst/introduction.rst
```diff
@@ -387,6 +392,22 @@ Alternatively you can specify the ``skip`` argument on the command line:
...
+
+    [*]
+    handler = file:pipeline.log
+    level = INFO
+
+which specifies that all log messages will be directed to a file called
+"pipeline.log" and messages at a severity level of INFO and above will be
+recorded.
+
 If there's no ``stpipe-log.cfg`` file in the working directory, which specifies
```
philhodge: The first sentence is now redundant.

It would be useful to point out that there is a very big advantage to using a log configuration file with a non-default file name.  If you use the default name and then a week later forget that stpipe-log.cfg is in the directory, your log output will disappear "into the void" (I'm writing from experience).  If you use a different name, it's safe to forget about it because it won't be used by default.

docs/jwst/introduction.rst
```diff
@@ -412,60 +433,39 @@ logging level designations of ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``, and
...
+pipeline or individual steps. For JWST, configuration files are retrieved from
+CRDS, just as with other reference files. If there is no match between a step,
+the input data, and CRDS, the coded defaults are used. These values can be
+overridden either by the command line options, as previously described, and by a
+local configuration file.
 
- $ collect_pipeline_cfgs .
+A configuration file should be used when there are parameters a user wishes to
+change from the default/CRDS version, but will not be changed on every run of
+the step. To create a configuration file with the default
```
philhodge: Will this work for a pipeline, or only for a step?  If I use `--save_parameters` when the class is for a pipeline, will I get the default ASDF configuration file for every step in the pipeline?

docs/jwst/references_general/references_general.rst
```diff
@@ -259,6 +260,17 @@ documentation on each reference file.
 | :ref:`WFSSBKG <wfssbkg_reffile>`                 | :ref:`background <background_step>`         |
 +--------------------------------------------------+---------------------------------------------+
 
+Step Parameters Reference Types
++++++++++++++++++++++++++++++++
+
+When each ``Step`` is instantiated, a CRDS look-up, based on the ``Step`` class
+name and input data, is made to retrieve a configuration file. The ``reftype``
+for such configuration files is ``pars-<class name``. For example, for the step
```
philhodge: `>` is missing.

docs/jwst/stpipe/config_asdf.rst
```diff
@@ -0,0 +1,77 @@
...
+      pedigree: GROUND
+      reftype: pars-makeliststep
+      telescope: JWST
+      title: MakeListStep default parameters
+      useafter: '1990-04-24T00:00:00'
+   parameters:
+      class: jwst.stpipe.tests.steps.MakeListStep
+   ...
+
+All of the keys found under ``meta`` are required, most of which are
```
philhodge: I would say "shown in the above example" rather than "found".  "Found" implies that everything you put in `meta` is required by CRDS.  Basically I'm saying we should be careful to distinguish between requirements and an example.

docs/jwst/stpipe/user_step.rst
```diff
@@ -13,27 +13,30 @@ parameters on it.
...
+description of the contents. The rest of this documented will focus on the step
+parameters themselves.
 
-``name`` defines the name of the step.  This is distinct from the
-class of the step, since the same class of Step may be configured in
-different ways, and it is useful to be able to have a way of
-distinguishing between them.  For example, when Steps are combined
-into :ref:`stpipe-user-pipelines`, a Pipeline may use the same Step class
-multiple times, each with different configuration parameters.
+All step parameters appear under the ``parameters:`` section and indented. The
```
philhodge: Suggestion:  "section and indented." --> "section, and they must be indented."

docs/jwst/stpipe/user_step.rst
```diff
@@ -13,27 +13,30 @@ parameters on it.
...
+amount of indentation does not matter, as long as all parameters are indented
+equally.
+
+Every step configuration file must contain the parameter ``class``, followed by
```
philhodge: The example in lines 28 - 30 of config_cfg.rst shows `name` first, then `class`.  Does the order matter?

docs/jwst/introduction.rst
```diff
@@ -412,60 +433,39 @@ logging level designations of ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``, and
...
+values, use the command: ::
```
ddavis-stsci: For some reason I'm finding the above very hard to parse. I think that you're saying that 
"A local configuration file should be used when there are parameters a user wishes to
change from the default/CRDS version for a custom run of the pipeline." 
I'm not sure why the "every run of the step" has to do with this. I can change the local config file every time if I like.

docs/jwst/introduction.rst
```diff
@@ -412,60 +433,39 @@ logging level designations of ``DEBUG``, ``INFO``, ``WARNING``, ``ERROR``, and
...
-
-For a given step, the step's cfg file specifies parameters and their default
-values; it includes parameters that are typically not changed between runs.
-Parameters that are usually reset for each run are not included in the cfg file,
-but instead specified on the command line. An example of a cfg file for the
-jump detection step is:
-::
+Once retrieved, the file can be edited, removing parameters that should be left
+at their default/CRDS values, and setting the remaining parameters to the
+desired values.
```
ddavis-stsci: I assume that removing the parameters that the user does not want to change is to make sure those are updated with new parameters with updated default values from CRDS? If so I'd just add that to the text.

docs/jwst/introduction.rst
stscieisenhamer: This should only operate on the specific step/pipeline.

docs/jwst/stpipe/user_step.rst
stscieisenhamer: No, order does not matter.

PR #3901, opened 2019-08-08
Straylight unittest

jemorrison:
First attempt on straylight unit test.
Some slight modifications of the code were made after review.


jwst/straylight/tests/test_straylight.py
```diff
@@ -0,0 +1,72 @@
...
+
+    slice_map[:,10:15] = 0
+    slice_map[:,30:35] = 0
+    image.data[:,10:15] = 0.5
+    image.data[:,30:35] = 0.5
+    roi = 4
+    power = 1
+
+    result = correct_mrs_modshepard(image, slice_map, roi, power)
+    assert type(image) is type(result)
```
philhodge: It would be much better to check that `result` has the values that you expect.

jwst/straylight/tests/test_straylight.py
```diff
@@ -0,0 +1,72 @@
...
+
+def test_correct_detector():
```
philhodge: IMHO, you could run the step off-line yourself with detector = 'MIRIFULONG', and that would be a sufficient test.  Remember that the unit tests are run every time someone submits a PR (in spacetelescope/jwst), and calling the step takes a lot of time, even if the input is small.  You could have a test such as this for a regression test, but I just don't think it's necessary.

jwst/straylight/straylight.py
```diff
@@ -226,40 +223,48 @@ def correct_mrs_modshepard(input_model, slice_map, roi, power):
 
     # find if any of the gap pixels are bad pixels - if so mark them
     mask_dq = input_model.dq.copy()
-    all_flags = (dqflags.pixel['DEAD'] + dqflags.pixel['HOT'])
+#    all_flags = (dqflags.pixel['DEAD'] + dqflags.pixel['HOT'])
+#    testflags = np.bitwise_and(mask_dq, all_flags)
     # where are pixels set to any one of the all_flags cases
```
philhodge: There's only one case, now, so this comment is a little out of date.

jwst/straylight/straylight.py
```diff
@@ -226,40 +223,48 @@ def correct_mrs_modshepard(input_model, slice_map, roi, power):
...
+    # avoid cosmic ray contamination. This check is used to screen out
+    # undetected cosmic rays. Only use the science data for this cosmic ray test.
     cosmic_ray_test = 0.02 * np.max(output.data[slice_map > 0])
     image_gap[image_gap > cosmic_ray_test] = 0
 
     image_gap[image_gap < 0] = 0   #set pixels less than zero to 0
     image_gap = convolve(image_gap, Box2DKernel(3))   #smooth gap pixels
-    image_gap *= mask   #reset science pixels to 0
+    image_gap *= mask   # reset science pixels to 0
     #we not not want the reference pixels to be used in the convolution
```
philhodge: typo:  "not not" --> "do not"

jwst/straylight/straylight_step.py
```diff
@@ -14,7 +14,7 @@ class StraylightStep (Step):
 
     spec = """
          method = option('Nearest','ModShepard',default='ModShepard') #Algorithm method
-         roi = float(default = 50.0) # Region of interest
+         roi = integer(default = 50) # Region of interest
```
philhodge: Does `roi` need to be an integer?  In line 285 of straylight.py, you divide `roi` by 2.0, which makes `half_roi` a float.  Shouldn't `roi` be a float as well?

jwst/straylight/tests/test_straylight.py
```diff
@@ -0,0 +1,124 @@
...
+    compare[14,:] = [3.0688972e+01, 3.0629198e+01, 3.0642084e+01, 3.0645052e+01, 1.6645306e-01,
+                     1.1563802e-03, 1.6646616e-01, 3.0641420e+01, 3.0642084e+01, 3.0641420e+01,
+                     1.6646616e-01, 1.1563802e-03, 1.6645306e-01, 3.0645052e+01, 3.0642084e+01,
+                     3.0629198e+01]
+    compare[15,:] = [3.0696348e+01, 3.0638458e+01, 3.0652958e+01, 3.0663357e+01, 2.7708188e-01,
+                     1.6659760e-01, 2.7704430e-01, 3.0657593e+01, 3.0652958e+01, 3.0657593e+01,
+                     2.7704430e-01, 1.6659760e-01, 2.7708188e-01, 3.0663357e+01, 3.0652958e+01,
+                     3.0638458e+01]
+
+    assert(np.allclose(compare, result.data))
```
philhodge: It would be safer to specify a tolerance when calling `np.allclose` for float data.  You could try something like rtol=1.e-6.  This comment also applies to lines 106 and 124 of this file.

jwst/straylight/straylight_step.py
jemorrison: I think routine shepard_2d_kernel is assuming roi is an integer.  
I program crashed when it was an odd vlalue.
I put in checks to make sure the roi is given as an even integer > 2 because that is what
the shepard_2d_kernel is assuming. 
I have change half_roi to be a integer - since it really should be since it is used in
xk, yk = np.meshgrid(np.arange(-half_roi, half_roi + 1),
                         np.arange(-half_roi, half_roi + 1))

nden:
@jemorrison If you rebase on master you will pick up the coverage report correctly.

jdavies-st:
The coverage report is indicating that the new code introduced in this PR does not have coverage at or above ~45%, i.e. the current level of the whole codebase.  That is the reason for the failed check.

PR #3904, opened 2019-08-09
Add unit tests to srctype

mfixstsci:
This PR adds a few more unit tests to the source type correction which now has 100 % coverage.


jwst/srctype/tests/test_srctype.py
```diff
@@ -151,3 +152,40 @@ def test_tso_types():
...
+    """ Test for when visit is tso.
+    """
+    input = datamodels.ImageModel((10,10))
+    input.meta.exposure.type = "FGS_DARK"
+    input.meta.visit.tsovisit = True
+    result = srctype.set_source_type(input)
+
+    assert(result.meta.target.source_type == "POINT")
+
+@pytest.mark.xfail
```
jdavies-st: Why is this set to xfail?  If you want to make sure that `srctype` produces an error on this type of data, then it would be better to do something like:
```
with pytest.raises(RuntimeError):
    srctype.set_source_type(input)
```
instead of `@pytest.mark.xfail`.  `xfail`ing a test means that there is an intention to eventually fix that failure at some point.  Our code raises a `RuntimeError` on purpose, so we need to capture that in the test.

jwst/srctype/tests/test_srctype.py
```diff
@@ -151,3 +152,40 @@ def test_tso_types():
     input.meta.visit.tsovisit = True
     output = srctype.set_source_type(input)
     assert output.meta.target.source_type == 'POINT'
+
+def test_nrs_msaspec():
+    """Test for when exposure type is NRS_MSASPEC
+    """
+    input = datamodels.MultiSlitModel()
+    input.meta.exposure.type = "NRS_MSASPEC"
```
jdavies-st: If you wanted to cover NRS_FIXEDSLIT too, you could parameterize this test.
```
@pytest.mark.parametrize("exptype", ["NRS_MSASPEC", "NRS_FIXEDSLIT"])
def test_nrs_msaspec(exptype):
    ...
    input.meta.exposure.type = exptype
    ...
```

jwst/srctype/tests/test_srctype.py
```diff
@@ -151,3 +152,40 @@ def test_tso_types():
...
+
+    assert(result.slits[0].source_type == 'POINT')
+    assert(result.slits[1].source_type == 'UNKNOWN')
+    assert(result.slits[2].source_type == 'EXTENDED')
+
+def test_is_tso():
+    """ Test for when visit is tso.
+    """
+    input = datamodels.ImageModel((10,10))
+    input.meta.exposure.type = "FGS_DARK"
```
jdavies-st: Should probably use a TSO exptype here.  In fact, there's a `test_tso_types()` test above.  Maybe parametrize that one and test all the TSO types?

jdavies-st: Something like
```
@pytest.mark.parametrize("exptype", ["NRS_BRIGHTOBJ", "NRC_TSGRISM", "NIS_SOSS",
    "MIR_LRS-SLITLESS"])
def test_tso_types(exptype):
    """Check that TSO exptypes get set to POINT"""
    input = datamodels.ImageModel()
    input.meta.observation.bkgdtarg = False
    input.meta.dither.primary_type = 'NONE'
    input.meta.visit.tsovisit = True
    input.meta.exposure.type = exptype

    result = srctype.set_source_type(input)

    assert result.meta.target.source_type == 'POINT'
```

PR #3905, opened 2019-08-09
Add [ephem] to install extras in JenkinsfileRT

jdavies-st:
SDP needs `jplephem` in the conda env .yaml file for their CI linked to our CI.  So even though we don't use it in our regression tests, it should be included in the build.

Also there is a problem with them being able to run `set_velocity_aberration.py`.


JenkinsfileRT
```diff
@@ -39,14 +39,12 @@ bc0.conda_packages = [
...
-    "pip install -e .[test]",
+    "pip install -e .[test,ephem]",
     "pip install pytest-xdist",
 ]
 bc0.test_cmds = [
-    "pytest -r sx -v -n 30 --bigdata --slow \
-    --basetemp=${pytest_basetemp}  --junit-xml=results.xml \
-    --log-level=INFO \
-    jwst/tests_nightly/general"
+    "pytest -r sxf -n 15 --bigdata --slow \
```
jhunkeler: `30` -> `15`: was this in case of RT jobs overlapping, or just so the host is still semi-usable for other things during runs?

scripts/set_velocity_aberration.py
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
```
jhunkeler: Hmm, `scripts/*` are installed to `$venv/bin` as executable scripts so wouldn't removing this be bad?

JenkinsfileRT
jdavies-st: This was just in anticipation of our eventual use of a dev build/test running in parallel.  And it turns out 15 or 30 has no effect on total run time as it is a few tests that take forever that drive that.

scripts/set_velocity_aberration.py
jdavies-st: Doh! Yes.  Since these are called directly.

Because `jplephem` wasn't imported properly this was showing up as different from the other scripts that were getting that line rewritten upon installation.  I'll will change this back.

JenkinsfileRT
jdavies-st: But I see we are doing this in other PRs, so I will nix this.

jdavies-st:
So the problem with `set_velocty_aberration.py` was that it had DOS-type line endings:
```
$ file scripts/*.py
scripts/move_wcs.py:                Python script text executable, ASCII text
scripts/okify_regression_tests.py:  Python script text executable, ASCII text
scripts/set_bary_helio_times.py:    Python script text executable, ASCII text
scripts/set_telescope_pointing.py:  Python script text executable, ASCII text
scripts/set_velocity_aberration.py: Python script text executable, ASCII text, with CRLF line terminators
scripts/utc_to_bary.py:             Python script text executable, ASCII text
```
which made it not work in SDP.  For some reason.

The solution:
```
$ dos2unix scripts//set_velocity_aberration.py 
dos2unix: converting file scripts//set_velocity_aberration.py to Unix format...
```
and now
```
$ file scripts/*.py
scripts/move_wcs.py:                Python script text executable, ASCII text
scripts/okify_regression_tests.py:  Python script text executable, ASCII text
scripts/set_bary_helio_times.py:    Python script text executable, ASCII text
scripts/set_telescope_pointing.py:  Python script text executable, ASCII text
scripts/set_velocity_aberration.py: Python script text executable, ASCII text
```

jdavies-st:
attn @capadgett

PR #3877, opened 2019-08-02
In support of JP-862: Fetch ASDF Step configurations from CRDS.

dmggh:
Added prototype code to retrieve the config file from CRDS, using look-up of step name from the step class. Still need: to more fully test all strun use cases (including for overrides),  include code to actually retrieve config file from CRDS,  continue work to incorporate/merge config info retrieved from CRDS,  improve code by additional use of argparse, to consider rework in light or Jonathan's recent related PR. This should not be merged.

PR #3921, opened 2019-08-15
Code to incorporate config info from crds into step parameters

stscirij:


jwst/stpipe/cmdline.py
```diff
@@ -16,6 +18,9 @@
     'debug', 'logcfg', 'verbose'
     ]
 
+def get_cfg_from_crds(step, filename):
```
stscieisenhamer: So, I believe this should be superseded by `crds_client.get_config_from_reference` in #3922

jwst/stpipe/cmdline.py
```diff
@@ -223,23 +231,38 @@ def just_the_step_from_cmdline(args, cls=None):
 
     debug_on_exception = known.debug
 
+    #
+    # This creates a config object from the spec file of the step class merged with
+    # the spec files of the superclasses of the step class and adds arguments for
+    # all of the expected reference files
     spec = step_class.load_spec_file(preserve_comments=True)
 
+    parameters_from_crds = get_cfg_from_crds(step_class, '')
```
stscieisenhamer: The updated call would be something like, parts coming from PR #3877, 
```
input_model = datamodels.open(positional[0])
crds_client.get_config_from_reference(step_class, input_model)
```
I'm guessing this will need to occur after `positional` is defined but before `_override_config_from_args` is called.

stscirij: Yes, and I need to convert the dictionary returned from get_config_from_reference to a ConfigObj as merge_config requires ConfigObjs and not dicts.

stscieisenhamer: should a ConfigObj just be returned anyways then? This will have to happen both here and other places.

stscieisenhamer: the other option is updating the ConfigObj directly from a dict as `Configobj.update(dict)`. whichever is more readable works for me.

PR #3922, opened 2019-08-15
Jp862 Implement get_config_from_reference

stscieisenhamer:
Resolves JP-862/#3814 

Implements function `jwst.stpipe.crds_client.get_config_from_reference` to retrieve parameters from CRDS for a given `Step` and dataset.


stscieisenhamer:
I think this is all working like: #3877 had elements of everything, from which, this takes the part about getting the crds data, and encapsulates it in the function. #3921 handles the incorporation into `cmdline` and, wherever the second point we had determined this needed to occur. So between this and #3921, I believe this covers #3877.

dmggh:
Ok, thanks.

stscieisenhamer:
I'm going to merge this in. Seems to cover everything from #3877 and, from the team review, sounds like @stscirij may be using it. If there are still issues, they'll be dealt with in later as issues.

PR #3924, opened 2019-08-15
JP-634 cube build unit test

jemorrison:


jwst/cube_build/file_table.py
```diff
@@ -120,7 +118,8 @@ def set_file_table(self,
                 else:
 
                     log.info('Instrument not valid for cube')
-        return instrument, detector
+#        return instrument, detector
```
stscieisenhamer: Any reason to leave this here? Unless there is a significant reason that is documented, its better practice to just delete such lines

stscieisenhamer:
LOL Tests added, coverage drops by 18%...
image

PR #3928, opened 2019-08-16
WIP Regtest redesign

### LINKED ISSUE #3895: Create a fixture which models the interaction between the test system and artifactory
Issue [JP-915](https://jira.stsci.edu/browse/JP-915) was created by Nadia Dencheva:

Create a fixture which models the interaction between the test system and artifactory

jdavies-st:
This PR is a WIP for redesigning our regression test infrastructure.

#### `RegtestData`
A class that holds state information that allows a test to communicate with bytesalad (Artifactory).  Attributes point to were input and truth data lives on Artifactory, and where outputs from the test live locally.  Also provides some convenience methods to get the input and truth data from Artifactory into the local directory where the test is running.

#### Fixtures that return a `RegtestData` instance to the test:

- `rtdata` returns an instance of `RegtestData` to a test for use.

- `rtdata_module` returns an instance of `RegtestData` to another fixture that can be scoped at function or module level, and then that fixture in turn runs a pipeline or step, and then those results get supplied to individual tests that can test the results of the pipeline or step in various ways.

#### Auto-upload of test outputs on test failure

This is a great feature of our current test infrastructure, and it is only partially implemented here so far.

Resolves #3895.


jdavies-st:
The `conftest.py` and example regression tests are in `jwst/regtest`.  To run them:
```
pytest --bigdata jwst/regtest/ --env=newdev
```

PR #3930, opened 2019-08-19
JP-633 Unit tests were added to `barshadow`

### LINKED ISSUE #3346: Barshadow pipeline step needs unit-testing
Issue [JP-633](https://jira.stsci.edu/browse/JP-633) was created by Macarena Garcia Marin:

The barshadow correction step in the JWST calibration pipeline does currently not have any test in the GitHub repository. To ensure the robustness of the pipeline and have a lower level of risk, the pipeline testing focus group is requesting unit tests to be included and routinely executed. Verifying individual code components will make easier the process of validating the pipeline. The requested code-coverage is 90%.

philhodge:
Closes #3346.

### DIFF
--- a/jwst/barshadow/tests/test_barshadow.py
+++ b/jwst/barshadow/tests/test_barshadow.py
@@ -85,30 +85,28 @@ def test_create_shadow():
     barshadow_model = datamodels.BarshadowModel(data1x1=d1x1, data1x3=d1x3)
     shutter_elements = bar.create_shutter_elements(barshadow_model)
 
-    shutter_status = "0001x1000"
+    shutter_status = "11001x11"
 
     shadow = bar.create_shadow(shutter_elements, shutter_status)
 
     # first
     assert np.allclose(shadow[0:500, :], d1x1[0:500, :], atol=1.e-10)
-    # open_closed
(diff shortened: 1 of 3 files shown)


jwst/barshadow/tests/test_barshadow.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def test_create_shadow():
+
+    d1x1 = rn.random_sample((1001, 101))
+    d1x3 = rn.random_sample((1001, 101))
+    barshadow_model = datamodels.BarshadowModel(data1x1=d1x1, data1x3=d1x3)
+    shutter_elements = bar.create_shutter_elements(barshadow_model)
+
+    shutter_status = "0001x1000"
```
stscirij: I don't think you can have a shutter status with 0s at the ends, this slit would just be "1x1".

philhodge: I appreciate now that this is not realistic, but is this actually a problem?  Using that string tests all of the "open_closed", etc., options.

stscirij: True, for the purposes of the unit tests this is perfectly fine.

PR #3939, opened 2019-08-21
Change stsci.image 'dev' dependency to a stable version floor

rendinam:
...now that a new release of that package is available.


nden:
Thanks @rendinam !

PR #3949, opened 2019-08-27
added unit tests for cube_build

jemorrison:
additional unit tests for cube_build

### DIFF
--- a/jwst/cube_build/tests/test_cube_build_step.py
+++ b/jwst/cube_build/tests/test_cube_build_step.py
@@ -20,84 +20,82 @@ def miri_cube_pars(tmpdir_factory):
     filename = str(filename.join('miri_cube_pars.fits'))
     hdu0 = fits.PrimaryHDU()
     hdu0.header['REFTYPE'] = 'CUBEPAR'
-    hdu0.header['INSTRUME']='MIRI'
-    hdu0.header['MODELNAM']='FM'
-    hdu0.header['DETECTOR']='N/A'
-    hdu0.header['EXP_TYPE']='MIR_MRS'
+    hdu0.header['INSTRUME'] = 'MIRI'
+    hdu0.header['MODELNAM'] = 'FM'
+    hdu0.header['DETECTOR'] = 'N/A'
+    hdu0.header['EXP_TYPE'] = 'MIR_MRS'
 
     # make the first extension
-    channel=np.array(['1','1','1','2','2','2','3','3','3','4','4','4'])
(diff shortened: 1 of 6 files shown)


jwst/cube_build/tests/test_cube_build_step.py
```diff
@@ -0,0 +1,139 @@
...
+from astropy.io import fits
+from jwst.datamodels import IFUImageModel
+from jwst.cube_build import CubeBuildStep
+from jwst.cube_build.file_table import ErrorNoAssignWCS
+from jwst.cube_build.cube_build import ErrorNoChannels
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+
+
+@pytest.fixture(scope='module')
+def miri_cube_pars(tmpdir_factory):
```
mfixstsci: Hey @jemorrison why did you create a the `cube_pars` reference file instead of using the configured file? Is there a `cube_pars` reference file?

jemorrison: @mfixstsci did you mean why did I create it for this test ?
I created it for the test because I did not think it was appropriate to use CRDS in a unit test. 
So I just hacked up a dummy one (but I actually used some of the code that creates the real one)

jemorrison: There is some complex logic in using the reference table and I thought I should check it in a unit test

mfixstsci: Hey @jemorrison, okay great. I was just curious. I noticed that you do this for all of the instruments so I figured there was a good reason for it, thanks!

jwst/cube_build/tests/test_wcs.py
```diff
@@ -173,15 +173,15 @@ def test_setup_wcs():
     footprint = (ra_min, ra_max, dec_min, dec_max, lambda_min, lambda_max)
 
     pars_cube = {
```
mfixstsci: Hey @jemorrison just to capture the reason for the change, why is this change necessary? Does the code behave differently if we use `None` or `0.0`?

mfixstsci: Is this because of the `assert` for `math.isclose`?

jemorrison: As I recall that was the reason -- to make assert for math.isclose work

stscieisenhamer:
@mfixstsci Is this all ok now? If so, could you approve.

mfixstsci:
Hey @stscieisenhamer everything looks good to me. I got all the answers I needed.

PR #3953, opened 2019-08-28
Jp 947

ddavis-stsci:
This is to fix the issue where fixed-slit spectra with a nod pattern excluded observations with PATTYPE=NULL from the level 3 associations. A new level 3 rule was put in place to capture these observations. 
From the tests, 
pytest -n 3 -ra --bigdata --basetemp=/Users/ddavis/Data/JWST/CTeam/test_JP-947 ~/src/JWST/scsb/jwst/jwst/tests_nightly/general/associations/  --junit-xml=results.xml
...
============= 106 passed, 14 skipped, 2 xfailed in 2230.92 seconds =============

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,11 @@ assign_wcs
 associations
 ------------
 
+- Update rules to make level 3 associations for slitless LRS mode [#3940]
+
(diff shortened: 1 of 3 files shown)


stscieisenhamer:
@ddavis-stsci I've approved and merged. If you haven't done so, run the regression suite locally to update artifactory before the next nightly run.

PR #3957, opened 2019-08-28
Regression test infrastructure redesign

jhunkeler:
#### `RegtestData`
A class that holds state information that allows a test to communicate with bytesalad (Artifactory).  Attributes point to were input and truth data files live locally and on Artifactory, and where outputs from the test live locally.

```
RegtestData.input
RegtestData.input_remote
RegtestData.output
RegtestData.truth
RegtestData.truth_remote
```

Also provides some convenience methods to get the input and truth data from Artifactory into the local directory where the test is running.

```
RegtestData.get_data()
RegtestData.get_truth()
```

#### Fixtures that return a `RegtestData` instance to the test:

- `rtdata` returns an instance of `RegtestData` to a test for use.

- `rtdata_module` returns an instance of `RegtestData` to another fixture that can be scoped at function or module level, and then that fixture in turn runs a pipeline or step, and then those results get supplied to individual tests that can test the results of the pipeline or step in various ways.

#### JSON breadcrumbs get left behind on failure
- As in our current system, if a test fails, a JSON specfile gets left behind, and Jenkins slurps these up and uploads the failed outputs to Artifactory
- Additionally, a JSON specfile that allows OKification gets left behind.  We will have to write a small script to utilize these to allow okifying.

Continuation of #3928


jdavies-st:
So a failing test with `assert 0` now uploads its failing output file to artifactory.

https://bytesalad.stsci.edu/artifactory/webapp/#/artifacts/browse/tree/General/jwst-pipeline-results/2019-10-10_jenkins-RT-JWST_RT_REFACTOR-42_0/test_miri_image2/jw00001001001_01101_00001_mirimage_cal.fits

Now to make it less hacky and more general.

jdavies-st:
This is ready for final review.  The more the merrier.

jdavies-st:
I'm not sure how we want to do the regression testing once this is merged.  We could have 2 separate Jenkins jobs.

One for the existing regression tests
```
pytest --bigdata --slow -k "not regtest"
```
and one for the new regression tests
```
pytest --bigdata --slow -k "not tests_nightly" --env=newdev
```
And of course separate coverage flags for each.  Does this seem reasonable?

jwst/regtest/conftest.py
```diff
@@ -0,0 +1,367 @@
...
+    outcome = yield
+    rep = outcome.get_result()
+
+    # set a report attribute for each phase of a call, which can
+    # be "setup", "call", "teardown"
+    setattr(item, "report_" + rep.when, rep)
+
+
+def postmortem(request, want_property):
+    if isinstance(want_property, str):
```
stscieisenhamer: docstring?

jwst/regtest/conftest.py
```diff
@@ -0,0 +1,367 @@
...
+        upload_schema = _func(schema_pattern)
+
+        # Write the schema to JSON
+        jsonfile = os.path.join(cwd, "{}_results.json".format(request.node.name))
+        with open(jsonfile, 'w') as outfile:
+            json.dump(upload_schema, outfile, indent=2)
+
+
+# @pytest.fixture(scope='function', autouse=True)
+def generate_artifactory_okify_json(request, artifactory_repos):
```
stscieisenhamer: docstring?

nden:
It looks good. I second the request for docstrings. It may be helpful also to describe the signature and purpose of the example test functions.

PR #3960, opened 2019-08-29
added one more test and a little pep8 clean up

jemorrison:
Added a simple to test for checking user input options


dmggh:
This looks good. The whitespace revisions look like they should satisfy pep8. Is there any need to test the default power of 1 in test_call_straylight1() ?


jemorrison:
I am leaving testing the default power of 1 for a regression test. Because setting default to reasonable values will result in the step being run to completely which is too much for a unit test.

Issue #3750, opened 2019-06-27
Cube FITS WCS issues with WAVE-TAB


stscijgbot:
Issue [JP-826](https://jira.stsci.edu/browse/JP-826) was created by David Law:

When calwebb_spec2 runs it produces quick-look cubes from each individual exposure using cube_build.  However, the WCS embedded in these data cubes is strange.  The figure below shows the result when I try to align the WCS coordinates between four different cases.  (A) is the result of running cube_build on an association file containing both individual exposures.  The dithered images of the source are combined properly, so clearly the WCS from assign_wcs is ok.  (B) is the result of running cube_build on an association file with just one exposure.  As expected the quality is lower, but the source location matches.  (C) and (D) are the s3d files produced by calwebb_spec2.  In both cases, WCS alignment is finding the lower-left corner of the image instead of the real source location.

Perhaps tellingly, reading (A) and (B) into ds9 works fine, while (C) and (D) give a warning "Warning: the WCS has no defined inverse.  Some functionality may not be available."  Why the difference?  The WCS of these cubes should be entirely defined by the simple fits header keywords, even though (C) and (D) contain information from both Ch1 and Ch2.

On poking at this a little deeper, I think the issue is that (C) and (D) by default use the WAVE-TAB approach.

I don't know whether ds9 works with WAVE-TAB or not, but the contents of the WCS-TABLE extension look weird anyway.  Instead of being 4149 elements long it has shape [1, 10420] where the last 6000 elements are zero.  Why is this?

[~morrison] [~dencheva] [~bushouse]

 

 

!cubes.png!

stscijgbot:
Comment by Jane Morrison: I have been using DS9 Version 8.0 rc5 and this version works with WAVE_TAB. What version are you using.

The default of calwebb_spec2 for a file is to produce a single IFU cube containing both channels. These cubes will have a WAVE_TAB and an non-linear wavelength range. 
So when you say you used dithered data, I am assuming you ran calwebb_spec3. The default IFU Cube here are single Band cubes so they have a linear wavelength range. 

The reason is the WCS-TABLE extension looks strange is because the only way (right now) that we can get  the non-linear WAVE_TAB to work in the pipeline is to hard code the WAVE_TAB to have 10402 elements
to the size [1, 10420] if no data exists for a wavelength plane then 0 is written. I believe we will need to
update astropy the option to produce non-linear wavelengths using WAVE_TAB. That is on the the to-do list, but not a simple task. 

[...]

PR #3976, opened 2019-08-30
Allow variable shape arrays in schemas for structured array fields

mcara:
This PR improves the logic in the `datamodels.properties._cast()` function so that now it can handle structured arrays who's fields can be arbitrarily shaped arrays. Previously schemas for structured arrays required a fixed shape to be defined which in turn led to issues such as https://github.com/spacetelescope/jwst/issues/3750

In addition, the new logic also allows specifying the ``ndim`` constrain in the schemas for the fields of structured arrays.

Finally, this PR modified the code for the ``cube_build`` so that wavetables are saved using arrays of appropriate length `=>` DS9 "frame matching by WCS" now should work with cubes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -51,6 +51,12 @@ calwebb_tso3
 
 - Update to exclude target_acquisitions from processing in the calwebb_tso3 pipeline [#3759]
 
+cube_build
+----------
+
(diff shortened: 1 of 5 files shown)


jemorrison:
Wonderful.

mcara:
@jdavies-st @jemorrison Thanks!

PR #3981, opened 2019-08-31
Remove unnecessary string-to-string conversions

### LINKED ISSUE #3978: Remove conversion of strings to strings
I have encountered numerous instances of cases when strings are converted to strings of the kind `str('string')` such as:

https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/tests/test_schema.py#L348-L350

This kind of conversion is unnecessary and should be removed.

mcara:
This PR removes unnecessary string-to-string conversions.

Closes #3978


mcara:
@jdavies-st Yeah, likely it was the result/consequence of `from future import unicode_literals`

PR #4020, opened 2019-09-13
Revert "Fixes JP-974 Footprint calculation for archive "

stscicrawford:
Reverts spacetelescope/jwst#4010

This is being reverted as this did not yet address the problem and created additional regressions.   This should be reviewed further and the root cause of the problem should be address.

### DIFF
--- a/jwst/cube_build/cube_build_step.py
+++ b/jwst/cube_build/cube_build_step.py
@@ -6,7 +6,7 @@ from . import cube_build
 from . import ifu_cube
 from . import data_types
 from ..assign_wcs.util import update_s_region_keyword
-from . import cube_build_wcs_util
+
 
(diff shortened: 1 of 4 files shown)


jemorrison:
Yes I agree lets revert this pr. In addition,  I have spent all day trying to figure out why the single IFU cubes cause a failure now  in mrs_imatch. These single cubes have a linear wavelength range so a different problem the errors we see with Wave Table form. By reverting this pr we can see if something unexpected was introduced by this pr.

stscicrawford:
Can you approve this @jemorrison ? Then I can revert

stscicrawford:
I also think I might have found some clues  -- I'll try to summarize and post them soon

PR #4021, opened 2019-09-14
JP-974 : temporary fix for crash in cube_build

nden:
This will eliminate the crash in cube_build although there might be some regression tests failing with differences. I suspect there's a deeper reason for the failure and this should be discussed further.


stscicrawford:
Closing for the time being but will re-open if further needed.

PR #4022, opened 2019-09-16
JP-622 Unit tests for photom have been added

philhodge:


jwst/photom/tests/test_photom.py
```diff
@@ -0,0 +1,1470 @@
...
+            (data, dq, err, var_p, var_r, var_f) = mk_data(shape)
+            input_model = datamodels.ImageModel(data=data, dq=dq, err=err)
+            # There is no wavelength attribute for ImageModel, but this
+            # should work anyway.
+            wl = mk_wavelength(shape, 0.6, 4.0, dispaxis=1)
+            input_model.wavelength = wl.copy()
+            input_model.var_poisson = var_p
+            input_model.var_rnoise = var_r
+            input_model.var_flat = var_f
+        else:
```
hbushouse: What exptype does this last branch get used for? Or is it just a catch-all for any random exptype?

jwst/photom/tests/test_photom.py
```diff
@@ -0,0 +1,1470 @@
...
+                input_model.slits.append(slit.copy())
+        else:                                   # NRC_IMAGE
+            (data, dq, err, var_p, var_r, var_f) = mk_data((128, 256))
+            input_model = datamodels.ImageModel(data=data, dq=dq, err=err)
+            input_model.var_poisson = var_p
+            input_model.var_rnoise = var_r
+            input_model.var_flat = var_f
+    elif instrument == 'MIRI':
+        if exptype == 'MIR_MRS':
+            (data, dq, err, var_p, var_r, var_f) = mk_data((128, 256))
```
hbushouse: Will the photom routines work OK on MIRI MRS data that isn't full-frame (you're creating a 128x256 subarray here)? If I recall correctly, one or more of the photom ref files for MRS mode is a full-frame image containing the flux calibration factors, hence it might not be possible to multiply it into a science data subarray.

jwst/photom/tests/test_photom.py
```diff
@@ -0,0 +1,1470 @@
...
+    max_r : float
+        Maximum value to assign when populating the relresponse array.
+
+    Returns
+    -------
+    ftab : `~jwst.datamodels.DataModel`
+        An open data model for a MIRI LRS photom reference file.
+    """
+
+    filter = ["F560W", "F770W", "F1000W"]
```
hbushouse: If this routine is being used to create a photom table for the LRS mode, then the filter value should be "P750L", which is the MIRI prism location in the filter wheel. The ones listed here are imaging filters.

jwst/photom/tests/test_photom.py
```diff
@@ -0,0 +1,1470 @@
...
+    ratio = output[iy, ix] / input[iy, ix]
+    result.append(math.isclose(ratio, compare, rel_tol=1.e-7))
+    assert np.alltrue(result)
+
+
+def test_miri_lrs():
+    """Test calc_miri, LRS data"""
+
+    input_model = create_input('MIRI', 'MIRIMAGE', 'MIR_LRS-FIXEDSLIT',
+                               filter='F770W')
```
hbushouse: For LRS, filter needs to be "P750L".

jwst/photom/tests/test_photom.py
philhodge: I added a comment to say that this is for NIS_IMAGE.

jwst/photom/tests/test_photom.py
philhodge: I wouldn't think so, no.  This only works because I created the "reference file" as well, and the function that creates it takes the image shape as an argument.

jwst/photom/tests/test_photom.py
philhodge: Thanks!  I fixed this, here and in line 1299.

philhodge:
> Everything looks good and quite exhaustive. Do you think you need to test photom_io() and save_area_info() separately or do you think that the tests you have done will take care of that?

Yes, these are tested.  I realize it's not obvious, though.  `photom_io` is called for each of the `calc_xxx` functions.  In `test_nirspec_fs` and `test_nirspec_bright` the corrections to the error and variance arrays are also checked; this is only done for these two functions because that covers the two sections in `photom_io` where the corrections are applied to those arrays.  With regard to `save_area_info`, testing that is the reason for the separate function `test_apply_photom`, using imaging data.  This is the only test using an area reference file (though just a dummy model, as in all the other tests in this module).

Issue #3999, opened 2019-09-09
regression test failure: One of the requested xi is out of bounds in dimension 0


stscijgbot:
Issue [JP-974](https://jira.stsci.edu/browse/JP-974) was created by Steven Crawford:

Three regression tests in the cube building step are failing with the following or similar error: '

One of the requested xi is out of bounds in dimension 0'

nden:
@jemorrison The `Tabular` model in astropy.modeling allows extrapolation by passing 
`bounds_error=False` to the initializer. I don't know wether extrapolation is a good thing here, just pointing out the capability.

stscijgbot:
Comment by Jane Morrison: The regression test is failing when cube_build is creating the types of cubes that have non-linear wavelength dimension. After the IFU cube is created the last line of the cube_build_step
updates the spatial wcs with this code:
footprint = cube.meta.wcs.footprint(axis_type="spatial")
this is where the error is occurring 
I think Nadia has pointed us in the right direction. It is probably related to her comment.
I could use a bit more information. For the "Tabular" model in astropy if "bounds_error" is False does that mean no extrapolation ?   How can we change it ? From the error of "One of the requested xi is out of bounds in dimension 0"  is it not clear if what type of extrapolation is being done.  It is occurring in /Users/morrison/anaconda/envs/dev_test/lib/python3.6/site-packages/scipy/interpolate/interpolate.py", 
 Since this is related to the "spatial" dimensions I am not sure why any extrapolating or interpolating is being done anyway. 
Could [~dencheva] or [~mcara] maybe expand on what they think is happening.

stscijgbot:
Comment by Mihai Cara: {quote}is it not clear if what type of extrapolation is being done{quote}

I do not know much about footprints ([~dencheva] is the expert), but if you are talking about _analytical_ non-linear models, I assume that _that_ model is being used. For example, if lambda=z^2 for 0<=z<1024 then for z=2000 the same model will be used => lambda=2000^2.

nden:
@jemorrison Is the spatial transform of a cube a `Tabular` model from modeling?

stscijgbot:
Comment by Jane Morrison: No the spatial dimensions should not be 'Tabular' only the wavelength dimension.
If the wavelength dimension is non-linear then the wavelength WCS is changed as follows:
ifucube_model.meta.wcsinfo.ctype3 = 'WAVE-TAB'
[...]

PR #4030, opened 2019-09-17
Fix a bug in casting of structured arrays

mcara:
Fixes a bug described in https://github.com/spacetelescope/jwst/issues/3999#issuecomment-532276865 This bug was introduced in a bug fix to a bug fix in #3995. Essentially the structured array schema validation AND type coercion code was (incorrectly) skipped when input is already a structured array.

CC: @stscicrawford @jemorrison

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -85,6 +85,10 @@ datamodels
   data types defined by schema of structured arrays when input values are not
   native Python types (tuples). [#3995]
 
+- Fixed an issue with the fix from [#3995] that was affecting "casting" to
+  data types defined by schema of structured arrays when input values are
+  already structured arrays. [#4030]
+
 exp_to_source
 -------------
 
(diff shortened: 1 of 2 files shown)


jemorrison:
I just pulled this fix into my branch and re-ran cube_build. I no longer get an error.
The IFU cube look good and have the correct wavelength range.
Super fantastic.

stscicrawford:
I've tried this on one of the failing tests, and the test runs to completion with what looks like the need to update the truth files.   

@jemorrison Can you please run the other regression tests and check the validity of the output files?

stscicrawford:
Thanks @jemorrison for the check and @mcara for the fix!

jemorrison:
Yep I will run it on all the regression test involving cube_build

jemorrison:
The failing regression tests now pass.  I am now updating the truth files.

jemorrison:
truth files on the 3 failing tests have been updated. I re-ran the regression tests and they all pass now

PR #4033, opened 2019-09-17
Calibration changed from dividing to multiplying across all modes.

dmggh:
This is part of the work needed to support JP-988 and JP-796.


stscicrawford:
I'm just marking this as 'Do not merge' until we have the regression tests passing as this will likely cause failures there as we do not have the new photom reference files submitted yet either.   So this will have to be merged coordinated with that work.

philhodge:
> This will need unit tests updated and a change log

There are comments in test_photom.py regarding the changes that are needed when replacing division with multiplication.

hbushouse:
Still just needs a change log entry.

dmggh:
Updated change log.

stscicrawford:
Closing as this has been incorporated into #4118

PR #4052, opened 2019-09-20
JP-1014: New imaging photom data models

hbushouse:
Initial set of changes to accommodate new photom reference file data models for imaging modes. The old data models for MIRI, NIRCam, and NIRISS (``MiriImgPhotomModel``, ``NircamPhotomModel``, and ``NirissPhotomModel``) are still in place, to allow for backwards compatibility, at least temporarily. The new data models for these 3 instruments are ``MirImgPhotomModel``, ``NrcImgPhotomModel``, and ``NisImgPhotomModel``.

The existing FGS data model ``FgsPhotomModel`` was modified in-place and hence will not work with the existing photom reference files in CRDS. This is going to cause regression test errors.


hbushouse:
@stscieisenhamer One potential solution to the issue with incompatibility of the modified FGS data model is to use a new class name for the modified model and leave the old one available for a while, for backwards compatibility. This is the approach taken with the new ``MirImgPhotomModel``, ``NrcImgPhotomModel``, and ``NisImgPhotomModel``: they're all new model names and hence the old ones are still available and unchanged. We could name the new FGS model ``FgsImgPhotomModel`` instead of the more generic ``FgsPhotomModel``. There's only 1 FGS photom data model to begin with and it only applies to imaging mode, because FGS only has imaging modes, so there really isn't a need to explicitly identify the FGS model with the "Img" label, but it doesn't hurt.

hbushouse:
Reviewers:

- The new models no longer inherit from the generic `PhotomModel`, because nearly every model is unique to its instrument mode, so there's no good reason to even have a base class anymore (seemed like a good idea 5 years ago). They inherit from `ReferenceFileModel` instead, just like all other reference file data models.

- The file name of the existing `MiriImgPhotomModel` was changed to include the extra "i" (miri instead of just mir), so that the file names for the new `MirImgPhotomModel` could use the 3-letter version of "mir".

hbushouse:
> In miriimg_photom.schema.yaml, should line 3 be $ref: referencefile.schema.yaml instead of photom.schema.yaml ?

No, sorry the MIRI changes were hard to follow, especially if you're just looking at diffs. The original MIRI data model was named ``MiriImgPhotomModel``, but the name of the schema file for that model was "mirimg_photom.schema.yaml" (the former has the 4-letter Miri, while the latter has the 3-letter mir). We wanted to use the 3-letter version for the new data model, so I renamed the existing "mirimg_photom.schema.yaml" to "miriimg_photom.schema.yaml", which still contains the old data model. And because the old models still use ``PhotomModel`` as a base class, the schema needs to reference "photom.schema.yaml" instead of "referencefile.schema.yaml". Only the new schemas reference "referencefile.schema.yaml."

PR #4053, opened 2019-09-22
Mrs imatch unit test

jemorrison:
Added a set of unit tests for mrs_imatch step
Updated _get_2d_pixgrid  because channel 1 and 4 are on the left side of the detector and channel 2 and 3 are on the right side


jemorrison:
@mcara 
Can you take a look at the modifications to get_2d_pixgrid. Channel 1 and 4 are on the left side of the detector and channel 2 and 3 are on the right side of the detector

jemorrison:
@philhodge 
could you look at test_apply_background.py. I am calling assign_wcs -which takes time but I did not know of another way to get the wcs set up correctly for this test. If you know of a quicker way let me know.

philhodge:
> @philhodge
> could you look at test_apply_background.py. I am calling assign_wcs -which takes time but I did not know of another way to get the wcs set up correctly for this test. If you know of a quicker way let me know.

You can create a wcs.  Just write a function and assign it to model.meta.wcs.  I've done this for a unit test, but I don't remember where.  I'll keep looking.

stscicrawford:
There does seem to be errors related to the test_configuration.py file with the travis tests.

stscieisenhamer:
@jemorrison The reason for the failure is that there is no `__init__.py` file in the `mrs_imatch/tests` directory. Add one and all will be right in the world. To check, you can do the following command locally:

```
pytest jwst/cube_build/tests/test_configuration.py jwst/mrs_imatch/tests/test_configuration.py
```
Suggestion courtesy of @jdavies-st

jemorrison:
I will add that file but I did not add the __init__.py to the test directory for the cube_build tests or the straylight tests and those passed travis testing. I guess I should go back and add this file to those tests ? I looked at a couple of other test directories and they don't have __init__.py Maybe they need too

Issue #3541, opened 2019-05-29
IRS2 extra reference pixels are not readout with TAs


stscijgbot:
Issue [JP-737](https://jira.stsci.edu/browse/JP-737) was created by Rosa Diaz:

According to Charles Proffitt:

"Note that when doing "IRS2 TA" the extra reference pixels are not actually read out as they are for regular IRS2 observations. So the TA images downlinked to the ground will look like non-IRS2 images. Other than the possible need for special versions of the reference files, the DMS calibration requirements for the "IRS2 TA" images should be the same as they are for regular TA images."

This will affect all TAs taken with the following read patterns NRSIRS2RAPID, NRSIRS2RAPIDD1, NRSIRS2RAPIDD2, and NRSIRS2RAPIDD6 available for TA

Does the Calibration Pipeline need to be changed to work on this case?

stscijgbot:
Comment by Howard Bushouse: If the image data from these TA's go through DMS with the substring "IRS2" contained anywhere in their READPATT keyword value, it will require many changes to both CAL and CRDS. There are several calwebb_detector1 pipeline steps that rely on the READPATT value to determine whether to handle the data in IRS2 format. If the READPATT says "IRS2", but the data aren't actually in IRS2 mode (i.e. the interleaved columns of reference pixels aren't in the data array), then the pipeline will crash because it'll assume the extra reference pixel columns are there. Additional hooks would be needed in the pipeline to figure out that the READPATT value is essentially lying to us and to treat the data as regular images instead.

There are also several NIRSpec CRDS reference files that use READPATT as a selector and in particular have unique reference files for the IRS2 patterns. This includes the DARK, MASK, READNOISE, REFPIX, and SUPERBIAS reference file types. So changes will be necessary in the reference file selection process too, in order to have "normal" reference files returned instead of the IRS2-specific reference files that would normally be chosen.

An example data set would help a lot in order for us to inspect the various meta data (keyword values) and from there decide how and what to change to accommodate these TA images.

stscijgbot:
Comment by Rosa Diaz: According to the linked JSOCINT-192 issue:
 * NIRSpec TA exposures with EXP_TYPE = ‘WATA’ or ‘MSATA’ and IRS2 data mode will not have reference pixels.
 * Process these TA/IRS2 exposures as if they were NRSRAPID.

[~bushouse], can the code be changed to look for substring IRS2 that does not have "D1", "D2", and "D6" or something along these lines?

[...]

PR #4054, opened 2019-09-23
JP-737 Don't use PATTTYPE to check for IRS2 mode

philhodge:
For science data, the value of the PATTTYPE keyword will contain "IRS2" if and only if the data were read out using IRS2.  Target acquisition data are never read out using IRS2, but PATTTYPE can still contain "IRS2".  Some but not all tests for IRS2 mode already use the image shape rather than the PATTTYPE keyword.

A new `is_irs2` function has been added to lib/pipe_utils.py, and the nearly identical function of the same name has been removed from the x_irs2.py files in ipc and saturation.  This function returns True if one axis of the input image is larger than 2048 (it will be 3200, for the current IRS2 parameters).  Files in refpix, ipc, and saturation have been modified to call the version of `is_irs2` in lib.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -89,9 +89,6 @@ datamodels
   data types defined by schema of structured arrays when input values are
   already structured arrays. [#4030]
 
-- Added "MIR_TACONFIRM" to the list of allowed EXP_TYPE values in the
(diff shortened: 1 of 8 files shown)


hbushouse:
Fixes #3541

PR #4063, opened 2019-09-25
Pin dependencies in requirements-sdp.txt for nightly regression test builds

### LINKED ISSUE #3037: Specfile for the calibration software for SDP
Issue [JP-503](https://jira.stsci.edu/browse/JP-503) was created by Steven Crawford:

For the installation and running of the science data pipeline, there needs to be a stable environment and set of dependencies.   Testing against different dependencies for users can be done in different ways, but installation and nightly regression testing done for the science data pipeline should be against this set of dependencies. 

Here are a few requirements that should be considered:  
 # At the start of a development cycle for a build, the dependencies that will be run in the pipeline should be set.  
 # The spec file should be part of the JWST pipeline repository. In this way, anyone working on the JWST pipeline can easily test changes to dependencies and can easily update the file.  
 # Each branch released for a build should have its own specfile.  This make its easy to test and avoid dependency updates especially during the middle of a build or for patch updates. 

In this ticket, a specfile is defined as the packages used  to build the JWST pipeline by using conda.   As the dependencies will be set and not changing during a build development cycle for the SDP, these do not have to built or use conda if there are better solutions.

Please add any other suggestions to this ticket that you think would be helpful.

### LINKED ISSUE #3988: Question on intended use of install_requires in setup.py
An approach being validated to provide environment deliveries to DMS generates an environment dump at the conclusion of each successful regression testing run. The nightly regression testing job performs a `pip install -e .[test,ephem]` for the 'stable-deps' build configuration (as seen in `JeninsfileRT`). This obtains dependencies as per the `install_requires` list in `setup.py`. However, currently there are three non-stable/unreleased dependencies in that list:

`astropy @ git+https://github.com/astropy/astropy@5f7c192#egg=astropy`
`gwcs @ git+https://github.com/spacetelescope/gwcs@3e2bc108e#egg=gwcs`
`photutils @ git+https://github.com/astropy/photutils@99f101be#egg=photutils`

The development snapshot dependencies above are problematic for this approach as they cannot be obtained from PyPI when the environment is to be recreated elsewhere, due to their not corresponding to released packages.

I notice there is a `requirements-dev.txt` file that seems to be intended to accommodate development snapshot dependencies, and a JWST-UNSTABLE Jenkins job that might be a more appropriate place to use the unreleased dependencies noted above.

Is it possible to designate the `stable-deps` build configuration and the `install_requires` list in `setup.py` to be used only for stable dependency sets, and by extension, only include stable dependencies in the `install_requires` list of `setup.py`?

attn: @jdavies-st, @nden

jdavies-st:
This will pin the dependencies (and dependencies of dependencies) for our regression test builds.  These pinned dependencies can then be used by SDP builds when we have successful regression test runs.

Should we also add a pinned dependency build to our per-PR CI tests (Travis and/or JenkinsCI)?

Resolves #3037 
Resolves #3988 
Closes #3747


stscicrawford:
This would meet my requirements for SDP installation/regression testing.   It might be good to add a description of this to the README as well.

jdavies-st:
So it turns out that `pymssql` does not like to get installed into a MacOS python 3.6 conda env.  It barfs.

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/142/execution/node/168/log/
```
Collecting pymssql==2.1.4 (from -r requirements-sdp.txt (line 26))
  Downloading https://files.pythonhosted.org/packages/2e/81/99562b93d75f3fc5956fa65decfb35b38a4ee97cf93c1d0d3cb799fffb99/pymssql-2.1.4.tar.gz (691kB)
    ERROR: Command errored out with exit status 1:
     command: /users/iraf/build/boyle/workspace/RT/jdavies-dev/miniconda/envs/tmp_env1/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/fy/_s99d0l14k598wxq803bjjr800000g/T/pip-install-0yt5fchc/pymssql/setup.py'"'"'; __file__='"'"'/private/var/folders/fy/_s99d0l14k598wxq803bjjr800000g/T/pip-install-0yt5fchc/pymssql/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base pip-egg-info
         cwd: /private/var/folders/fy/_s99d0l14k598wxq803bjjr800000g/T/pip-install-0yt5fchc/pymssql/
    Complete output (7 lines):
    /users/iraf/build/boyle/workspace/RT/jdavies-dev/miniconda/envs/tmp_env1/lib/python3.6/site-packages/setuptools/dist.py:45: DistDeprecationWarning: Do not call this function
      warnings.warn("Do not call this function", DistDeprecationWarning)
      File "<string>", line 1, in <module>
      File "/private/var/folders/fy/_s99d0l14k598wxq803bjjr800000g/T/pip-install-0yt5fchc/pymssql/setup.py", line 88, in <module>
        from Cython.Distutils import build_ext as _build_ext
    ModuleNotFoundError: No module named 'Cython'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```
I can reproduce this locally too.  Basically it will install under python 3.7 conda env, but not 3.6 on MacOS.

So this means our `requirements-sdp.txt` file should not include this dependency for MacOS.  Which is probably fine as it is only used in the `timeconversion` script to access an SDP database.  I.e. SDP uses it under Linux, but nobody (that we know of) uses it in the MacOS build, including us developers.

Based on discussion with @rendinam, I think we can exclude it from the pinned dependencies in `requirements-sdp.txt` and instead just pin it the `[ephem]` extras, so that it can be installed like `pip install .[ephem]` if needed.

jdavies-st:
Build on MacOS now works.  A regression test run on this branch are the same 10 failures we are seeing on master.  So I think this is good.  Will separately update the README and the Wiki.

Issue #4056, opened 2019-09-23
test_call_straylight1 intermittently fails


stscijgbot:
Issue [JP-1015](https://jira.stsci.edu/browse/JP-1015) was created by James Davies:

If one runs the `jwst` unit tests using `pytest-xdist`, the failure occasionally happens:
```
__________________________________________________________________ test_call_straylight1 ___________________________________________________________________
[gw0] darwin -- Python 3.7.4 /Users/jdavies/miniconda3/envs/jwst_yaml/bin/python

_jail = None, miri_mrs = <IFUImageModel(20, 20) from test_miri.fits>

    def test_call_straylight1(_jail, miri_mrs):
        """ test possible user options are set up correctly """
    
        collect_pipeline_cfgs('./config')
    
        # Test the ModShepard power is in the correct range
        # set the power to outside the upper range
        step = StraylightStep.from_config_file('config/straylight.cfg')
        step.override_straylight = 'dummy.asdf'
        step.method = 'ModShepard'
        step.power = 6
>       result = step.run(miri_mrs)

/Users/jdavies/dev/jwst/jwst/straylight/tests/test_straylight_step.py:53: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/Users/jdavies/dev/jwst/jwst/stpipe/step.py:385: in run
    step_result = self.process(*args)
/Users/jdavies/dev/jwst/jwst/straylight/straylight_step.py:37: in process
    " 0.1 to 5. It is set to {}".format(self.power))
/Users/jdavies/dev/jwst/jwst/associations/lib/log_config.py:168: in log
    self._log(level, line, args, **kwargs)
/Users/jdavies/miniconda3/envs/jwst_yaml/lib/python3.7/logging/__init__.py:1514: in _log
    self.handle(record)
/Users/jdavies/miniconda3/envs/jwst_yaml/lib/python3.7/logging/__init__.py:1524: in handle
    self.callHandlers(record)
/Users/jdavies/miniconda3/envs/jwst_yaml/lib/python3.7/logging/__init__.py:1586: in callHandlers
    hdlr.handle(record)
[...]

Issue #2159, opened 2018-06-14
Step architecture review/refactor


stscieisenhamer:
# Issue
In as much that most of the pipelines and steps needed for launch are implemented or at least defined, and external users are beginning to use the code, and that version 1.0 delivery is approaching, a number of issues regarding `Step` and pipeline development have been revealed and should be discussed and resolved.

This issue will be used as simply a collection and discussion of the set of current concerns regarding `Step` design, API, and usage. 

Please to update the list below with issues/questions/concerns to be addressed.
- `Step` API
    - Roles of `call`, `run`, and `process` #1098 
    - Refactor of `run`
    - reference file collection
    - `Step` responsibilities/granularity/conventions #1699 #1096 #943 
- I/O
    - Multiple input/outputs, both file and parameters #1694 
    - implicit versus explicit file I/O: Currently an uncomfortable mix
    - step input validation #1826 
- Configuration
    - dual configuration identity crisis or "where is truth"? In the `cfg` or in the `spec`? #965
    - configuration discoverability/documentation
    - config file generation
- Documentation
    - versioning #2014
    - Archive #2118
- Multiprocessing #1364 #1096 
- Crazy
    - Step/Pipeline factories based on input

melanieclarke:
Closing, since JP-818 is closed.

PR #4064, opened 2019-09-25
updated how straylight step checks power parameter

jemorrison:
Addresses part of #4056


jemorrison:
James - I think does what we want.
I added the suggestion you made. Took out the log.error. But I kept the log.warning. That is ok right ? The step is skipped if the power is ever out of range.

stscieisenhamer:
@jdavies-st could you add that subject to the #2159?

jdavies-st:
Test failure is some CRDS server issue in the JenkinsCI run that is unrelated to this PR.  Merging.

PR #4068, opened 2019-09-26
Add exptype keywords to new imaging photom data models

hbushouse:
#4052 created new data models for imaging-mode photom reference files, in support of JP-1014. That PR, however, forgot to include the CRDS selector keywords - EXP_TYPE, P_EXP_TY, and BAND - the schema, so that they're available in the data models. This PR fixes that omission.

It's likely the case that only the P_EXP_TY keyword is needed in the reference file headers, but it can't hurt to have both EXP_TYPE and P_EXP_TY defined, in case anyone wants to populate EXP_TYPE too.


hbushouse:
Jenkins failures appear to be unrelated and Travis tests finally passed, so I'm merging.

PR #4069, opened 2019-09-26
Added tests for the leastsqnrm module

dmggh:
For functions in the leastsqnrm module in the ami_analyze step, these are the unit tests.  There are 2 functions in that module that I did not write tests - they are custom least squares functions that are too complicated for unit tests, so they may be put into regression tests later.


nden:
There's one flake8 error on Travis that needs to be fixed.

PR #4070, opened 2019-09-26
JP-860: filter out asdf warnings related to extension version mismatch

### LINKED ISSUE #3810: Filter out warnings from asdf
Issue [JP-860](https://jira.stsci.edu/browse/JP-860) was created by Nadia Dencheva:

When files are opened as data models the following type of warnings are printed:

 

site-packages/asdf/asdf.py:202: UserWarning: File was created with extension 'gwcs.extension.GWCSExtension' from package gwcs-0.10.1.dev524, but older version gwcs-0.10.0 is installed
 warnings.warn(fmt_msg)

 

Asdf has an option to ignore these types of warnings and it should be propagated to the pipeline to allow the warnings to be suppressed.

nden:
When a file is read with an asdf extension version older than the version it was created with asdf raises a `Warning` like this one:
```
asdf/asdf.py:202: UserWarning: File 'poly.asdf' was created with extension 'astropy.io.misc.asdf.extension.AstropyAsdfExtension' from package astropy-4.0.dev25781, but older version astropy-3.2.1 is installed
  warnings.warn(fmt_msg)

```
This type of warning is raised also when reading a FITS file with a datamodel under those circumstances. This is really just an informational warning. Given the level of backwards compatibility and versioning in asdf the usual case is that the file is read and converted correctly. In fact, if it wasn't converted, an `Error` is raised. So given this I suggest we propagate the flag to the pipeline and set it to `True`, i.e. do not raise the warning by default.

Resolves #3810.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -88,6 +88,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
                                                     pass_invalid_values)
         self._strict_validation = self.get_envar("STRICT_VALIDATION",
                                                  strict_validation)
+        self._ignore_missing_extensions = ignore_missing_extensions
 
         kwargs.update({'ignore_missing_extensions': ignore_missing_extensions})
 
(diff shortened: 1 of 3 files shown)


jwst/datamodels/model_base.py
```diff
@@ -74,6 +74,11 @@ def __init__(self, init=None, schema=None,
         strict_validation: if true, an schema validation errors will generate
             an excption. If false, they will generate a warning.
```
hbushouse: Not a functional problem, but while you're modifying this module anyway, we should fix the typos in lines 74-75. Change to: "if true, schema validation errors will generate an exception. If false, ..."

nden: Will do. Actually I meant to fix all docstrings in the `__init__` since they are not correctly formatted.

nden:
@hbushouse The last commit is only docstring changes.

nden:
Ha, this is what we just discussed and I think we agreed we need to change the default behavior in asdf. It will take some time to implement though and the ability to control this flag within the pipeline is still needed.

nden:
Can this be merged?

PR #4071, opened 2019-09-27
Straylight unit test remove test now handled with ConfigObj

jemorrison:
@jdavies-st 
I removed from the unit test the tests on power and roi values. That was causing a regression test failure - conflict with ConfigOji


jemorrison:
closing this issue. It was resolved by #4064

PR #4075, opened 2019-09-30
Speed up ramp fit unit tests by using smaller arrays

jdavies-st:
No need for full-sized arrays in unit test when a single pixel is being tested.   This speeds up the tests significantly.  For example, the slowest:
```
17.08s call     jwst/ramp_fitting/tests/test_ramp_fit.py::TestMethods::test_nocrs_noflux[GLS]
```
And now:
```
0.20s call     jwst/ramp_fitting/tests/test_ramp_fit.py::TestMethods::test_nocrs_noflux[GLS]
```
Overall the runtime on my machine goes from 119 seconds down to 5 seconds.


karllark:
This seems like a reasonable update to me.  Might want to check with Mike Regan (@mwregan2) as he created the tests originally.

PR #4081, opened 2019-09-30
Add subarray keyword(s) to new Miri imaging photom model

hbushouse:
Yet another update to the new MIRI photom ref file data model for imaging mode, to include the subarray-related keywords in the schema, because subarray is used as a CRDS selector for MIRI photom ref files. In continued support of JP-1004 and JP-1014.


hbushouse:
Never mind!  SUBARRAY is NOT needed as a CRDS reference file selector (only as a table row selector) and hence the keyword is not needed in the header of the photom reference files. Closing this PR. Please ignore (nothing to see here - move along).

PR #4090, opened 2019-10-01
Jp 799 Integrate retrieval of Step parameters from CRDS

stscieisenhamer:


jwst/datamodels/steppars.py
```diff
@@ -0,0 +1,34 @@
...
+    },
+}
+
+
+class StepParsModel(DataModel):
+    """
+    A data model for `Step` parameters.
+    """
+    schema_url = "steppars.schema"
+    supported_formats = ['yaml', 'json', 'asdf']
```
jdavies-st: What does `supported_formats` do?  And does the datamodel actually have a serializer that can write these out as JSON or YAML?

jwst/datamodels/steppars.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def __init__(self, init=None, **kwargs):
+        super().__init__(init=init, **kwargs)
+        meta = copy(DEFAULT_META)
+        meta.update(self.meta.instance)
+        self.meta.instance.update(meta)
```
jdavies-st: Is there a reason this copy-update-update is done rather than relying on the schema to take care of default values?  If a default value is specified in the schema, then it should get populated, right?  Seems like the defaults are located in to places here.

jwst/stpipe/tests/test_step.py
```diff
@@ -5,10 +5,128 @@
...
+
+REFPIXSTEP_CRDS_MIRI_PARS = {
+    'class': jwst.refpix.refpix_step.RefPixStep,
+    'name': 'refpix',
+    'odd_even_columns': False,
+    'odd_even_rows': False,
+    'side_gain': 10.0,
+    'side_smoothing_length': 21,
+    'use_side_ref_pixels': False
+}
```
jdavies-st: This is only referenced in the test below.  Perhaps it should live in the test below and not be a global variable?

jwst/datamodels/steppars.py
stscieisenhamer: Absolutely nothing apparently. Will remove.

jwst/datamodels/steppars.py
stscieisenhamer: And the schema would take care of it. Will remove.

Note: These are leftover from, and still is the case, that the datamodel is not finished.

stscieisenhamer: Oh, now I remember. This was done to force the meta to be completely populated. Just having the schema definition was insufficient. Is there a way to get ever required schema element to actually exist in a newly created model?

jwst/stpipe/tests/test_step.py
stscieisenhamer: Maybe just one test today, but may be used by many tomorrow. *shrug* either way.

jwst/datamodels/schemas/steppars.schema.yaml
```diff
@@ -0,0 +1,77 @@
...
+            title: Instrument used to acquire the data
+            type: string
+            enum: [NIRCAM, NIRSPEC, MIRI, TFI, FGS, NIRISS, ANY, N/A]
+          detector:
+            title: Name of detector used to acquire the data
+            type: string
+            enum: [NRCA1, NRCA2, NRCA3, NRCA4, NRCALONG, NRCB1, NRCB2, NRCB3, NRCB4,
+              NRCBLONG, NRS1, NRS2, ANY, MIRIMAGE, MIRIFULONG, MIRIFUSHORT,
+              NIS, GUIDER1, GUIDER2, N/A]
+            description: Detector name.
```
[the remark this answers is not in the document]
stscieisenhamer: Actually, having the whole core schema may be appropriate. How would having the definitions of `fits_keyword` affect things? Or would this be benign enough when the format is just 'asdf'?

jwst/datamodels/steppars.py
jdavies-st: If you define defaults in the schema, then you can access them to get them populated in the instantiated `DataModel`.  It is the reason that there are those
```
    self.data = self.data
    self.dq = self.dq
    self.err = self.err
```
lines in some datamodels that have defaults set for these attributes.

jwst/datamodels/schemas/steppars.schema.yaml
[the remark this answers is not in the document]
jdavies-st: `fits_keyword` has no effect on the ability to write out the tree as an ASDF file.

stscieisenhamer:
Since all the review commentary are on the `Steppars` datamodel, which is not yet part of the configs-from-crds process, will go ahead and merge. There is an existing issue, [jp-904](https://jira.stsci.edu/browse/JP-904), for dealing with the datamodel.

PR #4096, opened 2019-10-02
JP-1032: Update spectroscopic photom data models

### LINKED ISSUE #4088: Create new spectroscopic mode photom reffile datamodels
Issue [JP-1032](https://jira.stsci.edu/browse/JP-1032) was created by Jonathan Eisenhamer:

Create new spectroscopic mode photom reffile datamodels

hbushouse:
Baseline version of updated/new spectroscopic mode photom reference file data models, per the specs outlined in JP-1018. The discussion of the specs is not quite finalized, so there may be some further updates before merging.

As with the imaging mode photom data models, a little rearranging of MIRI model class names was necessary to maintain backwards compatibility with the models and reffiles currently in use. The old MIRI MRS model is `MiriMrsPhotomModel`, which uses schema ``mirimrs_photom.schema.yaml``, while the new one is `MirMrsPhotomModel` and uses schema ``mirmrs_photom.schema.yaml`` (note that the only difference in the names is the extra "i" after "Mir" in the old model).

Fixes #4088

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -109,6 +109,7 @@ datamodels
   'astropy.io.misc.asdf.extension.AstropyAsdfExtension' from package astropy-4.0.dev24515,
   but older version astropy-3.2.1 is installed``. [#4070]
 
+- Added new spectroscopic mode photom reference file data models. [#4096]
 
 exp_to_source
 -------------
--- a/jwst/datamodels/__init__.py
+++ b/jwst/datamodels/__init__.py
@@ -41,9 +41,13 @@ from .nirspec_flat import NirspecFlatModel, NirspecQuadFlatModel
 from .outlierpars import OutlierParsModel
 from .pathloss import PathlossModel
 from .persat import PersistenceSatModel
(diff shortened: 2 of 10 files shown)


hbushouse:
Note that in addition to creating new spectroscopic data model classes, the new models are also taking advantage of the recent updates to `datamodels` that no longer require the shape/size of arrays imbedded in tables to be hardwired in the schema. So the hardwired sizes of the ``wavelength``, ``relresponse``, and ``reluncertainty`` arrays have been removed and instead the schema simply indicates that they must be 1-dimensional (``ndim: 1``).

hbushouse:
Fixes #4088

hbushouse:
Probably no drawbacks to using "data" that I can think of. The main drawback right now is the fact that the teams need to start building their ref files NOW and I doubt they'd be too thrilled with a last minute change. In a way it doesn't hurt to distinguish the content of this model as `phot_table`, since it is a completely unique entity relative to other data models that have standard `data`, `err`, and `dq` components.

PR #4100, opened 2019-10-02
Jp 1036 Update number of associations in test_generate to match number after NRS_MIMF fix (JP1000)

ddavis-stsci:
Update the number of expected associations in nightly tests test_generate from the current 98
to the 99 being created.


ddavis-stsci:
closes #4100

PR #4104, opened 2019-10-04
Replace RELIC with setuptools_scm

jhunkeler:
Note: 
I retained `__version_commit__` because `stpipe` stores the value as metadata. When not in development mode (i.e. no `.dev#+gHASH.dDATE` exists in `__version__`), the value is left as an empty string. If the string is empty `stpipe` will emit `RELEASE` instead of `unknown`


pllim:
Where is stpipe actually using this? Isn't stpipe upstream, so why would it care about jwst version?

PR #4107, opened 2019-10-07
Fix __version_commit__ regression

### LINKED ISSUE #4106: Interrupt: Regression after switching from relic to scm
Issue [JP-1040](https://jira.stsci.edu/browse/JP-1040) was created by Nadia Dencheva:

After switching from relic to scm many regression tests are failing because FITS keyword CAL_VCS is no longer populated.  According to the core schema:

CAL_VCS: Calibration software version control sys number

Trying to figure out what exactly this is supposed to be and if it's still needed.

jhunkeler:
Fixes the following bug introduced by #4104 :

```
[...]/validate.py:35: ValidationWarning: While validating calibration_software_revision the following error occurred:
  ('5a2e91e3',) is not of type 'string'
  
  Failed validating 'type' in schema:
      {'$schema': 'http://stsci.edu/schemas/asdf-schema/0.1.0/asdf-schema',
       'blend_table': True,
       'fits_keyword': 'CAL_VCS',
       'title': 'Calibration software version control sys number',
       'type': 'string'}
  
  On instance:
      ('5a2e91e3',)
    warnings.warn(errmsg, ValidationWarning)
```
Resolves #4106

### DIFF
--- a/jwst/__init__.py
+++ b/jwst/__init__.py
@@ -13,7 +13,7 @@ except DistributionNotFound:
 if '+' in __version__:
     commit = _regex_git_hash.match(__version__).groups()
     if commit:
-        __version_commit__ = commit
+        __version_commit__ = commit[0]
 
 if sys.version_info < (3, 5):
     raise ImportError("JWST requires Python 3.5 and above.")
--- a/jwst/tests/test_version.py
+++ b/jwst/tests/test_version.py
@@ -1,0 +1,8 @@
+def test_version_is_string():
+    from jwst import __version__
+    assert isinstance(__version__, str)
+
+
+def test_version_commit_is_string():
+    from jwst import __version_commit__
+    assert isinstance(__version_commit__, str)


nden:
Do I understand correctly that previously for releases `CAL_VCS` was set to `unknown` and after  #4104 it will be set to `RELEASE`?

jhunkeler:
That's correct, `CAL_VER` was permanently set to `unknown` when `relic` didn't initialize `jwst/version.py` (i.e. RTD builds). `setuptools_scm` doesn't create a separate module to import, so it's bit easier to manage.

Total failure: `__version_commit__ = ''`
Tagged release: `__version_commit__ = 'RELEASE'`
Post-tag development: `__version_commit__ = 'abcd1234`

nden:
OK, so for tests run on releases we will see this difference. I would think for releases it's better to set it to `__version__`?  But that may be decided and changed in a separate PR.

jhunkeler:
If we ever decide to print a version string, as a user, I'd rather see `1.0.0-RELEASE` than `1.0.0-1.0.0` :grin:

PR #4111, opened 2019-10-08
improve slit footprint computation

### LINKED ISSUE #3929: Improve footprint calculation for MOS data
Issue [JP-942](https://jira.stsci.edu/browse/JP-942) was created by Nadia Dencheva:

Improve footprint calculation for MOS data

nden:
Resolves #3929 

The footprint is computed using the transform from "slit_frame" to "world" .

TODO: 
[x] Verify the order of corners in `S_REGION`
   It looks like th eorder here is correct - clockwise - which means the rest of the footprints are incorrect. I'll open an issue about it.
[ X] Find out what is the best order of corner coordinates for mosviz
   ~After some off-line discussion it's still unclear what mosviz needs. Best if a new ticket is filed for that work.~ Got a confirmation that this is good for mosviz.

This will produce failures in some regression tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -143,6 +143,8 @@ extract_2d
 - For NIRCam TSO data, wavelengths are computed and assigned to the
   wavelength attribute. [#3863]
 
+- Improved the computation of ``S_REGION`` of a slit. [#4111]
+
 flat_field
 ----------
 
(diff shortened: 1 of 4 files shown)


nden:
@jhunkeler Are we somehow back to the previous coverage report?
I only added two comments :smirk:

PR #4113, opened 2019-10-09
ccd-601 consolidate CRDS imports

jaytmiller:
Consolidated all direct imports of CRDS modules in the jwst.stpipe.crds_client.   Imports of the crds.core.exceptions module were deferred/delayed behind the crds_client.get_exceptions_module() getter function.
This helps manage the import order of circular CAL and CRDS imports.
The imports are circular due to the CAL dependence on CRDS for best refererences,  and the CRDS dependence on CAL for datamodels support.

### DIFF
--- a/jwst/stpipe/crds_client.py
+++ b/jwst/stpipe/crds_client.py
@@ -30,6 +30,10 @@
 """This module defines functions that connect the core CRDS package to the
 JWST CAL code and STPIPE, tailoring it to work with DATAMODELS as inputs
 and provide results in the forms required by STPIPE.
+
+WARNING:  JWST CAL and CRDS have circular dependencies.  Do not use CRDS imports
+directly in modules other than this crds_client so that dependency order and
+general integration can be managed here.
 """
 
 import re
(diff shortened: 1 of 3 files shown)


jwst/stpipe/crds_client.py
```diff
@@ -38,6 +42,12 @@
 from crds.core import config, exceptions, heavy_client, log
 from crds.core import crds_cache_locking
 
+def get_exceptions_module():
+    """Provide external indirect access to the crds.core.exceptions module to
+    alleviate issues with circular dependencies.
+    """
+    return exceptions
+
```
jdavies-st: Is there a reason that one needs a function to return `exceptions`?  Since it is imported in `crds_client`, can't you just to the imports elsewhere as
```
>>> from jwst.stpipe.crds_client import exceptions
>>> raise exceptions.CrdsLookupError('foo')
crds.core.exceptions.CrdsLookupError: foo
```

jaytmiller: CRDS and JWST CAL have circular imports;  it's a code smell but I don't think there's an easy fix without some mutual refactoring:   CRDS needs datamodels.   CAL needs best references.

When conflicts have arisen I've resolved them with hand rolled lazy imports.

get_exceptions_module() decouples import timing so that 
      timing(crds_client import) !- timing (exceptions import).   
exceptions is not imported until absolutely needed.    

I want to maintain the pattern that crds imports only happen in crds_client  to help mitigate what I'll dub "circular import spaghetti".   If all the interfaces to crds_client are kept as functions,  all the CRDS imports in crds_client can be made lazy later only updating that one module.

PR #4114, opened 2019-10-09
Update to StepParsModel and minor documentation tweaks

stscieisenhamer:
Changes:
- Update `StepParsModel` to remover unnecessary code
- Update docstrings to remove sphinx errors

Notes:
The sphinx errors encountered when running locally are
not what RTD is seeing. Cannot reproduce RTD errors
but maybe this will push things a bit.

### DIFF
--- a/jwst/datamodels/steppars.py
+++ b/jwst/datamodels/steppars.py
@@ -27,7 +27,7 @@ class StepParsModel(DataModel):
     schema_url = "steppars.schema"
 
     def __init__(self, init=None, **kwargs):
-        super(StepParsModel, self).__init__(init=init, **kwargs)
+        super().__init__(init=init, **kwargs)
         meta = copy(DEFAULT_META)
         meta.update(self.meta.instance)
         self.meta.instance.update(meta)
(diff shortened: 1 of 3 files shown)


jwst/datamodels/steppars.py
```diff
@@ -25,10 +25,9 @@ class StepParsModel(DataModel):
     A data model for `Step` parameters.
     """
     schema_url = "steppars.schema"
-    supported_formats = ['yaml', 'json', 'asdf']
 
     def __init__(self, init=None, **kwargs):
-        super().__init__(init=init, **kwargs)
+        super(StepParsModel, self).__init__(init=init, **kwargs)
```
jdavies-st: This is unnecessary since python 3.  The original implementation is fine.

jwst/stpipe/step.py
```diff
@@ -23,6 +23,8 @@
 except ImportError:
     DISCOURAGED_TYPES = None
 
+from crds.core import exceptions
```
jdavies-st: To avoid circular dependencies, use `from jwst.stpipe.crds_client import exceptions`

jwst/datamodels/steppars.py
stscieisenhamer: ya, i was just trying this to see jp-1041 was affected by this. However, I have since figured out what is the basis of that issue.

PR #4117, opened 2019-10-10
JP-1035 JP-1041 Make properties pars and pars_models into methods

stscieisenhamer:
Resolves JP-1035 and, hopefully, JP-1041

Hate it when documentation build issues point out questionable code...not that defeat has yet been admitted.

### DIFF
--- a/jwst/lib/class_property.py
+++ b/jwst/lib/class_property.py
@@ -1,6 +1,65 @@
 """Allow property access to both instance and class attributes"""
+from functools import partial
 
-class ClassProperty:
+
+class ClassInstanceMethod:
+    """Allow a method to be both class and instance
+
+    This defines a descriptor that maps a function to
+    both the class and instance.
+
+    Parameters
+    ----------
+    fget: function
+        The function to use as the `__get__` method.
+
+    Notes
+    -----
+    On class access, only the `__get__` method of a descriptor is used.
+    All other descriptor methods are ignored. As such, allowing the definitions
+    of `__set__` and `__delete__` are not allowed.
+
+    Examples
+    --------
+    Usage is the same as any other class or instance method. However, access
+    is available from both the class and instances.
+
+    >>> class MyClass:
+    ...     @ClassInstanceMethod
+    ...     def baz(obj):
+    ...         return obj._baz
+    ...
+    ...     _baz = 'class'
+
+    >>> MyClass.baz
+    'class'
+
+    >>> mc = MyClass()
+    >>> mc.baz()
+    'class'
+
+    >>> mc._baz = 'instance'
+    >>> mc.baz()
+    'instance'
+
+    >>> MyClass.baz()
+    'class'
+    """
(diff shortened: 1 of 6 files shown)


jwst/lib/class_property.py
```diff
@@ -1,6 +1,65 @@
 """Allow property access to both instance and class attributes"""
+from functools import partial
 
-class ClassProperty:
+
+class ClassInstanceMethod:
+    """Allow a method to be both class and instance
```
nden: Pyhton class methods can be run on the class and the instance. It's sufficient to use the `@classmethod` decorator. 
So I guess I don't understand the purpose of this and how it's going to be used.

stscieisenhamer: The purpose is to have the method operate as both a class method and an instance method. When called from a class, like `classmethod`, have the first argument be the class. However, when called from an instance, have the first argument behave as `self` or the instance. Example below:

```
import inspect
from jwst.lib.class_property import ClassInstanceMethod

class A:
     @classmethod
     def only_class(cls):
         print(f'only_class: cls == {cls}')
         print('cls is class? {}'.format(inspect.isclass(cls)))
         print('type(cls) == {}'.format(type(cls)))

     @ClassInstanceMethod
     def both(obj):
         print(f'both: obj == {obj}')
         print('obj is class? {}'.format(inspect.isclass(obj)))
         print('type(obj) == {}'.format(type(obj)))
```

Then the following occurs. The first example is calling `only_class` and `both` on the class `A`. As intended, both methods behave the same.

```
>>> A.only_class()
only_class: cls == <class '__main__.A'>
cls is class? True
type(cls) == <class 'type'>

>>> A.both()
both: obj == <class '__main__.A'>
obj is class? True
type(obj) == <class 'type'>
```

In the second example, the methods are called on an instance of `A`. As expected, the classmethod `only_class` operates the same as above: The first argument is the class. However, `both` now operates as an instance method: The first argument is not the class, as above, but is the instance.

```
>>> a = A()

>>> a.only_class()
only_class: cls == <class '__main__.A'>
cls is class? True
type(cls) == <class 'type'>

>>> a.both()
both: obj == <__main__.A object at 0x1106e9510>
obj is class? False
type(obj) == <class '__main__.A'>
```

For the particular usage, this allows the methods `get_pars` and `get_pars_model` to operate on both a `Step` class and a `Step` instance. When given a class, the parameters will have values as given by the `spec` parameters. When used on an instance, those parameters will have the instance values, modified by configuration files and arguments used to instantiate the instance.

jdavies-st: I guess this is required by the not-very-good way that we instantiate `spec`.  It is defined with defaults and validated at the class level, but once you've instantiated the class, these are instance attributes that can be changed to anything without validation.

A better approach in the future would be to be a bit more pythonic and define these attributes as properties that get their defaults on `__init__` and have validation occur in the setter.

stscieisenhamer:
Are there any further comments or suggested changes by the reviewers?

PR #4118, opened 2019-10-10
Modify the photom step to use new data models

philhodge:
The `photom` step has been modified to use new `photom` data models.  This version is incompatible with the older `photom` data models.  The changes include:

For spectroscopic data, the data are multiplied by the 2D relative response term; previously the data were divided by this term.  The relevant code changes for this were copied from PR #4033.

For NIRSpec spectroscopic data for an extended target, the conversion factor will be divided by the pixel area in steradians, because the value in the new photom reference file converts to MJy, but we need MJy / sr.  This will also be done for NIRISS SOSS data if the target is extended, but that case should never be encountered.

The pixel area is gotten from the area reference file and copied to metadata.  This was done before but only for imaging data; now it is also done for spectroscopic data.  Another difference is that the pixel area is only gotten from the area reference file; pixel area keywords in the `photom` reference file are no longer checked.


jwst/photom/photom.py
```diff
@@ -634,7 +649,24 @@ def photom_io(self, tabdata, order=None):
 
         """
         # Get the scalar conversion factor from the PHOTMJSR column of the table row
-        conversion = tabdata['photmjsr']
+        try:
```
dmggh: Maybe in the docs include a short explanation of what's going on in this new code.

philhodge: Good point, thanks!  Yes, I'll update the docs.

docs/jwst/photom/main.rst
```diff
@@ -44,22 +45,30 @@ For spectroscopic data, if the photom step finds that the wavelength and relativ
 response arrays in the reference table row are populated, it loads those 1-D arrays
 and interpolates the response values into the 2-D space of the science image based
 on the wavelength at each pixel.
+For NIRSpec spectroscopic data and for NIRISS SOSS, the conversion terms in
+the photom reference file give results in flux density (MJy).  For point
+sources, the output of the photom step will be in these units.  For extended
+sources, however, the values will be divided by the average solid angle of a
```
hbushouse: Might want to just mention that the source type determination comes from the value of the SRCTYPE keyword set earlier in the pipeline flow (and for NIRSpec MOS/SLIT is determined on a slit-by-slit basis).

docs/jwst/photom/main.rst
```diff
@@ -44,22 +45,30 @@ For spectroscopic data, if the photom step finds that the wavelength and relativ
...
 the variance (VAR_POISSON, VAR_RNOISE, and VAR_FLAT) arrays.
-The correction values are divided into the SCI and ERR arrays, and the square
-of the correction values are divided into the variance arrays.
+The correction values are multiplied into the SCI and ERR arrays, and the square
+of the correction values are multiplied into the variance arrays.
 
 The scalar conversion constant is copied to the header keyword PHOTMJSR, which
-gives the conversion from DN/s to MegaJy/steradian that was applied to the data.
+gives the conversion from DN/s to megaJy/steradian (or to megajanskys, for
+NIRSpec and NIRISS SOSS extended sources, as described above) that was
```
hbushouse: The units are megajanskys for *point* sources, not extended.

docs/jwst/photom/main.rst
```diff
@@ -68,36 +77,37 @@ The process of attaching the pixel
...
+retrieves the scalar conversion constant, as well as the 1-D wavelength and
 relative response arrays, from that row. It also loads the IFU pixel area
 data from the pixel area reference file.
 
 It then uses the scalar conversion constant, the 1-D wavelength and relative
 response, and pixel area data to compute a 2-D sensitivity map (pixel-by-pixel)
 for the entire science image. The 2-D SCI and ERR arrays in the science
-exposure are divided by the 2D sensitivity map, which converts the science
+exposure are multiplied by the 2D sensitivity map, which converts the science
 pixels from units of DN/sec to mJy/arcsec\ :sup:`2`\ .
```
hbushouse: Will the NIRSpec IFU data still come out in mJy/arcsec^2, or will all modes be standardized to result in MJy or MJy/steradian? I think the units info here is outdated.

docs/jwst/photom/main.rst
```diff
@@ -68,36 +77,37 @@ The process of attaching the pixel
...
 MIRI MRS
 --------
 
 For the MIRI MRS mode, the photom reference file contains 2-D arrays of sensitivity
 factors and pixel sizes that are loaded into the step. As with NIRSpec IFU, the
 sensitivity and pixel size data are used to compute a 2-D sensitivity map
-(pixel-by-pixel) for the entire science image. This is divided into both
+(pixel-by-pixel) for the entire science image. This is multiplied into both
 the SCI and ERR arrays of the science exposure, which converts the pixel values
 from units of DN/sec to mJy/arcsec\ :sup:`2`\ .
```
hbushouse: Same question about resulting units. I'm guessing everything will now be in MJy/sr.

docs/jwst/references_general/photom_reffile.inc
```diff
@@ -36,130 +36,218 @@ The FITS primary HDU does not contain a data array.
...
++============+=======+================+===========+============+========================+
+| NIRCam     | WFSS  | filter         | string    | 12         | N/A                    |
++            +       +----------------+-----------+------------+------------------------+
+|            |       | pupil          | string    | 15         | N/A                    |
++            +       +----------------+-----------+------------+------------------------+
+|            |       | photmjsr       | float     | scalar     | MJy/steradian/(DN/sec) |
++            +       +----------------+-----------+------------+------------------------+
+|            |       | uncertainty    | float     | scalar     | MJy/steradian/(DN/sec) |
 +            +       +----------------+-----------+------------+------------------------+
 |            |       | order          | integer   | scalar     | N/A                    |
```
hbushouse: A minor nit, but the order of the columns has changed in the actual schema. It's 1) filter, 2) pupil, 3) order, 4) photmjsr, 5) uncertainty, 6) nelem, 7) wavelength, 8) relresponse, 9) reluncertainty. So "order" needs to be moved up a couple rows in this table.

jwst/photom/photom.py
```diff
@@ -233,14 +240,14 @@ def calc_nirspec(self, ftab, area_fname):
...
+                        self.input.data *= sens2d
+                        self.input.err *= sens2d
+                        self.input.var_poisson *= sens2d**2
+                        self.input.var_rnoise *= sens2d**2
                         if self.input.var_flat is not None and np.size(self.input.var_flat) > 0:
-                            self.input.var_flat /= sens2d**2
+                            self.input.var_flat *= sens2d**2
 
                         # Update BUNIT values for the science data and err
                         self.input.meta.bunit_data = 'mJy/arcsec^2'
```
hbushouse: I think this may need updating to the new units, i.e. MJy or MJy/sr, depending on the source type.

docs/jwst/photom/main.rst
philhodge: Thanks for your comments.  I'll fix all of these.

hbushouse:
There's also a MIRI MRS unit test that needs fixing (that's what's causing all the Travis failures)

philhodge:
> There's also a MIRI MRS unit test that needs fixing (that's what's causing all the Travis failures)

It was due to the change to divide by the pixel area instead of multiplying.  I thought I had run the unit tests after making that change, but obviously not.
I have fixed the unit test; thanks.

drlaw1558:
Following the approach for the other modes and outlined at https://jira.stsci.edu/browse/JP-1018, the new MIRI MRS PHOTOM files include the pixel area correction in the SCI extension so we should now have 
sens2d = ftab.data
It would be helpful to still read in and store the pixel area information, but it no longer needs to be included explicitly in sens2d since it is folded into ftab.data.
Likewise, units are now MJy/sr so the self.input.meta.bunit_data and self.input.meta.bunit_err should be updated.

philhodge:
@drlaw1558 thanks for the info about the pixel area for MIRI MRS.  I fixed this.

drlaw1558:
Thanks @philhodge 
Two remaining questions:
1) Can we also change the units declared in photom.py to be MJy/sr?  Using this PR cube_build is incorrectly reporting units as mJy/arcsec2, presumably since this is the information that it is being passed.
2) Can you remind me how the conv_factor is designed to be used?  If multiplying by sens2d is now putting the data in units of MJy/sr then should conv_factor be equal to 1?  I'm trying to make sure that this is set consistently in our new reference files.

philhodge:
@drlaw1558 yes, of course.  I had changed the documentation but then forgot to change the code.  Thanks for pointing this out.
Regarding `conv_factor` and keywords PHOTMJSR and PHOTUJA2, we are aware of the discrepancy between the keyword value and what was applied to the data, but the units are implied by the keyword names.  We have been thinking that these keywords are no longer useful, and that it might be better to remove them entirely.  This issue is on the agenda for the DMSWG meeting for Oct 23.

drlaw1558:
Ok, thanks @philhodge !
For the time being I'll modify the keywords so that I think they're consistent with the rest of the file, but anticipate that they might be dropped entirely soon.

PR #4120, opened 2019-10-10
Adding guider cds unit tests

### LINKED ISSUE #3351: Guider_cds pipeline step needs unit-testing
Issue [JP-641](https://jira.stsci.edu/browse/JP-641) was created by Macarena Garcia Marin:

The guider_cds step in the JWST calibration pipeline does currently not have any test in the GitHub repository. To ensure the robustness of the pipeline and have a lower level of risk, the pipeline testing focus group is requesting unit tests to be included and routinely executed. Verifying individual code components will make easier the process of validating the pipeline. The requested code-coverage is 90%.

mfixstsci:
Resolves #3351

### DIFF
--- a/jwst/guider_cds/tests/test_guider_cds.py
+++ b/jwst/guider_cds/tests/test_guider_cds.py
@@ -29,15 +29,14 @@ def test_get_dataset_info(make_guider_image):
 
     imshape, n_int, grp_time, exp_type = get_dataset_info(model)
 
-    assert (imshape == (10,10) and
-            n_int == model.data.shape[0] and
-            grp_time == model.meta.exposure.group_time and
-            exp_type == model.meta.exposure.type)
+    assert imshape == (10,10)
+    assert n_int == model.data.shape[0]
+    assert grp_time == model.meta.exposure.group_time
+    assert exp_type == model.meta.exposure.type
 
 
 def test_guider_cds_fineguide_mode(make_guider_image):
-    """Test the fine guiding mode
-    """
+    """Test the fine guiding mode"""
 
     model = make_guider_image
     model.meta.exposure.type = 'FGS_FINEGUIDE'
@@ -89,6 +88,8 @@ def test_guider_cds_acq_track_modes(exptype, make_guider_image):
 
 @pytest.mark.parametrize("exptype", ['FGS_ID-IMAGE', 'FGS_ID-STACK'])
 def test_guider_cds_id_modes(exptype, make_guider_image):
+    """Test fgs id exptypes"""
+
     model = make_guider_image
     model.meta.exposure.type = exptype
 
@@ -136,8 +137,8 @@ def test_table_extensions(make_guider_image):
 
     result = guider_cds(model)
 
-    assert(hasattr(result, 'planned_star_table') and
-           hasattr(result, 'flight_star_table') and
-           hasattr(result, 'pointing_table') and
-           hasattr(result, 'centroid_table') and
-           hasattr(result, 'track_sub_table'))
+    assert 'planned_star_table' in result
+    assert 'flight_star_table' in result
+    assert 'pointing_table' in result
+    assert 'centroid_table' in result
+    assert 'track_sub_table' in result
--- a/jwst/guider_cds/tests/test_guider_cds.py
+++ b/jwst/guider_cds/tests/test_guider_cds.py
@@ -1,0 +1,73 @@
+import numpy as np
+import pytest
+
+from jwst import datamodels
+from jwst.guider_cds.guider_cds import get_dataset_info, guider_cds
+
+
+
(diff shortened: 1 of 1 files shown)


mfixstsci:
This is ready for review

jwst/guider_cds/tests/test_guider_cds.py
```diff
@@ -0,0 +1,143 @@
...
+
+@pytest.fixture(scope='module')
+def make_guider_image():
+    """Generate science image"""
+
+    image = datamodels.GuiderRawModel()
+
+    image.meta.instrument.name = 'FGS'
+    image.meta.exposure.frame_time = 234.3423235
+    image.meta.exposure.ngroups = 4
```
philhodge: It looks as if ngroups is actually 10, based on the image shape (it's the number of integrations that's 4).  It isn't used except for a log message, so it probably doesn't matter.

jwst/guider_cds/tests/test_guider_cds.py
```diff
@@ -0,0 +1,143 @@
...
+    result = guider_cds(model)
+
+    assert result.meta.bunit_data == 'DN/s'
+
+def test_table_extensions(make_guider_image):
+    """Test that tables are assigned to result of pipeline"""
+
+    model = make_guider_image
+
+    model.planned_star_table = np.arange(0, 11)
```
philhodge: This is interesting; I'm surprised that it works.  These are tables, so I would expect the schema verification to fail.  I'm not really concerned, as long as it continues to work.

jwst/guider_cds/tests/test_guider_cds.py
```diff
@@ -0,0 +1,143 @@
...
+
+    for num_int in range(0, n_int):
+        data_sect = model.data[num_int, :, :, :]
+        first_4 = data_sect[:4, :, :].mean(axis=0)
+        last_4 = data_sect[-4:, :, :].mean(axis=0)
+        slope_int_cube[num_int, :, :] = last_4 - first_4
+
+        truth = slope_int_cube /  model.meta.exposure.group_time
+
+    assert np.array_equal(result.data, truth)
```
philhodge: It might not be safe to check for equality, since these values are float.  You could use `np.allclose` instead.

jwst/guider_cds/tests/test_guider_cds.py
```diff
@@ -0,0 +1,143 @@
...
+    model.centroid_table = np.arange(0, 11)
+    model.track_sub_table = np.arange(0, 11)
+
+    result = guider_cds(model)
+
+    assert(hasattr(result, 'planned_star_table') and
+           hasattr(result, 'flight_star_table') and
+           hasattr(result, 'pointing_table') and
+           hasattr(result, 'centroid_table') and
+           hasattr(result, 'track_sub_table'))
```
jdavies-st: This would better be broken up into individual asserts:
```
assert "planned_star_table" in result
assert "flight_star_table" in result
assert "pointing_table" in result
```
etc

jwst/guider_cds/tests/test_guider_cds.py
```diff
@@ -0,0 +1,143 @@
...
+
+    model = make_guider_image
+    model.meta.exposure.type = 'FGS_FINEGUIDE'
+
+    imshape, n_int, grp_time, exp_type = get_dataset_info(model)
+
+    assert (imshape == (10,10) and
+            n_int == model.data.shape[0] and
+            grp_time == model.meta.exposure.group_time and
+            exp_type == model.meta.exposure.type)
```
jdavies-st: Again, this would be better broken up into individual `assert` statements, as it will be easier to debug if something fails.

jwst/guider_cds/tests/test_guider_cds.py
mfixstsci: I see what you mean, I can change it now if you'd prefer it to have the same shape as the data.

jwst/guider_cds/tests/test_guider_cds.py
mfixstsci: Will make the change, thanks Phil!

jwst/guider_cds/tests/test_guider_cds.py
mfixstsci: Sounds good @jdavies-st, I went ahead and made the changes

jwst/guider_cds/tests/test_guider_cds.py
mfixstsci: I see, that makes sense. Changes made

mfixstsci:
I went ahead and broke up the assert statements and the changed the numpy functions that compare the arrays. If there is anything else, let me know @jdavies-st @philhodge

PR #4121, opened 2019-10-10
Unit tests for the analyticnrm2 module within the ami_analyze pipeline

dmggh:
Additional tests for ami_analyze. These test all but a few functions (pending) in the analyticnrm2 module.


jwst/ami/tests/test_ami_analyze.py
```diff
@@ -592,29 +594,201 @@ def test_hexee_glimit():
...
+        -0.50277564+0.86441695j ]]
+      )
+
+    assert_allclose( result, true_result, atol=1E-8 )
+
+
+def setup_SF():
+    ''' Initialize values for these parameters needed for the analyticnrm2 tests:
+
+        pixel (optional, via **kwargs) : float
```
philhodge: Is the "(optional, via **kwargs)" a cut & paste error?
You probably ought to include a "Returns" label and underline to make it clear that these are output values.

PR #4131, opened 2019-10-14
JP-1054: Fix mistake in calwebb_spec2 docs

hbushouse:
Changed the comment about the order of the `extract_2d` and `flat-field` steps to remove the mistaken reference to the `background` step.


hbushouse:
Fixes #4132

PR #4134, opened 2019-10-14
Use python 3.7 in regtest builds

### LINKED ISSUE #4124: Minimum versions of jwst dependencies
Issue [JP-1046](https://jira.stsci.edu/browse/JP-1046) was created by Nadia Dencheva:

Minimum versions for next jwst delivery

 

python: ~ 3.7

numpy: 1.16

 

Test configuration and installation files should be modified to use these versions.

jdavies-st:
In addition, `setup.py` has been updated to requiring:

`python>=3.6` <- an astropy requirement
`numpy>=1.16`

Resolves #4124

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@
 
 - Remove ``jwpsf`` module. [#3791]
 
+- Update dependencies ``python>=3.6`` and ``numpy>=1.16``. [#4134]
+
 assign_wcs
 ----------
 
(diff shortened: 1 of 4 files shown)


hbushouse:
When we recently discussed using 3.7.x for B7.4, it was suggested that we use 3.7.4 (IIRC). I see 3.7.2 referenced here. Is there a preference for one or the other?

jdavies-st:
Oops!  That's a mistake.  Thanks for catching it @hbushouse.  Will update.

Issue #3990, opened 2019-09-04
MIR_LRS-SLITLESS asn table is listing dedicated bkg exposures


stscijgbot:
Issue [JP-972](https://jira.stsci.edu/browse/JP-972) was created by David Davis:

nOPS5-OTB-20190522 porcessing is failing when the master background step is using the association file, [+jw00623-o026_20190808t215150_spec3_001.json+|https://pljwdmsweb.stsci.edu/owlgui/logs/directory_view?directory_path=/ifs/archive/ops/jwst/info/owl/logs/owlmgr_jw00623-o026_20190808t215150_spec3_001_1565329935.595263]. Dithering isn't allowed for LRS-SLITLESS mode so it is not clear that data should ever be taken in a way that allows for the master_background step. So the asn file should no contain dedicated background exposures. The attached files are from the A-string run that exhibits this error. This is reported in JWSTDMS-65

stscijgbot:
Comment by Rosa Diaz: This data is no-TSO and hence that is why we added a dither, as indicated in [https://innerspace.stsci.edu/display/jwstdms/DMS+Test+Data+Needs] MinAsn-13

stscijgbot:
Comment by Howard Bushouse: Per the discussion in JWSTDMS-65, we should change the level-3 ASN rules so that exposures with EXP_TYPE='MIR_LRS-SLITLESS' and TSOVISIT='F' do NOT have level-3 ASN's created, which will effectively stop the calibration of these types of exposures at the end of calwebb_spec2 processing.

stscijgbot:
Comment by Rosa Diaz: This data is non-TSO and hence that is why we added a dither, as indicated in [https://innerspace.stsci.edu/display/jwstdms/DMS+Test+Data+Needs] MinAsn-13

PR #4136, opened 2019-10-15
Jp 972 Exclude mir_lrs-slitless calibration data from level 3 processing

ddavis-stsci:
Update association rules to exclude exposures with EXP_TYPE='MIR_LRS-SLITLESS' and TSOVISIT='F' from level 3 associations. 

The local association tests all pass, 
pytest  ~/src/JWST/scsb/jwst/jwst/associations/tests
...
============ 133 passed, 2 xfailed, 1 warnings in 350.94s (0:05:50) ============

The regression tests currently fail and these four associations will need to be removed from the truth files. 
```
jw00623-a3001_20190607t021101_spec3_002_asn.json
jw00623-o023_20190607t021101_spec3_001_asn.json
jw00623-o024_20190607t021101_spec3_001_asn.json
jw00623-o026_20190607t021101_spec3_001_asn.json
```


jwst/associations/lib/rules_level3.py
```diff
@@ -277,6 +275,22 @@ def __init__(self, *args, **kwargs):
...
+                        name='restricted_slitless',
+                        sources=['exp_type'],
+                        value = ('mir_lrs-slitless')
+                    ),
+                    DMSAttrConstraint(
+                        name='tso_obs',
+                        sources=['tso_visit'],
+                        value = ('T')
+                    ),
+                ],
```
hbushouse: So, the effect of these constraints is to only allow exposures with both EXP_TYPE='MIR_LRS-SLITLESS' and TSO_VISIT=True to be included in a level-3 ASN? If so, that seems to do the trick.

hbushouse:
Fixes #3990

jwst/associations/lib/rules_level3.py
ddavis-stsci: That is correct. Observation 27 has an entry with EXP_TYPE=MIR_LRS-SLITLESS &  TSO_VISIT=T and level 3 products are created,  jw00623-o027_tso3_001_asn.json

Issue #4135, opened 2019-10-15
Add MIR_TACONFIRM to ASN rules


stscijgbot:
Issue [JP-1055](https://jira.stsci.edu/browse/JP-1055) was created by Howard Bushouse:

JWSTDMS-73 reports that EXP_TYPE='MIR_TACONFIRM' exposures will be added as confirmation images for MIRI LRS fixed-slit (MIR_LRS-FIXEDSLIT) observations. The level-2 and level-3 ASN rules need to be updated to include these exposures in the ASN's generated for LRS fixed-slit observations, with member type set to target-acquisition.

stscijgbot:
Comment by Howard Bushouse: This is high priority for including in B7.4

stscijgbot:
Comment by David Davis: Updated the ASN rules so that MIR_TACONFIRM is treated as a target acquisition. Since we don't have an example pool file for this I manually modified jw00623_20190625t115752_pool.csv so that one entry has the EXP_TYPE changed from MIR_TACQ to MIR_TACONFIRM.
 {{}}
{noformat}
jw00623029003_02101_00001_mirimage_uncal.fits|V00623029003P0000000002101|623|29|3|00623029003|2|PRIME_TARGETED_FIXED|1|01|1|MIR_TACONFIRM{noformat}
{{}}

The diff command shows the level 2 & 3 products for observation 29 are unchanged.
 `diff -s -I 'asn_pool' -I'constraints' output output_taconfirm/ | grep o029

1
{noformat}
81:Files output/jw00623-o029_image2_001_asn.json and output_taconfirm/jw00623-o029_image2_001_asn.json are identical
 ...
 192:Files output/jw00623-o029_image3_001_asn.json and output_taconfirm/jw00623-o029_image3_001_asn.json are identical`{noformat}
I've attached the updated association files.

stscijgbot:
Comment by Howard Bushouse: Fixed by [https://github.com/spacetelescope/jwst/pull/4149]

PR #4149, opened 2019-10-18
JP 1055 Added mir_taconfirm to the target acquisition list

ddavis-stsci:


ddavis-stsci:
Updated the ASN rules so that MIR_TACONFIRM is treated as a target acquisition. Since we don't have an example pool file for this I manually modified jw00623_20190625t115752_pool.csv so that one entry has the EXP_TYPE changed from MIR_TACQ to MIR_TACONFIRM. 
`jw00623029003_02101_00001_mirimage_uncal.fits|V00623029003P0000000002101|623|29|3|00623029003|2|PRIME_TARGETED_FIXED|1|01|1|MIR_TACONFIRM`

The diff command shows the level 2 & 3 products for observation 29 are unchanged. 
`diff -s -I 'asn_pool' -I'constraints' output output_taconfirm/ | grep o029

181:Files output/jw00623-o029_image2_001_asn.json and output_taconfirm/jw00623-o029_image2_001_asn.json are identical
...
192:Files output/jw00623-o029_image3_001_asn.json and output_taconfirm/jw00623-o029_image3_001_asn.json are identical`

I've attached the updated association files.

ddavis-stsci:
Closes JP-1055
closes #4135

ddavis-stsci:
Well github will not let me attach the .json files so the are now in the ticket (JP-1055).

PR #4168, opened 2019-10-23
JP-1063: Initial version of imaging APCORR ref file models

### LINKED ISSUE #4145: Create initial versions of imaging APCORR data models
Issue [JP-1063](https://jira.stsci.edu/browse/JP-1063) was created by Howard Bushouse:

JP-1022 gives the specifications for new data models that will be used for imaging-mode aperture correction (apcorr) reference files, as determined by the instrument teams. We need to create those data models, so that the teams can start using them to create initial versions of the reference files.

hbushouse:
New data models for the imaging mode aperture correction reference files, according to the specs provided in JP-1022.

Fixes #4145

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -119,6 +119,10 @@ datamodels
 
 - Added new spectroscopic mode photom reference file data models. [#4096]
 
+- Added new imaging mode aperture correction (apcorr) reference file data
+  models ``FgsImgApcorrModel``, ``MirImgApcorrModel``, ``NrcImgApcorrModel``,
+  and ``NisImgApcorrModel``. [#4168]
+
 exp_to_source
 -------------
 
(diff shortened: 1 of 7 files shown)


jdavies-st:
Given the proliferation of instrument-specific datamodels with obscure names, could a different possible approach be to have one `ApcorrModel` with multiple possible tables, one for each instrument/mode supported?  So there would be a `miri_img_apcorr_table` and a `fgs_img_apcorr_table`, etc.

Recall that a single schema can define multiple tables, but they don't need to be populated in any given instantiated model.  So the MIRI team could populate `miri_img_apcorr_table` and so on.  Any written out table with one of these tables populated would validate correctly against the schema.

Of course this would require that each table be read in a specific way in the code, but that is essentially already the case, as each table has different columns.  Adding a different attribute name for the table name would be relatively small change.

Just a thought.

Otherwise the PR looks fine.  =)

sosey:
just for added fodder for james comment on datamodels and schemas, we do this in other areas already, like the schemas for wavelengthrange, that cover multiple instruments and modes with one schema: https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/schemas/wavelengthrange.schema.yaml

PR #4173, opened 2019-10-23
JP-1053: Remove old photom data models

hbushouse:
One more component of the work for JP-1053, which is to remove the definitions of the old photom reference file data models. Also some additional updates to the photom docs, as well as adding the new apcorr data models (#4168). Theoretically this shouldn't break anything, because the updated photom code only uses the new data models.


jwst/datamodels/photom.py
```diff
@@ -74,69 +59,58 @@ class NrcWfssPhotomModel(ReferenceFileModel):
...
-        - pupil: str[12]
-        - order: int16
-        - photmjsr: float32
-        - uncertainty: float32
-        - nelem: int16
-        - wavelength: float32[5000]
-        - relresponse: float32[5000]
+    data : numpy array
+        An array-like object containing the pixel-by-pixel conversion values
+        in units of DN / sec / mJy / pixel.
```
philhodge: This needs parentheses.  Also, shouldn't it be (MJy / pixel) / (DN / s)?

hbushouse: Yes, I missed this one. Good eye!

hbushouse:
We've had a couple of clean regression test runs using the 0.14.0 release, so I'm going merge this and see what fallout occurs.

PR #4174, opened 2019-10-23
JP-1053 Fix bugs in photom

philhodge:
Function save_area_nirspec failed for NIRSpec IFU data because it was trying to assign a value to slit.meta.photometry.pixelarea_arcsecsq, but when running photom the input is a single 2048x2048 array, i.e. there are no slits.

Function photom_io failed for NIRISS extended-source data.  The conversion factor should be divided by the pixel area, but that value was None.

### DIFF
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -665,17 +665,25 @@ class DataSet():
                 else:
                     srctype = self.input.meta.target.source_type
                 if srctype is None or srctype.upper() != 'POINT':
-                    log.debug('Converting conversion factor from flux '
-                              'to surface brightness')
-                    conversion /= slit.meta.photometry.pixelarea_steradians
+                    if slit.meta.photometry.pixelarea_steradians is None:
+                        log.warning("Pixel area is None, so can't convert "
+                                    "flux to surface brightness!")
+                    else:
+                        log.debug("Converting conversion factor from flux "
+                                  "to surface brightness")
+                        conversion /= slit.meta.photometry.pixelarea_steradians
                 else:
                     unit_is_surface_brightness = False
             else:
                 srctype = self.input.meta.target.source_type
                 if srctype is None or srctype.upper() != 'POINT':
-                    log.debug('Converting conversion factor from flux '
-                              'to surface brightness')
-                    conversion /= self.input.meta.photometry.pixelarea_steradians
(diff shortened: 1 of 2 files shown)


jwst/photom/photom.py
```diff
@@ -665,17 +665,25 @@ def photom_io(self, tabdata, order=None):
                 else:
                     srctype = self.input.meta.target.source_type
                 if srctype is None or srctype.upper() != 'POINT':
-                    log.debug('Converting conversion factor from flux '
-                              'to surface brightness')
-                    conversion /= slit.meta.photometry.pixelarea_steradians
+                    if slit.meta.photometry.pixelarea_steradians is None:
```
hbushouse: For the case that failed in the regression tests, was the problem due to the pixel area missing from the reference file? Or was the loading of it missed somewhere else in the code?

jwst/photom/photom.py
```diff
@@ -912,9 +924,9 @@ def save_area_nirspec(self, pix_area):
             # There is a slice_id column for selecting a matching slice, but
             # we're just going to average the pixel area for all slices.
             pixel_area = np.nanmean(pixarea)
-            slit.meta.photometry.pixelarea_arcsecsq = pixel_area
-            slit.meta.photometry.pixelarea_steradians = \
-                slit.meta.photometry.pixelarea_arcsecsq * A2_TO_SR
+            self.input.meta.photometry.pixelarea_arcsecsq = pixel_area
```
hbushouse: Is the "input" model being updated in place throughout the step or is this the variable name chosen for the output model?

stscirij: Looks like self.input is a copy of the input datamodel, so isn't updated in place

jwst/photom/photom.py
philhodge: In function `save_area_info`, no value was being assigned to the `pixel_steradians` attribute if the relevant value was not present in the area reference file.  I think that's appropriate, but I needed to check for that case when `pixel_steradians` was used.

jwst/photom/photom.py
philhodge: Yes, `self.input` is modified in-place.

philhodge:
@stscirij @hbushouse , yes, sorry; I didn't mean that the input file/model itself was modified.  `self.input` is just the name used for the open model that is modified and returned, and it would have been better if it were called `self.output`.

PR #4178, opened 2019-10-24
move hashes to current astropy, gwcs, asdf dev versions

nden:
We need to run reg tests on this PR before merging.


jdavies-st:
Sounds good.

PR #4187, opened 2019-10-25
Add set_telescope_pointing test for NRC_TSGRISM

hbushouse:
Based on the recent problems associated with JWSTDMS-14, I'm adding a regression test of the ``set_telescope_pointing`` script for a NRC_TSGRISM exposure. The test uses the static version of the ``prd.db`` that we got from DMS some time ago, which is known to have out of date values compared to the current SIAF table, but it at least works to check the mechanics of the operations. It also includes the use of the RESTfull interface to the EngDB, so full computation of pointing via quaternions is done.

The input file has been stripped of all keywords (both FITS keywords and ASDF meta data) that will be populated by the script, just like the intermediate level-1b files when DMS runs the script during level-1b processing. The truth file is the result of running the test manually and has all of the expected WCS-related keywords in the SCI extension header, including the all-important XREF_SCI and YREF_SCI keywords needed for level-2b processing of NRC_TSGRISM exposures.

The appropriate input and truth files have already been deployed to artifactory, under "nircam/test_pointing/".


hbushouse:
I also added a few new log messages to the ``set_telescope_pointing`` script, just to make it easier to see what it's actually doing when running.

jwst/tests_nightly/general/nircam/test_nircam_steps_single.py
```diff
@@ -186,3 +189,31 @@ def test_wfs_combine2(self):
...
+        # Copy original version of file to test file, which will get overwritten by test
+        input_file = self.get_data(self.test_dir,
+                                   'jw00721012001_03103_00001-seg001_nrcalong_uncal_orig.fits')
+
+        # Get SIAF PRD database file
+        siaf_prd_loc = ['jwst-pipeline', self.env, 'common', 'prd.db']
+        siaf_path = get_bigdata(*siaf_prd_loc)
+
+        add_wcs(input_file, siaf_path=siaf_path,
+                engdb_url='http://twjwdmsemwebag.stsci.edu/JWDMSEngFqAccSide1/TlmMnemonicDataSrv.svc/')
```
jdavies-st: Is there a reason this is different from our default URL currently in the code?

https://github.com/spacetelescope/jwst/blob/d448b27c9cc34f5ea70fb19181e890fd660f99b3/jwst/lib/engdb_tools.py#L22

hbushouse: This is the url I got from the most recent DMS Ops processing logs. When I try to use the default from engdb_tools, it fails to connect. The combinations that seem to work include:
ENGDB_HOST = 'http://twjwdmsemweb.stsci.edu/' or 'twjwdmsemwebag.stsci.edu/'
but with the second level path being 'JWDMSEngFqAccSide1'. Without the substring 'Side1' it fails.

hbushouse: Perhaps we should update the default in engdb_tools.

hbushouse: Latest commit updates the default URL and uses it (instead of listing it as an explicit argument).

jdavies-st:
Looks like a `flake8` failure is preventing the CI from passing.

hbushouse:
> Looks like a `flake8` failure is preventing the CI from passing.

It's because I have to import ENGDB_Service, but it doesn't actually get referenced anywhere. Any suggestions?

PR #4193, opened 2019-10-28
JP-913 Update documentation for save-parameters

### LINKED ISSUE #3889: strun --save-parameters without input should not error
Issue [JP-913](https://jira.stsci.edu/browse/JP-913) was created by Jonathan Eisenhamer:

The command

{{strun jwst.stpipe.Step --save-parameters step_params.asdf}}

should not error. If no input is provided, just save the class's defaults. Otherwise, save the instance's values

stscieisenhamer:
Resolves #3889/[JP-913](https://jira.stsci.edu/browse/JP-913).

### DIFF
--- a/docs/jwst/introduction.rst
+++ b/docs/jwst/introduction.rst
@@ -150,10 +150,10 @@ where ``my_dark_current.cfg`` contains:
 
 The configuration parameters for each step can be saved to an ASDF file using
 the ``--save-parameters <filename>`` option. For example, to save the
-configuration for the `tweakreg` step, one would do the following:
+configuration for the `persistence` step, one would do the following:
 ::
 
- strun jwst.tweakreg.PersistenceStep jw00017001001_01101_00001_nrca1_uncal.fits --save-parameters persistence.asdf
+ strun jwst.persistence.PersistenceStep jw00017001001_01101_00001_nrca1_uncal.fits --save-parameters persistence.asdf
 
 This file can be edited and then used to run the step with the new configuration:
 ::
@@ -451,16 +451,16 @@ local configuration file.
 
 A configuration file should be used when there are parameters a user wishes to
 change from the default/CRDS version for a custom run of the step. To create a
-configuration file add ``--save-parameters <filename.asdf`` to the command:
+configuration file add ``--save-parameters <filename.asdf>`` to the command:
 
  strun <step.class> <required-input-files> --save-parameters <filename.asdf>
 
-For example, to get the parameters for the jump step, use:
+For example, to save the parameters used for a run of the ``jump`` step, use:
 ::
 
  strun jwst.jump.JumpStep jw82500001003_02101_00001_NRCALONG_uncal.fits --save-parameters jump_pars.asdf
 
-Once retrieved, the file can be edited, removing parameters that should be left
+Once saved, the file can be edited, removing parameters that should be left
 at their default/CRDS values, and setting the remaining parameters to the
 desired values.
 
--- a/docs/jwst/introduction.rst
+++ b/docs/jwst/introduction.rst
@@ -153,7 +153,7 @@ the ``--save-parameters <filename>`` option. For example, to save the
 configuration for the `tweakreg` step, one would do the following:
 ::
 
- strun jwst.tweakreg.PersistenceStep --save-parameters persistence.asdf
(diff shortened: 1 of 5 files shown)


docs/jwst/introduction.rst
```diff
@@ -153,7 +153,7 @@ the ``--save-parameters <filename>`` option. For example, to save the
 configuration for the `tweakreg` step, one would do the following:
 ::
 
- strun jwst.tweakreg.PersistenceStep --save-parameters persistence.asdf
+ strun jwst.tweakreg.PersistenceStep jw00017001001_01101_00001_nrca1_uncal.fits --save-parameters persistence.asdf
```
hbushouse: Sorry, but ``tweakreg`` and ``persistence`` are mutually exclusive steps. Pick one or the other.

docs/jwst/introduction.rst
```diff
@@ -451,18 +451,23 @@ local configuration file.
 
 A configuration file should be used when there are parameters a user wishes to
 change from the default/CRDS version for a custom run of the step. To create a
-configuration file with the default values, use the command: ::
+configuration file add ``--save-parameters <filename.asdf`` to the command:
```
hbushouse: missing the closing ">" at the end of filename.asdf

docs/jwst/introduction.rst
```diff
@@ -451,18 +451,23 @@ local configuration file.
...
 
- strun <step.class> --save-parameters <filename.asdf>
+ strun <step.class> <required-input-files> --save-parameters <filename.asdf>
 
 For example, to get the parameters for the jump step, use:
```
hbushouse: I'm not sure I would characterize this example as the way to just "get the parameters for the jump step". You're doing more than that: you're running the step and having it save the parameters for you. If you just wanted to get the parameters, you could do something like ``stspec jwst.jump.JumpStep``.

stscieisenhamer: Modified the documentation to make it clear it was a run of the step.

Though stspec gives the {{spec}} of a step, it currently does not produce anything directly useful in creating parameter files. That may be something to modify such that it would do so.

docs/jwst/introduction.rst
```diff
@@ -110,55 +110,30 @@ step by using the '-h' (help) argument to strun:
...
-``calwebb_detector1.cfg`` file to contain the following: ::
-
- name = "Detector1Pipeline"
- class = "jwst.pipeline.Detector1Pipeline"
-
-    [steps]
-      [[dark_current]]
-        override_dark = 'my_dark.fits'
-
-Note that simply removing the entry for a step from a pipeline cfg file will
```
hbushouse: I can't immediately tell if this info (about removing the entry for a step from a cfg file) is replicated somewhere else? Or has it been completely removed? This has been an issue for a lot of folks in the past, in that they think by not having any info for a step in the cfg will cause it to be skipped.

stscieisenhamer: Removed in that some of this is replicated later on, but the specific bit about the note is gone. I will recover this.

stscieisenhamer:
@nden @jdavies-st So how is this `doctest_skip` supposed to work?

docs/jwst/stpipe/config_asdf.rst
```diff
@@ -75,3 +268,28 @@ The one keyword to explain further is ``reftype``. This is what CRDS uses to
...
+
+.. doctest_skip::
+
+   >>> import asdf
+   >>> cfg = asdf.open('config.asdf')
+       #
+       # Modify cfg['parameters'] as necessary
+       #
+   >>> cfg.add_history_entry('Parameters modified for some reason')
+   >>> cfg.write_to('config_modified.asdf')
```
jdavies-st: This block should be something like:
```
    import asdf
    cfg = asdf.open('config.asdf') # doctest: +SKIP
    #
    # Modify cfg['parameters'] as necessary
    #
    cfg.add_history_entry('Parameters modified for some reason') # doctest: +SKIP
    cfg.write_to('config_modified.asdf') # doctest: +SKIP
```
and no need for the `.. doctest_skip::` directive.

jdavies-st: The `# doctest: +SKIP` won't show up in the final HTML document.

Issue #4017, opened 2019-09-12
Bounding box missing for NIS SOSS non-TSO data


stscijgbot:
Issue [JP-985](https://jira.stsci.edu/browse/JP-985) was created by David Davis:

When processing jw00307 observation o002, NIS SOSS  and is a non-TSO data set
{noformat}
TSOVISIT= F / Time Series Observation visit indicator {noformat}
The assign_wcs step runs to completion in the spec2 pipeline,
{noformat}
2019-09-10 15:05:58,529 - stpipe.Spec2Pipeline.assign_wcs - INFO - COMPLETED assign_wcs{noformat}
but does not supply a bounding box which is required for resample_spec
{noformat}
resample_spec.py", line 99, in build_interpolated_output_wcs
 grid = wcstools.grid_from_bounding_box(bb)
 File "/dms/local/jwst/pipeline/pkgs/miniconda3/envs/jwstdp-0.13.8-py36/lib/python3.6/site-packages/gwcs/wcstools.py", line 186, in grid_from_bounding_box
 if np.isscalar(bounding_box[0]):
TypeError: 'NoneType' object is not subscriptable
{noformat}
This may be due to the wcs being applied without being given a spectral order in the previous step. The log files from the run in JWSTDMS-83 are attached.

 The files can be found on the D-string under

/ifs/archive/test/jwst/store/rspencer/tests/complete_medium_01/dtest2/JWSTDP-2019.2.0-44~aeda1a8b/2019-09-10-144245/sdp

stscijgbot:
Comment by Philip Hodge: For NIRISS SOSS data, resample_spec.py needs to be modified to call assign_wcs.niriss.niriss_soss_set_input().  This function takes the input model and the spectral order as input arguments, and it returns the wcs function for the specified spectral order.  This wcs function will have a bounding_box attribute, in contrast to input_model.meta.wcs.

stscijgbot:
Comment by Steven Crawford: It might be good to have [~bushouse] chime in on this, but it would be good to verify that NIRISS SOSS non-TSO mode data should be process by level 3 or if these should just be skipped as they are a non-standard mode.

stscijgbot:
[...]

Issue #4183, opened 2019-10-24
Inconsistent tables in the documentation


stscijgbot:
Issue [JP-1089](https://jira.stsci.edu/browse/JP-1089) was created by Alicia Canipe:

This table indicates that SOSS exposures can go through either calwebb_spec3 or calwebb_tso3, depending on whether or not TSO=True: [https://jwst-pipeline.readthedocs.io/en/latest/jwst/pipeline/main.html#pipelines-vs-exposure-type]

However, this table does not include SOSS for calwebb_spec3: [https://jwst-pipeline.readthedocs.io/en/latest/jwst/pipeline/calwebb_spec3.html]

According to NIRISS: If one has a non-TSO SOSS exposure it would be processed through calwebb_spec3. There may be such science cases if the SOSS mode works well, so there is this provision for the possibility of non-TSO requests for SOSS observations.

stscijgbot:
Comment by Howard Bushouse: JP-985 is tracking the work to be done to the calwebb_spec3 pipeline to allow for NIRISS SOSS non-TSO exposures to process without failure. We will update the docs as part of the code updates.

hbushouse:
Fixed in #4194

stscijgbot:
Comment by Howard Bushouse: Fixed by [https://github.com/spacetelescope/jwst/pull/4194]

PR #4194, opened 2019-10-28
JP-985: Update calwebb_spec3 for non-TSO NIRISS SOSS

hbushouse:
Updated the ``calwebb_spec3`` pipeline module to handle NIRISS SOSS (NIS_SOSS) the same it already handles NIRCam and NIRISS WFSS exposures, which is to simply run ``extract_1d`` and ``combine_1d`` to produce a combined x1d product from all input members (no resampling, no outlier detection, ...). NIRISS SOSS data should only ever go through the ``calwebb_spec3`` pipeline when they are NOT obtained in TSO mode (TSO mode SOSS data go through the ``calwebb_tso3`` pipeline).

Also added entries to the default ``calwebb_spec3.cfg`` file for the ``combine_1d`` step.


hbushouse:
Fixes #4017

hbushouse:
Documentation updates fix https://jira.stsci.edu/browse/JP-1089 and #4183

docs/jwst/pipeline/calwebb_spec3.rst
```diff
@@ -80,8 +95,8 @@ Source-based calibrated data
 
 For NIRSpec fixed-slit, NIRSpec MOS, and NIRCam and NIRISS WFSS, which have a defined
 set of slits or sources, the data from the input calibrated exposures is reorganized
-by :ref:`exp_to_source <exp_to_source>` so that all of the instances of data for a
-particular source/slit are contained in a
+by the :ref:`exp_to_source <exp_to_source>` step so that all of the instances of data
```
philhodge: Just a question:  Why is "_step" included within the angle brackets for other step names but not for this one?

hbushouse: Just randomness of how the cross-reference labels were created/spelled in the corresponding rst files for the steps.

PR #4199, opened 2019-10-29
Jp 937 Create image3 associations for FGS_IMAGE exposures

ddavis-stsci:
Updated the logic for Constraint_Optical_Path in the level 3 rules to allow filter to be undefined and retain the constraint that subarray be defined.

### DIFF
--- a/jwst/associations/lib/dms_base.py
+++ b/jwst/associations/lib/dms_base.py
@@ -89,6 +89,7 @@ IMAGE2_SCIENCE_EXP_TYPES = [
     'nrc_coron',
     'nrc_tsimage',
     'nrs_mimf',
+    'fgs_image',
 ]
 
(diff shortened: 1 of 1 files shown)


ddavis-stsci:
All association tests pass
= 133 passed, 2 xfailed, 1 warnings in 368.27s (0:06:08) =

Three of the regression tests fail and will need 10 image3 json files added for programs jw86600, jw97012, and jw97013.

Fixes JP-937

hbushouse:
I noticed that in the image3 ASN file attached to JP-937 the input members are rate products. The inputs for level-3 ASN's should be cal products.

ddavis-stsci:
I'll take another look. I missed that.

hbushouse:
Results look good now. I approve. 👍

PR #4202, opened 2019-10-29
Update README to avoid duplication of DMS snapshot installation procedure

rendinam:
& add python version to end-user conda installation step to avoid unintentional use of newer python.

Since the DMS snapshot installation procedure was modified slightly with the 0.14.0 release, the documentation here was rendered obsolete. To avoid having to maintain it in two locations, this documentation now points to the procedures associated with each snapshot release.

Since python 3.8.0 was recently published as a conda package, the conda create command used without pinning obtains that version of python rather than 3.7.4, which is the latest tested version. Add a variant of the `conda create` command that pins the target python version if the latest is not wanted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,14 +15,6 @@ Installation
 
 The ``jwst`` package can be installed into a virtualenv or conda environment via ``pip``.  We recommend creating a fresh environment with only python installed.  Via conda:
 
-    conda create -n jwst_env python
-    conda activate jwst_env
-
-*Note: The conda create command above will obtain the latest available python
-version which may newer than the version against which the jwst package has been
-most extensivesly tested. An explicit version pin may be used here to obtain a
-particular python version. i.e*
-
     conda create -n jwst_env python=3.7.4
     conda activate jwst_env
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,6 +125,7 @@ Software vs DMS build version map
 
 | jwst tag | DMS build | CRDS_CONTEXT |   Date     |          Notes                           |
 | -------- | --------- | ------------ | ---------- | -----------------------------------------|
+|  0.14.0  | B7.4rc1   | 0560         | 10/25/2019 | First release candidate for B7.4         |
```
hbushouse: 👍

README.md
```diff
@@ -35,19 +43,21 @@ As can a particular commit hash:
...
 
-where `<URL>` is of the form:
+    https://github.com/astroconda/astroconda-releases/tree/master/jwstdp
 
-    Linux: http://ssb.stsci.edu/releases/jwstdp/0.13.7/latest-linux
-    OS X: http://ssb.stsci.edu/releases/jwstdp/0.13.7/latest-osx
+The version values shown there are the JWSTDP releases available to install.
+Installation procedures for each version are in the README.md file for that
+version. See the "Software vs DMS build version map" table below for a
+list of version tags corresponding to particular releases.
```
stscicrawford: I like the additional text that explains some more detail about the DMS release.  

I think though it is preferable to keep this as the one place for installation instructions, and this should be kept up to date.  Please add the instructions for the current latest release here as well.

README.md
```diff
@@ -18,6 +18,14 @@ The ``jwst`` package can be installed into a virtualenv or conda environment via
...
     conda activate jwst_env
 
+*Note: The conda create command above will obtain the latest available python
+version which may newer than the version against which the jwst package has been
+most extensivesly tested. An explicit version pin may be used here to obtain a
+particular python version. i.e*
+
+    conda create -n jwst_env python=3.7.4
+    conda activate jwst_env
+
```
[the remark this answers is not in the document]
philhodge: Typos in the quoted text:
"version which may newer than ..."
"most extensivesly ..."

rendinam:
Not entirely sure why the coverage claims to have decreased as a result of these commits, but this should be ready to merge.

jdavies-st:
Yeah project coverage is not currently reliable as it keys off regression test coverage as well.  Merging.

PR #4205, opened 2019-10-31
Remove/replace SUBPRISM in unit tests

hbushouse:
MIRI LRS slitless exposures are now coming out of DMS with SUBARRAY='SLITLESSPRISM' instead of the old 'SUBPRISM' value. Most MIRI reference files now use SLITLESSPRISM too. So this excises the use of 'SUBPRISM' from our code base.

### DIFF
--- a/jwst/photom/tests/test_photom.py
+++ b/jwst/photom/tests/test_photom.py
@@ -290,7 +290,7 @@ def create_input(instrument, detector, exptype,
             input_model.var_poisson = np.ones(shape, dtype=np.float32)
             input_model.var_rnoise = np.ones(shape, dtype=np.float32)
             input_model.var_flat = np.ones(shape, dtype=np.float32)
-            input_model.meta.subarray.name = 'SUBPRISM'     # matches 'GENERIC'
+            input_model.meta.subarray.name = 'FULL'
             input_model.meta.target.source_type = 'POINT'
             input_model.meta.photometry.pixelarea_arcsecsq = 0.0025
             input_model.meta.photometry.pixelarea_steradians = 0.0025 * A2_TO_SR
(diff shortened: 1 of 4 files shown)


jwst/photom/tests/test_photom.py
```diff
@@ -290,7 +290,7 @@ def create_input(instrument, detector, exptype,
             input_model.var_poisson = np.ones(shape, dtype=np.float32)
             input_model.var_rnoise = np.ones(shape, dtype=np.float32)
             input_model.var_flat = np.ones(shape, dtype=np.float32)
-            input_model.meta.subarray.name = 'SUBPRISM'     # matches 'GENERIC'
+            input_model.meta.subarray.name = 'FULL'
```
jdavies-st: Is this supposed to be changed to `SLITLESSPRISM`?

jwst/rscd/rscd_sub.py
```diff
@@ -196,7 +196,10 @@ def get_rscd_parameters(input_model, rscd_model):
     readpatt = input_model.meta.exposure.readpatt
     subarray = input_model.meta.subarray.name
 
-#    if subarray == 'SUBPRISM': subarray = 'SLITLESSPRISM'
+    # Check for old values of the MIRI LRS slitless subarray name
+    # in the science data and change to the new
+    if subarray.upper() == 'SUBPRISM': subarray = 'SLITLESSPRISM'
```
jdavies-st: nit: This would be more readable if it was 2 lines.

jwst/photom/tests/test_photom.py
hbushouse: No. The actual test case is simulating an LRS fixedslit exposure, which are always taken using SUBARRAY='FULL' (SUBPRISM is only used for slitless). So I don't know why the original was set that way. Apparently it doesn't make any difference to the actual test operations. But I reset the value to 'FULL' just to be consistent with the real params of a LRS fixedslit exposure.

philhodge: I don't think there was any reason why I used SUBPRISM.  It didn't matter, and I just wanted to test the row selection.

PR #4232, opened 2019-11-06
Jp 1120 list MIR_TACQ as target_acquisition in level 3 associations

ddavis-stsci:
Update for level 3 rules so that the target acquisitions in the associations are listed as exp_type = 'target_acquisition' and not as science exposures. 

The local tests pass pytest  ~/src/JWST/scsb/jwst/jwst/associations/tests,
`133 passed, 2 xfailed, 1 warnings in 374.61s (0:06:14) `
and in the regression tests five spec3 associations from jw00623 will need to be updated.


jwst/associations/lib/rules_level3_base.py
```diff
@@ -956,8 +956,17 @@ def get_exposure_type(self, item, default='science'):
...
         -------
         exposure_type : 'science'
-            Always returns as science
+            Returns as science for most Exposures
+        exposure_type : 'target_acquisition'
+            Returns target_acquisition for mir_tacq
         """
+        try:
+            exp_type = item['exp_type']
+            if exp_type =='mir_tacq':
```
hbushouse: A new MIRI exp_type that's in the process of being implemented is 'mir_taconfirm', so we should go ahead and include that value here too, in order to make this future-proof. We already allow for it in our data model schemas.

hbushouse: Another general question: all that's being checked for here is the specific 'mir_tacq' value. Target acq's from all instruments should be treated the same way, so is there some other rule(s) somewhere that handles those (e.g. nrc_tacq, nis_tacq, nrs_wata, ...)

ddavis-stsci: The code now checks in any of the exp_types are in the ACQ_EXP_TYPES and if so lists them in the association as target_acquisition. This list includes,     
    'mir_tacq',
    'mir_taconfirm',
    'nis_taconfirm',
    'nis_tacq',
    'nrc_taconfirm',
    'nrc_tacq',
    'nrs_confirm',
    'nrs_msata',
    'nrs_taconfirm',
    'nrs_tacq',
    'nrs_taslit',
    'nrs_wata',

PR #4244, opened 2019-11-08
Changed method to open pixel area map from datamodels.open to datamodels.PixelAreaModel

### LINKED ISSUE #4243: Pixel area map reference file not loaded properly
In 0be3208811c75d80fbcae32f4655ebd095a5a418 `photom.py` was updated to load the pixel area map using a generic `datamodels.open` instead of `datamodels.PixelAreaModel` (see [here](https://github.com/spacetelescope/jwst/commit/0be3208811c75d80fbcae32f4655ebd095a5a418#diff-73c7d1f525bfa73f37c12f62510511f4L752)). However, I'm having issues loading the pixel area map using the new method, receiving the following warning during Image2:
```
stpipe.Image2Pipeline.photom - WARNING - The PIXAR_SR keyword is missing from /grp/crds/test_cache/references/jwst/jwst_niriss_area_0011.fits
stpipe.Image2Pipeline.photom - WARNING - The PIXAR_A2 keyword is missing from /grp/crds/test_cache/references/jwst/jwst_niriss_area_0011.fits
stpipe.Image2Pipeline.photom - DEBUG - PIXAR_SR = None, PIXAR_A2 = None
```
The PIXAR_SR and PIXAR_A2 keywords are in both the 0th and SCI extensions of the referenced pixel area map. 

As a result, there is no `meta.photometry` info in the opened datamodel because it opens the pixel area map as a generic ReferenceImageModel. Consequently, in Image3.source_catalog objects are assigned NaN ABmag, which causes the pipeline to crash in Spec2.extract_2d because there are no valid targets:
```
stpipe.Spec2Pipeline.extract_2d - INFO - Grism object list created from source catalog:
```

The full traceback is attached if needed: [pipeline.log](https://github.com/spacetelescope/jwst/files/3826304/pipeline.log)

Luckily it's an easy fix for my purposes (testing the new photom files)- the pixel area map just needs to be opened using the explicit datamodel. 

Also, seems like this issue will become obsolete when issue #3837 is addressed.

jotaylor:
Changed method to open pixel area map from datamodels.open to datamodels.PixelAreaModel.

Resolves #4243

### DIFF
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -823,7 +823,7 @@ class DataSet():
         instrument = self.input.meta.instrument.name.upper()
 
         # Load the pixel area reference file
-        pix_area = datamodels.open(area_fname)
+        pix_area = datamodels.PixelAreaModel(area_fname)
 
         # Copy the pixel area data array to the appropriate attribute
         # of the science data model


hbushouse:
I think it's debatable as to which approach is better: using the generic ``datamodels.open()`` and relying on the DATAMODL keyword value vs. explicitly loading as ``datamodels.PixelAreaModel()``. It's true that all reference files (and science products for that matter) should have a DATAMODL keyword, but what's really important is that the format and content of the data in the file follow the expected definitions of the data model. Hence by explicitly opening with ``datamodels.PixelAreaModel`` the schema validation that takes place will instantly alert you when someone tries to use a file that doesn't conform to what we're expecting for an AREA reference file.

The same approach is taken in some of the pipeline modules. For example, because the ``calwebb_detector1`` pipeline will only work if you the input is a 4-D dataset conforming to a RampModel, the input is explicitly loaded with ``datamodels.RampModel()``. So if someone tries to feed it a countrate or calibrated file, it will throw an error immediately upon loading, rather than hitting some more obscure error later on when it tries to access non-existent dimensions of the input data.

philhodge:
There are four different pixel area reference file models.  There are a lot more photom reference models.  The photom step had been opening most of these files by explicitly specifying the model, and I changed them all to `datamodels.open` when all of the photom models changed.  I didn't think that it made sense for photom to have to check the instrument and mode just to open a reference file.  There's a lot of mode-specific code for applying the data model, but some is generic.

hbushouse:
Updated ref files that include the DATAMODL keyword will be delivered to CRDS, rendering this code change unnecessary. Closing.

PR #4256, opened 2019-11-12
Fix for JP-1008: White Light Step does not ignore NaNs in input.

dmggh:
NaNs are  now treated as 0 in the flux sum for each integration in the input.


dmggh:
Removed trailing white space to placate flake8.

PR #4265, opened 2019-11-14
Add Codecov flags unittests, nightly, newnightly

### LINKED ISSUE #4264: Update codecov to use flags indicating unittest vs regtest
Issue [JP-1141](https://jira.stsci.edu/browse/JP-1141) was created by James Davies:

Currently our codecov setup compares nightly regression test coverage to per-PR CI coverage and this can differ by 15% or more for the _project_ or even _patch_ checks.

So it would be good to compare apples to apples by using Codecov flags to indicate the type of test suite being run by any particular CI or nightly test runs, and only compare like to like.

This will also be useful in comparing the new regression test system to the old.

jdavies-st:
Sets up Codecov flags

- `unittests`
- `nightly`
- `newnightly`

So that our daily coverage testing is being tracked in parallel, and that there are apples-to-apples comparisons.

And only do coverage on the TravisCI builds and the nightly Jenkins RT builds.

Resolves #4264

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -15,5 +15,6 @@ omit =
     docs/*
     jwst/conftest.py
     jwst/tests*
-    jwst/*/tests/*
+    jwst/regtest
+    jwst/**/tests
     jwst/extern/*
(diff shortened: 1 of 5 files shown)


jdavies-st:
So it looks like we do not need to specify `target` coverage thresholds in our `codecov.yml`, as we would rather have increasing coverage be the threshold of a passing coverage check.

Screen Shot 2019-11-14 at 2 25 51 PM

PR #4268, opened 2019-11-14
Consolidate property discovery into postmortem function

jhunkeler:
Question:

Would it be better if `want_property` accepted a list and `yield`s each property as they're discovered?

```python
for key, value in postmortem(request, ['output', 'other', 'data']):
```

 Or should `want_property` accept a string and only return the value of the property?

```python
prop = postmortom(request, 'output')
if prop:
    thing = prop
    stuff.append(prop)
```

### DIFF
--- a/jwst/regtest/conftest.py
+++ b/jwst/regtest/conftest.py
@@ -43,12 +43,18 @@ def pytest_runtest_makereport(item, call):
 
 
 def postmortem(request, want_property):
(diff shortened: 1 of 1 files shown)


jdavies-st:
It probably doesn't matter too much, as for almost every use case, it is grabbing a single property, as there should only be one output per test.

PR #4270, opened 2019-11-15
Get the pixel area from the photom reference file

philhodge:
For all instruments other than NIRSpec, the pixel area (solid angle) is not gotten from the photom reference file instead of from the area reference file.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,8 @@ photom
 - Fixed a bug so that the reference table column "PHOTMJ" is used for NIRSpec IFU
   exposures. [#4263]
 
+- The pixel area is now gotten from the photom reference file. [#4270]
+
 0.14.1 (2019-11-11)
 ===================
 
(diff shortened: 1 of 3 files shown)


philhodge:
The IFU cube det_image_ch1-short_s3d.fits and also det_image_ch1-short_s3d.fits were made using version 0.14.0 of JWST.  A number of significant changes have been made to the pipeline since then, including changes to the units of an IFU cube (now MJy / sr).  You should do the test again using JWST 0.14.3, i.e. the latest develop version.  I would suggest starting with the raw file and running calwebb_detector1.py and calwebb_spec2.py.

hbushouse:
Where did these comments come from? I don’t see them anywhere within #4270. Is this something really old that github just finally decided to pass along?

philhodge:
Sorry, that was a dumb mistake!  I had a Jira ticket open in one tab, and I was looking at an old GitHub PR (this one) in another tab.  I intended to add a comment to the Jira ticket, but I added it to this PR instead.  When I realized my mistake, I deleted the comment from this PR (that's why you don't see it here), and I added it to the Jira ticket.

PR #4273, opened 2019-11-15
update asdf's hash

nden:
to the latest master


jdavies-st:
We need to update the version pin of `semantic-version` in `requirements-sdp.txt`.

jdavies-st:
Updated `semantic-version` pin in latest commit on this branch.

PR #4288, opened 2019-11-20
Adding white light unit tests

mfixstsci:
Ready for review.

### DIFF
--- a/jwst/white_light/tests/test_white_light.py
+++ b/jwst/white_light/tests/test_white_light.py
@@ -1,0 +1,110 @@
+from jwst import datamodels
+from jwst.white_light.white_light import white_light
+
+from astropy.time import Time, TimeDelta
+import math
+import numpy as np
+import pytest
+
+@pytest.fixture(scope='module')
+def make_datamodel():
+    """Make data for white light tests"""
+
+    model = datamodels.MultiSpecModel()
+    model.meta.exposure.group_time = 0.15904
+    model.meta.exposure.ngroups = 60
+    model.meta.exposure.start_time = 58627.0
+    model.meta.exposure.integration_start = 1
+    model.meta.exposure.integration_end = 2
+
+    # Make data arrays
+    flux = np.random.rand(200)*1e-9
+    wavelength = np.arange(11,13, step=0.01)
+    error = np.random.rand(200)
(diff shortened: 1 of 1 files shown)


jwst/white_light/tests/test_white_light.py
```diff
@@ -0,0 +1,110 @@
...
+    offset = int_start - int_num[0]
+    time_arr = np.zeros(ntables, dtype=np.float64)
+    time_arr[0 : 1] = mid_utc[offset : offset + ntables]
+    int_times = Time(time_arr, format='mjd', scale='utc')
+
+    # Sum the fluxes
+    fluxsums = data.spec[0].spec_table['FLUX'].sum()
+
+    assert result['MJD'] == int_times.mjd
+    assert result['whitelight_flux'] ==  fluxsums
```
dmggh: Is fluxsum a sum of random values from the flux array ? If so, is it guaranteed that you can test for equality here instead of using allclose() ?

jwst/white_light/tests/test_white_light.py
```diff
@@ -0,0 +1,110 @@
...
+    model.meta.exposure.integration_start = 1
+    model.meta.exposure.integration_end = 2
+
+    # Make data arrays
+    flux = np.random.rand(200)*1e-9
+    wavelength = np.arange(11,13, step=0.01)
+    error = np.random.rand(200)
+    surf_bright = np.zeros(200)
+    sb_error = np.zeros(200)
+    dq = np.ones(200)
```
dmggh: Doesn't this set the dq to 'bad' ? I don't see them reset, so perhaps this doesn't matter if the dq values are never used.

mfixstsci: @dmggh -- Correct, the step doesn't actually uses the DQ flags but the `spec_table` attribute of the `SpectrumModel` is expecting all of the columns to be populated.

jwst/white_light/tests/test_white_light.py
mfixstsci: Correct, this is what happens when I carelessly copy and paste. I will make the change! Thanks.

mfixstsci:
Went ahead and made the changes @dmggh

PR #4295, opened 2019-11-21
Be explicit in .travis.yml about Python versions

jdavies-st:
If we set the TravisCI versions like
```yaml
   include:
    - python: 3.6
    - python: 3.7
```
Then we get maybe old, maybe very old versions:

- python 3.6.7 https://travis-ci.org/spacetelescope/jwst/jobs/615155895#L567
- python 3.7.1 https://travis-ci.org/spacetelescope/jwst/jobs/615155896#L567

So let's be explicit that we want all the latest bugfix releases as well.
```yaml
   include:
    - python: 3.6.9
    - python: 3.7.5
```

Fixup from changes made in #3957

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,12 +27,12 @@ matrix:
   fast_finish: true
 
   include:
-    - python: 3.6
-    - python: 3.7
-    - python: 3.8
+    - python: 3.6.9
+    - python: 3.7.5
+    - python: 3.8.0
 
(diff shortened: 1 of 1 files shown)


jdavies-st:
This resolves the issue.  The build matrix now reports:

Screen Shot 2019-11-22 at 10 51 36 AM

And the actual builds show the latest version in the series:

Screen Shot 2019-11-22 at 10 51 53 AM

Issue #4261, opened 2019-11-14
Tweakreg failure when no sources are found


stscijgbot:
Issue [JP-1140](https://jira.stsci.edu/browse/JP-1140) was created by Howard Bushouse:

DMS testing of CAL 0.14.1 has revealed a failure in the `calwebb_image3` `tweakreg` step that occurs when no viable sources were found from the source catalogs. The old version of `tweakreg` had been made robust enough to handle this appropriately, by just skipping the remaining processing and returning the results unchanged. It appears that the recent switch to using the new `tweakwcs` library has lost some of this robustness.

An example traceback is:
{code:java}
2019-11-13 19:50:18,220 - stpipe.Image3Pipeline.tweakreg - WARNING - /dms/local/jwst/pipeline/pkgs/miniconda3/envs/jwstdp-0.14.1-py37/lib/python3.7/site-packages/photutils/detection/findstars.py:972: NoDetectionsWarning: Sources were found, but none pass the sharpness and roundness criteria.
  'and roundness criteria.', NoDetectionsWarning)

2019-11-13 19:50:18,221 - stpipe.Image3Pipeline.tweakreg - WARNING - No sources found in jw87700005001_02103_00001_nis_cal.fits.
2019-11-13 19:50:18,224 - stpipe.Image3Pipeline.tweakreg - INFO -
2019-11-13 19:50:18,224 - stpipe.Image3Pipeline.tweakreg - INFO - Number of image groups to be aligned: 4.
2019-11-13 19:50:18,224 - stpipe.Image3Pipeline.tweakreg - INFO - Image groups:
2019-11-13 19:50:18,249 - stpipe.Image3Pipeline.tweakreg - WARNING - /dms/local/jwst/pipeline/pkgs/miniconda3/envs/jwstdp-0.14.1-py37/lib/python3.7/site-packages/astropy/modeling/core.py:734: AstropyDeprecationWarning: Class TPCorr defines class attributes ``inputs``.
            This has been deprecated in v4.0 and support will be removed in v4.1.
            Starting with v4.0 classes must define a class attribute ``n_inputs``.
            Please consult the documentation for details.

  """, AstropyDeprecationWarning)

2019-11-13 19:50:18,250 - stpipe.Image3Pipeline.tweakreg - INFO - * Images in GROUP 'jw87700005001_02104_00001_nis_cal':
[...]

PR #4299, opened 2019-11-21
Improve tweakreg resilience to empty catalogs

mcara:
This PR fixes https://github.com/spacetelescope/jwst/issues/4261

### DIFF
--- a/jwst/tweakreg/tweakreg_step.py
+++ b/jwst/tweakreg/tweakreg_step.py
@@ -195,7 +195,7 @@ class TweakRegStep(Step):
                 self.skip = True
                 for model in images:
                     model.meta.cal_step.tweakreg = "SKIPPED"
-                return input
+                return images
 
             else:
                 raise e
--- a/jwst/tweakreg/tweakreg_step.py
+++ b/jwst/tweakreg/tweakreg_step.py
@@ -136,9 +136,9 @@ class TweakRegStep(Step):
             self.log.info('')
 
             # we need at least two exposures to perform image alignment
-            self.log.info("At least two exposures are required for image "
-                          "alignment.")
-            self.log.info("Nothing to do. Skipping 'TweakRegStep'...")
+            self.log.warning("At least two exposures are required for image "
+                             "alignment.")
+            self.log.warning("Nothing to do. Skipping 'TweakRegStep'...")
             self.skip = True
             for model in images:
                 model.meta.cal_step.tweakreg = "SKIPPED"
@@ -170,17 +170,35 @@ class TweakRegStep(Step):
             yoffset=self.yoffset
         )
 
-        align_wcs(
-            imcats,
-            refcat=None,
-            enforce_user_order=self.enforce_user_order,
-            expand_refcat=self.expand_refcat,
-            minobj=self.minobj,
-            match=tpmatch,
-            fitgeom=self.fitgeometry,
-            nclip=self.nclip,
-            sigma=(self.sigma, 'rmse')
-        )
+        try:
+            align_wcs(
+                imcats,
+                refcat=None,
+                enforce_user_order=self.enforce_user_order,
+                expand_refcat=self.expand_refcat,
+                minobj=self.minobj,
+                match=tpmatch,
+                fitgeom=self.fitgeometry,
+                nclip=self.nclip,
+                sigma=(self.sigma, 'rmse')
+            )
+
+        except ValueError as e:
+            msg = e.args[0]
+            if (msg == "Too few input images (or groups of images) with "
+                "non-empty catalogs."):
+                # we need at least two exposures to perform image alignment
+                self.log.warning(msg)
+                self.log.warning("At least two exposures are required for "
+                                 "image alignment.")
+                self.log.warning("Nothing to do. Skipping 'TweakRegStep'...")
+                self.skip = True
+                for model in images:
+                    model.meta.cal_step.tweakreg = "SKIPPED"
+                return input
+
+            else:
+                raise e
 
         for imcat in imcats:
             imcat.meta['image_model'].meta.cal_step.tweakreg = 'COMPLETE'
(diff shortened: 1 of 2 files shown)


jwst/tweakreg/tweakreg_step.py
```diff
@@ -170,17 +170,35 @@ def process(self, input):
...
+                "non-empty catalogs."):
+                # we need at least two exposures to perform image alignment
+                self.log.warning(msg)
+                self.log.warning("At least two exposures are required for "
+                                 "image alignment.")
+                self.log.warning("Nothing to do. Skipping 'TweakRegStep'...")
+                self.skip = True
+                for model in images:
+                    model.meta.cal_step.tweakreg = "SKIPPED"
+                return input
```
jdavies-st: This should probably return `images`, right?

mcara: No, `input` is perfect.

mcara: Yes, you are right.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -170,17 +170,35 @@ def process(self, input):
...
+                sigma=(self.sigma, 'rmse')
+            )
+
+        except ValueError as e:
+            msg = e.args[0]
+            if (msg == "Too few input images (or groups of images) with "
+                "non-empty catalogs."):
+                # we need at least two exposures to perform image alignment
+                self.log.warning(msg)
+                self.log.warning("At least two exposures are required for "
```
hbushouse: Too late now, but in hind sight I think this warning message could've been a bit more explicit. The problem condition that's being encountered is not the fact that there aren't enough images, it's the fact that there aren't enough images **WITH DETECTED SOURCES**. I think we'll get confused users saying "I have several input images, so why is it telling me I need at least two?".

jdavies-st: Is the message given above, "Too few input images (or groups of images) with non-empty catalogs" not clear enough?  Or is the term "detected sources" preferable?

hbushouse: If the message referred to in lines 188-189 is also seen by the user, then I think we're OK as-is. I had initially assumed that that message was being trapped and then only the one in lines 192-193 was being written out for the user.

hbushouse:
Just tested this fix on the example data in /grp/jwst/ssb/chartreuse/dms_bugs/jwstdms-245/ and it still fails. The new try/except block around the call to align_wcs() is correctly branching to the except part of the block, but it's looking for the wrong error message. The error message being returned is "Image catalog must contain at least one entry.". Because it's not finding this error message, the remaining part of the except block is not executed and the step still fails with a raised exception.

hbushouse:
When I modify the code to look for the appropriate error message, it then correctly skips the rest of the ``tweakreg`` step, but the only warnings sent to the user are:
```
2019-11-25 11:13:56,656 - stpipe.Image3Pipeline.tweakreg - WARNING - Image catalog must contain at least one entry.
2019-11-25 11:13:56,656 - stpipe.Image3Pipeline.tweakreg - WARNING - At least two exposures are required for image alignment.
2019-11-25 11:13:56,656 - stpipe.Image3Pipeline.tweakreg - WARNING - Nothing to do. Skipping 'TweakRegStep'...
```
So nowhere does it say anything about not having any sources in any of the images.

philhodge:
Testing on an error message seems particularly fragile to me.  In order to check for a particular error, if you don't want to bother setting up different classes, I would suggest assigning integer codes and testing for the code.

mcara:
> if you don't want to bother setting up different classes

I do not see the difference between testing on an integer or a bunch of characters (`int8`). Same thing to me. I thought about introducing a new exception class but it really does not make much sense from the point of view of the stand-alone `tweakwcs`.

jdavies-st:
Yes, agree that excepting on an error message is fragile.  But don't introduce integer error codes either.   The way to to this in python is to use an Exception class.  And if `tweakwcs` gives errors of the same class that needed differentiation by the wrapping code in `jwst.tweakreg`, then use a custom error class.

```python
try:
    stuff
except (ValueError, MyException):
    other stuff
```

And `MyException` should probably subclass `ValueError` instead of `Exception`.  Though maybe this is overkill, and `ValueError` is enough without the `if` block.

mcara:
@hbushouse 
>  The error message being returned is "Image catalog must contain at least one entry.". Because it's not finding this error message, the remaining part of the except block is not executed and the step still fails with a raised exception.

This is a very different message from what @jdavies-st and I were getting. Please make sure you are using correct environment with an updated `tweakwcs`. Something is not right here.

hbushouse:
You're absolutely right! I hadn't updated the rest of my environment in a while and was obviously running an older version of `tweakwcs`. After creating a new environment I now get:
```
2019-11-25 13:24:44,694 - stpipe.Image3Pipeline.tweakreg - WARNING - Group with ID 'jw87700005001_02101_00001_nis_cal' will not be aligned: empty source catalog
2019-11-25 13:24:44,698 - stpipe.Image3Pipeline.tweakreg - WARNING - Group with ID 'jw87700005001_02103_00001_nis_cal' will not be aligned: empty source catalog
2019-11-25 13:24:44,701 - stpipe.Image3Pipeline.tweakreg - WARNING - Group with ID 'jw87700005001_02104_00001_nis_cal' will not be aligned: empty source catalog
2019-11-25 13:24:44,708 - stpipe.Image3Pipeline.tweakreg - WARNING - Too few input images (or groups of images) with non-empty catalogs.
2019-11-25 13:24:44,708 - stpipe.Image3Pipeline.tweakreg - WARNING - At least two exposures are required for image alignment.
2019-11-25 13:24:44,708 - stpipe.Image3Pipeline.tweakreg - WARNING - Nothing to do. Skipping 'TweakRegStep'...
2019-11-25 13:24:44,711 - stpipe.Image3Pipeline.tweakreg - INFO - Step tweakreg done
2019-11-25 13:24:44,711 - stpipe.Image3Pipeline - INFO - Matching sky values across all input images...
```
just as it's supposed to be.

Sorry for the spam. Will close the ticket again.

Issue #3761, opened 2019-07-01
Update radial weight index in cube build


stscijgbot:
Issue [JP-834](https://jira.stsci.edu/browse/JP-834) was created by David Law:

The  default cube building approach uses a modified Shepard interpolation scheme in which the weights decrease with increasing radius.  The index of this weight function is subject to optimization; initial tests in 2016 explored 1/r, 1/r^2, and e^-r forms.  We now have simulated data with sufficient spectral fidelity to show in B7.3 testing that the 1/r and 1/r^2 forms produce undesirable artifacts in individual-spaxel spectra on the wings of unresolved point sources at the 5% level corresponding to the sharp cutoffs in the region-of-interest given in the cubepar reference file.

Substantially better performance is found using the e^(-r) form instead, which declines more smoothly with radius.

See example image attached showing a MIRI simulation of a flat spectrum comparing the spaxel-to-spaxel spectrum (bottom left) and total integration spectrum (bottom right).  This weight function improves residuals from ~2-4% to below 1%.  These are both 4-point dithered images in MIRI Ch1A, build using standalone cube-building code ([https://github.com/STScI-MIRI/miri3d/blob/master/miri3d/modshep/modshep.py).]

Fixing this would primarily require an update to the cubepar reference files to an an extra extension to describe parameters to use for the exponential profile.  An additional code update would be required to swap out the 1/r line for an e^-r line once draft reference files are available.

Tagging [~nluetzgendorf] as this would affect NIRSpec as well.

!specqual.png!

stscijgbot:
Comment by Nora Luetzgendorf: From the NIRSpec side it is ok to go ahead with this. As discussed with [~dlaw] and [~sargent] today, we have decided to wait until this is implemented into the pipeline to run tests with simulated NIRSpec data. We will try to find out if this is the same for NIRSPec (if e^r is better) or not and what are the best cubepar paramters. In the meantime [~sargent] will create another extension for the CubePars with dummy parameters for the e^r weighting so that the pipeline run through.

stscijgbot:
Comment by David Law: Adding a note that a draft MIRI updated cubepar file has now been produced using [https://github.com/STScI-MIRI/miri3d/blob/master/miri3d/cubepar/make_cubepar.py] that can be used for testing.

stscijgbot:
[...]

PR #4313, opened 2019-12-03
Ifucube radial weighting and new reference files 

jemorrison:
DO NOT MERGE UNTIL NEW CUBEPARS REFERENCE FILES IN CRDS.
The PR addresses #3761 
Updated cube_build to read in the new reference files that contain the information to set up the radial weighting scheme.

### DIFF
--- a/jwst/cube_build/instrument_defaults.py
+++ b/jwst/cube_build/instrument_defaults.py
@@ -726,7 +726,7 @@ class InstrumentInfo():
                  self.high_wroi,
                  self.high_power,
                  self.high_softrad,
-                 self.hight_scalerad)
+                 self.high_scalerad)
         return table
 
     def GetMIRISliceEndPts(self, parameter1):
--- a/jwst/cube_build/instrument_defaults.py
+++ b/jwst/cube_build/instrument_defaults.py
@@ -552,6 +552,20 @@ class InstrumentInfo():
         self.Info[parameter1]['xstart'] = x1
         self.Info[parameter1]['xend'] = x2
 
+    def SetMSM(self, parameter1, parameter2, sroi, wroi, power, softrad):
+        self.Info[parameter1][parameter2]['sroi'] = sroi
+        self.Info[parameter1][parameter2]['wroi'] = wroi
+        self.Info[parameter1][parameter2]['msm_power'] = power
+        self.Info[parameter1][parameter2]['softrad'] = softrad
+        self.Info[parameter1][parameter2]['scalerad'] = None
+
+    def SetEMSM(self, parameter1, parameter2, sroi, wroi, scalerad):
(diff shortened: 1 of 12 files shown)


jwst/cube_build/instrument_defaults.py
```diff
@@ -605,42 +683,50 @@ def GetSoftRad(self, parameter1, parameter2):
...
         return table
 
     def Get_high_table(self):
         table = (self.high_wavelength,
                  self.high_sroi,
                  self.high_wroi,
                  self.high_power,
-                 self.high_softrad)
+                 self.high_softrad,
+                 self.hight_scalerad)
```
philhodge: Shouldn't this be `high_scalerad`?

drlaw1558:
Some quick comments from testing things:
- Is there no check that the specified weighting function exists in the cubepar file?  I.e., if I call cube_build with weighting="aardvark" I just get ERROR RUNNING STEP 'CubeBuildStep':
    Floating point nan values are not allowed in FITS headers
after a while.  This crash will be what we see until cubepar gets updated in crds.

- Overriding the reference file with my local new cubepar file and weighting="emsm" I get a data cube that doesn't look like I'd expect, details over on Jira ticket.

drlaw1558:
It looks like @jemorrison latest commit fixed the weighting issue where the radial scale radius wasn't being converted to pixel units properly.  I'm going to do some more quick tests with the updated code, updated the new cubepar file and deliver it to crds.  Once it's active in CRDS I'll approve this PR.

jdavies-st:
Restarted TravisCI build now that new cubepars reffiles are in CRDS.

jwst/cube_build/instrument_defaults.py
jemorrison: yes - fixed

jemorrison:
@drlaw1558 
@hbushouse 
@stscieisenhamer 
@jdavies-st 
This PR is ready to merge. The reference files are in CRDS. This pr will result in cube_build using a different weighting function to combine data, so the regression tests will fail. Since we are in the middle of updating how the regression tests are being done I don't know if we should hold off on this one or merge it ?

drlaw1558:
Where are we on being able to merge this PR?  Since this is a blocker issue for other work it would be helpful to merge this asap since all of the reference files are ready.

jemorrison:
it is ready to be merged. It will change the output cube_build files. @hbushouse @stscieisenhamer  the regression tests will fail. Do you want me to update the truth files before we merge this ?
We are still running the old regression system nightly - correct ?

hbushouse:
Yes, we're still running both the old and new regression systems, at least for now. We're hoping to turn off the old ones soon. Go ahead and merge and then look at the differences that result, then update the truth files.

Issue #4317, opened 2019-12-03
Remove DataModel.meta.model_type


stscijgbot:
Issue [JP-1162](https://jira.stsci.edu/browse/JP-1162) was created by Jonathan Eisenhamer:

{{DataModel.meta.model_type}} is a side effect of inserting the FITS header keyword {{DATAMODL}} into FITS-formatted datamodel output.

The use of {{DATAMODL}} is a good thing, as it identifies the type of model that is represented by the FITS file. However, the use of this keyword should exist outside and behind-the-scenes of the {{DataModel}} written to/read from FITS process. It should not be in the schema on be a visible component of {{DataModel}}.

As such, it is currently too easy to overwrite this attribute with conflicting values, resulting in unreadable FITS files.

The only other current usage in code is to check the model type of an instance. These checks should be using {{isinstance}} against the appropriate class instead of using this attribute.

melanieclarke:
@emolter @braingram - is this an issue we'd like to pursue?  The associated JP ticket is still open with standard priority, but seems to be disconnected from this parent ticket.

I note that we are using meta.model_type in several places in the current pipeline code.

braingram:
My vote would be to close this as no longer an issue.

`meta.model_type` is set to the class name on save:
https://github.com/spacetelescope/stdatamodels/blob/703f0155dae0891559dd032fe91b844fc71afccc/src/stdatamodels/model_base.py#L514-L515
so any in-memory modification would be ignored on save and I don't see a way a non-readable FITS file could be created.

For ASDF files this is the only way the model type is saved (and how stdatamodels knows what model to construct).

Using isinstance instead of checking meta.model_type seems less error prone but would be a jwst (not stdatamodels) change.

melanieclarke:
Okay, thank you.  I will close this ticket and the associated JP ticket.  I'll file a new clean up ticket to use isinstance instead of meta.model_type in jwst.

PR #4318, opened 2019-12-04
Define DataModel.model_type and force that when saving

stscieisenhamer:
When saving `DataModel`s to FITS, there is a core header keyword that is set, `DATAMODL` with the name of the model. This is used as a very strong hint as to what model to create when the file is read back in. However, being part of the model's schema, `meta.model_type`, it is easy to change this value, potentially creating a file of one model type but `DATAMODL` indicates another.

This change simply forces `DATAMODL` to be the current model type on saving.

Note: This is a partial fix for issue [JP-1162](https://jira.stsci.edu/browse/JP-1162) / #4317

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -239,15 +239,12 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
                     setattr(self, attr, value)
 
     @property
-    def model_type(self):
-        match = re.search(r"(\w+)'", str(type(self)))
-        if match:
-            return match.group(1)
-        return "DataModel"
-        
+    def _model_type(self):
+        return self.__class__.__name__
+
     def __repr__(self):
         buf = ['<']
-        buf.append(self.model_type)
+        buf.append(self._model_type)
 
         if self.shape:
             buf.append(str(self.shape))
@@ -473,7 +470,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
         self.meta.date = current_date.value
 
         # Enforce model_type to be the actual type of model being saved.
-        self.meta.model_type = self.model_type
+        self.meta.model_type = self._model_type
 
     def save(self, path, dir_path=None, *args, **kwargs):
         """
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -5,6 +5,7 @@ Data model class heirarchy
 import copy
 import datetime
 import os
+import re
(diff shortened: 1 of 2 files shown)


jwst/datamodels/model_base.py
```diff
@@ -237,16 +238,16 @@ def __init__(self, init=None, schema=None,
...
-        buf = ['<']
+    @property
+    def model_type(self):
         match = re.search(r"(\w+)'", str(type(self)))
         if match:
-            buf.append(match.group(1))
-        else:
-            buf.append("DataModel")
+            return match.group(1)
+        return "DataModel"
```
jdavies-st: A better way to do this without `re`:
```
    @property
    def model_type(self):
        return self.__class__.__name__
```

jdavies-st: Or just despense with the property altogether and use `self.__class__.__name__` elsewhere.

jwst/datamodels/model_base.py
```diff
@@ -471,6 +472,9 @@ def on_save(self, path=None):
         current_date.format = 'isot'
         self.meta.date = current_date.value
 
+        # Enforce model_type to be the actual type of model being saved.
+        self.meta.model_type = self.model_type
```
jdavies-st: Currently, `model_type` is not set for a base `DataModel`.  This changes that behavior.  I haven't quite worked through the ramifications of that.

The one thing this will change, is that if an old JWST FITS file without the DATAMODL keyword is read in and then written back out, it will now have the DATAMODL keyword.  Before this PR, datamodels would not modify FITS files without the DATAMODL keyword.

jwst/datamodels/model_base.py
stscieisenhamer: I would rather use a property, then spreading dunder-usage throughout code. And this encapsulates the definition.

About using re/type vs. the dunder, this is just a personal preference of trying to avoid dunders vs. using available public api. For this particular case, using dunder attributes is fine.

jwst/datamodels/model_base.py
stscieisenhamer: Offline discussion: This was determined to be fine.

jwst/datamodels/model_base.py
stscieisenhamer: Offline discussion: Property is fine, but make is private, `_model_type`. There is an issue of being able to pretend to set this, but that is a larger issue with the `DataModel` api.

PR #4319, opened 2019-12-04
Update README.md with final 0.14.2 release and install info

jdavies-st:


jdavies-st:
Yeah, I checked with Mary.  👍

PR #4322, opened 2019-12-06
Jp 1100 Implement MultiExposureModel throughout calibration

### LINKED ISSUE #4195: Should we handle calwebb_spec3 x1d products differently?
Issue [JP-1100](https://jira.stsci.edu/browse/JP-1100) was created by Howard Bushouse:

While testing updates to the calwebb_spec3 pipeline to handle non-TSO NIRISS SOSS exposures I noticed the following behavior regarding the different x1d products that are being created for an input ASN that contains 2 science members:
{code:java}
...
2019-10-28 16:49:41,804 - stpipe.Spec3Pipeline.extract_1d - INFO - Saved model in jw00307-o002_t001_niriss_clear-gr700xd_0_x1d.fits
2019-10-28 16:49:42,378 - stpipe.Spec3Pipeline.extract_1d - INFO - Saved model in jw00307-o002_t001_niriss_clear-gr700xd_1_x1d.fits
2019-10-28 16:49:42,379 - stpipe.Spec3Pipeline.extract_1d - INFO - Step extract_1d done
2019-10-28 16:49:42,425 - stpipe.Spec3Pipeline.combine_1d - INFO - Step combine_1d running with args (<ModelContainer>,).
2019-10-28 16:49:42,425 - stpipe.Spec3Pipeline.combine_1d - INFO - Step combine_1d parameters are: {'exptime_key': 'exposure_time'}
2019-10-28 16:49:42,428 - stpipe.Spec3Pipeline.combine_1d - INFO - Using exposure time as the weight.
2019-10-28 16:49:43,248 - stpipe.Spec3Pipeline.combine_1d - INFO - Saved model in jw00307-o002_t001_niriss_clear-gr700xd_x1d.fits
2019-10-28 16:49:43,248 - stpipe.Spec3Pipeline.combine_1d - INFO - Step combine_1d done
2019-10-28 16:49:43,248 - stpipe.Spec3Pipeline - INFO - Ending calwebb_spec3
2019-10-28 16:49:43,248 - stpipe.Spec3Pipeline - INFO - Step Spec3Pipeline done{code}
These data do not get combined or resampled at the 2D stage (using resample_spec), so the extract_1d step is being called multiple times to work on each of the individual input members. That results in individual x1d products getting saved for each input. Then the combine_1d step combines those individual x1d's into a final, single x1d product. The final x1d product has an appropriate source-based file name, taken from the product name specified in the ASN table.

But notice that the products saved by the extract_1d step also use this same source-based file name as a root, and in order to avoid file name clobbering it adds a numerical suffix (0, 1, 2, ...) to each product.

This same type of behavior will occur when processing NIRCam and NIRISS WFSS data, because they also do not get combined at the 2D stage, but instead execute extract_1d multiple times and then call combine_1d to create a combined x1d product.

Two things concern me about this:

1) Will the archive be prepared to handle the file names coming out of the extract_1d step - the file names that have the extra numerical field added to them?

2) Technically it isn't appropriate to be using a source-based file name for the products coming out of extract_1d, because they still correspond to data from a single exposure. But if we change this behavior to using the original exposure-based file name roots, we'll then have issues with file name clobbering with the x1d products that were already created for each exposure at the end of calwebb_spec2 processing.

Perhaps one way out of this is to not have extract_1d save its products in this case. For the case of NIRISS SOSS they are identical to the x1d's created at the end of calwebb_spec2 and hence aren't needed. All that's really of interest is the final x1d combing out of combine_1d. So perhaps just skip saving to files at the end of extract_1d and pass the data in memory only to combine_1d.

The source-based file naming used by extract_1d is appropriate for the case of WFSS data, because they go through the exp_to_source step and get reorganized into source-based products before calling extract_1d. But even in this case the archive may not be prepared to accept/catalog files that have the extra numerical field in their names.

stscieisenhamer:
Resolves #4195 and JP-1100.

For multi-object modes, `SourceModelContainer` was used as a blanket wrapper around `MultiExposureModel`. However, this masked some necessary information resulting in mistreatment of data. The blanket use of `SourceModelContainer` is removed, and applied only in specific cases in specific steps. Otherwise, the native `MultiExposureModel` is implemented.

As a side effect, the desire of JP-1100, to have non-TSO NIS_SOSS save the intermediate extract_1d results in a single file, is implemented.

### DIFF
--- a/jwst/extract_1d/extract_1d_step.py
+++ b/jwst/extract_1d/extract_1d_step.py
@@ -107,8 +107,8 @@ class Extract1dStep(Step):
         elif isinstance(input_model, datamodels.MultiSlitModel):
             self.log.debug('Input is a MultiSlitModel')
         elif isinstance(input_model, datamodels.MultiExposureModel):
-            self.log.warning('Input is a MultiExposureModel, '
-                             'which is not currently supported')
+            self.log.debug('Input is a MultiExposureModel, '
+                           'which is only supported for multi-object modes')
         elif isinstance(input_model, datamodels.MultiProductModel):
             self.log.debug('Input is a MultiProductModel')
         elif isinstance(input_model, datamodels.IFUCubeModel):
@@ -195,5 +195,8 @@ class Extract1dStep(Step):
             result.meta.cal_step.extract_1d = 'COMPLETE'
 
         input_model.close()
+
+        if len(result) and result[0].meta.exposure.type == 'NIS_SOSS':
+            result = datamodels.MultiExposureModel(result)
 
         return result
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -98,6 +98,9 @@ class OutlierDetection:
         if isinstance(self.inputs, datamodels.ModelContainer):
             self.input_models = self.inputs
             self.converted = False
+        elif isinstance(self.inputs, datamodels.MultiExposureModel):
+            self.input_models = datamodels.SourceModelContainer(self.inputs)
+            self.converted = True
         else:
             self.input_models = datamodels.ModelContainer()
             num_inputs = self.inputs.data.shape[0]
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -471,6 +471,9 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
         current_date.format = 'isot'
         self.meta.date = current_date.value
 
+        # Force the model_type to be what this model actually is.
+        self.meta.model_type = self.__class__.__name__
+
     def save(self, path, dir_path=None, *args, **kwargs):
         """
         Save to either a FITS or ASDF file, depending on the path.
--- a/jwst/datamodels/multiexposure.py
+++ b/jwst/datamodels/multiexposure.py
@@ -3,8 +3,10 @@ from copy import deepcopy
 from asdf import schema as asdf_schema
 from asdf import treeutil, AsdfFile
 
+from .container import ModelContainer
 from .model_base import DataModel
 from .image import ImageModel
+from .properties import merge_tree
 from .slit import SlitModel, SlitDataModel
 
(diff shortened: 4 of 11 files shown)


stscieisenhamer:
@hbushouse This is a first pass at resolving the issues discussed in JP-1100.

In general, a general simplification of the cal code, especially Spec3Pipeline, has been done, allowing non-TSO NIS_SOSS to be saved as a multi-object single file.

There is test data in /grp/jwst/ssb/eisenhamer/jwst_issues/jp-1100. In particular, the case of interest is under the folder jwstdms-83. The new, temporary suffix is x1dexps.

If you can determine whether this was the desired behavior, that would be great.

jwst/extract_1d/extract_1d_step.py
```diff
@@ -196,4 +196,7 @@ def process(self, input):
 
         input_model.close()
 
+        if len(result) and result[0].meta.exposure.type == 'NIS_SOSS':
```
hbushouse: What about other modes, such as NRC_WFSS and NIS_WFSS, that might need to be handled similarly to SOSS? Or do they use a different type of model/container that doesn't apply here?

jwst/outlier_detection/outlier_detection.py
```diff
@@ -98,6 +98,9 @@ def _convert_inputs(self):
         if isinstance(self.inputs, datamodels.ModelContainer):
             self.input_models = self.inputs
             self.converted = False
+        elif isinstance(self.inputs, datamodels.MultiExposureModel):
```
hbushouse: Not really relevant at this point in the code, but in general how does `MultiExposureModel` differ from a generic `ModelContainer`? Couldn't the data from multiple exposures, which presumably are all in individual data models of some type, just as easily be stored and processed from within a `ModelContainer` that just contains the list of individual exposure models?

melanieclarke:
JP-1100 was fixed by #4846, so I think this PR is no longer needed. @stscieisenhamer, please reopen or file an issue if more discussion is needed.

PR #4328, opened 2019-12-09
Deprecate MIRIRampModel

jdavies-st:
Failing regression tests due to #4318 indicated that we still have tests that use `MIRIRampModel`.  It was subsumed into `RampModel` in #3546.

This PR finishes the job in removing it from our tests, deprecating it and updating the docs.  If someone tries to instantiate `MIRIRampModel` now, they will get a `RampModel` returned and a `DeprecationWarning`:
```python-console

In [2]: m = datamodels.MIRIRampModel()
/Users/jdavies/dev/jwst/jwst/datamodels/ramp.py:51: DeprecationWarning: MIRIRampModel is deprecated and will be removed.  Use RampModel.
  "Use RampModel.", category=DeprecationWarning)

In [3]: m
Out[3]: <RampModel(0, 0, 0, 0)>
```


stscieisenhamer:
Has anyone pinged the IDT?

hbushouse:
> Has anyone pinged the IDT?

Yes, I've talked to Steven Beard:
```
Howard,
 
MIRI isn’t using the MiriRampModel data model, but it is using the miri_ramp.schema.yaml schema file. Do we need to switch to ramp.schema.yaml? It should cause us any trouble because the two schemas look identical. I just need to make the switch before we update to the next version of the JWST software.
 
Thanks for the warning.
 
Cheers,
 
Steven
```
So he's going to switch their code to no longer use the miri_ramp.schema.

jdavies-st:
Then I think this can be merged.

PR #4333, opened 2019-12-10
Make Image2Pipeline consistent across different instantiation methods

jdavies-st:
We discovered this in the Hack Day today:
```
strun calwebb_image2.cfg input_rate.fits
```
produces a `_cal` and `_i2d` file for MIRI Imaging mode.  As expected, as this is the way SDP runs the pipeline.

But the following:
```
strun jwst.pipeline.Image2Pipeline input_rate.fits
```
produces only the `_cal` file.  And if we try to explicitly force writing out the resampled file:
```
strun jwst.pipeline.Image2Pipeline input_rate.fits --steps.resample.save_results=True
```
produces the `_cal` file and a `_resample` file, not the `_i2d` file.

From within python:
```
result = Image2Pipeline.call("input_rate.fits")
```
no files were written out, as expected, with the `_cal` product returned to `result` in memory.  But
```
result = Image2Pipeline.call("input_rate.fits", save_results=True)
```
writes out the `_cal` but not the `_i2d`.  And
```
result = Image2Pipeline.call("input_rate.fits", config_file="calwebb_image2.cfg")
```
writes out the `_cal` file and the `i2d`, with no way to control whether it is or not.

So big inconsistencies in how the pipeline works depending on how it is run.

This PR fixes those for `Image2Pipeline`.  So now using a config file or the class name from the command line always writes out the `_cal` and `_i2d` files.  This banks on the fact that `strun` always switches on `self.save_results` for the pipeline it is running.  Writing out the resampled produced can piggyback on that.

And the consistency is now from within python as well.  No files are written out unless specifically requested:
```
Image2Pipeline.call("input_rate.fits", save_results=True)
```
outputs `_cal` and `_i2d` files as one would expect, but without `save_results` writes out nothing.

### DIFF
--- a/jwst/pipeline/calwebb_image2.py
+++ b/jwst/pipeline/calwebb_image2.py
@@ -153,9 +153,9 @@ class Image2Pipeline(Pipeline):
         # regular 2D science image types
         if input.meta.exposure.type.upper() in self.image_exptypes and \
         len(input.data.shape) == 2:
-            i2d = self.resample(input)
-            if self.save_results:
-                self.save_model(i2d, suffix="i2d")
+            self.resample.save_results = self.save_results
+            self.resample.suffix = 'i2d'
+            self.resample(input)
 
         # That's all folks
         self.log.info(
--- a/jwst/pipeline/calwebb_image2.py
+++ b/jwst/pipeline/calwebb_image2.py
@@ -153,7 +153,9 @@ class Image2Pipeline(Pipeline):
         # regular 2D science image types
         if input.meta.exposure.type.upper() in self.image_exptypes and \
         len(input.data.shape) == 2:
-            self.resample(input)
+            i2d = self.resample(input)
+            if self.save_results:
+                self.save_model(i2d, suffix="i2d")
 
         # That's all folks
         self.log.info(
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,7 +16,12 @@ master_background
 
 - Updated to fill the asn table and asn pool names. [#4240]
 
-=======
+pipeline
+--------
+
+- Make the naming and writing out of the resampled results to an `i2d` file
+  in `Image2Pipeline` consistent between config and class invocations [#4333]
+
 tweakreg
 --------
 
--- a/jwst/pipeline/calwebb_image2.cfg
+++ b/jwst/pipeline/calwebb_image2.cfg
@@ -1,15 +1,9 @@
 name = "Image2Pipeline"
 class = "jwst.pipeline.Image2Pipeline"
-save_results = True
 
     [steps]
       [[bkg_subtract]]
       [[assign_wcs]]
-        config_file = assign_wcs.cfg
       [[flat_field]]
-        config_file = flat_field.cfg
       [[photom]]
-        config_file = photom.cfg
       [[resample]]
-        config_file = resample.cfg
-        save_results = true


hbushouse:
I find it very surprising that
```
strun jwst.pipeline.Image2Pipeline input_rate.fits --steps.resample.save_results=True
```
results in the resampled output having a "resample" suffix. Is that because it's the step infrastructure constructing the name, rather than having stpipe figure it out when running from a cfg file? I can understand getting "resample" when running the ``resample`` step alone, but surprised it does that in the context of running the whole ``Image2Pipeline`` pipeline.

jwst/pipeline/calwebb_image2.py
```diff
@@ -153,7 +153,9 @@ def process_exposure_product(
         # regular 2D science image types
         if input.meta.exposure.type.upper() in self.image_exptypes and \
         len(input.data.shape) == 2:
-            self.resample(input)
+            i2d = self.resample(input)
+            if self.save_results:
+                self.save_model(i2d, suffix="i2d")
```
hbushouse: Isn't there something built-in somewhere that automatically knows (via a dictionary list or something like that) what suffix to use for products coming from a particular step or pipeline? Hence I'm wondering if this explicit reference to the "i2d" suffix is necessary. I'm sure @stscieisenhamer knows the answer.

jdavies-st: The `i2d` suffix is only defined in `resample.cfg`.

https://github.com/spacetelescope/jwst/blob/0b595e33d332cede2a7a69500e9bcb601a151954/jwst/pipeline/resample.cfg#L11

That's the only place it is defined anywhere.  So if that config file doesn't get parsed, the default, which is `resample`, gets used.

jdavies-st:
Yeah, I'll wait for @stscieisenhamer review before merging.  Clearly, this is not the only case where `save_results` is built into a config file (unnecessarily I believe).  Would be worth looking at all pipelines and making them consistent.

jwst/pipeline/calwebb_image2.py
hbushouse: I guess I was thinking of the list that's in /lib/suffix.py

stscieisenhamer: `/lib/suffix` only defines known suffixes; there is no single mapping of suffixes to product results. not sure that is possible nor desired.

There is a larger architecture discussion here, but the convention seems to be to leave it up to the pipelines to define what their outputs are. Hence, the fact that the resample step defines its suffix could be a point of discussion anyways. It may have been more appropriate to not be defined there but in the `calwebb_image2.cfg` file.

As such, having the pipeline hardcode the suffix is OK.

About the code itself, the current way is OK. However, it one wanted to be a bit more "keeping with the architecture", the alternate implementation would be:
```
self.resample.save_results = self.save_results
self.resample.suffix = 'i2d'
i2d = self.resample(input)
```
This in case , if `resample` did something else that is dependent on `save_results` and its suffix, it now has all the knowledge necessary.

jdavies-st: Thanks Jonathan.  That's a good suggestion and is a bit cleaner.

I think having the `i2d` suffix defined in `resample.cfg` is actually not good, as it gives different behavior from running with the default config vs running with the Step class.  No other steps defined the default suffix in their config files.  `resample_spec.cfg` defines suffix to be `s2d` in a similar way.

jdavies-st:
I'm not going to remove `suffix = "i2d"` in the `resample.cfg` yet as there are other pipelines that may depend on having this set.  Same with the `resample_spec.cfg`.  We should revisit this later.

hbushouse:
First regtest run since merging shows that the miri test_image2pipeline1 test no longer saves any output products (no _cal or _i2d):
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/482/testReport/jwst.tests_nightly.general.miri.test_image2pipeline_1/TestImage2Pipeline/_stable_deps__test_image2pipeline1/

jdavies-st:
Since this uses `Step.call()`, it needs to have `save_results=True` set.  I'll fix the test.

PR #4334, opened 2019-12-10
Add RegtestData.get_asn() to get asn and members from Artifactory

### LINKED ISSUE #4325: Implement get_association() method in RegtestData class
Issue [JP-1166](https://jira.stsci.edu/browse/JP-1166) was created by James Davies:

Add a `get_association()` method to `RegtestData` that downloads the specified association from Artifactory, and then parses it and grabs all the members.

jdavies-st:
This convenience method is used to get the remote association JSON file and then iteratively grabs all the members as well.

Resolves #4325

### DIFF
--- a/jwst/regtest/regtestdata.py
+++ b/jwst/regtest/regtestdata.py
@@ -8,6 +8,8 @@ from ci_watson.artifactory_helpers import (
     BigdataError,
 )
 
+from jwst.associations import load_asn
+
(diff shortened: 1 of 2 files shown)


jdavies-st:
The tests for the `RegtestData` data retrieval methods will run in the nightly regression tests because they need access to Artifactory.

PR #4336, opened 2019-12-10
Update template MIRI Image2 regtest

jdavies-st:
- Fix bug in `diff.report()`
- Use `Step.from_cmdline()` instead of `Image2Pipeline.call()`


jdavies-st:
Yeah, we'll prune this one once we have more examples.

Issue #4207, opened 2019-10-31
Modify extract_1d because of changes to photom


stscijgbot:
Issue [JP-1106](https://jira.stsci.edu/browse/JP-1106) was created by Philip Hodge:

The output of the photom step can now have different units from what was assumed for the current version of extract_1d, so extract_1d must be modified to account for the differences.

If the photom step has not been run on the input to extract_1d, extract_1d should log a warning and set the keyword for units for the "flux" column to count rate.

stscijgbot:
Comment by Howard Bushouse: In addition to different units for the "flux" column if the photom step hasn't been run, the extraction itself should be modified to not multiply by the pixel area (see JP-859).

philhodge:
> Comment by Howard Bushouse: In addition to different units for the "flux" column if the photom step hasn't been run, the extraction itself should be modified to not multiply by the pixel area (see JP-859).

This has been done, and it is included in PR #4369 .

hbushouse:
Fixed by #4369

stscijgbot:
Comment by Howard Bushouse: Fixed by [https://github.com/spacetelescope/jwst/pull/4369]

Issue #4040, opened 2019-09-19
Integration number can be negative


stscijgbot:
Issue [JP-1005](https://jira.stsci.edu/browse/JP-1005) was created by Philip Hodge:

Integration number can be negative

hbushouse:
Fixed by #4369

stscijgbot:
Comment by Howard Bushouse: Fixed by work done for JP-1106.

PR #4345, opened 2019-12-16
JP-1178: Updates for NRC_TSGRISM with NINTS=1

hbushouse:
While experimenting with the possibility of processing TSO exposures that have NINTS=1, I've determined that minor changes to a few level-2b (calwebb_tso-spec2) routines would be necessary to allow for this. The changes mainly involve backing off of the hardwired checks in many of the dedicated TSO routines that only allow for processing of a ``CubeModel``, as well as associated changes here and there to account for the fact that the data arrays are 2D instead of the expected 3D.

The changes made here to the ``photom`` step ``save_area_info`` function are actually needed for ALL modes of processing, because right now that function isn't getting called when there's no AREA reference file for the data. It does still need to be called, because as of recently it was updated to retrieve the pixel area keyword values from the PHOTOM ref file instead of the AREA ref file. So it can still be called even when there isn't an AREA ref file and in fact *must* be called for many such mode, because the pixel area keyword values are needed in downstream processing (e.g. ``extract_2d`` and/or ``extract_1d``).


jwst/extract_1d/extract.py
```diff
@@ -2983,7 +2983,7 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
                     else:
                         wcs = input_model.meta.wcs
                     # There's no need to repeat this for each integration.
-                    if integ == 0:
+                    if integ <= 1:
```
hbushouse: The reason for this change is due to the fact that in upstream code ``integrations`` is set to "[-1]" when the input data have a shape of 2 (as opposed to 3), so when executing this loop ``integ`` is equal to -1. The test for a value of zero was then causing this part of the if-block to get skipped, resulting the the pixel area values not getting populated. Now the pixel area is populated.

jwst/extract_2d/grisms.py
```diff
@@ -189,9 +189,9 @@ def extract_tso_object(input_model,
         ymax = int(ymax)
 
         # cut it out, keep the entire row though
-        ext_data = input_model.data[:, ymin: ymax + 1, :].copy()
-        ext_err = input_model.err[:, ymin: ymax + 1, :].copy()
-        ext_dq = input_model.dq[:, ymin: ymax + 1, :].copy()
+        ext_data = input_model.data[..., ymin: ymax + 1, :].copy()
+        ext_err = input_model.err[..., ymin: ymax + 1, :].copy()
+        ext_dq = input_model.dq[..., ymin: ymax + 1, :].copy()
```
hbushouse: This change allows for both 2D and 3D array shapes.

jwst/photom/photom.py
```diff
@@ -990,8 +991,8 @@ def apply_photom(self, photom_fname, area_fname):
 
         # Load the pixel area reference file, if it exists, and attach the
         # reference data to the science model
-        if area_fname != 'N/A':
-            self.save_area_info(ftab, area_fname)
+        #if area_fname != 'N/A':
+        self.save_area_info(ftab, area_fname)
```
hbushouse: Always call the ``save_area_info`` function and handle the absence of an AREA reference file within the function itself.

jwst/extract_1d/extract.py
philhodge: This is JP-1005, GitHub #4040.  However, this whole section will be deleted by addressing JP-1106, GitHub #4207.

hbushouse:
Updated to conform to recent changes to extract_1d/extract.py (my change no longer needed). Going to merge in order to get the updates to photom installed, which are needed to fix other problems unrelated to the TSO NINTS=1 issue. The other changes to accommodate TSO NINTS=1 don't hurt anything else.

PR #4354, opened 2019-12-17
JP-1181: NIRSpec calwebb_image2 regression test

hbushouse:
calwebb_image2 test using a NIRSpec imaging-mode exposure as input (in particular, EXP_TYPE='NRS_CONFIRM'). These kinds of exposures don't get resampled, so the only output is a _cal product, which is compared by the test.

Input and truth files have been uploaded to artifactory, at the locations specified in the test.

### DIFF
--- a/jwst/regtest/test_nirspec_image2.py
+++ b/jwst/regtest/test_nirspec_image2.py
@@ -1,0 +1,21 @@
+import os
+
+import pytest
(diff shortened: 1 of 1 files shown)


jdavies-st:
(except you don't need the `import os`).  You could also shorten up the name (and truth subdir) by removing the `_cal` from the end if you wanted.  Optional.

PR #4356, opened 2019-12-18
JP-1167: caldetector1 regtest for MIRI imaging

jemorrison:
Do not merge. I need to put the files on artifactory. Also this test runs the pipeline twice which is not needed I just wanted a reviewer to look at it and tell me  if the way I set up the fixture 'run_pipeline' was fine or did I need to run the pipeline the way 'test_miri_caldetector1_rate' runs it.


jwst/regtest/test_miri_caldector1.py
```diff
@@ -0,0 +1,93 @@
...
+def test_miri_caldetector1_rate(_jail, rtdata, fitsdiff_default_kwargs):
+    rtdata.get_data("miri/image/jw00001001001_01101_00001_MIRIMAGE_uncal.fits")
+
+    collect_pipeline_cfgs("config")
+    args = ["config/calwebb_detector1.cfg", rtdata.input]
+    Step.from_cmdline(args)
+    rtdata.output = "jw00001001001_01101_00001_MIRIMAGE_rate.fits"
+
+    rtdata.get_truth("truth/test_miri_caldetector1/jw00001001001_01101_00001_MIRIMAGE_rate.fits")
+    assert rtdata.output != rtdata.truth
```
hbushouse: I don't think you need this. It's just comparing the output and truth file names, which should be the same except for the path.

jwst/regtest/test_miri_caldector1.py
```diff
@@ -0,0 +1,93 @@
...
+    step = Detector1Pipeline()
+    step.lastframe.save_results = True
+    step.firstframe.save_results = True
+    step.dq_init.save_results = True
+    step.saturation.save_results = True
+    step.rscd.save_results = True
+    step.linearity.save_results = True
+    step.dark_current.save_results = True
+    step.refpix.save_results = True
+    step.jump.save_results = True
```
hbushouse: Technically the `jump` step product should be identical to the `ramp` product that you get as a result of setting `save_calibrated_ramp=True`, so you don't really need both of them.

jwst/regtest/test_miri_caldector1.py
```diff
@@ -0,0 +1,93 @@
...
+    files += glob('*_dq_init.fits')
+    files += glob('*_saturation.fits')
+    files += glob('*_rscd.fits')
+    files += glob('*_firstframe.fits')
+    files += glob('*_lastframe.fits')
+    files += glob('*_linearity.fits')
+    files += glob('*_jump.fits')
+    files += glob('*_dark_current.fits')
+    files += glob('*_refpix.fits')
+    files += glob('*_ramp.fits')
```
hbushouse: Could omit either the `jump` or `ramp` product, since they're identical.

jemorrison: Should I also delete the test_miri_caldetector1_rate test since I think the parametrized test checks the rate file

jwst/regtest/test_miri_caldector1.py
```diff
@@ -0,0 +1,93 @@
...
+    'jw00001001001_01101_00001_MIRIMAGE_dq_init.fits',
+    'jw00001001001_01101_00001_MIRIMAGE_firstframe.fits',
+    'jw00001001001_01101_00001_MIRIMAGE_lastframe.fits',
+    'jw00001001001_01101_00001_MIRIMAGE_saturation.fits',
+    'jw00001001001_01101_00001_MIRIMAGE_dark_current.fits',
+    'jw00001001001_01101_00001_MIRIMAGE_refpix.fits',
+    'jw00001001001_01101_00001_MIRIMAGE_jump.fits',
+    'jw00001001001_01101_00001_MIRIMAGE_ramp.fits',],
+    ids=['rate', 'rateints', 'linearity', 'rscd', 'dq_int', 'firstframe',
+         'lastframe', 'saturation', 'dark_current', 'refpix', 'jump', 'ramp'])
```
jdavies-st: You could parametrize the suffix, and then you wouldn't need `ids`.  So something like:
```
@pytest.mark.parametrize("output", ['rate', 'rateints', 'linearity', 'rscd', ...])
```
and
```
rtdata.output = "jw00001001001_01101_00001_MIRIMAGE_" + output + ".fits"
```

jwst/regtest/test_miri_caldector1.py
jdavies-st: All these files are checked in the parametrized test below, so probably not necessary to do it here.  If you did want to verify there N files at the end of the pipeline run, you could just add a single `glob('*.fits')` and the assert in this test.

jwst/regtest/test_miri_image_caldector1.py
```diff
@@ -0,0 +1,45 @@
...
+    step.save_results = True
+    step.run(rtdata.input)
+    return rtdata
+
+
+@pytest.mark.bigdata
+@pytest.mark.parametrize("output", ['rate', 'rateints', 'linearity', 'rscd',
+                                    'dq_init', 'firstframe', 'lastframe',
+                                    'saturation', 'dark_current', 'refpix', 'jump'])
+def test_miri_image_caldetector1(run_pipeline, request, fitsdiff_default_kwargs, output):
```
hbushouse: Just for my own knowledge, what's the "request" argument do here?

jemorrison:
@hbushouse I changed the test as you suggested and now most of the tests fail because the save_results differ with the truth files.  The data differs (a lot like 98% of the pixels) . I am not sure I understand why this is. Why would changing how the pipeline is run result in different values ?

hbushouse:
Probably because when you ran it the first time to generate the truth files the ``ipc`` step got applied, whereas now that step is being skipped, hence the difference in results. One way to confirm is to run it again (locally) with ``--steps.ipc.skip=False`` added to the argument list and see if the results are now the same as the truth files.

hbushouse:
Just need to fix a trailing whitespace to placate flake8 and you should be good to go.

jwst/regtest/test_miri_image_caldector1.py
```diff
@@ -0,0 +1,48 @@
...
+    Step.from_cmdline(args)
+    return rtdata
+
+
+@pytest.mark.bigdata
+@pytest.mark.parametrize("output", ['rate', 'rateints', 'linearity', 'rscd',
+                                    'dq_init', 'firstframe', 'lastframe',
+                                    'saturation', 'dark_current', 'refpix',
+                                    'jump'])
+def test_miri_image_caldetector1(run_pipeline, request, fitsdiff_default_kwargs, output):
```
jdavies-st: I think a name `test_miri_image_detector1` here would be preferable.  "cal" is unnecessary just makes it longer.

jdavies-st:
Btw, how long does this test take to run?  I notice that you're not setting `--steps.jump.rejection_threshold` to be higher than the default 4 sigma.  Is it detecting a ton of jumps and then taking forever in ramp fit?  Or is it ok?

jemorrison:
I will change the name and I am running a timing of how long the test take now.

jemorrison:
It takes a pretty long time to run. About 1 hour. I can change the sigma threshold for the jump step to
a higher value.  6? Or do you recommend higher ?

hbushouse:
6 won't make much difference, compared to the default of 4. Set it to at least 25 or 50.

jemorrison:
Ok I will increase it to 25 first. But this is makes is 25 sigma right ? Gosh that is high- a statistically crazy value. But it is noisy is unrealistically large I guess that is ok.

jemorrison:
changes made after review. Reloaded truth files to artifactory

PR #4357, opened 2019-12-18
JP-1170 Unit test for NIRSpec subarray and grouped data.

philhodge:


philhodge:
Yes, of course.  I removed the first test.

PR #4359, opened 2019-12-18
Jp 1168 Update NIRCam level 2a regression test

ddavis-stsci:
Updated the NIRCam level 2a regression test. Running the test locally 
```
pytest --env=newdev --bigdata --basetemp=/Users/ddavis/src/JWST/develop/JP-1168/test_Dec18a ~/src/JWST/scsb/jwst/jwst/regtest/test_NIRcam_image2a.py
=========================================================== test session starts ============================================================
platform darwin -- Python 3.7.4, pytest-5.2.2, py-1.8.0, pluggy-0.13.0
plugins: xdist-1.30.0, requests-mock-1.7.0, doctestplus-0.4.0, openfiles-0.4.0, ci-watson-0.4, forked-1.1.3, cov-2.8.1, asdf-2.4.3a1.dev10+g1b41c6d
collected 4 items

../../scsb/jwst/jwst/regtest/test_NIRcam_image2a.py ....                                                                             [100%]

============================================================ 4 passed in 29.05s ============================================================   
```


jwst/regtest/test_NIRcam_image2a.py
```diff
@@ -0,0 +1,66 @@
...
+from glob import glob
+
+import pytest
+
+from astropy.io.fits.diff import FITSDiff
+from jwst.pipeline import Image2Pipeline
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+
+@pytest.mark.bigdata
+def test_NIRcam_image2a_cal(rtdata, fitsdiff_default_kwargs, _jail):
```
stscieisenhamer: I believe this test is a duplicate/subset of the test_NIRCam_image2a and can be safely removed.

jwst/regtest/test_NIRcam_image2a.py
```diff
@@ -0,0 +1,66 @@
+""" Test for the image2 pipeline using NIRCam image data. """
```
hbushouse: The intention of ticket JP-1168 was to write a test of the calwebb_detector1 (level-2a) pipeline, not the calwebb_image2 (level-2b) pipeline. So all references to image2 and ``Image2Pipeline`` need to be changed to ``Detector1Pipeline``. That also means the input needs to be an uncal product (instead of rate) and the outputs to be tested need to be _rate and _rateints. Alternatively, this code could be turned into satisfying JP-1179, which as intended to be a test of the ``Image2Pipeline`` on NIRCam images.

ddavis-stsci:
Rewrote test to test level2a (not image 2a) pipeline. The local pytest results,
= 5 passed, 2 warnings in 254.88s (0:04:14) ==

jwst/regtest/test_NIRcam_level2a.py
```diff
@@ -0,0 +1,66 @@
...
+    file and generate the stage 1 FITS files (rate, ratints, and trapsfilled)."""
+import os
+from glob import glob
+import pytest
+
+from astropy.io.fits.diff import FITSDiff
+from jwst.pipeline import Detector1Pipeline
+
+@pytest.mark.bigdata
+def test_NIRcam_stage1_rate(rtdata, fitsdiff_default_kwargs, _jail):
```
stscieisenhamer: This test is redundant; this case is also covered by test_NIRCam_stage1_completion. The intent of this test in the template was to demonstrate running just a single check.  Everyone has been doing this.

jwst/regtest/test_NIRcam_level2a.py
```diff
@@ -0,0 +1,66 @@
...
+
+@pytest.fixture(scope="module")
+def run_pipeline(jail, rtdata_module):
+    """Run calwebb_detector1 pipeline on NIRCam uncalibrated data. For these data the
+       rejection_threshold is increased from the default of 4 to 150 to reduce the
+       runtime to a managable level."""
+    rtdata = rtdata_module
+    rtdata.get_data("nircam/test_detector1pipeline/jw82500001003_02101_00001_NRCALONG_uncal.fits")
+
+    Detector1Pipeline.call(rtdata.input, steps={'jump': {'rejection_threshold': 150}},
```
hbushouse: As I mentioned in comments to #4356 it's not good to be running the tests via either the .call() or .run() methods, without using a cfg file, because the cfg files often have several params set to skip certain steps for certain modes of data. In the case of calwebb_detector1.cfg, it is hardwired to always skip the ``ipc`` step. By using the .call() method to run the pipeline, the ``ipc`` step will be applied, resulting in non-standard output.

You still have the ability to pass params as arguments when running from the command line. For example, this could be done with:
```
args = ["calwebb_detector1.cfg", rtdata.input, "--steps.jump.rejection_threshold=150", "--save_results=True"]
Step.from_cmdline(args)
```
of course you have to first collect the cfg file(s) or just include the calwebb_detector1.cfg file in the input data directory on artifactory.

Also, regarding the jump.rejection_threshold, we don't want to waste a lot of unnecessary time by having lots of bogus jumps flagged, but at the same time we also want to test the ability to detect and handle at least *some* flagged jumps. So take a look at the log from the jump step to verify that at least some (> 0) pixels got flagged as containing jumps. If not, you may need to decrease the rejection_threshold a bit till you find an appropriate "sweet spot."

ddavis-stsci:
Updated the test so that the intermediate steps are saved and checked. The tests show
`19 passed, 1 warnings in 2663.98s (0:44:23`
I'll have the check the speed again with a better connection, my home network is currently very slow.

Issue #4032, opened 2019-09-17
Bug in function to compute pixel solid angle


stscijgbot:
Issue [JP-999](https://jira.stsci.edu/browse/JP-999) was created by Philip Hodge:

Function pixel_area in extract_1d/util.py computes the solid angle of a pixel, based on the mean values of the differences in the right ascension and declination from pixel to pixel over the image. This function does not take into account the possibility that the right ascension can cross zero.

stscijgbot:
Comment by Jonathan Eisenhamer: No longer needed because pixel area comes from keywords. Part of the photom reference refactor.

PR #4369, opened 2019-12-19
JP-1106 Modify extract_1d because of changes to photom

philhodge:
The input file may have one set of units or another, depending on the instrument and whether the target is a point source or extended; this is due to recent changes to the `photom` step.  As a result, the way the table columns are computed by the `extract_1d` step depends on the configuration.

There's now a check for whether the `photom` step has been run, and if it has not been, the output from `extract_1d` will be in units of DN/s.

The pixel solid angle is now gotten from a header keyword, rather than being computed from the WCS.

The source type (point or extended) is gotten from either the primary header or an extension header, depending on the input data model.

Closes  JP-1106, GitHub #4207 and JP-859, GitHub #3477
These tickets can be closed:  JP-999, GitHub #4032 and JP-1005, GitHub #4040


hbushouse:
I'm going to merge and kick off a regtest run, so that we can get started ASAP on making any necessary truth file updates.

PR #4379, opened 2019-12-20
Jp 1165 Implement FGS GUIDER pipeline regression

### LINKED ISSUE #4324: REGTEST: Write new FGS guidestar level2a regression test
Issue [JP-1165](https://jira.stsci.edu/browse/JP-1165) was created by James Davies:

Implement FGS guidestar level2a regression test for ID and FINEGUIDE modes.

Data and current regtest exists for ID mode.

stscieisenhamer:
Resolves #4324/[JP-1165](https://jira.stsci.edu/browse/JP-1165)

### DIFF
--- a/jwst/regtest/test_fgs.py
+++ b/jwst/regtest/test_fgs.py
@@ -60,8 +60,8 @@ def run_guider_pipelines(jail, rtdata_module, request):
 @pytest.mark.bigdata
 @pytest.mark.parametrize(
     'run_guider_pipelines',
-    ['exptype_fgs_id_image'],
-    ids=['fgs_id_image'],
+    ['exptype_fgs_acq1', 'exptype_fgs_id_image', 'exptype_fgs_id_stack'],
+    ids=['fgs_acq1', 'fgs_id_image', 'fgs_id_stack'],
     indirect=True
 )
 @pytest.mark.parametrize(
@@ -71,11 +71,12 @@ def run_guider_pipelines(jail, rtdata_module, request):
 )
 def test_fgs_id_image(run_guider_pipelines, fitsdiff_default_kwargs, suffix):
     rtdata = run_guider_pipelines
-    rtdata.output = replace_suffix(
+    output = replace_suffix(
         os.path.splitext(os.path.basename(rtdata.input))[0], suffix
     ) + '.fits'
+    rtdata.output = output
 
-    rtdata.get_truth(os.path.join('fgs/truth/test_fgs', rtdata.output))
+    rtdata.get_truth(os.path.join('fgs/truth/test_fgs', output))
 
     diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
     assert diff.identical, diff.report()
--- a/jwst/regtest/test_fgs.py
+++ b/jwst/regtest/test_fgs.py
@@ -4,24 +4,81 @@ import os
 import pytest
 from astropy.io.fits.diff import FITSDiff
 
+from jwst.lib.suffix import replace_suffix
 from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
 from jwst.stpipe import Step
 
 
 @pytest.fixture(scope='module')
-def run_pipeline(jail, rtdata_module):
+def run_fgs_imaging_pipelines(jail, rtdata_module):
     """Run the pipelines"""
     rtdata = rtdata_module
     rtdata.get_data('fgs/level1b/exptype_fgs_image_uncal.fits')
 
     collect_pipeline_cfgs('config')
     args = [
-        'config/calwebb_guider.cfg',
-        rtdata.input
+        'config/calwebb_detector1.cfg',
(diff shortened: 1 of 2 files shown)


jwst/regtest/test_fgs.py
```diff
@@ -0,0 +1,88 @@
...
+    rtdata.output = output
+
+    rtdata.get_truth(os.path.join(truth_path, output))
+
+    diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
+    assert diff.identical, diff.report()
+
+
+@pytest.fixture(scope='module')
+def run_fgs_imaging_pipelines(jail, rtdata_module):
```
hbushouse: We didn't really need a test of the imaging (science) mode data, but it's OK to have it.

jwst/regtest/test_fgs.py
```diff
@@ -0,0 +1,88 @@
...
+    rtdata = rtdata_module
+    rtdata.get_data('fgs/level1b/' + request.param + '_uncal.fits')
+
+    collect_pipeline_cfgs('config')
+    args = [
+        'config/calwebb_guider.cfg',
+        rtdata.input,
+        '--steps.dq_init.save_results=true',
+        '--steps.guider_cds.save_results=true',
+        '--steps.flat_field.save_results=true',
```
hbushouse: Theoretically the output of the ``flat_field`` step, which is the last step in the pipeline, should be identical to the _cal product, so you don't really need to save and compare them. But again, it can't hurt (it's just extra time spent doing the same thing twice).

jwst/regtest/test_fgs.py
```diff
@@ -0,0 +1,88 @@
...
+        '--steps.refpix.save_results=true',
+        '--steps.firstframe.save_results=true',
+        '--steps.lastframe.save_results=true',
+        '--steps.linearity.save_results=true',
+        '--steps.dark_current.save_results=true',
+        '--steps.persistence.save_results=true',
+        '--steps.jump.save_results=true',
+        '--steps.ramp_fit.save_results=true',
+        '--steps.gain_scale.save_results=true',
+    ]
```
hbushouse: Do you also need to set the --steps.jump.rejection_threshold param to a high value, so that it doesn't take 2.5 hours to run?

jwst/regtest/test_fgs.py
stscieisenhamer: Part of my ignorance of FGS terminology. Realized later that the specification of GUIDESTAR does not, in fact, include the science side of things.

jwst/regtest/test_fgs.py
stscieisenhamer: Ah yes. Would it be appropriate to actually make this check against the "final" output instead? Or does any of this really matter?

jwst/regtest/test_fgs.py
stscieisenhamer: Actually thought I had taken that out, since it did nothing about the ramp_fit runtime. Whatever is the reason for the long running there, it seems to be invariant against what jump found.

hbushouse: Hmmm, so if it takes 2.5 hours to run, even when no jumps are flagged, then that's more incentive yet to not have the science-mode (FGS_IMAGE) test in there at all.

jwst/regtest/test_fgs.py
hbushouse: You're already doing a check against the "final" output by having "cal" in the parameterized output list below. So the "flat_field" could be removed, because it's redundant with "cal" (same data, just saved in a different file).

jwst/regtest/test_fgs.py
stscieisenhamer: Removing!

jwst/regtest/test_fgs.py
```diff
@@ -0,0 +1,48 @@
+"""Regression tests for FGS Guidestar in ID and FINEGUIDE modes"""
```
hbushouse: Minor nit, but the actual tests appear to be covering ID and ACQ1 modes, not FINEGUIDE.

hbushouse:
Can this be merged or are you still hoping to find some FINEGUIDE mode data to add to the test (or swap it for the ACQ1 data)?

stscieisenhamer:
Was looking for FGS_FINEGUIDE, but there isn't any in the SDP dataset. Unless anyone else has any in-hand, then this can be considered complete.

hbushouse:
> Was looking for FGS_FINEGUIDE, but there isn't any in the SDP dataset. Unless anyone else has any in-hand, then this can be considered complete.

Looks like they didn't copy over any of the guiding datasets and the originals have probably been wiped out on the D-string by now. I have one fineguide mode dataset that isn't too terribly old. It's at /grp/jwst/ssb/bushouse/jwst_data/FGS/Guiding/jw00624003002_gs-fg_2019036004816_uncal.fits. Could try that one.

stscieisenhamer:
Added the FINEGUIDE case.

PR #4380, opened 2019-12-21
JP-1196: Implement NIR WFSS Level2b spectral test

### LINKED ISSUE #4375: REGTEST: Write new WFSS level2b regression test
Issue [JP-1196](https://jira.stsci.edu/browse/JP-1196) was created by Howard Bushouse:

Create a regression test in the new regtest system for applying the ``calwebb_spec2`` (``Spec2Pipeline``) pipeline to a Wide Field Slitless Spectroscopy (WFSS) exposure from either NIRCam (EXP_TYPE='NRC_WFSS') or NIRISS (EXP_TYPE='NIS_WFSS'). This requires the use of a valid source catalog derived from calibration processing of associated direct images of the field. One such test for NIRISS WFSS exists in the old system.

stscieisenhamer:
Resolves #4375/[JP-1196](https://jira.stsci.edu/browse/JP-1196)

### DIFF
--- a/jwst/regtest/regtestdata.py
+++ b/jwst/regtest/regtestdata.py
@@ -38,7 +38,7 @@ class RegtestData:
         self.test_name = test_name
         self.traceback = traceback
 
-        # Initialize non-initialized attriburtes
+        # Initialize non-initialized attributes
         self.asn = None
 
     def __repr__(self):
--- a/jwst/regtest/regtestdata.py
+++ b/jwst/regtest/regtestdata.py
@@ -38,6 +38,8 @@ class RegtestData:
         self.test_name = test_name
         self.traceback = traceback
 
+        # Initialize non-initialized attriburtes
+        self.asn = None
 
     def __repr__(self):
         return pprint.pformat(
@@ -174,6 +176,7 @@ class RegtestData:
         # Get each member in the association as well
         with open(self.input) as fp:
             asn = load_asn(fp)
+        self.asn = asn
         for product in asn['products']:
             for member in product['members']:
                 fullpath = os.path.join(
--- a/jwst/regtest/test_niriss.py
+++ b/jwst/regtest/test_niriss.py
@@ -58,9 +58,11 @@ def run_nrs_wfss_spectral(jail, rtdata_module):
     return rtdata
 
 
-wfss_suffixes = ['assign_wcs', 'bsub', 'cal', 'extract_2d', 'flat_field', 'photom', 'srctype', 'x1d']
 @pytest.mark.bigdata
-@pytest.mark.parametrize('suffix', wfss_suffixes, ids=wfss_suffixes)
+@pytest.mark.parametrize(
+    'suffix',
+    ['assign_wcs', 'bsub', 'cal', 'extract_2d', 'flat_field', 'photom', 'srctype', 'x1d']
+)
 def test_nrs_wfss_spectral(run_nrs_wfss_spectral, fitsdiff_default_kwargs, suffix):
     """Regression test matching output files"""
     fitsdiff_default_kwargs['ignore_keywords'].append('SCATFILE')
--- a/jwst/regtest/test_niriss.py
+++ b/jwst/regtest/test_niriss.py
@@ -1,0 +1,67 @@
+"""Regression tests for NIRISS"""
+import os
+
+import pytest
+from astropy.io.fits.diff import FITSDiff
+
+from jwst.lib.suffix import replace_suffix
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+from jwst.stpipe import Step
+
+
+def is_like_truth(rtdata, fitsdiff_default_kwargs, suffix, truth_path='truth/niriss/test_niriss'):
+    """Compare step outputs with truth"""
+
+    # If the input was an association, the output should be the name of the product
+    # Otherwise, output is based on input.
+    if rtdata.asn:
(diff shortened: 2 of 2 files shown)


jwst/regtest/regtestdata.py
```diff
@@ -38,6 +38,8 @@ def __init__(self, env="dev", inputs_root="jwst-pipeline",
         self.test_name = test_name
         self.traceback = traceback
 
+        # Initialize non-initialized attriburtes
```
hbushouse: typo: "attriburtes" -> "attributes"

jwst/regtest/test_niriss.py
```diff
@@ -0,0 +1,67 @@
...
+def run_nrs_wfss_spectral(jail, rtdata_module):
+    """Run the pipelines"""
+    rtdata = rtdata_module
+    rtdata.get_asn('niriss/level2a/nir_wfss_spec2_asn.json')
+
+    collect_pipeline_cfgs('config')
+    args = [
+        'config/calwebb_spec2.cfg',
+        rtdata.input,
+        '--steps.bkg_subtract.save_results=true',
```
hbushouse: Not all of these steps get applied to WFSS data, such as imprint_subtract, msa_flagging, straylight, fringe, barshadow, and resample_spec, so at the very least it's not necessary to save all of these outputs and I'm also wondering if they get produced at all when a step is skipped? Does the pipeline only create the ones for steps that are actually run?

jwst/regtest/test_niriss.py
```diff
@@ -0,0 +1,67 @@
...
+        '--steps.resample_spec.save_results=true',
+        '--steps.cube_build.save_results=true',
+        '--steps.extract_1d.save_results=true',
+    ]
+    Step.from_cmdline(args)
+
+    return rtdata
+
+
+wfss_suffixes = ['assign_wcs', 'bsub', 'cal', 'extract_2d', 'flat_field', 'photom', 'srctype', 'x1d']
```
hbushouse: Ah, I see that here you narrow it down to only the products that are actually created.

jwst/regtest/test_niriss.py
```diff
@@ -0,0 +1,67 @@
...
+@pytest.mark.bigdata
+@pytest.mark.parametrize('suffix', wfss_suffixes, ids=wfss_suffixes)
+def test_nrs_wfss_spectral(run_nrs_wfss_spectral, fitsdiff_default_kwargs, suffix):
+    """Regression test matching output files"""
+    fitsdiff_default_kwargs['ignore_keywords'].append('SCATFILE')
```
hbushouse: Why do you want to ignore the SCATFILE keyword? Shouldn't it be the same every time the test is run?

jwst/regtest/regtestdata.py
```diff
@@ -174,6 +176,7 @@ def get_asn(self, path=None, docopy=None):
         # Get each member in the association as well
         with open(self.input) as fp:
             asn = load_asn(fp)
+        self.asn = asn
```
jdavies-st: This is a good addition.  It may help later on in OKifying assocation inputs if needed.  Thanks!

jwst/regtest/test_niriss.py
```diff
@@ -0,0 +1,67 @@
...
+        '--steps.extract_1d.save_results=true',
+    ]
+    Step.from_cmdline(args)
+
+    return rtdata
+
+
+wfss_suffixes = ['assign_wcs', 'bsub', 'cal', 'extract_2d', 'flat_field', 'photom', 'srctype', 'x1d']
+@pytest.mark.bigdata
+@pytest.mark.parametrize('suffix', wfss_suffixes, ids=wfss_suffixes)
```
jdavies-st: I don't think you need `ids` here since they are the same as the suffixes.  Without `ids`, the tests will have the same name when you do
```
pytest -v
```

jwst/regtest/test_niriss.py
stscieisenhamer: I was going to edit out the ones that didn't produce anything. Will do that.

jwst/regtest/test_niriss.py
stscieisenhamer: This is actually a question: This seems to get set when a catalog is specified through the association, and it is set to the whole path. As such, it changes per run.

jwst/regtest/test_niriss.py
hbushouse: Sounds good. In the meantime I've confirmed that steps that get skipped do NOT create an output, so leaving them in should be harmless.

jwst/regtest/test_niriss.py
hbushouse: Ah, I see. I guess that makes sense. The normal Ops scheme will have the source catalog file reference in the ASN file, so it's a good idea to then populate the SCATFILE keyword as a record of what was used to create the output products. But I'm surprised it keeps the whole path in the value. What would the whole path be in the Ops environment? Surely something that no mere mortal can get to.

stscieisenhamer: I would agree that it should just be the basename. Will look into that now.

stscieisenhamer: JP-1201 created for this parrticular issue. Otherwise, the test will continue to ignore this.

jwst/regtest/test_niriss.py
stscieisenhamer: Wasn't sure there was not some extra use in the rest of the testing architecture.

PR #4381, opened 2019-12-21
JP-1179: Create regression test for NIRCam image2 (plus stages 1 and 3)

### LINKED ISSUE #4349: REGTEST: Write new NIRCam imaging level2b regression test
Issue [JP-1179](https://jira.stsci.edu/browse/JP-1179) was created by Howard Bushouse:

Create a regression test in the new regtest system for applying the calwebb_image2 pipeline to NIRCam imaging data. Could be for a single exposure or multiple exposures that will naturally feed into a subsequent calwebb_image3 test for NIRCam imaging.

Could be piggy-backed onto the NIRCam imaging calwebb_detector1 test from JP-1168 (i.e. add a calwebb_image2 test to the calwebb_detector1 test that uses the output from the calwebb_detector1 test).

If creating a test that runs calwebb_image2 on multiple exposures - to feed into a subsequent calwebb_image3 test - use exposures from 4 of the NIRCam detectors that make up one of the NIRCam short-wavelength modules (A or B) that also simulate at least 3 dither positions (i.e. a total of 12 images). calwebb_image2 runtime should not be that expensive for 12 images (it's only 4 steps: assign_wcs, flat_field, photom, and resample).

jdavies-st:
This implements stages 1, 2 and 3 pipelines for Mirage-simulated NIRCam imaging dataset `jw42424001001`.

One `_uncal` file is processed through to the end of calwebb_image3.  The other 5 members of the association are processed from the `_rate` file through stage 2 and 3 pipelines.

Currently the dataset has the wrong pointing info, so the final `_i2d` mosaic doesn't look great.  Will resolve this with fixed input and truth data in the near future.

Also cleaned up the postmortem path handling, which should have no effect on functionality.

Test results:
```
$ pytest --bigdata --env=newdev test_nircam_image.py -v
============================ test session starts ============================
platform darwin -- Python 3.7.5, pytest-5.3.1, py-1.8.0, pluggy-0.13.1 -- /Users/jdavies/miniconda3/envs/jwst/bin/python
cachedir: .pytest_cache
plugins: asdf-2.4.3a1.dev13+gc1373db, xdist-1.30.0, requests-mock-1.7.0, openfiles-0.4.0, ci-watson-0.4, forked-1.1.3, doctestplus-0.5.0, cov-2.8.1
collected 15 items                                                                                                                                    

test_nircam_image.py::test_nircam_image_stages12[dq_init] PASSE        [  6%]
test_nircam_image.py::test_nircam_image_stages12[saturation] PASSED    [ 13%]
test_nircam_image.py::test_nircam_image_stages12[superbias] PASSED     [ 20%]
test_nircam_image.py::test_nircam_image_stages12[refpix] PASSED        [ 26%]
test_nircam_image.py::test_nircam_image_stages12[linearity] PASSED     [ 33%]
test_nircam_image.py::test_nircam_image_stages12[trapsfilled] PASSED   [ 40%]
test_nircam_image.py::test_nircam_image_stages12[dark_current] PASSED  [ 46%]
test_nircam_image.py::test_nircam_image_stages12[jump] PASSED          [ 53%]
test_nircam_image.py::test_nircam_image_stages12[rate] PASSED          [ 60%]
test_nircam_image.py::test_nircam_image_stages12[flat_field] PASSED    [ 66%]
test_nircam_image.py::test_nircam_image_stages12[cal] PASSED           [ 73%]
test_nircam_image.py::test_nircam_image_stages12[i2d] PASSED           [ 80%]
test_nircam_image.py::test_nircam_image_stage2_wcs PASSED              [ 86%]
test_nircam_image.py::test_nircam_image_stage3_i2d PASSED              [ 93%]
test_nircam_image.py::test_nircam_image_stage3_catalog PASSED          [100%]
```

Resolves #4349

### DIFF
--- a/jwst/regtest/test_nircam_image.py
+++ b/jwst/regtest/test_nircam_image.py
@@ -3,9 +3,12 @@ import os
 import pytest
 from astropy.io.fits.diff import FITSDiff
 from astropy.table import Table, setdiff
+from gwcs.wcstools import grid_from_bounding_box
+from numpy.testing import assert_allclose
 
 from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
 from jwst.stpipe import Step
+from jwst import datamodels
 
 
 @pytest.fixture(scope="module")
@@ -33,12 +36,13 @@ def run_pipelines(jail, rtdata_module):
     rtdata.input = "jw42424001001_01101_00001_nrca5_rate.fits"
     args = ["config/calwebb_image2.cfg", rtdata.input,
         "--steps.assign_wcs.save_results=True",
+        "--steps.flat_field.save_results=True",
         ]
     Step.from_cmdline(args)
 
     # Grab rest of _rate files for the asn and run image2 pipeline on each to
     # produce fresh _cal files for the image3 pipeline.  We won't check these
-    # or look at intermediate products
+    # or look at intermediate products, including the resampled i2d
     rate_files = [
     "nircam/image/jw42424001001_01101_00001_nrcb5_rate.fits",
     "nircam/image/jw42424001001_01101_00002_nrca5_rate.fits",
@@ -68,15 +72,13 @@ def run_pipelines(jail, rtdata_module):
 @pytest.mark.bigdata
 @pytest.mark.parametrize("suffix", ["dq_init", "saturation", "superbias",
     "refpix", "linearity", "trapsfilled", "dark_current", "jump", "rate",
-    "assign_wcs", "cal", "i2d"])
+    "flat_field", "cal", "i2d"])
 def test_nircam_image_stages12(run_pipelines, fitsdiff_default_kwargs, suffix):
     """Regression test of detector1 and image2 pipelines performed on NIRCam data."""
     rtdata = run_pipelines
     rtdata.input = "jw42424001001_01101_00001_nrca5_uncal.fits"
     output = "jw42424001001_01101_00001_nrca5_" + suffix + ".fits"
     rtdata.output = output
-    assert os.path.exists(rtdata.output)
-
     rtdata.get_truth("truth/test_nircam_image_stages/" + output)
 
     diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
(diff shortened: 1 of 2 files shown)


jwst/regtest/test_nircam_image.py
```diff
@@ -0,0 +1,115 @@
...
+        "--steps.dark_current.save_results=True",
+        "--steps.jump.save_results=True",
+        "--steps.jump.rejection_threshold=50.0",
+        ]
+    Step.from_cmdline(args)
+
+    # And run image2 pipeline on the produced _rate file
+    rtdata.input = "jw42424001001_01101_00001_nrca5_rate.fits"
+    args = ["config/calwebb_image2.cfg", rtdata.input,
+        "--steps.assign_wcs.save_results=True",
```
hbushouse: Could also save the ``flat_field`` step results, since that's before ``photom`` (and hence ahead of the _cal file). And actually, since all ``assign_wcs`` does is attach the gwcs object, without modifying pixel values in any way, testing the ``flat_field`` output would essentially check both ``assign_wcs`` and ``flat_field`` (2 for 1).

jwst/regtest/test_nircam_image.py
```diff
@@ -0,0 +1,115 @@
...
+        # "--steps.skymatch.save_results=True",
+        ]
+    Step.from_cmdline(args)
+
+    return rtdata
+
+
+@pytest.mark.bigdata
+@pytest.mark.parametrize("suffix", ["dq_init", "saturation", "superbias",
+    "refpix", "linearity", "trapsfilled", "dark_current", "jump", "rate",
```
hbushouse: If the data have NINTS>1 you could also include the _rateints file in the comparison. Even though it isn't used downstream, it's still useful to test as another output of the ``ramp_fit`` step.

jwst/regtest/test_nircam_image.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+@pytest.mark.bigdata
+def test_nircam_image_stage3_i2d(run_pipelines, fitsdiff_default_kwargs):
+    rtdata = run_pipelines
+    rtdata.input = "jw42424-o002_20191220t214154_image3_001_asn.json"
+    rtdata.output = "jw42424-o002_t001_nircam_clear-f444w_i2d.fits"
+    rtdata.get_truth("truth/test_nircam_image_stages/jw42424-o002_t001_nircam_clear-f444w_i2d.fits")
+
+    fitsdiff_default_kwargs['ignore_fields'] = ['date', 'filename']
```
hbushouse: Isn't the normal default list bigger than this? And instead of repopulating the list from scratch, is there a way that you can simply append "naxis1" and "tform*" to the existing list?

jwst/regtest/test_nircam_image.py
```diff
@@ -0,0 +1,115 @@
...
+    fitsdiff_default_kwargs['ignore_keywords'] += ['naxis1', 'tform*']
+    diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
+    assert diff.identical, diff.report()
+
+
+@pytest.mark.bigdata
+def test_nircam_image_stage3_catalog(run_pipelines):
+    rtdata = run_pipelines
```
hbushouse: So are all the pipelines actually being run twice: once in "test_nircam_image_stage3_id" and again here in "test_nircam_image_stage3_catalog"? Why can't the comparisons of the 2 different types of outputs (i2d and cat) be combined into a single test module? Neither of them is parameterized, so I would think that the straightforward pieces of code that do the comparisons could be merged together. Or do you want them split up so that they show up as separate failures (in the case where one or both of them fail their assert)?

jdavies-st: Yes, I want there to be a single file that needs to be OKified if a test fails.  And since the comparisons are not similar, better to break them up into separate tests.

jdavies-st: But now, the pipeline is just being run once.  Just separate tests that then test aspects of the pipeline.

jwst/regtest/test_nircam_image.py
jdavies-st: There is currently no `ignore_fields` in the default kwargs, but we could add it if needed.  It only becomes an issue here were we are comparing columns in the HDRTAB extension.

hbushouse:
Can we get the 2 fairly simple requests for changes added in and get this merged?

jwst/regtest/test_nircam_image.py
jdavies-st: NINTS=1 for this dataset.

jdavies-st:
I updated this PR
 - also test the WCS object after `assign_wcs` is run in stage 2 pipeline
 - add `ignore_fields` to the `fitsdiff_default_kwargs` fixture so that we don't have these problems comparing FITS files with HDRTAB extensions.

The whole test session takes 7m20s to run on my laptop.

hbushouse:
One little flake8 issue that needs to be fixed yet before this can be merged.

PR #4386, opened 2019-12-23
JP-1189: NIRSpec FS spec2 regtests

hbushouse:
This test module applies the calspec2 pipeline to 3 different flavors of NIRSpec fixed-slit exposures:

1. A full-frame exposure, so all slits will be extracted and processed
2. An ALLSLITS subarray exposure, so all slits will be extracted and processed
3. A S400A1 subarray exposure on the NRS2 detector, so only 1 slit is processed

The test compares all 8 flavors of intermediate and final output products created by each pipeline run.

### DIFF
--- a/jwst/regtest/test_nirspec_fs_spec2.py
+++ b/jwst/regtest/test_nirspec_fs_spec2.py
@@ -1,0 +1,58 @@
+import os
+
+import pytest
+from astropy.io.fits.diff import FITSDiff
+
+from jwst.lib.suffix import replace_suffix
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+from jwst.stpipe import Step
+
+file_roots = ['jw00023001001_01101_00001_nrs1_', 'jw93045010001_02101_00001_nrs2_', 'jwtest1013001_01101_00001_nrs1_']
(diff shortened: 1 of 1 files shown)


jwst/regtest/test_nirspec_fs_spec2.py
```diff
@@ -0,0 +1,58 @@
...
+
+import pytest
+from astropy.io.fits.diff import FITSDiff
+
+from jwst.lib.suffix import replace_suffix
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+from jwst.stpipe import Step
+
+file_roots = ['jw00023001001_01101_00001_nrs1_', 'jw93045010001_02101_00001_nrs2_', 'jwtest1013001_01101_00001_nrs1_']
+@pytest.fixture(scope="module", params=file_roots, ids=file_roots)
```
stscieisenhamer: I have learned that the specification of `ids` at this point is actually redundant; the parametrization accomplishes the same thing.

hbushouse: Good to know. I just copy your code, under the assumption that you know waaaay more than I do about how stuff works. ;-)

PR #4394, opened 2019-12-24
JP-1206: NIRCam TSImg tso2 and tso3 regtests

### LINKED ISSUE #4391: REGTEST: Write new NIRCam TS Image level2b regression test
Issue [JP-1206](https://jira.stsci.edu/browse/JP-1206) was created by Howard Bushouse:

Create a regression test in the new regtest system for applying the "calwebb_tso-image2"  pipeline to a NIRCam Time Series Image (EXP_TYPE='NRC_TSIMAGE') exposure. Could also include a subsequent test of the same exposure processed with the level-3 "calwebb_tso3" pipeline.

There is an existing level-3 test of this mode in the old regtest system, but nothing for level-2b (because the only simulated data that we have is the equivalent of a _cal product). Could use files with this exposure mode from the DMS test suite, but of course the actual image data in those is garbage.

hbushouse:
New regression test that runs the ``calwebb_tso-image2`` and ``calwebb_tso3`` pipelines on a set of NIRCam TS imaging data. The input data are from a simulation from INS of a point source in the SUB64P subarray. I copied the simulated data into 2 uncal files taken from the DMS test suite that also correspond to TSIMG mode on the b1 detector (B module, short-wave detector 1). Processed the uncal's through ``calwebb_tso1``, and the resulting _rateints products are used as input to the ``calwebb_tso-image2`` test. The resulting _calints products are used as input to the ``calwebb_tso3`` test. Having 2 exposures (even though the data are identical) tests the capability of the level-3 routines to handle multiple exposures. The only level-3 product for TSO imaging mode is the photometry catalog (_phot.ecsv), which is compared by the test.

Resolves #4391 and https://jira.stsci.edu/browse/JP-1206

### DIFF
--- a/jwst/regtest/test_nircam_tsimg.py
+++ b/jwst/regtest/test_nircam_tsimg.py
@@ -1,0 +1,62 @@
+import pytest
+from astropy.io.fits.diff import FITSDiff
+from astropy.table import Table, setdiff
+
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+from jwst.stpipe import Step
+
+
+@pytest.fixture(scope="module")
+def run_pipelines(jail, rtdata_module):
+    """Run stage 2 and 3 pipelines on NIRCam TSIMG data."""
+
+    rtdata = rtdata_module
+
+    collect_pipeline_cfgs("config")
+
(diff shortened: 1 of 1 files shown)


hbushouse:
Input and truth files have been uploaded to artifactory and a local run of the test shows that they all pass.

PR #4395, opened 2019-12-24
Jp 1199 Implement NIRSpec IFU Level2b regression

### LINKED ISSUE #4384: REGTEST: Write new NIRSpec IFU level2b regression test
Issue [JP-1199](https://jira.stsci.edu/browse/JP-1199) was created by Howard Bushouse:

Create a regression test in the new regtest system for applying the ``calwebb_spec2`` (``Spec2Pipeline``) pipeline to a NIRSpec IFU exposure, from either detector NRS1 or NRS2.

stscieisenhamer:
Resolves #4384/[JP-1199](https://jira.stsci.edu/browse/JP-1199).


stscieisenhamer:
@jdavies-st I'm beginning to generalize the pipeline testing. When possible, could you look at that aspect of testing?

jwst/regtest/test_niriss.py
```diff
@@ -2,68 +2,44 @@
...
-    rtdata.get_truth(os.path.join(truth_path, output))
-
-    diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
-    assert diff.identical, diff.report()
+from . import regtestdata as rt
 
 
 @pytest.fixture(scope='module')
-def run_nrs_wfss_spectral(jail, rtdata_module):
+def run_nis_wfss_spectral(jail, rtdata_module):
```
hbushouse: Wouldn't hurt to just spell out the whole instrument name (i.e. "niriss"), which also avoids all potential for confusion.

jwst/regtest/test_nirspec.py
```diff
@@ -0,0 +1,46 @@
...
+
+from . import regtestdata as rt
+
+
+@pytest.fixture(scope='module')
+def run_nrs_ifu(jail, rtdata_module):
+    """Run the pipelines"""
+    step_params = {
+        'input_path': 'nirspec/spectroscopic/nrs_ifu_uncal.fits',
+        'step': 'calwebb_detector1.cfg',
```
hbushouse: The request contained in JP-1199 was to test the ``calwebb_spec2`` pipeline on NIRSpec IFU data, not the ``calwebb_detector1`` pipeline.

stscieisenhamer: At the time, did not see any nirspec-related detector1 tests, so was going to cover both. However, if it is covered, I can pre-build the Stage 1 files.

hbushouse: We have ``regtest/test_nirspec_subarray.py``, which runs ``calwebb_detector1`` on a NIRSpec subarray exposure. So we do have at least one detector1 test on NIRSpec data, although it's subarray (intentionally so, in order to test the ``gain_scale`` step) and https://jira.stsci.edu/browse/JP-1169 is in progress to create a detector1 test on a full-frame NIRSpec exposure using the IRS2 readout pattern. So once the latter gets finished, we should have decent coverage for NIRSpec detector1.

jwst/regtest/test_niriss.py
```diff
@@ -2,68 +2,44 @@
...
-        '--steps.photom.save_results=true',
-        '--steps.resample_spec.save_results=true',
-        '--steps.cube_build.save_results=true',
-        '--steps.extract_1d.save_results=true',
-    ]
-    Step.from_cmdline(args)
-
-    return rtdata
+    step_params = {
+        'input_path': 'niriss/level2a/nir_wfss_spec2_asn.json',
```
hbushouse: The convention we've been trying to follow lately for the organization of input directories is to just have the data segregated by instrument and mode, without anything in terms of what's being tested. So a suitable input directory for this would be 'niriss/wfss/'.

jwst/regtest/test_nirspec_ifu_spec2.py
```diff
@@ -0,0 +1,45 @@
...
+import pytest
+
+from . import regtestdata as rt
+
+
+@pytest.fixture(scope='module')
+def run_nrs_ifu(jail, rtdata_module):
+    """Run the pipelines"""
+    step_params = {
+        'input_path': 'nirspec/spectroscopic/ifu/nrs_ifu_rate.fits',
```
hbushouse: I think we can do away with the very generic "spectroscopic" subdirectory under nirspec, so that we don't then have to go down yet another directory level for each specific spectroscopic mode. I would suggest just using 'nirspec/ifu/'.

stscieisenhamer: There are other things there that will need be dealt with, but that should probably be under its own issue.

PR #4396, opened 2019-12-26
JP-1205: NIRISS SOSS level-2b and level-3 regtests

### LINKED ISSUE #4392: REGTEST: Write new NIRISS SOSS level2b regression test
Issue [JP-1205](https://jira.stsci.edu/browse/JP-1205) was created by Howard Bushouse:

Create a regression test in the new regtest system for applying the "calwebb_tso-spec2"  pipeline to a NIRISS SOSS (EXP_TYPE='NIS_SOSS') exposure. Could also potentially include a subsequent test of the same exposure processed with the level-3 "calwebb_tso3" pipeline.

There are existing level-2b and level-3 tests of this exposure mode in the old regtest system, so those tests (and data files) could just be ported over to the new system.

hbushouse:
This provides ``calwebb_tso-spec2`` and ``calwebb_tso3`` regression tests for NIRISS SOSS mode data. It uses 2 segmented exposures, both of which are processed through ``calwebb_tso-spec2``, but only the results of the first one are checked by the test. Both resulting _calints products are then used as input to a ``calwebb_tso3`` test.

The data were built from some SOSS simulations we obtained a couple years ago from the NIRISS team, transplanting the data arrays into the results of a recent DMS run of the same data mode. It uses only 10 integrations per segment to keep run time reasonable and the actual data arrays are identical in both segment files. Only the INT_TIMES table in each segment makes it look like different integrations.

Input and truth files have been uploaded to artifactory and I've gotten a clean local run of the test.

Resolves #4392 and https://jira.stsci.edu/browse/JP-1205

### DIFF
--- a/jwst/regtest/test_niriss_soss.py
+++ b/jwst/regtest/test_niriss_soss.py
@@ -1,0 +1,76 @@
+import pytest
+from astropy.io.fits.diff import FITSDiff
+from astropy.table import Table, setdiff
+
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+from jwst.stpipe import Step
+
+
+@pytest.fixture(scope="module")
+def run_pipelines(jail, rtdata_module):
+    """Run stage 2-3 pipelines on NIRISS SOSS data."""
+    rtdata = rtdata_module
+    collect_pipeline_cfgs("config")
+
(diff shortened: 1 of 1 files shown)


hbushouse:
Note that the tso3 asn file also contains a target-acq exposure as a member, which is intended to test the ability of the ``calwebb_tso3`` pipeline to correctly ignore all non-science members.

PR #4397, opened 2019-12-27
JP-1204: NIRCam TSGRISM regression test

### LINKED ISSUE #4393: REGTEST: Write new NIRCam TS Grism level2b regression test
Issue [JP-1204](https://jira.stsci.edu/browse/JP-1204) was created by Howard Bushouse:

Create a regression test in the new regtest system for applying the "calwebb_tso-spec2"  pipeline to a NIRCam Time Series Grism (EXP_TYPE='NRC_TSGRISM') exposure. Could also potentially include a subsequent test of the same exposure processed with the level-3 "calwebb_tso3" pipeline.

Doesn't appear to be a level2b test of this mode in the old regression test system, probably due to the lack of appropriate test data (INS only supplied simulated _cal images for testing level-3 processing). Could potentially find suitable file(s) from the DMS test suites, although the actual image data in them are likely to be garbage.

hbushouse:
New regression test that performs ``calwebb_tso-spec2`` and ``calwebb_tso3`` processing on a NIRCam TSO grism (NRC_TSGRISM) exposure. It checks intermediate products from ``calwebb_tso-spec2``, as well as all final products from both pipelines.

The data were assembled from a recent DMS dataset that uses the same instrument setup, and imbedding simulated data into the SCI array. The assembled test data start at the _uncal level.

Note that this PR also includes a critical fix to the NRC_TSGRISM portion of the ``assign_wcs`` step. It had not yet been updated to switch from using the CRPIX1/CRPIX2 keyword values to determine the source location, to using the new XREF_SCI/YREF_SCI keywords instead. This was causing the wavelength array calculations to come out incorrectly, which subsequently lead to problems downstream. This has now been fixed.

All input/truth files needed by the test have been uploaded to artifactory and a local run of the test confirms success.

Resolves #4393 and https://jira.stsci.edu/browse/JP-1204

### DIFF
--- a/jwst/assign_wcs/nircam.py
+++ b/jwst/assign_wcs/nircam.py
@@ -190,13 +190,14 @@ def tsgrism(input_model, reference_files):
     # input into the forward transform is x,y,x0,y0,order
     # where x,y is the pixel location in the grism image
     # and x0,y0 is the source location in the "direct" image
-    # For this mode, the source is always at crpix1,crpis2
-    # discussion with nadia that wcsinfo might not be available
+    # For this mode, the source is always at crpix1 x crpix2, which
+    # are stored in keywords XREF_SCI, YREF_SCI.
+    # Discussion with nadia that wcsinfo might not be available
     # here but crpix info could be in wcs.source_location or similar
     # TSGRISM mode places the sources at crpix, and all subarrays
     # begin at 0,0, so no need to translate the crpix to full frame
     # because they already are in full frame coordinates.
-    xc, yc = (input_model.meta.wcsinfo.crpix1, input_model.meta.wcsinfo.crpix2)
+    xc, yc = (input_model.meta.wcsinfo.siaf_xref_sci, input_model.meta.wcsinfo.siaf_yref_sci)
     xcenter = Const1D(xc)
     xcenter.inverse = Const1D(xc)
     ycenter = Const1D(yc)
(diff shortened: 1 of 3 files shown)


jwst/assign_wcs/nircam.py
```diff
@@ -190,13 +190,14 @@ def tsgrism(input_model, reference_files):
...
-    # discussion with nadia that wcsinfo might not be available
+    # For this mode, the source is always at crpix1 x crpix2, which
+    # are stored in keywords XREF_SCI, YREF_SCI.
+    # Discussion with nadia that wcsinfo might not be available
     # here but crpix info could be in wcs.source_location or similar
     # TSGRISM mode places the sources at crpix, and all subarrays
     # begin at 0,0, so no need to translate the crpix to full frame
     # because they already are in full frame coordinates.
-    xc, yc = (input_model.meta.wcsinfo.crpix1, input_model.meta.wcsinfo.crpix2)
+    xc, yc = (input_model.meta.wcsinfo.siaf_xref_sci, input_model.meta.wcsinfo.siaf_yref_sci)
```
jdavies-st: Will this cause any old regression tests to fail?

hbushouse: It resulted in one unit test that needed to be updated (to also switch from crpixn to x/yref_sci), but not sure about regression tests. According to JP-811, we don't have any existing tests for NIRCam TS-GRISM mode (just NIRCam TS-Image). So I'm guessing we won't see any failures.

PR #4401, opened 2019-12-27
Set source_catalog.snr_threshold=20 for MIRI/NIRCam image3 regtests

jdavies-st:
This fixes the `test_miri_image_stage3_catalog` test that is currently failing

https://plwishmaster.stsci.edu:8081/job/RT/job/JWST_RT_REFACTOR/93/testReport/junit/jwst.regtest/test_miri_image/_stable_deps__test_miri_image_stage3_catalog/

 in the new regression test system because it is digging down into the noise of the maskbar to find sources and coming up with `NaN` for x and y centroids.  And of course `nan != nan` so the comparison fails.

The solution is to set the source catalog detection threshold higher to only detect the 5 simulated stars in the field.  It still detects a 6th non-star, but at least this blob has a centroid.

The truth file has already been updated on Artifactory.

```
$ pytest --bigdata --env=newdev jwst/regtest/ -k test_miri_image_stage3_catalog
============================= test session starts =============================
platform darwin -- Python 3.7.5, pytest-5.3.1, py-1.8.0, pluggy-0.13.1
plugins: asdf-2.4.3a1.dev13+gc1373db, xdist-1.30.0, requests-mock-1.7.0, openfiles-0.4.0, ci-watson-0.4, forked-1.1.3, doctestplus-0.5.0, cov-2.8.1
collected 140 items / 139 deselected / 1 selected                                                                                      

jwst/regtest/test_miri_image.py .                                         [100%]
========== 1 passed, 139 deselected, 38 warnings in 115.50s (0:01:55) ==========
```

### DIFF
--- a/jwst/regtest/test_miri_image.py
+++ b/jwst/regtest/test_miri_image.py
@@ -55,9 +55,10 @@ def run_pipelines(jail, rtdata_module):
     # image3 pipeline on all _cal files listed in association
     rtdata.get_data("miri/image/det_dithered_5stars_image3_asn.json")
     args = ["config/calwebb_image3.cfg", rtdata.input,
-        # Set some unique tweakreg param values needed for these data
+        # Set some unique param values needed for this data
         "--steps.tweakreg.snr_threshold=200",
-        "--steps.tweakreg.use2dhist=False"
+        "--steps.tweakreg.use2dhist=False",
+        "--steps.source_catalog.snr_threshold=20",
         ]
     Step.from_cmdline(args)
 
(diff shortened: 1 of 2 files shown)


jwst/regtest/test_miri_image.py
```diff
@@ -55,9 +55,10 @@ def run_pipelines(jail, rtdata_module):
     # image3 pipeline on all _cal files listed in association
     rtdata.get_data("miri/image/det_dithered_5stars_image3_asn.json")
     args = ["config/calwebb_image3.cfg", rtdata.input,
-        # Set some unique tweakreg param values needed for these data
+        # Set some unique param values needed for this data
```
hbushouse: I beg to differ: "data" is plural (as opposed to, say, a data file), hence "these" instead of "this".

<#grammar-police>

jdavies-st: Not sure what your agendum is here Howard, but I will change it back.  😁

jdavies-st:
The NIRCam imaging regtest is failing for the same reason, so also setting source catalog step to have high SNR detection threshold as well.

PR #4408, opened 2019-12-30
JP-1200: REGTEST for NIRSpec brightobj

### LINKED ISSUE #4383: REGTEST: Write new NIRSpec Bright-Object level2b regression test
Issue [JP-1200](https://jira.stsci.edu/browse/JP-1200) was created by Howard Bushouse:

Create a regression test in the new regtest system for applying the ``calwebb_spec2`` (``Spec2Pipeline``) pipeline to a NIRSpec IFU exposure, from either detector NRS1 or NRS2.

stscicrawford:
Added a test for nirspec bright obj:
 Regression test of calwebb_spec2 pipeline performed on NIRSpec fixed-slit data that uses the NRS_BRIGHTOBJ mode (S1600A1 slit).

Closes #4383

### DIFF
--- a/jwst/regtest/test_nirspec_brightobj.py
+++ b/jwst/regtest/test_nirspec_brightobj.py
@@ -18,7 +18,7 @@ def run_tso_spec2_pipeline(jail, rtdata_module, request):
     # Get the input exposure
     rtdata.get_data('nirspec/tso/jw84600042001_02101_00001_nrs2_rateints.fits')
 
-    # Run the calwebb_spec2 pipeline; 
+    # Run the calwebb_spec2 pipeline;
     collect_pipeline_cfgs("config")
     args = ["config/calwebb_tso-spec2.cfg", rtdata.input]
     Step.from_cmdline(args)
--- a/jwst/regtest/test_nirspec_brightobj.py
+++ b/jwst/regtest/test_nirspec_brightobj.py
@@ -1,0 +1,46 @@
+import os
+import pytest
+
+from astropy.io.fits.diff import FITSDiff
+
+from jwst.lib.suffix import replace_suffix
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+from jwst.stpipe import Step
+
+@pytest.fixture(scope="module")
+def run_tso_spec2_pipeline(jail, rtdata_module, request):
+    """Run the calwebb_spec2 pipeline performed on NIRSpec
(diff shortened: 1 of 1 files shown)


stscicrawford:
When I ran this locally, the only two outputs that were created were the ones that were given.  In the config file, it has `save_result` for `resample_spec`, `cube_build`, and `extract_1d`, and as this data set is a single long slit, I don't think it would build a cube.   However, if this is incorrect, then please let me know to add the relevant file.

philhodge:
`resample_spec` shouldn't be called for TSO data, so that leaves just the output of `extract_1d`.  There are some other steps for which you could check the output, and you can specify in the test that the output from those steps should be save by appending command-line arguments such as "--steps.photom.save_results=True".

The Travis failure seems to be due to flake8, trailing whitespace.

stscieisenhamer:
Adding the `save_results` to the assign_wcs, extract_2d, and flat_field steps should also produce output, unless there is some hidden logic that skips those steps. The photom step technically doesn't need an explicit one, since that is the `cal` product itself.

stscicrawford:
@stscieisenhamer -- the photom step did add an ouptut file when I had `save_results=True`.   I added it along with the results from assign_wcs, flatfield, and extract_2d.   So I think this should be ready to be merged now.

jdavies-st:
The fact that `--steps.photom.save_results` outputs an `_photom.fits` file which is indentical to but differently-named from the `_cal.fits` file is probably an indicator that we should rewrite the spec2 pipeline to use something like:
```
self.photom.save_results = True
self.photom.suffix = "cal"
```
or similar when saving the cal file.

stscieisenhamer:
This would be a general architecture/convention issue: The current convention is that whatever is the top-level step, in this case `Spec2Pipeline`, when run from `strun`, gets `save_results=True`. So, for the suggestion of making the substep `photom` be the one to save the final result, `Spec2Pipeline` would need `save_results=False` hardwired in there. Hence the current setup.  IMHO the current setup is current-convention/architecture "cleaner". But this would be a later discussion.

Concerning testing and wondering whether the test should save the `photom` step result and compare them, since it is really the `cal` file anyways, there is reasons for doing so:

- When writing tests, one can easily just go through all the substeps and request the results saved without having to have inherent knowledge about the particular pipeline. No harm done.
- If both are saved, the current test would be resilient to having something later in the pipeline modifying its output, but the test will still check how `photom` did anyways.
- If one really wanted to ensure that both are the same, then the extra test can be added to ensure equality.

PR #4409, opened 2019-12-30
JP-1152 Check whether the output pixel number is NaN

philhodge:
For each input spectrum, the array of corresponding pixel numbers in the output, combined spectrum is computed by using the inverse WCS function.  It is possible for some of these pixel numbers (probably only at an endpoint of the input spectrum) to be NaN due to the finite bounding box for the WCS.  A check for NaN was added.

While testing this change, another bug was uncovered.  In the section that tests for the possibility that there was no input data for some output pixel, the error estimate for flux was being assigned to the error estimate for surface brightness.  This has been fixed.

### DIFF
--- a/jwst/combine_1d/combine1d.py
+++ b/jwst/combine_1d/combine1d.py
@@ -187,6 +187,7 @@ class OutputSpectrumModel:
         self.weight = np.zeros(nelem, dtype=np.float64)
         self.count = np.zeros(nelem, dtype=np.float64)
 
+        n_nan = 0                                       # initial value
         for in_spec in input_spectra:
             # Get the pixel numbers in the output corresponding to the
             # wavelengths of the current input spectrum.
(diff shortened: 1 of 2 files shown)


jwst/combine_1d/combine1d.py
```diff
@@ -199,7 +200,12 @@ def accumulate_sums(self, input_spectra):
                 if in_spec.dq[i] & datamodels.dqflags.pixel['DO_NOT_USE'] > 0:
                     continue
                 # Round to the nearest pixel.
+                if np.isnan(out_pixel[i]):
```
hbushouse: Is there a way to do the search for NaN on the entire out_pixel array at once (i.e. outside of this loop over i), rather than having to do the test on every loop iteration? Or would waiting to do the check after the loop is done create problems for the calculations being done in the remainder of the loop? If it's not a big performance hit, just leave it as is.

philhodge: Yes, `np.isnan` can operate on an array and return a boolean array, but I would still need to check an element of that boolean array inside the loop.  I thought the check on NaN would be quick, but checking a boolean value has to be quicker.  I'll take the call to `np.isnan` out of the loop.

PR #4413, opened 2019-12-31
Updated dependencies to released versions of astropy, asdf and gwcs

jdavies-st:
Set the following minimum versions in `setup.py`:
 - `asdf>=2.5`
 - `astropy>=4.0`
 - `gwcs>=0.12`

So now we no longer have dev dependencies.

Also zero out the requirements in `requirements-sdp.txt` so that our regression testing will get the latests releases of everything.  We can pin these again as we get into our next dev cycle.


nden:
Just to confirm:  after renaming JenkinsRT_Unstable-->JenkinsRT_dev the Jenkins job was not changed. It still runs on the `move_transforms_out_of_jwst` branch, right?

jdavies-st:
It hasn't been running on that branch for a while.  It has been running on `master`.  I'm not sure when it was switched.

Let's switch it to `move_transforms_out_of_jwst`.  Can you rebase that branch so it picks up this latest change to the file name?

PR #4414, opened 2019-12-31
Jp 1169 NIRSpec MOS level2a regression test for IRS2 mode data

ddavis-stsci:
Test for the detector1 pipeline using NIRSpec data in IRS2 mode. This takes an uncal file and generates the stage 1 FITS files (rate) along with the intermediate products for testing.


jwst/regtest/test_NIRSpec_IRS2_detector1.py
```diff
@@ -0,0 +1,86 @@
...
+from astropy.io.fits.diff import FITSDiff
+from jwst.stpipe import Step
+
+@pytest.fixture(scope="module")
+def run_pipeline(rtdata_module, jail):
+    """Run calwebb_Detector1 pipeline on NIRSpec uncal data. This test is for
+       data taken with the IRS2 mode. The steps are similar to the NIRCam level
+       2a test but the IRS2 mode tests different paths in the same modules."""
+    rtdata = rtdata_module
+    rtdata.get_data("nirspec/test_detector1Pipeline/jw0010010_11010_nrs1_chimera_uncal.fits")
```
philhodge: The subdirectory after "nirspec/" is supposed to be the mode of the data.  I don't know exactly what "mode" values we should use, though.

jwst/regtest/test_NIRSpec_IRS2_detector1.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Step.from_cmdline(["jwst.pipeline.Detector1Pipeline", rtdata.input,
+                       "--save_calibrated_ramp=True",
+                       "--steps.group_scale.save_results=True",
+                       "--steps.dq_init.save_results=True",
+                       "--steps.saturation.save_results=True",
+                       "--steps.ipc.save_results=True",
```
philhodge: The `ipc` step is skipped, so there is no need to save the results.  Also, the `rscd` step is only used for MIRI data.

jwst/regtest/test_NIRSpec_IRS2_detector1.py
```diff
@@ -0,0 +1,86 @@
...
+@pytest.fixture(scope="module")
+def run_pipeline(rtdata_module, jail):
+    """Run calwebb_Detector1 pipeline on NIRSpec uncal data. This test is for
+       data taken with the IRS2 mode. The steps are similar to the NIRCam level
+       2a test but the IRS2 mode tests different paths in the same modules."""
+    rtdata = rtdata_module
+    rtdata.get_data("nirspec/test_detector1Pipeline/jw0010010_11010_nrs1_chimera_uncal.fits")
+
+    Step.from_cmdline(["jwst.pipeline.Detector1Pipeline", rtdata.input,
+                       "--save_calibrated_ramp=True",
```
philhodge: You don't need to save the calibrated ramp if you save the output from the `jump` step, because they would be the same.

jwst/regtest/test_NIRSpec_IRS2_detector1.py
```diff
@@ -0,0 +1,86 @@
...
+                       "--steps.ramp_fit.save_results=True",
+                       "--steps.gain_scale.save_results=True",
+                       "--steps.jump.rejection_threshold=20.0"])
+
+    return rtdata
+
+@pytest.mark.bigdata
+def test_NIRSpec_det1_completion(run_pipeline):
+    """ Check that the final and intermediate files are generated by the pipeline."""
+    files = glob('*_rate.fits')
```
philhodge: You could omit _rate and _ipc.

jwst/regtest/test_NIRSpec_IRS2_detector1.py
```diff
@@ -0,0 +1,86 @@
...
+    files += glob('*_saturation.fits')
+    files += glob('*_superbias.fits')
+    files += glob('*_0_ramp_fit.fits')
+    # There should be 13 outputs
+    assert len(files) == 13
+
+
+@pytest.mark.bigdata
+@pytest.mark.parametrize("output", [
+    'jw0010010_11010_nrs1_chimera_rate.fits',
```
philhodge: You could omit _rate and _ipc.

jwst/regtest/test_NIRSpec_IRS2_detector1.py
```diff
@@ -0,0 +1,86 @@
...
+                         ids=['rate','ramp','dark_current','dq_init','gain_scale',
+                              'group_scale','ipc','jump','linearity','refpix',
+                              'saturation','superbias','0_ramp_fit'])
+def test_NIRSpec_det1Pipeline(run_pipeline, fitsdiff_default_kwargs, output):
+    """
+    Regression test of calwebb_Detector1 pipeline performed on NIRSpec data.
+    """
+    rtdata = run_pipeline
+    rtdata.output = output
+    rtdata.get_truth(os.path.join("truth/nirspec/test_detector1Pipeline", output))
```
philhodge: The path should be "truth/test_nirspec_irs2_detector1/", though I'm not sure whether the directory below "truth/" should be lower case or match the case of the test name.

jwst/regtest/test_NIRSpec_IRS2_detector1.py
```diff
@@ -0,0 +1,86 @@
...
+                       "--steps.dark_current.save_results=True",
+                       "--steps.jump.save_results=True",
+                       "--steps.ramp_fit.save_results=True",
+                       "--steps.gain_scale.save_results=True",
+                       "--steps.jump.rejection_threshold=20.0"])
+
+    return rtdata
+
+@pytest.mark.bigdata
+def test_NIRSpec_det1_completion(run_pipeline):
```
jdavies-st: This completion test is unnecessary as you test for the exact same thing in the test below.

jwst/regtest/test_NIRSpec_IRS2_detector1.py
jdavies-st: Test names should be lowercase as it is a test function.  Functions in python are always lower case with underscores.  Only classes get CamelCase, but without underscores.  Never mix.  Same with this test filename - should be lower case.

jdavies-st:
Also, when I run this test locally, it gives a validation warning on the input data.
```
jwst/regtest/test_NIRSpec_IRS2_detector1.py::test_NIRSpec_det1Pipeline[rate]
  /Users/jdavies/dev/jwst/jwst/datamodels/validate.py:35: ValidationWarning: While validating meta.exposure.readpatt the following error occurred:
  'ALLIRS2' is not one of ['ACQ1', 'ACQ2', 'BRIGHT1', 'BRIGHT2', 'DEEP2', 'DEEP8', 'FAST', 'FASTGRPAVG', 'FGS', 'FGS60', 'FGS8370', 'FGS840', 'FGSRAPID', 'FINEGUIDE', 'ID', 'MEDIUM2', 'MEDIUM8', 'NIS', 'NISRAPID', 'NRS', 'NRSIRS2', 'NRSN16R4', 'NRSN32R8', 'NRSN8R2', 'NRSRAPID', 'NRSIRS2RAPID', 'NRSRAPIDD1', 'NRSRAPIDD2', 'NRSRAPIDD6', 'NRSSLOW', 'RAPID', 'SHALLOW2', 'SHALLOW4', 'SLOW', 'TRACK', 'ANY', 'N/A']
  
  Failed validating 'enum' in schema:
      {'$schema': 'http://stsci.edu/schemas/asdf-schema/0.1.0/asdf-schema',
       'enum': ['ACQ1',
                'ACQ2',
                'BRIGHT1',
                'BRIGHT2',
                'DEEP2',
                'DEEP8',
                'FAST',
                'FASTGRPAVG',
                'FGS',
                'FGS60',
                'FGS8370',
                'FGS840',
                'FGSRAPID',
                'FINEGUIDE',
                'ID',
                'MEDIUM2',
                'MEDIUM8',
                'NIS',
                'NISRAPID',
                'NRS',
                'NRSIRS2',
                'NRSN16R4',
                'NRSN32R8',
                'NRSN8R2',
                'NRSRAPID',
                'NRSIRS2RAPID',
                'NRSRAPIDD1',
                'NRSRAPIDD2',
                'NRSRAPIDD6',
                'NRSSLOW',
                'RAPID',
                'SHALLOW2',
                'SHALLOW4',
                'SLOW',
                'TRACK',
                'ANY',
                'N/A'],
       'fits_keyword': 'READPATT',
       'title': 'Readout pattern',
       'type': 'string'}
  
  On instance:
      'ALLIRS2'
    warnings.warn(errmsg, ValidationWarning)
```
It looks like the input data does not have a valid `meta.exposure.readpatt`.  Perhaps it should be set to `NRSIRS2`?

Finally, I'm getting some tests failing on several different pixels in the `fitsdiff` when I run locally on my MacOS laptop.  I'm wondering if there's some round-off errors somewhere that are affecting the results.  We are also seeing some of these issues in the coron tests and tests that use drizzle.  Hmm.

hbushouse:
It's possible that the schema validation warning could be coming from one of the NIRSpec ref files in CRDS. If the science data had that bad value, it wouldn't be allowed to write any of the output products.

PR #4417, opened 2020-01-03
JP-1210 Implement NIRSpec IFU Spec3Pipeline regression

### LINKED ISSUE #4405: REGTEST: Write new NIRSpec IFU level3 regression test
Issue [JP-1210](https://jira.stsci.edu/browse/JP-1210) was created by Howard Bushouse:

Create a regression test in the new regtest system for applying the "calwebb_spec3"  pipeline to an association of NIRSpec IFU exposures. Could potentially be combined with tests of the "calwebb_spec2" pipeline on NIRSpec IFU data (JP-1199). Should include data from both detectors (NRS1 and NRS2) for a given filter+grating combination, and ideally could include multiple filters with a given grating (data from different gratings do NOT get combined).

stscieisenhamer:
Resolves #4405/[JP-1210](https://jira.stsci.edu/browse/JP-1210)


stscieisenhamer:
IFU. I believe this one was on the order of 3.5 hours.

PR #4419, opened 2020-01-04
JP-1209 MIRI MRS Level3 regression

### LINKED ISSUE #4406: REGTEST: Write new MIRI MRS level3 regression test
Issue [JP-1209](https://jira.stsci.edu/browse/JP-1209) was created by Howard Bushouse:

Create a regression test in the new regtest system for applying the "calwebb_spec3"  pipeline to an association of MIRI MRS exposures. Could potentially be combined with tests of the "calwebb_spec2" pipeline on MIRI MRS data (JP-1198). Should at least include data from multiple sub-bands (short, medium, long) on either the IFUSHORT or IFULONG detector, and ideally would also include data from both detectors.

stscieisenhamer:
Resolves #4406/[JP-1209](https://jira.stsci.edu/browse/JP-1209)


stscieisenhamer:
Concering #1: The miri tests depends on code that was developed and waiting in PR #4417. That would should go through first.

Concerning #2: I think this is useful. It is to be used to re-run tests when the pipeline component takes a long time. This allows quick development of other aspects of a test. I will better document in the other PR.

jwst/regtest/test_miri_mrs.py
```diff
@@ -32,6 +32,25 @@ def run_spec2(jail, rtdata_module):
     return rt.run_step_from_dict(rtdata_module, **step_params)
 
 
+@pytest.fixture(scope='module')
+def run_spec3(jail, rtdata_module):
```
jdavies-st: This pipeline doesn't piggyback off any of the data in the spec2 pipeline?  I.e. is there any dataset that runs all the way through levels 2 and 3?

stscieisenhamer:
Yes, and I don't know what had changed because I created the data in the first place when running. There was one change to the association, plus a rebase. Will be investigating.

jwst/regtest/test_miri_mrs.py
stscieisenhamer: Will look into that.

stscieisenhamer:
@jdavies-st I've just done a rebase and ran the tests and am no longer getting the ValueError. If time permits, could you try again.

jwst/regtest/regtestdata.py
```diff
@@ -161,11 +161,27 @@ def get_truth(self, path=None, docopy=None):
...
+        ----------
+        path: str
+            The remote path
+
+        docopy : bool
+            Switch to control whether or not to copy a file
+            into the test output directory when running the test.
+            If you wish to open the file directly from remote
+            location or just to see path to source, set this to `False`.
+            Default: `True`
```
jdavies-st: It looks like the default is actually `docopy=None`.  Maybe it should be True.

jwst/regtest/regtestdata.py
```diff
@@ -212,6 +229,11 @@ def run_step_from_dict(rtdata, **step_params):
     step_params: dict
         The parameters defining what step to run with what input
 
+    Returns
+    -------
+    rtdata: RegistryData
```
jdavies-st: This should be a `RegtestData` class instance.

jwst/regtest/regtestdata.py
```diff
@@ -260,6 +286,11 @@ def run_step_from_dict_mock(rtdata, source,  **step_params):
     source: Path-like folder
         The folder to copy from. All regular files are copied.
 
+    Returns
+    -------
+    rtdata: RegistryData
+        Updated `RegistryData` object with inputs set.
```
jdavies-st: `RegtestData`

jwst/regtest/test_miri_mrs.py
```diff
@@ -1,14 +1,36 @@
...
 import pytest
 
+from jwst.associations import load_asn
+from jwst.lib.suffix import replace_suffix
+
 from . import regtestdata as rt
 
 
+# Define artifactory source and truth
+BIGDATA_PATH = 'miri/mrs'
```
jdavies-st: Could rename this path to something other than BIGDATA?  I fear confusion with TEST_BIGDATA.

jwst/regtest/test_miri_mrs.py
```diff
@@ -39,5 +107,49 @@ def run_spec2(jail, rtdata_module):
...
+        'ifushort_set2_1_mrs_imatch.fits',
+        'ifushort_set2_ch1-short_s3d.fits',
+        'ifushort_set2_ch1-short_x1d.fits',
+        'ifushort_set2_ch2-short_s3d.fits',
+        'ifushort_set2_ch2-short_x1d.fits',
+    ]
+)
+def test_spec3_multi(run_spec3_multi, fitsdiff_default_kwargs, output):
+    """Regression test matching output files"""
+    rt.is_like_truth(
```
jdavies-st: What does the `rt.is_like_truth` traceback look like when it fails?

Is there a good reason to have it in the `RegtestData` class as opposed to separate helper, i.e. a fixture or just an assertion helper?

stscieisenhamer: First note that this is not in `RegtestData` class, but in module `regtestdata.py`. As such, it is a helper to and a user of the `RegtestData` class, and hence why it is in the module. However, no strong feeling as to where.

jdavies-st: Doh!  My bad.  Carry on.  👍

jemorrison: How long does this test take ? I may be a good idea to test an IFULONG as well. The IFULONG is a different detector. The reference files for the IFULONG is a different set of columns. It would be nice to have an end-to-end (spec2 and spec3) test of the IFULONG.

stscieisenhamer: Not sure if end-to-end of IFULONG is necessary. However, would it affect how the "full spec3" test? I believe that would be included in the same association. If so, that would be worthwhile testing.

And, about run time, this is fairly short, I believe on the order of 20 minutes.

jemorrison: I am curious why did you write the RegTestData class rather than just use pytest.fixtures. This regression test seems to be set up differently than the other regression tests. Is it because you wanted to pass the results of the spec2 pipeline to the spec3 pipeline ?

stscieisenhamer: Yes, done that way for the chaining. Note however, it is still a `RegtestData` object, just the one returned from `run_spec2` instead of a clean one.  For chaining, this allows isolation and independent testing of the individual pipelines while making it clear that the `run_spec3` fixture is dependent on the result of the `run_spec2` fixture.

Issue #4246, opened 2019-11-11
ASDF extension in MIRI Team looking for schemas in wrong location


stscijgbot:
Issue [JP-1132](https://jira.stsci.edu/browse/JP-1132) was created by Nadia Dencheva:

Nadia and Howard,

Compiling the notes for the MIRI software meeting has reminded me of a JWST data models issue I forgot to report to you. The problem is described in Bugzilla ticket 596

http://www.miricle.org/bugzilla/show_bug.cgi?id=596

but here is a summary.

Back in July Nadia helped me to get the MIRI data models working with the jwst v0.13.8 release by making a following change within the data model base class, MiriDataModel.

The code within the constructor which defined where to find the data model schemas was changed from:

```
      if extensions is None:
           extensions = [MIRIExtension()]
```

to

```
        schema_path = os.path.join(URL_PREFIX, self.schema_url)
        asdf_file = AsdfFile()
        schema = asdf_schema.load_schema(schema_path,
                                         resolver=asdf_file.resolver,
                                         resolve_references=True)
        self._schema = mschema.merge_property_trees(schema)
```

The MIRIExtension class (derived from AsdfExtension) still exists but is no longer used. I discovered this change prevents data models which are derived from the MIRI models from working because it is no longer possible to reference the MIRI schemas. For example, the MiriTE package can define a data model which references a schema in the jwst package with the URL

```
$ref:http://stsci.edu/schemas/jwst_datamodel/<filename>.schema
```

but if the MiriTeam package defines a new data model based on an existing MIRI data model, it can’t reference the MIRI schemas using the URL

```
$ref:http://stsci.edu/schemas/miri_datamodel/<filename>.schema
```

Worse still, the data model isn’t able to access its own schemas. Specifying

```
   schema_url = "<filename>.schema.yaml"
```

attempts to load a file with an incorrect name (`<filename>.schema.yaml.yaml`) from the correct schema directory (the one owned by the new package) but

```
   schema_url = "<filename>.schema"
```

attempts to load a file with the correct name (`<filename>.schema.yaml`) from the wrong schema directory (the one owned by the MiriTE data models).

It looks like the new code added to the MiriModelBase class only solves the issue of referencing the JWST models from the MIRI data models. What do I need to do to allow a custom data model to reference the MIRI data models? Is the AsdfExtension class obsolete, or can I still use it? Do I need to add another call to “asdf_schema.load_schema” within a derived data model so its schema files are loaded from the directory owned by the local package?

[...]

PR #4435, opened 2020-01-07
Update datamodels schemas

eslavich:
This PR will make datamodels schemas more friendly to users who want to subclass DataModel (e.g., https://github.com/spacetelescope/jwst/issues/4246), and also lay the groundwork for a fix in asdf for https://github.com/spacetelescope/asdf/issues/683.  It's a large diff, but the changes were mostly automated.  Here's a summary:

- Change schema_url values in model classes to be absolute URIs, so that DataModel can be subclassed with schemas from other asdf extensions.
- Add an `id` field to each model schema so that relative references are unambiguous.
- Since I was in there anyway, add the missing `%YAML` directive and move the `$schema` declaration to the top to match the schemas in asdf-standard.
- Drop the `.yaml` extension from references in model schemas so that they can be treated as relative URIs instead of relative filesystem paths.

These changes are compatible with existing releases of asdf.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,10 @@ datamodels
 
 - Make memmap=False be the default in datamodels [#4445]
 
+- Update schemas to add the ``id`` field and switch relative references
+  from filesystem paths to URIs.  Make ``schema_url`` absolute to facilitate
+  subclassing DataModel with schemas from other asdf extensions. [#4435]
+
 extract_1d
 ----------
 
@@ -569,7 +573,7 @@ outlier_detection
 
 - Don't use NaNs or masked values in weight image for blotting. [#3651]
 
-- When calling cube_build for IFU data fixed selecting correct channels (MIRI) or 
+- When calling cube_build for IFU data fixed selecting correct channels (MIRI) or
   correct grating (NIRSPEC) [#4301]
 
 pipeline
--- a/jwst/datamodels/amilg.py
+++ b/jwst/datamodels/amilg.py
@@ -34,7 +34,7 @@ class AmiLgModel(DataModel):
     solns_table : numpy table
          Solutions table
     """
-    schema_url = "amilg.schema"
+    schema_url = "http://stsci.edu/schemas/jwst_datamodel/amilg.schema"
 
     def get_primary_array_name(self):
         """
--- a/jwst/datamodels/apcorr.py
+++ b/jwst/datamodels/apcorr.py
@@ -23,7 +23,7 @@ class FgsImgApcorrModel(ReferenceFileModel):
         - skyout: float32
 
     """
-    schema_url = "fgsimg_apcorr.schema"
+    schema_url = "http://stsci.edu/schemas/jwst_datamodel/fgsimg_apcorr.schema"
 
 
 class MirImgApcorrModel(ReferenceFileModel):
@@ -47,7 +47,7 @@ class MirImgApcorrModel(ReferenceFileModel):
         - skyout: float32
 
     """
-    schema_url = "mirimg_apcorr.schema"
+    schema_url = "http://stsci.edu/schemas/jwst_datamodel/mirimg_apcorr.schema"
 
 
 class NrcImgApcorrModel(ReferenceFileModel):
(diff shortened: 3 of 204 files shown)


nden:
It looks good in general.
One question though - should the `fits-schema` schema be applied to all files? Some of the schemas represent files which will never be converted to FITS, particularly the WCS reference files are almost all pure ASDF files. Should those use the `yaml-schema`?

nden:
It would be great if you can test [the MIRI team package](https://github.com/JWST-MIRI/MiriTE) with this PR.

nden:
I guess it doesn't really matter that the `fits-schema` is defined in schemas which support pure asdf files since the asdf schema is there too.

eslavich:
@nden you make a good point about `fits-schema`.  I agree that there aren't any practical consequences, but still it's probably confusing to people who look into the schemas and wonder why they need fits_keyword or fits_hdu.  I'm happy to change them but I'll need help identifying the ones that should use the vanilla asdf schema.

eslavich:
Here are the results from running the MiriTE/datamodels tests for various versions of jwst:

0.13.7:
```
279 passed, 6 warnings
```

master:
```
28 failed, 251 passed, 6 warnings
```

this branch:
```
28 failed, 251 passed, 6 warnings
```

It doesn't seem to make things worse, anyway.

nden:
Thanks for testing MiriTE!

The WCS schemas are:

distortion.schema
distortion_mrs.schema
specwcs.schema
specwcs_nircam_grism.schema
specwcs_niriss_grism.schema
regions.schema
wavelengthrange.schema
fpa.schema
ifupost.schema
ifuslicer.schema
msa.schema
disperser.schema
filteroffset.schema
ifufore.schema
camera.schema
collimator.schema
ote.schema
fore.schema
wavecorr.schema

stscicrawford:
Miri is having some problems with the current CRDS context, so that might be related to the current failures.    

It would also be good to see how this would impact the MIRI simulator as well and to reach out to the team for a review about this as I know they use the data models for their work.

jdavies-st:
One extra point about `fits-schema` is that it itself appears to not be set up correctly.

https://github.com/spacetelescope/jwst/blob/2a423ea17d704f487fc6b092c2f438097bc20e0c/jwst/datamodels/metaschema/fits-schema.yaml#L1-L10

Though maybe this is my misunderstanding of how metaschema validation works?  It seems that it is a metaschema that validates against itself - `$schema` and `id` are the same.  I would expect it to validate against either `asdf-schema-1.0.0` or `http://stsci.edu/schemas/yaml-schema/draft-01`?

Or is it because it has the following:

https://github.com/spacetelescope/jwst/blob/2a423ea17d704f487fc6b092c2f438097bc20e0c/jwst/datamodels/metaschema/fits-schema.yaml#L9-L10

that it can validate against itself?

None of the other metaschemas in `asdf-standard` validates against itself in this way.

If I do try to change `fits-schema` to validate against the asdf or yaml schemas above, it fails.

(All this is unrelated to this PR, but I figured I'd bring it up so as to gain some understanding)

jdavies-st:
Two more `fits-schema`-related points (again unrelated to this PR):

 1.  It might be a good idea to rename this schema at some point.  It is really an `asdf-in-fits-schema`.  There is another FITS schema in the `asdf` package

https://github.com/spacetelescope/asdf-standard/blob/master/schemas/stsci.edu/asdf/fits/fits-1.0.0.yaml

that is for stuffing a FITS format file into ASDF.  The opposite.
   
 2. And version `fits-schema`, even if it remains at 1.0.0 forever.

Thoughts?

rizeladiaz:
what is the context number that is having the issues?

jdavies-st:
@nden and @eslavich can we merge this?  Looks like some small conflicts need to be resolved due to some changes to ifucubepars.

eslavich:
I updated the WCS schemas and resolved the conflicts.  Still need to chat with MIRI simulator folks.

eslavich:
Per discussion on Slack, MIRI folks are satisfied with our tests against MiriTE.

eslavich:
@stscieisenhamer suggested that I run the regression tests on this branch, I'll merge once that's done.

jdavies-st:
Regression test run on this branch here:

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/153/

jdavies-st:
Regression tests on this branch pass.  Merging.

PR #4437, opened 2020-01-08
test for miri lrs tso1

jemorrison:
@jdavies-st 
Should I combine this test with test_miri_lrs_slitless.py ?
This pr is a test for miri lrs tso1 and the test_lrs_slitless.py is a test for tso2 and 3.
They do use different data.


jwst/regtest/test_miri_lrs_slitless_tso1.py
```diff
@@ -0,0 +1,50 @@
+import pytest
+from astropy.io.fits.diff import FITSDiff
+
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+from jwst.stpipe import Step
+
+
+DATASET_ID = "det_image_MIRIMAGE_P750Lexp1"
```
hbushouse: I think it would be best to use standard DMS style file names for the input and output products, because the infrastructure within ``stpipe`` is setup to look for and do certain things with the various parts of the standard file names. Even if you just have to make up something, like jw1234500101_etc. Better yet, since it appears that these data came from MIRISim, the data array itself should be copied into an actual MIRI LRS slitless _uncal file that was produced by DMS, so that we know for sure that everything about the file structure, organization, and header keywords is completely as expected from DMS.

jwst/regtest/test_miri_lrs_slitless_tso1.py
```diff
@@ -0,0 +1,50 @@
...
+        "config/calwebb_tso1.cfg",
+        rtdata.input,
+        "--steps.dq_init.save_results=True",
+        "--steps.saturation.save_results=True",
+        "--steps.rscd.save_results=True",
+        "--steps.linearity.save_results=True",
+        "--steps.dark_current.save_results=True",
+        "--steps.refpix.save_results=True",
+        "--steps.jump.rejection_threshold=25",
+        "--steps.jump.save_results=True"
```
hbushouse: What about the results of the ``firstframe`` and ``lastframe`` steps? Those could be included too.

jwst/regtest/test_miri_lrs_slitless_tso1.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@pytest.mark.bigdata
+@pytest.mark.parametrize("step_suffix", ['rate', 'rateints', 'linearity', 'rscd',
+                                         'dq_init',
+                                         'saturation', 'dark_current', 'refpix',
+                                         'jump'])
+def test_miri_lrs_slitless_tso1(run_tso1_pipeline, fitsdiff_default_kwargs, step_suffix):
+    """
+    Regression test of calwebb_detector1 pipeline performed on MIRI data.
```
hbushouse: Should say "calwebb_tso1" pipeline and also be more specific in that it's performed on MIRI LRS slitless data.

jwst/regtest/test_miri_lrs_slitless_tso1.py
jemorrison: @hbushouse I do not think the first or last frame is done for MIRI tso1.  I looked at the existing  cal webb_tso1.cfg and at the document K Gordon wrote https://outerspace.stsci.edu/display/JWSTCC/CALWEBB_DETECTOR1
and it looks like both steps are skipped.

jwst/regtest/test_miri_lrs_slitless_tso1.py
jemorrison: @bushhouse not sure what you mean by "the data array itself should be copied into an actual MIRI LRS slitless_uncal file." I had to take a simulated data set read it and and chop out 90% of the file.  Where can I get ahold of a DMS produced MIRI LRS slitless_cal file ?

hbushouse: Since this test runs the ``calwebb_tso1`` pipeline I assume that the simulated data is in raw 4-D format, correct? If so, then you can just copy the 4-D data array from "det_image_MIRIMAGE_P750Lexp1" into any of the MIRI LRS-SLITLESS uncal files produced by DMS B7.4 that are sitting in "/grp/jwst/ins/mary/b7.4_cal14_full/", such as:

/jw00859/jw00859001001_03102_00001_mirimage_uncal.fits
/jw00623/jw00623024001_02103_00001_mirimage_uncal.fits
/jw00623/jw00623026001_03103_00010_mirimage_uncal.fits
etc.

You may have to modify a few header keyword values for things like NGROUPS and NINTS to make them match the simulated data.

jemorrison: @hbushouse well I do not have permission to get data in that directory
Who should I contact so I can get permission ?

hbushouse: @mromelfanger In the meantime let me know what file(s) you'd like and I can grab them for you.

jemorrison: @hbushouse I just need any one of the files you suggested.

hbushouse: You can grab:
/grp/jwst/ssb/bushouse/jwst_data/MIRI/LRS/SLITLESS/jw00623026001_03106_00005_mirimage_uncal.fits

jemorrison:
@hbushouse  I put the tso1 test into the existing tests for miri_lrs_slitless.py to organize all the tests together. Transferring new input and truth files to artifactory now

hbushouse:
Any reason why this can't be merged? The PR has been approved and the data appears to be present in artifactory.

Issue #4404, opened 2019-12-30
REGTEST: Write new MIRI LRS fixed-slit level3 regression test


stscijgbot:
Issue [JP-1211](https://jira.stsci.edu/browse/JP-1211) was created by Howard Bushouse:

Create a regression test in the new regtest system for applying the "calwebb_spec3"  pipeline to an association of MIRI LRS fixed-slit exposures (e.g. multiple exposures resulting from an ALONG-SLIT-NOD dither pattern). Could potentially be combined with the existing "calwebb_spec2" test for MIRI LRS fixed-slit data (JP-1188).

PR #4440, opened 2020-01-08
JP-1211 MIRI LRS fixed-slit level3 regression test

ddavis-stsci:
Fixes JP-1211 and #4404

Local tests all pass
pytest --env=newdev --bigdata --basetemp=/Users/ddavis/src/JWST/develop/JP-1211/test_Jan06 ~/src/JWST/scsb/jwst/jwst/regtest/ -k test_miri_lrs_fs_spec3
...
================ 2 passed, 219 deselected, 2 warnings in 35.01s ================

This test should have better data as there are several warnings for "All-NaN slice encountered"

### DIFF
--- a/jwst/regtest/test_miri_lrs_fs_spec3.py
+++ b/jwst/regtest/test_miri_lrs_fs_spec3.py
@@ -1,0 +1,48 @@
+""" Test for the spec3 pipeline using MIR data in LRS mode. This takes
+    the an association and generates the level 3 products."""
+import os
+import pytest
+from astropy.io.fits.diff import FITSDiff
+
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+from jwst.stpipe import Step
+
+@pytest.fixture(scope="module")
+def run_pipeline(jail, rtdata_module):
+    """Run the calwebb_spec3 pipeline on an ASN of nodded MIRI LRS
+       fixedslit exposures."""
+
+    rtdata = rtdata_module
+
+    # Get the cfg files
+    collect_pipeline_cfgs("config")
+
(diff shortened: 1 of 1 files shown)


hbushouse:
@ddavis-stsci do you happen to recall (or can you figure out) where you got the input cal files for this test? While examining some recent regtest runs I've noticed that the SCI arrays of all the input cal files appear to be zero-filled, thus resulting in zero-filled combined data products from calwebb_spec3.

ddavis-stsci:
Hmmm. good question. I do have a note from when I was working on this 
that I was getting all 0 arrays.

I think the _rate files are from
/grp/jwst/ssb/bushouse/jwst_data/MIRI/LRS/Sims/along-slit-nod/set2

I think this was based on your comments in  JP-1188 "...his particular 
set of simulated data does NOT have the excessive levels of noise that 
some other MIRISim images have, so the jump and ramp_fit steps in 
calwebb_detector1 do not go crazy during processing"

I think most of my notes for this test are in a notepad in my office...

hbushouse:
I was just looking at those files and the rate files all look good (you can see a spectrum that moves back and forth, due to the along-slit-nod, from one exposure to the next). Only 1 of the exposures in that directory has a cal file (the first exposure in the series) and that looks good too. There's a slight difference in the file names for these vs. the ones in the spec3 regtest:

jw00623032001_310[24]_0000[12] vs. jw00623044001_0210[13]_0000[12]

so I'm guessing the ones in the regtest came from somewhere else.

I'm going to try creating a lev-3 ASN for the ones in my directory and see how they work when used for the spec3 test. If they look OK, then I'll just update the regtest to use this different set of inputs.

PR #4444, opened 2020-01-10
JP-1225 MIRI master_background tests copied to new regtest

philhodge:
Tests of the `master_background` step for MIRI data have been created in the new regression test system, using input data copied from the old regtests.  There is a test with a user-supplied background file, a test with each of LRS and MRS data for the case of nodded exposures, and a similar pair of tests for the case of dedicated background exposures.

### DIFF
--- a/jwst/regtest/test_miri_lrs_dedicated_mbkg.py
+++ b/jwst/regtest/test_miri_lrs_dedicated_mbkg.py
@@ -1,0 +1,40 @@
+import os
+import pytest
+from astropy.io.fits.diff import FITSDiff
+
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+from jwst.stpipe import Step
+
+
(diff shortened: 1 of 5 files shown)


jwst/regtest/test_miri_lrs_dedicated_mbkg.py
```diff
@@ -0,0 +1,40 @@
...
+    Step.from_cmdline(args)
+
+    return rtdata
+
+
+@pytest.mark.bigdata
+def test_miri_lrs_dedicated_mbkg(run_pipeline, fitsdiff_default_kwargs):
+    """Run a test for MIRI LRS data with dedicated background exposures."""
+
+    # Run the step and retrieve the output
```
jdavies-st: This comment is misleading.  The running of the pipeline actually happens on test setup and the it returns a `RegtestData` object to this variable.  Here it is just assigning it to a different name.

Issue #2887, opened 2018-11-20
Unit tests leave behind open files


jdavies-st:
The following unit tests leave behind open files:
```
[gw0] [ 10%] ERROR jwst/assign_wcs/tests/test_schemas.py::test_distortion_schema
[gw1] [ 12%] ERROR jwst/datamodels/tests/test_fits.py::test_delete 
[gw1] [ 12%] ERROR jwst/datamodels/tests/test_fits.py::test_extra_fits 
[gw1] [ 12%] ERROR jwst/datamodels/tests/test_fits.py::test_casting 
[gw1] [ 12%] ERROR jwst/datamodels/tests/test_fits.py::test_metadata_doesnt_override 
[gw1] [ 12%] ERROR jwst/datamodels/tests/test_fits.py::test_metadata_from_fits 
[gw1] [ 13%] ERROR jwst/datamodels/tests/test_history.py::test_history_from_model_to_fits 
[gw1] [ 13%] ERROR jwst/datamodels/tests/test_history.py::test_history_from_fits 
[gw1] [ 14%] ERROR jwst/datamodels/tests/test_models.py::test_subarray 
[gw1] [ 15%] ERROR jwst/datamodels/tests/test_models.py::test_from_fits_write 
[gw1] [ 16%] ERROR jwst/datamodels/tests/test_models.py::test_multislit_from_fits_image 
[gw1] [ 16%] ERROR jwst/datamodels/tests/test_models.py::test_multislit_copy 
[gw1] [ 18%] ERROR jwst/datamodels/tests/test_models.py::test_image_with_extra_keyword_to_multislit 
[gw1] [ 19%] ERROR jwst/datamodels/tests/test_open.py::test_open_image 
[gw1] [ 19%] ERROR jwst/datamodels/tests/test_open.py::test_open_reference_files 
[gw1] [ 20%] ERROR jwst/datamodels/tests/test_schema.py::test_mask_model 
[gw1] [ 21%] ERROR jwst/datamodels/tests/test_schema.py::test_validate_transform_from_file 
[gw3] [ 33%] ERROR jwst/datamodels/tests/test_fits.py::test_from_fits 
[gw3] [ 33%] ERROR jwst/datamodels/tests/test_fits.py::test_from_scratch 
[gw3] [ 42%] ERROR jwst/lib/tests/test_set_telescope_pointing.py::test_add_wcs_default 
[...]

PR #4445, opened 2020-01-10
Make memmap=False be the default in datamodels

### LINKED ISSUE #4431: ValueError: I/O operation on closed file
Issue [JP-1227](https://jira.stsci.edu/browse/JP-1227) was created by Jonathan Eisenhamer:

[Regression|https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/lastCompletedBuild/testReport/jwst.tests_nightly.general.nirspec.test_nirspec_msa_spec3/TestSpec3Pipeline/_stable_deps__test_nrs_msa_spec3/] jwst.tests_nightly.general.nirspec.test_nirspec_msa_spec3.TestSpec3Pipeline.[stable-deps] test_nrs_msa_spec3

fails with

ValueError: I/O operation on closed file

jdavies-st:
This turns off the memmap=True default when we open FITS files in datamodels. 

This is a potential bandaid fix for #4431/JP-1227.  It doesn't so much fix the problem as avoid the problem by not memory mapping.

Fixes #4431

Addresses some of #2887

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,8 @@ datamodels
 
 - Deprecate MIRIRampModel [#4328]
 
+- Make memmap=False be the default in datamodels [#4445]
+
 extract_1d
 ----------
 
--- a/docs/jwst/assign_wcs/asdf-howto.rst
+++ b/docs/jwst/assign_wcs/asdf-howto.rst
@@ -137,9 +137,9 @@ catch any errors due to inconsistent inputs/outputs or invalid parameters.
 
 To test the file, it can be read in again using the ``asdf.open()`` function:
 
->>> ff = asdf.open('reffile.asdf')
->>> model = ff.tree['model']
->>> model(1)
+>>> with asdf.open('reffile.asdf') as ff:
+...     model = ff.tree['model']
+...     model(1)
     -152.2
 
 
--- a/jwst/ami/tests/test_ami_interface.py
+++ b/jwst/ami/tests/test_ami_interface.py
@@ -5,17 +5,15 @@ from jwst.ami import AmiAnalyzeStep
 import jwst
 
 
-def test_ami_analyze_calints_fail(tmpdir):
+def test_ami_analyze_calints_fail():
     """Make sure ami_analyze fails if input is CubeModel (_calints)"""
-    input_file = str(tmpdir.join("ami_analyze_input.fits"))
     model = datamodels.CubeModel((25, 19, 19))
     model.meta.instrument.name = "NIRISS"
     model.meta.instrument.filter = "F277W"
     model.meta.observation.date = "2019-01-01"
     model.meta.observation.time = "00:00:00"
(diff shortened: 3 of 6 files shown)


hbushouse:
What sort of impact could this have on performance?

jdavies-st:
Good question.  Datamodels circumvent the default FITS lazy loading because they instantiate the `data`, `dq` and `err` arrays right away.  And as soon as the input FITS file is loaded into the first step, a copy is made in memory, and that copy is then used for every step subsequent.  So the memmaping should not effect anything downstream.  And we never write back to the input FITS file in the pipeline, so again, no reason really to be using memmap.

I'll run regression test run on this branch to look at performance.

And of course we can always revert this if needed.  The changes in this PR are minimal.

jdavies-st:
The flipside would be to actually thoroughly debug the code in `MultiExposureModel` and `exp_to_source` and see what is causing the issue.  There probably needs to be a `copy()` inserted in there somewhere, but I don't know where.  @stscieisenhamer might have a better idea?

That said, there's already a PR (#4322) that completely revamps those two sections of code for other reasons, and I don't want to put too much effort into the old code if it is being replaced with the new.  Does the offending test pass under the new code @stscieisenhamer?

stscieisenhamer:
Concerning performance: Just out of curiosity, running the regression without mmap may prove to be very enlightening anyways.

About #4322, I can see if there is this error, though, if there is, that would be enlightening in and of itself, since the point of the PR is to remove `SourceContainer`.

jdavies-st:
The regression test build on this branch:

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/152/

stscieisenhamer:
Looking at @jdavies-st regression, if anything, the performance seemed to improve.

jdavies-st:
Yeah, slight improvement in aggregate, which is not a surprise given how our datamodels and pipelines work.

jdavies-st:
Good question.  You can reach that info via the `_files_to_close` attribute which is where the underlying `HDUList` is stored.
```
In [25]: m = datamodels.open('/user/jdavies/jp-1179/jw42424001001_01101_00003_nrca4_cal.fits')

In [24]: m._files_to_close[0]._file.__dict__
Out[24]: 
{'strict_memmap': False,
 'simulateonly': False,
 '_mmap': None,
 'name': '/user/jdavies/jp-1179/jw42424001001_01101_00003_nrca4_cal.fits',
 'closed': False,
 'binary': True,
 'mode': 'readonly',
 'memmap': False,
 'file_like': False,
 'close_on_error': True,
 'compression': None,
 'readonly': True,
 'writeonly': False,
 '_file': <_io.FileIO name='/user/jdavies/jp-1179/jw42424001001_01101_00003_nrca4_cal.fits' mode='rb' closefd=True>,
 'fileobj_mode': 'rb',
 'size': 117515520}
```

PR #4448, opened 2020-01-14
JP-1226: Add nirspec masterbackground tests to the new test framework.

### LINKED ISSUE #4427: REGTEST: Move nirspec masterbackground tests to new regressions
Issue [JP-1226](https://jira.stsci.edu/browse/JP-1226) was created by Jonathan Eisenhamer:

Create new regression tests for the Master Background Subtraction step in the calwebb_spec3 pipeline. There are several different ways the step can be applied, including creating its own master background spectrum from individual x1d inputs versus using a user-supplied master background spectrum, as well as the creation of the master background spectrum from background regions within nodded exposures versus dedicated background exposures. All of these various options need to be covered, both for slit-like spectra (NIRSpec MOS and fixed-slit, MIRI LRS fixed-slit) as well as IFU spectra (NIRSpec IFU and MIRI MRS).

The tests could be constructed as standalone tests of just the master_background step itself (which is the case for the existing tests in the old regtest system) or they could be incorporated as part of overall calwebb_spec3 tests of the various spectroscopic modes mentioned above.

jwhite3:
This PR adds NIRSpec MasterBackground tests in the new testing framework.

Fixes #4427 and JP-1226.

### DIFF
--- a/jwst/regtest/test_nirspec_masterbackground.py
+++ b/jwst/regtest/test_nirspec_masterbackground.py
@@ -1,0 +1,169 @@
+import os
+import pytest
+import numpy as np
+
+from astropy.io.fits.diff import FITSDiff
+from numpy.testing import assert_allclose
+from typing import List
+
+from gwcs.wcstools import grid_from_bounding_box
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+from jwst.assign_wcs import nirspec
+from jwst.stpipe import Step
+from jwst import datamodels
+
+
+@pytest.fixture
+def run_nirspec_mbkg(jail, rtdata_module):
+    """Fixture-factory that runs the MasterBackground step for NIRSpec MasterBackground tests."""
+    rtdata = rtdata_module
+    collect_pipeline_cfgs('config')
+
+    def _set_pipeline_run(input_file: str, user_files: List[tuple] = None, asn: bool = False,
+                          options: List[str] = None):
+        """Set additional pipeline options for the different MasterBackground tests.
+        Additional user_files and their inputs (inputs are assumed to be artifactory paths) should be given as tuples
+        for the user_files argument.
+        Set asn to True if getting an association.
+        Add additional command line options with the options argument.
+        """
+        # Create optional arguments list using rtdata to find artifact paths
+        user_files = [
+            f'{arg}={rtdata.get_data(input_path)}' for arg, input_path in user_files
+        ] if user_files is not None else []
+
+        if options is None:
+            options = []
+
+        # Get the input data
+        if asn:
+            rtdata.get_asn(input_file)
+
+        else:
+            rtdata.get_data(input_file)
+
+        # Set arguments and execute
+        args = ['config/master_background.cfg', rtdata.input, '--save_results=True'] + user_files + options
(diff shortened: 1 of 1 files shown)


jwst/regtest/test_nirspec_masterbackground.py
```diff
@@ -0,0 +1,174 @@
...
+    rtdata.output = 'prism_sci_bkg_master_background.fits'
+
+    # Get the truth file
+    rtdata.get_truth(os.path.join("truth/test_nirspec_ifu_mbkg_user", 'prism_sci_bkg_master_background.fits'))
+
+    # Compare the results
+    diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
+    assert diff.identical, diff.report()
+
+    # Test 2  compare the science  data with no background to the output from the masterBackground Subtraction step
```
hbushouse: These additional tests that compare data processed one way to data from some other source were originally implemented in the old reg test system when we were in the early stages of developing the master background step and needed some sanity checks to verify that the overall algorithm was doing the right thing (i.e. getting close to the expected result generated by hand). Now that we're confident in the design of the algorithm itself, I think these types of additional tests could be removed. They add a lot of extra baggage and are difficult to maintain and update. All we really need are the very simple tests that run the step and compare its various forms of output products against truth files of the same product.

jwhite3: No problem! I can get rid of those

jwhite3:
After the discussion, I uploaded new truth files from a previous run to ensure that runs were consistent. After that, all tests passed

jwst/regtest/test_nirspec_masterbackground.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # Compare the results
+    diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
+    assert diff.identical, diff.report()
+
+
+@pytest.mark.parametrize(
+    'output_file', ['ifu_prism_source_on_NRS1_master_background.fits', 'ifu_prism_source_off_NRS1_o001_masterbg.fits']
+)
+def test_nirspec_ifu_nod_mbkg(run_nirspec_mbkg, fitsdiff_default_kwargs, output_file):
```
hbushouse: Minor nit, but I'm a consistency nut: How 'bout changing the name of the function to "test_nirspec_ifu_mbkg_nod", so that the naming is consistent with the other test functions above.

jwst/regtest/test_nirspec_masterbackground.py
```diff
@@ -0,0 +1,122 @@
...
+        'nirspec/ifu/nirspec_spec3_asn.json',
+        asn=True,
+        options=['--save_background=True']
+    )
+
+    rtdata.output = output_file
+
+    # Get the truth file
+    rtdata.get_truth(
+        os.path.join("truth/test_nirspec_ifu_nod_mbkg", output_file)
```
hbushouse: Same request here: change the name of the truth path to "test_nirspec_ifu_mbkg_nod"

jwhite3: Updated the test name and the corresponding truth path. Re-ran the tests and everything still passes

PR #4450, opened 2020-01-15
JP-1219: Add missing NIRSpec regtests

hbushouse:
Add new NIRSpec regression tests to fill holes in coverage from the old set. In particular, this adds tests for FS and IFU WCS computations, as well as exceptions that are often encountered due to no data falling on the NRS2 detector (for some filter+grating combinations) and missing MSA meta data files.

A few existing test modules have been updated to change the directory paths of their input and/or truth files, just to better and more consistently organize the data on artifactory. All necessary changes to artifactory have been made.

A few more holes still need to be filled, including a test of the ``imprint`` step in ``calwebb_spec2``.

### DIFF
--- a/jwst/regtest/test_nirspec_exceptions.py
+++ b/jwst/regtest/test_nirspec_exceptions.py
@@ -6,6 +6,7 @@ from jwst.stpipe import Step
 from jwst.assign_wcs.util import NoDataOnDetectorError
 from jwst.pipeline import Spec2Pipeline
 
+
 @pytest.mark.bigdata
 def test_nirspec_missing_msa_fail(_jail, rtdata, fitsdiff_default_kwargs, caplog):
     """
@@ -13,7 +14,7 @@ def test_nirspec_missing_msa_fail(_jail, rtdata, fitsdiff_default_kwargs, caplog
         that's missing an MSAMETFL. Exception should be raised.
     """
 
-    # Get the input file 
+    # Get the input file
     rtdata.get_data('nirspec/mos/f170lp-g235m_mos_observation-6-c0e0_001_dn_nrs1_mod.fits')
 
     # Run the calwebb_spec2 pipeline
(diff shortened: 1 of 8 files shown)


hbushouse:
Latest commit modifies test_nirspec_ifu.py to add an imprint image to the calwebb_spec2 test, to satisfy the missing test of the imprint step.

hbushouse:
Latest commit adds a ``calwebb_spec3`` test for NIRSpec fixed-slit (FS) exposures. Note that this also includes a temporary quick-fix to one of the ``extract_1d`` modules to get around the issue reported in JP-1144, which otherwise prevents these data from processing successfully.

jwst/regtest/test_nirspec_fs_spec3.py
```diff
@@ -0,0 +1,50 @@
...
+    # Get the cfg files
+    collect_pipeline_cfgs("config")
+
+    # Get the ASN file and input exposures
+    rtdata.get_asn('nirspec/fs/jw93045-o010_20180725t035735_spec3_001_asn.json')
+
+    # Run the calwebb_spec3 pipeline; save results from intermediate steps
+    args = ["config/calwebb_spec3.cfg", rtdata.input,
+            "--steps.outlier_detection.save_results=true",
+            "--steps.resample_spec.save_results=true",
```
jdavies-st: This `resample_spec.save_results=True` should not be necessary, right?  The pipeline itself makes sure this is written out.  Though maybe no harm if it is there?

hbushouse: That was my thought too: it's already hardwired in the ``calwebb_spec3.cfg`` file and potentially in the pipeline/step code, but I figured it can't hurt to make doubly sure it's set.

jwst/regtest/test_nirspec_exceptions.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Call the Spec2Pipeline
+    cmd = [
+        'strun',
+        'jwst.pipeline.Spec2Pipeline',
+        rtdata.input]
+
+    status = subprocess.run(cmd)
+
+    assert status.returncode == 64
```
nden: What does this return code mean?

hbushouse: I have no idea, but that's the value that comes back!

stscieisenhamer: https://jwst-pipeline.readthedocs.io/en/latest/jwst/introduction.html#exit-status

PR #4455, opened 2020-01-21
JP-1231 Update test data and refactor test

### LINKED ISSUE #4451: test_nirspec_ifu failing with ValueError: output array is read-only
Issue [JP-1231](https://jira.stsci.edu/browse/JP-1231) was created by Jonathan Eisenhamer:

New regression test {{test_nirspec_ifu.run_spec3_multi}} is failing with a {{ValueError: output array is read-only}}

This is most likely due to garbage data, but need be confirmed.

stscieisenhamer:
Resolves #4451/[JP-1231](https://jira.stsci.edu/browse/JP-1231)

### DIFF
--- a/jwst/regtest/test_nirspec_ifu.py
+++ b/jwst/regtest/test_nirspec_ifu.py
@@ -1,14 +1,36 @@
 """Regression tests for NIRSpec"""
+from pathlib import Path
 import pytest
 
+from jwst.associations import load_asn
+from jwst.lib.suffix import replace_suffix
+
 from . import regtestdata as rt
+
+# Define artifactory source and truth
+INPUT_PATH = 'nirspec/ifu'
+TRUTH_PATH = 'truth/test_nirspec_ifu'
 
 
 @pytest.fixture(scope='module')
 def run_spec2(jail, rtdata_module):
-    """Run the pipelines"""
+    """Run the Spec2Pipeline on a single exposure"""
+    rtdata = rtdata_module
+
+    # Setup the inputs
+    asn_name = 'single_nrs1_spec3_asn.json'
+    rtdata.get_data(INPUT_PATH + '/' + asn_name)
+    asn_path = rtdata.input
+    with open(asn_path, 'r') as asn_fh:
+        asn = load_asn(asn_fh)
+    member_path = Path(asn['products'][0]['members'][0]['expname'])
+    rate_path = member_path.stem
+    rate_path = replace_suffix(rate_path, 'rate')
+    rate_path = INPUT_PATH + '/' + rate_path + member_path.suffix
+
+    # Run the pipeline
     step_params = {
-        'input_path': 'nirspec/ifu/nrs_ifu_spec2_asn.json',
+        'input_path': rate_path,
         'step': 'calwebb_spec2.cfg',
         'args': [
             '--steps.bkg_subtract.save_results=true',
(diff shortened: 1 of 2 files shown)


jwst/regtest/test_nirspec_ifu.py
```diff
@@ -1,14 +1,36 @@
...
+    # Setup the inputs
+    asn_name = 'single_nrs1_spec3_asn.json'
+    rtdata.get_data(INPUT_PATH + '/' + asn_name)
+    asn_path = rtdata.input
+    with open(asn_path, 'r') as asn_fh:
+        asn = load_asn(asn_fh)
+    member_path = Path(asn['products'][0]['members'][0]['expname'])
+    rate_path = member_path.stem
+    rate_path = replace_suffix(rate_path, 'rate')
+    rate_path = INPUT_PATH + '/' + rate_path + member_path.suffix
```
philhodge: If Artifactory expects the path separator to be '/', what is the point of using the `pathlib` module to construct part of the path for a file name on Artifactory?

jwst/regtest/test_nirspec_ifu.py
```diff
@@ -1,14 +1,36 @@
...
+
+    # Run the pipeline
     step_params = {
-        'input_path': 'nirspec/ifu/nrs_ifu_spec2_asn.json',
+        'input_path': rate_path,
         'step': 'calwebb_spec2.cfg',
         'args': [
```
philhodge: Why save the output from the `straylight`, `fringe`, and `barshadow` steps when the data are NIRSpec IFU?

jwst/regtest/test_nirspec_ifu.py
```diff
@@ -1,14 +1,36 @@
...
 
 @pytest.fixture(scope='module')
 def run_spec2(jail, rtdata_module):
-    """Run the pipelines"""
+    """Run the Spec2Pipeline on a single exposure"""
+    rtdata = rtdata_module
+
+    # Setup the inputs
+    asn_name = 'single_nrs1_spec3_asn.json'
+    rtdata.get_data(INPUT_PATH + '/' + asn_name)
```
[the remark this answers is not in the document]
stscieisenhamer: 50/50 here: The original intent was to have the "chained" pipelines work off of one file, which this approach uses the association as the primary input. As such, one doesn't what to use `get_asn` because the test doesn't want the `cal` file; the `run_spec2` will produce that.

The other side is to use the single `rate` file as the primary input, and build the association for the `run_spec3` fixture.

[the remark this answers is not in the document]
hbushouse: Saying that "the test doesn't want the ``cal`` file" suggests that the ASN file in use is a level-3 ASN, where the members are in fact ``cal`` files? But this particular test module is running the ``spec2`` pipeline, so why are you using a level-3 ASN as input? Why not use an actual level-2 ("spec2") ASN that has ``rate`` file(s) as members?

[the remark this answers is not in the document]
stscieisenhamer: Agreed. Makes more sense to use the single file as the primary input. Will fix.

jwst/regtest/test_nirspec_ifu.py
stscieisenhamer: Though the current test does not have this, the test was using the local file path to get file name components. This has been removed.

jwst/regtest/test_nirspec_ifu.py
stscieisenhamer: Just the generic set of step parameters. If a step is not used, no output actually happens anyways.

jwst/regtest/test_nirspec_ifu.py
```diff
@@ -49,34 +100,52 @@ def run_spec3(jail, rtdata_module):
...
 
-    return rt.run_step_from_dict(rtdata_module, **step_params)
+    rtdata = rt.run_step_from_dict(rtdata, **step_params)
+    return rtdata
 
 
 @pytest.mark.bigdata
 @pytest.mark.parametrize(
     'suffix',
-    ['assign_wcs', 'cal', 'flat_field', 'imprint_subtract', 'msa_flagging', 'pathloss', 'photom', 's3d', 'srctype', 'x1d']
```
hbushouse: So has testing of the ``imprint`` step been removed from the test? This was added recently in order to fill a hole in NIRSpec testing. It requires running the ``calwebb_spec2`` pipeline with an ASN as input, which includes the imprint image as one of its members. And of course setting ``steps.imprint_subtract.save_results=true``.

stscieisenhamer: Aaaaannnnnnnddddd this would be the case of getting data stomped on. Will restore.

stscieisenhamer: Pushed. One more try...

PR #4459, opened 2020-01-21
Populate the background column

philhodge:
For the case that the target is a point source and the instrument is NIRSpec or NIRISS (SOSS), the input units are megajanskys, and the background was being set to zero.  The extracted background is now divided by the pixel area (solid angle) to convert to units of surface brightness.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,9 @@ extract_1d
 
 - Updated to work with the current output from photom [#4369]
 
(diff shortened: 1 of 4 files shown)


hbushouse:
Oh, what about an update to the ``extract_1d`` docs to mention that this now happens to background spectra.

philhodge:
That's a good point.  The docs are not up to date anyway; I'll make some changes.

Issue #4429, opened 2020-01-06
Make new regression tests pass


stscijgbot:
Issue [JP-1224](https://jira.stsci.edu/browse/JP-1224) was created by Steven Crawford:

For the new regression test system, make sure that all of the tests are passing

stscijgbot:
Comment by Jonathan Eisenhamer: h2. Run #109
 * jwst.regtest.test_lrs_slit_spec2: Bad CRDS flat_field retrieval. [~eisenham] cannot reproduce.
 * jwst.regtest.test_miri_image_detector1: Missing artifactory file. Issue is capitalization of the filename. [~eisenham] capitalized the artifactory files.
 * jwst.regtest.test_nirspec_ifu: [~eisenham] is currently testing.

stscijgbot:
Comment by Jonathan Eisenhamer: h2. Run #109
 * jwst.regtest.test_lrs_slit_spec2: Bad CRDS flat_field retrieval. [~eisenham] cannot reproduce; tests pass.
 * jwst.regtest.test_miri_image_detector1: Missing artifactory file. Issue is capitalization of the filename. [~eisenham] capitalized the artifactory files.
 * jwst.regtest.test_nirspec_ifu: [~eisenham] cannot reproduce; tests pass.
 * All other tests are fitsdiff issues, presumably all which can be okified.

stscijgbot:
Comment by Jonathan Eisenhamer: h2. Run #109
 * jwst.regtest.test_lrs_slit_spec2: Bad CRDS flat_field retrieval. [~eisenham] cannot reproduce; tests pass.
 * jwst.regtest.test_miri_image_detector1: Missing artifactory file. Issue is capitalization of the filename. [~eisenham] capitalized the artifactory files.
 * jwst.regtest.test_nirspec_ifu: [~eisenham] still testing...
 * All other tests are fitsdiff issues, presumably all which can be okified.

stscijgbot:
Comment by Jonathan Eisenhamer: h2. Run #109
 * jwst.regtest.test_lrs_slit_spec2: Bad CRDS flat_field retrieval. [~eisenham] cannot reproduce; tests pass.
 * jwst.regtest.test_miri_image_detector1: Missing artifactory file. Issue is capitalization of the filename. [~eisenham] capitalized the artifactory files.
 * jwst.regtest.test_nirspec_ifu: [~eisenham] reproducible and needs to be examined.
[...]

PR #4464, opened 2020-01-21
Get new regression tests to pass

jdavies-st:
This does the following:

 - Add `NAXIS1` and `TFORM*` to keywords and fields to ignore in the `fitsdiff_default_kwargs` fixture
 - Update the `diff_astropy_tables` fixture to show the differences detected.  And use it where there were failing tests with `astropy.table.setdiff` before.
 - Fix fixture return values in `test_miri_lrs_slitless.py` module
- Fix bugs in `test_nircam_setpointing` and update data on artifactory

Addresses #4429


jwst/regtest/conftest.py
```diff
@@ -193,55 +194,63 @@ def rtdata_module(artifactory_repos, envopt, request):
...
-    def _diff_astropy_tables(actual_path, truth_path, rtol=0.00001, atol=0.0000001):
-        actual = Table.read(actual_path)
+    def _diff_astropy_tables(result_path, truth_path, rtol=1e-5, atol=1e-7):
+        result = Table.read(result_path)
         truth = Table.read(truth_path)
 
         diffs = []
 
-        if actual.colnames != truth.colnames:
+        if truth.colnames != truth.colnames:
```
hbushouse: Shouldn't this be ``result.colnames`` on the left side?

jwst/regtest/conftest.py
```diff
@@ -193,55 +194,63 @@ def rtdata_module(artifactory_repos, envopt, request):
...
             diffs.append("Column names (or order) do not match")
 
-        if len(actual) != len(truth):
+        if len(truth) != len(truth):
```
hbushouse: Again, shouldn't this be ``len(result)`` on the left side?

jwst/regtest/conftest.py
```diff
@@ -193,55 +194,63 @@ def rtdata_module(artifactory_repos, envopt, request):
...
             diffs.append("Row count does not match")
 
         # If either the columns or the row count is mismatched, then don't
         # bother checking the individual column values.
         if len(diffs) > 0:
             return diffs
 
-        if actual.meta != truth.meta:
+        if truth.meta != truth.meta:
```
hbushouse: ``result.meta``

jwst/regtest/conftest.py
```diff
@@ -193,55 +194,63 @@ def rtdata_module(artifactory_repos, envopt, request):
...
-                    truth[col_name], actual[col_name], rtol=rtol, atol=atol
-                ):
-                    diffs.append(
-                        "Column '{col_name}' values do not match (within tolerances)"
-                    )
-            else:
-                if not (actual[col_name] == truth[col_name]).all():
-                    diffs.append("Column '{col_name}' values do not match")
+            try:
+                if truth[col_name].dtype != truth[col_name].dtype:
```
hbushouse: ``result[col_name]`` on the left?

jwst/regtest/conftest.py
```diff
@@ -193,55 +194,63 @@ def rtdata_module(artifactory_repos, envopt, request):
...
+                    diffs.append(f"Column '{col_name}' dtype does not match")
+                    continue
+
+                dtype = truth[col_name].dtype
+                if dtype.kind == "f":
+                    try:
+                        assert_allclose(truth[col_name], truth[col_name],
```
hbushouse: ``result[col_name]`` as first arg?

jwst/regtest/conftest.py
```diff
@@ -193,55 +194,63 @@ def rtdata_module(artifactory_repos, envopt, request):
...
+                            rtol=rtol, atol=atol)
+                    except AssertionError as err:
+                        diffs.append(
+                            f"Column '{col_name}' values do not match (within tolerances) \n{str(err)}"
+                        )
+                else:
+                    if not (truth[col_name] == truth[col_name]).all():
```
hbushouse: ``result[col_name]`` on left

jwst/regtest/test_miri_image.py
```diff
@@ -66,10 +66,12 @@ def run_pipelines(jail, rtdata_module):
...
 @pytest.mark.bigdata
-@pytest.mark.parametrize("suffix", ["a3001_crf", "assign_wcs", "cal", "dark_current",
-    "dq_init", "firstframe", "flat_field", "i2d", "lastframe", "linearity", "ramp",
-    "rate", "rateints", "refpix", "rscd", "saturation"])
-def test_miri_image_stages12(run_pipelines, fitsdiff_default_kwargs, suffix):
+@pytest.mark.parametrize("suffix", ["dq_init", "saturation", "refpix", "rscd",
+    "firstframe", "lastframe", "linearity", "dark_current", "ramp", "rate",
+    "rateints",
+    "assign_wcs", "flat_field", "cal", "i2d",
+    "a3001_crf"])
```
hbushouse: Just curious, is this ordering necessary so that it first tests all the results from ``detector1``, then the results from ``image2``, and finally the results from ``image3``? Otherwise it would be doing comparisons of files that hadn't been regenerated yet?

jdavies-st: No, not necessary, just a little clearer to me the reader as to what is getting tested.  I was confused as to why `a3001_crf` was getting tested at all, but then realized it was the stage3 processing that was producing this (and due to the non-standard naming) putting the asn id at the end.

jwst/regtest/conftest.py
jdavies-st: Yes to all of these.  Thanks for catching.  Fixing all of them.

jdavies-st: Also, `flake8` caught this bug in the CI for this PR:

https://travis-ci.org/spacetelescope/jwst/jobs/640132790#L595

PR #4473, opened 2020-01-22
JP-1218 Move association tests from tests_nightly to regtest

### LINKED ISSUE #4425: Move associations to the new regression suite
Issue [JP-1218](https://jira.stsci.edu/browse/JP-1218) was created by Steven Crawford:

Move associations to the new regression suite

ddavis-stsci:
The association tests from tests_nightly have been moved to the regtest test area for the new version of the regression tests. 

Resolves #4425


jwst/regtest/conftest.py
```diff
@@ -10,6 +10,7 @@
 from numpy.testing import assert_allclose
 
 from .regtestdata import RegtestData
+from jwst.tests_nightly.general.associations.sdp_pools_source import SDPPoolsSource
```
jdavies-st: This import should be
```python
from jwst.regtest.sdp_pools_source import SDPPoolsSource
```

ddavis-stsci: good catch! fixed.

ddavis-stsci: test names have been updated.

PR #4479, opened 2020-01-24
Modify test to create and compare ramp_fit optional product.

dmggh:
There is no existing test that creates and compares the ramp_fit optional output product, so modify an existing test to do so. Addresses JP-1236.

### DIFF
--- a/jwst/regtest/test_miri_image_detector1.py
+++ b/jwst/regtest/test_miri_image_detector1.py
@@ -23,7 +23,8 @@ def run_pipeline(jail, rtdata_module):
             "--steps.dark_current.save_results=True",
             "--steps.refpix.save_results=True",
             "--steps.jump.rejection_threshold=25",
-            "--steps.jump.save_results=True"]
+            "--steps.jump.save_results=True",
+            "--steps.ramp_fit.save_results=True"]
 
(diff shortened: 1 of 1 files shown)


jwst/regtest/test_miri_image_detector1.py
```diff
@@ -23,7 +23,9 @@ def run_pipeline(jail, rtdata_module):
             "--steps.dark_current.save_results=True",
             "--steps.refpix.save_results=True",
             "--steps.jump.rejection_threshold=25",
-            "--steps.jump.save_results=True"]
+            "--steps.jump.save_results=True",
+            "--steps.ramp_fit.save_opt=True",
+            "--steps.ramp_fit.save_results=True"]
```
hbushouse: Just curious: do both ``save_results`` and ``save_opt`` need to be turned on in order to get the optional product created? Or would it get created by just having ``save_opt`` set?

dmggh: Both save_results and save_opt need to be turned on in order to get the optional product calculated and saved. Setting save_opt causes the optional arrays to be created (but not saved), so the parameter name is a bit misleading. I think I implemented (and named) the step-specific save_opt before the more general save_results was available.

PR #4490, opened 2020-01-28
JP-1234: Add MIRI image2 pipeline wcs regression test

jemorrison:
I updated the first test for the WCS - imager.
I am still working on the other modes and will add those to this PR soon. 
I just wanted to make sure this is all that was needed for WCS testing. 
@hbushouse @jdavies-st

### DIFF
--- a/jwst/regtest/test_miri_image.py
+++ b/jwst/regtest/test_miri_image.py
@@ -4,7 +4,7 @@ from numpy.testing import assert_allclose
 from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
 from gwcs.wcstools import grid_from_bounding_box
 from jwst.stpipe import Step
-from jwst.datamodels import ImageModel
+from jwst  import datamodels
 
 @pytest.fixture(scope="module")
 def run_pipelines(jail, rtdata_module):
@@ -116,15 +116,20 @@ def test_miri_image_wcs(run_pipelines, fitsdiff_default_kwargs):
     rtdata = run_pipelines
 
     # get input assign_wcs and truth file
-    output = "det_image_1_MIRIMAGE_F770Wexp1_5stars_assign_wcs.fits"
-    truth_file = rtdata.get_truth('truth/test_miri_image_stages/'+output)
-
+    rtdata.output = "det_image_1_MIRIMAGE_F770Wexp1_5stars_assign_wcs.fits"
+    rtdata.get_truth("truth/test_miri_image_stages/" + rtdata.output)
     # Open the output and truth file
-    im = ImageModel(output)
-    im_ref = ImageModel(truth_file)
+    im = datamodels.open(rtdata.output)
+    im_truth = datamodels.open(rtdata.truth)
 
     x, y = grid_from_bounding_box(im.meta.wcs.bounding_box)
     ra, dec = im.meta.wcs(x, y)
-    raref, decref = im_ref.meta.wcs(x, y)
-    assert_allclose(ra, raref)
-    assert_allclose(dec, decref)
+    ratruth, dectruth = im_truth.meta.wcs(x, y)
+    assert_allclose(ra, ratruth)
+    assert_allclose(dec, dectruth)
+
+    # Test the inverse transform
+    xtest, ytest = im.meta.wcs.backward_transform(ra, dec)
+    xtruth, ytruth = im_truth.meta.wcs.backward_transform (ratruth, dectruth)
+    assert_allclose(xtest, xtruth)
+    assert_allclose(ytest, ytruth)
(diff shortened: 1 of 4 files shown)


jemorrison:
I could add a test to modes that use ra, dec to x,y - which would be those mode that use outlier detection. I think that is just Imager and MRS - I don't think the LRS uses outlier detection in that way.

hbushouse:
> I could add a test to modes that use ra, dec to x,y - which would be those mode that use outlier detection. I think that is just Imager and MRS - I don't think the LRS uses outlier detection in that way.

LRS fixed-slit should use WCS transforms in ``outlier_detection``, because the data are resampled. But LRS slitless does not. Slitless data are analyzed in detector space (no resampling) when doing ``outlier_detection``.

jwst/regtest/test_miri_image.py
```diff
@@ -108,3 +109,22 @@ def test_miri_image_stage3_catalog(run_pipelines, diff_astropy_tables):
...
     diff = diff_astropy_tables(rtdata.output, rtdata.truth, rtol=1e-4)
     assert len(diff) == 0, "\n".join(diff)
+
+
+@pytest.mark.bigdata
+def test_miri_image_wcs(run_pipelines, fitsdiff_default_kwargs):
+    rtdata = run_pipelines
+
+    # get input assign_wcs and truth file
+    output = "det_image_1_MIRIMAGE_F770Wexp1_5stars_assign_wcs.fits"
```
jdavies-st: This needs to be `rtdata.output = "det_image_1_MIRIMAGE_F770Wexp1_5stars_assign_wcs.fits"`

jwst/regtest/test_miri_image.py
```diff
@@ -108,3 +109,22 @@ def test_miri_image_stage3_catalog(run_pipelines, diff_astropy_tables):
...
+    truth_file = rtdata.get_truth('truth/test_miri_image_stages/'+output)
```
jdavies-st: `rtdata.get_truth` populates `rtdata.truth` automatically.  So no need for assignment here.  Just use rtdata.truth below for the path.

jwst/regtest/test_miri_image.py
```diff
@@ -108,3 +109,22 @@ def test_miri_image_stage3_catalog(run_pipelines, diff_astropy_tables):
...
+
+    # Open the output and truth file
+    im = ImageModel(output)
+    im_ref = ImageModel(truth_file)
```
jdavies-st: These should be `datamodels.open()` calls.  And it would be nice to avoid "ref" in variable names as it is ambiguous between truth files and CRDS reference files.  This is the reason to use "truth".

jwst/regtest/test_miri_lrs_slit_spec2.py
```diff
@@ -2,8 +2,11 @@
 import pytest
 from astropy.io.fits.diff import FITSDiff
 
+from numpy.testing import assert_allclose
 from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
 from jwst.stpipe import Step
+from jwst  import datamodels
```
hbushouse: flake8 might complain about the extra whitespace in "jwst  import"

jwst/regtest/test_miri_lrs_slit_spec2.py
```diff
@@ -33,7 +37,7 @@ def run_pipeline(jail, rtdata_module):
 
 @pytest.mark.bigdata
 @pytest.mark.parametrize("output",[
-    "bsub", "flat_field", "srctype", "cal", "x1d"])
+    "bsub", "flat_field", "assign_wcs","srctype", "cal", "x1d"])
```
hbushouse: flake8 will complain about the lack of whitespace between "assign_wcs" and "srctype"

jwst/regtest/test_miri_lrs_slitless.py
```diff
@@ -1,9 +1,12 @@
 import pytest
 from astropy.io.fits.diff import FITSDiff
+from numpy.testing import assert_allclose
 
 from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
 from jwst.stpipe import Step
+from gwcs.wcstools import grid_from_bounding_box
 from jwst.associations.asn_from_list import asn_from_list
+from jwst  import datamodels
```
hbushouse: extra whitespace between "jwst" and "import"

jwst/regtest/test_miri_mrs.py
```diff
@@ -2,9 +2,11 @@
 from pathlib import Path
 import pytest
 
+from numpy.testing import assert_allclose
 from jwst.associations import load_asn
 from jwst.lib.suffix import replace_suffix
-
+from jwst  import datamodels
```
hbushouse: extra whitespace

jdavies-st:
Looks great!  👍

PR #4491, opened 2020-01-29
Remove TPCorr

mcara:
This PR removes `TPCorr` WCS correction model. This model is now a compound model defined in `tweakwcs` using fundamental models defined in `astropy` and `gwcs`.

Until a new release of `gwcs` is made (and `tweakwcs`), `jwst` will need to depend on `dev` versions of these two packages if this PR is merged.

CC: @nden @hbushouse @jdavies-st

PR #4497, opened 2020-01-29
Revert "Remove TPCorr"

mcara:
Reverts spacetelescope/jwst#4491

Until `asdf` validation error is understood.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -63,12 +63,6 @@ stpipe
 
 - Fix sub-step nesting in parameter reference files [#4488]
 
-transforms
-----------
-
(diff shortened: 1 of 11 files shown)


mcara:
@jdavies-st If it is no rush, give me a few hours to see if I can fix the issue.

mcara:
Let's delay merging this for 1 day. I believe the workaround from https://github.com/spacetelescope/tweakwcs/issues/97#issuecomment-579992134 works. However, now tests fail because of different angle wrapping definition. I will try to have a PR tomorrow that takes care of these issues.

PR #4506, opened 2020-01-31
Added pytest_astropy_header to the README.  It is needed for tests to…

blakem:
… run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Note: If you wish to install directly from github, but also include the extra de
 
 Need other useful packages in your development environment?
 
-    pip install ipython flake8 pytest-xdist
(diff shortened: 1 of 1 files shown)


stscieisenhamer:
This may be a useful plugin, but it should not be a dependency in the code. How did you run the tests and at what point did the tests fail because of not having this?

jdavies-st:
No, it is not needed for tests to run.  There is currently an upstream dependency bug in `gwcs` that will produce a warning if `pytest-remotedata` plugin is not installed.  See

https://github.com/spacetelescope/gwcs/issues/279

But this is not an issue with `jwst`.  Is there something else we are missing?

jdavies-st:
Closing this as there does not seem to be an issue.

PR #4519, opened 2020-02-03
Only do post-regtest failure output upload if file exists

jdavies-st:
- Prevent Jenkins RT build failures when a test fails when pipeline is running and `rtdata.output` is either undefined or the file it refers to does not exist for some reason.
- Upload the `_okify.json` and `_rtdata.asdf` files to `jwst-pipeline-results` Artifactory repo for the OKify script to use.


jdavies-st:
So coverage has gone down slightly, but I can't figure out a way to test this using `pytest`, other than to just verify in practice that it works.

PR #4521, opened 2020-02-05
JP-1128 Do not overwrite rules with defaults

### LINKED ISSUE #4238: model_blender ignores anyOf items in the schemas
Issue [JP-1128](https://jira.stsci.edu/browse/JP-1128) was created by James Davies:

Any datamodel keyword that uses “anyOf” in its schema gets excluded from the the blend header table.  Currently this affects `CORONMSK`, `CHANNEL` and `SUBARRAY` keywords.

stscieisenhamer:
Resolves #4238/[JP-1128](https://jira.stsci.edu/browse/JP-1128).

### DIFF
--- a/jwst/model_blender/tests/test_blend.py
+++ b/jwst/model_blender/tests/test_blend.py
@@ -1,5 +1,5 @@
-import shutil
-import tempfile
+"""Test blend_models"""
+import pytest
 
 import numpy as np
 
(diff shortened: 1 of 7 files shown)


jwst/model_blender/tests/test_blend.py
```diff
@@ -1,55 +1,130 @@
...
-    shutil.rmtree(TMP_DIR)
-
-
-def test_blendmeta():
-    newmeta, newtab = blendmeta.get_blended_metadata(TMP_FILES)
-    for attr in INPUT_VALUES:
-        assert newmeta[attr] == OUTPUT_VALUES[attr]
+DATES = ['2017-11-30', '2017-11-11', '2017-12-10']
+INSTRUMENT_NAMES = ['NIRCAM'] * 3
+CORONAGRAPHS = ['4QPM', '4QPM_1065', '4QPM_1140']
```
jdavies-st: Which one of the above is the `anyOf`?  CORONOGRAPHS?

stscieisenhamer: Yep, CORONAGRAPHS was one of the offending parameters. I'll actually see if others exist with a different set of rules actually.

About `jail`, is there one lagging about? I did remove them and can't find any in `test_blend` module.

jdavies-st: Ah I see.  Yes, none are there, but you have moved the `jail` fixture in conftest.py up one level, which made me think you are using it here.

stscieisenhamer: Ah right, short-attention-span-theater: Yes, it was going to be used, but when the specific test didn't require it, forgot about it. I can back that out if desired.

PR #4541, opened 2020-02-07
JP-1266: Update NRS_FIXEDSLIT handling in SRCTYPE step

### LINKED ISSUE #4520: source_type and pathloss inconsistency for NIRSpec FS exposures
Issue [JP-1266](https://jira.stsci.edu/browse/JP-1266) was created by Howard Bushouse:

For NIRSpec fixed-slit (FS) exposures, the source_type step only populates the source type in the model.meta.target.source_type attribute, because there's only 1 value available for this type of exposure (there's no per-slit information, like there is for MOS mode). But when pathloss does its processing of NIRSpec FS exposures, it looks for the source type that's stored in each slit instance of the data model, i.e. slit.source_type. So the pathloss step is finding source types of None for all the slits and hence not applying the correct processing.

The source_type step should be updated to populate the slit.source_type attribute for each slit contained in the exposure.

hbushouse:
Updated the `srctype` step logic to handle input exposures that contain one or more slit instances. Previously only the SRCTYPE value in the primary header had been updated. Now the SRCTYPE value in each slit instance is also updated. For NIRSpec fixed-slit mode in particular, only the slit instance for the primary slit is updated with the SRCTYPE value derived by the step. Any additional slit instances get the default value for NRS_FIXEDSLIT, which is "POINT".

Updated the docs to reflect these changes.

Also a few minor updates to fix things in other docs (e.g. the `IFUCubeParsModel` was removed from datamodels in #4313, but had not been removed from the docs), and a few minor updates to log messages in other steps, just to make all steps consistent in their logging.

Fixes #4520 / [JP-1266](https://jira.stsci.edu/browse/JP-1266)


hbushouse:
Tests of the updated code have been run on NRC_TSGRISM, NIS_SOSS, NRS_BRIGHTOBJ, and different flavors of NRS_FIXEDSLIT (e.g. an ALLSLITS exposure, an exposure with a single slit, and an exposure where FXD_SLIT='NONE'). All cases work as desired.

jwst/srctype/srctype.py
```diff
@@ -45,6 +47,10 @@ def set_source_type(input_model):
             src_type = 'EXTENDED'
             log.info('Exposure is a background target; setting SRCTYPE = %s' % src_type)
 
+        elif pipe_utils.is_tso(input_model):
```
philhodge: I'm surprised that the user-supplied value (if any) doesn't have higher priority.  That's OK; I'm just surprised.

jwst/srctype/srctype.py
```diff
@@ -67,8 +73,30 @@ def set_source_type(input_model):
...
+            # value, which for NRS_FIXEDSLIT is 'POINT'.
+            default_type = 'POINT'
+            primary_slit = input_model.meta.instrument.fixed_slit
+            log.debug(' primary_slit = {}'.format(primary_slit))
+            for slit in input_model.slits:
+                if slit.name == primary_slit:
+                    slit.source_type = src_type
+                else:
+                    slit.source_type = default_type
+                log.debug(' slit {} = {}'.format(slit.name,slit.source_type))
```
philhodge: You might need a space after the comma.

docs/jwst/srctype/description.rst
```diff
@@ -5,19 +5,23 @@ determine whether a spectroscopic source should be considered to be a point
...
+For JWST observing modes that use a single primary target (e.g. MIRI MRS
+and LRS spectroscopy and NIRSpec IFU and Fixed-Slit spectroscopy), the observer
+has the option of designating a source type in the APT template for the
+observation. They have the choice of declaring whether or not the source
+should be considered extended. If they don't know the character of the source,
+they can also choose a value of "UNKNOWN." The observer's choice is passed along
 to DMS processing, which sets the value of the "SRCTYPE" keyword in the
 primary header of the uncalibrated (_uncal.fits) product that's used as input
 to the calibration pipeline. If the user has selected a value in the APT, the
 "SRCTYPE" keyword will be set to "POINT", "EXTENDED", or "UNKNOWN." If the
```
philhodge: Does this mean that SRCTYPE will be set to the value the use selected?

hbushouse: Yes. I could perhaps change the wording to make that more explicit.

jwst/srctype/srctype.py
hbushouse: I don't see any flake8 complaints, but I'll fix it anyway, just for readability.

hbushouse:
This will definitely change some regression test results, because the slit.source_type values were not populated before, in which case the ``pathloss` step assumed a value of EXTENDED. Now many of those slit instances will be set to the NRS_FIXEDSLIT default value of POINT, hence leading to differences in ``pathloss` results.

PR #4542, opened 2020-02-07
Remove TPCorr. V2

mcara:
Reverts spacetelescope/jwst#4497

From #4491 

> This PR removes `TPCorr` WCS correction model. This model is now a compound model defined in `tweakwcs` using fundamental models defined in `astropy` and `gwcs`.
> 
> Until a new release of `gwcs` is made (and `tweakwcs`), `jwst` will need to depend on `dev` versions of these two packages if this PR is merged.


jdavies-st:
I have a branch that fixes the MIRI imaging regression test (by allowing tweakwcs to work with only 5 stars) to actually run `TweakRegStep` and not skip it.  Let me run that too.

mcara:
In https://github.com/spacetelescope/tweakwcs/pull/96 I noticed some degradation of accuracy and I had to reduce the tolerance requirements in some unit tests to get them to pass. Just one example:

https://github.com/spacetelescope/tweakwcs/pull/96/files#diff-7965be8489a50e50218b20aa8de47426R32

(but there are many more cases when I had to increase tolerance by a factor of 100). I have no immediate explanation for this (no obvious reason). The changes that you are seeing are probably due to this accuracy loss.

I would prefer to understand the cause of this accuracy loss better and so if this PR can wait a little it would be great.

nden:
This can be merged after #4741 is merged and after a reabse.

setup.py
```diff
@@ -93,15 +93,15 @@ def get_transforms_data():
         'astropy>=4.0',
         'crds>=7.2.7',
         'drizzle>=1.13',
-        'gwcs>=0.12',
+        'gwcs @ git+https://github.com/spacetelescope/gwcs@1c1cb3bb35caddef80fb760ea68bc71e189d32de',
```
jdavies-st: This should be changed to `gwcs>=0.13`.

setup.py
```diff
@@ -93,15 +93,15 @@ def get_transforms_data():
...
         'jsonschema>=2.3,<4',
         'numpy>=1.16',
         'photutils>=0.7',
         'scipy>=1.0',
         'spherical-geometry>=1.2',
         'stsci.image>=2.3.3',
         'stsci.imagestats>=1.4',
-        'tweakwcs>=0.5.3',
+        'tweakwcs @ git+https://github.com/spacetelescope/tweakwcs@553ebd51262ca980c58aaf0a980c9cf746c08bb4',
```
jdavies-st: This should be changed to `tweakwcs>=0.6`?  Or whichever minimum version is appropriate.

mcara:
> I don't think we want to remove tpcorr ASDF type, schema and transform from our ASDF extension. It needs to still be able to read old data, right? Or are we happy dropping support for reading old Build 7.4.2 or earlier data? @nden @hbushouse

> If so, then no problem dropping them.

@jdavies-st @nden @hbushouse What is the conclusion on this? Do we keep `TPCorr` around? Would removing it create issues with old data?

hbushouse:
Is `TPCorr` a model that's only ever used within the WCS transforms computed by the `tweakreg` step? If so, then it only appears in data files that've previously been through `calwebb_image3` processing. We typically don't even save the intermediate (individual) exposure files that've been through `tweakreg`, we only save the final combined image that's built from the individual corrected exposures. And I'm assuming `TPCorr` is irrelevant and/or non-existent in those combined images, right? If all of the above is correct, then I don't there's really any old data with `TPCorr` that would ever need to be reloaded into any pipelines or steps. At worst, the data would just have to be reprocessed through `calwebb_image3` again.

Someone please correct me if my understanding is incomplete.

mcara:
@hbushouse Thank you

jdavies-st:
Agree with Howard.  The only files it will appear in are the `_crf` files, and those can be reprocessed.  Let's remove it.

mcara:
I messed-up things trying to resolve conflicts. I will close this PR and a new PR was created: https://github.com/spacetelescope/jwst/pull/4790

PR #4547, opened 2020-02-07
JP-1287: Fix a variety of cube_build and IFU outlier rejection issues

### LINKED ISSUE #4548: Bugs in IFU outlier rejection
Issue [JP-1287](https://jira.stsci.edu/browse/JP-1287) was created by David Law:

Ongoing testing of cube building and outlier rejection for the MIRI MRS has revealed a few issues.  In diagnosing and investigating these issues I've also implemented a variety of fixes for them in pull request https://github.com/spacetelescope/jwst/pull/4547.  Issues include:
 * Outlier rejection cannot call cube_build with 'emsm' weighting but is forced to use 'msm'.  I've modified the code to use 'emsm' for outlier rejection by default for the MIRI MRS, and ensure this doesn't get overwritten by hard-coded 'msm'.
 * The blotting stage has missing 'stripes' of detector pixels with zero value in the blotted data.  I've increased the ROI for blotting slightly to fix this.
 * Blotting fails if more than 2 detector pixels are within range of a cube spaxel because of an array size casting error.  I've rewritten the indexing function to concatenate x/y matches for both arrays consistently.
 * Blotting shifts input data by 1 pixel in x/y from the IFU cube, resulting in incorrect trace locations on the detector.  Not entirely sure why this is happening, but since the cube WCS looks ok I've applied a shift to the transform that fixes the problem in test data.
 * Cube building does not propagate ERR information from the 2d calibrated frames into the 3d data cubes.  I've installed and activated the plumbing to pass ERR through to output cubes.
 * Cube building should not multiply spatial ROI by 1.5 for single-exposure cubes if using 'emsm' weighting.  I've deactivated this multiplication if weighting='emsm'

Even with the above outlier rejection isn't rejecting outliers as expected, still investigating.

drlaw1558:
This pr addresses a variety of issues found in https://jira.stsci.edu/browse/JP-1287
Namely, 
* Outlier rejection cannot call cube_build with 'emsm' weighting but is forced to use 'msm'.  I've modified the code to use 'emsm' for outlier rejection by default for the MIRI MRS, and ensure this doesn't get overwritten by hard-coded 'msm'.
* The blotting stage has missing 'stripes' of detector pixels with zero value in the blotted data.  I've increased the ROI for blotting slightly to fix this.
* Blotting fails if more than 2 detector pixels are within range of a cube spaxel because of an array size casting error.  I've rewritten the indexing function to concatenate x/y matches for both arrays consistently.
* Blotting shifts input data by 1 pixel in x/y from the IFU cube, resulting in incorrect trace locations on the detector.  Not entirely sure why this is happening, but since the cube WCS looks ok I've applied a shift to the transform that fixes the problem in test data.
* Cube building does not propagate ERR information from the 2d calibrated frames into the 3d data cubes.  I've installed and activated the plumbing to pass ERR through to output cubes.
* Cube building should not multiply spatial ROI by 1.5 for single-exposure cubes if using 'emsm' weighting.  I've deactivated this multiplication if weighting='emsm'

Fixes #4548

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ associations
 - Update act_id format to allow base 36 values in product name [#4282]
 
 - Refactor association logging configuration [#4510]
+
 combine_1d
 ----------
 
@@ -48,6 +49,11 @@ master_background
 -----------------
 
 - Updated to fill the asn table and asn pool names. [#4240]
+
+model_blender
+-------------
+
+- Do not overwrite rules with defaults. [#4521]
 
 pipeline
 --------
--- a/jwst/cube_build/blot_cube_build.py
+++ b/jwst/cube_build/blot_cube_build.py
@@ -65,7 +65,7 @@ class CubeBlot():
 
         # using wcs of ifu cube determine ra,dec,lambda
         self.cube_ra, self.cube_dec, self.cube_wave = \
-            self.median_skycube.meta.wcs(xcube, ycube, zcube)
+            self.median_skycube.meta.wcs(xcube + 1, ycube + 1, zcube + 1)
 
         # pull out flux from the median sky cube that matches with
         # cube_ra,dec,wave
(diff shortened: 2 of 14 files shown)


drlaw1558:
Added an extra commit to insert pixel area header keywords required by extract_1d routine.

PR #4551, opened 2020-02-10
JP-952 Skip reading and validating non-datamodel members

ddavis-stsci:
Restricted the import and verification of association exptypes to those known about in dms_base to prevent the code from reading and trying to verify files that have no datamodel associated with them. 

local pytest results:
```
pytest --bigdata --basetemp=/Users/ddavis/src/JWST/develop/JP-952/Feb10/ ~/src/JWST/scsb/jwst/jwst/regtest/ -k test_associations 2>&1 | tee test_Feb10ab.log
============================= test session starts ==============================
platform darwin -- Python 3.7.4, pytest-5.3.5, py-1.8.1, pluggy-0.13.0
plugins: xdist-1.30.0, requests-mock-1.7.0, doctestplus-0.4.0, openfiles-0.4.0, ci-watson-0.4, asdf-2.5.1, forked-1.1.3, cov-2.8.1
collected 363 items / 282 deselected / 81 selected

../../scsb/jwst/jwst/regtest/test_associations_generate.py ss            [  2%]
../../scsb/jwst/jwst/regtest/test_associations_io.py ssss                [  7%]
../../scsb/jwst/jwst/regtest/test_associations_level3_product_names.py s [  8%]
                                                                         [  8%]
../../scsb/jwst/jwst/regtest/test_associations_main.py ssss              [ 13%]
../../scsb/jwst/jwst/regtest/test_associations_sdp_pools.py ......s...s. [ 28%]
s..ss....s.s.......xx..s...s...ss                                        [ 69%]
../../scsb/jwst/jwst/regtest/test_associations_standards.py ............ [ 83%]
............s                                                            [100%]

==== 56 passed, 23 skipped, 282 deselected, 2 xfailed in 1354.67s (0:22:34) ====
```

should also point out that the test case now works,
```
>>> from jwst import datamodels
>>> a = datamodels.open('test_asn.json')
>>>

```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,8 @@ combine_1d
 datamodels
 ----------
 
+- Skip reading and validating non-datamodel members [#3946]
+
 - Force data model type setting on save [#4318]
 
 - Deprecate ``MIRIRampModel`` [#4328]
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -8,6 +8,14 @@ import logging
 from asdf import AsdfFile
 from astropy.io import fits
 
+from jwst.associations.lib.dms_base import (
+    CORON_EXP_TYPES,
+    IMAGE2_NONSCIENCE_EXP_TYPES,
+    IMAGE2_SCIENCE_EXP_TYPES,
+    SPEC2_SCIENCE_EXP_TYPES,
+    ACQ_EXP_TYPES
+)
+
 from ..associations import (
     AssociationNotValidError,
     load_asn)
@@ -106,6 +114,12 @@ class ModelContainer(model_base.DataModel):
         elif is_association(init):
             self.from_asn(init)
         elif isinstance(init, str):
+            self.asn_exptypes = (CORON_EXP_TYPES +
+                                 IMAGE2_NONSCIENCE_EXP_TYPES +
+                                 IMAGE2_SCIENCE_EXP_TYPES +
+                                 SPEC2_SCIENCE_EXP_TYPES +
+                                 list(ACQ_EXP_TYPES))
+
             init_from_asn = self.read_asn(init)
             self.from_asn(init_from_asn, asn_file_path=init)
         else:


jwst/datamodels/container.py
```diff
@@ -106,6 +114,12 @@ def __init__(self, init=None, asn_exptypes=None, **kwargs):
         elif is_association(init):
             self.from_asn(init)
         elif isinstance(init, str):
+            self.asn_exptypes = (CORON_EXP_TYPES +
+                                 IMAGE2_NONSCIENCE_EXP_TYPES +
+                                 IMAGE2_SCIENCE_EXP_TYPES +
+                                 SPEC2_SCIENCE_EXP_TYPES +
+                                 list(ACQ_EXP_TYPES))
+
```
jdavies-st: Instead of being down here in the logic, defining `self.asn_exptypes` should be up top.  Maybe replace
```
self.asn_exptypes = asn_exptypes
```
with
```
self.asn_exptypes = asn_exptypes
if self.asn_exptypes is None:
    self.asn_exptypes = (CORON_EXP_TYPES + IMAGE2_NONSCIENCE_EXP_TYPES
    + IMAGE2_SCIENCE_EXP_TYPES + SPEC2_SCIENCE_EXP_TYPES
    + list(ACQ_EXP_TYPES))
```
or similar.  I.e. leave the possibility of specifying it on the `open()` call, but if not, then use the defaults.

Also, it looks like ACQ_EXP_TYPES is a tuple while the others are strings.  That looks like a bug.

CHANGES.rst
```diff
@@ -16,6 +16,8 @@ combine_1d
 datamodels
 ----------
 
+- Skip reading and validating non-datamodel members [#3946]
```
hbushouse: " ... non-datamodel members **of an association**"

jwst/datamodels/container.py
```diff
@@ -8,6 +8,14 @@
 from asdf import AsdfFile
 from astropy.io import fits
 
+from jwst.associations.lib.dms_base import (
+    CORON_EXP_TYPES,
+    IMAGE2_NONSCIENCE_EXP_TYPES,
+    IMAGE2_SCIENCE_EXP_TYPES,
+    SPEC2_SCIENCE_EXP_TYPES,
+    ACQ_EXP_TYPES
+)
```
hbushouse: These are all lists of EXP_TYPE values, which are recorded as keyword values in the headers of the member files. They aren't recorded in the ASN file itself. I'm guessing what you really meant is a list of ASN member types, which is recorded in the "exptype" attribute in the ASN json files, right? Those have values like "science", "background", "psf", "target-acquisition", etc., and I'm guessing are stored in some other macros somewhere within associations/lib.

hbushouse:
Is it really just non-datamodel members that we want to exclude, because that's what causes the error? For example, say at some point in the future we actually implement a data model for source catalogs (_cat.ecsv files). At that point it's fine to include them in the list of members to open and read when calling `datamodels.open`?

If so, then I agree with @stscieisenhamer that we shouldn't be doing checks based on either EXP_TYPE or member type ("exptype"). It should somehow be able to just test whether each thing in the input list is a data model or not.

jdavies-st:
The code for handling files of different types in `datamodels.open` is in `datamodels.filetype.check()`.  I suspect a solution lies there.  Handle the case of ECSV, and do the appropriate thing in whatever calls that function.

stscieisenhamer:
Won't fix. See originating issue JP-952 for details.

PR #4552, opened 2020-02-10
JP-1258: Remove  drizproduct and multiproduct

### LINKED ISSUE #4228: Build 7.3 resample_spec yields empty array
Issue [JP-1124](https://jira.stsci.edu/browse/JP-1124) was created by Beth Sargent:

While performing Build 7.3 testing of the extract1d step for MIRI LRS Fixed Slit data using the calwebb_spec3.cfg script (see JP-994), I found that the spectrum output by the extract1d step was zero at all wavelengths.  The simulations I used included signal in the _cal.fits files output by the calwebb_spec2.cfg script (which were input for the calwebb_spec3.cfg script), so the spectrum should have non-zero flux density values.

I checked the output of the preceding steps in the spec3 script.  For LRS Fixed Slit data, the step immediately preceding extract1d in spec3 is the resample_spec step.  The output of this step was an empty array, where all the pixels' values were equal to 0.  The step preceding this in spec3 for LRS Fixed Slit data is outlier_detection.  In this step, I did see signal like what I saw in the _cal.fits files that were output from the spec2 pipeline (and input for the spec3 pipeline).  So it appears that the resample_spec step is eliminating the signal in the data read in to it, so that it outputs an array with pixel values that are all zero.

### LINKED ISSUE #4507: Fix inconsistencies and problems in DrizProductModel and its use
Issue [JP-1258](https://jira.stsci.edu/browse/JP-1258) was created by Howard Bushouse:

We have several tickets open for problems that all trace back, in one way or another, to shortcomings or inconsistencies in the structure and content of the datamodels.DrizProductModel.

The most troublesome problem is the lack of certain meta data (keywords) not being defined in the DrizProductModel schema that are needed by some pipeline steps that use a DrizProductModel as input. These problems reported in JP-870 and JP-1144. A related ticket, JP-880, concerns meta data that is not or cannot be copied over from a DrizProductModel.

The schema for DrizProductModel also needs cleaning up, in order to remove ad hoc includes of sub-schemas like "lev3_prod.schema.yaml". The meta data contained in that schema should be reorganized to either move it to other schemas (e.g. core.schema) or create a more appropriate sub-schema that is relevant to all resampled products.

Related to this is simply the naming of DrizProductModel and the associated MultiProductModel. In particular, the name "MultiProductModel" is not very informative. Just what kind of "Product" is being referred to? We have lots of "products" created by the jwst pipeline. Both of these models should perhaps be renamed to something more obvious and suitable, such as "ResampledModel" and "MultiResampleModel" (other suggestions are welcome). And perhaps we may even need a model dedicated exclusively to resampled *spectral* data, because it typically requires lots of additional meta data (slit-related info) that is irrelevant to resampled *image* data.

jemorrison:
@hbushouse @jdavies-st @philhodge 
This PR is just to get show what I have done so far - and where I am stuck. It is not to be merged.
The plan is to use this new multiresamp model and delete the multiprod and drizproduct data models. 

I combined the drizproduct, multiprod  and lev3_prod into a single schema - multiresamp.schema
I probably should move the meta type data out and put it in the core schema  but first I need some help 
on how to define multiresamp.py the MultiResampModel.
How do I define it to make it be able to handle the  1 product case as we would have for imager resampled data  or the multiple products  case for multi slit data when both types of data are defined by multiresamp.schema (they just have single products or multiple products) ? It seems circular. 

Resolves #4507 / [JP-1258](https://jira.stsci.edu/browse/JP-1258)
Resolves #4228 / [JP-1124](https://jira.stsci.edu/browse/JP-1124)


jemorrison:
Second attempt to get this correct. I added a slitmeta.schema. 
@jdavies-st  I am not sure if I am defining the slitmeta schema correctly in 
multiresamp.schema or resamp.schema
Could you look at it and let know if that looks correct

jwst/datamodels/schemas/resamp.schema.yaml
```diff
@@ -0,0 +1,50 @@
...
+  properties:
+    meta:
+      type: object
+      properties:
+        bunit_data:
+          title: physical units of the array values
+          type: string
+          fits_hdu: SCI
+          fits_keyword: BUNIT
+        pointings:
```
hbushouse: It looks to me like 2 or 3 of these keywords are also used in `resample_spec` and hence presumably they get populated in what was the `MultiProductModel` (now `MultiResampModel`), so I guess we need these 3 keywords defined in `multiresamp.schema`. Or perhaps we can again consolidate things into one place and put all drizzle/resample-related keyword definitions in a sub-schema, like "resampmeta.schema", and include that sub-schema in both `resamp.schema` and `multiresamp.schema`.

jwst/datamodels/schemas/slitmeta.schema.yaml
```diff
@@ -0,0 +1,100 @@
...
+    xcen:
+      title: Center of shutter in MSA coordinates (X dir)
+      type: integer
+    ycen:
+      title: Center of shutter in MSA coordinates (Y dir)
+      type: integer
+- $ref: variance.schema
+- $ref: pathlosscorr.schema
+- $ref: bunit.schema
+- $ref: photometry.schema
```
hbushouse: `photometry.schema` and `wcsinfo.schema` are already included in the top-level schemas into which `slitmeta.schema` is being included, so we don't need/want them included here too. Same with `bunit.schema`, because the BUNIT keyword is already explicitly defined in the upper-level schemas.

jemorrison:
Added resampmeta.schema and added meta to both resampmeta.schema and slitmeta.schema have everything defined as meta data

jwst/datamodels/schemas/image.schema.yaml
```diff
@@ -7,6 +7,8 @@ allOf:
 - $ref: bunit.schema
 - $ref: photometry.schema
 - $ref: wcsinfo.schema
+- $ref: slitmeta.schema
```
jdavies-st: I don't think we want `slitmeta.schema` in `ImageModel`.

jwst/datamodels/schemas/slitmeta.schema.yaml
```diff
@@ -0,0 +1,98 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
+id: "http://stsci.edu/schemas/jwst_datamodel/slitmeta.schema"
+type: object
+  properties:
+    meta:
```
jdavies-st: I think moving all of these slit meta properties under `meta` is a big change that will probably break lots of code.  We should eventually do it, but probably in a different PR?  I believe there's another issue for this, and if not, we should file one.

jwst/datamodels/schemas/image.schema.yaml
jdavies-st: In fact, we probably want the `resampmeta.schema` here, right?  We also want `resampmeta.schema` included in the `slitdata.schema` I believe?

jwst/datamodels/schemas/resampmeta.schema.yaml
```diff
@@ -0,0 +1,27 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
+id: "http://stsci.edu/schemas/jwst_datamodel/resampmeta.schema"
+type: object
+  properties:
+    meta:
+      type: object
+      properties:
+        bunit_data:
```
hbushouse: We shouldn't need to have BUNIT defined explicitly here, because `bunit.schema` is already inherited in both `image.schema` and `slitdata.schema`.

Furthermore, we could easily move the remaining keywords - NDRIZ, TEXPTIME, and RESWHT - into `core.schema`. They could be put into a new sub-section of `core.schema`, entitled "Resampling parameters". They'll only show up in products when they've got values.

So basically we don't need `resampmeta.schema` at all.

jwst/datamodels/schemas/image.schema.yaml
hbushouse: See my comment below: I don't think we need `resampmeta.schema` at all. The few entries in it can be moved to `core.schema`.

hbushouse: I think we need `slitmeta.schema` included in `image.schema`, because there's at least one instance somewhere in the pipeline where a single instance of a spectroscopic slit gets resampled and was previously stored in a `DrizProductModel`. That was causing problems downstream, because `DrizProductModel` did not have slit-related keywords, like SLTNAME, which other code needed (that was the whole reason for this work to begin with).

I think it might the case where a NIRSpec fixed-slit exposure uses a subarray that only contains data from a single slit and that is/was getting resampled into a `DrizProductModel`, so presumably it would now get resampled into an `ImageModel`.

jwst/datamodels/schemas/resampmeta.schema.yaml
```diff
@@ -0,0 +1,27 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
+id: "http://stsci.edu/schemas/jwst_datamodel/resampmeta.schema"
+type: object
+  properties:
+    meta:
+      type: object
+      properties:
```
jdavies-st: This can become
```
allOf:
- type: object
   properties:
```

jwst/datamodels/schemas/slitmeta.schema.yaml
```diff
@@ -0,0 +1,98 @@
...
+      type: object
+      properties:
```
jdavies-st: This can become
```
allOf:
- type: object
   properties:
```

jwst/datamodels/schemas/image.schema.yaml
jdavies-st: Hmm, perhaps.  Maybe it should be resampled into a `SlitModel` instead?

jwst/datamodels/schemas/resampmeta.schema.yaml
jdavies-st: That makes sense.  Just make sure they map to `meta.resample.product_exposure_time`, `meta.resample.weight_type`, etc.  I.e. they should have a subheading `resample`.

jwst/datamodels/drizproduct.py
```diff
@@ -1,30 +0,0 @@
-from .model_base import DataModel
```
hbushouse: Need to also remove `DrizProductModel` and `MultiProductModel` from `datamodels/__init__.py`, as well as any occurrences in docs.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -2172,6 +2172,23 @@ properties:
             type: string
             fits_keyword: S_WHTLIT
             blend_table: True
+      resample:
+        title: Resample Parameters
+        type: object
+        properties:
+	  pointings:
```
hbushouse: What's the big colored bar here? Does that indicate use of a TAB or something? Check the formatting.

hbushouse: Oh, I know, the indenting is wrong. It needs to be indented another 4 spaces.

jwst/resample/resample_spec_step.py
```diff
@@ -43,7 +43,7 @@ def process(self, input):
         self.drizpars = kwargs
 
         if isinstance(input_models[0], MultiSlitModel):
-            # result is a MultiProductModel
+            # result is a MultiResampModel
```
hbushouse: All of these need to be changed to `MultiSlitModel`. We're not creating a `MultiResampModel` after all.

jwst/extract_1d/extract.py
```diff
@@ -2724,15 +2724,18 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
...
-        isinstance(input_model, datamodels.MultiSlitModel) or
-        isinstance(input_model, datamodels.MultiProductModel)):
+        isinstance(input_model, datamodels.MultiSlitModel):
+#        isinstance(input_model, datamodels.MultiSlitModel) or
+#        isinstance(input_model, datamodels.MultiProductModel)):
 
         if was_source_model:            # from a SourceModelContainer?
             slits = [input_model]
-        elif isinstance(input_model, datamodels.MultiSlitModel):
+        else isinstance(input_model, datamodels.MultiSlitModel):
```
hbushouse: This needs to remain `elif`, because it's got a clause following it. `else` can only be used "bare". This syntax error is causing all the CI tests to fail to build.

jwst/extract_1d/extract.py
```diff
@@ -2875,14 +2878,16 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
                         "so apply_nod_offset will be set to False",
                         source_type)
 
-        if isinstance(input_model, (datamodels.ImageModel,
-                                    datamodels.DrizProductModel)):
+#        if isinstance(input_model, (datamodels.ImageModel,
+#                                    datamodels.DrizProductModel)):
+
+        if isinstance(input_model, (datamodels.ImageModel)):
```
hbushouse: Don't need the extra parens around `datamodels.ImageModel`, because the argument now contains only 1 item.

jwst/datamodels/schemas/slitmeta.schema.yaml
```diff
@@ -0,0 +1,95 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
+id: "http://stsci.edu/schemas/jwst_datamodel/slitmeta.schema"
+type: object
+  properties:
```
hbushouse: I think you need to remove the 2 whitespaces at the beginning of this line, in order to get the syntax correct (and fix the unit test failures).

jemorrison: I have been struggling trying to get this line fixed in flake8. I added allOf at the top is that needed ?
So properties should not have 2 white spaces ? All the other schemas seem to have this. 
But flake8 does not like  line 6

jemorrison: Ok in the wcsinfo.schema there is no 'allOf' and there is not indenting of  of properties. I will try that format.

hbushouse:
https://github.com/spacetelescope/jwst/blob/master/jwst/ami/tests/test_ami_interface.py#L21 needs to be updated to create a `CubeModel`, not an `ImageModel` (an `ImageModel` can't be 3-D). That test was bad to begin with and only now showed up as such, due to our changes.

stscieisenhamer:
I've looked into the `append` issue: This actually is a fallout from removing the explicit creation of the `data` array: One cannot append to something that does not yet exist.

So @ddavis-stsci @eslavich: I seem to recall that there was a way of marking an item in the schema to not lazy instantiate? This would specifically be with the FITS HDU creation. Otherwise, we are going to have, in some places like `MultiSlitModel`, have the weird 'reference to create' statement.  For `MultiSlitModel`, the reference would just be `self.slits`.

jwst/datamodels/schemas/image.schema.yaml
philhodge: Currently in SlitDataModel, the attributes such as `source_type` are at the same level as `data`, `dq`, etc.  Is it the plan to retain this organization, or will `source_type`, etc., be moved under `meta`?

hbushouse: I think that the plan for now was to simply replicate the existing structure of MultiSlit. It's then a longer-term goal to fix all such instances of data models that have a list of objects so that all meta attributes for each instance are in fact under the "meta" hierarchy. Of course that will require a lot of step code updates too.

eslavich:
@stscieisenhamer there's something fishy going on in the `DataModel.update` method.  It's ending up setting the value for `slits` to an empty dict instead of an empty list, which later causes that call to `append` to fail.

stscieisenhamer:
Ya, just found that before seeing the comment. debugging...

stscieisenhamer:
@jemorrison Got another PR for you which should fix the `append` issue. Re-running the regressions, there are all sorts of truth differences that look expected. However two regressions failed another reason:

- jwst.regtest.test_nirspec_fs_spec2.test_nirspec_fs_spec2
- jwst.regtest.test_nirspec_mos_spec2.test_nirspec_mos_spec2

The reason is the same:

```
2020-02-28 12:23:19,960 - stpipe.Spec2Pipeline.extract_1d - INFO - Step extract_1d running with args (<MultiSlitModel from f170lp-g235m_mos_observation-6-c0e0_001_dn_nrs1_mod_s2d.fits>,).
2020-02-28 12:23:19,963 - stpipe.Spec2Pipeline.extract_1d - INFO - Step extract_1d parameters are: {'pre_hooks': [], 'post_hooks': [], 'output_file': None, 'output_dir': None, 'output_ext': '.fits', 'output_use_model': False, 'output_use_index': True, 'save_results': True, 'skip': False, 'suffix': 'x1d', 'search_output_file': True, 'input_dir': '/data1/jenkins/workspace/RT/eisenhamer-dev/clone/test_outputs/popen-gw24/test_nirspec_mos_spec2_run_pipeline0', 'smoothing_length': None, 'bkg_order': None, 'log_increment': 50, 'subtract_background': None, 'apply_nod_offset': None}
2020-02-28 12:23:20,318 - stpipe.Spec2Pipeline.extract_1d - INFO - Using EXTRACT1D reference file /grp/crds/cache/references/jwst/jwst_nirspec_extract1d_0003.json
  File "/data1/jenkins/workspace/RT/eisenhamer-dev/clone/jwst/pipeline/calwebb_spec2.py", line 95, in process
    asn.filename
  File "/data1/jenkins/workspace/RT/eisenhamer-dev/clone/jwst/pipeline/calwebb_spec2.py", line 309, in process_exposure_product
    x1d_result = self.extract_1d(result_extra)
  File "/data1/jenkins/workspace/RT/eisenhamer-dev/clone/jwst/stpipe/step.py", line 389, in run
  File "/data1/jenkins/workspace/RT/eisenhamer-dev/clone/jwst/extract_1d/extract_1d_step.py", line 193, in process
    was_source_model=False)
  File "/data1/jenkins/workspace/RT/eisenhamer-dev/clone/jwst/extract_1d/extract.py", line 2582, in run_extract1d
    apply_nod_offset, was_source_model)
  File "/data1/jenkins/workspace/RT/eisenhamer-dev/clone/jwst/extract_1d/extract.py", line 2789, in do_extract1d
    prev_offset, True, extract_params)
  File "/data1/jenkins/workspace/RT/eisenhamer-dev/clone/jwst/extract_1d/extract.py", line 3504, in extract_one_slit
    data = replace_bad_values(data, input_dq, wl_array)
  File "/data1/jenkins/workspace/RT/eisenhamer-dev/clone/jwst/extract_1d/extract.py", line 3577, in replace_bad_values
    mask = np.logical_or(mask, bad_mask)
ValueError: operands could not be broadcast together with shapes (15,1152) (0,) 
```

philhodge:
The error message "ValueError: operands could not be broadcast together with shapes (15,1152) (0,)" implies that the input slit (one of the slits in a MultiSlitModel) has a DQ array that exists but has a shape of (0,).  If the DQ array is not None, DQ should have the same shape as the SCI array.

hbushouse:
Could this somehow be related to the fact that resampled NIRSpec MOS/FS data used to be passed into `extract_1d` in a `MultiProductModel`, which has `con` and `wht` arrays, but no `dq`. Now they're being passed in a `MultiSlitModel`, which should have `con` and `wht` arrays populated by `resample_spec`, but also now has a `dq` array that hasn't been populated with anything (i.e. it has zero shape).

philhodge:
Why isn't dq None if it isn't populated?

hbushouse:
> Why isn't dq None if it isn't populated?

Good question. Could be one of those mysterious things that takes place way down in datamodels that automatically initializes an attribute without giving it any substance ("And the DQ was without form; and void.")

jemorrison:
This PR removed DrizProduct, MultiProduct and lev3_prod from the schemas. There is nothing currently set to if some tries to use these schemas - they just do not exist. Is this ok or do I need to add something ?

jemorrison:
added back the DrizProductModel and MultProductModel (ugly) and I tested when I tried to create a mode of that type it gave the warning and returned the model should be using.
@jdavies-st please look at what I did and confirm it looks ok
This should be the last thing for this issue

hbushouse:
Have you run all of this through the regtest suite recently, to confirm that it all still works? I expect there might be differences in the some of the regtests, but hopefully no actual errors.

jemorrison:
@hbushouse  Good question. After running the nirspec mos spec2 test the headers are different (as expected) but so are the extensions. There seems to be an ERR extension in the new data (in the fits file produced it is of size 0). The truth file has extensions: SCI, WHT, CON
The new files has extensions SCI,CON,WHT, ERR

hbushouse:
That zero-sized ERR extension should NOT be in the FITS file. The infrastructure is apparently smart enough to not include the DQ extension, so I wonder how/why the ERR is getting there. It should be only SCI, CON, WHT as before when using a `DrizXxxxx` model.

jwst/source_catalog/source_catalog_step.py
```diff
@@ -38,7 +38,7 @@ def process(self, input):
         npixels = self.npixels
         deblend = self.deblend
 
-        with DrizProductModel(input) as model:
+        with ImageModel(input) as model:
```
jdavies-st: This can be a standard `with datamodels.open(input) as model:`

jdavies-st:
I suspect the ERR array is appearing because it is being accessed by some function.

I really dislike this auto-creation.  The simple line:

```
hasattr(model, "error")
```

will always return True whether the `model.error` array exists or not, because it gets created when checked if it exists.  It gets created because the `__getattr__` method in datamodels does this, and also because it has a default value set.  Do either of these behaviors (auto-create and a default value of 0) make any sense?

jwst/pipeline/calwebb_coron3.py
```diff
@@ -154,10 +154,10 @@ def process(self, input):
 
         if result == resample_input:
             # Resampling was skipped,
-            #     yet we need to return a DrizProductModel, so...
+            #     yet we need to return an output resampled ImageModel, so...
             warn1 = 'Creating fake resample results until step is available'
             self.log.warning(warn1)
-            result = datamodels.DrizProductModel(data=resample_input[0].data,
+            result = datamodels.ImageModel(data=resample_input[0].data,
```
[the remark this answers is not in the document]
jemorrison: That block of code does not make sense now . 
why would "If result = resample_input" mean the resampling was skipped.

jemorrison:
@jdavies-st @stscieisenhamer 
Could you take a look at the datamodels.slit.py
Do we still want self.err = init.err and all those similar lines ?

jemorrison:
I am striking out on tracking down why the err extension is written to output file of size 0. 
I have managed to write all sorts of 0 array size data to the out put just by querying the size.
For example if at the end of resample_spec_step  I just add the lines
if(result.slits[0].dq.size ==0):
                    print('slit dq == 0')
The DQ array is then added to the resample output 2d of size 0.  This is maddening. 
Can someone suggest how I can track down when err is added - if I check the size
with err.size == 0 - then it creates it. 

Can we add some function that checks if array size is 0 then strips it from the result at the
end of the step ?

philhodge:
@jemorrison If you call `output_model.save` at various points in `resample_spec`, with different output file names (maybe just include a digit to distinguish one output file from another), that might work.  You could use `astropy.io.fits` to open the files and see which ones contain an ERR extension.  I would save the first one right after creating it, because if that one already has an ERR extension, the problem is likely not in `resample_spec` itself, but in `datamodels`.

stscieisenhamer:
@jemorrison Can you provide the data you are using to make this happen? I can track this down.

Or, if you are good with `pdb` and/or are feeling adventuresome, I would set a `breakpoint()` at this [line in properties.py](https://github.com/spacetelescope/jwst/blob/b220f8aca0fd9679753736e3e059520b2f896690/jwst/datamodels/properties.py#L148). This is where default arrays are created. Then, you can use the `u` command to go up the call stack and see where the reference is being made.

jemorrison:
getting closer to tracking down the problem with Phil suggestion.
in resample_spec.py in __init__
self.blank_output = datamodels.ImageModel(self.data_size) 
if I write this out I get an single SCI extension.

but a few lines later 
 self.blank_output.update(datamodels.ImageModel(self.input_models[0]._instance))
if I write this out I get a SCI extension of the correct size but an ERR extension 
of size 0. 
I believe that line is just to copy all the meta data in the Primary header from the first image
into the output - is there a way to do this without using "instance" which I think is creating
the empty error array.

stscieisenhamer:
The `_instance` must be making a reference to `err`. Either something has created ERR in the `_instance` or, if there is truly no `err`, there is another issue with `update`.

hbushouse:
A lot of the data models, like `SlitModel` have lines like the following in their _init_ methods:
```
            self.data = init.data
            self.dq = init.dq
            self.err = init.err
            self.area = init.area
            if init.hasattr('wavelength'):
                self.wavelength = init.wavelength
```
Should all array attributes, other than `data`, use the `hasattr` test as is done for optional arrays like `wavelength`, etc., because `err` and `dq` are also now optional in some models?

This may not be the cause of the zero-sized `err` extensions showing up in `ImageModel` output, but is ultimately related.

philhodge:
There are not many places in the pipeline code where `_instance` is used.  (I checked the step directories but none of the subdirectories, such as `tests/`.)  It's used in `datamodels` and `stpipe`, of course.  It's used in function `split_container` in master_background_step.py, and it's used in resample_spec.py.  So my guess is that it does not do anything beneficial in resample_spec.py, and it would be worth removing it as a test.

stscieisenhamer:
Depends on what `init` is. If `init` is a `DataModel`, this will always create the arrays. However, if `init` is not a `DataModel`, as this case since it is coming from the `_instance`, which should just be a `dict`, then yes, this check will work.

Without having the larger conversation about what a schema/array creation/default should be for `DataModel`, what is needed is a non-array-creating `hasattr`-like call.

Regardless, for this particular situation, I would really like to see what the `update(_instance)` is doing.

philhodge:
Would it be possible to use `dir(output_model)` and search the returned list?

jemorrison:
For resample_spec the _instance is just used to populate the Primary header of the resampled data with the information in the first set of multislit data to be resampled. 
Do we have another way to pull in all the meta data and shove it into the output data model

philhodge:
@jemorrison Wouldn't `self.blank_output.update(self.input_models[0])` be sufficient?  Other steps use `.update()` without `_instance`, so why would `resample_spec` need it?

jemorrison:
well this is interesting. 
In resample_step_spec (before it calls resample_spec)  it sets up the output data:
result = datamodels.MultiSlitModel()
result.update(input_models[0])
This pulls in the needed meta data to the resampled output

Now in resample_spec there are the lines that was giving us trouble.
self.blank_output = datamodels.ImageModel(self.data_size)
self.blank_output.update(self.input_models[0])

Notice I changed the update to just update and not _instance. This still produces a self.blank_outout with a err extension of 0 size. It must be because it is a ImageModel instead
of a MultiSlitModel (like in resample_spec_step). Not sure why it is .

So I think I do not need the self.blank_output.update(self.input_models[0]) at all since the meta
data is filled back in resample_spec_step.

jemorrison:
The NIRSPEc regression tests now pass (only difference with truth files is there is more information in the header - as expected). I am running the entire regression suite to see if
any other issues related to resample pop up

philhodge:
Would it be better to have the call to `.update()` in resample_spec.py than in resample_spec_step.py?  I'm not making a suggestion; I'm just asking.  I see that `.update()` is called in resample.py and not in resample_step.py, for instance.

jemorrison:
I think it is cleaner to call update in resample_spec_step. I did not add the update it was there before I just took out the update in resample_spec cause I do not think it is needed there and it is creating an
empty err array. 
In resample_spec_step the resampled data model MultiSlit is defined. This is the final output product. So it makes sense to me to use update here and populate the meta data. 
Resample_spec takes the container elements (the input model - and Image Model) and works on those. I don't see why update should be to be run on blank_output which is defined as an ImageModel.
I could be missing something but the regression test show the data agrees now it is just the added header values that are flagged.

jemorrison:
@philhodge Ok I see your point now. What I have implemented works if the input data is Multslit but
because I populate the meta data in process_multislit

if the input data is  ImageModel I do not have an equivalent update in resample_spec_step:process_slit.

jemorrison:
@hbushouse  when is the input to resample_spec_step an ImageModel - what type of data is that ?
I need to test that kind of data with this PR

philhodge:
@jemorrison Isn't MIRI LRS an ImageModel?

jemorrison:
Ah yes it is thanks

jemorrison:
in resample_spec_step if the data is an Image model it goes to _process_slit.
In _process_slit it there can be two types of data  produced a Single resampled  imageor a container of resampled images. When would we have the case that the input is an Image Model and out put is a
container of resampled images ?
I have tried to run MIRI LRS data through resample_spec. It runs but the output is all zeros ?
I looked at the input image and there is a spectrum. Is this a know error that does not relate to this PR  or is this error resulting from the changes in this PR. I ask because the regression test I got this from does not look at the resampled data and the comments say that is because there is a bug in resample for this data. Not sure it the error is what I am trying to fix or a different error

jemorrison:
it seems the slit location in the header for this MIRI_LRS data is incorrect. SLTSTRT1/2 = 1, SLTSIZE1/2 =0. I would think it might need that information.

jemorrison:
I think there is an error in the way 'update' works for ImageModel. WHen I use it in resample_spec for
dirzzled ImageModel data (MRS LRS data) the update creates an array of 0 size err.
This is not the case when the input is MultSlitModel - it works fine.

hbushouse:
The only way I can think of that _process_slit would result in a container of resampled images is if the input itself is a container of ImageModels. Can't think of any situations where this might happen, but the code is probably setup to handle it just in case.

The all zero resampled LRS image is not a result of this PR. As the comments in test_miri_lrs_spec2 clearly state:
```
    # Run the calwebb_spec2 pipeline; save results from intermediate steps
    # NOTE: THE RESAMPLE_SPEC STEP IS SKIPPED FOR NOW, BECAUSE IT HAS A BUG
    # (**_the s2d image is all zeros_**)
```
so that's an existing problem.

hbushouse:
The SLTSTRTn and SLTSIZEn are purely informational to let the user know what was used during the `extract_2d` step to create the slit cutout. Since LRS fixed-slit data don't go through `extract_2d` (the full image is left intact), these keywords are just set to default values.

hbushouse:
For the zero-sized ERR array added to ImageModel when it goes through .update, could it be the case that there's a reference to a zero-sized ERR array lurking down in the ASDF extension of the input LRS image, which is then causing an ERR array to appear as a result of .update? One way to find out would be to delete the ASDF extension from the test (rate) image and run it through calwebb_spec2 again to regenerate the WCS object and all other info in the ADSF extension.

jemorrison:
I removed the asdf extension and then re-ran calwebb_spec2 and the s2d still has a ERR with 0 size
This occurs after I use the update function to pull in the meta data

jemorrison:
I finally got it to work. I feel like it is an ugly solution but it works. Some one with more knowledge of datamodels probably has a better solution.
instead of drizzled_models[0].upate(input_model[0])

if we only grab the Primary and SCI meta data doing this
drizzled_models[0].update(input_models[0],only='PRIMARY')
drizzled_models[0].update(input_models[0],only='SCI')

 that seems to work. The meta data is pulled over and no err array is created

hbushouse:
I notice that image.schema inherits bunit.schema, which includes a keyword (BUNIT) defined to be in the ERR fits_hdu. Could it be that .update is creating that zero-length ERR extension simply because there's a keyword defined to be in its header? That would explain why a DQ extension doesn't get created, because it doesn't have any keywords defined for it in the schema. I wonder what would happen if you intentionally set model.meta.bunit_err = None. Would that cause the ERR extension to no longer get created?

jdavies-st:
Looking good.

I ran the regression tests on this PR (rebased to `master` at 5d83150)

Results

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/161/

Generally things look good.  But there are a number of non-spectroscopic pipelines that produce SLTSTRTn and SLTSIZEn keywords.  I don't think this is expected or desirable.  Is there a reason these keywords have default values defined?

The fact that these keywords and all the other slit-specific keywords do not live in `meta` but instead the top level may be causing unexpected behavior.

hbushouse:
That regression test run shows a number of problems.

The least severe of which is all the SLTxxxx keywords showing up in image headers in which they're totally irrelevant (i.e. i2d products). All of the default values listed in slitmeta.schema need to be removed, so that those keywords are optional (we don't need them in images, just spectra).

Second, I saw some results for MultiSlit products that had lost all of the WCS keywords from their SCI extension headers. Granted, we don't need the basic FITS WCS keywords like CRPIX, CRVAL, etc., but they're also missing useful ones like V2_REF, V3_REF, RA_REF, etc. Isn't wcsinfo.schema inherited within SlitModel or MultiSlitModel somewhere? That needs to be fixed.

Finally, there were some tests that showed outright errors due to CRDS complaining that it couldn't find the INSTRUME (model.instrument.name) keyword to do reference file lookups. That's not good.

jemorrison:
I think we have a problem. So when I remove all the slitmeta.schema default values and I run calspec2 on MIRI_LRS data it fails in flat_field because lib/reffile_utils.py can not determine the subarray properties.  
The 'try' block in trying to find the size of sci_model.xstart, xsize,ystart,ysize does not give an exception it is just that the values are None.  Reading the comments it was assumed that model.xstart and the others would only we set if  extract_2d was run. But now slitmeta.schema is setting these values to None. 

So instead of 'try' I could change it to check if any of the value are None but since this is a lib routine
and I guess a very general routine I thought I would check what others thought first.

hbushouse:
Subarray corners/sizes are stored in the SUBSTRT1/2 and SUBSIZE1/2 keywords, not the SLTSTRT1/2 and SLTSIZE1/2 keywords. The SUBSTRT and SUBSIZE keywords should be in every primary header of every product and are defined in the core schema (not in slit schema). The only time flat_field should be using SLIT corners (as opposed to SUBARRAY corners) is if the data has gone through extract_2d, which happens for NIRSpec MOS and FS, but not for MIRI LRS.

philhodge:
It's only for `MultiSlitModel` data that `slit` is passed to `apply_flat_field`.  For MIRI LRS, it should be the complete input model that is passed, and that should have the SUBSTRT1, etc., keywords, as @hbushouse pointed out.
Is the input MIRI LRS file an `ImageModel`?  Or was something done to it to make it a `MultiSlitModel`?

jemorrison:
slitmeta.schema defines
model.xstart, model.ystart, model.xsize, model.ysize

If you recall in Image.schema includes slitmeta.schema
So the MRS LRS which is an ImageModel will have model.xstart  =None, model.ystart =None and so on
The meta.subarray.subarray blah are set to correct values for this MIRI LRS data. The problem is
that model.xstart .... are all none

philhodge:
I see.  The `ref_matches_sci` function in reffile_utils.py first tries to get `sci_model.xstart`, etc., and then falls back to `sci_model.meta.subarray.xstart`.  It should get the correct attribute in the first place, depending on the input model.

hbushouse:
Perhaps this is even further motivation to never store resampled spectral data in an `ImageModel`, as is currently done for MIRI LRS fixed slit, but instead always store resampled spectra (even if it's a single spectrum) in a `SlitModel` or `MultiSlitModel` (with only 1 slit instance). Then we could drop slitmeta.schema from `ImageModel`.

jemorrison:
I think it would be much cleaner to not have MRS LRS slit resampled data  be an ImageModel. That is a bit confusing. 
So would I just change resample spec  and maybe extract_1d. The other steps would have MRS LRS data defined as ImageModel ?

jemorrison:
I think I know how to do this. In resample_spec if input is ImageModel then set up something like
new_input  = MultiSlitModel
new_input.update(input_models[0]) # to get the meta data
I think we only need the input.data array  (right?)
new_input.data  = input.data

jdavies-st:
In this PR, `ImageModel` and `SlitModel` have essentially merged into almost the same thing.  The only difference is that `SlitModel` allows a 3D `data` array.  Also `ImageModel` allows a `zeroframe` array and `SlitModel` allows a `barshadow` and `wavelength` array.  Metadata is the same, including `slitmeta.schema` living outside of `meta` in both.

Would it be good to revisit whether these need to be separate, and if they do, decide when an `ImageModel` might become a `SlitModel` in the spec2 pipeline?  At least for MIRI LRS?

My thought is that it makes sense for them to be different and to keep `slitmeta.schema` out of `ImageModel`.  But that means the spec3 pipeline cannot use `ImageModel`.

Is MIRI LRS the only one that remains an `ImageModel` in spec2 currently instead of getting extracted into a `MultiSlitModel`?

philhodge:
NIRISS SOSS is an ImageModel.

hbushouse:
That's why I suggested (earlier) to remove slitmeta from being inherited by `ImageModel`, to keep it distinct from `SlitModel`. I believe MIRI LRS fixed-slit is in fact the only mode to currently use `ImageModel` on input to `resample_spec` (and use `DrizProductModel` on output). The other modes that contain a single slit - NIRSpec fixed-slit and BrightObj - use `SlitModel`. BrightObj has to use `SlitModel`, because it uses a 3-D data array. I'm pretty sure the simpler case of regular single slit NIRSpec FS exposure also uses `SlitModel`, perhaps because `extract_2d` still gets run on it and hence it needs the slit-related keywords.

So simply swapping MIRI LRS fixed-slit to use `SlitModel` coming out of `resample_spec` should allow us to drop the inclusion of `slitmeta.schema` in `ImageModel`. The regtests can be used to confirm that for us.

hbushouse:
@philhodge Good point about NIRISS SOSS, but fortunately that doesn't go through `resample_spec`. Does it need any slit-related keywords in order to work properly in `extract_1d`? I'm guessing the answer is no, because up until now `ImageModel` never contained any slit-related keywords.

philhodge:
NIRISS SOSS needs the spectral order, but it's gotten from `wcsinfo` rather than from `slit`.
The current `extract_1d` expects an NRS_FIXEDSLIT exposure with just one slit to be an `ImageModel`.  This is the section where @hbushouse added the "temporary hack" to assign the slit name from `meta.instrument.fixed_slit`, but this should go away with the current PR.

jemorrison:
I am not converting the MIRI LRS image model to a slit model correctly.
At the every start of resample_spec_step 
I have added:
if isinstance(input, ImageModel):
            slit_model = datamodels.SlitDataModel()
            slit_model.update(input,only='PRIMARY')
            slit_model.update(input,only='SCI')
            slit_model.data = input.data
            input = slit_model
       
I do a input.save('temp.fits) to see what it looks like. It does not have the asdf extension. Problem
If I put slit_model.update(input) - dang it if I don't get an empty error array - again 

So later when it is trying to access wcs information - there is none. 
Ideas ?

jemorrison:
ok I was mistaken I do get the asdf extension.
Also if I set input.meta.bunit_err = None
then when I do slit_model.update(input)
I DO NOT get the err extension - so that was it.
A couple of problems - I can not get the Primary header meta data into slit_model
and I don't think it is correct to copy the 1032 X 1024 data into slit_model - I just want the slit region - right ?

jemorrison:
@stscieisenhamer, @hbushouse  - I just pushed this PR - this is what I am trying. The primary header information is not being updated.

jwst/resample/resample_spec_step.py
```diff
@@ -25,6 +25,23 @@ class ResampleSpecStep(ResampleStep):
     def process(self, input):
         input = datamodels.open(input)
 
+        if isinstance(input, ImageModel):
+            slit_model = datamodels.SlitDataModel()
```
hbushouse: I think you might want `datamodels.SlitModel()` here, instead of `SlitDataModel`.

jwst/resample/resample_spec_step.py
```diff
@@ -25,6 +25,23 @@ class ResampleSpecStep(ResampleStep):
...
+            input.meta.bunit_err = None  # this prevents error array populated in slit model
+            # copy the meta data from input do slit_model
+            slit_model.update(input)
+            slit_model.update(input,only='PRIMARY') # tried this it does not help populate Primary meta
+            slit_model.data = input.data
+            slit_model.meta  = input.meta
```
hbushouse: Why is this necessary, if .update() has been called? Shouldn't .update() have already copied over all the meta?

jwst/resample/resample_spec_step.py
hbushouse: Yes, doing some more checking, `SlitDataModel` does not inherit `core.schema`, while `SlitModel` does. So that's why there's no meta from the primary getting copied over by .update().

jwst/resample/resample_spec_step.py
jemorrison: Ok changing the data model to SlitModel - fixed the problem with the meta data not being copied over.

stscieisenhamer: Also, the statement `slit_model.meta  = input.meta` should not be done. This is making the two models share the meta structure. This is what the `update` should be handling.

stscieisenhamer: Looking a bit deeper, this section of code should look like:
```
            slit_model = datamodels.SlitModel()
            slit_model.update(input)
            slit_model.meta.wcs = input.meta.wcs
            slit_model.data = input.data
            input = slit_model
```

The next issue is that in `_process_slit`, the call `resamp.do_drizzle` is returning a result without the meta being properly updated. Not sure when that is supposed to occur, whether resample is supposed to take care of it or whether the callee is supposed to.

jemorrison:
slowly making some progress. Getting MIRI LRS to process through resample - but the out put is
44 X 387 - which I believe is wrong. It should be the full image.

jemorrison:
For MIRI LRS data I convert the input data to slit model. In resample it only resamples  the slit region and returns an output that is the MIRI LRS slit size. 
So do I then need to convert this data back to an image model to store the entire image and update slit region with the resampled region? 
What type of datamodel  do we want resample to return for MIRI LRS ?

hbushouse:
Examples of previously processed MIRI LRS fixed-slit exposures like this:
```
(jwstdev) macviper:b7.4_cal14_full bushouse$ fitsinfo jw00859/jw00859003001_03103_00001_mirimage_s2d.fits
Filename: jw00859/jw00859003001_03103_00001_mirimage_s2d.fits
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU     337   ()      
  1  SCI           1 ImageHDU        35   (44, 387)   float32
  2  CON           1 ImageHDU         9   (44, 387)   int32
  3  WHT           1 ImageHDU         9   (44, 387)   float32
  4  ASDF          1 BinTableHDU     11   1R x 1C   [21237B]   
```
indicate that it's expected to get only the slit region resampled, which makes sense, because that's the only region for which there's WCS info. The rest of it can't be resampled. So your latest results seem fine.

hbushouse:
We want the resampled result returned in a `SlitModel`, just like you've now got it.

jdavies-st:
Resampled MIRI LRS output is just the pixels within the slit bounding box, not the whole detector.  The WCS transforms are only defined within the slit.  So looking good.  👍

jemorrison:
Here is my next question. We started this PR to replace DrizProduct with ImageProduct and MultiProductModel with MultiSlitModel.
For  MIRI LRS resample the input is an ImageModel. Now a SlitModel is created and ImageModel is shoved into SliModel. So we return a SlitModel after drizzling.  
I hope  I have summarized that correctly. 
In __init__ to resample_spec 
self.blank_output = datamodels.ImageModel(self.data_size)
I think now that we have convert MRS LRS to SlitModel we want this be to
self.blank_output = datamodels.SlitModel(self.data_size)
self.data_size is (387,44)

Something just does not seem correct with the bounding box
The bounding box determined from resample_spec.build_interpolate_ouput_wcs is bounding box ((-0.5, 43.5), (-0.5, 386.5)).  
I would of thought that would of been the bounding box for LRS region in the the input (1024, 1032) image because this bounding box is used in resample_spec.do_drizzle

hbushouse:
I looked at a MIRI LRS fixed-slit cal product, which should be the equivalent of what's going into `resample_spec` and its bounding box is:
```
cal.meta.wcs.bounding_box                                                                                                                        
((302.5, 346.5), (6.5, 396.5))
```
which looks reasonable to me.

Are you saying that in the data model that's actually going into the drizzle routines the bounding box is different, i.e. ((-0.5, 43.5), (-0.5, 386.5))?

jemorrison:
The bounding box for the input Full MIRI LRS image seems correct.
The size of the output image is set by build_interpolated_output_wcs and that seems correct too
What I seem to be missing in how drizzle knows what region in the input image to use to resample.
I have hoping @jdavies-st  can clarify. gwcs_drizzle.add_image has parameters min,xmax,ymin and ymax which I think should contain the bb of in the input to resample. But none are passed to this routine from resample_spec so they are determined in the code gwcs_drizzle.dodrizzle.  do drizzle uses the input shape of the input image to set these values.  These x and y min and max values are what is passed to the c routine. So they are 0, 1032 and 0, 1024
For a test I hard coded xmin = 302, xmax=346, ymin = 6 and max = 396 in gwcs_drizzle.dodrizzle and the output resampled image contains the MIRI LRS spectrum.
I think resample_spec should be passing these values - has resample ever worked for the MIRI LRS ?

hbushouse:
Interesting. Yes, resample_spec used to work for MIRI LRS fixed-slit, but after the whole approach of constructing the output frame was changed to make it slit-based, rather than based on sky coords, it doesn't seem to work anymore. So perhaps something did go wrong in that refactoring, as you suggest.

jdavies-st:
Good comments Jane.  You're on the right track to fixing this.

And no, MIRI LRS has never worked it seems.  It wasn't even implemented before the refactor a year or two ago.

jemorrison:
I double checked how resample_spec works for the NIRSPEC FS data and since the input data
has already been chopped up into slits it works fine.
To me it seems if the xmin,xmax,ymin,ymax of the region is not provided by resample_spec to the drizzle routine then it use the entire input image. For NIRSPEC FS that is fine, but not for MIRI LRS.
I think we need to update resample_spec_step._process_slit - that when it calls resample. do_drizzle it passes xmin,xmax,ymin,ymax 
Could we add those parameters to **pars (drizpars)  or should we just pass them 
@jdavies-st  what you think ?

jemorrison:
I just pushed over updates. These fix the blank MIRI LRS spectrum in resampled data.
Now time to go back to running regression tests to make sure I have not broken anything else
and have fixed everything.

jwst/resample/resample_spec_step.py
```diff
@@ -24,9 +24,16 @@ class ResampleSpecStep(ResampleStep):
 
     def process(self, input):
         input = datamodels.open(input)
-
+        if isinstance(input, ImageModel):
+            slit_model = datamodels.SlitModel()
+            input.meta.bunit_err = None  # this prevents error array populated in slit model
```
hbushouse: Don't reset the value of `bunit_err` in the `input` model. This is causing the BUNIT keyword to get dropped from the header of the ERR extension of the cal file once it's written out at the end of `calwebb_spec2` processing (remember, the cal file doesn't get saved until the very end of the whole pipeline). Instead, reset `bunit_err` in `slit_model` after it's been created and updated (i.e. after line 30).

jwst/resample/resample_spec_step.py
```diff
@@ -108,23 +111,41 @@ def _process_slit(self, input_models):
...
-            model.meta.asn.table_name = input_models.meta.table_name
-            update_s_region_spectral(model)
+        bb = input_models[0].meta.wcs.bounding_box
+        ((x1, x2), (y1, y2)) = bb
+        xmin = int(min(x1, x2))
+        ymin = int(min(y1, y2))
+        xmax = int(max(x1, x2))
+        ymax = int(max(y1, y2))
+        resamp = resample_spec.ResampleSpecData(input_models, **self.drizpars)
+        drizzled_models = resamp.do_drizzle(xmin=xmin, xmax=xmax,
```
hbushouse: Passing xmin, xmax, etc. may have fixed the case of MIRI LRS fixed slit, but now a lot of NIRSpec MOS regtests are crashing in `do_drizzle`, which may be a result of this change.

jemorrison: So I thought that only MIRI LRS data would be the type of data that was ImageModel into resample_spec_step. So only MIRI LRS data would go to the routine _process_slit.
Is that a wrong assumption ?

jwst/resample/resample_spec_step.py
jemorrison: I just push over some code that I think should fix this. The issue is if I don't set bunit_err to
None before I use .update - then I get an error array in the output that I can not get rid of.
Look at what I just pushed over for ImageModel (reample_spec_step) and for Container data - __init__ in resample_spec to see if that will work. I have tested it on MIRI LRs and MIRI_LRS spec3 - which was  added a error array for the miri_lrs_spec3 test - that I missed before

jdavies-st:
Ran the regression tests on a rebased version of this branch yesterday afternoon at commit 085397e00b8b612e93c4bf20ab1c232d1c562ee9

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/162/

One can see the issue with the NIRSpec MOS that @hbushouse mentions above.  Ignore the extract_1d issues, which are unrelated and are seen in the normal regression test run last night.

https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/581/

I believe @philhodge  has a fix for these.

jwst/resample/resample_spec_step.py
```diff
@@ -23,15 +23,21 @@ class ResampleSpecStep(ResampleStep):
     """
 
     def process(self, input):
-        input = datamodels.open(input)
-
+        input_new = datamodels.open(input)
+        input_new.meta.bunit_err = None
```
jdavies-st: This can't work, as `input` may not have `meta.bunit_err`. I.e. if it is a MultiSlitModel or ModelContainer.

jwst/resample/resample_spec_step.py
jdavies-st: That was the original design, but it looks like `_process_slit` is being called for spec3 processing of NIRSpec MOS data.

hbushouse: Ah, right. I didn't notice that this was exclusive to the `process_slit` function, which only gets used if the input is a `SlitModel`. NIRSpec MOS input should be a `MultiSlitModel` and hence not go through this function. But something in the regtest run that @stscieisenhamer did using this code caused an error in `do_drizzle` when running `calwebb_spec3` on NIRSpec MOS data. I incorrectly connected the two.

jwst/resample/resample_spec_step.py
jdavies-st: You can set both the keyword and the `.err` attribute to None after running `update`.  Or delete them.

philhodge:
Some of the `extract_1d` regression test failures are expected and will need to be OKified (e.g. additional keywords in extract1d extension headers), and some will be fixed by PR #4693 which I merged recently.

jwst/resample/resample_spec_step.py
```diff
@@ -45,15 +51,14 @@ def process(self, input):
             kwargs = self._set_spec_defaults()
 
         self.drizpars = kwargs
-
         if isinstance(input_models[0], MultiSlitModel):
```
hbushouse: The regtest problem with NIRSpec MOS data in the level-3 pipeline occurs here and is due to the fact that the input models undergo a bunch of remanipulation before getting to this line. First, at the start of `calwebb_spec3` the `MultiSlitModel`s get reorganized into `MultiExposureModel`s by the `exp_to_source` process. Then at some point the container of `MultiExposureModel`s get reorganized into a `SourceModelContainer` and that process converts each model within the container back into a simple `SlitModel`. So the first model in the container, which is what's being tested here, is no longer `MultiSlitModel`, but is instead `SlitModel`, so it doesn't use `process_multislit`. Obviously this worked OK before, but now something has changed in the preceding code such that it doesn't work right anymore.

hbushouse: @jdavies-st  any suggestions?

hbushouse: Ah, I get it now. Even with the original code NIRSpec MOS data were being sent to the `process_slit` function (as opposed to `process_multislit`), because the data have been reorganized upstream in `calwebb_spec3` and hence each time you call `resample_spec` you're giving it data from only 1 slit. There are still multiple instances of data (from multiple exposures) for that slit, but it's still just 1 slit. So `process_slit` is being fed a container of multiple `SlitModel`s. The `process_multislit` function would only get used when running NIRSpec MOS data through the `calwebb_spec2` pipeline, where you've got a single instance of data for multiple slits.

So having the NIRSpec MOS data go through `process_slit` during `calwebb_spec3` is correct and hence the only way to fix the current situation is to add a check down in `process_slit` so that it only includes the xmin, xmax, ymin, ymax arguments in the call to `do_drizzle` if `input_models[0].meta.exposure.type == "MIR_LRS-FIXEDSLIT"`. Making that change results in successful NIRSpec MOS processing through `calwebb_spec3`.

Wow, this is really complicated stuff ...

jwst/resample/resample_spec.py
```diff
@@ -65,11 +65,10 @@ def __init__(self, input_models, output=None, **pars):
         self.blank_output = None
 
         # Define output WCS based on all inputs, including a reference WCS
-        # wcslist = [m.meta.wcs for m in self.input_models]
         self.output_wcs = self.build_interpolated_output_wcs()
-        self.blank_output = datamodels.DrizProductModel(self.data_size)
-
-        self.blank_output.update(datamodels.ImageModel(self.input_models[0]._instance))
+        self.blank_output = datamodels.SlitModel(self.data_size)
+        self.input_models[0].meta.bunit_err = None
```
hbushouse: This line is causing the BUNIT keyword to disappear from the ERR extension header in the output cal file when processing something like NIRSpec FS or MOS (in MultiSlitModel) through `calwebb_spec2`, because the cal file hasn't been saved yet from the input_model (it gets saved at the very end of the pipeline).

If I remove this line and instead add the following 2 lines after line 71:
```
self.blank_output.err = None
self.blank_output.meta.bunit_err = None
```
that retains the BUNIT keyword for the ERR extension in the output cal file, but unfortunately is still causing an ERR extension to show up in the resampled s2d file (along with SCI, WHT, and CON) and in this case it even has non-zero size. So why is that still showing up when the err array is set to None?

@stscieisenhamer @jdavies-st

jdavies-st: Instead of setting them to None:

```
del self.blank_output.meta.bunit_err
del self.blank_output.err
```

jemorrison: Thanks James. 
del self.blank_output.meta.bunit_err
also deletes the err array.

hbushouse: Ding! Ding! Ding! We have a winner. Simply deleting bunit_err seems to be enough. When I try to delete the err array itself, I get a "Attribute 'err' missing" error, but maybe that's because I deleted the default values for the all the data arrays in the slitdata.schema, so it doesn't get created to begin with. And now it no longer shows up in the output s2d file.

jemorrison:
Ok so much for the easily solution

self.blank_output = datamodels.SlitModel(self.data_size)
self.blank_output.update(self.input_models[0])
self.blank_output.meta.wcs = self.output_wc
self.blank_output.save('test2.fits')
del self.blank_output.err

This gives an error even though when I look at test2.fits - it has an error array.
File "/Users/morrison/git/jwst/jwst/resample/resample_spec.py", line 73, in __init__
    del self.blank_output.err
  File "/Users/morrison/git/jwst/jwst/datamodels/properties.py", line 337, in __delattr__
    "Attribute '{0}' missing".format(attr))
AttributeError: Attribute 'err' missing

Then if instead
del self.blank_output.meta.bunit_err
It runs but the s2d output has an empty error plane.
The only way I have found this to work is to set the **input model bunit_err = None**
I will try to look at the model._asdf.tree  and see if that helps.

jemorrison:
@hbushouse 
just deleting the self.blank_output.meta.bunit_err gave me an s2d empty error array for MIRI LRS data. 
What did you test it on ?

hbushouse:
I tested it on the NIRSpec MOS exposure used in the regtest "test_nirspec_mos_spec2", but also running against my checked-out version of your branch where I had also made some other mods like deleting the default values for dq, err, wht, con, etc. arrays in slitdata.schema. So that might be why I got no ERR extension at all. I'll try putting the default values back in and see what happens.

hbushouse:
One side effect of removing the default values for dq, err, wht, con, ... is that when I run a fitsdiff on the old and new s2d files, I get a lot of differences in values in all of the sci, wht, and con arrays, perhaps due to the fact that some pixels that aren't ever populated by resample_spec are now ending up with random values, instead of the usual defaults (e.g. zero). Perhaps something in resample_spec (and resample) itself should explicitly initialize the output arrays to zero, before then populating them.

hbushouse:
When I add the default values for the arrays back in, the fitsdiff differences go away, and I still don't get any ERR extensions in the output s2d file. I wonder why you're still getting them.

jemorrison:
I think the place to zero out the data is in resample.py __intit__ - the blank_output.

I still can not get del self.blank_output.meta.bunit_err not to produce an error array - at least for
MIRI LRS data. 
I made a couple of changed and I 'git diff' my changes and tried to back them out.  @hbushouse   in resample.py in the __init__ I commented out the line that does not work. Is that where you have it 
I don't get why I still get an empty error array in the s2d files.

philhodge:
Aren't we worrying too much about zero-size err arrays?  Isn't the only problem the regression tests?  They can be OKified, or smart-ified.  By the latter I mean that if two files differ only because one has a zero-size extension, that shouldn't matter.

jemorrison:
I think it would be confusing to the user to have an extension of the error that is zero in the output.
For multislit data this can mean many extensions that are zero. That just seems like something we need to fix

jwst/resample/resample_spec.py
jdavies-st: Yeah, the zero-size error array only get created when written to disk.  It needs to be created because on writing, the `bunit_err` in the ASDF tree gets mapped to that array.  If you don't have `bunit_err` in the tree, the extension doesn't get written.  So as long as the model hasn't been written out, deleting `bunit_err` is enough.

hbushouse:
The test I had run was calspec2 on a NIRSpec MOS `MultiSlitModel`, which uses function `process_multislit` and that did not show any ERR extensions. However, when I ran a calspec3 test on NIRSpec MOS data, which as we discovered earlier get reorganized such that they use the `process_slit` function (same as MIRI LRS), those did still show an empty ERR extension. So we still have a problem somewhere in the `process_slit` branch.

hbushouse:
I think that in general nothing in `resample` or `resample_spec` should ever being doing a generic `output.update(input)`, because the output model always has a different structure (i.e. different collection of data arrays) than the input. The only things they have in common is the primary header and the SCI extension (data array). So I think all uses of .update() should be modified to make 2 separate calls to:
```
   output.update(input, only="PRIMARY")
   output.update(input, only="SCI")
```
then you don't have to worry about deleting the `bunit_err` attribute or anything else. This approach also makes it future-proof. Right now it's the case that we only have 1 keyword defined that lives in the ERR extension and none in the DQ extension (or any others beyond SCI). So what happens in the future if we define more keywords for some of those extensions? We then have to come back and update all these code fragments that selectively delete meta attributes. Only updating the primary and SCI solves that problem once and for all (IMHO).

I'm currently trying that approach in my version of the branch and running a battery of tests.

hbushouse:
The last commits got us REALLY close. i2d resampled images are showing up with empty ERR extensions, because we were so focused on fixing `resample_spec` in the last round that we forgot about `resample`. Making the same change in `resample.__init__` to only update the primary and SCI hdu's should fix that.

jwst/datamodels/schemas/image.schema.yaml
```diff
@@ -35,5 +35,15 @@ allOf:
       fits_hdu: AREA
       default: 0.0
       datatype: float32
+    wht:
```
hbushouse: Just for the purpose of getting correct matching of extensions in fitsdiff comparisons used in the regtests, we need to have the order of the wht and con arrays match what was used in the old DrizProductModel, which had con first and then wht. So if you swap the ordering here we should be able to get complete fitsdiff results for the new vs old files.

jemorrison: done

jwst/extract_1d/extract.py
```diff
@@ -2862,20 +2869,18 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
                         "so apply_nod_offset will be set to False",
                         source_type)
 
-        if isinstance(input_model, (datamodels.ImageModel,
-                                    datamodels.DrizProductModel)):
+        if isinstance(input_model, datamodels.ImageModel):
```
hbushouse: There's one failing regtest (test_nirspec_fs_spec3) that shows we need a slight change in this line. The failing regtest is an instance of NIRSpec fixed-slit data for a single slit, which I think used to get resampled into a `DrizProductModel` (because it's only 1 slit). Initially this PR changed that result to an `ImageModel`, but now it's been changed again such that a resampled NIRSpec single slit gets stored in `SlitModel`. Hence it's no longer getting into this branch that sets the proper `slitname` value for it and hence later on it fails when trying to find a matching slit name ("apername") in the extract1d reference file. It's slitname is set to "NRS_FIXEDSLIT", but it needs to be "S400A1" in this case. So if you change this line to test for both `ImageModel` and `SlitModel`, it then works properly, e.g.
```
  if isinstance(input_model, (datamodels.ImageModel, datamodels.SlitModel)):
```

philhodge: That won't work.  NRS_BRIGHTOBJ data are SlitModel, and they can't be lumped into this case.  Brightobj data (currently) need slitname = "NRS_BRIGHTOBJ" in order to select the correct entry in the reference file.  I would much rather this type of data had the correct slit name, though.

philhodge: Though maybe ImageModel and 3-D SlitModel should be lumped into one case, with the number if iterations set to 1 for ImageModel.

hbushouse: Hmmm, good point about NRS_BRIGHTOBJ. Hadn't thought of that case. There's a dedicated extract1d ref file for NRS_BRIGHTOBJ that contains entries for both "S1600A1" and "NRS_BRIGHTOBJ". Since it's already a dedicated ref file for the bright object mode, I would think we could go down to just 1 entry in that ref file and have it use the proper aperture name, which is "S1600A1". The param values could be set to whatever's appropriate for when that slit is used for BRIGHTOBJ, and then meanwhile the "S1600A1" entry in the regular extract1d ref file would have params appropriate for non-TSO mode. Then the proposed logic change here should still be OK.

But what about NIS_SOSS? Does that also use `SlitModel`, with 3-D arrays, or does it use `CubeModel`?

jdavies-st:
Regression tests on this branch ran last night on commit 7692b1abce7ce0e7e99a0e814197ce1260780cd3

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/164/

jemorrison:
There is an elif that goes with if(instance, datamodels.ImageModel) 
and it is 
elif isinstance(input_model, (datamodels.CubeModel,
                                      datamodels.SlitModel)):

So not sure now about changing the if(instance, datamodels.ImageModel)  to
if(instance, (datamodels.ImageModel, datamodels.SlitModel))

is the elif one for NRS_BRIGHT_OBJ ?

jwst/extract_1d/extract.py
philhodge: The difference between those two entries (one for S1600A1 and the other for NRS_BRIGHTOBJ) is that the former has `extract_width` 7 and the latter has `extract_width` 32.

Plain old NIRISS SOSS is a 2-D ImageModel.  I don't know what data model is used for the corresponding TSO mode.

philhodge:
@jemorrison Yes, the SlitModel case is for NRS_BRIGHTOBJ, which is a 3-D SlitModel.  If you combine ImageModel and SlitModel into one case, NRS_BRIGHTOBJ data would fail because the ImageModel case does not handle 3-D data.

jwst/resample/resample.py
```diff
@@ -54,10 +54,11 @@ def __init__(self, input_models, output=None, **pars):
         # Define output WCS based on all inputs, including a reference WCS
         self.output_wcs = resample_utils.make_output_wcs(self.input_models)
         log.debug('Output mosaic size: {}'.format(self.output_wcs.data_size))
-        self.blank_output = datamodels.DrizProductModel(self.output_wcs.data_size)
+        self.blank_output = datamodels.ImageModel(self.output_wcs.data_size)
 
         # update meta data and wcs
-        self.blank_output.update(input_models[0])
+        self.blank_output.update(input_models[0], only='PRIMARY')
+        self.blank_output.update(input_models[0], only='SCI')
```
hbushouse: So the case of regular imaging data seems to not respond properly to this change. I'd go ahead and keep these new lines in there, but in addition we need to make a change in `resample_step.py` to add the following lines right after line 77:
```
  if hasattr(model.meta, "bunit_err") and model.meta.bunit_err is not None:
      del model.meta.bunit_err
```
Not elegant, but it finally gets rid of the empty ERR extensions in i2d products and hence fixes the current problems with several of the imaging-mode reg tests.

jwst/extract_1d/extract.py
jemorrison: could we do something like (I know if it not elegant) 
if input_model.meta.exposure.type != 'NRS_BRIGHTOBJ' and  isinstance(input_model, (datamodels.ImageModel, datamodels.SlitModel)):
and then also keep 
 elif isinstance(input_model, (datamodels.CubeModel,
                                      datamodels.SlitModel)):  (this should catch NRS_BRIGHTOBJ case)

hbushouse: Another solution that appears to work is to copy the existing the if-block
```
if input_model.meta.exposure.type == 'NRS_FIXEDSLIT':
```
down in the `elif` block that includes `CubeModel` and `SlitModel`. That way it's not necessary to add `SlitModel` to the branch that handles `ImageModel` and it leaves "NRS_BRIGHTOBJ" (and NIS_SOSS) data alone.

But if the two of you can come up with a more elegant way of combining these two branches, then none of this should be needed.

philhodge: Yes, I think it would work to move the section for using `fixed_slit` for `slitname` for NRS_FIXEDSLIT data to the block for CubeModel and SlitModel, and I would add ImageModel to that block (i.e. so it would be for ImageModel, CubeModel, and SlitModel).  Then it should be OK to delete the block that's currently for ImageModel.  The block for CubeModel, etc., already has a test for `len(shape) == 2`.  This might take some tinkering to make sure `slitname` gets set appropriately for all data of these types.
I would also suggest uncommenting the lines for setting `slitname = input_model.name` (which would be for NRS_BRIGHTOBJ), if possible.  And I agree with @hbushouse regarding replacing the reference file for brightobj data to have just one entry, for slit "S1600A1".

jemorrison:
I am hesitant to combine ImageModel, SlitModel and CubeModel in extract.py in this PR and do the simple thing Howard suggests -  But we only have Phil for this week so maybe we should tackle this while he is around.
@bhhouse - suggestion ?

hbushouse:
I would suggest the simple approach to start with, just to get everything working again and hopefully be able to merge this PR. Then once all of the fallout is cleaned up from that, we go back and refactor this section to simplify it. Hopefully we can do both of those within the next 4 days.

philhodge:
The "simple approach" sounds fine, and it would help to test whether the single-slit fixed slit would be processed correctly by that section that is currently for CubeModel and SlitModel.

jemorrison:
@philhodge  could you look at the changes to extract.py -I moved the NRS_FIXEDSLIT block and removed the comments to get the slit name for NRS_BRIGHTOBJ
do these look correct ?

hbushouse:
Once the conflict in extract.py is fixed, I think this is ready for another rebase (if necessary) and then another full run of reg tests. The remaining failures from the previous run are all expected differences.

jwst/extract_1d/extract.py
```diff
@@ -2968,12 +2966,16 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
...
         elif isinstance(input_model, (datamodels.CubeModel,
                                       datamodels.SlitModel)):
-
             slit = None
             # Replace the default value for slitname with a more accurate
             # value, if possible.
+            # next two lines are for NRS_BRIGHTOBJ
             if getattr(input_model, "name", None) is not None:
                 slitname = input_model.name
+            if input_model.meta.exposure.type == 'NRS_FIXEDSLIT':
```
philhodge: I would have put these two lines just after `slit = None`, but this is fine and can wait until later.

jwst/resample/resample_spec_step.py
```diff
@@ -108,23 +112,30 @@ def _process_slit(self, input_models):
...
+            ymin = int(min(y1, y2))
+            xmax = int(max(x1, x2))
+            ymax = int(max(y1, y2))
+            drizzled_models = resamp.do_drizzle(xmin=xmin, xmax=xmax,
+                                                ymin=ymin, ymax=ymax)
         else:
-            result = drizzled_models
+            drizzled_models = resamp.do_drizzle()
 
+        drizzled_models[0].update(input_models[0], only="PRIMARY")
```
hbushouse: These 2 lines doing updates are causing failures in the regtests for some of the exposure time related keywords in the output (resampled) image header. The drizzle routines compute new values for some of them, such as the total exposure time of all images that've been combined and put those values into the `drizzled_models` that's returned. But then these 2 update calls essentially reset all of the meta data back to their original input values, which is causing the differences in the regtests (e.g. EFFEXPTM is back to the original single exposure values, rather than the total). The original `process_slit` code did not do any update like this on the returned `drizzled_models`, so we shouldn't need it in this new version either. When I remove these lines (133-134), s2d products from `calwebb_spec3` go back to not showing any differences in the regtest comparisons.

jdavies-st:
Kicked off a regression test run on this PR branch (rebased against `master`) here

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/165/

hbushouse:
Soooo close! The only problem revealed by the last regtest run is the fact that an empty INT_TIMES table extension is occasionally showing up in calspec3 x1d products for observing modes that don't need the INT_TIMES table. There's already a hook in `extract_1d` to not copy the table data to the output for those modes, but the empty table extension is still showing up (which it didn't before). I've got a simple 2-line fix for that, which I'm testing right now on the failed regtests. When I'm convinced it's OK, I'll either push the commit directly to this PR or send you the lines to edit.

Note that quite a few of the regtest failures were actually differences due to other PR's merged into master recently. I'm going to update those truth files now, before we make another run.

PR #4561, opened 2020-02-12
JP-1129: Add guard for xref/yref; add unit tests

### LINKED ISSUE #4239: Make use of XSCI_REF/YSCI_REF more robust against missing keywords
Issue [JP-1129](https://jira.stsci.edu/browse/JP-1129) was created by Howard Bushouse:

There have been a number of occasions where one of the calwebb_tso3 steps crashes due to the XSCI_REF/YSCI_REF (or CRPIX1/CRPIX2) keywords not being present in the input header (due to upstream problems with WCS construction). The steps that make use of these keywords should have try/except blocks added so that more information error messages are emitted when these values aren't available.

jwhite3:
This PR adds exceptions in cases where `xref_xci` or `yref_sci` are missing in `assign_wcs.nircam.tsgrism`, `extract_2d.extract_tso_object`, and `tso_photomoetry.TSOPhotometryStep`.

Additionally, unit tests were added to test for cases where the exception should be raised.

Resolves #4239 / [JP-1129](https://jira.stsci.edu/browse/JP-1129)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 0.14.3 (Unreleased)
 ===================
 
+assign_wcs
+----------
+
+- A ``ValueError`` is now raised if input data is missing ``xref_sci`` or ``yref_sci`` keywords. [#4561]
+
 associations
 ------------
 
@@ -44,6 +49,8 @@ extract_2d
 - For GRISM data, the variance arrays and INT_TIMES table are copied to output,
   and keywords SLTSTRT1 and SLTSTRT2 are set to the pixel location of the
   cutout in the input file. [#4504]
+
+- A ``ValueError`` is now raised if the input data is missing ``xref_sci`` or ``yref_sci`` keywords. [#4561]
 
 master_background
 -----------------
@@ -118,6 +125,10 @@ wfs_combine
 - Use float64 data types internally in ``wfs_combine`` so as not to cause an
   error in ``scipy.signal.convolve``. [#4432]
 
+tso_photometry
+--------------
+
+- A ``ValueError`` is now raised if the input data for ``call`` is missing ``crpix1`` or ``crpix2`` keywords. [#4561]
 
 0.14.2 (2019-11-18)
 ===================
@@ -2076,8 +2087,6 @@ assign_wcs
 
 - fix input units to meters when filter=OPAQUE [#2134]
 
-- A ``ValueError`` is now raised if input data is missing ``xref_sci`` or ``yref_sci`` keywords. [#4561]
-
 
 associations
 ------------
@@ -2141,8 +2150,6 @@ extract_1d
 
 extract_2d
 ----------
-
-- A ``ValueError`` is now raised if the input data is missing ``xref_sci`` or ``yref_sci`` keywords. [#4561]
 
 firstframe
 ----------
@@ -2294,9 +2301,6 @@ tso_photometry
 - Added a new model for setting parameters for TSO photometry [#2239]
 
 - Add a  reference file for use with tso_photometry [#2254, #2264]
-
-- - A ``ValueError`` is now raised if the input data for ``call`` is missing ``crpix1`` or ``crpix2`` keywords. [#4561]
-
 
 tweakreg
 --------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2076,6 +2076,9 @@ assign_wcs
 
 - fix input units to meters when filter=OPAQUE [#2134]
 
+- A ``ValueError`` is now raised if input data is missing ``xref_sci`` or ``yref_sci`` keywords. [#4561]
+
+
 associations
 ------------
 
@@ -2139,6 +2142,7 @@ extract_1d
 extract_2d
 ----------
 
+- A ``ValueError`` is now raised if the input data is missing ``xref_sci`` or ``yref_sci`` keywords. [#4561]
 
 firstframe
 ----------
@@ -2290,6 +2294,9 @@ tso_photometry
 - Added a new model for setting parameters for TSO photometry [#2239]
 
 - Add a  reference file for use with tso_photometry [#2254, #2264]
+
+- - A ``ValueError`` is now raised if the input data for ``call`` is missing ``crpix1`` or ``crpix2`` keywords. [#4561]
+
 
 tweakreg
 --------
--- a/jwst/assign_wcs/nircam.py
+++ b/jwst/assign_wcs/nircam.py
@@ -198,6 +198,13 @@ def tsgrism(input_model, reference_files):
     # begin at 0,0, so no need to translate the crpix to full frame
     # because they already are in full frame coordinates.
     xc, yc = (input_model.meta.wcsinfo.siaf_xref_sci, input_model.meta.wcsinfo.siaf_yref_sci)
+
+    if xc is None:
+        raise ValueError('XREF_SCI is missing.')
+
+    if yc is None:
+        raise ValueError('YREF_SCI is missing.')
+
     xcenter = Const1D(xc)
     xcenter.inverse = Const1D(xc)
     ycenter = Const1D(yc)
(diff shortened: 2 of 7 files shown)


jwhite3:
Could use some help/ideas on testing `extract_tso_object` and `TSOPhotometryStep`. 

For `extract_tso_object`, my first attempt results in an `AttributeError` where the resulting model doesn't have the `source_catalog` key needed to run the function. I suppose I could change the order that the exceptions are triggered, but that seemed lazy. 

For `TSOPhotometryStep`, I wanted to see if anyone had any suggestions as this class has no direct tests.

jwhite3:
Figured out the issue with the tests for `extract_tso_object`.. wrong function was being called

jdavies-st:
For functions that don't currently have unit tests and seem to defy unit testing because of the complexity of the function, it might be worth a small refactor of the function - i.e. break it up into 2 or more functions so that each of the pieces can then more easily be unit tested. I don't know the specifics here, but sometimes that's worth a shot.

jwst/extract_2d/grisms.py
```diff
@@ -111,6 +111,12 @@ def extract_tso_object(input_model,
     if len(available_orders) > 1:
         raise NotImplementedError("Multiple order extraction for TSO not currently implemented")
 
+    if input_model.meta.wcsinfo.siaf_xref_sci is None:
+        raise ValueError('XREF_SCI is missing.')
+
+    if input_model.meta.wcsinfo.siaf_yref_sci is None:
+        raise  ValueError('YREF_SCI is missing.')
```
hbushouse: Minor nit: extra whitespace between `raise` and `ValueError`

CHANGES.rst
```diff
@@ -2076,6 +2076,9 @@ assign_wcs
 
 - fix input units to meters when filter=OPAQUE [#2134]
 
+- A ``ValueError`` is now raised if input data is missing ``xref_sci`` or ``yref_sci`` keywords. [#4561]
```
philhodge: It looks as if you added these notes near the end of CHANGES.rst.  They should go near the beginning of the file, under the heading "0.14.3 (Unreleased)".

jwst/assign_wcs/nircam.py
```diff
@@ -198,6 +198,13 @@ def tsgrism(input_model, reference_files):
     # begin at 0,0, so no need to translate the crpix to full frame
     # because they already are in full frame coordinates.
     xc, yc = (input_model.meta.wcsinfo.siaf_xref_sci, input_model.meta.wcsinfo.siaf_yref_sci)
+
```
philhodge: @nden Should 1 be subtracted from `xc` and from `yc`?

jwst/tso_photometry/tests/test_tso_photometry_step.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+
+from astropy.io import fits
+
+from jwst.tso_photometry.tso_photometry_step import TSOPhotometryStep
```
philhodge: Should this be a relative import?

jwst/assign_wcs/nircam.py
hbushouse: I don't remember if these values, which come from the SIAF, are 1-indexed or 0-indexed. @nden do you know?

philhodge: In `extract_tso_object` in extract_2d/grisms.py, 1 is subtracted from these values to get `source_xpos` and `source_ypos`, so it looks to me as if one or the other of these functions should be changed.

nden: If the transforms are defined using 0-based coordinates, yes. And normally we would want them 0-based. But I will have to double check that with the INS team. I suggest we merge this as is and file a ticket (I can file it). I also see use of `CRVAL1/2`  below which probably isn't necessary, we should use `RA/DEC_REF`.
I will review the tsgrism code after I finish with wfss.

nden: Sorry I didn't see the previous replies. If extract_2d is subtracting 1 and `source_x/ypos` are used then everything is correct. I'd prefer to subtract it here though when this is defined.
@hbushouse The SIAF values are 1-based.

jwst/tso_photometry/tests/test_tso_photometry_step.py
jwhite3: This `tests` subdirectory wasn't set up with `__init__.py`, so I left it as is. I've added the missing `__init__.py` so that relative imports can be used.

CHANGES.rst
jwhite3: Whoops! Moved the notes

jwst/assign_wcs/nircam.py
hbushouse: So, based on @nden comments, we should modify line 200 to subtract 1 from siaf_xref_sci and siaf_yref_sci when assigning their values to xc, yc, and then ... (see further comments below)

jwst/assign_wcs/nircam.py
```diff
@@ -198,6 +198,13 @@ def tsgrism(input_model, reference_files):
...
+    if xc is None:
+        raise ValueError('XREF_SCI is missing.')
+
+    if yc is None:
+        raise ValueError('YREF_SCI is missing.')
+
```
hbushouse: ... modify lines 213-216 to replace each instance of `crval1` with `ra_ref`, and each instance of `crval2` with `dec_ref`, and ... (more below)

jwst/extract_2d/grisms.py
```diff
@@ -111,6 +111,12 @@ def extract_tso_object(input_model,
     if len(available_orders) > 1:
         raise NotImplementedError("Multiple order extraction for TSO not currently implemented")
 
+    if input_model.meta.wcsinfo.siaf_xref_sci is None:
+        raise ValueError('XREF_SCI is missing.')
+
+    if input_model.meta.wcsinfo.siaf_yref_sci is None:
+        raise ValueError('YREF_SCI is missing.')
+
```
hbushouse: ... I was about to say change lines 132-133 to no longer subtract 1, because we're doing it earlier, but the subtraction that's done earlier is done on the values in the WCS transforms, not on the keyword values themselves, so the keyword values themselves will still be 1-indexed in lines 132-133 and hence still need to subtract 1.

One way around this would be to modify the meta data values themselves back in ``assign_wcs``, but I don't like th idea of modifying meta data values - they should always be kept at their original input values.

So is the change to subtract 1 in `assign_wcs/nircam.py` necessary after all?  @nden

jwst/assign_wcs/nircam.py
philhodge: @nden I'm not sure what you mean by "when this is defined."  Do you mean the transform that's defined in function `tsgrism` in assign_wcs/nircam.py?  It looks to me as if the wcs created by `extract_tso_object` in extract_2d/grisms.py is just the one from assign_wcs/nircam but with shifts and spectral order selection added.  That is, the wcs in extract_2d adds to rather than replaces the one from assign_wcs.  Am I misinterpreting the code in extract_2d?

nden: I think this goes beyond the scope of this PR. I am refactoring the code and will look into it in detail when I get to this function. My comment above was not meant to be a requirement for this PR, I just noticed and was surprised by the use of CRVALs.

hbushouse: That's fine with me, to confine the scope of this PR to only adding the raising of errors when the meta values aren't available, and not try to fix any other issues involving the actual use of xref_sci or crval1/2.

jwhite3:
@hbushouse , @philhodge , @nden It seems like in the conversation it was decided to address the discussed issues/questions in another PR. If that's the case, should I go ahead and merge this one?

philhodge:
Yes, I would say you should go ahead and merge it.

PR #4567, opened 2020-02-12
Add script okify_regtests

### LINKED ISSUE #4303: Hook up regtest okify script to new Okify infrastructure
Issue [JP-1155](https://jira.stsci.edu/browse/JP-1155) was created by James Davies:

The new regtest infrastructure currently leaves behind a JSON breadcrumb to okify the test via Artifactory's JFROG CLI.  This should be wrapped as it is in our current regtest okify script for the following use cases:

1. Okify files Artifactory > Artifactory
2. Okify files local disk > Artifactory

jdavies-st:
This adds the OKifying script to our new regreession test infrastructure.

To interactively OKify Jenkins RT/JWST build 540
```
$ okify_regtests 540
```
From a user perspective, it works the same as @eslavich 's script before.

Under the hood, it downloads the `_okify.json` and `_rtdata.asdf` breadcrumb files left behind by failed tests which are uploaded to Artifactory under the build number.  It then runs `jfrog rt cp` on the specfile, giving the user the traceback info and the option to okify or skip.

Resolves #4303


jdavies-st:
@hbushouse 
History Eraser Button

scripts/okify_regtests
```diff
@@ -0,0 +1,161 @@
...
+
+    for a,b in zip(specfiles, asdffiles):
+        if a.split(".")[0][:-6] != b.split(".")[0][:-7]:
+            raise RuntimeError("The _okify.json and _rtdata.asdf files are not matched")
+
+    return specfiles, asdffiles
+
+
+@contextmanager
+def cd(newdir):
```
eslavich: `pushd` might be a good name for this

scripts/okify_regtests
```diff
@@ -0,0 +1,161 @@
...
+
+    return sorted(glob("*_okify.json"))
+
+
+def artifactory_get_rtdata_asdf(build_number):
+    build_name = BUILD_NAME
+
+    props = f"build.number={build_number};build.name={build_name}"
+
+    # Retreive all the okify specfiles for failed tests.
```
eslavich: Is this comment relevant here?  Looks like it may have been copied/pasted from the previous function

scripts/okify_regtests
```diff
@@ -0,0 +1,161 @@
...
+    args = list(["jfrog", "rt", "dl"]
+        + [f"{ARTIFACTORY_REPO}/*/*_rtdata.asdf"]
+        + [f'--props="{props}"']
+        + ["--flat"])
+    cmd = " ".join(args)
+    subprocess.run(cmd, check=True, shell=True, capture_output=True)
```
eslavich: Is `shell=True` necessary?  It looks to me like the argument list is crafted for shellessness.

scripts/okify_regtests
```diff
@@ -0,0 +1,161 @@
...
+
+    props = f"build.number={build_number};build.name={build_name}"
+
+    # Retreive all the okify specfiles for failed tests.
+    args = list(["jfrog", "rt", "dl"]
+        + [f"{ARTIFACTORY_REPO}/*/*_okify.json"]
+        + [f'--props="{props}"']
+        + ["--flat"])
+    cmd = " ".join(args)
+    subprocess.run(cmd, check=True, shell=True, capture_output=True)
```
eslavich: Same question, `shell=True` may not be necessary

scripts/okify_regtests
```diff
@@ -0,0 +1,161 @@
...
+
+def artifactory_get_build_artifacts(build):
+    specfiles = artifactory_get_okify_specfiles(build)
+    asdffiles = artifactory_get_rtdata_asdf(build)
+
+    if len(specfiles) != len(asdffiles):
+        raise RuntimeError("Different number of _okify.json and _rtdata.asdf files")
+
+    for a,b in zip(specfiles, asdffiles):
+        if a.split(".")[0][:-6] != b.split(".")[0][:-7]:
```
eslavich: These index numbers are a little mysterious, may be worth defining constants:

```python
RTDATA_SUFFIX = "_rtdata.asdf"
SPECFILE_SUFFIX = "_okify.json"
```

And using those everywhere instead.

hbushouse:
https://youtu.be/JmyyR22RLGk

scripts/okify_regtests
jdavies-st: So this is something that I could use some input on.  I would prefer to use something like
```
subprocess.run(args, check=True)
```
But the args are then sanitized, and because one of the args has a `;` in it (I hate you JFROG CLI syntax), it gets munged and the command doesn't work.  Any other ideas on how to get `subprocess.run` not mangle string input?

jdavies-st: So it turns out I can avoid `--props="build.number=540;build.name=RT :: JWST"` by using instead `--build=RT :: JWST/540`.  No semicolons, and no need of `shell=True`.

👍

scripts/okify_regtests
jdavies-st: I combined these 2 functions into 1, and added an arg `suffix`.

PR #4573, opened 2020-02-14
JP-1264: Fix TSGRISM WCS subarray offset

### LINKED ISSUE #4513: Investigate f444w wavelength issues
Issue [JP-1264](https://jira.stsci.edu/browse/JP-1264) was created by Jonathan Eisenhamer:

See if the wavelengths assigned for F444W are reasonable when used in NIRCam GRISM exposures

hbushouse:
This implements the fix that was determined for the wavelength scale of NIRCam TSGRISM exposures, as discussed extensively in [JP-1264](https://jira.stsci.edu/browse/JP-1264). The fix is to set the x-shift added to the subarray WCS transform to zero, which corresponds to the fact that the entire x-axis worth of data is included in the subarray extraction. Detailed notes have been included in the code, for future maintenance.

Also cleaned up some of the existing code comments that were originally copied from the `extract_grism_objects` routine, but did not apply (at least not directly) to the `extract_tso_object` routine.

Minor updates to the `extract_2d` docs are forthcoming.

Fixes #4513

### DIFF
--- a/jwst/extract_2d/grisms.py
+++ b/jwst/extract_2d/grisms.py
@@ -195,7 +195,7 @@ def extract_tso_object(input_model,
         # subarray WCS transform. If the team ever decides to change the extraction limits,
         # the following two constants must be modified accordingly.
         xmin_ext = 0  # hardwire min x for extraction to zero
-        xmax_ext = input_model.data.shape[-1]  # hardwire max x for extraction to size of data
+        xmax_ext = input_model.data.shape[-1] - 1  # hardwire max x for extraction to size of data
 
         order_model = Const1D(order)
         order_model.inverse = Const1D(order)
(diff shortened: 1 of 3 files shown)


hbushouse:
Note to reviewers: the only substantive change to the actual code is the change in (new) 203 to set the x-shift in the additional transforms added to the cutout WCS object to the actual value used when doing the array slicing. All other changes to comments and log messages are purely cosmetic.

jwst/extract_2d/grisms.py
```diff
@@ -167,52 +170,66 @@ def extract_tso_object(input_model,
...
+        # above based on the min/max wavelength range and the source position to do the
+        # subarray extraction and set the subarray WCS accordingly. HOWEVER, the NIRCam
+        # team has asked for all data along the dispersion direction to be included in
+        # subarray cutout, so here we override the xmin/xmax values calculated above and
+        # instead hardwire the extraction limits for the x (dispersion) direction to
+        # cover the entire range of the data and use this new minimum x value in the
+        # subarray WCS transform. If the team ever decides to change the extraction limits,
+        # the following two constants must be modified accordingly.
+        xmin_ext = 0  # hardwire min x for extraction to zero
+        xmax_ext = input_model.data.shape[-1]  # hardwire max x for extraction to size of data
```
philhodge: It's a little confusing that `xmax_ext` here is a slice limit, rather than the actual upper limit.  That results in inconsistent notation when extracting a slice, e.g. `ymin: ymax + 1, xmin_ext:xmax_ext`.  If you subtract 1 here, you would need to add 1 when taking the slice; that would be more verbose, but perhaps less confusing because the notation for both x and y would be consistent.

hbushouse: I thought about that too while writing the code. I left it this way for simplicity, but I think I will go back and update it so that it works consistently as slice indexes.

philhodge: Aren't the bounding box limits inclusive?  I thought that was why `ymax` was the actual upper limit.

hbushouse: Yes, bounding box limits are inclusive, so in the y-axis they go from -0.5 to 63.5 (in 0-indexed python pixels) for the data array that is 64 pixels high. In the slicing done below, the upper limit on y is ymax+1, which causes the last pixel in the cutout to be number 63 (0-indexed). This change to setting "Shift(xmin_ext)" has no affect on the bounding box. It's still -0.5 < x < 2047.5 and -0.5 < y < 63.5, as it was before.

PR #4575, opened 2020-02-14
JP-954 Swap alpha and beta arrays; add support for a DQ extension

philhodge:
The `alpha` and `beta` arrays from the `IRS2Model` reference file were used incorrectly.  The `alpha` arrays are now used for the reference output, and the `beta` arrays are now used for the interleaved reference pixels.

If there is a DQ array in the reference file, this will now be used to set some of the reference pixel values to zero.

### DIFF
--- a/docs/jwst/references_general/refpix_reffile.inc
+++ b/docs/jwst/references_general/refpix_reffile.inc
@@ -29,9 +29,9 @@ READPATT   model.meta.exposure.readpatt
 
 Reference File Format
 +++++++++++++++++++++
-REFPIX reference files are FITS format, with 1 BINTABLE extension.
+REFPIX reference files are FITS format, with two BINTABLE extensions.
 The FITS primary HDU does not contain a data array.
-The BINTABLE extension is labeled with EXTNAME = "IRS2" and has the
+The first BINTABLE extension is labeled with EXTNAME = "IRS2" and has the
 following column characteristics:
 
 =======  =========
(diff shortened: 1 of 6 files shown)


jwst/datamodels/schemas/irs2.schema.yaml
```diff
@@ -29,3 +29,13 @@ allOf:
         datatype: float32
       - name: beta_3
         datatype: float32
+    dq_table:
+      title: Reference file for IRS2 refpix correction
```
hbushouse: The title for this extension/table should refer to that table, e.g. "Table of Data Quality information"

jwst/refpix/irs2_subtract_reference.py
```diff
@@ -138,6 +138,30 @@ def correct_model(input_model, irs2_model,
...
+            for row in range(len(output)):
+                part = output[row]
+                bits = decode_mask(output, mask, row)
+                log.debug("output {} DQ bits {}".format(part, bits))
+                for k in bits:
+                    ref = (scipix_n // 2 + k * (scipix_n + refpix_r) +
+                           2 * (odd_even[row] - 1))
+                    log.debug("bad interleaved reference at pixels {} {}"
+                              .format(ref, ref+1))
+                    data[integ, :, :, ref:ref+2] = 0.
```
hbushouse: Wow, that really is a complicated process for decoding the DQ mask info!

jwst/refpix/irs2_subtract_reference.py
```diff
@@ -261,6 +286,55 @@ def exclude_ref(output_model, irs2_mask):
         temp_array = output_model.err
         output_model.err = temp_array[..., irs2_mask]
 
+
+def decode_mask(output, mask, row):
+    """Interpret the MASK column of the DQ table.
+
+    As per the ESA CD{3 document:
```
hbushouse: Looks like a  typo: "CD{3" should probably be "CDP3"

hbushouse:
What about documentation? Does the `refpix` step doc need any updates to include descriptions of how the DQ mask info is applied?

philhodge:
Yes to all of those comments; thanks.  I'll fix these.

jwst/datamodels/irs2.py
```diff
@@ -11,10 +11,14 @@ class IRS2Model(DataModel):
     Parameters
     __________
     irs2_table : numpy table
-         Reference file for IRS2 refpix correction
-         A table with 8 columns and 2916352 (2048 * 712 * 2) rows.  All
-         values are float, but these are interpreted as alternating real
-         and imaginary parts (real, imag, real, imag, ...) of complex
-         values.  There are four columns for ALPHA and four for BETA.
+        Reference file for IRS2 refpix correction
```
jdavies-st: `"Table for IRS refpix correction"`  The table is not the file.

jwst/datamodels/schemas/irs2.schema.yaml
jdavies-st: The title above for `irs2_table` should also replace "Reference file" with "Table"

jwst/refpix/irs2_subtract_reference.py
jdavies-st: Could leave some comment here as to what is happening?  Why dividing, multiplying and adding 2?

jwst/refpix/irs2_subtract_reference.py
```diff
@@ -261,6 +286,55 @@ def exclude_ref(output_model, irs2_mask):
...
+    reference pixels, which are then read twice consecutively. Therefore,
+    the masks are 32 bit unsigned integers that encode bad interleaved
+    reference pixels/columns from left to right (increasing column index)
+    in the native detector frame. When a bit is set, the corresponding
+    reference data should not be used for the correction."
+
+    Parameters
+    ----------
+    output : 1-D ndarray, int
+        An array of amplifier output numbers, 1, 2, 3, or 3.
```
jdavies-st: 1, 2, 3 or _4_?

docs/jwst/references_general/refpix_reffile.inc
```diff
@@ -47,12 +47,52 @@ beta_2   float32
...
+"There is also a DQ extension that holds a binary table with three
+columns (OUTPUT, ODD_EVEN, and MASK) and eight rows. In the current
+IRS2 implementation, one jumps 32 times to odd and 32 times to even
+reference pixels, which are then read twice consecutively. Therefore,
+the masks are 32 bit unsigned integers that encode bad interleaved
+reference pixels/columns from left to right (increasing column index)
+in the native detector frame. When a bit is set, the corresponding
+reference data should not be used for the correction."
+
+I assumed that "native detector frame" in the above description referred to
```
hbushouse: We should probably change this to be less personal (i.e. not first-person). So perhaps "We assume that ..."

docs/jwst/references_general/refpix_reffile.inc
```diff
@@ -47,12 +47,52 @@ beta_2   float32
...
+the order that the data and interleaved reference pixels were read out from
+the detector, not the physical locations of the pixels on the detector.  The
+difference is that the readout direction changes when going from one
+amplifier output to the next; that is, the pixels are read out from left to
+right for the first and third outputs, and they are read out from right to
+left for the second and fourth outputs.  Furthermore, I assumed that for the
```
hbushouse: Same here: change to "Furthermore, we assume that ..."

jwst/refpix/irs2_subtract_reference.py
```diff
@@ -261,6 +286,55 @@ def exclude_ref(output_model, irs2_mask):
...
+    bits = np.where(temp > 0)[0]
+
+    # The bit number corresponds to a count of groups of reads of the
+    # interleaved reference pixels.  It wasn't clear to me whether bit
+    # number increases from left to right or right to left within a
+    # 32-bit unsigned integer.  It also wasn't clear whether the direction
+    # should follow the direction of reading within an amplifier output.
+    # The following is my interpretation of the description.
+    if part // 2 * 2 != part:
+        bits = 31 - bits
```
jdavies-st: Is there a way to use our mnemonics here, or use `np.bitwise_*`?

philhodge:
Thanks.  I'll fix these.

philhodge:
I think I've addressed all the comments.

PR #4577, opened 2020-02-14
Refactor unit tests for jwst submodules

jwhite3:
This PR adds the class `DataBundle` for creating DataModels and corresponding WCS instances for testing. Unit tests in the `assign_wcs` submodule have been refactored to use a fixture-factory for producing these bundles.

### DIFF
--- a/jwst/assign_wcs/tests/test_nircam.py
+++ b/jwst/assign_wcs/tests/test_nircam.py
@@ -8,169 +8,144 @@ of the results is verified by the team based on the specwcs reference
 file.
 
 """
-from numpy.testing.utils import assert_allclose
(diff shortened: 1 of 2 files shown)


jwhite3:
@jdavies-st I added you to this since we had chatted about it briefly.

Also, this draft currently only includes changes in `assign_wcs.tests.test_nircam` since I was testing out the idea before going through and refactoring the rest

jwhite3:
I think I'm actually leaning away from this implementation. I'm going to think about how to make some aspects more generalize, and then also I realized that I could just assign the wcs instance to the datamodel..

jwhite3:
Holding off on this. Closing

Issue #2219, opened 2018-07-03
Remove INDEF from code and reffiles


jdavies-st:
We currently use INDEF as a default parameter value in some `resample` and `outlier_detection` code, and in the reference file for these steps.  We should remove INDEF and be explicit about what values we want to put here, as INDEF is undefined outside of IRAF.

INDEF gets passed to the C `drizzle` code, as a fill value.  But the C `drizzle` code converts this to a fill value of 0., as INDEF doesn't exist in C either.  We should be explicit in our pipeline code.  Pass `fillval=0.` if we want 0., and get rid of INDEF.  If we want `fillval=np.nan`, then be explicit there too.

For backwards compatibility reasons, C `drizzle` respects INDEF, but we should be explicit rather than implicit in the `jwst` pipeline code and stop using it.

melanieclarke:
INDEF is no longer used.

PR #4578, opened 2020-02-15
Jp 1176 Hardwire required pipeline outputs in the pipeline

### LINKED ISSUE #4341: strun using config file should produce same results as with step class
Issue [JP-1176](https://jira.stsci.edu/browse/JP-1176) was created by James Davies:

Currently, some of our pipelines write out other files other than the one that is returned at the end of the pipeline.  In this case, what is written out can be different depending on whether the pipeline is run with the default config file or from the pipeline class itself (in python or using `strun`).  The file can or cannot be written out, and it can be written out with different extensions, depending how the pipeline is run.

For details on `Image2Pipeline`, see

https://github.com/spacetelescope/jwst/pull/4333

which also provides a solution for that pipeline.  Other pipelines that have this problem are `Spec2Pipeline`.  Possibly others.

Review all pipelines and make them at least consistent with each other, and consistent with the way DMS is running our code.

- pipeline-specific suffixes should be defined in the pipeline module itself, not the config file
- writing out those extra files should depend on `self.save_results`
- We should think hard about how to deal with subsequent command line args passed to the pipeline.

stscieisenhamer:
Resolves #4341/[JP-1176](https://jira.stsci.edu/browse/JP-1176)


jwst/pipeline/calwebb_image3.py
```diff
@@ -100,6 +100,7 @@ def process(self, input):
         self.save_model(result, suffix=self.suffix)
 
         self.log.info("Creating source catalog...")
+        self.source_catalog.save_results = self.save_results
```
hbushouse: So in all of these cases that now pass `self.save_results` (which is the pipeline-level argument) to the lower-level step arguments, where is the pipeline-level argument being set? Is that still in the pipeline cfg file? If so, then what about cases where the pipeline cfg isn't being used?

jwst/pipeline/resample.cfg
```diff
@@ -1,11 +1,2 @@
 name = "resample"
 class = "jwst.resample.ResampleStep"
-
-single = False
```
hbushouse: Are all of these set in the step `spec` so that we still get the same default values?

jwst/resample/resample_spec_step.py
```diff
@@ -18,6 +18,11 @@ class ResampleSpecStep(ResampleStep):
         A singe datamodel, a container of datamodels, or an association file
     """
 
+    # Spec is all the same except for the suffix
+    spec = """
+        suffix = string(default='s2d')
```
hbushouse: So are these now being set in both the calling pipeline module (e.g. `calwebb_spec2`, `calwebb_spec3`) and here in the step's `spec`?  I guess that covers cases where someone only runs the step and not the whole pipeline?

But if that's the case, why aren't we including `suffix` in the `spec` of lots more steps?

And does that mean that if someone runs the step by itself, the output file suffix will be `s2d`, instead of the usual `resample_spec`?

jwst/pipeline/resample.cfg
stscieisenhamer: Yes. This is actually preliminary work for the epic that deals with deprecating out the cfg specifications.

jwst/resample/resample_spec_step.py
stscieisenhamer: Nope, good catch. The suffix should not be set here but by the pipelines that use this step. Will need to fix that.

jwst/resample/resample_step.py
```diff
@@ -26,13 +26,14 @@ class ResampleStep(Step):
     """
 
     spec = """
-        pixfrac = float(default=None)
-        kernel = string(default=None)
-        fillval = string(default=None)
-        weight_type = option('exptime', default=None)
+        pixfrac = float(default=1.0)
+        kernel = string(default='square')
+        fillval = string(default='INDEF')
```
jdavies-st: I think `fillval` should remain `None` here.  See #2219

jwst/pipeline/calwebb_image3.py
stscieisenhamer: The pipeline level `save_results` is handled as before: either set in `strun` or, if a user is doing so programmatically, set explicitly. No change in behavior there.

The intent of this PR is that, for all pipelines, if the "expected results" of that pipeline come from more than one step, those steps are forced to follow the pipeline's state of `save_results`.

jwst/resample/resample_step.py
stscieisenhamer: Clarification: The original CFG file had `fillval=INDEF`, Would setting the default to `None` now without #2219 being dealt with, be OK?

stscieisenhamer: Offline conversation says this is ok to leave for now.

PR #4587, opened 2020-02-17
Jp 1265 Update schema-editor to match documentation and clarify execution

stscieisenhamer:
Resolves [JP-1265](https://jira.stsci.edu/browse/JP-1265)/#4518.

Note to reviewers: As important to the code review is the documentation review of the confluence document [Keyword Dictionary JSON vs. YAML Comparision Process](https://innerspace.stsci.edu/display/SCSB/Keyword+Dictionary+JSON+vs.+YAML+Comparison+Process). For issues with the document, please either fix directly or leave comments on the page.

Note to reviewers: There is a whole set of files which is a version of the Keyword Database. Its a bunch of JSON, but is all text and is used for testing.


jwst/datamodels/tests/test_schema_editor.py
```diff
@@ -0,0 +1,43 @@
+"""Tests for the schema editor
+
+Notes
+-----
+The tests do not have coverage. The original code was written without tests.
+The tests here are for later modifiations.
```
hbushouse: typo: "modifi**c**ations"

jwst/datamodels/tests/test_schema_editor.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+from os import path as ospath
+import pytest
+
+from ..schema_editor import Schema_editor
+
+
+def t_path(partial_path):
+    """Construction the full path for test files"""
```
hbushouse: "Construct"

jwst/datamodels/tests/test_schema_editor.py
```diff
@@ -0,0 +1,43 @@
...
+    keyword_db = t_path(ospath.join('data', 'jwstkd'))
+    return keyword_db
+
+
+def test_just_run(_jail, keyword_db):
+    """Just run the editor"""
+
+    editor = Schema_editor(
+        input=keyword_db,
+        add=True, delete=True, edit=True, rename=True, list=True
```
hbushouse: If the schema editor prompts for input while running, how does the test work?

jwst/datamodels/schema_editor.py
```diff
@@ -480,47 +470,60 @@ class Options:
...
-               ("delete", "fields in the model but not in the keyword db from the model") ,
-               ("edit", "fields found in both to match the values in the keyword db") ,
-               ("rename", "fields in the model to match the names in the keyword db") ,
-               ("list", "changes without making them"),
-               ("query", "for approval of changes before they are made"),
-               ("omit", "")
-            )
+        ("input", "directory name containing keyword database"),
+        ("output", "directory name model schemas will be written to"),
+        ("add", "fields in the keyword db but not in the model to the model"),
```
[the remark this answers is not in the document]
stscieisenhamer: Actually, this is technically right. These were apparently setup to read as one whole string, so that sentence is

"add fields in the keyword db but not in the model to the model"

This is how it appears when doing the "-h" or when that option is shown in the interactive mode.  I'll update to

"add fields in the keyword db but not in the model to the model schema"

jwst/datamodels/tests/test_schema_editor.py
stscieisenhamer: When run from the API, the editor does not run in interactive mode without being passed an `Options` object; see the `scripts/schema_editor` file for details. So, this is basically doing a run to just list what the changes are without actually doing any of them. Simply a canary test.

eslavich:
We're just checking in the keyword dictionary JSONs so that we can exercise the schema editor in the tests, right?  Do we need all of them, or would one file be sufficient?

stscieisenhamer:
Unfortunately, leaving any out will result in all sorts of diffs being generated. Thinking these are just text files, this wouldn't be an issue. However, looking now, I see the folder is rather large, 1.2M. Methinks this will end up going to the artifactory repo.

jwst/datamodels/schema_editor.py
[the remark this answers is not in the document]
hbushouse: Perhaps they would be more readable/intelligible if the object of the action immediately followed the action, e.g "add fields to the model schema that are in the keyword db but not the model", etc.

stscieisenhamer:
@hbushouse I implemented the exclude datamodel schema feature.

jwst/datamodels/schema_editor.py
```diff
@@ -480,47 +491,61 @@ class Options:
...
+        ("input", "directory name containing keyword database"),
+        ("output", "directory name model schemas will be written to"),
+        ("add", "fields that exist in the keyword db but are not in the model schemas"),
+        ("delete", "fields in the model but not in the keyword db from the model"),
+        ("edit", "fields found in both to match the values in the keyword db"),
+        ("rename", "fields in the model to match the names in the keyword db"),
+        ("list", "changes without making them"),
+        ("query", "for approval of changes before they are made"),
+        ("omit_file", "containing the list of parameters to omit", False),
+        ("exclude_file", "containing the list of DataModel schema to exclude from the comparisoni", False)
```
hbushouse: Typo: extra "i" way at the end of "comparison"

stscieisenhamer: Not so simple I guess: its YAML, so each line should begin with the "-", but otherwise one per line, and it is just the file name, i.e. "- core.schema.yaml".  I'll update the confluence page with examples.

Issue #4570, opened 2020-02-13
Correct some instances of PATTTYPE values in ASN rules


stscijgbot:
Issue [JP-1299](https://jira.stsci.edu/browse/JP-1299) was created by Howard Bushouse:

The module `associations/lib/rules_level3.py` has an instance of:
{code:java}
    name='patttype',
    sources=['patttype'],
    value=['2_point|4_point_nod|along_slit_nod'],
{code}
The correct syntax of the value strings is:

`2-point-nod`, `4-point-nod`, and `along-slit-nod`

hbushouse:
Fixed in #4592

stscijgbot:
Comment by Howard Bushouse: Fixed in [https://github.com/spacetelescope/jwst/pull/4592]

PR #4592, opened 2020-02-18
JP-1299 Update PATTTYPE values in ASN rules

ddavis-stsci:
Update rules_level3.py 
value=['2_point|4_point_nod|along_slit_nod']
to
value=['2-point-nod|4-point-nod|along-slit-nod']

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ assign_wcs
 associations
 ------------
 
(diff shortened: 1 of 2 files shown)


hbushouse:
Fixes #4570 / [JP-1299](https://jira.stsci.edu/browse/JP-1299)

PR #4593, opened 2020-02-18
Make DataModel.update no longer update from extra_fits

### LINKED ISSUE #4591: Datamodels.update() should not update from `extra_fits`.
Issue [JP-1310](https://jira.stsci.edu/browse/JP-1310) was created by James Davies:

It seems that FITS keywords that are not in the schema and are passed along to `extra_fits` get copied over via `model.update()` when a new datamodel is updated from an old one that has `extra_fits`.  This is not desirable.  Only keywords in the schema should be copied over.

Also maybe the default `DataModel.save` should not write out anything from `extra_fits`, and the user should be explicit here.  Something like `model.save(filename, extra_fits=True)`.

jdavies-st:
DataModel.update() method will now ignore `extra_fits` in the model that is providing the updated metadata, unless requested via keyword arg `extra_fits=True`.

Resolves #4591

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -948,12 +948,6 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
             if not protected_keyword(path):
                 set_hdu_keyword(self._instance, d, path)
 
-        # Perform updates to extra_fits area of a model
-
-        for hdu_name in hdu_names:
-            path = ['extra_fits', hdu_name, 'header']
-            set_hdu_keyword(self._instance, d, path)
-
         self.validate()
 
     def to_flat_dict(self, include_arrays=True):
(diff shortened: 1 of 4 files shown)


jwst/datamodels/model_base.py
```diff
@@ -948,12 +948,6 @@ def protected_keyword(path):
             if not protected_keyword(path):
                 set_hdu_keyword(self._instance, d, path)
 
-        # Perform updates to extra_fits area of a model
```
hbushouse: I would really like to see this controlled by an optional argument, rather than just eliminating the capability.

jdavies-st:
One thing I'm not entirely clear on in the `update()` method API is what `only` arg does.  The documentation is bad, and the default value of `""` is, um, not good.  @hbushouse do you have any insight what its purpose is?  I feel like there could be an odd interaction between `only` and `extra_fits` args.

hbushouse:
IIRC, the "only" arg was used to copy/update only meta data from the primary header (no extensions).

jdavies-st:
I think the tests now should cover the combos of `only` and `extra_fits`.  And it seems to behave how we expect it to.

I also found a test for this method over in `test_fits.py`.  That test was in the wrong location and was very simple and not very intuitive.  I removed it, and added another test in `test_models.py`.  I think the new tests cover all the use cases and give good coverage to the codebase.

PR #4597, opened 2020-02-20
JP-1201 Remove full path from SCATFILE keyword

### LINKED ISSUE #4387: SCATFILE is incorrectly set to full path of the catalog file
Issue [JP-1201](https://jira.stsci.edu/browse/JP-1201) was created by Jonathan Eisenhamer:

When {{SourceCatalogStep}} is run, the keyword {{SCATFILE}} should be set to the filename of the created catalog. However, it is currently being set to the full path of the catalog file.

ddavis-stsci:
Update assign_wcs so that only the filename is saved in the keyword SCATFILE.

This is tested in test_niriss_wfss.py and the resulting keyword is listed as 
```
# EXTENSION:    0
SCATFILE= 'nir_wfss_cat.ecsv'  / Output source catalog filename

```
Fixes JP-1201
Fixes #4387

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,7 @@
 
 assign_wcs
 ----------
+- Remove full path from SCATFILE keyword [#4387]
 
 - A ``ValueError`` is now raised if input data is missing ``xref_sci`` or ``yref_sci`` keywords. [#4561]
 
--- a/jwst/source_catalog/source_catalog_step.py
+++ b/jwst/source_catalog/source_catalog_step.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+import os
 from ..stpipe import Step
 from ..datamodels import DrizProductModel
 from . import source_catalog
@@ -56,7 +57,8 @@ class SourceCatalogStep(Step):
                 )
                 self.log.info('Wrote source catalog: {0}'
                               .format(cat_filepath))
-                model.meta.source_catalog.filename = cat_filepath
+                model.meta.source_catalog.filename = os.path.basename(
+                    cat_filepath)
 
         # nothing is returned because this is the last step
         return


jwst/assign_wcs/assign_wcs.py
```diff
@@ -62,6 +63,9 @@ def load_wcs(input_model, reference_files={}, nrs_slit_y_range=None):
         wcs = WCS(pipeline)
         output_model.meta.wcs = wcs
         output_model.meta.cal_step.assign_wcs = 'COMPLETE'
+        output_model.meta.source_catalog.filename = os.path.basename(
+            output_model.meta.source_catalog.filename)
```
hbushouse: This meta attribute is only going to be populated for exposures containing WFSS (grism) data; all others will be None. So a check is needed here to see if it exists. That's what's causing all the Travis errors.

hbushouse: Actually, this meta attribute is first populated in the `source_catalog/source_catalog_step.py` module. That's where it should be fixed to not include the full path when it gets populated. Of course it doesn't hurt to have another check here, in case, for example, a user populates it by hand with a full path included.

jwst/assign_wcs/assign_wcs.py
```diff
@@ -62,6 +63,13 @@ def load_wcs(input_model, reference_files={}, nrs_slit_y_range=None):
         wcs = WCS(pipeline)
         output_model.meta.wcs = wcs
         output_model.meta.cal_step.assign_wcs = 'COMPLETE'
+        if hasattr(output_model.meta, 'source_catalog'):
+            try:
+                output_model.meta.source_catalog.filename = os.path.basename(
+                    output_model.meta.source_catalog.filename)
+            except AttributeError:
+                pass
+
```
jdavies-st: This should not be in `assign_wcs.py`.  Setting a source catalog has nothing to do with `assign_wcs`.  I would completely remove this code.  Handle it where it is populated or actually used.  In addition to the two locations I mentioned aboe, it is also read here:

https://github.com/spacetelescope/jwst/blob/d086b9ac4120470e6b1b3c94dd6d786c47e67763/jwst/assign_wcs/util.py#L494

jdavies-st: Finally, if we want to catch the case were users put in bad values by hand into a FITS file, then I think we've lost the battle.  I'm happy to hard error there.  Users should be using a command line argument when running the pipeline if they have a user-supplied source catalog.

PR #4598, opened 2020-02-20
Check for a zero array before sigma clipping

stscieisenhamer:
Resolves [JP-1231](https://jira.stsci.edu/browse/JP-1231)/#4451.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -66,6 +66,8 @@ model_blender
 outlier_detection
 -----------------
 
+- Check for a zero array before sigma clipping [#4598]
+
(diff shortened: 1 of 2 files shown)


jdavies-st:
An upstream suggestion would be to have `astropy.stats.sigma_clipped_stats` to use only masks and not use the nan functions.  I.e. mask the nans on the way in, OR them with whatever else masked, and then do the stats using the non-nan functions.

stscieisenhamer:
WRT the astropy usage: This may actually be an astropy bug, due to some recent comments left on the related [numpy issue 12283](https://github.com/numpy/numpy/issues/12283), as @jdavies-st suggest.

PR #4599, opened 2020-02-20
Align to gaia jp 1184

stsci-hack:
The 'tweakreg' Step has been updated to perform an additional fit to the GAIA astrometric catalog.  Changes include:

  - A new module, astrometric_utils.py, was added to get the astrometric catalog from a web service
  - The fit to the catalog was added as an optional section of code after the relative fit completes
  - New parameters were added to the 'tweakreg.cfg' to control this fitting process

This PR addresses [PR JP-1184](https://jira.stsci.edu/browse/JP-1184). 

Testing of this PR will require Level 3 test data that simulates an actual field of the sky for fitting to the GAIA catalog.


jwst/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,179 @@
+import os
+import requests
+
+from astropy import table
+from astropy.table import Table
+from astropy.coordinates import SkyCoord
+from astropy import units as u
+
+from resample import resample_utils
```
hbushouse: Travis builds are failing on this. I think it needs to be "from ..resample import resample_utils" (i.e. a relative import)

jwst/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,179 @@
...
+
+Primary function for creating an astrometric reference catalog.
+
+"""
+
+
+def create_astrometric_catalog(input_models, catalog="GAIADR2", output="ref_cat.ecsv",
+                               gaia_only=False, table_format="ascii.ecsv",
+                               existing_wcs=None, num_sources=None):
+    """Create an astrometric catalog that covers the inputs' field-of-view.
```
hbushouse: Is this the field-of-view that is the union of all inputs?

jwst/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,179 @@
...
+        documented on the catalog web page.
+
+    output : str, optional
+        Filename to give to the astrometric catalog read in from the master
+        catalog web service.  If None, no file will be written out.
+
+    gaia_only : bool, optional
+        Specify whether or not to only use sources from GAIA in output catalog
+
+    existing_wcs : `~stwcs.wcsutil.HSTWCS`
```
hbushouse: Yeah, we probably aren't going to be feeding in any HSTWCS objects in the JWST pipeline ...

jwst/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,179 @@
...
+    Returns
+    -------
+    ref_table : `~astropy.table.Table`
+        Astropy Table object of the catalog
+
+    """
+
+    # start by creating a composite field-of-view for all inputs
+    # This default output WCS will have the same plate-scale and orientation
+    # as the first chip in the list, which for WFPC2 data means the PC.
```
hbushouse: This may need updating for JWST context. Probably something relevant to which, if any, NIRCam detector is normally first in the list. But if it's just the list of members as specified in the ASN table that's input to the `calwebb_image3` pipeline, the order can be semi-random.

jwst/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,179 @@
...
+            g = "-1"  # indicator for no source ID extracted
+        gaia_sources.append(g)
+
+    gaia_col = table.Column(data=gaia_sources, name='GaiaID', dtype='U25')
+    ref_table.add_column(gaia_col)
+
+    # sort table by magnitude, fainter to brightest
+    ref_table.sort('mag', reverse=True)
+
+    if num_sources is not None:
```
hbushouse: Would be nice to have a comment here that once again explains the significance of `num_sources` being negative and what happens to the list in that case.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -200,6 +207,61 @@ def process(self, input):
...
+                    output_name = 'fit_{}_ref.ecsv'.format(self.gaia_catalog.lower())
+                else:
+                    output_name = None
+                ref_cat = amutils.create_astrometric_catalog(self.input_models,
+                                                             self.gaia_catalog,
+                                                             output=output_name)
+
+                # Check that there are enough GAIA sources for a reliable/valid fit
+                num_ref = len(ref_cat)
+                if num_ref < self.minobj:
```
hbushouse: I guess this is using the same `minobj` step param that's used to do the existing relative alignment. Is there any reason why the two different processes might want/need to have their own unique param values (e.g. perhaps fewer sources are necessary to get a good absolute fit, than a relative fit)?

jwst/tweakreg/tweakreg_step.py
```diff
@@ -200,6 +207,61 @@ def process(self, input):
...
+                    msg = "Not enough GAIA sources for a fit: {}\n".format(num_ref)
+                    msg += "Skipping alignment to {} astrometric catalog!\n".format(self.gaia_catalog)
+                    raise ValueError(msg)
```
hbushouse: Is this raising of an error going to cause the step and pipeline to halt, or is the error trapped up above, so that the rest of the pipeline can proceed without any modifications to the WCS? What we prefer is the latter (in which case this might need to be changed to just a warning message, rather than raising an error).

jwst/tweakreg/tweakreg_step.py
```diff
@@ -200,6 +207,61 @@ def process(self, input):
...
+                # Set group_id to same value so all get fit as one observation
+                for imcat in imcats:
+                    imcat.meta['orig_group_id'] = imcat.meta['group_id']
+                    imcat.meta['group_id'] = 987654
```
hbushouse: Any particular significance to that number or just picked it at random?

jwst/tweakreg/tweakreg_step.py
```diff
@@ -55,8 +56,14 @@ class TweakRegStep(Step):
         fitgeometry = option('shift', 'rscale', 'general', default='general') # Fitting geometry
         nclip = integer(min=0, default=3) # Number of clipping iterations in fit
         sigma = float(min=0.0, default=3.0) # Clipping limit in sigma units
+
+        # Absolute astrometry fitting parameters:
+        align_to_gaia = boolean(default=True)  # Align to GAIA catalog
+        gaia_catalog = option('GAIADR2', 'GAIADR1', default='GAIADR2')
+        output_gaia = boolean(default=False)  # Write out GAIA catalog as a separate product
```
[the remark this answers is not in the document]
hbushouse: And now that you mention it, given that the web service is expanding to include catalogs other than Gaia, should we even be hardwiring the name "gaia" into the step arguments? Perhaps something more generic (not sure what).

jwst/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,179 @@
...
+    # This default output WCS will have the same plate-scale and orientation
+    # as the first chip in the list, which for WFPC2 data means the PC.
+    # Fortunately, for alignment, this doesn't matter since no resampling of
+    # data will be performed
+    if existing_wcs is not None:
+        outwcs = existing_wcs
+    else:
+        outwcs = resample_utils.make_output_wcs(input_models)
+    radius = compute_radius(outwcs)
+    ra, dec = outwcs.wcs.crval
```
nden: If `outwcs`  is the output of `resample_utils.make_output_wcs` then this is a GWCS object which has no attributes `wcs` or `wcs.crval`. However, there are still ways to find out the fiducial on the sky.

jwst/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,179 @@
...
+"""
+
+Utility functions for creating an astrometric reference catalog.
+
+"""
+
+def compute_radius(wcs):
+    """Compute the radius from the center to the furthest edge of the WCS."""
+
+    ra, dec = wcs.wcs.crval
```
nden: This assumes FITS WCS which is probably not correct.

jwst/tweakreg/tweakreg_step.py
stsci-hack: This Exception gets trapped at line 259 (or thereabouts) so that the only result is a warning being reported in the log and the rest of the processing proceeds normally with no changes to the WCS.

jwst/tweakreg/tweakreg_step.py
stsci-hack: This number is a random large number unlikely to be used in any program for the group_id identifier within any association.  The distinct value also makes it clear that it is an artificially assigned ID which can be recognized as being associated with alignment to GAIA as opposed to the group_id's used for relative alignment of the same observations performed earlier in this step.

jwst/pipeline/tweakreg.cfg
```diff
@@ -18,3 +18,6 @@ yoffset = 0.0
 fitgeometry = 'general'
 nclip = 3
 sigma = 3.0
+align_to_gaia = True
```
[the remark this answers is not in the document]
stsci-hack: We will just need to make sure that some test gets defined with data suitable for turning this step on so that it can be tested with the rest of the code.  This will admittedly need to wait for suitable data to be made available before this can happen.

jwst/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,179 @@
...
+
+    """
+    service_type = 'vo/CatalogSearch.aspx'
+    spec_str = 'RA={}&DEC={}&SR={}&FORMAT={}&CAT={}&MINDET=5'
+    headers = {'Content-Type': 'text/csv'}
+    fmt = 'CSV'
+
+    spec = spec_str.format(ra, dec, sr, fmt, catalog)
+    service_url = '{}/{}?{}'.format(SERVICELOCATION, service_type, spec)
+    rawcat = requests.get(service_url, headers=headers)
```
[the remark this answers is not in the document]
stsci-hack: I added this dependency to setup.py pinned to a minimum version which we have used for the last year with HST.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -55,8 +56,14 @@ class TweakRegStep(Step):
         fitgeometry = option('shift', 'rscale', 'general', default='general') # Fitting geometry
         nclip = integer(min=0, default=3) # Number of clipping iterations in fit
         sigma = float(min=0.0, default=3.0) # Clipping limit in sigma units
+
+        # Absolute astrometry fitting parameters:
```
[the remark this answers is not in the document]
stsci-hack: The spaces and comments will be removed to create a less understandable spec definition for this step.  May I then ask what should be done about the other JWST pipeline steps where this formatting was also done (see skymatch, for example)?

jwst/tweakreg/tweakreg_step.py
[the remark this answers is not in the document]
stsci-hack: I would argue that referring to these catalogs as gaia_catalog remains perfectly valid, even if we start using the updated PAN_STARRS catalog.  The primary reason is that ALL the absolute astrometry catalogs with improved accuracy we use or plan to use have been put into the GAIA frame.  Furthermore, merely implying use of GAIA provides an indication that this object refers to something related to absolute astrometry on the sky as opposed to merely a relative or instrumental catalog.  As a result, use of the term 'gaia' serves 2 purposes which support the correct use and understanding in the code.

jwst/tweakreg/astrometric_utils.py
stsci-hack: I revised this based on usage of the WCS objects in assign_wcs itself.  Now we just need some suitable complete WCS information from simulated data to make sure it works as expected.

jwst/tweakreg/astrometric_utils.py
stsci-hack: Again, this has been revised using the same information that assign_wcs relies on for its computations.

jwst/tweakreg/tweakreg_step.py
stsci-hack: A separate parameter, min_gaia, was added to allow fields with fewer GAIA sources to be aligned.  The default value of 5 is based on evaluation of the HST fits to GAIA, where we are able to fit >90% of the fields when we have at least 5 GAIA sources in the field.

jwst/tweakreg/astrometric_utils.py
stsci-hack: Done.  Now devs won't need to keep going back to the docstring to understand what is going on here...

jwst/tweakreg/astrometric_utils.py
stsci-hack: Done...

jwst/tweakreg/astrometric_utils.py
stsci-hack: Fixed. It now refers to it as a GWCS object.

jwst/tweakreg/astrometric_utils.py
stsci-hack: This function determines the full field-of-view for all the input models based on the available WCS information, and requests all the GAIA sources that fall in that field-of-view from the web service.  So, yes, it is intended to be the union of all the inputs.

jwst/tweakreg/astrometric_utils.py
stsci-hack: Fixed.

stsci-hack:
All comments have been addressed as well as possible.  No suitable or viable test data was available to test these changes, so it is relying solely on the current automated tests to make sure it doesn't break currently implemented functionality.  **This PR should only be merged after suitable test data can be provided and used to test these changes.**

stsci-hack:
I believe all the comments have been addressed with the commits 0472ea9 and eb40195.  The code, as of 0472ea9, can successfully generate the GAIA catalog for the output field of view.  This was demonstrated using the 'mosaic_long_asn2.json' ASN and images from the regression test data provided by @hbushouse .  The resulting GAIADR2 catalog was confirmed by displaying one of the input images in DS9 and reading in the reference catalog as a region file to generate the figures:
[Central region overlaid with GAIA sources](https://ssb.stsci.edu/hack/astrometry/nrca5_47Tuc_subpix_dither1_vs_GAIA.png)
[Full frame overlaid with GAIA sources](https://ssb.stsci.edu/hack/astrometry/nrca5_47Tuc_subpix_dither1_vs_GAIA_fullframe.png)

The second figure shows how the current catalog gets generated using a cone-search centered on the overall field of view.  This can be revised to use the footprint instead at a later date.  

Testing still needs to be done to insure that this catalog gets used correctly, and that the overall fit to this catalog gets done properly.  However, given a little more time, the current regression test data may be sufficient for this purpose.

hbushouse:
Nice! The figure of the central region shows what appears to be a consistent offset of all the Gaia sources relative to actual sources in the image, which I'm assuming shows that there is in fact an offset in the absolute WCS of the image?  Which would then get corrected once the solution is computed and applied?

stscicrawford:
One question I have -- for right now, would it make sense to set `align_to_gaia` to a default of False?   This seems like it could use more testing before being activated in the pipeline by default.

stsci-hack:
> One question I have -- for right now, would it make sense to set `align_to_gaia` to a default of False? This seems like it could use more testing before being activated in the pipeline by default.

This alignment step has been turned off by default in the 'tweakreg.cfg' file (line 21), if the code is actually checking that variable correctly...

stsci-hack:
> Nice! The figure of the central region shows what appears to be a consistent offset of all the Gaia sources relative to actual sources in the image, which I'm assuming shows that there is in fact an offset in the absolute WCS of the image? Which would then get corrected once the solution is computed and applied?

The offset seen will be interpreted as an offset in the absolute astrometry which should be identified and corrected by this step.  That makes this data appropriate for testing this new functionality.

stsci-hack:
The commit a4cbc26 modifies the tweakreg step to correctly call 'tweakwcs.align_wcs' for the fit to GAIA after performing the relative fit amongst the images.  The fit, though, only gets applied by tweakwcs to the GWCS object in each input exposure, so the 'to_fits_sip()' of the GWCS object (added in GWCS version 0.13) to update the FITS WCS representation as well. This allowed the test data to be updated based on the fit to the GAIA reference frame, although issues still remain in getting the final alignment to GAIA to better than 1 pixel as seen [this closeup of the center of the drizzled product](https://ssb.stsci.edu/hack/astrometry/center_vs_gaia.jpeg) with the GAIA sources overlaid on it.

hbushouse:
Is the inaccuracy of the alignment shown in the image referenced above due to actual inaccuracies in the tweakwcs solution for Gaia sources or just a shortcoming of the updated FITS WCS used to create the display?

setup.py
```diff
@@ -102,6 +102,7 @@ def get_transforms_data():
         'jsonschema>=2.3,<4',
         'numpy>=1.16',
         'photutils>=0.7',
+        'requests>=2.22',
```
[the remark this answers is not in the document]
stsci-hack: This pin for 'requests' is already in the 'requirements-sdp.txt' file.  I am not sure if anything else needs to be done...

jwst/tweakreg/tweakreg_step.py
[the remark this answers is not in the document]
stsci-hack: Good idea.  This has been changed as requested in e9076e7.

jwst/tweakreg/tweakreg_step.py
stsci-hack: Fixed in 1df551e (subject to continued testing...).

stsci-hack:
@jdavies-st You pointed out a number of things which definitely improve this code.  I have added (as of e9076e7) a unit test for each of the functions in 'astrometric_utils'.  However, I am still trying to get this code to update the header of the image correctly, so the regression test for this step will have to wait until I believe the code works right.  Otherwise, all comments made by all reviewers should be addressed.

jdavies-st:
Thanks Warren.  Changes look good.  I did look at this briefly yesterday, and I'll add a couple comments inline.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -200,12 +198,75 @@ def process(self, input):
...
+                #       This value should also be translated to the FITS WCSNAME keyword
+                #       IF that is what gets recorded in the archive for end-user searches
+                imcat.meta['image_model'].meta.wcs.name = "FIT-LVL3-{}".format(self.gaia_catalog)
+                
+                # Also update FITS representation 
+                gwcs_header = imcat.wcs.to_fits_sip(max_pix_error=0.1, 
+                                                max_inv_pix_error=0.1,
+                                                degree=3, 
+                                                npoints=128)
+                imcat.meta['image_model'].wcs = wcs.WCS(header=gwcs_header)
```
jdavies-st, line 269: I think this block is what is causing the problem in the existing regression tests.  If I delete this whole block, everything works fine.  Line 256 updates the `gwcs` object, and that's all that needs to be done.  We don't use or update any FITS WCS objects until we get rectified, resampled images, so we don't need to do that here.  And we are not currently not using the `gwcs.WCS.name` attribute, but we could in the future.  But any name including `FIT` and `GAIA` should be controlled by `self.align_to_gaia` of the step.

setup.py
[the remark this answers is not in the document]
jdavies-st: Ah, ok!  I forgot that was already in there, pulled in by our optional `[test]` dependenciesd, i.e. `ci_watson`.  Thanks for checking.

jwst/tweakreg/tweakreg_step.py
stsci-hack: This block of code deals with updating the input images's FITS WCS, something which apparently is not required.  It has been commented out in the latest commit, while the addition of the '.name' attribute to the WCS has been (rightfully) moved under the logic for successful GAIA fit only.  I would argue that this is required as the ONLY indication to the end-user (short of diving into the GWCS and hoping to recognize one of the elements there) that any additional alignment to GAIA has been performed on the input GWCS objects.  It is a key feature of the HST alignment strategy with INS and end-users already relying on that name.

hbushouse: As of right now we aren't including a FITS SIP representation in the unrectified products, so you're right that this update is not needed right now. There's a ticket in the backlog to add the SIP, so this code code be reactivated whenever that gets implemented.

jwst/tweakreg/tweakreg_step.py
mcara: `imcat.meta['group_id']` is used exclusively internally for alignment purpose and it is not stored anywhere in the `DataModel`. It can be anything. Setting it to 1 would be just as good as setting it to 987654.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -200,13 +202,94 @@ def process(self, input):
...
+                          fitgeom=self.fitgeometry,
+                          nclip=self.nclip,
+                          sigma=(self.sigma, 'rmse')
+                         )
+
+                # Reset group_id to original values
+                # Also, update/create the WCS .name attribute with information 
+                #  on this astrometric fit as the only record that it was successful
+                for imcat in imcats:
+                    imcat.meta['group_id'] = imcat.meta['orig_group_id']
```
mcara, line 267: I think this `'group_id'` is not used anywhere in the pipeline or stored/recorded in the `DataModel` and so there is no need to "restore" this id.

stsci-hack:
Change log entry has been added.

CHANGES.rst
```diff
@@ -174,6 +174,8 @@ tweakreg
 
 - Updated step arguments in the documentation. [#4723]
 
+- Added align_to_gaia processing as an option [#4599]
```
hbushouse, line 177: This needs to go up at the top of the file, under the 0.16.1 release (0.16.0 has already gone out).

stsci-hack: Rebased to latest version of JWST.   Updated correct version of CHANGES with this message under v0.16.1 in commit 78f34b6

mcara:
@stsci-hack I meant files from this PR should pass `flake8` - not files from _other_ packages. I do not think this PR should modify unrelated files but in the end this is up to the maintainers.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -200,13 +202,92 @@ def process(self, input):
...
+                          enforce_user_order=False,
+                          expand_refcat=False,
+                          minobj=self.minobj,
+                          match=tpmatch_gaia,
+                          fitgeom=self.fitgeometry,
+                          nclip=self.nclip,
+                          sigma=(self.sigma, 'rmse')
+                         )
+
+                # Reset group_id to original values
```
mcara, line 262: This comment is no longer valid

stsci-hack: Removed in c344917

jwst/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def create_astrometric_catalog(input_models, catalog="GAIADR2", output="ref_cat.ecsv",
+                               gaia_only=False, table_format="ascii.ecsv",
+                               existing_wcs=None, num_sources=None):
+    """Create an astrometric catalog that covers the inputs' field-of-view.
+
+    Parameters
+    ----------
+    input : str, list
```
nden, line 34: The actual parameter is called `input_models`, not `input` and I believe is supposed to be a list of datamodels, not file names.

jwst/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,184 @@
...
+        Filenames of images to be aligned to astrometric catalog
+
+    catalog : str, optional
+        Name of catalog to extract astrometric positions for sources in the
+        input images' field-of-view. Default: GAIADR2. Options available are
+        documented on the catalog web page.
```
nden, line 40: Where is this page? Is GAIA the only option?

jwst/tweakreg/astrometric_utils.py
```diff
@@ -0,0 +1,184 @@
...
+
+    output : str, optional
+        Filename to give to the astrometric catalog read in from the master
+        catalog web service.  If None, no file will be written out.
+
+    gaia_only : bool, optional
+        Specify whether or not to only use sources from GAIA in output catalog
+
+    existing_wcs : model
```
nden, line 49: The type here should be `~gwcs.wcs.WCS` in single back ticks. It would be clearer to add a sentence about what it is used for.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -200,13 +202,91 @@ def process(self, input):
...
                 imcat.meta['image_model'].meta.wcs = imcat.wcs
 
+                """
+                # Also update FITS representation in input exposures for
+                # subsequent reprocessing by the end-user.
+                # Not currently enabled, but may be requested later...
+                gwcs_header = imcat.wcs.to_fits_sip(max_pix_error=0.1,
+                                                max_inv_pix_error=0.1,
+                                                degree=3,
+                                                npoints=128)
```
nden, line 286: How could we update headers when a number of questions of how to do it are not answered? The first that comes to mind is that SIP writes out CD matrix. JWST uses and has in the headers a PC matrix. Would this write both and what does this mean?

jwst/tweakreg/tweakreg_step.py
```diff
@@ -200,13 +202,91 @@ def process(self, input):
...
+                        #       IF that is what gets recorded in the archive for end-user searches
+                        imcat.meta['image_model'].meta.wcs.name = "FIT-LVL3-{}".format(self.gaia_catalog)
+
         for imcat in imcats:
             imcat.meta['image_model'].meta.cal_step.tweakreg = 'COMPLETE'
             # retrieve fit status and update wcs if fit is successful:
-            fit_info = imcat.meta.get('fit_info')
-            if fit_info['status'] in 'SUCCESS':
+            if 'SUCCESS' in imcat.meta.get('fit_info')['status']:
                 imcat.meta['image_model'].meta.wcs = imcat.wcs
```
nden, line 277: Just for my understanding: is `imcat.wcs` a FIST WCS object or a GWCS object?

jwst/tweakreg/tweakreg_step.py
```diff
@@ -200,13 +202,91 @@ def process(self, input):
...
 
+                """
+                # Also update FITS representation in input exposures for
+                # subsequent reprocessing by the end-user.
+                # Not currently enabled, but may be requested later...
+                gwcs_header = imcat.wcs.to_fits_sip(max_pix_error=0.1,
+                                                max_inv_pix_error=0.1,
+                                                degree=3,
+                                                npoints=128)
+                imcat.meta['image_model'].wcs = wcs.WCS(header=gwcs_header)
```
nden, line 287: Where is wcs.WCS imported from here? I'm probably not seeing it.

jwst/engdblog/tests/test_engdblog.py
```diff
@@ -32,7 +32,7 @@ def test_barestring(caplog, engdb):
     assert mnemonic in result
     assert 'EngDBLogStep instance created' in caplog.text
     assert mnemonic in caplog.text
-    assert "Step EngDBLogStep running with args ('INRSI_GWA_Y_TILT_AVGED',).".format(mnemonic) in caplog.text
+    assert "Step EngDBLogStep running with args ('INRSI_GWA_Y_TILT_AVGED',{}).".format(mnemonic) in caplog.text
```
jdavies-st, line 35: Please remove this change.  This has been fixed elsewhere and is not related to this PR.

jwst/tweakreg/tweakreg_step.py
jdavies-st: Yeah, this whole block has been commented out with `"""`.  (Easy to miss).  We discussed this above in a previous round of review. Agree we don't want to do this right now.

hbushouse:
OK, @jdavies-st reports this PR branch got a clean regtest run after rebasing to master, so I'm merging (after long last).

hbushouse:
Remaining issues with non-optimal alignment of images using the Gaia catalog will be addressed separately in the future.

PR #4602, opened 2020-02-21
JP-1318: Keyword updates

hbushouse:
First round of updates to implement actions documented in JP-1265. Still waiting on a few outstanding questions to be answered from SDP in order to determine what further updates, if any, we need to make to our datamodels schemas.

Note that the duplicate keyword definitions that used to be in both `guider_raw.schema` and `guider_cal.schema` have now been consolidated into a single new schema called `guider_meta.schema`, which is referenced/included by both of the higher-level schemas.

Fixes [JP-1318](https://jira.stsci.edu/browse/JP-1318)


jwst/datamodels/schemas/extract1dimage.schema.yaml
```diff
@@ -22,7 +22,7 @@ allOf:
       type: integer
       fits_keyword: SPORDER
       fits_hdu: SCI
-    dispersion_axis:
+    dispersion_direction:
```
jdavies-st: What's the reason for this change?  I think `dispersion_axis` is much more descriptive.  The value can be 1 or 2, which is the FITS 1-indexed axis of the dispersion.  It's not really a direction.  It's an axis.

jwst/datamodels/schemas/multiextract1d.schema.yaml
```diff
@@ -26,7 +26,7 @@ allOf:
             type: integer
             fits_keyword: SPORDER
             fits_hdu: SCI
-          dispersion_axis:
+          dispersion_direction:
```
jdavies-st: Some comment here.  The title should probably reflect "axis" as well.

jwst/datamodels/schemas/extract1dimage.schema.yaml
jdavies-st: Ah, I see it is `dispersion_direction` in `wcsinfo`.  I would suggest changing it there (and where ever it used in the code) to `disperision_axis`.

hbushouse: When it was first implemented in both the KWD and our schema (wcsinfo) it was entered as `dispersion_direction`. When it then got added later to other schema, it was done using a different name. So this change is simply an effort to make them all consistent. It's easiest for us to change the inconsistencies in our schemas, rather than having the KWD change to match us.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -601,26 +601,52 @@ properties:
...
-              - type: number
-              - fits_keyword: EXPMID
-                title: UTC exposure mid time
-                blend_table: True
-                blend_rule: mean
+            title: UTC exposure mid time
+            type: number
+            fits_keyword: EXPMID
+            blend_table: True
+            blend_rule: mean
```
jdavies-st: Do we want to blend these in different ways?
```
mid_time:
  blend_rule: mean
```
and
```
mid_time_mjd:
  blend_rule: first
```
Do we even want to blend these?  These may not have any meaning in a combined image.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -601,26 +601,52 @@ properties:
...
+            fits_keyword: EXPEND
+            blend_table: True
+            blend_rule: last
+          end_time_mjd:
+            title: "[d] exposure end time in MJD"
+            type: number
+            fits_keyword: MJD-END
+            fits_hdu: SCI
+            blend_table: True
+            blend_rule: first
```
jdavies-st: Same `blend_rule` issue here.  Maybe this should be `last`?

jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -118,24 +118,24 @@ allOf:
...
-          time_scale:
+          time_sys:
             title: "principal time system for time-related keywords"
             type: string
             default: "UTC"
             fits_keyword: TIMESYS
             fits_hdu: EXTRACT1D
-          start_utc:
-            title: "UTC at start of integration [MJD]"
+          start_time_mjd:
```
jdavies-st: `start_time_mjd` and the others are also in the core schema under the same name and same FITS keyword name.  Here they are just mapped to the EXTRACT1D HDU.  I think this is fine, but you had mentioned in another context that this was undesirable (RA_TARG, DEC_TARG).  Thoughts?

philhodge: I didn't think these keywords (MJD-BEG, etc.) were used elsewhere, but maybe I just didn't see them.  These values are copied from the INT_TIMES table data to keywords in the extract_1d extension headers.

hbushouse: You're right that they weren't used previously, but there was a recent (external) request to add them to the SCI extension headers of all products, because they conform to the FITS time standard. They have the same values as our EXPSTART, EXPMID, EXPEND keywords, but those keywords don't conform to the standard. So now that the MJD-XXX keywords have been added to the core schema for SCI extensions, I wanted to make the definitions consistent in all places.

jwst/datamodels/schemas/core.schema.yaml
hbushouse: In a combined image EXPSTART/MJD-BEG should give the starting time of the first blended exposure, EXPEND/MJD-END give the end time of the last blended exposure, and EXPMID/MJD-MID give the mid-point. So yes they can be useful in a blended image header.

hbushouse: But I see that I mistakenly set all of the MJD-XXX keywords to blend_rule: first. I'll fix that. Thanks for catching!

jwst/datamodels/schemas/multispec.schema.yaml
hbushouse: The entry in core.schema places them in SCI extension headers, while this one puts them in EXTRACT1D extension headers. And since you never have both kinds of extensions in the same file, the keywords will only appear once.

PR #4603, opened 2020-02-21
Refactor WFSS transforms

nden:
Improved performance.

This PR uses the `expected values` in the wfss tests (assign_wcs/tests/test_wfss.py) but changes the calls to the functions to use the new interface.

### DIFF
--- a/jwst/assign_wcs/tests/test_wfss.py
+++ b/jwst/assign_wcs/tests/test_wfss.py
@@ -26,10 +26,6 @@ def create_slit(model, x0, y0, order):
     model = Mapping((0, 1, 0, 0, 0)) | (Shift(xmin) & Shift(ymin) &
                                         Const1D(x0) & Const1D(y0) & Const1D(order) )| model
     wcsobj = wcs.WCS([('det', model), ('world', None)])
-    # refactored call
-    # wcsobj.bounding_box = ((3, 7), (2, 5), (x0-1, x0+1), (y0-1, y0+1), (1, 3))
-
-    #wcsobj.bounding_box = ((xmin, xmax), (ymin, ymax))
     wcsobj.bounding_box = ((20, 25), (800, 805))
     slit = SlitModel()
     slit.meta.wcs = wcsobj
@@ -194,14 +190,6 @@ def test_NIRISSBackwardDispersion():
 
     slit = create_slit(forward_model, x0, y0, order)
     slit.meta.wcs.bounding_box = ((910, 916), (12,18))
-    #grid = grid_from_bounding_box(slit.meta.wcs.bounding_box)
-    #shape = grid[0].shape
-    #xdx = np.empty(shape, dtype=np.float64)
-    #ydy = np.empty(shape, dtype=np.float64)
-    #for j in range(shape[0]):
-        #for i in range(shape[1]):
-            #xdx[j, i], ydy[j, i], _, _, _ = model(grid[0][j, i], grid[1][j, i],
-                                                  #wavelength[j, i], 1)
 
     expected_xdx = np.array(
         [[641.69045022, 642.69044108, 643.69043194, 644.6904228, 645.69041366, 646.69040451, 647.69039537],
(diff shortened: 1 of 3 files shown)


jwst/assign_wcs/tests/test_wfss.py
```diff
@@ -235,21 +224,21 @@ def test_NIRISSBackwardDispersion():
          [15.66666438, 15.6661858 , 15.66570726, 15.66522876, 15.66475031, 15.66427189, 15.66379352]])
 
     # refactored call
-    # x, y = wcstools.grid_from_bounding_box(slit.meta.wcs.bounding_box)
-    # xdx, ydy, _, _ = model(x, y, wavelength, np.zeros(x.shape)+1)
+    x, y = grid_from_bounding_box(slit.meta.wcs.bounding_box)
+    xdx, ydy, _, _, _ = model(x, y, wavelength, np.zeros(x.shape)+1)
```
hbushouse: Up above in line 150 you changed this call to use `order` in place of the hardcoded value "1". Is the same change warranted here?

jwst/extract_2d/grisms.py
```diff
@@ -551,11 +551,13 @@ def compute_wavelength_array(slit):
...
+    #shape = grid[0].shape
+    #wavelength = np.empty(shape, dtype=np.float64)
     transform = slit.meta.wcs.forward_transform
-    for j in range(shape[0]):
-        for i in range(shape[1]):
-            wavelength[j, i] = transform(grid[0][j, i], grid[1][j, i])[2]
+    #for j in range(shape[0]):
+        #for i in range(shape[1]):
+            #wavelength[j, i] = transform(grid[0][j, i], grid[1][j, i])[2]
+    x, y = grid_from_bounding_box(slit.meta.wcs.bounding_box)
```
hbushouse: Woot!!! No more looping over pixels!

jwst/transforms/models.py
```diff
@@ -1406,25 +1444,44 @@ def evaluate(self, x, y, wavelength, order):
...
-        dy = self.ymodels[iorder][0](x, y) + t * self.ymodels[iorder][1](x, y)
-        # rotate by theta
+        xmodel = self.xmodels[iorder]
+        ymodel = self.ymodels[iorder]
+        lmodel = self.lmodels[iorder]
+
+        dx = xmodel[0](x, y) + t * xmodel[1](x, y)
+        dy = ymodel[0](x, y) + t * ymodel[1](x, y)
+
+        ## use that t to compute the dx and dy
```
hbushouse: Are all of these commented blocks new code that will be activated in the future?

nden:
One of the tests is failing the comparison with the previous version of the transforms. The refactored code uses an interpolation routine from scipy while the previous code used the one in numpy setting  wavelength values outside the bounding_box to the left and right corner. The current one cannot do this. Values outside the bounding box can be set to either NaN or extrapolated. I'd argue one of these is better than setting the wavelength to a known incorrect value.

jwst/assign_wcs/tests/test_wfss.py
nden: Yes, I'll change it. I'm running the tests for order=1 only.

jwst/transforms/models.py
nden: Everything is activated now, I will clean the code.

hbushouse:
> One of the tests is failing the comparison with the previous version of the transforms. The refactored code uses an interpolation routine from scipy while the previous code used the one in numpy setting wavelength values outside the bounding_box to the left and right corner. The current one cannot do this. Values outside the bounding box can be set to either NaN or extrapolated. I'd argue one of these is better than setting the wavelength to a known incorrect value.

Seems to me we have (or had) places downstream where wavelength values are set to NaN if they're outside of the bounding_box or are otherwise unable to be calculated for any reason, so there's precedence for this approach. @hodge can you confirm? I would vote for setting them to NaN, rather than setting them a bad value.

philhodge:
There are some places where we assume that wavelengths (or right ascension or declination) will be NaN for pixels that are outside the bounding box.  It's useful to have an unambiguous and simple test for points that are outside the bounding box.  In `photom`, we temporarily set elements of the conversion factor to NaN for pixels that have NaN wavelengths, then check for NaN in the array of conversion factors and replace those with zero.

PR #4604, opened 2020-02-21
use transform-1.2.0 in schemas

nden:
The schemas in datamodels were updated to use transform version 1.2.0. This fixes the problem with reading the Nirspec disperser reference file [seen here](https://travis-ci.org/spacetelescope/jwst/jobs/653521852#L1455) in our CI testing that uses dev versions of `asdf` and `astropy`.

The schemas in jwst/transforms will be dealt with separately.


jdavies-st:
This PR will be required for `jwst` to work with `astropy 4.0.1` patch when it is released (soon).

PR #4605, opened 2020-02-24
JP-1320: Allow FileNotFoundError to be raised in filetype.check; add test

jwhite3:
This PR makes the Traceback more specific between FileNotFoundError and when `filetype.check` raises a ValueError when it can't determine the filetype. 

Resolves [JP-1320](https://jira.stsci.edu/browse/JP-1320)


jwst/datamodels/tests/test_filetype.py
```diff
@@ -0,0 +1,18 @@
...
+def fake_pipeline():
+    return Spec2Pipeline()
+
+
+def test_pipeline_raises_filenotfound(fake_pipeline, capsys):
+    with pytest.raises(RuntimeError):
+        fake_pipeline.run('file_does_not_exist.fits')
+
+    assert (
+            "FileNotFoundError: [Errno 2] No such file or directory: 'file_does_not_exist.fits'"
```
philhodge: Why not just check for "FileNotFoundError"?

jdavies-st: I think the `Spec2Pipeline` catches `FileNotFoundError` and raises a `RuntimeError`.  Not great.  That needs to be fixed.

Perhaps simpler test that doesn't mix up `jwst.datamodels` and `jwst.pipeline` in the tests would be to just to open a non-existent file using `datamodels.open` and one should get a nice `pytest.raises(FileNotFoundError)`?

jwhite3: Turns out that error is captured and printed for the pipeline that I used (not sure if this is the case for all pipelines). For `Spec2Pipeline` if there are any errors during the course of executing the pipeline, the error is printed and then at the end a `RuntimeError` is raised. No other way to access it other than looking for it in the `stdout`

jwhite3: Also the reason that I used that pipeline in the test is because in the issue, JP-1320, that was where the clarity was desired.

jwhite3: @jdavies-st That's a good idea and preferable for sure. I wasn't sure where `check` was used, so I just defaulted to using a pipeline. I'll switch it out

philhodge: I just meant that you don't need to search for the entire string, that "FileNotFoundError" should be sufficient.  Searching on the entire string seems fragile to me.  For example, what if the error number differs on different operating systems?  What if in some version of Python they decide to use double quotes instead of single quotes around the file name?  What if they add a period after "Errno 2]"?  What if they decide to change from "No such file or directory" to "File <name> does not exist"?

jwhite3: @philhodge Yeah I see you're point, but at the same time I think a similar argument could be made in that looking for just `FileNotFoundError` is non-specific. What if there's more than one of those for some reason?

jwhite3: Either way, I think that issue will be solved by using `datamodels.open`

jwhite3:
@hbushouse @philhodge I thought better of it and just removed the specific case where the file isn't opened if it ends with "fits" or "asdf". This way, if the file doesn't actually exist, `FileNotFoundError` will be raised for any extension for the input string.

jwhite3:
Also, can anyone think of a reason that for input file strings with "fits" or "asdf" extensions why the `filetype` would be returned even if the file didn't exist? Is there any case where a DataModel would need to be created with a fake input path and work?

jwst/datamodels/tests/test_filetype.py
jdavies-st: Yeah, `Spec2Pipeline` captures all exceptions and then re-emits a `RuntimeError`.  This is bad needs fixing.

https://github.com/spacetelescope/jwst/blob/05ab872e7c833e08b9b3daa3f158f605bc71a3d6/jwst/pipeline/calwebb_spec2.py#L91-L112

jdavies-st:
Possibly when opening from an AWS S3 bucket?  That seems to be the failure mode right now in the unit tests.  @eslavich might have some insight as he implemented the S3 functionality.

eslavich:
Right, we support opening a datamodel from an S3 URI (s3://some-bucket/some/key.fits), but that URI can't be read with the open function.  I wouldn't recommend reading the S3 object just to inspect the first 5 bytes, since the underlying code actually downloads the entire file into memory.

Can we just check the file extension in all cases?  Do we need to support users reading files with missing/unusual extensions?

jwhite3:
Sounds reasonable to me @eslavich . Thoughts @hbushouse @jdavies-st @philhodge ?

jwhite3:
@hbushouse I can't speak to the other pipelines without some digging first, but at least in that example, the error would be captured and printed out, and then at the conclusion of the pipeline a `RunTimeError` would be raised. So the "real" error would not get lost in that sense.

jwhite3:
I've refactored the `check` function to not attempt to open the file if the input is a string. Instead, It'll now try to determine the file type based on the extension. 

I did make an assumption that input `json` files should have the file type `asn` is this true?

Also, raises a `ValueError` with unexpected file extensions or no extension

jwhite3:
A Note: The assumption that ".json" -> "asn" is currently safe as `json` reference files are not supported as DataModels

jwhite3:
@stscieisenhamer Think this is ok to merge considering what we found yesterday?

Issue #4611, opened 2020-02-24
Jump Step Default for max_cores should be 'one' instead of 'None'


mwregan2:
The value of None is not accepted in config files. It also doesn't make sense to have the default value of max_cores to be None. The correct value should be 'one'.

stscieisenhamer:
Is this issue still relevant?

hbushouse:
According to https://github.com/spacetelescope/jwst/blob/master/jwst/jump/jump_step.py#L21 "one" is not a currently valid option. So I guess this request would entail not just resetting the default value of the param, but completely replacing the option "none" with the option "one". Is that the intention @mwregan2 ?

PR #4612, opened 2020-02-24
Change default of max_cores to be one in jump step. Fix for issue #4611

mwregan2:
This is a small change to make the default value more logical and allow the default value to be stored in the config file.


jdavies-st:
We are moving away from storing default values in config files.  We want to make sure all defaults are stored in the class itself and only there, as that way there is one location for step default params.

If defaults aren't in the config, then the step defaults from the class are used, which is what we want.

And in fact most of these defaults in .cfg files will be removed in an upcoming PR.  We've already done that in most of the pipeline .cfgs.  It is a work in progress.  Eventually, we will ask people to use the CRDS pars-jumpstep files and have DMS doing the same.

mwregan2:
The PR doesn't change the cfg file. It just addresses changing the default to be 'one'.

hbushouse:
Need to remove the "all_xxx_radon.txt" files from the commit.

mwregan2:
There were three files that needed to be deleted. I though I'd gotten rid of them a lot time ago.

jwst/jump/jump.py
```diff
@@ -105,17 +104,17 @@ def detect_jumps (input_model, gain_model, readnoise_model,
...
     slices.insert(numslices - 1, (data[:, :, (numslices - 1) * yincrement:nrows, :],
                                  gdq[:, :, (numslices - 1) * yincrement:nrows, :],
                                  readnoise_2d[(numslices - 1) * yincrement:nrows, :],
-                                 rejection_threshold, frames_per_group, flag_4_neighbors,
+                                 rejection_threshold, num_groups, flag_4_neighbors,
                                  max_jump_to_flag_neighbors, min_jump_to_flag_neighbors))
     if numslices == 1:
         gdq, row_below_dq, row_above_dq = twopt.find_crs(data, gdq, readnoise_2d, rejection_threshold,
-                                                                        frames_per_group, flag_4_neighbors,
+                                                                        num_groups, flag_4_neighbors,
```
hbushouse: Previously "nframes" had been passed into `find_crs`, which is in turn used within that function to properly estimate the read noise per group. Now you're sending "ngroups" for that argument, so are you still getting the correct read noise estimate? The read noise per group (or per two adjacent groups) should depend on "nframes", not "ngroups", shouldn't it?

mwregan2: Well, I lost an update somewhere in the creation of this branch. You are right this should be nframes. That's the but that I introduced and fixed in the main pull request. Somehow I lost jump.py in my repo and somehow I coped in a version that had the ngroups bug.
  My git skills clearly need work. 
I'll fix it now

mwregan2: I reapplied the fix to match the version of jump.py in master and then put in the small change to the default value. I also updated the test case to correctly fail when the wrong value is passed.

jwst/jump/tests/test_detect_jumps.py
```diff
@@ -12,7 +12,7 @@ def test_nocrs_noflux(setup_inputs):
     all pixel values are zero. So slope should be zero
     """
     model1, gdq, rnModel, pixdq, err, gain = setup_inputs(ngroups=5)
-    out_model = detect_jumps(model1, gain, rnModel, 4.0,  1, 200, 4, True)
+    out_model = detect_jumps(model1, gain, rnModel, 4.0,  'one', 200, 4, True)
```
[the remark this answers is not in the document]
mwregan2: If max_cores is set to 'one', it will set num_slices to be 1. Else, it will look at the value of max_cores for a match and if it doesn't match set num_slices to 1. So, it's a little faster when max_cores is set to 'one' and logically cleaner than just letting it cascade through the ifs and end up with num_slices = 1.

jdavies-st:
I'm not sure this PR is necessary or desirable.

The case currently where `max_cores = None` actually does not guarantee that only one core is being used.  It just turns off all calls to python's `multiprocessing` module and runs the step as it was originally written before multiprocessing was added.  No multiprocessing pools are made.  It's all just skipped.

What is actually happening under the hood of `numpy` with numbers of cores is not controlled by the pipeline (or even python).  It is likely that more than one is being used depending on machine architecture, installation of `numpy` being used and user preferences (i.e. of `mkl` is used or `open_blas`, etc).  So I think `max_cores = "one"` is actually misleading.  This was the original reason by we used `None` instead of `"one"`.  See comments in #3440.

If `max_cores = None` is not descriptive enough for multiprocessing being turned off, then a different parameter name might be more appropriate?  Ideas?

mwregan2:
James,
  As you know, the jump step uses multi_processing and not multiple threads. Max_cores does refer to the maximum number of cores that will have processes running on them. Numpy runs multi-threaded under a single process.  Monitoring the execution using top or Activity Monitor shows the difference between the two.
  It could be that the problem is the overloading of max_cores. It is both a flag to turn on or off the multi-processing and a scaling of how much of your computer you want to dedicate to the multi-processing. 
  Personally, I think the overloading is okay but I did it. So, of course, I would think that.
  If you would prefer, I can switch to two parameters. A yes/no multi-processing flag and a max_cores parameter. If multi-processing is set to no, then the max_cores parameter would be ignored. In that case, I would propose that the default value of max_cores be 'quarter' and not allowing 'one' as an option.

jwst/jump/jump_step.py
```diff
@@ -16,7 +16,7 @@ class JumpStep(Step):
 
     spec = """
         rejection_threshold = float(default=4.0,min=0) # CR sigma rejection threshold
-        maximum_cores = option('quarter', 'half', 'all', default=None) # max number of processes to create
+        maximum_cores = option('one', 'quarter', 'half', 'all', default='one') # max number of processes to create
```
hbushouse, line 19: I notice that in the recently merged multi-processing version of the ramp_fit step this param is defined in the spec as:
```
maximum_cores = string(default='none')
```
Down within the ramp_fit step code it does allow for additional values like "quarter", "half", and "all", as is the case here in the jump step. So I'm guessing one or the other of the two steps should be updated to make the definition of `maximum_cores` consistent between the two. Because several options are actually allowed in both steps, I'm guessing the definition in the ramp_fit step should be the one to get updated to match the option list used here.

mwregan2:
Howard,
  You are correct. This is kind of confusing because there is an old PR that changed the default to ‘one’ for jump. James didn’t like it and preferred the default to be ‘none’. That’s why I made the default none for the ramp_fitting step. Even so, it should have listed all of the options. We can just delete the jump_step PR to update the default value.
  I can do a pull request to fix this and update the ramp_fit step documentation to reflect the new parameter  too. Is that okay with you?
Mike

hbushouse:
@mwregan2 If we just close this PR, that will leave the default for `maximum_cores` in this step set to python `None`, as opposed to the string value "none" that's used as the default now in `ramp_fitting`. So if you want to close this PR we should submit a new one that makes both steps consistent, by using the "option" mechanism to list all available options and set the default to string "none". If you want to do that, and make sure the docs for both steps are also updated, that would be great.

mwregan2:
Howard,
  Sounds good. Let’s close the old PR and get everything consistent.
Thanks
Mike

hbushouse:
Closing this PR in favor of a new one that will make the `maximum_cores` argument definition consistent for both `jump` and `ramp_fitting` steps.

PR #4613, opened 2020-02-24
Add ``source_ra`` and ``source_dec`` to Slit MOS attributes

nden:
Adds `Slit.source_ra` and `Slit.source_dec` for MOS data with values from the MSA metadata file.
These are saved to the headers of FITS SCI extensions as keywords `SRCRA` and `SRCDEC`.

### DIFF
--- a/jwst/datamodels/schemas/slitdata.schema.yaml
+++ b/jwst/datamodels/schemas/slitdata.schema.yaml
@@ -95,7 +95,7 @@ allOf:
       title: Center of shutter in MSA coordinates (Y dir)
       type: integer
     source_ra:
-      title: Right Ascention of the source
+      title: Right Ascension of the source
       type: number
       fits_keyword: SRCRA
       fits_hdu: SCI
--- a/jwst/datamodels/schemas/slitdata.schema.yaml
+++ b/jwst/datamodels/schemas/slitdata.schema.yaml
@@ -94,6 +94,16 @@ allOf:
     ycen:
       title: Center of shutter in MSA coordinates (Y dir)
       type: integer
+    source_ra:
+      title: Right Ascention of the source
+      type: number
+      fits_keyword: SRCRA
+      fits_hdu: SCI
+    source_dec:
+      title: Declination of the source
+      type: number
+      fits_keyword: SRCDEC
+      fits_hdu: SCI
     data:
       title: The science data
       fits_hdu: SCI
--- a/jwst/transforms/models.py
+++ b/jwst/transforms/models.py
@@ -673,7 +673,8 @@ class Gwa2Slit(Model):
         A slit is a namedtupe of type `~jwst.transforms.models.Slit`
         Slit("name", "shutter_id", "dither_position", "xcen", "ycen", "ymin", "ymax",
         "quadrant", "source_id", "shutter_state", "source_name",
-        "source_alias", "stellarity", "source_xpos", "source_ypos"])
+        "source_alias", "stellarity", "source_xpos", "source_ypos",
+        "source_ra", "source_dec"])
     models : list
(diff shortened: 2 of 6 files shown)


jwst/datamodels/schemas/slitdata.schema.yaml
```diff
@@ -94,6 +94,12 @@ allOf:
     ycen:
       title: Center of shutter in MSA coordinates (Y dir)
       type: integer
+    source_ra:
+      title: Right Ascention of the source
+      type: number
+    source_dec:
+      title: Declination of the source
+      type: number
```
nden: If these are to be saved to the FITS headers we need FITS keywords. Any suggestions?

philhodge: The target coordinates in the primary header have keywords TARG_RA, TARG_DEC.  We could use RA_TARG, DEC_TARG for the keywords in extension headers.  Or perhaps RA_SLIT, DEC_SLIT?

jdavies-st: Why not just use the existing FITS keywords TARG_RA and TARG_DEC, but put them in the correct SCI extension?  This would cause the least confusion I think.

jwst/transforms/models.py
```diff
@@ -36,12 +36,12 @@
 Slit = namedtuple('Slit', ["name", "shutter_id", "dither_position", "xcen", "ycen",
                            "ymin", "ymax", "quadrant", "source_id", "shutter_state",
                            "source_name", "source_alias", "stellarity",
-                           "source_xpos", "source_ypos", ])
+                           "source_xpos", "source_ypos", "source_ra", "source_dec"])
 """ Nirspec Slit structure definition"""
 
 
 Slit.__new__.__defaults__ = ("", 0, 0, 0.0, 0.0, 0.0, 0.0, 0, 0, "", "", "", "",
```
philhodge: It looks to me as if there should only be three empty strings at the end of this line, i.e. for shutter_state, source_name, and source_alias.  The definition of `Slit` has 17 elements, but this specification of default values has 18 elements.
There are some doc strings in this file that could be updated.

jwst/datamodels/schemas/slitdata.schema.yaml
hbushouse: Personally I don't like repurposing existing keyword names for other uses and especially don't want to confuse users by having them appear in both the primary and SCI headers, with different values. A lot of the existing source-related keywords that are populated in the SCI header for multi-slit data use the form "SRCxxxx" (e.g. SRCNAME, SRCTYPE, SRCXPOS, ...) so I suggest we continue that scheme and use "SRCRA" and "SRCDEC".

jdavies-st: Agree.

jwst/transforms/models.py
nden: fixed

nden: Which docstrings did you have in mind?

philhodge: In transforms/models.py, both `Gwa2Slit` and `Slit2Msa` have a `slits` parameter, and the "Parameters" section of their doc strings gives the contents of those tuples.

nden: Thanks, I didn't notice them. All comments have been addressed now.

jwst/datamodels/schemas/slitdata.schema.yaml
```diff
@@ -94,6 +94,16 @@ allOf:
     ycen:
       title: Center of shutter in MSA coordinates (Y dir)
       type: integer
+    source_ra:
+      title: Right Ascention of the source
```
philhodge: typo:  Ascension

PR #4614, opened 2020-02-25
Turn on tweakreg in MIRI imaging regtest

jdavies-st:
This allows the MIRI imaging test data with only 5 stars to be run through `tweakreg`.  Also:

- Set `rejection_threshold=200` in `JumpStep` for this dataset.  It was detecting lots of noise.
- break up the pipeline-running fixtures in the MIRI Imaging regtests to be a single fixture for each pipeline stage, so if I only want to run the ramp test, I don't have to run the stage 2 and 3 pipelines.  Makes everything more modular.
- Update the the `diff_astropy_tables` assertion helper to be more assertion-like.
- Clean up parametrization of `test_nirspec_fs_spec2`

### DIFF
--- a/jwst/regtest/test_miri_image.py
+++ b/jwst/regtest/test_miri_image.py
@@ -58,11 +58,10 @@ def run_pipelines(jail, rtdata_module):
         # Set some unique param values needed for these data
         "--steps.tweakreg.snr_threshold=200",
         "--steps.tweakreg.use2dhist=False",
+        "--steps.tweakreg.minobj=4",
         "--steps.source_catalog.snr_threshold=20",
         ]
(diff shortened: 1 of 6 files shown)


jwst/regtest/test_miri_image.py
```diff
@@ -7,7 +7,7 @@
 from jwst import datamodels
 
 @pytest.fixture(scope="module")
-def run_pipelines(jail, rtdata_module):
+def run_detector1(rtdata_module):
     """Run stage 1-3 pipelines on MIRI imaging data."""
```
hbushouse: Should update this comment, now that only stage 1 is being run in this module.

PR #4618, opened 2020-02-26
JP-1324 Change data type of OUTPUT and ODD_EVEN columns to uint8

philhodge:
In the schema for the data model `IRS2Model`, the DQ bintable extension has three columns.  Two of these columns had data type int16, but the reference files that were recently installed in CRDS have data type uint8.  The schema has been changed to agree with the reference files.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -124,6 +124,8 @@ refpix
 
 - Fixed bugs in PR #4575; added unit tests [#4596]
 
+- Changed the data type of columns OUTPUT and ODD_EVEN in DQ [#4618]
+
 set_telescope_pointing
 ----------------------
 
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -124,6 +124,8 @@ refpix
 
 - Fixed bugs in PR #4575; added unit tests [#4596]
 
+- Changed the data type of columns OUTPUT and ODD_EVEN in DQ [#4618]
```
jdavies-st: This is in the `IRS2Model` reference file DQ?  Might be good to indicate that here.

philhodge: Yes, good suggestion.

PR #4620, opened 2020-02-26
Add requirements-sdp.txt pinning dependencies

jdavies-st:
This will pin dependencies for our nightly regression tests.  We should keep these pinned until integration period for 7.4.2 is completely done.

Also pin `tweakwcs==0.5.3` in `setup.py` as the new version 0.6.0 requires an unrelased version of `gwcs` for `jwst` data.


nden:
Why are the versions currently limited by `7.4.2` ? We will soon have new releases of packages that we want to use for the next delivery. It'd be good to start testing with them.

hbushouse:
We're hoping to deliver a CAL release for B7.4.2 in the next 48 hours, so we don't want things changing (other than the CAL code) during that time. Once we make the delivery, we can open it up again.

What new releases were you hoping to include in the B7.4.2 patch?

nden:
I was not aware there's a patch in the works so I have nothing to include in 7.4.2.

jdavies-st:
@nden, yeah if you want to get anything else in this patch (including dev dependencies), we'll have to change them in lockstep in `setup.py` and `requirements-sdp.txt` as we did for the last release. 

This is just to make this patch period a little stable.  We can wipe `requirements-sdp.txt` in a week or two if everything looks good for SDP.  `tweakwcs` release yesterday already broke our regression tests, so this is to just make any more changes we need deliberate.

Issue #2390, opened 2018-08-20
log amount, --verbose default, and debug forced by default


stscieisenhamer:
# Issue
Prompted by #2383, where valued debug information is being commented out, a re-evaluation of the `Step` logging level may be in order.  Specific issues include:
- `--verbose`` essentially seems to be `DEBUG`-level logging
- `Step.py:308` forces `DEBUG` by default.
- Conflicting use of `--debug` with common practice: Usually this invokes the `DEBUG` level of output. However, for `Step`, this causes `pdb` to be started when an exception occurs.

Possible solutions:
- Update coding standards to define what is default output, `verbose` output and `debug` output.
- Add/modify the `Step` options to allow the delineation.
- Remove the default `DEBUG` in the code. 
- Update regression testing to, by default, use `verbose`, but allow to use `debug`.

pllim:
I would also like an easier to configure log verbosity. See spacetelescope/da5-notebooks#23 . cc @jdavies-st

jwhite3:
In PR #4621 , the `setLevel` call in the definition of the `Step` class is gone, so the logging level is no longer forced to `DEBUG`

melanieclarke:
Since the logging level is no longer forced to DEBUG, remaining issues are tracked by the general refactor requested in #4658.  I will close this ticket in favor of tracking logging issues there.

PR #4621, opened 2020-02-26
JP-1090: Remove setLevel calls

### LINKED ISSUE #4182: Remove all explicity logging.setLevel calls
Issue [JP-1090](https://jira.stsci.edu/browse/JP-1090) was created by Jonathan Eisenhamer:

The whole of JWST code is littered with logging.setLevel calls, short-circuiting any usefullness of the logging facility.

Remove them. Just remove them...please...gone...go...squashed...

jwhite3:
This PR removes most `setLevel` calls from the project files with the exceptions in `stpipe.log.py` and `stpipe.tests.steps.local.py`

Resolves #4182 / [JP-1090](https://jira.stsci.edu/browse/JP-1090)

### DIFF
--- a/jwst/ami/ami_analyze.py
+++ b/jwst/ami/ami_analyze.py
@@ -12,7 +12,6 @@ from . import webb_psf
 from . import leastsqnrm
 
 log = logging.getLogger(__name__)
-log.setLevel(logging.DEBUG)
 
 
 def apply_LG(input_model, filter_model, oversample, rotation):
--- a/jwst/ami/ami_average.py
+++ b/jwst/ami/ami_average.py
@@ -5,7 +5,6 @@ import logging
 from .. import datamodels
 
 log = logging.getLogger(__name__)
-log.setLevel(logging.DEBUG)
 
 
 def average_LG(lg_products):
--- a/jwst/ami/ami_normalize.py
+++ b/jwst/ami/ami_normalize.py
@@ -6,7 +6,6 @@
 import logging
 
 log = logging.getLogger(__name__)
-log.setLevel(logging.DEBUG)
 
 
 def normalize_LG(target_model, reference_model):
--- a/jwst/ami/analyticnrm2.py
+++ b/jwst/ami/analyticnrm2.py
@@ -11,7 +11,6 @@ import scipy.special
 from . import leastsqnrm
 
 log = logging.getLogger(__name__)
-log.setLevel(logging.DEBUG)
 
 def Jinc(x, y):
     """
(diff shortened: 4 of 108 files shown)


hbushouse:
Does this still allow DEBUG-level messages to be emitted from individual steps when calling `strun` with the `-v` or `--debug` arguments?

jwhite3:
@hbushouse I believe so. From what I understand, those `setLevel` calls were effectively overriding the logging utility in `jwst.stpipe.log.py`

I could be wrong though, so I tagged @stscieisenhamer for a review since he file the issue this is in reference to

jdavies-st:
Yup, logging still works fine with `strun` and running from within python.  Behavior is not changed.  If we do want behavior changed, we will have to fix the stpipe logging config.

So in the future we will have to fix the logging config if we want to be able to run within python and use `setLevel` interactively in the shell or a notebook to control the amount of logging.

jwhite3:
@jdavies-st couldn't you still do that by hijacking the logger with `log = getLogger()` and then `log.setLevel(...)`? I supposed you'd need to know the name though

jdavies-st:
Doing

```
import logging
log.setLevel("WARNING")
```

doesn't work.  But something similar should.  Yeah, not sure what the name should or would be.  @stscieisenhamer ?

jdavies-st:
Ah, but what does work now is:

```
import logging
log = logging.getLogger("stpipe")
log.setLevel("ERROR")
```

:tada:

jdavies-st:
Well, I should clarify.  It now works for any log messages emitted by `stpipe` and it's subclasses.  But for log messages emitted by `crds` or `datamodels`, it does nothing.

So in addition to this PR, we should think about how we want to configure a top-level logger for the package, if needed.  Or have our datamodels use the `stpipe` logger.  Or something else?

I think this PR is fine as-is though.

jwhite3:
Yep for those, you'd need to change the settings individually, but I _think_ that other stuff (like steps and whatnot) would get that level change right?

jdavies-st:
So there is a top-level `jwst` logger.  And also a `CRDS` one.  So if one sets those levels to ERROR or CRITICAL, one can get quiet operation now.

```
logging.getLogger("CRDS").setLevel("CRITICAL")
logging.getLogger("tweakwcs").setLevel("ERROR")
logging.getLogger("jwst").setLevel("ERROR")
```

:tada:

stscieisenhamer:
So, this PR is good to go, as far as what the issue was discussing. There is more to be said but realized this is not the place. However...

@jwhite3: Suggested reading: The [documentation about logging](https://jwst-pipeline.readthedocs.io/en/latest/jwst/introduction.html#logging-configuration). Note: This is supposedly the "official" way that logging was to be handled by STPIPE; all this normal, standard log stuff was not supposed to be used. Then review issues #2159, #2390, JP-564, JP-579.

The ultimate goal is to find a path back towards more standard log handling. Discussions will be had.

PR #4625, opened 2020-02-27
Skip the NIRSpec IFU single exposure calspec3 regtest

hbushouse:
Update the regtest `test_nirspec_ifu` module to skip the `calwebb_spec3` test that uses a single exposure as input. It takes a long time to run and the results (only applies the `cube_build` and `extract_1d` steps) are completely redundant with the results of those steps applied in the `calwebb_spec2` test that uses the same exposure.

Also entered some missing updates to the change log.

### DIFF
--- a/jwst/regtest/test_nirspec_ifu.py
+++ b/jwst/regtest/test_nirspec_ifu.py
@@ -47,6 +47,7 @@ def run_spec2(jail, rtdata_module):
     return rtdata
 
 
+@pytest.mark.skip(reason="single exposure processing is redundant with calspec2")
 @pytest.fixture(scope='module')
 def run_spec3(jail, run_spec2):
     """Run the Spec3Pipeline on the results from the Spec2Pipeline run"""
--- a/jwst/regtest/test_nirspec_ifu.py
+++ b/jwst/regtest/test_nirspec_ifu.py
@@ -47,7 +47,6 @@ def run_spec2(jail, rtdata_module):
     return rtdata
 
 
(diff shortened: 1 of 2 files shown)


jwst/regtest/test_nirspec_ifu.py
```diff
@@ -47,6 +47,7 @@ def run_spec2(jail, rtdata_module):
     return rtdata
 
 
+@pytest.mark.skip(reason="single exposure processing is redundant with calspec2")
```
jdavies-st: This is the fixture. You actually want to `pytest.mark.skip` the `test_spec3` test below that uses this `run_spec3` fixture.

jdavies-st:
Also, you may want to rebase this branch against `upstream/master` and force push.  Looks like you're pulling in other changes from some other branch.

jwst/regtest/test_nirspec_ifu.py
hbushouse: Thanks. Fixing that.

jwst/regtest/test_nirspec_ifu.py
```diff
@@ -104,6 +104,7 @@ def run_spec3_multi(jail, rtdata_module):
...
```
jdavies-st: Still on the wrong test.  Next test down.  `test_spec3`

hbushouse: D'oh!! My brain is fried ...

Issue #4588, opened 2020-02-17
Enable Fixed slit NRS_AUTOWAVE calspec2 processing


stscijgbot:
Issue [JP-1306](https://jira.stsci.edu/browse/JP-1306) was created by Robert Jedrzejewski:

Break out the fixed-slit autowave part of lamp processing

stscijgbot:
Comment by Jonathan Eisenhamer: Completed as far as specs given by INS are concerned.

PR #4626, opened 2020-02-27
Added code to enable calspec2 processing of fixed slit autowave NRS data

stscirij:
Addresses #4588 and [jp-1306](https://jira.stsci.edu/browse/JP-1306)


jwst/assign_wcs/nirspec.py
```diff
@@ -1693,6 +1697,31 @@ def _create_ifupost_transform(ifupost_slice):
...
+    """
+    lamp_mode = input_model.meta.instrument.lamp_mode.lower()
+    if lamp_mode in ['fixedslit', 'brightobj']:
+        return slits_wcs(input_model, reference_files, slit_y_range)
+    elif lamp_mode == 'ifu':
+        return ifu(input_model, reference_files, slit_y_range)
+    elif lamp_mode == 'msaspec':
+        return slits_wcs(input_model, reference_files, slit_y_range)
+    else:
+        return not_implemented_mode
```
hbushouse: Where is `not_implemented_mode` defined/set?

hbushouse: The `not_implemented_mode` function below requires 2 arguments.

jwst/assign_wcs/nirspec.py
```diff
@@ -368,6 +369,9 @@ def get_open_slits(input_model, reference_files=None, slit_y_range=[-.55, .55]):
         slits = [Slit('S1600A1', 3, 0, 0, 0, slit_y_range[0], slit_y_range[1], 5, 4)]
     elif exp_type == "nrs_lamp":
         slits = get_open_fixed_slits(input_model, slit_y_range)
+    elif exp_type == "nrs_autowave":
+        if lamp_mode == 'fixedslit':
```
nden: Should this also check for `lamp_mode=="brightobj"`?

nden: BTW, what are the possible values of `lamp_mode`? It may be good to add them as `enum` to the core schema.

jwst/extract_2d/nirspec.py
```diff
@@ -37,7 +37,7 @@ def nrs_extract2d(input_model, slit_name=None, apply_wavecorr=False, reference_f
     exp_type = input_model.meta.exposure.type.upper()
 
     wavecorr_supported_modes = ['NRS_FIXEDSLIT', 'NRS_MSASPEC', 'NRS_BRIGHTOBJ',
-                                'NRS_AUTOFLAT']
+                                'NRS_AUTOFLAT', 'NRS_AUTOWAVE']
```
nden: Further down at line 69 the code checks if exp_type is `BRIGHTOBJ`. Should it check also for `NRS_AUTOWAVE` and `LAMP_MODE=BRIGHTOBJ`?

nden:
Other questions:

- Currently the list of steps in the pipeline is determined by the value of the keyword `FILTER`. If `FILTER==OPAQUE` the pipeline stops at the MSA. Are observations with `EXP_TYPE=NRS_AUTOWAVE` guaranteed to have `FILTER=OPAQUE`? 
- The rmap for the `wavecorr` reference files does not contain `NRS_AUTOWAVE`. Should it?
- While looking at the rmaps I noticed that while the code handles `NRS_AUTOFLAT` (as MOS data)  the rmaps for the reference types do not include it.

jwst/assign_wcs/nirspec.py
stscirij: Agreed it needs arguments. 3 of them actually

jwst/assign_wcs/nirspec.py
stscirij: I'm trying to do these modes one at a time so that progress can be made rather than bundling up all the changes necessary, so I'll get to the other modes later.  I'll add an enum entry to the datamodel.

jwst/extract_2d/nirspec.py
stscirij: See comment above

stscirij:
Answers to the above:
- I believe NRS_AUTOWAVE files should have FILTER equal to 'OPAQUE', but that's something for the INS folks to confirm
- I think the rmap for wavecorr should contain NRS_AUTOWAVE
- There are a bunch of rmaps that will need to change to handle this, along with the selection rules that will now need to include lamp_mode as a selector

jwst/assign_wcs/nirspec.py
philhodge: IIUC, `not_implemented_mode` is not intended to actually be evaluated at the point where it is returned by `autowave`.  `autowave` normally returns a pipeline, but in the case of `not_implemented_mode`, `autowave` returns a function (that returns None if it is eventually evaluated).  So no, you don't want to add an argument list to it here.

jwst/assign_wcs/nirspec.py
nden: I only mentioned `BRIGHTOBJ` because it's essentially identical to fixed slits and is already covered below in `autowave`. But I agree one at a time is best.

nden:
needs a change log entry

jwst/assign_wcs/nirspec.py
stscirij: Yes but when `not_implemented_mode` returned by `autowave`, the arguments have been consumed in the `autowave` call, so they need to be put back when the function call is passed on

nden:
Can you rebase and squash? I see changes which don't belong here.

hbushouse:
Definitely need a rebase, so that reviewers don't have to pick their way through 137 files!

hbushouse:
Can this PR be closed, since it's now been superseded by #4710 ?

stscirij:
Superseded by #4710

PR #4638, opened 2020-02-28
Add build-backend to pyproject.toml so that pep517.build works

jdavies-st:
This will allow building the sdist and wheel via

```
python -m pep517.build .
```
replacing

```
python setup.py sdist
python setup.py bdist_wheel
```


jdavies-st:
I also added the build command above to our Jenkins build matrix so it is tested.

PR #4639, opened 2020-02-28
Split NIRSpec IFU regtests

hbushouse:
Split the spec2 and spec3 tests contained in `test_nirspec_ifu.py` into separate test modules, so that they can run in parallel, thus saving execution time for the whole regtest suite (these are the 2 longest running tests of all).

### DIFF
--- a/jwst/regtest/test_nirspec_ifu_spec2.py
+++ b/jwst/regtest/test_nirspec_ifu_spec2.py
@@ -1,0 +1,56 @@
+"""Regression tests for NIRSpec IFU"""
+import pytest
+
+from . import regtestdata as rt
+
+# Define artifactory source and truth
+INPUT_PATH = 'nirspec/ifu'
+TRUTH_PATH = 'truth/test_nirspec_ifu'
+
+
+@pytest.fixture(scope='module')
+def run_spec2(jail, rtdata_module):
+    """Run the Spec2Pipeline on a single exposure"""
+    rtdata = rtdata_module
+
+    # Setup the inputs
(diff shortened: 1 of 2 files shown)


jdavies-st:
The way these tests are set up, running the spec3 pipeline depends on having run the spec2 pipeline, just like our miri and nircam imaging tests.  I don't think this will work.

The solution probably lies in either having them not depend on each other (i.e. separate fixtures for separate input) or just `@pytest.mark.skip` until we can sort out why cube build is taking forever.  Or turning off `outlier_detection`.

hbushouse:
I don't think that's the case. The spec3 test uses inputs that are independent from the spec2 run and the fixtures also don't depend on one another. I successfully ran each of them, one at a time, locally. They have independent inputs and outputs.

hbushouse:
Ah, you're correct that `run_spec3` has `run_spec2` as an argument, but that test is the one that's being skipped. The only spec3 test that's running is `run_spec3_multi`, which has no dependencies on `run_spec2`.

hbushouse:
Perhaps I should just remove the code for `run_spec3` and `test_spec3` (leaving only the `multi` versions), as opposed to having it set to skip. If anyone ever removed the skip, it would then have problems because of the dependency on `run_spec2`.

jdavies-st:
Thanks for the clarification.  And yeah, this looks good.

We should remove the data from bytesalad for the test that was removed as well.

hbushouse:
I'll figure out which dataset(s) need to be removed and will do that.

I've also been promised a brand new set of recent NIRSpec IFU simulations from INS sometime this week, so I may end up replacing everything that's currently used in these tests.

PR #4640, opened 2020-02-28
2390 use log config

jwhite3:
This PR adds the ability to configure the logger from the `Step` API (which means also the `Pipeline` API). 

Also includes some minor refactoring of `LogConfig`, removal of module-level log definitions for step modules (the log is created during `Step.__init__()`), and a slight update to some command-line args: 
- renamed "debug" to "debug_run" to remove confusion with common usage of "--debug" as a means to configure the logging level to `DEBUG`
- Changed the "--verbose" help text to explicitly state that the argument sets the logging level to `DEBUG`

### DIFF
--- a/jwst/assign_wcs/assign_wcs_step.py
+++ b/jwst/assign_wcs/assign_wcs_step.py
@@ -5,9 +5,6 @@ import logging
 from .assign_wcs import load_wcs
 from .util import MSAFileError
 
-
-log = logging.getLogger(__name__)
-
 __all__ = ["AssignWcsStep"]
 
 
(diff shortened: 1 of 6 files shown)


jwhite3:
Note: `LogConfig.apply()` is applied to the root logger `stpipe`. This is to avoid duplicate messages being emitted in the case of the user specifying an additional `StreamHandler` for the child log, `stpipe.MyStep` or something along those lines. 

However, this itself could result in some strange behavior. Suppose, for example, that a user wants to run a couple of different steps individually and each time adds a `FileHandler`: 

```python

step1 = MyStep1(log_handler='file: output_for_1.log')
...  # does stuff

step2 = MyStep2(log_handler='file: output_for_2.log')
```

The user could assume that the log messages for 1 will go to `output_for_1.log` and messages for 2 will go to `output_for_2.log`, but this is not the case. I think what'll actually happen is that messages for `MyStep1` will go to `output_for_1.log`, but then messages for `MyStep2` will go to _both_ log files. This is because the root logger will at that point have two different `FileHandlers` as opposed to just one that's replaced. 

This is definitely fixable, but we should think about at what level. We could limit how users can configure the logging from the `Step` API, or we could make `LogConfig` smarter

jdavies-st:
I'm not actually sure why there is something called `self.log` in the `stpipe` classes.  Logs are not objects of specific classes.  They are global.  Thoughts?

jwhite3:
@jdavies-st I can't speak for the original idea for why it was implemented in this way, but I think that it could still be a useful convention. 

I think the current implementation is useful for users as it means that they won't have to worry about accessing the logs directly, but can do so through the step or pipeline that they care about, although I can't speak to the original decision for implementing the logs in this way for `stpipe`.  

We could also be more clear by perhaps instantiating the loggers globally (as is sort of done in some `_step` and `_pipeline` modules) and just have the step or pipeline classes carry the configuration of those logs?

jwhite3:
Holding off on logging changes. Closing

PR #4641, opened 2020-02-28
Update README 0.15.0 install instructions

jdavies-st:


README.md
```diff
@@ -13,42 +13,46 @@ JWST Calibration Pipeline
 Installation
 ------------
 
-The ``jwst`` package can be installed into a virtualenv or conda environment via ``pip``.  We recommend creating a fresh environment with only python installed.  If using conda, make sure you have a recent version of Anaconda or Miniconda installed.  Via conda:
+The `jwst` package can be installed into a virtualenv or conda environment via `pip`.  We recommend creating a fresh environment with only python installed.  If using conda, make sure you have a recent version of Anaconda or Miniconda installed.  Via conda:
 
-    conda create -n jwst_env python=3.7
+    conda create -n jwst_env python
```
hbushouse: We don't need to pin the python version any more?

README.md
```diff
@@ -13,42 +13,46 @@ JWST Calibration Pipeline
...
 
     pip install git+https://github.com/spacetelescope/jwst
 
 As can a particular commit hash:
 
     pip install git+https://github.com/spacetelescope/jwst@3f03323c
 
+To include extra testing dependencies, the syntax is as follows:
+
+    pip install "jwst[test] @ git+https://github.com/spacetelescope/jwst"
```
hbushouse: Is the whitespace in the middle of "@ git" intentional?

README.md
jdavies-st: Right.  The package and its dependencies all support python 3.8.  That said, the DMS distribution still installs python 3.7.6, as all the packages are pinned.

README.md
jdavies-st: Yes.  It is needed.  And also the quotes.

PR #4642, opened 2020-02-28
JP-1327 mrs_imatch polynomial nan check

jemorrison:
When the background polynomial is Nans set poly = None so that cube_build does not try and subtract it and set skip step.

### DIFF
--- a/jwst/mrs_imatch/mrs_imatch_step.py
+++ b/jwst/mrs_imatch/mrs_imatch_step.py
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+
 """
 JWST pipeline step for image intensity matching for MIRI images.
 
@@ -316,6 +316,10 @@ def _match_models(models, channel, degree, center=None, center_cs='image'):
 
         image_data.append(cm.data)
         sigma_data.append(sigmas)
+
+    print('Mask',mask_data)
+    print('Image data',image_data)
+
 
     bkg_poly_coef, mat, _, _, effc, cs = match_lsq(
         images=image_data,
--- a/jwst/mrs_imatch/mrs_imatch_step.py
+++ b/jwst/mrs_imatch/mrs_imatch_step.py
@@ -317,8 +317,17 @@ def _match_models(models, channel, degree, center=None, center_cs='image'):
         image_data.append(cm.data)
         sigma_data.append(sigmas)
 
-    print('Mask',mask_data)
-    print('Image data',image_data)
+    # leaving in below commented out lines for
+    # Mihia to de-bug step  when coefficients are NAN
+    #mask_array = np.asarray(mask_data)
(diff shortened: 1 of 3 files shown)


mcara:
Hm... May I understand why would polynomial coefficients be `NaN`? Is it another case of [`GIGO`](https://en.wikipedia.org/wiki/Garbage_in,_garbage_out)? If so, what kind of data are causing this? Do data have `NaN`? If so, can they be masked out?

jemorrison:
I do not know why the polynomials have NaNs. The data was not what I call normal data - most of the data contained large negative values. I don't think this data is representative of what we would get in flight. This PR just checks if the polynomial is negative and skips the step if it is .
I can look a bit more at the data and see if there are NaNs in the data.

jwst/mrs_imatch/mrs_imatch_step.py
```diff
@@ -337,17 +340,22 @@ def _match_models(models, channel, degree, center=None, center_cs='image'):
...
 
-    # set 2D models' background meta info:
-    for im, poly in zip(models, bkg_poly_coef):
-        im.meta.background.subtracted = False
-        im.meta.background.polynomial_info.append(
-            {
+    if np.isnan(bkg_poly_coef).any():
+        bkg_poly_coef = None
+        for im in models:
+            im.meta.cal_step.mrs_imatch = 'SKIPPED'
```
hbushouse: I wonder if it would be good to also set `im.meta.background.subtracted = False` here (as is done below), just in case anything downstream uses that value for anything (just to make sure it's set to something, rather than None). Or could the presence of a value (either True or False) trick something into thinking that the background has at least been fitted and in this case it hasn't?

mcara: Maybe the code (downstream) should _be made/forced_ to check `im.meta.cal_step.mrs_imatch == 'COMPLETE'` before using `im.meta.background` values.

mcara:
@jemorrison @hbushouse My concern about this fix is that it may hide other bugs instead of allowing us to catch and fix them and allow processing. This fix simply turns off backgrounds nd matching without understanding why did it happen.

For example, if current issue is due to presence of `NaN`s in input data, then a better fix would be to create e a mask that flags "bad" data assuming input data have no mask. If it has mask, then maybe we should understand why `NaN`s are not masked. If everything is good with the data, then maybe we should revisit the algorithm to see why/under what circumstances it returns `NaN`

jemorrison:
I checked the data going into the the routine that determines the background polynomials and no Nans were found in the data. This PR just skips the step. @mcara wants to follow up on this data and determine why the polynomial determination fails. We can do that with a different issue.

mcara:
I would like to get the data (arrays/masks/weights) that go into the sky matching routine. I will investigate the issue at a later time.

PR #4644, opened 2020-02-28
Have s3 tests create data on-the-fly

jdavies-st:


jwst/conftest.py
```diff
@@ -35,8 +35,13 @@ def mk_tmp_dirs():
 
 
 @pytest.fixture(autouse=True)
-def monkey_patch_s3_client(monkeypatch):
-    monkeypatch.setattr(s3_utils, "_CLIENT", lib_helpers.MockS3Client())
+def monkey_patch_s3_client(monkeypatch, request):
+    # If tmpdir is used in the test, then it is providing the file.  Map to it.
+    if "tmpdir" in request.fixturenames:
+        path = request.getfixturevalue("tmpdir")
```
eslavich: Should we create a special-purpose temp directory and fixture for this?  Something like `s3_root_dir`.  That would make what's going on clearer in the tests -- as it is, the implicit mapping of tmpdir to S3 URIs is a little mysterious.  It also may not always be desirable to treat all files in tmpdir like they're available from S3.

jwst/lib/tests/helpers.py
```diff
@@ -12,6 +12,12 @@
 
 
 class MockS3Client:
+    def __init__(self, s3_test_data_path=None):
+        if s3_test_data_path:
+            self.s3_test_data_path = s3_test_data_path
+        else:
+            self.s3_test_data_path = S3_TEST_DATA_PATH
```
eslavich: If we want to discourage writing tests that rely on data in S3_TEST_DATA_PATH, maybe the client should just raise errors when the path in the initializer is None.  Something like "whoops, your test is reading from S3, but you didn't set that up".

PR #4653, opened 2020-03-02
Fix NIRISS WFSS FWPOS angle bugs

nden:
This PR fixes two of the NIRISS WFSS tests. The results were wrong before because the tests did not take into account the angle of the Filter Wheel Position recorded in the sci data file. 
It turns out that the angle that the trace is rotated about is the difference between the actual FWPOS and a reference one recorded in the reference file. So the previous presumption that the value of the angle is 0 was off by about ~33 deg.


docs/jwst/assign_wcs/main.rst
```diff
@@ -97,11 +98,20 @@ For each exposure, assign_wcs uses reference files and WCS header keywords
 to create the WCS object. What reference files are retrieved
 from CRDS is determined based on EXP_TYPE and other keywords in the science file header.
 
+The assign_wcs step can accept the single slope image that is the result of averaging
+over all integrations or a cube in the case of TSO observations.
```
hbushouse: change to "or a 3D cube of integrations in the case of TSO exposures"

docs/jwst/assign_wcs/main.rst
```diff
@@ -97,11 +98,20 @@ For each exposure, assign_wcs uses reference files and WCS header keywords
...
 
-The assign_wcs step can accept any type of DataModel as input. In particular, for
-multiple-integration datasets the step will accept either of these data products:
-the slope results for each integration in the exposure, or the single slope image
-that is the result of averaging over all integrations.
+WFSS mode
+---------
+
+The WCS forward transforms for WFSS observations (``NIS_WFSS``, ``NRC_WFSS``, ``NRC_TSGRISM``)
```
hbushouse: Because NRC_TSGRISM is technically not a WFSS observation, perhaps change to "... transforms for slitless grism exposures ..."

docs/jwst/assign_wcs/main.rst
```diff
@@ -97,11 +98,20 @@ For each exposure, assign_wcs uses reference files and WCS header keywords
...
+take as input the ``x, y`` coordinates on the dispersed image, the ``x0, y0`` coordinate of
+the center of the object in the direct imahe and ``spectral order``. They return the ``x0, y0`` coordinate of the center
```
hbushouse: typo "imahe" -> "image"

docs/jwst/assign_wcs/main.rst
```diff
@@ -97,11 +98,20 @@ For each exposure, assign_wcs uses reference files and WCS header keywords
...
+of the object in the direct image, ``wavelength`` and ``spectral order``.
+
+For NIRISS WFSS data the reference files contains a reference value for the filter wheel
```
hbushouse: "files contains" are inconsistent. One of them needs to lose the "s" on the end: either "file contains" or "files contain", depending on whether there's more than one reference file.

jwst/transforms/models.py
```diff
@@ -1546,38 +1546,42 @@ def evaluate(self, x, y, x0, y0, order):
...
         dxr = astmath.SubtractUfunc()
         wavelength = dxr | tab | lmodel
-        model = Mapping((2, 3, 0, 2, 4)) | Const1D(x0) & Const1D(y0) & wavelength & Const1D(order)
+        model = Mapping((2, 3, 0, 2, 4)) | Const1D(x00) & Const1D(y00) & wavelength & Const1D(order)
         return model(x, y, x0, y0, order)
 
 
 class NIRISSForwardColumnGrismDispersion(Model):
-    """This model calculates the dispersion extent of NIRISS pixels.
+    """This model calculates the wavelengths for horicontally dispersed NIRISS grism data.
```
hbushouse: typo: "horicontally" -> "horizontally"

nden:
Fixed all typos found by @hbushouse . Needs a review.

jwst/transforms/models.py
```diff
@@ -1450,21 +1447,23 @@ def evaluate(self, x, y, wavelength, order):
...
+        The list of orders which are available to the model
     xmodels : list[tuples]
         The list of tuple(models) for the polynomial model in x
     ymodels : list[tuples]
         The list of tuple(models) for the polynomial model in y
     lmodels : list
         The list of models for the polynomial model in l
-    orders : list
-        The list of orders which are available to the model
+    theta : float
```
jdavies-st: It looks like the ASDF Type to serialize this properly already has `theta`?  Was this just a case of Type being more up-to-date than the model itself?

`order` is removed because it is only used in the call to `evaluate`, whereas `theta` is needed on class init?

docs/jwst/assign_wcs/main.rst
```diff
@@ -97,11 +98,20 @@ For each exposure, assign_wcs uses reference files and WCS header keywords
...
+over all integrations or a 3D cube of integrations in the case of TSO exposures.
 
-The assign_wcs step can accept any type of DataModel as input. In particular, for
-multiple-integration datasets the step will accept either of these data products:
-the slope results for each integration in the exposure, or the single slope image
-that is the result of averaging over all integrations.
+WCS of slitless grism exposures
+-------------------------------
+
+The WCS forward transforms for WFSS slitless grism exposures (``NIS_WFSS``, ``NRC_WFSS``, ``NRC_TSGRISM``)
```
hbushouse: Remove "WFSS" here, so it's just "... for slitless grism exposures", because TSGRISM is not WFSS.

jwst/transforms/models.py
nden: This line is just adding `theta` to the docstrings of the model. it was used and defined but not in the docstrings. The real issue was with the unit tests only. They did not use `theta` and because the angle of rotation is defined as the difference between theta_ref and theta, when theta is 0 the trace in the unit tests was rotated by 33 deg.

docs/jwst/assign_wcs/main.rst
nden: done

PR #4660, opened 2020-03-05
add info and search methods to models

### LINKED ISSUE #3114: DataModel.info() method needs improvement
- [ ] The current `info()` method doesn't work for `MultiSlitModel` or `ModelContainer`
- [ ] It returns a stream for all other datamodels, which is not useful for the user introspecting a datamodel.
- [ ] Since `info()` doesn't have any parameters, it's not clear it needs to be a method rather than a property.
- [ ] The ordering of array sizes in `info()` is FITS order (x, y), not python order (y, x). This will just cause confusion. it should be changed to be consistent with python.

The original issue was #2268.

### LINKED ISSUE #4579: Consider adding Datamodel.info and Datamodel.search
Issue [JP-1302](https://jira.stsci.edu/browse/JP-1302) was created by Nadia Dencheva:

asdf now implements "asdf.info()" and "asdf.search()" which allow easy visualization and search of the ASDF tree. Implementing the corresponding methods on DataModel is going to make jwst data models more user and developer friendly. These two methods will be publicly available in the next asdf release - (2.6?).

nden:
Resolves #4579
Resolves #3114

Requires asdf v 2.6.0 (to be released soon).

### DIFF
--- a/docs/jwst/datamodels/models.rst
+++ b/docs/jwst/datamodels/models.rst
@@ -163,6 +163,56 @@ description is::
 where the second argument is the dictionary with the keywords
 mentioned.
 
+Looking at the contents of a model
+----------------------------------
+
+Use ``model.info()`` to look at the contents of a data model. It renders
+the underlying ASDF tree starting at the root or a specified ``node``.
+The number of displayed rows is controlled by the ``max_row`` argument::
+
+  im.info()
+  root.tree (AsdfObject)
+  ├─asdf_library (Software)
+  │ ├─author (str): Space Telescope Science Institute
+  │ ├─homepage (str): http://github.com/spacetelescope/asdf
+  │ ├─name (str): asdf
+  │ └─version (str): 2.5.2a1.dev12+g12aa460
+  ├─history (dict)
+  │ └─extensions (list) ...
+  ├─data (ndarray): shape=(2048, 2048), dtype=float32
+  ├─dq (ndarray): shape=(2048, 2048), dtype=uint32
+  ├─err (ndarray): shape=(2048, 2048), dtype=float32
+  ├─meta (dict)
+  │ ├─aperture (dict) ...
+  │ ├─bunit_data (str): DN/s
+  │ ├─bunit_err (str): DN/s
+  │ ├─cal_step (dict) ...
+  │ ├─calibration_software_revision (str): 3bfd782b
+  │ ├─calibration_software_version (str): 0.14.3a1.dev133+g3bfd782b.d20200216
+  │ ├─coordinates (dict) ...
+  │ └─28 not shown
+  ├─var_poisson (ndarray): shape=(2048, 2048), dtype=float32
+  ├─var_rnoise (ndarray): shape=(2048, 2048), dtype=float32
+  └─extra_fits (dict) ...
+  Some nodes not shown.
+
+
+Searching a model
+-----------------
+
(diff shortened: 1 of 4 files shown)


docs/jwst/datamodels/models.rst
```diff
@@ -163,6 +163,56 @@ description is::
 where the second argument is the dictionary with the keywords
 mentioned.
 
+Looking at the contents of a model
+----------------------------------
+
+Use ``model.info()`` to look at the contents of a data model. It renders
+the underlying ASDF tree starting at the root or a specified ``node``.
+The n umber of displayed rows is controlled by the ``max_row`` argument::
```
hbushouse: typo: extra whitespace in "The n umber"

jwst/datamodels/model_base.py
```diff
@@ -368,71 +368,15 @@ def callback(schema, path, combiner, ctx, recurse):
...
-                type_info = '(' + ','.join(col_info) + ')'
-            else:
-                type_info = numpy_dtype_to_asdf_datatype(instance.dtype)[0]
-            return type_info
+    def info(self, **kwargs):
+        return self._asdf.info(**kwargs)
 
-        meta_info = get_meta_info(self.meta)
-        field_info = get_field_info('', self._instance)
+    info.__doc__ = asdf.info.__doc__
```
jdavies-st: Because these are methods on `self`, the docstrings are wrong in the first arg.  It doesn't exist in this implementation.

nden: Sorry, which implementation? This PR should be merged after the release of asdf 2.6. The asdf version requirement will change to >=2.6. Is this what you mean?

jdavies-st: No, `asdf.info.__doc__` shows

```
    Print a rendering of an ASDF tree or sub-tree to stdout.

    Parameters
    ----------
    node_or_path : str, pathlib.Path, asdf.asdf.AsdfFile, or any                    ASDF tree node
        The tree or sub-tree to render.  Strings and Path objects
        will first be passed to asdf.open(...).

    max_rows : int, tuple, or None, optional
        Maximum number of lines to print.  Nodes that cannot be
        displayed will be elided with a message.
        If int, constrain total number of displayed lines.
        If tuple, constrain lines per node at the depth corresponding             to the tuple index.
        If None, display all lines.
```

But because in `datamodels`, `info` is a method that operates on `self`, `node_or_path` is not part of the calling sequence.  The docstring should not include `node_or_path`.

nden:
Failures are not related.

jwst/datamodels/model_base.py
```diff
@@ -400,71 +399,17 @@ def callback(schema, path, combiner, ctx, recurse):
...
-        buffer.append(format_string % ('attribute', 'size', 'type'))
-        buffer.append(70 * '-')
-        for field in field_info:
-            buffer.append(format_string % field)
-        return "\n".join(buffer) + "\n"
+    try:
+        info.__doc__ = AsdfFile().info.__doc__
+        search.__doc__ = AsdfFile().search.__doc__
+    except AttributeError:
+        pass
```
jdavies-st, line 412: I think you can use the class instead of the instance here.  I.e.:

```suggestion
    try:
        info.__doc__ = AsdfFile.info.__doc__
        search.__doc__ = AsdfFile.search.__doc__
    except AttributeError:
        pass
```

PR #4666, opened 2020-03-08
JP-623: Add multi-chip unit tests for tweakreg

mcara:
Resolves [JP-623](https://jira.stsci.edu/browse/JP-623)

**Description:**

This PR adds multi-chip testing to the `tweakreg` step of the pipeline. There are two _very_ similar tests: `test_multichip_jwst_alignment()` and `test_multichip_alignment_step()`. The first test is almost identical to what is available in [`tweakwcs/tests/test_multichip_jwst.py`](https://github.com/spacetelescope/tweakwcs/blob/afd09784155479b616ee9f81a581d8faef93148f/tweakwcs/tests/test_multichip_jwst.py#L145) except it is designed to run with different versions of `tweakwcs` and also `jwst` (once `TPCorr` is removed via https://github.com/spacetelescope/jwst/pull/4542) In my opinion there is an advantage to have this test in the `jwst` pipeline as it allows more frequent tests of `tweakreg` step.

The second test, `test_multichip_alignment_step()`, essentially repeats the first test but makes it run through the entire step thus, at least partially, testing the step itself. However, because `tweakreg` step does not store `'fit_info'` in `meta`, this test does not check the alignment as thoroughly as the first test.

These tests are a variation of the corresponding FITS test in the `tweakwcs` package adapted to `gwcs.WCS()`.

These tests are currently failing due to a bug in how correction was computed for JWST in the `tweakwcs` package version `<0.6.0`. The tests will start passing once the pipeline will switch to using `tweakwcs` version > `0.6.0`. I wanted to make this PR with failing tests because it shows that old (<`0.6.0`) `tweakwcs` was buggy and this explains the differences in the regression tests seen in https://github.com/spacetelescope/jwst/pull/4542: the failing test should be OKified.

Checklist
- [x] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [x] Label(s)


mcara:
Strange... tests are actually passing in a couple of cases and fail with minimally larger errors in other cases. I do not know why but I was getting significantly larger errors with old `tweakwcs` JWST code. Maybe then the test failure in #4542 is unrelated to `tweakwcs` changes.

mcara:
@hbushouse @jdavies-st This PR now can be merged too and provides a quick test for `tweakreg` and jwst-specific `tweakwcs`.

In light of recent updates of dependences `setup.py` (`tweakwcs>=0.6.3` and `gwcs>=0.13.0`), I think https://github.com/spacetelescope/jwst/pull/4666/files#diff-d1305d2e41555c241d84e88d0af6cc5cR27-R91 could be removed from this test.

Let me know how do you want to proceed.

hbushouse:
@jdavies-st are you OK with merging this?  Jenkins failure seems to be unrelated.

mcara:
Here are my answers to above comments:

> It uses WFC3 UVIS headers to create a FITS WCS with SIP coeffs.
> There's lots of custom code here to construct a gwcs.WCS object from the FITS WCS, and it is unrelated to what we do in jwst.assign_wcs for the imaging modes.

This is not a test for `assign_wcs` but for `tweakreg`. Designing this simulation that checks correctness of computations is very time consuming and it was performed for HST images first. While what goes inside a WCS is not relevant for this test, it was easier to simulate a HST WCS because of standardized transformations.

> If it starts failing, where do we start?

That is a valid question. But I feel it is almost impossible to design any *meaningful* test for `tweakreg` that could point right away to the problem. The code in the "step" is almost an "atomic" operation in the sense that all meaningful operation is performed in the `tweakwcs` package. 

> There's several test data files and test setup overhead in these files, including a custom subclass of ImageModel.

Latest commit removed the "custom subclass" - now test uses standard `ImageModel`. 

> If the only reason to include it here instead of in tweakwcs package is so that it runs more often, that's not a good enough reason. We run regressions nightly, and tweakwcs runs its tests whenever the code there changes. That's all that should be needed.

Several reasons for including this test. First is that `tweakwcs` is a dependency of `jwst` and not the other way around. Importing `jwst` (that depends on `tweakwcs`) in order to test `JWST` corrector _in `tweakwcs`_ is complicated. Those tests do not run often and definitely do not keep up with changes in schemas, models, etc. It is very useful to run this more often (second reason). 

> If we want to verify that multi-chip alignment is working as we expect, adding a second chip to one of our existing NIRCam regtests might be the more simple and easier-to-maintain way.

It might if you could simulate a perfect multi-chip alignment in NIRCam. It took a lot of effort to do this for HST _correctly_ and _precisely_ (not just ballpark like current `fit_quality_is_good()` method that makes sure that found transformations are not too large).

-----------------

Being a multi-chip simulation, this test tests single-chip by necessity and exercises most of the code in the `tweakreg` step and a lot of JWST-specific and general algorithm in the `tweakwcs` package.

nden (approved):

PR #4665, opened 2020-03-06
Remove FITS WCS transforms from WFSS mode

nden:
Resolves #2533 
Resolves #4622

PR #4667, opened 2020-03-10
Revert "JP-1090: Remove setLevel calls"

jwhite3:
Reverts spacetelescope/jwst#4621

### DIFF
--- a/jwst/ami/ami_analyze.py
+++ b/jwst/ami/ami_analyze.py
@@ -12,6 +12,7 @@ from . import webb_psf
 from . import leastsqnrm
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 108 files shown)


hbushouse:
So now the only question is how many, if any, of the other PR's that've been merged since our delivery of 0.15.0 do we want to include in our redelivery (probably designated as 0.15.1) to DMS? Many of them were related to regression tests or other internals that don't affect the Ops usage of the pipeline, but the following ones do:
#4642 
#4653 
#4665 

Some of these could actually be useful to Ops (e.g. #4642 fixes a bug that was caught in our previous round of DMS testing), but I worry about including too much that could have unforeseen consequences.

PR #4676, opened 2020-03-12
Jp 1203 Add command line and environmental options to not retrieve steppars references

### LINKED ISSUE #4389: Add option to not pull Step parameters from CRDS
Issue [JP-1203](https://jira.stsci.edu/browse/JP-1203) was created by Jonathan Eisenhamer:

h2. What

Add the ability to run {{Step}} without querying CRDS for its step parameters. 
h2. Why

This would allow running steps without having any external interaction. Particularly useful when testing or when running in an environment where external parameters are not desired.
h2. Specification

This would be a step-local step parameter, allowing targeted disabling. Being a step parameter, no other interfaces need to affected. Working name of this parameter is {{pars_from_crds}} with a default value of {{True}}

A nice-to-have would be creation of an environmental variable, {{PARS_FROM_CRDS}}, that would globally control CRDS access.

Note that this would NOT affect other reference file access. This would only affect step parameter reference retrieval.

stscieisenhamer:
Resolves #4389/[JP-1203](https://jira.stsci.edu/browse/JP-1203)

### DIFF
--- a/jwst/stpipe/pipeline.py
+++ b/jwst/stpipe/pipeline.py
@@ -37,6 +37,7 @@ from . import config_parser
 from . import Step
 from . import crds_client
 from . import log
+from .step import get_disable_crds_steppars
 from ..lib.class_property import ClassInstanceMethod
 
 
@@ -171,8 +172,16 @@ class Pipeline(Step):
             The parameters as retrieved from CRDS. If there is an issue, log as such
             and return an empty config obj.
         """
+        pars_model = cls.get_pars_model()
         refcfg = ConfigObj()
         refcfg['steps'] = Section(refcfg, refcfg.depth + 1, refcfg.main, name="steps")
+
+        # Check if retrieval should be attempted.
+        if get_disable_crds_steppars():
+            log.log.debug(f'{pars_model.meta.reftype.upper()}: CRDS parameter reference retrieval disabled.')
+            return refcfg
+
+
         log.log.debug('Retrieving all substep parameters from CRDS')
         #
         # Iterate over the steps in the pipeline
@@ -183,7 +192,6 @@ class Pipeline(Step):
             )
         #
         # Now merge any config parameters from the step cfg file
-        pars_model = cls.get_pars_model()
         log.log.debug(f'Retrieving pipeline {pars_model.meta.reftype.upper()} parameters from CRDS')
         exceptions = crds_client.get_exceptions_module()
         try:
--- a/jwst/stpipe/step.py
+++ b/jwst/stpipe/step.py
@@ -272,7 +272,6 @@ class Step():
             variables on the new Step instance.
         """
         self._pars_model = None
-        self._disable_crds_steppars = None
         self._reference_files_used = []
         self._input_filename = None
         self._input_dir = None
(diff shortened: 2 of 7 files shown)


hbushouse:
While this is a convenient work around to avoid the ridiculous amount of time spent loading input datamodels to do the retrievals, eventually (once step param files exist in CRDS) we'll need to fix it at the source, i.e. in the calls to CRDS best ref itself.

stscieisenhamer:
Agreed, this is not the fix for the efficiency issues, which is to be handled in [JP-1322](https://jira.stsci.edu/browse/JP-1322).

jwst/stpipe/step.py
```diff
@@ -1296,3 +1302,31 @@ def _get_suffix(suffix, step=None, default_suffix=None):
     if suffix is None and step is not None:
         suffix = step.name.lower()
     return suffix
+
+
+def get_disable_crds_steppars(default=None):
```
jwhite3:
```suggestion
def get_disable_crds_steppars(flag=None):
```

The name `default` for this arg was a bit misleading to me

jwhite3: Of course, this would need to be refactored in the function as well

jwst/stpipe/step.py
```diff
@@ -1296,3 +1302,31 @@ def _get_suffix(suffix, step=None, default_suffix=None):
...
+    truths =  ('true', 'True', 't', 'yes', 'y')
+    if default:
+        if isinstance(default, bool):
+            return default
+        elif isinstance(default, str):
+            return default in truths
+        raise ValueError(f'default must be string or boolean: {default}')
+
+    flag = os.environ.get('STPIPE_DISABLE_CRDS_STEPPARS', '')
+    return flag in truths
```
jwhite3:
```suggestion
    if isinstance(default, bool):
        return default

    if isinstance(default, str):
        return default in truths
    
    if default is None:
        flag = os.environ.get('STPIPE_DISABLE_CRDS_STEPPARS', '')
        return flag in truths
    
    raise ValueError(f'If supplied, default must be string or boolean: {default}')
```

Breaking up the conditions here makes it much easier to read

jwst/stpipe/step.py
```diff
@@ -707,9 +708,14 @@ def get_config_from_reference(cls, dataset, observatory=None):
         # Get the root logger, since the following operations
         # are happening in the surrounding architecture.
         logger = log.delegator.log
+        pars_model = cls.get_pars_model()
+
+        # Check if retrieval should be attempted.
+        if get_disable_crds_steppars():
```
jwhite3: Shouldn't the arg be supplied here so that users can modify from the API? Or was it intentional to only get the info from the environment variable in this method?

jwst/stpipe/pipeline.py
```diff
@@ -171,8 +172,16 @@ def get_config_from_reference(cls, dataset, observatory=None):
             The parameters as retrieved from CRDS. If there is an issue, log as such
             and return an empty config obj.
         """
+        pars_model = cls.get_pars_model()
         refcfg = ConfigObj()
         refcfg['steps'] = Section(refcfg, refcfg.depth + 1, refcfg.main, name="steps")
+
+        # Check if retrieval should be attempted.
+        if get_disable_crds_steppars():
```
jwhite3: Same question as in the implementation for `Step`

jwst/stpipe/step.py
stscieisenhamer: This would have required a bit of refractor upstream, which, for this functionality, was deemed not necessary. If it turns out to be an in-demand item, will consider it then.

PR #4678, opened 2020-03-12
Add .sw[op] to gitignore

larrybradley:
For vim users, like me. 😃


jdavies-st:
This is probably fine, but I think a better policy is for users to have a global gitignore at  `~/.gitignore` or similar.

More info

https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

PR #4680, opened 2020-03-13
Add datamodels for reference files containing AB to Vega mag offsets

larrybradley:
This PR adds new imaging-mode Vega magnitude offsets (from AB magnitudes) reference file data models: `FgsImgVegaMagOffsetModel`, `MirImgVegaMagOffsetModel`, `NrcImgVegaMagOffsetModel`, and `NisImgVegaMagOffsetModel`.

The JWST Photometry Working Group (JPWG) has requested both AB and Vega magnitudes in the source catalog.  These new reference files contain a table of the offsets to convert from AB mag to Vega mag for each filter.

There should be only one reference file of this type for each instrument.  Because the level-3 resampled/combined `i2d` image may represents the combination of several detectors, there does not need to be a separate reference file for each detector.  The selectors for each instrument are:
 * FGS:  None (it's a single number)
 * NIRCam, NIRISS:  `filter`, `pupil`
 * MIRI:  `filter` (I don't think `subarray` is needed here because different `subarrays` could be combined?)

I also made the Vega mag offset reference files for each instrument.  I do not yet have the Vega mag offset for FGS, so I set it to zero for now.

### DIFF
--- a/docs/jwst/references_general/abvega_offset_reffile.inc
+++ b/docs/jwst/references_general/abvega_offset_reffile.inc
@@ -1,0 +1,68 @@
+.. _abvega_offset_reffile:
+
+ABVEGA_OFFSET Reference File
+----------------------------
+
+:REFTYPE: ABVEGA_OFFSET
+:Data model: `~jwst.datamodels.ABVegaOffsetModel`
+
+The ABVEGA_OFFSET reference file contains data necessary for converting
+from AB to Vega magnitudes.
+
+.. include:: ../references_general/abvega_offset_selection.inc
+
+.. include:: ../includes/standard_keywords.inc
+
+ABVEGA_OFFSET Reference File Format
++++++++++++++++++++++++++++++++++++
+
+ABVEGA_OFFSET reference files are in ASDF format.  The ABVEGA_OFFSET
+reference file contains tabular data in a key called
+``abvega_offset``.  The content of the table varies for different
+instrument modes, as shown in the tables below.
+
++------------+---------------+-----------+------------+----------+
+| Instrument | Column name   | Data type | Dimensions | Units    |
++============+===============+===========+============+==========+
+| FGS        | detector      | string    | 7          | N/A      |
++            +---------------+-----------+------------+----------+
+|            | abvega_offset | float     | scalar     | unitless |
++------------+---------------+-----------+------------+----------+
+
++------------+---------------+-----------+------------+----------+
+| Instrument | Column name   | Data type | Dimensions | Units    |
++============+===============+===========+============+==========+
+| MIRI       | filter        | string    | 12         | N/A      |
++            +---------------+-----------+------------+----------+
+|            | abvega_offset | float     | scalar     | unitless |
++------------+---------------+-----------+------------+----------+
+
++------------+---------------+-----------+------------+----------+
+| Instrument | Column name   | Data type | Dimensions | Units    |
++============+===============+===========+============+==========+
+| NIRCam     | filter        | string    | 12         | N/A      |
++ or         +---------------+-----------+------------+----------+
+| NIRISS     | pupil         | string    | 15         | N/A      |
++            +---------------+-----------+------------+----------+
+|            | abvega_offset | float     | scalar     | unitless |
++------------+---------------+-----------+------------+----------+
+
+
+Row Selection
+^^^^^^^^^^^^^
+
+A row of data within the reference table is selected by the pipeline
+step based on the optical elements in use for the exposure. The
+selection attributes are always contained in the first few columns of
+the table.  The last column contains the data needed to convert from
+AB to Vega magnitudes.  The row selection criteria for each
+instrument/mode are:
+
+* FGS:
+   - Detector
+* MIRI:
+   - Filter
+* NIRCam:
+   - Filter and Pupil
+* NIRISS:
+   - Filter and Pupil
--- a/docs/jwst/references_general/abvega_offset_selection.inc
+++ b/docs/jwst/references_general/abvega_offset_selection.inc
@@ -1,0 +1,16 @@
+Reference Selection Keywords for ABVEGA_OFFSET
+++++++++++++++++++++++++++++++++++++++++++++++
+
+CRDS selects appropriate ABVEGA_OFFSET references based on the
+following keywords.  ABVEGA_OFFSET is not applicable for instruments
+not in the table.  All keywords used for file selection are
+*required*.
+
+========== ======================================
+Instrument Keywords
+========== ======================================
+FGS        INSTRUME, EXP_TYPE, DATE-OBS, TIME-OBS
+MIRI       INSTRUME, EXP_TYPE, DATE-OBS, TIME-OBS
+NIRCam     INSTRUME, EXP_TYPE, DATE-OBS, TIME-OBS
+NIRISS     INSTRUME, EXP_TYPE, DATE-OBS, TIME-OBS
+========== ======================================
--- a/jwst/datamodels/schemas/abvega_offset.schema.yaml
+++ b/jwst/datamodels/schemas/abvega_offset.schema.yaml
@@ -5,43 +5,23 @@ id: "http://stsci.edu/schemas/jwst_datamodel/abvegamag_offset.schema"
 title: AB to Vega magnitude offset reference file model
 allOf:
 - $ref: referencefile.schema
-- anyOf:
-  - type: object
-    properties:
-      abvega_offset_detector:
-        title: Offsets to convert AB to Vega magnitudes (FGS)
-        type: object
-        properties:
+- type: object
+  properties:
+    abvega_offset:
+      title: Offsets to convert AB to Vega magnitudes
+      type: object
+      properties:
+        abvega_offset:
+          title: AB_mag - Vega_mag
+          type: number
+        anyOf:
           detector:
             title: Detector name
             type: string
-          abvega_offset:
-            title: AB_mag - Vega_mag
-            type: number
-  - type: object
-    properties:
-      abvega_offset_filter_pupil:
-        title: Offsets to convert AB to Vega magnitudes (NIRCam and NIRISS)
-        type: object
-        properties:
           filter:
             title: Filter wheel element name
             type: string
           pupil:
             title: Pupil wheel element name
             type: string
-          abvega_offset:
-            title: AB_mag - Vega_mag
-            type: number
-  - type: object
-    properties:
-      abvega_offset_filter:
-        title: Offsets to convert AB to Vega magnitudes (MIRI)
-        type: object
-        properties:
-          filter:
-            title: Filter wheel element name
-            type: string
-          abvega_offset:
-            title: AB_mag - Vega_mag
-            type: number
+...
--- a/jwst/datamodels/schemas/abvega_offset.schema.yaml
+++ b/jwst/datamodels/schemas/abvega_offset.schema.yaml
@@ -1,0 +1,47 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.0.0"
+id: "http://stsci.edu/schemas/jwst_datamodel/abvegamag_offset.schema"
+title: AB to Vega magnitude offset reference file model
+allOf:
+- $ref: referencefile.schema
+- anyOf:
+  - type: object
+    properties:
+      abvega_offset_detector:
+        title: Offsets to convert AB to Vega magnitudes (FGS)
+        type: object
+        properties:
+          detector:
+            title: Detector name
+            type: string
+          abvega_offset:
+            title: AB_mag - Vega_mag
+            type: number
+  - type: object
+    properties:
+      abvega_offset_filter_pupil:
+        title: Offsets to convert AB to Vega magnitudes (NIRCam and NIRISS)
+        type: object
+        properties:
+          filter:
+            title: Filter wheel element name
+            type: string
+          pupil:
+            title: Pupil wheel element name
+            type: string
+          abvega_offset:
+            title: AB_mag - Vega_mag
+            type: number
+  - type: object
+    properties:
+      abvega_offset_filter:
+        title: Offsets to convert AB to Vega magnitudes (MIRI)
+        type: object
+        properties:
+          filter:
+            title: Filter wheel element name
+            type: string
+          abvega_offset:
+            title: AB_mag - Vega_mag
+            type: number
--- a/jwst/datamodels/tests/data/nircam_abvega_offset.asdf
+++ b/jwst/datamodels/tests/data/nircam_abvega_offset.asdf
@@ -1,0 +1,61 @@
+#ASDF 1.0.0
+#ASDF_STANDARD 1.4.0
+%YAML 1.1
+%TAG ! tag:stsci.edu:asdf/
+--- !core/asdf-1.1.0
+asdf_library: !core/software-1.0.0 {author: Space Telescope Science Institute, homepage: 'http://github.com/spacetelescope/asdf',
+  name: asdf, version: 2.5.2}
+history:
+  entries:
+  - !core/history_entry-1.0.0 {description: File data provided by Kevin Volk in February
+      2020. Errors are at the 3-6% level., time: ! '2020-04-17 17:40:03'}
+  extensions:
+  - !core/extension_metadata-1.0.0
+    extension_class: asdf.extension.BuiltinExtension
+    software: {name: asdf, version: 2.5.2}
+  - !core/extension_metadata-1.0.0
+    extension_class: astropy.io.misc.asdf.extension.AstropyExtension
+    software: {name: astropy, version: 4.0.1.post1}
+abvega_offset: !<tag:astropy.org:astropy/table/table-1.0.0>
+  colnames: [filter, pupil, abvega_offset]
+  columns:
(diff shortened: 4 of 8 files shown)


larrybradley:
They are identical, but I don't think they can be combined.  Some of the filter names are the same for NIRCam and NIRISS.

jdavies-st:
That's OK though, as they will be different reference files.  The schemas only need be different if the definition of `dtype` of the columns is different in the reference file.  So one schema should be able to handle both reffile types:

jwst_nircam_vegaoffset_0001.fits
jwst_niriss_vegaoffset_0001.fits

A quick check to make sure this is the case would be to open your current NIRCam reffile using `NisImgVegaMagOffsetModel`.  If no errors, then you're good to go.

larrybradley:
I was using the `apcorr` schemas as a reference.  For those, `nisimg_apcorr.schema.yaml` and `nrcimg_apcorr.schema.yaml` are also identical (except for the `id` and `title`).

larrybradley:
Same for `nisimg_photom.schema.yaml` and `nrcimg_photom.schema.yaml`.

CC: @hbushouse

hbushouse:
Regarding identical data models: while it's nice to prevent the proliferation of model types, I think that in this case (and apcorr and photom) it's worth it, because at least one or more instruments need to have a unique model (e.g. miri and fgs). I think it would be a bit confusing to have a `MiriXxxModel`, `FgsXxxModel`, and then a generic `XxxModel` for NIRCam and NIRISS. That requires folks to remember which instruments have an instrument-specific model and which ones don't.

jwst/datamodels/schemas/fgsimg_vegamag_offset.schema.yaml
```diff
@@ -0,0 +1,17 @@
...
+title: FGS Vega magnitude offset data model
+allOf:
+- $ref: referencefile.schema
+- $ref: keyword_exptype.schema
+- $ref: keyword_pexptype.schema
+- type: object
+  properties:
+    vegamag_offset_table:
+      title: Vega magnitude offsets from AB magnitudes table
+      fits_hdu: VEGAMAG_OFFSET
```
hbushouse: Arrgh, these are going to be FITS tables?  How gross. ASDF would be so much nicer.

jwst/datamodels/vegamag_offset.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    A FGS imaging data model for reference files containing Vega
+    magnitude offsets from AB magnitudes.
+
+    Parameters
+    ----------
+    vegamag_table : numpy table
+        Vega magnitude offsets tables.
+        A table-like object containing row selection criteria made up of
+        instrument mode parameters and aperture correction factors
```
hbushouse: Whoops - copy-n-paste error: these tables don't contain aperture correction factors. Need to update (in all schemas).

jwst/datamodels/schemas/fgsimg_vegamag_offset.schema.yaml
sosey: Is there a specific reason they have to be FITS tables? ASDF would make for a files that is easier to understand and maintain.

larrybradley: No, I was just following the `photom` and `apcorr` examples, which are both FITS.  I can make them ASDF, although the INS teams seem to prefer FITS tables.

hbushouse: ASDF is so much simpler and more appropriate for simple table-based data like this. We already have a number of reference files in ASDF format (e.g. some of the WCS-related ref files), so there is precedence. Since you're the one creating them, I'd suggest using ASDF, which would also then help others to start down the ASDF road. We suggested ASDF for the apcorr ref files, but there seemed to be an assumption that a new format would take time to learn and time was of the essence (at least back then) to get them created and delivered.

jdavies-st:
Another way to do this would be to have a single datamodel, but use `anyOf` with the tables, so tables of any of the 3 formats would validate.  Something like:

```
%YAML 1.1
---
id: "http://stsci.edu/schemas/jwst_datamodel/vegamag.schema"
title: Vega magnitude offset data model
allOf:
- $ref: referencefile.schema
- anyOf:
  - type: object
    properties:
      vegamag_offset:
        title: Vega magnitude offsets from AB magnitudes table
        fits_hdu: VEGAMAG_OFFSET
        datatype:
        - name: vegamag_offset
          datatype: float32
  - type: object
    properties:
      vegamag_offset_filter:
        title: Vega magnitude offsets from AB magnitudes table
        fits_hdu: VEGAMAG_OFFSET_FILTER
        datatype:
        - name: filter
          datatype: [ascii, 12]
        - name: vegamag_offset
          datatype: float32
  - type: object
    properties:
      vegamag_offset_filter_pupil:
        title: Vega magnitude offsets from AB magnitudes table
        fits_hdu: VEGAMAG_OFFSET_FILTER_PUPIL
        datatype:
        - name: filter
          datatype: [ascii, 12]
        - name: pupil
          datatype: [ascii, 15]
        - name: vegamag_offset
          datatype: float32
...
```
The reader then does not need to worry about the instrument or datamodel type, only the type of table.

nden:
Using FITS tables for these seems like a very heavy machinery. Is there a specific request or decision for the format? In the case of FGS we are storing one number in a FITS binary table? Or did I misunderstand?

jwst/datamodels/schemas/fgsimg_vegamag_offset.schema.yaml
nden: We also have the `filteroffset` reference file type which may serve as an example.

jwst/datamodels/vegamag_offset.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class MirImgVegaMagOffsetModel(ReferenceFileModel):
+    """
+    A MIRI imaging data model for reference files containing Vega
+    magnitude offsets from AB magnitudes.
+
+    Parameters
+    ----------
+    vegamag_table : numpy table
```
nden: What is the expected parameter type?

nden: The model classes should create a valid data model when a file name is passed. Could you add a test?

pllim:
Did someone say Vega? Which Vega spectrum did you use? How did you calculate the offsets? Anything I need to know as `synphot` maintainer?

larrybradley:
I got the values from Kevin Volk.

larrybradley:
I've updated this PR:

* Changed the reference files from FITS binary tables to ASDF.  The initial version of the 4 ASDF references files (one per instrument) are here:  https://gist.github.com/larrybradley/e3073233eef8b2fbddecbea53afee3bb

* Replaced 4 datamodels and schemas with a single datamodel and schema that works for all instruments.

* Added a test - but this required adding a small ASDF file.

Because the instruments have different selectors, there are three flavors of astropy tables (with different column names) in the reference files.  The schema uses `anyOf` for the column names.  Please let me know if there is a better way to do this.  I could find no other examples of a single schema for all instruments like this.  The schema is valid and roundtrips correctly reading/writing the reference files.


CC: @nden, @jdavies-st, @stscieisenhamer, @hbushouse

jwst/datamodels/schemas/fgsimg_vegamag_offset.schema.yaml
larrybradley: Updated to use ASDF.

jwst/datamodels/vegamag_offset.py
larrybradley: Fixed.

jwst/datamodels/vegamag_offset.py
larrybradley: Done.

larrybradley:
Unrelated question:  Why do the docstrings for the datamodel classes have a `Parameters` section?  Shouldn't these items be in an `Attributes` section?

jwst/datamodels/tests/data/nircam_abvega_offset.asdf
```diff
@@ -0,0 +1,61 @@
...
+meta:
+  author: STScI
+  description: Offset from AB to Vega magnitudes
+  instrument: {name: NIRCAM}
+  model_type: ABVegaOffsetModel
+  pedigree: GROUND
+  reftype: abvegaoffset
+  telescope: JWST
+  title: AB to Vega magnitude offset reference file
+  userafter: '2015-01-01T00:00:00'
```
jdavies-st, line 60: `useafter` here

jwst/datamodels/schemas/abvega_offset.schema.yaml
```diff
@@ -0,0 +1,26 @@
...
+        anyOf:
+          detector:
+            title: Detector name
+            type: string
+          filter:
+            title: Filter wheel element name
+            type: string
+          pupil:
+            title: Pupil wheel element name
+            type: string
```
jdavies-st, line 26: schema should have a final line with `...`

jwst/datamodels/tests/test_models.py
```diff
@@ -691,3 +692,10 @@ def test_ifuimage():
 def test_datamodel_raises_filenotfound():
     with pytest.raises(FileNotFoundError):
         DataModel(init='file_does_not_exist.fits')
+
+def test_abvega_offset_model():
+    model = ABVegaOffsetModel('nircam_abvega_offset.asdf')
+    assert isinstance(model, ABVegaOffsetModel)
+    assert hasattr(model, 'abvega_offset')
+    assert isinstance(model.abvega_offset, Table)
+    assert model.abvega_offset.colnames == ['filter', 'pupil', 'abvega_offset']
```
jdavies-st, line 701: Maybe a `model.validate()` call would be good in the test.  Or better, make those required fields in the schema?  @nden I recall there being some `.validate()` methods you have written that do something like this.

jwst/datamodels/schemas/abvega_offset.schema.yaml
larrybradley: None of the schemas in `jwst/datamodels/schemas` end with `...`.  I think you mean ASDF file must end with `...`?

jwst/datamodels/tests/data/nircam_abvega_offset.asdf
larrybradley: Removed.

jwst/datamodels/schemas/abvega_offset.schema.yaml
jdavies-st: Actually both do, as it's a YAML standard.  We don't actually check for it in our checker right now, so ignore that.  We clearly have some cleaning up to do in the rest of our schemas.  =)

larrybradley:
`abvega_offset` is essentially unitless.  There's a `u.ABmag` unit, but not one for Vega mag.  It's not clear how one would represent a difference between ABmag and Vega mag.  There is no generic `u.mag` unit.  There is also a `u.Magnitude()` object in astropy for converting logarithmic units, but its usage is completely unintuitive to me and I think unnecessary here.

larrybradley:
I did try using the `$ref: "tag:astropy.org:astropy/table/table-1.0.0"` tag in the schema, but I could not get it to work and I didn't find any other example schemas that use it.

jwst/datamodels/schemas/abvega_offset.schema.yaml
larrybradley: I'm happy to be a trailblazer and add the trailing `...`. 😄

larrybradley: I've added the trailing `...`.

jwst/datamodels/tests/test_models.py
nden: `ReferenceFileeModel` has its own `validate()` method which checks for the presence of the common required  reference file keywords. When possible I also add `validate` on specific models to make sure they are consistent and that fields I know should be there are really present, for example models in wcs_ref_models.py.

larrybradley: I've added a custom `validate()` method to check that the expected columns for a given instrument are in the table.

docs/jwst/references_general/abvega_offset_reffile.inc
```diff
@@ -0,0 +1,67 @@
...
+:REFTYPE: ABVEGA_OFFSET
+
+The ABVEGA_OFFSET reference file contains data necessary for converting
+from AB to Vega magnitudes.
+
+.. include:: ../references_general/abvega_offset_selection.inc
+
+.. include:: ../includes/standard_keywords.inc
+
+APVEGA_OFFSET Reference File Format
```
hbushouse, line 13: typo: A*P*VEGA should be A*B*VEGA

docs/jwst/references_general/abvega_offset_reffile.inc
```diff
@@ -0,0 +1,67 @@
...
+^^^^^^^^^^^^^
+
+A row of data within the reference table is selected by the pipeline
+step based on the optical elements in use for the exposure. The
+selection attributes are always contained in the first few columns of
+the table.  The last column contains the data needed to convert from
+AB to Vega magnitudes.  The row selection criteria for each
+instrument/mode are:
+
+* FGS Image:
```
hbushouse, line 60: Minor nit, but probably don't need all the explicit references to "Image" mode in each of these entries, because this ref file only applies to imaging mode to begin with (as opposed to the apcorr or photom ref files).

docs/jwst/references_general/abvega_offset_reffile.rst
```diff
@@ -0,0 +1,5 @@
+:orphan:
```
hbushouse, line 1: I think that this scheme of having an (orphan) .rst file include a .inc file is only necessary when the ref file type is used by multiple steps and hence has to be referenced from multiple step docs (as is the case for apcorr, being used by both `source_catalog` and `extract_1d`. But since this ref file type is only used by one step (`source_catalog`) I think you can use the normal doc file layout used by other ref files that correspond to a single step, such as linearity and superbias. In that scheme, you only have the 2 .inc files (abvega_offset_reffile.inc and abvega_offset_selection.inc), where the abvega_offset_reffile.inc file contains the reference ".. _abvega_offset_reffile:" in its first line (i.e. just move line 3 of this file into the file above), and similarly move the line ".. _abvega_offset_selectors:" into abvega_offset_selection.inc, and remove the 2 .rst files.

hbushouse: You also need to update the file `docs/jwst/references_general/references_general.rst` to add entires for these new abvega_offset docs into the 2 tables that give reffile vs. step and step vs. reffile. And also update the reference files section of the `source_catalog` step docs to list the new abvega_offset ref file.

jwst/datamodels/tests/test_models.py
jdavies-st: Is there a reason to write a `validate()` method instead of using `required` in the schema directly?  We do use `required` often elsewhere.  An example from our transforms:

https://github.com/spacetelescope/jwst/blob/95b306c76a21a67816cc15407445a5448e2fd0e9/jwst/transforms/schemas/stsci.edu/jwst_pipeline/coords-0.7.0.yaml#L39

(I'm asking mostly for my own education here.)

larrybradley: I think `required` would work only for the `abvega_offset` column, because that's the only column that appears in every schema.  The other columns depend on the instrument.  I'm not a schema expert, so please let me know if there is a way to use `required` in combination with `anyOf` in `properties`.

docs/jwst/references_general/abvega_offset_reffile.rst
larrybradley: Thanks, @hbushouse.  I've made those changes.

docs/jwst/references_general/abvega_offset_reffile.inc
larrybradley: Fixed, thanks!

docs/jwst/references_general/abvega_offset_reffile.inc
larrybradley: Done.

docs/jwst/references_general/abvega_offset_selection.inc
```diff
@@ -0,0 +1,16 @@
+Reference Selection Keywords for ABVEGA_OFFSET
```
hbushouse, line 1: Need to add ".. _abvega_offset_selectors:" reference at the top of this file.

larrybradley: This page isn't referenced (i.e., using `:ref:` anywhere).  It's included (`.. include::`) in the `abvega_offset_reffile` docs.  Does it still need a reference tag?

hbushouse: Perhaps not. We have it all the other ref file docs, but you're right that if the reference is never used anywhere, it's obviously not necessary.

docs/jwst/references_general/abvega_offset_reffile.rst
larrybradley: Also, I didn't update the reference files section of the `source_catalog` step in this PR.  I updated that section to include both the `apcorr` and `abvega_offset` reference files in my upcoming `source_catalog` PR that makes use of these reference files.

Issue #3838, opened 2019-07-19
extract_1d should copy slit-specific data


stscijgbot:
Issue [JP-880](https://jira.stsci.edu/browse/JP-880) was created by Philip Hodge:

The input and output of extract_1d are different data models, so it is necessary to explicitly copy some metadata, such as slit information and target information.  This is currently done for some models, such as MultiSlitModel, but it is not done for some other models, such as DrizProductModel.  If slit and target information is present in the input data, it should be copied to the output.

stscijgbot:
Comment by Howard Bushouse: Fixed by https://github.com/spacetelescope/jwst/pull/4685

PR #4685, opened 2020-03-16
JP-1340 Pixel-by-pixel calls to wcs removed from extract_1d

philhodge:
The two sections in `extract_1d` where the wcs function was called in a loop over pixels have been removed, leaving just the calls to `wcs` with array arguments.

This also resolves JP-880, GitHub #3838.  Function `copy_keyword_info` is now called in two more sections, covering `ImageModel`, `DrizProductModel`, `CubeModel`, `SlitModel`.  `copy_keyword_info` also copies new slit attributes `source_ra` and `source_dec`.

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -1183,23 +1183,20 @@ class ExtractBase:
             The declination of the target, or None
         """
 
+        targ_ra = targ_dec = None
+
         if slit is None:
             targ_ra = input_model.meta.target.ra
             targ_dec = input_model.meta.target.dec
         else:
-            targ_ra = None
-            targ_dec = None
-            if hasattr(slit, 'source_ra'):
-                targ_ra = slit.source_ra
-            if hasattr(slit, 'source_dec'):
-                targ_dec = slit.source_dec
+            targ_ra = getattr(slit, 'source_ra', default=None)
+            targ_dec = getattr(slit, 'source_dec', default=None)
 
         if targ_ra is None or targ_dec is None:
             log.warning("Target RA and Dec could not be determined.")
-            targ_ra = None
-            targ_dec = None
+            targ_ra = targ_dec = None
 
-        return (targ_ra, targ_dec)
+        return targ_ra, targ_dec
 
     def offset_from_offset(self, input_model, slit, verbose):
         """Get nod/dither pixel offset from the target coordinates.
@@ -1233,14 +1230,14 @@ class ExtractBase:
 
         targ_ra, targ_dec = self.get_target_coordinates(input_model, slit)
         if targ_ra is None or targ_dec is None:
-            return (0., None)
+            return 0., None
 
         # Use the WCS function to find the cross-dispersion (XD) location that
         # is closest to the target coordinates.  This is the "actual" location
         # of the spectrum, so the extraction region should be centered here.
         locn_info = self.locn_from_wcs(input_model, slit, targ_ra, targ_dec,
                                        verbose)
-        (middle, middle_wl, locn) = locn_info
+        middle, middle_wl, locn = locn_info
         if middle is not None and verbose:
             log.debug("Spectrum location from WCS used column (or row) %d",
                       middle)
@@ -1917,10 +1914,10 @@ class ExtractModel(ExtractBase):
             x_array.fill((self.xstart + self.xstop) / 2.)
 
         if self.wcs is not None:
-            stuff = self.wcs(x_array, y_array)
-            ra = stuff[0]
-            dec = stuff[1]
-            wcs_wl = stuff[2]
+            coords = self.wcs(x_array, y_array)
+            ra = coords[0]
+            dec = coords[1]
+            wcs_wl = coords[2]
             # We need one right ascension and one declination, representing
             # the direction of pointing.
             mask = np.isnan(wcs_wl)
@@ -2825,6 +2822,9 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
             copy_keyword_info(slit, slit.name, spec)
             output_model.spec.append(spec)
     else:
+        # These default values for slitname are not really slit names, and
+        # slitname may be assigned a better value below.  See the sections
+        # for input_model being an ImageModel or a SlitModel.
         slitname = input_model.meta.exposure.type
         if slitname is None:
             slitname = ANY
@@ -2870,11 +2870,12 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
             # through processing. Once the DrizProductModel schema is
             # updated to carry over the necessary slit meta data, this
             # should no longer be needed. See JP-1144.
+            # Replace the default value for slitname with a more accurate
+            # value, if possible.
             if input_model.meta.exposure.type == 'NRS_FIXEDSLIT':
                 slitname = input_model.meta.instrument.fixed_slit
-            elif hasattr(input_model, "name") and input_model.name is not None:
+            elif getattr(input_model, "name", default=None) is not None:
                 slitname = input_model.name
-            # else slitname was assigned above
 
             prev_offset = OFFSET_NOT_ASSIGNED_YET
             for sp_order in spectral_order_list:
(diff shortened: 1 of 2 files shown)


jwst/extract_1d/extract.py
```diff
@@ -1874,34 +1913,10 @@ def extract(self, data, wl_array, verbose):
...
-                        log.warning("n_inputs for wcs function is %d",
-                                    n_inputs)
-                        log.warning("WCS function was expected to take "
-                                    "either 2 or 3 arguments.")
-                    ra[:] = -999.
-                    dec[:] = -999.
-                    wcs_wl[:] = -999.
-            else:
-                ra, dec, wcs_wl = self.wcs(x_array, y_array)
+            stuff = self.wcs(x_array, y_array)
```
hbushouse: Wow, that's a lot simpler than the previous 20 lines of code! 👍

jwst/extract_1d/extract.py
```diff
@@ -3274,9 +3273,9 @@ def populate_time_keywords(input_model, output_model):
             spec = output_model.spec[n]             # n is incremented below
             spec.int_num = int_num[row]
             spec.time_scale = "UTC"
-            spec.start_time_mjd = start_utc[row]
-            spec.mid_time_mjd = mid_utc[row]
-            spec.end_time_mjd = end_utc[row]
+            spec.start_utc = start_utc[row]
```
hbushouse: #4602 changed the meta attribute names "start_utc", "mid_utc", and "end_utc" to "start_time_mjd", "mid_time_mjd", and "end_time_mjd" in order to make them consistent with the names for these keywords that are now used in the core schema. This change is reverting that update. The new names are used in the multispec schema and hence the code here also needs to use the new names (the old ones aren't defined in the schema anymore).

philhodge: Thanks; I'll fix this.

jwst/extract_1d/extract.py
```diff
@@ -1874,34 +1917,10 @@ def extract(self, data, wl_array, verbose):
...
-                                    "either 2 or 3 arguments.")
-                    ra[:] = -999.
-                    dec[:] = -999.
-                    wcs_wl[:] = -999.
-            else:
-                ra, dec, wcs_wl = self.wcs(x_array, y_array)
+            stuff = self.wcs(x_array, y_array)
+            ra = stuff[0]
+            dec = stuff[1]
+            wcs_wl = stuff[2]
```
nden: I wondered why you prefer to unpack in separate assignments instead of 
`ra, dec, wcs_wl = self.wcs(x_array, y_array)`

nden: `mask` is defined twice in this section, on line 1926 and further on line 1939. 
Also, if this is to cover the NIrspec case, I think it would be best for future development to check for `NaN` in the wavelength arrays, not in the celestial coordinates. 
For example, Nirspec now populates the three arrays (ra, dec, lam) with `NaNs`, but one can argue that this is incorrect because celestial coordinates are always valid. The NaNs are forced only in the wavelength array because some slits fall outside the detector or pixels on the detector fall outside the slit. At some point in the future I plan to change the Nirspec implementation (when I figure out how) and have NaNs only in the wavelength arrays. This will allow us to compute the footprint on the sky for Nirspec slits. I hope this makes sense.

philhodge: I'll change the name of `get_celestial_coordinates`; thanks for pointing that out.
I assigned the output of `wcs` to a variable because the function can return either three or four items, and it's cleaner (I think) to just assign the output to a single variable (a tuple) and then pick out what I need, than to first find out how many values the function returns, have an `if` block to call the function and get either three or four values, and then only use three of them.
I'll change the assignment to `mask` to use wavelength, rather than getting `mask` once for right ascension and again for declination.  I do think it would be better to only have to check one thing (wavelength).

jwst/extract_1d/extract.py
```diff
@@ -1155,6 +1155,52 @@ def update_extraction_limits(self, ap):
...
+                targ_ra = slit.source_ra
+            if hasattr(slit, 'source_dec'):
+                targ_dec = slit.source_dec
+
+        if targ_ra is None or targ_dec is None:
+            log.warning("Target RA and Dec could not be determined.")
+            targ_ra = None
+            targ_dec = None
+
+        return (targ_ra, targ_dec)
```
jwhite3:
```suggestion
    targ_ra = targ_dec = None

        if slit is None:
            targ_ra = input_model.meta.target.ra
            targ_dec = input_model.meta.target.dec
        else:
            targ_ra = getattr(slit, 'source_ra', default=None)
            targ_dec = getattr(slit, 'source_dec', default=None)

        if targ_ra is None or targ_dec is None:
            log.warning("Target RA and Dec could not be determined.")
            targ_ra = targ_dec = None

        return targ_ra, targ_dec
```
Minor suggestion

jwst/extract_1d/extract.py
```diff
@@ -1180,18 +1229,18 @@ def offset_from_offset(self, input_model, slit, verbose):
...
             direction, at the middle of the spectrum in the dispersion
             direction.
-
-        verbose : bool
-            If True, log messages.
         """
 
+        targ_ra, targ_dec = self.get_target_coordinates(input_model, slit)
+        if targ_ra is None or targ_dec is None:
+            return (0., None)
```
jwhite3:
```suggestion
            return 0., None
```
Parentheses here aren't necessary

jwst/extract_1d/extract.py
```diff
@@ -1180,18 +1229,18 @@ def offset_from_offset(self, input_model, slit, verbose):
...
         # Use the WCS function to find the cross-dispersion (XD) location that
         # is closest to the target coordinates.  This is the "actual" location
         # of the spectrum, so the extraction region should be centered here.
-        (middle, middle_wl, locn) = self.locn_from_wcs(
-                                        input_model,
-                                        slit,
-                                        verbose)
+        locn_info = self.locn_from_wcs(input_model, slit, targ_ra, targ_dec,
+                                       verbose)
+        (middle, middle_wl, locn) = locn_info
```
jwhite3:
```suggestion
        middle, middle_wl, locn = locn_info
```
Don't need the parentheses here

jwst/extract_1d/extract.py
```diff
@@ -1874,59 +1917,28 @@ def extract(self, data, wl_array, verbose):
...
-                        log.warning("n_inputs for wcs function is %d",
-                                    n_inputs)
-                        log.warning("WCS function was expected to take "
-                                    "either 2 or 3 arguments.")
-                    ra[:] = -999.
-                    dec[:] = -999.
-                    wcs_wl[:] = -999.
-            else:
-                ra, dec, wcs_wl = self.wcs(x_array, y_array)
+            stuff = self.wcs(x_array, y_array)
```
jwhite3: `stuff` seems like a really ambiguous variable name. Is there another name that might be more clear while remaining concise?

jwst/extract_1d/extract.py
```diff
@@ -2885,6 +2872,9 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
             # should no longer be needed. See JP-1144.
             if input_model.meta.exposure.type == 'NRS_FIXEDSLIT':
                 slitname = input_model.meta.instrument.fixed_slit
+            elif hasattr(input_model, "name") and input_model.name is not None:
```
jwhite3:
```suggestion
            elif getattr(input_model, "name", default=None) is not None:
```

jwst/extract_1d/extract.py
```diff
@@ -2969,14 +2959,19 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
...
+                # intended to be a SlitModel object, but ImageModel now has
+                # the attributes we will look for in this function.
+                copy_keyword_info(input_model, slitname, spec)
                 output_model.spec.append(spec)
 
         elif isinstance(input_model, (datamodels.CubeModel,
                                       datamodels.SlitModel)):
 
             slit = None
+            if hasattr(input_model, "name") and input_model.name is not None:
```
jwhite3:
```suggestion
            if getattr(input_model, "name", default=None) is not None:
```

I've seen this bit a couple of times now, what does "else slit name was assigned above" mean here?

Could also just assign `slitname` directly perhaps: 
```python
slitname = getattr(input_model, 'name')
```

philhodge: @jwhite3 Those are all good suggestions; thanks.

jwhite3:
@philhodge looks like `getattr` doesn't take keyword arguments, so everywhere where `default` is specified with a keyword argument needs to be updated to a positional one. 

So: 
```python
slitname = getattr(input_model, 'name', None)
```
For example. 

Apologies for the incorrect recollection of that function!

philhodge:
Interesting!  Thanks for pointing that out, @jwhite3 .

PR #4693, opened 2020-03-17
JP-1359 Two recent changes to extract_1d were backed out

philhodge:
For WFSS and NRC_TSGRISM data, don't try to determine the pixel location of the target spectrum from the wcs.  Also, for NRS_BRIGHTOBJ data, leave `slitname` set to 'NRS_BRIGHTOBJ', rather than changing it to the actual slit name, 'S1600A1'.

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -1237,7 +1237,10 @@ class ExtractBase:
         # of the spectrum, so the extraction region should be centered here.
         locn_info = self.locn_from_wcs(input_model, slit, targ_ra, targ_dec,
                                        verbose)
-        middle, middle_wl, locn = locn_info
+        if locn_info is None:
+            middle = middle_wl = locn = None
+        else:
+            middle, middle_wl, locn = locn_info
         if middle is not None and verbose:
             log.debug("Spectrum location from WCS used column (or row) %d",
                       middle)
@@ -1292,7 +1295,8 @@ class ExtractBase:
 
         Returns
         -------
-        middle : int or None
+        tuple (middle, middle_wl, locn) or None
+        middle : int
             Pixel coordinate in the dispersion direction within the 2-D
             cutout (or the entire input image) at the middle of the WCS
             bounding box.  This is the point at which to determine the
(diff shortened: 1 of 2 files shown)


jwst/extract_1d/extract.py
```diff
@@ -1313,6 +1313,13 @@ def locn_from_wcs(self, input_model, slit, targ_ra, targ_dec, verbose):
             wavelength attribute or wcs function is not defined.
         """
 
+        # WFSS data are not currently supported.
+        if input_model.meta.exposure.type in WFSS_EXPTYPES:
+            log.warning("For exposure type %s, we currently can't use "
+                        "target coordinates to get location of spectrum.",
+                        input_model.meta.exposure.type)
+            return None, None, None
```
jwhite3: This seems odd, but I understand if this method is expected to return a 3-tuple. 

Would it be better maybe to assign those values to attributes instead and then return them?

jwhite3: This seems odd, but I understand if this method is expected to return a 3-tuple.

Would it be better maybe to assign those values to attributes instead and then return them?

jwst/extract_1d/extract.py
```diff
@@ -2876,6 +2882,7 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
                 slitname = input_model.meta.instrument.fixed_slit
             elif getattr(input_model, "name", None) is not None:
                 slitname = input_model.name
+            log.debug('slitname=%s', slitname)
```
jwhite3:
```suggestion
            log.debug(f'slitname={slitname}')
```

jwst/extract_1d/extract.py
```diff
@@ -2972,8 +2979,9 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
             slit = None
             # Replace the default value for slitname with a more accurate
             # value, if possible.
-            if getattr(input_model, "name", None) is not None:
-                slitname = input_model.name
+            # xxx if getattr(input_model, "name", None) is not None:
+            # xxx     slitname = input_model.name
+            log.debug('slitname=%s', slitname)
```
jwhite3:
```suggestion
            log.debug(f'slitname={slitname}')
```

jwst/extract_1d/extract.py
philhodge: That was odd, I agree.  I changed it to return a single None.

PR #4710, opened 2020-03-18
Changes to enable nirspec fixedslit autowave lamp mode in calspec2

stscirij:
Addresses JP-1306.  Replaces #4626, this should be easier to review as it's rebased

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,7 @@ ami
 assign_wcs
 ----------
 
-- Added code to enable processing of NIRSPEC FIXEDSLIT AUTOWAVE data [#4626]
+- Added code to enable processing of NIRSPEC FIXEDSLIT AUTOWAVE data [#4710]
 
 associations
 ------------
@@ -21,13 +21,16 @@ coron
 
 - Reorganized step documentation [#4697]
 
+datamodels
+----------
+
+- Updated core schema to include enum values for lamp_mode attribute [#4710]
+
 extract_1d
 ----------
 
 - Remove pixel-by-pixel calls to wcs; copy input keywords to output for
   more types of input data. [#4685]
-
-- Added code to enable processing of NIRSPEC FIXEDSLIT AUTOWAVE data [#4626]
 
 extract_2d
 ----------
@@ -35,10 +38,18 @@ extract_2d
 - Change the source type for NIRSpec MOS sources with stellarity = -1 from
   UNKOWN to POINT. [#4686]
 
+- Added code to enable processing of NIRSPEC FIXEDSLIT AUTOWAVE data [#4710]
+
 flatfield
 ---------
 
-- Added code to enable processing of NIRSPEC FIXEDSLIT AUTOWAVE data [#4626]
+- Added code to enable processing of NIRSPEC FIXEDSLIT AUTOWAVE data [#4710]
+
+pipeline
+--------
+
+- Added code to enable the ``calwebb_spec2`` pipeline to process NIRSPEC FIXEDSLIT
+  AUTOWAVE data [#4710]
 
 srctype
 -------
@@ -208,9 +219,6 @@ pipeline
 - Fixed the ``calwebb_tso3`` calculation of the number_of_integrations recorded
   in the photometric table product to avoid ``astropy.table`` merge conflicts.
   [#4502]
-
-- Added code to enable the ``calwebb_spec2`` pipeline to process NIRSPEC FIXEDSLIT
-  AUTOWAVE data [#4626]
 
 photom
 ------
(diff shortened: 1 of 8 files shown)


CHANGES.rst
```diff
@@ -6,6 +6,11 @@ ami
 
 - Reorganized step documentation [#4697]
 
+assign_wcs
+----------
+
+- Added code to enable processing of NIRSPEC FIXEDSLIT AUTOWAVE data [#4626]
```
nden: The PR number has changed now to #4710

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -511,6 +511,7 @@ properties:
           lamp_mode:
             title: Lamp operating mode
             type: string
+            enum: [BRIGHTOBJ, FIXEDSLIT, GRATING-ONLY, IFU, MSASPEC, NULL]
```
nden: What is the `GRATING-ONLY` mode?

nden:
The logic for wavecorr in extract_2d does not handle the case for `AUTOWAVE` and `AUTOFLAT`.
Is wavelength correction to be run for these modes? In any case the logic needs to be tweaked because currently it will apply the MOS reference file to any observation with EXP_TYPE different from `NRS_FIXEDSLIT` and `NRS_BRIGHTOBJ`.

https://github.com/spacetelescope/jwst/blob/master/jwst/extract_2d/nirspec.py#L277

CHANGES.rst
```diff
@@ -22,12 +27,19 @@ extract_1d
 - Remove pixel-by-pixel calls to wcs; copy input keywords to output for
   more types of input data. [#4685]
 
+- Added code to enable processing of NIRSPEC FIXEDSLIT AUTOWAVE data [#4626]
```
hbushouse: Was this entry meant to go under `extract_2d`?

jwst/datamodels/schemas/core.schema.yaml
stscirij: From jp-2927:
Comment by James Muzerolle: GRATING_ONLY is a limited special case that is used to move the grating wheel without taking any exposures (no lamp is ever turned on).  You can ignore it for the purposes of processing lamp data.

CHANGES.rst
stscirij: Thanks!  I'll change that

CHANGES.rst
stscirij: Yes, thanks for spotting that

nden:
@stscirij Do you want also to change the logic in [extract_2d/apply_zero_point_correction](https://github.com/spacetelescope/jwst/blob/master/jwst/extract_2d/nirspec.py#L277) to either not include the two modes mentioned above or handle them properly? Do you know what the iNS recommendation is for those modes?

stscirij:
I don't know what the INS recommendation is for that case.  However, the flat_field steps require division by the S-flat for NRS_LAMP LINE lamp exposures and AUTOWAVE exposures, but not for NRS_LAMP FLAT lamp exposures or AUTOFLAT.  I would interpret that to mean that wavelength calibration is to be done for line lamp data but not flatfield data.

stscirij:
Perhaps this should be handled by setting the steps.extract_2d.apply_wavecorr parameter in a calwebb_spec2.cfg file?

stscirij:
@nden I've just got back to looking at the wavelength correction for internal lamp modes.  I don't see how the wavelength correction can be done in exactly the same way that it was done for the science exposures because the source type for the lamp exposures won't be 'point' and the source_xpos won't be known (unless the values for the associated science exposures get transferred to the lamp exposures).  Any ideas?  Am I missing something?

hbushouse:
Do the specs (requirements) from the NIRSpec team indicate that any of the lamp mode exposures should receive the wavelength shift correction? As you say, I don't see how that's possible. That correction is only for point sources that are not centered in a slit/slitlet. Lamp exposures have no such concept. I think wavelength correction has to be skipped for these.

stscirij:
I'll ask.  Although one usually wants the same sequence of operations to be done on calibration exposures as are done on science exposures - unless the wavelength zero point correction is supposed to modify the science data to make it look like the source is in the center of the aperture and thus a non-wavelength corrected lamp exposure is appropriate?

hbushouse:
That's correct (wavecorr is simply to make a science exposure look like the source is centered). There's no concept of centering for lamp illumination that completely fills the slit, so it has to be skipped (same as for any extended source).

nden:
OK, in this case, these exposures need to be skipped in extact_2d/nirspec.wavecorr_supported_modes. Otherwise it looks good to me.

hbushouse:
@rij has this PR been superseded by #4889 ? If so, can this one be closed?

stscirij:
Yes, it's superseded by #4889, so I'll close it

PR #4721, opened 2020-03-20
Add a step generating WCS for a moving target

### LINKED ISSUE #4649: Implement MT WCS transform
Issue [JP-1336](https://jira.stsci.edu/browse/JP-1336) was created by Jonathan Eisenhamer:

Implement MT WCS transform

nden:
Resolves #4649

This is a working skeleton of the step - to provide context for questions.

The input is an association and the output is a `ModelContainer`. Is this OK?
The association is read into a `ModelContainer` and each model in the container is modified in place.

*Questions:* 

- The step iterates over `ModelContainer._models`. Is there a better way to do this, i.e. not using private variables?
- When I run the step on the command line it creates 4 new files on disk named `mt_asn_0_assign_mtwcs.fits`. What am I doing wrong or is this expected?

- What keywords define an exposure as a moving target (other than `MT_RA` and `MT_DEC`)?
    (EDIT: I found `model.meta.target.type` with a possible value of `moving`)

### DIFF
--- a/docs/jwst/assign_mtwcs/index.rst
+++ b/docs/jwst/assign_mtwcs/index.rst
@@ -1,0 +1,12 @@
+.. _assign_mtwcs_step:
+
+========================
+Assign Moving Target WCS
+========================
+
+.. toctree::
+   :maxdepth: 1
+
+   main.rst
+
+.. automodapi:: jwst.assign_mtwcs
--- a/docs/jwst/assign_mtwcs/main.rst
+++ b/docs/jwst/assign_mtwcs/main.rst
@@ -1,0 +1,30 @@
+
+Description
+===========
+
+``jwst.assign_mtwcs`` is run in the beginning of the level 3 JWST pipeline.
+It assigns a WCS to a ``Moving Target`` exposure. The table below defines the
+keywords specific to exposures with moving targets:
+
++--------------+----------------------+----------+------------------------------------------------+
+| FITS keyword | Data model attribute | Type     | Description                                    |
+|              |                      | (Value)  |                                                |
++==============+======================+==========+================================================+
+| TARGTYPE     | meta.target.type     | string   | Type of target                                 |
+|              |                      | (moving) |                                                |
++--------------+----------------------+----------+------------------------------------------------+
+| MT_RA,       | meta.wcsinfo.mt_ra,  | number,  | Average right ascension and declination        |
+| MT_DEC       | meta.wcsinfo.mt_dec  | number   | of the moving target during exposure [deg]     |
++--------------+----------------------+----------+------------------------------------------------+
+| MT_AVRA,     | meta.wcsinfo.mt_avra,| number,  | Right ascension and declination of the         |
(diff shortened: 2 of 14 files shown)


jdavies-st:
- `datamodels.open()` on an association returns a `ModelContainer` of the association science datamodels.  What you're doing looks good.

- To iterate over models in a container:

       for model in container:
           model.meta.wcs = blah

- `ModelContainer` is a list of datamodels which on serialization, calls the `save` method for each model.  If you want to write out specific filenames, one can specify a suffix or one can populate `meta.filename` in each datamodel.  Perhaps @stscieisenhamer can comment on why the `0` appears in the filename by default?  Is there a way to avoid this?

nden:
I guess the number in the name is the only way to generate different names.

The `test_suffix_existence` test is failing. Should I add suffixes somewhere?

jdavies-st:
The allowed suffixes are in

https://github.com/spacetelescope/jwst/blob/9781353c8a07a12e1c62227190e84f507639d819/jwst/lib/suffix.py#L63

I.e. these are the suffixes that will get replaced.  Would be nice if these were autogenerated on-the-fly.  That said, there is a function for generating them on-the-fly.

https://github.com/spacetelescope/jwst/blob/9781353c8a07a12e1c62227190e84f507639d819/jwst/lib/suffix.py#L257

stscieisenhamer:
The automatic output file name creation sees that it will be a list of files and will append an index number to the automatic name produced.  This is triggered by the fact it is saving a `ModelContainer`.

To avoid this, the following should occur:

- The `meta.filename` attribute for each model in the container needs to be set. Use `Step.make_output_path`, though really look at `Strep._make_output_path` for the parameters.
- To force using `meta.filename`, either modify your Step's `spec` definition or in the `process` method, set `self.output_use_model = True`

When creating the output names with `Step.make_output_path`, you will need to change `basepath` for each model in some way so that the names are unique.

stscieisenhamer:
Actually, about making the names unique: Most likely that will not be necessary, since the `meta.filename` attributes are probably already unique. If you have defined a new suffix, that will be sufficient.

The ideal spot for a new suffix is in the `spec` definition:

```
spec = """
suffix             = string(default=MyNewSuffix)        # Default suffix for output files
output_use_model   = boolean(default=True)      # When saving use `DataModel.meta.filename`
"""
```

nden:
@stscieisenhamer Thanks! That helped with file names. How do I fix [this error](https://travis-ci.org/github/spacetelescope/jwst/jobs/664980071#L1450)? Do I need to add the suffix somewhere in `jwst/lib`?

stscieisenhamer:
Yes, sorry:  Update `suffix._calculate_suffixes`.  

That should be made clear in the docstring. And probably renamed to something like `calculated_suffix_cache`. Its an efficiency thing so that the potentially long and fragile `find_suffixes` isn't called everytime. I'll file an issue to do that.

nden:
I'm in the process of writing the docs. Should I add the 4 simulated files as a regtest?

hbushouse:
> I'm in the process of writing the docs. Should I add the 4 simulated files as a regtest?

Yes, add the regtest!

PR #4722, opened 2020-03-24
add a regression test for moving target

nden:
This is my first regression tests in the new system. Comments welcome.

I'm waiting for artifactory to come back to upload the data.

### DIFF
--- a/jwst/lib/exposure_types.py
+++ b/jwst/lib/exposure_types.py
@@ -25,6 +25,6 @@ FGS_GUIDE_EXP_TYPES = [
 def is_moving_target(input_models):
     """ Determine if a moving target exposure."""
     model = input_models[0]
-    if model.meta.target.type == 'moving':
+    if model.meta.target.type.lower() == 'moving':
         return True
     return False
--- a/jwst/regtest/test_nircam_image.py
+++ b/jwst/regtest/test_nircam_image.py
@@ -129,3 +129,16 @@ def test_nircam_image_stage3_catalog(run_image3pipeline, rtdata_module, diff_ast
     rtdata.get_truth("truth/test_nircam_image_stages/jw42424-o002_t001_nircam_clear-f444w_cat.ecsv")
 
     assert diff_astropy_tables(rtdata.output, rtdata.truth, rtol=1e-4, atol=1e-5)
+
+
+@pytest.mark.bigdata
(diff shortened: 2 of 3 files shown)


hbushouse:
I agree with @ddavis-stsci that it's probably best to keep this test in a separate module (e.g. test_nircam_mtimage). The input data can live in the same directory as the other NIRCam imaging tests, but then there should be a separate truth directory that matches the test name (e.g. test_nircam_mtimage). Of course that means adding more code to the test module itself to run the pipeline, specify the input directory, etc.

nden:
I moved the test to a separate module and the data is uploaded to the suggested locations. The test passes locally. Ready for review. Not sure why the Travis tests disappeared.

jdavies-st:
CI test failures are unrelated, so fine to merge.

jwst/lib/exposure_types.py
```diff
@@ -25,6 +25,6 @@
 def is_moving_target(input_models):
     """ Determine if a moving target exposure."""
     model = input_models[0]
-    if model.meta.target.type == 'moving':
+    if model.meta.target.type.lower() == 'moving':
```
hbushouse: @jdavies-st points out that many regtests are now failing because we have a lot of data files for non-moving target exposures that don't have the TARGTYPE keyword. So this logic should be changed slightly to first check for existence of model.meta.target.type and default to normal processing if it's not there.

nden: Will do. 
Is this a required keyword? In the future can we expect that it will be always populated (by SDP)?

hbushouse: Yes, it's a required keyword in the KWD and SDP, and as such will always be populated with some string, where "UNKNOWN" is the default (in case SDP can't determine a value for it). So your logic is correct in that we should only ever apply mt processing if it's set to "MOVING". It's just some of our old test data that don't have it.

jdavies-st: Checking

```
if model.meta.target.type is not None and model.meta.target.type.lower() == 'moving'
```

does the trick.  There's another change I'm adding in to our miri image regtests based on the failures last night.  I can do both of these in a PR @nden.

It is just taking a very long time to run the tests from home.

nden: I saw your offer too late.

Issue #4337, opened 2019-12-11
outlier_detection not working


Guang91:
I'm running the pipeline on some MIRI images. The outlier_detection (stage 3) successfully detects some CR outliers. But it only changes the DQ values of CR pixels to 4. Then in the step (resampling), these pixels are considered "good" and the mosaic image still have these CR outliers. That is to say, outlier_detection does not affect the final science image. 

I think it's better to flag the CR pixels as a higher value (say 14) in the DQ map. Then the resampling step can reject the CR outliers. 

Is it a reasonable approach?    

Well... After setting it to 14, I realized too many pixels are identified as CR. The outlier_detection step seems to be *very* ineffective.

jdavies-st:
Hi @Guang91, I think this is probably fixed in #4726.  Could you verify that it is working properly for your simulated MIRI data?  We were not setting DO_NOT_USE properly, and this should now be fixed.

Generally, any pixels that are not to be used in a pipeline step are set to 1, i.e. DO_NOT_USE.  Every other bit flag is informational only.  It looks like we don't have this documented very well in the docs.  But to find out what is what:

```
In [21]: from jwst.datamodels.dqflags import pixel

In [22]: pixel
Out[22]: 
{'GOOD': 0,
 'DO_NOT_USE': 1,
 'SATURATED': 2,
 'JUMP_DET': 4,
 'DROPOUT': 8,
 'RESERVED_1': 16,
 'RESERVED_2': 32,
 'RESERVED_3': 64,
 'RESERVED_4': 128,
 'UNRELIABLE_ERROR': 256,
 'NON_SCIENCE': 512,
 'DEAD': 1024,
 'HOT': 2048,
 'WARM': 4096,
 'LOW_QE': 8192,
 'RC': 16384,
 'TELEGRAPH': 32768,
 'NONLINEAR': 65536,
 'BAD_REF_PIXEL': 131072,
 'NO_FLAT_FIELD': 262144,
 'NO_GAIN_VALUE': 524288,
 'NO_LIN_CORR': 1048576,
 'NO_SAT_CHECK': 2097152,
 'UNRELIABLE_BIAS': 4194304,
 'UNRELIABLE_DARK': 8388608,
 'UNRELIABLE_SLOPE': 16777216,
 'UNRELIABLE_FLAT': 33554432,
 'OPEN': 67108864,
 'ADJ_OPEN': 134217728,
 'UNRELIABLE_RESET': 268435456,
[...]

PR #4726, opened 2020-03-25
JP-787 Change outlier and resample DQ bit usage

### LINKED ISSUE #3641: Rework the way outlier_detection and resample make use of input DQ flags
Issue [JP-787](https://jira.stsci.edu/browse/JP-787) was created by Howard Bushouse:

The outlier_detection and resample steps currently use input DQ flags in a different way than other jwst steps. In particular, they use the "good_pixels" parameter to tell the step which DQ flags are to be considered as OK and hence to NOT reject pixels with those DQ flags. This logic needs to be inverted, such that a parameter indicates which DQ flag values to be considered as BAD and hence reject any pixels with those values.

stscieisenhamer:
Resolves #3641 
Resolves [JP-787](https://jira.stsci.edu/browse/JP-787)

### DIFF
--- a/docs/jwst/references_general/references_general.rst
+++ b/docs/jwst/references_general/references_general.rst
@@ -603,3 +603,47 @@ Bit  Value         Name              Description
 ===  ==========    ================  ===========================================
 
 Note: Words like "highly" and "large" will be defined by each instrument team.  They are likely to vary from one detector to another – or even from one observing mode to another.
+
+.. _`dq_parameter_specification`:
+
+Parameter Specification
+=======================
+
+There are a number of steps, such as :ref:`OutlierDetectionStep
+<outlier_detection_step>` or :ref:`SkyMatchStep <skymatch_step>`, that define
+what data quality flags a pixel is allowed to have to be considered in
+calculations. Such parameters can be set in a number of ways.
+
+First, the flag can be defined as the integer sum of all the DQ bit values from
+the input images DQ arrays that should be considered "good". For example, if
+pixels in the DQ array can have combinations of 1, 2, 4, and 8 and one wants to
+consider DQ flags 2 and 4 as being acceptable for computations, then the
+parameter value should be set to "6" (2+4). In this case a pixel having DQ values
+2, 4, or 6 will be considered a good pixel, while a pixel with a DQ value, e.g.,
+1+2=3, 4+8="12", etc. will be flagged as a "bad" pixel.
+
+Alternatively, one can enter a comma-separated or '+' separated list of integer
+bit flags that should be summed to obtain the final "good" bits. For example,
+both "4,8" and "4+8" are equivalent to a setting of "12".
+
+Finally, instead of integers, the JWST mnemonics, as defined above, may be used.
+For example, all the following specifications are equivalent:
+
+`"12" == "4+8" == "4, 8" == "JUMP_DET, DROPOUT"`
+
+.. note::
+   - The default value (0) will make *all* non-zero
+     pixels in the DQ mask be considered "bad" pixels and the
+     corresponding pixels will not be used in computations.
+
+   - Setting to `None` will turn off the use of the DQ array
+     for computations.
+
+   - In order to reverse the meaning of the flags
+     from indicating values of the "good" DQ flags
+     to indicating the "bad" DQ flags, prepend '~' to the string
+     value. For example, in order to exclude pixels with
+     DQ flags 4 and 8 for computations and to consider
+     as "good" all other pixels (regardless of their DQ flag),
+     use a value of ``~4+8``, or ``~4,8``. A string value of
+     ``~0`` would be equivalent to a setting of ``None``.
--- a/docs/jwst/references_general/references_general.rst
+++ b/docs/jwst/references_general/references_general.rst
@@ -572,7 +572,7 @@ Bit  Value         Name              Description
 1    2             SATURATED         Pixel saturated during exposure
 2    4             JUMP_DET          Jump detected during exposure
 3    8             DROPOUT           Data lost in transmission
-4    16            RESERVED
+4    16            OUTLIER           Removed by outlier detection
 5    32            RESERVED
 6    64            RESERVED
 7    128           RESERVED
--- a/docs/jwst/references_general/references_general.rst
+++ b/docs/jwst/references_general/references_general.rst
@@ -572,7 +572,7 @@ Bit  Value         Name              Description
 1    2             SATURATED         Pixel saturated during exposure
 2    4             JUMP_DET          Jump detected during exposure
 3    8             DROPOUT           Data lost in transmission
-4    16            OUTLIER           Removed by outlier detection
+4    16            OUTLIER           Flagged by outlier detection
 5    32            RESERVED
 6    64            RESERVED
 7    128           RESERVED
--- a/docs/jwst/skymatch/README.rst
+++ b/docs/jwst/skymatch/README.rst
@@ -137,6 +137,13 @@ The ``skymatch`` step has the following optional arguments:
   final "good" bits. For example, both ``4,8`` and ``4+8``
   are equivalent to setting ``dqbits`` to 12.
 
+  Finally, instead of integers, the JWST mnemonics for the bitflags may
+  be used. For example, all the following specifications are equivalent:
+
+  `12 == 4 + 8 == "4, 8" == "JUMP_DET, DROPOUT"`
+
+  The DQ flag mnemonics are found in :ref:`Data Quality Flags`.
+
   .. note::
     - The default value (0) will make *all* non-zero
       pixels in the DQ mask be considered "bad" pixels and the
--- a/docs/jwst/skymatch/README.rst
+++ b/docs/jwst/skymatch/README.rst
@@ -121,45 +121,9 @@ The ``skymatch`` step has the following optional arguments:
   value computations. Supported values are: 'mean', 'mode', 'midpt',
   and 'median'. (Default = 'mode')
 
-* ``dqbits`` (str): 
-  Integer sum of all the DQ bit values from the input images
-  DQ arrays that should be considered "good" when building masks for
-  sky computations. For example, if pixels in the DQ array can have
-  combinations of 1, 2, 4, and 8 and one wants to consider DQ
-  flags 2 and 4 as being acceptable for sky
-  computations, then ``dqbits`` should be set to 6 (2+4). In this
-  case a pixel having DQ values 2, 4, or 6 will be considered a good pixel,
-  while a pixel with a DQ value, e.g., 1+2=3, 4+8=12, etc. will be flagged as
-  a "bad" pixel.
-
-  Alternatively, one can enter a comma-separated or '+' separated list
-  of integer bit flags that should be summed to obtain the
-  final "good" bits. For example, both ``4,8`` and ``4+8``
-  are equivalent to setting ``dqbits`` to 12.
-
-  Finally, instead of integers, the JWST mnemonics for the bitflags may
-  be used. For example, all the following specifications are equivalent:
-
-  `12 == 4 + 8 == "4, 8" == "JUMP_DET, DROPOUT"`
-
-  The DQ flag mnemonics are found in :ref:`Data Quality Flags`.
-
-  .. note::
-    - The default value (0) will make *all* non-zero
-      pixels in the DQ mask be considered "bad" pixels and the
-      corresponding image pixels will not be used for sky computations.
-
-    - Set ``dqbits`` to `None` to turn off the use of image's DQ array
-      for sky computations.
-
-    - In order to reverse the meaning of the ``dqbits``
-      parameter from indicating values of the "good" DQ flags
-      to indicating the "bad" DQ flags, prepend '~' to the string
-      value. For example, in order to exclude pixels with
-      DQ flags 4 and 8 for sky computations and to consider
-      as "good" all other pixels (regardless of their DQ flag),
-      set ``dqbits`` to ``~4+8``, or ``~4,8``. A ``dqbits`` string value of
-      ``~0`` would be equivalent to setting ``dqbits=None``.
+* ``dqbits`` (str): The DQ bit values from the input images' DQ arrays that
+  should be considered "good" when building masks for sky computations. See
+  DQ flag :ref:`dq_parameter_specification` for details. (Default='0')
 
 * ``lower`` (float):
   An optional value indicating the lower limit of usable pixel
--- a/jwst/datamodels/dqflags.py
+++ b/jwst/datamodels/dqflags.py
@@ -5,7 +5,7 @@ The definitions are documented in the JWST RTD:
 https://jwst-pipeline.readthedocs.io/en/latest/jwst/references_general/references_general.html#data-quality-flags
 
 
-Impementation
+Implementation
 -------------
 
 The flags are implemented as "bit flags": Each flag is assigned a bit position
@@ -16,6 +16,9 @@ The data structure that stores bit flags is just the standard Python `int`,
 which provides 32 bits. Bits of an integer are most easily referred to using
 the formula `2**bit_number` where `bit_number` is the 0-index bit of interest.
 """
+from astropy.nddata.bitmask import interpret_bit_flags as ap_interpret_bit_flags
+
+from jwst.lib.basic_utils import multiple_replace
 
 # Pixel-specific flags
 pixel = {'GOOD':             0,      # No bits set, all is good
@@ -62,3 +65,44 @@ group = {'GOOD':       pixel['GOOD'],
          'JUMP_DET':   pixel['JUMP_DET'],
          'DROPOUT':    pixel['DROPOUT'],
 }
+
+
+def interpret_bit_flags(bit_flags, flip_bits=None):
+    """Converts input bit flags to a single integer value (bit mask) or `None`.
+
+    Wraps `astropy.nddate.bitmask.interpret_bit_flags`, allowing the JWST
+    bit mnemonics to be used in place of integers.
+
+    Parameters
+    ----------
+    bit_flags : int, str, list, None
+        See `astropy.nddate.bitmask.interpret_bit_flags`.
+        Also allows strings using JWST mnemonics
+
+    flip_bits : bool, None
+        See `astropy.nddate.bitmask.interpret_bit_flags`.
+
+    Returns
+    -------
+    bitmask : int or None
+        Returns an integer bit mask formed from the input bit value or `None`
+        if input ``bit_flags`` parameter is `None` or an empty string.
+        If input string value was prepended with '~' (or ``flip_bits`` was set
+        to `True`), then returned value will have its bits flipped
+        (inverse mask).
+
+    Examples
+    --------
+    Using JWST mnemonics:
+    TBD
+    """
+    bit_flags_dm = bit_flags
+    if isinstance(bit_flags, str):
+        dm_flags = {
+            key: str(val)
+            for key, val in pixel.items()
+        }
+        bit_flags_dm = multiple_replace(bit_flags, dm_flags)
+
+    return ap_interpret_bit_flags(bit_flags_dm, flip_bits=flip_bits)
+
--- a/jwst/datamodels/dqflags.py
+++ b/jwst/datamodels/dqflags.py
@@ -105,4 +105,3 @@ def interpret_bit_flags(bit_flags, flip_bits=None):
         bit_flags_dm = multiple_replace(bit_flags, dm_flags)
 
     return ap_interpret_bit_flags(bit_flags_dm, flip_bits=flip_bits)
-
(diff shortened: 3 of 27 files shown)


stscieisenhamer:
## Reviewers

The basics of moving from the definition of `good-bits` to `bad-bits` has been implemented. The PR has been put up to confirm both 1) this is what was desired and 2) have a preliminary verification of test results.

A quick test that does exercise the changes is `jwst/regtest/test_miri_image.py`. The changes are complicated because of the move of definition. Previously only data with DQ flags of [GOOD, SATURATED, JUMP_DET], value of `6`, were considered in calculations. Now, all DQ flags are valid except for flag of DO_NOT_USE.

stscieisenhamer:
## Question

`skymatch` has a similar parameter, `dqbits`. Not that this PR should handle this, but should `skymatch` be modified to be consistent with `outlier`?

jdavies-st:
Is there a reason to have `bad-bits` at all?  Can't resample just ignore anything that is set to `DO_NOT_USE`, like every other step and leave it at that?  And `outlier_detection` can set it's detected outliers to `DO_NOT_USE`?  This greatly would simplify the code.

Force steps upstream to make a decision about `DO_NOT_USE`.  If people want to add or subtract from `DO_NOT_USE` before resampling, then do it.  But `resample` should not be the place to do that.

Or if you disagree, and we do want to exclude specific DQ mnemonics other than `DO_NOT_USE` in `resample` then have an arg that is a list of the mnemonics.  The idea would be to avoid having to force users to add bits.  This would make the interface more consistent with the other pipeline steps and much easier to understand.

stscieisenhamer:
Good point. Trivial to do since it uses the same base function. All that is needed is to change the spec to a string.

stscieisenhamer:
And now for the "DOH" moment. If setup as a string, no code changes need be done, except for changing the spec's to strings, and making the default "~1". This means that for `good_bits`, all bits except for DO_NOT_USE are good.

So @hbushouse gets an executive decision: Leave the parameter as `good_bits` and change the default "~1". OR, change the interface to `bad_bits` and have the default as "1".

Either way, methinks having the ability to use the JWST mnemonics, such as "DO_NOT_USE", would also be advantageous. And straightforward to do.

philhodge:
My vote is to remove `good_bits` and not replace it with anything.  Just check for DO_NOT_USE, as @jdavies-st suggested.

hbushouse:
Haven't done my own review yet and only marginally keeping up with comments so far, but for the `outlier_detection` step in particular, the original purpose of `good_bits` (from HST use) was to allow pixels in the input images that had already gone through a previous round of CR detection and hence already had CR flags set in their DQ array, to still be used in the current round of CR detection (e.g. if a user wanted to experiment with different threshold levels during off-line use). Hence usually `good_bits` would be set to the CR flag value. When it got ported to the jwst pipeline, the flag value for saturation got added to that, because INS had asked for both CR and saturation flag values to be passed along in the output from the ramp fitting step, even if the pixel values themselves were OK (because ramp fitting had accounted for the saturation and CRs).

But I think that if we switch to having the routine either hardwired to ignore DO_NOT_USE or have a user param that defaults to DO_NOT_USE, then we essentially accomplish the same thing as the original intent of `good_bits`. Pixels flagged upstream with CR or saturation, but do NOT have DO_NOT_USE set, will be treated as good and used in the calculations. However, that would only work on the first pass of the data through the step, because if the step sets DO_NOT_USE on output (in addition to a CR flag), then on the next iteration of the step (if a user does that), those pixels will get excluded when we don't necessarily want them to be.

So I think we need a hybrid solution of some sort. Perhaps have the task hardwired to exclude input pixels with DO_NOT_USE, which could be overridden by a user param (`good_bits`?). For example, if a pixel was flagged as CR in a previous run of the step and hence has both DO_NOT_USE and CR flags set, but `good_bits` is set to the CR value, then that would tell the step to go ahead use that pixel despite the DO_NOT_USE flag.

Downstream, the `resample` step would/should then always ignore anything with DO_NOT_USE set.

stscieisenhamer:
The desired behavior can be accomplished now using the existing `good_bits` parameter. Reminder that the CR flag is "JUMP_DET", or a value of "4". The below also uses the, now implemented, ability to use the flag mnemonics in the `good-bits` specification:

The important syntax to not here is the "~". It means "invert the resulting bitmask".

- Default `good_bits` for all steps "~DO_NOT_USE"
   This allows all flags except DO_NOT_USE. This allows both running with data that has and has not already been CR flagged

- `outlier_detection_step.good_bits` == "~DO_NOT_USE, JUMP_DET"
    This would run outlier detection ignoring both DO_NOT_USE and JUMP_DET flagged pixels.

jdavies-st:
The CR flag should NOT be JUMP_DET.  `outlier_detection` should be setting DO_NOT_USE, and the pick your poison for the informational bit to set.  We don't currently have one.

But once `outlier_detection` is fixed to do this, then `resample` can just use DO_NOT_USE to mask pixels that are not to be used.

hbushouse:
Agreed. The only step that we should consider having a `good_bits` argument is `outlier_detection`. `resample` doesn't need one and should just be hardwired to reject anything with DO_NOT_USE (like other steps).

We also need to figure out a value for the new information flag to be set by `outlier_detection`, but to do that we need to cannibalize one of the existing flags, because we're up against the 32-bit limit and I do NOT want to go to 64 bits.

stscieisenhamer:
Weeeellllll, the current flag is [JUMP_DET](https://github.com/spacetelescope/jwst/blob/2defdd36b38c0d423de66a905cc9410687a4d5f3/jwst/outlier_detection/outlier_detection.py#L20). There is the set of `RESERVED`. Also note that the documenation does discuss using SATURATED, 2, as input to `good-bits`.  This is the "2+4=6" discussion.

`resample` will be modified as needed.

philhodge:
@hbushouse If the user reruns `outlier_detection` with different thresholds, wouldn't they want the rejection to be cumulative?  (For example, using a lower threshold but not flagging neighbors.)  If not, they could rerun the step with the previous input.  By making the flagging cumulative, the user has the choice of starting fresh or accumulating flags.  For HST, in at least some cases we flagged the DQ in the input to the CR rejection function, which is why that function had the option to clear the CR flag in the input before running.

stscieisenhamer:
Though, WRT resample, if the default for resample as "~DO_NOT_USE, <CRFLAG>" , would get the same effect.

hbushouse:
> ## Question
> `skymatch` has a similar parameter, `dqbits`. Not that this PR should handle this, but should `skymatch` be modified to be consistent with `outlier`?

Yes, all steps that have any kind of param like this should be modified to be consistent. At the very least, all steps should reject pixels with DO_NOT_USE, where it's appropriate to do so. And by "appropriate to do so" I mean it's only necessary to reject such pixels if their value is being used in some kind of calculation that affects the output, like computing the stack median for outlier_detection, average signal values in skymatch, etc. Simple steps that are just applying some correction/calculation to the pixels (e.g. dividing by the flat, subtracting the dark, ...) don't need to pay attention to this. If a pixel has a bad value, it'll just have a slightly different bad value in the result.

docs/jwst/outlier_detection/outlier_detection.rst
```diff
@@ -37,8 +37,8 @@ Specifically, this routine performs the following operations:
     save_intermediate_results: specifies whether or not to save any products
                                created during outlier_detection [default=False]
     resample_data: specifies whether or not to resample the input data [default=True]
-    good_bits: Sum of DQ integer values which should be considered good when
-               creating weight and median images [default=6]
+    bad_bits: Sum of DQ integer values which should be considered bad when
```
hbushouse: The purpose of having the flag mnemonics (DO_NOT_USE, JUMP_DET, ...) was so that users wouldn't have to remember numerical values and so that we can make changes to the numerical values without it affecting anything. So it would be better if a param like this (whether it's `good_bits` or `bad_bits`) could be a list of mnemonic names, rather than a numerical value.

jwst/datamodels/dqflags.py
```diff
@@ -1,41 +1,108 @@
...
+# Pixel-specific flags
+pixel = {'GOOD':             0,      # No bits set, all is good
+         'DO_NOT_USE':       2**0,   # Bad pixel. Do not use.
+         'SATURATED':        2**1,   # Pixel saturated during exposure
+         'JUMP_DET':         2**2,   # Jump detected during exposure
+         'DROPOUT':          2**3,   # Data lost in transmission
+         'RESERVED_1':       2**4,   #
+         'RESERVED_2':       2**5,   #
+         'RESERVED_3':       2**6,   #
+         'RESERVED_4':       2**7,   #
```
hbushouse: I suggest that for now we utilize one of the reserved bits and call it "OUTLIER" and have the `outlier_detection` step set that informational flag, as well as "DO_NOT_USE", for the outliers that it detects. That gives us a unique flag from "JUMP_DET".

jwst/outlier_detection/outlier_detection_scaled_step.py
```diff
@@ -31,7 +31,7 @@ class OutlierDetectionScaledStep(Step):
         scale = string(default='0.5 0.4')
         backg = float(default=0.0)
         save_intermediate_results = boolean(default=False)
-        good_bits = integer(default=6)
+        good_bits = string(default="~DO_NOT_USE")  # DQ flags to allow
```
hbushouse: So this presents a real conundrum. If the `outlier_detection` step itself sets both DO_NOT_USE and OUTLIER for pixels that it finds to be outliers, then if a user runs the step again to iterate on something like the rejection threshold, the pixels previously flagged with DO_NOT_USE won't get used in subsequent runs. And we can't implement logic like if a pixel has both DO_NOT_USE and OUTLIER set, assume it's just from a previous run of the step and go ahead and use the pixel, because the DO_NOT_USE may have been set for some other legit reason because the pixel really is bad. In which case, what's the point of having this param at all? Maybe we should consider just giving up on it and if folks want to iterate the step, they'll just have to reset the DQ flags manually before rerunning.

Oh, but wait!  We're not updating the input images in place, are we? The updated DQ flag values only appear in the *output* images that are created. So a user really could just go ahead and rerun the step on the original input images. This is what's different compared to the HST pipeline usage: those updated the inputs in place.

So maybe this default setting (use any pixel that's NOT flagged with DO_NOT_USE) is OK after all and actually achieves what we want!

Never mind ...

jwst/pipeline/outlier_detection.cfg
```diff
@@ -14,4 +14,4 @@ scale = '0.5 0.4'
 backg = 0.0        
 save_intermediate_results = False
 resample_data = True
-good_bits = 6
+good_bits = 'SATURATED, JUMP_DET'
```
hbushouse: We should be consistent with what's in the step spec, in other words set these to ~DO_NOT_USE.

jwst/datamodels/dqflags.py
stscieisenhamer: Sounds like a plan. The first reserved bit will be used.

Just for FYI, as implied earlier when determined what flags were being set, the DO_NOT_USE flag was never set for the marked outliers. This will be fixed also.

So, all discovered outliers will have DO_NOT_USE + OUTLIER set in the DQ array.

stscieisenhamer:
Though I should have done it before the rebase, there is a [regression of this PR](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/44/). The hard errors were all resolved by a rebase. The other errors are directly related to the change in outlier status, most likely due to the fact that DO_NOT_USE is actually being set. The CRF file difference look correct in that there are not any different pixels flagged, but the ones that are have the value `DO_NOT_SET + OUTLIER` instead of `JUMP_DET`. However, this clearly has an effect downstream. Presumption would be that if there were any negative effects, those would be separate issues, but it would be worth a quick check that nothing immediately abhorrent has been done.

docs/jwst/references_general/references_general.rst
```diff
@@ -572,7 +572,7 @@ Bit  Value         Name              Description
 1    2             SATURATED         Pixel saturated during exposure
 2    4             JUMP_DET          Jump detected during exposure
 3    8             DROPOUT           Data lost in transmission
-4    16            RESERVED
+4    16            OUTLIER           Removed by outlier detection
```
hbushouse: Minor nit, but technically the pixel isn't being removed by `outlier_detection`, it's only being flagged. It's then up to other steps to decide whether or not to remove it. So perhaps "Flagged by outlier detection".

docs/jwst/skymatch/README.rst
```diff
@@ -137,6 +137,13 @@ The ``skymatch`` step has the following optional arguments:
   final "good" bits. For example, both ``4,8`` and ``4+8``
   are equivalent to setting ``dqbits`` to 12.
 
+  Finally, instead of integers, the JWST mnemonics for the bitflags may
+  be used. For example, all the following specifications are equivalent:
+
+  `12 == 4 + 8 == "4, 8" == "JUMP_DET, DROPOUT"`
```
hbushouse: Cool - so the user can specify it either as numbers or mnemonics? Fancy!

hbushouse: So it looks like the only doc updates included here are for `skymatch`. What about `outlier_detection` and/or `resample`?

jwst/pipeline/outlier_detection.cfg
hbushouse: This isn't consistent with what's in the step code "spec" block.

jwst/pipeline/outlier_detection_scaled.cfg
```diff
@@ -13,4 +13,4 @@ snr = '4.0 3.0'
 scale = '0.5 0.4'
 backg = 0.0        
 save_intermediate_results = False
-good_bits = 6
+good_bits = 'SATURATED, JUMP_DET'
```
hbushouse: ditto

jwst/pipeline/outlier_detection_stack.cfg
```diff
@@ -14,4 +14,4 @@ scale = '0.5 0.4'
 backg = 0.0        
 save_intermediate_results = False
 resample_data = False
-good_bits = 6
+good_bits = 'SATURATED, JUMP_DET'
```
hbushouse: ditto

jwst/pipeline/outlier_detection.cfg
stscieisenhamer: DIdn't change this behavior to reduce the number of test failures; there would be enough from the regressions. As such, being "unit" tests, and being hard-wired, was not clear that the intent would not have been to set explicitly. Regardless, if the real need is to be "default", it can be made so.

stscieisenhamer:
## Question: DQ_DEF extension

When/where is this created? Is this during Level1 creation, or is this something we have control over.

hbushouse:
Currently the DQ_DEF extension only appears in CRDS reference files and is applied to the ref data when they're loaded into a datamodel (that allows the DQ extension in the ref files to use only a few bits and hence reduce data volume). It's not used or created at all for science products.

jwst/lib/suffix.py
```diff
@@ -297,8 +297,8 @@ def find_suffixes():
                 step = Step.from_config_file(
                     path.join(config_path, config_file)
                 )
-            except Exception:
-                pass
+            except Exception as err:
```
jdavies-st: Can we be more specific with the exception type here?  What exceptions does `from_config_file` raise?

jwst/lib/suffix.py
```diff
@@ -335,12 +335,12 @@ def load_local_pkg(fpath):
             module_path = '.'.join(module_path)
             try:
                 module = import_module(module_path)
-            except Exception:
-                logging.debug('Cannot load module "{}"'.format(module_path))
+            except Exception as err:
```
jdavies-st: Instead of `Exception`, it would be better to have `ImportError` here?

jwst/lib/tests/test_suffix.py
```diff
@@ -1,12 +1,17 @@
 """Test suffix replacement"""
 
+import logging
 import pytest
 
 from .. import suffix as s
 
+s_logger = logging.getLogger(s.__name__)
+s_logger.addHandler(logging.StreamHandler())
```
jdavies-st: What is the purpose of adding a logger and handler to the test module?  Is this to capture log messages?  Tests shouldn't be logging anything.

jwst/outlier_detection/outlier_detection_scaled_step.py
jdavies-st: Actually, we don't even need `good_bits` at all.  We are not modifying images in place.  We are always starting with input `_cal` files and generating `_crf` files.  I think `good_bits` is just confusing and not a desirable workflow.

jwst/pipeline/tests/data/cfgs/outlier_detection.cfg
```diff
@@ -14,4 +14,4 @@ scale = '0.5 0.4'
 backg = 0.0        
 save_intermediate_results = False
 resample_data = True
-good_bits = 4
+good_bits = 'JUMP_DET'
```
jdavies-st: What are all these config files doing in `jwst/pipeline/tests/data/cfgs/`?  This seems like a recipe for being out-of-date?

jwst/outlier_detection/outlier_detection_scaled_step.py
jdavies-st: Actually, we don't even need `good_bits` at all.  We are not modifying images in place.  We are always starting with input `_cal` files and generating `_crf` files.  I think `good_bits` is just confusing and not a desirable workflow.

jdavies-st: I suspect this was implemented for HST because all the instruments flag their DQ bits differently.  This is not the case with JWST.

jwst/lib/suffix.py
stscieisenhamer: Well, logically, this routine doesn't particularly care what the error is, simply that, for whatever reason, a step cannot be "bare instantiated".

That being said, and see more about the logging comment below, I am going to be filing a couple of issues that can do away with I believe most of the errors. So far, most of the issues are with defunct/missing/mis-spelled configs.

jwst/lib/tests/test_suffix.py
stscieisenhamer: The test itself is not logging anything. However, having the module-being-tested logging enabled does provide more info if a test failure occurs.  In this case, it has pointed out some other unrelated issues to be filed.

But, glad this was pointed out, because what the final form of this test was to also turn off the logging once tests are done. Next on the list.

jwst/pipeline/tests/data/cfgs/outlier_detection.cfg
stscieisenhamer: yep. should be removed once all configs have been removed/simplified.

jwst/lib/suffix.py
jdavies-st: Yeah, that makes sense.  That's exactly my concern.  By just blanket handling all exceptions, things get swept under the rug.  We want it to fail for reasons that we don't expect (missing or misspelled configs, for instance).  But another PR is good.  👍

jwst/outlier_detection/outlier_detection_scaled_step.py
stscieisenhamer: @hbushouse Just to clarify before merging, with regard to this thread, the `good_bits` parameter will be left in place?

hbushouse: It can't hurt to leave it in, at least for now. In the default mode it's at least doing the right thing now, in terms of rejecting pixels that have DO_NOT_USE in the input and also setting DO_NOT_USE in the output (for use downstream). It's likely that we don't really need it, because we aren't modifying the inputs in place, but it can't hurt to have the flexibility available (it's amazing what use cases folks will come up with that we never thought of).

jdavies-st:
Does this PR resolve #4337?

stscieisenhamer:
Yes, this should. Result should now be both DO_NOT_USE and the new OUTLIER flags should be set.

mcara:
Why was parameter `good_bits` of the ResampleStep removed and DQ flags hardcoded? 

CC: @hbushouse

hbushouse:
> Why was parameter `good_bits` of the ResampleStep removed and DQ flags hardcoded?

Not easy to remember the logic behind decisions made 3+ years ago, but as I recall there were 2 reasons:
1) The concept of specifying which DQ flags are considered "good" is backwards from the way all other JWST pipeline steps handle DQ flags, where the logic is reversed, in that it's the flags that are considered *bad* that are specified and used in rejecting pixels.
2) The overall philosophy within all JWST pipeline steps is that all DQ flags are considered to be "information only", i.e. here are some conditions that this pixel has, *except* for the DO_NOT_USE flag (bit 1). All decisions within all steps concerning which pixels to reject are based on whether or not DO_NOT_USE is set, period. No waffling or ability to bend the rules. If DO_NOT_USE is set, reject it; if not, keep it.

PR #4730, opened 2020-03-26
Jp 1384

dmggh:
GLS crashes for multiple integration datasets. See Git Hub PR#4727

### DIFF
--- a/jwst/jump/tests/test_detect_jumps.py
+++ b/jwst/jump/tests/test_detect_jumps.py
@@ -201,7 +201,7 @@ def test_twoints_onecr_10_groups_neighbors_flagged_multi(setup_inputs):
     """"
     A multiprocessing test that has two CRs on the boundary between two slices
     in different integrations. This makes sure that we are correctly flagging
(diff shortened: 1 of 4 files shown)


jwst/jump/tests/test_detect_jumps.py
```diff
@@ -447,8 +445,158 @@ def test_onecr_50_groups(setup_inputs):
...
+    grouptime = 3.0
+    ingain = 5
+    inreadnoise = np.float64(7)
+    ngroups = 10
+
+    model1, gdq, rnModel, pixdq, err, gain = \
+        setup_inputs( ngroups=ngroups, nrows=5, ncols=6, gain=ingain, readnoise=inreadnoise,
+                      deltatime=grouptime )
+
+    # two segments perfect fit, second segment has twice the slope
```
philhodge: What is the reason for having twice the slope after the jump?  I realize that's done in the other tests in test_detect_jumps.py, but I don't see the point.

dmggh: This test is Mike Regan's; I also don't understand the point and will look into this.

PR #4731, opened 2020-03-26
JP-1342 Fix unclosed file warnings

### LINKED ISSUE #4662: Unclosed files error from some pipeline steps
Issue [JP-1342](https://jira.stsci.edu/browse/JP-1342) was created by Howard Bushouse:

Python is reporting "unclosed file" errors after a few different steps shutdown. This includes the uncal.fits file after dq_init ends and the rateints.fits file after the calspec2 assign_wcs step ends. See the near-final comments in JP-1174 for more details and examples.

jwhite3:
This PR resolves `ResourceWarning`s triggered by unclosed files over the course of some pipeline runs. A call to `closeout` was added to the end of `Step.run` to ensure that inputs were properly closed. In addition, a simple `close` method was added to `ModelContainer` in order to close models in the container. 

Resolves #4662 / [JP-1342](https://jira.stsci.edu/browse/JP-1342)

### DIFF
--- a/jwst/stpipe/tests/test_step.py
+++ b/jwst/stpipe/tests/test_step.py
@@ -86,6 +86,8 @@ def test_parameters_from_crds_fail():
     pars = RefPixStep.get_config_from_reference(data)
     assert not len(pars)
 
+    data.close()
+
 
 @pytest.mark.parametrize(
     'cfg_file, expected',
@@ -110,7 +112,7 @@ def test_saving_pars(tmpdir):
     """Save the step parameters from the commandline"""
     cfg_path = t_path(join('steps', 'jwst_generic_pars-makeliststep_0002.asdf'))
     saved_path = tmpdir.join('savepars.asdf')
-    Step.from_cmdline([
+    step = Step.from_cmdline([
         cfg_path,
         '--save-parameters',
         str(saved_path)
@@ -118,6 +120,9 @@ def test_saving_pars(tmpdir):
     assert saved_path.check()
     saved = datamodels.StepParsModel(str(saved_path))
     assert saved.parameters == ParsModelWithPar3.parameters
+
+    step.closeout()
+    saved.close()
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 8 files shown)


jwhite3:
Not actually sure if it's a good idea for now to include the `open-files` option for pytest... I think quite a lot of tests would start failing

jwst/stpipe/tests/test_step.py
```diff
@@ -119,6 +121,9 @@ def test_saving_pars(tmpdir):
     saved = datamodels.StepParsModel(str(saved_path))
     assert saved.parameters == ParsModelWithPar3.parameters
 
+    step.closeout()
+    saved.close()
```
jdavies-st: Instead of `saved.close()` I would use a `with` context manager:

```python
with datamodels.StepParsModel(str(saved_path)) as saved:
    assert saved.parameters == ParsModelWithPar3.parameters
```

jwhite3:
So this solution only sort of works.. 

In thinking about how I was "fixing" the tests, I realized that that was dumb.. and that my solution wasn't actually having the intended result in all cases. 

This solution works in cases where the input for a `Step` has some sort of close method attached, but it does nothing in the case where an input is a path and a model is generated in `process` from that path. 

So, for example:

```python
Step.from_cmdline(args)  # args includes a path to an input rather than an input model
```

Should not require a call to `closeout` if the fix in `run` was doing it's job in all cases

jdavies-st:
Generally, the step itself should be using the `with` context manager any time it is opening files.  So yeah, probably fix it there instead of in the test if possible.

jdavies-st:
I'll add that very few of our steps or pipelines actually use `Step.open_model`, because that was added quite late in the game.  Could you remind us why it is preferable to use that instead of a simple `datamodels.open` at the beginning of each step or pipeline?

stscieisenhamer:
To abstract the `Step` I/O conventions. In this case, it is to satisfy honoring the `Step.input_dir` parameter.

jwhite3:
Pushed the discussed changes. 

For now, the plan is to open additional PRs for:
- Adding `open-files` option to the pytest configuration
- Refactoring all `jwst` `Step` and `Pipeline` `process` methods to use `Step.open_model` within a context manager (if possible) or at least close the data models explicitly when `process` opens an initial data model from input

jwhite3:
This PR is expanded further with #4740

PR #4732, opened 2020-03-27
JP-1334: Make MRS bg matching use EMSM weighting, apply sigma clipping

### LINKED ISSUE #4637: MRS imatch incorrectly deriving gradient
Issue [JP-1334](https://jira.stsci.edu/browse/JP-1334) was created by David Law:

Newly-simulated MRS data produced for end-to-end testing shows some bug in mrs_imatch in quick test reductions.

Specifically, with a simulated 2-pt dither and two-position mosaic, an imatched Ch2A cube shows a strong background gradient across the field of view that is not present in cubes build from individual exposures.  See attached illustration, for which all images are presented at the same wavelength and with the same greyscale stretch.

I have not yet had a chance to dig into this issue further, but wanted to file a ticket while I remember before going offline for a few days.  At a guess there may be some artifact in one or more frames (possible from incomplete cosmic ray detection, or unmasked detector artifacts) that is biasing the solution. !imatch.png!

drlaw1558:
This pull request fixes https://jira.stsci.edu/browse/JP-1334 regarding a bug in background matching for the MIRI MRS.  As per that ticket, the root problem is the with the small MRS FOV and the spatial undersampling of the instrument, point sources in the overlap region between adjacent exposures dramatically pulls off the derived background levels for the matching algorithm.

The fix here implements a sigma-clipped rejection of sources prior to passing the input cubes to the stock background solver (albeit possibly fixing the problem using the proverbial sledgehammer, for which performance could be optimized).

An additional fix makes MRS background matching use the 'EMSM' weighting function instead of the 'MSM' function for consistency with recent changes to the cube building and outlier rejection algorithms.

Fixes #4637 / [JP-1334](https://jira.stsci.edu/browse/JP-1334)


jdavies-st:
The failing test here is unrelated to the PR.  It is caused by a bugfix to `gwcs` released yesterday.

https://github.com/spacetelescope/gwcs/releases/tag/0.13.0

This should be fixed in a separate PR.  @nden

hbushouse:
Please add an entry to "CHANGES.rst" to document this update.

drlaw1558:
Updated CHANGES.rst

jwst/mrs_imatch/mrs_imatch_step.py
```diff
@@ -259,6 +259,7 @@ def _match_models(models, channel, degree, center=None, center_cs='image'):
     cbs.channel = str(channel)
     cbs.band = 'ALL'
     cbs.single = True
+    cbs.weighting = 'EMSM'
```
mcara: What does this do?

drlaw1558: This forces the cube building routine to use the EMSM instead of (currently) default MSM weighting scheme.  This is an exponential radial weight function instead of a 1/r that a previous ticket/PR demonstrated achieved better performance.  At some point Jane and I planned to make the default cube building for MRS use EMSM, but in the meantime this will ensure that background matching uses cubes with the fewest artifacts.

jwst/mrs_imatch/mrs_imatch_step.py
```diff
@@ -328,6 +329,31 @@ def _match_models(models, channel, degree, center=None, center_cs='image'):
...
+    # MRS fields of view are small compared to source sizes,
+    # and undersampling produces significant differences
+    # in overlap regions between exposures.
+    # Therefore use sigma-clipping to detect and remove sources
+    # (and unmasked outliers) prior to doing the background matching.
+    # Loop over input exposures
+    for i in range(0,len(models)):
+        image = image_data[i]
+        mask = mask_data[i]
+        weight = sigma_data[i]
```
mcara: I believe this could be written as:
```suggestion
    for image, mask, weight in zip(image_data, mask_data, sigma_data):
```

jwst/mrs_imatch/mrs_imatch_step.py
```diff
@@ -328,6 +329,31 @@ def _match_models(models, channel, degree, center=None, center_cs='image'):
...
+        # Do statistics wavelength by wavelength
+        for j in range(0,len(image)):
+            thisimg = image[j,:,:]
+            thismask = mask[j,:,:]
+            thiswght = weight[j,:,:]
```
mcara: Same could be done here:
```suggestion
        for thisimg, thismask, thiswght in zip(image, mask, weight):
```

jwst/mrs_imatch/mrs_imatch_step.py
```diff
@@ -328,6 +329,31 @@ def _match_models(models, channel, degree, center=None, center_cs='image'):
...
+            # Sigma clipped statistics, ignoring zeros where no data
+            _, themed, clipsig = sigclip(thisimg, mask_value=0.)
+            im1d = thisimg.ravel()
+            mask1d = thismask.ravel()
+            wght1d = thiswght.ravel()
+            # Reject beyond 3sigma
+            reject = np.where(np.abs(im1d - themed) > 3*clipsig)
+            im1d[reject] = themed
+            mask1d[reject] = 0
+            wght1d[reject] = 1e18
```
mcara: Is `ravel()` really necessary here? It seems to me this code should work as is

mcara: I think masking should be enough (data and weights are ignored if mask is 0):
```suggestion
            mask1d[reject] = 0
```

jwst/mrs_imatch/mrs_imatch_step.py
```diff
@@ -328,6 +329,31 @@ def _match_models(models, channel, degree, center=None, center_cs='image'):
...
+            thisimg = image[j,:,:]
+            thismask = mask[j,:,:]
+            thiswght = weight[j,:,:]
+            # Sigma clipped statistics, ignoring zeros where no data
+            _, themed, clipsig = sigclip(thisimg, mask_value=0.)
+            im1d = thisimg.ravel()
+            mask1d = thismask.ravel()
+            wght1d = thiswght.ravel()
+            # Reject beyond 3sigma
+            reject = np.where(np.abs(im1d - themed) > 3*clipsig)
```
mcara: I think it is important to have the value of "number of sigmas" (3 in this case) be settable for the step. For example, add a new parameter `nsig(ma)`. But, if you think that for now it would be better not to introduce new parameters to the step, then leave it as is.

drlaw1558: I'd been thinking about whether that was worthwhile doing now or delaying until sometime later; I'll test that out along with your other suggestions above (was sure you'd have some thoughts on how to make my clunky code more elegant!)

mcara: You know, maybe it is better not to do this now. Let's wait until there is need for this. It is more difficult to remove a parameter than to add one. Also add spaces around `*` in `3 * clipsig`.

drlaw1558:
Ok, I've integrated your suggested changes, and added some info to the documentation that this clipping is being done.

mcara:
Are there any unit/regression tests for this case @jdavies-st @hbushouse ?

hbushouse:
> Are there any unit/regression tests for this case @jdavies-st @hbushouse ?

No units tests, but there is 1 regression test in https://github.com/spacetelescope/jwst/blob/master/jwst/regtest/test_miri_mrs.py

The test "spec3_multi" runs the `calwebb_spec3` pipeline, including the `mrs_imatch` step on an association of several MRS exposures. I don't know off-hand the origin of the input data or how realistic it is.

drlaw1558:
@hbushouse Can you point me to the documentation site for how to get data out of artifactory?  This would be good for me to know anyway, and then I can check what truth data is being used for these tests at the moment.

hbushouse:
@jdavies-st do we have any documentation on artifactory use that @drlaw1558 can use? I can easily point you to the actual url, but someone probably needs to give you access first. Is that @rendinam ?

jdavies-st:
@drlaw1558 to get the data, just run the regression test.  It will download the data needed into a tmpdir and then run the pipeline and run the test on the pipeline results.

Instructions are in the repo README for running regression tests locally.  But from within your git checkout of the `jwst` repo at the top level (after a dev installation):

```
pytest --bigdata -v --basetemp=/path/to/scratchdir jwst/regtest/test_miri_mrs.py::spec3_multi
```

CHANGES.rst
```diff
@@ -37,6 +37,12 @@ master_background
 
 - Updated step arguments in the documentation. [#4723]
 
+mrs_imatch
+-----------------
```
pllim, line 41: Underline length must match number of characters in text.

jwst/mrs_imatch/mrs_imatch_step.py
```diff
@@ -328,6 +329,20 @@ def _match_models(models, channel, degree, center=None, center_cs='image'):
...
+    # Therefore use sigma-clipping to detect and remove sources
+    # (and unmasked outliers) prior to doing the background matching.
+    # Loop over input exposures
+    for image, mask in zip(image_data, mask_data):
+        # Do statistics wavelength by wavelength
+        for thisimg, thismask in zip(image, mask):
+            # Sigma clipped statistics, ignoring zeros where no data
+            _, themed, clipsig = sigclip(thisimg, mask_value=0.)
+            # Reject beyond 3 sigma
+            reject = np.where(np.abs(thisimg - themed) > 3 * clipsig)
```
pllim, line 344: Is `where` really needed?
```suggestion
            reject =np.abs(thisimg - themed) > 3 * clipsig
```

mcara: It is not _required_ for the logic to function but it may help with performance.For example, if you just leave `reject` as a full boolean array, then when you do something like `data[reject] = 0` it will have to go through all elements of the `data` array. On the other hand, if only a small number of elements are "rejected", then the length of the indices produced by `np.where` will be small and so `data[reject]` will be much faster because it would access only a small number of elements in the `data` array.

mcara: So, _it depends_ on whether you expect a large number of elements to be rejected (then do not use `np.where`) or a small number (do use `np.where`).

pllim: Yes, that makes sense. But `np.where` also has its own overhead but maybe insignificant for your use case. Probably premature optimization anyway, so it is fine to leave as is. Thanks for the clarification!

CHANGES.rst
drlaw1558: Done

hbushouse: I *think* it's the case that it's OK to have the underline be *longer* than the corresponding text, but not *shorter*. Of course it doesn't hurt to always make them match, just in case.

hbushouse:
It appears all serious issues have been addressed, including the conflict in the change log, and the CI tests have now passed. Is there anything else that needs to be done before merging or can we merge now?

mcara:
LGTM

drlaw1558:
I think it's set from my point of view; when I get a chance I'd like to follow up on the overall MRS regression test data conversation (thanks for the info!) but I think that's separable.

PR #4734, opened 2020-03-27
fix test - should compare 0-based coordinates

nden:
The 1 deg difference in the failing test is due to CDELT=1.


nden:
As expected the updated test now fails with the old version of gwcs in `requirements_sdp`.

jdavies-st:
We will have to eventually bump to `gwcs>=0.13` in `install_requires` so this test will always pass.

PR #4740, opened 2020-03-28
Resolve open file issues in unit tests

### LINKED ISSUE #4737: Add `open-files` option to default `pytest` configuration
Issue [JP-1387](https://jira.stsci.edu/browse/JP-1387) was created by James White:

Enabling the `pytest` plugin, `open-files` by default in the configuration file would enable checking for opened files throughout testing. 

 

This would help prevent the same issue from reemerging in the future, and will help with making sure that various stages of the Pipelines or Steps are properly handling files.

jdavies-st:
This PR is for a branch started a year ago.  I've rebased it and added a few fixes to it, so it is complimentary to #4731 I think.  @jwhite3 have a look.  I don't think it solves the regression test pipeline issues, but it does resolve the unit test issues and allows turning on the `pytest-openfiles` plugin.

Resolves #4737

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,5 @@ scripts/launch.json
 *.orig
 .ipynb_checkpoints
 */.ipynb_checkpoints/*.ipynb
+.coverage
+htmlcov
--- a/.gitignore
+++ b/.gitignore
@@ -33,5 +33,4 @@ scripts/launch.json
 *.orig
 .ipynb_checkpoints
 */.ipynb_checkpoints/*.ipynb
-.coverage
 htmlcov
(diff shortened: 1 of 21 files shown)


jwhite3:
@jdavies-st you're right, this is definitely a good expansion of the small set of changes in #4731 

Next step after this would be refactoring all implemented steps and pipelines to use `datamodels.open` within a context manager or close explicitly, and then hopefully we should see the `ResourceWarnings` go away

jdavies-st:
I've rebased this to your merged PR #4731 and have cleaned up some of the test instances of `Step` used so that they use a context manager properly.  This now allows us to get rid of lots of `model.close()` and `step.closeout()` calls in our tests.

At some point we should move all those example steps used for testing out of the `jwst/stpipe/tests/steps/__init__.py` module.  😬

PR #4741, opened 2020-03-28
Remove requirements from requirements-sdp.txt

jdavies-st:
This should get our regression tests and unit tests passing.  Both need `gwcs>=0.13`.

Recall that one of the builds in our TravisCI per-PR unit test runs uses `requirements-sdp.txt`.


jdavies-st:
We should repopulate the `requirements-sdp.txt` in a week or two once the versions of our main dependencies for build 7.5 are known.  These will include new versions of `gwcs`, `astropy`, `asdf` and `tweakwcs` at a minimum.

nden:
Do you need to change `JenkinsRT` to not use requirements-sdp?

jdavies-st:
No, we still use `-r requirements-sdp.txt` in the Jenkinsfiles, but because it is empty, nothing happens. 👍

PR #4745, opened 2020-03-30
Update readme.rd

mwregan2:
This PR attempts to clear up some ambiguities in the README.md file. The creation of the conda environment is separated out from the actual installation. The descriptions of how to install the three different types of pipelines are made more consistent. Also, CRDS is explained a little more.


README.md
```diff
@@ -13,18 +13,26 @@ JWST Calibration Pipeline
...
+    ./conda activate <env_name>
+
+## Below are directions on how to install three different versions of the pipeline.
+Note that you need to always create and activate your new evironment
+before proceeding.
 
 ### Installing for end-users ###
 
-To install a released (tagged) version, you can install directly from Github.  To install release `jwst 0.15.0`:
+If you want to run the latest fully tested version of the pipeline you should install a released (tagged) version. you can install these versions directly from Github.  For example, to install release `jwst 0.15.0`:
```
hbushouse, line 31: Minor nit, but the latest tagged version is 0.15.1 (corresponding to B7.4.2), so I'd change 0.15.0 to 0.15.1

README.md
```diff
@@ -36,22 +44,20 @@ To include extra testing dependencies, the syntax is as follows:
 
     pip install "jwst[test] @ git+https://github.com/spacetelescope/jwst"
 
-### Installing a DMS release ###
+### Installing a Data Management System (DMS) release ###
 
-We still package our releases to DMS via environment snapshots that specify the exact versions of all packages to be installed.
+We still package our releases to DMS via environment snapshots that specify the exact versions of all packages to be installed. This method should be more stable than the frist method because the latest versions of packages may cause problems.
```
hbushouse, line 49: typo: "frist" should be "first"

README.md
```diff
@@ -36,22 +44,20 @@ To include extra testing dependencies, the syntax is as follows:
...
 
 The latest release 0.15.0 may be installed by running the following commands in Linux or MacOS:
 
 Linux:
 
-    conda create -n jwstdp-0.15.0 --file https://ssb.stsci.edu/releases/jwstdp/0.15.0/conda_python_stable-deps.txt
-    conda activate jwstdp-0.15.0
+    conda install --file https://ssb.stsci.edu/releases/jwstdp/0.15.0/conda_python_stable-deps.txt
```
hbushouse, line 55: I hope it remains clear to someone doing this that they should first create a specific conda environment in which they want to install this particular version of the package. Without creating and activating a particular environment, it's just going to get installed in whatever environment is currently active or (if no environment is active) in their "base" environment.

README.md
```diff
@@ -13,18 +13,26 @@ JWST Calibration Pipeline
...
 ------------
 
-The `jwst` package can be installed into a virtualenv or conda environment via `pip`.  We recommend creating a fresh environment with only python installed.  If using conda, make sure you have a recent version of Anaconda or Miniconda installed.  Via conda:
+The `jwst` package can be installed into a virtualenv or conda environment via `pip`.  In general, we recommend creating a fresh environment with only python installed.  If using conda, make sure you have a recent version of Anaconda or Miniconda installed.  From a bash shell enter:
 
-    conda create -n jwst_env python
-    conda activate jwst_env
+    conda create -n <env_name> python
+    conda activate <env_name>
+If conda is not in your path, you will need to cd to the bin directory of your anaconda installation and enter:
```
stscieisenhamer, line 20: Always a bit leery of situations where "its not in the path": For general users, if that is so, then something is probably not right and there will be more issues later on.

jdavies-st: Agree.  Conda better be in your path.  And you need to fix your conda installation.  That is outside the scope of this document.

README.md
jdavies-st: I don't like this wording.  Latest versions of packages should not cause problems, because we test against the latest versions of packages, and generally latest patch versions of packages will have bugfixes, so it should not be an issue.

If you're installing a very old version of the pipeline, then using one of these DMS snapshots *might* be preferable for stability reasons, but is probably not preferred from a performance and bugfix standpoint.  It really depends on the dependency.

README.md
jdavies-st: We don't not want to make this change to the instructions.  We specifically recommend doing a `conda create` here so that the DMS environment is isolated.  We don't know what exists in an existing env and whether there will be conflicts with what is in the DMS build env.

README.md
```diff
@@ -36,22 +44,20 @@ To include extra testing dependencies, the syntax is as follows:
...
-    conda create -n jwstdp-0.15.0 --file https://ssb.stsci.edu/releases/jwstdp/0.15.0/conda_python_stable-deps.txt
-    conda activate jwstdp-0.15.0
+    conda install --file https://ssb.stsci.edu/releases/jwstdp/0.15.0/conda_python_stable-deps.txt
     pip install -r https://ssb.stsci.edu/releases/jwstdp/0.15.0/reqs_stable-deps.txt
 
 MacOS:
 
-    conda create -n jwstdp-0.15.0 --file https://ssb.stsci.edu/releases/jwstdp/0.15.0/conda_python_macos-stable-deps.txt
-    conda activate jwstdp-0.15.0
+    conda install --file https://ssb.stsci.edu/releases/jwstdp/0.15.0/conda_python_macos-stable-deps.txt
```
jdavies-st, line 60: Again, no.

README.md
mwregan2: I changed it to 0.15.1

README.md
mwregan2: @jdavies-st I was trying to figure out why you would choose option 2 over option 1. We have a general problem on the project of giving users options but not describing why you would use one option over the other. 
  So, should it say to choose the DMS option only if you are trying to run a non-current version?

README.md
mwregan2: The problem is that the current instructions create a new conda env for the DMS installation but not for the "end user" installation or the developer installation. This lack of symmetry confused me. The revised instruction create the env at the top and then say to choose one of the three methods. I'll try adding something to make it clearer.

README.md
mwregan2: Well, conda has never been in my path.
I'm running a bog-standard bash shell setup by ITSD. I can't be the only one without conda in their path.

README.md
mwregan2: My first draft put the env creation and activation at the start of each option. I removed it because it seemed redundant. I can go back to that, if people don't mind the redundancy. I do think the three options should be symmetrical.

README.md
```diff
@@ -13,18 +13,27 @@ JWST Calibration Pipeline
...
+The `jwst` package can be installed into a virtualenv or conda environment via `pip`.  In general, we recommend creating a fresh environment with only python installed.  If using conda, make sure you have a recent version of Anaconda or Miniconda installed.  From a bash shell enter:
+## Step 1. Create a Conda Environment ##
 
-    conda create -n jwst_env python
-    conda activate jwst_env
+    conda create -n <env_name> python
+    conda activate <env_name>
 
-### Installing for end-users ###
+## Below are directions on how to install three different versions of the pipeline.
```
jdavies-st, line 22: Change to "directions for 3 methods to install the pipeline".  All install the same version (same code) of the pipeline.

README.md
jdavies-st: Symmetry is not a good enough argument for changing this.  We are giving here the *exact* same installation procedure that DMS uses to install it in their systems.  That is the purpose of having this method of installation available here - to debug DMS issues.  End users, INS testers can use this method if they like but it is not necessary.

mwregan2: @jdavies-st The problem is that the instructions are not clear. It is not just me that thinks this is true. Others have commented on it too. I had to read it several times to figure out that the conda env created at the top needs to be used for the 1st and 3rd methods of installing but not for the second method. I've updated the new version to repeat the conda env creation and activation at the start of each of the three methods. 
So, should the DMS installation instructions say this is only for DMS?

hbushouse:
Superseded by #4779. Closing.

PR #4750, opened 2020-03-31
add pupil to filteroffset model and schema

### LINKED ISSUE #4739: Update filteroffset schema and model to allow for pupil as a selector
Issue [JP-1388](https://jira.stsci.edu/browse/JP-1388) was created by Nadia Dencheva:

Add "pupil" to filteroffset schema and model to accommodate new NIRISS and NIRCAM reference files.

nden:
Resolves #4739

### DIFF
--- a/jwst/datamodels/flat.py
+++ b/jwst/datamodels/flat.py
@@ -31,5 +31,4 @@ class FlatModel(ReferenceFileModel):
         self.dq = dynamic_mask(self)
 
         # Implicitly create arrays
-        self.dq = self.dq
         self.err = self.err
--- a/jwst/datamodels/reference.py
+++ b/jwst/datamodels/reference.py
@@ -1,6 +1,4 @@
 import warnings
-import sys
-import traceback
 
 from .model_base import DataModel
 from .dynamicdq import dynamic_mask
@@ -26,24 +24,24 @@ class ReferenceFileModel(DataModel):
         Convenience function to be run when files are created.
         Checks that required reference file keywords are set.
         """
-        try:
-            assert self.meta.description is not None
-            assert self.meta.telescope == 'JWST'
-            assert self.meta.reftype is not None
-            assert self.meta.author is not None
-            assert self.meta.pedigree is not None
-            assert self.meta.useafter is not None
-            assert self.meta.instrument.name is not None
-        except AssertionError:
-            if self._strict_validation:
-                raise
-            else:
-                tb = sys.exc_info()[-1]
-                tb_info = traceback.extract_tb(tb)
-                text = tb_info[-1][-1]
-                warnings.warn(text, ValidationWarning)
+        to_fix = []
+        to_check = ['description', 'reftype', 'author', 'pedigree', 'useafter']
+        for field in to_check:
+            if getattr(self.meta, field) is None:
+                to_fix.append(field)
+        if self.meta.instrument.name is None:
+            to_fix.append('instrument.name')
+        if self.meta.telescope != 'JWST':
+            to_fix.append('telescope')
+        if to_fix:
+            self.print_err(f'Model.meta is missing values for {to_fix}')
+        super().validate()
 
-        super(ReferenceFileModel, self).validate()
+    def print_err(self, message):
+        if self._strict_validation:
+            raise ValueError(message)
+        else:
+            warnings.warn(message, ValidationWarning)
 
 
 class ReferenceImageModel(ReferenceFileModel):
--- a/jwst/datamodels/schemas/filteroffset.schema.yaml
+++ b/jwst/datamodels/schemas/filteroffset.schema.yaml
@@ -8,6 +8,8 @@ allOf:
 - $ref: keyword_exptype.schema
 - $ref: keyword_filter.schema
 - $ref: keyword_pfilter.schema
+- $ref: keyword_channel.schema
+- $ref: keyword_module.schema
 - type: object
   properties:
     meta:
--- a/jwst/datamodels/schemas/filteroffset.schema.yaml
+++ b/jwst/datamodels/schemas/filteroffset.schema.yaml
@@ -20,13 +20,34 @@ allOf:
             - type: string
             - $ref: http://stsci.edu/schemas/asdf/unit/unit-1.0.0
     filters:
-      type: array
-      items:
-        type: object
-        properties:
-          name:
-            type: string
-          column_offset:
-            type: number
-          row_offset:
-            type: number
+      description: |
+        An array of dictionaries where either the
+        "name" key or the combination of "filter" and "pupil"
+        are matched with the SCI exposure metadata to get
+        values for "row_offset" and "column_offset".
+
+        The option with the "name" key is used for
+        MIRI data.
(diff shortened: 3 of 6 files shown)


jwst/datamodels/flat.py
```diff
@@ -31,5 +31,4 @@ def __init__(self, init=None, **kwargs):
         self.dq = dynamic_mask(self)
 
         # Implicitly create arrays
-        self.dq = self.dq
```
hbushouse: What does this change have to do with the filteroffset model?

jwst/datamodels/reference.py
```diff
@@ -26,24 +26,24 @@ def validate(self):
...
-        super(ReferenceFileModel, self).validate()
+        to_fix = []
+        to_check = ['description', 'reftype', 'author', 'pedigree', 'useafter']
+        for field in to_check:
+            if getattr(self.meta, field) is None:
+                to_fix.append(field)
+        if self.meta.instrument.name is None:
+            to_fix.append('instrument.name')
+        if self.meta.telescope != 'JWST':
+            to.fix.append('telescope')
```
hbushouse: typo: "to.fix" should be "to_fix"

jwst/datamodels/schemas/filteroffset.schema.yaml
```diff
@@ -30,3 +30,5 @@ allOf:
             type: number
           row_offset:
             type: number
+          pupil:
```
hbushouse: Now that there are entries for both filter and pupil, would it make sense to rename the existing "filters" attribute to something more generic and change "name" to "filter"?

hbushouse: And how does this addition affect the use of this model for MIRI, which only uses filter? Will they have to redeliver their ref file(s) to contain blank/dummy entries for "pupil"? Or is "pupil" somehow an optional attribute?

jwst/datamodels/flat.py
nden: Nothing. I looked at the file because of a naming conflict (fixed since then) and noticed that `self.dq` is defined twice, effectively overwriting the output of `dynamic_mask`. If you prefer I will put it back.

BTW those two statements (claiming to create arrays implicitly)  were removed in `image.py`. I'd like to understand the logic behind these statements and when to keep them.

jwst/datamodels/reference.py
nden: good catch, thanks!

jwst/datamodels/schemas/filteroffset.schema.yaml
nden: If we don't change `name` to `filter` adding `pupil` is transparent to the MIRI team.

nden: I take this back, we can allow both ways if the schema is something like:
```
    filters:
      anyOf:
        - type: array
          items:
            type: object
            properties:
              name:
                type: string
              column_offset:
                type: number
              row_offset:
                type: number
        - type:
          items:
            type: object
            properties:
              filter:
                type: string
              column_offset:
                type: number
              row_offset:
                type: number
              pupil:
                  type: string

```

which will produce files like
```
filters:
- {column_offset: 1, filter_name: F1, pupil: P1, row_offset: 1}
- {column_offset: 2, filter_name: F2, pupil: P2, row_offset: 2}
- {column_offset: 3, filter_name: F3, pupil: P3, row_offset: 3}
```

Another alternative is to write the files as matching lists of values as in 
```
column_offset: [-0.599, -0.143, 0.047, 0.0, -0.261, -1.07, -0.212, -0.287, 0.419,
  -0.076, -0.166, 0.145]
filter_names: [F090W, F115W, F140M, F150W, F158M, F200W, F277W, F356W, F380M, F430M,
  F444W, F480M]
meta: {date: '2020-04-01T08:33:21.293', filename: test1.asdf, model_type: FilteroffsetModel,
  telescope: JWST}
pupil_names: [F480M, F444W, F430M, F380M, F356W, F277W, F200W, F158M, F150W, F140M,
  F115W, F090W]
row_offset: [-0.007, 0.068, 0.016, 0.0, -0.322, 2.198, 0.254, -0.068, -0.327, -0.075,
  -0.003, -0.311]
```
That would be a different schema. IN both cases we can keep MIRI files readable.

I have no preference.

jwst/datamodels/flat.py
hbushouse: They're really only needed for models where you want to have a dq or err array, but there isn't a dq or err array in the file that's being loaded into the model. This is usually only the case for level-1b science products getting loaded into a `RampModel`. I see no need for it in any of the models for reference files.

jwst/datamodels/schemas/filteroffset.schema.yaml
hbushouse: I like the idea of the first scheme in which the files for NIRCam and NIRISS have columns of `filter`, `pupil`, `column_offset`, and `row_offset`, while the MIRI file remains the same.

jwst/datamodels/schemas/filteroffset.schema.yaml
```diff
@@ -8,6 +8,8 @@ allOf:
 - $ref: keyword_exptype.schema
 - $ref: keyword_filter.schema
```
hbushouse, line 9: Do we need FILTER in the headers of these ref files? It isn't currently used as a file selector in CRDS for MIRI, and it sounds like it won't be used as a selector for NIRCam or NIRISS either. Doesn't hurt to have it there, but I don't think it's needed. It's only needed as a key in the ref file data itself.

jwst/datamodels/wcs_ref_models.py
```diff
@@ -667,30 +667,46 @@ class FilteroffsetModel(ReferenceFileModel):
...
-        F1550C|F560W|F2550WR|FND|F2550W|F1500W|F1000W|F1065C|F770W|F1280W|"
-
-    def on_save(self, path=None):
         self.meta.reftype = self.reftype
+        if self.meta.instrument.name == "MIRI":
+            self.meta.instrument.detector = "MIRIMAGE"
+            self.meta.instrument.pfilter = "F1130W|F1140C|F2300C|F2100W|F1800W|\
+            F1550C|F560W|F2550WR|FND|F2550W|F1500W|F1000W|F1065C|F770W|F1280W|"
+        elif self.meta.instrument.name == "NIRCAM":
+            self.meta.instrument.pdetector = "NRCA1|NRCA2|NRCA3|NRCA4|NRCALONG|\
```
hbushouse, line 685: Since the NIRCam ref files aren't being selected by detector, do we need this? Not even sure it'll work, because the `pdetector` attribute isn't in the schema. Or is this needed for something internally?

nden: Yes, this should be removed. It's left from one of the many iterations I did.

jwst/datamodels/schemas/filteroffset.schema.yaml
nden: removed

PR #4751, opened 2020-03-31
Repin tweakwcs==0.5.3

jdavies-st:
Pinning `tweakwcs==0.5.3` as it has been until yesterday.  Yesterday we unpinned it, ran the regression tests using `tweakwcs 0.6.1` and saw errors in our regression tests.  The MIRI data failed.

https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/595/testReport/junit/jwst.regtest/test_miri_image/_stable_deps__test_miri_image3_i2d/

Any ideas @mcara?

The NIRCam imaging test showed differences using the new version of `tweakwcs`

https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/595/testReport/junit/jwst.regtest/test_nircam_image/_stable_deps__test_nircam_image_stage3_i2d/

But this I think is just a difference noted in the accuracy of the affine transform fit in the newer version?  The combined image looks good.

We can unpin this when the issues are sorted.  Maybe when #4542 is merged.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -101,7 +101,7 @@ setup(
         'spherical-geometry>=1.2',
         'stsci.image>=2.3.3',
         'stsci.imagestats>=1.4',
-        'tweakwcs>=0.5.3',
+        'tweakwcs==0.5.3',
     ],
     extras_require={
         'docs': DOCS_REQUIRE,


nden:
My understanding is that #4542 needs tweakwcs >0.5.3. @mcara can you clarify the order these PRs should be merged given that gwcs v.13 is used.

jdavies-st:
Yeah, the removal of `tpcorr` does need `tweakwcs 0.6.1`.  That PR should have unpinning this as part of it.

For now, it would be good to get a clean regtest run, as there were a bunch of other changes to drizzled products, that off-by-one bug now fixed in `gwcs 0.13` (that I believe explains the changes in results of S_REGION) that would be nice to get okified.

Issue #4753, opened 2020-03-31
Error being hidden when referred to config files are missing


stscijgbot:
Issue [JP-1393](https://jira.stsci.edu/browse/JP-1393) was created by Jonathan Eisenhamer:

When a cfg file has references to other cfg files, but those other cfg files are missing, no error is reported and the process stops silently. An error should be reported.

stscijgbot:
Comment by Jonathan Eisenhamer: Resolved by [PR #4761|https://github.com/spacetelescope/jwst/pull/4761]

PR #4468, opened 2020-01-22
Fix strun --version bug

jdavies-st:
Now:
```
$ strun --version
0.14.3a1.dev107+gb6b7c008.d20200122
```

Fixes #4467

Issue #2440, opened 2018-08-30
Disallow known optical paths that produce no data on NIRSpec NRS2 detector


stscieisenhamer:
# Issue
There are known optical paths for NIRSpec where there will be no data on the NRS2 detector. Modify the Level2/Level3 rules as needed to not include exposures that fall in this category.

# Resources
From [this comment](https://github.com/spacetelescope/jwst/issues/2336#issuecomment-412113876) in #2336:

> So, the reverse then, is the cases we know that will not have NRS2 as listed in #2151. Note that this discussion is specific to actual science; all the calibration-type exposures are already handled in their own ways, for example see #1188.

# Notes
With the resolution of #2336, which handles both this situation and others where the Level2 calibration needs be run to determine whether NRS2 science exists, the resolution of this issue is more of a preemptive feature that will create more science-correct associations, but otherwise would have no real affect on processing.

nden:
I have a recollection this was fixed already. @stscieisenhamer ?

stscieisenhamer:
Yes, this is resolved by PR #5504

Issue #2336, opened 2018-08-08
No IFU slices fall on detector NRS2


stscieisenhamer:
# Issue
From SDP data jw84700,  a run of `calwebb_spec2` reports the following:
<details><summary>Error Log</summary>
<p>

```
$ strun cfgs/calwebb_spec2.cfg /ifs/int/jwstc/store/doggett/tests/WholeWorks/cstring/JWSTDP-2018_2-180802/2018-08-06-153204/sdp/asn_creation/cal/level2/jw84700-o006_20180806t170209_spec2_003_asn.json --output_dir level2b
2018-08-08 02:13:56,219 - stpipe - WARNING - /ifs/int/jwstc/info/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/photutils/background/core.py:71: AstropyDeprecationWarning: "iters" was deprecated in version 3.1 and will be removed in a future version. Use argument "maxiters" instead.
  def __init__(self, sigma_clip=SigmaClip(sigma=3., iters=5)):

2018-08-08 02:13:56,220 - stpipe - WARNING - /ifs/int/jwstc/info/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/photutils/background/core.py:114: AstropyDeprecationWarning: "iters" was deprecated in version 3.1 and will be removed in a future version. Use argument "maxiters" instead.
  def __init__(self, sigma_clip=SigmaClip(sigma=3., iters=5)):

2018-08-08 02:13:56,224 - stpipe - WARNING - /ifs/int/jwstc/info/eisenham/anaconda3/envs/condadev/lib/python3.6/site-packages/photutils/background/background_2d.py:285: AstropyDeprecationWarning: "iters" was deprecated in version 3.1 and will be removed in a future version. Use argument "maxiters" instead.
  sigma_clip=SigmaClip(sigma=3., iters=10),

2018-08-08 02:13:56,922 - stpipe.Spec2Pipeline - INFO - Spec2Pipeline instance created.
2018-08-08 02:13:56,924 - stpipe.Spec2Pipeline.bkg_subtract - INFO - BackgroundStep instance created.
2018-08-08 02:13:56,926 - stpipe.Spec2Pipeline.assign_wcs - INFO - AssignWcsStep instance created.
2018-08-08 02:13:56,927 - stpipe.Spec2Pipeline.imprint_subtract - INFO - ImprintStep instance created.
[...]

PR #4761, opened 2020-04-01
JP-1393 Re-enable exeption tracebacks from strun for issues outside step processing

stscieisenhamer:
PR #4468 removed exception traceback printing from `strun` to avoid double tracebacks on failure. However, the side effect is that any error happening outside of step processing now went unreported, causing silent failures. In particular, config file errors would cease processing without reason.

This refactor restores `strun` but removes the traceback from `cmdline.step_from_cmdline`, ensuring that all errors get reported and avoiding the double traceback.

### DIFF
--- a/scripts/strun
+++ b/scripts/strun
@@ -25,6 +25,10 @@ if __name__ == '__main__':
     try:
         step = Step.from_cmdline(sys.argv[1:])
     except NoDataOnDetectorError:
+        import traceback
+        traceback.print_exc()
         sys.exit(64)
     except Exception:
+        import traceback
+        traceback.print_exc()
         sys.exit(1)
--- a/scripts/strun
+++ b/scripts/strun
@@ -24,10 +24,20 @@ if __name__ == '__main__':
 
     try:
         step = Step.from_cmdline(sys.argv[1:])
+
     except NoDataOnDetectorError:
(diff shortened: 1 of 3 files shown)


stscieisenhamer:
Resolves #4753 
Resolves [JP-1393](https://jira.stsci.edu/browse/JP-1393)

scripts/strun
```diff
@@ -25,6 +25,10 @@ if __name__ == '__main__':
     try:
         step = Step.from_cmdline(sys.argv[1:])
     except NoDataOnDetectorError:
+        import traceback
+        traceback.print_exc()
         sys.exit(64)
```
hbushouse, line 30: Nothing to do with this PR, but we should have some documentation in the code here as to the significance of the exit status = 64, as described in painful detail in #2336.

And speaking of #2336, it appears that the work to modify the ASN generator rules to exclude some NRS2 datasets (#2440) has never been implemented? We should do that.

jdavies-st:
The failing `flake8` check is unrelated, but you may want to fix it here.  Not sure how that got in to `master`.

scripts/strun
stscieisenhamer: Added some documentation.

About the associations, I hear that the new gitbot is about ready. When that happens, a new jira issue should be created linking to #2336.

PR #4765, opened 2020-04-02
Add metadata blending to resample_spec

### LINKED ISSUE #4724: calwebb_spec3 needs to run model_blender on combined output
Issue [JP-1383](https://jira.stsci.edu/browse/JP-1383) was created by Howard Bushouse:

The calwebb_spec3 pipeline needs to apply the model_blender.blendmeta function to the combined s2d/s3d products that it creates, in the same way that the calwebb_image3 pipeline does for combined i2d products. The calwebb_ami3 and calwebb_coron3 pipelines also already do this for their combined products.

In the calwebb_image3 pipeline, it's the resample step applies blendmeta to the i2d product that it creates. The cube_build step in calwebb_spec3 already calls blendmeta for its s3d products, so that particular case is probably already covered. So that suggests that the only remaining update needs to be made to the resample_spec step used in calwebb_spec3, to blend the meta data for combined s2d products.

jwhite3:
This PR adds metadata blending to `resample_spec` and therefore the `Spec3Pipeline`

Resolves #4724 / [JP-1383](https://jira.stsci.edu/browse/JP-1383)

### DIFF
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -15,7 +15,7 @@ from gwcs import coordinate_frames as cf
 from .. import datamodels
 from . import gwcs_drizzle
 from . import resample_utils
-
+from ..model_blender import blendmeta
 
 CRBIT = np.uint32(datamodels.dqflags.pixel['JUMP_DET'])
 
@@ -192,6 +192,14 @@ class ResampleSpecData:
 
         return output_wcs
 
+    def blend_output_metadata(self, output_model):
+        """Create new output metadata based on blending all input metadata."""
+        # Run fitsblender on output product
+        output_file = output_model.meta.filename
+
+        log.info(f'Blending metadata for {output_file}')
+        blendmeta.blendmodels(output_model, inputs=self.input_models, output=output_file)
+
     def do_drizzle(self, xmin=0, xmax=0, ymin=0, ymax=0, **pars):
         """ Perform drizzling operation on input images's to create a new output
         """
@@ -227,6 +235,9 @@ class ResampleSpecData:
             bb = resample_utils.wcs_bbox_from_shape(output_model.data.shape)
             output_model.meta.wcs.bounding_box = bb
             output_model.meta.filename = obs_product
+
+            if self.drizpars['blendheaders']:
+                self.blend_output_metadata(output_model)
 
             exposure_times = {'start': [], 'end': []}
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -85,6 +85,8 @@ resample_spec
 - Fix bug that was causing resampled MIRI LRS fixed-slit data to be all zero.
   [#4552]
 
+- Enable model metadata blending [#4765]
+
 source_catalog
 --------------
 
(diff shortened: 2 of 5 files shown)


jwst/resample/resample_spec.py
```diff
@@ -228,6 +236,9 @@ def do_drizzle(self, xmin=0, xmax=0, ymin=0, ymax=0, **pars):
             output_model.meta.wcs.bounding_box = bb
             output_model.meta.filename = obs_product
 
+            if self.drizpars['blendheaders']:
+                self.blend_output_metadata(output_model)
+
```
hbushouse, line 241: I notice that the implementation in `resample` has a couple extra lines here to save the value of the DATAMODL (meta.model_type) keyword from the original output model and then store it again after it gets clobbered by the blending process (i.e. the output model_type gets overwritten by the model_type from the inputs during blending). This may not be necessary anymore, now that we're storing resampled data in the same kind of model as the input data, but it would be worth checking to confirm that we no longer need to worry about that. A few regtests should tell us the answer.

jwhite3:
@stscieisenhamer whoops.. I keep forgetting to update the changelist! Sorry about that, I'll push the changes soon

jwst/resample/resample_spec.py
jwhite3: Do you want to include that in the scope of this PR, or check in on it in a different one?

hbushouse: I guess we can just merge this as-is and watch for any fallout from the regtests. A patch could be added later, if necessary.

jwhite3: No worries, I'll check it out now and see what happens. Shouldn't take long in any case

jwhite3: @hbushouse So, I commented out the lines that you mentioned in `resample` and then ran the `test_nircam_image` regression tests. Looks like they passed, and I was able to confirm that the test cases included model blending.

hbushouse: OK, good. I would also try running a couple spectroscopic regtests on this branch to make sure we're also good for spectroscopic resampled data models (e.g. SlitModel, MultiSlitModel).

jwhite3: @hbushouse In running the regression tests under `test_nirspec_fs_spec3` the differences seem to make sense (I think). For example, in the primary header, some date and time related keywords are different, and other differences are a result of the number of extensions being different (due to blending)

hbushouse: That sounds as expected. Some of the date/time keywords get modified by blending to correspond to the min/max/sum of all inputs. So I think we're good!

jwhite3: Just tried it for `nirspec_mos_spec3`, and there are the same differences as expected, although it appears for one of the parametrized sources there are differences in the `crf` file. Any thoughts for that one?

hbushouse: Differences in pixel values?

jwhite3: In the DQ images it looks like

hbushouse: Did you happen to rebase against master, which includes the PR from @stscieisenhamer on DQ handling in the `outlier_detection` step? If not, I've also occasionally seen small numbers of differences in `crf` files just due to differences in numerical precision when running the test on my own machine vs jwcalibdev, such that some pixel values are just slightly different enough to cross from one side of the detection threshold to the other, resulting in more or fewer DQ flags.

hbushouse:
We've got a clean regtest run as a baseline for comparison, so I'm merging.

PR #4766, opened 2020-04-02
JP-936 Calspec 3 for Resampling Multslit data

jemorrison:
@jdavies-st 
This pr is the first step of trying to define the out wcs using all the slits. But this PR is not  right - it fails.  I get an error down in gwcs_drizzle which I think is related to an incorrect out_wcs that is now being created. 
 I could use a bit of help in what is going on in build_interpolated_output_wcs.

x_tan_array, y_tan_array - these are found from inverse transform of ra, dec of the slit values  and are the  undistorted x,y on the detector ? Do I have that correct ?
I am not sure what pix_to_ra is. It seems to set up a transform of x in the slit bases reference system to ra. 
When a single input model was being used this consisted of 23 values. But now that I have appended the arrays together this consist of 90 value. So the code pix_to_ra = fitter(fit_model, np.arange(x_tan_array.shape[0]), x_tan_array) is not going to work correctly 
should I  group all the data from NRS1 and separately for NRS2. 
I am not sure how to define the final transform: 
transform = mapping | (pix_to_ra & pix_to_dec | undist2sky) & pix_to_wavelength 
Given that we have data from NRS 1 and NRS 2 that defines the final WCS.

Suggests of what I should change  ?


jemorrison:
@jdavies-st 
This new code is my effort to find the output WCS that includes the data from a single slit from different nods and the two detectors. It still is not correct because I get an error down in resample_utils. reproject. 
sky = forward_transform(x,y) result in all Nans

jemorrison:
@jdavies-st  or @hbushouse  or @stscieisenhamer 
In resample_spec I reconfigured how build_interpolated_output_wcs works so that it uses the
entire set of slits to define the output wcs. I left in the code if there is only one slit to test against.
When there is 1 slit and I used the new method of defining the output WCS the output WCS is basically the same as the old way. It is 1 pixel larger in the spatial dimension.
The code runs into an error when I trying to use 2 slits in resample and let the output wcs be defined using those 2 slits. The error occurs in resample_utilts._reproject

Basically what is happening 
pixmap = np.dstack(reproject(in_wcs, out_wcs)(grid[0], grid[1]))

in_wcs = input model 
out_wcs- is the final WCS using all the data
grid[0], grid[1] is based on bb of the input model

I believe that backward transform is not working correctly.For the forward transform takes the x,y and projects these values to the sky.:
 sky = forward_transform(x, y)
Flat_sky is formed from  the three different axis of the sky values. I get an error in backward transform 
File "/Users/morrison/git/jwst/jwst/resample/gwcs_drizzle.py", line 429, in dodrizzle
    pixmap = resample_utils.calc_gwcs_pixmap(input_wcs, output_wcs, insci.shape)
  File "/Users/morrison/git/jwst/jwst/resample/resample_utils.py", line 117, in calc_gwcs_pixmap
    pixmap = np.dstack(reproject(in_wcs, out_wcs)(grid[0], grid[1]))
  File "/Users/morrison/git/jwst/jwst/resample/resample_utils.py", line 190, in _reproject
    det = backward_transform(*tuple(flat_sky))

........ File "/Users/morrison/anaconda/envs/dev3/lib/python3.7/site-packages/scipy/interpolate/interpolate.py", line 2628, in interpn
    "ascending" % i)
ValueError: The points in dimension 0 must be strictly ascending

I think this is telling me that the flat_sky has a problem in that the points are not ascending.
So I check each axis of flat sky to see if they are "strictly ascending" and they are not
 I have gone through each axis (ra, dec,wavelength) of flat_sky and check if the axes are ascending. None of them are. I am a bit confused because at this stage all I have done is
gone from input model - > input wcs -> sky . So  that should be fine.
Maybe I am not understanding the error.  Maybe the error is occurring in the mapping from
output wcs to det.  At this stage I am not sure what array needs to be ascending. James do you have any idea ?

jemorrison:
Ok I finally figured out why when running resample using two files from NRS1 that the output WCS was producing errors in the backward transform. The wavelength output WCS is a table. So  using two files from the same NRS1 and just appending the wavelengths from each input model together (and then sorting it) caused the wavelength array to have duplicates (which totally makes sense)
THe inverse transform is then confused because there are two values of every wavelength in the array of wavelengths. So I need to either remove duplicates so define the output wavelength array as linear ? Using all the wavelengths from the appended input models to for this linear wavelength 
Does it seem reasonable to have the wavelength output dimension as linear ?

hbushouse:
Seems to me that the most appropriate thing to do is simply remove duplicates from the combined wavelength table, rather than trying to compute a new set of linearized values.

jdavies-st:
What you are trying to do is essentially splice together the wavelength arrays, i.e. something akin to what `combine1d` does.  Though you are doing it with a 2D array of wavelengths instead of a 1D array.

jwst/resample/resample_spec.py
```diff
@@ -90,105 +91,298 @@ def build_interpolated_output_wcs(self, refmodel=None):
...
+                ra, dec, lam = np.array(wcs(*grid))
+
+                spectral_axis = find_dispersion_axis(model)
+                spatial_axis = spectral_axis ^ 1
+            # print('spectral axis',spectral_axis)
+            # print('spatial axis',spatial_axis)
+            # Compute the wavelength array, trimming NaNs from the ends
+                wavelength_array = np.nanmedian(lam, axis=spectral_axis)
+                wavelength_array = wavelength_array[~np.isnan(wavelength_array)]
+                all_wavelength.append(wavelength_array)
```
jdavies-st, line 227: I don't think we can just append here.  It is important to not just have the wavelengths, but to know the points (pixel positions) that map to those wavelengths)

jemorrison:
Here is my plan - which may or may not work with the way the WCS is set up. Append all the wavelengths together and create a table that defines the wavelength to output grid.
We use the input wcs to convert pixels to wavelength and the output WCS takes these wavelength and converts them to output grid pixels. If we need to go from output wavelength to detector pixels (maybe for outlier detection) then we need to go from output wavelength to detector using the inverse wcs of the model for that detector.  
It seems resample might not be set up to work like this. It becomes a problem in keeping the information of what wavelength maps to the detector pixel because NRS1 and NRS2 are going to
have such different detector to wavelength mapping. If it is necessary to go from output wcs (wavelength) to detector pixels with a single transform that is going to hard because we need to know which detector the data comes from.

jemorrison:
@hbushouse @jdavies-st @jmuzerolle 
Ok I think I am closer. I think it is fine appending the wavelengths as I have done.  When working with the WCS resample_spec always ties an input wcs with the output wcs.  I have tested it on
the NRS 1 and NRS 2 nodded data. James M I put the resampled image on
/grp/jwst/ssb/chartreuse/calspec3/JP-936.
It is really long (uses full wavelength range) but at the end of the NRS wavelength range the image looks a little funny. I have included a portion of the image in the comment. It included the end of the NRS1, the gap, and the beginning for NRS2. James M is this what you are expecting ?
Screen Shot 2020-04-10 at 12 55 12 PM

jemorrison:
@hbushouse  @jdavies-st 
With this new push of code I fixed a MIRI LRS problem. I have left the old code in resample_spec and I set a flag if the variable 'do = 1' the old code will run if there is only 1 input model.
I ran some MIRI_LRS data that subtracted a background nod through calspec2 using the old way and the new code. There should be NO difference in the output. But there is the slit is flipped in the image - see image below. I cannot see what I am missing - what would cause this ?
The old way is on the left and the new way is on the right
Screen Shot 2020-04-12 at 11 06 57 AM

hbushouse:
The spatial orientation of the LRS results seem to be the same, but it appears that the dispersion direction has flipped vertically. I recall that the original dispersion direction for LRS has wavelength increasing towards the bottom, i.e. it has negative parity w.r.t. the y image axis. Perhaps the new way of combining wavelength arrays from the multiple inputs results in wavelengths always increasing with y (row number)? Not necessarily a bad thing, but the MIRI team might want the wavelength direction to remain the same.

jemorrison:
Ok I figured out why the wavelengths increase in an opposite direction. I sort the wavelength array. The plan was to append all the wavelengths together and then sort it at the end to figure out
the mapping in the output WCS from wavelength to output pixels. 
So if the MIRI LRS team wants the wavelength to decrease up the image I can re-order the wavelength. I can search on detector and if detector is MIRI I can flip the wavelength array

hbushouse:
I'm assuming the MIRI LRS folks would like the dispersion direction to stay the same in resampled data, otherwise it could be confusing to users. So, yes, you should just make sure the sort on wavelength goes in the negative direction if exp_type='MIR_LRS-FIXEDSLIT'.

jemorrison:
Added a fix for MIRI LRS. Now a nodded group of MIRI LRS data produces an image with wavelength decreasing up the y axis and we have two negative slits in the image. See below
Screen Shot 2020-04-13 at 2 12 30 PM

jmuzerolle:
Thanks, Jane, this is definitely getting closer!  There seem to be a few lingering issues - the individual nods aren't overlapping correctly (positive traces at the top and bottom, negative in the middle, while it should be the opposite), and the size of the gap is smaller than it should be by roughly an order of magnitude.  Also, there appear to be some non-zero pixels inside the gap that shouldn't be there.


Best,

James


jemorrison:
@hbushouse @jdavies-st 
The code is ready for review

jwst/resample/resample_spec.py
```diff
@@ -91,73 +92,155 @@ def build_interpolated_output_wcs(self, refmodel=None):
...
+            # 5. using x_tan, y_tan perform a linear fit to find spatial sampling
+            # first input model sets intializes wavelength array and defines
+            # the spatial scale of the output wcs
+            if im == 0:
+                for iw in wavelength_array:
+                    all_wavelength.append(iw)
+
+                lam_center_index = int((bb[spectral_axis][1] -
+                                            bb[spectral_axis][0]) / 2)
+                if not spatial_axis:
```
hbushouse, line 144: I know this syntax works mathematically and programmatically, but to a human reading it the immediate interpretation is "we're working with an axis that isn't spatial". I think it would be more obvious for readability to just use "if spatial_axis == 0"

jwst/resample/resample_spec.py
```diff
@@ -91,73 +92,155 @@ def build_interpolated_output_wcs(self, refmodel=None):
...
+            # it is assumed the spatial sampling is the same for all the input
+            # models. So we can use the first input model to set the spatial
+            # sampling.
+            # Note: using the first input model sets up the spatial output to match
+            # this  model. For nodded (background subtracted data)
+            # this means the positive and negative traces will have the same
+            # configurations as the first input model.
+
+            # Steps to do this for first input model:
+            # 1. find the center of slit in wavelength
```
hbushouse, line 130: I know these are just comments and hence don't affect the actual processing, but when I read "center of slit", I immediately picture in my mind a physical, rectangular slit, that has its long axis in the spatial direction (which is natural for a spectrograph, because you want to he spectral dimension of the slit to be narrow) and hence the center of the slit refers to something in the spatial (long) dimension of the slit itself. I think what you're referring to here is the center/middle of the spectral trace along the wavelength axis, right? So perhaps this should be reworded along those lines, e.g. find the middle of the spectrum in wavelength. It's the spectrum we're operating on, not the entrance slit.

jwst/resample/resample_spec.py
```diff
@@ -91,73 +92,155 @@ def build_interpolated_output_wcs(self, refmodel=None):
...
+                native2celestial = RotateNative2Celestial(ra_center_pt, dec_center_pt, 180)
+                undist2sky1 = tan | native2celestial
+                # Filter out RuntimeWarnings due to computed NaNs in the WCS
+                warnings.simplefilter("ignore")
+                # at this center of slit find x,y tangent proction - x_tan, y_tan
+                x_tan, y_tan = undist2sky1.inverse(ra, dec)
+                warnings.resetwarnings()
+
+                # pull out data from center
+                if not spectral_axis:
```
hbushouse, line 165: same comment on syntax as above

jwst/resample/resample_spec.py
```diff
@@ -91,73 +92,155 @@ def build_interpolated_output_wcs(self, refmodel=None):
...
+        # done looping over set of models
+
+        all_ra = np.hstack(all_ra_slit)
+        all_dec = np.hstack(all_dec_slit)
+        all_wave = np.hstack(all_wavelength)
+        all_wave = all_wave[~np.isnan(all_wave)]
+        all_wave = np.sort(all_wave,axis=None)
         # Tabular interpolation model, pixels -> lambda
+        wavelength_array = np.unique(all_wave)
+        if self.input_models[0].meta.exposure.type == 'MIR_LRS-FIXEDSLIT' :
```
hbushouse, line 214: Should add a comment here as to why this is done for MIRI LRS

jwst/resample/resample_spec_step.py
```diff
@@ -76,8 +77,9 @@ def _process_multislit(self, input_models):
         """
         containers = multislit_to_container(input_models)
         result = datamodels.MultiSlitModel()
+
         result.update(input_models[0], only="PRIMARY")
-        result.update(input_models[0], only="SCI")
+        #result.update(input_models[0], only="SCI")
```
hbushouse, line 82: Why aren't we updating the SCI header any more?

jwst/resample/resample_spec.py
```diff
@@ -183,23 +298,14 @@ def build_interpolated_output_wcs(self, refmodel=None):
...
-
+        output_array_size[spatial_axis] = x_size
         # turn the size into a numpy shape in (y, x) order
         self.data_size = tuple(output_array_size[::-1])
         bounding_box = resample_utils.wcs_bbox_from_shape(self.data_size)
         output_wcs.bounding_box = bounding_box
 
         return output_wcs
 
-    def blend_output_metadata(self, output_model):
```
hbushouse, line 195: Why is the blending operation being removed? We still need that? I noticed that the import of blendmeta was removed up above too. Is this just a conflict in code versions?

jemorrison:
@jdavies-st 
I realized that I removed from resample_spec the ability to use refmodel.  Does that seem ok
to you James and if it does then I should remove in  ? 
def build_interpolated_output_wcs(self, refmodel=None)

hbushouse:
Merging. Any remaining refactoring that doesn't directly affect the results can be continued later.

PR #4767, opened 2020-04-02
Use bitwise operator to check for presence of flag

eslavich:
In the midst of a conversation with @jdavies-st about DQ flags, I made a search through this codebase for cases where we check for flag presence with `==` instead of `&`.  This is the one example that turned up.

### DIFF
--- a/jwst/ramp_fitting/utils.py
+++ b/jwst/ramp_fitting/utils.py
@@ -543,7 +543,7 @@ def calc_slope_vars(rn_sect, gain_sect, gdq_sect, group_time, max_seg):
         del wh_good
 
         # Locate any CRs that appear before the first SAT group...
-        wh_cr = np.where( gdq_2d_nan[i_read, :] == dqflags.group['JUMP_DET'])
+        wh_cr = np.where( gdq_2d_nan[i_read, :] & dqflags.group['JUMP_DET'] > 0 )
 
         # ... but not on final read:
         if (len(wh_cr[0]) > 0 and (i_read < nreads-1) ):
--- a/jwst/ramp_fitting/utils.py
+++ b/jwst/ramp_fitting/utils.py
@@ -543,7 +543,7 @@ def calc_slope_vars(rn_sect, gain_sect, gdq_sect, group_time, max_seg):
         del wh_good
 
         # Locate any CRs that appear before the first SAT group...
-        wh_cr = np.where( gdq_2d_nan[i_read, :] & dqflags.group['JUMP_DET'] > 0 )
+        wh_cr = np.where( gdq_2d_nan[i_read, :].astype(np.int32) & dqflags.group['JUMP_DET'] > 0 )
 
         # ... but not on final read:
         if (len(wh_cr[0]) > 0 and (i_read < nreads-1) ):


jwst/ramp_fitting/utils.py
```diff
@@ -543,7 +543,7 @@ def calc_slope_vars(rn_sect, gain_sect, gdq_sect, group_time, max_seg):
         del wh_good
 
         # Locate any CRs that appear before the first SAT group...
-        wh_cr = np.where( gdq_2d_nan[i_read, :] == dqflags.group['JUMP_DET'])
+        wh_cr = np.where( gdq_2d_nan[i_read, :] & dqflags.group['JUMP_DET'] > 0 )
```
jdavies-st, line 546: Unfortunately `gdq_2d_nan` is a float here, so it won't work.  Not sure why it needs to be a float.

https://github.com/spacetelescope/jwst/blob/2b54a372d588091f2faa9a9e212f59ff132c3174/jwst/ramp_fitting/utils.py#L519

eslavich: Whoops!  Guess it's a float to support nan values?  Anyway, I'll close this, seems like my "fix" would do more harm than good.

eslavich: Aaaaaand it's back!

dmggh:
Here is a simple example showing the intent of the code in locating cosmic rays. For a SATURATED group 7 (which the ramp fitting code has set to nan, and set all later groups to nan) and a JUMP_DET in groups 3 and 5:

In [66]: gdq = np.zeros(10, dtype=np.float32)                                                  

In [67]: gdq[7:] = np.nan                                                                       

In [68]: gdq[3] = dqflags.group['JUMP_DET']                                                    

In [69]: gdq[5] = dqflags.group['JUMP_DET']       

In [70]: gdq                                                                                   
Out[70]: array([ 0.,  0.,  0.,  4.,  0.,  4.,  0., nan,  nan,  nan], dtype=float32)

In [71]: wh_cr = np.where( gdq == dqflags.group['JUMP_DET'])                                                                       

In [72]: wh_cr                                                                                 
Out[73]: (array([3,5]),)

jdavies-st:
Consider your example slightly modified:

```
In [1]: import numpy as np

In [2]: from jwst.datamodels import dqflags

In [3]: gdq = np.zeros(10, dtype=np.float32)

In [4]: gdq[7:] = np.nan

In [5]: gdq[3] = dqflags.group['JUMP_DET']

In [6]: gdq[5] = (dqflags.group['JUMP_DET'] | dqflags.group['DO_NOT_USE'])

In [7]: gdq
Out[7]: array([ 0.,  0.,  0.,  4.,  0.,  5.,  0., nan, nan, nan], dtype=float32)

In [8]: np.where(gdq == dqflags.group['JUMP_DET'])
Out[8]: (array([3]),)
```

The last comparison missing the the JUMP_DET flag at index 5.  I believe that's the issue that Ed is trying to fix.

dmggh:
Okay, I see. The step code may already (I will have to check) correctly handle the case where DO_NOT_USE is also flagged, but I will look more carefully at the code in question.

dmggh:
In your last command, recasting the DQ array before doing a bitwise_and seems to have the desired effect:

In [8]: np.where(np.bitwise_and(gdq.astype(np.int32), dqflags.group['JUMP_DET']))                       
Out[8]: (array([3, 5]),)

eslavich:
Casting to int32 turns those np.nan values into REFERENCE_PIXEL?

```
In [13]: np.bitwise_and(np.array([np.nan]).astype(np.int32), dqflags.pixel['REFERENCE_PIXEL']) == dqflags.pixel['REFERENCE_PIXEL']
Out[13]: array([ True])
```

dmggh:
Seems so, but since conversion of a nan to a ref pixel leaves the JUMP_DET flag unset, and since wh_cr, that is assigned to it - 

wh_cr = np.where(np.bitwise_and(gdq.astype(np.int32), dqflags.group['JUMP_DET']))

... is deleted soon thereafter ( before there are any effects elsewhere), this conversion looks ok to me.

PR #4770, opened 2020-04-06
JP-1401: Spectroscopic APCORR data models

### LINKED ISSUE #4768: CLONE - Create initial versions of imaging APCORR data models
Issue [JP-1401](https://jira.stsci.edu/browse/JP-1401) was created by Howard Bushouse:

JP-1022 gives the specifications for new data models that will be used for imaging-mode aperture correction (apcorr) reference files, as determined by the instrument teams. We need to create those data models, so that the teams can start using them to create initial versions of the reference files.

hbushouse:
New datamodels schemas for spectroscopic aperture correction reference files, as specified in [JP-1049](https://jira.stsci.edu/browse/JP-1049). Still need to add the new models to the reference file docs.

Fixes #4768 / [JP-1401](https://jira.stsci.edu/browse/JP-1401)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,8 @@ datamodels
 - Remove lev3_prod schema and move resample-related keywords to
   core schema. [#4552]
 
+- Add data models for spectroscopic mode APCORR reference files. [#4770]
+
 extract_1d
 ----------
 
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Software vs DMS build version map
 
 | jwst tag | DMS build | CRDS_CONTEXT |   Date     |          Notes                           |
 | -------- | --------- | ------------ | ---------- | -----------------------------------------|
-|  0.15.1  | B7.4.2rc2 | 0586         | 03/10/2020 | Second release candidate for B7.4.2      |
+|  0.15.1  | B7.4.2    | 0586         | 03/10/2020 | Final release candidate for B7.4.2       |
 |  0.15.0  | B7.4.2rc1 | 0585         | 02/28/2020 | First release candidate for B7.4.2       |
 |  0.14.2  | B7.4      | 0570         | 11/18/2019 | Final release candidate for B7.4         |
 |  0.14.1  | B7.4rc2   | 0568         | 11/11/2019 | Second release candidate for B7.4        |
--- a/docs/jwst/references_general/apcorr_reffile.inc
+++ b/docs/jwst/references_general/apcorr_reffile.inc
@@ -1,0 +1,280 @@
+APCORR Reference File
+---------------------
+
+:REFTYPE: APCORR
+
+The APCORR reference file contains data necessary for correcting
+extracted imaging and spectroscopic photometry to the equivalent
+of an infinite aperture. It is used within the
+:ref:`source_catalog <source_catalog_step>` step for imaging and
+within the :ref:`extract_1d <extract_1d_step>` step for
+spectroscopic data.
+
+.. include:: ../references_general/apcorr_selection.inc
+
+.. include:: ../includes/standard_keywords.inc
+
+Type Specific Keywords for APCORR
++++++++++++++++++++++++++++++++++
+In addition to the standard reference file keywords listed above,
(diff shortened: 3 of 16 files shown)


jdavies-st:
Is there any chance that someone will try to use a NIRSPec IFU APCORR reffile for NIRSPec MOS?  If the selector takes care of this and the header keywords can differentiate the two reffiles, I think it would make sense to write a single `ApcorrModel` datamodel to handle all cases.  The schema then would be something like:

```
%YAML 1.1
---
id: "http://stsci.edu/schemas/jwst_datamodel/apcorr.schema"
title: Aperture correction data model
allOf:
- $ref: referencefile.schema
- anyOf:
  - type: object
    properties:
      apcorr_table:
        title: Aperture correction factors table (NIRSpec Fixed Slit)
        fits_hdu: APCORR
        datatype:
        - name: filter
          datatype: [ascii, 12]
        - name: grating
          datatype: [ascii, 15]
        - name: slit
          datatype: [ascii, 15]
        - name: wavelength
          datatype: float32
          ndim: 1
        - name: nelem_wl
          datatype: int16
        - name: size
          datatype: float32
          ndim: 2
        - name: nelem_size
          datatype: int16
        - name: pixphase
          datatype: float32
          ndim: 1
        - name: apcorr
          datatype: float32
          ndim: 3
        - name: apcorr_err
          datatype: float32
          ndim: 3
  - type: object
    properties:
      apcorr_table:
        title: Aperture correction factors table (NIRSpec MOS and IFU)
        fits_hdu: APCORR
        datatype:
        - name: filter
          datatype: [ascii, 12]
        - name: grating
          datatype: [ascii, 15]
        - name: wavelength
          datatype: float32
          ndim: 1
        - name: nelem_wl
          datatype: int16
        - name: size
          datatype: float32
          ndim: 2
        - name: nelem_size
          datatype: int16
        - name: pixphase
          datatype: float32
          ndim: 1
        - name: apcorr
          datatype: float32
          ndim: 3
        - name: apcorr_err
          datatype: float32
          ndim: 3
...
```
etc, listing all the possible layouts.  The `anyOf` takes care of this. Then there is just a single datamodel, and as long as the reference file conforms to one of those schema layouts, it will validate.  The metadata in the reffile has everything else needed to ID it as NIRSpec MOS or FS or whatever.  And there's only one datamodel. I differentiate the titles of the tables above for clarity.

hbushouse:
@jdavies-st I suppose the two NIRSpec model schemas could be combined in the way you suggest, but you could also then say why not include the schemas for all instrument modes in a single schema? A couple of them are identical (NrcWfss and NisWfss) and others have only slight variations (like the slight differences between NIRSpec MOS/IFU and FS). I think the instrument teams are already used to having dedicated models per mode.

jdavies-st:
Yes, that's exactly my point.  Combine them all into a single datamodel.  There's no need for one for each instrument as they are fundamentally the same except for the table columns.  In order to read the columns correctly, one will need to know what intrument/mode it is for, but one doesn't need a separate class for this, only information on instrument/mode, which is in the header.

stscieisenhamer:
And for the last 2 cents: 

Agree that getting away from instrument-specific models would be really nice.

Ideally, it seems that mode-specific models would be useful, such as ApcorrIFU, ApcorrFS, ApcorrWFSS. This would allow algorithms to decide usage based on mode/type of reference file given. This would disconnect from what the input data is and allow for more generalized code.

However, given the the current code has so much wired in based EXP_TYPE, etc. already, having everything just in one datamodel seems reasonable.

hbushouse:
I sympathize with the "one data model to rule them all" philosophy, but unfortunately the imaging-mode apcorr models are already in place and have been used to generate reference files, and folks have been waiting a looooong time to finalize the spec models, so I really hesitate to go back to the teams at this point and ask if it's OK to change everything again. They want/need these models yesterday.

jdavies-st:
Fine to keep the imaging and spectral ones separate for now.  Fix the imaging ones later.

Why keep propagating huge numbers of models with impossible-to-remember names and lots of repeated code that doesn't actually do anything.  Make it simple.

PR #4771, opened 2020-04-06
JP-1382: Add updates to source models to include header info

### LINKED ISSUE #4725: calwebb_spec3 exp_to_source does not copy/update primary header keywords
Issue [JP-1382](https://jira.stsci.edu/browse/JP-1382) was created by Howard Bushouse:

The calwebb_spec3 pipeline call to the exp_to_source/multislit_to_container function does not copy/update primary header meta data, leaving the primary header of the resulting source-based cal products nearly empty (only the required FITS keywords appear).

The primary header of the resulting source-based cal files should at least contain a copy of the primary header from the first exposure being included in the reorganization process and ultimately could/should even call the model_blender.blendmeta function to create a properly blended header from all the input exposures (and a HDRTAB extension containing all of the original meta data from each exposure).

jwhite3:
This PR resolves #4725 / [JP-1382](https://jira.stsci.edu/browse/JP-1382) where results from `exp_to_source` did not include any header information from the original exposures. Now, each new `MultiExposureModel` will contain header information from the input `MultiSlitModels`

### DIFF
--- a/jwst/exp_to_source/exp_to_source.py
+++ b/jwst/exp_to_source/exp_to_source.py
@@ -36,13 +36,11 @@ def exp_to_source(inputs):
 
     for exposure in inputs:
         log.info(f'Reorganizing data from exposure {exposure.meta.filename}')
-        initial = True
 
         for slit in exposure.slits:
             log.debug(f'Copying source {slit.source_id}')
             result_slit = result[str(slit.source_id)]
             result_slit.exposures.append(slit)
-            result[str(slit.source_id)].exposures.append(slit)
             merge_tree(result_slit.exposures[-1].meta.instance, exposure.meta.instance)
 
             if result_slit.meta.instrument.name is None:
--- a/jwst/exp_to_source/exp_to_source.py
+++ b/jwst/exp_to_source/exp_to_source.py
@@ -7,8 +7,7 @@ from collections.abc import Callable
 
 from ..datamodels import (
     MultiExposureModel,
-    SourceModelContainer,
-    DataModel)
+    SourceModelContainer)
 from ..datamodels.properties import merge_tree
 
 __all__ = ['exp_to_source', 'multislit_to_container']
--- a/jwst/exp_to_source/exp_to_source.py
+++ b/jwst/exp_to_source/exp_to_source.py
@@ -49,6 +49,8 @@ def exp_to_source(inputs):
 
                 result_slit.update(exposure)
 
+            result_slit.meta.filename = None  # Resulting merged data doesn't come from one file
+
         exposure.close()
 
     # Turn off the default factory
--- a/jwst/exp_to_source/exp_to_source.py
+++ b/jwst/exp_to_source/exp_to_source.py
@@ -7,8 +7,8 @@ from collections.abc import Callable
 
 from ..datamodels import (
     MultiExposureModel,
-    SourceModelContainer
-)
+    SourceModelContainer,
+    DataModel)
 from ..datamodels.properties import merge_tree
 
 __all__ = ['exp_to_source', 'multislit_to_container']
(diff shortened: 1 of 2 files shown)


jwhite3:
I will note, that since the information is being included via `update`, there will be no "HDRTAB" extension since the models are not being "blended"

jwhite3:
@hbushouse @stscieisenhamer Actually have a question: 

 One of the unit tests for `exp_to_source` will fail now due to this assertion:

```python
assert outputs[str(slit.source_id)].meta.filename != in_model.meta.filename
```

This due to the `update` call adding in things like `filename` and such from the input models. Is this an issue? If so, then what should `filename` actually be? In terms of the test, it just expects that it's _not_ the input filename rather than suggesting what it _should_ be

stscieisenhamer:
After having just created each `source_id` model, `meta.filename` for each of the new models should be nothing; there is no originating file.

jwhite3:
Ok, then the expected behavior should then be to remove filename after the updates for the new models?

stscieisenhamer:
To further expound, the `meta.filename` of the source-based models cannot be derived because they are created from a number of input files. I think it reasonable that `meta.filename` be explicitly cleared at this point.

jwhite3:
@stscieisenhamer Any other keywords that should be cleared as well along the same line of thought?

jwhite3:
I suppose another approach would be instead of calling `update`, "blend" all of the headers from the input exposure-based data for each resulting slit

stscieisenhamer:
Offhand no. I'm actually a bit surprised that `meta.filename` is actually included, but that is another discussion... In general, `update` has some exclusions built-in, for reasons such as this; they make no sense to transfer between models.

hbushouse:
> Offhand no. I'm actually a bit surprised that `meta.filename` is actually included, but that is another discussion... In general, `update` has some exclusions built-in, for reasons such as this; they make no sense to transfer between models.

I think that copying meta.filename can make sense for steps that are creating an output that is still the same organizational scheme as the input (e.g. all the level-2a and 2b steps that use exposure-based files as both input and output), even if the product type is changing (e.g. going from ImageModel to SlitModel). The stpipe file handling infrastructure handles the appropriate updating of the product type suffix from input to output when the output is serialized.

jwst/exp_to_source/exp_to_source.py
```diff
@@ -33,15 +33,24 @@ def exp_to_source(inputs):
...
-            merge_tree(
-                result[str(slit.source_id)].exposures[-1].meta.instance,
-                exposure.meta.instance
-            )
+            log.debug(f'Copying source {slit.source_id}')
+            result_slit = result[str(slit.source_id)]
+            result_slit.exposures.append(slit)
+            merge_tree(result_slit.exposures[-1].meta.instance, exposure.meta.instance)
+
+            if result_slit.meta.instrument.name is None:
```
stscieisenhamer, line 46: Question: This check is to ensure that the update happens only once?

Quesrtion: The next check on whether inputs is a `DataModel`: The input is always a list, no? What situation is this true?

jwhite3: Not actually sure; essentially I just pulled these specific changes from the other PR as a starting point like we said. I'll look into it

jwhite3: I think that the first condition checks to see whether additional header info should be added. 

The second might be to handle the case where the input isn't a list, but a MultiExposureModel or other model for a single exposure?

stscieisenhamer: Go figure and explains why I'm questioning them. The first check makes sense, but why a non-list would ever be provided here, or allowed, I'm not sure of. I would remove that and ensure at least unit tests pass. Otherwise, should be good.

jwst/exp_to_source/exp_to_source.py
```diff
@@ -33,15 +33,24 @@ def exp_to_source(inputs):
...
+                if isinstance(inputs, DataModel):
+                    result_slit.update(inputs)
+
+                result_slit.update(exposure)
```
jwhite3, line 50:
```suggestion
                result_slit.update(exposure)
```

jwst/exp_to_source/exp_to_source.py
```diff
@@ -7,8 +7,8 @@
 
 from ..datamodels import (
     MultiExposureModel,
-    SourceModelContainer
-)
+    SourceModelContainer,
+    DataModel)
```
hbushouse, line 11: Due to the most recent commit `DataModel` is no longer used and hence flake8 is throwing an error in the Travis tests. Fix this and it should be good to go.

jwhite3, line 11:
```suggestion
    SourceModelContainer)
```

PR #4772, opened 2020-04-07
Update tsophot.schema.yaml

jdavies-st:
`TsoPhotModel` can only get serialized to an ASDF file, so it only needs to be validated against `asdf-schema-1.0.0`.  It does not use the `fits_keyword` and `fits_hdu` attributes.

### DIFF
--- a/jwst/datamodels/schemas/tsophot.schema.yaml
+++ b/jwst/datamodels/schemas/tsophot.schema.yaml
@@ -1,6 +1,6 @@
 %YAML 1.1
 ---
-$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
+$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.0.0"
 id: "http://stsci.edu/schemas/jwst_datamodel/tsophot.schema"
 title: TSO image photometry reference file model (tso_photometry step)
 allOf:


eslavich:
Well this is an interesting riddle.  With this change, `tsophot.schema.yaml` looks like this:

```yaml
%YAML 1.1
---
$schema: "http://stsci.edu/schemas/asdf/asdf-schema-1.0.0"
id: "http://stsci.edu/schemas/jwst_datamodel/tsophot.schema"
title: TSO image photometry reference file model (tso_photometry step)
allOf:
- $ref: referencefile.schema
# ...
```
but the referenced schema, `referencefile.schema.yaml`, looks like this:
```yaml
%YAML 1.1
---
id: "http://stsci.edu/schemas/jwst_datamodel/referencefile.schema"
type: object
properties:
  meta:
    type: object
    properties:
      date:
        title: Date this file was created (UTC)
        type: string
        fits_keyword: DATE
# ...
```
The [jsonschema spec](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.8.1.1) states that subschemas "MUST NOT" include the `$schema` keyword, but it isn't clear how that's supposed to work with a `$ref`.  I know the asdf library just wholesale copies in the entire contents of the referenced schema, including any nested `$schema` field.  I wonder how the validator should behave if the referenced schema has a different metaschema from the referencing schema?  At worst, that could mean that the schemas were written under entirely different jsonschema drafts and aren't compatible with one another.

I don't have a well-formed opinion here, but as I'm writing this I'm starting to think that `tsophot.schema.yaml` has to use the http://stsci.edu/schemas/fits-schema/fits-schema metaschema, even though it doesn't include `fits_keyword` or `fits_hdu` -- otherwise, the complete schema after resolving `$ref` won't be valid.

This may be the opposite of what I've advocated for in the past, this new angle hadn't occurred to me until now.

jdavies-st:
Thanks Ed.  That's helpful.  Maybe that's another angle for our schema tester to make sure things are consistent in the core library.  Not sure how to properly handle that in `asdf`, though that's clearly outside the scope of this PR.

The `asdf` pytest schema tester currently says both versions (old and this) of the `tsophot.schema` are valid.  That is a good chin scratcher.

Anyway, looks like I can close this PR as we don't want to make this change now, or probably at all in the future.

We might want to revisit other schemas in this repo too and get some consistency once things are tightened up in `asdf`.

Thanks!

PR #4776, opened 2020-04-08
Update MIRI filteroffset to use generic model

drlaw1558:
This updates the assign_wcs step of the pipeline to work with MIRI filteroffset files that have been updated to the new generic format.  It also removes the old MIRI-specific filteroffset schema.  This update will likely result in pipeline failures until the new CRDS reference file is in place.


jwst/datamodels/schemas/filteroffset.schema.yaml
```diff
@@ -27,28 +27,14 @@ allOf:
...
+        type: object
+        properties:
+          filter:
+            type: string
+          pupil:
+            type: string
+          row_offset:
+            type: number
+          column_offset:
+            type: number
```
jdavies-st, line 40: The schema should probably be kept as is, as that way old reference file(s) will still be readable. Delivering new reference files will work using the old schema since they satisfy the 2nd `anyOf` type.

Or is it inconceivable that once the new reffile is delivered, anyone will use the old reffiles for any data?  I.e. will they be removed from rmaps?

drlaw1558: With the change to use the generic file format, reference files in the old format will be unusable by the master branch and any future tagged releases.  As such, it would only be folks using old builds such as B7.3 that used the old files, for which I believe the old schema would be available anyway?

As such I don't think there's a benefit to keeping the old schema around; indeed, removing it is probably good since anything in that format will be incompatible with the pipeline.

hbushouse: Since the `assign_wcs` code is also being updated to match the schema change, then I think @drlaw1558 is correct that new code will only work with new ref files and at the same time old code could be used to work with old ref files (requiring the use of certain CRDS_CONTEXT values with the new and old code as well).

Of course this also means that this PR can't be merged until the new/updated ref files are in place in CRDS, right? Because the new code won't work with the existing ref files.

CHANGES.rst
```diff
@@ -6,6 +6,11 @@ ami
 
 - Reorganized step documentation [#4697]
 
+assign_wcs
+---
```
hbushouse, line 10: Underline needs to be same length as section title in line 9

jwst/datamodels/schemas/filteroffset.schema.yaml
```diff
@@ -27,28 +27,14 @@ allOf:
...
-            properties:
-              filter:
-                type: string
-              pupil:
-                type: string
-              row_offset:
-                type: number
-              column_offset:
-                type: number
+      items:
```
hbushouse, line 30: I notice that the old schema had an "outer" attribute of:
```
 - type: array
    items:
```
which is not used in the new schema. Is that OK?

CHANGES.rst
drlaw1558: Fixed

jwst/datamodels/schemas/filteroffset.schema.yaml
drlaw1558: I think someone more knowledgeable than I about datamodel schemas would have to comment; from comparing this to other schemas I concluded that this outer attribute was probably just a special wrapper required in order to have multiple schema options for both the old miri format file and the new generic file.  With the simplification to just one schema I think it's not necessary, but it would be good to confirm that.

nden: Yes, the type needs to be specified at the same level as `description`, as in 

```
filters:
  description: |
    An array of dictionaries where either the
        ...
  type: array
  items:
      type: object
      properties:
         filter:
           type: string
         pupil:
           type: string

```

hbushouse:
The updated MIRI filteroffset ref file has been available for a while now on the crds test server and is just waiting to be migrated to the Ops side so that it's done simultaneously with the merging of this PR. We'd like to do that ASAP, in order to get the code and ref file updates included in our next B7.5 release. So can I get approvals from the reviewers for this? Thanks.

nden:
@hbushouse My comment above has not been addressed - the `filters` field has no type. I cannot approve it but can  push a change to fix it. How can I get the new file to test it.

hbushouse:
@stscieisenhamer can you point us to the new MIRI filteroffset ref file on the CRDS test server?

stscieisenhamer:
I believe the context on the test server is 0597. Note that this is a branch context; it will have nothing else of the other new new reference formats.

hbushouse:
So it looks like the new ref file is either https://jwst-crds-test.stsci.edu/browse/jwst_miri_filteroffset_0004.asdf or https://jwst-crds-test.stsci.edu/browse/jwst_miri_filteroffset_0005.asdf

Not sure which one is supposed to be the latest/greatest.

drlaw1558:
Looks like 0005 is the one with my latest updates

drlaw1558:
I just added the type field to the schema following @nden example

hbushouse:
Just waiting now for the updated ref file to be merged/installed/migrated over from crds-test to the crds ops server, at which point I'll merge this PR.

hbushouse:
Updated filteroffset ref file delivered in CRDS_CONTEXT 616. Merging.

PR #4778, opened 2020-04-09
Default outlier rejection parameters for MRS

drlaw1558:
With the merge of https://github.com/spacetelescope/jwst/pull/4726 outlier rejection is now working for the MRS, but with the default parameters in the outlier rejection config file it is now rejecting all point sources as outliers.  As such the calibration pipeline cannot be used until proper parameters are provided to the routine.

This PR implements a one-line stopgap fix to outlier rejection to force it to use reasonable parameters for MRS observations; once a longer-term solution such as instrument-specific parameter reference files is implemented this line should be removed.


drlaw1558:
Added a changelog entry

PR #4779, opened 2020-04-10
Rewrite README installation instructions

hbushouse:
Reorganize and rewrite the installation instructions in the README, in an effort to clear up some ambiguities and unanswered questions that many installers had. More information is provided as to why the different methods are the way they are, while at the same time attempting to make all of them as consistent as possible, in order to reduce confusion. A complete set of commands is included for each method, so that folks don't have to jump around looking for different pieces that make up the whole process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,73 +13,116 @@ JWST Calibration Pipeline
 Installation
 ------------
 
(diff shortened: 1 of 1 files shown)


hbushouse:
@mwregan2 for some reason I can't seem to assign you as a reviewer for this PR. Please review anyway. Thanks!

hbushouse:
OK, thanks @mwregan2 and @jdavies-st. I'll go ahead and merge this PR and at the same time just close #4745. Thanks @mwregan2 for spurring us to finally make this all more understandable for our users.

PR #4781, opened 2020-04-10
Lock down asdf version to < 2.6

eslavich:
This PR prevents jwst from pulling in the as-yet-unreleased asdf 2.6.  That release will include a change to validator behavior that will require a small update to jwst/datamodels/fits_support.py.

@jdavies-st did I do the right thing in requirements-sdp.txt?

### DIFF
--- a/requirements-sdp.txt
+++ b/requirements-sdp.txt
@@ -1,1 +1,1 @@
-# Nothing here right now
+asdf==2.5.2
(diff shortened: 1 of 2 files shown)


requirements-sdp.txt
```diff
@@ -1 +1 @@
-# Nothing here right now
\ No newline at end of file
+asdf==2.5.2
```
jdavies-st, line 1: We don't want anything here right now.  Eventually this file will get populated via `pip freeze` for build 7.5.

PR #4782, opened 2020-04-10
Refactor wcs_from_footprints to use gwcs.WCS objects instead of FITS keywords

### LINKED ISSUE #3250: Don't use CDELTn in wcs_from_footprints()
Instead, use the gwcs object to compute an x and y scale at the reference pixel.

Prompted by #3247.

### LINKED ISSUE #3900: Unexpected PC matrix behaviour for MIRI imaging
Issue [JP-921](https://jira.stsci.edu/browse/JP-921) was created by David Law:

One long-term development item for simulated MIRI imaging data is to have an 'approximately' correct WCS in the Lvl1b simulated headers (CRPIX, CRVAL, etc).  In order to do so the PC matrix will need to be set to some angle to approximately describe the rotation of the detector with respect to RA/DEC.

However, in testing such changes it seems that changing the PC matrix in Lvl1b data has undesirable affects on the drizzled Lvl2b data downstream.  I'm not sure why this is the case- assign_wcs should set the WCS solution based entirely on the reference distortion files and the five header keywords V2_REF, V3_REF, RA_REF, DEC_REF, ROLL_REF.  Any values of the PC matrix, CRPIX, etc keywords should be irrelevant to the pipeline-derived distortion solution.

Why is this?  Is the resample step simply propagating PC values in the Lvl2a headers without computing new ones?

### LINKED ISSUE #4784: Computed PC matrix for Imaging modes uses incorrect angle
Issue [JP-1404](https://jira.stsci.edu/browse/JP-1404) was created by James Davies:

Currently `jwst.lib.set_telescope_pointing.update_wcs_from_telem` computes the PC matrix for imaging modes using `PA_V3`.

https://github.com/spacetelescope/jwst/blob/d234bf18d734dfe3f059b6a6705ce5bd8719992c/jwst/lib/set_telescope_pointing.py#L440-L445

This is probably incorrect, as we want the rotation angle at the aperture reference point, i.e. `ROLL_REF`.  Same error for FGS guiding mode

https://github.com/spacetelescope/jwst/blob/d234bf18d734dfe3f059b6a6705ce5bd8719992c/jwst/lib/set_telescope_pointing.py#L328-L331

jwhite3:
This PR refactors `wcs_from_footprints` and some functions that it depends on to utilize the `wcs` object from `refmodel` rather than using FITS-dependent keywords. 

Resolves #3250
Resolves #3900 / [JP-921](https://jira.stsci.edu/browse/JP-921)
Resolves #4784

### DIFF
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -111,7 +111,7 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray]) -> float:
 
     crpix = np.array(wcs.backward_transform(*fiducial))
     crpix_with_offsets = np.vstack((crpix, crpix + (1, 0), crpix + (0, 1))).T
-    crval_with_offsets = wcs(*crpix_with_offsets)
+    crval_with_offsets = wcs(*crpix_with_offsets, with_bounding_box=False)
 
     coords = SkyCoord(ra=crval_with_offsets[0], dec=crval_with_offsets[1], unit="deg")
     xscale = np.abs(coords[0].separation(coords[1]).value)
@@ -120,7 +120,7 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray]) -> float:
     return np.sqrt(xscale * yscale)
 
 
-def calc_rotation_matrix(angle, vparity=1):
+def calc_rotation_matrix(angle, v3i_yang, vparity=1):
     """ Calculate the rotation matrix.
 
     Parameters
@@ -148,10 +148,12 @@ def calc_rotation_matrix(angle, vparity=1):
        ----------------
 
     """
-    pc1_1 = vparity * np.cos(angle)
-    pc1_2 = np.sin(angle)
-    pc2_1 = vparity * -np.sin(angle)
-    pc2_2 = np.cos(angle)
+    rel_angle = angle - (vparity * v3i_yang)
+
+    pc1_1 = vparity * np.cos(rel_angle)
+    pc1_2 = np.sin(rel_angle)
+    pc2_1 = vparity * -np.sin(rel_angle)
+    pc2_2 = np.cos(rel_angle)
 
     return [pc1_1, pc1_2, pc2_1, pc2_2]
 
@@ -215,7 +217,11 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
         # Need to put the rotation matrix (List[float, float, float, float]) returned from calc_rotation_matrix into the
         # correct shape for constructing the transformation
         pc = np.reshape(
-            calc_rotation_matrix(np.deg2rad(refmodel.meta.wcsinfo.roll_ref), vparity=refmodel.meta.wcsinfo.vparity),
+            calc_rotation_matrix(
+                np.deg2rad(refmodel.meta.wcsinfo.roll_ref),
+                refmodel.meta.wcsinfo.v3yangle,
+                vparity=refmodel.meta.wcsinfo.vparity)
+            ,
             (2, 2)
         ).T
 
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -6,6 +6,7 @@ import logging
 import functools
 import numpy as np
 
+from astropy.coordinates import SkyCoord
 from astropy.utils.misc import isiterable
 from astropy.io import fits
 from astropy.modeling import models as astmodels
@@ -88,6 +89,70 @@ def reproject(wcs1, wcs2, origin=0):
     return _reproject
 
 
+def compute_scale_transform(refwcs, fiducial):
+    """Compute scaling transform
+    """
+    x0, y0 = refwcs.backward_transform(*fiducial)
+
+    x1 = x0 + 1
+    y1 = y0 + 1
+    ra0, dec0 = refwcs(x0, y0)
+    ra_xdir, dec_xdir = refwcs(x1, y0)
+    ra_ydir, dec_ydir = refwcs(x0, y1)
+
+    position0 = SkyCoord(ra=ra0, dec=dec0, unit='deg')
+    position_xdir = SkyCoord(ra=ra_xdir, dec=dec_xdir, unit='deg')
+    position_ydir = SkyCoord(ra=ra_ydir, dec=dec_ydir, unit='deg')
+
+    xscale = np.abs(position0.separation(position_xdir).value)
+    yscale = np.abs(position0.separation(position_ydir).value)
+    scale = np.sqrt(xscale * yscale)
+
+    return astmodels.Scale(scale) & astmodels.Scale(scale)
+
+
+def calc_rotation_matrix(angle, vparity=1):
+    """ Calculate the rotation matrix.
+
+    Parameters
+    ----------
+    angle : float in radians
+        The angle to create the matrix.
+
+    vparity : int
+        The x-axis parity, usually taken from
+        the JWST SIAF parameter VIdlParity.
+        Value should be "1" or "-1".
+
+    Returns
+    -------
+    matrix: [pc1_1, pc1_2, pc2_1, pc2_2]
+        The rotation matrix
+
+    Notes
+    -----
+    The rotation is
+
+       ----------------
+       | pc1_1  pc2_1 |
+       | pc1_2  pc2_2 |
+       ----------------
+
+    where:
+        pc1_1 = vparity * cos(angle)
+        pc1_2 = sin(angle)
+        pc2_1 = -1 * vparity * sin(angle)
+        pc2_2 = cos(angle)
+    """
+
+    pc1_1 = vparity * np.cos(angle)
+    pc1_2 = np.sin(angle)
+    pc2_1 = vparity * -np.sin(angle)
+    pc2_2 = np.cos(angle)
+
+    return [pc1_1, pc1_2, pc2_1, pc2_2]
+
+
 def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=None, domain=None):
     """
     Create a WCS from a list of input data models.
@@ -122,17 +187,20 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
     """
     bb = bounding_box
     wcslist = [im.meta.wcs for im in dmodels]
+
     if not isiterable(wcslist):
         raise ValueError("Expected 'wcslist' to be an iterable of WCS objects.")
+
     if not all([isinstance(w, WCS) for w in wcslist]):
         raise TypeError("All items in wcslist are to be instances of gwcs.WCS.")
+
     if refmodel is None:
         refmodel = dmodels[0]
     else:
         if not isinstance(refmodel, DataModel):
             raise TypeError("Expected refmodel to be an instance of DataModel.")
 
-    fiducial = compute_fiducial(wcslist, bb)
+    fiducial = compute_fiducial(wcslist, bb)  # possibly retrieve from reference wcs
 
     prj = astmodels.Pix2Sky_TAN()
 
@@ -140,29 +208,33 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
         transform = []
         wcsinfo = pointing.wcsinfo_from_model(refmodel)
         sky_axes, spec, other = gwutils.get_axes(wcsinfo)
-        rotation = astmodels.AffineTransformation2D(wcsinfo['PC'])
+
+        pc = np.reshape(
+            calc_rotation_matrix(refmodel.meta.wcsinfo.roll_ref, vparity=refmodel.meta.wcsinfo.vparity), (2, 2)
+        ).T
+        rotation = astmodels.AffineTransformation2D(pc)
         transform.append(rotation)
+
         if sky_axes:
-            cdelt1, cdelt2 = wcsinfo['CDELT'][sky_axes]
-            scale = np.sqrt(np.abs(cdelt1 * cdelt2))
-            scales = astmodels.Scale(scale) & astmodels.Scale(scale)
-            transform.append(scales)
+            transform.append(compute_scale_transform(refmodel.meta.wcs, fiducial))
 
         if transform:
             transform = functools.reduce(lambda x, y: x | y, transform)
 
     out_frame = refmodel.meta.wcs.output_frame
-    wnew = wcs_from_fiducial(fiducial, coordinate_frame=out_frame,
-                             projection=prj, transform=transform)
+    wnew = wcs_from_fiducial(fiducial, coordinate_frame=out_frame, projection=prj, transform=transform)
 
     footprints = [w.footprint().T for w in wcslist]
     domain_bounds = np.hstack([wnew.backward_transform(*f) for f in footprints])
+
     for axs in domain_bounds:
         axs -= axs.min()
+
     bounding_box = []
     for axis in out_frame.axes_order:
         axis_min, axis_max = domain_bounds[axis].min(), domain_bounds[axis].max()
         bounding_box.append((axis_min, axis_max))
+
     bounding_box = tuple(bounding_box)
     ax1, ax2 = np.array(bounding_box)[sky_axes]
     offset1 = (ax1[1] - ax1[0]) / 2
@@ -171,6 +243,7 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
 
     wnew.insert_transform('detector', offsets, after=True)
     wnew.bounding_box = bounding_box
+
     return wnew
 
 
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -12,7 +12,7 @@ from astropy.io import fits
 from astropy.modeling import models as astmodels
 from astropy.table import QTable
 from astropy.constants import c
-from typing import Union
+from typing import Union, List
 
 from gwcs import WCS
 from gwcs.wcstools import wcs_from_fiducial, grid_from_bounding_box
@@ -97,6 +97,7 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray]) -> float:
     ----------
     wcs :
         Reference WCS object from which to compute a scaling factor.
+
     fiducial :
         Input fiducial of (RA, DEC) used in calculating reference points.
 
@@ -111,7 +112,7 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray]) -> float:
 
     crpix = np.array(wcs.backward_transform(*fiducial))
     crpix_with_offsets = np.vstack((crpix, crpix + (1, 0), crpix + (0, 1))).T
-    crval_with_offsets = wcs(*crpix_with_offsets, with_bounding_box=False)
+    crval_with_offsets = wcs(*crpix_with_offsets)
 
     coords = SkyCoord(ra=crval_with_offsets[0], dec=crval_with_offsets[1], unit="deg")
     xscale = np.abs(coords[0].separation(coords[1]).value)
@@ -120,17 +121,19 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray]) -> float:
     return np.sqrt(xscale * yscale)
 
 
-def calc_rotation_matrix(angle, v3i_yang, vparity=1):
-    """ Calculate the rotation matrix.
+def calc_rotation_matrix(angle: float, v3i_yang: float, vparity: int = 1) -> List[float]:
+    """Calculate the rotation matrix.
 
     Parameters
     ----------
-    angle : float in radians
-        The angle to create the matrix.
+    angle :
+        The angle in radians to create the matrix.
+
+    v3i_yang :
+        The angle between ideal Y-axis and V3 in radians.
 
     vparity : int
-        The x-axis parity, usually taken from
-        the JWST SIAF parameter VIdlParity.
+        The x-axis parity, usually taken from the JWST SIAF parameter VIdlParity.
         Value should be "1" or "-1".
 
     Returns
@@ -148,6 +151,9 @@ def calc_rotation_matrix(angle, v3i_yang, vparity=1):
        ----------------
 
     """
+    if vparity not in (1, -1):
+        raise ValueError(f'vparity should be 1 or -1. Input was: {vparity}')
+
     rel_angle = angle - (vparity * v3i_yang)
 
     pc1_1 = vparity * np.cos(rel_angle)
@@ -205,7 +211,8 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
         if not isinstance(refmodel, DataModel):
             raise TypeError("Expected refmodel to be an instance of DataModel.")
 
-    fiducial = compute_fiducial(wcslist, bb)  # possibly retrieve from reference wcs
+    fiducial = compute_fiducial(wcslist, bb)
+    ref_fiducial = compute_fiducial([refmodel.meta.wcs])
 
     prj = astmodels.Pix2Sky_TAN()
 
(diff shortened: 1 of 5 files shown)


jwhite3:
Currently some regression tests are failing since the values used in computing the scaling and the PC matrix are slightly different, which results in keyword differences. 

In addition, a couple of tests in `test_nircam_image` are failing due to the bounding box that's created in `wcs_from_footprints` being full of `NaNs`, but I'm not sure why that is

jwst/assign_wcs/util.py
```diff
@@ -122,47 +187,54 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
...
     prj = astmodels.Pix2Sky_TAN()
 
     if transform is None:
         transform = []
         wcsinfo = pointing.wcsinfo_from_model(refmodel)
         sky_axes, spec, other = gwutils.get_axes(wcsinfo)
-        rotation = astmodels.AffineTransformation2D(wcsinfo['PC'])
+
+        pc = np.reshape(
+            calc_rotation_matrix(refmodel.meta.wcsinfo.roll_ref, vparity=refmodel.meta.wcsinfo.vparity), (2, 2)
```
nden, line 213: `roll_ref` is in deg. Do you need to convert to rad?

jwst/assign_wcs/util.py
```diff
@@ -88,6 +89,70 @@ def _reproject(x, y):
...
+def compute_scale_transform(refwcs, fiducial):
+    """Compute scaling transform
+    """
+    x0, y0 = refwcs.backward_transform(*fiducial)
+
+    x1 = x0 + 1
+    y1 = y0 + 1
+    ra0, dec0 = refwcs(x0, y0)
+    ra_xdir, dec_xdir = refwcs(x1, y0)
+    ra_ydir, dec_ydir = refwcs(x0, y1)
```
nden, line 101: Could you put the inputs in an array and do the transformations once?

jwst/assign_wcs/util.py
```diff
@@ -88,6 +89,70 @@ def _reproject(x, y):
...
+       | pc1_1  pc2_1 |
+       | pc1_2  pc2_2 |
+       ----------------
+
+    where:
+        pc1_1 = vparity * cos(angle)
+        pc1_2 = sin(angle)
+        pc2_1 = -1 * vparity * sin(angle)
+        pc2_2 = cos(angle)
+    """
```
nden, line 146: The order of the matrix in the comments is `pc11, pc21, ...`, and normally it's `pc11, pc12,...`.
It's probably best to remove the comments unless there's a special meaning. They are a copy of the code.

jwst/assign_wcs/util.py
jwhite3: I did not! I'll fix that

jwst/assign_wcs/util.py
```diff
@@ -88,6 +89,70 @@ def _reproject(x, y):
     return _reproject
 
 
+def compute_scale_transform(refwcs, fiducial):
+    """Compute scaling transform
```
jdavies-st, line 93: Could you add a docstring indicating what `refwcs` and `fiducial` are?  I think you could probably rename `refwcs` to be just `wcs` and indicate it is a `gwcs.WCS` object.  I believe `fiducial` is a tuple of (RA, Dec)?

jwst/assign_wcs/util.py
```diff
@@ -88,6 +89,70 @@ def _reproject(x, y):
...
+
+    position0 = SkyCoord(ra=ra0, dec=dec0, unit='deg')
+    position_xdir = SkyCoord(ra=ra_xdir, dec=dec_xdir, unit='deg')
+    position_ydir = SkyCoord(ra=ra_ydir, dec=dec_ydir, unit='deg')
+
+    xscale = np.abs(position0.separation(position_xdir).value)
+    yscale = np.abs(position0.separation(position_ydir).value)
+    scale = np.sqrt(xscale * yscale)
+
+    return astmodels.Scale(scale) & astmodels.Scale(scale)
```
jdavies-st, line 111: Also, you may just want the function to return the `scale` itself, and convert that to Astropy models in the calling function.  This gives this function and `calc_rotation_matrix` a symmetry and makes them more generally useable for both `gwcs.WCS` and `astropy.wcs.WCS` perhaps?

jwst/assign_wcs/util.py
jwhite3: Actually I'm not sure. For computing the "initial" points, I think that makes sense, but is there a way to do that with the `WCS` object without splitting an "initial point" array and "delta points" array to provide the appropriate inputs (x1, y0 and x0, y1)? I'm not very familiar with the `WCS` API yet, so if you had something in mind, let me know!

jwst/assign_wcs/util.py
jwhite3: I like this

jwst/assign_wcs/util.py
jdavies-st: Something like this can work:
```
crpix = refwcs.backward_transform(*fiducial)
crpix_with_offsets = np.vstack((crpix, crpix + (0, 1), crpix + (1, 0))).T
crval_with_offsets = refwcs(*crpix_with_offsets)

coords = SkyCoord(ra=crval_with_offsets[0], dec=crval_with_offsets[1], unit="deg")
xscale = np.abs(coords[0].separation(coords[1]).value)
yscale = np.abs(coords[0].separation(coords[2]).value)
```

jdavies-st: The wcs can operate on a numpy array.  So `wcs(x, y)` where x and y are each arrays of x values and y values.  That `vstack` and transpose manipulate the array to have that form so it can be indexed by 0 and 1.  This assumes that `fiducial` is a tuple of (RA, Dec), which is I think correct for this function?

jwhite3: Got it, thanks @jdavies-st 

Also, yes, I believe that's correct

jdavies-st: Oh, it looks like I flipped `xscale` and `yscale` above.  WCS takes x first then y, so an offset (0, 1) is an offset in y, not x.

jdavies-st:
Yeah, not exactly sure why we're seeing a bounding box failure on the NIRCam calwebb_image3 pipeline regression test.  For the single-resampled `_i2d` file at the end of calwebb_image2 there is no problem.  🤔 

The failure is in `outler_detection` which is only run in calwebb_image3.  But not clear why the bounding box is returning NaNs.

```
____________________________________________ ERROR at setup of test_nircam_image_stage3_i2d _____________________________________________

run_image2pipeline = None
rtdata_module = {'input': '/Users/jdavies/scratch/test_nircam_image_run_detector1pipeline0/jw42424-o002_20191220t214154_image3_001_asn...',
 'truth_remote': 'jwst-pipeline/dev/truth/test_nircam_image_stages/jw42424001001_01101_00001_nrca5_assign_wcs.fits'}
jail = local('/Users/jdavies/scratch/test_nircam_image_run_detector1pipeline0')

    def run_image3pipeline(run_image2pipeline, rtdata_module, jail):
        """Run calwebb_image3 on NIRCam imaging long data"""
        rtdata = rtdata_module
        # Grab rest of _rate files for the asn and run image2 pipeline on each to
        # produce fresh _cal files for the image3 pipeline.  We won't check these
        # or look at intermediate products, including the resampled i2d
        rate_files = [
        "nircam/image/jw42424001001_01101_00001_nrcb5_rate.fits",
        "nircam/image/jw42424001001_01101_00002_nrca5_rate.fits",
        "nircam/image/jw42424001001_01101_00002_nrcb5_rate.fits",
        "nircam/image/jw42424001001_01101_00003_nrca5_rate.fits",
        "nircam/image/jw42424001001_01101_00003_nrcb5_rate.fits",
        ]
        for rate_file in rate_files:
            rtdata.get_data(rate_file)
            args = ["config/calwebb_image2.cfg", rtdata.input,
                "--steps.resample.skip=True"]
            Step.from_cmdline(args)
    
        # Get the level3 assocation json file (though not its members) and run
        # image3 pipeline on all _cal files listed in association
        rtdata.get_data("nircam/image/jw42424-o002_20191220t214154_image3_001_asn.json")
        args = ["config/calwebb_image3.cfg", rtdata.input,
            # Comment out following lines, as the dataset is currently broken
            # "--steps.tweakreg.save_results=True",
            # "--steps.skymatch.save_results=True",
            "--steps.source_catalog.snr_threshold=20",
            ]
>       Step.from_cmdline(args)

../../dev/jwst/jwst/regtest/test_nircam_image.py:72: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../dev/jwst/jwst/stpipe/step.py:156: in from_cmdline
../../dev/jwst/jwst/stpipe/cmdline.py:346: in step_from_cmdline
    step.run(*positional)
../../dev/jwst/jwst/stpipe/step.py:392: in run
../../dev/jwst/jwst/pipeline/calwebb_image3.py:83: in process
    input_models = self.outlier_detection(input_models)
../../dev/jwst/jwst/stpipe/step.py:392: in run
../../dev/jwst/jwst/outlier_detection/outlier_detection_step.py:167: in process
    step.do_detection()
../../dev/jwst/jwst/outlier_detection/outlier_detection.py:190: in do_detection
    sdriz = resample.ResampleData(self.input_models, single=True,
../../dev/jwst/jwst/resample/resample.py:55: in __init__
../../dev/jwst/jwst/resample/resample_utils.py:47: in make_output_wcs
    output_wcs.data_size = shape_from_bounding_box(output_wcs.bounding_box)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

bounding_box = ((nan, nan), (nan, nan))

    def shape_from_bounding_box(bounding_box):
        """ Return a numpy shape based on the provided bounding_box
        """
        size = []
        for axs in bounding_box:
            delta = axs[1] - axs[0]
>           size.append(int(delta + 0.5))
E           ValueError: cannot convert float NaN to integer

../../dev/jwst/jwst/resample/resample_utils.py:63: ValueError
```

jwhite3:
Note that this PR will update some of the comparison values in `test_set_telescope_pointing.py`. This is due to different angles being used in calculations resulting in different values than before.

Also note that for `update_wcs_from_telem` I did not update the function to use `roll_ref` because it looks like that value is calculated from other parameters (including `pa_v3`), but let me know if that's not correct

jwst/assign_wcs/util.py
```diff
@@ -122,47 +192,63 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
...
         raise TypeError("All items in wcslist are to be instances of gwcs.WCS.")
+
     if refmodel is None:
         refmodel = dmodels[0]
     else:
         if not isinstance(refmodel, DataModel):
             raise TypeError("Expected refmodel to be an instance of DataModel.")
 
-    fiducial = compute_fiducial(wcslist, bb)
+    fiducial = compute_fiducial(wcslist, bb)  # possibly retrieve from reference wcs
```
jdavies-st, line 208: What is this comment here?  Not sure the meaning or context.

jwst/lib/set_telescope_pointing.py
```diff
@@ -325,10 +330,17 @@ def update_wcs_from_fgs_guiding(model, default_pa_v3=0.0, default_vparity=1):
...
+    except AttributeError:
+        logger.warning(f'Keyword "V3I_YANG" not found. Using {default_v3yangle} as default value.')
+
+        v3i_yang = default_v3yangle
+
     (model.meta.wcsinfo.pc1_1,
      model.meta.wcsinfo.pc1_2,
      model.meta.wcsinfo.pc2_1,
-     model.meta.wcsinfo.pc2_2) = calc_rotation_matrix(pa_rad, vparity=vparity)
+     model.meta.wcsinfo.pc2_2) = calc_rotation_matrix(pa_rad, v3i_yang, vparity=vparity)
```
jdavies-st, line 343: `v3i_yang` also needs to be converted to radians before being sent along to `calc_rotation_matrix()`

jwst/lib/set_telescope_pointing.py
```diff
@@ -441,7 +453,7 @@ def update_wcs_from_telem(
          model.meta.wcsinfo.pc1_2,
          model.meta.wcsinfo.pc2_1,
          model.meta.wcsinfo.pc2_2) = calc_rotation_matrix(
-             wcsinfo.pa * D2R, vparity=siaf.vparity
+             wcsinfo.pa * D2R, model.meta.wcsinfo.v3yangle, vparity=siaf.vparity
```
jdavies-st, line 456: Again, convert `v3yangle` to radians before sending it along to the function.  I think this probably explains the big differences seen below in some of the tests computing the PC matrix.

jwst/assign_wcs/util.py
```diff
@@ -111,7 +111,7 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray]) -> float:
 
     crpix = np.array(wcs.backward_transform(*fiducial))
     crpix_with_offsets = np.vstack((crpix, crpix + (1, 0), crpix + (0, 1))).T
-    crval_with_offsets = wcs(*crpix_with_offsets)
+    crval_with_offsets = wcs(*crpix_with_offsets, with_bounding_box=False)
```
jdavies-st, line 114: Perhaps instead of fixing this here, it would be better to fix it in whatever is calling this function in `outlier_detection` in the blotting code?  I suspect that it is using the wrong fiducial.  Instead of using the fiducial RA/Dec at the reference point of the distorted frame, it is using the RA/Dec reference point of the much larger rectified frame?

jwhite3: Good point!

jwst/lib/set_telescope_pointing.py
jwhite3: I've been forgetting that on all of the new angle args.. 🤦

jwst/assign_wcs/util.py
jwhite3: It was a note that I had from our conversation that I forgot to remove. Sorry about that

jwst/lib/set_telescope_pointing.py
```diff
@@ -441,7 +453,7 @@ def update_wcs_from_telem(
          model.meta.wcsinfo.pc1_2,
          model.meta.wcsinfo.pc2_1,
          model.meta.wcsinfo.pc2_2) = calc_rotation_matrix(
-             wcsinfo.pa * D2R, vparity=siaf.vparity
+             wcsinfo.pa * D2R, np.deg2rad(model.meta.wcsinfo.v3yangle), vparity=siaf.vparity
```
jdavies-st, line 456: Since `D2R` is likely almost exactly the same as `np.deg2rad`, could we use one or the other here?

jwst/lib/tests/test_set_telescope_pointing.py
```diff
@@ -243,10 +243,10 @@ def test_add_wcs_default(data_file):
...
     assert np.isclose(model.meta.wcsinfo.cdelt1, 3.0555555e-5)
     assert np.isclose(model.meta.wcsinfo.cdelt2, 3.0555555e-5)
-    assert np.isclose(model.meta.wcsinfo.pc1_1, -1.0)
-    assert np.isclose(model.meta.wcsinfo.pc1_2, 0.0)
-    assert np.isclose(model.meta.wcsinfo.pc2_1, 0.0)
-    assert np.isclose(model.meta.wcsinfo.pc2_2, 1.0)
+    assert np.isclose(model.meta.wcsinfo.pc1_1, -0.7431448254773942)
+    assert np.isclose(model.meta.wcsinfo.pc1_2, 0.6691306063588582)
+    assert np.isclose(model.meta.wcsinfo.pc2_1, 0.6691306063588582)
+    assert np.isclose(model.meta.wcsinfo.pc2_2, 0.7431448254773942)
```
jdavies-st, line 249: These changes in the PC matrix seem pretty large for what is expected to be a pretty small change in the pointing.  The old pointing was essentially no rotation on the sky, and this new one shows ~42 degrees rotation.  Same for following 3 tests.  Do we understand this?

jwst/lib/set_telescope_pointing.py
```diff
@@ -304,14 +308,15 @@ def update_wcs_from_fgs_guiding(model, default_pa_v3=0.0, default_vparity=1):
...
     except AttributeError:
         logger.warning(
-            'Keyword `PA_V3` not found. Using {} as default value'.format(
-                default_pa_v3
+            'Keyword `ROLL_REF` not found. Using {} as default value'.format(
+                default_roll_ref
             )
         )
-        pa = default_pa_v3
+        pa = default_roll_ref
```
jdavies-st: Are these doing the same thing?  It looks like `pa` is getting populated regardless of the `except` block, then also getting populated in the `except` block.

Also, if `pa` is the position angle at the aperture reference point, i.e `roll_ref`, the variable should just be called `roll_ref` to avoid confusion.

jwst/assign_wcs/util.py
```diff
@@ -88,7 +90,81 @@ def _reproject(x, y):
...
+
+    return np.sqrt(xscale * yscale)
+
+
+def calc_rotation_matrix(angle: float, v3i_yang: float, vparity: int = 1) -> List[float]:
+    """Calculate the rotation matrix.
+
+    Parameters
+    ----------
+    angle :
```
jdavies-st, line 129: It might be worth either calling this `angle` `roll_ref` instead, or indicating that that is the angle that one needs here.  I.e. the position angle of the V3 axis of the telescope translated to the aperture reference point.

From the JWST keyword dictionary for ROLL_REF:

"Telescope roll angle of V3 North over East at the ref. point (deg)"

jwst/lib/set_telescope_pointing.py
jwhite3: I was running into a case where `model.meta.wcsinfo.roll_ref` was returning `None` rather than raising an attribute error, which then caused errors elsewhere

jwhite3:
Also, just to note (seeing all of the additional commits) as I was working, I rebased this branch on to master so that I could work through some of the regression test failures locally due to other changes. However, after updating this branch I saw that the regression tests that I was running were now failing due to a crds error (possibly due to an update to crds as well?): 

```
crds.core.exceptions.CrdsLookupError: Error determining best reference for 'apcorr'  =   Unknown reference type 'apcorr'
```

jwst/lib/set_telescope_pointing.py
jdavies-st: Yeah, it should always return `None` if it is not populated, as it is in the schema for all datamodels.  I.e. it should never raise `AttributeError`, though I suppose one can't be too careful?

jwhite3: Hmmm, well the reason that I implemented it in this way is because I was trying to stick with what was previously done with the other values (in the try/except blocks). Perhaps those should be updated as well?

jwst/lib/tests/test_set_telescope_pointing.py
jwhite3: For these unit tests I think the `v3yangle` is 42 deg across the board, so I think that that accounts for it?

nden:
This needs to be rebased to remove the extra changes from other PRs before it can be reviewed.

jwhite3:
@nden This happened exactly because I rebased onto master locally (which I did again and had to resolve some conflicts in the most recent set of commits) and then pushed. I'm not sure what's happening at this point

jwhite3:
@nden I figured out what the issue was, and it should be all fixed now

jwst/lib/set_telescope_pointing.py
```diff
@@ -343,15 +347,16 @@ def update_wcs_from_fgs_guiding(model, default_pa_v3=0.0, default_vparity=1):
 
     # Get position angle
     try:
-        pa = model.meta.wcsinfo.pa_v3
+        roll_ref = model.meta.wcsinfo.roll_ref if model.meta.wcsinfo.roll_ref is not None else default_roll_ref
```
nden, line 350: I am missing where `default_roll_ref` is defined?

jwst/assign_wcs/util.py
```diff
@@ -122,47 +198,63 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
...
+
+        # Need to put the rotation matrix (List[float, float, float, float]) returned from calc_rotation_matrix into the
+        # correct shape for constructing the transformation
+        pc = np.reshape(
+            calc_rotation_matrix(
+                np.deg2rad(refmodel.meta.wcsinfo.roll_ref),
+                np.deg2rad(refmodel.meta.wcsinfo.v3yangle),
+                vparity=refmodel.meta.wcsinfo.vparity),
+            (2, 2)
+        ).T
```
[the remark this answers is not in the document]
jwhite3: The result of doing `np.reshape(calc_rotation_matrix_results), (2,2))` with the order of the results being `[pc1_1, pc1_2, pc2_1, pc2_2]` ends up being: 
```
[[pc1_1, pc1_2],
 [pc2_1, pc2_2]]
```

From the docstring in `calc_rotation_matrix`, I thought that the array should be:
```
[[pc1_1, pc2_1],
 [pc1_2, pc2_2]]
```
Which requires the transpose. It's totally possible that I misunderstood, so please let me know if this isn't right!

jwst/lib/set_telescope_pointing.py
jwhite3: It's an argument for the function (line 322)

jwst/assign_wcs/util.py
```diff
@@ -88,7 +90,81 @@ def _reproject(x, y):
     return _reproject
 
 
-def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=None, domain=None):
+def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray]) -> float:
+    """Compute scaling transform.
+
+    Parameters
+    ----------
+    wcs :
```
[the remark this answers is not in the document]
jwhite3: Ok, sure I can do that. I hesitated since I used annotations in the function definition (I thought that it would be redundant to specify again), but I can change that

jdavies-st:
While I think this is good for merging now, I think it would be nice to eventually write unit tests for `calc_rotation_matrix` and `compute_scale`.  I think these are pretty trivial, but would verify that they are doing what we think they should be.  Maybe pick some easy parameter space values for the rotation matrix, like 45 deg, 90 deg, 1 radian, etc.  And for scale, again pick easy geometric ratios like 45 deg triangles or 30-60 triangles.

jwst/assign_wcs/util.py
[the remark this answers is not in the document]
jdavies-st: Is the docstring in FITS/Fortran order?  Do we need C/Python order, i.e. rows, columns?  I don't know.  Any ways to double-check this @nden?

jdavies-st:
Oh, and to show that this PR fixes the problem originally found in our MIRI imaging regression test.  Results of resampled `_i2d` image from this branch on the left side.  Original on the right side.  It no longer has the unnecessary tilt.  🎉 

Screen Shot 2020-04-30 at 11 06 46 PM

jwst/assign_wcs/util.py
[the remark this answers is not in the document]
jwhite3: The first commit where `calc_rotation_matrix` occurs (that I could find) was created by @stscieisenhamer as a part of #1855 and includes the note with revision number 564cfeca2096c65681f197b7511c771e2fbdb5bb

@stscieisenhamer do you remember anything on this?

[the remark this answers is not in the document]
nden: The code in that commit looks correct, it's just that the comment is transposed. I think it was a simple typo in the comment and removing the transpose operation in this PR is correct.

PR #4787, opened 2020-04-14
JP-1389 Reduce detector1 pipeline memory footprint

ddavis-stsci:
Updated the detector1 pipeline to explicitly copy the variable with the data to a new name, close and then delete the old variable before calling the next step in the pipeline. Monitoring the amount of swap memory used is shows that this decreased significantly over the original pipeline processing.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -82,6 +82,8 @@ outlier_detection
 pipeline
 --------
 
+- Update ``calwebb_detector1`` to reduce the memory used in processing. [#4643]
+
 - Update ``calwebb_coron3`` to return ``ImageModel`` instead of ``DrizProductModel``,
   when necessary. [#4552]
 
--- a/jwst/pipeline/calwebb_detector1.py
+++ b/jwst/pipeline/calwebb_detector1.py
@@ -76,104 +76,45 @@ class Detector1Pipeline(Pipeline):
             # the steps are in a different order than NIR
             log.debug('Processing a MIRI exposure')
 
-            input1 = self.group_scale(input)
-            input1 = input.copy()
-            input.close()
-            del input
-            input1 = self.dq_init(input1)
-            input2 = input1.copy()
-            input1.close()
-            del input1
-            input2 = self.saturation(input2)
-            input3 = input2.copy()
-            input2.close()
-            del input2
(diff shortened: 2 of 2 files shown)


jdavies-st:
The only thing complicating my above notes is if there's still a reference to a datamodel within a step and it is assign to `self`, i.e. an instance of the step.  If so, then that datamodel (and its allocated memory) will persist after the step's `process()` method finishes, and it will persist throughout the processing of the detector1 pipeline, because the first thing `Detector1Pipeline` does is create instances of each step and assign them to a class variable.  So the datamodel would persist because the step persists within the pipeline.

https://github.com/spacetelescope/jwst/blob/97adef48fc100af2cb2e1c7e7a14a6eb80f336e2/jwst/pipeline/calwebb_detector1.py#L44-L59

If a step is doing this, then it would be important to set that class variable, say something like `self.input_datamodel` to `None` before exiting `process()` so it doesn't persist for the whole pipeline.  Are any of the steps in this pipeline doing that?

hbushouse:
@jdavies-st regarding your comment about the step instances persisting until the whole pipeline shuts down (which might be part of our problem of memory not being freed), would it help to somehow delete each step instance (in the pipeline module) immediately after it's been called (i.e. before calling the next step)?

jdavies-st:
It could, but that would have unintended consequences.  Better to make sure each step is doing the correct thing.

1. Identify which step is hanging on to lots of memory.

2. See if that step class has any instance variables (`self.foo`) pointing to large datamodels or `numpy` arrays.

3. Change that behavior if it is not needed, or make sure those instance variables are set to `None` after they are no longer needed.

hbushouse:
Do you mean stuff like this:
https://github.com/spacetelescope/jwst/blob/master/jwst/refpix/reference_pixels.py#L147
where a step assigns a big block of data to the step instance itself?

jdavies-st:
Yup, exactly like that.  That array will persist until `Detector1Pipeline` instance gets destroyed.

ddavis-stsci:
and this is cleaned up like in jump_step with

             gain_model.close()
             readnoise_model.close()

hbushouse:
When you think you have a version that's ready for an initial merge, reset the requests for approval and we'll get it through the mill.

PR #4789, opened 2020-04-15
Turn off pytest-openfiles plugin in setup.cfg

jdavies-st:
We need to turn this off for now until we can debug why our regression tests are leaving behind open files when they fail via `FITSDiff`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ results_root = jwst-pipeline-results
 doctest_plus = true
 doctest_rst = true
 text_file_format = rst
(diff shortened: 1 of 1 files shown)


jdavies-st:
And the cause of this open file issue is that `astropy.io.fits.FITSDiff` opens files with `memmap=True` which then doesn't clean up references to the memmap object, so that `close()` doesn't actually close the file.  By design.

This is a known, documented problem in the `astropy.io.fits` library.  A solution will be to make `memmap=False` for `FITSDiff` by default.

PR #4790, opened 2020-04-15
Remove TPCorr

mcara:
2nd attempt to remove `TPCorr`. This is a re-incarnation of #4542 :

This PR removes `TPCorr` WCS correction model. This model is now a compound model defined in `tweakwcs` using fundamental models defined in `astropy` and `gwcs`.


mcara:
@hbushouse No changes. There were changes due to a mistake in `tweakwcs v0.6.2` but in the latest release `0.6.3` that bug should have been fixed and therefore I expect no changes at all. Hopefully, @jdavies-st or @nden can confirm this from regression tests.

hbushouse:
So is it OK to merge this now? Everything we need to support it (e.g. gwcs) is already in place?

nden:
Yes, I believe so. Both gwcs and tweakwcs are on pypi.

mcara:
Everything should be on `pypi`. I also run regression test for `miri` and it had the same failures as before but there was no change in data. With the older version of the `tweakwcs` (version `0.6.2`) there was a difference in alignment and hence in data of the resampled images. Not with the latest version.

However, can this wait until tomorrow to see that all regression tests pass?

mcara:
> it had the same failures as before

I mean that the tests even if they are failing, they are failing in exactly the same way as before when it was using `TPCorr` and `tweakwcs==0.5.3`. I mean that I do not see any meaningful difference in the failures.

hbushouse:
Yes, our nightly regtest run is going now, so I'll wait till that's done to merge this.

PR #4815, opened 2020-04-15
Ramp multiprocessing

mwregan2:
This is a working version of adding multiprocessing to the ramp step. Because of the inability of multiprocessing to pickle the various data models, I had to change the code to pass the various arrays and header values as explicit parameters.
There are also a couple of minor changes. The power value for the optimal fitting is now interpolated to avoid discontinuities in the slopes at the SNR boundaries. Also, most of the code was moved to double precision to avoid rounding errors. I did this when trying to track down why I was getting different results from two tests with the same input parameters. In the end it was because the input read noise was being changed. These changes could be backed out but I think defaulting to double precision is a good practice. 
Testing on the ODL server showed a speed up of a factor of 13 (27 min -> 2 min) for a large MIRSIM multi-int exposure and a factor of 17 for a small one. 
More tests are needed and the doc strings need to be updated.


dmggh:
I notice there are no updates to gls_fit.py, which contains the majority of generalized least squares routines. A very general question - has this PR been tested on the GLS algorithm ?

mwregan2:
David,
  I haven’t tested it on that. I’ll update the code to have the current version and make the changes.
Mike

mwregan2:
The latest updates have all the unit tests working and I've added multiprocessing to the GLS option. GLS seems to benefit from multiprocessing even more than the baseline OLS method.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -105,9 +118,286 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
 
     return new_model, int_model, opt_model, gls_opt_model
 
+def ols_ramp_fit_multi(input_model, buffsize, save_opt, readnoise_2d, gain_2d,
+                 weighting, max_cores):
+    """
+       Setup the inputs to ols_ramp_fit with and without multiprocessing. The input inputs will be sliced into the
```
hbushouse, line 124: typo: "input inputs"

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -105,9 +118,286 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+       number of cores that are being used for multiprocessing. Because the data models cannot be pickeled, only
```
hbushouse, line 125: typo: pickeled -> pickled

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -105,9 +118,286 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+
+       gls_opt_model : GLS_RampFitModel object or None
+           Object containing optional GLS-specific ramp fitting data for the
+           exposure
+       """
+    if max_cores == 'none':
+        number_slices = 1
+    else:
+        num_cores = multiprocessing.cpu_count()
+        log.debug("Found %d possible cores to use for ramp fitting " % num_cores)
```
hbushouse, line 179: future proof using `log.debug(f'Found {num_cores} possible cores to use for ramp fitting')`

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -105,9 +118,286 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+            opt_model.var_poisson = opt_var_poisson
+            opt_model.var_rnoise = opt_var_rnoise
+            opt_model.yint = opt_yint
+            opt_model.sigyint = opt_sigyint
+            opt_model.pedestal = opt_pedestal
+            opt_model.weights = opt_weights
+            opt_model.crmag = opt_crmag
+        return out_model, int_model, opt_model
+    else:
+        log.debug("number of processes being used is %d" % number_slices)
```
hbushouse, line 229: update to `log.debug(f'number of processes being used is {number_slices}')`

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -105,9 +118,286 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+        groupdq_slice = input_model.groupdq[:, :, (number_slices - 1) * rows_per_slice: total_rows, :].copy()
+        pixeldq_slice = input_model.pixeldq[(number_slices - 1) * rows_per_slice: total_rows, :].copy()
+        slices.insert(number_slices - 1, (data_slice, err_slice, groupdq_slice, pixeldq_slice, buffsize, save_opt,
+                                          readnoise_slice, gain_slice, weighting,
+                                          input_model.meta.instrument.name, input_model.meta.exposure.frame_time,
+                                          input_model.meta.exposure.ngroups, input_model.meta.exposure.group_time,
+                                          input_model.meta.exposure.groupgap, input_model.meta.exposure.nframes,
+                                          input_model.meta.exposure.drop_frames1, int_times))
+
+        log.debug("Creating %d processes for ramp fitting " % number_slices)
```
hbushouse, line 262: update to `log.debug(f'Creating {number_slices} processes for ramp fitting')`

mwregan2:
Howard,
  I saw that but I don’t understand it. I’ll look into how to get them out of the pull request.
Mike

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -105,9 +118,286 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+        # Create new model for the primary output.
+        actual_segments = real_results[0][20]
+        actual_CRs = real_results[0][21]
+        int_model, opt_model, out_model = create_output_models(input_model,
+                            number_of_integrations, save_opt, total_cols, total_rows,
+                                                               actual_segments, actual_CRs)
+        # iterate over the number of slices and place the results into the output models
+        for resultslice in real_results:
+            if len(real_results) == k + 1:  # last result
+                out_model.data[k * rows_per_slice: total_rows, :] = resultslice[0]
```
dmggh, line 278: Perhaps mapping the index to resultslice something more descriptive would help clarify this.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -242,26 +582,32 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
     if np.all(np.bitwise_and( first_gdq, dqflags.group['SATURATED'] )):
-        new_model, int_model, opt_model = utils.do_all_sat( model, imshape,
-                                                            n_int, save_opt )
+        new_model, int_model, opt_model = utils.do_all_sat(inpixeldq, groupdq, imshape, n_int, save_opt )
+        if (save_opt):
+            actual_segments = 0
+            actual_CRs = 0
 
-        return new_model, int_model, opt_model
+        return new_model.data, new_model.dq, new_model.var_poisson, new_model.var_rnoise, new_model.err, \
```
dmggh, line 604: These 20+ values returned could be consolidated, if the 3 separate models were returned.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -242,26 +582,32 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+           int_model.data, int_model.dq, int_model.var_poisson, int_model.var_rnoise, int_model.err, int_model.int_times, \
+           opt_model.slope, opt_model.sigslope, opt_model.var_poisson, opt_model.var_rnoise, opt_model.yint, opt_model.sigyint, \
+           opt_model.pedestal, opt_model.weights, opt_model.crmag, actual_segments, actual_CRs
 
     # Get max number of segments fit in all integrations
-    max_seg = calc_num_seg(gdq_cube, n_int)
+    max_seg, num_CRs = calc_num_seg(gdq_cube, n_int)
```
dmggh, line 610: Is num_CRs used ?

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -755,14 +1086,53 @@ def ols_ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
...
+        opt_var_rnoise = None
+        opt_yint = None
+        opt_sigyint = None
+        opt_pedestal = None
+        opt_weights = None
+        opt_crmag = None
+        actual_segments = 0
+        actual_CRs = 0
+
+    return new_model.data, new_model.dq, new_model.var_poisson, new_model.var_rnoise, new_model.err, \
```
dmggh, line 1128: Similar comment as above to consolidate.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -807,172 +1177,150 @@ def gls_ramp_fit(model,
...
+            pedestal_int = np.zeros((n_int,) + imshape, dtype=np.float32)
+            # If there are no cosmic rays, set the last axis length to 1.
+            shape_ampl = (n_int, imshape[0], imshape[1], max(1, max_num_cr))
+            ampl_int = np.zeros(shape_ampl, dtype=np.float32)
+            ampl_err_int = np.zeros(shape_ampl, dtype=np.float32)
+
+##Loop over number of processes
+        for i in range(number_slices - 1):
+            readnoise_slice = readnoise_2d[i * rows_per_slice: (i + 1) * rows_per_slice, :]
+            gain_slice = gain_2d[i * rows_per_slice: (i + 1) * rows_per_slice, :]
```
dmggh, line 1265: Common indices duplicated: perhaps this can be simplified.

jwst/ramp_fitting/ramp_fit.py
mwregan2: fixed

jwst/ramp_fitting/ramp_fit.py
mwregan2: fixed

jwst/ramp_fitting/ramp_fit.py
mwregan2: changed

jwst/ramp_fitting/ramp_fit.py
mwregan2: changed

jwst/ramp_fitting/ramp_fit.py
mwregan2: I agree that returned this many values is awkward but the models cannot be returned directly. This is because the multiprocessing process uses pickle to pass the parameters to the parallel process and the data models cannot be pickled due to something odd in the ASDF format.
I spent a while trying alternative methods but the only thing that works is passing the numpy arrays in and out as parameters.

jwst/ramp_fitting/ramp_fit.py
mwregan2: It is not used on this call but is used in the call from the OLS fit section.

jwst/ramp_fitting/ramp_fit.py
mwregan2: I've updated the all of the result indices. If you thing the new way is clearer. I can update the input indices in a similar way

jwst/ramp_fitting/ramp_fit.py
dmggh: Hmmm. I will let you know if I have any ideas.

dmggh:
Yes, that seems clearer. Thanks.

jwst/ramp_fitting/ramp_fit.py
mwregan2: I have changed all the indices to use explicit variables to make it easier to read.

jwst/jump/jump.py
```diff
@@ -122,7 +122,10 @@ def detect_jumps (input_model, gain_model, readnoise_model,
     else:
         log.info("Creating %d processes for jump detection " % numslices)
         pool = multiprocessing.Pool(processes=numslices)
+        #Starts each slice in it's own process. Starmap allows more than one parameter to be passed.
```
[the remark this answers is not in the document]
mwregan2: I fixed that line and a few others in that file.

mwregan2:
Is there something that people are waiting for that I should do for the PR?

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -1936,6 +2465,7 @@ def fit_lines(data, mask_2d, rn_sect, gain_sect, ngroups, weighting):
     else: # unsupported weighting type specified
         log.error('FATAL ERROR: unsupported weighting type specified.')
 
+    slope_s[good_pix] = slope
     slope_s[good_pix] = slope
```
hbushouse, line 2469: It looks to me like these 2 lines (2468-2469) are identical. Why the duplication?

PR #4816, opened 2020-04-15
Update flatfield and linearity docs

### LINKED ISSUE #4524: Linearity correction docs should state that MIR data is not bias subtracted
Issue [JP-1268](https://jira.stsci.edu/browse/JP-1268) was created by Alicia Canipe:

From [~kgordon]:
 * For the linearity correction, readthedocs should state that MIR data is not bias subtracted

### LINKED ISSUE #4526: Flat field readthedocs has NIRSpec info in non-NIRSpec section
Issue [JP-1271](https://jira.stsci.edu/browse/JP-1271) was created by Alicia Canipe:

From [~kgordon] :
 * The readthedocs for the flat field correction has NIRSpec info in a non-NIRspec section.

hbushouse:
Update the flat-field docs to move/remove old NIRSpec info from the non-NIRSpec section of the docs. Also updated the name of the argument used to save the on-the-fly flat as used in the code. Update the linearity docs to make it explicit that MIRI data do not get bias subtracted before doing linearity correction (or ever).

Fixes #4524 / [JP-1268](https://jira.stsci.edu/browse/JP-1268)
Fixes #4526 / [JP-1271](https://jira.stsci.edu/browse/JP-1271)

### DIFF
--- a/docs/jwst/linearity/description.rst
+++ b/docs/jwst/linearity/description.rst
@@ -4,12 +4,16 @@ Description
 Assumptions
 -----------
 It is assumed that the input science exposure data for near-IR instruments
-have had the superbias subtraction applied, therefore the correction
-coefficients stored in the linearity reference files for those instruments
-must have been derived from data that also had the zero group subtracted.
+have had the :ref:`superbias <superbias_step>` subtraction step applied,
+therefore the correction coefficients stored in the linearity reference files
+for those instruments must have been derived from data that has also been
+bias subtracted.
+MIRI data, on the other hand, do not receive bias subtraction
+(see :ref:`calwebb_detector1 <calwebb_detector1>`) and hence the linearity
+correction is derived and applied in the raw data space.
 
 It is also assumed that the saturation step has already been applied to
-the science data, so that saturation flags are set in the GROUPDQ array of
+the input data, so that saturation flags are set in the GROUPDQ array of
 the input science data.
 
 Algorithm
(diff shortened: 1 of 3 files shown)


docs/jwst/linearity/description.rst
```diff
@@ -4,12 +4,16 @@ Description
...
-have had the superbias subtraction applied, therefore the correction
-coefficients stored in the linearity reference files for those instruments
-must have been derived from data that also had the zero group subtracted.
+have had the :ref:`superbias <superbias_step>` subtraction step applied,
+therefore the correction coefficients stored in the linearity reference files
+for those instruments must have been derived from data that has also been
+bias subtracted.
+MIRI data, on the other hand, do not receive bias subtraction
+(see :ref:`calwebb_detector1 <calwebb_detector1>`) and hence the linearity
+correction is derived and applied in the raw data space.
```
karllark, line 13: The part that could cause confusion is stating that the correction is derived and applied in "raw data space."  What does "raw data space" mean in this context?  Maybe just replace this with the equivalent from the NIR text.  E.g., "..and hence the linearity correction is derived from data that has not been bias subtracted."

hbushouse: Agreed. I was having a hard figuring out the proper words to use to express something like "absolute DN" (as opposed to DN's above bias). I'll make the change you suggest.

hbushouse: Fixed in most recent commit.

hbushouse:
@karllark github seems to think this still needs approval, so I recycled the review request. Can you approve again?

hbushouse:
Oh, never mind. The problem is that you aren't a user with write access to the repo and hence your approval doesn't count (nothing personal). I'll get an approval from someone with authority.

karllark:
I knew that already.  No authority is the way I like it!  :-P

PR #4817, opened 2020-04-15
Turn off memmap for FITSDiff in regtests

jdavies-st:
This should solve our issues with FITSDiff leaving behind open files when a difference is found.

Also, this updates the patch versions of python used in TravisCI.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ results_root = jwst-pipeline-results
 doctest_plus = true
 doctest_rst = true
 text_file_format = rst
-addopts = --show-capture=no
+addopts = --show-capture=no --open-files
 
 [bdist_wheel]
 # This flag says that the code is written to work on both Python 2 and Python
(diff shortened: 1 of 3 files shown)


hbushouse:
> This should solve our issues with FITSDiff leaving behind open files when a difference is found.

*should*?  Do we know for sure?

setup.cfg
```diff
@@ -37,7 +37,7 @@ results_root = jwst-pipeline-results
 doctest_plus = true
 doctest_rst = true
 text_file_format = rst
-addopts = --show-capture=no
+addopts = --show-capture=no --open-files
```
hbushouse, line 40: Oh boy, here we go again!  What could possibly go wrong?

jdavies-st:
Well it solves the known issue.  After being bit twice on this, I'm no longer speaking in absolutes.  😒

jdavies-st:
Btw, this is now fixed over in `astropy` as well.

https://github.com/astropy/astropy/pull/10159

PR #4842, opened 2020-04-21
JP-1403 Fix handling of the boolean-like environmental variables PASS_INVALID_VALUES and STRICT_VALIDATION

### LINKED ISSUE #4777: Environmental variables should not simply override class init arguments
Issue [JP-1403](https://jira.stsci.edu/browse/JP-1403) was created by Jonathan Eisenhamer:

In `jwst.datamodels.model_base.DataModel`, two environmentals, if set, override the following class arguments:
 * PASS_INVALID_VALUES overrides pass_invalid_values
 * STRICT_VALIDATION overrides strict_validation

In general, environmentals should not blindly override function/class arguments. Environmentals should only be used if the arguments have not been explicitly set. If an argument has been explicitly set, the environmental should not be used.

stscieisenhamer:
Resolves #4777 
Resolves [JP-1403](https://jira.stsci.edu/browse/JP-1403)

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -93,15 +93,14 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
             Additional arguments passed to lower level functions.
         """
 
-        # Override value of validation parameters
-        # if environment value set
+        # Override value of validation parameters if not explicitly set.
         if not pass_invalid_values:
-            pass_invalid_values = self.get_envar("PASS_INVALID_VALUES",
+            pass_invalid_values = self.get_envar_as_boolean("PASS_INVALID_VALUES",
                                                  False)
         self._pass_invalid_values = pass_invalid_values
         if not strict_validation:
-            strict_validation = self.get_envar("STRICT_VALIDATION",
-                                                     False)
+            strict_validation = self.get_envar_as_boolean("STRICT_VALIDATION",
+                                               False)
         self._strict_validation = strict_validation
         self._ignore_missing_extensions = ignore_missing_extensions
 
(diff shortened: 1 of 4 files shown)


jwst/datamodels/model_base.py
```diff
@@ -306,14 +313,31 @@ def close(self):
             if fd is not None:
                 fd.close()
 
-    def get_envar(self, name, value):
+    def get_envar_as_boolean(self, name, default=False):
```
jdavies-st, line 316: This should probably be a private method `_get_envar_as_boolean`.  I don't think we want it as part of the pubic API of the class.

jwst/datamodels/tests/test_models.py
```diff
@@ -702,3 +703,75 @@ def test_abvega_offset_model():
...
+        ('y', True, True),
+        ('Y', False, True),
+        ('Y', True, True),
+        ('yes', False, True),
+        ('yes', True, True),
+        ('Yes', False, True),
+        ('Yes', True, True),
+        ('YES', False, True),
+        ('YES', True, True),
+    ]
```
jdavies-st, line 763: Did you get all permutations?  😂 👍

jwst/datamodels/tests/test_models.py
```diff
@@ -702,3 +703,75 @@ def test_abvega_offset_model():
...
+
+@pytest.fixture(scope='module')
+def empty_model():
+    """Create an empy model"""
+    return DataModel()
+
+
+@pytest.fixture(scope='module')
+def jail_environ():
+    """Lock cheanges to the environment"""
```
jwhite3, line 715:
```suggestion
    """Lock changes to the environment"""
```

jwst/datamodels/tests/test_models.py
stscieisenhamer: No, but there was actually good reason for these. One would be surprised...

The frustration level for this bit-o-code was higher than it really should have been.

stscieisenhamer:
Not sure why the jenkins build is hung; it reports done and good. Merging.

PR #4846, opened 2020-04-22
JP-1399: Reorganize spec3 data products for WFSS data

### LINKED ISSUE #4762: calwebb_spec3 product updates for WFSS and non-TSO SOSS
Issue [JP-1399](https://jira.stsci.edu/browse/JP-1399) was created by Alicia Canipe:

+*Update needed*+

Need to use a unique product type suffix for combine_1d (step 3) products, such as “c1d” (combined 1d). Then SCSB would need to change the behavior of extract_1d in step 2 to store all the results for a given source in a single file, still using the x1d product suffix.

 

+*Background from [~bushouse]*+

When spectroscopic data from any mode that does not allow for (or require) resampling/drizzling goes through the calwebb_spec3 pipeline, the combination of data is done on the 1-D extracted spectra, rather than first producing a combined 2-D image and then extracting from that. This includes NIRCam and NIRISS WFSS modes, as well as corner cases like non-TSO NIRISS SOSS (TSO modes go through calwebb_tso3). In these cases the only calwebb_spec3 steps that are applied are:

 
 # Reorganization of exposure-based cal files into source-based cal files
 # 1-D extraction (via extract_1d) of spectra from each instance in each source-based cal file, resulting in multiple x1d products for each source
 # Combination of the multiple 1-D spectra for each source (via combine_1d), resulting in a final x1d product for each source

 

Unfortunately the source-based products resulting from steps 2 and 3 both use the “x1d” product type suffix, so in order to avoid file name collisions the multiple results from step 2 currently use file names like:

    jw00307-o002_t001_niriss_clear-gr700xd_0_x1d.fiits

    jw00307-o002_t001_niriss_clear-gr700xd_1_x1d.fiits

    jw00307-o002_t001_niriss_clear-gr700xd_2_x1d.fiits

   …

    jw00307-o002_t001_niriss_clear-gr700xd_<n>_x1d.fiits

 

while the final product from step 3 uses file names of the form:

 

    jw00307-o002_t001_niriss_clear-gr700xd_x1d.fiits

 

This situation is bad for 2 reasons:
 # The extra numerical suffix in the step 2 product file names is non-standard and means SDP and Archive have to do special things to handle them.
 # Having spectra for the same source spread out across multiple files in the step 2 results is annoying for users. It would be much more convenient to have all the spectra for a given source stored in multiple extensions in a single file (as is done for TSO modes).

 

We can’t simply switch to storing all the multiple spectra for a given source in a single step 2 product, because then there’d be a name collision with the step 3 product.

 

Personally I think the most straight-forward approach to solving this would be to use a unique product type suffix for combine_1d (step 3) products, such as “c1d” (combined 1d). We could then change the behavior of extract_1d in step 2 to store all the results for a given source in a single file, still using the x1d product suffix.

hbushouse:
This modifies the type and organization of level-3 data products created by `calwebb_spec3` for WFSS observations. Specifically:

1) Instead of having `extract_1d` create individual `x1d` files for each instance of data for a source, it now creates a single `x1d` file that contains the multiple instances of 1-D spectra for the source.

2) The `combine_1d` step now uses a unique product suffix of `c1d` for its products, instead of `x1d`.

This PR also includes an update to the existing `test_niriss_wfss` regtest module to add a `calwebb_spec3` test (previously it only tested `calwebb_spec2`). Necessary input and truth files have been uploaded to artifactory.

Note that I've made some additional (benign) updates to the `extract_1d` code while implementing the necessary changes, because the data flow through the code is highly convoluted and not terribly well documented. So I've added some documentation here and there to hopefully start improving that situation.

Fixes #4762 / [JP-1399](https://jira.stsci.edu/browse/JP-1399)


hbushouse:
> Looks good.
> 
> The only thing that I'd say is we should really get away from using `input` as a variable (not just here) since that's also a python built-in function

Agreed. I changed "input" to "input_temp" (very original name) in the latest commit.

The latest commit also includes updates to all the relevant docs that pertain to x1d and c1d products and the calwebb_spec3 pipeline.

Issue #4760, opened 2020-04-01
Update set_telescope_pointing to populate Moving Target coords


stscijgbot:
Issue [JP-1398](https://jira.stsci.edu/browse/JP-1398) was created by Howard Bushouse:

The new "assign_mtwcs" step used in the level-3 imaging and spectroscopic pipelines relies on having the keywords MT_RA and MT_DEC present in the SCI extension header of moving target data products. The set_telescope_pointing script needs to be updated to populate these 2 keywords from the CRVAL1A, CRVAL2A keyword values that will be present in the Moving Target Position table extension header in level-1b products for moving targets. These keywords only get added/populated for exposures of moving targets. They will be absent for all others.

stscijgbot:
Comment by Howard Bushouse: Fixed by [https://github.com/spacetelescope/jwst/pull/4847]

PR #4847, opened 2020-04-22
JP-1398 Populate Moving Target coords in set_telescope_pointing

ddavis-stsci:
Update for set_telescope_pointing to add or update the MT_RA and MT_DEC fields for data with target type of 'Moving'. 

currently all tests pass but it does seem to be rather slow. 

```
pytest --bigdata --basetemp= ~/src/JWST/scsb/jwst/jwst/regtest/ -k test_nircam_mtimage
=== test session starts ===
platform darwin -- Python 3.7.4, pytest-5.3.5, py-1.8.1, pluggy-0.13.0
plugins: xdist-1.30.0, requests-mock-1.7.0, doctestplus-0.4.0, openfiles-0.4.0, ci-watson-0.4, asdf-2.5.1, forked-1.1.3, cov-2.8.1
collected 377 items / 373 deselected / 4 selected

../../scsb/jwst/jwst/regtest/test_nircam_mtimage.py ....                                                             [100%]
...
-- Docs: https://docs.pytest.org/en/latest/warnings.html
=== 4 passed, 373 deselected, 2 warnings in 1978.24s (0:32:58) ===

```

closes issue #4760 and JP-1398


jwst/lib/set_telescope_pointing.py
```diff
@@ -194,6 +195,10 @@ def add_wcs(filename, default_pa_v3=0., siaf_path=None, engdb_url=None,
         reduce_func=reduce_func,
         **transform_kwargs
     )
+
+    if model.meta.target.type.lower() == 'moving':
```
jdavies-st, line 199: `lower()` will fail if `meta.target.type` is `None`, i.e not set, which might be the case for some data.

jwst/lib/set_telescope_pointing.py
```diff
@@ -204,6 +209,37 @@ def add_wcs(filename, default_pa_v3=0., siaf_path=None, engdb_url=None,
...
+    """Add/update the Moving target header keywords
+
+    If the target type is "moving_target" check for the moving target position
+    table. If this is available calculate the moving target position keywords
+    and insert or update MT_RA & MT_DEC.
+    """
+
+    if model.extra_fits.hasattr('MOVING_TARGET_POSITION'):
+        time_mt = model.extra_fits.MOVING_TARGET_POSITION.data.time
+        exp_midpt_mjd = model.meta.exposure.mid_time #- 0.207
```
jdavies-st, line 222: What does `#- 0.207` mean?  It should be removed if not needed, or clarified.

jwst/lib/set_telescope_pointing.py
```diff
@@ -204,6 +209,37 @@ def add_wcs(filename, default_pa_v3=0., siaf_path=None, engdb_url=None,
...
 
+def update_mt_kwds(model):
+    """Add/update the Moving target header keywords
+
+    If the target type is "moving_target" check for the moving target position
+    table. If this is available calculate the moving target position keywords
+    and insert or update MT_RA & MT_DEC.
+    """
+
+    if model.extra_fits.hasattr('MOVING_TARGET_POSITION'):
```
jdavies-st, line 220: We should make access to this table not go through `extra_fits`.  Can we add this as an allowable table in the schema?

jwst/lib/set_telescope_pointing.py
jdavies-st: Oh, and this is actually why the unit tests are failing.  Check for `None` first.

ddavis-stsci:
Updated the tests to address James' comments. Now the tests run in less than 2 min!

CHANGES.rst
```diff
@@ -122,6 +122,11 @@ resample_spec
 
 - Enable model metadata blending [#4765]
 
+  set_telescope_pointing
```
hbushouse, line 125: Not sure, but sphinx might complain about the extra whitespace at the beginning of the line here (or the fact that that makes it extend past the end of the subsequent underline).

jwst/lib/set_telescope_pointing.py
```diff
@@ -194,6 +195,11 @@ def add_wcs(filename, default_pa_v3=0., siaf_path=None, engdb_url=None,
         reduce_func=reduce_func,
         **transform_kwargs
     )
+
+    if model.meta.target.type:
```
hbushouse, line 199: For safety's sake, should this have a check for None or AttributeError around it, just in case the input doesn't have the TARGTYPE keyword? That should be a pretty basic keyword (and hence in every "real" dataset coming through DMS), but you can never be too careful (IMHO).

jwst/lib/set_telescope_pointing.py
hbushouse: Agree, we should add the table definition to the level-1b schema and then access it through the data model.

jwst/lib/set_telescope_pointing.py
```diff
@@ -204,6 +210,37 @@ def add_wcs(filename, default_pa_v3=0., siaf_path=None, engdb_url=None,
...
+
+    if model.extra_fits.hasattr('MOVING_TARGET_POSITION'):
+        time_mt = model.extra_fits.MOVING_TARGET_POSITION.data.time
+        exp_midpt_mjd = model.meta.exposure.mid_time #- 0.207
+        # check to see if the midpoint of the observation is contained within
+        # the timerange of the MT table
+        if time_mt[0] <= exp_midpt_mjd <= time_mt[-1]:
+            ra = model.extra_fits.MOVING_TARGET_POSITION.data.moving_target_RA
+            dec = model.extra_fits.MOVING_TARGET_POSITION.data.moving_target_Dec
+            f_ra = interp1d(time_mt,ra)
```
hbushouse, line 229: Huh, surprised flake8 didn't complain about the lack of whitespace after the comma in the argument list ...

jdavies-st: Our `flake8` checks are lenient here because of all our `numpy` code that does things like `array[0,1,2]` instead of `array[0, 1, 2]`.  🤷‍♀️

jwst/regtest/test_nircam_mtimage.py
```diff
@@ -18,3 +21,30 @@ def test_nircam_image_moving_target(rtdata, fitsdiff_default_kwargs):
     fitsdiff_default_kwargs["atol"] = 1e-5
     diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
     assert diff.identical, diff.report()
+
+@pytest.mark.parametrize("input_file",
+                         [
+                             'jw00634_nrcblong_mttest_tnotinrange_uncal.fits',
+                             'jw00634_nrcblong_no_mtt_uncal.fits',
+                             'jw00634_nrcblong_mttest_uncal.fits',
+                         ])
```
jdavies-st, line 30: To make the parametrization a little clearer, could you add an `ids` kwarg to the parametrize function?
```
ids=["midpt_not_in_mt_table_range", "no_mt_table", "with_mt_table"]
```
That way the test name will have the above, useful info, instead of the filename.

PR #4851, opened 2020-04-24
JP-1415: Update resample to reject NON_SCIENCE pixels

### LINKED ISSUE #4850: Resample needs to reject MIRI pixels flagged as NON_SCIENCE
_Issue [JP-1415](https://jira.stsci.edu/browse/JP-1415) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The MIRI imager flat-field reference files use the NON_SCIENCE DQ flag to indicate pixels that are outside of the relevant portion of the detector for imaging, LRS, and coronagraphic modes, so that the cal software has a way to distinguish the active areas. The NON_SCIENCE DQ flags are propagated into the cal products during calwebb_image2 processing. Right now the resample step is creating combined i2d products for MIRI imaging mode that contain pixels from the entire detector. The resample step needs to be modified to reject any pixels flagged as NON_SCIENCE, in addition to rejecting any pixels flagged as DO_NOT_USE.

hbushouse:
Updated the resample step module to by default reject all pixels with DQ flags of NON_SCIENCE, in addition to DO_NOT_USE. This is especially necessary for MIRI Imaging exposures, where the portion of the detector not covered by direct images (e.g. the LRS and 4QPM areas) is flagged as NON_SCIENCE.

Fixes #4850 / [JP-1415](https://jira.stsci.edu/browse/JP-1415)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -117,6 +117,9 @@ resample
   The parameter ``good_bits`` has been removed in favor of allowing all
   DQ flags except for ``DO_NOT_USE``
 
+- Updated to reject pixels with DQ flag NON_SCIENCE, in addition to
+  DO_NOT_USE. [#4851]
+
 resample_spec
 -------------
 
--- a/jwst/resample/resample_step.py
+++ b/jwst/resample/resample_step.py
@@ -15,8 +15,8 @@ log.setLevel(logging.DEBUG)
 __all__ = ["ResampleStep"]
 
 
-# Force use of all DQ flagged data except for DO_NOT_USE
-GOOD_BITS = '~DO_NOT_USE'
+# Force use of all DQ flagged data except for DO_NOT_USE and NON_SCIENCE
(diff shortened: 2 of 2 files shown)


hbushouse:
New weight SCI and WHT images from one of the MIRI regtests shows that pixels in the LRS and 4QPM areas are now zeroed out (but coronagraphic mask area is not).

dithered_sci

dithered_wht

hbushouse:
I thought (briefly) about the `outlier_detection` step too, but in that case I think it's OK for it to work on NON_SCIENCE pixels, because at least some of them are seeing the sky and hence there can be legit outliers to detect and flag. Remember these are not bad or bogus pixels, they're simply ones that we don't want included in the final combined science image. So regardless of what `outlier_detection` does with the NON_SCIENCE pixels, they'll now get excluded from the combined image.

PR #4858, opened 2020-04-27
JP-1416: Fix outlier_detection background subtraction bug

### LINKED ISSUE #4302: Background over-subtracted
I was running the pipeline on a set of simulated MIRI images (3 dithers). After finishing stage 1 & 2 procedures, I got three images with background ~ 8. Then I run stage 3, and merged the three images together. The merged image has background ~ -8, instead of ~0. 

I investigated each step in stage 3, i.e. tweakreg -> skymatch -> outlier_detection -> resample. After skymatch, the three images still have background ~ 8. But after outlier_detection, the background became ~ 0. After resample (which merges the three images), the background became ~ -8. So, outlier_detection and resample both subtracted the background, leading to an over-subtraction. 

Any idea what's wrong?

Thanks,
Guang

jdavies-st:
Work on a copy instead of the input data.

Add unit tests for outlier detection for imaging modes.
 - change existing test to check that the correct DQ bits are set
 - add test that runs 3 identical (except for noise) images with a single source through `OutlierDetectionStep` and verify that nothing is changed if there are no outliers

Allow `assign_wcs.pointing.create_fitswcs` to be used without a `WAVE-TAB` and use a `CoordinateFrame` object in the input frame.

Fixes #4302 / [JP-1416](https://jira.stsci.edu/browse/JP-1416).

Also reverts #5124 as we now require `gwcs>=0.14`.


jwst/outlier_detection/outlier_detection.py
```diff
@@ -65,7 +65,8 @@ def __init__(self, input_models, reffiles=None, **pars):
             - resample_suffix
 
         """
-        self.inputs = input_models
+        self.inputs = input_models.copy()
```
jdavies-st, line 68: Here's the fix.  🥇

stscieisenhamer (approved):

hbushouse (approved):

jwst/outlier_detection/outlier_detection.py
hbushouse: @jdavies-st is the king of one-liners.

hbushouse:
Needs change log entries.

jwst/assign_wcs/pointing.py
```diff
@@ -223,10 +223,13 @@ def create_fitswcs(inp, input_frame=None):
...
-        sp_axis = spectral_axes[0]
-        if wcsinfo['CTYPE'][sp_axis] == 'WAVE-TAB':
-            wavetable = inp.wavetable
-        transform = fitswcs_transform_from_model(wcsinfo, wavetable)
+        if spectral_axes:
+            sp_axis = spectral_axes[0]
+            if wcsinfo['CTYPE'][sp_axis] == 'WAVE-TAB':
+                wavetable = inp.wavetable
+        else:
+            wavetable = None
```
nden, line 231: wavetable is defined already on line 224.

jwst/assign_wcs/util.py
```diff
@@ -259,13 +259,8 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
...
     input_frame = dmodels[0].meta.wcs.input_frame
-    wnew = wcs_from_fiducial(fiducial, coordinate_frame=out_frame, projection=prj, transform=transform)
-    # temporary fix before gwcs 0.14 is released
-    # wnew = wcs_from_fiducial(fiducial, coordinate_frame=out_frame, projection=prj,
-    #                          transform=transform, input_frame=input_frame)
-    pipe = wnew.pipeline[:]
-    pipe[0] = (input_frame, pipe[0][1])
-    wnew = WCS(pipe)
+    wnew = wcs_from_fiducial(fiducial, coordinate_frame=out_frame, projection=prj,
+                             transform=transform, input_frame=input_frame)
```
nden, line 263: Yes, the fix in gwcs is released now - good you remembered 👍

nden (approved):

jwst/assign_wcs/util.py
jdavies-st: The deprecation warning you put into `gwcs` helped.  I was seeing:

```
/Users/jdavies/miniconda3/envs/jwst/lib/python3.8/site-packages/gwcs/wcs.py:973: DeprecationWarning: Indexing a WCS.pipeline step is deprecated. Use the `frame` and `transform` attributes instead.
    warnings.warn("Indexing a WCS.pipeline step is deprecated. "
```

jwst/assign_wcs/pointing.py
jdavies-st: Ah, I don't need the `else`.  Will fix.

jdavies-st:
Before merging, I'm running regression tests on this branch.  I expect it will affect all `crf`, `s2d` and `i2d` products.

jdavies-st:
And it looks like this fix causes a bunch of errors for the spectra modes as generally these get passed a `SourceModelContainer` that is a `MultiExposureModel` masquerading as a `ModelContainer`, and it does not have a `copy()` method.  So I'l have to implement a more subtle fix further down in the code.

hbushouse:
Well of course! Nothing is ever that easy. I guess we need to rescind your regal status.

jdavies-st:
It looks like it is now a 2-line fix.  The background level gets subtracted once in `resample`, and once in `flag_cr`.  The issue here is that both operate on the inputs in-place, and do not make copies.  *tsk* *tsk*

jwst/assign_wcs/pointing.py
```diff
@@ -223,19 +223,17 @@ def create_fitswcs(inp, input_frame=None):
...
         output_frame = frame_from_model(wcsinfo)
-    #elif isinstance(inp, str):
-        #transform = create_fitswcs_transform(inp)
-        #output_frame = frame_from_fits(inp)
     else:
         raise TypeError("Input is expected to be a DataModel instance or a FITS file.")
 
     if input_frame is None:
-        input_frame = "detector"
+        input_frame = cf.Frame2D(name="detector")
```
jdavies-st, line 236: So this is not quite correct.  As originally implemented, the string "detector" is generic and stand in for a 2D image input frame or 3D cube input frame.  If I want to actually use a `CoordinateFrame` object here instead, especially for `cube_build` which uses this function to generate the `gwcs.WCS` object for the final IFU cubes, it needs to have something like `Frame3D` coordinate frame as input, which doesn't currently exist.  Can I use `CoordinateFrame` directly?  @nden

jdavies-st: Or should I just switch it back to the string input frame, as it was?

jdavies-st: I implemented a solution that checks WCSAXES keyword and uses `Frame2D` or `CoordinateFrame(naxes=3)` appropriately.  And fixed up the tests.

PR #4867, opened 2020-04-29
JP-1413 NIRSpec IFU level-3 ASNs to segregate obs based on the grating

ddavis-stsci:
Modified the asn rules so that grating NIRSpec IFU observations are placed in separate association files. Many truth files will need to be updated.


ddavis-stsci:
This should not affect the MIRI IFU data.

hbushouse:
A unit test involving nirspec product name construction is failing and needs to be fixed.

jdavies-st:
This was merged with failing tests.  Now `master` is failing.  Please fix or revert.

PR #4873, opened 2020-04-30
Populate meta.asn.table_name when association opened in ModelContainer

jdavies-st:
Populate `meta.asn.table_name` in each member of the association that is loaded into a `ModelContainer`.

Related to #4871 #4865 and #4870.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,9 @@ datamodels
   (remove TIME-END; add TDB-BEG, TDB-MID, TDB-END, XPOSURE, TELAPSE)
   [#4925]
 
+- Populate meta.asn.table_name when an association is loaded into a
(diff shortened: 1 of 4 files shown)


jdavies-st:
Agreed.

I should probably double-check this works when there's a catalog `.ecsv` file in the association.  I suspect we need a `try` block in there.  I'll add that to the tests as well.

jdavies-st:
Well, it looks like the kwarg `asn_exptypes="science"` needs to be added to `datamodels.open()` for associations that have non-datamodels such as `.ecsv` catalogs.  So it never even makes it to opening.  I added a `try/except` just in case the attributes are not there for some reason.  I also expanded the one test we have to exercise the `asn_exptypes` keyword, which was not tested.

PR #4882, opened 2020-05-01
JP-1405 Investigate Cube build speed

jemorrison:
This is not to be merged. I pulled out the general blotting and made a routine that blots for MIRI and one the blots for NIRSPEC so I could investigate why NIRSPEC takes so long to do blotting that is used in outlier detection. This PR is just what I have done because I do have question on the results of blotting the median sky cube to each slice on the detector.

For the example I am using I have a NRS1 and NRS2 exposure. The median cube that is created  in outlier detection has 4, 389,760 values.
When I blot this median cube back to the first NIRSPEC slice there are 2,048, 273 mapped back values that fall in the slice. 
For slice 2: 2,003,114
For slice 3: 2,051,264 
around 2 million for each slice
I do not think that is possible. We only have 4 million values in the median cube. The slices on NIRPSPEC do not overlap - so I am not sure what is going on in the invert mapping:
In detail:
for each slice grab the wcs of input model:
slice_wcs = nirspec.nrs_wcs_set_input(model, ii)
x_slice, y_slice = slice_wcs.invert(self.cube_ra, self.cube_dec, self.cube_wave)
then check the bb of the slice and only use x_slice, y_slice in the bb.
those are the number reported above. 
Not sure what is going on here  - maybe @jdavies-st or @nden  can comment

### DIFF
--- a/jwst/cube_build/blot_cube_build.py
+++ b/jwst/cube_build/blot_cube_build.py
@@ -101,10 +101,17 @@ class CubeBlot():
 
         log.info('Number of input models %i ', len(self.input_models))
 
-    # ************************************************************************
-    def blot_images(self):
-        """ Core blotting routine
 
+    def blot_images(self):
+        if self.instrument == 'MIRI':
+            blotmodels = self.blot_images_miri()
+        elif self.instrument == 'NIRSPEC':
+            blotmodels = self.blot_images_nirspec()
+        return blotmodels
+    # ************************************************************************
+    def blot_images_miri(self):
+        """ Core blotting routine for MIRI
+        
         This is the main routine for blotting the median sky image back to
         the detector space and creating a blotting image for each input model
         1. Loop over every data model to be blotted and find ra,dec,wavelength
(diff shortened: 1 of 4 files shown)


nden:
> then check the bb of the slice and only use x_slice, y_slice in the bb.
> those are the number reported above.

@jemorrison How do you check that the values are within the bounding_box? Is it done [here](https://github.com/spacetelescope/jwst/pull/4882/files#diff-a4e2fc43d21355928c30e1156295b32aR319)? If so you need to check if the values are within the bounding_box rather than checking for NaNs. In this case a result of NaN most probably means it started with ra or dec being NaN. I may be missing something.

jemorrison:
@nden 
I do both. First I check for Nans and then I check the bound box as follows (does this look correct). I do double check that the final x_slice and y_slice values are within the xlimit and ylimit 
I print the values to the screen and look at them and they all look correction but there seems 
to be too many of them.

                xlimit, ylimit = slice_wcs.bounding_box
                xuse = np.logical_and(x_slice >= xlimit[0], x_slice <= xlimit[1])
                yuse = np.logical_and(y_slice >= ylimit[0], y_slice <= ylimit[1])
                print('for slice x y range',ii+1,xlimit[0],xlimit[1],ylimit[0],ylimit[1])
                fuse = np.logical_and(xuse, yuse)
               x_slice = x_slice[fuse]
                y_slice = y_slice[fuse]

I am now running an example of only using dithered data from a single detector and grating setting.

jemorrison:
@nden I was tried to do outlier detection when only a single detector and single grating is used for 3 dithered files. In this case the median sky is a better presentation of how to do outlier rejection. I still have the same problem the median sky cube has 2,168,064 elements
Blotting this back to slice 1 on the first input model and  limiting the region by the bound box
results in 1,764,707 values in the first slice BB: xlimits 563.90 to  1001.40 , ylimits: 778.6 to  818.46

I checked the x and y limits of of slice after the BB check and they are all in that slice according to BB. 
Same sort of number for all the other slices.

hbushouse:
@jemorrison Will this ever get used/merged or will it be completely replaced by whatever fixed/new methods are to be implemented for doing NIRSpec IFU blotting? In other words, can we go ahead and close this or do you still need the PR branch for testing?

jemorrison:
This problem is being addresses with JP-1569 and that ticket will include a PR to make blotting work for NIRSPEC

PR #4885, opened 2020-05-02
JP-1426: Updates for new SRCTYAPT keyword

### LINKED ISSUE #4883: Update srctype step to use new SRCTYAPT keyword
_Issue [JP-1426](https://jira.stsci.edu/browse/JP-1426) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

JWSTKD-354 requested the addition of the new keyword SRCTYAPT, which is to contain the value of the source type set by the user in the APT. The existing SRCTYPE keyword is then used by the calwebb_spec2 srctype step to record the source type actually used during calibration. SRCTYAPT preserves (as information only) the original value selected by the user.

In addition, the SRCTYPE keyword needs to get moved from the primary header to the SCI extension header, so that it's a self-contained piece of meta data associated with each slit instance.

The datamodels schemas need to be updated to include SRCTYAPT and to move SRCTYPE to the SCI extension. The srctype step code needs to be updated to use SRCTYAPT as input and SRCTYPE as output.

hbushouse:
Update all necessary datamodel schemas to add the new keyword SRCTYAPT to the primary header and move the SRCTYPE keyword to the SCI and EXTRACT1D headers. Also update the `srctype` step to use SRCTYAPT as input and SRCTYPE as output, with lots of allowances for old datasets that don't have the SRCTYAPT keyword.

A couple of updates were needed in other routines, like `calwebb_tso3` and `extract_1d`, to prevent output files from having empty SCI extensions, just because of the presence of the SRCTYPE keyword defined in the SCI HDU.

This will of course create lots of regtest failures due to the new and relocated keywords, but it's necessary now, because SDP has included SRCTYAPT in their B7.5 code.

Fixes #4883 / [JP-1426](https://jira.stsci.edu/browse/JP-1426)

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -2586,8 +2586,7 @@ def run_extract1d(input_model, refname, smoothing_length, bkg_order,
     # Remove target.source_type from the output model, so that it
     # doesn't force creation of an empty SCI extension in the output
     # x1d product just to hold this keyword.
-    if hasattr(output_model.meta.target, "source_type"):
-        del output_model.meta.target.source_type
+    output_model.meta.target.source_type = None
 
     return output_model
 
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -2583,6 +2583,12 @@ def run_extract1d(input_model, refname, smoothing_length, bkg_order,
                                 log_increment, subtract_background,
                                 apply_nod_offset, was_source_model)
 
+    # Remove target.source_type from the output model, so that it
+    # doesn't force creation of an empty SCI extension in the output
+    # x1d product just to hold this keyword.
+    if hasattr(output_model.meta.target, "source_type"):
+        del output_model.meta.target.source_type
+
     return output_model
 
 
--- a/jwst/pipeline/calwebb_tso3.py
+++ b/jwst/pipeline/calwebb_tso3.py
@@ -137,9 +137,7 @@ class Tso3Pipeline(Pipeline):
             # Remove source_type from the output model, if it exists,
             # to prevent the creation of an empty SCI extension just
             # for that keyword.
-            if (hasattr(x1d_result.meta.target, "source_type") and
-                x1d_result.meta.target.source_type is not None):
-                    del x1d_result.meta.target.source_type
+            x1d_result.meta.target.source_type = None
 
             # For each exposure in the TSO...
             for cube in input_models:
--- a/jwst/pipeline/calwebb_tso3.py
+++ b/jwst/pipeline/calwebb_tso3.py
@@ -117,7 +117,7 @@ class Tso3Pipeline(Pipeline):
         # regardless of how many members there may be...
         phot_result_list = []
         if (input_exptype == 'NRC_TSIMAGE' or
-            (input_exptype == 'MIR_IMAGE'  and input_tsovisit)):
+            (input_exptype == 'MIR_IMAGE' and input_tsovisit)):
             # Create name for extracted photometry (Level 3) product
             phot_tab_suffix = 'phot'
 
@@ -132,7 +132,14 @@ class Tso3Pipeline(Pipeline):
             # define output for x1d (level 3) products
             x1d_result = datamodels.MultiSpecModel()
             # TODO: check to make sure the following line is working
-            x1d_result.update(input_models[0])
+            x1d_result.update(input_models[0], only="PRIMARY")
+
+            # Remove source_type from the output model, if it exists,
+            # to prevent the creation of an empty SCI extension just
+            # for that keyword.
+            if (hasattr(x1d_result.meta.target, "source_type") and
+                x1d_result.meta.target.source_type is not None):
+                    del x1d_result.meta.target.source_type
 
             # For each exposure in the TSO...
             for cube in input_models:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,6 +44,9 @@ datamodels
 
 - Added FASTGRPAVG[8,16,32,64] to the READPATT keyword allowed values. [#4818]
 
+- Added the SRCTYAPT keyword and moved SRCTYPE to the SCI extension header of
+  all applicable data model schemas. [#4885]
+
 exp_to_source
 -------------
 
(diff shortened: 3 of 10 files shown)


jwst/extract_1d/extract.py
```diff
@@ -2583,6 +2583,12 @@ def run_extract1d(input_model, refname, smoothing_length, bkg_order,
                                 log_increment, subtract_background,
                                 apply_nod_offset, was_source_model)
 
+    # Remove target.source_type from the output model, so that it
+    # doesn't force creation of an empty SCI extension in the output
+    # x1d product just to hold this keyword.
+    if hasattr(output_model.meta.target, "source_type"):
```
jdavies-st, line 2589: Because `source_type` is in the schema, `hasattr` will always evaluate to true.

Deleting the attribute is maybe not great?  Can you set it to `None` instead?  It shouldn't get written out then.

```
In [13]: model = datamodels.ImageModel()

In [14]: model.meta.wcsinfo.roll_ref = 45.

In [15]: model.save("foo.fits")
Out[15]: 'foo.fits'

In [16]: !fitsinfo foo.fits
Filename: foo.fits
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU       7   ()      
  1  SCI           1 ImageHDU        10   ()      
  2  ASDF          1 BinTableHDU     11   1R x 1C   [523B]   

In [17]: !fitsheader foo.fits
# HDU 0 in foo.fits:
SIMPLE  =                    T / conforms to FITS standard                      
BITPIX  =                    8 / array data type                                
NAXIS   =                    0 / number of array dimensions                     
EXTEND  =                    T                                                  
DATE    = '2020-05-02T21:16:22.311' / [yyyy-mm-ddThh:mm:ss.ss] UTC date file cre
FILENAME= 'foo.fits'           / Name of the file                               
DATAMODL= 'ImageModel'         / Type of data model                             

# HDU 1 in foo.fits:
XTENSION= 'IMAGE   '           / Image extension                                
BITPIX  =                    8 / array data type                                
NAXIS   =                    0 / number of array dimensions                     
PCOUNT  =                    0 / number of parameters                           
GCOUNT  =                    1 / number of groups                               
EXTNAME = 'SCI     '           / extension name                                 
                                                                                
        WCS parameters                                                          
                                                                                
ROLL_REF=                 45.0 / [deg] V3 roll angle at the ref point (N over E)

# HDU 2 in foo.fits:
XTENSION= 'BINTABLE'           / binary table extension                         
BITPIX  =                    8 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                  523 / length of dimension 1                          
NAXIS2  =                    1 / length of dimension 2                          
PCOUNT  =                    0 / number of group parameters                     
GCOUNT  =                    1 / number of groups                               
TFIELDS =                    1 / number of table fields                         
TTYPE1  = 'ASDF_METADATA'                                                       
TFORM1  = '523B    '                                                            
EXTNAME = 'ASDF    '           / extension name                                 

In [18]: model.meta.wcsinfo.roll_ref = None

In [19]: model.save("foo.fits")
Out[19]: 'foo.fits'

In [20]: !fitsinfo foo.fits
Filename: foo.fits
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU       7   ()      
  1  ASDF          1 BinTableHDU     11   1R x 1C   [509B]   

In [21]: !fitsheader foo.fits
# HDU 0 in foo.fits:
SIMPLE  =                    T / conforms to FITS standard                      
BITPIX  =                    8 / array data type                                
NAXIS   =                    0 / number of array dimensions                     
EXTEND  =                    T                                                  
DATE    = '2020-05-02T21:17:21.005' / [yyyy-mm-ddThh:mm:ss.ss] UTC date file cre
FILENAME= 'foo.fits'           / Name of the file                               
DATAMODL= 'ImageModel'         / Type of data model                             

# HDU 1 in foo.fits:
XTENSION= 'BINTABLE'           / binary table extension                         
BITPIX  =                    8 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                  509 / length of dimension 1                          
NAXIS2  =                    1 / length of dimension 2                          
PCOUNT  =                    0 / number of group parameters                     
GCOUNT  =                    1 / number of groups                               
TFIELDS =                    1 / number of table fields                         
TTYPE1  = 'ASDF_METADATA'                                                       
TFORM1  = '509B    '                                                            
EXTNAME = 'ASDF    '           / extension name                                 
```

jwst/pipeline/calwebb_tso3.py
```diff
@@ -132,7 +132,14 @@ def process(self, input):
...
             # TODO: check to make sure the following line is working
-            x1d_result.update(input_models[0])
+            x1d_result.update(input_models[0], only="PRIMARY")
+
+            # Remove source_type from the output model, if it exists,
+            # to prevent the creation of an empty SCI extension just
+            # for that keyword.
+            if (hasattr(x1d_result.meta.target, "source_type") and
+                x1d_result.meta.target.source_type is not None):
+                    del x1d_result.meta.target.source_type
```
jdavies-st, line 142: Same here.  I think

```
x1d_result.meta.target.source_type = None
```

would suffice.

jwst/extract_1d/extract.py
hbushouse: I swear I had initially tried just setting it to None, but then got "attribute not found" errors, so I went to using "hasattr". But now that you pointed this out, I switched it all back to not using "hasattr" and just setting to None and it all seems to be working for quite a number of spectral-based regtests I've tried it on. So ... go figure. Latest commit will have the updates.

jdavies-st: Great!

I thought about this a bit more, and I think perhaps the more elegant solution is to not have anything in `core.schema` that gets mapped to non-PRIMARY HDUs.  Instead have a separate `keyword_foo` schema or `sci` schema that maps those, and then you can be selective of what gets written out.  The reason the x1d files are getting the keyword (possibly) in the first place is because it is in `core.schema`.  If it is not, but included in every datamodel that needs it, we can avoid having `x1d` files get it altogether.

This doesn't solve the issue that Jane was seeing in BUNIT in the ERR array of `ImageModel`, as that has to be there.  But in that case updating just the primary array via `update()` is the solution.

PR #4889, opened 2020-05-04
Enable calspec2 processing of nirspec lamp exposures

stscirij:
Addresses JP-1429


stscirij:
The test errors about 'slit frame and msa frame are not in sequence' come from the assign_wcs task when no ifu_fore, ifu_post and ifu_slicer reference files are selected

jwst/assign_wcs/nirspec.py
```diff
@@ -171,7 +172,14 @@ def ifu(input_model, reference_files, slit_y_range=[-.55, .55]):
...
     filter = input_model.meta.instrument.filter
-
+    # Check for ifu reference files
+    if reference_files['ifufore'] is None and \
+       reference_files['ifuslicer'] is None and \
+       reference_files['ifupost'] is None:
+        # No ifu reference files, won't be able to create pipeline
+        log_message = 'No ifufore, ifuslicer or ifupost reference files'
+        log.critical(log_message)
+        raise RuntimeError(log_message)
```
hbushouse, line 182: Does the RuntimeError get trapped up in the pipeline module, so that it can shut down cleanly?

jwst/assign_wcs/nirspec.py
```diff
@@ -358,16 +375,23 @@ def get_open_slits(input_model, reference_files=None, slit_y_range=[-.55, .55]):
     """Return the opened slits/shutters in a MOS or Fixed Slits exposure.
     """
     exp_type = input_model.meta.exposure.type.lower()
-    if exp_type in ["nrs_msaspec", "nrs_autoflat"]:
+    lamp_mode = input_model.meta.instrument.lamp_mode
+    if type(lamp_mode) == str:
+        lamp_mode = lamp_mode.lower()
+    else:
```
hbushouse, line 381: Is this just to guard against input files that don't have the OP_MODE keyword?

jwst/pipeline/calwebb_spec2.py
```diff
@@ -220,15 +221,17 @@ def process_exposure_product(
         # Technically there should be just one.
         # We'll just get the first one found
         imprint = members_by_type['imprint']
-        if exp_type in ['NRS_MSASPEC', 'NRS_IFU'] and \
-           len(imprint) > 0:
+        if len(imprint) > 0 and (exp_type in ['NRS_MSASPEC', 'NRS_IFU'] or \
+           is_nrs_ifu_flatlamp(input)):
```
hbushouse, line 225: Do we expect anyone to ever take associated imprint (light leak) exposures along with lamp exposures?

jwst/assign_wcs/nirspec.py
stscirij: calwebb_spec2 traps all exceptions in assign_wcs, so it should look like any other assign_wcs problem

jwst/assign_wcs/nirspec.py
stscirij: Yes, I put this code in before I changed datamodels to set a default value for lamp_mode, it was barfing against trying to change a NoneType to lower().  It shouldn't be needed now.

jwst/pipeline/calwebb_spec2.py
stscirij: That's up to the association rules.  According to the INS requirements table, imprint_subtract is supposed to be run for NRS_LAMP exposures with lamp_mode = 'IFU', although they have a footnote in that says "leakage exposures not specifiable as such in lamp template, so no automatic association?"

stscirij:
The decision about which steps to run was delegated to the pars files for each exposure type, so until those files have been put into CRDS steps like srctype, photom, resample/cube_build and extract_1d will run.  I haven't tested against non_lamp exposures yet, I was going to run some of the regtests before merging.

hbushouse:
> The decision about which steps to run was delegated to the pars files for each exposure type, so until those files have been put into CRDS steps like srctype, photom, resample/cube_build and extract_1d will run. I haven't tested against non_lamp exposures yet, I was going to run some of the regtests before merging.

But I thought it wasn't possible for some of these lamp type exposures to make it through steps like cube_build and resample_spec? So are they all going to crash?

You've already added some logic here and there within `calwebb_spec2` itself to decide whether to run a particular step, like msa_flagging. So does that mean some of the logic will be in the pipeline and some will be in the .cfg (par) files?

stscirij:
> > The decision about which steps to run was delegated to the pars files for each exposure type, so until those files have been put into CRDS steps like srctype, photom, resample/cube_build and extract_1d will run. I haven't tested against non_lamp exposures yet, I was going to run some of the regtests before merging.
> 
> But I thought it wasn't possible for some of these lamp type exposures to make it through steps like cube_build and resample_spec? So are they all going to crash?
> 
> You've already added some logic here and there within `calwebb_spec2` itself to decide whether to run a particular step, like msa_flagging. So does that mean some of the logic will be in the pipeline and some will be in the .cfg (par) files?

It's a bit of a balancing act.  If we handled msa flagging with pars files, almost every type of data would need a pars file to make sure that step was skipped, so it makes sense to put the conditions for running the step in the code.  For the steps like srctype, photom and extract_1d which are to be run by the majority of exposure types, it makes more sense (to me) to handle the exceptions using pars files.  If it's going to cause the pipeline to crash by running these steps with the default settings, I'll put code in calwebb_spec2 to make sure that doesn't happen.

jwst/assign_wcs/nirspec.py
nden: I've been thinking forever (and never did anything about it) that something like this is needed in the assign_wcs_step.py module when the reference files are pulled from CRDS. assign_wcs requires and will not run without reference files. Some of them apply to all modes but it's nice that you checked for the ifu specific ones here.

jwst/assign_wcs/nirspec.py
```diff
@@ -211,7 +219,19 @@ def ifu(input_model, reference_files, slit_y_range=[-.55, .55]):
 
     det, sca, gwa, slit_frame, msa_frame, oteip, v2v3, world = create_frames()
 
-    if input_model.meta.instrument.filter != 'OPAQUE':
+    exp_type = input_model.meta.exposure.type.upper()
+
+    is_lamp_exposure = exp_type in ['NRS_LAMP', 'NRS_AUTOWAVE', 'NRS_AUTOFLAT']
```
stscieisenhamer, line 224: There is a number of copies of this statement. Would this be appropriate to place in the `exptypes` library as function?

stscirij: Yes, I did make a bunch of boolean functions in exptypes, but this code was written before I did that and I didn't update it

nden:
The test failures seem real.

nden:
This needs changes in CRDS before it can be merged.

Issue #3979, opened 2019-08-31
Structured array properties of DataModels should not be converted to FITS_rec 


mcara:
PR https://github.com/spacetelescope/jwst/pull/1754 introduced automatic conversion of structured array properties to `FITS_rec`. In my opinion, `DataModel` attributes should have the type described in corresponding schemas which is unexpected. Therefore, I think https://github.com/spacetelescope/jwst/pull/1754 should be rolled back and handling of units in FITS tables should be handled separately.

CC: @hbushouse @perrygreenfield @nden

hbushouse:
I agree that having things converted to FITS_rec within a datamodel is a contrived solution to problem of attaching units (and any other column attributes). If someone can find a way to handle these types of column attributes that doesn't require conversion to FITS_rec (at least not within the datamodel itself - conversion should happen only when serializing to a FITS file), I'm all for reverting #1754. But we do need the column units - so that problem needs to be figured out first.

mcara:
I agree with you that units must work. I believe it is possible but it will not be trivial and quite time consuming to do it carefully.

jdavies-st:
Labelling this as a bug, as the result of this behavior means that any units (i.e. TUNIT) described in a FITS binary table get dropped on the floor when read in as a datamodel.

melanieclarke:
@emolter @braingram - do you know if this is still an issue?

braingram:
I would say this can be closed.

https://github.com/spacetelescope/stdatamodels/issues/240 detailed a way to use TUNIT keywors with stdatamodels. To copy the [last comment of that closed issue](https://github.com/spacetelescope/stdatamodels/issues/240#issuecomment-2651989979)
> This could be implemented via one of the strategies described here. However, there is currently no need for the feature and this work can be deferred indefinitely until such a need arises

PR #4902, opened 2020-05-06
JP-1107: Implement aperture correction in extract_1d_step

### LINKED ISSUE #4206: Apply a spectroscopic aperture correction in extract_1d
Issue [JP-1107](https://jira.stsci.edu/browse/JP-1107) was created by Philip Hodge:

The extract_1d step needs to apply a correction to the flux to account for the finite size of the extraction aperture, for some modes. The size and location of the extraction region may need to be taken from the aperture correction reference file, rather than from the extract_1d reference file.

jwhite3:
This PR adds functionality to generate and apply an aperture correction in extract1d. 

Resolves #4206 / [JP-1107](https://jira.stsci.edu/browse/JP-1107)


jwhite3:
Note: This shouldn't be merged until the reference files are available.

jwst/assign_wcs/util.py
```diff
@@ -100,25 +100,40 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray]) -> float:
         Reference WCS object from which to compute a scaling factor.
 
     fiducial : tuple
-        Input fiducial of (RA, DEC) used in calculating reference points.
+        Input fiducial of (RA, DEC) or (RA, DEC, Wavelength) used in calculating reference points.
+
+    spec : bool
```
jdavies-st, line 105: This name is a bit vague - `spec` can mean specification or spectral.  I would make it `spectral` or even `spectral_wcs`.

Or maybe the function doesn't need this kwarg and can figure out internally if it is a spectral WCS?  I.e. the output frame will have both spatial and spectral components.

jwst/assign_wcs/util.py
```diff
@@ -100,25 +100,40 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray]) -> float:
...
+        Switch to use if input is a spectral WCS
+
+    disp_axis : int
+        Dispersion axis integer. Assumes the same convention as `wcsfino.dispersion_direction`
```
jdavies-st, line 109: `wcsinfo` typo

jwst/extract_1d/tests/test_apply_apcorr.py
```diff
@@ -0,0 +1,169 @@
+import pytest
+import numpy as np
+import os
+
+from astropy.io import fits
+from astropy.table import Table
+
+from ...datamodels import DataModel
```
jdavies-st, line 8: Imports in test modules should be absolute instead of relative.

jwst/extract_1d/extract.py
```diff
@@ -2518,9 +2520,8 @@ def separate_target_and_background(self, ref):
         return (mask_target, mask_bkg)
 
 
-def run_extract1d(input_model, refname, smoothing_length, bkg_order,
-                  log_increment, subtract_background, apply_nod_offset,
-                  was_source_model=False):
+def run_extract1d(input_model, extract_ref_name, smoothing_length, bkg_order, log_increment, subtract_background,
```
hbushouse, line 2523: I know that when adding a new argument to a function it's obviously simplest to just append it to the end of the list, but wouldn't it make more sense (stylistically or whatever) to put extract_ref_name and apcorr_ref_name back-to-back in the list somewhere? No impact on functionality of course, just API style.

jwst/extract_1d/extract.py
```diff
@@ -2530,7 +2531,7 @@ def run_extract1d(input_model, refname, smoothing_length, bkg_order,
     input_model : data model
         The input science model.
 
-    refname : str
+    extract_ref_name : str
         The name of the reference file, or "N/A".
```
hbushouse, line 2535: Perhaps expand this to "The name of the EXTRACT1D reference file", because there's no longer a single ("the") reference file for the step.

jwst/extract_1d/extract.py
```diff
@@ -2562,14 +2563,25 @@ def run_extract1d(input_model, refname, smoothing_length, bkg_order,
...
-    """
 
+    """
     # Read and interpret the reference file.
-    ref_dict = load_ref_file(refname)
+    ref_dict = load_ref_file(extract_ref_name)
+
+    apcorr_table = apcorr_sizeunits = None
+    if apcorr_ref_name is not None or apcorr_ref_name != 'N/A':
+        with fits.open(apcorr_ref_name) as ap_ref:
```
hbushouse, line 2580: Don't we have data models defined for apcorr ref files and hence should/could be using datamodels.open() here, rather than direct FITS I/O? We try to avoid any direct I/O as much as possible and let datamodels handle it.

jwst/extract_1d/extract.py
```diff
@@ -2562,14 +2563,25 @@ def run_extract1d(input_model, refname, smoothing_length, bkg_order,
...
+            apcorr_table = ap_ref['APCORR'].data.copy()
+            apcorr_sizeunits = ap_ref['APCORR'].header['SIZEUNIT']
```
hbushouse, line 2582: If we loaded apcorr into a data model, you could then just pass the data model around, rather than having to store and pass around separate objects containing the table data, units, etc.

jwst/extract_1d/extract.py
```diff
@@ -2578,10 +2590,19 @@ def run_extract1d(input_model, refname, smoothing_length, bkg_order,
...
+    output_model = do_extract1d(
+        input_model,
+        ref_dict,
+        smoothing_length,
+        bkg_order,
+        log_increment,
+        subtract_background,
+        apply_nod_offset,
+        was_source_model,
+        apcorr_table=apcorr_table,
```
hbushouse, line 2603: See above comment about passing apcorr as a datamodel, rather than multiple data objects.

jwst/extract_1d/extract.py
```diff
@@ -2753,61 +2777,77 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
...
 
             try:
-                (ra, dec, wavelength, temp_flux, background,
-                 npixels, dq, prev_offset) = extract_one_slit(
-                                        input_model, slit, -1,
-                                        prev_offset, True, extract_params)
+                ra, dec, wavelength, temp_flux, background, npixels, dq, prev_offset = extract_one_slit(
+                    input_model,
+                    slit,
+                    -1,
```
hbushouse, line 2846: Is there some way to make it more obvious exactly what this value is referring to in the argument list, e.g. "sp_order = -1" or whatever it is? Otherwise the reader/maintainer has to go find the extract_one_slit() function and count down through the list of arguments to figure out what it is.

jwst/extract_1d/extract.py
```diff
@@ -2859,14 +2903,29 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
             spec.spectral_order = sp_order
             spec.dispersion_direction = extract_params['dispaxis']
             copy_keyword_info(slit, slit.name, spec)
+
+            if source_type is not None and source_type.upper() == 'POINT' and apcorr_table is not None:
+                log.info('Applying Aperture correction.')
+                if instrument == 'NIRSPEC':
```
hbushouse, line 2909: Maybe some comments would be good here to explain why these wl values are being set the way they are.

jwst/extract_1d/extract_1d_step.py
```diff
@@ -73,7 +73,7 @@ class Extract1dStep(Step):
     apply_nod_offset = boolean(default=None)
     """
```
hbushouse, line 74: I'm thinking it couldn't hurt to add an optional argument here to allow for turning the apcorr on and off, e.g. "apply_apcorr = boolean(default=True)", at least to support testing (i.e. turn it off and verify that results from before have not changed). There may also be times where a user doesn't want to apply it. Of course you also then need to add logic into the code to make use of that param.

jwst/extract_1d/tests/test_apply_apcorr.py
```diff
@@ -0,0 +1,169 @@
...
+}
+
+
+@pytest.fixture(
+    scope='module',
+    params=[
+        ('MIRI', 'MIR_LRS-FIXEDSLIT'),
+        ('MIRI', 'MIR_LRS-SLITLESS'),
+        ('MIRI', 'MIR_MRS'),
+        ('NIRCAM', 'NRC_GRISM'),
```
hbushouse, line 25: If this is for the NIRCam time series grism mode, then it should be "NRC_**TS**GRISM"

jwst/extract_1d/extract.py
jdavies-st: I assume this is because datamodels table reader drops units defined in TUNIT on the floor?  This really should be fixed in the datamodels infrastructure.

The bug is somewhere in `datamodels.util.gentle_asarray()`:

https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/util.py#L326-L377

The `unit` info is the the column, but we just use the `dtype`.  There's no unit info in the `dtype`.  The above code should probably be refactored to use `column` objects instead of `dtype`.

jdavies-st: There's even an open issue for this

https://github.com/spacetelescope/jwst/issues/3979

Tackle in an upcoming sprint?  🤞

hbushouse: We added a dedicated keyword to the headers of the apcorr ref files to store the units, separate from TUNITn, in order to work around this problem. So as long as there's a meta data entry for that special keyword in the datamodel schema for the ref file, we should be good (i.e. the data and units will both be available in the model).

hbushouse: > There's even an open issue for this
> 
> #3979
> 
> Tackle in an upcoming sprint? 🤞

Are you volunteering?

jwst/assign_wcs/util.py
jwhite3: Ah, so something like: 

```python
if 'SPECTRAL' in wcs.output_frame.axes_type:
    ...  # do the things for spectral WCSs
```

jwst/extract_1d/apply_apcorr.py
```diff
@@ -0,0 +1,271 @@
...
+        },
+        'NIRSPEC': {
+            'IFU': {'instrument': ['filter', 'grating']},
+            'MSASPEC': {'instrument': ['filter', 'grating']},
+            'FIXEDSLIT': {'instrument': ['filter', 'grating']},  # Slit is also required; passed in as init arg
+            'BRIGHTOBJ': {'instrument': ['filter', 'grating']}
+        },
+        'NIRCAM': {
+            'WFSS': {'instrument': ['filter', 'pupil']},
+            'NRC_GRISM': {'instrument': ['filter', 'pupil']}
```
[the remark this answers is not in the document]
jwhite3: Ok, then it might be as simple as removing that. 

If I remember correctly, I tried to create the different branches based on possible `EXPTYPE`s in JWST-CRDS, but if it's not going to be used, then removing it should work just fine

jwst/extract_1d/tests/test_apply_apcorr.py
jwhite3: Same as the other comment above. I think the correct answer is to remove it

jwst/extract_1d/extract.py
jwhite3: Yep, so originally I was using `jwst.datamodels.open`, but we were having issues with consistently being able to extract the units for one of the columns. Additionally, some of the reference files don't contain a `DATAMODL` keyword, and I wasn't sure how to get around that without writing a function with a hardcoding mapping of some kind. Thoughts?

hbushouse: You can work around the missing `DATAMODL` keyword by explicitly opening with the required/desired data model, e.g. `jwst.datamodels.NrcWfssApcorrModel`, but that of course has the added burden of having separate open calls for every mode of data. We should just enforce the presence of `DATAMODL` in the reference files.

jwst/extract_1d/apply_apcorr.py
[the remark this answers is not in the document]
jwhite3: Your correct! The rmap definitely has issues of this exact nature (I forget which mode I was testing... might've even been this one). A CRDS issue was filed for it. 

@stscieisenhamer do you think that the CRDS issue for the APCORR rmap covers this? Or is it too specific

jwst/extract_1d/extract.py
jwhite3: @hbushouse that's what I was thinking as well and I think it would allow for trimming number of functions that do exactly that for other steps.

jwst/extract_1d/apply_apcorr.py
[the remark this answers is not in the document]
jwhite3: Found the issue: [CRDS-355](https://jira.stsci.edu/browse/CRDS-355)

jwst/assign_wcs/util.py
```diff
@@ -100,25 +100,39 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray]) -> float:
...
-        Scaling factor for x and y.
+        Scaling factor for x and y or dispersion direction.
 
     """
-    if len(fiducial) != 2:
-        raise ValueError(f'Input fiducial must contain only (RA, DEC); Instead recieved: {fiducial}')
+    spectral = 'SPECTRAL' in wcs.output_frame.axes_type
+
+    if spectral and disp_axis is None:
+        raise ValueError(f'If input WCS is spectral, a disp_axis must be given')
```
jwhite3, line 117:
```suggestion
        raise ValueError('If input WCS is spectral, a disp_axis must be given')
```

jwhite3:
Note: (just in case) these changes are also affected by [AL-335](https://jira.stsci.edu/browse/AL-335) and should probably not be merged until that issue is resolved as it appears that reading in the APCORR files will result in errors due to this bug. After that issue is merged, additional testing should be done to ensure that the problem is resolved

hbushouse:
Problem with datamodels traced to one badly formed ref file, which has been fixed and will be redelivered to CRDS soon. NIRCam apcorr rmap was also recently updated to set NRC_TSGRISM to N/A, so that mode will not get apcorr applied. Once the fixed MIRI ref file is available in CRDS, this should finally be ready to merge. There are no outstanding issues related to datamodels.

jwhite3:
Note on the execution path of extraction for `extract_1d`: 

- Starts with `extract_1d_step`
  - `process` calls `extract.run_extract1d`
    - `run_extract1d` calls `extract.do_extract1d` (`do_extract1d` is the main driver of most extractions)
      - `do_extract1d` generally calls `extract_one_slit`
        - `extract_one_slit` uses instances of either `ExtractModel` or `ImageExtractModel` which further configure extraction
          - `ExtractModel`/`ImageExractModel` both call the method `extract`
            - The `extract` method calls the `extract1d` function from `extract1d.py` (this is where the extraction is happening)
      - However for **IFU** data, `do_extract1d` actually calls `ifu_extract1d` from `extract1d.ifu.py`
        - `ifu_extract1d` either calls `extract_ifu` or `image_extract_ifu` each of which is it's own, complete extraction algorithm independent from everything else

@jdavies-st Let me know if this format isn't clear and I can attempt to write it out a bit differently

hbushouse:
Regression tests showing expected differences, but no errors. Merging.

PR #4905, opened 2020-05-06
Support asdf 2.6.x

### LINKED ISSUE #4864: Call asdf.schema.get_validator with _visit_repeat_nodes flag
Version 2.6.0 of the asdf library improved validator efficiency by visiting nodes only once each, even if they were represented in the tree multiple times.  Unfortunately, this breaks datamodels fits output, which rides along with asdf validation and expects that nodes will be visited every time they show up in the tree.  Included in asdf 2.6.0 is a new flag, `_visit_repeat_nodes`, which will cause the validator to revert to its original behavior.

Currently the asdf requirement is restricted to `~=2.5.0`.  When we want to upgrade asdf, we'll need to change that to `>=2.6` and add `_visit_repeat_nodes=True` to the call to `get_validator` in `jwst.datamodels.fits_support._save_from_schema`. 

asdf 2.6.0 is compatible with both astropy 4.0.x and 4.1.x, but 4.1.x requires it, so making this change is a prerequisite to upgrading astropy.

eslavich:
The asdf 2.6.0 release includes a validator performance improvement that skips nodes that have already been visited during the current validation pass.  This package relies on the old behavior to ensure that fits headers have been fully populated, so here I'm setting a flag that will cause asdf to continue to revisit nodes.  This flag is not available in earlier versions of asdf so we need to inspect the asdf version before attempting to use it.

Resolves #4864

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,11 @@ associations
 ------------
 - Modify NIRSpec IFU level-3 ASN rules to include only one grating per association [#4845]
 
+datamodels
+----------
+
+- Relax asdf requirement and use validator flag when asdf 2.6.x is installed [#4905]
+
 0.16.0 (2020-05-04)
 ===================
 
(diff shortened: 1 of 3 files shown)


jdavies-st:
The argument for including this in any 0.16.0 patches is that even though SDP is getting `asdf 2.5.2` for their Build 7.5, future users of `jwst 0.16` outside of SDP may want to use `astropy 4.1` with it, which is yet to be released.  And it will require `asdf 2.6`.

nden:
This needs a rebase.
@hbushouse Can this PR be merged or are we waiting for something to happen in DMS?

PR #4909, opened 2020-05-07
Revert RT setup back to using CRDS ops

hbushouse:
Revert JenkinsRT setup back to using crds ops server and latest context, now that all updates have been migrated from test.


jdavies-st:
Should this be milestoned for 7.5?

hbushouse:
I sometimes don't bother with milestones for things that are strictly related to testing or infrastructure, because they don't directly affect our delivered software.

jdavies-st:
True, but if we ever want to run this version of our code through the regression tests again, it might be good to have up-to-date data within the repo of what the expected CRDS setup was.

jdavies-st:
I say this as the 7.5 milestone stuff is what is going to go into our 0.16.1 patch.

Issue #4845, opened 2020-04-22
Modify NIRSpec IFU level-3 ASN rules to include only one grating


stscijgbot-jp:
_Issue [JP-1413](https://jira.stsci.edu/browse/JP-1413) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

If a NIRSpec IFU (EXP_TYPE='NRS_IFU') observation contains a mixture of exposures using multiple gratings (e.g. g235h, g395h) the current ASN rules create spec3 ASN's that contain the exposures for all gratings. In theory this is acceptable, because there is some wavelength overlap between adjacent gratings and hence merged spectral cubes can be created. However, at least right now there are tremendous performance issues associated with having such large ASN's, especially if multi-point and multi-tile dither/mosaic patterns are in use. There are occasions where such large ASN's have also run into memory errors in Ops processing (e.g. JWSTDMS-44).

It would be prudent, therefore, to at least temporarily modify the ASN rules to create separate spec3 ASN's for each grating value encountered in a NIRSpec IFU observation, to help alleviate the processing performance problems.

stscijgbot-jp:
_Comment by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse) on [JIRA](https://jira.stsci.edu/browse/JP-1413?focusedCommentId=460309#comment-460309):_

The test data directory includes the original pool file "jw00626_20200307t140611_pool.csv" and the original spec3 ASN file for observation 29 "jw00626-o029_20191129t190337_spec3_001_asn.json". That spec3 ASN contains a total of 16 exposures, 8 that use grating g235h and 8 that use grating g395h.

With the modified ASN rules, 2 spec3 files should be created, with one containing the 8 g235h exposures, and the other containing the 8 g395h exposures. Examples of such spec3 ASN's (created manually) are "jw00626-o029_20191129t190337_spec3_002_asn.json" and "jw00626-o029_20191129t190337_spec3_003_asn.json".

stscijgbot-jp:
_Comment by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis) on [JIRA](https://jira.stsci.edu/browse/JP-1413?focusedCommentId=461655#comment-461655):_

Updated the asn rules to segregate observations based on the grating in use. An example for jw00626-o029 is attached above.

stscijgbot-jp:
[...]

PR #4926, opened 2020-05-11
Jp 1413 Updates to split grating observations into separate asns

ddavis-stsci:
Updated the asn rules to segregate observations based on the grating in use. This touched many more modules than I expected. 

closes JP-1413, #4845


ddavis-stsci:
The travis failure is from pkg_resources.VersionConflict, which is ignorable.

PR #4930, opened 2020-05-11
Fix nonsensical premature closing of FITS file of a DataModel

stscieisenhamer:
Or, well methinks its nonsensical anyways, but the PR review will indicate otherwise.

In general, when a `DataModel` is opened from a FITS file, the filehandle for the FITS file is held internally until the `DataModel` is `close`ed garbage-collected (maybe). However, ran across this bit of code in which the FITS file is opened in a context manager and the file handled is stored for later closing. This would be the nonsensical point: either a context manager should not be used OR the filehandle should not be stored for later re-closing.

Though having the FITS file open is an integral part of `DataModel` operation, this appears to have not had any operational effect. But, current work on `DataModel` reading efficiency has actually caused the expected "I/O operation on closed file" error.

Because of the confusing nature of the current code, and with an eye to possible future issues, this PR, which removes the context manager bit and leaving closing to the intended (?) closing of the DataModel itself, is being filed.

The set of file changes are as follows
    - The fix is in `model_base.py`
    - Many tests and a couple cube-related modules have been modified to use contexts as needed.
    - Finally, and my apologies, a number of tests are changed only for PEP8 issues.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -173,12 +173,11 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
                 else:
                     init_fitsopen = init
 
-                with fits.open(init_fitsopen, memmap=memmap) as hdulist:
-                    asdffile = fits_support.from_fits(hdulist,
-                                                  self._schema,
-                                                  self._ctx,
-                                                  **kwargs)
-                    self._files_to_close.append(hdulist)
+                hdulist = fits.open(init_fitsopen, memmap=memmap)
+                asdffile = fits_support.from_fits(
+                    hdulist, self._schema, self._ctx, **kwargs
+                )
+                self._files_to_close.append(hdulist)
 
             elif file_type == "asdf":
                 asdffile = self.open_asdf(init=init, **kwargs)
@@ -274,6 +273,10 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
 
         return "".join(buf)
 
+    def __del__(self):
+        """Ensure closure of resources when deleted."""
+        self.close()
+
     @property
     def override_handle(self):
         """override_handle identifies in-memory models where a filepath
@@ -304,7 +307,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
         _drop_array(self._instance)
 
     def close(self):
-        if not self._iscopy and self._asdf is not None:
+        if not self._iscopy and getattr(self, '_asdf', None) is not None:
             self._asdf.close()
             self._drop_arrays()
 
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -307,7 +307,7 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
         _drop_array(self._instance)
 
     def close(self):
-        if not self._iscopy and getattr(self, '_asdf', None) is not None:
+        if not self._iscopy and self._asdf is not None:
             self._asdf.close()
             self._drop_arrays()
 
--- a/jwst/datamodels/tests/test_models.py
+++ b/jwst/datamodels/tests/test_models.py
@@ -617,12 +617,18 @@ def test_hasattr():
 
 
 def test_validate_on_read():
+    """ Test for proper validation error
+
+    Note: The FITS file is opened separately in order to properly close
+    the file.
+    """
     schema = ImageModel((10, 10))._schema.copy()
     schema['properties']['meta']['properties']['calibration_software_version']['fits_required'] = True
+    hduls = fits.open(FITS_FILE)
 
     with pytest.raises(jsonschema.ValidationError):
-        with ImageModel(FITS_FILE, schema=schema, strict_validation=True):
-            pass
+        ImageModel(hduls, schema=schema, strict_validation=True)
+    hduls.close()
 
 
 def test_validate_required_field():
--- a/jwst/datamodels/tests/test_models.py
+++ b/jwst/datamodels/tests/test_models.py
@@ -624,11 +624,10 @@ def test_validate_on_read():
     """
     schema = ImageModel((10, 10))._schema.copy()
     schema['properties']['meta']['properties']['calibration_software_version']['fits_required'] = True
-    hduls = fits.open(FITS_FILE)
 
-    with pytest.raises(jsonschema.ValidationError):
-        ImageModel(hduls, schema=schema, strict_validation=True)
-    hduls.close()
+    with fits.open(FITS_FILE) as hduls:
+        with pytest.raises(jsonschema.ValidationError):
+            ImageModel(hduls, schema=schema, strict_validation=True)
 
 
 def test_validate_required_field():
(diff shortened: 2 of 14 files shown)


jwst/datamodels/tests/test_models.py
```diff
@@ -615,12 +615,18 @@ def test_info():
...
 
 def test_validate_on_read():
+    """ Test for proper validation error
+
+    Note: The FITS file is opened separately in order to properly close
+    the file.
+    """
     schema = ImageModel((10, 10))._schema.copy()
     schema['properties']['meta']['properties']['calibration_software_version']['fits_required'] = True
+    hduls = fits.open(FITS_FILE)
```
jwhite3, line 625: In this case I think using the context manager should work right?

```python
with fits.open(FITS_FILE) as hduls:
    with pytest.raises(jsonschema.ValidationError):
        ImageModel(hduls, schema=schema, strict_validation=True)
```

stscieisenhamer: If I remember correctly, I did try that and it didn't work, in the sense that the pytest openfiles plugin still failed. Will try again just to make sure.

stscieisenhamer: Well, it did work this time. Something else must have been in play at the time. 👍

jwst/datamodels/model_base.py
```diff
@@ -174,12 +174,11 @@ def __init__(self, init=None, schema=None, memmap=False,
                 else:
                     init_fitsopen = init
 
-                with fits.open(init_fitsopen, memmap=memmap) as hdulist:
-                    asdffile = fits_support.from_fits(hdulist,
-                                                  self._schema,
-                                                  self._ctx,
-                                                  **kwargs)
-                    self._files_to_close.append(hdulist)
```
jwhite3, line 182: Perhaps the intent was that the underlying fits file wasn't meant to remain open after reading with `DataModel`? In that case this would make sense because you'd still need to be careful about closing the fits file even after using a context manager (especially with `memmap=True`) since `io.fits` makes it possible to access memory-mapped files even outside of the context. 

On the other hand, if the underlying fits file is supposed to remain accessible after reading in with a `DataModel`, then I agree, this doesn't make sense to me either.

jwst/datamodels/tests/test_models.py
jwhite3: Weird!

jwst/datamodels/model_base.py
```diff
@@ -305,7 +308,7 @@ def _drop_array(d):
         _drop_array(self._instance)
 
     def close(self):
-        if not self._iscopy and self._asdf is not None:
+        if not self._iscopy and getattr(self, '_asdf', None) is not None:
```
jdavies-st, line 311: `self._asdf` should always exist, as it contains the `AsdfFile` instance.  When does it not exist?

jwst/datamodels/tests/test_models.py
```diff
@@ -615,12 +615,17 @@ def test_info():
...
+    """
     schema = ImageModel((10, 10))._schema.copy()
     schema['properties']['meta']['properties']['calibration_software_version']['fits_required'] = True
 
-    with pytest.raises(jsonschema.ValidationError):
-        with ImageModel(FITS_FILE, schema=schema, strict_validation=True):
-            pass
+    with fits.open(FITS_FILE) as hduls:
+        with pytest.raises(jsonschema.ValidationError):
+            ImageModel(hduls, schema=schema, strict_validation=True)
```
jdavies-st, line 628: Why doesn't `with ImageModel():` context manager close it?  It should.

jwst/datamodels/model_base.py
jdavies-st: `hdulist` is not accessed anywhere else in __init__`.  The line is to extract the `AsdfFile` instance from the hdulist["ASDF"] extension and validate it using the given schema.  Once this is done, the hdulist is no longer needed here I think.

Would it make sense to leave the `with` context manager and not append `hdulist` to `self._files_to_close`?

stscieisenhamer: As per concensus during SCSB meeting [Datamodels, ASDF, and FITS efficiency](https://innerspace.stsci.edu/display/SCSB/Datamodels%2C+ASDF%2C+and+FITS+efficiency), the coded behavior will remain the same. Any resource opened by `DataModel` will remain open until `DataModel` is itself closed.

jwst/datamodels/tests/test_models.py
stscieisenhamer: That behavoir would be bad form. Nothing, if passed on open resource, should inherently close that resource. `DataModel` should not presume intent. The application may be doing something else with that open resource.

stscieisenhamer:
Regarding the comment beginning with "hdulist is not accessed anywhere else in init...". This may be true for `DataModel`, but, as discussed during the team meeting, it is accessed by the underlying `asdf` I/O.

jwst/datamodels/model_base.py
stscieisenhamer: Hmmm. there was a test which required this, but reverting is not finding a failing test. It may be this change should actually be part of the FITS-hash PR. I will revert this and we'll wait and see if it crops up again.

jwst/stpipe/tests/test_pipeline.py
```diff
@@ -60,9 +60,9 @@ class MultiplyBy2(Step):
 
     def process(self, image):
         with datamodels.ImageModel(image) as dm:
-            with datamodels.ImageModel() as dm2:
-                dm2.data = dm.data * 2
-                return dm2
+            dm2 = datamodels.ImageModel()
+            dm2.data = dm.data * 2
+            return dm2
```
jdavies-st, line 65: Is removing the with necessary since `dm2` is not associated with any file? Agree that your fix makes it clear that the `with` context manager is not necessary.

jwst/datamodels/model_base.py
```diff
@@ -274,6 +273,10 @@ def __repr__(self):
 
         return "".join(buf)
 
+    def __del__(self):
+        """Ensure closure of resources when deleted."""
+        self.close()
```
jdavies-st, line 278: Does this do anything?  I.e. shouldn't `close` already have been called before the `__del__` destructor ever gets invoked?

jwst/stpipe/tests/test_pipeline.py
stscieisenhamer: With the current API, using the context manager on a `DataModel` that is being passed on is actually wrong. If a `DataModel` is to be treated as an I/O stream, then having a model be returned after leaving a context where it would be closed is simply against the API. Returning a `DataModel` implies that said `DataModel` will be accessed/operate on/modified later, so closing the `DataModel` is semantically incorrect.

Agreed that today, with the code base as per this PR, this example is works, since it has no other resources opened. However, as per the API, this is not guaranteed and could very well change in the future.

Also agreed that this does need to be worked out a bit more. JP-1568 has been created to account for such issues.

jwst/datamodels/model_base.py
stscieisenhamer: This is to handle the situation of when a `DataModel` is deleted without being closed, usually when the object is garbage collected.  This was prompted by unit tests.

However, as before, there is still some discussion to be had around close vs deletion, as noted in JP-1568

PR #4931, opened 2020-05-12
Fix flake8 F523 and F541 bugs dealing with .format() and f-strings

jdavies-st:
The new version of `flake8` has uncovered some long-existing string formatting bugs in our code.  And some new-ish ones.  Seen here

https://travis-ci.org/github/spacetelescope/jwst/jobs/685867988#L216-L220

```
./jwst/stpipe/subproc.py:64:28: F523 '...'.format(...) has unused arguments at position(s): 2
./jwst/srctype/srctype.py:56:29: F541 f-string is missing placeholders
./jwst/srctype/srctype.py:60:25: F541 f-string is missing placeholders
./jwst/engdblog/tests/test_engdblog.py:35:12: F523 '...'.format(...) has unused arguments at position(s): 0
```

And of course these are failing on `master` in our PEP8 check.  This PR fixes them.

### DIFF
--- a/jwst/engdblog/tests/test_engdblog.py
+++ b/jwst/engdblog/tests/test_engdblog.py
@@ -32,7 +32,7 @@ def test_barestring(caplog, engdb):
     assert mnemonic in result
     assert 'EngDBLogStep instance created' in caplog.text
     assert mnemonic in caplog.text
-    assert "Step EngDBLogStep running with args ('INRSI_GWA_Y_TILT_AVGED',).".format(mnemonic) in caplog.text
+    assert f"Step EngDBLogStep running with args ('{mnemonic}')." in caplog.text
     assert '{}[2016-01-01:2016-01-31] = '.format(mnemonic) in caplog.text
     assert 'Step EngDBLogStep done' in caplog.text
 
--- a/jwst/srctype/srctype.py
+++ b/jwst/srctype/srctype.py
@@ -53,11 +53,11 @@ def set_source_type(input_model):
             user_type = input_model.meta.target.source_type_apt
             log.info(f'Input SRCTYAPT = {user_type}')
             if user_type is None:
-                log.warning(f'SRCTYAPT keyword not found in input; using SRCTYPE instead')
+                log.warning('SRCTYAPT keyword not found in input; using SRCTYPE instead')
                 user_type = input_model.meta.target.source_type
                 input_model.meta.target.source_type_apt = user_type
         except AttributeError:
-            log.warning(f'SRCTYAPT keyword not found in input; using SRCTYPE instead')
+            log.warning('SRCTYAPT keyword not found in input; using SRCTYPE instead')
             user_type = input_model.meta.target.source_type
             input_model.meta.target.source_type_apt = user_type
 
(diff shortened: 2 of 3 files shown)


jwst/engdblog/tests/test_engdblog.py
```diff
@@ -32,7 +32,7 @@ def test_barestring(caplog, engdb):
     assert mnemonic in result
     assert 'EngDBLogStep instance created' in caplog.text
     assert mnemonic in caplog.text
-    assert "Step EngDBLogStep running with args ('INRSI_GWA_Y_TILT_AVGED',).".format(mnemonic) in caplog.text
+    assert f"Step EngDBLogStep running with args ('{mnemonic}')." in caplog.text
```
hbushouse, line 35: Was the original line including the mnemonic name twice: once in the hardcoded 'INRSI_GWA_Y_TILT_AVGED' and then a second time via '.format(mnemonic)'?

jwst/srctype/srctype.py
```diff
@@ -53,11 +53,11 @@ def set_source_type(input_model):
             user_type = input_model.meta.target.source_type_apt
             log.info(f'Input SRCTYAPT = {user_type}')
             if user_type is None:
-                log.warning(f'SRCTYAPT keyword not found in input; using SRCTYPE instead')
+                log.warning('SRCTYAPT keyword not found in input; using SRCTYPE instead')
```
hbushouse, line 56: So just for my own knowledge, you don't include the 'f' formatting specifier when it's just a plain text string with no values from variables to be inserted?

jwst/engdblog/tests/test_engdblog.py
jdavies-st: Good question.  I think it should be included once.  Hard to know, as this test is set to `xfail`.

stscieisenhamer: just a buggy line never checked for due to the specified `xfail`. Fix is correct.

jwst/srctype/srctype.py
stscieisenhamer: harsh pep8 checking *oof*

jdavies-st: Correct.  It doesn't result in an error, but it adds extra processing time because the interpreter has to treat it as an f-string, and look for expressions to evalutate at runtime.  `flake8` considers that on par with unnecessary imports and other "cruft" in code.

```
In [1]: %timeit foo = "This is a nice string"
12.8 ns ± 0.14 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)

In [2]: %timeit foo = f"This is a nice string"
13.1 ns ± 0.134 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)

In [3]: %timeit foo = f"This is a nice {'string'}"
71.6 ns ± 0.36 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

jwst/stpipe/subproc.py
```diff
@@ -61,7 +61,7 @@ def process(self, *args):
         newargs = []
         for i, arg in enumerate(args):
             if isinstance(arg, datamodels.DataModel):
-                filename = "{0}.{1:04d}.{1}".format(
+                filename = "{0}.{1:04d}.{2}".format(
```
stscieisenhamer, line 64: And here is the good reason for harshness. Wow, this would have never worked.  *oof*

jdavies-st: Agreed.  I'm happy to accept harsh checking if it uncovers actual bugs, as is the case here.

PR #4936, opened 2020-05-12
Update minumum dependencies

### LINKED ISSUE #4921: Add TravisCI job with oldest supported dependencies
And automate the install based off our requirements in `setup.py`.  This will help us keep our minimum dependencies up-to-date.

jdavies-st:
We have not updated the minimum versions for several of our package dependencies in years (ever?).  This is an attempt to do that and to automate it so that they stay updated in the future.

- Update the `install_requires` and `test` dependencies to a minimum state that work with our code today.
- Add `minimum_deps` script that can create a `requirements-min.txt` file from `setup.py`.  This won't become a static file in our repo.  It gets generated by the script.  Thanks for the discussion @eslavich!
- Add Travis CI job that runs our unit tests with these minimum dependencies.

So in the future, if a new PR adds a line of code that only works in `numpy 1.18`, then the above Travis CI job will try to run it with `numpy 1.16.0` (or whatever the minimum requirement we have is), and it will fail.  The developer will then have the option of not using that line of code, or can have a conversation with the rest of the team about why we should add `numpy>=1.18` to our package install requirements.

This also consolidates the Travis CI job matrix:
- test with latest versions of dependencies
- test with oldest allowed versions of dependencies
- test with what we are delivering to SDP
- test against development versions of dependencies (allowed failure)
- test against CRDS_CONTEXT=jwst-edit (allowed failure)

Also updated the Gaia unit tests that were recently added.  They were failing under the old versions of `pytest-openfiles`, leaving the asdf file open.

Resolves #4921


scripts/minimum_deps
```diff
@@ -0,0 +1,45 @@
...
+import urllib.request
+import json
+import warnings
+
+
+def get_minimum_version(requirement):
+    """Return minimum version available on PyPi for a given version specification"""
+    content = urllib.request.urlopen(f'https://pypi.python.org/pypi/{requirement.name}/json').read()
+    versions = sorted([pkg_resources.parse_version(v) for v in json.loads(content)["releases"].keys()])
+    # print(requirement)
```
eslavich, line 15: Coupla stray prints here

scripts/minimum_deps
```diff
@@ -0,0 +1,45 @@
...
+    # print(versions)
+    for version in versions:
+        if version in requirement:
+            return version
+    # If there was no version specifier in setup.py, return the 2nd lowest
```
eslavich, line 20: I don't think this will actually happen if there was no version specifier -- the `pkg_resources.Requirement` object returns True for all versions if the specifier is missing.  One situation that will cause this is when setup.py contains a requirement for which no version exists.  Seems like that should result in an exception?

scripts/minimum_deps
```diff
@@ -0,0 +1,45 @@
...
+    # version available on PyPi, avoiding 0.0 versions
```
eslavich, line 21: I looked a few packages and discovered that the `0` version is not consistent.  The earliest version of `asdf` is `0`, but for `scipy` it's `0.4.4`.

scripts/minimum_deps
jdavies-st: Ah, right.  It looks it grabs the first version it comes across if there's no version set in `setup.py` requirements.  Hmm.

That said, the purpose of the script is to maintain minimum versions.  So now we have minimum versions for all packages.

I should fix this though.

scripts/minimum_deps
jdavies-st: Right, I saw that too.

scripts/minimum_deps
jdavies-st: Fixed.

scripts/minimum_deps
jdavies-st: I've clarified comments and added some warning messages.  The script can be used as is I think to iteratively find a solution for the lowest useable version of a package dependency (which is how I used it).  Once the lowest versions are set in `install_requires`, then it should never hit that logic in the code.

I.e. instead of raising an error, I want it to return a possibly not-working version of the dependency so that it can be tested and shown to be too low a version.  This will force maintainers to update that package in `install_requires` to one that works.

PR #4940, opened 2020-05-13
JP-1444: Remove path from SCATFILE keyword for NIRISS WFSS in Spec2Pipeline

### LINKED ISSUE #4923: WFSS SCATFILE keyword value needs to have path stripped
_Issue [JP-1444](https://jira.stsci.edu/browse/JP-1444) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

B7.4.2 calspec3 processing of NIRISS WFSS observations in pid 625 obs 30 fails due to schema validation errors for the SCATFILE keyword value in the input cal products, which have a very very long path name included in the file name (e.g. "/ifs/archive/test/jwst/store/doggett/tests/nOPS6/dtest1/JWSTDP-2020.1.0-62~9db2427c/2020-03-04-035303/sdp/asn_creation/cal/level2/jw00625-o030_t006_niriss_clear-f090w_cat.ecsv")

In normal Ops processing the source catalog file name is retrieved by the calwebb_spec2 pipeline from the input spec2 ASN file and the value gets stored in the SCATFILE keyword. This operation takes place here: <https://github.com/spacetelescope/jwst/blob/master/jwst/pipeline/calwebb_spec2.py#L167>

The file name value needs to have the path trimmed before getting stored in the meta attribute.

Note that #4597 made a similar change to strip the path from the value stored in SCATFILE, but that was within the source_catalog step and pertains to the source catalog file generated by that step for imaging data. This use of SCATFILE is distinct and pertains to WFSS data going through the calwebb_spec2 pipeline.

This problem is preventing any level-3 products from being created for WFSS observations.

jwhite3:
This PR removes the path from the source catalog file before setting it as the `SCATFILE` keyword value in the `Spec2Pipeline`. 

Resolves #4923 / [JP-1444](https://jira.stsci.edu/browse/JP-1444)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,7 +22,7 @@ pipeline
   types from an input ASN. [#4937]
 
 - Updated the calwebb_spec2 pipeline to only use the basename of the source
-  catalogue file when updating the source_catalogue keyword for WFSS inputs.
+  catalog file when updating the source_catalogue keyword for WFSS inputs.
   [#4940]
 
 rscd
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,10 @@ pipeline
 - Updated calwebb_image3 pipeline to only load science and background member
   types from an input ASN. [#4937]
 
+- Updated the calwebb_spec2 pipeline to only use the basename of the source
+  catalogue file when updating the source_catalogue keyword for WFSS inputs.
+  [#4940
+
 rscd
 ----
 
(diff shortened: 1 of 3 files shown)


jwhite3:
Note that this PR will result in regression test failures for `test_niriss_wfss` tests. The reason for this is that beforehand `SCATFILE` was being appended to the the list of ignored-keywords in the `fits.Diff`, which has now been removed.

jwhite3:
I was just waiting for the PR number I swear...

hbushouse:
We should put checks for "catalogue" and "colour" in flake8 ...

jwhite3:
Or I could just update my vocabulary to not include "ye olde english"

PR #4949, opened 2020-05-14
Performance improvement in GAIA alignment and bug fix wcs name

mcara:
This PR turns off optimization of the order in which catalogs are aligned to GAIA since there is only one "group catalog" and this turns off computation of image footprints on the sky.

Bug fix: the way previous code was written, it would have resulted in `wcs.name` being overwritten (actually lost) in the next loop. This PR fixes that.

If this PR is merged before next release - no changelog entry should be required.


mcara:
Ah, it also updates requirements to `tweakreg>=0.6.4`

hbushouse:
Fixes JP-1441

Issue #4663, opened 2020-03-06
Reading MultiSlit model is very slow


stscijgbot:
Issue [JP-1343](https://jira.stsci.edu/browse/JP-1343) was created by Nadia Dencheva:

WFSS exposures through the pipeline are processed relative fast now. A recent test with ~800 slits took ~10 min through spec2 and saving the `_cal` product to disk. The resulting `_cal` file has ~8800 FITS extensions.

Opening the file with datamodels takes an extremely long time (I waited for a day and canceled the process). 

 

OTH opening the ASDF extension of that file with pure asdf and reading the tree is not slow so the bottleneck is in datamodels.

stscijgbot:
Comment by James Davies: Can you put an example _cal file on central store somewhere?

nden:
/user/dencheva/grisms

stscijgbot:
Comment by Howard Bushouse: Another example has been copied to /grp/jwst/ssb/chartreuse/dms_bugs/jp-1343/final_output_caldet1_NRS1_srctype.fits. It came from a tester in the INS NIRSpec team and contains 268 extracted slit instances, for a total of 1877 FITS extensions. It takes mere seconds to load using astropy.io.fits.open(), but about 90 minutes to load using jwst.datamodels.open().

stscijgbot:
Comment by Howard Bushouse: Comments/suggestions from [~dencheva]:
{noformat}
1) In a MultiSlitModel have .slits be a list of SlitData models, not Slit models 
and don't run .update() on them because every Slit model now contains the entire 
core schema values. Megan says she thought there was a reason to do SlitModel 
instead of SlitData model but I've done some preliminary tests and it seems this 
is fine to do. (extract_2d){noformat}
{noformat}
2) Provide an option to not run datamodels.properties.load_from_schema, because 
it checks every subschema for FITS_KEYWORD and gets the FITS keyword from the 
fits_hdu hdu. This option should be off by default, but on in the pipeline. The 
reason we have this behavior is to ensure updates to the FITS files are picked 
up by the data model (which isn't necessary in the pipeline setting).
{noformat}

stscijgbot:
Comment by Howard Bushouse: More from [~dencheva]:
{noformat}
3) Provide an option to not validate (off by default) (in datamodels){noformat}
[...]

PR #4951, opened 2020-05-15
Limit looping over FITS HDU's for schema arrays

stscieisenhamer:
WIP [JP-1343](https://jira.stsci.edu/browse/JP-1343) / #4663 

Looping was being performed over all available HDU's
to resolve arrays. However, only the maximum EXTVER for
any named HDU need be considered; anything after that will
guarantee no data.

### DIFF
--- a/jwst/datamodels/fits_support.py
+++ b/jwst/datamodels/fits_support.py
@@ -457,7 +457,7 @@ def _load_from_schema(hdulist, schema, tree, context):
 
     # Determine maximum EXTVER that could be used in finding named HDU's.
     # This is needed to constrain the loop over HDU's when resolving arrays.
-    max_extver = calculate_max_extver(hdulist)
+    max_extver = max(hdu.ver for hdu in hdulist) if len(hdulist) else 0
 
     def callback(schema, path, combiner, ctx, recurse):
         result = None
@@ -597,23 +597,3 @@ def from_fits_hdu(hdu, schema):
         data._coldefs._listeners = listeners
 
     return data
-
-
-def calculate_max_extver(hdulist):
-    """Return the maxiumum EXTVER possible
-
-    Parameters
-    ----------
-    hdulist: FITS hdulist
-        The HDUlist.
-
-    Returns
-    -------
-    max_extver: int
-        The maximum EXTVER for any named HDU.
-    """
-    if not len(hdulist):
-        return 0
-    hdu_counts = Counter(hdu.name for hdu in hdulist)
-    max_extver = max(hdu_counts.values())
-    return max_extver
--- a/jwst/datamodels/fits_support.py
+++ b/jwst/datamodels/fits_support.py
@@ -1,4 +1,3 @@
-from collections import Counter
 import datetime
 import os
 import re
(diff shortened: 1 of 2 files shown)


jwst/datamodels/fits_support.py
```diff
@@ -589,3 +597,23 @@ def from_fits_hdu(hdu, schema):
...
+    Returns
+    -------
+    max_extver: int
+        The maximum EXTVER for any named HDU.
+    """
+    if not len(hdulist):
+        return 0
+    hdu_counts = Counter(hdu.name for hdu in hdulist)
+    max_extver = max(hdu_counts.values())
+    return max_extver
```
jdavies-st, line 619: You can probably avoid this function and the `Counter` instance with

```
max_extver = max([hdu.ver for hdu in hdulist])
```

jdavies-st: Also, max_extver is always going to be 1 or more if there are any HDUs, which of course there has to be at least 1 for a valid FITS file.  And `hdu.ver` is 1-indexed, not zero-indexed.  I don't know how that affects the `recurse` call in the `callback()` function above.

stscieisenhamer: Oh, oh good optimization. 👍 

About the 1-index: That is handled in the schema walking. And yes, there will always be at least 1 HDU. However, note that this is invoked only when there is an array being handled, so there really aught to be at least one to loop anyways.

jwst/datamodels/fits_support.py
```diff
@@ -489,7 +495,7 @@ def callback(schema, path, combiner, ctx, recurse):
         if schema.get('type') == 'array':
             has_fits_hdu = _schema_has_fits_hdu(schema)
             if has_fits_hdu:
-                for i in range(len(hdulist)):
+                for i in range(max_extver):
```
hbushouse, line 498: So help me understand what was and is now happening. Say you have a file that had 5 tuples of (sci,err,dq), for a total of 15 hdu's (16 counting the primary). In this case `len(hdulist)=16`, while `max_extver=5`, and hence you're reducing the number of searches by a factor of 3?

stscieisenhamer: And we have a winner! Yep, that is exactly it. The [original issue](https://jira.stsci.edu/browse/JP-1343) has a full explanation, including suggestions on where to go from here.

hbushouse: Amazing that such a simple 2-3 line change can get us a factor 7x reduction in processing time. Awesome!

hbushouse: Makes me worry about how many other similar instances are lurking all over the depths of the datamodels code ...

stscieisenhamer:
Regarding the schema+FITS+ASDF code, as stated in the main issue, there isn't much left to explore. There may be minor optimizations, but nothing grand as this particular instance. As per a standup discussion, a possibility is to work backwards: Take what exists and only pick out those parts of the schema to validate. However, two issues with that: First, to do such an operation, the schema has to be walked anyways to determine structure. Second, the matter of the schema defaults: They would still need to be found, and hence the schema is again walked. The walking itself is actually very tight; such code is pretty standard for any tree structure API.

Beyond what is further discussed in the original issue, there is actually some confidence that this was the 1-sigma improvement; all else would be 2 or 3 sigma fixes. Well, barring a complete refactor/redefinition of `DataModels`, like not doing the FITS thing at all 😄

hbushouse:
@stscieisenhamer are you OK with merging this so that we could kick the tires with a regtest run?

hbushouse:
OK, kick starting a regtest run ...

jwst/datamodels/fits_support.py
jdavies-st: Cool.  Understand it better now.  Thanks.

This PR is mega! 🚀

PR #4958, opened 2020-05-18
Refactor photom to also update multislit models' bunit keys

jwhite3:
This PR resolves [JP-1434](https://jira.stsci.edu/browse/JP-1434). 

Fixes a bug where input `MultiSlitModel` `bunit` keys were not being updated with new flux units which resulted in incorrect units in the headers of the individual slit products due to header blending down the line

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,11 @@ mrs_imatch
 - Fix ``mrs_imatch`` to avoid calls to ``sigma_clipped_stats`` with all-zero
   arrays. [#4944]
 
+photom
+------
+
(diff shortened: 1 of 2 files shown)


jwhite3:
@hbushouse Well sort of, but thankfully the fix seems to be that easy

I think it was a conscious decision for some reason not to include `bunit` keyword updates in the `MultiSlitModel` and only apply them in the contained `SlitModels`. My guess is that this used to not matter since `resample_spec` didn't blend the headers of the individual slit data, but after we enabled header blending, `bunit` was getting "updated" back to the original units of "DN/s"

hbushouse:
Needs a change log entry.

PR #4959, opened 2020-05-18
JP-1151: Fix NIRSpec MOS PPS reference point bug.

### LINKED ISSUE #4291: PPS reference point for MSA shutter location has changed
Issue [JP-1151](https://jira.stsci.edu/browse/JP-1151) was created by James Muzerolle:

There has been a change in how PPS (specifically, the MSA planning tool in APT) calculates the shutter position for each source in a MOS exposure. It now defines the origin as the top *right* corner of the shutter, as opposed to the top left corner.  DMS currently converts the position to a reference frame with the origin at the shutter center (as expected by the instrument model) using the latter definition, which means the slit_x position is now reversed.  The correct transformation from the values in the MSA metafile should now be: slit_x = 0.5 - slit_x(metafile), slit_y = 0.5-slit_y(metafile).  I'm not sure if this has already been communicated; I was not aware of the change until I discovered a discrepancy when comparing the slit_x,slit_y values from the data model of a processed exposure from the NOPS-6 OTB run with the values in the associated MSA metafile.

jwhite3:
This PR resolves #4291 / [JP-1151](https://jira.stsci.edu/browse/JP-1151) and includes a fix in the conversion from the PPS coordinate frame to the Model frame.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ assign_wcs
 
 - Update keyword and attribute usage around SkyObject to reflect updated keywords. [#4943]
 
-- Refactor PPS origin to bottom left rather than top left.
+- Refactor PPS origin of NIRSpec MOS shutters from top left to bottom left. [#4959]
 
 associations
 ------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ assign_wcs
 
 - Update keyword and attribute usage around SkyObject to reflect updated keywords. [#4943]
 
+- Refactor PPS origin to bottom left rather than top left.
+
 associations
 ------------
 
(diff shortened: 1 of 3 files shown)


jwhite3:
I know that unit tests will need to be updated, but I wanted to double check that this was correct before going through and updating those

CHANGES.rst
```diff
@@ -6,6 +6,8 @@ assign_wcs
 
 - Update keyword and attribute usage around SkyObject to reflect updated keywords. [#4943]
 
+- Refactor PPS origin to bottom left rather than top left.
```
hbushouse, line 9: Would be good to be a little more explicit in terms of the instrument mode, i.e. "Refactor PPS origin of NIRSpec MOS shutters to bottom left, rather than top left"

jwst/assign_wcs/nirspec.py
```diff
@@ -647,11 +647,11 @@ def get_open_msa_slits(msa_file, msa_metadata_id, dither_position,
         The source x,y position in the shutter is given in the msa configuration file,
         columns "estimated_source_in_shutter_x" and "estimated_source_in_shutter_y".
         The source position is in a coordinate system associated with each shutter whose
-        origin is the upper left corner of the shutter, positive x is to the right
+        origin is the lower left corner of the shutter, positive x is to the right
         and positive y is downwards.
```
hbushouse, line 651: positive y is now *upwards*

jwhite3: Ah, I missed that one! Thanks

Issue #4556, opened 2020-02-11
Request to skip outlier detection for coronagraphic data


stscijgbot:
Issue [JP-1294](https://jira.stsci.edu/browse/JP-1294) was created by Alicia Canipe:

The Coronagraphy WG is requesting that outlier detection be skipped for coronagraphic data. They are concerned that the speckle noise will be flagged as outliers. This would require an update to the code, the software readthedocs, the JWST Calibration WG specifications, and the JDox pages.

stscijgbot:
Comment by Misty Cracraft: The MIRI team discussed this in our meeting today, and this definitely needs discussion in the calwg. They said outlier detection might be in the wrong place in coron3, so would need moved rather than being completely removed, but a discussion is needed to determine the final arrangement of steps. For now, skipping it as a default might be a workaround. Right now, though, outlier_detection will flag the speckles that remain after psf subtraction is performed.

stscijgbot:
Comment by Karl Gordon: If the threshold noise level used in the outlier detection is measured from the image stack, would this mean the speckles would not be flagged as outliers?  This was the CapWG requested algorithm, but not what is currently implemented.

stscijgbot:
Comment by Karl Gordon: CalWG requested.

stscijgbot:
Comment by Alicia Canipe: Just adding a note to say the request for updating calwebb_coron3 (and calwebb_tso3) outlier detection step to use the image stack instead of propagated uncertainties is here: https://jira.stsci.edu/browse/JP-1280

[~hagan] [~jgirard] [~bsunnquist]   FYI

If CWG and CalWG decide to go a different route, I'll close the ticket.

stscijgbot:
Comment by Alicia Canipe: From [~jgirard] : It seems to me than this flow order would make more sense
{{stack_refs}}
{{align_refs}}
{{outlier_detection}}
{{klip}}
{{resample}}

stscijgbot:
Comment by Brendan Hagan: Thanks [~acanipe]. The CWG is still discussing about how best to handle this step and what recommendations we should make to the CalWG.

stscijgbot:
[...]

PR #4968, opened 2020-05-19
JP-1294 Update coron3 for new outlier detection application

stscieisenhamer:
Resolves [JP-1294](https://jira.stsci.edu/browse/JP-1294) / #4556

### DIFF
--- a/docs/jwst/pipeline/calwebb_coron3.rst
+++ b/docs/jwst/pipeline/calwebb_coron3.rst
@@ -17,24 +17,25 @@ below.
 +---------------------------------------------------+
 | calwebb_coron3                                    |
 +===================================================+
+| :ref:`outlier_detection <outlier_detection_step>` |
++---------------------------------------------------+
 | :ref:`stack_refs <stack_refs_step>`               |
 +---------------------------------------------------+
 | :ref:`align_refs <align_refs_step>`               |
 +---------------------------------------------------+
 | :ref:`klip <klip_step>`                           |
 +---------------------------------------------------+
-| :ref:`outlier_detection <outlier_detection_step>` |
-+---------------------------------------------------+
 | :ref:`resample <resample_step>`                   |
 +---------------------------------------------------+
 
 The high-level processing provided by these steps is:
 
-1) Accumulate all reference PSF images into a single product
-2) Align every PSF image to every science target image
-3) Compute an optimal PSF fit and subtract it from every science target image
-4) Compare the PSF-subtracted target images against one another to identify outliers
-5) Combine the PSF-subtracted and CR-flagged images into a single resampled image
+1) CR-flag all PSF and science target exposures
+2) Accumulate all reference PSF images into a single product
+3) Align every PSF image to every science target image
+4) Compute an optimal PSF fit and subtract it from every science target image
+5) Compare the PSF-subtracted target images against one another to identify outliers
+6) Combine the PSF-subtracted and CR-flagged images into a single resampled image
 
 Currently the individual steps shown above can only be run in a convenient way by
 running the ``calwebb_coron3`` pipeline on an association (ASN) file that lists the
@@ -54,7 +55,7 @@ Inputs
 :File suffix: _calints
 
 The input to ``calwebb_coron3`` must be in the form of an ASN file that lists
-one or more exposures of a science target and a reference PSF target.
+one or more exposures of a science target and one or more  reference PSF targets.
 The individual target and reference PSF exposures should be in the form of 3D
 calibrated ("_calints") products from :ref:`calwebb_image2 <calwebb_image2>`
 processing. Each pipeline step will loop over the 3D stack of per-integration images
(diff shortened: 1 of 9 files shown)


jwst/pipeline/calwebb_coron3.py
```diff
@@ -85,21 +85,28 @@ def process(self, input):
...
+
         # Stack all the PSF images into a single CubeModel
         psf_stack = self.stack_refs(psf_models)
         psf_models.close()
 
         # Save the resulting PSF stack
         self.save_model(psf_stack, suffix='psfstack')
 
-        # Call the sequence of steps align_refs, klip, and outlier_detection
+        # Call the sequence of steps outlier_detection, align_refs, and klip.
```
hbushouse, line 100: Minor nit, but should remove the period at the end of this phrase, because the thought is continued on the next comment line.

jwst/pipeline/calwebb_coron3.py
```diff
@@ -118,9 +124,7 @@ def process(self, input):
                 suffix='psfsub', acid=acid
             )
 
-            # Create a ModelContainer of the psf_sub results to send to
-            # outlier_detection
-            self.log.debug('Building ModelContainer of klip results')
+            # Create a ModelContainer of the psf_sub results.
```
hbushouse, line 127: So, for my understanding, it's still necessary to build a `ModelContainer` in order to send the proper "stack" of inputs to the `resample` step? It looks like the `ModelContainer` contains a stack of psf-subtracted images from all integrations from all target exposures, right? Rather than sending to `resample` a `ModelContainer` consisting of multiple `CubeModels`, each of which has it's own stack of images (per integration).

jwst/pipeline/calwebb_coron3.py
```diff
@@ -130,21 +134,6 @@ def process(self, input):
...
-                err_str1 = "Creating Level 2c output with updated DQ arrays..."
-                self.log.info(err_str1)
-                lev2c_model = psf_sub.copy()
-                # Replace Level 2b product DQ array with Level 2c DQ array
-                for i in range(len(target_models)):
-                    lev2c_model.dq[i] = target_models[i].dq
-                lev2c_model.meta.cal_step.outlier_detection = 'COMPLETE'
-                self.save_model(lev2c_model, output_file=target_file,
-                                suffix='crfints', acid=acid)
-
```
hbushouse, line 147: So the level-2c's for both psf and target exposures are now created within the `outlier_detection` step itself, rather than up here in the pipeline?

stscieisenhamer: At the moment, yes. This would be more in-line with other pipelines. The `psfsub` product has always been saved, as an archive product. Once JP-1308 is complete, ideally the DQ should be propagated through. If not, well, that should become a new issue.

jwst/pipeline/calwebb_coron3.py
stscieisenhamer: I'll dive into this a bit more and see if a simplification can happen; less transformation the better.

stscieisenhamer: To resample everything down to a single image, the integrations need to be unwrapped and placed in a single-level list. However, I have refactored this bit to be more generic.

jwst/datamodels/tests/test_cube.py
```diff
@@ -0,0 +1,55 @@
...
+def make_container():
+    """Create the container to test"""
+    size = 5
+    cube = CubeModel((size, size, size))
+    cube.meta.target.proposer_name = 'JWST regresstion test'
+    for idx in range(size):
+        arr = np.zeros((size, size))
+        cube.data[idx] = arr + idx
+        cube.dq[idx] = arr + (idx * 2)
+        cube.err[idx] = arr + (idx * 3)
```
jdavies-st, line 19: Is this part necessary?  The above `cube = CubeModel((size, size, size))` has already initialized the values to zeros.  And it looks like you don't use the values anywhere in the tests.

jwst/pipeline/calwebb_coron3.py
```diff
@@ -118,50 +130,16 @@ def process(self, input):
...
-            # Append results from this target exposure to resample input model
-            for i in range(len(target_models)):
-                resample_input.append(target_models[i])
+            # Split out the integrations into separate models
+            # in a ModelContainer to pass to `resample`
+            for model in psf_sub.to_container():
+                resample_input.append(model)
```
jdavies-st, line 136: Can you just do

```
result = self.resample(psf_sub.to_container())
```
instead?

jwst/datamodels/tests/test_cube.py
stscieisenhamer: oh yea. was going to use that as an actual data check, but that is probably not necessary. Will remove.

jwst/pipeline/calwebb_coron3.py
stscieisenhamer: The resample step is outside of the loop and is called only once to operate on all input science files. This is to pass to resample all the integrations of all the input sciences as a flat list.

jdavies-st: Ah, I missed the indentation. 👍

docs/jwst/pipeline/calwebb_coron3.rst
```diff
@@ -17,24 +17,25 @@ below.
...
-1) Accumulate all reference PSF images into a single product
-2) Align every PSF image to every science target image
-3) Compute an optimal PSF fit and subtract it from every science target image
-4) Compare the PSF-subtracted target images against one another to identify outliers
-5) Combine the PSF-subtracted and CR-flagged images into a single resampled image
+1) CR-flag all PSF and science target exposures
+2) Accumulate all reference PSF images into a single product
+3) Align every PSF image to every science target image
+4) Compute an optimal PSF fit and subtract it from every science target image
+5) Compare the PSF-subtracted target images against one another to identify outliers
```
hbushouse, line 37: This line needs to be removed - it's replaced by item 1 (we still have a total of 5 steps, not 6).

PR #4970, opened 2020-05-20
JP-1462: Fix bug that results in obscure linear algebra error

jwhite3:
This PR resolves [JP-1462](https://jira.stsci.edu/browse/JP-1462) in which an error was being raised when attempting to construct a polynomial fit with `LinearLSQFittter` and `Polynomial1D` with inputs of length 1. The minimum number of "good points" is now 2.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,10 @@
 0.16.2 (unreleased)
 ===================
 
+extract_1d
+----------
 
+- Fix bug in creating a polynomial fit used in background extraction. [#4970]
 
 0.16.1 (2020-05-19)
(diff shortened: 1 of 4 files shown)


stscicrawford:
Good catch to determine the problem!

Can you also:
- add a unit test for `_fit_background_model` ?
- add a note to the `_fit_background_model` and `extract1d ` documentation about the behavior and error cases when using `bkg_order`?

jwhite3:
@stscieisenhamer Right, I wasn't sure how much detail to go into or all of the spots I'd need to mention this update, so I figured I'd start there.

PR #4973, opened 2020-05-21
JP-1308 Correction for bad pixels in Target and PSF images

### LINKED ISSUE #4590: Add a bad pixel correction to calwebb_coron3
Issue [JP-1308](https://jira.stsci.edu/browse/JP-1308) was created by Julien Girard:

{{calwebb_coron3}}'s most critical step is the image registration: {{align_refs}}

It appears that it works fine on the precursor noisefree data used for the initial tests by the Coron Working Group. The reference star images with small shifts (small grid dithers) are properly registered.

But, when we encapsulate those images into the real subarrays and run {{MIRAGE}}, the alignment step does not produce the expected results. (see video made by Ben Sunnquist).

We think this is because of the presence of numerous bad pixels. Brendan Hagan and I think we should add a {{bad_pixel_correction}} step before {{align_refs}} and {{klip}}. {{outlier_detection}} also makes sense to be moved (JP-1251) as such:

{{outlier_detection}}
 {{bad_pixel_correction}} [new step]
 {{stack_refs}}
 {{align_refs}}
 {{klip}}
 {{resample}}

We will investigate how to perform this {{bad_pixel_correction}}. Brendan says that NICMOS PCA ({{KLIP}}) post-processing also required it.

The goal of this issue is therefore to add a step that will make the {{align_refs}} step *robust* to noisy data even for faint stars.

ddavis-stsci:
Fixes #4590 / [JP-1308](https://jira.stsci.edu/browse/JP-1308) by adding bad pixel replacement to target and psf images at the beginning of the `align_refs` step.


jwst/coron/smooth_psf.py
```diff
@@ -0,0 +1,78 @@
...
+    try:
+        data_array = input_array[xc - hbox:xc + hbox, yc - hbox: yc + hbox]
+        dq_array = input_dq_array[xc - hbox:xc + hbox, yc - hbox: yc + hbox]
+    except IndexError:
+        # If the box is outside the data return 0
+        log.warning('Box for median filter is outside the data.')
+        return 0.
+
+    filtered_array = data_array[dq_array != dqflags.pixel['DO_NOT_USE']]
+    median_value = np.median(filtered_array)
```
jwhite3, line 42: Is it expected that `filtered_array` doesn't have `NaN` values at all? Otherwise, could use `np.nanmedian`

jwst/coron/smooth_psf.py
```diff
@@ -0,0 +1,78 @@
...
+        box_size : scalar
+            box size for the median smoothing
+    """
+
+    n_ints, _, _ = psf_model.data.shape
+    for nimage in range(n_ints):
+        psf_int = psf_model.data[nimage]
+        psf_dq = psf_model.dq[nimage]
+        bad_locations  = np.where(np.equal(psf_dq, dqflags.pixel['DO_NOT_USE']))
+        if bad_locations:
```
jwhite3, line 67: check out `np.put`: https://numpy.org/doc/stable/reference/generated/numpy.put.html

Might be of use here

jwst/coron/smooth_psf.py
ddavis-stsci: The DO_NOT_USE flag should cover the case for NaN's

jwhite3: Ah, ok, then in that case, when would `np.median` return `NaN`?

ddavis-stsci: As long as the flag is accurate np.median will return a Nan if the no values are left after the flagging. np.nanmedian will also return a NaN in that case. 
Do you have another case in mind?

jwst/coron/smooth_psf.py
ddavis-stsci: I didn't know about this function. Just what I needed!

jwst/coron/align_refs_step.py
```diff
@@ -38,6 +41,16 @@ def process(self, target, psf):
             # Open the input psf images
             psf_model = datamodels.open(psf)
 
+            # Retrieve the lenth of the smoothing box for the filter
```
stscieisenhamer, line 44: Two issues: One is that this check should not be needed. Any `spec` parameter will always be defined. If a parameter is not given a value, and the `spec` does not define a default, the upstream config parsing will raise an exception.

This check is probably in there because there is a name mismatch. The `spec` defines `smoothing_box_length`, but here `self.smoothing_length` is queried for.

jwst/coron/smooth_psf.py
jwhite3: No, I don't think so! I was trying to figure out whether or not we could be more explicit in trying to catch any issues here

jwst/pipeline/align_refs.cfg
```diff
@@ -1,2 +1,4 @@
-name = "alignrefs"
+name = "align_refs"
 class = "jwst.coron.AlignRefsStep"
+
+smoothing_box_length = 4
```
stscieisenhamer, line 4: Hidden developer information that is not clear: Since we're in the process of migrating from `cfg` files to crds-delivered step parameter files, all default specifications have been removed from the `cfg` files. Having the file itself with with the `name` and `class` are fine, but nothing else.

jwst/pipeline/calwebb_coron3.cfg
```diff
@@ -4,6 +4,7 @@ class = "jwst.pipeline.Coron3Pipeline"
     [steps]
       [[stack_refs]]
       [[align_refs]]
-      [[klip]]
+        config_file = align_refs.cfg
```
stscieisenhamer, line 7: Similar to the previous comment of removing defaults, the line with `config_file` should be removed. Having the steps listed is fine, but otherwise no other default specifications.

jwst/pipeline/calwebb_coron3.cfg
```diff
@@ -4,6 +4,7 @@ class = "jwst.pipeline.Coron3Pipeline"
...
+      [[klip]] 
       [[outlier_detection]]
```
stscieisenhamer, line 9: Something I missed in the related moving of the outlier detection PR: The `[[outlier_detection]]` step should now be the first step. This doesn't affect operations, but is just for consistency with order.

jwst/coron/align_refs_step.py
ddavis-stsci: I should never change my mind about a variable name halfway through the process..., thanks for catching this again.

jwst/pipeline/align_refs.cfg
ddavis-stsci: How do I get this into the crds step parameter files?

jwst/coron/align_refs_step.py
```diff
@@ -38,6 +41,16 @@ def process(self, target, psf):
...
+            try:
+                box_size = self.smoothing_length
+            except AttributeError:
+                box_size = 4
+                self.log.warning('Smoothing length not found, set to default')
+
+            # Smooth the psf images
+            psf_model = smooth_psf(psf_model, box_size)
```
hbushouse, line 52: The smoothing operation needs to be applied to both the psf and target images, so this should also be called using `target_model` as input. And given that it's being applied to both sets of images (not just the PSF's) perhaps the name of the routine should be changed to something more generic than `smooth_psf`.

jwst/coron/smooth_psf.py
```diff
@@ -0,0 +1,78 @@
...
+    if np.isnan(median_value):
+        # If the median fails return 0
+        log.warning('Median filter returned N/A setting value to 0.')
+        median_value = 0.
+
+    return median_value
+
+
+def smooth_psf(psf_model, box_size):
+    """ Routine to median smooth the psf images after flagging and replacing the bad pixels.
```
hbushouse, line 53: If this comment is correct, in terms of the order of operations, I think it's backwards of the approach that the CoronWG folks used in their example notebook. They first performed a simple `median_filter` of the entire image. creating a temporary smoothed image, and then copied individual pixel values from that smoothed image to replace values in the original image that have DQ flags set. From their notebook:

```
   for i in np.arange(nints):
        # Get the data for this integration
        data_i = np.copy(data[i])
        dq_i = np.copy(dq[i])
        
        # Identify the bad pixels of interest based on 'dq_bits' as 1
        dq_i[dq_i & dq_bits == 0] = 0
        dq_i[dq_i != 0] = 1
        
        # Median-smooth the science data
        smoothed = median_filter(data_i, box_size)
        
        # Replace the bad pixels with their corresponding median-smoothed values
        data_i[dq_i != 0] = smoothed[dq_i != 0]
        data_new[i] = data_i
```

hbushouse: So only the bad pixels get modified, as opposed to smoothing the entire image.

jwst/pipeline/align_refs.cfg
stscieisenhamer: The fact that in the `spec` the default is defined as 4, there is no need to specify it in the cfg file.  If one wanted to change from the default, it can be added to their local copy of the cfg file. But, as delivered in the package, there is no reason to have defaults specified. This will be true for step parameter files, if and when they are ever created by INS.

jwst/coron/smooth_psf.py
hbushouse: That's correct. The images to be passed along to the rest of the processing (the rest of `align_refs` and subsequent steps) should be the ones that've only had values for bad pixels replaced. Not the ones that've been smoothed over the entire image.

hbushouse: And both the psf and target images need to have this bad pixel replacement applied to them, because the process of aligning is done via some cross-correlation like operation (possibly involving FFT's) between each psf image and each target image, so spurious pixel values in either of them will corrupt the alignment results. The badpix-replaced images for both sets should also then get passed along to subsequent steps.

jwst/pipeline/align_refs.cfg
ddavis-stsci: I've updated the config files and removed the try block.

jwst/coron/smooth_psf.py
ddavis-stsci: The code is updated to replace bad pixels with the median value calculated from a box around the location of the bad pixel. This is applied to both the psf and target image.

jwst/pipeline/calwebb_coron3.cfg
ddavis-stsci: Moved this to the first step

jwst/coron/align_refs_step.py
```diff
@@ -1,6 +1,8 @@
+""" Smooth and align psf image with target image."""
```
hbushouse, line 1: I would reword this to "Replace bad pixels and align psf ..."

jwst/coron/align_refs_step.py
```diff
@@ -38,6 +41,15 @@ def process(self, target, psf):
             # Open the input psf images
             psf_model = datamodels.open(psf)
 
+            # Retrieve the lenth of the smoothing box for the filter
```
hbushouse, line 44: typo in "lenth"

jwst/coron/align_refs_step.py
```diff
@@ -38,6 +41,15 @@ def process(self, target, psf):
...
+            box_size = self.smoothing_box_length
+
+            # Smooth the psf images
```
hbushouse, line 47: Again, I would not refer to this as a smoothing operation. The bad pixel replacement values are being derived by applying a kind of smoothing operation to the images, but overall the resulting images are not being smoothed. We're simply replacing individual bad pixels. So "Replace bad pixels in the psf images" and "Replace bad pixels in the target images".

jwst/coron/median_replace_img.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+
+    n_ints, _, _ = img_model.data.shape
+    for nimage in range(n_ints):
+        img_int = img_model.data[nimage]
+        img_dq = img_model.dq[nimage]
+        # check to see if any of the pixels are flagged
+        if np.count_nonzero(img_dq == dqflags.pixel['DO_NOT_USE']) > 0:
+            bad_locations  = np.where(np.equal(img_dq, dqflags.pixel['DO_NOT_USE']))
+            # fill the bad pixel values with the media of the data in a box region
```
hbushouse, line 68: typo in "media" (should be median)

jwst/coron/median_replace_img.py
```diff
@@ -0,0 +1,77 @@
...
+    hbox = int(bsize/2)
+
+    # Extract the region of interest for the data
+    try:
+        data_array = input_array[xc - hbox:xc + hbox, yc - hbox: yc + hbox]
+        dq_array = input_dq_array[xc - hbox:xc + hbox, yc - hbox: yc + hbox]
+    except IndexError:
+        # If the box is outside the data return 0
+        log.warning('Box for median filter is outside the data.')
+        return 0.
```
hbushouse, line 37: Hmmm, so this is not going to allow for replacement of any bad pixels that're within half the box width of the image borders, instead replacing them with a value of zero. Given that later on in the alignment a weight/mask is applied that essentially gives zero weight to pixels near the image perimeter, hopefully the fact that they haven't had a real replacement done to their values will be OK. I guess we'll see what happens in testing.

jwst/coron/median_replace_img.py
```diff
@@ -0,0 +1,77 @@
...
+
+    filtered_array = data_array[dq_array != dqflags.pixel['DO_NOT_USE']]
+    median_value = np.median(filtered_array)
+
+    if np.isnan(median_value):
+        # If the median fails return 0
+        log.warning('Median filter returned N/A setting value to 0.')
```
hbushouse, line 44: typo: "N/A" should be "NaN"

jwst/coron/align_refs_step.py
ddavis-stsci: Done

jwst/coron/median_replace_img.py
ddavis-stsci: That is what I was counting on here. In general I'd use the median of the outer part of the image, or perhaps the outermost values of a radial profile.

hbushouse:
Just realized this should also include updates to the docs for the `align_refs` step, to add info about the new bad pixel correction/replacement.

docs/jwst/pipeline/calwebb_coron3.rst
```diff
@@ -121,11 +123,17 @@ ASN file, e.g. "jw86073-a3001_t001_nircam_f140m-maskbar_psfstack.fits."
 :Data model: `~jwst.datamodels.QuadModel`
 :File suffix: _psfalign
 
-For each science target exposure, all of the reference PSF images in the
-"_psfstack" product are aligned to each science target integration and saved to
-a 4D "_psfalign" product by the :ref:`align_refs <align_refs_step>` step. The output file
-name is exposure-based, with the addition of the associated candidate ID, e.g.
-"jw8607342001_02102_00001_nrcb3_a3001_psfalign.fits."
+For each science target exposure, all of the reference PSF images in
```
hbushouse, line 126: Oh wait - This info should go into the description section of the `align_refs` step doc, rather than here in the `calwebb_coron3` pipeline doc. https://jwst-pipeline.readthedocs.io/en/latest/jwst/align_refs/description.html
That's why the descriptions here are so terse: there's more detailed info in the docs for each step.

ddavis-stsci: Maybe I'm just very confused. I thought that the html was generated from the rst files with sphinx.

hbushouse: No, you're not confused, just my message was confusing. The rst file that needs updating is https://github.com/spacetelescope/jwst/blob/master/docs/jwst/align_refs/description.rst

Issue #4050, opened 2019-09-20
MRS alpha-beta cubes for commissioning


stscijgbot:
Issue [JP-1013](https://jira.stsci.edu/browse/JP-1013) was created by David Law:

In order to support commissioning activities to determine the MRS PSF (known to be different in the alpha-beta directions) it will be necessary to construct alpha-beta cubes for dithered observations.  These are cubes in which the along-slice and across-slice directions are aligned with the X/Y spaxel axes of the cube.

This has been on the long-term agenda of the MIRI team for a while, but was pending an easy solution of how it might be implemented that is now resolved.

Essentially, what is necessary is a data cube with astronomical RA/DEC coordinates in which the rotation of the cube is set such that the along-slice (alpha) direction is along the X-axis of the cube.  This can be achieved simply by applying a rotation angle Theta when computing the Xi-Eta pixel coordinates from the input RA/DEC coordinatates, where Theta is defined as the position angle of the along-slice direction in the first exposure (computed using the embedded wcs).  Specification of a keyword when running cube_build would thus determine whether the position angle of the cube is 0 (which it always is at the moment) or allow it to be aligned with the IFU hardware.  Both kinds of cubes would be equally astronomically valid.

Technical note: alpha may not increase to the right, and beta may not increase up in these cubes.  The alpha/beta coordinate system contains some flips relative to V2/V3 (and by extension RA/DEC) that should not be reproduced so that these cubes are simply rotated with respect to the standard PA=0 cubes.  This may mean, for instance, that beta increases downwards.

stscijgbot:
Comment by Alicia Canipe: Added to the list of issues for prioritization: [https://outerspace.stsci.edu/display/JWSTPWG/Calibration+Pipeline+Deadlines+and+INS+Issues]

stscijgbot:
Comment by David Law: From an email thread:
Rather than trying to do any actual cube construction in a distorted reference frame (requiring mapping coordinates from one image to another), the simplest way to implement it is as a rotation in the xi-eta cube coordinate frame with respect to RA/DEC.  
 
[...]

PR #4974, opened 2020-05-21
Derotated cubes

jemorrison:
@drlaw1558 
This is my first attempt at aligning the cubes with alpha-beta plane. I more work to do. The rotation angle is not exactly right. I cheated and just did what you did but I picked the wavelength to be the median wavelength  alpha, beta (0,0) and alpha, beta (0, -1)
Maybe I should do something more complex. 
Also for now I figured out the size of the cube in the unrotated system - I can figure it out after rotation - for now I just wanted to have the same size images to see if they looked about right. 
In the image blow left unrotated, right rotated. The unrotated one looks like a square in a rotated system - the rotated one looks more like an rectangle. So I might have something wrong here.
Screen Shot 2020-05-21 at 2 43 19 PM


jemorrison:
By the way the rotation angle was -112 deg. I used det_image_seq1_MIRIFUSHORT_12LONGexp1_cal.fits from that set of dithered images you pointed me at in Feb.

hbushouse:
Fixes #4050 / [JP-1013](https://jira.stsci.edu/browse/JP-1013)

jemorrison:
@drlaw1558 
I think this PR is about done and ready for you and James M to test. One last thing - is when the cubes are built in 'ifualign' the output cubes are rotated but because of how I determined the size on the sky they have an extra padding around the edges that I would like to get rid of.
I set up the size of the cubes in ifu_cube.set_geometry.   I use the min and max ra and dec on the sky to set up the size. I have played around with different methods but I still can not get it trimmed correctly. Do you have a suggestion ?  Below is a picture of what I mean 
Screen Shot 2020-06-08 at 10 11 57 AM

jemorrison:
@drlaw1558 
I am not sure the new way of determining the size of the output IFU cube (in ifu_cube.set_geometry) is correct. I am mapping the ra, dec min max to the tangent plane and determining the size by the length of the distance between these corners. 
The pytest test_setup_wcs now have the axes lengths switched - probably because this test is a dec=45. 
The brute force method would be to map all the ra and dec to tangent plane and then find size but would be slower.

jemorrison:
@drlaw1558 
In a picture I drew it seemed that all I need to do is multiple the distance is set_geometry  by cos of rotation angle. I tried it and it seemed to work. I am not sure if it a general fix. Let me know what you think.  I will try it on more examples

drlaw1558:
Hm, you mean just multiplying the corner distance by cos(rotangle)?  I don't think that could work as a general solution, if it were 90 degrees then that would shrink the field of view to zero.  I think part of the problem is that when we take min/max of the input ra/dec values then we get the ra of the left and right corners, and the dec of the top and bottom, but not full ra/dec information about any corner.
Another option would be to identify the input pixels where the min/max ra/dec values occur, and get both ra/dec values for these pixels.  That would give you full position information for all four corners, and then those four corner coordinates could be mapped to the rotated tangent plane to figure out the necessary boundaries.  Might be faster than mapping all of the input points just to find the min/max.

jemorrison:
@drlaw1558 
I can not figure out the correct size of the rotated 'ifualign' cube.
I map all the points to ra, dec. I store the ra, dec corners (ra max, associated dec value), (ra min, associated dec value), (associated ra, dec min), (associated ra, dec max)
Then using the ra, dec center (still defined by ra center = (ra min + ra max)/2 and dec_center = (dec_min + dec max)/2 I convert these corners on the sky to the tangent plane values and using the rotation angle determined. 
I then find the min and max of the 4 corners to set up the size of the ifu cube.
This is not correct.
Do you see what I am doing wrong ?

drlaw1558:
@jemorrison A few quick thoughts on looking through the code:
- set_geometry doesn't have a default value for rot_angle, so it's crashing when I run using 'skyalign' at the moment
- The distances in set_geometry are getting multiplied by the cosine of rot_angle; in the test case I'm working with this is producing negative distances which confuses things
- set_geometry is only being passed four numbers spatially; ramin/max, and decmin/max.  It doesn't actually have the ra/dec values of the four corners; as such it only knows how to make the bounding box bigger, not smaller.  If we compute ra1/dec1, ra2/dec2, ra3/dec3, ra4/dec4 of the real corners and pass those in as the footprint, then the radec2std call in set_geometry should automatically get the right xi1/eta1, etc.  xi_min will then just be np.min([xi1,xi2,xi3,xi4]) and etc without needing to do any computation of distances and cosine terms.

jemorrison:
Ok I guess I did not push over the new code. For 'ifualign' I use set_geometry_ifu (for now). Once we figure this out I can probably have one generic module.  I will push over the code now

jemorrison:
@drlaw1558 
After looking at the code once more I think I fixed the problem with the ifualign cubes.
Test this PR and see what you think. All the types should be working ' sky align', 'ifualign' and 'internal_cal'
I still need to clean up the code and remove commented out code and unused modules - but before I do that I want to make sure it is working as expected. I also have to fix some of the unit tests.

drlaw1558:
That looks much better.  Just a few comments:
- I might be seeing an unexpected 90 degree rotation compared to before; not sure whether this is the code or my test data though.  The Leuven server seems to be offline, so it might take me a little while to sort out what's going on here, stay tuned.
- It looks like for some reason the field of view is cropped a little too close.  The ifualign cubes have no blank space around them at all, while I'd expect some simply because the edges of the IFU field are a little ragged.  Likewise, the skyalign cubes seem to be losing the very tips of the IFU field out of the cubes.  Presumably this is something to do with how the edges map into integer pixels.  Can we add a 1-pixel buffer around what the FOV calculation thinks should be the FOV to help deal with this?  (I.e., calculate what we think the box size needs to be, then add 2 more spaxels in width and height?)
- It looks like you've changed the default weight function internal to the cube building code, but I think we need to change it in cube_build.cfg as well (since that would override the code internal version).  Similarly, we'll need to change the default weighting in the config file to away from 'world'

jemorrison:
@drlaw1558 
I am reviewing how the size of the output cube is established to see if it is off by 1 (or why it seems the data is getting cut off).
I did find one thing to improve with the 'internal"cal' cubes. In these types of cubes I force
a 1 to 1 mapping between the beta value and slice number. To do this I have to update the cdelt2 default, which is the default for making sky cubes that will be dithered. So the default cdelt2 is a bit small for these types of cubes. Should I reset the cdelt1 = new cdelt2 (othersize the cubes get a bit stretched out in along slice dimension). I am not sure how much the MIRI team will use these types of cube not that I am updating cube build I wanted to review how these cubes are constructed.

drlaw1558:
Ok, ignore my previous comment about a 90 degree rotation, that was just an oddity of my test data.  With a new fresh simulation I don't see that.
Regarding internal_cal:
- Can we make the error message if trying to pass cube build a multi-exposure association say something more obvious, like 'multiple exposures isn't allowed for choice of internal_cal coordinate system'?  At the moment it reports 'For interpolation = area, only one file can be used to created the cube'
- I don't think these cubes will see a lot of use for MIRI, but I think it would be good to keep a square spaxel size, otherwise it will suggest unless someone looks closely at the headers that the PSF is elongated.  We know that's likely to be the case anyway, so anything that further confuses that measurement isn't ideal.  So- yes, I think set cdelt1 = new cdelt2 for these cubes.

jemorrison:
@drlaw1558  While I have my head back into how cube_build works I am cleaning up a few things.
Do we want to keep weighting = miripsf ?
Weighting based on the psf ? Which maybe we only use for channel 1 a
If we do we should test it - I have not tried to use it in a long time.

drlaw1558:
Personally I'm not keen on the miripsf weighting option; I think it's too complicated, doesn't offer clear advantages over the regular cube building (even at 1A), and it hasn't been tested.  However, it was added based on discussion with the IDT MRS team, so I wouldn't want to remove the code and all of the work that's gone into it, as it may be something that we want to look closely at again after commissioning.  Can we simply disable the option at the top level (so that no end-users get misled by it), but leave all of the code intact?

jemorrison:
@jmuzerolle @drlaw1558 
Just a couple of lingering issues left.
James for NIRSPEC ''internal_cal' IFU cube do you want the wavelength coordinates of the IFU cube to be left in meters ? I converted it to microns because it was easier for me to check that it was working the way it should.
For both of you. When making 'internal_cal' cubes what do we want the ctype1 and ctype2 output WCS to be ? In the case of MIRI - long ago- and I don't know if we want to keep this - we set the ctype1 = MRSAL1A  ctype2 = MRSBE1A for - for channel 1A
so basically MRSAL +'CHANNEL# and SUBCHANNEL' for ctype1 and likewise MRSBE + 'CHANNEL # and SUBCHANNEL'. I am not sure why we did that.  We can keep that for MIRI and add something like it for NIRSPEC. Or just redefine it all together. 
I sort of remember giving Nadia those definition so - I need to check if it is somehow in the gwcs package.

jmuzerolle:
Hi Jane,


I'd prefer microns for the wavelength coordinates.


How about "NRSSLICERX" and "NRSSLICERY" for ctype1/2?  Units in meters is fine for these.


James


drlaw1558:
Keeping the MRSAL1A type structure sounds fine to me

drlaw1558:
@jemorrison Looks like there's some kind of missing keyword error when the mrs_imatch step in spec3 calls cube building now; abbreviated, I get 
  File "/Users/drlaw/jwcode/jwst/jwst/mrs_imatch/mrs_imatch_step.py", line 263, in _match_models
    cube_models = cbs.process(models)
  File "/Users/drlaw/jwcode/jwst/jwst/cube_build/cube_build_step.py", line 337, in process
    cube_container = thiscube.build_ifucube_single()
  File "/Users/drlaw/jwcode/jwst/jwst/cube_build/ifu_cube.py", line 845, in build_ifucube_single
    scalerad_pixel)
TypeError: match_det2cube_msm() missing 2 required positional arguments: 'cube_debug' and 'debug_file'

jemorrison:
@drlaw1558  I fixed the code. Try it again 
I added some debug information in 'ifu_cube.py' that was not there when cube_build is built with 'single' option

drlaw1558:
@jemorrison Looks like that fixed the issue.  IFU boundaries are still seeming to be a bit clipped though; I just submitted a PR against this PR that should fix that.

jemorrison:
@drlaw1558 You mentioned above that you submitted at PR against this PR to make the cubes bigger.
What PR is this change in ? I can not find it.

drlaw1558:
https://github.com/jemorrison/jwst/pull/3

drlaw1558:
Still running a few tests, but generally looking good.  We should update CHANGES.rst and the documentation as well to make sure that the new coordinate system names make it into readthedocs.

jemorrison:
@drlaw1558 Looking at the regression test runs I don't think we need to expand the cubes by 1 pixel. They seem to be too big now and before I don't think any data was cut off. Comments ?

drlaw1558:
cubes
Attached is an image of before/after the 1 pixel expansion; you can see the difference in the extrema.  Arguably 2 pixels might be better, but it's easy enough to change later if necessary.

jemorrison:
@drlaw1558 thanks for those images that helps.

drlaw1558:
I don't understand what the failing tests are about, but from the science side I think this PR is looking good.  Some documentation issues remain, but perhaps it's easiest to punt those to the documentation ticket (JP-1456) instead.

jemorrison:
@drlaw1558 what data set did you make those two - old and new images. I want to double check how the size is determined. I don't know why we need to inflate the size by 1 to 2 pixels. So let me run some tests with whatever data set you are using to see if I can find why the size is too small.

jemorrison:
@drlaw1558 I just tried to run with weighting  = miripsf and cube build does not work with that option.
I have removed it from the options to use for weighting so cubes can not be built this way but if you think
the MRS team might want to try and build cubes like that in commissioning I should probably figure out why 
it does not work. We can open a ticket later on this if we need too.

drlaw1558:
@jemorrison I think that makes sense; that's a separable issue that we can return to when necessary.  Just sent you a link to the test data that I was showing above via email.

jemorrison:
@drlaw1558 
I rediscovered why it looks like the cube size is too small and we have to increase the size by 1 around the border.
The roi spatial size is much larger than the spaxel spatial size. For a 4 dithered data of MIRISHORT channel A the spaxel spatial size is 0.13 arc seconds, but the roi spatial size is 0.31 arc seconds. So even when we increase the size of the cube by 1 on all 4 edges the roi spatial size is large enough that the edge spaxels can find a point cloud member even on the edge. I just wanted to understand why after increasing the size to point that no edge spaxel should have an overlap we still had overlap. I am not going to make any changes - just something to remember for the future.

PR #4977, opened 2020-05-22
Remove packaging dependency; remove vendorized pyparsing

jdavies-st:
We now depend on `packaging` since #4905, but it is not in our requirements.  It only gets pulled in by `pytest`.

So without installing `pytest` currently, our runtime code in `datamodels` doesn't work.  It fails:

```
from jwst.datamodels import DataModel
  File "/Users/jdavies/miniconda3/envs/packaging/lib/python3.8/site-packages/jwst/datamodels/__init__.py", line 5, in <module>
    from .model_base import DataModel
  File "/Users/jdavies/miniconda3/envs/packaging/lib/python3.8/site-packages/jwst/datamodels/model_base.py", line 25, in <module>
    from . import fits_support
  File "/Users/jdavies/miniconda3/envs/packaging/lib/python3.8/site-packages/jwst/datamodels/fits_support.py", line 5, in <module>
    from packaging import version
ModuleNotFoundError: No module named 'packaging'
```

So use `pkg_resources.parse_version` instead.

Also replace the vendorized `pyparsing` in `jwst.extern` with the actual package that is actively developed in is a dependency of basic python infrastructure.


jdavies-st:
Failing tests are unrelated and due to the MAST catalog server being down for maintenance.  Merging.

PR #4875, opened 2020-04-30
Ignore doctestplus import errors in ramp_fit

jdavies-st:
This fix ignores pytest collection errors we've seen with the recently released `pytest-doctestplus 0.6` which now raises a collection error if modules in the repo are not importable.  This is probably good behavior, but we have several modules in the repo that are not importable currently:

```
================================================================= ERRORS =================================================================
__________________________________________________ ERROR collecting docs/exts/numfig.py __________________________________________________
docs/exts/numfig.py:1: in <module>
    from docutils.nodes import figure, caption, Text, reference, raw, SkipNode, Element
E   ModuleNotFoundError: No module named 'docutils'
______________________________________________ ERROR collecting jwst/ramp_fitting/mc_2d.py _______________________________________________
jwst/ramp_fitting/mc_2d.py:43: in <module>
    import energy_dists
E   ModuleNotFoundError: No module named 'energy_dists'
__________________________________________ ERROR collecting jwst/ramp_fitting/rotate_dataset.py __________________________________________
jwst/ramp_fitting/rotate_dataset.py:11: in <module>
    from . import rotate_utils
jwst/ramp_fitting/rotate_utils.py:8: in <module>
    import Rotatefitscube
E   ModuleNotFoundError: No module named 'Rotatefitscube'
___________________________________________ ERROR collecting jwst/ramp_fitting/rotate_utils.py ___________________________________________
jwst/ramp_fitting/rotate_utils.py:8: in <module>
    import Rotatefitscube
E   ModuleNotFoundError: No module named 'Rotatefitscube'
```

PR #4979, opened 2020-05-26
Verify install_requires in setup.py is up-to-date

jdavies-st:
- Add a new `verify_install_requires` script that imports all submodules in the package
- Add a Travis job that runs the above script.  This makes sure that all the packages/modules in `jwst` are importable with the dependencies listed in setup.py `install_requires`.

The key here is that none of the `[test]` dependencies get installed in the Travis job.  So this check happens without `[test]` dependencies.  This is to avoid having our runtime code depend on the `[test]` dependencies which has happened a few times in the recent past.

This is a follow-on to #4977 to verify the dependencies added there and to keep them updated in the future.

Also a follow-on to #4875 in that it removes the non-importable code from `jwst.ramp_fitting`.


jdavies-st:
Yup.  This code is not importable.  It doesn't work.  I will leave it to @dmggh to verify.

dmggh:
Yes, these can be removed: mc_3d.py, rotate_dataset.py, and rotate_utils.py, as neither energy_dists, Rotatefitscube are needed.

PR #4980, opened 2020-05-26
Check astropy 4.1rc1

jdavies-st:
Verifying our unit tests work with `astropy 4.1rc1`.


jdavies-st:
The jwst CI tests with the latest dependencies and `astropy 4.1rc1` pass.  See checks below

But it does not pass with oldest allowed dependencies, i.e. with `asdf 2.5.0`.  And this is somewhat expected.

This happens because astropy lists `asdf>=2.6` has an extras dependency, and these do not get checked at runtime in the same way that it would be if in `install_requires`. So instead one gets cryptic errors when using `asdf<2.6`.  So any package that depends on `astropy` and `asdf` cannot rely on `astropy` to tell it what version of `asdf` it needs.  It needs to do the minimum versioning itself.

So to summarize:

1. Users of `jwst` could use `astropy==4.0` and `asdf>=2.5` together without issue.

2. Users of `jwst` who want to use `astropy==4.1` need `asdf>=2.6`, but astropy doesn't enforce this at runtime, only if you install astropy with it's `[all]` extras.  So it is kicking the ball back up to `jwst` to manage this, which is virtually impossible if we want to maintain compatibility with `astropy==4.0` and `asdf~=2.5.0`.  The only solution to allow use of both versions of astropy is to bump the requirement in jwst that `asdf>=2.6`.

jdavies-st:
The regression tests fail.

```
>               wcs_truth = nirspec.nrs_wcs_set_input(im_truth, name)

../../../jwst/regtest/test_nirspec_wcs.py:48: 
../../../jwst/assign_wcs/nirspec.py:1542: in nrs_wcs_set_input
    slit2detector = slit_wcs.get_transform('slit_frame', 'detector')
../../../../miniconda/envs/tmp_env0/lib/python3.7/site-packages/gwcs/wcs.py:127: in get_transform
    transforms = [tr.inverse for tr in transforms[::-1]]
../../../../miniconda/envs/tmp_env0/lib/python3.7/site-packages/gwcs/wcs.py:127: in <listcomp>
    transforms = [tr.inverse for tr in transforms[::-1]]

self = <CompoundModel(matrix_0=[[0.000018, 0.      ], [0.      , 0.000018]], translation_0=[0., 0.], offset_1=-0.03817084, of...9=[0., 0.], offset_10=0.0001439, offset_11=0.29360602, angles_13=[ 0.05879583,  0.01342577, -0.43673554, -4.1       ])>

    @property
    def inverse(self):
        if self.has_inverse():
            if self._user_inverse is not None:
                return self._user_inverse
            return self._inverse
        else:
>           raise NotImplementedError("Inverse function not provided")
E           NotImplementedError: Inverse function not provided

../../../../miniconda/envs/tmp_env0/lib/python3.7/site-packages/astropy/modeling/core.py:2956: NotImplementedError
```

This suggests that the PR that fixed this and is milestoned for 4.1

https://github.com/astropy/astropy/pull/10384

did not make it into the release.  Looking at the log of commits for the `v4.1rc` tag, this seems to be the case.

jdavies-st:
So @nden and @hbushouse , I think the way forward is to update `setup.py` `install_requires` to have `asdf>=2.6` on the same day that `astropy 4.1` is released.  Is there any drawback to that?

And that will only be for master.  Not sure if we'll have to cut a new release for `stable`.

hbushouse:
Is there any drawback to jwst functionality to not allowing `asdf < 2.6` anymore? If not, then your suggestion of setting `asdf >= 2.6` when astropy 4.1 comes out seems appropriate.

jdavies-st:
I don't think there's a drawback.  So that's probably the most reliable way forward.

jdavies-st:
Closing this as it has served its purpose in checking `astropy 4.1rc1`.

nden:
👍 on setting the minimum values. As for timing - should we wait for the gwcs release that sets those minimum versions as well? It should come shortly after the astropy release.

jdavies-st:
This package already works with `asdf 2.6`, so I think we can set that as a minimum version anytime and it will be OK.

Will the next release of `gwcs` require `astropy 4.1`?  If so, then yeah, we may want to bump both those minimum versions when `gwcs` comes out.

Issue #2824, opened 2018-11-01
stpipe.Step does respect --save_results=False


jdavies-st:
Data in `/user/jdavies/test_writing_bryan/`

```
(jwst_dev) rey> collect_pipeline_cfgs ./config
(jwst_dev) rey> strun config/flat_field.cfg jw00017001001_01101_00001_NRCA1_ramp_fit.fits 
2018-11-01 12:32:18,950 - stpipe.flat_field - INFO - FlatFieldStep instance created.
2018-11-01 12:32:18,975 - stpipe.flat_field - INFO - Step flat_field running with args ('jw00017001001_01101_00001_NRCA1_ramp_fit.fits',).
2018-11-01 12:32:34,401 - stpipe.flat_field - INFO - Saved model in jw00017001001_01101_00001_NRCA1_ramp_fit_flat_field.fits
2018-11-01 12:32:34,401 - stpipe.flat_field - INFO - Step flat_field done
(jwst_dev) rey> ls
config/								jw00017001001_01101_00001_NRCA1_ramp_fit_flat_field.fits
jw00017001001_01101_00001_NRCA1_ramp_fit.fits
(jwst_dev) rey> rm jw00017001001_01101_00001_NRCA1_ramp_fit_flat_field.fits 
(jwst_dev) rey> strun config/flat_field.cfg jw00017001001_01101_00001_NRCA1_ramp_fit.fits --save_results=False
2018-11-01 12:33:06,878 - stpipe.flat_field - INFO - FlatFieldStep instance created.
2018-11-01 12:33:06,898 - stpipe.flat_field - INFO - Step flat_field running with args ('jw00017001001_01101_00001_NRCA1_ramp_fit.fits',).
2018-11-01 12:33:11,136 - stpipe.flat_field - INFO - Saved model in jw00017001001_01101_00001_NRCA1_ramp_fit_flat_field.fits
2018-11-01 12:33:11,136 - stpipe.flat_field - INFO - Step flat_field done
(jwst_dev) rey> ls
config/								jw00017001001_01101_00001_NRCA1_ramp_fit_flat_field.fits
jw00017001001_01101_00001_NRCA1_ramp_fit.fits
```

I would think that writing out of files should be controlled here
https://github.com/spacetelescope/jwst/blob/39a2149e2e8a4e7c2c3925370c9f3632b00ebca7/jwst/stpipe/step.py#L442

But that doesn't seem to be the case because of
[...]

PR #4990, opened 2020-05-27
JP-872: Refactor Image3Pipeline to use Step infrastructure

### LINKED ISSUE #3829: Update calwebb_image3 pipeline to "modern" standards
Issue [JP-872](https://jira.stsci.edu/browse/JP-872) was created by Howard Bushouse:

As reported in JP-858, the calwebb_image3 pipeline does things very differently than the more recently developed level-3 pipelines like calwebb_spec3. For example, it explicitly saves some intermediate products, like the crf's from outlier_detection, rather than letting the step infrastructure handle it via the use of "save_results". It also has some log messages in the pipeline regarding the execution of individual steps, which should be moved into the steps themselves (or just eliminated).

jwhite3:
This PR refactors `Image3Pipeline` to use the `Step` infrastructure to save results of the included steps and removes extra step logging at the pipeline level (to be more consistent with other pipelines)

Resolves #3829 / [JP-872](https://jira.stsci.edu/browse/JP-872)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,11 @@ associations
 
 - Update diagrams to change sloper to detector1. [#4986]
 
+pipeline
+--------
+
+- Refactor the ``Image3Pipeline`` to use ``stpipe`` infrastructure. [#4990]
+
 datamodels
 ----------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,11 +6,6 @@ associations
 
 - Update diagrams to change sloper to detector1. [#4986]
 
-pipeline
---------
-
-- Refactor the ``Image3Pipeline`` to use ``stpipe`` infrastructure. [#4990]
-
 datamodels
 ----------
 
@@ -22,6 +17,11 @@ extract_1d
 ----------
 
 - Fix bug in creating a polynomial fit used in background extraction. [#4970]
+
+pipeline
+--------
+
+- Refactor the ``Image3Pipeline`` to use ``stpipe`` infrastructure. [#4990]
 
 stpipe
 ------
--- a/jwst/pipeline/calwebb_image3.py
+++ b/jwst/pipeline/calwebb_image3.py
@@ -32,13 +32,14 @@ class Image3Pipeline(Pipeline):
     """
 
     # Define alias to steps
-    step_defs = {'assign_mtwcs': assign_mtwcs_step.AssignMTWcsStep,
-                 'tweakreg': tweakreg_step.TweakRegStep,
-                 'skymatch': skymatch_step.SkyMatchStep,
-                 'outlier_detection': outlier_detection_step.OutlierDetectionStep,
-                 'resample': resample_step.ResampleStep,
-                 'source_catalog': source_catalog_step.SourceCatalogStep
-                 }
+    step_defs = {
+        'assign_mtwcs': assign_mtwcs_step.AssignMTWcsStep,
+        'tweakreg': tweakreg_step.TweakRegStep,
+        'skymatch': skymatch_step.SkyMatchStep,
+        'outlier_detection': outlier_detection_step.OutlierDetectionStep,
+        'resample': resample_step.ResampleStep,
+        'source_catalog': source_catalog_step.SourceCatalogStep
+    }
 
     def process(self, input_data):
         """
@@ -65,15 +66,16 @@ class Image3Pipeline(Pipeline):
 
         with datamodels.open(input_data, asn_exptypes=asn_exptypes) as input_models:
             # If input is an association, set the output to the product name.
-            try:
-                self.output_file = input_models.meta.asn_table.products[0].name
-            except AttributeError:
-                pass
+            if self.output_file is None:
+                try:
+                    self.output_file = input_models.meta.asn_table.products[0].name
+                except (AttributeError, IndexError):
+                    pass
 
             # Check if input is single or multiple exposures
             try:
                 has_groups = len(input_models.group_names) > 1
-            except Exception:
+            except (AttributeError, TypeError, KeyError):
                 has_groups = False
 
             if isinstance(input_models, datamodels.ModelContainer) and has_groups:
@@ -86,17 +88,4 @@ class Image3Pipeline(Pipeline):
                 input_models = self.outlier_detection(input_models)
 
             result = self.resample(input_models)
-
-            try:
-                result.meta.asn.pool_name = input_models.meta.asn_table.asn_pool
-                result.meta.asn.table_name = os.path.basename(input_data)
-                result.meta.filename = input_models.meta.asn_table.products[0].name
-            except Exception:
-                pass
-
             self.source_catalog(result)
-            # NOTE: source_catalog step writes out the catalog in .ecsv format
-            # In the future it would be nice if it was returned to the pipeline,
-            # and then written here.  A datamodel for .ecsv might be required.
-
-        self.log.info('Ending calwebb_image3')
--- a/jwst/pipeline/calwebb_image3.py
+++ b/jwst/pipeline/calwebb_image3.py
@@ -29,7 +29,6 @@ class Image3Pipeline(Pipeline):
     """
 
     spec = """
-        suffix = string(default='i2d')
     """
 
     # Define alias to steps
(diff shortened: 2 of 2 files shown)


jwst/pipeline/calwebb_image3.py
```diff
@@ -41,77 +40,63 @@ class Image3Pipeline(Pipeline):
...
-
-        self.log.info("Resampling images to final result...")
-        result = self.resample(input_models)
-
-        try:
-            result.meta.asn.pool_name = input_models.meta.asn_table.asn_pool
-            result.meta.asn.table_name = os.path.basename(input)
-            result.meta.filename = input_models.meta.asn_table.products[0].name
-        except Exception:
-            pass
```
jwhite3, line 106: I noticed that some of the other pipelines (for example `Spec3Pipeline`) don't include this section, so I wasn't sure if it was necessary or not.

jwst/pipeline/calwebb_image3.py
```diff
@@ -41,77 +40,63 @@ class Image3Pipeline(Pipeline):
...
+            # If input is an association, set the output to the product name.
+            try:
+                self.output_file = input_models.meta.asn_table.products[0].name
+            except AttributeError:
+                pass
+
+            # Check if input is single or multiple exposures
+            try:
+                has_groups = len(input_models.group_names) > 1
+            except Exception:
```
jwhite3, line 76: This needs a more specific exception(s). My guess would be to check for both an `AttributeError` as well as `TypeError`, but I wasn't sure. 

Same for lines 68-71

jwst/pipeline/calwebb_image3.py
```diff
@@ -41,77 +40,63 @@ class Image3Pipeline(Pipeline):
...
+                result.meta.filename = input_models.meta.asn_table.products[0].name
+            except Exception:
+                pass
+
+            self.source_catalog(result)
+            # NOTE: source_catalog step writes out the catalog in .ecsv format
+            # In the future it would be nice if it was returned to the pipeline,
+            # and then written here.  A datamodel for .ecsv might be required.
+
+        self.log.info('Ending calwebb_image3')
```
jwhite3, line 102: I added this to be consistent with some of the other pipelines, but It's kind of redundant since `Pipeline` includes a "done message"

jwst/pipeline/calwebb_image3.py
```diff
@@ -41,77 +40,63 @@ class Image3Pipeline(Pipeline):
...
-            result.meta.filename = input_models.meta.asn_table.products[0].name
-        except Exception:
-            pass
-
-        self.save_model(result, suffix=self.suffix)
-
-        self.log.info("Creating source catalog...")
+
+        # Configure settings for saving results files
+        self.outlier_detection.suffix = 'crf'
```
jdavies-st, line 58: Is this going to get the `asn_id` correct?  Notice the filenames in the `_crf.fits` checks in the regression tests for this pipeline.  The `asn_id` is the "a3001" and similar strings.

jwst/pipeline/calwebb_image3.py
jdavies-st: Agree.  You may want to see what the `ModelContainer.group_names` property returns if there are none.

jwst/pipeline/calwebb_image3.py
jwhite3: I made sure to run the miri and nircam image regression tests, and from those runs, nothing seemed amiss

jwst/pipeline/calwebb_image3.py
```diff
@@ -41,77 +40,63 @@ class Image3Pipeline(Pipeline):
...
+
+            try:
+                result.meta.asn.pool_name = input_models.meta.asn_table.asn_pool
+                result.meta.asn.table_name = os.path.basename(input_data)
+                result.meta.filename = input_models.meta.asn_table.products[0].name
+            except Exception:
+                pass
+
+            self.source_catalog(result)
+            # NOTE: source_catalog step writes out the catalog in .ecsv format
```
stscieisenhamer, line 98: This comment seemed to be there originally, but is nonsensical. This is operating by design. IMHO this comment should be removed.

jwst/pipeline/calwebb_image3.py
stscieisenhamer: IMHO agree that it is redundant and this logging should be removed. Probably should be removed from all other pipelines also, but that is another PR....

jwst/pipeline/calwebb_image3.py
hbushouse: I think at least some of that might be included down in the `resample` step itself when the output data model (`result`) is created and has `.update()` called to copy meta from one of the inputs. I guess the regtests will tell us for sure (i.e. we'll see if ASNPOOL and ASNTABLE are now missing from the outputs). And I'm pretty sure that `result.meta.filename` is set down within the `datamodels` code itself when any file gets saved to disk.

jwst/pipeline/calwebb_image3.py
```diff
@@ -41,77 +40,63 @@ class Image3Pipeline(Pipeline):
...
-        except Exception:
-            pass
-
-        self.save_model(result, suffix=self.suffix)
-
-        self.log.info("Creating source catalog...")
+
+        # Configure settings for saving results files
+        self.outlier_detection.suffix = 'crf'
+        self.outlier_detection.save_results = self.save_results
```
hbushouse, line 59: This looks like it's assigning the value of `save_results` that applies to the whole pipeline to the individual steps. What happens if a user specifies an override, e.g. `--steps.outlier_detection.save_results=False`? Is the user's value going to be overridden by the pipeline setting?

jwst/pipeline/calwebb_image3.py
hbushouse: I think this comment was written based on 1) the assumption that the source catalog product was the final, main product of this pipeline, and hence the creation of the file should be handled by the pipeline, and 2) the pipeline was not able to save that file, because there's no corresponding data model for it (hence it must be saved by the step itself).

But in reality it's the resampled (i2d) product that is regarded as the final, main product and that's the one that's saved by the pipeline (or by stpipe when the pipeline finishes). Hence you're correct that this comment really doesn't apply and can be removed.

jwst/pipeline/calwebb_image3.py
jwhite3: I think it would, yes, and is probably not desired now that I think of it. However, this is the same practice used in the `Spec3Pipeline` too. 

To fix this then, I suppose that I would add the suffix back into the spec string, and remove those explicit assignments.

jwhite3: Then, I guess that we'd also need to fix that in `Spec3Pipeline` too

hbushouse: Someone could/should run a quick test of seeing if the command-line overrides actually work.

jwhite3: I think that the intent here was that the pipeline should produce these output files by default, but now that I'm thinking about it, I'm not actually sure how you'd control that at the pipeline level while still allowing for overrides for individual steps. 

If you removed those assignments in `process`, the only way to set it (I think.. isn't the default for `save_results` for the steps `False`?) would be to configure it after instantiation or with a config file, right?

Another approach that I see in other pipelines is to add additional arguments in `spec` for use in switching saving results on/off

hbushouse: The default setting for `save_results` in the `stpipe` `step` class is `False`. And this same operation of setting the value for individual steps, like `resample`, to the step value is also used in some other pipelines, like `calwebb_image2`. So now I don't understand how those pipeline are actually managing to create output products from these individual steps when the default seems to be `False`.  @stscieisenhamer any wisdom here?

jdavies-st: `save_results` defaults to `False` for all steps/pipelines.  When you use `strun` on the command line, it sets `save_results` for the pipeline (or whatever strun is running) to `True`.  So I think the above is valid.  It allows for nothing to be written out if being run from within python, and for the normal things to be written out when running using `strun`.

That said, if you want the user to be able to specify turning off writing out the `outlier_detection` results, then you could check first if the user has set it?  But that's tricky because it's an order-of-operations issue.  Hmm.

jwhite3: What if we added an argument in `spec`, something like: 
```python
spec = """
    default_save_results = bool(default=True)
"""
```

Combined with: 

```python
if self.default_save_results:
    # Configure settings for saving results files
    self.outlier_detection.suffix = 'crf'
    self.outlier_detection.save_results = self.save_results

    self.resample.suffix = 'i2d'
    self.resample.save_results = self.save_results

    self.source_catalog.save_results = self.save_results
```

That way, users could have the same "default results" output just like when executing from the command line, but they could also disable that and set things manually or with a config file. Seems like that might cover both cases, while by default providing the output products that the stakeholders seem to expect

hbushouse: > `save_results` defaults to `False` for all steps/pipelines. When you use `strun` on the command line, it sets `save_results` for the pipeline (or whatever strun is running) to `True`. So I think the above is valid. It allows for nothing to be written out if being run from within python, and for the normal things to be written out when running using `strun`.
> 
> That said, if you want the user to be able to specify turning off writing out the `outlier_detection` results, then you could check first if the user has set it? But that's tricky because it's an order-of-operations issue. Hmm.

Ah, that's right. The default is `False` for the purpose of running interactively within python, where you normally just want the results passed back in a data model and not saved to disk. When running from command line it's automatically reset to `True`, otherwise there's no way to get the results.

The default of `True` is necessary for all the steps that create required products, which includes `resample`, `outlier_detection`, `source_catalog`, `extract_1d`, etc. (the Ops environment always wants those results). I suppose a user might want to be able to turn off some of those occasionally, but it's more import to allow them to simply turn on the saving of results for intermediate steps that don't normally save their products.

jwhite3: Just to provide resolution to this thread: 

This is actually a known issue ( #2824 #2723 ), and is outside of the scope of this PR, and will be addressed later.

jwst/pipeline/calwebb_image3.py
jwhite3: Removed this section and it resulted in no changes to ASNPOOL or ASNTABLE in regression tests

jdavies-st:
Just fix this errant import

https://travis-ci.org/github/spacetelescope/jwst/jobs/692309787#L215

jwst/pipeline/calwebb_image3.py
```diff
@@ -29,89 +29,63 @@ class Image3Pipeline(Pipeline):
...
-    def process(self, input):
+    step_defs = {
+        'assign_mtwcs': assign_mtwcs_step.AssignMTWcsStep,
+        'tweakreg': tweakreg_step.TweakRegStep,
+        'skymatch': skymatch_step.SkyMatchStep,
+        'outlier_detection': outlier_detection_step.OutlierDetectionStep,
+        'resample': resample_step.ResampleStep,
+        'source_catalog': source_catalog_step.SourceCatalogStep
+    }
+
```
hbushouse, line 43: Just for my own edification, what's gained or changed by moving `step_defs` out of the `process` method?

jwhite3: I'm not sure what you mean. `step_defs` is a class attribute, and was never part of the `process` method. I think this is actually a part of the `Pipeline` infrastructure. 

As I was going through, I just changed some indentation is all

hbushouse: Ah, OK. Perhaps it was just the layout of the diff that made me think it used to be inside of `process`, but it really wasn't.

CHANGES.rst
```diff
@@ -6,6 +6,11 @@ associations
 
 - Update diagrams to change sloper to detector1. [#4986]
 
+pipeline
```
hbushouse, line 9: Minor nit, but `pipeline` entries should be further down in the log to maintain alphabetical order (i.e. after `datamodels`)

jwhite3: D'oh!

Issue #4988, opened 2020-05-27
Add note to installation instructions that Windows compatibility is not supported 


stscijgbot-jp:
_Issue [JP-1474](https://jira.stsci.edu/browse/JP-1474) was created on JIRA by [Alicia Canipe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=acanipe):_

We should add a note to the installation instructions (and mabye ReadTheDocs/JDox?) that, while the software should install and run for Windows, it is not currently supported. This question came up in the DMSWG meeting today.

stscijgbot-jp:
_Comment by [Alicia Canipe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=acanipe) on [JIRA](https://jira.stsci.edu/browse/JP-1474?focusedCommentId=465909#comment-465909):_

[Steven Crawford](https://jira.stsci.edu/secure/ViewProfile.jspa?name=scrawford) FYI

stscijgbot-jp:
_Comment by [Steven Crawford](https://jira.stsci.edu/secure/ViewProfile.jspa?name=scrawford) on [JIRA](https://jira.stsci.edu/browse/JP-1474?focusedCommentId=465912#comment-465912):_

Thanks Alicia - I'm adding [Mark Kyprianou](https://jira.stsci.edu/secure/ViewProfile.jspa?name=kyp).  This is our current position and you are right we should add a note.   

We could try to consider adding in Windows builds on conda as we have had requested from other users as well for this.   However, there is no guarantee that all the dependencies and everything will work and that will require more testing and support.

jdavies-st:
We don't know if it runs on Windows because we don't test it.  Maybe a good first step would be to add a job to our Travis CI matrix that runs our unit tests on a Windows box and see what breaks.  And then fix it.  Then I think we can say something more definitive in our documentation.

jdavies-st:
Based on the results of #4993, there are 163 unit test failures in `jwst` when run on Windows.

https://travis-ci.org/github/spacetelescope/jwst/jobs/692239993#L76133-L76304

A number of these come via `crds`, which also does not support Windows.

```
>               raise CrdsDownloadError(
                    "Failed caching mapping files:", str(exc)) from exc
[...]

PR #4993, opened 2020-05-28
Add Windows job to TravisCI as allowed_failure

jdavies-st:
A start at #4988


jdavies-st:
Closing this PR.  It can be resurrected again when `crds` supports Windows and we decide that `jwst` will do so as well.

PR #4998, opened 2020-05-28
JP-1410 Add blend rule for detector

### LINKED ISSUE #4681: Incorrect meta.instrument.detector in combined i2d products
The `resample` step can combine multiple detectors (this will be typical for NIRCam), thus the `meta.instrument.detector` value should be set to `None` (according to @jdavies-st).

The `jw42424-o002_t001_nircam_clear-f444w_i2d.fits` file produced by the `test_nircam_image.py` regression test is the combination of both NIRCam LW detectors.  However, its `meta.instrument.detector` is incorrectly set to `NRCALONG` instead of `None`.

stscieisenhamer:
Resolves [JP-1410](https://jira.stsci.edu/browse/JP-1410)
Resolves #4681

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -386,6 +386,7 @@ properties:
             type: string
             fits_keyword: MODULE
             blend_table: True
+            blend_rule: multi
           channel:
             title: Instrument channel
             type: string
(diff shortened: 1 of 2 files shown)


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -377,9 +377,10 @@ properties:
             type: string
             enum: [NRCA1, NRCA2, NRCA3, NRCA4, NRCALONG, NRCB1, NRCB2, NRCB3, NRCB4,
               NRCBLONG, NRS1, NRS2, ANY, MIRIMAGE, MIRIFULONG, MIRIFUSHORT,
-              NIS, GUIDER1, GUIDER2, N/A]
+              NIS, GUIDER1, GUIDER2, N/A, MULTIPLE]
             fits_keyword: DETECTOR
             blend_table: True
+            blend_rule: multi
```
hbushouse, line 383: While we're at it, we should make the same addition to the `module` keyword, because data from multiple modules can be combined into a single image.

PR #5007, opened 2020-06-01
JP-1472 Blend headers just from each input science model, not every integration

### LINKED ISSUE #4985: HDRTAB should contain only a single entry for each exposure
_Issue [JP-1472](https://jira.stsci.edu/browse/JP-1472) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

When CubeModel data are used as input to steps like resample, which run blendheaders to create a HDRTAB extension, they are typically reorganized into a ModelContainer where each ImageModel in the container is the image from a single plane (integration) in the original CubeModel, so that the resample step itself doesn't have to reorganize the data before combining. This has an undesirable side effect in that the blendheaders routine then treats each of those images contained in the ModelContainer as is it were from a separate exposure and hence creates multiple rows in the HDRTAB table that contain identical data (because they're all from the same exposure).

An example of this behavior is when running the calwebb_coron3 pipeline on exposures that have NINTS>1, resulting in NINTS rows for each original exposure in the HDRTAB extension of the i2d product.

Updates should be made to only create a single row in the HDRTAB for a given filename value of the inputs.

stscieisenhamer:
Resolves #4985 
Resolve [JP-1472](https://jira.stsci.edu/browse/JP-1472)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,11 +6,6 @@ associations
 
 - Update diagrams to change sloper to detector1. [#4986]
 
-coron3
-------
-
-- Blend headers just from each input science model, not every integration. [#5007]
-
 datamodels
 ----------
 
@@ -29,6 +24,8 @@ pipeline
 --------
 
 - Refactor the ``Image3Pipeline`` to use ``stpipe`` infrastructure. [#4990]
+
+- Blend headers just from each input science model, not every integration. [#5007]
 
 stpipe
 ------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,7 +25,7 @@ pipeline
 
 - Refactor the ``Image3Pipeline`` to use ``stpipe`` infrastructure. [#4990]
 
-- Blend headers just from each input science model, not every integration. [#5007]
+- Fix ``Coron3Pipeline`` to blend headers just from each input science model, not every integration. [#5007]
 
 stpipe
 ------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,11 @@ associations
 
 - Update diagrams to change sloper to detector1. [#4986]
 
+coron3
+------
+
+- Blend headers just from each input science model, not every integration. [#5007]
+
 datamodels
 ----------
 
--- a/jwst/pipeline/calwebb_coron3.cfg
+++ b/jwst/pipeline/calwebb_coron3.cfg
@@ -7,3 +7,4 @@ class = "jwst.pipeline.Coron3Pipeline"
       [[klip]]
       [[outlier_detection]]
       [[resample]]
+        blendheaders = false
--- a/jwst/pipeline/calwebb_coron3.cfg
+++ b/jwst/pipeline/calwebb_coron3.cfg
@@ -7,4 +7,3 @@ class = "jwst.pipeline.Coron3Pipeline"
       [[klip]]
       [[outlier_detection]]
       [[resample]]
-        blendheaders = false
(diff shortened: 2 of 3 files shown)


jwst/pipeline/calwebb_coron3.cfg
```diff
@@ -7,3 +7,4 @@ class = "jwst.pipeline.Coron3Pipeline"
       [[klip]]
       [[outlier_detection]]
       [[resample]]
+        blendheaders = false
```
jdavies-st, line 10: Should this also be set in `calwebb_coron3.py` `Coron3Pipeline .spec` class attribute?

CHANGES.rst
```diff
@@ -6,6 +6,11 @@ associations
 
 - Update diagrams to change sloper to detector1. [#4986]
 
+coron3
```
hbushouse, line 9: Since this is a pipeline module change, this should go under a "pipeline" heading and then referenced as "calwebb_coron3".

jwst/pipeline/calwebb_coron3.cfg
hbushouse: I would say yes, otherwise if someone runs via the `Coron3Pipeline` class, without the config file, it won't work right.

stscieisenhamer: As far as I know, hierarchical definitions in `spec` cannot be done. `spec` defines parameters for the specific class; it cannot affect other things.

This, along with the nirspec lamp processing and possible others, REQUIRE some sort of config file setup, at the moment, to operate correctly. This is due to both the above situation with `spec`, and the fact that CRDS-retrieved parameter files have not been implemented. Also, the design discussion of "are pipelines in full control" vs "pipelines a thin veneer over" their substeps, need some resolution. Hearken back to the very recent discussion of `save_results`.

This current implementation is "baseline" with what has been implemented. Options are:

1. Leave as is
2. Hardwire `blendheaders` into the code
3. Go ahead and create/ingest the necessary CRDS step parameter file for coron3.

My initial impulse was option 2.

Note: If set wrong, science itself is not muddled. The results keywords will be correct. Just the HDRTAB would be unnecessarily repeated, with the wrong filenames.

Discuss!

hbushouse: So if it doesn't work to set `blendheaders = False` for the `resample` step in the `spec` definition, can it at least be set somewhere in the `Coron3Pipeline` process method itself, like we do with `save_results` settings for some steps in pipeline modules?

jdavies-st: If we are going with option 2, then setting it by hand, outside of `spec`, is probably the consistent (if not elegant) solution.  Somewhere at the top of process:

```
self.resample.blendheaders = False
```

A user won't be able to modify it, but we don't actually want them to be able to.

jdavies-st: And then you can remove it from the `.cfg` file, as it will be redundant.

stscieisenhamer: Consensus is to hardwire. Done.

stscieisenhamer:
Concerning tso3 processing: For both imaging and spectrographic, the whitelight ecsv file has a very minimal header. If the intent was to reproduce a more full-featured header here, then work is to be done.

Similarly with the spectrographic x1d result. The header is only taken from the first input exposure. Again, if a more "true", blended header is desired. Work is to be done.

If either/both cases above should be fixed, a separate issue/PR would probably be more appropriate.

stscieisenhamer:
Issue JP-1486 created which focuses on the tso3 issues.

CHANGES.rst
```diff
@@ -25,6 +25,8 @@ pipeline
 
 - Refactor the ``Image3Pipeline`` to use ``stpipe`` infrastructure. [#4990]
 
+- Blend headers just from each input science model, not every integration. [#5007]
```
hbushouse, line 28: Should mention that this change was applied in the `Coron3Pipeline` (or `calwebb_coron3`)

PR #5015, opened 2020-06-03
JP-1490: skip assign_mtwcs if target coords missing

### LINKED ISSUE #5014: assign_mtwcs fails when MT_RA/DEC keywords are missing
_Issue [JP-1490](https://jira.stsci.edu/browse/JP-1490) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

Datasets in the mt1_OTB test suite fail in the assign_mtwcs step with the following traceback:
```java
----> 1 mt_avra=mt_ra.mean()


~/anaconda3/envs/jwstdev/lib/python3.8/site-packages/numpy/core/_methods.py in _mean(a, axis, dtype, out, keepdims)
    149             is_float16_result = True
    150 
--> 151     ret = umr_sum(arr, axis, dtype, out, keepdims)
    152     if isinstance(ret, mu.ndarray):
    153         ret = um.true_divide(


TypeError: unsupported operand type(s) for +: 'NoneType' and 'NoneType'
```
when attempting to compute the average of the MT_RA values in the collection of input images, due to the fact that SDP is not yet populating the MT_RA/DEC keywords during level-1b processing.

Need to make this calculation more robust against this kind of failure.

hbushouse:
If any of the input images to the `assign_mtwcs` step are missing the MT_RA or MT_DEC keyword values, issue a warning and skip the step.

Fixes #5014 / [JP-1490](https://jira.stsci.edu/browse/JP-1490)

### DIFF
--- a/jwst/assign_mtwcs/assign_mtwcs_step.py
+++ b/jwst/assign_mtwcs/assign_mtwcs_step.py
@@ -40,10 +40,4 @@ class AssignMTWcsStep(Step):
         # Apply the step
         result = assign_moving_target_wcs(input)
 
-        # Check for a null result
-        if result is None:
-            for model in input:
-                model.meta.cal_step.assign_mtwcs = 'SKIPPED'
-            return input
-
         return result
--- a/jwst/assign_mtwcs/assign_mtwcs_step.py
+++ b/jwst/assign_mtwcs/assign_mtwcs_step.py
@@ -1,9 +1,8 @@
 #! /usr/bin/env python
+import logging
 from ..stpipe import Step
 from .. import datamodels
-import logging
 from .moving_target_wcs import assign_moving_target_wcs
-
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


jwst/assign_mtwcs/assign_mtwcs_step.py
```diff
@@ -22,17 +21,29 @@ class AssignMTWcsStep(Step):
...
             return input
 
+        # Apply the step
         result = assign_moving_target_wcs(input)
+
+        # Check for a null result
+        if result is None:
+            for model in input:
+                model.meta.cal_step.assign_mtwcs = 'SKIPPED'
+            return input
```
nden, line 47: This is a side comment, if you want to leave it as is that's fine. And I wrote this initially so it's really on me. 

The code would be a little bit more readable and symmetric if `assign_moving_target_wcs` returned the input model and set the status instead of returning None.

Issue #5022, opened 2020-06-08
Add default to datamodels 'lamp_mode' entry


stscijgbot-jp:
_Issue [JP-1496](https://jira.stsci.edu/browse/JP-1496) was created on JIRA by [Robert Jedrzejewski](https://jira.stsci.edu/secure/ViewProfile.jspa?name=rij):_

Since the datamodel parameter meta.instrument.lamp_mode will be used as a parkey for reference files, and since some of our test data doesn't have the OPMODE keyword, set the default value of meta.instrument.lamp_mode to 'NONE' so that CRDS can do a lookup even if the OPMODE keyword isn't present in the FITS header.  Also add enum entries for this attribute.

hbushouse:
Fixed by #5024

stscijgbot-jp:
_Comment by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse) on [JIRA](https://jira.stsci.edu/browse/JP-1496?focusedCommentId=468927#comment-468927):_

Fixed by #5024

PR #5024, opened 2020-06-09
Changes to datamodels to add default value for meta.instrument.lamp_mode

stscirij:
As it will be used as a parkey in CRDS ref file selection and we want there to be a default value even if the OPMODE keyword is missing from the science data headers.  Also added enum list.


stscirij:
Fixes #5022/[JP-1496](https://jira.stsci.edu/browse/JP-1496)

PR #5034, opened 2020-06-11
Fix a bug in ramp_fit multiprocessing code

hbushouse:
PR #4815 was a missing a line to populate the ERR array in the output rateints product for the case of single processor mode. Also added missing change log entry for the addition of multiprocessing to the step.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -172,6 +172,8 @@ def ols_ramp_fit_multi(input_model, buffsize, save_opt, readnoise_2d, gain_2d,
            Object containing optional GLS-specific ramp fitting data for the
            exposure
        """
+
+    # Determine number of slices to use for multi-processor computations
     if max_cores == 'none':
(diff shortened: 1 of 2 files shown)


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -205,15 +212,20 @@ def ols_ramp_fit_multi(input_model, buffsize, save_opt, readnoise_2d, gain_2d,
...
+                     input_model.meta.exposure.groupgap, input_model.meta.exposure.nframes,
+                     input_model.meta.exposure.drop_frames1, int_times)
+
+        # Populate the rateints output model
         if number_of_integrations > 1:
             int_model.data = int_data
             int_model.dq = int_dq
             int_model.var_poisson = int_var_poisson
             int_model.var_rnoise = int_var_rnoise
+            int_model.err = int_err
```
hbushouse, line 226: Reviewers: this is the only real code change. All others are purely cosmetic for readability.

PR #5036, opened 2020-06-11
filteroffset correction - fix #2

### LINKED ISSUE #2: Move src/tweakreg to tweakreg subpackage?
Makes more sense than sitting in the top level.

nden:
Apply filteroffset correction only if the correction is present in the reference file.


nden:
@hbushouse Regression tests pass now.Should have run them before.

PR #5040, opened 2020-06-15
Temporarily xfail regtest that has rmap problem

hbushouse:
Temporarily xfail the NIRSpec `calwebb_detector1` regtest that throws a bestref error due to an rmap problem, because it appears it's going to be a while yet till the rmap gets fixed. At least we'll get a passing regtest run this way.

### DIFF
--- a/jwst/regtest/test_nirspec_subarray.py
+++ b/jwst/regtest/test_nirspec_subarray.py
@@ -6,6 +6,8 @@ from astropy.io.fits.diff import FITSDiff
 from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
 from jwst.stpipe import Step
 
+from crds.exceptions import CrdsLookupError
+
 """
 nrs1_group_subarray.fits                the input (uncal) file
 nrs1_group_subarray_group_scale.fits    output from group_scale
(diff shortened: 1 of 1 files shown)


hbushouse:
> Can you add a `strict=True` to the xfail args, so that when this rmap problem is fixed and it passes, we will get an error in the test that xpassed? And we can then un-xfail it. And perhaps a `raises=CrdsLookupError` so that if it fails for a different reason due to the fixed rmap, we will also know?
> 
> You'll have to add a
> 
> ```
> from crds.exceptions import CrdsLookupError
> ```
> 
> at the top if you use the `raises` kwarg.
> 
> This should make it easier to get back to working order when the rmap issue is fixed.

Like this? (see latest commit)

hbushouse:
Jenkins tests seem to be hanging during unit tests for `ramp_fitting`, even though the same tests run just fine in Travis and on my own system. Another active PR is showing the same thing (Jenkins hangs, while Travis runs fine). Any thoughts from anyone?

jdavies-st:
Yeah, that way if it does pass because the rmap gets fixed and doesn't fail with a `CrdsLookupError`, then we'll get

```
jwst/regtest/test_nirspec_subarray.py::test_nirspec_detector1[group_scale] FAILED       [100%]

=================================== FAILURES ====================================
_____________________ test_nirspec_detector1[group_scale] _______________________
[XPASS(strict)] known CRDS rmap issue
```

PR #5045, opened 2020-06-16
Don't leave behind FITS files in ramp_fitting multiprocessing tests

jdavies-st:
These 2 tests leave behind FITS files in the repo.  They should not.


jdavies-st:
Well, for some reason TravisCI checks are not being linked properly here.  Here's the test run.

https://travis-ci.org/github/spacetelescope/jwst/builds/699070434

PR #5046, opened 2020-06-16
Disable ssbjenkins builds

jdavies-st:
This will disable the per-PR "Kurt Russell" ssbjenkins builds until we sort out the `multiprocessing` core/threading issue.

Apparently the way to do this is to rename our `Jenkinsfile` to something else.


jdavies-st:
The TravisCI build which doesn't seem to be linking properly from the interface.

https://travis-ci.org/github/spacetelescope/jwst/builds/699074201

PR #5047, opened 2020-06-16
Fix NIRSPEC IFU slices problem in JP-932

### LINKED ISSUE #3910: IFU not mapping trace to sky for all wavelengths
Issue [JP-932](https://jira.stsci.edu/browse/JP-932) was created by Nora Luetzgendorf:

When building the cube for several NIRSpec modes, the first ~20 wavelength bins of some slices (not all) do not seem to have data mapped to the sky. It seems like this is a problem that arises when assigning pixels positions on the sky and is visible in the traces of the level 2b products. Example data can be found at /grp/jwst/ssb/chartreuse/nirspec_ifudata/CV3

 

The attachment shows the problem on a cube.

jemorrison:
This PR fixes the NIRSPEC slices not mapping properly.
A fix was required in pathloss and photom to remove wavelengths with NANs

Resolves #3910 / [JP-932](https://jira.stsci.edu/browse/JP-932)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,6 +47,12 @@ master_background
 
 - Fix open files bug [#4995]
 
+pathloss
+--------
+- fix bug in NIRSpec IFU data that causes valid pixel dq flags to set to 
+  NON-SCIENCE in the region of an overlapping bounding box slice [#5047]
+  
+
 pipeline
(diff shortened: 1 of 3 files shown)


jemorrison:
Yes it is ready to merge

PR #5079, opened 2020-06-18
extract_1d style; readability updates

jwhite3:
This PR provides style/readability updates to `extract.py`. Some minor refactoring was included.


jwst/extract_1d/extract.py
```diff
@@ -197,39 +197,41 @@ def get_extract_parameters(ref_dict,
...
+
         for aper in ref_dict['apertures']:
-            if 'id' in aper and aper['id'] != "dummy" and \
-               (aper['id'] == slitname or aper['id'] == "ANY" or
-                slitname == "ANY"):
+            if (
+                    'id' in aper
+                    and aper['id'] != "dummy"
+                    and (aper['id'] == slitname or aper['id'] == "ANY" or slitname == "ANY")
+            ):
```
nden, line 234: Just curious - was this block produced by an automatic code formatter?

jwhite3: Nope! 
It's definitely a style that's similar to some common code formatters though (I think [Black](https://pypi.org/project/black/) does something similar), but for most of this I was doing it "by hand" with the help of an IDE (which does help with automatically adjusting indentation and such to match some standard) so that I could also work through the logic as I went

jwst/extract_1d/extract.py
```diff
@@ -1793,42 +1655,49 @@ def assign_polynomial_limits(self, verbose):
...
 
-    def extract(self, data, wl_array, verbose):
+    def extract(
+            self,
+            data: np.ndarray,
+            wl_array: Union[np.ndarray, None],
+            verbose: bool
+    ) -> Tuple[
+        float, float, np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray
+    ]:
```
jwhite3, line 1700: In cases where you want to annotate a return that's super long (like this) it is possible to define your own type beforehand: 

```python
EXRACT_RESULTS = Tuple[
          float, float, np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray
      ]
def extract(
    self, data: np.ndarray, wl_array: Union[np.ndarray, None], verbose: bool
)
```
This can be useful especially if several functions/methods return a similarly complex type. 

However, I personally would use this sparingly as it might add an unnecessary layer of obscurity.

jwst/extract_1d/extract.py
```diff
@@ -22,68 +26,67 @@
...
 
+
 class InvalidSpectralOrderNumberError(Extract1dError):
     """The spectral order number was invalid or off the detector."""
     pass
 
 
-def load_ref_file(refname):
+def load_ref_file(refname: str) -> dict:
     """Open the reference file.
```
hbushouse, line 90: Is this used for both extract1d and apcorr ref files?

jwst/extract_1d/extract.py
```diff
@@ -239,17 +248,19 @@ def get_extract_parameters(ref_dict,
...
                     continue
+
                 spectral_order = aper.get("spectral_order", sp_order)
+
                 if spectral_order == sp_order or spectral_order == ANY:
                     extract_params['match'] = EXACT
                     extract_params['spectral_order'] = sp_order
-                    # Note that extract_params['dispaxis'] is not assigned.
-                    # This will be done later, possibly slit by slit.
+                    # Note: extract_params['dispaxis'] is not assigned. This is be done later, possibly slit by slit.
```
hbushouse, line 260: Fix the grammar in "This is be done"

jwst/extract_1d/extract.py
```diff
@@ -539,31 +539,34 @@ def update_from_shape(ap, im_shape):
...
 
 
-def aperture_from_wcs(wcs, verbose):
+def aperture_from_wcs(wcs: WCS, verbose: bool) -> Union[NamedTuple, None]:
     """Get the limits over which the WCS is defined.
 
     Parameters
     ----------
-    wcs : data model
+    wcs : data model ?
```
hbushouse, line 569: I'm guessing it's not a data model, but a WCS object.

jwst/extract_1d/extract.py
```diff
@@ -2733,118 +2568,127 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
...
+
             source_type = slit.source_type
+
             if source_type != 'POINT':
                 apply_nod_offset = False
-                log.warning("SRCTYPE = '%s'; correcting for nod/dither "
-                            "offset will only be done for a point source, "
-                            "so apply_nod_offset will be set to False",
-                            source_type)
+                log.warning(
```
hbushouse, line 2605: I don't think this needs to be a warning. Working on extended sources is perfectly natural and hence there's nothing bad about it. So change this to `log.info` instead.

jwst/extract_1d/extract.py
```diff
@@ -2861,87 +2705,99 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
...
-            pixel_solid_angle = 1.                      # not needed
+            pixel_solid_angle = 1.  # not needed
 
         source_type = input_model.meta.target.source_type
-        input_units_are_megajanskys = (photom_has_been_run and
-                                       source_type == 'POINT' and
-                                       (instrument == 'NIRSPEC' or
-                                        exp_type == 'NIS_SOSS'))
+
+        input_units_are_megajanskys = (
```
hbushouse, line 2736: This logic seems dangerous. If at some point it's decided to change the calibrated units for one of these modes, someone has to make sure to update this code. Wouldn't it be better to just rely on the BUNIT keyword value? Of course if we're only making style updates here, I guess that's not an appropriate change.

jwst/extract_1d/extract.py
```diff
@@ -2861,87 +2705,99 @@ def do_extract1d(input_model, ref_dict, smoothing_length=None,
...
+                and (instrument == 'NIRSPEC' or exp_type == 'NIS_SOSS')
+        )
+
         if source_type != 'POINT':
             apply_nod_offset = False
-            log.warning("SRCTYPE = '%s'; correcting for nod/dither "
-                        "offset will only be done for a point source, "
-                        "so apply_nod_offset will be set to False",
-                        source_type)
+            log.warning(
```
hbushouse, line 2744: Again, change to `log.info`

jwhite3:
@hbushouse I noticed that a few of your comments seemed to be in light of the impending APCORR implementation. I'll be sure to address those comment and log changes in that PR

jwst/extract_1d/extract.py
jwhite3: I was confused on what the object was supposed to be since it says `wcs` but then describes the parameter as a "data model," so I was hoping that somebody might have a bit more contextual info here, but I think I agree with your guess!

jwst/extract_1d/extract.py
```diff
@@ -22,68 +26,67 @@
...
 Extended summary
 ----------------
-For full-frame input data, keyword SLTNAME may not be populated, so the
-slit name will be set to this string to indicate that the first slit in
-the reference file should be used.
-A slit name in the reference file can also be ANY, in which case the
-reference information for that slit will be regarded as matching any slit
-name from the input data.
+For full-frame input data, keyword SLTNAME may not be populated, so the slit name will be set to this string to indicate 
+that the first slit in the reference file should be used.
```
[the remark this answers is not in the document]
jwhite3: Yes, I think so

jwst/extract_1d/extract.py
jwhite3: No, it's not. In #4902 this function is renamed to specifically say that it's opening the extract1d file

jwst/extract_1d/extract.py
jwhite3: I agree, there's quite a lot of refactoring that needs to be done, this included. But, yes, I think that that should go in a separate "refactor extract 1d" PR

jwst/extract_1d/extract.py
```diff
@@ -774,16 +802,15 @@ def compare_stop(stop_ref, stop_wcs):
     value : int or float
         The stop limit, possibly constrained by the WCS stop limit.
     """
-
-    if stop_ref <= stop_wcs:            # ref is inside WCS limit
+    if stop_ref <= stop_wcs:  # ref is inside WCS limit
```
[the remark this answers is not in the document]
hbushouse: done

jwst/extract_1d/extract.py
```diff
@@ -738,16 +765,17 @@ def compare_start(start_ref, start_wcs):
     value : int or float
         The start limit, possibly constrained by the WCS start limit.
     """
-
-    if start_ref >= start_wcs:          # ref is inside WCS limit
+    if start_ref >= start_wcs:  # ref is inside WCS limit
```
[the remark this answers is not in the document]
hbushouse: done

PR #5094, opened 2020-06-20
Reactivate failing NIRSpec regtest

hbushouse:
Now that the NIRSpec readnoise rmap has been fixed in the latest CRDS context, the NIRSpec regtest that had been xfail-ed a couple weeks ago can be reinstated, because ref file retrieval is successful again.


jdavies-st:
Remove the `crds.exceptions.CrdsLookupError` import before merging.

PR #5095, opened 2020-06-23
Jp 918 Correct bar shadow parity bug for yslit

### LINKED ISSUE #3897: NIRSpec bar shadow parity bug
Issue [JP-918](https://jira.stsci.edu/browse/JP-918) was created by James Muzerolle:

There is a possible parity error in the bar shadow correction.  Our preliminary validation test shows discrepancies at the edges of MOS spectral traces; I can get much better agreement by reversing the sign of the slit_y parameter in our code.    The problem is in the equation for the reference file row index (line 92 in bar_shadow.py), I think the sign in front of "yslit" should be positive, not negative:

yrow = index_of_fiducial_in_array + yslit * shutter_height

ddavis-stsci:
This is a correction for the fact that the yslit values in a 2D science spectrum _decrease_ from bottom to top but in the reference file they _increase_ from bottom to top. 

Fixes #3897 / [JP-918](https://jira.stsci.edu/browse/JP-918)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,10 @@ associations
 
 - Update diagrams to change sloper to detector1. [#4986]
 
+  barshadow
+---------
+- Correct bar shadow parity bug for yslit. [#5095]
+
 combine_1d
 ----------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,7 +16,7 @@ associations
 
 - Update diagrams to change sloper to detector1. [#4986]
 
-  barshadow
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -16,6 +16,10 @@ associations
 
 - Update diagrams to change sloper to detector1. [#4986]
 
+  barshadow
```
hbushouse, line 19: Not sure it matters, but you might need to remove the extra whitespaces in front of this section title.

PR #5096, opened 2020-06-23
Unset drop frames1 crashes ramp fit

mwregan2:
If the FITS header keyword DRPFRMS1 is not present, then the model attribute drop_frames1 has a value of None. The protection logic that checked for this error was skipped in the latest version of ramp_fit. This PR adds a test case that recreates the problem and reproduces the check for the unset value. If it is None then it is set to 0.


dmggh:
About 'maximum_cores' - I admit this is a bit picky, but I find 'one' (or 'single_core') to be more intuitive than 'none' to designate that only a single core should be used. The other strings ('quarter', etc) designate the fraction of the maximum number of cores, but one/none/single_core is different.  Also, 'none' (the string) may be casually misinterpreted as the python None (NoneType).  I'm not sure what the best answer is.

mwregan2:
David,
  I’m with you. James didn’t like ‘one’ and preferred ‘none’. One is more in family. If you can convince James it would be easy to change
Mike

mwregan2:
Howard,
  I fixed the extra white space issue but I don’t know what changing the log entry means.
Mike

hbushouse:
@mwregan2 edit the file https://github.com/spacetelescope/jwst/blob/master/CHANGES.rst to add another entry under the ramp_fitting heading that mentions this update and the PR number.

mwregan2:
Howard,
  Done, It’s a separate PR since my fork is behind master right now.
Mike

PR #5107, opened 2020-06-29
Make calwebb_tso3 more robust

hbushouse:
A user recently reported that MIRI imaging TSO data were going through the spectroscopic branch of the ``calwebb_tso3`` pipeline, which turned out to be due to the fact that the TSOVISIT keyword was missing. These updates make the ``calwebb_tso3`` pipeline more robust to that kind of obvious problem with the input data and causes it to issue an error message and abort. Also needed to update the ``is_tso`` helper function, because it was assuming that any ``CubeModel`` data were TSO, which is not correct: coronagraphy data is also ``CubeModel``. Also just did a little general clean-up and addition of comments, etc. to the modules involved, as well as adding a description of the keyword logic to the docs.

At some point we should probably also try to figure out how to update the ``extract_1d`` and ``white_light`` steps to have them figure out if they've actually been given spectroscopic data as input and if not, then abort. Before the updates here, MIRI imaging data actually somehow made it through both of those steps, which is crazy.

### DIFF
--- a/jwst/lib/pipe_utils.py
+++ b/jwst/lib/pipe_utils.py
@@ -3,10 +3,6 @@
 import numpy as np
 
 from ..associations.lib.dms_base import TSO_EXP_TYPES
-from ..datamodels import CubeModel
-
-# Model types that typically represent TSO's
-TSO_MODEL_TYPES = (CubeModel,)
 
 
 def is_tso(model):
@@ -42,9 +38,6 @@ def is_tso(model):
             is_tso = False
     except AttributeError:
         pass
-
-    # Check on model type
-    is_tso = is_tso or isinstance(model, TSO_MODEL_TYPES)
 
     # We've checked everything.
     return is_tso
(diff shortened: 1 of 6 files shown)


jwst/lib/pipe_utils.py
```diff
@@ -3,10 +3,6 @@
 import numpy as np
 
 from ..associations.lib.dms_base import TSO_EXP_TYPES
-from ..datamodels import CubeModel
-
-# Model types that typically represent TSO's
-TSO_MODEL_TYPES = (CubeModel,)
```
jdavies-st, line 9: One of the tests in
```
jwst/lib/tests/test_pipe_utils.py
```
tests this.  You'll want to modify or remove the test.

hbushouse: Done (in latest commit)

hbushouse:
For some unknown reason github thinks the Travis runs haven't completed yet, but they have (successfully). Some I'm merging.

Issue #4540, opened 2020-02-07
outlier_detection fails if parameter save_intermediate_results is set to True


stscijgbot:
Issue [JP-1283](https://jira.stsci.edu/browse/JP-1283) was created by Juergen Schreiber:

See the following in Image3Pipeline that fails at least in version 0.13.8 but I think

it is not resolved in newer versions.

dm_3_container = datamodels.ModelContainer("IMA_asn.json")

Image3Pipeline.call(dm_3_container, save_results=True, output_dir=output_dir,
 steps=\{'skymatch':{'skip':True},
 'source_catalog':\{'save_results':True, 'kernel_fwhm': 5.0, 'snr_threshold': 20.},
 'outlier_detection':\{'save_intermediate_results': True}})

 

with the following error message indicating that save_model is not inherited from Step in the OutlierDetection class.

 

~/anaconda/envs/miricle.devel/lib/python3.6/site-packages/jwst-0.13.8-py3.6.egg/jwst/outlier_detection/outlier_detection_step.py in process(self, input)
 164 # Set up outlier detection, then do detection
 165 step = detection_step(self.input_models, reffiles=reffiles, **pars)
--> 166 step.do_detection()
 167 
 168 if self.input_container:

~/anaconda/envs/miricle.devel/lib/python3.6/site-packages/jwst-0.13.8-py3.6.egg/jwst/outlier_detection/outlier_detection.py in do_detection(self)
 193 if save_intermediate_results:
 194 log.info("Writing out resampled exposures...")
--> 195 self.save_model(
 196 model,
 197 output_file=model.meta.filename,

AttributeError: 'OutlierDetection' object has no attribute 'save_model'

stscijgbot:
Comment by James Davies: This is the same as https://jira.stsci.edu/browse/JP-1078

I've linked that issue to this.

stscijgbot:
Comment by Sarah Kendrew: I would like to add that being able to produce & inspect intermediate output files is critical to our (MIRI) commissioning analysis work (for which we have rehearsals coming up in spring) - some feedback on this ticket or any workarounds would be much appreciated.

stscijgbot:
Comment by Misty Cracraft: Ors Detre has a workaround bit of code for this.

Hi Misty, Dave,

Yes, it is available on the GitHub of CAP-201:

[https://github.com/JWST-MIRI/miricap201]

[...]

PR #5108, opened 2020-06-29
JP-1283 Fix bug in outlier_detection to save intermediate results

ddavis-stsci:
Updated the outlier detection code to save intermediate results when save_intermediate_results = True.
Fixes [JP-1283](https://jira.stsci.edu/browse/JP-1283) , #4540

### DIFF
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -194,11 +194,16 @@ class OutlierDetection:
             for model in drizzled_models:
                 if save_intermediate_results:
                     log.info("Writing out resampled exposures...")
-                    self.save_model(
-                        model,
-                        output_file=model.meta.filename,
-                        suffix=self.resample_suffix
+                    self.save_results = True
+                    self.output_use_model = model.meta.filename
+                    self.log = log
+                    Step.save_model(
+                        self,
+                        model=model,
+                        output_file=model.meta.filename.lstrip("_"),
+                        suffix=self.resample_suffix.lstrip("_").replace('.fits','')
                     )
+                    self.save_resutls = False
         else:
             drizzled_models = self.input_models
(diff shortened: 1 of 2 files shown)


jwst/outlier_detection/outlier_detection.py
```diff
@@ -194,11 +194,16 @@ def do_detection(self):
...
-                        output_file=model.meta.filename,
-                        suffix=self.resample_suffix
+                    self.save_results = True
+                    self.output_use_model = model.meta.filename
+                    self.log = log
+                    Step.save_model(
+                        self,
+                        model=model,
+                        output_file=model.meta.filename.lstrip("_"),
+                        suffix=self.resample_suffix.lstrip("_").replace('.fits','')
```
hbushouse, line 204: Typically the inputs to `outlier_detection` are individual cal products with file names like "jw00632011001_04102_00001_nrca3_cal.fits". If `lstrip("_")` removes everything to the left of the first occurrence of "_" in that name, won't it be left with just "jw00632011001"? Or does it work from the last occurrence of "_"? I guess ultimately the desired file name is "jw00632011001_04102_00001_nrca3_i2d.fits", right? Or something else, cuz I guess that could conflict with existing files of that same name that were created by the `calwebb_image2` pipeline.

stscieisenhamer: Ahhh yes. Trying to do saving while not in a Step object. The solution should be just below this code section. In the master branch, look to line 219.

jwst/outlier_detection/outlier_detection.py
```diff
@@ -194,11 +194,16 @@ def do_detection(self):
...
                     )
+                    self.save_resutls = False
```
jdavies-st, line 206: typo `resutls`.

PR #5110, opened 2020-06-30
Jp 1457 Add a hash set/check to DataModel I/O to check whether schema traversal is necessary

### LINKED ISSUE #4952: Add hash checking of the FITS component to DataModel I/O
_Issue [JP-1457](https://jira.stsci.edu/browse/JP-1457) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

An optimization that can greatly speed up the reading of FITS files as `DataModels` is to skip the FITS verification/loading routine\` load_from_schema`. This is the routine that reads the FITS headers, according to the provided schemas, and loads ASDF tree. This is the slowest part of the I/O since it requires walking the schema, often over arrays of schemas.

This routine is necessary when either the FITS file does not have the ASDF component already, or to ensure that any changes to FITS keywords are incorporated into the ASDF component.

When a `DataModel` is written to file, the ASDF component is added to the file and contains all the information that is also in the FITS headers. Hence, on read, if it can be determined that the FITS component had not changed, the ASDF component can simply be read and the `load_from_schema` routine skipped.

This can be checked programmatically when, on write, a hash of the FITS headers is calculated and added to the ASDF component. Then, on read, the FITS hash is re-calculated. If it is the same as the stored hash, then the routine can safely be skipped.

In resolving JP-1343, this approach was experimented with and a few issues where found, none of which were show-stopping in nature.

The first is the data components. Data arrays are kept in a FITS HDU and not duplicated in the ASDF tree, conserving memory. This is implemented in ASDF itself using a lazy-load method. A side effect of `load_from_schema` is to trigger the loading of these data segments. Skipping this is fine; the lazy loading kicks in when the data are actually referred to later on. However, this requires the FITS file to remain open during the life of the `DataModel`. Again, this is fine except there is inconsistency in `DataModes` code itself; sometimes the FITS file is left open, sometimes not. An initial PR #&#x2060;4930, resolves this issue. However, this does cascade into how DataModels are further copied, and closed, in application code. Some examination of this will need be done. Or the option is to just do the data-schema loading regardless. This component of the load is quite quicker than the FITS keywords, only because there are not that many of them compared to the number of keywords to load.

Another thing that `load_from_schema` does is determine which HDUs are not part of the schema at all, placing them in the `extra_fits` part of the ASDF tree. A way of determining extra_fits will need be done without having to incur the full cost of the loading.

stscieisenhamer:
Resolves #4952 
Resolves [JP-1457](https://jira.stsci.edu/browse/JP-1457)

### DIFF
--- a/jwst/datamodels/dqflags.py
+++ b/jwst/datamodels/dqflags.py
@@ -59,12 +59,11 @@ pixel = {'GOOD':             0,      # No bits set, all is good
 
 # Group-specific flags. Once groups are combined, these flags
 # are equivalent to the pixel-specific flags.
-group = {
-    'GOOD':       pixel['GOOD'],
-    'DO_NOT_USE': pixel['DO_NOT_USE'],
-    'SATURATED':  pixel['SATURATED'],
-    'JUMP_DET':   pixel['JUMP_DET'],
-    'DROPOUT':    pixel['DROPOUT'],
+group = {'GOOD':       pixel['GOOD'],
+         'DO_NOT_USE': pixel['DO_NOT_USE'],
+         'SATURATED':  pixel['SATURATED'],
+         'JUMP_DET':   pixel['JUMP_DET'],
+         'DROPOUT':    pixel['DROPOUT'],
 }
 
 
@@ -118,20 +117,23 @@ def bit_flags_to_mnemonics(bit_flags):
 
     Returns
     -------
-    mnemonics : [str[,...]]
-        List of mnemonics represented by the set bits
+    mnemonics : {str[,...]}
+        Set of mnemonics represented by the set bits
 
     Examples
     --------
     >>> bit_flags_to_mnemonics(1)
-    ['DO_NOT_USE']
+    {'DO_NOT_USE'}
 
-    >>> bit_flags_to_mnemonics(7)
-    ['DO_NOT_USE', 'SATURATED', 'JUMP_DET']
+    >>> bit_flags_to_mnemonics(7)             #doctest: +SKIP
+    {'JUMP_DET', 'DO_NOT_USE', 'SATURATED'}
+
+    >>> bit_flags_to_mnemonics(7) == {'JUMP_DET', 'DO_NOT_USE', 'SATURATED'}
+    True
     """
-    mnemonics = [
+    mnemonics = {
         mnemonic
         for mnemonic, value in pixel.items()
         if (bit_flags & value)
-    ]
+    }
     return mnemonics
--- a/jwst/datamodels/dqflags.py
+++ b/jwst/datamodels/dqflags.py
@@ -59,11 +59,12 @@ pixel = {'GOOD':             0,      # No bits set, all is good
 
 # Group-specific flags. Once groups are combined, these flags
 # are equivalent to the pixel-specific flags.
-group = {'GOOD':       pixel['GOOD'],
-         'DO_NOT_USE': pixel['DO_NOT_USE'],
-         'SATURATED':  pixel['SATURATED'],
-         'JUMP_DET':   pixel['JUMP_DET'],
-         'DROPOUT':    pixel['DROPOUT'],
+group = {
+    'GOOD':       pixel['GOOD'],
+    'DO_NOT_USE': pixel['DO_NOT_USE'],
+    'SATURATED':  pixel['SATURATED'],
+    'JUMP_DET':   pixel['JUMP_DET'],
+    'DROPOUT':    pixel['DROPOUT'],
 }
 
 
@@ -105,3 +106,32 @@ def interpret_bit_flags(bit_flags, flip_bits=None):
         bit_flags_dm = multiple_replace(bit_flags, dm_flags)
 
     return ap_interpret_bit_flags(bit_flags_dm, flip_bits=flip_bits)
+
+
+def bit_flags_to_mnemonics(bit_flags):
+    """Interpret value as bit flags and return the mnemonics
+
+    Parameters
+    ----------
+    bit_flags : int-like
+        The value to interpret as bit flags
+
+    Returns
+    -------
+    mnemonics : [str[,...]]
+        List of mnemonics represented by the set bits
+
+    Examples
+    --------
+    >>> bit_flags_to_mnemonics(1)
+    ['DO_NOT_USE']
+
+    >>> bit_flags_to_mnemonics(7)
+    ['DO_NOT_USE', 'SATURATED', 'JUMP_DET']
+    """
+    mnemonics = [
+        mnemonic
+        for mnemonic, value in pixel.items()
+        if (bit_flags & value)
+    ]
+    return mnemonics
--- a/jwst/datamodels/dqflags.py
+++ b/jwst/datamodels/dqflags.py
@@ -107,33 +107,43 @@ def interpret_bit_flags(bit_flags, flip_bits=None):
     return ap_interpret_bit_flags(bit_flags_dm, flip_bits=flip_bits)
 
 
-def bit_flags_to_mnemonics(bit_flags):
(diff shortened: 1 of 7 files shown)


stscieisenhamer:
Merge only after PR #4930 is merged.

jwst/datamodels/tests/test_models.py
```diff
@@ -49,6 +49,95 @@ def teardown():
     shutil.rmtree(TMP_DIR)
 
 
+@pytest.fixture(scope='module')
+def empty_model():
+    """Create an empy model"""
+    return DataModel()
```
jdavies-st, line 55: This is potentially dangerous to scope this to "module", unless there's a good reason. Better to leave the scope at function, so there's not any unintended effects on one test from another, and dependency on run order.

jwst/outlier_detection/outlier_detection.py
```diff
@@ -497,7 +497,7 @@ def flag_cr(sci_image, blot_image, **pars):
     log.debug("Pixels in cr_mask:  {}".format(count_cr))
 
     # Update the DQ array in the input image in place
-    np.bitwise_or(sci_image.dq, np.invert(cr_mask) * CRBIT, sci_image.dq)
+    sci_image.dq = np.bitwise_or(sci_image.dq, np.invert(cr_mask) * CRBIT)
```
jdavies-st, line 500: Hey does this fix a known bug?!

jwst/datamodels/dqflags.py
```diff
@@ -105,3 +106,32 @@ def interpret_bit_flags(bit_flags, flip_bits=None):
...
+    """Interpret value as bit flags and return the mnemonics
+
+    Parameters
+    ----------
+    bit_flags : int-like
+        The value to interpret as bit flags
+
+    Returns
+    -------
+    mnemonics : [str[,...]]
```
eslavich, line 121: Would it be helpful to return these as a `set` instead of a `list`?  Not that the list will ever be long enough to gain any efficiency there (membership test may even be slower for such a short list), but the set intersection/union/etc methods might be handy.

jwst/datamodels/fits_support.py
```diff
@@ -594,3 +636,91 @@ def from_fits_hdu(hdu, schema):
...
+    Parameters
+    ----------
+    skip_fits_update : bool
+        Regardless of FIT hash check, attempt to skip if requested.
+
+    hdulist : astropy.io.fits.HDUList
+        The input FITS information
+
+    asdf_struct : asdf.ASDFFile
+        The association ASDF structure
```
eslavich, line 656: Was this meant to be `associated`?

jwst/datamodels/fits_support.py
```diff
@@ -594,3 +636,91 @@ def from_fits_hdu(hdu, schema):
...
+        The FITS structure.
+
+    Returns
+    -------
+    fits_hash : str
+        The hash of all HDU headers.
+    """
+    fits_hash = hashlib.sha256()
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore', AstropyWarning)
```
eslavich, line 720: Worth a comment noting what warning in particular is being ignored

jwst/datamodels/tests/test_models.py
```diff
@@ -49,6 +49,95 @@ def teardown():
...
+
+
+@pytest.fixture 
+def jail_environ():
+    """Lock cheanges to the environment"""
```
eslavich, line 60: `cheanges` --> `changes`

jwst/datamodels/tests/test_models.py
```diff
@@ -49,6 +49,95 @@ def teardown():
...
+    original = os.environ.copy()
+    try:
+        yield
+    finally:
+        os.environ = original
+
+
+@pytest.fixture(scope='module')
```
eslavich, line 68: This one probably also ought to be a function-scoped fixture.  We're only using it in one function anyway, so the module-scoping doesn't save us anything.

jwst/datamodels/util.py
```diff
@@ -444,3 +445,30 @@ def create_history_entry(description, software=None):
     if software is not None:
         entry['software'] = software
     return entry
+
+
+def get_envar_as_boolean(name, default=False):
```
eslavich, line 450: I realize this is existing code that was just copied to a new module, but I would whitelist the falsey values here and raise `ValueError` for an unrecognized string.  This would prevent the code from silently treating strings like "true " (note trailing space) as `False`.

jwst/datamodels/tests/test_models.py
stscieisenhamer: This one is generating files, so it does not suffer the same issues as an open `DataModel` as before. And, though it is just being used in one test function, that function is parametrized and is re-executed, currently, 6 times.

As such, module scope should be appropriate.

eslavich: But since it's a module-scoped fixture, it only generates those files once _per module_, and re-uses the generated result for every test in the module.  Those 6 executions of the parametrized test are sharing the same files, is that the intention?

eslavich: Here's an example:

```python
import pytest

def some_module_fixture():
    return {"values": []}

@pytest.mark.parametrize("value", [1, 2, 3])
def test_something(some_module_fixture, value):
    some_module_fixture["values"].append(value)
    assert some_module_fixture["values"] == [value]
```

The first of the parametrized tests succeeds, but the next two fail because the values start accumulating in the array:

```
some_module_fixture = {'values': [1, 2]}, value = 2

    @pytest.mark.parametrize("value", [1, 2, 3])
    def test_something(some_module_fixture, value):
        some_module_fixture["values"].append(value)
>       assert some_module_fixture["values"] == [value]
E       assert [1, 2] == [2]
E         At index 0 diff: 1 != 2
E         Left contains one more item: 2

asdf/tests/test_fixture.py:10: AssertionError
```

jwst/outlier_detection/outlier_detection.py
stscieisenhamer: Actually no and this was one of the issues to be brought up during the team meeting but time ran out. Before this PR, all data arrays are loaded by this point, so numpy functions that allow operation in-place work fine. However, with this PR, having the data arrays loaded are not guaranteed. Thus, at this point the numpy arrays can get a `Block` node object. This routine does a check to ensure that the output can be written to, which now fails because of the unknown `Block` object type.

jwst/datamodels/tests/test_models.py
stscieisenhamer: Claim red-herring argument! This is true for any fixture scoped greater than function. If a fixture is being misused, then expect the consequences.  Current fixture is creating resources that is appropriate for the test in use. 🐟

stscieisenhamer:
WRT the `bit_flags_to_mnemonics` function: Have done the suggested rename and added an argument for the different mnemonic mappings.

jwst/datamodels/fits_support.py
```diff
@@ -462,7 +462,7 @@ def _load_from_schema(hdulist, schema, tree, context, skip_fits_update=False):
         if not any(isinstance(hdu, fits.BinTableHDU) for hdu in hdulist if hdu.name != 'ASDF'):
             log.debug('Skipping FITS updating completely.')
             return known_keywords, known_datas
-        log.debug('Skipping FITS keyword updating. FITS Table and data HDU updating will occur.')
+        log.debug('Skipping FITS keyword updating. FITS BinTableHDU and ImageHDU updating will occur.')
```
jdavies-st, line 465: The debug text should be something like `"Skipping FITS keyword updating except for BinTableHDU and it's associated header keywords"`.  `ImageHDU` objects are never updated under any circumstances, because they are not stored separately in the `AsdfInFITS` tree.

PR #5112, opened 2020-06-30
JP-1556: Store both versions of pathloss correction

### LINKED ISSUE #5103: Save both pathloss point source and uniform source correction arrays
_Issue [JP-1556](https://jira.stsci.edu/browse/JP-1556) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

As part of the implementation of NIRSpec master background updates outlined in JP-1213, it would facilitate corrections needed to support IFU mode if the calwebb_spec2 pathloss step computed and saved both point source and uniform source versions of the correction, at least for the NIRSpec IFU mode (and possibly FS and MOS too). Modify the pathloss step to compute and save both versions to the output data model. The data model schema itself may need to be updated to define both versions of the extension.

hbushouse:
Update all of the `datamodels` schemas that contain a "pathloss" array to now contain both a "pathloss_ps" and "pathloss_un" (point source and uniform, respectively). Update the `pathloss` step to always compute both 2D versions, apply the appropriate one, and store both versions in the output. For NIRISS SOSS, which only has one version of the pathloss correction, store as "pathloss_ps", because it's assumed that only point sources will be observed with that mode.

Fixes #5103 / [JP-1556](https://jira.stsci.edu/browse/JP-1556)

### DIFF
--- a/jwst/datamodels/ifuimage.py
+++ b/jwst/datamodels/ifuimage.py
@@ -23,9 +23,6 @@ class IFUImageModel(DataModel):
     zeroframe : numpy float32 array
          Zeroframe array
 
-    area : numpy float32 array
-         Pixel area map array
-
     var_poisson : numpy float32 array
          variance due to poisson noise
 
@@ -35,8 +32,14 @@ class IFUImageModel(DataModel):
     wavelength : numpy float32 array
          wavelength
 
-    pathloss : numpy float32 array
-         pathloss correction
+    pathloss_ps : numpy float32 array
+         pathloss correction for point source
+
+    pathloss_un : numpy float32 array
+         pathloss correction for uniform source
+
+    area : numpy float32 array
+         Pixel area map array
     """
     schema_url = "http://stsci.edu/schemas/jwst_datamodel/ifuimage.schema"
 
@@ -47,12 +50,12 @@ class IFUImageModel(DataModel):
             self.data = init.data
             self.dq = init.dq
             self.err = init.err
-            if init.hasattr('area'):
-                self.area = init.area
             if init.hasattr('var_poisson'):
                 self.var_poisson = init.var_poisson
             if init.hasattr('var_rnoise'):
                 self.var_rnoise = init.var_rnoise
+            if init.hasattr('area'):
+                self.area = init.area
             return
 
         super(IFUImageModel, self).__init__(init=init, **kwargs)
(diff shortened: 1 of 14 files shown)


jwst/datamodels/ifuimage.py
```diff
@@ -35,8 +32,14 @@ class IFUImageModel(DataModel):
...
 
-    pathloss : numpy float32 array
-         pathloss correction
+    pathloss_ps : numpy float32 array
+         pathloss correction for point source
+
+    pathloss_un : numpy float32 array
+         pathloss correction for uniform source
+
+    area : numpy float32 array
```
hbushouse, line 41: Note to reviewers: the `area` object just moved in the order, so that it shows up after all the other extensions in the actual FITS files (that's just my OCD kicking in).

jwst/datamodels/ifuimage.py
```diff
@@ -47,12 +50,12 @@ def __init__(self, init=None, **kwargs):
...
             self.dq = init.dq
             self.err = init.err
-            if init.hasattr('area'):
-                self.area = init.area
             if init.hasattr('var_poisson'):
                 self.var_poisson = init.var_poisson
             if init.hasattr('var_rnoise'):
                 self.var_rnoise = init.var_rnoise
+            if init.hasattr('area'):
+                self.area = init.area
```
hbushouse, line 58: Same OCD mod here. Nothing substantive.

jwst/pathloss/pathloss.py
```diff
@@ -10,9 +10,10 @@
 log.setLevel(logging.DEBUG)
 
 
-# There are 30 slices in the NIRSPEC IFU, numbered from 0 to 29
+# There are 30 slices in the NIRSpec IFU, numbered from 0 to 29
```
hbushouse, line 13: A LOT of mods in this module are purely cosmetic, just for readability, flake8, etc. Only substantive changes directly involve `pathloss_ps` and `pathloss_un`.

jwst/datamodels/ifuimage.py
jdavies-st: I think names like `pathloss_point` and `pathloss_uniform` would be a bit more descriptive.

hbushouse: Agreed. I wanted to keep the EXTNAME values short, but the data array names could be spelled out more fully.

hbushouse: Fixed in latest commit.

hbushouse:
Also added some units tests in `datamodels`.

Issue #5035, opened 2020-06-11
NIRSpec BOTS data with NINT=3000 crashes TSO3


stscijgbot-jp:
_Issue [JP-1505](https://jira.stsci.edu/browse/JP-1505) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

When processing a NIRSpec BOTS exposure containing 3000 integrations through calwebb_tso3, the pipeline crashed at the outlier detection step with the error message "combine: too many arrays".  Is there a limit on the number of integrations the step can handle?  I was able to successfully process a similar exposure with only 600 integrations.

stscijgbot-jp:
_Comment by [David Davis](https://jira.stsci.edu/secure/ViewProfile.jspa?name=ddavis) on [JIRA](https://jira.stsci.edu/browse/JP-1505?focusedCommentId=470614#comment-470614):_

This is a limitation of [stsci.image](https://github.com/spacetelescope/stsci.image)/[src](https://github.com/spacetelescope/stsci.image/tree/master/src)/*_combinemodule.c.*  The module has a limit of 1800 input images 
```java
#define MAX_ARRAYS 1800
```
As we requested. ([Bump number up as H. Bushouse requested](https://github.com/spacetelescope/stsci.image/commit/0271b8d70f63608b6701c58910e33d3be8811988)).

What should we request for TSO observations? Is there a practical limit?

stscijgbot-jp:
_Comment by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol) on [JIRA](https://jira.stsci.edu/browse/JP-1505?focusedCommentId=470929#comment-470929):_

Yes, the number of integrations per exposure cannot exceed 65535.

stscijgbot-jp:
_Comment by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse) on [JIRA](https://jira.stsci.edu/browse/JP-1505?focusedCommentId=471285#comment-471285):_

[...]

PR #5114, opened 2020-06-30
JP-1505 Limit number of images to median combine in outlier detection

ddavis-stsci:
Avoids surpassing max limit imposed in `stsci.image.median`

Fixes [JP-1505](https://jira.stsci.edu/browse/JP-1505) , #5035

Passes outlier_detection unit tests and miri_mrs outlier_detection tests.


jwst/outlier_detection/outlier_detection.py
```diff
@@ -291,8 +291,16 @@ def create_median(self, resampled_models):
         # Compute median of stack of images using `badmasks` to remove
         # low-weight values.  In the future we should use a masked array
         # and np.median
-        median_image = median(resampled_sci, nlow=nlow, nhigh=nhigh,
-                              badmasks=badmasks)
+        max_images_for_median = 1024
```
jdavies-st, line 294: Looking at the comment above, we could avoid using `stsci.image.median` completely by just using `np.median` and using a masked array.  Or by turning the masked values into nans and using `np.nanmedian`.  Either would work, and would use all the data to create the median.  Using `np.nanmedian` might take less memory and be faster.

The only reason we are using `stsci.image.median` is for the case when there are several images in the stack, and one _might_ want to toss out the high and low values before doing the median, though we don't use this mode at all in pipeline processing.

Or one could take a hybrid approach and use `stsci.image.median` for cases were n <= 1024, and just do a straight median using `np.nanmedian` for the >1024  case, as there's no chance throwing out a few high or low cases will matter at all if there are >1024 images in the stack.

That said, even if we are using plain `np.nanmedian` here, on a stack of 3000 integrations, the memory footprint would be very high for a 2048x2048 image. I assume we are not using full arrays here but subarrays?

ddavis-stsci: From my point of view I'd ask what are we trying to accomplish? So to determine a robust median value as a baseline  to look for anomalous values then 1K inputs should be fine. I'm not sure what adding another 50, 100, or 2000 images will accomplish. If the additional images change the results significantly then we have a different problem and that should be addressed separately. So for outlier detection I think just picking the first N images should work well. 

If we need a median image for a large number of images then moving away from stimage might be a good idea. As you say the memory footprint for thousands of 1024^2 images would be an issue.

hbushouse: We should never have a situation of a very large number of large images. The current problem case is for a very large stack of 2D spectra, but the 2D cutouts are not terribly big. Still, if we ever get a situation where there are 30-60k of these (which is allowed in the APT), we could be in trouble in terms of memory footprint.

I agree that we should just take this opportunity to switch all cases to using `np.nanmedian` and be done with the limitation imposed in `stsci.image.median`. I also think it's appropriate to limit the number of inputs we use in `np.nanmedian`, because we could get into memory footprint issues in the future if we just blindly try to use all inputs. Setting a limit of 1024 is a reasonable baseline. Can always come back and tweak that if testing shows a different value works better.

hbushouse: Can we somehow use the existing `badmasks` to, for example, intentionally set those pixels to NaN so that they won't be used in the `np.nanmedian` calculation? Or use them to create a masked array to send to `np.nanmedian`.

hbushouse:
@ddavis-stsci I just checked out your PR branch and ran it on one regression test (so far), which was the test_miri_image.py module, running only the "test_miri_image_image3" test. It showed differences in the crf output file, but only in the DQ array. Nothing in the SCI or ERR arrays. I'll try some more tests. In which test(s) were you seeing the NaN's showing up?

hbushouse:
@ddavis-stsci you're right that the entire step appears to work on the input data models in place, rather than making the output as a copy of the input (like all other steps do), and hence any changes to the input (such as your new code to replace bad pixels with NaN) get propagated to the output products. A little playing around shows that adding a `.copy()` operation in the 2 lines:
```
resampled_sci = [i.data.copy() for i in resampled_models]
resampled_weight = [i.wht.copy() for i in resampled_models]
```
seems to fix the propagation problem. In fact you might not need it for the `resampled_weight`, because it looks like that doesn't get modified at all.

PR #5122, opened 2020-07-01
JP-1442 fix some issues with wfss mode

### LINKED ISSUE #4736: WFSS extraction boxes are too short
Issue [JP-1385](https://jira.stsci.edu/browse/JP-1385) was created by Camilla Pacifici:

Everything runs to completion, but the extraction boxes in spec2 are not long enough both in the R and C grism. This is a problem, because a significant part of the spectrum is cut. See screenshot.

!Screen Shot 2020-03-27 at 2.36.18 PM.png|width=415,height=400!

### LINKED ISSUE #4919: Investigate possible issues with WFSS WCSs
_Issue [JP-1442](https://jira.stsci.edu/browse/JP-1442) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

Investigate and fix if necessary issues with assign_wcs and extract_2d for WFSS mode:

 
 * list of extraction orders is not respected in extract_2d
 * "mmag_extract" is not passed as a parameter in extract_2d_step
 * 
 <https://github.com/spacetelescope/jwst/blob/master/jwst/assign_wcs/util.py#L652>
xmax,ymax, .... are absolute - they should be compared to xstart+xsize, not xsize
 * 
- <https://github.com/spacetelescope/jwst/blob/master/jwst/extract_2d/grisms.py#L406>
Should it be using xstart + xsize and ystart+ysize to pick up subarrays
 * util.create_grism_objects possibly marks too many object as "partial order", can the algorithm be improved
 * x-x0 can be negative  in image to grism_image transforms. In grismconf it is never negative. Should an absolute value be used or some other limit?
 * improve documentation

### LINKED ISSUE #5126: magnitude cut in extract2d for WFSS data
_Issue [JP-1566](https://jira.stsci.edu/browse/JP-1566) was created on JIRA by [Takahiro Morishita](https://jira.stsci.edu/secure/ViewProfile.jspa?name=tmorishita):_

This is not a bug report but suggestion for possible convenience:

Extract2dStep has an argument for magnitude cut, "mmag_extract", but this is currently only for NRC_TSGRISM. 
<https://github.com/spacetelescope/jwst/blob/f6863653995ed46a09fc8da7348a4de48cd09c2c/jwst/extract_2d/extract_2d.py#L67-L78>

We, NIRISS WFSS working group, think this option would be very helpful for users who wish to extract only bright objects, rather than everything in the field. The magnitude cut can be applied universally to the sources.

nden:
Fixes #4919 / [JP-1442](https://jira.stsci.edu/browse/JP-1442)

Fixes #5126 / [JP-1556](https://jira.stsci.edu/browse/JP-1566)

Fixes #4736/ [JP-1385](https://jira.stsci.edu/browse/JP-1385)

### DIFF
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -643,15 +643,21 @@ def create_grism_bbox(input_model,
 
                     xstack = np.hstack([x1, x2])
                     ystack = np.hstack([y1, y2])
-                    xmin = np.min(xstack)
-                    xmax = np.max(xstack)
-                    ymin = np.min(ystack)
-                    ymax = np.max(ystack)
 
-                    xmin = int(xmin) - input_model.meta.subarray.xstart
-                    xmax = int(xmax) - input_model.meta.subarray.xstart
-                    ymin = int(ymin) - input_model.meta.subarray.ystart
-                    ymax = int(ymax) - input_model.meta.subarray.ystart
+                    # Subarrays are only allowed in nircam tsgrism mode. The polynomial transforms
+                    # only work with the full frame coordinates. The code here is called during extract_2d,
+                    # and is creating bounding boxes which should be in the full frame coordinates, it just
+                    # uses the input catalog and the magnitude to limit the objects that need bounding boxes.
+
+                    # Tsgrism is always supposed to have the source object at the same pixel, and that is
+                    # hardcoded into the transforms. At least a while ago, the 2d extraction for tsgrism mode
+                    # didn't call this bounding box code. So I think it's safe to leave the subarray
+                    # subtraction out, i.e. do not subtract x/ystart.
+
+                    xmin = int(np.min(xstack))
+                    xmax = int(np.max(xstack))
+                    ymin = int(np.min(ystack))
+                    ymax = int(np.max(ystack))
 
                     # don't add objects and orders which are entirely
                     # off the detector partial_order marks partial
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -583,7 +583,7 @@ def create_grism_bbox(input_model,
     # dispersion coefficients
 
     sky_to_detector = input_model.meta.wcs.get_transform('world', 'detector')
-    sky_to_grism = input_model.meta.wcs.get_transform('world', 'grism_detector')
+    sky_to_grism = input_model.meta.wcs.backward_transform
 
     # Get the disperser parameters which have the wave limits
     with WavelengthrangeModel(reference_files['wavelengthrange']) as f:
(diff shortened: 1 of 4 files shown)


jwst/assign_wcs/util.py
```diff
@@ -615,25 +615,24 @@ def create_grism_bbox(input_model,
...
+                    ystack = np.hstack([y1, y2])
+                    xmin = np.min(xstack)
+                    xmax = np.max(xstack)
+                    ymin = np.min(ystack)
+                    ymax = np.max(ystack)
+
+                    xmin = int(xmin) - input_model.meta.subarray.xstart
+                    xmax = int(xmax) - input_model.meta.subarray.xstart
+                    ymin = int(ymin) - input_model.meta.subarray.ystart
+                    ymax = int(ymax) - input_model.meta.subarray.ystart
```
nden, line 635: @sosey It seems `xmin`, `ymin`, ..., are in full frame coordinates. Since the bounding box is constructed form them  the origin of the subarray needs to be subtracted. Does this sound right? I haven't see a simulated data set with a subarray to test it.

hbushouse: The APT NIRCam WFSS template has the subarray choice hardwired to "FULL" and the NIRISS WFSS template doesn't even have a subarray field (hence default is FULL). So you may not need to worry about this.

sosey: If I remember correctly, subarrays are only allowed in nircam tsgrism mode. The polynomial transforms only work with the full frame coordinates. The code here is called during extract_2d, and is creating bounding boxes which should be in the full frame coordinates, it just uses the input catalog and the magnitude to limit the objects  that need bounding boxes.  

tsgrism is always supposed to have the source object at the same pixel, and that is hardcoded into the transforms. At least a while ago, the 2d extraction for tsgrism mode didn't call this bounding box code. So I think you're safe to leave the subarray subtraction out.  You might add a note here for the benefit of future code peepers if you agree :)

nden: @sosey Another quick question - what's the reason for having these as integer values?

nden: After looking at the code again, the bounding boxes of the slits are created from the shape of the extracted array. The `xmin`, ..., values are used to define the extraction indices of the arrays, so that's why they need to be integers. So ignore my question.

sosey: I think it was just to give an integer pixel value for the bounding box corners. If that doesn't matter you can exclude the int. I just noticed that later on there is a round() called as well. I think both are definitely not needed if you decide to keep the integer value in the order bounding.

nden:
The requested changes were made. This PR needs lifting the change request so it can be merged.

nden:
A Niriss slit with the source and the extraction box:

one_slit

nden:
For the record this is how NIRCAM WFSS looks:

nrc_slit

PR #5125, opened 2020-07-02
JP-1557: Update master_background for NRS IFU pathloss

### LINKED ISSUE #5104: Update the master background step to correct NIRSpec IFU point sources
_Issue [JP-1557](https://jira.stsci.edu/browse/JP-1557) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

As outlined in JP-1213 and JP-1493, the calwebb_spec3 master background step needs to be updated to apply corrections to the 2D master bkg image when processing NIRSpec IFU point sources. If the science exposure contains a point source (as indicated by the SRCTYPE keyword), the 2D bkg image that is created by inflating the 1D master bkg spectrum to the wavelength grid of the science image, needs to be corrected as follows:

MBKG(corrected) = MBKG(raw) * PL(uniform) / PL(point)

where

PL(uniform) is the 2D pathloss correction array appropriate for a uniform source

PL(point) is the 2D pathloss correction array appropriate for a point source

PL(uniform) and PL(point) will be saved in the "cal" product of the science data that is input to the master bkg step (see JP-1556).

Once the master bkg 2D array has had these corrections applied, it can be subtracted from the science exposure as normal.

hbushouse:
Updated the `expand_to_2d` module to call the new function `correct_nrs_ifu_bgk` when the science data is a point source, in order to apply the point source vs. uniform source pathloss mods to the computed 2D background, before subtracting from the science data.

Fixes #5104 / [JP-1557](https://jira.stsci.edu/browse/JP-1557)

### DIFF
--- a/jwst/master_background/nirspec_corrections.py
+++ b/jwst/master_background/nirspec_corrections.py
@@ -1,0 +1,26 @@
+import logging
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+
+def correct_nrs_ifu_bkg(input):
+    """Apply point source vs. uniform source pathloss adjustments
+    to a NIRSpec IFU 2D master background array.
+
+    Parameters
+    ----------
+    input : `~jwst.datamodels.IFUImageModel`
+        The input background data.
+
+    Returns
+    -------
+    input : `~jwst.datamodels.IFUIMAGEModel`
+        An updated (in place) version of the input with the data
+        replaced by the corrected 2D background.
+    """
+
+    log.info('Applying point source pathloss updates to IFU background')
+    input.data *= (input.pathloss_point / input.pathloss_uniform)
+
+    return input
--- a/jwst/master_background/nirspec_corrections.py
+++ b/jwst/master_background/nirspec_corrections.py
@@ -21,6 +21,23 @@ def correct_nrs_ifu_bkg(input):
     """
 
     log.info('Applying point source pathloss updates to IFU background')
-    input.data *= (input.pathloss_point / input.pathloss_uniform)
+
+    # Try to load the appropriate pathloss correction arrays
+    try:
+        pl_point = input.getarray_noinit('pathloss_point')
+    except AttributeError:
+        log.warning('Pathloss_point array not found in input')
+        log.warning('Skipping pathloss background updates')
+        return input
+
+    try:
+        pl_uniform = input.getarray_noinit('pathloss_uniform')
+    except AttributeError:
+        log.warning('Pathloss_uniform array not found in input')
+        log.warning('Skipping pathloss background updates')
+        return input
+
+    # Apply the corrections
+    input.data *= (pl_point / pl_uniform)
 
     return input
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -64,7 +64,7 @@ master_background
 
 - Fix open files bug [#4995]
 
-- Update to include pathloss corrections to NIRSpec IFU background [#xxxx]
+- Update to include pathloss corrections to NIRSpec IFU background [#5125]
 
 outlier_detection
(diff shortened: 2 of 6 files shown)


jwst/master_background/nirspec_corrections.py
```diff
@@ -0,0 +1,26 @@
...
+
+    Returns
+    -------
+    input : `~jwst.datamodels.IFUIMAGEModel`
+        An updated (in place) version of the input with the data
+        replaced by the corrected 2D background.
+    """
+
+    log.info('Applying point source pathloss updates to IFU background')
+    input.data *= (input.pathloss_point / input.pathloss_uniform)
```
jdavies-st, line 24: What happens here if someone turns off the pathloss correction step?  Might want a `try` block.

hbushouse: Good point. But that raises an issue I was having while trying to run my first test on this new code. I was using an old cal file that still had the old "pathloss" array, but not the new "pathloss_point" or "pathloss_uniform" arrays. I could NOT find a useful way to test for the true existence of the new arrays. Any use of `hasattr(model, 'pathloss_point')` always returns `True`, because it's defined in the schema, `model.pathloss_point.shape` returns a valid looking `(2048,2048)`, etc. About the only way I could tell whether the data were real or not was to see if the data array was filled with zeros (which for some kinds of corrections could be valid). How can you reliably test for the true existence of an attribute like this in our datamodels?!

jdavies-st: That is quite irritating.  @stscieisenhamer is there a trick for doing this?

jwst/master_background/nirspec_corrections.py
```diff
@@ -21,6 +21,23 @@ def correct_nrs_ifu_bkg(input):
...
 
     log.info('Applying point source pathloss updates to IFU background')
-    input.data *= (input.pathloss_point / input.pathloss_uniform)
+
+    # Try to load the appropriate pathloss correction arrays
+    try:
+        pl_point = input.getarray_noinit('pathloss_point')
+    except AttributeError:
+        log.warning('Pathloss_point array not found in input')
+        log.warning('Skipping pathloss background updates')
```
jdavies-st, line 30: It might be good to set the step status keyword to "SKIPPED" in this block and the one below.

hbushouse: But the master background step itself is not being skipped in this case. Only the application of the corrections to the master bkg signal. Perhaps we should in fact skip the entire step if the corrections can't be applied.

jdavies-st: Ah, right.  I guess if the pathloss arrays are not there, the pathloss step was not run?  Does that mean this is just fine as is?

hbushouse: That's a good point. I was thinking of the case where someone was using old input files that don't have the proper pathloss arrays attached, but this would actually properly handle the case of the pathloss step having been skipped (I think). In that case the master bkg should already match the actual bkg, so there's no need to remove the previously applied correction.

hbushouse: But overall it would perhaps be best to skip the whole step, so that we don't deliver improperly calibrated results from the master bkg step. If someone has turned off pathloss upstream, they're going to get bad calibration on several fronts (i.e. flux cal won't be right, ...)

jdavies-st: Hmm, at that point perhaps we are getting to the point where things need to be managed by the pipeline.  If people decide to skip steps that are essential, then they do so at their own risk?  Perhaps your original approach is correct.  Just return the input, throw a big warning in the log, and be done with it.

PR #5127, opened 2020-07-02
 Added nan replacment routine.

dmggh:
As part of JP-1449, this update replaces nans in the input Sci array with the median of surrounding good pixels.

### DIFF
--- a/jwst/ami/ami_analyze.py
+++ b/jwst/ami/ami_analyze.py
@@ -13,7 +13,7 @@ from . import leastsqnrm
 from . import utils
 
 log = logging.getLogger(__name__)
-log.setLevel(logging.DEBUG)  ###### 0706 
+log.setLevel(logging.DEBUG)
 
 def apply_LG(input_model, filter_model, oversample, rotation):
     """
--- a/jwst/ami/ami_analyze.py
+++ b/jwst/ami/ami_analyze.py
@@ -13,7 +13,7 @@ from . import leastsqnrm
 from . import utils
 
 log = logging.getLogger(__name__)
-
+log.setLevel(logging.DEBUG)  ###### 0706 
 
 def apply_LG(input_model, filter_model, oversample, rotation):
     """
@@ -85,6 +85,8 @@ def apply_LG(input_model, filter_model, oversample, rotation):
 
     input_data = new_img_model.data.copy()
     input_model.data = input_data.astype(np.float64)
+
+    del datamodel_img_model, new_img_model
 
     subarray = input_model.meta.subarray.name.upper()
     if subarray == 'FULL':
--- a/jwst/ami/ami_analyze.py
+++ b/jwst/ami/ami_analyze.py
@@ -10,9 +10,9 @@ from .. import datamodels
 from .nrm_model import NrmModel
 from . import webb_psf
 from . import leastsqnrm
+from . import utils
 
 log = logging.getLogger(__name__)
-log.setLevel(logging.DEBUG)
 
 
 def apply_LG(input_model, filter_model, oversample, rotation):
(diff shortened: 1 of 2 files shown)


jwst/ami/utils.py
```diff
@@ -285,3 +287,100 @@ def findmax(mag, vals, mid=1.0):
...
+    try:
+        data_array = input_array[xc - hbox:xc + hbox, yc - hbox: yc + hbox]
+        dq_array = input_dq_array[xc - hbox:xc + hbox, yc - hbox: yc + hbox]
+    except IndexError:
+        # If the box is outside the data return 0
+        log.warning('Box for median filter is outside the data.')
+        return 0.
+
+    filtered_array = data_array[(dq_array != dqflags.pixel['DO_NOT_USE']) &
+                     (np.isfinite(data_array))]
```
hbushouse, line 332: You might not need to include the extra filter `np.isfinite()` if you just use `np.nanmedian()` in the next line.

dmggh: Right, and it simplifies the code.

jwst/ami/ami_analyze.py
```diff
@@ -10,9 +10,9 @@
 from .nrm_model import NrmModel
 from . import webb_psf
 from . import leastsqnrm
+from . import utils
 
 log = logging.getLogger(__name__)
-log.setLevel(logging.DEBUG)
```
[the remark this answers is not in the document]
dmggh: Inadvertent removal, I've reinstated it.

jwst/ami/ami_analyze.py
```diff
@@ -76,6 +76,15 @@ def apply_LG(input_model, filter_model, oversample, rotation):
...
     input_data = input_model.data.astype(np.float64)
+    input_dq = input_model.dq
+
+    datamodel_img_model = datamodels.ImageModel(data=input_data, dq=input_dq)
+    box_size = 4
+
+    new_img_model = utils.img_median_replace(datamodel_img_model, box_size)
+
+    input_data = new_img_model.data.copy()
+    input_model.data = input_data.astype(np.float64)
```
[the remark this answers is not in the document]
dmggh: I think it does go away, but explicitly deleted them for clarity.

jwst/ami/utils.py
```diff
@@ -285,3 +288,99 @@ def findmax(mag, vals, mid=1.0):
...
+    # Extract the region of interest for the data
+    try:
+        data_array = input_array[xc - hbox:xc + hbox, yc - hbox: yc + hbox]
+        dq_array = input_dq_array[xc - hbox:xc + hbox, yc - hbox: yc + hbox]
+    except IndexError:
+        # If the box is outside the data return 0
+        log.warning('Box for median filter is outside the data.')
+        return 0.
+
+    filtered_array = data_array[dq_array != dqflags.pixel['DO_NOT_USE']]
```
hbushouse, line 332: Hmmm, not exactly sure how equality comparisons turn out when we're interested in the settings of individual bits, but I'm worried that simply using `!=` will only exclude pixels that have a DQ value exactly equal to `DO_NOT_USE`. But in most cases there's also another informational DQ value set along with `DO_NOT_USE`, in order to indicate why that pixel is bad. So will this statement correctly exclude pixels that have multiple flags set along with `DO_NOT_USE`?  Might need to use one of the bitwise comparators in some way.

hbushouse: Perhaps something like:
```
goodpixels = np.where(np.bitwise_and(dq, dqflags.pixel['DO_NOT_USE']) == 0)
```

dmggh: The bitwise check does it, irrespective of other DQ bits.

PR #5133, opened 2020-07-08
JP-1555: Move NIRSpec wavecorr to separate step

### LINKED ISSUE #5102: Create a calspec2 step for the NIRSpec wavecorr operation
_Issue [JP-1555](https://jira.stsci.edu/browse/JP-1555) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The NIRSpec wavelength correction for sources that are off-center (in the dispersion direction) within a Fixed Slit or a MOS slitlet is currently embedded with the calwebb_spec2 extract_2d step. In order to more easily implement the updates to NIRSpec master background subtraction outlined in JP-1213 and JP-1493, it's necessary to have that correction applied in a separate step, so that it is more controllable. Create a new step in the calwebb_spec3 flow, immediately following extract_2d, that is only applied to NIRSpec FS (NRS_FIXEDSLIT, NRS_BRIGHTOBJ) and MOS (NRS_MSASPEC) exposures that updates the 2D wavelength array created in the extract_2d step.

hbushouse:
Removed the various functions from the `extract_2d/nirspec.py` module that were used to compute the wavelength correction for off-center NIRSpec MOS and FS sources and put them into their own `wavecorr` step. Updated the `calwebb_spec2` pipeline to insert `wavecorr` into the flow, after `extract_2d`. Also rearranged the `calwebb_spec2` flow so that the `srctype` step gets called immediately after `extract_2d` for NIRSpec MOS and FS, so that the new `wavecorr` step can use the source type values. This allowed the removal of the duplication of MOS source type determination (based on `stellarity`) that had been in the `extract_2d` routines.

In the original `extract_2d` versions of the wavecorr routines, the Slit object that's in use during extraction of MOS spectra was used to get the `shutter_id` that's needed in one of the lower-level wavecorr routines. The Slit objects are no longer available when the data get to the new `wavecorr` step, but most of the meta data originally contained in the Slit object is stored as attributes in the `SlitModel` - all except for `shutter_id`. So `shutter_id` is now also saved to each MOS `SlitModel`, so that it's available for use in the `wavecorr` step.

This is all part of the infrastructure changes needed to support the corrections to NIRSpec master background subtraction.

Fixes #5102 / [JP-1555](https://jira.stsci.edu/browse/JP-1555)


hbushouse:
@nden OK, I think this is complete now and ready for a review. I've tested on several modes (NIRSpec FS, MOS, IFU) that get wavecorr applied and a few that don't (MIRI LRS slit and slitless). Seems to be working with no ill effects.

jwst/wavecorr/wavecorr_step.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    This step applies wavelength offsets to off-center NIRSpec sources.
+    """
+
+    spec = """
+    """
+
+    reference_file_types = ['wavecorr']
+
+    def process(self, input, *args, **kwargs):
```
stscieisenhamer, line 20: Two point: `process` never gets `kwargs` and the specification should be removed.

Also, unless `process` is intended to get a list, it should only define the argument(s) it needs. Presuming if this case is `input`, then it should be just `input`.

Also, as has been pointed out by others, `input` is a Python builtin function name and we really should start avoiding using that as a variable.

CHANGES.rst
```diff
@@ -28,68 +33,65 @@ barshadow
...
+- Trim MT_RA, MT_DEC keyword comments to fit within FITS record. [#4994]
 
 - Add enum list and default value of 'NONE' for ``meta.instrument.lamp_mode`` [#5022]
 
 - Add TIMEUNIT keyword to schemas. [#5109]
 
-- Split ``pathloss`` object into ``pathloss_ps`` and ``pathloss_un``. [#5112]
+- Split ``pathloss`` object into ``pathloss_ps`` and ``pathloss_un`` in schemas. [#5112]
 
 - Added "PERSISTENCE" DQ flag definition. [#5137]
```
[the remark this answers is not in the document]
hbushouse: Is there a preferred convention for this? Or is it important enough to even worry about?

jwst/wavecorr/wavecorr_step.py
hbushouse: This was a copy-n-paste from the previous incarnation of the function that lived in the `extract_2d` step, so I'll update as suggested.

hbushouse: Fixed in latest commit

PR #5134, opened 2020-07-08
cube_build - change default weighting back to 'msm'

jemorrison:
Changed default weighting from 'emsm' back to 'msm', at least temporarily until NIRSpec cubepars ref file is updated to contain params for 'emsm' mode. Also added checks for invalid/non-existent params in ref file.


jemorrison:
Ok I believe I got all the errors running the code taken care of by changing back to msm.  I did find one error I needed to fix with how outlier detection using the single IFU cubes- so that was good. 
Now there should just be truth file differences.

PR #5140, opened 2020-07-10
JP-1431: add parameter extract_height to wfss extraction

### LINKED ISSUE #4896: WFSS extraction size for point sources needs to be a fixed value that the user can set
_Issue [JP-1431](https://jira.stsci.edu/browse/JP-1431) was created on JIRA by [Kevin Volk](https://jira.stsci.edu/secure/ViewProfile.jspa?name=kvolk):_

Currently the extraction box size in the WFSS case is determined by the segmentation map width as projected in the cross-dispersion direction.  For point sources this is not what is required.  Any object identified as a point source or likely point source from the source catalogue should have a fixed width extraction box in pixels.  This width needs to be a parameter for the step so that the user can adjust it as needed.  The width will need to vary with the filter for NIRISS because of the change in PSF FWHM between the F090W and F200W filters, and so it will need to be set in a CRDS parameter file for the step eventually, but for now it is sufficient if the user can specify the width.  One way to specify the width is to have 2N+1 pixels total, N pixels on either side of the trace peak position.  Currently the aperture corrections for WFSS are defined in terms of this type of width with N varying from 3 to 10 (I believe).  The APCORR file can be extended to larger N values if needed, but since the maximum FWHM estimate is about 1.3 pixels at 2.2 microns an N value that is 5 or 6 times larger than the FWHM should be sufficient.

The issue with the current method is that the extraction width varies widely from one point source to another depending on (a) the brightness of the point source and (b) the contrast between the point source and any background emission that may be present.  This is not a desired feature of the extractions.  The bright point sources should, at least ideally, use the same extraction width as the fainter ones in a given filter.  We want the user to be able to define the value if they wish, since they may want smaller 2-D cut-outs for more crowded fields or larger ones for more sparse fields.

Another problem with the current method is that we have no indication that the PSFs are significantly elliptical, but the segmentation map does not necessarily project the same way onto the two dispersion directions for the GR150R and GR150C grisms.  Having the same extraction cross-width in both grisms is better than having a large variation, because the spectra should then come out with similar total signals rather than having significantly different signals.  While in principle the APCORR values should correct for this, the S/N of the spectra may not be that similar if the extraction widths are very different.

nden:
Fixes #4896 / [JP-1431](https://jira.stsci.edu/browse/JP-1431)


nden:
@hbushouse Is the `extraction_height` parameter supposed to be a full or half height?

hbushouse:
JP-1431 requests that the extraction parameter be a half-height, such that param=N results in a total height of 2N+1.

jwst/assign_wcs/util.py
```diff
@@ -506,7 +506,8 @@ def get_object_info(catalog_name=None):
 def create_grism_bbox(input_model,
                       reference_files,
                       mmag_extract=99.0,
-                      extract_orders=None):
+                      extract_orders=None,
+                      extract_height=None):
```
jdavies-st, line 510: If the keyword is `extract_height` it would be clearer if it was the full height of the extraction.  And this is already what the current keyword is.

https://github.com/spacetelescope/jwst/blob/ff9ea4017ebdc62309cf97661a5093f72c402722/jwst/extract_2d/extract_2d.py#L37-L40

If it is to be the half height, then `extraction_half_height` would be clearer to users I think.

jwst/assign_wcs/util.py
```diff
@@ -562,8 +563,17 @@ def create_grism_bbox(input_model,
     instr_name = input_model.meta.instrument.name
     if instr_name == "NIRCAM":
         filter_name = input_model.meta.instrument.filter
+        if input_model.meta.instrument.pupil.endswith('R'):
+            # _extraction_height_direction is used if extract_height is not None
+            _extract_height_direction = 'y'
+        else:
+            _extract_height_direction = 'x'
```
jdavies-st, line 570: The DISPAXIS or `meta.wcsinfo.dispersion_direction` keyword should give the dispersion direction without having to rely on filter name parsing.  Or move the filter name parsing to the dispersion direction function

https://github.com/spacetelescope/jwst/blob/ff9ea4017ebdc62309cf97661a5093f72c402722/jwst/lib/dispaxis.py#L6-L8

hbushouse:
> Does this also allow all the grism extractions to have their height set to a specific value, not just SOSS? That could be useful for the WFSS modes as well I suspect.

This PR **IS** for WFSS mode, not SOSS.

jdavies-st:
Doh!  _brainfart_

Then the `extract_height` keyword should be consistent with what it is for SOSS, i.e. the full height of the extraction box.  Having the same keyword mean 2 different things with two different datasets using the same code is not good.

nden:
@jdavies-st SOSS does not go through extract_2d. The parameter is for TSGRISM mode. But it's a good point. @hbushouse Would it be possible to change the TSGRISM parameter to "half height" as well or does the INS team wants is to be a full height. If so we'll have to have two parameters.

hbushouse:
I know the NIRCam TSO folks were specific about the placement and height of the extraction box that they wanted for TSGRISM. If the current height happens to be an odd number, then it too could be expressed as a half-height, like WFSS. But I'm guessing that the default values will be different for TSGRISM and WFSS, so it's not possible to combine into a single parameter, unless we create separate extract_2d parameter files and install them in CRDS. During processing, the appropriate one would be selected based on EXP_TYPE.

nden:
In this case I'll create a new parameter `extract_half_hight_grism` and say the rest is out of scope for this PR. If there are better ideas, please speak.

hbushouse:
That's fine, but I would use the term "wfss" in the parameter name, instead of the generic "grism", because TSGRISM and WFSS both use a grism. This parameter will be specific to WFSS only. Perhaps "wfss_extract_half_height".

nden:
I took the opportunity to refactor `create_grism_bbox` by splitting it into two functions (unfortunately still long), one that performs the computations and one that sets the input parameters. I also added another optional parameter, `wavelength_range`, to `create_grim_bbox` which gives more control over setting the extraction parameters.

docs/jwst/extract_2d/main.rst
```diff
@@ -78,6 +78,80 @@ of ``GrismObjects`` outside of these, the ``GrismObject`` itself can be imported
 The dispersion direction will be documented by copying keyword "DISPAXIS"
 (1 = horizontal, 2 = vertical) from the input file to the output cutout.
 
+Following are examples of how to customize the list of grism objects used in extract_2d.
+The input file must have a WCS assigned to it by running ``assign_wcs``. The default values
+for  wavelength range and the spectral orders are stored in the ``wavelengthrange``
+reference file which can be retrieved from CRDS. A user may supply a different
+wavelength range by passing `None` to ``reference_files``. In this case the spectral
+orders to be extracted and their corresponding wavelength range will be taken
+from the ``wavelength_range`` parameter.
```
hbushouse, line 87: How does a user specify both wavelength ranges AND spectral orders via a single parameter? Or is that explained later?

docs/jwst/extract_2d/main.rst
```diff
@@ -78,6 +78,80 @@ of ``GrismObjects`` outside of these, the ``GrismObject`` itself can be imported
...
+  >>> from jwst.extract_2d import Extract2dStep
+  >>> step = Extract2dStep()
+  >>> refs = {}
+  >>> for ref_type in step.reference_file_types:
+  ...     refs[ref_type] = step.get_reference_file(input_model, ref_type)
+  >>> print(refs)
+  {'wavelengthrange': '/crds/references/jwst/niriss/jwst_niriss_wavelengthrange_0002.asdf'}
+
+Create a list of grism objects for a specified spectral order with a limited
+minimum magnitude, and a specified half height of the extraction box in
```
hbushouse, line 107: Should note that the user-specified extraction height is only applied to grism objects with a source type of POINT (non-point sources should use the normal bb limits).

docs/jwst/extract_2d/main.rst
```diff
@@ -133,11 +207,13 @@ For NIRCam and NIRISS WFSS, the ``extract_2d`` step has three optional arguments
...
+For NIRCam TSGRISM, the ``extract_2d`` step has one optional argument:
 
 ``--extract_height``
   int. The cross-dispersion size (in units of pixels) to extract.
+
+
+For NIRCam and NIRISS WFSS mode, the ``extract_2d`` step has one optional argument:
+
+  ``--wfss_extract_half_height``
+    int. The cross-dispersion half size, in pixels.
```
hbushouse, line 219: Again, note that this only applies to sources tagged in the source catalog as point-like (based on the catalog entry "is_star" - boolean).

jwst/assign_wcs/util.py
```diff
@@ -521,19 +523,32 @@ def create_grism_bbox(input_model,
...
+    extract_orders : list, optional
         The list of orders to extract, if specified this will
-        override the orders listed in the wavelengthrange reference file
+        override the orders listed in the wavelengthrange reference file.
+        If ``None``, the default one in the wavelengthrange reference file is used.
+    wfss_extract_half_height : int, optional
+        Cross-dispersion extraction half height in pixels, WFSS mode.
+        Overwrites the computed extraction height in ``GrismObject.order_bounding.``
+        If ``None``, it's computed from the segementation map,
+        using the min and max wavelegnth for each of the orders that
```
hbushouse, line 540: typo in "wavelegnth"

jwst/assign_wcs/util.py
```diff
@@ -567,49 +580,55 @@ def create_grism_bbox(input_model,
     else:
         raise ValueError("create_grism_object works with NIRCAM and NIRISS WFSS exposures only.")
 
+    if reference_files is None:
+        # Get the list of extract_orders and lmin, lmax from wavelength_range.
+        if wavelength_range is None:
+            message = "If reference files are not supplied, ``wavelength_range`` must be provided."
+            raise TypeError(message)
+    else:
+        # Get the list of extract_orders and lmin, lmax from the ``wavelengthrange11 reference file.
```
hbushouse, line 589: typo "wavelengthrange11"

jwst/assign_wcs/util.py
```diff
@@ -567,49 +580,55 @@ def create_grism_bbox(input_model,
...
         log.error(err_text)
         raise ValueError(err_text)
 
-    log.info("Getting objects from {}".format(input_model.meta.source_catalog))
+    log.info(f"Getting objects from {input_model.meta.source_catalog}")
+
+    return _create_grism_bbox(input_model, mmag_extract, wfss_extract_half_height, wavelength_range)
+
+
+def _create_grism_bbox(input_model, mmag_extract=99.0,
```
hbushouse, line 618: Do we want mmag hardwired to 99 or allow a user value to be passed in?

jwst/extract_2d/extract_2d_step.py
```diff
@@ -16,7 +16,8 @@ class Extract2dStep(Step):
     spec = """
         slit_name = string(default=None)
         extract_orders = int_list(default=None)  # list of orders to extract
-        extract_height =  integer(default=None)  # extraction height in pixels
+        extract_height =  integer(default=None)  # extraction height in pixels, TSGRISM mode
```
hbushouse, line 19: Since we now have 2 extraction height params, maybe we should be explicit in the names for both of them which mode they apply to, such as `tsgrism_extract_height`.

jwst/extract_2d/grisms.py
```diff
@@ -299,6 +300,9 @@ def extract_grism_objects(input_model,
         Compute a wavelength array for the datamodel.  Computationally
         expensive, but saves doing it repeatedly later on in pipeline.
 
+    wfss_extract_half_heigh : int, (optional)
```
hbushouse, line 303: typo: missing the trailing "t" on height

jwst/assign_wcs/util.py
```diff
@@ -658,6 +677,17 @@ def create_grism_bbox(input_model,
                     xmax = int(np.max(xstack))
                     ymin = int(np.min(ystack))
                     ymax = int(np.max(ystack))
+                    if wfss_extract_half_height is not None:
```
hbushouse, line 680: Is the source catalog info for each source known at this point in the processing? This param value needs to be applied ONLY if the source being processed has `is_star = True` in the source catalog (only applied to point sources).

docs/jwst/extract_2d/main.rst
nden: The `wavelength_range` parameter is a dictionary of 
`{spectral_order: (lam_min, lam_max)}`

docs/jwst/extract_2d/main.rst
nden: will make that check and point this pout in the documentation

jwst/assign_wcs/util.py
nden: The default value is 99, a user can change it.

jwst/assign_wcs/util.py
nden: yes, the source catalog is available, will include this check

docs/jwst/extract_2d/main.rst
nden: The catalog I have lists `is_star` as float, not boolean, and all values are NaNs.
```
<Column name='is_star' dtype='float64' description='Flag indicating whether the source is a star' length=116>
nan
nan
nan
nan

```
Do you know if this column is computed in photutils or source_catalog?

nden: From `source_catalog`:
```
@lazyproperty
    def is_star(self):
        """
        Flag indicating whether the source is a star.

        2020.03.02: The JWST Photometry Working Group has not determined
        the criteria for this flag.
        """

        # TODO: need algorithm for this flag
        #is_star = np.random.randint(2, size=len(self.id))
        #return is_star.astype(bool)
        return self.null_column
```

This is where NaNs are returned.

nden:
I added the check for `is_star` with a made up value. This will need to be revised after the decision of how to populate the column. I also addressed the rest of the comments.

docs/jwst/extract_2d/main.rst
hbushouse: An algorithm was specified a month or two ago, but it hasn't been implemented yet. It will be Boolean when implemented, with a value of `True` for point sources.

jwst/assign_wcs/util.py
```diff
@@ -659,6 +680,21 @@ def create_grism_bbox(input_model,
                     ymin = int(np.min(ystack))
                     ymax = int(np.max(ystack))
 
+                    # The ``72`` criteria is a placeholder fof when there's a decision
+                    # on how to populate the ``is_star`` column in the catalog.
+                    # See ``is_star`` property in source_catalog.py
+                    if wfss_extract_half_height is not None and obj.is_star > 72:
```
hbushouse, line 686: Is there a way to test for `NaN` versus `True`, such that if `NaN`, don't apply, and if `True`, then do apply it.

docs/jwst/extract_2d/main.rst
```diff
@@ -133,11 +208,13 @@ For NIRCam and NIRISS WFSS, the ``extract_2d`` step has three optional arguments
...
 
 
-For NIRCam TSGRISM, the ``extract_2d`` step has two optional arguments:
-
-``--extract_orders``
-  list. The list of orders to extract. The default is taken from the ``wavelengthrange``
-  reference file.
+For NIRCam TSGRISM, the ``extract_2d`` step has one optional argument:
 
 ``--extract_height``
```
hbushouse, line 213: Should update this now to ``--tsgrism_extract_height``

jwst/assign_wcs/tests/data/step_SourceCatalogStep_cat.ecsv
```diff
@@ -23,6 +23,7 @@
 # - {name: orientation_sky, unit: deg, datatype: float64}
 # - {name: isophotal_abmag, datatype: float64}
 # - {name: isophotal_abmag_err, datatype: float32}
+# - {name: is_star, datatype: float64}
```
hbushouse, line 26: Perhaps you should modify the data file used in the test to have the `is_star` column be Boolean and modify the entries in the ecsv file itself to also be boolean for that column (perhaps with a mixture of True and False values).

jwst/assign_wcs/util.py
```diff
@@ -659,6 +680,21 @@ def create_grism_bbox(input_model,
                     ymin = int(np.min(ystack))
                     ymax = int(np.max(ystack))
 
+                    # The ``72`` criteria is a placeholder fof when there's a decision
```
hbushouse, line 683: You can get rid of this comment now, since you've already changed `obj.is_star` to be used as a boolean.

jwst/lib/catalog_utils.py
```diff
@@ -118,6 +119,8 @@ class SkyObject(namedtuple('SkyObject', ("id",
         Upper left corner of the minimum bounding box
     sky_bbox_ur : `~astropy.coordinates.SkyCoord`
         Upper right corner of the minimum bounding box
+    is_star : float
```
hbushouse, line 122: Need to change the data type to `bool`

PR #5149, opened 2020-07-17
First of several new modules for ami_analyze updates.

dmggh:
First pass of some of the new functions (in new modules) needed for ami_analyze updates, as part of what is needed for JP-1571. These modules will need adjustments as additional functions are updated and added, so should not be merged.  Documentation also needs completion.

### DIFF
--- a/jwst/ami/InstrumentData.py
+++ b/jwst/ami/InstrumentData.py
@@ -1,0 +1,894 @@
+#
+#  Module for defining data format, wavelength info, an mask geometry for these
+#   instruments: GPI NRM, NIRISS AMI, VISIR SAM
+#
+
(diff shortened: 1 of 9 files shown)


dmggh:
This new module (InstrumentData.py) needs testing with other updated modules in the step code, replacing hard-coded calls for file access with CRDS routine calls, and ''fixing" remaining flake8 warnings. Again, DO_NOT_MERGE.

dmggh:
With this commit of the updated analyticnrm2.py and the new nrm_core.py, all of the needed *initial* updates to the ami_analyze step code have been made. This code should not be merged, as I will now test a local copy of this updated step code.

PR #5150, opened 2020-07-17
Add bandit and pep517 build to Travis jobs

jdavies-st:
This pulls over the two jobs that used to run on ssbjenkins.stsci.edu

1. the security audit using `bandit`
2. the dist build.

This also removes the `Jenkinsfile` that used to run on ssbjenkins.  And ignores some warnings we're seeing from astropy in testing.


.travis.yml
```diff
@@ -62,6 +62,16 @@ jobs:
...
 
+    - name: Build distribution
+      env:
+        PIP_DEPENDENCIES=pep517
+        TEST_COMMAND="python -m pep517.build ."
+
+    - name: Security check
+      env:
+        PIP_DEPENDENCIES=".[test] bandit"
+        TEST_COMMAND="bandit --exclude 'jwst/*test*' -r jwst scripts"
```
nden, line 73: The `exclude` flag is not picked up, perhaps the name is different.
I've done the bandit configuration in gwcs/.bandit.yaml and it works.
I suggest the same configuration is done for jwst.
Most (if not all) of the assert errors I see are in the test files. When those are cleaned
we can see if there are anything remaining to fix.

jdavies-st:
Btw, even though this check was being run on ssbjenkins and the build was passing over there, it was actually failing:

https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fjwst/detail/master/678/pipeline/33#step-250-log-1053

It was just not reporting it as a failure.  Run under Travis, it is now correctly reporting it as a failure.

```
Code scanned:
	Total lines of code: 71864
	Total lines skipped (#nosec): 0

Run metrics:
	Total issues (by severity):
		Undefined: 0.0
		Low: 104.0
		Medium: 7.0
		High: 1.0
	Total issues (by confidence):
		Undefined: 0.0
		Low: 0.0
		Medium: 0.0
		High: 112.0
```

.bandit.yaml
```diff
@@ -0,0 +1,7 @@
+exclude_dirs:
+  - jwst/*test*
+  - jwst/**/*test*
+  - jwst/fits_generator
+
+skips:
+  - B322
```
jdavies-st, line 7: This check is not needed as we are python3-only, and `input` is safe in python3.

jwst/datamodels/make_header.py
```diff
@@ -323,7 +323,7 @@ def import_objects(package, line):
     local_variables = {}
     global_variables = {'__package__':package}
     try:
-        exec(line, global_variables, local_variables)
+        exec(line, global_variables, local_variables) # nosec
```
jdavies-st, line 326: This whole module does a bunch of source code parsing and executing imports found at the top of various pipeline modules.  It would be better to use `importlib` and `ast` for all of this if anyone want to take a crack at it in the future.

PR #5152, opened 2020-07-18
Fix extract2d-related bugs due to wavecorr

hbushouse:
PR #5133 left a few bugs behind in extract_2d-related operations for TSGRISM and WFSS modes, due to the fact that extract_2d no longer uses a list of multiple reference files, but instead just the single wavelengthrange ref file. Failed to update called functions that were expecting a list of ref files to be passed in. This fixes all related functions to only expect the wavelengthrange ref file.

### DIFF
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -504,7 +504,7 @@ def get_object_info(catalog_name=None):
 
 
 def create_grism_bbox(input_model,
-                      reference_file,
+                      reference_files,
                       mmag_extract=99.0,
                       extract_orders=None):
     """Create bounding boxes for each object in the catalog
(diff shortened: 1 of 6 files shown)


jwst/assign_wcs/util.py
```diff
@@ -521,8 +521,8 @@ def create_grism_bbox(input_model,
     ----------
     input_model : `jwst.datamodels.ImagingModel`
         Data model which holds the grism image
-    reference_files : dict
-        Dictionary of reference files
+    reference_file : str
```
nden, line 524: I have a slight preference of keeping this a dictionary. It's more general and in case we need another reference file in the future we won't have to change the interface again.

hbushouse: Yes, half way through the process of ripping out the "genericness" that thought occurred to me too. I guess I'll put it back in and make sure all the other lower-level functions still treat it appropriately.

hbushouse: No reference files at all are needed for the branch of extract_2d that handles NIRSpec (FS and MOS), because wavecorr is now in its own step. So the only functions that need any ref files included in their calls are the ones for tso and grism, which I'll restore.

PR #5157, opened 2020-07-21
JP-1150: extract_1d updates for using source position

### LINKED ISSUE #4289: NIRSpec extract_1d aperture should be centered at planned source location
Issue [JP-1150](https://jira.stsci.edu/browse/JP-1150) was created by James Muzerolle:

The extract_1d step as applied to NIRSpec FS or MOS data needs to be updated to account for the actual position of a target source within an aperture in each exposure.  Currently, a single extraction aperture location is taken from the extract_1d reference file, which cannot account for cases where a source is located at a different aperture position, such as in a dither or nod, or the MOS case generally.

As a default, the pipeline should calculate the aperture center location in the rectified 2D spectrum based on the planned position of the source that is passed by PPS.  In the FS case, the target sky coordinates and exposure pointing keywords should be sufficient to do this.  For MOS, the relative shutter position for each source given in the associated MSA metafile should be used.  The extract_1d reference file can then be used for the extraction aperture size.

For FS observations that include multiple slits (ALLSLITS subarray or FULL FRAME), only the primary target and its slit matters for source centering and extraction aperture size.  There is no way currently (or probably ever) for a user to plan a FS observation that centers more than one target in more than one slit at a time, and in any case the only metadata returned by PPS would be for the primary target.  Also, the nod/dither pattern is specific to the primary slit, so won't be optimized for any other slit if observed.  In general, the other slits will only be useful, if at all, as master backgrounds that a user could apply should the nod subtraction produce non-optimal results, or if nodding is not done in the observation (e.g., if the primary source is significantly extended).  Thus, I recommend that the pipeline automatically extract the other slits, if present, as backgrounds.

hbushouse:
First round of updates to the `extract_1d` step to use the source coordinates to compute the cross-dispersion center for the extraction aperture. A lot of the machinery was already in place, it just needed some logic updates here and there to make use of this capability at appropriate times and for appropriate exposure types (and not for others). Note that it's not obvious that the actual WCS backward transform that's used to compute x/y image location from source RA/Dec in NIRSpec MOS mode is actually working correctly, because the resulting pixel positions tend to be somewhat crazy (and negative a lot of times). Other modes, such as NIRSpec FS and MIRI LRS fixedslit seem to return good values from the backward transform.

Fixes #4289 / [JP-1150](https://jira.stsci.edu/browse/JP-1150)


hbushouse:
Note that a very large fraction of the code changes are purely cosmetic changes to log messages (making them shorter to fit on the screen, making the info in them more explicit, etc.)

hbushouse:
Latest commit changes the "apply_nod_offset" and "nod_correction" params to "use_source_posn" and "position_correction", because the use of the source position to center the extraction aperture does not just apply to cases where sources move due to nods or dithers. It also applies to the NIRSpec MOS case where not all sources are perfectly centered in their slitlets to begin with, regardless of any nodding/dithering that may be going on. So the more generic names are appropriate. Also changed all log messages to use this more generic wording (i.e. removing references to nod/dither).

jwst/extract_1d/extract.py
```diff
@@ -1208,17 +1210,20 @@ def get_target_coordinates(
         else:
             targ_ra = getattr(slit, 'source_ra', None)
             targ_dec = getattr(slit, 'source_dec', None)
+            if targ_ra is None or targ_dec is None:
```
hbushouse, line 1213: This is necessary for NIRSpec FS mode, where the input data are in the form of SlitModels, but there's no source RA/Dec info on a per-slit basis, because there's only 1 main target. So in this case it's necessary to revert to using the generic targ_ra/targ_dec values.

PR #5184, opened 2020-07-23
Fix bug in new WFSS extraction code

hbushouse:
Fix a bug in the new WFSS extraction region code in assign_wcs functions, which is referring to `model.meta.wcsinfo.dispaxis` instead of the correct `model.meta.wcsinfo.dispersion_direction`. Regression tests caught this error, causing all WFSS level-2b processing to fail.


nden:
There's another case in test_grisms.py

https://github.com/spacetelescope/jwst/pull/5140/files#diff-6591c2034d14eb7eba21f7a1829a35b9R396

My bad, sorry about this.

PR #5185, opened 2020-07-23
Wrap first angle at 360 in V23ToSky.cartesian2spherical

### LINKED ISSUE #5148: Negative RAs sometimes returned from assign_wcs

mcara:
Fixes #5148 / [JP-1585](https://jira.stsci.edu/browse/JP-1585) (hopefully - I have not yet got test files).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -180,6 +180,11 @@ stpipe
 
 - Remove further sloper references. [#4989]
 
+transforms
+----------
+
+- Wrap first spherical angle at 360 degrees in ``V23ToSky``. [#5185]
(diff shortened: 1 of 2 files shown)


mcara:
In the future, however, `assign_wcs.pointing.v23tosky` should transition to using [`gwcs.geometry.SphericalToCartesian` and `gwcs.geometry.CartesianToSpherical` models](https://github.com/spacetelescope/gwcs/blob/master/gwcs/geometry.py#L62-L208)

nden:
Testing with the failing data shows the issue has been fixed.

PR #5187, opened 2020-07-23
Make stable and unstable JenkinsfileRT consistent

jdavies-st:
The two files are the same now except for:

```
$ diff -y JenkinsfileRT JenkinsfileRT_dev --suppress-common-lines
jobconfig.enable_env_publication = true			      |	jobconfig.enable_env_publication = false
// Build and test with dependencies specified in requirements |	// Build and test with dependencies specified in requirements
bc0.name = 'stable-deps'				      |	bc0.name = 'unstable-deps'
bc0.pip_reqs_files = ['requirements-sdp.txt']		      |	bc0.pip_reqs_files = ['requirements-dev.txt']
    "pytest --cov-report=xml --cov -r sxf -n 30 --bigdata --s |	    "pytest -r sxf -n 30 --bigdata --slow \
bc1.name = 'macos-stable-deps'				      |	bc1.name = 'macos-unstable-deps'
```

### DIFF
--- a/JenkinsfileRT_dev
+++ b/JenkinsfileRT_dev
@@ -1,6 +1,17 @@
 // [skip ci] and [ci skip] have no effect here.
 if (utils.scm_checkout(['skip_disable':true])) return
 
+def PipInject(String reqs) {
+    def result = []
+    if (reqs.trim().isEmpty()) {
+        return result
+    }
+    for (req in reqs.split('\n')) {
+        result += "pip install $req"
+    }
+    return result
+}
+
 withCredentials([string(
     credentialsId: 'jwst-codecov',
     variable: 'codecov_token')]) {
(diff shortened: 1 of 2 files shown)


JenkinsfileRT_dev
```diff
@@ -1,24 +1,14 @@
 // [skip ci] and [ci skip] have no effect here.
 if (utils.scm_checkout(['skip_disable':true])) return
 
-def PipInject(String reqs) {
```
eslavich, line 4: This is a feature that I regularly use.  I don't know if it's worth keeping around just for me.

jdavies-st: Ah, you do use it?!  Ok, then we'll keep it in.

eslavich: This is typically how I test an asdf branch, just add a `git+` requirement to the input box on the parametrized build.  Saves me from having to create a jwst branch with an updated requirements-dev.txt

jdavies-st: I will add that capability to the normal nightly build too then.

PR #5196, opened 2020-07-26
JP-1585: do not allow negative sky values

### LINKED ISSUE #5148: Negative RAs sometimes returned from assign_wcs

nden:
Fixes #5148 
Alternative to #5195

This has the same effect as #5195 but does not require a special model.
If transforms are to be removed from jwst in the future we should go with this implementation.
There's a branch `move_transforms_out_of jwst` which replaces the `V23ToSky` transform with rotations from astropy.
Worked on this with @mcara.

Most of the failing regression tests pass, some fail with small differences.


nden:
Superseded by #5206

PR #5201, opened 2020-07-27
JP-1562 - incorrect surface brightness for IFU data in extract 1d

### LINKED ISSUE #5117: Incorrect surface brightness computation in Extract1D
_Issue [JP-1562](https://jira.stsci.edu/browse/JP-1562) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

When extract1d is run on an extended source image, the FITS binary table containing the extracted spectrum contains both a 'flux' and a 'surf_bright' vector.  The 'flux' vector in this case is simply the sum of all of the individual spaxels (converted from MJy/sr to MJy/pixel units) given in units of Jy.  The 'surf_bright' vector is the flux vector divided by the total cube area, given in units of MJy/sr.

The implementation of this is incorrect; based on the test data linked above it looks like it is simply dividing the flux vector by the total cube area, not the area of the cube in which the IFU footprint actually sits.  This is an important difference; because the IFU field of view can be (and generally is) tilted in the data cubes, it tends to be surrounded by lots of zeros that are outside the true footprint and should not be included in the area computation.  In this particular case for instance, the 3d cube has all spaxels with surface brightness about 50 MJy/sr around plane 400.  The surface brightness vector from extract1d instead measures 29 MJy/sr, which is incorrect by a factor corresponding to the fractional area of the actual IFU footprint within the data cube.

jemorrison:
@drlaw1558 
I think this is what we need for 'extract_ifu' still thinking about the 'image_extract_ifu'
Look it over

Fixes #5117 / [JP-1562](https://jira.stsci.edu/browse/JP-1562)

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -2626,7 +2626,6 @@ def do_extract1d(
     # Setup the output model
     output_model = datamodels.MultiSpecModel()
 
-    print('**************** in do_extract1d')
     if hasattr(input_temp, "int_times"):
         output_model.int_times = input_temp.int_times.copy()
 
@@ -2677,7 +2676,6 @@ def do_extract1d(
                 f"Turning on source position correction for exp_type = {exp_type}"
             )
 
-    print('was_source_model',was_source_model)
     # Handle inputs that contain one or more slit models
     if was_source_model or isinstance(input_model, datamodels.MultiSlitModel):
         if was_source_model:   # SourceContainer has a single list of SlitModels
@@ -3213,6 +3211,7 @@ def do_extract1d(
         or extract_ref_dict['need_to_set_to_complete']
     ):
         output_model.meta.cal_step.extract_1d = 'COMPLETE'
+
 
     return output_model
 
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -2609,7 +2609,6 @@ def do_extract1d(
         A new MultiSpecModel containing the extracted spectra.
     """
 
-    log.info('in do extract1d')
     extract_ref_dict = ref_dict_sanity_check(extract_ref_dict)
 
     if isinstance(input_model, datamodels.SourceModelContainer):
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -2609,6 +2609,7 @@ def do_extract1d(
         A new MultiSpecModel containing the extracted spectra.
     """
 
+    log.info('in do extract1d')
     extract_ref_dict = ref_dict_sanity_check(extract_ref_dict)
 
     if isinstance(input_model, datamodels.SourceModelContainer):
@@ -2625,6 +2626,7 @@ def do_extract1d(
     # Setup the output model
     output_model = datamodels.MultiSpecModel()
 
+    print('**************** in do_extract1d')
     if hasattr(input_temp, "int_times"):
         output_model.int_times = input_temp.int_times.copy()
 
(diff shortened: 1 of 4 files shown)


drlaw1558:
Remind me what the idea behind the 'image_extract_ifu' routine is?  Something to force extraction from a specific area corresponding to some input image/cube file?  I'm not aware of any cases in which this would be run, is the functionality there just in case it's useful in the future?

jemorrison:
Extract 1D for IFU data can be done two ways. The first method (default) uses the JSON reference file to set up the extraction region. The second method (I am still researching) but according to comments on ifu.image_extract_ifu:
"_One of the requirements for this step is that for an extended target,  the entire aperture is supposed to be extracted (with no background  subtraction).  It doesn't make any sense to use an image extract1d reference file to extract the entire aperture; a trivially simple JSON extract1d reference file would do.  Therefore, we assume that if the user specified a reference file in image format, the user actually wanted that extract1d reference file to be used, so we will ignore the requirement in this case."_
@drlaw1558 @jmuzerolle can either of you think of when the second image_extract_ifu routine would be used and the user would provide a reference file in image form. Howard was thinking this was for NIRSS SOSS mode - but this routine is only called if the data has a datamodel of IFUCubeModel.

jemorrison:
@drlaw1558 @jmuzerolle 
Ok I found a test that is using this image_extract_ifu routine that might better explain what the purpose of this routine is.
See extract_1d/tests/test_ifu_image.py

drlaw1558:
Looks like there are still a few bugs in the area normalization.  EXTENDED sources look ok in an intial test, but for POINT sources the normalization of BACKGROUND is still incorrect.  Looking at the code, I believe this is because the actual annulus area is computed around line 421 of ifu.py, but it's never passed back out of this routine.  When the normalization actually happens around line 110 of ifu.py it uses npixels_temp, which is the area of the circular flux aperture, not the area of the background annulus.

There also seems to be a glitch when the loop over wavelengths (starting around line 397) hits the first wavelength in which there are zero valid pixels between the two channels of data.  This triggers the else clause that 'Background annulus has no area so background subtraction will be turned off', which turns off subtract_background and effectively breaks us out of the loop so that the background is never computed for the second channel worth of data at longer wavelengths. 

Relating to the Cal WG discussion earlier today on JP-1563 I think we want to be sure while we're tinkering with this code that it always computes the background regardless of whether subtract_background is True.

hbushouse:
Regarding the change to always compute the background, but only subtract it when requested, I think it would be cleaner for tracking purposes to only fix the reported bugs in this PR and then make the change to always compute background in the work that will be done (separately) to implement the broader changes discussed this morning.

drlaw1558:
@hbushouse Sure; my point was that the not-always-computing-background quality check triggers are what's breaking us out of the computation loop too early in the above fix to the area normalization.

jemorrison:
@drlaw1558 
Line ~447 temp_flux[k] = temp_flux[k] - background[k] * normalization
where normalization:
if annulus_area > 0.:
                normalization = aperture_area / annulus_area

Do you think we need to pass back the number of pixels in the annulus and use that in 
background /= npixels_temp (line 110) instead of npixels_temp  or should we pass back the 'normalization' factor

drlaw1558:
My feeling is that it's cleaner to pass back the number of pixels in the annulus instead of the 'normalization' factor, so extract_ifu would be returning the source flux, the background flux, and the number of pixels in each.  It's the annulus area that we care about in order to get a background in surface brightness units, it shouldn't need to be tied to what's going on with the science aperture.

jemorrison:
@drlaw1558  I believe these updates should fix the problems  you noted in the last review. Could you run this branch and let me know what you think

hbushouse:
@jemorrison the only thing we're missing is a change log entry. If you can add that, I'll approve and merge.

jwst/extract_1d/extract.py
```diff
@@ -2609,6 +2609,7 @@ def do_extract1d(
         A new MultiSpecModel containing the extracted spectra.
     """
 
+    log.info('in do extract1d')
```
hbushouse, line 2612: We should remove this from the production code.

PR #5202, opened 2020-07-28
JP-1546: Updates for MIRI MRS TSO processing

### LINKED ISSUE #5087: Implement minimal support for MIRI MRS TSO mode
_Issue [JP-1546](https://jira.stsci.edu/browse/JP-1546) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

Time series is available as a special requirement for MIRI MRS observations in the cycle 1 APT. A plan for minimal calibration pipeline support is being created by the MIRI team, as discussed in <https://jira.stsci.edu/browse/JSOCINT-421.> Cal pipeline updates are needed to implement changes to MRS calibration for TSO mode.

Also relates to <https://jtar.stsci.edu/browse/JWSTSOCINTDR-5363>

hbushouse:
As part of the minimal processing support for MIRI MRS exposures taken in TSO mode, the calspec2 fringe step has been requested to be applied, as it is for non-TSO MRS data. This updates the fringe step code to allow for both 2D and 3D inputs. In the case of 3D input, the 2D fringe correction is simply broadcast to all planes of the 3D cube.

Fixes #5087 / [JP-1546](https://jira.stsci.edu/browse/JP-1546)

### DIFF
--- a/jwst/fringe/fringe.py
+++ b/jwst/fringe/fringe.py
@@ -61,14 +61,12 @@ def apply_fringe(input_model, fringe):
     # Initialize the output model as a copy of the input
     output_model = input_model.copy()
 
-    fringe_data = fringe.data
-    # fringe_dq = fringe.dq
-
     # Fringe-correct data and error arrays, applying correction only
-    #    to pixels having good reference values
-    good_pix = np.isfinite(fringe_data)
-    output_model.data[good_pix] /= fringe_data[good_pix]
-    output_model.err[good_pix] /= fringe_data[good_pix]
+    # to pixels having good calibration values
+    fringe_data = fringe.data
+    fringe_data[np.isnan(fringe_data)] = 1.0
+    output_model.data /= fringe_data
+    output_model.err /= fringe_data
 
 ### 05/22/14: For now, commenting out the following updating of the output
 ###    DQ based on the DQ of the reference file. This is done now because the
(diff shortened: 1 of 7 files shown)


hbushouse:
> Nailed the association modification!

Yep, most of the time involved simply studying all the existing rules till I found a similar use of a Constraint() as what I needed. After that it was just copy-n-paste. Oh, and testing of course!

jwst/fringe/fringe.py
```diff
@@ -77,12 +75,11 @@ def apply_fringe(input_model, fringe):
 ###    will be logged for now. This behavior may be changed later.
 ###
     # set DQ flag for bad pixels in the fringe
-#       dq_mask = fringe_dq * 0
-#       dq_mask[np.where(fringe_dq != 0)] = dqflags.pixel['DEAD']
+#   dq_mask = fringe.dq * 0
```
jemorrison, line 78: I am just curious I would of thought that if the fringe_dq !=0 then the dq_mask would of been set to NON_SCIENCE rather than DEAD.

hbushouse: No idea. That's just a holdover from the original code in the step, along with the commentary that it's all deactivated for now, because the DQ values in the reference file don't actually correspond to bad pixels (or some such). You'll need to talk with the original developer of the step and ref file to sort it out. Now, who could that be ...

PR #5206, opened 2020-07-29
JP-1585: Wrap first angle at 360 in forward V2V3 to sky and at 180 for inverse.

### LINKED ISSUE #5148: Negative RAs sometimes returned from assign_wcs

mcara:
Fixes #5148 / [JP-1585](https://jira.stsci.edu/browse/JP-1585). Improves upon #5185 by restoring wrap angle for `V2` to 180 degrees so that inversion polynomials that take from detector to V2V3 frames now can work as designed.

Closes #5195
Closes #5185 

This PR replaces `V23ToSky` model with "standard" models from `astropy` and `gwcs`.

### DIFF
--- a/jwst/lib/set_telescope_pointing.py
+++ b/jwst/lib/set_telescope_pointing.py
@@ -532,11 +532,6 @@ def update_s_region(model, siaf):
     )
     # Execute IdealToV2V3, followed by V23ToSky
     from ..transforms.models import IdealToV2V3
-    v2_ref_deg = model.meta.wcsinfo.v2_ref / 3600  # in deg
-    v3_ref_deg = model.meta.wcsinfo.v3_ref / 3600  # in deg
-    roll_ref = model.meta.wcsinfo.roll_ref
-    ra_ref = model.meta.wcsinfo.ra_ref
-    dec_ref = model.meta.wcsinfo.dec_ref
     vparity = model.meta.wcsinfo.vparity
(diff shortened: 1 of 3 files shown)


hbushouse:
If this approach is approved, does it supersede both #5195 and #5196 ?

mcara:
Yes, it supercedes #5185 and #5195. It also removes the need for custom model `V23ToSky`. However, I did not remove `V23ToSky` because I am not sure how to do this safely (so that it works with old files) @nden

jwst/lib/set_telescope_pointing.py
```diff
@@ -547,15 +548,12 @@ def update_s_region(model, siaf):
     )
     v2, v3 = idltov23(xvert, yvert)  # in arcsec
 
-    # Convert to deg
-    v2 = v2 / 3600  # in deg
-    v3 = v3 / 3600  # in deg
     angles = [-v2_ref_deg, v3_ref_deg, -roll_ref, -dec_ref, ra_ref]
     axes = "zyxyz"
```
jdavies-st, line 552: These two lines need to be removed, as `angles` and `axes` are no longer used.

nden:
The failures in the code style check need to be fixed but otherwise looks good.
We can close delete now the branch move-transforms-out-of-jwst, the same changes are implemented in this PR.

mcara:
PEP8 fixed in the last commit (squashed onto previous one).

PR #5211, opened 2020-07-30
JP-1296: Updates for NINTS=1 coron exposures

### LINKED ISSUE #4566: Allow exposures with NINT=1 as input to coron3
Issue [JP-1296](https://jira.stsci.edu/browse/JP-1296) was created by Ben Sunnquist:

NIRCam coronagraphy exposures with 1 integration (NINT=1) only produce cal.fits images during calwebb_image2 processing (rather than calints.fits), but the next stage in the pipeline for coronagraphy exposures (calwebb_coron3) only accepts calints.fits files as inputs. Running cal.fits images through coron3 produces the attached error.

Coronagraphy exposures with NINT=1 are used in NIRCam Commissioning programs, and may be used during science operations as well.

hbushouse:
Several updates are contained here, in order to allow NINTS=1 exposures to be used in later stages of the pipeline for coronagraphy observations.

1) Updated the `ramp_fit` step so that it ALWAYS creates a "rateints" product, even when NINTS=1. In this case, the data is the same as the "rate" product, except for the fact that it's stored in a `CubeModel` with the 3rd dimension having length 1. This allows the "rateints" products to flow through `calwebb_image2`, producing "calints" products, which are then used as input to `calwebb_coron3`, with no changes needed to the ASN rules to use different products (e.g. "cal" instead of "calints").

2) Updated parts of the `calwebb_coron3` pipeline and the `outlier_detection` step to properly handle the `CubeModel`s that have NINTS=1. In this case, the `outlier_detection` step automatically gets skipped, due to lack of sufficient data. Various bits of logic in the pipeline and outlier step needed updating to handle all of this properly.

3) Updated the docs to indicate that the "rateints" product is no longer created only when NINTS>1.

Fixes #4566 / [JP-1296](https://jira.stsci.edu/browse/JP-1296)

### DIFF
--- a/jwst/model_blender/blendmeta.py
+++ b/jwst/model_blender/blendmeta.py
@@ -233,6 +233,9 @@ def get_blended_metadata(input_models, verbose=False):
     else:
         new_table = None
 
+    for model in input_models:
+        model.close()
+
     return new_meta, new_table
 
 
--- a/jwst/outlier_detection/outlier_detection_step.py
+++ b/jwst/outlier_detection/outlier_detection_step.py
@@ -37,7 +37,7 @@ class OutlierDetectionStep(Step):
 
     Parameters
     -----------
-    input : asn file or ModelContainer
+    user_input : asn file or ModelContainer
         Single filename association table, or a datamodels.ModelContainer.
 
     """
@@ -64,9 +64,9 @@ class OutlierDetectionStep(Step):
         search_output_file = boolean(default=False)
     """
 
-    def process(self, input):
+    def process(self, user_input):
         """Perform outlier detection processing on input data."""
-        with datamodels.open(input) as input_models:
+        with datamodels.open(user_input) as input_models:
             self.input_models = input_models
             if not isinstance(self.input_models, datamodels.ModelContainer):
                 self.input_container = False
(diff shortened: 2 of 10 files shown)


jwst/model_blender/blendmeta.py
```diff
@@ -233,6 +233,9 @@ def get_blended_metadata(input_models, verbose=False):
     else:
         new_table = None
 
+    for model in input_models:
+        model.close()
+
```
hbushouse, line 238: Added this to prevent warnings about unclosed files at the end of the pipeline processing.

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -37,7 +37,7 @@ class OutlierDetectionStep(Step):
 
     Parameters
     -----------
-    input : asn file or ModelContainer
+    user_input : asn file or ModelContainer
```
hbushouse, line 40: Trying to get away from using `input`

hbushouse:
Updates to datamodels/util.py are just changes to formatting of log messages and using `log.warning()` instead of `warnings.warn()`. Nothing else substantive to actual processing.

PR #5214, opened 2020-07-31
Spec2 step refactor

stscieisenhamer:
This is a basic refactor of `Spec2Pipeline`. There is a bunch of internal logic to decide on step execution intermingled with the actual execution of the steps.  Also, order of steps is dependent on the science being processed, again, intermingled with the aforementioned logic and flow of the main processing method. All this intermingling makes for hard-to-read code and adding any addition steps difficult.

The refactor separates out the execution decision logic from the actual step execution. This allows the different required orders of step execution to also be consolidated into specific methods. This will make implementation of the NIRSpec MOS Master Background Subtraction easier to implement.

Another potential advantage is that steps, for the most part, no longer simply "disappear". The execution decision logic uses the `skip` attribute to skip unwanted steps, providing a log trail (in debugging mode) as to why the step was skipped.

Finally, ideally the execution decision logic should really be based on by the CRDS step parameter reference files. Once this are fully populated, this logic can be completely removed from the pipeline.


jwst/pipeline/calwebb_spec2.py
```diff
@@ -27,6 +27,9 @@
 
 __all__ = ['Spec2Pipeline']
 
+# Classify various exposure types.
+WFSS_TYPES = ["NIS_WFSS", "NRC_WFSS"]
```
hbushouse, line 31: Might want to add "NRC_GRISM" to this list just for backwards compatibility (it's the old form of NRC_WFSS).

jwst/pipeline/calwebb_spec2.py
```diff
@@ -201,80 +206,23 @@ def process_exposure_product(
...
+        # Steps whose order is the same for all types of input.
+        input = self.bkg_subtract(input, members_by_type['background'])
+        input = self.imprint_subtract(input, members_by_type['imprint'])
+        input = self.msa_flagging(input)
 
         # The order of the next few steps is tricky, depending on mode:
         # WFSS/Grism data need flat_field before extract_2d, but other modes
         # need extract_2d first. Furthermore, NIRSpec MOS and FS need
         # srctype and wavecorr before flat_field.
         if exp_type in ['NRC_WFSS', 'NIS_WFSS', 'NRC_TSGRISM']:
```
hbushouse, line 218: Maybe use another macro here that includes all modes that use a grism, e.g. GRISM_TYPES? It consists of WFSS_TYPES + 'NRC_TSGRISM'.

jwst/pipeline/calwebb_spec2.py
```diff
@@ -201,80 +206,23 @@ def process_exposure_product(
...
+            input = self._process_grism(input)
             # Apply flat-field correction
-            input = self.flat_field(input)
-            input = self.extract_2d(input)
-            input = self.srctype(input)
+        elif exp_type in ['NRS_FIXEDSLIT', 'NRS_BRIGHTOBJ', 'NRS_MSASPEC', 'NRS_LAMP']:
+            input = self._process_nirspec(input)
```
hbushouse, line 222: Technically this isn't a generic nirspec routine, because it doesn't apply to NRS_IFU. So maybe a somewhat less generic name for the method (_process_nirspec_slits?)

jwst/pipeline/calwebb_spec2.py
```diff
@@ -336,3 +284,115 @@ def process_exposure_product(
         )
 
         return result
+
+    def _step_verification(self, exp_type, members_by_type, multi_int):
+        """Verify whether requested steps can operated on the given data
```
hbushouse, line 289: typo: "operated" should not be past tense

jwst/pipeline/calwebb_spec2.py
```diff
@@ -336,3 +284,115 @@ def process_exposure_product(
...
+
+        Thought ideally this would all be controlled through the pipeline
```
hbushouse, line 291: typo: "Thought" should be "Though"

jwst/pipeline/calwebb_spec2.py
```diff
@@ -336,3 +284,115 @@ def process_exposure_product(
...
+                # Setup for saving
+                self.bkg_subtract.suffix = 'bsub'
+                if multi_int:
+                    self.bkg_subtract.suffix = 'bsubints'
+
+                # Backwards compatibility
+                if self.save_bsub:
+                    self.bkg_subtract.save_results = True
+            else:
+                self.log.debug('Science does not allow direct background subtraction. Skipping "bkg_subtract".')
```
hbushouse, line 315: Should this be `info` level, rather than `debug`, or would that clutter up the log for too many modes that don't allow for direct background subtraction?

jwst/pipeline/calwebb_spec2.py
```diff
@@ -336,3 +284,115 @@ def process_exposure_product(
...
+                    self.log.warning('Wrong number of imprint members')
+                members_by_type['imprint'] = imprint[0]
+            else:
+                self.log.debug('Science does not allow imprint processing. Skipping "imprint_subtraction".')
+                self.imprint_subtract.skip = True
+
+        # Check for NIRSpec MSA bad shutter flagging.
+        if not self.msa_flagging.skip and exp_type not in ['NRS_MSASPEC', 'NRS_IFU']:
+            self.log.debug('Science does not allow MSA flagging. Skipping "msa_flagging".')
+            self.msa_flagging.skip = True
```
hbushouse, line 333: So what happens if someone creates an off-line ASN table that has multiple products defined, just as an easy way to process a bunch of exposures all at once, in a loop, and there's a mixture of exposures that allow for and don't allow for a given step like msa_flagging? Will setting the skip param to `True` for the first exposure in the list then cause it to get skipped for all remaining exposures or will the vetting be started over from scratch for each exposure?

Encountered this problem in the `calwebb_coron3` pipeline with the `outlier_detection` step. Once the `skip` param is set to `True` for the first exposure, it doesn't even look to see if it's possible to apply to subsequent exposures.

jwst/pipeline/calwebb_spec2.py
stscirij: Also, NRS_LAMP exposures are going to be covered by the pars reference files, and the set of steps to be run differs from that for regular NIRSPEC science data

jwst/pipeline/calwebb_spec2.py
```diff
@@ -336,3 +284,115 @@ def process_exposure_product(
...
+
+        # Check for straylight correction for MIRI MRS.
+        if not self.straylight.skip and exp_type != 'MIR_MRS':
+            self.log.debug('Science does not allow stray light correction. Skipping "straylight".')
+            self.straylight.skip = True
+
+        # Apply the fringe correction for MIRI MRS
+        if not self.fringe.skip and exp_type != 'MIR_MRS':
+            self.log.debug('Sicnece does not allow fringe correction. Skipping "fringe".')
```
stscirij, line 342: Typo: Sicnece -> Science

jwst/pipeline/calwebb_spec2.py
```diff
@@ -336,3 +284,115 @@ def process_exposure_product(
...
+        # Apply pathloss correction to NIRSpec and NIRISS SOSS exposures
+        if not self.pathloss.skip and exp_type not in ['NRS_FIXEDSLIT', 'NRS_MSASPEC', 'NRS_IFU', 'NIS_SOSS']:
+            self.log.debug('Science does not allow pathloss correction. Skipping "pathloss".')
+            self.pathloss.skip = True
+
+        # Apply barshadow correction to NIRSPEC MSA exposures
+        if not self.barshadow.skip and exp_type != 'NRS_MSASPEC':
+            self.log.debug('Science does not allow barshadow correction. Skipping "barshadow".')
+            self.barshadow.skip = True
+
```
stscirij, line 354: Would it be clearer to say "Science *data* does not allow xxx correction"?

jwst/pipeline/calwebb_spec2.py
stscieisenhamer: Actually not sure. Guess I did debug just to keep the noise down, but in the goal of making things more explicit, and this does seem rather important info, bumping up to info sounds reasonable.

hbushouse: Now that I see that there are so many of these, for all the various steps that do not apply to a given exp_type, it probably would be log overload if these were info level. So just leave them at debug.

jwst/pipeline/calwebb_spec2.py
stscieisenhamer: Yes, the described scenario is exactly what would happen. This is a general problem rooted in the instantiate-once/use-many-times model of Steps. For this particular instance, IMHO, the `spec2` looping over products was a mistake from the start. There are a number of approaches on this, see JP-1580, and it is probably best to deal with separately.

stscieisenhamer:
> This looks like an upgrade in clarity, although there is still some exp_type level branching at the end 

Good point! Tunnel-vision didn't even get that far. Should incorporate that also.

jwst/pipeline/calwebb_spec2.py
stscieisenhamer: @stscirij Though the set of steps are different, is the order of those steps the same as what is done in `_process_nirspec`? The key between these methods is that ordering is different. If steps are skipped, that is OK and is expected.

stscieisenhamer:
> > This looks like an upgrade in clarity, although there is still some exp_type level branching at the end
> 
> Good point! Tunnel-vision didn't even get that far. Should incorporate that also.

Actually, this set of conditions is dependent on previous results. This most likely can be incorporated into a process method, but it is not clear to me yet how. Will leave as is.

PR #5217, opened 2020-08-03
JP-1489: Allow calc_gwcs_pixmap to compute NaNs for drizzle

### LINKED ISSUE #5013: resample_spec results for NIRSpec MOS data still contain artifacts
_Issue [JP-1489](https://jira.stsci.edu/browse/JP-1489) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

MOS s2d products generated by resample_spec are still showing spurious artifacts and/or incomplete sampling.  I tested two different data sets: an internal lamp exposure processed through calwebb_spec2 (the MOS_CV3_products directory indicated above), and a set of simulated exposures containing point sources processed through calwebb_spec3 (the Sim_products directory).  Both show sampling artifacts of various kinds.

The first attachment shows the result for slit 53 from the MOS_CV3 exposure (NRS1 detector only), with the input unrectified 2D spectrum (calibrated, with flat_field skipped) at the top and the s2d product at the bottom.  There are two artifacts of note here: (1) there is a spurious signal at the center of the trace (y=2) extending from roughly x~120 to 280, where there is an abrupt discontinuity; (2) rows 0, and to a lesser extent, 1, contain signal that appears to be completely unrelated to the original spectrum.  Perhaps there is some sort of error in how the input flux is redistributed onto the output grid?  Artifact (1) looks suspiciously like the extra signal seen in the unrectified spectrum at y~9, which is overlapping flux from a failed open shutter.  Artifacts similar to (2) appear in almost all of the spectra in this exposure.

The second attachment shows the result from the same exposure, but for slit 10, NRS2 detector.  Artifact (2) can be seen in row y=0.  In addition, there are blocky areas with inexplicably large flux in rows y=2, 3, and 4.

The next three attachments show results from the simulated data set, which contained spectra of dozens of separate sources.  In this case, the level 3 products include combination of three exposures at three nod positions.  The first two plots show source 30263 - one with unrectified spectra for one of the three nods, both detectors, and the other with the combined s2d product.  The s2d product exhibits three unexpected artifacts: (1) the distribution of flux throughout is unexpectedly patchy, with brighter blocks (e.g., between x~380-500, 550-750, 880-980, 1180-1250); (2) there is spurious signal in rows y=1-4 between x~0-450 (perhaps similar to artifact (2) from the CV3 data); (3) there is unexpected signal in the gap between the detectors (x~980) in row y=4 (there should be zero signal in all rows here). Artifact (1) appears in most of the source spectra, to varying degrees of blockiness.  The last attachment, the s2d product for source 9427, shows a particularly bad example of this, and looks almost as if parts of the three nods were not lined up properly.

As a side note, I've checked resample_spec results for FS data, and they are generally much better. There are low-level blocky features in the narrowest slit, S1600A1, but the rest of the (longer) slits look excellent.  So, there's some evidence that the quality of the resampled output depends on the size of the input spectra in the cross-dispersion dimension.  Perhaps this has something to do with the limited range of sky coordinates?

Note this ticket is a follow-up of two previous tickets on separate aspects of the resampling (JP-369 and JP-936), and is intended to supersede them both.

This ticket has critical priority for both Commissioning and science data because the quality of 1D extracted spectra depend on the fidelity of the 2D resampling, and there is no alternative method of extraction.

jdavies-st:
Replacing NaNs with -1 in the pixel mapping for `drizzle` was causing these artifacts.  When this was originally implemented, `drizzle` pixmap barfed when there were NaNs in the pixmap.

And of course the reason this is showing up only in NIRSpec, as that NIRSpec is the only instrument where the WCS produces NaNs for pixels in the WCS bounding box.  This is why we did not see it elsewhere.

See below for what the fix does for single-resampled `s2d` files in `calwebb_spec2` for NIRSpec MOS data:

Screen Shot 2020-08-05 at 11 23 08 AM

Resolves #5013 / [JP-1489](https://jira.stsci.edu/browse/JP-1489)

### DIFF
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -556,13 +556,16 @@ def gwcs_blot(median_model, blot_img, interp='poly5', sinscl=1.0):
     log.debug("Pixmap shape: {}".format(pixmap[:, :, 0].shape))
     log.debug("Sci shape: {}".format(blot_img.data.shape))
 
-    # median_model_pscale = median_model.meta.wcsinfo.cdelt1
-    # blot_pscale = blot_img.meta.wcsinfo.cdelt1
-
     pix_ratio = 1
     log.info('Blotting {} <-- {}'.format(blot_img.data.shape, median_model.data.shape))
 
     outsci = np.zeros(blot_img.shape, dtype=np.float32)
+
+    # Currently tblot cannot handle nans in the pixmap, so we need to give some
+    # other value.  -1 is not optimal and may have side effects.  But this is
+    # what we've been doing up until now, so more investigation is needed
+    # before a change is made.  Preferably, fix tblot in drizzle.
+    pixmap[np.isnan(pixmap)] = -1
     tblot(median_model.data, pixmap, outsci, scale=pix_ratio, kscale=1.0,
                    interp=interp, exptime=1.0, misval=0.0, sinscl=sinscl)
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -193,6 +193,11 @@ ramp_fitting
 
 - Update to always create the rateints product, even when NINTS=1. [#5211]
 
+resample_spec
+-------------
+
+- Fix artifacts in resampled NIRSpec slit data caused by NaNs in the WCS [#5217]
+
 source_catalog
 --------------
 
--- a/jwst/resample/resample_utils.py
+++ b/jwst/resample/resample_utils.py
@@ -76,7 +76,6 @@ def calc_gwcs_pixmap(in_wcs, out_wcs, shape=None):
 
     grid = wcstools.grid_from_bounding_box(bb)
     pixmap = np.dstack(reproject(in_wcs, out_wcs)(grid[0], grid[1]))
-    pixmap[np.isnan(pixmap)] = -1
 
     return pixmap
 


jwst/outlier_detection/outlier_detection.py
```diff
@@ -556,13 +556,16 @@ def gwcs_blot(median_model, blot_img, interp='poly5', sinscl=1.0):
...
     pix_ratio = 1
     log.info('Blotting {} <-- {}'.format(blot_img.data.shape, median_model.data.shape))
 
     outsci = np.zeros(blot_img.shape, dtype=np.float32)
+
+    # Currently tblot cannot handle nans in the pixmap, so we need to give some
+    # other value.  -1 is not optimal and may have side effects.  But this is
+    # what we've been doing up until now, so more investigation is needed
+    # before a change is made.  Preferably, fix tblot in drizzle.
+    pixmap[np.isnan(pixmap)] = -1
```
hbushouse, line 568: So have you seen any bad side effects from this addition, since the same thing was causing a problem in forward drizzling?

hbushouse: Ah, of course now that I think about it a bit more, I guess this really isn't a new thing for blotting. It was already there due to the resetting of `pixmap` values in `resample_utils`? And hence all you're doing is preserving that pre-existing behavior for blotting?

jdavies-st: Well, this is what it was doing before for all pixmap creation.  Since we don't want it generally, but need it specifically for  blotting, I moved it out of `calc_gwcs_pixmap` and made it specific for blotting.

I think it would be good to look at this, but I think there are other issue in outlier detection that preventing good testing of this at the moment.  So basically this line of code just keeps the status quo for blotting in outlier detection.

hbushouse: Yeah, that's what I figured out post-comment.

PR #5231, opened 2020-08-06
JP-1626: Fix NIRSpec IFU product names in spec3 ASN's

### LINKED ISSUE #5212: Association rule Asn_IFUGrating is create malformed Level3 product names
_Issue [JP-1626](https://jira.stsci.edu/browse/JP-1626) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

An example product name being created for Asn_IFUGrating is as follows:

jw00736-o038_t001_nirspec_tg235h

which is fine. However, the code takes this as the full root and, when appending the rest of the optical path, does the following

jw00736-o038_t001_nirspec_tg235h_g140h-f100lp

when it should be doing

jw00736-o038_t001_nirspec_tg235h-g140h-f100lp

since the `tg235h` is actually part of the optical path.

hbushouse:
This updates the level-3 ASN rules to remove the errant "t" prepended to the grating name for NIRSpec IFU product names that get stored in spec3 ASN files, and updates the part of the `cube_build` step that composes output file names to take into account the fact that the input product base name may already contain the grating (and hence don't add a second one). Rather than having `cube_build` decide whether or not it needs to add a grating name, the existing one is simply removed and then all the remaining logic for adding grating and filter names to the product name remains the same.

Fixes #5212 / [JP-1626](https://jira.stsci.edu/browse/JP-1626)

### DIFF
--- a/jwst/cube_build/ifu_cube.py
+++ b/jwst/cube_build/ifu_cube.py
@@ -1096,18 +1096,18 @@ class IFUCubeData():
             if np.isnan(self.weight_power):
                 self.weight_power = None
 
-        log.debug(f'spatial size {self.spatial_size:.7f}')
-        log.debug(f'spectral size {self.spectral_size:.7f}')
-        log.debug(f'spatial roi {self.rois:.7f}')
-        log.debug(f'wave min and max {self.wavemin:.7f} {self.wavemax:.7f}')
-        log.debug(f'linear wavelength {self.linear_wavelength:.7f}')
-        log.debug(f'roiw {self.roiw:.7f}')
+        log.debug(f'spatial size {self.spatial_size}')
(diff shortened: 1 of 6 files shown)


jwst/cube_build/ifu_cube.py
```diff
@@ -192,6 +192,14 @@ def define_cubename(self):
                         '_single_s3d.fits'
 # ________________________________________________________________________________
             elif self.instrument == 'NIRSPEC':
+
```
hbushouse, line 195: This block of code is the only real change to the `cube_build` step. All the other edits are fixes to the cosmetics of some log messages.

PR #5232, opened 2020-08-06
Update setup.cfg package metadata

jdavies-st:
This updates the package metadata and does general cleanup of the `setup.py` and `setup.cfg` files, opting for the declarative approach in `setup.cfg` as much as possible.

This also fixes the `twine check` warning we are currently seeing:

```
Checking dist/jwst-0.16.3.dev94+gaaa0ebfb-py3-none-any.whl: PASSED, with warnings
  warning: `long_description_content_type` missing. defaulting to `text/x-rst`.
Checking dist/jwst-0.16.3.dev94+gaaa0ebfb.tar.gz: PASSED, with warnings
  warning: `long_description_content_type` missing. defaulting to `text/x-rst`.
```


setup.cfg
```diff
@@ -1,10 +1,66 @@
 [metadata]
-package_name = jwst
-description = JWST Pipeline
-author = STScI
-license = BSD
-edit_on_github = False
-github_project = spacetelescope/jwst
+name = jwst
+description = Python library for science observations from the James Webb Space Telescope
+long_description = file: README.md
```
nden, line 4: This is what the front page on PyPI will be. Do we want to put there all that is in the `README`?
All the information about DMS builds and links to Jenkins?

setup.cfg
```diff
@@ -1,10 +1,66 @@
...
+setup_requires =
+    setuptools_scm
+install_requires =
+    asdf>=2.5
+    astropy>=4.0
+    crds>=7.4.1.3
+    drizzle>=1.13
+    gwcs>=0.13.0
+    jsonschema>=3.0.1
+    numpy>=1.16
```
nden, line 27: This looks pretty old too. Should we go with v 0.17 as the oldest? The current is 0.19.

setup.cfg
```diff
@@ -1,10 +1,66 @@
...
+    photutils>=0.7
+    pyparsing>=2.2
+    requests>=2.22
+    scipy>=1.1.0
```
nden, line 31: Current release is 1.5. Should we up this too, e.g. to 1.3? I haven't kept up with scipy dev recently.

setup.cfg
```diff
@@ -1,10 +1,66 @@
...
+    Programming Language :: Python :: 3
+
+[options]
+zip_safe = False
+python_requires = >=3.6
+setup_requires =
+    setuptools_scm
+install_requires =
+    asdf>=2.5
+    astropy>=4.0
```
nden, line 22: Perhpas set this to `4.0.1.post1`. There was an issue with 4.0.

setup.cfg
jdavies-st: We actually test these in our "Oldest Dependencies" Travis CI job. I.e. `numpy 1.16` actually does work with all the other old dependencies.  So while people should be using newer versions, it is not required.  If it eventually stops working that Oldest Dependencies Travis job will start failing and we'll have to update this.

And that oldest dependencies Travis job actually gets the spec from `install_requires` minimums.

setup.cfg
jdavies-st: Same here.  It works, so while people should use the newer one, it is not required.

setup.cfg
jdavies-st: It was some install issue for certain platforms, right?  We currently test with 4.0.0 on Linux without problem, but perhaps other platforms (Windows) it is not good?

setup.cfg
jdavies-st: That's a good question.  Maybe we should break out our README.md into chunks and only include one of those chunks on the PyPi page?

jdavies-st: @hbushouse, @stscieisenhamer, any opinions on what sections in our current README, if anything, should we not have on the PyPi landing page?

Some examples of some other README files for packages we all use:

https://pypi.org/project/pytest/
https://pypi.org/project/numpy/
https://pypi.org/project/astropy/
https://pypi.org/project/matplotlib/
https://pypi.org/project/asdf/
https://pypi.org/project/crds/

As you can see, they run the gamut of just some links, to the full help of how to use the package.

setup.cfg
nden: Ah, I wasn't aware of that job. Awesome!

jdavies-st: Yes, there's a script that generates a `requirements-min.txt` file.  The Travis job runs it, installs those deps and runs the unit tests on that installation to verify they are correct.

https://github.com/spacetelescope/jwst/blob/master/scripts/minimum_deps

We also have a script that verifies that `install_requires` lists everything that the package actually needs at runtime.  I.e. if someone adds a runtime dependency on `pytest` or other test or doc dependencies, then this will catch it.

https://github.com/spacetelescope/jwst/blob/master/scripts/verify_install_requires

We had both the above things cause our `install_requires` to get out-of-date, and in the latter case, break a delivered tagged version of the code.

setup.cfg
```diff
@@ -1,10 +1,66 @@
 [metadata]
-package_name = jwst
-description = JWST Pipeline
-author = STScI
-license = BSD
-edit_on_github = False
-github_project = spacetelescope/jwst
+name = jwst
+description = Python library for science observations from the James Webb Space Telescope
```
hbushouse, line 3: I've never really liked this description. It's far too vague, in my opinion (but then so is the package name "jwst"). We're definitely not the only package out there that contains software for science observations from JWST. Our nextdoor neighbors in DATB have pkgs that contain just as much code for JWST science observations from JWST. What's distinctive about ours is the fact that it's a package of software for *calibrating* science observations from JWST.

setup.cfg
hbushouse: Definitely leave out the section with the DMS builds table and the regression tests. What about unit tests - I guess that's still of use to the public? What about JupyterHub - is that info for the public?

setup.cfg
```diff
@@ -1,10 +1,66 @@
...
+long_description = file: README.md
+long_description_content_type = text/markdown
+author = JWST Pipeline developers
```
hbushouse, line 6: This is not the whole JWST pipeline (that includes all of DMS). We need to be careful to qualify every instance of "pipeline" with "calibration".

setup.cfg
jdavies-st: Ok, so something like `Library for calibration of science observations from the James Webb Space Telescope`?

setup.cfg
jdavies-st: In order to break up the README into chunks, we would have to switch from markdown to restructured text.  Perhaps just a link to our existing `README.md` would be OK for the long description on PyPi?

jdavies-st: So for this PR, I'm just removing the long description and adding in the links to our github page and documentation on RTD in the standard PyPi way.  We can rethink how we want to present this information on PyPi, if needed, in a separate PR and maybe after we've already at least done a release to PyPi.  I think this will suffice for now, while not putting on PyPi all the gory detail we have in our current README.md.

pyproject.toml
```diff
@@ -1,7 +1,5 @@
 [build-system]
-requires = [
-    "setuptools>=38.2.5",
-    "setuptools_scm",
-    "wheel",
-]
+requires = ["setuptools>=42", "setuptools_scm[toml]>=3.4", "wheel"]
 build-backend = "setuptools.build_meta"
+
+[tool.setuptools_scm]
```
eslavich, line 5: Does this empty section have an effect of some kind?

setup.cfg
```diff
@@ -1,10 +1,68 @@
...
-package_name = jwst
-description = JWST Pipeline
-author = STScI
-license = BSD
-edit_on_github = False
-github_project = spacetelescope/jwst
+name = jwst
+description = Library for calibration of science observations from the James Webb Space Telescope
+author = JWST calibration pipeline developers
+license = BSD 3-Clause License
```
eslavich, line 5: Should this be the "official" short identifier?  Only asking because that turned up as a warning when I last submitted an update to asdf on one of the conda channels.  Per https://spdx.org/licenses/BSD-3-Clause.html, `BSD-3-Clause` is the short identifier for this license.

jdavies-st: Ah, thanks!

We are not delivering to any conda channels, but it would be good to standardize this.  That said, this was my example

https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

💭

pyproject.toml
jdavies-st: It does.  It's the equivalent of putting `use_scm_version=True` in `setup.py`.

https://pypi.org/project/setuptools-scm/

setup.cfg
stscieisenhamer: I would have to check, but I think James White added some type notation. Is python 3.6 still compatible with that?

jdavies-st:
Yes, `python>=3.5` supports `typing` in the standard library.  👍

setup.cfg
jdavies-st: I've updated to `BSD-3-Clause`.

PR #5236, opened 2020-08-10
JP-1587 Implement inverse of the FlatFieldStep

stscieisenhamer:
After many red-herrings and other issues, the inversion of `FlatFieldStep` is (near) ready.

Issues/conditions:

~1. Inversion is not done for any other instrument except NIRSpec. A `NotImplementedError` is currently produced. Just haven't looked that far.~ Done.
2. Question: For the override of the flat field, how should that be recorded in the headers. Reference files are populated, but not sure if the user-supplied file should go into the `FLAT` reference or as something else. Storing in `FLAT` may be misleading, especially for NIRSpec, since that reference file type is not used at all.

### DIFF
--- a/jwst/flatfield/flat_field_step.py
+++ b/jwst/flatfield/flat_field_step.py
@@ -92,7 +92,9 @@ class FlatFieldStep(Step):
 
             # Record the user-supplied flat as the FLAT reference type for recording
             # in the result header.
-            self._reference_files_used.append(('flat', Path(self.user_supplied_flat).name))
+            self._reference_files_used.append(
+                ('flat', reference_file_models['user_supplied_flat'].meta.filename)
+            )
 
         # Do the flat-field correction
         output_model, flat_applied = flat_field.do_correction(
--- a/jwst/flatfield/flat_field_step.py
+++ b/jwst/flatfield/flat_field_step.py
@@ -43,6 +43,7 @@ class FlatFieldStep(Step):
     spec = """
         save_interpolated_flat = boolean(default=False) # Save interpolated NRS flat
         user_supplied_flat = string(default=None)  # User-supplied flat
+        inverse = boolean(default=False)  # Invert the operation
     """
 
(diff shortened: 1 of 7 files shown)


hbushouse:
Regarding the recording of an override flat in the header, for the non-NIRSpec cases it's easy, because it should just get recorded in the `R_FLAT` keyword. In normal NIRSpec cases, where a user provides overrides for one or more of the DFLAT, FFLAT, or SFLAT ref files, those too can be stored in the corresponding `R_DFLAT`, `R_FFLAT`, and `R_SFLAT` keyword values. But in the case where a single combined flat is provided on input for use with NIRSpec, the solution is not obvious. Perhaps in that case the name of it should just get stored in the `R_FLAT` keyword and don't populate the `R_[DFS]FLAT` keywords at all (such that they won't appear in the header). I think that's the best we can do.

jwst/flatfield/flat_field_step.py
```diff
@@ -87,10 +90,15 @@ def process(self, input):
                 'user_supplied_flat': datamodels.open(self.user_supplied_flat)
             }
 
+            # Record the user-supplied flat as the FLAT reference type for recording
+            # in the result header.
+            self._reference_files_used.append(('flat', Path(self.user_supplied_flat).name))
```
hbushouse, line 95: What happens here if the user supplied flat is just in the form of a datamodel, as opposed to an actual file on disk? And isn't that the way it's going to work in the nrsmsb stuff in calwebb_spec2, i.e. we'll just be passing in a previously computed datamodel?

stscieisenhamer: Right, this would fail painfully. 👍

hbushouse: Anything worth failing is worth failing painfully.

stscieisenhamer:
Have not clue what is going on with the document build. Cannot reproduce locally. Will merge since this is just to a feature branch and we'll see if it keeps up.

PR #5242, opened 2020-08-12
Implement correction parameter API for FlatFieldStep

stscieisenhamer:
Implement the "easy/lazy" method of passing correction parameters around for FlatFieldStep.

Side note: The regressions discovered a dangling-if-without-else condition. See line 293 in flat_field.py. Basically, there was the possibility that NIRSPec-like model could be passed in, but have no flat field applied silently. Do not know if the condition could have happened in real life. However, because of changes where the reference files are no longer always required, the condition became possible. A runtime exception has been added just to ensure that the condition does not go unnoticed.

### DIFF
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -18,7 +18,7 @@ log.setLevel(logging.DEBUG)
 
 MICRONS_100 = 1.e-4                     # 100 microns, in meters
 
-# This is for NIRSpec.  
+# This is for NIRSpec.
 FIXED_SLIT_TYPES = ["NRS_LAMP", "NRS_BRIGHTOBJ", "NRS_FIXEDSLIT"]
 NIRSPEC_SPECTRAL_EXPOSURES = ['NRS_BRIGHTOBJ', 'NRS_FIXEDSLIT', 'NRS_IFU', 'NRS_MSASPEC']
 
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -74,8 +74,7 @@ def do_correction(input_model,
     # NIRSpec spectrographic data are processed differently from other
     # types of data (including NIRSpec imaging).  The test on flat is
     # needed because NIRSpec imaging data are processed by do_flat_field().
-    if input_model.meta.instrument.name == 'NIRSPEC' and \
-       input_model.meta.exposure.type in NIRSPEC_SPECTRAL_EXPOSURES:
+    if input_model.meta.exposure.type in NIRSPEC_SPECTRAL_EXPOSURES:
         flat_applied = do_nirspec_flat_field(output_model, fflat, sflat, dflat,
                                              user_supplied_flat=user_supplied_flat,
                                              inverse=inverse)
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -18,8 +18,9 @@ log.setLevel(logging.DEBUG)
 
 MICRONS_100 = 1.e-4                     # 100 microns, in meters
 
-# This is for NIRSpec.  These exposure types are all fixed-slit modes.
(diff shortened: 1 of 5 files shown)


jwst/flatfield/flat_field.py
```diff
@@ -73,7 +74,8 @@ def do_correction(input_model,
     # NIRSpec spectrographic data are processed differently from other
     # types of data (including NIRSpec imaging).  The test on flat is
     # needed because NIRSpec imaging data are processed by do_flat_field().
-    if input_model.meta.instrument.name == 'NIRSPEC' and flat is None:
+    if input_model.meta.instrument.name == 'NIRSPEC' and \
```
hbushouse, line 77: It's not obvious to me that you need to check both instrument name and exp_type here, since it would be pretty impossible to have instrument name be anything but NIRSPEC if the exp_type is in that list, but of course it can't hurt anything either.

jwst/flatfield/flat_field.py
```diff
@@ -288,6 +290,9 @@ def do_nirspec_flat_field(output_model, f_flat_model, s_flat_model, d_flat_model
                                    .format(type(output_model)))
             return nirspec_ifu(output_model, f_flat_model, s_flat_model, d_flat_model, dispaxis,
                                user_supplied_flat=user_supplied_flat, inverse=inverse)
+        else:
+            raise RuntimeError(f'No flat field algorithm exists for handling data {output_model}')
```
hbushouse, line 294: Is `output_model` a datamodel here and, if so, how does that get printed into a log message?

jwst/pipeline/calwebb_spec2.py
```diff
@@ -421,12 +421,10 @@ def _process_nirspec_masterbackground(self, data):
         """
         # First pass: just do the calibration to determine the correction
         # arrays.
-        pre_calibrated, ff_corrections = self.flat_field(
-            data, force_extended=True, return_corrections=True
-        )
-        pre_calibrated = self.pathloss(pre_calibratedforce_extended=True, return_corrections=True)
-        pre_calibrated = self.barshadow(pre_calibrated, force_extended=True, return_corrections=True)
-        pre_calibrated = self.photom(pre_calibrated, force_extended=True, return_corrections=True)
+        pre_calibrated = self.flat_field(data)
```
hbushouse, line 424: Don't we want to still force all of these to be treated as extended sources, so that the computed corrections are appropriate for application to the master bkg data later on in the 2nd pass?

jwst/flatfield/flat_field.py
stscieisenhamer: Yes, that is probably right. Will remove the instrument check. Just being too conservative on the changes.

jwst/flatfield/flat_field.py
stscieisenhamer: The repr of data models are that strings like "ModelType(size) filename". These happen all the time now in logging.

jwst/pipeline/calwebb_spec2.py
stscieisenhamer: Yes, and that will be coming with the Pathloss step.

stscieisenhamer:
About the "where are the correction parameters stored": It is in the step attribute "correction_pars". This is set when a Step is run. If a Step is then re-run or has the Step.correction_pars explicitly set, and the attribute Step.use_correction_pars is set, then it is used instead of recalculation based on reference files, etc.

Currently there is no option to save as part of the Step. But that is something to be discussed.

hbushouse:
So the correction arrays (or a datamodel containing them) are somehow stored in memory somewhere within the Step class instance?

jwst/pipeline/calwebb_spec2.py
hbushouse: Ok, so you'll be adding that param back into the calling sequence for the other steps (pathloss, barshadow, photom)?

PR #5243, opened 2020-08-12
JP-1641: Update calami3 to exclude targ_acqs from processing

### LINKED ISSUE #5240: NIRISS ami3 associations are writing calibration output files for exptype=target_acquisition members
_Issue [JP-1641](https://jira.stsci.edu/browse/JP-1641) was created on JIRA by [Mike Swam](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mswam):_

ami3 calibrations at Level3 are generating output Level2c files for the exptype=target_acquisition members.  Since the target acq members are shared between asn tables in the same program, this means multiple ami3 runs are generating the same output file, which are colliding and causing checksum mismatches in FITS keywords found during INGEST_2C (see JSDP-1571).

Howard and Jonathan point out that ami3 should skip over generating output products for the exptype=target_acquisition members.

hbushouse:
Updated the `calwebb_ami3.py` module to only send PSF and science target members from the input ASN through the processing steps, so as to exclude any other member types, such as target_acquisition. Comparison with previous results shows that "ami" products are no longer created for target_acq members and the results of all other processing are identical.

Fixes #5240 / [JP-1641](https://jira.stsci.edu/browse/JP-1641)

### DIFF
--- a/jwst/pipeline/calwebb_ami3.py
+++ b/jwst/pipeline/calwebb_ami3.py
@@ -2,7 +2,6 @@
 import logging
 import os.path as op
 
-from .. import datamodels
 from ..stpipe import Pipeline
 
 # step imports
--- a/jwst/pipeline/calwebb_ami3.py
+++ b/jwst/pipeline/calwebb_ami3.py
@@ -44,8 +44,7 @@ class Ami3Pipeline(Pipeline):
         # Load the input association table
         asn = self.load_as_level3_asn(input)
 
-        # We assume there's one final product defined by the
-        # association
+        # We assume there's one final product defined by the association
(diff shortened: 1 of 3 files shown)


jwst/pipeline/calwebb_ami3.py
```diff
@@ -110,29 +118,15 @@ def process(self, input):
 
                 # Perform blending of metadata for all inputs to this
                 # output file
-                self.log.info(
-                    'Blending metadata for averaged psf...'
-                )
-                self.log.info(
```
hbushouse, line 116: Note sure why these log messages were here and of what use they were, so I got rid of them.

PR #5259, opened 2020-08-18
Use python 3.8 in nightly regression tests

jdavies-st:


hbushouse:
Reverting until we can get `pymssql` issues under 3.8 figured out.

PR #5264, opened 2020-08-19
use gwcs 0.14.0 for sdp delivery

nden:
This PR sets the version of gwcs in SDP to 0.14.0 and the version of asdf to 2.7.1
It includes some minor changes to the Nirspec pipeline to take advantage of the fact that the WCS pipeline is not a list of `wcs.Step` instances instead of a list of (frame, transform) tuples and the `transform` and `frame` of a step should be accessed as attributes. Indexing the step still works but raises a `DeprecationWarning`.

Update: Bumped the version numbers of gwcs and asdf in setup.py


nden:
Regression tests passed with this version of gwcs.

PR #5267, opened 2020-08-19
NIRspec lamps new PR to check for conflicts

stscirij:
Not sure whether the original PR (#4889) is good any more


jwst/assign_wcs/nirspec.py
```diff
@@ -211,7 +219,19 @@ def ifu(input_model, reference_files, slit_y_range=[-.55, .55]):
 
     det, sca, gwa, slit_frame, msa_frame, oteip, v2v3, world = create_frames()
 
-    if input_model.meta.instrument.filter != 'OPAQUE':
+    exp_type = input_model.meta.exposure.type.upper()
+
+    is_lamp_exposure = exp_type in ['NRS_LAMP', 'NRS_AUTOWAVE', 'NRS_AUTOFLAT']
+
+    if input_model.meta.instrument.filter == 'OPAQUE' or is_lamp_exposure:
+        # If filter is "OPAQUE" or if internal lamp exposure the pipeline stops at the MSA.
```
hbushouse, line 227: The term "pipeline" is pretty over-subscribed. Perhaps mention explicitly that we're talking about the NIRSpec WCS pipeline in this case.

stscirij: OK

stscirij:
The only regression test failures should be the introduction of the OPMODE keyword for most NIRSPEC datasets

PR #5277, opened 2020-08-24
JP-1653 changed default weighting to emsm

jemorrison:
@drlaw1558 
This should do it. I am checking on what the nightly regression test use for the CRDS context. Once we confirm it includes the new NIRSpec Cube pars reference file we will be able to merge this PR


jemorrison:
@hbushouse  or @stscieisenhamer 
 Ok for me to merge ?

PR #5278, opened 2020-08-25
Add ability to change artifactory environment

stscieisenhamer:
Add a new Jenkins build string parameter "ARTIFACTORY_ENV" so that the regressions can use a different artifactory environment instead of just "dev".

To setup Jenkins builds, in "Configure", add a new string parameter called "ARTIFACTORY_ENV". Make the default "dev", and set the description to "Set which environment to use from the jwst-pipeline artifactory to test against."


jdavies-st:
Would this be better to merge into master (and then update the `jp-1213-nrsmsb` branch) so that all of our Jenkins builds can use it?  If it is in that feature branch, it will eventually get into master anyway once merged.

stscieisenhamer:
Just doing it here because this feature needs this now. I can add to master in a bit.

PR #5279, opened 2020-08-25
Reset to EMSM cube defaults

drlaw1558:
Adding a couple of places where cube building weight function had been changed to MSM that were missed in https://github.com/spacetelescope/jwst/pull/5277/


drlaw1558:
@jemorrison

jemorrison:
@drlaw1558  Good catch I forgot about how cube build is called from outlier detection. For mrs_imatch it was emsm.

PR #5286, opened 2020-08-27
Remove old regtests and regtest infrastructure

jdavies-st:
We no longer run or maintain these tests, nor use this infrastructure.  Probably best to remove it.


hbushouse (approved):

jdavies-st:
This also means we can remove the `jwst-pipeline/old` data on bytesalad.

PR #5295, opened 2020-08-31
Fix for JP-1642: NIRSpec IFU failure in calwebb_spec2

jemorrison:
Fixes [JP-1642](https://jira.stsci.edu/browse/JP-1642)


jemorrison:
The PR will close JP-1642.
If ifucube contains all 0 data for every wavelength plane then self.skip = True and calspec2 does not write out the s3d or x1d data

Issue #5290, opened 2020-08-28
Installable package should be smaller


stscijgbot-jp:
_Issue [JP-1662](https://jira.stsci.edu/browse/JP-1662) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

Currently the installed `jwst` package size is 30-40MB because a lot of large test data is packaged with the installed `sdist` and `bdist_wheel`.  We should stop packaging all this large data.

Perhaps this means packaging no large data and marking the tests that rely on it so that they only run when the package is installed from github?

jdavies-st:
Related to 
https://github.com/spacetelescope/jwst/issues/3751

jdavies-st:
The culprits >100K:

```
$ find jwst -type f -exec du -h {} + | sort --human-numeric-sort --reverse
 40M    jwst/datamodels/tests/data/nircam_flat.fits
 16M    jwst/datamodels/tests/data/nircam_readnoise.fits
 16M    jwst/datamodels/tests/data/nircam_mask.fits
 16M    jwst/datamodels/tests/data/nircam_gain.fits
7.7M    jwst/extract_1d/tests/data/jwst_nirspec_apcorr_fs_dummy.fits
7.5M    jwst/lib/tests/data/acs_tlm_data_4_stsci_dms_jitter_file_mod_RA_DEC_PA.csv
5.4M    jwst/lib/tests/notebooks/full_table.csv
5.3M    jwst/exp_to_source/tests/data/sources/ori_jwst_nod1_cal.fits
5.2M    jwst/datamodels/tests/data/test.fits
3.8M    jwst/associations/tests/notebooks/make_tests.ipynb
1.9M    jwst/assign_wcs/tools/nirspec/jw95065006001_01_msa.fits
1.9M    jwst/assign_wcs/tests/data/msa_configuration.fits
1.9M    jwst/assign_wcs/tests/data/Trace_IFU_Slice_00_SMOS-MOD-G1M-17-5344175105_30192_JLAB88.fits
1.8M    jwst/lib/tests/data/engdb_jw00703/sa_zaducmdy_data.json
1.7M    jwst/lib/tests/data/engdb_jw00703/sa_zaducmdx_data.json
1.7M    jwst/assign_wcs/tests/data/Trace_SLIT_A_200_1_V84600010001P0000000002101_39547_JLAB88.fits
[...]

PR #5296, opened 2020-08-31
Remove 88MB of test data from the repo; simplify test using it

jdavies-st:
So let's remove the 88MB of data files used by the very simple `test_open_reference_files()` and replace the test with a parametrized test with better coverage using `tempfiles`.

Also, some of the reffiles are out-of-date, so we shouldn't even be doing tests like this, i.e. packaging a reffile to test a general feature.

On `master`:
```
$ ls -lh dist/*
-rw-rw-r--  1 jdavies  STSCI\science    30M Sep  7 20:35 dist/jwst-0.17.1.dev1+g8e212f12-py3-none-any.whl
-rw-rw-r--  1 jdavies  STSCI\science    38M Sep  7 20:35 dist/jwst-0.17.1.dev1+g8e212f12.tar.gz
```

This PR
```
$ ls -lh dist/*
-rw-rw-r--  1 jdavies  STSCI\science   3.6M Sep  7 17:12 dist/jwst-0.17.1.dev11+g5f9aee86-py3-none-any.whl
-rw-rw-r--  1 jdavies  STSCI\science    12M Sep  7 17:12 dist/jwst-0.17.1.dev11+g5f9aee86.tar.gz
```
Addresses #5290


jwst/regtest/test_miri_image.py
```diff
@@ -68,7 +68,7 @@ def run_image3(run_image2, rtdata_module):
         "--steps.tweakreg.snr_threshold=200",
         "--steps.tweakreg.use2dhist=False",
         "--steps.tweakreg.minobj=4",
-        "--steps.source_catalog.snr_threshold=20",
+        "--steps.source_catalog.snr_threshold=10",
```
jdavies-st, line 71: This change is to get back to detecting the 5 sources in the MIRI Image regtest.  One source got overlooked when the background subtraction bug was fixed.  See #4858.

hbushouse (approved):

jwst/regtest/test_miri_image.py
jdavies-st: After this is merged, we'll have to OKify the source catalog test in `regtests/test_miri_image.py`.

PR #5299, opened 2020-09-01
JP-1665 Remove unwarranted closing of datamodels in blendheaders

### LINKED ISSUE #5297: Image3Pipeline operation on closed file
_Issue [JP-1665](https://jira.stsci.edu/browse/JP-1665) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Following command ends with an "Operation on closed file" error.

strun cfgs/calwebb_image3.cfg jw00617-o083_20200901t055714_image3_001_asn.json

stscieisenhamer:
Resolves [JP-1665](https://jira.stsci.edu/browse/JP-1665)
Resolves #5297

### DIFF
--- a/jwst/regtest/test_nircam_image.py
+++ b/jwst/regtest/test_nircam_image.py
@@ -72,10 +72,9 @@ def run_image3pipeline(run_image2pipeline, rtdata_module, jail):
     Step.from_cmdline(args)
 
 
-@pytest.fixture(scope="module")
-def run_image3_closedfile(rtdata_module, jail):
+@pytest.fixture()
+def run_image3_closedfile(rtdata, jail):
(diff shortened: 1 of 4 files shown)


jwst/regtest/test_nircam_image.py
```diff
@@ -72,6 +72,19 @@ def run_image3pipeline(run_image2pipeline, rtdata_module, jail):
     Step.from_cmdline(args)
 
 
+@pytest.fixture(scope="module")
```
jdavies-st, line 75: If you scope it like this, this data will be written to the same tmpdir as the other dataset in this module.  Probably not what we want.  Probably better to make this function-scoped and and use the `rtdata` fixture instead of the `rtdata_module` fixture since there is only one test function that calls this fixture.

PR #5302, opened 2020-09-02
JP-509 Implement NIRSpec MOS master background subtraction calwebb_spec2

### LINKED ISSUE #3073: NIRSpec MOS master background subtraction
Issue [JP-509](https://jira.stsci.edu/browse/JP-509) was created by Howard Bushouse:

The spectroscopic master background subtraction for NIRSpec MOS mode will form a master background spectrum from the data provided by one or more background shutters.

The background shutters are identified in the MSA meta data file that is one of the inputs to the calwebb_spec2 pipeline when processing MOS exposures.

The use of dedicated background shutters and nodding of exposures (also used to subtract background) are mutually exclusive in the APT. Hence dedicated background shutters will only appear for exposures that do not use nodding and hence will be single exposures.

Within the calwebb_spec2 processing flow, the background shutters will need to be processed all the way to the stage of 1-D extraction (extract_1d step), at which point the 1-D spectra for all background shutters will be averaged together to form the master background spectrum. The 1-D master background spectrum is then re-projected back into the 2-D space of each defined source (non-background) slitlet and subtracted from the 2-D source slitlet data. The source slitlet data are then processed to completion, namely through the stage of extract_1d.

stscieisenhamer:
Resolves [JP-509](https://jira.stsci.edu/browse/JP-509)
Resolves #3073


CHANGES.rst
```diff
@@ -142,14 +142,15 @@ extract_2d
...
 
-
 flat_field
 ----------
 - For NIRSpec BOTS and ALLSLITS add the slit start corner to the subarray start corner
   when determining what region of the flat_field reference files to extract. [#5269]
 
 - Enable NIRSpec lamp processing in calspec2 pipeline. [#5267]
 
+- Implement providing a user-supplied flat field to which overrides all references [#5233]
```
hbushouse, line 152: Fix grammatical typo ("to which overrides")

hbushouse: flat_field was also updated to implement inverse correction

JenkinsfileRT
```diff
@@ -64,7 +64,8 @@ bc0.build_cmds = [
 bc0.build_cmds = PipInject(env.OVERRIDE_REQUIREMENTS) + bc0.build_cmds
 bc0.test_cmds = [
     "pytest --cov-report=xml --cov -r sxf -n 30 --bigdata --slow \
-    --basetemp=${pytest_basetemp} --junit-xml=results.xml --dist=loadscope",
+    --basetemp=${pytest_basetemp} --junit-xml=results.xml --dist=loadscope \
+    --env=${env.ARTIFACTORY_ENV} ",
```
jdavies-st, line 68: When this gets merged, we need up update all the build jobs (`JWST`, `JWST_UNSTABLE`, `jdavies-dev`, and `eslavich-dev`) that run the regression tests with this parameter.

jwst/pipeline/calwebb_spec2.cfg
```diff
@@ -9,6 +9,8 @@ save_results = True
       [[imprint_subtract]]
       [[msa_flagging]]
       [[extract_2d]]
+      [[master_background]]
```
hbushouse, line 12: The master_background step itself never actually gets called, even when processing MOS, does it? We just call various functions within master_background directly. So it's not obvious to me why this needs to be added to the cfg's and to the calwebb_spec2 pipeline module itself.

jwst/regtest/test_nirspec_brightobj.py
```diff
@@ -49,3 +52,31 @@ def test_nirspec_brightobj_spec2(run_tso_spec2_pipeline, fitsdiff_default_kwargs
     # Compare the results
     diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
     assert diff.identical, diff.report()
+
+
+@pytest.mark.bigdata
+def test_flat_field_step_user_supplied_flat(jail, rtdata_module, fitsdiff_default_kwargs):
```
jdavies-st, line 58: Since this test doesn't share any data with the other tests in this module, it should use the `rtdata` fixture which is function-scoped.  Data for this test will then be segregated from data from the other tests in this module.

jwst/regtest/test_nirspec_brightobj.py
```diff
@@ -49,3 +52,31 @@ def test_nirspec_brightobj_spec2(run_tso_spec2_pipeline, fitsdiff_default_kwargs
...
+    rtdata.output = 'flat_fielded_step_user_supplied.fits'
+    data_flat_fielded.write(rtdata.output)
+
+    rtdata.get_truth('truth/test_nirspec_brightobj_spec2/flat_fielded_step_user_supplied.fits')
+    diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
+    assert diff.identical, diff.report()
+
+
+@pytest.mark.bigdata
+def test_ff_inv(jail, rtdata_module, fitsdiff_default_kwargs):
```
jdavies-st, line 74: Same with this one.  Use `rtdata` instead of `rtdata_module`.  And `jail` is not necessary, as `rtdata` already includes the function-scoped `_jail` fixture.

jwst/regtest/test_nirspec_fs_spec2.py
```diff
@@ -63,3 +67,62 @@ def test_nirspec_fs_spec2(run_pipeline, fitsdiff_default_kwargs, suffix):
     # Compare the results
     diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
     assert diff.identical, diff.report()
+
+
+@pytest.mark.bigdata
+def test_pathloss_corrpars(jail, rtdata_module):
```
jdavies-st, line 73: function scoped here

jwst/regtest/test_nirspec_fs_spec2.py
```diff
@@ -63,3 +67,62 @@ def test_nirspec_fs_spec2(run_pipeline, fitsdiff_default_kwargs, suffix):
...
+    bad_slits = []
+    for idx, slits in enumerate(zip(corrected.slits, corrected_corrpars.slits)):
+        corrected_slit, corrected_corrpars_slit = slits
+        if not np.allclose(corrected_slit.data, corrected_corrpars_slit.data, equal_nan=True):
+            bad_slits.append(idx)
+    assert not bad_slits, f'correction_pars failed for slits {bad_slits}'
+
+
+@pytest.mark.bigdata
+def test_pathloss_inverse(jail, rtdata_module):
```
jdavies-st, line 93: function scoped here too.

jwst/regtest/test_nirspec_fs_spec2.py
```diff
@@ -63,3 +67,62 @@ def test_nirspec_fs_spec2(run_pipeline, fitsdiff_default_kwargs, suffix):
...
+    for idx, slits in enumerate(zip(data.slits, corrected_inverse.slits)):
+        data_slit, corrected_inverse_slit = slits
+        non_nan = ~np.isnan(corrected_inverse_slit.data)
+        if not np.allclose(data_slit.data[non_nan], corrected_inverse_slit.data[non_nan]):
+            bad_slits.append(idx)
+    assert not bad_slits, f'Inversion failed for slits {bad_slits}'
+
+
+@pytest.mark.bigdata
+def test_pathloss_source_type(jail, rtdata_module):
```
jdavies-st, line 114: function-scoped

jwst/regtest/test_nirspec_ifu_spec2.py
```diff
@@ -54,3 +63,95 @@ def test_spec2(run_spec2, fitsdiff_default_kwargs, suffix):
     """Regression test matching output files"""
     rt.is_like_truth(run_spec2, fitsdiff_default_kwargs, suffix,
                      truth_path=TRUTH_PATH)
+
+
+@pytest.mark.bigdata
+def test_nirspec_ifu_user_supplied_flat(jail, rtdata_module, fitsdiff_default_kwargs):
```
jdavies-st, line 69: All the new tests in this module should be function-scoped too.  Remove `jail` and `rtdata_module`, replace with `rtdata`.

jwst/regtest/test_nirspec_image2.py
```diff
@@ -17,3 +21,49 @@ def test_nirspec_image2(_jail, rtdata, fitsdiff_default_kwargs):
 
     diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
     assert diff.identical, diff.report()
+
+
+@pytest.mark.bigdata
+def test_flat_field_step_user_supplied_flat(jail, rtdata_module, fitsdiff_default_kwargs):
```
jdavies-st, line 27: Same here.  `rdata` for function-scoped for all new tests.

jwst/pipeline/calwebb_spec2.cfg
stscieisenhamer: Presuming that this should switchable, i.e., to perform or not perform, how would this be indicated? Would it be switched/piggy-backed on the existing `background` step?

Also, ideally, this should be encapsulated into a step. This wasn't done initially because it wasn't clear how exactly to implement, originally. This should be done.  However, since it is not done that way now, we can defer to whatever the answer is to the first question.

JenkinsfileRT
stscieisenhamer: would things horribly crash without the jobs being updated. just curious.

jdavies-st: Yes.  Especially since there's no definition of this anywhere in the JenkinsfileRT script other than here.  Notice how `env.OVERRIDE_REQUIREMENTS` is sorted out at the top of this script.

jwst/pipeline/calwebb_spec2.cfg
hbushouse: Agree that it should be switchable, at least eventually, in case anyone ever does NOT want it applied. Also, is there existing logic in the flow that will skip all of the master bkg operations for MOS mode if the regular `background` step was applied previously (i.e. for simple nodding cases)? Separate from this PR, we also need to update the `calwebb_spec3` `master_background` step to test whether the master bkg has already been subtracted here in `calwebb_spec2` (for MOS mode) and skip the level-3 step. Finally, does the current `calwebb_spec2` set of updates include setting the S_xxxxxx keyword for `master_background` to COMPLETE if/when it gets applied? The level-3 step logic would need that to figure out if it's already been done.

stscieisenhamer: spec2: Will leave in the pseudo-mbs step. I will put in the logic to make both background steps mutually exclusive and add the S_XXXXX keyword when it is done.

spec3: Will file issue(s) to capture the necessary changes. Will these need be resolved for B7.6?

JenkinsfileRT
stscieisenhamer: I can add similar code, defaulting to `dev`.

jwst/pipeline/calwebb_spec2.cfg
hbushouse: I think at least the spec2 items should be resolved for B7.6. spec3 update should be easy, but isn't imperative for B7.6.

hbushouse (approved):

PR #5311, opened 2020-09-03
Pointing analysis tools

stscieisenhamer:
Prompted by a request from DMS management, develop a couple tools to review/analyze pointing for exposures that have gone through `set_telescope_pointing`.

Two new commands are implemented:

- v1_calculate: Calculate V1 over a time period.
- pointing_summary: Summarize various pointing information in a table.


jwst/lib/pointing_summary.py
```diff
@@ -0,0 +1,146 @@
...
+import logging
+
+from astropy.coordinates import SkyCoord
+from astropy.table import Table
+import astropy.units as u
+
+import jwst.datamodels as dm
+
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
```
eslavich, line 41: What's the advantage of adding a `NullHandler` vs leaving the handler list empty?

jwst/lib/pointing_summary.py
```diff
@@ -0,0 +1,146 @@
...
+
+__all__ = ['Delta', 'calc_pointing_deltas', 'calc_deltas']
+
+# Basic delta structure
+Delta = namedtuple('Delta', 'target, v1, refpoint, delta_v1, delta_refpoint')
```
eslavich, line 46: This might be nice as a class with `delta_v1` and `delta_refpoint` as property methods.  Could add a `__repr__` to make logging convenient, too.

eslavich: (never mind, this isn't used outside of this class, is it)

jwst/lib/set_telescope_pointing.py
```diff
@@ -1545,3 +1594,52 @@ def pointing_from_average(mnemonics):
...
     # That's all folks
     return pointing
+
+
+def fill_mnemonics_chronologically(mnemonics, filled_only=True):
+    """Return time-ordered mnemonic list with progressive values
+
+    The different set of mnemonics used for observatory orientation
+    appear at different cadences. This routine creates a time-ordered dictionary
+    with all the mnemonics for each time found in the engineering. For mnemnonics
```
eslavich, line 1604: Possible typo: "in the engineering"

jwst/lib/tests/data/calc_deltas_truth.ecsv
```diff
@@ -0,0 +1,58 @@
+# %ECSV 0.9
```
eslavich, line 1: Wow this is like bizarro ASDF

jwst/lib/v1_calculate.py
```diff
@@ -0,0 +1,118 @@
...
+    v1_table = Table(v1_dict)
+    return v1_table
+
+
+def v1_calculate_over_time(obsstart, obsend, **calc_wcs_from_time_kwargs):
+    """Calculate V1 over the given time period
+
+    Parameters
+    ----------
+    obsstart, obbsend : float
```
eslavich, line 60: Typo in "obbsend"

eslavich (commented):
I'm not qualified to review what this is actually doing, but the code seems well structured to me.  I nitpicked a few docstrings.

hbushouse (approved):
Other than all the typos already pointed out by @eslavich looks good to me.

jwst/lib/pointing_summary.py
stscieisenhamer: The "sheepish" answer is that this is what [Python recommends](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library).

jwst/lib/pointing_summary.py
stscieisenhamer: Good point, or actually maybe that data class stuff. But, until a real use-case comes up, the line-in-sand is here.

jdavies-st (commented):
Instead of adding more test data to `jwst/lib/tests/data/`, would it be possible for the tests to generate these on-the-fly using the `tmpdir` fixture?  If not the `ecsv`, at least the FITS file?

stscieisenhamer:
About the test data: Removed the file `empty.fits`. However, the truth files, all the `ecsv`, have been left. That data would simply be replicated in the test itself, so there would be no savings.

jdavies-st (approved):
Looks good!

PR #5317, opened 2020-09-09
JP-891 Provide option to save the calculated master background in both 1D and 2D forms

stscieisenhamer:
Resolves [JP-891](https://jira.stsci.edu/browse/JP-891)

Notes

For the new NIRSpec MOS spec2 version, the implementation has been changed from an embedded, psuedo-step, to an actual step, MasterBackgroundNRSSlitsStep. This makes the management of saving the products more in-line with other steps, and also fixes a number of shortcomings with the pseudo-step approach.


jwst/regtest/test_nirspec_masterbackground.py
```diff
@@ -1,12 +1,90 @@
...
+def test_nirspec_spec2_mbkg(suffix, run_spec2_mbkg, fitsdiff_default_kwargs):
+    """Run spec2 with master background"""
+    rtdata = run_spec2_mbkg
+    rt.is_like_truth(rtdata, fitsdiff_default_kwargs, suffix, truth_path='truth/test_nirspec_mos_mbkg_user')
+
+    # Compare results
+    #rtdata.output = "nrs_mos_with_bkgslits_cal.fits"
+    #rtdata.get_truth('truth/test_nirspec_mos_mbkg_user/nrs_mos_with_bkgslits_cal.fits')
+    #diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
+    #assert diff.identical, diff.report()
```
jdavies-st, line 85: Looks like this test is commented out?

jwst/stpipe/step.py
```diff
@@ -1264,6 +1264,30 @@ def get_pars_model(step, full_spec=True):
...
+            Attributes to update. If a value is another dict,
+            recursively update those attributes.
+
+        Notes
+        -----
+        `attributes` is presumed to have been produced by the
+        `Step.get_pars` method. As such, the "steps" key is treated
+        special in that it is a dict whose keys are the steps assigned
+        directly as attributes to the current step. This is standard
+        practice for `Pipeline`-based steps.
```
jdavies-st, line 1282: It is not at all clear what this method does.  Also, it looks a bit dangerous.  Wouldn't it be better to have `Step.get_pars` have a post-hook that does the update, so the functionality here is tied to `Step.get_pars`?  Especially if the assumption is the dict is produced by `Step.get_pars`.

jwst/master_background/master_background_nrs_slits_step.py
```diff
@@ -0,0 +1,235 @@
...
+        data : `~jwst.datamodels.MultiSlitModel`
+            The data to operate on.
+
+        user_background : None, string, or `~jwst.datamodels.CombinedSpecModel`
+            Optional user-supplied master background 1D spectrum, path to file
+            or opened datamodel
+
+        Returns
+        -------
+        masterbkg_1d, masterbkg_2d : `~jwst.datamodels.CombinedSpecModel`, `~jwst.datamodels.MultiSlitModel`
```
jdavies-st, line 180: Is the background just a single file?  Is there a reason to return it in a `MultiSlitModel` when a `SlitModel` would suffice?

jdavies-st (commented):
Looks good.  Step architecture looks great.  Just a few comments/questions below.

jwst/regtest/test_nirspec_masterbackground.py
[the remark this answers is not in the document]
stscieisenhamer: It is and replaced by the lines above it. It was leftover of development and will delete. 😊

jwst/master_background/master_background_nrs_slits_step.py
stscieisenhamer: This is accurate. The `MultiSlitModel` is where all the slits have been replaced with the 2D expanded background for each slit. This is equivalent to the `background_2d` in `MasterBackgroundStep`, but for each slit. See the func `jwst.master_background.nirspec_utils.map_to_science_slits`

jwst/master_background/master_background_nrs_slits_step.py
```diff
@@ -0,0 +1,235 @@
...
+            getattr(self, step).update(pars)
+
+    @contextmanager
+    def preserve_pars(self):
+        """Ensure no changes to the Step's parameters"""
+        saved_pars = self.get_pars()
+        try:
+            yield saved_pars
+        finally:
+            self.update(saved_pars)
```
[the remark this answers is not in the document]
stscieisenhamer: Yes, this is exactly what happens, and this insulates against those changes. Offline discussion is probably best.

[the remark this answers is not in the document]
stscieisenhamer: Offline discussion clarified the algorithm. Actions concerning this point:

- More commentary
- Clearer name for the context manager, such as `preserve_step_parameters`

jwst/stpipe/step.py
stscieisenhamer: Offline discussion clarified the need. Actions are:

- More commentary
- Protect against updating attributes that are not already part of the relevant steps.
- Rename to a more specific function, such as `update_step_parameters`

hbushouse (approved):
LGTM

PR #5321, opened 2020-09-10
JP-1681: Fix empty int_times table in rateints products

### LINKED ISSUE #5320: rateints product has zero rows in INT_TIMES table
_Issue [JP-1681](https://jira.stsci.edu/browse/JP-1681) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

B7.6rc1 is creating rateints products from the ramp_fitting step that have zero rows in the INT_TIMES table extension, even when the table is fully populated in the input ramp file.

hbushouse:
Fix `ramp_fit` to populate the `int_times` attribute in the `rateints` products. I think it got removed during the rework to add multi-processing.

Fixes #5320 / [JP-1681](https://jira.stsci.edu/browse/JP-1681)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -53,6 +53,10 @@ pipeline
 
 - Include the per-slit failure traceback in any RuntimeError raised in Spec2Pipeline. [#5315]
 
+ramp_fitting
+------------
+
+- Reinstate copying of INT_TIMES table to output rateints product for TSO exposures. [#5321]
 
 tso_photometry
 --------------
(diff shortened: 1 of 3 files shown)


jdavies-st (commented):
Looks good. Could you add a small addition to the unit tests that checks this new code, or maybe reproduces the old error and shows this fixes it?

hbushouse:
> Looks good. Could you add a small addition to the unit tests that checks this new code, or maybe reproduces the old error and shows this fixes it?

I *could* create such a test, but it would take a little work to build (on the fly) an appropriate input image that contains a populated INT_TIMES table, rather than having it use an existing FITS file. So, I'll think about it  ;-)

dmggh (approved):

jdavies-st:
The existing tests have fixtures that produce on-the-fly data with int times populated.

https://github.com/spacetelescope/jwst/blob/adb39f7c2c0cb579dfa50f3d1551a26bcbc96a50/jwst/ramp_fitting/tests/test_ramp_fit.py#L671

I think you just need to write a very short test that checks that its there at the end, or add it to one of the existing tests.

https://github.com/spacetelescope/jwst/blob/master/jwst/ramp_fitting/tests/test_ramp_fit.py

PR #5325, opened 2020-09-14
change METERS to METER

jemorrison:


jemorrison:
This PR is just a simple fix in changing METERS to METER for NIRSPEC internal_cal output WCS unit for axis 1 and 2. It failed in a WCS check (that I is new - because I tested making these types of cubes when I added internal_cal as an type)

hbushouse (approved):

Issue #4996, opened 2020-05-28
Provide support for Windows


stscijgbot-jp:
_Issue [JP-1477](https://jira.stsci.edu/browse/JP-1477) was created on JIRA by [Steven Crawford](https://jira.stsci.edu/secure/ViewProfile.jspa?name=scrawford):_

We have had several requests for Windows support for the JWST calibration software.   At this time, we do not test or develop to support Windows.   An initial test run indicates there will need to be substantial work to CRDS and the JWST calibration software to support Windows environment.  We will also need to set up regular testing and distribution of the software under Windows.     

 

At this time, there is no plan to support Windows.

nden:
A number of recent PRs fixed this, the latest #7491

stscijgbot-jp:
_Comment by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw) on [JIRA](https://jira.stsci.edu/browse/JP-1477?focusedCommentId=690048#comment-690048):_

Still no plan to support Windows right now, setting to On Hold.

Issue #3864, opened 2019-07-29
Step.from_config_file should be able to take pathlib.Path object


stscijgbot:
Issue [JP-899](https://jira.stsci.edu/browse/JP-899) was created by Jonathan Eisenhamer:

{{Step.from_config_file}} cannot take a {{pathlib.Path}} object.

jdavies-st:
Related to #2892.  Basically the whole `jwst` package doesn't consistently handle `pathlib.Path` objects.

melanieclarke:
Transferring to stdatamodels.  The associated JP ticket is still open, with low priority.

PR #5345, opened 2020-09-16
Allow PurePath objects to initialize/open datamodels

### LINKED ISSUE #2892: datamodels should accept a pathlib.Path object as init

stscieisenhamer:
Just because this has been annoying enough.

Resolves #2892

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -5,6 +5,7 @@ Data model class heirarchy
 import copy
 import datetime
 import os
+from pathlib import PurePath
 import sys
 import warnings
 
@@ -170,7 +171,9 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
             asdffile = fits_support.from_fits(init, self._schema, self._ctx,
                                               **kwargs)
 
-        elif isinstance(init, (str, bytes)):
+        elif isinstance(init, (str, bytes, PurePath)):
+            if isinstance(init, PurePath):
+                init = str(init)
             if isinstance(init, bytes):
                 init = init.decode(sys.getfilesystemencoding())
             file_type = filetype.check(init)
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,11 @@ env:
     - TEST_COMMAND="pytest --cov-report=xml --cov=./"
     - PIP_DEPENDENCIES=.[test]
 
+cache:
+  - pip
+  - directories:
+    - /tmp/crds_cache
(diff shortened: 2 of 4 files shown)


hbushouse:
Can someone enlighten this python neophyte as to just what a `PurePath` object is and what's the use case in the jwst pipeline driving this change?

stscieisenhamer:
The module is the [pathlib](https://docs.python.org/3/library/pathlib.html) module, which provides a more object-oriented approach to handling file and path names. For a full discussion, [Trey Hunner](https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/) has a bit to say on the subject.

jdavies-st:
Not just you @stscieisenhamer .  It resolves an issue I filed a while back.  Updated your message up top.

There's also #3864

jwst/datamodels/model_base.py
```diff
@@ -170,7 +171,9 @@ def __init__(self, init=None, schema=None, memmap=False,
             asdffile = fits_support.from_fits(init, self._schema, self._ctx,
                                               **kwargs)
 
-        elif isinstance(init, (str, bytes)):
+        elif isinstance(init, (str, bytes, PurePath)):
+            if isinstance(init, PurePath):
+                init = str(init)
```
jdavies-st, line 176: So this is just to allow `Path` input, but internally, everything is still a path string?

Is it worth trying to use `Path` objects throughout the package and converting string paths into `Path` objects?  Or is that too much to bite off for now?

jdavies-st (approved):
Looks good as is.  See comment below.

jwst/datamodels/model_base.py
stscieisenhamer: It probably is worth it, but that should have a design discussion and an issue to update the internals.

Practically, I did start down the rabbit-hole but, on the third module, made the executive decision to abort. The goal was to allow the user-facing API to handle Path. Updating the internals is a different task.

jdavies-st: Agree.  Going through this exercise though would catch instances where we are assuming filesystems are unix (vs Windows), and would allow us to support Windows with simpler code.  See #4996.

PR #5347, opened 2020-09-17
jp-1690 calspec3  cube_build clean up 

jemorrison:
Clean up of making single cubes and blotted cubes.
1.   If making SINGLE type check the input file sent to cube_build and only use those of the correct channel/grating.
2.  When blotting a median IFUcube back to the detector look at the channel or grating and only blot those files that cover the same channel/grating.
3. If all the data  on the input file has been flagged as DO_NOT_USE (a by product of outlier rejection) skip the file in creating the ifucube.

### DIFF
--- a/jwst/cube_build/blot_cube_build.py
+++ b/jwst/cube_build/blot_cube_build.py
@@ -101,7 +101,7 @@ class CubeBlot():
         """ Prints the basic paramters of the blot image and median sky cube
         """
         log.info('Information on Blotting')
-        log.info('Working with instrument %s ', self.instrument)
+        log.info('Working with instrument %s', self.instrument)
         log.info('shape of sky cube %f %f %f',
                  self.median_skycube.data.shape[2],
                  self.median_skycube.data.shape[1],
--- a/jwst/cube_build/blot_cube_build.py
+++ b/jwst/cube_build/blot_cube_build.py
@@ -41,21 +41,23 @@ class CubeBlot():
         # Pull out the needed information from the Median IFUCube
         self.median_skycube = median_model
         self.instrument = median_model.meta.instrument.name
-        self.detector = median_model.meta.instrument.detector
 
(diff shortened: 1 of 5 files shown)


jwst/cube_build/blot_cube_build.py
```diff
@@ -80,23 +82,31 @@ def __init__(self, median_model, input_models):
...
+                self.input_models.append(model)
     # **********************************************************************
 
     def blot_info(self):
         """ Prints the basic paramters of the blot image and median sky cube
         """
         log.info('Information on Blotting')
-        log.info('Working with instrument %s %s', self.instrument,
-                 self.detector)
+        log.info('Working with instrument %s ', self.instrument)
```
hbushouse, line 104: Minor nit, but could remove unused white space after %s

jwst/cube_build/ifu_cube.py
```diff
@@ -638,18 +642,28 @@ def build_ifucube(self):
 
                     coord1, coord2, wave, flux, err, slice_no, rois_pixel, roiw_pixel, weight_pixel,\
                         softrad_pixel, scalerad_pixel, alpha_det, beta_det = pixelresult
+                    # check that there is valid data returned
+                    # If all the data is flagged as DO_NOT_USE - not common then log warning and skip data
+                    nn = wave.size
+                    no_data = False
+                    if nn == 0:
```
hbushouse, line 649: Is `nn` used anywhere else? If not, why not just consolidate these lines into:
```
if wave.size == 0:
    no_data = True
else:
    no_data = False
```

hbushouse (commented):
Looks OK overall, just a couple minor comments. Oh, and of course it needs a change log entry eventually.

hbushouse (approved):
Looks OK to me. Still needs a change log entry before merging.

PR #5353, opened 2020-09-22
JP-1668: Fix bug in extract_1d target coords

### LINKED ISSUE #5301: NIRSpec MOS level 3 extract_1d not centering correctly
_Issue [JP-1668](https://jira.stsci.edu/browse/JP-1668) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

When processing a set of simulated MOS data through calwebb_spec3, the extract_1d extraction aperture is not being centered at the location of the source for any of the sources in the parent exposures.  I see a warning in each case, like this:

2020-09-01 17:09:32,414 - stpipe.Spec3Pipeline.extract_1d - INFO - Turning on source position correction for exp_type = NRS_MSASPEC

2020-09-01 17:09:32,430 - stpipe.Spec3Pipeline.extract_1d - WARNING - WCS implies the target is at -1129.17, which is outside the bounding box,

2020-09-01 17:09:32,430 - stpipe.Spec3Pipeline.extract_1d - WARNING - so we can't get spectrum location using the WCS

2020-09-01 17:09:32,431 - stpipe.Spec3Pipeline.extract_1d - INFO - Using extraction limits: xstart=0, xstop=1297, ystart=10.0, ystop=14.0

I've checked the WCS in the s2d product for one of the sources, and both the source RA and DEC (as taken from the source_ra/dec attributes in the s2d data model) line up exactly with the source spectral trace (see attachment, which is the source corresponding to the above warning messages).

The extraction does seem to be working correctly at level 2b (the x1d products look great!), so whatever the issue is here must be specific to the level 3 version of the step.

hbushouse:
Updated the `get_target_coordinates` function in `extract_1d` step so that when the input model is a `SlitModel` it gets the target coords from the slit source_ra and source_dec attributes, instead of defaulting to the generic TARG_RA/TARG_DEC values (which for something like NIRSpec MOS is going to be very wrong).

Fixes #5301 / [JP-1668](https://jira.stsci.edu/browse/JP-1668)

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -1204,15 +1204,23 @@ class ExtractBase(abc.ABC):
         targ_dec : float or None
             The declination of the target, or None
         """
-        if slit is None:
-            targ_ra = input_model.meta.target.ra
-            targ_dec = input_model.meta.target.dec
-        else:
+        if slit is not None:
+            # If we've been passed a slit object, get the RA/Dec
+            # from the slit source attributes
             targ_ra = getattr(slit, 'source_ra', None)
             targ_dec = getattr(slit, 'source_dec', None)
             if targ_ra is None or targ_dec is None:
                 targ_ra = input_model.meta.target.ra
                 targ_dec = input_model.meta.target.dec
+        elif isinstance(input_model, datamodels.SlitModel):
+            # If the input model is a single SlitModel, again
+            # get the coords from the slit source attributes
+            targ_ra = getattr(input_model, 'source_ra', None)
+            targ_dec = getattr(input_model, 'source_dec', None)
+        else:
+            # Otherwise get it from the generic target coords
+            targ_ra = input_model.meta.target.ra
+            targ_dec = input_model.meta.target.dec
 
         if targ_ra is None or targ_dec is None:
             log.warning("Target RA and Dec could not be determined")
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -1204,24 +1204,26 @@ class ExtractBase(abc.ABC):
         targ_dec : float or None
             The declination of the target, or None
         """
+        targ_ra = None
+        targ_dec = None
+
         if slit is not None:
             # If we've been passed a slit object, get the RA/Dec
             # from the slit source attributes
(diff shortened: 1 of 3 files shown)


jwst/extract_1d/extract.py
```diff
@@ -1204,15 +1204,23 @@ def get_target_coordinates(
...
-        else:
+        if slit is not None:
+            # If we've been passed a slit object, get the RA/Dec
+            # from the slit source attributes
             targ_ra = getattr(slit, 'source_ra', None)
             targ_dec = getattr(slit, 'source_dec', None)
             if targ_ra is None or targ_dec is None:
                 targ_ra = input_model.meta.target.ra
                 targ_dec = input_model.meta.target.dec
+        elif isinstance(input_model, datamodels.SlitModel):
```
stscieisenhamer, line 1215: Since SlitModel also uses `getattr`, the situation that either are `None` still exists and should have the same check as the `slit` case. Suggested code:
```
targ_ra = None
targ_dec = None
if slit is not None:
    targ_ra = getattr...
    targ_de = getattr...
elif isinstance(input_model....
   targ_ra = getattr...
   targ_dec = getattr...
if targ_ra is None or targ_dec is None:
    targ_ra = input_model...
    targ_dec = input_model...
```

hbushouse: Fixed in latest commit.

stscieisenhamer (approved):

jdavies-st (approved):
LGTM.

I'll say that the calling sequence of `get_target_coordinates()` is very odd.  The optional input `slit` is actually contained in the primary input `input_model`, so really one is a copy of a portion of the other when it is a `MultiSlitModel`.  ❓ 

Otherwise, these changes look good.

hbushouse:
Yep, there are several low-level functions that use that kind of argument list: (input_model, slit, ...). I think because there needed to be a way for the functions to distinguish what they should be working on: if they get the entire input_model, which is often a MultiSlitModel, they don't know which slit instance to work on, so the specific slit instance is also passed. Meanwhile, there are cases where what needs to be worked on is the input_model itself, in the form of a simple ImageModel. This whole steps needs an enema.

PR #5354, opened 2020-09-23
JP-1541 Implement memory check in resample to prevent huge arrays

### LINKED ISSUE #5081: Prevent resample and outlier_detection from trying to create ludicrously big mosaics
_Issue [JP-1541](https://jira.stsci.edu/browse/JP-1541) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

There are some error cases in DMS tests where either missing or bad engineering data results in coordinates varying wildly from image to image within an association slated for mosaic combination. This results in the calwebb_image3 resample step trying to create an insanely large output frame for the mosaic and going through many failure/retry cycles of trying to allocate memory for it, ultimately crashing.

DMS management would like to try to prevent the crashing by having some sort of sanity check put into the resample step that will trap attempts to create an output larger than some reasonable threshold and shutdown cleanly with a clear error message. Examples of failures have been trying allocate 10's of TiB of memory, so this gives us an idea as to where the threshold should be set.

stscieisenhamer:
Resolves [JP-1541](https://jira.stsci.edu/browse/JP-1541)
Resolves #5081 

A predictive memory check is implemented in the `resample` module. The steps `OutlierDetectionStep` and `ResampleStep`, both of which have the potential of building very large images, make use of this check through the parameter `allowed_memory`. The value represents the fraction of available memory to use in instantiating the output data model. `None` means do not check.

If the output is too large, steps will raise `OutputTooLargeError` exception. This exception is subclassed from `RuntimeError`.

There is also an environmental defined, DMODEL_ALLOWED_MEMORY, that is queried if the step parameter is `None`.

### DIFF
--- a/jwst/datamodels/util.py
+++ b/jwst/datamodels/util.py
@@ -518,29 +518,6 @@ def check_memory_allocation(shape, allowed=100, model_type=None, include_swap=Tr
     -------
     can_instantiate, required_memory : bool, number
         True if the model can be instantiated and the predicted memory footprint.
-
-    Examples
-    --------
-    >>> check_memory_allocation((1, 1))
-        (True...
-
-    >>> check_memory_allocation((1e10, 1e10))
-        (False...
-
-    >>> check_memory_allocation((1e10, 1e10), allowed=None)
-        (True...
-
-    >>> with_swap = get_available_memory(True)
-    >>> without_swap = get_available_memory(False)
-    >>> dim_size = int((with_swap / 4)**0.5) - 100
-    >>> check_memory_allocation((dim_size, dim_size), allowed=100)
-        (True,...
-
-    >>> check_memory_allocation((dim_size, dim_size), allowed=1)
-        (False,...
-
-    >>> check_memory_allocation((dim_size, dim_size), allowed=100, include_swap=False)
-        (False,...
     """
     # Create the unit shape
     unit_shape = (1,) * len(shape)
--- a/jwst/datamodels/util.py
+++ b/jwst/datamodels/util.py
@@ -522,7 +522,7 @@ def check_memory_allocation(shape, allowed=None, model_type=None, include_swap=T
     """
     # Determine desired allowed amount.
     if allowed is None:
-        allowed = os.environ.get('ALLOWED_MEMORY', None)
+        allowed = os.environ.get('DMODEL_ALLOWED_MEMORY', None)
         if allowed is not None:
             allowed = int(allowed)
 
--- a/jwst/datamodels/util.py
+++ b/jwst/datamodels/util.py
@@ -6,6 +6,8 @@ import sys
 import warnings
 import os
 from os.path import basename
+import psutil
+from psutil._common import bytes2human
 
 import numpy as np
 from astropy.io import fits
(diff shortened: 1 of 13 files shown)


jwst/datamodels/util.py
```diff
@@ -6,6 +6,11 @@
 import warnings
 import os
 from os.path import basename
+from platform import system as platform_system
+from plistlib import loads as plist_loads
+import psutil
+from psutil._common import bytes2human
```
eslavich, line 12: This is pretty short, self-contained function:

https://psutil.readthedocs.io/en/latest/#bytes-conversion

Maybe worth vendorizing so we don't have to worry about that underscore module changing?

jwst/datamodels/util.py
```diff
@@ -6,6 +6,11 @@
 import warnings
 import os
 from os.path import basename
+from platform import system as platform_system
+from plistlib import loads as plist_loads
+import psutil
```
eslavich, line 11: Should `psutil` be added to `install_requires` in setup.cfg?

jwst/datamodels/util.py
```diff
@@ -488,3 +493,157 @@ def get_envar_as_boolean(name, default=False):
...
+    -------
+    available : number
+        The amount available.
+    """
+    vm_stats = psutil.virtual_memory()
+    available = vm_stats.available
+    if include_swap:
+
+        # Attempt to determine amount of free disk space on the boot partition.
+        run_info = run(['/usr/sbin/bless', '--getBoot', '--plist'], capture_output=True)
```
eslavich, line 637: This is the result on my laptop (macOS 10.15.6):

```
$ /usr/sbin/bless --getBoot --plist
Can't access "efi-boot-device" NVRAM variable
```

jwst/datamodels/util.py
```diff
@@ -488,3 +493,157 @@ def get_envar_as_boolean(name, default=False):
...
+        boot_volume = plist_loads(run_info.stdout)['Boot Volume']
+        partitions = psutil.disk_partitions()
+        for partition in partitions:
+            if partition.device == boot_volume:
+                swap = psutil.disk_usage(partition.mountpoint).free
```
eslavich, line 642: Maybe leave 10% or so of the boot volume for temp files or whatever else it needs?

jwst/resample/resample.py
```diff
@@ -54,6 +58,14 @@ def __init__(self, input_models, output=None, **pars):
         # Define output WCS based on all inputs, including a reference WCS
         self.output_wcs = resample_utils.make_output_wcs(self.input_models)
         log.debug('Output mosaic size: {}'.format(self.output_wcs.data_size))
+        can_allocate, required_memory = datamodels.util.check_memory_allocation(
+            self.output_wcs.data_size, pars['allowed_memory'], datamodels.ImageModel
+        )
+        if not can_allocate:
+            raise RuntimeError(
```
eslavich, line 65: I'd recommend a dedicated exception class for this, so we're not converting all `RuntimeError` to SKIP

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -163,13 +165,20 @@ def process(self, input_data):
 
             # Set up outlier detection, then do detection
             step = detection_step(self.input_models, reffiles=reffiles, **pars)
-            step.do_detection()
+            try:
+                step.do_detection()
+            except RuntimeError as exception:
+                self.log.error(f'{exception}')
+                state = 'SKIP'
```
eslavich, line 172: The issue description claims that the desired behavior is to "shutdown cleanly with a clear error message".  Skipping with a log message seems more hazardous, since a user might not notice that the step had been skipped entirely.

jwst/datamodels/util.py
stscieisenhamer: With the default being "no check", and otherwise the user having "free reign" on how they want to manage their memory, I have a slight preference for not having any hidden padding in there. But not wedded to either way.

jwst/resample/resample.py
stscieisenhamer: Though a comment on this: the stpipe architecture has some sort of global try...except thing. I will see what that is really supposed to be doing. Otherwise, agreed.

jwst/outlier_detection/outlier_detection_step.py
stscieisenhamer: Either way. Part of the point of stpipe is to do things like this, and there are other cases. So, the user should expect such things. However, the community "we" have not been diligent on this point, so other reviewers chime in for further opinion.

The commentary may be a bit optimistic though. :-)

jwst/datamodels/util.py
stscieisenhamer: Go figure. Well, delving more around, it seems a more reliable method is using `/private/var/vm`. Will do that and avoid being "bless"ed.

jwst/outlier_detection/outlier_detection_step.py
stscieisenhamer: @hbushouse A statement you just made in the integration meeting probably defines what should happen: If science processing cannot continue, the pipeline will raise an exception. Paraphrased anyways. Seems like this situation would count as "processing cannot continue". As such, the exception will not be caught, though sufficient error messages should make the situation clear.

jwst/resample/resample.py
```diff
@@ -1,6 +1,7 @@
 import logging
 from collections import OrderedDict
 import numpy as np
+from psutil._common import bytes2human
```
eslavich, line 4: This is still using the old `bytes2human`

jwst/resample/resample.py
```diff
@@ -11,7 +12,11 @@
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
-__all__ = ["ResampleData"]
+__all__ = ["OutputTooLargeError", "ResampleData"]
+
+
+class OutputTooLargeError(RuntimeError):
+    """Raised when the output is too large for in-memory instanitation"""
```
eslavich, line 19: Typo in `instanitation`

jwst/datamodels/util.py
eslavich: That makes sense, fine by me.

eslavich (approved):
LGTM!

jwst/lib/basic_utils.py
```diff
@@ -2,6 +2,41 @@
...
+
+    Examples
+    --------
+    >>> bytes2human(10000)
+        '9.8K'
+
+    >>> bytes2human(100001221)
+        '95.4M'
+    """
+    symbols = ('K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y')
```
hbushouse, line 29: Wow, you really made this future proof! Can't wait to have enough RAM to utilize "Y"!

hbushouse (approved):
LGTM. Should update the docs for outlier_detection and resample to mention the new param and its use.

jwst/lib/basic_utils.py
stscieisenhamer: Wonder if Amazon can provide that. Wonder what the cost would be!

stscieisenhamer:
Docs are the last thing on the list. Wanted to get a review in to make sure the solution was in the ballpark.

hbushouse (approved):
LGTM

PR #5363, opened 2020-09-27
Do not attempt prefetch on pipelines that are set to not allow prefetch

### LINKED ISSUE #5350: Spec2Pipeline.master_background step causing unneeded precache fetching
_Issue [JP-1712](https://jira.stsci.edu/browse/JP-1712) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Using the given data, an attempt to retrieve the skipped-step `pathloss` reference file is made. If `master_background` is skipped, the pipeline runs as expected.

stscieisenhamer:
Resolves [JP-1712](https://jira.stsci.edu/browse/JP-1712)
Resolves #5350 

If a `Pipeline` has `prefetch_references == False`, do not attempt to prefetch any reference files for the Pipeline, or its steps.

Reviewers: The fix is the one-liner at pipeline.py:93

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -103,6 +103,11 @@ scripts
 
 - Add pointing analysis commands v1_calculate and pointing_summary. [#5311]
 
+stpipe
+------
+
+- Do not attempt prefetch on pipelines that are set to not allow prefetch. [#5363]
+
 ramp_fitting
 ------------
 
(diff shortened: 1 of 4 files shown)


hbushouse (approved):
Needs change log entry.

PR #5367, opened 2020-09-29
JP-1731: Add DO_NOT_USE to pixels with all groups saturated

### LINKED ISSUE #5364: Ramp fitting not marking pixels with no good samples with DQ='DO_NOT_USE'
_Issue [JP-1731](https://jira.stsci.edu/browse/JP-1731) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

A user reports that calwebb_detector1 and calwebb_image2 processing of imaging data with some pixels saturated in all groups of all integrations results in cal products with SCI=0 and DQ='SATURATED', but otherwise no other indication that the pixels should not be used. They should also have DQ='DO_NOT_USE' set (in addition to SATURATED).

hbushouse:
Updated the `ramp_fitting` function `utils.fix_sat_ramps` to add the DQ flag 'DO_NOT_USE' to pixels that have all groups flagged as saturated. Also greatly simplified the whole function to get rid of looping over pixels and just use native `numpy` array indexing.

Fixes #5364 / [JP-1731](https://jira.stsci.edu/browse/JP-1731)

### DIFF
--- a/jwst/ramp_fitting/tests/test_ramp_fit.py
+++ b/jwst/ramp_fitting/tests/test_ramp_fit.py
@@ -331,8 +331,8 @@ class TestMethods:
         np.testing.assert_allclose(slopes[0].data[50, 50], cds_slope, 1e-6)
         #expect SATURATED
         assert slopes[0].dq[50, 51] == 2 # is there a better way to do this test?
-        #expect SATURATED since 1st group is Saturated
-        assert slopes[0].dq[50, 52] == 2 # is there a better way to do this test?
+        #expect SATURATED and DO_NOT_USE, because 1st group is Saturated
+        assert slopes[0].dq[50, 52] == 3 # is there a better way to do this test?
 
     def test_four_groups_oneCR_orphangroupatend_fit(self, method):
         model1, gdq, rnModel, pixdq, err, gain = setup_inputs(ngroups=4,gain=1,readnoise=10)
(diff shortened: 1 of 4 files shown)


jwst/ramp_fitting/tests/test_ramp_fit.py
```diff
@@ -331,8 +331,8 @@ def test_two_groups_fit(self, method):
         np.testing.assert_allclose(slopes[0].data[50, 50], cds_slope, 1e-6)
         #expect SATURATED
         assert slopes[0].dq[50, 51] == 2 # is there a better way to do this test?
-        #expect SATURATED since 1st group is Saturated
-        assert slopes[0].dq[50, 52] == 2 # is there a better way to do this test?
+        #expect SATURATED and DO_NOT_USE, because 1st group is Saturated
+        assert slopes[0].dq[50, 52] == 3 # is there a better way to do this test?
```
stscieisenhamer, line 335: For readability, one can use the `datamodels.dqflags.pixel` or `group` to give meaning to the value `3`.

stscieisenhamer (approved):

jwst/ramp_fitting/tests/test_ramp_fit.py
hbushouse: Updated in latest commit.

PR #5370, opened 2020-09-29
JP-1732: NRS MOS updates for lamp exposures with no sources

### LINKED ISSUE #5365: Update assign_wcs for NIRSpec MOS to allow for empty SOURCE_INFO table
_Issue [JP-1732](https://jira.stsci.edu/browse/JP-1732) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

NIRSpec internal lamp exposures (NRS_LAMP) taken with OPMODE='MSASPEC' used pre-defined MSA configurations (all open, all closed, long-slit, ...) that do not have any source info associated with them, in which case the SOURCE_INFO table in the MSA metadata file will have zero rows in it. Make any necessary updates to the MOS branch of assign_wcs to handle an empty SOURCE_INFO table, in which case all defined slitlets in the SHUTTER_INFO table should be categorized as background slits (as is already done for real background slitlets in science exposures).

hbushouse:
Updates to a couple of steps to properly handle NRS_LAMP exposures in MSASPEC mode, which don't have any sources defined in their MSA metadata files. The `assign_wcs` step was already setup to handle this properly, in that any slitlet without a source assigned to it is classified as a background slitlet, so all lamp exposure slitlets get processed as background type. Added some checks to the new NIRSpec master_background step so that it does not compute or apply the master bkg spectrum when there are no background slits available or no source slits to apply it to (as is the case for lamp exposures). Also a fix to `flat_field` to properly classify NRS_LAMP exposures as spectroscopic type, rather than imaging.

Fixes #5365 / [JP-1732](https://jira.stsci.edu/browse/JP-1732)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,15 @@
 0.17.2 (unreleased)
 ===================
 
+cube_build
+----------
+
+- When making SINGLE type cubes for outlier detection or mrs_imatch data not in the 
+  appropriate channel/grating is skipped [#5347]
+
+- If outlier detection has flagged all the data on a input file as DO_NOT_USE, then
+  skip the file in creating an ifucube [*5347]
+
 extract_1d
 ----------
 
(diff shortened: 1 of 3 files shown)


stscieisenhamer (approved):
obligatory "changelog update" notice.

PR #5371, opened 2020-09-29
JP-1679: Skip serializing None in datamodels

### LINKED ISSUE #5314: datamodels should not serialize null values
_Issue [JP-1679](https://jira.stsci.edu/browse/JP-1679) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

Currently, `datamodels` relies on undocumented behavior in `asdf` to strip elements in the tree that are set to `None` out of the tree before validating and writing the file out.

This behavior on the part of `asdf` is going away.  See

asdf-format/asdf#536
asdf-format/asdf#863

Using the above branch of `asdf`, this is a fairly typical error we get currently on `master`:

```
E           RuntimeError: Traceback (most recent call last):
E             File "/Users/jdavies/dev/jwst/jwst/pipeline/calwebb_spec2.py", line 112, in process
E               result = self.process_exposure_product(
E             File "/Users/jdavies/dev/jwst/jwst/pipeline/calwebb_spec2.py", line 240, in process_exposure_product
E               calibrated = self._process_nirspec_slits(calibrated)
E             File "/Users/jdavies/dev/jwst/jwst/pipeline/calwebb_spec2.py", line 376, in _process_nirspec_slits
E               calibrated = self.srctype(calibrated)
E             File "/Users/jdavies/dev/jwst/jwst/stpipe/step.py", line 453, in run
E               self.save_model(result, idx=idx)
E             File "/Users/jdavies/dev/jwst/jwst/stpipe/step.py", line 878, in save_model
E               output_path = model.save(
E             File "/Users/jdavies/dev/jwst/jwst/datamodels/model_base.py", line 476, in save
E               self.to_fits(output_path, *args, **kwargs)
E             File "/Users/jdavies/dev/jwst/jwst/datamodels/model_base.py", line 587, in to_fits
E               with fits_support.to_fits(self._instance, self._schema) as ff:
E             File "/Users/jdavies/dev/jwst/jwst/datamodels/fits_support.py", line 396, in to_fits
E               _save_from_schema(hdulist, tree, schema)
E             File "/Users/jdavies/dev/jwst/jwst/datamodels/fits_support.py", line 353, in _save_from_schema
E               validator.validate(tree, _schema=schema)
E             File "/Users/jdavies/miniconda3/envs/jwst/lib/python3.8/site-packages/jsonschema/validators.py", line 353, in validate
E               raise error
E           jsonschema.exceptions.ValidationError: None is not one of ['EXTENDED', 'POINT', 'UNKNOWN']
E           
E           Failed validating 'enum' in schema['properties']['meta']['properties']['target']['properties']['source_type_apt']:
E               OrderedDict([('title', 'Source type from APT (point/extended)'),
E                            ('type', 'string'),
E                            ('enum', ['EXTENDED', 'POINT', 'UNKNOWN']),
E                            ('fits_keyword', 'SRCTYAPT'),
E                            ('blend_table', True)])
E           
E           On instance['meta']['target']['source_type_apt']:
E               None
```

Solutions could be

- have `datamodels` itself strip attributes set to `None` out of the tree before validating and serializing, retaining that behavior in `datamodels` and not relying on `asdf` to do so.

- don't strip `None` values at all, and fix all the areas in `jwst` where we assign them, instead using `del <attribute>` instead.  This would work for us, but users might run into issues if they've been relying on this behavior in the past and have used it in their own code.

Related, simply accessing an attribute sets it to `None` and returns it.  So checking if something exists makes it exists and usually with an invalid `None` value that is not explicitly allowed by the schemas.  So this will probably have to change as well.

Perhaps don't create attributes on access?  Just return `None` to preserve the old behavior.  And create the attribute only on assignment.

jdavies-st:
From `asdf 2.8` onwards, `None` will not be stripped from the ASDF tree before serialization.  But `jwst.datamodels` currently depends on this behavior for a variety of reasons.

Lots of pipeline steps populate things in `meta` with `None`.  If we don't strip `None`, it causes a validation error for many of the items in our schemas, because `None` is not an allowed type.  The schema is expecting a string or number, not `None`.

So the solution is to strip any `None` values before save, and avoid populating attributes with `None` internally.

Other clean up:
 - warnings were being emitted by `ModelContainer` for instruments that were not NIRCam.  A useless warning that I put in several years ago to detect malformed NIRCam simulated data.  No longer needed.

HT: @eslavich for way forward and the first commit.  🚀 

Resolves #5314 / [JP-1679](https://jira.stsci.edu/browse/JP-1679)

### DIFF
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -1,7 +1,6 @@
 import copy
 from collections import OrderedDict
 import os.path as op
-import warnings
 import re
 import logging
 
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -348,12 +348,6 @@ class ModelContainer(model_base.DataModel):
                                              ''.join(params[3:6]), params[6]]))
                 model.meta.group_id = group_id
             except TypeError:
-                params_dict = dict(zip(unique_exposure_parameters, params))
-                bad_params = {'meta.observation.'+k:v for k, v in params_dict.items() if not v}
-                warnings.warn(
-                    'Cannot determine grouping of exposures: '
-                    '{}'.format(bad_params)
-                    )
                 model.meta.group_id = 'exposure{0:04d}'.format(i + 1)
 
     @property
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -213,7 +213,8 @@ class ResampleSpecData:
             wavelength_array = np.flip(wavelength_array,axis=None)
 
         pix_to_wavelength = Tabular1D(lookup_table=wavelength_array,
-                                      bounds_error=False, name='pix2wavelength')
+                                      bounds_error=False, fill_value=None,
+                                      name='pix2wavelength')
 
         # Tabular models need an inverse explicitly defined.
         # If the wavelength array is decending instead of ascending, both
@@ -227,7 +228,8 @@ class ResampleSpecData:
             lookup_table = lookup_table[::-1]
         pix_to_wavelength.inverse = Tabular1D(points=points,
                                               lookup_table=lookup_table,
-                                              bounds_error=False, name='wavelength2pix')
+                                              bounds_error=False, fill_value=None,
+                                              name='wavelength2pix')
 
         # For the input mapping, duplicate the spatial coordinate
         mapping = Mapping((spatial_axis, spatial_axis, spectral_axis))
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -390,7 +390,7 @@ class ResampleSpecData:
             output_model.meta.resample.weight_type = self.drizpars['weight_type']
             output_model.meta.resample.pointings = pointings
 
-            # Update mutlislit slit info on the output_model. This is needed
+            # Update slit info on the output_model. This is needed
             # because model.slit attributes are not in model.meta, so the
             # normal update() method doesn't work with them.
             for attr in ['name', 'xstart', 'xsize', 'ystart', 'ysize',
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -213,7 +213,7 @@ class ResampleSpecData:
             wavelength_array = np.flip(wavelength_array,axis=None)
 
         pix_to_wavelength = Tabular1D(lookup_table=wavelength_array,
-                                      bounds_error=False, fill_value=None, name='pix2wavelength')
+                                      bounds_error=False, name='pix2wavelength')
 
         # Tabular models need an inverse explicitly defined.
         # If the wavelength array is decending instead of ascending, both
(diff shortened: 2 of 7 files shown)


jwst/datamodels/container.py
```diff
@@ -348,12 +347,6 @@ def _assign_group_ids(self):
                                              ''.join(params[3:6]), params[6]]))
                 model.meta.group_id = group_id
             except TypeError:
-                params_dict = dict(zip(unique_exposure_parameters, params))
```
eslavich, line 351: What was the issue here?  Not that I'd miss this code, just curious.

jwst/resample/resample_spec.py
```diff
@@ -390,7 +390,9 @@ def do_drizzle(self, xmin=0, xmax=0, ymin=0, ymax=0, **pars):
             output_model.meta.resample.weight_type = self.drizpars['weight_type']
             output_model.meta.resample.pointings = pointings
 
-            # Update mutlislit slit info on the output_model
+            # Update mutlislit slit info on the output_model. This is needed
```
eslavich, line 393: Not your fault, but there's a typo in `mutlislit`

eslavich (approved):
One extremely minor optional comment.  LGTM!

jwst/datamodels/container.py
jdavies-st: Yeah, this is a very unnecessary warning that applies to all data that isn't NIRCam.  So you are cursed to get this warning if you're not using NIRCam data.  Not a good warning.

jwst/resample/resample_spec.py
jdavies-st: Fixed.

hbushouse (approved):

hbushouse:
Sorry, I updated the change log this morning for my own changes in another PR, so now you've got conflicts.

jdavies-st:
No problem Howard.  I've added a simple unit test to this PR as well, to show this works, though clearly this is being tested elsewhere implicitly as well.

jdavies-st:
This PR cannot be merged until astropy 4.0.2 is released which contains this PR:

https://github.com/astropy/astropy/pull/10674

jdavies-st:
So, after discussion with @eslavich and @nden today, it looks like this PR is fine as is and can be merged.  We will only see failures if `asdf 2.8` is released before the fix https://github.com/astropy/astropy/pull/10674 makes it into an astropy release.  Which should not happen, as that will cause all sorts of other problems for packages outside of `jwst`.  So this is good as is, as we will continue to use it with `asdf 2.7.1` but will be ready when `asdf 2.8` is released.

jdavies-st:
It looks like the small changes in resampling I was seeing were not due floating point accuracy issues, but to the change in `Tabular1D` of not passing `fill_value=None`.  I mistakenly thought that was the default value, but the default value is actually NaN.  So I've reverted it back to `fill_value=None`, which causes the tabular model to extrapolate beyond the bounding box if needed.  Running the regression tests again now to make sure everything is good.

PR #5375, opened 2020-09-30
JP-1645 Supress CRDS unknown reftype errors while retrieving step parameter reference files.

### LINKED ISSUE #5251: Failures to get pars reference files in prefetch cause log messages
_Issue [JP-1645](https://jira.stsci.edu/browse/JP-1645) was created on JIRA by [Robert Jedrzejewski](https://jira.stsci.edu/secure/ViewProfile.jspa?name=rij):_

When step pars reference files are requested from CRDS when running a pipeline, log messages are printed if there is no such pars reference file found in CRDS.  Find a way to disable these messages, as there are many and they are distracting.

stscieisenhamer:
Resolves [JP-1645](https://jira.stsci.edu/browse/JP-1645)
Resolves #5251 

Currently, the CRDS database is sparsely populated with respect to all the possible step parameter reftypes. As such, under normal usage, it is expected to not find a requested reftype. As such, suppress CRDS errors unless debugging is in effect.

### DIFF
--- a/jwst/stpipe/tests/test_crds.py
+++ b/jwst/stpipe/tests/test_crds.py
@@ -38,8 +38,8 @@ class CrdsStep(Step):
 
 def test_steppars_error_supression(_jail, caplog):
     """Test supression of CRDS errors while retrieving step parameters"""
-    model = datamodels.ImageModel((10, 10))
-    model.save('image.fits')
+    with datamodels.ImageModel((10, 10)) as model:
+        model.save('image.fits')
 
     args = ['jwst.stpipe.tests.steps.WithDefaultsStep', 'image.fits']
     Step.from_cmdline(args)
--- a/jwst/stpipe/tests/test_crds.py
+++ b/jwst/stpipe/tests/test_crds.py
@@ -6,9 +6,11 @@ import pytest
 
 from astropy.io import fits
 
+import crds
+from jwst import datamodels
+
 from .. import Step
 from ..import crds_client
-import crds
 
 TMP_DIR = None
 TMP_FITS = None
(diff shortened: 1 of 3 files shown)


jwst/stpipe/tests/test_crds.py
```diff
@@ -34,6 +36,17 @@ def process(self, input_file):
         return datamodels.DataModel()
 
 
+def test_steppars_error_supression(_jail, caplog):
+    """Test supression of CRDS errors while retrieving step parameters"""
+    model = datamodels.ImageModel((10, 10))
+    model.save('image.fits')
```
jdavies-st, line 42: This needs to be done in a `with` context manager.  You're leaving an open file here.

jdavies-st (commented):
🎉  It will be nice to not see these error messages from CRDS anymore.

Just a minor comment below.

hbushouse (approved):
Other than what @jdavies-st said, looks OK to me.

jwst/stpipe/tests/test_crds.py
stscieisenhamer: Will do, but technically no file is open, unless the below is a bug:

```
>>> import jwst.datamodels as dm
>>> m = dm.ImageModel((10, 10))
>>> m.save('junk.fits')
    'junk.fits'
>>> m._files_to_close
    []
```

jdavies-st: Interesting!  I was always under the impression that once something was written, it was then tied to the file.  This a nice, positive thing to discover!

jdavies-st (approved):

stscieisenhamer:
Changing the status of the PR to `Draft`. Going to pursue a more preferred solution described in [CRDS-381](https://jira.stsci.edu/browse/CRDS-381).

stscieisenhamer:
This PR has been rendered unnecessary by [CCD_783](https://jira.stsci.edu/browse/CCD-783)

PR #5377, opened 2020-10-01
Updated unit tests to work with code using new algorithm.

dmggh:
These unit tests have been updated to work with the updated ami_analyze algorithm.


hbushouse (commented):
I think the code updates look OK, but there are a couple of minor pep8 fixes needed and it appears that we're having test timeout issues again with Travis. Something seems to be causing them to hang or just run very slowly.

dmggh:
I can't determine the severity of the Travis error. Is this timeout benign ?

jdavies-st:
The timeout is caused by CRDS server problems we are having right now for all `jwst` builds on Travis.  There's still a bug in this PR though:

https://travis-ci.org/github/spacetelescope/jwst/jobs/732019105#L218

dmggh:
It's hard for me to be sure by looking, but this latest Travis failure  ( event #9231.10 from 
https://travis-ci.org/github/spacetelescope/jwst/jobs/732117182 ),  seems to be (somehow) related to importing the optional uncertainties package - a dependency you mentioned earlier.

I agree that the code wouldn't suffer if it was removed.  Do you think that if I merely remove the affected code in leastsqnrm.py (the sole python file it appears in) the build should then succeed ?  I suppose it can also be removed from setup.cfg.  I also agree that poppy, bloated as it is, should be left in for now.

jdavies-st (approved):
Looks good.  I think the Travis CI problems on Friday were system-wide at ST, preventing Travis from accessing CRDS in a timely manner for CI to run the unit tests.

hbushouse (approved):

PR #5385, opened 2020-10-06
JP-1726 Do not allow target acqs to be considered TSO

### LINKED ISSUE #5357: Bug in ASN level-2b rules for identifying TSO exposures
_Issue [JP-1726](https://jira.stsci.edu/browse/JP-1726) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The MIRI LRS slitless TSO observation 00623055 includes a set of TACQ, TACONFIRM, and LRS-SLITLESS (science) exposures. All 3 are tagged with TSOVISIT=T. The TACQ has a pointing type of "target_acq", while the TACONFIRM and LRS-SLITLESS exposures both have a pointing type of "science". The combination of TSOVISIT + science pointing type for the TACONFIRM exposure causes it to be identified by the level-2b ASN Rules as an exposure that should be processed at the per-integration level and hence uses the "rateints" product as input to level-2b processing, resulting in a "calints" level-2b product. A TACONFIRM exposure should instead be processed at the exposure level (rate, cal products), just like the TACQ exposure.

stscieisenhamer:
Resolves JP-1726
Resolves #5357 

Exclude all target acquisition exposures from being processed at integrations.

### DIFF
--- a/jwst/associations/lib/dms_base.py
+++ b/jwst/associations/lib/dms_base.py
@@ -358,6 +358,10 @@ class DMSBaseMixin(ACIDMixin):
         if item['pntgtype'] != 'science':
             return False
 
+        # Target acquisitions are never TSO
+        if item['exp_type'] in ACQ_EXP_TYPES:
+            return False
+
         # Setup exposure list
         all_exp_types = TSO_EXP_TYPES.copy()
         if other_exp_types:


jwst/associations/lib/dms_base.py
```diff
@@ -358,6 +358,10 @@ def is_item_tso(self, item, other_exp_types=None):
         if item['pntgtype'] != 'science':
             return False
 
+        # Target acquisitions are never TSO
+        if item['exp_type'] in ACQ_EXP_TYPES:
+            return False
+
```
hbushouse, line 364: Might as well go ahead and add 'NRS_VERIFY' to the list of `ACQ_EXP_TYPES` for JP-1423 and get it over with now.

hbushouse (approved):
LGTM. Requisite change log reminder ...

ddavis-stsci (approved):

stscieisenhamer:
Actually removing the addition of NRS_VERIFY to the association list. There are ramifications in test that require schema updates. Leaving to its own issue, JP-1423.

PR #5389, opened 2020-10-07
JP-1400: Add pixel_scale_ratio parameter to resample

### LINKED ISSUE #4764: Include all parameters used for resample step in the config file
Issue [JP-1400](https://jira.stsci.edu/browse/JP-1400) was created by Swara Ravindranath:

The config file for resample step only includes some parameters and not all of them that are used in the drizzle step. For instance, the drizzle scale parameter which is the size of the output pixels relative to the input pixels is not listed in the config file. The parameter is default to 1.0, but users would want to drizzle to a finer pixel scale for dithered images to reconstruct the fully-sampled PSF.  This ticket is to include all the parameters used in the resample step to the config file so they are visible the users. The user should be able to change the parameters (eg; change scale= 1.0 to scale = 0.5), and run the pipeline steps on the command line using the updated values.

jdavies-st:
This adds `pixel_scale_ratio` parameter to `ResampleStep` and `ResampleSpecStep` to allow one to specify a finer output grid in the resampled image.  So a ratio of 0.5 will give twice as fine output grid - a 10x10 input image will get resampled to ~20x20 assuming no rotation.

Resolves #4764 / [JP-1400](https://jira.stsci.edu/browse/JP-1400)


hbushouse:
Is this ready for review or is there more work to do?

jdavies-st:
Still a draft PR.  A bit more work to do on it today.

Tracked down a bug (currently on `master`)

https://github.com/spacetelescope/jwst/blob/354074e61b8275d378e35e5e83047e337952ef7f/jwst/resample/resample.py#L155-L163

that scales the flux of the data by a small fractional value, i.e. the ratio of the CDELT1 in the input and input images, which will be slightly different as the input values are taken from the SIAF and the output is the non-distorted scale that is square, i.e. `CDELT1 == CDELT2`.  And this scaling happens down in the drizzle code, but I'm pretty sure we don't want to be scaling flux by anything in the drizzle code.  I.e. we should have `pscale_ratio=1.0` for all cases.  Any spatial scaling between input and output frame should be happening in the output `gwcs.WCS` object, and drizzle should just handle the flux appropriately, i.e. conserve it.  So it would be good to write a test that verifies this is happening properly.

hbushouse:
I would suggest filing a separate ticket (and working a separate PR) for the investigation and potential fix of the scaling issue, since it's independent from the original request to just add user settable parameters.

jwst/resample/gwcs_drizzle.py
```diff
@@ -196,13 +196,10 @@ def add_image(self, insci, inwcs, inwht=None,
...
-                            self.outsci, self.outwht, self.outcon,
-                            expin, in_units, wt_scl,
-                            pscale_ratio=pscale_ratio, uniqid=self.uniqid,
-                            xmin=xmin, xmax=xmax, ymin=ymin, ymax=ymax,
-                            pixfrac=self.pixfrac, kernel=self.kernel,
-                            fillval=self.fillval)
+        dodrizzle(insci, inwcs, inwht, self.outwcs, self.outsci, self.outwht,
+                  self.outcon, expin, in_units, wt_scl, uniqid=self.uniqid,
+                  xmin=xmin, xmax=xmax, ymin=ymin, ymax=ymax,
+                  pixfrac=self.pixfrac, kernel=self.kernel, fillval=self.fillval)
```
jdavies-st, line 202: It's not clear to me why `dodrizzle` actually ever needs to pass `pscale_ratio` to the underlying `tdriz` C function.  All scaling is taken care of by the input and output frame WCS objects.  So I've removed it from the calling signature here and down at the call to `tdriz`.

jwst/resample/resample.py
```diff
@@ -152,22 +159,18 @@ def do_drizzle(self):
...
-                wcslin_pscale = img.meta.wcsinfo.cdelt1
-
                 inwht = resample_utils.build_driz_weight(img,
-                    weight_type=self.drizpars['weight_type'],
-                    good_bits=self.drizpars['good_bits'])
+                    weight_type=self.weight_type,
+                    good_bits=self.good_bits)
                 driz.add_image(img.data, img.meta.wcs, inwht=inwht,
-                        expin=img.meta.exposure.exposure_time,
-                        pscale_ratio=outwcs_pscale / wcslin_pscale)
```
jdavies-st, line 163: This here was a bug, as this ratio was always very close to 1, but never 1.

jwst/resample/gwcs_drizzle.py
hbushouse: It might be worth running this by either @mcara or @stsci-hack to confirm.

jwst/assign_wcs/util.py
```diff
@@ -210,6 +220,8 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
     bounding_box : tuple, optional
         Bounding_box of the new WCS.
         If not supplied it is computed from the bounding_box of all inputs.
+    pscale_ratio : float, optional
+        A scale representing the ratio of pixel scale of input to output.
```
hbushouse, line 224: I think wording "... ratio of input to output pixel scale" would be a little more straightforward for a reader to parse.

jdavies-st:
When merged, it will cause some differences in the imaging resampled results in our regression tests, due to the above bugfix.

jwst/resample/resample_utils.py
```diff
@@ -6,45 +6,42 @@
...
-    if naxes == 3:
-        # THIS BLOCK CURRENTLY ISN"T USED BY resample_spec
-        pass
-    elif naxes == 2:
-        output_wcs = wcs_from_footprints(input_models)
+    if naxes == 2:
+        output_wcs = wcs_from_footprints(input_models, pscale_ratio=pscale_ratio)
         output_wcs.data_size = shape_from_bounding_box(output_wcs.bounding_box)
+    else:
+        raise RuntimeError("Input WCS needs 2 axes in in output frame."
```
hbushouse, line 43: typo: " axes in in"

hbushouse (commented):
Overall looks OK to me. Just a couple minor questions/comments. Does this require updates to the step docs to mention that pixel scale is now a user-settable step param?

nden:
@stsci-hack Please review this.

jwst/assign_wcs/util.py
jdavies-st: Fixed here and elsewhere.  Yes, clearer.

jwst/resample/resample_utils.py
jdavies-st: Fixed.

jdavies-st:
Step docs are now updated.

jwst/resample/gwcs_drizzle.py
```diff
@@ -196,13 +196,10 @@ def add_image(self, insci, inwcs, inwht=None,
         wt_scl = 1.0  # hard-coded for JWST count-rate data
         self.increment_id()
 
-        dodrizzle(insci, inwcs, inwht, self.outwcs,
-                            self.outsci, self.outwht, self.outcon,
-                            expin, in_units, wt_scl,
-                            pscale_ratio=pscale_ratio, uniqid=self.uniqid,
```
stsci-hack, line 202: The 'pscale_ratio' parameter also needs to be removed from the signature of the 'add_image' function (line 117?).

jwst/resample/gwcs_drizzle.py
stsci-hack: The parameter `pscale_ratio` is not necessary at this point, as the full transformation from input to output WCS has already been determined before calling this code when the `self.outwcs` WCS was defined/provided by the user.  So, this actually cleans up the code.

stsci-hack (commented):
Generally, this change seems fine.  I did find a lingering artifact which should be investigated (see other comment).

jwst/resample/gwcs_drizzle.py
hbushouse: Are you - by implication - confirming that `pscale_ratio` should not be used at all in `dodrizzle` (and `add_image`)?

stsci-hack: It already is not used by the code, making this parameter in the function signature unnecessary...if I didn't overlook any use of this parameter in my review, that is. ;-)

stsci-hack:
Please note:  a default plate_scale_ratio (output/input) that is not identically 1 stems from the distortion model and how it is defined.  This does not imply any rescaling of flux, since drizzle is flux-conserving.  The distortion model for many HST modes results in output WCS plate scales which are intentionally a little different from the input reference pixel (before resampling), which is also slightly different from any other pixel in the frame.  The original comment about CDELT1 and the pscale_ratio just did not seem to me to take this information into account.  It, however, was a good idea to clean up the APIs regarding pscale_ratio in general and enable the user to have more flexibility in setting it.

jdavies-st:
Generally what's your view about what the public interface should be to `cdrizzle.tdriz`?  Since all scaling is happening in the pixmap, which is generated by `wcs_in.forward_transform --> wcs_out.backward_transform`, does it even make sense to have that parameter available in the call signature?  I guess this is more of a question for the `drizzle` package than for this package.  =)

All that said, it is available, and it was being used, and the effect of it was not to scale the image spatially as expected, but to scale down the flux.  So clearly we need to understand what the C code is doing and rectify or remove it.

jwst/resample/gwcs_drizzle.py
jdavies-st: Thanks for catching that.  Removed.

hbushouse (approved):

mcara (approved):
LGTM

hbushouse:
Is everyone happy enough with this to merge?

docs/jwst/resample/arguments.rst
```diff
@@ -12,30 +12,36 @@ image.
 
 ``--kernel`` (str, default='square')
   The form of the kernel function used to distribute flux onto the output
-  image.
+  image.  Available kernels are `square`, `gaussian`, `point`, `tophat`, `turbo`,
+  `lanczos2`, and `lanczos3`.
+
+``--pixel_scale_ratio`` (float, default=1.0)
+  Ratio of input to output pixel scale.  A value of 0.5 means the output image
+  would have 4 pixels sampling each input pixel.
```
mcara, line 20: I think something here is contradictory.

> Ratio of input to output pixel scale.

So, if input image has pixels with pixel scale of 0.1'' and output image has a finer grid in which pixels have 0.05'', pixel ratio should be `input/output = 2` but second sentence suggests the opposite: `0.5`. This is not right.

jwst/resample/tests/test_resample_step.py
```diff
@@ -127,18 +120,129 @@ def test_spatial_transform_miri():
...
+
+
+@pytest.mark.parametrize("ratio", [0.5, 0.7, 1.0])
+def test_pixel_scale_ratio_imaging(nircam_rate, ratio):
+    im = AssignWcsStep.call(nircam_rate)
+    result1 = ResampleStep.call(im)
+    result2 = ResampleStep.call(im, pixel_scale_ratio=ratio)
+    assert_allclose(np.array(result1.data.shape), np.array(result2.data.shape) * ratio, rtol=1, atol=1)
+    e1 = 50
+    e2 = int(50 / ratio)
```
mcara, line 247: Again, I think I am confused about the definition of the `pixel_scale_ratio`. The usage here is as if `pixel_scale_ratio=output_pscale/input_pscale`.

jwst/resample/tests/test_resample_step.py
```diff
@@ -127,18 +120,129 @@ def test_spatial_transform_miri():
...
+    assert np.mean(result1.data[e1:-e1,e1:-e1]) == np.mean(result2.data[e2:-e2,e2:-e2])
```
mcara, line 248: This is where I think things are wrong: the flux in each pixel has to be rescaled by `pixel_scale_ratio`. This test shows that the flux of a star (if one were to do photometry on the resampled image) would quadruple. For the very least this is how I would interpret this for HST images. I do not know what if there are any *fundamental* @hbushouse differences between HST and JWST images. To the best of my knowledge, I think JWST images are fundamentally like `FLT * PHOTFLAM`, i.e., instead of counts or count-rates they have units of flux. If so, the test above shows that the flux is not preserved.

mcara (changes requested):
I do not believe flux is preserved in this PR after resampling and it should be.

mcara:
@hbushouse 
> Is everyone happy enough with this to merge?

I would like to understand why pixel values in input and resampled images are the same (as unit test indicates). I believe this is wrong but it is possible I am missing something in my understanding of what pixels in JWST images represent.

I would hold merging for now

mcara:
I think that a more rigorous test would be to simulate a "star" using a simple Gaussian model with a finite base/extension and normalize the total flux of this star in the input image to, e.g., 1 or 100.  *Make sure that* position of the "star" in the input image is such that *it fits completely, after resampling, in the output image*. Then resample it to a finer grid and do aperture photometry on it. Aperture can be the entire image. The total flux of the star in the output image should be the same as in the input image (1 or 100).

mcara:
My comments above are based on my experience with HST images. For example, take an ACS image and set (for simplicity) `EXPTIME=1` and set _all_ pixel values to 1. Then drizzle it using `drizzlepac`'s `AstreoDrizzle()` to a grid that is 1/2 in scale to the input pixel scale. You will get an output (resampled) image with constant value 0.25.

For reference, I find this useful: https://www.cv.nrao.edu/course/astr534/Brightness.html

Now, HST images are in units of electrons (or counts and then we need to multiply by gain but let's forget this detail). So, when multiplied by `PHOTFLAM` one would get units of flux density (erg/cm**2/s/Hz). We kind of can replace here `erg/cm**2` with `Jy` (forget scale). So, roughly, `FLT*PHOTFLAM` has units of `Jy/s/Hz` - flux density.

On the other hand, I am being told that units of JWST images are `Jy/sr` so each pixel must be multiplied by pixel area are to get the total flux. This is very different from FLT images in which allows photometry to be performed directly on them by simply adding all pixels. It appears this wouldn't be possible to do (without extra processing) for JWST because each input pixel has a different pixel scale and intensity was already flat-fielded. I am sure I am missing some details of how things are in JWST and most likely I am so wrong on this, but at least I would like people to review these issues.

This also means that the way photometry is performed on JWST images may be different from how it is done on HST image because now one needs to take into account pixel area of individual pixels. Would "standard" tools like `photutils` work on such images without modification? CC: @larrybradley

jdavies-st:
Thanks for the feedback Mihai.  I agree we should get this sorted out for imaging in this PR.

larrybradley:
@mcara JWST level-3 images are rectified such that each pixel has the same area.  In that case, standard tools (including `photutils`) work fine (the pixel area (in sr) is a constant that gets multiplied to the data).  The source catalog is derived from level-3 data.

In the case where each pixel has a different area (e.g., JWST level-2 products), I think the images would need to be multiplied by a pixel-area map before one could do photometry.  Do level-2 products come with pixel-area maps?

mcara:
Yes, I am aware it will be easy for drizzled images but my concern was mostly about images that are inputs to drizzle - may choose to perform photometry on un-resampled images.

jdavies-st:
@mcara, let's not worry about photometry on unresampled images in this thread.  That's a separate issue, unrelated to my concern here.

Perhaps some input from @karllark would enlighten us here and get us thinking about this correctly.

Say I decide to set `pixel_scale_ratio=0.5`, which means my resampled pixel grid will be double the input.  A 10x10 input image will become 20x20 in the resampled output.

And assume my input image has a uniform flux of 4 across the whole image, and I resample it with the above setting.  Should each input pixel of value 4 map to 4 output pixels of value 1 each?  Or should each pixel have a value of 4 in the output?  What would be expected for imaging where the input image is in MJy/sr?

I'm not thinking about this correctly I think.

Btw, we covered some of the same ground in 

https://jira.stsci.edu/browse/JP-950

but the result of that discussion still left me wanting clarity.

How this should work for 2D spectra in the spatial and spectral directions when the units are MJy/sr but are dispersed (microns?) is also not clear to me.

hbushouse:
And of course substitute "surface brightness" everywhere that the word "flux" appears in the previous comment ...

karllark:
> Perhaps some input from @karllark would enlighten us here and get us thinking about this correctly.
> 
> Say I decide to set `pixel_scale_ratio=0.5`, which means my resampled pixel grid will be double the input. A 10x10 input image will become 20x20 in the resampled output.
> 
> And assume my input image has a uniform flux of 4 across the whole image, and I resample it with the above setting. Should each input pixel of value 4 map to 4 output pixels of value 1 each? Or should each pixel have a value of 4 in the output? What would be expected for imaging where the input image is in MJy/sr?

As the units of the images are surface brightness (i.e., MJy/sr), then the surface brightness should be the same regardless of the pixel scale of the final rectified image.  In other words, 4 across the whole image and pixel_scale_ratio = 0.5 would result in 4 in the resampled output.

larrybradley:
@jdavies-st In your example, I think the rescaled pixels should also have a value of 4 (otherwise, you would be changing the surface brightness).  The key point is the pixel *area* is now smaller (by a factor of 4), so the total flux (e.g., in Jy) should be the same.  `meta.photometry.pixelarea_steradians` needs to account for the new resampled pixel size.

karllark:
> @mcara, let's not worry about photometry on unresampled images in this thread. That's a separate issue, unrelated to my concern here.

Please tag me in the discussion for photometry on the unresampled images.  There are some subtleties what depend on how the photometry is done that will be important to get right.

karllark:
> How this should work for 2D spectra in the spatial and spectral directions when the units are MJy/sr but are dispersed (microns?) is also not clear to me.

There should be no difference between imaging and spectroscopy for resampling images in MJy/sr.  The units are per unit frequency already (part of the Jy unit) for both cases.

hbushouse:
> > How this should work for 2D spectra in the spatial and spectral directions when the units are MJy/sr but are dispersed (microns?) is also not clear to me.
> 
> There should be no difference between imaging and spectroscopy for resampling images in MJy/sr. The units are per unit frequency already (part of the Jy unit) for both cases.

Although keep in mind that not all 2D JWST spectra are in surface brightness. NIRSpec point sources and NIRISS SOSS are in units of flux (Jy). At least I think that's the correct list of modes that are in Jy.

mcara:
I think this PR can be approved then. Although, I still think that the ultimate (unit) test is to get the photometry correctly for the example that I provided above (a finite Gaussian or even something like a top-hat to detect any edge effects). For this it is important to know how to compute flux in the input image and to compare it to the flux measured in the output image.

mcara:
@hbushouse 

> Although keep in mind that not all 2D JWST spectra are in surface brightness. NIRSpec point sources and NIRISS SOSS are in units of flux (Jy). At least I think that's the correct list of modes that are in Jy.

Will those images will be processed with this resample step? If so, I think it should detect image units of input images and then re-scale intensities accordingly.

Also, what would be the output expected from the resample step for NIRSpec point sources and NIRISS SOSS?

hbushouse:
> @hbushouse
> 
> > Although keep in mind that not all 2D JWST spectra are in surface brightness. NIRSpec point sources and NIRISS SOSS are in units of flux (Jy). At least I think that's the correct list of modes that are in Jy.
> 
> Will those images will be processed with this resample step? If so, I think it should detect units of input images and then re-scale intensities accordingly.
> 
> Also, what would the output from the resample step for NIRSpec point sources and NIRISS SOSS expected to be?

Actually, NIRISS SOSS spectra do not get resampled, so we don't have to worry about that case. But NIRSpec point sources do get resampled (by the corresponding resample_spec step) and are in the rectified space of the slit as projected on the sky.

jdavies-st:
So it looks like we need special handling of cases where `BUNIT == "MJy"`, i.e. for NIRSpec FS point sources.  For those cases, we do have to pass this scale parameter all the way down to the `cdrizzle.tdriz` code.  Lovely.

mcara (approved):

jdavies-st:
Issues #5426 and #5427 describe further related work still needed that was not done here.

PR #5392, opened 2020-10-08
Update source_catalog argument list in step docs

hbushouse:
Remove `kernel_xsize` and `kernel_ysize` arguments from `source_catalog` step docs, because those params were removed from the step in a recent update.


larrybradley (approved):
Thanks, @hbushouse

Issue #5381, opened 2020-10-02
Int times table no longer readable with dev astropy


stscijgbot-jp:
_Issue [JP-1740](https://jira.stsci.edu/browse/JP-1740) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

In our CI job using astropy master, we see the following failure when trying to read int times tables:

<https://travis-ci.org/github/spacetelescope/jwst/jobs/732001794#L756-L824>

And I'm pretty sure it is caused by this PR recently merged into astropy master:

astropy/astropy#10768

stscijgbot-jp:
_Comment by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva) on [JIRA](https://jira.stsci.edu/browse/JP-1740?focusedCommentId=489752#comment-489752):_

Here's a minimal failing example:

 
```java
import numpy as np
from jwst import datamodels

times = np.zeros((5,7))
r = datamodels.RampModel(int_times=times)
c = datamodels.CubeModel()
c.int_times = r.int_times```

PR #5394, opened 2020-10-09
Rework handling of NO_SAT_CHECK pixels in saturation step

hbushouse:
Rework the way the saturation step handles pixels flagged with NO_SAT_CHECK in the saturation reference file. There are many occasions where pixels are flagged this way because they never actually go into full-well saturation before hitting the 16-bit A-to-D converter limit of 65535 DN (so no full-well saturation threshold can be determine). This leaves pixels unflagged in the science data, even though they have one or more groups in which the values are pegged at 65535. Without flagging, the ramp_fit step tries to do a fit and results in nonsensical slopes. So this change sets the saturation flagging threshold for NO_SAT_CHECK pixels at 65535, so any that are pegged at that value will get flagged as saturated and hence ramp_fit will ignore them.

This is a follow-on to the saturation-related updates in JP-1731.

### DIFF
--- a/jwst/saturation/saturation.py
+++ b/jwst/saturation/saturation.py
@@ -99,7 +99,7 @@ def do_correction(input_model, ref_model):
     # Save the saturation flags in the output GROUPDQ array
     output_model.groupdq = groupdq
 
-    n_sat = len(np.where(np.any(np.any(groupdq, axis=0), axis=0))[0])
+    n_sat = np.any(np.any(np.bitwise_and(groupdq, sat_flag), axis=0), axis=0).sum()
     log.info(f'Detected {n_sat} saturated pixels')
 
     # Save the NO_SAT_CHECK flags in the output PIXELDQ array
--- a/jwst/saturation/saturation.py
+++ b/jwst/saturation/saturation.py
@@ -1,4 +1,4 @@
-#
+
 #  Module for 2d saturation
 #
 import logging
@@ -13,6 +13,8 @@ import numpy as np
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
+SATURATED = dqflags.pixel['SATURATED']
+NO_SAT_CHECK = dqflags.pixel['NO_SAT_CHECK']
 ATOD_LIMIT = 65535.  # Hard DN limit of 16-bit A-to-D converter
 
 
@@ -40,7 +42,6 @@ def do_correction(input_model, ref_model):
     ramp_array = input_model.data
     nints = ramp_array.shape[0]
     ngroups = ramp_array.shape[1]
-    sat_flag = dqflags.group['SATURATED']
     detector = input_model.meta.instrument.detector
 
     # If NIRSpec IRS2 readout mode was used, create a mask
@@ -65,15 +66,13 @@ def do_correction(input_model, ref_model):
         ref_sub_model.close()
 
     # For pixels flagged in reference file as NO_SAT_CHECK,
-    # set the output dq mask to NO_SAT_CHECK and set the
-    # saturation check threshold to the A-to-D converter limit.
-    wh_sat = np.bitwise_and(sat_dq, dqflags.pixel['NO_SAT_CHECK'])
-    sat_dq[wh_sat == dqflags.pixel['NO_SAT_CHECK']] = dqflags.pixel['NO_SAT_CHECK']
-    sat_thresh[wh_sat == dqflags.pixel['NO_SAT_CHECK']] = ATOD_LIMIT
-    del wh_sat
+    # set the saturation check threshold to the A-to-D converter limit.
+    sat_thresh[np.bitwise_and(sat_dq, NO_SAT_CHECK) == NO_SAT_CHECK] = ATOD_LIMIT
 
-    # Reset NaN values in the saturation threshold array
-    correct_for_NaN(sat_thresh, sat_dq)
+    # Also reset NaN values in the saturation threshold array to the
+    # A-to-D limit and flag them with NO_SAT_CHECK
+    sat_dq[np.isnan(sat_thresh)] |= NO_SAT_CHECK
+    sat_thresh[np.isnan(sat_thresh)] = ATOD_LIMIT
 
     # Loop over integrations and groups, checking for pixel values
     # that are above the saturation threshold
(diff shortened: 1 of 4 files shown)


hbushouse:
Note to reviewers: 90% of the changes are simply cosmetic and an effort to clean up and make the code more readable. The only substantive change is to replace HUGENUM=100000 with ATOD_LIMIT=65535, which becomes the threshold value used for pixels flagged with NO_SAT_CHECK.

jwst/saturation/saturation.py
```diff
@@ -13,89 +13,103 @@
...
-                          groupdq[ints, plane:, :, :])
+                flagarray[:, :] = np.where(ramp_array[ints, group, :, :] >= sat_thresh,
+                                           sat_flag, 0)
+            np.bitwise_or(groupdq[ints, group:, :, :], flagarray,
+                          groupdq[ints, group:, :, :])
 
+    # Save the saturation flags in the output GROUPDQ array
     output_model.groupdq = groupdq
+
+    n_sat = len(np.where(np.any(np.any(groupdq, axis=0), axis=0))[0])
```
hbushouse, line 102: Any of you numpy ninjas out there have a better, more concise way to do this counting of saturated pixels? I just want the number of pixels flagged as saturated in at least 1 integration and 1 group (i.e. I do NOT want to have multiple counts due to the pixel being saturated in more than 1 group or more than 1 integration).

hbushouse: Whoops. Actually that's the wrong computation. It's counting any pixel that has a non-zero flag value in the groupdq array. I only want ones that have the SATURATION flag. Stay tuned for next commit ...

jemorrison:
I think this NO_SAT_CHECK is something what was in the old MIRI saturation files. We have updated our files so we no longer have this confusion. I just don't think the new files have been delivered to CRDS.  We now supply a single value for all the pixels. This single values is determined using only pixels that saturate and it is the minimum values of the saturating pixels.

jwst/saturation/tests/test_saturation.py
```diff
@@ -213,11 +213,11 @@ def test_nans_in_mask(setup_nrc_cube):
     correct_for_NaN(satmap.data, satmap.dq)
```
jdavies-st, line 213: I don't think this needs to be in the test, as this gets done in `do_correction()`.

jwst/saturation/tests/test_saturation.py
```diff
@@ -213,11 +213,11 @@ def test_nans_in_mask(setup_nrc_cube):
...
     output = do_correction(data, satmap)
 
-    # Check that NaN reference value gets reset to HUGE_NUM
+    # Check that NaN reference value gets reset to ATOD_LIMIT
     # Check that reference DQ is set to NO_SAT_CHECK
     # Check that output GROUPDQ is not flagged as saturated
     # Check that output PIXELDQ is set to NO_SAT_CHECK
-    assert satmap.data[500, 500] == huge_num
+    assert satmap.data[500, 500] == ATOD_LIMIT
     assert satmap.dq[500, 500] == dqflags.pixel['NO_SAT_CHECK']
```
jdavies-st, line 221: May want to have a test where one of the input DQ arrays has another bit set and then makes sure output doesn't blow away that bit.  See my comments above.

jdavies-st (commented):
Good cleanup.  See comments below for some more suggestions and a possible bug.

Also, is the limit 65535 or 2**16, i.e. 65536?  I suspect the former because of zero indexing?

jwst/saturation/tests/test_saturation.py
```diff
@@ -195,7 +195,7 @@ def test_nans_in_mask(setup_nrc_cube):
     ngroups = 5
     nrows = 2048
     ncols = 2048
```
jdavies-st, line 197: Also, the test data could be massively scaled down.  20 instead of 2048?  And edit pixel [5,5]?

jwst/saturation/saturation.py
```diff
@@ -13,89 +13,103 @@
...
+    output_model : `~jwst.datamodels.RampModel`
+        Data model with saturation flags set in GROUPDQ array
     """
 
-    ramparr = input_model.data
-    # Was IRS2 readout used?
+    ramp_array = input_model.data
+    nints = ramp_array.shape[0]
+    ngroups = ramp_array.shape[1]
+    sat_flag = dqflags.group['SATURATED']
```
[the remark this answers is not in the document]
hbushouse: Implemented in next commit.

jwst/saturation/saturation.py
```diff
@@ -13,89 +13,103 @@
...
-    correct_for_NaN(satmask, dqmask)
-
-    dq_flag = dqflags.group['SATURATED']
+    # For pixels flagged in reference file as NO_SAT_CHECK,
+    # set the output dq mask to NO_SAT_CHECK and set the
+    # saturation check threshold to the A-to-D converter limit.
+    wh_sat = np.bitwise_and(sat_dq, dqflags.pixel['NO_SAT_CHECK'])
+    sat_dq[wh_sat == dqflags.pixel['NO_SAT_CHECK']] = dqflags.pixel['NO_SAT_CHECK']
+    sat_thresh[wh_sat == dqflags.pixel['NO_SAT_CHECK']] = ATOD_LIMIT
+    del wh_sat
```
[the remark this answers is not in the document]
hbushouse: Looking at the original code, it first found all the places in the ref file DQ array that have NO_SAT_CHECK and then set all those pixels equal to NO_SAT_CHECK. So it wasn't actually making any change to the NO_SAT_CHECK values themselves, but instead was wiping out any other DQ values that might've been set for those pixels. These values eventually then get propagated into the science data PIXELDQ array. I think we should be propagating *all* DQ values from the ref file, so I'm modifying this logic in the way you suggest, by simply removing line 71, because there's no reason to reset values in the array that were used to find the NO_SAT_CHECK flags in the first place.

jwst/saturation/saturation.py
```diff
@@ -13,89 +13,103 @@
...
 
-    nints = ramparr.shape[0]
-    ngroups = ramparr.shape[1]
+    # Reset NaN values in the saturation threshold array
+    correct_for_NaN(sat_thresh, sat_dq)
```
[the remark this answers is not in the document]
hbushouse: Done

jwst/saturation/tests/test_saturation.py
hbushouse: Done

jwst/saturation/tests/test_saturation.py
```diff
@@ -195,7 +195,7 @@ def test_nans_in_mask(setup_nrc_cube):
...
-    huge_num = 100000.
+    ATOD_LIMIT = 65535.
```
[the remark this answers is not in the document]
hbushouse: Done

jwst/saturation/tests/test_saturation.py
hbushouse: Agree. It's been removed.

jwst/saturation/tests/test_saturation.py
hbushouse: Done

jdavies-st (approved):
Looks good!  And your simplified fix is even better.  👍

hbushouse:
> CI failures are unrelated and are due to `astropy 4.0.2` coming out and #5381 not fixed yet.

So *someone* should fix it!

jdavies-st:
Yes, and possibly do a 0.17.2 patch release once it is fixed?

Issue #5376, opened 2020-09-30
Minimum Number of Groups for Jump Detection should be 5 not 3


mwregan2:
The algorithm described in Anderson and Gordon (2011) requires a minimum of five groups to work. This is because the algorithm always rejects the largest difference and then finds the median of the remaining differences. Five groups have four differences yielding 3 differences after the rejection. Three is the minimum number of differences for finding the median if one of the differences is an outlier. 
  The current code will try to find jumps if there are at least 3 groups. This should be changed to be a minimum of five groups.

PR #5396, opened 2020-10-09
Set jump step minimum number of groups to 5

mwregan2:
This is a fix for issue #5376. The minimum number of frames to invoke the twopoint_difference routine was changed to be 5 instead of 3. This is the minimum number for the current algorithm. In the future the algorithm could be enhanced to find jumps in shorter integrations.
Two test cases were added to show that the step is skipped when the number of frames is 4 and that it runs when the number of frames is five.
Additionally, the comments in twopoint_difference were cleaned up by removing obsolete comments.


jwst/jump/jump_step.py
```diff
@@ -30,8 +30,8 @@ def process(self, input):
             tstart = time.time()
             # Check for an input model with NGROUPS<=2
             ngroups = input_model.data.shape[1]
-            if ngroups <= 2:
-                self.log.warning('Can not apply jump detection when NGROUPS<=2;')
+            if ngroups <= 4:
+                self.log.warning('Can not apply jump detection when NGROUPS<=4;')
```
hbushouse, line 34: Can you walk me through the logic again that says we need more than 4 groups? I know you can't detect a jump with only 2 differences (3 groups), but isn't it enough to have 3 differences (4 groups)?

jwst/jump/tests/test_jump_step.py
```diff
@@ -245,3 +245,35 @@ def test_two_group_integration(generate_miri_reffiles, max_cores, setup_inputs):
     out_model = JumpStep.call(model1, override_gain=override_gain,
                               override_readnoise=override_readnoise, maximum_cores=max_cores)
     assert(out_model.meta.cal_step.jump == 'SKIPPED')
+
+@pytest.mark.parametrize("max_cores", MAXIMUM_CORES)
```
hbushouse, line 249: Because this test is simply dealing with the mechanics of skipping when NGROUPS is below the required threshold does it really need to execute using all available settings of MAXIMUM_CORES? Wouldn't just 1 be sufficient? The number of cores should have no effect on the skipping logic.

jwst/jump/tests/test_jump_step.py
```diff
@@ -245,3 +245,35 @@ def test_two_group_integration(generate_miri_reffiles, max_cores, setup_inputs):
...
+    ysize = 102
+    model1, gdq, rnModel, pixdq, err, gain = setup_inputs(ngroups=ngroups,
+        nrows=ysize, ncols=xsize,
+        gain=ingain, readnoise=inreadnoise, deltatime=grouptime)
+    out_model = JumpStep.call(model1, override_gain=override_gain,
+                              override_readnoise=override_readnoise, maximum_cores=max_cores)
+    assert(out_model.meta.cal_step.jump == 'SKIPPED')
+
+@pytest.mark.parametrize("max_cores", MAXIMUM_CORES)
+def test_five_group_integration(generate_miri_reffiles, max_cores, setup_inputs):
```
hbushouse, line 266: We've already got a number of existing tests that use ngroups > 4, so is this one really necessary? Just adds to the total time to execute the tests. It doesn't appear to be looking for anything specific in the results (e.g. which jumps were detected) other than the fact that it simply ran.

jwst/jump/jump_step.py
mwregan2: The algorithm uses the median of the first differences after excluding the largest difference. To get a reliable median you need three samples. So, adding in the difference that is excluded, we need four differences for the algorithm to work. This means we need five frames.
Making a change so that four frames gives a good result would not be hard.  My goal with this change was to make simple change first and at least match the input minimum frames to the algorithm minimum frames. This change is low risk (except for the cases of three and four frame integrations where the expected results could change)

jwst/jump/tests/test_jump_step.py
mwregan2: I agree. I was just following the pattern of the other test cases. I'll change it.

jwst/jump/jump_step.py
hbushouse: Ah, right - I forgot about the exclusion of the largest diff, thus requiring one more.

jwst/jump/tests/test_jump_step.py
mwregan2: The idea is to pair the two test cases for the largest failing number of frames with the smallest working number of frames. By making it explicit that the only thing that changed was the number of frames required, when we change the minimum number of frames. I'll be removing the parameterize for these to test cases making over time increase quite small.

hbushouse (commented):
Looks good now. Just needs two remaining minor updates: Update the step docs (https://github.com/spacetelescope/jwst/blob/master/docs/jwst/jump/description.rst) to change the statement about the min number of groups required, and add an entry to the change log (CHANGES.rst) in the top 0.17.2 section.

hbushouse (approved):

PR #5398, opened 2020-10-12
Add environmental and pytest argument parameters to the Jenkins build parameters

stscieisenhamer:
Adding two new build parameters to the Jenkins build. Neither of these parameters need be defined in the Jenkins build; defaults will be used.

PYTEST_ARGS: A `String` parameter that specifies the additional arguments to the `pytest` command, nominally the list of files to include in the test. If unspecified or blank, all files in the project will be considered for testing. For example, to search for tests in only `jwst/regtest`, set to:

```
jwst/regtest
```

ENV_VARS: A `MultiLine String` specifying a list of environmental variables to set. If unspecified or blank, no extra envrionmental variables are set. For example, to change the CRDS server to use the DEV server:

```
CRDS_PATH=/data1/crds_caches/crds_cache_dev
CRDS_SERVER_URL=https://jwst-crds-dev.stsci.edu
```

Note about the above example: The path is a path on the Jenkins server, currently `jwcalibdev`. Any path writeable by the Jenkins process can be used. The example does show a valid path.

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -17,9 +17,9 @@ if (env.ARTIFACTORY_ENV) {
     artifactoryenv = env.ARTIFACTORY_ENV
 }
 
-test_items = ""
-if (env.TEST_ITEMS) {
-    test_items = env.TEST_ITEMS
+pytest_args = ""
+if (env.PYTEST_ARGS) {
+    test_items = env.PYTEST_ARGS
 }
 
 withCredentials([string(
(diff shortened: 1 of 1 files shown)


hbushouse (approved):
Looks OK to me, but I'll leave it to the Jenkins/test aficionados for real confirmation.

stscieisenhamer:
CI failures unrelated to this PR.

stscieisenhamer:
Note that I cannot configure the JWST or JWST_UNSTABLE jobs. However, the updates will work out-of-box with those jobs since all new parameters are optional.

Have also updated the jwst github wiki.

PR #5401, opened 2020-10-12
JP-1308 coron3 pipeline align_refs updates

bsunnquist:
This PR updates the CORON3 align_refs step's bad pixel correction with the following tweaks as discussed in the comments of JP-1308 (https://jira.stsci.edu/browse/JP-1308):

- find the bad pixel locations using np.bitwise_and() rather than np.equal(), i.e. find all pixels that have a DO_NOT_USE flag set (e.g. 17, 1, 5, …), not just those exactly equal to 1.
- add an option to specify the DQ bits to consider bad like in the example notebook, to gives the teams flexibility to minimize what is smoothed.
- add +1 to the second x,y indexes of the median box, so the box is equal length on either side of the bad pixel (also change the default box size to 3 since this means the box length will always be odd).

I've updated the documentation as well, and tested these updates on our simulated coro data. Let me know if there are any issues.


jwst/coron/median_replace_img.py
```diff
@@ -1,42 +1,57 @@
...
+        Input data quality array
+
+    bsize : scalar
+        Box size of the data to extract
+
+    bad_bitvalue : int
+        The sum of all of the DQ bit values to consider bad. Setting to 0
+        will treat all pixels as good.
+
+    xc: scalar
```
hbushouse, line 29: Really minor nit, but should have a whitespace preceding the colon (i.e. "xc : scalar") to follow conventions. Same in line 32 for `yc`.

hbushouse (approved):
Looks good to me. Just one really minor comment on formatting.

stscieisenhamer (approved):

jwst/coron/median_replace_img.py
bsunnquist: whoops - fixed!

PR #5404, opened 2020-10-13
JP-1735: Reduce jump step memory usage

### LINKED ISSUE #5368: Reduce jump step memory usage
_Issue [JP-1735](https://jira.stsci.edu/browse/JP-1735) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

A significant fraction of the memory errors that occur during calwebb_detector1 processing of TSO exposures in the Ops environment occur in the jump step, when it tries to allocate memory for any one of the large temporary arrays it uses during processing.

One the worst (if not the worst) offenders is the `all_ratios` array, which is created at <https://github.com/spacetelescope/jwst/blob/master/jwst/jump/twopoint_difference.py#L36.> It is a 4-D floating-point array of the same size as the original SCI and ERR arrays, hence consuming a large amount of memory. A quick scan of the code suggests that it may not be necessary to make this array 4-D (covering all integrations in the exposure), but could instead be a 3-D array specific to each integration. This needs to be investigated and updated, if possible.

hbushouse:
Updated the `twopoint_difference` module in the `jump` step to try to eliminate some of the largest memory hogs. The single biggest change with the largest effect on memory usage is moving the flagging of neighbors from the end of the routine (after all integrations have been processed) to being inside the main loop over integrations. This completely eliminates the need for the `all_ratios` array, which was a huge 4-D float array the same size as the 4-D SCI data.

Also eliminated some of the smaller intermediate arrays that were used to do things like storing indexes of bad or saturated pixels. In addition, there's been just a huge amount of cosmetic cleanup to try to make the code more readable and just more efficient overall.

Somehow, these changes actually fixed some previously undetected problems with the wrong groups getting flagged as having a jump (e.g. a jump is detected in group 5 for a pixel, but group 1 ended up getting the DQ flag set). Differences in the regressions show that the new code is producing the correct results in all cases.

Fixes #5368 / [JP-1735](https://jira.stsci.edu/browse/JP-1735)

### DIFF
--- a/jwst/jump/tests/test_detect_jumps.py
+++ b/jwst/jump/tests/test_detect_jumps.py
@@ -506,7 +506,7 @@ def test_proc(setup_inputs):
     ngroups = 10
 
     model1, gdq, rnModel, pixdq, err, gain = \
-        setup_inputs( ngroups=ngroups, nrows=5, ncols=6, nints=2, gain=ingain, readnoise=inreadnoise,
+        setup_inputs( ngroups=ngroups, nrows=25, ncols=6, nints=2, gain=ingain, readnoise=inreadnoise,
                       deltatime=grouptime )
 
     model1.data[0, 0, 2, 3] = 15.0
@@ -520,11 +520,11 @@ def test_proc(setup_inputs):
     model1.data[0, 8, 2, 3] = 170.0
     model1.data[0, 9, 2, 3] = 180.0
 
-    out_model_a = detect_jumps( model1, gain, rnModel, 4.0, 'half', 200, 4, True )
-    out_model_b = detect_jumps( model1, gain, rnModel, 4.0, None, 200, 4, True )
+    out_model_a = detect_jumps( model1, gain, rnModel, 4.0, None, 200, 4, True )
+    out_model_b = detect_jumps( model1, gain, rnModel, 4.0, 'half', 200, 4, True )
     assert( out_model_a.groupdq == out_model_b.groupdq ).all()
 
-    out_model_c = detect_jumps( model1, gain, rnModel, 4.0, 'All', 200, 4, True )
+    out_model_c = detect_jumps( model1, gain, rnModel, 4.0, 'all', 200, 4, True )
     assert( out_model_a.groupdq == out_model_c.groupdq ).all()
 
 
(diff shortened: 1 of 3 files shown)


stscieisenhamer (approved):

jwst/jump/tests/test_detect_jumps.py
```diff
@@ -506,7 +506,7 @@ def test_proc(setup_inputs):
     ngroups = 10
 
     model1, gdq, rnModel, pixdq, err, gain = \
-        setup_inputs( ngroups=ngroups, nrows=5, ncols=6, nints=2, gain=ingain, readnoise=inreadnoise,
+        setup_inputs( ngroups=ngroups, nrows=25, ncols=6, nints=2, gain=ingain, readnoise=inreadnoise,
```
hbushouse, line 509: Reviewers: these couple of tests that employ multi-processing had their image sizes increased a little just to make sure there were enough rows in the image to sensibly divide up between the multiple processes.

jwst/jump/twopoint_difference.py
```diff
@@ -28,65 +32,57 @@ def find_crs(data, group_dq, read_noise, rej_threshold, nframes, flag_4_neighbor
...
-        nan_pixels = np.where(np.isnan(first_diffs))
-        first_diffs[nan_pixels] = 100000.
-
+        # note: roll the ngroups axis of data array to the end, to make
+        # memory access to the values for a given pixel faster.
+        # New form of the array has dimensions [nrows, ncols, ngroups].
+        first_diffs = np.diff(np.rollaxis(data[integ], axis=0, start=3), axis=2)
         positive_first_diffs = np.abs(first_diffs)
 
+        # sat_groups is a 3D array that is true when the group is saturated
```
jdavies-st, line 68: ```# sat_groups is a 3D boolean array that is true ...```

jdavies-st (approved):
Looks good to me.  Easier to read code is always a plus.

How much memory does this save?  Did you profile it at all?

hbushouse:
Travis error for latest commit seems to be an unrelated connection error during a tweakreg test.

PR #5406, opened 2020-10-14
Use Codecov carryforward flags

### LINKED ISSUE #4920: Add Carryforward Codecov flags to repo
This new feature of flags in Codecov

https://blog.codecov.io/2020/04/21/new-test-only-what-you-change-with-carryforward-flags/

is what our original implementation of flags was trying to accomplish.  Let's try implementing this new feature and see if it works.

jdavies-st:
Make the `nightly` tests flag _carryforward_ so that we can see the coverage of the repo from both unit and nightly tests combined.  This should also only compare CI unit tests to previous coverage of CI unit tests, not to coverage of the nightly tests.  🤞 

Also, make threshold be 0.1% so that when a few lines of code are deleted, the `codecov/project` check doesn't fail because the denominator has gotten smaller.

Resolves #4920

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,14 +1,23 @@
 coverage:
   status:
     project:
-      default: off
+      default:
+        target: auto
+        threshold: 0.1%
+        flags:
+          - unit
+          - nightly
       unit:
-        flags: unit
+        target: auto
+        flags:
+          - unit
       nightly:
-        flags: nightly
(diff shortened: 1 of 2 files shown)


hbushouse (approved):

jdavies-st:
Oops, forgot turn turn on reporting for the nightly tests.  Done now.

PR #5411, opened 2020-10-16
Simplify test setup fixture in jump detection

jdavies-st:
Follow on from #5404 and #5412.

Cleans up `setup_inputs` test fixture in `test_detect_jumps.py` to only return what the tests need, and to reduce redundancy with defaults already in use in the runtime code.

### DIFF
--- a/jwst/jump/tests/test_detect_jumps.py
+++ b/jwst/jump/tests/test_detect_jumps.py
@@ -34,10 +34,12 @@ def test_nocrs_noflux_badgain_pixel(setup_inputs):
     gain.data[7, 7] = -10  # bad gain
     gain.data[17, 17] = np.nan  # bad gain
     out_model = detect_jumps(model, gain, rnoise, 4.0,  1, 200, 4, True)
-    assert np.bitwise_and(out_model.pixeldq[7, 7], NO_GAIN_VALUE)
-    assert np.bitwise_and(out_model.pixeldq[7, 7], DO_NOT_USE)
-    assert np.bitwise_and(out_model.pixeldq[17, 17], NO_GAIN_VALUE)
-    assert np.bitwise_and(out_model.pixeldq[17, 17], DO_NOT_USE)
+
+    # 2 bits are set for each pixel, so use bitwise_and to check is set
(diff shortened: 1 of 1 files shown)


jwst/jump/tests/test_detect_jumps.py
```diff
@@ -354,79 +351,78 @@ def test_crs_on_edge_with_neighbor_flagging(setup_inputs):
...
     assert (out_model.groupdq[0, 5, 5, -1] == 0)# The one not to flag
     # flag CR and three neighbors of last column CR
-    assert (4 == out_model.groupdq[0, 5, 15, 1027])
-    assert (4 == out_model.groupdq[0, 5, 15, 1026])
-    assert (4 == out_model.groupdq[0, 5, 16, 1027])
-    assert (4 == out_model.groupdq[0, 5, 14, 1027])
+    assert (4 == out_model.groupdq[0, 5, 15, -1])
+    assert (4 == out_model.groupdq[0, 5, 15, -2])
+    assert (4 == out_model.groupdq[0, 5, 16, -1])
+    assert (4 == out_model.groupdq[0, 5, 14, -1])
```
jdavies-st, line 425: It's not clear to me the original test was testing this properly.  The input array size is

```
nrows = 1024
ncols = 1032
```

but the checks were actually not being made at the edge on columns 1030 and 1031, but in columns 1026 and 1027.  I suspect these are reference pixels, but these tests don't know anything about reference pixels.

For the rows, the check was being done correctly, at row 1022 and 1023.

Regardless, both tests pass in the new version of the test.

hbushouse: Yes, the last few cols are ref pixels, hence the tests were being done in a few cols from the edge. But then that means they aren't really "edge" pixels, because the data array that's being examined by the jump step includes the extra cols of ref pixels.

hbushouse:
Ah crud. I didn't know you were working on this, so now we have competing PR's (#5412). That'll be fun to try to resolve conflicts!

jdavies-st:
Feel free to merge #5412 first.  Happy to rebase this one.

hbushouse (approved):

PR #5412, opened 2020-10-16
Add another jump unit test

hbushouse:
Added another new jump unit test based on some saturated data that was getting handled incorrectly before #5404 was merged. Also did a lot of general cleanup to the test module and reduced the sizes of many tests, since typically only a few pixel values are being tested anyway.

### DIFF
--- a/jwst/jump/tests/test_detect_jumps.py
+++ b/jwst/jump/tests/test_detect_jumps.py
@@ -8,21 +8,23 @@ from jwst.jump.jump import detect_jumps
 import multiprocessing
 from jwst.datamodels import dqflags
 
+
 def test_nocrs_noflux(setup_inputs):
     """"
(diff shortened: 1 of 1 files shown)


jwst/jump/tests/test_detect_jumps.py
```diff
@@ -215,6 +234,51 @@ def test_multiple_neighbor_jumps_firstlastbad(setup_inputs):
     assert_array_equal(out_model.groupdq[0,:,4,4], [1, 0, 0, 0, 0, 0, 0, 0, 0, 1])
 
 
+def test_nirspec_saturated_pix(setup_inputs):
```
hbushouse, line 237: This is the actual new test. All other changes are just cleanup and cosmetics (and some pep8 fixes).

jdavies-st (approved):
👏

PR #5413, opened 2020-10-19
JP-1748 Remove logging from DataModel.close

stscieisenhamer:
Resolves followup issue with JP-1748.

Since `DataModel.close` is called from `DataModel.__del__`, a `DataModel` that is still open when a Python process closes, it is possible that the logging module is no longer available, resulting in the error

```
ImportError: sys.meta_path is None, Python is likely shutting down
```

Since logging at this point is normally not required, simply remove the logging statement.

### DIFF
--- a/jwst/datamodels/model_base.py
+++ b/jwst/datamodels/model_base.py
@@ -338,7 +338,12 @@ class DataModel(properties.ObjectNode, ndmodel.NDModel):
 
             for fd in self._files_to_close:
                 if fd is not None:
-                    log_with_traceback(f'{__name__}: {self}: Closing {fd}')
+
+                    # Logging is normally disabled since this method is called
+                    # from __del__, which does not guarantee that logging
+                    # is available.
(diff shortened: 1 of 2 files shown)


hbushouse (approved):

jwst/datamodels/model_base.py
```diff
@@ -338,7 +338,12 @@ def close(self):
 
             for fd in self._files_to_close:
                 if fd is not None:
-                    log_with_traceback(f'{__name__}: {self}: Closing {fd}')
```
eslavich, line 341: Another option perhaps is to wrap this in `if sys.meta_path is not None:` and then change the log level to debug?

stscieisenhamer: Did consider an in-code fix, but felt that debugging messages at this level is a bit deep anyways. Debugging at this point is to actually figure out problems in other places, so hopefully not needed often and would crowd out other more useful messages.

PR #5414, opened 2020-10-19
JP-1478: extract_1d enhancements for NRC_TSGRISM

### LINKED ISSUE #5000: Enable use of extract_1d reference file for NRC_TSGRISM data
_Issue [JP-1478](https://jira.stsci.edu/browse/JP-1478) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

The extract_1d step currently does not allow NRC_TSGRISM observations to make use of the extract_1d reference file, which can be used to define background regions and enable column by column background subtraction. Without a column by column background subtraction, the extracted spectrum is much more noisy than it would be otherwise, limiting the transit sensitivity. The lack of background subtraction also means that any residual amplifier-dependent variation in the background is not removed.

hbushouse:
Updates to the extract_1d step to allow for an EXTRACT1D reference file to be used for NRC_TSGRISM mode, so that necessary column-by-column background extraction and subtraction can be performed. Previously the code was hardwired to not use EXTRACT1D ref file params for NRC_TSGRISM mode, which defaulted to very simplistic extraction with no background subtraction.

In addition to enabling the use of the reference file for this mode, a new parameter `bkg_fit` has been added, which now allows the user to specify whether to perform the default polynomial fit to background values or do simple mean and median computations. Functions to perform the mean and median background computations have been added to the code. Note that even when mean/median are used, the background model is still created in the form of a polynomial, but in this case it's a zero-order polynomial with the 0th-order coefficient set to the mean/median. That allows the background model to be used as-is in the downstream extraction functions.

Documentation for the extract_1d step and EXTRACT1D reference file have also been updated to account for the changes and to give more details on how the various extraction parameters are actually used in the code.

Fixes #5000 / [JP-1478](https://jira.stsci.edu/browse/JP-1478)

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -277,7 +277,9 @@ def get_extract_parameters(
         extract_params['src_coeff'] = None
         extract_params['bkg_coeff'] = None  # no background sub.
         extract_params['smoothing_length'] = 0  # because no background sub.
+        extract_params['bkg_fit'] = None  # because no background sub.
         extract_params['bkg_order'] = 0  # because no background sub.
+        extract_params['subtract_background'] = False
 
         if use_source_posn is None:
             extract_params['use_source_posn'] = False
@@ -322,7 +324,9 @@ def get_extract_parameters(
                         shape = input_model.data.shape
                         extract_params['src_coeff'] = None
                         extract_params['bkg_coeff'] = None
+                        extract_params['bkg_fit'] = None
                         extract_params['bkg_order'] = 0
+                        extract_params['subtract_background'] = False
                         extract_params['use_source_posn'] = False
                         extract_params['xstart'] = 0
                         extract_params['xstop'] = shape[-1] - 1
@@ -337,6 +341,9 @@ def get_extract_parameters(
                         if extract_params['bkg_coeff'] is not None:
                             extract_params['subtract_background'] = True
                             extract_params['bkg_fit'] = aper.get('bkg_fit', 'poly')
+                        else:
+                            extract_params['bkg_fit'] = None
+                            extract_params['subtract_background'] = False
                         extract_params['independent_var'] = aper.get('independent_var', 'pixel').lower()
 
                         if bkg_order is None:
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -29,7 +29,8 @@ from .apply_apcorr import select_apcorr
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
-WFSS_EXPTYPES = ['NIS_WFSS', 'NRC_WFSS', 'NRC_GRISM', 'NRC_TSGRISM']
+#WFSS_EXPTYPES = ['NIS_WFSS', 'NRC_WFSS', 'NRC_GRISM', 'NRC_TSGRISM']
+WFSS_EXPTYPES = ['NIS_WFSS', 'NRC_WFSS', 'NRC_GRISM']
 """Exposure types to be regarded as wide-field slitless spectroscopy."""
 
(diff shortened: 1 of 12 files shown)


hbushouse:
By the way, these changes include a fix for another previously undiscovered bug at https://github.com/spacetelescope/jwst/pull/5414/files#diff-8e03eb4c28c8d698c106149c2306a1becf8f1a7c4e4027bf5dfeae9bc76395abR1782. The wrong object was being returned from the function that was setting up background region coeffs, such that even when the extract1d ref file specified background regions, they were not getting used and hence no background extraction or subtraction was ever actually happening. This leads to differences (good ones!) in at least the NIRISS SOSS regression tests, because it has background regions defined in its extract1d ref file.

hbushouse:
OK, the last commit contained a few minor updates to make the new code work for all existing modes and also updated the TSGRISM regression test to use the prototype tsgrism extract1d ref file. That can be reverted once there's a ref file available in CRDS.

This is now ready for review.

hbushouse:
Of course the changes to step docs and code related to anything for IFU data will get superseded in the near future when @jemorrison finishes all the updates for IFU extraction.

stscieisenhamer (approved):

jwst/extract_1d/extract.py
```diff
@@ -266,7 +275,11 @@ def get_extract_parameters(
         extract_params['ystop'] = shape[-2] - 1  # last pixel in Y
         extract_params['extract_width'] = None
         extract_params['src_coeff'] = None
-        extract_params['bkg_coeff'] = None
+        extract_params['bkg_coeff'] = None  # no background sub.
+        extract_params['smoothing_length'] = 0  # because no background sub.
+        extract_params['bkg_fit'] = None  # because no background sub.
+        extract_params['bkg_order'] = 0  # because no background sub.
+        extract_params['subtract_background'] = False
```
jdavies-st, line 282: It looks like this block is not unit tested.  Perhaps a small unit test that just checks that this function works as expected when called with `ref_dict=None`

jwst/extract_1d/extract.py
```diff
@@ -325,6 +338,12 @@ def get_extract_parameters(
                     else:
                         extract_params['src_coeff'] = aper.get('src_coeff')
                         extract_params['bkg_coeff'] = aper.get('bkg_coeff')
+                        if extract_params['bkg_coeff'] is not None:
+                            extract_params['subtract_background'] = True
+                            extract_params['bkg_fit'] = aper.get('bkg_fit', 'poly')
+                        else:
+                            extract_params['bkg_fit'] = None
+                            extract_params['subtract_background'] = False
```
jdavies-st, line 346: Same here.  Not unit tested, but perhaps a small test could target some of this.

jdavies-st (approved):
This looks good.

Some specific comments below about test coverage.  The `extract.py` module is only unit tested at the 9% level, which is pretty crazy given it is 3791 lines long.  Would be nice to toss in a unit test on the areas you touched so we can start getting some coverage.

jwst/extract_1d/extract.py
hbushouse: Agreed. Fortunately it is at least tested by 1 or 2 regression tests, specifically the WFSS mode, which does not use an extract1d ref file and hence the `ref_dict` is None.

jwst/extract_1d/extract.py
hbushouse: Again, this is at least tested in the regression tests, where the NIRISS SOSS extract1d ref file specifies `bkg_coeff` values, but none of the other spectral modes do, so both branches do get tested in regtest. But unit test coverage certainly couldn't hurt.

PR #5420, opened 2020-10-21
Split up some long-running regtest modules

hbushouse:
Splitting up some of the regtest modules that contain both spec2 and spec3 tests for MIRI and NIRSpec IFU modes, which are notoriously slow to run. Segregated the spec2 and spec3 tests into their own modules and also put the NIRSpec IFU spec2 step-specific tests into their own module (separate from the generic spec2 pipeline test).

### DIFF
--- a/jwst/regtest/test_miri_mrs_spec3.py
+++ b/jwst/regtest/test_miri_mrs_spec3.py
@@ -1,0 +1,123 @@
+"""Regression tests for MIRI MRS modes"""
+import os
+import pytest
+
+from . import regtestdata as rt
+
+from astropy.io.fits.diff import FITSDiff
+from jwst.stpipe import Step
+from jwst.pipeline.collect_pipeline_cfgs import collect_pipeline_cfgs
+from jwst.associations.asn_from_list import asn_from_list
+
+# Define artifactory source and truth
+INPUT_PATH = 'miri/mrs'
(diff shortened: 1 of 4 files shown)


jwst/regtest/test_miri_mrs_spec3.py
```diff
@@ -0,0 +1,123 @@
...
+
+@pytest.fixture(scope='module')
+def run_spec3(jail, rtdata_module):
+    """Run the Spec3Pipeline on the single cal result from the Spec2Pipeline run"""
+    rtdata = rtdata_module
+
+    collect_pipeline_cfgs("config")
+
+    # Note that we use the truth file from spec2 processing as the input to spec3
+    rtdata.get_data(TRUTH_PATH + '/' + 'ifushort_ch12_cal.fits')
```
hbushouse, line 25: Note to reviewers: the previous tests ran spec2 and spec3 in sequence, such that this spec3 test would then use the spec2 result directly as input. Because they're now in separate modules we can't use the spec2 result directly, so instead I've set it up to use the truth file from the spec2 test. So at least whenever spec2 results change and the truth file gets updated, it should then flow into the spec3 test.

stscieisenhamer (approved):

jdavies-st (approved):

PR #5422, opened 2020-10-21
JP-1357: Code, tests, and docs to update saturation step to test for groups at the A/D floor (0 DN)

### LINKED ISSUE #4690: Saturation step update needed
Issue [JP-1357](https://jira.stsci.edu/browse/JP-1357) was created by Alicia Canipe:

From Karl's baseline cross-check:
 * We recently added a check for "low" saturation where the DN values are 0, which is missing from the documentation and the code
 * This was found in MIRI data, but there was general agreement that we should do this for all data to make sure we catch this if it were to happen in the NIR.

Karl will update ReadTheDocs and the code. Tagging [~morrison]

karllark:
Closes #4690.

Fixes [JP-1357](https://jira.stsci.edu/browse/JP-1357)

One side effect of this update to the saturation check is that values of zero for a group is now flagged as AD_FLOOR and DO_NOT_USE.  This is fine for real data, but for simulated data where it might be common to initialize all the values to zero, this might now cause issues.

### DIFF
--- a/jwst/saturation/saturation.py
+++ b/jwst/saturation/saturation.py
@@ -24,8 +24,9 @@ def do_correction(input_model, ref_model):
     """
     Short Summary
     -------------
-    Apply saturation flagging based on threshold values stored in the
-    saturation reference file.
+    Apply flagging for saturation based on threshold values stored in the
+    saturation reference file and A/D floor based on testing for 0 DN values.
+    For A/D floor flagged groups, the DO_NOT_USE flag is also set.
 
     Parameters
     ----------
@@ -38,7 +39,7 @@ def do_correction(input_model, ref_model):
     Returns
     -------
     output_model : `~jwst.datamodels.RampModel`
-        Data model with saturation flags set in GROUPDQ array
+        Data model with saturation and A/D floor flags set in GROUPDQ array
     """
 
     ramp_array = input_model.data
@@ -87,7 +88,7 @@ def do_correction(input_model, ref_model):
                 # check for high saturation
                 flag_temp = np.where(sci_temp >= sat_thresh, SATURATED, 0)
                 # check for low saturation
-                flaglow_temp = np.where(sci_temp <= 0, AD_FLOOR, 0)
+                flaglow_temp = np.where(sci_temp <= 0, AD_FLOOR | DONOTUSE, 0)
                 # Copy temps into flagarrays.
                 x_irs2.to_irs2(flagarray, flag_temp, irs2_mask, detector)
                 x_irs2.to_irs2(flaglowarray, flaglow_temp, irs2_mask, detector)
@@ -97,7 +98,7 @@ def do_correction(input_model, ref_model):
                                            SATURATED, 0)
                 # check for low saturation
                 flaglowarray[:, :] = np.where(ramp_array[ints, group, :, :] <= 0,
-                                              AD_FLOOR, 0)
+                                              AD_FLOOR | DONOTUSE, 0)
             # for high saturation, the flag is set in the current plane
             # and all following planes.
             np.bitwise_or(groupdq[ints, group:, :, :], flagarray,
--- a/jwst/saturation/saturation.py
+++ b/jwst/saturation/saturation.py
@@ -13,7 +13,9 @@ import numpy as np
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
+DONOTUSE = dqflags.pixel['DO_NOT_USE']
 SATURATED = dqflags.pixel['SATURATED']
+AD_FLOOR = dqflags.pixel['AD_FLOOR']
 NO_SAT_CHECK = dqflags.pixel['NO_SAT_CHECK']
 ATOD_LIMIT = 65535.  # Hard DN limit of 16-bit A-to-D converter
 
@@ -85,7 +87,7 @@ def do_correction(input_model, ref_model):
                 # check for high saturation
                 flag_temp = np.where(sci_temp >= sat_thresh, SATURATED, 0)
                 # check for low saturation
-                flaglow_temp = np.where(sci_temp <= 0, SATURATED, 0)
+                flaglow_temp = np.where(sci_temp <= 0, AD_FLOOR, 0)
                 # Copy temps into flagarrays.
                 x_irs2.to_irs2(flagarray, flag_temp, irs2_mask, detector)
                 x_irs2.to_irs2(flaglowarray, flaglow_temp, irs2_mask, detector)
@@ -95,7 +97,7 @@ def do_correction(input_model, ref_model):
                                            SATURATED, 0)
                 # check for low saturation
                 flaglowarray[:, :] = np.where(ramp_array[ints, group, :, :] <= 0,
-                                              SATURATED, 0)
+                                              AD_FLOOR, 0)
             # for high saturation, the flag is set in the current plane
             # and all following planes.
             np.bitwise_or(groupdq[ints, group:, :, :], flagarray,
@@ -109,6 +111,8 @@ def do_correction(input_model, ref_model):
 
     n_sat = np.any(np.any(np.bitwise_and(groupdq, SATURATED), axis=0), axis=0).sum()
     log.info(f'Detected {n_sat} saturated pixels')
+    n_floor = np.any(np.any(np.bitwise_and(groupdq, AD_FLOOR), axis=0), axis=0).sum()
+    log.info(f'Detected {n_floor} AD FLOOR pixels')
 
     # Save the NO_SAT_CHECK flags in the output PIXELDQ array
     if is_irs2_format:
--- a/jwst/saturation/saturation.py
+++ b/jwst/saturation/saturation.py
@@ -74,26 +74,35 @@ def do_correction(input_model, ref_model):
     sat_dq[np.isnan(sat_thresh)] |= NO_SAT_CHECK
     sat_thresh[np.isnan(sat_thresh)] = ATOD_LIMIT
 
-    # Loop over integrations and groups, checking for pixel values
-    # that are above the saturation threshold
     flagarray = np.zeros(ramp_array.shape[-2:], dtype=groupdq.dtype)
+    flaglowarray = np.zeros(ramp_array.shape[-2:], dtype=groupdq.dtype)
     for ints in range(nints):
         for group in range(ngroups):
-            # Update the 4D groupdq array with the saturation flag. The
-            # flag is set in the current group and all following groups.
+            # Update the 4D groupdq array with the saturation flag.
             if is_irs2_format:
                 sci_temp = x_irs2.from_irs2(ramp_array[ints, group, :, :],
-                                            irs2_mask,
-                                            detector)
+                                            irs2_mask, detector)
+                # check for high saturation
                 flag_temp = np.where(sci_temp >= sat_thresh, SATURATED, 0)
-                # Copy flag_temp into flagarray.
+                # check for low saturation
+                flaglow_temp = np.where(sci_temp <= 0, SATURATED, 0)
+                # Copy temps into flagarrays.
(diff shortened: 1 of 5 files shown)


karllark:
@jemorrison @mwregan2

jwst/saturation/saturation.py
```diff
@@ -74,26 +74,35 @@ def do_correction(input_model, ref_model):
...
+                # Copy temps into flagarrays.
                 x_irs2.to_irs2(flagarray, flag_temp, irs2_mask, detector)
-
+                x_irs2.to_irs2(flaglowarray, flaglow_temp, irs2_mask, detector)
             else:
+                # check for high saturation
                 flagarray[:, :] = np.where(ramp_array[ints, group, :, :] >= sat_thresh,
                                            SATURATED, 0)
+                # check for low saturation
+                flaglowarray[:, :] = np.where(ramp_array[ints, group, :, :] <= 0,
```
hbushouse, line 97: I was about to suggest combining the two thresholds into a single `np.where()` and hence just use one flagarray, but then realized that the high saturation flags need to be kept separate so that they can be used to flag subsequent groups.

jwst/saturation/tests/test_saturation.py
```diff
@@ -42,6 +42,70 @@ def test_basic_saturation_flagging(setup_nrc_cube):
...
 
 
+def test_low_saturation_flagging(setup_nrc_cube):
+    """Check that the saturation flag is set when a pixel value is zero or
+    negative."""
+
+    # Create inputs, data, and saturation maps
+    ngroups = 5
+    nrows = 2048
+    ncols = 2048
```
hbushouse, line 52: Don't need 2k rows and columns when only pixel 5,5 is being checked in the test. Could make it much smaller like 10 or 20 to save time when running the test.

jwst/saturation/tests/test_saturation.py
```diff
@@ -42,6 +42,70 @@ def test_basic_saturation_flagging(setup_nrc_cube):
...
+    data.data[0, 1, 5, 5] = 0  # Signal at bottom rail - low saturation
+    data.data[0, 2, 5, 5] = 20
+    data.data[0, 3, 5, 5] = 40
+    data.data[0, 4, 5, 5] = 60
+
+    # frames that should be flagged as saturation (low)
+    satindxs = [0, 1]
+
+    # Set saturation value in the saturation model
+    satmap.data[500, 500] = satvalue
```
hbushouse, line 68: If you reduce nrows/ncols above, then of course change this to something like `satmap.data[8, 8]`

jwst/saturation/tests/test_saturation.py
```diff
@@ -42,6 +42,70 @@ def test_basic_saturation_flagging(setup_nrc_cube):
...
+
+
+def test_low_and_high_saturation_flagging(setup_nrc_cube):
+    """Check that the saturation flag is set when a pixel value is zero or
+    negative and when the pixel is above the high saturation threshold."""
+
+    # Create inputs, data, and saturation maps
+    ngroups = 5
+    nrows = 2048
+    ncols = 2048
```
hbushouse, line 84: Same comment regarding image size.

jwst/saturation/saturation.py
karllark: I had this same thought.  I may even have written it this way to start.  But then had the same realization.  Great minds?

jwst/saturation/tests/test_saturation.py
karllark: Done.

jwst/saturation/tests/test_saturation.py
karllark: Done.

jwst/saturation/tests/test_saturation.py
karllark: Done.

mwregan2:
I think we need to make concurrent changes to ramp_fitting in this pull request. Given that your using the same flag to mean two different things, the logic in ramp fitting will be more complicated. It already has to deal with a lot of edge cases and now almost every one of those cases will also have a permutation where the ramp starts and zero. Alternatively, there might be a way to preprocess the ramp (within ramp fitting) to remove the floored values from the fit and then run ramp fitting like normal.

karllark:
> I think we need to make concurrent changes to ramp_fitting in this pull request. Given that your using the same flag to mean two different things, the logic in ramp fitting will be more complicated. It already has to deal with a lot of edge cases and now almost every one of those cases will also have a permutation where the ramp starts and zero. Alternatively, there might be a way to preprocess the ramp (within ramp fitting) to remove the floored values from the fit and then run ramp fitting like normal.

Can you be more specific?  If saturated groups are not used in the fit already, then it is not clear to me how flagging the 0 DN (and below) groups requires changes to the ramp fitting.  I have been assuming that the ramp fitting does not use any saturated groups in the fitting and fits whatever remaining segments with good groups.

mwregan2:
The first rule is "assume nothing". Overloading the saturation flag with two meanings will mess up ramp fitting. Here's what the code says: 
"If the initial group of a ramp is saturated, it is assumed that all groups are saturated."
This is a completely natural assumption. 
I think there should be a different flag for A/D floor. If not, then we need to add a bunch of test cases to ramp fitting with A/D floor values at the beginning of ramps.

karllark:
Argh.  I'm fine with a different flag for low sat/AD floor.  @hbushouse : is this possible?  @mwregan2 : will a different flag solve the issue with ramp fitting?

karllark:
> "If the initial group of a ramp is saturated, it is assumed that all groups are saturated."

While this might be a reasonable assumption, why is it made?  Isn't the goal of flagging groups prior to ramp fitting to allow ramp fitting not to have specific knowledge of the details of the ramp flagging, just to use the groups that are not flagged (or only have advisory flags)?

dmggh:
This assumption was made years ago based on our understanding at that time. Do we want to reconsider it ?

karllark:
Only if we can't use a different flag for low saturation that will remove low saturated groups from the ramp fitting.  ;-)

jemorrison:
Reading the comments can you handle this like the baseline RSCD correction flags data and set the dqflags.groupdq = DO_NOT_USE.

karllark:
> Reading the comments can you handle this like the baseline RSCD correction flags data and set the dqflags.groupdq = DO_NOT_USE.

Great idea!  I'll make this change.

hbushouse:
While using DO_NOT_USE instead of SATURATION will fix the problem from a mechanical standpoint, it does go against the overall DQ flag philosophy of always having an informational flag to go with DO_NOT_USE in order to document *why* it's not to be used. I wouldn't mind creating a new flag that's specific for low saturation, but given that we're already at the limit of the uint32 data type for the DQ array, that would mean either bumping the data type up to uint64 (which I'd *really* like to avoid) or someone does some house cleaning on the existing flags to free up a spot. It seems to me there are several in there that were defined years ago before we even had real data for some of the instruments and have never actually been used.

karllark:
Ok.  Well, it sounds like I should wait to update this PR until I hear what dqflag I should use for low saturation.

jemorrison:
I think we already broke that idea. For MIRI the reason the data is 0 is mostly because of the RSCD effect. In the RSCD step we set the groups we don't want to use to DO_NOT_USE. So if we are going to create a new flag for low saturation you really should create a new flag for when we drop frames in the RSCD step. I actually would love the clean up the DQ flags. A few years ago Nadia asked me to go through the flags (there is a confluence page someplace) and I found some  different flags are being used to mean the same thing but for different instruments. I will look for the confluence page

hbushouse:
Maybe we just need a new DQ flag value that says "MIRI", thus indicating all data are suspect?  ;-)

jemorrison:
I searched confluence and I could not find the "report" I put together on how the DQ flags are used by pipeline, where there are duplicate flags meaning different things used in reference files that could be combined, and I think there was even a flag that was not being use. For example I know MIRI is not using NO_SAT_CHECK anymore in the saturation reference files (or will after I deliver a new file) but that is still in list. I think it was put in for MIRI.  Also do we need a flag for BAD_REF_PIXEL when we could just combine REFERENCE_PIXEL and DO_NOT_USE ?  I did that study about 2 years ago any way. We would have to review things again.  If it is time we reviewed the DQFlags and clean them up - I can open a JIRA ticket and I don't mind doing the research unless Howard has someone else in mind for the review.

jemorrison:
@karllark @hbushouse  I was looking back over tickets that I needed to do something. Should we file a ticket to clean up the DQ flags so we can add an option for flagging low saturated data and data flagged as being rejected for RSCD effect ?

karllark:
Definitely would be good to have a DQ flag for low saturation.

I was wondering if I should move forward with the PR w/o the new DQ flag.  Just flag the frames as DO_NOT_USE (like RSCD) and then do a separate PR once we have the appropriate DQ flag (maybe even the same PR for adding a DQ flag to RSCD).  What do others think?

jemorrison:
I am going to put together a JP ticket for the DMS WG to look at. We plan on trying to clean up the DQ flags so we can have a specific flag for low saturation. I will flag you when I get this new ticket done. We really should also add a flag for RSCD rejects.

hbushouse:
> I will **flag** you when I get this new ticket done.

I see what you did there.

mwregan2:
Karl
  Please stop calling this “low saturation”. It has nothing to do with saturation.
Mike

jemorrison:
@hbushouse @karllark  Since I think it will take a few weeks for JP-1800 (cleaning up DQ flags) to  come to any conclusion I think it would be good to get this part of Build 7.7. So @hbushouse  do you think we can just flag these types of low DN as DO_NOT_USE for now. We can always come back later and update the DQ flag with a new flag (once we get one)

jemorrison:
@karllark @mwregan2  Just had a MIRI tag up. Howard says that since this flagging is done on a group level we can take 
one of the reserved ones. So we can take RESERVED_3 and call it something that makes sense. Low_DN  or Zero_DN - any better ideas ?

karllark:
> Low_DN or Zero_DN - any better ideas ?

How about "AD_FLOOR" to make @mwregan2 happy?

mwregan2:
I like Karl’s version.

jemorrison:
Sounds perfect. Karl you ok with updating the dqflags and updating this step with that dqflag. I believe you have to combine with DO_NOT_USE for the jump step to reject it from the fit.

karllark:
> Sounds perfect. Karl you ok with updating the dqflags and updating this step with that dqflag. I believe you have to combine with DO_NOT_USE for the jump step to reject it from the fit.

Yes.  Will work on doing this in the next few days.

hbushouse:
@karllark don't like being a nag, but we'd like to get this merged in the next day or two, so that we have time to include it in some regression test runs before we make our first B7.7 delivery early next week (12/21 or 12/22).

karllark:
I am working on this.  Will push to get it done today.  Reminders/pings are always a good idea!

karllark:
If/when the tests pass, I would consider this PR ready from my side.

jemorrison (approved):
Looks good to me.  Nice to also see those additional tests.

hbushouse (approved):

hbushouse:
Is this change supposed to apply to only MIRI data or all instruments? I'm seeing changes in some regression tests for both MIRI and NIRSpec.

karllark:
All instruments.  In theory, none of the NIR detectors should have this issue, but when this was discussed in the CalWG having the check for the NIR and MIR was seen as a good thing.  Just in case.  And made it easier to implement.

If you are seeing changes for NIRSpec, is this for simulated or real data?  If simulated, were some pixel values set to zero be default even if this is not expected for real data?

hbushouse:
The NIRSpec data set is garbage test pattern data from OTIS, so I guess it's no surprise that it's got some zeros in it. Thanks for the confirmation.

PR #5428, opened 2020-10-26
Updates to extract_1d logging

hbushouse:
Minor updates to logging and code documentation in the extract_1d step, mostly pertaining to cases where looping over integrations is being done for TSO data. No change to functionality.


stscieisenhamer (approved):
Approved.

However, for future comment: The general use of logging by this module is not good, by either Python conventions or STPIPE usage. That is for another day.

hbushouse:
> Approved.
> 
> However, for future comment: The general use of logging by this module is not good, by either Python conventions or STPIPE usage. That is for another day.

Just out of morbid curiosity, in what way(s) is it not good?

PR #5431, opened 2020-10-26
Documentation update on ways to call the pipeline

### LINKED ISSUE #5425: Confusing directives for .run() and .call()
_Issue [JP-1756](https://jira.stsci.edu/browse/JP-1756) was created on JIRA by [Alicia Canipe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=acanipe):_

We've had issues in the past with confusion about using different syntax for the pipeline. Many users run the pipeline within ipython, especially since examples are provided to them in Jupyter notebooks. 

The documentation here: <https://jwst-pipeline.readthedocs.io/en/latest/jwst/introduction.html#running-from-within-python>
provides them with instructions for the .call() method only, though most people are using the .run() method to be able to change the parameters on the fly. 

However, the .run() method doesn't accept config files or arguments overriding the config file: see #1098 

and the .call() method doesn't accept overrides: see <https://stsci.slack.com/archives/CBQQMLA7Q/p1591735522253800> 

So if a user wants to override the config file, they have to use the .call() method, but if they want to override any parameters, they have to use the .run() method. This kind of information isn't clear in the documentation and has tripped INS up a few times.

bhilbert4:
This PR makes some small tweaks/additions to the documentation on the ways to call pipelines and steps from within python. The changes highlight the differences between the `call` and `run` methods and describe how to provide input parameters for each. After numerous questions/errors from users in the building on the details for how to call the pipeline, I was motivated to try and clarify the documentation. 

I realize that there is upcoming work on the methods used to call the pipeline and the details/best practices may change. I figured that these changes will still help people using the pipeline between now and then, and perhaps these changes can be used as a springboard for future documentation updates after any changes to the methods.

Resolves #5425

### DIFF
--- a/docs/jwst/introduction.rst
+++ b/docs/jwst/introduction.rst
@@ -162,7 +162,7 @@ step instance directly.
 
 The ``call`` method creates a new instance of the class and runs the pipeline or
 step. Optional parameter settings can be specified by supplying a configuration file,
-or via keyword. Examples are shown on the :ref:`Execute via call()<call_examples>` page.
+or via keyword arguments. Examples are shown on the :ref:`Execute via call()<call_examples>` page.
 ::
 
  from jwst.pipeline import Detector1Pipeline
--- a/docs/jwst/introduction.rst
+++ b/docs/jwst/introduction.rst
@@ -162,7 +162,7 @@ step instance directly.
 
 The ``call`` method creates a new instance of the class and runs the pipeline or
 step. Optional parameter settings can be specified by supplying a configuration file,
-or via keyword. Examples are shown on the :ref:`Execute via call()<call-examples>` page.
+or via keyword. Examples are shown on the :ref:`Execute via call()<call_examples>` page.
 ::
 
  from jwst.pipeline import Detector1Pipeline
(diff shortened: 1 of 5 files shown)


docs/jwst/introduction.rst
```diff
@@ -157,7 +157,12 @@ Running From Within Python
 ==========================
 
 You can execute a pipeline or a step from within python by using the
-``call`` method of the class:
+``call`` or ``run`` methods of the class, or by calling the pipeline or
+step instance directly.
+
+The ``call`` method creates a new instance of the class and runs the pipeline or
+step. Optional parameter settings can be specified by supplying a configuration file,
+or via keyword. Examples are shown on the :ref:`Execute via call()<call_examples>` page.
```
hbushouse, line 165: Can you think of a way to make it more explicit here that you're referring to keyword arguments passed to the method, as opposed to the generic way that folks immediately think of the term "keyword" (i.e. a FITS header keyword). Perhaps just "or via keyword arguments".

docs/jwst/stpipe/user_step.rst
```diff
@@ -224,7 +225,12 @@ very useful if one wants to setup the step's attributes first, then run it::
     mystep.override_sflat = ‘sflat.fits’
     output = mystep.run(input)
 
-Using the `.run()` method is the same as calling the instance or class directly.
+`input` in this case can be a fits file containing the appropriate data, or the output
+of a previously run step/pipeline, which is an instance of a particular :ref:`datamodel<datamodels>`.
+
+Unlike in the use of ``call``, a configuration file supplied whle instantiating ``run()`` will be ignored.
```
hbushouse, line 231: typo "whle" -> "while"

bhilbert4: Fixed

docs/jwst/introduction.rst
bhilbert4: I changed it to "keyword arguments", which I think makes it clear.

hbushouse (approved):

Issue #3132, opened 2019-02-15
Feature Request: Separate out data models into its own package


pllim:
I thought I opened an issue about it but I cannot find it.

I need the data models to read `jwst` ASDF files but I don't want to ever run `jwst` pipeline. When I install `jwst` to just get to the data models, I am not thrilled on how many other dependencies it is installing, which I am pretty sure I don't need to just for *reading* the data. Would be nice if data models is its own package. It probably only would pull `gwcs` and `pyyaml` with it, instead of these plethora of packages?

See also:

* spacetelescope/drizzle#33 -- FIXED thanks to @jhunkeler 
* spacetelescope/stsci.image#9

pllim:
Another motivation is that I don't want to install `jwst` natively on Windows because it is not guaranteed to work on Windows.

pllim:
And in case you are wondering, I tried to install `jwst` on Windows 10 and failed.
```
    utest_cdrizzle.c
    c:\...\drizzle\drizzle\src\tests\fct.h(213): fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory
    error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\Community\\VC\\Tools\\MSVC\\14.14.26428\\bin\\HostX86\\x64\\cl.exe' failed with exit status 2

    ----------------------------------------
Command "C:\...\Miniconda3\envs\py37\python.exe -u -c "import setuptools, tokenize;__file__='C:\\...\\AppData\\Local\\Temp\\pip-install-5xl1o1iu\\drizzle\\setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record C:\...\AppData\Local\Temp\pip-record-rkb03kjc\install-record.txt --single-version-externally-managed --compile" failed with error code 1 in C:\...\AppData\Local\Temp\pip-install-5xl1o1iu\drizzle\
```
As a result, I am unable to read JWST Level 2 ASDF-in-FITS portion of the data on Windows. Even when @drdavella told me how to load the ASDF part without `jwst`, it fails with errors about `TaggedDict` without `jwst` installed.

jhunkeler:
That might be an easy fix in `drizzle`...

```c
/* Platform indepedent pipe functions. TODO: Look to figure this out in a way
that follows the ISO C++ conformant naming convention. */
#if defined(WIN32)
[...]

PR #5433, opened 2020-10-26
JP-1779: Move core jwst.datatmodels functionality to stdatamodels

### LINKED ISSUE #5469: jwst should use stdatamodels
_Issue [JP-1779](https://jira.stsci.edu/browse/JP-1779) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

Move jwst to using stdatamodels. Do this on a separate branch in the `jwst` repo.

 - Use the jwst core schema with stdatamodels. This will require a subclass of stdatamodels.DataModel called JwstDataModel which points to the JWST core schema.

- jwst.datamodels.open will remain implemented.

- Update tests and other subclasses to use JwstDataModel where appropriate, i.e. where the core schema is needed.

jdavies-st:
This PR removes the core `jwst.datamodels.model_base` code and replaces it with a dependency on the `stdatamodels` package, which is where all that code went.

https://github.com/spacetelescope/stdatamodels/tree/master/src/stdatamodels

Additionally, since `stdatamodels.DataModel` doesn't have a core schema that corresponds with the JWST core schema, a new base class `JwstDataModel` is introduced here as the new base class with the JWST core schema, though we've kept `jwst.datamodels.DataModel` as well with the option to deprecate it.  Thoughts?

Of course this is motivated by the desire for `jwst` and `romancal` to share the core `stdatamodels` code.

Updates to `msaflagopen` tests are completely unrelated, but they were the canary that helped me find a bug in the `crds_client`, so I kept those changes for clarity.

Resolves #5469 / [JP-1779](https://jira.stsci.edu/browse/JP-1779)
Addresses part of #3132


eslavich:
> Any suggestions on a solution @eslavich ?

How about we compare the DATAMODL string to the class's string name?  That seems equivalent to me, since DATAMODL starts off as a string anyway.

jdavies-st:
That seems reasonable.  Then the comparison can be done without having to load up the package (which may not exist in the installation).

jdavies-st:
So the failures seen in this PR relating to table validation

https://travis-ci.org/github/spacetelescope/jwst/jobs/739800949#L1086-L1090

are a byproduct of fixing bugs in the validator on the `stdatamodels` side.  See:

https://github.com/spacetelescope/stdatamodels/pull/4/files#diff-b7092a6c38c74202a51899938a5c9451867217ef795268cc6ae555474a377c0fL329-R331

https://github.com/spacetelescope/stdatamodels/pull/4/files#diff-28744f508b621b779032bc4f0f0542d33778e66674941ec524d213cea968c24eL65-R50

for instance.

@eslavich and I discussed this, and since the solutions to get non-buggy performance with correct validation are to rewrite a number of schemas, it would be better to back out this validator bugfix in `stdatamodels`, get `jwst.datamodels` using `stdatamodels` as in this PR, then fix things together in future PRs piecemeal.  That will speed along the merging of this PR with mininmal behavior changes and without wholesale changing of schemas in `jwst.datadmodels`.

jdavies-st:
The fixes have been made in `stdatamodels` 

https://github.com/spacetelescope/stdatamodels/pull/19
https://github.com/spacetelescope/stdatamodels/pull/21
https://github.com/spacetelescope/stdatamodels/pull/22

to enable this PR.

One further fix to get the dev job to pass is required.

https://github.com/spacetelescope/stdatamodels/issues/3

But since that's an allowed failure right now, it's not blocking this PR.

jwst/master_background/tests/test_master_background.py
```diff
@@ -111,3 +114,34 @@ def test_copy_background_to_surf_bright():
...
+                    },
+                ]
+            }
+        ]
+    }
+
+    sci, bkg = split_container(container)
+
+    assert sci[0].meta.filename == "foo.fits"
+    assert bkg[0].meta.filename == "bar.fits"
```
eslavich, line 147: The code style check is complaining about this missing newline

.travis.yml
```diff
@@ -37,6 +37,9 @@ jobs:
       install:
         - pip install -U pip certifi
         - pip install $PIP_DEPENDENCIES
+        # Since there is no version of stdatamodels on PyPI, install from
+        # Github master until it is released to PyPI
+        - pip install git+https://github.com/spacetelescope/stdatamodels
```
eslavich, line 42: Is this necessary?  I would have thought your setup.cfg requirement would be enough to get it installed from source.

jwst/datamodels/tests/test_history.py
```diff
@@ -1,96 +0,0 @@
-import datetime
```
eslavich, line 1: Should we keep this file for now, since these tests aren't running yet in stdatamodels?

requirements-dev.txt
```diff
@@ -1,6 +1,7 @@
-git+https://github.com/spacetelescope/asdf#egg=asdf
-git+https://github.com/astropy/astropy#egg=astropy
-git+https://github.com/spacetelescope/crds#egg=crds
-git+https://github.com/spacetelescope/gwcs#egg=gwcs
-git+https://github.com/astropy/photutils#egg=photutils
-git+https://github.com/spacetelescope/tweakwcs#egg=tweakwcs
+git+https://github.com/spacetelescope/asdf
```
eslavich, line 1: Can't make an omelette without breaking a few packages

.travis.yml
jdavies-st: Unfortunately it is, as for this job the pip install script is

```
pip install -r requirements-min.txt .[test]
```

where `requirements-min.txt` is generated by the `minimum_deps` script, which takes the `install_requires` and finds the minimum version available on pypi and uses that.  For `stdatamodels`, the minimum and only version is 0.0.0, which doesn't work.  So we need to install the latest version for that package.  Once we deliver a proper version of `stdatamodels` to pypi, we can remove this.

jwst/datamodels/tests/test_history.py
jdavies-st: We could. It's not testing anything in `jwst.datamodels` though.  We'll just have to make a note to port the tests over to `stdatamodels`.

Is there a reason these tests are disabled in `stdatadmodels`?  Is it just a lack of a `history` schema entry location?

requirements-dev.txt
jdavies-st: Ha!  The `egg=` portion here is not necessary, as the package name and egg name are the same.

jwst/datamodels/tests/test_history.py
eslavich: Asking the good questions!  I tried enabling them and they all passed.  I think they were failing before due to the model class test in the FITS hash code, which we fixed.  I'll open a PR over there to get those running.

jdavies-st:
Btw, the project-level `codecov` coverage here is less because we are deleting a bunch of well-covered code, leaving behind less-well-covered code.

kmacdonald-stsci (commented):
This all looked fine.  The changes looked like they were largely label changes to deal with `stdatamodels` being its own separate package, instead of coupled with `jwst`.  Besides one of the style comments I had, is there a need for `JwstDataModels` rather than simply using `DataModels` from the `stdatamodels`?

jwst/datamodels/tests/test_schema.py
```diff
@@ -376,12 +372,10 @@ def test_table_array_convert():
     Test that structured arrays are converted when necessary, and
     reused as views when not.
     """
-    from jwst.datamodels import util
 
     table_schema = {
         "allOf": [
-            mschema.load_schema(os.path.join(URL_PREFIX, "image.schema"),
-                resolver=asdf.AsdfFile().resolver,
+            mschema.load_schema("http://stsci.edu/schemas/jwst_datamodel/core.schema",
```
nden, line 378: Should  `image.schema` be tested here?

eslavich (approved):
🥳

jwst/msaflagopen/tests/test_msa_open.py
```diff
@@ -72,9 +80,10 @@ def test_get_failed_open_shutters():
     # get_failed_open_shutters returns 3 flaggable states
     # state, Internal state, and TA state.
     for shutter in result:
-        assert (shutter['state'] == 'open' or
-                shutter['Internal state'] == 'open' or
-                shutter['TA state'] == 'open')
+        assert shutter['state'] == 'open' or \
```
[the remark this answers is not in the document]
jdavies-st: Good point.  I'll change it to having `or` at the beginning.  I was changing this as part of a larger cleanup of problems with assert using parentheses:

```python
>>> assert 1 == 2
AssertionError
>>> assert 1 == 2, "hi"
AssertionError: hi
>>> assert (1 == 2)
AssertionError
>>> assert(1 == 2)
AssertionError
>>> assert(1 == 2, "hi")
```

That last one evaluates to `True` and is a bug.

Generally, it's a bad idea to use the `assert` keyword with parenthesis, as it's not a function.  I found a bug in similar test elsewhere in the repo recently here, so I was just being proactive.  Maybe a bit too proactive.  =)

jwst/datamodels/tests/test_schema.py
jdavies-st: A base `DataModel` was being tested before

https://github.com/spacetelescope/jwst/pull/5433/files#diff-c83baf50e80f78e21b7a376fe4208a97ea9421e27f8ca0b3a12121820c8ce588L417

And I've just changed it to a `JwstDataModel` which only depends on the core schema.

So I think this is a bit clearer and consistent, especially since nothing specific to `ImageModel` is used in these tests.

jdavies-st:
I've rebased to resolve conflicts with #5467

jdavies-st:
And a note to @hbushouse , this PR introduces a dev dependency on the unreleased version of `stdatamodels` for the time being.  We can probably resolve this at any time by just doing a release of `stdatamodels` to pypi, but I just wanted to make this clear.  Hopefully we can get a release of `stdatamodels` out very soon, maybe the next few weeks?

The reason I bring this up is that having this dev dependency will prevent us from delivering `jwst` to pypi if we needed to do a release.  But as long as we release `stdatamodels` well before our next schedule release of `jwst`, we should be good.

PR #5437, opened 2020-10-27
JP-1764:Update ami_analyze to extract subarray from full-frame images

### LINKED ISSUE #5436: Extract SUB80 subarray from full-frame AMI exposures
_Issue [JP-1764](https://jira.stsci.edu/browse/JP-1764) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The new version of AMI LG processing implemented in #5390 no longer includes the on-the-fly extraction of a subarray from the input data when the exposure is full-frame, as was done in the old version of the step. This process needs to be added back into the new code, because processing times for any data larger than a few hundred x a few hundred pixels are unacceptably slow and there's no useful data further away than about 40-50 pixels from the target. The extraction should be the equivalent of having used the SUB80 subarray to take the exposure. The AMI group agrees that this should can can be added back into the new code.

hbushouse:
Update the `ami_analyze` module to extract the equivalent of a SUB80 subarray from input images that were taken in full-frame mode.

Fixes #5436 / [JP-1764](https://jira.stsci.edu/browse/JP-1764)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@
 ami
 ---
 
-- Update code and unit tests to use new ami analyze algorithm [#5390]
+- Update code and unit tests to use new ami_analyze algorithms [#5390]
+
+- Update ami_analyze to extract a SUB80 subarray from full-frame images [#5437]
 
 assign_wcs
 ----------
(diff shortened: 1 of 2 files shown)


dmggh (approved):
Apparently that was lost in the update. Thanks for catching and reinstating this !

hbushouse:
It was one of the DMS test suites that found it - insanely long processing time for a bunch of full-frame exposures in a single ASN.

PR #5440, opened 2020-10-28
Fix _apply_sky in skymatch for single group input

mcara:
Fixes a bug in `skymatch.skymatch._apply_sky()` that would result in a crash when `skymethod` contains 'global' and input image is a single image group whose sky cannot be computed, e.g., because its images contains no "good" pixels.

This bug was revealed after PR #5423 which permitted processing of single image/group inputs.

Fixes failing test `test_nircam_image.py::test_image3_closedfile`. The test will still be failing for two reasons:
1. Extra keyword 'S_SKYMAT' - that's expected
2. Small differences in many image pixels. This failure is not related to this PR and probably is caused by unrelated changes. This test was not OKified before because it was crashing.

### DIFF
--- a/jwst/skymatch/skymatch.py
+++ b/jwst/skymatch/skymatch.py
@@ -359,9 +359,7 @@ def match(images, skymethod='global+match', match_down=True, subtract=False):
             log.info(" ")
             if minsky is None:
                 log.warning("   Unable to compute \"global\" sky value")
-                sky_deltas = len(sky_deltas) * [None]
-            else:
-                sky_deltas = len(sky_deltas) * [minsky]
+            sky_deltas = len(sky_deltas) * [minsky]
             log.info("   \"Global\" sky value correction: {} "
                      "[not converted]".format(minsky))
 
@@ -382,24 +380,26 @@ def match(images, skymethod='global+match', match_down=True, subtract=False):
 
 
 def _apply_sky(images, sky_deltas, do_global, do_skysub, show_old):
+    img_type = ['Image', 'Group']
+
     for img, sky in zip(images, sky_deltas):
-        img_type = 'Image' if isinstance(img, SkyImage) else 'Group'
+        is_group = not isinstance(img, SkyImage)
 
         if do_global:
             if sky is None:
-                valid = img.is_sky_valid
+                valid = img[0].is_sky_valid if is_group else img.is_sky_valid
                 sky = 0.0
             else:
                 valid = True
 
         else:
-            valid = sky is None
-            if valid:
+            valid = sky is not None
+            if not valid:
                 log.warning("   *  {:s} ID={}: Unable to compute sky value"
-                            .format(img_type, img.id))
+                            .format(img_type[is_group], img.id))
                 sky = 0.0
 
-        if img_type == 'Group':
+        if is_group:
             # apply sky change:
             old_img_sky = [im.sky for im in img]
             if do_skysub:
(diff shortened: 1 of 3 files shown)


jwst/skymatch/skymatch.py
```diff
@@ -382,24 +380,26 @@ def match(images, skymethod='global+match', match_down=True, subtract=False):
...
                 valid = True
 
         else:
-            valid = sky is None
-            if valid:
+            valid = sky is not None
+            if not valid:
                 log.warning("   *  {:s} ID={}: Unable to compute sky value"
-                            .format(img_type, img.id))
+                            .format(img_type[is_group], img.id))
```
hbushouse, line 399: Is it kosher to use a boolean (`is_group`) as an index into a list? I guess it sort of works by having False and True equate to 0 and 1, but ...

mcara: Is there anything kosher in Python? Unfortunately not a Python priest here. booleans are a subclass of `int` and so I do not think there is anything wrong with this.

hbushouse (approved):

jwst/skymatch/skymatch.py
mcara: Possible alternatives:
```
img_type[int(is_group)]
```
or
```
img_type[is_group + 0]
```
I like this last one above :)

Other alternatives (both overkill IMO): make `img_type` a dictionary or have an `if` statement.

mcara: I found something: https://www.python.org/dev/peps/pep-0285/ - see point 6:

> Should bool inherit from int?
>
> => Yes.
>
> In an ideal world, bool might be better implemented as a separate integer type that knows how to perform mixed-mode arithmetic. However, inheriting bool from int eases the implementation enormously (in part since all C code that calls PyInt_Check() will continue to work -- this returns true for subclasses of int). Also, I believe **this is right in terms of substitutability: code that requires an int can be fed a bool and it will behave the same as 0 or 1.**

hbushouse: You like the last one above, because it's clever (adding an int value will cause the result to be int), while I like the first one (explicitly casting the result to int), because it's explicit (and hence doesn't require a reader to think about what's happening).

mcara: Ha! priest has spoken (see same PEP "Compatibility" section):

> Because of backwards compatibility, the bool type lacks many properties that some would like to see. For example, arithmetic operations with one or two bool arguments is allowed, treating False as 0 and True as 1. **Also, a bool may be used as a sequence index.**
>
> I don't see this as a problem, and I don't want evolve the language in this direction either. I don't believe that a stricter interpretation of "Booleanness" makes the language any clearer.

jwst/skymatch/skymatch.py
```diff
@@ -382,24 +380,26 @@ def match(images, skymethod='global+match', match_down=True, subtract=False):
 
 
 def _apply_sky(images, sky_deltas, do_global, do_skysub, show_old):
+    img_type = ['Image', 'Group']
```
jdavies-st, line 383: I would add a comment here,

```
# The following to be indexed by boolean is_group
img_type = ['Image', 'Group']
```
to make this clear.

jdavies-st (approved):
Looks good.  Just a request for a comment.

jwst/skymatch/skymatch.py
mcara: Done

mcara: Done again

hbushouse (approved):

PR #5441, opened 2020-10-28
JP-1751 Retrieve step parameter references from CRDS instrument "calpars"

### LINKED ISSUE #5416: Update stpipe to retreive step parameter references from CALPARS
_Issue [JP-1751](https://jira.stsci.edu/browse/JP-1751) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Update the `stpipe` architecture to retrieve the step parameter references from the new pseudo-instrument `calpars`.

## Implementation

* Update `get_config_from_reference` to replace `meta.instrument.name`  with `calpars`
* Save original instrument in a top level keyword, such as `science_instrument`

stscieisenhamer:
Resolves [JP-1751](https://jira.stsci.edu/browse/JP-1751)
Resolves #5416 

[CCD-739](https://jira.stsci.edu/browse/CCD-739) implements a new "instrument" called `calpars`. Within this instrument, all the step parameter reference files are located. To retrieve these references, the `meta.instrument.name` must be overridden in the `DataModel` being used as the reference file selector.

In addition, a new utility function, `jwst.stpipe.utilities.all_steps`, is implemented. `all_steps` returns all classes within the package that subclass `Step`.

### DIFF
--- a/jwst/stpipe/step.py
+++ b/jwst/stpipe/step.py
@@ -729,6 +729,14 @@ class Step():
         logger = log.delegator.log
         pars_model = cls.get_pars_model()
 
+        # If the dataset is not an operable DataModel, log as such and return
+        # an empty config object
+        try:
+            model = dm_open(dataset)
+        except (IOError, TypeError, ValueError):
+            logger.warning(f'Input dataset is not a DataModel.')
+            disable = True
+
         # Check if retrieval should be attempted.
         if disable is None:
             disable = get_disable_crds_steppars()
@@ -737,14 +745,18 @@ class Step():
             return config_parser.ConfigObj()
 
         # Reassign and remember the instrument.
-        extra_pars = {'calpars_instrument': dataset.meta.instrument.name,
+        if isinstance(model, ModelContainer):
+            instrument = model[0].meta.instrument.name
+        else:
+            instrument = model.meta.instrument.name
+        extra_pars = {'calpars_instrument': instrument,
                       'meta.instrument.name': 'calpars'}
 
         # Retrieve step parameters from CRDS
         logger.debug(f'Retrieving step {pars_model.meta.reftype.upper()} parameters from CRDS')
         exceptions = crds_client.get_exceptions_module()
         try:
-            ref_file = crds_client.get_reference_file(dataset,
+            ref_file = crds_client.get_reference_file(model,
                                                       pars_model.meta.reftype,
                                                       observatory=observatory,
                                                       asn_exptypes=['science'],
--- a/jwst/stpipe/step.py
+++ b/jwst/stpipe/step.py
@@ -714,6 +714,14 @@ class Step():
         step_parameters : configobj
             The parameters as retrieved from CRDS. If there is an issue, log as such
             and return an empty config obj.
+
+        Notes
+        -----
+        The CRDS lookup normally uses `META.INSTRUMENT.NAME` as the primary key, to distinguish
+        references at the instrument `imap` level. However, the step parameter reference files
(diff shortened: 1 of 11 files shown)


jwst/stpipe/step.py
```diff
@@ -714,28 +714,53 @@ def get_config_from_reference(cls, dataset, observatory=None, disable=None):
         step_parameters : configobj
             The parameters as retrieved from CRDS. If there is an issue, log as such
             and return an empty config obj.
+
+        Notes
+        -----
+        The CRDS lookup normally uses `META.INSTRUMENT.NAME` as the primary key, to distinguish
+        references at the instrument `imap` level. However, the step parameter reference files
+        are all grouped in a pseudo-instrument "CALPARS". As such, the instrument name is set to
+        "CALPARS". If an rmap requires knowledge of the original instrument, this is located
```
hbushouse, line 723: I assume (and from what I've seen in the new code above) that it's not the case that the instrument name is simply being reset to "calpars", because how then would a call to CRDS find any instrument-level reference files? But in reality it's being given multiple instrument names, one of which is "calpars" and the other is the original instrument (e.g. "miri", "nircam", ...)? Is that correct? If so, maybe the Notes and/or code comments could make that a little more clear.

jwst/stpipe/tests/test_utilities.py
```diff
@@ -0,0 +1,34 @@
+"""Test utility funcs"""
+from jwst.stpipe.utilities import all_steps
+
+
+# Snapshot of known steps
+KNOWN_STEPS = set([
+    'AlignRefsStep','Ami3Pipeline', 'AmiAnalyzeStep', 'AmiAverageStep',
```
hbushouse, line 7: Wow. Who/what/where attempts to keep this list up to date as we continue development and create/rename/remove step classes?

hbushouse (approved):
I can't spot anything obviously wrong, but that doesn't mean much. I will rely on the kindness of CRDS infrastructure experts to really shake this down.

jwst/stpipe/tests/test_utilities.py
stscieisenhamer: This could be parred down to a short-list that will probably not change and would take less maintenance.

However, the intent of `all_steps` is to return all steps, and this should be expected to fail in create/delete cases.  In particular, as a canary-test, unintended changes to step definitions will be caught. As to maintenance, this would be no different than any other code change that requires unit test justification. If the value vs. work is considered insufficient, then, as stated, the test could be limited.

jwst/stpipe/utilities.py
```diff
@@ -117,3 +136,99 @@ def islist_tuple(obj):
     Return True if `obj` is either a list or a tuple. False otherwise.
     """
     return isinstance(obj, tuple) or isinstance(obj, list)
+
+
+def all_steps():
```
eslavich, line 141: Is it safe to assume that the steps will be used in at least one pipeline?  If so, this might be a reasonable substitute:

```python
def all_steps():
    from jwst import pipeline

    steps = {}

    for pipeline_name in pipeline.__all__:
        pipeline_class = getattr(pipeline, pipeline_name)
        steps[pipeline_name] = pipeline_class
        for step_class in pipeline_class.step_defs.values():
            steps[step_class.__name__] = step_class

    return steps
```

stscieisenhamer:
As documented in [crds-381](https://jira.stsci.edu/browse/CRDS-381), the general approach has been deemed sub-optimal. Removing this PR.

PR #5442, opened 2020-10-28
JP-1734: Return correct DataModel subclass when reading an ASDF file

### LINKED ISSUE #5153: Datamodels should be able to read and write ASDF files
Data models have an attribute `DataModel.meta.model_type` which is used to initialize the correct data model from a FITS file. This is written to the FITS keyword `DATAMODEL`. 
 `DataModel.meta.model_type` is also written to an ASDF file but is not taken into acocunt when reading an ASDF file with `datamodels.open`  which does not initialize the correct data model and instead creates an instance of the base `DataModel` class.

Adding this keyword to `datamodels.open` when an asdf file is read will allow using the `open` function with reference files without having to specify the exact model the reference file represents. 

It will also help with using datamodels for the Roman pipeline.

kmacdonald-stsci:
JP-1734: Datamodels should be able to read and write ASDF files

Changed datamodel.open to read an ASDF file and instantiate a datamodel class listed in the ASDF file.

Fixes #5153 / [JP-1734](https://jira.stsci.edu/browse/JP-1734)

### DIFF
--- a/jwst/datamodels/tests/test_open.py
+++ b/jwst/datamodels/tests/test_open.py
@@ -257,3 +257,12 @@ def t_path(partial_path):
     """Construction the full path for test files"""
     test_dir = os.path.join(os.path.dirname(__file__), 'data')
     return os.path.join(test_dir, partial_path)
+
+
+def test_open_asdf_datamodel_class(tmpdir):
+    path = str(tmpdir.join("image_model.asdf"))
+    model = ImageModel((10, 10))
+    model.save(path)
+
+    with datamodels.open(path) as m:
+        assert isinstance(m, ImageModel)
--- a/jwst/datamodels/tests/test_open.py
+++ b/jwst/datamodels/tests/test_open.py
@@ -17,6 +17,8 @@ from jwst.datamodels import (DataModel, ModelContainer, ImageModel,
 from jwst import datamodels
 from jwst.datamodels import util
 
+import asdf
+
 # Define artificial memory size
 MEMORY = 100  # 100 bytes
 
@@ -278,3 +280,15 @@ def test_open_asdf_no_datamodel_class(tmpdir, suffix):
 
     with datamodels.open(path) as m:
         assert isinstance(m, DataModel)
+
+
+def test_open_asdf(tmpdir):
+    path = str(tmpdir.join(f"straight_asdf.asdf"))
+    tree = {"foo": 42, "bar": 13, "seq": np.arange(100)}
+    with asdf.AsdfFile(tree) as af:
+        af.write_to(path)
+
+    with datamodels.open(path) as m:
+        assert isinstance(m, DataModel)
+
+
--- a/jwst/datamodels/tests/test_open.py
+++ b/jwst/datamodels/tests/test_open.py
@@ -283,7 +283,7 @@ def test_open_asdf_no_datamodel_class(tmpdir, suffix):
 
 
 def test_open_asdf(tmpdir):
-    path = str(tmpdir.join(f"straight_asdf.asdf"))
+    path = str(tmpdir.join("straight_asdf.asdf"))
     tree = {"foo": 42, "bar": 13, "seq": np.arange(100)}
     with asdf.AsdfFile(tree) as af:
         af.write_to(path)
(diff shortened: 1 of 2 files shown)


jdavies-st (commented):
Looking good.  You'll want to update the CHANGES.rst file in the top "unreleased" part under a `datamodels` section.

jwst/datamodels/util.py
```diff
@@ -20,6 +20,8 @@
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
+import asdf
```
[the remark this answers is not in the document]
kmacdonald-stsci: I moved this import to above `numpy`.

jwst/datamodels/util.py
```diff
@@ -112,8 +114,19 @@ def open(init=None, memmap=False, **kwargs):
             return container.ModelContainer(init, **kwargs)
 
         elif file_type == "asdf":
-            # Read the file as asdf, no need for a special class
-            return model_base.DataModel(init, **kwargs)
+            # Add a case statement here.
```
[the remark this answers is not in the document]
kmacdonald-stsci: I removed this comment.

jwst/datamodels/util.py
```diff
@@ -195,15 +208,22 @@ def open(init=None, memmap=False, **kwargs):
     return model
 
 
-def _class_from_model_type(hdulist):
+def _class_from_model_type(init):
     """
     Get the model type from the primary header, lookup to get class
```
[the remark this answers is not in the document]
kmacdonald-stsci: Expanded the docstring to include `Parameter` and `Return`.

jwst/datamodels/tests/test_open.py
```diff
@@ -257,3 +257,12 @@ def t_path(partial_path):
     """Construction the full path for test files"""
     test_dir = os.path.join(os.path.dirname(__file__), 'data')
     return os.path.join(test_dir, partial_path)
+
+
+def test_open_asdf_datamodel_class(tmpdir):
+    path = str(tmpdir.join("image_model.asdf"))
```
jdavies-st, line 263: Could you look at parametrizing this test so that it also gets checked for `fits` files?  So something like:

```
@pytest.mark.parametrize("suffix", ["asdf", "fits"])
def test_open_asdf_datamodel_class(tmpdir, suffix):
      path = str(tmpdir.join(f"image_model.{suffix}"))
```

And also check for the ".asdf" case that when there is no `model_type` set, that you get a `DataModel` returned.  Maybe there's already an existing test that is checking the ".fits" case?  Feel free to add a new test that checks both.  Or do it in the same test above as a separate check.

Looking at codecov, those areas of `_class_from_model_type` are not covered.

https://codecov.io/gh/spacetelescope/jwst/compare/a8d5086eb22efc3a19ace4c1dadce63db89f9af2...f4b59215e765618cab1ca4a96dfec4771a20caed/diff

jwst/datamodels/util.py
```diff
@@ -195,15 +208,30 @@ def open(init=None, memmap=False, **kwargs):
...
+    if init:
+        if isinstance(init, fits.hdu.hdulist.HDUList):
+            primary = init[0]
+            model_type = primary.header.get('DATAMODL')
+        elif isinstance(init, asdf.AsdfFile):
+            tree = init.tree
+            if 'meta' in tree and 'model_type' in tree['meta']:
+                model_type = tree['meta']['model_type']
+            else:
+                model_type = None
```
jdavies-st, line 234: Not sure why this line is unreachable by the tests.  Odd.

jdavies-st (approved):
Looks good.  Minor comment below.  Ready to merge I think unless you want to investigate the comment below.

jwst/datamodels/util.py
jdavies-st: Ah, right.  So then maybe that `test_open_asdf_no_datamodel_class` is not doing anything for us.  Maybe better to have a test that writes out a file not using `datamodels`, just `asdf`, then try reading it in.

jwst/datamodels/tests/test_open.py
kmacdonald-stsci: Added parametrization for `test_open_asdf_datamodel_class` and added tests `test_open_asdf_no_datamodel_class` and `test_open_asdf` to increase code coverage.

jwst/datamodels/util.py
kmacdonald-stsci: Added a test that tests `datamodels.open` on an ASDF file created using `asdf.write_to` that doesn't have a datamodel in its tree.

PR #5444, opened 2020-10-28
Remove V1 readthedocs config file

eslavich:
We've had this lingering V1-style config file which only recently started breaking the readthedocs build.  It seems like they must have changed the order of precedence of the files when multiple are present?  In any case, removing the old config seems to fix the build, here's a pass on this branch:

https://readthedocs.org/projects/jwst-pipeline/builds/12209132/

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,4 +1,0 @@
-conda:
-    file: .rtd-environment.yml
-python:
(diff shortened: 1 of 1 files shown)


jdavies-st (approved):
Thanks.  This was an oversight on my part.

PR #5462, opened 2020-11-03
JP-1775: Update flatfield to compute uniform and ps corrections

### LINKED ISSUE #5457: Update FlatField step to compute and store uniform and point source corrections
_Issue [JP-1775](https://jira.stsci.edu/browse/JP-1775) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

To implement the NIRSpec fixed-slit master bkg corrections outlined in JP-1761, the flat-field step has to be updated to compute and store corrections for both uniform and point sources when processing the primary slit of a NIRSpec fixed-slit exposure and the slit contains a point source. The uniform source version of the correction must be computed based on the original (uncorrected) wavelength assignments, which are retrievable from the WCS object, while the point source version of the correction (which gets applied to the data) must be based on the corrected wavelengths stored in the 2D wavelength array.

hbushouse:
Updated the NIRSpec fixed-slit and msa branch of the flatfield step to compute both uniform and point source versions of the flat when processing the primary slit of a fixed-slit exposure and it contains a point source. The uniform source version uses the original wavelengths, uncorrected by the wavecorr step, while the point source version uses the wavecorr results as stored in the 2D wavelength array.

Fixes #5457 / [JP-1775](https://jira.stsci.edu/browse/JP-1775)

### DIFF
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -178,12 +178,12 @@ def apply_flat_field(science, flat, inverse=False):
     bad_flag = dqflags.pixel['DO_NOT_USE'] + dqflags.pixel['NO_FLAT_FIELD']
 
     flat_nan = np.isnan(flat_data)
-    flat_dq[flat_nan] = np.bitwise_or(flat_dq[flat_nan],bad_flag)
+    flat_dq[flat_nan] = np.bitwise_or(flat_dq[flat_nan], bad_flag)
 
     # Find pixels in the flat that have a value of zero, and set
     # DQ = DO_NOT_USE + NO_FLAT_FIELD
     flat_zero = np.where(flat_data == 0.)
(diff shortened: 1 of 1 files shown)


hbushouse:
Note to reviewers: the way this update is setup relies on the fact that as of right now the `wavecorr` step (which precedes `flatfield`) only applies the wavelength corrections to the 2D wavelength array carried along with each slit and does NOT update the WCS object. So this code takes advantage of that by recomputing the 2D wavelength from the WCS object when it needs uncorrected values, while using the existing 2D wavelength array when it needs corrected values. @dencheva (or a new WCS person) will be changing this behavior at some point, so that the `wavecorr` corrections are added to the WCS object as another transform. When that happens, this code will need updating to compute the uncorrected wavelengths from the WCS object by stopping at the appropriate transform (i.e. not including the wavecorr transform).

stscieisenhamer (approved):

PR #5467, opened 2020-11-05
Move utility functions out of datamodels.schema

eslavich:
This PR moves a couple of utility functions out of `datamodels.schema` and into the only modules that use them, so that they can be removed from stdatamodels without impacting jwst.

See https://github.com/spacetelescope/stdatamodels/pull/25


jdavies-st (approved):
🚀

PR #5468, opened 2020-11-06
Add utility asn_gather

stscieisenhamer:
`asn_gather` is a new API and command-line utility that basically allows one to copy an association, and all its members, from one place to another.

### DIFF
--- a/jwst/associations/asn_gather.py
+++ b/jwst/associations/asn_gather.py
@@ -12,7 +12,7 @@ logger.addHandler(logging.NullHandler())
 LogLevels = [logging.WARNING, logging.INFO, logging.DEBUG]
 
 
-def asn_gather(association, destination=None, exp_types=None,
+def asn_gather(association, destination=None, exp_types=None, exclude_types=None,
                shellcmd='rsync -urv --no-perms --chmod=ugo=rwX'):
     """Copy/Move members of an association from one location to another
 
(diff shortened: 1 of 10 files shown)


jwst/associations/asn_gather.py
```diff
@@ -0,0 +1,160 @@
...
+    parser.add_argument(
+        'association',
+        help='Association to gather.'
+    )
+    parser.add_argument(
+        'destination',
+        help='Folder to copy the association to.'
+    )
+    parser.add_argument(
+        '-t', '--exp-types', default=None, dest='exp_types',
```
hbushouse, line 136: You know, we never should've called this "exptype" 5 years ago when we invented it, because there's too much potential for confusion with EXP_TYPE. Should've called it "memtype" (member type) instead. But that's water under the bridge ...

hbushouse (approved):
Works for me.

ddavis-stsci (approved):
Would it be useful to have an option in exp_types to be able to exclude a type. For instance everything but target acquisitions? But this should be for a future
enhancement. Looks good to me.

jwst/associations/asn_gather.py
stscieisenhamer: Have had this thought myself. Though we will probably need to implement backward compatibility, an issue should be filed to make such a change. Will file said issue.

PR #5472, opened 2020-11-06
JP-1745: Remove uncertainities dependency

### LINKED ISSUE #5397: Find and remove/replace the imp module
_Issue [JP-1745](https://jira.stsci.edu/browse/JP-1745) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Some recent PR has introduced a module that uses the deprecated `imp` module. Find and replace by the `importlib` module.

jdavies-st:
The use of the `uncertainties` package is optional in the `jwst.ami` code.  But it brings in the unmaintained package `future` that are raising deprecation warnings in python 3.8+.  So let's not use it.

Resolves #5397 / [JP-1745](https://jira.stsci.edu/browse/JP-1745)

### DIFF
--- a/jwst/ami/leastsqnrm.py
+++ b/jwst/ami/leastsqnrm.py
@@ -3,10 +3,6 @@ import logging
 import numpy as np
 import numpy.linalg as linalg
 from scipy.special import comb, jv
-try:
-    from uncertainties import unumpy
-except ImportError:
-    pass
 
 from . import hexee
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):

stscieisenhamer (approved):
Rhetorical question: Not sure what "optional" means in this situation, but has it been confirmed that pipelines, etc., run without importing this module? Does the import occur only when the "optional" code is executed?

jdavies-st:
The import is in a try block in `jwst.ami.leastsqnrm`, and the code checks to see `if` it has been imported before using it, `else` uses equivalent code that gives equivalent results in the unit tests.  Maybe it would be good to just completely remove all that code and cut down on the amount of untested stuff?

stscieisenhamer:
Agree, if it is easy enough to have no dependency on the module, then extricate!

PR #5479, opened 2020-11-13
JP-1696  Asn_Lv3SpecAux: Add optical element constraint

### LINKED ISSUE #5333: ASN bkg rules and regression test bugs
_Issue [JP-1696](https://jira.stsci.edu/browse/JP-1696) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The level-3 ASN rule `Asn_Lv3SpecAux` is designed to handle level-3 associations that contain dedicated background exposures (BKGDTARG=T) and will handle MIRI MRS, MIRI LRS fixed-slit, and NIRSpec fixed-slit observations containing dedicated bkg exposures. Meanwhile, the `Asn_IFUGratingBkg` rule handles NIRSpec IFU observations with dedicated backgrounds. The latter rule contains a product name override that results in a proper product name template/root that gets stored in the resulting ASN file.

The `Asn_Lv3SpecAux` rule, on the other hand, does **not** have a product name override and hence the default product name template is generated. The problem with this is that the `Asn_Lv3SpecAux` rule does not contain any constraint of type "opt_elem", which causes the routine that creates the default product name to substitute the string "clear" for the optical element suffix part of the product name. This is incorrect, because there are in fact optical elements (grating, prism, filter) in the beam for MIR_MRS, MIR_LRS-FIXEDSLIT, and NRS-FIXEDSLIT modes, hence the "clear" is deceiving. It also then leads to incorrect product name construction when the actual grating/filter/prism names get appended to the product name using an underscore, leading to a list of multiple optical elements that are separated by an underscore, instead of the proscribed hyphen. This in turn leads to file name parsing errors in CAOM during archive ingest.

Finally, some of the input pool files used in asn regression tests have mock observations for science + dedicated background exposures that use the same target_id number for both the science and bkg exposures. The science and bkg exposures need to have different target_id values. This leads to incorrect and sometimes duplicate product names.

stscieisenhamer:
Resolves [JP-1696](https://jira.stsci.edu/browse/JP-1696)
Resolves #5333 

Add optical element constraint, `opt_elem`, to rule Asn_Lv3SpecAux to ensure correct Stage 3 science combining and proper product name creation.

Refactor rules Asn_IFUGrating and Asn_IFUGratingBkg to simplify constraints.

Add method `Constraint.get_all_attr` that retrieves the values for the specified attribute for all defined `Constraints`.

Add property `Constraint.dup_names` returning a list of duplicate assigned names for a given `Constraint`.

Update association tests to with appropriate identifications to better isolate failure cases.


ddavis-stsci (approved):
LGTM

hbushouse (approved):
Looks good to me. The one failing Travis test may be unrelated (problem with WCS somewhere).

Oh, requisite reminder about a change log entry.

stscieisenhamer:
Seriously, the changelog push didn't get in. Will do in the other outstanding pr....

PR #5484, opened 2020-11-16
Enable resample_spec for NIRSpec line lamp exposures

### LINKED ISSUE #5366: Update resample_spec to use MSA frame for NRS lamp exposure outputs
_Issue [JP-1733](https://jira.stsci.edu/browse/JP-1733) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

JP-1427 requests updates to resample_spec, cube_build, and extract_1d to work in "internal" instrument coordinates (MSA plane in this case), rather than sky coords when processing NIRSpec lamp exposures. This ticket covers the updates necessary to the resample_spec step to have it create its output image in the MSA plane, rather than sky coords.

stscirij:
Addresses [JP-1733](https://jira.stsci.edu/browse/JP-1733)

Fixes #5366

### DIFF
--- a/jwst/pipeline/calwebb_spec2.py
+++ b/jwst/pipeline/calwebb_spec2.py
@@ -5,7 +5,7 @@ import traceback
 
 from .. import datamodels
 from ..assign_wcs.util import NoDataOnDetectorError
-from ..lib.exposure_types import is_nrs_ifu_flatlamp, is_nrs_ifu_linelamp
+from ..lib.exposure_types import is_nrs_ifu_flatlamp, is_nrs_ifu_linelamp, is_nrs_slit_linelamp
 from ..stpipe import Pipeline
 
 # step imports
@@ -252,6 +252,11 @@ class Spec2Pipeline(Pipeline):
            and not isinstance(calibrated, datamodels.CubeModel):
 
             # Call the resample_spec step for 2D slit data
+            resampled = self.resample_spec(calibrated)
+
+        elif is_nrs_slit_linelamp(calibrated):
+
+            # Call resample_spec for NRS 2D line lamp slit data
             resampled = self.resample_spec(calibrated)
 
         elif (exp_type in ['MIR_MRS', 'NRS_IFU']) or is_nrs_ifu_linelamp(calibrated):
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -73,6 +73,7 @@ class ResampleSpecData:
 
         # Define output WCS based on all inputs, including a reference WCS
         self.output_wcs = self.build_interpolated_output_wcs()
+        print(self.output_wcs.__repr__)
         self.blank_output = datamodels.SlitModel(self.data_size)
         self.blank_output.update(self.input_models[0], only="PRIMARY")
         self.blank_output.update(self.input_models[0], only="SCI")
@@ -157,15 +158,18 @@ class ResampleSpecData:
                 ra_center_pt = np.nanmean(ra_center)
                 dec_center_pt = np.nanmean(dec_center)
 
-                # ra and dec this converted to tangent projection
-                tan = Pix2Sky_TAN()
-                native2celestial = RotateNative2Celestial(ra_center_pt, dec_center_pt, 180)
-                undist2sky1 = tan | native2celestial
-                # Filter out RuntimeWarnings due to computed NaNs in the WCS
-                warnings.simplefilter("ignore")
-                # at this center of slit find x,y tangent proction - x_tan, y_tan
-                x_tan, y_tan = undist2sky1.inverse(ra, dec)
-                warnings.resetwarnings()
+                if model.meta.wcs.output_frame.name == 'world':
+                    # ra and dec this converted to tangent projection
+                    tan = Pix2Sky_TAN()
+                    native2celestial = RotateNative2Celestial(ra_center_pt, dec_center_pt, 180)
+                    undist2sky1 = tan | native2celestial
+                    # Filter out RuntimeWarnings due to computed NaNs in the WCS
+                    warnings.simplefilter("ignore")
+                    # at this center of slit find x,y tangent proction - x_tan, y_tan
+                    x_tan, y_tan = undist2sky1.inverse(ra, dec)
+                    warnings.resetwarnings()
+                else:
+                    x_tan, y_tan = ra, dec
 
                 # pull out data from center
                 if spectral_axis == 0:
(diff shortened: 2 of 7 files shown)


jwst/resample/resample_spec.py
```diff
@@ -157,15 +157,18 @@ def build_interpolated_output_wcs(self, refmodel=None):
...
+                    tan = Pix2Sky_TAN()
+                    native2celestial = RotateNative2Celestial(ra_center_pt, dec_center_pt, 180)
+                    undist2sky1 = tan | native2celestial
+                    # Filter out RuntimeWarnings due to computed NaNs in the WCS
+                    warnings.simplefilter("ignore")
+                    # at this center of slit find x,y tangent proction - x_tan, y_tan
+                    x_tan, y_tan = undist2sky1.inverse(ra, dec)
+                    warnings.resetwarnings()
+                else:
+                    x_tan, y_tan = ra, dec
```
hbushouse, line 171: Some commentary might be useful here to explain what's happening and why.

hbushouse (commented):
Looks good overall. Just a few minor changes/additions requested. As well as the requisite reminder for a change log entry.

jwst/pipeline/calwebb_spec2.py
```diff
@@ -254,6 +254,11 @@ def process_exposure_product(
             # Call the resample_spec step for 2D slit data
             resampled = self.resample_spec(calibrated)
 
+        elif is_nrs_slit_linelamp(calibrated):
```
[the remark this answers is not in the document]
stscirij: Yes, I could, but then I think the if clause gets overly complicated, IMHO

jwst/resample/resample_spec.py
```diff
@@ -157,15 +157,18 @@ def build_interpolated_output_wcs(self, refmodel=None):
...
-                tan = Pix2Sky_TAN()
-                native2celestial = RotateNative2Celestial(ra_center_pt, dec_center_pt, 180)
-                undist2sky1 = tan | native2celestial
-                # Filter out RuntimeWarnings due to computed NaNs in the WCS
-                warnings.simplefilter("ignore")
-                # at this center of slit find x,y tangent proction - x_tan, y_tan
-                x_tan, y_tan = undist2sky1.inverse(ra, dec)
-                warnings.resetwarnings()
+                if model.meta.wcs.output_frame.name == 'world':
+                    # ra and dec this converted to tangent projection
```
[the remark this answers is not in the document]
stscirij: This is a holdover from before my changes, I'll clean up

jwst/resample/resample_spec.py
stscirij: Again, will comment to clarify, here and elsewhere

jwst/resample/resample_spec.py
```diff
@@ -157,15 +157,20 @@ def build_interpolated_output_wcs(self, refmodel=None):
...
+                    # convert ra and dec to tangent projection
+                    tan = Pix2Sky_TAN()
+                    native2celestial = RotateNative2Celestial(ra_center_pt, dec_center_pt, 180)
+                    undist2sky1 = tan | native2celestial
+                    # Filter out RuntimeWarnings due to computed NaNs in the WCS
+                    warnings.simplefilter("ignore")
+                    # at this center of slit find x,y tangent projection - x_tan, y_tan
+                    x_tan, y_tan = undist2sky1.inverse(ra, dec)
+                    warnings.resetwarnings()
+                else:
```
nden, line 170: I wonder if this `else` statement is too general. If the general idea is to have coordinates on a plane and If the output_frame is `V2V3` the code may not work correctly since `V2V3` is a spherical system. I'm not sure if it's better to exclude `V2V3` or to consider only `msa_frame`. The first option allows for other frames to work so it may be prefferable.

jwst/resample/resample_spec.py
```diff
@@ -289,8 +294,11 @@ def build_interpolated_output_wcs(self, refmodel=None):
         native2celestial = RotateNative2Celestial(ra_center_final, dec_center_final, 180)
         undist2sky = tan | native2celestial
 
-        # find the spatial size of the output - same in x,y
-        x_tan_all, y_tan_all = undist2sky.inverse(all_ra, all_dec)
+        if model.meta.wcs.output_frame.name == 'world':
+            # find the spatial size of the output - same in x,y
+            x_tan_all, y_tan_all = undist2sky.inverse(all_ra, all_dec)
```
nden, line 299: The definition of `undist2sky` can go in the if statement.

jwst/resample/resample_spec.py
```diff
@@ -301,11 +309,17 @@ def build_interpolated_output_wcs(self, refmodel=None):
...
+            transform = mapping | (pix_to_xtan & pix_to_ytan) & pix_to_wavelength
 
         det = cf.Frame2D(name='detector', axes_order=(0, 1))
-        sky = cf.CelestialFrame(name='sky', axes_order=(0, 1),
-                                reference_frame=coord.ICRS())
+        if model.meta.wcs.output_frame.name == 'world':
+            sky = cf.CelestialFrame(name='sky', axes_order=(0, 1),
+                                    reference_frame=coord.ICRS())
+        else:
+            sky = cf.Frame2D(name='sky', axes_order=(0, 1))
```
nden, line 322: The name is a bit misleading since it's definitely not `sky`. Is something like 
`resampled_+ {model.meta.wcs.output_frame.name}` better?

jwst/resample/resample_spec.py
stscirij: It's a tricky one - currently resample_spec only 'knows' how to do sky/world and msa_frame, so we could just test those cases and punt if there's a different output frame.  Alternatively, maybe we could do something like
`if Unit("deg") in model.meta.wcs.output_frame.unit or Unit("arcsec") in model.meta.wcs.output_frame.unit:`
to capture the sky-like cases, and otherwise do the cartesian cases

jwst/resample/resample_spec.py
stscirij: Yes, I thought the naming was only temporary for the purposes of doing the transformation, but now I see that it appears in the meta.wcs.output_frame of the slits in the _s2d file.

stscieisenhamer (approved):
Maybe worth a separate issue to refactor that `build_interpolated_output_wcs` to try to cut down the special case handling.

hbushouse (approved):
LGTM

PR #5486, opened 2020-11-16
Regtest: Remove superfluous test_miri_image_detector1

jdavies-st:
We have coverage of the MIRI Imaging detector1 pipeline here

https://github.com/spacetelescope/jwst/blob/16432c3179439a9cd0d50e2187ce4d1ccb349945/jwst/regtest/test_miri_image.py#L76-L82

When merged, remove the test and truth data on bytesalad.


hbushouse:
I think the reason we have 2 detector1 tests for MIRI is because one of them is full-frame and the other is subarray (perhaps the LRS Slitless subarray). There's enough different processing in some of the steps to warrant testing both.

hbushouse (approved):
OK, I guess you're right that this test really does duplicate the one in the test_miri_image.py module. They use different exposures, but both exposures are regular full-frame imaging and have all other similar characteristics, such that they test all the same steps and branches within steps. Plus this one has many more groups per integration and hence takes longer to run.

So I guess it's fine to deep six this one.

Note that there are a couple of different permutations of the file name(s) sitting on artifactory (upper-case and lower-case), so all of them should be removed.

hbushouse:
Note that a MIRI subarray exposure gets tested in calwebb_detector1 in the test_miri_lrs_slitless.py module, so we have coverage for that too.

PR #5489, opened 2020-11-18
JP-1755: Rename class RSCD_Step to preserve CRDS functionality. (#5424)

### LINKED ISSUE #5424: Rename class RSCD_Step to RscdStep
_Issue [JP-1755](https://jira.stsci.edu/browse/JP-1755) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Currently, the RSCD step has the class name `RSCD_Step`. Besides being conventionally not-so-right, being mixed CamelCase with underscore, this form also breaks the CRDS functionality that retrieves step parameter reference files: In CRDS, underscores mean something in reftype names, and this underscore adds one too many.

Rename to `RscdStep`.

tapastro:
Fixes #5424 to rename RSCD step.

Addresses JP-1755.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,8 @@ jobs:
 
     - name: Dev dependencies in requirements-dev.txt
       install:
-        - pip install -r requirements-dev.txt -e .[test]
+        - pip install -r requirements-dev.txt
+        - pip install -e .[test]
 
     - name: Warnings treated as Exceptions
       script:
@@ -94,7 +95,8 @@ jobs:
   allow_failures:
     - name: Dev dependencies in requirements-dev.txt
       install:
-        - pip install -r requirements-dev.txt -e .[test]
+        - pip install -r requirements-dev.txt
+        - pip install -e .[test]
 
     - name: Warnings treated as Exceptions
       script:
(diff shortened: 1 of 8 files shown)


jemorrison (approved):
Looks good to me.

jdavies-st:
This change from `RSCD_Step` to `RscdStep` will have a knock-on effect on the default suffix that products from the step will have.  it will now be `_rscdstep.fits` instead of `rscd_step.fits`.

That's what the following test failure is telling you

https://travis-ci.com/github/spacetelescope/jwst/jobs/442552646#L735-L755

There's also some regression tests in `jwst/regtest` that check for outputs based on suffix.  Might want to check those as well and make sure they pass, and if not, change the expected suffix.

stscieisenhamer (approved):

tapastro:
The addition of a deprecated step name (to allow for outdated step name calls to still run, for those who have code calling RSCD_Step) has caused an extra suffix to be found by lib/suffix.find_suffixes(). As the deprecated step class will not create any files with this suffix, I've added the suffix to the list of SUFFIXES_TO_DISCARD. 

These suffixes are removed from the set of suffixes using the set method difference_update, which does not throw an error if the set does not include the item being discarded, so this list may gather clutter unless maintained.

jdavies-st (approved):
Looks great!

And it looks like since the regression tests use the config files, the suffix there is keyed off `name` in the default config file, so we get `jw1234_blah_blah_rscd.fits`.  Fun!

PR #5492, opened 2020-11-18
Implement utility function all_steps and fix crds reference file retrieval for non-datamodels

stscieisenhamer:
This is code that is fallout from various other issues, JP-1751, CCD-381, and CCD-783.

The changes are

- `jwst.stpipe.utilities.all_steps`: Return all classes defined in the package that are subclassed from `Step`. This includes a slight refactor of the `suffix` code.
- Bug fix in `jwst.stpipe.Step.get_config_from_reference` to not attempt retrieval of step parameter references files when input is not a `DataModel`.

### DIFF
--- a/jwst/stpipe/utilities.py
+++ b/jwst/stpipe/utilities.py
@@ -49,6 +49,7 @@ NON_STEPS = [
     'Pipeline',
     'Step',
     'SystemCall',
+    'TestLinearPipeline',
 ]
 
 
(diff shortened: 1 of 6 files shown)


jwst/stpipe/utilities.py
```diff
@@ -29,10 +29,29 @@
 """
 Utilities
 """
+from copy import copy
+from importlib import import_module
 import inspect
+import logging
 import os
+from os import path
```
tapastro, line 37: File contains sys.path, os.path and path calls - might clean this up for readability (suggest os.path?).

tapastro (approved):

PR #5498, opened 2020-11-30
Allow packages installed from GitHub to pass through the minimum_deps script

eslavich:
This will hopefully prevent the "oldest dependency versions" job from failing when a package (like stdatamodels) is being installed from GitHub instead of pypi.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,6 @@ jobs:
       install:
         - pip install -U pip certifi
         - pip install -r requirements-min.txt -e .[test]
-        # Since there is no version of stdatamodels on PyPI, install from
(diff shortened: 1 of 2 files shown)


jdavies-st (approved):
🚀

jdavies-st:
Yeah, looks like the old code was a stub that never worked.  Glad this is sorted.  Should work with all versions of `pip`.

jdavies-st:
Looks like we could remove the

https://github.com/spacetelescope/jwst/blob/6ce0ea6536294124b0c0047bc0689cd4d157da14/.travis.yml#L40-L42

lines as well.

stscieisenhamer (approved):

PR #5500, opened 2020-11-30
JP-1763: fix dq flag handling in wfs_combine

### LINKED ISSUE #5435: wfs_combine is setting all pixels with  DQ>0 as DO_NOT_USE pixels
_Issue [JP-1763](https://jira.stsci.edu/browse/JP-1763) was created on JIRA by [Marcio Melendez Hernandez](https://jira.stsci.edu/secure/ViewProfile.jspa?name=melendez):_

I have been creating WFSC combined products, using  `wfs_combine` which is a step that combines a pair of dithered WFSC. Per the [documentation](https://jwst-pipeline.readthedocs.io/en/stable/jwst/wfs_combine/main.html?highlight=_wfscmb#creating-the-combined-image), it seems that bad pixels in the combined product are defined as having DQ>0 but it should be DQ=1, otherwise, the final combined image could have more “DO_NOT_USE” pixels with SCI values of zero than the individual dither pointing. In the attached example, each dither point have 0.076% of “DO_NOT_USE”  pixels, whereas the **_wfscmb** image has 3.76% of bad pixels because is adding a flag to everything where both dither point have DQ>0.  This could makes a _wfscmb image unusable for science because of all the zero values/(DQ=1). In the attached image, blue pixels are those flag as “DO_NOT_USE”.

!image-2020-10-27-13-10-58-593.png!

jemorrison:
Update DQ handling in `wfs_combine` so that only pixels with the DO_NOT_USE flag are considered bad, rather than any pixel with DQ > 0.

Fixes [JP-1763](https://jira.stsci.edu/browse/JP-1763)
Fixes #5435

### DIFF
--- a/jwst/wfs_combine/wfs_combine.py
+++ b/jwst/wfs_combine/wfs_combine.py
@@ -139,8 +139,8 @@ class DataSet():
             # 1a. create image to smooth by first setting bad DQ pixels equal
             #     to mean of good pixels
             data_1 = self.input_1.data.astype(np.float64)
-            wh_bad_1 = np.where(np.bitwise_and(self.input_1.dq, dqflags.pixel['DO_NOT_USE']) == 1)
-            wh_good_1 = np.where(np.bitwise_and(self.input_1.dq, dqflags.pixel['DO_NOT_USE']) == 0)
+            wh_bad_1 = np.where(np.bitwise_and(self.input_1.dq, dqflags.pixel['DO_NOT_USE']))
+            wh_good_1 = np.where(np.logical_not(np.bitwise_and(self.input_1.dq, dqflags.pixel['DO_NOT_USE'])))
 
             data_1[wh_bad_1] = data_1[wh_good_1].mean()
 
@@ -323,10 +323,10 @@ class DataSet():
         err_data_1 = im_1_a.err.copy()
         err_data_2 = im_2_a.err.copy()
 
-        bad1 = np.bitwise_and(dq_data_1, dqflags.pixel['DO_NOT_USE']) = 1
-        good1 = np.bitwise_and(dq_data_1, dqflags.pixel['DO_NOT_USE']) = 0
-        bad2 = np.bitwise_and(dq_data_2, dqflags.pixel['DO_NOT_USE']) = 1
-        good2 = np.bitwise_and(dq_data_2, dqflags.pixel['DO_NOT_USE']) = 0
+        bad1 = np.bitwise_and(dq_data_1, dqflags.pixel['DO_NOT_USE'])
+        good1 = np.logical_not(np.bitwise_and(dq_data_1, dqflags.pixel['DO_NOT_USE']))
+        bad2 = np.bitwise_and(dq_data_2, dqflags.pixel['DO_NOT_USE'])
+        good2 = np.logical_not(np.bitwise_and(dq_data_2, dqflags.pixel['DO_NOT_USE']))
 
         wh_1_bad_2_good = bad1 & good2
         wh_1_bad_2_bad =  bad1 & bad2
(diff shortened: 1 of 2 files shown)


jwst/wfs_combine/wfs_combine.py
```diff
@@ -139,8 +139,9 @@ def create_aligned_2(self):
             # 1a. create image to smooth by first setting bad DQ pixels equal
             #     to mean of good pixels
             data_1 = self.input_1.data.astype(np.float64)
-            wh_bad_1 = np.where(self.input_1.dq != 0)
-            wh_good_1 = np.where(self.input_1.dq == 0)
+            wh_bad_1 = np.where(np.bitwise_and(self.input_1.dq, dqflags.pixel['DO_NOT_USE']) == 1)
```
hbushouse, line 142: Travis is blowing up on these statements. You don't need the `==1` qualification, because `np.bitwise_and()` already equates to 1=True for pixels that satisfy the condition. All you need is:

wh_bad_1 = np.where(np.bitwise_and(self.input_1.dq, dqflags.pixel['DO_NOT_USE']))

jwst/wfs_combine/wfs_combine.py
```diff
@@ -139,8 +139,9 @@ def create_aligned_2(self):
...
+            wh_good_1 = np.where(np.bitwise_and(self.input_1.dq, dqflags.pixel['DO_NOT_USE']) == 0)
+
```
hbushouse, line 144: Need similar change as above, but in this case you just want:

wh_good_1 = np.where(np.logical_not(np.bitwise_and(self.input_1.dq, dqflags.pixel['DO_NOT_USE'])))

jwst/wfs_combine/wfs_combine.py
```diff
@@ -322,10 +323,15 @@ def create_combined(self, im_1_a, im_2_a):
         err_data_1 = im_1_a.err.copy()
         err_data_2 = im_2_a.err.copy()
 
-        wh_1_bad_2_good = (dq_data_1 != 0) & (dq_data_2 == 0)
-        wh_1_bad_2_bad = (dq_data_1 != 0) & (dq_data_2 != 0)
-        wh_1_good_2_good = (dq_data_1 == 0) & (dq_data_2 == 0)
-        wh_1_good_2_bad = (dq_data_1 == 0) & (dq_data_2 != 0)
+        bad1 = np.bitwise_and(dq_data_1, dqflags.pixel['DO_NOT_USE']) = 1
```
hbushouse, line 326: same changes as above, just without the `np.where()`

jwst/wfs_combine/wfs_combine.py
```diff
@@ -433,7 +439,8 @@ def interp_array(sci_data, dq_data):
         interpolated SCI image
     """
 
-    wh_bad_dq = np.where(dq_data != 0)
+
+    wh_bad_dq = np.where(np.bitwise_and(dq_data, dqflags.pixel['DO_NOT_USE']) == 1)
```
hbushouse, line 443: Same change needed here

hbushouse (changes requested):

hbushouse (approved):
Looks good now.

jemorrison:
@hbushouse  is it ok to merge this PR. I think the 2 failing test are just code coverage

jwst/wfs_combine/wfs_combine.py
```diff
@@ -322,10 +323,15 @@ def create_combined(self, im_1_a, im_2_a):
         err_data_1 = im_1_a.err.copy()
         err_data_2 = im_2_a.err.copy()
 
-        wh_1_bad_2_good = (dq_data_1 != 0) & (dq_data_2 == 0)
-        wh_1_bad_2_bad = (dq_data_1 != 0) & (dq_data_2 != 0)
-        wh_1_good_2_good = (dq_data_1 == 0) & (dq_data_2 == 0)
-        wh_1_good_2_bad = (dq_data_1 == 0) & (dq_data_2 != 0)
+        bad1 = np.bitwise_and(dq_data_1, dqflags.pixel['DO_NOT_USE'])
```
jdavies-st, line 326: `np.bitwise_and` is returning an array of 0 and 1, so a bitmask.  This is then used to index another array as if it were a boolean array below, which it is not.  This needs to be converted to a boolean if it is to be used for indexing, otherwise it is just changing indexes 0 and 1 multiple times.

Same for 2 lines down.

PR #5504, opened 2020-12-02
JP-1466 Restrict association creation based on optical path for NIRSpec Fixed-slit and IFU

### LINKED ISSUE #4971: Rules for excluding NIRSpec NRS1 or NRS2 detector from processing
_Issue [JP-1466](https://jira.stsci.edu/browse/JP-1466) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

For NIRSpec fixed slit mode with certain disperser and slit combinations, spectra do not fall on one of the two detectors.  In these instances, the calwebb_spec3 pipeline currently crashes because no level 2b products are generated.  The level 3 associations should not include any files for the detector with no data.  Here are the relevant combinations for which the indicated detector contains **no** data and should be **excluded** from the association:

 

NRS1:

slit S200B1 + PRISM/CLEAR, G140M/F100LP, G235M/F170LP, G395M/F290LP, G140H/F100LP, G235H/F170LP, G395H/F290LP

 

NRS2:

all slits other than S200B1 + PRISM/CLEAR, G140M/F070LP, G140M/F100LP, G235M/F170LP, G395M/F290LP, G140H/F070LP

stscieisenhamer:
Resolves [JP-1466](https://jira.stsci.edu/browse/JP-1466)
Fixes #4971


ddavis-stsci (approved):
Looks good.

jwst/associations/asn_gather.py
```diff
@@ -35,6 +36,10 @@ def asn_gather(association, destination=None, exp_types=None, exclude_types=None
     exclude_types : [str[,...]] or None
         List of exposure types to exclude.
 
+    source_folder : str or None
+       Folder where the members originate from.
```
hbushouse, line 40: I guess we can let you get away with ending a sentence with a preposition, after all it's only a doc string ...  :-)

jwst/associations/lib/dms_base.py
```diff
@@ -803,3 +832,20 @@ def item_getattr(item, attributes, association=None):
...
+    try:
+        _, detector = item_getattr(item, ['detector'])
+        _, filter = item_getattr(item, ['filter'])
+        _, grating = item_getattr(item, ['grating'])
+        _, slit = item_getattr(item, ['fxd_slit'])
+    except KeyError:
+        return False
+
+    # Reduce all A slits to just 'a'.
+    if slit != 's200b1':
```
hbushouse, line 848: Is the slit name guaranteed to be in lower case?

hbushouse (approved):
LGTM

Requisite reminder about change log entry.

hbushouse:
looks like some unit tests need fixing as a result of these changes

jwst/associations/asn_gather.py
stscieisenhamer: 🤦

jwst/associations/lib/dms_base.py
stscieisenhamer: yes, the association architecture forces all this. It is encapsulated in the `AssociationPool` class.

PR #5507, opened 2020-12-03
JP-1347: Add SIP approximation to meta.wcsinfo for imaging modes

### LINKED ISSUE #4673: Add SIP solution to JWST WCS
Issue [JP-1347](https://jira.stsci.edu/browse/JP-1347) was created by Steven Crawford:

Recent updates to gWCS have enabled a tool that can convert a gWCS WCS into a SIPs WCS.    This potentially provides a more accurate FITS WCS though there will still be some limitations.   

This could be added as a step in the processing pipeline to provide the SIP solution so that a more accurate WCS could be supplied with our calibrated, un-resampled data products.

cshanahan1:
This adds a SIP approximation to the assign_wcs step for imaging modes.

gwcs.to_fits_sip is used to compute the linear & polynomial distortion terms. The degree of the polynomial is not supplied, but chosen from the best fit of degrees <= 6. The PC/cdelt/naxis keywords in wcsinfo are removed, and the CD matrix terms as well as all the polynomial coefficients are added. The updated wcsinfo schema includes these new keywords. 

@cshanahan1 Remaining tasks:

- [x] Add a `True/False` parameter to `assign_wcs` which enables/disables the approximation
- [x] Add a test which creates a FITS WCS object, evaluates it and compares the results to the GWCS object (I'm curious what the difference is, should be small).

- [x] Figure out what conditions creates a FITS WCS header without SIP coefficients (The Travis failure)

- [x] Add the rest of the `to_fits_sip` arguments as parameters to the `assign_wcs` step. These (with defaults) are: `max_pix_error=0.25`, `degree=None`,`max_inv_pix_error=0.25`, `inv_degree=None`,`npoints=32`

- [x] Add a Changelog entry

Resolves #4673 / [JP-1347](https://jira.stsci.edu/browse/JP-1347)

### DIFF
--- a/docs/jwst/assign_wcs/main.rst
+++ b/docs/jwst/assign_wcs/main.rst
@@ -22,6 +22,12 @@ input frame ``detector`` to a frame ``v2v3``. This part of the WCS pipeline may 
 intermediate coordinate frames. The basic WCS keywords are used to create
 the transform from frame ``v2v3`` to frame ``world``.
 
+For image display with software like DS9 that relies on specific WCS information, a SIP-based
+approximation to the WCS is fit. The results are FITS keywords stored in
+``model.meta.wcsinfo``. This is not an exact fit, but is accurate to ~0.25 pixel and is sufficient
+for display purposes. This step, which occurs for imaging modes early, is performed by default but
+can be switched off, and parameters controlling the fit can also be adjusted. 
+
 
 
 Basic WCS keywords and the transform from ``v2v3`` to ``world``
--- a/jwst/assign_wcs/assign_wcs_step.py
+++ b/jwst/assign_wcs/assign_wcs_step.py
@@ -1,6 +1,7 @@
 #! /usr/bin/env python
 from ..stpipe import Step
 from .. import datamodels
+from ..lib.exposure_types import IMAGING_TYPES
 import logging
 from .assign_wcs import load_wcs
 from .util import MSAFileError
@@ -41,6 +42,12 @@ class AssignWcsStep(Step):
     """
 
     spec = """
+        sip_approx = boolean(default=True)  # enables SIP approximation for imaging modes.
+        sip_max_pix_error = float(default=0.25)  # max err for SIP fit, forward.
+        sip_degree = integer(max=6, default=None)  # degree for forward SIP fit, None to use best fit.
+        sip_max_inv_pix_error = float(default=0.25)  # max err for SIP fit, inverse.
+        sip_inv_degree = integer(max=6, default=None)  # degree for inverse SIP fit, None to use best fit.
+        sip_npoints = integer(default=32)  #  number of points for SIP
         slit_y_low = float(default=-.55)  # The lower edge of a slit.
         slit_y_high = float(default=.55)  # The upper edge of a slit.
 
@@ -83,5 +90,44 @@ class AssignWcsStep(Step):
             slit_y_range = [self.slit_y_low, self.slit_y_high]
             result = load_wcs(input_model, reference_file_names, slit_y_range)
 
+        if not (result.meta.exposure.type.lower() in IMAGING_TYPES and self.sip_approx):
+            return result
+
+        # fit sip approx., degree is chosen by best fit
+        try:
+            fit_sip_hdr = result.meta.wcs.to_fits_sip(None, self.sip_max_pix_error,
+                          self.sip_degree, self.sip_max_inv_pix_error, self.sip_inv_degree,
+                          self.sip_npoints)
+        except ValueError:
+            return result
+
+        # if a polynomial was fit (not just linear terms), but the
+        # best fit > 6th degree, fit a 6th degree instead.
+        # and give up if that doesn't work.
+        if 'SIP' in fit_sip_hdr['ctype1'] and (fit_sip_hdr['a_order'] > 6 or fit_sip_hdr.get('ap_order', 1) > 6):
+            fwd_deg = min(fit_sip_hdr['a_order'], 6)
+            inv_deg = min(fit_sip_hdr.get('ap_order', 1), 6)
+            try:
+                fit_sip_hdr = result.meta.wcs.to_fits_sip(None, self.sip_max_pix_error,
+                              fwd_deg, self.sip_max_inv_pix_error, inv_deg, self.sip_npoints)
+            except ValueError:
+                return result
+
+        # maintain convention of lowercase keys
+        fit_sip_hdr = {k.lower(): v for k, v in fit_sip_hdr.items()}
+
+        # update meta.wcs_info with fit keywords except for naxis*
+        for key in ['naxis1', 'naxis2']:
+            del fit_sip_hdr[key]
+
+        # update meta.wcs_info with fit keywords
+        result.meta.wcsinfo.instance.update(fit_sip_hdr)
+
+        # delete naxis, cdelt, pc from wcsinfo
+        rm_keys = ['naxis', 'cdelt1', 'cdelt2',
+                    'pc1_1','pc1_2', 'pc2_1', 'pc2_2']
+        for key in rm_keys:
+            if key in result.meta.wcsinfo.instance:
+                del result.meta.wcsinfo.instance[key]
 
         return result
--- a/jwst/assign_wcs/tests/test_wcs.py
+++ b/jwst/assign_wcs/tests/test_wcs.py
@@ -1,13 +1,53 @@
 import pytest
+from numpy import zeros
 from numpy.testing import assert_allclose
 from astropy import units as u
 from astropy import wcs
 from astropy.io import fits
 from asdf.tests import helpers
 
-from jwst.assign_wcs import pointing
+from jwst.assign_wcs import AssignWcsStep, pointing
 from jwst.transforms import models
-from jwst.datamodels import ImageModel, CubeModel
+from jwst.datamodels import ImageModel, CubeModel, open
+from gwcs.wcstools import grid_from_bounding_box
+
+
+def create_hdul(wcskeys={
+                      'wcsaxes': 2,
+                      'ra_ref': 22.02351763251896,
+                      'dec_ref': 11.99875540218638,
+                      'v2_ref': 86.039011,
+                      'v3_ref': -493.385704,
+                      'roll_ref': 0.005076934167039675},
+                data_shape=(2048, 2048)):
+
+    """Create nircam hdulist specifically for the SIP test."""
+    hdul = fits.HDUList()
+    phdu = fits.PrimaryHDU()
+    phdu.header['DATAMODL'] = 'ImageModel'
+    phdu.header['TELESCOP'] = "JWST"
+    phdu.header['FILENAME'] = "test+F444W"
+    phdu.header['INSTRUME'] = 'NIRCAM'
+    phdu.header['CHANNEL'] = 'LONG'
+    phdu.header['DETECTOR'] = 'NRCALONG'
+    phdu.header['FILTER'] = 'F444W'
+    phdu.header['PUPIL'] = 'CLEAR'
+    phdu.header['MODULE'] = 'A'
+    phdu.header['TIME-OBS'] = '16:58:27.258'
+    phdu.header['DATE-OBS'] = '2021-10-25'
+    phdu.header['EXP_TYPE'] = 'NRC_IMAGE'
+    scihdu = fits.ImageHDU()
+    scihdu.header['EXTNAME'] = "SCI"
+    scihdu.header['SUBARRAY'] = 'FULL'
+
+    scihdu.header.update(wcskeys)
+
+    scihdu.data = zeros(data_shape)
+
+    hdul.append(phdu)
+    hdul.append(scihdu)
+
+    return hdul
 
 
 @pytest.fixture
@@ -83,15 +123,15 @@ def test_v23_to_sky():
     """
     Test taken from INS report.
     """
-    ra_ref = 165 # in deg
-    dec_ref = 54 # in deg
-    v2_ref = -503.654472 / 3600 # in deg
-    v3_ref = -318.742464 / 3600 # in deg
-    r0 = 37 # in deg
+    ra_ref = 165  # in deg
+    dec_ref = 54  # in deg
+    v2_ref = -503.654472 / 3600  # in deg
+    v3_ref = -318.742464 / 3600  # in deg
+    r0 = 37  # in deg
 
-    v2 = 210 # in deg
-    v3 = -75 # in deg
-    expected_ra_dec = (107.12810484789563, -35.97940247128502) # in deg
+    v2 = 210  # in deg
+    v3 = -75  # in deg
+    expected_ra_dec = (107.12810484789563, -35.97940247128502)  # in deg
     angles = [-v2_ref, v3_ref, -r0, -dec_ref, ra_ref]
     axes = "zyxyz"
     v2s = models.V23ToSky(angles, axes_order=axes)
@@ -155,3 +195,93 @@ def test_create_fitswcs(tmpdir, create_model_3d):
     # test serialization
     tree = {'wcs': w3d}
     helpers.assert_roundtrip_tree(tree, tmpdir)
+
+
+def test_sip_approx(tmpdir):
+    # some of the wcs info
+    true_wcs = {
+                'ctype1': 'RA---TAN-SIP',
+                'ctype2': 'DEC--TAN-SIP',
+                'crpix1': 1025.0,
+                'crpix2': 1025.0,
+                'crval1': 22.023508708126627,
+                'crval2': 11.998764139672561,
+                'cd1_1': -1.7436649443640038e-05,
+                'cd1_2': -2.0849272804435688e-08,
+                'cd2_1': -4.616114099568815e-08,
+                'cd2_2': 1.7519905482510155e-05,
+                'a_order': 3,
+                'b_order': 3,
+                'a_0_2': -1.5520006349224245e-06,
+                'a_0_3': 4.3714933936267264e-13,
+                'a_1_1': -1.1604291350572094e-05,
+                'a_1_2': 1.7125037791723067e-09,
+                'a_2_0': 1.9225112320436695e-06,
+                'a_2_1': -7.926931183344235e-11,
+                'a_3_0': 1.5974716218080076e-09,
+                'b_0_2': -6.7554035599749904e-06,
+                'b_0_3': 1.6911444547233376e-09,
+                'b_1_1': 3.6345928479129676e-06,
+                'b_1_2': -8.64927324063469e-11,
+                'b_2_0': 4.915411904251364e-06,
+                'b_2_1': 1.5701011999755348e-09,
+                'b_3_0': -2.9577497324389284e-12,
+                'sipmxerr': 3.950294659161568e-11,
+                'sipiverr': 0.11047793758675752,
+                'ap_0_1': 3.988114493629691e-07,
+                'ap_0_2': 1.5486628623564042e-06,
+                'ap_0_3': 3.784084622814559e-11,
+                'ap_1_0': -7.65706188115184e-06,
+                'ap_1_1': 1.149989280312782e-05,
+                'ap_1_2': -1.4994544929498265e-09,
+                'ap_2_0': -1.904283418222981e-06,
+                'ap_2_1': -4.507459210843236e-11,
+                'ap_3_0': -1.630572028632995e-09,
+                'bp_0_1': -7.46518175488653e-06,
+                'bp_0_2': 6.690996263911794e-06,
+                'bp_0_3': -1.5898038512674783e-09,
+                'bp_1_0': 4.524878821020556e-07,
+                'bp_1_1': -3.60126568682908e-06,
+                'bp_1_2': -4.4691036548847474e-11,
+                'bp_2_0': -4.9044713849991446e-06,
+                'bp_2_1': -1.7110976247273783e-09,
+                'bp_3_0': 3.895381500452847e-11,
+                'ap_order': 3,
+                'bp_order': 3}
+
+    hdu1 = create_hdul()
+    im = ImageModel(hdu1)
+
+    pipe = AssignWcsStep()
+    result = pipe.call(im)
+
+    # check that result.meta.wcsinfo has correct
+    # values after SIP approx.
+    wcs_info = result.meta.wcsinfo.instance
+
+    # make sure all expected keys are there
+    assert set(true_wcs.keys()).issubset(set((wcs_info.keys())))
+
+    # and make sure they match
+    for key in true_wcs:
+        if 'ctype' in key:
+            assert true_wcs[key] == wcs_info[key]
+        else:
+            assert_allclose(true_wcs[key], wcs_info[key])
+
+    # evaulate fits wcs and gwcs, make sure they agree
+    grid = grid_from_bounding_box(result.meta.wcs.bounding_box)
+    gwcs_ra, gwcs_dec = result.meta.wcs(*grid)
+    fits_wcs = wcs.WCS(wcs_info)
+    fitswcs_res = fits_wcs.pixel_to_world(*grid)
+
+    assert_allclose(fitswcs_res.ra.deg, gwcs_ra)
+    assert_allclose(fitswcs_res.dec.deg, gwcs_dec, atol=1.5e-6)
+
+    # now write the file out, read it back in, and check that the fit values are preserved
+    path = str(tmpdir.join("tmp_sip_wcs.fits"))
+    result.write(path)
+
+
+    with open(path) as result_read:
+        result.meta.wcsinfo == result_read.meta.wcsinfo
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,8 @@ assign_wcs
 
 - Enable resample_spec for NIRSpec line lamp exposures [#5484]
 
+- Added SIP approximation to WCS for imaging modes. FITS WCS keywords added to meta.wcsinfo. [#5507]
+
 associations
 ------------
 
--- a/jwst/datamodels/schemas/wcsinfo.schema.yaml
+++ b/jwst/datamodels/schemas/wcsinfo.schema.yaml
@@ -215,6 +215,30 @@ properties:
             fits_keyword: PC3_3
             fits_hdu: SCI
             blend_table: True
+          cd1_1:
+            title: linear transformation matrix element, CD matrix
+            type: number
+            fits_keyword: CD1_1
+            fits_hdu: SCI
+            blend_table: True
+          cd1_2:
+            title: linear transformation matrix element, CD matrix
+            type: number
+            fits_keyword: CD1_2
+            fits_hdu: SCI
+            blend_table: True
+          cd2_1:
+            title: linear transformation matrix element, CD matrix
+            type: number
+            fits_keyword: CD2_1
+            fits_hdu: SCI
+            blend_table: True
+          cd2_2:
+            title: linear transformation matrix element, CD matrix
+            type: number
+            fits_keyword: CD2_2
+            fits_hdu: SCI
+            blend_table: True
           ps3_0:
             title: Coordinate table extension name
             type: string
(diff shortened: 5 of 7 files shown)


jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -83,5 +86,48 @@ def process(self, input, *args, **kwargs):
...
+                    # check if polynomial terms were actually fit, or just linear.
+                    sip = False or ('SIP' in fit_sip_hdr['ctype1'])
+
+                    # if best fit > 6th degree, fit a 6th degree instead.
+                    # and give up if that doesn't work.
+                    if sip:
+                        if fit_sip_hdr['a_order'] > 6:
+                            try:
+                                fit_sip_hdr = result.meta.wcs.to_fits_sip(None, max_pix_error,
+                                              6, max_inv_pix_error, 6, npoints)
```
mcara, line 116: Even though it is not likely, in principle it is possible that the degree for the inverse polynomial be smaller than for the forward one. The logic here forces inverse to be 6 (even though it might be 5) if forward's degree is > 6.

NOTE: I do not think this is critical to be fixed.

In addition, I wish, as suggested in https://github.com/spacetelescope/gwcs/pull/286#issuecomment-592650047, `degree` could be a list of degrees that `to_fits_sip()` would consider for fitting. Then this logic here would not be necessary at all if you could set `degree=[2,3,4,5,6]`. CC: @perrygreenfield

mcara (changes requested):

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -51,7 +52,9 @@ class AssignWcsStep(Step):
                             'fore', 'fpa', 'msa', 'ote', 'ifupost',
                             'ifufore', 'ifuslicer']
 
-    def process(self, input, *args, **kwargs):
+    def process(self, input, sip_approx=True, max_pix_error=0.25,
+                degree=None, max_inv_pix_error=0.25, inv_degree=None,
+                npoints=32, *args, **kwargs):
```
mcara, line 57: @jdavies-st Shouldn't a step's parameters be specified through the `spec` "docstring"-like class member of the step? If this is not going to be modified ever, why even bother with this and not hardcode all these parameters?

mcara: None of these parameters are documented. Shouldn't they be? Are these "private" parameters that users should not use? Then why have them? If they are supposed to be public - please add documentation.

perrygreenfield: I'm probably responsible for that. Mea culpa.

jwst/assign_wcs/assign_wcs_step.py
perrygreenfield: Unless the accuracy for the inverse is to be poor, I have a hard time thinking of a case where it would be lower.

mcara: I do not disagree with that.

But things would be way easier/simpler if you would allow `degree` to be a list in `gwcs`.

jwst/assign_wcs/assign_wcs_step.py
nden: @mcara is right, all input parameters need to go in the spec section to be picked up by `configobj`.
@cshanahan1 Follow the examples in this file and other *_step.py files i fnecessary and add all new parameters to `spec`. They also need to be added to `jwst/pipeline/assign_wcs.cfg`.

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -83,5 +86,48 @@ def process(self, input, *args, **kwargs):
...
+                    # fit sip approx., degree is chosen by best fit
+                    try:
+                        fit_sip_hdr = result.meta.wcs.to_fits_sip(None, max_pix_error,
+                                      degree, max_inv_pix_error, inv_degree,
+                                      npoints)
+                    except ValueError:
+                        return result
+
+                    # check if polynomial terms were actually fit, or just linear.
+                    sip = False or ('SIP' in fit_sip_hdr['ctype1'])
```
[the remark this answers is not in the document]
nden: There's a case where the fit does not return SIP coefficients, only the primary WCS , i.e. there's no distortion. We still want to save this header.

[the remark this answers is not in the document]
nden: It needs a comment to clarify this.

jwst/assign_wcs/assign_wcs_step.py
nden: ⬆️

nden (changes requested):

nden:
The failing tests are a real problem but it's in gwcs. When a WCS without a bounding box is attempted to be fit the algorithm fails because it cannot determine the sampling. This is not going to be a problem in the pipeline (most likely) because the WCS is attached to a data file and this means the bounding_box is always available.

jwst/assign_wcs/assign_wcs_step.py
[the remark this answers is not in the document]
mcara: Still, I do not see the need for `False`. It is absolutely unnecessary. Constant `False` is never `True` so if `X` is always `False`, `X or Y` is equivalent to just `Y`.

perrygreenfield:
But is SIP useful without WCS? I don't see the point of it in any other context.

jwst/assign_wcs/tests/test_wcs.py
```diff
@@ -155,3 +158,108 @@ def test_create_fitswcs(tmpdir, create_model_3d):
...
+    'bp_1_1': -3.60126568682908e-06,
+    'bp_1_2': -4.4691036548847474e-11,
+    'bp_2_0': -4.9044713849991446e-06,
+    'bp_2_1': -1.7110976247273783e-09,
+    'bp_3_0': 3.895381500452847e-11,
+    'ap_order': 3,
+    'bp_order': 3}
+
+    data_path = os.path.split(os.path.abspath(data.__file__))[0]
+    test_file = os.path.join(data_path, 'jw42424001001_01101_00001_nrca5_rate.fits')
```
jdavies-st, line 215: A slightly more reliable pattern would be

```
test_file = os.path.join(
    os.path.dirname(__file__), "data", "jw42424001001_01101_00001_nrca5_rate.fits"
)
```

and you can remove the `from . import data` above as well.

But I agree that generating a file on-the-fly via a pytest fixture is probably a better solution that committing a FITS file to the repo.

jwst/assign_wcs/tests/test_wcs.py
```diff
@@ -155,3 +158,108 @@ def test_create_fitswcs(tmpdir, create_model_3d):
...
+
+    pipe = AssignWcsStep()
+    result = pipe.process(test_file)
```
jdavies-st, line 218: Better to not call `process()` directly.  Instead:

```
result = AssignWcsStep.call(test_file)
```

as this will do all the step setup that is required, such as grabbing reffiles.

jwst/assign_wcs/tests/test_wcs.py
```diff
@@ -155,3 +158,108 @@ def test_create_fitswcs(tmpdir, create_model_3d):
...
+    wcs_info = result.meta.wcsinfo.instance
+
+    # make sure all expected keys are there
+    assert set(true_wcs.keys()).issubset(set((wcs_info.keys())))
+
+    # and make sure they match
+    for key in true_wcs:
+        try: # make sure numerical values match
+            k1 = float(true_wcs[key])
+            k2 = float(wcs_info[key])
```
jdavies-st, line 231: You shouldn't have to cast these to float.  They should already be floats I think?

jwst/assign_wcs/tests/test_wcs.py
```diff
@@ -155,3 +158,108 @@ def test_create_fitswcs(tmpdir, create_model_3d):
...
+            assert_allclose(k1, k2)
+        except ValueError: # and str values match
+            assert true_wcs[key] == wcs_info[key]
```
jdavies-st, line 234: Generally, a try/except is a big no-no in tests, because it can mask unexpected behavior.  In this case, it might be better to pick out `ctype1` and `ctype2` explicitly and compare them.  Makes it much clearer what the test is doing, and more strict.

Same above with the couple integer values that need to be compared.

jwst/assign_wcs/tests/test_wcs.py
```diff
@@ -155,3 +158,108 @@ def test_create_fitswcs(tmpdir, create_model_3d):
...
+
+    # evaulate fits wcs and gwcs, make sure they agree
+    x = arange(10)
+    y = arange(10)
```
jdavies-st, line 238: It would probably be better to evaluate over a grid over the entire image.  `gwcs` has a good tool for this:

```python

grid = grid_from_bounding_box(result.meta.wcs.bounding_box)
gwcs_ra, gwcs_dec = result.meta.wcs(*grid)
fitswcs_result = fits_wcs.pixel_to_world(*grid)
```

`grid_from_bounding_box` also has a `step` arg that allows one to not use every pixel, but a sparse grid.  So `step=10` would make a grid of every 10th pixel over the full range of the bounding box.

jwst/assign_wcs/tests/test_wcs.py
```diff
@@ -155,3 +158,108 @@ def test_create_fitswcs(tmpdir, create_model_3d):
...
+    gwcs_ra, gwcs_dec = result.meta.wcs(x, y)
+    fits_wcs = wcs.WCS(wcs_info)
+    fitswcs_res = fits_wcs.pixel_to_world(x, y)
+
+    assert_allclose(fitswcs_res.ra.deg, gwcs_ra)
+    assert_allclose(fitswcs_res.dec.deg, gwcs_dec)
+            
+    #now write the file out, read it back in, and check that the fit values are preserved
+    result.write('tmp.fits')
+    result_read = ImageModel('tmp.fits')
```
jdavies-st, line 248: Use your pytest `tmpdir` fixture for this:

```python
path = str(tmpdir.join("tmp_sip_wcs.fits"))
result.write(path)
```
and use the `with` context manager to open:

```python
with datamodels.open(path) as result_read:
    assert result.meta.wcsinfo == result_read.meta.wcsinfo
```

jwst/assign_wcs/tests/test_wcs.py
```diff
@@ -155,3 +158,108 @@ def test_create_fitswcs(tmpdir, create_model_3d):
...
+    # make sure all expected keys are there
+    assert set(true_wcs.keys()).issubset(set((wcs_info_read.keys())))
+
+    for key in true_wcs:
+        try: # make sure numerical values match
+            k1 = float(true_wcs[key])
+            k2 = float(wcs_info_read[key])
+            assert_allclose(k1, k2)
+        except ValueError: # and str values match
+            assert true_wcs[key] == wcs_info_read[key]
```
jdavies-st, line 261: You're basically checking round-tripping here.  The check above is more strict and will verify that the thing you had before writing is the thing you now have on read.

jdavies-st (commented):
Looks good!  Some comments below.

Fixes needed are in the passing of step config arguments via `spec` and replacing the FITS file with one created on-the-fly.  You can see how this is done here:

https://github.com/spacetelescope/jwst/blob/785d976a54b3de8e125b39f8a5f44f1bf76bfa6b/jwst/wfs_combine/tests/test_wfs_combine.py#L14-L60

which creates 2 files and association.  In your case, one file, and just return the file in memory.  No need to write it out via the fixture.

jwst/assign_wcs/tests/test_wcs.py
jdavies-st: Ah, I see that you are casting all the keys and then when strings can't be cast, trapping the exception.  Better would to be explicit in the test.  No `try/except` blocks.

jwst/assign_wcs/tests/test_wcs.py
jdavies-st: Another thing you can do if you want to make sure that you've done the new schema keywords correctly and do not get any validation warnings on read or write is to do something like:

```python
from stdatamodels.validate import ValidationWarning

with pytest.warns(None) as record:
    result.write(path)
    assert len(record) == 0

with pytest.warns(None) as record:
    with datamodels.open(path) as result_read:
        assert len(record) == 0
```

jdavies-st:
For the failing unit test

https://travis-ci.com/github/spacetelescope/jwst/jobs/456720593#L768

Feel free to do

```python
AssignWcsStep.call(input_file, sip_approx=False)
```

jwst/assign_wcs/tests/test_wcs.py
cshanahan1: I changed this and now assert_allclose fails with the default tolerance on comparing Dec. Still, the maximum difference is still only ~1.4e-06 degrees so I'll just increase the tolerance and make a note.

mcara:
@perrygreenfield What do you refer to in 

> But is SIP useful without WCS? I don't see the point of it in any other context.

?

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -83,5 +90,44 @@ def process(self, input, *args, **kwargs):
             slit_y_range = [self.slit_y_low, self.slit_y_high]
             result = load_wcs(input_model, reference_file_names, slit_y_range)
 
+            if self.sip_approx:
+                if self.sip_degree is not None and self.sip_degree > 6:
```
mcara, line 94:
```suggestion
                if self.sip_degree and self.sip_degree > 6:
```

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -83,5 +90,44 @@ def process(self, input, *args, **kwargs):
...
+                    raise ValueError(f"SIP polynomial degree of {self.sip_degree} must be <=6")
+                if self.sip_inv_degree is not None and self.sip_inv_degree > 6:
```
mcara, line 96:
```suggestion
                if self.sip_inv_degree and self.sip_inv_degree > 6:
```

jwst/assign_wcs/assign_wcs_step.py
cshanahan1: i'm going to leave this unresolved, it seems like this could be an improvement later?

jwst/assign_wcs/assign_wcs_step.py
cshanahan1: @mcara where should i add documentation, unless you were referring to adding documentation in gwcs? I wrote short descriptions of the parameters in ``spec``, let me know if this is sufficient!

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -83,5 +90,44 @@ def process(self, input, *args, **kwargs):
...
+                                      self.sip_degree, self.sip_max_inv_pix_error, self.sip_inv_degree,
+                                      self.sip_npoints)
+                    except ValueError:
+                        return result
+
+
+                    # if a polynomia was fit (not just linear terms), but the
+                    # best fit > 6th degree, fit a 6th degree instead.
+                    # and give up if that doesn't work.
+                    if 'SIP' in fit_sip_hdr['ctype1'] and fit_sip_hdr['a_order'] > 6:
```
mcara, line 113: I think this still does not cover all situations. For example, what if forward polynomial degree is 6 (OK) but inverse is 7? This logic would not detect this situation.

I would do it this way:
```
                    if 'SIP' in fit_sip_hdr['ctype1'] and (fit_sip_hdr['a_order'] > 6 or fit_sip_hdr['ap_order'] > 6):
                        fwd_deg = min(fit_sip_hdr['a_order'], 6)
                        inv_deg = min(fit_sip_hdr['ap_order'], 6)
                        result.meta.wcs.to_fits_sip(None, self.sip_max_pix_error,
                                            fwd_deg, self.sip_max_inv_pix_error, inv_deg, self.sip_npoints)
```

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -83,5 +86,48 @@ def process(self, input, *args, **kwargs):
             slit_y_range = [self.slit_y_low, self.slit_y_high]
             result = load_wcs(input_model, reference_file_names, slit_y_range)
 
+            if sip_approx:
+                if degree is not None:
+                    if degree > 6:
+                        raise ValueError('Degree for SIP approximation must be less or equal to 6.')
```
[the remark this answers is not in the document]
cshanahan1: ok, im going to get rid of the check/error im raising then and just leave it up to that

jwst/assign_wcs/tests/test_wcs.py
jdavies-st: Perhaps an easier validation check to make sure the new keywords being added comply with the schema is:

```
with pytest.warns(None) as record:
    result.validate()
    assert len(record) == 0
```

which should do a full validation of model, including the newly-populated `meta.wcsinfo` section.

jwst/pipeline/assign_wcs.cfg
```diff
@@ -1,4 +1,10 @@
 name = "assign_wcs"
 class = "jwst.assign_wcs.AssignWcsStep"
+sip_approx = True  # enables SIP approximation for imaging modes.
+sip_max_pix_error = 0.25  # imaging modes only
+sip_degree = None  # imaging modes only
+sip_max_inv_pix_error = 0.25  # imaging modes only
+sip_inv_degree = None  # imaging modes only
+sip_npoints = 32  # imagine modes only
```
jdavies-st, line 8:
```suggestion
sip_npoints = 32  # imaging modes only
```

jwst/assign_wcs/assign_wcs_step.py
mcara: @cshanahan1 Take a look at https://github.com/spacetelescope/jwst/pull/5507#pullrequestreview-549584346 That fixes, I believe, the first part + another issue described in the comment - you do not detect the situation when inverse degree is over 6 but forward is `<=6`.

The other issue with simplification of the logic - that one requires changes in `gwcs` and it cannot be done now.

jwst/assign_wcs/assign_wcs_step.py
mcara: @nden would know better but my guess would be for something like https://github.com/spacetelescope/jwst/blob/master/docs/jwst/assign_wcs/main.rst. Take a look at https://github.com/spacetelescope/jwst/blob/master/docs/jwst/tweakreg/README.rst

nden: Yes, add information about the SIP approximation to the `Description` section of assign_wcs.

nden (commented):
This looks like it's almost ready to go. I have two additional minor comments.

perrygreenfield:
@mcara (previously responded to in slack but putting it here for historical purposes) regarding my comment about bounding boxes, the only use of SIP is for WCS where a bounding box should be required. It has no real use elsewhere in modeling, so that it fails without a bounding box should not be a problem (should a bounding box be required by this model if it isn't already, perhaps, though I'm not sure how that interacts with what GWCS does with bounding boxes)

jwst/pipeline/assign_wcs.cfg
cshanahan1: haha! whoops, i must have been imagining things.

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -83,5 +90,43 @@ def process(self, input, *args, **kwargs):
             slit_y_range = [self.slit_y_low, self.slit_y_high]
             result = load_wcs(input_model, reference_file_names, slit_y_range)
 
+            if self.sip_approx:
+                # check if its imaging mode
+                if result.meta.exposure.type.lower() in IMAGING_TYPES:
+
+                    # fit sip approx., degree is chosen by best fit
+                    try:
+                        print('self.sip_degree', self.sip_degree)
```
mcara, line 99: Did you forget this debugging `print`?
```suggestion
```

jwst/pipeline/assign_wcs.cfg
```diff
@@ -1,4 +1,10 @@
 name = "assign_wcs"
 class = "jwst.assign_wcs.AssignWcsStep"
+sip_approx = True  # enables SIP approximation for imaging modes.
+sip_max_pix_error = 0.25  # imaging modes only
+sip_degree = None  # imaging modes only
```
jdavies-st, line 5: `None` is not an allowed value in `configobj` for a value that needs an integer.  I would recommend removing this and `sip_inv_degree` below.  And in fact, remove all the new ones here, as they are taken care of in the step's `"spec"` attribute.

This is the disadvantage of using `configobj` to do our validation and is a known issue.

Btw, this is what is causing the suffix unit test failure.

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -83,5 +86,48 @@ def process(self, input, *args, **kwargs):
...
+
+                    # maintain convention of lowercase keys
+                    fit_sip_hdr = {k.lower(): v for k, v in fit_sip_hdr.items()}
+
+                    # update meta.wcs_info with fit keywords
+                    result.meta.wcsinfo.instance.update(fit_sip_hdr)
+
+                    # delete naxis, cdelt, pc from wcsinfo
+                    rm_keys = ['naxis1', 'naxis2', 'cdelt1', 'cdelt2', 'pc1_1',
+                               'pc1_2', 'pc2_1', 'pc2_2']
```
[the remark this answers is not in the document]
cshanahan1: I think this is the only suggestion I haven't addressed. so the point of this is just for clarity? I'll add this before the header update, then after go ahead and just delete PC/cdelt?

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -83,5 +90,45 @@ def process(self, input, *args, **kwargs):
             slit_y_range = [self.slit_y_low, self.slit_y_high]
             result = load_wcs(input_model, reference_file_names, slit_y_range)
 
+            if self.sip_approx:
+                # check if its imaging mode
+                if result.meta.exposure.type.lower() in IMAGING_TYPES:
```
mcara, line 95: From what I see, these `if`s can also be joined with an `and`:
```suggestion
            if result.meta.exposure.type.lower() in IMAGING_TYPES and self.sip_approx:
```
and then decrease indentation of the rest of the block.

mcara: Even better:
```python
if not (result.meta.exposure.type.lower() in IMAGING_TYPES and self.sip_approx):
    return result
```
and then de-dent by 8 spaces!

jwst/assign_wcs/assign_wcs_step.py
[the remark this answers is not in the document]
jdavies-st: Yeah, just for clarity in terms of reading the code and figuring out what is going on.  You could also just add a comment indicating why `naxis*` is being removed.  Up to you.  But yeah, just clarity for others (and your future self).  =)

mcara:
What is the reason `naxis` is deleted in https://github.com/spacetelescope/jwst/pull/5507/files#diff-c1ecc261f156c297269fb06a9718273df2f77ac75ba422ac0a0699c22579658cR127 ?

CC: @nden @jdavies-st

jwst/pipeline/assign_wcs.cfg
cshanahan1: Ok, I removed all of these parameters from the config file.

jwst/extract_2d/tests/test_grisms.py
```diff
@@ -148,7 +148,7 @@ def create_tso_wcsimage(filtername="F277W", subarray=False):
     im.meta.wcsinfo.siaf_xref_sci = 887.0
     im.meta.wcsinfo.siaf_yref_sci = 35.0
     aswcs = AssignWcsStep()
-    return aswcs.process(im)
+    return aswcs.process(im, sip_approx=False)
```
nden, line 151: Hm, I'm surprised you have to disable sip approximation in this tests. The test is for NRC_WFSS (grism) observations and normally should not go through sip approximation.

jdavies-st:
@mcara, the `to_fits_sip()` method in `gwcs.WCS` returns a FITS header that has `naxis` in it, but it is not needed to be assign because that is handled by the `astropy.io.fits` library.  I.e. it's not a keyword header that we manage in `model.meta.wcsinfo` for any particular FITS extensions that get written.

jwst/master_background/tests/test_master_background.py
```diff
@@ -45,7 +45,7 @@ def science_image():
     image.meta.wcsinfo.roll_ref = 0
     image.meta.wcsinfo.ra_ref = 0
     image.meta.wcsinfo.dec_ref = 0
-    image = AssignWcsStep.call(image)
+    image = AssignWcsStep.call(image, sip_approx=False)
```
nden, line 48: The above comment applies here as well. This is an LRS (i.e. spectral observation).

mcara:
@jdavies-st Thanks! @nden already explained this: I missed that `wcsinfo` is just a dictionary with essential stuff and not a full header.

nden (commented):
I see some tests had to exclude spectral observation from sip approximation which shouldn't happen if `IMAGING_TYPES` includes only imaging observations. Need to double check what `IMAGING_TYPES` is.

jwst/extract_2d/tests/test_grisms.py
cshanahan1: oh whoops, i was just blindly going through and switching it off every time AssignWcsStep was called in the tests. ill fix that.

mcara (approved):
Looks good to me. Just squash all the commits.

jdavies-st:
Btw, this block of new code that is not covered by a test:

Screen Shot 2020-12-15 at 11 43 44 AM

Might be worth adding a test to cover it, to make sure that it's working as expected.  I.e force a fitting with 7th order, and see if it triggers this.  Or is this block of code unreachable by a test in that one can't actually request a 7th order fit?  And one would only get that if the `fit_sip_hdr` returned something >6?  Or does it not do that and this whole block is mute?

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -83,5 +90,44 @@ def process(self, input, *args, **kwargs):
             slit_y_range = [self.slit_y_low, self.slit_y_high]
             result = load_wcs(input_model, reference_file_names, slit_y_range)
 
+        if not (result.meta.exposure.type.lower() in IMAGING_TYPES and self.sip_approx):
```
nden, line 93: @cshanahan1 I think this is the wrong condition that makes sip approximation run for spectral types. We need this to evaluate to True in order to skip SIP processing. As it is written it will evaluate to False for spectral observations when `self.sip_approx=True`. If it's written as

` if result.meta.exposure.type.lower() not in IMAGING_TYPES or not self.sip_approx:`

mcara: > @cshanahan1 I think this is the wrong condition that makes sip approximation run for spectral types. We need this to evaluate to True in order to skip SIP processing. As it is written it will evaluate to False for spectral observations when `self.sip_approx=True`. If it's written as
> 
> ` if result.meta.exposure.type.lower() not in IMAGING_TYPES or not self.sip_approx:`

I disagree. If image type's is not in `IMAGING_TYPES` then  `result.meta.exposure.type.lower() in IMAGING_TYPES` is `False` and combined with the `not` in front => `True` => `return result` without fitting SIP.

mcara: Am I missing something?

nden: I agree :-) with your disgreement

cshanahan1:
> Btw, this block of new code that is not covered by a test:
> 
> Screen Shot 2020-12-15 at 11 43 44 AM
> 
> Might be worth adding a test to cover it, to make sure that it's working as expected. I.e force a fitting with 7th order, and see if it triggers this. Or is this block of code unreachable by a test in that one can't actually request a 7th order fit? And one would only get that if the `fit_sip_hdr` returned something >6? Or does it not do that and this whole block is mute?
> 
> This is why the codedov/patch check below is failing.
> 
> That said, if it's a real "failure" mode of the fitting method, and we can't get to it via calling the step with allowed ranges of params (<=6), then maybe skip testing it.

I can't force > 6 order polynomial, the only way i could make this block run is if i made an image with a wcs that is artificially created including a 7th order polynomial, then allow it to choose the best fit which should be 7th, in which case this block would run. do you think this is important to add @jdavies-st ?

nden (approved):

jdavies-st:
@cshanahan1, no not too important.  Skip it.  👍 

Just add an entry to CHANGES.rst for this PR, following the examples already there under an `assign_wcs` section.

jdavies-st (approved):

jdavies-st:
Btw, when running our basic NIRCam imaging regression test using this branch, I see the following `fitsdiff` differences:

```
E              Keyword CRPIX1   has different values:
E                 a> 1025.0
E                 b> 1024.5
E              Keyword CRPIX2   has different values:
E                 a> 1025.0
E                 b> 1024.5
```

amongst the expected ones.  The current CRPIX has the half-pixel value, from the distortion solution.  I suspect that this is because the `fit_sip_hdr` method picks the middle of the image to do a SIP polynomial fit, but does so with an integer-valued pixel?

There shouldn't be any knock-on effects of changing CRPIXn at all, will there be @nden?  It shouldn't matter that CRPIXn no longer corresponds to what is in the `model.meta.wcs` GWCS object?

Btw, `CRPIXn` is not currently used at all by `resample` or `outlier_detection`.  It used to be.

hbushouse:
As long as CRVALn are still consistent with CRPIXn, I don't see any downside to the half pixel shift in CRPIXn.

jdavies-st:
Btw, running `jwst/regtest/test_miri_image.py` with this branch produces some nice SIP coeff `_cal.fits` images that now load up in ds9 and align really nicely.  The `test_nircam_image.py` regression test `_cal.fits` files mentioned above also look equally great.  🚀

mcara:
> It shouldn't matter that CRPIXn no longer corresponds to what is in the `model.meta.wcs` GWCS object?

_In principle_, it moves the center of projection. This can have effect on true "projection effects" (that is, effects of projecting a sphere onto a plane). However, 1/2 pixel for an ACS-like image would have such a small effect due to this projection, that we can really neglect it. However, it is 1/2 pixel for this test image but it could be more for another image.

Preferably, `to_fits_sip()` should be able to take optional `crpix` parameter which, when provided, would replace https://github.com/spacetelescope/gwcs/blob/45f65154ac57b8714aa46e2cf12b415bbcc3cd45/gwcs/wcs.py#L1435-L1436

CC: @perrygreenfield 

Moreover, this is really a bug in the above code in `to_fits_sip` since it should add `+ 1` to get the correct number of `naxis1,2`. CC: @perrygreenfield

mcara:
> It shouldn't matter that CRPIXn no longer corresponds to what is in the `model.meta.wcs` GWCS object?

One more thing... Like it or not, `tweakreg` step modifies the "real" (`gwcs`) WCS but it does not update `wcsinfo` and so things get out of sync anyway.

jdavies-st:
An unforeseen issue I see is that for the resampled data produced in the `calwbb_image3`, the meta `update` pulls in all the SIP stuff from input `_cal.fits` images now, but that is not wanted or needed in the drizzled output.  These means that the `_i2d.fits` files now have both PC and CD matrix, the wrong CTYPEn and all the SIP coeffs.

So that's a problem that needs to be fixed.  Here or a quick follow-on PR.

Running this test shows the problem:

```
$ pytest --bigdata -v --basetemp=/tmp/scratch jwst/regtest -k test_miri_image3_i2d
```

jdavies-st:
I suspect that fixing the resampled data could happen here in this function:

https://github.com/spacetelescope/jwst/blob/5a8757199e3cdf01c75d2be2ed27cee4b6e2e1f2/jwst/resample/resample.py#L180-L196

and it would involve deleting any SIP coeffs-related attributes in `wcsinfo`.

mcara:
> I suspect that fixing the resampled data could happen here in this function:
> ......................................
> and it would involve deleting any SIP coeffs-related attributes in `wcsinfo`.

In a resampled image, WCS is not (or should not be) simply the input's image's WCS. For example, if by resampling you mean HST-like drizzle, computed WCS for the drizzled image is "rectified" and no skews are present in the CD matrix. So, in a drizzled image's WCS almost everything is different from input image's WCS. Removing SIP is not enough. But maybe I mis-understand the issue.

jdavies-st:
Right, the same is true here for JWST, but currently we don't clean out any SIP coeffs or CD-matrix related stuff.  And we do a `dict`-type update from the first input image to populated `meta` in the output image, and that includes `wcsinfo`.  We then overwrite any PC-matrix related keywords.  It was not foreseen that there would be a CD matrix and SIP coeffs there as well.

Maybe there's a better, more programmatic way to do this.  Would be nice if `gwcs` could control what should and should not appear in any SIP or PC/CDELT WCS it generates.

Anyway, I have a fix in a separate PR.  So I think we can merge this and fix the resample issue separately.

mcara:
We had only 18 more comments to go to get to 100

hbushouse:
I tend to agree with @mcara on the issue regarding the WCS for the resampled image. It's a whole new output frame that bears little, if any, resemblance to the old one(s) from which it was formed and hence the resampled image should essentially start with a blank slate as far as the WCS object (and FITS WCS keywords in meta.wcsinfo) are concerned and create a new WCS from scratch, with no inheriting from the old and hence no need to delete items one-by-one from the inherited set.

hbushouse:
T-16

hbushouse:
A quick browse of last night's regression test run shows the expected dozens of cases of output images that now have all the extra SIP keywords in their SCI extension headers. I also noticed that the CDELTn and PCn_m keywords are gone, having been replaced by CDn_m keywords. Do we have any idea whether the removal of CDELT and PC is going to cause problems for anyone? I assume that nothing in the Cal pipeline code itself was using those (should be using everything from the gWCS object), but there could be some fallout with user code here and there.

docs/jwst/assign_wcs/main.rst
```diff
@@ -22,6 +22,12 @@ input frame ``detector`` to a frame ``v2v3``. This part of the WCS pipeline may
 intermediate coordinate frames. The basic WCS keywords are used to create
 the transform from frame ``v2v3`` to frame ``world``.
 
+For image display with software like DS9 that relies on specific WCS information, a SIP-based
+approximation to the WCS is fit. The results are FITS keywords stored in
+``model.meta.wcsinfo``. This is not an exact fit, but is accurate to ~0.25 pixel and is sufficient
+for display purposes. This step, which occurs for imaging modes early, is performed by default but
```
hbushouse, line 28: Do you really mean "early" here, or "only"?

cshanahan1: I meant only, i'll fix that.

mcara:
@hbushouse 
>  Do we have any idea whether the removal of CDELT and PC is going to cause problems for anyone?

With regard to "any idea" - strictly speaking no, we don't. I can _imagine_ a situation when a user (internal or external) has code that accesses `pc1_1` or `cdelt1` members but I doubt this would happen in practice. However, there is no practical solution to this ... I think: it is either CD or PC formalism and only CD works with SIP and so we do have to break things (in case someone is accessing these values directly) in order to get better approximation. Keeping both keywords in the header is not a good approach: it is hard to keep in sync (CD matrix is PC * linear_correction) and, as you saw, currently `to_fits_sip` modifies CRPIX and so PC/CDELT is just no longer correct. I think it has to be discarded.

perrygreenfield:
T-15

jdavies-st:
A quick update here, since this seems where all the discussion is happening.

The knock-on effects of populating a CD matrix in the `_cal.fits` files has been handled for `calwebb_image3` and `resample`.  I think the extra thing to consider is that when the image3 pipeline runs `tweakreg`, it updates the gwcs object and then that updated one gets written out later in the `_crf.fits` file.  So I think it would be good to updated the SIP fit in `tweakreg` step as well.

mcara:
In https://github.com/spacetelescope/jwst/pull/5507#issuecomment-745491495 I wrote:

> Moreover, this is really a bug in the above code in `to_fits_sip` since it should add `+ 1` to get the correct number of `naxis1,2`.

@jdavies-st pointed that the computed difference value is correct but it gets rounded to an integer. Indeed, bounding box limits usually are 1/2 "wider" to include the edge of the pixels and hence the difference between the upper and lower limits already includes this `+1` pixel that I was talking about. => Not a bug!

jdavies-st:
So @hbushouse  OKified the regression tests this morning before we all woke up.  All looked as expected except for some NIRCam time-series imaging with a 64x64 subarray.  It failed outright when trying to do the fit:

https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/977/testReport/junit/jwst.regtest/test_nircam_tsimg/_stable_deps__test_nircam_tsimg_stage2_calints_/

The traceback:

```pytb
/data1/jenkins/workspace/RT/JWST/clone/jwst/assign_wcs/assign_wcs_step.py:98: in process
    fit_sip_hdr = result.meta.wcs.to_fits_sip(None, self.sip_max_pix_error,
/data1/jenkins/workspace/RT/JWST/miniconda/envs/tmp_env0/lib/python3.8/site-packages/gwcs/wcs.py:1461: in to_fits_sip
    ud, vd = _make_sampling_grid(2 * npoints, bounding_box)
/data1/jenkins/workspace/RT/JWST/miniconda/envs/tmp_env0/lib/python3.8/site-packages/gwcs/wcs.py:1769: in _make_sampling_grid
    y, x = np.mgrid[: ymax + 1: stepsize_y, : xmax + 1: stepsize_x]

self = <numpy.lib.index_tricks.MGridClass object at 0x7feb64007310>
key = (slice(None, 63.5, 0), slice(None, 63.5, 0))

    def __getitem__(self, key):
        try:
            size = []
            typ = int
            for k in range(len(key)):
                step = key[k].step
                start = key[k].start
                if start is None:
                    start = 0
                if step is None:
                    step = 1
                if isinstance(step, complex):
                    size.append(int(abs(step)))
                    typ = float
                else:
                    size.append(
>                       int(math.ceil((key[k].stop - start)/(step*1.0))))
E                       OverflowError: cannot convert float infinity to integer

/data1/jenkins/workspace/RT/JWST/miniconda/envs/tmp_env0/lib/python3.8/site-packages/numpy/lib/index_tricks.py:162: OverflowError
```

It looks like `_make_sampling_grid()` is setting the slice step size to zero.  A more verbose traceback:

```pytb
~/miniconda3/envs/jwst/lib/python3.9/site-packages/gwcs/wcs.py in to_fits_sip(self=<WCS(output_frame=world, input_frame=detector, f....0002777777777777778 -0.0420034575 .. -84.67665)>, bounding_box=((-0.5, 62.5), (-0.5, 62.5)), max_pix_error=0.25, degree=None, max_inv_pix_error=0.25, inv_degree=None, npoints=32, verbose=False)
   1459         undist_x, undist_y = ntransform(u, v)
   1460         # Double sampling to check if sampling is sufficient.
-> 1461         ud, vd = _make_sampling_grid(2 * npoints, bounding_box)
        ud = undefined
        vd = undefined
        global _make_sampling_grid = <function _make_sampling_grid at 0x7fbde031fe50>
        npoints = 32
        bounding_box = ((-0.5, 62.5), (-0.5, 62.5))
   1462         undist_xd, undist_yd = ntransform(ud, vd)
   1463         # Determine approximate pixel scale in order to compute error threshold

~/miniconda3/envs/jwst/lib/python3.9/site-packages/gwcs/wcs.py in _make_sampling_grid(npoints=64, bounding_box=((-0.5, 62.5), (-0.5, 62.5)))
   1767     stepsize_y = int(ysize / npoints)
   1768     # Ensure last row and column are part of the evaluation grid.
-> 1769     y, x = np.mgrid[: ymax + 1: stepsize_y, : xmax + 1: stepsize_x]
        y = undefined
        x = undefined
        global np.mgrid = <numpy.lib.index_tricks.MGridClass object at 0x7fbdb1f12790>
        ymax = 62.5
        stepsize_y = 0
        xmax = 62.5
        stepsize_x = 0
   1770     x[:, -1] = xsize - 1
   1771     y[-1, :] = ysize - 1

~/miniconda3/envs/jwst/lib/python3.9/site-packages/numpy/lib/index_tricks.py in __getitem__(self=<numpy.lib.index_tricks.MGridClass object>, key=(slice(None, 63.5, 0), slice(None, 63.5, 0)))
    160                 else:
    161                     size.append(
--> 162                         int(math.ceil((key[k].stop - start)/(step*1.0))))
        global int = undefined
        global math.ceil = <built-in function ceil>
        key = (slice(None, 63.5, 0), slice(None, 63.5, 0))
        k.stop = undefined
        start = 0
        step = 0
    163                 if (isinstance(step, float) or
    164                         isinstance(start, float) or

OverflowError: cannot convert float infinity to integer
```

@perrygreenfield any ideas what might be happening?  I can provide the dataset if helpful.

jdavies-st:
It looks like the following

https://github.com/spacetelescope/gwcs/blob/45f65154ac57b8714aa46e2cf12b415bbcc3cd45/gwcs/wcs.py#L1766-L1767

can return 0 when `xsize < npoints`.  This will be the case for all subarrays 64 pixels in size or smaller.

jdavies-st:
I think our choice is to do one of the following:

- [ ] Don't fit SIP coeffs to any subarrays 64 pix or smaller.  So modify `assign_wcs`.
- [ ] But for this particular dataset, the subarray is exactly 64x64, but the bounding box is computed by `bounding_box_from_subarray()` to be `((-0.5, 62.5), (-0.5, 62.5))`.  Is that correct?  Shouldn't it be `((-0.5, 63.5), (-0.5, 63.5))`?  If that's a bug in `bounding_box_from_subarray()` then, that would fix the current test, but not the general case.
- [ ] Fix the problem in `gwcs` so it handles images smaller than 64x64 and release a patch version of `gwcs` to be used by `jwst`.

Thoughts?  @cshanahan1 @nden @perrygreenfield @hbushouse

jdavies-st:
It seems `bounding_box_from_subarray()` actually makes the bounding box 1 pixel too small in each dimension, and there's a unit test testing for the wrong size as well:

https://github.com/spacetelescope/jwst/blob/c129fee33dbdd7c99a427e4425b81b940ab49dd0/jwst/assign_wcs/tests/test_util.py#L98-L104

For non-subarrays, the bounding box for full frame NIRCam images looks like this, relative to the array size:

```
In [56]: model = datamodels.open("jw42424001001_01101_00001_nrca5_cal.fits")

In [57]: model.meta.wcs.bounding_box
Out[57]: ((-0.5, 2047.5), (-0.5, 2047.5))

In [58]: model.data.shape
Out[58]: (2048, 2048)
```

I'll write an issue and fix.

mcara:
I think the code in `to_fits_sip()` should be made more robust to this kind of issues and raise a meaningful exception.

mcara:
I think I will take back my comment in https://github.com/spacetelescope/jwst/pull/5507#issuecomment-746532214; the lines:

https://github.com/spacetelescope/gwcs/blob/45f65154ac57b8714aa46e2cf12b415bbcc3cd45/gwcs/wcs.py#L1435-L1436

seem to be problematic if `xmin>>1`: then `crpix` could be outside the image. I think the correct formula should have `+` instead of `-`. CC: @perrygreenfield

PR #5509, opened 2020-12-04
Reset logging configuration after test

eslavich:
The test in jwst/stpipe/tests/test_logger.py changes the stpipe logging configuration but doesn't set it back, which I think is responsible for some of the issues we've been seeing with pytest-openfiles checks.  This PR adds a fixture that restores the default logging configuration after that test.


jdavies-st (approved):
THANK YOU!

PR #5510, opened 2020-12-04
Require stdatamodels from pypi

eslavich:
Hot off the presses!

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
     requests>=2.22
     scipy>=1.1.0
     spherical-geometry>=1.2.18
-    stdatamodels @ git+https://github.com/spacetelescope/stdatamodels
+    stdatamodels~=0.1.0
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -36,7 +36,7 @@ install_requires =
     requests>=2.22
     scipy>=1.1.0
     spherical-geometry>=1.2.18
-    stdatamodels @ git+https://github.com/spacetelescope/stdatamodels
+    stdatamodels~=0.1.0
```
jdavies-st, line 39: Do you want to pin it to `0.1.x`?  Or allow to float to `0.2.x` when released?

eslavich: Yea, it should be pinned to 0.1.x, since 0.2.0 is going to include that astropy Time change that we'll need to update jwst to support.

jdavies-st (approved):

PR #5512, opened 2020-12-04
Add automatic meta.date management to JwstDataModel

eslavich:
To support Roman storing astropy.time.Time objects in meta.date, the code that automatically maintains that attribute will be removed from stdatamodels starting in version 0.2.0 (not yet released).  This PR re-implements that feature in JwstDataModel using a new hook that we added to the base DataModel class.

This PR should *not* be merged before jwst 0.18.0 is released, since stdatamodels 0.2.0 will include other changes that may cause trouble for jwst.  I'm mainly opening the PR now so I don't forget about it.

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -73,6 +73,7 @@ bc0 = new BuildConfig()
 bc0.nodetype = 'jwst'
 bc0.name = 'stable-deps'
 bc0.env_vars = env_vars
+bc0.conda_ver = '4.9.2'
 bc0.conda_packages = [
     "python=${python_version}",
 ]
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -73,7 +73,6 @@ bc0 = new BuildConfig()
 bc0.nodetype = 'jwst'
 bc0.name = 'stable-deps'
 bc0.env_vars = env_vars
-bc0.conda_ver = '4.8.4'
 bc0.conda_packages = [
     "python=${python_version}",
 ]
--- a/JenkinsfileRT_dev
+++ b/JenkinsfileRT_dev
@@ -74,6 +74,7 @@ bc0 = new BuildConfig()
 bc0.nodetype = 'jwst'
 bc0.name = 'unstable-deps'
 bc0.env_vars = env_vars
+bc0.conda_ver = '4.9.2'
 bc0.conda_packages = [
     "python=${python_version}",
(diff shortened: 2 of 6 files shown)


jdavies-st (commented):
Will this change allow us to unpin `stdatamodels~=0.1.0`?  If so, we should do it here in this PR.  Though maybe there will be other stuff in 0.2.0 that is unaddressed here?

jdavies-st:
Just a note here that the changes over at `stdatamodels` to allow for an astropy time object in `meta.date` is causing our `jwst` nightly regression tests with dev dependencies (including dev `stdatamodels`) to fail, but that is expected.

https://plwishmaster.stsci.edu:8081/job/RT/job/JWST_UNSTABLE/427/

Those will fail until this PR is merged.

jdavies-st (commented):
It's probably OK to now get this PR to a state to be merged, now that 0.18.x is done.

The minimum will be unpinning `stdatamodels` in setup.cfg

https://github.com/spacetelescope/jwst/blob/3914b95672188f700b2b21a6057e5e52627a56c4/setup.cfg#L45

Might be good to run some regression tests as well after rebasing.

eslavich:
Regression test run on this branch with the stdatamodels release candidate: https://plwishmaster.stsci.edu:8081/job/RT/job/eslavich-dev/63/

JenkinsfileRT
```diff
@@ -73,7 +73,6 @@ bc0 = new BuildConfig()
 bc0.nodetype = 'jwst'
 bc0.name = 'stable-deps'
 bc0.env_vars = env_vars
-bc0.conda_ver = '4.8.4'
```
jdavies-st, line 76: Removing this means the default conda version will be used, which might be out-of-date.  We should probably just bump this to the latest for now until we fix the Jenkins CI helpers to get the latest, i.e.

https://github.com/spacetelescope/jenkins_shared_ci_utils/blob/37606ed9c9a9133e7983183e52976683d8d25ece/vars/utils.groovy#L106-L108

jdavies-st (approved):

JenkinsfileRT
eslavich: Oops, yes, I have a PR open over there (https://github.com/spacetelescope/jenkins_shared_ci_utils/pull/80) but I agree we should wait until that's actually merged.

Issue #5511, opened 2020-12-04
Standardize wfs_combine


stscijgbot-jp:
_Issue [JP-1793](https://jira.stsci.edu/browse/JP-1793) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

The `WfsCombineStep` has a number oddities that make it difficult to work with.  These should be resolved.

* It only accepts an assocation as input.  Generally steps also accept open datamodels, or in this case perhaps a ModelContainer would be appropriate.
* The step does not return anything.  It merely writes out the results.  These makes it very difficult to test.  It should return either a DataModel or ModelContainer as appropriate
* It always writes out the results.  But steps shouldn't do this.  They should only write out results if asked to.
* It doesn't know whether it is a Step or Pipeline.  It seems to be being used as a pipeline, and perhaps that's why it has hard-coded writing of files and no return at the end of the step.

jdavies-st:
Some good comments here

https://github.com/spacetelescope/jwst/pull/5519#discussion_r537820643

On returning the model with the correct name without having to write it out manually, as we currently do.

stscijgbot-jp:
_Comment by [Alicia Canipe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=acanipe) on [JIRA](https://jira.stsci.edu/browse/JP-1793?focusedCommentId=564213#comment-564213):_

Just checking in on this, since it isn't quite clear... is this fix implemented? If so, should we ask wavefront experts to test? 

[Marcio Melendez Hernandez](https://jira.stsci.edu/secure/ViewProfile.jspa?name=melendez) FYI

stscijgbot-jp:
_Comment by [Marcio Melendez Hernandez](https://jira.stsci.edu/secure/ViewProfile.jspa?name=melendez) on [JIRA](https://jira.stsci.edu/browse/JP-1793?focusedCommentId=564218#comment-564218):_

[Alicia Canipe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=acanipe) I can give it a try but it seems that it is more about the wfs_combine method that the actual science quality output, i.e., images. Not sure what should be the expectation about the output.

stscijgbot-jp:
[...]

PR #5519, opened 2020-12-07
JP-1796: Fix wfs_combine bugs; add unit tests

### LINKED ISSUE #5515: Resolve regression tests with wfs_combine
_Issue [JP-1796](https://jira.stsci.edu/browse/JP-1796) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

jdavies-st:
Fixes some bugs introduced in #5500 that showed up in the `wfs_combine` regression tests.  Adds some unit tests for the expected behavior.

Resolves #5515 / [JP-1796](https://jira.stsci.edu/browse/JP-1796)

### DIFF
--- a/jwst/wfs_combine/wfs_combine.py
+++ b/jwst/wfs_combine/wfs_combine.py
@@ -21,7 +21,7 @@ BLUR_SIZE = 10  # size of gaussian kernel for convolution
 N_SIZE = 2  # size of neighborhood in create_griddata_array
 
 
-class DataSet():
+class DataSet:
     """
     Two dithered input wavefront sensing images to be combined
 
--- a/jwst/wfs_combine/wfs_combine.py
+++ b/jwst/wfs_combine/wfs_combine.py
@@ -1,14 +1,12 @@
-#! /usr/bin/env python
-#
-#  wfs_combine.py - combine 2 dithered PSF images
-#
-import numpy as np
 import logging
-from .. import datamodels
-from ..datamodels import dqflags
+
+import numpy as np
 from scipy.interpolate import griddata
 from scipy.signal import convolve
 from scipy import mgrid
+
+from .. import datamodels
+from ..datamodels import dqflags
 
 
 log = logging.getLogger(__name__)
@@ -19,6 +17,9 @@ OFF_DELTA = 2  # for searching +/-2 around nominal offsets
 PSF_SIZE = 200  # half width of psf for 12 waves
 BLUR_SIZE = 10  # size of gaussian kernel for convolution
 N_SIZE = 2  # size of neighborhood in create_griddata_array
+
+
+DO_NOT_USE = dqflags.pixel["DO_NOT_USE"]
 
 
 class DataSet:
@@ -138,11 +139,9 @@ class DataSet:
             # 1. Create smoothed image of input SCI data of image #1
             # 1a. create image to smooth by first setting bad DQ pixels equal
             #     to mean of good pixels
-            data_1 = self.input_1.data.astype(np.float64)
-            wh_bad_1 = np.where(np.bitwise_and(self.input_1.dq, dqflags.pixel['DO_NOT_USE']))
-            wh_good_1 = np.where(np.logical_not(np.bitwise_and(self.input_1.dq, dqflags.pixel['DO_NOT_USE'])))
-
-            data_1[wh_bad_1] = data_1[wh_good_1].mean()
+            data_1 = self.input_1.data.astype(np.float)
+            bad1 = np.bitwise_and(self.input_1.dq, DO_NOT_USE).astype(np.bool)
+            data_1[bad1] = data_1[~bad1].mean()
 
             # 1b. Create smoothed image by smoothing this 'repaired' image
             g = gauss_kern(BLUR_SIZE, sizey=None)
@@ -196,7 +195,7 @@ class DataSet:
             log.info('Values for the refined offsets are, for x,y : %s %s',
                      self.off_x, self.off_y)
 
-        log.info('Final values for x,y offsets:%s %s', self.off_x, self.off_y)
+        log.info(f"Final x,y offset in pixels: {self.off_x} {self.off_y}")
 
         # Do final alignment for original (not interpolated) image #2
         data_2_a, dq_2_a, err_2_a = self.apply_final_offsets()
@@ -272,7 +271,7 @@ class DataSet:
 
         return off_x, off_y
 
-    def create_combined(self, im_1_a, im_2_a):
+    def create_combined(self, image1, image2):
         """
         Short Summary
         -------------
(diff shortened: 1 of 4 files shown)


jwst/wfs_combine/wfs_combine.py
```diff
@@ -315,42 +314,36 @@ def create_combined(self, im_1_a, im_2_a):
...
 
-        sci_data_1 = im_1_a.data.astype(np.float64)
-        sci_data_2 = im_2_a.data.astype(np.float64)
-
-        dq_data_1 = im_1_a.dq.copy()
-        dq_data_2 = im_2_a.dq.copy()
-        err_data_1 = im_1_a.err.copy()
-        err_data_2 = im_2_a.err.copy()
-
-        bad1 = np.bitwise_and(dq_data_1, dqflags.pixel['DO_NOT_USE'])
```
jdavies-st, line 326: The bug was here (and 2 lines down).  `np.bitwise_and` returns an array of 0 and 1 values, and cannot be used to index another array without being converted to a boolean array.  Once converted, it can have the True/False values flipped via that `~` operator.

jemorrison (approved):

jemorrison:
Ha I see my problem (bug). Hopefully I will not make that mistake again.

jwst/wfs_combine/wfs_combine_step.py
```diff
@@ -47,14 +46,18 @@ def process(self, input_table):
...
-            output_model.meta.asn.table_name = op.basename(input_table)
+            output_model.meta.asn.table_name = os.path.basename(input_table)
 
             # Save the output file
             self.save_model(
                 output_model, suffix='wfscmb', output_file=outfile, format=False
             )
 
-        return None
+        return output_model
```
hbushouse, line 63: So lines 59-61 are still in there to save the results to a file, as well as now passing `output_model` back to `stpipe`. Is `stpipe` going to try to save the model to a file too?

hbushouse (approved):
Looks great overall (lots of nice cleanup), but just one question about saving the results.

jwst/wfs_combine/wfs_combine_step.py
jdavies-st: That's a good question.  Let me check in the regression test run.

jdavies-st: Yes, it auto write out each result under the wrong name.  Lovely.  Let me fix this.

jdavies-st: So if I put it into a `if self.save_results:` block, it will then save it out properly. One then needs to supress the autosaving because there's the `return` statement.  Perhaps @stscieisenhamer has a better way of handling this.  I think this can be taken care of in issue #5511, as currently this step theoretically could return multiple products, but has no way to do so without utilizing a ModelContainer.

jdavies-st:
> Ha I see my problem (bug). Hopefully I will not make that mistake again.

Count on making it again.  😅 

I've made this mistake countless times myself, and did while trying to debug this.  Writing the test sorted me out.

jwst/wfs_combine/wfs_combine_step.py
hbushouse: What kind of wrong name is `stpipe` using when it saves to file? We should be able to tell it the proper product name suffix. @stscieisenhamer is the expert in that.

jdavies-st: The following:

```
$ ls /tmp/scratch/test_nircam_wfs_run_pipelines0/
config/
jw00632-o003_20191210t194815_wfs-image2_001_asn.json
jw00632-o003_20191210t194815_wfs-image3_001_asn.json
jw00632-o003_20191210t194815_wfs-image3_001_asn_wfscombine.fits <<<<<<<<<<<<
jw00632-o003_20191210t194815_wfs-image3_002_asn.json
jw00632-o003_20191210t194815_wfs-image3_002_asn_wfscombine.fits <<<<<<<<<<<<
jw00632-o003_t001_nircam_f212n-wlp8-nrca4_wfscmb-05.fits
jw00632-o003_t001_nircam_f212n-wlp8-nrca4_wfscmb-05_refine.fits
jw00632003002_03105_00001_nrca4_cal.fits
jw00632003002_03105_00001_nrca4_rate.fits
jw00632003002_03105_00002_nrca4_cal.fits
jw00632003002_03105_00002_nrca4_rate.fits
truth/
```

jdavies-st: The correct names are the `*wfscmb-05*.fits` ones which are generated from the name template in the association file.  How to get that into `meta.filename` of the returned model is not entirely clear to me.

stscieisenhamer: I believe the original code, where `return None` is used, was to handle the situation that the `stpipe` architecture did not handle multiple outputs. If there is not a reason to change this, I would suggest leaving it as-is.

However, if there is a need/desire to have the results returned, then the current way is as follows:

- Place each result in a `ModelContainer`
- For each resulting model, ensure that `result.filename = which_set['name']`
- ensure that the step parameter `self.output_use_mode = True`
- ensure that the step parameter `self.suffix = 'wfscb'`
- Remove the `save_model` statement

Ignoring possible other details being forgotten, this would fix the `return None`. IMHO, may want to leave the `return None` as-is.

stscieisenhamer: Though untested, theoretically the above setup should work when the return is a list of DataModels, without use a `ModelContainer`.  Did I say just leave the `return None` as-is?

jdavies-st: There's a very strong use case for returning something, as this is a step, and in order to test it, one needs the results of the step.

jdavies-st: And one also needs control over the `format` arg in `Step.save_model(format=False)`.  Without this, one gets:

```
jw00632-o003_t001_nircam_f212n-wlp8-nrca4_{suffix}-05_wfscmb.fits
```

jdavies-st: The formatting template in the association file is:

```
"name": "jw00632-o003_t001_nircam_f212n-wlp8-nrca4_{suffix}-05",
```

jdavies-st: So the current state of this PR produces the correct name, doesn't produce the incorrect name, and does so by hijacking the knowledge of when stpipe turns on `save_results`.  This is not ideal, but given that doing it properly would require a small refactor, I don't think we can do it before 7.7

hbushouse: I think the only time this step would return multiple results is if you gave it an input ASN file that defines multiple products. This is not the case in normal operations (only 1 product per ASN).

jdavies-st: I added some comments to why it is implemented in this hacky way now.  In the future, perhaps we can refactor `Step.save_model` to expose `format` as a class or instance attribute.

Once tests pass I will merge.

hbushouse (approved):
I'm Howard Bushouse and I approve this PR.

PR #5524, opened 2020-12-09
JP-1801: Update MIRI LRS slitless docs

### LINKED ISSUE #5522: Update pipeline stages table for MIRI LRS w/ TSOVISIT=False
_Issue [JP-1801](https://jira.stsci.edu/browse/JP-1801) was created on JIRA by [Alicia Canipe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=acanipe):_

According to this ticket: <https://jira.stsci.edu/browse/JWSTDMS-462> MIRI LRS slitless w/ TSOVISIT=False shouldn't be run through level-3. We need to update the [Stages vs. Exposure Type](https://jwst-pipeline.readthedocs.io/en/latest/jwst/pipeline/main.html#pipelines-vs-exposure-type) table to reflect that change. 

 

[Sarah Kendrew](https://jira.stsci.edu/secure/ViewProfile.jspa?name=skendrew) FYI

hbushouse:
Update the exposure-type vs pipeline modules chart in the pipeline docs to show that MIRI LRS slitless exposures with TSOVISIT=False do not receive any level 3 processing. Also fixed the entry for NRS lamp exposures, now that we're capable of processing them through level 2b.

Fixes #5522 / [JP-1801](https://jira.stsci.edu/browse/JP-1801)

### DIFF
--- a/docs/jwst/pipeline/main.rst
+++ b/docs/jwst/pipeline/main.rst
@@ -148,7 +148,7 @@ mode are marked as "N/A" in the TSOVISIT column.
 +---------------------+----------+-------------------+-----------------------+------------------+
 | | MIR_LRS-SLITLESS  | True     | calwebb_tso1      | calwebb_tso-spec2     | calwebb_tso3     |
 +                     +----------+-------------------+-----------------------+------------------+
-|                     | False    | calwebb_detector1 | calwebb_spec2         | calwebb_spec3    |
+|                     | False    | calwebb_detector1 | calwebb_spec2         | N/A              |
 +---------------------+----------+-------------------+-----------------------+------------------+
 | | MIR_MRS           | N/A      | calwebb_detector1 | calwebb_spec2         | calwebb_spec3    |
 +---------------------+----------+-------------------+-----------------------+------------------+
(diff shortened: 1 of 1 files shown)


aliciacanipe (approved):
Looks good, thanks Howard!

stscieisenhamer (approved):
Nit: The TSO flags are in opposite order across some entries, sometimes TRUE is first, other times not.

Also, does this make the table documented in level2_asn_rules.rst obsolete?  Did find that some entries in that table are wrong, as below. But maybe that table should be removed.

- nrs_autoflat: should be nrslamp-spec2
-nrs_autowave: should be nrslamp-spec3
- nrs_brightobj: tso-spec2

and possibly others for the "either/or" case.

tapastro (approved):
Looks good, I didn't know this page of the docs existed! I wish the readthedocs would adapt the width of the column a bit, having to scroll for this table is a bit of a nuisance.

hbushouse:
> Nit: The TSO flags are in opposite order across some entries, sometimes TRUE is first, other times not.

I think that was done in order to have the default mode appear first. For example MIRI LRS Slitless and NIRISS SOSS would have TSO listed first, because that's the default, while modes like NIRCam and MIRI imaging would have TSO listed second.

hbushouse:
> Also, does this make the table documented in level2_asn_rules.rst obsolete? Did find that some entries in that table are wrong, as below. But maybe that table should be removed.
> 
> * nrs_autoflat: should be nrslamp-spec2
>   -nrs_autowave: should be nrslamp-spec3
> * nrs_brightobj: tso-spec2
> 
> and possibly others for the "either/or" case.

Those are the correct values for NRS lamps and Brightobj, now that we're actually able to process lamp exposures through level-2b.

Issue #5523, opened 2020-12-08
Add SUBPXPTS as source in constraints everywhere SUBPXPNS is used


stscijgbot-jp:
_Issue [JP-1802](https://jira.stsci.edu/browse/JP-1802) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Due to changes in dithering keywords, SUBPXPNS has been renamed as SUBPXPTS. For backwards compatibility, add SUBPXPTS as a source for all constraints that use SUBPXPNS.

stscijgbot-jp:
_Comment by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham) on [JIRA](https://jira.stsci.edu/browse/JP-1802?focusedCommentId=499558#comment-499558):_

Resolved by [PR#&#x2060;5525](https://github.com/spacetelescope/jwst/pull/5525)

PR #5525, opened 2020-12-09
JP-1802 Add new dither keyword subpxpts to constraints

stscieisenhamer:
Resolves [JP-1802](https://jira.stsci.edu/browse/JP-1802) / #5523

### DIFF
--- a/jwst/associations/lib/rules_level2b.py
+++ b/jwst/associations/lib/rules_level2b.py
@@ -760,8 +760,8 @@ class Asn_Lv2NRSFSS(
                                 sources=['numdthpt'],
                             ),
                             DMSAttrConstraint(
-                                name='subpxpns',
-                                sources=['subpxpns'],
+                                name='subpxpts',
+                                sources=['subpxpns', 'subpxpts'],
                             ),
                             SimpleConstraint(
                                 value='science',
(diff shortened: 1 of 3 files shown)


jwst/associations/lib/rules_level2b.py
```diff
@@ -760,8 +760,8 @@ def __init__(self, *args, **kwargs):
                                 sources=['numdthpt'],
                             ),
                             DMSAttrConstraint(
-                                name='subpxpns',
-                                sources=['subpxpns'],
+                                name='subpxpts',
+                                sources=['subpxpns', 'subpxpts'],
```
hbushouse, line 764: Yay for backwards compatibility!

hbushouse (approved):
Requisite reminder for change log entry ...

jwst/associations/lib/rules_level2b.py
stscieisenhamer: and, believe it or not, by design!

hbushouse: And what a clever design. One instance here that lists both keywords as sources can then be referenced anywhere else simply using the single constraint name "subpxpts" and it works like magic!

tapastro (approved):
☑️ 

This got me reading associations.lib.constraint, which I might have to save for another day...

PR #5529, opened 2020-12-11
Fix units in tso photometry catalog

larrybradley:
This PR updates the `tso_photometry` step for input data units in `MJy/sr`.  The output catalog columns are now astropy `Quantity` objects with units of `Jy`.

Addresses part of https://jira.stsci.edu/browse/JP-1559

CC: @hbushouse, @tapastro

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -176,6 +176,8 @@ tso_photometry
 - Place aperture using header keywords XREF_SCI and YREF_SCI instead of
   CRPIX1 and CRPIX2 [#5533]
 
+- Fixed the flux units in the output photometry catalog. [#5529]
+
 tweakreg
 --------
 
--- a/jwst/tso_photometry/tso_photometry.py
+++ b/jwst/tso_photometry/tso_photometry.py
@@ -6,6 +6,7 @@ import logging
 import numpy as np
 from astropy.table import QTable
 from astropy.time import Time, TimeDelta
+import astropy.units as u
 import photutils
 from photutils import CircularAperture, CircularAnnulus
 
@@ -58,6 +59,15 @@ def tso_aperture_photometry(datamodel, xcenter, ycenter, radius, radius_inner,
         phot_aper = CircularAperture((xcenter, ycenter), r=radius)
         bkg_aper = CircularAnnulus((xcenter, ycenter), r_in=radius_inner,
                                    r_out=radius_outer)
+
+    # convert the input data and errors from MJy/sr to Jy
+    if datamodel.meta.bunit_data != 'MJy/sr':
+        raise ValueError('data is expected to be in units of MJy/sr')
+    factor = 1.e6 * datamodel.meta.photometry.pixelarea_steradians
+    datamodel.data *= factor
+    datamodel.err *= factor
+    datamodel.meta.bunit_data = 'Jy'
+    datamodel.meta.bunit_err = 'Jy'
 
     aperture_sum = []
     aperture_sum_err = []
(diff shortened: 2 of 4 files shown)


tapastro (approved):
This code looks good to me, though upon trying it out, I've found an issue with datamodel.meta.photometry.pixelarea_steradians for all MIRI image data I've looked at. The difference between pixelarea_steradians and pixelarea_arcsecsq should be a constant value of 2.3504e-11, and the difference between these two FITS header keywords (PIXAR_SR and PIXAR_A2) is something else... possibly a MIRIsim issue?

hbushouse:
@tapastro PIXAR_SR and PIXAR_A2 are read from the PHOTOM ref file during the photom step and stored in the science product keywords. So if there's a discrepancy in the values, that suggests an error in the values stored in the MIRI imaging photom ref files.

hbushouse (approved):
Code wise it looks fine to me. Is this now consistent with the units of photometry produced by the source_catalog step, for normal imaging?

Don't forget to add a change log entry (in CHANGES.rst).

tapastro:
> @tapastro PIXAR_SR and PIXAR_A2 are read from the PHOTOM ref file during the photom step and stored in the science product keywords. So if there's a discrepancy in the values, that suggests an error in the values stored in the MIRI imaging photom ref files.

@hbushouse Ah, I tried to grep a source for the values but didn't understand how the photom step worked. It does appear there, perhaps I should bring it up with Jonathan? A CRDS ticket?

hbushouse:
Any errors in the keyword values in reference files would be the responsibility of the instrument team to fix and redeliver.

larrybradley:
@hbushouse I added a changelog entry.

CHANGES.rst
```diff
@@ -277,6 +277,8 @@ tso_photometry
 - Fix a bug in the computation of integration time stamps when the INT_TIMES
   table is not available. [#5318]
 
+- Fixed the flux units in the output photometry catalog. [#5529]
```
hbushouse, line 280: This needs to be moved up into the 0.17.2 (unreleased) section of the log file. Create a tso_photometry subsection up there (e.g. up around line 170).

jwst/tso_photometry/tso_photometry.py
```diff
@@ -59,6 +60,15 @@ def tso_aperture_photometry(datamodel, xcenter, ycenter, radius, radius_inner,
         bkg_aper = CircularAnnulus((xcenter, ycenter), r_in=radius_inner,
                                    r_out=radius_outer)
 
+    # convert the input data and errors from MJy/sr to Jy
+    if datamodel.meta.bunit_data != 'MJy/sr':
```
hbushouse, line 64: Probably want to add a test for existence here, in case the BUNIT keyword isn't in the header.

skendrew:
this should be queried with the MIRI teamers who are responsible for this reference file - I believe that's Macarena Garcia Marin. I will drop her a quick message.

jwst/tso_photometry/tso_photometry.py
larrybradley: done

CHANGES.rst
larrybradley: done

jwst/tso_photometry/tso_photometry.py
hbushouse: I don't see a code change. Or has the updated version not been committed yet?

larrybradley: pushed now, thanks.

larrybradley:
The failing tests are all unrelated to this PR.

jwst/tso_photometry/tso_photometry.py
hbushouse: What we have here is failure to communicate ... when I said "add a test for existence" I was referring to a check in the code to make sure that `datamodel.meta.bunit_data` actually exists, before trying to do a test on its value, not the addition of a unit test for the case where the keyword doesn't exist (although the latter doesn't hurt either). I'm just worried that the step itself is going to crash with some kind of weird NoneType error if `datamodel.meta.bunit_data` isn't defined or has the value `None`. So check that the attribute `is not None` before doing the value comparison.

larrybradley: I did both.  Is this not what you were looking for:  https://github.com/spacetelescope/jwst/pull/5529/files#diff-486ca4d2c6af260ec03cb90061a92b296db28d424ccb5904b1f11f69f07349fdR40-R43

larrybradley: https://github.com/spacetelescope/jwst/pull/5529/commits/0cf23cef8fa62ab6d943bc3d6a226db65d2c5127

hbushouse: D'oh! Yep, that's it. I was looking for it in this module and didn't see it in the step module. Never mind ...

larrybradley:
Can I merge this given that the test failures are all unrelated to this PR?

hbushouse:
Yes, go ahead and merge.

PR #5532, opened 2020-12-14
JP-1795 Asn_Lv2WFSS: Add segmentation map exposure

### LINKED ISSUE #5514: Update WFSS lvl2b ASN rules to include segmentation map
_Issue [JP-1795](https://jira.stsci.edu/browse/JP-1795) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The WFSS contamination correction scheme will need access to the segmentation map created from the direct images. Therefore the ASN rules for creating level-2b ASN tables for WFSS exposures need to be updated to list the segmentation map file, along with the existing source catalog file and direct image names.

stscieisenhamer:
Resolves [JP-1795](https://jira.stsci.edu/browse/JP-1795)
Fixes #5514

### DIFF
--- a/jwst/associations/lib/rules_level2b.py
+++ b/jwst/associations/lib/rules_level2b.py
@@ -637,7 +637,7 @@ class Asn_Lv2WFSS(
             for idx in sorted(direct_idxs, reverse=True)
         ))
 
-        # Add the Level3 catalog and direct image members
+        # Add the Level3 catalog, direct image, and segmentation map members
         lv3_direct_image_root = DMS_Level3_Base._dms_product_name(self)
         members.append(
             Member({
@@ -649,6 +649,12 @@ class Asn_Lv2WFSS(
             Member({
                 'expname': lv3_direct_image_root + '_cat.ecsv',
                 'exptype': 'sourcecat'
+            })
+        )
+        members.append(
+            Member({
+                'expname': lv3_direct_image_root + '_seg.fits',
+                'exptype': 'segmap'
             })
         )
 
--- a/jwst/associations/lib/rules_level2b.py
+++ b/jwst/associations/lib/rules_level2b.py
@@ -653,7 +653,7 @@ class Asn_Lv2WFSS(
         )
         members.append(
             Member({
-                'expname': lv3_direct_image_root + '_seg.fits',
+                'expname': lv3_direct_image_root + '_segm.fits',
                 'exptype': 'segmap'
             })
         )
--- a/jwst/associations/tests/test_wfss.py
+++ b/jwst/associations/tests/test_wfss.py
@@ -38,12 +38,14 @@ def test_niriss_wfss():
         for asn in asns
     }
 
-    # Ensure catalog name is correct in the spec2 associations
-    source_cat = asn_by_type['image3']['products'][0]['name']
-    source_cat += '_cat.ecsv'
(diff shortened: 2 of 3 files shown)


ddavis-stsci (approved):
LGTM

hbushouse (approved):
Looks good. 

Will it cause any harm to have the seg file listed in the ASN even if the file itself doesn't exist (i.e. until such time as the source_catalog step is updated to produce seg files)? None of the pipelines or steps have yet been updated to try to make use of the seg file, so nothing should actually try to open it.

Oh, and requisite change log entry reminder ...

stscieisenhamer:
Nothing cal pipeline related will fail; if a pipeline doesn't tried to use it, it might as well not exist.

The issue will actually be with SDP: Associations will not be executed unless all files in the association exist. Since this file is not yet produced by anything, these will hang processing in SDP.

hbushouse:
> The issue will actually be with SDP: Associations will not be executed unless all files in the association exist. Since this file is not yet produced by anything, these will hang processing in SDP.

OK, that's not good. It would mean that any WFSS exposure would never get calwebb_spec2 processing triggered. So ... do we not merge this PR into the build for 7.7?

hbushouse:
Unrelated, but does anyone have any clues what's causing the failure of the Security Check in Travis CI for all PR's today?

tapastro (approved):
Looks good 👍

stscieisenhamer:
We should discuss with SDP how they would like to handle. They can selectively choose which `exptype`s to actively search for. This case would be to not search for `segmap` until otherwise indicated. However, I do not know how invasive this is to their process. A standup topic!

stscieisenhamer:
This PR cannot be merged until [JP-1778](https://jira.stsci.edu/browse/JP-1778) is resolved.

jwst/associations/lib/rules_level2b.py
```diff
@@ -597,6 +597,12 @@ def add_catalog_members(self):
                 'exptype': 'sourcecat'
             })
         )
+        members.append(
+            Member({
+                'expname': lv3_direct_image_root + '_seg.fits',
```
hbushouse, line 602: The PR that updated the `source_cat` step used a suffix of "segm", not "seg"

jwst/associations/tests/test_wfss.py
```diff
@@ -38,12 +38,14 @@ def test_niriss_wfss():
         for asn in asns
     }
 
-    # Ensure catalog name is correct in the spec2 associations
-    source_cat = asn_by_type['image3']['products'][0]['name']
-    source_cat += '_cat.ecsv'
+    # Ensure catalog and segmentation map names are correct in the spec2 associations
+    l3name = asn_by_type['image3']['products'][0]['name']
+    source_cat = l3name + '_cat.ecsv'
+    segmap = l3name + '_seg.fits'
```
hbushouse, line 44: Should be "segm"

hbushouse (changes requested):
Needs a couple updates to change instances of "seg" to "segm"

jwst/associations/lib/rules_level2b.py
tapastro: 😬

hbushouse (approved):
LGTM

PR #5533, opened 2020-12-14
JP 1559: Fix tso_photometry keywords for aperture location

### LINKED ISSUE #5113: MIRI TSO photometry not using the right positional coordinates
_Issue [JP-1559](https://jira.stsci.edu/browse/JP-1559) was created on JIRA by [Sarah Kendrew](https://jira.stsci.edu/secure/ViewProfile.jspa?name=skendrew):_

The TSO photometry step (in CALTSO3) does not place the photometric aperture in the correct location for MIRI data. The CRPIX keywords are set to the left-bottom of the array, rather than the nominal pointing position at the centre of the array (where the target is located in our test dataset). The issue was raised by [Misty Cracraft](https://jira.stsci.edu/secure/ViewProfile.jspa?name=cracraft) in testing, and the following captures the salient points from a Slack discussion between [Misty Cracraft](https://jira.stsci.edu/secure/ViewProfile.jspa?name=cracraft) , [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies) , [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse) , [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw) and [Sarah Kendrew](https://jira.stsci.edu/secure/ViewProfile.jspa?name=skendrew) :
 - The tso_photometry pipeline step assumes that the target is at the nominal pointing location which should be pulled from the SIAF (the XSciRef and YSciRef values).

 - Ideally the pipeline should use the TARG_RA and TARG_DEC keywords recording the location of the source, which the embedded WCS can then map to a specific detector location. But this has not yet been implemented given many problems with these keywords being incorrect in test activities.

 - further comment from [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse) : The CRPIX1/2 and X/YREF_SCI values both come from the SIAF and are added during level-1b processing, within the set_telescope_pointing script that computes and populates all WCS-related keywords. So these are NOT set by the Cal pipeline itself, but rather are already in the level-1b products used as input to the Cal pipeline. Both sets of those keywords come from the same SIAF entries, namely XSciRef and YSciRef. Sticking their values into CRPIX1/2 for imaging is a holdover from the old days. Once we dropped FITS WCS keywords for spectra we invented the X/YREF_SCI keyword to take their place. So we could presumably update both set_telescope_pointing and tso_photometry to populate and use X/YREF_SCI for both imaging and spectra and be done with CRPIX.

tapastro:
Updated the keyword usage for aperture placement from CRPIX1/2 to X/YREF_SCI. 

Related to resolving JP-1559.
Fixes #5113

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -162,6 +162,12 @@ tweakreg
 - Add support for the new ``fitgeom`` mode: ``'rshift'`` that can fit only
   for shifts and a rotation. [#5475]
 
+white_light
+-----------
+
+- Add support for step parameters ``min_wavelength`` and ``max_wavelength`` to modify
+  the wavelength region over which the flux integration is calculated. [#5501]
+
 0.17.1 (2020-09-15)
 ===================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -160,7 +160,7 @@ tso_photometry
 --------------
 
 - Place aperture using header keywords XREF_SCI and YREF_SCI instead of
-  CRPIX1 and CRPIX2
+  CRPIX1 and CRPIX2 [#5533]
 
 tweakreg
 --------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -156,6 +156,12 @@ stpipe
 
 - Implement utility function all_steps and fix crds reference file retrieval for non-datamodels [#5492]
 
+tso_photometry
+--------------
+
+- Place aperture using header keywords XREF_SCI and YREF_SCI instead of
+  CRPIX1 and CRPIX2
+
 tweakreg
 --------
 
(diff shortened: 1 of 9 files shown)


hbushouse (approved):
Looks good. Just need to add a change log entry.

CHANGES.rst
```diff
@@ -156,6 +156,12 @@ stpipe
 
 - Implement utility function all_steps and fix crds reference file retrieval for non-datamodels [#5492]
 
+tso_photometry
+--------------
+
+- Place aperture using header keywords XREF_SCI and YREF_SCI instead of
```
stscieisenhamer, line 162: Add the PR number here.

stscieisenhamer (approved):
Besides a minor update to the changelog, LGTM.

PR #5536, opened 2020-12-15
Remove deprecated method get_children from XML elements

stscirij:
Also PATTSIZE needs to be a string


hbushouse (approved):
Requisite reminder for a change log entry ...

PR #5538, opened 2020-12-15
Update resample to handle SIP coeffs in input data

jdavies-st:
A follow-on from #5507 to allow `resample` to handle input data that now has CD matrix and SIP coeffs.

Addresses #4673 / [JP-1347](https://jira.stsci.edu/browse/JP-1347)

### DIFF
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -184,13 +184,13 @@ class ResampleData:
         Update FITS WCS keywords of the resampled image.
         """
         # Delete any SIP-related keywords first
-        pattern = r"(cd[123]_[123]|ap?_|bp?_)"
+        pattern = r"^(cd[12]_[12]|[ab]p?_\d_\d|[ab]p?_order)$"
         regex = re.compile(pattern)
-        # Make a copy so we don't delete keys in the model while iterating
-        wcsinfo = model.meta.wcsinfo.instance.copy()
-        for item in wcsinfo:
-            if regex.match(item):
-                del model.meta.wcsinfo.instance[item]
+
+        keys = list(model.meta.wcsinfo.instance.keys())
+        for key in keys:
+            if regex.match(key):
+                del model.meta.wcsinfo.instance[key]
 
         # Write new PC-matrix-based WCS based on GWCS model
         transform = model.meta.wcs.forward_transform
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -1,4 +1,6 @@
 import logging
+import re
+
 import numpy as np
 
 from . import gwcs_drizzle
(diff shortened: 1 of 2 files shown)


jdavies-st:
@mcara and @hbushouse this fixes the issue, but I agree that cleaning out `meta.wcsinfo` completely would be preferable.  Is there anything that _ought_ to stay in `meta.wcsinfo` in a resampled product other than the computed FITS WCS?

jdavies-st:
Currently, the whole meta tree is populated from the first input file:

https://github.com/spacetelescope/jwst/blob/1245580a4ea4918137e2268dbcc8e469331605ff/jwst/resample/resample.py#L85-L88

We don't have a way to ignore certain branches with `update()` but of course we could nuke it from space after doing `update()` if needed.  💭

hbushouse:
You can run .update() on only the primary header, which will skip all the FITS WCS stuff in the SCI extension (model.meta.wcsinfo). Of course I don't know what it does with the WCS object (model.meta.wcs), which isn't relegated to a particular FITS extension.

hbushouse:
I get a warm and fuzzy feeling from the suggestion of `model.meta.wcs.nukeit()`.

hbushouse:
Can you just skip https://github.com/spacetelescope/jwst/blob/1245580a4ea4918137e2268dbcc8e469331605ff/jwst/resample/resample.py#L88 and build the meta.wcs from scratch? And also do `model.meta.wcsinfo=None` and rebuild it from scratch?

hbushouse:
Oh, sorry. https://github.com/spacetelescope/jwst/blob/1245580a4ea4918137e2268dbcc8e469331605ff/jwst/resample/resample.py#L88 is already creating the WCS object from scratch, by copying the *output* WCS. For a minute there I thought it was copying the *input* WCS. Never mind.

hbushouse (approved):
I guess this works. I assume that `output.meta.wcs` is already clean, because it's a fresh copy of the WCS created for the output frame?

Don't forget to add a change log entry.

jwst/resample/resample.py
```diff
@@ -181,6 +183,16 @@ def update_fits_wcs(self, model):
         """
         Update FITS WCS keywords of the resampled image.
         """
+        # Delete any SIP-related keywords first
+        pattern = r"(cd[123]_[123]|ap?_|bp?_)"
+        regex = re.compile(pattern)
+        # Make a copy so we don't delete keys in the model while iterating
+        wcsinfo = model.meta.wcsinfo.instance.copy()
```
cshanahan1, line 190: (per our slack convo) change this to list(model.meta.wcsinfo.instance.keys()) instead of copying the whole thing.

jwst/resample/resample.py
```diff
@@ -181,6 +183,16 @@ def update_fits_wcs(self, model):
         """
         Update FITS WCS keywords of the resampled image.
         """
+        # Delete any SIP-related keywords first
+        pattern = r"(cd[123]_[123]|ap?_|bp?_)"
```
cshanahan1, line 187: this will delete anything with 'a_', 'b_' in it, so maybe make this a bit more strict to avoid deleting any future keywords that might contain these strings.

cshanahan1 (commented):
Looks good to me! I made some minor suggestions.

jwst/resample/resample.py
jdavies-st: Thanks.  Yeah, let me make this much more strict.

jwst/resample/tests/test_resample_step.py
```diff
@@ -235,3 +235,23 @@ def test_pixel_scale_ratio_imaging(nircam_rate, ratio):
     area1 = result1.meta.photometry.pixelarea_steradians
     area2 = result2.meta.photometry.pixelarea_steradians
     assert_allclose(area1 * ratio**2, area2, rtol=1e-6)
+
+
+def test_sip_coeffs_do_not_propagate(nircam_rate):
+    im = AssignWcsStep.call(nircam_rate, sip_degree=2)
```
jdavies-st, line 241: Btw, if I didn't specify `sip_degree=2` here, the sip fitting would fit with just a 1st order affine transformation, i.e. just a CD matrix, so no SIP coeffs and ctypes didn't have the `-SIP` suffix.

Issue #5453, opened 2020-10-30
set_telescope_pointing should use proposal info when engineering is all zero


stscijgbot-jp:
_Issue [JP-1771](https://jira.stsci.edu/browse/JP-1771) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

In `set_telescope_pointing`, there are conditions where the pointing mnemonics are available, yet are all zero. In this situation, the pointing keywords should default to the proposal values.

stscijgbot-jp:
_Comment by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse) on [JIRA](https://jira.stsci.edu/browse/JP-1771?focusedCommentId=500830#comment-500830):_

Fixed in #5540

PR #5540, opened 2020-12-16
Set_telescope_pointing to handle zero values for pointing mnemonics in engineering database

stscirij:
Addresses #5453 and [JP-1771](https://jira.stsci.edu/browse/JP-1771)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -143,6 +143,7 @@ set_telescope_pointing
 ----------------------
 
 - Add logging of the found quaternion information [#5495]
+- Handle cases where engineering database's pointing mnemonics are all zero over the requested time range [#5540]
 
 skymatch
 --------
--- a/jwst/lib/set_telescope_pointing.py
+++ b/jwst/lib/set_telescope_pointing.py
@@ -592,8 +592,12 @@ def calc_wcs_over_time(obsstart, obsend, engdb_url=None, tolerance=60, reduce_fu
     vinfos = list()
 
     # Calculate WCS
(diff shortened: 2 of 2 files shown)


hbushouse (approved):
Looks good to me.

I assume that the errors raised in the lower-level routines get trapped at the upper level, so that processing continues, but with default values used for the pointing (e.g. values from TARG_RA/DEC)? We don't want level 1 processing falling dead in its tracks for all the cases of bad data we have in some of the DMS test suites.

hbushouse:
There's 1 flake8 error that needs fixing.

stscieisenhamer (approved):

stscirij:
> Looks good to me.
> 
> I assume that the errors raised in the lower-level routines get trapped at the upper level, so that processing continues, but with default values used for the pointing (e.g. values from TARG_RA/DEC)? We don't want level 1 processing falling dead in its tracks for all the cases of bad data we have in some of the DMS test suites.



> Looks good to me.
> 
> I assume that the errors raised in the lower-level routines get trapped at the upper level, so that processing continues, but with default values used for the pointing (e.g. values from TARG_RA/DEC)? We don't want level 1 processing falling dead in its tracks for all the cases of bad data we have in some of the DMS test suites.

Yes, set_telescope_pointing traps errors if the engineering database can't be opened or if the telemetry values are zero, and uses the header keywords instead, but only if --allow-defaults is specified on the command line

v1_calculate exits in those 2 cases, but it isnt used to update the WCS, just to check that things are working

PR #5548, opened 2020-12-17
 Update Ami analyze to input two new input parameters: JP-1742

### LINKED ISSUE #5388: Add parameters for updated ami_analyze algorithm
_Issue [JP-1742](https://jira.stsci.edu/browse/JP-1742) was created on JIRA by [David Grumm](https://jira.stsci.edu/secure/ViewProfile.jspa?name=grumm):_

The additional parameters (although not required) in the configuration file improve the created Affine2d object in the updated algorithm.

jemorrison:
Updates for ami analyze for JP-1742
The two new input parameters are 'set_psf_offset_find_rotation' and 'set_rotsearch_d' 

These seems a bit long. When I was testing I tripped up the exact name of the parameter. Maybe we could use something shorter- maybe 'set_psf_offset'  'set_rotsearch' - suggestions - or just leave it as it is.

Fixes #5388 / [JP-1742](https://jira.stsci.edu/browse/JP-1742)

### DIFF
--- a/jwst/ami/ami_analyze.py
+++ b/jwst/ami/ami_analyze.py
@@ -17,7 +17,7 @@ log.addHandler(logging.NullHandler())
 
 
 def apply_LG_plus(input_model, filter_model, oversample, rotation,
-                  psf_offset_find_rotation, rotsearch_parameters):
+                  psf_offset, rotsearch_parameters):
     """
     Short Summary
     -------------
@@ -79,7 +79,7 @@ def apply_LG_plus(input_model, filter_model, oversample, rotation,
     filt = "F430M"
     rotsearch_d = np.arange(rotsearch_parameters[0], rotsearch_parameters[1], rotsearch_parameters[2])
 
-    affine2d = find_rotation(data[:,:], psf_offset_find_rotation, rotsearch_d,
+    affine2d = find_rotation(data[:,:], psf_offset, rotsearch_d,
                    mx, my, sx, sy, xo, yo,
                    PIXELSCALE_r, dim, bandpass, oversample, holeshape)
 
--- a/jwst/ami/ami_analyze.py
+++ b/jwst/ami/ami_analyze.py
@@ -16,7 +16,8 @@ log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
 
-def apply_LG_plus(input_model, filter_model, oversample, rotation):
+def apply_LG_plus(input_model, filter_model, oversample, rotation,
+                  psf_offset_find_rotation, rotsearch_parameters):
     """
     Short Summary
     -------------
@@ -66,9 +67,7 @@ def apply_LG_plus(input_model, filter_model, oversample, rotation):
     #   x0, y0: offsets in pupil space
     mx,my,sx,sy,xo,yo, = (1.0,1.0, 0.0,0.0, 0.0,0.0)
 
-    psf_offset_find_rotation = (0.0,0.0)
     psf_offset_ff = None
-    rotsearch_d = None
 
     lamc = 4.3e-6
     oversample = 11
@@ -78,7 +77,7 @@ def apply_LG_plus(input_model, filter_model, oversample, rotation):
     PIXELSCALE_r = pixelscale_as * arcsec2rad
     holeshape = 'hex'
     filt = "F430M"
-    rotsearch_d = np.arange(-3, 3.1, 1)
+    rotsearch_d = np.arange(rotsearch_parameters[0], rotsearch_parameters[1], rotsearch_parameters[2])
 
     affine2d = find_rotation(data[:,:], psf_offset_find_rotation, rotsearch_d,
                    mx, my, sx, sy, xo, yo,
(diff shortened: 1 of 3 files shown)


jwst/ami/ami_analyze_step.py
```diff
@@ -11,6 +11,8 @@ class AmiAnalyzeStep(Step):
     spec = """
         oversample = integer(default=3, min=1)  # Oversampling factor
         rotation = float(default=0.0)           # Rotation initial guess [deg]
+        set_psf_offset_find_rotation = string(default='0.0 0.0') # Psf offset values to use to create the model array
```
hbushouse, line 14: Definitely agree that shorter names would be good for both of these. Perhaps just `psf_offset` and `rotation_search`?

jwst/ami/ami_analyze_step.py
```diff
@@ -32,8 +34,18 @@ def process(self, input):
         # Retrieve the parameter values
         oversample = self.oversample
         rotate = self.rotation
-        self.log.info('Oversampling factor = %d', oversample)
-        self.log.info('Initial rotation guess = %g deg', rotate)
+
+        # pull out parameters that are strings and change to floats
+
+        psf_offset = self.set_psf_offset_find_rotation.split(' ')
+        psf_offset_find_rotation = (float(psf_offset[0]), float(psf_offset[1]))
```
jdavies-st, line 41: Slightly more readable would be:

```python
psf_offset_find_rotation = [float(a) for a in self.set_psf_offset_find_rotation.split()]
```

Same below for `rotsearch_parameters`.

jwst/ami/ami_analyze_step.py
```diff
@@ -32,8 +34,18 @@ def process(self, input):
...
+        rotsearch = self.set_rotsearch_d.split(' ')
+        rotsearch_parameters= [float(rotsearch[0]), float(rotsearch[1]), float(rotsearch[2])]
+
+        self.log.info(f'Oversampling factor =  {oversample}')
+        self.log.info(f'Initial rotation guess = {rotate} deg')
+        self.log.info(f'Initial values to use for psf offset = {psf_offset_find_rotation}')
+        self.log.info(f'Initial values to use for rotation search {rotsearch_parameters}')
```
jdavies-st, line 48: This logging already happens when the step is initialized.  It's not as nicely formatted, but it is there.  It then gets logged again, each on its own line:

```
2020-12-17 14:09:38,885 - stpipe.Ami3Pipeline.ami_analyze - INFO - Step ami_analyze parameters are: {'pre_hooks': [], 'post_hooks': [], 'output_file': None, 'output_dir': None, 'output_ext': '.fits', 'output_use_model': False, 'output_use_index': True, 'save_results': False, 'skip': False, 'suffix': None, 'search_output_file': True, 'input_dir': '/private/var/folders/jg/by5st33j7ps356dgb4kn8w900001n5/T/pytest-of-jdavies/pytest-30/test_niriss_ami3_run_pipeline0', 'oversample': 3, 'rotation': 1.49}
2020-12-17 14:09:38,885 - stpipe.Ami3Pipeline.ami_analyze - INFO - Oversampling factor = 3
2020-12-17 14:09:38,885 - stpipe.Ami3Pipeline.ami_analyze - INFO - Initial rotation guess = 1.49 deg
```

Would be nice if we could make `stpipe` print these to be more readable by using `pprint.pformat` or similar.

Fine to keep as is though I guess, as it was there originally.

hbushouse (commented):
Overall code structure looks fine. Just use shorter names everywhere for the params themselves.

And don't forget a change log entry.

jwst/ami/ami_analyze.py
```diff
@@ -78,7 +77,7 @@ def apply_LG_plus(input_model, filter_model, oversample, rotation):
     PIXELSCALE_r = pixelscale_as * arcsec2rad
     holeshape = 'hex'
     filt = "F430M"
-    rotsearch_d = np.arange(-3, 3.1, 1)
+    rotsearch_d = np.arange(rotsearch_parameters[0], rotsearch_parameters[1], rotsearch_parameters[2])
```
jdavies-st, line 80: It's odd that the defaults are such that the purpose of `[-3, 3.1, 1]` is to make an evenly-spaced interval between -3 and 3, spaced by 1.  But `[-3, 4, 1]` will produce the exact same thing.

```
In [25]: np.arange(-3, 3.1, 1)
Out[25]: array([-3., -2., -1.,  0.,  1.,  2.,  3.])

In [27]: np.arange(-3, 4., 1)
Out[27]: array([-3., -2., -1.,  0.,  1.,  2.,  3.])
```

Might be worth documenting that this is "start, stop, step" or similar in the spec comment.

jwst/ami/ami_analyze_step.py
jemorrison: I will make the change but it is not more readable to me. In fact I had to stare at that a while to understand what it was going

jdavies-st (approved):
I verified that the regression tests on this branch pass.  🎉

hbushouse (approved):

PR #5554, opened 2020-12-17
NRS slits should be larger than 1 pixel

### LINKED ISSUE #5539: Debug and fix error reported in JWSTDMS-461
_Issue [JP-1816](https://jira.stsci.edu/browse/JP-1816) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

JWSTDMS-461 reports a crash in the calwebb_spec2 flatfield step for a NIRSpec MOS exposure. It's crashing when trying to process a slit that had 2D cutout extents of [(2050, 2048) (718, 741)] in the extract_2d step, which clearly puts it off the edge of the image. So this slit should not have been created in the first place.

nden:
Resolves #5539 / JWSTDMS-461 / [JP-1816](https://jira.stsci.edu/browse/JP-1816)

Mark Nirspec slits which project on less than one pixel as invalid.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,8 @@ assign_wcs
 
 - Fix bug where subarray bounding boxes were 1 pixel too small. [#5543]
 
+- Mark Nirspec slits which project on less than one pixel as invalid. [#5554]
+
 associations
 ------------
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):
Needs a change log entry.

PR #5557, opened 2020-12-18
Constrain SIP approximation to use wcsinfo.crpix and simplify code

mcara:
Constrain SIP approximation to use `wcsinfo.crpix1,2` (keep it fixed to what was before) and simplify code that constrains the SIP fit to use polynomial degrees <= 6. This is possible due to https://github.com/spacetelescope/gwcs/pull/339

A follow on from #5538 and #5507


jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -94,25 +97,16 @@ def process(self, input, *args, **kwargs):
             return result
 
         # fit sip approx., degree is chosen by best fit
+        sip_degree = range(1, _MAX_SIP_DEGREE) if self.sip_degree is None else self.sip_degree
+        sip_inv_degree = range(1, _MAX_SIP_DEGREE) if self.sip_inv_degree is None else self.sip_inv_degree
+        crpix = [result.meta.wcsinfo.crpix1, result.meta.wcsinfo.crpix2]
```
jdavies-st, line 102: It would be nice to not depend on crpix values as inputs.  We don't need to constrain the SIP to use the same ones as are done for the gwcs model.  Though clearly we could.  Is there a good reason to do this?

mcara: What do you mean by "depend" and why it would be nice? I thought during one of the tag-ups you mentioned that for some instruments `crpix` is in a corner of the chip instead of being near the center of the bounding box. That was the motivation for https://github.com/spacetelescope/gwcs/pull/337 in the first place.

mcara: Plus, there is no harm in not modifying CRPIX value already recorded in `wcsinfo`.

mcara: That is, I see absolutely no benefit in moving/modifying `wcsinfo.crpix1,2`.

jdavies-st: Maybe if `CRPIXn` is not defined, default to the center?  So if `crpix = [None, None]`, set `crpix = None` and then do the fit?

mcara: Done

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -94,25 +97,17 @@ def process(self, input, *args, **kwargs):
...
+        crpix = None if None in crpix else crpix
         try:
             fit_sip_hdr = result.meta.wcs.to_fits_sip(None, self.sip_max_pix_error,
-                          self.sip_degree, self.sip_max_inv_pix_error, self.sip_inv_degree,
-                          self.sip_npoints)
+                          sip_degree, self.sip_max_inv_pix_error, sip_inv_degree,
+                          self.sip_npoints, crpix=crpix)
```
jdavies-st, line 107: Could you specify these as kwargs, and not depend on arg order?  So:

```
to_fits_sip(max_pix_error=self.sip_max_pix_error, degree=sip_degree, ...
```

jdavies-st (approved):
Looks great!

mcara:
DO NOT MERGE: I need to squash commits and put back requirements for `gwcs` version

jdavies-st:
FYI @hbushouse , this PR will require re-okifying the regression tests that use SIP fitting.

jdavies-st:
Oh and the failure seen on the dev build against `gwcs` master on this PR:

https://travis-ci.com/github/spacetelescope/jwst/jobs/462584512#L893-L903

is due to the change in `gwcs.WCS.to_fits_sip()` in how it finds the center of the bounding box.  Before it was the nearest integer pixel, now it is the true center.

https://github.com/spacetelescope/gwcs/pull/338

This PR should fix that test.  Eitehr change the expected result in the test, or specify crpix in the call to `to_fits_sip()`.

mcara:
Merge at will

nden (approved):

Issue #5565, opened 2020-12-23
FITS cannot write NAN value possibly because SKYMATCH GLOBAL is NAN


stscijgbot-jp:
_Issue [JP-1830](https://jira.stsci.edu/browse/JP-1830) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

`calwebb_image3` is failing with a `ValueError: Floating point nan values are not allowed in FITS headers.` Full traceback is below.

Examining behavior pre-0.18.0, it is noted that the `skymatch` setting of the "Global" sky value correction changed from zero to NAN.

Full traceback:
$ strun cfgs/calwebb_image3.cfg jw00617-o083_20201223t045707_image3_001_asn.json
2020-12-23 07:57:09,604 - CRDS - INFO -  Fetching  /Users/eisenham/Downloads/crds-cache/jwst_op/mappings/jwst/jwst_nircam_gain_0010.rmap    2.3 K bytes  (1 / 3 files) (0 / 7.4 K bytes)
2020-12-23 07:57:09,727 - CRDS - INFO -  Fetching  /Users/eisenham/Downloads/crds-cache/jwst_op/mappings/jwst/jwst_nircam_0145.imap    4.5 K bytes  (2 / 3 files) (2.3 K / 7.4 K bytes)
2020-12-23 07:57:09,844 - CRDS - INFO -  Fetching  /Users/eisenham/Downloads/crds-cache/jwst_op/mappings/jwst/jwst_0645.pmap      580 bytes  (3 / 3 files) (6.8 K / 7.4 K bytes)
2020-12-23 07:57:11,309 - stpipe.Image3Pipeline - INFO - Image3Pipeline instance created.
2020-12-23 07:57:11,310 - stpipe.Image3Pipeline.assign_mtwcs - INFO - AssignMTWcsStep instance created.
2020-12-23 07:57:11,312 - stpipe.Image3Pipeline.tweakreg - INFO - TweakRegStep instance created.
2020-12-23 07:57:11,313 - stpipe.Image3Pipeline.skymatch - INFO - SkyMatchStep instance created.
2020-12-23 07:57:11,314 - stpipe.Image3Pipeline.outlier_detection - INFO - OutlierDetectionStep instance created.
2020-12-23 07:57:11,315 - stpipe.Image3Pipeline.resample - INFO - ResampleStep instance created.
[...]

PR #5567, opened 2020-12-24
Fix a crash in skymatch when using mode on 1 pixel

mcara:
Fixes https://github.com/spacetelescope/jwst/issues/5565 (SKYMATCH GLOBAL is NAN)


jdavies-st (approved):
Concise Fix!  Thanks Mihai.

PR #5569, opened 2020-12-26
Enable Github actions for CI

### LINKED ISSUE #5490: Configure Github Actions for jwst
_Issue [JP-1786](https://jira.stsci.edu/browse/JP-1786) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

Following a recommendation by the chief engineers we need to move the jwst tests from Travis to Github Actions.

Jenkins used to run on pull requests. Not any more?

jdavies-st:
This transitions us from Travis CI to Github Actions as our per-PR CI provider.  I tried to keep the testing matrix consistent with what we had on Travis.  Some changes:

- we now easily test on MacOS
- we no longer test using dev dependencies `requirements-dev.txt` as an allowed failure. On the flip side, we still run our dev dependencies tests (unit + regression tests) on Jenkins every night, so we should be able to keep up with any issues there.
- we've removed the job where we turn warnings into errors (it was an allowed failure).  I won't miss this at all, as it wasn't very useful.
- generally, our "allowed failure" jobs are no longer in the matrix.  This is one missing feature of Github Actions.

I've added `pip` and CRDS caching between builds.

Resolves #5490 / [JP-1786](https://jira.stsci.edu/browse/JP-1786)
Closes #4304


.travis.yml
```diff
@@ -1,7 +1,7 @@
 os: linux
 dist: xenial
 language: python
-python: 3.8.6
+python: 3.9.0
```
stscieisenhamer, line 4: Question: Is the intent to run "latest" python here, or "latest 3.9", or be specific to 3.9.0? If some form of latest is intended, should the specification be more general, such as "3.9"?

stscieisenhamer (approved):

.travis.yml
jdavies-st: For Travis, if you specify "3.9" it will pick a version, not necessarily the latest.  So if you want the latest, you need to specify it.  This is most important to get more recent versions of the packaged tools with that version of python, i.e. `setuptools` and `pip`.  They don't have a good way of indicating "install the latest 3.8 or 3.9".  But this is an idiosyncracy of Travis.  I'm completely deleting the Travis YAML file in this PR.

jdavies-st: Oh, and 3.9.0 is the only 3.9 currently.  So I was someone over-specifying it here, but better to do that than to leave it at "3.9" and have people think that when 3.9.1 comes out in a month or two, that it will somehow automatically be used.  It won't.

jdavies-st:
So it doesn't appear that one can allow the suite of tests to pass if one of the constituent checks has failed.

I'm fine removing the "Warnings treated as Exceptions" job, as I think that adds minimal useful information for us on a per-PR basis.

Would it be better to move the dev dependencies test into a separate `ci_cron.yml` that ran it as a daily cron job?  Would we lose track of it?

stscieisenhamer:
"Daily cron job"? Not sure what this means. This would be on github/travis? or locally in jenkins? Having CI on areas that are not high priority run elsewhere/not-every-PR sounds reasonable.

Currently, as it seems to be, "master" is, for all practical purposes, "development". As such "development CI" means "pass unit tests on minimum and target configurations". This does not mean, "pass full regression on all configs and prepare for future". So, dropping things like "warnings as exceptions" on a per-PR basis seems reasonable.

Now, to make a stable, the full test suite needs to pass, plus future consideration. However, that does not need to be a per-PR process. As such, if these tests are run cron-job, the the dev-to-stable process should ensure that all cron-job/hand-run tests are documented, so nothing is missed.

Was this a response to the question?

jdavies-st:
Thanks Jonathan.  Yeah, we currently already do a daily run of the full unit + regression tests using dev dependencies on Jenkins in the wee hours of the morning.  So maybe having a further run of them is not needed.  I had originally added these dev dependency tests to the per-PR CI so that they would be more visible while we were making major changes to `asdf` and `jwst.datamodels`->`stdatamodels`.  They have been useful in that regard.

I'm happy to drop the warnings as errors job, as it hasn't proven to be very useful.

kmacdonald-stsci (approved):

kmacdonald-stsci:
Everything looks good.

pllim (commented):
`twine>=3.3` has `twine check --strict ...` option that fails the check if there is warning. FYI.

Also if you want the ability to skip CI and cancel duplicate workflows, define a job like this in CI YAML and have all the other jobs `needs` it. 😉 

```yaml
  cancel_ci:
    name: Mandatory checks before CI
    runs-on: ubuntu-latest
    steps:
    - name: Failure means CI is skipped on purpose
      uses: pllim/action-skip-ci@main
      with:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
    # This should only run if we did not skip CI
    - name: Cancel Previous Runs
      uses: styfle/cancel-workflow-action@0.6.0
      with:
        access_token: ${{ secrets.GITHUB_TOKEN }}
```

jdavies-st:
Thanks @pllim.  So the above allows a commit to have `[ci skip]` in the commit message and that short-circuits CI?

pllim:
@jdavies-st , there are actually two things there:

1. `pllim/action-skip-ci` would result in a workflow similar to https://github.com/kali-maa/astropy/actions/runs/427044876 but it is important that you make every other jobs `needs` it in some way. I couldn't get it to just cancel the job due to restricted access to token from a fork.
2. `styfle/cancel-workflow-action` would cancel older runs if, say, you push multiple commits to the PR or a branch within short time frame. This behavior is similar to what CircleCI and Travis CI do (the latter didn't do so good at it but at least they supported it). But this is written by third-party, so I am thinking maybe it is safer to call it by the hash like `uses: styfle/cancel-workflow-action@ce17749`. I did look at the code and didn't find anything fishy (to the extend my TS skills allow) but there is no guarantee going forward.

Also, I saw that GitHub posted a new warning on `pull_request_target:` section and now I am confused as to whether it is really safer or we should just stick with `pull_request:`. I asked GitHub help desk and waiting to hear back.

jdavies-st:
Thanks!  Those look like nice features for the future.

Btw, the astropy Github actions YAML scripts were very helpful in figuring all this out.  And I haven't figured it all out yet.  😅

jdavies-st:
Oops.  I had removed the `--pyargs` job from the Travis matrix when I added tox in #5499.  Adding it back in this PR, and adding it to the Github actions matrix.

hbushouse (approved):
LGTM

pllim:
To close the loop on `pull_request_target`, I was wrong. It shouldn't be used in CI. See https://securitylab.github.com/research/github-actions-preventing-pwn-requests . Sorry for the confusion!

PR #5570, opened 2020-12-27
JP-1540: Fix bugs in wavecorr

### LINKED ISSUE #5080: NIRSpec wavecorr values appear to be incorrect
_Issue [JP-1540](https://jira.stsci.edu/browse/JP-1540) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The magnitude of the wavelength correction applied in extract_2d to NIRSpec MOS slitlets containing off-center point sources appears to be too small. Experiments done by moving a test source from one side (in the x or dispersion direction) of a slitlet to the other result in no change at all to the values in the output WAVELENGTH array of the cal file.

nden:
Closes #5080

This resolves part of JP-1540, fixing the current implementation.

### DIFF
--- a/jwst/wavecorr/wavecorr.py
+++ b/jwst/wavecorr/wavecorr.py
@@ -102,6 +102,8 @@ def apply_zero_point_correction(slit, reffile):
                                                  aperture_name, dispersion)
     ## TODO: set a DQ flag to a TBD value for pixels where dq_lam == 0.
     ## The only purpose of dq_lam is to set that flag.
+    
+    # Wavelength is in um, the correction is computed in meters.
     slit.wavelength = lam - corr * 10 ** 6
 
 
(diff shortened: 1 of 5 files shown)


jwst/wavecorr/wavecorr.py
```diff
@@ -102,7 +102,7 @@ def apply_zero_point_correction(slit, reffile):
                                                  aperture_name, dispersion)
     ## TODO: set a DQ flag to a TBD value for pixels where dq_lam == 0.
     ## The only purpose of dq_lam is to set that flag.
-    slit.wavelength = lam - corr
+    slit.wavelength = lam - corr * 10 ** 6
```
hbushouse, line 105: A comment here would be useful, to indicate the purpose of the scaling (e.g. converting meters to microns)

hbushouse (approved):
Just one minor comment. The rest all looks good.

PR #5571, opened 2020-12-30
Add new extract_1d regtest for MIRI MRS

hbushouse:
Added a new regtest module to run just the `extract_1d` step on a MIRI MRS data cube, with the source type set to POINT, in order to test the point-source branches within `extract_1d`. The data is from a set of simulations created by the MIRI team, which actually contains an extended source, but I just reset the source type value to say POINT.


jdavies-st (approved):

jemorrison (commented):
looks good

PR #5574, opened 2020-12-30
JP-1834: remove defaults from cfg files

### LINKED ISSUE #5573: source_catalog step crashing for some FGS images
_Issue [JP-1834](https://jira.stsci.edu/browse/JP-1834) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

[Mike Swam](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mswam) reports a handful of the calwebb_image3 failures all involving FGS imaging, due to the source_catalog step crashing with a traceback of:
```java
2020-12-30 14:28:54,198 - stpipe.source_catalog - INFO - Instrument: FGS
2020-12-30 14:28:54,199 - stpipe.source_catalog - INFO - Detector: GUIDER2
2020-12-30 14:28:54,199 - stpipe.source_catalog - INFO - Subarray: FULL
2020-12-30 14:28:54,257 - stpipe.source_catalog - DEBUG - No ASDF information found. Cannot skip updating from FITS headers.
2020-12-30 14:28:54,330 - stpipe.source_catalog - INFO - AB to Vega magnitude offset 1.22852
2020-12-30 14:29:00,181 - stpipe.source_catalog - INFO - Detected 2036 sources
----------------------------------------------------------------------
ERROR RUNNING STEP 'SourceCatalogStep':
    positions must not contain any non-finite (e.g., NaN or inf)
    positions
----------------------------------------------------------------------
  File "/Users/bushouse/JWST/softw/jwst/scripts/strun", line 26, in <module>
  File "/Users/bushouse/JWST/softw/jwst/jwst/stpipe/step.py", line 164, in from_cmdline
  File "/Users/bushouse/JWST/softw/jwst/jwst/stpipe/cmdline.py", line 346, in step_from_cmdline
    step.run(*positional)
  File "/Users/bushouse/JWST/softw/jwst/jwst/stpipe/step.py", line 400, in run
  File "/Users/bushouse/JWST/softw/jwst/jwst/source_catalog/source_catalog_step.py", line 113, in process
    catalog = catobj.catalog
  File "/Users/bushouse/anaconda3/envs/jwstdev/lib/python3.9/site-packages/astropy/utils/decorators.py", line 758, in _get_
    val = self.fget(obj)
  File "/Users/bushouse/JWST/softw/jwst/jwst/source_catalog/source_catalog.py", line 1341, in catalog
    self.set_aperture_properties()
  File "/Users/bushouse/JWST/softw/jwst/jwst/source_catalog/source_catalog.py", line 928, in set_aperture_properties
    apertures = [CircularAperture(self.xypos, radius) for radius in
  File "/Users/bushouse/JWST/softw/jwst/jwst/source_catalog/source_catalog.py", line 928, in <listcomp>
    apertures = [CircularAperture(self.xypos, radius) for radius in
  File "/Users/bushouse/anaconda3/envs/jwstdev/lib/python3.9/site-packages/photutils/aperture/circle.py", line 148, in _init_
    self.positions = positions
  File "/Users/bushouse/anaconda3/envs/jwstdev/lib/python3.9/site-packages/photutils/aperture/attributes.py", line 62, in _set_
    self._validate(value)
  File "/Users/bushouse/anaconda3/envs/jwstdev/lib/python3.9/site-packages/photutils/aperture/attributes.py", line 80, in _validate
    raise ValueError('{} must not contain any non-finite (e.g., NaN '
ValueError: positions must not contain any non-finite (e.g., NaN or inf) positions
```

hbushouse:
Remove the non-default param values from the `source_catalog.cfg` file, allowing the step to always run with defaults. This fixes the errors reported in [JP-1834](https://jira.stsci.edu/browse/JP-1834).

At the same time, I've also gone through *all* cfg files and removed all default param values, including pointers to cfg files in some of the pipeline module cfg's, leaving only non-default param settings that are in fact needed for special cases.

Fixes #5573 / [JP-1834](https://jira.stsci.edu/browse/JP-1834)

### DIFF
--- a/jwst/pipeline/calwebb_ami3.cfg
+++ b/jwst/pipeline/calwebb_ami3.cfg
@@ -4,9 +4,6 @@ save_averages = True
 
     [steps]
       [[ami_analyze]]
-        config_file = ami_analyze.cfg
       [[ami_average]]
-        config_file = ami_average.cfg
       [[ami_normalize]]
-        config_file = ami_normalize.cfg
 
--- a/jwst/pipeline/calwebb_detector1.cfg
+++ b/jwst/pipeline/calwebb_detector1.cfg
@@ -4,34 +4,19 @@ save_calibrated_ramp = False
 
     [steps]
       [[group_scale]]
-        config_file = group_scale.cfg
       [[dq_init]]
-        config_file = dq_init.cfg
       [[saturation]]
-        config_file = saturation.cfg
       [[ipc]]
         skip = True
       [[superbias]]
-        config_file = superbias.cfg
       [[refpix]]
-        config_file = refpix.cfg
       [[rscd]]
-        config_file = rscd.cfg
       [[firstframe]]
-        config_file = firstframe.cfg
       [[lastframe]]
-        config_file = lastframe.cfg
       [[reset]]
-        config_file = reset.cfg
       [[linearity]]
-        config_file = linearity.cfg
       [[dark_current]]
-        config_file = dark_current.cfg
       [[persistence]]
-        config_file = persistence.cfg
       [[jump]]
-        config_file = jump.cfg
       [[ramp_fit]]
-        config_file = ramp_fit.cfg
       [[gain_scale]]
-        config_file = gain_scale.cfg
--- a/jwst/pipeline/calwebb_tso3.cfg
+++ b/jwst/pipeline/calwebb_tso3.cfg
@@ -8,6 +8,4 @@ scale_detection = False
         config_file = outlier_detection_tso.cfg
       [[tso_photometry]]
       [[extract_1d]]
-        config_file = extract_1d.cfg
       [[white_light]]
-        config_file = white_light.cfg
--- a/jwst/pipeline/calwebb_tso3.cfg
+++ b/jwst/pipeline/calwebb_tso3.cfg
@@ -1,8 +1,6 @@
 name = "Tso3Pipeline"
 class = "jwst.pipeline.Tso3Pipeline"
 
-scale_detection = False
-
     [steps]
       [[outlier_detection]]
         config_file = outlier_detection_tso.cfg
--- a/jwst/pipeline/cube_build.cfg
+++ b/jwst/pipeline/cube_build.cfg
@@ -1,16 +1,3 @@
 name = "cube_build"
 class = "jwst.cube_build.CubeBuildStep"
-channel = "all"
-band = "all"
-filter = "all"
-grating = "all"
-scale1 = 0.0
-scale2 = 0.0
-scalew = 0.0
-weighting = "emsm"
-coord_system = 'skyalign'
-rois = 0.0
-roiw = 0.0
 weight_power = 2.0
-single = false
-output_type = "band"
--- a/jwst/pipeline/outlier_detection_stack.cfg
+++ b/jwst/pipeline/outlier_detection_stack.cfg
@@ -1,17 +1,4 @@
 name = "outlier_detection_stack" 
 class = "jwst.outlier_detection_stack.OutlierDetectionStackStep"
 
-weight_type = 'exptime'
-pixfrac = 1.0
-kernel = 'square'
-fillval = 'INDEF'
-nlow = 0
-nhigh = 0
-maskpt = 0.7
-grow = 1
-snr = '4.0 3.0'
-scale = '0.5 0.4'
-backg = 0.0        
-save_intermediate_results = False
 resample_data = False
-good_bits = '~DO_NOT_USE'
--- a/jwst/pipeline/outlier_detection_stack.cfg
+++ b/jwst/pipeline/outlier_detection_stack.cfg
@@ -1,4 +1,2 @@
 name = "outlier_detection_stack" 
 class = "jwst.outlier_detection_stack.OutlierDetectionStackStep"
-
-resample_data = False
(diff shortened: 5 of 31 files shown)


jwst/pipeline/calwebb_ami3.cfg
```diff
@@ -4,9 +4,6 @@ save_averages = True
 
     [steps]
       [[ami_analyze]]
-        config_file = ami_analyze.cfg
       [[ami_average]]
-        config_file = ami_average.cfg
       [[ami_normalize]]
-        config_file = ami_normalize.cfg
```
jdavies-st, line 11: Do the sections need to stay in the config?  I suppose that's useful for a future user to add things to those section?  But they don't do anything on their own, right?

jwst/pipeline/calwebb_detector1.cfg
```diff
@@ -4,34 +4,19 @@ save_calibrated_ramp = False
 
     [steps]
       [[group_scale]]
-        config_file = group_scale.cfg
       [[dq_init]]
-        config_file = dq_init.cfg
       [[saturation]]
-        config_file = saturation.cfg
       [[ipc]]
         skip = True
```
jdavies-st, line 10: Is there a reason we even have `IPCStep` in the pipeline if it is skipped?

jwst/pipeline/calwebb_tso3.cfg
```diff
@@ -8,6 +8,4 @@ scale_detection = False
         config_file = outlier_detection_tso.cfg
```
jdavies-st, line 8: `scale_detection = False` above can probably be removed, as it is set in the pipeline `spec`

jwst/pipeline/cube_build.cfg
```diff
@@ -1,16 +1,3 @@
...
-filter = "all"
-grating = "all"
-scale1 = 0.0
-scale2 = 0.0
-scalew = 0.0
-weighting = "emsm"
-coord_system = 'skyalign'
-rois = 0.0
-roiw = 0.0
 weight_power = 2.0
```
jdavies-st, line 3: A reason to keep `weight_power` here?  This is different from the default in the step.  Perhaps the default in `CubeBuildStep.spec` should be changed and this removed?

jwst/pipeline/outlier_detection_stack.cfg
```diff
@@ -1,17 +1,4 @@
...
-fillval = 'INDEF'
-nlow = 0
-nhigh = 0
-maskpt = 0.7
-grow = 1
-snr = '4.0 3.0'
-scale = '0.5 0.4'
-backg = 0.0        
-save_intermediate_results = False
 resample_data = False
```
jdavies-st, line 4: `resample_data = False` in the step spec, so you can probably remove it here too.

stscieisenhamer (commented):
Agree with other comments. Linked this issue to JP-1030, since all cfgs are involved.

jwst/pipeline/calwebb_ami3.cfg
hbushouse: You're correct that they have no effect on their own. My thinking was to just leave them there to make it easier when folks want to add param settings for a given step. Just serves as a template/skeleton for them to work with.

jwst/pipeline/calwebb_detector1.cfg
hbushouse: We left it in the code so that it's at least available should anyone want to run it on their own. There are ref files for some instruments. But for now it's skipped by default in normal Ops processing.

jwst/pipeline/calwebb_tso3.cfg
hbushouse: Whoops, missed that one. Good catch!  Will remove.

jwst/pipeline/cube_build.cfg
hbushouse: I left it in for now because it *is* different than the default and couldn't tell if that was intentional or not. The default in the step spec is zero. @jemorrison or @drlaw1558 can either of you tell us which is the preferred default (0 or 2)?

jwst/pipeline/outlier_detection_stack.cfg
hbushouse: Ah, right. The default is `True` in the regular `OutlierDetectionStep`, so I thought I'd leave it in. But I see that the defaults are different in `OutlierDetectionStackStep`.

jwst/pipeline/calwebb_tso3.cfg
hbushouse: Done.

jwst/pipeline/outlier_detection_stack.cfg
hbushouse: Done

stscieisenhamer (approved):

jwst/pipeline/cube_build.cfg
drlaw1558: Hm, 2 should be the default, as this is 1/r^2 MSM weighting.  However, the difference is less important at the moment since EMSM is the current default weighting scheme so the MSM parameters are less likely to be used.

hbushouse: Thanks for the confirmation. So in this case the value in the cfg file is correct, but the default in the step code is not. So we should fix the default value in the step code and then remove this one remaining param from the cfg file.

PR #5577, opened 2020-12-31
TST: Allow skip and cancel CI

pllim:
Allow skipping and cancelling CI in GitHub Actions.

The skip CI part is demonstrated in this PR. I have `[skip ci]` in the commit message, so the skip CI job would fail and prevent the rest from running.

cc @jdavies-st @eslavich 

Close #5591

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,14 +23,19 @@ jobs:
   cancel_ci:
     name: Mandatory checks before CI
     runs-on: ubuntu-latest
+    outputs:
+      run_next: ${{ steps.skip_ci_step.outputs.run_next }}
     steps:
     - name: Failure means CI is skipped on purpose
       uses: pllim/action-skip-ci@main
+      id: skip_ci_step
       with:
+        NO_FAIL: true
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     # This should only run if we did not skip CI
     - name: Cancel Previous Runs
       uses: styfle/cancel-workflow-action@ce17749
+      if: steps.skip_ci_step.outputs.run_next == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -20,9 +20,32 @@ env:
...
 
 jobs:
+  cancel_ci:
+    name: Mandatory checks before CI
+    runs-on: ubuntu-latest
+    outputs:
+      run_next: ${{ steps.skip_ci_step.outputs.run_next }}
+    steps:
+    - name: Set output for skip CI
+      uses: pllim/action-skip-ci@main
```
jdavies-st, line 30: Would be nice if this `action-skip-ci` action was moved to either astropy or spacetelescope, depending on scope.

pllim: Likely `astropy` but probably won't be resolved for 1-2 weeks until I get okay from Astropy people. This is not the only package using it now, so when that happens, I plan to batch PR the change. But if you are not comfortable, you are welcome to wait as well.

pllim: Update: `astropy` is using it now, so I think it is just a matter of time that this Action moves into `astropy` org, but it cannot happen before the next infrastructure meeting scheduled for 2021-01-25.

jdavies-st: Thanks @pllim!  How about we merge this then, and then when it moves you can just add us to your batch PR?

pllim: Yes, that is my plan. Thanks for the merge. Good to see my random pet project getting some use. 😅

jdavies-st: Here here to random pet projects!  🎉

PR #5579, opened 2020-12-31
TST: Add weekly cron

pllim:
To catch upstream changes. I guess you might find daily too daunting, so I have it run weekly. If you only want subset of the jobs to run in cron, we can either add a `if` or have a separate YAML.

cc @jdavies-st


jdavies-st (approved):
LGTM.  Thanks!

PR #5580, opened 2020-12-31
TST: Test on push to release branches

pllim:
You might want to also run the CI on push to release branches. In older branches without a workflow YAML, nothing will happen.

Depends on your backport/release workflows though.

cc @jdavies-st

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
-      - 0.*
+      - '*.x'
     tags:
       - '*'
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - 0.*
```
jdavies-st, line 7: That's a good idea.  I guess Travis does this automatically.  Our release branches are always have names like `A.B.x`, so perhaps a `"*.x"` would be better?

jdavies-st (commented):
We don't really maintain old release branches for `jwst` right now, but I think we might in the future once 1.0.0 is out.

jdavies-st (approved):
Looks great!  Thanks.

PR #5582, opened 2020-12-31
TST: Pin twine for --strict

pllim:
I see that you use the new `--strict` option but it is only available for `twine>=3.3`. This pins `twine` in the setup to prevent developers having older twine in their tox env encountering error. Should be uncontroversial enough?

cc @jdavies-st

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -26,7 +26,7 @@ description =
 extras = test
 
 # Pass through the following environment variables which may be needed for the CI
-passenv = 
+passenv =
     TOXENV
     CI
     CODECOV_*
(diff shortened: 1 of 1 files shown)


jdavies-st (approved):
Thanks for correcting my oversight!

jdavies-st:
Btw, what are your thoughts on the `pep517.check` we do?  Would it be better to use `pip wheel` for this, as this is what is used to install the package?

Currently, our auto-publishing to PyPI action uses `pep517.build`, but it only builds and delivers the sdist.

pllim:
I don't distribute wheels myself (maybe I should but that is a different story), so I don't have any useful thoughts to share. `astropy` does, so perhaps you can inquire over at Astropy Slack (`#infrastructure` or `#releasing`). In addition, @jhunkeler might have thoughts since he dealt with a lot of releases in his SSB days.

PR #5583, opened 2020-12-31
TST: Add CodeQL check

pllim:
Adds a check using https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/about-code-scanning#about-codeql . I don't see any C-ext or Cython, so I only enabled Python check.

* Includes cancel/skip CI workflow similar to #5577 , as the check can take a while, so you want to be able to skip it in PR.
* Depending on how #5580 go, might also want to change the `push` branches to match.

Not sure how useful this really is compared to `bandit`, but doesn't hurt? 🤷 

cc @jdavies-st

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,94 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+
+jobs:
+  cancel_ci:
+    name: Failure means CI cancelled on purpose
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


jdavies-st:
Is this check more robust than `bandit`?  Is there a strong reason we are currently using `bandit` at ST?

I like the fact that I can run `bandit` locally.  If this CodeQL is just a Github action, then I have to push regularly (and deal with the rest of the CI matrix running) every time I want to iterate on some security problem.

pllim:
Re: Robustness -- Not sure. I originally heard about `bandit` from @nden, so maybe she could clarify?

Re: Running locally -- Seems possible but the setup is more work than `bandit`, see https://codeql.github.com/docs/codeql-for-visual-studio-code/analyzing-your-projects/

Re: Running CI matrix to fix security bug -- Yeah, the inability to easily skip CI in Actions is kinda painful sometimes. CodeQL here is a GitHub Action (https://github.com/github/codeql-action). I wonder if I can modify [action-skip-ci](https://github.com/pllim/action-skip-ci) to take custom directives; that way, you can use different commit message directive to skip different parts of Actions, but let's come back to this if you decide you do want CodeQL in your pipeline.

jdavies-st:
Closing this PR, as we currently are going to continue to use `bandit` as is standard practice at ST DMD.

We can revisit CodeQL in the future and resurrect this PR if we feel we need to have something more than `bandit`.

pllim:
For future reference, if you change your mind, I added ability to customize the skip directive. You can see how it is used over at https://github.com/spacetelescope/synphot_refactor/blob/6346230fbd3de442828ea71b607b44f022998900/.github/workflows/codeql-analysis.yml#L29

PR #5584, opened 2020-12-31
JP-1407 Update MIRI LRS-Fixedslit along-slit-nod backgrounds strategies

### LINKED ISSUE #4814: Updates to background subtraction strategies for MIRI LRS
Issue [JP-1407](https://jira.stsci.edu/browse/JP-1407) was created by Sarah Kendrew:

MIRI LRS uses mandatory dithering for non-TSO observations (i.e. with LRS slit). This results in file sets being created that the JWST pipeline is not currently able to process correctly in the background subtraction steps. 

LRS (Kendrew) will create an overview of the different use cases for LRS slit observations, with recommended background subtraction approach, for input from CalWG and DMSWG.

stscieisenhamer:
Resolves [JP-1407](https://jira.stsci.edu/browse/JP-1407)
Resolves #4814 

Other changes:

- Resolves issue with duplicate Level 3 associations. Original algorithm only checked for duplicate product names. However, for Level 3, product names are often different, but membership is the same. 
- Fixes tests where the value of PATTTYPE being used was `along_slit_nod` and not the correct `along-slit-nod`.


jwst/associations/lib/dms_base.py
```diff
@@ -74,23 +74,23 @@
...
-    'mir_4qpm',
     'nis_ami',
     'nis_image',
-    'nrc_image',
     'nrc_coron',
-    'nrc_tsimage',
+    'nrc_image',
     'nrs_mimf',
-    'fgs_image',
+    'nrc_tsimage',
```
hbushouse, line 93: If you're trying to achieve strict alphabetical ordering of these elements, `nrc_tsimage` needs to be above `nrs_mimf` in the list ;-)

jwst/associations/lib/rules_level2b.py
```diff
@@ -252,7 +252,7 @@ def __init__(self, *args, **kwargs):
                     DMSAttrConstraint(
                         name='patttype',
                         sources=['patttype'],
-                        value=['2-point-nod|4-point-nod|along-slit-nod'],
+                        value=['2-point-nod|4-point-nod'],
```
hbushouse, line 255: So I'm trying to wrap my head around the effect of this double negative: `along-slit-nod` used to be included in a `notany` constraint, which means we were *not* creating `spec2` ASN's for exposures with `along-slit-nod`? And now that that pattern type has been removed from the `notany` constraint, we *are* creating `spec2` ASN's for them?

But if that's the case, why aren't we creating `spec2` ASN's for dithered exposures using `2-point-nod` and `4-point-nod`? I'm not sure why those pattern types are treating differently, since they all should allow for potential image-from-image background subtraction, if the target is a point source.

ddavis-stsci (approved):
Looks good to me.

jwst/associations/lib/rules_level2b.py
stscieisenhamer: First: Correct, this "notany" means that none of the contained constraints can match. So, if patttype had any of the given values, the match would fail.

The removal of along-slit-nod allows this association to create the dedicated background associations for along-slit-nod, if and when they exist, as per the request of the issue.

The other patttypes, which are NIRSpec-specific, are handled by [Asn_Lv2NRSIFUNod](https://github.com/spacetelescope/jwst/blob/93b8d5dacbf15fb8b74c482b32df387e8d741042/jwst/associations/lib/rules_level2b.py#L783). Hence they are not allowed in this association.

hbushouse (approved):

hbushouse:
Are we OK to merge?

PR #5585, opened 2021-01-02
JP-1743: Update FILETYPE keyword in standard products

### LINKED ISSUE #1699: Pipeline modules need to update FILETYPE keyword
SDP populates the FILETYPE keyword with 'uncalibrated' when it creates level-2b (_uncal.fits) files. That value is currently retained throughout all levels of cal pipeline processing. Each pipeline module should update the value in its output product(s) at the end of its processing, with a list of appropriate values that are still TBD.

hbushouse:
Updated all of the pipeline modules to update the FILETYPE value for at least the main product that they produce, as well as intermediate products that are saved by the pipeline module itself. Several of the major level 3 steps, such as `outlier_detection`, `resample`, and `extract_1d`, have their products created by `stpipe` when the step finishes and returns control to the pipeline module. Hence for those the step module itself has to do the updating, because the product has already been saved by the time control is returned to the pipeline module.

This only updates FILETYPE for standard products that are routinely created and archived during Ops processing. It would be far too messy and complicated to have every step make an update. If anyone has suggestions for alternate FILETYPE values for any of the products, feel free to suggest.

Fixes #1699 / [JP-1743](https://jira.stsci.edu/browse/JP-1743)

### DIFF
--- a/jwst/combine_1d/combine_1d_step.py
+++ b/jwst/combine_1d/combine_1d_step.py
@@ -24,5 +24,6 @@ class Combine1dStep(Step):
         with datamodels.open(input_file) as input_model:
             result = combine1d.combine_1d_spectra(input_model,
                                                   self.exptime_key)
+        result.meta.filetype = '1d combined spectrum'
 
         return result
--- a/jwst/cube_build/cube_build_step.py
+++ b/jwst/cube_build/cube_build_step.py
@@ -354,7 +354,8 @@ class CubeBuildStep (Step):
         for cube in cube_container:
             footprint = cube.meta.wcs.footprint(axis_type="spatial")
             update_s_region_keyword(cube, footprint)
-        if status_cube ==1:
+            cube.meta.filetype = '3d ifu cube'
+        if status_cube == 1:
             self.skip = True
 
         return cube_container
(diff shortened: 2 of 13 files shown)


jdavies-st (approved):
Are these FILETYPE strings defined anywhere?  Looks a bit like the wild west.

Would it be good to have an enum in the schema for these?

Or maybe that's too much and none of this matters?  We can't just remove this keyword?

hbushouse:
> Are these FILETYPE strings defined anywhere? Looks a bit like the wild west.

No, not defined anywhere. I just made them up.

> Would it be good to have an enum in the schema for these?

Couldn't hurt, but not really necessary.

> Or maybe that's too much and none of this matters? We can't just remove this keyword?

Just discussed removal of the keyword in our team-coffee sprint planning and we're going to file a JWSTKD ticket to have that group discuss the possibility of removing the keyword (assuming of course that no other subsystem is using it for anything).

jdavies-st:
Is there a SDP database that associates the filename suffix, i.e. `_uncal` or `_crf` with some text in a database?  If so, that might be the text that it would be worth using if we decide we need to keep this keyword.

I don't think those suffix endings are in the keyword dictionary anywhere, but have to be in some other database.  Maybe MAST?

jdavies-st:
In our own documentation, we have descriptive phrases for these suffixes.

https://jwst-pipeline.readthedocs.io/en/latest/jwst/data_products/science_products.html

PR #5587, opened 2021-01-04
Update cube_build weight_power default value

hbushouse:
This is a follow-on to https://github.com/spacetelescope/jwst/pull/5574 which fixed issues related to param settings in cfg files. @drlaw1558 reported that the default value for `weight_power` should be 2, which it was set to in the `cube_build.cfg` file, but in the step spec it was set to zero. So if the step is ever run without the .cfg file you'd get the wrong value. This updates the default value in the step spec to 2.0 and removes the param from the .cfg file.


jemorrison (approved):

jemorrison:
Just to be clear. With weighting set to default 'emsm'  'weight_power' is not used. 
Also the real default value for 'weight_power' is held in the cubepars reference file. I mention that because maybe it is confusing the way it is set up now. If the user wants to override the value in that is in the cubepars reference file then they can  use this option. I set it the default values in the spec before there was a cubepars reference file. But now I wonder if it adds confusion since this step works different than other steps.

drlaw1558 (approved):

PR #5588, opened 2021-01-04
JP-1622: Allow one combine1d extension per order in output

### LINKED ISSUE #5204: WFSS combine_1d step does not output combined spectrum for both spectral orders with F090W filter
_Issue [JP-1622](https://jira.stsci.edu/browse/JP-1622) was created on JIRA by [Swara Ravindranath](https://jira.stsci.edu/secure/ViewProfile.jspa?name=swara):_

The Calwebb_spec2 and Calwebb_spec3 processing correctly outputs the 1st order and 2nd order spectra for grism combinations that use the F090W filter. The 1st and 2nd order are placed in different extensions as expected. So the 2D cutouts and 1D extracted outputs are as expected.

However, in the final combine_1d output of the Calwebb_spec3 processing there is only one extension that has the spectrum. It appears that the first four x_1d files are being combined and the output is not correct. Also, there is no separate output for 1st order and 2nd order spectra. I have attached 2 figures here that show the information on the header extensions, because showing the figure with spectrum does not convey the issue.

tapastro:
Resolves [JP-1622](https://jira.stsci.edu/browse/JP-1622)
Resolves #5204 

After modeling the new schema for combine1d output to follow multispec.schema, I pulled the spec_table specification out of both schema systems to avoid redundant definitions and make future edits more straightforward.

PS: Unsure of what milestone, if any, to assign.


jwst/datamodels/schemas/multispec.schema.yaml
```diff
@@ -14,27 +14,7 @@ allOf:
...
-              datatype: float64
-            - name: DQ
-              datatype: uint32
-            - name: BACKGROUND
-              datatype: float64
-            - name: BERROR
-              datatype: float64
-            - name: NPIXELS
-              datatype: float64
+            $ref: spectable.schema
```
hbushouse, line 17: Nice consolidation of schema elements!

hbushouse (commented):
Looks good overall. Just a few comment/questions to consider before finalizing.

hbushouse:
I just did a spot check of a few x1d products for WFSS and NIRISS SOSS modes and they do in fact have several useful slit/source-related keywords in the EXTRACT1D extension header, such as SLTNAME, SLITID, SOURCEID, SRCXPOS, SRCYPOS, SLIT_RA, SLIT_DEC, SPORDER, and DISPAXIS. I believe all of those are defined in `multispec.schema`.

Unfortunately it would not be trivial to extract those keyword definitions to a separate schema that could be imported by several others, because the `fits_hdu` values need to be different depending on which data model they're being used in. For example we need `fits_hdu = EXTRACT1D` for models used by x1d products, while we need `fits_hdu = COMBINE1D` for c1d products from the `combine_1d` step.

tapastro:
> I just did a spot check of a few x1d products for WFSS and NIRISS SOSS modes and they do in fact have several useful slit/source-related keywords in the EXTRACT1D extension header, such as SLTNAME, SLITID, SOURCEID, SRCXPOS, SRCYPOS, SLIT_RA, SLIT_DEC, SPORDER, and DISPAXIS. I believe all of those are defined in `multispec.schema`.
> 
> Unfortunately it would not be trivial to extract those keyword definitions to a separate schema that could be imported by several others, because the `fits_hdu` values need to be different depending on which data model they're being used in. For example we need `fits_hdu = EXTRACT1D` for models used by x1d products, while we need `fits_hdu = COMBINE1D` for c1d products from the `combine_1d` step.
> 
> Or perhaps it's time to abandon the `EXTRACT1D` and `COMBINE1D` EXTNAME's and just have them all use the generic `SCI` extension name, as is the case for all other data products. That would solve the `fits_hdu` problem.

For both the EXTRACT1D and WCS information, I just did minimal effort and tried to keep existing functionality without adding much. The originial combine1d output files are quite bare in their detail, and I felt that adding the SPORDER was necessary given the change, but I haven't given much thought to the other keywords. Besides modifying EXTNAME, we could add the most relevant to the combinedmultispec schema properties and populate when available?

hbushouse:
I agree that if the changes incorporated here at least fix the problem that was originally reported, we can tackle any other desired changes to the output product in another ticket/PR.

hbushouse (approved):
Just needs a change log entry and it'll be ready to merge, after which we'll give it a shakedown with the regtests.

PR #5590, opened 2021-01-05
Remove unused LinearPipeline class

eslavich:
This removes the LinearPipeline class, which is a simplified base class for pipelines whose steps can feed output into one another without requiring additional Python glue code.  In practice all pipelines have required at least some glue, so the simplified base class was never used.  I'd like to remove it so that we don't have to maintain it in the new stpipe package.

I'll create a ticket to remove the pars-linearpipeline reference type from crds.


jdavies-st (approved):
🎉

hbushouse (approved):
LGTM. Just needs a change log entry before merging.

PR #5591, opened 2021-01-05
Add a skip ci feature to the CI workflow

eslavich:
This replicates the Travis feature where `[skip ci]` or `[ci skip]` in the commit message causes tests to be skipped.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,29 @@ env:
   CRDS_CLIENT_RETRY_DELAY_SECONDS: 20
 
 jobs:
+  output-commit-message:
+    name: Output commit message
+    runs-on: ubuntu-latest
+    outputs:
+      message: ${{ steps.get-commit-message.outputs.message }}
+    steps:
+      - name: Checkout code (pull request)
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 1
+      - name: Checkout code (other events)
+        if: github.event_name != 'pull_request'
+        uses: actions/checkout@v2
+        with:
+          fetch_depth: 1
+      - id: get-commit-message
+        name: Get commit message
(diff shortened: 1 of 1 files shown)


pllim:
Duplicate functionality as #5577 but different approach.

pllim:
p.s. Is the addition of a file named `foo` intentional?

eslavich:
> p.s. Is the addition of a file named foo intentional?

Yea, that was just the change I made to show the behavior when `[skip ci]` is present.  I'll remove that commit if we end up deciding to merge this.

eslavich:
I attempted to make this version a separate action, but evidently we're not permitted to nest an action (like checkout) within another action.  I think @pllim's javascript action is the way to go, though I'm in favor of setting an output and checking it with an `if:` so that we can avoid a red X failure when the CI is skipped.

pllim:
Yeah, composite action isn't a thing yet though it is widely requested. You can wait, but I am not sure for how long.

> I'm in favor of setting an output and checking it

Looks like there is an option to set output in https://github.com/actions/toolkit/tree/main/packages/core . But I have to think about this. Maybe I'll make it an option whether to fail or just set the output, but that also complicates the documentation and logic.

pllim:
For now though, if you can endure a red X, my action already works, as illustrated in the other PR, in case you need it immediately.

jdavies-st:
Yeah, I don't think we want to endure red Xs for skipped CI.  For `jwst`, the check should pass if CI is skipped.

Same is true for allowed failures, though that's a separate issue.

The way all this worked on Travis CI was very useful and we'd like to see that identical functionality in Github actions as well.

I would very much be in favor of the best of both these PRs getting put together.

pllim:
That makes sense. I am looking into it.

pllim:
I updated #5577 with your requested feature. Took me a while to get the output syntax right, but it seems to work now.

eslavich:
Ok thanks a lot!  Closing this one.

pllim:
Here is the issue about composite action actions/runner#646

PR #5592, opened 2021-01-07
Remove unused private method from outlier detection steps

eslavich:
I noticed that these methods seem to be unused (the string `_build_reffile_container` does not occur anywhere else in the repo).  Removing them will save me the trouble of updating them for changes to the `reference_uri_to_cache_path` method.


jdavies-st (approved):
Looks good.  It used to be that `gain` and `readnoise` reffiles were required for this step.

```python
```

but the last ones were removed in #1758.

stsci-hack (approved):
This simply removes vestigial code, which is always a good thing.

PR #5593, opened 2021-01-07
Fix internal_cal cubes for NIRSpec Lamp exposures 

jemorrison:
Need a better fix once we figure out how to determine what the units of the wavelength are. Works for now


hbushouse (approved):

stscirij (approved):
Looks perfectly reasonable to me

PR #5594, opened 2021-01-07
Fix bug in latest extract_1d update

hbushouse:
Fixes a bug introduced in the latest `extract_1d` updates. There's a certain corner case (WFSS observations) where the data coming into `run_extract1d` are in a ModelContainer, as opposed to a simple data model, hence trying to access `model.meta.exposure.type` throws an AttributeError, because it needs to be `model[0].meta.exposure.type`.


jdavies-st (approved):

jemorrison (commented):
good catch.

Issue #5562, opened 2020-12-21
Enable cube_build in the spec2pipeline for NRS lamp exposures


stscijgbot-jp:
_Issue [JP-1828](https://jira.stsci.edu/browse/JP-1828) was created on JIRA by [Robert Jedrzejewski](https://jira.stsci.edu/secure/ViewProfile.jspa?name=rij):_

The pars-spec2pipeline reference file needs to be updated to enable the cube-build step and set the coord_system parameter to internal_cal

Also the cubepar rmap needs to be updated to select the reference file based on EXP_TYPE and OPMODE/meta.instrument.lamp_mode and a new cubepar reference file delivered.

stscijgbot-jp:
_Comment by [Robert Jedrzejewski](https://jira.stsci.edu/secure/ViewProfile.jspa?name=rij) on [JIRA](https://jira.stsci.edu/browse/JP-1828?focusedCommentId=501499#comment-501499):_

[James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol) is there any reason to use a different cubepar reference file or different cube_build parameters for NIRSpec lamp data as compared to science data, apart from setting coord_system to 'internal_cal'?  Also, should the cubepar reference file select on the LAMP keyword (i.e. do you think flat lamp exposures should use a different reference file from line lamp exposures?  I know flat lamp exposures aren't run through cube_build in the default spec2pipeline, but in principle you might want to run it offline)

stscijgbot-jp:
_Comment by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol) on [JIRA](https://jira.stsci.edu/browse/JP-1828?focusedCommentId=502363#comment-502363):_

[...]

PR #5599, opened 2021-01-08
fix wavelength when building internal_cal for NIRSpec lamp exposures

jemorrison:
Supports #5562 / [JP-1828](https://jira.stsci.edu/browse/JP-1828)


hbushouse (approved):
Is there a Jira ticket that this work goes with?

jemorrison:
This PR is in support of JP-1828

hbushouse:
Is this ready to be merged, so that we can run regtests on it?

PR #5600, opened 2021-01-09
Make stpipe more observatory-agnostic

eslavich:
This PR changes `stpipe.crds_client` to accept parameter dictionaries instead of models, adds new `JwstPipeline` and `JwstStep` base classes that each include a method that wraps datamodels.open, and changes `Pipeline` and `Step` to call that method whenever a model needs to be opened.  This will allow us to better support both Roman and JWST in a shared stpipe package.

I successfully ran the regression tests on this branch [here](https://plwishmaster.stsci.edu:8081/job/RT/job/eslavich-dev/55/).


hbushouse:
Because these look like fairly significant structural changes in stpipe and datamodels, I would like to hold off on merging this until after we've completed all release candidates for JWST DMS B7.7 sometime in the next couple weeks.

jwst/datamodels/container.py
```diff
@@ -381,6 +381,50 @@ def close(self):
...
+    def _open_first_science_exposure(self):
+        """
+        Open first model with exptype SCIENCE, or the first model
+        if none exists.
+
+        Returns
+        -------
+        stdatamodels.DataModel
+        """
+        for exposure in self.meta.asn_table.products[0].members:
```
jdavies-st, line 419: This `meta.asn_table` is not always guaranteed to be populated since one can use a container without an association.  So for robustness, this might have to be in a try/except.  The alternative way, if there's no association info on what is "SCIENCE" or not, is to just open the first item in the container:

```python
        return datamodel_open(self[0])
```

jwst/datamodels/container.py
```diff
@@ -381,6 +381,50 @@ def close(self):
...
+        Get the CRDS observatory for this container.  Used when selecting
+        step/pipeline parameter files when the container is a pipeline input.
+
+        Returns
+        -------
+        str
+        """
+        # Eventually ModelContainer will also be used for Roman, but this
+        # will work for now:
+        return "jwst"
```
jdavies-st, line 396: Since `ModelContainer` subclasses `JwstDataModel`, is this needed?

jwst/datamodels/model_base.py
```diff
@@ -7,9 +7,32 @@ class JwstDataModel(_DataModel):
...
+        """
+        Get parameters used by CRDS to select references for this model.
+
+        Returns
+        -------
+        dict
+        """
+        return {
+            key: val for key, val in self.to_flat_dict(include_arrays=False).items()
+            if isinstance(val, (str, int, float, complex, bool))
```
jdavies-st, line 31: I assume this avoids grabbing an complex objects like `gwcs.WCS` objects from `meta` and even the data arrays?  Ideally, we would like something that works like `to_flat_dict()` but only on `self.meta`, right?

jwst/datamodels/model_base.py
```diff
@@ -7,9 +7,32 @@ class JwstDataModel(_DataModel):
...
+        }
+
 
 # We may want to deprecate DataModel
 # @deprecate_class(JwstDataModel)
-class DataModel(_DataModel):
-
-    schema_url = "http://stsci.edu/schemas/jwst_datamodel/core.schema"
+class DataModel(JwstDataModel):
+    pass
```
jdavies-st, line 38: What is the reason for and effect of this change?

jwst/stpipe/pipeline.py
```diff
@@ -194,29 +190,25 @@ def get_config_from_reference(cls, dataset, observatory=None, disable=None):
...
+                refcfg['steps'][cal_step] = cal_step_class.get_config_from_reference(model)
+            #
+            # Now merge any config parameters from the step cfg file
+            log.log.debug(f'Retrieving pipeline {pars_model.meta.reftype.upper()} parameters from CRDS')
+            try:
+                ref_file = crds_client.get_reference_file(model.get_crds_parameters(),
+                                                        pars_model.meta.reftype,
+                                                        model.crds_observatory)
+            except (AttributeError, crds_client.CrdsError):
+                log.log.debug(f'{pars_model.meta.reftype.upper()}: No parameters found')
```
jdavies-st, line 205: Can we clean up the import so we don't have `log.log...`?

```python
from .log import log
```

jdavies-st:
@hbushouse, Ed suggested we make the base of this PR an epic branch that will stick around for a few weeks.  We can keep it up-to-date with `master`, develop on it, and then merge after all the B7.7 development is done.

jwst/datamodels/container.py
eslavich: This was in anticipation of your project to make `ModelContainer` independent of `DataModel`.  We're going to need a separate method here anyway since eventually we should pull this from one of the member models, but I didn't want to do that right away since at the moment that requires opening the model a second time.

jwst/datamodels/container.py
eslavich: @stscirij I think this is your code that I copied over from crds_client.py, do you have an opinion on this?

jwst/datamodels/model_base.py
eslavich: Yes CRDS only accepts a handful of basic types (we'll be adding `astropy.time.Time` soon I think).

jwst/datamodels/model_base.py
eslavich: I didn't want to have to define `crds_observatory` and `get_crds_parameters` twice.  I can't think of any downsides.

jwst/datamodels/container.py
perrygreenfield: This is a more general comment but I didn't see how to put it at the top. This is all being done with data models in the JWST package. How do you see this merging with stdatamodels? (Sorry if you answered this elsewhere.)

eslavich: @jdavies-st is working on moving `ModelContainer` to stdatamodels.  Once that's done we may or may not need a JWST-specific ModelContainer subclass.  If we do we can put this hard-coded "jwst" in there, if we don't then we'll have to retrieve the observatory code from one of the container's member models.

eslavich:
I created a new branch called "stpipe-transition" and changed the base branch of this PR to that.

perrygreenfield (approved):
I don't think I know stpipe well enough to really do a good review, but what I saw looks fine.

jdavies-st (approved):
🚀

nden (approved):

jwst/lib/suffix.py
```diff
@@ -267,7 +267,7 @@ def find_suffixes():
     is worth doing dynamically or only as a utility to update
     a static list.
     """
-    from jwst.stpipe import Step
+    from jwst.stpipe.step import Step
```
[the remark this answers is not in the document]
eslavich: This comment saved the day!  I thought this was due to an oddity in how "Spec2NRSLamp" is defined, but it turns out code in `Step` requires that `JwstPipeline` be a subclass of `JwstStep`.  I updated the `JwstPipeline` definition accordingly and now we can use `JwstStep` here like everywhere else.

PR #5601, opened 2021-01-09
JP-1694: fix outlier detection

### LINKED ISSUE #5332: outlier_detection flagging all non-source pixels
_Issue [JP-1694](https://jira.stsci.edu/browse/JP-1694) was created on JIRA by [Misty Cracraft](https://jira.stsci.edu/secure/ViewProfile.jspa?name=cracraft):_

Data are made up of eight simulated images with a starfield of ~50 stars. There are four dither positions and two exposures at each dither. Previously, when this data set was tested through build 7.5, the outlier_detection step was subtracting the sky (background) and leaving the sky averaging 0, with specific background pixels at negative values after the subtraction. All of those negative values were being flagged as outliers. When I re-tested these files through the development version of 7.6 listed above, the outlier detection step is no longer subtracting the background in the output crf files (output of outlier detection). However, now all non-source pixels are being flagged as outliers by the step, which is resulting in a completely invalid combined image.The pixels are being flagged as 17 or 21 which indicates both outlier detected and 'DO_NOT_USE' are being set, which makes combination difficult if all pixels are being rejected.

I used the following parameters in running calwebb_image3. 

fwhm=3.762 # Gaussian kernel FWHM of objects expected, default=2.5
 minobj=5 # minimum number of objects needed to match positions for a good fit, default=15
 snr= 250 # signal to noise threshold, default=5
 sigma= 3 # clipping limit, in sigma units, used when performing fit, default=3
 fit_geom='shift' # ftype of affine transformation to be considered when fitting catalogs, default='general'
 use2dhist=False # boolean indicating whether to use 2D histogram to find initial offset, default=True

I have the cal files, .crf files output from outlier detection to show the flagging in the DQ array and the combined image stored in the data location listed above. Only the files pertaining to this test are in that folder.

jemorrison:
This PR goes a long way to fixing combined data for MIRI.
the parameter good_bits has to be set to (~DO_NOT_USE+NON_SCIENCE) to ignore both NON_SCIENCE data and BAD data set with DO_NOT_USE.
In addition the noise image used to flag outliers was updated to account for the background subtracted region

Resolves #5332 / [JP-1694](https://jira.stsci.edu/browse/JP-1694)


hbushouse (commented):
I assume that the computed background needs to be added back into the blotted image because the individual cal image that it's going to be compared to still has the background signal in it, right? If so, that makes sense and hence that change seems appropriate.

I don't, however, see any change here related to the proper setting of good_bits.

jemorrison:
I put the change for the good_bits in the cfg file I was using.  I am not sure how other instruments  handle NON_SCIENCE. So I don't k now if that is a global default change or something we put in the code or something that gets added to the pars-outlierdetection ? I am still trying to understand how the 'pars' method will work.

hbushouse:
We recently stripped out all of the default parameter entries from all .cfg files for pipeline and steps, so now the defaults are set only in each step's `spec` block in the actual code module. For the outlier_detection step we have
https://github.com/spacetelescope/jwst/blob/master/jwst/outlier_detection/outlier_detection_step.py#L62
which sets the default `good_bits` to only `~DO_NOT_USE`. So changing that to also include `NON_SCIENCE` in your .cfg file is fine for the purpose of experimenting (that is exactly what param settings in .cfg are for - using non-default values), but if we determine that this is appropriate as a new default value, then it should get modified in the `spec` settings in the code itself.

Theoretically, however, the approach we have taken in most, if not all, pipeline steps with regard to which DQ values to use to signify that a pixel is to be ignored is in fact just the `DO_NOT_USE` value. All other flag values are deemed informational only. So in this case, if we really want `NON_SCIENCE` pixels to be excluded from use in any/all steps, those pixels should also have the `DO_NOT_USE` flag set for them (in addition to `NON_SCIENCE`). If that were done, then no changes would be needed for the `good_bits` parameter here. For the particular case of MIRI imaging, we'd need to find out where the `NON_SCIENCE` DQ values are getting set (e.g. are they propagated in from the FLAT ref files?) and, if so, possibly update those FLAT ref files  to also set `DO_NOT_USE` for those pixels.

jemorrison:
@hbushouse  on the reason the background has to be added - yes the cal image still has the background signal in it. 

@cracraft  @mwregan2  @karllark Is there any reason why the pixels in the lower 3 masks are set to NON_SCIENCE and not NON_SCIENCE+DO_NOT_USE ?
These pixels should not  in the calimage3 pipeline because using them messes up the outlier detection step. They get included in the median image. It would be cleaner if we added DO_NOT_USE to their dqflags.

cracraft:
Going back through emails with Ors, I found this note about the masking: Yes, the 4QPM should not be used. I did not mask the 4QPMs for 2 reasons. 1.) for onboard flats I needed them. 2.) Could be useful for image registration or could contain information that is valuable. Although I should mark NON_SCIENCE, to remove them from the mosaics, I totally agree. Is NO_FLAT_FIELD needed too? 

We also talked about which mask values to use (though didn't discuss DO_NOT_USE) in JP-164.

jemorrison:
I can not find where MIRI Imager pixels are marked at NON_SCIENCE in the pipeline. Could someone point me ? I did a grep on NON_SCIENCE in the pipeline that pointed me to photom step. But when I look at the code it seems NON_SCIENCE is set for NIRSPec data but I can not find where it is set for MIRI imager data.

cracraft:
NON_SCIENCE DQ values are in the flat files. It's not set as part of the pipeline, just read in from reference files.

jemorrison:
@cracraft so would it cause down stream problems if as part of the flat field step we set pixels with NON_SCIENCE to NON_SCIENCE + DO_NOT_USE.  You mentioned above that on board flats need these regions. Is that for making the flat fields not after they are applied.

cracraft:
Ors would have to answer why he wanted to use the flags he did, and what effect he anticipated it having through the pipeline. I don't know if that would be a problem or not. I don't see that it would, but others should weigh in on that.

jdavies-st:
I think it is fine to change the default in `OutlierDetectionStep.spec` to add NON_SCIENCE.  MIRI is the only one that uses this, and I think it won't have an effect on any other instruments.

hbushouse:
> I think it is fine to change the default in `OutlierDetectionStep.spec` to add NON_SCIENCE. MIRI is the only one that uses this, and I think it won't have an effect on any other instruments.

That's a solution that achieves the proper results, but goes against the philosophy of having all steps only reject pixels flagged with `DO_NOT_USE` by default. Any other flag values, like `NON_SCIENCE`, are to be considered as information only and not cause for rejection. So it would be better to add `DO_NOT_USE` flags to these `NON_SCIENCE` pixels where necessary.

jdavies-st:
Agree Howard, but as I mentioned in a previous thread on this, NON_SCIENCE might be DO_NOT_USE for one step, but fine to use on another step.  I.e. if two steps that have different requirements, `flat_field` and `outlier_detection` in this case, there's no resolving it using a single dq bit in a single reffile.

In the end, this DQ flag is being propagated from the flat reference file.  That doesn't seem ideal in controlling what happens in `outlier_detection`.  Ideally you'd want this DQ flag populated with DO_NOT_USE specifically for the level3 pipeline.

jemorrison:
@hbushouse 
@jdavies-st 
This PR is now ready for review. I added DO_NOT_USE to the pixels of NON_SCIENCE data after flat fielding.
So now adding the additional flag of NON_SCIENCE to good_bits is not needed in outlier_detection

jwst/flatfield/flat_field.py
```diff
@@ -215,6 +215,11 @@ def apply_flat_field(science, flat, inverse=False):
     # Combine the science and flat DQ arrays
     science.dq = np.bitwise_or(science.dq, flat_dq)
 
+    # Find all pixels in the flat that have a DQ value of NON_SCIENCE
+    # add the DO_NOT_USE flag to these pixels. We don't want to use these pixels
+    # in futher steps
+    flag_nonsci = np.bitwise_and(science.dq, dqflags.pixel['NON_SCIENCE']).astype(np.bool)
+    science.dq[flag_nonsci] = np.bitwise_or(science.dq[flag_nonsci], dqflags.pixel['DO_NOT_USE'])
```
jdavies-st, line 222: Adding NON_SCIENCE doesn't seem to have any effect on the issue described in the ticket.  I.e. in the dataset from the ticket, running `calwebb_image3` starting with the `_cal.fits` files as they are resolves the issue with just the changes in `outlier_detection`, without having to rerun `flat_field` on the input data.

So what is the purpose of this section of code?  Currently the NON_SCIENCE area in MIRI data gets set to zero in the resampled `_i2d.fits` output.  I.e., I don't see a problem with what is happening currently.

jdavies-st (commented):
Thanks for the excellent debug here Jane!  Just some questions and comments below.

jemorrison:
The preferred method is for  outlier detection to have good_bits = DO_NOT_USE only. If we do not go back to the flat field step and flag NON_SCIENCE data as also DO_NOT_USE then  this data is  in the combined image. This region should be in the combined image. The longer term fix is to get the MIRI team to set the NON_SCIENCE pixels in the flat field reference file to DO_NOT_USE + NON_SCIENCE.  We can then take out this code.

jemorrison:
The preferred method is for  outlier detection to have good_bits = DO_NOT_USE only. If we do not go back to the flat field step and flag NON_SCIENCE data as also DO_NOT_USE then  this data is  in the combined image. This region should NOT be in the combined image. The longer term fix is to get the MIRI team to set the NON_SCIENCE pixels in the flat field reference file to DO_NOT_USE + NON_SCIENCE.  We can then take out this code.

jdavies-st:
That makes sense Jane.  Thanks.  I think what I'm seeing is that `outlier_detection` is using the areas that are NON_SCIENCE, but because in the final resample of `calwebb_image3`, it throws out NON_SCIENCE in addition to DO_NOT_USE, it is masking whatever is happening in outlier.  I.e. here

https://github.com/spacetelescope/jwst/blob/2bbb67734e8e09b8f0b9a1d968c62228f9aba147/jwst/resample/resample_step.py#L19

jemorrison:
@jdavies-st 
I can confirm that the line in resample_step.py setting the GOOD_BIT is confusing the issue. It is hard coding the values. I went in and changed GOOD_BITS ='~DO_NOT_USE'  in the line and the resampled data now contains the NON_SCIENCE pixels. I still think it is best to flag these NON_SCIENCE pixels as also DO_NOT_USE because they are properly excluded from the i2d, median and blot data. We don't have to flag them in FLAT_FIELD it just seemed the place to do it since the NON_SCIENCE flag is coming from there.

jemorrison:
@jdavies-st  Can we ignore the Code style check - it seems to be related to a long line in a file that is not even part of the PR. 
I don't know why it is throwing an error

hbushouse:
> @jdavies-st Can we ignore the Code style check - it seems to be related to a long line in a file that is not even part of the PR.
> I don't know why it is throwing an error

I fixed the long line issue in a PR I recently merged. So it'll be gone once this gets merged with master.

CHANGES.rst
```diff
@@ -1,6 +1,14 @@
 0.18.2 (unreleased)
 ===================
 
+flat_field
```
hbushouse, line 4: Minor nit, but we like to keep the change entires in alphabetical order of step/pipeline name, so these should go below `cube_build` in the list.

hbushouse (commented):
Is the comment by @jdavies-st accurate, in that flagging the NON_SCIENCE pixels with DO_NOT_USE makes no difference to the final product? Does it make any difference in outlier_detection?

Ah, never mind. Just caught up on the rest of the comment thread here, which verifies that we do in fact want the NON_SCIENCE to have DO_NOT_USE added to them. So that means the only thing that needs changing yet is the change log.

jdavies-st (approved):
I am giving this PR my highest approval.  🥇 

🌟 🌟 🌟 🌟 

This not only fixes the reported issue for MIRI, but it also makes `outlier_detection` actually work for all the other instruments.

For instance, here's NIRCam imaging with 3 NRCB5 detectors at different roll angles.

Screen Shot 2021-01-14 at 3 51 56 PM

On the left is the i2d file for `jwst 0.18.1`.  On the right is this branch.  Outliers are actually detected and removed.

Rockon @jemorrison 🎉

CHANGES.rst
jemorrison: Ok I think I finally have the change log correct. I just edited it on GitHub

hbushouse: Yep, looks good to me now.

hbushouse (approved):
Nailed it.

hbushouse:
As soon as CI tests complete successfully I'll merge (hopefully in time for the nightly regtest run).

PR #5602, opened 2021-01-09
JP-1762: Add VA correction to WCS

mcara:
This PR adds support for velocity aberration correction in WCS. It needs thorough review.

It was decided, after a couple of iterations, that DVA correction will be applied as a separate transformation from the `v2v3` frame to a new, VA-corrected, `v2v3` frame that will be called `'v2v3vacorr'` in WCS pipeline.

The code applies a constant scale DVA correction in the (almost-)tangent plane `v2-v3` at Colin Cox suggestion. However, for when non-linear effects may become important, this PR also includes commented out full set of transformations from the spherical `v2,v3` coordinates to a tangent plane and then applying constant scale transformation in that plane and then backward transformation to the (DVA-corrected) spherical coordinates `v2,v3`.

### DIFF
--- a/jwst/assign_wcs/miri.py
+++ b/jwst/assign_wcs/miri.py
@@ -112,6 +112,11 @@ def imaging_distortion(input_model, reference_files):
     with DistortionModel(reference_files['distortion']) as dist:
         distortion = dist.model
 
+    # Apply differential velocity aberration (DVA) correction:
+    va_corr = pointing.va_corr_model(input_model)
+    if va_corr is not None:
+        distortion |= va_corr
+
     # Check if the transform in the reference file has a ``bounding_box``.
     # If not set a ``bounding_box`` equal to the size of the image.
     try:
@@ -195,6 +200,11 @@ def lrs_distortion(input_model, reference_files):
     # full array to v2v3 transform for the ordinary imager
     with DistortionModel(reference_files['distortion']) as dist:
         distortion = dist.model
+
+    # Apply differential velocity aberration (DVA) correction:
+    va_corr = pointing.va_corr_model(input_model)
+    if va_corr is not None:
+        distortion |= va_corr
 
     # Combine models to create subarray to v2v3 distortion
     if subarray2full is not None:
@@ -484,6 +494,8 @@ def abl_to_v2v3l(input_model, reference_files):
     # used to read the wavelength range
     channels = [c + band for c in channel]
 
+    va_corr = pointing.va_corr_model(input_model)
+
     with DistortionMRSModel(reference_files['distortion']) as dist:
         v23 = dict(zip(dist.abv2v3_model.channel_band, dist.abv2v3_model.model))
 
@@ -500,6 +512,11 @@ def abl_to_v2v3l(input_model, reference_files):
         ident1 = models.Identity(1, name='identity_lam')
         ident1._inputs = ('lam',)
         chan_v23 = v23[c]
+
+        # Apply differential velocity aberration (DVA) correction:
+        if va_corr is not None:
+            chan_v23 |= va_corr
+
         v23chan_backward = chan_v23.inverse
         del chan_v23.inverse
         v23_spatial = chan_v23
--- a/jwst/assign_wcs/miri.py
+++ b/jwst/assign_wcs/miri.py
@@ -113,7 +113,7 @@ def imaging_distortion(input_model, reference_files):
         distortion = dist.model
 
     # Apply differential velocity aberration (DVA) correction:
-    va_corr = pointing.va_corr_model(input_model, fast_corr=True)
+    va_corr = pointing.va_corr_model(input_model)
     if va_corr is not None:
         distortion |= va_corr
 
@@ -202,7 +202,7 @@ def lrs_distortion(input_model, reference_files):
         distortion = dist.model
 
     # Apply differential velocity aberration (DVA) correction:
-    va_corr = pointing.va_corr_model(input_model, fast_corr=True)
+    va_corr = pointing.va_corr_model(input_model)
     if va_corr is not None:
         distortion |= va_corr
 
@@ -494,7 +494,7 @@ def abl_to_v2v3l(input_model, reference_files):
     # used to read the wavelength range
     channels = [c + band for c in channel]
 
-    va_corr = pointing.va_corr_model(input_model, fast_corr=True)
+    va_corr = pointing.va_corr_model(input_model)
 
     with DistortionMRSModel(reference_files['distortion']) as dist:
         v23 = dict(zip(dist.abv2v3_model.channel_band, dist.abv2v3_model.model))
--- a/jwst/assign_wcs/miri.py
+++ b/jwst/assign_wcs/miri.py
@@ -113,7 +113,7 @@ def imaging_distortion(input_model, reference_files):
         distortion = dist.model
 
     # Apply differential velocity aberration (DVA) correction:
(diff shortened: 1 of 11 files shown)


jwst/assign_wcs/pointing.py
```diff
@@ -11,6 +16,9 @@
 
 from stdatamodels import DataModel
 
+_RAD2ARCSEC = 3600.0 * np.rad2deg(1.0)
+_ARCSEC2RAD = 1.0 / _RAD2ARCSEC
```
nden, line 20: Remove these, they are not used

jwst/assign_wcs/nirspec.py
```diff
@@ -142,6 +147,7 @@ def imaging(input_model, reference_files):
                             (sca, det2gwa),
                             (gwa, gwa2msa),
                             (msa_frame, None)]
+        # TODO: Add DVA correction
```
nden, line 150: There's no need to add VA correction here since this pipeline does not go to sky but stops at the MSA.

jwst/assign_wcs/nirspec.py
```diff
@@ -231,12 +237,13 @@ def ifu(input_model, reference_files, slit_y_range=[-.55, .55]):
                     (slit_frame, slit2slicer),
                     ('slicer', slicer2msa),
                     (msa_frame, None)]
+        # TODO: Add DVA correction
```
nden, line 240: No need for VA correction here, ti stops at the MSA.

jwst/assign_wcs/nirspec.py
```diff
@@ -345,14 +352,15 @@ def slitlets_wcs(input_model, reference_files, open_slits_id):
                         (gwa, gwa2slit),
                         (slit_frame, slit2msa),
                         (msa_frame, None)]
+        # TODO: Add DVA correction
```
nden, line 355: and here, no va correction

mcara:
Thanks @nden! I removed `TODO` comments, unused imports and fixed a bug due to which `v2_ref` and `v3_ref` were not converted to degrees.

jwst/assign_wcs/pointing.py
```diff
@@ -237,3 +237,110 @@ def create_fitswcs(inp, input_frame=None):
...
+    #     'zy',
+    #     name='det_to_optic_axis'
+    # )
+    #
+    # va_corr = (
+    #     unit_conv | s2c | rot | va_scale_corr |
+    #     rot.inverse | c2s | unit_conv.inverse
+    # )
+
+    va_corr.name = 'VA_Correction'
```
jdavies-st, line 344: The name should be descriptive, and bits are free.  "Spacecraft Velocity Aberration Correction" would be useful.

jwst/assign_wcs/tests/test_pointing.py
```diff
@@ -0,0 +1,66 @@
...
+from .test_nircam import create_hdul
+
+
+def create_imaging_datamodel(v2_ref, v3_ref, va_scale):
+    hdul = create_hdul()
+    datamodel = ImageModel(hdul)
+    datamodel.meta.velocity_aberration.scale_factor = va_scale
+    datamodel.meta.wcsinfo.v2_ref = v2_ref
+    datamodel.meta.wcsinfo.v3_ref = v3_ref
+    return datamodel
```
jdavies-st, line 17: This could be a Pytest fixture, and it would be much clearer that the tests depend on it.

jdavies-st (commented):
This looks good!

Just a couple minor comments below.  Not sure why we're seeing small changes in the computed coordinates, mostly in S_REGION.

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/220/testReport/

jwst/assign_wcs/miri.py
```diff
@@ -159,8 +164,17 @@ def lrs(input_model, reference_files):
     # v2v3 spatial component
     v2v3_spatial = cf.Frame2D(name='v2v3_spatial', axes_order=(0, 1), unit=(u.arcsec, u.arcsec),
                               axes_names=('v2', 'v3'))
+    v2v3vacorr_spatial = cf.Frame2D(
+        name='v2v3vacorr_spatial',
+        axes_order=(0, 1),
+        unit=(u.arcsec, u.arcsec),
+        axes_names=('v2', 'v3')
```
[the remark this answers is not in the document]
mcara: I do not know... I just copied the frame that was already used in the `assign_wcs` for a given instrument. I simply appended `vacorr` to the frame name. No other changes. I also noticed sometimes frames are called `v2` and sometimes `V3` - see https://github.com/spacetelescope/jwst/issues/5746 point `#4`. Maybe this should be noted there?

jwst/assign_wcs/pointing.py
mcara: @nden What do you think? I personally think that `VA_Correction` should be OK but if you think a longer string is more desirable - it is easy to change this.

jwst/assign_wcs/pointing.py
```diff
@@ -237,3 +237,110 @@ def create_fitswcs(inp, input_frame=None):
...
+    else:
+        if kwargs:
+            raise ValueError(
+                "'v2_ref', 'v3_ref', and 'va_scale' cannot be provided when "
+                "'datamodel' is not None."
+            )
+        try:
+            va_scale = float(datamodel.meta.velocity_aberration.scale_factor)
+            v2_ref = float(datamodel.meta.wcsinfo.v2_ref)
+            v3_ref = float(datamodel.meta.wcsinfo.v3_ref)
```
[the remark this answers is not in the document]
mcara: > You could just put the equation in the try block and if any of the items are None, you're going to get a TypeError.

Isn't what I am already doing? I did not know that `meta.velocity_aberration.scale_factor` and `wcsinfo.v2_ref` and `meta.wcsinfo.v3_ref` are *always guaranteed* to exist. If that is the case, I'll be happy to remove attribute error from being caught.

mcara:
> Just a couple minor comments below. Not sure why we're seeing small changes in the computed coordinates, mostly in S_REGION.
> 
> https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/220/testReport/

I investigated this and it looks that `VA_SCALE` was poulated with a value != 1 and hence now that VA correction is accounted for in the WCS, changes in produced world coordinates are expected. I think what we see here is reasonable.

jwst/assign_wcs/miri.py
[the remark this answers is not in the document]
jdavies-st: I see.  Maybe for a future "consistency and cleanup" PR then.  Agree to keep it as for this PR.

jwst/assign_wcs/pointing.py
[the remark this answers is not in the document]
jdavies-st: Yeah, they are always guaranteed to exist because they're in the schema and in `meta`.

jdavies-st (approved):
Thanks Mihai.  Yeah, it looks like some of our test data, probably from OTB or SDP processing, already has VA_SCALE set, so nice to see it exercised in your code.

jwst/assign_wcs/pointing.py
nden: Models can be referenced by name so I'd say shorter is better and not having spaces in the name is preferable.

mcara:
these reviewers...

nden (approved):
LGTM other than the code style issue

nden:
Needs a change log.

PR #5615, opened 2021-01-12
Fix wavelength tolerance for non-linear solutions

drlaw1558:
Bugfix to cube_build wavelength tolerances for non-linear wavelength solutions.  @jemorrison @hbushouse


drlaw1558:
Issue was that linear wavelength solutions use self.roiw while non-linear use self.roiw_table.  It isn't known until slightly later in the code which pieces of the table to use, and rather than introduce some kind of circular dependency I just declared the maximum roiw from the table (which should be ok as this is just a preselection convenience, with the actual selection of points done later).  Jane, see if you think this would have any issues.

jemorrison (approved):
Yeah that looks like it should fix the problem.

hbushouse (approved):

hbushouse:
Kicking off another round of regtests. Cross your fingers ...

PR #5620, opened 2021-01-14
JP-1407 Update MIRI LRS-Fixedslit along-slit-nod backgrounds strategies try 2

### LINKED ISSUE #4814: Updates to background subtraction strategies for MIRI LRS

stscieisenhamer:
Resolves [JP-1407](https://jira.stsci.edu/browse/JP-1407)
Resolves #4814

Other changes:

- Resolves issue with duplicate Level 3 associations. Original algorithm only checked for duplicate product names. However, for Level 3, product names are often different, but membership is the same.
- Fixes tests where the value of PATTTYPE being used was along_slit_nod and not the correct along-slit-nod.


hbushouse (approved):
LGTM. Fixed the change log so that the entry now appears up in the 0.18.2 section.

PR #5635, opened 2021-01-18
JWSTDMS-410 Asn_Lv2NRSLAMPSpectral: Break out the negative cases

stscieisenhamer:
Original test hid bug in how the constraints were setup. New case using JW00620, which covers a number of different LAMP modes, has been added to the regression.

Note: When this PR is merged, the regression data found in `jwst-pipeline/jwstdms-420-nolamp/associations` need to replace development data `dev/associations`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 associations
 ------------
 
+- JWSTDMS-410 Asn_Lv2NRSLAMPSpectral: Break out the negative cases [#5635]
+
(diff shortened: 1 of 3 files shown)


hbushouse (approved):

stscieisenhamer:
I am moving the appropriate artifactory data now.

hbushouse:
Thanks. I tried doing the move yesterday, but it's not obvious that it got moved properly or that I moved/copied the right stuff.

PR #5642, opened 2021-01-19
update to docs  for extract1d for IFU data 

jemorrison:
Documentation updates made in JP-1736.


docs/jwst/references_general/apcorr_reffile.inc
```diff
@@ -263,18 +240,55 @@ The row selection criteria for each instrument+mode are:
...
+varies with wavelength and the contents of the files are shown below.
+
+:Data model: `~jwst.datamodels.MirMrsApcorrModel`
+
++------------+-------+-------------+-----------+------------+----------+
+| Instrument | Mode  | Column name | Data type | Dimensions | Units    |
++============+=======+=============+===========+============+==========+
+| MIRI       | MRS   | wavelength  | float     | 1D array   | micron   |
++            +       +-------------+-----------+------------+----------+
+|            |       | radius      | float     | 1D array   | arcsec   |
```
drlaw1558, line 271: radius, apcor, and apcorr_err are all 2d arrays.  At the moment the 2nd dimension doesn't add information, but in future it could be used to provide different aperture corrections for different radii.

docs/jwst/extract_1d/description.rst
```diff
@@ -239,25 +239,27 @@ direction perpendicular to dispersion.
...
-will be used in the extraction process.
+circular annulus for background extraction and subtraction. The size of the extraction
+region and the background annulus size varies with wavelength. 
+The extraction related vectors are found in the asdf extract1d reference file.
+For each element in the ``wavelength`` vector there are three size components: ``radius``, ``inner_bkg``, and
+``outer_bkg``. The radius vector sets the extraction size; while ``inner_bkg`` and ``outer_bkg`` specify the
+limits of an annular background aperture.
+The extraction size parameters are given in units of arcseconds and converted to units of pixels
+in the extraction process. 
+
```
drlaw1558, line 257: Is it worth mentioning the axis_ratio and axis_pa vectors?  These are in the file, but are just placeholders for potential future functionality.

drlaw1558:
Might also be worth clarifying that for IFU mode the extraction location is either the RA/DEC target location indicated by the header (point sources), the center of the cube if the header target location is undefined (point sources) or the entire FOV (extended sources).  I.e., the --use_source_posn doesn't work (see https://jira.stsci.edu/browse/JP-1750)

drlaw1558 (approved):

jemorrison:
@hbushouse 
I would like to refer to the table in srctype step on single source observations when explaining how use_source_posn is set. How can I do that instead of just putting in the link to src type step :ref:`srctype <srctype_step> ?
Or maybe you have a better idea.

hbushouse:
> @hbushouse
> I would like to refer to the table in srctype step on single source observations when explaining how use_source_posn is set. How can I do that instead of just putting in the link to src type step :ref:`srctype <srctype_step> ?
> Or maybe you have a better idea.

If you add a tag/label to /docs/jwst/srctype/description.rst just before the table, like `.. _srctype_table:` (or whatever you want to call it), you can then reference it in the extract_1d doc using `:ref:`srctype_table``.

jemorrison:
@skendrew 
I updated the documentation for extract 1d for some IFU work. I added some clarification here for the use_source_posn and added a link to the srctype table. Is this good or do you want more description  ?

skendrew:
@jemorrison the documentation update looks good to me!

docs/jwst/extract_1d/description.rst
```diff
@@ -239,25 +247,27 @@ direction perpendicular to dispersion.
 
 Extraction for 3D IFU Data
 --------------------------
-For IFU cube data, 1D extraction is contolled by a different set of EXTRACT1D
-reference file parameters.
-Note that for an extended source, anything specified in the reference file
+For IFU cube data, 1D extraction is controlled by a different set of EXTRACT1D
```
hbushouse, line 250: This phrase seems to end unexpectedly and doesn't flow to the wording in the next line

docs/jwst/extract_1d/description.rst
```diff
@@ -239,25 +247,27 @@ direction perpendicular to dispersion.
...
+the RA/DEC target location indicated by the header or the center of the cube if the header
+target location is undefined. For extended source data, anything specified in the reference file
 or step arguments will be ignored; the entire image will be
-extracted, and no background subtraction will be done.
+extracted, and no background subtraction will be done. For IFU data the ``use_source_posn`` option is not used.
```
hbushouse, line 254: Really? Why isn't use_source_posn used when we know the data are using a point-source nod pattern? If the hardwired position of the extraction region in the ref file is set to the center of the cube, isn't that going to lead to bad extractions when nodding/dithering is involved and hence the source is not centered spatially in the cube?

docs/jwst/references_general/apcorr_selection.inc
```diff
@@ -4,12 +4,12 @@ CRDS selects appropriate APCORR references based on the following keywords.
...
 Instrument Keywords
-========== ======================================
+========== ======================================================
 FGS        INSTRUME, EXP_TYPE, DATE-OBS, TIME-OBS
 MIRI       INSTRUME, EXP_TYPE, DATE-OBS, TIME-OBS
 NIRCam     INSTRUME, EXP_TYPE, DATE-OBS, TIME-OBS
 NIRISS     INSTRUME, EXP_TYPE, DATE-OBS, TIME-OBS
-NIRSpec    INSTRUME, EXP_TYPE, DATE-OBS, TIME-OBS
-========== ======================================
+NIRSpec    INSTRUME, EXP_TYPE, FILTER, GRATING,DATE-OBS, TIME-OBS
```
hbushouse, line 14: need a whitespace following the comma in "GRATING,DATE-OBS"

docs/jwst/extract_1d/description.rst
jemorrison: The center of the extraction region is set by the target ra/dec and if that is not given then it is set by the center of the IFU. If this method does not work for nodded IFU data then we will need to update the ifu.py routine in extract.py. Nothing about use_source_posn is used in the routine.
@drlaw1558 comments ? I think the extract ifu.py is optimized for MIRI (does MIRI MRS even use Nodding ?)

hbushouse: OK, what you described is basically the same functionality as use_source_posn. For non-IFU modes, when use_source_posn=True, the center of the extraction is computed from the target RA/Dec, and when use_source_posn=False the position is based on the params from the extract1d ref file. So I guess we're OK. It just might be confusing to users that use_source_posn only applies to non-IFU data.

jemorrison: I was confused myself how the extraction parameters for different modes was working. It took me a while to understand it (and I am still learning). One reason why I explicitly said 'use_source_posn was not used in IFU data' was so that the user would not use it. Or try to use it and wonder why it did nothing.  Maybe we need to revisit this with NIRSpec Nodded IFU data. I will leave that to the NIRSpec experts.

hbushouse: If, as you said above, the center of the extraction region is set by the target ra/dec, that should work equally well for both MRS and NRS IFU dithers/nods, because it will force the extraction aperture to follow the source around the field of view. Unless of course what you described above is only being done for MRS data and not for NRS IFU, in which case that would be a problem. I would hope/assume that both types of IFU data are treated the same way.

jemorrison:
both IFU modes are treated the same way. Should I remove the line "For IFU data the 'use_source_posn` option is not used". Is that making it confusing ?

hbushouse:
No, I would not remove that line, because it's important for the users to know that that param has no effect on processing IFU data, so perhaps just rearrange the paragraph such that it starts with that line and then goes on to explain what happens instead.  e.g. something like "The `use_source_posn` parameters is not used when extracting spectra from an IFU cube. Instead, the extraction aperture is centered at the RA/Dec of the target, ...."

hbushouse (approved):

PR #5644, opened 2021-01-20
JP-1622: Combine1D should output separate extensions for each spectral order

### LINKED ISSUE #5204: WFSS combine_1d step does not output combined spectrum for both spectral orders with F090W filter

tapastro:
Resolves [JP-1622](https://jira.stsci.edu/browse/JP-1622)
Resolves #5204

First PR had a code error, using an incorrect index parameter. This fix should create distinct extensions with appropriate wavelength and flux vectors.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,12 @@ associations
 
 - Update MIRI LRS-Fixedslit ALONG-SLIT-NOD backgrounds strategies [#5620]
 
+combine1d
+---------
+
+- Fixed code error in combine1d, creating extensions per spectral order
+  with the same input data [#5644]
+
 cube_build
 ----------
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):

tapastro:
One day I will get git.

PR #5648, opened 2021-01-20
Add hook to finalize Step results

eslavich:
This creates a `Step.finalize_result` hook that allows `Step` subclasses to set mission-specific metadata on each result before it is saved.  I also moved the `Step.record_step_status` method to `JwstStep` since it's only used in the master background steps and the Roman schemas don't include those metadata fields.

### DIFF
--- a/jwst/stpipe/jwst.py
+++ b/jwst/stpipe/jwst.py
@@ -1,9 +1,15 @@
 """
 JWST-specific Step and Pipeline base classes.
 """
+from stdatamodels import DataModel
+
+from .. import __version_commit__, __version__
+from .. import datamodels
+from ..datamodels import ModelContainer
+
+from . import crds_client
 from .step import Step
 from .pipeline import Pipeline
-from .. import datamodels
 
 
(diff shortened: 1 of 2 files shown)


jdavies-st (approved):
Looks good.

`record_step_status()` was added to the infrastructure only last year, which is why `master_background` is the only step to use it.  Other steps should probably be using it as well too.

I suspect `romancal` will need similar metadata and similar functionality.

jdavies-st:
Btw, `record_step_status()` is required as a callable method because there's no automagical mapping between a step name and what it's status metadata keyword is.  If there was, then it could just be implemented via a hook and private method, with the step itself probably not needing to call it under normal circumstances, and if it did need to call it, it wouldn't need args.

eslavich:
> Btw, record_step_status() is required as a callable method because there's no automagical mapping between a step name and what it's status metadata keyword is.

Maybe we ought to add a class attribute to the step alongside `spec` and `reference_file_types` that specifies the metadata keyword?

PR #5650, opened 2021-01-20
Move class_property module into stpipe

eslavich:
The class_property module contains general purpose tools but is not currently used anywhere but stpipe.  Moving it into stpipe will allow stpipe to exist independently of the jwst package.


jdavies-st:
This `class_property` class was introduced because of the issues with order-of-operations in instantiation and needing to have a property available on the class before `__init__`.  I suspect (and would hope) that a redesign of the `Step` class and how it operates would negate the need of this class.

jdavies-st (approved):

PR #5652, opened 2021-01-20
Add overrideable remove_suffix method to Step

eslavich:
This PR adds a `Step.remove_suffix` method that is then overridden in `JwstStep`.  This breaks away more JWST-specific code from the base `Step` class.

### DIFF
--- a/jwst/stpipe/step.py
+++ b/jwst/stpipe/step.py
@@ -501,7 +501,7 @@ class Step(abc.ABC):
         str
             Separator that delimited the original suffix.
         """
-        return name
+        return name, "_"
 
     def prefetch(self, *args):
         """Prefetch reference files,  nominally called when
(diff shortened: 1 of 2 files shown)


jwst/stpipe/step.py
```diff
@@ -485,6 +484,25 @@ def finalize_result(self, result, reference_files_used):
...
+
+        Returns
+        -------
+        str
+            Filename with any known suffix removed.
+        str
+            Separator that delimited the original suffix.
+        """
+        return name
+
```
jdavies-st, line 505: This looks like a dummy implementation that does nothing.  Would it be better to return `NotImplemented`?

jdavies-st:
Also just a general comment that currently the coverage on these changes is not getting compared to any baseline.  We'll want to make sure we have a good measurement when we merge this into `master`, and add some unit tests if needed.

jwst/stpipe/step.py
eslavich: This seemed like a reasonable default behavior since the original method returns the unmodified name if it doesn't contain any of the known substrings.  I notice now that I didn't return the second string here (the separator) so I should at least fix that.

jdavies-st (approved):
🚀

stscieisenhamer (approved):

PR #5653, opened 2021-01-21
JP-1860: Fix ramp_fit counting of segments

### LINKED ISSUE #5630: ramp_fit crashing for some NIRCam images in DMS B7.7 testing
_Issue [JP-1860](https://jira.stsci.edu/browse/JP-1860) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The ramp_fit step is crashing with an "index out of bounds" error for several NIRCam images in the DMS B7.7 test suites. The images are full-frame with NGROUPS=4 and NINTS=1.

hbushouse:
Update the function in ramp_fitting that attempts to estimate the max number of segments that will be needed to fit any pixel, given the number of CR and DO_NOT_USE flags that break up each pixel's ramp. Previously the function only looked for CR (JUMP_DET) flags, but a recent update to the saturation step allows for adding DO_NOT_USE flags to groups with values below the A-to-D floor (< 0). So now the ramp for a given pixel can be broken into multiple segments due to both JUMP_DET and DO_NOT_USE flags, so we need to check for both. Checking for only JUMP_DET was causing the data arrays to be created too small to hold data from all segments, leading to an "index out of bounds" error when trying to stuff data into the array.

Fixes #5630 / [JP-1860](https://jira.stsci.edu/browse/JP-1860)

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -30,6 +30,9 @@ from . import utils
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
+DO_NOT_USE = dqflags.group['DO_NOT_USE']
+JUMP_DET = dqflags.group['JUMP_DET']
+
 BUFSIZE = 1024 * 300000  # 300Mb cache size for data section
 
 
@@ -201,6 +204,7 @@ def ols_ramp_fit_multi(input_model, buffsize, save_opt, readnoise_2d, gain_2d,
     # Call ramp fitting for the single processor (1 data slice) case
     if number_slices == 1:
         max_segments, max_CRs = calc_num_seg(input_model.groupdq, number_of_integrations)
+        log.debug(f"Max segments={max_segments}")
         int_model, opt_model, out_model = create_output_models(input_model,
                                             number_of_integrations, save_opt, total_cols, total_rows,
                                             max_segments, max_CRs)
@@ -2941,11 +2945,10 @@ def calc_num_seg(gdq, n_int):
     max_cr = 0  # max number of CRS for all integrations
 
     # For all 2d pixels, get max number of CRs or DO_NOT_USE flags along their
-    # ramps, to use as a surrogate for the number of semiramps along the ramps
-    for nint in range(n_int):  # loop over integrations
-        gdq_cr = np.bitwise_and(gdq[nint], dqflags.group['JUMP_DET'] | dqflags.group['DO_NOT_USE'])
-        temp_max_cr = int((gdq_cr.sum(axis=0)).max())
-        max_cr = max(max_cr, temp_max_cr)
+    # ramps, to use as a surrogate for the number of segments along the ramps
+    # Note that we only care about flags that are NOT in the first or last groups,
+    # because exclusion of a first or last group won't result in an additional segment.
+    max_cr = np.count_nonzero(np.bitwise_and(gdq[:, 1:-1], JUMP_DET | DO_NOT_USE), axis=1).max()
 
     # Do not want to return a value > the number of groups, which can occur if
     #  this is a MIRI dataset in which the first or last group was flagged as
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -2940,12 +2940,12 @@ def calc_num_seg(gdq, n_int):
     """
     max_cr = 0  # max number of CRS for all integrations
 
-    # For all 2d pixels, get max number of CRs along their ramps
-    # to use as a surrogate for the number of semiramps along the ramps
+    # For all 2d pixels, get max number of CRs or DO_NOT_USE flags along their
+    # ramps, to use as a surrogate for the number of semiramps along the ramps
     for nint in range(n_int):  # loop over integrations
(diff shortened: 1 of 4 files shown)


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -2940,12 +2940,12 @@ def calc_num_seg(gdq, n_int):
...
 
-    # For all 2d pixels, get max number of CRs along their ramps
-    # to use as a surrogate for the number of semiramps along the ramps
+    # For all 2d pixels, get max number of CRs or DO_NOT_USE flags along their
+    # ramps, to use as a surrogate for the number of semiramps along the ramps
     for nint in range(n_int):  # loop over integrations
-        gdq_cr = np.bitwise_and( gdq[nint, :, :, :], dqflags.group['JUMP_DET'])
-        temp_max_cr = int((gdq_cr.sum(axis=0)).max()/dqflags.group['JUMP_DET'])
-        max_cr = max( max_cr, temp_max_cr )
+        gdq_cr = np.bitwise_and(gdq[nint], dqflags.group['JUMP_DET'] | dqflags.group['DO_NOT_USE'])
```
jdavies-st, line 2946: The expression:

```python
dqflags.group['JUMP_DET'] | dqflags.group['DO_NOT_USE']
```
always evaluates to 5.  So the above is really:

```python
gdq_cr = np.bitwise_and(gdq[nint], 5)
```
which will pick out all the 1, 4 and 5 values in `gdq`.  For example:

```
In [24]: gdq                                                                                                          
Out[24]: array([0, 1, 4, 5, 8])

In [25]: np.bitwise_and(gdq, 5)                                                                                       
Out[25]: array([0, 1, 4, 5, 0])
```

But then when you sum these, the 4 and 5 are going to give you a much higher count.  Instead of summing you may want to count them using `np.count_nonzero`.  Summing was fine when the values were all 4 and that could be corrected by dividing by 4, but not now.

jdavies-st: Also, please write a quick unit test for this so we don't break it again.  😄

hbushouse: Hmmm, could've sworn I had put in an appropriate expression that basically returned an array of 1's and 0's (equivalent to True/False), such that the .sum() really did just count up the number of hits, as opposed to the sum of all the flag values. Will look at that again.

Unit tests: to quote the Beatles, "yeah, yeah, yeah, ..."

jdavies-st: Also also, I would pick either:

```python
np.bitwise_and(gdq[nint], np.bitwise_or(dqflags.group['JUMP_DET'], dqflags.group['DO_NOT_USE']))
```

or 

```python
gdq[int] & (dqflags.group['JUMP_DET'] | dqflags.group['DO_NOT_USE'])
```

not mix up the function calls and operators (which point to the function calls).

Also also also, might be nice to put something like

```python
DO_NOT_USE = dqflags.group['DO_NOT_USE']
```

globally in this module as we've done elsewhere to make these sorts of things a bit more clear and concise:

```python
gdq[int] & (JUMP_DET | DO_NOT_USE)
```

stscieisenhamer (commented):
Besides the existing commentary, LGTM.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -2940,12 +2944,11 @@ def calc_num_seg(gdq, n_int):
...
-    # to use as a surrogate for the number of semiramps along the ramps
-    for nint in range(n_int):  # loop over integrations
-        gdq_cr = np.bitwise_and( gdq[nint, :, :, :], dqflags.group['JUMP_DET'])
-        temp_max_cr = int((gdq_cr.sum(axis=0)).max()/dqflags.group['JUMP_DET'])
-        max_cr = max( max_cr, temp_max_cr )
+    # For all 2d pixels, get max number of CRs or DO_NOT_USE flags along their
+    # ramps, to use as a surrogate for the number of segments along the ramps
+    # Note that we only care about flags that are NOT in the first or last groups,
+    # because exclusion of a first or last group won't result in an additional segment.
+    max_cr = np.count_nonzero(np.bitwise_and(gdq[:, 1:-1], JUMP_DET | DO_NOT_USE), axis=1).max()
```
hbushouse, line 2951: note that max_cr is bumped up by 1 later in line 2956, to account for the fact that n flags implies n+1 segments.

stscieisenhamer: Minor comment: Is there a reason not to bump now? Or is the later use a conceptually different value, segments, as opposed to "maximum rejects".

stscieisenhamer (approved):

hbushouse:
Errors in CI appear to be due to access problems with crds_cache. Trying again.

jwst/ramp_fitting/ramp_fit.py
hbushouse: No really good reason I suppose, although they are a bit different conceptually. This line simply computes the max number "interrupts" in any ramp, and then that number of interrupts turns into n+1 segments for fitting. Besides, the line is so long already, I didn't want to add yet another operation on the end of it!

PR #5657, opened 2021-01-22
JP-1873: Fix set_telescope_pointing TSO logic

### LINKED ISSUE #5656: Fix set_telescope_pointing handling of TSO exposures
_Issue [JP-1873](https://jira.stsci.edu/browse/JP-1873) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

A recent change to the way the set_telescope_pointing script determines whether an exposure is to be treated as TSO has a gap in the logic. The change was to only look at the value of TSOVISIT. But there is at least one corner case where EXP_TYPE=NRC_TSGRISM, TSOVISIT=False, and NINTS=1 (due to weird combination of APT params from NIRCam engineering template) that currently is not treated as TSO (because TSOVISIT=F), but it would be better for downstream processing to treat it as TSO. This will allow the exposure to fail more gracefully in calwebb_tso3 processing than the way it's currently crashing in calwebb_spec2.

Upgrade the logic in set_telescope_pointing to force any exposures using one of the hardwired TSO EXP_TYPEs (NRC_TSIMAGE, NRC_TSGRISM, NRS_BRIGHTOBJ) to always be treated as TSO, regardless of the values of TSOVISIT and NINTS.

hbushouse:
Updated the TSO check in set_telescope_pointing to use a combination of the `is_tso()` function and EXP_TYPE values, instead of relying solely on the value of TSOVISIT. This handles corner cases that have TSOVISIT=False and NINTS=1, which would normally come back as False for the TSO test, to come back as True when EXP_TYPE is a hardwired TSO mode.

Fixes #5656 / [JP-1873](https://jira.stsci.edu/browse/JP-1873)

### DIFF
--- a/jwst/lib/set_telescope_pointing.py
+++ b/jwst/lib/set_telescope_pointing.py
@@ -12,8 +12,10 @@ import numpy as np
 
 from ..assign_wcs.util import update_s_region_keyword, calc_rotation_matrix
 from ..assign_wcs.pointing import v23tosky
+from ..associations.lib.dms_base import TSO_EXP_TYPES
 from ..datamodels import Level1bModel
 from ..lib.engdb_tools import ENGDB_Service
+from ..lib.pipe_utils import is_tso
 from .exposure_types import IMAGING_TYPES, FGS_GUIDE_EXP_TYPES
 
 TYPES_TO_UPDATE = set(list(IMAGING_TYPES) + FGS_GUIDE_EXP_TYPES)
@@ -1516,8 +1518,12 @@ def populate_model_from_siaf(model, siaf):
 
     # For TSO exposures, also populate XREF_SCI/YREF_SCI keywords,
     # which are used by the Cal pipeline to determine the
-    # location of the source
-    if model.meta.visit.tsovisit:
+    # location of the source.
+    # Note that we use a combination of the is_tso function and
+    # a check on EXP_TYPE, because there are rare corner cases
+    # where EXP_TIME=NRC_TSGRISM, TSOVISIT=False, NINTS=1, which
+    # normally return False, but we want to treat it as TSO anyway.
+    if is_tso(model) or model.meta.exposure.type.lower() in TSO_EXP_TYPES:
         logger.info('TSO exposure:')
         logger.info(' setting xref_sci to {}'.format(siaf.crpix1))
         logger.info(' setting yref_sci to {}'.format(siaf.crpix2))
--- a/jwst/lib/set_telescope_pointing.py
+++ b/jwst/lib/set_telescope_pointing.py
@@ -12,7 +12,6 @@ import numpy as np
 
 from ..assign_wcs.util import update_s_region_keyword, calc_rotation_matrix
 from ..assign_wcs.pointing import v23tosky
-from ..associations.lib.dms_base import TSO_EXP_TYPES
 from ..datamodels import Level1bModel
 from ..lib.engdb_tools import ENGDB_Service
 from ..lib.pipe_utils import is_tso
@@ -1523,7 +1522,7 @@ def populate_model_from_siaf(model, siaf):
     # a check on EXP_TYPE, because there are rare corner cases
     # where EXP_TIME=NRC_TSGRISM, TSOVISIT=False, NINTS=1, which
     # normally return False, but we want to treat it as TSO anyway.
-    if is_tso(model) or model.meta.exposure.type.lower() in TSO_EXP_TYPES:
+    if is_tso(model) or model.meta.exposure.type.lower() in ['nrc_tsimage', 'nrc_tsgrism']:
         logger.info('TSO exposure:')
         logger.info(' setting xref_sci to {}'.format(siaf.crpix1))
         logger.info(' setting yref_sci to {}'.format(siaf.crpix2))
--- a/jwst/lib/tests/test_pipe_utils.py
+++ b/jwst/lib/tests/test_pipe_utils.py
@@ -54,12 +54,16 @@ def test_is_tso_nrcgrism_nints1():
     model.meta.exposure.type = "NRC_TSGRISM"
     model.meta.visit.tsovisit = False
     model.meta.exposure.nints = 10
-    assert pipe_utils.is_tso(model) is True
 
+    # with NINTS>1, should be True
+    assert pipe_utils.is_tso(model)
+
+    # with NINTS=1, should be False
     model.meta.exposure.nints = 1
-    assert pipe_utils.is_tso(model) is False
+    assert not pipe_utils.is_tso(model)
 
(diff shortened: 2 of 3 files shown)


hbushouse:
Look @jdavies-st I added unit tests!

nden:
Does `NRS_BRIGHTOBJ` go through tso_photometry?

hbushouse:
> Does `NRS_BRIGHTOBJ` go through tso_photometry?

It goes through the calwebb_tso3 pipeline, but not the tso_photometry step. tso_photometry is only for TSO imaging, not spectroscopy.

jdavies-st:
> Look @jdavies-st I added unit tests!

b9789d4f40897f4c60d06824cae22c0b

jwst/lib/tests/test_pipe_utils.py
```diff
@@ -47,6 +48,20 @@ def test_is_tso_from_tsoflag(tsovisit, expected):
     assert pipe_utils.is_tso(model) is expected
 
 
+def test_is_tso_nrcgrism_nints1():
+    """Test is_tso with NRC_TSGRISM and NINTS=1"""
+    model = datamodels.JwstDataModel()
+    model.meta.exposure.type = "NRC_TSGRISM"
+    model.meta.visit.tsovisit = False
+    model.meta.exposure.nints = 10
+    assert pipe_utils.is_tso(model) is True
```
jdavies-st, line 57: `assert pipe_utils.is_tso(model)` is sufficient, as it returns `True`.

jwst/lib/tests/test_pipe_utils.py
```diff
@@ -47,6 +48,20 @@ def test_is_tso_from_tsoflag(tsovisit, expected):
...
+
+    model.meta.exposure.nints = 1
+    assert pipe_utils.is_tso(model) is False
```
jdavies-st, line 60: `assert not pipe_utils.is_tso(model)` is probably a bit clearer.

jwst/lib/tests/test_pipe_utils.py
```diff
@@ -47,6 +48,20 @@ def test_is_tso_from_tsoflag(tsovisit, expected):
...
+
+    assert (is_tso(model) or model.meta.exposure.type.lower() in dms_base.TSO_EXP_TYPES) is True
```
jdavies-st, line 62: You don't need to `assert is True`, because the assertion itself returns `True` or `False`.

jdavies-st (approved):
Horray!  Just some minor comments on using `assert` in tests below.

jwst/lib/set_telescope_pointing.py
```diff
@@ -1516,8 +1518,12 @@ def populate_model_from_siaf(model, siaf):
...
     # For TSO exposures, also populate XREF_SCI/YREF_SCI keywords,
     # which are used by the Cal pipeline to determine the
-    # location of the source
-    if model.meta.visit.tsovisit:
+    # location of the source.
+    # Note that we use a combination of the is_tso function and
+    # a check on EXP_TYPE, because there are rare corner cases
+    # where EXP_TIME=NRC_TSGRISM, TSOVISIT=False, NINTS=1, which
+    # normally return False, but we want to treat it as TSO anyway.
+    if is_tso(model) or model.meta.exposure.type.lower() in TSO_EXP_TYPES:
```
nden, line 1526: Since `NRS_BRIGHTOBJ` is also in `TSO_EXP_TYPES` and it doesn't need those two keywords perhaps change this line to
`if is_tso(model) or model.meta.exposure.type.lower() in ['NRC_TSIMAGE', NRS_TSGRISM']"`
It may be confusing to have those two keywords in NRS_BRIGHTOBJ files.

hbushouse: Instead of picking and choosing which of the various TSO modes to put XREF_SCI and YREF_SCI into, I thought it would be simpler and more consistent to just put them in *all* TSO modes. We need them for all TSO imaging and we also need them for at least NRC_TSGRISM on the spectroscopic side, so why not just put them into all TSO exposures.

nden: IMHO they are a kludge and in the case of `NRS_BRIGHTOBJ` only create clutter. Their meaning is (at least for NRC_TSGRISM) that the source is located there which isn't true for `NRS_BRIGHTOBJ` and may be confusing to end users.

hbushouse: OK boss, I'll fix it in the next commit.

jwst/lib/tests/test_pipe_utils.py
hbushouse: done

jwst/lib/tests/test_pipe_utils.py
hbushouse: done

jwst/lib/tests/test_pipe_utils.py
hbushouse: done

jwst/lib/set_telescope_pointing.py
hbushouse: done

jwst/lib/set_telescope_pointing.py
```diff
@@ -1516,8 +1517,12 @@ def populate_model_from_siaf(model, siaf):
...
     # For TSO exposures, also populate XREF_SCI/YREF_SCI keywords,
     # which are used by the Cal pipeline to determine the
-    # location of the source
-    if model.meta.visit.tsovisit:
+    # location of the source.
+    # Note that we use a combination of the is_tso function and
+    # a check on EXP_TYPE, because there are rare corner cases
+    # where EXP_TIME=NRC_TSGRISM, TSOVISIT=False, NINTS=1, which
+    # normally return False, but we want to treat it as TSO anyway.
+    if is_tso(model) or model.meta.exposure.type.lower() in ['nrc_tsimage', 'nrc_tsgrism']:
```
stscieisenhamer, line 1525: Big clarification question: `is_tso` enforces NINTS >1. Conceptually, does TSO data have to have NINTS>1? I.e., if this particular check was ignored, and all other indicators say `is_tso` is True, that any code would be correct in continuing processing such data as TSO?

stscieisenhamer (approved):

jwst/lib/set_telescope_pointing.py
hbushouse: Unfortunately we've now gotten into gray territory where the answer to NINTS > 1 is "it depends". There are some cases where it's OK and even desired to allow NINTS=1 and others where it isn't. Hence this code essentially overrides `is_tso` in the particular case of NRC_TSGRISM, TSOVISIT=F, and NINTS=1.

hbushouse: Otherwise, you're right that it would've been far simpler to just change `is_tso` itself to allow NINTS=1.

PR #5662, opened 2021-01-25
Fix premature closing of file in apcorr reffile tests

jdavies-st:
These are failing using `asdf` master.  Not sure why they are not failing now, as a table is being accessed after the file is closed.  @eslavich ?

https://github.com/jdavies-st/asdf/runs/1760011694?check_suite_focus=true#step:6:456


eslavich (approved):
This LGTM.  I'm not sure why the test succeeds with asdf 2.7.x, but due to memory mapping and lazy array loading this is safer practice anyway.

jemorrison (approved):

PR #5665, opened 2021-01-25
Make pars-related Step methods instance methods only

eslavich:
This is some simplification of the pars-related Step methods:

- Make `get_pars` and `get_pars_model` instance methods instead of hybrid instance/class methods.  We aren't currently using them as class methods anyway.
- Construct the CRDS reftype for pars files from the `Step` class name only.  Some code existed that used the `Step` instance `name` property in the reftype, but that wasn't reachable from the CLI.

Successful regression test run on this branch here: https://plwishmaster.stsci.edu:8081/job/RT/job/eslavich-dev/57/


stscieisenhamer (approved):

jdavies-st:
Well, this looks great.  How does this resolve the discussion in https://github.com/spacetelescope/jwst/pull/4117#discussion_r334320015?

Issue #5658, opened 2021-01-23
Computation of velocity aberrated RA, DEC and scale


mcara:
In https://github.com/spacetelescope/jwst/pull/5602 I implemented _constant scale correction_ to be applied to `gwcs` in order to obtain Velocity Aberration (VA)-free coordinates. That correction relies on constant scale value supplied in `datamodel.meta.velocity_aberration.scale_factor`. My understanding is that this value is computed by https://github.com/spacetelescope/jwst/blob/master/jwst/lib/set_velocity_aberration.py

I do not know where this script is called and how it is used. However, I have noticed a couple of apparent inconsistencies in the returned values of the [`aberration_offset()`](https://github.com/spacetelescope/jwst/blob/4a60fe1ad92a50a38a15e25b43402ac024c4406e/jwst/lib/set_velocity_aberration.py#L90-L127) function:

1. It returns _difference_ between corrected RA, DEC and uncorrected values;
2. [Even more weird is that] It returns above differences converted to radians;
3. However, [datamodel's schema](https://github.com/spacetelescope/jwst/blob/4a60fe1ad92a50a38a15e25b43402ac024c4406e/jwst/datamodels/schemas/core.schema.yaml#L1292-L1305) indeed says (confirms) that the `DVA_RA` and `DVA_DEC` are in `rad`.

Here are my comments and questions:
1. `DVA` implies "differential" effects of VA, that is, effects left over after correcting for the absolute ("offsets") aberration. However, `DVA_RA` and `DVA_DEC` seem to be the absolute corrections, not the differential that are (mostly) accounted for by a constant scale correction.
2. Another issue is that reported values are deltas. Shouldn't we report corrected `RA` and `DEC` instead of "corrections to old values"? This will also make life easier since after adding "corrections" to `RA` and `DEC` one would need to take care of the 180 and 360 angle wrapping.
3. Based on points 1. and 2. above, I would suggest renaming these names in `meta` to `VACORR_RA` and `VACORR_DEC` (or `VA_RA` and `VA_DEC` or something along these lines) _that report **corrected(?) `RA` and `DEC` values in degrees**_.
[...]

PR #5666, opened 2021-01-26
Improve code for computing VA scale and offsets

mcara:
This PR improves code in `lib/set_velocity_aberration.py` based on suggestions from Colin Cox:
1. Simplify formula for computing constant scale (neglect higher-order terms);
2. Improve formula for computing RA, DEC offsets based on example code supplied by Colin Cox;
3. Overall refactoring.

Closes: #5658
Related to [JP-1762](https://jira.stsci.edu/browse/JP-1762)


hbushouse:
Need to update unit tests in test_velocity_aberration.py, which are causing CI to fail.

hbushouse (commented):
Code refactoring looks good to me. Just need to fix the the unit tests that are failing.

mcara:
@hbushouse @ColinRCox @antonkoekemoer @nden Could you please take a look at https://github.com/spacetelescope/jwst/issues/5658 and comment on the issues either here or there or ping someone who is knowledgeable in how this code and its computed values are intended to be used.

Thanks!

mcara:
@hbushouse I have made several significant changes to code and schemas. Please re-review this PR. For schema review, maybe @jdavies-st could take a look (don't I need to update a schema version somewhere?)

Here is the outline of changes:

1. I have combined computation of scale with the computation of apparent RA and DEC;
2. Instead of "offsets" - the differences between RA (DEC) in the moving telescope and barycentric frames, now the values in the moving frame are reported (apparent RA and DEC);
3. FITS keywords have been changed to `VA_RA` and `VA_DEC` (I wanted to have `VA_REF_DEC` but it does not fit into 8 characters; the reason for `REF` is that script computes VA angles of `meta.wcsinfo.ra_ref` and `meta.wcsinfo.dec_ref`; the reason I renamed `DVA` to `VA` is because `D` comes from differential which, I believe, does not apply to these quantities).
4. `VA_RA` and `VA_DEC` are returned in degrees and not radians;
5. In `jwst/datamodels/schemas/core.schema.yaml` I have renamed `ra(dec)_offset` to `va_ra(dec)_ref`;
6. I have somewhat modified Colin Cox' code to return RA in the range `[0, 360)` instead of `[-180, 180)` (by re-using `gwcs.geometry.CartesianToSpherical` model).
7. Updated unit tests. Previous tests were using `RA_ref=0` and `DEC_ref=0` which made test for the case of zero velocity to not be able to distinguish between offsets and absolute angle values.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1293,15 +1293,15 @@ properties:
...
+          va_ra_ref:
+            title: "[deg] Velocity aberration apparent reference RA"
             type: number
-            fits_keyword: DVA_RA
+            fits_keyword: VA_RA
             blend_table: True
-          dec_offset:
-            title: "[rad] Velocity aberration correction Dec offset"
+          va_dec_ref:
+            title: "[dec] Velocity aberration apparent reference Dec"
```
hbushouse, line 1302: typo in "dec" - should be "de**g**"

jwst/lib/set_velocity_aberration.py
```diff
@@ -45,95 +45,50 @@ def aberration_scale(velocity_x, velocity_y, velocity_z,
...
 
-    dot_prod = (velocity_x * targ_x +
-                velocity_y * targ_y +
-                velocity_z * targ_z)
-    cos_theta = dot_prod / speed
-    # This sin_theta is only valid over the range [0, pi], but so is the
-    # angle between the velocity vector and the direction toward the target.
-    sin_theta = math.sqrt(1. - cos_theta**2)
+    apparent_ra: float
+        Aparent star position in the moving telescope frame.
```
hbushouse, line 60: typo: "Aparent" should be "Ap**p**arent"

jwst/lib/set_velocity_aberration.py
```diff
@@ -45,95 +45,50 @@ def aberration_scale(velocity_x, velocity_y, velocity_z,
...
 
-    tan_theta_p = sin_theta / (gamma * (cos_theta + beta))
-    theta_p = math.atan(tan_theta_p)
+    apparent_dec: float
+        Aparent star position in the moving telescope frame.
```
hbushouse, line 63: same typo here "Ap**p**arent"

jwst/lib/set_velocity_aberration.py
```diff
@@ -45,95 +45,50 @@ def aberration_scale(velocity_x, velocity_y, velocity_z,
...
-
-    Returns
-    -------
-    delta_ra, delta_dec: float
-        The offset to be added to the input RA/Dec, in units of radians.
     """
+    beta = np.array([velocity_x, velocity_y, velocity_z]) / SPEED_OF_LIGHT
+    beta2 = np.dot(beta, beta)
+    if beta2 == 0.0:
+        return 1.0, targ_ra, targ_dec
```
hbushouse, line 69: The old version of the code issued a warning when the velocity is zero. That would be good to maintain here, just for traceability in the processing log.

CHANGES.rst
```diff
@@ -12,6 +12,10 @@ lib
 
 - Make EngDB_Value public for JSDP use [#5663]
 
+- Update code in ``set_velocity_aberration.py`` functions based on Colin Cox
+  suggestions (simplify DVA scale computation and improve offset
+  computation). [#5666]
+
```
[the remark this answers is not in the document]
mcara: yes, I realized that too late. I'll add more to the description.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1293,15 +1293,15 @@ properties:
         title: Velocity aberration correction information
         type: object
         properties:
-          ra_offset:
-            title: "[rad] Velocity aberration correction RA offset"
+          va_ra_ref:
```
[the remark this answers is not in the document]
mcara: > So unless there RA, Dec values are at the reference point ...

It seems that they are. Take a look at the original code: https://github.com/spacetelescope/jwst/blob/1f4db5cffbf5df050220eb91b9da20cee27149f2/jwst/lib/set_velocity_aberration.py#L143-L150

In the script in the functions that compute VA effects, input arguments are called `targ_ra` and `targ_dec` and I find this confusing. In the function that updates FITS header, `targ_ra=ra_ref`, ... I think it is important to preserve the _actual_ meaning of the keywords.

jwst/lib/set_velocity_aberration.py
mcara: The old warning `logger.warning('Speed is zero. Forcing scale to 1.0')` is interesting. When speed is 0 then mathematically scale is 1. There is nothing exceptional here. Saying the code is "forcing" scale to be 1 is kind of a forceful statement :) I think from the caller's perspective it should not matter how the code handles special cases. In fact, this `if` can be safely removed from the code if we are talking only about scale. It is here because of the aberrated RA and Dec - that formula should be taking a limit in mathematical sense when `v->0`. In addition, it is really an extreme check to make code work even when JWST is "launched" in a garage. We do not reaaly expect a flying JWST with speed 0.

However, if you are thinking of detecting case when software incorrectly passes 0 speed instead of `v != 0` - then I hope someone will be reading those log files. Otherwise I would suggest using an `assert` or raising some other exception.

mcara: Let me know what do you prefer.

ColinRCox:
A zero speed will not interfere with the scale calculation but the RA and Dec calculation will have a divide by zero problem. By the way beta2 is the square of the velocity so I would prefer to name it something like beta_squared to make it clear.

jwst/datamodels/schemas/core.schema.yaml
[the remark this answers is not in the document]
hbushouse: Ah ha, so it appears that these coords are in fact the RA, Dec of the reference point, which is not necessarily the same as the coords of the target (if the target isn't at the reference point). Yes, it is unfortunate and confusing when the variable names used are "targ_ra" and "targ_dec", which makes you think it's the same as the TARG_RA, TARG_DEC keywords, but in fact they correspond to keywords RA_REF, DEC_REF. So in that case I withdraw my previous comment and think it's OK to leave the "ref" suffix on the meta attribute names. Too bad the stupid FITS standard won't allow us the room to add "REF" to the keyword names too.

mcara:
@hbushouse I addressed (locally) most comments. I have a question about the use of `REF(ref)`. Are you OK with `va_ref_ra` (or maybe make it `va_ra_ref` to make it consistent with `wcsinfo.ra_ref`)? Also, what about changing FITS keywords from `VA_RA` and `VA_DEC` to `VA_RFDEC` (the most I can fit in 8 characters) or `VA_RDEC`?

The log message is back but modified:

```python
logger.warning('Observatory speed is zero. Setting VA scale to 1.0')
```

jwst/datamodels/schemas/core.schema.yaml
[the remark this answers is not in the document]
mcara: I was writing https://github.com/spacetelescope/jwst/pull/5666#issuecomment-768484499 in parallel with you. I was suggesting to add at least an `R` in front of `RA->RRA` and `DEC->RDEC` just to alert readers and make it obvious that something is different about this RA and DEC.

hbushouse (approved):

ColinRCox:
The formula for computing the scale is exact. It is fairly simple but no higher terms have been omitted.

PR #5669, opened 2021-01-26
JP-1875: Make _get_wcs_values_from_siaf function public for JSDP use

### LINKED ISSUE #5663: Make _get_wcs_values_from_siaf Function Public
_Issue [JP-1875](https://jira.stsci.edu/browse/JP-1875) was created on JIRA by [Jesse Doggett](https://jira.stsci.edu/secure/ViewProfile.jspa?name=doggett):_

 SDP code to support moving targets needs to call the jwst.lib.set_telescope_pointing._get_wcs_values_from_siaf. However, since it is private with the leading underscore, the function is duplicated in the SDP code. If it could be made public, then the duplicate code in SDP could be removed.

This ticket will be done when _get_wcs_values_from_siaf has been renamed to get_wcs_values_from_siaf.

tapastro:
Resolves [JP-1875](https://jira.stsci.edu/browse/JP-1875)
Resolves #5663 

Also changed _EngDB_Value to EngDB_Value, per ticket comment request.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,8 +1,15 @@
 0.18.4 (unreleased)
 ===================
 
+lib
+---
 
+- Make EngDB_Value public for JSDP use [#5663]
 
+set_telescope_pointing
+----------------------
+
+- Make get_wcs_values_from_siaf public for JSDP use [#5663]
 
 
 0.18.3 (2021-01-25)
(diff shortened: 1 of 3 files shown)


hbushouse (approved):
LGTM. Just needs a change log entry.

jbdoggett (approved):

PR #5671, opened 2021-01-26
JP-1878: Update srctype Step with new default SRCTYPE for NRS_FIXEDSLIT 

### LINKED ISSUE #5668: Update srctype step to set non-primary NRS slits to EXTENDED
_Issue [JP-1878](https://jira.stsci.edu/browse/JP-1878) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

As indicated in JP-1150 and JP-1576, the default source type (SRCTYPE) for all slits other than the primary slit in a NIRSpec FIXEDSLIT exposure should be EXTENDED, rather than POINT, because they will almost always contain background. The srctype step needs to be updated accordingly.

tapastro:
Resolves [JP-1878](https://jira.stsci.edu/browse/JP-1878)
Resolves #5668 

Updated srctype step for NRS_FIXEDSLIT default SRCTYPE to EXTENDED rather than POINT, for non-primary slits.


hbushouse (approved):
LGTM.

hbushouse:
Whoops - need unit test updates, now that we've changed the behavior of the code.

tapastro:
> Whoops - need unit test updates, now that we've changed the behavior of the code.

If I read test_nrs_fixedslit correctly, the primary slit is S200A1, which has a user/test-defined srctype of 'EXTENDED'. The code change should now assert that non-primary slits are 'EXTENDED' as well, rather than the past test assertion of 'POINT'. Let me know if I've done this correctly! My assertion is not made with 100% confidence.

jwst/srctype/tests/test_srctype.py
```diff
@@ -175,9 +175,9 @@ def test_nrs_fixedslit():
 
     result = srctype.set_source_type(input)
 
-    assert(result.slits[0].source_type == 'POINT')
+    assert(result.slits[0].source_type == 'EXTENDED')
     assert(result.slits[1].source_type == 'EXTENDED')
-    assert(result.slits[2].source_type == 'POINT')
+    assert(result.slits[2].source_type == 'EXTENDED')
```
jdavies-st, line 180: Not your mistake, but `assert` is a statement, not a function, and does not need parenthesis. So:

```python
assert result.slits[0].source_type == 'EXTENDED'
assert result.slits[0].source_type == 'EXTENDED'
assert result.slits[0].source_type == 'EXTENDED'
```
is preferable

hbushouse: I like the idea of changing the unit test so that the primary slit is now POINT, just to make it different from the others.

hbushouse:
>  My assertion is not made with 100% confidence.
I see what you did there ...

Should also check the docs for the srctype step, to see if there's any mention of the default for NRS_FIXEDSLIT.

hbushouse (approved):

PR #5676, opened 2021-01-26
JP-1667: Update documentation for optimal weighting in ramp_fitting

### LINKED ISSUE #5300: Unclear definition of "weighting" in ramp_fitting step from docs
_Issue [JP-1667](https://jira.stsci.edu/browse/JP-1667) was created on JIRA by [Nestor Espinoza](https://jira.stsci.edu/secure/ViewProfile.jspa?name=nespinoza):_

It is currently unclear from the documentation what the "weighting" parameter in the ramp_fitting step is doing. Naively, "weighting" might be associated with performing an estimate of the slope weighting for the errors, but that wouldn't formally be Ordinary Least Squares which is the default method when running the ramp_fitting step (ols).

This is important to understand because, in general, for exoplanet time-series observations it seems that setting "weighting = unweighted" gives much better results than the default "weighted = optimal" version. The attached slides show an example of this with a NIRCam simulation analyzed by [Nikolay Nikolov](https://jira.stsci.edu/secure/ViewProfile.jspa?name=nnikolov).

Based on tests made in [this notebook](https://github.com/nespinoza/jwst_pipeline_testing/blob/master/Stage%201%20testing%20-%20NIRISS%20SOSS.ipynb) (which were made with NIRISS/SOSS simulations), my hunch is that the "weighting" parameter is just defining what to do in cases when there is a DQFLAG different than zero (with "unweighted" meaning nothing is done). Whatever the case, it might be perhaps optimal to have "weighted = unweighted" as the default for TSO observations, as with the typical small number of groups per integration (and hence, frame-times per group), it might be more likely that jumps are actually due to statistical fluctuations and not due to cosmic rays. We will discuss this anyways on our next TSO WG meeting before opening an "Improvement" ticket on this matter. We want this ticket to be solely about defining what the "weighting" parameter actually does.

tapastro:
Resolves [JP-1667](https://jira.stsci.edu/browse/JP-1667)
Resolves #5300 

Added an explanation of the optimal weighting algorithm to the documentation for the ramp_fitting step.


CHANGES.rst
```diff
@@ -12,6 +12,11 @@ lib
 
 - Make EngDB_Value public for JSDP use [#5663]
 
+ramp_fitting
+------------
+
+- Update documentation to define optimal weighting algorithm [#5300]
```
hbushouse, line 18: Actual PR number is 5676, not 5300.

hbushouse (approved):
Other than the PR number in the change log the doc updates themselves look good.

CHANGES.rst
tapastro: May or may not have been listing issue numbers in the changelog instead of PR numbers... 😬 fixing now.

tapastro:
Closing this ticket to make one with updated PR number fixes for recent PRs.

PR #5677, opened 2021-01-26
TST: Action needs long hash now

pllim:
Action needs long hash now. https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions

You merged #5675 too fast... 😆

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     # This should only run if we did not skip CI
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@ce17749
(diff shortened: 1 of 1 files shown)


jdavies-st:
Why point to a hash instead of a tag?  Or `master`?

pllim:
Because for this particular Action, it doesn't come from myself or a trusted organization, so I am following advise from https://julienrenaux.fr/2019/12/20/github-actions-security-risk/

pllim:
p.s. I did attempt to vet the code that was published at that hash and I didn't find anything fishy, but I cannot guarantee it going forward, so the pin.

jdavies-st (approved):

pllim:
p.s. Though just yesterday, I also found out about token limitations and that this cancelling is not working for PRs anyway, see astropy/astropy#11276 . I am still chatting with GitHub help desk and have not reached a conclusion yet on how to follow up. FYI.

Issue #3874, opened 2019-08-01
Formalize StepParsModel


stscijgbot:
Issue [JP-904](https://jira.stsci.edu/browse/JP-904) was created by Jonathan Eisenhamer:

{{StepParsModel}} was created to help resolve jp-875 and jp-876. It was based reference file schema and discussions with CRDS about required keywords. However, it should be reviewed for completeness.

There is also an issue: The {{StepParsModel.parameters.class}} attribute, required by the current stpipe architecture, cannot be used; a Python protected attribute issue. The value itself can still be accessed through {{StepParsModel.parameters.instance['class']}}, and it is preserved through writing/reading of the model. However, this may be an issue if the model is used for CRDS.

nden:
@stscieisenhamer With the work in #5678 I think this can be closed?

stscijgbot:
Comment by Jonathan Eisenhamer: Agreed, since such a thing no longer exists, this can go!

PR #5678, opened 2021-01-27
Replace StepParsModel with non-DataModel-derived StepConfig

eslavich:
This drops StepParsModel and replaces it with a class called StepConfig.  I also changed the format of the ASDF config file (the old ASDF configs should still read correctly).

The sdp-requirements.txt job is failing because this implementation requires the fabled asdf 2.8.


jwst/stpipe/config.py
```diff
@@ -0,0 +1,217 @@
...
+        StepConfig schema URI (must be registered with the ASDF entry point).
+
+    Raises
+    ------
+    asdf.ValidationError
+    """
+    # TODO: We should add a method on AsdfFile to facilitate this:
+    schema = asdf.schema.load_schema(schema_uri, asdf_file.resolver)
+    tagged_tree = asdf.yamlutil.custom_tree_to_tagged_tree(asdf_file.tree, asdf_file)
+    asdf.schema.validate(tagged_tree, asdf_file, schema)
```
jdavies-st, line 217: If one loads this using the `custom_schema` kwarg when creating the `AsdfFile` instance, one can just call `af.validate()` to do the validation.

jwst/stpipe/resources/schemas/step_config-0.1.0.yaml
```diff
@@ -0,0 +1,23 @@
...
+description: >
+  File containing saved configuration for an stpipe Step
+  instance.
+
+type: object
+properties:
+  parameters:
+    type: object
+    properties:
+      class:
```
jdavies-st, line 15: `class` is a reserved word in Python and should not be used.  Though clearly if we change it, there will be a small amount of backwards incompatibility.

jwst/stpipe/resources/schemas/step_config-1.0.0.yaml
```diff
@@ -0,0 +1,29 @@
...
+$schema: http://stsci.edu/schemas/yaml-schema/draft-01
+id: asdf://stsci.edu/stpipe/schemas/step_config-1.0.0
+title: Calibration pipeline Step config file
+description: >
+  File containing saved configuration for an stpipe Step
+  instance.
+
+type: object
+properties:
+  class:
```
jdavies-st, line 12: Again, we shouldn't use `class` as a name here.

setup.cfg
```diff
@@ -28,11 +28,13 @@ python_requires = >=3.7
 setup_requires =
     setuptools_scm
 install_requires =
-    asdf>=2.7.1
+    # asdf 2.8.0 is unreleased so it needs to be installed from master:
+    asdf@git+https://github.com/asdf-format/asdf
```
jdavies-st, line 32: This will commit us to releasing `asdf 2.8` before we can do another official release of `jwst`.

jdavies-st (commented):
Looking good.  Some comments below.

jwst/stpipe/config.py
eslavich: That's true.  But this module didn't necessarily load the file and doesn't have control over what `custom_schema` was specified.

jwst/stpipe/resources/schemas/step_config-0.1.0.yaml
eslavich: Why is that relevant in an ASDF file?  This gets mapped to the string key "class" in a dict.

setup.cfg
eslavich: @nden do you think we'll have time to get that asdf release out?  I estimate I'll need about a week's worth of mopping up before the code is ready to go, plus some additional time for thorough testing.

jwst/stpipe/resources/schemas/step_config-0.1.0.yaml
jdavies-st: Ah true.  When this was a datamodel, it meant that it had an attribute named `class` which caused problems in testing.  See #3874.

Carry on.

stscieisenhamer (approved):

setup.cfg
eslavich: I'm going to replace this with an asdf 2.7 extension, since it's straightforward given that we only have schemas to add and no tags, and gives us one less thing to worry about.

PR #5685, opened 2021-01-28
JP-1880: added source_type, x & y extraction center to extracted 1d spec for IFU data

### LINKED ISSUE #5674: SRCTYPE not propagating to x1d products for IFU modes
_Issue [JP-1880](https://jira.stsci.edu/browse/JP-1880) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The SRCTYPE keyword is not getting copied to the output "spec" datamodel within the ifu.py module of the extract_1d step, which means it doesn't show up in the x1d product headers anywhere. This is a problem for downstream data analysis tools that rely on that keyword.

jemorrison:
This address JP-1880 - adding SRCTYPE to x1d products for IFU data. In addition, the datamodels.multispec was added updated to include the x and y center of the extraction region:  extraction_x & extraction_y. 

Fixes #5674 / [JP-1880](https://jira.stsci.edu/browse/JP-1880)

### DIFF
--- a/jwst/datamodels/schemas/slitmeta.schema.yaml
+++ b/jwst/datamodels/schemas/slitmeta.schema.yaml
@@ -79,6 +79,18 @@ properties:
     default: 0.0
     fits_keyword: SRCYPOS
     fits_hdu: SCI
+  extr_x:
+    title: X center of extraction region
+    type: number
+    default: 0.0
+    fits_keyword: EXTR_X
+    fits_hdu: SCI
+  extr_y:
+    title: Y center of extraction region
+    type: number
+    default: 0.0
+    fits_keyword: EXTR_Y
+    fits_hdu: SCI
   shutter_state:
     title: All (open and closed) shutters in a slit
     type: string
--- a/jwst/datamodels/schemas/slitmeta.schema.yaml
+++ b/jwst/datamodels/schemas/slitmeta.schema.yaml
@@ -79,18 +79,6 @@ properties:
     default: 0.0
     fits_keyword: SRCYPOS
     fits_hdu: SCI
-  extr_x:
-    title: X center of extraction region
-    type: number
-    default: 0.0
-    fits_keyword: EXTR_X
-    fits_hdu: SCI
-  extr_y:
-    title: Y center of extraction region
-    type: number
-    default: 0.0
-    fits_keyword: EXTR_Y
-    fits_hdu: SCI
   shutter_state:
     title: All (open and closed) shutters in a slit
     type: string
--- a/jwst/extract_1d/ifu.py
+++ b/jwst/extract_1d/ifu.py
@@ -9,7 +9,8 @@ from .apply_apcorr import select_apcorr
 from ..assign_wcs.util import compute_scale
 
 from .. import datamodels
-from ..datamodels import dqflags
+from ..datamodels import dqflags, SpecModel
+
 from . import spec_wcs
 from scipy.interpolate import interp1d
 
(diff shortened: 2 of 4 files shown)


jwst/datamodels/schemas/slitmeta.schema.yaml
```diff
@@ -79,6 +79,18 @@ properties:
     default: 0.0
     fits_keyword: SRCYPOS
     fits_hdu: SCI
+  extr_x:
```
hbushouse, line 82: These keywords will only ever appear in extracted (x1d) products, hence they don't need to be in this schema. This schema is only used for pre-extraction products.

jwst/extract_1d/ifu.py
```diff
@@ -1064,3 +1076,26 @@ def shift_ref_image(mask, delta_y, delta_x, fill=0):
     temp[..., oslice_y, oslice_x] = mask[..., islice_y, islice_x]
 
     return temp
+
+
+def copy_keyword_info_ifu( source_type, x_center, y_center, slitname , spec: SpecModel):
```
hbushouse, line 1081: I guess the concept of this function is OK, but its version in the module for non-IFU data is specifically copying meta entries from the input `SlitModel` to the output `SpecModel`. Here you're just copying individual, independent variable values that aren't part of a model into `SpecModel` and hence you're forced to send all of those variables into the function as arguments. So I'm not really sure it's worth the effort of having a separate function to do this, as opposed to just having lines 1099-1101 appear up above at line 158.  But I don't have a strong feeling either way. It's OK to keep it too.

jwst/extract_1d/ifu.py
```diff
@@ -1064,3 +1076,26 @@ def shift_ref_image(mask, delta_y, delta_x, fill=0):
...
+        Metadata will be copied from the input IFUcube to output `spec`.
+
+    slitname : str or None
+        The name of the slit.
+
+    spec : One element of MultiSpecModel.spec
+        Metadata attributes will be updated in-place.
+
+    """
+    if slitname is not None and slitname != "ANY":
```
hbushouse, line 1096: I think this duplicates what's already been done up above around line 156 right before this function is called.

hbushouse:
The description box above needs to be populated with an actual description of what's been done and why. It also needs to have references to ticket JP-1880 and the associated github issue number.

jwst/extract_1d/ifu.py
jemorrison: I removed it. At first I thought we might want to copy a bunch of keywords coming from different locations so I modeled
it after what was in extract.py. But it is just a few simple lines of code now. So it just added unneeded complexity.

jwst/datamodels/schemas/slitmeta.schema.yaml
jemorrison: Ok thanks for the clarification. I was misunderstanding what the slitmeta schema was

jemorrison:
Something is not correct. The extract1d output has an empty SCI extension. I need to work on this issue some more. Do not merge this

jwst/extract_1d/ifu.py
```diff
@@ -663,6 +673,9 @@ def image_extract_ifu(input_model, source_type, extract_params):
     n_bkg : ndarray, 1-D, float64
         For each slice, this is the number of pixels that were added
         together to get background.
+
+    x_center, y_center : float
```
hbushouse, line 677: These have been added to the "Returns" list in the doc string here, but they aren't actually returned by the function.

hbushouse:
Unit tests are failing due to the additional arguments added to the function call.

jwst/extract_1d/ifu.py
jemorrison: hah that was an oversight. That function is used in the pipeline but I guess it is in the tests.

hbushouse:
CI flake8 errors are in another module (conftest), so I think this PR is good to go.

hbushouse (approved):

PR #5692, opened 2021-02-01
JP-1805 Fix premature model closing in group_scale_step

stscieisenhamer:
Partially resolves [JP-1805](https://jira.stsci.edu/browse/JP-1805)
Blocked by [AL-456](https://jira.stsci.edu/browse/AL-456)

Main issue is inappropriate context manager usage in `group_scale_step`.

Secondary issue is that `stdatamodels.model_base.DataModel` does not manage opened file resources across shallow copies.

### DIFF
--- a/jwst/group_scale/group_scale.py
+++ b/jwst/group_scale/group_scale.py
@@ -9,7 +9,7 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 
-def do_correction(input_model):
+def do_correction(model):
     """
     Short Summary
     -------------
@@ -22,34 +22,26 @@ def do_correction(input_model):
 
     Parameters
     ----------
-    input_model: data model object
-        science data to be corrected
-
-    Returns
-    -------
-    output_model: data model object
-        rescaled science data
-
+    model: data model object
+        science data to be corrected. Model is modified in place
+        but also returned.
     """
 
     # Get the meta data values that we need
-    nframes = input_model.meta.exposure.nframes
-    frame_divisor = input_model.meta.exposure.frame_divisor
+    nframes = model.meta.exposure.nframes
+    frame_divisor = model.meta.exposure.frame_divisor
     if (nframes is None) or (frame_divisor is None):
         log.warning('Necessary meta data not found')
         log.warning('Step will be skipped')
-        input_model.meta.cal_step.group_scale = 'SKIPPED'
-        return input_model
-
-    # Create output as a copy of the input science data model
-    output_model = input_model.copy()
+        model.meta.cal_step.group_scale = 'SKIPPED'
+        return
 
     log.info('NFRAMES={}, FRMDIVSR={}'.format(nframes, frame_divisor))
     log.info('Rescaling all groups by {}/{}'.format(frame_divisor, nframes))
 
     # Apply the rescaling to the entire data array
     scale = float(frame_divisor) / nframes
-    output_model.data *= scale
-    output_model.meta.cal_step.group_scale = 'COMPLETE'
+    model.data *= scale
+    model.meta.cal_step.group_scale = 'COMPLETE'
 
-    return output_model
+    return
--- a/jwst/group_scale/group_scale.py
+++ b/jwst/group_scale/group_scale.py
@@ -23,8 +23,7 @@ def do_correction(model):
     Parameters
     ----------
     model: data model object
-        science data to be corrected. Model is modified in place
-        but also returned.
+        science data to be corrected. Model is modified in place.
     """
 
     # Get the meta data values that we need
--- a/jwst/group_scale/group_scale_step.py
+++ b/jwst/group_scale/group_scale_step.py
@@ -52,6 +52,6 @@ class GroupScaleStep(Step):
                 return result
 
             # Do the scaling
-            result = group_scale.do_correction(result)
+            group_scale.do_correction(result)
 
         return result
--- a/jwst/group_scale/group_scale_step.py
+++ b/jwst/group_scale/group_scale_step.py
@@ -18,18 +18,21 @@ class GroupScaleStep(Step):
         # Open the input data model
         with datamodels.RampModel(input) as input_model:
 
+            # Always work on a copy
+            result = input_model.copy()
+
             # Try to get values of NFRAMES and FRMDIVSR to see
             # if we need to do any rescaling
-            nframes = input_model.meta.exposure.nframes
-            frame_divisor = input_model.meta.exposure.frame_divisor
+            nframes = result.meta.exposure.nframes
+            frame_divisor = result.meta.exposure.frame_divisor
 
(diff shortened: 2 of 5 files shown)


hbushouse (commented):
Looks good. Needs a change log entry.

And just for my own edification, what was the problem with using the "with" construct in the process method? Is it because there are several opportunities for it to skip out and return early and doing so doesn't close the input_model?

jdavies-st (commented):
Same question as Howard.  What does removing the `with` context manager solve?

The problem with shallow copies has been addressed in 

https://github.com/spacetelescope/stdatamodels/pull/59
https://github.com/spacetelescope/stdatamodels/pull/60

Does this not solve the issue here?

What's unique about `GroupScaleStep` that's different from other pipeline steps that makes us not want to use the `with` context manager?

jwst/group_scale/group_scale_step.py
```diff
@@ -16,39 +16,39 @@ class GroupScaleStep(Step):
...
-            nframes = input_model.meta.exposure.nframes
-            frame_divisor = input_model.meta.exposure.frame_divisor
-
-            # If we didn't find NFRAMES, we don't have enough info
-            # to continue. Skip the step.
-            if nframes is None:
-                self.log.warning('NFRAMES value not found')
-                self.log.warning('Step will be skipped')
-                input_model.meta.cal_step.group_scale = 'SKIPPED'
-                return input_model
```
eslavich, line 32: The context manager was removed so that these early returns wouldn't return a closed model, right?  Maybe we should return `input_model.copy()` instead of removing the context manager?  Otherwise `input_model` doesn't get closed if we get to `return result` at the end.

stscieisenhamer:
Note first that he file management issue fixed in stdatamodels does not address this issue, though it does make this even more of an issue, since files are now being properly managed, well with some caveats to be discussed future-wise.

This issue is just a bad use of context management: The result, having been created by the context manager, is what is being returned. However, as soon as the return executes, the result becomes invalid, since it is leaving the context manager. One can demonstrate this with the example below. 

And yes, this has been a problem in the past and fixed. At the moment, other known uses of context management are OK, since, at some point in the context, the result returned is copied from the context-managed resource. 

Fail Example:
```
from contextlib import contextmanager

@contextmanager
def athing():
    thing = {'value': "a value", 'valid': True}
    try:
        yield thing
    finally:
        thing['valid'] = False

def working_athing():
    with athing() as mything:
        return mything

outside_thing = working_athing()
print(outside_thing)
=> {'value': 'a value', 'valid': False}
```

jwst/group_scale/group_scale_step.py
jdavies-st: This is exactly my thought too.

Generally, pipeline steps should not be modifying their inputs at all, which means we should be returning a deep copy of the input here.  And actually copying it early, before metadata is modified.

jdavies-st:
Also, if this is a general problem with this step, why haven't our unit or regression tests picked it up?  Are users running this in a way we are not?

jwst/group_scale/group_scale_step.py
stscieisenhamer: Forgot about that point and agree. Will open-then-copy.

stscieisenhamer:
About why this hasn't been seen: There is an exposure and regression added that does hit the bug; file is now on artifactory. If I recall, I believe there is/are other data array(s) in the example that the equivalent data, already in the regression, do not have. The theory is that these extra arrays are not directly accessed. As such, the current premature closing of the input later causes the file-not-open issue when those arrays are accessed, usually when results are written to files.

jwst/group_scale/group_scale_step.py
```diff
@@ -18,37 +18,40 @@ def process(self, input):
         # Open the input data model
         with datamodels.RampModel(input) as input_model:
 
+            # Always work on a copy
+            result = input_model.copy()
```
eslavich, line 22: Plagiarism!

jdavies-st (approved):
Looks good!

jwst/group_scale/group_scale_step.py
```diff
@@ -18,37 +18,40 @@ def process(self, input):
...
                 self.log.info('NFRAMES and FRMDIVSR are equal; correction not needed')
                 self.log.info('Step will be skipped')
-                input_model.meta.cal_step.group_scale = 'SKIPPED'
-                return input_model
+                result.meta.cal_step.group_scale = 'SKIPPED'
+                return result
 
             # Do the scaling
-            result = group_scale.do_correction(input_model)
+            result = group_scale.do_correction(result)
```
eslavich, line 55: We make another copy inside of `group_scale.do_correction`:

https://github.com/spacetelescope/jwst/blob/master/jwst/group_scale/group_scale.py#L45

Should we skip that now?  Otherwise the result from this method will get overwritten by the result from `do_correction`, and the original result won't get closed.

jwst/group_scale/group_scale_step.py
eslavich: Ironically, I had to copy and paste the word "plagiarism" from a dictionary site in order to get the spelling correct

hbushouse (approved):

jwst/group_scale/group_scale_step.py
stscieisenhamer: Agree to remove that copy; there is definitely enough copying going around.

However, a future discussion point: At this level, the code should not care whether a DataModel is closed or not: the user-level api should simply presume python scoping is in effect, end-of-story. If insufficient, that is actually a bug in the internals of DataModel, not the user code.

And FYI on this particular case: For the copied `result`, there would be nothing to close anyways. `DataModel.copy` is a deep copy (had to look it up), and no files are involved, so `DataModel.close` is a no-op, the in-memory stuff is still around, and won't be taken care of until gc-ed anyways.

eslavich: Oops, right!  Still can't keep it straight which one is deep copy and which is shallow 🤦

jdavies-st: Yeah, I think the choice of names could have been done wiser in the past.

jwst/group_scale/group_scale.py
```diff
@@ -22,34 +22,26 @@ def do_correction(input_model):
...
-        science data to be corrected
-
-    Returns
-    -------
-    output_model: data model object
-        rescaled science data
-
+    model: data model object
+        science data to be corrected. Model is modified in place
+        but also returned.
```
eslavich, line 27: "but also returned" is no longer accurate?

eslavich (approved):
One possible issue with a docstring but feel free to merge after you dispense with that

jwst/group_scale/group_scale.py
stscieisenhamer: bah! forgot to remove that...

PR #5696, opened 2021-02-04
JP-1856: Trap zero-weight columns in extract_1d

### LINKED ISSUE #5625: extract_1d crashing in bkg fitting for garbage NIRISS SOSS data
_Issue [JP-1856](https://jira.stsci.edu/browse/JP-1856) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The extract_1d step in calwebb_tso-spec2 is crashing for a NIRISS SOSS exposure that's filled with garbage (OTB) data. Lots of processing warning messages about too few valid background pixels at a given wavelength (over and over and over again) and then ultimately dies completely in one of the astropy fitting routines.

Low priority to fix, due to garbage data, but worth looking into to make sure background fitting is robust at the level of, say, no valid pixels at a particular wavelength, and what happens with the background value for such instances.

hbushouse:
Updated the `fit_background_model` function in extract1d.py to check for zero weights before calling `LinearLSQFitter` to fit background values in a given image column. When the total weight of all pixels is zero, it causes an error return from the fitter.

Also includes some flake8 clean-up and other cosmetic changes.

Fixes #5625 / [JP-1856](https://jira.stsci.edu/browse/JP-1856)

### DIFF
--- a/jwst/extract_1d/extract1d.py
+++ b/jwst/extract_1d/extract1d.py
@@ -21,6 +21,7 @@ __author__ = 'Mihai Cara'
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
+
 def extract1d(image, lambdas, disp_range,
               p_src, p_bkg=None, independent_var="wavelength",
               smoothing_length=0, bkg_fit="poly", bkg_order=0, weights=None):
@@ -195,7 +196,7 @@ def extract1d(image, lambdas, disp_range,
         temp_image = image
 
     #################################################
-    ##         Perform spectral extraction:        ##
+    #          Perform spectral extraction:         #
     #################################################
 
     bkg_model = None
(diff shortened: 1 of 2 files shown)


jwst/extract_1d/extract1d.py
```diff
@@ -452,6 +454,9 @@ def _fit_background_model(image, x, j, bkglim, bkg_fit, bkg_order):
     wht = wht[good]
     y = y[good]
 
+    if wht.sum() == 0:
```
hbushouse, line 457: Reviewers: This is the only substantive change to fix the bug. All other changes are purely cosmetic.

tapastro: Do you think a log message/warning is in order, or should this effectively never run unless the data aren't reasonable?

tapastro (approved):
LGTM, with a suggestion of logging/warning if necessary.

jwst/extract_1d/extract1d.py
hbushouse: Yes, both here and in line 450 (where npts is insufficient) this function returns 0 for the number of bkg region points used, which in turn triggers a warning message in the calling function.

tapastro: Ah, I didn't do due diligence on tracing what the zeroed-out polynomial fit would cause downstream. Sounds good!

Issue #5480, opened 2020-11-13
Convert ModelContainer to a container type


stscijgbot-jp:
_Issue [JP-1784](https://jira.stsci.edu/browse/JP-1784) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

Currently `ModelContainer` subclasses `DataModel` and has a schema, but neither of these things are necessary, as it is a container of datamodels, not a datamodel itself.  And the schema does nothing, as the container itself never gets serialized, only the datamodels it contains do (and those have schemas).  See

#1251

So we need an initial implementation of `ModelContainer` that does not subclass datamodels, but has most of the same functionality.  It should conform to the `collections.abc.MutableSequence` container object I suspect.

In so doing, we can move ModelContainer into `stdatamodels` and allow it to be used by `romancal`.

It seems likely that JWST-specific functionality, for now, will have to remain in the `jwst` package. This includes reading in associations.

This may require some changes to step functions that check if something is a `DataModel` instance.  If these steps do accept containers, they should only be checking that it conforms to `collections.abc.MutableSequence` at a minimum, or that it is an `stdatamodels.ModelContainer` if that's necessary.

The goal here is to make `stpipe` not depend on `jwst.datamodels`.

stscijgbot-jp:
_Comment by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva) on [JIRA](https://jira.stsci.edu/browse/JP-1784?focusedCommentId=648683#comment-648683):_

Resolved in #6094

PR #5697, opened 2021-02-04
Break stpipe dependence on ModelContainer

eslavich:
It turns out, stpipe only uses ModelContainer in its capacity as a sequence -- iterating over the component models and performing operations on them.  This PR makes ModelContainer a subclass of collections.abc.Sequence and changes stpipe to perform `isinstance(foo, Sequence)` checks instead of `isinstance(foo, ModelContainer)`.  This should be enough to allow us to leave ModelContainer behind when we move stpipe to an independent package.

Regression test run: https://plwishmaster.stsci.edu:8081/job/RT/job/eslavich-dev/59

### DIFF
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -1,5 +1,6 @@
 import copy
 from collections import OrderedDict
+from collections.abc import Sequence
 import os.path as op
 import re
 import logging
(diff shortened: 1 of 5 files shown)


jdavies-st (approved):
Well, this is fantastically simple and solves the issue without having to really change `ModelContainer` at all.  The class already implements the minimum methods to be a `Sequence` subclass.  We will want it to be a `MutableSequence`, but that won't affect `stpipe` code at all.

This is really great.  🚀

jdavies-st:
Btw, this is the minimal first step in solving #5480 / [JP-1784](https://jira.stsci.edu/browse/JP-1784).

Step 2 will be to make `ModelContainer` subclass `MutableSequence`, which will simplify the `datamodels.ModelContainer` code a bit more.

Step 3 will be to peel apart the JWST-specific part (if, in fact, any JWST-specific part is even needed) and put the general code in `stdatamodels`.

PR #5698, opened 2021-02-04
Use parent datamodels_open method if available

eslavich:
This is a sneaky dependence on `JwstStep` that I hadn't noticed before.  I updated `Step.datamodels_open` to use the parent class's implementation if available, which allows the hook subclasses to be mission-agnostic.

### DIFF
--- a/jwst/stpipe/function_wrapper.py
+++ b/jwst/stpipe/function_wrapper.py
@@ -29,7 +29,7 @@
 """
 A Step whose only purpose is to wrap an ordinary function.
 """
-from . import Step
+from .step import Step
 
 
 class FunctionWrapper(Step):
--- a/jwst/stpipe/step.py
+++ b/jwst/stpipe/step.py
@@ -757,7 +757,7 @@ class Step:
         # If the dataset is not an operable DataModel, log as such and return
         # an empty config object
         try:
-            model = cls.datamodels_open(dataset)
+            model = cls._datamodels_open(dataset)
         except (IOError, TypeError, ValueError):
             logger.warning('Input dataset is not a DataModel.')
             disable = True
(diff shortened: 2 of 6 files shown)


jwst/stpipe/function_wrapper.py
```diff
@@ -29,7 +29,7 @@
 """
 A Step whose only purpose is to wrap an ordinary function.
 """
-from . import Step
+from .step import Step
```
jdavies-st, line 32: This is to get the base `Step` class instead of the renamed `JwstStep` class?

jwst/stpipe/step.py
```diff
@@ -1093,12 +1092,12 @@ def closeout(self, to_close=None, to_del=None):
...
         gc.collect()
 
-    @abc.abstractclassmethod
+    @classmethod
     def datamodels_open(cls, init, **kwargs):
         """
         Wrapper around observatory-specific datamodels.open function.
         """
-        pass
+        raise NotImplementedError(f"{cls.__name__} does not implement datamodels_open")
```
jdavies-st, line 1100: What is the purpose of `datamodels_open` class method, given that we already have `open_model`?  I see it is called by `open_model` method below, but is this supposed to be part of the top-level user interface for `Step`?

I guess I'm asking if it should be a private method instead?

jwst/stpipe/function_wrapper.py
eslavich: Yes, this is to better simulate how the module will behave once it's moved into the stpipe package.

jwst/stpipe/step.py
eslavich: `datamodels_open` is a class method, which the CLI code needs in order to open the dataset and request step parameters from CRDS.  I think it would be reasonable to make it private, at least until a user demonstrates a clear need, so I'll go ahead and do that.

jdavies-st (approved):
Looks good!

PR #5700, opened 2021-02-05
JP-1910: Fix ramp_fit handling of saturated ngroups=2 data

### LINKED ISSUE #5715: ramp_fit setting pixels to zero when only 1st group is good
_Issue [JP-1910](https://jira.stsci.edu/browse/JP-1910) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The ramp_fit step is setting the final slope to zero for pixels in an NGROUPS=2 exposure that have good data in the first group and saturated in the second group (i.e. the only good group is the first one). Because there's one good group the DO_NOT_USE flag is correctly not being set in the output, but that allows the zero SCI value to be used in downstream steps, which is not appropriate. A non-zero value for the slope should be computed from the 1st good group.

hbushouse:
Updated the low-level `fit_2_group` function in ramp_fitting to properly handle the situation where NGROUPS=2 and the group 2 is saturated, but group 1 is OK. Hence the group 1 value can be used to compute a countrate. The problem in the existing code was that the variance for a pixel like this was being set to 0 in `fit_2_group`, but when those results got returned to the calling function it was throwing out all results with variance=0, so slopes and other quantities were not getting computed.

I've simply set the variance to 1 in this function, which allows for the proper computation of slope, variances, and error once it returns to the calling routine.

Fixes #5715  / [JP-1910](https://jira.stsci.edu/browse/JP-1910)

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -2895,7 +2895,9 @@ def fit_2_group(slope_s, intercept_s, variance_s, sig_intercept_s,
     if (len(wh_sat1[0]) > 0):
         data0_slice = data[0, :, :].reshape(npix)
         slope_s[wh_sat1] = data0_slice[wh_sat1]
-        variance_s[wh_sat1] = 0.
+        # set variance non-zero because calling function uses variance=0 to
+        # throw out bad results; this is not bad
+        variance_s[wh_sat1] = 1.
         sig_slope_s[wh_sat1] = 0.
         intercept_s[wh_sat1] = 0.
         sig_intercept_s[wh_sat1] = 0.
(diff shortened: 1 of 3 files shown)


hbushouse:
Also added a new unit test to check this condition. In doing so I reworked the unit test framework a bit to allow for tests that use more than 1 pixel at a time.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -2895,7 +2905,9 @@ def fit_2_group(slope_s, intercept_s, variance_s, sig_intercept_s,
     if (len(wh_sat1[0]) > 0):
         data0_slice = data[0, :, :].reshape(npix)
         slope_s[wh_sat1] = data0_slice[wh_sat1]
-        variance_s[wh_sat1] = 0.
+        # set variance non-zero because calling function uses variance=0 to
```
hbushouse, line 2908: Note to reviewers: this is the only actual change in the code needed to fix the bug. All the rest is cosmetic.

tapastro (approved):
Just got to the email requesting review... LGTM.

stscieisenhamer (approved):

PR #5695, opened 2021-02-04
Merge stpipe-transition branch to master

eslavich:
This includes the following PRs:

https://github.com/spacetelescope/jwst/pull/5600
https://github.com/spacetelescope/jwst/pull/5645
https://github.com/spacetelescope/jwst/pull/5648
https://github.com/spacetelescope/jwst/pull/5650
https://github.com/spacetelescope/jwst/pull/5652
https://github.com/spacetelescope/jwst/pull/5665
https://github.com/spacetelescope/jwst/pull/5678
https://github.com/spacetelescope/jwst/pull/5697
https://github.com/spacetelescope/jwst/pull/5698

Regression test run is ~~pending~~ passing here: https://plwishmaster.stsci.edu:8081/job/RT/job/eslavich-dev/58

PR #5720, opened 2021-02-05
Integrate stpipe package

eslavich:
Here is the final volume in our [thrilling series](https://github.com/spacetelescope/jwst/pull/5695) on stpipe.  This adds the stpipe package as a dependency, removes code from jwst.stpipe that has been copied to the new package, and updates relevant imports.

We'll want to make a release of stpipe with a real version number before we merge this PR.

Regression test run on this branch: https://plwishmaster.stsci.edu:8081/job/RT/job/eslavich-dev/61


jwst/assign_wcs/util.py
```diff
@@ -40,25 +41,6 @@ def __init__(self, message):
...
-    configurations of the MSA, it is possible that no dispersed spectra will
-    appear on NRS2. This is not a failure of calibration, but needs to be
-    called out in order for the calling architecture to be aware of this.
-
-    """
-
-    def __init__(self, message=None):
-        if message is None:
-            message = 'WCS solution indicate that no science is in the data.'
-        super(NoDataOnDetectorError, self).__init__(message)
```
jdavies-st, line 59: This exception really is specific to JWST and `assign_wcs`.  Do you think it is worthwhile having it in `stpipe`?  Is there a reason it needs to be in there?

jdavies-st: Ah, I see it is used in `strun`.

https://github.com/spacetelescope/jwst/blob/0382b8dbd663c4c1b61f0f91a0402ce19fe3d881/scripts/strun#L28

But this is a very specific.  Would it be better to have a more general exception to return 64 to include in `strun` checking, include that in `stpipe`, and then have this very specific one subclass it for JWST use?

eslavich: Ah I didn't realize it was specific.  How about we keep the exception in jwst but make it subclass SystemExit:

```python
class NoDataOnDetectorError(SystemExit):
    def __init__(self):
        super().__init__(64)

raise NoDataOnDetectorError()
```
After running that in a script `$?` reports 64 as expected.

eslavich: I added `SystemExit` handling to stpipe in a new PR [here](https://github.com/spacetelescope/stpipe/pull/12).  I'll update this branch to add back in `NoDataOnDetectorError` and make it subclass `SystemExit`.

jdavies-st: Raising `SystemExit` I think is fine.  It is a little risky to `except SystemExit`, though in the case of the CLI it should fine I think.  👍

eslavich: You make a good point in the stpipe PR that we don't need to catch it anyway...

jdavies-st: Thinking about this a bit more, we generally only want the `strun` CLI to be raising `SystemExit`.  And this is specifically for DMS processing.  For general user say, running some NIRSpec data through a Jupyter notebook, it would be pretty annoying to get an unhandled `SystemExit` turning off their python session (killing their notebook kernel) every time there was a no data on one of the detectors.  🤔 

Thoughts @hbushouse and @nden ?

eslavich: Hmm... another fine point.  It doesn't exit a notebook kernel but you do get an ugly message:
image

eslavich: I added an `stpipe.exceptions.StpipeExitException` and switched the base class on `NoDataOnDetectorError` to that.

jdavies-st: That's a perfect solution.  🥇 

Yeah, raising `SystemExit` under the default python interpreter does exit:

```
$ python
Python 3.9.1 (default, Dec 11 2020, 06:28:49) 
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> raise SystemExit
$ 
```

jwst/assign_wcs/util.py
```diff
@@ -56,16 +55,11 @@ class NoDataOnDetectorError(SystemExit):
...
-        self._message = message
-
-    def __repr__(self):
-        return f"NoDataOnDetectorError({repr(self._message)})"
-
-    def __str__(self):
-        return self._message
+        # The first argument instructs stpipe CLI tools to exit with status
+        # 64 when this exception is raised.
+        super().__init__(64, message)
```
jdavies-st, line 62: This looks great.

jdavies-st (approved):
Looks good!  After `stpipe 0.1` is released, I think we can change the dependency in this PR and merge away.

eslavich:
Regression test run with release version of stpipe:

https://plwishmaster.stsci.edu:8081/job/RT/job/eslavich-dev/62

eslavich:
stpipe 0.1.0 is released and the final regression test run passed, so here we go!

PR #5727, opened 2021-02-08
Add dev dependencies workflow as weekly cron

jdavies-st:
We are not running unit tests using `requirements-dev.txt`.  This remedies that problem by running it at least weekly.  If it fails, it will not interfere with approval of PRs, though we will get emails about it.  And it can also be triggered manually if needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,8 +10,7 @@ on:
   pull_request:
   schedule:
     # Weekly Monday 9AM build
-    # * is a special character in YAML so you have to quote this string
-    - cron: '0 9 * * 1'
+    - cron: "0 9 * * 1"
 
 env:
   CRDS_SERVER_URL: https://jwst-crds.stsci.edu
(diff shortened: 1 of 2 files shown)


jdavies-st:
Example of this workflow running is here:

https://github.com/spacetelescope/jwst/runs/1990497333?check_suite_focus=true

where I temporarily turned it on for `pull_request`.  Oh look, there's a failure.

hbushouse (approved):

.github/workflows/ci.yml
```diff
@@ -20,9 +19,6 @@ env:
   CRDS_CLIENT_RETRY_DELAY_SECONDS: 20
 
 jobs:
-
-  # The rest only run if above are done
```
pllim, line 24: Ops, my bad for leaving this in. Thanks for cleaning up after me.

pllim (approved):
I don't grok the actual test goals but in principle, I don't see anything wrong. 👍 

p.s. I am not as bothered by that one failure as I am with the 110 warnings. 😆

PR #5730, opened 2021-02-09
JP-1778: Output segmentation map for source catalog in Image3

### LINKED ISSUE #5461: Output the segmentation map for the source catalog in Image3
_Issue [JP-1778](https://jira.stsci.edu/browse/JP-1778) was created on JIRA by [Swara Ravindranath](https://jira.stsci.edu/secure/ViewProfile.jspa?name=swara):_

The image3 step currently outputs a source catalog which uses the photutils segmentation for source identification. However, the pipeline does not save and output the segmentation map. The segmentation map is a very useful diagnostic for many purposes which includes checking the source parameters, identifying spurious sources, verifying spectral extraction boxes used in the Spec2 stage, and also for various post pipeline analysis. Users can run photutils offline with the parameters in the source_catalog.cfg file, but the source IDs may not match the ones from the pipeline catalog and the catalogs may not be identical (at least, based on past experience with SExtractor segmentation maps). This ticket is to ask for the pipeline to provide the segmentation maps associated with the source catalogs that are output by Image3 step.

The segmentation map will also be critical for WFSS contamination (JP-897)

tapastro:
Resolves [JP-1778](https://jira.stsci.edu/browse/JP-1778)
Resolves #5461 

This implements the basic functionality of outputting a fits file with suffix '_segm' from the source catalog step, which contains a single SCI extension holding the segmented image. I've updated the primary header of the _segm.fits file from the input model. 

Should a unit test be added? Will regression tests need to be updated?


stscieisenhamer:
at least a regression test. unit tests would be nice, but can be tricky due to need for and/or construct data.

jwst/source_catalog/source_catalog_step.py
```diff
@@ -119,4 +119,9 @@ def process(self, input_model):
                 model.meta.source_catalog = os.path.basename(cat_filepath)
                 self.log.info(f'Wrote source catalog: {cat_filepath}')
 
+                segm_model = datamodels.ImageModel(segment_img.data)
+                segm_model.update(model, only="PRIMARY")
+                self.save_model(segm_model, suffix='segm')
+                self.log.info('Wrote segmentation map to segm.fits.')
```
stscieisenhamer, line 125: Include the segment file name in the log message. After saving, it should be available in `segm_model.filename`.  May be `segm_model.meta.filename`.

stscieisenhamer (changes requested):
About testing: At least update a regression test. Suggest finding one that runs `Image3Pipeline`, and ensure that the respective truth folder has the new file.

Also update CHANGES.

jwst/source_catalog/source_catalog_step.py
```diff
@@ -119,4 +119,9 @@ def process(self, input_model):
                 model.meta.source_catalog = os.path.basename(cat_filepath)
```
hbushouse, line 119: Since we're saving the name of the source catalog in a header keyword of the image product, I feel that for completeness we should also save the name of the segmentation map file. We'll need to create a new keyword for this in `core.schema.yaml`, perhaps called `SEGMFILE` (similar to `SCATFILE`) and give it a datamodel name of `segmentation_map`. So would need to add a line here (or below) to populate that keyword.

hbushouse (commented):
Who knew it would be this simple? Looks good, but agree that regression tests should be updated (one or all of them that run calwebb_image3) to now also start checking the segm file output.

CHANGES.rst
```diff
@@ -44,6 +44,12 @@ ramp_fitting
 
 - Update documentation to define optimal weighting algorithm [#5682]
 
+source_catalog
```
hbushouse, line 47: Should also now add an entry to the change log under `datamodels`, because of the new SEGMFILE keyword that's being added.

hbushouse (approved):
Other than adding one more entry to the change log, it looks good to me now.

stscieisenhamer (approved):

PR #5734, opened 2021-02-09
TST: Remove custom CI skip

pllim:
See https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,19 +23,10 @@ jobs:
   cancel_ci:
     name: Mandatory checks before CI
     runs-on: ubuntu-latest
-    outputs:
-      run_next: ${{ steps.skip_ci_step.outputs.run_next }}
     steps:
-    - name: Set output for skip CI
-      uses: OpenAstronomy/action-skip-ci@main
-      id: skip_ci_step
-      with:
-        NO_FAIL: true
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    # This should only run if we did not skip CI
+    # TODO: Need to fix for pull_request
     - name: Cancel Previous Runs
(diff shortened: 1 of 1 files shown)


jdavies-st:
Whoa!  They listened to the users!

Now if they would just implement `allowed_failure`.  😅

pllim:
Bad news is that currently the built-in CI skipping skips everything in `.github/workflow` but that isn't a problem for your current use case.

jdavies-st:
Ah, so if you have a commit at HEAD that has `[ci skip]`, then it will skip the merged job on `master` plus any cron jobs?

jdavies-st:
Btw, is there a reason to keep the `cancel_ci` job in this workflow?  Or can all of that be removed now?

pllim:
Actually, I just merged the equivalent PR on `astropy` and it looks like the merge commit only has PR title, not the individual commit messages, so the merge ran the CI. I suspect cron would behave the same (I'll find out tomorrow). But if you have `[ci skip]` in the PR title, you deserve it (LoL).

Well, I have a fix in for cancelling at styfle/cancel-workflow-action#55 but it is just sitting there. You can choose to remove this altogether, keep it as is until my PR is merged (if ever), or use my forked branch; but I think that is out of scope here.

jdavies-st:
I see.  Cool.  We'll just leave it here then.

I suspect you would run into the problem if you did "Rebase and merge" which would put that rebased commit with `[ci skip]` in the title at the HEAD of master.  So I see where this could be a problem.  We had the same problem back when we had our own Jenkins per-PR CI.  Nothing changes.

jdavies-st (approved):
Looks good.  Thanks!

pllim:
Yeah... at this point, the rule is server side now and out of our control. The only thing we can do is workaround it.

PR #5737, opened 2021-02-11
Fix CRDS_CONTEXT setting in CI workflow

### LINKED ISSUE #5733: GitHub Actions reports syntax error in ci.yml
https://github.com/spacetelescope/jwst/blob/bf24eaeaa5825f95439956e4e053f51b49655fb6/.github/workflows/ci.yml#L67-L68

If you open that file for edit using GitHub UI, you will see a complain that saying something like "expect string, blah blah, but an object is found" -- maybe you cannot change env that way inside a matrix; not sure.

jdavies-st:
- Set CRDS_CONTEXT=jwst-edit in tox instead ci.yml
 - Add a header line to the pytest report to show the used CRDS_CONTEXT.

        $ pytest jwst/datamodels/tests/test_validation.py 
        ============================== test session starts ==============================
        platform darwin -- Python 3.9.1, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
        crds_context: jwst_0672.pmap
        rootdir: /Users/jdavies/dev/jwst, configfile: setup.cfg
        plugins: requests-mock-1.8.0, cov-2.11.1, ci-watson-0.5, xdist-2.2.1, asdf-2.7.2, openfiles-0.5.0, forked-1.3.0, doctestplus-0.9.0
        collected 3 items                                                               

        jwst/datamodels/tests/test_validation.py ...                              [100%]

        =============================== 3 passed in 0.81s ===============================

Resolves #5733.

### DIFF
--- a/jwst/conftest.py
+++ b/jwst/conftest.py
@@ -5,7 +5,6 @@ import pytest
 import inspect
 
 from stdatamodels import s3_utils
-from stpipe.crds_client import get_context_used
 
 from jwst.associations import (AssociationRegistry, AssociationPool)
 from jwst.associations.tests.helpers import t_path
@@ -112,8 +111,3 @@ class TestDescriptionPlugin:
             yield
             if self.desc:
                     self.terminal_reporter.write(f'\n{self.desc} ')
-
-
-def pytest_report_header(config):
-    """Add CRDS_CONTEXT to pytest report header"""
-    return f"crds_context: {get_context_used('jwst')}"
--- a/jwst/conftest.py
+++ b/jwst/conftest.py
@@ -4,10 +4,12 @@ import tempfile
 import pytest
 import inspect
 
+from crds.client.api import get_mapping_names, get_default_context
+from stdatamodels import s3_utils
+
 from jwst.associations import (AssociationRegistry, AssociationPool)
 from jwst.associations.tests.helpers import t_path
 from jwst.lib.tests import helpers as lib_helpers
-from stdatamodels import s3_utils
 
 
 @pytest.fixture(scope='session')
@@ -75,10 +77,12 @@ def jail(request, tmpdir_factory):
     yield newpath
     os.chdir(old_dir)
 
+
 @pytest.mark.trylast
 def pytest_configure(config):
     terminal_reporter = config.pluginmanager.getplugin('terminalreporter')
     config.pluginmanager.register(TestDescriptionPlugin(terminal_reporter), 'testdescription')
+
 
 class TestDescriptionPlugin:
     """Pytest plugin to print the test docstring when `pytest -vv` is used.
@@ -108,3 +112,17 @@ class TestDescriptionPlugin:
             yield
             if self.desc:
                     self.terminal_reporter.write(f'\n{self.desc} ')
+
+
+def get_crds_context():
+    """Return the in-use CRDS_CONTEXT for pytest report header"""
+    if "CRDS_CONTEXT" in os.environ:
+        pmap = [a for a in get_mapping_names(os.getenv("CRDS_CONTEXT")) if "pmap" in a][0]
+    else:
+        pmap = get_default_context()
+    return pmap
+
+
+def pytest_report_header(config):
+    """Add CRDS_CONTEXT to pytest report header"""
+    return f"crds_context: {get_crds_context()}"
--- a/jwst/conftest.py
+++ b/jwst/conftest.py
@@ -4,8 +4,8 @@ import tempfile
 import pytest
 import inspect
 
-from crds.client.api import get_mapping_names, get_default_context
 from stdatamodels import s3_utils
+from stpipe.crds_client import get_context_used
 
 from jwst.associations import (AssociationRegistry, AssociationPool)
 from jwst.associations.tests.helpers import t_path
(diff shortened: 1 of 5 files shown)


scripts/echo_crds_context
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+if [[ "${CRDS_CONTEXT}" ]]; then
+    echo $(crds list --contexts ${CRDS_CONTEXT} --resolve-contexts)
```
pllim, line 4: This is a step backwards if you plan on supporting Windows in the future.

scripts/echo_crds_context
```diff
@@ -0,0 +1,7 @@
...
+else
+    crds list --operational-context
+fi
```
jdavies-st, line 7: This script gives a consistent way to print out the current CRDS_CONTEXT pmap, regardless of how it is set.  Observe usage in where context is not set (and gets the operational context), set to the floating "jwst-edit" or set to a specific context.

```
$ unset CRDS_CONTEXT
$ echo_crds_context 
jwst_0672.pmap
$ export CRDS_CONTEXT=jwst-edit
$ echo_crds_context 
jwst_0674.pmap
$ export CRDS_CONTEXT=jwst_0673.pmap
$ echo_crds_context 
jwst_0673.pmap
$ unset CRDS_CONTEXT
$ echo_crds_context 
jwst_0672.pmap
```

pllim:
Makes me curious though... how were things passing when it was wrong?

scripts/echo_crds_context
jdavies-st: Doh!  You're right.  Any suggestions on something that could be used instead?  Do I really have to write a github action to do this?  😂 

Of course CRDS and it's CLI is not Windows compatible at all right now, so that's probably the larger hurdle.  Maybe the new CLI for CRDS would actually have the above functionality as a single command?  Can only hope.

jdavies-st:
> Makes me curious though... how were things passing when it was wrong?

I think the bad YAML was just being ignored.  And none of our unit tests actually depend on CRDS_CONTEXT being set to `jwst-edit`.  We only do it here to catch early anything buggy reffiles that might have got put into CRDS, so we catch them before that becomes the operational context.

scripts/echo_crds_context
pllim: This is a little frustrating because I thought CRDS is implemented in Python in the first place! Is there no way to call this from the Pythonic API?

Regardless, perhaps you want to look at https://docs.pytest.org/en/stable/example/simple.html#adding-info-to-test-report-header instead of creating a custom script like this. If Python API isn't an option, at least code it up such that it is a no-op if Windows is detected?

jdavies-st: Yes, implemented in python, but the CLI is a non-trivial map to the underlying library code.  I think I've figured it out.

And thanks for the excellent report header integration suggestion.  This makes it much slicker, and is now available to users even without using `tox`.  🎉

pllim (commented):
Does seem cleaner, but I don't see the custom header in your test log. This is a known issue that header doesn't display in `tox` unless you move `conftest.py` to root level. Is there a reason why it has to be embedded inside `jwst` dir?

jwst/conftest.py
```diff
@@ -108,3 +112,17 @@ def pytest_runtest_logstart(self, nodeid, location):
             yield
             if self.desc:
                     self.terminal_reporter.write(f'\n{self.desc} ')
+
+
+def get_crds_context():
```
eslavich, line 117: There's a [method](https://github.com/spacetelescope/stpipe/blob/0.1.0/src/stpipe/crds_client.py#L179) in `stpipe.crds_client` that returns this value.  It uses `crds.heavy_client.get_processing_mode` which is a good choice because it returns the actual context that the crds client is using.

eslavich (commented):
Looks good, just a comment about an stpipe method that may be of use

jwst/conftest.py
jdavies-st: Thanks!  Perfect.  Changed the code.

jdavies-st:
> Does seem cleaner, but I don't see the custom header in your test log. This is a known issue that header doesn't display in `tox` unless you move `conftest.py` to root level.

That's kinda nutty.  When I run this locally under tox, it does appear.  😕 

And now I see the difference.  If I let tox run pytest at the top level of the repo with no subdir args, it doesn't print the header.  I.e.:

```
tox -e py39-sdpdeps
```

If a specify the `jwst` subdir in the repo, then it does print.

```
tox -e py39-sdpdeps jwst
```

and that seems to be the case with `--pyargs` too.  🙄

jdavies-st:
Well, putting it in a root-level `conftest.py` fixes it.  Crazytown.

I think this PR is ready.

eslavich (approved):
LGTM

pllim:
Is `0676` the `jwst-edit` you are looking for? If so, this is perfect, as the solution is very clean. Congrats!

```
============================= test session starts ==============================
platform linux -- Python 3.8.7, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
cachedir: .tox/sdpdeps/.pytest_cache
crds_context: jwst_0676.pmap
rootdir: /home/runner/work/jwst/jwst, configfile: setup.cfg
plugins: cov-2.11.1, openfiles-0.5.0, doctestplus-0.9.0, ci-watson-0.5, requests-mock-1.8.0, asdf-2.7.2
collected 2417 items / 1 skipped / 2416 selected
```

jdavies-st:
Yes!  And thanks for the reviews @pllim and @eslavich.

3 heads together is better than one.  🙈 🙉 🙊

PR #5738, opened 2021-02-11
JP-1657: Make ivm default weight_type for resample

### LINKED ISSUE #800: Implement IVM weighting in resample step
This is a reminder that IVM (inverse variance map) weighting is needed to produce photometric errors in the source catalog step.  Currently the `wht_type` `'IVM'` option is disabled:

https://github.com/STScI-JWST/jwst/blob/master/jwst/resample/resample_spec.py#L506

jdavies-st:
This adds `weight_type="ivm"` as the default weight scheme for `ResampleStep`.

This will obviously cause changes in the _s2d.fits, _i2d.fits, _x1d.fits and _cat.ecsv outputs in our regression tests.  Further, some of the input _rate.fits files (NIRSpec MOS) do not have `var_rnoise`, so when it is accessed, it is all zeros, which gives zero weight to everything in the output.  So those will have to have dummy `var_rnoise` arrays added to the input files.

Resolves #800 / [JP-1657](https://jira.stsci.edu/browse/JP-1657)

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -82,7 +82,6 @@ bc0.build_cmds = [
     "pip install -e .[test,ephem]",
     "pip install pytest-xdist pytest-sugar",
     "pip freeze",
-    'echo "CRDS_CONTEXT = $(crds list --contexts $CRDS_CONTEXT --mappings | grep pmap)"',
 ]
 bc0.build_cmds = PipInject(env.OVERRIDE_REQUIREMENTS) + bc0.build_cmds
 bc0.test_cmds = [
--- a/jwst/resample/gwcs_drizzle.py
+++ b/jwst/resample/gwcs_drizzle.py
@@ -12,9 +12,8 @@ class GWCSDrizzle:
     """
     Combine images using the drizzle algorithm
     """
-    def __init__(self, product, outwcs=None, single=False,
-                 wt_scl="exptime", pixfrac=1.0, kernel="square",
-                 fillval="INDEF"):
+    def __init__(self, product, outwcs=None, single=False, wt_scl=None,
+                 pixfrac=1.0, kernel="square", fillval="INDEF"):
         """
         Create a new Drizzle output object and set the drizzle parameters.
 
@@ -33,9 +32,9 @@ class GWCSDrizzle:
 
         wt_scl : str, optional
             How each input image should be scaled. The choices are `exptime`
-            which scales each image by its exposure time, `expsq` which scales
-            each image by the exposure time squared, or an empty string, which
-            allows each input image to be scaled individually.
+            which scales each image by its exposure time or `expsq` which scales
+            each image by the exposure time squared.  If not set, then each
+            input image is scaled by its own weight map.
 
         pixfrac : float, optional
             The fraction of a pixel that the pixel flux is confined to. The
@@ -63,7 +62,10 @@ class GWCSDrizzle:
         self.outexptime = 0.0
         self.uniqid = 0
 
-        self.wt_scl = wt_scl
+        if wt_scl is None:
+            self.wt_scl = ""
+        else:
+            self.wt_scl = wt_scl
         self.kernel = kernel
         self.fillval = fillval
         self.pixfrac = pixfrac
@@ -89,22 +91,13 @@ class GWCSDrizzle:
             self.outcon = np.reshape(self.outcon, (1,
                                      self.outcon.shape[0],
                                      self.outcon.shape[1]))
-
-        elif self.outcon.ndim == 3:
-            pass
-
-        else:
+        elif self.outcon.ndim != 3:
             raise ValueError("Drizzle context image has wrong dimensions: \
                 {0}".format(product))
 
         # Check field values
         if not self.outwcs:
             raise ValueError("Either an existing file or wcs must be supplied")
-
-        if util.is_blank(self.wt_scl):
-            self.wt_scl = ''
-        elif self.wt_scl != "exptime" and self.wt_scl != "expsq":
-            raise ValueError("Illegal value for wt_scl: %s" % self.wt_scl)
 
         if out_units == "counts":
             np.divide(self.outsci, self.outexptime, self.outsci)
--- a/jwst/resample/gwcs_drizzle.py
+++ b/jwst/resample/gwcs_drizzle.py
@@ -1,7 +1,6 @@
 import numpy as np
 
 from drizzle import util
-from drizzle import doblot
 from drizzle import cdrizzle
 from . import resample_utils
 
@@ -198,34 +197,6 @@ class GWCSDrizzle:
                   self.outcon, expin, in_units, wt_scl, uniqid=self.uniqid,
                   xmin=xmin, xmax=xmax, ymin=ymin, ymax=ymax,
                   pixfrac=self.pixfrac, kernel=self.kernel, fillval=self.fillval)
-
-    def blot_image(self, blotwcs, interp='poly5', sinscl=1.0):
-        """
-        Resample the output image using an input world coordinate system.
-
-        Parameters
-        ----------
-
-        blotwcs : wcs
-            The world coordinate system to resample on.
-
-        interp : str, optional
-            The type of interpolation used in the resampling. The
-            possible values are "nearest" (nearest neighbor interpolation),
-            "linear" (bilinear interpolation), "poly3" (cubic polynomial
-            interpolation), "poly5" (quintic polynomial interpolation),
-            "sinc" (sinc interpolation), "lan3" (3rd order Lanczos
-            interpolation), and "lan5" (5th order Lanczos interpolation).
-
-        sincscl : float, optional
-            The scaling factor for sinc interpolation.
-        """
-
-        util.set_pscale(blotwcs)
-        self.outsci = doblot.doblot(self.outsci, self.outwcs, blotwcs,
-                                    1.0, interp=interp, sinscl=sinscl)
-
-        self.outwcs = blotwcs
 
     def increment_id(self):
         """
--- a/jwst/resample/gwcs_drizzle.py
+++ b/jwst/resample/gwcs_drizzle.py
@@ -31,8 +31,8 @@ class GWCSDrizzle:
             provided, the WCS is taken from product.
 
         wt_scl : str, optional
-            How each input image should be scaled. The choices are `exptime`
-            which scales each image by its exposure time or `expsq` which scales
+            How each input image should be scaled. The choices are `exptime`,
+            which scales each image by its exposure time, or `expsq`, which scales
             each image by the exposure time squared.  If not set, then each
             input image is scaled by its own weight map.
 
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -41,7 +41,7 @@ class ResampleData:
          (eventually) a record of metadata from all input models.
     """
     def __init__(self, input_models, output=None, single=False, blendheaders=True,
-        pixfrac=1.0, kernel="square", fillval="INDEF", weight_type="exptime",
+        pixfrac=1.0, kernel="square", fillval="INDEF", weight_type="ivm",
         good_bits=0, pscale_ratio=1.0, **kwargs):
         """
         Parameters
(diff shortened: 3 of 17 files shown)


jwst/resample/gwcs_drizzle.py
```diff
@@ -34,9 +32,9 @@ def __init__(self, product, outwcs=None, single=False,
 
         wt_scl : str, optional
             How each input image should be scaled. The choices are `exptime`
-            which scales each image by its exposure time, `expsq` which scales
-            each image by the exposure time squared, or an empty string, which
-            allows each input image to be scaled individually.
+            which scales each image by its exposure time or `expsq` which scales
```
hbushouse, line 35: OK, if you're fixing punctuation you need to add commas at the ends of the words immediately preceding each occurrence of "which", i.e. "The choices are `exptime`, which scales each image by its exposure time, or `expsq`, which scales each image ...".

jwst/resample/gwcs_drizzle.py
```diff
@@ -199,34 +191,6 @@ def add_image(self, insci, inwcs, inwht=None, xmin=0, xmax=0, ymin=0, ymax=0,
                   xmin=xmin, xmax=xmax, ymin=ymin, ymax=ymax,
                   pixfrac=self.pixfrac, kernel=self.kernel, fillval=self.fillval)
 
-    def blot_image(self, blotwcs, interp='poly5', sinscl=1.0):
```
hbushouse, line 202: I'm assuming we don't need this function any more?

jwst/resample/resample.py
```diff
@@ -164,8 +163,7 @@ def do_drizzle(self):
                 inwht = resample_utils.build_driz_weight(img,
                     weight_type=self.weight_type,
                     good_bits=self.good_bits)
-                driz.add_image(img.data, img.meta.wcs, inwht=inwht,
-                    expin=img.meta.exposure.exposure_time)
```
hbushouse, line 168: So given that exptime is still an allowed choice for weighting (just no longer the default), what happens when someone selects exptime but the exposure_time is no longer being passed to the lower-level functions?

jwst/resample/resample_step.py
```diff
@@ -68,7 +68,7 @@ def process(self, input):
             self.log.info('Drizpars reference file: {}'.format(ref_filename))
             kwargs = self.get_drizpars(ref_filename, input_models)
         else:
-            # Deal with NIRSpec which currently has no default drizpars reffile
+            # Deal with FGS which currently has no default drizpars reffile
```
hbushouse, line 71: Why does the comment here say "FGS", but the log message in the next line says "NIRSpec"? It would be better to stay instrument-agnostic and just say there's no ref file, period, with no mention of instruments.

jwst/resample/resample_step.py
```diff
@@ -148,15 +147,13 @@ def get_drizpars(self, ref_filename, input_models):
...
-        # Define the keys to pull from drizpars reffile table.  Note the
-        # step param 'weight_type' is 'wht_type' in the FITS binary table.
+        # Define the keys to pull from drizpars reffile table.
         # All values should be None unless the user set them on the command
         # line or in the call to the step
         drizpars = dict(
             pixfrac=self.pixfrac,
             kernel=self.kernel,
             fillval=self.fillval,
-            wht_type=self.weight_type,
```
hbushouse, line 159: Is wht_type still present in the ref file, but we're just ignoring it? So what happens if a user intentionally tries to change the behavior by modifying the values in the ref file?

jwst/resample/resample_utils.py
```diff
@@ -140,33 +140,35 @@ def _reproject(x, y):
...
     exptime = model.meta.exposure.exposure_time
 
-    if weight_type == 'error':
-        err_model = np.nan_to_num(model.err)
-        inwht = (exptime / err_model)**2 * dqmask
-        log.debug("DEBUG weight mask: {} {}".format(type(inwht), np.sum(inwht)))
-    # elif weight_type == 'ivm':
-    #     _inwht = img.buildIVMmask(chip._chip,dqarr,pix_ratio)
+    if weight_type == 'ivm':
+        with np.errstate(divide="ignore", invalid="ignore"):
```
hbushouse, line 149: Ooh, nice. I have to remember to use this in the future.

jwst/resample/tests/test_resample_step.py
```diff
@@ -3,14 +3,16 @@
...
-from jwst.datamodels import ImageModel
+from jwst.datamodels import ImageModel, ModelContainer
 from jwst.assign_wcs import AssignWcsStep
 from jwst.extract_2d import Extract2dStep
 from jwst.resample import ResampleSpecStep, ResampleStep
 
 @pytest.fixture
 def nirspec_rate():
-    shape = (2048, 2048)
+    ysize = 2048
```
hbushouse, line 13: Hey, I thought we were supposed to keep test images *small*?!

hbushouse (commented):
Did you ever figure out why the current WHT arrays looked like they had Poisson noise included, even though it was supposed to be weighting by only exptime?

JenkinsfileRT
```diff
@@ -82,7 +82,6 @@ bc0.build_cmds = [
     "pip install -e .[test,ephem]",
     "pip install pytest-xdist pytest-sugar",
     "pip freeze",
-    'echo "CRDS_CONTEXT = $(crds list --contexts $CRDS_CONTEXT --mappings | grep pmap)"',
```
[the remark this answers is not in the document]
jdavies-st: The CRDS_CONTEXT is now printed out by default every time you run any test in the `jwst` repo.  See #5737.  I had forgot to remove it in our Jenkinsfiles in that PR.  Oops!

jwst/resample/gwcs_drizzle.py
jdavies-st: Yeah, it's not used at all, and it's incorrect for `jwst` which uses `gwcs.WCS` objects.  I think it was just copied over from `drizzle` accidentally.

The only time we use `blot` is in `outlier_detection`, and that calls the C blot routine directly, not through this function.  Clearly a refactor in the future would be desirable, but I could say that about many aspects of `resample`.  😅

jwst/resample/resample.py
jdavies-st: Ah, good catch.  Let me look at my unit tests and see what I missed.

That raises the larger question, should we even allow exposure time to be used?  It clearly gives the _wrong_ answer for multi-read up-the-ramp data, so maybe it should just be completely removed?

jwst/resample/resample_step.py
jdavies-st: Doh!  Didn't catch the log message.  Agree it should be instrument-agnostic.  And once we move to `pars-resamplestep` CRDS reffiles, we should just remove all of this (and the `drizpars` CRDS reffile) to boot.  A future PR.

jwst/resample/resample_step.py
jdavies-st: We are ignoring it.  All the default reffiles in CRDS have this set as `exptime`, and it would necessarily override whatever I set as the default in `ResampeStep.spec`.  So we are ignoring it.  And we are going to be deprecating replacing `drizpars` reffiles with `pars-resamplestep` reffiles anyway, so I really don't want to create a new drizpars reffile for every instrument or encourage users to do so either.  It's a dead end.

jwst/resample/resample_utils.py
jdavies-st: Yes, useful!  Though it has it's pitfalls. I only did it after I made sure the arrays looked how I expected them to, as the warning messages were telling me something.  For instance, dividing by zero variance produced `np.inf`, but the original code here that used `np.nan_to_num` then turned that `np.inf` into:

```
In [2]: np.nan_to_num(np.inf)                                                                                           
Out[2]: 1.7976931348623157e+308
```

which is clearly wrong.  Hence the call below to `np.isfinite`.

jwst/resample/tests/test_resample_step.py
jdavies-st: I think this needs to stay large because it is NIRSpec for some reason?  I didn't change it from the original test that was put here (not by me).  It still runs fast, because it is not doing any jump detection or ramp fitting.

jdavies-st:
As mentioned above, some of our input regression test data is very old and predates the population of `var_rnoise` arrays in ramp fitting.  This is the case for the simulated NIRSpec MOS data we have and a couple other datasets.  For real data, there will always be `var_rnoise`.  For this old data, is mocking it up with ones acceptable?  Any other ideas?

Without mocking it up, the `var_rnoise` attributes gets auto-created with zeros, and then everything has zero weight and you get no data in the output.

jdavies-st:
> Did you ever figure out why the current WHT arrays looked like they had Poisson noise included, even though it was supposed to be weighting by only exptime?

Yes, and it turns out that got fixed in the 0.17.x releases.  🙄

jwst/resample/resample.py
hbushouse: So what's the verdict here? Does it still work when selecting "exptime" weighting?

hbushouse:
> As mentioned above, some of our input regression test data is very old and predates the population of `var_rnoise` arrays in ramp fitting. This is the case for the simulated NIRSpec MOS data we have and a couple other datasets. For real data, there will always be `var_rnoise`. For this old data, is mocking it up with ones acceptable? Any other ideas?
> 
> Without mocking it up, the `var_rnoise` attributes gets auto-created with zeros, and then everything has zero weight and you get no data in the output.

Yes, I guess we'll need to fix the input files to the regtests by just adding some generic readnoise value. The simulated NRS data we get from the team is already in the form of a countrate image (equivalent to rate file) and hence it doesn't go through caldetector1 or ramp_fit to get `var_rnoise` populated. So just look at one of the NRS readnoise ref files and estimate a suitable value from that.

hbushouse (approved):

jwst/resample/resample_step.py
```diff
@@ -34,7 +34,7 @@ class ResampleStep(Step):
         pixfrac = float(default=1.0)
         kernel = string(default='square')
         fillval = string(default='INDEF')
-        weight_type = option('exptime', default='exptime')
+        weight_type = option('ivm', 'exptime', default='ivm')
```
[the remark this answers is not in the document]
jdavies-st: Ah, thanks for catching that @eslavich .  Will fix.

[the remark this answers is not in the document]
jdavies-st: Fixed.

jdavies-st:
> Yes, I guess we'll need to fix the input files to the regtests by just adding some generic readnoise value.

As you suggested, perhaps it would be best to check if the array exists.  If it doesn't then use weight of 1 for everything and emit a warning.  This would be a more general purpose solution that chasing down all our data that doesn't have a `var_rnoise` array in it.

jdavies-st:
So, for datasets that start `calwebb_spec2` with no `var_rnoise` array, it gets created pretty early on as soon as the first error propagation step happens.  Certainly by `flat_field`.  So by the time we get to `resample` it is already zeros and exists.  This presents a problem for dealing with it in a smart way in `resample`.

Also, at this point, it also has NaNs in it due scaling it by the flat, so it's a mess - a combination of zeros and NaNs.

Will think on this a bit more and find a solution.

jdavies-st:
So the solution is to raise a `ValueError` if the weight map is all zeros, or a combination of zeros and non-finite numbers (`np.inf`, `np.nan`) with a useful error message.  For data coming through Ops, all should have `var_rnoise` arrays.

jdavies-st:
I have updated the rate.fits files in our nightly regression tests that are old enough to not have `var_rnoise` arrays.  And added one to the NIRSpec MOS observation that is simulated as a rate file and was not run through ramp fitting.

This PR is done.  Next step is to figure out what goes into the `err` array for these resampled data.

hbushouse:
Given that JP-1567 was really just concerned with turning on IVM weighting, which this PR accomplishes, I wonder if it would be cleanest to also close JP-1567 and start a new ticket to continue the discussion of how to create an ERR array in resampled products. They really are 2 different things (as you've explained to me).

jdavies-st:
Well, it turns out raising errors in `build_driz_weight` raises havoc throughout the rest of our unit tests.  I've opted for warnings for now.  But happy to revisit or modify in a subsequent PR.  I'm going to merge this and kick off regtests.

PR #5739, opened 2021-02-11
Changing the interface to function `ols_ramp_fit`.

kmacdonald-stsci:
Refactoring `ramp_fit.py`, changing the interface of `ols_ramp_fit`.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -36,6 +36,81 @@ JUMP_DET = dqflags.group['JUMP_DET']
 BUFSIZE = 1024 * 300000  # 300Mb cache size for data section
 
 
+class RampInput:
+    """
+    Captures the input model data needed for ramp fitting.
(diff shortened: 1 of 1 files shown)


jdavies-st (commented):
Another way to implement this would be to use Python's built-in `namedtuple` class factory.  See https://docs.python.org/3/library/collections.html#collections.namedtuple

```python
RampInput = namedtuple("RampInput",["data", err", "groupdq", "pixeldq", "instrume", "frame_time", "ngroups", "group_time", "groupgap", "nframes", "dropframes1"])
```

which essentially allows you to define your data container class in a single line.  And then, once instantiated, you can access the elements like

```python
ramp_input = RampInput(foo, bar, ...)
ramp_input.err += 1
print(ramp_input.group_time)
```

dmggh:
Ken, I think your refactored code (and James' suggestion) is an improvement in terms of organization and readablilty.  Just my 2 cents.

PR #5743, opened 2021-02-12
JP-1563: sigma-clip master background

### LINKED ISSUE #5118: Master background uses incorrect surface brightness estimation
_Issue [JP-1563](https://jira.stsci.edu/browse/JP-1563) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

In testing the master background step, I noticed that the lvl3 association files are set up such that it requires the input background to be provided via the x1d spectrum constructed from individual files in spec2.  This will result in significant quality problems because of how extract_1d works.

Specifically, for extended source observations (e.g., dedicated backgrounds) extract1d in spec2 sums the spectrum of the entire data cube and divides by the area to get the effective surface brightness (although see <https://jira.stsci.edu/browse/JP-1562> regarding a bug in this computation).  Since this is a sum, and not a sigma-clipped mean or median, it is not robust against outlying artifacts such as unfiltered cosmic rays or unexpected sources.  As shown in the attached image, such artifacts can cause spikes in the 1d spectrum that would then be replicated to the master background used throughout the data cube.

Similarly, the example attached also shows a falloff at the ends of a band where fewer and fewer IFU slices contribute to a given wavelength.  Although spec3 builds data cubes that do not use these wavelength ranges, spec2 does use them since it combines information from both channels into a single spectrum.

The best solution from the perspective of master background subtraction is to use sigma clipping or a median estimator to maximize the quality of the 1d spectrum prior to using it as the master background.  However, this probably wouldn't be desirable for extract1d to do in general, as extract1d is more frequently used to get the spectrum of extended sources where you **don't** want to clip away signal from your object.

The cleanest solution may be for master_background to do its own computation of the 1d background from input 3d background cubes, or even from the _cal.fits background data (i.e., building its own cubes as necessary).

Tagging [Anton Koekemoer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=koekemoe) since the solution to this should involved the Cal WG before moving to the DMS WG.

jemorrison:
The PR address JP-1563. For IFU data for an extended source, the extracted region is sigma clipped to reject outliers. The sigma-clipped result is converted to units of surface brightness and stored in the 'background' column of the extract_1d output.  

Fixes #5118 / [JP-1563](https://jira.stsci.edu/browse/JP-1563)

### DIFF
--- a/jwst/combine_1d/combine1d.py
+++ b/jwst/combine_1d/combine1d.py
@@ -18,7 +18,7 @@ class InputSpectrumModel:
         wavelength
         flux
         error
-        surf_bright
+        background
         sb_error
         dq
         nelem
@@ -53,12 +53,12 @@ class InputSpectrumModel:
         self.flux = spec.spec_table.field("flux").copy()
         self.error = spec.spec_table.field("error").copy()
         try:
-            self.surf_bright = spec.spec_table.field("surf_bright").copy()
+            self.background = spec.spec_table.field("background").copy()
             self.sb_error = spec.spec_table.field("sb_error").copy()
         except KeyError:
-            self.surf_bright = np.zeros_like(self.flux)
+            self.background = np.zeros_like(self.flux)
             self.sb_error = np.zeros_like(self.flux)
-            log.warning("There is no SURF_BRIGHT column in the input.")
+            log.warning("There is no BACKGROUND column in the input.")
         self.dq = spec.spec_table.field("dq").copy()
         self.nelem = self.wavelength.shape[0]
         self.unit_weight = False        # may be reset below
@@ -88,7 +88,7 @@ class InputSpectrumModel:
         self.wavelength = None
         self.flux = None
         self.error = None
-        self.surf_bright = None
+        self.background = None
         self.sb_error = None
         self.dq = None
         self.nelem = 0
@@ -104,7 +104,7 @@ class OutputSpectrumModel:
         wavelength
         flux
         error
-        surf_bright
+        background
         sb_error
         dq
         weight
@@ -118,7 +118,7 @@ class OutputSpectrumModel:
         self.wavelength = None
         self.flux = None
         self.error = None
-        self.surf_bright = None
+        self.background = None
         self.sb_error = None
         self.dq = None
         self.weight = None
@@ -182,7 +182,7 @@ class OutputSpectrumModel:
 
         self.flux = np.zeros(nelem, dtype=np.float64)
         self.error = np.zeros(nelem, dtype=np.float64)
-        self.surf_bright = np.zeros(nelem, dtype=np.float64)
+        self.background = np.zeros(nelem, dtype=np.float64)
         self.sb_error = np.zeros(nelem, dtype=np.float64)
         self.dq = np.zeros(nelem, dtype=dq_dtype)
         self.weight = np.zeros(nelem, dtype=np.float64)
@@ -215,7 +215,7 @@ class OutputSpectrumModel:
                 self.dq[k] |= in_spec.dq[i]
                 self.flux[k] += in_spec.flux[i] * weight
                 self.error[k] += (in_spec.error[i] * weight)**2
-                self.surf_bright[k] += (in_spec.surf_bright[i] * weight)
+                self.background[k] += (in_spec.background[i] * weight)
                 self.sb_error[k] += (in_spec.sb_error[i] * weight)**2
                 self.weight[k] += weight
                 self.count[k] += 1.
@@ -235,7 +235,7 @@ class OutputSpectrumModel:
             self.wavelength = self.wavelength[index]
             self.flux = self.flux[index]
             self.error = self.error[index]
-            self.surf_bright = self.surf_bright[index]
+            self.background = self.background[index]
             self.sb_error = self.sb_error[index]
             self.dq = self.dq[index]
             self.weight = self.weight[index]
@@ -249,7 +249,7 @@ class OutputSpectrumModel:
 
         if not self.normalized:
             sum_weight = np.where(self.weight > 0., self.weight, 1.)
-            self.surf_bright /= sum_weight
+            self.background /= sum_weight
             self.flux /= sum_weight
             self.error = np.sqrt(self.error / sum_weight)
             self.sb_error = np.sqrt(self.sb_error / sum_weight)
@@ -274,7 +274,7 @@ class OutputSpectrumModel:
         data = np.array(list(zip(self.wavelength,
                                  self.flux,
                                  self.error,
-                                 self.surf_bright,
+                                 self.background,
                                  self.sb_error,
                                  self.dq,
                                  self.weight,
@@ -287,7 +287,7 @@ class OutputSpectrumModel:
         self.wavelength = None
         self.flux = None
         self.error = None
-        self.surf_bright = None
+        self.background = None
         self.sb_error = None
         self.dq = None
         self.weight = None
--- a/jwst/combine_1d/combine1d.py
+++ b/jwst/combine_1d/combine1d.py
@@ -18,7 +18,7 @@ class InputSpectrumModel:
         wavelength
         flux
         error
-        background
+        surf_bright
         sb_error
         dq
         nelem
@@ -53,12 +53,12 @@ class InputSpectrumModel:
         self.flux = spec.spec_table.field("flux").copy()
         self.error = spec.spec_table.field("error").copy()
         try:
-            self.background = spec.spec_table.field("background").copy()
+            self.surf_bright = spec.spec_table.field("surf_bright").copy()
             self.sb_error = spec.spec_table.field("sb_error").copy()
         except KeyError:
-            self.background = np.zeros_like(self.flux)
+            self.surf_bright = np.zeros_like(self.flux)
             self.sb_error = np.zeros_like(self.flux)
-            log.warning("There is no BACKGROUND column in the input.")
+            log.warning("There is no SURF_BRIGHT column in the input.")
         self.dq = spec.spec_table.field("dq").copy()
         self.nelem = self.wavelength.shape[0]
         self.unit_weight = False        # may be reset below
@@ -88,7 +88,7 @@ class InputSpectrumModel:
         self.wavelength = None
         self.flux = None
         self.error = None
-        self.background = None
+        self.surf_bright = None
         self.sb_error = None
         self.dq = None
         self.nelem = 0
@@ -104,7 +104,7 @@ class OutputSpectrumModel:
         wavelength
         flux
         error
-        background
+        surf_bright
         sb_error
         dq
         weight
@@ -118,7 +118,7 @@ class OutputSpectrumModel:
         self.wavelength = None
         self.flux = None
         self.error = None
-        self.background = None
+        self.surf_bright = None
         self.sb_error = None
         self.dq = None
         self.weight = None
@@ -182,7 +182,7 @@ class OutputSpectrumModel:
 
         self.flux = np.zeros(nelem, dtype=np.float64)
         self.error = np.zeros(nelem, dtype=np.float64)
-        self.background = np.zeros(nelem, dtype=np.float64)
+        self.surf_bright = np.zeros(nelem, dtype=np.float64)
         self.sb_error = np.zeros(nelem, dtype=np.float64)
         self.dq = np.zeros(nelem, dtype=dq_dtype)
         self.weight = np.zeros(nelem, dtype=np.float64)
@@ -215,7 +215,7 @@ class OutputSpectrumModel:
                 self.dq[k] |= in_spec.dq[i]
                 self.flux[k] += in_spec.flux[i] * weight
                 self.error[k] += (in_spec.error[i] * weight)**2
-                self.background[k] += (in_spec.background[i] * weight)
+                self.surf_bright[k] += (in_spec.surf_bright[i] * weight)
                 self.sb_error[k] += (in_spec.sb_error[i] * weight)**2
                 self.weight[k] += weight
                 self.count[k] += 1.
@@ -235,7 +235,7 @@ class OutputSpectrumModel:
             self.wavelength = self.wavelength[index]
             self.flux = self.flux[index]
             self.error = self.error[index]
-            self.background = self.background[index]
+            self.surf_bright = self.surf_bright[index]
             self.sb_error = self.sb_error[index]
             self.dq = self.dq[index]
             self.weight = self.weight[index]
@@ -249,7 +249,7 @@ class OutputSpectrumModel:
 
         if not self.normalized:
             sum_weight = np.where(self.weight > 0., self.weight, 1.)
-            self.background /= sum_weight
+            self.surf_bright /= sum_weight
             self.flux /= sum_weight
             self.error = np.sqrt(self.error / sum_weight)
             self.sb_error = np.sqrt(self.sb_error / sum_weight)
@@ -274,7 +274,7 @@ class OutputSpectrumModel:
         data = np.array(list(zip(self.wavelength,
                                  self.flux,
                                  self.error,
-                                 self.background,
+                                 self.surf_bright,
                                  self.sb_error,
                                  self.dq,
                                  self.weight,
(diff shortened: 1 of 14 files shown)


jwst/extract_1d/ifu.py
```diff
@@ -1080,3 +1124,54 @@ def shift_ref_image(mask, delta_y, delta_x, fill=0):
     temp[..., oslice_y, oslice_x] = mask[..., islice_y, islice_x]
 
     return temp
+
+def sigma_clip_extended_region(data, mask_targ, wmap, sigma_clip):
+    """ Convert 2D back mask to 3D and flag pixels identified as outliers with 0
```
jdavies-st, line 1129:
```suggestion
    """Convert 2D background mask to 3D and flag pixels identified as outliers with 0
```

jdavies-st: Can you write a unit test for this function?  It is not tested at all.

jwst/extract_1d/ifu.py
```diff
@@ -1080,3 +1124,54 @@ def shift_ref_image(mask, delta_y, delta_x, fill=0):
...
+        For each wavelength plane perform sigma clipping of extracted region
+
+    Parameters
+    ----------
+    data: 3-D
+        Input data array to perform extraction from
```
jdavies-st, line 1135: 3-D is not an input type.  Is this an `ndarray`?

jwst/extract_1d/ifu.py
```diff
@@ -1080,3 +1124,54 @@ def shift_ref_image(mask, delta_y, delta_x, fill=0):
...
+        pixel is in the extraction region.
+    wmap: ndarray, 3-D
+       weight map for IFU
+    sigma_clip : float
+        outlier sigma clipping parameter
+
+    Returns
+    -------
+    temp : ndarray, same type and shape as `mask_bkg`
+        A copy of `mask_bkg`, but outliers flagged as 0
```
jdavies-st, line 1148: This docstring doesn't coincide with what the function actually returns.

jwst/extract_1d/extract1d.py
```diff
@@ -24,7 +24,7 @@
 
 def extract1d(image, lambdas, disp_range,
               p_src, p_bkg=None, independent_var="wavelength",
-              smoothing_length=0, bkg_fit="poly", bkg_order=0, weights=None):
+              smoothing_length=0, bkg_fit="poly", bkg_order=0, bkg_sigma_clip=0, weights=None):
```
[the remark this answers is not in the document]
jemorrison: I realize that is a mistake I should not pass it to extract1d.py . I was going to but then it was decided this is only for IFU type data. I will remove it.

jwst/extract_1d/ifu.py
jemorrison: whoops - I fixed that

jwst/datamodels/schemas/combinedspectable.schema.yaml
```diff
@@ -11,7 +11,7 @@ datatype:
   datatype: float64
 - name: ERROR
   datatype: float64
-- name: SURF_BRIGHT
+- name: BACKGROUND
```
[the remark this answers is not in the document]
jemorrison: That was the intent. Currently it is a bit confusing using surf_bright since depending on the type of data sometimes is it from the surf_bright column of extract 1d and sometimes it is from the background column. By changing it I was trying to make it completely clear what it was.  I hope this does not create too many new issues.

jemorrison:
@hbushouse Using this branch I get a failure in test_nirspec_masterbackground.py the test: test_nirspec_specc_mbkg - which is a masterbackgrund on MOS data. The result seem wrong with this branch.  The resulting nrs_mos_with_bkglsits_masterbg2d.fits has slits with zero data.  I think I need to update more code in the master background step for MOS data. I am working on figure how what I need to now

hbushouse:
The NIRSpec mbkg test is a case of MOS mode with dedicated background slits in the exposure, hence it's running the calspec2 pipeline (not calspec3), because that's where NRS MOS mbkg is now performed. It is still calling the master_background step, but calling it from calspec2. I helped write the calspec2 stuff for MOS mode, so I'll take a look at the code to remind myself what it's doing with the data from the background slits and where it gets stored.

jwst/combine_1d/combine1d.py
```diff
@@ -52,12 +52,12 @@ def __init__(self, ms, spec, exptime_key):
         self.flux = spec.spec_table.field("flux").copy()
         self.error = spec.spec_table.field("error").copy()
         try:
-            self.surf_bright = spec.spec_table.field("surf_bright").copy()
+            self.background = spec.spec_table.field("background").copy()
```
hbushouse, line 55: You can't do this. The combine_1d step's main purpose is for combining 1D spectra of science targets and hence the primary data of interest will be contained in the `flux` and `surf_bright` columns, which will usually already have had background subtraction applied (if it's possible to do background subtraction for the particular mode). This is why many of the regtests are now failing, including the NRS MOS mode, because it uses combine_1d to combine source-like spectra from slits that happen to contain only background signal, and hence all of the data of interest are in the `flux` and `surf_bright` columns, which are now being ignored by this step.

jwst/datamodels/schemas/combinedspectable.schema.yaml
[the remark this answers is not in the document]
hbushouse: We can NOT do this. The output of combine_1d must continue to contain `flux` and `surf_bright` columns. That's where all of the data of interest is stored when combine_1d is used to do things like combining the 1D spectra of WFSS sources.

jwst/combine_1d/combine1d.py
hbushouse: Unfortunately this indicates that we need to carefully rethink all the various scenarios of how master_bkg is applied and where the data "live" in each particular case. For cases that use nodded exposures of point sources, the bkg data naturally live in the "background" column of the incoming x1d product. For the case of dedicated bkg exposures, which are naturally treated as extended sources, the extracted data come from the "source" region of the image/cube, which would normally get stored in the "flux" and "surf_bright" columns of the x1d product, but we agreed to also put a copy of the data in the "background" column. But then in the case of a user-supplied spectrum that had its origins who knows where, we can't make any assumptions at all as to what x1d column contains the background signal we need. The assumption till now was that it was in the flux/surf_bright column. The same is true for the NRS MOS case with dedicated bkg slits. The data in those slits is extracted using a single extraction region and hence it naturally gets stored in flux/surf_bright, not background. Oy, what a mess.

jemorrison: Here was my thinking. The masterbackground step is now reading in the background column from x1d not the surf_brightness. So I thought I needed to change combine 1d to combine the background columns instead of the surface brightness columns. So I should go back and undo the combine 1d changes. But when combine 1d is called to combine background data how do we handle that ?

hbushouse: "But when combine_1d is called to combine background data how do we handle that?"   That's the big $$$ question. Without having spent more than 5-10 mins thinking about it so far, I'm beginning to wonder if we need a param in the master_bkg step that tells it which column to use, which would be user-settable to allow for customizations. Or at the very least it would have to be set by the master_bkg code based on some hierarchy of logic, covering all the use cases mentioned above. It would be a little like the old way of copying background into surf_bright (or vice versa), but instead of actually copying data, we just tell it which column to use.

hbushouse: And perhaps a switch/param in combine_1d as well, that tells it to combine the background data, instead of the flux/surf_bright data? Or it could just always combine all of them and then we leave it up to the downstream step (i.e. master_bkg) to decide which one it needs to use.

jemorrison: what if in masterbackground we go back and copy the needed information to surface brightness right in the beginning.
I think it would only have to change IFU extended source data - the background information is copied to surface brightness. 
Now the user background is another thing. If I should go back and change it to using the surf brightness column I can and then later after a more formal review we could decide on a clear path forward. Would that cover everything ?

jemorrison:
@hbushouse 
If we have the user provided background in the 'background' column (or a pointer to which column to use ) is there ever
a case where master background would read anything but the 'background column' from the various x1d files ?
Then could we add a column to the combine 1d data mode that contains background (in addition to surface brightness). 
Looking at the combine1d code it I don't think we would get any error if it by default  combined the surface brightness and background data (separately)The master background step would know to only use the background column.
Would there be a problem with changing the schema for the combine 1d schema? I obviously neglected how combine 1d was combining surface brightness for other types of data so I don't know how MOS data deals with a "background" - I think the backgrounds are defined in the MSA file so (and handled completely differently). Except there is a regression test in test_nirspec_masterbackgound.py (test_nirspec_mos_mbkg_user) that subtracts a background using a user provided background. In that case I assume the special MOS calspec2 stuff (subtracting the background) is not done and it is like other spec data. Would this plan cause problems with WFSS data (that I think also uses combine 1d to combine spectra)

jemorrison:
@hbushouse I removed the code that was causing problems. This PR was to determine the background from extended IFU dedicated backgrounds using sigma clipping. I don't have test data like this so I an uncertain what is set in the header that flags this type of data. Master_backbround_step will copy the 'background' column to the 'surf_bright' column for
model.meta.observation.bkgdtarg == False. I am not sure what this keyword is set for IFU dedicated background exposures.
I might need another check in the if statement in master_background_step

hbushouse:
> @hbushouse I removed the code that was causing problems. This PR was to determine the background from extended IFU dedicated backgrounds using sigma clipping. I don't have test data like this so I an uncertain what is set in the header that flags this type of data. Master_backbround_step will copy the 'background' column to the 'surf_bright' column for
> model.meta.observation.bkgdtarg == False. I am not sure what this keyword is set for IFU dedicated background exposures.
> I might need another check in the if statement in master_background_step

Yes, you will almost certainly need to expand that check logic, because any dedicated background exposure should have bkgdtarg=True, which means in the current code the data from background would not get copied to surf_bright, but for MIR_MRS (and presumably NRS_IFU too?) with bkgdtarg=True we now want the data copied from background to surf_bright.

jemorrison:
@hbushouse There is one regression test that I have a question about.
It is the test 'test_nirspec_ifu_mbkg_nod' which is the last test in test_nirspec_masterbackground.py
The asn file has 
1 science exposure: ifu_prism_source_on_NRS1_cal.fits
1 x1d background: ifu_prism_source_off_NRS1_x1d.fits
The x1d file has information in surf_bright and background column (which seems correct because the source_type = POINT).

The value the x1d meta.observation.bkgdtarg  is NoneType. So in masterbackground step the background is not copied to the surface brightness to be used as the background. The surf_bright is used. 
I don't know I get a bit confused on how background subtraction is done with nodded data. The background is from an "source_off" so maybe the surf_bright column does represent the background. 
I just wanted to double check this with you.

hbushouse:
> @hbushouse There is one regression test that I have a question about.
> It is the test 'test_nirspec_ifu_mbkg_nod' which is the last test in test_nirspec_masterbackground.py
> The asn file has
> 1 science exposure: ifu_prism_source_on_NRS1_cal.fits
> 1 x1d background: ifu_prism_source_off_NRS1_x1d.fits
> The x1d file has information in surf_bright and background column (which seems correct because the source_type = POINT).
> 
> The value the x1d meta.observation.bkgdtarg is NoneType. So in masterbackground step the background is not copied to the surface brightness to be used as the background. The surf_bright is used.
> I don't know I get a bit confused on how background subtraction is done with nodded data. The background is from an "source_off" so maybe the surf_bright column does represent the background.
> I just wanted to double check this with you.

That's a bad input x1d file being used in the test. It's missing the BKGDTARG keyword. It should be in the primary header and set to False (it's a nodded source exposure, not an exposure of a dedicated bkg target). If you do that, the existing code will then correctly copy the background column data to surf_bright and use it to form the master bkg spectrum. The fact that the BKGDTARG keyword is missing sounds like a previously uncaught error in the test and could lead to differences in results now when it starts being handled correctly.

jemorrison:
@hbushouse Should I got ahead and update the 5 x1d files on artifactory so the regression tests contain data in the background column ?

hbushouse:
yes

jemorrison:
new files uploaded to artifactory

jwst/extract_1d/ifu.py
```diff
@@ -495,31 +507,56 @@ def extract_ifu(input_model, source_type, extract_params):
...
                                             method=method, subpixels=subpixels)
             background[k] = float(bkg_table['aperture_sum'][0])
             temp_flux[k] = temp_flux[k] - background[k] * normalization
 
+        # Extended source data - background determined from sigma clipping
+        if source_type == 'EXTENDED':
+            bkg_data = data[k, :, :]
+            # pull out the data where weight map => 1. We do not want to use
+            # the edge data that is zero to define the statistics on clipping
+            bkg_stat_data = bkg_data[temp == 1]
```
hbushouse, line 533: I see `temp_flux` defined and used above, but I don't see `temp`. Would be nice to use a variable name that gives a better idea of what's in it.

jwst/extract_1d/ifu.py
```diff
@@ -1080,3 +1125,53 @@ def shift_ref_image(mask, delta_y, delta_x, fill=0):
     temp[..., oslice_y, oslice_x] = mask[..., islice_y, islice_x]
 
     return temp
+
+def sigma_clip_extended_region(data, mask_targ, wmap, sigma_clip):
+    """ sigma clip the extraction region
+
+    Parameters
+    ----------
+    data: ndarray, 3-D
```
hbushouse, line 1134: need whitespace before colon

jwst/extract_1d/ifu.py
```diff
@@ -1080,3 +1125,53 @@ def shift_ref_image(mask, delta_y, delta_x, fill=0):
...
+        Input data array to perform extraction from
+
+    mask_targ : ndarray, 2-D or 3-D
+        Mask of pixels defining the extended source region. A value of 1 indicated
+        pixel is in the extraction region.
+    wmap: ndarray, 3-D
```
hbushouse, line 1140: need whitespace before colon

hbushouse (commented):
Overall functionality looks OK, just needs a moderate amount of code cleanup.

jwst/extract_1d/ifu.py
```diff
@@ -299,9 +308,10 @@ def extract_ifu(input_model, source_type, extract_params):
     dq : ndarray, 1-D, uint32
         The data quality array.
 
-    npixels_annulus : ndarray, 1-D, float64
-        For each slice, this is the number of pixels that were added
-        together to get `temp_flux` for an annulus region.
+    npixels_bkg : ndarray, 1-D, float64
```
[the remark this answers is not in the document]
jemorrison: I was just matching what 'npixels' was. All the other parameters are float64. It probably does not matter if it float 32 or 64, but if we change it to float32 should we not also change 'npixels' to be consistent.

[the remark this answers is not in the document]
hbushouse: OK, I see that the new array is just a renamed version of the old one, so we might as well just leave things as they were (i.e. keep the float64).

jwst/extract_1d/ifu.py
```diff
@@ -495,31 +507,56 @@ def extract_ifu(input_model, source_type, extract_params):
...
                                             method=method, subpixels=subpixels)
             background[k] = float(bkg_table['aperture_sum'][0])
             temp_flux[k] = temp_flux[k] - background[k] * normalization
 
+        # Extended source data - background determined from sigma clipping
+        if source_type == 'EXTENDED':
+            bkg_data = data[k, :, :]
+            # pull out the data where weight map => 1. We do not want to use
+            # the edge data that is zero to define the statistics on clipping
+            bkg_stat_data = bkg_data[temp_weightmap == 1]
```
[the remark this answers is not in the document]
jemorrison: I will clean up the comment to make it clear. weight_map > = 1 is the same as temp_weightmap = 1 because of how  temp_weightmap was defined
temp_weightmap = weightmap[k,:,:]
        temp_weightmap[temp_weightmap>1] = 1

hbushouse (approved):

PR #5752, opened 2021-02-15
move output_ext to jwst.Step

nden:
This PR 

- defines a default `output_ext` from the base `stpipe.Step` class to `JWSTStep` to prevent romancal from writing FITS files by default. I tested this on one step using `strun`.
- Unpinned stdatamodels and tested with the latest release - 0.2.
- This revealed a straggler `Model._files_to_close` attribute which was replaced with the new `FileReference` object.

This PR depends on spacetelescope/stpipe#17

### DIFF
--- a/jwst/stpipe/core.py
+++ b/jwst/stpipe/core.py
@@ -14,6 +14,11 @@ from ..lib.suffix import remove_suffix
 
 
 class JwstStep(Step):
+    
+    spec = """
+    output_ext         = string(default='.fits')     # Default type of output
+    """
+
     @classmethod
     def _datamodels_open(cls, init, **kwargs):
         return datamodels.open(init, **kwargs)
--- a/jwst/stpipe/core.py
+++ b/jwst/stpipe/core.py
@@ -14,7 +14,7 @@ from ..lib.suffix import remove_suffix
 
 
 class JwstStep(Step):
-    
+
     spec = """
     output_ext         = string(default='.fits')     # Default type of output
     """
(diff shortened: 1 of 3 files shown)


stscieisenhamer (approved):

eslavich (approved):
This looks good, I think we'll want to keep `output_ext` on the `stpipe.Step` class but we should remove the default there so that subclasses are forced to specify it.

nden:
This should be merged after the change to stpipe.

hbushouse (approved):

setup.cfg
```diff
@@ -42,7 +42,7 @@ install_requires =
     requests>=2.22
     scipy>=1.1.0
     spherical-geometry>=1.2.18
-    stdatamodels~=0.1.0
+    stdatamodels
```
eslavich, line 45: stdatamodels 0.2.0 requires https://github.com/spacetelescope/jwst/pull/5512, which we'll hopefully merge today.  I'd like to require `<1.0` here and also for stpipe, in case we need to make breaking changes for romancal once jwst is frozen.

jwst/stpipe/core.py
```diff
@@ -14,6 +14,11 @@
 
 
 class JwstStep(Step):
+
+    spec = """
+    output_ext         = string(default='.fits')     # Default type of output
```
eslavich, line 19: I did some testing this morning and discovered that identically named parameters in subclasses automatically override the parent class's definition.  I think we can merge this without waiting for an stpipe release.

stscieisenhamer: Can confirm and is by design, I believe.

nden:
Any reason not to merge this?

eslavich:
Looks ready to go to me!

PR #5753, opened 2021-02-16
Remove hard-coded MRS outlier detection parameters

drlaw1558:
Removing hard-coded MRS outlier detection parameters that we inserted as a stopgap until parameter reference files were available (PR # 4778).  Now those files have been delivered, we should remove the stopgap code.


drlaw1558:
@jemorrison @hbushouse

jemorrison:
@drlaw1558  Can you remind me exactly which parameter reference file you updated to put these numbers in

drlaw1558:
New outlier rejection parameter reference file: https://jira.stsci.edu/browse/JWSTMIRI-195

hbushouse (approved):

jemorrison:
@drlaw1558 Is the reason NIRSPEC does not need a special parameter reference file is because the detaults  in the code spec : scale = string(default='0.5 0.4') are fine for NIRSPEC ?

drlaw1558:
@jemorrison To my knowledge NIRSpec have not tested the defaults in the code spec, but I doubt that those values would perform well given the undersampling of the instrument.

jemorrison (approved):
Looks good.

jemorrison:
@drlaw1558 before we forget we should probably ping a NIRSpec team member to study the outlier detection routine and decide if they also need a parameter reference file for NIRSpec IFU outlier detection. I just want to make sure the NIRSpec team is aware of this issue.

PR #5754, opened 2021-02-16
JP-1906: Fix handling of SRCTYPE for MOS

### LINKED ISSUE #5714: inconsistent flux units for NIRSpec extended source spectra
_Issue [JP-1906](https://jira.stsci.edu/browse/JP-1906) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

I'm seeing some inconsistent results regarding the flux units for NIRSpec MOS data with extended sources.  These are ground-based internal lamp exposures processed as if they were on-sky (EXP_TYPE = NRS_MSASPEC).  In one case (LINE1 lamp), the units look as expected - surface brightness, MJy/sr.  However, in another case (REF lamp), the units are in flux density, MJy, despite each source/shutter having been identified as EXTENDED by the srctype step.  It's possible I've done something wrong somewhere in creating the headers, but I can't seem to find anything amiss there.  I've also double-checked the photom and area reference files, those all have the expected (dummy) values.  In the verbose output, the value of PHOTMJSR given for each slit is 1, which is incorrect - it should be 4.25e10.

hbushouse:
This update to the srctype step simply unsets the value of the global target-based source_type in the data model when processing NIRSpec MOS exposures, so that it can't mistakenly get used later on, because MOS mode data should always and only use the slit-specific value of srctype. This was triggering a problem downstream in the photom step, which was making use of the model.meta.target.source_type value and ignoring the slit-specific values.

Fixes #5714 / [JP-1906](https://jira.stsci.edu/browse/JP-1906)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -71,6 +71,10 @@ srctype
 - Changed default SRCTYPE for non-primary NIRSpec slits in a FIXEDSLIT
   exposure to 'EXTENDED' rather than 'POINT' [#5671]
 
+- Changed logic for handling NIRSpec MOS exposures to blank out the "global"
+  value of SRCTYPE, to ensure that only the individual slit-specific values
+  of SRCTYPE get used downstream. [#5754]
+
 stpipe
 ------
 
--- a/jwst/srctype/srctype.py
+++ b/jwst/srctype/srctype.py
@@ -146,6 +146,10 @@ def set_source_type(input_model):
 
             log.info(f'source_id={slit.source_id}, stellarity={stellarity:.4f}, type={slit.source_type}')
 
+        # Remove the global target source type, so that it never mistakenly
+        # gets used for MOS data, which should always use slit-specific values
+        input_model.meta.target.source_type = None
+
     # Set all TSO exposures to POINT
     elif pipe_utils.is_tso(input_model):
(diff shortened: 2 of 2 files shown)


tapastro (approved):
LGTM - any need for this fix to be available to wfs modes? I see they also go through spec2 pipeline processing, but I'm not sure how they differ in processing/source assignment compared to MOS data. More of a question for my education, maybe...

hbushouse:
> LGTM - any need for this fix to be available to wfs modes? I see they also go through spec2 pipeline processing, but I'm not sure how they differ in processing/source assignment compared to MOS data. More of a question for my education, maybe...

WFSS modes currently don't make any use of SRCTYPE in any step of the pipeline, because up until recently there was no way to determine a SRCTYPE setting for each slit/source instance in a WFSS grism image. All of the extraction params are based solely on the size of the region assigned to each source in the segmentation map. However, thanks to a recent update to the source_type step that sets the value "is_star" in the output source catalog, we could actually use that value when the source catalog is read into the extract_2d step to process the grism exposure and set SRCTYPE for each slit/source in the extract_2d output. The value still wouldn't get used anywhere downstream, but at least it would get passed along as useful info to the user.

PR #5755, opened 2021-02-16
Remove vendorized configobj

eslavich:
I forgot to remove this when extracting code into stpipe.  We can use `stpipe.extern.configobj` instead.


jdavies-st (approved):
🎉 

You beat me to it!  😅

nden (approved):

PR #5759, opened 2021-02-17
JP-1841: Add source id to combine1d extension headers

### LINKED ISSUE #5605: Add useful slit/source-related keywords to combine1d headers
_Issue [JP-1841](https://jira.stsci.edu/browse/JP-1841) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The table extension header in combine_1d (c1d) output products currently doesn't contain any of the slit/source-related keywords that are available in the extension headers of the input x1d products. Some of these could be useful to users and hence should be copied over. Right now the step is setup to only copy meta data from the primary HDU when creating the c1d output product.

At the same time we should think about possibly changing the EXTNAME's of the table extensions in both x1d and c1d products from their current values (EXTRACT1D, COMBINE1D) to the generic value "SCI". This would greatly simplify schema construction (making it possible to reuse schemas for multiple data models) and would make the naming consistent with other science products.

tapastro:
Resolves [JP-1841](https://jira.stsci.edu/browse/JP-1841)
Resolves #5605 

The immediate urge to replace the entry in the multicombinedspec.schema under items with a $ref to combinedspec.schema is thwarted by the output file now containing extraneous IMAGE extensions, likely due to contained refs to core.schema or int_times.schema. As is, some redundant definitions are present.


CHANGES.rst
```diff
@@ -9,6 +9,12 @@ associations
 - Added new Lvl2 rule, Asn_Lv2NRSLAMPImage, to run Image2 pipeline for NRSLAMP
   exposures with OPMODE=image [#5740]
 
+combine1d
```
hbushouse, line 12: minor nit, but the full name for the step has an underscore in it, i.e. "combine_1d" (like extract_1d)

jwst/datamodels/schemas/combinedspec.schema.yaml
```diff
@@ -15,3 +15,9 @@ allOf:
       default: 1
       fits_keyword: SPORDER
       fits_hdu: COMBINE1D
+    source_id:
+      title: Source ID
+      type: integer
+      default: 0
```
hbushouse, line 21: I notice this entry has a default value, while the one in multicombinedspec does not. I think we could leave out the default value.

hbushouse (commented):
Just a couple very minor comments. Otherwise looks good.

jwst/datamodels/schemas/combinedspec.schema.yaml
tapastro: Thought I had removed those!

hbushouse (approved):

jdavies-st:
So looking at the regtests failures for this:

```
——————————————————————————————————————————— test_nis_wfss_spec3[s00002-c1d] ————————————————————————————————————————————
run_nis_wfss_spec3 = {'input': '/data1/jenkins/workspace/RT/JWST/clone/test_outputs/popen-gw18/test_niriss_wfss_run_nis_wfss_spec20/jw00625...',
 'truth_remote': 'jwst-pipeline/dev/truth/test_niriss_wfss/jw00625-o030_s00002_niriss_f090w-gr150c-gr150r_c1d.fits'}
suffix = 'c1d', source_id = 's00002'
fitsdiff_default_kwargs = {'atol': 1e-05, 'ignore_fields': ['DATE', 'CAL_VER', 'CAL_VCS', 'CRDS_VER', 'CRDS_CTX', 'NAXIS1', ...], 'ignore_hdus': ['ASDF'], 'ignore_keywords': ['DATE', 'CAL_VER', 'CAL_VCS', 'CRDS_VER', 'CRDS_CTX', 'NAXIS1', ...], ...}

    @pytest.mark.bigdata
    @pytest.mark.parametrize('suffix', ['cal', 'x1d', 'c1d'])
    @pytest.mark.parametrize('source_id', ['s00001', 's00002'])
    def test_nis_wfss_spec3(run_nis_wfss_spec3, suffix, source_id, fitsdiff_default_kwargs):
        """Regression test of the calwebb_spec3 pipeline applied to NIRISS WFSS data"""
    
        # Run the pipeline
        rtdata = run_nis_wfss_spec3
    
        # Get outputs and truths
        output = "jw00625-o030_" + source_id + "_niriss_f090w-gr150c-gr150r_" + suffix + ".fits"
        rtdata.output = output
        rtdata.get_truth("truth/test_niriss_wfss/" + output)
    
        # Compare the results
        fitsdiff_default_kwargs['atol'] = 1e-5
>       assert diff.identical, diff.report()
E       AssertionError: 
E          fitsdiff: 4.2
E          a: /data1/jenkins/workspace/RT/JWST/clone/test_outputs/popen-gw18/test_niriss_wfss_run_nis_wfss_spec20/jw00625-o030_s00002_niriss_f090w-gr150c-gr150r_c1d.fits
E          b: /data1/jenkins/workspace/RT/JWST/clone/test_outputs/popen-gw18/test_niriss_wfss_run_nis_wfss_spec20/truth/jw00625-o030_s00002_niriss_f090w-gr150c-gr150r_c1d.fits
E          HDU(s) not to be compared:
E           ASDF
E          Keyword(s) not to be compared:
E           CAL_VCS CAL_VER CRDS_CTX CRDS_VER DATE NAXIS1 TFORM*
E          Table column(s) not to be compared:
E           CAL_VCS CAL_VER CRDS_CTX CRDS_VER DATE NAXIS1 TFORM*
E          Maximum number of different data values to be reported: 10
E          Relative tolerance: 1e-05, Absolute tolerance: 1e-05
E         
E         Extension HDU 1 (COMBINE1D, 1):
E         
E              Headers have different number of cards:
E               a: 29
E               b: 28
E              Extra keyword 'SOURCEID' in a: 2
E         
E         Extension HDU 2 (COMBINE1D, 2):
E         
E              Headers have different number of cards:
E               a: 29
E               b: 28
E              Extra keyword 'SOURCEID' in a: 2
E         
E       assert False
E        +  where False = <astropy.io.fits.diff.FITSDiff object at 0x7fad7354d070>.identical
```

Does this seem reasonable?  I see that the value for both HDU of SOURCEID is 2.  Expected?

hbushouse:
Yes, that's the expected result. The c1d products are per source, so all data within a given c1d will have the same SOURCEID. And the reason for 2 extensions is because there are 2 spectral orders present (1 extension per order).

PR #5761, opened 2021-02-18
JP-1919: Fix photom handling of NRS IFU extended sources

### LINKED ISSUE #5735: problem with extended source flux units for NIRSpec MOS & IFU 2D spectra
_Issue [JP-1919](https://jira.stsci.edu/browse/JP-1919) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

There are some inconsistencies with how the flux units are applied to IFU and MOS data with extended sources.  For the IFU case, the SCI headers for the cal and s3d products correctly give BUNIT='MJy/sr'.  However, the value for PHOTMJSR is not correct - it appears to be equal to the value of PHOTMJ (1.0), while it should be PHOTMJ/PIXAR_SR.

In the MOS case, the BUNIT keyword for each slit has the incorrect value 'MJy', despite having SRCTYPE = EXTENDED.  PHOTMJSR again has a value of 1.0, although that is technically consistent with the BUNIT.

hbushouse:
Update the function in the photom step that handles NIRSpec IFU data, so that it checks for SRCTYPE='EXTENDED' and converts the calibration given in MJy to MJy/sr, and update all necessary keyword values appropriately. Previously it had been just applying the point-source calibration (in MJy) to all IFU data, regardless of source type.

Fixes #5735 / [JP-1919](https://jira.stsci.edu/browse/JP-1919)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -69,6 +69,12 @@ lib
 
 - Make get_wcs_values_from_siaf public for JSDP use [#5669]
 
+photom
+------
+
+- Fixed handling of NIRSpec IFU extended source data, so that the flux
+  calibration gets converted to surface brightness [#5761]
+
 ramp_fitting
 ------------
 
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -235,6 +235,8 @@ class DataSet():
             row = find_row(ftab.phot_table, fields_to_match)
             if row is None:
                 return
+
+            # MSA (MOS) data
             if (isinstance(self.input, datamodels.MultiSlitModel) and
                 self.exptype == 'NRS_MSASPEC'):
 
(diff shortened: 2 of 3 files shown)


hbushouse:
Reviewers: the real meat of the update starts here https://github.com/spacetelescope/jwst/pull/5761/files#diff-b9378ce0b10971befd7443bb2b038a52954fb85c6a06906b7d0d6db5cb6a4057R258

Most of the rest is just cosmetic cleanup of the code.

hbushouse:
And just to keep @jdavies-st happy, I'll see if I can add some unit test coverage.

tapastro (approved):
Code looks good to me. Just bring along a changelog update with that unit test. 😉

hbushouse:
Working on it ...

hbushouse:
OK, after doing some researching, the possibility of a unit test for NRS IFU mode is very low, due to the fact that the calibration requires a valid WCS, which has to be consistent with the fake data arrays, and an IFU WCS is *very* complex and hence would take a lot of work to construct on the fly. And for it to be truly valid it would have to be a full-frame 2048x2048 image, so storing an actual data file in the unit test directory would take up a fair amount of space. So I'm going to make it a regtest instead.

jdavies-st (commented):
It's hard to follow the logic of the setting of `unit_is_surface_brightness` as it only seems to be getting set in `else` blocks, but later on it is needed to be defined for an `if` block.

Would it be good to set it to `True` or `False` on `init` so we can guarantee it is always set to one or the other, then change it as needed in the `else` blocks?

hbushouse:
@jdavies-st it only gets used again with the same level "else" block in which it was set. Anywhere outside of that block it's not used.

Issue #5779, opened 2021-02-22
investigate duplicate product names found in test suite


stscijgbot-jp:
_Issue [JP-1936](https://jira.stsci.edu/browse/JP-1936) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

After resolution of JP-1903, it was noticed that a number of the regression tests are producing the duplicate product name warning. Investigate the causes.

stscijgbot-jp:
_Comment by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham) on [JIRA](https://jira.stsci.edu/browse/JP-1936?focusedCommentId=511430#comment-511430):_

Issue found is that the warnings are also produced for Level 2 associations. For Level 2, this is actually a common occurrence, handled by the generator, that does not warrant the warning message.

Note for SDP: The v1.0.0 release will produce these spurious warnings. This will be fixed in the patch.

stscijgbot-jp:
_Comment by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham) on [JIRA](https://jira.stsci.edu/browse/JP-1936?focusedCommentId=511599#comment-511599):_

Resolved by [PR#&#x2060;5780](https://github.com/spacetelescope/jwst/pull/5780)

stscijgbot-jp:
_Comment by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham) on [JIRA](https://jira.stsci.edu/browse/JP-1936?focusedCommentId=511430#comment-511430):_

Issue found is that the warnings are also produced for Level 2 associations. For Level 2, this is actually a common occurrence, handled by the generator, that does not warrant the warning message.

Note for SDP: The v1.0.0 release will produce these spurious warnings. This will be fixed in the patch.

stscijgbot-jp:
_Comment by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham) on [JIRA](https://jira.stsci.edu/browse/JP-1936?focusedCommentId=511599#comment-511599):_

Resolved by [PR#&#x2060;5780](https://github.com/spacetelescope/jwst/pull/5780)

PR #5721, opened 2021-02-05
JP-1903 Warn about duplicate product names and do not write duplicate associations

stscieisenhamer:
Resolves #5716 / [JP-1903](https://jira.stsci.edu/browse/JP-1903)

A warning will be generated that looks like the following. One of any duplicate association will still be created, normally the first one found.

```
RuntimeWarning: Duplicate associations exist: ['jw00632-o003_t001_nircam_f212n-wlp8-nrca1_{suffix}-05']
RuntimeWarning: Duplicates will be removed, leaving only one of each.
```

PR #5780, opened 2021-02-22
JP-1936 Ignore duplicate product names while handling Level 2 associations

stscieisenhamer:
Resolves [JP-1936](https://jira.stsci.edu/browse/JP-1936) / #5779 

JP-1903 #5721 introduced a warning when duplicate associations product names are found. For Level 3 this is a crucial warning. However, for Level 2, this warning is meaningless; duplicate product names are expected, since all product names are just exposure names. Hide the warning while processing Level 2 associations.

Side note: the duplicate products are filtered out, regardless of the warning level.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,11 +1,10 @@
 1.0.1 (unreleased)
 ==================
 
+associations
+------------
 
-
-
-
-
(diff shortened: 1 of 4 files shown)


hbushouse (approved):
Needs change log entry

PR #1918, opened 2018-04-19
Activate WFSS background subtraction

hbushouse:
Update the background_subtraction step to activate the loading and subtracting of WFSSBKG reference images for WFSS exposures. This simply involves uncommenting lines that were already included in the routine, but were waiting for the ref files to become available.

Fixes #1917.

PR #5781, opened 2021-02-23
JP-1918: Remove errant background exposures from Coron3 associations

### LINKED ISSUE #5732: Coron3 ASN lists x1dints product as background member
_Issue [JP-1918](https://jira.stsci.edu/browse/JP-1918) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

As noted in JWSTDMS-486 the coron3 ASN jw00628-c1023_coron3_001_asn.json correctly includes a background member from Obs 86, but the product type for background exposure is x1dints. Coronagraphy is an imaging mode and hence a cal (or calints) product should be used for background.

tapastro:
Resolves [#1918](https://jira.stsci.edu/browse/JP-1918)
Resolves #5732 

In the end, the solution to the errant x1dints entry in the initial association was to have it removed with a new constraint, as it was a background exposure. This caused a duplicate association warning, as the group association entries now exactly overlapped with the coronography association. The end result removes an association from the output of asn_generate.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,10 @@
 1.0.1 (unreleased)
 ==================
 
+associations
+------------
 
+- Constraint added to Asn_Lv3Coron to remove background exposures [#5781]
 
 
 
--- a/jwst/associations/lib/rules_level3.py
+++ b/jwst/associations/lib/rules_level3.py
@@ -166,6 +166,14 @@ class Asn_Lv3Coron(AsnMixin_Science):
                     force_reprocess=ProcessList.EXISTING,
                     only_on_match=True,
                 ),
+                Constraint(
+                    [DMSAttrConstraint(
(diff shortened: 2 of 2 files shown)


hbushouse (approved):
Requisite change log reminder ...

hbushouse:
You realize, of course, that as soon as we implement this, the Coron WG will make a request to allow for subtraction of dedicated background images during level 3 coron processing.  ;-)

tapastro:
> You realize, of course, that as soon as we implement this, the Coron WG will make a request to allow for subtraction of dedicated background images during level 3 coron processing. ;-)

Jonathan was curious why the onlyif rule was in place on the existing constraint in the association rule - it seems to specifically _allow_ background exposures in...

PR #5784, opened 2021-02-25
JP-1887 Refectored 'ols_ramp_fit' to be a more manageable function.

kmacdonald-stsci:
The `ols_ramp_fit` function was over 700 lines long.  I broke down each major step into functions to make the code easier to read.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -102,12 +102,12 @@ def ramp_fit(model, buffsize, save_opt, readnoise_model, gain_model,
         # Get readnoise array for calculation of variance of noiseless ramps, and
         #   gain array in case optimal weighting is to be done
         frames_per_group = model.meta.exposure.nframes
-        readnoise_2d, gain_2d = utils.get_ref_subs(model, readnoise_model,
-                                                   gain_model, frames_per_group)
+        readnoise_2d, gain_2d =\
+            utils.get_ref_subs(model, readnoise_model, gain_model, frames_per_group)
 
-        new_model, int_model, opt_model = \
-               ols_ramp_fit_multi(model, buffsize, save_opt, readnoise_2d,
-               gain_2d, weighting, max_cores)
+        new_model, int_model, opt_model =\
+            ols_ramp_fit_multi(model, buffsize, save_opt, readnoise_2d,
+                               gain_2d, weighting, max_cores)
         gls_opt_model = None
 
     # Update data units in output models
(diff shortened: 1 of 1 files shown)


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -538,72 +527,11 @@ def ols_ramp_fit(data, err, groupdq, inpixeldq, buffsize, save_opt, readnoise_2d
...
-        # Next block is to satisfy github issue 1681:
-        # "MIRI FirstFrame and LastFrame minimum number of groups"
-        if (ngroups < 2):
-            log.warning('MIRI datasets require at least 2 groups/integration')
-            log.warning('(NGROUPS), so will not process this dataset.')
-            return None, None, None
+        miri_ans = ramp_fit_miri_multigroups(data, err, groupdq, cubeshape, imshape, nreads, ngroups)
+        if miri_ans is None:
+            return None * 22
+        data, err, groupdq, cubeshape, imshape, nreads, ngroups, first_gdq = miri_ans
```
jdavies-st, line 540: `miri_ans` could be a little clearer what it is, though this can maybe be cleaned up when there's a data class for the returned items.

Is this `ramp_fit_miri_multigroups()` function only specific to MIRI?  If not, it might be good to give it a more general name, especially if it might be useful for Roman or any other sort of ramp fitting.

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -616,6 +544,249 @@ def ols_ramp_fit(data, err, groupdq, inpixeldq, buffsize, save_opt, readnoise_2d
...
+
+    Parameters
+    ----------
+    data: 4-D image cube with dimensions (integrations, groups, rows, columns)
+    err: Error array
+    groupdq: GroupDQ array
+    cubeshape: Image cube dimensions per integration
+    imshape: Image shape per group
+    nreads: Number of reads
+    ngroups: Number of groups
```
jdavies-st, line 650: Could you make these docstrings more `numpy` standard?:

```
data : np.ndarray
    4-D image cube with dimensions (integrations, groups, rows, columns)
err : np.ndarray
```

etc

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -2946,13 +3166,15 @@ def calc_num_seg(gdq, n_int):
     gdq : float, 3D array
         cube of GROUPDQ array for a data
 
-    n_int : int
+    n_int : int (unused)
```
jdavies-st, line 3175: If this is unused, is there a reason to keep it in the call signature?

kmacdonald-stsci: It's used in a test and I have not changed the test, yet.

jwst/ramp_fitting/ramp_fit.py
kmacdonald-stsci: Yes.  I will expand these to conform better to expected standards.

jwst/ramp_fitting/ramp_fit.py
kmacdonald-stsci: This function is only called for MIRI images that have more than 1 nreads.  The shape of the 4-D cube gets changed based on removing beginning and end groups in MIRI images.

PR #5787, opened 2021-02-25
JP-1911: Bump asdf==2.7.3 in requirements-sdp.txt

### LINKED ISSUE #5699: Datamodel I/O efficiencies causing error in copy()
There seems to be a datamodel issue with running the ramp output of the development version of MIRISim though any pipeline build >=7.6, which we can't pin down. This occurs when the dq_init step tries to make a copy of the ramp datamodel. The error can be recreated using the following code snipped with the ramp file available here (the trace is at the end of this report):

https://www.dropbox.com/s/nxzgahfveejzhqy/det_image_seq1_MIRIMAGE_F1000Wexp1.fits?dl=0 

`from jwst import datamodels`
`dm = datamodels.RampModel('det_image_seq1_MIRIMAGE_F1000Wexp1.fits')`
`dmc = dm.copy()`

I found an issue with a similar error, though not identical, in the following: 

https://github.com/spacetelescope/jwst/issues/5527

I tried the fix of setting SKIP_FITS_UPDATE environment variable to false to disable some of the I/O efficiencies and this fixes the problem described above so it appears to be related.

==========================================
  File "/Users/patrickkavanagh/anaconda3/anaconda3/envs/miricle.devel/lib/python3.8/site-packages/jwst/datamodels/model_base.py", line 352, in copy
    self.clone(result, self, deepcopy=True, memo=memo)
  File "/Users/patrickkavanagh/anaconda3/anaconda3/envs/miricle.devel/lib/python3.8/site-packages/jwst/datamodels/model_base.py", line 331, in clone
    instance = copy.deepcopy(source._instance, memo=memo)
  File "/Users/patrickkavanagh/anaconda3/anaconda3/envs/miricle.devel/lib/python3.8/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/Users/patrickkavanagh/anaconda3/anaconda3/envs/miricle.devel/lib/python3.8/copy.py", line 296, in _reconstruct
    value = deepcopy(value, memo)
  File "/Users/patrickkavanagh/anaconda3/anaconda3/envs/miricle.devel/lib/python3.8/copy.py", line 151, in deepcopy
    copier = getattr(x, "__deepcopy__", None)
  File "/Users/patrickkavanagh/anaconda3/anaconda3/envs/miricle.devel/lib/python3.8/site-packages/asdf/tags/core/ndarray.py", line 362, in __getattr__
    return getattr(self._make_array(), attr)
  File "/Users/patrickkavanagh/anaconda3/anaconda3/envs/miricle.devel/lib/python3.8/site-packages/asdf/tags/core/ndarray.py", line 267, in _make_array
    self._array = np.ndarray(
ValueError: strides is incompatible with shape of requested array and size of buffer

jdavies-st:
Pulls in the bugfix release `asdf 2.7.3` into `requirements-sdp.txt`.  Previously, the `asdf` library was writing out unreadable FITS files for numpy arrays that were views over a larger arrays and were stored in as `AsdfInFITS`.

Resolves #5699 / [JP-1911](https://jira.stsci.edu/browse/JP-1911)

### DIFF
--- a/requirements-sdp.txt
+++ b/requirements-sdp.txt
@@ -4,7 +4,7 @@
 #     pip install pytest-xdist
 #     pip freeze | grep -v jwst.git >> requirements-sdp.txt
 apipkg==1.5
-asdf==2.7.2
+asdf==2.7.3
 astropy==4.2
 attrs==20.3.0
 certifi==2020.12.5


eslavich (approved):
LGTM

PR #5788, opened 2021-02-25
NIRISS WFSS set SRC_TYPE in extract_2d JP-1917

### LINKED ISSUE #5731: SRCTYPE header keyword not populated in x1d files (NIRISS)
_Issue [JP-1917](https://jira.stsci.edu/browse/JP-1917) was created on JIRA by [Camilla Pacifici](https://jira.stsci.edu/secure/ViewProfile.jspa?name=cpacifici):_

The SRCTYPE header keyword in each extension of the x1d files is populated with 'UNKNOWN', instead of using the information from the column 'is_star' in the ecsv catalog used for extraction.

For the JWST data analysis tools and for creating parsers for all the instruments and modes, it would be important to have this keyword populated consistently among the different instruments.

jemorrison:
This pr concerns JP-1917.
The GrismObject defined in transforms/model.py was updated to include 'is_star'
assign_wcs/util.py also had to be updated to pass this information along.

The code that sets SRC_TYPE for each slit in a NIRISS WFSS is in extract_2d/grism.py

A question I have is in the srctype step there is code that sets input_model.meta.target.source_type = src_type.
The log message to screen for running NIRISS WFSS in spec2 using a source catalog reports:
2021-02-25 15:56:15,579 - stpipe.Spec2Pipeline.srctype - WARNING - EXP_TYPE NIS_WFSS not applicable to this operation
2021-02-25 15:56:15,580 - stpipe.Spec2Pipeline.srctype - WARNING - Setting SRCTYPE = UNKNOWN
But I am not sure what input_model.meta.target.source_type is For NIRISS WFSS. We need to set the SRCTYPE for each slit not the input_model.meta.target.source_type.
Running this branch on NIRISS WFSS data results in the x1d files with SRCTYPE = 'POINT' for the is_star = True values in the source catalog.

Fixes #5731 / [JP-1917](https://jira.stsci.edu/browse/JP-1917)


jemorrison:
@hbushouse  @sosey  I don't think these changes will affect other GRISM type data. The GrismObject may contain the is_star information but it is only used in extract_2d/grism.py so I don't think these changes will break any thing else.

hbushouse:
How about the idea of adding `is_star` to the `SlitModel` schema, but only as a model attribute and not tied to a FITS keyword (i.e. don't give it a `fits_keyword` entry in the schema definition), and set it for each WFSS slit instance created in extract_2d/grisms.py (it would not get set for any other exposure types that use `SlitModel`), such that we can then add a special WFSS processing block in the `srctype` step to simply set the SRCTYPE keyword for each slit instance based on the value of `is_star`.

I know it's a bit hokey to define a new model attribute just to carry one piece of info to the very next step where it'll get replaced or superseded by another model attribute, but at least that way we maintain consistency in terms of having the `srctype` step actually set the SRCTYPE keyword value for all exposure types (rather than having it happen in `extract_2d` for WFSS exposures).

jemorrison:
Just to be clear we would leave the updates to the GrismObject into include is_star (since that information is read from the source catalog). But add is_star to slit_model and set it in extract_2d/grism.py Then we can set the SRCTYPE in the scrtype step where it really should be.
That is cleaner than setting SRCTYPE in extract_2d

hbushouse:
> Just to be clear we would leave the updates to the GrismObject into include is_star (since that information is read from the source catalog). But add is_star to slit_model and set it in extract_2d/grism.py Then we can set the SRCTYPE in the scrtype step where it really should be.
> That is cleaner than setting SRCTYPE in extract_2d

Right. Include `is_star` in the GrismObjects when they're created, then have `extract_2d` store the value of `is_star` into the slit.meta when it creates each slit instance, then finally have `srctype` read `is_star` from the slit meta and set SRCTYPE. The operation in `srctype` would then look very much like what it does for NIRSpec MOS data, where it loops over all the slits contained in the input `MultiSlitModel` and sets SRCTYPE for each slit based on the value of `stellarity`. In this case, for WFSS data, it would use `is_star` instead of `stellarity`.

jwst/extract_2d/grisms.py
```diff
@@ -476,7 +476,10 @@ def extract_grism_objects(input_model,
                 # The overall subarray offset is recorded in model.meta.subarray.
                 # nslit = obj.sid - 1  # catalog id starts at zero
                 new_slit.name = "{0}".format(obj.sid)
-                new_slit.source_type = 'UNKNOWN'
+                if obj.is_star:
+                    new_slit.is_star = True
```
hbushouse, line 480: I don't think any logic is needed here to make a choice of what to set, but instead simply set `new_slit.is_star = obj.is_star`, in order to carry along the value of `is_star` to the `srctype` step, where it will set the appropriate value of source_type.

jwst/srctype/srctype.py
```diff
@@ -156,6 +156,14 @@ def set_source_type(input_model):
         log.info(f'Input is a TSO exposure; setting default SRCTYPE = {src_type}')
         input_model.meta.target.source_type = src_type
 
+    # FOR NIR_WFSS check slit values of is_star to set SRCTYPE
```
hbushouse, line 159: This needs to be done for all WFSS modes, so that includes `NRC_WFSS` and `NIS_WFSS`. This means that some of the change log entries above need to be updated to indicate that these changes apply to WFSS in general, not just NIS_WFSS.

hbushouse (changes requested):

jwst/extract_2d/grisms.py
jemorrison: yes I meant to remove the settling of source type. Some how my push did not include this update

jwst/srctype/srctype.py
jemorrison: I am not all the familiar with WFSS mode. I did a search on WFSS in the pipeline and frequently get these modes:
WFSS_EXPTYPES = ['NIS_WFSS', 'NRC_WFSS', 'NRC_GRISM', 'NRC_TSGRISM']
NRC_TSGRISM is a mode that is already handled in the srctype code. But I could not find NRC_GRISM 
Do I need to add it to this   too ?

CHANGES.rst
```diff
@@ -8,12 +8,30 @@ associations
...
 
+datamodels
+----------
+
+- Added is_star to slitmeta [#5788]
+
+extract_2d
+----------
+
+- For NIS_WFSS removed setting sorce_type to UNKNOW, added setting is_star to slitmeta [#5788]
```
hbushouse, line 19: this should be for all WFSS now, not just NIS_WFSS

hbushouse (commented):
Looks good, except for one minor wording issue in change log.

hbushouse (approved):

PR #5794, opened 2021-02-26
TST: Fix cancel workflow for PR

pllim:
Similar to astropy/astropy#11330

cc @jdavies-st 

p.s. I think this should be the last of the Action stuff, so you will probably not get PRs from me no more. Merge this and enjoy the peace!


jdavies-st (approved):
🎉 Thanks @pllim !

PR #5805, opened 2021-03-03
Make PEP8 checks more strict

jdavies-st:
You may have assumed we were already doing these checks, but we were not.  This adds the following checks to our automated `flake8` checks, and updates the code to fix violations of this via `autopep8`:

```
E201 | Whitespace after '('
E202 | Whitespace before ')'
E211 | Whitespace before '('
E221 | Multiple spaces before operator
E222 | Multiple spaces after operator
E271 | Multiple spaces after keyword
E272 | Multiple spaces before keyword
E273 | Tab after keyword
E274 | Tab before keyword
E275 | Missing whitespace after keyword
E701 | Multiple statements on one line (colon)
E702 | Multiple statements on one line (semicolon)
E703 | Statement ends with a semicolon
E704 | Multiple statements on one line (def)
E712 | Comparison to true should be 'if cond is true:' or 'if cond:'
E713 | Test for membership should be 'not in'
E714 | Test for object identity should be 'is not'
```

See https://www.flake8rules.com for details.

Using `autopep8` like:

```
autopep8 jwst -i -r --select E27
```

for most of these, and by hand on the few cases where it refused to offer a solution.

My plan was to do this just for `resample` before doing some refactoring, but I got a bit carried away.  And we should be doing these checks anyway.


dmggh:
I can tolerate all of these style enforcements, with the exceptions of these 2:

E201 | Whitespace after '('
E202 | Whitespace before ')'

When I code or read anyone else's code, I find it *extremely* annoying to have to parse through statements that do not have whitespace after '(' and before ')' as too many characters are often jammed together. When coding I explicitly violate these 2 checks so that it is easier to read.

I consider :
      myfunction( argument1, argument2, argument3 )
much easier to read than :
     myfunction(argument1, argument2, argument3)

The whitespace makes code *much* easier to read, and as long as the convention used is consistent within each module, it shouldn't matter.

The purpose of *style* checking should be some standardization to make code easier to read for everyone. These are not syntax errors, so why enforce them ? Style should not be treated as syntax.

Just my 2 cents.

jdavies-st:
> I consider :
> myfunction( argument1, argument2, argument3 )
> much easier to read than :
> myfunction(argument1, argument2, argument3)

This PEP8 standard is pretty non-controversial.  I.e. I've never seen any major Python project violate it.  We shouldn't be either.  Also, we're [supposed to be following PEP8](https://github.com/spacetelescope/style-guides).

If you find the standard PEP8 compliant code hard to read, maybe use a syntax highlighter?  All editors have them, even `vim` and `emacs` have highlighters for python available.

```python
def myfunction(argument1, argument2, argument3):
    pass
```

We all share maintenance of the whole code base, so it's best to be PEP8 compliant, so that we are all coding to the same style standards, otherwise half the diffs in a PR are each coder's preferences for numbers or spaces or indenting, which makes doing code review rather labor intensive if lots of the changes are just style preferences.  A major point of PEP8 is to make the code easy to read and the style as standardized as possible so that any Python developer can be dropped into a project and not look around in horror.

jdavies-st:
Codecov failures here are expected as there were a good number of 1-line `if` statements that got converted to 2 lines, and none of that code is covered in tests, so we see the number of uncovered lines of code there double for each instance of this.

hbushouse (approved):
I've often seen infractions by eye when reviewing PR's and wondered why they weren't getting flagged by the codestyle check. Now I know. Thanks for updating.

PR #5812, opened 2021-03-05
Finish adding PEP8 checks

jdavies-st:
This is a follow on to #5805.  Same method.  This PR adds all the remaining flake8 (pycodestyle) `E` checks except  spacing around commas, which for numpy indices can be a bit long if we enforce them.  This also means we are not enforcing single spaces after commas in arg lists, unfortunately.

```
E231 | Missing whitespace after ',', ';', or ':'
E241 | Multiple spaces after ','
```
Addresses #2552


jwst/pathloss/pathloss.py
```diff
@@ -130,7 +130,7 @@ def calculate_pathloss_vector(pathloss_refdata,
         crval3 = pathloss_wcs.crval3
         cdelt3 = pathloss_wcs.cdelt3
         for i in np.arange(wavesize):
-            wavelength[i] = crval3 +(float(i+1) - crpix3) * cdelt3
+            wavelength[i] = crval3 + (float(i+1) - crpix3) * cdelt3
```
kmacdonald-stsci, line 133: There should be spaces before and after the '+'.

jwst/ramp_fitting/utils.py
```diff
@@ -51,16 +51,15 @@ def __init__(self, n_int, imshape, max_seg, nreads, save_opt):
...
-            self.sigyint_seg=np.zeros((n_int,)+(max_seg,)+imshape,dtype=np.float32)
-            self.sigslope_seg=np.zeros((n_int,)+(max_seg,)+imshape,dtype=np.float32)
-            self.inv_var_seg=np.zeros((n_int,)+(max_seg,)+imshape,dtype=np.float32)
-            self.firstf_int=np.zeros((n_int,) + imshape, dtype=np.float32)
-            self.ped_int=np.zeros((n_int,) + imshape, dtype=np.float32)
-            self.cr_mag_seg=np.zeros((n_int,)+(nreads,)+imshape, dtype=np.float32)
-            self.var_p_seg=np.zeros((n_int,)+(max_seg,)+imshape,dtype=np.float32)
-            self.var_r_seg=np.zeros((n_int,)+(max_seg,)+imshape,dtype=np.float32)
-
+            self.yint_seg = np.zeros((n_int,)+(max_seg,)+imshape,dtype=np.float32)
```
kmacdonald-stsci, line 54: There should be spaces before and after the '+'.

jwst/ramp_fitting/utils.py
```diff
@@ -612,8 +605,8 @@ def calc_slope_vars(rn_sect, gain_sect, gdq_sect, group_time, max_seg):
     #   checked for and handled later
     warnings.filterwarnings("ignore", ".*invalid value.*", RuntimeWarning)
     warnings.filterwarnings("ignore", ".*divide by zero.*", RuntimeWarning)
-    den_r3[wh_seg_pos] = 1./(segs_beg_3[wh_seg_pos] **3. -
-                               segs_beg_3[wh_seg_pos]) # overwrite where segs>1
+    den_r3[wh_seg_pos] = 1./(segs_beg_3[wh_seg_pos] ** 3. -
```
kmacdonald-stsci, line 608: There should be spaces before and after the '/'.

kmacdonald-stsci (approved):

jwst/ami/analyticnrm2.py
```diff
@@ -30,7 +30,7 @@ def jinc(x, y):
         2d jinc at the given coordinates, with NaNs replaced by pi/4.
     """
     R = (jinc.d / jinc.lam) * jinc.pitch *  \
-           np.sqrt((x - jinc.offx)*(x - jinc.offx) + \
+           np.sqrt((x - jinc.offx)*(x - jinc.offx) +
```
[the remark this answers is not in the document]
jdavies-st: I hadn't done the `autopep8` for E226. Agree that was an oversight.  I've ran those changes, and it corrects all the issues you saw.  Thanks!

jdavies-st:
I'm not sure why the project Codecov measurements are down.  The patch failure is understandable, because I'm changing a lot of lines of code that are untested in the unit tests.  Maybe this explains the project level failure as well?  Once the regtests run, we'll be back to the original coverage?

jwst/ami/ami_analyze.py
```diff
@@ -54,9 +54,9 @@ def apply_LG_plus(input_model, filter_model, oversample, rotation,
         ysize = 80
         xstop = xstart + xsize - 1
         ystop = ystart + ysize - 1
-        input_model.data = input_model.data[ystart-1:ystop, xstart-1:xstop]
-        input_model.dq = input_model.dq[ystart-1:ystop, xstart-1:xstop]
-        input_model.err = input_model.err[ystart-1:ystop, xstart-1:xstop]
+        input_model.data = input_model.data[ystart - 1:ystop, xstart - 1:xstop]
```
hbushouse, line 57: This is actually one place where my personal preference is to disagree with codestyle, in that putting in white space then "attaches" the "1" visually with the 2nd argument of the slice, rather than keeping it with the first. But I'm willing to forgo personal preference in this case.

jwst/cube_build/cube_build_wcs_util.py
```diff
@@ -62,7 +61,7 @@ def find_corners_MIRI(input, this_channel, instrument_info, coord_system):
         valid = np.where(yedge < 1023)
         xedge = xedge[valid]
         yedge = yedge[valid]
-        coord2b, coord1b, lamb = detector2alpha_beta(xedge,yedge) #lam ~0 for this transform
+        coord2b, coord1b, lamb = detector2alpha_beta(xedge,yedge)  # lam ~0 for this transform
```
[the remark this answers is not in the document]
hbushouse: Huh, how did that sneak past the checker?

jwst/ami/ami_analyze.py
jdavies-st: Because we are specifically ignoring E231 (enforce whitespace after `,` and `:`), I did not bother to run `autopep8` with this code to do any corrections. We could turn it on and do the checks.  This would then enforce that the following:

```python
input_model.dq = input_model.dq[ystart-1:ystop, xstart-1:xstop]
```

becomes

```python
input_model.data = input_model.data[ystart - 1: ystop, xstart - 1: xstop]
```

This is where going part way, encourages one to go all the way.  Thoughts?  I find either of these more readable than what I corrected it to, but we now don't allow the former because we want spaces between arithmatic operators for this PR.  Hmm.

jwst/cube_build/cube_build_wcs_util.py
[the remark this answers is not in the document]
jdavies-st: Yeah, because we are not enforcing E231, space after the comma, I didn't run `autopep8` with this code.  Clearly in this case it would be preferable.  But this would also enforce slicing to be done with spaces after both `,` and `:`.

jwst/ami/ami_analyze.py
hbushouse: Personally I think the first example above is fine. The 2nd ends up with everything so spread out that it actually makes it harder to follow (IMHO). I think space before and after arithmetic operators should only be enforced when used as part of an actual statement that's computing something, but not when they're used with arguments to a function or slicing.

By the way, is there an "Oxford comma" option for the checker to apply?  I'm a BIG believer in that one.

jwst/cube_build/cube_build_wcs_util.py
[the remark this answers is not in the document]
hbushouse: Does E231 only apply to argument lists or anything with a comma?  Personally I think it's useful to enforce this one (makes the argument list easier for humans to parse).

jwst/ami/ami_analyze.py
jdavies-st: I agree.

jwst/cube_build/cube_build_wcs_util.py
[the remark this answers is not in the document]
jdavies-st: Unfortunately, E231 is an all-or-nothing check.  You either enforce spacing for arg lists and numpy array slicing or you don't.  It enforces a space after `,`, `:`, `;` anytime they're used.

hbushouse (approved):
I , hereby ap prove these + change-s to format ting .

PR #5818, opened 2021-03-08
Add module for the starfinder kernel

larrybradley:
This PR adds a new private module for the starfinder kernel, which prevents importing a fragile private class from `photutils` (which recently moved the private class, breaking code here).  See

https://github.com/astropy/photutils/pull/1173
https://github.com/astropy/photutils/pull/1175
https://github.com/astropy/photutils/pull/1177

cc: @jdavies-st

### DIFF
--- a/jwst/source_catalog/source_catalog.py
+++ b/jwst/source_catalog/source_catalog.py
@@ -22,7 +22,6 @@ from photutils import __version__ as photutils_version
 from photutils import Background2D, MedianBackground
 from photutils import detect_sources, deblend_sources, source_properties
 from photutils import CircularAperture, CircularAnnulus, aperture_photometry
-from photutils.detection.findstars import _StarFinderKernel
 from photutils.utils._wcs_helpers import _pixel_scale_angle_at_skycoord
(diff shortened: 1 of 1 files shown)


jdavies-st (commented):
What's the reason that we can't just call the public class `DAOStarFinder` in `jwst`?  Are we doing something that's not part of the public interface?  And if so, should it be part of the public interface?

larrybradley:
@jdavies-st The source catalog step uses segmentation to detect all sources, `DAOStarFinder` only finds stars :smiley:.  However, the working group did want to get the DAO sharpness and roundness parameters from the source pixels with the segmentation map -- hence the need for a hybrid approach here.

I've updated this PR to replace the private `_StarFinderKernel` (basically a 2D circular Gaussian normalized to zero) in the code using astropy's `Gaussian2DKernel`.

jdavies-st (approved):
Looks good.  Thanks @larrybradley 

I ran the NIRCam imaging regression tests that run `source_catalog` with this branch, and all looks good.

Issue #5701, opened 2021-02-05
Increase test coverage of ramp_fitting


stscijgbot-jp:
_Issue [JP-1887](https://jira.stsci.edu/browse/JP-1887) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

Study the code in jwst.ramp_fitting and increase test coverage.

Identify code representing standalone algorithms for possibly including in a shared library.

PR #5819, opened 2021-03-08
JP-1887: Refactoring `ols_ramp_fit` into smaller more manageable functions

kmacdonald-stsci:
Addresses #5701 / [JP-1887](https://jira.stsci.edu/browse/JP-1887)

This is one in a series of PR's designed to refactor ramp fitting to make it more readable, maintainable, and reusable.  

This PR is used to break down the function `ols_ramp_fit`.  It is over 700 lines long, but has four distinct steps, which have been encapsulated in four different functions that `ols_ramp_fit` calls.  
- `ramp_fit_miri_multigroups()`, deals with the special case of MIRI data with multiples groups in an integration, resizing the data for discarded groups.
- `ramp_fit_first_pass()`, does the "first pass" over the data to compute the initial values related to slope computation.
- `ramp_fit_second_pass()`, does the "second pass" over the data using the intermediate results from the first pass to compute values related to variances of estimates.
- `ramp_fit_overall_slopes_and_variances()`, uses the values computed in the first and second passes over the date to compute the overall slopes and variances for each pixel.

### DIFF
--- a/jwst/ramp_fitting/ramp_fit.py
+++ b/jwst/ramp_fitting/ramp_fit.py
@@ -433,7 +433,6 @@ def create_output_models(input_model, number_of_integrations, save_opt, total_co
     return int_model, opt_model, out_model
 
 
-# ols_ramp_fit BEGIN
 def ols_ramp_fit(data, err, groupdq, inpixeldq, buffsize, save_opt, readnoise_2d, gain_2d,
                  weighting, instrume, frame_time, ngroups, group_time, groupgap, nframes,
                  dropframes1, int_times):
@@ -550,10 +549,10 @@ def ols_ramp_fit(data, err, groupdq, inpixeldq, buffsize, save_opt, readnoise_2d
 
     if instrume == 'MIRI' and nreads > 1:
 
-        # The function could return None if the removed groups leaves no data to be 
-        # processed.  If this is the case, return None for all expected variables 
+        # The function could return None if the removed groups leaves no data to be
+        # processed.  If this is the case, return None for all expected variables
         # returned by ramp_fit
-        miri_ans = ramp_fit_miri_multigroups(data, err, groupdq, cubeshape, 
+        miri_ans = ramp_fit_miri_multigroups(data, err, groupdq, cubeshape,
                                              imshape, nreads, ngroups)
         if miri_ans is None:
             return None * 22
@@ -584,13 +583,13 @@ def ols_ramp_fit(data, err, groupdq, inpixeldq, buffsize, save_opt, readnoise_2d
     #   median slopes from the 'First Pass'. These variances are due to Poisson
     #   noise only, read noise only, and the combination of Poisson noise and
     #   read noise. The integration-specific variances are 3D arrays, and the
-    #   segment-specific variances are 4D arrays . 
+    #   segment-specific variances are 4D arrays.
     second_pass_ans = ramp_fit_second_pass(
             n_int, nrows, groupdq, cubeshape, imshape, gain_2d, readnoise_2d,
             group_time, max_seg, med_rates, num_seg_per_int)
 
-    var_p3, var_r3, var_p4, var_r4, var_both4, var_both3, inv_var_both4, \
-        s_inv_var_p3, s_inv_var_r3, s_inv_var_both3 = second_pass_ans 
+    var_p3, var_r3, var_p4, var_r4, var_both4, var_both3, inv_var_both4 = second_pass_ans[:7]
+    s_inv_var_p3, s_inv_var_r3, s_inv_var_both3 = second_pass_ans[7:]
 
 
     # Now that the segment-specific and integration-specific variances have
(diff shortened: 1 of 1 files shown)


jwst/ramp_fitting/ramp_fit.py
```diff
@@ -549,79 +537,332 @@ def ols_ramp_fit(data, err, groupdq, inpixeldq, buffsize, save_opt, readnoise_2d
...
+    groupdq: ndarray
+        GroupDQ array
 
-        # Next block is to satisfy github issue 1681:
-        # "MIRI FirstFrame and LastFrame minimum number of groups"
-        if ngroups < 2:
-            log.warning('MIRI datasets require at least 2 groups/integration')
-            log.warning('(NGROUPS), so will not process this dataset.')
-            return None, None, None
+    cubeshape: tuple
```
jdavies-st, line 661: Is `cubeshape` deterministically derived from `data.shape` in some way?  From the comment, it looks like `cubeshape = data[0].shape`.  If that's the case, then this doesn't need to be passed separately?

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -549,79 +537,332 @@ def ols_ramp_fit(data, err, groupdq, inpixeldq, buffsize, save_opt, readnoise_2d
...
+        Image cube dimensions per integration
 
-    if ngroups == 1:
-        log.warning('Dataset has NGROUPS=1, so count rates for each integration')
-        log.warning('will be calculated as the value of that 1 group divided by')
-        log.warning('the group exposure time.')
+    imshape: tuple
```
jdavies-st, line 664: Same question here.  Is `imshape = data[0][0].shape`?

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -549,79 +537,332 @@ def ols_ramp_fit(data, err, groupdq, inpixeldq, buffsize, save_opt, readnoise_2d
...
+        Image shape per group
 
-    # Calculate effective integration time (once EFFINTIM has been populated
-    #   and accessible, will use that instead), and other keywords that will
-    #   needed if the pedestal calculation is requested. Note 'nframes'
-    #   is the number of given by the NFRAMES keyword, and is the number of
-    #   frames averaged on-board for a group, i.e., it does not include the
-    #   groupgap.
+    nreads: int
```
jdavies-st, line 667: Can this be derived from `data.shape` as well?

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -549,79 +537,332 @@ def ols_ramp_fit(data, err, groupdq, inpixeldq, buffsize, save_opt, readnoise_2d
...
+        Number of reads
+
+    ngroups: int
```
jdavies-st, line 670: And this?

jwst/ramp_fitting/ramp_fit.py
```diff
@@ -1193,6 +1550,12 @@ def gls_ramp_fit(input_model, buffsize, save_opt, readnoise_model, gain_model, m
     gain_model : instance of gain model
         Gain for all pixels.
 
+    max_cores : string
+        Number of cores to use for multiprocessing. If set to 'none' (the default),
+        then no multiprocessing will be done. The other allowable values are 'quarter',
+        'half', and 'all'. This is the fraction of cores to use for multi-proc. The
+        total number of cores includes the SMT cores (Hyper Threading for Intel).
+
```
jdavies-st, line 1558: Does `gls_ramp_fit` actually allow multiprocessing?  I didn't think it did.

jdavies-st (commented):
Looks good Ken.  Just a couple questions below.

jwst/ramp_fitting/ramp_fit.py
kmacdonald-stsci: I didn't add the parameter to the function.  I simply added to the doctoring to capture all the parameters.  I have not looked at this function.  The only reason I noticed it is because I created the function `compute_slices`.

jwst/ramp_fitting/ramp_fit.py
kmacdonald-stsci: The variable `groups` is gotten from `input_model.meta.exposure.ngroups` in the `RampModel` passed to `ramp_fit`.  It is not part of the shape of `data`.

jwst/ramp_fitting/ramp_fit.py
kmacdonald-stsci: Yes. That's part of what I'm currently refactoring. The primary focus of this PR is to break down ols_ramp_fit into a more manageable size calling sub-functions. The follow on PR's that will come will contain the type of refactoring you are describing, simplifying the variables used and passed around.

jwst/ramp_fitting/ramp_fit.py
kmacdonald-stsci: Yes. That's part of what I'm currently refactoring. The primary focus of this PR is to break down ols_ramp_fit into a more manageable size calling sub-functions. The follow on PR's that will come will contain the type of refactoring you are describing, simplifying the variables used and passed around.

jwst/ramp_fitting/ramp_fit.py
hbushouse: Be very careful with the use and meaning of `nreads` vs `ngroups`. In the instruments, individual detector reads can be averaged on-board and then downlinked as a single group. When the data hit the ground they are 4-dimensional: [ncols, nrows, ngroups, nints] in conventional (e.g. FITS) notation, which is [nints, ngroups, nrows, ncols] in python array notation. The data in the `ngroups` dimension can consist of 1 or more detector reads. So in this particular use in the `ramp_fit` step I'm guessing `nreads` refers to the actual number of detector readouts that took place in the instrument, which is not necessarily the same as `ngroups` (because, as mentioned above, 1 or more reads can get averaged into a group). So `nreads` can't reliably be determined from the shape of the 4-D data array. Only `ngroups` can.

kmacdonald-stsci: Upon clarification, the use of `nreads` and `ngroups` are the same values taken from two different locations in the `RampModel` class.  in effect, these two variables are redundant that makes the code confusing.  I will remove this redundancy and use the correct `ngroups` variable name.

jdavies-st (approved):
Cool.  I think this is good to go then as is.

Issue #3962, opened 2019-08-29
interpolate_onto_grid function in pathloss.py doesn't return correct interpolation values


mfixstsci:
`interpolate_onto_grid` does a linear interpolation using array slicing/indexing to avoid element wise looping over the wavelength grid. Using the parameters provided in `test_interpolate_onto_gird` in `test_pathloss.py` (https://github.com/spacetelescope/jwst/pull/3871/files#diff-e233168f815af7352933af27315ac47bR283) it is easy to predict the result of a basic linear interpolation.  We expect the output array to look like:

```
[[1.1 2.2 3.3 4.4 5.5 6.6 7.7 8.8 9.9 nan]
 [nan nan nan nan nan nan nan nan nan nan]
 [nan nan nan nan nan nan nan nan nan nan]
 [nan nan nan nan nan nan nan nan nan nan]
 [nan nan nan nan nan nan nan nan nan nan]
 [nan nan nan nan nan nan nan nan nan nan]
 [nan nan nan nan nan nan nan nan nan nan]
 [nan nan nan nan nan nan nan nan nan nan]
 [nan nan nan nan nan nan nan nan nan nan]
 [nan nan nan nan nan nan nan nan nan nan]]
```

which match the results from `np.iterp`. Currently the array that is returned from `interpolate_onto_grid` with these same parameters looks like:

```
[[ 1.11  2.44  3.99  5.76  7.75  9.96 12.39 15.04 17.91   nan]
 [  nan   nan   nan   nan   nan   nan   nan   nan   nan   nan]
 [  nan   nan   nan   nan   nan   nan   nan   nan   nan   nan]
 [  nan   nan   nan   nan   nan   nan   nan   nan   nan   nan]
 [  nan   nan   nan   nan   nan   nan   nan   nan   nan   nan]
 [  nan   nan   nan   nan   nan   nan   nan   nan   nan   nan]
 [  nan   nan   nan   nan   nan   nan   nan   nan   nan   nan]
 [  nan   nan   nan   nan   nan   nan   nan   nan   nan   nan]
 [  nan   nan   nan   nan   nan   nan   nan   nan   nan   nan]
 [  nan   nan   nan   nan   nan   nan   nan   nan   nan   nan]]
```

I spoke with @stscirij and he believes that the bug might be caused by the result of the `fraction` variable calculated in `interpolate_onto_grid`.

hbushouse:
[...]

PR #5821, opened 2021-03-09
Removed skip for test_interpolate_onto_grid, function works as designed

stscirij:
This addresses #3962.  The test passes and I checked the operation of the function by hand and it is working as intended.


hbushouse:
There's a flake8 failure that needs fixing.

PR #5823, opened 2021-03-10
Add logging to srctype WFSS branch

hbushouse:
Trivial update to add logging info when setting SRCTYPE values for WFSS exposures, as is already done for MOS mode.


jemorrison (approved):
Looks good. I guess I forgot to add that information

PR #5828, opened 2021-03-10
JP-1955: Index copy of array for ami_analyze full frame slicing

### LINKED ISSUE #5827: ASDF update (2.7.3) prevents current slicing operation in ami_analyze
_Issue [JP-1955](https://jira.stsci.edu/browse/JP-1955) was created on JIRA by [Tyler Pauly](https://jira.stsci.edu/secure/ViewProfile.jspa?name=tpauly):_

For full-frame observations running through calwebb_ami3, ami_analyze includes indexing operations that are no longer supported by ASDF 2.7.3. Lines 57:59 now must include a 
```python
{...}.copy()
```
.

tapastro:
Resolves [JP-1955](https://jira.stsci.edu/browse/JP-1955)
Resolves #5827 

Working to get the relevant test data in a public directory so this can be checked by someone else, but the logic seems clear.

James suggested a unit test to cover this block, which I'll look into.


jwst/ami/ami_analyze.py
```diff
@@ -54,9 +54,9 @@ def apply_LG_plus(input_model, filter_model, oversample, rotation,
         ysize = 80
         xstop = xstart + xsize - 1
         ystop = ystart + ysize - 1
-        input_model.data = input_model.data[ystart-1:ystop, xstart-1:xstop]
-        input_model.dq = input_model.dq[ystart-1:ystop, xstart-1:xstop]
-        input_model.err = input_model.err[ystart-1:ystop, xstart-1:xstop]
+        input_model.data = input_model.data[ystart-1:ystop, xstart-1:xstop].copy()
+        input_model.dq = input_model.dq[ystart-1:ystop, xstart-1:xstop].copy()
+        input_model.err = input_model.err[ystart-1:ystop, xstart-1:xstop].copy()
```
jdavies-st, line 59: Looking at the whole function here, it's doing something insidious.  It says it is taking `input_model` as input and returning `output_model`, but it is actually modifying (permanently) `input_model` into a view in the process.  Even your fix here, changes input model to be a subarray of what it was on the way in.

Either way it is bad practice.

That said, I don't know if anything downstream depends on this bad practice.  It likely does, so fixing this may be complicated.

tapastro: Your fears are warranted! I had assumed the change in output array size was intended, as a sort of psf measurement per source? But from the documentation of ami_average:

> The ami_average step produces a single output file, having the same format as the input files, where the data for the 8 file components are the averages from the list of input files.

If the input is a full-frame image, the output of both ami_average and ami_normalize do not follow this statement, but are instead the size of a SUB80 frame.

hbushouse: The ami_average and ami_normalize steps use completely different datamodels for their inputs/outputs that are not at all image-like. They are lists of computed fringe parameters. ami_analyze uses a normal `ImageModel` as input, but the output is an `AmiLgModel`. So the input to ami_analyze is just thrown away when processing is done; it doesn't move on. Hence modifying it in place in this subarray extraction operation is likely to be harmless.

tapastro: So the documentation of the other steps refer not to the ami pipeline, but to the individual step. I suppose that makes sense.

jdavies-st: The bug was happening though because this newly-now-a-view subarray was being written out to a FITS file, and `asdf` doesn't like writing out views to FITS files anymore because the FITS file ends up being corrupted.

So maybe a question is, if this gets written out, should it be a subarray or the original size?

hbushouse: Why is it getting written out? Only the `output_model` should get written out by either the step or the calwebb_ami3 pipeline when the step returns `output_model` to it. There's no need to write `input_model` to a file when the step completes, because it's not the intended output ("these aren't the data you're looking for ...")

jdavies-st: Great question.  Do we have the SDP log that originally discovered this bug? Maybe it can be added to issue #5827 .

tapastro: Added it to the ticket comments.

jdavies-st: I'll reiterate though that at no point in `AmiAnalyzeStep` or the functions in this module is there a copy made of the input array, but it is modified.  Probably be good to make a copy and pass that along so that the original is not modified in-place.

hbushouse: I guess that makes sense for folks who might be calling `ami_analyze` interactively and don't want to have their input model get munged, cuz they might want to use it again for another call or something. So I agree that we should make a copy of `input_model` and then do the slicing on the copy and leave `input_model` alone. Just need to make sure that the new copy gets passed along properly to the rest of the function.

hbushouse: So I think this can be implemented by making a copy of `input_model` to some other instance that then gets fed into the rest of processing, but do not copy it to `output_model` (like other steps do). `output_model` is created from scratch as an `AmiLgModel` and populated with the appropriate output data by one of the lower-level functions.

jwst/ami/ami_analyze.py
```diff
@@ -41,6 +42,8 @@ def apply_LG_plus(input_model, filter_model, oversample, rotation,
         Fringe analysis data
 
     """
+    # Create copy of input_model to avoid overwriting input
+    input_copy = datamodels.ImageModel(input_model.copy())
```
jdavies-st, line 46: You don't need to copy it and then pass in as an `ImageModel`.  That's extra overhead that will then create a new shallow copy pointer to the new deep copy.  A simple deep copy will do I think:

```python
input_copy = input_model.copy()
```

And then the `datamodels` import is also not needed.

tapastro: Could I also remove the copy()s in the if block, as I'm now modifying the copy rather than the original?

hbushouse: yes, you can then just slice `input_copy`

jdavies-st: Yes!  Absolutely.  I missed those.

hbushouse (approved):
LGTM

PR #5837, opened 2021-03-12
Move non-runtime ramp_fitting code to tests

jdavies-st:
This PR moves the following:

```
 rename jwst/ramp_fitting/{ => tests}/compare_cr_files.py (100%)
 rename jwst/ramp_fitting/{ => tests}/compare_cr_navg_files.py (100%)
 rename jwst/ramp_fitting/{ => tests}/compare_crs.py (100%)
 rename jwst/ramp_fitting/{ => tests}/create_cube.py (100%)
 rename jwst/ramp_fitting/{ => tests}/mc_3d.py (100%)
```

These modules are essentially test helpers for creating and comparing test data, so should be in the test area.  They are not runtime code.

I dont' think we want to get rid of these, as they can probably be used in the future in automated unit tests.  Relocating them to the `tests` subdir will make it clear this is testing code, not runtime code, and they won't get counted as untested runtime code by Codecov.

This will increase the coverage of the `ramp_fitting` submodule as we've been currently counting these as runtime code.

Addresses #5701 / [JP-1887](https://jira.stsci.edu/browse/JP-1887)


kmacdonald-stsci (approved):

dmggh:
That's fine to move these tests for now. These old development scripts were written long before the ramp fitting code reached it's current state.

Issue #5826, opened 2021-03-10
ramp_fit tests run twice, don't test GLS


stscijgbot-jp:
_Issue [JP-1954](https://jira.stsci.edu/browse/JP-1954) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

Tests in

`jwst/ramp_fitting/tests/test_ramp_fit::TestMethods`

are parametrized by fit method, GLS or OLS, but only one of the tests actually passes the test method onto the `ramp_fit` function, which means that every test is run twice and falsely reports that it passes with method OLS and GLS:

```
$ pytest jwst/ramp_fitting/tests/test_ramp_fit.py::TestMethods -v
=================================================== test session starts ====================================================
platform darwin -- Python 3.9.1, pytest-6.2.2, py-1.10.0, pluggy-0.13.1 -- /Users/jdavies/miniconda3/envs/jwst/bin/python
cachedir: .pytest_cache
crds_context: jwst_0687.pmap
rootdir: /Users/jdavies/dev/jwst, configfile: setup.cfg
plugins: requests-mock-1.8.0, cov-2.11.1, ci-watson-0.5, xdist-2.2.1, asdf-2.7.2, openfiles-0.5.0, forked-1.3.0, doctestplus-0.9.0
collected 48 items                                                                                                         

jwst/ramp_fitting/tests/test_ramp_fit.py::TestMethods::test_nocrs_noflux[OLS] PASSED                                 [  2%]
jwst/ramp_fitting/tests/test_ramp_fit.py::TestMethods::test_nocrs_noflux[GLS] PASSED                                 [  4%]
jwst/ramp_fitting/tests/test_ramp_fit.py::TestMethods::test_nocrs_noflux_firstrows_are_nan[OLS] PASSED               [  6%]
jwst/ramp_fitting/tests/test_ramp_fit.py::TestMethods::test_nocrs_noflux_firstrows_are_nan[GLS] PASSED               [  8%]
jwst/ramp_fitting/tests/test_ramp_fit.py::TestMethods::test_error_when_frame_time_not_set[OLS] PASSED                [ 10%]
jwst/ramp_fitting/tests/test_ramp_fit.py::TestMethods::test_error_when_frame_time_not_set[GLS] PASSED                [ 12%]
[...]

PR #5878, opened 2021-03-16
JP-1954: Changed tests to parametrize the choice of ramp fitting algorithm

kmacdonald-stsci:
The tests set up a class to parametrize testing `ramp_fitting` using ordinary least squares (OLS) and generalized least squares (GLS) as algorithms to compute slopes and variances for ramps.  However, the tests had OLS hardcoded as the algorithm to use, so the same test twice, instead of testing GLS.

Addresses #5826 
Fixes [JP-1954](https://jira.stsci.edu/browse/JP-1954)


hbushouse (approved):
Just curious to know if you've run the new tests locally and if so, how much longer the GLS version takes to run compared to OLS?  Although if the input data arrays are small, there might not be much noticeable difference. What really matters is how they compare on normal-sized exposures.

melanieclarke:
@kmacdonald-stsci - this PR looks like it was overtaken by the work on OLS and likelihood algorithms for ramp fitting.  Is it okay to close?

PR #5891, opened 2021-03-18
Update and fix NIRSpec MOS master bkg regtests

hbushouse:
**Description**

This PR addresses a problem I recently noticed in the master bkg regression tests for the NIRSpec MOS mode. First, the existing test that used a user-supplied master bkg spectrum was still running the calspec3 master bkg step, instead of using the new NIRSpec-slit-specific version of the master bkg step that gets called from calspec2. I removed that entire test and replaced it with one that runs the data through calspec2, along with a user-supplied mbkg spectrum. I also fixed the naming and location of truth files for the equivalent test that runs MOS data through calspec2 and computes the mbkg spectrum from available bkg slits in the exposure.

All necessary input files and truth files for the new test have been uploaded to artifactory and the location of truth files for the non-user-supplied case has been moved.

Related to [JP-1935](https://jira.stsci.edu/browse/JP-1935)

Checklist
- [X] Tests
- [ ] Documentation  (N/A: no docs for regtests)
- [ ] Change log (N/A: not required for regtests)
- [X] Milestone
- [X] Label(s)

### DIFF
--- a/jwst/regtest/test_nirspec_masterbackground.py
+++ b/jwst/regtest/test_nirspec_masterbackground.py
@@ -35,6 +35,29 @@ def run_spec2_mbkg(jail, rtdata_module):
     return rtdata
 
 
+@pytest.fixture(scope='module')
+def run_spec2_mbkg_user(jail, rtdata_module):
+    """Run Spec2 on MSA data with a user-supplied master bkg spectrum"""
+    rtdata = rtdata_module
+
+    # Get data
+    rtdata.get_data('nirspec/mos/nrs_mos_3pointnod_1_msa.fits')
+    rtdata.get_data('nirspec/mos/jw00626030001_02103_00001_nrs1_masterbg1d.fits')
+    rtdata.get_data('nirspec/mos/jw00626030001_02103_00001_nrs1_rate.fits')
+    collect_pipeline_cfgs('config')
+
+    # Run the pipeline
+    step_params = {
+        'step': 'calwebb_spec2.cfg',
+        'args': [
+            '--steps.master_background.skip=false',
+            '--steps.master_background.user_background=jw00626030001_02103_00001_nrs1_masterbg1d.fits'
+        ]
+    }
+    rtdata = rt.run_step_from_dict(rtdata, **step_params)
+    return rtdata
+
+
 def test_masterbkg_rerun(rtdata):
     """Test to ensure sequential runs of the step are consistent"""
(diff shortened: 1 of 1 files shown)


tapastro (approved):
Looks good to me, but a question: it looks as though the default level 3 master background step has a .cfg file you could use with strun for non-nirspec-slit observations, but it looks like here you use the calwebb_spec2 input to pull the nrs_slits-specific step. Is this the only easy way to interact with this step? I'm still getting my head around strun calls without cfg input, so maybe there's another way.

hbushouse:
> Looks good to me, but a question: it looks as though the default level 3 master background step has a .cfg file you could use with strun for non-nirspec-slit observations, but it looks like here you use the calwebb_spec2 input to pull the nrs_slits-specific step. Is this the only easy way to interact with this step? I'm still getting my head around strun calls without cfg input, so maybe there's another way.

Theoretically I believe you could test only the master bkg subtraction by just calling the nirspec-slits version of the master bkg step, but in order for it to work right (i.e. give the right scientific answer) it has to be called mid-stream within the calwebb_spec2 processing flow (e.g. after the exposure has been processed up through the extract_2d and srctype steps) and then the resulting output of the mbkg step gets reinjected into the remaining flow of calwebb_spec2. So in order to just call the nirspec-slits version of mbkg by itself, the input data would need to have that partial calwebb_spec2 processing applied to them. I figured it was easy and more thorough to just run the entire calwebb_spec2 pipeline on a given exposure, let it do all the complicated data flow handling, and then test the final calwebb_spec2 output products (rather than just testing the result of mbkg by itself). That way we also make sure that changes in calwebb_spec2 itself don't have any unintended side effects on the mbkg process.

tapastro:
> Theoretically I believe you could test only the master bkg subtraction by just calling the nirspec-slits version of the master bkg step, but in order for it to work right (i.e. give the right scientific answer) it has to be called mid-stream within the calwebb_spec2 processing flow (e.g. after the exposure has been processed up through the extract_2d and srctype steps) and then the resulting output of the mbkg step gets reinjected into the remaining flow of calwebb_spec2. So in order to just call the nirspec-slits version of mbkg by itself, the input data would need to have that partial calwebb_spec2 processing applied to them. I figured it was easy and more thorough to just run the entire calwebb_spec2 pipeline on a given exposure, let it do all the complicated data flow handling, and then test the final calwebb_spec2 output products (rather than just testing the result of mbkg by itself). That way we also make sure that changes in calwebb_spec2 itself don't have any unintended side effects on the mbkg process.

That makes sense to me. PR review and a lesson, what a deal!

PR #5892, opened 2021-03-18
Fixed README so that some commands are `zsh` compatible.

WilliamJamieson:
**Description**
There were some commands listed in the `README.md` which caused errors when running in the `zsh` shell. To fix this I added escapes around these commands which allow them to be run in the `zsh` shell properly.

Checklist
- [ ] Tests

- [x] Documentation

- [ ] Change log

- [x] Milestone

- [x] Label(s)


jdavies-st (approved):
Looks good.  Tried these out, and it works on both bash and zsh.  👍 

Thanks!

PR #5898, opened 2021-03-19
enable stcal

nden:
**Description**

This is the first PR which enables a new library for code common to
jwst and romancal. Sofar the only code moved to stcal is the code
manipulating bitmasks. It is used by `MaskModel`, `outlier_detection` and `resample`.

Checklist
- [ ] Tests

- [ ] Documentation

- [x] Change log

- [x] Milestone

- [x] Label(s)


jwst/datamodels/dqflags.py
```diff
@@ -16,9 +16,11 @@
 which provides 32 bits. Bits of an integer are most easily referred to using
 the formula `2**bit_number` where `bit_number` is the 0-index bit of interest.
 """
-from astropy.nddata.bitmask import interpret_bit_flags as ap_interpret_bit_flags
 
-from jwst.lib.basic_utils import multiple_replace
+# These imports are here for backwards compatibility
+from astropy.nddata.bitmask import interpret_bit_flags as ap_interpret_bit_flags # NOQA
```
hbushouse, line 21: What does "NOQA" stand for in these lines?

hbushouse (approved):
Looks OK to me, other than the one minor question I have.

jwst/datamodels/dqflags.py
nden: Flake8 complains that these are imported but not used. 
A comment `#NOQA` signals flake8 to exclude this line from the check.

hbushouse: Ah, I didn't know that little trick. And it also doesn't complain about the fact that there aren't 2 spaces before the "#"?

jdavies-st: "No Quality Assurance" I believe.  👍

jwst/datamodels/mask.py
```diff
@@ -22,7 +23,7 @@ def __init__(self, init=None, **kwargs):
         super(MaskModel, self).__init__(init=init, **kwargs)
 
         if self.dq is not None or self.dq_def is not None:
-            self.dq = dynamic_mask(self)
+            self.dq = dynamic_mask(self, pixel)
```
jdavies-st, line 26: The old  calling sequence of `dynamic_mask()` did not require a specification of which mnemonic map to use, i.e. `pixel` or `group`.  I see the version in `stcal` now does.  This is a breaking change.  I don't see this used outside of `jwst`, `romancal` or `stcal`, so perhaps this is OK?  Or would it be good to have default specified in the version now in `stcal`?

Also, the `jwst/datamodels/dynamicdq.py` module still exists in this package.  We should probably remove it in this PR, or just keep it as if we are now using the version in `stcal`.  It is used in a bunch of our datamodels:

https://github.com/spacetelescope/jwst/search?q=dynamicdq

and of course `romcal` too.

jdavies-st (commented):
Looks good, though I think the `dynamicdq.py` module needs to be removed in this package.  It doesn't seem to be used anywhere on Github besides our internal packages, even though it is technically part of the public interface.

jwst/datamodels/mask.py
nden: I think placing a default `pixel` in stcal may lead to problems, especially if the function is called with the old signature which uses the default `pixel` definitions but the jwst ones have changed. I'd rather have code like that fail and be corrected.

Good point for the `dynamicdq` module in jwst! I missed that. Update coming.

jdavies-st: Agree!  👍

jdavies-st (approved):
Looks good!

Not sure why the coverage is reported to have dropped, but it seems to be a problem with Codecov, not this PR.

PR #5901, opened 2021-03-19
JP-2004: Update documentation deprecate primary use of CFG files

### LINKED ISSUE #2921: Add call method example that includes the "steps=" argument
It might be helpful for pipeline users if there was an example in the documentation that shows how to run the pipeline from within Python without a config file, using the "steps" argument instead, e.g.:

`Image3Pipeline.call("asn.json", steps={"tweakreg":{"skip":True}, "skymatch":{"skip":True}, "source_catalog":{"output_file":"test_cat.ecsv"}})`

This issue came up from a discussion on Slack about how to get the Image3 pipeline to save out the catalog using this method.

### LINKED ISSUE #5890: Review/update package documentation for the New Config World Order
_Issue [JP-2004](https://jira.stsci.edu/browse/JP-2004) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Update documentation on usage of pipelines and steps with regards to defaults and default retrieval of step parameters from CRDS. Changes include

* Verify how to retrieve/create ASDF configs
* Remove normal usages of CFG files and use class specification of pipelines
* Create a deprecation page to document relation between new and old usage

stscieisenhamer:
Resolves [JP-2004](https://jira.stsci.edu/browse/JP-2004)
Resolves [JP-2042](https://jira.stsci.edu/browse/JP-2042)
Fixes #5890
Fixes #2921 

**Description**

This PR updates the documentation to deprecate the use of `cfg` files as the primary identifier of and use for running the pipelines and steps.

Checklist
- [ ] Tests

- [x] Documentation

- [x] Change log

- [x] Milestone

- [x] Label(s)


docs/jwst/pipeline/main.rst
```diff
@@ -229,15 +204,94 @@ Control (WFS&C) use a dedicated processing flow through the pipeline stages.
...
  - Stage 3: The ASN generator identifies pairs of dithered WFS&C images to be
    combined via the "PATTTYPE" keyword value "WFSC". The resulting ASN
    uses the :ref:`calwebb_wfs-image3 <calwebb_wfs-image3>` pipeline
    for stage 3 processing. This pipeline consists of the single step
    :ref:`wfs_combine <wfs_combine_step>`.
+
+Configuration File Deprecation
+==============================
+
+Up to version 1.1.1, the primary way specific pipelines were referred to was by
```
hbushouse, line 222: We're only at 1.1.0 right now, so did you mean that version? I doubt we'll ever have a 1.1.1, so the next potential release would be 1.2.0 (probably an interim release before B7.8) and then B7.8 would be higher yet.

docs/jwst/pipeline/main.rst
```diff
@@ -229,15 +204,94 @@ Control (WFS&C) use a dedicated processing flow through the pipeline stages.
...
+their configuration file name, i.e. ``calwebb_detector1.cfg``. These
+configuration files were delivered as part of the JWST calibration package.
+Below is the table that matched configuration file to observing mode it was
+intended to be used with.
+
+Post-1.1.1, configuration files are no longer the primary identifier of
```
hbushouse, line 228: Post-1.1.0

docs/jwst/pipeline/main.rst
```diff
@@ -229,15 +204,94 @@ Control (WFS&C) use a dedicated processing flow through the pipeline stages.
...
+``jwst.pipeline.Detector1Pipeline``, or by their simple name, or alias, i.e.
+``calwebb_detector1``. How a pipeline is run is determined by the input data and
+what step parameter reference file in CRDS is selected by that data. The reftype
+for each pipeline, or step, is determined by appending the class name of the
+step to the string ``pars-``. For example, the reftype for
+``jwst.pipeline.Detector1Pipeline`` is ``pars-detector1pipeline``. Which
+specific reference file for a reftype is then determined by the data, just as
+with any other reference file.
+
+As a result, there are a few pipelines that no longer exist explicitly by name,
```
hbushouse, line 239: "there are a few pipeline configurations that no longer exist" ?  We haven't gotten rid of any actual pipelines.

docs/jwst/pipeline/main.rst
```diff
@@ -229,15 +204,94 @@ Control (WFS&C) use a dedicated processing flow through the pipeline stages.
...
+since they were only a configuration file for an already existing pipeline
```
hbushouse, line 240: Sorry, word snob: I like to use "since" when referring to time (e.g. "since we last met") but use "because" for a causal relationship like this one.

docs/jwst/pipeline/main.rst
```diff
@@ -229,15 +204,94 @@ Control (WFS&C) use a dedicated processing flow through the pipeline stages.
...
+| calwebb_tso1.cfg          | `jwst.pipeline.Detector1Pipeline` | calwebb_detector1 |
++---------------------------+-----------------------------------+-------------------+
+| calwebb_tso-image2.cfg    | `jwst.pipeline.Image2Pipeline`    | calwebb_image2    |
++---------------------------+-----------------------------------+-------------------+
+| calwebb_tso-spec2.cfg     | `jwst.pipeline.Spec2Pipeline`     | calwebb_spec2     |
++---------------------------+-----------------------------------+-------------------+
+| calwebb_wfs-image2.cfg    | `jwst.pipeline.Image2Pipeline`    | calwebb_image2    |
++---------------------------+-----------------------------------+-------------------+
+
+The deprecated, up to version 1.1.1, configuration to mode mapping. This table is given only
```
hbushouse, line 260: Perhaps move the "up to version ..." to the end of the sentence, so that the thought doesn't get broken up.

docs/jwst/pipeline/calwebb_detector1.rst
```diff
@@ -15,54 +15,54 @@ are in the form of one or more ramps (integrations) containing accumulating
...
-There are two unique configuration files to be used to control this pipeline,
-depending on whether the data are to be treated as a Time Series Observation (TSO).
-Non-TSO exposures use the ``calwebb_detector1`` configuration, which applies all
-applicable steps to the data. The ``calwebb_tso1`` configuration, on the other hand,
-should be used for TSO exposures, for which some steps are set to be skipped by
-default (see the list of steps in the table below). Both configurations call the
-``Detector1Pipeline``; the only difference is which steps are applied.
+There are two unique configurations used to control this pipeline,
+depending on whether the data are to be treated as a Time Series Observation
+(TSO). The configuration is controlled by CRDS and the reftype
```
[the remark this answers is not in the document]
stscieisenhamer: Agreed with the wording change.

Question though, and this applies to a number of other cases in the documentation: Information such as this is beginning to cross the line of "what is and how does one use the software" to "use-case scenarios and science-oriented discussion of how the software is utilized". Granted it is a fuzzy line, and this may be the only place such information is currently discussed, but such information feels like it should be in some sort of user-handbook about the specific instruments and observation modes and their calibration. Oh wait, such a thing exists: JDocs! 

With the move to a more data-driven process, this type of documentation really doesn't belong here. Of course, this is a much bigger re-org that will not be solved by this PR, but should be considered in future revisions.

docs/jwst/pipeline/main.rst
stscieisenhamer: No, no actual pipelines have been removed; they never existed. For example, the "pipeline" referred to as "calwebb_wfs-image2.cfg" was never a pipeline. It is just a specific configuration for Image2Pipeline. There was never a class `jwst.pipeline.WfsSpec2Pipeline` or any other code defining a "calwebb_wfs-image2" pipeline.

As such, as we move away from using config files and depending on CRDS, there is no `strun calwebb_wfs-image2" command. It will simply be `strun calwebb_image2` and the data and CRDS selection will choose the right configuration.

At the moment, the configuration file `calwebb_wfs-image2.cfg` is still in the package, and will appear when `collect_pipeline_cfgs` is run. However, the file, as all other CFG files, are now devoid of any actual settings; it is all CRDS-dependent.

docs/jwst/pipeline/calwebb_detector1.rst
```diff
@@ -15,54 +15,54 @@ are in the form of one or more ramps (integrations) containing accumulating
...
+``pars-detector1pipeline``. In general, for Non-TSO exposures, all applicable
+steps are applied to the data. For TSO exposures, some steps are set to be
+skipped by default (see the list of steps in the table below).
 
-The list of steps applied by the ``Detector1Pipeline`` pipeline is shown in the
-table below. Note that MIRI exposures use some instrument-specific steps and some of
-the steps are applied in a different order than for Near-IR (NIR) instrument exposures.
+As of CRDS context ``jwst_0704``, the list of steps applied by the
```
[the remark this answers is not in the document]
stscieisenhamer: Agreed, as this is another case that may more appropriately belong in other documentation elsewhere.

However, seeing that this is a major shift in paradigm, the general idea that "at version XXX and crds context 704, a major change in functionality occurs...". Will see where such a statement can be place.

docs/jwst/pipeline/calwebb_dark.rst
```diff
@@ -32,8 +32,8 @@ pipeline, the order of steps is a bit different for MIRI exposures.
 |                                       | :ref:`rscd <rscd_step>`                 |
 +---------------------------------------+-----------------------------------------+
 
-.. [1] The :ref:`ipc <ipc_step>` step is currently set to be skipped by default in the
-   "calwebb_dark.cfg" configuration file for all instruments.
+.. [1] By default, the step parameter reference `pars-darkpipeline`
```
hbushouse, line 35: Could probably remove the word "step" here, because in this particular example `pars-darkpipeline` is a param file for a pipeline, not a step. I know that we developers know that a pipeline inherits from `Step` and hence in terms of python objects they're kind of the same thing, but your average Joe user isn't going to or want to know that.

docs/jwst/pipeline/calwebb_detector1.rst
```diff
@@ -15,54 +15,54 @@ are in the form of one or more ramps (integrations) containing accumulating
...
+|                                            |         |         | :ref:`refpix <refpix_step>`             | |check| | |check| |
++--------------------------------------------+---------+---------+-----------------------------------------+---------+---------+
+| :ref:`jump <jump_step>`                    | |check| | |check| | :ref:`jump <jump_step>`                 | |check| | |check| |
++--------------------------------------------+---------+---------+-----------------------------------------+---------+---------+
+| :ref:`ramp_fitting <ramp_fitting_step>`    | |check| | |check| | :ref:`ramp_fitting <ramp_fitting_step>` | |check| | |check| |
++--------------------------------------------+---------+---------+-----------------------------------------+---------+---------+
+| :ref:`gain_scale <gain_scale_step>`        | |check| | |check| | :ref:`gain_scale <gain_scale_step>`     | |check| | |check| |
++--------------------------------------------+---------+---------+-----------------------------------------+---------+---------+
+
+.. [1] By default, the step parameter reference `pars-detector1pipeline`
```
hbushouse, line 64: same comment here about the use of "step"

docs/jwst/pipeline/calwebb_wfs-image2.rst
```diff
@@ -3,8 +3,15 @@
 calwebb_wfs-image2: Stage 2 WFS&C Processing
 ============================================
 
+:Deprecated post-1.1.1:
```
hbushouse, line 6: Should be post-1.1.0

docs/jwst/pipeline/main.rst
```diff
@@ -39,60 +39,35 @@ observing modes, including:
 
 - :ref:`calwebb_dark <calwebb_dark>` for processing dark exposures
 - :ref:`calwebb_guider <calwebb_guider>` for calibrating FGS guide star data
-- :ref:`calwebb_wfs-image2 <calwebb_wfs-image2>` for stage 2 WFS&C processing
```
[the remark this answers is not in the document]
hbushouse: Oh, OK, I get it now. There's no longer any need for `calwebb_wfs-image2`, because the step skipping is handled by the `pars-image2pipeline` ref file, in the same way there's no need for `calwebb_tso1` or `calwebb_tso-spec2` anymore. Never mind.

docs/jwst/stpipe/config_asdf.rst
```diff
@@ -3,7 +3,11 @@
 ASDF Configuration Files
```
hbushouse, line 3: Not sure we should continue to call them "configuration" files, because of the historical connection between that term and an actual .cfg file. I know we're using the word "configuration" in a generic sense here, but to avoid possible confusion, we should perhaps always now refer to them as "parameter" files.

docs/jwst/introduction.rst
```diff
@@ -78,12 +78,21 @@ desired step or pipeline (see `Configuration Files`_ below for more details).
...
 
 For example, running the full stage 1 pipeline or an individual step by
-referencing their class names is done as follows:
+referencing their class name is done as follows:
 ::
 
   $ strun jwst.pipeline.Detector1Pipeline jw00017001001_01101_00001_nrca1_uncal.fits
   $ strun jwst.dq_init.DQInitStep jw00017001001_01101_00001_nrca1_uncal.fits
 
+Most pipelines also have an "alias", or short name, that can be used instead of the
```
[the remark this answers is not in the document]
stscieisenhamer: Checked, yes all the pipelines do. It is the steps that do not. Will fix the wording.

docs/jwst/introduction.rst
```diff
@@ -371,19 +373,19 @@ Another argument available to all steps in a pipeline is ``skip``.
 If ``skip=True`` is set for any step, that step will be skipped, with the
 output of the previous step being automatically passed directly to the input
 of the step following the one that was skipped. For example, if you want to
-skip the linearity correction step, edit the calwebb_detector1.cfg file to
+skip the linearity correction step, edit the calwebb_detector1.asdf file to
```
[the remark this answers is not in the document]
hbushouse: This still needs to be cleared up too.

docs/jwst/introduction.rst
```diff
@@ -460,11 +462,10 @@ configuration file add ``--save-parameters <filename.asdf>`` to the command:
 
 $ strun <step.class> <required-input-files> --save-parameters <filename.asdf>
 
-For example, to save the parameters used for a run of the ``calwebb_image2.cfg`` pipeline, use:
+For example, to save the parameters used for a run of the ``calwebb_image2`` pipeline, use:
 ::
 
-$ collect_pipeline_cfgs .
-$ strun calwebb_image2.cfg jw82500001003_02101_00001_NRCALONG_rate.fits --save-parameters my_image2.asdf
+$ strun calwebb_image2 jw82500001003_02101_00001_NRCALONG_rate.fits --save-parameters my_image2.asdf
```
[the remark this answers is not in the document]
hbushouse: ?

hbushouse (commented):
Don't forget a change log entry.

docs/jwst/stpipe/config_asdf.rst
```diff
@@ -27,43 +33,51 @@ The remaining sections will describe the file format and contents.
...
+     coord_system: skyalign
+     filter: all
+     grating: all
+     input_dir: ''
+     output_ext: .fits
+     output_type: band
+     output_use_index: true
+     output_use_model: true
+     post_hooks: []
+     pre_hooks: []
```
jdavies-st, line 67: None of these general Step parameters applicable to all steps should ever be included in a parameter reference file.  And in fact most of a step-specific parameters shouldn't either.  Generally, the rule should be that if you are selecting on an instrument with CRDS selectors, only those parameters that change from the code default based on instrument and CRDS selectors should be included.

docs/jwst/stpipe/config_asdf.rst
```diff
@@ -75,65 +89,61 @@ The first 5 lines, up to and including the "---" line, define the file as an
...
+++++++++++
 
-All other keys are the parameters and their values to be used when the
-step/pipeline is run. The order of the parameters does not matter. Except for
-the ``class`` key, no other parameter needs to be specified. If not defined, the
-default, as defined in the code, will be used.
+``parameters`` contains all the parameters to pass onto the step. The order of
+the parameters does not matter. It is not necessary to specify all parameters
+either. If not defined, the default, as defined in the code or values from CRDS
+parameter references, will be used.
```
jdavies-st, line 120: As mentioned above, I think it's important to note that for a `pars-foo` reffile for an instrument with CRDS selectors, only those parameters that change from the code default based on instrument and CRDS selectors should be included.

This is important to have here, as already many of the `pars-foo` reffiles delivered do not do this and blindly list all the parameters, regardless of whether they are different based on the instrument or the CRDS selectors for that instrument pars file.  This is bad, since it means that the defaults in the code will never get applied, even if they get changed in the code during commissioning.  In order to get the new defaults (say for a global parameter that is not instrument-specific) the teams will have to deliver new `pars-foo` reffiles.

I see this is covered a bit below in "Completeness", but I don't think we can repeat this enough.

docs/jwst/stpipe/config_asdf.rst
```diff
@@ -27,43 +33,51 @@ The remaining sections will describe the file format and contents.
...
+   --- !core/asdf-1.1.0
+   asdf_library: !core/software-1.0.0 {author: Space Telescope Science Institute, homepage: 'http://github.com/spacetelescope/asdf',
+     name: asdf, version: 2.7.3}
+   history:
+     extensions:
+     - !core/extension_metadata-1.0.0
+       extension_class: asdf.extension.BuiltinExtension
+       software: !core/software-1.0.0 {name: asdf, version: 2.7.3}
+   class: jwst.cube_build.cube_build_step.CubeBuildStep
+   name: CubeBuildStep
```
jdavies-st, line 54: Good to see that `class` and `name` have been pulled out of the parameters section.

docs/jwst/stpipe/config_cfg.rst
```diff
@@ -3,6 +3,15 @@
...
 =========================
 
+.. note::
+
+   The ``cfg`` format can still be used but is deprecated in favor of
+   :ref:`config_asdf_files`. Please convert any processes that use ``cfg`` files
+   to the ``ASDF`` format. Note also that all ``cfg`` files that are currently
+   being delivered in the package and retrieved using ``collect_pipeline_cfgs``
+   set no parameters; files are empty. All steps query CRDS parameter references
+   for any data-dependent parameter settings.
```
jdavies-st, line 13: ", or use the code defaults."

docs/jwst/srctype/description.rst
```diff
@@ -92,19 +92,16 @@ background.
...
-used in the APT, nor a single keyword in the science product, to record the type of
-each source. For these exposures, a stellarity value can be supplied by
-the observer for each source used in the MSA Planning Tool (MPT). The
-stellarity values are
-in turn passed from the MPT to the MSA metadata (_msa.fits) file
-created by DMS and used in the calibration pipeline. The stellarity
-values from the MSA metadata file are loaded for each source/slitlet
-by the ``assign_wcs`` step of the ``calwebb_spec2`` pipeline and then
-evaluated by the ``srctype`` step to determine whether each source
+NIRSpec MOS +++++++++++ For NIRSpec MOS exposures (EXP_TYPE="NRS_MSASPEC"),
```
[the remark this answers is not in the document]
stscieisenhamer: About the comment on calwebb_wfs-image2 and no alias: There is no class to define an alias for. The class `Image2Pipline` already has its own alias, `calwebb_image2`. Since the alias definition is not a list, no more aliases can be defined.

If it is felt that there really has to be an alias, a placeholder class could be created that is simply a child of Image2Pipeline. 

Otherwise, the deprecation notice could be stronger, possibly removing all the information on this page except for the deprecation.

docs/jwst/stpipe/config_asdf.rst
stscieisenhamer: Thanks for pointing out that there are already references submitted. Appropriate messaging to INS will be going out shortly.

hbushouse (approved):

hbushouse:
Doc build has failed due to 2 errors in .rst files that need fixing.

PR #5902, opened 2021-03-19
Fix test that fails with asdf master

eslavich:
**Description**

This test has been failing when run with asdf master due to the change in behavior around serializing `None` values.  This PR will allow the test to pass under both asdf 2.7.x and the upcoming asdf 2.8.

Checklist
- [x] Tests

- [ ] Documentation

- [ ] Change log

- [ ] Milestone

- [ ] Label(s)


jdavies-st (approved):
LGTM.

PR #5913, opened 2021-03-26
JP-1960: Updates for background subtraction docs

### LINKED ISSUE #5835: Enhance documentation for spectroscopic background subtractions
_Issue [JP-1960](https://jira.stsci.edu/browse/JP-1960) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

Given the various different ways of doing background subtraction for different spectroscopic modes, users are often confused as to which method is used for a particular instrument mode and/or observation setup, and in what pipeline (calspec2 or calspec3) the subtraction actually occurs. We need to improve our documentation to try to explain this more clearly.

hbushouse:
Closes #5835 

Resolves [JP-1960](https://jira.stsci.edu/browse/JP-1960)

**Description**

This PR addresses shortcomings in the existing docs for background subtraction, particularly for spectroscopic modes and the difficult NIRSpec MOS master background subtraction. I've added a generic doc on the entire topic of background subtraction, which gives a somewhat high-level overview of all the different background subtraction methods that are available for both imaging and spectroscopy and where they live in the pipeline. The docs for individual steps still give all of the details of each method.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [X] Milestone
- [X] Label(s)


stscieisenhamer (approved):

docs/jwst/background_subtraction/main.rst
```diff
@@ -0,0 +1,183 @@
...
+------------
+Background subtraction for imaging data is currently only available in two places
+within the calibration pipeline:
+
+1. Image-from-image subtraction during :ref:`calwebb_image2 <calwebb_image2>`
+   processing in the :ref:`background <background_step>` step. The background images
+   come from observations of a dedicated background target.
+2. Local background subtraction for individual sources in the
+   :ref:`calwebb_image3 <calwebb_image3>` pipeline
+   :ref:`source_catalog <source_catalog_step>` step
```
jdavies-st, line 21: Currently in resampling, there's implicit background subtraction happening when we produce the `i2d` files.  I.e. the values computed in `skymatch` get subtracted before drizzling happens, which applies the `skymatch` results, but don't then go back and add back in the initial pedestal.  Is that a bug or feature?

I'm of the opinion that we should probably be adding back in the pedestal value so that the resampled images have the same mean background level as the reference input image.

docs/jwst/background_subtraction/main.rst
```diff
@@ -0,0 +1,183 @@
...
+observations of a designated background target have been obtained. For spectroscopic
+modes this is possible either through observations of a designated background target
+or when nodded exposures of a point-like target are obtained (e.g. using the MIRI LRS
+"ALONG-SLIT-NOD" dither pattern for an LRS fixed slit observation). Exposures from
+one nod position can be used as background for exposures at the other nod position,
+assuming the source is point-like.
+
+In either instance, the exposures to be used as background are included in the
+`image2` or `spec2` ASN file used to process the science target exposures, where
+the background exposures are labeled with an ASN member type of "background".
```
jdavies-st, line 125: > `image2` or `spec2` association (ASN) file used...

> with an ASN member `exptype` of "background".

or has "exptype" now permanently changed to "type"

jdavies-st (approved):
Looks good.  Just some minor comments/questions below.

docs/jwst/background_subtraction/main.rst
hbushouse: No, the attribute name in the ASN file is still "exptype" (unfortunately). I just referred to it in the text of the docs as a member type, because I think that wording is less confusing than saying "exposure type", because that can be interpreted as "EXP_TYPE".

docs/jwst/background_subtraction/main.rst
hbushouse: Hmmm, forgot about the background subtraction that can be done in `skymatch` and/or `resampling`. I think it's probably *not* desirable to have `resample` just go ahead and subtract background without it being a controllable option. So I guess I should mention these steps too when I list the possible ways of doing imaging background subtraction.

hbushouse: I guess we should perhaps first consult the CalWG to make sure they agree that the `resample` step should not be doing it blindly and automatically. At the very least it should be a user-settable param, just like it is in `skymatch`.

jdavies-st: Precisely why I'm thinking about it right now.

Of course we don't propagate errors for any background subtraction right now, so it's all "unaccounted" at the moment.

hbushouse: OK, could you file a JP ticket asking whether we should turn off background subtraction and assign it to Anton?

PR #5915, opened 2021-03-26
JP-1885 Changes for when there are only three and four good groups for jump detection

### LINKED ISSUE #5688: Enhance jump detection algorithm to work with fewer groups
_Issue [JP-1885](https://jira.stsci.edu/browse/JP-1885) was created on JIRA by [Anton Koekemoer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=koekemoe):_

This proposed enhancement to the baseline algorithm was received from Mike Regan who has discussed it also with Karl Gordon, who concurs with this.

Currently, the baseline jump detection algorithm only works on datasets that have at least 5 groups (ie, ngroup=5 or larger). However, it is expected that a significant amount of data will be obtained with fewer groups. Therefore, an enhancement is proposed to the baseline algorithm that will enable it to work with datasets that have fewer groups, with the following changes proposed:

For integrations with ngroup=4, do not exclude the highest difference before taking the median.

For integrations with ngroup=3, exclude the larger of the two differences, if it is above some threshold (probably different than the normal threshold).

-For multiple integration exposures, process all the differences in the exposure at once. This allows jumps to be found in all integrations with at least 2 groups. This would really help TSO observations.-   **_This portion was moved to JP-2517 to be tracked separately._**

mwregan2:
Closes #5688 

Resolves [JP-1885](https://jira.stsci.edu/browse/JP-1885) (partially)

**Description**

This PR addresses the first half of JP-1885. It adds the capability to detect jumps in three and four frame integrations. The baseline version only operates on integrations with at least five frame.

Checklist
- [X] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


dmggh (commented):
This latest set of updates seem fine. I do have a few mostly minor editorial comments about two_point_difference.py :

1) How are pixels with only a single good group handled within find_crs( ) ? If such pixels are filtered out before find_crs() is called perhaps clarifying that in the docs would help.

2) Some variables have long names, such as 'three_diff_rejection_threshold' (30 characters ). While very descriptive, readability of statements having several such variables could be improved, with shorter names and docstrings for example.

3) 'ndiffs- number_CRs_found - pixel_sat_groups'  is used 4 times within a while loop; perhaps use a single variable instead to ease readability.

4) In : 
 def get_clipped_median_array(num_differences, diffs_to_ignore, differences, sorted_index)
" This routine will return the clipped median for the input array."
What is the 'input array' ? I can't tell, based on the names of the arguments. Documentation (docstrings) of the input arguments (including types) would help clarify this.

hbushouse (commented):
Overall code structure looks good. Needs a change log entry, as well as modifications to get flake8 CI test to pass.

jwst/jump/tests/test_twopoint_difference.py
```diff
@@ -379,7 +403,8 @@ def test_11grps_0cr_3donotuse(setup_cube):
     gdq[0, 0, 100, 100] = dqflags.group['DO_NOT_USE']
     gdq[0, 9, 100, 100] = dqflags.group['DO_NOT_USE']
     gdq[0, 10, 100, 100] = dqflags.group['DO_NOT_USE']
-    out_gdq, row_below_gdq, row_above_gdq = find_crs(data, gdq, read_noise, rej_threshold, rej_threshold, rej_threshold, nframes, False, 200, 10)
+    out_gdq, row_below_gdq, row_above_gdq = find_crs(data, gdq, read_noise, rej_threshold, rej_threshold, rej_threshold,
+                                                     nframes, False, 200, 10)
     assert np.array_equal([0, 0, 0, 0, 0, 0, 0, 0], out_gdq[0, 1:-2, 100, 100])
```
dmggh, line 410: Why don't the next 2 steps have asserts (and therefore are set to be skipped) ?

mwregan2:
I think I've addressed all of David and Howard's comments. I've also updated read-the-docs and the inline comments to match the current version. I've fixed all the PEP8 problems and fixed the bugs I introduced fixing the PEP8 problems too.

hbushouse (commented):
Just a couple minor things and it'll be good to go.

hbushouse (approved):
Looks good now.

PR #5926, opened 2021-04-01
JP-1929: Added code to do refpix correction for NIR subarrays using 4 readout amplifiers

### LINKED ISSUE #5886: Update refpix to handle 4-amp subarrays
_Issue [JP-2001](https://jira.stsci.edu/browse/JP-2001) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

Update the "refpix" step to handle 4-amp subarrays, using the same logic that's applied to 4-amp full-frame readouts. Requirements are documented in JP-1929.

stscirij:
Closes #5886
Resolves [JP-1929](https://jira.stsci.edu/browse/JP-1929)

**Description**

This PR addresses JP-1929.  Some NIR subarrays will be taken using 4 readout amplifiers, but the current implementation of the refpix correction for subarrays assumes 1 readout amplifier.  Modify the code to use all 4 output amplifiers independently if NOUTPUTS=4, and also use side reference pixels if available.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


docs/jwst/refpix/description.rst
```diff
@@ -107,6 +107,9 @@ and/or rotated to convert to the detector frame.
 NIR Data
 ++++++++
 
+For single amplifier readout (NOUTPUTS keyword = 1):
+----------------------------------------------------
```
hbushouse, line 111: Might need a different section indicator for these. These should be a sub-section under "NIR Data", but it's using the same rst section markup as the higher level section "Subarrays". Perhaps "~~~~~~~" will work?

jwst/refpix/reference_pixels.py
```diff
@@ -1758,6 +1842,7 @@ def reference_pixel_correction(input_dataset):
         Corrected dataset
 
     """
+
     input_dataset.do_corrections()
 
-    return input_dataset
+    return
```
hbushouse, line 1848: So, why is the dataset no longer being returned to the calling step?

hbushouse (commented):
Looks good overall. Just a couple minor comments/questions.

Needs a change log entry.

hbushouse:
We should add a subarray with NOUTPUTS=4 case to the regtest suite.

jwst/refpix/reference_pixels.py
stscirij: This was in response to a flake8 item: the returned dataset wasn't being used, so it wasn't necessary to return it.  The dataset that is passed as a function argument is altered in place, so there's no need to return a copy of it.

hbushouse (approved):

PR #5927, opened 2021-04-02
JP-1949: MOS 0-sized bounding box extraction bug

### LINKED ISSUE #5807: NIRSpec MOS bug with 0-sized 2D extractions
_Issue [JP-1949](https://jira.stsci.edu/browse/JP-1949) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

I've run into a bug processing a simulated MOS exposure.  For this particular case, the spec2 pipeline crashes at the flat_field step for the NRS1 detector because one of the slits has an x-dimension of zero.  Normally, assign_wcs skips any slits for which the bounding box is completely outside the detector, but for some reason that is not happening here.  The pipeline does run successfully for NRS2.

tapastro:
Closes #5807 

Resolves [JP-1949](https://jira.stsci.edu/browse/JP-1949)

**Description**

This PR addresses a bug found when validating slits in the assign_wcs/nirspec code. Near-duplicate code had a missing slit_id shift caused by 0- vs 1- indexing of fixed slits vs msa slitlets - a slit on the edge of the detector in one bounding box was off the detector when calculated correctly. This near-duplicate code has also been cleaned up by calling a common function instead.

Note: Have not yet checked existing regtests, but could reasonably expect this to cause issues, as this moved the bounding box by ~2.5 pixels. Checking on this. 

Note2: Checked, the two nirspec_mos_spec2 regtests run without error.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,10 @@ assign_wcs
 - Changed evaluation of grism bounding box center from averaged extrema of
   transformed bounding box to transformed centroid of source_cat object [#5809]
 
+- Added pixel shift to non-fixed slits due to 0-indexing in NIRSpec slit
+  validation code, fixing difference between bounding box locations during the
+  separate halves of assign_wcs runs [#5927]
+
 associations
 ------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,7 +12,7 @@ assign_wcs
 - Changed evaluation of grism bounding box center from averaged extrema of
   transformed bounding box to transformed centroid of source_cat object [#5809]
 
-- Added pixel shift to non-fixed slits due to 0-indexing in NIRSpec slit
+- Added pixel shift to msa slits due to 0-indexing in NIRSpec slit
   validation code, fixing difference between bounding box locations during the
   separate halves of assign_wcs runs [#5927]
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,9 +12,9 @@ assign_wcs
 - Changed evaluation of grism bounding box center from averaged extrema of
   transformed bounding box to transformed centroid of source_cat object [#5809]
 
-- Added pixel shift to msa slits due to 0-indexing in NIRSpec slit
-  validation code, fixing difference between bounding box locations during the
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -12,6 +12,10 @@ assign_wcs
 - Changed evaluation of grism bounding box center from averaged extrema of
   transformed bounding box to transformed centroid of source_cat object [#5809]
 
+- Added pixel shift to non-fixed slits due to 0-indexing in NIRSpec slit
```
hbushouse, line 15: Maybe phrase this in the positive sense, rather than negative, i.e. "Added pixel shift to MSA slits due to ..."

jwst/assign_wcs/nirspec.py
```diff
@@ -1657,31 +1657,24 @@ def _is_valid_slit(domain):
...
-                valid = _is_valid_slit(bb)
-                if not valid:
-                    log.info("Removing slit {0} from the list of open slits because the "
-                             "WCS bounding_box is completely outside the detector.".format(slit.name))
-                    log.debug("Slit bounding_box is {0}".format(bb))
-                    idx = np.nonzero([s.name == slit.name for s in open_slits])[0][0]
-                    open_slits.pop(idx)
 
-    msa.close()
+    slit2msa = slit_to_msa(open_slits,reference_files['msa'])
```
hbushouse, line 1663: Uh oh. Flake8 isn't going to like the fact that there's no whitespace after the comma in the function argument list.

tapastro: Hm, didn't see it fail the style check, but fixed it up.

hbushouse: I guess we don't have that particular test activated. I've just gotten so used to it, that I assume it'll fail the check.

CHANGES.rst
tapastro: Covered this in the most recent commit.

hbushouse (approved):
LGTM

PR #5929, opened 2021-04-06
JP-1829: Fix resample_spec output size from input images crossing RA=0

### LINKED ISSUE #5564: resample_spec output frame is spatially too small
_Issue [JP-1829](https://jira.stsci.edu/browse/JP-1829) was created on JIRA by [Beth Sargent](https://jira.stsci.edu/secure/ViewProfile.jspa?name=sargent):_

Originally, I was testing extract1d in Build 7.5, and I had developed notebooks for this (please see JP-1804).  When I use the notebook for testing extract1d for LRS Fixed Slit as part of calwebb_spec3 in Build 7.6, I get an error.  When I look into the error, I see a problem with the output of resample_spec in calwebb_spec3 for Build 7.6.  The output array is all zeros and is too narrow.  In Build 7.5, resample_spec seemed to work correctly.

jdavies-st:
Closes #5564
Resolves [JP-1829](https://jira.stsci.edu/browse/JP-1829)

**Description**

This PR fixes the case where the WCS of the input images spans RA=0.

- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -205,15 +205,12 @@ class ResampleSpecData:
             this_maxw = np.max(wavelength_array)
             all_minw = np.min(all_wavelength)
             all_maxw = np.max(all_wavelength)
-
             if this_minw < all_minw:
                 addpts = wavelength_array[wavelength_array < all_minw]
-                for ip in range(len(addpts)):
-                    all_wavelength.append(addpts[ip])
+                all_wavelength = np.append(all_wavelength, addpts)
             if this_maxw > all_maxw:
                 addpts = wavelength_array[wavelength_array > all_maxw]
-                for ip in range(len(addpts)):
-                    all_wavelength.append(addpts[ip])
+                all_wavelength = np.append(all_wavelength, addpts)
 
         # done looping over set of models
         all_ra = np.hstack(all_ra_slit)
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -142,20 +142,20 @@ class ResampleSpecData:
             # first input model sets intializes wavelength array and defines
             # the spatial scale of the output wcs
             if im == 0:
-                for iw in wavelength_array:
-                    all_wavelength.append(iw)
+                all_wavelength = np.append(all_wavelength, wavelength_array)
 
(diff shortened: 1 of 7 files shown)


jwst/resample/resample_spec.py
```diff
@@ -142,20 +142,20 @@ def build_interpolated_output_wcs(self, refmodel=None):
...
                 else:
-                    ra_center = ra[:, lam_center_index]
-                    dec_center = dec[:, lam_center_index]
-                # find the ra and dec for this slit using center of slit
-                ra_center_pt = np.nanmean(ra_center)
-                dec_center_pt = np.nanmean(dec_center)
+                    ra_slice = ra[:, lam_center_index]
+                    dec_slice = dec[:, lam_center_index]
+                # wrap RA if near zero
+                ra_center_pt = np.nanmean(wrap_ra(ra_slice))
```
jdavies-st, line 157: This was the bug.  For RA~0, the mean of the RA slice was the mean of zeros and 359, so an output frame centered at RA~180 deg.

nden:
LGTM.
Since there'e nothing specific to cube_build in `wrap_ra`, I'd suggest moving it to a more central place so it can be reused.

jdavies-st:
Yeah, I'd like to do a follow-up PR that does a bit of refactoring to `wrap_ra`, replacing it with a function that uses `astropy.coordinates.SkyCoord` midpoint functionality.  As it is, we only deal with the discontinuity at RA=0, but there's also a discontinuity at zenith.  Instead of reinventing the wheel, we should just be using `SkyCoord` objects to deal with it correctly.  I'll file a separate issue.

hbushouse (approved):
LGTM.  So this only affected `resample_spec` and not `resample` (for images)?

jdavies-st:
@hbushouse right, it was only in the `build_interpolated_output_wcs()` function used for spectral slits.  For imaging, we use the `footprint()` of each WCS and a separate `wcs_from_fiducial()` function from `gwcs`.

jdavies-st:
Added one last commit to fix a bug in my code that the NIRSpec MOS lev3 regression tests revealed.

Issue #5798, opened 2021-03-01
Generate ERR array in resampled data


stscijgbot-jp:
_Issue [JP-1944](https://jira.stsci.edu/browse/JP-1944) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

Create an error array in resampled output data that is generated from the IVM weight and var_poisson

stscijgbot-jp:
_Comment by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies) on [JIRA](https://jira.stsci.edu/browse/JP-1944?focusedCommentId=516862#comment-516862):_

After discussion with interested parties, the *initial* implementation to populate the output resampled `err` array should do the following:

* Drizzle `1 / err**2` with unity weighting as if it was the science data, and convert back to rms units and store it in the `err` attribute in the resampled datamodel.
* We don't want to weigh this by inverse variance.  The weight map should be unity.
* Handle any zero variance pixel values as NaN when we take the inverse.

In addition to the above new `err` in the resampled output, we should do the exact same thing to each variance array.

* Drizzle each of `1 / var_rnoise`, `1 / var_poisson`, `1 / var_flat` in the same way as above, as if it was the science data, again with unity weighting, and then convert back to variance units, writing it to its respective attribute in the output resampled datamodel.
* Again handle zero variance as NaN during this process.

The `weight` attrribute (WHT in the FITS file) will still just contain the cumulative weight of the drizzled science data.  That will not change.  Toss aside and WHT and CON produced in the drizzling of the error and variance arrays described above.

Note: Resampled `var_rnoise` should be equivalent to the weight map `weight` produced when drizzling the science array.  Write a unit test that verifies this, and then only do the calculation once at runtime.

stscijgbot-jp:
_Comment by [Karl Gordon](https://jira.stsci.edu/secure/ViewProfile.jspa?name=kgordon) on [JIRA](https://jira.stsci.edu/browse/JP-1944?focusedCommentId=516875#comment-516875):_

Minor comment.  The WHT should equal 1 / var_rnoise mosaic image.

stscijgbot-jp:
_Comment by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies) on [JIRA](https://jira.stsci.edu/browse/JP-1944?focusedCommentId=516883#comment-516883):_

[...]

PR #5941, opened 2021-04-07
Propagate variance arrays into SlitModel used as input for ResampleSpecStep

jdavies-st:
Addresses #5798 
Addresses [JP-1944](https://jira.stsci.edu/browse/JP-1944)

**Description**

This PR allows propagation of the variance arrays of the inputs to `ResampleSpecStep`.  Previously these were being dropped on the floor when converting an `ImageModel` into a `SlitModel`, which happens in the case of MIRI LRS.

Regression tests pass on this branch.

This is of course needed if we are to use the variance arrays for weighting and to compute uncertainties in the resampled products.

- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -112,6 +112,11 @@ refpix
 
 - Fixed a bug introduced in #5926 that affected refpix calibration of 1-amp NIR subarrays [#5937]
 
+resample
+--------
+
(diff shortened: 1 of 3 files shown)


hbushouse (approved):
LGTM

PR #5946, opened 2021-04-08
JP-2002: Update ASN rules for TSGRISM with PUPIL=CLEAR

### LINKED ISSUE #5888: Don't create ASN's for EXP_TYPE='NRC_TSGRISM' and PUPIL='CLEAR'
_Issue [JP-2002](https://jira.stsci.edu/browse/JP-2002) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

Some NIRCam commissioning and calibration programs use the APT Engineering Imaging template with one of the subarrays intended for time-series grism, which results in EXP_TYPE='NRC_TSGRISM', but leave the pupil wheel set to clear. This results in exposures that look like TSO grism, because of the EXP_TYPE value, but really aren't (no grism in the beam). These exposures should only be processed up through level-2a and stop there. Hence the ASN rules need to be updated so that no ASN's are created for exposures with EXP_TYPE='NRC_TSGRISM' and PUPIL='CLEAR'.

hbushouse:
Closes #5888 
Resolves [JP-2002](https://jira.stsci.edu/browse/JP-2002)

**Description**

This PR updates the level2b and level3 ASN rules for TSO modes to exclude any exposures that have EXP_TYPE='NRC_TSGRISM' and PUPIL='CLEAR', which can arise when using the NIRCam Engineering Imaging template in APT.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


jwst/associations/lib/rules_level2b.py
```diff
@@ -319,6 +319,23 @@ def __init__(self, *args, **kwargs):
...
+                        ),
+                        DMSAttrConstraint(
+                            name='pupil',
+                            sources=['pupil'],
+                            value=['clear'],
+                        )],
+                    )
+                ],
+                reduce=Constraint.notany
+            )
```
hbushouse, line 338: Note that this dual-level Constraint structure is necessary in order to force the individual constraints on exp_type and pupil value be treated as a logical AND, i.e. the `reduce` operation will now only kick in if both of the inner constraints are true.

jwst/associations/lib/rules_level3.py
```diff
@@ -667,6 +667,25 @@ def __init__(self, *args, **kwargs):
                 ],
                 reduce=Constraint.notany
             ),
+            # Don't allow NIRCam engineering mode
+            # with PUPIL='CLEAR' in tso3
+            Constraint(
+                [
+                    Constraint([
+                        DMSAttrConstraint(
+                            name='restricted_grism',
```
hbushouse, line 676: Because the containing Asn Class already has another constraint involving exp_type, I wasn't sure if it was OK to reuse the same `name` value, so I followed the lead of another rule and made up a different name, `restricted_grism`.

jwst/associations/lib/rules_level3.py
```diff
@@ -667,6 +667,25 @@ def __init__(self, *args, **kwargs):
...
+                            sources=['exp_type'],
+                            value = ('nrc_tsgrism')
+                        ),
+                        DMSAttrConstraint(
+                            name='grism_clear',
```
hbushouse, line 681: Same thing here, where I didn't want to reuse the name `pupil`, so I used `grism_clear` instead.

ddavis-stsci (approved):
LGTM

tapastro (approved):
^What he said. LGTM

PR #5951, opened 2021-04-09
JP-2035: Ramp fit ols gls sep

kmacdonald-stsci:
Closes #5943

Resolves [JP-2035](https://jira.stsci.edu/browse/JP-2035)

**Description**

This PR separates the OLS and GLS algorithms to their own files, instead of both in the one `ramp_fit.py` file.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

PR #5953, opened 2021-04-12
JP-2036: refactor `fit_next_segment`

### LINKED ISSUE #5944: Refactor `fit_next_segment` in `ramp_fit.py`
_Issue [JP-2036](https://jira.stsci.edu/browse/JP-2036) was created on JIRA by [Kenneth MacDonald](https://jira.stsci.edu/secure/ViewProfile.jspa?name=kmacdonald):_

This function is very long and hard to follow being over 400 lines long and a series of special cases to handle while fitting a segment in a ramp.  It would benefit from being broken down in smaller chunks with function calls, rather than everything being done in this function.

kmacdonald-stsci:
Closes #5944

Resolves [JP-2036](https://jira.stsci.edu/browse/JP-2036)

**Description**

This PR breaks down the large function `fit_next_segment`.  It is long because it processes a series of special cases.  Each case has been moved to its own function.

Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/ramp_fitting/ols_fit.py
+++ b/jwst/ramp_fitting/ols_fit.py
@@ -637,22 +637,6 @@ def ols_ramp_fit_single(
     orig_ngroups = ngroups
     orig_cubeshape = (ngroups, nrows, ncols)
 
-    '''
-    # For MIRI datasets having >1 group, if all pixels in the final group are
-    #   flagged as DO_NOT_USE, resize the input model arrays to exclude the
-    #   final group.  Similarly, if leading groups 1 though N have all pixels
-    #   flagged as DO_NOT_USE, those groups will be ignored by ramp fitting, and
-    #   the input model arrays will be resized appropriately. If all pixels in
-    #   all groups are flagged, return None for the models.
-    if input_model.meta.instrument.name == 'MIRI' and ngroups > 1:
-        # The function returns False if the removed groups leaves no data to be
-        # processed.  If this is the case, return None for all expected variables
-        # returned by ramp_fit
-        miri_ans = discard_miri_groups(input_model)
-        if miri_ans is not True:
-            return [None] * 3
-    '''
-
     if ngroups == 1:
         log.warning('Dataset has NGROUPS=1, so count rates for each integration')
         log.warning('will be calculated as the value of that 1 group divided by')
--- a/jwst/ramp_fitting/ols_fit.py
+++ b/jwst/ramp_fitting/ols_fit.py
@@ -105,7 +105,7 @@ def ols_ramp_fit_multi(
     #   flagged as DO_NOT_USE, those groups will be ignored by ramp fitting, and
     #   the input model arrays will be resized appropriately. If all pixels in
     #   all groups are flagged, return None for the models.
-    if input_model.meta.instrument.name == 'MIRI' and input_model.data.shape[1]> 1:
+    if input_model.meta.instrument.name == 'MIRI' and input_model.data.shape[1] > 1:
         miri_ans = discard_miri_groups(input_model)
         # The function returns False if the removed groups leaves no data to be
         # processed.  If this is the case, return None for all expected variables
--- a/jwst/ramp_fitting/ols_fit.py
+++ b/jwst/ramp_fitting/ols_fit.py
@@ -99,19 +99,35 @@ def ols_ramp_fit_multi(
     total_cols = input_model.data.shape[3]
     number_of_integrations = input_model.data.shape[0]
 
+    # For MIRI datasets having >1 group, if all pixels in the final group are
+    #   flagged as DO_NOT_USE, resize the input model arrays to exclude the
+    #   final group.  Similarly, if leading groups 1 though N have all pixels
+    #   flagged as DO_NOT_USE, those groups will be ignored by ramp fitting, and
+    #   the input model arrays will be resized appropriately. If all pixels in
+    #   all groups are flagged, return None for the models.
+    if input_model.meta.instrument.name == 'MIRI' and input_model.data.shape[1]> 1:
+        miri_ans = discard_miri_groups(input_model)
+        # The function returns False if the removed groups leaves no data to be
+        # processed.  If this is the case, return None for all expected variables
+        # returned by ramp_fit
+        if miri_ans is not True:
+            return [None] * 3
+
     # Call ramp fitting for the single processor (1 data slice) case
     if number_slices == 1:
         max_segments, max_CRs = calc_num_seg(input_model.groupdq, number_of_integrations)
         log.debug(f"Max segments={max_segments}")
 
-        # Create output models to be populated after ramp fitting.
-        int_model, opt_model, out_model = \
-            create_output_models(input_model, number_of_integrations, save_opt,
-                                 total_cols, total_rows, max_segments, max_CRs)
-
         # Single threaded computation
         new_mdl, int_mdl, opt_res = ols_ramp_fit_single(
             input_model, int_times, buffsize, save_opt, readnoise_2d, gain_2d, weighting)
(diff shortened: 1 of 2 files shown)


dmggh (commented):
A few comments, but otherwise looks good.

jwst/ramp_fitting/ols_fit.py
```diff
@@ -0,0 +1,3491 @@
...
+            wh_check, start, end_st, end_heads, got_case, ngroups)
+
+    # CASE OTHER: all other types of segments not covered earlier. No segments
+    #   handled here have adequate data, but the stack arrays are updated.
+    wh_check = np.asarray(np.where(~pixel_done & ~got_case))
+    if len(wh_check[0]) > 0:
+        fit_next_segment_all_other(wh_check, start, end_st, end_heads, ngroups)
+
+    return f_max_seg, num_seg
+
```
dmggh, line 1871: You may want to separate out the related fit_next_XYZ ('case') functions into another file.

hbushouse: I second the above suggestion.

jwst/ramp_fitting/ols_fit.py
```diff
@@ -0,0 +1,3491 @@
...
+    input_model.meta.exposure.groupgap = groupgap
+    input_model.meta.exposure.nframes = nframes
+    input_model.meta.exposure.drop_frames1 = dropframes1
+
+    # Compute ramp fitting
+    new_model, int_model, opt_res = ols_ramp_fit_single(
+        input_model, int_times, buffsize, save_opt, readnoise_2d, gain_2d, weighting)
+
+    if new_model is None:
+        return None * 22
```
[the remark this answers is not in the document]
kmacdonald-stsci: It does cause a failure.  I will change this, but this brings up a larger issue of what to do if a slice fails.  Starting at line 

https://github.com/spacetelescope/jwst/blob/7908fbe3996fe343675ecc9fac2e90dd431dcb9e/jwst/ramp_fitting/ols_fit.py#L173 

the processing assumes non-`None` values and continues that way for the loop to populate the output models.  If any particular slice returns `None` values, should the final results be None or on that slice?  If the latter, the output models need to be created using a different method because it's possible for the first slice to be `None`.

These issues will be corrected in JP-2037, since this is the multiprocessing section of the code.

jwst/ramp_fitting/ols_fit.py
kmacdonald-stsci: The `ols_fit.py` file is a large file and could be broken down into several smaller files.  Should other parts of the code be identified to separate into their own file?

jwst/ramp_fitting/ols_fit.py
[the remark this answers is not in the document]
hbushouse: I would think that if any slice fails for some reason, the whole process should fail (i.e. don't create an output that has gaps in it).

nden:
Needs a rebase - looks like it picked up changes from #5951

jwst/ramp_fitting/ols_fit.py
dmggh: To further segregate related functions in ols_fit.py,  here are some rough groupings of similar functions that you could consider moving into other modules.  (This is just based on a quick review, and there is some overlap here  - consider this just as a guideline): 

# Output model-related 
def set_output_models( )
def create_output_models( )

# OLS ramp fitting
def ols_ramp_fit_multi( )
def ols_ramp_fit_sliced()
def ols_ramp_fit_single()

# JWST-specific
def discard_miri_groups( )

# Slope fitting
def ramp_fit_slopes( )
def ramp_fit_compute_variances( )
def ramp_fit_overall( ) # needs epydoc description
def calc_slope( )

# Perhaps move to utils.py
def calc_power(snr)
def interpolate_power(snr) # needs epydocs

# Fitting particular segments/ramps
def fit_next_segment( ) 
def fit_lines( )
def fit_single_read( )
def fit_double_read( )
def fit_1_group( )
def fit_2_group( )
def calc_num_seg( )

# Calculation of goodness-of-fit parameters
def calc_unwtd_fit( )
def calc_opt_fit( )
def calc_unwtd_sums( )
def calc_opt_sums( )

jwst/ramp_fitting/tests/test_ramp_fit.py
```diff
@@ -681,13 +682,39 @@ def test_miri_first_last():
...
 
     np.testing.assert_allclose(new_mod.data, 10. / 3., rtol=1E-5)
 
 
+def test_miri_no_good_pixel():
+    ''' 
+    With no good data, MIRI will remove all groups where all pixels are bad.
+    If all groups are bad, NoneType is returned for all return values from
+    ramp_fit.
+    '''
```
dmggh, line 696: How does "no good data" differ from "all groups are bad" ?

dmggh (commented):
2 questions about a test. Otherwise looks good.

jwst/ramp_fitting/tests/test_ramp_fit.py
```diff
@@ -681,13 +682,39 @@ def test_miri_first_last():
...
+
+    # Dummy non-zero data to make sure if processing occurs a non-NoneType gets
+    # returned.
+    model1.data[0, :, 0, 0] = np.array([-200., -500.], dtype=np.float32)
+    model1.data[0, :, 0, 1] = np.array([-300., -600.], dtype=np.float32)
+    model1.data[0, :, 1, 0] = np.array([-200., 900.], dtype=np.float32)
+    model1.data[0, :, 1, 1] = np.array([-600., -400.], dtype=np.float32)
+
+    # For all pixels, set gdq for 0th and final groups to DO_NOT_USE
+    model1.groupdq[:, :, :, :] = DO_NOT_USE
```
dmggh, line 710: Doesn't that set ALL groups to DO_NOT_USE ?

jwst/ramp_fitting/tests/test_ramp_fit.py
kmacdonald-stsci: I don't understand this question.

jwst/ramp_fitting/tests/test_ramp_fit.py
kmacdonald-stsci: Yes.

dmggh:
1. Does "With no good data, MIRI will remove all groups where all pixels are bad." mean the same as:

For a group with no good data (meaning all pixels in that group are bad) , MIRI will remove the group.


The first sound likes no good data in *all* groups.


2. Because of :

    # Set all groups to DO_NOT_USE

    model1.groupdq[:, :, :, :] = DO_NOT_USE



. . . does it matter that the data has some if its values set to non-zero ?

jwst/ramp_fitting/tests/test_ramp_fit.py
kmacdonald-stsci: The purpose of the test is to cover the portion of the code where `discard_miri_groups` ends up discarding all groups, i.e., discarding all data.  Because no test covered this code, the bug you found earlier never got detected.  Once I realized this case wasn't being covered I added this test.  I changed the wording of the comments to clarify this better.  Let me know if I need to further clarify the comments.

Additionally, I moved the `discard_miri_groups` to be one of the first functions called.  The discarding of groups in MIRI datasets applies to the dataset as a whole.  For single threading, where it was originally was fine, but for multithreading the effect was to call `discard_miri_groups` for each slice, which is inappropriate.

dmggh (approved):
Looks good to me.

nden:
@hbushouse Can this be merged?

hbushouse (approved):

PR #5954, opened 2021-04-13
JP-824: Add option to save combined background image

### LINKED ISSUE #3746: Add option to background step to save combined background image
Issue [JP-824](https://jira.stsci.edu/browse/JP-824) was created by Howard Bushouse:

The Cal WG has requested that an option be added to the calwebb_image2 and calwebb_spec2 background step to allow the combined background image to be saved, when desired. This is similar to the flat-field step capability to save the flat-field image that's constructed on the fly.

tapastro:
Closes #3746

Resolves [JP-824](https://jira.stsci.edu/browse/JP-824)

**Description**

This PR adds the option to save the combined background image model, in a method pulled directly from the optional flat_field output. Currently will not work as written, oops. Do we really need to close that bkg_model?

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/background/background_sub.py
+++ b/jwst/background/background_sub.py
@@ -11,7 +11,7 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 
-def background_sub(input_model, bkg_list, sigma, maxiters):
+def background_sub(input_model, bkg_list, sigma, maxiters, save_combined_background):
     """
     Short Summary
     -------------
@@ -45,7 +45,7 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
     bkg_model.close()
 
     # We're done. Return the result.
-    return result
+    return bkg_model, result
 
 
 def average_background(bkg_list, sigma, maxiters):
--- a/jwst/background/background_sub.py
+++ b/jwst/background/background_sub.py
@@ -42,7 +42,7 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
     result = subtract_images.subtract(input_model, bkg_model)
 
     # Close the average background image and update the step status
-    #bkg_model.close()
+    # bkg_model.close()
 
     # We're done. Return the result.
     return bkg_model, result
--- a/jwst/background/background_sub.py
+++ b/jwst/background/background_sub.py
@@ -99,7 +99,6 @@ def average_background(bkg_list, sigma, maxiters):
 
     # Compute the combined ERR as the uncertainty in the mean
     avg_bkg.err = (np.sqrt(merr.sum(axis=0)) / (num_bkg - merr.mask.sum(axis=0))).data
-    # avg_bkg.err[merr.mask.sum(axis=0)] = np.nan
 
     return avg_bkg
 
--- a/jwst/background/background_sub.py
+++ b/jwst/background/background_sub.py
@@ -42,7 +42,7 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
     result = subtract_images.subtract(input_model, bkg_model)
 
     # Close the average background image and update the step status
-    bkg_model.close()
+    #bkg_model.close()
 
     # We're done. Return the result.
     return bkg_model, result
--- a/jwst/background/background_sub.py
+++ b/jwst/background/background_sub.py
@@ -11,7 +11,7 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 
-def background_sub(input_model, bkg_list, sigma, maxiters, save_combined_background):
+def background_sub(input_model, bkg_list, sigma, maxiters):
     """
     Short Summary
     -------------
--- a/jwst/background/background_sub.py
+++ b/jwst/background/background_sub.py
@@ -41,9 +41,6 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
     log.debug(' subtracting avg bkg from {}'.format(input_model.meta.filename))
     result = subtract_images.subtract(input_model, bkg_model)
 
-    # Close the average background image and update the step status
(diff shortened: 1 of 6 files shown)


hbushouse (commented):
Code looks good. Should update the background step docs to mention this, and of course the requisite change log entry.

hbushouse:
I think we might have at least 1 regtest that applies the `background` step during `calwebb_spec2` processing (might be a MIRI LRS slit observation?), which could be updated to include saving the background and doing a comparison on it. Then you can even check off the `Tests` box above!

jdavies-st:
Yeah, the regression test is 

https://github.com/spacetelescope/jwst/blob/434c5406c45f6be5c3d73f40fe9e71f122c185fa/jwst/regtest/test_miri_lrs_slit_spec2.py#L28-L34

Also, looking at the test, it seems we are skipping resample in that test for the following reason:

https://github.com/spacetelescope/jwst/pull/4373

We should turn resampling back on, and test the `s2d` file.

tapastro:
It took me a bit to find some code I could test this with, having forgotten to check the regtest data first. Using the data from the test mentioned above, I'm getting an error when attempting the save_model call; the traceback goes through many lines of of astropy fits io but ends in a numpy core line with "NotImplementedError: MaskedArray.tofile() not implemented yet." Has this been seen before? Is some model massaging required?

hbushouse:
#4373 Ah, from the great regtest marathon, when all @hbushouse and @jdavies-st did for Christmas was write tests!  Those were the days. ;-)

tapastro:
@hbushouse @jdavies-st It took me a bit to find some code I could test this with, having forgotten to check the regtest data first. Using the data from the test mentioned above, I'm getting an error when attempting the save_model call; the traceback goes through many lines of of astropy fits io but ends in a numpy core line with "NotImplementedError: MaskedArray.tofile() not implemented yet." Has this been seen before? Is some model massaging required? It seems as though the culprit is the err array in the ImageModel. I can't believe this is the first time an err is being saved...

jwst/background/background_sub.py
```diff
@@ -42,10 +42,10 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
     result = subtract_images.subtract(input_model, bkg_model)
 
     # Close the average background image and update the step status
```
stscieisenhamer, line 44: Suggest removing the commented code completely; the model is being returned and there should not be even a hint of closing it.

stscieisenhamer (approved):

stscieisenhamer (dismissed):
Need changelog!

tapastro:
> Need changelog!

@stscieisenhamer , do you have a suggestion on the comment I made above regarding issues printing an ImageModel with an included MaskedArray to file?

stscieisenhamer:
Sorry, missed that comment. The issue is not that an err array is being saved, but this is a masked err array. Others a bit more knowledgeable of the code base may know different, however I do not believe any of the `DataModel` arrays can be numpy masked arrays. The masking is explicit in the dq array, and all other data arrays should be non-masked. This may be the first time an attempt at saving a `DataModel` with masked arrays has been attempted, and the claim is that all arrays should be unmasked and that the dq array is consistent.

hbushouse:
So the obvious next question is where/how one of the arrays in the background model is ending up as a masked array. Have you been able to figure that out?  Perhaps by whatever routine is being used to create the combined/sigma-clipped array?

tapastro:
> So the obvious next question is where/how one of the arrays in the background model is ending up as a masked array. Have you been able to figure that out? Perhaps by whatever routine is being used to create the combined/sigma-clipped array?

Finding that wasn't too difficult - the err array is explicitly initialized as a masked array. I've been futzing around with the output though, to understand exactly how I should be "collapsing" it - the mask is in place to utilize the results of a sigma_clip on the background data. The combined err does an uncertainty calculation, but it calls the masked array without using .data or .mask, and I'm not sure exactly what is being fed into the line ([Line here](https://github.com/spacetelescope/jwst/blob/e2a402acdf007d5b6b21bd168376ea425cadbedb/jwst/background/background_sub.py#L104)). If I were to force-apply the mask to the data, is the goal of the sigma_clip to remove those entries from the error calculation?

hbushouse (commented):
Looks like this latest update should fix the model saving problem. Can remove the 1 commented line.

Still needs a change log entry and step docs update.

jwst/regtest/test_miri_lrs_slit_spec2.py
```diff
@@ -30,15 +30,16 @@ def run_pipeline(jail, rtdata_module):
             "--save_bsub=true",
             "--steps.assign_wcs.save_results=true",
             "--steps.flat_field.save_results=true",
-            "--steps.srctype.save_results=true"]
+            "--steps.srctype.save_results=true",
+            "--steps.bkg_subtract.save_combined_background=true"]
```
jdavies-st, line 34: While you'r at it, could you delete the `"--steps.resample_spec.skip=true",` in this test and remove the comment above it about there being a bug.  But is fixed and results look good.

I've already added the truth file to artifactory, so it should run without problem, though you might see small precision differences when you run it locally.

tapastro: Yeah, I wasn't sure if your comment above should be done here or elsewhere, but I can throw it in for good measure. I haven't added the truth file for the saved background, but hoping to okify that, if that works... 😬

jdavies-st: I can walk you through adding it over slack if you like.

hbushouse (approved):
Looks good. All the boxes checked!

PR #5962, opened 2021-04-14
TST: Use astropy nightly build

pllim:
**Description**

This PR attempts to use `astropy` nightly build instead of building from source.

PROS: Faster run time for test suite!

CONS: You won't see new changes from "astropy dev" until they go into the next nightly build. And if build is broken, you need to bug Stuart Mumford.

Checklist

- [x] Tests -- This controls all the dev tests (*my preciousssss*)
- [x] Documentation -- N/A
- [x] Change log -- N/A
- [ ] Milestone -- Up to you
- [ ] Label(s) -- Up to you

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 git+https://github.com/spacetelescope/asdf
-git+https://github.com/astropy/astropy
+--extra-index-url=https://pkgs.dev.azure.com/astropy-project/astropy/_packaging/nightly/pypi/simple/ --pre astropy
 git+https://github.com/spacetelescope/crds
 git+https://github.com/spacetelescope/gwcs
 git+https://github.com/astropy/photutils


pllim:
I actually can't tell where you actually test using astropy dev, though it is defined in your `tox.ini`... @jdavies-st ?

jdavies-st:
Thanks @pllim.  We only test with dev dependencies once-a-week as a Monday morning cron job in the ci_cron.yml workflow.  It can also can be kicked off via manual dispatch.

So speed is probably not high priority for us, though for manual dispatch, having to wait for a nightly build might not be useful, so maybe building from scratch is preferable.

pllim:
OK in that case, feel free to close this without merge. I just vaguely remember someone asked for this before but maybe it wasn't you or even this package. 😆 

If you change your mind, now you know and knowledge is power, kids.

jdavies-st:
Thanks @pllim .  Yeah, I think I'll close.

I recall being interested in getting the `numpy` nightly build (done by Anaconda folks?) to add to our dev builds?  I recall we had a hard time doing the build under Travis.  Not sure we've tried it with Github actions.  So currently our dev dependencies build does not use dev numpy.

pllim:
`astropy` pulls numpy nightly (more like weekly actually), so you can look there.

PR #5969, opened 2021-04-19
Speed up NRS IFU footprint computation

nden:
**Description**

This PR speeds up the computation of  NIRSpec IFU footprints by using the WCS transforms directly,
instead of through the WCS object. Most of the transforms are reused and only slice specific differences are 
computed.

The computed spectral range is saved in `meta.wcsinfo.spectral_range`. It's not saved in the FITS file.
This is necessary for improving cube_build performance.

The "true" bounding box in the test was computed using the previous implementation with evaluating the WCS object.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


jwst/assign_wcs/tests/test_nirspec.py
```diff
@@ -857,3 +858,91 @@ def test_functional_ifu_prism():
...
+                (132.5285162704088, 183.27350269292484)),
+           27: ((169.8529496136358, 1640.778485168005),
+                (1826.028691168028, 1880.9336718824313)),
+           28: ((94.71390837793813, 1555.4048050512263),
+                (82.94691422559131, 133.63901517357235)),
+           29: ((172.3681094850081, 1643.685604697228),
+                (1874.8184744639657, 1929.9072657798927))}
+
+    hdul = create_nirspec_ifu_file("F290LP", "G140M")
+    im = datamodels.ImageModel(hdul)
```
hbushouse, line 926: Should this be an `IFUImageModel` or does it not matter to the usage here?

jwst/assign_wcs/util.py
```diff
@@ -938,36 +938,71 @@ def update_s_region_keyword(model, footprint):
...
         The output of assign_wcs.
     mod : module
         The imported ``nirspec`` module.
+
+    Returns
+    -------
+    footprint : ndarray
+        The spatial footprint
+    spctral_region : tuple
+        The wavwlength range for the observation.
```
jemorrison, line 964: typo - should be wavelength

jemorrison (approved):
Except for a few typos it looks fine to me.

jemorrison:
Can we merge this so I can use it in cube_build.

PR #5970, opened 2021-04-19
Update NIRISS SOSS tso3 regression test

hbushouse:
**Description**

This PR addresses an issue in the calwebb_tso3 regression test using NIRISS SOSS data. Recent updates to the outlier_detection step algorithm resulted in massive numbers of false positive detections in the particular dataset used in the regression test, just due to unrealistically noisy data. This update simply increases the value of the "snr" parameter in the outlier_detection step used in the test to reduce the number of detections.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [X] Milestone
- [X] Label(s)


jdavies-st (approved):
Looks good.

Do we know why the SOSS data has such elevated noise?

hbushouse:
> Looks good.
> 
> Do we know why the SOSS data has such elevated noise?

Not really. It's really old simulated SOSS data from the team. The noise is across integrations, since that's the dimension that's being used in the outlier detection now. Can't think of anything in the rest of the calibration process that would create undo amounts of variation from integration to integration (only within an integration, e.g. noisy darks). So I guess the original raw data were that way.

PR #5971, opened 2021-04-19
removing wcs keywords after resample

### LINKED ISSUE #2225: Remove JWST-specific WCS header keywords from resampled FITS files
Based on discussions in PR #2222, we should remove the following keywords once they are no longer relevant to the data, i.e. after an image has been resampled and it is given a standard FITS WCS:

V2_REF
V3_REF
RA_REF
DEC_REF
ROLL_REF

That way there is no confusion as to whether these keywords still apply to the resampled file or not (they don't).

cshanahan1:
WCS keywords ra/dec/roll/v2/v3_ref are removed from im.meta.wcsinfo after ResampleStep.

Fixes #2225 / [JP-2039](https://jira.stsci.edu/browse/JP-2039)

### DIFF
--- a/jwst/resample/resample_step.py
+++ b/jwst/resample/resample_step.py
@@ -93,6 +93,13 @@ class ResampleStep(Step):
         else:
             result = resamp.output_models
 
+        # remove irrelevant WCS keywords
+        rm_keys = ['v2_ref', 'v3_ref', 'ra_ref', 'dec_ref', 'roll_ref',
+                   'v3yangle', 'vparity']
+        for key in rm_keys:
+            if key in result.meta.wcsinfo.instance:
+                del result.meta.wcsinfo.instance[key]
+
         return result
 
     def update_phot_keywords(self, model):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -166,6 +166,8 @@ resample
 
 - Propagate variance arrays into ``SlitModel`` used as input for ``ResampleSpecStep`` [#5941]
 
+- Remove certain WCS keywords that are irrelevant after resampling. [#5971]
+
 source_catalog
 --------------
 
--- a/jwst/resample/tests/test_resample_step.py
+++ b/jwst/resample/tests/test_resample_step.py
@@ -380,3 +380,18 @@ def test_build_interpolated_output_wcs(miri_rate_pair):
     # Make sure the output slit size is larger than the input slit size
     # for this nodded data
     assert driz.data_size[1] > ra.shape[1]
+
+
+def test_wcs_keywords(nircam_rate):
+    # make sure certain wcs keywords are removed after resample
+
+    im = AssignWcsStep.call(nircam_rate)
+    result = ResampleStep.call(im)
+
+    assert result.meta.wcsinfo.v2_ref is None
+    assert result.meta.wcsinfo.v3_ref is None
+    assert result.meta.wcsinfo.ra_ref is None
+    assert result.meta.wcsinfo.dec_ref is None
+    assert result.meta.wcsinfo.roll_ref is None
+    assert result.meta.wcsinfo.v3yangle is None
+    assert result.meta.wcsinfo.vparity is None


hbushouse:
Should probably also remove V3I_YANG and VPARITY. They're part of the whole set imported from the SIAF (along with V2_REF and V3_REF) and pertinent only to the particular aperture that was in use. Any others that anyone can think of?  @jdavies-st @nden

jwst/resample/resample_step.py
```diff
@@ -93,6 +93,13 @@ def process(self, input):
         else:
             result = resamp.output_models
 
+        # remove irrelevant WCS keywords
+        rm_keys = ['v2_ref', 'v3_ref', 'ra_ref', 'dec_ref', 'roll_ref',
+                   'vi_yang', 'vparity']
```
hbushouse, line 98: vi_yang should be "v3yangle"  (see https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/schemas/wcsinfo.schema.yaml#L301)

jdavies-st (approved):
This PR looks good.  Needs a change log in CHANGES.rst

@hbushouse and @jemorrison are these keywords also left behind in other level3 products such as those produced by `cube_build`?  Or is it only `resample` that's doing a `model.update()` from one of the input images that gets these populated?

hbushouse:
I don't know for sure how `cube_build` is populating its output meta data. If it is doing a `model.update()`, then these keywords will also be in those output models and should be removed, just as you're doing here for `resample`. So take a look at the `cube_build` code to confirm.

hbushouse:
And since you brought it up, what about `resample_spec`?  That may need a similar update.

jemorrison:
cube_build is using a 'update'

ifucube_model.update(model_ref)
where model_ref if the first model in the set of data to combine

nden:
I wonder if this needs some more thought, especially for imaging mode.
If resample runs in stage 2 and removes the keywords would this be a problem for stage 3.
Also, at least for HST alignment with GAIA, the code makes "sanity" checks that alignment did not introduce huge errors, e.g. by identifying the wrong sources in a crowded field, by comparing with the original pointitng.

hbushouse:
> I wonder if this needs some more thought, especially for imaging mode.
> If resample runs in stage 2 and removes the keywords would this be a problem for stage 3?
> Also, at least for HST alignment with GAIA, the code makes "sanity" checks that alignment did not introduce huge errors, e.g. by identifying the wrong sources in a crowded field, by comparing with the original pointitng.

The i2d (or s2d or s3d) product created at the end of level-2b processing (by calwebb_image2 or calwebb_spec2) for each exposure is a "dead end" product and is not used in any level-3 processing. Only the level-2b "cal" products are passed along to level-3 processing, which are still in unresampled space and hence still have their original full complement of WCS-related keywords.

hbushouse:
I was under the impression that @nden 's question was wondering whether the *removal* of these JWST pointing keywords from products coming out of the resample step in level-2b would cause problems downstream, but since those products aren't used downstream, their absence should not be a problem. Hence I would not try to put in special hooks to make sure those keywords stay in when running at level-2b, but get removed when running at level-3. Just remove them all the time.

hbushouse:
I just looked at an s3d (cube_build) product coming out of the latest software and they too still have the JWST aperture-specific keywords like V2_REF, V3_REF, RA_REF, DEC_REF, etc. So if we're going to remove those keywords from resampled images I feel we should do the same for resampled IFU products and hence a similar update should be made to the cube_build step. I'll leave it to @jemorrison to decide whether we should include that in this PR or if she wants to do a separate PR for cube_build (since she's likely to know immediately where in the cube_build code this should be done).

If the decision is to do a separate PR for cube_build, then I think this one is ready to merge, if there are no objections?

hbushouse (commented):
Still needs a change log entry (in CHANGES.rst) before merging.

jemorrison:
If it is ok I would prefer to do this in the current PR I have for cube build ($5991). I just hate conflicts that  I have to resolve. I think the current cube build PR (#5991) will get reviewed and merged soon. Or if folks want I can do it as a separate PR  after I get 5991 merged

hbushouse:
OK, so the cube_build mods will be done in another PR, so this one just needs a change log entry before it can be merged.

nden (approved):

hbushouse (approved):

PR #5508, opened 2020-12-03
JP-1791: Initial commit of WFSS contam correction step and INS grism library modules

hbushouse:
These are the code modules from Nor Pirzkal's various grism repos that we need in order to support the WFSS contamination corrections. These modules do the work of creating simulated spectra for sources identified in direct images. These will all need a lot of clean-up and rearranging to get them to work from within the jwst pipeline environment. Just getting them into the repo to start with. Some of them may end up going away eventually. The functions/methods in these modules will be called by a new calwebb_spec2 step that will do decontamination of WFSS spectra.

Resolves [JP-1791](https://jira.stsci.edu/browse/JP-1791)
Resolves #5503 
Resolves [JP-1924](https://jira.stsci.edu/browse/JP-1924)
Resolves #5750

PR #5978, opened 2021-04-22
Algorithm for clipping rectangles to rectangular windows

mcara:
This PR adds an algorithm for clipping multiple rectangles to clipping windows. Rectangles to be clipped must have sides parallel to coordinate axes. `get_clipped_pixels()` returns indices of the clipped pixels and their area.

This PR is to support commit https://github.com/spacetelescope/jwst/commit/af1d903e1aeff917397e86b65e26b8733df8f2c8 in #5508

CC: @hbushouse @nden

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,6 @@ classifiers =
 zip_safe = False
 python_requires = >=3.7
 setup_requires =
-    numpy
     setuptools_scm
 install_requires =
     asdf>=2.7.1
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,6 +26,7 @@ classifiers =
 zip_safe = False
 python_requires = >=3.7
 setup_requires =
+    numpy
     setuptools_scm
 install_requires =
     asdf>=2.7.1
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,32 @@
-from os.path import basename
-from setuptools import setup, find_packages
+import sys
+import importlib
 from glob import glob
+from os.path import basename
+from setuptools import setup, find_packages, Extension, _install_setup_requires
 
+
+try:
+    from distutils.config import ConfigParser
+except ImportError:
+    from configparser import ConfigParser
+
+conf = ConfigParser()
+conf.read(['setup.cfg'])
+
+# Get some config values
+metadata = dict(conf.items('metadata'))
+PACKAGENAME = metadata['name']
+
+options = dict(conf.items('options'))
+SETUP_REQUIRES = [
+    s for s in map(str.strip, options['setup_requires'].splitlines()) if s
+]
+INSTALL_REQUIRES = [
+    s for s in map(str.strip, options['install_requires'].splitlines()) if s
+]
 
 scripts = [s for s in glob("scripts/*") if basename(s) != "__pycache__"]
+
 # package_data values are glob patterns relative to each specific subpackage.
 package_data = {
     "": [
@@ -19,23 +42,64 @@ package_data = {
         "tests/data/*.txt",
         "*.yaml",
     ],
+
     "jwst.fits_generator": [
         "templates/*.inc",
         "templates/*.txt",
         "tests/okfile/*.prop",
     ],
+
     "jwst.lib": ["tests/data/siaf.db"],
+
     # Include the rules .py files in associations test data
     "jwst.associations": ["tests/data/*.py"],
+
+    # Include C extensions
+    "jwst.lib.src": ["*.c"],
+
     # Include the transforms schemas
     "jwst.transforms": ["schemas/stsci.edu/jwst_pipeline/*.yaml"],
     "jwst.stpipe.resources": ["schemas/*.yaml"],
 }
 
+# Install packages required for this setup to proceed:
+_install_setup_requires(dict(setup_requires=SETUP_REQUIRES))
+for dep_pkg in SETUP_REQUIRES:
+    try:
+        importlib.import_module(dep_pkg)
+    except ImportError:
+        print(f"{dep_pkg} is required in order to install '{PACKAGENAME}'.\n"
+              f"Please install {dep_pkg} first.",
+              file=sys.stderr)
+        exit(1)
+
+# Setup C module include directories
+import numpy  # noqa: E402
+include_dirs = [numpy.get_include()]
+
+# Setup C module macros
+define_macros = [('NUMPY', '1')]
+
+# Handle MSVC `wcsset` redefinition
+if sys.platform == 'win32':
+    define_macros += [
+        ('_CRT_SECURE_NO_WARNING', None),
+        ('__STDC__', 1)
+    ]
+
 setup(
     use_scm_version=True,
-    setup_requires=["setuptools_scm"],
+    setup_requires=SETUP_REQUIRES,
+    install_requires=INSTALL_REQUIRES,
     scripts=scripts,
     packages=find_packages(),
     package_data=package_data,
+    ext_modules=[
+        Extension(
+            'jwst.lib.winclip',
+            ['jwst/lib/src/winclip.c'],
+            include_dirs=include_dirs,
+            define_macros=define_macros
+        )
+    ],
 )
--- a/setup.py
+++ b/setup.py
@@ -1,32 +1,10 @@
-import sys
-import importlib
-from glob import glob
 from os.path import basename
-from setuptools import setup, find_packages, Extension, _install_setup_requires
+from setuptools import setup, find_packages, Extension
+from glob import glob
+import numpy
 
-
-try:
-    from distutils.config import ConfigParser
-except ImportError:
-    from configparser import ConfigParser
-
-conf = ConfigParser()
-conf.read(['setup.cfg'])
-
-# Get some config values
-metadata = dict(conf.items('metadata'))
-PACKAGENAME = metadata['name']
-
-options = dict(conf.items('options'))
-SETUP_REQUIRES = [
-    s for s in map(str.strip, options['setup_requires'].splitlines()) if s
-]
-INSTALL_REQUIRES = [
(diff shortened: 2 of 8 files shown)


hbushouse:
The new algorithm itself seems to be fine to me, but there are obviously errors in the package setup, given that the test module is not able to import `winclip`. These seem to be similar to the problems I had when trying to get setup to build things properly when including a C module. I can get it to build OK in my local checkout, so that I can run the code, but could never find the right way to get it to build properly in the CI systems.

setup.cfg
```diff
@@ -26,6 +26,7 @@ classifiers =
 zip_safe = False
 python_requires = >=3.7
 setup_requires =
+    numpy
```
jdavies-st, line 29: This is not really needed here, because by the time `setup` is run, numpy needs to be installed already.  Instead you need to list it in pyproject.toml, and use `oldest-supported-numpy` instead, as that's what will be used to build the C code.

See

https://github.com/astropy/astropy/blob/5f8866e41164293e463a3cc53fadb5566ba488cc/pyproject.toml#L7

for an example.

setup.py
```diff
@@ -19,23 +42,64 @@
...
+    define_macros += [
+        ('_CRT_SECURE_NO_WARNING', None),
+        ('__STDC__', 1)
+    ]
+
 setup(
     use_scm_version=True,
-    setup_requires=["setuptools_scm"],
+    setup_requires=SETUP_REQUIRES,
+    install_requires=INSTALL_REQUIRES,
```
jdavies-st, line 93: These are not necessary (nor any of the code above defining the global variables), as pyproject.toml makes sure that `numpy` is installed before `setup()` is run.  And the meta already gets pulled from setup.cfg, as specified in pyproject.toml.

setup.py
```diff
@@ -19,23 +42,64 @@
...
+# Setup C module macros
+define_macros = [('NUMPY', '1')]
+
+# Handle MSVC `wcsset` redefinition
+if sys.platform == 'win32':
+    define_macros += [
+        ('_CRT_SECURE_NO_WARNING', None),
+        ('__STDC__', 1)
+    ]
+
```
jdavies-st, line 89: I believe none of this except `include_dirs` is needed, as again, pyproject.toml assures that `numpy` is installed before `setup()` is run.  Also we don't support Windows (currently).

setup.py
mcara: But if this is not incorrect, i.e., it does not hurt, I would prefer to keep this code in.

As a test, I created a new/clean Python environment and I commented out the code that you say is optional and tried to install `jwst` into this environment. Here is the error that I get:

```python
(clnenv) dacia:jwst mcara$ pip install .
Processing /.../jwst
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
   command: /.../clnenv/bin/python3 /.../clnenv/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /var/folders/kb/78_n024d2jb8p24bf6t2c3_c0001_0/T/tmpbnsu8zik
       cwd: /private/var/folders/kb/78_n024d2jb8p24bf6t2c3_c0001_0/T/pip-req-build-ul60c3p9
  Complete output (16 lines):
    File "/.../clnenv/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py", line 280, in <module>
      main()
    File "/.../clnenv/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py", line 263, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/.../clnenv/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py", line 114, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/private/var/folders/kb/78_n024d2jb8p24bf6t2c3_c0001_0/T/pip-build-env-5c_84cv7/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 154, in get_requires_for_build_wheel
      return self._get_build_requires(
    File "/private/var/folders/kb/78_n024d2jb8p24bf6t2c3_c0001_0/T/pip-build-env-5c_84cv7/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 135, in _get_build_requires
      self.run_setup()
    File "/private/var/folders/kb/78_n024d2jb8p24bf6t2c3_c0001_0/T/pip-build-env-5c_84cv7/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 150, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 77, in <module>
      import numpy
  ModuleNotFoundError: No module named 'numpy'
  ----------------------------------------
```

Should users be able to install `jwst` by clonning it and then running `pip install .`?

setup.py
```diff
@@ -1,9 +1,32 @@
...
+metadata = dict(conf.items('metadata'))
+PACKAGENAME = metadata['name']
+
+options = dict(conf.items('options'))
+SETUP_REQUIRES = [
+    s for s in map(str.strip, options['setup_requires'].splitlines()) if s
+]
+INSTALL_REQUIRES = [
+    s for s in map(str.strip, options['install_requires'].splitlines()) if s
+]
```
jdavies-st, line 26: This is all handled from pyproject.toml build backend, and already currently works. I.e. you don't need to parse setup.cfg.  This already happens because

```
build-backend = "setuptools.build_meta"
```

setup.py
jdavies-st: Yes, if you have numpy listed in `pyproject.toml`, then it _will_ work.  This is the modern way to do it.

jdavies-st: I.e. the only reason you are getting to the point of not having `numpy` in the above traceback is because it is not listed in pyproject.toml as it needs to be.  `jwst` is PEP517-compliant, and this is how one guarantees that the build requirements are installed.

mcara: @jdavies-st Alright. I moved `'numpy'` requirement to `pyproject.toml` and now `pip install .` works but the now `python setup.py install` fails:

```python
jwst mcara$ python setup.py install
WARNING: The wheel package is not available.
  File "setup.py", line 78, in <module>
    import numpy
ModuleNotFoundError: No module named 'numpy'
```

jdavies-st: You need `import numpy` at the top of `setup.py`.  pyproject.toml will guarantee it is installed in the isolated build environment, but you'll still need to import it.

Be sure to use `oldest-supported-numpy` in pyproject.toml, so that we avoid the problems we saw a few months ago when the numpy folks delevered 1.20 and we were still using 1.19 in our deliveries to SDP and it broke the installs.

mcara:
@hbushouse When testing this PR with #5508 (or the other way) you may want to uncomment the following lines in `winclip.c`: https://github.com/spacetelescope/jwst/blob/24b647240aad5d5be9a7e6466a43819568e5fb41/jwst/lib/src/winclip.c#L186-L193
in order to simulate the accuracy loss in the original code [to get results closer to the original code].

setup.py
mcara: @jdavies-st OK. Removed support for installing/compiling on Windows and support for installing using `python setup.py install` in https://github.com/spacetelescope/jwst/pull/5978/commits/a9de43db2bba745cd793f57d20023c7d1b58477a

mcara:
I have no idea how to get this PR to pass this rigorous `codecov` test.

tox.ini
```diff
@@ -38,6 +38,12 @@ passenv =
     TEST_BIGDATA
 
 commands =
+    # The following fix for C-extensions was copied from
+    # https://github.com/spacetelescope/synphot_refactor/blob/
+    # d4aa97f47a2b83e4b255f42087be446f94906354/tox.ini#L56-L58
+    # FIXME: Not sure why need this for tox to see the C-ext
```
pllim, line 44: If you find a solution, please circle back to spacetelescope/synphot_refactor#295 . Thanks!

pllim:
I don't think codecov measures C-extension accurately. 🤷

tox.ini
mcara: I think I found one for `jwst` but it does not work for `synphot_refactor` yet (one failing test). I'll make another PR for `jwst` pipeline first and and then if you like it - you can try it on `synphot_refactor`.

The biggest issue that's killing this idea for `synphot_refactor` is your import of `_astropy_init` (due to affiliated package status).

mcara:
> I don't think codecov measures C-extension accurately. 🤷

No, hire coverage measures coverage of `setup.py` and somehow some imports are not covered.

tox.ini
pllim: Why would `_astropy_init` break `tox`? 🤔

jdavies-st:
> here coverage measures coverage of `setup.py` and somehow some imports are not covered.

It should not be covering `setup.py`.  That's my mistake in our coverage setup.  Feel free to add it setup.py to the omits:

https://github.com/spacetelescope/jwst/blob/0d5269d2760fcd2966fe8480cee10da4492e8939/setup.cfg#L137-L151

tox.ini
```diff
@@ -59,8 +65,11 @@ commands_pre =
     pip install -r requirements-min.txt
 
 [testenv:pyargs]
-changedir = {homedir}
+# changedir = {homedir}
 commands =
+    # FIXME: Not sure why need this for tox to see the C-ext
+    python setup.py build_ext --inplace
+    # End of FIXME
```
jdavies-st, line 72: Commenting out the `changedir` directive will break the purpose of doing `pytest --pyargs`, which is to test the package on the installed version of the package in `site-packages`.  `pip install .` should be installing the build C lib in `site-packages` here.

hbushouse (approved):
Approving now, with the understanding that issues concerning the "correct" way to build C extensions in setup.py and tox.ini will be resolved independently in the near future after this is merged.

PR #5981, opened 2021-04-22
JP-1785: Remove sys.path calls from various locations

### LINKED ISSUE #5487: Remove sys.path calls from associations, stpipe, lib
_Issue [JP-1785](https://jira.stsci.edu/browse/JP-1785) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

There's a number of calls to `sys.path` where the system path is explicitly set in order load .py rules files.  This has some odd knock-on side effects seen in our testing when comparing testing on the checked out repo vs testing against the package installed in site-packages.

An example in the code:

<https://github.com/spacetelescope/jwst/blob/64f2cda84f5b40b102663f10f8826189a55a3372/jwst/associations/registry.py#L475>

Better to use `importlib` to do this.  See

<https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly>

<https://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path>

for further info.

All occurrences:

<https://github.com/spacetelescope/jwst/search?q=%22sys.path%22>

tapastro:
Closes #5487

Resolves JP-1785

**Description**

This PR addresses sys.path calls that explicitly set the path, which have been causing issues "when comparing testing on the checked out repo vs testing against the package installed in site-packages." 

I've addressed the first example location, but may want some feedback on what to do in the case of the associations files using sys.path, e.g.: [here](https://github.com/spacetelescope/jwst/blob/0382b8dbd663c4c1b61f0f91a0402ce19fe3d881/jwst/associations/tests/data/rules_basic.py).

Checklist
- [ ] Tests

- [ ] Documentation

- [ ] Change log

- [x] Milestone

- [x] Label(s)


jdavies-st (commented):
Looking good!  There's still some use of `sys.path` in `jwst.associations`.

https://github.com/spacetelescope/jwst/search?q=sys.path

Ignore the occurrence in `docs`, as that's only used by `sphinx` when building our docs.

Also, that jupyter notebook under `jwst/stpipe/tests` should probably not be there?  It is neither a test nor runtime code.  And there's already a test that checks suffixes.  It should probably be just removed, right @stscieisenhamer ?

Finally, when the issue was filed, `stpipe` was still part of `jwst`.  So it would be great to fix those too, separately, if you're feeling jaunty.  =)

https://github.com/spacetelescope/stpipe/search?q=sys.path

again, ignoring the instance in `docs`.

stscieisenhamer (approved):
Go ahead and remove find_all_suffixes.ipynb

tapastro:
Updating the PR branch and using a more fresh base branch, closing this PR.

PR #5983, opened 2021-04-23
Removing ramp fit from JWST and refer to stcal for ramp fit.

kmacdonald-stsci:
Closes #

Resolves JP-

**Description**

Removing ramp fitting from JWST to use `stcal` instead.

Checklist
- [ ] Tests

- [ ] Documentation

- [ ] Change log

- [ ] Milestone

- [ ] Label(s)


jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -1,8 +1,15 @@
 #! /usr/bin/env python
 
+import numpy as np
```
dmggh, line 3: Here process( ) is calling stcal's ramp_fit( ) with a  JWST RampModel as an input argument - I thought the plan was to call the common ramp_fit() with either a generic datamodel or numpy arrays; perhaps the JWST RampModel is generic enough.

Also, a minor comment: we should update -    
 # Only algorithm allowed for Build 7.1
  # Only weighting allowed for Build 7.1

kmacdonald-stsci: For the moment, the assumption is that the RampModel will be common, or at least common enough with relevant ramp fitting key words being the same, across pipelines.  I agree that the documentation should clearly state the only algorithms available right now are ordinary least squares with "optimal" weighting.

nden:
The tests are failing because there's no `__init__.py` in the ramp_fit directory. Add an empty `__init__.py`

kmacdonald-stsci:
I see a `__init__.py` file in my branch:

https://github.com/kmacdonald-stsci/jwst/blob/fd162e100379c159088d05c8eae5477cf534fa6d/jwst/ramp_fitting/__init__.py#L1

It's not empty because it needs to expose `RampFitStep`.

jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -75,6 +75,126 @@ def compute_int_times(input_model):
     return int_times
 
 
+def create_output_image_model(input_model, image_info):
+    """
+    Creates output ImageModel using keyworsd from the inut RampModel and the
```
dmggh, line 80: keyworsd -> keywords

jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -151,25 +272,38 @@ def process(self, input):
...
-        if out_model is not None:
+        # Save the OLS optional fit product, if it exists
+        if opt_info is not None:
+            opt_model = create_optional_results_model(opt_info)
+            self.save_model(opt_model, 'fitopt', output_file=self.opt_name)
+
+        if image_info is not None:
+            input_model.int_times = old_int_times
+            out_model = create_output_image_model(input_model, image_info)
+            if out_model is not None:
```
dmggh, line 290: I don't see how out_model could be None here.

dmggh (commented):
Looks good so far.

jwst/ramp_fitting/ramp_fit_step.py
kmacdonald-stsci: Changed spelling.

jwst/ramp_fitting/ramp_fit_step.py
kmacdonald-stsci: Removed check.

kmacdonald-stsci:
This PR is deprecated.

PR #5985, opened 2021-04-23
Switch to absolute imports in tests

mcara:
Use absolute import for unit tests. + an attempt to fix issues related to pytest/tox/c-extensions interaction.

Follow-on to #6001.
Addresses #5989 / [JP-2059](https://jira.stsci.edu/browse/JP-2059)

### DIFF
--- a/jwst/datamodels/multislit.py
+++ b/jwst/datamodels/multislit.py
@@ -13,7 +13,7 @@ class MultiSlitModel(JwstDataModel):
     This model has a special member `slits` that can be used to
     deal with an entire slit at a time.  It behaves like a list::
 
-       >>> from jwst.datamodels.slit import SlitModel
+       >>> from jwst.datamodels import SlitModel
        >>> multislit_model = MultiSlitModel()
        >>> multislit_model.slits.append(SlitModel())
        >>> multislit_model[0]
--- a/jwst/datamodels/multislit.py
+++ b/jwst/datamodels/multislit.py
@@ -13,7 +13,7 @@ class MultiSlitModel(JwstDataModel):
     This model has a special member `slits` that can be used to
     deal with an entire slit at a time.  It behaves like a list::
 
-       >>> from .slit import SlitModel
+       >>> from jwst.datamodels.slit import SlitModel
        >>> multislit_model = MultiSlitModel()
        >>> multislit_model.slits.append(SlitModel())
        >>> multislit_model[0]
--- a/jwst/datamodels/multispec.py
+++ b/jwst/datamodels/multispec.py
@@ -12,7 +12,7 @@ class MultiSpecModel(JwstDataModel):
     This model has a special member `spec` that can be used to
     deal with an entire spectrum at a time.  It behaves like a list::
 
-       >>> from . import SpecModel
+       >>> from jwst.datamodels.spec import SpecModel
        >>> multispec_model = MultiSpecModel()
        >>> multispec_model.spec.append(SpecModel())
        >>> multispec_model.spec[0] # doctest: +SKIP
--- a/jwst/datamodels/multispec.py
+++ b/jwst/datamodels/multispec.py
@@ -12,7 +12,7 @@ class MultiSpecModel(JwstDataModel):
     This model has a special member `spec` that can be used to
     deal with an entire spectrum at a time.  It behaves like a list::
 
-       >>> from jwst.datamodels.spec import SpecModel
+       >>> from jwst.datamodels import SpecModel
        >>> multispec_model = MultiSpecModel()
        >>> multispec_model.spec.append(SpecModel())
        >>> multispec_model.spec[0] # doctest: +SKIP
--- a/jwst/datamodels/specprofile.py
+++ b/jwst/datamodels/specprofile.py
@@ -11,7 +11,7 @@ class SpecProfileModel(ReferenceFileModel):
     This model has a special member `profile` that can be used to
     deal with an entire spectral profile at a time.  It behaves like a list::
 
-       >>> from . import SpecProfileSingleModel
+       >>> from jwst.datamodels.specprofile import SpecProfileSingleModel
        >>> specprofile_model = SpecProfileModel()
        >>> specprofile_model.profile.append(SpecProfileSingleModel())
        >>> specprofile_model.profile[0] # doctest: +SKIP
--- a/jwst/datamodels/specprofile.py
+++ b/jwst/datamodels/specprofile.py
@@ -11,7 +11,7 @@ class SpecProfileModel(ReferenceFileModel):
     This model has a special member `profile` that can be used to
     deal with an entire spectral profile at a time.  It behaves like a list::
 
-       >>> from jwst.datamodels.specprofile import SpecProfileSingleModel
+       >>> from jwst.datamodels import SpecProfileSingleModel
        >>> specprofile_model = SpecProfileModel()
        >>> specprofile_model.profile.append(SpecProfileSingleModel())
        >>> specprofile_model.profile[0] # doctest: +SKIP
--- a/jwst/datamodels/spectrace.py
+++ b/jwst/datamodels/spectrace.py
@@ -11,7 +11,7 @@ class SpecTraceModel(ReferenceFileModel):
     This model has a special member `trace` that can be used to
     deal with an entire spectral trace at a time.  It behaves like a list::
 
-       >>> from . import SpecTraceSingleModel
+       >>> from jwst.datamodels.spectrace import SpecTraceSingleModel
        >>> spectrace_model = SpecTraceModel()
        >>> spectrace_model.trace.append(SpecTraceSingleModel())
        >>> spectrace_model.trace[0] # doctest: +SKIP
--- a/jwst/datamodels/spectrace.py
+++ b/jwst/datamodels/spectrace.py
@@ -11,7 +11,7 @@ class SpecTraceModel(ReferenceFileModel):
     This model has a special member `trace` that can be used to
     deal with an entire spectral trace at a time.  It behaves like a list::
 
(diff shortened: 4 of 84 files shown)


mcara:
Related to comment https://github.com/spacetelescope/jwst/pull/5978#discussion_r619404694, I do not believe these tests run on installed package before, at least not since https://github.com/spacetelescope/jwst/pull/5737. Instead they run in the source directory. I think before that PR, the tests used to run in the _installed_ directory. Here is [an example dating back about 2 weeks] - before my PR #5978 - (https://github.com/spacetelescope/jwst/runs/2325199492):
tox-in-src1

Therefore, the workaround in #5978 of using `python setup.py build_ext --inplace` and running tests in the source directory did not result in change of how tests were run. [This does not mean that this is a good way of running tests]

jdavies-st (commented):
I think all the changes here to test modules to have absolute imports are good and needed.  👏 

Changes to `tox.ini` are probably not ideal, though I don't have a better solution at the moment.  Give me some time to think about this problem.

jdavies-st:
The coverage has gone down I believe because now under "Latest dependency versions" job, you're running many fewer tests.  2007 tests

https://github.com/spacetelescope/jwst/pull/5985/checks?check_run_id=2450271997#step:10:94

whereas currently on master that job is collecting 2486 tests.

https://github.com/spacetelescope/jwst/runs/2422915561?check_suite_focus=true#step:10:50

mcara:
@jdavies-st Yeah, but which one is better? Currently on the master you seem to be running coverage tests on source files themselves and yaml files and `rst` files and whatnot. Is this how you want them?

mcara:
Screenshot_20210427-154520_Edge

jdavies-st:
Yeah, the number of tests collected by `pytest` is way down - missing 400 tests.  The knock-on effect is that it has a coverage effect.

On `master`, the source files are getting run by `pytest-doctestplus` plugin for doctests in the docstrings, the yaml files by the `asdf` schema tester plugin, and rst by the `pytest-doctestplus` plugin to make sure docs with code examples work.

Perhaps you're not getting pytest config options correctly from `setup.cfg`?  Those are all configure there.

mcara:
> Yeah, the number of tests collected by `pytest` is way down - missing ~400 tests. The knock-on effect is that it has a coverage effect.
> 
> On `master`, the source files are getting run by `pytest-doctestplus` plugin for doctests in the docstrings, the yaml files by the `asdf` schema tester plugin, and rst by the `pytest-doctestplus` plugin to make sure docs with code examples work.
> 
> Perhaps you're not getting pytest config options correctly from `setup.cfg`? Those are all configure there.

That's strange since in https://github.com/spacetelescope/jwst/blob/40b7b0fdeaac00e486dbed09860804c67f912417/setup.cfg#L137-L151 apparently you are trying to exclude `docs` directory from being included and yet, on master, currently those `rst` files are being picked up: https://github.com/spacetelescope/jwst/runs/2422915561?check_suite_focus=true#step:10:52

mcara:
Also, WRT

> Yeah, the number of tests collected by `pytest` is way down - missing ~400 tests. The knock-on effect is that it has a coverage effect.

Shouldn't coverage test measure the coverage of the Python code and not include the `yaml` files?

mcara:
@jdavies-st I feel like that `conftest.py` (see https://github.com/spacetelescope/jwst/pull/5737) in the root directory is messing thing up and making it very difficult to set up tests correctly. It seems like it is changing directories in a strange way. Even if I provide absolute path to schemas - it still does not run `asdf` tests.

Would it be possible for you to re-design https://github.com/spacetelescope/jwst/pull/5737 and then I could rebase my PR against master?

mcara:
And I know that `tox.ini` is picking up `setup.cfg` settings because I can see the effects of changes in `[pytest]` section on tests.

jdavies-st:
Apparently the reason `tox` is picking up `setup.cfg` is because we have the `{toxinidir}/docs` part in there.  Without that, it would not be able to find it with `--pyargs`.

mcara:
> Apparently the reason `tox` is picking up `setup.cfg` is because we have the `{toxinidir}/docs` part in there. Without that, it would not be able to find it with `--pyargs`.

Not really. That is how you had it before and maybe that is how you have it now. However, my PR originally used `pytest -c {toxinidir}/setup.cfg` to specify path to configuration file. It worked.

jwst/datamodels/multislit.py
```diff
@@ -13,7 +13,7 @@ class MultiSlitModel(JwstDataModel):
     This model has a special member `slits` that can be used to
     deal with an entire slit at a time.  It behaves like a list::
 
-       >>> from .slit import SlitModel
+       >>> from jwst.datamodels.slit import SlitModel
```
jdavies-st, line 16: This would be better as

```
>>> from jwst.datamodels import SlitModel
```

as that's the recommended canonical import.

jwst/datamodels/multispec.py
```diff
@@ -12,7 +12,7 @@ class MultiSpecModel(JwstDataModel):
     This model has a special member `spec` that can be used to
     deal with an entire spectrum at a time.  It behaves like a list::
 
-       >>> from . import SpecModel
+       >>> from jwst.datamodels.spec import SpecModel
```
jdavies-st, line 15:
```
>>> from jwst.datamodels import SpecModel
```

jwst/datamodels/specprofile.py
```diff
@@ -11,7 +11,7 @@ class SpecProfileModel(ReferenceFileModel):
     This model has a special member `profile` that can be used to
     deal with an entire spectral profile at a time.  It behaves like a list::
 
-       >>> from . import SpecProfileSingleModel
+       >>> from jwst.datamodels.specprofile import SpecProfileSingleModel
```
jdavies-st, line 14:
```
>>> from jwst.datamodels import SpecProfileSingleModel
```

jwst/datamodels/spectrace.py
```diff
@@ -11,7 +11,7 @@ class SpecTraceModel(ReferenceFileModel):
     This model has a special member `trace` that can be used to
     deal with an entire spectral trace at a time.  It behaves like a list::
 
-       >>> from . import SpecTraceSingleModel
+       >>> from jwst.datamodels.spectrace import SpecTraceSingleModel
```
jdavies-st, line 14:
```
>>> from jwst.datamodels import SpecTraceSingleModel
```

jwst/datamodels/wavemap.py
```diff
@@ -11,7 +11,7 @@ class WaveMapModel(ReferenceFileModel):
     This model has a special member `map` that can be used to
     deal with an entire wavelength map at a time.  It behaves like a list::
 
-       >>> from . import WaveMapSingleModel
+       >>> from jwst.datamodels.wavemap import WaveMapSingleModel, WaveMapModel
```
jdavies-st, line 14:
```
>>> from jwst.datamodels. import WaveMapSingleModel, WaveMapModel
```

jwst/stpipe/tests/test_step.py
```diff
@@ -592,14 +584,14 @@ def test_crds_override():
 
 
 def test_omit_ref_file():
-    from .steps import OptionalRefTypeStep
+    from jwst.stpipe.tests.steps import OptionalRefTypeStep
```
jdavies-st, line 587: Since you moved the other imports to the top of the module, might be worth moving this one too.

jwst/stpipe/tests/test_step.py
```diff
@@ -592,14 +584,14 @@ def test_crds_override():
...
 
     step = OptionalRefTypeStep(override_to_be_ignored_ref_type="")
     step.process()
 
 
 def test_search_attr():
-    from .steps import SavePipeline
+    from jwst.stpipe.tests.steps import SavePipeline
```
jdavies-st, line 594: And this one.

tox.ini
```diff
@@ -58,6 +59,7 @@ install_command = python -m pip install --no-deps {packages}
 commands_pre =
     # Generate a requirements-min.txt file
     minimum_deps
+    python -m pip install --upgrade pip
```
jdavies-st, line 62: So in the CI workflow we have a step that upgrades `pip` automatically.  So this is redundant.  It could be argued that all this should be in `tox` instead of `ci.yml`, but then we should do it for all envs, not just a few.

tox.ini
```diff
@@ -69,10 +71,12 @@ commands =
 [testenv:regtests]
 description = run tests with --bigdata and --slow flags
 commands =
+    python -m pip install --upgrade pip
```
jdavies-st, line 74: See comment above.

tox.ini
```diff
@@ -69,10 +71,12 @@ commands =
...
     pytest --bigdata --slow --basetemp={homedir}/scratch {posargs}
 
 [testenv:twine]
 description = check that the package builds sdist/wheel and that twine uploads
+changedir = {toxinidir}
```
jdavies-st, line 79: This should be unnecessary as there's no `changedir` in `[testenv]` to override.  I.e. this is doing nothing here.

tox.ini
```diff
@@ -83,6 +87,7 @@ commands =
 
 [testenv:style]
 description = check code style, e.g. with flake8
+changedir = {toxinidir}
```
jdavies-st, line 90: Same here.

jdavies-st:
> my PR originally used `pytest -c {toxinidir}/setup.cfg` to specify path to configuration file. It worked.

That's a good way to do it.  👍

jwst/datamodels/multislit.py
mcara: of course! I must have missed some docstrings

jdavies-st (approved):
LGTM!  Thanks @mcara.

PR #5986, opened 2021-04-23
JP-2054 Add CFG file deprecation notice

### LINKED ISSUE #5979: Document deprecation of cfg files as default
_Issue [JP-2054](https://jira.stsci.edu/browse/JP-2054) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Incorporate the attached Word document content into the package documentation.

stscieisenhamer:
Fixes #5979 
Resolves [JP-2054](https://jira.stsci.edu/browse/JP-2054)

**Description**

This PR addresses a request from INS to add the deprecation notice that has been mailed around
to the internal JWST user groups.

Checklist
- [x] Tests
- [x] Documentation
- [ ] ~Change log~No code/api changes so no change log required
- [x] Milestone
- [x] Label(s)


docs/jwst/cfg_deprecation.rst
```diff
@@ -0,0 +1,85 @@
+.. _cfg_usage_deprecation_notice:
+
+CFG Usage Deprecation Notice
+============================
+
+As of March 18, a significant change to how JWST pipelines operate was completed
```
hbushouse, line 6: For the sake of future-proofing this, you should add the year to the date.

docs/jwst/introduction.rst
```diff
@@ -111,7 +121,7 @@ file from CRDS when running a pipeline:
 
     $ strun calwebb_detector1 jw00017001001_01101_00001_nrca1_uncal.fits
           --steps.dark_current.override_dark='my_dark.fits'
-    $ strun calwebb_detector1.cfg jw00017001001_01101_00001_nrca1_uncal.fits
+    $ strun calwebb_detector1.asdf jw00017001001_01101_00001_nrca1_uncal.fits
```
hbushouse, line 124: Do we even want to include the use of param ref files in examples of command line usage? Given that under normal circumstances the param files are pulled from CRDS and hence the users won't or don't need to download them locally and reference them on the command line, why should we even introduce new users to that way of running things right off the bat? I would think that the first example(s) for new users of how to run from the command line should just use the pipeline alias(s), e.g. "strun calwebb_detector1 blee.fits".  Later, after the initial introduction of this way of running things, could we also show how to run things by referencing a param file directly, if that's needed by a user in order to customize these (e.g. non-standard skipping of steps).

docs/jwst/cfg_deprecation.rst
stscieisenhamer: ✅

docs/jwst/introduction.rst
stscieisenhamer: Agree with the sentiment. This will take a bit of rework but a useful one. Currently there is no one specific section where all the parameter file stuff falls under. Right now, it sort of spread between introduction, the code-specific step documentation, and the format explanations. So, in intro, will leave out the running with bit, but have a pointer to a new "parameter files and how to use them" page with implications that this is "advanced usage".

jdavies-st (approved):

docs/jwst/introduction.rst
stscieisenhamer: I'm about to push a commit that goes a bit in the desired direction. However, I haven't completely re-written things, because I'm not sure where we are at on the API rewrite using the simplified interface.  I've tried to get enough in to at least clarify how best to do things as they stand.

hbushouse (approved):
LGTM

PR #5997, opened 2021-04-29
JP-1611: Calculate photometric errors using resampled total-error array

### LINKED ISSUE #5182: Error arrays in source catalog
_Issue [JP-1611](https://jira.stsci.edu/browse/JP-1611) was created on JIRA by [Steven Crawford](https://jira.stsci.edu/secure/ViewProfile.jspa?name=scrawford):_

While testing data that was created with Mirage on the Jupyterhub platforms, the linked to data set was processed with the default configurations for the pipeline.   

The source catalog step successfully executed and produced a catalog, but all of the error columns are filled with a values of nan.  

An example file is attached along with links to the data and notebook used.

larrybradley:
This PR can be merged despite the resampled total-error array still being work-in-progress.  The `model.err` arrays are currently all zeros, so the photometry errors in the source catalog will be zeros until `model.err` is populated with the resampled total-error array.

Fixes #5182 / [JP-1611](https://jira.stsci.edu/browse/JP-1611)

### DIFF
--- a/docs/jwst/source_catalog/main.rst
+++ b/docs/jwst/source_catalog/main.rst
@@ -53,6 +53,10 @@ semimajor and semiminor axis lengths, orientation of the major axis,
 and sky coordinates at corners of the minimal bounding box enclosing
 the source.
 
+Photometric errors are calculated from the resampled total-error
+array contained in the ``ERR`` (``model.err``) array. Note that this
+total-error array includes source Poisson noise.
+
 Output Products
 ---------------
 
--- a/docs/jwst/source_catalog/main.rst
+++ b/docs/jwst/source_catalog/main.rst
@@ -53,12 +53,6 @@ semimajor and semiminor axis lengths, orientation of the major axis,
 and sky coordinates at corners of the minimal bounding box enclosing
 the source.
 
-.. Note::
-
-   Errors are only created when an image has an error extension.  Products
-   created from the resampling step currently do not have an error extension
-   and the error columns are currently filled with a value of NaN.
-
 Output Products
 ---------------
 
--- a/jwst/source_catalog/source_catalog.py
+++ b/jwst/source_catalog/source_catalog.py
@@ -433,29 +433,6 @@ def make_segment_img(data, threshold, npixels=5.0, kernel=None, mask=None,
     return segm
 
 
-def calc_total_error(model):
-    """
-    Calculate the total error array.
-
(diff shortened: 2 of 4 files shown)


docs/jwst/source_catalog/main.rst
```diff
@@ -53,12 +53,6 @@ semimajor and semiminor axis lengths, orientation of the major axis,
 and sky coordinates at corners of the minimal bounding box enclosing
 the source.
 
-.. Note::
```
hbushouse, line 56: I agree that this note can now be removed, given the updates to the code, but is there anywhere else in the docs where we at least mention that all of the photometric errors listed above are computed based on the ERR (model.error) array of the input data, as well as mentioning that the ERR array contains all noise/error sources (e.g. Poisson, read noise, flat-field, ...) computed upstream in the pipeline? If not, I think it would be good to add just a sentence or two about this somewhere.

jwst/source_catalog/source_catalog.py
```diff
@@ -656,15 +631,6 @@ def set_segment_properties(self):
 
         return
 
-    @lazyproperty
```
hbushouse, line 659: Yay, no more columns full of NaN!

hbushouse (commented):
Looks good overall. Just one comment regarding docs.

jwst/source_catalog/source_catalog.py
larrybradley: 👍

docs/jwst/source_catalog/main.rst
larrybradley: I added a note in the docs about the source of the photometric errors.  I stated that this "total" error array *includes* source Poisson noise.  "total" here means combining all of the propagated error planes, but I'm don't think it includes the flat-field (@jdavies-st ?).  In any case, "total" could change in the future if more error arrays are propagated.  I think the important thing to point out here is that the source Poisson noise is included, which may be unexpected for those familiar with HST data.

hbushouse: I think our ERR does in fact also include flat-field variance, but you're right that it's probably not necessary to list every kind of noise source that is included. As you suggest, the important one is simply the Poisson term, which is different than HST. So wording like "... total error array, which includes Possion noise, ..." should be fine.

hbushouse (approved):

PR #5998, opened 2021-04-30
Replace nn_abmag with nn_label column

larrybradley:
This is a followup to #5996.

To further remove the dependency of other catalog values on the `is_star` value (and to remove redundant columns), this PR replaces the `nn_abmag` column (which represented the aperture-corrected total AB mag if the nearest-neighbor is star, otherwise its isophotal (segment-based) AB mag was used), with a simple `nn_label` column which gives the source label number of nearest-neighbor.  The user can then use whatever magnitude they want for that nearest-neighbor source.


hbushouse (approved):
LGTM

PR #6001, opened 2021-04-30
Move CRDS_CONTEXT reporter to pytest plugin

### LINKED ISSUE #5989: Fix CI  testing so it is testing against installed dependencies
_Issue [JP-2059](https://jira.stsci.edu/browse/JP-2059) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

Currently the `--pyargs` job is not testing against installed dependencies. Fix it.

jdavies-st:
Closes #5989
Resolves [JP-2059](https://jira.stsci.edu/browse/JP-2059)

**Description**

This PR fixes the problem of `--pyargs` not testing against the installed version of the package.

 - Move the top-level `conftest.py` to a pytest plugin.  This was the main thing that was breaking `--pyargs`.  An interaction between that and our layout not being a `src`-layout.
- Make the other jobs use `usedevelop = true` in tox.ini

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)


mcara:
This looks good. However, in relation to https://github.com/spacetelescope/jwst/pull/5985#issuecomment-828721566, I must note that coverage tests in this PR run on source directory and not on the installed package.

For example, here is coverage test:

test-coverage

and here is the `--pyargs` test:

test-pyargs

You can see that there are about 447 different tests.

mcara (approved):
LGTM

jdavies-st:
Right, this basically just reverts to how things worked before I broke it - i.e. run tests on source code with coverage, and then have a separate job that runs everything against installed, no coverage.

One of the limitations is the `asdf` schema tester plugin, which doesn't work with `--pyargs`, unfortunately.

mcara:
> One of the limitations is the `asdf` schema tester plugin, which doesn't work with `--pyargs`, unfortunately.

Yes, I've noticed that too.

mcara:
> One of the limitations is the `asdf` schema tester plugin, which doesn't work with `--pyargs`, unfortunately.

Yes, I've noticed that too, CC: @eslavich

jdavies-st:
I think the absolute import changes in #5985 are totally worth doing though.  And good.  Maybe rebase against `master` once this is merged, and take out the changes to `setup.cfg` and `tox.ini`?

mcara:
yes

eslavich:
> One of the limitations is the `asdf` schema tester plugin, which doesn't work with `--pyargs`, unfortunately.

One problem is that the asdf plugin is enabled in setup.cfg, which isn't available in TOXENV=pyargs because the working directory isn't the repo root.  I don't know how to resolve this in a repo layout without a "src" directory.

PR #6005, opened 2021-05-04
JP-1300:TSGRIM and WFSS improvements

### LINKED ISSUE #4572: Refactor TSGRISM WCS transforms
Issue [JP-1300](https://jira.stsci.edu/browse/JP-1300) was created by Howard Bushouse:

Refactor the WCS transforms used for NRC_TSGRISM exposures.

nden:
Closes #4572 

Resolves [JP-1300](https://jira.stsci.edu/browse/JP-1300)

**Description**

- Replace FITS WCS keywords in TSGRISM mode with RA/DEC_REF
- Add Spectral system info to slits in WFSS mode

Checklist
- [x] Tests
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/assign_wcs/nircam.py
+++ b/jwst/assign_wcs/nircam.py
@@ -240,10 +240,10 @@ def tsgrism(input_model, reference_files):
     ycenter = Const1D(yc)
     ycenter.inverse = Const1D(yc)
 
-    setra = Const1D(input_model.meta.wcsinfo.crval1)
-    setra.inverse = Const1D(input_model.meta.wcsinfo.crval1)
-    setdec = Const1D(input_model.meta.wcsinfo.crval2)
-    setdec.inverse = Const1D(input_model.meta.wcsinfo.crval2)
+    setra = Const1D(input_model.meta.wcsinfo.ra_ref)
+    setra.inverse = Const1D(input_model.meta.wcsinfo.ra_ref)
+    setdec = Const1D(input_model.meta.wcsinfo.dec_ref)
+    setdec.inverse = Const1D(input_model.meta.wcsinfo.dec_ref)
 
     # x, y, order in goes to transform to full array location and order
     # get the shift to full frame coordinates
--- a/jwst/extract_2d/grisms.py
+++ b/jwst/extract_2d/grisms.py
@@ -209,7 +209,7 @@ def extract_tso_object(input_model,
         log.info("Spectral trace extents: (xmin: {}, ymin: {}), "
                  "(xmax: {}, ymax: {})".format(xmin, ymin, xmax, ymax))
         log.info("Extraction limits: (xmin: {}, ymin: {}), "
-                 "(xmax: {}, ymax: {})".format(xmin_ext, ymin, xmax, ymax))
+                 "(xmax: {}, ymax: {})".format(xmin_ext, ymin, xmax_ext, ymax))
 
         # Cut out the subarray from the input data arrays
         ext_data = input_model.data[..., ymin: ymax + 1, xmin_ext:xmax_ext + 1].copy()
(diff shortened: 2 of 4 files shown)


jwst/assign_wcs/nircam.py
```diff
@@ -240,10 +240,10 @@ def tsgrism(input_model, reference_files):
     ycenter = Const1D(yc)
     ycenter.inverse = Const1D(yc)
 
-    setra = Const1D(input_model.meta.wcsinfo.crval1)
```
hbushouse, line 243: Given that we supposedly removed FITS WCS keywords from spectroscopic products quite a while ago already, how was this working? `crval1` and `crval2` should not have been in the meta info.

jwst/extract_2d/grisms.py
```diff
@@ -208,8 +206,10 @@ def extract_tso_object(input_model,
         ymax = int(ymax)
 
         log.info("WCS made explicit for order: {}".format(order))
-        log.info("Spectral trace extents: (xmin:{}, ymin:{}), (xmax:{}, ymax:{})".format(xmin, ymin, xmax, ymax))
-        log.info("Extraction limits: (xmin:{}, ymin:{}), (xmax:{}, ymax:{})".format(xmin_ext, ymin, xmax_ext, ymax))
+        log.info("Spectral trace extents: (xmin: {}, ymin: {}), "
+                 "(xmax: {}, ymax: {})".format(xmin, ymin, xmax, ymax))
+        log.info("Extraction limits: (xmin: {}, ymin: {}), "
+                 "(xmax: {}, ymax: {})".format(xmin_ext, ymin, xmax, ymax))
```
hbushouse, line 212: I realize this is the same code as before, just reformatted, but why does this message use `xmin_ext`, while all the other items being printed are just plain `ymin`, `xmax`, `ymax`?

jwst/extract_2d/grisms.py
```diff
@@ -467,9 +465,11 @@ def extract_grism_objects(input_model,
                 new_slit.meta.wcsinfo.spectral_order = order
                 new_slit.meta.wcsinfo.dispersion_direction = \
                     input_model.meta.wcsinfo.dispersion_direction
+                new_slit.meta.wcsinfo.specsys = input_model.meta.wcsinfo.specsys
+                new_slit.meta.coordinates = input_model.meta.coordinates
                 new_slit.meta.wcs = subwcs
                 if compute_wavelength:
-                    log.debug("Computing wavelengths (this takes a while ...)")
+                    log.debug("Computing wavelengths")
```
hbushouse, line 472: So, it doesn't take a while anymore, or you just don't want to publicize the fact that it does?  ;-)

hbushouse (commented):
Looks good. Just a few comments and minor questions.

jwst/assign_wcs/nircam.py
nden: Good question. I think this is a subtle and very nasty bug due to the way modeling and datamodels work.
`input_model.meta.wcsinfo.crval1` returns `None` because crval1 is not set but is in the schema.
`modeling.Const1D` has a default parameter value (amplitude=1) which is set if `None` is passed.
The effect is that `setra` and `setdec` are set to 1 and the step works without crashing.

jwst/extract_2d/grisms.py
nden: The team has decided they want to cover the entire detector in the `x` direction, that's why it's hard coded to `xmin_ext=0`.

jwst/extract_2d/grisms.py
nden: The transforms were reworked a while back to evaluate on arrays instead of individual pixels but I missed this comment a the time.

jwst/extract_2d/grisms.py
hbushouse: I see that in the old code the extraction limits (line 212) arguments include `xmin_ext` and `xmax_ext`, while the new code uses `xmin_ext` and `xmax`. Should that instead be `xmax_ext`?

hbushouse:
RTD build issue unrelated to this PR. Fixed in #6017 .

PR #6027, opened 2021-05-14
small fix for type of maximum_cores in jump detection

cshanahan1:
Jump detection wants a string to set the max number of cores used ('half', 'all', etc...). The default is 'none' but the code was checking for None type instead of the string.

### DIFF
--- a/jwst/jump/jump.py
+++ b/jwst/jump/jump.py
@@ -32,11 +32,12 @@ def detect_jumps(input_model, gain_model, readnoise_model,
     image.  Also, a 2-dimensional read noise array with appropriate values for
     each pixel is passed to the detection methods.
     """
-    if max_cores is None:
+    if max_cores is 'none' or None:
         numslices = 1
+        og.info("Using 1 core for jump detection ")
+
     else:
         num_cores = multiprocessing.cpu_count()
-        log.info("Found %d possible cores to use for jump detection " % num_cores)
         if max_cores == 'quarter':
             numslices = num_cores // 4 or 1
         elif max_cores == 'half':
@@ -45,6 +46,7 @@ def detect_jumps(input_model, gain_model, readnoise_model,
             numslices = num_cores
         else:
             numslices = 1
+        log.info(f"Found {num_cores} possible cores to use for jump detection, using {numslices} ")
 
     # Load the data arrays that we need from the input model
     output_model = input_model.copy()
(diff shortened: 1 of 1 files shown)


dmggh (commented):
Either way works - in the 'else' block, numslices would be set to 1 at the bottom anyway.

cshanahan1:
Yeah that's true, I just noticed it was logging 'found N cores for jump detection' which was a bit confusing when I was first looking at this.

jdavies-st:
We should have never allowed the string `"none"` to be a legit value of `maximum_cores`, as it makes no sense.  "none" cannot be a maximum value.  The default used internally used to be `None`, i.e not set, as is normal for python kwargs.

https://github.com/spacetelescope/jwst/commit/0a3a0696911f072c0ba90663147d648874e9f885

If you wanted to turn on multiprocessing, you could give it a maximum_cores value, and it would use it.

It was changed about 11 months ago (see https://github.com/spacetelescope/jwst/pull/5041), and clearly not done correctly.  And I for what reason?  Confusion?

So this PR fixes that implementation, but do we really want to be allowing `"none"` as a parameter value?  Or has that boat sailed?  @hbushouse

hbushouse (approved):
I agree with previous comments that 1) seeing the message about "Found N cores ..." often catches me off guard too, because I immediately think that's how many cores it has decided to use, even when I'm running with the no multiprocessing, and 2) this doesn't really end up changing the functionality at all (i.e. you still end up with `numslices=1`, but at least it has the benefit of bypassing the log message and hence avoids the potential confusion.

jwst/jump/jump.py
```diff
@@ -31,7 +31,7 @@ def detect_jumps (input_model, gain_model, readnoise_model,
     image.  Also, a 2-dimensional read noise array with appropriate values for
     each pixel is passed to the detection methods.
     """
-    if max_cores is None:
+    if max_cores is 'none' or None:
         numslices = 1
     else:
         num_cores = multiprocessing.cpu_count()
```
hbushouse, line 37: Maybe it would be more useful and less confusing to the user to move the log message (in the next line) to the end of the block where numslices is known and then change to message to include both available cores and actual number in use, e.g. "Found <n> cores; using <m>".

cshanahan1: I agree, I wanted to change this as well it threw me off. It does eventually say 'Creating N processes for jump detection', I'll combine those into a single message.

hbushouse (approved):

PR #6033, opened 2021-05-15
JP-1944: resample cleanup

jdavies-st:
Addresses #5798 / [JP-1944](https://jira.stsci.edu/browse/JP-1944)

**Description**

This PR should be uncontroversial.  It doesn't change anything user-facing.  It does some necessary cleanup of the resample code, without changing any of the core functionality nor test results.  Highlights:

 - Make calling of `ResampleData` and `ResampleSpecData` both return the resampled datamodel(s).  Before only one returned anything.
 - Remove `single` as kwarg from `GWCSDrizzle` class init.  It doesn't do anything there.
 - Always apply the DQ array, even if weighting is not specified (no effect now, as we always apply weight).  But we may want to not apply any weights in stage2 resampling.  Weighting has no meaning with a single input, but currently we do weight, which may have bad floating point effects, i.e multiplying by a small number and then dividing by that same number.  Regardless if we weight or not, bad pixels should always be excluded.
 - Importantly, adjust tolerances on the regression tests for `i2d` and `s2d` products so that they pass on jwcalibdev and on our development machines (MacOS laptops).  This is necessary right now because `drizzle` does all its computations in single precision, float32.  This very much speeds development to see a test passing.

More PRs to come built on top of this one.

Checklist
- [x] Tests
- [ ] ~Documentation~
- [ ] ~Change log~
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -186,11 +186,11 @@ class OutlierDetection:
         if pars['resample_data']:
             # Start by creating resampled/mosaic images for
             # each group of exposures
-            sdriz = resample.ResampleData(self.input_models, single=True,
+            resamp = resample.ResampleData(self.input_models, single=True,
                                           blendheaders=False, **pars)
-            drizzled_models = sdriz.do_drizzle()
-            for model in drizzled_models:
-                if save_intermediate_results:
+            drizzled_models = resamp.do_drizzle()
+            if save_intermediate_results:
+                for model in drizzled_models:
                     log.info("Writing out resampled exposures...")
                     model_output_path = self.make_output_path(
                         basepath=model.meta.filename,
@@ -206,8 +206,7 @@ class OutlierDetection:
                     good_bits=pars['good_bits'])
 
         # Initialize intermediate products used in the outlier detection
-        median_model = datamodels.ImageModel(
-            init=drizzled_models[0].data.shape)
+        median_model = datamodels.ImageModel(drizzled_models[0].data.shape)
         median_model.update(drizzled_models[0])
         median_model.meta.wcs = drizzled_models[0].meta.wcs
 
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -188,8 +188,7 @@ class OutlierDetection:
             # each group of exposures
             sdriz = resample.ResampleData(self.input_models, single=True,
                                           blendheaders=False, **pars)
-            sdriz.do_drizzle()
-            drizzled_models = sdriz.output_models
+            drizzled_models = sdriz.do_drizzle()
             for model in drizzled_models:
                 if save_intermediate_results:
                     log.info("Writing out resampled exposures...")
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -187,7 +187,7 @@ class OutlierDetection:
             # Start by creating resampled/mosaic images for
             # each group of exposures
             resamp = resample.ResampleData(self.input_models, single=True,
-                                          blendheaders=False, **pars)
+                                           blendheaders=False, **pars)
             drizzled_models = resamp.do_drizzle()
             if save_intermediate_results:
                 for model in drizzled_models:
--- a/jwst/regtest/test_fgs_image2.py
+++ b/jwst/regtest/test_fgs_image2.py
@@ -25,5 +25,10 @@ def test_fgs_image2(run_fgs_image2, rtdata_module, fitsdiff_default_kwargs, suff
 
     rtdata.get_truth(f"truth/test_fgs_image2/{output}")
 
+    # Adjust tolerance for machine precision with float32 drizzle code
+    if suffix == "i2d":
+        fitsdiff_default_kwargs["rtol"] = 3e-3
+        fitsdiff_default_kwargs["atol"] = 2e-2
+
     diff = FITSDiff(rtdata.output, rtdata.truth, **fitsdiff_default_kwargs)
     assert diff.identical, diff.report()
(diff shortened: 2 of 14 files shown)


jwst/regtest/test_fgs_image2.py
```diff
@@ -25,5 +25,10 @@ def test_fgs_image2(run_fgs_image2, rtdata_module, fitsdiff_default_kwargs, suff
 
     rtdata.get_truth(f"truth/test_fgs_image2/{output}")
 
+    # Adjust tolerance for machine precision with float32 drizzle code
+    if suffix == "i2d":
+        fitsdiff_default_kwargs["rtol"] = 3e-3
```
hbushouse, line 30: Wow, the differences can be that big?  If I'm reading this right, that's 0.3%, which is a lot (IMHO)!

jwst/regtest/test_nirspec_fs_spec3.py
```diff
@@ -34,13 +34,12 @@ def test_nirspec_fs_spec3(run_pipeline, rtdata_module, fitsdiff_default_kwargs,
...
     rtdata.output = output
-
-    # Get the truth files
     rtdata.get_truth(f"truth/test_nirspec_fs_spec3/{output}")
 
-    # Set looser tolerance for resampled s2d comparison
+    # Adjust tolerance for machine precision with float32 drizzle code
     if suffix == "s2d":
-        fitsdiff_default_kwargs["atol"] = 1e-5
+        fitsdiff_default_kwargs["rtol"] = 1e-2
```
hbushouse, line 41: This is up to 1%, right? That's huge. I don't like the idea of the routine tests running day after day on jwcalibdev being able to hide that big of a difference. I think I would be more in favor of having a smaller tolerance and just have folks realize that it's normal to see differences between their system and jwcalibdev.

hbushouse (approved):
Mechanically this all looks fine. I'm just a little uneasy about the relatively large tolerances that are being set for some of the regtests.

Will change log entries be added later in subsequent PR's that build on this one?

jwst/regtest/test_fgs_image2.py
jdavies-st: Yes.  This is why it is important to get the C drizzle code to use `float64` instead of `float32`.

jwst/regtest/test_nirspec_fs_spec3.py
jdavies-st: It is.  It's only a few pixels (7 in this case) that have that high of a tolerance.  But if the test is supposed to be effective at allowing development and catching bugs, this tolerance will catch bugs, but will allow development.

jdavies-st: And I'm pretty sure the affected pixels are edge effects in this case were NIRSpec FS has `NaN` mapping for some of the pixels in the bounding box for the curved spectral trace, i.e. pixels that do not come through the optics from the sky.  So this is sort-of-a thresholding effect on pixels of the trace that are either in-or-out.

jwst/outlier_detection/outlier_detection.py
```diff
@@ -186,12 +186,11 @@ def do_detection(self):
...
-                                          blendheaders=False, **pars)
-            sdriz.do_drizzle()
-            drizzled_models = sdriz.output_models
-            for model in drizzled_models:
-                if save_intermediate_results:
+            resamp = resample.ResampleData(self.input_models, single=True,
+                                           blendheaders=False, **pars)
+            drizzled_models = resamp.do_drizzle()
+            if save_intermediate_results:
+                for model in drizzled_models:
```
mcara, line 193: Oh, yes!

mcara (approved):
LGTM aside from an [optional] disagreement on removing explicit parameter name `init=...`.

jwst/resample/resample_utils.py
```diff
@@ -163,7 +163,7 @@ def build_driz_weight(model, weight_type=None, good_bits=None):
     else:
         warnings.warn("weight_type set to None.  Setting drizzle weight map to 1",
                       RuntimeWarning)
-        result = np.ones(model.data.shape, dtype=model.data.dtype)
+        result = np.ones(model.data.shape, dtype=model.data.dtype) * dqmask
```
mcara, line 166: Is `dqmask` here always `1` or `0` (or `True/False`)? Also, shouldn't it have the same shape as `model.data`? If so, wouldn't it be more clear to do `dqmask.astype(model.data.dtype.type)`? (or, of `dqmask` is not `1` or `0` - then (dqmask == 0).astype(model.data.dtype.type)`)?

mcara: I see on line `148` it is a boolean array => all good.

jwst/outlier_detection/outlier_detection.py
```diff
@@ -207,8 +206,7 @@ def do_detection(self):
                     good_bits=pars['good_bits'])
 
         # Initialize intermediate products used in the outlier detection
-        median_model = datamodels.ImageModel(
-            init=drizzled_models[0].data.shape)
+        median_model = datamodels.ImageModel(drizzled_models[0].data.shape)
```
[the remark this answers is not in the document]
jdavies-st: We routinely pass various things to `init`, i.e shapes, other datamodels, arrays, etc.  This is literally the only place in the code were we specify `init=Foo`.  `init` is not very descriptive or useful anyway.  So just making it consistent with the ~1000 initializations of datamodel classes.

jwst/regtest/test_nirspec_fs_spec3.py
jdavies-st: Good question!  I would prefer to actually fix `drizzle` to use `float64` internally.  Maybe you know somebody who is good at C programming?!  😉

mcara: It is strange that there is a loss of about 4-5 digits. This does not seem "right". Why is this happening?

PR #6038, opened 2021-05-18
JP-1950: Update ramp fit unit tests

### LINKED ISSUE #3879: GroupDQ index bug in ramp_fit.py
This is a simple issue where the same effect has been introduced twice in subsequent lines of code, so can be fixed by changing one of the lines.
The group indices of the GROUPDQ section need to be shifted by 1 in order to line up with the differences array.
Line 348 does this shift by 1 by extracting the GROUPDQ section with gdq_sect[1:,:,:].
https://github.com/spacetelescope/jwst/blob/ad87fa493f02bc5ea7c6c215a32db9bebd2d1fb5/jwst/ramp_fitting/ramp_fit.py#L348-L349

Line 355 does the same shift again by setting the indices as i_group-1 instead of i_group.
https://github.com/spacetelescope/jwst/blob/ad87fa493f02bc5ea7c6c215a32db9bebd2d1fb5/jwst/ramp_fitting/ramp_fit.py#L355

Only one of the above shifts is required.

Running version 0.13.7.

kmacdonald-stsci:
Closes #3879
Resolves [JP-1950](https://jira.stsci.edu/browse/JP-1950)

**Description**

This PR addresses unit tests for ramp fitting.   A bug fix was found in the computation of the median first differences of a pixel ramp.   Correcting this bug means revising unit and regression tests.  The bug fix effected two unit test cases.  These tests have been updated to account for the bug fix.  Work needs to be done to correct regression tests.

Bugfix in https://github.com/spacetelescope/stcal/pull/12

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


dmggh (approved):
Looks good to me. Does 'Work needs to be done to correct regression tests.' merely mean the truth files need to be updated ?

kmacdonald-stsci:
Yes.  I should have made that more clear.  The slope estimate change done in this PR effects many calculations.  The regression test truth files need to be adjusted to account for this change.

jdavies-st (commented):
Looks good.  Just need to update `setup.cfg` and `requirements-sdp.txt` to have the new version.  >= in the first, == in the 2nd.  Then all the tests will pass.

jdavies-st (approved):
Looks good.  Merge when tests pass.

Issue #6034, opened 2021-05-17
Create a tweakreg regression test


stscijgbot-jp:
_Issue [JP-2084](https://jira.stsci.edu/browse/JP-2084) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

Create a regression test running tweakreg with align_to_gaia enabled.

stscijgbot-jp:
_Comment by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse) on [JIRA](https://jira.stsci.edu/browse/JP-2084?focusedCommentId=538885#comment-538885):_

Fixed by #6040

PR #6040, opened 2021-05-19
JP-2084: Add a regression test for aligning to GAIA

### LINKED ISSUE #5455: Tweakreg fails when align_to_gaia is True
_Issue [JP-1773](https://jira.stsci.edu/browse/JP-1773) was created on JIRA by [Alicia Canipe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=acanipe):_

When running tweakreg for a set of 8 NIRCam LW images with align_to_gaia=True, the step fails with:
```java
LinAlgError: Singular matrix.
```
The full traceback is attached.

If I run the step on this dataset with align_to_gaia=False, the step runs to completion (full pipeline log attached in "tweakreg_passing.txt").

The data and association used are copied into the directory provided (██████████████████████████████████████████

mcara:
Closes https://github.com/spacetelescope/jwst/issues/6034
Resolves [JP-2084](https://jira.stsci.edu/browse/JP-2084)

This is a follow-on to changes in downstream packages that fix #5455 

Dataset is the LMC calibration field, simulated in `mirage`.

### DIFF
--- a/jwst/regtest/test_nircam_align_to_gaia.py
+++ b/jwst/regtest/test_nircam_align_to_gaia.py
@@ -1,5 +1,4 @@
 import pytest
-from astropy.io.fits.diff import FITSDiff
 from gwcs.wcstools import grid_from_bounding_box
 from numpy.testing import assert_allclose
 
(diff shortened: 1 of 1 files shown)


jdavies-st (approved):
This is perfect!  🚀 🥇 

In the future this fixture could also be used to test other aspects of this data.  Btw, the dataset is the LMC calibration field, simulated with `mirage`.

PR #6036, opened 2021-05-17
JP-1944: Resample error and variance arrays for imaging data

jdavies-st:
Addresses #5798 / [JP-1944](https://jira.stsci.edu/browse/JP-1944)

**Description**

This PR adds resampled variance and error arrays for imaging data run through `resample`.  The new error arrays in the resampled `i2d` files have the same names as in the `cal` files:

 - `ERR`
 - `VAR_RNOISE`
 - `VAR_POISSON`
 - `VAR_FLAT`

This PR also:

 - Refactors `ResampleData` to split out the processing of `single` resampling, which is used for `outlier_detection` from the case used in stage2 and stage3 `ResampleStep`, where 1 or more inputs get resampled to a single output.
 - Fixes a bug in the computation of total effective exposure time.
 - Removes unused `ResampleData.update_driz_outputs` method (it's never been used as far back as we've been on Github)
 - Marks the NIRSpec IFU regtest as `slow`, same as the other IFU `cube_build` tests.

Expected results of this PR in the regression tests are:
 - Lots of new ERR and VAR_* arrays.
 - New `EFFEXPTM` and `TEXPTIME` keyword values (tripled or quadrupled depending on the dataset).

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

PR #6041, opened 2021-05-19
JP-1944: Resample error and variance arrays for 2D spectral data

### LINKED ISSUE #5798: Generate ERR array in resampled data

jdavies-st:
Resolves #5798 / [JP-1944](https://jira.stsci.edu/browse/JP-1944)

**Description**

This PR adds resampled variance and error arrays for spectral data run through `resample_spec`.  The new error arrays in the resampled `s2d` files have the same names as in the `cal` files:

 - `ERR`
 - `VAR_RNOISE`
 - `VAR_POISSON`
 - `VAR_FLAT`

This PR also:

 - Makes the calling of drizzle from `ResampleStep` and `ResampleSpecStep` much more similar.
 - Use the refactored code already implemented in `ResampleData` by subclassing it for `ResampleSpecData`.  In a future PR, we will get rid of `ResampleSpecData` altogether and have spectral/imaging differences handled at the top level by the `Step` code.
 - No longer delete `bunit_err`.  We need it.

This branch contains all the changes from #6036, so that should be merged first and this rebased.  Reviewers feel free to hold off until the other is merged and this is rebased.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -103,8 +103,7 @@ class ResampleData:
         output_file = output_model.meta.filename
 
         log.info('Blending metadata for {}'.format(output_file))
-        blendmeta.blendmodels(output_model, inputs=self.input_models,
-                              output=output_file)
+        blendmeta.blendmodels(output_model, inputs=self.input_models, output=output_file)
 
     def resample_many_to_many(self):
         """Resample many inputs to many outputs where outputs have a common frame.
@@ -175,7 +174,13 @@ class ResampleData:
         output_model.err = np.sqrt(output_model.var_rnoise + output_model.var_poisson
                                    + output_model.var_flat)
 
-        self.update_fits_wcs(output_model)
+        # TODO: The following two methods and calls should be moved upstream to
+        # ResampleStep and ResampleSpecStep respectively
+        if isinstance(output_model, datamodels.ImageModel):
+            self.update_fits_wcs(output_model)
+        if isinstance(output_model, datamodels.SlitModel):
+            self.update_slit_metadata(output_model)
+
         self.update_exposure_times(output_model)
         self.output_models.append(output_model)
 
@@ -240,7 +245,7 @@ class ResampleData:
 
     @staticmethod
     def drizzle_arrays(insci, inwht, input_wcs, output_wcs, outsci, outwht, outcon,
-                       uniqid=1, xmin=0, xmax=0, ymin=0, ymax=0,
+                       uniqid=1, xmin=None, xmax=None, ymin=None, ymax=None,
                        pixfrac=1.0, kernel='square', fillval="INDEF"):
         """
         Low level routine for performing 'drizzle' operation on one image.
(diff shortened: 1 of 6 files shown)


hbushouse:
#6036 just got merged, so this PR can be rebased now.

jwst/resample/resample.py
```diff
@@ -377,6 +377,14 @@ def drizzle_arrays(insci, inwht, input_wcs, output_wcs, outsci, outwht, outcon,
         if outcon.ndim == 3:
             outcon = outcon[planeid]
 
+        if xmin is xmax is ymin is ymax is None:
+            bb = input_wcs.bounding_box
+            ((x1, x2), (y1, y2)) = bb
```
mcara, line 382: very minor thing: IMO no need to split this into two lines

mcara (approved):

jwst/resample/resample.py
```diff
@@ -377,6 +377,14 @@ def drizzle_arrays(insci, inwht, input_wcs, output_wcs, outsci, outwht, outcon,
         if outcon.ndim == 3:
             outcon = outcon[planeid]
 
+        if xmin is xmax is ymin is ymax is None:
```
hbushouse, line 380: I did not know this kind of syntax worked!  Cool.  I assume it treats the multiple arguments with a logical OR, i.e. if any one of them is None, it's True? Or is it an AND?

mcara: Not a very intuitive syntax, IMO but it's OK. I am actually not sure why it is necessary to check all of them. Wouldn't  just `xmin is None` be enough? What will happen if `xmin` is `None` and `xmax=100` (for example)?

jdavies-st: Yeah, checking one is probably fine.  I'm the only one calling all this internal code anyway.

This only works for `None` as it is a singleton, i.e. there is only of it.  There's no instance or copy.  So if `xmin is None` and `xmax is None`, they are both just the same thing, so `xmin is xmax`.

jwst/resample/resample.py
jdavies-st: Thanks.  Copy-and-pasted from somewhere else where `bb` was used subsequently 200 lines down.  😂

jdavies-st:
> Looks good. So the actual resampling of variance arrays is handled in the now generic `resample_many_to_many` function and hence doesn't appear explicitly as a change in this particular PR, right?

Exactly.  I did the work once.  A future PR after B7.8 will be to fully merge `ResampleData` and `ResampleSpecData` instead of subclassing.  No need to have two classes when one is sufficient.

PR #5958, opened 2021-04-14
JP-2045: Use Python 3.9 for SDP CI and regtests

jdavies-st:
Closes #5957 
Closes #5747
Resolves [JP-2045](https://jira.stsci.edu/browse/JP-2045)

**Description**

This PR does the following:
   - use Python 3.9 for SDP CI and nightly regtest builds
   - ~add Python 3.10-dev to the CI matrix~ Oops, the numpy folks accidentally prevented their code from being built under Python 3.10 because they pinned `wheel` version in pyproject.toml
   - remove pip caching in CI workflow (it masks build issues, see #5744 )
   - add `asdf` install so `crds` caching works in CI workflow

Checklist
- [x] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

PR #6042, opened 2021-05-19
Pin dependencies for Build 7.8

jdavies-st:
Replaces #5958 

**Description**

This PR pins direct and indirect dependencies via `requirements-sdp.txt` for Build 7.8 delivery.

It also turns off pip caching for both our CI and nightly regression tests.  This will allow us to discover when building of our dependency packages fails.  Normally `pip` uses the already-built binary wheel that it caches for convenience.  Good for speed, but not for catching problems.

Checklist
- [ ] ~Tests~
- [ ] ~Documentation~
- [ ] ~Change log~
- [x] Milestone
- [x] Label(s)


JenkinsfileRT_dev
```diff
@@ -80,8 +76,8 @@ bc0.conda_packages = [
 ]
 bc0.pip_reqs_files = ['requirements-dev.txt']
 bc0.build_cmds = [
-    "pip install -e .[test,ephem]",
-    "pip install pytest-xdist pytest-sugar",
```
hbushouse, line 84: Oh good. We need to cut back on sugar.

hbushouse (commented):
We have agreed with the other DMS subsystems to *not* use Python 3.9 for B7.8, so we need to pin to the latest 3.8.x patch somewhere for the B7.8 delivery. Is that set somewhere in here? If not, it needs to be.

jdavies-st:
We are already pinned to 3.8.x for nightly regtests here

https://github.com/spacetelescope/jwst/blob/f9ad731293df7137289d957b7963a86311f978c9/JenkinsfileRT#L46

jdavies-st:
Build from this branch:

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/243/

JenkinsfileRT_dev
jdavies-st: That plugin was supposed to make the output pretty and also show the failure tracebacks right away in the output instead of you having to wait until the tests are finished to see what failed. But it doesn't work with Jenkins log capturing for some reason.

hbushouse: So, no sugar tonight

hbushouse (approved):

Issue #5904, opened 2021-03-23
source_catalog:  aperture photometry needs to be updated for non-native pixel scales


larrybradley:
The `source_catalog` step needs to be updated to account for non-native detector pixel scales. The `pixel_scale_ratio` parameter was added to the `resample` step in https://github.com/spacetelescope/jwst/pull/5389.  Aperture photometry currently assumes native detector pixels scales (the aperture sizes in pixels are defined in the `apcorr` reference files).

This work is awaiting the `pixel_scale_ratio` value being output in the resampled `i2d` data.

PR #6044, opened 2021-05-19
Pixel scale ratio keyword

jdavies-st:
Address #5904

**Description**

This PR adds `meta.photometry.pixel_scale_ratio` which gets propagated from `ResampleStep`.  This is needed if the non-native pixel scale is used in the resampled products so that the photometry step can use the aperture reference files which have aperture sizes in native pixels.

This is needed by `source_catalog`.

- [x] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)


larrybradley (approved):
Thanks, @jdavies-st.  LGTM!

hbushouse (commented):
Functionally this is fine, but given that we already have some other drizzle-related keywords in the "Resample" section of the primary header (NDRIZ, TEXPTIME, RESWHT), I'm thinking it might be better to put this new keyword in that section, so they're all together? The datamodel path would be `model.meta.resample.pixel_scale_ratio`.

Any objections to that change?

larrybradley:
@hbushouse Yes, I think `model.meta.resample` is better.

Also, I recommend that `pixfrac` be added to `model.meta.resample` (with my scientist hat on).  But, I don't *need* that for source catalog.

hbushouse:
> @hbushouse Yes, I think `model.meta.resample` is better.
> 
> Also, I recommend that `pixfrac` be added to `model.meta.resample` (with my scientist hat on). But, I don't _need_ that for source catalog.

I was wondering about pixfrac. That is an available param in the resample step, right?  I agree that it would be good to add that one too.

jdavies-st:
I'll make these changes and file a JWST Keyword dictionary ticket to update.

With my scientist hat on, shouldn't `pixfrac` always be 1?  😄 

And just because HST stores everything and the kitchen sink in the header, we generally do not do that for JWST.  I.e. very few steps record anything in the header about how they were run.  The log records this.  Do we want to go down that path?

jdavies-st:
As a cautionary tale, I will point you to the old `level3prod` schema fragment that was removed when be deprecated `DrizProductModel` and the other one.

https://github.com/spacetelescope/jwst/blob/2defdd36b38c0d423de66a905cc9410687a4d5f3/jwst/datamodels/schemas/lev3_prod.schema.yaml

CHANGES.rst
```diff
@@ -194,6 +194,9 @@ resample
 
 - Propgagate error and variance arrays in ``ResampleStep`` for imaging data. [#6036]
 
+- Record ``pixel_scale_ratio`` from ``ResampleStep`` in header keyword PXSCLRT
+  or ``meta.photometry.pixel_scale_ratio``, for later use in aperture photometry. [#6044]
```
hbushouse, line 198: Need to update this now to `meta.resample.pixel_scale_ratio`

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -2330,6 +2330,16 @@ properties:
             type: string
             enum: [exptime, ivm]
             fits_keyword: RESWHT
+          pixfrac:
+            title: Drizzle parameter describing pixel shrinking
+            type: number
+            fits_keyword: PIXFRAC
+            fits_hdu: SCI
```
hbushouse, line 2337: Shouldn't these be in the primary header now, cuz they're aren't grouped with the photometry keywords anymore.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -2330,6 +2330,16 @@ properties:
             type: string
             enum: [exptime, ivm]
             fits_keyword: RESWHT
+          pixfrac:
```
hbushouse, line 2333: I don't see `pixfrac` getting populated anywhere in the `resample` code

jdavies-st: doh!  Will fix.

jdavies-st:
Another question, should these be in the PRIMARY header or the SCI header?  I suspect the PRIMARY since other resample params are written there and also all data arrays will be resampled.  I will change that too, as currently it is SCI.

hbushouse:
> Another question, should these be in the PRIMARY header or the SCI header? I suspect the PRIMARY since other resample params are written there and also all data arrays will be resampled. I will change that too, as currently it is SCI.

Yes, primary.

hbushouse (approved):
LGTM

jwst/datamodels/schemas/core.schema.yaml
jdavies-st: Done

CHANGES.rst
jdavies-st: Done.

PR #6045, opened 2021-05-19
Scale circular apertures for non-native pixel scales in resampled data

### LINKED ISSUE #5904: source_catalog:  aperture photometry needs to be updated for non-native pixel scales

larrybradley:
Closes #5904

This PR depends on the metadata added in #6044.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -213,6 +213,9 @@ source_catalog
 - The ``is_star`` column was replaced by a ``is_extended`` column with
   inverted boolean values. [#6018]
 
+- Circular aperture sizes now scale in the case of non-native pixel
+  scales in the resampled image. [#6045]
+
 srctype
 -------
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):
Looks good. Should the step docs also be updated to mention the use of this param?

Will wait to merge until #6044 has been merged, which provides the new keyword being used here.

larrybradley:
I'm not sure if additional docs are needed.  I think users would already expect this behavior.

PR #6046, opened 2021-05-20
JP-2044 Add rule Asn_MIRMRSBackground to treat background as science

### LINKED ISSUE #5955: MIRI MRS background spec3 associations should not have filter specification
_Issue [JP-2044](https://jira.stsci.edu/browse/JP-2044) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

MIRI MRS background observations should produce Level 3 associations so they are calibrated as science. Currently, the rule `Lv3SpecAux` handles this. This rule does not understand that MIRI MRS has no filter specification. As such, the `name` key is populated with a filter name, always `clear`. However, `Spec3Pipeline` reworks the name to remove this specification.

The issue is that downstream components, such as archive, use the association specification to search for files, which is now wrong.

Rework the rules such that MIRI MRS background observations are handled by a rule similar to MIRI MRS science, `Lv3MIRIMRS`, which creates the correct name.

stscieisenhamer:
Resolves #5955 [JP-2044](https://jira.stsci.edu/browse/JP-2044)

**Description**

This PR adds new rule `Asn_MIRMRSAux` to treat MRS backgrounds observations as science.

Checklist
- [ ] Tests
- [x] Documentation
- [ ] Change log
- [ ] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/associations/lib/rules_level3.py
+++ b/jwst/associations/lib/rules_level3.py
@@ -18,7 +18,7 @@ __all__ = [
     'Asn_Lv3Image',
     'Asn_Lv3SpecAux',
     'Asn_Lv3MIRMRS',
-    'Asn_Lv3MIRMRSAux',
+    'Asn_Lv3MIRMRSBackground',
     'Asn_Lv3NRSFSS',
     'Asn_Lv3NRSIFU',
     'Asn_Lv3NRSIFUBackground',
@@ -317,7 +317,7 @@ class Asn_Lv3MIRMRS(AsnMixin_Spectrum):
 
 
 @RegistryMarker.rule
-class Asn_Lv3MIRMRSAux(AsnMixin_AuxData, AsnMixin_Spectrum):
+class Asn_Lv3MIRMRSBackground(AsnMixin_AuxData, AsnMixin_Spectrum):
     """Level 3 MIRI MRS Association Auxiliary data
 
     Characteristics:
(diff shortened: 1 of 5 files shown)


hbushouse:
Needs a flake8 fix. And the run-time errors should be fixed by rebasing to include latest updates to ramp_fit unit tests.

jwst/associations/lib/rules_level3.py
```diff
@@ -313,18 +313,54 @@ def __init__(self, *args, **kwargs):
...
 
-        product_name = 'jw{}-{}_{}_{}'.format(
-            self.data['program'],
-            self.acid.id,
-            target,
-            instrument
-        )
-        return product_name.lower()
+@RegistryMarker.rule
+class Asn_Lv3MIRMRSAux(AsnMixin_AuxData, AsnMixin_Spectrum):
```
hbushouse, line 320: If this essentially creates the same kind of ASN's as `Lv3NRSIFUBackground` should we perhaps name it `Lv3MIRMRSBackground` for consistency?

hbushouse (commented):
Looks good. Just one question about the new class name.

stscieisenhamer:
Upstream/master is the one failing the test. I can fix truth if that is acceptable.

stscieisenhamer:
and magically the error went away...alrighty then.

hbushouse (approved):
LGTM

PR #6050, opened 2021-05-21
JP-2087: Updates for JWST_XYZ keyword changes

### LINKED ISSUE #6048: Update scripts to account for moving JWST_[XYZ] keywords to SCI header
_Issue [JP-2087](https://jira.stsci.edu/browse/JP-2087) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

JWSTKD-435 (and related tickets) ended up moving the JWST_[XYZ] and JWST_[DXDYDZ] keywords from the primary header to the SCI extension header. All scripts, such as set_velocity_aberration.py, that use direct FITS I/O to read/write these keywords need to be updated to access them in the SCI HDU, not the primary HDU.

hbushouse:
Closes #6048 

Resolves [JP-2087](https://jira.stsci.edu/browse/JP-2087)

**Description**

This PR makes updates for recent changes in the KWD and SDP to the JWST position/velocity keywords, in particular the movement of the JWST_[XYZ] and JWST_[DXYZ] keywords from the primary header to the SCI extension header. Scripts that use these keywords have been updated to account for that change. The core datamodel schema has also been updated to move those keywords to the SCI header, and to add the new OBSGEO-xyz keywords. The OBSGEO-xyz keywords give the JWST position and velocity in the geocentric frame (in units of meters and meters/sec), while the old JWST_xyz keywords are being changed to give the position and velocity in the barycentric frame (in units of km and km/s). Note that the datamodel attribute names for all of the keywords have been modified, to distinguish between the bary and geo frames.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [X] Milestone
- [X] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -100,6 +100,11 @@ datamodels
 - Added ``FULLP`` to SUBARRAY enum list in core, subarray,
   and keyword_psubarray schemas [#5947]
 
+- Moved JWST_[XYZ] and JWST_[DXDYDZ] keywords from primary to SCI extension
+  header and updated their comment fields to indicate they'll now be in the
+  barycentric frame. Also added the new OBSGEO[XYZ] keywords to the SCI
+  extension header, which are in the geocentric frame. [#6050]
+
 documentation
 -------------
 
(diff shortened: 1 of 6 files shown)


stscieisenhamer (approved):
LGTM

Need an issue to convert this utility to using `DataModel`.

hbushouse:
> Need an issue to convert this utility to using `DataModel`.

That would probably be good. I think all of the scripts applied during level-1b processing were originally written using fitsio because it wasn't obvious whether the jwst.datamodels package would be available in that environment (depending on how they got executed). But given that `set_telescope_pointing` already migrated to using datamodels quite a while ago, we should probably migrate the rest of them too.

PR #6051, opened 2021-05-21
Change segmentation map dtype to uint32

### LINKED ISSUE #5933: Segmentation maps should be written as integers
_Issue [JP-2029](https://jira.stsci.edu/browse/JP-2029) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

Currently segmentation maps from `source_catalog` are written out and saved as an `ImageModel`, storing the map itself in the `data` attribute.  But this is a float32 attribute, and so the map, which needs to be positive integers (uint32) gets converted to floats.

Either a new attribute for `ImageModel` that is uint32 should be added, or more likely, a new datamodel that has a `data` attribute defined in the schema as uint32 should be used.

Btw, how many sources are in an image?  Less than ~65k?  If so, we could go with uint16.  But maybe we should be conservative.

jdavies-st:
Resolves #5933 / [JP-2029](https://jira.stsci.edu/browse/JP-2029)

**Description**

This PR adds a new datamodel, `SegmentationMapModel` that has an `uint32` data array for storing the segmentation map output from `source_catalog`.  Before this was being stored in an `ImageModel` as a `float32`.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


hbushouse (commented):
Should also add a change log entry under datamodels to indicate that a new model has been created, i.e. `SegmentationMapModel`, because this is a bigger change than simply changing the dtype of one data array.

hbushouse (approved):
Looks good. Just need to resolve conflicts in change log (thanks to me for merging another PR in the meantime!)

PR #6055, opened 2021-05-21
Add migrate_data script to make older x1d files readable by later versions of jwst

eslavich:
**Description**

This PR adds a `migrate_data` script that will update the EXTRACT1D table to the new format introduced in #6014.  It needs documentation and a test of some kind, but I'd like to get feedback first in case I need to make significant changes to the script.

Checklist
- [ ] Tests

- [ ] Documentation

- [ ] Change log

- [ ] Milestone

- [ ] Label(s)


scripts/migrate_data
```diff
@@ -0,0 +1,153 @@
...
+
+    with fits.open(filename, memmap=False, mode=mode) as hdul:
+        model_type = hdul[0].header.get("DATAMODL")
+        jwst_version = hdul[0].header.get("CAL_VER")
+
+        if not (model_type and jwst_version):
+            print(f"Unable to migrate {filename}: DATAMODL and CAL_VER keywords are required")
+            return
+
+        match = re.match(r'^[0-9]+\.[0-9]+\.[0-9]+', jwst_version)
```
hbushouse, line 95: So does this checking of CAL_VER value mean that the script will only work for x1d's that were created using 1.1.0? We might need to allow it to go back a bit further than that, whenever the `SpecModel` had it's last change before this big one.

hbushouse (commented):
Looks reasonable overall. At first I thought the name of the script was too generic, but then I saw that was intentional, because migrating the current x1d files is just one change allowed for. Future expansion: what a concept.

scripts/migrate_data
eslavich: The requirement `"<= 1.1.0"` is interpreted as "any cal version that satisfies this requirement needs to be migrated".  So, it's going to try to migrate any version before 1.1.0 as well.  Maybe that's too many?  It looks like the change that removed NET and added SURF_BRIGHT was released in 0.13.2, so I can make that the lower bound.

hbushouse: Sounds good.

scripts/migrate_data
```diff
@@ -0,0 +1,153 @@
...
+def migrate_spec_model_1_1_0(hdul):
+    for hdu in hdul:
+        if hdu.name == "EXTRACT1D":
+            hdu.data = rebuild_spec_table(hdu.data)
+
+
+def migrate_multispec_model_1_1_0(hdul):
+    for hdu in hdul:
+        if hdu.name == "EXTRACT1D":
+            hdu.data = rebuild_spec_table(hdu.data)
```
jdavies-st, line 64: These functions are identical.  Is there a reason they are differently-named?  Can't the `_MIGRATE_METHODS` dict below point to a single function for the two different datamodel classes?

eslavich: Hmm... good point, I thought I was going to need different logic for MultiSpecModel but they sure are identical.

jdavies-st: I made the exact same discovery when writing the readers for `specutils` for these a year ago.  😅

jdavies-st (approved):
This looks great.  And the updates to `RegtestData` are a nice enhancement as well.  🚀

hbushouse:
Should be ready to merge, after fixing conflicts in change log.

PR #6072, opened 2021-05-25
AL-552: Tests and step code needed to change to accommodate for new parameter in `ramp_fit`.

kmacdonald-stsci:
Resolves AL-552

**Description**

To make ramp fitting code pipeline agnostic DQ flags are now passed as a parameter to `ramp_fit`.  The new parameter to `ramp_fit` required changes to the tests and step code.  All JWST ramp fit unit tests and regression tests pass with these changes.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -231,11 +232,18 @@ def process(self, input):
             else:
                 input_model.int_times = None
 
+            ramp_fit_dq_flags = {
+                "DO_NOT_USE": dqflags.pixel["DO_NOT_USE"],
+                "JUMP_DET": dqflags.pixel["JUMP_DET"],
+                "SATURATED": dqflags.pixel["SATURATED"],
+                "NO_GAIN_VALUE": dqflags.pixel["NO_GAIN_VALUE"],
+                "UNRELIABLE_SLOPE": dqflags.pixel["UNRELIABLE_SLOPE"],
+            }
```
jdavies-st, line 241: I think originally some of these DQ flags were from `jwst.datamodels.dqflags.group` dict.

https://github.com/spacetelescope/jwst/blob/f8d9122b78dd4c4f34df5ca0199e0857116ed0a3/jwst/ramp_fitting/ramp_fit.py#L33-L34

And some from the pixel dict:

https://github.com/spacetelescope/jwst/blob/f8d9122b78dd4c4f34df5ca0199e0857116ed0a3/jwst/ramp_fitting/ramp_fit.py#L1547

Though it seems the group and pixel dq flags are identical.  (above code snippets from release 1.1.0)

Is there a reason to redefine them here in the calling code instead of just shuffling the full dictionary on to `ramp_fit` and letting it use them as it used them before?

jwst/ramp_fitting/tests/test_ramp_fit.py
```diff
@@ -7,9 +7,17 @@
...
-DO_NOT_USE = dqflags.group['DO_NOT_USE']
-JUMP_DET = dqflags.group['JUMP_DET']
-SATURATED = dqflags.group['SATURATED']
+test_dq_flags = {
+    "DO_NOT_USE": dqflags.pixel["DO_NOT_USE"],
+    "JUMP_DET": dqflags.pixel["JUMP_DET"],
+    "SATURATED": dqflags.pixel["SATURATED"],
+    "NO_GAIN_VALUE": dqflags.pixel["NO_GAIN_VALUE"],
+    "UNRELIABLE_SLOPE": dqflags.pixel["UNRELIABLE_SLOPE"],
+}
```
jdavies-st, line 16: Again, no need for this.  Just use the existing dicts in `jwst.datamodels.dqflags.pixel` and `jwst.datamodels.dqflags.group`.

jwst/ramp_fitting/tests/test_ramp_fit_cases.py
```diff
@@ -18,6 +19,18 @@
 #  their SCI values (these are mostly for my reference).
 #
 
+test_dq_flags = {
+    "DO_NOT_USE": dqflags.pixel["DO_NOT_USE"],
+    "JUMP_DET": dqflags.pixel["JUMP_DET"],
+    "SATURATED": dqflags.pixel["SATURATED"],
+    "NO_GAIN_VALUE": dqflags.pixel["NO_GAIN_VALUE"],
+    "UNRELIABLE_SLOPE": dqflags.pixel["UNRELIABLE_SLOPE"],
+}
```
jdavies-st, line 28: Same here.

jwst/ramp_fitting/ramp_fit_step.py
kmacdonald-stsci: The `group` flags are defined from the `pixel` dictionary.  I simply created these test DQ flag dictionaries as a demonstration that the values could be defined to be anything, so long as they are consistent, as well as to show the required key words to pass to `ramp_fit`.

kmacdonald-stsci: I changed the call to `ramp_fit` to simply pass `dqflags.pixel`, rather than create a separate dictionary to pass.

jwst/ramp_fitting/tests/test_ramp_fit.py
kmacdonald-stsci: I changed the code to `test_dq_flags = dqflags.pixel`.

jwst/ramp_fitting/tests/test_ramp_fit_cases.py
kmacdonald-stsci: I changed the code to `test_dq_flags = dqflags.pixel`.

jdavies-st (approved):
This looks good.

This PR assumes that `dqflags.group` dict is a subset of `dqflags.pixel` and that those mnemonics are interchangeable.  Which they _currently_ are.  And that was the intention, but that is not actually formalized.

So I think this is OK.

Approving, though merging will have to wait until all the changes in `stcal` are merged and released.

Will file a separate issue/PR for formalize `dqflags.pixel` being a superset of `dqflags.group`.

jdavies-st:
FWIW, `dqflags.group` being a subset of `dqflags.pixel` was formalized in #4726

jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -162,10 +165,13 @@ def create_optional_results_model(opt_info):
         weights=weights,
         crmag=crmag)
 
+    opt_model.meta.filename = input_model.meta.filename
+    opt_model.update(input_model)  # ... and add all keys from input
```
hbushouse, line 169: I think that calling .update will include the copying of meta.filename from the input to the output model, so line 168 may not even be necessary. Also, do we really want to set the input and output file names to be the same? Shouldn't the opt_model get saved to a file name that at least has a different file type suffix? Or is that handled later?

hbushouse (commented):
Needs a change log entry (in CHANGES.rst)

jwst/ramp_fitting/ramp_fit_step.py
jdavies-st: That's done later in the Step code:

https://github.com/spacetelescope/jwst/blob/8a8f148d4ab6310282ac76162890d44b81122840/jwst/ramp_fitting/ramp_fit_step.py#L240-L243

jdavies-st:
Unit and regression tests pass pointing to `stcal 0.2.2`

https://plwishmaster.stsci.edu:8081/job/RT/job/jdavies-dev/262/

Merging.

PR #6086, opened 2021-05-27
JP-2102: Add new keywords from dictionary

### LINKED ISSUE #6071: Add new keywords to datamodels schemas for B7.8
_Issue [JP-2102](https://jira.stsci.edu/browse/JP-2102) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

Several new keywords were implemented in the KWD for B7.8 that have not yet been added to the datamodels schemas. The new keywords and their properties are listed below.

hbushouse:
Closes #6071 

Resolves [JP-2102](https://jira.stsci.edu/browse/JP-2102)

**Description**

This PR adds several new keywords to the core schema, in order to keep in synch with the JWST Keyword Dictionary at the time of the B7.8 delivery.

Checklist
- [ ] Tests
- [ ] ~~Documentation~~
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -47,6 +47,11 @@ properties:
         type: string
         fits_keyword: PRD_VER
         blend_table: True
+      oss_software_version:
+        title: Observatory Scheduling Software (OSS) version number
+        type: string
+        fits_keyword: OSS_VER
+        blend_table: True
       calibration_software_version:
         title: Calibration software version number
         type: string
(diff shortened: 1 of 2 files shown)


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1140,6 +1155,14 @@ properties:
             type: integer
             fits_keyword: NRIMDTPT
             blend_table: True
+          nod_type:
```
tapastro, line 1158: I assume you have the source for these lists, but I would've guessed there'd be a 2- and 5-POINT-NOD in addition to the 3-POINT-NOD listed here. Any idea what gives?

hbushouse: This keyword only applies to NIRSpec MOS exposures where the names of the nod pattern use the x-SHUTTER-SLITLET syntax (nods within a slitlet). The PATTTYPE keyword has values like 2-POINT-NOD, 3-POINT-NOD, etc., which is used for NIRSpec FS and other modes that are nodding within a physical slit (as opposed to a slitlet construct). Yes, the values shown here come from what's currently in the JWST Keyword Dictionary, which is based on all known values defined in PPS.

tapastro (approved):
LGTM

PR #6081, opened 2021-05-26
JP-2106: Update keyword schema enum lists

hbushouse:
Closes #6078 

Resolves [JP-2106](https://jira.stsci.edu/browse/JP-2106)

**Description**
Update the enum lists for a few keywords to bring them into synch with what's in the KWD for B7.8

Checklist
- [ ] Tests
- [ ] ~~Documentation~~
- [x] Change log
- [x] Milestone
- [x] Label(s)

PR #6088, opened 2021-05-28
Fix ENGQLPTG value populated by set_telescope_pointing

hbushouse:
This PR is a follow on to https://github.com/spacetelescope/jwst/pull/6081/ and JP-2106, which changed the allowed list of values for the ENGQLPTG keyword. The simple value "CALCULATED" is no longer allowed, so the set_telescope_pointing script is being updated to instead use the new allowed value of "CALCULATED_ORIGINAL".

Checklist
- [ ] Tests
- [ ] ~~Documentation~~
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,12 @@ general
 
 - Make CRDS context reporting pytest plugin disabled by default. [#6070]
 
+lib
+---
+
+- Updated set_telescope_pointing to populate ENGQLPTG keyword with new
+  allowed values [#6088]
+
 1.2.0 (2021-05-24)
 ==================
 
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -61,8 +67,6 @@ assign_wcs
 associations
 ------------
 
-- Asn_Lv2NRSLAMPSpectral: Allow msaspec only if msametfl is available. [#6085]
```
hbushouse, line 64: this is the second copy of this entry, so I removed it

tapastro (approved):
LGTM

PR #6090, opened 2021-06-01
JP-2068: Fix invalid ECSV test data headers

### LINKED ISSUE #6003: Table.read() error when reading ECSV using astropy dev
_Issue [JP-2068](https://jira.stsci.edu/browse/JP-2068) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

We are seeing a failure in our weekly cron job of unit tests using dev dependencies.  This one is because of some changes to the ECSV format in astropy dev:

<https://github.com/spacetelescope/jwst/runs/2489783314?check_suite_focus=true#step:9:51414>

The failing tests in the `jwst` repo are:

```
test_calc_deltas
test_from_models
```

failing with the same error in `Table.read()` when reading an ECSV file:

```
E               ValueError: datatype 'object' of column 'exposure' is not in allowed values ('bool', 'int8', 'int16', 'int32', 'int64', 'uint8', 'uint16', 'uint32', 'uint64', 'float16', 'float32', 'float64', 'float128', 'string')
```

Also, the test should be using a context manager when opening the ECSV file using `Table.read()`, hence all the open-files ERRORs, though clearly these 2 tests are contaminating others, somehow.

jdavies-st:
Resolves #6003 / [JP-2068](https://jira.stsci.edu/browse/JP-2068)

**Description**

This PR fixes 2 truth tables that have `dtype=object` in the ECSV headers.  That is apparently no longer allowed in astropy dev.  Changing it to `string` solves the problem.

The failing `Table.read()` call was also causing open file handle errors, which probably means that the internals of that function are not quite handling files correctly.  Will file an issue in `astropy` for both of these.

Also cleaned up reading of the FITS files with a context manager.  Apparently there is no `with` context manager for `Table.read`.

- [x] Tests
- [ ] ~Documentation~
- [ ] ~Change log~
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/lib/tests/data/calc_deltas_truth.ecsv
+++ b/jwst/lib/tests/data/calc_deltas_truth.ecsv
@@ -1,7 +1,7 @@
 # %ECSV 0.9
 # ---
 # datatype:
-# - {name: exposure, datatype: object}
+# - {name: exposure, datatype: string}
 # - {name: target.ra, unit: deg, datatype: float64}
 # - {name: target.dec, unit: deg, datatype: float64}
 # - {name: v1.ra, unit: deg, datatype: float64}
--- a/jwst/lib/tests/data/v1_calc_truth.ecsv
+++ b/jwst/lib/tests/data/v1_calc_truth.ecsv
@@ -1,7 +1,7 @@
 # %ECSV 0.9
 # ---
 # datatype:
-# - {name: source, datatype: object}
+# - {name: source, datatype: string}
 # - {name: obstime, datatype: string}
 # - {name: ra, datatype: float64}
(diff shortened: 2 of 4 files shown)


stscieisenhamer (approved):

jdavies-st:
Failures in unit tests

```
FAILED jwst/assign_wcs/tests/test_nircam.py::test_traverse_wfss_grisms - asse...
FAILED jwst/assign_wcs/tests/test_nircam.py::test_traverse_tso_grism - assert...
```

due to wonky CRDS reffile is unrelated.  Merging.

PR #6101, opened 2021-06-03
JP-2051: WFS combine align dithers and Update Refine

### LINKED ISSUE #5974: wfs_combine to create images with the same dither position
_Issue [JP-2051](https://jira.stsci.edu/browse/JP-2051) was created on JIRA by [Marcio Melendez Hernandez](https://jira.stsci.edu/secure/ViewProfile.jspa?name=melendez):_

The wfs_combine step of the WFS&C JWST pipeline combines a two-point dither set of exposures. However, the resulting combined image take the X,Y dither coordinate from the first dither in the pair of images.  This is a problem because for an observation with many dither exposures inside, the L3 combined image will alternate between dither positions. 

It is important that each of the dither combine images (when there are multiple exposures in a single observation) are all relative to the same dither location, base.  The example below, is from the Coarse Phasing template where it is clear that the dither position alternates in a A B, B, A, A, etc pattern and not A, B A, B which is what the wfs_combine is creating. 

 

Tar Tile Exp Dith Aperture Name             Target       RA         Dec        BaseX    BaseY    DithX    DithY    V2       V3       IdlX     IdlY   Level      Type      ExPar  DkPar   dDist          

  1    1   1    1 NRCA3_DHSPIL_SUB96        9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.000   +0.000  +52.191 -526.208   +0.000   +0.000 TARGET     SCIENCE        0      0  0.000 (base)

  1    1   1    2 NRCA3_DHSPIL_SUB96        9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.160   +0.000  +52.031 -526.207   +0.160   +0.000 DITHER     SCIENCE        0      0  0.160          

  1    1   2    1 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.160   +0.000  +52.031 -526.207   +0.160   +0.000 FILTER     SCIENCE        0      0  0.000          

  1    1   2    2 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.000   +0.000  +52.191 -526.208   +0.000   +0.000 DITHER     SCIENCE        0      0  0.160          

  1    1   3    1 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.000   +0.000  +52.191 -526.208   +0.000   +0.000 FILTER     SCIENCE        0      0  0.000          

  1    1   3    2 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.160   +0.000  +52.031 -526.207   +0.160   +0.000 DITHER     SCIENCE        0      0  0.160          

  1    1   4    1 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.160   +0.000  +52.031 -526.207   +0.160   +0.000 FILTER     SCIENCE        0      0  0.000          

  1    1   4    2 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.000   +0.000  +52.191 -526.208   +0.000   +0.000 DITHER     SCIENCE        0      0  0.160          

  1    1   5    1 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.000   +0.000  +52.191 -526.208   +0.000   +0.000 FILTER     SCIENCE        0      0  0.000          

  1    1   5    2 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.160   +0.000  +52.031 -526.207   +0.160   +0.000 DITHER     SCIENCE        0      0  0.160          

  1    1   6    1 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.160   +0.000  +52.031 -526.207   +0.160   +0.000 FILTER     SCIENCE        0      0  0.000          

  1    1   6    2 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.000   +0.000  +52.191 -526.208   +0.000   +0.000 DITHER     SCIENCE        0      0  0.160          

  1    1   7    1 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.000   +0.000  +52.191 -526.208   +0.000   +0.000 FILTER     SCIENCE        0      0  0.000          

  1    1   7    2 NRCA3_DHSPIL              9 TYC-9182 +102.91771  -69.53910   +0.000   +0.000   +0.160   +0.000  +52.031 -526.207   +0.160   +0.000 DITHER     SCIENCE        0      0  0.160       
## APT ID 1159 (Coarse Phasing) 

Example of simulations are under, including association files 

/ifs/jwst/tel/TeamPractices/OTE-22/output

mwregan2:
Closes #5974 

Resolves [JP-2051](https://jira.stsci.edu/browse/JP-2051)

This PR addresses JP-2051 that refers to the location of the combined image when WFS-COMBINE is run. The changes are to always shift the images in the positive x pixel direction. If the calculated offset is negative the first image will be shifted rather than the second image.
  While making the changes for JP-2051, I added test cases for the refine option. This showed that the baseline code's cross correlation algorithm was not correct. It was unable to find the correct offset. I updated the algorithm to be correct, added 27 new unit tests, and updated the output FITS header and logs to contain the calculated shifts.   

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/wfs_combine/tests/test_wfs_combine.py
+++ b/jwst/wfs_combine/tests/test_wfs_combine.py
@@ -5,6 +5,7 @@ from jwst.assign_wcs import AssignWcsStep
 from jwst.associations.asn_from_list import asn_from_list
 from jwst.wfs_combine import WfsCombineStep
 from jwst.wfs_combine import wfs_combine
+import astropy.modeling as modeling
 import numpy as np
 
 
@@ -26,11 +27,15 @@ def add_point_source(inarray, scale, xcen, ycen, sigx, sigy):
     for y in range(inarray.shape[0]):
         for x in range(inarray.shape[1]):
             outarray[y, x] = scale * gaussian2d(x, y, xcen, ycen, sigx, sigy)
+#            outarray[y, x] = modeling.functional_models.Gaussian2D(amplitude=scale,
+#                                                                   x_mean=xcen, y_mean=ycen,
+#                                                                   x_stddev=sigx, y_stddev=sigy)
     return outarray
 
 
 @pytest.fixture(scope="module")
-def wfs_association(tmp_path_factory, imsize=10):
+def wfs_association(tmp_path_factory):
+    imsize = 10
     tmp_path = tmp_path_factory.mktemp("wfs")
     im1 = datamodels.ImageModel((imsize, imsize))
     im1.meta.observation.exposure_number = 1
--- a/jwst/wfs_combine/tests/test_wfs_combine.py
+++ b/jwst/wfs_combine/tests/test_wfs_combine.py
@@ -14,7 +14,7 @@ SATURATED = datamodels.dqflags.pixel["SATURATED"]
 
 
 def gaussian(x, mu, sig):
-    return 1. /(np.sqrt(2 * np.pi) * sig) * np.exp(-np.power((x - mu) /sig, 2) /2)
+    return 1. / (np.sqrt(2 * np.pi) * sig) * np.exp(-np.power((x - mu) / sig, 2) / 2)
 
 
 def gaussian2d(x, y, mux, muy, sigx, sigy):
@@ -121,7 +121,7 @@ def test_create_combined(_jail, wfs_association,
 
 def test_shift_order_no_refine_no_flip(wfs_association):
     path_asn, path1, path2 = wfs_association
-    nircam_pixel_size = 0.031 /3600.
+    nircam_pixel_size = 0.031 / 3600.
     delta_pixel = 5
     with datamodels.open(path2) as im2:
         im2.meta.wcsinfo = {
--- a/jwst/wfs_combine/tests/test_wfs_combine.py
+++ b/jwst/wfs_combine/tests/test_wfs_combine.py
@@ -6,7 +6,6 @@ from jwst.associations.asn_from_list import asn_from_list
 from jwst.wfs_combine import WfsCombineStep
 from jwst.wfs_combine import wfs_combine
 import numpy as np
-from astropy.io import fits
 
 
 GOOD = datamodels.dqflags.pixel["GOOD"]
@@ -15,7 +14,7 @@ SATURATED = datamodels.dqflags.pixel["SATURATED"]
 
 
 def gaussian(x, mu, sig):
-    return 1./(np.sqrt(2 * np.pi) * sig) * np.exp(-np.power((x - mu)/sig, 2)/2)
+    return 1. /(np.sqrt(2 * np.pi) * sig) * np.exp(-np.power((x - mu) /sig, 2) /2)
 
 
 def gaussian2d(x, y, mux, muy, sigx, sigy):
@@ -122,7 +121,7 @@ def test_create_combined(_jail, wfs_association,
 
 def test_shift_order_no_refine_no_flip(wfs_association):
     path_asn, path1, path2 = wfs_association
-    nircam_pixel_size = 0.031/3600.
+    nircam_pixel_size = 0.031 /3600.
     delta_pixel = 5
     with datamodels.open(path2) as im2:
         im2.meta.wcsinfo = {
@@ -135,7 +134,7 @@ def test_shift_order_no_refine_no_flip(wfs_association):
         im2.save(path2)
     wfs = wfs_combine.DataSet(path1, path2, 'outfile.fits', do_refine=False, flip_dithers=False, psf_size=50,
                               blur_size=10, n_size=2)
-    output_model = wfs.do_all
+    wfs.do_all
     assert wfs.input_1.meta.observation.exposure_number == '1'
     assert wfs.input_2.meta.observation.exposure_number == '2'
     assert wfs.off_x == abs(delta_pixel)
@@ -160,7 +159,7 @@ def test_shift_order_no_refine_with_flip(wfs_association):
         im2.save(path2)
     wfs = wfs_combine.DataSet(path1, path2, 'outfile.fits', do_refine=False, flip_dithers=True, psf_size=50,
                               blur_size=10, n_size=2)
-    output_model = wfs.do_all
+    wfs.do_all
     wfs.input_1.close()
     wfs.input_2.close()
 #    assert wfs.input_1.meta.observation.exposure_number == '2'
@@ -223,7 +222,7 @@ def test_refine_no_error(wfs_association, xshift, yshift, xerror, yerror, flip_d
         im2.save(path2)
     wfs = wfs_combine.DataSet(path1, path2, 'outfile.fits', do_refine=True, flip_dithers=flip_dithers, psf_size=50,
                               blur_size=10, n_size=2)
-    output_model = wfs.do_all
+    wfs.do_all
     assert np.max(abs(wfs.diff)) < 0.001
     if delta_x_pixel + xerror >= 0 or not flip_dithers:
         assert wfs.off_x == delta_x_pixel + xerror
@@ -259,7 +258,7 @@ def test_refine_with_error(wfs_association):
         im2.save(path2)
     wfs = wfs_combine.DataSet(path1, path2, 'outfile.fits', do_refine=True, flip_dithers=True, psf_size=50,
                               blur_size=10, n_size=2)
-    output_model = wfs.do_all
+    wfs.do_all
     assert wfs.input_1.meta.observation.exposure_number == '1'
     assert wfs.input_2.meta.observation.exposure_number == '2'
     assert wfs.off_x == delta_pixel + shift_error
--- a/jwst/wfs_combine/tests/test_wfs_combine.py
+++ b/jwst/wfs_combine/tests/test_wfs_combine.py
@@ -4,6 +4,7 @@ from jwst import datamodels
 from jwst.assign_wcs import AssignWcsStep
 from jwst.associations.asn_from_list import asn_from_list
 from jwst.wfs_combine import WfsCombineStep
+from astropy.io import fits
 
 
 GOOD = datamodels.dqflags.pixel["GOOD"]
@@ -94,3 +95,22 @@ def test_create_combined(_jail, wfs_association,
     # Check that results are as expected
     assert result.data[5, 5] == result_data
     assert result.dq[5, 5] == result_dq
+
+
+def test_shift_order():
+    local_path = "/users/mregan/Documents/wfsc/"
+    path1 = str(local_path + "jw01465091001_02105_00001_nrca3_cal.fits")
+    path2 = str(local_path + "jw01465091001_02105_00002_nrca3_cal.fits")
+    asn = asn_from_list([path1, path2],
+                        product_name='jw00077_02105_nrca3_combine')
+    asn.data["program"] = "00077"
+    asn.data["asn_type"] = "wfs-image2"
+    asn.sequence = 1
+    asn_name, serialized = asn.dump(format="json")
+    path_asn = local_path + asn_name
+    with open(path_asn, "w") as f:
+        f.write(serialized)
+    result = WfsCombineStep.call(path_asn, do_refine=False)
+    fits.writeto(local_path+"combined_image"+asn.data["program"]+".fits", result.data, overwrite=True)
+    result = WfsCombineStep.call(path_asn, do_refine=True)
+    fits.writeto(local_path + "combined_image_refined" + asn.data["program"] + ".fits", result.data, overwrite=True)
--- a/jwst/wfs_combine/tests/test_wfs_combine.py
+++ b/jwst/wfs_combine/tests/test_wfs_combine.py
@@ -138,7 +138,7 @@ def test_shift_order_no_refine_no_flip(wfs_association):
         im2.save(path2)
     wfs = wfs_combine.DataSet(path1, path2, 'outfile.fits', do_refine=False, flip_dithers=False, psf_size=50,
                               blur_size=10, n_size=2)
-    wfs.do_all
+    wfs.do_all()
     assert wfs.input_1.meta.observation.exposure_number == '1'
     assert wfs.input_2.meta.observation.exposure_number == '2'
     assert wfs.off_x == abs(delta_pixel)
@@ -163,7 +163,7 @@ def test_shift_order_no_refine_with_flip(wfs_association):
         im2.save(path2)
     wfs = wfs_combine.DataSet(path1, path2, 'outfile.fits', do_refine=False, flip_dithers=True, psf_size=50,
                               blur_size=10, n_size=2)
-    wfs.do_all
+    wfs.do_all()
     wfs.input_1.close()
     wfs.input_2.close()
 #    assert wfs.input_1.meta.observation.exposure_number == '2'
@@ -226,7 +226,7 @@ def test_refine_no_error(wfs_association, xshift, yshift, xerror, yerror, flip_d
         im2.save(path2)
     wfs = wfs_combine.DataSet(path1, path2, 'outfile.fits', do_refine=True, flip_dithers=flip_dithers, psf_size=50,
                               blur_size=10, n_size=2)
-    wfs.do_all
+    wfs.do_all()
     assert np.max(abs(wfs.diff)) < 0.001
     if delta_x_pixel + xerror >= 0 or not flip_dithers:
         assert wfs.off_x == delta_x_pixel + xerror
@@ -262,7 +262,7 @@ def test_refine_with_error(wfs_association):
         im2.save(path2)
     wfs = wfs_combine.DataSet(path1, path2, 'outfile.fits', do_refine=True, flip_dithers=True, psf_size=50,
                               blur_size=10, n_size=2)
-    wfs.do_all
+    wfs.do_all()
     assert wfs.input_1.meta.observation.exposure_number == '1'
     assert wfs.input_2.meta.observation.exposure_number == '2'
     assert wfs.off_x == delta_pixel + shift_error
--- a/jwst/wfs_combine/tests/test_wfs_combine.py
+++ b/jwst/wfs_combine/tests/test_wfs_combine.py
@@ -5,7 +5,6 @@ from jwst.assign_wcs import AssignWcsStep
 from jwst.associations.asn_from_list import asn_from_list
 from jwst.wfs_combine import WfsCombineStep
 from jwst.wfs_combine import wfs_combine
-import astropy.modeling as modeling
 import numpy as np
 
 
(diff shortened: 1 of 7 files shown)


jwst/wfs_combine/wfs_combine.py
```diff
@@ -90,12 +94,27 @@ def do_all(self):
...
+            self.off_y = -1 * self.off_y
+            data_c, dq_c, err_c, diff_c = self.create_combined(im_2_a, im_1_a)
+        log.info(f"Final x, y offset in pixels: {self.off_x} {self.off_y}")
+
+        self.diff = diff_c
         # Create a new model using the combined arrays...
         new_model = datamodels.ImageModel(data=data_c, dq=dq_c, err=err_c)
         new_model.update(self.input_1)
-
+        print(" Refine value {}".format(self.do_refine))
```
hbushouse, line 112: need to change the print statement to `log.info()`

jwst/wfs_combine/wfs_combine.py
```diff
@@ -116,7 +135,7 @@ def create_aligned_2(self):
            smoothed image, and taking the mean of the coordinates of these
            pixels.  The x- and y-mean defines the centroid location.
         3. Set the limits of the subarrays for interpolation by taking this
-           centroid +/- PSF_SIZE and adding the BLUR_SIZE, taking the edges
+           centroid +/- psf_size and adding the BLUR_SIZE, taking the edges
```
hbushouse, line 138: It looks like `BLUR_SIZE` was removed above, so should this commentary be changed?

hbushouse: Oh, never mind. `BLUR_SIZE` is no longer a macro defined in the code, but passed in as a parameter.

jwst/wfs_combine/wfs_combine.py
```diff
@@ -591,54 +643,16 @@ def get_overlap(sci_int_1, sci_int_2, nom_off_x, nom_off_y):
...
 
-    sub_1 = sci_int_1[:fy - iy, :fx - ix]
-    sub_2 = sci_int_2[iy:fy, ix:fx]
+    sub_1 = sci_int_1[initial_y_1:final_y_1, initial_x_1:final_x_1]
+    sub_2 = sci_int_2[initial_y_2:final_y_2, initial_x_2:final_x_2]
 
     return sub_1, sub_2
 
 
-def optimize_offs(sci_nai_1, sci_nai_2):
```
hbushouse, line 603: This original scheme was the one requested by the WFS&C team quite a few years ago. Does the new cross-correlation scheme below perform essentially the same way, and if not, has the change been approved by the WFS&C team?

jwst/wfs_combine/wfs_combine_step.py
```diff
@@ -60,7 +65,7 @@ def process(self, input_table):
             # Save the output file
             if self.save_results:
                 self.save_model(
-                    output_model, suffix='wfscmb', output_file=outfile, format=False
+                    output_model, suffix='wfscmb.fits', output_file=outfile, format=False
```
hbushouse, line 68: I don't think `fits` should be included in the specification of the file name suffix. `suffix` is intended to specify only the final suffix contained in the root part of the file name. The file name extension is set within either the datamodels methods or stpipe based on the file format used (FITS in this case).

jwst/wfs_combine/tests/test_wfs_combine.py
```diff
@@ -89,8 +111,157 @@ def test_create_combined(_jail, wfs_association,
         im1.save(path1)
         im2.save(path2)
 
-    result = WfsCombineStep.call(path_asn, do_refine=False)
+    result = WfsCombineStep.call(path_asn, do_refine=False, flip_dithers=False, psf_size=50,
```
hbushouse, line 114: The `WfsCombineStep` class is no longer directly imported above, just the `wfs_combine` module. So does this need to be `result = wfs_combine.WfsCombineStep.call(...)`?

jwst/wfs_combine/tests/test_wfs_combine.py
```diff
@@ -4,18 +4,36 @@
...
 SATURATED = datamodels.dqflags.pixel["SATURATED"]
 
 
+def gaussian(x, mu, sig):
+    return 1. / (np.sqrt(2 * np.pi) * sig) * np.exp(-np.power((x - mu) / sig, 2) / 2)
+
+
+def gaussian2d(x, y, mux, muy, sigx, sigy):
+    return gaussian(x, mux, sigx) * gaussian(y, muy, sigy)
+
```
dmggh, line 22: astropy.modeling has a Gaussian1D and Gaussian2D

jwst/wfs_combine/wfs_combine.py
```diff
@@ -28,7 +21,8 @@ class DataSet:
 
     """
```
dmggh, line 22: For many functions, there are new parameters which should be added to the epydocs.

jwst/wfs_combine/wfs_combine.py
```diff
@@ -591,54 +643,16 @@ def get_overlap(sci_int_1, sci_int_2, nom_off_x, nom_off_y):
...
-            if (this_diff > max_diff):
-                max_diff = this_diff
-                max_off_x = off_x
-                max_off_y = off_y
-
-    return max_off_x, max_off_y
-
-
-def calc_cor_coef(sci_nai_1, sci_nai_2, off_x, off_y):
+def calc_refined_offsets(sci_nai_1, sci_nai_2, off_x, off_y, psf_size):
```
dmggh, line 655: I second Howard's question. Also, a description of this new 'refined' method should be documented in the epydocs and in readthedocs.

jwst/wfs_combine/wfs_combine.py
mperrin: Hi @hbushouse - @mwregan2 is now part of the WFS&C team, and has our 👍 for these changes. Thanks for checking! These improvements are in response to issues we've encountered during the various WF Team Practices and LRE sims.

jwst/wfs_combine/wfs_combine.py
mwregan2: removed it.

jwst/wfs_combine/wfs_combine_step.py
mwregan2: removed ".fits" from name.

jwst/wfs_combine/tests/test_wfs_combine.py
mwregan2: I'm not clear about the issue. This is working.

jwst/wfs_combine/tests/test_wfs_combine.py
mwregan2: I thought that there should be one. I tried using this but it is not a drop-in replacement. If you want, I'll figure out how to make it work.

jwst/wfs_combine/wfs_combine.py
```diff
@@ -81,21 +86,42 @@ def do_all(self):
...
+        if self.off_x < 0 and self.flip_dithers:
+            self.input_1.close()
+            self.input_2.close()
+            self.input_1 = datamodels.open(self.file2)
+            self.input_2 = datamodels.open(self.file1)
+            log.info('File 1 to combine: %s', self.file2)
+            log.info('File 2 to combine: %s', self.file1)
+        else:
+            log.info('File 1 to combine: %s', self.file1)
+            log.info('File 2 to combine: %s', self.file2)
```
jdavies-st, line 101: Having files opened in `__init__()` and closed in `do_all()` is not ideal.  File management should probably be happening all in `__init__()`.  Perhaps this should go in `__init__()` for the class?

jwst/wfs_combine/tests/test_wfs_combine.py
```diff
@@ -4,18 +4,36 @@
...
+    outarray = np.zeros_like(inarray)
+    for y in range(inarray.shape[0]):
+        for x in range(inarray.shape[1]):
+            outarray[y, x] = scale * gaussian2d(x, y, xcen, ycen, sigx, sigy)
+    return outarray
+
+
 @pytest.fixture(scope="module")
-def wfs_association(tmp_path_factory):
+def wfs_association(tmp_path_factory, imsize=10):
```
jdavies-st, line 33: pytest fixtures should not get arguments, i.e. `imsize=10` does not belong here.

jwst/wfs_combine/tests/test_wfs_combine.py
jdavies-st: There's the `wfs_combine` submodule and the `wfs_combine` subpackage, both named the same. This is causing confusion I think.

jwst/wfs_combine/tests/test_wfs_combine.py
```diff
@@ -89,8 +111,157 @@ def test_create_combined(_jail, wfs_association,
...
+        im2.save(path2)
+    wfs = wfs_combine.DataSet(path1, path2, 'outfile.fits', do_refine=False, flip_dithers=False, psf_size=50,
+                              blur_size=10, n_size=2)
+    wfs.do_all
+    assert wfs.input_1.meta.observation.exposure_number == '1'
+    assert wfs.input_2.meta.observation.exposure_number == '2'
+    assert wfs.off_x == abs(delta_pixel)
+    assert wfs.off_y == 0
+    wfs.input_1.close()
+    wfs.input_2.close()
```
jdavies-st, line 143: Ideally `DataSet` class would have an `__enter__` and `__exit__` method each defined so that one could do 

```
with wfs_combine.DataSet(path1, path2, ...):
    do things
```

with no calls to `.close()` needed.  Because of course if any of these asserts fail, then the files get left open.

jwst/wfs_combine/wfs_combine_step.py
jdavies-st: Actually, `self.suffix` should be defined for the class, and then it will work as intended.  This can be done in the `spec` class attribute so that there's a default, but that it can be overridden if needed.

jwst/wfs_combine/wfs_combine.py
jdavies-st: Or better yet, define `self.__enter__` and `self.__exit__` methods for the class, so that the `with` context manager can be used.

jwst/wfs_combine/tests/test_wfs_combine.py
hbushouse: Ah, OK. So I withdraw my original comment.

jwst/wfs_combine/wfs_combine_step.py
mwregan2: James, can you point me to an example of some code doing this? I've never really groked the pipeline class structure.

jwst/wfs_combine/wfs_combine.py
mwregan2: done

jwst/wfs_combine/wfs_combine.py
mwregan2: I've updated the epydocs and the readthedocs to match the current version.

jwst/wfs_combine/wfs_combine.py
```diff
@@ -1,9 +1,9 @@
 import logging
 
 import numpy as np
+import scipy.signal
 from scipy.interpolate import griddata
 from scipy.signal import convolve
```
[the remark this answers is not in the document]
mwregan2: done

jwst/wfs_combine/wfs_combine.py
```diff
@@ -58,14 +52,25 @@ def __init__(self, infile_1, infile_2, outfile, do_refine):
...
+        self.off_y = 0
+        self.flt_off_x = 0
+        self.flt_off_y = 0
+        self.diff = np.ndarray(shape=(2 * psf_size + 1, 2 * psf_size + 1))
+        self.flip_dithers = flip_dithers
+        self.psf_size = psf_size
+        self.blur_size = blur_size
+        self.n_size = n_size
+
+    @property
```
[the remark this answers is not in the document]
mwregan2: This was original code. I removed it.

jwst/wfs_combine/wfs_combine.py
mwregan2: I was trying to fit within the framework of the original code. Having a single-use class for this step doesn't make much sense to me and seems to be used to just have global variables (class attributes) to minimize parameter passing.
The close within the if statement is because we need to switch the contents of input_1 and input 2.  This is only called when the images need to be flipped. The files are left open in both cases of the if and are closed when the step finishes.  If I move the open to __enter__ and close to __exit__, I still have to close it and reopen it when the images need to be flipped.

jwst/wfs_combine/tests/test_wfs_combine.py
mwregan2: done

jwst/wfs_combine/tests/test_wfs_combine.py
mwregan2: I agree that __enter__ and __exit__ are a more symmetrical way to go but I don't know that it is worth the effort. I worry about unintended consequences. This code will only run during commissioning and it doesn't seem to be worth the effort.
I don't see major consequences of tests failing and thus files being left open. 
If I need to, I'll make the changes.

jwst/wfs_combine/wfs_combine.py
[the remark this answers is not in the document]
mwregan2: Removing it caused the class to become a function and all tests failed. I've replaced it. Clearly, this routine should be refactored but given the fact that it is only used in commissioning and the need to get it into the flight build, I think we should just leave this in.

jwst/wfs_combine/wfs_combine_step.py
jdavies-st: Here's how its done in `SourceCatalogStep`.  You can do the same here I think.

https://github.com/spacetelescope/jwst/blob/fc8a4077904d27898a8090ff467634c36435565a/jwst/source_catalog/source_catalog_step.py#L41

jwst/wfs_combine/wfs_combine.py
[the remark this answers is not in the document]
jdavies-st: No, it's wasn't originally like this.  You changed it.  See the diff here



You'll have to rewrite your new tests that call `DataSet.do_all` to be `DataSet.do_all()`.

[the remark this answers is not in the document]
hbushouse: I think this issue is worth resolving, i.e. remove the `property` tag and change the calls to use `DataSet.do_all()` (including in the main step class at https://github.com/spacetelescope/jwst/pull/6101/files#diff-07348f2cb4b6795e7d5f12d6d4741f5e33126ecc478ad35f67c20ccd5b56e041R54)

hbushouse (commented):
I think the remaining unresolved issues here can be left as is, since they aren't major show stoppers, other than removing the use of `@property` (see other in-line comment) and adding `suffix` to the `spec` block in `wfs_combine_step`. The '.fits' extension should not be hardcoded in the call to save the output.

jwst/wfs_combine/tests/test_wfs_combine.py
mwregan2: One thing: I'm wrong about the code only running during commissioning. It will run whenever we do a wavefront sensing observation. The current plan is this will happen every two weeks. Even so, the orphan open files would only happen when running tests that fail.

jwst/wfs_combine/wfs_combine.py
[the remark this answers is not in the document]
mwregan2: I agree that @property is new. I have no idea how it got there though. I've removed it and updated the calls to do_all().

jwst/wfs_combine/wfs_combine_step.py
mwregan2: Is this code section even supposed to be here? I can't find the equivalent in any other step. When I try to run the step setting save_result to True, it crashes with an incorrect file extension exception. 
The code after this sets save_results to False. So, I'm really confused. The comments claim that Format=True needs to be set. I don't know what that does but something is wrong,.

jwst/wfs_combine/wfs_combine_step.py
```diff
@@ -60,8 +65,7 @@ def process(self, input_table):
             # Save the output file
             if self.save_results:
                 self.save_model(
-                    output_model, suffix='wfscmb', output_file=outfile, format=False
-                )
+                    output_model, output_file=outfile, format=False)
```
hbushouse, line 68: I think you do need to set the suffix somewhere, either right here when `save_model` is being called (as before), or include it as a param in the `spec` block above (https://github.com/spacetelescope/jwst/pull/6101/files#diff-07348f2cb4b6795e7d5f12d6d4741f5e33126ecc478ad35f67c20ccd5b56e041R17). Otherwise stpipe will likely use the default suffix based on the step name (wfscombine), which is not the official suffix expected for this product (downstream WFS software and the archive are expecting "wfscmb"). This step is setup differently than all other steps (which don't specific suffix names), because it's never run as part of a larger pipeline module. For steps that are run from a pipeline, the output data model is passed back to the pipeline and the pipeline takes care of saving it to a file. That's why you don't see this kind of operation in any other steps.

hbushouse (commented):
In addition to handling the remaining suffix issue, there were also a couple of flake8/pycodestyle failures that need to be cleaned up (unused import, etc.). Once these are addressed I think it'll be ready to go.

hbushouse:
Oh, it also still needs a change log entry.

jwst/wfs_combine/wfs_combine_step.py
hbushouse: When you run the step standalone, e.g. via the command line using "strun jwst.wfs_combine.WfsCombineStep <some_asn.json>" what kind of product type suffix does it use when writing the combined output image to disk? If `suffix` is not specified anywhere, I think it's going to use a generic file name suffix, rather than the "wfscmb" string that we want for this product. `suffix` should be added to the `spec` block up at line 22 and set to a default value of "wfscmb".

mwregan2: I added the suffix of "wfscmb" to the block in line 23. I checked in the debugger and suffix is set correctly. The old value was "wfscombinestep". So, it was not horrible just not the expected value.

hbushouse (approved):
Looks ready to go.

hbushouse:
By the way, I'm setting the milestone for this one to B7.8.1, because I assume this would qualify as being necessary to support commissioning activities, right?

mperrin:
Hi @hbushouse that's correct, necessary for commissioning activities. Much appreciated!

PR #6120, opened 2021-06-04
JP-2128: Switch default interpolation method in blot to poly5 instead of linear

### LINKED ISSUE #6116: outlier_detection blotting segfaults on linux
_Issue [JP-2128](https://jira.stsci.edu/browse/JP-2128) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

On Linux, calwebb_spec->outlier_detection ends in a segfault during blotting. On mac this runs fine.

mcara:
Resolves #6116 / [JP-2128](https://jira.stsci.edu/browse/JP-2128)

This PR replaces the default interpolator ('linear') in blot (used in `outlier_detection`) used when interpolator is not specified with `'poly5'`. The `'poly5'` interpolator is the default one used in `drizzlepac` and it is extensively used (~= tested).

This PR does not solve the underlying issue in the `drizzle` package but should alleviate the issue described in #6116.

NOTE to maintainers. This is not a bug in the `outlier_detection` step per se and so I did not add the `bug` label but feel free to classify it as you wish.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -240,6 +240,10 @@ master_background
 outlier_detection
 -----------------
 
+- Avoid using 'linear' interpolation method as default for ``blot`` due to
+  a bug in the implimentation of the bilinear interpolator in the ``drizzle``
+  package. Now the default value will be 'poly5'. [#6116]
+
 - Outlier detection on non-dithered images is implemented with a simple sigma
   clipping, dithered outlier detection cleaned up and HST specific steps removed
   and additional tests added. [#5822]
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -304,7 +304,7 @@ class OutlierDetection:
 
     def blot_median(self, median_model):
         """Blot resampled median image back to the detector images."""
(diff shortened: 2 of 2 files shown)


nden (approved):

jdavies-st (approved):
So this solves the problem, i.e. moves back to poly5 interpolation due to whatever buggy behavior exists in bilinear.  Agree this this probably the best we can do right now to get a working `outlier_detection` step out that doesn't segfault for B7.8.

That said, I did a comparison of our regression test data produced in this PR with poly5 and the previous linear interpolation, and the linear interpolation looks better for NIRCam long detectors.  It flags more real cosmic rays.

A reason may be that for the default pipeline, we are currently drizzling to the output image with the same pixel scale as the input.  So generally the resampled image and the blotted image are similarly well- or under-sampled.  Clearly it would be better to investigate this further, as this observation is based on 1 NIRCam LONG simulated dataset.

MIRI imaging data look equally good with both types of interpolation, most of the differences coming along the edges of the imaging area.

Tagging @karllark here for any input.

PR #6130, opened 2021-06-08
JP-2091: Fix datamodels.open interaction with pathlib.Path objects

### LINKED ISSUE #6054: datamodels.open improperly handle pathlib.Path objects
_Issue [JP-2091](https://jira.stsci.edu/browse/JP-2091) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

`datamodels.open` improperly handle `pathlib.Path` objects as demonstrated below:

`>>> from pathlib import Path`
`>>> from jwst import datamodels`
`>>> image_model = datamodels.ImageModel((50, 50))`
`>>> print(image_model)`
`<ImageModel(50, 50)>`
`>>> image_model.save('image_model.fits')`
`'image_model.fits'`
`>>> image_model_path = Path('image_model.fits')`
`>>> from_path = datamodels.open(image_model_path)`
`>>> print(from_path)`
`<JwstDataModel from image_model.fits>`
`>>> from_str = datamodels.open(str(image_model_path))`
`>>> print(from_str)`
`<ImageModel(50, 50) from image_model.fits>`

tapastro:
Closes #6054

Resolves JP-2091

**Description**

This PR currently does the minimum possible work to fix the issue - if open() sees a Path object, it converts to string. I don't know if something more was wanted, but reading old tickets regarding the use of Path objects in the pipeline, it seems that the intended effect is to allow their use externally but not promote their use internally?

Checklist
- [x] Tests

- [ ] Documentation

- [x] Change log

- [x] Milestone

- [x] Label(s)


jwst/datamodels/tests/test_models.py
```diff
@@ -59,6 +60,17 @@ def make_models(tmp_path):
     }
 
 
+def test_init_from_pathlib(tmp_path):
+    """Test initializing model from a PurePath object"""
+    path = tmp_path / "pathlib.fits"
```
stscieisenhamer, line 65: Is `tmp_path` already a `Path` object? Or something else?

stscieisenhamer (commented):
About promoting internally: It is worth exploring. There may be points where the `str` is needed, but may be few and the conversion could be dealt with there.

jwst/datamodels/tests/test_models.py
tapastro: I honestly don't know where/how tmp_path gets initialized, but it seems to be an environment where pytest creates fits files during a run, based on the fixture above this test. I believe it should be a string, but I convert it to a Path object in the datamodels.open() call. I notice in the test comment that I say it uses a PurePath object, though I believe it's just a Path object, if that matters. I can clear that up.

stscieisenhamer: It actually is a Path object; [see the docs](https://docs.pytest.org/en/6.2.x/tmpdir.html). So you should be able to get away with not actually re-Path-ing it.

tapastro:
Regarding the current path it takes:

It does not trigger any of the isinstance() checks and gets down to the _class_from_shape() call before a model is instantiated - because shape is defined as an empty tuple and hdulist is defined as an empty dict, this call doesn't cause an error but finds len(shape)==0 and creates a JwstDataModel ([L305](https://github.com/spacetelescope/jwst/blob/a42adb25d24494d9f4253f670c258e1187253550/jwst/datamodels/util.py#L305)). The file is currently not accessed if .open() is provided a Path object - the current fix changes that, but an explicit handler for a Path object rather than a string conversion could be an option if this is too ham-fisted.

jdavies-st:
Tyler it looks like the bug seen above in @stscieisenhamer's bug report is that `datamodels.open()` opens the written out `ImageModel` as a generic `JwstDataModel` instead of as the correct `ImageModel` type.  So that would be the test case to add, which it looks like you have.

My comment above about our tests are that we do have some tests that test aspects of this, but clearly not this buggy case.

jdavies-st:
This looks good to me.  If we handle as `isinstance(inti, pathlib.Path)` explicitly then that object gets passed along to whichever thing is doing the opening, and that might be a bigger can of worms.  This seems good for now.  We mostly just want to allow the public interface to use `pathlib.Path` objects.

PR #6131, opened 2021-06-08
JP-2104: Do not consider GROUP candidates for WFS&C observations

### LINKED ISSUE #6075: Do not consider GROUP candidates for WFS&C observations
_Issue [JP-2104](https://jira.stsci.edu/browse/JP-2104) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

WFS&C observations should not produce associations for any GROUP association candidate.

Given the attached pool, the command

`{{asn_generate -i c1001`}}

should produce no associations.

tapastro:
Closes #6075

Resolves [JP-2104](https://jira.stsci.edu/browse/JP-2104)

**Description**

This PR adds an extra validation check for Lv3_WFSCMB to ensure that the association id type is not 'group'.

Checklist
- [ ] Tests

- [ ] Documentation

- [x] Change log

- [x] Milestone

- [x] Label(s)


tapastro:
@stscieisenhamer Not entirely clear to me why 'member' is a required argument to a validity update, but it seems to force 2 arguments into the function so I need to allow for it. Otherwise working as intended?

stscieisenhamer:
The validation bit has some tech-debt in there which needs refactor at some point in the distant future. Will review momentarily.

hbushouse (approved):

PR #6138, opened 2021-06-09
 fix option to create channel cubes - JP 2105

### LINKED ISSUE #6076: Per-channel cube build option isn't working
_Issue [JP-2105](https://jira.stsci.edu/browse/JP-2105) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

Running cube_build with the output_type = 'channel' option should create an output cube for each channel of data, but instead crashes with
TypeError: 'list' object cannot be interpreted as an integer
in v1.2.0 of the pipeline (not yet an option in the dropdown list above).  Jane has already been able to reproduce this bug using existing data.

jemorrison:
Fixes #6076 
Resolves [JP-2105](https://jira.stsci.edu/browse/JP-2105)

**Description**

The option to make MIRI with output_type = channel was not working. This option takes a list of files and creates IFU Cubes for each channel (the input data can be from difference subchannels - bands)

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


drlaw1558 (approved):
Looks good to me; confirmed that this fixes the problem and produces cube_build per-channel behavior as expected.

stscieisenhamer (approved):

hbushouse:
Does this really relate to JP-2015 in some way, which pertains to implementing some of the `cube_build` functions in C? The bug may have been discovered while doing work for JP-2015, but as far as I can tell this PR does nothing to actually implement/resolve that ticket. Hence the reference to it should be removed.

hbushouse:
> Does this really relate to JP-2015 in some way, which pertains to implementing some of the `cube_build` functions in C? The bug may have been discovered while doing work for JP-2015, but as far as I can tell this PR does nothing to actually implement/resolve that ticket. Hence the reference to it should be removed.

Never mind. The relevant ticket is actually JP-2105. I've fixed all the references.

PR #6144, opened 2021-06-11
JP-2089: Rename association rule to Asn_Lv3NRCWFSS

### LINKED ISSUE #6052: Rename `Asn_Lv3SpectralSource` to `Asn_Lv3NRCWFSS`
_Issue [JP-2089](https://jira.stsci.edu/browse/JP-2089) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Rename `Asn_Lv3SpectralSource` to `Asn_Lv3NRCWFSS`. This rule now only affects NIRCam WFSS mode so the name should match.

tapastro:
Closes #6052

Resolves [JP-2089](https://jira.stsci.edu/browse/JP-2089)

**Description**

This PR renames an association rule from a generic title to NIRCam-specific, as it now only covers NIRCam WFSS.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/associations/lib/rules_level3.py
+++ b/jwst/associations/lib/rules_level3.py
@@ -23,7 +23,7 @@ __all__ = [
     'Asn_Lv3NRSIFU',
     'Asn_Lv3NRSIFUBackground',
     'Asn_Lv3SlitlessSpectral',
-    'Asn_Lv3SpectralSource',
+    'Asn_Lv3NRCWFSS',
     'Asn_Lv3SpectralTarget',
     'Asn_Lv3TSO',
     'Asn_Lv3WFSCMB',
@@ -565,7 +565,7 @@ class Asn_Lv3SlitlessSpectral(AsnMixin_Spectrum):
 
 
 @RegistryMarker.rule
-class Asn_Lv3SpectralSource(AsnMixin_Spectrum):
+class Asn_Lv3NRCWFSS(AsnMixin_Spectrum):
     """Level 3 slit-like, multi-object spectrographic Association
 
     Characteristics:
@@ -604,7 +604,7 @@ class Asn_Lv3SpectralSource(AsnMixin_Spectrum):
         ])
 
         # Check and continue initialization.
-        super(Asn_Lv3SpectralSource, self).__init__(*args, **kwargs)
+        super(Asn_Lv3NRCWFSS, self).__init__(*args, **kwargs)
 
     @property
     def dms_product_name(self):
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,10 @@
 1.2.4 (unreleased)
 ==================
 
+associations
+------------
 
+- Renamed Asn_Lv3SpectralSource to Asn_Lv3NRCWFSS [#6144]
 
 
 1.2.3 (2021-06-08)


stscieisenhamer (approved):

hbushouse:
Is there a corresponding rule of a different name that handles construction of spec3 ASN's for *NIRISS* WFSS observations? If so, is there a need to have separate rules for NIRCam vs NIRISS WFSS, i.e. why would they require different handling?

hbushouse:
Ah, found the corresponding rule for NIRISS WFSS: "Lv3WFSSNIS".  I suggest we try to be consistent in our naming schemes, many of which take the form `Asn_Lv3[instrument][mode]`, such as the new `Asn_Lv3NRCWFSS` implemented here. Hence I would suggest that we also change the name of the corresponding NIRISS WFSS rule to `Asn_Lv3NISWFSS`.

jwst/associations/lib/rules_level3.py
```diff
@@ -565,7 +565,7 @@ def __init__(self, *args, **kwargs):
 
 
 @RegistryMarker.rule
-class Asn_Lv3SpectralSource(AsnMixin_Spectrum):
+class Asn_Lv3NRCWFSS(AsnMixin_Spectrum):
     """Level 3 slit-like, multi-object spectrographic Association
```
hbushouse, line 569: These seem like pretty generic descriptions for a rule that specifically applies to only NIRCam WFSS mode. "Level 3 slit-like, multi-object" sounds to me like it should also include NIRSpec MOS, but it doesn't.

hbushouse: Also, what's the purpose of the references to `nrs_autoflat` and `nrs_autowave` in the rules below? What do these have to do with NIRCam WFSS?

jwst/associations/lib/rules_level3.py
```diff
@@ -604,7 +604,7 @@ def __init__(self, *args, **kwargs):
         ])
```
hbushouse, line 604: Looks to me like the inclusion of `Constraint_MSA()` a few lines above means this is also going to generate spec3 ASN's for NRC_MSASPEC observations, hence the rule name `NRCWFSS` seems too specific. Should we pull the NIRSpec stuff into a separate (new) rule?

jwst/associations/lib/rules_level3.py
stscieisenhamer: good catch. those should be removed.

jwst/associations/lib/rules_level3.py
stscieisenhamer: Hmmmm, maybe jumped the gun on the renaming. I was just looking at the primary exp_type check.

So, this rule originally handled all "source-focused" associations. Then niriss wfss got removed and I mistakenly thought that was all there was to it.

Two ways to go about this: Either leave the original rule as is, except for commentary about "except NIRISS WFSS". Or split out and make a rule for each of the modes. Original philosophy was to make the rules as "science-focused" as possible and not "instrument-mode-focused", so the my preference is the former option.

hbushouse: The existing `WFSSNIS` rule has constraints on exp_type ('nis_wfss'), filter ('gr150r', 'gr150c'), and pupil. Meanwhile the `SpectralSource` rule being addressed here only seems to use exp_type ('nrc_wfss') to select NIRCam WFSS exposures. Any ideas as to why the one for NRC is so simple and generic, while the one for NIS is more specific? Just wondering how hard it might be to create a generic `WFSS` rule that could apply to both NRC and NIS and then have a separate one for NRS MSASPEC.

Alternatively, is there a way to combine all 3 into a single rule that covers all multi-object modes (NRS MOS, NRC WFSS, and NIS WFSS)?

stscieisenhamer: About the last point: One could combine, but that would create an extremely complicated rule, the main differential being the grism/catalog aspect. Not sure that would ever be de-buggable.

About why NIRISS Grism is different: Would have to explore more, but the split occurred at PR #1983.

hbushouse: It looks to me like we had a single, generic `Asn_WFSS` rule, which was not actually completely generic, because it only included exp_type `nis_wfss`, up until https://github.com/spacetelescope/jwst/pull/2001 was introduced, which renamed it to the specific `Asn_WFSSNIS` rule. I'm guessing we had the unique rule for nis_wfss due to issues with using the filter and pupil values to construct the appropriate opt_elem pieces of the product name(s) and perhaps also for use in determining the correct corresponding catalog and direct image names?

But given that NIRCam also uses filter and pupil, why don't we need those special hooks for nrc_wfss? Perhaps it's because the NIRISS grisms happen to be physically installed in the filter wheel, while for NIRCam they're in the pupil wheel?

stscieisenhamer: Just for history/context: The original idea was to have just two spectrographic rules:

- Single Object/target-based
- Mulit-object/source-based

Then reality started hitting. Fix Slit/Slitless, whichever mode is the source-based one, had to be separated out. Then WFSS was actually finalized, originally based on NIRISS, which was called Asn_WFSS. The presumption was that NIRCam would also fall under there, but either didn't need to or possibly it should but just hasn't been migrated. The final "source" mode is IFU, but that either never made it in or was determined to be to different to be grouped.

So, with 20-20 hindsight, the ideal groupings would be, again trying to be science-based as opposed to instrument:

- Single Object/Target
- IFU
- MSA
- Fixed slit/slitless (which ever is the source one)
- Grism/catalog

and the suggestion on naming would be `Asn_Lv3[mode][instrument]` where `instrument` is the optional part, reverse of what was suggested above. 

Methinks this issue should just be closed out and a new issue (again methinks there may be an old github issue already) created to deal with just the overall rename/refactor. There is an issue about removing the redundant `Asn_` prefix; tech debt from a previous version.

hbushouse: I agree with closing this PR and related ticket/issue and filing a new ticket aimed at a bit more ambitious refactoring/reorganizing of the various spectroscopic rule classes. This PR is not needed to fix an existing problem, it was only for the purpose of doing some clean-up.

PR #6148, opened 2021-06-14
Initial soss code

talensgj:
This PR concerns the addition of a specialised extraction method to the Extract1dStep for NIRISS SOSS observations. This specialised extraction deals with the overlap between the different spectral orders. I apologise if the format of this description does not comply with the standards of the project, it is my first time making a PR for this project.

The code in this PR is still under development and the purpose of this PR is to provide a space for discussion with StSci on the best way to implement some aspects of this new extraction, and to merge when we have a satisfactory 1st working version.


talensgj:
The first issue I would like to discuss is the best way to branch the pipeline towards the specialized extraction in case of a NIRISS SOSS observation. I have currently added a tentative if statement to the existing Extract1dStep, however I currently see no overlap between the parameters and reference files used by the existing extraction and the NIRISS SOSS extraction. I would therefore like to discuss if a new Step made specifically for the SOSS extraction might be preferable over adding it to the existing Extract1DStep.

tapastro:
Hi Geert Jan, welcome to the repo! I'll try to help with implementing this extraction method. 

The Extract1dStep attempts to be the singular location for extractions of this type, so for organization's sake we would like to keep this method under the coverage of this step. We will use something just like what you've done to branch SOSS observations from the top-level Extract1dStep to different functions, but we can keep those functions stored in the step directory.

In terms of orthogonal parameters and reference files, we have other steps that have encountered this issue, e.g. the AssignWcsStep. Looking [here](https://github.com/spacetelescope/jwst/blob/master/jwst/assign_wcs/assign_wcs_step.py) you can see that many reference file types may be provided, depending on instrument. For now, you can add your new reference file types to the reference_file_types list, and once they live on CRDS with appropriate maps they should be called when SOSS data is provided to the step.

As a side note: I am currently in the early stages of refactoring this step, though changes where you are forking, at the Extract1dStep level, are likely to be minimal.

talensgj:
Hi Tyler, 

Alright, that makes sense on the issue of orthogonal parameters.

When you say "We will use something just like what you've done to branch SOSS observations from the top-level Extract1dStep to different functions, but we can keep those functions stored in the step directory." do you refer to my use of a soss_extract submodule? This was another thing I wanted to ask about. I'm currently using a submodule out of convenience, but if it is preferred I can consolidate it down to 1 or 2 files in the step directory once everything is in place.

hbushouse:
Should we just go ahead and close this PR (without merging), since it's been superseded by #6467?

tapastro:
> Should we just go ahead and close this PR (without merging), since it's been superseded by #6467?

Seems appropriate!

hbushouse:
Closing, because it's been replaced by #6467

PR #6160, opened 2021-06-21
Open specwcs reffile with context manager

jdavies-st:
This PR fixes a problem we saw in a unit test where the `specwcs` reffile was incompletely downloaded from CRDS and so `asdf.open()` failed.  When it failed, it was not closed, as it never went through `__exit__`.

~Not sure if this is a problem with the `with` context manager in `asdf.open()`, i.e. is heavy-lifting happening in `AsdfFile.__init__` that should be happening in `AsdfFile.__enter__`?  Not sure.~

~Regardless, in this case, as soon as `asdf.open(file)` was called, it opens it without using `__enter__`.  Is this a bug or feature in `asdf` @eslavich ?  Or just a small bug here.~

Separate issue filed in https://github.com/asdf-format/asdf/issues/1006

Also cleaned up a try/except block below.  `velosys` is guaranteed to be an attribute as it is in the schema.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/assign_wcs/niriss.py
+++ b/jwst/assign_wcs/niriss.py
@@ -135,29 +135,25 @@ def niriss_soss(input_model, reference_files):
     try:
         # We'd like to open this file as a DataModel, so we can consolidate
         # the S3 URI handling to one place.  The S3-related code here can
-        # be removed once that has been changed.
+        # be removed once we have a specwcs DataModel subclass.
         if s3_utils.is_s3_uri(reference_files['specwcs']):
-            wl = asdf.open(s3_utils.get_object(reference_files['specwcs']))
+            uri = s3_utils.get_object(reference_files['specwcs'])
         else:
-            wl = asdf.open(reference_files['specwcs'])
-        with wl:
-            wl1 = wl.tree[1].copy()
-            wl2 = wl.tree[2].copy()
-            wl3 = wl.tree[3].copy()
-    except Exception:
-        raise IOError('Error reading wavelength correction from {}'.format(reference_files['specwcs']))
+            uri = reference_files['specwcs']
+        with asdf.open(uri) as af:
+            wl1 = af.tree[1].copy()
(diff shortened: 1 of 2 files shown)


jwst/assign_wcs/niriss.py
```diff
@@ -135,29 +135,25 @@ def niriss_soss(input_model, reference_files):
...
-            wl1 = wl.tree[1].copy()
-            wl2 = wl.tree[2].copy()
-            wl3 = wl.tree[3].copy()
-    except Exception:
-        raise IOError('Error reading wavelength correction from {}'.format(reference_files['specwcs']))
-
-    try:
-        velosys = input_model.meta.wcsinfo.velosys
-    except AttributeError:
-        pass
```
jdavies-st, line 153: No need for a try/except block here as it will never fail to have this attribute since it is in the wcsinfo schema.

jwst/assign_wcs/niriss.py
```diff
@@ -135,29 +135,25 @@ def niriss_soss(input_model, reference_files):
     try:
         # We'd like to open this file as a DataModel, so we can consolidate
         # the S3 URI handling to one place.  The S3-related code here can
-        # be removed once that has been changed.
+        # be removed once we have a specwcs DataModel subclass.
         if s3_utils.is_s3_uri(reference_files['specwcs']):
-            wl = asdf.open(s3_utils.get_object(reference_files['specwcs']))
+            uri = s3_utils.get_object(reference_files['specwcs'])
```
eslavich, line 140: This accepts a URI and returns an `io.BytesIO` (not another URI)

jdavies-st: I changed the variable name to make this clear.

eslavich (approved):
LGTM

PR #6165, opened 2021-06-21
Don't cache CRDS reffiles between Github Actions jobs

jdavies-st:
This turns off Github Actions caching between jobs.  It has not been working correctly lately, and seems to be causing issues as well.

Using a cache only seems to work on common branches, i.e. so `master` can use caches created on `master`, but not those created on branches on forks.  See the following:

https://github.community/t/actions-cache-cache-not-being-hit-despite-of-being-present/17956

These issues need to be resolved before it is turned on again.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [ ] Label(s)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,18 +81,6 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Get CRDS context
-        id: crds-context
-        run: |
-          pip install crds
-          echo "::set-output name=pmap::$(crds list --resolve-contexts --contexts | cut -f 1 -d ".")"
-
-      - name: Restore CRDS cache
-        uses: actions/cache@v2
-        with:
-          path: ${{ env.CRDS_PATH }}
-          key: crds-${{ matrix.toxenv }}-${{ steps.crds-context.outputs.pmap }}
-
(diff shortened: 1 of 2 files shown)


eslavich:
It looks like branches should be able to access caches generated on [master](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key), but even that may not be worth it in our case since the limit per repository is only [5 GB](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy).

eslavich (approved):
LGTM

jdavies-st:
I'm noting here that the problem with this implementation of caching the contents of CRDS_PATH was that the caching was occuring on all CI jobs, including those that didn't run the tests and populate CRDS_PATH with a snapshot of the reffiles needed by the tests.  So for instance, the codestyle/flake8 job ran, created an empty cache for the non-existent files in CRDS_PATH (no tests needing them ran), and since this job finishes first, that empty cache is used by every subsequent job.

Caching the contents of CRDS_PATH can only happen if `pytest` is run in the job.

PR #6170, opened 2021-06-23
Test with safe casting

eslavich:
**Description**

This PR is to run the tests with a branch of stdatamodels that raises errors on unsafe array casts.  It should not be merged.

Checklist
- [ ] Tests

- [ ] Documentation

- [ ] Change log

- [ ] Milestone

- [ ] Label(s)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,7 +43,7 @@ install_requires =
     scipy>=1.1.0
     spherical-geometry>=1.2.20
     stcal>=0.2.1
-    stdatamodels>=0.2.0,<1.0
+    stdatamodels @ git+https://github.com/eslavich/stdatamodels.git@eslavich-strict-gentle-asarray
     stpipe>=0.1.0,<1.0
     stsci.image>=2.3.3
     tweakwcs>=0.7.2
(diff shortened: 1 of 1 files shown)


jdavies-st:
So there's a lot of

```
ValueError: Array dtype float64 cannot be safely cast to schema dtype float32
```

here.  Does this mean we have to do lots of `.astype(np.float32)` calls (which makes a truncated copy) on arrays before assigning them back to the datamodel attribute that has the schema checking its dtype?

I've always been skeptical of keeping our data arrays as `float32` in our datamodel schemas.  If we are going to do safe casting checks (which we should do), this will add a lot of by-hand casting when we want to be doing calculations in 64-bit.

I.e. the disconnect of wanting the in-memory representation of all our data arrays to be 64-bit, but when finally written out to a FITS file to be 32-bit (to save storage costs?) is not rearing its head.  Again.

And there's already a lot of casting to 64-bit of 32-bit arrays that are read in.

Can we revisit this?  What are the trade-offs?

melanieclarke:
Closing - it looks like this PR is no longer needed for testing.

Issue #527, opened 2016-12-09
Move MultiSlit slit attributes to meta within the slit


jdavies-st:
Right now `xstart`, `xsize` and `name` are appended directly to an individual slit in `extract_2d`, while the wcs object is attached to the the slit's `meta`.  This causes problems when repackaging MultiSlit data (say from exposure-based to source-based) with running the `update()` method, which only copies over items from meta.

Currently for MultiSlitModel `model`:

```
model.slits[0].xstart
model.slits[0].xsize
model.slits[0].name
```

Possible change:

```
model.slits[0].meta.xstart
model.slits[0].meta.xsize
model.slits[0].meta.name
```

These values (and the others) should be updated in the schema, so that they get validated when written.  Code upstream would have to be changed too.

Interested in people's thoughts on this.

hbushouse:
I agree that these values should be part of a meta attribute in a slit instance. They start out as part of .meta in the original ImageModel, so they ought to also be within .meta when a MultiSlit is created from that ImageModel. This should be a fairly simple change (I would think) in the MultiSlit schema and also then offers the advantage that they get copied when using the .update method.

hbushouse:
From what I can tell it looks like the MultiSlitModel schema is still setup such that attributes like xstart, xsize, etc. are not contained in a meta object. Is this still an issue to be pursued? If so, it's a big enough change to enough code that we should wait for B7.2.
@nden @jdavies-st

nden:
I am in the process of refactoring this code in #1070 and this will be resolved there. Slip to 7.2.

stscicrawford:
Closing as refactoring occurred in #1070.  If there was more work or that was not the case, please re-open or create a new ticket.

jdavies-st:
Reopening as this work has not been done yet.

nden:
@stscijgbot-jp

stscijgbot-jp:
_This issue is tracked on JIRA as [JP-2175](https://jira.stsci.edu/browse/JP-2175)._

stscijgbot-jp:
_Comment by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva) on [JIRA](https://jira.stsci.edu/browse/JP-2175?focusedCommentId=536531#comment-536531):_

[Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse) Just to confirm, this essentially means moving the attributes under slitmeta,

[...]

PR #6175, opened 2021-06-30
JP-2074: Update wfss_contam to process multiple orders

### LINKED ISSUE #6015: Update wfss_contam to process all spectral orders
_Issue [JP-2074](https://jira.stsci.edu/browse/JP-2074) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The prototype version of the wfss_contam step is hardwired to only use spectral order 1 when generating simulated contamination spectra and subtracting them from a science exposure. This needs to be enhanced to cover all spectral orders for which calibration data are available for each NIRCam and NIRISS WFSS mode.

hbushouse:
Closes #6015

Resolves [JP-2074](https://jira.stsci.edu/browse/JP-2074)

**Description**

This PR addresses updates the baseline wfss_contam step code to process all available spectral orders in a given grism exposure. Some restructuring of the underlying INS code was necessary to do this in a simple way, such that the initialization of the Observation class for the grism exposure no longer does anything that's order-dependent. All of the order dependencies are now contained in the functions that actually create the dispersed spectra.

Also includes some code cleanup and adds hooks for saving 2 kinds of intermediate data files.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/wfss_contam/wfss_contam.py
+++ b/jwst/wfss_contam/wfss_contam.py
@@ -1,22 +1,21 @@
 #
 #  Top level module for WFSS contamination correction.
 #
-from os.path import splitext
+import logging
+import multiprocessing
+import numpy as np
 
 from jwst import datamodels
 from .observations import Observation
 from .sens1d import get_photom_data
-from ..lib.suffix import replace_suffix
-
-import logging
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 
-def contam_corr(input_model, waverange, photom):
+def contam_corr(input_model, waverange, photom, max_cores):
     """
-    The main WFSS contam correction function
+    The main WFSS contamination correction function
 
     Parameters
     ----------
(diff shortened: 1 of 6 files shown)


jwst/wfss_contam/wfss_contam.py
```diff
@@ -26,71 +25,159 @@ def contam_corr(input_model, waverange, photom):
...
+    contam_model.slits.extend(slits)
 
     # Set the step status to COMPLETE
     output_model.meta.cal_step.wfss_contam = 'COMPLETE'
 
-    return output_model
+    return output_model, simul_model, contam_model
+
+
+def copy_slit_info(input_slit, output_slit):
```
stscieisenhamer, line 159: should this be a method of the model (SlitModel)?

stscieisenhamer (approved):
LGTM

jwst/wfss_contam/wfss_contam.py
hbushouse: That would be convenient, because it's needed in at least a couple different steps in the pipeline (e.g. transferring slit meta data from 2D cutouts to 1D spectra in extract_1d). I tried using `output.update(input, only="SCI")`, but that had no effect, for some reason. Different observing modes have different subsets of the slit meta data populated, but that could be handled in a generic method by only copying those items that exist in the input.

jdavies-st: This is because of #527.  `update()` only works on things in `meta`.

hbushouse: Ah, of course. And of the slit "meta" data is not within the actual `meta` tree. Another good reason to move it all into `meta` (some day).

jdavies-st: We could do that next week (if its prioritized).  😉 

We could also move the gWCS attribute `wcs` out of `meta`, as it doesn't belong there either.

hbushouse: The main concern with finally making that move is whether and how much existing code base (outside of the pipeline) might get broken.

jdavies-st: We could leave behind a pointer (i.e. keep it working) and issue a `DeprecationWarning`.  And leave it like that for a while.  But then at least internally `update` would work as we expect, and we could remove a bunch of duplicate code that tries to work around this.

PR #6218, opened 2021-07-14
JP-678 Update okifying to handle full folder updates for associations

### LINKED ISSUE #3444: Ensure association regression tests upload failed regressions to artifactory
Issue [JP-678](https://jira.stsci.edu/browse/JP-678) was created by Jonathan Eisenhamer:

In Jenkins, results from failed regressions are uploaded to artifactory. However, association regressions do not upload on failure. Implement such functionality.

stscieisenhamer:
Closes #3444

Resolves [JP-678](https://jira.stsci.edu/browse/JP-678)

**Description**

The okifying system is updated to handle association tests that require full folder updates
to the regression truth data.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


stscieisenhamer:
@jdavies-st ready for review.

hbushouse (commented):
Looks OK to me, but I'll leave the definitive assessment to @jdavies-st .  Have you been able to verify that none of these changes messes up the okifying of "normal" tests in any way?

stscieisenhamer:
Yes, that is basically the reason for the last flurry of commits.

hbushouse (approved):
Fix the conflicts in the change log and let's merge this puppy.

jwst/regtest/regtestdata.py
```diff
@@ -52,18 +53,30 @@ def __init__(self, env="dev", inputs_root="jwst-pipeline",
...
+    @property
+    def input(self):
+        return self._input
+
+    @input.setter
+    def input(self, value):
+        if value:
+            self._input = os.path.abspath(value)
+        else:
+            self._input = value
```
jdavies-st, line 79: Ok, this diff is impossible to follow, as I assume you've changed things, but since you've also re-ordered the methods, it's impossible to follow.  Could you keep the order the same, so it is clear to see what has changed?

Feel free to reorder them in a follow-up PR.

jwst/regtest/sdp_pools_source.py
```diff
@@ -2,10 +2,12 @@
...
 # SDP pool lists for test parametrization.
 # #############################################################
-from jwst.tests.base_classes import BaseJWSTTest
+import os
 
+from jwst.regtest.regtestdata import RegtestData
 
-class SDPPoolsSource(BaseJWSTTest):
+
+class SDPPoolsSource(RegtestData):
```
jdavies-st, line 10: Glad to see `BaseJWSTTest` dropped.  Can we now delete that class and maybe that whole module?

scripts/okify_regtests
```diff
@@ -46,6 +47,53 @@ def artifactory_copy(specfile, dry_run=False):
...
+    # the spec to perform the delete.
+    with open(specfile) as fh:
+        spec = json.load(fh)
+    pattern = spec['files'][0]['pattern'] + '/'
+    target = spec['files'][0]['target']
+
+    # Remove the target
+    folder = target + Path(pattern).stem
+    args = ['jfrog', 'rt', 'del', folder, '--quiet=true'] + jfrog_args
+    subprocess.run(args, check=True)
```
jdavies-st, line 66: Why is deleting the folder necessary?  Is it because the number and names of the new files might be different?

jdavies-st (commented):
Looks good.  Some questions below.

jwst/regtest/conftest.py
```diff
@@ -121,18 +123,28 @@ def artifactory_result_path():
...
+                upload_schema = generate_upload_schema(rtdata.output,
+                                                       rtdata.remote_results_path,
+                                                       schema=upload_schema)
+            elif rtdata.okify_op == 'folder_copy':
+                output = rtdata.output + '/'
+                target = rtdata.remote_results_path + os.path.basename(rtdata.output) + '/'
+                upload_schema = generate_upload_schema(output, target, schema=upload_schema)
+            else:
+                raise RuntimeError(f'Unknown artifactory operation: {rtdata.okify_op}')
+
```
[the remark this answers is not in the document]
stscieisenhamer: ✅

jwst/regtest/regtestdata.py
stscieisenhamer: ✅

jwst/regtest/sdp_pools_source.py
stscieisenhamer: ✅

scripts/okify_regtests
stscieisenhamer: Exactly. Any rule change, or even other code changes, can produce a completely new set of files, sometimes more, sometimes less, files with the same name but different contents, or the same contents but in different names, etc.

It is easier to remove the folder completely than keep track of files. The file-based operations, currently, do not add missing files nor does it delete truth files if non-existent. This may be a feature to add, but not in this PR. The intent is exactly this: If anything is different in a folder, remove and rebuild the folder.

jwst/regtest/conftest.py
```diff
@@ -196,6 +209,18 @@ def rtdata_module(artifactory_repos, envopt, request, jail):
     return _rtdata_fixture_implementation(artifactory_repos, envopt, request)
 
 
+def _sdpdata_fixture_implementation(artifactory_repos, envopt, request):
+    """Provides the RemoteResource class"""
+    inputs_root, results_root = artifactory_repos
+    return SDPPoolsSource(env=envopt, inputs_root=inputs_root,
+                          results_root=results_root)
```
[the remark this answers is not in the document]
stscieisenhamer: The generic part of okify-ing files vs. folder is a generic part of `RegtestData`. 

However, the need for `SDPPoolsSource` is because of the need to use a common set of data between both test construction, see `conftest.pytest_generatetests`, and the test itself. There may be way of generalizing that but that will take some more pytest-foo. Regardless, I'll update some docstrings to make this clear. And maybe lightning will strike.

jwst/regtest/conftest.py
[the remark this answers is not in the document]
stscieisenhamer: No, it doesn't overwrite. The issue is that the "targets" of the okifying files is different than the target of the actual output. Hence the schema snippets need to be different. The snippets are actually appended together, through the use of the `schema=schema_to_append` keyword argument.

There are a couple of approaches here that have different ramifications other places, and this one happens to be the least obtrusive.

hbushouse:
pep8 is a cruel task master

stscieisenhamer:
OK, things are back to working. @jdavies-st, as noted in above commentary, some things were left as-is. If there are still concerns, offline discussion will need be had; further work will need a mind-meld.

jwst/regtest/regtestdata.py
```diff
@@ -253,11 +248,6 @@ def to_asdf(self, path):
         af = asdf.AsdfFile(tree=tree)
         af.write_to(path)
 
-    @classmethod
-    def open(cls, filename):
-        with asdf.open(filename) as af:
-            return cls(**af.tree)
-
```
jdavies-st, line 260: Is there a reason `open()` is removed?  I think this is necessary to parse the `RegtestData` object that gets serialized to ASDF when okifying.

jdavies-st (commented):
Looks good.  Some questions below.

Also, it would be nice to delete the `jwst.tests.base_classes` module and move whatever is there to a better location.  Not necessary in this PR.

jwst/regtest/regtestdata.py
stscieisenhamer: was lost in the undoing of the code rearrangement. that operation turned out to be much nastier than it should have been.

stscieisenhamer: ✅

jdavies-st (approved):
Looks good!

PR #6228, opened 2021-07-19
JP-1932: Update DARK model schemas to use uint32 for DQ

### LINKED ISSUE #5775: Inconsistent data types in dark reference file DQ extension
_Issue [JP-1932](https://jira.stsci.edu/browse/JP-1932) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

Looking over the schema for the dark current reference file, it appears that the data type associated with the DQ extensions is inconsistent with that of other reference files. In the mask reference file the data are 32-bit unsigned integers. But in the DQ extension of the dark current reference file, the data are 16-bit unsigned integers. When following the standard dq value definitions, pixels flagged as UNRELIABLE_DARK should have a value of 2^23, which is larger than a 16-bit integer can contain. In this case, any pixels flagged as UNRELIABLE_DARK in the input array when creating a reference file would have their values reassigned when the file is saved.

I've looked through the schema for many of the other reference file types, and in all other cases, the DQ data are defined as 32-bit integers.

hbushouse:
Closes #5775

Resolves [JP-1932](https://jira.stsci.edu/browse/JP-1932)

**Description**

This PR updates the DQ array definition in the DARK ref file data model schema to be uint32, instead of the old uint16. This model apparently got missed when all the others were updated to uint32.

Checklist
- [ ] ~~Tests~~
- [ ] ~~Documentation~~
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,8 @@ datamodels
 
 - Add NOUTPUTS keyword to the `DarkModel` schema. [#6213]
 
+- Update `DarkModel` to use uint32 for DQ array. [#6228]
+
 extract_1d
 ----------
 
--- a/jwst/datamodels/dark.py
+++ b/jwst/datamodels/dark.py
@@ -15,7 +15,7 @@ class DarkModel(ReferenceFileModel):
     data : numpy float32 array
          Dark current array
 
-    dq : numpy uint16 array
+    dq : numpy uint32 array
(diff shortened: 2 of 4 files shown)


stscirij (approved):
LGTM

PR #6233, opened 2021-07-22
JP-2211: Fix wfss_contam handling of filter names

### LINKED ISSUE #6231: Use PUPIL value for filter name in NIRISS wfss_contam processing
_Issue [JP-2211](https://jira.stsci.edu/browse/JP-2211) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The wfss_contam step uses the value of the FILTER keyword to make selections from the wavelengthrange and photom ref files. This works fine for NIRCam WFSS mode, which has gratings in the PUPIL wheel and cross-filters in the FILTER wheel. NIRISS, however, has the gratings in the FILTER wheel and cross-filter in the PUPIL wheel, so the use of those two keyword values needs to be swapped in the code when processing NIRISS WFSS grism exposures.

hbushouse:
Closes #6231 

Resolves [JP-2211](https://jira.stsci.edu/browse/JP-2211)

**Description**

This PR fixes a bug in the `wfss_contam` step when processing NIRISS WFSS grism exposures. The filters used in NIRISS WFSS are physically located in the PUPIL wheel, so you have to use the PUPIL keyword value for the filter name everywhere in the step code where the filter name is needed.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -66,6 +66,8 @@ wfss_contam
 
 - Added step documentation [#6210]
 
+- Fixed handling of filter/pupil names for NIRISS WFSS mode [#6233]
+
 1.2.3 (2021-06-08)
 ==================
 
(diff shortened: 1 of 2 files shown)


nden (approved):
LGTM

PR #6234, opened 2021-07-23
Jump flags all pixels for MIRI 3 & 4 group ints

### LINKED ISSUE #6223: Jump step on 3 and 4 group data seeming to flag nearly all pixels as jumps
_Issue [JP-2206](https://jira.stsci.edu/browse/JP-2206) was created on JIRA by [Misty Cracraft](https://jira.stsci.edu/secure/ViewProfile.jspa?name=cracraft):_

As part of testing the updates to jump step that apply the correction to 3 and 4 group data, I created two new simulations with MIRISim, one with 3 groups and 2 ints, and the other with 4 groups and 2 ints. When I ran them through the pipeline (using a simulator specific readnoise file, skipping refpix and using a threshold of 10), the output rate images seem to have nearly every pixel flagged as 4 (jump detected).

Command used:

$ strun calwebb_detector1 det_image_seq1_MIRIMAGE_F1130Wexp1_4groups.fits --steps.jump.override_readnoise='███████████████████████████████████████████████████ --steps.ramp_fit.override_readnoise='███████████████████████████████████████████████████ --steps.refpix.skip=True --steps.jump.rejection_threshold=10

I'm not sure what happened with this, but I'm quite sure we don't want all pixels flagged. The uncal (*groups.fits) files are also in the specified directory, so the input and output files are available for examination.

 

After confirming that this is a bug, I re-set the team priority to critical.

This update was meant to allow jump detection with 3 and 4 group data. As it stands currently, all pixels are being flagged as jumps. This will prevent ramp fitting happening accurately in any data with 3 and 4 group data. For any observations with these small numbers of groups to give accurate results, this bug needs to be fixed.

mwregan2:
Closes #6223 

Resolves JP-2206

This PR addresses the problem that when the MIRI first frame is flagged as DO_NOT_USE the jump step would flag all the pixels in three and four group integrations. This fix also address a bug that caused it flag single good group pixels under certain conditions. Both problems were recreated with new tests.

Checklist
- [ ] Tests

- [ ] Documentation

- [ ] Change log

- [ ] Milestone

- [ ] Label(s)

### DIFF
--- a/jwst/jump/jump.py
+++ b/jwst/jump/jump.py
@@ -73,11 +73,13 @@ def detect_jumps(input_model, gain_model, readnoise_model,
     if len(wh_g[0] > 0):
         pdq[wh_g] = np.bitwise_or(pdq[wh_g], dqflags.pixel['NO_GAIN_VALUE'])
         pdq[wh_g] = np.bitwise_or(pdq[wh_g], dqflags.pixel['DO_NOT_USE'])
+        gain_2d[wh_g] == 0
 
     wh_g = np.where(np.isnan(gain_2d))
     if len(wh_g[0] > 0):
         pdq[wh_g] = np.bitwise_or(pdq[wh_g], dqflags.pixel['NO_GAIN_VALUE'])
         pdq[wh_g] = np.bitwise_or(pdq[wh_g], dqflags.pixel['DO_NOT_USE'])
+        gain_2d[wh_g] = 0.0  # a pixel with all NaNs messes up the two point difference step
 
     # Apply gain to the SCI, ERR, and readnoise arrays so they're in units
     # of electrons
--- a/jwst/jump/twopoint_difference.py
+++ b/jwst/jump/twopoint_difference.py
@@ -75,8 +75,10 @@ def find_crs(data, group_dq, read_noise, normal_rejection_threshold,
 
         # Make all the first diffs for saturated groups be equal to
         # 100,000 to put them above the good values in the sorted index
+        max_value_index = np.nanargmax(positive_first_diffs, axis=2)
         first_diffs[np.isnan(first_diffs)] = 100000.
-
+#        not_sat = first_diffs < 100000.
+#        max_value_index = np.where(not_sat.any(1), )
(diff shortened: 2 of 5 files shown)


jwst/jump/jump.py
```diff
@@ -73,11 +73,13 @@ def detect_jumps(input_model, gain_model, readnoise_model,
     if len(wh_g[0] > 0):
         pdq[wh_g] = np.bitwise_or(pdq[wh_g], dqflags.pixel['NO_GAIN_VALUE'])
         pdq[wh_g] = np.bitwise_or(pdq[wh_g], dqflags.pixel['DO_NOT_USE'])
+        gain_2d[wh_g] == 0
```
hbushouse, line 76: Should be just a single "=" sign here, right? You want to set the values to zero, not evaluate them.

jwst/jump/twopoint_difference.py
```diff
@@ -105,20 +107,11 @@ def find_crs(data, group_dq, read_noise, normal_rejection_threshold,
...
-        # Extract from the sorted group indices the index of the largest non-saturated group.
+        # Get the row and column indices of pixels whose largest non-saturated ratio is above the threshold
         row, col = np.where(number_sat_groups >= 0)
-        max_index1d = sort_index[row, col, max_value_index[row, col]]
-        max_index1 = np.reshape(max_index1d, (nrows, ncols))  # reshape to a 2-D array
-        max_ratio2d = np.reshape(ratio3d[row, col, max_index1[row, col]], (nrows, ncols))
-        max_index1d = sort_index[row, col, 1]
-        max_index2d = np.reshape(max_index1d, (nrows, ncols))
-        last_ratio = np.reshape(ratio3d[row, col, max_index2d[row, col]], (nrows, ncols))
+        max_ratio2d = np. reshape(ratio3d[row, col, max_value_index[row, col]], (nrows, ncols))
```
hbushouse, line 113: extra white space in "np. reshape"

hbushouse (commented):
Couple of small fixes needed. Also need to rebase to resolve conflicts with change log (which was recently updated for the 1.3.0 release)

PR #6243, opened 2021-07-27
JP-2103 : Update slit.source_x(y)pos (NRS FS mode) after computing them in wavecorr

### LINKED ISSUE #6074: The source_xpos parameter not being calculated for FS data?
_Issue [JP-2103](https://jira.stsci.edu/browse/JP-2103) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

In processing a set of FS simulated dithered exposures, which included both primary and spectral dithers, both the source_xpos and source_ypos parameters are zero for all dither positions.  The former in particular should be nonzero for dither positions in which the source is not at the center of the slit in the dispersion direction, which affects the wavecorr step.  In this case, no wavelength corrections are applied even in cases where they should be.

nden:
Closes #6074 

Resolves [JP-2103](https://jira.stsci.edu/browse/JP-2103)

**Description**

In Nirspec FS mode the wavecorr step computes the position of the source in 
the aperture using `meta.dither.x_offset` and `meta.dither.y_offset` but did not update `slit.source_xpos` and `slit.source_ypos`. This PR updates the two attributes with the computed [position of the source in the slit.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -59,6 +59,12 @@ source_catalog
 
 - Fixed issue with non-finite positions for aperture photometry. [#6206]
 
+wavecorr
+--------
+
+- Location of source in NIRspec fixed slit updated
+  (keywords ``SCRCXPOS``, ``SRCYPOS``). [#6243]
+
 wfss_contam
 -----------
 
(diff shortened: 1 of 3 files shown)


hbushouse (approved):
LGTM

Issue #6105, opened 2021-06-03
Increase test coverage of extract_1d


stscijgbot-jp:
_Issue [JP-2120](https://jira.stsci.edu/browse/JP-2120) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

Current test coverage of extract_1d is 77.01 %. 

In preparation for refactoring extract_1d add tests to modules/functions that are currently not tested.

 
||•||•||•||Coverage||
|[ *_init_*.py](https://codecov.io/gh/spacetelescope/jwst/src/master/jwst/extract_1d/*_init_*.py)|2|2|0|0|100.00%|
|[ apply_apcorr.py](https://codecov.io/gh/spacetelescope/jwst/src/master/jwst/extract_1d/apply_apcorr.py)|165|158|0|7|95.76%|
|[ extract.py](https://codecov.io/gh/spacetelescope/jwst/src/master/jwst/extract_1d/extract.py)|1,595|1,046|0|549|65.58%|
|[ extract1d.py](https://codecov.io/gh/spacetelescope/jwst/src/master/jwst/extract_1d/extract1d.py)|271|215|0|56|79.34%|
|[ extract_1d_step.py](https://codecov.io/gh/spacetelescope/jwst/src/master/jwst/extract_1d/extract_1d_step.py)|92|74|0|18|80.43%|
|[ ifu.py](https://codecov.io/gh/spacetelescope/jwst/src/master/jwst/extract_1d/ifu.py)|591|465|0|126|78.68%|
|[ spec_wcs.py](https://codecov.io/gh/spacetelescope/jwst/src/master/jwst/extract_1d/spec_wcs.py)|24|24|0|0|100.00%|

PR #6246, opened 2021-07-27
JP-2120: Increase test coverage for ImageModel through Extract1dStep

tapastro:
Addresses (but does not close) #6105 

Addresses [JP-2120](https://jira.stsci.edu/browse/JP-2120)

**Description**

This PR adds test coverage for ImageModel inputs to Extract1dStep. **WIP**

Checklist
- [x] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -27,6 +27,8 @@ from . import ifu
 from . import spec_wcs
 from .apply_apcorr import select_apcorr
 
+from json.decoder import JSONDecodeError
+
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):
Even though this new test only covers ImageModel as used for the input science exposure, and not an ImageModel used as a reference file, it's still a worthwhile addition to the test suite, because it's quite possible that some users will not want to resample MIRI LRS slit spectra.

tapastro:
Had to add/modify some of the code in extract - some of the duplicate methods present no longer had equal lengths of arguments, so a dummy was added. The error statement for attempting to open a FITS with json.load may have also changed, so the JSONDecodeError was imported for that. Otherwise, new rtdata uploaded to artifcatory and we'll see if we get some better coverage...

PR #6251, opened 2021-07-28
JP-2152: Refactor resample_variance_array to use less memory

### LINKED ISSUE #6154: Extreme memory consumption in Image3Pipeline.resample
_Issue [JP-2152](https://jira.stsci.edu/browse/JP-2152) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

In DMS, `calwebb_image3` crashes with memory issues in `resample`; see attached log. The memory footprint at time of failure is ~65GB.

The command to use is

`strun calwebb_image3 jw00624-o029_20210610t121017_image3_003_asn.json`

When running on a local macos laptop, the processing completes, but the following behavior is seen:
 * for nearly all processing, the process memory is between 10-20GB
 * in `resample`, during the drizzling, as noted by the output, the memory usage does not significantly change.
 * However, after the drizzling logging finishes, and no other output occurs, the memory usage starts rising dramatically, topping out ~75GB
 * processing then proceeds with memory dropping back to the nominal ~10GB
 * This pattern repeats for each of the data arrays that get drizzled, including var_poisson, var_rnoise, etc.

Implication that something, after drizzle completes for each array, consumes massive amount of memory. Presumption is that this is at the failure line: resample.py:219

jdavies-st:
Resolves #6154 / [JP-2152](https://jira.stsci.edu/browse/JP-2152)

**Description**

This PR fixes the extreme memory consumption seen in resampling of variance arrays.  For the dataset that triggered this issue, memory consumption goes from ~29GB down to ~1.6GB.

This also fixes a bug in how the the resampled variances were combined previously.  Previously the combined variances were being over-estimated.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -1,6 +1,5 @@
 import logging
 import re
-import warnings
 
 import numpy as np
 from drizzle import util
@@ -188,48 +187,40 @@ class ResampleData:
     def resample_variance_array(self, name, output_model):
         """Resample variance arrays from self.input_models to the output_model
 
-        Loop through self.input_models and resample the `name` variance array
-        to the same name in output_model.  This modifies output_model in-place.
+        Resample the `name` variance array to the same name in output_model,
+        using a cummulative sum instead of weighted average.
+
+        This modifies output_model in-place.
         """
+        output_wcs = output_model.meta.wcs
+        output_array = np.zeros_like(output_model.data)
         outwht = np.zeros_like(output_model.data)
         outcon = np.zeros_like(output_model.con)
-        output_wcs = output_model.meta.wcs
-        output_arrays = []
 
         log.info(f"Resampling {name}")
+        uniqid = 1
         for model in self.input_models:
-            # Unity weight but ignore pixels that are NON_SCIENCE or REFERENCE_PIXEL
-            inwht = resample_utils.build_driz_weight(model, weight_type=None,
-                                                     good_bits="~NON_SCIENCE+REFERENCE_PIXEL")
-            input_array = getattr(model, name)
-            output_array = np.zeros_like(output_model.data)
+            with np.errstate(divide="ignore"):
+                inv_var = np.reciprocal(getattr(model, name))
+            inwht = np.ones_like(model.data)
 
-            # Resample the variance array
-            self.drizzle_arrays(input_array, inwht, model.meta.wcs,
+            # Resample the inverse variance array
+            self.drizzle_arrays(inv_var, inwht, model.meta.wcs,
                                 output_wcs, output_array, outwht, outcon,
                                 pixfrac=self.pixfrac, kernel=self.kernel,
-                                fillval=np.nan)
-            output_arrays.append(output_array)
+                                fillval=np.nan, uniqid=uniqid)
+            uniqid += 1
 
-        # Stack and coadd as inverse variance if there any to stack
-        if len(output_arrays) > 1:
-            stacked = np.stack(output_arrays)
-            with np.errstate(divide="ignore"):
-                inv_var = np.reciprocal(stacked)
-            inv_var[~np.isfinite(inv_var)] = np.nan
+        # Drizzle produces a weighted average of the new image with the
+        # existing cumulative image.  But we want a weighted sum.  We have
+        # specified all the input weights to be unity above, which means to get
+        # a sum, we just multiply the final cummulative weighted average image
+        # by the total cummulative weights to get the sum.
+        output_array *= outwht
 
-            with warnings.catch_warnings():
-                warnings.filterwarnings(action="ignore",
-                                        message="Mean of empty slice",
-                                        category=RuntimeWarning)
-                inv_var_mean = np.nanmean(inv_var, axis=0)
-            with np.errstate(divide="ignore"):
-                output_err_array = np.reciprocal(inv_var_mean)
-        else:
-            output_err_array = output_arrays[0]
-        output_err_array[~np.isfinite(output_err_array)] = np.nan
-
-        setattr(output_model, name, output_err_array)
+        output_variance = np.reciprocal(output_array)
+        output_variance[~np.isfinite(output_variance)] = np.nan
+        setattr(output_model, name, output_variance)
 
     def update_exposure_times(self, output_model):
         """Modify exposure time metadata in-place"""
@@ -249,7 +240,7 @@ class ResampleData:
     @staticmethod
     def drizzle_arrays(insci, inwht, input_wcs, output_wcs, outsci, outwht, outcon,
                        uniqid=1, xmin=None, xmax=None, ymin=None, ymax=None,
-                       pixfrac=1.0, kernel='square', fillval="INDEF"):
+                       pixfrac=1.0, kernel='square', fillval="INDEF", wtscale=1.0):
         """
         Low level routine for performing 'drizzle' operation on one image.
 
@@ -408,7 +399,7 @@ class ResampleData:
             kernel=kernel,
             in_units="cps",
             expscale=1.0,
-            wtscale=1.0,
+            wtscale=wtscale,
             fillstr=fillval
         )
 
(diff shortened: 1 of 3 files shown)


jdavies-st:
Unit test coverage percentage is lower because this PR deletes a bunch of well-covered code, leaving behind less well-covered code.

hbushouse (approved):
LGTM

mcara:
Are there any unit or regression tests that cover this?

mcara (changes requested):
Drizzling input images and then adding reciprocals of drizzled images is not the same (mathematically) as adding drizzled images of reciprocals of input images.

In my testing, I wrote out variance (error) arrays produced by resample function when running  the `test_miri_image.py::test_miri_image3_catalog` test. The difference in the values of error arrays is large, i.e., a typical value is 1.6e-6 (master) and 1e-6 (this PR). Maybe something else is contributing to this difference as well.

It is not necessary to collect all drizzled images to compute variance array in the original algorithm (in order to reduce the memory).

jdavies-st:
> Drizzling input images and then adding reciprocals of drizzled images is not the same (mathematically) as adding drizzled images of reciprocals of input images.

Ah!  Thanks for catching this @mcara.  So which is the correct way to do it?  The way I originally implemented it, or this?  Thoughts?

The idea here is to use standard uncertainty propagation, so add the inverse variances to get the final.  But I suspect the difference is when the input image is drizzled, the distribution of values in the drizzle process amongst adjacent pixels is mathematically different as you suggest above.

mcara:
> The idea here is to use standard uncertainty propagation, so add the inverse variances to get the final.

This doesn't sound right to me, at least not for drizzled images. Why _inverse_ variance?

> So which is the correct way to do it? The way I originally implemented it, or this?

Probably neither. Also, I can't understand how error propagation (using drizzling) works for "translating" (converting) variances of input images to the output (resampled) grid. Even if it is true (=possible), it must be for a very limited, special, theoretical situation. That is, I have serious doubts, at least at first glance, about correctness of resampling variances of input images.

Second issue is the variance of combined drizzled images. Assuming d = d1 + d2 + d3 + ... (i.e., total drizzled image is a sum of "single sci" images, using HST nomenclature), then I would think v = v1 + v2 + v3 + ... [See the formula for uncorrelated variances in https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Linear_combinations for a more general linear combination case]. Again, where does _inverse_ come from?

Ignoring my first objection (resampling variance array to get variance of resampled image), thinking about the original question ("which is the correct way to do it...?"), I would say the first way was "better" than this PR: at least it was not drizzling _inverse_ variances **but** it was computing _harmonic mean_ instead of the _arithmetic mean_.

mcara:
I also think the code needs a regression test that can test created variance arrays against some "truth" before another PR is made. It really helps because it allows easy checking that new algorithm is equivalent to the original one.

jdavies-st:
> > The idea here is to use standard uncertainty propagation, so add the inverse variances to get the final.
> 
> This doesn't sound right to me, at least not for drizzled images. Why _inverse_ variance?

This is what @karllark and I had already discussed as the close-to-correct way to do this when originally implementing this, not taking into account correlated noise, which is being ignored in the baseline implementation.  The reason here is that each pixel in an output _i2d drizzled image is a weighted _average_ of the input pixels.  And we are weighting by inverse variance (noise, sigma squared).  So in this special case of weighting by the inverse variance, the variance in the output is computed as the reciprocal of the sum of the weights (where the weights are inv var).  See Taylor "Error Analysis" pg 176

IMG_8791

And of course in resampled case here, the weights are the inverse variance modulated by the fraction pixel factor in drizzling, so there's correlation, which we are not taking into account right now, but hope to in the future.



> Second issue is the variance of combined drizzled images. Assuming d = d1 + d2 + d3 + ... (i.e., total drizzled image is a sum of "single sci" images, using HST nomenclature), then I would think v = v1 + v2 + v3 + ... [See the formula for uncorrelated variances in https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Linear_combinations for a more general linear combination case]. Again, where does _inverse_ come from?

See above.

> Ignoring my first objection (resampling variance array to get variance of resampled image), thinking about the original question ("which is the correct way to do it...?"), I would say the first way was "better" than this PR: at least it was not drizzling _inverse_ variances **but** it was computing _harmonic mean_ instead of the _arithmetic mean_.

Agree.  Will fix.  I.e. do it the way it was originally done, but more efficiently.

jdavies-st:
> I also think the code first needs a regression test that can test created variance arrays against some "truth" before another PR is made.

There's existing regression tests for imaging and spectra.  Will report here on results once I've fixed this PR.

jdavies-st:
So to be clear here, in the notes in #5798 we discussed, "drizzle 1 / err**2 with unity weighting", but in fact, what was implemented in #6036 (which I believe is correct) is that we drizzle the variance with unity weighting, then for level 3 coadded images, the final variance in the coadded _i2d.fits image is `var_f = 1 / sum(1 / var_i)` with unity weighting on each _input_ variance image, which of course means that the output resampled variance image has unity weighting on average where there is coverage, but pixel-to-pixel may deviate from unity weighting because input pixels do not map exactly to output pixels.

mcara (approved):

mcara:
@jdavies-st Could you add a regression test with a reference ("truth") being the original result? It would be helpful for this and future variance-related PRs but also to make sure that this part of the code is working as expected.

jwst/resample/resample.py
```diff
@@ -188,48 +187,41 @@ def resample_many_to_one(self):
...
+            # Add the inverse of the resampled variance to a running sum
             with np.errstate(divide="ignore"):
-                inv_var = np.reciprocal(stacked)
-            inv_var[~np.isfinite(inv_var)] = np.nan
-
-            with warnings.catch_warnings():
-                warnings.filterwarnings(action="ignore",
-                                        message="Mean of empty slice",
-                                        category=RuntimeWarning)
-                inv_var_mean = np.nanmean(inv_var, axis=0)
```
jdavies-st, line 225: So this line is a bug in the original code.  Instead of of summing the inverse variances, I was taking the mean along the stack with no reference to number of images in the overlapping regions.  Which means the final error array was overestimated.  Ack.

So in our regression test for NIRCam imaging with 3 dithered pointings.  The error array in the resampled image was computed as the following for regions with 1, 2 and 3 overlapping pointings, respectively:

```
In [10]: truth.err[2250:2300,3400:3450].mean()
Out[10]: 0.024356956

In [12]: truth.err[2100:2150,3400:3450].mean()
Out[12]: 0.019771107

In [16]: truth.err[1850:1900,3400:3450].mean()
Out[16]: 0.017140672
```

This PR correctly sums the inverse variances, so that for similar variances, the combined variance of N images shows a reduction in sigma by `sqrt(N)` where there are N overlapping resampled images.

I.e. for the same regions above:

```
In [9]: result.err[2250:2300,3400:3450].mean()
Out[9]: 0.024356956

In [11]: result.err[2100:2150,3400:3450].mean()
Out[11]: 0.01714876

In [15]: result.err[1850:1900,3400:3450].mean()
Out[15]: 0.01403387
```

And it scales as ~`1 / sqrt(N)` as expected:

```
In [17]: 0.024356956 / np.sqrt(2)
Out[17]: 0.017222968756662363

In [18]: 0.024356956 / np.sqrt(3)
Out[18]: 0.014062495103239871
```

jdavies-st:
Finally, with tweakreg turned off, the code in this PR at commit 41ed771 uses 1.6 GB of memory compared to ~29GB with `master` for the dataset in JP-2152.

jdavies-st:
> I also think the code first needs a regression test that can test created variance arrays against some "truth" before another PR is made. It really helps because it allows easy checking that new algorithm is equivalent to the original one.

I've added a unit test.

mcara:
Thanks!

jwst/resample/resample.py
jdavies-st: I've put the above into a unit test.

PR #6252, opened 2021-07-28
JP-2152: Add an upper tweak threshold to tweakreg

### LINKED ISSUE #6154: Extreme memory consumption in Image3Pipeline.resample

jdavies-st:
Resolves #6154 / [JP-2152](https://jira.stsci.edu/browse/JP-2152)

**Description**

This PR adds an upper tweak threshold to `TweakregStep`.  I.e. any tweaks larger than 10 arcsec are probably due to poor data, poor catalogs, or both, and should be rejected.

This hard codes the threshold to be 10 arcsec.  But honestly, we could also use the existing `self.threshold` parameter in tweakreg of 1.0 arcsec as well.  Thoughts?  I don't want to add a separate parameter yet, as I don't think this is something users need to play with right now.  There are already too many knobs in this step to play with.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/tweakreg/tweakreg_step.py
+++ b/jwst/tweakreg/tweakreg_step.py
@@ -7,6 +7,9 @@ JWST pipeline step for image alignment.
 from os import path
 
 from astropy.table import Table
+from astropy import units as u
+from astropy.coordinates import SkyCoord
+from gwcs.wcstools import grid_from_bounding_box
 from tweakwcs.imalign import align_wcs
 from tweakwcs.tpwcs import JWSTgWCS
 from tweakwcs.matchutils import TPMatch
@@ -207,6 +210,15 @@ class TweakRegStep(Step):
             else:
                 raise e
 
+        for imcat in imcats:
+            if not self.fit_quality_is_good(imcat):
+                self.log.warning("WCS has been tweaked by more than 10 arcsec")
+                self.log.warning("Skipping 'TweakRegStep'...")
+                self.skip = True
+                for model in images:
+                    model.meta.cal_step.tweakreg = "SKIPPED"
+                return images
+
         if self.align_to_gaia:
             # Get catalog of GAIA sources for the field
             #
@@ -299,6 +311,28 @@ class TweakRegStep(Step):
                 """
 
         return images
+
+    def fit_quality_is_good(self, imcat):
+        """Check that the newly tweaked wcs hasn't gone off the rails"""
+        tolerance = 10.0 * u.arcsec
+
+        wcs = imcat.meta['image_model'].meta.wcs
+        twcs = imcat.wcs
+
+        grid = grid_from_bounding_box(wcs.bounding_box, step=100)
+        # There's a bug in grid_from_bounding_box that makes the final grid
+        # item outside the bounding box.  So trim off the last one.  And trim
+        # the first one to avoid edge effects.
+        grid = grid[:,1:-1,1:-1]
+        ra, dec = wcs(*grid)
+        tra, tdec = twcs(*grid)
+
+        skycoord = SkyCoord(ra=ra, dec=dec, unit="deg")
+        tskycoord = SkyCoord(ra=tra, dec=tdec, unit="deg")
+
+        separation = skycoord.separation(tskycoord)
+
+        return (separation < tolerance).all()
 
     def _imodel2wcsim(self, image_model):
         # make sure that we have a catalog:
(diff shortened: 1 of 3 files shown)


hbushouse (approved):
Code looks good. In terms of how to implement the upper threshold, if you don't feel well about hard coding the 10 arcsec limit, but at the same time don't want to add yet another step param, what about doing something like 10 * `self.threshold`?

jwst/tweakreg/tweakreg_step.py
```diff
@@ -300,6 +312,28 @@ def process(self, input):
...
+        tolerance = 10.0 * u.arcsec
+
+        wcs = imcat.meta['image_model'].meta.wcs
+        twcs = imcat.wcs
+
+        grid = grid_from_bounding_box(wcs.bounding_box, step=100)
+        # There's a bug in grid_from_bounding_box that makes the final grid
+        # item outside the bounding box.  So trim off the last one.  And trim
+        # the first one to avoid edge effects.
+        grid = grid[:,1:-1,1:-1]
```
mcara, line 326: I think this is an overkill. Using the corners of the bounding box should be enough: these are extreme points of the image and should reveal large shift, scale, and/or rotation.

mcara (approved):
I approve this PR because it should do what it is supposed to do. However, for performance reasons, I highly recommend simplifying the grid to just 4 corners of the bounding box. I would be surprised if inner pixels can have larger separation than the largest separation of the 4 corners.

jwst/tweakreg/tweakreg_step.py
jdavies-st: Great feedback.  `wcs.footprint().T` gives just the bounding box corners in RA/Dec, so I'll just use those.

jdavies-st:
I used `10 * self.tolerance` as you suggested @hbushouse .  We can reduce the multiplicative factor in the future if needed.

I changed from using the a grid of points to just the corners, i.e. `wcs.footprint().T` as you suggested @mcara .  This is much better and clearer.

Finally I added a unit test to make sure the function is working as expected, and doing so made me change the calling signature to be a bit more test-friendly.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -312,21 +313,12 @@ def process(self, input):
...
-        # There's a bug in grid_from_bounding_box that makes the final grid
-        # item outside the bounding box.  So trim off the last one.  And trim
-        # the first one to avoid edge effects.
-        grid = grid[:,1:-1,1:-1]
-        ra, dec = wcs(*grid)
-        tra, tdec = twcs(*grid)
+        tolerance = 10.0 * self.tolerance * u.arcsec
 
+        ra, dec = wcs.footprint(axis_type="spatial").T
+        tra, tdec = twcs.footprint(axis_type="spatial").T
```
nden, line 321: I think this would be the right approach but is it expected to work with spectral resampling? The Nirspec slits will produce NaNs as a footprint. Somewhere in the jwst code there might be a function to deal with this but I'm not sure and will have to look fo rit.

hbushouse: tweakreg isn't used for spectra, only images

nden: Duh!

PR #6254, opened 2021-07-29
Changing ramp fitting multiprocessing tests

kmacdonald-stsci:
Closes #

Resolves JP-

**Description**

This PR addresses the ramp fitting multiprocessing tests.  These were skipped because the old automated test system didn't support multiprocessing.  The automated test system now runs on github actions.  The multiprocessing tests have been modified to properly work with the new refactored interface of ramp fitting.

Checklist
- [x] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/ramp_fitting/tests/test_ramp_fit.py
+++ b/jwst/ramp_fitting/tests/test_ramp_fit.py
@@ -125,7 +125,7 @@ def test_gls_vs_ols_two_ints_ols():
     np.testing.assert_allclose(slopes_gls[0].data[50, 50], 150.0, 1e-6)
 
 
-@pytest.mark.skip(reason="Jenkins environment does not correctly handle multi-processing.")
+# @pytest.mark.skip(reason="Jenkins environment does not correctly handle multi-processing.")
 def test_multiprocessing():
     nints, ngroups, nrows = 3, 25, 100
     ncols = nrows  # make sure these are the same, so the loops below work
@@ -147,6 +147,10 @@ def test_multiprocessing():
         model1.data[2, j + 1, :, :] = model1.data[2, j, :, :] + delta_plane1 + delta_plane2
     model1.data = np.round(model1.data + np.random.normal(0, 5, (nints, ngroups, ncols, nrows)))
 
+    model2 = model1.copy()
+    gain2 = gain.copy()
+    rnoise2 = rnoise.copy()
+
     # TODO change this to be parametrized once GLS gets working.
     algo = "OLS"
     # algo = "GLS"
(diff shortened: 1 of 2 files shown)


jwst/ramp_fitting/tests/test_ramp_fit.py
```diff
@@ -125,7 +125,7 @@ def test_gls_vs_ols_two_ints_ols():
     np.testing.assert_allclose(slopes_gls[0].data[50, 50], 150.0, 1e-6)
 
 
-@pytest.mark.skip(reason="Jenkins environment does not correctly handle multi-processing.")
+# @pytest.mark.skip(reason="Jenkins environment does not correctly handle multi-processing.")
```
hbushouse, line 128: So does the Jenkins environment now handle multi-processing?

jwst/ramp_fitting/tests/test_ramp_fit.py
```diff
@@ -147,19 +147,23 @@ def test_multiprocessing():
         model1.data[2, j + 1, :, :] = model1.data[2, j, :, :] + delta_plane1 + delta_plane2
     model1.data = np.round(model1.data + np.random.normal(0, 5, (nints, ngroups, ncols, nrows)))
 
+    model2 = model1.copy()
+    gain2 = gain.copy()
+    rnoise2 = rnoise.copy()
```
hbushouse, line 152: Just curious: what's the purpose or need for copies of these?

hbushouse:
Are there any github issues or JP tickets that need to be referenced here?

jwst/ramp_fitting/tests/test_ramp_fit.py
kmacdonald-stsci: I don't think so.  These tests should work in github actions.

jwst/ramp_fitting/tests/test_ramp_fit.py
kmacdonald-stsci: Processing modifies input data, so if these copies are not done, the single processing and multiprocessing will process slightly different data.

jwst/ramp_fitting/tests/test_ramp_fit.py
hbushouse: So if we enable them to run in CI/GA, will they cause a problem when they get run by our nightly regression tests in Jenkins?

kmacdonald-stsci: I don't know if they effect nightly regression tests.  I will find out.  The problem is that since Jenkins didn't actually have multiple processors, these tests weren't actually testing multiprocessing.  These tests passed, but ran on a single processor, giving the false impression multiprocessing got tested and passed.

jwst/ramp_fitting/tests/test_ramp_fit.py
hbushouse: OK, so at least we know that it won't cause any problems/crashes in the Jenkins environment. It just won't give a useful result.

jwst/ramp_fitting/tests/test_ramp_fit.py
jdavies-st: These were originally written by @mwregan2 and were skipped for CI because TravisCI couldn't run them in multiprocessing mode.  Now that we're doing Github Actions, we have 2 cores available, so as long as we use "all" cores, then this will be testing what we want it to test.  On the flip side, when run on the 30-core machine we run our nightly regression tests on under Jenkins, it might also split them over 30 cores.  Which again should be fine, and they should pass there too.

jdavies-st (commented):
This looks good, but the tests are failing.

Does this work with the released version of `stcal 0.2.2`?

kmacdonald-stsci:
No.  These tests do not work with the released STCAL 0.2.2.  That version of STCAL does not have multiprocessing implemented, so returns `NoneType` for ramp fitting when requesting multiprocessing.  These tests require a version of STCAL that is merged with https://github.com/spacetelescope/stcal/pull/30

nden:
@kmacdonald-stsci Point this PR to stcal/master, and revert it once tests pass.
In setup.cfg : 

stcal @ git+https://github.com/spacetelescope/stcal

setup.cfg
```diff
@@ -42,7 +42,7 @@ install_requires =
     requests>=2.22
     scipy>=1.1.0
     spherical-geometry>=1.2.20
-    stcal>=0.2.2
+    stcal@git+https://github.com/spacetelescope/stcal
```
jdavies-st, line 45: stcal 0.2.3 is out now, so let's change this to

```
stcal>=0.2.3
```

After that, ready to merge.

jdavies-st (approved):

jdavies-st:
Tests pass against `stcal 0.2.3` in CI here and separately in regression tests.  Merging.

PR #6286, opened 2021-08-11
JP-2252: Fix jwst.lib.pointing_summary docstring example

### LINKED ISSUE #6285: pointing_summary doctest failing
_Issue [JP-2252](https://jira.stsci.edu/browse/JP-2252) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

In our weekly test against dev dependencies, the doctest in `jwst.lib.pointing_summary` is failing due to a change in `Table.*_repr_*` in astropy dev.

<https://github.com/spacetelescope/jwst/runs/3277416942?check_suite_focus=true#step:6:64>

The documentation example should be modified to not depend on the repr, and instead just print out some column values that are useful.

jdavies-st:
Closes #6285
Resolves [JP-2252](https://jira.stsci.edu/browse/JP-2252)

**Description**

This PR fix `jwst.lib.pointing_summary` docstring example to work under `astropy 5.0`.  `Table.__repr__` has changed between `astropy 4.3` and the dev version of `astropy 5.0`, so avoid using it in the documentation example.

Our weekly regression tests against dev dependencies (`requirements-dev.txt`) will pass once this is merged.

Checklist
- [x] Tests
- [x] Documentation
- [ ] ~Change log~
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/lib/pointing_summary.py
+++ b/jwst/lib/pointing_summary.py
@@ -20,13 +20,17 @@ Examples
         (91.08142005, -66.60547869)>, refpoint=<SkyCoord (ICRS): (ra, dec) in deg
         (90.70377653, -66.59540224)>, delta_v1=<Angle 0.13712727 deg>, delta_refpoint=<Angle 0.04044315 deg>)
 
->>> calc_deltas([im])
-    <Table length=1>
-      exposure                  target                ...    delta_refpoint
-                               deg,deg                ...
-       object                   object                ...       float64
-    ------------ ------------------------------------ ... -------------------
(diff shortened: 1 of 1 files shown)


hbushouse (approved):
The original output was a lot nicer, giving a snapshot of the whole table, but if that style isn't available anymore I guess there's not much we can do about it.

stscieisenhamer (approved):
This is more than sufficient. Previously there was not a check on the actual columns, so that is an improvement. After that, the single value check is fine.

jdavies-st:
Dev dependency job now passes.

https://github.com/spacetelescope/jwst/actions/runs/1124843617

PR #6321, opened 2021-09-02
JP-2012 Bug fix: Change MasterBackgroundNrsSlitsStep name

### LINKED ISSUE #5910: Add aliases to all steps
_Issue [JP-2012](https://jira.stsci.edu/browse/JP-2012) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Currently, alias entry-points are defined for all pipeline classes, so that the alias can be used with `strun`. Do this also for all steps.

Consider creating a utility to create the entry-point list from the code, similar to how the suffix list is created.

tapastro:
Closes #5910

Resolves [JP-2012](https://jira.stsci.edu/browse/JP-2012)

**Description**

This PR addresses issues caused by JP-2012 to the regression tests - namely, that step parameter reference files link to the old naming scheme and refer to the MasterBackgroundNrsSlitsStep as master_background, while this had changed to master_background_nrs so that step_def names could match the aliases introduced. To fix this bug (as well as modify the existing name that no longer applies), this PR clears up some naming discrepancies but also changes MasterBackgroundNrsSlitsStep to a more reasonable MasterBackgroundMosStep.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/docs/jwst/master_background/description.rst
+++ b/docs/jwst/master_background/description.rst
@@ -338,7 +338,7 @@ During :ref:`calwebb_spec2 <calwebb_spec2>` processing, all source and backgroun
 slits are first partially calibrated up through the :ref:`extract_2d <extract_2d_step>`
 and :ref:`srctype <srctype_step>` steps of :ref:`calwebb_spec2 <calwebb_spec2>`,
 which results in 2D cutouts for each slit with the source type identified. At this
-point the `master_background_nrs_slits` step is applied, which is a unique version
+point the :ref:`master_background_mos <master_background_mos_step` step is applied, which is a unique version
 of the step specifically tailored to NIRSpec MOS mode. 
 
 This version of the master background step completes the remaining calibration
@@ -369,7 +369,7 @@ as follows:
 
 1) Process all slitlets in the MOS exposure up through the
    :ref:`extract_2d <extract_2d_step>` and :ref:`srctype <srctype_step>` steps
-2) The `master_background_nrs_slits` step temporarily applies remaining calibration
+2) The :ref:`master_background_mos <master_background_mos_step>` step temporarily applies remaining calibration
    steps up through :ref:`photom <photom_step>` to all slits, treating them all as
    extended sources (appropriate for background signal), and saving the extended
    source correction arrays for each slit in an internal copy of the data model
--- a/docs/jwst/master_background/description.rst
+++ b/docs/jwst/master_background/description.rst
@@ -338,7 +338,7 @@ During :ref:`calwebb_spec2 <calwebb_spec2>` processing, all source and backgroun
 slits are first partially calibrated up through the :ref:`extract_2d <extract_2d_step>`
 and :ref:`srctype <srctype_step>` steps of :ref:`calwebb_spec2 <calwebb_spec2>`,
 which results in 2D cutouts for each slit with the source type identified. At this
-point the :ref:`master_background_mos_step` step is applied, which is a unique version
+point the `master_background_mos_step` step is applied, which is a unique version
 of the step specifically tailored to NIRSpec MOS mode. 
 
 This version of the master background step completes the remaining calibration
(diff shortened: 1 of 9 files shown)


docs/jwst/master_background/description.rst
```diff
@@ -338,7 +338,7 @@ During :ref:`calwebb_spec2 <calwebb_spec2>` processing, all source and backgroun
 slits are first partially calibrated up through the :ref:`extract_2d <extract_2d_step>`
 and :ref:`srctype <srctype_step>` steps of :ref:`calwebb_spec2 <calwebb_spec2>`,
 which results in 2D cutouts for each slit with the source type identified. At this
-point the `master_background_nrs_slits` step is applied, which is a unique version
+point the `master_background_mos_step` step is applied, which is a unique version
```
hbushouse, line 341: don't need to include "_step" in the name here, do we? We didn't include that suffix in the old name.

docs/jwst/master_background/description.rst
```diff
@@ -369,7 +369,7 @@ as follows:
 
 1) Process all slitlets in the MOS exposure up through the
    :ref:`extract_2d <extract_2d_step>` and :ref:`srctype <srctype_step>` steps
-2) The `master_background_nrs_slits` step temporarily applies remaining calibration
+2) The `master_background_mos_step` step temporarily applies remaining calibration
```
hbushouse, line 372: see above

jwst/master_background/master_background_mos_step.py
```diff
@@ -107,7 +107,7 @@ def process(self, data):
             if not self.force_subtract and \
                'COMPLETE' in [data_model.meta.cal_step.back_sub, data_model.meta.cal_step.master_background]:
                 self.log.info('Background subtraction has already occurred. Skipping.')
-                self.record_step_status(data, 'master_background', False)
+                self.record_step_status(data, 'master_background_mos', False)
```
hbushouse, line 110: We only have 1 keyword "S_MSBSUB" to hold the step completion status for both `master_background` and `master_background_mos`. Does that work OK?  Will the `record_step_status` function be able to accept 2 different step names as input to update the value of the same keyword?

hbushouse (commented):
I heartily approve of this name change, because I always thought that `nrs_slits` was too ambiguous - could mean either fixed slits or MOS slits (technically "slitlets").

We will of course want to inform the NIRSpec team in INS of the name change, in case they use it any test scripts or notebooks.

jwst/master_background/master_background_mos_step.py
tapastro: Good catch - I hamfisted the find-and-replace and didn't check results because of regtest/CRDS woes... Running spec2 on some regtest MOS data shows that this leaves S_MSBSUB off the list of calibration step keywords listed in the header.

docs/jwst/master_background/description.rst
tapastro: ✅

hbushouse (approved):

hbushouse:
I assume that merging this PR by itself still won't fix the issues in the regtests, because the pars-spec2pipeline files in CRDS will still be trying to use a different name for this step, right?

tapastro:
> I assume that merging this PR by itself still won't fix the issues in the regtests, because the pars-spec2pipeline files in CRDS will still be trying to use a different name for this step, right?

That's right, the only way to fix it immediately would be to revert the step_def name of master_background_nrs in Spec2Pipeline back to master_background, which we're avoiding due to the alias-step_def connection combined with avoiding duplicate aliases... This only changes the CRDS requirement from master_background_nrs to master_background_mos.

tapastro:
> > I assume that merging this PR by itself still won't fix the issues in the regtests, because the pars-spec2pipeline files in CRDS will still be trying to use a different name for this step, right?
> 
> That's right, the only way to fix it immediately would be to revert the step_def name of master_background_nrs in Spec2Pipeline back to master_background, which we're avoiding due to the alias-step_def connection combined with avoiding duplicate aliases... This only changes the CRDS requirement from master_background_nrs to master_background_mos.

Though this now also requires the change in class name specified in some step-pars files from the old master_background_nrs_slits_step.MasterBackgroundNrsSlitsStep to master_background_mos_step.MasterBackgroundMosStep, e.g. [here](https://jwst-crds.stsci.edu/browse/jwst_niriss_pars-spec2pipeline_0001.asdf)

PR #6324, opened 2021-09-03
JP-2027 & JP-2083: fix resample spec spatial offsets

### LINKED ISSUE #5931: Combined resampled spectra show offset in spatial direction
_Issue [JP-2027](https://jira.stsci.edu/browse/JP-2027) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

As a follow on to JP-1829, fixing that bug revealed that the output WCS frame for combined, resampled spectra is offset spatial from what it should be.  It is the correct width, but with the dataset from that example, it is offset by ~10 pixels because it uses one of the previously-computed mappings from x,y to RA,Dec and doesn't compute a new one.

This is demonstrated in new test (with the relevant part commented out) in PR

#5929

This is a bug in `build_interpolated_output_wcs()`.  That function needs a good refactor.

While this bug is seen in MIRI LRS slit data, it is very likely also in NIRSpec FS and MOS data as well, as the same function is used.  It's just much more visible in MIRI as the slit is much longer than the typical MSA slitlet.

### LINKED ISSUE #6032: resample problem for NIRSpec FS level 3
_Issue [JP-2083](https://jira.stsci.edu/browse/JP-2083) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

The output of level 3 processing of a set of FS simulated exposures does not look as expected.  The data comprise a set of 9 simulated exposures taken with a 3-point primary nod pattern with spectral subdithers.  For this level 3 test, I only included 3 exposures, the ones exactly at the primary nod positions.  In the s2d product, there are single positive and negative source traces, with a second negative trace cut off at the bottom of the 2D image (see first attachment). There are two additional areas of negative flux near the top, separated by wide swaths of pixels with values of 0.  Looking at the CON extension, the coverage map looks funny - the largest area of overlap has a value of 6, while I would have expected 7 (as 3 nodded exposures went into this); there is only a very small region of pixels with a value of 7 at the extreme right end of the 2D image (see second attachment).  There are also disconnected regions with a value of 0, as indicted in the actual spectrum.  As far as I can tell, the coordinate keywords are all populated correctly; the RA and DEC of the source traces in the level 2b products seem to be consistent among the three exposures.

Note: the level 3 association file I used is also in the central store directory, name "spec3_short_asn.json".  (The other spec3 json file is for the entire set of 9 exposures, whose combined product I haven't included.  That one has an even worse problem with the source traces not overlapping, but I'm still checking that the coordinate keywords are correct.)

jdavies-st:
Resolves #5931 / [JP-2027](https://jira.stsci.edu/browse/JP-2027)
Resolves #6032 / [JP-2083](https://jira.stsci.edu/browse/JP-2083)

**Description**

This PR addresses 2 issues seen in resampling NIRSpec spectra where

- Input FS spectra dithered along the spectral axis (orthogonal to slit) end up spatially offset in the resampled output.
- The output footprint of the resampled 2D spectra is the correct size but offset from what it should be.

Some further work is needed to make sure the output resampled 2D spectra has a WCS that round trips.  And tests need to be added.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/datamodels/schemas/slitmeta.schema.yaml
+++ b/jwst/datamodels/schemas/slitmeta.schema.yaml
@@ -107,3 +107,9 @@ properties:
     type: number
     fits_keyword: SRCDEC
     fits_hdu: SCI
+  slit_ymin:
+    title: Bottom of slit in slit_frame relative to (0, 0) center
+    type: number
+  slit_ymax:
+    title: Top of slit in slit_frame relative to (0, 0) center
+    type: number
--- a/jwst/datamodels/schemas/slitmeta.schema.yaml
+++ b/jwst/datamodels/schemas/slitmeta.schema.yaml
@@ -33,6 +33,7 @@ properties:
     title: Slitlet ID
     type: integer
     default: 0
+    
     fits_keyword: SLITID
     fits_hdu: SCI
   shutter_id:
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -4,8 +4,8 @@ import warnings
 import numpy as np
 from astropy import coordinates as coord
 from astropy import units as u
-from astropy.modeling.models import (Mapping, Tabular1D, Linear1D,
-                                     Pix2Sky_TAN, RotateNative2Celestial)
+from astropy.modeling.models import (Mapping, Tabular1D, Linear1D, Shift,
+                                     Pix2Sky_TAN, RotateNative2Celestial, Identity)
 from astropy.modeling.fitting import LinearLSQFitter
 from gwcs import wcstools, WCS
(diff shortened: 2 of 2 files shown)


hbushouse:
@jdavies-st is this PR testable as-is? Nadia suggested that it might be incomplete.

hbushouse:
Given that the changes here are likely superseded by #6747 , can this PR be closed?

mcara:
#6747 replaces this PR

PR #6326, opened 2021-09-03
Nirspec  improvements in blotting for outlier detection, JP-2050 

jemorrison:
Partially resolves [JP-2050](https://jira.stsci.edu/browse/JP-2050)

**Description**

There was a bug in the NIRSpec IFU blotting. An new routine was added to assign_wcs.util.in_ifu_slice that determines which ra, dec, lambda of the median cube fall in a NIRSpec slice. blot_cube_cube.py was updated to use this information

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/cube_build/ifu_cube.py
+++ b/jwst/cube_build/ifu_cube.py
@@ -202,7 +202,7 @@ class IFUCubeData():
                         '_internal_s3d.fits'
                 if self.output_type == 'single':
                     newname = self.output_name_base + ch_name + '-' + b_name + \
-                        '-single_s3d.fits'
+                        '_single_s3d.fits'
             # ________________________________________________________________________________
             elif self.instrument == 'NIRSPEC':
 
@@ -221,7 +221,7 @@ class IFUCubeData():
                 fg_name = fg_name.lower()
                 newname = self.output_name_base + fg_name + '_s3d.fits'
                 if self.output_type == 'single':
-                    newname = self.output_name_base + fg_name + '-single_s3d.fits'
+                    newname = self.output_name_base + fg_name + '_single_s3d.fits'
                 if self.coord_system == 'internal_cal':
                     newname = self.output_name_base + fg_name + '_internal_s3d.fits'
         # ______________________________________________________________________________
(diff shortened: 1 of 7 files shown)


hbushouse:
Looks like a couple of unit tests need updating (see CI failures)

jwst/cube_build/ifu_cube.py
```diff
@@ -221,7 +221,7 @@ def define_cubename(self):
                 fg_name = fg_name.lower()
                 newname = self.output_name_base + fg_name + '_s3d.fits'
                 if self.output_type == 'single':
-                    newname = self.output_name_base + fg_name + '_single_s3d.fits'
+                    newname = self.output_name_base + fg_name + '-single_s3d.fits'
                 if self.coord_system == 'internal_cal':
                     newname = self.output_name_base + fg_name + '_internal_s3d.fits'
```
hbushouse, line 226: If the "single" file names are being changed to use a hyphen as the separator, instead of underscore, should the "internal" names be changed the same way here?

hbushouse (approved):
Looks OK mechanically to me, although I'm no expert on IFU WCS transforms. My eyes are picking up what I'm guessing will be flake8 failures, but I don't see the results of that in the latest CI test run. It's only showing the results of the docs build. Should try to trigger a rerun of all CI tests.

eslavich:
I've been running into macOS errors as well, but they don't seem to be consistent (I've seen a failure and a success after rerunning the same commit).

The doc build failure can be resolved by rebasing on master.

jemorrison:
@eslavich 
So I did a 'git fetch upstream'
'git rebase upstream/master'
But a git status says nothing changed so I have nothing to add with a commit.  In the past I have gone in and touched a file to make something to commit and then pushed it up. But there must be a better way of 'rebasing on master'

eslavich:
> But a git status says nothing changed so I have nothing to add with a commit.

That only happens if there's a conflict to resolve.  It sounds like git was able to rebase automatically, which means all you need to do afterwards is force push to the remote.

jemorrison:
@hbushouse  merged code to hopefully fix NIRSpec blotting. So you can continue with testing

PR #6327, opened 2021-09-03
JP-2210: Improvements to datamodels.open with ASDF files

### LINKED ISSUE #6229: datamodels.open does not permit strict_validation for .asdf files
_Issue [JP-2210](https://jira.stsci.edu/browse/JP-2210) was created on JIRA by [Edward Slavich](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eslavich):_

The datamodels.open function accepts a strict_validation flag when opening a .fits file, but for .asdf files it causes the following error:

TypeError: Unexpected keyword argument 'strict_validation'

It looks like we need to be more selective about the arguments that we pass to asdf.

eslavich:
Closes #6229 

Resolves [JP-2210](https://jira.stsci.edu/browse/JP-2210)

**Description**

This PR makes a few improvements to `datamodels.open` when opening a model from an ASDF file:

- Implement the `memmap` argument for ASDF files
- Emit a warning when the ASDF file is missing meta.model_type (previously this warning only existed for FITS missing DATAMODL).
- Fix bug that prevented extra arguments (like strict_validation) from being used when opening a model from an ASDF file.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


jwst/datamodels/tests/test_open.py
```diff
@@ -287,3 +288,41 @@ def test_open_asdf(tmp_path):
...
+    """
+    Test that unrecognized kwargs to the datamodels.open function
+    are passed on to the model class constructor.  Similar to the
+    above, except the invalid file must be created differently
+    because DataModel can't save an invalid .fits file.
+    """
+    file_path = tmp_path / "test.fits"
+
+    model = ImageModel()
+    model.save(file_path)
```
jdavies-st, line 320: It looks like this part is also exercising creating a datamodel with init=None instead of a shape as in the above test?

jwst/datamodels/util.py
```diff
@@ -57,16 +57,26 @@ def open(init=None, memmap=False, **kwargs):
...
+             If `True`, arrays will be validated against ndim, max_ndim, and datatype
+             validators in the schemas.
 
         - FITS
 
-           skip_fits_update - bool or None
+           cast_fits_arrays : bool
+             If `True`, arrays will be cast to the dtype described by the schema
+             when read from a FITS file.
+             If `False`, arrays will be read without casting.
```
jdavies-st, line 77: This is a new parameter available in stdatamodels 0.3.  If I'm using an earlier one, will it just ignore it?  Or do we need to specify `stdatamodels>=0.3` in setup.cfg for this PR?

jdavies-st (approved):

jwst/datamodels/tests/test_open.py
eslavich: Ah, oops, that was unintentional, thanks.

jwst/datamodels/util.py
eslavich: Yes excellent point!  We'll need the requirement I think.

PR #6329, opened 2021-09-07
created 3rd party package bayesicfitting

jemorrison:
Closes #

Resolves JP-1273

**Description**
The 3rd party package BayesFitting used by the MIRI residual_fringe step has been packaged and put into jwst/extern/bayesicfitting.
The url for the original code is https://www.bayesicfitting.nl
This PR addresses the fitting used in residual_fringe step.

Checklist
- [ ] Tests

- [ ] Documentation

- [ ] Change log

- [ ] Milestone

- [ ] Label(s)

### DIFF
--- a/jwst/extern/bayesicfitting/BaseFitter.py
+++ b/jwst/extern/bayesicfitting/BaseFitter.py
@@ -1,0 +1,949 @@
+import numpy as numpy
+import math
+import warnings
+import matplotlib.pyplot as plt
+from astropy.table import Table
+
+from .ImageAssistant import ImageAssistant
+from .MonteCarlo import MonteCarlo
+from .ConvergenceError import ConvergenceError
+from . import Tools
+from . import Plotter
+from .Formatter import formatter as fmt
+
+__author__ = "Do Kester"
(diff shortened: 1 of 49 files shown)


jemorrison:
@jdavies-st  are the CI failures something I need to fix ?
Do I need to add this directory to another place in setup.cfg

eslavich:
Adding an empty `__init)__.py` to the jwst/extern and jwst/extern/bayesicfitting directories will fix the ImportErrors.

jemorrison:
@eslavich  Do you know what the CI/SDP dependencies fail means - can you see what I am missing ?

eslavich:
I think it's running docstring tests in the new module?  I don't think we need to do that, those can be excluded in setup.cfg by adding `jwst/extern` here: https://github.com/spacetelescope/jwst/blob/master/setup.cfg#L121

Probably should also exclude that directory in the coverage config:

https://github.com/spacetelescope/jwst/blob/master/setup.cfg#L142

eslavich:
Here too the doc build can be fixed by rebasing on master.

PR #6332, opened 2021-09-08
JP-2224: Fixed a bug in flatfield for brightobj mode

### LINKED ISSUE #6268: NIRSpec - spec2 flat field: calculated flat fails validation test for BOTS data NRS1 detector
_Issue [JP-2224](https://jira.stsci.edu/browse/JP-2224) was created on JIRA by [Maria Pena-Guerrero](https://jira.stsci.edu/secure/ViewProfile.jspa?name=pena):_

During testing of build 7.8 (pipeline version 1.2.3), we found that the pipeline's calculated flat field does not match the NIRSpec team's calculated flat. This issue was addressed around the build 7.5 timeframe, and we have been since reporting the reappearance of the discrepancy in the flats, only for detector NRS1. For detector NRS2  both calculations of the flat match to computer precision (about 10^-7 difference between the on-the-fly calculated flats).

The NIRSpec team's flat field code can be found at: <https://github.com/spacetelescope/nirspec_pipe_testing_tool/blob/master/nirspec_pipe_testing_tool/calwebb_spec2_pytests/auxiliary_code/flattest_fs.py>

stscirij:
Closes #6268 

Resolves [JP-2224](https://jira.stsci.edu/browse/JP-2224)

**Description**

This PR addresses JP-2224.  The cutout of the S-flat component of the flatfield was calculated incorrectly by not including the slit offset as well as the subarray offset. 

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -72,6 +72,9 @@ flatfield
 - Updated flatfield step docs to include complete details on how the
   variance and error arrays are updated. [#6245]
 
+- Fixed a bug in flatfield for brightobj mode where the S-flat cutout
+  was calculated incorrectly by not accounting for the slit offset [#6332]
+
 jump
 ----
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):
Looks good. And I see that the change is confined to a routine that's only used for BrightObj mode, so that answers my question as to whether it would also affect other modes like the generic fixed-slit and MOS.

PR #6335, opened 2021-09-09
Add NRS BOTS interpolated flat regtest

hbushouse:
**Description**

This PR adds another regression test to the NIRSpec BOTS test suite to test the interpolated flat created for a BOTS exposure on NRS1, which was the problematic case reported and fixed in [JP-2224](https://jira.stsci.edu/browse/JP-2224).

Checklist
- [x] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)


stscirij (approved):
This looks good - the dataset has a slit xstart that is not 1, so it will test the correct application of the offset

hbushouse:
Failure in SDP dependencies CI test is unrelated (due to unexpected changes to NIRSpec specwcs ref files).

PR #6338, opened 2021-09-13
Add niriss soss datamodel

sosey:
Closes #

Resolves JP-

**Description**

This PR comes from a discussion with @eslavich about reference file validation and adds a datamodel for the niriss SOSS wavelength tables, the specwcs reference file...it likely still needs work, tests, a doc update and should be reviewed by @nden  or someone familiar with niriss soss mode before acceptance.  There's an example new reference file as well on CS. 

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/assign_wcs/niriss.py
+++ b/jwst/assign_wcs/niriss.py
@@ -15,11 +15,10 @@ from ..transforms.models import (NirissSOSSModel,
                                  NIRISSForwardRowGrismDispersion,
                                  NIRISSBackwardGrismDispersion,
                                  NIRISSForwardColumnGrismDispersion)
-from ..datamodels import (ImageModel, 
-                          NIRISSGrismModel, 
+from ..datamodels import (ImageModel,
+                          NIRISSGrismModel,
                           DistortionModel,
                           NirissSOSSTableModel)
-from stdatamodels import s3_utils
 from ..lib.reffile_utils import find_row
 
 log = logging.getLogger(__name__)
@@ -144,21 +143,10 @@ def niriss_soss(input_model, reference_files):
     world = cf.CompositeFrame([sky, spec], name='world')
 
     try:
-        # We'd like to open this file as a DataModel, so we can consolidate
-        # the S3 URI handling to one place.  The S3-related code here can
-        # be removed once we have a specwcs DataModel subclass.
-        # if s3_utils.is_s3_uri(reference_files['specwcs']):
-        #     bytesio_or_path = s3_utils.get_object(reference_files['specwcs'])
-        # else:
-        #     bytesio_or_path = reference_files['specwcs']
-        # with asdf.open(bytesio_or_path) as af:
-        #     wl1 = af.tree[1].copy()
-        #     wl2 = af.tree[2].copy()
-        #     wl3 = af.tree[3].copy()
         with NirissSOSSTableModel(reference_files['specwcs']) as f:
-            wl1 = f.models[0] # order 1
-            wl2 = f.model[1]  # order 2
-            wl3 = f.model[2]  # order 3
+            wl1 = f.models[0]  # order 1
+            wl2 = f.models[1]  # order 2
+            wl3 = f.models[2]  # order 3
 
     except IOError as e:
         raise IOError(f"Error reading wavelength correction from {reference_files['specwcs']}") from e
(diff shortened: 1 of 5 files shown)


jwst/assign_wcs/niriss.py
```diff
@@ -108,12 +110,20 @@ def niriss_soss(input_model, reference_files):
...
     -----
-    It includes tWO coordinate frames -
+    It includes two coordinate frames -
     "detector" and "world".
 
-    It uses the "specwcs" reference file.
+    It uses the "specwcs" reference file type.
     """
 
+    # The input is the grism image
```
hbushouse, line 119: Which kind of NIRISS "grism" image are we talking about here? The GR700XD grism that's used for SOSS mode, or one of the GR150R or GR150C grisms that's used for WFSS mode?

jwst/assign_wcs/niriss.py
```diff
@@ -108,12 +110,20 @@ def niriss_soss(input_model, reference_files):
...
+    if not isinstance(input_model, ImageModel):
```
hbushouse, line 120: SOSS mode data are usually taken as a time-series observation, which are always processed on a per-integration basis, and hence the input to the spec2 and spec3 pipeline levels is expected to be a `CubeModel` (3-D stack of integrations), not an `ImageModel`. Unfortunately, it's not mandatory that SOSS uses the time-series approach, so the input could be either `ImageModel` (2D) or `CubeModel` (3D). You never know.

jwst/assign_wcs/niriss.py
```diff
@@ -133,18 +143,12 @@ def niriss_soss(input_model, reference_files):
...
-            bytesio_or_path = reference_files['specwcs']
-        with asdf.open(bytesio_or_path) as af:
-            wl1 = af.tree[1].copy()
-            wl2 = af.tree[2].copy()
-            wl3 = af.tree[3].copy()
-    except Exception as e:
+        with NirissSOSSTableModel(reference_files['specwcs']) as f:
+            wl1 = f.models[0]  # order 1
+            wl2 = f.models[1]  # order 2
+            wl3 = f.models[2]  # order 3
```
hbushouse, line 149: Is it guaranteed that data for all 3 orders will always be present in the ref file?

jwst/datamodels/__init__.py
```diff
@@ -79,7 +79,7 @@
                              RegionsModel, WavelengthrangeModel, CameraModel, CollimatorModel, OTEModel,
                              FOREModel, FPAModel, IFUPostModel, IFUFOREModel, IFUSlicerModel, MSAModel,
                              FilteroffsetModel, DisperserModel, NIRCAMGrismModel, NIRISSGrismModel,
-                             WaveCorrModel)
+                             NirissSOSSTableModel,WaveCorrModel)
```
hbushouse, line 82: I'm guessing flake8 won't like the missing white space after the comma here

jwst/assign_wcs/niriss.py
```diff
@@ -133,18 +143,12 @@ def niriss_soss(input_model, reference_files):
...
+
+    except IOError as e:
         raise IOError(f"Error reading wavelength correction from {reference_files['specwcs']}") from e
```
hbushouse, line 152: Minor nit and not part of the PR changes, but is it technically correct to refer to the data in the ref file as a wavelength "correction"? I assumed it was just data that defined the wavelength scale to begin with, as opposed to adding some kind of correction to it.

jwst/datamodels/wcs_ref_models.py
```diff
@@ -341,6 +341,56 @@ def to_fits(self):
...
+        self.meta.instrument.detector = "NIS"
+        self.meta.exposure.type = "NIS_SOSS"
+        self.meta.reftype = self.reftype
+
+    def validate(self):
+        super(NirissSOSSTableModel, self).validate()
+        try:
+            assert isinstance(self.meta.input_units, (str, u.NamedUnit))
+            assert isinstance(self.meta.output_units, (str, u.NamedUnit))
+            assert self.meta.instrument.name == "NIRISS"
```
hbushouse, line 377: Given that all of these values are hardwired just 10 lines above, is it even possible for the asserts to ever fail and hence are they really useful?

hbushouse (commented):
The new model itself looks OK to me, but lots of comments/questions elsewhere.

melanieclarke:
@tapastro - this PR looks like it is probably no longer needed, or else would need significant work to bring it up to date.  Is it okay to close, or does this need a ticket filed to follow up?

hbushouse:
If you do end up implementing this, be sure to pay attention to the extremely useful and wise comments left by a reviewer about 3 years ago!  ;-)

sosey:
I have zero memory of any of this 😆

tapastro:
This PR has an aura about it - I don't recall seeing this ping either...

Closing the PR. It had a good run. No longer applicable, three years later.

PR #6346, opened 2021-09-13
JP-2122: Refactor of do_extract1d

tapastro:
Partially resolves JP-2122

**Description**

This PR collects redundant code found under do_extract1d and attempts to reduce clutter and increase readability. This is the first pass to check unit tests, with some function renaming and improved documentation to follow soon.

Checklist
- [ ] Tests

- [ ] Documentation

- [ ] Change log

- [x] Milestone

- [x] Label(s)


tapastro:
Alright, I clearly do not understand what is going on regarding the reverted alias PRs. I don't see the class_alias code in master, yet it shows as a diff in this PR. What gives?

stscieisenhamer (approved):
Short story: LGTM

Long story: On initial look, there are a couple of details that will need to be explored. One issue is the use of a local `verbose` flag when all that work should be handled by logging control. A step back reviewing how the step using this code is designed is probably also warranted. Any further improvements, if any, would need to be framed in a larger context.  As such, the current changes are a good basis for further work, if deemed necessary.

PR #6349, opened 2021-09-14
Residual Fringe fix for applying step on a file instead of an asn

jemorrison:
Resolves JP-1273

**Description**

The PR 6211 for adding the residual fringe step to the pipeline  did not work correctly if the input was a filename instead of an association.
This PR allows the input to be a filename.  Now the residual_fringe step can take either an association or a filename.
This PR also fixed an error in how suffix was initially defined. 

Checklist
- [ ] Tests
- [ ] Documentation
- [X] Change log
- [X] Milestone
- [X] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1990,6 +1990,9 @@ residual_fringe
 
 - Fixed suffix defined in spec [#6347]
 
+- Fixed an error when a filename was give as the input to apply the residual fringe correction [#6349]
+  
+
 set_telescope_pointing
 ----------------------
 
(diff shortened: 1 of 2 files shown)


hbushouse:
So if the step can work on a single file/exposure, why is it slated to be part of the spec3 pipeline, where an association of multiple exposures is being processed, as opposed to being in the spec2 pipeline and having it applied to one exposure at a time?

jemorrison:
@hbushouse The step just runs over a single file - if there are more in an association it just loops over each file separately.
The fits for the fringes will be better determined if the mrs_imatch step has already been applied to the data. So it should occur after mrs_imatch but before  the final cube_build in calspec3.

stscieisenhamer (approved):

hbushouse (approved):

PR #6362, opened 2021-09-22
JP-2289: Fix specwcs in wcs_ref_models

drlaw1558:
Resolves JP-2289

This PR fixes a bug in wcs_ref_models where SpecwcsModel was not of type ReferenceFileModel, causing validation to attempt to compare against astropy standard models instead of JWST specwcs models.  Also add an on_save setting of self.meta.reftype that was missing and causing validation errors, and add some missing allowed BAND values for MIRI MRS distortion and regions files.  Fix an incorrect comment on FilteroffsetModel.

### DIFF
--- a/jwst/datamodels/wcs_ref_models.py
+++ b/jwst/datamodels/wcs_ref_models.py
@@ -138,7 +138,8 @@ class DistortionMRSModel(ReferenceFileModel):
             assert self.meta.instrument.name == "MIRI"
             assert self.meta.exposure.type == "MIR_MRS"
             assert self.meta.instrument.channel in ("12", "34", "1", "2", "3", "4")
-            assert self.meta.instrument.band in ("SHORT", "LONG", "MEDIUM")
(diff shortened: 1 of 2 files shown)


jwst/datamodels/wcs_ref_models.py
```diff
@@ -166,6 +167,9 @@ class SpecwcsModel(_SimpleModel):
     schema_url = "http://stsci.edu/schemas/jwst_datamodel/specwcs.schema"
     reftype = "specwcs"
 
+    def on_save(self, path=None):
+        self.meta.reftype = self.reftype
+
```
jdavies-st, line 172: This is not needed now as `_SimpleModel.on_save` is already defined, and `SpecwcsModel` inherits it.

drlaw1558: Removed, and confirmed that the files still look ok.

stscieisenhamer (approved):

jdavies-st (approved):
Looks great!  Thanks for the contribution, as always.

Issue #6301, opened 2021-08-20
Allow necessary image drizzling parameters to be passed into resample step


stscijgbot-jp:
_Issue [JP-2261](https://jira.stsci.edu/browse/JP-2261) was created on JIRA by [Anton Koekemoer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=koekemoe):_

Necessary image drizzling parameters need to be able to be passed into the resample step, to be able to override the pipeline defaults (which would still be used if no such parameters are passed). This is a critical capability to enable users to carry out their science on the resampled images. For example, the WCS of the output frame needs to be able to be passed into the resample step, so that users can control the layout of resulting drizzled images (including, for example, the ability to drizzle images in different filters onto the same pixel grid, as well as the ability to specify the WCS in general). This capability is readily available in astrodrizzle and is widely used by the community, and offering a similar degree of flexibility in the jwst pipeline is critical to enable users to carry out their science. More details are provided in the comments section below.

stscijgbot-jp:
_Comment by [Anton Koekemoer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=koekemoe) on [JIRA](https://jira.stsci.edu/browse/JP-2261?focusedCommentId=545460#comment-545460):_

For specifying the output WCS, the following parameters would need to be passed to the resample step, and, if present, these would be used to define the output WCS - these are described here in terms of FITS keywords for clarity, but once they are passed into the resample code they can be translated into gwcs.WCS objects or other necessary quantities, as needed:

naxis1: x-dimension of output drizzled image (units=pixels, type=int)
naxis2: y-dimension of output drizzled image (units=pixels, type=int)
crpix1: x-coordinate of reference pixel in output drizzled image (units=pixels, type=float)
crpix2: y-coordinate of reference pixel in output drizzled image (units=pixels, type=float)
crval1: Right Ascension of reference pixel in output drizzled image (units=degrees, type=float)
crval2: Declination of reference pixel in output drizzled image (units=degrees, type=float)
rotation: Position angle of output drizzled image y-axis, relative to North, using the same rotation convention as the "final_rot" parameter in Astrodrizzle (units=degrees, type=float)
pixel_scale: Pixel scale of output drizzled image, in both x and y (units=arcseconds, type=float)

[...]

Issue #6358, opened 2021-09-21
Contradictory meaning of pixel_scale_ratio in resample step


mcara:
According to the [definition of `pixel_scale_ratio` in the `ResampleStep`](https://github.com/spacetelescope/jwst/blob/82d5766b7e08c7e39d15dd3e6ec701b50d93995c/jwst/resample/resample_step.py#L38), `pixel_scale_ratio` is:

> Ratio of output to input pixel scale.

This parameter is passed _directly_ (i.e., unmodified) to [`make_output_wcs`](https://github.com/spacetelescope/jwst/blob/82d5766b7e08c7e39d15dd3e6ec701b50d93995c/jwst/resample/resample_utils.py#L20) where the same parameter is defined as **opposite**:

> Ratio of input to output pixel scale.

The actual computation performed is that of "Ratio of input to output pixel scale".

I just would like to make sure this is the desired behavior (meaning docs for the `ResampleStep`) should be corrected. 

In addition, in the unit test, the relative tolerance is very very large:

https://github.com/spacetelescope/jwst/blob/82d5766b7e08c7e39d15dd3e6ec701b50d93995c/jwst/resample/tests/test_resample_step.py#L227

For an image of size 1000 pixels, this is equivalent to testing:

```
|size_a - size_b| <= 1 + 1 * 1000 = 1001
```

in which case an image of size 1000 would be close enough both to an image of 2000 pixels and to an image of 500 pixels.

PR #6364, opened 2021-09-23
Allow user-specified custom WCS parameters in the resample step

mcara:
Fixes https://github.com/spacetelescope/jwst/issues/6301 https://github.com/spacetelescope/jwst/issues/6358


mcara:
@jdavies-st  I think `output_wcs.data_size` could be replaced with `array_shape` which is now populated correctly. The question is: should I do this in this PR or to have a separate PR just for that.

jdavies-st:
> @jdavies-st I think `output_wcs.data_size` could be replaced with `array_shape` which is now populated correctly. The question is: should I do this in this PR or to have a separate PR just for that.

Feel free to do it here.  It's a good improvement.  Attaching an unofficial attribute to a WCS object was always dicey.

jwst/resample/resample_step.py
```diff
@@ -35,7 +35,14 @@ class ResampleStep(Step):
         kernel = string(default='square')
         fillval = string(default='INDEF')
         weight_type = option('ivm', 'exptime', default='ivm')
-        pixel_scale_ratio = float(default=1.0) # Ratio of output to input pixel scale
+        outnx = integer(min=1, default=None)
+        outny = integer(min=1, default=None)
```
jdavies-st, line 39: Don't like these names, but I don't have a better idea myself.  `output_size`?

Related, it would be nice if we could pass lists and tuples to the steps.  Theoretically we can do this via:

```
    shape = integer_list(min=list(1, 1), default=None)
```

but I don't know how the command line interface will deal with this.

mcara: I like this idea but `min=list(1, 1)` does not work (at least not like you would expect it to work: `min`, `max` is used to impose restrictions on the _number of elements_ in the list - not their values).

hbushouse (commented):
Looks good to me. Before we finalize this, however, it would be useful to post info in the JP-2261 ticket as to what new parameters are being added, so that the INS folks have a chance to review and comment.

mcara:
PR https://github.com/spacetelescope/stpipe/pull/33 fixes the issue with list arguments.

I have also discussed offline with @jdavies-st the issue of `array_shape` vs a property of the `Resample*` class itself and the decision was made to keep proposed change.

PR #6369, opened 2021-09-29
JP-1729: Assign int_num to extensions in extract_1d 

### LINKED ISSUE #5360: labeling of NIRSpec BOTS integrations
_Issue [JP-1729](https://jira.stsci.edu/browse/JP-1729) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

The level 3 "x1dints" product for NIRSpec BOTS data contains separate extensions for each integration and detector obtained in the exposure.  Currently, there is no obvious way to distinguish between each extension in terms of its corresponding detector and integration, which is necessary in order for a user to be able to combine the data on both detectors for a given integration.  Is the ordering of the extensions based on the original order in the raw exposure file, and is it the same for both detectors?  Even if so, I suggest adding the integration number and detector name (NRS1 or NRS2) to the header of each extension, so that there is no ambiguity for the user.

tapastro:
Closes #5360

Resolves JP-1729

**Description**

This PR utilizes the dormant 'INT_NUM' extension keyword in the MultiSpecModel datamodel to track the integration number of each extraction.

Checklist
- [ ] Tests

- [ ] Documentation

- [ ] Change log

- [x] Milestone

- [x] Label(s)

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -3860,6 +3860,9 @@ def create_extraction(extract_ref_dict,
         spec.dispersion_direction = extract_params['dispaxis']
         copy_keyword_info(meta_source, slitname, spec)
 
+        if integ > -1:
+            spec.int_num = integ
+
         if source_type is not None and source_type.upper() == 'POINT' and apcorr_ref_model is not None:
             log.info('Applying Aperture correction.')
             # NIRSpec needs to use a wavelength in the middle of the range rather then the beginning of the range


stscieisenhamer (approved):
Wow, now that was straight-forward.

hbushouse (approved):
This obviously fixes the problem, but would it be more appropriate to just call `populate_time_keywords` for this instrument mode? Of course you'd have to make sure there are hooks in that function to handle the case where INT_TIMES doesn't exist (e.g. still go ahead and populate INT_NUM without the times).

tapastro:
> This obviously fixes the problem, but would it be more appropriate to just call `populate_time_keywords` for this instrument mode? Of course you'd have to make sure there are hooks in that function to handle the case where INT_TIMES doesn't exist (e.g. still go ahead and populate INT_NUM without the times).

Yes, the issue posed by this ticket at its core is the lack of integration number for a dataset with an empty INT_TIMES table. The pipeline does run populate_time_keywords, but it doesn't populate with an empty INT_TIMES table. An argument could be made that INT_NUM isn't really a "time keyword", and any extraction input with multiple integrations will get covered by the new code. However the old code will work for any code with a proper INT_TIMES table.  I could see many avenues here, including a (likely over-engineered for the result) change to populate_time_keywords that would fill INT_NUM even with an empty INT_TIMES table. I'm not sure what the best course of action is.

hbushouse:
It may also end up being confusing for future maintenance to have the INT_NUM keyword populated in two different places in the code, depending on whether the INT_TIMES table has data. So even though it may be more work right now, I think that it'd be better in the long run if the `populate_time_keywords` function were updated to *always* populate INT_NUM, even when the INT_TIMES data is not available. At least that way it's all happening in only one place (despite the misnomer of INT_NUM not being a time-related keyword).

tapastro:
The forced assignment of INT_NUM even with an empty INT_TIMES table is now in populate_time_keywords - requesting another pass over the code before I merge.

PR #6373, opened 2021-10-04
Turn off MacOS job in CI

jdavies-st:
**Description**

This PR turns off the MacOS job in our CI workflow.  Github `macos-latest` runners are notoriously slow and these jobs fail more than 50% of the time, which is not good for reliable CI.  We should re-enable these when Github sorts out their MacOS runners to be more reliable.

Hopefully this won't bite as, as almost every developer is developing and running all these tests locally on MacOS.

Checklist
- [ ] ~Tests~
- [ ] ~Documentation~
- [ ] ~Change log~
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,10 +66,12 @@ jobs:
             python-version: 3.9
             toxenv: security
 
-          - name: macOS
-            os: macos-latest
-            python-version: 3.9
-            toxenv: py39
+          # MacOS job is flaky on Github actions and fails routinely.
+          # Re-enable when things get more reliable.
+          # - name: macOS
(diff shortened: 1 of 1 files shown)


hbushouse (approved):
Sounds a little risky, because we often forget to go back and reimplement things that get turned off, but I guess it's OK.

jdavies-st:
I filed an issue to eventually turn this back on.  #6375.

jdavies-st:
I'll add that if we can eventually get CRDS reffile caching between jobs turned back on, and maybe un-parallelize the `macos-latest` job, then it should be much more reliable.

stscieisenhamer (approved):

PR #6376, opened 2021-10-04
JP-1793: Standardize wfs_combine Step

### LINKED ISSUE #5511: Standardize wfs_combine

tapastro:
Closes #5511 

Resolves JP-1793

**Description**

This PR addresses issues with the Step/Pipeline identity crisis that calwebb_wfs-image3 - err, wfs_combine - found itself in. 

The step now uses datamodels.open() instead of load_as_level3_asn() - this should allow the step to take a ModelContainer as input, though the formatting will need to follow the level 3 associations, i.e. presence of member pairs under products, expnames provided, etc. Documentation might be on the to-do list to make this clear...

The step now follows a more traditional structure for saving results, though this requires an [stpipe PR](https://github.com/spacetelescope/stpipe/pull/29) to be merged that allows for Step instances to access the save_model(format=...) parameter.

Checklist
- [ ] Tests

- [ ] Documentation

- [ ] Change log

- [x] Milestone

- [x] Label(s)

### DIFF
--- a/jwst/wfs_combine/tests/test_wfs_combine.py
+++ b/jwst/wfs_combine/tests/test_wfs_combine.py
@@ -119,8 +119,8 @@ def test_create_combined(_jail, wfs_association,
                                  blur_size=10, n_size=2)
 
     # Check that results are as expected
-    assert result.data[5, 5] == result_data
-    assert result.dq[5, 5] == result_dq
+    assert result[0].data[5, 5] == result_data
+    assert result[0].dq[5, 5] == result_dq
 
 
 def test_shift_order_no_refine_no_flip(wfs_association):
--- a/jwst/wfs_combine/tests/test_wfs_combine_step.py
+++ b/jwst/wfs_combine/tests/test_wfs_combine_step.py
@@ -75,7 +75,7 @@ def test_step_pos_shift_no_refine_no_flip(wfs_association):
 
     wfs = WfsCombineStep.call(path_asn, do_refine=False, flip_dithers=False, psf_size=50,
                               blur_size=10, n_size=2)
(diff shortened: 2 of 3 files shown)


stscieisenhamer (approved):
LGTM

I am presuming the `stpipe` issue will need to be implemented before this is merged? Has an issue been filed against `stpipe` and is it linked to jp-1793?

tapastro:
> LGTM
> 
> I am presuming the `stpipe` issue will need to be implemented before this is merged? Has an issue been filed against `stpipe` and is it linked to jp-1793?

The stpipe issue does need to be in first, as this uses the new Step.output_use_format parameter.

I've tried to dot my github i's and cross the JIRA t's, but I'm not sure the linking of the stpipe issue to JP-1793 has gone through - I may need another link somewhere? The stpipe issue is [here](https://github.com/spacetelescope/stpipe/issues/30) and the PR [here](https://github.com/spacetelescope/stpipe/pull/29)

PR #6390, opened 2021-10-12
Remove redundant level-setting, logs in Pipelines, use logging ancestry

tapastro:
Closes #[TBD]
Resolves [JP-TBD](https://jira.stsci.edu/browse/JP-TBD)

**Description**

This PR addresses some hanging questions about logging implementation and utilizes the logging ancestry by defining the package logger in jwst/__init__.py - from here, loggers underneath jwst will inherit the level set there. Additionally, a few of the pipelines were using loggers set in calwebb_{pipeline}.py, when they have a log already defined when created as a sub-class of Pipeline with self.log.

I've tested this with a variety of calls with strun and pipelines/steps, using --verbose and without, and no features appear to have changed, but I could be missing something. I'd appreciate any suggestions on how this might have snuck a broken thing into the logging setup that I could have missed.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/__init__.py
+++ b/jwst/__init__.py
@@ -1,9 +1,13 @@
 import re
 import sys
+import logging
 from pkg_resources import get_distribution, DistributionNotFound
 
 __version_commit__ = ''
 _regex_git_hash = re.compile(r'.*\+g(\w+)')
+
+log = logging.getLogger('jwst')
+log.setLevel(logging.DEBUG)
 
 try:
     __version__ = get_distribution(__name__).version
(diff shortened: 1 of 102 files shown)


melanieclarke:
@tapastro - is this PR still needed? It looks related to [JP-1866](https://jira.stsci.edu/browse/JP-1866), but would need significant work to bring it up to date.

tapastro:
The PR is probably too dusty to bother reviving it, but work on jwst and stpipe logging would probably still be useful. This work branched off of work done for JP-2258.

PR #3557, opened 2019-05-30
Fix crashes in skymatch when some footprints almost entirely overlap

mcara:
This PR essentially re-implements enhancement in `stsci.skypac` in https://github.com/spacetelescope/stsci.skypac/pull/49

PR #6421, opened 2021-10-29
Fix crashes in skymatch when some footprints almost entirely overlap

mcara:
This PR extends the fix from https://github.com/spacetelescope/jwst/pull/3557 to the `SkyImage` class in order to mitigate issues reported by @jemorrison in https://jira.stsci.edu/browse/JP-1670

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -190,6 +190,12 @@ residual_fringe
 ---------------
  - Added documentation on step [#6387]
 
+skymatch
+--------
+
+- Improved reliability when matching sky in images with very close sky
+  footprints. [#6421]
+
 source_catalog
 --------------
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):
LGTM. I assume it actually works, right? ;-)

mcara:
> I assume it actually works, right? ;-)

mcara:
> I assume it actually works, right? ;-)

Who knows... It worked for the other class and my _hope_ is this will work for @jemorrison 's issue. Ideally, @jemorrison would clone my branch and test it with a data set.

However, it is easier to just merge this and rebase @jemorrison 's PR and then maybe tweak tolerance.

mcara:
... and if tolerance needs adjustment, make another PR just for tolerance value.

jemorrison:
@mcara @hbushouse  I grabbed the fixed PR (6421) and that fixed the problem with this set of data crashing. The data I was testing runs all the way through calimage3

mcara:
@jemorrison Thanks for confirming this!

PR #6422, opened 2021-10-29
JP-2306: New SOSS datamodel and core attributes for SOSS extraction parameters

### LINKED ISSUE #6408: Data model updates for new NIRISS SOSS extraction results
_Issue [JP-2306](https://jira.stsci.edu/browse/JP-2306) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The new NIRISS SOSS 1-D extraction methods developed by the NIRISS IDT result in additional meta data and extraction-related data arrays that are not defined in our current 1-D spectral data model (`SpecModel`). Either the `SpecModel` needs to be updated to allow for the additional data and meta data, or if that turns out to be not possible, a new 1-D spectral model should be created for the unique form of the SOSS spectra.

tapastro:
…xtraction Parameters

Closes #6408 
Resolves [JP-2306](https://jira.stsci.edu/browse/JP-2306)

**Description**

This PR is the first go at creating a new datamodel for the NIRISS SOSS extraction model traces and pixel weights, as well as new header keywords that detail parameters used in the extraction. 

This version hard-codes the three orders for traces and pixel weights, as opposed to a version that would avoid code duplication at the cost of over-engineering - thoughts?

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -2437,6 +2437,38 @@ properties:
             title: Pixel scale ratio relative to native detector scale
             type: number
             fits_keyword: PXSCLRT
+      soss_extract1d:
+        title: NIRISS SOSS Extract1D Parameters
+        type: object
+        properties:
+          width:
+            title: Aperture width used in extraction
+            type: number
+            fits_keyword: SOSS_WDTH
+          tikhonov_factor:
+            title: Tikhonov factor used in extraction
+            type: number
+            fits_keyword: TIKFAC
+          delta_x:
+            title: X-shift used in extraction
+            type: number
+            fits_keyword: SOSS_DLTX
+          delta_y:
+            title: Y-shift used in extraction
+            type: number
+            fits_keyword: SOSS_DLTY
+          theta:
+            title: Rotation angle used in extraction
+            type: number
+            fits_keyword: SOSS_ROTN
+          oversampling:
+            title: Oversampling used in extraction
+            type: number
+            fits_keyword: OVERSAMP
+          threshold:
+            title: Threshold used in extraction
+            type: number
+            fits_keyword: THRESHLD
       background:
         title: Background information
         type: object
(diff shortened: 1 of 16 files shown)


loicalbert:
Thanks for implementing the output data model changes that we requested. This looks good. I will send an email to Howard and Nadia to seek a meeting to discuss a few things. Schedule and deadlines, the code merging process,  the review process, etc.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -2437,6 +2437,38 @@ properties:
...
+            type: number
+            fits_keyword: SOSS_ROTN
+          oversampling:
+            title: Oversampling used in extraction
+            type: number
+            fits_keyword: OVERSAMP
+          threshold:
+            title: Threshold used in extraction
+            type: number
+            fits_keyword: THRESHLD
```
hbushouse, line 2471: In addition to these new header keywords to store the value of extraction params, we should also add new R_xxxxxx keywords to record the names of the new ref files being used for the SOSS extraction (unless of course those are already included in the code updates being made by the NIRISS team).

jwst/datamodels/sossmodel.py
```diff
@@ -0,0 +1,40 @@
+from .model_base import JwstDataModel
+
+__all__ = ['SossModel']
+
+
+class SossModel(JwstDataModel):
```
hbushouse, line 6: I wonder if we could use a slightly less generic class name for this model. Perhaps something like `SossExtractModel`, just so it's more obvious that it's associated with stuff used as part of the extraction?

jwst/datamodels/schemas/core.schema.yaml
tapastro: Seems unlikely, I'll add them. Didn't realize there were entries for these 😨

tapastro: Seems unlikely, it's easy to forget or miss, at least from my experience. 😬  I've added them.

hbushouse (approved):
LGTM

PR #6450, opened 2021-11-11
JP-2304: Fix arguments to DarkCurrentStep's make_output_path

### LINKED ISSUE #6402: dark_current error when using `dark_output` - redundant spec parameter?
_Issue [JP-2304](https://jira.stsci.edu/browse/JP-2304) was created on JIRA by [Tyler Pauly](https://jira.stsci.edu/secure/ViewProfile.jspa?name=tpauly):_

From a discussion on Slack with [Jonathan Aguilar](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jaguilar):

I’m trying to run Stage 1 from a python terminal (ok, a notebook). Could someone kindly point me to an example for how to pass arguments to steps? In this particular case, I’m trying to write out the frame-averaged dark data to a file called stage1_dark. However, my current attempts to set the jwst.pipeline.calwebb_detector1.Detector1Pipeline.dark_current.dark_output parameter fail with the following message:

```python
~/Projects/miniconda3/envs/mirisim/lib/python3.8/site-packages/jwst/dark_current/dark_current_step.py in process(self, input)
     39             dark_output = self.dark_output
     40             if dark_output is not None:
---> 41                 dark_output = self.make_output_path(
     42                     None, basepath=dark_output, ignore_use_model=True
     43                 )

TypeError: _make_output_path() got multiple values for argument 'basepath'
```

I was able to get the Step to produce an output file with an altered file name via the default Step parameter `output_file`:

```python
result = Detector1Pipeline.call(input_file, steps={‘dark_current’:{‘output_file’:‘stage1, ‘suffix’:‘dark’}})
```

However, I'm curious why `dark_output` exists and if it has additional functionality over `output_file` that I'm not understanding. If it does have value, there remains the bug that providing it a reasonable filename results in an error.

tapastro:
Closes #6402 
Resolves [JP-2304](https://jira.stsci.edu/browse/JP-2304)

**Description**

This PR changes the arguments provided to the DarkCurrentStep's make_output_path, such that the step parameter `dark_output` will now provide the frame-averaged darks in the filename specified, as per the documentation.

Aside: is anyone utilizing this output? With this sitting broken for so long, does this mean there are no tests verifying the contents of this file? 😨 

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -107,6 +107,12 @@ datamodels
 
 - Added the ``MirLrsPathlossModel`` for use in the ``pathloss` step. [#6435]
 
+dark_current
+------------
+
+- Fixed bug during save of optional averaged darks output, bug with
+  providing step a file instead of a datamodel, added regression test [#6450]
+
 extract_1d
 ----------
 
--- a/jwst/dark_current/dark_current_step.py
+++ b/jwst/dark_current/dark_current_step.py
@@ -55,13 +55,13 @@ class DarkCurrentStep(Step):
                 input_model, dark_model, dark_output
             )
 
-        out_data, dark_data = result
+            out_data, dark_data = result
 
-        if dark_data is not None and dark_data.save:
-            save_dark_data_as_dark_model(dark_data, dark_model, instrument)
-        dark_model.close()
+            if dark_data is not None and dark_data.save:
+                save_dark_data_as_dark_model(dark_data, dark_model, instrument)
+            dark_model.close()
 
-        out_ramp = dark_output_data_2_ramp_model(out_data, input_model)
+            out_ramp = dark_output_data_2_ramp_model(out_data, input_model)
 
         return out_ramp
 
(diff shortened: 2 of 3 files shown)


stscieisenhamer (changes requested):
Since it was broken for so long, there probably is not a test. A test should be implemented.

hbushouse (commented):
Looks good to me.

Requisite change log request ...

hbushouse:
A quick "grep dark_output *.py" of the regtest directory reveals no hits, hence this is not being tested. Sounds like someone just volunteered to create one.

hbushouse:
Here's an example of an uncal exposure that uses NFRAMES>1 and hence could be used as input to a regtest:
/grp/jwst/ssb/bushouse/jwst_data/NIRCam/jw00312007001_02102_00001_nrcblong_uncal.fits
For the purpose of the regtest, you could process it up through the step in calwebb_detector1 preceding dark_current and then just run dark_current in the test.

tapastro:
Note: Two files will need to be added to artifactory for the new regression test.

stscieisenhamer (approved):

tapastro:
Files successfully uploaded to Artifactory.

PR #6453, opened 2021-11-12
Fix incorrect removal of PC and CDELT info from FITS WCS

mcara:
Resolves [AL-607](https://jira.stsci.edu/browse/AL-607)

**Description**

This PR fixes a bug in `assign_wcs` due to which WCS elements (`CD` matrix, `CDELT`, ...) are removed _after_ header has been updated with the new header from `to_sip_fits()`. Instead this clean-up, if needed, should happen before header is updated.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/assign_wcs/assign_wcs_step.py
+++ b/jwst/assign_wcs/assign_wcs_step.py
@@ -114,15 +114,11 @@ class AssignWcsStep(Step):
         except ValueError:
             return result
 
+        # update meta.wcs_info with fit keywords except for naxis*
+        del fit_sip_hdr['naxis*']
+
         # maintain convention of lowercase keys
         fit_sip_hdr = {k.lower(): v for k, v in fit_sip_hdr.items()}
-
-        # update meta.wcs_info with fit keywords except for naxis*
-        for key in ['naxis1', 'naxis2']:
-            del fit_sip_hdr[key]
-
-        # update meta.wcs_info with fit keywords
-        result.meta.wcsinfo.instance.update(fit_sip_hdr)
 
         # delete naxis, cdelt, pc from wcsinfo
         rm_keys = ['naxis', 'cdelt1', 'cdelt2',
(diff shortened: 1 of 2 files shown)


jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -121,14 +121,14 @@ def process(self, input, *args, **kwargs):
...
-
         # delete naxis, cdelt, pc from wcsinfo
         rm_keys = ['naxis', 'cdelt1', 'cdelt2',
                    'pc1_1', 'pc1_2', 'pc2_1', 'pc2_2']
         for key in rm_keys:
             if key in result.meta.wcsinfo.instance:
                 del result.meta.wcsinfo.instance[key]
 
+        # update meta.wcs_info with fit keywords
+        result.meta.wcsinfo.instance.update(fit_sip_hdr)
```
nden, line 132: In this case we still end up with `naxis` in `meta.wcsinfo`. I think if you add it to the list of attributes to be remove in line 121 this should fix it.

nden (changes requested):

jwst/assign_wcs/assign_wcs_step.py
```diff
@@ -118,11 +118,7 @@ def process(self, input, *args, **kwargs):
         fit_sip_hdr = {k.lower(): v for k, v in fit_sip_hdr.items()}
 
         # update meta.wcs_info with fit keywords except for naxis*
-        for key in ['naxis1', 'naxis2']:
-            del fit_sip_hdr[key]
-
-        # update meta.wcs_info with fit keywords
-        result.meta.wcsinfo.instance.update(fit_sip_hdr)
+        del fit_sip_hdr['naxis*']
```
nden, line 121: This doesn't work, raises an error.

mcara: Yes, I just moved this before the line in which the header is converted to a dictionary.

mcara:
This is ready for review/merging

nden (approved):
lgtm, thanks

PR #6454, opened 2021-11-12
JP-2291: Fix bug in barshadow for off-center sources

### LINKED ISSUE #6365: NIRSpec master background subtraction incomplete coverage for nods
_Issue [JP-2291](https://jira.stsci.edu/browse/JP-2291) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

Testing some variations of MOS master background subtraction using dedicated background shutters in the same MSA configuration (i.e., the spec2 version, not spec3)... I'm using a simulation of 3-shutter nodded exposures, creating a simplified MSA metafile with just one science slitlet and multiple background slitlets. For the first nod, in which the source is in the middle shutter of the slitlet, I get a good result where the background is subtracted relatively cleanly across the entire 2D spectrum of the slitlet:

 

!image.png!

However, when I do the same thing for the second and third nods, the resampled background is present in the 2D space of the single shutter containing the source, but is close to zero for the other two shutters of the slitlet. This results in the background being subtracted for only the source shutter:

!Screen Shot 2021-09-24 at 11.03.52 AM.png!

The resampling of the combined 1D background needs to be done over the entire slitlet. Otherwise the 1D extraction will include unsubtracted background, unless a small extraction aperture is used (which may be undesirable if the source is significantly extended, a primary use case of master background subtraction). Is there some part of the MB processing that depends on the nod position that could be causing this behavior? Perhaps the inverse bar shadow correction?

hbushouse:
Closes #6365
Resolves [JP-2291](https://jira.stsci.edu/browse/JP-2291)

**Description**

This PR fixes a bug in the barshadow step that was leading to the correction array being calculated incorrectly for slitlets in which the source is indicated as being off-center (in the spatial direction) in the slitlet. The bar-shadow correction only applies to extended/uniform sources in the first place, which are assumed to fill the slitlet (or at best the distribution is not known), hence the correction should not depend at all on any information from the MSA meta data that indicates which shutter within the slitlet contains the center of the source. It should be completely independent of source location in the slitlet. The fix removes all effects on the calculation of the correction that were due to source location within the slitlet.

The previously incorrect results were affecting the results of the master_background subtraction in instances where a point source is located off-center in a slitlet (e.g. due to nodding) and the barshadow correction is computed for the background/sky signal in the slitlet. The resulting correction applied to the slitlet was asymmetric, leading to incorrect background/sky subtraction.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/assign_wcs/nirspec.py
+++ b/jwst/assign_wcs/nirspec.py
@@ -634,8 +634,8 @@ def get_open_msa_slits(msa_file, msa_metadata_id, dither_position,
             else:
                 jmin = min([s['shutter_column'] for s in slitlets_sid])
                 jmax = max([s['shutter_column'] for s in slitlets_sid])
-                j = jmin + (jmax - jmin) // 2 + 1
-            ymax = 0.5 + margin + (jmax - j) * 1.15
+                j = jmin + (jmax - jmin) // 2
+            ymax = yhigh + margin + (jmax - j) * 1.15
             ymin = -(-ylow + margin) + (jmin - j) * 1.15
             quadrant = slitlets_sid[0]['shutter_quadrant']
             ycen = j
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,9 @@ assign_wcs
 - Changed in_ifu_slice in util.py to return the indices of elements in slice.
   Also the x tolerance on finding slice elements was increased. [#6326]
 
+- Fixed bug in NIRSpec MOS slitlet meta data calculations for background slits
+  consisting of multiple shutters. [#6454]
+
 associations
 ------------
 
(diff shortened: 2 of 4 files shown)


stscirij (approved):
Looks good, now we don't care where the 'x' is in shutter_status, or even if there is one :)

jwst/assign_wcs/nirspec.py
```diff
@@ -634,8 +634,8 @@ def get_open_msa_slits(msa_file, msa_metadata_id, dither_position,
             else:
                 jmin = min([s['shutter_column'] for s in slitlets_sid])
                 jmax = max([s['shutter_column'] for s in slitlets_sid])
-                j = jmin + (jmax - jmin) // 2 + 1
-            ymax = 0.5 + margin + (jmax - j) * 1.15
+                j = jmin + (jmax - jmin) // 2
```
nden, line 637: I finally convinced myself that this change is correct. It is important that `j` and `ycen` are 1-based and with this change they are because the msa metadata file `shutter_column` values are 1-based. They need to be 1-based in order to compute the correct [shutter_id](https://github.com/spacetelescope/jwst/pull/6454/files#diff-45fd9f060b3edd128f0e133edcd2124e8ae1b90140b72adffe3a4e2f715049dfL673). 
The extra `+1` previously may have been based on an assumption that `jmin` and `jmax` are 0-based.

nden (approved):

jwst/assign_wcs/nirspec.py
hbushouse: And it does seem to work correctly now too, based on actual data coming out of the step. In the past, if a background slit contained 3 shutters in columns 21, 22, 23 (for example), it set j=ycen=23 and the shutter_state string was "11x". Both results indicating the source location in an end shutter. Now it gives j=ycen=22 and "1x1" (source in the middle).

PR #6457, opened 2021-11-15
JP-2103: Fix calculation of source_xpos in wavecorr

### LINKED ISSUE #6074: The source_xpos parameter not being calculated for FS data?

### LINKED ISSUE #6287: wavecor step not applying any correction to NIRSpec FS dithered data
_Issue [JP-2253](https://jira.stsci.edu/browse/JP-2253) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

I have a set of simulated FS exposures in which the spectral dither pattern was used. In cases where the source is not centered in the slit in the dispersion direction, a wavelength correction should be applied by the wavecorr step. In this case, the SRCXPOS values for the 3 dithered exposures are -0.29, -0.42, 0.54, and so the wavelength values per pixel should be different among the exposures, as well as relative to the default slit-centered values. However, when I compare the wavelength arrays before and after the wavecorr step, the differences are all exactly zero, which indicates that no correction was actually applied.

In detail, what I did was to run the extract_2d step and save to data model "d1e2d", then ran the sourcetype step with that as input, and then finally the wavecorr step with output saved the data model "d1wavecor". To calculate the differences, I then subtracted d1e2d.slits[0].wavelength - d1wavecor.slits[0].wavelength. All values within the valid spectral trace are zero. Are the latter wavelengths supposed to be updated with the corrected values, or should I have used a different pipeline product such as the final cal.fits output?

nden:
Closes #6074
Resolves [JP-2103](https://jira.stsci.edu/browse/JP-2103)
Closes #6287
Resolves [JP-2253](https://jira.stsci.edu/browse/JP-2253)

**Description**

This PR corrects the computation of the position of the source in the NRS fixed slit.

It also removes `vparity` from the transform from Ideal to V2V3 which was found to be incorrect.
The method `absolute2fractional` was removed because the returned `source_xpos` position is essentially
the fractional position.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/wavecorr/wavecorr.py
+++ b/jwst/wavecorr/wavecorr.py
@@ -4,7 +4,9 @@ slits in which the source is off center. The correction is applied
 only to point sources.
 
 The algorithm uses a reference file which is a look up table of
-wavelenght_correction as a function of slit_x_position and wavelength.
+wavelength_correction as a function of slit_x_position and wavelength.
+The ``x`` direction is the one parallel to dispersion/wavelength for
+both MOS and FS slits.
 
 The slit_x_position is determined in the following way.
 
@@ -279,7 +281,7 @@ def get_source_xpos(slit, slit_wcs, lam):
     log.debug("wcsinfo: {0}, {1}, {2}, {3}".format(v2ref, v3ref, v3idlyangle, vparity))
     # Compute the location in V2,V3 [in arcsec]
     xv, yv = idl2v23(xoffset, yoffset)
-    log.info(f'xoffset, yoffset, {xoffset}, {yoffset}, {lam}')
+    log.info(f'xoffset, yoffset, {xoffset}, {yoffset}')
 
     # Position in the virtual slit
     xpos_slit, ypos_slit, lam_slit = slit.meta.wcs.get_transform('v2v3', 'slit_frame')(
@@ -288,6 +290,6 @@ def get_source_xpos(slit, slit_wcs, lam):
     slit.source_xpos = xpos_slit
     slit.source_ypos = ypos_slit
     log.debug('Source X/Y position in V2V3: {0}, {1}'.format(xv, yv))
-    log.info('Source X/Y position in the slit: {0}, {1}, {2}, {3}'.format(xpos_slit, ypos_slit, lam, lam_slit))
+    log.info('Source X/Y position in the slit: {0}, {1}'.format(xpos_slit, ypos_slit))
 
     return xpos_slit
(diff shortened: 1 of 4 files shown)


jwst/wavecorr/wavecorr.py
```diff
@@ -1,7 +1,30 @@
...
-#  Module for applying wavelength corrections to NIRSpec MOS and FS
-#  slit instances in which the source is off center.
-#
+"""
+Module for applying wavelength corrections to NIRSpec MOS and FS
+slits in which the source is off center. The correction is applied
+only to point sources.
+
+The algorithm uses a reference file which is a look up table of
+wavelenght_correction as a function of slit_x_position and wavelength.
```
hbushouse, line 7: typo: wavelenght -> wavelength

jwst/wavecorr/wavecorr.py
```diff
@@ -1,7 +1,30 @@
...
+
+The slit_x_position is determined in the following way.
```
hbushouse, line 9: Perhaps it would be useful to mention somewhere near the start that the "x" direction is the one parallel to dispersion/wavelength for both MOS and FS slits.

jwst/wavecorr/wavecorr.py
```diff
@@ -261,55 +276,18 @@ def get_source_xpos(slit, slit_wcs, lam, msa_model):
     vparity = slit.meta.wcsinfo.vparity
 
     idl2v23 = trmodels.IdealToV2V3(v3idlyangle, v2ref, v3ref, vparity)
+    log.debug("wcsinfo: {0}, {1}, {2}, {3}".format(v2ref, v3ref, v3idlyangle, vparity))
     # Compute the location in V2,V3 [in arcsec]
     xv, yv = idl2v23(xoffset, yoffset)
+    log.info(f'xoffset, yoffset, {xoffset}, {yoffset}, {lam}')
```
hbushouse, line 282: You're printing 3 values but only 2 of them are named in the text part of the message.

nden: fixed

jwst/wavecorr/wavecorr.py
nden: done

hbushouse (approved):
Looks good, but I think it needs a rebase to fix conflicts with recent updates to change log.

PR #6458, opened 2021-11-16
JP-2329 Add pysiaf to SDP dependency

stscieisenhamer:
Resolves [JP-2329](https://jira.stsci.edu/browse/JP-2329)

**Description**

This PR addresses adds `pysiaf` to the SDP dependencies.

Also, the list of SDP dependencies is now `sdp`, instead of just `ephem`.

Checklist
- [ ] Tests
- [ ] ~~Documentation~~
- [ ] ~~Change log~~
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -76,7 +76,7 @@ bc0.conda_packages = [
 ]
 bc0.pip_reqs_files = ['requirements-sdp.txt']
 bc0.build_cmds = [
-    "pip install -e .[test,ephem] --no-cache-dir",
+    "pip install -e .[test,sdp] --no-cache-dir",
     "pip install pytest-xdist",
     "pip freeze",
 ]
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,6 +51,8 @@ install_requires =
     tweakwcs>=0.7.2
     
 [options.extras_require]
+aws =
+    stsci-aws-utils>=0.1.2
 docs =
     matplotlib
     sphinx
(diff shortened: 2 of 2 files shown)


JenkinsfileRT
```diff
@@ -76,7 +76,7 @@ bc0.conda_packages = [
 ]
 bc0.pip_reqs_files = ['requirements-sdp.txt']
 bc0.build_cmds = [
-    "pip install -e .[test,ephem] --no-cache-dir",
+    "pip install -e .[test,sdp] --no-cache-dir",
```
hbushouse, line 79: Are the SDP folks in the habit of installing with this argument set? Or is it done for them automagically?

hbushouse (approved):
Looks reasonable to me. To we require change log entries for installation-related stuff like this?

JenkinsfileRT
stscieisenhamer: Unless the discussion with SPB reveals otherwise, I believe this is the path. The theory is that this build is used to produce the dependency files found on artifactory with which SDP uses to build JWST. Initial test will be to run the regression job after merging this, if the regression is successful, or maybe even if it isn't.

hbushouse:
Shall we merge and see what happens in a regtest run or wait for more eyes to have a look first?

stscieisenhamer:
I believe we can merge when ready. Can always back it out. As long as the regression job is in a state, ideally passing, that will demonstrate success on this.

PR #6459, opened 2021-11-16
JP-2291: minor tweak to barshadow update

hbushouse:
Resolves [JP-2291](https://jira.stsci.edu/browse/JP-2291)

**Description**

This PR adds a minor tweak to the `bar_shadow` updates implemented in PR #6454. In the original PR, small changes were made to the bar shadow correction array even for slits that contained a centered source, while the update was intended to fix only slits that have off-center sources. So this adds an extra hook to first check to see if the source position is at the center of the slitlet shutter list and, if it is, don't modify the original `yslit` data. This has the benefit of keeping regressions to a minimum and also keeps us in sync with the NIRSpec team results for slitlets that have a centered source (otherwise there is a small difference).

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,7 +52,7 @@ barshadow
   centering of the source within the slitlet, because for extended/uniform
   sources the centering is irrelevant. This fixes bugs encountered when
   computing the correction for background signal contained within slits with
-  an off-center point source. [#6454]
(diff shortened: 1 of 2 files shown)


stscirij (approved):
Looks good

Issue #6434, opened 2021-11-03
extract_1d bug for NIRSpec IFU cubes with internal_cal coordinates


stscijgbot-jp:
_Issue [JP-2322](https://jira.stsci.edu/browse/JP-2322) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

I'm encountering a spec2 pipeline crash at the extract_1d step when processing IFU data taken with an internal lamp. The CRDS rmap for the step has not yet been updated to handle internal lamps, so I specified a local reference file using an override. It's possible that I did not modify the reference file correctly (the only thing I changed was the exposure type, from NRS_IFU to NRS_LAMP). However, the reference file is asdf format, while the error implies that the pipeline is expecting an image format, which seems strange.

The full error message:

[...]

PR #6460, opened 2021-11-16
Extract1d ifu internal jp 2322

jemorrison:
Closes JP-2322
Resolves [JP-2322](https://jira.stsci.edu/browse/JP-2322)

**Description**
NRS_LAMP needed to be added to the options of ext_types that have an extract1d reference file in asdf format. 

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ X] Milestone
- [X ] Label(s)


jwst/extract_1d/extract.py
```diff
@@ -134,13 +134,12 @@ def open_extract1d_ref(refname: str, exptype: str) -> dict:
 
     if refname == "N/A":
         ref_dict = None
-    elif exptype in ['MIR_MRS', 'NRS_IFU']:
+    elif exptype in ['MIR_MRS', 'NRS_IFU', 'NRS_LAMP']:
         # read in asdf file
         extract_model = datamodels.Extract1dIFUModel(refname)
```
tapastro, line 139: MRS and NRS_IFU exptypes necessitate an IFUModel - do all NRS_LAMP exposures produce an IFUModel, or can NRS_LAMP be assigned to data from fixed slits or the MSA?

tapastro: I think I might be misunderstanding NIRSpec data generally - should it always be a cube if EXP_TYPE is NRS_LAMP, because it's just detector reads?

hbushouse: No, EXP_TYPE='NRS_LAMP' can come in any form of IFU, MOS, or FS mode for NIRSpec and hence could be `IFUImageModel` for IFU mode and perhaps something like `MultiSlitModel` for MOS and FS.

hbushouse:
@jemorrison I know it's just tedious bookkeeping, but please fix the information in the description above: "Closes" should list the *github issue number* (#6434  in this case) and the whitespace inside the brackets for the checklist items has to be removed (in order for them to get rendered properly).

jwst/extract_1d/extract.py
```diff
@@ -134,13 +134,12 @@ def open_extract1d_ref(refname: str, exptype: str) -> dict:
 
     if refname == "N/A":
         ref_dict = None
-    elif exptype in ['MIR_MRS', 'NRS_IFU']:
+    elif exptype in ['MIR_MRS', 'NRS_IFU', 'NRS_LAMP']:
```
hbushouse, line 137: I'm not sure this simple addition will work for all modes. It will be OK for lamp exposures that happen to be taken in IFU mode (OPMODE='IFU'), but what about lamp exposures taken in MOS or FS mode (OPMODE='MSASPEC' or 'FIXEDSLIT'). We don't want those latter types going down the branch that only works for IFU mode.

jwst/extract_1d/extract.py
jemorrison: Ok I thought NRS_LAMP data in calspec2 meant it was going to create an IFU cube.   If this is not the case we could just check if the data was IFUModel - then read as asdf file.

hbushouse: That sounds like it could work. If lamp exposures for the other modes already work correctly and we're only trying to fix IFU lamp exposures, then you could just key off of either the datamodel type or OPMODE='IFU'.

jemorrison:
@hbushouse @tapastro  does this new fix look right. Only IFUtype data have an extract1d ref file in asdf. 
If that changes in the future then we will need to revisit this code.

jemorrison:
@hbushouse I did not realize the Closed referred to the original GitHub issue. Now I do. I thought it was strange I had to type the JP number in 2 places.  I really think I need a tutorial on how to properly document these tickets.

tapastro (commented):
One alternative to this, which may prevent further issues if the built-in assumptions of {Only ASDFs for IFUs, Only JSONs for other non-ImageRefFiles} could be something like:

```
def load_ref(ref_file):
    file_stream = open(ref_file)
    try:
        ref_dict = json.load(file_stream)
        ref_dict['ref_file_type'] = 'FILE_TYPE_JSON'
        file_stream.close()
    except (UnicodeDecodeError, JSONDecodeError) as e:
        print(f"{e} for {ref_file}")
        file_stream.close()
        try:
            with fits.open(ref_file) as openfits:
                fits_present = 1
        except OSError as e:
            print(f"{e} for {ref_file}")
            fits_present = 0
        if fits_present:
            extract_model = datamodels.MultiExtract1dImageModel(ref_file)
            ref_dict = {'ref_file_type': 'FILE_TYPE_IMAGE', 'ref_model': extract_model}
        else:
            extract_model = datamodels.Extract1dIFUModel(ref_file)
            ref_dict = dict()
            ref_dict['ref_file_type'] = 'FILE_TYPE_ASDF'
            ref_dict['ref_model'] = extract_model
```

Where we just try to load as JSON - if a JSONDecodeError, check if file can be loaded as FITS - if OSError from bad load (this step might be sketchy?), only reference structure left is the ASDF file containing an Extract1dIFUModel. If either of the first two work, the relevant ref_dict should be made.

jemorrison:
@tapastro  Let me test you idea. I like that better than just assuming IFU extract1d reference files are asdf

jemorrison:
@tapastro  I updated the how the extract1d reference file is read in.

jwst/extract_1d/extract.py
```diff
@@ -106,7 +107,7 @@ class ContinueError(Exception):
     pass
 
 
-def open_extract1d_ref(refname: str, exptype: str) -> dict:
+def open_extract1d_ref(refname: str, exptype: str, model_type: str) -> dict:
```
tapastro, line 110: Looks good! I think this argument can now be removed, though.

tapastro (changes requested):

jwst/extract_1d/extract.py
```diff
@@ -2656,9 +2656,11 @@ def run_extract1d(
     """
     # Read and interpret the extract1d reference file.
     try:
-        ref_dict = open_extract1d_ref(extract_ref_name, input_model.meta.exposure.type)
+        ref_dict = open_extract1d_ref(extract_ref_name, input_model.meta.exposure.type,
+                                      input_model.meta.model_type)
     except AttributeError:  # Input is a ModelContainer of some type
-        ref_dict = open_extract1d_ref(extract_ref_name, input_model[0].meta.exposure.type)
+        ref_dict = open_extract1d_ref(extract_ref_name, input_model[0].meta.exposure.type,
```
tapastro, line 2662: Should've quoted this too, for the same reason.

jwst/extract_1d/extract.py
jemorrison: Good catch I forgot about that

jwst/extract_1d/extract.py
jemorrison: @tapastro I think this is ready to be merged

tapastro (approved):
LGTM

hbushouse (approved):

PR #6462, opened 2021-11-17
JP-2332: Ensure NRS_LAMP imaging data properly skips flatfielding

### LINKED ISSUE #6461: Flat-field step can't handle NRS_LAMP with OPMODE=IMAGE
_Issue [JP-2332](https://jira.stsci.edu/browse/JP-2332) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

Program 450, observation 64 contains a number of NIRSpec lamp exposures that are failing in the calwebb_image2 flat_field step. The exposures have EXP_TYPE=NRS_LAMP, GRATING=MIRROR, and OPMODE=IMAGE. The NIRSpec team requests that these kinds of exposures receive calwebb_image2 processing, hence JP-1876 was implemented to create level-2b "image2" ASN's for these exposures. That work was successful and the ASN's are being created, but when the calwebb_image2 processing reaches the flat-field step, it fails with the error "No flat field algorithm exists for handling data".

The flat_field step needs to be fixed to allow these exposures to pass. The CRDS rmap for the NIRSpec FLAT ref file has an entry that will map to a ref file of "N/A" for these exposures, at which point the flat_field step will automatically skip processing.

Note that this will leave assign_wcs as the only active step in calwebb_image2 to be applied to these kinds of lamp exposures, which is as desired by the NIRSpec team.

tapastro:
Closes #6461
Resolves [JP-2332](https://jira.stsci.edu/browse/JP-2332)

**Description**

This PR adds an extra check on NRS_LAMP exposures that route lamp_mode=IMAGE to the imaging portion of FlatFieldStep, where an additional check was needed such that if no flat is given or provided, the step will be skipped.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -9,14 +9,14 @@ import numpy as np
 from gwcs.wcstools import grid_from_bounding_box
 
 from .. import datamodels
-from .. datamodels import dqflags
-from .. lib import reffile_utils
-from .. assign_wcs import nirspec
+from ..datamodels import dqflags
+from ..lib import reffile_utils
+from ..assign_wcs import nirspec
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
-MICRONS_100 = 1.e-4                     # 100 microns, in meters
+MICRONS_100 = 1.e-4  # 100 microns, in meters
 
 # This is for NIRSpec.
 FIXED_SLIT_TYPES = ["NRS_LAMP", "NRS_AUTOWAVE", "NRS_BRIGHTOBJ", "NRS_FIXEDSLIT"]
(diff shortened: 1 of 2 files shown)


tapastro:
I let flake run wild on this file, causing a lot of extra line changes that aren't relevant.. also still failed the flake check. 👎

jwst/flatfield/flat_field.py
```diff
@@ -202,14 +208,14 @@ def apply_flat_field(science, flat, inverse=False):
     # Update the variances using BASELINE algorithm.  For guider data, it has
     # not gone through ramp fitting so there is no Poisson noise or readnoise
     if not isinstance(science, datamodels.GuiderCalModel):
-        flat_data_squared = flat_data**2
+        flat_data_squared = flat_data ** 2
         science.var_poisson /= flat_data_squared
         science.var_rnoise /= flat_data_squared
-        science.var_flat = science.data**2 / flat_data_squared * flat_err**2
+        science.var_flat = science.data ** 2 / flat_data_squared * flat_err ** 2
```
hbushouse, line 214: Curse flake8. To me the original syntax (without whitespace) is more natural and readable (to a human).

hbushouse (approved):
Looks good. And now I understand why this wasn't working properly before, with the N/A result from CRDS. Most steps check for an N/A ref file way up at the beginning of the step module already and never get to the point of calling the correction if the ref file is N/A. But because flatfield has multiple ref files, for different modes, it can't do a reasonable check on the ref files until it's gone through more logic to figure out what kind of data it's working on.

Oh, requisite request for change log entry ...

jwst/flatfield/flat_field.py
tapastro: Agreed 😞

PR #6464, opened 2021-11-18
JP-2333: Load level3 ASN using method rather than force-load with datamodels

### LINKED ISSUE #6463: wfs_combine crashes due to missing TA exposures in ASN
_Issue [JP-2333](https://jira.stsci.edu/browse/JP-2333) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

There are some kinds of Wavefront Sensing (WFS) observations that contain target-acq exposures that only get processed through level-2a (rate product). But the wfs-image3 ASN that's created lists the cal products (by default) for the TA exposures. Even though the TA exposures are not used in any of the wfs_combine processing, the way in which the ASN file is loaded (using the generic `datamodels.open(asn_file)`) causes the existence of all files in the ASN to be checked and hence fails when the cal product for the TA exposure(s) can't be found:

FileNotFoundError: [Errno 2] No such file or directory: 'jw00632020001_02102_00001_nrca3_cal.fits'

The way in which the input ASN table is opened/loaded needs to be changed to work in the same way as other level-3 pipelines (e.g. calwebb_image3, calwebb_coron3, etc.), which are correctly setup to be able to ignore members that aren't needed, so that it doesn't crash when those files aren't present.

tapastro:
Closes #6463 
Resolves [JP-2333](https://jira.stsci.edu/browse/JP-2333)

**Description**

This PR changes the wfs_combine step (pipeline?) to use the Step.load_as_level3_asn() method rather than datamodels.load() for ingesting the association - this allows for operations on the contents of the association file without requiring all listed files in the association to be present.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/wfs_combine/wfs_combine_step.py
+++ b/jwst/wfs_combine/wfs_combine_step.py
@@ -31,46 +31,46 @@ class WfsCombineStep(Step):
         self.name_format = False
 
         # Load the input ASN table
-        with datamodels.open(input_table) as asn_table:
+        asn_table = self.load_as_level3_asn(input_table)
 
-            self.log.info('Using input table: %s', input_table)
-            self.log.info('The number of pairs of input files: %g', len(asn_table.meta.asn_table.products))
+        self.log.info('Using input table: %s', input_table)
+        self.log.info('The number of pairs of input files: %g', len(asn_table['products']))
 
-            output_container = datamodels.ModelContainer()
+        output_container = datamodels.ModelContainer()
 
-            # Process each pair of input images listed in the association table
-            for which_set in asn_table.meta.asn_table.products:
+        # Process each pair of input images listed in the association table
+        for which_set in asn_table['products']:
 
-                # Get the list of science members in this pair
-                science_members = [
-                    member
-                    for member in which_set.members
-                    if member.exptype.lower() == 'science'
-                ]
-                infile_1 = science_members[0].expname
-                infile_2 = science_members[1].expname
-                outfile = which_set.name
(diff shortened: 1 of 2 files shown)


jwst/wfs_combine/wfs_combine_step.py
```diff
@@ -31,46 +31,46 @@ def process(self, input_table):
...
+            wfs.input_1.close()
+            wfs.input_2.close()
+
+            # Update necessary meta info in the output
+            output_model.meta.cal_step.wfs_combine = 'COMPLETE'
+            output_model.meta.asn.pool_name = asn_table['asn_pool']
+            output_model.meta.asn.table_name = os.path.basename(input_table)
+            output_model.meta.filename = which_set['name']
+
+            output_container.append(output_model)
```
hbushouse, line 73: Because we're no longer using a "with" block here, should we explicitly do an `asn_table.close()` here or just assume it'll be taken care of by python when it goes out of scope?

tapastro: I copied my work from [here](https://github.com/spacetelescope/jwst/blob/master/jwst/pipeline/calwebb_coron3.py), and the asn is not explicitly closed, so I think it should be fine. I initially tried to just replace the datamodels.load() with the load_as... in the with block, but the method is not configured to work in such a way.

hbushouse (approved):
LGTM

PR #6475, opened 2021-11-23
JP-2284: Refactor residual fringe background fit to use astropy splines.

WilliamJamieson:
**Description**

This PR switches the residual fringe background fit step from using `BayesicFitting` based splines to the new `astropy` based splines. This is part of the effort to remove the dependency of `jwst` on `BayesicFitting`. Note that this will change the results of the residual fringe correction slightly due to numerical differences in how `astropy` (`scipy` based) splines are computed/fit vs how `BayesicFitting` splines are computed/fit.

Note that the bugfix astropy/astropy#12523 is required for this code to function properly.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [ ] Label(s)


hbushouse:
This PR can be closed, right?  It appears that it's been superseded by 2 other PRs.

PR #6478, opened 2021-11-29
JP-2331 Cube build footprint

### LINKED ISSUE #6456: Cube build uses incorrect footprint computation
_Issue [JP-2331](https://jira.stsci.edu/browse/JP-2331) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

Sometime since jwst build 1.1.0, some revisions between cube_build and assign_wcs changes has meant that the IFU cube footprint is now being calculated incorrectly for the 'skyalign' method.  It appears that position information from the long-wavelength side of the detector is getting used to determine the field boundaries without first trimming to only the wavelength range of interest in the cube itself.  This only seems to affect 'skyalign' cubes, not 'ifualign' or 'internal_cal', and seems to be due to line 1228 of ifu_cube.py relying on a default footprint computation rather than using the dedicated find_corners routine.

Screenshot attached shows the difference between a cube created with build 1.1.0 (on the left) vs any build since (1.2.x, 1.3.x, or latest master).

jemorrison:
Closes #6456 
Resolves [JP-2331](https://jira.stsci.edu/browse/JP-nnnn)

**Description**
This PR concerns a bug fix for MIRI MRS IFU cubes. 
This PR fixes a bug in calculating the IFU footprint for cubes not created by calspec2 pipeline. In the calspec2 pipeline a single IFU cube is created for each exposure. This IFU cube will contain both channels on the detector. In this case (output_type = multi) and the spatial and spectral footprint on the sky are determined in assign_wcs and are the footprints for both channels.  Assign_wcs stores the spatial footprint in input_model.meta.wcsinfo.s_region and the spectral footprint in input_model.meta.wcsinfo.spectral_region.  Because the MIRI channels are designed to be roughly concentric the s_region in wcsinfo is correct for the larger FOV (channel 2 and channel 4). When creating IFU cubes that contain one  band (or one channel) and for channel with the smaller FOV on the detector (channel 1 or channel 3)  we do not want to use the footprint determined by assign_wcs because the footprint in wcsinfo is too large. In this case  we need to calculate the footprint in cube_build using only data for the appropriate channel. 

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ X] Milestone
- [ X] Label(s)


jwst/cube_build/ifu_cube.py
```diff
@@ -1227,6 +1227,17 @@ def setup_ifucube_wcs(self):
                 world = False
                 if self.coord_system == 'skyalign':
                     world = True
+                # Do not use the spatial or spectral region found wcs if
```
drlaw1558, line 1230: Suggest:
Do not use the default spatial or spectral region found in the wcs if

drlaw1558 (approved):
Minor comment on the comment, but otherwise lgtm.  Results as expected and fix the problem.

jwst/cube_build/ifu_cube.py
jemorrison: done

nden (approved):
lgtm

Not sure why the hidden file is removed in this PR.
I restarted the tests to clear the Py 3.10 job.

PR #6482, opened 2021-11-29
JP-2339 Move setting of the default method to `calc_transforms`

stscieisenhamer:
Resolves [JP-2339](https://jira.stsci.edu/browse/JP-2339)

**Description**

This PR moves the use of the default method to the function that actually needs to use a method: calc_transforms

This should address an incompatibility introduced in a previous PR.

Checklist
- [x] Tests
- [ ] ~~Documentation~~
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/lib/set_telescope_pointing.py
+++ b/jwst/lib/set_telescope_pointing.py
@@ -928,6 +928,8 @@ def calc_transforms(t_pars: TransformParameters):
     transforms : `Transforms`
         The list of coordinate matrix transformations
     """
+    t_pars.method = t_pars.method if t_pars.method else Methods.default
+
     transforms = t_pars.method.func(t_pars)
     return transforms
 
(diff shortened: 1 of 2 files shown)


jwst/lib/set_telescope_pointing.py
```diff
@@ -928,6 +928,8 @@ def calc_transforms(t_pars: TransformParameters):
     transforms : `Transforms`
         The list of coordinate matrix transformations
     """
+    t_pars.method = t_pars.method if t_pars.method else Methods.default
```
jbdoggett, line 931: It'd be better for TransformationParameters.method to be initialized with Methods.default rather than have users of the class figure out a default unless there's a special value needed for a particular situation.

jbdoggett (changes requested):

jwst/lib/set_telescope_pointing.py
stscieisenhamer: Good catch. This was leftover from having to deal with two different methods for COARSE vs. TRACK. But that has been handled, so a proper default can be added.

jbdoggett: You should now be able to take out this line.

jbdoggett (changes requested):

stscieisenhamer:
Decided to leave in, for interface consistency. It will still be true of that is None, the default will be used, in case somehow it becomes None.

jbdoggett (approved):

tapastro (approved):
LGTM

PR #6491, opened 2021-12-06
JP-2286: Refactor the Bayesian evidence step of residual fringe correction

WilliamJamieson:
**Description**

This PR switches the Bayesian evidence step of the residual fringe correction from using the `BayesicFitting` library to using astropy based models and fitters. Note that this will change the results of the residual fringe correction, and these changes should be investigated throughly. The changes are due to differences in how the numerical approximations of the Bayesian evidence are computed (`scipy/astropy` vs `BayesicFitting` based computations). 

Note that this depends on PR #6475 (and should be merged after that PR is merged).

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [ ] Label(s)


hbushouse:
How does this PR relate to all the other ones currently open for residual fringe work, such as #6475 , #6761 , and #6739 ? Can this one be closed, because it's obsolete?

WilliamJamieson:
> How does this PR relate to all the other ones currently open for residual fringe work, such as #6475 , #6761 , and #6739 ? Can this one be closed, because it's obsolete?

Not quite yet. This is the second "phase" (it refactors the second step/part) of the residual fringe correction refactor. The other PRs are related to the first phase of the residual fringe correction.

hbushouse:
Any updates on the status of this? The residual fringe correction step is now available in the cal pipeline. Is this work meant to be an enhancement to the baseline algorithm, in order to remove a dependency on `BayesicFitting` (which I assume is in use in the baseline version)?

WilliamJamieson:
> Any updates on the status of this? The residual fringe correction step is now available in the cal pipeline. Is this work meant to be an enhancement to the baseline algorithm, in order to remove a dependency on `BayesicFitting` (which I assume is in use in the baseline version)?

The issue is that astropy has extreme bottle necks on some of the fitting involved, which means that this causes the residual fringe to take >5x longer than the current code. It will require extremely careful benchmarking to narrow down the issues. For now I'm keeping it around because it works (but it is much slower), or for the event that we absolutely need to drop the `BaysicFitting` dependency.

PR #6493, opened 2021-12-07
JP-2326: Adding changes to the change log due to changes in STCAL bug fix

### LINKED ISSUE #6445: Incorrect slope calculation when jumps near start of ramp
_Issue [JP-2326](https://jira.stsci.edu/browse/JP-2326) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

One of the MIRI IDT members (Javier Alvarez) noticed some oddities in working with simulated data that we've run down to a bug in how ramp fitting behaves when cosmic rays occur near the start of a ramp on anything other than the first integration.

In the data set linked above pixel x,y=415,480 (0-indexed) shows this problem.  As illustrated in ramp.png, the ramp for this pixel shows a jump between group 2 and 3 for integration number 2 (1-indexed).  The first two groups are flagged according to GROUPDQ though (DONOTUSE) so I would expect that ramp fitting would be insensitive to this jump.  Halting ols_fit around line 1335 indicates that this pixel is being split into two segments (opt_res.slope_seg[1,:,480,415]) with slopes of 4.5e2 and -1.3e-1 respectively.

Clearly, despite the fact that group 2 is flagged in GROUPDQ it is nonetheless still being used in determining the slope for this segment.  This problem only seems to occur for integrations 2 and 3, not integration 1.  Multiple such artifacts are visible in the rateints data for integrations 2 and 3, but not for integration 1.

From poking at this a little further, I think the problem is that the RSCD step is setting the entirety of the 2nd group for everything **except** the first integration to DONOTUSE (as intended), but this is **not** being respected by ramp fitting.

The jump step seems to successfully ignore the 2nd group DONOTUSE, and thus doesn't look for any jumps between groups 2 and 3.  When ramp_fit runs though, ols_fit.discard_miri_groups looks for where DONOTUSE is set in all pixels (line 687) and trims those groups out before passing the arrays on to the actual ramp fitting.  However, while DONOTUSE is set everywhere in group 2 for integrations 2 and above, it is not set in group 2 for integration 1.  As written, line 687 will therefore not trim off group 2 for any integration, and will use the data therein during ramp fitting with no knowledge of any cosmic ray jumps here (since jump step skipped looking for jumps here as these pixels aren't meant to be used).

Essentially, because discard_miri_groups isn't throwing away group 2 of these later integrations, any cosmic rays between groups 2 and 3 aren't being discarded at all.

In order to test this I forced DONOTUSE to be set for group 2 in integration 1 as well; in this case discard_miri_groups sees the flag and trims both groups 1 and 2.  The final rate.fits image (i.e., combined across all integrations) with this modification is MUCH better than the current pipeline version; see beforeafter.png attached).

I'm not sure what the best overall solution is, but this needs to be fixed in order to not have many unpleasant artifacts in the slope data.

kmacdonald-stsci:
Closes #6445 
Resolves [JP-2326](https://jira.stsci.edu/browse/JP-2326)

**Description**

This PR simply adds a jwst change log entry for a fix made in stcal.

Checklist
- [ ] Tests
- [ ] Documentation
- [X] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -221,7 +221,7 @@ ramp_fitting
 ------------
 
 - Fix issue with inappropriately including a flagged group at the beginning
-  of a ramp segment. [telescope/stcal#68]
+  of a ramp segment. [spacetelescope/stcal#68]
 
 - Pixels with negative median rates will have VAR_POISSON set to zero.
   [spacetelecope/stcal#59]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -220,6 +220,9 @@ pipeline
 ramp_fitting
 ------------
 
+- Fix issue with inappropriately including a flagged group at the beginning
+  of a ramp segment. [telescope/stcal#68]
+
 - Pixels with negative median rates will have VAR_POISSON set to zero.
   [spacetelecope/stcal#59]
 


CHANGES.rst
```diff
@@ -220,6 +220,9 @@ pipeline
 ramp_fitting
 ------------
 
+- Fix issue with inappropriately including a flagged group at the beginning
+  of a ramp segment. [telescope/stcal#68]
```
nden, line 224: We don't like ground based telescopes. :-)

nden:
```suggestion
  of a ramp segment. [spacetelescope/stcal#68]
```

nden (approved):

hbushouse (approved):

PR #6503, opened 2021-12-14
JP-1750: Update extract_1d to allow user-specified aperture center

### LINKED ISSUE #5415: Allow Extract1D to override source location
_Issue [JP-1750](https://jira.stsci.edu/browse/JP-1750) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

As currently written, the IFU version of Extract1D doesn't allow users to override the source location at which to perform aperture extraction.  There is no fundamental reason why it shouldn't however, and it would be beneficial both for testing purposes and science analysis if this could be easily overridden rather than relying exclusively on the target location in the headers.  For instance, running Extract1D in an offline capacity on multiple sources in a large extended mosaic would mean that users don't need to write their own software to do this very basic operation.

There are a few nuances, such as also allowing users to override the source type (POINT vs EXTENDED) in order to actually get the point source extraction routine to run at a fixed location in a mosaic that might be nominally EXTENDED due to the APT dithering options, but these should not overly complicate the issue.

I'm unsure what the status of source location overrides is for other instruments, so it's probably worth a cal WG discussion just to make sure that everyone is on the same page.

Priority is 'Nice to Have', but not critical.

hbushouse:
Closes #5415 
Resolves [JP-1750](https://jira.stsci.edu/browse/JP-1750)

**Description**

This PR adds a new argument "center_xy" to the extract_1d step, which can be used to override the steps computation of the center of the extraction aperture from 3D IFU cubes. The argument values are a list of x, y coords, in units of pixels. It only applies to IFU data and can only be applied when the source is POINT like (SRCTYPE=POINT). Extended sources still default to extracting the entire cube.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [X] Label(s)


setup.cfg
```diff
@@ -45,7 +45,7 @@ install_requires =
     spherical-geometry>=1.2.20
     stcal>=0.4.1
     stdatamodels>=0.3.0,<1.0
-    stpipe>=0.1.0,<1.0
+    stpipe>=0.3.1,<1.0
```
tapastro, line 48: Should this be here?

jwst/extract_1d/ifu.py
```diff
@@ -92,6 +92,19 @@ def ifu_extract1d(input_model, ref_dict, source_type, subtract_background,
...
 
+    # If the user supplied extraction center coords,
+    # load them into extract_params for use later.
+    if center_xy is not None:
+        if len(center_xy) == 2:
+            extract_params['x_center'] = int(center_xy[0])
+            extract_params['y_center'] = int(center_xy[1])
+            log.info(f'Using user-supplied x_center={center_xy[0]}, y_center={center_xy[1]}')
+        else:
+            log.warning('Incorrect number of values in center_xy; should be two.')
```
tapastro, line 103: Might still need to add extract_params['x_center'] = None in this else: block, or they will be undefined for the following check on None on L428.

jwst/extract_1d/extract.py
```diff
@@ -2800,6 +2809,12 @@ def do_extract1d(
         reference file (or the default position, if there is no reference
         file) will be shifted to account for source position offset.
 
+    center_xy : int or None
```
tapastro, line 2812: Might be a nitpick, but should this type def be list[int] or None, or iterable[int] or None, rather than int or None?

tapastro (commented):
Looks good, just minor comments.

setup.cfg
hbushouse: The latest (or nearly latest) version of stpipe is required in order to allow for proper parsing of arguments that contain a list of values (which is what `center_xy` needs).

jwst/extract_1d/extract.py
hbushouse: No idea

jwst/extract_1d/ifu.py
hbushouse: Good catch. Fixed in next commit. I initialize them to None and then set them only if given by the user.

tapastro (approved):
LGTM

PR #6509, opened 2021-12-15
Update linearity test  in jwst pipeline caused by changes in  STCAL PR 65 

jemorrison:
This PR is to support JP-2311  and STCAL PR [#65](https://github.com/spacetelescope/stcal/pull/65)

**Description**

The linearity step on STCAL was updated to deal with the case where all the linearity coefficients =0. It has been assumed that  this is a mistake in the reference file. If the linearity term of the linearity correction is zero then the resulting data will be zero. PR 65 on STCAL checks if the linear term is zero and if it is it sets the value to 1 and sets the DQflag in the linearity model to "NO_LIN_CORR'.  These pixels will have no linearity correction applied to them and their DQFlag will include NO_LIN_CORR. 
One of the CI tests is to run the unit test for the linearity step in the JWST Pipeline.
This PR fixes cases where the test did not set the linear term to a value. The default value was 0, so the returned DQ for the pixel now included NO_LIN_CORR.
.
This PR has to be merged before the STCAL code (PR65) can be merged because the test is failing.

Checklist
- [X] Tests
- [ ] Documentation
- [X] Change log
- [X] Milestone
- [X] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -193,6 +193,8 @@ linearity
 
 - Use the common code in STCAL for linearity correction. [#6386]
 
+- Update of linearity test to support STCAL PR65 [#6509]
+
 outlier_detection
 -----------------
 
--- a/jwst/linearity/tests/test_linearity.py
+++ b/jwst/linearity/tests/test_linearity.py
@@ -61,7 +61,7 @@ def test_coeff_dq():
 
     # test case where all coefficients are zero, the linearity reference file may
     # not mark these pixels as NO_LIN_CORR. The code will mark these pixels as
-    # NO_LIN_COR
+    # NO_LIN_CORR
     ref_model.coeffs[:, 25, 25] = 0.0
     im.data[0, 50, 25, 25] = 600.0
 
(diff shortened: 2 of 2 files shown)


jemorrison:
This test is going to fail in CI test because we have not merged the code over on stcal #65. But that code can not be merged until we fix this test.

jemorrison:
@nden @stscirij 
I made a change to STCAlL(see link below) that that does not apply the linearity correction if the linear coefficient  in the correction =0.  The CI tests on STCAL that run the jwst unit tests  failed because how this test was set up. The test was set up in some case where all the coefficients =0 but it was not testing the linearity corrected data but the DQ returned. However now with STCAL change - it changes the DQ to NO_LIN_CORR for these pixels. So I need to update this test. However this test fails because the STCAL code is not merged into master. 
https://github.com/spacetelescope/stcal/pull/65

I need someone to approve one of these to get the ball moving. I picked this one to start the process because teamcoffee does not have the approval to approve the STCAL code.

stscirij (approved):
LGTM

jemorrison:
@nden I believe this test is failing because the changes I made on STCAL are not part of the release version. What do you suggest I do with the PR ?

hbushouse:
Recent release of stcal 0.4.3 now includes the needed updates to the linearity step to support the upgraded tests in this PR. So I'm merging.

PR #6519, opened 2021-12-27
Temp fix for failing NRS lamp regtests

hbushouse:
**Description**

This PR makes a temporary change to 2 NIRSpec lamp exposure regression tests, to avoid a failure caused by recent updates to CRDS ref files. The NIRSpec team recently submitted new pars-spec2pipeline ref files that enable the extract_1d step for lamp exposures, but skip the aperture correction. The CRDS bestref that runs at the beginning of pipeline instantiation doesn't know apcorr is going to be skipped and hence tries to find an appropriate APCORR ref file and fails (there aren't any for lamp modes). This will be fixed, but not for a while yet. So in the meantime I've modified the 2 failing regtests to skip the extract_1d step for lamp exposures. This can be reactivated once CRDS updates are in place (post-B7.9?).

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/regtest/test_nirspec_lamp_fs_spec2.py
+++ b/jwst/regtest/test_nirspec_lamp_fs_spec2.py
@@ -23,7 +23,9 @@ def run_pipeline(rtdata_module):
             "--steps.extract_2d.save_results=true",
             "--steps.flat_field.save_results=true",
             # The pars file skips resampling by default.  We want to test it.
-            "--steps.resample_spec.skip=False",]
+            "--steps.resample_spec.skip=False",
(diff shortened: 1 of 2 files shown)


nden (approved):

nden:
How will we remember to turn this back on when the par files are fixed? Another issue?

hbushouse:
> How will we remember to turn this back on when the par files are fixed? Another issue?

[JP-2363](https://jira.stsci.edu/browse/JP-2363)

PR #6520, opened 2021-12-27
Try restricting sphinx-automodapi

eslavich:
**Description**

This PR restricts sphinx-automodapi to < 0.14.0 to avoid https://github.com/astropy/sphinx-automodapi/issues/141 which is causing the jwst doc build to fail.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

PR #6522, opened 2021-12-27
DOC: Blender rules need __all__

pllim:
Reverts spacetelescope/jwst#6520

Instead of pinning automodapi, just don't try to import everything.

Because I am lazy, I am using GitHub interface. I will fix the `__all__` in subsequent commit.

### DIFF
--- a/docs/jwst/model_blender/blender_rules.rst
+++ b/docs/jwst/model_blender/blender_rules.rst
@@ -40,6 +40,4 @@ above.  This definition illustrates how several rules are simply interfaces for
 numpy array operations, while others are defined internally to `model_blender`.
 
 .. automodapi:: jwst.model_blender.blendrules
-   :skip: OrderedDict
-   :skip: Time
-   :skip: time
+  :no-inheritance-diagram:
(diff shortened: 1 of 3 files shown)


pllim:
Yikes, I guess some code is relying on the lack of `__all__` to work... 😬

eslavich:
> Yikes, I guess some code is relying on the lack of __all__ to work... 😬

Those failures are in master too, I think since the rtd build succeeded this will work just fine.

pllim:
@eslavich , good to know! I was afraid `blendrules` has dark magic like `stpyfits`. I am glad you replied before I went too deep into the rabbit hole...

pllim:
p.s. Who merged PR with failing CI? 😜

eslavich (approved):
LGTM, way better than pinning the dependency.  Thanks!

pllim:
Not saying that astropy/sphinx-automodapi#141 doesn't need to be fix but I have no idea when that will happen.

hbushouse (approved):
LGTM. Are we agreed that this is the final/best solution?

hbushouse:
All of the CI failures are unrelated and due to something we're also trying to fix in a linearity step test.

PR #6532, opened 2021-12-30
review oldest deps

nden:
~Closes #~
~Resolves [JP-nnnn](https://jira.stsci.edu/browse/JP-nnnn)~

**Description**

This PR bumps up minimum versions of some dependencies.
gwcs >=0.18 to pickup `CompoundBoundingBox` and `asdf-wcs-schemas`
scipy>=1.6.0 to satisfy photutils requirements

This should fix the failing `oldest-dependencies` GA job .

Checklist
~- [ ] Tests~
~- [ ] Documentation~
~- [ ] Change log~
- [x] Milestone
- [x] Label(s)


nden:
@eslavich Should we bump the asdf version? Any others?

hbushouse (approved):
Works for me.

eslavich (commented):
Yes let's bump minimum asdf up to 2.8.2, I think this PR is necessary for jwst: https://github.com/asdf-format/asdf/pull/1029.

The asdf 2.8.3 release was just a fix of the tests to allow the library to be packaged for Debian, so we don't need to require that.

eslavich:
The current "Oldest dependency versions" failure is due to astropy 5 having a minimum Python version that exceeds the job's version (3.7).

nden:
Ah, right. Since astropy 5.0 is the LTS release I think we should drop support for Python 3.7 and numpy 1.17.
Any objections?

hbushouse:
> Ah, right. Since astropy 5.0 is the LTS release I think we should drop support for Python 3.7 and numpy 1.17. Any objections?

Make it so.

PR #6543, opened 2022-01-04
JP-2373: Fix tolerances in failing unit tests

### LINKED ISSUE #6542: Fix unit test tolerance limits for changes in Linux numpy 1.22
_Issue [JP-2373](https://jira.stsci.edu/browse/JP-2373) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

Are friends at numpy have switched to the Intel Short Vector Math Library in release 1.22 (came out 12/31/2021), but only for Linux distributions. Because all of our CI tests run on Linux, this change is resulting in tiny numerical differences in a few unit tests, all of which are in the /jwst/lib/tests/ directory (e.g. test_fgs_pointing, test_set_telescope_pointing, test_velocity_aberration). The failing unit tests should be modified to set an appropriate tolerance for the assert comparisons.

hbushouse:
Closes #6542 
Resolves [JP-2373](https://jira.stsci.edu/browse/JP-2373)

**Description**

This PR adds the use of "isclose" to some unit tests in order to allow for specification of tolerances, to allow for differences between results on Linux vs Mac, thanks to latest numpy 1.22 release.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [X] Milestone
- [X] Label(s)


hbushouse:
This first commit is basically a draft, to test whether the methods work at all. Some of which are clunky. May need iterating, based on the CI test results to see if these changes actually work.

jwst/lib/tests/test_set_telescope_pointing.py
```diff
@@ -590,7 +591,8 @@ def test_add_wcs_method_full_siafdb(eng_db_ngas, data_file, tmp_path):
 
         with datamodels.open(DATA_PATH / expected_name) as expected:
             for meta in METAS_EQUALITY:
-                assert model[meta] == expected[meta]
+                #assert model[meta] == expected[meta]
+                assert np.isclose(model[meta], expected[meta], atol=1e-13)
```
nden, line 595: This will not work because the `meta` contains values of type `str` and those cannot be compared with `isclose`. 
One way to do this is to move all numerical values from `METAS_EQUALITY` to `METAS_ISCLOSE` and leave the string values in `METAS_EQUALITY` where they can be directly compared as before.

hbushouse: Yep, of course. Why didn't I see that?  Latest commit does different comparisons depending on datatype of arguments. Cross your fingers ...

jwst/lib/pointing_summary.py
```diff
@@ -26,11 +26,11 @@
 >>> print(t["delta_v1"])
           delta_v1
     -------------------
-    0.13712727164010646
+    0.13712727164010
```
nden, line 29: This is an output alignment issue Move lines 27-29 4 spaces left and keep the previous value

hbushouse:
@stscieisenhamer OK, I've got it down to only 1 test that's still failing, but it appears to be something that's being run/built automatically from example code that's included in the doc string of the /lib/pointing_summary.py module and is not obvious to me how to fix. Here's the latest error report:

`017 >>> calc_pointing_deltas(im)
018     Delta(target=<SkyCoord (ICRS): (ra, dec) in deg
019         (90.75541667, -66.56055556)>, v1=<SkyCoord (ICRS): (ra, dec) in deg
020         (91.08142005, -66.60547869)>, refpoint=<SkyCoord (ICRS): (ra, dec) in deg
021         (90.70377653, -66.59540224)>, delta_v1=<Angle 0.13712727 deg>, delta_refpoint=<Angle 0.04044315 deg>)
022 
023 >>> t = calc_deltas([im])
024 >>> print(t.columns)
025     <TableColumns names=('exposure','target','v1','refpoint','delta_v1','delta_refpoint')>
026 >>> print(t["delta_v1"])
Differences (unified diff with -expected +actual):
    @@ -1,3 +1,3 @@
    -          delta_v1
    -    -------------------
    -    0.13712727164010
    +      delta_v1     
    +-------------------
    +0.13712727164010546

/home/runner/work/jwst/jwst/jwst/lib/pointing_summary.py:26: DocTestFailure`

nden:
We need to bump the version of spherical_geometry up to the latest release to pick up the fix @mcara made. I guess it can be done here.

nden:
@hbushouse I made a comment on how to fix it above.

jwst/lib/pointing_summary.py
hbushouse: Ah, thanks. @stscieisenhamer ignore my question (below), since Nadia just answered it.

nden:
Is it worth running the regtests on this to see if there are other cases of numerical differences?

hbushouse:
> Is it worth running the regtests on this to see if there are other cases of numerical differences?

Can't hurt. Once I get everything working here I'll kick off a regtest run against the PR.

stscieisenhamer (approved):
LGTM

nden:
The Py 310 test  failure is real.
@stscieisenhamer Is this simply an example and is it OK to skip that doctest?

hbushouse:
> The Py 310 test failure is real. @stscieisenhamer Is this simply an example and is it OK to skip that doctest?

Or is there some way to truncate or round-off the values, as with the other tests? But I'm not sure how to do that when it's code being executed in a doc string.

jwst/lib/pointing_summary.py
```diff
@@ -24,13 +24,13 @@
 >>> print(t.columns)
     <TableColumns names=('exposure','target','v1','refpoint','delta_v1','delta_refpoint')>
 >>> print(t["delta_v1"])
-          delta_v1
```
hbushouse, line 27: Did I remove too many whitespaces here? The following 2 lines needed to have 4 spaces removed, while this one was 6, in order to get it to line up visually the same as the others. How does the spacing make a difference to the tests? The removal of 4 spaces in lines 31-33 seems to have worked for all test instances.

hbushouse:
> We need to bump the version of spherical_geometry up to the latest release to pick up the fix @mcara made. I guess it can be done here.

What's the latest version number? What we have in setup.cfg right now is >=1.2.20

jwst/lib/pointing_summary.py
stscieisenhamer: One can round in doctests by simply printing a truncated version of the results. For this particular example use the following. Using old-style printing to cut down on quotation mark issues.

```
>>> "%.5f" % t["delta_v1"]

Comment (stscieisenhamer) [reply jwst/lib/pointing_summary.py:27]: And of course, fix the comparison to match the precision.

Comment (stscieisenhamer) [reply jwst/lib/pointing_summary.py:27]: and maybe more decimal places? looks like no more than 13.

Comment (hbushouse) [reply jwst/lib/pointing_summary.py:27]: > And of course, fix the comparison to match the precision.

Sorry for being dim, but fix which comparison to match what precision? And where are more decimal places needed? It seems to me that part of the problem we already have is that too many decimal places are being used for the comparisons and we need to cut them back.

Comment (stscieisenhamer) [reply jwst/lib/pointing_summary.py:27]: Hmmm, I see the issue. because it is printing a column. Will have to play but almost assuredly there is a way to limit the precision of the printing of table columns. If not another way is to assign the value then print it. As follows:

```
>>> delta_v1 = t["delta_v1"][0]
>>> %.13f % delta_v1
0.137127271640

nden (approved):

mcara:
This test should have given exact results:

https://github.com/hbushouse/jwst/blob/a7c88e75cf0cc304b16ff3c7e5b91a94a2e58c5d/jwst/lib/tests/test_velocity_aberration.py#L26-L30

It is disappointing that one cannot get exact results anymore.

PR #6545, opened 2022-01-05
JP-2356: Add an option for user to supply wavelength ranges to ignore

### LINKED ISSUE #6515: Add wavelengths to ignore in residual fringe correction
_Issue [JP-2356](https://jira.stsci.edu/browse/JP-2356) was created on JIRA by [Jane Morrison](https://jira.stsci.edu/secure/ViewProfile.jspa?name=morrison):_

In the original documentation the MIRI team provided on the residual fringe correction there was an option to exclude fringe corrections for user supplied wavelength ranges.  This option got left off of the work of putting this step in the pipeline (JP-1273). This PR will add this option. 

From the original documentation provided by Patrick (documentation found in JP-1273)
"As described in Sect 2, fitting residual fringes may erroneously fit and remove features of
astrophysical origin. This is mitigated against by only fitting frequencies within a range around
the peak frequency determined by test measurements, defined in the FRINGEFREQ CDP.
However, a user may wish to explicitly exclude a wavelength range from consideration as a
further safeguard in post-pipeline processing. To this end, the
ResidualFringeCorrection
class can take an optional user input
that holds a 2-column array of lower and upper bounds of the wavelength ranges to be ignored"

jemorrison:
Closes #6515 
Resolves [JP-2356](https://jira.stsci.edu/browse/JP-2356)

**Description**

This PR adds a command line option for the user to give a minimum and maximum wavelength. Data in this wavelength range will be ignored in the residual fringe correction. 

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [X] Milestone
- [X ] Label(s)


jemorrison:
@hbushouse @nden @stscieisenhamer 
I don't think the Jenkins failures are related to this PR.  Can I ignore them or what should I do about them. I did a rebase  with master before I made the pr ?

nden:
@jemorrison Now that #6543 was merged, rebase again on master.

jwst/residual_fringe/residual_fringe.py
```diff
@@ -164,6 +163,14 @@ def do_correction(self):
             y, x = np.mgrid[:ysize, :xsize]
             _, _, wave_map = self.input_model.meta.wcs(x,y)
 
+            # if the user wants to ignore some values use the wave_map array to set the corresponds
```
hbushouse, line 166: should the last word be "corresponding"?

jwst/residual_fringe/residual_fringe_step.py
```diff
@@ -31,6 +33,31 @@ def process(self, input):
...
+
+        min_num = len(ignore_regions['min'])
+
+        if self.ignore_region_max is not None:
+            for region in self.ignore_region_max:
+                ignore_regions['max'].append(float(region))
+        max_num = len(ignore_regions['max'])
+
+        if max_num != min_num:
+            self.log.error("Number of minimum and maximum wavelengths to ignore are not the same")
```
hbushouse, line 53: After issuing the error message, doesn't it then need to abort or something? How can it proceed with mismatched specifications?

hbushouse (commented):
The docs for the step need to be updated to list these new arguments.

jwst/residual_fringe/residual_fringe_step.py
jemorrison: I guess I forgot how self.log.errror works I thought it stopped if it got an error. But I guess not. I will fix that.

jemorrison: fix this I put in a exception

jwst/residual_fringe/residual_fringe_step.py
```diff
@@ -31,6 +33,32 @@ def process(self, input):
...
+
+        ignore_regions['num'] = min_num
+
+        if min_num > 0:
+            self.log.info('Ignoring {} wavelength regions'.format(min_num))
+            raise IncorrectIgnoreRegions("Number of ignore_region_min does not match ignore_region_max")
```
hbushouse, line 59: Shouldn't this exception occur after line 53, where a mismatch between min_num/max_num is detected? This block of code is part of the normal operations.

jwst/residual_fringe/residual_fringe_step.py
```diff
@@ -124,9 +152,17 @@ def process(self, input):
...
                                                            self.residual_fringe_filename,
                                                            self.regions_filename,
+                                                           self.ignore_regions,
                                                            **pars)
             result = rfc.do_correction()
             result.meta.cal_step.residual_fringe = 'COMPLETE'
             self.output_models.append(result)
 
         return self.output_models
+
```
hbushouse, line 162: I don't think you need this special class just to raise an exception above. Just use a standard 

`raise ValueError("Number of ignore_region_min does not match ignore_region_max")` above.

docs/jwst/residual_fringe/arguments.rst
```diff
@@ -3,4 +3,15 @@
 Step Arguments
 ==============
 
-The residual fringe step has no step-specific arguments.
+The residual fringe step has two step arguments that can be used to ignore determining a residual fringe
+correction in a wavelength range. These two parameters must have the same number of elements.
```
hbushouse, line 7: Minor nit, but I think cleaner wording would be: "... two step arguments that can be used to specify wavelength regions in which no correction will be determined. The two arguments give lists of minimum and maximum wavelength values, respectively, for the regions to be ignored. The two lists must contain an equal number of elements."

docs/jwst/residual_fringe/arguments.rst
```diff
@@ -3,4 +3,15 @@
 Step Arguments
 ==============
 
-The residual fringe step has no step-specific arguments.
+The residual fringe step has two step arguments that can be used to ignore determining a residual fringe
+correction in a wavelength range. These two parameters must have the same number of elements. 
+
+``ignore_region_min [list]``
```
hbushouse, line 9: Might want to use different formatting for consistency with other docs, such as:
```
``--ignore_region_min`` (float, default=None)
    The minimum wavelengths for the region(s) to be ignored, given as a comma-separated list of floats.
```

jwst/residual_fringe/residual_fringe_step.py
jemorrison: Yes. My stupid mistake

jemorrison:
@hbushouse  Does this make sense. When I try and run residual_fringe correction I still have to use the --override_fringefreq.  I thought it was just not in CRDS yet.  But it should be. In  CRDS it lists a type as 'fringe' .  I think that should have been 'fringefreq'. I changed the code to use
reference_file_types = ['fringe','regions']. It still does not work. Do I need to go and change the datamodel to 'fringe'

jemorrison:
@hbushouse Ok on further inspect that 'fringe' must be the fringe flat. the fringefreq is not in CRDS. Is it on a test CRDS. So we just have to use the override_fringefreq for now.

hbushouse:
> @hbushouse Ok on further inspection that 'fringe' must be the fringe flat. the fringefreq is not in CRDS. Is it on a test CRDS. So we just have to use the override_fringefreq for now.

If I change both instances of "fringefreq" to just "fringe" in the residual_fringe_step.py module and run using the default CRDS server and context, it successfully finds and retrieves a jwst_miri_fringe_xxxx.fits ref file to use.

hbushouse:
The fringefreq ref files are currently in https://jwst-crds-test.stsci.edu/context_table/jwst_0780.pmap, so yes it's the test server. If you set your CRDS_SERVER environment variable, you should be able to get to them when running the pipeline/step. I also noticed that there's no info anywhere in our readthedocs about either the residual_fringe step or the fringefreq ref file (at least not that I could find).

hbushouse (approved):
This all looks good to me now and all of the CI tests passed on the latest run.

jemorrison:
@hbushouse I will put in a separate PR to describe the fringefreq reference file. That file is complicated and I would like to figure out my problem building the docs so I can review it as I write it

hbushouse:
@jemorrison If you're OK with current commits (any last minute changes you want to make yet?), we can go ahead and merge.

PR #6546, opened 2022-01-05
JP-2154: Clarify wording in skymatch docs

### LINKED ISSUE #6161: Skymatch step documentation seems incorrect
_Issue [JP-2154](https://jira.stsci.edu/browse/JP-2154) was created on JIRA by [Mattia Libralato](https://jira.stsci.edu/secure/ViewProfile.jspa?name=libra):_

The documentation of the skymatch step of the pipeline ([here|#limitations-and-discussions]]) states: 

"Since sky subtraction is performed on flat-fielded but not distortion corrected images, it is important to keep in mind that flat-fielding is performed to obtain uniform surface brightness and not flux. This distinction is important for images that have not been distortion corrected. As a consequence, it is advisable that point-like sources be masked through the user-supplied mask files.”

I am not sure this part is correct. The pixel-area map is already applied to the stage-2 cal.fits images, thus the value of each pixel already takes into account for the effects of the distortion. Even if this is not true, the problem/limitation described above would be also true for a pixel that is not part of a point source but, for example, of the background.

Could these lines be something left from the HST pipeline documentation ([here](https://stsci-skypac.readthedocs.io/en/latest/skymatch.html))? For HST, flc/flt images are flat-fielded but not distortion corrected and their pixel values are not corrected for pixel area. Thus, the HST pipeline needs to correct for the pixel-area differences when creating the drc/drz images.

tapastro:
Closes #6161 
Resolves [JP-2154](https://jira.stsci.edu/browse/JP-2154)

**Description**

This PR addresses some possibly confusing wording in the skymatch documentation. 

Checklist
~~- [ ] Tests~~
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -306,6 +306,9 @@ skymatch
 - Updated code in ``skymatch.region.py`` with latest improvements and bug fixes
   from ``stsci.skypac.regions.py``. [#6451]
 
+- Updated documentation to clarify details of flat-fielding versus distortion
+  corrections [#6546]
+
 source_catalog
 --------------
 
--- a/docs/jwst/skymatch/README.rst
+++ b/docs/jwst/skymatch/README.rst
@@ -172,13 +172,12 @@ the outcome of the matching (sky subtraction in general) algorithm:
 
 * Because sky subtraction is performed on *flat-fielded* but
   *not distortion corrected* images, it is important to keep in mind
-  that flat-fielding is performed to obtain correct surface brightness
-  for all pixels. Extracted source measurements will be affected by sampling
-  and distortion, however, and so distortion corrections will be necessary
-  before such extraction occurs. As a consequence, it is advisable that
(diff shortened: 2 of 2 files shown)


mcara (approved):
LGTM

CHANGES.rst
```diff
@@ -306,6 +306,9 @@ skymatch
 - Updated code in ``skymatch.region.py`` with latest improvements and bug fixes
   from ``stsci.skypac.regions.py``. [#6451]
 
+- Updated documentation to clarify details of flat-fielding versus distortion
+  corrections [#6546]
```
mcara, line 310: I do not believe documentation updates need a changelog entry. What is the policy with regard to this in this repository, @hbushouse

hbushouse: Yes, it's fine (and encouraged) to have change log entries for doc updates.

PR #6548, opened 2022-01-07
JP-2371 Remove MIR_FLATMRS from Asn_Lv3MIRMRS rule

### LINKED ISSUE #6538: Eliminate flats from spec3 ASN tables
_Issue [JP-2371](https://jira.stsci.edu/browse/JP-2371) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

spec3 type ASN files for MIRI MRS observations that contain flat-field (lamp) exposures currently include entries for the flat exposures, in addition to the science exposures. This is causing problems in SDP workflows, because the flat products are not getting copied to the level-3 input directory and hence ASN completeness is not satisfied, so processing doesn't get triggered for the ASN.

The level-3 ASN rules for spec3 associations should be updated to **not** include the flat exposures to begin with. They are not needed for any level-3 processing anyway.

stscieisenhamer:
Closes #6538
Resolves [JP-2371](https://jira.stsci.edu/browse/JP-2371)

**Description**

Do not add MIR_FLATMRS exposures to the Level 3 associations for MIRI MRS observations.

Checklist
- [x] Tests
- [ ] ~Documentation~
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,8 @@ assign_wcs
 associations
 ------------
 
+- Remove MIR_FLATMRS from Asn_Lv3MIRMRS rule. [#6548]
+
 - Fix bug causing ``pytest`` to encounter an error in test collection when
   running with recent commits to ``astropy`` main (``5.0.dev``). [#6176]
 
--- a/jwst/associations/lib/rules_level3.py
+++ b/jwst/associations/lib/rules_level3.py
@@ -294,11 +294,7 @@ class Asn_Lv3MIRMRS(AsnMixin_Spectrum):
             DMSAttrConstraint(
                 name='exp_type',
                 sources=['exp_type'],
-                value=(
-                    'mir_mrs'
-                    '|mir_flatmrs'
-                ),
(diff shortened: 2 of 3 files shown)


hbushouse (approved):
Looks good to me, assuming all CI tests pass. Could use a little more explanation of the change in the "description" field of the PR.

hbushouse:
Oh, and what about any related documentation? I wonder if flats are mentioned anywhere in the docs and, if so, are they mentioned in the context of being in level 3 ASN's? (which would need changing now)

stscieisenhamer:
Did go through everything, and there is nothing Level 3 specific. The only extensive set of documentation is for Level 2. There are a number of issue, one of which is JP-1059, to fill out the documentation for individual rules.

hbushouse:
Clean test run (at least for the ones we care about). Merging.

stscieisenhamer:
i've already have okify data...will do so momentarily.

Issue #5091, opened 2020-06-19
1D spectroscopic extraction upgrades for NIRISS SOSS


stscijgbot-jp:
_Issue [JP-1550](https://jira.stsci.edu/browse/JP-1550) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

Implement the enhanced 1D extraction techniques for NIRISS SOSS mode as outlined by the NIRISS IDT in JP-201.

stscijgbot-jp:
_Comment by [Anton Koekemoer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=koekemoe) on [JIRA](https://jira.stsci.edu/browse/JP-1550?focusedCommentId=512559#comment-512559):_

Set priority of this ticket to calwg_critical to match the already existing priority of the corresponding ticket JP-201

stscijgbot-jp:
_Comment by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse) on [JIRA](https://jira.stsci.edu/browse/JP-1550?focusedCommentId=565309#comment-565309):_

Baseline code commit in #6467

stscijgbot-jp:
_Comment by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse) on [JIRA](https://jira.stsci.edu/browse/JP-1550?focusedCommentId=566628#comment-566628):_

It has been discovered that the baseline code for the new extraction algorithm is intended to work on data that are in units of DN/sec, i.e. not flux calibrated. This requires a change to the pipeline work flow, such that the photom step in level-2b calibration is turned off for SOSS exposures. There may be a desire to then apply flux calibration to the 1D spectrum after it's been extracted. This will require some small modifications to the photom step (to handle 1D data as input for SOSS, rather than 2D). An investigation also needs to be done to see if having the 2D SOSS data in units of DN/s will have any effects on the early steps of the level-3 pipeline, before 1D extraction is performed.

A baseline version of the new extraction technique is included in the B7.9 Cal pipeline code, but because there wasn't sufficient development time remaining to make the above changes to flux calibration, the old "boxcar" extraction method is still applied to SOSS data by default. An option was added to the extract_1d step to allow the new method to be applied during off-line processing, with the caveat that the overall scale of the resulting 1D spectrum will be incorrect, due to the flux calibration.

[...]

PR #6551, opened 2022-01-07
JP-1550: Turn off ATOCA by default

tapastro:
Related to [JP-1550](https://jira.stsci.edu/browse/JP-1550) and #5091 

**Description**

This PR turns off the ATOCA SOSS extraction algorithm by default, as more testing with the photom step turned off will be required before we can verify reasonable results. The algorithm can be activated using a temporary extract_1d step parameter soss_atoca, which defaults to False and triggers box extraction. Additionally, this removes the source_type attribute from the ATOCA output to prevent creation of an extraneous SCI extension in the x1dints output.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -176,6 +176,9 @@ extract_1d
 - Delivery of new algorithm `ATOCA` for SOSS extraction, along with four new reference
   files: speckernel, specprofile, spectrace and wavemap. [#6467]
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):
Looks good. Should have a change log entry.

tapastro:
> Looks good. Should have a change log entry.

Pushed the changelog commit to the wrong branch, oops.

PR #6552, opened 2022-01-07
Fix jump tests

cshanahan1:
Fixes unit test `test_nirspec_saturated_pix` for jump detection to account for changes in https://github.com/spacetelescope/stcal/pull/72

In this test, a test dataset is constructed and four pixels ((1, 1), (2, 2), (3, 3), 4,4)) are given test scenarios to see if flagging for saturated pixels is done correctly. 

The tests for pixels (1, 1) and (3, 3) did not have different results - while the internal calculations changed based on changes to the code, the resulting pixels flagged as jumps ended up being the same.

for the test case of pixel (2, 2), 5 of the 7 groups in the test data are set to be saturated, meaning there is only one non saturated diff so no jumps should be flagged. previously, it was incorrectly flagging the 3rd group as a jump when NO groups should have been flagged, so the truth value of this test has changed. I kept this test but changed the truth to assert that NO groups are flagged in this case. 

pixel(4, 4) was the same case as (2, 2), where 5 out of 7 groups were saturated so no jumps should be flagged as well.. I deleted this redundant example.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -193,6 +193,12 @@ jump
 
 - Updated jump detection step to use common code moved to stcal [#6089]
 
+- In stcal (pr #72), several changes were made to fix existing bugs in the
+  twopoint difference routine for jump detection. Some of these issues
+  resulted in jumps erroneously being flagged for pixels with only two
+  usable groups (i.e one usable difference). This PR on the JWST side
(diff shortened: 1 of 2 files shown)


hbushouse (approved):
LGTM

PR #6567, opened 2022-01-13
Update CRDS dependency

stscieisenhamer:
**Description**

Update the CRDS dependency to the Build 7.9 versions of CRDS

Checklist
- [ ] Tests
- [ ] ~Documentation~
- [ ] ~Change log~
- [x] Milestone
- [x] Label(s)


setup.cfg
```diff
@@ -31,7 +31,7 @@ install_requires =
     asdf>=2.8.2
     astropy>=5.0.0
     BayesicFitting>=2.7.2
-    crds>=11.5.2
+    crds~=11.6.0
```
hbushouse, line 34: For my own education, what kind of comparison does the "~" do? Is that like any 11.6.x subversion?

hbushouse (approved):
Looks OK to me.

setup.cfg
hbushouse: So I guess you could use ">=11.6.0" to get any 11.6.x subversion, but that would also allow future 11.7, 12.0, etc. releases to satisfy the condition. This way you're still restricting it to 11.6.x only. Interesting.

stscieisenhamer: Feel more comfortable with this, at least for Build releases. development and master can be reverted to a looser restriction.

PR #6568, opened 2022-01-13
SPB-778 Bugfix: JWST dependency uploads occuring with failed regressions

shahm0:
SPB-778 Added a failure threshold so that failed regressions do not trigger an update during the build process

Resolves [SPB-778](https://jira.stsci.edu/browse/SPB-778)

**Description**

This PR fixes the creation of python dependency files by jwst regression test runs, so that the dependency files are only created and uploaded to artifactory when there were no failures in the regression test run. Right now they're being created for every run.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -90,6 +90,7 @@ bc0.test_cmds = [
     'codecov --token=${codecov_token} -F nightly',
 ]
 bc0.test_configs = [data_config]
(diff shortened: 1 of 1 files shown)


jneu (approved):
lgtm, but wait for @stscieisenhamer to review as well

PR #6580, opened 2022-01-18
JP-2024: Update default value of skymatch's 'skymethod' step parameter

### LINKED ISSUE #5923: Should resample automatically subtract background?
_Issue [JP-2024](https://jira.stsci.edu/browse/JP-2024) was created on JIRA by [James Davies [X]](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

I believe the answer should be "no".  But currently `resample` does subtract off the sky background in `calwebb_image3` when it has background measurements available.  A scalar sky value gets computed earlier on in the stage3 pipeline via `skymatch`.

The single-resampled images produced at the end of the stage2 pipeline do not have the background subtracted, as they don't run through `skymatch`.

The current implementation checks if `skymatch` has computed a background level for each exposure.  If it has, it subtracts it before resampling.  So the output resampled images currently have a mean sky background of zero.

I suspect what it should do is subtract off those computed sky backgrounds for each exposure, drizzle into the common output image, and the add back the sky background after drizzling is completed, so that the output drizzled image has the same background as the input images.

But which sky background from which input image should be added back in?  The mean of the input sky backgrounds?  The minimum?  Some weighted mean?

I'm primarily concerned about what should be the default for SPD processing of the associations that we will be producing. I suspect all exposures in these associations will be taken close to the same time and should thus have pretty similar zodiacal light characteristics so pretty similar backgrounds.

tapastro:
Closes #5923 
Resolves [JP-2024](https://jira.stsci.edu/browse/JP-2024)

**Description**

This PR addresses background subtraction issues found in the resample step due to its interaction with the default skymatch step parameter skymethod being set to 'global+match'. This was found to have a simple solution of changing the default to 'match'.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 1.4.2 (Unreleased)
 ==================
 
+skymatch
+--------
+
+- Changed default value of ``skymethod`` step parameter to 'match' [#6580]
 
 1.4.1 (2022-01-15)
 ==================
--- a/docs/jwst/skymatch/README.rst
+++ b/docs/jwst/skymatch/README.rst
@@ -90,7 +90,7 @@ The ``skymatch`` step has the following optional arguments:
 * ``skymethod`` (str):
   The sky computation algorithm to be used.
   Allowed values: {``local``, ``global``, ``match``, ``global+match``}
-  (Default = ``global+match``)
+  (Default = ``match``)
 
 * ``match_down`` (boolean):
   Specifies whether the sky *differences* should
--- a/jwst/skymatch/skymatch_step.py
+++ b/jwst/skymatch/skymatch_step.py
@@ -34,7 +34,7 @@ class SkyMatchStep(Step):
 
     spec = """
         # General sky matching parameters:
-        skymethod = option('local', 'global', 'match', 'global+match', default='global+match') # sky computation method
+        skymethod = option('local', 'global', 'match', 'global+match', default='match') # sky computation method
         match_down = boolean(default=True) # adjust sky to lowest measured value?
         subtract = boolean(default=False) # subtract computed sky from image data?
 


hbushouse (commented):
Changes look trivial enough and are appropriate. But now I have a question for resampling/drizzling experts @mcara @stsci-hack : will this change have any kind of undesirable effect on subsequent steps like `outlier_detection` and `resample`? If I remember correctly, `outlier_detection` uses the previously computed sky values to temporarily remove sky signal from each image when it's doing its comparisons to find outliers. Now that the value computed by `skymatch` is only the delta between images, that subtraction will only equalize the images, but will leave a global sky signal in them. Is that OK? Is it good enough to at least have the images equalized, so that comparisons are valid when looking for outliers?  Any similar potential issues with anything in resampling/drizzling?

mcara:
> background subtraction issues found in the resample step due to its interaction with the default skymatch step parameter skymethod being set to 'global+match'.

I would be helpful to know what the issues are. From reading #5923 it is not obvious (at least not to me who did not participate in the meetings) what is the real issue that is being solved here.

To answer the question:

1. From the "pure" resample point of view - there should not be any difference between "global+match" and "match". Both should "equalize" background levels in the images making them suitable for resampling. [Whether this is done or not - I do not know. I was not involved in the development of the resample and/or CR-rejection steps]
2. I think "global+match" has potential to do more harm for the CR-rejection step than just "match". It all depends on the implementation of that code.

I will post more comments on the #5923 issue. But this PR looks fine in achieving what was decided in #5923.

mcara (approved):

stscieisenhamer (approved):

hbushouse:
I'm satisfied from comments and testing that this change is fine and should not have any negative consequences on other steps. outlier_detection will still do comparisons of equalized images, because it accounts for the BKGLEVEL internally, it'll just be comparing images that have higher overall signal levels than before. And the combined results of resample/drizzle will be fine, without any mismatches between images, but again will just have the global sky still in them (which is apparently what INS wants).

hbushouse (approved):

hbushouse:
Failing CI test is totally unrelated. Merging.

PR #6681, opened 2022-01-19
JP-2495: Change restriction on Level-2b association creation to nrc_image instead of nrc_tacq

### LINKED ISSUE #5655: adjust AsnGenerator rules to avoid Level2b table creation for workflows that stop at Level1b
_Issue [JP-2495](https://jira.stsci.edu/browse/JP-2495) was created on JIRA by [Mike Swam](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mswam):_

As of DMS b7.7, certain instrument modes will only process through Level1b, and not proceed on to Level2a.  In some situations those datasets still get a Level2b association table written by AsnGenerator, which will never result in calling calibration because the requisite input Level2a files will never be generated.  The AsnGenerator rules should be adjusted to exclude creating Level2b asn tables for this particular instrument mode that will stop at Level1b:

EXP_TYPE = NRC_IMAGE  and 
DMS_NOTE = WFSC_LOS_JITTER

This should have been the requested behavior under JP-1872, but that ticket description was written incorrectly, so the rule change made under that ticket can be backed-out in favor of the one above.

tapastro:
Closes #5655 
Resolves [JP-2495](https://jira.stsci.edu/browse/JP-2495)

**Description**

This PR prohibits the creation of level 2b associations if `dms_note` is `wfsc_los_jitter` and `exptype` = `nrc_image`, rather than the first implementation that prohibited associations for  `exptype` = `nrc_tacq`.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 1.4.2 (Unreleased)
 ==================
 
+associations
+------------
+
+- Changed restriction on Level2b creation for ``NRC_TACQ`` exposures
+  to ``NRC_IMAGE`` to allow asn creation for tacq but not science [#6681]
+
 outlier_detection
 -----------------
 
--- a/jwst/associations/lib/rules_level2b.py
+++ b/jwst/associations/lib/rules_level2b.py
@@ -112,7 +112,7 @@ class Asn_Lv2ImageNonScience(
                     DMSAttrConstraint(
                         name='exp_type',
                         sources=['exp_type'],
-                        value='nrc_tacq'
+                        value='nrc_image'
                     ),
 
                 ],


stscieisenhamer (approved):
LGTM modulo changelog

tapastro:
In the original ticket, I mentioned adding a new pool to Artifactory to test this addition to the association rules, but I didn't seem to include the updated pool name in the regtest. Should I add that in this ticket? I have it on hand, just never delivered.

tapastro:
Scratch that, I forgot how associations are pulled into the regtest - I didn't see it on the "list", but the list is only pools for which "special handling" is required. I checked the last run on Jenkins and it's listed as passing (the specific pool is jw00632_20210921t193255_pool.csv). I think this should cause some regtest failures that will need okify'ing, but this PR should be ready to merge if that's the plan.

hbushouse:
Can't merge without a change log entry ...

hbushouse (approved):

PR #6696, opened 2022-01-27
Clean up Extract1d logs for multi-integration extractions

tapastro:
Closes #----
Resolves [JP-nnnn](https://jira.stsci.edu/browse/JP-nnnn)

**Description**

This PR makes the log more readable for extractions with many to hundreds of integrations, removing repeated warnings and including the integration in the default log output.

Can create an issue if needed for bookkeeping.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -2887,7 +2887,7 @@ def do_extract1d(
                     log.info("Spectral order 0 is a direct image, skipping ...")
                     continue
 
-                log.info(f'Processing spectral order {sp_order}')
+                log.debug(f'Processing spectral order {sp_order}')
 
                 try:
                     output_model = create_extraction(
@@ -2932,7 +2932,7 @@ def do_extract1d(
                         log.info("Spectral order 0 is a direct image, skipping ...")
                         continue
 
-                log.info(f'Processing spectral order {sp_order}')
+                log.debug(f'Processing spectral order {sp_order}')
 
                 try:
                     output_model = create_extraction(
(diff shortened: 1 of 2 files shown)


jwst/extract_1d/extract.py
```diff
@@ -2887,7 +2887,7 @@ def do_extract1d(
                     log.info("Spectral order 0 is a direct image, skipping ...")
                     continue
 
-                log.info(f'Processing spectral order {sp_order}')
+                log.debug(f'Processing spectral order {sp_order}')
```
hbushouse, line 2890: What's the order of looping within the overall extract_1d process, when you have multiple sources, multiple integrations, and multiple orders? I like the idea of seeing the looping over spectral orders, but perhaps not if it's repeated for every integration.

jwst/extract_1d/extract.py
```diff
@@ -3372,6 +3372,7 @@ def extract_one_slit(
         exp_type = slit.meta.exposure.type
 
     if integ > -1:
+        log.info(f"Extracting integration {integ}")
```
hbushouse, line 3375: bump this up by 1 to make the log message use 1-indexed counting

jwst/extract_1d/extract1d.py
```diff
@@ -263,11 +263,11 @@ def extract1d(image, var_poisson, var_rnoise, var_flat, lambdas, disp_range,
 
             if bkg_npts == 0:
                 bkg_model = None
-                log.warning(f"Not enough valid pixels to determine background "
+                log.debug(f"Not enough valid pixels to determine background "
```
hbushouse, line 266: I think this seems appropriate. It could be useful info for debugging or tracking down exactly what happened with a given spectrum, but not part of the routine logging.

hbushouse (commented):
Looks good to me. Just needs a change log entry.

PR #6699, opened 2022-01-31
JP-2494: Change association 'asn_pool' to include file extension

### LINKED ISSUE #6680: ASNPOOL keyword should include the extension
_Issue [JP-2494](https://jira.stsci.edu/browse/JP-2494) was created on JIRA by [Lisa Gardner](https://jira.stsci.edu/secure/ViewProfile.jspa?name=gardner):_

The ASNPOOL keyword should include the file extension.
Example: jw00625_20210713t194701_pool is the value ASNPOOL in the FITS header.

The file name of the actual Association Pool file being ingested would be like jw00625_20210713t194701_pool.csv. 

The ASNTABLE keyword does include the the extension so the change to ASNPOOL keyword will make it consistent with the ASNTABLE keyword.

tapastro:
Closes #6680 
Resolves [JP-2494](https://jira.stsci.edu/browse/JP-2494)

**Description**

This PR adds the file extension to the `asn_pool` entry in associations during their creation. I tested this with asn_generate locally, but I don't know that I've covered all cases.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/associations/lib/rules_level3_base.py
+++ b/jwst/associations/lib/rules_level3_base.py
@@ -242,9 +242,9 @@ class DMS_Level3_Base(DMSBaseMixin, Association):
             if self.data['asn_pool'] == 'none':
                 self.data['asn_pool'] = basename(
                     item.meta['pool_file']
-                ).split('.')[0]
+                )
                 parsed_name = re.search(
-                    _DMS_POOLNAME_REGEX, self.data['asn_pool']
+                    _DMS_POOLNAME_REGEX, self.data['asn_pool'].split()[0]
                 )
                 if parsed_name is not None:
                     pool_meta = {
--- a/jwst/associations/lib/rules_level3_base.py
+++ b/jwst/associations/lib/rules_level3_base.py
@@ -244,7 +244,7 @@ class DMS_Level3_Base(DMSBaseMixin, Association):
                     item.meta['pool_file']
                 )
                 parsed_name = re.search(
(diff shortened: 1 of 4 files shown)


tapastro:
Just realized as I got home that I didn't give split an input, though I don't know if that code is used either. I'll fix it when I get in tomorrow!

stscieisenhamer (approved):
LGTM

jwst/associations/lib/rules_level3_base.py
```diff
@@ -242,9 +242,9 @@ def update_asn(self, item=None, member=None):
             if self.data['asn_pool'] == 'none':
                 self.data['asn_pool'] = basename(
                     item.meta['pool_file']
-                ).split('.')[0]
+                )
                 parsed_name = re.search(
-                    _DMS_POOLNAME_REGEX, self.data['asn_pool']
+                    _DMS_POOLNAME_REGEX, self.data['asn_pool'].split()[0]
```
hbushouse, line 247: Yep, here's where you need to have an argument for `split`, i.e. `split(".")`

hbushouse (commented):
Other than the missing argument to `split` (as you already realized), it looks good.

hbushouse:
> LGTM

Look more carefully at https://github.com/spacetelescope/jwst/pull/6699/files#diff-25f0067c8586a7bb44a877e3454c1441e1410efa812d3080dfeda9788a68d796R247

PR #6709, opened 2022-02-02
JP-2508: Exclude NRC_TSGRISM + module B exposures from association creation

### LINKED ISSUE #6705: Update level2b ASN rules for NRC grism Module B subarray exposures
_Issue [JP-2508](https://jira.stsci.edu/browse/JP-2508) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

As described in JSOCINT-576 and JWSTDMS-632, some NIRCam exposures during commissioning will be using Module B subarrays that are normally dedicated to the NRC_TSGRISM observing mode. Currently, Module B is not supported for science observations with the NRC_TSGRISM mode and hence when such exposures reach the assign_wcs step in the calwebb_spec2 pipeline, the pipeline aborts with an error message indicating that they're not supported.

Fortunately the use of these exposures during commissioning only requires level-1b and level-2a products. So we need to update the level-2b ASN rules to exclude construction of any ASN's for these kinds of exposures, which will effectively end processing at level-2a.

tapastro:
Closes #6705 
Resolves [JP-2508](https://jira.stsci.edu/browse/JP-2508)

**Description**

This PR adds additional constraints to levels 2 and 3 association creation for exposures with exp_type='nrc_tsgrism' and detector='nrcblong'. 

Milestone?

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


jwst/associations/lib/rules_level2_base.py
```diff
@@ -794,6 +794,19 @@ def __init__(self):
...
+                    DMSAttrConstraint(
+                        sources=['detector'],
+                        value='nrcblong'
+                    ),
+                    DMSAttrConstraint(
+                        sources=['exp_type'],
+                        value='nrc_tsgrism'
+                    )
+                ],
+                reduce=Constraint.notall
```
hbushouse, line 808: Are you sure this should use `notall` as opposed to `notany`?  I have a hard time remembering exactly how the logic works.

tapastro: I think I'll have to defer to @stscieisenhamer on this, when he gets a chance - my thought was that exclusion is needed if both constraints are met, i.e. don't constrain nrc_tsgrism entries that have detector != nrcblong, but constrain entries with both - so not all constraints? But that goes against the logic in the constraint above the one I added that uses notany, so I may not have full comprehension...

tapastro: An explanation that satisfies me - the existing constraint using `notany` above the new constraint has two members, one of which looks for 'nirspec' in the detector column. I can't find an instance of a pool file with 'nirspec' listed in detector, just NRS1/NRS2 in nirspec's case. So as written, I think it's acting as a single constraint looking for 'filter'=='opaque' and excluding.

hbushouse: Sounds like a bug to me. That 'nirspec' constraint should be looking at 'instrument', not 'detector'. But then the `notall` vs `notany` logic may need changing too.

tapastro: Should I fix it here or separate into another ticket?

hbushouse:
Looking at the APT for program 809, obs 31 uses full-frame and hence doesn't get any TSO modes mistakenly assigned. Obs 32 and 33 use Subgrism256 and hence get TSGRISM assigned, but it's only Module A, so they're OK. Obs 34 uses Subgrism64 and hence get TSGRISM assigned, but it's only Module A, so it's OK. Obs 36 and 37 use Module B with Full-frame, so they don't get TSGRISM assigned.

Obs 35 is the only one that uses Module B and Subgrism256, so the exposures get TSGRISM assigned. There are 2 exposures, due to the use of a 2x1 mosaic special requirement. So the original ASN rules produced 2 spec2 ASN's (one for each of the 2 exposures), and 1 tso3 ASN, for the combined level-3 processing.

The new rules still create all of the same ASN's for all obs EXCEPT 35, where the 2 spec2 and 1 tso3 ASN's are now missing. This is exactly as desired.

tapastro:
This PR does not affect any of the tests, implying a lack of coverage - presumably this is a niche issue and doesn't require a test? Or should a pool be added to the burgeoning association test directory?

hbushouse (approved):

PR #6720, opened 2022-02-08
Add ability to override source type in Source Type step (JP-2155)

### LINKED ISSUE #6162: Add ability to override source type in Source Type step
_Issue [JP-2155](https://jira.stsci.edu/browse/JP-2155) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

A common issue when testing MIRI MRS simulated data is that we have to override the source type assigned by the SRCTYPE step since the relevant APT keywords are not present in simulated data.  So far, I've done this by hacking the FITS header keywords directly after the SRCTYPE step runs, but from reviewing various GO programs it has become clear that some science programs will want to also override this keyword (e.g., point sources in programs that would otherwise look like extended sources).

It would be much cleaner to be able to pass a manual override step-specific argument to the SRCTYPE step declaring a specific source type rather than relying on directly editing the FITS files in the middle of a pipeline run.  It seems like this could be of broader utility beyond MRS as well, albeit at low priority.

dmggh:
Closes #6162
Resolves [JP-2155](https://jira.stsci.edu/browse/JP-2155)

**Description**
This update allows the SRCTYPE to be overridden on the command line when processing exposures that contain a single pre-defined target, i.e. MIR_LRS-FIXEDSLIT, MIR_LRS-SLITLESS, MIR_MRS, NRC_TSGRISM, NRS_FIXEDSLIT, NRS_BRI
GHTOBJ, and NRS_IFU. Other EXP_TYPEs will be ignored. The change log and documentation will soon also be updated. This PR supersedes #6692 .

Checklist
- [X] Tests
- [ ] Documentation
- [ ] Change log
- [X] Milestone
- [X] Label(s)

### DIFF
--- a/jwst/srctype/srctype.py
+++ b/jwst/srctype/srctype.py
@@ -7,9 +7,10 @@ log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
 
-def set_source_type(input_model):
+def set_source_type(input_model, source_type=None):
     """
-    Set the source_type, based on APT input or default values.
+    Set source_type based on APT input, user specification, exposure type,
+    or default values.
 
     Parameters
     ----------
@@ -17,6 +18,9 @@ def set_source_type(input_model):
                   `~jwst.datamodels.IFUImageModel`, `~jwst.datamodels.MultiSlitModel`,
                   or `~jwst.datamodels.SlitModel`
         The data model to be processed.
+
+    source_type : str,  {POINT, EXTENDED}
+        User-requested value for source type.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


hbushouse (commented):
Code updates look good. Just needs a change log entry and updates to the step docs to give info about this new step param.

dmggh:
I would prefer to merge this as is, as soon as possible. I will then include the updated docs and change log, as I wrote in the description. Trying to include them all in the same PR is what I tried earlier, but it resulted in several days delay and a throughly mangled PR as I tried to convince GitHub to realize that I had correctly updated the changelog.

hbushouse:
> I would prefer to merge this as is, as soon as possible. I will then include the updated docs and change log, as I wrote in the description. Trying to include them all in the same PR is what I tried earlier, but it resulted in several days delay and a throughly mangled PR as I tried to convince GitHub to realize that I had correctly updated the changelog.

Adding additional commits is usually not a problem, but I guess in this case we can go ahead and do them separately. Merging this one now.

dmggh:
Thanks.

PR #6722, opened 2022-02-08
 JP-2515: Workaround for excessive align_refs logging for MIRI coronagraphy

### LINKED ISSUE #6724: Fix align_refs warnings for large numbers of DO_NOT_USE pixels
_Issue [JP-2515](https://jira.stsci.edu/browse/JP-2515) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

[Jonathan Aguilar](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jaguilar) reports:

When I run align_refs on simulated data, there’s a median filter than runs and throws a warning if the median comes back NaN. This warning is passed to the logger. Unfortunately because of the way the coronagraph subarrays are set up on MIRI, this warning gets thrown over 10,000 times and the logging really slows down the step. Like, by a factor of 1000. Is there a way we can make this median filter a little smarter?

aggle:
Closes #6724 
Resolves [JP-2515](https://jira.stsci.edu/browse/JP-2515)

**Description**

The MIRI coronagraphs have large regions of pixels marked NON_SCIENCE, which trigger thousands of logging messages when a median filter is passed over them in the calwebb_coron3.align_refs stage. These median filter sets these pixels to 0, so this fix identifies these pixels and sets them to 0 before the median filter is executed. This speeds up the time it takes to complete the align_refs stage from one or more days on a Macbook Pro to a few minutes or less.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


jwst/coron/median_replace_img.py
```diff
@@ -47,7 +48,7 @@ def median_fill_value(input_array, input_dq_array, bsize, bad_bitvalue, xc, yc):
         dq_array = input_dq_array[xc - hbox:xc + hbox + 1, yc - hbox: yc + hbox + 1]
     except IndexError:
         # If the box is outside the data return 0
-        log.warning('Box for median filter is outside the data.')
+        # log.warning('Box for median filter is outside the data.')
```
hbushouse, line 51: Should reinstate this warning for the final version. Or does it get triggered a lot for some kinds of data?

jwst/coron/median_replace_img.py
```diff
@@ -81,6 +82,7 @@ def median_replace_img(img_model, box_size, bad_bitvalue):
     Returns
     -------
     img_model : image model
+
```
hbushouse, line 85: The extra blank line is not needed here.

jwst/coron/median_replace_img.py
```diff
@@ -89,9 +91,15 @@ def median_replace_img(img_model, box_size, bad_bitvalue):
         img_int = img_model.data[nimage]
         img_dq = img_model.dq[nimage]
         bad_locations = np.where(np.bitwise_and(img_dq, bad_bitvalue))
-
+        # if it's MIRI, only use the bad locations that are in the science aperture
+        # set the others to 0 with no logging message
+        # bad_locations, zero_locations = filter_bad_locations(img_model, bad_locations)
+        bad_locations, non_science = separate_non_science_pixels(img_dq, bad_locations)
+        img_model.data[nimage, non_science[0], non_science[1]] = 0
```
hbushouse, line 98: I'm confused by the fact that values in `img_model.data` are being set to zero here, but later in line 105 it's `img_int` that's passed to the `median_fill_value` function, which has not had any values reset. Is that correct? Or does it work out properly because `img_int` is just a view of (pointer to) `img_model.data` and hence it has the same values?

jwst/coron/median_replace_img.py
```diff
@@ -101,3 +109,34 @@ def median_replace_img(img_model, box_size, bad_bitvalue):
...
+    """
+    For the median filter, we don't care about the NON_SCIENCE pixels, but they
+    produce a ton of warnings that clog up the alignment algorithm and make it
+    run really slowly. In this function, we take all the bad pixels and pull out
+    the ones with NON_SCIENCE flags so we can set them to 0 without running the
+    median filter.
+
+    Parameters
+    ----------
+    img_dq: 2-D dq image
```
hbushouse, line 124: Formatting of the parameter definitions here should follow the same syntax as what's seen above for `median_replace_img`

jwst/coron/median_replace_img.py
```diff
@@ -101,3 +109,34 @@ def median_replace_img(img_model, box_size, bad_bitvalue):
...
+    Output
+    ------
+    Tuple of (bad_locations, non_science) pixels
+
+    """
+    def is_science(pix):
+        # return True if pixel is for science, False if flagged NON_SCIENCE
+        val = img_dq[pix[0], pix[1]]
+        flags = dqflags.dqflags_to_mnemonics(val, dqflags.pixel)
+        if "NON_SCIENCE" in flags:
```
hbushouse, line 136: Alternative method here, which might be faster than using `dqflags_to_mnemonics`, would be something like:
```
flags = np.bitwise_and(val, dqflags.pixel['NON_SCIENCE'])
if flags.any():
    return False
...
```

hbushouse (commented):
Overall approach looks good. Just a couple questions/comments.

jwst/coron/median_replace_img.py
aggle: Oops, you're correct. That's not the warning I'm worried about. I must have commented it out by accident. Thanks very much for catching it.

jwst/coron/median_replace_img.py
aggle: Removed

jwst/coron/median_replace_img.py
aggle: Ah, that's a good point. It works because it's all passed by reference, but I changed img_model.data to img_int for consistancy.

jwst/coron/median_replace_img.py
aggle: Fixed.

jwst/coron/median_replace_img.py
aggle: Thanks for the suggestion!

aggle:
The failing/error tests don't seem to have anything to do with MIRI coronagraphy, or to be calling my changes.

jwst/coron/median_replace_img.py
```diff
@@ -89,9 +90,16 @@ def median_replace_img(img_model, box_size, bad_bitvalue):
         img_int = img_model.data[nimage]
         img_dq = img_model.dq[nimage]
         bad_locations = np.where(np.bitwise_and(img_dq, bad_bitvalue))
-
+        # if it's MIRI, only use the bad locations that are in the science aperture
```
hbushouse, line 93: The comment suggests that this new processing is intended only for MIRI data, but there's no logic that determines whether it's MIRI data. If it's equally OK for NIRCam (or any coronagraphic data on which this might get used at some point in the future), then perhaps the mention of MIRI should be removed. Right now the new code is in fact being applied to all inputs.

jwst/coron/median_replace_img.py
```diff
@@ -89,9 +90,16 @@ def median_replace_img(img_model, box_size, bad_bitvalue):
...
+        # set the others to 0 with no logging message
+        # bad_locations, zero_locations = filter_bad_locations(img_model, bad_locations)
```
hbushouse, line 95: Should remove this commented line of code, since it's not being used.

hbushouse (commented):
Looking good. Just a couple remaining minor questions. Can you also please add an entry in the change log (file CHANGES.rst in the top repo directory) to document this change. It can go under an entry for `align_refs`. Note that the github number that gets included in the change log entry (for tracking purposes) is the number of the PR (i.e. 6722).

I also wonder if anything needs to be added to the docs for the `align_refs` step, to mention this new behavior. That would go in /docs/jwst/align_refs/description.rst.

jwst/coron/median_replace_img.py
aggle: There used to be a check for MIRI in the old filtering function, but I forgot to put it in the new one when I rewrote it to not use SIAF. I will check with the NIRCam people to see if they are OK with this behavior; otherwise I'll put it back. Thanks for checking.

aggle: > I also wonder if anything needs to be added to the docs for the align_refs step, to mention this new behavior. That would go in /docs/jwst/align_refs/description.rst.

I don't think so, since the NON_SCIENCE pixels actually get masked out in the alignment algorithm by the PSF_MASK reference file.

hbushouse (approved):
LGTM

hbushouse:
@bsunnquist do you have any comments or suggestions before we merge this?

bsunnquist (approved):
Yep looks good - looks like this ultimately results in no changes to NIRCam, which is fine since we don't have these large NON_SCIENCE pixel areas.

aggle:
Thanks for all your suggestions @bsunnquist and thanks @hbushouse for holding my hand through all the changes. I saw that this is part of build 8, which is due in May. Does that mean it's merged with the "stable" branch, or is that when it will be merged with "master"? I ask because I work with some people who are doing lots of MIRI coronagraphy simulations to prepare for ERS data and I'm sure they'll want to know. I guess I can always refer them to the master branch of my fork of the pipeline.

hbushouse:
> I saw that this is part of build 8, which is due in May. Does that mean it's merged with the "stable" branch, or is that when it will be merged with "master"? I ask because I work with some people who are doing lots of MIRI coronagraphy simulations to prepare for ERS data and I'm sure they'll want to know. I guess I can always refer them to the master branch of my fork of the pipeline.

As soon as we merge this PR the update/fix will be available in the master branch for anyone to use in an off-line environment (by installing the jwst pkg from github). It just won't be in an official delivery to Ops until B8.0 goes out.

aggle:
Ah I understand. Thanks very much!

PR #6726, opened 2022-02-11
JP-2387: Update skymatch step docs for clarity

### LINKED ISSUE #6581: Skymatch documentation for "global+match" not clear
_Issue [JP-2387](https://jira.stsci.edu/browse/JP-2387) was created on JIRA by [Mattia Libralato](https://jira.stsci.edu/secure/ViewProfile.jspa?name=libra):_

The documentation of the skymatch step regarding the option "global+match" is not clear. Regardless whether "subtract" is True or False, the output i2d image of the resample step has the background subtracted.

The current documentation says:

 
```java
’global+match’ : first use ‘match’ method to equalize sky values between images and then find a minimum “global” sky value in all input images.
Note
This is the recommended setting for images containing diffuse sources (e.g., galaxies, nebulae) covering significant parts of the image.
```
The "match" option already matches the background to that of the minimum (or maximum depending on the "match_down" option) background among all images.

 

Furthermore, the readthedocs documentation also writes:
```java
The 'global+match' algorithm combines 'match' and 'global' methods in order to overcome the limitation of the 'match' method described in the note above: it uses 'global' algorithm to find a baseline sky value common to all input images and the 'match' algorithm to “equalize” sky values in the mosaic. Thus, the sky value of the “reference” image will be equal to the baseline sky value (instead of 0 in 'match' algorithm alone).
```
Because of how the pipeline is designed (see comments in [JP-2024](https://jira.stsci.edu/browse/JP-2024), when using "global+match" the variable that stores the background of each image initially saves the bkg difference (with "match") and then the code adds to it the minimum background among all images (with "global"). Regardless if "subtract" is True or False, the final i2d resampled image has the background subtracted. If you use only "match", the background in the final image is equal to the minimum (or maximum) bkg value among all images. Thus, I would say that the last sentence is not correct.

 

I think the documentation should better describe what this option allows the user to do, and what are the differences with the "match" option.

hbushouse:
Closes #6581 
Resolves [JP-2387](https://jira.stsci.edu/browse/JP-2387)

**Description**

This PR updates the docs for the skymatch step, to clarify the differences between the various sky computation methods and add some example results (to better illustrate the methods). Just some general text cleanup as well.

Checklist
- [ ] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


docs/jwst/skymatch/description.rst
```diff
@@ -0,0 +1,239 @@
...
+to remove bad pixels from being used in computations of the sky statistics.
+
+In addition to the classic "local" method,
+two other methods have been introduced: "global" and
+"match", as well as a combination of the two -- "global+match".
+
+- The "global" method computes a single sky value across *all*
+  input images. The resulting *single sky value* is then considered to be
+  the background in *all input images*. The global value is always based on
+  the image with the lowest sky signal level.
```
tapastro, line 47: So the algorithm of global to select the image is "local" at its base, i.e. the global method uses the lowest "local" value as the global value? On first read, it wasn't clear exactly what the " is always based on" means  - perhaps just 

> The global value uses the local method on each image and selects the lowest value.

docs/jwst/skymatch/description.rst
```diff
@@ -0,0 +1,239 @@
...
+in the images that correspond to common (overlap) regions on the celestial
+sphere for both images being compared (comparison is pair-wise).
+This method is often used when there are no large
+empty sky regions in the images, such as when a diffuse extended source occupies
+most of the view. This method cannot measure true background, but
+rather additive corrections that need to be applied to the input images so that
+the total signal from the same part of the sky is equal in all images.
+
+The computed sky value for each image is stored in the BKGLEVEL keyword of the
+output data products. The BKGSUB keyword (boolean) is also populated to indicate
```
tapastro, line 119: Are the values in the table above the ones that are stored in the BKGLEVEL keyword? If so, this behavior makes me a bit uneasy - the step output seems very modal in whether it's giving you the actual estimate of the background (for local or global), a value to match other exposures (match_down or match_up), or an estimate of the background like the first, but instead of the background value, a number to be removed from the image to reach a true background level. Is that correct? 

If something else is stored in BKGLEVEL, it might be useful to clarify.

tapastro (commented):
Just a couple of comments.

docs/jwst/skymatch/description.rst
```diff
@@ -0,0 +1,239 @@
...
+the ``skymatch`` step in a matching mode, it compares *total* signal levels
+in *the overlap regions* (instead of doing this comparison on a per-pixel basis)
+of a set of input images and computes the signal offsets for each image that will
+minimize -- in a least squares sense -- the residuals across the entire set.
+This comparison is performed directly on the input images without resampling
+them onto a common grid. The overlap regions are computed directly on the sky
+(celestial sphere) for each pair of input images. By default the sky value
+computed for each image is recorded, but not actually subtracted from the
+images. Also note that the meaning of "sky background" depends on the chosen
+sky computation method.
```
mlibralato, line 20: True, but I would also add that it is not only sky. There are other contributions, for example thermal self-emission.

docs/jwst/skymatch/description.rst
mlibralato: I agree, it is not very clear. Does this mean that, if the "subtract" flag is set to True, the background that is subtracted to all images is exactly the lowest one and the differences between images are not taken in account?

docs/jwst/skymatch/description.rst
```diff
@@ -0,0 +1,239 @@
...
+
+  Note that if the argument "match_down=True", matching will be done to the image
+  with the lowest sky value, and if "match_down=False" it will be done to the
+  image with the highest value.
+
+- The "global+match" algorithm combines the "global" and "match" methods.
+  It uses the "global" algorithm to find a baseline sky value common to all
+  input images and the "match" algorithm to equalize sky values among images.
+  The direction of matching (to the lowest or highest) is again controlled by
+  the "match_down" argument.
```
mlibralato, line 68: What is the difference between "global+match" and "local"? The example in the table below (very nice) shows that the result is the same.

docs/jwst/skymatch/description.rst
mlibralato: I believe that the BKGLEVEL is the saved in the table above. Maybe adding new keyword saving the method used could help to remember what BKGLEVEL refers to.

docs/jwst/skymatch/description.rst
```diff
@@ -0,0 +1,239 @@
...
+  too strong, clipping may converge to different sky values in
+  different images even when factoring in the true difference
+  in the sky background between the two images.
+
+- In general images can have different true background values
+  (we could measure it if images were not affected by large diffuse
+  sources). However, arguments such as ``lower`` and ``upper`` will
+  apply to all images regardless of the intrinsic differences
+  in sky levels.
+
```
mlibralato, line 236: I would add what happens if you have a mosaic with images not all overlapping. For example, if you have two groups of images not overlapping with each other, what does 'skymatch' do? Does it work as if the two groups are two distinct mosaic? I think this is particularly important when 'match' is used

mlibralato (commented):
I added a few comments.

docs/jwst/skymatch/description.rst
hbushouse: Yes, "global" by itself essentially uses the "local" method to find the background level in each image independently and then use the lowest result as the background for the entire set. That single (lowest) value is what will be used to populate the BKGLEVEL keyword in all images and what would be used to do the subtraction if that parameter were turned on. That would result in one image having (near) zero bkg, while the remaining ones would have some positive residual in them.

docs/jwst/skymatch/description.rst
hbushouse: Yes, the values listed in the table are the ones that would end up being stored in the BKGLEVEL keyword for each image. Of course for a mode like "match", that value is not the true total background in each image, it's only the value necessary to get them all to match one another. They could all still have some (nearly) common residual background signal in them. So reporting, in another keyword, the bkg method that was used could be very useful, so that folks understand exactly what the values in BKGLEVEL mean. Perhaps a new keyword like "BKGMETH". It would not be possible to store the method used in the comment portion of the BKGLEVEL keyword, because the comments come from a single, static definition of the keyword.

docs/jwst/skymatch/description.rst
hbushouse: Good question. I'll check into that.

docs/jwst/skymatch/description.rst
mlibralato: I agree. In general the name of the method is a bit misleading but if the documentation is clear, it is fine to keep the name as it is.

docs/jwst/skymatch/description.rst
mlibralato: OK, I understand how comments work. I agree that adding a new keyword would help to keep track of things.

mlibralato (commented):
@hbushouse A few replies to your comments

docs/jwst/skymatch/description.rst
hbushouse: I will add the BKGMETH keyword via another PR. I've included mention of it in the latest commits to the docs here.

docs/jwst/skymatch/description.rst
hbushouse: The latest commit mentions the fact that the term "sky" is being used generically in the docs and can include other contributions that just true sky.

docs/jwst/skymatch/description.rst
hbushouse: I've done the experiment of having gaps on the sky and have included it in the latest examples. Bottom line: gaps are not good for the match method.

hbushouse:
@tapastro and @mlibralato I've added some more new info to the docs and have done a little more rearranging as well. You can see the fully rendered result at https://jwst-pipeline--6726.org.readthedocs.build/en/6726/jwst/skymatch/index.html if that makes it easier to review.  Hopefully this now addresses all previous questions and comments.

hbushouse:
New keyword "BKGMETH" is implemented via #6736

docs/jwst/skymatch/description.rst
```diff
@@ -0,0 +1,247 @@
...
++-------+-------+--------+-------+--------------+
+| 110   |  110  |   90   |    10 |    96.25     |
++-------+-------+--------+-------+--------------+
+|  95   |   95  |   90   |  8.75 |     95       |
++-------+-------+--------+-------+--------------+
+|  90   |   90  |   90   |  3.75 |     90       |
++-------+-------+--------+-------+--------------+
+| 100   |  100  |   90   | 13.75 |    100       |
++-------+-------+--------+-------+--------------+
+
```
mlibralato, line 175: Thanks for this example. I am a bit puzzled by the results of the three images at the bottom. Why is the result for "match" showing decimal values as well? Also, why is the result of "global+match" showing the same results for local for the last three images (as it should) but strange values for the top  four images?

mlibralato (approved):
The new changes are fine for me.

PR #6727, opened 2022-02-11
JP-2071: Adding a One Good Group Ramp Suppression Switch

### LINKED ISSUE #6013: Add ramp_fit parameter to turn on/off one-group slope calculations
_Issue [JP-2071](https://jira.stsci.edu/browse/JP-2071) was created on JIRA by [Michael Regan](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mregan):_

Currently, the ramp fitting step will fit ramps with only a single useable group using the difference between the first group and the superbias. This assumes that the superbias reference file used darks taken at the same temperature as the current observations. Because there will be secular drifts in temperature over the life of the observatory (especially during commissioning), there can be a mismatch in temperature which will lead to the wrong slope.
It would be best to have a parameter value that can turn this off and on with the default being off until this is tested with on-orbit observations.

kmacdonald-stsci:
Closes #6013 
Resolves [JP-2071](https://jira.stsci.edu/browse/JP-2071)

**Description**

This PR adds switch to suppress computations for ramps that have only one good group in them.  This is done by find those ramps, then setting the DQ flag for the one good group in that ramp to `DO_NOT_USE`.  This makes that ramp a zero group ramp and all computations will be done for that ramp the same as if the ramp had no good groups.

The `ramp_fit` function in STCAL added `suppress_one_group` to it's signature due to this PR.  This causes many ramp fitting tests to fail, since the STCAL PR has not yet been merged.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/ramp_fitting/ramp_fit_step.py
+++ b/jwst/ramp_fitting/ramp_fit_step.py
@@ -182,7 +182,7 @@ class RampFitStep(Step):
         int_name = string(default='')
         save_opt = boolean(default=False) # Save optional output
         opt_name = string(default='')
-        suppress_one_group = boolean(default=False) # Suppress ramps with only one good group
+        suppress_one_group = boolean(default=True) # Suppress ramps with only one good group
         maximum_cores = option('none', 'quarter', 'half', 'all', default='none') # max number of processes to create
     """
 
--- a/jwst/ramp_fitting/ramp_fit_step.py
+++ b/jwst/ramp_fitting/ramp_fit_step.py
@@ -182,6 +182,7 @@ class RampFitStep(Step):
         int_name = string(default='')
         save_opt = boolean(default=False) # Save optional output
         opt_name = string(default='')
+        suppress_one_group = boolean(default=False) # Suppress ramps with only one good group
         maximum_cores = option('none', 'quarter', 'half', 'all', default='none') # max number of processes to create
     """
 
@@ -240,7 +241,8 @@ class RampFitStep(Step):
             image_info, integ_info, opt_info, gls_opt_model = ramp_fit.ramp_fit(
                 input_model, buffsize,
                 self.save_opt, readnoise_2d, gain_2d, self.algorithm,
-                self.weighting, max_cores, dqflags.pixel)
+                self.weighting, max_cores, dqflags.pixel,
+                suppress_one_group=self.suppress_one_group)
 
         # Save the OLS optional fit product, if it exists
         if opt_info is not None:
--- a/jwst/ramp_fitting/tests/test_ramp_fit_step.py
+++ b/jwst/ramp_fitting/tests/test_ramp_fit_step.py
@@ -254,9 +254,9 @@ def test_int_times2(generate_miri_reffiles, setup_inputs):
 
 def one_group_suppressed(suppress, setup_inputs):
     """
-    Tests three pixel ramps.
+    Creates three pixel ramps.
     The first ramp has no good groups.
-    The second ramp has one good groups.
+    The second ramp has one good group.
     The third ramp has all good groups.
 
     Sets up the models to be used by the tests for the one
@@ -305,7 +305,7 @@ def test_one_group_not_suppressed(setup_inputs):
     """
     Tests three pixel ramps.
     The first ramp has no good groups.
-    The second ramp has one good groups.
+    The second ramp has one good group.
     The third ramp has all good groups.
 
     Verify that when the suppress switch is turned off the second ramp is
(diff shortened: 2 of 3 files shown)


jwst/ramp_fitting/tests/test_ramp_fit_step.py
```diff
@@ -246,3 +250,117 @@ def test_int_times2(generate_miri_reffiles, setup_inputs):
     assert cube_model is not None
 
     assert(len(cube_model.int_times) == nints)
+
+
+def one_group_suppressed(suppress, setup_inputs):
+    """
+    Tests three pixel ramps.
+    The first ramp has no good groups.
+    The second ramp has one good groups.
```
dmggh, line 259: 'groups' -> 'group'

jwst/ramp_fitting/tests/test_ramp_fit_step.py
```diff
@@ -246,3 +250,117 @@ def test_int_times2(generate_miri_reffiles, setup_inputs):
     assert cube_model is not None
 
     assert(len(cube_model.int_times) == nints)
+
+
+def one_group_suppressed(suppress, setup_inputs):
+    """
+    Tests three pixel ramps.
```
dmggh, line 257: Should that be "Creates three pixel ramps" ?

jwst/ramp_fitting/tests/test_ramp_fit_step.py
```diff
@@ -246,3 +250,117 @@ def test_int_times2(generate_miri_reffiles, setup_inputs):
...
+        maximum_cores="none")
+
+    return slopes, cube_model, dims
+
+
+def test_one_group_not_suppressed(setup_inputs):
+    """
+    Tests three pixel ramps.
+    The first ramp has no good groups.
+    The second ramp has one good groups.
```
dmggh, line 308: 'groups' -> 'group'

jwst/ramp_fitting/tests/test_ramp_fit_step.py
```diff
@@ -246,3 +250,117 @@ def test_int_times2(generate_miri_reffiles, setup_inputs):
...
+    cdata_check[0, 0, :] = np.array([np.nan, 0., 1.0000001])
+    cdata_check[1, 0, :] = np.array([1.0000001, 1.0000001, 1.0000001])
+    np.testing.assert_allclose(cube_model.data, cdata_check, tol)
+
+
+def test_one_group_suppressed(setup_inputs):
+    """
+    Tests three pixel ramps.
+    The first ramp has no good groups.
+    The second ramp has one good groups.
```
dmggh, line 345: 'groups' -> 'group'

dmggh (approved):
Looks good to me. Just a few trivial typos in comments.

jwst/ramp_fitting/tests/test_ramp_fit_step.py
kmacdonald-stsci: Fixed typo.

jwst/ramp_fitting/tests/test_ramp_fit_step.py
kmacdonald-stsci: Changed wording.

jwst/ramp_fitting/tests/test_ramp_fit_step.py
kmacdonald-stsci: Fixed typo.

jwst/ramp_fitting/tests/test_ramp_fit_step.py
kmacdonald-stsci: Fixed typo.

jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -182,6 +182,7 @@ class RampFitStep(Step):
         int_name = string(default='')
         save_opt = boolean(default=False) # Save optional output
         opt_name = string(default='')
+        suppress_one_group = boolean(default=False) # Suppress ramps with only one good group
```
hbushouse, line 185: The requested behavior from the Cal WG was that the slope calculation from only 1 group should NOT be used by default, so if this switch turns off the 1-group calculation, then the default value of it should be `True`, at least for now. Given the fact that `suppress_one_group=False` is kind of a case of double negatives, I wonder if it would be better to change the name and sense of the switch so that it's in the positive sense, e.g. something like `use_one_group` or `one_group_slope`. Setting that to `True` would enable the current calculation of slopes from only 1 group, while setting it to `False` would disable it. In this case the default would be `False`.

kmacdonald-stsci: Changed the default to be True, instead of False.

kmacdonald-stsci:
The PR does not correctly address the feature request.

PR #6732, opened 2022-02-15
Fix IFU x1d error propagation

drlaw1558:
Closes #
Resolves https://jira.stsci.edu/browse/JP-2522

**Description**

This PR fixes propagation of errors for IFU data cubes through the Extract1D stage.   Extract1D relies on having individual poisson/readnoise/flat components of the variance; with this PR it uses this framework to pass through a generalized error term even in the absence of breaking out the components individually in the data cubes.

Checklist
- [ ] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


drlaw1558:
Updated pull request after approval by Cal WG.

tapastro:
I don't see anything wrong with the code additions, but with the limited IFU data I have on hand, I'm not convinced it's working properly. Have you run some data through this and found the results acceptable? If not, there may be a bug in the existing extract_1d code, or perhaps I just have questionable err arrays in my (the regression test) data.

drlaw1558:
It looks reasonable to me in the example files that I tested (from a recently-updated https://github.com/STScI-MIRI/MRS-ExampleNB/blob/main/Notebook3/MRS_Notebook3.ipynb)  Pipeline 1d error visually similar to a rough estimate that I made directly from the cubes themselves using numpy.  There were a lot of unrelated error bugs in the past, so it's possible that the regtest data are out of date.

tapastro:
I downloaded your data and the error array comes out looking fine - pending the merge conflict, looks good to me!

drlaw1558:
Ah, a collision on the CHANGES.rst file update; should I merge main into my branch?

nden:
The one failure is due to a bad CRDs connection. Merging.

PR #6737, opened 2022-02-18
JP-2071: Update ramp_fit to toggle one-group slope calculations

### LINKED ISSUE #6013: Add ramp_fit parameter to turn on/off one-group slope calculations

kmacdonald-stsci:
Closes #6013 
Resolves [JP-2071](https://jira.stsci.edu/browse/JP-2071)

**Description**

A `suppress_one_group` boolean variable has been added as a parameter to the ramp fitting step class.

When true (the default) saturated ramps with only the 0th group being good will be treated as completely saturated ramps. The calculations will be done as if all calculations are saturated, but the DQ flags will be set as if the first group is good, i.e., the pixel DQ will not have DO_NOT_USE set.

When false saturated ramps with only the 0th group being good will be treated as the normal special case of having only one good group.

Many unit tests are expected to fail until the corresponding PR in STCAL has been merged.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


CHANGES.rst
```diff
@@ -48,6 +48,9 @@ ramp_fitting
   the various flavors of variance and ERR stored in the output
   products [#6715]
 
+- Adding feature to turn off calculations of ramps with good 0th group,
+  but all other groups are saturated. [#]
```
hbushouse, line 52: add 6737 to PR number

jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -240,7 +241,11 @@ def process(self, input):
             image_info, integ_info, opt_info, gls_opt_model = ramp_fit.ramp_fit(
                 input_model, buffsize,
                 self.save_opt, readnoise_2d, gain_2d, self.algorithm,
-                self.weighting, max_cores, dqflags.pixel)
+                self.weighting, max_cores, dqflags.pixel, suppress_one_group=self.suppress_one_group)
+
+            print("-" * 80)
```
hbushouse, line 246: don't want to leave these in the final product

hbushouse (commented):
Looks OK. Just needs a couple minor updates before finalizing.

jwst/ramp_fitting/tests/test_ramp_fit_step.py
```diff
@@ -246,3 +247,232 @@ def test_int_times2(generate_miri_reffiles, setup_inputs):
...
+        override_gain=gmodel,
+        override_readnoise=rnModel,
+        suppress_one_group=suppress,
+        maximum_cores="none")
+
+    return slopes, cube_model, dims
+
+
+def test_one_group_not_suppressed_one_integration(setup_inputs):
+    slopes, cube, dims = one_group_suppressed(1, False, setup_inputs)
```
dmggh, line 316: Include short description of test purpose (also for later tests)

dmggh (approved):
Looks good to me.  Just a couple of minor docs comments.

CHANGES.rst
kmacdonald-stsci: I corrected this.

jwst/ramp_fitting/tests/test_ramp_fit_step.py
kmacdonald-stsci: I included short descriptions for the tests.

CHANGES.rst
```diff
@@ -46,7 +46,10 @@ ramp_fitting
 
 - Updated step docs to clarify exactly what calculations are used for
   the various flavors of variance and ERR stored in the output
-  products [#6715]
+  products [#6737]
```
hbushouse, line 49: The 6737 reference needs to go down into line 52 (the entry for this PR). The reference to 6715 needs to stay as is (that was my PR for doc updates).

jwst/ramp_fitting/ramp_fit_step.py
kmacdonald-stsci: I removed these.

hbushouse (approved):
LGTM.  I guess we have to merge the stcal PR first and then this one?

hbushouse:
It occurs to me that the docs for the ramp_fit step will also need to be updated to list this new argument and what it does.

kmacdonald-stsci:
I updated the `arguments.rst` documentation for ramp fitting.

hbushouse:
Failure in CI for oldest dependencies is due to the use of stcal 0.6.0, which does not support the new "suppress_one_group" step param. It is supported in stcal 0.6.1, which is used in the other passing CI tests.

PR #6739, opened 2022-02-21
JP-2284 Residual Fringe correction using astropy for fitting + other improvements needed to support change

### LINKED ISSUE #6350: Refactor initial background spline fit to use astropy
_Issue [JP-2284](https://jira.stsci.edu/browse/JP-2284) was created on JIRA by [William Jamieson](https://jira.stsci.edu/secure/ViewProfile.jspa?name=wjamieson):_

Currently, the residual fringe correction step relies on the external dependency [BayesicFitting](https://github.com/dokester/BayesicFitting). This library is a mathematical modeling and data fitting library intended for astronomy applications based on the Herschel Common Science System. This library is largely incompatible with astropy and its modeling package, which is the common mathematical modeling and fitting package used throughout the rest of the JWST pipeline. Moreover, BayesicFitting is not as well supported or maintained as astropy. Thus it is undesirable to include another external dependency on a less well maintained library, when much of that functionality can be derived directly from astropy.

The first step of the residual fringe correction involves fitting a spline model to the data in order to fit the background of the data. This step relies entirely on models and fitters included in the BayesicFitting package. Thus this background step must be refactored so that it only relies on astropy derived or compatible models and fitters instead.

jemorrison:
This is a replacement for JP-2284 (#6475)
Closes #6350 
Resolves [JP-2284](https://jira.stsci.edu/browse/JP-2284)

**Description**

This PR includes the updates in JP-2284 plus additional changes provided by the MIRI instrument team (Patrick Kavanagh). Updates to the residual_fringe code were needed to support changing to astropy fitting of the background for columns with less than ideal data. 
The description in JP-2284 is "This PR switches the residual fringe background fit step from using BayesicFitting based splines to the new astropy based splines. This is part of the effort to remove the dependency of jwst on BayesicFitting. Note that this will change the results of the residual fringe correction slightly due to numerical differences in how astropy (scipy based) splines are computed/fit vs how BayesicFitting splines are computed/fit.
Note that the bugfix https://github.com/astropy/astropy/pull/12523 is required for this code to function properly."

The astropy updates in 12523 have already been included into astropy.

This ticket is closes the step 1 in the parent ticket JP-2344. 

Closes #6475 and closes #6761

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


jemorrison:
@WilliamJamieson  Does this PR look ok ?

jwst/residual_fringe/fitter.py
```diff
@@ -0,0 +1,92 @@
...
+
+class ChiSqOutlierRejectionFitter:
+    """Chi squared statistic for outlier rejection"""
+
+    def __init__(self, fitter, domain=None, tolerance=0.0001):
+        self.fitter = fitter
+        self.tolerance = tolerance
+
+        if domain is None:
+            self.domain = 10
```
WilliamJamieson, line 14: Is `self.domain` ever set to anything other than 10? if not maybe we should drop it.

jwst/residual_fringe/residual_fringe.py
```diff
@@ -134,7 +132,7 @@ def do_correction(self):
 
         self.background_fit = np.zeros(self.input_model.data.shape)
         self.knot_locations = np.empty(self.input_model.data.shape)
```
WilliamJamieson, line 134:
```suggestion
        self.knot_locations = np.full_like(self.input_model.data, np.NaN)
```
This is one way to create an array of that shape filled with `np.NaN` values.

WilliamJamieson (changes requested):
One additional general question: are there any unit tests planned for testing these changes?

jwst/residual_fringe/fitter.py
jemorrison: Looking at the old code it seems domain is always set to 10. Should we just define it as self.domain  = 10 in the initialization of the ChiSqOutlierRejectFitter and not even check if it is None

jemorrison: There are unit tests planned. Do you have any suggestions for some

jemorrison:
@WilliamJamieson I was thinking about how to test the background step as a unit test. I was thinking of a test that is part unit and part regression. What if I pull out a good column of data and pull out the background fit and background subtracted col.  I think you developed the code using a "good column". The unit test could read in the 1 column of data, call the background fitting and then compare the results with the stored "regression" data in the unit test. 
I don't think it would take long since we are just testing a 1 X 1024 column of data.  Sound ok - or do you have a better idea ?

WilliamJamieson:
> @WilliamJamieson I was thinking about how to test the background step as a unit test. I was thinking of a test that is part unit and part regression. What if I pull out a good column of data and pull out the background fit and background subtracted col. I think you developed the code using a "good column". The unit test could read in the 1 column of data, call the background fitting and then compare the results with the stored "regression" data in the unit test. I don't think it would take long since we are just testing a 1 X 1024 column of data. Sound ok - or do you have a better idea ?

This is the general approach to what I was thinking. I might also add that we could do essentially the same thing with a "bad column" in a separate test(s).

jemorrison:
A bad column is a good addition, since they can be tricky.

jemorrison:
@WilliamJamieson  Take a look at the unit test I added for the backgrounds.
I wrote out the input data that goes to the fit_1d_complex and  I wrote out the background fit. 
I think pulled out 2 columns of data - one right in the center of slice and one on the edge that had caused problems. 
I then used these two columns in a unit test to test that the saved background fit and the computed background fit called in the unit test are the same. 
What do you think ? Good enough ?

WilliamJamieson:
@jemorrison can you resolve the merge conflict? That seems to be preventing the tests from being run.

jwst/residual_fringe/fitter.py
jemorrison: I added a unit test for the background fitting

jemorrison:
@WilliamJamieson @hbushouse do we need to fix the CI test - Oldest dependency versions ? or can we just merge

WilliamJamieson:
> @WilliamJamieson @hbushouse do we need to fix the CI test - Oldest dependency versions ? or can we just merge

See #6755 for discussion.

hbushouse:
> > @WilliamJamieson @hbushouse do we need to fix the CI test - Oldest dependency versions ? or can we just merge
> 
> See #6755 for discussion.

If you're confident that the "KeyError: 'slice'" error in that CI run is due to the lack of support for numpy 1.18, then I'd say we can ignore it and merge.

WilliamJamieson:
> > > @WilliamJamieson @hbushouse do we need to fix the CI test - Oldest dependency versions ? or can we just merge
> > 
> > 
> > See #6755 for discussion.
> 
> If you're confident that the "KeyError: 'slice'" error in that CI run is due to the lack of support for numpy 1.18, then I'd say we can ignore it and merge.

@hbushouse It turns out it was an astropy bug which was fixed in the 5.0.1 release. I have made that change here and it looks like everything is passing now.

jemorrison:
@hbushouse  can we merge this PR ? If so I will close some of the other ones that are duplicates

hbushouse:
> @hbushouse can we merge this PR ? If so I will close some of the other ones that are duplicates

But I thought that this PR was going to get closed, because it's been replaced by #6761 . So which PR is the active one with the correct code and configuration?

WilliamJamieson:
> But I thought that this PR was going to get closed, because it's been replaced by #6761 . So which PR is the active one with the correct code and configuration?

@hbushouse sorry for the confusion. I think local git issues caused the creation of #6761, while I was still working on this one. All of the changes that fix the issues with the oldest dependencies have only been resolved in this PR.

PR #6749, opened 2022-03-02
JP-2536 Reset model updates 

jemorrison:
Resolves [JP-2536](https://jira.stsci.edu/browse/JP-2536)

**Description**

This PR updates the reset model to be in line with what CRDS is expecting. The reset model is an integration and group dependent correction. CRDS is expected the NINTS and NGROUPS keywords as part of the reset model. This was discovered as J Morrison was creating new reset reference files to be delivered to CRDS.
There are also updates to the reset step to use the NINTS and NGROUP instead of the shape of the data to determine how many integration and groups there are to be corrected. 

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,6 +41,8 @@ datamodels
 - Added the new keyword "BKGMETH" for use in the ``skymatch`` step.
   [#6736]
 
+- Updated reset model to include NINTS, NGROUPS keywords and the subarray.schema [#6749]
+
 extract_1d
 ----------
 
@@ -60,6 +62,12 @@ ramp_fitting
   the various flavors of variance and ERR stored in the output
   products [#6715]
 
+reset
+-----
+
+- Reading nints and ngroups from  model.meta for reset reference file and data instead of using the
+  shape of the data to define these values [#6749]
+  
 set_telescope_pointing
 ----------------------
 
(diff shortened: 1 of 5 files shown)


hbushouse (commented):
Code updates look good to me. Is it safe to assume that all the CI failures are due to either reset ref files or science files not having the nints or ngroups keywords in them?  The error messages imply that these values are None when trying to be accessed by some of the unit tests. Need to get these fixed before merging the PR.

hbushouse:
Oh, and now that you're changing both code and model schemas, it would be good to have a JP ticket to track this, at least for historical purposes, even though you're creating the ticket after the fact.

jemorrison:
I think there might be something wrong with my branch. I did this  pip install -e ".[tests]" (in the directory with  setup.cfg) 
But in the reset/test directory when I type 'pytest' or 'pytest test_reset_sub.py'
I get an error:
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --open-files
  inifile: /Users/morrison/git/jwst/setup.cfg
  rootdir: /Users/morrison/git/jwst


Am I doing this wrong ?

hbushouse (approved):

PR #6751, opened 2022-03-03
JP-2534 Set CRVAL* from GS_* for guider exposures

stscieisenhamer:
Resolves [JP-2534](https://jira.stsci.edu/browse/JP-2534)

**Description**

This PR addresses implements the setting  of CRVAL1/CRVAL2 for guider products in `set_telescope_pointing`.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [ ] Label(s)

### DIFF
--- a/jwst/lib/set_telescope_pointing.py
+++ b/jwst/lib/set_telescope_pointing.py
@@ -634,8 +634,10 @@ def update_wcs_from_fgs_guiding(model, default_roll_ref=0.0, default_vparity=1, 
 
     For Fine Guidance guiding observations, nearly everything
     in the `wcsinfo` meta information is already populated,
-    except for the PC matrix. This function updates the PC
+    except for the PC matrix and CRVAL*. This function updates the PC
     matrix based on the rest of the `wcsinfo`.
+
+    CRVAL* values are taken from GS_RA/GS_DEC.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


nden (approved):

jwst/lib/set_telescope_pointing.py
```diff
@@ -684,6 +686,10 @@ def update_wcs_from_fgs_guiding(model, default_roll_ref=0.0, default_vparity=1,
         model.meta.wcsinfo.pc2_2
     ) = calc_rotation_matrix(roll_ref, np.deg2rad(v3i_yang), vparity=vparity)
 
+    # Set CRVAL as the guide star coordinates.
+    model.meta.wcsinfo.crval1 = model.meta.guidestar.gs_ra
+    model.meta.wcsinfo.crval2 = model.meta.guidestar.gs_dec
+
```
hbushouse, line 692: So apparently the storing of these same values that happens in the block at https://github.com/spacetelescope/jwst/blob/a278351fc50a9336d07b7abc9b02f1eefe3f4ea9/jwst/lib/set_telescope_pointing.py#L2458 isn't sufficient for getting crvaln populated already?

stscieisenhamer: No. The guiding exposures take a completely different approach. I haven't tried to find the original documentation so I am not sure why; way to long ago for memory. I believe it is because guiding is "self-referential": It is reporting the information that is needed, so it already knows.  The TRACK algorithm, for regular science, is then based on this information to transform to the science detectors.  COARSE is the only one that needs the original observatory orientation, since guiding is not in use.

stscieisenhamer:
the failing job is the usual crds sync issue. Merging anyways.

PR #6752, opened 2022-03-03
Drop references to transform-1.2.0 from datamodel schemas

eslavich:
**Description**

This PR addresses an issue with recent updates to the transform schemas that use jsonschema features that stdatamodels doesn't know how to handle.  Since the transform-1.2.0 schema is mainly used as documentation that an astropy model is expected, we can replace it with actual documentation in the description field which is more appropriate to this situation anyway.

Checklist
- [ ] ~Tests~
- [ ] ~Documentation~
- [x] Change log
- [ ] Milestone
- [x] Label(s)


WilliamJamieson (approved):
LGTM

PR #6760, opened 2022-03-08
JP-2530 Create new datamodel GuiderStreamModel and enable WCS determination for it

stscieisenhamer:
Resolves [JP-2530](https://jira.stsci.edu/browse/JP-2530)
Resolves [JP-2532](https://jira.stsci.edu/browse/JP-2532)

**Description**

This PR addresses allows WCS determination for Guider "stream" format data. In order to make this possible, a new model, GuiderStreamModel, is also created.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [ ] Label(s)


jwst/datamodels/schemas/guider_stream.schema.yaml
```diff
@@ -0,0 +1,61 @@
...
+- $ref: bunit.schema
+- $ref: wcsinfo.schema
+- type: object
+  properties:
+    data:
+      title: The science data
+      fits_hdu: SCI
+      default: 0
+      ndim: 3
+      datatype: uint16
```
tapastro, line 17: One note: the ndim listed here is 3, and all stream files provided in the data directory have NAXIS = 3, but the TR that you provided for documentation would appear to indicate that the gs-acq streams would have NAXIS = 4. That might just be out of date. The ticket that spawned this issue (JWSTDMS-641) related to a gs-acq1_..._stream.fits file, for what it's worth.

stscieisenhamer: Good catch. The authors have been asked.

stscieisenhamer:
This PR is being withdrawn. STREAM format should not get WCS. Also, there is question as to whether the current data format for STREAM is correct.

PR #6762, opened 2022-03-10
Define MIRI IFU point source correction data model

### LINKED ISSUE #6049: IFU detector-based spectral extraction for point sources
_Issue [JP-2088](https://jira.stsci.edu/browse/JP-2088) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

There are multiple percent-level effects that will influence extracted spectra of point sources that cannot be corrected for with the currently-implemented method of building data cubes and then extracting spectra from those cubes.  Indeed, building data cubes for known point sources isn't the best way of working with the data as it introduces unnecessary covariance.

We therefore propose a mechanism for extraction of point source spectra directly from the calibrated 2d detector data in a manner that can easily account for across-slice wavelength offsets, across-slice transmission losses, and the MRS spectral leak.  Each of these effects is impossible to generically correct for in data cube space, but simple to address when extracting a point source directly from the 2d data.

This mechanism may at some point make sense to incorporate into the Optimal pipeline architecture; in the meantime we're looking for Cal WG feedback on the methodology and the potential for incorporating it into offline analysis tools.

Discussion scheduled for July 6 Cal WG, to be presented by Yannis Argyriou.

drlaw1558:
Closes #6049 
Resolves [JP-2088](https://jira.stsci.edu/browse/JP-2088)

**Description**

This PR implements a data model for a new reference file type defining MIRI IFU point source correction factors for detector-based extraction. This is for future use in the Cal pipeline. Needed to support [CRDS-541](https://jira.stsci.edu/browse/CRDS-541).

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


drlaw1558:
Updated to merge latest master into the branch and add a change log entry, should now be ready for review.

CHANGES.rst
```diff
@@ -58,6 +58,9 @@ datamodels
 
 - Correcting the default ZEROFRAME allocation. [#6791]
 
+- Add the new MIRI IFU point source correction reference file data model
+  of type ifuptcor. [#6762]
```
hbushouse, line 62: Perhaps better to mention the actual data model class name here, rather than the module name, e.g. `MiriIFUPtCorrModel`.

jwst/datamodels/__init__.py
```diff
@@ -27,6 +27,7 @@
 from .ifucube import IFUCubeModel
 from .ifucubepars import NirspecIFUCubeParsModel, MiriIFUCubeParsModel
 from .ifuimage import IFUImageModel
+from .ifuptcor import MiriIFUPtCorrModel
```
hbushouse, line 30: Do we know if this will only ever be used for MIRI MRS or is there a chance it could be used for NIRSpec IFU too? If it is MIRI-specific, then the module name should probably indicate that, e.g. "mrsptcor" or "miriptcor". Unlike modules such as photom or pathloss, this will likely only ever contain a single MIRI-specific datamodel.

jwst/datamodels/schemas/miri_ifuptcor.schema.yaml
```diff
@@ -0,0 +1,82 @@
...
+      fits_hdu: ERR
+      default: 0.0
+      ndim: 2
+      datatype: float64
+    dq:
+      title: Data quality array
+      fits_hdu: DQ
+      default: 0
+      ndim: 2
+      datatype: int64
```
hbushouse, line 27: I believe most (if not all) DQ arrays up to this point are uint32. Is there a need for int64?

hbushouse (commented):
Structures look fine, just a few questions regarding naming conventions. Unfortunately we started down a bad road a while ago already with mixing references to MIRI MRS data with labels of "MRS" and "IFU", in no consistent fashion. Not sure how feasible it is at this point to try to get back to some kind of consistency.

jwst/datamodels/schemas/miri_ifuptcor.schema.yaml
drlaw1558: No, not sure why that was set (values are all zero anyway).  I've changed to int32.

jwst/datamodels/schemas/miri_mrsptcorr.schema.yaml
```diff
@@ -24,7 +24,7 @@ allOf:
       fits_hdu: DQ
       default: 0
       ndim: 2
-      datatype: int64
+      datatype: int32
```
hbushouse, line 27: actually should be *u*int32 (unsigned)

drlaw1558: Hm, what's the best way to convert an existing array from regular int to uint?  Manually add an offset value and hack the BZERO in the header?  Passing it through np.uint looks like it goes to uint64 not uint32.

drlaw1558:
Added exptype and pexptype to schema based on parallel conversation for https://github.com/spacetelescope/jwst/pull/6800

jwst/datamodels/schemas/miri_mrsptcorr.schema.yaml
hbushouse: Can you just do an `<my_array>.astype(np.uint32)`?

drlaw1558: Done

hbushouse (approved):

PR #6765, opened 2022-03-10
JP-2526: Fix kdtree memory issues on Linux systems

larrybradley:
Resolves [JP-2526](https://jira.stsci.edu/browse/JP-2526)

**Description**

This PR fixes the KDTree calculation to use only non-finite source positions to prevent memory issues on Linux systems. It also filters a RuntimeWarning for non-finite source positions.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -71,12 +71,10 @@ reset
 - Read NINTS and NGROUPS from model.meta for reset reference file and data instead of using the
   shape of the data to define these values [#6749]
 
-
 residual_fringe
 ---------------
 
 - Replaced fitting the background with an astropy fitting package [#6739]
-
 
 set_telescope_pointing
 ----------------------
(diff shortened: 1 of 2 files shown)


larrybradley:
For non-finite source positions, `nn_dist` is assigned to `np.nan` and `nn_label` is assigned to -1.  I chose -1 to keep the `nn_label` as `int` type (inserting `np.nan` requires float type).

CHANGES.rst
```diff
@@ -97,6 +95,12 @@ skymatch
 
 - Updated to populate the "BKGMETH" keyword in output files. [#6736]
 
+source_catalog
+--------------
+
+- Fixed the KDTree calculation to use only non-finite source positions
```
hbushouse, line 101: I think you mean *finite* here! We're NOT using the non-finite positions.

hbushouse (approved):
Other than the one minor change log comment, LGTM.

hbushouse:
Hmmm, one of the new lines of code seems to be causing a CI failure.

CHANGES.rst
larrybradley: Thanks! 😃

larrybradley:
The SDP failure doesn't seem related.  I'll investigate the other one.  The pipeline ran fine locally for me on the troublesome dataset.

PR #6766, opened 2022-03-10
Fix sharpness and roundness to use source centroid instead of peak

larrybradley:
**Description**

This PR is a follow-on to #6765.  While investigating that issue, I found that the sharpness and roundness parameters are using the source peak position instead of its centroid.  This PR fixes that.  Sources that have non-finite centroid positions will have `np.nan` for these parameters.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -104,11 +104,13 @@ source_catalog
 - Fixed the KDTree calculation to use only finite source positions to
   prevent memory issues on Linux systems. [#6765]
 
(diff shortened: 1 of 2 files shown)


hbushouse:
Now that #6765 has been merged, this needs a change log update to fix conflicts.

larrybradley:
rebased.

PR #6767, opened 2022-03-10
JP-2530 Further restrict default models that can be updated

stscieisenhamer:
Resolves [JP-2530](https://jira.stsci.edu/browse/JP-2530)

**Description**

PR#6735 allowed any datamodel to be modified, with warnings if an unexpected model was found. This had the unintended side effect in ground processing for failing on certain observatory products that were not intended to be used. This PR re-instates some of the restrictions, but allow the option of modifying anything.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -78,6 +78,8 @@ set_telescope_pointing
 
 - Set CRVAL* from GS_* for guider exposures. [#6751]
 
+- Further restrict default models that can be updated. [#6767]
+
 skymatch
 --------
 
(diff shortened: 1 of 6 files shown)


tapastro (approved):
LGTM - seems clear that the new default is to allow only the expected file/model types through, unless explicitly flagged to try allowing everything through?

stscieisenhamer:
@tapastro Yes, that is the intent: Default is restricted because otherwise it is "your mileage may vary".

hbushouse (approved):
LGTM

stscieisenhamer:
failing test is the usual crds/CI issue and is a non-blocker.

PR #6771, opened 2022-03-14
Residual fringe regression test

jemorrison:
This PR is part of JP-2344 - the parent ticket of the residual fringe correction 

**Description**

This PR adds a regression test for the residual fringe corrrection. Ground test data was converted to DMS format. The ground test data is a good example of data with fringes.  The input and truth files have been added to artifactory

Checklist
- [X] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -60,6 +60,11 @@ extract_1d
 - Add separate behavior for 2D vs (3D data with only one image)
   by passing appropriate integ value [#6745]
 
+regtest
+-------
+
+- Added a residual fringe correction test [#6771]
+
 pipeline
 --------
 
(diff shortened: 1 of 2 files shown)


stscieisenhamer:
In general, I mark a test slow if it takes more than a minute, two max, to run. The idea is that one can run through all the unit tests or fast regression tests quickly; ideally < 5 minutes, so any single test that consumes too much time makes the "quick test" non-useful. Once one is satisfied with the "quick test", then adding the `--slow` option can be done, as needed, or let the Jenkins run take care of the slow tests.

jemorrison:
Ok I will make this a slow - because it takes longer than 1 to 2 minutes

stscieisenhamer:
@jemorrison No, this is an unrelated issue. I'll go ahead and force the merge if good with you

PR #6772, opened 2022-03-15
JP-2540: Remove existence and usage of FILETYPE meta keyword

tapastro:
Closes #
Resolves [JP-2540](https://jira.stsci.edu/browse/JP-2540)

**Description**

This PR removes the FILETYPE keyword from pipeline products and all assignments to that keyword inside the pipeline. 

NOTE: should I also remove the keyword from the referencefile schema?

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/combine_1d/combine_1d_step.py
+++ b/jwst/combine_1d/combine_1d_step.py
@@ -24,6 +24,5 @@ class Combine1dStep(Step):
         with datamodels.open(input_file) as input_model:
             result = combine1d.combine_1d_spectra(input_model,
                                                   self.exptime_key)
-        result.meta.filetype = '1d combined spectrum'
 
         return result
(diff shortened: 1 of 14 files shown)


stscieisenhamer (approved):
Wow, that was a bit more involved than expected.

LGTM

Though tests pass, it may be worth trawling through `stpipe`  to ensure there isn't some hidden crevice where this may be hiding.

hbushouse (approved):
LGTM.

Is there any mention of this keyword and its' possible values in any of the docs?

hbushouse:
CI failure is unrelated.

tapastro:
I don't see any mention of filetype (apart from a separate object/usage wrt. datamodels import) in stpipe and docs - I think we're good to go.

PR #6778, opened 2022-03-25
JP-2561 Add option --force-level1bmodel

stscieisenhamer:
Resolves [JP-2561](https://jira.stsci.edu/browse/JP-2561)

**Description**

This PR adds the command-line option `--force-level1bmodel` for the situation where an input file
does not define the DATAMODL keyword and, instead of attempting to guess at the model type,
force the file to be opened as a `Level1bModel`.

Checklist
- [ ] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -118,6 +118,14 @@ srctype
 
 - Add command line option to override source type [#6720]
 
+1.4.6 (2022-03-25)
+==================
+
+set_telescope_pointing
+----------------------
+
+- Add option --force-level1bmodel. [#6778]
+
 1.4.5 (2022-03-23)
 ==================
 
--- a/jwst/lib/set_telescope_pointing.py
+++ b/jwst/lib/set_telescope_pointing.py
@@ -373,7 +373,8 @@ class TransformParameters:
         return r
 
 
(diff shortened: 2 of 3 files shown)


hbushouse (approved):
Code looks fine, but one question about the use of this option. If the default if False and SDP/DMS run it that way in level-1b processing aren't they still going to get the errors that Mike was reporting? Or will they need to run it with the new arg set to True?

CHANGES.rst
```diff
@@ -118,6 +118,14 @@ srctype
 
 - Add command line option to override source type [#6720]
 
+1.4.6 (2022-03-25)
```
hbushouse, line 121: Is it too late to add (cherry pick) this change into the existing 1.4.5 tag/branch, rather than having to create yet another new one?

hbushouse:
Why is the "Documentation" check box checked above? I don't see any documentation for this new param anywhere (other than in the code doc string).

stscieisenhamer:
Yes, SDP will need to add the option `--force-level1bmodel` to their wrapper script.

PR #6777, opened 2022-03-22
JP-2027: Fix horizontal centering of resampled MIRI-LRS spectra

mcara:
Resolves [JP-2027](https://jira.stsci.edu/browse/JP-2027)

**Description**

This PR fixes the horizontal displacement of resampled spectra for MIRI-LRS. 
This is not a complete re-design of the `build_interpolated_output_wcs()` function but just a patch to fix horizontal displacement.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

PR #6747, opened 2022-02-28
JP-2083: Re-design algorithm for NIRspec output WCS for resample step

mcara:
Closes #6032 #6324
Resolves [JP-2083](https://jira.stsci.edu/browse/JP-2083)

**Description**

This PR significantly re-designs the algorithm for computing output WCS for `resample_spec` for NIRspec. It solves issues described in https://jira.stsci.edu/browse/JP-2083 with resampled spectra: unexplained shifts and features in the resampled spectra.

The fundamental idea is the same as in https://github.com/spacetelescope/jwst/pull/6324 but with enhanced algorithms and bug fixes.

CC: @hbushouse @nden @jmuzerolle 

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

PR #6779, opened 2022-03-28
JP-2027: Restrict fix from 6777 to vertical spectra only

mcara:
Resolves [JP-2027](https://jira.stsci.edu/browse/JP-2027)

**Description**

@hbushouse reported:

> After merging PR https://github.com/spacetelescope/jwst/pull/6777 we're now seeing expected differences in MIRI LRS fixed-slit data (which is good), but we're also seeing differences in regtest results for NIRSpec fixed-slit and MOS data, which are not expected. 

This PR restricts the fix from #6777 to vertical spectra exclusively. I do not *visually see* significant differences for NIRSpec with unarmed eye but I can see how regtests would report differences.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [ ] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -85,6 +85,8 @@ ramp_fitting
 resample
 --------
 
+- Restrict fix from #6777 to "vertical" spectra (i.e., MIRI LRS) only. [#6779]
+
 - Fixed ``resample_spec`` output spectrum centering issue for MIRI LRS fixed-slit. [#6777]
 
 residual_fringe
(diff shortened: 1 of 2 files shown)


hbushouse:
I believe we can close this PR, because we're now convinced that the new results for NIRSpec data are OK, right @mcara ?

mcara:
I do not know what this code will be used for. https://github.com/spacetelescope/jwst/pull/6747 and https://github.com/spacetelescope/jwst/pull/6780 introduced a new approach for NIRSpec data. For MIRI (vertical spectra) and *other horizontal spectra*, the code was "fixed" (centering stuff) in https://github.com/spacetelescope/jwst/pull/6777. 
This PR undoes the fix from https://github.com/spacetelescope/jwst/pull/6777 for *other horizontal spectra*, which, since https://github.com/spacetelescope/jwst/pull/6747 and https://github.com/spacetelescope/jwst/pull/6780 no longer include NIRSpec.

What I want to say is: Do you still want to undo the centering fix for the horizontal spectra (other than NIRSpec)? If yes, then this PR should be merged, if not - then it can be closed.

PR #6781, opened 2022-04-01
Make fit_quality_is_good private and rename it

mcara:
**Description**

This PR is basically a maintenance edition. Here I make method `fit_quality_is_good()` of the `tweakreg` step private (I do not think there is a reason for it to be public) and also I rename it to something that is better suited for what it tests. This function was first introduced in https://github.com/spacetelescope/jwst/pull/6252

@hbushouse Please let me know if you think I should officially deprecate this method instead of renaming it right away. IMO, it is highly unlikely anyone would have used it since it was introduced.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -122,6 +122,13 @@ srctype
 
 - Add command line option to override source type [#6720]
 
+tweakreg
+--------
+
+- Make ``fit_quality_is_good()`` member private and rename it to
+  ``_is_wcs_correction_small()``. [#6781]
+
+
 1.4.6 (2022-03-25)
 ==================
 
@@ -2288,6 +2295,7 @@ tweakreg
 --------
 
 - Updated step arguments in the documentation. [#4723]
+
 
 wfs_combine
 -----------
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -2289,6 +2289,10 @@ tweakreg
 
 - Updated step arguments in the documentation. [#4723]
 
+- Make ``fit_quality_is_good()`` member private and rename it to
```
hbushouse, line 2292: The location of this entry in the change log puts it way down into the section for a very old release. It should be moved to the top section under "unreleased".

hbushouse (commented):
Looks OK to me, other than needing to fix the location of the change log entry.

CHANGES.rst
mcara: My bad

mcara: Fixed changelog

stscieisenhamer (approved):
LGTM as long as Howard's comment is dealt with.

hbushouse (approved):

PR #6783, opened 2022-04-01
JP-898: Update calwebb_spec2 to create WFSS e/sec product

### LINKED ISSUE #3860: Intermediate products in e-/sec for WFSS
Issue [JP-898](https://jira.stsci.edu/browse/JP-898) was created by Alicia Canipe:

{color:#000000}Instrument teams are discussing the need for non-background subtracted, non-flat-fielded, WFSS slope images in e-/sec with valid WCS as an intermediate product of the Stage 2 pipeline. Currently, users will have to take Stage 1 outputs (slope images), install and run the pipeline to add the WCS, and then apply the gain conversion to get units back to e-/sec. However, this is seen as significantly less important and more of a convenience for users than other WFSS requests. It's being added here for tracking purposes. 
{color}

hbushouse:
Closes #3860
Resolves [JP-898](https://jira.stsci.edu/browse/JP-898)

**Description**

This PR updates the WFSS processing branch of the calwebb_spec2 pipeline to allow it save an optional intermediate product that is a version of the grism image converted to units of e-/sec. The image is saved immediately after the flat-field step, before extract_2d has been applied, so that it is the full grism image before any extractions of spectra have been done. The image data are converted to units of e-/sec by loading the gain reference file, computing the mean gain from the reference data, and multiplying the grism image by the mean gain. The units of the variance and error arrays are also converted. The resulting image is saved as new "esec" product type.

Creating and saving this image is optional and is controlled by a new pipeline param "save_wfss_esec", which has a default setting of `False`.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,4 @@
-1.4.6 (unreleased)
+1.4.7 (unreleased)
 ==================
 
 associations
@@ -70,7 +70,10 @@ regtest
 pipeline
 --------
 
-- Improve memory performance of calwebb_detector1 pipeline [#6758]
+- Improve memory performance of `calwebb_detector1` pipeline [#6758]
+
+- Update the `calwebb_spec2` pipeline to allow for the creation of an
+  optional WFSS product that's in units of e-/sec [#6783]
 
 ramp_fitting
 ------------
--- a/docs/jwst/pipeline/calwebb_spec2.rst
+++ b/docs/jwst/pipeline/calwebb_spec2.rst
@@ -112,7 +112,7 @@ If set to ``True``, an intermediate image product is created for WFSS exposures 
 is in units of electrons/sec, instead of the normal DN/sec units that are used throughout
 the rest of processing. This product can be useful for doing off-line specialized
 processing of WFSS images. This product is created after the :ref:`background <background_step>`
-and :ref:`flat-field <flat_field_step>` steps have been applied, but before the
+and :ref:`flat-field <flatfield_step>` steps have been applied, but before the
 :ref:`extract_2d <extract_2d_step>` step, so that it is the full WFSS image. The conversion
 to units of electrons/sec is accomplished by loading the :ref:`GAIN <gain_reffile>` reference file,
 computing the sigma-clipped mean gain across all pixels, and multiplying the WFSS image by the
--- a/docs/jwst/pipeline/calwebb_spec2.rst
+++ b/docs/jwst/pipeline/calwebb_spec2.rst
@@ -113,9 +113,9 @@ The ``calwebb_spec2`` pipeline has two optional arguments.
   and :ref:`flat-field <flatfield_step>` steps have been applied, but before the
   :ref:`extract_2d <extract_2d_step>` step, so that it is the full WFSS image. The conversion
   to units of electrons/sec is accomplished by loading the :ref:`GAIN <gain_reffile>` reference file,
-  computing the sigma-clipped mean gain across all pixels, and multiplying the WFSS image by the
-  mean gain.  The intermediate file will have a product type of "_esec". Only applies to
-  WFSS exposures.
+  computing the mean gain across all pixels (excluding reference pixels), and multiplying the WFSS
+  image by the mean gain.  The intermediate file will have a product type of "_esec".
+  Only applies to WFSS exposures.
 
 Inputs
 ------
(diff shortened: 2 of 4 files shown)


KevinVolkSTScI (commented):
Another comment is that we may want to mask off the reference pixels and use the normal numpy mean on the masked image rather than using sigma_clipped_stats, although I have to see how much this difference makes in the output mean value before I can say more.

hbushouse:
> One quick comment looking at the code: for the mean gain one should always use the mean of the full image irrespective of whether a subarray is being used. Some NIRISS calibration exposures use subarrays (I am not sure about NIRCam) but for these we still want the same mean gain as is used in the full frame case. Whatever gain variations there are across the detector are assumed to be taken out in the flat fielding step.

Theoretically this shouldn't be an issue, because (as a comment in the code indicates) WFSS observations should always be taken in full-frame mode and hence the full-frame gain ref image should always be retrieved and used in the computation. If for some bizarre reason the WFSS science image were taken in subarray mode, I could remove the logic that extracts a matching subarray from the gain ref file and always use the full gain image. That should guarantee that the mean of the full gain image is always computed.

hbushouse:
> Another comment is that we may want to mask off the reference pixels and use the normal numpy mean on the masked image rather than using sigma_clipped_stats, although I have to see how much this difference makes in the output mean value before I can say more.

I thought of that too and did a bit of experimenting before implementing the code. Example gain ref files for both NIRCam and NIRISS, which have the ref pixels set to either 0 or 1, seemed to get clipped out nicely due to the very small sigma of the remaining good values up around 1.6-2.0. The sigma is on the order of 0.03 or something like that, at least after a few clipping iterations.

KevinVolkSTScI:
Howard, how much did the mean gain change between the two algorithms?  If the change is say sub-1% then I would not worry about which method is used.  But if one saw a 10% change that would be of concern as to which is the one to use.

We have photometric standard star observations in the WFSS sub-arrays for NIRISS.  While these subarray are not used for science, I think that they should be treated the same way as the science exposures in this step.

tapastro (changes requested):
Code looks good to me - but could maybe add a regression test for coverage's sake, just using whatever WFSS products we already hold?

hbushouse:
> Howard, how much did the mean gain change between the two algorithms? If the change is say sub-1% then I would not worry about which method is used. But if one saw a 10% change that would be of concern as to which is the one to use. The sigma implies of order 3% variations might be possible, which would be in a grey area between these two.
> 
> We have photometric standard star observations in the WFSS sub-arrays for NIRISS. While these subarray are not used for science, I think that they should be treated the same way as the science exposures in this step.

For the currently active NIRISS gain ref file, /grp/crds/jwst/references/jwst/jwst_niriss_gain_0006.fits, which has all of the reference pixels set to 1.0, I get the following results:
`
In [10]: sigma_clipped_stats(gain_model.data)
Out[10]: (1.6214292, 1.6101412, 0.035471577)

In [11]: sigma_clipped_stats(gain_model.data[4:-4,4:-4])
Out[11]: (1.6214294, 1.6101407, 0.03547038)

In [12]: gain_model.data[4:-4,4:-4].mean()
Out[12]: 1.6221113
`
So the overall mean without clipping and without the ref pix is 1.6221, which is identical to the value stored in the extension header of the gain ref file. The clipped results with and without the ref pix are essentially identical, because the clipping is removing the ref pix. The difference between the clipped results 1.62143 and the simple mean 1.6221 is 0.04%. The sigma computed by the clipped stats routine is 0.03547, which is consistent with the small variations in the gain image.

hbushouse:
@KevinVolkSTScI  I'm happy to switch to using the simple mean of all the non-reference pixels, if you're more comfortable with that.

hbushouse:
> Code looks good to me - but could maybe add a regression test for coverage's sake, just using whatever WFSS products we already hold?

Added the esec product to existing regtest.

tapastro (approved):
LGTM

KevinVolkSTScI:
I have been bogged down with meetings and other stuff this afternoon.  Given the values of the mean gain shown by Howard it does not matter which method is used, so going with the current code would be fine.

KevinVolkSTScI:
I would still be more comfortable if the code just took the mean of the full frame and applied it to all observations.  The NIRISS WFSS sub-arrays are in an area where the gain is higher than average, as I recall.  Maybe this will have no effect on the reduction of standard star spectra (we of course do not want any contamination in these observations so perhaps these products in e/s will never be used anyway).

KevinVolkSTScI (approved):

hbushouse:
The last commit modifies the new code to compute the simple mean of the entire gain image, excluding only reference pixels and pixels with gain values of NaN and <= 0, instead of using the sigma-clipped mean.

PR #6784, opened 2022-04-04
JP-2564 Update Reset Step to work with segmented TSO data

jemorrison:
Resolves [JP-2564](https://jira.stsci.edu/browse/JP-2564)

**Description**

This PR fixes the reset step to work with segmented data. The reset step is integration dependent. There is a correction for the first integration and then a correction for all other integrations. In the case of segmented data, the first integration in the file is not necessarily the first integration in the exposure. The integration number in the exposure needs to be use in the step rather than the assuming the first integration in the file is the first integration in the exposure. 

Checklist
- [X] Tests
- [ ] Documentation
- [X] Change log
- [X] Milestone
- [X] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -67,6 +67,11 @@ regtest
 
 - Added a residual fringe correction test [#6771]
 
+reset
+-----
+
+- Fix bug in how segemented data is corrected [#6784]
+
 pipeline
 --------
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):
Looks like this should work OK. Just one question regarding the application to only some of the groups.

PR #6787, opened 2022-04-06
JP-2565 Implement PoolRow to avoid deep copy of the AssociationPool table

stscieisenhamer:
Resolves [JP-2565](https://jira.stsci.edu/browse/JP-2565)

**Description**
A necessary operation during generation is the deep copying of rows from the `AssociationPool`. However, because `AssociationPool` is subclassed from`astropy.table.Table`, each row has a pointer back to the table it belongs to, each row deep copy would copy the whole table. This causes exponential memory use as the tables grow in length.

A new class, `PoolRow`, is implemented to return a stripped-down version of each row, without the extra baggage.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 associations
 ------------
 
+- Implement PoolRow to avoid deep copy of the AssociationPool table [#6787]
+
 - Added valid optical paths for NRS_LAMP observations to generate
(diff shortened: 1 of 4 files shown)


tapastro (approved):
LGTM - tested and reproduced the lowered memory usage.

PR #6788, opened 2022-04-06
JP-2570: WFSS spec2 updates to handle crowded fields

hbushouse:
Resolves [JP-2570](https://jira.stsci.edu/browse/JP-2570)

**Description**

This PR makes a couple of updates to the calwebb_spec2 background step as applied to WFSS exposures. First, it checks the mask created for use in scaling the bkg ref image to make sure it has enough non-source pixels to work with, in order to avoid crashing on images that have every pixel covered by a source spectrum. Second, it adds the "mmag_extract" parameter to the background step (just like extract_2d), so that a user can set the minimum brightness of sources to be used during the background subtraction process. It also just cleans up some of the logging messages in a couple other steps that relate to WFSS processing.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


tapastro:
Code looks good, but is this the final form, or are we being asked to provide an N_sources cutoff instead of the current magnitude cutoff?

hbushouse:
> Code looks good, but is this the final form, or are we being asked to provide an N_sources cutoff instead of the current magnitude cutoff?

Correct. We'll be expanding this to include that new param as well. This initial set of updates at least avoids the crashes that are currently happening in the background statistics computation.

hbushouse:
Latest commit adds capabilities for the new "wfss_nbright" param in the "extract_2d" step. This is used to limit the list of sources processed to the N brightest (based on isophotal_abmag). It will have a default value of 300, as suggested by WFSS working group reps. Had to make changes in some related modules to fully implement this.

hbushouse:
@nden @tapastro ready for review (again)

jwst/extract_2d/extract_2d.py
```diff
@@ -18,7 +18,8 @@ def extract2d(input_model,
               tsgrism_extract_height=None,
               wfss_extract_half_height=None,
               extract_orders=None,
-              mmag_extract=99.):
+              mmag_extract=99.,
+              nbright=999):
```
tapastro, line 22: Should this default be here, if we have a default set above this scope?

jwst/extract_2d/grisms.py
```diff
@@ -273,7 +273,8 @@ def extract_grism_objects(input_model,
                           extract_orders=None,
                           mmag_extract=99.,
                           compute_wavelength=True,
-                          wfss_extract_half_height=None):
+                          wfss_extract_half_height=None,
+                          nbright=999):
```
tapastro, line 277: Same question on this default value.

jwst/extract_2d/grisms.py
```diff
@@ -318,10 +322,13 @@ def extract_grism_objects(input_model,
...
-    magnitude-cutoff value. Sources with magnitudes fainter than the
-    extraction cutoff (MMAG_EXTRACT)  will not be extracted, but are
+    magnitude-cutoff value. The total list of objects to extract is limited
+    by both MMAG_EXTRACT and NBRIGHT. Sources with magnitudes fainter than the
+    extraction cutoff (MMAG_EXTRACT) will not be extracted, but are
     accounted for when computing the spectral contamination and background
-    estimates. The default extraction value is 99 right now.
+    estimates. The default value is 99 right now.
+    NBRIGHT further limits the list to the NBRIGHT brightest objects.
+    The default value is 999 right now.
```
tapastro, line 331: Should this be 300 now?

tapastro (commented):
Looks good, just some comments.

jwst/extract_2d/extract_2d.py
hbushouse: You're right that we have a default value set in the step class, but I was thinking of a use case where someone might call the `extract2d` function directly, without including `nbright`, in which case it would use this (different) default.

hbushouse:
Further investigation into `astropy.table.QTable` sort capabilities shows that we could in fact sort the object table on the `isophotal_abmag` column, but the object list is only in the form of a QTable for a very short time when it's first loaded in the `get_object_info` function in assign_wcs/util.py. It is immediately rewritten as a list of `GrismObject`s, which no longer has a sort capability, because it's a list of objects. So I'm going to leave the current code as-is that manually sorts the list of objects by magnitude.

jwst/extract_2d/grisms.py
hbushouse: The 300 value will be used only for crowded filed situations, which will be set in the pars-spec2pipeline ref file. The default in the code will now be 1000 (see the latest commit) for all other situations.

jwst/assign_wcs/util.py
```diff
@@ -822,10 +829,35 @@ def _create_grism_bbox(input_model, mmag_extract=99.0,
...
+        mags = []
+        for obj in grism_objects:
+            mags.append(obj.isophotal_abmag)
+        indxs = np.argsort(mags)
+
+        # Create a final grism object list containing only the N brightest objects
+        final_objects = []
+        for i in list(indxs[:nbright]):
+            final_objects.append(grism_objects[i])
+        del grism_objects
```
nden, line 854: One way to collapse 845 - 853 
```
indxs = np.argsort([obj.isophotal_abmag for obj in grism_objects])
final_objects = []
final_objects = [grism_objects[i] for i in indxs[:nbright]]
```

nden: It's possible also to sort the original table in get_object_info

```
catalog =QTable.read(catalog_file, format='ascii.ecsv')
catalog.sort('isophotal_abmag')
```

However, this appears to be much slower than working with lists.

nden (approved):

jwst/assign_wcs/util.py
hbushouse: > One way to collapse 845 - 853
> 
> ```
> indxs = np.argsort([obj.isophotal_abmag for obj in grism_objects])
> final_objects = []
> final_objects = [grism_objects[i] for i in indxs[:nbright]]
> ```

Implemented in latest commit. Thanks!

tapastro (approved):
LGTM

Issue #6359, opened 2021-09-21
Possibly incorrect computation of "crpix" (or offsets) in assign_wcs' wcs_from_footprints


mcara:
In [`wcs_from_footprints`](https://github.com/spacetelescope/jwst/blob/82d5766b7e08c7e39d15dd3e6ec701b50d93995c/jwst/assign_wcs/util.py#L196), see https://github.com/spacetelescope/jwst/blob/82d5766b7e08c7e39d15dd3e6ec701b50d93995c/jwst/assign_wcs/util.py#L294-L298
input coordinates are shifted by the coordinates of the middle of the bounding box. In FITS WCS, this corresponds to CRPIX. 
Normally, the value of CRPIX is tied to CRVAL ("fiducial" point in this code). Setting CRPIX independently from CRVAL results in a shift in the coordinate system.

For the resample step this may not be relevant as long as input images are aligned relative to each other. However, this may result in incorrect WCS (in the absolute coordinate system, or, i.e., relative to GAIA) of the resampled product.

At first glance, fiducial point is computed as the mean of the cartesian spherical (on the celestial sphere) coordinates of the bounding boxes' corners (on the sky) and so computing "CRPIX" as mean of bounding boxes' corners of the images (in image coordinates - NOTE: this plane is tangent to the sphere) may seem like a reasonable estimate of CRPIX. However, IMO, this is very likely is not correct because the transformation from a tangent plane (image) to the spherical coordinates is non-linear.

hbushouse:
Fixed by #6789

PR #6789, opened 2022-04-07
JP-2563: Fix WCS bounding box offset in level 3 mosaic output

tapastro:
Closes #
Resolves [JP-2563](https://jira.stsci.edu/browse/JP-2563)
Resolves [JP-2558](https://jira.stsci.edu/browse/JP-2558)

**Description**

This PR corrects the computation of crpix from the midpoint of the axes' extents to a backward transform of the fiducial point into pixel coordinates. 

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


jwst/assign_wcs/util.py
```diff
@@ -366,15 +366,30 @@ def compute_fiducial(wcslist, bounding_box=None):
...
+        x_mid = (np.max(np.cos(lat) * np.cos(lon)) +
+                 np.min(np.cos(lat) * np.cos(lon))) / 2.
+        y_mid = (np.max(np.cos(lat) * np.sin(lon)) +
+                 np.min(np.cos(lat) * np.sin(lon))) / 2.
+        z_mid = (np.max(np.sin(lat)) + np.min(np.sin(lat))) / 2.
+        lon_fiducial = np.rad2deg(np.arctan2(y_mid, x_mid)) % 360.0
+        lat_fiducial = np.rad2deg(np.arctan2(z_mid, np.sqrt(x_mid ** 2 + y_mid ** 2)))
         fiducial[spatial_axes] = lon_fiducial, lat_fiducial
     if spectral_footprint.any():
         fiducial[spectral_axes] = spectral_footprint.min()
```
stsci-hack, line 395: These lines in the function should be removed entirely as they are completely replaced by the code you added above that works perfectly well based on my own tests with the changes.

stsci-hack (changes requested):
Clean up the function as noted and I think you have a solution that works as well as possible.

jwst/assign_wcs/util.py
tapastro: Yeah, sorry - that commit was a code-share more than a well-written commit with intent to merge 😬 

I'm still seeing a slight exclusion of data in my test - not sure if I'm missing some slight shift somewhere. Also thinking I might need to take the mean in spherical coordinate space to avoid issues crossing 359->0 in RA deg space, but I'll try to clean this up and get a better commit in place.

stsci-hack: No need to worry.  This is one advantage of the PR, it can be modified as many times as needed before being merged, and allows (as you are doing here) others to see what you are doing.  I know I have thought I removed such extra code before only to find it there later.  So far, so good.

nden: I think it's better to work in 3D cartesian space than in spherical coordinates. And yes, there's still an ofset.

jwst/assign_wcs/util.py
```diff
@@ -366,15 +366,32 @@ def compute_fiducial(wcslist, bounding_box=None):
...
+                *w.footprint(bounding_box=bounding_box).T) for w in wcslist])
+
+            output_bounding_box = []
+            for axis in wcslist[i].output_frame.axes_order:
+                axis_min, axis_max = domain_bounds[axis].min(), domain_bounds[axis].max()
+                output_bounding_box.append((axis_min, axis_max))
+
+            sky_bounds = np.array(wcs_init.forward_transform(*output_bounding_box))
+            lon[i * 2:(i + 1) * 2] = sky_bounds[0]
+            lat[i * 2:(i + 1) * 2] = sky_bounds[1]
```
nden, line 386: I think lines 374-386 recompute `spatial_footprint` in a roundabout way.
I would revert this code.

jwst/assign_wcs/util.py
```diff
@@ -366,15 +366,32 @@ def compute_fiducial(wcslist, bounding_box=None):
...
-        x_mean = np.mean(np.cos(lat) * np.cos(lon))
-        y_mean = np.mean(np.cos(lat) * np.sin(lon))
-        z_mean = np.mean(np.sin(lat))
-        lon_fiducial = np.rad2deg(np.arctan2(y_mean, x_mean)) % 360.0
-        lat_fiducial = np.rad2deg(np.arctan2(z_mean, np.sqrt(x_mean ** 2 + y_mean ** 2)))
+        x_mid = (np.max(np.cos(lat) * np.cos(lon)) +
+                 np.min(np.cos(lat) * np.cos(lon))) / 2.
+        y_mid = (np.max(np.cos(lat) * np.sin(lon)) +
+                 np.min(np.cos(lat) * np.sin(lon))) / 2.
+        z_mid = (np.max(np.sin(lat)) + np.min(np.sin(lat))) / 2.
```
nden, line 392: It may make sense to precompute `cos(lat)`, `cos(lon)`, ... , and use them without recomputing.

nden:
I think I have a fix which happens to be also a fix to #6359. Could you make the above changes and I'll push to this branch.

stsci-hack (approved):
(Review based on commit 6b0e3cb): Perfectly centered results in my tests using this code, and the computations more closely match the overall logic used successfully to perform the same type of computation for HST data.

jwst/assign_wcs/util.py
```diff
@@ -278,7 +278,7 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
                 fiducial[k] = crval[i]
                 i += 1
 
-    ref_fiducial = compute_fiducial([refmodel.meta.wcs])
+    ref_fiducial = np.array([refmodel.meta.wcsinfo.ra_ref, refmodel.meta.wcsinfo.dec_ref])
```
stsci-hack, line 281: If a reference model is specified on input, then this fiducial should be based at that reference model's reference point, not just a random position in the center of it's bounding box.  It may not make much of a difference, but this will affect the computation of the final output plate scale later and will now accurately describe the plate scale of the desired output reference model.  It may prove worthwhile to add a comment to the code to this effect for later maintenance.

nden (approved):

PR #6791, opened 2022-04-11
JP0-2506: Fix ZEROFRAME Default Allocation Dimensions

### LINKED ISSUE #6703: Incorrect zeroframe dimensions when reading data with RampModel
_Issue [JP-2506](https://jira.stsci.edu/browse/JP-2506) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

I'm getting some strange results when I try reading in some NIRISS dark current files using RampModel. The data array in the file is composed of a single integration with 100 groups. When I use RampModel, the model.data array looks correct, but the model.zeroframe array has a shape of 100x2048x2048, when I was expecting to see 1x2048x2048. 

Note that the data I am looking at are relatively old darks, and the uncal files do not contain a ZEROFRAME extension. So this may be a problem with the files, or a problem in RampModel with the way the zeroframe is defined in cases where it is not present in the input file. Since it seems like a rare occurrence to see an uncal file with no zeroframe extension, this may not be a high priority problem, but I wanted to report it just to be sure.

kmacdonald-stsci:
Closes #6703 
Resolves [JP-2506](https://jira.stsci.edu/browse/JP-2506)

**Description**

This PR addresses the incorrect ZEROFRAME dimension allocation.  Currently the ZEROFRAME default dimensions are (ngroups, nrows, ncols), but the dimensions should be (nints, nrows, ncols).  

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


CHANGES.rst
```diff
@@ -92,6 +92,8 @@ ramp_fitting
 - Adding feature to turn off calculations of ramps with good 0th group,
   but all other groups are saturated. [#6737]
 
+- Correcting the default ZEROFRAME allocation. [#6791]
```
hbushouse, line 95: This entry should go under the datamodels section. Not *everything* is for ramp_fitting! ;-)

hbushouse (approved):
LGTM. Just need to fix the change log entry.

CHANGES.rst
kmacdonald-stsci: I was using the six degrees of separation of ramp fitting for the change log, but I'll do it your way 😆

hbushouse (approved):

PR #6793, opened 2022-04-13
JP-2573: Allow reference file to specify extraction region for extended sources

tapastro:
Resolves [JP-2573](https://jira.stsci.edu/browse/JP-2573)

**Description**

This PR addresses lacking functionality in user-specified extraction regions via reference files for extended sources. It also found a bug in that a user-specified bkg_fit was not being utilized in get_extraction_parameters.

Checklist
- [x] Tests
- [x] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)


jwst/extract_1d/extract.py
```diff
@@ -338,50 +338,45 @@ def get_extract_parameters(
...
-                        extract_params['subtract_background'] = False
-                        extract_params['use_source_posn'] = False
-                        extract_params['xstart'] = 0
-                        extract_params['xstop'] = shape[-1] - 1
-                        extract_params['ystart'] = 0
-                        extract_params['ystop'] = shape[-2] - 1
-                        extract_params['extract_width'] = None
-                        extract_params['independent_var'] = 'pixel'
-                        extract_params['position_correction'] = 0  # default value
+                        extract_params['xstart'] = aper.get('xstart', 0)
```
hbushouse, line 342: What's the 2nd argument of the `aper.get()` function? The default value to use if the attribute doesn't exist?

jwst/extract_1d/extract.py
```diff
@@ -338,50 +338,45 @@ def get_extract_parameters(
...
 
                     if meta.target.source_type == "EXTENDED":
-                        log.info("Target is extended, so the entire region will be extracted.")
                         shape = input_model.data.shape
-                        extract_params['src_coeff'] = None
-                        extract_params['bkg_coeff'] = None
-                        extract_params['bkg_fit'] = None
-                        extract_params['bkg_order'] = 0
-                        extract_params['subtract_background'] = False
-                        extract_params['use_source_posn'] = False
```
hbushouse, line 348: So does the removal of this setting mean that we will always try to use the computed source position to set the center of the extraction region for extended sources? I'm not sure that's a good idea. That could end up overriding the start/stop limits entered by the user, or even if the user doesn't give any it could also end up trying to shift the default limits (the whole region) later on in the processing. Although there might be guard rails in place to make sure the shifted start/stop values don't go beyond the limits of the data array.

jwst/extract_1d/extract.py
tapastro: Yes, this will return 0 if `xstart` is not in the aper dict.

jwst/extract_1d/extract.py
tapastro: After a regression test run, it appears I did a *bad thing*. Looking into it.

tapastro:
Need to show a difference between a user specifying `bkg_fit` versus the default value, so setting the default to None, which will be set to 'poly' as before, if no value is found in the reference file.

hbushouse (approved):
LGTM, except of course for the lack of a change log entry ;-)

stscieisenhamer (changes requested):
Documentation stickler: The extract_1d documentation needs to be updated. The obvious one is the change in default for `poly`. Not sure about how the rest affects the docs but it should be reviewed.

stscieisenhamer (approved):

stscieisenhamer:
changelog?

PR #6796, opened 2022-04-15
JP-2566: Fix resample_spec parameter handling

hbushouse:
Resolves [JP-2566](https://jira.stsci.edu/browse/JP-2566)

**Description**

This PR makes changes to the resample and resample_spec steps to allow for proper use of user-supplied parameter values. In the case where there's no DRIZPARS parameter ref file from which to load parameter values (as is the case for NIRSpec), the old code was hardwiring parameters to default values, thus ignoring any user-supplied parameter values. The function `_set_spec_defaults` has been drastically modified to fix this situation. The current flow is to first initialize the values in `kwargs` from the default values specified in the step "spec". It then applies values retrieved from the active params, thus passing along any user-supplied param values for items like `pixfrac`, `fillval`, and `weight_type`. No hardwired values are assigned any more in the code.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/resample/resample_step.py
+++ b/jwst/resample/resample_step.py
@@ -78,17 +78,18 @@ class ResampleStep(Step):
         if ref_filename != 'N/A':
             self.log.info('Drizpars reference file: {}'.format(ref_filename))
             kwargs = self.get_drizpars(ref_filename, input_models)
-
         else:
             # If there is no drizpars reffile
             self.log.info("No DRIZPARS reffile")
             kwargs = self._set_spec_defaults()
 
+        kwargs['allowed_memory'] = self.allowed_memory
+        kwargs['weight_type'] = str(self.weight_type)
+
         # Issue a warning about the use of exptime weighting
         if self.weight_type == 'exptime':
             self.log.warning("Use of EXPTIME weighting will result in incorrect")
             self.log.warning("propagated errors in the resampled product")
-        kwargs['allowed_memory'] = self.allowed_memory
 
         # Custom output WCS parameters.
         # Modify get_drizpars if any of these get into reference files:
--- a/jwst/resample/resample_step.py
+++ b/jwst/resample/resample_step.py
@@ -81,14 +81,14 @@ class ResampleStep(Step):
 
         else:
             # If there is no drizpars reffile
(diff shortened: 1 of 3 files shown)


jwst/resample/resample_step.py
```diff
@@ -225,20 +226,16 @@ def _set_spec_defaults(self):
...
-            kwargs['pixfrac'] = 1.0
-        if kwargs['kernel'] is None:
-            kwargs['kernel'] = 'square'
-        if kwargs['fillval'] is None:
-            kwargs['fillval'] = 'INDEF'
-        if kwargs['weight_type'] is None:
-            kwargs['weight_type'] = 'ivm'
+        kwargs['pixfrac'] = self.pixfrac
+        kwargs['kernel'] = str(self.kernel)
+        kwargs['fillval'] = str(self.fillval)
```
tapastro, line 231: Should there be similar lines in resample_spec for pixfrac, kernel, and fillval as what you've added for weight_type? I don't know if they get called/used in the spectral version, but I do at least see the pixfrac value is added to the output product in resample_spec_step's _process_multislit().

tapastro (approved):
Looks good, just a question on some other parameters involved in resample vs. resample_spec.

jwst/resample/resample_step.py
hbushouse: The code you're seeing here is the `_set_spec_defaults` function, which lives in the resample_step module, but is called by both resample and resample_spec. So in fact the identical settings are being used for both.

tapastro: Sorry, I didn't read closely enough, though a question still remains - upon looking more closely, it looks like get_drizpars() tries to deal with the combination of user vs. reference values for these keywords. But in the case where drizpars exists, would pixfrac still get overridden to a value of 1.0 because it's the default spec value, even though the user has not specified it? The extract_1d bkg_fit issue has me tuned into this problem now.

tapastro: From a quick test with NIRCam's drizpars file, it looks like the reference values are being ignored because the default spec values are not None.

PR #6800, opened 2022-04-19
Initial datamodel/schema for MiriMrsXArtCorrModel

drlaw1558:
**Description**

This PR implements an initial data model and schema for the MIRI MRS Cross Artifact correction reference file ([CRDS-545](https://jira.stsci.edu/browse/CRDS-545))

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/datamodels/mrsxartcor.py
+++ b/jwst/datamodels/mrsxartcor.py
@@ -1,0 +1,25 @@
+from .reference import ReferenceFileModel
+
+__all__ = ['MiriMrsXArtCorrModel']
+
+
+class MiriMrsXArtCorrModel(ReferenceFileModel):
+    """
+    A data model for MIRI MRS cross-artifact corrections file.
+
+    Parameters
+    __________
+    init : any
+        Any of the initializers supported by `~jwst.datamodels.DataModel`.
+
+    ch1a_table : numpy table
+         Cross artifact correction parameters for Channel 1A
+
+    ch1b_table : numpy table
+         Cross artifact correction parameters for Channel 1B
+
+    ch1c_table : numpy table
+         Cross artifact correction parameters for Channel 1C
+
+    """
+    schema_url = "http://stsci.edu/schemas/jwst_datamodel/miri_mrsxartcor.schema"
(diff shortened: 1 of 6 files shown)


jwst/datamodels/mrsxartcor.py
```diff
@@ -0,0 +1,25 @@
+from .reference import ReferenceFileModel
+
+__all__ = ['MiriMrsXArtCorrModel']
+
+
+class MiriMrsXArtCorrModel(ReferenceFileModel):
```
stscieisenhamer, line 6: Suggest name change to begin with just `Mir` to be `MirMrsXArtCorr`. This is then consistent with other MIRI-specific models.

hbushouse (approved):
Looks good to me. Only thing missing is the addition of documentation in the reference files section of RTD, but that can come later once this is actually in use.

stscieisenhamer (changes requested):

jwst/datamodels/mrsxartcor.py
stscieisenhamer: Thought I do note there is an `MiriIFUCubeParsModel`, but others, such as `MirLrsPhotomModel` have just the three letters.

hbushouse: I actually see a mixture of "Mir", "Miri", and "MIRI" in use in various datamodel class names, but I think the trend of the more recent ones is in the direction of "Mir".

jwst/datamodels/schemas/miri_mrsxartcorr.schema.yaml
```diff
@@ -0,0 +1,63 @@
+%YAML 1.1
+---
+$schema: "http://stsci.edu/schemas/fits-schema/fits-schema"
+id: "http://stsci.edu/schemas/jwst_datamodel/miri_mrsxartcorr.schema"
+title: MIRI MRS Cross Artifact Corrections
+allOf:
+- $ref: referencefile.schema
```
stscieisenhamer, line 7: Will need the EXP_TYPE and P_EXP_TY keywords in the schema. These can be inserted using
```
```

stscieisenhamer (changes requested):

jwst/datamodels/schemas/miri_mrsxartcorr.schema.yaml
drlaw1558: Added.

stscieisenhamer (approved):

PR #6801, opened 2022-04-19
JP-2506: Adding a conditional to the creation of the ZEROFRAME array.

kmacdonald-stsci:
Closes #
Resolves [JP-2506](https://jira.stsci.edu/browse/JP-2506)

**Description**

This PR addresses creating the ZEROFRAME array when desired in the RampModel.  A conditional has now been added to the `__init__` function to check if the model is being created with a tuple or the `self.meta.exposure.zero_frame` flag is set to `True`.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/datamodels/ramp.py
+++ b/jwst/datamodels/ramp.py
@@ -44,14 +44,15 @@ class RampModel(JwstDataModel):
         self.groupdq = self.groupdq
         self.err = self.err
 
-        try:
-            self.getarray_noinit("zeroframe")
-        except AttributeError:
-            # If "zeroframe" is not in the instance, create a zero array with
-            # the correct dimensions.
-            nints, ngroups, nrows, ncols = self.data.shape
-            dims = (nints, nrows, ncols)
-            self.zeroframe = np.zeros(dims, dtype=self.data.dtype)
+        if isinstance(init, tuple) or self.meta.exposure.zero_frame == True:
+            try:
+                self.getarray_noinit("zeroframe")
+            except AttributeError:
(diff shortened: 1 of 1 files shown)


jwst/datamodels/ramp.py
```diff
@@ -44,14 +44,15 @@ def __init__(self, init=None, **kwargs):
...
 
-        try:
-            self.getarray_noinit("zeroframe")
-        except AttributeError:
-            # If "zeroframe" is not in the instance, create a zero array with
-            # the correct dimensions.
-            nints, ngroups, nrows, ncols = self.data.shape
-            dims = (nints, nrows, ncols)
-            self.zeroframe = np.zeros(dims, dtype=self.data.dtype)
+        if isinstance(init, tuple) or self.meta.exposure.zero_frame == True:
```
hbushouse, line 47: Not 100% sure we need the check against `meta.exposure.zero_frame`, because if we're loading an existing file into a model and it has the zeroframe extension in the file already (and hence `meta.exposure.zero_frame` will be True), we don't need to do anything, because the zeroframe attribute will already be populated in the datamodel (so we don't even need to try `getarray_noinit("zeroframe")`). But I guess it can't hurt to have it there.

I tried this against a couple of existing files that do and do not have zeroframe in them and the results are as desired, as is the case where I create a model from scratch (i.e. init is a tuple). So I guess it's good to go.

hbushouse (approved):

hbushouse:
There's one pycodestyle complaint that needs fixing.

PR #6809, opened 2022-04-28
JP-2569: Made defaults for some tweakreg parameters more robust

mcara:
Partially Resolves [JP-2569](https://jira.stsci.edu/browse/JP-2569)

**Description**

This PR updates ``searchrad``, ``tolerance``, and ``separation`` parameters for the ``trweakreg`` step to, hopefully, be more balanced. Also ``separation`` should be larger than ``tolerance``.

Checklist
- [ ] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/docs/jwst/tweakreg/README.rst
+++ b/docs/jwst/tweakreg/README.rst
@@ -101,14 +101,14 @@ The ``tweakreg`` step has the following optional arguments:
   for matching. (Default=15)
 
 * ``searchrad``: A `float` indicating the search radius in arcsec for a match.
-  (Default=1.0)
+  (Default=2.0)
 
 * ``use2dhist``: A boolean indicating whether to use 2D histogram to find
   initial offset. (Default=True)
 
-* ``separation``: Minimum object separation in arcsec. (Default=0.5)
+* ``separation``: Minimum object separation in arcsec. (Default=1.0)
 
-* ``tolerance``: Matching tolerance for ``xyxymatch`` in arcsec. (Default=1.0)
+* ``tolerance``: Matching tolerance for ``xyxymatch`` in arcsec. (Default=0.7)
 
 * ``xoffset``: Initial guess for X offset in arcsec. (Default=0.0)
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -198,6 +198,9 @@ tweakreg
 - Make ``fit_quality_is_good()`` member private and rename it to
   ``_is_wcs_correction_small()``. [#6781]
 
+- Change default settings for ``searchrad``, ``tolerance``, and ``separation``
+  parameters for the ``tweakreg`` step. [#6809]
+
 - Change default value of ``brightest`` parameter in the ``tweakreg`` step. [#6810]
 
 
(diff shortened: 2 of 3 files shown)


docs/jwst/tweakreg/README.rst
```diff
@@ -101,14 +101,14 @@ The ``tweakreg`` step has the following optional arguments:
...
 
 * ``searchrad``: A `float` indicating the search radius in arcsec for a match.
-  (Default=1.0)
+  (Default=2.0)
 
 * ``use2dhist``: A boolean indicating whether to use 2D histogram to find
   initial offset. (Default=True)
 
-* ``separation``: Minimum object separation in arcsec. (Default=0.5)
+* ``separation``: Minimum object separation in arcsec. (Default=1.0)
```
hbushouse, line 109: For NIRCam short-wave detectors, with a pixel scale of ~0.03", this implies a separation of ~33 pixels. So this setting means that detected sources that are closer to one another than 33 pixels will not be used in the solution? Given that you have 2048x2048 pixel space to work with, I guess this may be reasonable and still give you plenty of sources to work with, but it seems to me (IMHO) that you could have very valid results for sources closer together than that, especially since the NIRCam short-wave detectors are undersampled at short wavelengths (hence the PSF is very compact). But I'll leave it to folks with more experience at this to decide.

hbushouse: I guess this is also a strong motivation for having parameter ref files in CRDS that tailor the defaults to each detector/filter/wavelength combination.

docs/jwst/tweakreg/README.rst
```diff
@@ -101,14 +101,14 @@ The ``tweakreg`` step has the following optional arguments:
   for matching. (Default=15)
 
 * ``searchrad``: A `float` indicating the search radius in arcsec for a match.
-  (Default=1.0)
+  (Default=2.0)
```
hbushouse, line 104: I guess we'll have to wait till we've had more experience with real on-sky JWST observations using fine guide mode to have a better feel for the accuracy of blind pointing. I think everyone expects it to be better than 1 arcsec. I'm worried we may be driven right now by somewhat spurious and atypical results coming out of some commissioning data that were obtained in off-nominal circumstances (e.g. coarse guiding mode, etc.).

hbushouse (approved):
These suggested changes seem OK to me, despite the fact that they will have no effect on actual processing, because of the existence of CRDS parameter ref files for all imaging instruments that will override these code values. So in that sense, it doesn't really matter what we set these values to!

docs/jwst/tweakreg/README.rst
mcara: See my comment below - https://github.com/spacetelescope/jwst/pull/6809#issuecomment-1112252807 but there is very little difference between instruments in terms of alignment errors in units of arcsec (not pixels). These are mostly dictated by pointing accuracy of the observatory. If you are aligning images to an external catalog such as GAIA - then it is about absolute pointing errors and if you are aligning  a set of images relative to each other - then we are talking about relative error in alignment (which could be smaller then absolute errors) - I should add this to my longer comment below.

stsci-hack: For alignment, we only need to use sources with the highest probability of being cross-matched correctly as opposed to trying to cross-match every source in the field which can lead to a lot of confusion in crowded fields.   So, a larger separation in concert with reducing the number of sources used for alignment makes sense.

stsci-hack (approved):
These changes should not cause any problems as far as I can tell.

PR #6810, opened 2022-04-28
JP-2583: Adjust tweakreg's 'brightest' and update docs

mcara:
Resolves [JP-2583](https://jira.stsci.edu/browse/JP-2583)

**Description**

This PR updates  the default value of the ``brightest``  parameter of the ``tweakreg`` step to, hopefully, be more balanced. Also update docs to reflect this new value.

Checklist
- [ ] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


stsci-hack (approved):
This is similar to the value used for HST alignment (which is 250/chip), a value which has worked well.  Hopefully, this value will vary for each detector, since each detector's field-of-view and bandpass will affect the typical number of sources observed by the detector.

PR #6811, opened 2022-04-28
Add a link to reference file types table (JP-2533)

dmggh:
Resolves [JP-2533](https://jira.stsci.edu/browse/JP-2533)

**Description**
Add a link in the documentation to the step-specific reference file type table.

Checklist
- [ ] Tests
- [X] Documentation
- [ ] Change log
- [X] Milestone
- [x] Label(s)

### DIFF
--- a/docs/jwst/introduction.rst
+++ b/docs/jwst/introduction.rst
@@ -40,8 +40,8 @@ Information about all the reference files used by the Calibration Pipeline can b
 :ref:`reference_file_information`,
 as well as in the documentation for each Calibration Step that uses a reference file.
 Information on reference file types and their correspondence to calibration steps is
-described within the table for Reference File Types at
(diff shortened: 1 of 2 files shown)


dmggh:
Is a more specific link to the table desired ?

hbushouse (commented):
The link added in line 44 just duplicates the link that already exists in line 40 and hence doesn't really provide any added value. I think what they're looking for is a link to the actual table found in the page at https://jwst-pipeline.readthedocs.io/en/latest/jwst/references_general/references_general.html#reference-file-types. It might be necessary to add a reference tag for the table itself and then make the reference here point directly to that table, so that users don't have to go digging for it from the more general "reference_file_information" page.

hbushouse (approved):

PR #6813, opened 2022-04-28
Updates for source_catalog using photutils 1.4.0

larrybradley:
**Description**

This PR updates the minimum requirement to `photutils 1.4.0` and makes a few improvements to the `source_catalog` step (taking advantage of v1.4.0).

Most of the changes in the PR are simply due to breaking up the original `source_catalog.py` module into 3 separate modules:  one for reference data, one for source detection, and one for the source catalog.  The only substantive changes are:

* The image is now convolved at the start of the step and then this convolved image is used in the detection, deblending, and source catalog functions.  Previously the convolved image was (re-)computed in each step -- this change results in a performance improvement.  The convolution is performed using `astropy` `convolve` with a mask (for `nan` image values or where the image weight is 0) instead of `scipy` `convolve`.  Because of the `mask`, there are some differences in the output source catalog.  The segmentation images are identical.

* The catalog metadata now includes version numbers of all the dependencies.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [ ] Label(s)


larrybradley:
The CI test failure is unrelated.  It appears to be in `test_multichip_jwst_alignment` and caused by a failure to download a file from a remote server.

PR #6818, opened 2022-05-02
added in new param to flag neighboring saturated pixels

cshanahan1:
Add in parameter that controls how many pixels adjacent to a saturated pixel are flagged to account for charge spilling. The default is 1 (i.e, flagging the 8 adjacent pixels).


CHANGES.rst
```diff
@@ -187,6 +187,11 @@ residual_fringe
 
 - Replaced fitting the background with an astropy fitting package [#6739]
 
+saturation
+----------
+
+- New parameter to control how many adjacent-to-saturated pixels are flagged [#6818]
```
nden, line 193:
```suggestion
- New parameter to control how many adjacent-to-saturated pixels are flagged [spacetelescope/stcal#6818]
```

hbushouse (commented):
Also need to modify the saturation_step.py module to add this as a param in the "spec" block of the step class, so that users can set non-default values when desired.

CHANGES.rst
hbushouse: I would modify the text to cover the updates to saturation more generally, e.g. "Updated to allow the step to flag neighbors of saturated pixels, which is controlled by the new step param ``n_pix_grow_sat``."  (i.e. we're talking about actual changes to step algorithm, in addition to the simple addition of a step param)

cshanahan1:
> 

looks like saturation_step doesn't have a 'spec' block at all, ill add one with all the parameters

hbushouse:
> > 
> 
> looks like saturation_step doesn't have a 'spec' block at all, ill add one with all the parameters

Yeah, it didn't have any user-settable params until now. So it'll be a spec block with just this one param.

cshanahan1:
> > > 
> > 
> > 
> > looks like saturation_step doesn't have a 'spec' block at all, ill add one with all the parameters
> 
> Yeah, it didn't have any user-settable params until now. So it'll be a spec block with just this one param.

i deleted my comment when i realized thats what was going on haha

jwst/saturation/saturation_step.py
```diff
@@ -16,6 +16,11 @@ class SaturationStep(Step):
 
     class_alias = "saturation"
 
+    spec = """
+        n_pix_grow_sat = int(default=1) # number of layers adjacent pixels to flag
+    """
```
hbushouse, line 21: Will need to now pass `self.n_pix_grow_sat` as an argument to both `irs2_flag_saturation` and `flag_saturation`, so that it eventually makes its way down to the low-level functions.

jwst/saturation/saturation.py
```diff
@@ -88,13 +94,18 @@ def irs2_flag_saturation(input_model, ref_model):
...
+        charge spilling.
+
     Returns
     -------
     output_model : `~jwst.datamodels.RampModel`
         Data model with saturation, A/D floor, and do not use flags set in
         the GROUPDQ array
     """
-
+    print(n_pix_grow_sat)
```
hbushouse, line 108: Need to remove this print statement.

nden (approved):

hbushouse:
Full regression test run against this PR branch looks good. So I'm merging.

hbushouse (approved):

PR #6820, opened 2022-05-03
Update cube defaults to drizzle, and documentation accordingly

drlaw1558:
Resolves [JP-2590](https://jira.stsci.edu/browse/JP-2590)

**Description**

This PR changes the default cube building weight function from EMSM to 3d drizzle, and updates the documentation accordingly.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/docs/jwst/cube_build/main.rst
+++ b/docs/jwst/cube_build/main.rst
@@ -57,6 +57,20 @@ a single NIRSpec SCA. In high resolution mode the 30 slices are imaged on the tw
 Terminology
 -----------
 
+General IFU Terminology
++++++++++++++++++++
+
+``pixel``
+  A pixel is a physical 2-D element of the detector focal plane arrays.
+
+``spaxel``
+  A spaxel is a 2-D spatial element of an IFU rectified data cube.  Each spaxel in a data cube
+  has an associated spectrum composed of many voxels.
+
+``voxel``
+  A voxel is 3-D volume element within an IFU rectified data cube.  Each voxel has two spatial dimensions and one
+  spectral dimension.
+
 MIRI Spectral Range Divisions
 +++++++++++++++++++++++++++++
 We use the following terminology to define the spectral range divisions of MIRI:
@@ -94,9 +108,6 @@ G395H    F290LP  2.87 - 5.27
 =======  ======  ====================
 
 * Approximate wavelength ranges are given to aid in explaining  how to build NIRSpec IFU cubes, see `NIRSpec Spectral configuration <https://jwst-docs.stsci.edu/jwst-near-infrared-spectrograph/nirspec-observing-modes/nirspec-ifu-spectroscopy#NIRSpecIFUSpectroscopy-Spectralconfigurations>`_.
- 
-
-
 
 Types of Output Cubes
 ---------------------
@@ -157,7 +168,7 @@ WMAP     3      2 spatial and 1 spectral  integer
 =======  =====  ========================  =========
 
 The SCI image contains the surface brightness of cube spaxels in units of MJy/steradian. The wavelength dimension of the IFU cube
-can either be linear or non-linear. If the wavelength is non-linear then the IFU cube contain data from more than one band.  A
+can either be linear or non-linear. If the wavelength is non-linear then the IFU cube contains data from more than one band.  A
 table containing the wavelength of each plane is provided and conforms to the  'WAVE_TAB' fits convention. The wavelengths in the table are read in from the cubepar reference file.  The ERR image contains the
 uncertainty on the SCI values, the DQ image contains the data quality flags for each spaxel, and the WMAP image
 contains the number of point cloud elements contained in the region of interest of the spaxel. The data quality flag does not propagate the
@@ -202,10 +213,10 @@ same resolution are combined together in an IFU cube. The default output spatial
 There is an option to create IFU cubes in the coordinate system of the NIRSpec or MIRI MIRS local ifu slicer plane (see
 :ref:`arguments`, coord_system='internal_cal'). 
 
-The pixels on each exposure that are to be  included in the output are mapped to the cube coordinate system. This input-to-output
+The pixels on each exposure that are to be  included in the output are mapped to the cube coordinate system. This
 pixel mapping is determined via a series of chained mapping transformations derived from the WCS of each input image and the
 WCS of output cube. The mapping process corrects for the optical distortions and uses the spacecraft telemetry information
-to map each pixel location to its projected location in the cube coordinate system.
+to map each pixel to its projected location in the cube coordinate system.
 
 The mapping process results in an irregular spaced "cloud of points" that sample the specific intensity
 distribution at a series of locations on the sky. A schematic of this process is shown
@@ -222,7 +233,7 @@ are from exposure two.
 
 Each point in the cloud represents a measurement of the specific intensity (with corresponding uncertainty)
 of the astronomical scene at a particular location.  The final data cube is constructed by combining each of the
-irregularly-distributed samples of the scene into a regularly-sampled grid in three dimensions for which each
+irregularly-distributed samples of the scene into a regularly-sampled **voxel** grid in three dimensions for which each
 **spaxel** (i.e., a spatial pixel in the cube) has a spectrum composed of many spectral elements.
 
 .. _weighting:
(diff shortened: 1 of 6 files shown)


docs/jwst/cube_build/main.rst
```diff
@@ -57,6 +57,20 @@ a single NIRSpec SCA. In high resolution mode the 30 slices are imaged on the tw
 Terminology
 -----------
 
+General IFU Terminology
++++++++++++++++++++
```
hbushouse, line 61: The line of pluses here needs to be as long as the text above, otherwise sphinx will complain (and give bad results).

docs/jwst/cube_build/main.rst
```diff
@@ -57,6 +57,20 @@ a single NIRSpec SCA. In high resolution mode the 30 slices are imaged on the tw
...
+
+``pixel``
+  A pixel is a physical 2-D element of the detector focal plane arrays.
+
+``spaxel``
+  A spaxel is a 2-D spatial element of an IFU rectified data cube.  Each spaxel in a data cube
+  has an associated spectrum composed of many voxels.
```
hbushouse, line 68: Sorry, but if a spaxel is 2D and a voxel is 3D, I would assume that a voxel is composed of a string of spaxels, where each spaxel contains data at a different wavelength. So how then can a spaxel be composed of many voxels? Seems to me like it would be the other way around. But then I'm not all that familiar with IFU lingo.

docs/jwst/cube_build/main.rst
```diff
@@ -157,7 +168,7 @@ WMAP     3      2 spatial and 1 spectral  integer
 =======  =====  ========================  =========
 
 The SCI image contains the surface brightness of cube spaxels in units of MJy/steradian. The wavelength dimension of the IFU cube
-can either be linear or non-linear. If the wavelength is non-linear then the IFU cube contain data from more than one band.  A
+can either be linear or non-linear. If the wavelength is non-linear then the IFU cube contains data from more than one band.  A
```
hbushouse, line 171: Really minor nit, but shouldn't if-then clauses be punctuated in the middle with a comma, i.e. "If the wavelength is non-linear, then the IFU cube ..."?

jwst/cube_build/cube_build_step.py
```diff
@@ -161,12 +161,19 @@ def process(self, input):
             self.pars_input['output_type'] = 'single'
             self.log.info('Cube Type: Single cubes')
             self.pars_input['coord_system'] = 'skyalign'
-            self.interpolation = 'pointcloud'
 
-            # Don't allow anything but msm or emsm weightings
-            if ((self.weighting != 'msm') and (self.weighting != 'emsm')):
-                self.weighting = 'emsm'
+            # Don't allow anything but drizzle, msm, or emsm weightings
+            if ((self.weighting != 'msm') and (self.weighting != 'emsm') and (self.weighting != 'drizzle')):
```
hbushouse, line 166: Not a big deal, but this line could be made a little cleaner to read using:
```
if self.weighting not in ['msm', 'emsm', 'drizzle']:
```

docs/jwst/cube_build/main.rst
drlaw1558: Thinking of a data cube like a city full of skyscrapers (wavelength in the vertical direction), ’spaxel’ describes a particular city block, while ‘voxel’ describes a specific floor on that block.

docs/jwst/cube_build/main.rst
drlaw1558: derp, fixed.

docs/jwst/cube_build/main.rst
drlaw1558: Not sure, but fixed.

jwst/cube_build/cube_build_step.py
drlaw1558: Done

hbushouse:
The unit test cube_build/tests/test_cube_build_step.py needs updating, due to the change in default weight type. That's what's causing all the CI failures.

hbushouse (approved):
Looks good to me.

hbushouse:
@drlaw1558 @jemorrison we're seeing some hard crashes in cube_build after merging this PR when running against some of our regression tests that use the `--coord_system=internal_cal` option. This includes "test_miri_cubebuild.py::test_cube_build_miri_internal_cal" and "test_nirspec_ifu_internal_cal.py::test_cube_build_nirspec_internal_cal". 

Tracebacks have this:
```
/Users/bushouse/JWST/softw/jwst/jwst/cube_build/cube_build_step.py:329: in process
    cube_result = thiscube.build_ifucube()
/Users/bushouse/JWST/softw/jwst/jwst/cube_build/ifu_cube.py:632: in build_ifucube
    cdelt3_mean = np.nanmean(self.cdelt3_normal)
<__array_function__ internals>:180: in nanmean
    ???
/Users/bushouse/anaconda3/envs/jwstdev/lib/python3.9/site-packages/numpy/lib/nanfunctions.py:1049: in nanmean
    avg = _divide_by_count(tot, cnt, out=out)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

a = None, b = 1, out = None

    def _divide_by_count(a, b, out=None):
        """
        Compute a/b ignoring invalid results. If `a` is an array the division
        is done in place. If `a` is a scalar, then its type is preserved in the
        output. If out is None, then then a is used instead so that the
        division is in place. Note that this is only called with `a` an inexact
        type.
    
        Parameters
        ----------
        a : {ndarray, numpy scalar}
            Numerator. Expected to be of inexact type but not checked.
        b : {ndarray, numpy scalar}
            Denominator.
        out : ndarray, optional
            Alternate output array in which to place the result.  The default
            is ``None``; if provided, it must have the same shape as the
            expected output, but the type will be cast if necessary.
    
        Returns
        -------
        ret : {ndarray, numpy scalar}
            The return value is a/b. If `a` was an ndarray the division is done
            in place. If `a` is a numpy scalar, the division preserves its type.
    
        """
        with np.errstate(invalid='ignore', divide='ignore'):
            if isinstance(a, np.ndarray):
                if out is None:
                    return np.divide(a, b, out=a, casting='unsafe')
                else:
                    return np.divide(a, b, out=out, casting='unsafe')
            else:
                if out is None:
                    # Precaution against reduced object arrays
                    try:
                        return a.dtype.type(a / b)
                    except AttributeError:
>                       return a / b
E                       TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'

/Users/bushouse/anaconda3/envs/jwstdev/lib/python3.9/site-packages/numpy/lib/nanfunctions.py:225: TypeError
=============================================================== short test summary info ===============================================================
FAILED test_nirspec_ifu_internal_cal.py::test_cube_build_nirspec_internal_cal - TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'
```

Could you look into this?

hbushouse:
The regression tests also show that keyword values for ROIS, ROIW, and WPOWER have changed from non-zero to zero in the new outputs. Is that an expected consequence of using the drizzle weighting option?

hbushouse:
Also seeing at least one regression test of the whole calwebb_spec3 pipeline using MRS data where the mrs_imatch step (using the drizzle weighting) says "Background can not be determined, skipping mrs_imatch". Using the "emsm" weighting it succeeds. Have you tried applying mrs_imatch to real data using the drizzle option?  Maybe it's just a problem with our (old) regression test data.

drlaw1558:
The ROIS, ROIW, and WPOWER keywords aren't applicable to drizzle, so it wouldn't surprise me if those have no values.  mrs_imatch works using drizzle, I've tested it and not seen any issues so perhaps it's a reg test issue.  These changes shouldn't have affected internal_cal as that's an entirely different approach; will take a look.

hbushouse:
I tried the "drizzle" weighting in mrs_imatch applied to commissioning observation jw01024-o001, using the 4-dithers of the MIRIFUSHORT and BAND=short data. Despite the fact that these data look fine, I still got the message "Background can not be determined, skipping mrs_imatch". The step code indicates that this happens when any of the background fitting polynomial coeffs come back with a NaN value.

Do you have specific examples of where it does work OK with the drizzle weighting?

drlaw1558:
Ok, I've identified the issue with the hard crash; it's due to the interaction between how self.interpolation self.weighting and self.coord_system all interact with each other.  That's an easy one-line fix.  How should I go about doing so?  A new PR?

mrs_imatch has been completing ok for me on jw01024_o001; investigating that further now.

hbushouse:
Yes, go ahead and submit a new PR for the internal_cal one-line fix.

As for mrs_imatch, it doesn't crash, but if you look at the log from calwebb_spec3 it shows that near the end of mrs_imatch processing (after it's built all the individual cubes and attempted to fit the background in the overlap regions), it comes back with those messages saying that it can't succeed in doing the background subtraction. It then moves on to the remaining pipeline steps.

drlaw1558:
internal_cal crash fixed by https://github.com/spacetelescope/jwst/pull/6826

drlaw1558:
Ok; that was a pretty obvious failure case that I should have seen originally.  The issue with mrs_imatch was that when passed weighting=drizzle for this particular kind of (Single) cube it was still calling the emsm core cube building algorithm.  That of course failed, and returned all-zero cubes which mrs_imatch had no idea what to do with.  My previous testing must have been calling this in a different way so didn't trigger this case.

I've added a drizzle core wrapper call to the Single type cube construction possibilities so that it should pass into the correct type of cube building with the expected arguments (largely just pasting existing code from other places).

See https://github.com/spacetelescope/jwst/pull/6827

Tests well on jw01024_o001

PR #6825, opened 2022-05-04
JP-2593: Add more keywords to flat schema

nden:
Resolves [JP-2593](https://jira.stsci.edu/browse/JP-2593)

**Description**

This PR adds additional keywords to the flat schema to be used in the CRDS selectors

`readpatt`, `preadpatt`, `band` - for use by MIRI

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/datamodels/schemas/flat.schema.yaml
+++ b/jwst/datamodels/schemas/flat.schema.yaml
@@ -7,6 +7,10 @@ allOf:
 - $ref: subarray.schema
 - $ref: keyword_filter.schema
 - $ref: keyword_pfilter.schema
+- $ref: keyword_grating.schema
+- $ref: keyword_pgrating.schema
+- $ref: keyword_readpatt.schema
+- $ref: keyword_preadpatt.schema
 - $ref: keyword_pupil.schema
 - $ref: keyword_ppupil.schema
 - $ref: keyword_channel.schema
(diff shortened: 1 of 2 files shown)


jwst/datamodels/schemas/flat.schema.yaml
```diff
@@ -7,6 +7,10 @@ allOf:
 - $ref: subarray.schema
 - $ref: keyword_filter.schema
 - $ref: keyword_pfilter.schema
+- $ref: keyword_grating.schema
```
hbushouse, line 10: If grating and pgrating are for NIRSpec, I don't think they belong in this schema. NIRSpec uses the dflat, fflat, and sflat schemas, not the generic "flat" schema.

hbushouse: https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/schemas/nirspec_flat.schema.yaml#L8

nden: removed

hbushouse (approved):

jwst/datamodels/schemas/flat.schema.yaml
nden: And added `band` for MIRI as I see it in the selector

PR #6835, opened 2022-05-06
JP-2343: Add CRDS context log message to JwstStep

### LINKED ISSUE #6486: Step/Pipelines should log the CRDS context
_Issue [JP-2343](https://jira.stsci.edu/browse/JP-2343) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Steps/Pipelines should log the CRDS context being used.

Note that there is a verbosity factor involved: Ideally all Steps and Pipelines that use CRDS should log this info. However, when running a Pipeline, that will be quite a bit of the same info being logged, as the Pipeline and all executed Steps will be doing so. Consider methods of mitigation.

tapastro:
Closes #6486
Resolves [JP-2343](https://jira.stsci.edu/browse/JP-2343)

**Description**

This PR adds a log message to report the CRDS context by having JwstPipeline report in finalize_result. It adds the same log in JwstStep finalize_result, but is only provided if the step has no parent, i.e. standalone step running vs. a step within a pipeline.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/stpipe/core.py
+++ b/jwst/stpipe/core.py
@@ -11,7 +11,10 @@ from stpipe import crds_client
 from stpipe import Step
 from stpipe import Pipeline
 from ..lib.suffix import remove_suffix
+import logging
 
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
 
 class JwstStep(Step):
 
@@ -82,6 +85,7 @@ class JwstStep(Step):
                         getattr(result.meta.ref_file, ref_name).name = filename
                 result.meta.ref_file.crds.sw_version = crds_client.get_svn_version()
                 result.meta.ref_file.crds.context_used = crds_client.get_context_used(result.crds_observatory)
+                log.info(f"Results used CRDS context: {result.meta.ref_file.crds.context_used}")
 
     def record_step_status(self, datamodel, cal_step, success=True):
         """Record whether or not a step completed in meta.cal_step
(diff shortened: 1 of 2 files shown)


jwst/stpipe/core.py
```diff
@@ -11,7 +11,10 @@
 from stpipe import Step
 from stpipe import Pipeline
 from ..lib.suffix import remove_suffix
+import logging
 
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
```
stscieisenhamer, line 17: Why is the level being set?

stscieisenhamer (commented):
Besides changelog, LGTM.

Beautiful!

jwst/stpipe/core.py
tapastro: Good question - it was a brainless copy-paste from a different step, but I don't think it's needed anywhere. It does exist in nearly every python file that gets imported into a _step.py file, though. Could be something to cleanse the pipeline of, in another PR.

stscieisenhamer: Ah right. There is actually an issue about this: [JP-1090](https://jira.stsci.edu/browse/JP-1090)

stscieisenhamer (approved):

hbushouse (approved):
LGTM

tapastro:
@cshanahan1 Nadia asked me to ping you on this PR for Roman purposes - I don't know if it'll be helpful or not 😬

hbushouse:
flake8 is not happy with you

PR #6838, opened 2022-05-12
JP-2569: Change fitgeom default value from general to rshift

mcara:
Resolves [JP-2562](https://jira.stsci.edu/browse/JP-2562)
Resolves [JP-2569](https://jira.stsci.edu/browse/JP-2569)

**Description**

At the CalWG meeting it was requested that the default value for ``fitgeometry`` be changed from its current value of `'general'` to `'rshift'`.

I want to note here that the 'general' setting is required for solving exact mathematical equations needed to solve for mis-alignments containing caused by just shifts and/or rotations. However, CalWG believes that keeping the number of degrees of freedom low is more important for the type of alignment that will be encountered in the pipeline. 

Checklist
- [ ] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,11 +16,18 @@ extract_1d
 
 - Fix bug in SOSS algorithm for bad data by replacing source of possible
   infinite values with NaNs, caused by zero division [#6836]
-  
+
 stpipe
 ------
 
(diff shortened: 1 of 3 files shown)


hbushouse (approved):
Looks good

PR #6841, opened 2022-05-12
JP-2596: Eliminate duplicate copies in dark step

hbushouse:
Resolves [JP-2596](https://jira.stsci.edu/browse/JP-2596)

**Description**

This PR removes the copy operation from input to output models in the event the lower-level dark subtraction routines end up skipping processing. A copy is already being made in the lower-level routines, so this eliminates yet another (unnecessary) copy of the whole input data model, thus saving memory.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,11 @@ cube_build
 
 - Fix for residual spectral tearing in MIRI MRS multiband cubes [#6786]
 
+dark_current
+------------
+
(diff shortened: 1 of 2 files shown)


nden (approved):
lgtm

kmacdonald-stsci (commented):
Looks good.

PR #6844, opened 2022-05-16
JP-2610: refpix documentation update

stscirij:
Closes #
Resolves [JP-2610](https://jira.stsci.edu/browse/JP-2610)

**Description**

Minor improvement to the refpix documentation.  State explicitly in the step descriptions that the odd_even_rows parameter is ignored for NIR data and odd_even_columns is ignored for MIRI data

Checklist
- [ ] Tests
- [x] Documentation
- [x] Change log
- [ ] Milestone
- [x] Label(s)


hbushouse (commented):
Documentation updates looks fine, but I wonder if we could carry this even further into the code to make it less confusing when the step is running. For example, the block of code starting at https://github.com/spacetelescope/jwst/blob/master/jwst/refpix/refpix_step.py#L62 could be modified to only print the values of params that actually apply to the type of exposure being processed.

nden:
Something is wrong with this PR. A rebase may help.

stscirij:
Withdrawn in favor #6872

PR #6815, opened 2022-05-01
Restore multiproc for jump

mwregan2:
I don't know if there is a JIRA ticket for this. This PR restores the multiprocessing capability to the Jump step. This can significantly reduce the time for the jump step to run. Note that there is a parallel PR in STCAL that needs to be done at the same time for this to work.

Closes #
Resolves [JP-nnnn](https://jira.stsci.edu/browse/JP-nnnn)

**Description**

This PR addresses ...

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [ ] Label(s)

PR #6845, opened 2022-05-16
JP-2581: allow multiprocessing for jump detection

cshanahan1:
Re-enable multiprocessing for jump detection. 

required for https://github.com/spacetelescope/stcal/pull/87


hbushouse (approved):
Looks good. 

I seem to recall, however, that the code over in stcal allows for integer values of `max_cores`, so I'm just curious why all of these tests needed to have the value changed from 1 to `none`?

cshanahan1:
> Looks good.
> 
> I seem to recall, however, that the code over in stcal allows for integer values of `max_cores`, so I'm just curious why all of these tests needed to have the value changed from 1 to `none`?

I changed it from allowing integer values to just allowing strings. JumpStep spec only allows this to be strings so I just kept that there to avoid confusion.

hbushouse:
The CI tests all passed, when using your stcal branch. So that can be changed back now and we'll go ahead and merge.

By the way, how does this relate to PR #6815 ? Looks like it duplicates the code updates, but diverges in terms of unit tests.

cshanahan1:
I just updated the existing unit tests here, it looks like that PR adds a new one.

hbushouse:
> > Looks good.
> > I seem to recall, however, that the code over in stcal allows for integer values of `max_cores`, so I'm just curious why all of these tests needed to have the value changed from 1 to `none`?
> 
> I changed it from allowing integer values to just allowing strings. JumpStep spec only allows this to be strings so I just kept that there to avoid confusion.

OK, sounds good.

PR #6851, opened 2022-05-19
JP-2613: ZEROFRAME is all zeros after the linearity step

kmacdonald-stsci:
Resolves [JP-2613](https://jira.stsci.edu/browse/JP-2613)

**Description**

The `ZEROFRAME` array is all zeros after the linearity step.  This is caused by an error in calling the `linearity_correction` function in STCAL.  The `zframe` parameter is defaulted to `None`.  The default value was being passed, instead of the `ZEROFRAME` array when present.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,12 @@
 ==================
 
 
+linearity
+---------
+
+- Correcting bug when using ZEROFRAME data. [#6851]
+
+
 1.5.1 (2022-05-17)
 ==================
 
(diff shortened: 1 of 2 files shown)


hbushouse (approved):
LGTM

kmacdonald-stsci:
> CI failure looks like another unrelated glitch?

The failure is in `jwst/assign_wcs/tests/test_miri.py` and appears to be CRDS related.

PR #6854, opened 2022-05-19
JP-2603: SOSS errors for negative infinity, spline fit

tapastro:
Closes #
Resolves [JP-2603](https://jira.stsci.edu/browse/JP-2603)

**Description**

This PR addresses a prior fix that caught infinite values to also catch negative infinity, and it implements a generic try/except to catch errors raised during an optional spline fit to instead use a generic array of estimate values.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 1.5.2 (unreleased)
 ==================
 
+extract_1d
+----------
+
+- In SOSS ATOCA, catch negative infinite values in centroid finder;
+  catch spline-fit errors in first order flux estimate [#6854]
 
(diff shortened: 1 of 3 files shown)


hbushouse (approved):
Looks good.  Though I wonder how many more of these we'll continue to find as all of our bad test data continues to work its way further and further through the step each time you fix something upstream.

tapastro:
The files you've sent so far will all run to completion, so it's just a matter of how many varieties of bad data we have created, I think.

PR #6855, opened 2022-05-20
Use datamodels.open in calwebb_coron3 (JP-2116)

dmggh:
Closes #
Resolves [JP-2116](https://jira.stsci.edu/browse/JP-2116)

**Description**
Use the datamodels.open() method to open a coronagraphic  ASN file in the cal, and extract the 'science' and 'psf' members.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [X] Label(s)


jwst/pipeline/calwebb_coron3.py
```diff
@@ -46,17 +46,25 @@ class Coron3Pipeline(Pipeline):
...
+        """
         self.log.info('Starting calwebb_coron3 ...')
+        asn_exptypes = ['science', 'psf']
 
-        # Load the input association table
-        asn = self.load_as_level3_asn(user_input)
-        acid = asn.get('asn_id', '')
+        # Create a DM object using the association table
+        input_models = datamodels.open(user_input, asn_exptypes=asn_exptypes)
+        acid = input_models.meta.__dict__['_instance']['asn_table']['asn_id']
```
stscieisenhamer, line 63: To avoid the `__dict__`, should be able to do something similar as in line 66 below

`acid = input_models.meta.asn_table.asn_id`

jwst/pipeline/calwebb_coron3.py
```diff
@@ -158,17 +166,22 @@ def process(self, user_input):
...
+
             completed = 'SKIPPED'
         if completed == 'COMPLETE':
             self.log.debug(f'Blending metadata for {result}')
             blendmeta.blendmodels(result, inputs=targ_files)
 
         try:
-            result.meta.asn.pool_name = asn['asn_pool']
+            result.meta.asn.pool_name = \
+                input_models.meta.\
```
stscieisenhamer, line 179: As in line 66 above, should be able to do

`result.meta.asn.pool_name = input_models.meta.asn_table.asn_pool`

stscieisenhamer (commented):
For examples of how to use ModelContainer, and getting pool information, see `calwebb_tso3.py`.

nden:
The failure in the tests looks significant. I don't think it's caused by this PR, it's just the first time we see it (I think). And it may be caused by a new astropy release.

dmggh:
From the "SDP dependencies in requirements-sdp.txt" CI log failure,  the version used is 5.1, which was released 5/24/22 - yesterday !

stscieisenhamer:
The one CI-failure is not relevant to this PR and can be ignored.

tapastro:
The new test is failing in (I believe) the expected fashion, as it's missing the recent PR to stdatamodels. Still working on documentation updates...

jwst/pipeline/calwebb_coron3.py
```diff
@@ -46,17 +46,25 @@ class Coron3Pipeline(Pipeline):
...
-        asn = self.load_as_level3_asn(user_input)
-        acid = asn.get('asn_id', '')
+        # Create a DM object using the association table
+        input_models = datamodels.open(user_input, asn_exptypes=asn_exptypes)
+        acid = input_models.meta.asn_table.asn_id
 
         # We assume there's one final product defined by the association
-        prod = asn['products'][0]
-        self.output_file = prod.get('name', self.output_file)
+        prod = input_models.meta.asn_table.products[0].instance
```
stscieisenhamer, line 66: Is the `.instance` needed here? It may be but I am surprised, since at this point `ans_table` should simply be pointing to the astropy table.

hbushouse: That's the way it's also done in calwebb_spec3:
https://github.com/spacetelescope/jwst/blame/master/jwst/pipeline/calwebb_spec3.py#L117
but still not sure that proves it's really necessary. Blame suggests that we ask @jdavies-st

hbushouse (commented):
These updates look good in principal. I'm thinking we could take this even farther and update the way we handle the "science" vs. "psf" members of the input ASN. For example, the logic starting at line 80 to construct the lists "psf_files" and "targ_files" could also be updated to be done the same way the calwebb_spec3 pipeline constructs and uses lists of "science" and "background" members (see https://github.com/spacetelescope/jwst/blob/master/jwst/pipeline/calwebb_spec3.py#L118 and following).

jwst/pipeline/calwebb_coron3.py
```diff
@@ -46,17 +46,25 @@ class Coron3Pipeline(Pipeline):
...
+        self.output_file = input_models.meta.asn_table.products[0].name
```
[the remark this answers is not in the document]
dmggh: There is :
        input_models.meta.asn_table.products[0].name

   but not :
       input_models.meta.asn_table.products[0].instance.name

    . . .  so we can't get `name` directly from `prod` that way, since
     prod = input_models.meta.asn_table.products[0].instance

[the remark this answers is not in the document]
stscieisenhamer: Hmm, `asn_table` should be the association structure. So, the whole specification to the name should be
```
prod = input_models.meta.asn_table.products[0]
self.output_file = prod['name']
```

As mentioned previously, `ModelContainer.meta.asn_table` is just the association `dict` structure, as described [in the docs](https://jwst-pipeline.readthedocs.io/en/latest/jwst/associations/level3_asn_technical.html#logical-structure).

@hbushouse mentions that the `instance` is also done in Spec3Pipeline, which should also be re-examined. Looking at `ModelContainer`, `asn_table` is just the `dict` structure.

stscieisenhamer (changes requested):

PR #6861, opened 2022-05-24
JP-2587 Check spec2 data is IFUImageModel for straylight step 

jemorrison:
Resolves [JP-2587](https://jira.stsci.edu/browse/JP-2587)

**Description**

This PR checks that the input data to the straylight step has a data model type = IFUImageModel. If it does not, then the step is skipped. 

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [X] Milestone
- [X] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 1.5.3 (unreleased)
 ==================
 
+straylight
+----------
+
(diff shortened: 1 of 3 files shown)


jemorrison:
How do I get it to pass so I can merge it ?

PR #6870, opened 2022-06-06
JP-2624: Update ami_average for different size inputs

hbushouse:
Resolves [JP-2624](https://jira.stsci.edu/browse/JP-2624)

**Description**

This PR updates the ami_average step to be able to handle inputs coming from the ami_analyze step that have different sizes (from exposure to exposure) for the `fit_image` and `resid_image` images. It gets around this by averaging to the minimum size of all the inputs, trimming extra rows/cols from inputs that are larger. Because they all have the target at the center of each image, this still keeps the target aligned in the averaging process, because the images are trimmed symmetrically from all edges.

Also updated the ami_analyze step to only include the primary header when it calls `model.update()` on the output image, so that the outputs don't end up with empty SCI extensions. The SCI extensions don't belong in this kind of product.

Checklist
- [x] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,19 @@
 1.5.3 (unreleased)
 ==================
 
+ami_analyze
+-----------
+
+- Fixed the creation of the output product so that it no longer contains
+  an empty "SCI" extension. [#6870]
+
+ami_average
+-----------
+
+- Updated the step to handle inputs with different sizes for `fit_image` and
+  `resid_image`. Larger inputs are trimmed to match the size of the smallest
+  input. [#6870]
+
 documentation
 -------------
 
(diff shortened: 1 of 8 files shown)


tapastro (approved):
The code looks good. I don't know much about this, but are the ami input images always going to be odd?

hbushouse:
> The code looks good. I don't know much about this, but are the ami input images always going to be odd?

These images are created by the ami_analyze step and are a modeled fit to the actual input images. The way they are created in the ami_analyze step they will always have odd dimensions (because they're created with dimensions of 2*n+1, where n is the distance of the peak from the nearest image edge). So the images from different exposures will always vary in size by multiples of 2 pixels (e.g. 65, 67, 69, ...), hence making it easy to strip off the same number of pixels from each edge to make them the same size.

PR #6871, opened 2022-06-07
xfail NIRCam grism tests until ref files are fixed

hbushouse:
**Description**

This PR temporarily xfails a couple of unit tests in assign_wcs, which are failing (different results) due to new NIRCam specwcs ref files installed in CRDS recently. They will soon be replaced by yet newer versions. So xfail the tests for now, so that CI tests can pass again.

Checklist
- [x] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/assign_wcs/tests/test_nircam.py
+++ b/jwst/assign_wcs/tests/test_nircam.py
@@ -190,12 +190,14 @@ def traverse_wfss_trace(pupil):
     assert orderdet == orderin
 
 
+@pytest.mark.xfail
 def test_traverse_wfss_grisms():
     """Check the dispersion polynomials for each grism."""
     for pupil in ['GRISMR', 'GRISMC']:
         traverse_wfss_trace(pupil)
 
 
+@pytest.mark.xfail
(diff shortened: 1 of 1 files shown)


jwst/assign_wcs/tests/test_nircam.py
```diff
@@ -190,12 +190,14 @@ def traverse_wfss_trace(pupil):
     assert orderdet == orderin
 
 
+@pytest.mark.xfail
```
stscieisenhamer, line 193: A useful feature of `xfail` is to give a reason. This will be helpful to indicate why and when the `xfail` can be removed.
```
xfail(reason='Fails due to JP-XXXX')
```

stscieisenhamer (approved):
Besides suggestion, LGTM

jwst/assign_wcs/tests/test_nircam.py
hbushouse: Updated in latest commit.

PR #6874, opened 2022-06-08
Reduce memory usage in skymatch

mcara:
**Description**

This PR reduces memory usage when running `skymatch` step when input to the step is an ASN table. This is accomplished by saving large image and mask arrays to temporary files and retrieving them on demand. This PR will have no effect when input to the step is a list of already loaded into memory data models.

NOTE: Modifications to `jwst/datamodels/container.py` were designed by @stsci-hack who provided the file to me. I made minor adjustments, I think to `models_grouped`.

Checklist
- [ ] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -19,6 +19,8 @@ __doctest_skip__ = ['ModelContainer']
 
 __all__ = ['ModelContainer']
 
+_ONE_MB = 1 << 20
+
 # Configure logging
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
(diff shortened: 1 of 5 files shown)


jwst/datamodels/container.py
```diff
@@ -445,6 +467,77 @@ def ind_asn_type(self, asn_exptype):
...
+            data_item_size = model.data.itemsize
+            data_item_type = model.data.dtype
+            model.close()
+        del model
+
+        section_nrows = min(imrows, int(self.buffer_size / (imcols * data_item_size)))
+
+        if section_nrows == 0:
+            self.buffer_size = imcols * data_item_size
+            print("WARNING: Buffer size is too small to hold a single row.\n"
```
hbushouse, line 499: Don't want print statements in production code. Change to log message.

hbushouse (commented):
Looks OK to me. Just one trivial comment. Would be good to have @stsci-hack review for compatibility with his work on changing datamodels for use in outlier_detection.

jwst/datamodels/container.py
mcara: Looking at the code for this function, I think it needs serious revision. I'll do my best but @stsci-hack needs to review this. In principle he can fix the code later when working on `outlier_detection`. I am not using this part of the code.

mcara: @hbushouse @stsci-hack done (fixed `set_buffer` and `print()`->logging).

stsci-hack (commented):
I agree that print statements should not be used and support changing those calls to logging calls in this version of the updated module.  These changes can be ported over to the outlier-detection update as soon as the code meets the needs of this PR.

PR #6875, opened 2022-06-08
JP-2637: Only load science members in calwebb_image3

hbushouse:
Resolves [JP-2637](https://jira.stsci.edu/browse/JP-2637)

**Description**

This PR fixes a problem inadvertently implemented by a well-meaning, but misguided developer a couple years ago, in which the calwebb_image3 pipeline module was updated to not blindly load all members from the input ASN, because the presence of target-acq exposures caused problems (see #4937). The update copied what was done in calwebb_spec3, which loads only science+background members. But while spec3 needs background members, image3 does NOT. This is causing problems when calwebb_image3 tries to create mosaics (either in outlier_detection or resample) that contain science and bkg observations that are great distances from one another on the sky.

This update further limits the calwebb_image3 pipeline to only load science members, period. Nothing else. 

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,12 +20,16 @@ documentation
 - Updated the docs for ``calwebb_detector1`` pipeline, as well as the
   ``linearity``, ``refpix``, ``ramp_fit``, ``saturation``, and ``superbias``
   steps to include information on the handling of NIRCam "Frame 0" data.
+  [#6868]
 
 pipeline
 --------
 
 - Add check to ensure SOSS ``extract_1d`` return is not None, to
-  avoid photom errors in Spec3Pipeline and Tso3Pipeline [#6863]
+  avoid photom errors in ``Spec3Pipeline`` and ``Tso3Pipeline``. [#6863]
+
+- Updated the ``calwebb_image3`` pipeline to only science members from the
+  input ASN table. [#6875]
 
 resample
 --------


stscieisenhamer (approved):

tapastro:
I saw a discussion you had on the ops slack channel regarding 
"they're all from an asn ID labeled as background, which means they would've been used individually to subtract background from their corresponding science target images during calwebb_image2, but now it's just creating a mosaic of the background images by themselves, because there might be something interesting in them."

- does this change mean that a background mosaic will now _not_ be constructed by ops? I don't know what the desired configuration/output of image3 is for this kind of input, but I am curious if this could be defined.

stscieisenhamer:
If there is a desire to treat-backgrounds-as-science-and-make-l3-products, there solution would be a rule to create associations where the backgrounds are actually marked as "science". I do not remember if there was a request for this or not.

Regardless, the current functionality of `calwebb_image3` does not apply backgrounds to science hence `calwebb_image3` should not even be loading them. Also need to investigate why the backgrounds are even in the image3 associations; unless there was a request to add them as informational, there is no reason for them to be there.

hbushouse:
> If there is a desire to treat-backgrounds-as-science-and-make-l3-products, there solution would be a rule to create associations where the backgrounds are actually marked as "science". I do not remember if there was a request for this or not.
> 
> Regardless, the current functionality of `calwebb_image3` does not apply backgrounds to science hence `calwebb_image3` should not even be loading them. Also need to investigate why the backgrounds are even in the image3 associations; unless there was a request to add them as informational, there is no reason for them to be there.

I thought we already had an ASN rule in place to create image3 asn's for background only exposures. If we don't, we should, because I'm sure that was requested way back.

stscieisenhamer:
I do not see a rule to create background-as-science-image3 associations. This will need an issue.

Just investigated why there are background members and best as I can see, there should not be. Will take this back to the originating issue and figure that out.

hbushouse:
Regression test run against this PR branch is clean, so I'm merging.

hbushouse:
Updated the image3 ASN file used in the current NIRCam calwebb_image3 regression test to include several (non-existent) background members, to verify that this change stays in effect in the future.

PR #6878, opened 2022-06-08
JP-2632: Change suffix for Lv3_Image members with imaging exp_types and background member_exptype

tapastro:
Closes #
Resolves [JP-2632](https://jira.stsci.edu/browse/JP-2632)

**Description**

This PR adds a check to exp_type before assigning a suffix to an association member with the member_exptype of 'background'. If the exp_type is in the level2 imaging exptypes, use the `cal` suffix rather than `x1d`.

@hbushouse 's recent work on JP-2637 makes me wonder if, rather than fix the suffix, should this PR instead be removing image3 members with member_exptype of background? Or is it fine to leave the exclusion to the pipeline code?

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


hbushouse:
While this change may fix the situation originally reported in JP-2632, I'm also thinking that imaging type ASN's shouldn't have (or at least don't need) background members included at all. I don't recall a directive or request to include them as information only, as is the case for target-acqs. Given that the existing logic that needed fixing was totally devoted to backgrounds used for spectroscopic processing, I'm thinking that's fairly hard evidence that backgrounds were never intended for level-3 imaging asn's. 

So I vote that we modify this to further modify the rules to not include backgrounds in the first place. The only time backgrounds should appear in a level-3 imaging ASN is when we want to create a level-3 product out of the backgrounds themselves (i.e. treat them as science, by themselves).

stscieisenhamer (changes requested):

jwst/associations/lib/rules_level2b.py
```diff
@@ -69,11 +69,20 @@ def __init__(self, *args, **kwargs):
             Constraint_Base(),
             Constraint_Mode(),
             Constraint_Image_Science(),
-            Constraint_Single_Science(self.has_science),
             Constraint(
                 [Constraint_TSO()],
                 reduce=Constraint.notany
-            )
+            ),
```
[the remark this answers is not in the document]
stscieisenhamer: I've looked into this further, and it is a bug in `Constraint_Single_Science`. However, fixing that would require a bit of refactoring of a couple of rules.

Since this mimics what is happening for spectra, I say let's go with this. An issue will be filed for the constraint bug.

stscieisenhamer:
The Level3 background/x1d issue is not new. See #5781

Can only guess that this is somehow related to spectra masterbackground work. Otherwise, looking back through history there is no indication that backgrounds have to be in Level 3 associations.

And, again just self-confirming that the foggy memories do have some basis in fact, like fairly tales, backgrounds pre-masterbackground, were completely removed from Level 3 associations, #2096.

masterbackground fits the name; it has taken over everything.

stscieisenhamer (approved):

jwst/associations/lib/rules_level3.py
```diff
@@ -237,6 +245,53 @@ def _init_hook(self, item):
...
+        # Setup constraints
+        self.constraints = Constraint([
+            Constraint_Optical_Path(),
+            Constraint_Target(association=self),
+            Constraint_Image(),
+            DMSAttrConstraint(
+                name='wfsvisit',
+                sources=['visitype'],
+                value='((?!wfsc).)*',
+                required=False
```
hbushouse, line 270: Does the `required=False` turn this into a constraint that excludes WFS&C exposures from this type of ASN? If so, it's not obvious (at least to me) from the naming/syntax. To me saying that something is not required is equivalent to saying it's optional and hence could be included, but doesn't need to be.

hbushouse (approved):
I *think* this all looks OK, modulo my one question. The real proof of course is in what it actually produces when applied to a pool.

Oh, needs a change log entry.

jwst/associations/lib/rules_level3.py
tapastro: I'll wait with you to hear from Jonathan on the reality of this, but my understanding is that the required check is to prevent member candidates without an entry from causing problems. The constraint on value will exclude the WFS&C entries because they should always contain a value for `wfsvisit`.

hbushouse: Oh yeah, I didn't notice (till now) the presence of the "!" in front of "wfsc", hence it is excluding wfsc exposures. Never mind!

stscieisenhamer: FYI: The `required` option, which yes there needs to be documentation on this, refers to whether the sources are actually available.

Using the current rule as an example: If the column `visitype` was not present in the pool or the value is a non-value, such as `null`, the constraint could not be checked. The default, `required=True`, says the column has to exist, and the constraint would fail. However, with `required=False`, the constraint assumes truth.

Two reasons for this, the original one being that different incantations of pools generated by SDP rapidly changed what columns were in them during development. This allowed using older pools.

The second reason is that what columns are filled in is mode-dependent. If a column is not relevant for a particular mode, a "non-value" is assigned. Depending on the original source of each column, the non-value could be any number of things, such as `null`, `None`, etc. This option allows a constraint to be defined that may cover a number of different modes and ignore sometimes-defined columns amongst those modes. This happens most often when a constraint covers something like "all imaging exposure types among all instruments", but a related constraint may be applied to a column that is only defined for a particular instrument in imaging mode.

PR #6882, opened 2022-06-13
JP-2016: Flat_field step should ignore flatfield pixels with NO_FLAT_FIELD DQ bit set

stscirij:
Closes #
Resolves [JP-2016](https://jira.stsci.edu/browse/JP-2016)

**Description**

Set the DO_NOT_USE DQ bit in the flatfield if the NO_FLAT_FIELD dq bit is set.  This will ensure the flatfield value is set to 1.0 at those pixels.

Code now agrees with the documentation

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


hbushouse (approved):
LGTM

PR #6886, opened 2022-06-15
JP-2626: Add check to all lv3_asns to prevent single-obs group candidates

tapastro:
Closes #
Resolves [JP-2626](https://jira.stsci.edu/browse/JP-2626)

**Description**

This PR adds a finalization check to level 3 association candidates that prevents the creation of any `group` association candidate with members from only one group.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


tapastro:
Latest commit to fix last regtest error from Jenkins runs, should be in a workable state now.

jwst/associations/lib/rules_level3_base.py
```diff
@@ -478,6 +478,21 @@ def ok_candidate(self, member=None):
         """
         return self.acid.type.lower() not in INVALID_AC_TYPES
 
+    def finalize(self):
+        if self.acid.type.lower() == 'group':
+            if len(self.current_product['members']):
+                try:
+                    # Janky extraction of obs_num from filenames
```
hbushouse, line 485: "Janky"  Is that a technical term?

hbushouse (approved):
Looks OK to me, but I'm no asn rule expert.

hbushouse:
py3.10 CI failure is just another connection problem retrieving files from crds.

jwst/associations/lib/rules_level3_base.py
```diff
@@ -478,6 +478,21 @@ def ok_candidate(self, member=None):
         """
         return self.acid.type.lower() not in INVALID_AC_TYPES
 
+    def finalize(self):
+        if self.acid.type.lower() == 'group':
```
stscieisenhamer, line 482: This is possibly too restrictive? In fact, I wonder if it should be the other way: `acid.type != 'observation'`. All other candidate types should be multi-observation, though coronagraphic/ami should be reviewed. There is a possibility that those modes could be still valid for single observations.

jwst/associations/lib/rules_level3_base.py
stscieisenhamer: I like "janky".

However, there is a more "association-like" way of doing this. There is an "informational" constraint, called `Constraint_Obsnum`, that is part of the `AsnMixin_Science` class. The whole point of this constraint is to just gather all the observation numbers that are part of the association.

The multi-observation check can then be to ensure that the list of `self.constraints.['obs_num'].found_values` is greater than one.

stscieisenhamer: This will take away the filename dependency, and make it contolled by the input association pool.

stscieisenhamer (changes requested):

jwst/associations/lib/rules_level3_base.py
stscieisenhamer: Some further thoughts:
- This method should be moved to `Asn_MixinScience`
- Ensure that parent class `finalize` is also called, adding the `super` call at the end of the method.

PR #6895, opened 2022-06-21
JP-2651: Fully Saturated Exposures To Return NoneType

kmacdonald-stsci:
Closes #
Resolves [JP-2651](https://jira.stsci.edu/browse/JP-2651)

**Description**

This PR addresses crashed for fully saturated exposures because the variable `int_model` never gets created due to ramp fitting returning `NoneType` for integration information.  For fully saturated exposures, all return variables are now `NoneType` and the ramp fit step returns `NoneType` for both return variables.  This needed to be properly handled by the pipeline, since datamodels were expected for each variable.

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/pipeline/calwebb_detector1.py
+++ b/jwst/pipeline/calwebb_detector1.py
@@ -134,7 +134,6 @@ class Detector1Pipeline(Pipeline):
             input, ints_model = self.ramp_fit(input)
 
         # apply the gain_scale step to the exposure-level product
-        # XXX JP-2651
         if input is not None:
             self.gain_scale.suffix = 'gain_scale'
             input = self.gain_scale(input)
@@ -156,7 +155,6 @@ class Detector1Pipeline(Pipeline):
         return input
 
     def setup_output(self, input):
-        # XXX JP-2651
         if input is None:
             return None
         # Determine the proper file name suffix to use later


jwst/pipeline/calwebb_detector1.py
```diff
@@ -134,8 +134,12 @@ def process(self, input):
...
 
         # apply the gain_scale step to the exposure-level product
-        self.gain_scale.suffix = 'gain_scale'
-        input = self.gain_scale(input)
+        # XXX JP-2651
+        if input is not None:
+            self.gain_scale.suffix = 'gain_scale'
+            input = self.gain_scale(input)
+        else:
+            log.info("NoneType returned from ramp_fit")
```
hbushouse, line 142: The message could be expanded just a bit here to also mention that the `gain_scale` step will be skipped.

kmacdonald-stsci:
@nden I will check to see if romancal needs a similar PR.

hbushouse (approved):

hbushouse:
Still needs a change log entry.

hbushouse:
Everything, including tests, looks good now. Merging.

PR #6896, opened 2022-06-21
JP-2562: Skip tweakreg step when multiple matches to a single reference source are detected

mcara:
**Description**

Skip tweakreg step when multiple matches to a single reference source are detected. This PR would reduce number of crashes even when match parameters are sub-optimal by skipping `tweakreg` step altogether. See [JP-2562](https://jira.stsci.edu/browse/JP-2562).  Also addresses issues listed in [DMSOPS-445](https://jira.stsci.edu/browse/DMSOPS-445).

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 1.5.4 (unreleased)
 ==================
 
+tweakreg
+--------
+
+Added check for multiple matches to a single reference source and skip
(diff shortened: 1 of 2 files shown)


hbushouse (approved):
Looks good.

PR #6901, opened 2022-06-24
JP-2611: Update saturation handling of NO_SAT_CHECK

hbushouse:
Resolves [JP-2611](https://jira.stsci.edu/browse/JP-2611)

**Description**

This PR updates the NIRSpec IRS2 branch for handling pixels flagged as NO_SAT_CHECK in the saturation reference file DQ array or have thresholds of NaN in the saturation reference file. These pixels used to have their internal thresholds set to the A-to-D limit of 65535, but some pixels can attain that value and hence would get flagged when we don't want them to be. So now the internal threshold is set above the A-to-D limit, ensuring that they'll never get flagged (because no pixels can ever have values above the 65535 limit). Also updated the documentation to reflect this change.

The corresponding change for non-IRS2 data has been made in the `stcal` package in https://github.com/spacetelescope/stcal/pull/106

Checklist
- [ ] ~Tests~  (unit tests have been updated in the stcal package)
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 1.5.4 (unreleased)
 ==================
 
+saturation
+----------
+
+- Updated to set the internal threshold for NO_SAT_CHECK and NaN pixels above the
+  A-to-D limit, so that they never get flagged as saturated. [#6901]
+
 skymatch
 --------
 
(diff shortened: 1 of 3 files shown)


tapastro (approved):
Looks good to me.

cshanahan1 (approved):
looks good

PR #6904, opened 2022-06-28
JP-2527: Improve memory usage in outlier_detection and resample

stsci-hack:
Closes #
Resolves [JP-2527](https://jira.stsci.edu/browse/JP-2527) <Implement memory-saving improvements in outlier_detection>

**Description**

The changes included here update the **outlier_detection** and **resample** steps in order to minimize the memory usage by processing the **ImageModels** on disk instead of memory.  This required updates to the **ModelContainer** class in datamodels to support **ImageModels** on disk rather than keeping them all open in memory.  The initial updates to the **ModelContainer** were implemented in #6874.  The **outlier_detection** step was also modified to compute the median image section-by-section as well.  Finally, new parameters were added to the **outlier_detection** step to control whether or not to process inputs in memory as originally delivered for pipeline use or to work with inputs written out to disk to minimize memory use.  

All these changes resulted in a reduction of memory use during the on-disk processing of the 200 exposure dataset specified in [JP-2527](https://jira.stsci.edu/browse/JP-2527) to allow it to complete processing while using less that 30Gb active memory (according to 'top').  

A new test was also added to the **oulier_detection** step tests to cover the case of working with inputs that are 
on disk instead of in memory.  

Checklist
- [x] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/datamodels/container.py
+++ b/jwst/datamodels/container.py
@@ -495,7 +495,7 @@ class ModelContainer(JwstDataModel, Sequence):
 
         self.buffer_size = min_buffer_size if buffer_size is None else (buffer_size * _ONE_MB)
 
-        section_nrows = min(imrows, self.buffer_size // min_buffer_size)
+        section_nrows = min(imrows, int(self.buffer_size // min_buffer_size))
 
         if section_nrows == 0:
             self.buffer_size = min_buffer_size
--- a/jwst/outlier_detection/outlier_detection.py
+++ b/jwst/outlier_detection/outlier_detection.py
@@ -3,6 +3,7 @@
 from functools import partial
 import logging
 import warnings
+import os
 
 from guppy import hpy
 
@@ -15,6 +16,7 @@ from jwst import datamodels
 from jwst.resample import resample
 from jwst.resample.resample_utils import build_driz_weight, calc_gwcs_pixmap
 from jwst.stpipe import Step
+from jwst.datamodels.util import open as datamodel_open
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
@@ -69,7 +71,6 @@ class OutlierDetection:
         """
         self.inputs = input_models
         self.reffiles = reffiles
-        self.in_memory = pars.get('in_memory', False)
 
         self.outlierpars = {}
         if 'outlierpars' in reffiles:
@@ -185,10 +186,14 @@ class OutlierDetection:
         hp.setrelheap()
         
         self._convert_inputs()
-        self.build_suffix(**self.outlierpars)        
+        self.build_suffix(**self.outlierpars)
 
         pars = self.outlierpars
         save_intermediate_results = pars['save_intermediate_results']
+        h = hp.heap()
+        log.info(f"Initilialized using {h.size / (1024*1024):.3f} Mb")
+        log.info(f"self.input_models._models: {self.input_models._models}")
+        
         if pars['resample_data']:
             # Start by creating resampled/mosaic images for
             # each group of exposures
@@ -196,17 +201,11 @@ class OutlierDetection:
                                            blendheaders=False, **pars)
             drizzled_models = resamp.do_drizzle()
             h = hp.heap()
-            log.info("\n==== Memory Usage ====\n")
-            log.info(f"ResampleData used {h.size} bytes")
-            log.info(f"{h}")
-            if save_intermediate_results:
-                for model in drizzled_models:
-                    log.info("Writing out resampled exposures...")
-                    model_output_path = self.make_output_path(
-                        basepath=model.meta.filename,
-                        suffix='outlier_i2d')
-                    model.save(model_output_path)
-                  
+            log.info(f"RESAMPLING used {h.size / (1024*1024):.3f} Mb")
+
+            h = hp.heap()
+            log.info(f"Results saved using {h.size / (1024*1024):.3f} Mb")
+
         else:
             # for non-dithered data, the resampled image is just the original image
             drizzled_models = self.input_models
@@ -216,47 +215,51 @@ class OutlierDetection:
                     weight_type='ivm',
                     good_bits=pars['good_bits'])
 
-        hp.setrelheap()  # reset to get relative memory use of next steps
+        h = hp.heap()
+        log.info(f"Starting MEDIAN using {h.size / (1024*1024):.3f} Mb")
+        log.info(f"self.input_models._models: {self.input_models._models}")
 
         # Initialize intermediate products used in the outlier detection
-        median_model = datamodels.ImageModel(drizzled_models[0].data.shape)
-        median_model.update(drizzled_models[0])
-        median_model.meta.wcs = drizzled_models[0].meta.wcs
+        dm0 = datamodel_open(drizzled_models[0])
+        median_model = datamodels.ImageModel(dm0.data.shape)
+        # median_model = datamodels.ImageModel(drizzled_models[0].data.shape)
+        median_model.update(dm0)
+        median_model.meta.wcs = dm0.meta.wcs
+        dm0.close()
+        del dm0
 
         # Perform median combination on set of drizzled mosaics
         median_model.data = self.create_median(drizzled_models)
+        median_model_output_path = self.make_output_path(
+            basepath=median_model.meta.filename,
+            suffix='median')
+        median_model.save(median_model_output_path)
         
-        if save_intermediate_results:
-            median_output_path = self.make_output_path(
-                basepath=self.input_models[0].meta.filename,
-                suffix='median'
-            )
-            log.info("Writing out MEDIAN image to: {}".format(
-                median_output_path
-            ))
-            median_model.save(median_output_path)
-
+        h = hp.heap()
+        log.info(f"Finished MEDIAN using {h.size / (1024*1024):.3f} Mb")
+        log.info(f"self.input_models._models: {self.input_models._models}")
+        h = hp.heap()
+        log.info(f"Starting BLOT using {h.size / (1024*1024):.3f} Mb")
         if pars['resample_data']:
             # Blot the median image back to recreate each input image specified
             # in the original input list/ASN/ModelContainer
             blot_models = self.blot_median(median_model)
-            if save_intermediate_results:
-                log.info("Writing out BLOT images...")
-                for model in blot_models:
-                    model_path = self.make_output_path(
-                        basename=model.meta.filename,
-                        suffix='blot'
-                    )
-                    model.save(model_path)
+
+            h = hp.heap()
+            log.info(f"Finished BLOT using {h.size / (1024*1024):.3f} Mb")
+
         else:
             # Median image will serve as blot image
-            blot_models = datamodels.ModelContainer()
+            blot_models = datamodels.ModelContainer(open_models=False)
             for i in range(len(self.input_models)):
                 blot_models.append(median_model)
 
         # Perform outlier detection using statistical comparisons between
         # each original input image and its blotted version of the median image
         self.detect_outliers(blot_models)
+
+        h = hp.heap()
+        log.info(f"Finished DETECT using {h.size / (1024*1024):.3f} Mb")
 
         # clean-up (just to be explicit about being finished with
         # these results)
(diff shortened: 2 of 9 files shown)


hbushouse:
A dozen or so outlier_detection unit tests are failing in CI, many involving unknown key "hpy". Are those expected? Need to fix or update the unit tests.

hbushouse:
Also have flake8/pycodestyle CI failures that need fixing.

jwst/outlier_detection/outlier_detection.py
```diff
@@ -81,6 +82,9 @@ def __init__(self, input_models, reffiles=None, **pars):
             partial(Step._make_output_path, None)
         )
 
+        self.hpy = pars['hpy']
```
hbushouse, line 85: Just curious, what is the "hpy" attribute of pars? I don't see it being used anywhere (but maybe it's just hidden from the diff display).

jwst/outlier_detection/outlier_detection.py
```diff
@@ -206,38 +205,29 @@ def do_detection(self):
...
-            )
-            log.info("Writing out MEDIAN image to: {}".format(
-                median_output_path
-            ))
-            median_model.save(median_output_path)
-
+        median_model_output_path = self.make_output_path(
+            basepath=median_model.meta.filename,
+            suffix='median')
+        median_model.save(median_model_output_path)
```
hbushouse, line 221: So the median image is just saved all the time now, rather than being controlled by a step param? I assume that's necessary in order to keep memory usage down?

jwst/outlier_detection/outlier_detection.py
```diff
@@ -259,45 +249,50 @@ def create_median(self, resampled_models):
...
+                badmasks.append(badmask)
+
+           # Fill resampled_sci array with nan's where mask values are True
+            for f1, f2 in zip(resampled_sci, badmasks):
+                for elem1, elem2 in zip(f1, f2):
+                   elem1[elem2] = np.nan
+           
+            del badmasks
+                
+            # For a of stack of images with "bad" data replaced with Nan
```
hbushouse, line 288: typo: one too many "of"s in here

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -58,17 +58,22 @@ class OutlierDetectionStep(Step):
         snr = string(default='5.0 4.0')
         scale = string(default='1.2 0.7')
         backg = float(default=0.0)
-        save_intermediate_results = boolean(default=False)
+        save_intermediate_results = boolean(default=True)
```
hbushouse, line 61: Is this param even used any more? It looked to me like all the intermediate files are always saved now.

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -58,17 +58,22 @@ class OutlierDetectionStep(Step):
...
         resample_data = boolean(default=True)
         good_bits = string(default="~DO_NOT_USE")  # DQ flags to allow
         scale_detection = boolean(default=False)
         search_output_file = boolean(default=False)
         allowed_memory = float(default=None)  # Fraction of memory to use for the combined image.
+        in_memory = boolean(default=False)
+        output_use_index = boolean(default=False)
+        output_use_model = boolean(default=False)
+        save_results = boolean(default=True)
```
hbushouse, line 70: I believe ``save_results`` is a standard param that's hardwired into every `Step` object, so is it necessary to define it here too?

jwst/resample/resample.py
```diff
@@ -132,15 +134,22 @@ def resample_many_to_many(self):
         Used for outlier detection
         """
         for exposure in self.input_models.models_grouped:
-            output_model = self.blank_output.copy()
+            output_model = self.blank_output  #.copy()
+            output_root = '_'.join(exposure[0].meta.filename.replace(
+                '.fits', '').split('_')[:-1])
+            output_model.meta.filename = f'{output_root}_outlier_i2d.fits'
```
hbushouse, line 140: Not sure we should be hardwiring file names to use the ".fits" extension. Theoretically data models can be serialized to different formats (fits, ASDF, etc.).

hbushouse (commented):
Looks good at a high level (have to confess I haven't verified the accuracy of every single line of code!). Just a few questions here and there, for clarification.

hbushouse:
Still have some flake8/pycodestyle issues to fix.

jwst/outlier_detection/outlier_detection.py
```diff
@@ -307,23 +302,33 @@ def blot_median(self, median_model):
...
-            blotted_median.dq = None
+            blotted_median.err *= 0.0  # None
+            blotted_median.dq *= 0     # None
+
             # apply blot to re-create model.data from median image
             blotted_median.data = gwcs_blot(median_model, model, interp=interp,
                                             sinscl=sinscl)
-            blot_models.append(blotted_median)
+                                            
+            blotted_median.save(model_path)
```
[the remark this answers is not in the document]
stsci-hack: Definitely.  There will be 1 blotted image per input which can also soak up a LOT of memory as well.

jwst/outlier_detection/outlier_detection_step.py
stsci-hack: Probably not.  It has been removed in commit 351c28d5.

jwst/outlier_detection/outlier_detection_step.py
stsci-hack: Removed as part of commit 211f90d5.

jwst/outlier_detection/outlier_detection.py
stsci-hack: Cleaned up as of commit ea67e64.

jwst/outlier_detection/outlier_detection.py
stsci-hack: That is correct.

jwst/outlier_detection/outlier_detection.py
stsci-hack: This was originally added to allow the memory inspection object from 'guppy' to be passed around to provide better stats on the memory usage during processing.  All references to this module were removed, except this overlooked reference.  It has been removed in commit 446410ea.

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -64,11 +64,15 @@ class OutlierDetectionStep(Step):
         scale_detection = boolean(default=False)
         search_output_file = boolean(default=False)
         allowed_memory = float(default=None)  # Fraction of memory to use for the combined image.
+        in_memory = boolean(default=False)
+        output_use_index = boolean(default=False)
+        output_use_model = boolean(default=False)
```
mcara, line 69: I do not believe these three new step parameters are used anywhere in the step. If that is correct, these parameters should be removed.
```suggestion
```

jwst/outlier_detection/outlier_detection.py
```diff
@@ -307,23 +298,33 @@ def blot_median(self, median_model):
...
+
             # apply blot to re-create model.data from median image
             blotted_median.data = gwcs_blot(median_model, model, interp=interp,
                                             sinscl=sinscl)
-            blot_models.append(blotted_median)
+
+            blotted_median.save(model_path)
+            blot_models.append(model_path)
+            blotted_median.close()
+            # blot_models.append(blotted_median)
```
mcara, line 326: Is this needed in the final (production) code?
```suggestion
```

jwst/outlier_detection/outlier_detection.py
```diff
@@ -182,20 +183,14 @@ def do_detection(self):
...
                                            blendheaders=False, **pars)
             drizzled_models = resamp.do_drizzle()
-            if save_intermediate_results:
-                for model in drizzled_models:
-                    log.info("Writing out resampled exposures...")
-                    model_output_path = self.make_output_path(
-                        basepath=model.meta.filename,
-                        suffix='outlier_i2d')
-                    model.save(model_output_path)
+
```
mcara, line 193: I do not understand why `if save_intermediate_results: ...` block was removed. Is saving intermediate results no longer needed?

jwst/outlier_detection/outlier_detection_step.py
hbushouse: I saw in the diffs where the median and blotted images were being automatically saved to disk, but don't recall seeing if the individually resampled inputs are saved. Are they?

jwst/datamodels/container.py
```diff
@@ -495,7 +495,7 @@ def set_buffer(self, buffer_size, overlap=None):
 
         self.buffer_size = min_buffer_size if buffer_size is None else (buffer_size * _ONE_MB)
 
-        section_nrows = min(imrows, self.buffer_size // min_buffer_size)
+        section_nrows = min(imrows, int(self.buffer_size // min_buffer_size))
```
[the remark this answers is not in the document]
hbushouse: Granted, floor division always at least results in a whole number, and if both the inputs to the division are ints the result should also be int, but it doesn't do any harm to make sure the result really is int.

jwst/outlier_detection/outlier_detection_step.py
hbushouse: They may be standard params that are defined in the standard `Step` class and hence you're right that they're not needed here (already available in the step object created for the process).

jwst/outlier_detection/outlier_detection.py
hbushouse: It's my understanding that all intermediate results are now always saved to disk (instead of keeping them in memory only), hence there's no need for the optional param anymore. Not sure what happens if the new param `in_memory` is activated.

hbushouse:
The fact that some of the unit test results have changed worries me a bit. We shouldn't see any change in results from these updates, just changes in file/data handling.

hbushouse:
Just kicked off a full regression test run against this PR branch and already seeing errors in the log output. Can't tell what they are until it finishes and see the full report. Will let you know.

hbushouse:
Full regtest report is available at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/350/testReport/

Some of the "errors" are just the same differences we're seeing here in the unit tests, which in the context of a Jenkins run show up as an error, rather than just a difference. But also seeing other actual errors when applying to images from other instruments (e.g. MIRI). Examples:

```
/data1/jenkins/workspace/RT/JWST-Developers-Pull-Requests/clone/jwst/outlier_detection/outlier_detection.py:213: in do_detection
    median_model.data = self.create_median(drizzled_models)
/data1/jenkins/workspace/RT/JWST-Developers-Pull-Requests/clone/jwst/outlier_detection/outlier_detection.py:256: in create_median
    for (resampled_sci, resampled_weight, (row1,row2)) in resampled_sections:

self = <ModelContainer>

    def get_sections(self):
        """Iterator to return the sections from all members of the container."""
    
>       for k in range(self.n_sections):
E       TypeError: 'float' object cannot be interpreted as an integer
```

and

```
>           if not self.in_memory:
E           AttributeError: 'ResampleSpecData' object has no attribute 'in_memory'
```

mcara:
> The fact that some of the unit test results have changed worries me a bit. We shouldn't see any change in results from these updates, just changes in file/data handling.

I totally agree. I would not expect any changes but if there are _expected_ changes then they should be clearly stated (i.e., this code also fixes a bug; personally I would like then the bug and momory reduction to be split into two separate PRs).

mcara:
> Some of the "errors" are just the same differences we're seeing here in the unit tests, which in the context of a Jenkins run show up as an error, rather than just a difference. But also seeing other actual errors when applying to images from other instruments (e.g. MIRI). Examples:
> 
> ```
>     def get_sections(self):
>         """Iterator to return the sections from all members of the container."""
>     
> >       for k in range(self.n_sections):
> E       TypeError: 'float' object cannot be interpreted as an integer
> ```

Are you saying that even with explicit `int` in code like this:
```
nsec = int((imrows - self.overlap) // nbr)
```
number of section can be a float?

One good reason for removing explicit `int` is that if code is buggy somewhere and produces floats instead of integer, it will be more likely for the code to fail somewhere downstream which is good IMO because it gives a chance to detect the bug. Using `int` unnecessarily can mask a bug.

jwst/outlier_detection/outlier_detection_step.py
stsci-hack: This happens in the new code for 'resample.py' starting at line 168 (see the diffs).

jwst/resample/resample.py
stsci-hack: The `.fits` extension was replaced with the extension from the original input model 'meta.filename' instead in commit 88cc36f.

jwst/datamodels/container.py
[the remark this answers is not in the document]
stsci-hack: The use of 'int()' is not present in the version of 'datamodels.container' actually on master after merging #6874.

jwst/outlier_detection/outlier_detection.py
stsci-hack: This parameter actually controls the behavior of writing out all the final resampled products when they are created. Unfortunately, it was not practical to keep the rest of the **outlier_detection** step compatible with both in-memory and on-disk processing, so this parameter does not get used for the **outlier_detection** products (intermediate or final).   As such, the intermediate results are always written out to disk making the '**save_intermediate_results**' code duplicative during processing.

jwst/outlier_detection/outlier_detection.py
stsci-hack: Removed in [43ea4f9](https://github.com/spacetelescope/jwst/pull/6904/commits/43ea4f99f01bf738e2818f3e1009a1ea56f30e3b)

jwst/outlier_detection/outlier_detection.py
```diff
@@ -206,38 +205,29 @@ def do_detection(self):
                     good_bits=pars['good_bits'])
 
         # Initialize intermediate products used in the outlier detection
-        median_model = datamodels.ImageModel(drizzled_models[0].data.shape)
-        median_model.update(drizzled_models[0])
-        median_model.meta.wcs = drizzled_models[0].meta.wcs
+        dm0 = datamodel_open(drizzled_models[0])
+        median_model = datamodels.ImageModel(dm0.data.shape)
+        # median_model = datamodels.ImageModel(drizzled_models[0].data.shape)
```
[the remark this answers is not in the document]
stsci-hack: Removed in [43ea4f9](https://github.com/spacetelescope/jwst/pull/6904/commits/43ea4f99f01bf738e2818f3e1009a1ea56f30e3b)

jwst/datamodels/container.py
```diff
@@ -504,7 +504,7 @@ def set_buffer(self, buffer_size, overlap=None):
             section_nrows = 1
 
         nbr = section_nrows - self.overlap
-        nsec = (imrows - self.overlap) // nbr
+        nsec = int((imrows - self.overlap) // nbr)
```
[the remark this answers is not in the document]
stsci-hack: Actually, the use of 'int()' here should not be necessary with the fix in [b434083](https://github.com/spacetelescope/jwst/pull/6904/commits/b43408359ce41052d1274a3e19ea7b7b22f1cafc)

jwst/datamodels/container.py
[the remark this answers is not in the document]
stsci-hack: See comment at end for commit [b434083](https://github.com/spacetelescope/jwst/pull/6904/commits/b43408359ce41052d1274a3e19ea7b7b22f1cafc)

stsci-hack:
Summary of responses to review comments:

- **use of 'int()' in `datamodels.container`:**  This actually turns out to be required since the buffer size used for this computation is a float and '3.4 //2' actually returns a floating point value of '1.0'.  Fixed in b434083.
-  **Exception in Jenkins test for resample_spec**: This should be addressed by 7b6122f which defines the missing `in_memory` attribute for the ResampleSpecData class. 
- **Unused new parameters for outlier_detection step**:  They were removed in aad4d58.
- **Unit tests for outlier_detection**:  These tests should now pass as a result of d058ef4.
-

mcara (approved):
LGTM

hbushouse:
Oooh, look at all of those passing CI tests! Going to kick off another regtest run against the PR branch now, to see if yesterday's errors are now fixed.

hbushouse:
The full regression test run against this PR branch no longer shows any errors (yay!), but does show a number of crf and i2d images with a small number of slightly different pixel values (some of which correspond to pixels that either used to be flagged as outlier and now aren't, or vice versa). Not sure why these would change, but it is a very small number of pixels. Perhaps the slight rearrangement in processing has caused some of them to have ever so slightly different values that have now crossed from one side of the detection threshold to the other. I'll do a little more investigating to see if I can spot anything obvious.

hbushouse:
Good news: latest regression test run against the PR branch shows no differences for the MIRI and NIRCam imaging results, and the only differences for NIRSpec spectra are due to the addition of the empty AREA extension in the output files (no differences in pixel values - I verified that).

hbushouse (approved):
Tests are good! I'm approving and merging.

PR #6910, opened 2022-07-05
JP-2633: Add unit to time keyword comment strings

tapastro:
Closes #
Resolves [JP-2633](https://jira.stsci.edu/browse/JP-2633)

**Description**

This PR adds the unit (MJD) to the barycentric and heliocentric start, mid and end time header keyword comment strings.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -1406,17 +1406,17 @@ properties:
             fits_keyword: BARTDELT
             blend_table: True
           barycentric_expstart:
-            title: [d] Barycentric exposure start time in MJD
+            title: "[d] Barycentric exposure start time in MJD"
             type: number
             fits_keyword: BSTRTIME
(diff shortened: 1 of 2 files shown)


jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1406,17 +1406,17 @@ properties:
             fits_keyword: BARTDELT
             blend_table: True
           barycentric_expstart:
-            title: Barycentric exposure start time
+            title: Barycentric exposure start time in MJD
```
hbushouse, line 1409: The latest on-line version of the KWD has the comment field of all these keywords as "[d] Barycentric exposure start time in MJD", so we should make ours match.

hbushouse: No, the "[d]" indicates units of days.

PR #6911, opened 2022-07-05
Update default parameters for source catalog step

larrybradley:
**Description**

This PR updates the default background boxsize and minimum number of pixels in the source catalog step:
 * The background boxsize was increased to prevent oversubtraction of the background in sources smaller than the boxsize.  A larger size gives a more "global" vs. "local" background and prevents oversubtraction.  I recently observed this in simulated ERS (CEERS) data.
 * The number of pixels was increased to prevent the detection of many tiny sources (which may not be real sources).

Ideally the source catalog default parameters would be instrument/detector specific and set by parameter reference files, but I don't think those have been delivered yet by the instrument teams.

Checklist
- [ ] Tests
- [x] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,12 @@ skymatch
 
 - Fixed a couple errors in the step documentation. [#6891]
 
+source_catalog
+--------------
+
+- Updated the default background box size and minimum number of pixels.
+  [#6911]
+
 tweakreg
 --------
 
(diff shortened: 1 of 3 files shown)


hbushouse:
Sometimes the default values of params are listed in the docs for the step. Have you checked those to see if they need updating?

hbushouse (approved):

hbushouse:
All CI failures are due to expected differences from recently updated NIRSpec ref files in CRDS. Unrelated to this PR.

PR #6912, opened 2022-07-06
JP-2667: Fix calwebb_tso3 failure when checking for null results

hbushouse:
Resolves [JP-2667](https://jira.stsci.edu/browse/JP-2667)

**Description**

This PR makes a change to the logic used near the end of the pipeline to check for all null results coming back from either the `tso_photometry` or `white_light` steps. A recent numpy update (1.23.0) changed a FutureWarning to an actual error, related to the behavior of astropy.tables objects. The new, simpler logic works with numpy 1.23 and is backwards compatible with previous versions.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [ ] Label(s)

### DIFF
--- a/jwst/pipeline/calwebb_tso3.py
+++ b/jwst/pipeline/calwebb_tso3.py
@@ -187,7 +187,8 @@ class Tso3Pipeline(Pipeline):
         input_models.close()
 
         # Check for all null photometry results before saving
-        if phot_result_list.count(None) == len(phot_result_list):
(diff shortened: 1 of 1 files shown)


tapastro (approved):
Looks good to me.

hbushouse:
CI failures are due to the recent NIRSpec ref file updates (unrelated).

PR #6913, opened 2022-07-06
XFail NRS unit tests for new ref files

hbushouse:
**Description**

This PR temporarily xfails several NIRSpec WCS unit tests, due to differences from in-flight updates to wcs reference files. Will take some time before we can determine if the new results are valid and update the truth values.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [ ] Label(s)


tapastro (approved):
Looks like the sdp-deps tests failed due to the recurring CRDS connection problem; from the other tests, it looks like you got them all.

PR #6916, opened 2022-07-08
Switch to using external wiimatch package

mcara:
**Description**

This PR removes internal `wiimatch` subpackage and switches to external one: https://github.com/spacetelescope/wiimatch. This is was prompted by work on https://jira.stsci.edu/browse/JP-2607 and it is simply inefficient to maintain two identical packages. `jwst` pipeline should use the stand-alone version.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)


hbushouse (approved):
Looks good to me.

PR #6921, opened 2022-07-13
JP-2580 update loading of drizpars table

jemorrison:
Resolves [JP-2580](https://jira.stsci.edu/browse/JP-2580)

**Description**
The PR updates the resample_step to use the values read in from the drizpars reference file. The parameters in the reference file: pixfrac, weight_type, kernel, and fillval need to have defaults of None in the step spec. If they are not None, then the code assumes that the user has overridden the values.  In addition, the code was expecting pixel_scale_ratio to be in the drizpars reference file, instead of wht_type (weight_type). Since the reference file and data model expect the name of weight_type to be wht_type., self.wht_type was added and set to self.weight_type.

Checklist
- [ ] Tests
- [ ] Documentation
- [x] Change log
- [x] Milestone
- [x] Label(s)

### DIFF
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -39,7 +39,7 @@ class ResampleData:
     """
 
     def __init__(self, input_models, output=None, single=False, blendheaders=True,
-                 pixfrac=1.0, kernel="square", fillval="INDEF", weight_type="ivm",
+                 pixfrac=1.0, kernel="square", fillval="INDEF", wht_type="ivm",
                  good_bits=0, pscale_ratio=1.0, pscale=None, **kwargs):
         """
         Parameters
@@ -65,8 +65,13 @@ class ResampleData:
         self.pixfrac = pixfrac
         self.kernel = kernel
         self.fillval = fillval
-        self.weight_type = weight_type
+        self.weight_type = wht_type
         self.good_bits = good_bits
+
+        log.info(f"Driz parameter kernal: {self.kernel}")
+        log.info(f"Driz parameter pixfrac: {self.pixfrac}")
+        log.info(f"Driz parameter fillval: {self.fillval}")
+        log.info(f"Driz parameter weight_type: {self.weight_type}")
 
         ref_wcs = kwargs.get('ref_wcs', None)
         out_shape = kwargs.get('out_shape', None)
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -68,7 +68,7 @@ class ResampleData:
         self.weight_type = wht_type
         self.good_bits = good_bits
 
-        log.info(f"Driz parameter kernal: {self.kernel}")
+        log.info(f"Driz parameter kernel: {self.kernel}")
         log.info(f"Driz parameter pixfrac: {self.pixfrac}")
         log.info(f"Driz parameter fillval: {self.fillval}")
         log.info(f"Driz parameter weight_type: {self.weight_type}")
--- a/jwst/resample/resample_spec_step.py
+++ b/jwst/resample/resample_spec_step.py
@@ -25,6 +25,7 @@ class ResampleSpecStep(ResampleStep):
     class_alias = "resample_spec"
 
     def process(self, input):
+        self.wht_type = self.weight_type
         input_new = datamodels.open(input)
 
         # Convert ImageModel to SlitModel (needed for MIRI LRS)
@@ -61,11 +62,10 @@ class ResampleSpecStep(ResampleStep):
             kwargs['blendheaders'] = self.blendheaders
 
         kwargs['allowed_memory'] = self.allowed_memory
-        kwargs['weight_type'] = str(self.weight_type)
         kwargs['output'] = output
 
         # Issue a warning about the use of exptime weighting
-        if self.weight_type == 'exptime':
+        if self.wht_type == 'exptime':
             self.log.warning("Use of EXPTIME weighting will result in incorrect")
             self.log.warning("propagated errors in the resampled product")
 
--- a/jwst/resample/resample_step.py
+++ b/jwst/resample/resample_step.py
@@ -36,7 +36,7 @@ class ResampleStep(Step):
         pixfrac = float(default=1.0) # change back to None when drizpar reference files are updated
         kernel = string(default='square') # change back to None when drizpar reference files are updated
         fillval = string(default='INDEF' ) # change back to None when drizpar reference files are updated
-        weight_type = option('ivm', 'exptime', None, default='ivm')  # change back to None when drizpar reference files are updated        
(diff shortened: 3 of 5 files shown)


jemorrison:
There is one remaining issue - a pytest for NIRSPEC fails. I am not sure why:
jwst/resample/tests/test_resample_step.py::test_nirspec_wcs_roundtrip
  /Users/morrison/git/jwst/jwst/resample/resample_spec.py:117: RuntimeWarning: invalid value encountered in double_scalars
    wmean_s = np.sum(sd[good_s]) / total

I will look for the NIRSpec data to run resample_spec with and see if I can find the problem.

jwst/resample/resample.py
```diff
@@ -65,9 +65,14 @@ def __init__(self, input_models, output=None, single=False, blendheaders=True,
         self.pixfrac = pixfrac
         self.kernel = kernel
         self.fillval = fillval
-        self.weight_type = weight_type
+        self.weight_type = wht_type
         self.good_bits = good_bits
 
+        log.info(f"Driz parameter kernal: {self.kernel}")
```
hbushouse, line 71: typo: "kernal" should be "kernel"

Also, are these the values coming from the spec, i.e. before the drizpars ref file has been loaded?

jwst/resample/resample_spec_step.py
```diff
@@ -61,11 +62,10 @@ def process(self, input):
             kwargs['blendheaders'] = self.blendheaders
 
         kwargs['allowed_memory'] = self.allowed_memory
-        kwargs['weight_type'] = str(self.weight_type)
```
hbushouse, line 64: Is this going away because we want it to come from the drizpars file, instead of the step spec?

jwst/resample/resample_step.py
```diff
@@ -140,6 +140,7 @@ def get_drizpars(self, ref_filename, input_models):
         pixfrac = float(default=None)
         kernel = string(default=None)
         fillval = string(default=None)
+        wht_type = option('ivm', 'exptime', None,default=None)
```
hbushouse, line 143: Need whitespace before "default"

hbushouse (commented):
The overall strategy looks good to me. Just a few minor comments and questions.

Oh, needs a change log entry.

jwst/resample/resample.py
jemorrison: I added printing the driz parameters because they can be filled from either the reference file or the driz ref files.
These values are the final ones that are used in resample. So they are after the drizpars ref file has been loaded.

jwst/resample/resample_spec_step.py
jemorrison: I removed that line because now weight_type is set to 'None' in spec and if the user does not set it, then the value in the drizpar reference file is use.  In a fews lines above this: kwargs = self.get_drizpars(ref_filename, input_models)
the kwargs['wht_type'] is filled in.

jwst/resample/resample_step.py
```diff
@@ -33,10 +33,10 @@ class ResampleStep(Step):
...
 
     spec = """
-        pixfrac = float(default=1.0)
-        kernel = string(default='square')
-        fillval = string(default='INDEF')
-        weight_type = option('ivm', 'exptime', default='ivm')
+        pixfrac = float(default=None)
+        kernel = string(default=None)
+        fillval = string(default=None)
+        weight_type = option('ivm', 'exptime', None,default=None)
```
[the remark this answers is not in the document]
jemorrison: I need a whitespace before 'default' ? That is not how the other parameters are set up.

And yes I had to add the extra option of None because when the default is None the step failed because None was not a valid option.

jwst/resample/resample_spec_step.py
hbushouse: That's what I thought. Just wanted to confirm.

jemorrison:
Ok I did not realize the FGS data used resample. I will set up the defaults like NIRSPEC does.

jemorrison:
@hbushouse  The defaults set for NIRSPEC in resample_step.py are not necessarily NIRSPEC specific (just the comments mention NIRSpec). In CRDS the drizpars  file for FGS is  N/A. So it should work for FGS.Could you point me at some FGS data to run resample on and I will see what is going on.


if ref_filename != 'N/A':
            self.log.info('Drizpars reference file: {}'.format(ref_filename))
 else:
            # If there is no drizpars reffile
            self.log.info("No DRIZPARS reffile")

hbushouse:
Yes, theoretically the existing if-block that calls the function to set default values whenever drizpars is N/A should take care of it, but the problem might be due to the fact that that function only populates `kwargs` items, whereas something like `self.pixfrac` is still set to `None` and may try to get used somewhere later on. The test I ran (that crashed) was the test_fgs_image2.py module in jwst/jwst/regtest/. So you could try the files used in that test. You can also run that test to see the error that's occurring.

jemorrison:
@hbushouse  Yes of course the parameters that have None as a default now and do not have a drizpars reference file will still be None. So I set up defaults for these parameters in _set_spec_defaults. I tested it on test_fgs_image2.py and that passed.

hbushouse:
Another minor issue, related to the fact that several of the `self.xxxx` items are now `None` throughout much of the processing, I noticed that the PIXFRAC keyword was missing from the output file (it used to be there, along with a few other resample-related keywords). It's because the meta attribute for it gets set here:
https://github.com/spacetelescope/jwst/blob/master/jwst/resample/resample_step.py#L130
but that references `self.pixfrac` which is still set to `None` even at the end of processing. I wonder if it would be worthwhile, once the `kwargs` have been populated with values from the drizpars table, to then copy those kwargs values back into the various `self.xxxxx` items? At the very least, the storing of those output keyword values needs to be updated to use the values from `kwargs` instead of `self.xxxx`. I would also suggest moving the two lines starting at https://github.com/spacetelescope/jwst/blob/master/jwst/resample/resample_step.py#L129 out of the `update_phot_keywords` functions (because these aren't photometry keywords) and put them instead right before that function is called, e.g. here https://github.com/spacetelescope/jwst/blob/master/jwst/resample/resample_step.py#L114 where other keywords are updated. And for the PIXFRAC keyword the meta population needs to be done using:
`model.meta.resample.pixfrac = kwargs['pixfrac']`, because `self.pixfrac` is still `None`.

hbushouse:
Needs a fix for failing flake8 test.

hbushouse (approved):

mcara:
I do not know many things about reference files but I know that `pixel_scale_ratio` should not have been commented out in https://github.com/jemorrison/jwst/blob/844646c47071683df3428454e295e616a09fab89/jwst/resample/resample_step.py#L187

I did not do a full investigation but I believe this PR has caused the issue described the help desk ticket INC0180832, that is, `pixel_scale_ratio` is ignored altogether.

CC: @cshanahan1

PR #6922, opened 2022-07-13
Fix: skymatch not saving subtracted values for input ASN

### LINKED ISSUE #6920: SkyMatchStep measuring sky background, subtracting but not passing the subtracted data to next step in STAGE3
Problem in 1.5.3, related to rework on the memory usage of SkyMatchStep in #6874. Downgrading to 1.5.2 solves the issue.

Problem: SkyMatchStep measures sky background level, prints it to the screen, subtracts it (in skymatch.py) but then returns the input prior to the subtraction to the next step in STAGE3 pipeline. Final results are not sky subtracted while subtract=True. 

After some debugging I suspect the problem is related to the changes in #6874, specially the return of `process()` in `SkyMatchStep` and how the data is stored throughout that Step class.

line 128 SkyMathStep.py
`
 return input if self._is_asn else img
`

`input` does not hold the updated data sky subtracted, while img does. My understanding is that the else statement here is never used as you always need to run it with an ASN file (at least from command line which is the only way I've been using the pipeline). 

Here's the histogram for the flux of a single _cal.fits file using the same ASN config and SkyMatchStep setup with 1.5.2 and 1.5.3. Only change is the version

image

mcara:
Closes #6920

This PR fixes the bug due to which `skymatch` step was not saving changes to the data models (computed/subtracted sky levels) when inputs are ASN files.

CC: @astroferreira

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/jwst/skymatch/skymatch_step.py
+++ b/jwst/skymatch/skymatch_step.py
@@ -63,7 +63,10 @@ class SkyMatchStep(Step):
 
     def process(self, input):
         self.log.setLevel(logging.DEBUG)
-        self._is_asn = datamodels.util.is_association(input) or isinstance(input, str)
+        # for now turn off memory optimization until we have better machinery
+        # to handle outputs in a consistent way.
+        self._is_asn = False
+        # self._is_asn = datamodels.util.is_association(input) or isinstance(input, str)
 
         img = datamodels.ModelContainer(
             input,
--- a/jwst/skymatch/skymatch_step.py
+++ b/jwst/skymatch/skymatch_step.py
@@ -206,14 +206,18 @@ class SkyMatchStep(Step):
         return sky_im
 
     def _set_sky_background(self, image, sky, step_status):
+
         if self._is_asn:
-            image = datamodel_open(image)
+            dm = datamodel_open(image)
+        else:
+            dm = image
 
         if step_status == "COMPLETE":
-            image.meta.background.method = str(self.skymethod)
-            image.meta.background.level = sky
-            image.meta.background.subtracted = self.subtract
-        image.meta.cal_step.skymatch = step_status
+            dm.meta.background.method = str(self.skymethod)
+            dm.meta.background.level = sky
+            dm.meta.background.subtracted = self.subtract
+        dm.meta.cal_step.skymatch = step_status
 
         if self._is_asn:
-            image.close()
+            dm.save(image)
+            dm.close()
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,8 +9,13 @@ general
 resample
 --------
 
-- Changed parameters read in from drizpar reference file to have a value of None in Spec [#6921] 
+- Changed parameters read in from drizpar reference file to have a value of None in Spec [#6921]
 
+skymatch
+--------
+
+- Fixed a bug in `skymatch` due to which subtracted values were not saved
+  in the input `cal` files when input was an association table. [#6922]
 
 
 source_catalog


jwst/skymatch/skymatch_step.py
```diff
@@ -206,14 +206,18 @@ def _imodel2skyim(self, image_model):
...
-            image.meta.background.subtracted = self.subtract
-        image.meta.cal_step.skymatch = step_status
+            dm.meta.background.method = str(self.skymethod)
+            dm.meta.background.level = sky
+            dm.meta.background.subtracted = self.subtract
+        dm.meta.cal_step.skymatch = step_status
 
         if self._is_asn:
-            image.close()
+            dm.save(image)
```
hbushouse, line 222: Is this actually updating the input cal files on disk? If so, that means folks will need to be careful to keep original versions of their cal files somewhere, for fear of getting overwritten. Can't the results just be passed in memory to the next pipeline steps?

mcara: If we can keep all the files in memory, then what was the point of trying to reduce the memory usage in https://github.com/spacetelescope/jwst/pull/6874.

We have discussed previously (offline) that https://github.com/spacetelescope/jwst/pull/6874 makes sense only when inputs are lists of file names (i.e., ASN). When all input files are already loaded in memory, then the optimization from https://github.com/spacetelescope/jwst/pull/6874 and changes to the `ModelContainer` not to open input files are almost pointless (this step does not use too much extra memory compared to just input files, maybe extra 12% for boolean masks). What I want to say is that let's say we want to match sky in 1000 input images provided as file names and the images are so large that we cannot load them all at once. Then, the optimization from https://github.com/spacetelescope/jwst/pull/6874 would open/load only two files at a given time.

If input images are file names, would it be a problem to update them? Or do you want to save them under different file name?

mcara: Ah, and if `skysub` is true, then not only `meta` is modified but data arrays as well.

hbushouse: Just trying to understand new flow of things based on #6874 and this PR. In the case of "normal" pipeline processing, where the entire calwebb_image3 pipeline is run start to finish, using an ASN table as the initial input, I'm assuming that once the first step in the pipeline executes (tweakreg in this case) all data are then passed on as an already open `ModelContainer` (right?). Hence in that scenario, the input to `skymatch` is a `ModelContainer` and if so, then the changes implemented in #6874 have no effect and all images are still held in memory during `skymatch` execution? And any updates (e.g. keywords giving the computed "sky" values) are present in the data models that are then passed on to the next step (`outlier_detection` in this case) and no intermediate products are saved to disk that contain those `skymatch` updates. Is this correct so far?

mcara: Yes. That is correct. For the whole memory saving thing to make sense, input files should not have been loaded into memory (at once). If they are (there is enough memory to do this), it makes no sense to write these files to disk and load them one at a time, (just for fun).

As we have discussed over slack, this memory saving stuff will need some changes to how pipeline works (i.e., passing unopened files between steps). My work in #6874 is in preparation of those changes. The is, I updated code to support limited memory usage, but it makes sense only under certain circumstances (ie., ability to load data on demand and not getting a list of models already loaded).

hbushouse: OK, thanks for the explanation. I think I'm all caught up now.

So with regards to this PR, when the modified dm's get saved, they should get saved to files with a different product suffix, so that the original input "cal" files don't get overwritten. Let's use the same "bsub" suffix that's used for other types of background-subtracted products (I know this isn't truly a background subtraction, but it's very similar, and I don't want to throw a new product type into all the DMS subsystems). So instead of just doing a simple `dm.save(image)` it would require use of `.save_model` with appropriate file name suffix specified, such as the way it's done here in the level-2b background step:
https://github.com/spacetelescope/jwst/blob/master/jwst/background/background_step.py#L97
Just use the suffix "bsub".

mcara:
Actually, doesn't this step, when run as part of the pipeline and asked to save results, saves output files with suffix `_skymatch`? And also, shouldn't it save a ASN file as well with these `_skymatch` or `_bsub` files?

mcara:
I made another commit to turn off memory optimization code until we have a better machinery to handle outputs that are only file names in a more uniform/consistent was across all steps.

hbushouse (approved):
Looks good.

PR #6926, opened 2022-07-15
Update NIS_SOSS regtest with in-flight data

hbushouse:
This PR addresses updates our regression test of NIRISS SOSS data to use in-flight data. The data itself consists of an exposure that has been segmented (which is usually the case for TSO data), where I'm including just the first 2 segments and have also trimmed down the data in each segment to only contain 10 integrations each (originally 190 each), just to save processing time.

**Checklist**
- [ ] added relevant label(s)


tapastro (approved):
LGTM

PR #6929, opened 2022-07-18
Fix inability of residual fringe code to run on MRS LONG detector

drlaw1558:
This PR resolves the issue discussed in testing for https://jira.stsci.edu/browse/JP-1273 in which the residual fringe step does not run properly on LONG MRS detector data.

**Checklist**
- [x] added relevant label(s)


hbushouse:
A few flake8 complaints that need fixing.

hbushouse:
If folks think this is ready and OK to merge, so that we can get it into our last B8.1 release candidate today, I'll go ahead and do so.

hbushouse (approved):

drlaw1558:
Should be set; seems to be performing as expected in the cases that I've run.

PR #6930, opened 2022-07-18
JP-2677: Copy int_times keywords to output spectrum extensions for SOSS ATOCA

tapastro:
Resolves [JP-2677](https://jira.stsci.edu/browse/JP-2677)

This PR adds back the int_times keywords to SOSS spectral products that have gone missing since the implementation of the ATOCA algorithm.

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 1.6.3 (unreleased)
 ==================
 
+extract_1d
+----------
+
+- Update int_times keywords in SOSS spectral output [#6930]
+
 general
 -------
 
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -3097,8 +3097,8 @@ def populate_time_keywords(
 
     n = 0  # Counter for spectra in output_model.
 
-    for j in range(num_j):  # for each spectrum or order
-        for k in range(num_integ):  # for each integration
+    for k in range(num_integ):  # for each spectrum or order
+        for j in range(num_j):  # for each integration
             row = k + offset
             spec = output_model.spec[n]  # n is incremented below
             spec.int_num = int_num[row]
--- a/jwst/extract_1d/soss_extract/soss_extract.py
+++ b/jwst/extract_1d/soss_extract/soss_extract.py
@@ -790,5 +790,6 @@ def run_extract1d(input_model, spectrace_ref_name, wavemap_ref_name,
     if pipe_utils.is_tso(input_model):
         log.info("Populating INT_TIMES keywords from input table.")
         populate_time_keywords(input_model, output_model)
+        output_model.int_times = input_model.int_times.copy()
 
     return output_model, output_references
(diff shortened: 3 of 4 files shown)


tapastro:
@hbushouse I was initially confused on what the expected output changes were - in reality, this was not complicated. I figured an INT_TIMES extension was desired in the output MultiSpecModel, but the simple change correctly adds the keyword values to each spectral output extension.

stscieisenhamer (approved):
Nice! Short and sweet.

hbushouse:
Spot checking results using this code we do now in fact get the INT_TIMES table extension in the x1dints files coming out of both calwebb_spec2 and calwebb_tso3 processing, as well as extension header keywords. But I think something might be wrong in either the ordering of the EXTRACT1D extensions in the x1dints file or the values of the SPORDER and INT_NUM keywords in each extension header. For example, I'm seeing:
```
extver  int_num  sporder
1           181          1
2           182         2
3           183         3
4           184         1
5           185         2
6           186         3
7           187         1
etc.
```
I would think that int_num would increase while sporder stays constant, then int_num would start over when sporder incremented by 1 (e.g. [181,1], [182,1], [183,1], ..., [190,1], [181,2], [182,2], [183,2], ..., [190,2], [181,3], [182,3], ...)

tapastro:
The new commit to create a concatenated int_times table should work for segmented data, but I don't know if we expect all data to have a split, but single 'final', 'master' int_times table. 

Also, it's going to fail on the regtest dataset, because len(int_times)=11, but len(data[:])=10.

hbushouse:
This PR could benefit greatly from a rebase. When I ran regtests against it I got hundreds of differences due to some keyword updates that've been merged since this was created, not to mention all the flake8 updates. It would be much easier to sort through the regtest results without all of those "false positives".

tapastro:
> This PR could benefit greatly from a rebase. When I ran regtests against it I got hundreds of differences due to some keyword updates that've been merged since this was created, not to mention all the flake8 updates. It would be much easier to sort through the regtest results without all of those "false positives".

I was having problems finding complete SOSS datasets to test on that didn't trip an error related to the reference file issue that we thought we had mitigated. That ended up creating the other PR to update the SOSS algorithm, which also contains this fix. We can (should?) keep this fix independent of the algorithm update PR, but it may require an update to the SOSS regression data.

tapastro:
From local testing, it doesn't appear to cause errors, so okify'ing should be enough?

hbushouse:
> From local testing, it doesn't appear to cause errors, so okify'ing should be enough?

I assume it at least shows the expected differences from now having the INT_TIMES table extension and other header keywords present?

tapastro:
> > From local testing, it doesn't appear to cause errors, so okify'ing should be enough?
> 
> I assume it at least shows the expected differences from now having the INT_TIMES table extension and other header keywords present?

Realizing I may not have the updated test set locally - updating my local cache and re-running... The output showed lengthy failures for each test, which upon examination may indicate that I still have the full-size files and not the truncated versions.

tapastro:
Alright, seeing expected failures (but no errors) with updated test data - only three of seven fail.

hbushouse (approved):
OK, so the code looks good and it sounds like test results are also good. Will merge.

PR #6933, opened 2022-07-20
automatically add labels to pull requests based on changed files

zacharyburnett:
This GitHub workflow runs a labeler action on new / edited pull requests that scans the filenames of changed files and applies relevant labels. This process is configured in `.github/labeler.yml`, which I copied and modified from https://github.com/spacetelescope/stsynphot_refactor/blob/master/.github/labeler.yml

An example of this labeler working can be seen here: https://github.com/spacetelescope/stsynphot_refactor/pull/153
image

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -11,6 +11,22 @@ installation:
   - 'requirements-*.txt'
   - MANIFEST.in
 
+# --------------------------------------- testing ---------------------------------------
+
+testing:
+  - '**/tests/**'
+  - .github/workflows/*.yml
+  - .bandit.yaml
+  - .codecov.yml
+  - conftest.py
+  - tox.ini
+
+regression_testing:
+  - 'jwst/regtest/**'
+  - 'Jenkinsfile*'
+
+# --------------------------------------- modules ---------------------------------------
+
 ami:
   - 'jwst/ami/**'
 
@@ -31,10 +47,6 @@ bar shadow:
 
 combine_1d:
   - 'jwst/combine_1d/**'
-
-coron3 pipeline:
-  - 'jwst/coron/**'
-  - jwst/pipeline/calwebb_coron3.py
 
 cube_build:
   - 'jwst/cube_build/**'
@@ -120,18 +132,11 @@ persistence:
 photom:
   - 'jwst/photom/**'
 
-pipeline:
-  - 'jwst/pipeline/**'
-
 ramp_fitting:
   - 'jwst/ramp_fitting/**'
 
 refpix:
   - 'jwst/refpix/**'
-
-regression_testing:
-  - 'jwst/regtest/**'
-  - 'Jenkinsfile*'
 
 resample:
   - 'jwst/resample/**'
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,211 @@
...
+  - 'jwst/**/*wfss*/**/*'
+
+NIRSPEC:
+  - 'jwst/**/*nirspec*/**/*'
+
+NIRCAM:
+  - 'jwst/**/*nircam*/**/*'
+
+NIRISS:
+  - 'jwst/**/*niriss*/**/*'
```
zacharyburnett, line 211: I'm not sure if each instruments has specific files that are exclusive to them, so I used general patterns here.

hbushouse: It's rare, but there are a few instances here and there of modules that have an instrument-specific name.

.github/labeler.yml
```diff
@@ -0,0 +1,215 @@
...
+
+barshadow:
+  - 'jwst/barshadow/**'
+
+combine_1d:
+  - 'jwst/combine_1d/**'
+
+coron3 pipeline:
+  - 'jwst/coron/**'
+  - jwst/pipeline/calwebb_coron.py
```
hbushouse, line 37: I think the actual module name is calwebb_coron**3**.py

.github/labeler.yml
```diff
@@ -0,0 +1,215 @@
...
+mrs_imatch:
+  - 'jwst/mrs_imatch/**'
+
+msa_flagging:
+  - 'jwst/msaflagopen/**'
+
+outlier_detection:
+  - 'jwst/outlier_detection/**'
+
+path loss:
```
hbushouse, line 114: does the pathloss label have a whitespace in the middle of it?

.github/labeler.yml
```diff
@@ -0,0 +1,215 @@
...
+  - 'jwst/**/*nirspec*'
+  - 'jwst/**/*nirspec*/**'
+
+NIRCAM:
+  - 'jwst/**/*nircam*'
+  - 'jwst/**/*nircam*/**'
+
+NIRISS:
+  - 'jwst/**/*niriss*'
+  - 'jwst/**/*niriss*/**'
```
hbushouse, line 215: You're missing FGS and MIRI.

hbushouse (commented):
Looks OK, other than minor comments/questions?  Will be eager to see how it actually works.

.github/labeler.yml
zacharyburnett: yes indeed it is `path loss`, you can see here: https://github.com/spacetelescope/jwst/labels?q=path

Should the label be renamed?

hbushouse: No, that's OK. I just didn't realize it was spelled that way.

zacharyburnett:
> Looks OK, other than minor comments/questions? Will be eager to see how it actually works.

Yeah, it should make PRs easier!

.github/labeler.yml
```diff
@@ -0,0 +1,223 @@
...
+assign_wcs:
+  - 'jwst/assign_wcs/**'
+
+associations:
+  - 'jwst/associations/**'
+
+background:
+  - 'jwst/background/**'
+
+barshadow:
```
hbushouse, line 29: It looks like the "bar shadow" label also has a whitespace in it.

.github/labeler.yml
```diff
@@ -0,0 +1,223 @@
...
+  - 'jwst/barshadow/**'
+
+combine_1d:
+  - 'jwst/combine_1d/**'
+
+coron3 pipeline:
```
hbushouse, line 35: could add similar entries for the other pipelines: image2, image3, spec2, spec3, ami3, tso3, detector1

nden (approved):

PR #6934, opened 2022-07-21
JP-2688: Skymatch - Update datamodel data when subtract is True

mcara:
Resolves [JP-2688](https://jira.stsci.edu/browse/JP-2688)

Memory saving features introduced a bug in the `skymatch` step due to which background value is not subtracted from image model data even though step parameter `subtract` is set to True. This is due to how data now are written/saved to temporary files and the machinery involved in this. Due to these changes, the background was subtracted from a copy of the datamodel's array instead from original array.

This PR makes sure that model data are updated with matched/subtracted data values at the end of the step upon successful completion of the step.

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,13 @@ general
 
 -
 
+skymatch
+--------
+
+- Fixed a bug in ``skymatch`` due to which computed background values were
+  not subtracted from image data when ``subtract=True``. [#6934]
+
+
 1.6.2 (2022-07-19)
 ==================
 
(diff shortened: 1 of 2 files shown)


mcara:
@hbushouse I do not think Python 3.10 failures are related to this PR.

hbushouse:
So #6922 did not completely fix this problem?

hbushouse:
> @hbushouse I do not think Python 3.10 failures are related to this PR.

Correct. It's the usual kind of hiccup we often get with access to CRDS ref files curing CI tests.

hbushouse (approved):
Look good. Tested on MIRI and NIRCam image3 regression test data and verified that pre-fix the skymatch results have nothing subtracted and after the fix they are subtracted by their respective BKGLEVEL values. Also verified that with subtract=False there's no change to anything in the regression tests.

PR #6935, opened 2022-07-22
JP-2690: Convenience function for updating FITS WCS info

mcara:
Resolves [JP-2690](https://jira.stsci.edu/browse/JP-2690)

Numerous users both on Slack and in the helpdesk are asking about updating FITS WCS info from the current GWCS stored in data model's `meta.wcs`. This PR adds a convenience function `update_fits_wcsinfo()` to `assign_wcs.util` module and modifies `assign_wcs` to call this convenience function.

**Checklist**
- [x] added entry in `CHANGES.rst` within the relevant release section
- [ ] updated or added relevant tests
- [x] updated relevant documentation
- [x] added relevant milestone
- [x] added relevant label(s)

PR #6936, opened 2022-07-22
JP-2691: Update FITS WCS wcsinfo field at the end of tweakreg

mcara:
Resolves [JP-2691](https://jira.stsci.edu/browse/JP-2691)
Resolves [JP-2586](https://jira.stsci.edu/browse/JP-2586)

This PR implements `FITS WCS` update at the end of `tweakreg` step if the `GWCS` of a data model was "tweaked" by `tweakreg`.

This PR will need to be rebased after #6935 is merged. It depends on that #6935. Initially unit tests will fail because of this.

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,12 @@ skymatch
 - Fixed a bug in ``skymatch`` due to which computed background values were
   not subtracted from image data when ``subtract=True``. [#6934]
 
+tweakreg
+--------
+
+- ``tweakreg`` step now updates FITS WCS stored in ``datamodel.meta.wcsinfo``
+  from data model's tweaked GWCS. [#6936]
+
 
 1.6.2 (2022-07-19)
 ==================
--- a/jwst/tweakreg/tweakreg_step.py
+++ b/jwst/tweakreg/tweakreg_step.py
@@ -16,7 +16,7 @@ from tweakwcs.matchutils import TPMatch
 # LOCAL
 from ..stpipe import Step
 from .. import datamodels
-
+from ..assign_wcs.util import update_fits_wcsinfo
 from . import astrometric_utils as amutils
 from .tweakreg_catalog import make_tweakreg_catalog
 
@@ -319,16 +319,9 @@ class TweakRegStep(Step):
 
                 imcat.meta['image_model'].meta.wcs = imcat.wcs
 
-                """
                 # Also update FITS representation in input exposures for
                 # subsequent reprocessing by the end-user.
-                # Not currently enabled, but may be requested later...
-                gwcs_header = imcat.wcs.to_fits_sip(max_pix_error=0.1,
-                                                max_inv_pix_error=0.1,
-                                                degree=3,
-                                                npoints=128)
-                imcat.meta['image_model'].wcs = wcs.WCS(header=gwcs_header)
-                """
+                update_fits_wcsinfo(imcat.meta['image_model'])
 
         return images
 


hbushouse (commented):
Looks pretty simple and straightforward to me. Will reserve final judgement (i.e. approving) until it's been rebased and we see the new CI results.

hbushouse (approved):

hbushouse:
The full regtest run performed after merging this PR shows the expected differences in the FITS WCS keywords in the header of the crf files (downstream of tweakreg), but the new files are also missing the sets of AP_n_m and BP_n_m keywords that used to be there (propagated from the cal files upstream). Is that expected? Are those associated with the inverse transforms, which aren't stored any more?

```
       Headers have different number of cards:
        a: 96
        b: 122
       Extra keyword 'AP_0_2' in b: 1.12990966654351e-06
       Extra keyword 'AP_0_3' in b: -1.8557669161344e-09
       Extra keyword 'AP_0_4' in b: 8.83309765685681e-13
       Extra keyword 'AP_1_1' in b: 1.56948423810618e-05
       Extra keyword 'AP_1_2' in b: 9.2559179012512e-09
       Extra keyword 'AP_1_3' in b: -1.3312158936888e-11
       Extra keyword 'AP_2_0' in b: -6.9752001346959e-07
       Extra keyword 'AP_2_1' in b: -2.8854553824736e-09
       Extra keyword 'AP_2_2' in b: -1.7076522617462e-12
       Extra keyword 'AP_3_0' in b: 2.50796460211884e-09
       Extra keyword 'AP_3_1' in b: -1.1625910573439e-11
       Extra keyword 'AP_4_0' in b: 1.52481086499306e-13
       Extra keyword 'AP_ORDER' in b: 4
       Extra keyword 'BP_0_2' in b: 4.21639400716975e-07
       Extra keyword 'BP_0_3' in b: 2.96279802623005e-08
       Extra keyword 'BP_0_4' in b: -1.8459602814165e-11
       Extra keyword 'BP_1_1' in b: 3.40273661794458e-06
       Extra keyword 'BP_1_2' in b: -6.2225014148017e-09
       Extra keyword 'BP_1_3' in b: 2.59262374303019e-12
       Extra keyword 'BP_2_0' in b: 7.09838057497431e-06
       Extra keyword 'BP_2_1' in b: 1.60973519139947e-08
       Extra keyword 'BP_2_2' in b: -1.6867248128436e-11
       Extra keyword 'BP_3_0' in b: -1.3126268291304e-09
       Extra keyword 'BP_3_1' in b: 9.46599997948238e-14
       Extra keyword 'BP_4_0' in b: -3.4624772588674e-12
       Extra keyword 'BP_ORDER' in b: 4
```
(note that image "b" is the truth file, so it has keywords in it that the new file does not)

mcara:
Yes Howard, it is expected. `AP_*` and `BP_*` are basically useless given how software out there works. These coefficients give polynomial approximation of the inverse of the forward(pixel->world) distortion. None of HST image headers provide these and everything works fine because `astropy.wcs.WCS.all_world2pix()` (inverse transformation) iterates until it finds a solution. One could question this approach ("wouldn't a pre-computed approximation be faster?) but in practice numerical solution is only marginally slower (if at all) while providing higher accuracy.

I know that `astropy.wcs.WCS` is ignoring inverse coefficients. I am not familiar with AST WCS but a quick search of their documentation gives:

> **SipReplace:** This attribute is a boolean value which specifies how SIP keywords should be handled when reading
a FITS-WCS encoded header using the astRead function. ... **If SipReplace is non-zero, then any SIP keywords describ-ing the inverse transformation (i.e. from WCS to pixel coordinates) are ignored. Instead a new
inverse transformation is found by performing a fit to the forward transformation.** ... **The default value is 1.**

This was what I was asking in https://github.com/spacetelescope/jwst/pull/6935#issuecomment-1192799483 (I did leave them in the initial assignment in ``assign_wcs`` because it was there before).

hbushouse:
OK, thanks. I will go ahead and update the regtest truth files.

nden:
Actually, since these WCS is there for DS9, it may be good to have the inverses in the headers.
@mcara is right that if you are running the code the iterative inverse is probably more accurate but DS9 just uses what's in the headers.

mcara:
Isn't DS9 using just AST WCS?

mcara:
> Isn't DS9 using plain AST WCS?

To test this, I created three identical files except for their FITS WCS: two have inverse SIP coefficients (2nd and 4th order) and another one does not. I put the three file on the central storage /user/mcara/inv-sip/:
```
model_fwd_sip.fits - uses only forward SIP coefficents
model_fwd_inv2_sip.fits - has inverse SIP coefficients of AP_ORDER=BP_ORDER=2
model_fwd_inv4_sip.fits - has inverse SIP coefficients of AP_ORDER=BP_ORDER=4
```
To evaluate SIP/no-SIP difference:
```python
from astropy.io import fits
from astropy.wcs import WCS
with fits.open('model_fwd_sip.fits') as h:
    w = WCS(h[1].header, h)
    
print(w.wcs_world2pix(*w.all_pix2world(1, 1, 0), 0))
print(w.all_world2pix(*w.all_pix2world(1, 1, 0), 0))
```
shows that the error of not using SIP is about 0.7 pixels:
```
[array(1.7385737), array(1.68617794)]
[array(1.00000001), array(1.00000001)]
```
This kind of error would be visible in DS9.

Load the images in DS9, lock the frame to WCS and blink them. I see no difference in DS9 between 4th order (higher accuracy inverse SIP) and the header with no inverse SIP. If DS9 could not handle SIP without inverse coefficients, a difference of 0.7 pixels would be easily visible in DS9. **I conclude DS9 can deal with SIP correctly without inverse SIP coefficients.** On the other hand, if you compare 2nd (less accurate) and 4th order or 2nd order inverse with SIP without inverse coefficients - you will notice a slight motion between frames. This shows DS9 will pick inverse SIP coefficients, **if provided**. Indeed, looking at the release notes for DS9 version 8.3:

> 05.23.2021 WCS: Make SipReplace=0 as the default behavior.

So, if a less accurate SIP inverse is provided - DS9 will prefer to use it over the more accurate iterative inverse. I still think it is better not to provide inverse SIP. However, I'll make a PR just in case and you can decide whether you want to merge it.

PR #6938, opened 2022-07-22
JP-2687: Allow alignment of a single image/group to Gaia bypassing first alignment stage

mcara:
Resolves [JP-2687](https://jira.stsci.edu/browse/JP-2687)

This PR addresses the issue of `tweakreg` step exiting when only one input image is provided. This behavior was the desired one when there was only one "relative alignment" stage in `tweakwcs`. With the addition of alignment to Gaia stage, it now makes sense to bypass first stage and allow single images to align to GAIA. This PR enables this logic.

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,8 +16,8 @@ ami_analyze
 -----------
 
 - Revert Fourier Transform code to avoid using Poppy which was recently updated
-  to use a different sign convention.[#6967]  
-  
+  to use a different sign convention.[#6967]
+
 assign_wcs
 ----------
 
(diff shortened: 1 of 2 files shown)


mcara:
Anyone?

stsci-hack (approved):
This is a simple enough change in the logic that looks good to me.

hbushouse (approved):
Looks good to me.

hbushouse:
Everybody happy with this and OK to merge?

hbushouse:
Oh, looking at the checklist for things to be done I see documentation mentioned. Should the tweakreg step docs be updated to indicate that it's now possible to use a single image as input if you're just interested in aligning to Gaia (or eventually a user-supplied catalog)?

mcara:
> ... indicate that it's now possible to use a single image as input if you're just interested in aligning to Gaia (or eventually a user-supplied catalog)?

I do not think that is needed: I believe it is obvious that one needs a reference (either an external catalog or another image from which to extract a reference catalog) to which an image can be "aligned". This PR was to fix a breakdown in this expectation: users were requesting alignment to an external catalog (GAIA) but the code would crash for a single image. So the bug was in not meeting normal expectation.

hbushouse:
OK, sounds reasonable. Merging.

PR #6939, opened 2022-07-24
tweakreg should report inverse SIP coefficients in FITS WCS

mcara:
This PR enables `tweakreg` to report inverse SIP coefficients in data model's `meta.wcsinfo` when computing FITS WCS from GWCS.

I believe it is better not to include these coefficients as most common WCS software would resort to a more accurate iterative solution, but I make this PR for in case it is needed. See #6936 discussions for more details.

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,8 +16,8 @@ ami_analyze
 -----------
 
 - Revert Fourier Transform code to avoid using Poppy which was recently updated
-  to use a different sign convention.[#6967]  
-  
+  to use a different sign convention.[#6967]
+
 assign_wcs
 ----------
 
@@ -67,6 +67,9 @@ tweakreg
 
 - The ``tweakreg`` step now masks ``NON_SCIENCE`` pixels when
   calculating the source detection theshold. [#6940]
+
+- ``tweakreg`` will report inverse SIP coefficients in updated
+  FITS WCS. [#6939]
 
 1.6.2 (2022-07-19)
 ==================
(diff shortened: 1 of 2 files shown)


hbushouse:
Given the information provided in JP-2691, which suggests that ds9 can compute it's own inverse when not provided in the header and what it computes appears to be quite accurate, I don't see the need for this. Of course it also can't really hurt to have it, in case other tools can't compute their own inverse.

mcara:
> Of course it also can't really hurt to have it...

It can, if provided inverse SIP approximation is not very accurate due to how DS9 has been set-up designed to use this approximation first, if available.

hbushouse:
> > Of course it also can't really hurt to have it...
> 
> It can, if provided inverse SIP approximation is not very accurate due to how DS9 has been set-up designed to use this approximation first, if available.

Good point.

mcara:
On the other hand, current level of accuracy (0.005 pixels) for the inverse if high enough for visualization applications (DS9) and `astropy.wcs.WCS.all_worls2pix()` ignores inverse SIP. So does AST (by default). But I do not know what else is out there that may or may not use inverse SIP. So, by having high enough accuracy, we can mitigate the issue of some applications such as DS9 are using inv SIP.

I just don't want us to arrive at a final solution.

hbushouse:
Well, if we can easily provide a fairly accurate reverse solution, then I guess we're back to the state of "we might as well put it in, because it shouldn't do any damage". What do you think @nden ?

nden:
👍 for adding it

mcara:
Again, I recommend not to write out inverse SIP but I understand if this is desired and that is the only reason for this PR.

PR #6940, opened 2022-07-25
Mask NON_SCIENCE pixels when calculating the detection threshold

larrybradley:
This PR updates the detection threshold calculation to mask `NON_SCIENCE` pixels (e.g., MIRI).

xref: INC0180066

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,6 +54,8 @@ tweakreg
 - ``tweakreg`` step now updates FITS WCS stored in ``datamodel.meta.wcsinfo``
   from data model's tweaked GWCS. [#6936, #6947]
 
+- The ``tweakreg`` step now masks ``NON_SCIENCE`` pixels when
+  calculating the source detection theshold. [#6940]
 
 1.6.2 (2022-07-19)
 ==================
--- a/jwst/tweakreg/tweakreg_catalog.py
+++ b/jwst/tweakreg/tweakreg_catalog.py
@@ -86,7 +86,7 @@ def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
     if sources:
         catalog = sources[columns]
     else:
-        catalog = Table(names=columns, dtype=(np.int_, np.float_, np.float_,
-                                              np.float_))
+        # return an empty table
+        catalog = Table(names=columns, dtype=(int, float, float, float))
 
     return catalog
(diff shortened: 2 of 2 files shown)


mcara (approved):
LGTM. Many thanks!

tapastro (approved):
LGTM

CHANGES.rst
```diff
@@ -33,6 +33,8 @@ tweakreg
 - ``tweakreg`` step now updates FITS WCS stored in ``datamodel.meta.wcsinfo``
   from data model's tweaked GWCS. [#6936, #6947]
 
+- The ``tweakreg`` step now masks ``NON_SCIENCE`` pixels when
+  calculating the detection theshold. [#6940]
```
mcara, line 37:
```suggestion
  calculating source detection threshold. [#6940]
```

larrybradley: Updated.

hbushouse:
Was there ever a JP ticket created to track this problem? @tapastro

tapastro:
Looks like one was never made - I can create one for tracking purposes, if you'd like (and I think I've been better about creating them over the past week or so.)

hbushouse:
Yes, it's good to have them in the JP project so that everyone else knows what changes and fixes are being implemented.

hbushouse:
A question was raised as to why NON_SCIENCE pixels needed to be rejected, because they're also always flagged as DO_NOT_USE and (theoretically) we should always be rejecting DO_NOT_USE pixels from everything. But from looking at this code, it appears that the only dq flag that's paid attention to is NON_SCIENCE (and previously no attention was given to any flags). Is that true?  If so, are there other places in the step where we should be rejecting/masking pixels based on DO_NOT_USE?  @larrybradley @mcara

mcara:
I do not know enough about precise meanings of all these flags (i.e., whether `DO_NOT_USE` is a subset of `NON_SCIENCE`). With regard to the last part of the question: "No", source finding is the only part of the code that cares about DQ flags.

hbushouse:
But even if this is the only place we should be paying attention to DQ flags, I think it would be better to upgrade the previous upgrade to mask out both NON_SCIENCE **and** DO_NOT_USE. There may be a quite a few other bad pixels in the individual images that are flagged as DO_NOT_USE but not NON_SCIENCE.

mcara:
Is there a JIRA ticket for adding `DO_NOT_USE`?

hbushouse:
Wait for it ...

hbushouse:
JP-2814

PR #6942, opened 2022-07-26
JP-2694: reset input model after Source Catalog Step

cshanahan1:
A copy of the input model is not made in SourceCatalogStep since there is no return model (just the table). In the step, the units of the data/err arrays are converted and the background is subtracted from the data. If someone attempts successive runs of the step, it will fail because of these changes to the input.

This PR resets the model's data/err arrays and metadata information about units so that the input remains unchanged.

Resolves [JP-2694](https://jira.stsci.edu/browse/JP-2694)

### DIFF
--- a/jwst/source_catalog/source_catalog.py
+++ b/jwst/source_catalog/source_catalog.py
@@ -125,6 +125,22 @@ class JWSTSourceCatalog:
         self.model.meta.bunit_data = unit.name
         self.model.meta.bunit_err = unit.name
 
+    def convert_from_jy(self):
+        """
+        Convert the data and errors from Jy to MJy/sr and change from
+        `~astropy.unit.Quantity` objects to `~numpy.ndarray`.
+        """
+
+        to_mjy_sr = 1.e6 * self.model.meta.photometry.pixelarea_steradians
+        self.model.data /= to_mjy_sr
+        self.model.err /= to_mjy_sr
+
+        self.model.data = self.model.data.value  # remove units
+        self.model.err = self.model.err.value  # remove units
+
+        self.model.meta.bunit_data = 'MJy/sr'
+        self.model.meta.bunit_err = 'MJy/sr'
+
     @staticmethod
     def convert_flux_to_abmag(flux, flux_err):
         """
@@ -947,5 +963,8 @@ class JWSTSourceCatalog:
         self.meta['aperture_params'] = self.aperture_params
         self.meta['abvega_offset'] = self.abvega_offset
         catalog.meta.update(self.meta)
+
+        # reset units on input model back to MJy/sr
+        self.convert_from_jy()
 
         return catalog
--- a/jwst/source_catalog/tests/test_source_catalog.py
+++ b/jwst/source_catalog/tests/test_source_catalog.py
@@ -1,5 +1,6 @@
 import pytest
 import numpy as np
+from numpy.testing import assert_allclose
 from photutils.datasets import make_gwcs
 from jwst.datamodels import ImageModel
 from ..source_catalog_step import SourceCatalogStep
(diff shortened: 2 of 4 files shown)


nden:
Don't forget to add the JP ticket number

jwst/source_catalog/source_catalog.py
```diff
@@ -125,6 +125,22 @@ def convert_to_jy(self):
         self.model.meta.bunit_data = unit.name
         self.model.meta.bunit_err = unit.name
 
+    def convert_from_jy(self):
+        """
+        Convert the data and errors from MJy/sr to Jy and convert to
+        `~astropy.unit.Quantity` objects.
```
larrybradley, line 131:
```suggestion
        Convert the data and errors from Jy to MJy/sr and change
        from `~astropy.unit.Quantity` to `~numpy.ndarray`.
```

jwst/source_catalog/tests/test_source_catalog.py
```diff
@@ -122,3 +122,28 @@ def test_model_without_apcorr_data(nircam_model_without_apcorr):
...
     assert cat is None
+
+
+def test_input_model_reset(nircam_model):
+    """ Changes to input model data are made in SourceCatalogStep - make sure that
+        these changes are correctly reversed so the input model data/err arrays
+        remain unchanged after processing (to avoid copying datamodel),
+        and that the units are in MJy/sr before and after."""
+
+    original_data = nircam_model.data
```
larrybradley, line 133:
```suggestion
    original_data = nircam_model.data.copy()
```
If you don't copy the data here, it will be a reference  to the model data (i.e., the test below will always pass even if the model data changes).

jwst/source_catalog/tests/test_source_catalog.py
```diff
@@ -122,3 +122,28 @@ def test_model_without_apcorr_data(nircam_model_without_apcorr):
...
+    original_data_unit = nircam_model.meta.bunit_data
+    original_err = nircam_model.err
```
larrybradley, line 135:
```suggestion
    original_err = nircam_model.err.copy()
```

jwst/source_catalog/tests/test_source_catalog.py
```diff
@@ -122,3 +122,28 @@ def test_model_without_apcorr_data(nircam_model_without_apcorr):
...
+    assert(original_data_unit == original_err_unit == 'MJy/sr')
+
+    step = SourceCatalogStep(snr_threshold=0.5, npixels=5,
+                             bkg_boxsize=50, kernel_fwhm=2.0,
+                             save_results=False)
+
+    step.run(nircam_model)
+
+    assert(np.all(original_data == nircam_model.data))
+    assert(np.all(original_err == nircam_model.err))
```
larrybradley, line 147:
```suggestion
    assert_array_equal(original_data, nircam_model.data))
    assert_array_equal(original_err, nircam_model.err)
```

You will also need to add this line to the top imports:  `from numpy.testing import assert_array_equal`.

`assert_array_equal` is better because it compares NaNs.

larrybradley (changes requested):
Thanks, @cshanahan1.  I have a few minor comments.

larrybradley:
The RTD failure is unrelated.  It appears the RTD build has been failing for the past ~3 days.

hbushouse:
Wouldn't it be way simpler to just go ahead and make a copy of the input datamodel and work on the copy? I know we've been trying to avoid making too many, unnecessary copies of data in order to not overload memory usage, but for this step it's a single image being used as input. Granted, that image could be a big mosaic, but still just a single `ImageModel`.

cshanahan1:
@hbushouse i had originally just made it a copy but larry suggested avoiding that if possible. what do you think? it would just need to copy the data/err arrays, then reset the unit in the header.

larrybradley:
I wanted to avoid making copies if possible because the mosaics could be very large.  Copying requires more memory usage and more cpu time, which I thought was unnecessary given this is the last step of the pipeline.  I hadn't considered that an offline (not operations) user would want to re-run this step multiple times in succession -- but I don't really know if that's a common use case.  The changes here are straightforward (and all done in place).

hbushouse (commented):
If we're sticking with this method (reverting the data, as opposed to copying it at the beginning) the updates requested by @larrybradley need to be implemented.

cshanahan1:
in the original test i wrote for this, which was failing, there were a good number of pixels (all located in the bright source in the test image) that exceeded the 1.e-7 threshold to pass assert_allclose. i didn't want to just brush a potential issue under the rug by increasing the tolerance, but after discussing with @larrybradley we decided that the discrepancy was probably because the input data is float32 while the background array is float64. i increased the tolerance to 5.e-7 and the test passes.

larrybradley (approved):
Thanks, @cshanahan1!

PR #6943, opened 2022-07-26
JP-2693: Flag groups do_not_use after detected ramp jumps

karllark:
Resolves [JP-2693](https://jira.stsci.edu/browse/JP-2693)

This PR implements the enhancement to the jump step where groups after detected jumps are flagged do_not_use. This "corrects" transients that have been seen to make the slope after a jump different than before. This PR allows for flagging to be different for two different jump thresholds.

The main changes are in `stcal` and this PR provides the needed interface for `jwst`.

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -55,6 +55,12 @@ extract_1d
 - Update ``int_times`` keywords and copy the ``INT_TIMES`` table extension to SOSS
   spectral output (x1d) files [#6930]
 
+jump
+----
+
+- Added flagging after detected ramp jumps based on two DN thresholds and
+  two number of groups to flag [#6943]
+
 master_background
 -----------------
 
--- a/docs/jwst/jump/arguments.rst
+++ b/docs/jwst/jump/arguments.rst
@@ -38,14 +38,14 @@ The ``jump`` step has five optional arguments that can be set by the user:
   not have its neighbors flagged. The goal is to prevent flagging jumps that would be too
   small to significantly affect the slope determination.  The default value is 10.
 
-*  ``--after_jump_flag_dn1'': A floating point value in units of DN that gives the
+*  ``--after_jump_flag_dn1``: A floating point value in units of DN that gives the
    minimum ramp jump for after jump flagging for the 1st after ramp flagging.
 
-* ``--after_jump_flag_time1'': A floating point value in units of seconds that gives the
+* ``--after_jump_flag_time1``: A floating point value in units of seconds that gives the
   time since the ramp jump to flag groups as ramp jumps for the 1st after jump flagging.
 
-*  ``--after_jump_flag_dn1'': A floating point value in units of DN that gives the
+*  ``--after_jump_flag_dn1``: A floating point value in units of DN that gives the
    minimum ramp jump for after jump flagging for the 2nd after ramp flagging.
 
-* ``--after_jump_flag_time1'': A floating point value in units of seconds that gives the
+* ``--after_jump_flag_time1``: A floating point value in units of seconds that gives the
   time since the ramp jump to flag groups as ramp jumps for the 2nd after jump flagging.
--- a/docs/jwst/jump/arguments.rst
+++ b/docs/jwst/jump/arguments.rst
@@ -38,4 +38,14 @@ The ``jump`` step has five optional arguments that can be set by the user:
   not have its neighbors flagged. The goal is to prevent flagging jumps that would be too
   small to significantly affect the slope determination.  The default value is 10.
 
+*  ``--after_jump_flag_dn1'': A floating point value in units of DN that gives the
+   minimum ramp jump for after jump flagging for the 1st after ramp flagging.
 
+* ``--after_jump_flag_time1'': A floating point value in units of seconds that gives the
+  time since the ramp jump to flag groups as ramp jumps for the 1st after jump flagging.
+
+*  ``--after_jump_flag_dn1'': A floating point value in units of DN that gives the
+   minimum ramp jump for after jump flagging for the 2nd after ramp flagging.
+
+* ``--after_jump_flag_time1'': A floating point value in units of seconds that gives the
+  time since the ramp jump to flag groups as ramp jumps for the 2nd after jump flagging.
--- a/docs/jwst/jump/description.rst
+++ b/docs/jwst/jump/description.rst
@@ -34,7 +34,7 @@ output PIXELDQ array.
 The SCI and ERR arrays of the input data are not modified.
 
 The current implementation uses the two-point difference method described
-in Anderson&Gordon2011_.
+in `Anderson & Gordon (2011) <https://ui.adsabs.harvard.edu/abs/2011PASP..123.1237A>`_.
 
 Two-Point Difference Method
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -59,6 +59,15 @@ The two-point difference method is applied to each integration as follows:
   is used rather than the clipped median.
 * If there are only two differences (three groups), the smallest one is compared to the larger
   one and if the larger one is above a threshold, it is flagged as a jump.
+
+* If flagging of the 4 neighbors is requested, then the 4 adjacent pixels will
+  have ramp jumps flagged in the same group as the central pixel as long as it has
+  a jump between the min and max requested levels for this option.
+
+* If flagging of groups after a ramp jump is requested, then the groups in the
+  requested time since a detected ramp jump will be flagged as ramp jumps if
+  the ramp jump is above the requested threshold.  Two thresholds and times are
+  possible for this option.
 
 Note that any ramp values flagged as SATURATED in the input GROUPDQ array
 are not used in any of the above calculations and hence will never be
@@ -87,5 +96,3 @@ The use of the reference files is flexible. Full-frame reference
 files can be used for all science exposures, in which case subarrays will be
 extracted from the reference file data to match the science exposure, or
 subarray-specific reference files may be used.
-
-.. _Anderson&Gordon2011: https://ui.adsabs.harvard.edu/abs/2011PASP..123.1237A
--- a/jwst/jump/jump.py
+++ b/jwst/jump/jump.py
@@ -11,7 +11,9 @@ log.setLevel(logging.DEBUG)
 def run_detect_jumps(input_model, gain_model, readnoise_model,
                      rejection_thresh, three_grp_thresh, four_grp_thresh,
                      max_cores, max_jump_to_flag_neighbors,
-                     min_jump_to_flag_neighbors, flag_4_neighbors):
+                     min_jump_to_flag_neighbors, flag_4_neighbors,
+                     after_jump_flag_dn1, after_jump_flag_time1,
+                     after_jump_flag_dn2, after_jump_flag_time2):
 
     # Runs `detect_jumps` in stcal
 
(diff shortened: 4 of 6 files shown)


dmggh:
I just changed ticket number '2963' -> '2693' in description.

karllark:
If understand things correctly, the tests are failing as the version of stcal used for the tests does not include the changes needed for this PR.  After the stcal PR has been merged, then maybe the tests will pass.

The tests pass on my computer when the appropriate branch on stscal (jump_transient_flag) is used.

karllark:
Ready for review.  Likely, the corresponding PR in stcal should be reviewed at the same time.

CHANGES.rst
```diff
@@ -48,6 +48,12 @@ datamodels
 
 - Add P_SUBARR keyword to the `DarkModel` schema. [#6951]
 
+jump
+----
+
+- Added flagging after detected ramp jumps based on two DN thresholds and
```
cshanahan1, line 54: you changed this to electron threshold in stcal

docs/jwst/jump/arguments.rst
```diff
@@ -38,4 +38,14 @@ The ``jump`` step has five optional arguments that can be set by the user:
   not have its neighbors flagged. The goal is to prevent flagging jumps that would be too
   small to significantly affect the slope determination.  The default value is 10.
 
+*  ``--after_jump_flag_dn1``: A floating point value in units of DN that gives the
```
cshanahan1, line 41: update this to be in sync with what is in stcal now, the same docstrings there and e instead of dn

docs/jwst/jump/description.rst
```diff
@@ -60,6 +60,15 @@ The two-point difference method is applied to each integration as follows:
 * If there are only two differences (three groups), the smallest one is compared to the larger
   one and if the larger one is above a threshold, it is flagged as a jump.
 
+* If flagging of the 4 neighbors is requested, then the 4 adjacent pixels will
+  have ramp jumps flagged in the same group as the central pixel as long as it has
+  a jump between the min and max requested levels for this option.
+
+* If flagging of groups after a ramp jump is requested, then the groups in the
+  requested time since a detected ramp jump will be flagged as ramp jumps if
+  the ramp jump is above the requested threshold.  Two thresholds and times are
```
cshanahan1, line 69: what does 'two threshold and times' mean here?

jwst/jump/jump.py
```diff
@@ -11,7 +11,11 @@
 def run_detect_jumps(input_model, gain_model, readnoise_model,
                      rejection_thresh, three_grp_thresh, four_grp_thresh,
                      max_cores, max_jump_to_flag_neighbors,
-                     min_jump_to_flag_neighbors, flag_4_neighbors):
+                     min_jump_to_flag_neighbors, flag_4_neighbors,
+                     after_jump_flag_dn1=0.0,
```
cshanahan1, line 15: change param name to match stcal

jwst/jump/jump_step.py
```diff
@@ -22,6 +22,10 @@ class JumpStep(Step):
         flag_4_neighbors = boolean(default=True) # flag the four perpendicular neighbors of each CR
         max_jump_to_flag_neighbors = float(default=1000) # maximum jump sigma that will trigger neighbor flagging
         min_jump_to_flag_neighbors = float(default=10) # minimum jump sigma that will trigger neighbor flagging
+        after_jump_flag_dn1 = float(default=0) # 1st flag groups after jump above DN threshold
```
cshanahan1, line 25: change names and descriptions to match stcal

jwst/jump/jump.py
cshanahan1, line 15: also you don't need to set the defaults here since they are set in 'spec'

jwst/jump/jump_step.py
```diff
@@ -22,6 +22,10 @@ class JumpStep(Step):
...
+        after_jump_flag_time1 = float(default=0) # 1st flag groups after jump groups within specified time
```
cshanahan1, line 26: should be int

jwst/jump/jump_step.py
```diff
@@ -22,6 +22,10 @@ class JumpStep(Step):
...
+        after_jump_flag_dn2 = float(default=0) # 2nd flag groups after jump above DN threshold
+        after_jump_flag_time2 = float(default=0) # 2nd flag groups after jump groups within specified time
```
cshanahan1, line 28: should also be int

jwst/jump/jump_step.py
```diff
@@ -70,7 +78,11 @@ def process(self, input):
             result = run_detect_jumps(input_model, gain_model, readnoise_model,
                                       rej_thresh, three_grp_rej_thresh, four_grp_rej_thresh, max_cores,
                                       max_jump_to_flag_neighbors, min_jump_to_flag_neighbors,
-                                      flag_4_neighbors)
+                                      flag_4_neighbors,
```
cshanahan1, line 81: change param names

jwst/jump/jump.py
```diff
@@ -43,7 +53,11 @@ def run_detect_jumps(input_model, gain_model, readnoise_model,
                                     four_grp_thresh, max_cores,
                                     max_jump_to_flag_neighbors,
                                     min_jump_to_flag_neighbors,
-                                    flag_4_neighbors, dqflags.pixel)
+                                    flag_4_neighbors, dqflags.pixel,
+                                    after_jump_flag_dn1=after_jump_flag_dn1,
```
hbushouse, line 57: Given that all of the function arguments are specified in this call, in the proper order, I don't think it's really necessary to explicitly assign the values to each argument name for the last 4 (new) arguments. Could simply it to just listing the variables that are being passed, i.e. (..., dqflags.pixel, after_jump_flag_dn1, after_jump_flag_n1, ...)

jwst/jump/jump_step.py
hbushouse, line 81: Same  comment as above: I don't think it's necessary to equate all the new values to argument names, just pass them in the correct order. Of course it doesn't hurt to do it this way; just takes up more real estate in the code.

jwst/jump/jump_step.py
hbushouse, line 26: I'm wondering about how the default case (everything set to zero) ends up getting handled in the lower-level functions. If the time arguments are zero, that will equate to 0 groups getting flagged, but a DN level of zero (which gets converted to zero electrons) will cause all jumps to be above the threshold and hence a whole lot of pixels could end up getting calculations done on them only to find that 0 groups get flagged for them. Is there something down in the lower-level code that checks for a threshold of zero before bothering to do the calculations?

cshanahan1: i was also wondering this. i think a check in stcal for all 0 values should be done to skip this, and make it clear that its being skipped

docs/jwst/jump/description.rst
cshanahan1: i figured this out based on context, but rephrase since you're not using 'time' anymore

jwst/jump/jump.py
hbushouse: I think the confusion here is that it's setup intentionally for the user to give these argument values in units of DN, because that's what they see in the ramp data files, but because the lower-level routines in stcal work on the data after they've been (temporarily) converted to units of electrons, these argument values also get converted down in those routines and hence the name change. At the top levels they're DN, but in the lower level stcal functions they're electrons. So this apparent inconsistency in the naming is OK (IMHO).

docs/jwst/jump/description.rst
hbushouse: Yes, the user is still giving the argument values in units of time, but then those are converted into the equivalent number of groups down in the lower-level functions.

docs/jwst/jump/arguments.rst
hbushouse: User gives these values in units of DN. They get converted internally to electrons for use in the stcal functions.

jwst/jump/jump_step.py
karllark: This is done in stcal.  If time[1,2] is zero, then skip that after flagging.

karllark: To be clear if n[1,2] is zero this happens.  As time[1,2] -> n[1,2] via the group time.

jwst/jump/jump.py
cshanahan1: ok i see this now - i will remove my comments about this.

the input to detect_jumps in stcal is DN, its converted there to e-, and the input to find_crs in then electrons. i would vote for doing the conversion in side find_crs to keep all the units the same. i just see this possibly getting confusing when setting up tests that run the full step vs. calling find_crs directly.

docs/jwst/jump/description.rst
cshanahan1: ok, i guess i don't see the benefit of supplying time over just giving group number, but if this is more intuitive to users then thats fine

karllark: Giving it in time means that we don't need to supply different values of the number of groups to flag for different subarrays.  The transient seen after the ramp jumps is seen and expected to be time dependent only.  Hence, this make sense and makes it easier to implement.

jwst/jump/jump.py
karllark: Like in stcal, the defaults are needed for the tests to run.  Otherwise, these 4 parameters need to be set in all the tests.

jwst/jump/jump.py
```diff
@@ -11,7 +11,11 @@
 def run_detect_jumps(input_model, gain_model, readnoise_model,
                      rejection_thresh, three_grp_thresh, four_grp_thresh,
                      max_cores, max_jump_to_flag_neighbors,
-                     min_jump_to_flag_neighbors, flag_4_neighbors):
+                     min_jump_to_flag_neighbors, flag_4_neighbors,
+                     after_jump_flag_dn1=0.0,
+                     after_jump_flag_time1=0,
```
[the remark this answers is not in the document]
karllark: Good point.  Done.

jwst/jump/jump.py
```diff
@@ -23,6 +27,12 @@ def run_detect_jumps(input_model, gain_model, readnoise_model,
     err = input_model.err
     output_model = input_model.copy()
 
+    # determine the number of groups that coorespond to the after_jump times
```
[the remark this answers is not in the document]
karllark: Fixed.

jwst/jump/jump.py
```diff
@@ -23,6 +27,12 @@ def run_detect_jumps(input_model, gain_model, readnoise_model,
...
+    # needed as the group time is not passed to detect_jumps
```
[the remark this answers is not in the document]
karllark: Fixed.

jwst/jump/jump.py
karllark: Good point.  Fixed.

jwst/jump/jump_step.py
karllark: There are tests for this behavior in stcal.

jwst/jump/jump_step.py
karllark: Fixed.

jwst/jump/jump.py
karllark: I did the conversion at this point as it is where the data is converted to electrons.  Seemed like the consistent place to do it.

hbushouse (approved):
LGTM. All review comments have been addressed.

hbushouse:
@cshanahan1 are you good with this now that all review comments have been addressed?

hbushouse:
@karllark I think this PR branch may need to be rebased from master. Looks like conflicts are preventing the CI tests from running.

karllark:
> @karllark I think this PR branch may need to be rebased from master. Looks like conflicts are preventing the CI tests from running.

Done.

cshanahan1 (approved):

hbushouse:
I think the remaining CI failures, which are confined to ramp_fitting unit tests, are due to other PR's included in the stcal 1.1.0 release, and maybe fixed by #6949 once it gets merged. Retesting that PR right now.

hbushouse:
OK, #6949 has now been merged to update the truth values in the unit tests that were causing CI failures. So one more rebase to kickoff the tests again.

PR #6946, opened 2022-07-28
JP-2686: Support user-supplied reference files for stage 2 of WCS alignment

mcara:
Resolves [JP-2686](https://jira.stsci.edu/browse/JP-2686)

This PR add support for user-specified reference astrometric catalogs to be used _in place of_ GAIADR1 or GAIADR2 catalogs.

**Checklist**
- [x] added relevant label(s)


docs/jwst/tweakreg/README.rst
```diff
@@ -52,6 +52,13 @@ The results of this one fit then gets back-propagated to all the
 input images to align them all to the astrometric reference frame while
 maintaining the relative alignment between the images.
 
+For this part of alignment, instead of 'GAIADR1' or 'GAIADR2', users can
+supply an external reference catalog by providing path to an existing
```
hbushouse, line 56: typo: change "providing path" to "providing a path"

hbushouse (commented):
Looks straightforward enough. Just one minor comment regarding a typo.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -250,17 +250,35 @@ def process(self, input):
...
             else:
                 output_name = None
-            ref_cat = amutils.create_astrometric_catalog(images,
-                                                         self.gaia_catalog,
-                                                         output=output_name)
+
+            self.gaia_catalog = self.gaia_catalog.strip()
+            gaia_cat_name = self.gaia_catalog.upper()
+
+            if gaia_cat_name in ['GAIADR2', 'GAIADR1']:
```
stsci-hack, line 257: Actually, for HST, we are already using 'GAIAeDR3' as the most up-to-date catalog.  It may be easier to maintain/update, if the list of supported pre-defined catalogs were defined as a module level (global?) variable that gets referenced here.  That makes it more obvious what catalogs are supported and adding to that list in one place becomes easier when new catalogs are supported/used.

jwst/tweakreg/tweakreg_step.py
```diff
@@ -52,7 +68,7 @@ class TweakRegStep(Step):
         nclip = integer(min=0, default=3) # Number of clipping iterations in fit
         sigma = float(min=0.0, default=3.0) # Clipping limit in sigma units
         align_to_gaia = boolean(default=False)  # Align to GAIA catalog
```
hbushouse, line 70: The more I look at all of these arguments I wonder if we need large-scale change in their names (similar to what some of us were discussing in a meeting yesterday). If we're now allowing `gaia_catalog` to point to any user-supplied catalog, then it just doesn't seem to make sense to me to keep the name "gaia" in all the other arguments that control how that catalog gets used. Why should the user specify ``align_to_gaia=True`` when they're aligning to their own catalog (not Gaia). Similarly for `min_gaia` and `save_gaia_catalog` (the latter of which probably isn't even needed or useful when the user is supplying the catalog to begin with - why save it again for them?).

The existing changes in the PR certainly make it possible, in the short term, for a user to specify their own catalog, but in long term I think it needs more thought and rework. What new param names like `align_to_refcat`, `ref_catalog`, `min_refcat`?

@stsci-hack @nden @mairanteodoro any thoughts?

mcara: Yes, that is what @mairanteodoro and I were proposing: to get rid of `gaia` in parameter names. It is just that the work on exposing hidden parameters for the alignment of all images *at once* is work that @mairanteodoro is doing. Once that work is done, we should have a separate PR that deprecates the use of old parameters and introduces new parameters. I do not think this should be done here, unless you insist.

I would actually suggest that we get rig of `align_to_gaia(or anything)` altogether and simply specifying a reference catalog (user-supplied or `'GAIADR2'`) in `abs_refcat` would signal to run the absolute alignment.

hbushouse (approved):

PR #6948, opened 2022-07-28
Add ModelList to ModelContainer parents

tapastro:
This PR adds a parent ABC to ModelContainer, so that behavior in stpipe can be altered if step/pipeline input is a ModelContainer rather than a JwstDataModel.

Need to wait for stdatamodels release containing ModelList addition (currently expected to be 0.4.4.)

**Checklist**
- [x] added relevant label(s)


stscieisenhamer (commented):
Question: Should `ModelList` be the one to inherit `Sequence`? May be an executive decision, but it may make things cleaner to read, and enforce the behavior.

tapastro:
I wrote this thinking Sequence existed only as a 'marker' parent class, i.e. we weren't using any of its functionality in ModelContainer. Do you know if this is true or not? 

I tried to verify by putting all of my PRs in stpipe/stdatmodels/jwst in one environment, and while I don't see any errors, there could be an error in how I put all of the new code in the conda environment... 🤷

stscieisenhamer:
Ah, if `Sequence` is only a marker also, then should it be removed completely, since that is what `ModelList` is supposed to do?

tapastro:
> Ah, if `Sequence` is only a marker also, then should it be removed completely, since that is what `ModelList` is supposed to do?

I just wrote out a message saying that's what my PRs intended to do, then checked to see I left Sequence in the list of parent classes... I'll have to run the local tests again to see if Sequence is necessary.

jwst/datamodels/container.py
```diff
@@ -26,7 +26,7 @@
 logger.addHandler(logging.NullHandler())
 
 
-class ModelContainer(JwstDataModel, Sequence):
+class ModelContainer(JwstDataModel, Sequence, ModelList):
```
stscieisenhamer, line 29: Ahh, here is where `Sequence` is still present. As per offline discussion, since `Sequence` was just a marker, I think it should be removed from here.

stscieisenhamer (changes requested):
changelog

stscieisenhamer (approved):

stscieisenhamer:
This PR is still marked as draft. When ready, please rebase and un-draft.

tapastro:
This is no longer marked as draft, but it does rely on a new stdatamodels build before merging can be performed.

stscieisenhamer:
is this still relevant?

tapastro:
No longer needed - ModelLibrary is here to save the day!

PR #6953, opened 2022-07-29
Add details on when crds environmentals be defined

stscieisenhamer:
This PR clarifies when CRDS environmental variables should be defined when using Python scripts.

**Checklist**
- ~updated or added relevant tests~
- [x] added relevant label(s)

### DIFF
--- a/docs/jwst/introduction.rst
+++ b/docs/jwst/introduction.rst
@@ -461,14 +461,14 @@ see :ref:`Execute via call()<call_examples>`::
 For more details on the different ways to run a pipeline step, see
 the :ref:`Configuring a Step<configuring-a-step>` page.
 
-CRDS Environmentals
--------------------
+CRDS Environment Variables
+--------------------------
 
-The CRDS environmental variables need to be defined *before* importing anything
-from `jwst` or `crds`. In general, any scripts should assume the environmentals
-have been set before the scripts have run. If one needs to define the CRDS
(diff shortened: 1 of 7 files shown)


tapastro (approved):
LGTM. I hope people read far enough down the page to see this! 😬

stscieisenhamer:
About reading down to this point: Ya, true. This is the first point that the "in a python script" context is discussed. Placing it anywhere else would be out-of-context, I think.

tapastro:
> About reading down to this point: Ya, true. This is the first point that the "in a python script" context is discussed. Placing it anywhere else would be out-of-context, I think.

I agree - my first thought was to push it up near the introduction of the CRDS environment variables, but you don't need to worry about this for CLI usage, so I think you chose the best spot.

docs/jwst/introduction.rst
```diff
@@ -461,6 +461,25 @@ see :ref:`Execute via call()<call_examples>`::
 For more details on the different ways to run a pipeline step, see
 the :ref:`Configuring a Step<configuring-a-step>` page.
 
+CRDS Environmentals
+-------------------
+
+The CRDS environmental variables need to be defined *before* importing anything
```
nden, line 467: I always used `environment` as opposed to `environmental` variable. Are they interchangeable?

nden (approved):

PR #6954, opened 2022-07-30
JP-2707: Fix excessive zeroes in resampled variance

mcara:
This PR fixes a bug in how resampled variances are computed due to which resampled error array tends to have an excessive number of zero-valued pixels. By this I mean that even though there were multiple images contributing to one output pixel in SCI, the corresponding pixel in the ERR extension was 0. The reason for this was that current algorithm computes a sum of inverse variances from each resampled image. If variance of a pixel in an input image is 0 then `1/variance` would be `np.inf`. Even if other input images have valid pixels and contribute to the output pixel, their variances, even if finite and >0 cannot clear `np.inf`. In the end, the final error array computed as `1/sum(1/variances)` resulting in the final value being 0 (for resampled variance). That is, something like this would happen for a single pixel:

```
1/(1/0 + 1/finite2a + 1/finite3a + ...) = 1/(inf + finite2b + finite 3b + ...) = 1/inf = 0
```

This PR fixes this bug by tracking more carefully individual pixel contribution and removing from computations pixels with zero weights or zero variance.

The issue of excessive number of zero-valued pixels was reported in the help desk ticket INC0180024.

This PR does not address the validity of resampled variance or error computations in itself - only the excessive zeroes issue.

Resolves [JP-2707](https://jira.stsci.edu/browse/JP-2707)

**Checklist**
- [x] added relevant label(s)

### DIFF
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -281,15 +281,16 @@ class ResampleData:
             outwht = np.zeros_like(output_model.data)
             outcon = np.zeros_like(output_model.con)
 
-            # Resample the variance array.  Use fillval=np.inf so that when we
-            # take the reciprocal for summing, it is zero where there is zero weight
+            # Resample the variance array. Fill "unpopulated" pixels with NaNs.
             self.drizzle_arrays(variance, inwht, model.meta.wcs,
                                 output_wcs, resampled_variance, outwht, outcon,
                                 pixfrac=self.pixfrac, kernel=self.kernel,
                                 fillval=np.nan)
 
-            # Add the inverse of the resampled variance to a running sum
-            mask = (outwht > 0) & (resampled_variance > 0)
+            # Add the inverse of the resampled variance to a running sum.
+            # Update only pixels (in the running sum) with valid new values:
+            mask = np.logical_and(outcon, resampled_variance > 0)
+
             inverse_variance_sum[mask] = np.nansum(
                 [inverse_variance_sum[mask], np.reciprocal(resampled_variance[mask])],
                  axis=0
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -219,8 +219,16 @@ class ResampleData:
         self.resample_variance_array("var_rnoise", output_model)
         self.resample_variance_array("var_poisson", output_model)
         self.resample_variance_array("var_flat", output_model)
-        output_model.err = np.sqrt(output_model.var_rnoise + output_model.var_poisson
-                                   + output_model.var_flat)
+        output_model.err = np.sqrt(
+            np.nansum(
+                [
+                    output_model.var_rnoise,
+                    output_model.var_poisson,
+                    output_model.var_flat
+                ],
+                axis=0
+            )
+        )
 
         # TODO: The following two methods and calls should be moved upstream to
         # ResampleStep and ResampleSpecStep respectively
@@ -243,9 +251,7 @@ class ResampleData:
         This modifies output_model in-place.
         """
         output_wcs = output_model.meta.wcs
-        inverse_variance_sum = np.zeros_like(output_model.data)
-
-        tmask = None
+        inverse_variance_sum = np.full_like(output_model.data, np.nan)
 
         log.info(f"Resampling {name}")
         for model in self.input_models:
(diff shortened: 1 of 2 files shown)


mcara:
I would expect regression tests to fail with reported differences in the resampled ERR and VAR_* extensions of the form:
```
0 (OLD) -> something finite != 0 (NEW)
```
Old finite non-zero values should not change.

jwst/resample/resample.py
```diff
@@ -275,17 +286,19 @@ def resample_variance_array(self, name, output_model):
             self.drizzle_arrays(variance, inwht, model.meta.wcs,
```
perrygreenfield, line 286: The previous comment is now wrong.

jwst/resample/resample.py
```diff
@@ -275,17 +286,19 @@ def resample_variance_array(self, name, output_model):
...
                                 output_wcs, resampled_variance, outwht, outcon,
                                 pixfrac=self.pixfrac, kernel=self.kernel,
-                                fillval=np.inf)
+                                fillval=np.nan)
 
             # Add the inverse of the resampled variance to a running sum
-            with np.errstate(divide="ignore"):
-                inverse_variance_sum += np.reciprocal(resampled_variance)
+            mask = (outwht > 0) & (resampled_variance > 0)
```
perrygreenfield, line 292: To save me rummaging through other code can you say where `outwht` gets set to non-zero values? Presumably in the above `self.drizzle_arrays` call?

perrygreenfield (commented):
Code comment error and a question about where something happens.

jwst/resample/resample.py
mcara: I simplified mask computation. Checking for `resampled_variance > 0` should be enough.

mcara:
Style errors are unrelated to my PR and I do not understand why test is testing non-my-pr files.

perrygreenfield (approved):
LGTM

mcara:
@hbushouse @stscieisenhamer Why is the code style check failing on this PR?

perrygreenfield:
I believe Howard is on vacation. You might motivate him with a donut.

mcara:
:doughnut:

stscieisenhamer:
The code style checks are real. They are all instances of `if(condition)` and it should be `if (condition)`; there is no space between the `if` and the `(condtion)` in parenthesis.

Very odd since none of these changes are in this pr,  but all of these should be fixed. It looks like someone did a mass replacement that replaced more than intended.

mcara:
> The code style checks are real. They are all instances of `if(condition)` and it should be `if (condition)`; there is no space between the `if` and the `(condtion)` in parenthesis.
> 
> Very odd since none of these changes are in this pr, but all of these should be fixed. It looks like someone did a mass replacement that replaced more than intended.

I know they are real, but is it _me_ who has to fix these issues _in this PR_? I do not think I should be fixing these errors: they all come from many different PRs created by various users (some PRs are > 1 year old). IMO, if you want to have this fixed, it should be done in a separate style-fix PR.

It is odd that tests report errors in files not modified by this PR. I do not think this should be happening.

hbushouse:
> [🍩 🍩 🍩]` `[🍩 🍩 🍩 🍩]` ` [🍩 🍩 🍩]`*`[🍩 🍩 🍩 🍩]`=`🍩 [🍩 🍩 🍩]` `[🍩 🍩 🍩 🍩]` ` [🍩 🍩 🍩]
> 
> @hbushouse

So if I eat two arrays of 12 donuts each, it's equivalent to only 1 donut?  Cool!

mcara:
@hbushouse I fixed the math. Alignment is harder.

mcara:
I meant
```python
np.dot(np.full((4, 3), np.nan), np.full((3, 4), np.nan))
```

mcara:
Should I close this PR and open another one an hope that magic will happen and that the PEP8 test will not test files not modified by my PR?

stscieisenhamer:
The PEP8 tests everything, so re-opening the PR will not change that. I will merge the PR and open a new issue to fix the PEP8 issues.

stscieisenhamer:
Going back through blame, this has exists for 4+ years. There must be a change in flake8 or its CI configuration causing this check now; a good catch even if it is a bit late.

mcara:
> The PEP8 tests everything, so re-opening the PR will not change that.

Not true: other open PRs, such as https://github.com/spacetelescope/jwst/pull/6955, pass all tests. Why CI is not testing _all_ files for those PRs and only for this one?

hbushouse:
Check with @zacharyburnett as to whether something may have changed recently in the flake8 setup or use in the CI tests.

PR #6964, opened 2022-08-05
JP-2819: Populate wavelength extension for MIRI LRS slitless

### LINKED ISSUE #6963: No wavelength solution for MIRI/LRS in _calints files after Spec2Pipeline
As of right now, the WAVELENGTH extension of all my MIRI slitless LRS _calints files (point source, TSO) is completely filled with zeros. I believe this should be fairly easy to populate (and the extract1d step does calculate the wavelength solution for the _x1dints files) and don't see why it couldn't be computed given that the extension is populated for other instruments. I have skipped the photom step and can't say for sure yet that the wavelength solution is still missing if the photom step is run, but running the photom step should not be required to get a wavelength solution. I believe it should be happening in the assign_wcs step.

tapastro:
Resolves [JP-2819](https://jira.stsci.edu/browse/JP-2819)

Closes #6963 

This PR adds a function to assign_wcs to populate the wavelength extension of the calints level 2 output for MIRI LRS slitless data. 

Not sure if this is addressing pixel locations appropriately (0-indexed) or if this addition is warranted - would like review on this approach.

**Checklist**
- [x] added relevant label(s)


hbushouse (commented):
Looks OK to me, but @nden should have the definitive word.

jwst/assign_wcs/util.py
```diff
@@ -1420,3 +1420,16 @@ def update_fits_wcsinfo(datamodel, max_pix_error=0.01, degree=None, npoints=32,
...
+
+
+def populate_wavelength_extension(model):
+    """For MIRI LRS Slitless, no step currently
+    assigns values to model.wavelength - do it here
+    """
+    data_shape = model.data.shape
+    if len(data_shape) > 2:
+        data_shape = data_shape[-2:]
+    index_array = np.indices(data_shape[::-1])
```
nden, line 1432: I think the `x`, `y` inputs should be generated from the `bounding_box`, because I believe these need to be in full frame coordinates. The way to do this is
```

x, y = grid_from_bounding_box(data_model.wcs.bounding_box)
```

tapastro: This is the sort of thing I expected to miss - thanks for the heads up! I get a similar solution with the new code, but the coords did shift a small amount.

nden: `x`, and `y` should be integers (the center of the pixel). Is this what you get from `grid_from_bounding_box`?

jwst/assign_wcs/util.py
```diff
@@ -1420,3 +1420,17 @@ def update_fits_wcsinfo(datamodel, max_pix_error=0.01, degree=None, npoints=32,
...
+def populate_wavelength_extension(model):
+    """For MIRI LRS Slitless, no step currently
+    assigns values to model.wavelength - do it here
+    """
+    data_shape = model.data.shape
+    if len(data_shape) > 2:
+        data_shape = data_shape[-2:]
+
+    bbox_grid = grid_from_bounding_box(model.meta.wcs.bounding_box)
+    model.wavelength = model.meta.wcs(*bbox_grid)[2]
```
hbushouse, line 1434: It suddenly occurs to me, does the `wavelength` attribute even exist yet in the datamodel at this point in processing for LRS slitless data? I believe it comes into the spec2 pipeline as a `CubeModel` and then sometime later gets transformed to a `SlitModel`?  Is the wavelength attribute defined for a `CubeModel`?

hbushouse: Never mind. Just answered my own question: https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/schemas/cube.schema.yaml#L39

Yep, it does exist.

jwst/assign_wcs/util.py
tapastro: Yes, the return works as you expected - x is a 2d array of x-positions, while y is a 2d array of y-positions, such that x[i] and y[i] specify a given pixel. They may be of type float (I checked yesterday but can't remember now), but the values are ints.

nden (approved):

hbushouse (approved):

hbushouse:
Rats, should've asked if you had run regtests on this before I merged it. I'm doing a run now and a handful of errors have already shown up in the log: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/1938/console

hbushouse:
Yes, these are not failures/differences (which I would expect), these are ERRORs.

hbushouse:
I'm running test_miri_lrs_slitless locally right now to see if that's the one that's going bad.

hbushouse:
Yep, the LRS slitless test is now crashing in the photom step. I wonder if something in the photom step was also creating the 2D wavelength array and it doesn't need to do that anymore, since it should already be available in the input.

tapastro:
The change from the initial commit to the use of the bbox changed the wavelength array shape from the equivalent shape of the data to the shape of the bbox, which is smaller than the data array. This is causing a mismatch when applying an operation to the wavelength array.

tapastro:
@nden Looking further into the relative indices, it appears that the use of bbox serves to truncate the first method to only include the non-nan extent - i.e. the bbox is working as expected, but it is a mismatch to the data array shape. Was the intent of the bbox use (i.e. the full frame coordinates mentioned above) meant to include the location on the full frame instead of the subarray? I tried to offset the indices generated in the first method by the xstart and ystart values defined in meta.subarray, but there is no defined wcs for those pixel positions.

tapastro:
FWIW, bbox_grid appears to be in the same pixel-space as the original index_array, but due to the truncation of the original space to exclude fully-nan regions, the extent of the bbox grid is x:{4:71}, y:{5:394}.

PR #6967, opened 2022-08-06
Revert Fourier Transform code to avoid Poppy's recently updated sign convention (JP-2618)

dmggh:
Poppy was recently updated to change the sign convention used in exponents in the Fourier transform calculation. This update reverts the ami_analyze code to use an earlier code, in the matrix_dft.py module.

Resolves [JP-2618](https://jira.stsci.edu/browse/JP-2618)

**Checklist**
- [X] added relevant label(s)

### DIFF
--- a/jwst/ami/matrix_dft.py
+++ b/jwst/ami/matrix_dft.py
@@ -225,12 +225,6 @@ class MatrixFourierTransform:
         Deprecated. Use poppy.conf.default_logging_level to set DEBUG level
         logging.
 
-    Example
-    -------
-    >>> mft = MatrixFourierTransform()
-    >>> result = mft.perform(pupil, nlamD, npix, offset)
-
-
     History
     -------
     Code by Sivaramakrishnan based on Soummer et al.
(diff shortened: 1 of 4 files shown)


jwst/ami/matrix_dft.py
```diff
@@ -0,0 +1,352 @@
...
+        * SYMMETRIC spreads the zero-order term evenly between the center
+          four pixels
+    verbose : bool
+        Deprecated. Use poppy.conf.default_logging_level to set DEBUG level
+        logging.
+
+    Example
+    -------
+    >>> mft = MatrixFourierTransform()
+    >>> result = mft.perform(pupilArray, focalplane_size, focalplane_npix)
```
hbushouse, line 231: This line appears to be killing all of the CI tests, due to a problem with building the doc string. The argument names don't match what's used in the actual function definition.

hbushouse:
A regression test run against this branch shows what I assume are expected differences in the ami_analyze outputs (https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/372/#showFailuresLink) and the one unit test failure due to the doc string problem that's also killing the CI tests here. So once that unit test gets fixed, this should be good to go.

hbushouse (approved):
Looks good now. Will merge when some other tests of previous mergers have completed.

PR #6957, opened 2022-08-01
JP-2654  Fix bug in NIRSPEC DQ flagging  and JP-2662 for MIRI moving target

jemorrison:
Resolves [JP-2654](https://jira.stsci.edu/browse/JP-2654)
Resolves [JP-2662](https://jira.stsci.edu/browse/JP-2662)

This PR addresses a bug in the C code used by cube_build that sets the DQ flag for NIRSpec IFU data. After a review of the code it was determined the flagging was not being done correctly. I re-wrote the code to make it easier to follow. In the process of re-writing the code I fixed the segmentation fault error. 
When the above bug was fixed, a second bug was discovered. The NIRSpec data for this ticket is moving target data. This second bug was because drizzling was not working correctly for moving target data.  This was fixed and this fix resolved JP-2662

**Checklist**
- [X] added entry in `CHANGES.rst` within the relevant release section
- [ ] updated or added relevant tests
- [ ] updated relevant documentation
- [X] added relevant milestone
- [X] added relevant label(s)

PR #6971, opened 2022-08-10
JP-2654 JP-2662: Fix bug in NIRSpec IFU DQ flagging and handling of moving target exposures

jemorrison:
Resolves [JP-2654](https://jira.stsci.edu/browse/JP-2654)
Resolves [JP-2662](https://jira.stsci.edu/browse/JP-2662)

This PR is a duplicate of [PR 6957](https://github.com/spacetelescope/jwst/pull/6957). That PR was made before all the flake8 changes. There were too many conflicts once that PR was rebased with master, so I pull in the relevant changes from that PR and put them here.

This PR addresses a bug in the C code used by cube_build that sets the DQ flag for NIRSpec IFU data. After a review of the code it was determined the flagging was not being done correctly. I re-wrote the code to make it easier to follow. In the process of re-writing the code I fixed the segmentation fault error. 

When the above bug was fixed, a second bug was discovered. The NIRSpec data for this ticket is moving target data. This second bug was because drizzling was not working correctly for moving target data.  This was fixed, which resolved JP-2662

**Checklist**
- [X] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,12 @@ cube_build
 - Remove trailing dash from IFU cube filenames built from all subchannels.
   Sort subchannels present by inverse alphabetical order to ensure
   consistent filename creation across processing runs. [#6959]
+  
+- Re-wrote c code for NIRSpec dq flagging. In addition,if data is
+  moving_target data, then do not use s_region values in cal files,
+  but map all the pixels to sky to find cube footprint. Updated drizzle
+  code to use wcs of output frame to account for moving target data 
+ 
 
 datamodels
 ----------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,7 +35,7 @@ cube_build
 - Re-wrote c code for NIRSpec dq flagging. In addition,if data is
   moving_target data, then do not use s_region values in cal files,
   but map all the pixels to sky to find cube footprint. Updated drizzle
-  code to use wcs of output frame to account for moving target data 
+  code to use wcs of output frame to account for moving target data  [#6971]
  
 
 datamodels
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,10 +32,12 @@ cube_build
   Sort subchannels present by inverse alphabetical order to ensure
   consistent filename creation across processing runs. [#6959]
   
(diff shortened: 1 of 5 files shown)


CHANGES.rst
```diff
@@ -31,6 +31,12 @@ cube_build
 - Remove trailing dash from IFU cube filenames built from all subchannels.
   Sort subchannels present by inverse alphabetical order to ensure
   consistent filename creation across processing runs. [#6959]
+  
+- Re-wrote c code for NIRSpec dq flagging. In addition,if data is
```
hbushouse, line 35: I think these are two distinct issues and hence would be nice if they were listed as separate entries in the change log (just so the 2nd one doesn't go unnoticed). It's OK to have to entries reference the same PR number.

jwst/cube_build/cube_build_wcs_util.py
```diff
@@ -153,7 +153,7 @@ def find_corners_NIRSPEC(input, instrument_info, coord_system):
     lambda_slice = np.zeros(nslices * 2)
     k = 0
     # for NIRSPEC there are 30 regions
-    log.info('Looping over slices to determine cube size .. this takes a while')
+    log.info('Looping over slices to determine cube size')
```
hbushouse, line 156: So it's faster now? ;-)

hbushouse (commented):
Looks good. Just one minor comment on the change log entry.

jwst/cube_build/cube_build_wcs_util.py
jemorrison: Yes it is faster. But that was work that was on the wcs side of things. I think William might have worked on it

hbushouse (approved):

hbushouse:
Looks good. Merging.

hbushouse:
It appears that a workaround for the C compiler problems we're having with this PR would be to simply change the 3 lines in cube_dq_utils.c that declare a datatype for a loop variable within the "for" statement itself (e.g. "for (int i = 0, ....)" so that the variable is declared outside of the "for" statement. The lines are 349, 392, and 637. That would be a fairly simple fix.

And these are the only places in the C functions that declare datatype for an iterator within the "for" statement itself. All others are declared outside of the loop. That's why we haven't seen this error before.

PR #6983, opened 2022-08-15
JP-2816: Update outlier docs

stsci-hack:
Resolves [JP-2816](https://jira.stsci.edu/browse/JP-2816)
Resolves [JP-2820](https://jira.stsci.edu/browse/JP-2820)

This PR updates the documentation related to the memory model implemented under #6904 for [JP-2527](https://jira.stsci.edu/browse/JP-2527).  The changes include:
  - Adding a section to the outlier_detection algorithm discussion on the overall memory model. 
  - Updated the ModelContainer docstring to add Notes describing the optional parameters defined and used for the outlier_detection memory improvements
  - Added a Note to the ResampleStep docstring to describe the new ``in_memory`` parameter.
  - Updates initialization of the ModelContainter based on ``in_memory`` parameter. [JP-2820]

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -68,7 +68,7 @@ outlier_detection
   New parameters were aded to the step to control this functionality. [#6904]
 
 - Updated documentation of memory model and new parameters for memory use in
-  outlier_detection and resample steps. [#xxxx]
+  outlier_detection and resample steps. [#6983]
 
 resample
 --------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -70,6 +70,8 @@ outlier_detection
 - Updated documentation of memory model and new parameters for memory use in
   outlier_detection and resample steps. [#6983]
 
+- Fix reading of the source_type attribute for NIRSpec IFU data. [#6980]
+
 resample
 --------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -67,6 +67,9 @@ outlier_detection
   input ``ImageModels`` that are saved to disk instead of keeping them in memory.
   New parameters were aded to the step to control this functionality. [#6904]
 
+- Updated documentation of memory model and new parameters for memory use in
+  outlier_detection and resample steps. [#xxxx]
+
 resample
 --------
 
--- a/docs/jwst/outlier_detection/outlier_detection.rst
+++ b/docs/jwst/outlier_detection/outlier_detection.rst
@@ -68,14 +68,16 @@ final output product.  Specifically,
       be on the order of 2Gb or more.
 
     * The median combination step then needs to have all pixels at the same position on
-      the sky in memory in order to perform the median computation.  The simplest implmentation
-      for this step would require keeping all resampled outputs in memory at the same time.
+      the sky in memory in order to perform the median computation.  The simplest implementation
+      for this step requires keeping all resampled outputs fully in memory at the same time.
 
 Many Level 3 products only include a modest number of input exposures which can be
-processed using less than 32Gb of memory at a time.  However, given the number of
-ways this memory limit can be exceeded, the overall memory model for the outlier detection
-was revised to minimize the memory usage at the expense of file I/O with limited control
-over this through the ``in_memory`` parameter when set to `False`.
+processed using less than 32Gb of memory at a time.  However, there are a number of
+ways this memory limit can be exceeded.  This has been addressed by implementing an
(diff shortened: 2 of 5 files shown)


docs/jwst/outlier_detection/outlier_detection.rst
```diff
@@ -50,6 +50,55 @@ Specifically, this routine performs the following operations:
...
+
+    * The initial resample step creates an output product for EACH input that is the
+      same size as the final
+      output product, which for imaging modes can span all chips in the detector while
+      also accounting for all dithers.  For some Level 3 products, each resampled image can
+      be on the order of 2Gb or more.
+
+    * The median combination step then needs to have all pixels at the same position on
+      the sky in memory in order to perform the median computation.  The simplest implmentation
+      for this step would require keeping all resampled outputs in memory at the same time.
```
hbushouse, line 72: I'm not sure I'd say "would require". It *does* require keeping all outputs in memory and that's the way the initial implementation has already been done, which is still available to use (it is still available, right?). So I would say "... for this step requires keeping all ..."

docs/jwst/outlier_detection/outlier_detection.rst
```diff
@@ -50,6 +50,55 @@ Specifically, this routine performs the following operations:
...
+
+Many Level 3 products only include a modest number of input exposures which can be
+processed using less than 32Gb of memory at a time.  However, given the number of
+ways this memory limit can be exceeded, the overall memory model for the outlier detection
+was revised to minimize the memory usage at the expense of file I/O with limited control
```
hbushouse, line 77: It's true that if folks are interested in the history of the step implementation we should say it was revised. But I think most folks only care about the current state, so perhaps we should just say "the overall memory model for the outlier detection includes options to minimize the memory usage ..." (or something like that).

docs/jwst/outlier_detection/outlier_detection.rst
```diff
@@ -50,6 +50,55 @@ Specifically, this routine performs the following operations:
...
+over this through the ``in_memory`` parameter when set to `False`.
```
hbushouse, line 78: Perhaps break this line up so that there's just a short intro to the info contained in the following bullet list, e.g. "... through the use of the in_memory parameter. When in_memory=False: "

stsci-hack:
The change in b4f30a4 to apply the ``in_memory`` parameter to the ModelContainer initialization was tested using the data specified in the original ticket JP-2527.  The results of using ``in_memory=True`` and ``in_memory=False`` were then compared and there were no pixel differences at all in the SCI or VAR extensions, and only expected differences in the ASDF extension.  Additionally, 'top' was used to track the memory use and runtime during each run to get:

in_memory            max memory                   runtime (minutes)
--------------         ---------------                  --------------------
True                       6.3g                                 3:16.56
False                      5.4g                                 5:43.76

The memory usage does not appear to be improved as much as expected due to how I ran the test interactively.  These results do confirm the expected behavior, though, with in_memory requiring more memory but running faster.

docs/jwst/outlier_detection/outlier_detection.rst
stsci-hack: Good point.  Revised accordingly in 8280daf.

docs/jwst/outlier_detection/outlier_detection.rst
stsci-hack: Revised under 8280daf.

docs/jwst/outlier_detection/outlier_detection.rst
stsci-hack: Revised under 8280daf.

CHANGES.rst
```diff
@@ -67,6 +67,9 @@ outlier_detection
   input ``ImageModels`` that are saved to disk instead of keeping them in memory.
   New parameters were aded to the step to control this functionality. [#6904]
 
+- Updated documentation of memory model and new parameters for memory use in
+  outlier_detection and resample steps. [#xxxx]
```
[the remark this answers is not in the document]
stsci-hack: Updated in 8280daf.

hbushouse:
I think there's only 1 place where actual code was changed in this PR, but just to be safe it would be good to run a full regression test against it. See the instructions at https://github.com/spacetelescope/jwst/wiki/Running-Regression-Tests-Against-PR-Branches for doing that.

PR #6989, opened 2022-08-16
JP-2803: Propagate asn keywords to resample output products for level 2 processing

tapastro:
Resolves [JP-2803](https://jira.stsci.edu/browse/JP-2803)

Closes #

This PR adds assignment of `model.meta.asn.pool_name` and `{...}.table_name` to level 2b i2d products - this required propagating these keywords through an intermediate ModelContainer in the resample step.

There was discussion in the JIRA ticket of propagating values from the first contained model if any ModelContainer was found have no assigned values, but I felt that this covers the issue more cleanly - if there is a need for other behavior, please let me know.

NB: If a user runs e.g. `strun calwebb_image2 some_rate.fits`, the i2d output will have values of 'singleton' for both of these keywords. If using a MAST-provided level2 asn, the values propagate as expected.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -78,7 +78,7 @@ resample
   pixels. [#6954]
 
 - Propagate ``asn.pool_name`` and ``asn.table_name`` through step ModelContainer
-  for level 2 processing of single input datamodels
+  for level 2 processing of single input datamodels [#6989]
 
 skymatch
 --------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -77,6 +77,9 @@ resample
   resampled error map contained an excessive number of zero-valued
   pixels. [#6954]
 
+- Propagate ``asn.pool_name`` and ``asn.table_name`` through step ModelContainer
+  for level 2 processing of single input datamodels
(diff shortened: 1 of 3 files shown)


hbushouse (approved):
Looks good to me.

Could resample_spec have the same issue for the s2d products created from a single input cal file?

hbushouse:
Don't forget to run regtests against the PR branch!

tapastro:
I have a Jenkins-RT run going, but I'll restart now that I've added a line to resample_spec - it looks like the table_name was being added in the style of taking the first contained model keyword values and assigning to the ModelContainer, so I've added a line for the pool_name as well.

stscieisenhamer (approved):

tapastro:
Regressions show 31 failures, all of which are keyword additions due to the new code.

hbushouse:
So this is ready to merge, but I'm going to wait a bit until a regtest run for another recent merge is finished and we have a chance to update truth files for that one (so that the differences don't get mingled together).

PR #6998, opened 2022-08-19
JP-2704 MIRI MRS Failures in Cube when there is no valid data on rate/cal image

jemorrison:
Resolves [JP-2704](https://jira.stsci.edu/browse/JP-2704)

Closes #2709

This PR addresses when no valid data is found on the detector because all the pixels are marked as DO_NOT_USE. 
An additional bug fix was made for correctly updating the naxis3 value of the final cube when wavelength planes are removed from the IFUCube because they contain no valid data. 

**Checklist for maintainers**
- [ ] added relevant label(s)


drlaw1558 (approved):
Looks ok to me; tested that it fixes the crash that we were seeing in https://jira.stsci.edu/browse/JP-2704

jwst/cube_build/ifu_cube.py
```diff
@@ -2117,20 +2143,19 @@ def setup_final_ifucube_model(self, model_ref):
...
 
                 if self.linear_wavelength:
                     self.crval3 = self.zcoord[remove_start]
                 else:
-                    self.wavelength_table = self.wavelength_table[remove_start:self.naxis3 - remove_final]
+                    self.wavelength_table = self.wavelength_table[remove_start: (self.naxis3 - remove_final)]
                     self.crval3 = self.wavelength_table[0]
 
+                # update WCS information if removing wavelengths from the IFU Cube
+                self.naxis3 = self.naxis3 - (remove_start + remove_final)
```
hbushouse, line 2158: Is this the same fix that was recently implemented in another PR? And if so, is the arithmetic correct here in terms of needing to add or subtract 1 to match up with the python slicing behavior?

hbushouse (commented):
So what happens when there's no good data and this function returns a list of `None` values? I assume the higher-level calling routines are already setup to handle that and skip the creation of an s3d file? And the pipeline flow also properly recognizes that no s3d was created and hence the extract_1d step also gets skipped?

hbushouse:
What's the reference to #2709 above in the description? That's a very old github issue that doesn't appear to be related to this PR at all.

jwst/cube_build/ifu_cube.py
jemorrison: The is correct. No need to add 1.  I was wrong before

jemorrison: So for all the data I have tested a cube is created - it is just really small. The extract1d step does run on the small cube. Let me force a situation where all the data is rejected and no cube is created and see what happens in extract1d

jemorrison:
I checked and if running in calspec2 and all the wavelength planes in a cube have zero data, then a log.error message is given. No IFU cube is created  and the extract1d step is skipped.

hbushouse (approved):
Looks good.

PR #6999, opened 2022-08-19
JP-2627: Fixes to pixel scale in resample

### LINKED ISSUE #5426: Propagate correct pixel scale for photometry keywords in resample
_Issue [JP-1757](https://jira.stsci.edu/browse/JP-1757) was created on JIRA by [James Davies](https://jira.stsci.edu/secure/ViewProfile.jspa?name=jdavies):_

They keywords PIXAR_SR and PIXAR_A2 are populated in the photometry step, and these represent a "nominal" pixel area in the unrectified data, often taken to be the area of the pixel at either aperture reference point or the distortion model reference point.

So depending on how resample then sets the output rectified pixel scale, these may not be exact matches anymore for the resampled image even when pixel_scale_ratio=1.0.  So there may be some small errors in the pixel scale currently in resampled data.

These two keyword values are supposed to be the area of the pixel that has a value of 1.0 in the pixel area reference file.  And depending on the distortion solution, that may not be the pixel at the center of the input image that is used to compute the output pixel scale in `resample`.  So this needs resolving.  It's possible `resample` will need to depend on the pixel area reference files.

None of this work should affect that actual surface brightness scaling done in resample for data that is in MJy/sr.  Only compute those scales correctly.

cshanahan1:
The parameter 'pix_scale_ratio' was not being passed to ResampleData, so the default of 1.0 was always being used - this PR fixes that. Additionally, if 'pixel_scale' is provided, it overrides 'pix_scale_ratio' -  in this case, the header value for `pixel_scale_ratio` was not being recalculated to reflect the absolute pixel scale passed in. The source_catalog step uses this value from the header, so it needed to be accurately calculated. 

Resolves [JP-2627](https://jira.stsci.edu/browse/JP-2627) and [JP-1757](https://jira.stsci.edu/browse/JP-1757)
Closes #5426

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,6 +47,8 @@ cube_build
 
 - Add a check in the process of building a cube to confirm that there is valid data on the detector. [#6998]
 
+- Fix a bug when user changes the spatial scale [#7002]
+
 datamodels
 ----------
 
--- a/jwst/cube_build/ifu_cube.py
+++ b/jwst/cube_build/ifu_cube.py
@@ -304,12 +304,15 @@ class IFUCubeData():
         # center of spaxels
         self.xcoord = np.zeros(self.naxis1)
         xstart = xi_min + self.cdelt1 / 2.0
-
         self.xcoord = np.arange(start=xstart, stop=xstart + self.naxis1 * self.cdelt1, step=self.cdelt1)
 
         self.ycoord = np.zeros(self.naxis2)
         ystart = eta_min + self.cdelt2 / 2.0
         self.ycoord = np.arange(start=ystart, stop=ystart + self.naxis2 * self.cdelt2, step=self.cdelt2)
+        # depending on the naxis and cdelt values the x,ycoord can have 1 more element than naxis.
+        # Clean up arrays dropping extra values at the end.
+        self.xcoord = self.xcoord[0:self.naxis1]
+        self.ycoord = self.ycoord[0:self.naxis2]
 
         xv,yv = np.meshgrid(self.xcoord, self.ycoord)
         self.xcenters = xv.flatten()
@@ -334,7 +337,7 @@ class IFUCubeData():
             self.crpix3 = 1.0
             zstart = self.lambda_min + self.cdelt3 / 2.0
             self.zcoord = np.arange(start=zstart, stop=self.lambda_max, step=self.cdelt3)
-
+            self.zcoord = self.zcoord[0:self.naxis3]
         else:
             self.naxis3 = len(self.wavelength_table)
             self.zcoord = np.asarray(self.wavelength_table)
--- a/jwst/cube_build/src/cube_match_sky_driz.c
+++ b/jwst/cube_build/src/cube_match_sky_driz.c
@@ -135,8 +135,8 @@ int match_driz(double *xc, double *yc, double *zc,
 
   double *fluxv, *weightv, *varv, *ifluxv;  // vector for spaxel
 
-  int k,j,ix1,ix2,iy1,iy2,iz1,iz2, iw1, iw2;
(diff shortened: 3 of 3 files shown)


jwst/resample/resample.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, input_models, output=None, single=False, blendheaders=True,
                 ``output_shape`` is in the ``x, y`` order.
         """
         self.input_models = input_models
-
+        print('in resample data - pscale_ratio, pscale =', pscale_ratio, pscale)
```
hbushouse, line 60: need to remove this

jwst/resample/resample_step.py
```diff
@@ -79,6 +79,7 @@ def process(self, input):
         for reftype in self.reference_file_types:
             ref_filename = self.get_reference_file(input_models[0], reftype)
 
+        ref_filename='N/A'
```
hbushouse, line 82: Might want to add a comment here that this line will need to get removed when good drizpars ref files get delivered to CRDS and we want to start using them.

jwst/resample/resample_step.py
```diff
@@ -114,7 +116,14 @@ def process(self, input):
             util.update_s_region_imaging(model)
             model.meta.asn.pool_name = input_models.asn_pool_name
             model.meta.asn.table_name = input_models.asn_table_name
-            model.meta.resample.pixel_scale_ratio = self.pixel_scale_ratio
+
+            # if pixel_scale exists, it will override pixel_scale_ratio.
+            # calculate the actual value of pixel_scale_ratio based on pixel_scale
+            # because source_catalog uses this value from the header.
+            if not self.pixel_scale:
```
hbushouse, line 123: The default value of `pixel_scale` set in the "spec" block is None. Does this evaluate properly, or should it be 
`if self.pixel_scale is None`?

jwst/resample/resample_step.py
```diff
@@ -114,7 +116,14 @@ def process(self, input):
...
+                model.meta.resample.pixel_scale_ratio = self.pixel_scale_ratio
+            else:
+                model.meta.resample.pixel_scale_ratio = (self.pixel_scale ** 2) / self.meta.photometry.pixelarea_arcsecsq
```
hbushouse, line 126: Resetting the value of `pixel_scale_ratio` will certainly fix the situation with updating the pixel area keywords later on, but I'm wondering what might happen in the drizzle routines themselves when both `pixel_scale` and `pixel_scale_ratio` are now set to non-default values. Does the `make_output_wcs` function, for example, know enough to only pay attention to one or the other (e.g. only use `pixel_scale` when it's set)?

hbushouse (commented):
Just a few questions.

hbushouse:
Needs a change log entry, flake8 updates, and it appears that several unit tests are failing in the CI tests.

jwst/resample/resample_step.py
cshanahan1: yes, the presence of pscale will override pixel_scale_ratio in make_output_wcs. i will double check that this doesn't mess things up anywhere else.

stsci-hack: It would be helpful to have a log message reporting on HOW the value of `pixel_scale_ratio` gets set; one version reporting it was set based on the user input value of the ratio, while the other version reports that the user provided value of `pixel_scale` was used instead.   This will make it unambiguous as to what was actually used, given that a choice was made between 2 possible user inputs.

mcara: I agree with @stsci-hack .

mcara (changes requested):
A few comments and also, please, add a changelog entry.

PR #7000, opened 2022-08-24
Add an ORCID to CITATION.cff

jdavies-st:
Yes, my name is very common, so this should help.  😅 

**Checklist for maintainers**
- [x] added relevant label(s)


hbushouse (approved):
Yeah, folks around here use your name a LOT.

PR #7005, opened 2022-08-26
JP-2819: Bugfix - ensure wavelength array is of equivalent shape to data array

tapastro:
Resolves [JP-2819](https://jira.stsci.edu/browse/JP-2819)

Closes #

This PR fixes a bug in the previous PR #6964, which truncated the wavelength array extent for nan regions, causing array shape mismatch issues between the wavelength and data. This PR uses the data array shape to create the wavelength array.

**Checklist for maintainers**
- [x] added relevant label(s)


tapastro:
Starting the Jenkins job now: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/400/

hbushouse:
Yes, that kind of change log entry is fine.

hbushouse:
In addition to some unrelated differences in NIRCam results (due to updated ref files), the regtest run shows the expected additional wavelength extension in some of the slitless products, but the calints also shows slight differences in the SCI and ERR arrays. Could this be due to slightly different wavelength values being computed than before, which is then leading to slightly different flux cal values during application of photom? If so, how do we know which wavelength values are correct?

hbushouse:
Tracing through photom to see how wavelengths were computed previously for LRS slitless, I'm assuming it ends up calling `create_2d_conversion`, which in turn calls the `lib.wcs_utils.get_wavelengths` function, which does its computations starting here https://github.com/spacetelescope/jwst/blob/master/jwst/lib/wcs_utils.py#L65

I'm wondering how that matches (or doesn't) what's being done in this PR (conceptually similar, but details differ). Also wondering if the `get_wavelengths` function could just be called from this location in assign_wcs, instead of using another set of similar code.

nden:
IIRC, originally the `wavelength` array was intended only for Nirspec data to allow to do the wavelength correction step, which prevented getting a complete WCS result by simply evaluating the WCS. The intention was to all other modes to simply evaluate the WCS. Computing `wavelength` arrays seems like a shift in policy. 
I wonder if it can be misleading to users, since it is strictly one use case - evaluating the WCS at the center of each pixel. If the WCS result at the edges is needed that would need evaluation on different inputs.

hbushouse:
For the case of LRS slitless I don't see a problem with populating the wavelength array. The wavelengths don't get corrected or modified by anything else in the pipeline and should give the same results as before when going through the extract_1d step. This is because the extract_1d step first checks to see if the wavelength array is populated and, if so, use it rather than computing the wavelengths on the fly. And the way it computed the wavelengths on the fly is identical to what's being done in this PR. It also seems foolish to have the wavelength array in the data products if it's not populated. Of course that could be solved by just making sure that the wavelength attribute in the datamodel is set to None, so that it doesn't get created in the LRS slitless products.

hbushouse:
@tapastro were you going to modify the PR code yet to just call the `get_wavelengths` function, rather than duplicating the code that's in it here?

tapastro:
Yes, sorry - I was running regtests against this commit to see if any functional difference existed. I'll clean it up and ping a review request.

tapastro:
Realized I hadn't run regtests against the code update - started one here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/408/

hbushouse:
The regtest results show the expected differences due to the presence of the wavelength extension in products that didn't have it before, but it's also showing 100% differences in the wavelength extension values for the calints product: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/408/testReport/junit/jwst.regtest/test_miri_lrs_slitless/_stable_deps__test_miri_lrs_slitless_tso_spec2_calints_/

Have you looked at the wavelength values in the new results to make sure they're OK?

tapastro:
> The regtest results show the expected differences due to the presence of the wavelength extension in products that didn't have it before, but it's also showing 100% differences in the wavelength extension values for the calints product: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/408/testReport/junit/jwst.regtest/test_miri_lrs_slitless/_stable_deps__test_miri_lrs_slitless_tso_spec2_calints_/
> 
> Have you looked at the wavelength values in the new results to make sure they're OK?

The new wavelengths look correct - they are nan where they aren't defined, and real-valued where they are. The current truth value is an array of zeroes.

hbushouse:
Interesting that the current calints truth file has all the wavelengths zero. They should've been calculated and populated during the photom step and hence included in the calints results. Weird.

tapastro:
Yeah, this was the main issue raised in the initial issue #6963 - it's quite possible I solved this in a more roundabout way than needed, as I didn't realize this info should be calculated in photom. I assumed such information would populated once available, in assign_wcs.

hbushouse (approved):

hbushouse:
You're right. In the existing flow the 2D array of wavelengths is *computed* during the photom step, but only used locally to then compute the flux conversion factors for each pixel (because they're wavelength-dependent). That wavelength array was never stored in the `model.wavelength` attribute (just garbage collected at the end of the step).

PR #7008, opened 2022-08-29
JP-2038: Enforcing no path data in association files and expanding documentation

### LINKED ISSUE #5950: Inconsistent assumed file locations in association files
_Issue [JP-2038](https://jira.stsci.edu/browse/JP-2038) was created on JIRA by [Bryan Hilbert](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hilbert):_

I'm developing a set of notebooks for the JWebbinars that show how to run the various stages of the pipeline when using imaging data. I've noticed what seems to be an inconsistency in the assumed file locations when association files list relative paths between level 2 and level 3. 

For calwebb_image2, it seems that relative paths within the association file are interpreted as being relative to the directory that the pipeline is running in, while for calwebb_image3 the relative paths in the association file are interpreted as being relative to where the association file is sitting.

Here are the details of my set up:

I'm working in a directory called 'imaging_mode'. In this directory are my notebooks for running the pipelines. In a subdirectory called 'Stage1' I have rate files and a level 2 association file. In a subdirectory called 'Stage2' I have _cal.fits files and a level 3 association file. 

When running my stage 2 notebook in the 'imaging_mode' directory, I call calwebb_image2 and provide it with the association file 'Stage1/level2_lw_asn.json'. Within that association file, I have the member files listed as (e.g.) 'Stage1/my_file_rate.fits'. In this configuration, calwebb_image2 runs successfully. If I change the members in the association file by removing the 'Stage1', then the pipeline fails because it doesn't find the rate files.

However, when running calwebb_image3 in a similar way, I encounter problems. Again, I run my notebook in the 'imaging_mode' directory. I call the pipeline and give it my association file, which is in the Stage2 subdirectory: 'Stage2/level3_lw_asn.json'. Within the association file, if I list the members as (e.g.) 'Stage2/my_file_cal.fits', (which is the case that worked for calwebb_image2), the pipeline fails because it says it is looking for 'imaging_mode/Stage2/Stage2/my_file_cal.fits'. If I change the association file to remove the subdirectory from each member, then calwebb_image3 looks in the correct directory and runs successfully.

kmacdonald-stsci:
Resolves [JP-2038](https://jira.stsci.edu/browse/JP-2038)

Closes #5950

This PR addresses inconsistencies found when using relative path data in association files for the `expname`.  All input files, including association files and the files listed in them, must be in the same directory, therefore path information is not needed and can cause ambiguities when used.  To eliminate users from entering path data manually, an exception is now raised when opening an association file with path data in `expname`.

**Checklist for maintainers**
- [x] added relevant label(s)


drlaw1558:
Per discussion on https://jira.stsci.edu/browse/JP-2038 I have serious concerns about merging this PR

kmacdonald-stsci:
> Per discussion on https://jira.stsci.edu/browse/JP-2038 I have serious concerns about merging this PR

The JWST pipeline was designed to use only file names for `expname` with no path data, expecting all input files to be in the same input directory as the association file, for reasons given by James Davis and Rosa Diaz.  This limitation was documented in many places of the documentation.  The inconsistency was found due to a user incorrectly using an association file in a way that it was not designed to be used.  The code has been modified to enforce the correct usage of association files and the documentation has been expanded to make even clearer that no path data can be in association files `expname`.

drlaw1558:
Nonetheless, there is a fundamental difference between the pipeline sometimes breaking when used in unintended ways, and deliberately disabling a currently-working approach used by a large segment of the community.  We should document that using them in this way can have problems, and potentially even print a warning message when paths are detected in asn files, but not deliberately put roadblocks in the way of folks using the pipeline to process their own data.

kmacdonald-stsci:
> Nonetheless, there is a fundamental difference between the pipeline sometimes breaking when used in unintended ways, and deliberately disabling a currently-working approach used by a large segment of the community. We should document that using them in this way can have problems, and potentially even print a warning message when paths are detected in asn files, but not deliberately put roadblocks in the way of folks using the pipeline to process their own data.

I do not think any effort should be made to fix any errors in the JWST code to fix any problems encountered when using path data in association files, since it was purposefully designed to not use paths.  That said, since many people do this anyway, raising an exception may be overkill.  A warning saying something to the effect that association files are not designed to incorporate path data may be a better approach.

docs/jwst/associations/level2_asn_technical.rst
```diff
@@ -133,7 +133,10 @@ members *required*
 association. Each member has the following keywords.
 
 expname *required*
-  The exposure file name
+  The exposure file name.  This must be a filename only, with no path.  This
+  file must be in the same directory as the association file, so path is not
+  needed.  If path data is included, an exception is raised when loading the
```
hbushouse, line 138: Is it still true that an exception will be raised if a path is included or has it been downgraded to a warning?

hbushouse (approved):
Looks OK. I'm not sure if the docs should say that an exception will be raised if the inputs contain a path, because we aren't actually raising an exception (just a warning). But maybe it's OK to word it that way just to scare people away from using a path.

PR #7010, opened 2022-08-31
JP-2861: Refactor Asn_Lv2WFSS to better descriminate what direct images should be used

stscieisenhamer:
Resolves [JP-2861](https://jira.stsci.edu/browse/JP-2861)

This PR addresses the issue of extremely long running times while handling NIRCam WFSS mode.

The issue is that all image exposures were considered possibilities of being the direct image for the spectral data. This resulted in exponentially growing re-processing lists and hence exponentially growing runtimes.

Solution was to refactor the Asn_Lv2WFSS rule to not consider images that do not share either channel (NIRCam) or pupil (NIRISS) in the constraints. This reduces the re-process list size and the list itself is cleared without having exposures constantly being added. This brings runtime to manageable levels.

**Checklist for maintainers**
- [x] added relevant label(s)


jwst/associations/lib/rules_level2_base.py
```diff
@@ -1027,3 +1030,147 @@ def _init_hook(self, item):
...
+        if not sciences:
+            raise AssociationNotValidError(
+                '{} has no required science exposure'.format(
+                    self.__class__.__name__
+                )
+            )
+        science = sciences[0]
+
+        # Get the exposure sequence for the science. Then, find
+        # the direct image greater than but closest to this value.
```
hbushouse, line 1058: What exactly does "direct image greater than" translate to? Does that refer to the chronological order in which the exposures were obtained and hence suggests that the chosen direct image must be taken *after* the grism/science exposure? I'm not sure there's any enforced ordering of the direct and grism exposures in a WFSS observation.

jwst/associations/lib/rules_level2_base.py
```diff
@@ -1027,3 +1030,147 @@ def _init_hook(self, item):
...
+        Notes
+        -----
+        This is an override for the method in `DMSBaseMixin`.
+        The optical element is retrieved from the chosen direct image
+        found in `self.direct_image`, determined in the `self.finalize`
+        method.
+        """
+        item = self.direct_image.item
+        opt_elems = []
+        for keys in [['filter', 'band'], ['pupil', 'grating']]:
```
hbushouse, line 1165: I see `grating` referenced here in association (so to speak) with `pupil`. Keep in mind that in the NIRISS instrument the grisms used for WFSS are in the filter wheel and the blocking filters that go with them are in the pupil wheel (the opposite of NIRCam). So I hope that's accounted for here.

jwst/associations/lib/rules_level2b.py
```diff
@@ -599,195 +600,42 @@ class Asn_Lv2WFSS(
...
                     only_on_match=True,
                 ),
+            ], reduce=Constraint.any),
+            DMSAttrConstraint(
+                name='instrument',
+                sources=['instrume'],
+            ),
+            DMSAttrConstraint(
+                name='mode',
+                sources=['channel', 'pupil'],
```
hbushouse, line 625: For NIRCam, the `channel` keyword designates the wavelength channel of the instrument (short vs. long). The grisms only exist in the long wavelength channel and hence I would assume that the direct images that go with them would also normally be from the long channel, but I'm not sure that's strictly enforced anywhere (maybe it is in APT?). But I assume the purpose of this constraint is to just remove NIRCam short-wave exposures from consideration when building the level-2b ASN, right?

hbushouse (commented):
Looks great coding wise. Just a few comments and questions.

jwst/associations/lib/rules_level2_base.py
stscieisenhamer: The "order" referred to here is the "EXPSPCIN" order. It is the order defined by APT. This is seen in the bottom of either WFSS template under "Exposure Sequence Display". Direct images, when requested, are always after the science that requested it, where the "Direct Image" checkbox is set in the above "Science (Grism) Exposures".

I do not know if this enforces sequencing on the observatory itself, but it does define the EXPSPCIN values. So, one wants to find the nearest direct image that has an EXPSPCIN value greater than the current science one. You actually fixed a bug in this last year; the logic was reversed?

jwst/associations/lib/rules_level2b.py
stscieisenhamer: Yes, and in fact, in the latest push, this is changed to "detector" to make it even more restrictive. Speeds up association generation by about 40%. As best as I can tell with NIRCam, one always gets both short and long with any exposures, so one will always have a long channel, and in fact, the same long detector, as the science.

This was the problem originally; such constraints were not being made in the Constraints part of the rules, but after the fact. As a result, all short channel exposures were being added to the mix. This just made an aweful lot of extra work, with a 5X factor occurring with each actual grism exposure. Hence the exponential of the runtime.

jwst/associations/lib/rules_level2_base.py
hbushouse: So how and where is this now fixed?  I still see `pupil` being loaded as a constraint for NIS (below). Is that being used for the filter value, in the same way `filter` is used for the filter value in NRC?  I'm confused.

stscieisenhamer: Yes. NIRISS essentially has two filter wheels, one of which has some pupil-like spots. Hence the one wheel is called "pupil". The "filter" wheel has the grisms in it. Basically all mixed up. [See figure 3 here](https://jwst-docs.stsci.edu/jwst-near-infrared-imager-and-slitless-spectrograph#JWSTNearInfraredImagerandSlitlessSpectrograph-Opticalelements).

NIRCam was much cleaner, putting the grisms in their "pupil" wheel and all filters are in the "filter" wheel.

😞

hbushouse (approved):
OK, I guess it looks good now.

hbushouse:
Needs a regression test run.

stscieisenhamer:
[Regression run](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/275/)

PR #7012, opened 2022-09-01
Update NIRISS WFSS regtest with in-flight data

hbushouse:
This PR updates the NIRISS WFSS regression test to use recent in-flight data and the latest "specwcs" calibrations as part of the WCS. All input and truth files have been uploaded to artifactory and runs successfully when tested standalone.

Also just restructured the way the test modules themselves are written (cuz I just didn't like the old style!)

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/regtest/test_niriss_wfss.py
+++ b/jwst/regtest/test_niriss_wfss.py
@@ -2,37 +2,46 @@
 import pytest
 from astropy.io.fits.diff import FITSDiff
 
-from jwst.regtest import regtestdata as rt
+from jwst.stpipe import Step
 
 
 @pytest.fixture(scope='module')
 def run_nis_wfss_spec2(jail, rtdata_module):
-    """Run the calwebb_spec2 pipeline"""
-    step_params = {
-        'input_path': 'niriss/wfss/nir_wfss_spec2_asn.json',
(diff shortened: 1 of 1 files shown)


hbushouse:
Regtest run looks good: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/411/
There were 10 failures, but they're all due to a new NIRCam gain ref file sneaking into CRDS since the last time truths were updated (totally unrelated).

nden (approved):

dmggh (commented):
For my own education, looking at the readthedocs for this calwebb_spec2 mode, it appears that the wfss contamination step should be included but was removed.

hbushouse:
> For my own education, looking at the readthedocs for this calwebb_spec2 mode, it appears that the wfss contamination step should be included but was removed.

Yes, the `wfss_contam` step is available in the pipeline, but right now it's always skipped by default, because it's still a prototype and is somewhat buggy. Even in the old version of the test module, when it was telling the pipeline to save the `wfss_contam` results, it wasn't producing anything, because the step is skipped.

dmggh:
Seems reasonable. Thanks.

PR #7013, opened 2022-09-01
JP-2899: Fix pedestal residual in MRS straylight

drlaw1558:
This PR fixes residual pedestal offsets observed in MIRI MRS rate images described by https://jira.stsci.edu/browse/JP-2899

**Checklist for maintainers**
- [x] added relevant label(s)


jwst/straylight/straylight.py
```diff
@@ -182,5 +182,17 @@ def correct_xartifact(input_model, modelpars):
     model[:, 0:4] = 0.0
     output.data = output.data - model
 
+    # Remove any remaining pedestal stray light values based on inter-channel region
+    # chosen based on analysis of multiple bands of data from various programs
+    if (channel == '12'):
+        pedestal = np.nanmedian(output.data[280:740, 503:516])
```
hbushouse, line 188: My typical reaction to hardcoded values like is not a good one, but if you're really sure that these will work universally, I'm OK with it.

hbushouse (approved):
Seems straightforward enough.

hbushouse:
I'll kick off a regression test run against the PR branch, just to make sure nothing breaks anywhere.

hbushouse:
Regression test results are here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/416/testReport/

There's one unrelated difference and 6 expected (I assume) differences for calwebb_spec2 downstream products for an MRS exposure.

drlaw1558:
Right, I'd expect some value changes to all of these products.  We'll see more such changes as a result of the new darks that we're working on delivering over the next 2 days as well, so what's the best way to update the truth values for these examples?  Wait until all code/ref file changes are in and then update the values?  Can you remind me which file is being used for these tests?

hbushouse:
> Right, I'd expect some value changes to all of these products. We'll see more such changes as a result of the new darks that we're working on delivering over the next 2 days as well, so what's the best way to update the truth values for these examples? Wait until all code/ref file changes are in and then update the values? Can you remind me which file is being used for these tests?

Now that I know which tests are going to fail just due to this PR alone, I'll kick off another full run against master branch right after merging this PR and will hopefully be able to easily figure which changes are which and will update the truth files. For the record, we're now back to running our regression tests against CRDS-OPS, because PUB can sometimes fall behind a bit, and because some recent updates that were needed by the latest Cal code were only delivered to Ops.

PR #7017, opened 2022-09-02
Set DQ of ref pixels to DO_NOT_USE after refpix correction

### LINKED ISSUE #7015: Reference pixels set to DO_NOT_USE in DQ is fragile
Currently, the way that reference pixels get set to DO_NOT_USE in `calwebb_detector1` is by relying on them being set to zero in the GAIN reference file.  If they are set to zero there (and there's not a strong reason they should be), then `ramp_fit` flags any gain values <0 or NaN as NO_GAIN_VALUE and DO_NOT_USE here

https://github.com/spacetelescope/stcal/blob/8e00460560eea757f6959ef17b5877583bc60de8/src/stcal/ramp_fitting/utils.py#L981-L1017

If they are not set to zero in the GAIN reference file, as was the case briefly with NIRCam gain reference files this week, then the reference pixels never get set to DO_NOT_USE, `ramp_fitting` fits them, and they end up in the output _i2d mosaic produced by `calwebb_image3`, producing unwanted artifacts.  See right side below.

Screen Shot 2022-09-01 at 12 00 15

Setting GAIN reffile ref pix to zero so that they get flagged as DO_NOT_USE in `ramp_fitting` is a fragile and roundabout way to get that job done.

@hbushouse is there a good reason to not set the reference pixels DQ values to DO_NOT_USE at the end of the `refpix` step?  They've been corrected and now can be ignored from here on out.

jdavies-st:
Closes #7015 

This PR addresses the current fragility of how reference pixels are set to DO_NOT_USE in `calwebb_detector1`.  Currently they rely on the GAIN reference file having zeros in the reference pixel locations, and then `ramp_fitting` sets those pixels to DO_NOT_USE in the science data DQ.

This PR sets the DQ of reference pixels to DO_NOT_USE at the end of the `refpix` step.  It does require that `DQInitStep` label these properly as REFERENCE_PIXELS, but this seems to be the case currently for the currently selected MASK reference files for all instruments.

Is this the best way to do this?  Thoughts?

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -577,6 +577,11 @@ ramp_fitting
   removing NaNs from the rateints product and setting appropriate DQ
   flags. [#6949]
 
+refpix
+------
+
+- Set DQ of ref pixels to DO_NOT_USE after `refpix` correction [#7017]
+
 resample
 --------
 
--- a/docs/jwst/refpix/description.rst
+++ b/docs/jwst/refpix/description.rst
@@ -77,6 +77,8 @@ NIR Detector Data
        subtracted from the full group on a row-by-row basis.  Note that the ``odd_even_rows``
        parameter is ignored for NIR data when the side reference pixels are processed.
     #. Transform the data back to the JWST focal plane, or DMS, frame.
+    #. Flag reference pixel locations in DQ array as DO_NOT_USE now that the correction is
+       complete, so that they are not used in ramp fitting or any other downstream steps.
 
 MIR Detector Data
 +++++++++++++++++
--- a/jwst/refpix/reference_pixels.py
+++ b/jwst/refpix/reference_pixels.py
@@ -47,10 +47,13 @@ import logging
 from ..datamodels import dqflags
 from ..lib import reffile_utils
 
+
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
 
-#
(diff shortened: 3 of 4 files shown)


CHANGES.rst
```diff
@@ -577,6 +577,11 @@ ramp_fitting
   removing NaNs from the rateints product and setting appropriate DQ
   flags. [#6949]
 
+refpix
+------
+
+- Set DQ of ref pixels to DO_NOT_USE after `refpix` correction [#7017]
+
```
zacharyburnett, line 584:
```suggestion
```

now that #8671 is merged (switching change log handling to `towncrier`) this change log entry should be a file in `changes/` instead:
```shell
echo "Set DQ of ref pixels to ``DO_NOT_USE`` after ``refpix`` correction" > changes/7017.refpix.rst
```

(new PRs will include instructions on how to do this)

zacharyburnett (changes requested):

CHANGES.rst
zacharyburnett: (ignore this if this PR is no longer active or needed)

melanieclarke:
I'm going to close this PR since it's now very out of sync with main.  

The issue is still tracked in [JP-3703](https://jira.stsci.edu/browse/JP-3703) and #7015.  If INS decides this change is desirable, we can make a new PR.

PR #7019, opened 2022-09-02
JP-2866: Fix photom bug for NIRSpec FS data

hbushouse:
Resolves [JP-2866](https://jira.stsci.edu/browse/JP-2866)

This PR fixes a bug in the photom step, in the part that handles NIRSpec FS exposures. Two unique computations of the wavelength-dependent flux conversion factors are performed, one for applying to point sources and the other to extended sources. Unfortunately the results of the first calculation were being fed in as an input to the second round of calculations. This update saves the initial scalar conversion factor to a new variable, which gets used as input to both calls, without it getting overwritten by the results of the first call. Have confirmed using a test photom ref file that this now works properly.

**Checklist for maintainers**
- [x] added relevant label(s)


tapastro (approved):
LGTM

hbushouse:
Regression test run shows expected differences in 3 NIRSpec FS results: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/412/

PR #6987, opened 2022-08-16
JP-2678: Expose additional parameters in TweakRegStep.

mairanteodoro:
Resolves [JP-2678](https://jira.stsci.edu/browse/JP-2678)

This PR exposes the following parameters that were previously hardcoded:

- `abs_minobj`;
- `abs_searchrad`;
- `abs_use2dhist`;
- `abs_separation`;
- `abs_tolerance`;
- `abs_fitgeometry`;
- `abs_nclip`;
- `abs_sigma`.

**Checklist for maintainers**
- [x] added entry in `CHANGES.rst` within the relevant release section
- [ ] updated or added relevant tests
- [x] updated relevant documentation
- [x] added relevant milestone
- [x] added relevant label(s)
- [x] ran regression tests, post a link to the Jenkins job below.
      [How to run regression tests on a PR](https://github.com/spacetelescope/jwst/wiki/Running-Regression-Tests-Against-PR-Branches)
- [ ] Make sure the JIRA ticket is [resolved properly](https://github.com/spacetelescope/jwst/wiki/How-to-resolve-JIRA-issues)

**Regression tests results**
Regression tests results can be found [here](https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/401/).

PR #7023, opened 2022-09-03
Remove min_gaia and rename gaia->abs parameters

mcara:
1. This PR fixes a bug from PR #6987 which introduced `abs_minobj` parameter which should have _replaced_ `min_gaia` but it inadvertently left it in the step `spec` and also in the code.2. 

2. This PR renames `gaia`-specific parameters to be in line with the new convention (`abs`) adopted in #6987. This is continuation of the work started in #6946. In particular, @hbushouse see https://github.com/spacetelescope/jwst/pull/6946#discussion_r944534937.

Specifically, this PR renames: ``gaia_catalog``->``abs_refcat`` and ``save_gaia_catalog``->``save_abs_catalog``.

``save_abs_catalog`` is used to save used GAIADR? catalogs to disk (but it will also re-save user-supplied reference catalogs if users choose to do so - hopefully they won't).

This PR also removes `align_to_gaia` boolean step parameter since it now can be
  indicated via `abs_refcat` parameter. When `abs_refcat` is None or an
  empty string, alignment to an absolute astrometric catalog will be turned
  OFF. When `abs_refcat` is a non-empty string, alignment to an absolute
  astrometric reference catalog will be turned ON.

It is not clear to me if there is any mechanism for deprecating step's `spec` attributes. So this might introduce some backward incompatibility... Hopefully a minimal one.

**Checklist for maintainers**
- [x] added relevant label(s)

Regression Tests: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/415/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,8 +18,8 @@ photom
 straylight
 ----------
 
-- Fix a bug in calibrated MRS slopes from small pedestal rate offsets between exposures by reintroducing zeropoint
-  subtraction using dark regions of MRS detectors. [#7013]
+- Fix a bug in calibrated MRS slopes from small pedestal rate offsets between
+  exposures by reintroducing zeropoint subtraction using dark regions of MRS detectors. [#7013]
 
 tweakreg
 --------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,22 @@ photom
   flux conversion factors get computed correctly for both point-source and
   extended-source data. [#7019]
 
+tweakreg
+--------
+
+- Renamed ``gaia`` suffix to ``abs`` for the absolute astrometry
+  parameters. Specifically, ``gaia_catalog``->``abs_refcat``,
+  ``save_gaia_catalog``->``save_abs_catalog``. [#7023]
+
+- Removed ``align_to_gaia`` boolean step parameter since it now can be
+  indicated via ``abs_refcat`` parameter. When ``abs_refcat`` is None or an
+  empty string, alignment to an absolute astrometric catalog will be turned
+  OFF. When ``abs_refcat`` is a non-empty string, alignment to an absolute
+  astrometric reference catalog will be turned ON.
+
+- Bug fix: removed ``min_gaia`` which should have been removed in the
+  PR 6987. [#2023]
+
 1.7.0 (2022-09-01)
 ==================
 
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -8,6 +8,22 @@ photom
...
+  ``save_gaia_catalog``->``save_abs_catalog``. [#7023]
+
+- Removed ``align_to_gaia`` boolean step parameter since it now can be
+  indicated via ``abs_refcat`` parameter. When ``abs_refcat`` is None or an
+  empty string, alignment to an absolute astrometric catalog will be turned
+  OFF. When ``abs_refcat`` is a non-empty string, alignment to an absolute
+  astrometric reference catalog will be turned ON.
+
+- Bug fix: removed ``min_gaia`` which should have been removed in the
+  PR 6987. [#2023]
```
hbushouse, line 25: typo here in the PR number: 2023 should be 7023

docs/jwst/tweakreg/README.rst
```diff
@@ -191,7 +191,19 @@ Parameters used for absolute astrometry using a reference catalog:
 * ``sigma``: A positive `float` indicating the clipping limit, in sigma units,
   used when performing fit. (Default=3.0)
 
-Parameters used for absolute astrometry using a reference catalog:
+**Absolute Astrometric fitting parameters:**
```
hbushouse, line 194: There are several other `abs_xxxx` params described above. Why aren't they in this section that's dedicated to absolute fitting?

docs/jwst/tweakreg/README.rst
```diff
@@ -191,7 +191,19 @@ Parameters used for absolute astrometry using a reference catalog:
...
+
+Parameters used for absolute astrometry to a reference catalog.
+
+* ``abs_refcat``: String indicating what astrometric catalog should be used.
+  Currently supported options: `None`, '', 'GAIADR2' and 'GAIADR1' or a path
```
hbushouse, line 199: Minor nit pick, but it seems to me that a more logical order for the options would be something like GAIADR1, GAIADR2, path to a user-supplied cat, None, or '' (i.e. leave the null options till the end).

CHANGES.rst
mcara: either a typo or astigmatism

hbushouse (approved):
LGTM

hbushouse:
Arrrrrgh!  The link to the regtest run was posted, but no mention was made of the fact that it showed 3 errors due to this change. I guess it's partly my own fault for not looking at the results myself before merging.

Anyway, the regtest module https://github.com/spacetelescope/jwst/blob/master/jwst/regtest/test_nircam_align_to_gaia.py needs to be updated to take into account the change in step parameter names. @mcara

PR #7027, opened 2022-09-07
Fix failing assign_wcs test due to gwcs changes

mcara:
@jdavies-st  reported that pytest jwst/assign_wcs/tests/test_wcs.py::test_sip_approx was failing and my investigation showed that it happened after recent changes in `gwcs` reporting SIP fit errors from arcseconds to pixels (since fit quality is specified in pixels so it was good to be consistent). The only parameter that is different is `sipmxerr` and the test is expected to fail when using master branch of `gwcs`. Once there is a release of `gwcs`, this test will need to be updated.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/assign_wcs/tests/test_wcs.py
+++ b/jwst/assign_wcs/tests/test_wcs.py
@@ -1,4 +1,6 @@
 import pytest
+from packaging.version import Version
+
 from numpy import zeros
 from numpy.testing import assert_allclose
 from astropy import units as u
@@ -6,6 +8,7 @@ from astropy import wcs
 from astropy.io import fits
 from astropy.modeling.models import RotationSequence3D
 
+import gwcs
 from gwcs.wcstools import grid_from_bounding_box
 from gwcs.geometry import SphericalToCartesian, CartesianToSpherical
 
@@ -189,6 +192,8 @@ def test_create_fitswcs(tmpdir, create_model_3d):
     assert_allclose((ra, dec), (gra, gdec))
 
 
+@pytest.mark.xfail(Version(gwcs.__version__) > Version('0.18.1'),
+                   reason="Changes in fit errors units in gwcs.to_sip_fits")
 def test_sip_approx(tmpdir):
     # some of the wcs info
     true_wcs = {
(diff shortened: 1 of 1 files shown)


jdavies-st (approved):
🚀

jwst/assign_wcs/tests/test_wcs.py
```diff
@@ -189,6 +192,10 @@ def test_create_fitswcs(tmpdir, create_model_3d):
     assert_allclose((ra, dec), (gra, gdec))
 
 
+@pytest.mark.xfail(
+    Version('0.18.1') < Version(gwcs.__version__) < Version('0.18.2'),
```
hbushouse, line 196: Maybe I don't understand the subtleties python arithmetic used in this case, but how can a value be both greater than 0.18.1 and less than 0.18.2 at the same time? Should one of those entries perhaps have an equals sign added to it, e.g. `0.18.1 <=`?

jdavies-st:
Btw, this was the failure on the cron job:

https://github.com/spacetelescope/jwst/runs/8182928669?check_suite_focus=true#step:6:145

hbushouse:
> Btw, this was the failure on the cron job:
> 
> https://github.com/spacetelescope/jwst/runs/8182928669?check_suite_focus=true#step:6:145

It's only a change of about 14-15 orders of magnitude. What's the big deal? This is astronomy, after all.  ;-)

jwst/assign_wcs/tests/test_wcs.py
jdavies-st: gwcs 0.18.2 is not released yet, and the current dev version is 0.18.2a1.dev18+ge638714.

```python
>>> from packaging.version import Version
>>> Version('0.18.1') < Version("0.18.2a1.dev18+ge638714") < Version('0.18.2')
True
```

Dev versions are in between patch versions and always start with the unreleased version.  So 0.18.2dev < 0.18.2.

hbushouse: Ah ha!  Learn something new every day. See, I told you it was probably because I didn't completely understand the math involved. So ... never mind!

hbushouse (approved):

PR #7033, opened 2022-09-08
JP-2627: Fixes to pixel scale in resample and photometry keywords

### LINKED ISSUE #5426: Propagate correct pixel scale for photometry keywords in resample

cshanahan1:
(This PR is replacing #6999 )

The parameter 'pix_scale_ratio' was not being passed to ResampleData, so the default of 1.0 was always being used - this PR fixes that. Additionally, if 'pixel_scale' is provided, it overrides 'pix_scale_ratio' -  in this case, the header value for `pixel_scale_ratio` was not being recalculated to reflect the absolute pixel scale passed in. The source_catalog step uses this value from the header, so it needed to be accurately calculated. 

Resolves [JP-2627](https://jira.stsci.edu/browse/JP-2627) and [JP-1757](https://jira.stsci.edu/browse/JP-1757)
Closes #5426

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -64,7 +64,6 @@ class ResampleData:
                 all products in memory.
         """
         self.input_models = input_models
-
         self.output_filename = output
         self.pscale_ratio = pscale_ratio
         self.single = single
@@ -88,7 +87,10 @@ class ResampleData:
         rotation = kwargs.get('rotation', None)
 
         if pscale is not None:
+            log.info(f'Output pixel scale: {pscale} arcsec.')
             pscale /= 3600.0
+        else:
+            log.info(f'Output pixel scale ratio: {pscale_ratio}')
 
         # Define output WCS based on all inputs, including a reference WCS
         self.output_wcs = resample_utils.make_output_wcs(
--- a/jwst/resample/resample_step.py
+++ b/jwst/resample/resample_step.py
@@ -76,18 +76,19 @@ class ResampleStep(Step):
             # resample can only handle 2D images, not 3D cubes, etc
             raise RuntimeError("Input {} is not a 2D image.".format(input_models[0]))
 
-        # Get drizzle parameters reference file
-        self.wht_type = self.weight_type
-        for reftype in self.reference_file_types:
-            ref_filename = self.get_reference_file(input_models[0], reftype)
-
-        if ref_filename != 'N/A':
-            self.log.info('Drizpars reference file: {}'.format(ref_filename))
+        #  Get drizzle parameters reference file, if there is one
+        if 'drizpars' in self.reference_file_types:
+            ref_filename = self.get_reference_file(input_models[0], 'drizpars')
+            self.log.info('Using drizpars reference file: {}'.format(ref_filename))
             kwargs = self.get_drizpars(ref_filename, input_models)
-        else:
-            # If there is no drizpars reffile
-            self.log.info("No DRIZPARS reffile")
+        else:  # no drizpars reference file found
+            ref_filename = 'N/A'
+
+        if ref_filename == 'N/A':
+            self.log.info('No drizpars reference file found.')
             kwargs = self._set_spec_defaults()
+
+        self.wht_type = self.weight_type
 
         kwargs['allowed_memory'] = self.allowed_memory
 
@@ -105,6 +106,7 @@ class ResampleStep(Step):
         kwargs['crval'] = _check_list_pars(self.crval, 'crval')
         kwargs['rotation'] = self.rotation
         kwargs['pscale'] = self.pixel_scale
+        kwargs['pscale_ratio'] = self.pixel_scale_ratio
         kwargs['in_memory'] = self.in_memory
 
         # Call the resampling routine
@@ -116,7 +118,14 @@ class ResampleStep(Step):
             util.update_s_region_imaging(model)
             model.meta.asn.pool_name = input_models.asn_pool_name
             model.meta.asn.table_name = input_models.asn_table_name
-            model.meta.resample.pixel_scale_ratio = self.pixel_scale_ratio
+
+            # if pixel_scale exists, it will override pixel_scale_ratio.
+            # calculate the actual value of pixel_scale_ratio based on pixel_scale
+            # because source_catalog uses this value from the header.
+            if not self.pixel_scale:
+                model.meta.resample.pixel_scale_ratio = self.pixel_scale_ratio
+            else:
+                model.meta.resample.pixel_scale_ratio = (self.pixel_scale ** 2) / model.meta.photometry.pixelarea_arcsecsq
             model.meta.resample.pixfrac = kwargs['pixfrac']
             self.update_phot_keywords(model)
 
@@ -129,9 +138,9 @@ class ResampleStep(Step):
     def update_phot_keywords(self, model):
         """Update pixel scale keywords"""
         if model.meta.photometry.pixelarea_steradians is not None:
-            model.meta.photometry.pixelarea_steradians *= self.pixel_scale_ratio**2
+            model.meta.photometry.pixelarea_steradians *= model.meta.resample.pixel_scale_ratio**2
         if model.meta.photometry.pixelarea_arcsecsq is not None:
-            model.meta.photometry.pixelarea_arcsecsq *= self.pixel_scale_ratio**2
+            model.meta.photometry.pixelarea_arcsecsq *= model.meta.resample.pixel_scale_ratio**2
 
     def get_drizpars(self, ref_filename, input_models):
         """
@@ -187,7 +196,7 @@ class ResampleStep(Step):
             pixfrac=self.pixfrac,
             kernel=self.kernel,
             fillval=self.fillval,
-            wht_type=self.wht_type
+            wht_type=self.weight_type
             # pscale_ratio=self.pixel_scale_ratio, # I think this can be removed JEM (??)
         )
 
--- a/jwst/resample/tests/test_resample_step.py
+++ b/jwst/resample/tests/test_resample_step.py
@@ -422,9 +422,9 @@ def test_resample_variance(nircam_rate, n_images):
     result = ResampleStep.call(c, blendheaders=False)
 
     # Verify that the combined uncertainty goes as 1 / sqrt(N)
-    assert_allclose(result.err[5:-5,5:-5].mean(), err / np.sqrt(n_images), atol=1e-5)
-    assert_allclose(result.var_rnoise[5:-5,5:-5].mean(), var_rnoise / n_images, atol=1e-7)
-    assert_allclose(result.var_poisson[5:-5,5:-5].mean(), var_poisson / n_images, atol=1e-7)
+    assert_allclose(result.err[5:-5, 5:-5].mean(), err / np.sqrt(n_images), atol=1e-5)
+    assert_allclose(result.var_rnoise[5:-5, 5:-5].mean(), var_rnoise / n_images, atol=1e-7)
+    assert_allclose(result.var_poisson[5:-5, 5:-5].mean(), var_poisson / n_images, atol=1e-7)
 
 
 @pytest.mark.parametrize("shape", [(0, ), (10, 1)])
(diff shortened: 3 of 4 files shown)


jwst/resample/resample.py
```diff
@@ -89,6 +88,9 @@ def __init__(self, input_models, output=None, single=False, blendheaders=True,
 
         if pscale is not None:
             pscale /= 3600.0
+            log.info(f'Output pixel scale: {pscale} deg.')
```
hbushouse, line 91: Because the user is supplying the parameter value in units of arcsecs, perhaps it would be better to report it in the log while it's still in those units, rather than after it gets converted to degrees (just so the user has direct confirmation that the value they provided is the one being used). That could be done by simply moving this up 1 line before `pscale` gets converted.

jwst/resample/resample.py
```diff
@@ -545,4 +547,4 @@ def update_slit_metadata(self, model):
                 pass
             else:
                 if val is not None:
-                    setattr(model, attr, val)
+                    setattr(model, attr, val)
```
hbushouse, line 550: Probably need to remove the blank line after this, in order to keep flake8 happy.

jwst/resample/resample_step.py
```diff
@@ -116,8 +116,16 @@ def process(self, input):
...
             model.meta.asn.table_name = input_models.asn_table_name
-            model.meta.resample.pixel_scale_ratio = self.pixel_scale_ratio
+
+            # if pixel_scale exists, it will override pixel_scale_ratio.
+            # calculate the actual value of pixel_scale_ratio based on pixel_scale
+            # because source_catalog uses this value from the header.
+            if not self.pixel_scale:
+                model.meta.resample.pixel_scale_ratio = self.pixel_scale_ratio
+            else:
+                model.meta.resample.pixel_scale_ratio = (self.pixel_scale ** 2) / model.meta.photometry.pixelarea_arcsecsq
```
hbushouse, line 126: So this is calculating the ratio of the *areas* of output vs. input. Is that the way `pixel_scale_ratio` is supposed to be used, as opposed to the linear ratio? It could be, I'm just not sure.

hbushouse: Oh, and the computation is done using `pixelarea_arcsecsq`; is `pixel_scale` in units of arcsec at this point in the code or is it in degrees? If the latter, the difference in units needs to be taken into account.

jwst/resample/resample_step.py
```diff
@@ -116,8 +116,16 @@ def process(self, input):
...
             model.meta.resample.pixfrac = kwargs['pixfrac']
+            print('before', self.pixel_scale_ratio)
```
hbushouse, line 128: need to remove this print statement

jwst/resample/resample_step.py
```diff
@@ -128,10 +136,11 @@ def process(self, input):
 
     def update_phot_keywords(self, model):
         """Update pixel scale keywords"""
+        print('inside', self.pixel_scale_ratio)
```
hbushouse, line 139: remove print

jwst/resample/resample_step.py
```diff
@@ -128,10 +136,11 @@ def process(self, input):
...
         if model.meta.photometry.pixelarea_steradians is not None:
-            model.meta.photometry.pixelarea_steradians *= self.pixel_scale_ratio**2
+            model.meta.photometry.pixelarea_steradians *= model.meta.resample.pixel_scale_ratio**2
```
hbushouse, line 141: The fact that you're taking the square of `pixel_scale_ratio` here suggests that it's in linear units, as opposed to the ratio of the pixel areas. But up above it looked like it was being handled as if it were already a ratio of areas. Check for consistency. Maybe I'm just losing track of the units during the step flow.

hbushouse (commented):
Needs a change log entry and see the various comments and questions.

jwst/resample/resample_step.py
```diff
@@ -76,19 +76,18 @@ def process(self, input):
...
-        else:
-            # If there is no drizpars reffile
-            self.log.info("No DRIZPARS reffile")
+        else:  # no drizpars reference file found
+            ref_filename = 'N/A'
+            self.log.info('No drizpars reference file found.')
             kwargs = self._set_spec_defaults()
 
+        self.wht_type = self.weight_type
+
```
mcara, line 90: Should we even test that 'drizpars' is in `self.reference_file_types`? We just added 'drizpars' to `self.reference_file_types` on line 52.

The bigger issue is that `'No drizpars reference file found.'` message is printed only when 'drizpars' IS NOT IN `self.reference_file_types` (which is practically never) but not when CRDS returns 'N/A' in this line:

```python
ref_filename = self.get_reference_file(input_models[0], 'drizpars')
```

So, I think a separate `if` is required outside of this `if`:

```python
if ref_filename == 'N/A':
    self.log.info('No drizpars reference file found.')
````

File: jwst/resample/tests/test_resample_step.py [hunk 8]
```diff
@@ -493,3 +493,39 @@ def test_custom_wcs_pscale_resample_imaging(nircam_rate, ratio):
...
+    res = ResampleStep.call(im, pixel_scale=0.04, pixel_scale_ratio=0.7)
+    assert np.allclose(res.meta.resample.pixel_scale_ratio, 0.04**2 / pixarea)
+
+    # just pixel_scale
+    res = ResampleStep.call(im, pixel_scale=0.04)
+    assert np.allclose(res.meta.resample.pixel_scale_ratio, 0.04**2 / pixarea)
+
+    # just pixel_scale_ratio
+    res = ResampleStep.call(im, pixel_scale_ratio=0.7)
+    assert res.meta.resample.pixel_scale_ratio == 0.7
```
mcara, line 514: I do not think that tests should be checking [only] that correct value is populated/recorded in the `meta`. I think the test should be checking that output image of this step actually was created taking into account these parameters.

mcara (changes requested):
I added a couple of my own comments and also all @hbushouse comments should be addressed.

hbushouse:
Given this statement in the docs for `pixel_scale_ratio`: "A value of 0.5 means the output image would have 4 pixels sampling each input pixel." it sounds to me like it's a linear ratio. Hence a value of 0.5 or 2.0 would mean a factor of 4 difference in pixel area.

mcara:
> Given this statement in the docs for `pixel_scale_ratio`: "A value of 0.5 means the output image would have 4 pixels sampling each input pixel." it sounds to me like it's a linear ratio. Hence a value of 0.5 or 2.0 would mean a factor of 4 difference in pixel area.

That is correct. Hence:
```python
model.meta.resample.pixel_scale_ratio = np.sqrt((self.pixel_scale ** 2) / model.meta.photometry.pixelarea_arcsecsq)
```

jwst/resample/resample.py
cshanahan1: done

jwst/resample/resample_step.py
cshanahan1: its in degrees at this point so ive fixed that. still thinking about how the ratio should be calculated

jwst/resample/resample_step.py
```diff
@@ -76,19 +76,18 @@ def process(self, input):
...
+        #  Get drizzle parameters reference file, if there is one
+        if 'drizpars' in self.reference_file_types:
+            ref_filename = self.get_reference_file(input_models[0], 'drizpars')
+            self.log.info('Using drizpars reference file: {}'.format(ref_filename))
             kwargs = self.get_drizpars(ref_filename, input_models)
-        else:
-            # If there is no drizpars reffile
-            self.log.info("No DRIZPARS reffile")
+        else:  # no drizpars reference file found
+            ref_filename = 'N/A'
```
jdavies-st, line 85:
```suggestion
            ref_filename = "NULL"
```

jwst/resample/resample_step.py
hbushouse: Based on more recent info it appears that `pixel_scale_ratio` is in fact in linear space, not in terms of area, so the only remaining question is whether this scaling is in the right sense. Given that the documentation says that a `pixel_scale_ratio` value of 0.5 results in 4 output pixels for every 1 input pixel, the output pixel area is in fact 4 times smaller and hence multiplying the original area by 0.5*0.5=0.25 appears to be the correct thing to do.

cshanahan1:
for reasons explained in https://jira.stsci.edu/browse/JP-2914, i am just going to leave the tests here as tests that the header keywords in meta.resample and meta.photometry are correct rather than the output data size. once that issue is resolved, i will extend the tests here.

cshanahan1:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/425/#showFailuresLink

18 tests are failing but many seem to be due to a new distortion file, or the error '"ValueError: Input file path does not have an extension: N/A" so I'm not sure how many are related to these changes. I would expect changes only if the pixel scale ratio differs from the default.

hbushouse:
> https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/425/#showFailuresLink
> 
> 18 tests are failing but many seem to be due to a new distortion file, or the error '"ValueError: Input file path does not have an extension: N/A" so I'm not sure how many are related to these changes. I would expect changes only if the pixel scale ratio differs from the default.

Yes, the 15 failures for test_niriss_wfss are due to CRDS updates overnight and are expected. The 3 test_fgs_image2 failures are the same tests that showed (expected) differences due to the merger of #7037 late yesterday, but I'm not sure why they're actually throwing errors in your run. Perhaps try rebasing this PR branch to pick up the updates from the other PR and try again?  I've also updated the truth files for the wfss tests, so those should no longer fail.

cshanahan1:
i re ran the regression tests and the failures all appear to be unrelated https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/426/testReport/

can this be merged?

jwst/resample/resample_step.py
hbushouse: I don't see how line 128 has been fixed to take into account the fact that `pixel_scale` is in units of degrees at this point in the code. It's still dividing `pixel_scale` by `pixelarea_arcsecsq`, the latter of which is in units of arcsec.

hbushouse: Oh wait, maybe I see it now. `pixel_scale` is the original value provided by the user, which is in units of arcsec, but `pscale`, which is used in the actual drizzling code, is the one that's in units of degrees. OK, never mind. And `pixel_scale_ratio` is supposed to be in the sense of output / input (e.g. a value of 0.5 mean the output pixel size will be one-half that of the input), so it appears that the order of the calculation is also correct (if the output `pixel_scale` is smaller than the original, then we'll get a number < 1.

hbushouse (approved):
OK, I think everything looks correct now and I'm fine with having the tests only verify the value(s) of the output keywords, rather than the actual pixel size/area in the resampled image. That can come later.

mcara:
> OK, I think everything looks correct now and I'm fine with having the tests only verify the value(s) of the output keywords, rather than the actual pixel size/area in the resampled image. That can come later.

Actually, you've merged incorrect code. The equation used in this PR is just plain wrong because it computes pixel ratio as ratio of output to input **PIXEL AREAS**, while the correct formula was discussed in https://github.com/spacetelescope/jwst/pull/7033#issuecomment-1240918136

cshanahan1:
@mcara oh shoot, i thought i had changed that. ill open a new pr to swap that out.

mcara:
Since this PR seems to be related to some photometry issue, shouldn't a unit test designed to test correcness of photometry catch this?

cshanahan1:
> Since this PR seems to be related to some photometry issue, shouldn't a unit test designed to test correcness of photometry catch this?

yes, but i was encountering some issues with this (see [JP-2914](https://jira.stsci.edu/browse/JP-2914),). I think my original assessment in that ticket was incorrect because i wasn't accounting for the fact that the data units are surface brightness and not flux, but even when i accounted for that i was getting a discrepancy in the input vs. output

PR #7039, opened 2022-09-11
JP-2645 - Snowball and Shower flagging in Jump

mwregan2:
Resolves [JP-2645](https://jira.stsci.edu/browse/JP-2645)

Add parameters to allow flagging of snowballs and showers. Corresponding PR in stcal is https://github.com/spacetelescope/stcal/pull/118

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,13 @@ datamodels
 
 - Added writing S_RESFRI for residual_fringe and R_FRIFRQ for FRINGEFREQ reference file. 
 
+jump
+----
+
+- First version of snowball/shower flagging for the jump step
+  JP-#2645. This code will not be actiavated without either a set of
+  parameter reference files or a command line override. [#7039]
+
 master_background
 -----------------
 
--- a/docs/jwst/jump/arguments.rst
+++ b/docs/jwst/jump/arguments.rst
@@ -49,3 +49,15 @@ The ``jump`` step has five optional arguments that can be set by the user:
 
 * ``--after_jump_flag_time1``: A floating point value in units of seconds that gives the
   time since the ramp jump to flag groups as ramp jumps for the 2nd after jump flagging.
+
+* ``--expand_large_events``:  A boolean parameter that controls whether the jump step will expand the number of pixels that are flagged around large cosmic ray events. These are know as "snowballs" in the near-infrared detectors and "showers" for the MIRI detectors. In general, this should be set to True.
+
+* ``--min_jump_area``: The minimum number of contiguous pixels needed to trigger the expanded flagging of large cosmic rays events.
+
+* ``-min_sat_area``:  The minimum number of saturated pixels required to meet "sat_required_snowball".
+
+* ``--expand_factor``: A multiplicative factor applied to the enclosing circle or ellipse. This larger area will have all pixels flagged as having a jump.
+
+* ``--use_ellipses``:  A boolean value that if True will cause minimum enclosing ellipses to be fit to the jump contour rather than the default circles. This should be set to True for MIRI observations.
+
+* ``--sat_required_snowball``: A boolean value that if True requires that there are saturated pixels within the enclosed jump circle.
--- a/docs/jwst/jump/description.rst
+++ b/docs/jwst/jump/description.rst
@@ -96,3 +96,17 @@ The use of the reference files is flexible. Full-frame reference
 files can be used for all science exposures, in which case subarrays will be
 extracted from the reference file data to match the science exposure, or
 subarray-specific reference files may be used.
+
+Large Events (Snowballs and Showers)
+====================================
+All the detectors on JWST are affected by large cosmic ray
+events. While these events, in general, affect a large number of
+pixels, the more distinguishing characteristic is that they are
+surrounded by a halo of pixels that have a low level of excess
+counts. These excess counts are, in general, below the detection
+threshold of normal cosmic rays.
+
+To constrain the effect of this halo the jump step will fit ellipses or circles that enclose the large events and expand the ellipses and circles by the input expansion_factor and mark them as jump.
+
+The two types of detectors respond differently. The large events in the near infrared detectors are almost always circles with a central region that is saturated. The saturated core allows the search for smaller events without false positives.
+The MIRI detectors do not, in general, have a saturated center and are only rarely circular. Thus, we fit the minimum enclosing ellipse and do not require that there are saturated pixels within the ellipse.
--- a/jwst/jump/jump_step.py
+++ b/jwst/jump/jump_step.py
@@ -26,6 +26,12 @@ class JumpStep(Step):
         after_jump_flag_time1 = float(default=0) # 1st flag groups after jump groups within specified time
         after_jump_flag_dn2 = float(default=0) # 2nd flag groups after jump above DN threshold
         after_jump_flag_time2 = float(default=0) # 2nd flag groups after jump groups within specified time
+        min_sat_area = float(default=1.0) # minimum required area for the central saturation of snowballs
+        min_jump_area = float(default=5.0) # minimum area to trigger large events processing
+        expand_factor = float(default=2.0) # The expansion factor for the enclosing circles or ellipses
+        use_ellipses = boolean(default=False) # Use an enclosing ellipse rather than a circle for MIRI showers
+        sat_required_snowball = boolean(default=True) # Require the center of snowballs to be saturated
+        expand_large_events = boolean(default=False) # must be True to trigger snowball and shower flagging
     """
 
     reference_file_types = ['gain', 'readnoise']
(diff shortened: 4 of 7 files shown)


mwregan2:
I can't figure out why the unit tests are failing. They work for me. The assert error is saying something about file (s) not being closed. It is referring to the gain and read noise files.

CHANGES.rst
```diff
@@ -222,6 +257,10 @@ tweakreg
 
 - Refactored code to work with changes in ``tweakwcs`` version 0.8.0. [#7006]
 
+source_catalog
```
hbushouse, line 260: Some of these change log entries were originally entered long enough ago that they show up under the wrong release headings, because they didn't actually get merged until after 1.7.2 was released and hence now belong up in the top section for 1.7.3 development. I think that's been fixed in the version that's on master now. We can always fix any conflicts separately from this PR.

docs/jwst/jump/arguments.rst
```diff
@@ -49,3 +49,15 @@ The ``jump`` step has five optional arguments that can be set by the user:
 
 * ``--after_jump_flag_time1``: A floating point value in units of seconds that gives the
   time since the ramp jump to flag groups as ramp jumps for the 2nd after jump flagging.
+
+* ``--expand_large_events``:  A boolean parameter that controls whether the jump step will expand the number of pixels that are flagged around large cosmic ray events. These are know as "snowballs" in the near-infrared detectors and "showers"for the MIRI detectors. In General, this should be set to True.
```
hbushouse, line 53: Need a whitespace in `"showers"for`

hbushouse: Oh, and don't need to capitalize "General".

docs/jwst/jump/description.rst
```diff
@@ -96,3 +96,17 @@ The use of the reference files is flexible. Full-frame reference
 files can be used for all science exposures, in which case subarrays will be
 extracted from the reference file data to match the science exposure, or
 subarray-specific reference files may be used.
+
+Large Events (Snowballs and Showers)
+====================================
+All the detectors on JWST are effected by large cosmic ray
```
hbushouse, line 102: effected should be affected

docs/jwst/jump/description.rst
```diff
@@ -96,3 +96,17 @@ The use of the reference files is flexible. Full-frame reference
...
+events. While these events, in general, effect a large number of
```
hbushouse, line 103: same here: should be "affected", which is the verb version; "effect" is the noun version  (e.g. the snowball effect)

docs/jwst/jump/description.rst
```diff
@@ -96,3 +96,17 @@ The use of the reference files is flexible. Full-frame reference
...
+pixels, the more distinguishing characteristic is that they are
+surrounded by a halo of pixels that have a low level of excess
+counts. These excess counts are, in general, below the detection
+threshold of normal cosmic rays.
+
+To constrain the effect of this halo the jump step will fit ellipses or circles that enclose the large events and expand the ellipses and circles by the input expansion_factor and mark them as jump.
+
+The two types of detectors respond differently. The large events in the near infrared detetcors are almost always circles with a central region that is saturated. The saturated core allows the search for smaller events without false positives.
```
hbushouse, line 111: typo in "detetcors"

jwst/jump/jump_step.py
```diff
@@ -26,6 +26,12 @@ class JumpStep(Step):
         after_jump_flag_time1 = float(default=0) # 1st flag groups after jump groups within specified time
         after_jump_flag_dn2 = float(default=0) # 2nd flag groups after jump above DN threshold
         after_jump_flag_time2 = float(default=0) # 2nd flag groups after jump groups within specified time
+        min_sat_area = float(default=1.0) # minimum required area for the central saturation of snowballs
+        min_jump_area = float(default=5.0) # minimum area to trigger large events processing
+        expand_factor = float(default=2.0) # The expansion factor for the enclosing circles or ellipses
+        use_ellipses = boolean(default=False) # Use an enclosing ellipse rather than a circle for MIRI showers
+        sat_required_snowball = boolean(default=True) # Require the center of snowballs to be saturated
+        expand_large_events = boolean(default=False) # must be True to trigger snowball and shower flagging
```
hbushouse, line 34: So you're leaving this turned off by default. Is that just to play it safe until there's been more experience with the processing or is it strictly intended to be used during off-line reprocessing only?

jwst/jump/tests/test_jump_step.py
```diff
@@ -27,6 +27,7 @@ def generate_miri_reffiles(tmpdir_factory):
     gain_model.meta.subarray.xsize = xsize
     gain_model.meta.subarray.ysize = ysize
     gain_model.save(gainfile)
+    gain_model.close()
```
hbushouse, line 30: Hadn't noticed until now that these tests are generating files saved to disk. Are those being deleted later or are they just left sitting around in the test directories?

hbushouse (commented):
Overall looks good. Just minor comments about wording and typos.

cshanahan1:
I see that you added some tests in stcal for some of the functions you added there, but I think it would be good to have a few tests in JWST as well for the full step using these new options

jwst/jump/tests/test_jump_step.py
cshanahan1: they are written out to one of the pytest directories. i think you can override a reference file with a datamodel instead of a file path right, so these wouldn't have to be written out? i can open another pr to change this if thats the case

cshanahan1:
@mwregan2 (summarizing some of our call earlier)

i added a test to demonstrate the use of the `sat_required_snowball` flag, and it doesn't seem to be working (maybe an error on my end?). what i did was add in circles of pixels in the 1st (not 0th) read of the input data. inside those circles, and in every read after, there is a smaller circle of saturation flags in the core in the dq array in the input model. from what i understand of the code, what should be happening here is that the clusters of jump flags should be flagged as snowballs only if they also have saturated pixels in a circular shape within the boundaries, which they do in this case, so the jump flags should be expanded. it doesn't detect the snowballs in that read and therefore does not expand the radius of the jump dq flags as expected. additionally, it does detect a 'snowball' in the last read (where there is none). while i don't totally understand what is going on, and again it might just be an error on my end, i think it might have something to do with the calculation of the `new_flagged_pixels` array on line 349 in stcal. 

for now i have commented this test out

setup.cfg
```diff
@@ -43,7 +43,7 @@ install_requires =
     scikit-image>=0.17.2
     scipy>=1.6.0
     spherical-geometry>=1.2.22
-    stcal>=1.1.0
+    stcal @ git+https://github.com/mwregan2/stcal.git@snowball_shower
```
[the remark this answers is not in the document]
cshanahan1: done

cshanahan1 (approved):

jwst/jump/jump_step.py
mwregan2: I'm playing it safe. Karl G. and I need to do some work to determine what the best starting set of parameters are for the instruments.

jwst/jump/tests/test_jump_step.py
mwregan2: Without the close all the tests were failing with an error about files not being closed. This must have been due to some other change since I haven't touched the JWST unit tests for Jump.

hbushouse (approved):

hbushouse:
I'm assuming that there's no real benefit in running a regtest against the PR branch, because the new stuff is turned off by default, hence we don't expect the new code to get exercised at all.  Given that, I'm thinking we're ready to merge. Any objections?

mwregan2:
Yes, this will be transparent until we add the parameter reference files.

hbushouse:
Full regression test run on the master branch after merging this PR was clean, as expected (no changes).

mwregan2:
The tests were failing because files were not being closed. I just added the close to get the existing tests to work. I have no idea what changed to require that files be closed.


From: Clare Shanahan ***@***.***>
Reply-To: spacetelescope/jwst ***@***.***>
Date: Wednesday, October 5, 2022 at 3:28 PM
To: spacetelescope/jwst ***@***.***>
Cc: Michael Regan ***@***.***>, Author ***@***.***>
Subject: Re: [spacetelescope/jwst] JP-2645 - Snowball and Shower flagging in Jump (PR #7039)


@cshanahan1 commented on this pull request.

In jwst/jump/tests/test_jump_step.py<https://urldefense.com/v3/__https:/github.com/spacetelescope/jwst/pull/7039*discussion_r988270460__;Iw!!CrWY41Z8OgsX0i-WU-0LuAcUu2o!21jzkLU4neZF2_rgqTv6-tJwerwwXgFElihdguVOHjE7KStWNiepU8Uu8DTxuxKvxBB3gAp7Nu1_FuWPvXnA698$>:

> @@ -27,6 +27,7 @@ def generate_miri_reffiles(tmpdir_factory):

     gain_model.meta.subarray.xsize = xsize

     gain_model.meta.subarray.ysize = ysize

     gain_model.save(gainfile)

+    gain_model.close()

they are written out to one of the pytest directories. i think you can override a reference file with a datamodel instead of a file path right, so these wouldn't have to be written out? i can open another pr to change this if thats the case

PR #7042, opened 2022-09-13
Update WCS (or slit metadata) after image (or spectrum) resampling.

mairanteodoro:
Resolves [JP-2648](https://jira.stsci.edu/browse/JP-2648)

This PR updates the resample and resample_spec steps so that they update the WCS information to match the resampled output products.

**Checklist for maintainers**
- [x] added relevant label(s)

The regression test results can be seen here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/432/


nden:
I don't think using `update_fits_wcsinfo` is appropriate here because it will try to compute a distortion model and express it as SIP. The output is a resampled image and should not have any distortion.

mcara:
> I don't think using `update_fits_wcsinfo` is appropriate here because it will try to compute a distortion model and express it as SIP. The output is a resampled image and should not have any distortion.

It should be fine but set `degree` to 1: https://github.com/spacetelescope/jwst/blob/ed9a26aabf6af645c06b78d8f995bcd24ed6c01b/jwst/assign_wcs/util.py#L1251. Also lower `npoints` to 8 - it should be enough and less prone to accuracy loss.

mairanteodoro:
How about using [`resample.update_fits_wcs()`](https://github.com/spacetelescope/jwst/blob/b48f408b998f24c4ff00d3109043b2ad1d9afde1/jwst/resample/resample.py#L493) instead of `util.update_fits_wscinfo()`? The former is already being used in [`resample.resample_many_to_one()`](https://github.com/spacetelescope/jwst/blob/b48f408b998f24c4ff00d3109043b2ad1d9afde1/jwst/resample/resample.py#L243).

As a matter of fact, there's this comment in `resample.resample_many_to_one()`:

```
# TODO: The following two methods and calls should be moved upstream to
# ResampleStep and ResampleSpecStep respectively
if isinstance(output_model, datamodels.ImageModel):
    self.update_fits_wcs(output_model)
if isinstance(output_model, datamodels.SlitModel):
    self.update_slit_metadata(output_model)
```

Note that the method that's being called to resample the images for which the misalignment issue was observed is `resample.resample_many_to_many()`, which **does not** call `self.update_fits_wcs()`.

jwst/resample/resample.py
```diff
@@ -237,10 +236,8 @@ def resample_many_to_one(self):
             )
         )
 
-        # TODO: The following two methods and calls should be moved upstream to
-        # ResampleStep and ResampleSpecStep respectively
-        if isinstance(output_model, datamodels.ImageModel):
-            self.update_fits_wcs(output_model)
+        # TODO: The following method and call should be moved upstream to
```
stsci-hack, line 239: Should the `self.update_slit_metadata` method be moved up into the ResampleSpecStep class, like you are did for `self.update_fits_wcs`?  Not doing so would seem to introduce an inconsistency between the imaging and slit resample operations.

mairanteodoro: @stsci-hack I was not sure whether I should do that for this ticket or open a new ticket for spectroscopy and do it then. Given your input, I'll do it right away.

mcara (commented):
LGTM

CHANGES.rst
```diff
@@ -1,6 +1,19 @@
 1.7.3 (unreleased)
 ==================
 
+resample_spec_step
```
hbushouse, line 4: This heading can be changed to just "resample_spec"

CHANGES.rst
```diff
@@ -1,6 +1,19 @@
...
+------------------
+
+- Move update_slit_metadata out of ResampleData and into ResampleSpecStep. [#7042]
+
+resample_step
```
hbushouse, line 9: Same here: can be just "resample"

hbushouse (approved):
Looks good to me. Just a minor nit about the change log entries.

hbushouse:
Do we know why we're seeing lots of little changes in the regtest results, like SLITID keywords disappearing from some headers and values of other keywords like SLITSTRT1, SLITSTRT2 changing?  Is it perhaps due to this PR branch being behind master, such that the truth files are more current?

hbushouse:
Updated the PR branch from master and then reran the regtests. New results are here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/440/testReport/
Still seeing lots of changes in header keywords for s2d (resampled) files and the x1d's that are extracted from the s2d's. Is something in the new meta data update resulting in some keywords being removed or given different values? This is only showing up for resampled spectra (not images), which only get a meta data update, no FITS WCS update.

CHANGES.rst
mairanteodoro: Done.

CHANGES.rst
mairanteodoro: Done.

mairanteodoro:
> 

Not sure what is causing those errors. I will take a look into it and if I can't find the cause, I'll roll back the spec resampling for now.

PR #7044, opened 2022-09-13
JP-2913: Fix the coron median replace algorithm

hbushouse:
Resolves [JP-2913](https://jira.stsci.edu/browse/JP-2913)

This PR updates the median image replacement routine used in the coronagraphic "align_refs" step so that it rejects and replaces pixels that are NaN in addition to pixels that are flagged as bad (some NaNs are not flagged, which is another issue).

The presence of NaN's causes the various FFT routines used for alignment and shifting to return updated images that are all NaN (which is not a good thing).

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 1.7.3 (unreleased)
 ==================
 
+align_refs
+----------
+
+- Upgrade the median image replacement routine to also replace NaN pixels,
+  in addition to pixels flagged as bad. [#7044]
+
+
 1.7.2 (2022-09-12)
 ==================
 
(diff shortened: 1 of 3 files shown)


tapastro (approved):
Changes look good, pending a regtest run.

hbushouse:
Reg test results are here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/424/  All green.

hbushouse:
Updated the existing regression test to use these in-flight data that caused the problem with NaN's.

tapastro (approved):
LGTM

PR #7046, opened 2022-09-14
JP-2917: Master_background reopens files from asn_table, potentially ignoring input datamodels

tapastro:
Resolves [JP-2917](https://jira.stsci.edu/browse/JP-2917)

Closes #

This PR addresses a Help Desk issue where moving target IFU data being fed into calwebb_spec3 was not getting the mtwcs fix propagated through to cube_build. I traced this to a method in master_background that needs to check the asn_table for 'science' or 'background' labels on each member to sort for master_background creation and subsequent subtraction. However, this asn_table may have expnames that no longer refer to the datamodels that were provided to master_background, as in the case of data that goes through assign_mtwcs first. In this case, the ModelContainer sent to master_background contains models with suffixes of 'assign_mtwcs.fits' rather than 'cal.fits' - the sorting was opening datamodels based on the asn expnames, which refer back to the 'cal.fits' files.

Not sure how best to fix this issue yet, but I created a kluge to start. This will break if the assign_mtwcs suffix is changed. There should be a method to match the datamodels in the input to ``split_container`` with the datamodels listed in the asn_table, but I don't know enough about filename structure to guarantee a clean fix. Calling for help - @stscieisenhamer ? 

**Checklist for maintainers**
- [x] added relevant label(s)


jwst/master_background/master_background_step.py
```diff
@@ -231,12 +231,19 @@ def split_container(container):
...
+    # Check to see if assign_mtwcs has been run - if so, need suffix replacement
+    assign_mtwcs = False
+    for f in container:
+        if f.meta.cal_step.assign_mtwcs == 'COMPLETE':
+            assign_mtwcs = True
     for product in asn['products']:
         for member in product['members']:
             if member['exptype'].lower() == 'science':
-                science.append(datamodels.open(member['expname']))
+                science.append(datamodels.open(member['expname'].replace('cal', 'assign_mtwcs')))
```
stscieisenhamer, line 246: Concerning this particular point: There is `jwst.lib.suffix.replace_suffix` that should be used for this operation.

stscieisenhamer:
Concerning the general question about matching models to table entries: If this is truly needed, that should be a feature added to `ModelContainer`, since that is where the table is read/opened. Operations on `ModelContainer` would then book-keep this linkage.

Question is why is this needed? If a `ModelContainer` is already in-hand, why is there need to reference back to the originating the association and match on names? What code is doing this? Seems like something in the not-too-distant past dealt with this also?

tapastro:
Filed a [JP ticket](https://jira.stsci.edu/browse/JP-2917) on the issue, since this was reported from the Help Desk.

There is one small lingering issue from a code style perspective: the container method ``container.ind_asn_type(asn_exptype)`` returns a list of indices to the container's model list that matches the provided association expytpe. However, the list of models, as currently written, is indicated to be an internal variable by way of the name ``container._models``. Is there any reason to suggest this? If there is, should the method also be indicated as such?

jwst/datamodels/container.py
```diff
@@ -477,15 +478,11 @@ def ind_asn_type(self, asn_exptype):
...
-        for m in self.meta.asn_table.products[0].members:
-            if m.exptype.lower() == asn_exptype:
-                file_path, file_name = op.split(m.expname)
-                names.append(file_name)
         ind = []
         for i, model in enumerate(self._models):
-            if model.meta.filename in names:
-                ind.append(i)
+            if model.meta.asn.exptype:
+                if model.meta.asn.exptype.lower() == asn_exptype:
```
tapastro, line 484: I feel like there must be a better way of doing this - checking to ensure the entry exists, then applying .lower() to the result if so. Any thoughts? I was envisioning something like a dict get() method that would provide an empty string if no value was present, but maybe I'm overthinking it.

stscieisenhamer: if there is a chance this does not exist, then the `get` is the way to go. As for the default, maybe the default should be `science`, since, at this point, if there is no knowledge, the presumption is that any data is science.

jwst/master_background/master_background_step.py
```diff
@@ -231,14 +231,21 @@ def split_container(container):
...
     background = datamodels.ModelContainer()
     science = datamodels.ModelContainer()
-    for product in asn['products']:
-        for member in product['members']:
-            if member['exptype'].lower() == 'science':
-                science.append(datamodels.open(member['expname']))
-            if member['exptype'].lower() == 'background':
-                background.append(datamodels.open(member['expname']))
+
+    ind_science = container.ind_asn_type('science')
```
stscieisenhamer, line 238: Why are these numpy arrays as opposed to just lists? This does not seem necessary? Is there some type of math happening on `DataModel`s?  This would allow a more concise code something like:
```
for ind_science in container.ind_asn_type('science'):
    science.append(container._models[ind_science])
```

stscieisenhamer (approved):
A couple of commentary, but otherwise LGTM.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -110,6 +110,11 @@ properties:
             type: string
             fits_keyword: ASNTABLE
             blend_table: True
+          exptype:
+            title: Exposure type provided for this exposure in the association
+            type: string
+            fits_keyword: ASNEXPTP
```
hbushouse, line 116: If this is really intended only for internal use within the code and only ever accessed via the datamodel attribute, then it doesn't need to have a corresponding FITS keyword defined for it. Just leave out this one line in the schema definition. The datamodel attribute will still be there and useable in the code, but it just won't get stored in a keyword when/if the datamodel is serialized to a FITS file, which in this case is fine (IMHO). Oh, and that means the following line (blend_table) can be removed too.

hbushouse:
In the description above you mention that the `assign_mtwcs` step is modifying the file names of each datamodel in the input `ModelContainer` (or at least the ones that are "science" members) such that they now contain suffixes of "assign_mtwcs" rather than their original "cal", but I don't see anything explicit in the `assign_mtwcs` step that's doing this to the data models. Or is it being done behind the scenes by stpipe due to the fact that the "suffix" argument for the Step class is set to that value?

Regardless of how or where that's happening, I like the idea of adding the ASN member type to the datamodels anyway. That's good info to have and to carry along for whenever it might be needed later on, rather than having to go back to the original ASN info itself to figure it out.

jwst/datamodels/container.py
tapastro: Sorry to have left this sitting, going through the Help Desk wilderness...

Where exactly can get be applied here? I tried something like model.get('meta.asn.exptype'), but get an AttributeError from the datamodel having no 'get' method. Setting the default to science does seem appropriate.

jwst/master_background/master_background_step.py
tapastro: They were coded as numpy arrays prior to my meddling, and I wasn't sure why - so in fear of breaking something, I left it. I'll just tear off the bandaid and see if any pain surfaces.

jwst/datamodels/schemas/core.schema.yaml
tapastro: Sounds good - wasn't sure if there was a requirement to transfer these datamodel attributes to the FITS or not, but agreed that this one can stay internal.

tapastro:
Regression tests here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/430/

tapastro:
I think this should be go to go - there are failures in the regression test runs, though they appear to be expected, in a way. Output products now inherit the pool and asn table names - I thought this functionality already existed, but perhaps the changes to master_background fixed an unknown bug here.

PR #7047, opened 2022-09-16
JP-2918: Fix bad resampled frames for NRS MOS

hbushouse:
Resolves [JP-2918](https://jira.stsci.edu/browse/JP-2918)

This PR fixes the calculation of resampled output frames for NIRSpec slit data (either fixed-slit or MOS) by making the algorithm less sensitive to the presence of negative spectral traces in the input images (which occurs when nodding is used to do background subtraction).

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -99,6 +99,9 @@ class ResampleSpecData(ResampleData):
         else:
             refmodel = self.input_models[0]
 
+        refmodel_data = refmodel.data.copy()
+        refmodel_data[refmodel_data < 0.] = np.NaN
+
         refwcs = refmodel.meta.wcs
 
         s2d = refwcs.get_transform('slit_frame', 'detector')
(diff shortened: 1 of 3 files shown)


hbushouse:
This initial commit is a crude first attempt that appears to do a decent job of fixing the problem. It simply ignores all pixels with values < 0 when computing the weighted means used to find the target location and hence makes it less susceptible to negative spectral traces and even just random large negative values.

mcara (commented):
Seems reasonable, except, I am not sure about the need of the quite non-informative logging messages introduced in this PR.

hbushouse:
> Seems reasonable, except, I am not sure about the need of the quite non-informative logging messages introduced in this PR.

Looks like our comments just crossed paths!  The extra logging messages were just for my debugging work and will be removed again once we're OK with the fix.

jwst/resample/resample_spec.py
```diff
@@ -99,6 +99,9 @@ def build_nirspec_output_wcs(self, refmodel=None):
         else:
             refmodel = self.input_models[0]
 
+        refmodel_data = refmodel.data.copy()
+        refmodel_data[refmodel_data < 0.] = np.NaN
```
mcara, line 103: One possible way to improve the logic would be to find the minimum value in data and subtract it from data:
```suggestion
        min_refmodel_data = np.nanmin(refmodel_data)
        refmodel_data -= min_refmodel_data
```

mcara: Another approach would be to add back subtracted sky if that value was recorded in the metadata.

mcara:
@hbushouse What actually prompted this PR? What was happening when data had negative values? `wmean_s` is essentially a "center of mass" estimate. IRAF's `imexamine` actually used to take data and subtract mean value from data before computing CofM, see, e.g., https://github.com/mcara/imexamine-centroid/blob/f696adae0aaf15eb4de4c3b6474903f018390b6f/centroid.py#L258-L282

jwst/resample/resample_spec.py
hbushouse: Unfortunately this is not a case of having a constant background subtracted. It's images with the spectra in different places that've been subtracted from one another, so you're left with both positive and negative traces of what is essentially the same spectrum.

hbushouse:
> @hbushouse What actually prompted this PR? What was happening when data had negative values?

See the last comment in JP-2918. Briefly, sometimes it was getting values of `wmean_s` that were like -3510. When that value was used in the wcs transform to go to the sky frame, it was estimating source RA/Dec values that were way off from the real values. For example, it would compute RA~105, Dec~13, when the real RA/Dec were +155.9 and -57.8, respectively. That in turn caused it to use really big offsets ("ts") that got added to the y slit min/max. So the whole output frame was totally wrong.

jwst/resample/resample_spec.py
hbushouse: I implemented your suggested change to renormalize to the min value. Seems to work well without explicitly excluding some of the data (like my initial approach did).

hbushouse:
CI errors are unrelated problems with CodeCov of some kind.

mcara (approved):
LGTM

PR #7048, opened 2022-09-16
fix definition of pixel ratio

cshanahan1:
Fixes an error introduced by #7033 where the pixel scale ratio was incorrectly defined - it is now correctly computed as a ratio of pixel areas. 

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,10 +20,8 @@ resample
 - Fix calculation of 'pixel_scale_ratio' when 'pixel_scale' parameter is
   supplied, as well as fix a bug where this value was not being properly passed
   to ResampleStep, and another where photometry keywords weren't being updated
-  correctly to reflect the correct pixel scale ratio. [#7033]
+  correctly to reflect the correct pixel scale ratio. [#7033, #7048]
 
-- Changed computation of meta.resample.pixel_scale_ratio to correctly be the
-  ratio of input / output areas [#7048]
 
 tweakreg
 --------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,9 @@ resample
   to ResampleStep, and another where photometry keywords weren't being updated
   correctly to reflect the correct pixel scale ratio. [#7033]
 
+- Changed computation of meta.resample.pixel_scale_ratio to correctly be the
+  ratio of input / output areas [#7048]
+
 tweakreg
 --------
 
--- a/jwst/resample/resample_step.py
+++ b/jwst/resample/resample_step.py
@@ -77,6 +77,7 @@ class ResampleStep(Step):
             raise RuntimeError("Input {} is not a 2D image.".format(input_models[0]))
 
         #  Get drizzle parameters reference file, if there is one
+        self.wht_type = self.weight_type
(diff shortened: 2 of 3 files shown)


mcara (approved):
LGTM.

CHANGES.rst
```diff
@@ -22,6 +22,9 @@ resample
   to ResampleStep, and another where photometry keywords weren't being updated
   correctly to reflect the correct pixel scale ratio. [#7033]
 
+- Changed computation of meta.resample.pixel_scale_ratio to correctly be the
+  ratio of input / output areas [#7048]
```
mcara, line 26: I think you could have simply appended 7048 to the line above: `... pixel scale ratio [#7033, #7048]`

mcara:
I do not believe failing tests are related to this PR.

hbushouse (changes requested):
Found another error in the original PR that was merged. The changes implemented in terms of handling the drizpars reference file are now producing an error when the "get_reference_file" query to CRDS comes back with "N/A" (i.e. there's no drizpars ref file in CRDS applicable to the particular instrument+mode). The check for `ref_filename == 'N/A'` in line 87 of resample_step.py needs to be executed *before* the call to `get_drizpars` is made in line 83. If `get_drizpars` is called with a ref file name of "N/A", it throws an error.

jwst/resample/resample_step.py
```diff
@@ -79,14 +79,15 @@ def process(self, input):
...
-            self.log.info('Using drizpars reference file: {}'.format(ref_filename))
-            kwargs = self.get_drizpars(ref_filename, input_models)
         else:  # no drizpars reference file found
             ref_filename = 'N/A'
 
         if ref_filename == 'N/A':
             self.log.info('No drizpars reference file found.')
             kwargs = self._set_spec_defaults()
+        else:
+        	self.log.info('Using drizpars reference file: {}'.format(ref_filename))
```
hbushouse, line 89: The indent error here (too much indent) is causing code compilation to fail.

hbushouse: Since everything else appears to be correct, I'm going to go ahead and fix the error myself right in the PR branch, so that tests can be run.

hbushouse:
> I do not believe failing tests are related to this PR.

You can believe anything you want, but this was in fact leading to 2 different kinds of errors. One has been fixed, now I'm in the process of fixing the other. Always, always, always run a full regression test. Never rely on the built-in CI tests only.

cshanahan1:
@hbushouse should i run the regression tests on this branch then before we merge?

hbushouse:
> @hbushouse should i run the regression tests on this branch then before we merge?

I did one run earlier this morning, which revealed the remaining error (having to do with initializing `self.wht_type`) and am about to kick off another run now that that problem has been fixed too in the code.

hbushouse:
Yay!  CI tests have passed and the full regtest run is clean: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/428/

I'm merging.

hbushouse (approved):
It's all good now.

PR #7049, opened 2022-09-20
JP-2679: Update bkg_subtract to handle 3D coronagraphic data

dmggh:
Resolves [JP-2679](https://jira.stsci.edu/browse/JP-2679)

This PR updates the background subtraction step to accept rateints (3D) input BKG exposures. There is no requirement that these multi-integration exposures have the same number of exposures. The average over all integrations and all files will be subtracted from each integration of the input SCI exposure. I also embellished existing epydocs.

**Checklist for maintainers**
- [X] added relevant label(s)


jwst/background/background_sub.py
```diff
@@ -26,85 +26,171 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
...
         background-subtracted target data model
 
     """
+    # Subtract the average background from the member
+    log.debug(' subtracting avg bkg from {}'.format(input_model.meta.filename))
 
     # Compute the average of the background images associated with
     # the target exposure
-    bkg_model = average_background(bkg_list, sigma, maxiters)
+    if input_model.meta.model_type == "ImageModel":
```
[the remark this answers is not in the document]
hbushouse: In fact, now that I think about it more, I'm not sure why `data_dim` is even needed, at least here. It's actually the dimensions (2D vs 3D) of the *background* images that we really care about, as opposed to the dimensions of the science data (input_model). Once the background data have all been averaged together properly, they'll always end up as a 2D image to be subtracted from the input. At that point is doesn't matter what the dimensions of the input are, because you can just rely on numpy array broadcasting to subtract the 2D bkg image from the input, regardless of whether the input is 2D or 3D. Just like we do in the flat_field step where a 2D flat is divided into input data that are either 2D or 3D.  Just use an expression like `input_model.data -= avg_bkg.data` and numpy array broadcasting will do the right thing regardless of whether `input_model.data` is 2D or 3D.

jwst/background/background_sub.py
```diff
@@ -26,85 +26,171 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
...
+    elif data_dim == 3:  # rateints
+        result = subtract_images.subtract_ints(input_model, bkg_model)
 
-    # We're done. Return the result.
+    # We're done. Return the background model and result.
     return bkg_model, result
 
 
-def average_background(bkg_list, sigma, maxiters):
+def average_background(bkg_list, sigma, maxiters, data_dim):
```
hbushouse, line 69: don't need `data_dim` to be passed here. All you need to care about in this function is the dimensions of the bkg images, once they get loaded.

jwst/background/background_sub.py
```diff
@@ -26,85 +26,171 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
...
+        raise ValueError("Expected either a CubeModel or an ImageModel.")
 
-    # Subtract the average background from the member
-    log.debug(' subtracting avg bkg from {}'.format(input_model.meta.filename))
-    result = subtract_images.subtract(input_model, bkg_model)
+    bkg_model = average_background(bkg_list, sigma, maxiters, data_dim)
+
+    if data_dim == 2:  # rate
+        result = subtract_images.subtract(input_model, bkg_model)
+    elif data_dim == 3:  # rateints
```
hbushouse, line 62: Don't need a separate (new) function to handle 3D data. Just call the original `subtract` function as written. The numpy array arithmetic at https://github.com/spacetelescope/jwst/blob/master/jwst/background/subtract_images.py#L32 will automatically subtract the 2D avg bkg from each integration of the input data if the input is 3D.

jwst/background/subtract_images.py
```diff
@@ -40,3 +41,44 @@ def subtract(model1, model2):
 
     # Return the subtracted model
     return output
+
+
+def subtract_ints(sci_mod, bkg_mod):
```
hbushouse, line 46: Don't need this new function, as described above.

jwst/background/background_sub.py
```diff
@@ -26,85 +26,171 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
...
-    merr = np.ma.masked_array(cerr, mask=mdata.mask)
+        # Compute the combined ERR as the uncertainty in the mean
+        avg_bkg.err = (np.sqrt(merr.sum(axis=0)) / (num_bkg - merr.mask.sum(axis=0))).data
 
-    # Compute the combined ERR as the uncertainty in the mean
-    avg_bkg.err = (np.sqrt(merr.sum(axis=0)) / (num_bkg - merr.mask.sum(axis=0))).data
+        return avg_bkg
 
-    return avg_bkg
+    else:  # data_dim = 3; background rateints input
```
hbushouse, line 148: I think this could be structured in a much more simple way. You don't really need to completely separate branches to do all the steps depending on whether the bkg models are 2D or 3D. You could keep the original loop over the individual bkg files and then simply insert a check here https://github.com/spacetelescope/jwst/blob/master/jwst/background/background_sub.py#L82 that checks the dimensions of `bkg_model.data` and if it's 3D, then insert the new code that first averages the integrations of that bkg file into a single 2D image and then store the result in the accumulators `cdata` and `cerr`, just as it's done if the bkg_model is 2D. And then do the remaining sigma-clipping and averaging on the stack of 2D bkg images as before (i.e. using the existing code starting at https://github.com/spacetelescope/jwst/blob/master/jwst/background/background_sub.py#L90

hbushouse (commented):
I think this should be reorganized to be much simpler, without all the new functions and separate branches to handle 2D vs 3D bkg data. See all the individual comments.

jwst/background/background_sub.py
```diff
@@ -26,85 +26,171 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
...
         list of background exposure file names
 
+    sigma: float, optional
+        Number of standard deviations to use for both the lower
+        and upper clipping limits.
+
+    maxiters: int or None, optional
+        Maximum number of sigma-clipping iterations to perform
+
+    bkg_model: JWST data model
```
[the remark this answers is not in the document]
dmggh: Removed it.

jwst/background/background_sub.py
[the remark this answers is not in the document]
dmggh: Resolved this.

jwst/background/background_sub.py
dmggh: Resolved this.

jwst/background/background_sub.py
dmggh: Resolved this.

jwst/background/subtract_images.py
dmggh: Agreed; resolved this.

jwst/background/background_sub.py
dmggh: Right. Code has been simplified.

dmggh:
I need to update the reg test truth file.

dmggh:
I just updated the reg test truth file.

jwst/background/background_sub.py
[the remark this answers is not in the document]
hbushouse: But I'm still seeing it here in the list of param info.

jwst/background/background_sub.py
```diff
@@ -39,53 +49,86 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
...
     Returns:
     --------
-
     avg_bkg: data model
         The averaged background exposure
-
     """
 
+    # Determine the type and dimensionality of the background files
+    try:
```
hbushouse, line 83: The input background exposures can be more than just `ImageModel` and `CubeModel`. IFU exposures, for example, will be `IFUImageModel`. So I think it's best to not try to get them to conform to any particular datamodel and just figure out the dimensions using
```
bkg_model = datamodels.open(bkg_list[0])
image_shape = bkg_model.data.shape
bkg_dim = len(image_shape)
```
bkg_dim will have values of either 2 or 3.

hbushouse: So this logic can be collapsed to just those 3 lines of code suggested above and `image_shape` is still available for use below, without the need for separate try/except blocks.

jwst/background/background_sub.py
```diff
@@ -39,53 +49,86 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
...
     # Loop over the images to be used as background
     for i, bkg_file in enumerate(bkg_list):
         log.info(f'Accumulate bkg from {bkg_file}')
-        bkg_model = datamodels.ImageModel(bkg_file)
 
-        # Initialize the avg_bkg model, if necessary
-        if avg_bkg is None:
-            avg_bkg = datamodels.ImageModel(bkg_model.shape)
+        if bkg_dim == 2:  # do what the old code does
+            bkg_model = datamodels.ImageModel(bkg_file)  # 2D
```
hbushouse, line 106: We probably need to change this to a generic `datamodels.open()` rather than forcing it into an `ImageModel`. I didn't realize till now that the old code was doing that. It shouldn't be. Clearly when the step was first written we only envisioned it being used for simple imaging observations. But now we're throwing a number of different model types at it.

hbushouse: It's OK to still create `avg_bkg` as an `ImageModel` (up above in line 96), because all we need for the average background data internally is a simple 2D image model.

jwst/background/background_sub.py
```diff
@@ -39,53 +49,86 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
...
+
+            # Accumulate the data from this background image
+            cdata[i] = bkg_model.data  # 2D slice
+            cerr[i] = bkg_model.err * bkg_model.err
+            avg_bkg.dq = np.bitwise_or(avg_bkg.dq, bkg_model.dq)
+
+            bkg_model.close()
+
+        if bkg_dim == 3:
+            bkg_model = datamodels.CubeModel(bkg_file)  # 3D
```
hbushouse, line 116: Here too we should just open using the generic `datamodels.open()` to allow for any kind of 3D input model.

hbushouse (approved):
Looks good.

hbushouse:
Regression test results are at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/433/

All but 1 of the 17 failures are irrelevant (due to the PR branch being out of date with respect to master). The results for the 1 relevant test are here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/433/testReport/junit/jwst.regtest/test_miri_coron_rateint/_stable_deps__test_miri_coron/  and show an error in the new code.

jwst/background/background_sub.py
```diff
@@ -90,6 +90,9 @@ def average_background(bkg_list, sigma, maxiters):
     cdata = np.zeros(((num_bkg,) + image_shape))
     cerr = cdata.copy()
 
+    if bkg_dim == 3:
+        accum_dq_arr = np.zeros((image_shape), dtype=np.int32)
```
hbushouse, line 94: Might want to use *u*int32 for the data type, for consistency with the DQ arrays in the models.

dmggh:
These 3 failing CI tests appear unrelated to my updates.

hbushouse:
Latest regtest run is at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/436/

All but 1 failure is unrelated (differences due to the PR branch being out of date with master) and the 1 relevant test of coron data shows just trivial differences in a few output pixels.  So overall it looks good.

dmggh:
So if I okify the test, should that allow me to merge the PR ?

hbushouse:
I'll go ahead and merge and will okify the new result after the next regular regtest run.

dmggh:
Thanks !

PR #7051, opened 2022-09-22
JP-2681 add residual_fringe to calwebb_spec2 pipeline 

jemorrison:
Resolves [JP-2681](https://jira.stsci.edu/browse/JP-2681)

This PR incorporates the MRS residual fringe correction into the calwebb_spec2 pipeline. 

**Checklist for maintainers**
- [x] added relevant label(s)
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/439/testReport/junit/jwst.regtest/test_miri_residual_fringe/_stable_deps__test_residual_fringe_cal/


drlaw1558:
Will try to take a look at this tomorrow; hopefully updating the calspec2 param ref file won't crash the old pipeline adding a step it doesn't expect.  Might be good to update the regression test data to flight at the same time- would suggest using some 2C data from program 1524.

jemorrison:
Ok thanks I will grab some of the new flight data and update the regression tests.

hbushouse:
> Will try to take a look at this tomorrow; hopefully updating the calspec2 param ref file won't crash the old pipeline adding a step it doesn't expect. Might be good to update the regression test data to flight at the same time- would suggest using some 2C data from program 1524.

Oh, that could be a problem (updating the pars ref file with a new step name). I just did a quick experiment of running "strun calwebb_spec2 jw_blee_rate.fits --steps.residual_fringe.skip=True" and it threw an error for "unrecognized arguments". 

You can still go ahead and deliver a new pars ref file, but we'll have to coordinate the activation of the CRDS context that it's in with the release of the pipeline version that contains the step. So the new pars ref files definitely can NOT be delivered to Ops any time soon and should also not even go to CRDS-PUB, because theoretically PUB should support only stuff that's actually publicly released. So this would have to go to CRDS-TEST and then deployed to PUB when we release Cal for B9.0 (I think?). Should check with ReDCaT folks about the proper logistics for this.

drlaw1558:
Yep, new param ref file crashes the current pipeline in my tests as well.  Rather than waiting until B9.0 and worrying about coordinating updates to avoid crashes it sounds much cleaner and easier to do as you suggest and hardwire skip=True, at least for the time being.  That way we can make the code available for interested users much sooner as well.  I can simply deliver a new param ref file after the next build comes out.

jemorrison:
in the spec of the step I set:
 skip = boolean(default = True)

But when I run it on a file the residual fringe step still runs. I there something else I need to do ?

hbushouse:
That should do it, just like we do for `wfss_contam` at https://github.com/spacetelescope/jwst/blob/master/jwst/wfss_contam/wfss_contam_step.py#L22

Is there anything else, like in a pars ref file or something, that might be overriding it?

drlaw1558:
One oddity that I've noticed on testing this branch; if residual_fringe.skip is True, there's no logging information written out about how the step was skipped, and no FITS header keywords saying that it was skipped.  This is different from the performance I'm used to from other steps, which usually start the step, prints a log message about skipping the step, and sets SKIPPED into the relevant S_ header keyword.

If residual_fringe.skip is False then the step runs as expected, prints logging information and populates header keywords.  There is, however, an unusual log message in a different font/color from all other log messages (white background in a Notebook run instead of the usual pink):
Type of calibrated <class 'jwst.datamodels.ifuimage.IFUImageModel'>
Why is this message done via 'print' instead of self.log ?

jemorrison:
David the strange print message was something I forgot to remove. It is not there any more. 
I am trying to figure out why when the step is skipped it does not write any log information or write the information to the header that the step is skipped.

jemorrison:
@hbushouse Take a look at how I set up the residual_fringe in the table for calwebb_spec2.rst. I wanted to make clear that is was skipped so I added a subscript 2 and then a note for this subscript that the step by default is skipped.

CHANGES.rst
```diff
@@ -28,6 +28,11 @@ master_background
   ``asn_table``; instead sort input datamodels by new
   ``model.meta.asn.exptype`` metadata. [#7046]
 
+pipeline
+--------
+
+- Added residual_fringe correction to the calspec2 pipeline
```
hbushouse, line 34: Need to reference the PR number at the end of the entry.

docs/jwst/pipeline/calwebb_spec2.rst
```diff
@@ -40,54 +40,58 @@ TSO exposures. The instrument mode abbreviations used in the table are as follow
...
+| :ref:`wavecorr <wavecorr_step>`                          | |c| | |c| |     |     |     |     |      |      |        |     |
++----------------------------------------------------------+-----+-----+-----+-----+-----+-----+------+------+--------+-----+
+| :ref:`flat_field <flatfield_step>`\ :sup:`1`             | |c| | |c| | |c| | |c| | |c| | |c| |  |c| | |c|  |  |c|   | |c| |
++----------------------------------------------------------+-----+-----+-----+-----+-----+-----+------+------+--------+-----+
+| :ref:`straylight <straylight_step>`                      |     |     |     |     |     | |c| |      |      |        |     |
++----------------------------------------------------------+-----+-----+-----+-----+-----+-----+------+------+--------+-----+
+| :ref:`fringe <fringe_step>`                              |     |     |     |     |     | |c| |      |      |        |     |
++----------------------------------------------------------+-----+-----+-----+-----+-----+-----+------+------+--------+-----+
+| :ref:`residual_fringe <residual_fringe_step>` \ :sup:`2` |     |     |     |     |     | |c| |      |      |        |     |
```
hbushouse, line 70: I thought I remembered hearing that `residual_fringe` had to be done after the `photom` step, because it was setup to work on flux-calibrated data?

docs/jwst/residual_fringe/main.rst
```diff
@@ -27,8 +26,6 @@ The ``residual_fringe`` step can accept several different forms of input data, i
 
   - an association table (in json format) containing a list of input files
```
hbushouse, line 27: Is it still true that this step can accept an ASN table with a list of input files?  Or just a single input file/datamodel?

jwst/pipeline/calwebb_spec2.py
```diff
@@ -276,6 +280,8 @@ def process_exposure_product(
             # Call the cube_build step for IFU data;
             # always create a single cube containing multiple
             # wavelength bands
+            # for MIR_MRS data call the residual fringe correction
```
hbushouse, line 283: Looks to me like the step is being called up in line 257, so why is the comment down here?

jwst/residual_fringe/residual_fringe_step.py
```diff
@@ -104,61 +96,42 @@ def process(self, input):
...
-                                                           **pars)
-            result = rfc.do_correction()
-            result.meta.cal_step.residual_fringe = 'COMPLETE'
-            self.output_models.append(result)
-
-        return self.output_models
+        # 2. correct the  model
+        # 3. return from step
+
+        # Open the residual fringe reference file
```
hbushouse, line 106: Minor nit, but technically neither of the reference files is being "opened" by these calls. All you're doing is finding the names of the right files to use. They get opened later on.

hbushouse (commented):
Just a few minor comments and questions.

docs/jwst/pipeline/calwebb_spec2.rst
jemorrison: you are right. I fixed the table. It is in the correct place when it is called in the pipeline

jwst/pipeline/calwebb_spec2.py
jemorrison: Forgot to move comment when I moved where the code was run

jemorrison:
@drlaw1558 Try it now. You should now have the S_RESFRI=SKIPPED by default and you should also have logging information printed to the screen saying it is skipping the step.

drlaw1558:
I can see the expected logging information on the step now, both when run and skipped.  Still seems to be missed information about the reference file used though; I get the expected CRDS prefetch information but the FITS header doesn't seem to set the R_RESFRI or R_FRIFRQ (whatever the keyword is meant to be) giving the name of the FRINGEFREQ reference file used in the run.

Likewise, the logging information should probably say 'Using FRINGEFREQ reference file:' instead of 'Using Residual FRINGE reference file:'

jwst/pipeline/calwebb_spec2.py
```diff
@@ -252,6 +254,9 @@ def process_exposure_product(
         else:
             calibrated = self._process_common(calibrated)
 
+        # for MIR_MRS data call the residual fringe correction
```
hbushouse, line 257: I wonder if it would be a little cleaner (in terms of code complexity) to move this call into the `_process_common` function, along with the `fringe` step. The setting of the `skip` param will still control whether or not it gets applied to data for different modes.

jemorrison: David - currently there is no keyword for the FRINGEFREQ in the core schema. I need to make up one. Do you have a preference R_FRIFRQ ?

drlaw1558: R_FRIFRQ makes sense to me, but I defer to whatever you and Howard think most consistent with the other keywords.

jemorrison: @drlaw1558 have you used the parameters 'ignore_region_min' or 'ignore_region_max' to not used wavelengths. I can not get these parameters to work anymore

drlaw1558: Can't say as I've ever tried using those before

drlaw1558 (approved):
Looks good to me!  Seeing the expected performance and logging information.

jemorrison:
one of the unit tests fails now because there is a error in stpipe how it works with lists as input parameters.https://github.com/spacetelescope/stpipe/issues/71

Hopefully we can still merge with this error - or I could comment out the test until the fix is made.

jemorrison:
@hbushouse 
I put the link to the regression test. The difference in the residual_fringe tests is now S_RESFRI and R_FRIFRQ. Is written. I don't think the other differences matter.  Should I update the truth file ? I am not sure how I get access to the one created with this regression test. Or should we just wait and Okify them ?
I think all the other CI failures are related to the error about not being about the parse the list parameter - which Mihai already filed a ticket for.

hbushouse:
Once the latest run of the CI tests has finished (hoping that a few more of them might succeed this time around) I think it'll be ready to merge. I agree that the regtest results are OK. We'll okify all the truth files after we've merged and done another regular full regtest run.

hbushouse:
At least one of the CI tests is showing an error in jwst/jwst/residual_fringe/tests/test_configuration.py

jemorrison:
Yes this test is where I discovered that there is now a bug in parsing a parameter with a list. I can comment out that test until that gets fixed if you want

hbushouse:
> Yes this test is where I discovered that there is now a bug in parsing a parameter with a list. I can comment out that test until that gets fixed if you want

Just use the pytest.mark.xfail command to mark it as failing for now.

jemorrison:
Added pytest.mark.xfail to test. This PR should be ready to go

PR #7057, opened 2022-10-06
Change to new asdf resource entry point

WilliamJamieson:
This PR switches the datamodels to using the new `asdf.resources` entry-point for the datamodels schemas.  

**Checklist for maintainers**
- [x] added relevant label(s)


WilliamJamieson:
I'm not sure why there are test failures. I cannot reproduce these locally.

jwst/datamodels/integration.py
```diff
@@ -0,0 +1,27 @@
+import sys
```
stscieisenhamer, line 1: docstrings for module and function should be added.

WilliamJamieson: See 69ea324 for documentation.

stscieisenhamer: LGTM

stscieisenhamer (approved):

hbushouse:
Is it important to try to include this PR in our next B9.0 release candidate, which is expected to be formed either today or next Monday?  What kind of risks does it entail, if we try to include it at the last minute for the B9.0 delivery?

WilliamJamieson:
> Is it important to try to include this PR in our next B9.0 release candidate, which is expected to be formed either today or next Monday? What kind of risks does it entail, if we try to include it at the last minute for the B9.0 delivery?

I don't think this is critical. Though it would be nice, as it might help solve some of the weird install/upgrade issues that users are having (schemas going missing when they upgrade in their environment). It totally eliminates the old entry point for the datamodels so if jwst can still write files which use datamodels it should be fully functional.

hbushouse:
Full regression test run is at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/443/

PR #7069, opened 2022-10-14
Revert "JP-2812 : Allow user-provided custom image catalogs for tweakreg"

zacharyburnett:
Reverts spacetelescope/jwst#7022


hbushouse:
@mcara  Do we still need the revert or will there be a PR with a fix for master?

mcara:
No need. The bug was fixed 🤞 (pending user reports, of course) in #7071

PR #7295, opened 2022-10-16
JP-2971 Make the GWCSDrizzle.outcon attribute a property with setter

stscieisenhamer:
This forces the context array to be in sync with the products `con` attribute.

Resolves [JP-2971](https://jira.stsci.edu/browse/JP-2971)

This PR addresses an issue found in by a user where the output of `calwebb_image3` had a context array that made no sense. The issue was that the context array was being calculated but not attached to the output model.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,11 @@ datamodels
 
 - Fix handling of ASN directory path by the ``ModelContainer``. [#7071]
 
+resample
+--------
+
+- Make the GWCSDrizzle.outcon attribute a property with setter [#7295]
+
 set_telescope_pointing
 ----------------------
 
--- a/docs/jwst/data_products/science_products.rst
+++ b/docs/jwst/data_products/science_products.rst
@@ -373,33 +373,33 @@ different set of data arrays than other science products. Resampled 2-D images a
 ``i2d`` products and resampled 2-D spectra are stored in ``s2d`` products.
 The FITS file structure for ``i2d`` and ``s2d`` products is as follows:
 
-+-----+-------------+----------+-----------+---------------+
-| HDU | EXTNAME     | HDU Type | Data Type | Dimensions    |
-+=====+=============+==========+===========+===============+
-|  0  | N/A         | primary  | N/A       | N/A           |
-+-----+-------------+----------+-----------+---------------+
-|  1  | SCI         | IMAGE    | float32   | ncols x nrows |
-+-----+-------------+----------+-----------+---------------+
-|  2  | ERR         | IMAGE    | float32   | ncols x nrows |
-+-----+-------------+----------+-----------+---------------+
-|  3  | CON         | IMAGE    | int32     | ncols x nrows |
-+-----+-------------+----------+-----------+---------------+
-|  4  | WHT         | IMAGE    | float32   | ncols x nrows |
-+-----+-------------+----------+-----------+---------------+
-|  5  | VAR_POISSON | IMAGE    | float32   | ncols x nrows |
-+-----+-------------+----------+-----------+---------------+
-|  6  | VAR_RNOISE  | IMAGE    | float32   | ncols x nrows |
-+-----+-------------+----------+-----------+---------------+
-|  7  | VAR_FLAT    | IMAGE    | float32   | ncols x nrows |
-+-----+-------------+----------+-----------+---------------+
-|     | HDRTAB*     | BINTABLE | N/A       | variable      |
-+-----+-------------+----------+-----------+---------------+
-|     | ASDF        | BINTABLE | N/A       | variable      |
-+-----+-------------+----------+-----------+---------------+
++-----+-------------+----------+-----------+-------------------------+
+| HDU | EXTNAME     | HDU Type | Data Type | Dimensions              |
++=====+=============+==========+===========+=========================+
+|  0  | N/A         | primary  | N/A       | N/A                     |
++-----+-------------+----------+-----------+-------------------------+
+|  1  | SCI         | IMAGE    | float32   | ncols x nrows           |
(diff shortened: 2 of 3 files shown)


stscieisenhamer:
[Regression test](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/284/)

stscieisenhamer:
Don't know who else in gwcs/drizzle/resample world should review.

stscieisenhamer:
CI tests have the usual CRDS issues, but otherwise are clean.

hbushouse (approved):

hbushouse:
> [Regression test](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/284/)
> 
> Results summary: All differences are the expected change in dimensionality of the CON array from 2D to 3D.

Why are the CON arrays now 3D?  The resampled images are 2D.

mcara:
> > [Regression test](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/284/)
> > Results summary: All differences are the expected change in dimensionality of the CON array from 2D to 3D.
> 
> Why are the CON arrays now 3D? The resampled images are 2D.

I do not know exactly the issue this PR is trying to address but 2D or 3D is due to the number of input images: each bit in an `int` shows which input image contributed to a given resampled pixel. Since `int`has 32 bits (in the FITS world and hence probably it is the limit supported by `drizzle` code), there is need for multiple layers on ints (hence 3rd dimension), to represent contributions from images number 0-31, 32-63, 64-95, 96-127, ...

mcara (changes requested):
I do not think this is correct. I am still processing this and so I may change my mind, but I think this is wrong

mcara:
My reply to JP-2971:

> I do not understand at all the description of the problem. Line 93 simply insures that CON array is a 3D array making the input 2D CON array the first plane in a 3D CON array. Since I did not participate in the development of this step I lack the CONtext as to why this is done like this but I hope there is a good reason for doing it like that.

More over,

1. [Line 62](https://github.com/spacetelescope/jwst/blob/c3a4a0591074fa632e159266b24b51da69018ba6/jwst/resample/gwcs_drizzle.py#L62) is unnecessary and should be removed.
2. The entire code in [`increment_id`](https://github.com/spacetelescope/jwst/blob/c3a4a0591074fa632e159266b24b51da69018ba6/jwst/resample/gwcs_drizzle.py#L206-L209) relies on 3D dimensionality of CON array. What if `product.con` is a 2D array?

stscieisenhamer:
CON arrays are 3D. See the [__init__](https://github.com/spacetelescope/jwst/blob/c3a4a0591074fa632e159266b24b51da69018ba6/jwst/resample/gwcs_drizzle.py#L92), the [increment_id method](https://github.com/spacetelescope/jwst/blob/c3a4a0591074fa632e159266b24b51da69018ba6/jwst/resample/gwcs_drizzle.py#L189), and [dodrizzle](https://github.com/spacetelescope/jwst/blob/c3a4a0591074fa632e159266b24b51da69018ba6/jwst/resample/gwcs_drizzle.py#L348) itself.

The reason is the that CON array it a bit indicator of which input images contribute to a pixel. If there are >32 images, the bit marking overflows into the next dimension.

If the number of input images is <= 32, it could remain 2D. However, I would vote for consistency, and always leave it 3D.

Also note that since it is a bitflag, the type should be uint32; testing that now.

mcara:
> If the number of input images is <= 32, it could remain 2D. However, I would vote for consistency, and always leave it 3D.

You kind of _have to_ vote for consistency since the rest of the code seems to me to depend on 3D.

> Also note that since it is a bitflag, the type should be uint32; testing that now.

I agree it _preferably_ should be `unsigned integer` but it does not _have to_. But if this is possible to fix, and not cause too much trouble (i.e., backward compatibility), I am for it.

mcara:
Oy... Why is [`product`](https://github.com/spacetelescope/jwst/blob/c3a4a0591074fa632e159266b24b51da69018ba6/jwst/resample/gwcs_drizzle.py#L25) described as "optional" in the docs?

stscieisenhamer:
And, just to clarify, and what may have already been mentioned: The dimensionality of the array is not changing. It has always been 3D. The problem is as follows: The input `product` is also the output of the process, and the GWCSDrizzle object depends on using the product's arrays. However, with the CON array and the GWCSDrizzle.outcon variables, since the array is both reshaped and concatenated to, the equivalency between `product.con` and `GWCSDrizzle.outcon` is broken. As a result, the reshaping and growing of the context array never ends up in the final product.

And just seeing Mihai's about the docs: Agreed, `product` is absolutely necessary. And agree that those initialization lines, all of them, are unnecessary.

mcara (approved):
Upon further digging through the code, I think this could be a good fix. Going to give it emergency approval.

mcara:
I also recommend to modify an existing regression test to check that CON is populated correctly

stscieisenhamer:
Changing the schema did add some extra work, so not bothering. Updating some docs but otherwise will finish up.

I'll okify the regressions once merged.

mcara:
With regard to `uint` - I do not see how changing from `int` to `uint` would cause any backward compatibility issues. I think you should make CON to be `uint` as this is what people _expect_.

stscieisenhamer:
Hmmm, issue with changing to uint32 appears to be in `cdrizzle.tdriz`. It errors with "Invalid context array". So, going to leave the schema as-is, and can open new issue to consider what that means later on.

mcara:
@stscieisenhamer Merge at will

mcara:
> Hmmm, issue with changing to uint32 appears to be in cdrizzle.tdriz.

Sigh

PR #7312, opened 2022-10-24
JP-2982: Fix extract_2d to skip zero-length cutouts

hbushouse:
Resolves [JP-2982](https://jira.stsci.edu/browse/JP-2982)

This PR updates the extract_2d step so that it skips processing of any 2D cutouts that are defined with a length of <1 pixel in either of their dimensions, as computed from the WCS-related info for the cutout.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/extract_2d/grisms.py
+++ b/jwst/extract_2d/grisms.py
@@ -422,7 +422,7 @@ def extract_grism_objects(input_model,
             # don't extract anything that ended up with zero dimensions in one axis
             # this means that it was identified as a partial order but only on one
             # row or column of the detector
-            if ymax - ymin > 0 and xmax - xmin > 0:
+            if ymax - ymin > 0.5 and xmax - xmin > 0.5:
                 subwcs = copy.deepcopy(inwcs)
                 log.info("Subarray extracted for obj: {} order: {}:".format(obj.sid, order))
                 log.info("Subarray extents are: "
--- a/jwst/extract_2d/grisms.py
+++ b/jwst/extract_2d/grisms.py
@@ -422,7 +422,7 @@ def extract_grism_objects(input_model,
             # don't extract anything that ended up with zero dimensions in one axis
             # this means that it was identified as a partial order but only on one
             # row or column of the detector
-            if ymax - ymin > 0.5 and xmax - xmin > 0.5:
+            if ymax - ymin > 0 and xmax - xmin > 0:
                 subwcs = copy.deepcopy(inwcs)
                 log.info("Subarray extracted for obj: {} order: {}:".format(obj.sid, order))
                 log.info("Subarray extents are: "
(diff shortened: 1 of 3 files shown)


hbushouse:
So far I've been able to answer one of my questions, specifically why the xmin/xmax values shown for the objects above are coming out in floating-point values, rather than whole pixels. The bbox for each object is first computed to the nearest whole pixel indexes here in the create_grism_bbox function:

https://github.com/spacetelescope/jwst/blob/master/jwst/assign_wcs/util.py#L752

However, just a few lines later the x or y limits for the object are recomputed if the object is classified as a point source:

https://github.com/spacetelescope/jwst/blob/master/jwst/assign_wcs/util.py#L761

The new limits are left in floating-point indexes. Not sure if that was intentional or not.

hbushouse:
I've also discovered that the way these float limits are then later truncated to integer pixel indexes results in the wrong total width for point sources that have their limits set by the `wfss_extract_half_height` parameter. The intention was that if wfss_extract_half_height=N, the total width would be 2*N+1. But we are getting widths of 2*N.

jwst/extract_2d/grisms.py
```diff
@@ -447,8 +447,8 @@ def extract_grism_objects(input_model,
                                                  ycenter_model &
                                                  order_model) | tr
                 from gwcs.utils import _toindex
```
nden, line 449: Not sure why this is here. If not specific reason it can go at the top of the module.

nden (approved):
LGTM. I expect failures in testing.

jwst/extract_2d/grisms.py
hbushouse: Moved to top in latest commit.

hbushouse:
Regression test run shows expected differences in sizes of the 2D cutouts for many WFSS sources: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/451/  No errors or unintented changes.

jwst/extract_2d/grisms.py
```diff
@@ -441,14 +442,14 @@ def extract_grism_objects(input_model,
...
                 tr = Mapping((0, 1, 0, 0, 0)) | (Shift(xmin) & Shift(ymin) &
                                                  xcenter_model &
                                                  ycenter_model &
                                                  order_model) | tr
-                from gwcs.utils import _toindex
-                y_slice = slice(_toindex(ymin), _toindex(ymax))
-                x_slice = slice(_toindex(xmin), _toindex(xmax))
+
+                y_slice = slice(_toindex(ymin), _toindex(ymax) + 1)
+                x_slice = slice(_toindex(xmin), _toindex(xmax) + 1)
```
tapastro, line 451: Is there any worry with making both axes inclusive then running into a slice of e.g. [n:2049] in the dispersion direction, putting you off the detector? I may not be comprehending this change fully.

hbushouse: That had occurred to me too, but theoretically it should not happen. The coords we get back from all the transforms are zero-indexed, so the highest that xmax/ymax should ever be is 2047 (or NAXISn - 1). So when we add 1 to end slice limit the max we should ever get is 2048, which is valid for a data array of size 2048.

tapastro: Interesting - I only ask because this asdf issue with WFSS data has shown some cutouts (in log outputs, at least) to be something like xmin: 1255, xmax: 2048, so I wasn't sure if we were already reaching the edge before this PR. I'll just fetch master and make sure nothing breaks. 🙂

PR #7323, opened 2022-10-28
separate tox testenvs to cache CRDS reference files retrieved by initial test

zacharyburnett:
This PR sets up dependent CI jobs and attempts to cache CRDS reference files, succeeding https://github.com/spacetelescope/jwst/pull/6867 with suggestions from https://github.com/spacetelescope/jwst/pull/6867#pullrequestreview-1159731951.

**Checklist for maintainers**
- [ ] added relevant label(s)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,12 +83,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-      - run: pip install tox crds
-      - run: echo "CRDS_CONTEXT=$(crds list --operational-context)" >> $GITHUB_ENV
-        if: github.event.inputs.crds_context == ''
-      - run: echo "CRDS_CONTEXT=${{ github.event.inputs.crds_context }}" >> $GITHUB_ENV
-        if: github.event.inputs.crds_context != ''
-      - run: crds sync --contexts ${{ env.CRDS_CONTEXT }}
+      - run: pip install tox
       - uses: actions/cache@v3
         with:
           path: ${{ env.CRDS_PATH }}
@@ -116,12 +111,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-      - run: pip install tox crds
-      - run: echo "CRDS_CONTEXT=$(crds list --operational-context)" >> $GITHUB_ENV
-        if: github.event.inputs.crds_context == ''
-      - run: echo "CRDS_CONTEXT=${{ github.event.inputs.crds_context }}" >> $GITHUB_ENV
-        if: github.event.inputs.crds_context != ''
-      - run: crds sync --contexts ${{ env.CRDS_CONTEXT }}
+      - run: pip install tox
       - uses: actions/cache@v3
         with:
           path: ${{ env.CRDS_PATH }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,18 +25,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        toxenv: [ check-style, check-security, check-install ]
+        python-version: [ '3.9' ]
+        os: [ ubuntu-latest ]
         include:
           - name: Code style check
-            os: ubuntu-latest
-            python-version: 3.9
             toxenv: check-style
           - name: Security audit
-            os: ubuntu-latest
-            python-version: 3.9
             toxenv: check-security
           - name: Verify install_requires in setup.py
-            os: ubuntu-latest
-            python-version: 3.9
             toxenv: check-install
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        toxenv: [ test ]
+        toxenv: [ test-xdist ]
         python-version: [ '3.x' ]
         os: [ ubuntu-latest ]
     steps:
@@ -111,7 +111,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        toxenv: [ test-pyargs ]
+        toxenv: [ test-pyargs-xdist ]
         python-version: [ '3.x' ]
         os: [ ubuntu-latest ]
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,6 @@ jobs:
           - name: Verify install_requires in setup.py
             toxenv: check-install
     steps:
-      - run: echo ${{ format('{0}/{1}', env.CRDS_PATH, 'references') }}
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
@@ -65,7 +64,7 @@ jobs:
       - uses: actions/cache@v3
         with:
           path: ${{ env.CRDS_PATH }}
-          key: crds-reference-files-${{ hashFiles(format('{0}/{1}', env.CRDS_PATH, 'references')) }}
+          key: crds-reference-files-${{ hashFiles(format('{0}/references/*', env.CRDS_PATH)) }}
       - run: tox -e ${{ matrix.toxenv }}
   test:
     name: test (${{ matrix.toxenv }}, Python ${{ matrix.python-version }}, ${{ matrix.os }})
@@ -88,7 +87,7 @@ jobs:
       - uses: actions/cache@v3
         with:
           path: ${{ env.CRDS_PATH }}
-          key: crds-reference-files-${{ hashFiles(format('{0}/{1}', env.CRDS_PATH, 'references')) }}
+          key: crds-reference-files-${{ hashFiles(format('{0}/references/*', env.CRDS_PATH)) }}
       - run: tox -e ${{ matrix.toxenv }}
       - if: ${{ contains(matrix.toxenv,'-cov') }}
         uses: codecov/codecov-action@v3
@@ -116,7 +115,7 @@ jobs:
       - uses: actions/cache@v3
         with:
           path: ${{ env.CRDS_PATH }}
-          key: crds-reference-files-${{ hashFiles(format('{0}/{1}', env.CRDS_PATH, 'references')) }}
+          key: crds-reference-files-${{ hashFiles(format('{0}/references/*', env.CRDS_PATH)) }}
       - run: tox -e ${{ matrix.toxenv }}
   build:
     name: build distribution (${{ matrix.toxenv }}, Python ${{ matrix.python-version }}, ${{ matrix.os }})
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -19,86 +25,134 @@ env:
...
+      - run: pip install tox crds
+      - run: echo "CRDS_CONTEXT=$(crds list --operational-context)" >> $GITHUB_ENV
+        if: github.event.inputs.crds_context == ''
+      - run: echo "CRDS_CONTEXT=${{ github.event.inputs.crds_context }}" >> $GITHUB_ENV
+        if: github.event.inputs.crds_context != ''
+      - run: crds sync --contexts ${{ env.CRDS_CONTEXT }}
+      - uses: actions/cache@v3
+        with:
+          path: ${{ env.CRDS_PATH }}
+          key: crds-reference-files-${{ hashFiles(format('{0}/references/*', env.CRDS_PATH)) }}
```
jdavies-st, line 78: Perfect.

tox.ini
```diff
@@ -36,45 +58,38 @@ passenv =
...
 commands =
-    !cov: pytest -n auto {posargs}
-    cov: pytest -n auto --cov-report=xml --cov=. --cov-config=setup.cfg {posargs}
-
+    pip freeze
+    pytest -n auto \
+    cov: --cov=. --cov-config=setup.cfg --cov-report=term-missing --cov-report=xml \
+    warnings: -W error \
+    regtests: --bigdata --slow --basetemp={homedir}/test_outputs \
+    {posargs}
```
jdavies-st, line 68: Is the `\` in each of the 4 lines all pointing to `{posargs}`?  This is a confusing syntax.  Wouldn't be clearer to have:

```
pytest -n auto {posargs}
cov: --cov=. --cov-config=setup.cfg --cov-report=xml {posargs}
...
```

etc?

tox.ini
```diff
@@ -36,45 +58,38 @@ passenv =
...
     TEST_BIGDATA
-
 usedevelop = true
-
 commands =
-    !cov: pytest -n auto {posargs}
-    cov: pytest -n auto --cov-report=xml --cov=. --cov-config=setup.cfg {posargs}
-
+    pip freeze
+    pytest -n auto \
```
jdavies-st, line 64: Generally using `pytest -n auto` I don't think is desirable.  This is only useful for running regtests, which actually doesn't run on Github actions but runs under Jenkins, where there are dozens of cores available.  Github actions only has a couple cores available.

.github/workflows/ci.yml
```diff
@@ -19,86 +25,134 @@ env:
...
 jobs:
-  tox:
+  check:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        toxenv: [ check-style, check-security, check-install ]
+        python-version: [ '3.9' ]
```
jdavies-st, line 35: Might want to move checks to py3.10.

.github/workflows/ci_cron.yml
```diff
@@ -25,32 +25,9 @@ jobs:
...
-      - name: Run tests
-        run: tox -e ${{ matrix.toxenv }}
-
-      - name: Upload coverage to codecov
-        if: ${{ contains(matrix.toxenv,'-cov') }}
-        uses: codecov/codecov-action@v3
-        with:
-          file: ./coverage.xml
-          flags: unit
-          fail_ci_if_error: true
```
jdavies-st, line 56: Getting rid of coverage here is probably fine, but you should be using the cache action here too for restoring the CRDS cache data, as this job routinely fails because of CRDS access, as it did this morning.

jdavies-st (commented):
This looks good!  Some comments below.

tox.ini
zacharyburnett: These are so that several different args might be added or not added; the line might be any one of 
```
pytest -n auto {posargs}
pytest -n auto --cov=. --cov-config=setup.cfg --cov-report=term-missing --cov-report=xml {posargs}
pytest -n auto --cov=. --cov-config=setup.cfg --cov-report=term-missing --cov-report=xml -W error {posargs}
pytest -n auto -W error {posargs}
pytest -n auto -W error --bigdata --slow --basetemp={homedir}/test_outputs {posargs}
```
etc.

.github/workflows/ci.yml
```diff
@@ -19,86 +25,134 @@ env:
...
+          fetch-depth: 0
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - run: pip install tox crds
+      - run: echo "CRDS_CONTEXT=$(crds list --operational-context)" >> $GITHUB_ENV
+        if: github.event.inputs.crds_context == ''
+      - run: echo "CRDS_CONTEXT=${{ github.event.inputs.crds_context }}" >> $GITHUB_ENV
+        if: github.event.inputs.crds_context != ''
+      - run: crds sync --contexts ${{ env.CRDS_CONTEXT }}
```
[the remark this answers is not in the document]
jdavies-st: Hmm, yeah, that's a problem, and always a challenge with using the Github cache action.  I struggled with it here

https://github.com/spacetelescope/jwst/pull/6165/files

and eventually removed the caching.  The challenge is to get the cache created the first time, and then you're good to go.  I was having the problem of the non-pytest checks finishing first, then caching nothing.

tox.ini
jdavies-st: Ah, I understand now.  That's clever.  Carry on.  🚀

tox.ini
jdavies-st: Ah, that's clever!  Carry on.  🚀

.github/workflows/ci.yml
```diff
@@ -19,86 +19,160 @@ env:
...
-            os: ubuntu-latest
-            python-version: 3.9
-            toxenv: twine
-
           - name: Code style check
-            os: ubuntu-latest
-            python-version: 3.9
-            toxenv: style
-
+            toxenv: check-style
```
jdavies-st, line 33: These and the other 2 are already in the matrix above.  Probably no need to add them here.

.github/workflows/ci.yml
```diff
@@ -19,86 +19,160 @@ env:
...
-          pip install tox
-
-      - name: Run tests
-        run: tox -e ${{ matrix.toxenv }}
-
-      - name: Upload coverage to codecov
-        if: ${{ contains(matrix.toxenv,'-cov') }}
+      - run: pip install tox
+      - run: tox -e ${{ matrix.toxenv }}
+  quick_test:
```
jdavies-st, line 47: Is the sole purpose of this quick_test to generate a cache that the other tests can use?

.github/workflows/ci.yml
```diff
@@ -19,86 +19,160 @@ env:
...
+          path: ${{ env.CRDS_PATH }}
+          key: crds-reference-files-${{ env.REFERENCE_FILES_HASH }}
+      - run: echo ${{ env.REFERENCE_FILES_HASH }} > reference_files_hash.txt
+      - uses: actions/upload-artifact@v3
+        with:
+          name: reference_files_hash
+          path: reference_files_hash.txt
+  test:
+    name: ${{ matrix.toxenv }} (Python ${{ matrix.python-version }}, ${{ matrix.os }})
+    needs: [ quick_test ]
```
jdavies-st, line 77: Since these tests don't even kick off until `quick_test` is finished, this is going to ~double the runtime of the test suite.

.github/workflows/ci_cron.yml
```diff
@@ -25,32 +25,13 @@ jobs:
...
-      - name: Install tox
-        run: |
-          pip install tox
-
-      - name: Run tests
-        run: tox -e ${{ matrix.toxenv }}
-
-      - name: Upload coverage to codecov
-        if: ${{ contains(matrix.toxenv,'-cov') }}
-        uses: codecov/codecov-action@v3
```
jdavies-st, line 52: Removing coverage reporting here is probably a good idea since nothing is changing on `master` when this cron job runs.

jdavies-st (commented):
This is a hard nut to crack.

I really like the refactor of `tox.ini`.  Structurally much better and easier to work with, and importantly, a good distinction between checks and tests.  🚀 

I don't like the `needs: [ quick_test ]` dependency, as this dramatically increases the runtime of the test suite, likely doubling it.  And it seems the main reason behind it is to get CRDS caching to work.  Would be better to get the caching to work and keep the flat, concurrent set of checks we already have, I think.

The main problem with the CRDS cache is that we are trying to restore a cache whose contents we are using to make a hash.  But it needs to be there before we can hash it, but we need to hash it before it is restored.  Chicken, egg.

The next-best thing would be to hash files that exist after repo checkout, or some other state that corresponds with a state of the CRDS cache.  While the cache of files needed for the tests will not change with every increment of the default context, it's probably the best proxy we have.  Of course it needs to be created first and cached before it can be used, so still the chicken-egg problem, which I recall having a helluva time trying to get sorted when I first implemented (then unimplemented) this caching long ago.

I've been pinging some ideas around today and might have a solution based off the nice tox refactor you did here.  Look for a PR.

.github/workflows/ci.yml
```diff
@@ -19,86 +19,160 @@ env:
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        toxenv: [ test-xdist, test-oldestdeps, test-sdpdeps-xdist ]
+        python-version: [ '3.8', '3.9', '3.10', '3.11' ]
+        os: [ ubuntu-latest, macos-latest ]
```
jdavies-st, line 84: And I don't think we should be running on macos.  The runners are too slow, and they don't test anything extra compared to the linux runners.  If our test suite was 3 minutes, fine.  But it is over an hour on the macos runners.

Likewise, oldestdeps probably only needs to run on the oldest supported version of python (and in fact probably only can), and the sdpdeps probably only should run on whatever version of python DMS/SDP is currently using, python 3.9 currently, i.e. whatever our JenkinsfileRT says.

.github/workflows/ci.yml
```diff
@@ -19,86 +19,184 @@ env:
...
-      - name: Run tests
-        run: tox -e ${{ matrix.toxenv }}
-
-      - name: Upload coverage to codecov
-        if: ${{ contains(matrix.toxenv,'-cov') }}
+      - run: pip install tox
+      - uses: actions/cache@v3
+        with:
+          path: .tox
+          key: tox-${{ matrix.toxenv }}-${{ env.pythonLocation }}-${{ runner.os }}-${{ hashFiles('pyproject.toml', 'setup.*') }}
```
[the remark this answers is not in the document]
jdavies-st: Oh, nevermind, I see this has been removed.  👍

jdavies-st:
The other thing to keep in mind is that caches created in branches cannot be used on master, though caches created on master can be used on branches.  So this complicates any rollout.  Probably will have to be 2 successive PRs.  Chicken.  Egg.

PR #7331, opened 2022-11-02
remove duplicate enum values from core.schema.yaml

braingram:
jsonschema draft 4 (used by asdf) states that enum values MUST be unique. This was unchecked prior to a recent commit python-jsonschema/jsonschema@2db9c58

[core.schema.yaml](https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/schemas/core.schema.yaml#L1061) contains duplicate dithering types for the different instruments (for example CYCLING for NIRSpec IFU and MIRI imaging)

This commit removes the duplicates in a way that might impair user readability of the file (as types are no longer logically grouped to instrument).

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/datamodels/schemas/core.schema.yaml
+++ b/jwst/datamodels/schemas/core.schema.yaml
@@ -1058,7 +1058,7 @@ properties:
               PFLAT_CORON1065, PFLAT_CORON1140, PFLAT_CORON1550, PFLAT_CORONLYOT,
 
               # MIRI imaging
-              CYCLING, REULEAUX, SPARSE-CYCLING,
+              REULEAUX,
               2-POINT, 4-POINT-SETS,
               2-POINT-MIRI-F770W-WITH-NIRCAM, 3-POINT-MIRI-F770W-WITH-NIRCAM,
               4-POINT-MIRI-F770W-WITH-NIRCAM, 9-POINT-MIRI-F770W-WITH-NIRCAM,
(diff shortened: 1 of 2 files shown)


hbushouse (commented):
Rather than just removing the duplicate entries, can we instead create a new commented section entitled something like "Common to \<mode1\> and \<mode2\>", where mode1, mode2 are things like "MIRI imaging" and "NIRCam", as appropriate for each instance? That way we still keep track of which instrument/mode each allowed value belongs too. Simply removing an entry like "2-POINT" from the MIRI MRS section leaves us (the humans) not knowing that "2-POINT" is valid for MIRI MRS.

jwst/datamodels/schemas/core.schema.yaml
```diff
@@ -1049,17 +1049,28 @@ properties:
...
             enum:
               # FGS in parallel with NIRCam and NIRISS
-             [FULL, FULLBOX, INTRAMODULE, INTRAMODULEBOX, INTRAMODULEX,
-              INTRASCA, SUBARRAY_DITHER, WFSC, AMI, IMAGING, MIMF, WFSS,
+              # shared with NIRCam: FULL, FULLBOX, INTRAMODULE, INTRAMODULEBOX,
+              #   INTRAMODULEX, INTRASCA, WFSC,
+              # shared with NIRISS: AMI, IMAGING, MIMF, WFSS,
+             [SUBARRAY_DITHER,
+
+              # common to NIRISS and FGS in parallel with NIRCam and NIRISS
```
braingram, line 1057: Thanks @hbushouse for looking over the proposed changes. I added comments in the form you suggested so common modes are separated. I also added comments listing the shared modes alongside exclusive modes so hopefully all modes for a given instrument are grouped.

braingram:
@hbushouse I updated the changelog. Let me know if there's anything else I can do to help with this pr. This is my first jwst PR so I'm not quite sure about labels and other items on the checklist. Please let me know if there's anything I should change or add.

hbushouse (approved):
I think this looks OK. Is there a way to confirm that we haven't lost any values in the process of doing the rearrangement? Just want to make sure we don't have any holes in coverage.

PR #7332, opened 2022-11-03
JP-2907 refactor item reprocessing for efficiency, refactor how background associations are configured

### LINKED ISSUE #7266: asn_generator runs for excessively long time
_Issue [JP-2907](https://jira.stsci.edu/browse/JP-2907) was created on JIRA by [Hien Tran](https://jira.stsci.edu/secure/ViewProfile.jspa?name=htran):_

ops saw another case for 2128:1 in which the asn_generator ran for a long time. unlike the case in <https://jira.stsci.edu/browse/JP-2861> this one eventually completed, but only after ~24 hrs.

the attached pool file shows that obs 1 has 1200 members for two instruments (prime MIR + NRC parallel) all of imaging type.

initial test run shows that the updated rules in jp-2861 didn't appear to help for this case.

### LINKED ISSUE #7268: refactor reprocessing list to do more conditioning
_Issue [JP-2902](https://jira.stsci.edu/browse/JP-2902) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

When items are to be reprocessed, the current logic creates an individual reprocessing list for each item. This is inefficient in 1) a bunch of reprocessing lists based on the same rule and 2) the same item being added numerous times, again with the same rule.

Reprocessing lists should go through some conditioning before the next list is taken on to remove duplicates lists and items.

stscieisenhamer:
Resolves [JP-2907](https://jira.stsci.edu/browse/JP-2907)
Resolves [JP-2902](https://jira.stsci.edu/browse/JP-2902)

Closes #7266
Closes #7268

During operations, it was found some association generation was taken ~24 hours to run. A refactor of how the processing lists are handled and how "blanket" associations, those that add target acquisitions and backgrounds, are defined have increased speed by a factor of 15.

During the association refactor, also fixed the following issues
- Miri Fixed-slit along-slit-nod not calculating correct background
- Duplicate members, mostly target acquisitions
- Improve both verbose and debug output from the command-line

**Checklist for maintainers**
- [x] added relevant label(s)


stscieisenhamer:
[Regression test](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/285/)

kmacdonald-stsci (approved):

hbushouse (approved):
I have relatively little idea how most of this is actually working, but it looks good to me!

stscieisenhamer:
Would it be inappropriate to respond: You and me both?

hbushouse:
> Would it be inappropriate to respond: You and me both?

Based on that, I'm now not sure we should blindly accept your statement that the regtest failures are OK.

stscieisenhamer:
Actually thanks for the reminder! Someone should review the changes in the Fixed Slit along-slit-nod results. Previously, these were not treated as the other nod modes, where the nods are backgrounds to each other. Now, they are. But confirmation that this is correct should be had.

hbushouse:
So how does one easily examine the changes in the regtest results? Is there a way to run locally and/or some how find the pile of asn files generated from a given run and see if the ones for MIRI LRS were constructed properly?

hbushouse:
I've done a more detailed examination of the new regtest results, as well as results for a nearly identical observation setup in recent in-flight data, and compared the two. Here's what I've found for the in-flight data.

In-flight program 01189 obs 002 uses MIRI LRS fixed-slit with the ALONG-SLIT-NOD pattern and 2 exposures at each nod/dither point, resulting in 4 exposures with 1-2 at nod position 1 and 3-4 at nod position 2. The asn generator creates 4 "spec2" asn's, one for each of the 4 exposures, where exposures 3+4 are listed as background members for exposures 1 and 2, and exposures 1+2 are listed as background members for exposures 3 and 4 (all as it should be). In addition, there is 1 "spec3" asn created:

jw01189-c1000_20220929t154421_spec3_001_asn.json: Contains the 4 "cal" products as science members, the 4 "x1d" products of the same exposures listed as background members, and 1 targ-acq member.

What's curious, however, is that pid-01189 obs-019 appears to have the identical setup (along-slit-node with 2 exposures per dither/nod), yet for that observation the asn generator creates the "spec3" ASN "jw01189-o019_20221015t225428_spec3_001_asn.json". It has exactly the same kind of content as the c1000 asn for obs-002 (4 "cal" science products, 4 "x1d" background products, and 1 targ-acq), but why did it get created as an observation-level ASN, while the one for obs-002 was created as a group ASN (c1000)?

The pool file "jw01189_20220929t154421_pool.csv" is available if anyone wants to investigate.

hbushouse:
The example observation(s) used in the regression test (pid-00720 obs-044 and obs-045) use a similar setup, with the along-slit-nod pattern and 2 exposures per dither/nod, but in this case obs-045 is added as a dedicated background observation (using the same setup as obs-044). So in this case the generator creates 4 "spec2" asns for obs-044, again correctly assigning exposures 3+4 as background for exposures 1 and 2, and vice versa. But for obs-045 (of a bkg target) there are no background members listed in the 4 "spec2" asns. This seems reasonable, in that the dedicated bkg target exposures should not get any bkg subtraction themselves.

A total of 3 "spec3" ASN's are created by the new code in this PR:

jw00720-o045_pool-030-mir-lrs-nods-bkg_spec3_00001_asn.json  is the spec3 asn for the dedicated bkg observation. It contains only the 4 "cal" products from the bkg target exposures, to be combined on their own into Level 3 products for the bkg target only. This is expected.

jw00720-o044_pool-030-mir-lrs-nods-bkg_spec3_00001_asn.json is the spec3 asn for the science target observation. It contains the 4 "cal" products for the science target, as well as the "x1d" products for those same 4 exposures of the science target, but listed as background members. So in this case the Level 3 product will use its own data to do master bkg subtraction. I don't think this is correct, for the reasons explained in the next comment.

jw00720-c1001_pool-030-mir-lrs-nods-bkg_spec3_00001_asn.json contains the 4 "cal" products of the science target, the 4 "x1d" products of the bkg target (obs-045) listed as background members, as well as the 4 "x1d" products of the science target also listed as background members. So now we're mixing science target and bkg target data for use in master bkg subtraction.

hbushouse:
Regarding the cases mentioned above, for both in-flight and regtest example observations, where the 4 "x1d" products of the science target are being included in the "spec3" ASN in order to use them for master bkg subtraction, at first I thought this might be a problem, but have determined that it isn't (sort of). The science target observations will have BKGTARG=False in their meta data, which will cause the master bkg step to use the extracted spectrum from the "background" column of the x1d files listed in the spec3 ASN. So that should be OK. Unfortunately right now the default extract1d ref file used for MIRI LRS fixed-slit data does not specify any background regions to extract, so the background column in the x1d tables are filled with zeros, hence no actual bkg will be subtracted. But no harm is done either.

hbushouse:
Well, I went back to MAST and retrieved the most recent version of the jw01189 pool file (the one I grabbed earlier didn't have exposures for obs-019 added to it yet) and when I manually reran asn_generate on it (using no options, and using the github master branch code, i.e. not including this PR), I now get only 1 "spec3" ASN for each of jw01189 obs-002 and obs-019 and they're both the observation-level ASN's (no group ASN's). So it appears that the problem with duplicate group and obs ASN's was already taken care of, at least for this observation scenario.

hbushouse:
So, given all of the above analysis, I now feel safe in declaring that the new results for MIRI LRS produced by this PR, where you get both a group and observation level spec3 ASN created - but they do NOT have identical content - is OK and hence I approve of this PR!

hbushouse:
For the record, I've put all the files I used to testing/analyzing in /grp/jwst/ssb/chartreuse/dms_bugs/jp-2907/

PR #7333, opened 2022-11-04
Cache CRDS_PATH reffile contents in CI workflows

jdavies-st:
This PR re-enables caching of the CRDS_PATH for the CI workflows, after having been turned off in #6165.  It should make the per-PR CI much more reliable (no failed reffile downloads) and slightly faster overall.

This takes advantage of @zacharyburnett 's excellent tox.ini refactor in PR #7323 (with some minor additions), and uses the new clean structure to distinguish which tox envs are running unit tests which use CRDS and which are not.

Here we use the CRDS_CONTEXT as the cache key.  The context will change more often than the contents of CRDS_PATH itself, but that's pretty much the only way to do it that makes sense, as we are trying to cache the snapshot of reffiles that are pulled during the running of the unit tests, and thus need a proxy for the contents of this snapshot before the tests are run.  We cannot hash the contents of CRDS_PATH as it doesn't exist until after the tests are run, but we need the hash before the tests are run.  Chicken.  Egg.

This means that every increment of the default CRDS_CONTEXT will mean that the first CI test run will not have a cached CRDS_PATH, but after it runs, it will create one, and then all subsequent runs will use it until the next CRDS_CONTEXT change.

I've tested this separately in the CI system by running one of the test matrix jobs, and results can be seen here:

https://github.com/jdavies-st/jwst/actions/workflows/ci.yml

particularly the "CI use caching" job numbers 89 through 92, where in 4 stages, respectively, the

1. tests run without cache, cache is created for the first time upon test completion
2. cache is successfully used the tests
3. CRDS_CONTEXT is (artificially) changed, so no cache is restored or used, but a new one is made after the tests finish
4. new cache is used on tests

I've previously also checked that running matrix works with the cache both when created for the first time and incremented with a new key.

Note:
- Caching CRDS_PATH theoretically reduces runtime, but not by much.  More importantly, it reduces the downloading of files from the crds server, which means the CI should be more reliable.  So this PR means slightly reduced runtime for unit tests overall, but much more reliable, i.e. higher likelihood of finishing with out CRDS download errors causing problems.
- This PR also caches the pip wheel cache for the `actions/setup-python@v4` which saves having to download the likely packages from pypi.  This works the same here as on one's local machine, which also maintains a pip cache of packages that have been pulled from pypi.  The install step still queries pypi.org to get the latest version of a package dependency and will download it if not in the pip cache.  This is all default pip behavior.  This should speed up installs a bit and has no downsides.
- We get the default CRDS_CONTEXT without having to separately install `crds` to do so by using the crds server json api.  This speeds things up considerably.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,64 +20,15 @@ env:
   CRDS_CLIENT_RETRY_DELAY_SECONDS: 20
 
 jobs:
-  tox:
-    name: ${{ matrix.name }}
+  check:
+    name: ${{ matrix.toxenv }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - name: Python 3.10 Tests
-            python-version: '3.10'
-            os: ubuntu-latest
-            toxenv: py310
-
-          - name: Latest dependency versions w/coverage
-            os: ubuntu-latest
-            python-version: 3.9
-            toxenv: py39-cov
-
-          - name: Oldest dependency versions
-            os: ubuntu-latest
-            python-version: 3.8
-            toxenv: py38-oldestdeps-cov
-
-          - name: SDP dependencies in requirements-sdp.txt
-            os: ubuntu-latest
-            python-version: 3.9
-            toxenv: sdpdeps
-
-          - name: Installed package with --pyargs
-            os: ubuntu-latest
-            python-version: 3.9
-            toxenv: pyargs
-
-          # - name: Verify install_requires in setup.py
-          #   os: ubuntu-latest
-          #   python-version: 3.9
-          #   toxenv: verify-install-requires
-
-          # - name: Build distribution
-          #   os: ubuntu-latest
-          #   python-version: 3.9
-          #   toxenv: twine
-
-          # - name: Code style check
-          #   os: ubuntu-latest
-          #   python-version: 3.9
-          #   toxenv: style
-
-          # - name: Security audit
-          #   os: ubuntu-latest
-          #   python-version: 3.9
-          #   toxenv: security
-
-          # MacOS job is flaky on Github actions and fails routinely.
-          # Re-enable when things get more reliable.
-          # - name: macOS
-          #   os: macos-latest
-          #   python-version: 3.9
-          #   toxenv: py39
+        toxenv: [check-style, check-security, check-dependencies, check-build]
+        python-version: ['3.x']
+        os: [ubuntu-latest]
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


nden:
Zach will review when he's back from vacation.

zacharyburnett (approved):
this LGTM! Are there any instances where we might want to explicitly run tests with a previous CRDS context with `workflow_dispatch`?

jdavies-st:
> Are there any instances where we might want to explicitly run tests with a previous CRDS context with `workflow_dispatch`?

Possibly!  Maybe a separate PR for that?  Here I'm mostly trying to make the CI fast and reliable.

Before this is merged, someone who has admin credentials for this repo will have to change which are the required tests to pass.  They've been renamed, and rejiggered here, so for instance merging is blocked because the "Code style check" test hasn't passed yet, but it's there, just labeled "CI / check-style".

Feedback on the CI check names/descriptions would be appreciated.  This changes them, hopefully in a more structured way.  Is this an improvement for the daily developers or just another irritating change?  We _can_ go back to the old names.

hbushouse (approved):
I have no idea if this code will actually do what you claim it does, but I'm very much in favor of the claim!

hbushouse:
> Before this is merged, someone who has admin credentials for this repo will have to change which are the required tests to pass. They've been renamed, and rejiggered here, so for instance merging is blocked because the "Code style check" test hasn't passed yet, but it's there, just labeled "CI / check-style".

Can you indicate which names have been changed, so that we can make sure to cover all of them when modifying the list of required passing tests?

> Feedback on the CI check names/descriptions would be appreciated. This PR changes them, hopefully in a more structured way. Is this an improvement for the daily developers or just another irritating change? We _can_ go back to the old names.

The new ones look OK to me.

zacharyburnett:
> After poking around in the page of required CI test settings, it appears that we may have to merge this PR before the new test names will show up and hence allow me to change the list. So any objections to merging this now? I have the power to merge this PR even though the (old) required tests are not passing. (it's good being king ...)

I'm fine with that, though perhaps we should rebase first to see if the `xdist-cov` test will pass with new main code

zacharyburnett:
are these errors important?
```python
metacls = <class 'enum.EnumType'>, cls = 'ListCategory'
bases = (<enum 'ListCategory'>,), classdict = {}, boundary = None
_simple = False, kwds = {}, ignore = ['_ignore_'], key = '_ignore_'

    def __new__(metacls, cls, bases, classdict, *, boundary=None, _simple=False, **kwds):
        # an Enum class is final once enumeration items have been defined; it
        # cannot be mixed with other types (int, float, etc.) if it has an
        # inherited __new__ unless a new __new__ is defined (or the resulting
        # class will fail).
        #
        if _simple:
            return super().__new__(metacls, cls, bases, classdict, **kwds)
        #
        # remove any keys listed in _ignore_
        classdict.setdefault('_ignore_', []).append('_ignore_')
        ignore = classdict['_ignore_']
        for key in ignore:
            classdict.pop(key, None)
        #
        # grab member names
>       member_names = classdict._member_names
E       AttributeError: 'dict' object has no attribute '_member_names'
```
https://github.com/spacetelescope/jwst/actions/runs/3489141823/jobs/5840043311#step:7:362

hbushouse:
Probably need @stscieisenhamer to take a look.

jdavies-st:
Odd error.  It was not there before `master` was merged in.  We were also not testing against Python 3.11 either.

I don't understand `jwst.associations`, but the bug is here:

https://github.com/spacetelescope/jwst/actions/runs/3489141823/jobs/5840043311#step:7:338

https://github.com/spacetelescope/jwst/blob/f6b3973c79f387253f7d09fd24c3a27aeb81cad8/jwst/associations/registry.py#L323

where a ListCategory of type ListCategory, which subclasses Enum is being dynamically created.  But you can't do this with Enums.  Here's the minimally reproducible example:

```python
>>> from enum import Enum
>>> type("foo", (Enum,), {})
  File "/Users/jdavies/miniconda3/envs/jwst/lib/python3.11/enum.py", line 482, in __new__
    member_names = classdict._member_names
                   ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute '_member_names'
```
and instead you should do

```python
>>> Enum("foo", {})
<enum 'foo'>
```

But maybe the bug is that `ListCategory` should not be there in the first place?  All the other rules added are rules and constraints.

stscieisenhamer:
No doubt I am being over-clever, as usual. However, point of information: If installing jwst from master, all tests succeed under Python 3.11. It is only this PR that the errors appears.  I can investigate further next week.

hbushouse:
> No doubt I am being over-clever, as usual. However, point of information: If installing jwst from master, all tests succeed under Python 3.11. It is only this PR that the errors appears. I can investigate further next week.

So ... something in this PR is causing the failures or what?

zacharyburnett:
Ok, then I think this PR should be merged now, and then another PR opened for fixes for Python 3.11 for those tests.

Alternatively, we could try and solve those issues here.

stscieisenhamer:
I'm good with merging this now and investigating the related errors later. @zacharyburnett however you wish to handle the out-of-date issue and then merge is good.

PR #7336, opened 2022-11-04
JP-2996: Remove duplicate application of pixel area to NRS_IFU extended data

tapastro:
Resolves [JP-2996](https://jira.stsci.edu/browse/JP-2996)

Closes #

This PR addresses an issue seen when corrections to the NRS_IFU pixel area reference file were tested. Changes to the pixel area of scale N were seen applied to the data array at scale N**2, indicating that the pixel areas were being divided out twice.

In testing edits to the code, it was noticed that the pixel area was being divided out of all NRS_IFU data, but because the current dummy files are filled with values of 1, it went unnoticed. Additionally, the pixel area division was being done in units of square arcseconds, which required adding a factor of A2_TO_SR to the code where the pixel area is divided out slice-by-slice. 

The duplicate removed was dividing the conversion factor by the mean pixel area across all slices, which was redundant but also imprecise.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,6 +47,10 @@ photom
 
 - Cutout pixel area array to match the subarray of the science data. [#7319]
 
+- Remove duplicated division of pixel area during photometric calibration
+  of NIRSpec IFU data with EXTENDED source type; correct units in pixel area
+  division to sr from square arcseconds [#7336]
+
 resample
 --------
 
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -309,8 +309,12 @@ class DataSet():
                 # Compute relative sensitivity for each pixel based
                 # on its wavelength
                 sens2d = np.interp(wave2d, waves, relresps)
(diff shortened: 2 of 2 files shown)


tapastro:
Regtests: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/456/

hbushouse (approved):
I think this appears to make sense now. Instead of dividing by the average pixel area when source_type is extended, it now divides by the area of individual pixels (and in the proper units). And only does the division once.

hbushouse:
The only failures in the regtest run were unrelated changes due to updated MIRI reference files. I guess this suggests that we don't have any NIRSpec IFU extended source data in our regtests?

tapastro:
I think our IFU regressions only cover extended source extraction, actually - but this issue only becomes apparent when the area reference file was changed. It would have also become a problem for point sources, had the area reference files not been filled with ones.

This looks like a good test to update with in-flight data...

PR #7343, opened 2022-11-09
JP-2912: Store offsets for WFSS source contamination correction in subarray metadata

### LINKED ISSUE #7269: Offset from full coordinates - wfss contamination step
_Issue [JP-2912](https://jira.stsci.edu/browse/JP-2912) was created on JIRA by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva):_

The contamination step needs the offset of the cutout from the origin of the full frame. This is currently encoded in the WCS transforms as an offset or free coefficient in the Polynomial, depending on the model type. The contamination step retrieves the offset by indexing the compound model and the index is hard coded. This is not a robust solution because if the overall compound models in the WCS pipeline changes in the future the index of the model with the offset of interest may change.

Find a better way to access this offset. One way would be to introduce a dedicated attribute to store the offset. This woul dbe done in assign_wcs when the compound model is created.

tapastro:
Resolves [JP-2912](https://jira.stsci.edu/browse/JP-2912)

Closes #7269 

This PR removes directly calling offsets from the grism_wcs transform's CompoundModel and instead adds them to the datamodel metadata for more robust querying. However, I'm not sure exactly what the wfss_contam code is doing, so I need some review/assistance on one of my edits.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/wfss_contam/disperse.py
+++ b/jwst/wfss_contam/disperse.py
@@ -57,9 +57,11 @@ def dispersed_pixel(x0, y0, width, height, lams, flxs, order, wmin, wmax,
         Whether to allow for the SED of the object to be extrapolated when it does not fully cover the
         needed wavelength range. Default if False.
     xoffset : int
-        Pixel offset to apply when computing the dispersion (accounts for padded direct images)
+        Pixel offset to apply when computing the dispersion (accounts for offset from source cutout to
+        full frame)
     yoffset : int
-        Pixel offset to apply when computing the dispersion (accounts for padded direct images)
+        Pixel offset to apply when computing the dispersion (accounts for offset from source cutout to
+        full frame)
 
     Returns
     -------
--- a/jwst/wfss_contam/disperse.py
+++ b/jwst/wfss_contam/disperse.py
@@ -81,16 +81,6 @@ def dispersed_pixel(x0, y0, width, height, lams, flxs, order, wmin, wmax,
     sky_to_imgxy = grism_wcs.get_transform('world', 'detector')
     imgxy_to_grismxy = grism_wcs.get_transform('detector', 'grism_detector')
 
-    # The WCS comes from the first 2D cutout in the grism image, which has
-    # the offset from the full-frame origin to the cutout origin in it.
-    # We'll use these values later to get back to full-frame coords.
-    try:
-        offset_x = -imgxy_to_grismxy.offset_1
-        offset_y = -imgxy_to_grismxy.offset_2
-    except AttributeError:
-        offset_x = -imgxy_to_grismxy.offset_6
-        offset_y = -imgxy_to_grismxy.offset_7
-
     # Setup function for retrieving flux values at each dispersed wavelength
     if len(lams) > 1:
         # If we have direct image flux values from more than one filter (lambda),
@@ -113,8 +103,8 @@ def dispersed_pixel(x0, y0, width, height, lams, flxs, order, wmin, wmax,
     # then finally convert to x/y in the grism image frame
     x0_sky, y0_sky = seg_wcs(x0, y0)
     x0_xy, y0_xy, _, _ = sky_to_imgxy(x0_sky, y0_sky, 1, order)
-    xwmin, ywmin = imgxy_to_grismxy(x0_xy + offset_x, y0_xy + offset_y, wmin, order)
-    xwmax, ywmax = imgxy_to_grismxy(x0_xy + offset_x, y0_xy + offset_y, wmax, order)
+    xwmin, ywmin = imgxy_to_grismxy(x0_xy + xoffset, y0_xy + yoffset, wmin, order)
+    xwmax, ywmax = imgxy_to_grismxy(x0_xy + xoffset, y0_xy + yoffset, wmax, order)
     dxw = xwmax - xwmin
     dyw = ywmax - ywmin
 
@@ -141,7 +131,7 @@ def dispersed_pixel(x0, y0, width, height, lams, flxs, order, wmin, wmax,
     # then convert to x/y in grism frame.
     x0_sky, y0_sky = seg_wcs([x0] * n_lam, [y0] * n_lam)
     x0_xy, y0_xy, _, _ = sky_to_imgxy(x0_sky, y0_sky, lambdas, [order] * n_lam)
-    x0s, y0s = imgxy_to_grismxy(x0_xy + offset_x, y0_xy + offset_y, lambdas, [order] * n_lam)
+    x0s, y0s = imgxy_to_grismxy(x0_xy + xoffset, y0_xy + yoffset, lambdas, [order] * n_lam)
 
     # If none of the dispersed pixel indexes are within the image frame,
     # return a null result without wasting time doing other computations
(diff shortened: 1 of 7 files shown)


jwst/wfss_contam/observations.py
```diff
@@ -186,9 +186,9 @@ def disperse_all(self, order, wmin, wmax, sens_waves, sens_resp, cache=False):
                 self.cached_object[i]['maxy'] = []
 
             # Disperse object "i"
-            self.disperse_chunk(i, order, wmin, wmax, sens_waves, sens_resp)
+            self.disperse_chunk(i, order, wmin, wmax, sens_waves, sens_resp, [0, 0])
```
tapastro, line 189: The offsets for this bit of code are the part I'm not sure of - it looks like prior to my changes, it was going the same route as source-by-source dispersion, so perhaps the proper offsets need to be passed down this chain for each source cutout as well? @hbushouse

tapastro:
I took Nadia's suggestion to add these offsets to the individual SlitModel's subarray xstart/ystart keywords (i.e. `model.slits[0].meta.subarray.xstart`), as they appear to be unused. The initial implementation had added keywords to the already busy wcsinfo branch of the metadata.

jwst/wfss_contam/disperse.py
```diff
@@ -113,8 +103,8 @@ def flux(x):
     # then finally convert to x/y in the grism image frame
     x0_sky, y0_sky = seg_wcs(x0, y0)
     x0_xy, y0_xy, _, _ = sky_to_imgxy(x0_sky, y0_sky, 1, order)
-    xwmin, ywmin = imgxy_to_grismxy(x0_xy + offset_x, y0_xy + offset_y, wmin, order)
-    xwmax, ywmax = imgxy_to_grismxy(x0_xy + offset_x, y0_xy + offset_y, wmax, order)
+    xwmin, ywmin = imgxy_to_grismxy(x0_xy + xoffset, y0_xy + yoffset, wmin, order)
```
hbushouse, line 106: I don't see where xoffset/yoffset are being set/initialized before they get used here.

hbushouse: Oh wait, they're passed in as arguments in the call to this function, right?

tapastro: Yes, prior to this PR they were unused inputs to this function.

hbushouse (commented):
OK, I'm trying to trace my way through the existing code to see where/when the old offset_x/offset_y were set and used to figure out where you need to be sending the new values. Originally they were only ever used within disperse.py/dispersed_pixel() and used to be grabbed from the "grism_wcs" that's passed as an argument to dispersed_pixel. But now they aren't in "grism_wcs" anymore, but in the slit[i].meta instead and that is NOT passed to dispersed_pixel, so you have to pass them in some other way.

"dispersed_pixel" is only ever called from observations.py using the "pars" object to provide all of the arguments to dispersed_pixel, including xstart/ystart, which is what you're now using to pass the offset values. So it would work to just populate self.xstart/self.ystart earlier in the calling routines, e.g. in Observation.__init__, but unfortunately they get used for something else before they are then included in the list of pars/args to pass to dispersed_pixel. So we don't want to overload them. 

So my suggestion at this point would be to pass xoffset/yoffset as *new* pair of additional args to dispersed_pixel and leave the existing xstart/ystart alone (even though they aren't being used right now down in dispersed_pixel). In wfss_contam.py/contam_corr() you could retrieve xoffset/yoffset from input_model.slits[0].meta at the same place where "grism_wcs" is copied from input_model.slits[0].meta.wcs, then pass them as 2 additional args (or a tuple) when initializing "obs = Observation(image_names, ...)", then in Observation.__init__() add them as "self.x/yoffset" (or as a "self.offset" tuple), and finally add them as 2 additional args when the "pars" object is built. They will then be 2 additional args passed to "dispersed_pixel", which can be utilized where needed.

Clear as mud?

hbushouse:
It really would've been easier to have them stuffed somewhere within "grism_wcs", because at least then they'd be immediately available where needed (rather than having to pass them along from function to function).

tapastro:
That follows the current functionality, except we don't appear to have much control of how things are stored under the wcs or its transforms, which has led to this PR...

It wasn't clear to me that an attribute of the Observation class would work for these parameters, as the offsets change between sources despite all being part of a single Observation.

tapastro:
Figured out the issues I created for myself - should be running properly now. Testing locally on some public data, with regtest set here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/471/

tapastro:
I think the last regtest run showed I was out of sync with master and needed a rebase. Second set running now:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/480/

tapastro:
Regtest exposed poor assumption that each source could only have one slit entry - incorrect for multi-order extractions. New set of tests here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/482/

tapastro:
Think I've got the regression test down to platform/compiler differences, so running one more set: 
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/483/

jwst/wfss_contam/wfss_contam.py
```diff
@@ -106,15 +108,24 @@ def contam_corr(input_model, waverange, photom, max_cores):
...
                       boundaries=[0, 2047, 0, 2047], max_cpu=ncpus)
 
+    # Create dict of offsets to pass to disperse_all to create simulated image
+    # with sources in correct locations
+    offset_dict = dict()
+    for slit in output_model.slits:
+        source_entry = offset_dict.get(int(slit.source_id), dict())
+        # Add entry in source dict for each order
+        source_entry[slit.meta.wcsinfo.spectral_order] = (slit.xstart - 1, slit.ystart - 1)
+        offset_dict[int(slit.source_id)] = source_entry
```
hbushouse, line 121: If there's more than one slit per source_id (i.e. multiple orders), will the dictionary have redundant entries and/or one entry overwriting another? Or is there another "dimension" to the dictionary entry for a given source_id that contains the unique data for each order?

hbushouse (commented):
Looks like it all makes sense now, modulo one question about how the offsets dictionary is constructed.

jwst/wfss_contam/wfss_contam.py
tapastro: I could add a comment here, but the dict.get() method will pull the entry for the `source_id` if it exists, or return an empty `dict()` if it is not yet in `offset_dict.keys()`. The next lines add a new entry to the dict and then assign the updated dict to the parent `offset_dict`. Not the cleanest, but it worked best with the code in `wfss_contam.observations.Observation.disperse_all`, where the simulated images are built by order - in this way, it checks the presence of a source, then checks on order before simulating the source into the full frame.

tapastro:
One more try 🤞 https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/485/

hbushouse (approved):
Looks good, I guess(?). Much simpler than the previous version, which is always good.

hbushouse:
Same old failures in the CI tests that are awaiting another PR to fix them, so I think this looks good and am going to merge.

PR #7345, opened 2022-11-10
JP-2972: Add doc about context image and a tool to decode it

mcara:
Resolves [JP-2972](https://jira.stsci.edu/browse/JP-2972)

This PR documents "context" arrays created by the resample step. In addition, it also adds a utility function, `decode_context()` that can be used to identify input images that have contributed to a resampled pixel.

**Checklist for maintainers**
- [x] added relevant label(s)


mcara:
If anyone is an expert in doctest, please let me know if

```
.. doctest::
     :options: +SKIP
```

in [main.rst](https://github.com/spacetelescope/jwst/pull/7345/files#diff-271e04bc4e69bfa861825f876710d367c4d7c7bed56c84ba4fda57ed594e7dd4R88-R89) is the right way of turning off doctest for the next block. When docs are build locally, unfortunately they look like this:
doctest-skip

mcara:
From https://sphinx-rtd-trial.readthedocs.io/en/1.0.8/ext/doctest.html:


> **.. doctest:: [group]**
> A doctest-style code block. You can use standard doctest flags for controlling how actual output is compared with what you give as output. ...
> 
> This directive supports two options:
> 
> - **hide**, a flag option, hides the doctest block in other builders. ...
> 
> - **options**, a string option, can be used to give a comma-separated list of doctest flags that apply to each example in the tests. (You still can give explicit flags per example, with doctest comments, but they will show up in other builders too.)
>

mairanteodoro:
> 

@mcara I assume the issue with skipping doctest has been resolved, correct?

mairanteodoro (approved):
Looks good to me.

stsci-hack (commented):
The actual code and implementation look really good and efficient!  It would be worth responding to my comment about the documentation before merging this PR, though.

mcara:
SCI,1 and SCI,2 are so HST :) Not applicable to JWST.

docs/jwst/resample/main.rst
```diff
@@ -34,10 +34,75 @@ WCS of each input image and the WCS of the defined output product.
...
+pixel in the resampled data array. Context image uses 32 bit integers to encode
+this information and hence it can keep track of only 32 input images.
+First bit corresponds to the first input image, second bit corrsponds to the
+second input image, and so on. If the number of input images is larger than 32,
+then it is necessary to have multiple context images ("planes") to hold
+information about all input images
+with the first plane encoding which of the first 32 images contributed
+to the output data pixel, second plane representing next 32 input images
+(number 33-64), etc. For this reason, context array is a 3D array of the type
+``numpy.int32`` and shape ``(np, ny, nx)`` where ``nx`` and ``ny``
```
hbushouse, line 53: not sure we need to say the type is "**numpy**.int32". Isn't just generic "int32" sufficient?

hbushouse: Oh and actually isn't it *unsigned* 32-bit, i.e. uint32?

hbushouse (commented):
Looks good. Just a couple of minor comments and questions.

docs/jwst/resample/main.rst
mcara: 1. `numpy.int32` is a real and actual type of the context array. There is no generic `int32` in Python.
2. I think we should modify the type of the array from `int32` to `uint32`. But since it this PR deals primarily with documenting context arrays + a convenience function, I thought to leave change of type to a separate PR. Moreover, changing type would involve modifying image model's schema.

hbushouse: Sounds like a plan.

hbushouse (approved):

PR #7349, opened 2022-11-14
Update NIRSpec IFU spec2 regtest with in-flight data

hbushouse:
This PR updates the existing test_nirspec_ifu_spec2 regression test to use in-flight data (a single IFU exposure).

It also updates 1 line of the extract_1d/ifu.py module to switch the use of the deprecated `np.bool` to plain `bool` (gets rid of an execution time warning).

**Checklist for maintainers**
- [ ] added relevant label(s)

### DIFF
--- a/jwst/extract_1d/ifu.py
+++ b/jwst/extract_1d/ifu.py
@@ -435,7 +435,7 @@ def extract_ifu(input_model, source_type, extract_params):
     dq = np.zeros(shape[0], dtype=np.uint32)
 
     # This boolean mask will be used to mask any bad voxels in a given plane
-    bmask = np.zeros([shape[1], shape[2]], dtype=np.bool)
+    bmask = np.zeros([shape[1], shape[2]], dtype=bool)
 
     # If the user supplied extraction center coords, use them and
     # ignore all other source type and source position values
--- a/jwst/regtest/test_nirspec_ifu_spec2.py
+++ b/jwst/regtest/test_nirspec_ifu_spec2.py
@@ -10,11 +10,11 @@ TRUTH_PATH = 'truth/test_nirspec_ifu'
 
 @pytest.fixture(scope='module')
 def run_spec2(jail, rtdata_module):
-    """Run the Spec2Pipeline on a single exposure"""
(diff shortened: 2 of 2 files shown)


hbushouse:
Regression test run: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/461/

tapastro (approved):
Do you know if this file has source type set to extended by default? It may be worth keeping the extended test to validate both extraction methods.

Otherwise, LGTM.

hbushouse:
Yes, the new exposure does have SRCTYPE='EXTENDED', so there's no need for that additional (separate) check now. That's why the individual test of the photom step was removed. Should've pointed that out in the description.

tapastro:
> Yes, the new exposure does have SRCTYPE='EXTENDED', so there's no need for that additional (separate) check now. That's why the individual test of the photom step was removed. Should've pointed that out in the description.

Any value in setting up a second test forcing source_type to 'POINT', just for testing purposes? It will add coverage of some code that extended source extraction doesn't touch.

hbushouse:
> > Yes, the new exposure does have SRCTYPE='EXTENDED', so there's no need for that additional (separate) check now. That's why the individual test of the photom step was removed. Should've pointed that out in the description.
> 
> Any value in setting up a second test forcing source_type to 'POINT', just for testing purposes? It will add coverage of some code that extended source extraction doesn't touch.

I suppose that couldn't hurt. I could make a test that uses the same input, but forces POINT source type. I'll work on that.

hbushouse:
Latest commit adds back the photom test and also includes an extract_1d test, both using an input that's flagged as point source (the full calwebb_spec2 test uses an extended source). I've determined that photom and extract_1d are the only 2 steps that are source-type dependent, at least for IFU data.

jemorrison:
Looks good to me

tapastro (approved):
Looks like you're providing two new files, one for each of the tests of point-source data processing - is that right?

Looks good - I figured you'd just run spec2 and specify the source type in the srctype step params, but I guess that goes through unnecessary cube building.

PR #7352, opened 2022-11-15
(JP-2991) Require asdf-transform-schemas greater than 0.3.0

dmggh:
Resolves [JP-2991](https://jira.stsci.edu/browse/JP-2991)

Closes #

This PR requires the asdf transform version to be at least 0.3.0, by updating setup.cfg. This was prompted by a user unable to open files from MAST using the JWST data models .

**Checklist for maintainers**
- [X] updated relevant documentation
- [X] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,4 @@
-1.8.4 (unreleased)
+1.8.5 (unreleased)
 ==================
 
 assign_wcs
@@ -34,13 +34,6 @@ datamodels
 
 - Remove duplicates and add comments to core.schema dithering types [#7331]
 
-documentation
--------------
-
-- Update deprecation notice with copyedit changes [#7348]
-
-- Clarify how to manage a local CRDS cache [#7350]
-
 
 extract_1d
 ----------
@@ -59,7 +52,13 @@ flatfield
 
 - JP-2993 Update the flat-field ERR computation for FGS guider mode exposures to
   combine the input ERR and the flat field ERR in quadrature. [#7346]
+
   
+general
+-------
+
+- Add requirement for asdf-transform-schemas >= 0.3.0 [#7352]
+
 guider_cds
 ----------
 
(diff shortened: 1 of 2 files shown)


dmggh:
I suspect "asdf-transform" in the dependency is not quite correct - suggestions ?

nden:
It's `asdf_transform_schemas`.

CHANGES.rst
```diff
@@ -91,6 +91,11 @@ resample
   images that have contributed with flux to an output (resampled)
   pixel. [#7345]
 
+transforms
+----------
+
+- Require the asdf transform version to be at least 0.3.0 [#7352]
```
hbushouse, line 97: Let's be very explicit here in terms of exactly what's being referenced. It should say "Add requirement for asdf-transform-schemas >= 0.3.0".  I think this should also go under a heading of "general" in the change log, because it's an installation/setup item.

setup.cfg
```diff
@@ -28,6 +28,7 @@ setup_requires =
     setuptools_scm
 install_requires =
     asdf>=2.13.0
+    asdf_transform_schemas>=0.3.0
```
hbushouse, line 31: Needs to be `asdf-transform-schemas` (with dashes, not underscores)

tapastro:
However, in python, importing requires the underscores and rejects the dashes... 🤔 I'd guess that we want the underscores here, as it's dealing with the package in a python environment.

hbushouse:
> However, in python, importing requires the underscores and rejects the dashes... 🤔 I'd guess that we want the underscores here, as it's dealing with the package in a python environment.

Huh, really? We've got other entries in setup.cfg that use dashes, such as:
    spherical-geometry>=1.2.22
    asdf-astropy>=0.2.2

nden:
I think both spellings are valid in the setup file. However, when importing a package underscores are necessary. To keep it consistent we can spell it with dashes here.

dmggh:
So I will go with "asdf-transform-schemas" - are we in agreement ?

hbushouse (approved):
Looks good. Just needs a rebase to fix conflicts (I'm assuming with the change log).

dmggh:
These code coverage errors appear to be unrelated.

PR #7363, opened 2022-11-22
JP-2988: Update ramp_fit tests for change in saturation flag propagation

kmacdonald-stsci:
Resolves [JP-2988](https://jira.stsci.edu/browse/JP-2988)

Closes #

This PR addresses the updated way the saturation flags propagate through to the rateints and rates product.  This update affected CI tests that needed to be changed to reflect the new way the saturation flag gets set.  In the rateints product, the saturation flag only gets set if all groups in an integration are saturated; in this case the integration also gets flagged with the DO_NOT_USE flag.  In the rates product, the saturation flag only gets set if all integrations are marked SATURATED; in this instance, the pixel also gets flagged as DO_NOT_USE.

**Checklist for maintainers**
- [x] added relevant label(s)


hbushouse (approved):
Code changes look "GOOD" to me. Add a change log entry (in the jwst repo) that references the actual ramp_fit code changes in the stcal PR, so that we have a record of that on the jwst side.

hbushouse:
I assume all of the CI failures won't go away until the changes in stcal get merged and released.

kmacdonald-stsci:
> I assume all of the CI failures won't go away until the changes in stcal get merged and released.

Yes.  The STCAL PR https://github.com/spacetelescope/stcal/pull/125 will need to be merged first.  Further, even after merging, I expect many regression test failures due to the new data quality propagation rules.

CHANGES.rst
```diff
@@ -85,6 +85,12 @@ photom
   of NIRSpec IFU data with EXTENDED source type; correct units in pixel area
   division to sr from square arcseconds [#7336]
 
+ramp_fitting
+--------
```
hbushouse, line 89: Still need to fix the underline here, so that it matches the length of the text above.

CHANGES.rst
```diff
@@ -85,6 +85,12 @@ photom
...
+
+- Change the propagation of the SATURATED flag to be done only for complete
+  saturation. [#7363]
```
hbushouse, line 92: Oh, and also reference the stcal PR number here, in addition to 7363.

hbushouse:
stcal 1.2.2 has just been released, which contains the updates to the ramp_fit step needed to support this PR. So add to this PR a change to the setup.cfg file that pins stcal to >=1.2.2, and then this should be mergeable.

CHANGES.rst
kmacdonald-stsci: Fixed.

CHANGES.rst
kmacdonald-stsci: I linked the STCAL PR number here.

kmacdonald-stsci:
> stcal 1.2.2 has just been released, which contains the updates to the ramp_fit step needed to support this PR. So add to this PR a change to the setup.cfg file that pins stcal to >=1.2.2, and then this should be mergeable.

I updated the setup.cfg file to reflect the new STCAL release.

hbushouse:
I believe most of the failing CI tests are due to recent changes in the way the "cv2" package is imported, which may not be included in this branch yet.

hbushouse:
Checking with @zacharyburnett to see if the change to the way `cv2` is imported in the jump step code in `stcal` might be causing these problems.

zacharyburnett:
It looks like this slipped past the downstream tests for `stcal`; the issue is that `stcal` is not unconditionally requiring `opencv-python`, but the `jump_step` tests are using `stcal` code that makes ellipses / circles with it. I've made a PR #7377 that adds it to testing, and also conditionally skips these tests if it's not installed.

hbushouse:
@zacharyburnett So once #7377 is merged and we then rebase this PR, the errors should go away (theoretically)?

zacharyburnett:
@hbushouse yes, assuming the tests pass over there

hbushouse:
Yay, CI tests look good now, except for the unrelated errors in xdist. So I'm merging.

jdavies-st:
This PR is connect to the merging of spacetelescope/stcal#125

PR #7374, opened 2022-12-01
JP-2900 Redux : Sort combine1d wavelength array prior to creation of spectral wcs

tapastro:
Resolves [JP-2900](https://jira.stsci.edu/browse/JP-2900)

Closes #

This PR addresses a bug found in reprocessing JP-2900 after asdf 2.14.1 was released, where one of the WFSS sources creates a non-monotonic wavelength array in combine_1d. I don't fully comprehend the creation process of the wavelength array, so this fix may be hasty.

However, in checking our regtests for c1d outputs, there is only one regression test covering combine1d outputs! Apart from NIS_WFSS, no other spec3 product checks against c1d files. Might fix that here too, or in a related PR.

**Checklist for maintainers**
- [x] added relevant label(s)


hbushouse (approved):
Simple enough change, which should do the trick.

Needs a change log entry.

I think the main reason that combine_1d is only covered by 1 regression test is the fact that the only mode that uses the step is WFSS. All other spec modes do their combining at the 2D "image" level and then create a single 1D extracted spectrum from that combined 2D product. And no combined products are created at all for TSO modes.

hbushouse:
Have you done a regtest run?

tapastro:
> Have you done a regtest run?

My focus has been pinballing between a few issues, and I left this as a bit of a draft PR. I'll finish it up now (changelog/regtests). I suppose one test may be enough, but I kind of figured we'd have a check against c1d output for at least both wfss modes. Maybe it isn't necessary.

hbushouse:
> I suppose one test may be enough, but I kind of figured we'd have a check against c1d output for at least both wfss modes. Maybe it isn't necessary.

Once we have valid reference files and test data to support both NIRCam and NIRISS WFSS, we should make tests for both.

tapastro:
Regression tests: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/479

hbushouse:
Getting CI failures that should be fixed by #7377. Once that PR is merged, you'll need to rebase this PR.

hbushouse:
CI's look OK now after updating from master (and hence pulling in the ramp_fit changes).

PR #7375, opened 2022-12-02
JP-3009: Add cleanups for files left behind while running tests

WilliamJamieson:
Resolves [JP-3009](https://jira.stsci.edu/browse/JP-3009)

This PR adds fixtures to a few tests so that those tests do not leave files generated and saved while running the tests behind. The unit tests for JWST should not leave behind "artifacts" after running as this could interfere with re-running tests and/or lead to those files accidentally being picked up by git.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/associations/tests/test_level3_wfscmb.py
+++ b/jwst/associations/tests/test_level3_wfscmb.py
@@ -80,14 +80,14 @@ def jitter_present(jitter, associations):
     return False
 
 
-def test_level3_wfscmb_jitter_suppression():
+def test_level3_wfscmb_jitter_suppression(tmp_path):
     """
     Make sure no candidate from the pool with DMS_NOTE equal to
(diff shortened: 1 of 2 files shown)


kmacdonald-stsci (approved):

hbushouse:
Is this essentially a fix for [JP-3009}(https://jira.stsci.edu/browse/JP-3009), or does this only cover a subset of tests with this problem?

WilliamJamieson:
> Is this essentially a fix for [JP-3009](https://jira.stsci.edu/browse/JP-3009), or does this only cover a subset of tests with this problem?

Didn't know at all about JP-3009, I just noticed this while locally testing #7376 and it bothered me. I believe this does in fact resolve JP-3009!

hbushouse (approved):

PR #7376, opened 2022-12-02
Add memray support

WilliamJamieson:
This PR adds usage of the `pytest-memray` plugin to the tests. This is part of the effort to begin tracking memory usage in JWST (and other packages, see [SCSB-57](https://jira.stsci.edu/browse/SCSB-57).

Some questions:

- Should we turn `pytest-memray` on by default? Unfortunately, it currently is not supported by Windows and will cause the unit tests to fail during collection if it is used there.
- Does using `pytest-memray` need documentation somewhere in the JWST docs?
- Should we start decorating some of the tests with memory usage limits, see the `pytest-memray` [docs](https://pytest-memray.readthedocs.io/en/latest/usage.html#markers)? (Maybe this should be done in a series of follow-up PRs)

In a future version of `pytest-memray` there should be support for catching memory leaks, see bloomberg/pytest-memray#52. This maybe very useful for catching future memory issues.

Note that until bloomberg/pytest-memray#60 is resolved, we should not merge this PR.

**Checklist for maintainers**
- [ ] added relevant label(s)


setup.cfg
```diff
@@ -78,6 +78,7 @@ test =
     pytest>=6.0.0
     pytest-cov>=2.9.0
     pytest-doctestplus>=0.10.0
+    pytest-memray
```
kmacdonald-stsci, line 81: Does this require a version number?

kmacdonald-stsci (approved):

stscieisenhamer:
Unless there are cases/reasons to check memory usage during CI, I would think we do not want it on by default.  However, if this feature can be test-specific, developing some canary-in-the-mine tests would be useful.

stscieisenhamer (approved):

PR #7391, opened 2022-12-09
JP-3019: Remove code trimming zero-valued cube planes

drlaw1558:
Resolves [JP-3019](https://jira.stsci.edu/browse/JP-3019)

Given other updates in the past couple years this is no longer necessary, and always producing cubes of known size is helpful for both science analyses and MAST.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,7 +42,7 @@ cube_build
   from 0 to nan. [#7337]
 
 - Remove code trimming zero-valued planes from cubes, so that cubes of fixed length will always
-  be produced. Move nan-value setting to below spectral tear cleanup. []
+  be produced. Move nan-value setting to below spectral tear cleanup. [#7391]
 
 datamodels
 ----------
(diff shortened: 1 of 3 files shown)


jemorrison (commented):
Looks good to me

jemorrison (approved):

hbushouse:
So, just curious, instead of trimming out wavelength planes and/or spaxels that have no good data, they are now kept and all set to NaN?

hbushouse:
Is there any mention of trimming wavelength planes in the docs that now needs updating to indicate that they are NOT trimmed?

drlaw1558:
@hbushouse That's right, now they will be NaNs instead.  There is thus the possibility that we have some all-NaN cubes, but that it itself should be informative.  I looked for any documentation that might need to be updated already- the only relevant section was the one changed above (specifying NaN values aren't exclusively outside the footprint if there's bad data).

hbushouse:
Looks like @jemorrison ran regression tests against this PR branch at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/489/ and I'd say all of the differences look like they're expected based on these code changes. So I think this is ready for prime time.

PR #7405, opened 2022-12-20
JP-2013 Fix NIRSpec IFU NOD background subtract and include leakcal in the Level2 associations

stscieisenhamer:
WIP [JP-2013](https://jira.stsci.edu/browse/JP-2013)

This PR addresses the association creation portion of JP-2013. There were two problems with the rule Asn_Lv2NRSIFUNod:

- Mistakenly other dithers and mosaic tiles were being included in the association
- Leakcals/imprint were not included in assocations

The rule has been modified to

- Include leakcal/imprint exposures
- Limit the science/backgrounds to just the nodded dithers for a particular pointing.

**Checklist for maintainers**
- [x] added relevant label(s)
- [ ] ~Make sure the JIRA ticket is [resolved properly](https://github.com/spacetelescope/jwst/wiki/How-to-resolve-JIRA-issues)~
- [x] Comment related JIRA tickets as appropriate.

### DIFF
--- a/jwst/associations/lib/rules_level2b.py
+++ b/jwst/associations/lib/rules_level2b.py
@@ -835,7 +835,7 @@ class Asn_Lv2NRSIFUNod(
         - Spectral-based NIRSpec IFU multi-object science exposures
         - Single science exposure
         - Handle 2 and 4 point background nodding
-        - Included related imprint exposures
+        - Include related imprint exposures
     """
 
     def __init__(self, *args, **kwargs):
--- a/jwst/associations/lib/rules_level2b.py
+++ b/jwst/associations/lib/rules_level2b.py
@@ -864,10 +864,6 @@ class Asn_Lv2NRSIFUNod(
                 name='mosaic_tile',
                 sources=['mostilno'],
             ),
-            DMSAttrConstraint(
-                name='expcount_science',
-                sources=['expcount'],
-            ),
         ])
 
         # Now check and continue initialization.
--- a/jwst/associations/lib/rules_level2b.py
+++ b/jwst/associations/lib/rules_level2b.py
@@ -835,6 +835,7 @@ class Asn_Lv2NRSIFUNod(
         - Spectral-based NIRSpec IFU multi-object science exposures
         - Single science exposure
         - Handle 2 and 4 point background nodding
+        - Included related imprint exposures
     """
 
     def __init__(self, *args, **kwargs):
@@ -854,15 +855,19 @@ class Asn_Lv2NRSIFUNod(
                 force_unique=False
             ),
             DMSAttrConstraint(
-                name='expspcin',
-                sources=['expspcin'],
-            ),
-            DMSAttrConstraint(
                 name='patttype',
                 sources=['patttype'],
                 value=['2-point-nod|4-point-nod'],
                 force_unique=True
-            )
+            ),
+            DMSAttrConstraint(
+                name='mosaic_tile',
+                sources=['mostilno'],
+            ),
+            DMSAttrConstraint(
+                name='expcount_science',
+                sources=['expcount'],
+            ),
         ])
 
         # Now check and continue initialization.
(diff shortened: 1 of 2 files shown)


stscieisenhamer:
[Regression test](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/296/)

tapastro (commented):
When I test this on the pool under `leaknod` in the ticket directory, I see e.g. an association with 8 members: one science, three background and four imprint. Is that desired behavior? `calwebb_spec2` seems to only want one imprint per association.

jwst/associations/lib/rules_level2b.py
```diff
@@ -835,6 +835,7 @@ class Asn_Lv2NRSIFUNod(
         - Spectral-based NIRSpec IFU multi-object science exposures
         - Single science exposure
         - Handle 2 and 4 point background nodding
+        - Included related imprint exposures
```
hbushouse, line 838: typo: "Included" -> "Include"

hbushouse:
> When I test this on the pool under `leaknod` in the ticket directory, I see e.g. an association with 8 members: one science, three background and four imprint. Is that desired behavior? `calwebb_spec2` seems to only want one imprint per association.

Yes, this will definitely require updates to the `imprint_subtract` step. When the step is called from `Spec2Pipeline` it passes the names of all imprint members (as a list in the 2nd argument), but the step itself tries to treat it as a single model.

For the case of dithered exposures, with imprints taken at each dither position, is the desired behavior to average all the imprints and then subtract the average (as is done for backgrounds), or is it supposed to be one-to-one science-minus-imprint?  If so, then @tapastro is right that the current behavior of including multiple imprint members in the spec2 ASN is not correct.

stscieisenhamer:
The behavior described in JP-2013 is that when there are leakcals at each dither position, it is a one-to-one correspondence between the science/background and its corresponding leakcal. The leakcals do not get averaged or processed in any other way.

See [James' comment](https://jira.stsci.edu/browse/JP-2013?focusedId=620702&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-620702) for details.

kmacdonald-stsci (approved):

hbushouse:
> The behavior described in JP-2013 is that when there are leakcals at each dither position, it is a one-to-one correspondence between the science/background and its corresponding leakcal. The leakcals do not get averaged or processed in any other way.
> 
> See [James' comment](https://jira.stsci.edu/browse/JP-2013?focusedId=620702&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-620702) for details.
> 
> This is why there are four leakcals: one for the science dither, and three for the three background dithers.

Ah, so the presence of 4 leakcal members in the spec2 ASN file is expected and correct, but it's up to the `Spec2Pipeline` to properly make use of them all, such that it subtracts 1 leakcal from each of the 4 (1 science + 3 bkg) on-sky exposures. How will the pipeline know which leakcal goes with which on-sky exposure?  Had a similar issue when updating the `tweakreg` step to allow for a source catalog for each input science image: how to indicate which catalog goes with which image. That was solved by making the name of the catalog file an attribute of the science member itself.

jemorrison:
The pipeline is supposed to look at the dither number to know which leakcal to subtract from the science or background image. I am currently testing that.

hbushouse:
> The pipeline is supposed to look at the dither number to know which leakcal to subtract from the science or background image. I am currently testing that.

In that case, I hereby approve this PR.

hbushouse (approved):

hbushouse:
Started another regtest run at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/514 because the previous run had too many hundreds of failures to sort out.

jemorrison (approved):
Seems reasonable - but association rule code is not my strong point

hbushouse:
Regression test run was clean (other than expected errors).

PR #7408, opened 2022-12-21
Fix incorrect deallocation attempts of unallocated memory in cube builder

mcara:
This PR addresses fixes several bugs in how C code used by the cube builder may attempt to deallocate unallocated memory resulting in core dump. This happens when memory allocation fails for one array and then code attempts to deallocate several other arrays never allocated before. Due to this bug a memory allocation error is never raised but instead the whole process exits.

This PR also fixes a bug in `alloc_flux_arrays()` due to which 1 is not returned on memory allocation failure resulting in the code not detecting memory allocation failures.

Help desk issue INC0184892.

**Checklist for maintainers**
- [x] added relevant label(s)
- [x] ran regression tests: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/511/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,6 +44,10 @@ cube_build
 - Remove code trimming zero-valued planes from cubes, so that cubes of fixed length will always
   be produced. Move nan-value setting to below spectral tear cleanup. [#7391]
 
+- Fix several bugs in memory management in the C code for cube build which
+  would result in attempts to deallocate memory that was never allocated
+  resulting in core dump. [#7408]
+
 datamodels
 ----------
 
--- a/jwst/cube_build/src/cube_match_internal.c
+++ b/jwst/cube_build/src/cube_match_internal.c
@@ -6,12 +6,12 @@ each slice are assumed to not overlap on the output plane with pixels from other
 breaks the 3-D representation of the pixel's coordinates in the output plane to 2D. Standard
 drizzling type routines are used to calculate the percentage of each pixel's overlap with
 the output spaxels. The percentage of the pixel area  overlapped onto an output
-spaxel is used to weight the pixel flux for determining the spaxel flux. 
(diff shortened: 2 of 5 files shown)


drlaw1558:
Interesting; I've seen this kind of crash quite a few times, and thought it was simply what happened when the desired cubes were too big.  Is the thought that, with these new fixes, the pipeline will now fail on such big data cubes with a more informative error message?

hbushouse (approved):
Looks reasonable to me (once I was able to find the few lines of actual new code, buried in all the generic clean-up changes).

hbushouse:
Regression test failures are unrelated.

mcara:
> Looks reasonable to me (once I was able to find the few lines of actual new code, buried in all the generic clean-up changes).

@hbushouse This was not intentional but my editor automatically strips unnecessary white space at the end of lines. Also tabs->spaces, etc. Helps keep things in agreement with PEP8. I know this is C code but the editor simply strips trailing white spaces.

mcara:
> Interesting; I've seen this kind of crash quite a few times, and thought it was simply what happened when the desired cubes were too big. Is the thought that, with these new fixes, the pipeline will now fail on such big data cubes with a more informative error message?

Yes, this will accomplish two things: 1) not crash the Python interpreter/notebook kernel (so that you can simply re-run a command instead of re-launching Python session) and 2) provide a more informative message.

hbushouse:
> > Looks reasonable to me (once I was able to find the few lines of actual new code, buried in all the generic clean-up changes).
> 
> @hbushouse This was not intentional but my editor automatically strips unnecessary white space at the end of lines. Also tabs->spaces, etc. Helps keep things in agreement with PEP8. I know this is C code but the editor simply strips trailing white spaces.

Not a problem. I like clean code!

Issue #7409, opened 2022-12-21
Opencv dependency in stcal being optional causes uncaught error in jump step


m-samland:
After a clean install of the newest version (1.8.5), running the Detector1 pipeline now crashes if parameter `jump.use_ellipses = True`. This is because `stcal` requires opencv-python to work, which is now an [optional requirement and not automatically installed](https://github.com/spacetelescope/stcal/commit/6fe77933f83419b50cd3360c4c9f6fc392aecb05). I'm not sure why this packages was made optional, but since it was, I think there should either be a fallback that does not require opencv, or proper error handling that tells JWST pipeline users why the pipeline is failing if functions are used that require opencv. Right now it simply fails throwing a generic NameError on the line the function is called. (Or maybe it should just be added in again as a requirement, but I am not privy to the reasons it was removed in the first place.)

There is a warning that is raised when opencv cannot be imported in stcal, but it doesn't actually cause an Exceptions when an opencv function is called and opencv is not installed.

zacharyburnett:
I made `opencv` optional because certain users had difficulty building it on their platforms (https://github.com/spacetelescope/stcal/pull/126) which was blocking installation; you are right, however, that the warning is not descriptive enough. I'll work on a better exception flow, and also a fallback to analogous `scikit-image` ellipse functions

zacharyburnett:
https://github.com/spacetelescope/stcal/pull/136 adds better / more descriptive warnings to this situation. I am currently working on porting analogous functionality to `scikit-image`, which should be installed alongside `jwst`, in https://github.com/spacetelescope/stcal/pull/138

jdavies-st:
This is still broken on the latest release.  My workflow on MacOS 10.15 Catalina:

```
$ conda create -n jwst python=3.10 -qy
$ conda activate jwst
(jwst) $ pip install jwst
...
[...]

PR #7384, opened 2022-12-08
changes for `tox>=4.0` release

zacharyburnett:
Closes #7383 

This PR addresses the issue raised in #7383 

**Checklist for maintainers**
- [ ] added entry in `CHANGES.rst` within the relevant release section
- [x] updated or added relevant tests
- [ ] updated relevant documentation
- [ ] added relevant milestone
- [x] added relevant label(s)
- [ ] ran regression tests, post a link to the Jenkins job below.
      [How to run regression tests on a PR](https://github.com/spacetelescope/jwst/wiki/Running-Regression-Tests-Against-PR-Branches)
- [ ] Make sure the JIRA ticket is [resolved properly](https://github.com/spacetelescope/jwst/wiki/How-to-resolve-JIRA-issues)

PR #7414, opened 2022-12-22
test opencv functionality separately

zacharyburnett:
Addresses #7409
related to https://github.com/spacetelescope/stcal/pull/136

This PR runs the tests requiring `opencv` in a separate `toxenv` with `opencv-python` installed, to test whether conditional usage of ellipse construction is handled gracefully.

**Checklist for maintainers**
- [ ] added relevant label(s)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,9 @@ jobs:
           - toxenv: test-oldestdeps-xdist-cov
             os: ubuntu-latest
             python-version: '3.8'
+          - toxenv: test-opencv-xdist
+            os: ubuntu-latest
+            python-version: '3.10'
     steps:
       - uses: actions/checkout@v3
         with:
--- a/jwst/jump/tests/test_jump_step.py
+++ b/jwst/jump/tests/test_jump_step.py
@@ -310,7 +310,8 @@ def test_three_group_integration(generate_miri_reffiles, setup_inputs):
                               override_readnoise=override_readnoise, maximum_cores='none')
     assert out_model.meta.cal_step.jump == 'COMPLETE'
 
-@pytest.mark.skipif(not OPENCV_INSTALLED, reason="`opencv-python` not installed")
+
+@pytest.mark.xfail(not OPENCV_INSTALLED, reason="`opencv-python` not installed")
 def test_snowball_flagging_nosat(generate_nircam_reffiles, setup_inputs):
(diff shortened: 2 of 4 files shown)


hbushouse (approved):

jdavies-st:
This PR breaks CRDS caching between CI jobs.  The reason is that the `opencv` job only tests some very small subset of the unit test suite and pulls virtually nothing from CRDS, yet in uses and contributes to the general caching used by the jobs that run all the unit tests.  Because this job is short, it finishes before any of the other tests and then sets the cache when a new key become available (incremented CRDS context).  And of course uses virtually no reference files, so it is creating a mostly empty cache (instead of the normal ~350MB one).  The normal unit tests never get a change to set the cache using the new key.

The way to fix this is to move this `test-opencv` tox job into a different job in the workflow that does not use the cache action.

zacharyburnett:
> This PR breaks CRDS caching between CI jobs. The reason is that the `opencv` job only tests some very small subset of the unit test suite and pulls virtually nothing from CRDS, yet it uses and updates the general caching used by the jobs that run all the unit tests. Because this job is short, it finishes before any of the other tests and then sets the cache when a new key become available (incremented CRDS context). And of course uses virtually no reference files, so it is creating a mostly empty cache (instead of the normal ~350MB one). The normal unit tests never get a chance to set the cache using the new key.
> 
> The way to fix this is to move this `test-opencv` tox job into a different job in the workflow that does not use the cache action.

This should be fixed by adding a dedicated caching job in https://github.com/spacetelescope/jwst/pull/7384/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f

jdavies-st:
No, it is still broke, and that dedicated caching job makes it worse, as it runs `crds sync --contexts` in the unit test jobs, something that is never needed, as that only pulls over the imap, rmap and pmaps (~1.8MB), not any actual reference files (~350MB needed for the unit tests).  Most of the *maps are not needed, and none of the needed reference files are cached.  See the output of the caching job here:

https://github.com/spacetelescope/jwst/actions/runs/4009075221/jobs/6884024021#step:5:19

It has been broke since this PR.  It used to restore ~350MB of reference files:

https://github.com/spacetelescope/jwst/actions/runs/3758608010/jobs/6387151447#step:6:20

See the discussion here about why `crds sync --contexts` is not necessary:

https://github.com/spacetelescope/jwst/pull/7323#discussion_r1009544177

PR #7425, opened 2023-01-06
wfss_contam use datamodel in with context

braingram:
When the DataModel is cleaned up (__del__ is called) during garbage collection it will close any opened hdulists. This will make any subsequent attempts to load data from the hdulist fail (with ValueError: I/O operation on closed file).

DataModel prior to the below PR contained a self reference which created a reference cycle which made python fail to collect the object during generation 0 passes with the garbage collector. This meant that the hdulist was not immediately closed when the model fell out of scope.

Code in wfss_contam relied on this persistance as the model created in the following line is no longer referenced on subsequent lines.

dimage = datamodels.open(dir_image_name).data

python will attempt to clean up the opened data model and fail, keeping the hdulist (stored at dir_image_name) open. This allowed later lines that reference dimage to load the array data within the hdulist.

Fixing the reference cycle in stdatamodels in this PR: https://github.com/spacetelescope/stdatamodels/pull/109 makes the datamodel more easily garbage collected which causes wfss_contam to fail due to the hdulist closing when the model is collected.

To keep the model in scope, the above line is replaced with a with context to make it explicit that the datamodel is held open which it's contents are read.

This is work towards fixing JP-2798.

**Checklist for maintainers**
- [ ] added relevant label(s)


braingram:
Regression tests running here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/525/

hbushouse (commented):
Code looks good, as do all the CI and regtest results. But I would like it to have a change log entry (in CHANGES.rst). It's not a change in functionality or results, but enough of a change in the code that I'd like to have a record of it (as a bug fix).

CHANGES.rst
```diff
@@ -6,6 +6,11 @@ documentation
 
 - Remove references to pub server [#7421]
 
+wfss_contam
+-----------
+
+- Open image models in with context to keep models open while accessing contents [#7425]]
```
hbushouse, line 12: Perhaps put "with" in quotes, so folks know it's the name of an actual python code element (as opposed to a typo in the message!). There's also an extra "]" at the end of the line.

Oh, and this is not your fault, but the log file obviously didn't get updated to account for the fact that 1.9.2 got released on Wednesday. So can you give the 1.9.2 section a date of 2023-01-04, start a new section above it for 1.9.3 (marked as unreleased), and finally move this entry to the 1.9.3 section.

braingram: Thanks for looking it over. Sorry about the extra "]".

I made the updates to the changelog. Please let me know if they look ok.

hbushouse: Looks good. Thanks!

hbushouse (approved):

PR #7426, opened 2023-01-06
JP-2013 NIRSpec leakcal and nodded exposures 

### LINKED ISSUE #5911: Investigate ASN rules for NIRSpec leakcal and nodded exposures
_Issue [JP-2013](https://jira.stsci.edu/browse/JP-2013) was created on JIRA by [Mark Kyprianou](https://jira.stsci.edu/secure/ViewProfile.jspa?name=kyp):_

For 00675012, investigate the asn rules for the combination of leakcal and nodded exposures.

From Howard:

None of the leakcal exposures got included in any of the lvl2b ASN files (spec2 asn’s) for the science exposures (03102 and 03104) and hence none of the science exposures had the leakcal subtracted during level-2b processing. Off the top of my head I’m guessing this might be due to the fact that there’s a mixture of 2 different kinds of associated exposures included in obs 12:
 # The leakcal’s
 # Nodded exposures

 

The ASN rules are already setup to look for the 2-POINT-NOD dither pattern, which is in use here, and use the exposures from the opposite nod position as background exposures, to be subtracted from each science exposure as it goes through the calwebb_spec2 pipeline. The spec2 ASN files that were created for each science exposure do list the appropriate opposite nod exposures as background members. But they don’t include the leakcal exposures. I’m wondering if the ASN rules can only handle one situation at a time (e.g. either background or leakcal, but not both) and somehow the background association won out over the leakcals.

jemorrison:
Resolves [JP-2013](https://jira.stsci.edu/browse/JP-2013)

Closes #5911

This PR follows changes made in how level-2b associations are created for NIRSpec exposures with both leakcal and nodded exposures (#7405). The calspec2 pipeline has been changed to subtract leakcal images from science and background data before the background step is run on the science data. 

**Checklist for maintainers**
- [x] added relevant label(s)
      [How to run regression tests on a PR]: results

https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/534/


jemorrison:
Running regressions tests now

jemorrison:
The regression test show 8 failures. 4 failures are from tests that has an extra keyword S_IMPRINT SKIPPED. 
In those 4 causes the imprint step should not be run because the data is not NIRSPEC IFU data. I am not sure why the S_IMPRINT was not set to SKIP before. I down loaded on of the test data: test_miri_lrs_slit_spec. I ran calspec2 on the data and by default the IMPRINT step is set to skip. Now it is written in the header and before it was not. Maybe something else changes to does this now ?
The 4 other tests are related to NIRISS SOSS and I do not think are related to this PR.

jwst/pipeline/calwebb_spec2.py
```diff
@@ -237,8 +238,26 @@ def process_exposure_product(
...
+
+        # Leakcal subtraction (imprint)  occurs before background subtraction on a per-exposure basis.
+        # If there is only one `imprint` member, this imprint exposure is subtracted from all the
+        # science and background exposures.  Otherwise, there will be as many `imprint` members as
+        # there are science plus background members. Which imprint to subtract from which science/background
+        # is determined by the matching values of the header keyword PATT_NUM, or
+        # `DataModel.meta.dither.position_number`.
+
+        # find the position number of all the leak cal observations
+        imprint_pos_no = self._imprint_pos_no(members_by_type)
```
stscieisenhamer, line 250: The function of identifying the imprints to use for the given input is entirely an internal function of the `imprint_step` itself. As such, this function should be moved from here to being defined in and being done by the `imprint_step` and removed from `calwebb_spec2` and removing the `imprint_pos_no` argument on the `imprint_step.process`.

This simplifies `calwebb_spec2` and keeps the internal workings of `imprint_step` contained.

Agreed that this does add some rerunning of code, since that determination will be done for every invocation of `imprint_step`. However, it is a low-cost rerun and the aforementioned simplification of the logic is worth it, IMHO.

hbushouse: The comment for this line of code says that you're finding the dither positions of the leakcal exposures, but `members_by_type` contains the list of *all* ASN members. Or is the `imprint_pos_no` function limiting the search to only leakcal members?

jwst/pipeline/calwebb_spec2.py
```diff
@@ -237,8 +238,26 @@ def process_exposure_product(
...
+
+        calibrated = self.imprint_subtract(calibrated, members_by_type['imprint'], imprint_pos_no)
```
hbushouse, line 252: I think I agree with @stscieisenhamer that all of the functionality associated with figuring out dither position numbers and linking leakcals to other members should be contained within the `imprint` step itself, rather than adding all of that boiler plate code up here in the pipeline module. That would also mean that the argument list to `imprint_subtract` could remain simpler, in that it should only need to be sent `calibrated` and `members_by_type`. The population of the `imprint_pos_no` list could be done down within the step, as opposed to constructing it up here and then passing it down. It isn't needed anywhere in the pipeline module itself, only in the `imprint` step.

jwst/pipeline/calwebb_spec2.py
```diff
@@ -237,8 +238,26 @@ def process_exposure_product(
...
+
+        # for each background image subtract an associated leak cal
+        for i, bkg_file in enumerate(members_by_type['background']):
```
hbushouse, line 255: Why are there 2 calls to `imprint_subtract` (one in line 252 and another here in this loop)? Couldn't there just be one call to the step and it would then do the subtraction from all science and background members?

docs/jwst/imprint/description.rst
```diff
@@ -9,8 +9,8 @@ MOS and IFU exposures by the MSA structure. This is accomplished by
 subtracting a dedicated exposure taken with all MSA shutters closed and the
 IFU entrance aperture blocked.
 
-The step has two input parameters: the target exposure and the imprint
-exposure. These arguments can be provided as either a file name
+The step has three input parameters: the target exposure, the imprint
+exposure, and the dithered position number of the imprint exposure. These arguments can be provided as either a file name
```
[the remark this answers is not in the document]
jemorrison: fixed

jwst/pipeline/calwebb_spec2.py
jemorrison: Jonathan advised that the standard practice for a step is to operate on 1 data model. So that is why I designed it the way it is. If you prefer the imprint step to operate on science and backgrounds I can change the code.

jemorrison:
I re-ran the regression tests and now there are only 4 "failures" - these are just 1 difference in header that now has the IMPRINT step set to skipped. In these cases it should be set to skipped because the data is not NIRSpec IFU data

stscieisenhamer (approved):
LGTM.

hbushouse (approved):

PR #7430, opened 2023-01-13
JP-2813: Allow adjustments of imaging WCS in CAL-like data models

mcara:
Resolves [JP-2813](https://jira.stsci.edu/browse/JP-2813)

This PR adds a script `jwst/scripts/adjust_wcs.py` and a function to `jwst/tweakreg/utils.py` that allows users to apply adjustments to rotations defining an imaging WCS of calibrated images (`*_cal.fits`, `*_skymatch.fits`, `

**Checklist for maintainers**
- [x] added relevant label(s)


nden:
In case you are not aware of its existence and if you think it's useful - there's a file type check function in [stdatamodels.](https://github.com/spacetelescope/stdatamodels/blob/master/src/stdatamodels/filetype.py)

mcara:
@nden @hbushouse I am in particular interested about the choice of letters for the script's options:
```
    % adjust_wcs [-h] (-u | --suffix SUFFIX | -f FILE) [--overwrite] [-x] [-w]
                  [-r RA_DELTA] [-d DEC_DELTA] [-o ROLL_DELTA] [-s SCALE_FACTOR]
                  [-v] [input ...]
```

For example, is `-o` OK for the roll angle? or should I make it `-R`. But then I am afraid it will be easier to confuse it with `-r` used for `RA`. [Maybe] Should I remove these short options altogether for WCS adjustment parameters (RA, DEC, ROLL, SCALE)?

jwst/tweakreg/utils.py
```diff
@@ -0,0 +1,86 @@
...
+        Adjusted WCS object.
+    """
+    # find the last frame in the pipeline that starts with 'v2v3':
+    pipeline = deepcopy(wcs.pipeline)
+    for step in pipeline[::-1]:
+        if (
+            step.frame.name and
+            step.frame.name.startswith('v2v3') and
+            step.transform.name == 'v23tosky'
+        ):
```
nden, line 58: Is this necessary or can we just take the last step?

jwst/scripts/adjust_wcs.py
```diff
@@ -0,0 +1,320 @@
...
+    wcs_pars = [
+        options.ra_delta,
+        options.dec_delta,
+        options.roll_delta,
+        options.scale_factor
+    ]
+
+    if wcs_pars == [0.0, 0.0, 0.0, 1.0]:
+        logger.warning(
+            "All WCS adjustment parameters (delta RA, DEC, roll, and scale "
```
nden, line 216: Could you use the actual parameter names here. It may be a bit confusing that some of them are not "deltas"

jwst/scripts/adjust_wcs.py
```diff
@@ -0,0 +1,320 @@
...
+            "factor) have default values (identical transformation)."
+        )
+        logger.warning("Nothing to do. Quitting.")
```
nden, line 219: I think we are using the `warnings` module for warnings elsewhere, so let's stick to that.

jwst/scripts/adjust_wcs.py
```diff
@@ -0,0 +1,320 @@
...
+    if 0 in ftypes:
+        logger.error(f"Input file '{files[ftypes.index(0)]}' is neither "
+                     "an ASDF nor a FITS data file.")
+        logger.error("Script execution aborted.")
+        return
+
+    for fname, ftype in zip(files, ftypes):
+        data_model = jwst.datamodels.open(fname)
+
+        if jwst.datamodels.JwstDataModel == type(data_model):
```
nden, line 246: What use case does this check cover?

jwst/scripts/adjust_wcs.py
```diff
@@ -0,0 +1,320 @@
...
+            is_data_model = False
+            asdf_file = asdf.open(fname, mode='rw' if options.update else 'r')
+            in_wcs = asdf_file['wcs']
```
nden, line 249: I'm not sure I completely follow - if `asdf_file` is supposed to be e.g. a `cal.asdf` file opened with asdf, then the WCS object is in `asdf_file['meta']['wcs']. I may be misreading this.

jwst/tweakreg/utils.py
mcara: We could take the last step. This is just an extra check that that frame name and transform name have expected values, i.e., that we have a transformation that we expect to have. Probably it is possible to do a more rigorous check but I think proposed check is a minimal effort to verify WCS structure before proceeding.

jwst/scripts/adjust_wcs.py
mcara: Will do

jwst/scripts/adjust_wcs.py
mcara: `logger.warning` or `log.warning` are throughout the pipeline code. Just an example:

jwst/scripts/adjust_wcs.py
mcara: Both issues above are related and maybe there is a better solution/approach. The issue is how can I know where a WCS object in an `asdf` file is stored? In all JWST imaging data models, it is in `meta.wcs` (or `['meta']['wcs']`) but if saving a WCS manually in a "simple" `asdf` file (as opposite to a full data model). If trying to open a non-data-model file returns a `jwst.datamodels.JwstDataModel` then it is clear this is not a `ImageModel` or some other product created by the pipeline => it must be a "simple asdf" in which user stored just the WCS object.

I thought that maybe if an user saves just the WCS object to an `asdf` file then we could agree to save it in `['wcs']` key (which should have been documented, BTW). It seems counter-intuitive to require that WCS object be written at `['meta']['wcs']` for a non-data-model `asdf` file. But maybe we should require that WCS object be stored in `['meta']['wcs']` or maybe anywhere and have another parameter which could be used to indicate the location of the WCS object in an asdf file.

nden: > The issue is how can I know where a WCS object in an asdf file is stored? 

If you are talking about the user provided asdf file with a wcs object, I think we require that it is in the root of the tree, i.e. `asdf_file.tree['wcs']` or `asdf_file['wcs']`. We also require that the input file with the wcs is an asdf file and assume so. We are not allowing arbitrary files and location of the WCS but defining how the file should look like.

mcara: So, `if jwst.datamodels.JwstDataModel == type(data_model):` is essentially to distinguish between user-created asdf file and a valid `DataModel` (or its subclass)

mcara: > I think we require that it is in the root of the tree, i.e. asdf_file.tree['wcs'] or asdf_file['wcs']. We also require that the input file with the wcs is an asdf file and assume so.

Do you mean that this script should not support updating the WCS directly on a data model file? That is, should users extract the WCS from a data model, write it to an ASDF file, run this script, extract updated WCS and update the WCS in the data model and write the updated data model file? It think it would be much more convenient if we actually support data models directly as inputs.

jwst/scripts/adjust_wcs.py
```diff
@@ -0,0 +1,251 @@
...
+        )
+        logger.info("Nothing to do. Quitting.")
+        return
+
+    if options.expand:
+        files = []
+        for f in options.arg0:
+            files.extend(glob.glob(f))
+    else:
+        files = options.arg0
```
nden, line 207: Is the `expand` option necessary? Isn't `glob.glob(options.arg0)` equivalent to the logic here returning a list of file names in both the case of wild cards and single file input?

jwst/tweakreg/utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+    .. warning::
+        This function is not designed to handle neither FITS WCS nor
+        GWCS of resampled images. It is designed specifically for GWCS of
+        calibrated imaging data models that can be used as input to Stage 3
+        of the JWST pipeline (with suffixes '_cal', '_tweakreg', '_skymatch').
+
+    Parameters
+    ----------
+    wcs : gwcs.WCS
```
nden, line 23:
```suggestion
    wcs : `gwcs.WCS`
```

jwst/tweakreg/utils.py
```diff
@@ -0,0 +1,86 @@
...
+    scale_factor : float, optional
+        A multiplicative scale factor to be applied to the current scale
+        (if any) in the WCS. If input ``wcs`` does not have a scale factor
+        applied, it is assumed to be 1. The scale factor is applied in
+        a tangent plane perpendicular to the V1 axis of the telescope.
+
+    Returns
+    -------
+
+    wcs : gwcs.WCS
```
nden, line 48:
```suggestion
    wcs : `gwcs.WCS`
```

jwst/scripts/adjust_wcs.py
```diff
@@ -0,0 +1,251 @@
...
+# DAMAGE.
+#
+"""
+Apply adjustments to an imaging JWST GWCS object. Input files can be
+``DataModel``s serialized to ASDF or ASDF-in-FITS files, or they could be
+"simple" ASDF files storing just the GWCS model in the root of the ASDF file
+under the key ``'wcs'``, i.e., ``asdf_file.tree['wcs']``.
+
+Examples
+--------
```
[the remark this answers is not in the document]
mcara: No, it doesn't. Changing `%` to `>>>` below in examples would trigger doctest.

jwst/scripts/adjust_wcs.py
```diff
@@ -0,0 +1,251 @@
...
+        root = root + sep + new_suffix
+    base = root + ext
+
+    new_file_name = os.path.join(directory, base)
+
+    return new_file_name
+
+
+def main():
+    import argparse
```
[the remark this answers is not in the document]
mcara: No, will move up.

jwst/scripts/adjust_wcs.py
mcara: Yes, there could be multiple input files that all need to be globbed, for example:

```
adjust_wcs jw_f425_miri_*_cal.fits jw_f1080_nircam_*_tweakreg.fits
```

Yes, we could replace this with `jw_f*.fits` but allowing multiple input files enables a better/finer control control over file selection.

nden: In my opinion this is going too far and limiting it to one option will make the interface cleaner.

nden (approved):
Other than the opinion about the `expand` option I think this now looks good.

jwst/scripts/adjust_wcs.py
```diff
@@ -172,15 +164,14 @@ def main():
...
-                "only one input file allowed with argument -f/--file"
-            )
+    files = []
+    for f in options.arg0:
+        files.extend(glob.glob(f))
+
+    if len(files) > 1:
+        parser.error(
+            "argument -f/--file: not allowed with multiple input files"
+        )
```
nden, line 174: Is this. missing a check for the `-f` option or `options.file` as well?

CHANGES.rst
```diff
@@ -267,6 +267,12 @@ resample
 
 - Fixed a bug in the definition of the output WCS for NIRSpec. [#7359]
 
+scripts
```
hbushouse, line 270: These change log entries are buried way down in the 1.9.0 release section. They both need to be moved up to the current 1.9.5 (unreleased) section.

hbushouse (approved):
Looks good. Just need to fix the change log entries.

CHANGES.rst
mcara: Done

PR #7433, opened 2023-01-13
JP-2928: Update FGS image2 test with flight data

dmggh:
Partially resolves [JP-2928](https://jira.stsci.edu/browse/JP-2928)

This PR updates the FGS image2 reg test with flight data
**Checklist for maintainers**
- [X] added relevant label(s)
- [X] ran regression tests, post a link to the Jenkins job below.


CHANGES.rst
```diff
@@ -6,6 +6,11 @@ documentation
 
 - Remove references to pub server [#7421]
 
+guider_cds
```
hbushouse, line 9: This regtest runs the regular calwebb_image2 pipeline for FGS science-mode data and hence is not associated with the guider_cds step (which is used exclusively for FGS guide-mode data). So this entry should not have a guider_cds heading. It should be under calwebb_image2.

hbushouse (changes requested):

CHANGES.rst
hbushouse: Oh, and you also need a fresh version of the change log. The unrelased version is now up to 1.9.4 (1.9.3 has been released).

hbushouse (approved):

PR #7439, opened 2023-01-24
JP-3076 : Move `jwst.datamodels` to `stdatamodels.jwst.datamodels`

WilliamJamieson:
Resolves [JP-3076](https://jira.stsci.edu/browse/JP-3076)

This PR removes the datamodels in `jwst.datamodels` from `jwst` and moves them to `stdatamodels`, see matching PR spacetelescope/stdatamodels#112 for where they are being moved. It then turns the `jwst.datamodels` module into a wrapper that points to `stdatamodels.jwst.datamodels`, which raises a deprecation warning.

**Checklist for maintainers**
- [x] added relevant label(s)


docs/getting_started/quickstart.rst
```diff
@@ -72,7 +75,7 @@ be obtained. For more information, see :ref:`reference_files_crds`.
 **4. Running the Pipeline**
 
 With `jwst` installed and CRDS configured for JWST, you can now run the pipeline
-and use JWST `datamodels`.
+and use JWST `gatamodels`.
```
stscieisenhamer, line 78: `gatamodels`? guessing a bad keystroke here.

stscieisenhamer (commented):
Have some questions about putting `ModelContainer` into `associations`:

Primary is a use-case question: Do we know that `jdaviz` and other DATB code do not use `ModelContainer`? Or in general, if the goal was to import `jwst.datamodels` as a separate package, and that `ModelContainer` was part of `jwst.datamodels`, is that OK with the API-breaking?

Other is, if the above is not an issue, whether `ModelContainer` should be in associations, or the more general `lib` module. `ModelContainers`, besides being able to take associations as input, have nothing to do with associations. It is more of a higher-level DataModel management class.

WilliamJamieson:
> Have some questions about putting `ModelContainer` into `associations`:
> 
> Primary is a use-case question: Do we know that `jdaviz` and other DATB code do not use `ModelContainer`?

`ModelContainer` is pretty specific to data-handling within `jwst` (and really ought to be a private class to `jwst` if we can make it so). This being said, `jdaviz` makes no reference to it in their codebase.
 
> Or in general, if the goal was to import `jwst.datamodels` as a separate package, and that `ModelContainer` was part of `jwst.datamodels`, is that OK with the API-breaking?

So right now the API doesn't break at all because the `jwst.datamodels` sub-package still "exists" as a thin import from `stdatamodels.jwst` and then inserts the `ModelContainer` and `SourceModelContainer` imports from `jwst.assocations` (for now). I depreciated that module so a warning tests issued on import from it as the API could technically diverge between the two. I can add a test which imports from `stdatamodels.jwst.datamodels` and checks that everything imported in that is imported in `jwst.datamodels` and drop the warning. Then there is not API change.

> Other is, if the above is not an issue, whether `ModelContainer` should be in associations, or the more general `lib` module. `ModelContainers`, besides being able to take associations as input, have nothing to do with associations. It is more of a higher-level DataModel management class.

I'm fine with it going wherever, I stuck it in `jwst.associations` because thats what broke its ability to move out of `jwst`.

stscieisenhamer:
About where `ModelContainer` lives, I would suggest moving out to `lib`.

About how to get `ModelContainer`: If I understand right, if one just imports `stdatamodels.jwst.datamodels`, one will not get `ModelContainer` because `ModelContainer` requires `jwst`? That is the breaking API that was my concern. If DATB code doesn't care, then it is not a big issue, I'm guessing. I presume it will be documented that `stdatamodels.jwst.datamodels` is a complete replacement for `jwst.datamodels` except for `ModelContainer`?

WilliamJamieson:
> I presume it will be documented that `stdatamodels.jwst.datamodels` is a complete replacement for `jwst.datamodels` except for `ModelContainer`?

Correct, though if they are willing to use jwst then the jwst.datamodels API can be made to be the same (and evolve with stdatamodels). 

My understanding is limited to jdaviz where they want to be able to open the files jwst produces without needing all of jwst for the file to successfully open (an implied dependence on the datamodels rather than an explicit one). Since ModelContainer is not written to any of those files it isn't necessary for this purpose.

nden (approved):
LGTM

docs/getting_started/quickstart.rst
hbushouse: Those are the datamodels for use in Florida.

docs/jwst/datamodels/index.rst
```diff
@@ -1,16 +0,0 @@
-.. _data-models:
```
hbushouse, line 1: Would it be possible to retain at least some single, simple rst file in this location within jwst that just points/links over to the top-level datamodels docs that will now be over in `stdatamodels`?  I don't like the idea of folks having to manually go looking for them in some other package, because they are an integral part of the jwst pipeline.

hbushouse (commented):
Is there any place where the use of the old `from jwst import datamodels` will generate either a warning or deprecation notice, or will it just work transparently and silently?

docs/jwst/datamodels/index.rst
WilliamJamieson: Yes, but this will require a release of `stdatamodels` with the datamodels present so that I can build intersphinx links.

hbushouse:
> > Is there any place where the use of the old `from jwst import datamodels` will generate either a warning or deprecation notice, or will it just work transparently and silently?
> 
> In terms of importing from the `jwst.datamodels` subpackage (and not a submodule), no there will not be a deprecation notice and should continue to work so long as the tests in: `jwst.datamodels.tests.test_api` all pass w.r.t the version of `stdatamodels` being used. That set of tests ensures that the top-level API remains synchronized.
> 
> But, imports from modules within the `jwst.datamodels` package which previously worked will break, i.e. imports of the form (`dqflags` can be any module in `jwst.datamodels`):
> 
> ```python
> from jwst.datamodels import dqflags
> from jwst.datamodels.dqflags import pixel
> ```
> 
> will be broken.
> 
> I could create thin modules for all the sub-modules to ensure these don't break either, and figure out some API tests to ensure nothing is missed. Though I would be limited to what is present in `__all__` statements, which does not cover every possibility of what could be imported.

The `dqflags` stuff is certainly used a lot in the actual pipeline code, but I'm not sure how often it's used in other stuff, like scripts/notebooks that folks in INS use for either data analysis or reference file creation. I suppose we could try to find out from them how pervasive it is in their existing stuff and hence how much work it would be for them to update them all.

WilliamJamieson:
> The `dqflags` stuff is certainly used a lot in the actual pipeline code, but I'm not sure how often it's used in other stuff, like scripts/notebooks that folks in INS use for either data analysis or reference file creation. I suppose we could try to find out from them how pervasive it is in their existing stuff and hence how much work it would be for them to update them all.

As far as the `jwst` package is concerned, all the imports of `jwst.datamodels` have been updated to `stedatamodels.jwst.datamodels` (except of course the `ModelContainer` and `SourceModelContainer`) imports. I have no idea for outside packages.

I guess it is safest to add thin submodules which at least replicate the `__all__` API between `jwst.datamodels` and `stdatamodels.jwst.datamodels`. If someone is importing something outside the `__all__` API on a module, then they are playing dangerous games, because the `__all__` API defines the "publicly" available imports from that file and everything else contained in that module are "private". Breaking the "private" API is an entirely reasonable thing to do.

jemorrison:
The MIRI team uses the dqflags and jwst.datamodels frequently in Jupiter notebooks - both to make reference files and to do analysis.

WilliamJamieson:
> The MIRI team uses the dqflags and jwst.datamodels frequently in Jupiter notebooks - both to make reference files and to do analysis.

I will sync the API so that the public API does not get broken. It will basically be impossible to not break the private API.

hbushouse (approved):

PR #7443, opened 2023-01-30
(JP-3029) Update to calculate separate RN variance

dmggh:
Resolves [JP-3029](https://jira.stsci.edu/browse/JP-3029)

Closes #

This PR calculates a readnoise variance that can be used by the resampling step, using UNDERSAMP-groups that have been flagged by the undersampling_correction step. This functionality requires that PR# 7396 for JP-3015 be merged

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/ramp_fitting/ramp_fit_step.py
+++ b/jwst/ramp_fitting/ramp_fit_step.py
@@ -213,7 +213,6 @@ class RampFitStep(Step):
             max_cores = self.maximum_cores
             readnoise_filename = self.get_reference_file(input_model, 'readnoise')
             gain_filename = self.get_reference_file(input_model, 'gain')
-
             log.info('Using READNOISE reference file: %s', readnoise_filename)
             log.info('Using GAIN reference file: %s', gain_filename)
 
@@ -241,6 +240,11 @@ class RampFitStep(Step):
 
             int_times = input_model.int_times
 
+            # Before the first ramp_fit() call, copy the input model
+            # ("_W" for weighting) for input to the 2nd ramp_fitting call.
+            input_model_W = copy.copy(input_model)
+
+            # Run ramp_fit() the first time, ignoring all DO_NOT_USE groups
             image_info, integ_info, opt_info, gls_opt_model = ramp_fit.ramp_fit(
                 input_model, buffsize,
                 self.save_opt, readnoise_2d, gain_2d, self.algorithm,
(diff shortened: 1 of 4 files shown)


dmggh:
This PR needs additional testing, unit tests, and updated documentation and change log.

jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -242,11 +246,43 @@ def process(self, input):
...
-            self.save_model(opt_model, 'fitopt', output_file=self.opt_name)
+            # To calculate the readnoise variance for weighting, ramp_fitting
+            # will be run a second time - this time with the DO_NOT_USE flags
+            # removed from the UNDERSAMP-flagged pixels. To reduce execution time,
+            # fitting will effectively be performed only on the UNDERSAMP-flagged
+            # pixels, by flagging all other pixels as DO_NOT_USE.
+
+            # First create an intermediate model ('_W' for weighting) to
+            # input to a 2nd ramp_fitting call.
+            input_model_W = copy.copy(input_model)
```
kmacdonald-stsci, line 257: The input model gets modified during `ramp_fit`, such as NaNs getting set in the SCI array, as well as DO_NOT_USE and SATURATION flags possibly getting set in the group DQ array.  Should this copy occur before the first call to `ramp_fit`?

jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -242,11 +246,43 @@ def process(self, input):
...
+            image_info_W, integ_info_W, opt_info_W, gls_opt_model_W = ramp_fit.ramp_fit(
+                input_model_W, buffsize,
+                self.save_opt, readnoise_2d, gain_2d, self.algorithm,
+                self.weighting, max_cores, dqflags.pixel, suppress_one_group=self.suppress_one_group)
 
+            # The VAR_RNOISE arrays are written out to the out_model via the image_info tuple,
+            # and similarly for the integration and optional models. Create new info tuples
+            # to be composed of the original values plus the weighted RN variance value. The
+            # 3rd component of each contains the RN variance.
+            image_info_new = (image_info[0], image_info[1], image_info[2], image_info_W[3], image_info[4])
```
kmacdonald-stsci, line 277: This will fail if `image_info` is NoneType.  Checking that non-NoneType values are returned from `ramp_fit` should occur before trying to access elements of the returns.

jwst/ramp_fitting/ramp_fit_step.py
```diff
@@ -242,11 +246,43 @@ def process(self, input):
...
+            # pixels, by flagging all other pixels as DO_NOT_USE.
+
+            # First create an intermediate model ('_W' for weighting) to
+            # input to a 2nd ramp_fitting call.
+            input_model_W = copy.copy(input_model)
+
+            # Flag all pixels as DO_NOT_USE, then remove the DO_NOT_USE flag
+            # from the UNDERSAMP-flagged pixels.
+            gdq = input_model_W.groupdq.copy()
+            wh_undersamp = np.where(np.bitwise_and(gdq.astype(np.int32), dqflags.group['UNDERSAMP']))
```
kmacdonald-stsci, line 262: A check should be done to make sure a group has this flag set.  All this extra processing should be done only if the length of `wh_undersamp[0]` is greater than 0.  Otherwise, this is extra processing that doesn't add anything.

kmacdonald-stsci (changes requested):

jwst/ramp_fitting/ramp_fit_step.py
dmggh: Good point. Made that change.

jwst/ramp_fitting/ramp_fit_step.py
dmggh: Updated that and related code.

jwst/ramp_fitting/ramp_fit_step.py
dmggh: Yes. Adding that check first.

dmggh:
To assess the effect of my code changes on ramp fitting performance, I am testing the code on larger datasets that have been run through the undersampling correction step.
 
In my first test I just wanted to get a baseline estimate of the execution times and memory usage using the latest version of my code.  I have run the undersampling correction step on 2 versions of Kevin Volk's sample file that has 3 integrations, 50 reads per integration and is 1Kx1K to create an input file for ramp fitting. Both runs were with the save optional argument set to true ( not the default ), and not using command line overrides for the 2 reference files.

Running on my macbook pro laptop ( 32 GB, 8 cores, 2.3 GHz ) I ran the ramp fitting step on 2 datasets :
a ) Dataset on which the undersampling correction step has been run :
The total execution time of the ramp fitting step was approximately 20 minutes, with ~10 minutes spent in each of the
2 calls of the ramp_fit() routine that are called within the step. Memory usage reached a maximum of 14 GB.

b ) A version of the same dataset that has not been run through the undersampling correction step:
Because this dataset has no groups flagged as undersample-corrected, my updated ramp fitting code skips the second call  of the ramp_fit() routine.  Memory usage reached a maximum of 9GB. As expected, the total execution time is roughly 10 minutes - half the execution time of 'a)'.

I will create other datasets whose size may be more representative of the larger datasets we will encounter.
 
I am looking into code optimizations to reduce the memory in the case ramp_fit() is called twice. Although I had earlier looked into ways to calculate the new variety of readnoise variance without rerunning the entire ramp_fit routine a second time (as the code now does) with the modified  GROUPDQ, but I will look into this further.

dmggh:
This update bypasses the second fitting call, by calling a new routine to calculate the readnoise variances. Needs additional testing, so should remain a draft for now.

dmggh:
This PR has been superseded by #7484, so am closing this PR (#7443)

PR #7446, opened 2023-01-30
JP-3024: Allow MIRI LRS fixed slit rateints data through calwebb_spec2

tapastro:
Resolves [JP-3024](https://jira.stsci.edu/browse/JP-3024)

Closes #

This PR fixes an issue preventing by-integration data for MIRI LRS fixed slit exposures from processing through calwebb_spec2 - the pathloss step presumed that the pathloss correction array and the data array were both two dimensional, which is only true when using rate files as input. This PR repeats the pathloss correction across each integration, and it was found to be the only fix necessary to allow for processing to complete.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/pathloss/pathloss.py
+++ b/jwst/pathloss/pathloss.py
@@ -980,8 +980,19 @@ def _corrections_for_lrs(data, pathloss):
 
         # Save the corrections. The `data` portion is the correction used.
         # The individual ones will be saved in the respective attributes.
-        correction = type(data)(data=pathloss_2d)
-        correction.pathloss_point = pathloss_2d
+        if isinstance(data, datamodels.ImageModel):
+            correction = type(data)(data=pathloss_2d)
+            correction.pathloss_point = pathloss_2d
+        elif isinstance(data, datamodels.CubeModel):
+            pathloss_3d = np.repeat(pathloss_2d[np.newaxis, :, :], len(data.data), axis=0)
+            correction = type(data)(data=pathloss_3d)
+            correction.pathloss_point = pathloss_3d
+        else:
+            errmsg = (f"Pathloss step received unexpected datamodel type {type(data)}, "
+                      f"which it cannot process. Exiting.")
+            log.critical(errmsg)
+            raise RuntimeError(errmsg)
+
         correction.wavelength = wavelength_array
 
     else:


tapastro:
Should consider adding test coverage? According to JIRA comments - "TSO mode with the slit is not actually approved yet for implementation", so may be jumping the gun, or might just be a good idea.

tapastro:
Started a regression test run: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/556/

hbushouse (commented):
This looks fine, but I'm surprised the existing code doesn't just automatically handle the application of a 2D correction array to 3D science data array by array broadcasting. For example, the flatfield step does this seamlessly at https://github.com/spacetelescope/jwst/blob/master/jwst/flatfield/flat_field.py#L209
where it's even noted in the comments that if the input `science.data` is 3D, the 2D `flat_data` array will be broadcast across the 3rd dimension. So why doesn't the same thing automatically happen at https://github.com/spacetelescope/jwst/blob/master/jwst/pathloss/pathloss.py#L571 (and following lines)?

hbushouse:
> One option could be to simply leave it as an array and let it broadcast. The existing code fails because it attempts to create a new datamodel of the input model's type, then assign the 2d correction to it as the data array. This fails when the new instance is a CubeModel and is given 2d data.
> 
> It wasn't clear to me if the existing approach had an unforeseen (to me) benefit, so I kept the model around...

Ah, I see. If the correction arrays are always 2D, then would it be possible to always create a new data model for the correction that is of type `ImageModel`, rather than using the same type as the input data? Broadcasting should then work when trying to do something like `science_model_3d.data *= correction_model_2d.data`.

tapastro:
I think I gave too much credit to the existing structure - I figured there might be a reason for the correction to match the input model type, otherwise why code it this way? But it does run (and reads much more clearly) by explicitly calling an ImageModel.

hbushouse:
> I think I gave too much credit to the existing structure - I figured there might be a reason for the correction to match the input model type, otherwise why code it this way? But it does run (and reads much more clearly) by explicitly calling an ImageModel.

Never assume that the original code was the result of divine inspiration. ;-)

hbushouse (approved):
Wow, so that's a waaaay simpler implementation than the first attempt! A one-liner!

hbushouse:
Is there a recent regtest run for the latest code?

hbushouse:
> Yes, I forgot to link it here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/557/
> 
> It had a couple of strange (unrelated?) errors.

Yes, those same 2 failures have been popping up randomly over the past month or so. Ignore them.

hbushouse:
I'm guessing we don't have any regtests using rateints/calints for MIRI LRS slit, because it's not a supported mode (yet). I guess we can wait till it is supported and we have real data.

tapastro:
Yeah, I mentioned in my first comment here that because it technically isn't an approved implementation, it could probably wait. I'll add a post-it to the monitor 😉

PR #7447, opened 2023-01-30
JP-2862: Push DO_NOT_USE flags to groupdq array from mask

tapastro:
Resolves [JP-2862](https://jira.stsci.edu/browse/JP-2862)

Closes #

This PR sets the DO_NOT_USE bits of the groupdq array to match the mask used in dq initialization - previously, they were only used to set the bits of the pixeldq array. 

NOTE: While I see the changes in the results of the dq_init step, the final rate/rateints files I used for testing (a MIRI fixed slit exposure) do not show much measurable difference in the rates, and no discernable difference in the final dq array. Querying ticket opener on a more rigorous test dataset.

**Checklist for maintainers**
- [x] added relevant label(s)


stscieisenhamer (commented):
initially looks good. re-request review when pr is ready.

hbushouse (commented):
Code looks good (nice and simple). Needs a change log entry yet.

tapastro:
Regression test run started here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/574/

Also cleaned up the changelog - some duplicate entries and lack of order.

tapastro:
> Regression test run started here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/574/
> 
> Also cleaned up the changelog - some duplicate entries and lack of order.

Whew, there are a lot of failures - likely expected, as any changes to the dq array at the dq_init stage will cause a lot of downstream effects.

hbushouse:
Your regtest run also had the misfortune of picking up unrelated changes due to 1) updated NIRCam saturation ref file in CRDS, and 2) changes to ramp_fitting that recently got merged. All truth files related to those changes have now been updated, so you might want to try your regtest run again. But yes, there will still be lots of downstream diffs due to the fact that dq_init is the first step in the *entire* pipeline!

hbushouse:
The CI test failures are also due to unrelated changes in ramp_fitting, which have now been resolved. So another iteration of those tests is also in order.

hbushouse (commented):
Code updates are simple enough and appear to be having the desired effect, based on differences showing up in the regtest results. Just need to (again) update the change log (it's been a mess recently, due to so many PR's all making different changes to it) and run another round of all the tests.

tapastro:
Regtest round 2: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/576/

hbushouse:
Latest regtest (round 2) results look OK to me. I'm just curious about the presence of pixels that now have values of zero in the variance and err arrays, which I guess are the pixels that are now being completely flagged out by the additional propagation of DO_NOT_USE values to ramp fitting. What I'm curious about is the fact that they are set to zero, as opposed to something like NaN. @kmacdonald-stsci can you verify whether we expect *variance* values to be zero for pixels with no good data?

hbushouse:
> > Latest regtest (round 2) results look OK to me. I'm just curious about the presence of pixels that now have values of zero in the variance and err arrays, which I guess are the pixels that are now being completely flagged out by the additional propagation of DO_NOT_USE values to ramp fitting. What I'm curious about is the fact that they are set to zero, as opposed to something like NaN. @kmacdonald-stsci can you verify whether we expect _variance_ values to be zero for pixels with no good data?
> 
> Yes. Bad variances are set to zero. During computation bad variance values are set to `LARGE_VARIANCE` defined in utils.py. This allows the calculations to work without worrying about a divide by zero or affecting the final results for values that do have good values. At the end variances are checked to see if they are above `LARGE_VARIANCE_THRESHOLD` and if so set to zero.

OK, thanks for confirmation. In that case, I think these results all look good.

hbushouse (approved):

PR #7451, opened 2023-02-03
JP-3083: Fix photom handling of NRS FS source types

hbushouse:
Resolves [JP-3083](https://jira.stsci.edu/browse/JP-3083)

This PR fixes the logic in the photom step that handles NIRSpec fixed-slit data, so that it uses the correct source type for each slit, which was assigned up-stream in the srctype step. This is important, because point sources are left in units of MJy, while extended sources are converted to surface brightness (MJy/sr).

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -853,13 +853,9 @@ class DataSet():
                                                      dqflags.pixel['DO_NOT_USE'])
             if not self.inverse:
                 if unit_is_surface_brightness:
-                    log.debug(' unit_is_surf_bright is True')
-                    log.debug(' in photom_io MultiSlit; set to MJy/sr')
                     slit.meta.bunit_data = 'MJy/sr'
                     slit.meta.bunit_err = 'MJy/sr'
                 else:
-                    log.debug(' unit_is_surf_bright is False')
-                    log.debug(' in photom_io MultiSlit; set to MJy')
                     slit.meta.bunit_data = 'MJy'
                     slit.meta.bunit_err = 'MJy'
             else:
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -134,9 +134,26 @@ class DataSet():
         self.slitnum = -1
         self.specnum = -1
         self.inverse = inverse
-        self.source_type = source_type
-        if self.source_type is None and model.meta.target.source_type is not None:
-            self.source_type = model.meta.target.source_type.upper()
+        self.source_type = None
+
+        # For MultiSlitModels, only set a generic source_type value for the
+        # entire datamodel if the user has set the source_type parameter.
+        # Otherwise leave the generic source_type set to None, which will
+        # force the use of per-slit source_types later in processing.
+        if isinstance(model, datamodels.MultiSlitModel):
+            if source_type is not None:
+                self.source_type = source_type
+
+        # For non-MultiSlitModel inputs, where there's only 1 target and
+        # one source_type, use the user-provided source_type value, if it
+        # exists. Otherwise, use the generic source_type value provided in
+        # the input model (if it exists).
+        else:
+            if source_type is not None:
+                self.source_type = source_type
+            else:
+                if model.meta.target.source_type is not None:
+                    self.source_type = model.meta.target.source_type.upper()
 
         # Create a copy of the input model
         self.input = model.copy()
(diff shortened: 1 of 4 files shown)


hbushouse:
This version of the code succeeds at applying the proper flux calibration to the data in each slit and setting the BUNIT keywords in the output to the proper values for each slit. However, something in the downstream resample_spec step is mucking with the meta data and causing the final cal product to still have BUNIT values that are all DN/s. I'm guessing this is happening when the input model goes through the `exp_to_source` conversion at the beginning of resample_spec. Will likely need @stscieisenhamer 's expertise to figure out the problem in there.

hbushouse:
Regtest run at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/563/

hbushouse:
Regtest results show many expected differences, but checking each one carefully to make sure they really are correct.

hbushouse:
Check all differences that showed up in the regtest run and have confirmed that most all of them are expected and desired, based on the fixes to the photom step and also the change to try to make the resample_spec step stop modifying its input, leaving the final cal files with incorrect meta values. A few of the changes are still not correct, in that the exp_to_source process is incorrectly copying the old BUNIT keyword values of "DN/s" to the output resampled (s2d) files, instead of propagating the correct slit-dependent values of either "MJy" or "MJy/sr", depending on whether each slit contains a POINT or EXTENDED source. The proper BUNIT values are being set by the photom step and are now staying that way in the final cal files, but are incorrect in the s2d files. @stscieisenhamer is working on a fix for that in https://jira.stsci.edu/browse/JP-3088

jwst/photom/photom.py
```diff
@@ -134,9 +134,26 @@ def __init__(self, model, inverse=False, source_type=None, correction_pars=None)
...
+
+        # For non-MultiSlitModel inputs, where there's only 1 target and
+        # one source_type, use the user-provided source_type value, if it
+        # exists. Otherwise, use the generic source_type value provided in
+        # the input model (if it exists).
+        else:
+            if source_type is not None:
+                self.source_type = source_type
+            else:
+                if model.meta.target.source_type is not None:
```
tapastro, line 155: Is there a situation where `model.meta.target.source_type` could (should) be None? And if it is, will self.source_type be None as well because it hasn't been set up to this point?

jwst/photom/photom.py
```diff
@@ -691,7 +708,7 @@ def photom_io(self, tabdata, order=None):
             conversion = tabdata['photmj']              # unit is MJy
             if isinstance(self.input, datamodels.MultiSlitModel):
                 slit = self.input.slits[self.slitnum]
-                if self.exptype == 'NRS_MSASPEC':
+                if self.exptype in ['NRS_MSASPEC', 'NRS_FIXEDSLIT']:
```
tapastro, line 711: Is there an exptype where this conditional is intentionally excluding the following line?

tapastro (approved):
LGTM, with a couple questions on some conditionals.

jwst/photom/photom.py
hbushouse: I believe that's a holdover from the days when we had to be able process groundbased test data, which was static and did not evolve to keep up with pipeline development (as we added new keywords, etc.), and hence some of those old datasets may not have `source_type` set. In those cases, suitable default values for each slit are set during the `srctype` step, so the individual slit values *should* always have a value.

jwst/photom/photom.py
hbushouse: It's complicated logic that was imposed way back to handle the MOS mode master background subtraction processing in calspec2, where you occasionally have to force the step to use a given source_type (supplied as an argument to the step), rather than having it choose based on the existing `slit.source_type`. `self.source_type` is the value passed in as an argument and acts as an override. We need to use the `slit.source_type` for FS mode and this is the only if-block where that gets used, so I had to add it along with MOS. Other multiple-slit modes, like WFSS, don't need that logic (allowing for the possibility of an override). It's related to all the complicated logic to set `self.source_type` starting back up at https://github.com/spacetelescope/jwst/pull/7451/files#diff-b9378ce0b10971befd7443bb2b038a52954fb85c6a06906b7d0d6db5cb6a4057R137  At least that's my story and I'm sticking with it.

PR #7464, opened 2023-02-13
JP-3084: Separate photometric calibration factors for point vs. uniform use

tapastro:
Resolves [JP-3084](https://jira.stsci.edu/browse/JP-3084)

Closes #

This PR addresses an inconsistency in how the scalar conversion factor in the `photom` step is passed through various point vs. extended source conditionals, leading to a 2D conversion array stored in `model.slits[i].photom_uniform` not containing a conversion to MJy/sr. 

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -826,8 +826,6 @@ class DataSet():
                                                                    self.exptype, conversion,
                                                                    waves, relresps, order)
 
-                    slit.photom_uniform = conversion
-
             elif isinstance(self.input, datamodels.MultiSpecModel):
 
                 # This input does not require a 2d conversion, but a 1d interpolation on the
--- a/jwst/photom/photom.py
+++ b/jwst/photom/photom.py
@@ -721,6 +721,7 @@ class DataSet():
                                   "to surface brightness")
                         conversion /= slit.meta.photometry.pixelarea_steradians
                 else:
+                    conversion_uniform = conversion / slit.meta.photometry.pixelarea_steradians
                     unit_is_surface_brightness = False
             elif isinstance(self.input, datamodels.MultiSpecModel):
                 # MultiSpecModel output from extract1d should not require this area conversion?
@@ -735,6 +736,7 @@ class DataSet():
                                   "to surface brightness")
                         conversion /= self.input.meta.photometry.pixelarea_steradians
                 else:
+                    conversion_uniform = conversion / slit.meta.photometry.pixelarea_steradians
                     unit_is_surface_brightness = False
 
         # Store the conversion factor in the meta data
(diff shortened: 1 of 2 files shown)


tapastro:
Regression tests: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/575/

hbushouse (commented):
I think the new logic here all looks correct. Of course the PHOTOM_UN array (at least for NIRSpec) still won't have values that will actually convert to surface brightness; they'll just be 1/area, right? I guess we'll have to see if that's good enough to get the master background stuff to work correctly.

tapastro:
After fixing my naming mistake which got the CI tests running, I'm re-running the regression test here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/577/

hbushouse:
Many of the regression test failures are due to cal files now containing more HDU's than before. Is that expected? Were some extensions, like PHOTOM_UN or PHOTOM_PS, not being included before the fix? Given the difference in HDU's, it's hard to make sense of the comparisons.

hbushouse (approved):
I think the code all looks correct.

tapastro:
The changes (apart from one which shows the fix including 1/pixelarea_steradians) are outputs that now include the PHOTOM_UN extension which previously did not. I don't know if that is in the "spec" for these outputs, but it seemed like the existing was inconsistent in its decision to save (previously, PHOTOM_UN was saved only for _point-like_ sources), so I think all results are expected and positive changes.

hbushouse:
Doing some archival research into the original implementation of computing and saving both point-source and uniform-source versions of the photometric calibration for NIRSpec (as documented in [JP-1776](https://jira.stsci.edu/browse/JP-1776) and related/child tickets), the only time we need to compute and store both versions is when the slit being processed contains a point source. In that case, both the photom_ps and photom_un arrays are needed later in the master background subtraction step, in order to correct the background data (which are calibrated as uniform source) to match that of a point source. The photom_ps array is not needed for slits that contain an extended/uniform source.

jwst/photom/photom.py
```diff
@@ -795,35 +797,37 @@ def photom_io(self, tabdata, order=None):
...
                                                                    use_wavecorr=True)
                     slit.photom_point = conversion  # store the result
 
                 else:
-                    conversion, no_cal = self.create_2d_conversion(self.input.slits[self.slitnum],
+                    conversion, no_cal = self.create_2d_conversion(slit,
                                                                    self.exptype, conversion,
                                                                    waves, relresps, order)
+
+                    slit.photom_uniform = conversion
```
hbushouse, line 829: Now that I've done the historical research, I don't think this addition is needed. It's in the branch of the if/else that applies if the slit does NOT contain a point source, in which case we don't need to store either `photom_point` or `photom_uniform`. Just apply the computed uniform-source correction to the data, without storing it in the output.

hbushouse (approved):
I (re)approve.

PR #7466, opened 2023-02-15
JP-3095: Fix extract_1d usage of use_source_posn param

hbushouse:
Resolves [JP-3095](https://jira.stsci.edu/browse/JP-3095)

This PR updates the logic in the lower-levels of the extract_1d code to properly handle the various ways the "use_source_posn" can be specified, including command line overrides and from within the extract1d reference file. The previous logic was automatically turning the param on for certain exposure types, such as MIRI MRS and LRS fixed-slit, and NIRSpec IFU and slits, too high up in the code, before the information from the extract1d ref file had been loaded, thus overriding any settings in the ref file (if present). These changes move the logic down lower, so that the automatic override is only applied after the ref file params have been loaded and checked. The new logic how honors the appropriate hierarchy of precedence, in that it gives highest precedence to user-supplied values on the command line, then the ref file value (if any), and finally the automatic setting for some exposure types.

This fixes an issue encountered when the MIRI team decided they wanted to turn off the use of "use_source_posn" for LRS fixed-slit data by setting the param in the extract1d ref file and it was being ignored.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,13 @@ datamodels
 
 - Move ``jwst.datamodels`` out of ``jwst`` into ``stdatamodels.jwst.datamodels``. [#7439]
 
+extract_1d
+----------
+
+- Fix the logic for handling the ``use_source_posn`` parameter, so that proper
+  precedence is given for command line override, reference file settings, and
+  internal decisions of the appropriate setting (in that order). [#7466]
+
 other
 -----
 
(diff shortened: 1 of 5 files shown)


hbushouse:
Regression test running at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/578/

tapastro (approved):
LGTM - much simpler and easier to read. The one regtest failure is a bit strange - but as it matches the exposure type that caused this ticket, perhaps it's now functioning as intended, therefore a change is expected?

hbushouse:
> LGTM - much simpler and easier to read. The one regtest failure is a bit strange - but as it matches the exposure type that caused this ticket, perhaps it's now functioning as intended, therefore a change is expected?

Yes, the exposure type is one that had problems (LRS slit), but the test itself is to check the application of an image-based reference file, rather than the typical json extract1d ref file (the image-based version just uses a bit mask of which pixels to include in the extraction), so I'm not sure why those results would change. It appears it only resulted in slight differences in the wavelength values for the extracted 1D spectrum. Will need to investigate.

PR #7471, opened 2023-02-27
JP-3106: Fix input NaNs in residual fringe

drlaw1558:
Fixes an issue with MRS Residual Fringe crashing on input NaN values (https://jira.stsci.edu/browse/JP-3106) caused by recent merges of https://jira.stsci.edu/browse/JP-3014 and https://jira.stsci.edu/browse/JP-3087

Convert NaN values to zero at the start of this routine, then repopulate as NaN at the end since the Bayesic routines don't handle either NaN or zero well, but there is already infrastructure in place to deal with input zeros.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -64,6 +64,11 @@ resample
   a bug due to which parts of input images may not be present in the output
   resampled image under certain circumstances. [#7460]
 
+residual_fringe
+---------------
+
+- Fix bug in Residual Fringe step handling of NaNs in input []
+
 scripts
 -------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -67,7 +67,7 @@ resample
 residual_fringe
 ---------------
 
-- Fix bug in Residual Fringe step handling of NaNs in input []
+- Fix bug in Residual Fringe step handling of NaNs in input [#7471]
 
 scripts
 -------
--- a/jwst/residual_fringe/residual_fringe.py
+++ b/jwst/residual_fringe/residual_fringe.py
@@ -87,7 +87,7 @@ class ResidualFringeCorrection():
         # Set them to 0 for the residual fringe routine
         # They will be re-added at the end
         output_data = self.model.data.copy()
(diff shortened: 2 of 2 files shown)


CHANGES.rst
```diff
@@ -64,6 +64,11 @@ resample
   a bug due to which parts of input images may not be present in the output
   resampled image under certain circumstances. [#7460]
 
+residual_fringe
+---------------
+
+- Fix bug in Residual Fringe step handling of NaNs in input []
```
hbushouse, line 70: Need to add the PR number here (7471)

hbushouse (approved):
Looks good to me. Just one minor update needed for the change log entry.

hbushouse:
Oh, there's also a CI code style error that needs fixing:
jwst/residual_fringe/residual_fringe.py:90:60: E712 [*] Comparison to `False` should be `cond is False`

hbushouse:
Started regression test run at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/584/

drlaw1558:
> Oh, there's also a CI code style error that needs fixing: jwst/residual_fringe/residual_fringe.py:90:60: E712 [*] Comparison to `False` should be `cond is False`

Not sure I follow that issue; the offending line seems to be 
nanval_indx = np.where(np.isfinite(output_data) == False)
but using
nanval_indx = np.where(np.isfinite(output_data) is False)
just causes it to return no pixels at all.

I could probably do
nanval_indx = np.where(~np.isfinite(output_data))
but I'm not sure what the CI stylebot wants.

hbushouse:
What about inverting the logic by just using `np.where(np.isnan(output_data))` such that you don't need a comparison at all?

jemorrison (approved):
Looks good.

drlaw1558:
The problem with using np.isnan is that it wouldn't pick up on any np.inf values that may happen to be in the data.  I don't remember seeing these usually, but given that the following code is fragile against non-finite values then testing for np.isfinite seems more robust.

PR #7473, opened 2023-03-01
Control whether to use absolute or relative values in skymatch overlaps

### LINKED ISSUE #7472: skymatch: "match" should minimize difference in overlaps, not in whole images
It appears that in the skymatch "match" step, the minimization occurs between the entire images, not the overlap images. At least, I think that's what happens in the 
`A, W = _overlap_matrix(images, apply_sky=apply_sky)`
step in `skymatch.py`. This is fine for mostly blank images where you can calculate a "true" background, but for images heavily full of emission that isn't symmetric about the overlap points, this can cause really noticeable steps between tiles in the final mosaic. We've found for the PHANGS work that actually turning off the skymatch can help, but still isn't ideal. We have found that mosaicking each field separately and then stitching together using e.g. Montage has produced better results that the pipeline, but obviously this isn't a great solution.

Likely, what should happen is that calc_sky should be called for each part of the tile that overlaps, and then those values should be minimized. I don't think this would be a huge amount of rewriting, just pulling out the overlap regions rather than relying on the statistics for the whole tile. I think this is the intent, reading the docs: "For each pair of images, the sky mismatch is computed only in the regions in which the two images overlap on the sky." but this does not appear to be the case looking into the code

thomaswilliamsastro:
Closes #7472 (I misunderstood what the code was doing, but this PR fixes the issue we have)

This PR adds a new parameter to the skymatch step, `apply_sky`, to decouple the `subtract` parameter from absolute/relative comparisons in the overlap regions of `match`. This is vital for observations where emission fills the tiles, and is not consistent across them. I've used `apply_sky` to be consistent with the parameter name passed to the `_find_optimum_sky_deltas` call in `skymatch.py`.

For backwards compatibility, this defaults to None, which will then set apply_sky = not subtract, the old behaviour.

To illustrate, here's the F770W image for NGC1087, all with 'global+match'. The difference is the left has subtract=True, middle has subtract=False, and right has subtract=True and apply_sky=True. These are all locked to the same scale. The steps between tiles are super apparent in the left and middle, but are much reduced in the right.

Screenshot 2023-03-01 at 11 24 54

**Checklist for maintainers**
- [ ] added relevant label(s)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ htmlcov
 requirements-min.txt
 .tox/
 .tmp/
+.idea/
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -166,7 +166,8 @@ skymatch
 --------
 
 - Added an ``apply_sky`` parameter, to switch between using relative or absolute
-  sky values in match overlaps. [#7473]
+  sky values in match overlaps. Also changes how bounding box is calculated for
+  SkyImage, to avoid large empty areas in MIRI observations [#7473]
 
 straylight
 ----------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -179,6 +179,12 @@ scripts
 
 - Offload ``minimum_deps`` script to ``minimum_dependencies`` package [#7463]
 
+skymatch
+--------
+
+- Added an ``apply_sky`` parameter, to switch between using relative or absolute
+  sky values in match overlaps.
+
 transforms
 ----------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -162,6 +162,12 @@ set_telescope_pointing
 - Fill values of ``TARG_RA`` and ``TARG_DEC`` with ``RA_REF`` and ``DEC_REF``
   if target location is not provided, e.g. for pure parallel observations [#7512]
 
+skymatch
+--------
+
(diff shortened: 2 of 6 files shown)


thomaswilliamsastro:
I've also made an edit to the way the bounding polygons are created, since this was causing some issues. For the MIRI data, it doesn't make sense to simply have a rectangle that covers the whole chip, since much of it is full of non-science coronagraph. It can lead to the overlap fluxes being calculated in regions where science data is present in the image but not the overlap (or vice versa). There are now edits that remove this and instead take a convex hull of the selected usable data. Obviously this still isn't perfect but is much more representative of the overlaps you want to calculate

I'm not sure if this requires any updates to regression tests. It removes the stepsize parameter since this is not needed for the convex hull calculation, which may be an issue. Will make edits if people point me to them!

mcara:
Could you please share the data that are causing the issue? If data are proprietary, contact me offline or via help desk to set-up how to share data.
Thanks

thomaswilliamsastro:
It shows up for a number of the PHANGS-JWST targets. The one I've posted above is the F770W observations of NGC1087, which aren't proprietary:

jw02107001001_02101_0000[1-4]_mirimage
jw02107001002_02101_0000[1-4]_mirimage

hbushouse:
Closing per directive in #7472

PR #7474, opened 2023-03-01
include `opencv-python` in instructions to generate SDP requirements

zacharyburnett:
Resolves [JP-3109](https://jira.stsci.edu/browse/JP-3109)

https://github.com/spacetelescope/stcal/pull/126 and https://github.com/spacetelescope/stcal/pull/136 made `opencv-python` an optional dependency (as built binaries are not available on all user platforms). This PR adds `opencv-python` to the instructions for creating `requirements-sdp.txt`, to ensure the environment has `opencv-python` installed.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/requirements-sdp.txt
+++ b/requirements-sdp.txt
@@ -12,5 +12,5 @@
 #
 #     conda create -n sdp python -y
 #     conda activate sdp
-#     pip install -e .[test]
+#     pip install -e .[test] "opencv-python>=4.6.0.66"
 #     pip freeze | grep -v jwst >> requirements-sdp.txt


zacharyburnett:
Options going forward:
1. restore `opencv-python` to a hard dependency
2. use a less heavy dependency
    - `scikit-image` and `shapely` can perform similar functionality: https://github.com/spacetelescope/stcal/pull/138)
    - > Astropy `regions` can do this easily. You can make ellipses and circles and then export them as pixel masks. And also interface with `shapely` if needed.
    https://github.com/spacetelescope/jwst/issues/7409#issuecomment-1416074160 
4. use binary operations instead of ellipse / circle fitting 
    > Perhaps going too far, a much simpler approach would be to not fit circles and ellipses at all (which can lead to fitting errors and masking half the image), but to use scipy.ndimage.binary_dilation instead. Just grow regions. And select regions of a minimum area to grow using scipy.ndimage.binary_erosion. This is vastly simpler and probably more robust. And handles shapes that are not exactly circles or ellipses.
https://github.com/spacetelescope/jwst/issues/7409#issuecomment-1416074160

hbushouse (approved):

PR #7480, opened 2023-03-06
remove upper pins on spacetelescope projects

zacharyburnett:
This PR addresses issues with testing that arise from upper pins on dependencies.

**Checklist for maintainers**
- [x] added relevant label(s)


nden:
Is this going to create issues to external users the next time e.g. stcal releases incompatible changes? Or is this meant to be temporary while we are testing and developing and then pinning again before the next release?

braingram:
Thanks!

This should unblock the development on stdatamodels by allowing the CI there to test against the development version of jwst and stdatamodels (which has a version above the upper pin that is removed in this PR).

For more context see the stpipe PR (stpipe was checking the jwst version requirements at runtime which made a forced install of incompatible dependencies not possible): https://github.com/spacetelescope/stpipe/pull/84

WilliamJamieson:
> Is this going to create issues to external users the next time e.g. stcal releases incompatible changes? Or is this meant to be temporary while we are testing and developing and then pinning again before the next release?

Adding upper pins to dependencies for libraries causes all sorts of issues downstream and causes all sorts of unintended consequences, see https://iscinumpy.dev/post/bound-version-constraints/ for a better analysis than I can give.

In general, it is up to external users to manage their dependencies not us. In particular, downstream packages to JWST should in fact be testing regularly against the development version of JWST. As for individual users, we should give clear guidance on to how to update their environments. Namely, `pip install --upgrade` does not work as expected, though we can limit some issues via changing the lower pin on say `stcal` as needed.

nden:
I only quickly skimmed that link and I agree with what it says. I'm not sure it fully applies for the jwst package, since it's not exactly a library.
The issues arising from releasing unpinned jwst are listed and were discussed here
https://innerspace.stsci.edu/display/SCSB/Development+Workflow+-+stcal%2C+stpipe%2C+jwst%2C+romancal

We could go over them and discuss them again if you htink it's necessary. OTH, I think it's OK to unpin while in development unless there's something that needs to be avoid in a specific dependency version.

WilliamJamieson:
Given that our release style guilde https://github.com/spacetelescope/style-guides/blob/master/guides/release-workflow.md calls for releases to occur from their own distinct branches, I think it a reasonable compromise for JWST would be to pin dependencies only on these release branches only.

This would limit help calls in the same way the current pins do, while making it much simpler to do development (running CI) across multiple packages, which is something that is often required for JWST. This would fix both issues at the expense of making the release process slightly more complicated (CI checks and automations could limit these issues).

zacharyburnett:
> ... I think it a reasonable compromise for JWST would be to pin dependencies only on these release branches only.

That's a good idea, we could definitely incorporate that into the process (and also consistently release from dedicated branches)

setup.cfg
```diff
@@ -43,9 +43,9 @@ install_requires =
     scikit-image>=0.17.2
     scipy>=1.6.0,<1.10.0
     spherical-geometry>=1.2.22
-    stcal>=1.3.4,<1.4
-    stdatamodels>=1.1.0,<1.2.0
-    stpipe>=0.4.5,<1.0
+    stcal>=1.3.4
+    stdatamodels>=1.1.0
```
hbushouse, line 47: Now that https://github.com/spacetelescope/stdatamodels/pull/127 has been merged, we should push the minimum for stdatamodels up to 1.2.0, so that jwst gets the updates from that PR.

hbushouse (approved):

PR #7483, opened 2023-03-08
JP-3086 reduce memory needed to run cube build

jemorrison:
Resolves [JP-3086](https://jira.stsci.edu/browse/JP-3086) and 
Resolves [JP-3016](https://jira.stsci.edu/browse/JP-3016)

Closes #

This PR addresses the cube_build step requiring large amounts of memory.  The bulk of the changes were done in the c code. Instead of forming and allocating the arrays needed to keep track of the spaxel flux, weight, variance, and dq; the arrays are initialized to the correct size and set to zero in the python code and passed to the c code. This is similar to how the drizzlepac code works. The arrays are populated in the c code and passed back. All creating and deleting of arrays is now done in the python code. The c code does not allocate or de-allocate any memory. This also has a benefit of making the c code simpler. 
In addition, there is an option added to cube_build called 'in_memory'. This concept was taken from the outlier detection code. If this option is set to 'in_memory' = True (default) then all the input data is stored in memory and the output IFU cubes are also stored in memory. If 'in_memory' = False then the input files are written to the local disk and only the filenames are stored. In addition, the IFUcube are written to disk and the IFUCube model is a list of filenames. This option is not a helpful as it seems. I think because of how the pipeline works - still the data models are stored in memory before the step even gets to cube_build (in the setup of reading the association). For now it is an option that people can try if they run in memory issues running locally.

**Checklist for maintainers**
- [x] added relevant label(s)


mcara:
I am not sure I will have time to review this PR in a timely manner. However, reading these comments:

> This PR addresses the cube_build step requiring large amounts of memory. The bulk of the changes were done in the c code. Instead of forming and allocating the arrays needed to keep track of the spaxel flux, weight, variance, and dq; the arrays are initialized to the correct size and set to zero in the python code and passed to the c code. This is similar to how the drizzlepac code works. The arrays are populated in the c code and passed back. All creating and deleting of arrays is now done in the python code. The c code does not allocate or de-allocate any memory. This also has a benefit of making the c code simpler.

makes me wonder how does this address this PR's title "reduce memory needed to run cube build". It actually seems (from comments) that the code was made less efficient. What am I missing.

jemorrison:
Any suggestions on why I am getting the CI dependencies failure ? 
This compiles fine on my Mac.

mcara:
Screen Shot 2023-03-08 at 1 23 08 PM

These are the errors.

jemorrison:
@mcara The code was not freeing the memory in the c code properly. For large associations the memory was growing so large it could not be processed. I could not figure out why the memory was not being freed. I enlisted Warren's help and he pointed me to how drizzle works - - making the arrays in python rather than allocating them in c. The memory is not cleaned up and does not grow to huge sizes (terabytes).

jwst/cube_build/src/cube_match_internal.c
```diff
@@ -220,45 +205,66 @@ PyArrayObject * ensure_array(PyObject *obj, int *is_copy) {
...
-  PyArrayObject *spaxel_flux_arr=NULL, *spaxel_weight_arr=NULL, *spaxel_var_arr=NULL;
-  PyArrayObject *spaxel_iflux_arr=NULL;
-  npy_intp npy_ncube = 0;
-
-  if (!PyArg_ParseTuple(args, "iiiddddOOOOOOOOOOiOO:cube_wrapper_internal",
+  PyArrayObject *a1, *a2, *a3, *a4, *lam1, *lam2;
+  PyArrayObject *lam3, *lam4, *flux, *err, *acoord, *zcoord;
+  PyArrayObject *spaxel_flux, *spaxel_weight, *spaxel_var, *spaxel_iflux;
+  
+  const int max_size_error = 80;
```
mcara, line 244: Now `const int max_size_error = 80;` is not used anywhere. The correct way of doing this in C is to use macros:
```C
#define MAX_ERROR_LENGTH 80
```
somewhere at the top of the file and then inside the function do something like:
```C
char error[MAX_ERROR_LENGTH] = "Hello World";
```

mcara: But why limit error message length at all? Is that important?

mcara:
I do not think this is the best approach to split the code into two parts - Python and C in order to delegate memory management to Python. I think you should investigate where is the memory leak occurring in the C code and fix those bugs.

jemorrison:
@mcara I spent weeks trying to figure out where the memory error was occurring. In the end I decided it was the wrong approach and forming the array in python and having python handle the memory management seemed much cleaner. It is what drizzle does and that has well tested code for memory. At this stage I am think it was the wrong approach to allocate memory in c and pass back the arrays back to python. I could not dellocate the memory in the python code but I did not need those arrays any more.

mcara:
@jemorrison Have you tried this [advice on changing the format from `O` to `N`](https://stackoverflow.com/a/28906538/8033585)?

mcara:
@nden @hbushouse In light of my previous comment, probably all `Py_Build('(OOO)', arr1, ...)` occurrences - see https://github.com/spacetelescope/jwst/search?l=C&q=Py_buildvalue - should be revised to use `N` instead of `O`. See stackoverflow link in my previous comment above and from https://numpy.org/doc/stable/user/c-info.how-to-extend.html:

> The [Py_BuildValue](https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue) (format_string, c_variables…) function makes it easy to build tuples of Python objects from C variables. Pay special attention to the difference between ‘N’ and ‘O’ in the format string or you can easily create memory leaks. The ‘O’ format string increments the reference count of the [PyObject](https://docs.python.org/3/c-api/structures.html#c.PyObject)* C-variable it corresponds to, while the ‘N’ format string steals a reference to the corresponding [PyObject](https://docs.python.org/3/c-api/structures.html#c.PyObject)* C-variable. You should use ‘N’ if you have already created a reference for the object and just want to give that reference to the tuple. You should use ‘O’ if you only have a borrowed reference to an object and need to create one to provide for the tuple.

jemorrison:
@mcara Ok I will open a new branch and change the PY_BuildValue and see if that solves the memory issue.

jemorrison:
@mcara I changed how I am using Py_Build and using N rather than 0. The memory does seem to freed up better now, but in the test I have done  the memory usage is about  as good as this PR. I made number of additional changes to help with memory in this PR. Since OPS really needs this fix I suggest that we merge this PR (especially since I am out of town all next week). I can then follow up and study better if the code is slower when the arrays are created in python rather than allocating the memory in C.  And which method is better for handling memory in extremely large data sets. If there is a good reason why we should allocate the arrays in C then I can  take the changes in this PR, not related to how the arrays are passed around that help other memory issues, and merge then with using Py_Build using NNNN.  Does that sound reasonable ?

mcara (changes requested):
From the software point of view I cannot approve this PR but I cannot judge scientific side of this PR or the importance of getting it in.

@drlaw1558 @hbushouse @nden Please override my review based on your assessment of priorities.

I think that the simplest thing is to replace `O` with `N` (as explained in one of my previous comments) in the old code and then rebase this PR based on the "fixed" version of the old code and implement whatever further improvements this PR has. It seems to me this code may crash under certain circumstances but I cannot evaluate whether the "right" combination of parameters is possible in practice for this to happen.

jwst/cube_build/ifu_cube.py
```diff
@@ -676,6 +700,11 @@ def build_ifucube(self):
                         regions = list(range(start_region, end_region + 1))
 
                         for i in regions:
+                            spaxel_flux = np.zeros(total_num, dtype=np.float64)
+                            spaxel_weight = np.zeros(total_num, dtype=np.float64)
+                            spaxel_var = np.zeros(total_num, dtype=np.float64)
+                            spaxel_iflux = np.zeros(total_num, dtype=np.float64)
```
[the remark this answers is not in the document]
jemorrison: There are too many comments for me to try before I leave. I think this is just going to have to wait until I get back.
Yes the arrays need to be set to zero in the loop.

drlaw1558:
From the science standpoint this seems to be working well.  Previously my desktop could build a 5-28 micron cube from a 4-point dither but crashed when trying to build a 5-28 micron cube from a 12-point dither (i.e., 72 input cal files).  Using this PR it succeeds in both the 12-point dither and even a 57-point dither (342 input cal files).

hbushouse:
A quick grep of all the C routines in cube_build/src/ shows that there are only 8 occurrences of `Py_BuildValue`. If the same fix to memory usage can be accomplished through trivial changes to those 8 lines of code, I agree that that's probably a better way to go, at least for now. Why add unnecessary complexity to the existing python modules, when the same result can be accomplished via these simple updates to the arguments used in `Py_BuildValue`.

PR #7493, opened 2023-03-13
JP-3128: Fix memory leaks in C code

mcara:
Resolves [JP-3086](https://jira.stsci.edu/browse/JP-3086)
Resolves [JP-3128](https://jira.stsci.edu/browse/JP-3128)

This PR fixes memory leaks in C code due to incorrect reference count of returned arrays. This PR addresses a part of #7483 (other improvements in that PR should be merged after that PR is re-based after this PR is merged). Affected packages are: ``cube_build``, ``wfss_contam``, and ``straylight``.

This PR implements the recommendation from https://numpy.org/doc/stable/user/c-info.how-to-extend.html:

> The [Py_BuildValue](https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue) (format_string, c_variables…) function makes it easy to build tuples of Python objects from C variables. **Pay special attention to the difference between ‘N’ and ‘O’ in the format string or you can easily create memory leaks.** The ‘O’ format string increments the reference count of the [PyObject](https://docs.python.org/3/c-api/structures.html#c.PyObject) C-variable it corresponds to, while **the ‘N’ format string steals a reference to the corresponding [PyObject](https://docs.python.org/3/c-api/structures.html#c.PyObject) C-variable. You should use ‘N’ if you have already created a reference for the object and just want to give that reference to the tuple.** You should use ‘O’ if you only have a borrowed reference to an object and need to create one to provide for the tuple.

**Checklist for maintainers**
- [x] added relevant label(s)

Regression test: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/598/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,8 +24,11 @@ other
 
 - Update ``minimum_deps`` script to use ``importlib_metadata`` and ``packaging``
   instead of ``pkg_resources``. [#7457]
+
 - Switch ``stcal.dqflags`` imports to ``stdatamodels.dqflags``. [#7475]
 
+- Fix memory leaks in packages that use C code: ``cube_build``,
+  ``wfss_contam``, and ``straylight``. [#7493]
 
 outlier_detection
 -----------------
@@ -54,7 +57,7 @@ pipeline
 
 - Update the calwebb_spec2 pipeline to move the MIRI MRS straylight step to before
   the flat-fielding step. [#7486]
-  
+
 - Update the calwebb_spec2 pipeline to make a deep copy of the current results before
   calling the ``resample_spec`` and ``extract_1d`` steps, to avoid issues with the
   input data accidentally getting modified by those steps. [#7451]
--- a/jwst/cube_build/src/blot_median.c
+++ b/jwst/cube_build/src/blot_median.c
@@ -1,8 +1,8 @@
 /*
- 
+
 Main function for Python: blot_wrapper
 
-Python signature: 
+Python signature:
             result = blot_wrapper(roi_det, blot_xsize, blot_ysize, xstart, xsize2,
                                   xcenter, ycenter,
                                   x_cube, y_cube, flux_cube)
@@ -46,7 +46,7 @@ Returns
 blot_flux : numpy.ndarray
   IFU spaxel cflux
 blot_weight : numpy.ndarray
-  IFU spaxel weight 
+  IFU spaxel weight
 */
 
 #include <stdlib.h>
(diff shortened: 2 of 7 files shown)


mcara:
I do not see any _new_ failures in the regression tests that I could attribute to the changes proposed in this PR but all tests run in the last month had failures.

hbushouse (approved):
Changes look good to me (very straightforward), and I agree that the failures in the regtest run are unrelated (the truths have since been updated).

Would be nice if someone's got the time to run this PR branch against one of the MIRI MRS mosaics that had been a big memory hog, to verify the same memory reduction as the PR submitted by @jemorrison.

hbushouse:
Does this really fix [JP-3016](https://jira.stsci.edu/browse/JP-3016), which is not related to memory leaks?

mcara:
No. My bad

mcara:
And https://github.com/spacetelescope/jwst/pull/7483 should be rebased after this one is merged.

mcara:
> Would be nice if someone's got the time to run this PR branch against one of the MIRI MRS mosaics that had been a big memory hog, to verify the same memory reduction as the PR submitted by @jemorrison.

From https://github.com/spacetelescope/jwst/pull/7483#issuecomment-1464233450 see:

> I changed how I am using Py_Build and using N rather than 0. The memory does seem to freed up better now, but in the test I have done the memory usage is about as good as this PR.

mcara:
Screen Shot 2023-03-14 at 9 58 40 AM

The bot keeps removing the relevant label

drlaw1558 (approved):
Looks like this is working well.  I tested that (as expected) results from the straylight step are identical before/after the change.  I also tested trying to build a large (78 input files) cube that would ordinarily crash my laptop.  With the current pipeline memory usage grew to about 120 GB before crashing; with this branch memory usage topped out about 17 GB and successfully built the cube.

hbushouse:
OK, it sounds like everyone is on-board with this so I'm going to merge. Will need to coordinate with @jemorrison 's PR once she gets back.

jemorrison:
@hbushouse @mcara @drlaw1558 
Looks like you solved the problem much easier than what I had done. Thanks a ton @mcara. The old PR did a few things that may not be needed at all. It added an option (like outlier detection had) 'in_memory" which if the memory needed was too large it would write the data to disk and then open each file and work with the data. It takes longer to do this. Is this something we might still need - it is only used if 'in_memory = FALSE' (by default it is True).  I will test the updated cube_build and that super large set (over 500 files) and see if helps. This 'in_memory' option also added writing out the IFU cubes to disk as they were created and passed the filenames to extract1d. Again not sure if that is needed.

mcara:
> The old PR did a few things that may not be needed at all. It added an option (like outlier detection had) 'in_memory" which if the memory needed was too large it would write the data to disk and then open each file and work with the data. It takes longer to do this. Is this something we might still need - it is only used if 'in_memory = FALSE' (by default it is True). I will test the updated cube_build and that super large set (over 500 files) and see if helps. This 'in_memory' option also added writing out the IFU cubes to disk as they were created and passed the filenames to extract1d. Again not sure if that is needed.

Why not? If this is a  good option to decrease memory usage and the code is already written - I think you should consider it. Also, it may still be worth moving memory allocation to Python if that memory can be re-used between the calls in a loop (if there is one). I am not that familiar with the code to give my opinion on it.

My only comment about the other PR (which I think should be rebased and reviewed again) was that there were a couple of places in which I thought the code could crash and in the other places it was creating new arrays when I thought those could be reused. I think it may be worth pursuing it. This PR was a stop-gap fix to give more time to review the other one.

PR #7496, opened 2023-03-14
Bugfix for failing security checks

WilliamJamieson:
I noticed that the `check-security` CI job started failing recently. This is my attempt to fix this. 

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/tweakreg/astrometric_utils.py
+++ b/jwst/tweakreg/astrometric_utils.py
@@ -17,6 +17,8 @@ if ASTROMETRIC_CAT_ENVVAR in os.environ:
 else:
     SERVICELOCATION = DEF_CAT_URL
 
+TIMEOUT = 30.0  # in seconds
+
 """
 
 Primary function for creating an astrometric reference catalog.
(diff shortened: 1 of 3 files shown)


jwst/tweakreg/astrometric_utils.py
```diff
@@ -17,6 +17,8 @@
 else:
     SERVICELOCATION = DEF_CAT_URL
 
+TIMEOUT = 30.0  # in seconds
```
WilliamJamieson, line 20: Note that this choice was entirely a guess. The security check failure is due to the lack of a timeout value being passed to `requests.get`. It is considered "best-practice" to pass some sort of "reasonable" timeout value to this so that it will not remain open for too long.

30s seems reasonable to me (30s is the de facto suggestion), but I don't know what the expected response time for the service being accessed is. If the service can be slow, I suggest we give a much larger value.

zacharyburnett (approved):
Looks good, it passes the bandit check. I don't know what the typical return time of the server should be, but 30 seconds for a response seems to be reasonable to me. If we run into issues we can change it in the future

stscieisenhamer (approved):

hbushouse (approved):
Works for me.

PR #7497, opened 2023-03-14
Fix logic in MIRI LRS spec3 regtest

hbushouse:
This PR fixes the currently flawed logic in the test_miri_lrs_slit_spec3 regression test module. The problem is that the spec3 pipeline is run multiple times on the same input data, using different specifications for how to create the output/combined image WCS, but all of those different runs get saved using the same s2d product file name. That's fine for most of the tests, because most result in a combined image of the same size, but the very last iteration creates a bigger image. When the "okify_regtests" script is used to update truth files, that last s2d file, which is different size, gets used to update the truth files for all tests, leading to subsequent failures. This PR updates the test module to save the result of the last iteration in a different output file, so that it doesn't clobber the truths for the other iterations.

**Checklist for maintainers**
- [x] added relevant label(s)


hbushouse:
New truth files have been uploaded to artifactory.

PR #7499, opened 2023-03-15
return opencv-python to a hard dependency

zacharyburnett:
Related to [JP-3120](https://jira.stsci.edu/browse/JP-3120)

This PR adds `opencv-python` back to the hard dependencies in `setup.cfg`. This will prevent installation on platforms that do not have a compiled binary for `opencv-python`, such as MacOS Mojave + Python 3.10.

**Checklist for maintainers**
- [ ] added relevant label(s)

### DIFF
--- a/jwst/jump/tests/test_jump_step.py
+++ b/jwst/jump/tests/test_jump_step.py
@@ -7,12 +7,7 @@ from stdatamodels.jwst.datamodels import GainModel, ReadnoiseModel, RampModel
 
 from jwst.jump import JumpStep
 
-try:
-    import cv2 as cv # noqa: F401
-
-    OPENCV_INSTALLED = True
-except ImportError:
-    OPENCV_INSTALLED = False
+import cv2 as cv
 
 MAXIMUM_CORES = ['none', 'quarter', 'half', 'all']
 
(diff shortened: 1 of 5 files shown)


jwst/jump/tests/test_jump_step.py
```diff
@@ -7,13 +7,6 @@
 
 from jwst.jump import JumpStep
 
-try:
```
hbushouse, line 10: Were there other modules that also had this conditional import? Or are they over in stcal?

hbushouse (commented):
Looks OK, but let's be sure to get a regtest run against this branch before we approve and merge.

jwst/jump/tests/test_jump_step.py
zacharyburnett: yes the conditional import is in `stcal`, so installing it as part of `jwst` should prevent the "no opencv installed" warning from occurring upstream

hbushouse: So in other words, even though the conditional import may no longer be necessary over in stcal, it doesn't hurt to leave it in there, because the attempted import will succeed and everything will be good.

zacharyburnett:
regtest run is at https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST-Developers-Pull-Requests/detail/JWST-Developers-Pull-Requests/603/pipeline

hbushouse:
All of the failures from the regtest are expected differences due to more pixels being flagged by the snowball/shower correction in the jump step (which then propagates all the way downstream). So it looks good. I'm merging!

hbushouse (approved):

PR #7507, opened 2023-03-24
JP-3148: Trap source catalog errors in tweakreg

hbushouse:
Resolves [JP-3148](https://jira.stsci.edu/browse/JP-3148)

This PR simply adds a try/except block to the tweakreg step code where it attempts to build source catalogs for each input image. This addresses situations where something goes wrong in the source catalog generation (e.g. all pixels flagged as bad, etc.), preventing it from completing successfully. Instead of aborting with an error, it now emits a warning and just skips the rest of the tweakreg step. The subsequent pipeline steps can proceed without it.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/tweakreg/tweakreg_catalog.py
+++ b/jwst/tweakreg/tweakreg_catalog.py
@@ -1,3 +1,5 @@
+import logging
+
 from astropy.table import Table
 import numpy as np
 from photutils.detection import DAOStarFinder
@@ -5,6 +7,9 @@ from photutils.detection import DAOStarFinder
 from stdatamodels.jwst.datamodels import dqflags, ImageModel
 
 from ..source_catalog.detection import JWSTBackground
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
 
 
 def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
@@ -73,10 +78,17 @@ def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
                       dqflags.pixel['DO_NOT_USE']) &
                      model.dq).astype(bool)
 
-    bkg = JWSTBackground(model.data, box_size=bkg_boxsize,
-                         coverage_mask=coverage_mask)
+    columns = ['id', 'xcentroid', 'ycentroid', 'flux']
+    try:
+        bkg = JWSTBackground(model.data, box_size=bkg_boxsize,
+                             coverage_mask=coverage_mask)
+        threshold_img = bkg.background + (snr_threshold * bkg.background_rms)
+    except ValueError as e:
+        log.warning(f"Error determining sky background: {e.args[0]}")
+        # return an empty catalog
+        catalog = Table(names=columns, dtype=(int, float, float, float))
+        return catalog
 
-    threshold_img = bkg.background + (snr_threshold * bkg.background_rms)
     threshold = np.median(threshold_img)  # DAOStarFinder requires float
 
     daofind = DAOStarFinder(fwhm=kernel_fwhm, threshold=threshold,
@@ -85,7 +97,6 @@ def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
                             peakmax=peakmax)
     sources = daofind(model.data, mask=coverage_mask)
 
-    columns = ['id', 'xcentroid', 'ycentroid', 'flux']
     if sources:
         catalog = sources[columns]
     else:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -96,7 +96,6 @@ ramp_fitting
 - Update ramp fitting to calculate separate readnoise variance for processing
   ``undersampling_correction`` output [#7484]
 
-
 resample
 --------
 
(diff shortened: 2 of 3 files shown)


hbushouse:
Full regtest started at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/613/

tapastro (approved):
LGTM

mcara (commented):
I am concerned that this fix skips tweakreg step altogether when there could be sources found in other images in an "alignment group" that could allow alignment of all images in the group.

mcara:
An I feel the fix should probably go to:

https://github.com/spacetelescope/jwst/blob/af63faf04760112c16c036cfefa2e5209efc049e/jwst/tweakreg/tweakreg_catalog.py#L71-L93

and just return an empty catalog (just like the second branch on the last `IF` in the code above). The `tweakreg_step` should be able to handle empty catalogs.

It is important to get @larrybradley opinion on this.

mcara:
Specifically, I would propose that the original code in `tweakreg_catalog.py`:

```python
    bkg = JWSTBackground(model.data, box_size=bkg_boxsize,
                         coverage_mask=coverage_mask)

    threshold_img = bkg.background + (snr_threshold * bkg.background_rms)
    threshold = np.median(threshold_img)  # DAOStarFinder requires float

    daofind = DAOStarFinder(fwhm=kernel_fwhm, threshold=threshold,
                            roundhi=roundhi, brightest=brightest,
                            peakmax=peakmax)
    sources = daofind(model.data, mask=coverage_mask)

    columns = ['id', 'xcentroid', 'ycentroid', 'flux']
    if sources:
        catalog = sources[columns]
    else:
        # return an empty table
        catalog = Table(names=columns, dtype=(int, float, float, float))

    return catalog
``` 

be replaced with (proposed fix):

```python
    columns = ['id', 'xcentroid', 'ycentroid', 'flux']
    try:
        bkg = JWSTBackground(model.data, box_size=bkg_boxsize,
                             coverage_mask=coverage_mask)
    except ValueError as e:
        log.warning(f"Error in determining sky background: {e.args[0]}")
        # return an empty table
        catalog = Table(names=columns, dtype=(int, float, float, float))
        return catalog

    threshold_img = bkg.background + (snr_threshold * bkg.background_rms)
    threshold = np.median(threshold_img)  # DAOStarFinder requires float

    daofind = DAOStarFinder(fwhm=kernel_fwhm, threshold=threshold,
                            roundhi=roundhi, brightest=brightest,
                            peakmax=peakmax)
    sources = daofind(model.data, mask=coverage_mask)

    if sources:
        catalog = sources[columns]
    else:
        # return an empty table
        catalog = Table(names=columns, dtype=(int, float, float, float))

    return catalog
```

(This also would need some import of the logging module at the top of the `tweakreg_catalog` module.)

Also, possibly it would be even better to fix `JWSTBackground` to set a flag when background determination fails and then have ``tweakreg_catalog` inspect the return status from `JWSTBackground`.

larrybradley:
@mcara 's proposed changes looks good to me, but I would also suggest adding a warning log message in the `except` block.

@mcara What happens if *all* the catalogs are empty?  Does `tweakreg` give an error?

mcara:
> @mcara 's proposed changes look good to me, but I would also suggest adding a warning log message in the `except` block.

I think there is a log in the except block

> @mcara What happens if _all_ the catalogs are empty? Does `tweakreg` give an error?

It should be fine.

tapastro:
> I am concerned that this fix skips tweakreg step altogether when there could be sources found in other images in an "alignment group" that could allow alignment of all images in the group.

If an image has no catalog, how it will be aligned to the others that have sources? What would be the reference point from which to move the WCS?

jwst/tweakreg/tweakreg_catalog.py
```diff
@@ -73,10 +78,17 @@ def make_tweakreg_catalog(model, kernel_fwhm, snr_threshold, sharplo=0.2,
...
+    columns = ['id', 'xcentroid', 'ycentroid', 'flux']
+    try:
+        bkg = JWSTBackground(model.data, box_size=bkg_boxsize,
+                             coverage_mask=coverage_mask)
+        threshold_img = bkg.background + (snr_threshold * bkg.background_rms)
+    except ValueError as e:
+        log.warning(f"Error determining sky background: {e.args[0]}")
+        # return an empty catalog
+        catalog = Table(names=columns, dtype=(int, float, float, float))
+        return catalog
```
mcara, line 90:
```suggestion
    try:
        bkg = JWSTBackground(model.data, box_size=bkg_boxsize,
                             coverage_mask=coverage_mask)
    except ValueError as e:
        log.warning(f"Error determining sky background: {e.args[0]}")
        # return an empty catalog
        catalog = Table(names=columns, dtype=(int, float, float, float))
        return catalog
        
    threshold_img = bkg.background + (snr_threshold * bkg.background_rms)
```

mcara (changes requested):

jwst/tweakreg/tweakreg_catalog.py
hbushouse: That approach of only including line 83 (bkg = JWSTBackground()) in the `try` block doesn't work, because that line only creates an instance of the "bkg" object. It's line 91 (threshold_img = bkg.background ....) that actually throws the error and hence it needs to be in the `try` block.  Trust me, I tried it that way first and it still crashed.

mcara: Ah, OK.

mcara (approved):
LGTM

mcara:
> Regression test run started at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/615

I can't see how anything can fail here (but I was wrong many times before)

mcara:
I think it would be worth testing that logging works properly. You do not have a unit test for the case when the exception is raised. However, locally you could move the log message from the `except` to the `try` block and check that it indeed appears in the log file. (then move it back; that is, do the test locally only)

jwst/tweakreg/tweakreg_catalog.py
larrybradley: Yes, the background is not actually calculated until the `bkg.background` (lazy)property is requested.

larrybradley (approved):

PR #7508, opened 2023-03-24
JP-3147 Treat PSF exposures as science for Level 2 association processing

stscieisenhamer:
Resolves [JP-3147](https://jira.stsci.edu/browse/JP-3147)

This PR addresses and issue with coronographic observations found while investigating JWSTDMS-770 and documented in JP-3147. The issue is that the PSF observations Level 2 associations were not including their linked background observations. Technically, this is because PSF Image 2 associations were being processed by the rule {{Asn_Lv2ImageSpecial}}, which specifically disallows background candidates.

The solution is that, for Level 2 associations, PSF observations can be treated as normal science, which is handled by the rule {{Asn_Lv2Image}}. This is done by changing the exposure type from "psf" to "science".

**Checklist for maintainers**
- [ ] ~updated relevant documentation~
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@ associations
 
 - Remove extra reprocessing of existing associations. [#7506]
 
+- Treat PSF exposures as science for Level 2 association processing. [#7508]
+
 calwebb_detector1
 -----------------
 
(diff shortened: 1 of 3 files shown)


stscieisenhamer:
[Regression run](https://plwishmaster.stsci.edu:8081/job/RT/job/eisenhamer-dev/319/)

Note: This run is doing only the association tests.

hbushouse (approved):
Looks good to me, but just one question. It looks like there are now multiple definitions of `get_exposure_type` in different classes. Is that kosher?

stscieisenhamer:
That is using standard method overriding as part of the OO approach used by the generated. This is kosher and should be expected.

PR #7515, opened 2023-03-30
Use correct bits in resample_variance_array

### LINKED ISSUE #7514: Error arrays often include pixels not in the science array
The error array in the final level 3 products can often have errors for pixels not in the science array (see attached image). This is caused by the pipeline not properly carrying forward good bits when creating the various error arrays.

Screenshot 2023-03-30 at 12 51 23

thomaswilliamsastro:
Closes #7514

This PR addresses a small bug in the output error arrays, that in MIRI data these will often include the coronagraph observations even if this is not included in the output science array

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/resample/resample.py
+++ b/jwst/resample/resample.py
@@ -285,7 +285,7 @@ class ResampleData:
             inwht = resample_utils.build_driz_weight(
                 model,
                 weight_type=None,
-                good_bits="~NON_SCIENCE+REFERENCE_PIXEL"
+                good_bits=self.good_bits
             )
 
             resampled_variance = np.zeros_like(output_model.data)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -130,6 +130,8 @@ resample
   a bug due to which parts of input images may not be present in the output
   resampled image under certain circumstances. [#7460]
 
+- Carry through good bits correctly for the variance array [#7515]
+
 residual_fringe
 ---------------
 


jwst/resample/resample.py
```diff
@@ -285,7 +285,7 @@ def resample_variance_array(self, name, output_model):
             inwht = resample_utils.build_driz_weight(
                 model,
                 weight_type=None,
-                good_bits="~NON_SCIENCE+REFERENCE_PIXEL"
+                good_bits=self.good_bits
```
hbushouse, line 288: So for the specific case of something like MIRI imager, which also contains data from the coronagraphic mask areas, the coronagraphic areas are not included in the drizzled SCI image because they have something set in `good_bits` that excludes them and hence by also applying `good_bits` to the variance/weight array they'll get excluded from that too?

thomaswilliamsastro: Exactly. In any case, the science image only considers these good_bits anyway, so the error array should do the same to be consistent

mcara:
I wonder what was the reason to hard code these DQ flags in the first place @jdavies-st ?

mcara (approved):

thomaswilliamsastro:
I've rebased onto master but don't have merge privileges. @mcara @hbushouse can one of you merge this in?

jdavies-st:
> I wonder what was the reason to hard code these DQ flags in the first place @jdavies-st ?

Well, the whole "good_bits" was a hack, a carryover from drizzlepac, and a way of handling DQ values at odds with how they are handled everywhere else in every other step for JWST.  For JWST one doesn't define good bits.  They're all assumed to be good. You have to define specific ones that are bad, not just assume that everything greater than zero is bad.  There's a long-standing issue to refactor this, but it looks like it hasn't happened yet.

hbushouse:
> > I wonder what was the reason to hard code these DQ flags in the first place @jdavies-st ?
> 
> Well, the whole "good_bits" was a hack, a carryover from drizzlepac, and a way of handling DQ values at odds with how they are handled everywhere else in every other step for JWST. For JWST one doesn't define good bits. They're all assumed to be good. You have to define specific ones that are bad, not just assume that everything greater than zero is bad. There's a long-standing issue to refactor this, but it looks like it hasn't happened yet.

The primary reason it hasn't happened yet is because someone left STScI!  ;-)

hbushouse:
> I've rebased onto master but don't have merge privileges. @mcara @hbushouse can one of you merge this in?

Will merge after we've had a chance to run regression tests against the PR branch. Stay tuned.

mcara:
Regression tests: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/623/

hbushouse:
Regtest run 623 seems to have gotten clobbered by test 624. So I've started another run for this branch: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/625

PR #7516, opened 2023-03-30
(JP-3094) NIRSpec FS rateints flat_field processing

dmggh:
Resolves [JP-3094](https://jira.stsci.edu/browse/JP-3094)

Closes #

This PR updates flat fielding to be able to process NIRSpec Fixed Slit _rateints input.

**Checklist for maintainers**
- [X] updated relevant documentation
- [X] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -64,6 +64,7 @@ extract_1d
 
 - Edit surface brightness unit strings for parsing by ``astropy.units`` [#7511]
 
+
 jump
 ----
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -148,7 +148,7 @@ resample
 
 - Carry through good bits correctly for the variance array [#7515]
 
- -Update ``resample_spec`` to be skipped for NIRSpec fixed slit modes with
+- Update ``resample_spec`` to be skipped for NIRSpec fixed slit MultiSlitModel
   rateints input as that mode is not allowed. [#7516]  
 
 residual_fringe
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,15 @@ calwebb_detector1
 - Added regression test for ``calwebb_detector1`` pipeline which now
   includes ``undersampling_correction``. [#7509]
 
+
+calwebb_spec2
+-------------
+
+ -Update ``resample_spec`` and ``extract_1d`` step code to be skipped for NIRSpec
+  fixed slit modes with rateints input to allow calwebb_spec2 to run to completion for
+  this mode. [#7516] 
+
+
 cube_build
 ----------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -79,6 +79,7 @@ jump
   emission far below the single pixel SNR. It also allows detected
   showers to flag groups after the detection. [#7478]
 
+
 other
 -----
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,15 +27,6 @@ calwebb_detector1
 - Added regression test for ``calwebb_detector1`` pipeline which now
   includes ``undersampling_correction``. [#7509]
 
-
-calwebb_spec2
--------------
-
- -Update ``resample_spec`` and ``extract_1d`` step code to be skipped for NIRSpec
-  fixed slit modes with rateints input to allow calwebb_spec2 to run to completion for
-  this mode. [#7516] 
-
-
 cube_build
 ----------
 
@@ -73,6 +64,8 @@ extract_1d
 
 - Edit surface brightness unit strings for parsing by ``astropy.units`` [#7511]
 
+- Update to be skipped for NIRSpec fixed slit modes with rateints input as that
+  mode is not allowed. [#7516]
 
 jump
 ----
@@ -87,7 +80,6 @@ jump
   Shower code is completely new and is now able to find extended
   emission far below the single pixel SNR. It also allows detected
   showers to flag groups after the detection. [#7478]
-
 
 other
 -----
@@ -155,6 +147,9 @@ resample
   resampled image under certain circumstances. [#7460]
 
 - Carry through good bits correctly for the variance array [#7515]
+
+ -Update ``resample_spec`` to be skipped for NIRSpec fixed slit modes with
+  rateints input as that mode is not allowed. [#7516]  
 
 residual_fringe
 ---------------
--- a/docs/jwst/pipeline/calwebb_spec2.rst
+++ b/docs/jwst/pipeline/calwebb_spec2.rst
@@ -243,8 +243,13 @@ The input to the ``Spec2Pipeline`` pipeline is a countrate exposure, in the form
 of either "_rate" or "_rateints" data. A single input file can be processed or an
 ASN file listing multiple inputs can be used, in which case the processing steps
 will be applied to each input exposure, one at a time.
-If "_rateints" products are used as input, each step applies its algorithm to each
-integration in the exposure, where appropriate.
+If "_rateints" products are used as input, for modes other than NIRSpec Fixed Slit,
+each step applies its algorithm to each integration in the exposure, where appropriate.
+For the NIRSpec Fixed Slit mode, although the :ref:`flat_field <flat_field_step>`
+will accept "_rateints" products as input, the ``calwebb_spec2`` pipeline will currently
+fail at the :ref:`resample_spec <resample_step>` step which does not support
+multiple integration input products.  In the future, the :ref:`resample_spec <resample_step>`
+step will be modified to be skipped if its input is a "_rateints" product. 
 
 Note that the steps :ref:`background <background_step>` and :ref:`imprint <imprint_step>`
 can only be executed when the pipeline is given an ASN file as input, because they rely on
--- a/docs/jwst/pipeline/calwebb_spec2.rst
+++ b/docs/jwst/pipeline/calwebb_spec2.rst
@@ -243,13 +243,14 @@ The input to the ``Spec2Pipeline`` pipeline is a countrate exposure, in the form
 of either "_rate" or "_rateints" data. A single input file can be processed or an
 ASN file listing multiple inputs can be used, in which case the processing steps
 will be applied to each input exposure, one at a time.
+
 If "_rateints" products are used as input, for modes other than NIRSpec Fixed Slit,
 each step applies its algorithm to each integration in the exposure, where appropriate.
 For the NIRSpec Fixed Slit mode, although the :ref:`flat_field <flat_field_step>`
 will accept "_rateints" products as input, the ``calwebb_spec2`` pipeline will currently
-fail at the :ref:`resample_spec <resample_step>` step which does not support
-multiple integration input products.  In the future, the :ref:`resample_spec <resample_step>`
-step will be modified to be skipped if its input is a "_rateints" product. 
+skip both the :ref:`resample_spec <resample_step>` step and the
+:ref:`extract_1d <extract_1d_step>` step, because neither step supports
+multiple integration input products for this mode.
 
 Note that the steps :ref:`background <background_step>` and :ref:`imprint <imprint_step>`
 can only be executed when the pipeline is given an ASN file as input, because they rely on
--- a/docs/jwst/pipeline/calwebb_spec2.rst
+++ b/docs/jwst/pipeline/calwebb_spec2.rst
@@ -246,8 +246,7 @@ will be applied to each input exposure, one at a time.
 
 If "_rateints" products are used as input, for modes other than NIRSpec Fixed Slit,
 each step applies its algorithm to each integration in the exposure, where appropriate.
-For the NIRSpec Fixed Slit mode, although the :ref:`flat_field <flat_field_step>`
-will accept "_rateints" products as input, the ``calwebb_spec2`` pipeline will currently
+For the NIRSpec Fixed Slit mode the ``calwebb_spec2`` pipeline will currently
 skip both the :ref:`resample_spec <resample_step>` step and the
 :ref:`extract_1d <extract_1d_step>` step, because neither step supports
 multiple integration input products for this mode.
--- a/jwst/extract_1d/extract_1d_step.py
+++ b/jwst/extract_1d/extract_1d_step.py
@@ -172,7 +172,6 @@ class Extract1dStep(Step):
 
         # Open the input and figure out what type of model it is
         input_model = datamodels.open(input)
-        input_model_copy = input_model.copy()  # used if step is skipped
 
         was_source_model = False  # default value
         if isinstance(input_model, datamodels.CubeModel):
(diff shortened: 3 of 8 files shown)


dmggh:
For the NIRSPec Fixed Slit mode with _rateints input, the calwebb_spec2 no longer fails at flat fielding due to a shape error. The pipeline will currently fail at the resample-spec step as that step does not support multiple integration input products. 
In the future, the resample-spec step will be modified to be skipped if its input is a "_rateints" product.

docs/jwst/pipeline/calwebb_spec2.rst
```diff
@@ -243,8 +243,13 @@ The input to the ``Spec2Pipeline`` pipeline is a countrate exposure, in the form
...
 will be applied to each input exposure, one at a time.
-If "_rateints" products are used as input, each step applies its algorithm to each
-integration in the exposure, where appropriate.
+If "_rateints" products are used as input, for modes other than NIRSpec Fixed Slit,
+each step applies its algorithm to each integration in the exposure, where appropriate.
+For the NIRSpec Fixed Slit mode, although the :ref:`flat_field <flat_field_step>`
+will accept "_rateints" products as input, the ``calwebb_spec2`` pipeline will currently
+fail at the :ref:`resample_spec <resample_step>` step which does not support
+multiple integration input products.  In the future, the :ref:`resample_spec <resample_step>`
+step will be modified to be skipped if its input is a "_rateints" product.
```
nden, line 252: The future is now. We need a clean run through calwebb_spec2, not piecewise changes.

hbushouse: Agreed. We can't merge code that we know will fail and describe the expected failure in the docs. Something has to be done to the code to avoid the failure, even if that just means adding logic to skip the resample_spec step.

dmggh: I agree with you both. I'm working on that code change right now.

nden:
Tests are failing

jwst/extract_1d/extract_1d_step.py
```diff
@@ -186,6 +187,15 @@ def process(self, input):
         elif isinstance(input_model, ModelContainer):
             self.log.debug('Input is a ModelContainer')
         elif isinstance(input_model, datamodels.MultiSlitModel):
+            #  If input is a 3D rateints (which is unsupported) skip the step
+            try:
+                if len((input_model[0]).shape) == 3:
+                    self.log.warning('ext_1d_step step will be skipped as 3D input is unsupported')
```
hbushouse, line 193: The log already tells you what step is being run, so this can just say "3D input is not supported; step will be skipped"

jwst/extract_1d/extract_1d_step.py
```diff
@@ -186,6 +187,15 @@ def process(self, input):
...
+                    input_model_copy.meta.cal_step.extract_1d = 'SKIPPED'
```
hbushouse, line 194: Again, don't see the need for a copy here. Just update the meta value in the original input model and return it.

jwst/regtest/test_nirspec_fs_spec2.py
```diff
@@ -135,3 +136,26 @@ def test_pathloss_source_type(rtdata):
...
+    """
+    rtdata = rtdata_module
+
+    rtdata.get_data("nirspec/fs/jw01128004001_03102_00001_nrs1_new_rateints.fits")
+
+    # Run the begining of the spec2 pipeline on a (3D) _rateints file
+    args = ["calwebb_spec2", rtdata.input,
+            "--steps.pathloss.skip=True",
+            "--steps.photom.skip=True",
+            "--steps.resample_spec.skip=True",
```
hbushouse, line 154: Given that this PR updates the resample_spec and extract_1d steps to skip when given 3D input, is it necessary to explicitly skip them here? Won't it just happen automatically?

jwst/resample/resample_spec_step.py
```diff
@@ -29,6 +28,7 @@ class ResampleSpecStep(ResampleStep):
     def process(self, input):
         self.wht_type = self.weight_type
         input_new = datamodels.open(input)
+        input_new_copy = input_new.copy()  # used if step is skipped
```
hbushouse, line 31: Don't need the copy.

jwst/resample/resample_spec_step.py
```diff
@@ -83,6 +83,17 @@ def process(self, input):
 
         # Call resampling
         self.drizpars = kwargs
+
+        #  If input is a 3D rateints (which is unsupported) skip the step
+        try:
+            if len((input_models[0][0]).shape) == 3:
+                self.log.warning('Resample spec step will be skipped')
+                input_new_copy.meta.cal_step.resample_spec = 'SKIPPED'
```
hbushouse, line 91: Just update the original input, without copying.

jwst/resample/resample_spec_step.py
```diff
@@ -83,6 +83,17 @@ def process(self, input):
 
         # Call resampling
         self.drizpars = kwargs
+
+        #  If input is a 3D rateints (which is unsupported) skip the step
+        try:
+            if len((input_models[0][0]).shape) == 3:
+                self.log.warning('Resample spec step will be skipped')
```
hbushouse, line 90: Should say why the step is being skipped, e.g. "3D input data not supported; step will be skipped"

CHANGES.rst
```diff
@@ -27,6 +27,13 @@ calwebb_detector1
 - Added regression test for ``calwebb_detector1`` pipeline which now
   includes ``undersampling_correction``. [#7509]
 
+calwebb_spec2
```
hbushouse, line 30: It isn't the calwebb_spec2 pipeline module that's being updated here, it's the individual steps (resample_spec and extract_1d)

jwst/extract_1d/extract_1d_step.py
```diff
@@ -172,6 +172,7 @@ def process(self, input):
 
         # Open the input and figure out what type of model it is
         input_model = datamodels.open(input)
+        input_model_copy = input_model.copy()  # used if step is skipped
```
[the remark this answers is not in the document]
dmggh: Fixed

jwst/extract_1d/extract_1d_step.py
dmggh: fixed

jwst/regtest/test_nirspec_fs_spec2.py
dmggh: Right

jwst/resample/resample_spec_step.py
dmggh: Fixed

jwst/resample/resample_spec_step.py
dmggh: Fixed

docs/jwst/pipeline/calwebb_spec2.rst
```diff
@@ -243,8 +243,14 @@ The input to the ``Spec2Pipeline`` pipeline is a countrate exposure, in the form
 of either "_rate" or "_rateints" data. A single input file can be processed or an
 ASN file listing multiple inputs can be used, in which case the processing steps
 will be applied to each input exposure, one at a time.
-If "_rateints" products are used as input, each step applies its algorithm to each
-integration in the exposure, where appropriate.
+
+If "_rateints" products are used as input, for modes other than NIRSpec Fixed Slit,
+each step applies its algorithm to each integration in the exposure, where appropriate.
+For the NIRSpec Fixed Slit mode, although the :ref:`flat_field <flat_field_step>`
```
hbushouse, line 249: You don't need to single out the flat_field step here, because it's not unique in terms of handling 3D data. This PR makes it equal to all the other spec2 steps which were already capable of handling 3D data. So I think this middle phrase should be dropped and just have it say "For the NIRSpec Fixed Slit mode the calwebb_spec2 pipeline will ..."

jwst/regtest/test_nirspec_fs_spec2.py
```diff
@@ -135,3 +136,25 @@ def test_pathloss_source_type(rtdata):
...
+       only up to and including the flat-field step. This is a test that the
+       flat-field step does not fail processing this multi-integration input.
+    """
+    rtdata = rtdata_module
+
+    rtdata.get_data("nirspec/fs/jw01128004001_03102_00001_nrs1_new_rateints.fits")
+
+    # Run the begining of the spec2 pipeline on a (3D) _rateints file
+    args = ["calwebb_spec2", rtdata.input,
+            "--steps.pathloss.skip=True",
```
hbushouse, line 153: I would let the pathloss and photom steps be applied in this test. That way we have at least one test that sends 3D data through all spec2 steps, except of course resample_spec and extract_1d, so it'll serve as a test of the 3D handling for all steps.

jwst/regtest/test_nirspec_fs_spec2.py
```diff
@@ -135,3 +136,25 @@ def test_pathloss_source_type(rtdata):
             if not np.allclose(slit.data, slit.pathloss_uniform, equal_nan=True):
                 bad_slits.append(idx)
     assert not bad_slits, f'Force to uniform failed for slits {bad_slits}'
+
+
+@pytest.mark.bigdata
+def test_short_calwebb_spec2_rateints(rtdata_module):
```
hbushouse, line 142: Since this will now act as a generic test of sending a 3D FS (rateints) file through all of spec2, I would just call it "test_nirspec_fs_rateints_spec2" (in keeping with existing test names in this module).

CHANGES.rst
dmggh: Fixed

hbushouse:
Error in RTD build: jwst-pipeline/checkouts/7516/docs/jwst/pipeline/calwebb_spec2.rst:247: WARNING: undefined label: 'flat_field_step'

jwst/regtest/test_nirspec_fs_spec2.py
dmggh: Done

jwst/regtest/test_nirspec_fs_spec2.py
dmggh: Done

docs/jwst/pipeline/calwebb_spec2.rst
dmggh: Done

dmggh:
That string does not appear in the code I just pushed ( about 12 minutes ago ), so that error should no longer occur.

jwst/resample/resample_spec_step.py
```diff
@@ -83,6 +82,17 @@ def process(self, input):
 
         # Call resampling
         self.drizpars = kwargs
+
+        #  If input is a 3D rateints (which is unsupported) skip the step
+        try:
+            if len((input_models[0][0]).shape) == 3:
```
nden, line 88: I am trying to understand this logic. `input_models`, according to the docs, can be a `MultiSlitModel`, a `SlitModel`, an`ImageModel`, or a `ModelContainer`. In all cases it is turned into a `ModelContainer` by the time it reaches this logic. The `ModelContainer` can have models which are "single" models, like `SlitModel` or models which are instances of `MultiSlitModel`. 
I think the logic here works correctly in the case when the input is a `MultiSlitModel`. Otherwise it raises an `AssetionError` but this error is raised for unrelated reasons. Essentially it is raised in the code which flattens the `meta` and allows access using  `.` (dots), as in `model.meta....` and has nothing to do with the data shape. I think better logic is necessary.

nden:
A regression test run is needed.

dmggh:
I started reg test run #632 at Thu Apr  6 01:45:12 EDT 2023

jwst/resample/resample_spec_step.py
dmggh: For the rateints check for NIRSpec fixed slits, I updated the code to also check if the input is a MultiSlitModel.

nden: I don't think this was the right solution. 

The goal is to check the shape of the data array. The current implementation checks that an `AssertionError` is raised as a sign that the data array is 3D and this is incorrect. This `AssertionError` is raised in the base `DataModel` class, in the function that allows `.` dot access to the `meta` attribute (for example, instead of typing `model['meta'].instance['instrument']['name']` one can type `model.meta.instrument.name`. ) The code in the base class that does this uses `__getitem__` and expects the key to be of type string, not integer as is the typical case when indexing. Internally the function uses `assert`, hence the `AssertionError`, and it has nothing to do with the shape of the data array. So a `try ... except` block that catches an `AssertionError` is not an evidence that the array is 3D.

There are various alternative solutions, the simplest one may be to switch the order of checks in the condition from
```
if (len((input_models[0][0]).shape) == 3) and model_is_msm:
```
to
```
if model_is_msm and len((input_models[0][0]).shape) == 3:
```
the difference being that if it's not a `MultiSlit` model it doesn't reach the code that raises `AssertionError`. 

That said, this check should be done in the beginning of the function, immediately after opening the input and then it will be much simpler and faster because it won't have to package the input data into a `ModelContainer`. A `MultiSlit` model is created in `extract_2d`. It's either a grism observation or a Nirspec observation, excluding BOTS mode (which creates a single `SlitModel`). Since grism observations are not resampled I think it's OK to check that the input is a `MultiSlitModel` immediately after the input is opened (as you do now on line 33) and if so check that the data array of the first slit is 3D. I think that would be the cleanest solution.

nden (changes requested):
Needs a better check for 3D data in resample_spec and extract_1d.

jwst/extract_1d/extract_1d_step.py
```diff
@@ -186,6 +186,15 @@ def process(self, input):
...
             self.log.debug('Input is a ModelContainer')
         elif isinstance(input_model, datamodels.MultiSlitModel):
+            #  If input is a 3D rateints (which is unsupported) skip the step
+            try:
+                if len((input_model[0]).shape) == 3:
+                    self.log.warning('3D input is unsupported; step will be skipped')
+                    input_model.meta.cal_step.extract_1d = 'SKIPPED'
+                    return input_model
+            except AssertionError:
+                pass
```
nden, line 196: The same problem with the `try ... except` block here as in resample_step - it's incorrect to check for `AssertionError`. The code already has a check for `MultiSlitModel`. All that's needed is to check that the data array in the first slit is 3D, i.e. `input_model[0].data.shape == 3`

hbushouse: Ah, yes I see it now too (mind was numb the first time through). There's no need for a try/except block here. Nothing is going to throw an error that needs to be caught. Simply check for a `MultiSlitModel` as input and then check the shape of that model. If it's 3D, skip.  Same would apply to the test in the resample_spec step.

dmggh: Made those updates to extract_1d_step.py  and resample_spec_step.py

nden:
Code looks good now. Is there a regression test run?

dmggh:
Will start a reg test run now.

nden:
I ran [another regression test job](https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/634/#showFailuresLink) because the branch was not set up correctly and it was running with jwst/master.

nden (approved):

hbushouse (approved):
code looks good now, and all regtests and CI's look good.

PR #7521, opened 2023-03-31
Cube Build JP 3016 allow custom suffix and other improvements

### LINKED ISSUE #7367: cube_build does not allow a custom suffix
`CubeBuildStep` ignores custom suffixes.

https://github.com/spacetelescope/jwst/blob/5a0634e30e04fd422403afda357f6e4fdf64409f/jwst/cube_build/cube_build_step.py#L78

It should allow something like:

```python
step = CubeBuildStep()
step.suffix = "test_s3d"
```

Currently this is ignored.

Probably something like:

```python
if self.suffix is None:
    self.suffix = 's3d'
```

would be sufficient.

jemorrison:
Resolves [JP-3016](https://jira.stsci.edu/browse/JP-3016)

Closes #7367 

This PR addresses allowing the user to provide a custom suffix.  Some improvements were made to manage the memory more efficiently. A few arrays that were not being used any longer were removed.  In addition, it removed the option for users with MIRI data to use the coord_system='internal_cal'. 'internal_cal' was designed for NIRSpec and never tested in commissioning. Instead miri users wanting an IFUcube aligned with the ifu plane should use 'coord_system=ifualign'.

A very large effort was made to see if having an option 'in_memory' the way outlier_detection uses such an option was tested. Basically setting in_memory=False would write out out IFU cubes and single IFUCubes if doing outlier detection to disk instead of storing them in memory.  When just build typical IFUCube this did not save much memory. So I have put adding this option to cube_build until we better know how outlier detection for IFU is going to be re-designed. 

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/cube_build/cube_build.py
+++ b/jwst/cube_build/cube_build.py
@@ -16,7 +16,6 @@ class CubeData():
 
     def __init__(self,
                  input_models,
-                 #input_filenames,
                  par_filename,
                  **pars):
         """ Initialize the high level of information for the ifu cube
@@ -31,15 +30,12 @@ class CubeData():
         Parameters
         ----------
         input_models : list of data models
-        input_files : str
-          list of fits filenames
         par_filename: str
           cube parameter reference filename
         pars : dictionary holding top level cube parameters
         """
 
         self.input_models = input_models
-        #self.input_filenames = input_filenames
         self.par_filename = par_filename
         self.single = pars.get('single')
         self.channel = pars.get('channel')
@@ -49,7 +45,6 @@ class CubeData():
         self.weighting = pars.get('weighting')
         self.output_type = pars.get('output_type')
         self.instrument = None
-        self.in_memory = pars.get('in_memory')
 
         self.all_channel = []
         self.all_subchannel = []
@@ -82,9 +77,8 @@ class CubeData():
 # Read in the input data (association table or single file)
 # Fill in MasterTable   based on Channel/Subchannel  or filter/grating
 # ______________________________________________________________________________
-        master_table = file_table.FileTable(self.in_memory)
+        master_table = file_table.FileTable()
         instrument = master_table.set_file_table(self.input_models)
-                                                 #self.input_filenames)
 # _______________________________________________________________________________
 # find out how many files are in the association table or if it is an single
 # file store the input_filenames and input_models
--- a/jwst/cube_build/cube_build.py
+++ b/jwst/cube_build/cube_build.py
@@ -16,7 +16,7 @@ class CubeData():
 
     def __init__(self,
                  input_models,
-                 input_filenames,
+                 #input_filenames,
                  par_filename,
                  **pars):
         """ Initialize the high level of information for the ifu cube
@@ -39,7 +39,7 @@ class CubeData():
         """
 
         self.input_models = input_models
-        self.input_filenames = input_filenames
+        #self.input_filenames = input_filenames
         self.par_filename = par_filename
         self.single = pars.get('single')
         self.channel = pars.get('channel')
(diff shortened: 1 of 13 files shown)


jemorrison:
There are changes in cube_build_step at the end of the program that pull out how we deal with the single IFU cube case the typical IFUs. This was done when I was writing out the output files for testing in_memory=False option. Depending on how we update the outlier detection code I might need to write out the files to disk - so I just kept these changes.

drlaw1558:
Can't say as it's obvious to me what the majority of those changes do, but presumably the various test cases have unchanged results?

jwst/cube_build/cube_build.py
```diff
@@ -16,7 +16,6 @@ class CubeData():
 
     def __init__(self,
                  input_models,
-                 input_filenames,
```
hbushouse, line 19: This change in the API is, I believe, responsible for all the unit tests that are throwing hard errors. The unit tests need updating. See the CI test results for the errors.

hbushouse:
Regression tests started at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/626

jwst/cube_build/ifu_cube.py
```diff
@@ -853,6 +856,7 @@ def build_ifucube_single(self):
                 ifucube_model, status = result
 
                 single_ifucube_container.append(ifucube_model)
+                print('Cube build length', len(single_ifucube_container))
```
hbushouse, line 859: Ack, no print statements!

hbushouse:
Another regtest run started at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/628

hbushouse:
Still getting some hard errors in regression tests: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/628/testReport/

jwst/cube_build/ifu_cube.py
```diff
@@ -843,6 +845,7 @@ def build_ifucube_single(self):
                     self.spaxel_var = self.spaxel_var + np.asarray(spaxel_var, np.float64)
                     self.spaxel_iflux = self.spaxel_iflux + np.asarray(spaxel_iflux, np.float64)
                     result = None
+                    del result, spaxel_flux, spaxel_var, spaxek_iflux
```
hbushouse, line 848: This is what's killing all the regression tests: there's a typo in "spaxek_iflux" (last argument in the list). Should be "spaxel_iflux".

hbushouse: Surprised the pep8/codestyle checks didn't find that.

hbushouse:
Latest regtest run has solved all errors but one: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/630/testReport/jwst.regtest/test_miri_cubebuild/_stable_deps__test_cube_build_miri_internal_cal/
It's a file not found error, perhaps due to the fact that the new suffix handling might be creating an output file with a different name?

jemorrison:
Ha. We do not allow miri internal_cal any more. That was never confirmed in commissioning. Instead for these types of cubes the team used coord_system='ifu_align'.
I can update the test and create a truth file for this type. 
We do allow NIRSPEC to use internal_cal.

hbushouse:
Yay! We **FINALLY** got an error-free regtest run at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/631/.  Just shows 1 minor difference in a keyword comment field, which is expected. So I think this is good to go!

jemorrison:
Hallelujah !

hbushouse (approved):

PR #7526, opened 2023-04-07
Remove pytest-openfiles

WilliamJamieson:
`pytest-openfiles` has been deprecated and is interfering with the use of fixtures in our test code (ones that handle files). This PR switches `ResourceWarnings` into errors which will catch almost all cases of files left open. The only cases it does not catch is if an open file handle is assigned to a global variable as the "warning" will not be emitted until python stops running, which occurs after pytest has completed.

Similar to spacetelescope/romancal#666

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 1.10.1 (unreleased)
 ===================
 
+- Remove use of deprecated ``pytest-openfiles`` ``pytest`` plugin. This has been replaced by
+  catching ``ResourceWarning``s. [#7526]
 
 
 1.10.0 (2023-04-03)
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -92,6 +92,9 @@ other
 
 - add `opencv-python` to hard dependencies for usage of snowball detection in the jump step in `stcal` [#7499]
 
+- Remove use of deprecated ``pytest-openfiles`` ``pytest`` plugin. This has been replaced by
```
hbushouse, line 95: This new entry needs to go into the unreleased (1.10.1) section at the top of the log (version 1.10.0 has already been released)

hbushouse (commented):
Just fix the log entry location and it'll be good to go.

nden (approved):

PR #7541, opened 2023-04-11
Enable memory saving code in skymatch

mcara:
This PR enables memory saving code in the `skymatch` step first introduced in https://github.com/spacetelescope/jwst/pull/6874. Unfortunately, I do not think that the pipeline allows steps to pass ASN file names as inputs and outputs and so that memory-saving code was disabled in https://github.com/spacetelescope/jwst/pull/6922.

Unfortunately, the way the code was disabled does not allow for the memory-saving code to be covered by unit tests.

This PR allows enabling this code at the time of step object initialization but it is still disabled by default and there is no step argument to enable it. However, this change allows unit testing and experimental usage of this feature for development purpose.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/skymatch/skymatch_step.py
+++ b/jwst/skymatch/skymatch_step.py
@@ -18,7 +18,7 @@ from astropy.nddata.bitmask import (
 )
 
 from stdatamodels.jwst.datamodels.dqflags import pixel
-from stdatamodels.jwst.datamodels.util import open as datamodel_open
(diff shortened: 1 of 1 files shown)


mcara:
Regression tests are running in https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/645/

hbushouse (approved):

PR #7543, opened 2023-04-12
Remove `codecov` dependency

WilliamJamieson:
`codecov` was yanked from PyPi the morning of April 12, 2023. This is causing installs for testing `jwst` to fail. This PR is an attempt to fix that.

**Checklist for maintainers**
- [N/A] ~updated or added relevant tests~
- [N/A] ~updated relevant documentation~
- [N/A] ~added relevant milestone~
- [x] added relevant label(s)
- [N/A] ~Make sure the JIRA ticket is [resolved properly](https://github.com/spacetelescope/jwst/wiki/How-to-resolve-JIRA-issues)~

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,8 @@ other
 - Remove use of deprecated ``pytest-openfiles`` plugin. This has been replaced by
   catching ``ResourceWarning``s. [#7526]
 
+- Remove use of ``codecov`` package. [#7543]
+
 photom
 ------
 
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -89,7 +89,9 @@ bc0.test_cmds = [
     --env=${artifactoryenv} ${pytest_args}",
     // The following line needs single quotes to avoid insecure interpolation
     // of the codecov_token
-    'codecov --token=${codecov_token} -F nightly',
+    "curl -Os https://uploader.codecov.io/latest/linux/codecov",
+    "chmod +x codecov",
+    './codecov --token=${codecov_token} -F nightly',
 ]
 bc0.test_configs = [data_config]
 bc0.failedFailureThresh = 0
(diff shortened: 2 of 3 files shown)


hbushouse:
Any explanation from codecov as to *why* it as yanked from pypi?

Also, does this removal just mean that codecov won't be run during CI and/or regression tests?

WilliamJamieson:
> Any explanation from codecov as to _why_ it as yanked from pypi?
> 
> Also, does this removal just mean that codecov won't be run during CI and/or regression tests?

Apparently the `codecov` package has been deprecated for over a year now, so I guess someone decided to fully decommission it.

Also, we do not make any use of `codecov` anymore in `jwst` as all of the coverage is handled by the `pytest-cov` plugin which functions perfectly fine.

WilliamJamieson:
> It looks like `codecov` is used here:
> 
> https://github.com/spacetelescope/jwst/blob/33f102e778fab17f34607c6e2303b5cac88382d0/JenkinsfileRT#L92

This should be a replacement for it: https://github.com/codecov/uploader.

hbushouse:
> > It looks like `codecov` is used here:
> > https://github.com/spacetelescope/jwst/blob/33f102e778fab17f34607c6e2303b5cac88382d0/JenkinsfileRT#L92
> 
> This should be a replacement for it: https://github.com/codecov/uploader.

So does something need to be added to this PR to make use of github.com/codecov/uploader or is that already handled?

zacharyburnett:
> > > It looks like `codecov` is used here:
> > > https://github.com/spacetelescope/jwst/blob/33f102e778fab17f34607c6e2303b5cac88382d0/JenkinsfileRT#L92
> > 
> > 
> > This should be a replacement for it: https://github.com/codecov/uploader.
> 
> So does something need to be added to this PR to make use of github.com/codecov/uploader or is that already handled? Or do we just eliminate its use?

If time is of the essence, we can eliminate it now and add `codecov-uploader` in another PR

hbushouse (approved):

PR #7544, opened 2023-04-12
disable MacOS regression tests on Jenkins

zacharyburnett:
MacOS regression tests have numerical differences from the Linux ones; additionally, the Boyle Jenkins machine is currently unstable.

**Checklist for maintainers**
- [N/A] ~added entry in `CHANGES.rst` within the relevant release section~
- [N/A] ~updated or added relevant tests~
- [N/A] ~updated relevant documentation~
- [N/A] ~added relevant milestone~
- [x] added relevant label(s)
- [N/A] ~Make sure the JIRA ticket is [resolved properly](https://github.com/spacetelescope/jwst/wiki/How-to-resolve-JIRA-issues)~

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -79,6 +79,7 @@ bc0.build_cmds = [
     "pip install certifi -U --force-reinstall",
     "pip install -e .[test,sdp] --no-cache-dir",
     "pip install pytest-xdist",
+    "pip install -r requirements-sdp.txt",
     "pip freeze",
 ]
 bc0.build_cmds = PipInject(env.OVERRIDE_REQUIREMENTS) + bc0.build_cmds
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -100,19 +100,7 @@ bc0.failedFailureThresh = 0
 bc1 = utils.copy(bc0)
 bc1.nodetype = 'macos'
 bc1.name = 'macos-stable-deps'
-bc1.build_cmds = [
-    "pip install certifi -U --force-reinstall",
-    "pip install -e .[test,sdp] --no-cache-dir",
-    "pip install git+https://github.com/spacetelescope/stdatamodels.git@master git+https://github.com/spacetelescope/stcal.git@main",
-    "pip install pytest-xdist",
-    "pip freeze",
-]
-bc1.build_cmds = PipInject(env.OVERRIDE_REQUIREMENTS) + bc1.build_cmds
-// bc1.test_cmds = [
-    // "pytest --cov-report=xml --cov=./ -r sxf -n auto --bigdata --slow \
-    // --basetemp=${pytest_basetemp} --junit-xml=results.xml --dist=loadscope \
-    // --env=${artifactoryenv} ${pytest_args}",
-// ]
+bc1.test_cmds = []
 bc1.test_configs = []
 
 utils.run([jobconfig, bc0, bc1])
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -78,7 +78,6 @@ bc0.pip_reqs_files = ['requirements-sdp.txt']
 bc0.build_cmds = [
     "pip install certifi -U --force-reinstall",
     "pip install -e .[test,sdp] --no-cache-dir",
-    "pip install git+https://github.com/spacetelescope/stdatamodels.git@master git+https://github.com/spacetelescope/stcal.git@main",
     "pip install pytest-xdist",
     "pip freeze",
 ]
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -108,11 +108,11 @@ bc1.build_cmds = [
     "pip freeze",
 ]
 bc1.build_cmds = PipInject(env.OVERRIDE_REQUIREMENTS) + bc1.build_cmds
-bc1.test_cmds = [
-    "pytest --cov-report=xml --cov=./ -r sxf -n auto --bigdata --slow \
-    --basetemp=${pytest_basetemp} --junit-xml=results.xml --dist=loadscope \
-    --env=${artifactoryenv} ${pytest_args}",
-]
+// bc1.test_cmds = [
+    // "pytest --cov-report=xml --cov=./ -r sxf -n auto --bigdata --slow \
+    // --basetemp=${pytest_basetemp} --junit-xml=results.xml --dist=loadscope \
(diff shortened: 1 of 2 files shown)


JenkinsfileRT
```diff
@@ -108,11 +108,11 @@ bc1.build_cmds = [
     "pip freeze",
 ]
 bc1.build_cmds = PipInject(env.OVERRIDE_REQUIREMENTS) + bc1.build_cmds
```
hbushouse, line 110: By leaving the `bc1.build_cmds` uncommented, does that mean it's still going to do the build, but just not actually run the tests? Or is the build not executed until `bc1.test_cmds` is executed?

hbushouse:
> @hbushouse should I also disable the MacOS tests on the `JenkinsfileRT_dev` build as well?

Yes. That one's almost more important to shutdown than the regular RT, because that's the one that gets used multiple times a day by folks trying to test their PR's.  Whereas the regular RT typically only gets run once or twice a day. But it still should be removed from both for now.

hbushouse:
> > @hbushouse should I also disable the MacOS tests on the `JenkinsfileRT_dev` build as well?
> 
> Yes. That one's almost more important to shutdown than the regular RT, because that's the one that gets used multiple times a day by folks trying to test their PR's. Whereas the regular RT typically only gets run once or twice a day. But it still should be removed from both for now.

Whoops, sorry. My bad. JenkinsfileRT is used by both our regular nightly RT runs and the Developer's PR branch runs. JenkinsfileRT_dev is only used by the weekly devdeps test. So perhaps we could leave it in JenkinsfileRT_dev, so that we at least attempt to get a Mac run once a week. So please ignore my previous reply.

JenkinsfileRT
zacharyburnett: It should only run the build commands on `utils.run()`, so yes in this current config it will build and not test

hbushouse: So isn't it a waste of time to create the build and then not run it?  Why not also comment out the build?

JenkinsfileRT_dev
```diff
@@ -103,11 +103,6 @@ bc1.build_cmds = [
     "pip freeze",
 ]
 bc1.build_cmds = PipInject(env.OVERRIDE_REQUIREMENTS) + bc1.build_cmds
-bc1.test_cmds = [
```
hbushouse, line 106: Wait, I thought we had decided to leave the Mac build and test active in the `dev` file, which gets run once a week by the devdeps test.

zacharyburnett: I realized that this should be copied from the `bc0` anyway, so having it here is redundant; I'll check that it runs in the job.

zacharyburnett:
https://github.com/spacetelescope/jwst/blob/33f102e778fab17f34607c6e2303b5cac88382d0/JenkinsfileRT#L81

Is this still necessary? It might cause issues with the downstream ops build if these versions are not pinned to releases

JenkinsfileRT
zacharyburnett: https://github.com/spacetelescope/jwst/blob/33f102e778fab17f34607c6e2303b5cac88382d0/JenkinsfileRT#L97-L99

I forgot, this is what creates the `macos-stable_deps.txt`, so it is necessary to have here

zacharyburnett:
regtests at https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST-Developers-Pull-Requests/detail/JWST-Developers-Pull-Requests/648/pipeline

hbushouse:
> https://github.com/spacetelescope/jwst/blob/33f102e778fab17f34607c6e2303b5cac88382d0/JenkinsfileRT#L81
> 
> Is this still necessary? It might cause issues with the downstream ops build if these versions are not pinned to releases

I thought that when we do releases we always create a special requirements file that has absolutely *everything* pinned to a specific version. We just like to let a few of them (stcal, stdatamodels) float here, so that we're always using the latest code for those particular dependencies.

zacharyburnett:
> > https://github.com/spacetelescope/jwst/blob/33f102e778fab17f34607c6e2303b5cac88382d0/JenkinsfileRT#L81
> > 
> > Is this still necessary? It might cause issues with the downstream ops build if these versions are not pinned to releases
> 
> I thought that when we do releases we always create a special requirements file that has absolutely _everything_ pinned to a specific version. We just like to let a few of them (stcal, stdatamodels) float here, so that we're always using the latest code for those particular dependencies.

We do, `requirements-sdp.txt`. Unfortunately, however, in the build pipeline `requirements-sdp.txt` and / or `requirements-dev.txt` are installed *before* the `build_cmds` commands, so keeping these here would override the SDP pins in the final build environment. The files that are pushed to Artifactory and then deployed to DMS are derived from the final build environment in Jenkins, so don't necessarily align with `requirements-sdp.txt`

JenkinsfileRT
```diff
@@ -78,8 +78,8 @@ bc0.pip_reqs_files = ['requirements-sdp.txt']
 bc0.build_cmds = [
     "pip install certifi -U --force-reinstall",
     "pip install -e .[test,sdp] --no-cache-dir",
-    "pip install git+https://github.com/spacetelescope/stdatamodels.git@master git+https://github.com/spacetelescope/stcal.git@main",
     "pip install pytest-xdist",
+    "pip install -r requirements-sdp.txt",
```
zacharyburnett, line 82: Actually, I can try just installing it here; if this works, then we can also use the master commits of the other dependencies.

zacharyburnett:
regtests are blocked by #7543

hbushouse:
#7543 is merged now, so you can rebase and try the tests.

zacharyburnett:
regtest running at https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST-Developers-Pull-Requests/detail/JWST-Developers-Pull-Requests/652/pipeline

hbushouse:
@zacharyburnett @WilliamJamieson Is it OK for me to go ahead and merge this, so that all of our other tests we're trying to run (e.g. on codecov changes) will run faster and cleaner?

zacharyburnett:
> @zacharyburnett @WilliamJamieson Is it OK for me to go ahead and merge this, so that all of our other tests we're trying to run (e.g. on codecov changes) will run faster and cleaner?

yes it looks like this is going through successfully in the regtests, so we can merge it and sort out any issues later

hbushouse (approved):

PR #7547, opened 2023-04-12
re-add latest-commit installs of stdatamodels and stcal

zacharyburnett:
This PR addresses the discrepancies in truth files from the change in #7544 ; it appears that using an additional call to `pip install -r requirements-sdp.txt` within the build command is a valid workaround for the previously stated issue.

**Checklist for maintainers**
- [N/A] added entry in `CHANGES.rst` within the relevant release section
- [N/A] updated or added relevant tests
- [N/A] updated relevant documentation
- [N/A] added relevant milestone
- [x] added relevant label(s)
- [N/A] Make sure the JIRA ticket is [resolved properly](https://github.com/spacetelescope/jwst/wiki/How-to-resolve-JIRA-issues)

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -79,6 +79,8 @@ bc0.build_cmds = [
     "pip install certifi -U --force-reinstall",
     "pip install -e .[test,sdp] --no-cache-dir",
     "pip install pytest-xdist",
+    // these latest-commit installs are overridden by `requirements-sdp.txt` if it is populated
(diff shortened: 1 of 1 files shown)


zacharyburnett:
regression tests running at https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST-Developers-Pull-Requests/detail/JWST-Developers-Pull-Requests/654/pipeline

hbushouse (approved):
The test log shows that it's again using the master branches of stcal and stdatamodels, so it looks like this works as intended.

Although we should also confirm that when we do a release with a populated version of requirements-sdp.txt it does NOT use the master branches.

PR #7550, opened 2023-04-13
Refactor 1D flat interpolation for NIRSpec flatfield

melanieclarke:
The flatfield step for NIRSpec IFU data currently has a significant bottleneck in the combine_fast_slow function, in which a 1D flat array is interpolated onto the wavelengths for a 2D array with different wavelength sampling.  The interpolation takes place inside a loop over the 2D image pixels, which is inherently slow in Python.

This PR rearranges the interpolation code to use numpy interpolation functions on the whole 2D array at once, which is much faster than the previous version.  In local benchmarks on a single IFU test file, the previous version takes 240 seconds to perform flat fielding; the new version takes 100 seconds.  The output files from both versions are identical.

**Checklist for maintainers**
- [ ] added relevant label(s)


jemorrison (approved):
Looks good. This is a much cleaner way to interpret the data over the wavelengths.

hbushouse:
Regression test is running at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/679/

hbushouse:
Interesting that the regression tests are showing very small numerical differences in the results downstream of flat-fielding for NIRSpec MOS and fixed-slit modes. No differences for NIRSpec IFU data.

https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/679/testReport/junit/

melanieclarke:
> Interesting that the regression tests are showing very small numerical differences in the results downstream of flat-fielding for NIRSpec MOS and fixed-slit modes. No differences for NIRSpec IFU data.

I was looking specifically at IFU data because the performance issue most impacts that mode, but the same code is used to make MOS and fixed slit flats.  I suspect the difference is just in precision loss in the different interpolation methods.

PR #7556, opened 2023-04-17
JP-3219: Update ASN rules for NRC SW coronagraphy

hbushouse:
Resolves [JP-3219](https://jira.stsci.edu/browse/JP-3219)

This PR updates various level-2 and level-3 ASN rules to properly handle NIRCam coronagraphy observations that use full-frame readout and hence we get data from all 4 short-wave detectors, even when the occulted target only appears in 1 detector. The desired outcome is the following:

1) Continue to create "image2" ASN's for the detector(s) that have the target in them, using rateints files as input, so that calints are created and fed downstream for regular coronagraphic processing.
2) For the detectors that don't have the target in them, create an "image2" ASN that uses the rate file as input, so that they can go through regular imaging-mode processing downstream.
3) For the detectors with the target, also create another "image2" ASN that uses the rate file as input, so that a normal cal file can be used in downstream processing to treat them like regular images.
4) Construct a "coron3" ASN that uses calints files as input (as usual), but only includes files for the detector(s) that have th e target in them.
5) Construct an "image3" ASN that uses the cal files from *all* detectors, in order to construct a regular imaging mosaic from all the data.

**Checklist for maintainers**
- [x] added relevant label(s)


jwst/associations/lib/dms_base.py
```diff
@@ -753,6 +798,28 @@ def __init__(self, *args, **kwargs):
...
+class Constraint_Coron(Constraint):
+    """Match on Coronagraphic Observations"""
+    def __init__(self, *args, **kwargs):
+        super(Constraint_Coron, self).__init__(
+            [
+                DMSAttrConstraint(
+                    sources=['pntgtype'],
+                    value='science'
+                ),
+                Constraint(
```
stscieisenhamer, line 810: Should not need the wrapping `Constraint`. The DMSAttrConstraint alone should be fine.

The logic is that any value of CORON_EXP_TYPES is good, and whichever is chosen, will be the one used for this association.

jwst/associations/lib/rules_level2b.py
```diff
@@ -52,6 +54,84 @@
 # --------------------------------
 # Start of the User-level rules
 # --------------------------------
+@RegistryMarker.rule
+class Asn_Lv2Coron(
+        AsnMixin_Lv2Coron,
```
stscieisenhamer, line 59: `AsnMixing_Lv2Coron` is a duplicate of `AsnMixin_Image`. One should be able to just use `AsnMixin_image` at this point.

jwst/associations/lib/rules_level2b.py
```diff
@@ -52,6 +54,84 @@
...
+        - Exclude NIRCam SW detectors that don't have occulter
+    """
+
+    def __init__(self, *args, **kwargs):
+
+        # Setup constraints
+        self.constraints = Constraint([
+            Constraint_Base(),
+            Constraint_Mode(),
+            Constraint_Image_Science(),
```
stscieisenhamer, line 80: `Constraint_Image_Science` is to match any Image-based exposure. This rule only want coronagraphic, which looks to be handled by `Constraint_Coron`, below. Should be able to remove this constraint.

stscieisenhamer (changes requested):
Overall, looks good. Some minor changes requested.

Put in slack discussion about the issue with not getting both associations for target detectors.

jwst/associations/lib/dms_base.py
hbushouse: fixed in latest commit

jwst/associations/lib/rules_level2b.py
hbushouse: fixed in latest commit

jwst/associations/lib/rules_level2b.py
hbushouse: fixed in latest commit

hbushouse:
The most recent commit tries to make the decision of whether to use "ints" products in level2b and level3 ASN's more generic. Much of the original code was completely hardwired around the assumption (years ago) that this would only ever be done for TSO data. Now we also have it done for Coron and soon the NIRISS folks want to switch AMI to using "ints" products too. So I made the computation of the `use_integrations` argument in the `make_member` function more generic. That also eliminates the need for custom versions of `make_member` for special cases, like the latest NIRCam coron stuff.

hbushouse:
Latest commit adds the new "Lv3NRCCoronImage" rule, which is specifically designed to create a regular imaging (image3) ASN for this special case of NIRCam SW full-frame coron exposures that includes multiple detectors. Decided it would be cleaner to have a separate, dedicated rule for this special case, rather than trying to make potentially confusing mods to the existing "Lv3Image" rule.

jwst/associations/tests/test_main.py
```diff
@@ -17,12 +17,12 @@
...
 @pytest.mark.parametrize(
     'case', [
-        (None, 62),  # Don't re-run, just compare to the generator fixture results
+        (None, 65),  # Don't re-run, just compare to the generator fixture results
         (['-i', 'o001'], 2),
         (['-i', 'o001', 'o002'], 3),
-        (['-i', 'c1001'], 1),
-        (['-i', 'o001', 'c1001'], 3),
-        (['-i', 'c1001', 'c1002'], 2),
+        (['-i', 'c1001'], 0),
```
hbushouse, line 23: @stscieisenhamer Can't easily tell what exactly these particular tests are checking for, so I'm not sure if the fact that some of them are now returning a zero count is OK. Any thoughts?

hbushouse:
Started regtest run at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/680

hbushouse:
@stscieisenhamer Several regression tests are showing differences (as expected), but I'm having a hard time interpreting the output. Many say "Left has ..." and "Right has ...". What are Left and Right, i.e. which one is the "truth" vs. "new" results? And is there a way to somehow dump more detailed results of the tests to better see exactly what the differences are?

hbushouse:
Running the latest version of the code on the pool file "jw00016_20230331t130733_pool.csv" (from the OTB run) shows we're 98% of the way there. All of the desired "image2" asn's are created, which use both rate and rateints as inputs for the detector with the target and only rate for the others, resulting in all the necessary cal and calints files. And we get the desired "image3" asn that uses "cal" files as input from all detectors to create a regular image3 mosaic. It also gives us the correct "coron3" asn that uses the calints files as input for the science and psf pointings of only the detector with the mask.

However, what we're also getting is another "image3" asn file that uses cal files as input for all detectors, which is correct, but it includes both science **and psf** exposures. We only want the science members, because including the psf members in the image mosaic will cause the mosaic to have a disjointed region of the sky stuck onto it somewhere. This was created by the new `Lv3NRCCoronImage` class, so that may need some tweaking to exclude psf members.

hbushouse:
OK, that was an easy fix to get rid of the psf members from the image3 asn's. Latest code is now giving us all the proper image2, image3, and coron3 asn's, using all the proper rate/rateints/cal/calints inputs.  Huzzah!

jwst/associations/tests/test_main.py
stscieisenhamer: Bad commentary on the test. I'll check to see what is the intent and suggest changes.

stscieisenhamer (commented):
LGTM.

I'll go through the regressions and `test_main` to ensure nothing else horrendous has occurred.

So, for the moment, the official result is only "Comment" until the regression review is complete.

stscieisenhamer:
The regression review shows only expected changes. Explanations are

The following are the expected double associations for both the rate and rateints. This is seen in the test results in the errors of "duplicate product names" and the unaccounted members, one of which has "rate" members and the other has "rateints" members.  The test function will need to be updated to account for the new definition of duplicate. I will handle that and PR against this PR.
- jw88600_20171108T051731_pool
- jw10005_20181020T033546_pool
- jw93025_20171108T062313_pool
- pool_018_all_exptypes

The following are all expected changes to the NIS AMI going from rate to rateints. This is seen from the test results as again unaccounted for members, one having "rate" and the other "rateints". The resolution  for these tests will just be okification.
- jw87600_20180824T213416_pool
- jw00625_20190603t233254_pool
- pool_014_ami_niriss
- pool_018_all_exptypes

`test_main` is basically a canary test. If it changes, but nothing else does, such as the regressions, then something has changed that is not covered by the regressions. However, if the regressions have changes, and they are all known/accepted, as in this case, then just fixing its truths is good.

hbushouse:
Ack, thanks for pointing out the change to NIS_AMI (going from rate to rateints). That made me realize that I had jumped the gun on that one. The AMI folks are in the midst of making changes to the calwebb_ami3 processing to make it take calints as input, rather than the current cal. So I went ahead and added that change into the rules here, because I knew it would be needed in the future. But, as they say, the future is NOT now. I need to temporarily deactivate that one until the Cal code is ready for the switch.

hbushouse:
@stscieisenhamer Latest regtest run is ready for another review at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/693/  There should NOT be any AMI-related changes this time. Coron only.

hbushouse:
I've tested the latest commits on the jw00016 and jw01386 example pools. 

The results for jw00016, which only contains NIRCam coron observations using the short-wave detectors in full-frame readout, give the desired new results, where some of the exposures are used in both rate/cal form and rateints/calints form.

For jw01386, which contains a mixture of NRC coron, MIRI coron, NIS AMI, NRS IFU, and MIRI MRS, all of the non-NIRCam results are the same as before and correct. For the NIRCam coron observations, however, which only use subarray readouts from the NRCA5 (a.k.a. nrcalong) detector, we're getting double image2 asn's created, one using rate input for each exposure and a second using rateints input for each exposure. For this case, where there's only 1 detector readout to being with, which contains the occulted source, there's no need for the duplicate image2 asn's. We only need the regular rateints/calints products, in order to use during normal coron3 processing. I suppose it doesn't hurt to have the duplicate image2 asn's, because the resulting cal products just won't be used at all for any level-3 processing, but it's a waste of time to do the processing for them.

Is there some clever way that the new/updated Lv2 rules can figure out that they only need to create duplicate image2 asn's when exposure type is NRC_CORON and subarray is FULL?  If subarray is not FULL, we don't need the duplicates.

hbushouse:
D'oh! I think the problem is in the new `nrccoron_valid_detector` function in dms_base.py. It's only ever doing the check if SUBARRAY=FULL, hence any "normal" coron observations that use a subarray (i.e. SUBARRAY != FULL) are always returning True, which I believe is then leading to the double image2 asn's. I'll do some experimenting.

hbushouse:
Results of latest regtest run are here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/713/  No failures. Not sure that's actually expected or not?!

jwst/associations/lib/dms_base.py
```diff
@@ -753,6 +852,33 @@ def __init__(self, *args, **kwargs):
         )
 
 
+class Constraint_Coron(Constraint):
```
stscieisenhamer, line 855: I believe this class can be removed. It is not used anywhere.

stscieisenhamer (approved):
There is one code-cleanup thing that could be done. But otherwise, LGTM.

Submit a re-review when ready, regardless of whether the code cleanup occurs or not.

jwst/associations/lib/diff.py
```diff
@@ -344,27 +462,30 @@ def compare_product_membership(left, right):
                     'Left {left_expname}:{left_exptype}'
                     ' != Right {right_expname}:{right_exptype}'
                     ''.format(left_expname=left_member['expname'], left_exptype=left_member['exptype'],
-                          right_expname=right_member['expname'], right_exptype=right_member['exptype'])
+                              right_expname=right_member['expname'], right_exptype=right_member['exptype'])
```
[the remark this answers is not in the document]
hbushouse: No, it looks like that was intentional in order to line up the first character of line 465 with the first character inside the parentheses of line 464 (i.e. line up "left_expname" and "right_expname"). Code style seems OK with it.

Issue #162, opened 2016-08-16
Level 3 processing doesn't pay attention to strun --output_dir param


hbushouse:
Mike Swam reports that when he tried to run wfs_combine level-3 processing using strun and specified the optional --output_dir param to designate a particular output directory in which to place the output product, the output product was still created in the working directory.

I'm guessing this is due to the fact that with level-3 processing, via any task like wfs_combine or calwebb_image3, it's the task/pipeline module itself that's saving the output file, using the name specified in the input ASN table, and is ignoring anything specified via the --output_dir param on the command line. With level-2 processing tasks/pipelines, the output product model is passed back up to stpipe to let it handle the creation of the output file and in that case stpipe does pay attention to any --output_dir that was specified.

Level-3 tasks and pipeline modules, such as wfs_combine, calwebb_image3, calwebb_spec3, etc. should be upgraded to use the output_dir path specified on the strun command line.

stscieisenhamer:
Just as an example of use with the association product names:
```
with open('the_association', 'r') as asn_fh:
    asn = Association.load(asn_fh)
my_result_model = code_returning_some_jwst_datamodels_result_model(asn)
output_dir = '/the/output/directory/path/from--output_dir'
product_type = 'i2d' # Assuming CALIMAGE3
product_name = asn['products'][0]['name']
full_output_path = output_dir + '/' + product_name.format(product_type=product_type)
my_result_model.save(full_output_path)
```

hbushouse:
Regarding line 4 in the above example, if the user specifies the --output_dir param to strun, it will show up as an attribute of the pipeline or step that's being run, as self.output_dir. Hence pipeline modules can check for "self.output_dir is not None" and if so, then be sure to prepend the value of self.output_dir to the output file name.

hbushouse:
Oh, and instead of using the direct approach of appending the path and filename just by using the "+" operator, it might be safer and more OS-benign to use the os.path.join function to do that, i.e. full_output_path = os.path.join(output_dir, product_name).

PR #7559, opened 2023-04-19
JP-3047 Change how NIRSpec wavelengths are set in ifucube

jemorrison:
Resolves  part 2 of [JP-3047](https://jira.stsci.edu/browse/JP-3047) (Homogenize the use of the wavemin/wavemax parameters defining output cubes between MIRI and NIRSpec. ....NIRSpec to use cubepar to define default wavelength ranges.)

Closes #

NOTE an update to stdatamodels (#162) is required to be merged before this PR is merged. Stdatamodels PR 162 updated the NIRSpec cubepar data model to read in 3 wavelength arrays when the drizzle weighting method is used. In order to use this updated stdatamodels the requirement in the jwst package that stdatamodels could be higher than 1.4.0 was removed from
requirements-max.txt

Also **before this PR can be merged** the new NIRSpec cubepars reference file need to delivered to CRDS.

This PR modifies cube_build to use data from the cubepars reference file to define the wavelength ranges of the NIRSpec IFU cubes. Before, the wavelength range was determined from the data.   The NIRSpec team has delivered a new cubepar reference file with  a wavelength array for the PRISM, MEDIUM, and HIGH resolution gratings for the weighting method of drizzle. Currently these wavelength arrays are only needed if data from multiple gratings  (of the same resolution) are combined.  NIRSpec associations are by default only made from 1 grating. But the user can make associations contain multiple gratings and adding the wavelength arrays for the drizzle method to cubepars  allows the code to run if in the future the associations are made from multiple gratings.  

**Checklist for maintainers**
- [ ] updated or added relevant tests - ran unit tests and all passed.
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,11 +1,13 @@
 1.10.3 (unreleased)
 ===================
 
+
 datamodels
 ----------
 
 - Removed use of deprecated ``stdatamodels.jwst.datamodels.DataModel`` class from
   all steps and replaced it with ``stdatamodels.jwst.datamodels.JwstDataModel``. [#7571]
+
 
 documentation
 -------------
@@ -25,6 +27,13 @@ flat_field
 ----------
 
 - Refactored NIRSpec 1D flat interpolation for improved performance. [#7550]
+
+
+cube_build
+----------
+- Wavelength range of NIRSpec IFU cubes are set by cubepars reference file and
+  a new cubepars reference file with wavelength arrays for drizzle implemented. [#7559]
+  
 
 other
 -----
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,7 +32,7 @@ flat_field
 cube_build
 ----------
 - Wavelength range of NIRSpec IFU cubes are set by cubepars reference file and
-  a new cubepars reference file with wavelength arrays for drizzle implemented. [#7559]
+  a new cubepars reference file with wavelength arrays for drizzle is implemented. [#7559]
   
 
 other
--- a/jwst/cube_build/cube_build_io_util.py
+++ b/jwst/cube_build/cube_build_io_util.py
@@ -131,7 +131,6 @@ def read_cubepars(par_filename,
     elif instrument == 'NIRSPEC':
         with datamodels.NirspecIFUCubeParsModel(par_filename) as ptab:
             number_gratings = len(all_grating)
-
             for i in range(number_gratings):
                 this_gwa = all_grating[i]
                 this_filter = all_filter[i]
@@ -142,13 +141,11 @@ def read_cubepars(par_filename,
                     table_spectralstep = tabdata['SPECTRALSTEP']
                     table_wavemin = tabdata['WAVEMIN']
                     table_wavemax = tabdata['WAVEMAX']
-
                     if this_gwa == table_grating and this_filter == table_filter:
                         instrument_info.SetSpatialSize(table_spaxelsize, this_gwa, this_filter)
                         instrument_info.SetSpectralStep(table_spectralstep, this_gwa, this_filter)
                         instrument_info.SetWaveMin(table_wavemin, this_gwa, this_filter)
                         instrument_info.SetWaveMax(table_wavemax, this_gwa, this_filter)
-
                 #  modified Shepard method 1/r weighting
                 if weighting == 'msm':
                     for tabdata in ptab.ifucubepars_msm_table:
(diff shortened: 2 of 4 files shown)


jemorrison:
Running regression tests now. There will be differences in NIRSpec IFU cubes since the wavelength ranges is determined differently.

CHANGES.rst
```diff
@@ -1,6 +1,10 @@
 1.10.3 (unreleased)
 ===================
 
+cube_build
+----------
+- Wavelength range of NIRSpec IFU cubes are set by cubepars reference file #7559
```
hbushouse, line 6: PR number needs to be enclosed in square brackets at the end of the line. Can't be properly parsed this way.

hbushouse: Oh, and regarding the statement itself, I thought the description of the PR said that the wavelength range is not yet specified in the cubepars ref file, so in the meantime this PR will use the wavelength range from the msm/emsm mode. This says it's coming from cubepars. And the sentence is just a statement of fact: it doesn't have a verb that says what's been changed.

jwst/cube_build/cube_build_io_util.py
```diff
@@ -234,6 +231,19 @@ def read_cubepars(par_filename,
                     instrument_info.SetHighEMSMTable(table_wave, table_sroi,
                                                      table_wroi, table_scalerad)
 
+            elif weighting == 'drizzle':
+                # TODO when we get a new cubepars reference file with wavelength
```
hbushouse, line 235: Oh, so the wavelength range is in fact being read from the cubepars ref file, but just from a different section/extension of the ref file. So the description needs to be updated to make it more explicit exactly what's being changed and how.

hbushouse:
Where's the link to the regression test run?

CHANGES.rst
jemorrison: Fixed log entry. Updated description of PR. NIRSpec has already provided me with an updated reference file. Waiting until the jwst pipeline can use stadatamodels >= 1.4.1

DO NOT MERGE PR UNTIL NEW REF FILE IS IN CRDS AND JWST pipeline can use stdatamodels >= 1.4.1

jemorrison:
I originally testing this PR using the NIRSpec reference file in CRDS and hard coding the cube_build code to read in the wavelength arrays for emsm. I thought it would take the NIRSpec team some time to make a new cube pars reference file. Since I originally put this PR together the NIRSpec team sent me a new cubepars reference file.
I have updated stdatamodels to use this new file - but I cannot test it in this PR because I can not get the JWST pipeline to use a newer stdatamodels version. I want to test this file in the JWST pipeline before I tell NIRSpec to put the file on CRDS and to get the stdatamodels PR into the system.

jemorrison:
@hbushouse 
Ok I have completed the work for this. This included an update the stdatamodels to update the NIRSpec cubepar reference file. The new file has 3 more extensions. I believe the order of things are
1. Merge stdatamodels to get new Nirspec cubepar format
2. Deliver new NIRSpec file to CRDS
3. Merge this PR
I don't think it matters the order of 1 and 2. I tested using
1. existing cubepar reference file  in CRDS with new code in stdatamodels with standard processing and it worked fine
2.  new cube pars reference file and the old code  and standard processing and the code ran. 
 
This is because the new wavelength arrays are only used by cube_build when I forced an association to have  to have different gratings (but same resolution) AND --output_type='multi'. For standard processing cube_build will make an ifucube for each grating in the association.

jwst/cube_build/cube_build_io_util.py
```diff
@@ -130,8 +130,8 @@ def read_cubepars(par_filename,
     # Read in NIRSPEC Values
     elif instrument == 'NIRSPEC':
         with datamodels.NirspecIFUCubeParsModel(par_filename) as ptab:
+            print(par_filename)
```
hbushouse, line 133: remove the print statement or change to a log message

jwst/cube_build/ifu_cube.py
```diff
@@ -1339,23 +1340,13 @@ def setup_ifucube_wcs(self):
...
-        # for NIRSPec wavelength range determined from the data
-        if self.instrument == 'NIRSPEC':
-            final_lambda_min = min(lambda_min)
-            final_lambda_max = max(lambda_max)
-            if self.wavemin_user:
-                final_lambda_min = self.wavemin
-
-            if self.wavemax_user:
-                final_lambda_max = self.wavemax
+        log.debug(f'final wave using data:   {min(lambda_min), max(lambda_max)}')
```
hbushouse, line 1343: If the wavelength limits are now determined from the cubepars ref file for both MRS and NRS-IFU, instead of using the data itself, should this log message be updated, so that it doesn't say "using data"?

hbushouse (approved):
Everything looks good now.

hbushouse:
> I just tested running the new code with old cubepars reference file and it runs with the current default association rules - only one grating per association. If you try and run it using more than 1 grating /filter by editing an association and adding files to it so that the IFUcube will have a non-linear wavelength dimension then the code will fail because there is no wavelength array.

OK, having it work OK for default processing, but crash when someone does something unusual seems acceptable (if they're doing something unusual, they should also use a ref file that supports it).

PR #7569, opened 2023-04-26
JP- 3163 nirspec for Point Source data saved in surface brightness 

jemorrison:
Resolves [JP-3163](https://jira.stsci.edu/browse/JP-3163)

Closes #

This PR changes the photom step for NIRSpec IFU point source data to save it in units of surface brightness (MJy/sr). This change was made because cube_build was designed to work in surface brightness units. The extract_1d step for NIRSpec IFU point source data also had to be updated to convert the extracted data from surface brightness units to flux units.

**Checklist for maintainers**
- [x] added relevant label(s)
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/704/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,10 +15,10 @@ documentation
 - Update tutorial usage of ``jump.threshold`` to ``jump.rejection_threshold``. [#7572]
 
 
-extract_1d:
------------
-For NIRSpec IFU point source data remove options assuming data is in flux units. Treat these data
-the same as all other data and convert extracted values from surface brightness to flux. [#7569]
+extract_1d
+----------
+-For NIRSpec IFU point source data remove options assuming data is in flux units. Treat these data
+ the same as all other data and convert extracted values from surface brightness to flux. [#7569]
 
 
 flat_field
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,7 +12,14 @@ documentation
 
 - Fix bugs in implementation of ``pixel_replace`` documentation. [#7565]
 
+
 - Update tutorial usage of ``jump.threshold`` to ``jump.rejection_threshold``. [#7572]
+
+extract_1d:
+-----------
+For NIRSpec IFU point source data remove options assuming data is in flux units. Treat these data
+the same as all other data and convert extracted values from surface brightness to flux. [#7569]
+
 
 flat_field
 ----------
@@ -34,6 +41,10 @@ pathloss
 
 - Bug fix for NIRSpec fixed-slit data to remove double application of correction
   factors. [#7566]
+
+photom
+------
+- For NIRSpec IFU point source data save data in units of surface brightness [#7569]
 
 ramp_fitting
 ------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,8 +12,8 @@ documentation
 
 - Fix bugs in implementation of ``pixel_replace`` documentation. [#7565]
 
-
 - Update tutorial usage of ``jump.threshold`` to ``jump.rejection_threshold``. [#7572]
+
 
 extract_1d:
 -----------
--- a/docs/jwst/cube_build/main.rst
+++ b/docs/jwst/cube_build/main.rst
@@ -28,11 +28,11 @@ Assumptions
 It is assumed that the ``assign_wcs`` step has been applied to the data, attaching the distortion and pointing
 information to the image(s). It is also assumed that the ``photom`` step has been applied to convert the pixel
 values from units of count rate to surface brightness. This step will only work with MIRI or NIRSpec IFU data.
-The cube_build algorithm is a flux conservation method and requires the input data to be in unit of surface brightness.
-The NIRSpec calibration plan for point source data is designed to produce units of flux from the calwebb_spec2 pipeline.
-For NIRSpec IFU point source data the calwebb_spec2 pipeline divide the flux values by a pixel area map to produce pseudo
+The cube_build algorithm is a flux conserving method and requires the input data to be in units of surface brightness.
+The NIRSpec calibration plan for point source data is designed to produce units of flux density from the calwebb_spec2 pipeline.
+For NIRSpec IFU point source data the calwebb_spec2 pipeline divides the flux values by a pixel area map to produce pseudo
 surface brightness units (MJy/steradian). This allows the cube_build program to conserve flux when it combines and resamples
-the data. True fluxes are produced only at the extract_1d step, in which a 1D spectrum is extracted from the cube using an
+the data. True fluxes are produced only at the :ref:`extract_1d_step <extract_1d_step>`, in which a 1D spectrum is extracted from the cube using an
 appropriate extraction aperture, with resulting units of Jy.
 
 Instrument Information
--- a/docs/jwst/cube_build/main.rst
+++ b/docs/jwst/cube_build/main.rst
@@ -28,6 +28,12 @@ Assumptions
 It is assumed that the ``assign_wcs`` step has been applied to the data, attaching the distortion and pointing
 information to the image(s). It is also assumed that the ``photom`` step has been applied to convert the pixel
 values from units of count rate to surface brightness. This step will only work with MIRI or NIRSpec IFU data.
+The cube_build algorithm is a flux conservation method and requires the input data to be in unit of surface brightness.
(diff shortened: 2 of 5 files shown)


jemorrison:
Need to update documentation to reflect that NIRSpec IFU point source data is in units of surface brightness after photom step.

jemorrison:
running regression tests

jwst/photom/photom.py
```diff
@@ -328,11 +324,9 @@ def calc_nirspec(self, ftab, area_fname):
                 # on its wavelength
                 sens2d = np.interp(wave2d, waves, relresps)
                 sens2d *= tabdata['photmj']  # include the initial scalar conversion factor -> MJ
-                # This line used to be applied to all IFU data, but I think this was an error -
-                # masked by the fact that the initial pixel area maps for IFU data were
-                # arrays of 1-values (in arcsec**2). If srctype==POINT, do not apply.
-                if self.source_type.upper() != 'POINT':
-                    sens2d /= area2d * A2_TO_SR  # divide by pixel area * A2_TO_SR -> MJ/sr
+                # Apply to all data (both point source and extended)
```
hbushouse, line 327: Instead of just saying "Apply", which leaves open the question of "apply what and why?" make this comment more explicit, e.g. "Convert all data (both point source and extended) to surface brightness"

jwst/photom/photom.py
```diff
@@ -328,11 +324,9 @@ def calc_nirspec(self, ftab, area_fname):
...
+                sens2d /= area2d * A2_TO_SR  # divide by pixel area * A2_TO_SR -> MJ/sr
```
hbushouse, line 328: minor typo at the end of the comment string: "MJ" should be "MJy"

hbushouse:
Please provide a link to the regression test run.

jmuzerolle:
Just a quick comment - Leonardo Ubeda will be testing the new code early next week.
As to the question of units, we will have to be clear in the documentation that the surface brightness of the point source cube is not strictly correct because of the way the pathloss correction is applied. As is/was the case for the flux density units, you only get the correct answer once you perform the aperture extraction to 1D. If users want "true" surface brightness, they would have to rerun stage 2 in its entirety with srctype=extended. This units change is just a convenience to ensure that the cube_build resampling as it's currently constructed conserves flux for point sources. (Ok, not so quick a comment)

jwst/extract_1d/ifu.py
```diff
@@ -80,10 +80,6 @@ def ifu_extract1d(input_model, ref_dict, source_type, subtract_background,
     else:
         log.info(f"Source type = {source_type}")
 
-    # The input units will normally be MJy / sr, but for NIRSpec point-source
-    # spectra the units will be MJy.
-    input_units_are_megajanskys = (source_type == 'POINT' and instrument == 'NIRSPEC')
```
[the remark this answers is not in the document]
jemorrison: input_units_are_megajanskys was only set for NIRSPec POINT SOURCE data. It is not used in the module any more

jmuzerolle:
@jemorrison No, the pathloss should be left as-is. That correction is meant to account for diffraction losses as a function of source position in an IFU slice, and is different from the correction for extended sources. Based on analysis done so far, the effect is probably at the few percent level, so may not be significant for most use cases. But mathematically, this gives the more correct result for point sources.
I think a clarification in the cube_build documentation is a good idea, and we will also update our relevant JDox pages (probably the pipeline caveats page, to start).

hbushouse:
The regtest results at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/704/testReport/ seem consistent with what we'd expect from these changes:

1) The output from the photom step now has values that are much bigger, because it's now MJy/sr instead of MJy.
2) The values of "background" in x1d files from extract_1d are now much smaller, because they are no longer being divided by a very small number (hence old results were much bigger).

So I think this all looks good.

hbushouse (approved):
Code changes, CI tests, and regression test results all look good, so I'm approving. Will wait to merge in case you want to make more updates to the docs (e.g. cube_build, to warn users about "faux" surface brightness units).

jemorrison:
I do what to add something to cube_build docs. I am working on the language I just do not want to confuse the user to what the units are

jemorrison:
@jmuzerolle or @hbushouse (struggling a bit to add information on point source NIRSpec data being in "surface brightness" and not confuse the user". Here is my stab at what I will add to Output Cube Format section. Comments ? 

The cube_build algorithm is a flux conservation method and requires the input data to be in unit of surface brightness. The NIRSpec calibration plan for point source data is designed to produce units of flux from the calwebb_spec2 pipeline. For NIRSpec IFU point source data the calwebb_spec2 pipeline divide the flux values by a pixel area map to produce pseudo surface brightness units (MJy/steradian). This allows the cube_build program to conserve flux when it combines and resamples the data. The extract_1d step accounts for this pseudo surface brightness and produces data in true flux units (MJy).

jmuzerolle:
@jemorrison For the last sentence, I would say "True fluxes are produced only at the extract_1d step, in which a 1D spectrum is extracted from the cube using an appropriate extraction aperture, with resulting units of Jy." Everything else looks fine.

docs/jwst/cube_build/main.rst
```diff
@@ -28,6 +28,12 @@ Assumptions
 It is assumed that the ``assign_wcs`` step has been applied to the data, attaching the distortion and pointing
 information to the image(s). It is also assumed that the ``photom`` step has been applied to convert the pixel
 values from units of count rate to surface brightness. This step will only work with MIRI or NIRSpec IFU data.
+The cube_build algorithm is a flux conservation method and requires the input data to be in unit of surface brightness.
```
hbushouse, line 31: I think I would say "... is a flux *conserving* method ...".  And towards the end of the line "unit" should be "units".

CHANGES.rst
```diff
@@ -14,6 +14,13 @@ documentation
 
 - Update tutorial usage of ``jump.threshold`` to ``jump.rejection_threshold``. [#7572]
 
+
+extract_1d:
```
hbushouse, line 18: This entry is not formatted correctly. Get rid of the colon at the end of the step name and indent the following text lines, preceded by a dash "-"  (as with all other log entries)

docs/jwst/cube_build/main.rst
```diff
@@ -28,6 +28,12 @@ Assumptions
...
+The NIRSpec calibration plan for point source data is designed to produce units of flux from the calwebb_spec2 pipeline.
+For NIRSpec IFU point source data the calwebb_spec2 pipeline divide the flux values by a pixel area map to produce pseudo
```
hbushouse, line 33: "divide" -> "divides"

docs/jwst/cube_build/main.rst
```diff
@@ -28,6 +28,12 @@ Assumptions
...
+surface brightness units (MJy/steradian). This allows the cube_build program to conserve flux when it combines and resamples
+the data. True fluxes are produced only at the extract_1d step, in which a 1D spectrum is extracted from the cube using an
```
hbushouse, line 35: Could make the "extract_1d step" a link to that step's docs.

CHANGES.rst
```diff
@@ -14,6 +14,13 @@ documentation
 
 - Update tutorial usage of ``jump.threshold`` to ``jump.rejection_threshold``. [#7572]
 
+
+extract_1d
+----------
+-For NIRSpec IFU point source data remove options assuming data is in flux units. Treat these data
```
hbushouse, line 20: Need a whitespace between the dash and the first character of text. So the 2nd line needs to be intended one more space too.

hbushouse:
@jmuzerolle Are you good with the latest changes?  If so, we can merge this.

jmuzerolle:
@hbushouse The documentation looks good. Leonardo is testing the PR now, I don't know if you want to wait for those results, or merge first?

leonardoubeda:
@jemorrison and @jmuzerolle  I tested this fix on the standard solar-type star P330e from programme 01538. I reprocessed the photom step and created a new data cube. I did not pay attention to outlier detection to speed up the process. I see a huge improvement in the flux calibration. The attached plot shows the flux comparison in blue (pipeline 1.9.4) and orange (this JP-3163 code). The red spectrum is the CALSPEC model for this source for reference.

flux_comparison_p330e

hbushouse:
Thanks @leonardoubeda for testing and confirmation. I'm going to merge the updates now.

PR #7573, opened 2023-04-27
JP-3227: Add analog of HST's tweakback tool

mcara:
Resolves [JP-3227](https://jira.stsci.edu/browse/JP-3227)

This PR adds functionality similar to HST's `tweakback` task to be able to transfer an alignment solution from a file to another.

**Checklist for maintainers**
- [x] added relevant label(s)
- [ ] ran regression tests, post a link to the Jenkins job below. - **no need to run regression tests (new feature)**

### DIFF
--- a/jwst/tweakreg/utils.py
+++ b/jwst/tweakreg/utils.py
@@ -1,10 +1,56 @@
 from copy import deepcopy
 
+from astropy.modeling.rotations import RotationSequence3D
 from gwcs.wcs import WCS
+import numpy as np
 from tweakwcs.correctors import JWSTWCSCorrector
 from tweakwcs.linearfit import build_fit_matrix
 
+from ..datamodels import ImageModel
+from ..assign_wcs.util import update_fits_wcsinfo
 from ..assign_wcs.pointing import _v23tosky
+
+
+_RAD2ARCSEC = 3600.0 * np.rad2deg(1.0)
+
+
+def _wcsinfo_from_wcs_transform(wcs):
+    frms = wcs.available_frames
+    if 'v2v3' not in frms or 'world' not in frms or frms[-1] != 'world':
+        raise ValueError(
+            "Unsupported WCS structure."
+        )
+
+    # Initially get v2_ref, v3_ref, and roll_ref from
+    # the v2v3 to world transform. Also get ra_ref, dec_ref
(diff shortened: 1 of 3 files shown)


jwst/tweakreg/utils.py
```diff
@@ -84,3 +130,145 @@ def adjust_wcs(wcs, delta_ra=0.0, delta_dec=0.0, delta_roll=0.0,
...
+        raise ValueError(
+            "Both 'matrix' and 'shift' must be either None or not None."
+        )
+
+    elif matrix is None and shift is None:
+        # use the correction that was applied to the "reference" WCS - the
+        # WCS from the "from_image":
+        if 'v2v3corr' not in refwcs.available_frames:
+            raise ValueError(
+                "The WCS of the 'from_image' data model does not has not been "
```
hbushouse, line 247: minor typo here: "... does not has not been", which I guess should be "has not been"

jwst/tweakreg/utils.py
```diff
@@ -84,3 +130,145 @@ def adjust_wcs(wcs, delta_ra=0.0, delta_dec=0.0, delta_roll=0.0,
...
+                "previously corrected. A corrected WCS is need in order to "
```
hbushouse, line 248: minor typo: "need" should be "needed"

hbushouse (commented):
Other than a couple of minor typos it looks OK to me.

One general question: it looks like this will just be available as a function that can be called from the `tweakreg.utils` module, right?  It's not a new standalone step in the jwst package.

jwst/tweakreg/utils.py
mcara: Sigh

jwst/tweakreg/utils.py
mcara: > One general question: it looks like this will just be available as a function that can be called from the `tweakreg.utils` module, right? It's not a new standalone step in the jwst package.

That is correct. It is a just a tool for post-processing/custom processing. It is not a step. Could be made into a script too, if there is a strong interest.

hbushouse (approved):
LGTM

@nden any comments?

nden:
The imports don't look good to me. I'll create a PR to fix them.

PR #7321, opened 2022-10-27
JP-2931: Guard against nearly identical points in nirspec resampled WCS

mcara:
Resolves [JP-2931](https://jira.stsci.edu/browse/JP-2931)

Closes #7293

This PR fixes a crash observed in some datasets caused by the presence of almost the same points in the list of wavelengths in output spectral WCS.

**Checklist for maintainers**
- [x] added entry in `CHANGES.rst` within the relevant release section
- [ ] updated or added relevant tests
- [ ] updated relevant documentation
- [x] added relevant milestone
- [x] added relevant label(s)
- [x] ran regression tests, post a link to the Jenkins job below.
      [How to run regression tests on a PR](https://github.com/spacetelescope/jwst/wiki/Running-Regression-Tests-Against-PR-Branches)
- [ ] Make sure the JIRA ticket is [resolved properly](https://github.com/spacetelescope/jwst/wiki/How-to-resolve-JIRA-issues)

Regression tests *are still running*: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/450/

PR #7579, opened 2023-05-01
JP-2858: rework resample_spec for MIRI LRS

nden:
Resolves [JP-2858](https://jira.stsci.edu/browse/JP-2858)

Closes #

This PR reworks the computation of the spatial component of the output WCS in resample_spec for MIRI LRS. It reuses the function that computes the output WCS for imaging mode and replaces the axis on the detector corresponding to the spectral axis with what is computed by the code specific to wavelength computation, to determine the bounding_box of the new output WCS. The wavelength axis computation is kept the same as before, i.e. a table of wavelengths is constructed and interpolated during evaluation.

I think the main issue solved by this method is taking into account the roll angle of the telescope during the observation. The method computes a point on the detector where the fiducial on the sky is projected using the inverse transform of the WCS. This requires passing a wavelength to the `assign_wcs.util.compute_scale`  method (I used the mean of the wavelengths) and changing the signature of the method to accept a wavelength. This method is used for imaging and spectral observations. For imaging mode it is a transparent change - the new `wavelength` parameter is ignored. When a spectral observation is detected, a `wavelength` parameter is required and this is what the change in `ifu.py` for Nirspec is.

This method is implemented as a new function specific to `MIRI-LRS-FIXEDSLIT` exposure type. If testing proves it works better than the current methods we should consider expanding it to other exposure types.

The regression tests were run in https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/709. They show errors for Nirspec which I think are bugs - essentially calling `interpolate` too strictly, without allowing for extrapolation or providing a `fill_value`. These are caused by the change in `compute_scale` and need to be tracked down.

This method was tested only with the data in the JP-2858 ticket. It needs to be tested with the Spec3Pipeline.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -328,6 +328,8 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
         if sky_axes.any():
             if spec.any():
                 dispaxis = refmodel.meta.wcsinfo.dispersion_direction
+            else:
+                dispaxis = None
             if not pscale:
                 pscale = compute_scale(refmodel.meta.wcs, ref_fiducial,
                                        pscale_ratio=pscale_ratio, disp_axis=dispaxis,
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -17,6 +17,7 @@ from typing import Union, List
 from gwcs import WCS
 from gwcs.wcstools import wcs_from_fiducial, grid_from_bounding_box
 from gwcs import utils as gwutils
+from gwcs import coordinate_frames as cf
 from stpipe.exceptions import StpipeExitException
 
 from stdatamodels.jwst.datamodels import JwstDataModel
@@ -136,7 +137,6 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray],
         crpix = np.array(wcs.invert(fiducial[0], fiducial[1], wavelength))
     else:
         crpix = np.array(wcs.invert(*fiducial))
-    crpix = np.array(wcs.invert(*fiducial))
 
     delta = np.zeros_like(crpix)
     spatial_idx = np.where(np.array(wcs.output_frame.axes_type) == 'SPATIAL')[0]
@@ -325,7 +325,7 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
         rotation = astmodels.AffineTransformation2D(pc, name='pc_rotation_matrix')
         transform.append(rotation)
 
-        if sky_axes:
+        if sky_axes.any():
             if spec.any():
                 dispaxis = refmodel.meta.wcsinfo.dispersion_direction
             if not pscale:
@@ -338,7 +338,7 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
             transform = functools.reduce(lambda x, y: x | y, transform)
 
     if isinstance(refmodel.meta.wcs.output_frame, cf.CompositeFrame):
-        out_frame.frames[0]
+        out_frame = refmodel.meta.wcs.output_frame.frames[0]
     else:
         out_frame = refmodel.meta.wcs.output_frame
     input_frame = refmodel.meta.wcs.input_frame
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -101,7 +101,8 @@ def reproject(wcs1, wcs2):
 
 
 def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray],
-                  disp_axis: int = None, pscale_ratio: float = None) -> float:
+                  disp_axis: int = None, pscale_ratio: float = None,
+                  wavelength: float=None) -> float:
     """Compute scaling transform.
 
     Parameters
@@ -118,6 +119,9 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray],
     pscale_ratio : int
         Ratio of input to output pixel scale
 
+    wavelegnth : float
+        Central wavelength. Used in evaluating the inverse WCS transform for spectral data.
+
     Returns
     -------
     scale : float
@@ -128,7 +132,10 @@ def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray],
 
     if spectral and disp_axis is None:
         raise ValueError('If input WCS is spectral, a disp_axis must be given')
-
+    if spectral:
+        crpix = np.array(wcs.invert(fiducial[0], fiducial[1], wavelength))
+    else:
+        crpix = np.array(wcs.invert(*fiducial))
     crpix = np.array(wcs.invert(*fiducial))
 
     delta = np.zeros_like(crpix)
@@ -199,7 +206,7 @@ def calc_rotation_matrix(roll_ref: float, v3i_yang: float, vparity: int = 1) -> 
 
 def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=None,
                         pscale_ratio=None, pscale=None, rotation=None,
-                        shape=None, crpix=None, crval=None):
+                        shape=None, crpix=None, crval=None, wavelength=None):
     """
     Create a WCS from a list of input data models.
 
@@ -256,6 +263,10 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
     crval : tuple of float, None, optional
         Right ascension and declination of the reference pixel. Automatically
         computed if not provided.
+    wavelegnth : float
+         Central wavelength or other appropriate wavelength value to pass to
+         the inverse transform when computing the reference point on the detector,
+         as well as the scale.
 
     """
     bb = bounding_box
@@ -271,9 +282,9 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
         refmodel = dmodels[0]
     else:
         if not isinstance(refmodel, JwstDataModel):
-            raise TypeError("Expected refmodel to be an instance of DataModel.")
+            raise TypeError("Expected refmodel {refmodel} to be an instance of JwstDataModel.")
 
-    fiducial = compute_fiducial(wcslist, bb)
+    fiducial = compute_fiducial(wcslist, bb)[:2] # spatial only
     if crval is not None:
         # overwrite spatial axes with user-provided CRVAL:
         i = 0
@@ -289,7 +300,12 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
     if transform is None:
         transform = []
         wcsinfo = pointing.wcsinfo_from_model(refmodel)
-        sky_axes, spec, other = gwutils.get_axes(wcsinfo)
+
+        axes_type = np.array(refmodel.meta.wcs.output_frame.axes_type)
+        sky_axes = np.array(refmodel.meta.wcs.output_frame.axes_order)[axes_type == "SPATIAL"]
+        spec = np.array(refmodel.meta.wcs.output_frame.axes_order)[axes_type == "SPECTRAL"]
+        if spec and wavelength is None:
+            raise ValueError("wavelelngth is a required parameter for spectral data.")
 
         # Need to put the rotation matrix (List[float, float, float, float])
         # returned from calc_rotation_matrix into the correct shape for
@@ -310,26 +326,32 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
         transform.append(rotation)
 
         if sky_axes:
+            if spec.any():
+                dispaxis = refmodel.meta.wcsinfo.dispersion_direction
             if not pscale:
                 pscale = compute_scale(refmodel.meta.wcs, ref_fiducial,
-                                       pscale_ratio=pscale_ratio)
+                                       pscale_ratio=pscale_ratio, disp_axis=dispaxis,
+                                       wavelength=wavelength)
             transform.append(astmodels.Scale(pscale, name='cdelt1') & astmodels.Scale(pscale, name='cdelt2'))
 
         if transform:
             transform = functools.reduce(lambda x, y: x | y, transform)
 
-    out_frame = refmodel.meta.wcs.output_frame
+    if isinstance(refmodel.meta.wcs.output_frame, cf.CompositeFrame):
+        out_frame.frames[0]
+    else:
+        out_frame = refmodel.meta.wcs.output_frame
     input_frame = refmodel.meta.wcs.input_frame
     wnew = wcs_from_fiducial(fiducial, coordinate_frame=out_frame, projection=prj,
                              transform=transform, input_frame=input_frame)
 
-    footprints = [w.footprint().T for w in wcslist]
+    footprints = [w.footprint().T[:2] for w in wcslist]
     domain_bounds = np.hstack([wnew.backward_transform(*f) for f in footprints])
     axis_min_values = np.min(domain_bounds, axis=1)
     domain_bounds = (domain_bounds.T - axis_min_values).T
 
     output_bounding_box = []
-    for axis in out_frame.axes_order:
+    for axis in input_frame.axes_order:
         axis_min, axis_max = domain_bounds[axis].min(), domain_bounds[axis].max()
         output_bounding_box.append((axis_min, axis_max))
 
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -82,7 +82,7 @@ class ResampleSpecData(ResampleData):
             if resample_utils.is_sky_like(
                 self.input_models[0].meta.wcs.output_frame
             ):
-                if self.input_models[0].metaexposure.type == "MIR_LRS-FIXEDSLIT":
+                if self.input_models[0].meta.exposure.type == "MIR_LRS-FIXEDSLIT":
                     self.output_wcs = self.build_miri_lrs_output()
                 elif self.input_models[0].meta.instrument.name != "NIRSPEC":
                     self.output_wcs = self.build_interpolated_output_wcs()
@@ -315,7 +315,7 @@ class ResampleSpecData(ResampleData):
 
         return y_slit_min, y_slit_max
 
-    def build_miri_lrs_output():
+    def build_miri_lrs_output(self):
         """ Build MIRI LRS FIXEDSLIT output WCS. """
         all_wavelength = []
         all_ra_slit = []
--- a/jwst/resample/resample_spec.py
+++ b/jwst/resample/resample_spec.py
@@ -17,7 +17,7 @@ from stdatamodels.jwst import datamodels
 
 from jwst.datamodels import ModelContainer
 
-from ..assign_wcs.util import wrap_ra
+from ..assign_wcs.util import wrap_ra, wcs_from_footprints
 from . import resample_utils
 from .resample import ResampleData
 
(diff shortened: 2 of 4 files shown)


nden:
A comparison of the extracted spectra from the resampled and unresampled data for the two nods in JP-2858:


miri1



miri2

jwst/assign_wcs/util.py
```diff
@@ -101,7 +102,8 @@ def _reproject(x, y):
 
 
 def compute_scale(wcs: WCS, fiducial: Union[tuple, np.ndarray],
-                  disp_axis: int = None, pscale_ratio: float = None) -> float:
+                  disp_axis: int = None, pscale_ratio: float = None,
+                  wavelength: float=None) -> float:
```
hbushouse, line 106: code style might complain about the lack of whitespace around the "=" here?

jwst/resample/resample_spec.py
```diff
@@ -313,6 +315,91 @@ def _max_virtual_slit_extent(self, wcs_list, target_ra, target_dec):
...
+
+        # done looping over set of models
+        all_wave = np.hstack(all_wavelength)
+        all_wave = all_wave[~np.isnan(all_wave)]
+        all_wave = np.sort(all_wave, axis=None)
+        # Tabular interpolation model, pixels -> lambda
+        wavelength_array = np.unique(all_wave)
+        # Check if the data is MIRI LRS FIXED Slit. If it is then
+        # the wavelength array needs to be flipped so that the resampled
+        # dispersion direction matches the dispersion direction on the detector.
```
hbushouse, line 361: Isn't the "upside-down" dispersion direction for LRS FS already encoded by setting dispaxis=-2 (i.e. *negative* 2) in the meta data? Doesn't that somehow automatically get accounted for when constructing the spectral WCS axis and associated wavelength array?

hbushouse (commented):
Will need a change log entry.

hbushouse:
Does this new approach somehow ignore or workaround the assertions by @mcara that there are problems in the WCS structure/content of the *unrectified* LRS fixed-slit images, which then translated into problems with the resample WCS too? In other words, are there still unresolved problems in the original WCS?

nden:
I don't think we can improve the WCS of the unrectified spectra. Unlike imaging mode all the light on the detector comes from one source on the sky, and the unrectified transform is tied to the zero-point of the calibration in the reference file.

jwst/resample/resample_spec.py
nden: This is code related to the wavelength axis computation which I kept from the previous implementation. For this data 
```
cal.meta.wcsinfo.dispersion_direction
Out[43]: 2
```

jwst/assign_wcs/util.py
nden: yeah, I'll fix those

jwst/resample/resample_spec.py
hbushouse: Huh, you're right. In all the post-assign_wcs products for LRS FS it says "DISPAXIS=2". I could've sworn we used signed integers for that keyword to document the *direction* of increasing wavelength, along with the axis. But I guess not.

jwst/resample/resample_spec.py
```diff
@@ -313,6 +315,91 @@ def _max_virtual_slit_extent(self, wcs_list, target_ra, target_dec):
...
+            warnings.resetwarnings()
+            wavelength_array = wavelength_array[~np.isnan(wavelength_array)]
+
+            all_wavelength = np.append(all_wavelength, wavelength_array)
+            this_minw = np.min(wavelength_array)
+            this_maxw = np.max(wavelength_array)
+            all_minw = np.min(all_wavelength)
+            all_maxw = np.max(all_wavelength)
+            if this_minw < all_minw:
+                addpts = wavelength_array[wavelength_array < all_minw]
```
mcara, line 347: I would add some `delta` to this inequality. For example,

```python
deltas = np.ediff1d(wavelength_array)
delta_min = deltas[:5].mean()
delta_max = deltas[-5:].mean()
if this_minw < all_minw - 0.3 * delta_min:
    addpts = wavelength_array[wavelength_array < (all_minw - 0.3 * delta_min)]
.......
if this_maxw > all_maxw + 0.3 * delta_max:
    addpts = wavelength_array[wavelength_array > (all_maxw + 0.3 * delta_max)]
```

This helps avoid situations in which `this_minw` is barely smaller than `all_minw`. I have encountered this in the past, see https://github.com/spacetelescope/jwst/pull/7321

jwst/resample/resample_spec.py
```diff
@@ -313,6 +315,91 @@ def _max_virtual_slit_extent(self, wcs_list, target_ra, target_dec):
...
+                all_wavelength = np.append(all_wavelength, addpts)
```
mcara, line 348: If you want to keep the the wavelengths sorted, this should be modified:
```suggestion
                all_wavelength = np.append(addpts , all_wavelength)
```
**NOTE:** The above suggestion assumes wavelengths in `wavelength_array` are always increasing.

jwst/resample/resample_spec.py
```diff
@@ -313,6 +315,91 @@ def _max_virtual_slit_extent(self, wcs_list, target_ra, target_dec):
...
+            if this_maxw > all_maxw:
+                addpts = wavelength_array[wavelength_array > all_maxw]
+                all_wavelength = np.append(all_wavelength, addpts)
+
+        # done looping over set of models
+        all_wave = np.hstack(all_wavelength)
+        all_wave = all_wave[~np.isnan(all_wave)]
+        all_wave = np.sort(all_wave, axis=None)
```
mcara, line 356: sorting wouldn't be necessary if `all_wavelength` is kept sorted as suggested above.

jwst/resample/resample_spec.py
```diff
@@ -313,6 +315,91 @@ def _max_virtual_slit_extent(self, wcs_list, target_ra, target_dec):
...
+        # Tabular interpolation model, pixels -> lambda
+        wavelength_array = np.unique(all_wave)
```
mcara, line 358: Is this necessary?

mcara (commented):
This is just first pass through the code. I will try now to understand more of the big picture.

jwst/resample/resample_spec.py
```diff
@@ -313,6 +315,91 @@ def _max_virtual_slit_extent(self, wcs_list, target_ra, target_dec):
...
+            warnings.simplefilter("ignore")
+            wavelength_array = np.nanmedian(lam, axis=spectral_axis)
+            warnings.resetwarnings()
+            wavelength_array = wavelength_array[~np.isnan(wavelength_array)]
+
+            all_wavelength = np.append(all_wavelength, wavelength_array)
+            this_minw = np.min(wavelength_array)
+            this_maxw = np.max(wavelength_array)
+            all_minw = np.min(all_wavelength)
+            all_maxw = np.max(all_wavelength)
```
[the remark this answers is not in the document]
nden: I reused the previous code for the wavelength computation but I'll review it now and address your comments.

mcara:
> Does this new approach somehow ignore or workaround the assertions by @mcara that there are problems in the WCS structure/content of the _unrectified_ LRS fixed-slit images, which then translated into problems with the resampled WCS too? In other words, are there still unresolved problems in the original WCS?

@hbushouse @nden Here is another example that illustrates what I believe is a problem with input (CAL) images' WCS:

```python
import numpy as np

from jwst.datamodels import ImageModel
from jwst.assign_wcs.util import wcs_bbox_from_shape

image_model = ImageModel('jw02072001001_06101_00001_mirimage_cal.fits')

targ_ra = image_model.meta.target.ra
targ_dec = image_model.meta.target.dec

deg2pix = (3600 / 0.11)

bb = wcs_bbox_from_shape(image_model.data.shape)
gy, gx = grid_from_bounding_box(bb)
gx = gx.ravel()
gy = gy.ravel()

ra, dec, lam = image_model.meta.wcs(gx.ravel(), gy.ravel())
mask = np.isfinite(ra) + np.isfinite(dec) + np.isfinite(lam)

dra = np.abs(ra[mask] - targ_ra) * deg2pix
ddec = np.abs(dec[mask] - targ_dec) * deg2pix
gx = gx[mask]
gy = gy[mask]

euclid = np.sqrt(dra**2 + ddec**2)

idx = np.argsort(euclid)

print(euclid[idx])
```
Running this I get:
```
[ 3.35653571  3.35653607  3.3565382  ... 33.57244034 33.5780937
 33.58374708]
```
How is this even possible? How can every pixel in the image be at least 3.5 (equivalent) pixels away from the target on the sky? Am I doing something wrong? Am I not understanding how this particular WCS works?

For an imaging WCS I would be very worried if CRPIX is at the center of the image and every pixel's sky coordinates were 4 pixels (equivalent on the sky) away from CRVAL. Either I am making a mistake or CAL WCS is wrong.

tapastro:
> > Does this new approach somehow ignore or workaround the assertions by @mcara that there are problems in the WCS structure/content of the _unrectified_ LRS fixed-slit images, which then translated into problems with the resampled WCS too? In other words, are there still unresolved problems in the original WCS?
> 
> @hbushouse @nden Here is another example that illustrates what I believe is a problem with input (CAL) images' WCS:
> 
> ```python
> import numpy as np
> from gwcs.wcstools import grid_from_bounding_box
> 
> from jwst.datamodels import ImageModel
> from jwst.assign_wcs.util import wcs_bbox_from_shape
> 
> image_model = ImageModel('jw02072001001_06101_00001_mirimage_cal.fits')
> 
> targ_ra = image_model.meta.target.ra
> targ_dec = image_model.meta.target.dec
> 
> deg2pix = (3600 / 0.11)
> 
> bb = wcs_bbox_from_shape(image_model.data.shape)
> gy, gx = grid_from_bounding_box(bb)
> gx = gx.ravel()
> gy = gy.ravel()
> 
> ra, dec, lam = image_model.meta.wcs(gx, gy)
> mask = np.isfinite(ra) + np.isfinite(dec) + np.isfinite(lam)
> 
> dra = np.abs(ra[mask] - targ_ra) * deg2pix
> ddec = np.abs(dec[mask] - targ_dec) * deg2pix
> gx = gx[mask]
> gy = gy[mask]
> 
> euclid = np.sqrt(dra**2 + ddec**2)
> 
> idx = np.argsort(euclid)
> 
> print(euclid[idx])
> ```
> 
> Running this I get:
> 
> ```
> [ 3.35653571  3.35653607  3.3565382  ... 33.57244034 33.5780937
>  33.58374708]
> ```
> 
> How is this even possible? How can every pixel in the image be at least 3.5 (equivalent) pixels away from the target on the sky? Am I doing something wrong? Am I not understanding how this particular WCS works?
> 
> For an imaging WCS I would be very worried if CRPIX is at the center of the image and every pixel's sky coordinates were 4 pixels (equivalent on the sky) away from CRVAL. Either I am making a mistake or CAL WCS is wrong.

That program is observing target of opportunity supernovae, so it may be possible that the target RA/Dec were placeholders (or at least poorly described pending target acquisition)? Seems possible at least that the Target coords are bad rather than (or at least in addition to) the WCS.

mcara:
@tapastro Thanks! Useful info. I will continue on JIRA

jwst/assign_wcs/util.py
```diff
@@ -271,9 +282,9 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
         refmodel = dmodels[0]
     else:
         if not isinstance(refmodel, JwstDataModel):
-            raise TypeError("Expected refmodel to be an instance of DataModel.")
+            raise TypeError("Expected refmodel {refmodel} to be an instance of JwstDataModel.")
```
mcara, line 285:
```suggestion
            raise TypeError(f"Expected refmodel {refmodel} to be an instance of JwstDataModel.")
```

jwst/assign_wcs/util.py
```diff
@@ -271,9 +282,9 @@ def wcs_from_footprints(dmodels, refmodel=None, transform=None, bounding_box=Non
...
 
-    fiducial = compute_fiducial(wcslist, bb)
+    fiducial = compute_fiducial(wcslist, bb)[:2] # spatial only
```
mcara, line 287: pep8
```suggestion
    fiducial = compute_fiducial(wcslist, bb)[:2]  # spatial only
```

jwst/resample/resample_spec.py
```diff
@@ -313,6 +315,91 @@ def _max_virtual_slit_extent(self, wcs_list, target_ra, target_dec):
 
         return y_slit_min, y_slit_max
 
+    def build_miri_lrs_output(self):
+        """ Build MIRI LRS FIXEDSLIT output WCS. """
+        all_wavelength = []
+        all_ra_slit = []
+        all_dec_slit = []
+
+        for im, model in enumerate(self.input_models):
```
mcara, line 324:
```suggestion
        for model in self.input_models:
```

nden:
I appreciate the review of the old code! 
The point of the draft PR is to assess the new approach of computing the spatial transform.

mcara:
> I appreciate the review of the old code!
> The point of the draft PR is to assess the new approach of computing the spatial transform.

That takes time... but now _I think_ I am ready to review the essence.

1. I think almost any way of constructing a linear WCS would work (projections are generally non-linear; here by linear I mean pre-projection transformations that are not distorted + rotations). The issue, I believe, is to get axes oriented as close as possible to the "correct orientation". What is "correct orientation"? I am not sure but I will discuss this below.
2. The approach in this PR uses information from `wcsinfo` such as roll angle to construct the linear part of the WCS transformation in a similar fashion to how the WCS of the CAL images is constructed. The old approach tries to estimate this by fitting lines to the RA and DEC (projected to a plane) values of all valid pixels in the CAL images. It does this _separately_ for `xtan` and `ytan` coordinates. I believe the approach in this PR is more robust.

### Correct Orientation

(In my opinion) One of the goals of resampling spectra is to separate spatial and spectral axes. For example, wavelength changes only as we move along let's say the X axis and spatial component changes only along the Y axis of the resampled image. Not being an expert in MIRI LRS I may be wrong in what follows but I think the only way to do this is to have the Y axis in the resampled image to be along the "slit". Then, the spatial coordinate system should be chosen in such a way that as we "move" (vary celestial coordinates) along the projection of the slit on the sky, those changes to celestial coordinates correspond to changes in `y` coordinate in the resampled image.

Can either the proposed PR or old code accomplish this separation? By construction, these WCSes (old and new - this PR) seem to accomplish this by forcing a separation of coordinates using `Mapping`. The old PR uses the coordinate along the spacial axis in the resampled image and the fitted lines to convert from image to tangent plane and then to world. The new PR does the same but passes this coordinate (2x) to the WCS constructed from footprints:

```python
mapping = Mapping((spatial_axis, spatial_axis, spectral_axis))
wnew = wcs_from_footprints(self.input_models, wavelength=wave_cen)
# define the output wcs
transform = mapping | (wnew.forward_transform & pix_to_wavelength)
```
This concerns me a little bit because, assuming that Y axis is spatial, then `y` is converted to celestial coordinates using `forward_transform(y, y)`. But does the `forward_transform` take into account that first and second arguments are equal (i.e., that we are "moving" along the diagonal of the `forward_transform`'s detector space)? It may be an unexpected additional 45 degree rotation ... For example, why not do something like this:

```python
mapping = Mapping((spatial_axis, spatial_axis, spectral_axis)) | (Const1D(0) & Identity(1) & Identity(1))
```

Basically, why do inputs to `forward_transform` for a diagonal?

Also, if there are distortions in transformations from slit to sky then I think there is no single orientation that would work for all pixels. Distortions themselves will not go away just because constructed resampled image's WCS is rectified. I think this change in the orientation of axes along the slit (or slit not being a straight line on the sky) will translate in distortions (compression/dilation) along the Y (spatial) axis in the resampled image.

I will follow up in a few hours with some images.

mcara:
> Can either the proposed PR or old code accomplish this separation? By construction, these WCSes (old and new - this PR) seem to accomplish this by forcing a separation of coordinates using Mapping.

By "seem to accomplish" I was referring to the separation part only - not necessarily to tracing the projection of the slit on the sky.

mcara:
So, I was wrong in my previous statement (switched old with new S2D files). It actually seems like resampled images created with this PR have very large roundtrip errors _on the sky_.

To check this I wrote the following code:

```python
import numpy as np
from jwst.datamodels import ModelContainer, ImageModel

DEG2PIXEQ = 3600 / 0.11


def round_trip_2x(cal_ra, cal_dec, cal_lam, resampled_model, name=None):
    wcs = resampled_model.meta.wcs
    # to image coordinates:
    x1, y1 = wcs.invert(cal_ra, cal_dec, cal_lam)
    # back to sky:
    ra, dec, lam = wcs(x1, y1)
    # back to image:
    x2, y2 = wcs.invert(ra, dec, lam)
    xerr = np.sqrt((x2 - x1)**2 + (y2 - y1)**2)
    werr = np.sqrt((cal_ra - ra)**2 + (cal_dec - dec)**2) * DEG2PIXEQ
    
    if name is None:
        name = resampled_model.meta.filename
    
    print(f"\n*** Round-tripping for '{name}':")
    print("-----------------------------------")
    print(f"     0th (input world):       {cal_ra},  {cal_dec},  {cal_lam}")
    print(f"     1st (image):             {x1},  {y1}")
    print(f"     3rd (rnd-tripped world): {ra},  {dec},  {lam},  err: {werr}")
    print(f"     2nd (rnd-tripped image): {x2},  {y2},  err: {xerr}")

    
image_no = 1  # or 2

cal_model = ImageModel(f'jw02072001001_06101_0000{image_no}_mirimage_cal.fits')
# processed with the code from this PR:
resampled_model_pr = ImageModel(f'jw02072001001_06101_0000{image_no}_mirimage_s2d.fits')
# original S2D files from JIRA (current code):
resampled_model_master = ImageModel(f'orig/jw02072001001_06101_0000{image_no}_mirimage_s2d.fits')

print(f"TARGET RA, DEC: {cal_model.meta.target.ra}, {cal_model.meta.target.dec}")

# Visually, in DS9, choose a point on the trace:
x0, y0 = [(315.5, 135.5), (332.5, 135.5)][image_no - 1]
# compute this pixel's worldcoordinates using CAL WCS:
cal_ra, cal_dec, cal_lam = cal_model.meta.wcs(x0, y0)

# round trip current code's resampled  WCS and proposed WCS in PR:
round_trip_2x(cal_ra, cal_dec, cal_lam, resampled_model_master, 'CURRENT CODE')
round_trip_2x(cal_ra, cal_dec, cal_lam, resampled_model_pr, 'NEW CODE')
```

This code takes a pixel identified visually to be on the trace and computes world coordinates using CAL WCS. Then it computes pixel coordinates (2nd) in the resampled image and then back sky coordinates (3rd) from resampled pixel coordinates (and then computes again resampled pixel cords from round-tripped world coordinates - 4th).

The result was:

```
TARGET RA, DEC: 64.9725, -54.94808055555552

*** Round-tripping for 'CURRENT CODE':
-----------------------------------
     0th (input world):       64.97259964527014,  -54.94805319581023,  12.027506561725518
     1st (image):             13.850541629668987,  128.45135982693856
     3rd (rnd-tripped world): 64.97259964519108,  -54.94808306577748,  12.027506561725518,  err: 0.9775625645440174
     2nd (rnd-tripped image): 13.850541629629609,  128.45135982693856,  err: 3.937827841582475e-11

*** Round-tripping for 'NEW CODE':
-----------------------------------
     0th (input world):       64.97259964527014,  -54.94805319581023,  12.027506561725518
     1st (image):             14.393441819667217,  128.45135982693856
     3rd (rnd-tripped world): 64.97185191532206,  -54.94815425313875,  12.027506561725518,  err: 24.693647021974645
     2nd (rnd-tripped image): 14.393441819986514,  128.45135982693856,  err: 3.192965891685162e-10
```
By varying `x` coordinates, I got errors of up to 74 pixel-equivalent on the sky for the image created using this PR.  Look at **3rd (rnd-tripped world)**  error: 24.69 pixel-equivalent. For the original image (from JIRA) the error was < 1 pixel.

This may be due to that rotation issue (diagonal) mentioned above.

mcara:
Also, I will reproduce here the images that I put on JIRA showing traces on the CAL and S2D images (obtained using the original and this PR code). Assuming target RA and DEC in `meta` is wrong as @tapastro suggested, I computed traces that produce minimal errors in RA (between RA of trace's pixels and target RA) - red color - and traces that produce minimal errors in DEC - blue. (more details on JIRA).

#### CAL images:

target_traces

#### S2D resampled images obtained with released code:

target_traces_s2d

#### S2D resampled images obtained with this PR's code:

target_traces_s2d_newver

The conclusion is that this PR does not fix the issue of incorrect target RA&DEC coordinates. However, as you indicate, the extracted spectrum is better. This may be due to the code picking up the trace that minimizes DEC instead of RA.

Issue #4436, opened 2020-01-07
Update documentation on creating a new model


eslavich:
The [Creating a new model](https://jwst-pipeline.readthedocs.io/en/latest/jwst/datamodels/new_model.html) page is out-of-date.  At the very least it needs updates to the schema URIs, but probably more issues will turn up when one of us attempts to follow the instructions.

nden:
The new location of the page is https://github.com/spacetelescope/stdatamodels/blob/master/docs/source/jwst/datamodels/new_model.rst

melanieclarke:
It looks like these docs got some updates in https://github.com/spacetelescope/stdatamodels/pull/171; transferring to stdatamodels in case there is more work to do.

PR #7590, opened 2023-05-18
Jp-3207 ifu outlier detection

jemorrison:
Resolves [JP-3207](https://jira.stsci.edu/browse/JP-3207)

Closes #

This PR changes the way outlier detection is done on IFU data, implementing a completely new algorithm developed by @drlaw1558. A change to the outlierpars data models was needed [spacetelescope/stdatamodels#164 for stdatamodels]. In addition, a new data model outlierifuoutput was added to stdatamodels to support analysis of the new algorithm. If save_intermediate_results = True then an output file following outlierifuoutput is created. 

Order of work flow:

1. Get PR 164 in stdatamodels merged. (Done)
2. Get PR 167 in stdatamodels merged (Done)
3. Add the new NIRSpec and MIRI outlierpars reference files. [if not in CRDS will use defaults set in outlier_detection. We can use these defaults for NIRSpec get the PR merged and later load the NIRSpec ref file if they need more time]
4. Merge this PR. 

**Checklist for maintainers**
- [x] added relevant label(s)


jemorrison:
@hbushouse I did not mean to put requirements-max.txt into this PR. I think I removed it. But could you check I did that correctly before we merge

jemorrison:
@hbushouse So how do we record changes made in  datamodels in  stdatamodels ? I put a note under datamodels and I put a note in the stdatamodels CHANGE log.

hbushouse:
> @hbushouse So how do we record changes made in datamodels in stdatamodels ? I put a note under datamodels and I put a note in the stdatamodels CHANGE log.

You can include a link to the stdatamodels PR using the following syntax:
```
[#7590, spacetelescope/stdatamodels#164]
```

drlaw1558:
pars files are ready for both outlier_detection and an updated spec3.  The spec3 file contains updated extract1d calls too for the MRS PR on extract1d changes.
Question though: when these pars files go live, they contain keywords not in the current MAST pipeline, and offline testing suggests that this will cause crashes in old code versions.  How should we deal with that?

nden (changes requested):

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,228 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+
+    def _find_detector_parameters(self):
+        if self.input_models[0].meta.instrument.name.upper() == 'MIRI':
+            diffaxis = 1
+        elif self.input_models[0].meta.instrument.name.upper() == 'NIRSPEC':
+            diffaxis = 0
+        ny, nx = self.inputs[0].data.shape
+        return (diffaxis, ny, nx)
+
+    def convert_inputs(self):
```
nden, line 95: Why is this method necessary? I don't see the `converted` flag used anywhere. Is it to comply with the base class?

nden:
@jemorrison I believe `requirements-max.txt` is removed erroneously in this PR.

Also have you run the code with at least two different detectors? I'm trying to figure out the logic.

braingram:
This PR should include an update to:
https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/__init__.py
as the jwst test suite compares `stdatamodels.jwst.datamodels` to `jwst.datamodels`
https://github.com/spacetelescope/jwst/blob/21c7b517ad3fa4d516555d41847b1d7ca8f93b92/jwst/datamodels/tests/test_api.py#L30
and the addition of a new submodule and model in https://github.com/spacetelescope/stdatamodels/pull/164/ will cause the above test to fail.

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,228 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+        Parameter
+        ---------
+        input_model: ~jwst.datamodels.RampModel
+
+        opt_info: tuple
+        The output arrays needed for the OultierOutputModel.
+
+        Return
+        ---------
+        opt_model: OultierOutputModel
```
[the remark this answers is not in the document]
jemorrison: fixed

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,228 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+            normarr=normarr,
+            minnorm=minnorm)
+        opt_model.meta.kernel_xsize = kernsize_x
+        opt_model.meta.kernel_ysize = kernsize_y
+        opt_model.meta.threshold_percent = threshold_percent
+        return opt_model
+
+    def _find_detector_parameters(self):
+        if self.input_models[0].meta.instrument.name.upper() == 'MIRI':
+            diffaxis = 1
```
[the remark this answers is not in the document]
jemorrison: I just took that from the original code David Law wrote. It is the axis that the differences are done on.
Which is the opposite axis of the dispersion axis. MIRI's dispersion axis is y , the spectral traces are along the y -axis. But the diffaxis for MIRI is 1 (so differences are made left-right)

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,228 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+                comb[1, :, :] = np.abs(rightdiff)
+                combdiff = np.nanmin(comb, axis=0)
+                diffarr[j, :, :] = combdiff
+                j = j + 1
+
+        # minarr final minimum combined differences, size: ny X nx
+        minarr = np.nanmin(diffarr, axis=0)
+        # store where the minarr is nan (neighbor pixels have nan so differences produces a nan)
+        nanminarr = np.isnan(minarr)
+        nanindx = np.where(nanminarr)
```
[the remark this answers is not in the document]
jemorrison: I removed this one nanindx - it is used if 'if_second_check ' is True

jemorrison:
@nden I have run this code on NIRSpec and MIRI data using an association for each instrument that contains 2 detectors. The algorithm pulls the data from 1 detector and is basically trying to find bad pixels that are not in the bad pixel mask because the bad pixels vary with time. So it looks for outliers with adjacent pixels that are outliers on all then input data for that detector.

docs/jwst/outlier_detection/arguments.rst
```diff
@@ -1,8 +1,8 @@
 .. _outlier_detection_step_args:
 
-Step Arguments
-==============
-The `outlier_detection` step has the following optional arguments
+Step Arguments for Non-IFU data
+===============================
+The `outlier_detection` step not non-IFU data has the following optional arguments
```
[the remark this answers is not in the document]
jemorrison: fixed

docs/jwst/outlier_detection/arguments.rst
```diff
@@ -76,3 +76,21 @@ that control the behavior of the processing:
...
+Step Arguments for IFU data
+===========================
+The `outlier_detection` step for IFU data has the following optional arguments
+that control the behavior of the processing:
+
+``--kernel_size`` (string, default='7 7')
+  The size of the kernel to use to normalize the pixel differences. The kernel size
+  must only contain odd values. 
+
+``--threshold_percent`` (float, default='99.8')
```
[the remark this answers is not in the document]
jemorrison: removed ''

docs/jwst/outlier_detection/outlier_detection_ifu.rst
```diff
@@ -5,39 +5,24 @@ Outlier Detection for IFU Data
...
- * Blot median cube to match each original 2D input image
-
-   - Resampled/blotted images are written out to disk if 
-     the ``save_intermediate_results`` parameter is set to `True`
-
- * Perform statistical comparison between blotted image and original image to identify outliers
- * Update input data model DQ arrays with mask of detected outliers
+basic outlier detection algorithm searching for pixels that are consistent outliers
+on the data after the :ref:`calwebb_spec2 <calwebb_spec2>`. After launch it was discovered the
+bad pixels on the MIRI detectors vary with time. The pixels varied from usable to unsable, and at times,
```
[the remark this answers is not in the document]
jemorrison: fixed

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,228 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+        # send input_models back to outlier_detection.py.
+        # self.input_moels is  that is what is returned from outlier_detection.py
+        self.detect_outliers_ifu(self.input_models)
+
+    def flag_outliers(self, idet, uq_det, ndet_files,
+                      diffaxis, nx, ny,
+                      kern_size, threshold_percent,
+                      save_intermediate_results,
+                      ifu_second_check):
+        """Flag outlier pixels in DQ of input images."""
```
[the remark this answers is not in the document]
jemorrison: Forgot to add that - added information on input data

jwst/outlier_detection/outlier_detection_ifu.py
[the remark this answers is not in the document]
jemorrison: @nden see if the use of nanindx now ok. I have np.where because the location of nans is used in other arrays. Maybe there is a smarter way to do this without np.where

jemorrison:
@nden I needed requirements-max.txt locally to be changed so that I could use the newer version that I created to update the outlierParsModel. I added that to the PR by mistake. I guess I did not mean to remove requirements-max.txt completely just my update to it. Is that what you are saying I did ?

jemorrison:
@nden made the changes you requested. I remove a few things that were not needed. I am going to run the regression tests now- more just to make sure nothing breaks. The regression results for IFU outlier detection will be completely different

jemorrison:
@braingram I added OutlierIFUOutput to the jwst __init__.py and now I am getting and error in the check-style ci test. Did I not do it correctly ?

braingram:
> @braingram I added OutlierIFUOutput to the jwst **init**.py and now I am getting and error in the check-style ci test. Did I not do it correctly ?

Thanks! It looks like `check-style` is complaining about an unused import. Adding `OutlierIFUOutput` to `__all__` in `jwst.datamodels.__init__.py` should fix that.

docs/jwst/outlier_detection/arguments.rst
```diff
@@ -76,3 +76,21 @@ that control the behavior of the processing:
...
+===========================
+The `outlier_detection` step for IFU data has the following optional arguments
+that control the behavior of the processing:
+
+``--kernel_size`` (string, default='7 7')
+  The size of the kernel to use to normalize the pixel differences. The kernel size
+  must only contain odd values. 
+
+``--threshold_percent`` (float, default=99.8)
+The threshold (in percent) of the normalized minimum pixel difference used to identify bad pixels.
```
hbushouse, line 91: Might need to indent these 2 lines of text (91+92) to match the indent of the start of text in line 90, in order to make docs build correctly.

docs/jwst/outlier_detection/outlier_detection_ifu.rst
```diff
@@ -5,39 +5,24 @@ Outlier Detection for IFU Data
...
-
- * Blot median cube to match each original 2D input image
-
-   - Resampled/blotted images are written out to disk if 
-     the ``save_intermediate_results`` parameter is set to `True`
-
- * Perform statistical comparison between blotted image and original image to identify outliers
- * Update input data model DQ arrays with mask of detected outliers
+basic outlier detection algorithm searching for pixels that are consistent outliers
+on the data after the :ref:`calwebb_spec2 <calwebb_spec2>`. After launch it was discovered the
```
hbushouse, line 9: the end of this sentence "after the calwebb_spec2." doesn't make sense. Should it be something like "... pixels that are consistent outliers in the calibrated images created by the calwebb_spec2 pipeline"?

docs/jwst/outlier_detection/outlier_detection_ifu.rst
```diff
@@ -5,39 +5,24 @@ Outlier Detection for IFU Data
...
+bad pixels on the MIRI detectors vary with time. The pixels varied from usable to unusable, and at times,
+back to usable  on a time frame that was too short (sometimes as short as 2 days)  to fold into the bad pixel mask applied
+in the :ref:`calwebb_detector1 <calwebb_detector1>`.
```
hbushouse, line 12: "... applied in the calwebb_detector1 *pipeline*"

docs/jwst/outlier_detection/outlier_detection_ifu.rst
```diff
@@ -5,39 +5,24 @@ Outlier Detection for IFU Data
...
+basic outlier detection algorithm searching for pixels that are consistent outliers
+on the data after the :ref:`calwebb_spec2 <calwebb_spec2>`. After launch it was discovered the
+bad pixels on the MIRI detectors vary with time. The pixels varied from usable to unusable, and at times,
+back to usable  on a time frame that was too short (sometimes as short as 2 days)  to fold into the bad pixel mask applied
+in the :ref:`calwebb_detector1 <calwebb_detector1>`. 
+At this time it is believed that NIRSpec IFU data also have bad pixels that vary with time, though the time variation is
+still under study.
+
+An algorithm was developed to flag pixels that are  outliers when compared to their neighbors for a set of
+input files contained in association. The neighbor pixel differences are defined by the dispersion axis. For MIRI data,
```
hbushouse, line 17: "... contained in *an* association."

hbushouse: Not obvious to me what "The neighbor pixel differences are defined by the dispersion axis" really means. Does it perhaps mean that the differences between neighbors are computed in the direction of the dispersion axis? And hence differences between pixels in the spatial direction are ignored?

docs/jwst/outlier_detection/outlier_detection_ifu.rst
```diff
@@ -5,39 +5,24 @@ Outlier Detection for IFU Data
...
+the dispersion axis is along the y-axis and differences are found to the left and right of every
+science pixel. For NIRSpec data, the dispersion axis is along the x-axis and differences are
+found to between the pixels above and below every science pixel. The pixel differences for each input model
```
hbushouse, line 20: "found to between"  remove "to"

docs/jwst/outlier_detection/outlier_detection_ifu.rst
```diff
@@ -5,39 +5,24 @@ Outlier Detection for IFU Data
...
+in the :ref:`calwebb_detector1 <calwebb_detector1>`. 
+At this time it is believed that NIRSpec IFU data also have bad pixels that vary with time, though the time variation is
+still under study.
+
+An algorithm was developed to flag pixels that are  outliers when compared to their neighbors for a set of
+input files contained in association. The neighbor pixel differences are defined by the dispersion axis. For MIRI data,
+the dispersion axis is along the y-axis and differences are found to the left and right of every
+science pixel. For NIRSpec data, the dispersion axis is along the x-axis and differences are
+found to between the pixels above and below every science pixel. The pixel differences for each input model 
+in the associations is determined and is stored in a stack of pixel differences.
```
hbushouse, line 21: associations should probably be singular here (i.e. "association") There's only 1 association being processed at a time.

docs/jwst/outlier_detection/outlier_detection_ifu.rst
```diff
@@ -5,39 +5,24 @@ Outlier Detection for IFU Data
...
+
+An algorithm was developed to flag pixels that are  outliers when compared to their neighbors for a set of
+input files contained in association. The neighbor pixel differences are defined by the dispersion axis. For MIRI data,
+the dispersion axis is along the y-axis and differences are found to the left and right of every
+science pixel. For NIRSpec data, the dispersion axis is along the x-axis and differences are
+found to between the pixels above and below every science pixel. The pixel differences for each input model 
+in the associations is determined and is stored in a stack of pixel differences. 
+For each pixel the minimum difference
+through this stack is determined and normalized. The normalization uses a local median of the difference array
+(set by the kernel size). The input model DQ array is flagged as an outlier if this normalized minimum difference
```
hbushouse, line 24: The DQ array can't be flagged as an outlier. However, pixels that are seen as outliers can be flagged *in* the DQ array.

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -22,28 +16,30 @@ class OutlierDetectionIFU(OutlierDetection):
...
+      2. Loop over cal files
+         a. read in science data
+         b. Store computed neighbor differences for all the pixels.
+            The neighbor pixel  differences are defined by the dispersion axis.
+            For MIRI (disp axis = 1) the neighbors to find differences  are to the left and right of pixel
+            For NIRSpec (disp axis = 0) the neighbors to find the differences are above and below the pixel
+      3. For each input file store the  minimum of the pixel neighbor differences
+      4. Comparing all the differences from all the input data find the minimum neighbor difference
+      5. Normalize minimum difference to local median of difference array
+      6. select outliers by flagging those normailzed minimum values > thershold_percent
```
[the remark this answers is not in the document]
hbushouse: "thershold_percent" -> "threshold_percent"  (typo)

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,228 @@ def __init__(self, input_models, reffiles=None, **pars):
...
-    def _convert_inputs(self):
+        OutlierDetection.__init__(self, input_models, reffiles=reffiles, **pars)
+
+    def create_optional_results_model(self, opt_info):
+        """
+        Creates an OutlierOutputModel from the computed arrays from outlier detection on IFU data.
+
+        Parameter
+        ---------
+        input_model: ~jwst.datamodels.RampModel
```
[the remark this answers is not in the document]
hbushouse: For the purpose of doc strings, are we leaving the references as "jwst.datamodels.XXXX" or are we updating them to point to `stdatamodels`, i.e. `~stdatamodels.jwst.datamodels.RampModel`?  @nden

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -142,7 +148,7 @@ def process(self, input_data):
             elif exptype in IFU_SPEC_MODES:
                 # select algorithm for IFU data
                 detection_step = outlier_registry['ifu']
-                pars['resample_suffix'] = 's3d'
+                pars['resample_suffix'] = 'crf'
```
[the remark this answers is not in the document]
hbushouse: I'm also puzzled by this. Previously the s3d suffix was defined for attaching to the files created by "resampling" each input cal file into a cube, during the process of doing IFU outlier detection. But the new method no longer creates any resampled products while doing outlier detection right?  So I'm guessing this line could actually just be deleted, because it's no longer needed?

hbushouse:
This PR needs to add a jwst.datamodels.outlierifuoutput.py module that contains a pointer to the new `OutlierIFUOutputModel` model sitting over in `stdatamodels`, in the same way that all other models do, e.g. https://github.com/spacetelescope/jwst/blob/master/jwst/datamodels/linearity.py

docs/jwst/outlier_detection/outlier_detection_ifu.rst
jemorrison: Opposite - neighbors are in the spatial direction. I have changed the language - see if makes sense now. 

An algorithm was developed to flag pixels that are  outliers when compared to their neighbors for a set of
input files contained in an association. The neighbor pixel differences are the neighbors in spatial direction. For MIRI data, the dispersion axis is along the y-axis,  neighbor differences are found to the left and right of every
science pixel. For NIRSpec data, the dispersion axis is along the x-axis,  neighbor differences are
found to between the pixels above and below every science pixel

hbushouse:
Not sure why the CI tests are failing now when trying to import the new outlierifuoutput module from `stdatamodels`:
`jwst/datamodels/__init__.py:44: in <module>
    from stdatamodels.jwst.datamodels.outlierifuoutput import OutlierIFUOutputModel
E   ModuleNotFoundError: No module named 'stdatamodels.jwst.datamodels.outlierifuoutput'`
The module is definitely there in stdatamodels==1.5.0.

braingram:
> Not sure why the CI tests are failing now when trying to import the new outlierifuoutput module from `stdatamodels`: `jwst/datamodels/__init__.py:44: in <module> from stdatamodels.jwst.datamodels.outlierifuoutput import OutlierIFUOutputModel E ModuleNotFoundError: No module named 'stdatamodels.jwst.datamodels.outlierifuoutput'` The module is definitely there in stdatamodels==1.5.0.

It looks like the PR adding `outlierifuoutput` was merged Jun 2 (after the 1.5 stdatamodels release on May 16 based off commit dates).
https://github.com/spacetelescope/stdatamodels/commits/master

hbushouse:
> > Not sure why the CI tests are failing now when trying to import the new outlierifuoutput module from `stdatamodels`: `jwst/datamodels/__init__.py:44: in <module> from stdatamodels.jwst.datamodels.outlierifuoutput import OutlierIFUOutputModel E ModuleNotFoundError: No module named 'stdatamodels.jwst.datamodels.outlierifuoutput'` The module is definitely there in stdatamodels==1.5.0.
> 
> It looks like the PR adding `outlierifuoutput` was merged Jun 2 (after the 1.5 stdatamodels release on May 16 based off commit dates). https://github.com/spacetelescope/stdatamodels/commits/master

??? Yet the outlierifuoutput module was included in PR 164, which according to the stdatamodels change log was included in release 1.5.0. So ... something's messed up somewhere in the documentation.  Ultimately I guess we need another stdatamodels release.

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -22,28 +16,30 @@ class OutlierDetectionIFU(OutlierDetection):
...
+      2. Loop over cal files
+         a. read in science data
+         b. Store computed neighbor differences for all the pixels.
+            The neighbor pixel  differences are defined by the dispersion axis.
+            For MIRI (disp axis = 1) the neighbors to find differences  are to the left and right of pixel
+            For NIRSpec (disp axis = 0) the neighbors to find the differences are above and below the pixel
+      3. For each input file store the  minimum of the pixel neighbor differences
+      4. Comparing all the differences from all the input data find the minimum neighbor difference
+      5. Normalize minimum difference to local median of difference array
+      6. select outliers by flagging those normailzed minimum values > threshold_percent
```
hbushouse, line 38:
```suggestion
      6. Select outliers by flagging those normailzed minimum values > threshold_percent
```

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+        # set up array to hold group differences
+        diffarr = np.zeros([ndet_files, ny, nx])
+        j = 0
+        for i, model in enumerate(self.input_models):
+            sci = model.data
+            dq = model.dq
+            detector = model.meta.instrument.detector.lower()
+            # only use data from the same detector
+            if detector == uq_det[idet]:
+                bad = np.where(np.bitwise_and(dq, dqflags.pixel['DO_NOT_USE']).astype(bool))
```
hbushouse, line 196: I think you should be able to do:
```
bad = np.bitwise_and(dq, dqflags.pixel['DO_NOT_USE']).astype(bool)
```
and then use `bad` in the same way in subsequent lines as an array index. It voids using `np.where`.

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+                # There could be a large number of pixels with a sci value of NaN
+                # but the dq flag of DO_NOT_USE has not been set.
+                # This can occur in Non-science regions of the detector.
+                check = np.where(
+                    np.logical_and(~np.bitwise_and(dq, dqflags.pixel['DO_NOT_USE']).astype(bool),
+                                   np.isnan(sci)))
+                log.debug("Number of pixels DQ was not set to DO_NOT_USE and Sci array was Nan{} ".
+                          format(len(check[0])))
+                # set all pixels with dq = DO_NOT_USE to have sci values of Nan
+                bad = np.where(np.bitwise_and(dq, dqflags.pixel['DO_NOT_USE']).astype(bool))
```
hbushouse, line 274: Same here. You shouldn't need the `np.where`. Without it, `bad` will be a full-sized array of booleans (True/False), which can then be used as array indexes in the next line.

hbushouse:
@zacharyburnett @braingram Is there anyway you know of to force the CI tests to use stdatamodels master - just for running CI on this PR - instead of using the latest released version of stdatamodels?

hbushouse:
Regression test run is running at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/740/ using stdatamodels@master (to pick up all the models changes).

nden:
You should be able to use
'stdatamodels @ git+https://github.com/spacetelescope/stdatamodels.git@master',

CHANGES.rst
```diff
@@ -25,13 +25,20 @@ cube_build
 ----------
 
 - Remove deleting the ``spaxel_dq`` array twice when using a weighting method of either msm or emsm. [#7586]
+  
+- Updated to read wavelength range for NIRSpec IFU cubes from the cubepars reference file,
+  instead of setting it based on the data. This makes use of new NIRSpec IFU cubepars reference
+  files with wavelength arrays for the drizzle method. [#7559]
```
nden, line 31:
```suggestion
  files with wavelength arrays for the drizzle method. [#7590]
```

nden, line 31: Why is this log message here?

CHANGES.rst
```diff
@@ -54,12 +61,7 @@ flat_field
 
 - Refactored NIRSpec 1D flat interpolation for improved performance. [#7550]
 
-cube_build
-----------
-
-- Updated to read wavelength range for NIRSpec IFU cubes from the cubepars reference file,
-  instead of setting it based on the data. This makes use of new NIRSpec IFU cubepars reference
-  files with wavelength arrays for the drizzle method. [#7559]
+- Added log messages for reporting flat reference file(s) used. [#7592]
```
nden, line 64: This seems out of place too.

jwst/outlier_detection/outlier_detection.py
```diff
@@ -383,6 +383,10 @@ def detect_outliers(self, blot_models):
             for i in range(len(self.input_models)):
                 self.inputs.dq[i, :, :] = self.input_models[i].dq
 
+    def detect_outliers_ifu(self, inputs):
+        # Update in dq and science flags
+        self.input_models = inputs
```
nden, line 388: I'm not sure what this method does. It is called in `outlier_detection_ifu.do_detection` where effectively it does
`self.input_models = self.input_model`.  Unless I'm missing something.

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+        Parameters
+        ----------
+        input_model: ~stdatamodels.jwst.datamodels.RampModel
+
+        opt_info : tuple
+        The output arrays needed for the OultierIFUOutputModel.
+
+        Returns
+        ---------
+        opt_model : OutlierOutputModel
```
nden, line 72:
```suggestion
        opt_model : OutlierIFUOutputModel
```

nden: and below

CHANGES.rst
hbushouse: It had to be put in to remove conflicts, since this PR has been sitting around for a while as others have been merged, including at least one merger that had that log entry in the wrong place (so I moved it to the right place, under `cube_build`).

hbushouse:
> You should be able to use 'stdatamodels @ git+https://github.com/spacetelescope/stdatamodels.git@master',

That's the correct way to run regression tests using master, but I don't know how to specify such a thing for the *CI* tests. They just do their own thing without any obvious way to override and add build parameters.

braingram:
> > You should be able to use 'stdatamodels @ git+https://github.com/spacetelescope/stdatamodels.git@master',
> 
> That's the correct way to run regression tests using master, but I don't know how to specify such a thing for the _CI_ tests. They just do their own thing without any obvious way to override and add build parameters.

One option would be to change the requirement in `setup.cfg` from:
https://github.com/spacetelescope/jwst/blob/343edb70ad1bfcba292791c5ff43f06152f18dc2/setup.cfg#L47
to the suggestion by @nden
```
stdatamodels @ git+https://github.com/spacetelescope/stdatamodels.git@master
```

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+        if ifu_second_check:
+            # store where the minarr is nan (neighbor pixels have nan so differences produces a nan)
+            nanminarr = np.isnan(minarr)
+            nanindx = np.where(nanminarr)
+
+        # Update DQ flag
+        for i in range(len(self.input_models)):
+            model = self.input_models[i]
+            sci = model.data
+            dq = model.dq
```
nden, line 259: It looks like these are worked with only if the detector matches, should be in the `if` statement.

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+                # but the dq flag of DO_NOT_USE has not been set.
+                # This can occur in Non-science regions of the detector.
+                check = np.where(
+                    np.logical_and(~np.bitwise_and(dq, dqflags.pixel['DO_NOT_USE']).astype(bool),
+                                   np.isnan(sci)))
+                log.debug("Number of pixels DQ was not set to DO_NOT_USE and Sci array was Nan{} ".
+                          format(len(check[0])))
+                # set all pixels with dq = DO_NOT_USE to have sci values of Nan
+                bad = np.where(np.bitwise_and(dq, dqflags.pixel['DO_NOT_USE']).astype(bool))
+                sci[bad] = np.nan
```
nden, line 275: Following Howards's question about changing the data - was this intentional?

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+            ndet_files = int(len(indx))
+            self.flag_outliers(idet, uq_det, ndet_files,
+                               diffaxis, nx, ny,
+                               kern_size, threshold_percent,
+                               save_intermediate_results,
+                               ifu_second_check)
+
+        # send input_models back to outlier_detection.py.
+        # self.input_moels is  that is what is returned from outlier_detection.py
+        self.detect_outliers_ifu(self.input_models)
```
nden, line 148: What does this do? Is it necessary?

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+            If True then perform a secondary check searching for outliers. This will set outliers
+            whereever the difference array of adjacent pixels is a Nan. 
+        """
+
+        # set up array to hold group differences
+        diffarr = np.zeros([ndet_files, ny, nx])
+        j = 0
+        for i, model in enumerate(self.input_models):
+            sci = model.data
+            dq = model.dq
```
nden, line 192: Since `model.data` and `model.dq` are worked with only when the detector matches, these can go in the `if` block.

jwst/outlier_detection/outlier_detection_ifu.py
nden: 👍 Second that.

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+            detector = model.meta.instrument.detector.lower()
+            # only use data from the same detector
+            if detector == uq_det[idet]:
+                bad = np.where(np.bitwise_and(dq, dqflags.pixel['DO_NOT_USE']).astype(bool))
+                # set all science data that have DO_NOT_USE to NAN
+                sci[bad] = np.nan
```
nden, line 198: This changes the science data. Was it intentional?

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+        # Normalise the differences to a local median image to deal with ultra-bright sources
+        normarr = medfilt(minarr, kernel_size=kern_size)
+        nfloor = np.nanmedian(minarr)/3
+        normarr[normarr < nfloor] = nfloor  # Ensure we never divide by a tiny number
+        minarr_norm = minarr / normarr
+        # Percentile cut of the central region (cutting out weird detector edge effects)
+        pctmin = np.nanpercentile(minarr_norm[4:ny-4, 4:nx-4], threshold_percent)
+        log.debug("Flag pixels with values above {} {}: ".format(threshold_percent, pctmin))
+        # Flag everything above this percentile value
+        indx = np.where(minarr_norm > pctmin)
```
nden, line 234:
```suggestion
        indx = minarr_norm > pctmin
```

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+                    np.logical_and(~np.bitwise_and(dq, dqflags.pixel['DO_NOT_USE']).astype(bool),
+                                   np.isnan(sci)))
+                log.debug("Number of pixels DQ was not set to DO_NOT_USE and Sci array was Nan{} ".
+                          format(len(check[0])))
+                # set all pixels with dq = DO_NOT_USE to have sci values of Nan
+                bad = np.where(np.bitwise_and(dq, dqflags.pixel['DO_NOT_USE']).astype(bool))
+                sci[bad] = np.nan
+
+                # Basic setting outliers: flagging those at are found in from Percentage cut
+                sci[indx] = np.nan
```
nden, line 278: And this change to sci?

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+                        len(indx[0]), uq_det[idet]))
+                    log.info("Number of outlier pixels flagged in second check: {} on detector {} ".format(
+                        nadditional, uq_det[idet]))
+
+                total_bad = len(indx[0]) + nadditional
+                percent_cr = total_bad / (model.data.shape[0] * model.data.shape[1]) * 100
+                log.info(f"Total #  pixels flagged as outliers: {total_bad} ({percent_cr:.2f}%)")
+                # update model
+                model.dq = dq
+                model.data = sci
```
nden, line 305: There's no need for this assignment since the `sci` and `dq` arrays were not copies but references so `model.data` and `model.dq` were modified along the way.

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,258 @@ def __init__(self, input_models, reffiles=None, **pars):
...
+
+                nadditional = 0
+                # Second level of setting outliers: flagging pixels were minarr was a Nan
+                # This will also catch pixels that have a sci of Nan but the DQ flags did
+                # not have DO_NOT_USE set
+                if ifu_second_check:
+                    # For counting purposes, count the number of science values that were valid (not Nan)
+                    # after basic flagging in the nanminarr region that will now  be flagged as a Nan.
+                    additional = np.where(~np.isnan(sci[nanindx]))
+                    nadditional = len(additional[0])
```
nden, line 290: The above two lines can be written as `nadditional = (~np.isnan(sci[nanindx])).sum()`. It's ~4x faster.

jwst/outlier_detection/outlier_detection_ifu.py
jemorrison: yes I intentionally made the science data Nan for outliers. I might have misunderstood a conversion on a tag up about this. I think it LESS confusing if the science data is flagged as a NAN if the DQ is DO_NOT _USE. That is basically what is done in cal detector 1. I also think it is less confusing to the users if BAD data has a science value of NAN. For NIRSpec there where 1000s of input data that had DQ set to DO_NOT_USE but the input data was not Nan and there were science data that was NAN but the DQ flags did not have DO_NOT_USE. Super confusing.  I first had to clean up all the input data so I was not using bad data  or a bunch of nans in the comparison.  I worked with an NIRSpec user heavily with this data (it started as a help desk ticket). I strongly recommend that if the DQ flag is set to DO_NOT_USE the Science pixel also is flagged as Nan. It is cleaner. But I will take that out if you disagree.

jwst/outlier_detection/outlier_detection_ifu.py
jemorrison: done

jwst/outlier_detection/outlier_detection_ifu.py
jemorrison: @nden @hbushouse The above line gives as different number of pixels being flagged.
I use the len(indx[0]) later in the code  to tell me how many pixels are flagged. I think this is useful information to provide to the user. It could help tell us if something is not working correctly if it flags 10,000 of pixels. Is there another way to find out the number of times minarr_norm> pctmin without using np.where.

jemorrison: Ok I used Nadia's trick elsewhere and to get the number of indx = minarr_norm > pctmin I just have to find indx.sum(). So guess that is better than np.where

jwst/outlier_detection/outlier_detection.py
jemorrison: I originally had a problem getting the step up update the out dq pane because of the way outlier_detection_step is set up. I have it all cleaned up now

jemorrison:
@nden @hbushouse 
This PR should now be ready for final review. Thanks for all the suggestions it has made this step better (and I learned a few things along the way).

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -22,28 +16,30 @@ class OutlierDetectionIFU(OutlierDetection):
...
+      2. Loop over cal files
+         a. read in science data
+         b. Store computed neighbor differences for all the pixels.
+            The neighbor pixel  differences are defined by the dispersion axis.
+            For MIRI (disp axis = 1) the neighbors to find differences  are to the left and right of pixel
+            For NIRSpec (disp axis = 0) the neighbors to find the differences are above and below the pixel
+      3. For each input file store the  minimum of the pixel neighbor differences
+      4. Comparing all the differences from all the input data find the minimum neighbor difference
+      5. Normalize minimum difference to local median of difference array
+      6. select outliers by flagging those normailzed minimum values > threshold_percent
```
hbushouse, line 38:
```suggestion
      6. Select outliers by flagging those normalized minimum values > threshold_percent
```

jwst/outlier_detection/outlier_detection_ifu.py
```diff
@@ -53,203 +49,253 @@ def __init__(self, input_models, reffiles=None, **pars):
...
-    def _convert_inputs(self):
-        self.input_models = self.inputs
-        self.converted = False
+        Creates an OutlierOutputModel from the computed arrays from outlier detection on IFU data.
+
+        Parameter
+        ---------
+        input_model: ~stdatamodels.jwst.datamodels.RampModel
+
+        opt_info: tuple
```
hbushouse, line 66:
```suggestion
        Parameters
        ----------
        input_model : ~stdatamodels.jwst.datamodels.IFUImageModel

        opt_info : tuple
```

hbushouse (approved):

hbushouse:
Another (hopefully final?) regtest run started at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/743/

hbushouse:
Not sure I understand why the CI tests are still failing due to some kind of error with not finding jwst.datamodels.outlierifuoutput.py, even though that module is now included in this PR and I *think* it also has the appropriate change to jwst/setup.cfg to use stdatamodels/master. @zacharyburnett @braingram any ideas?

braingram:
> Not sure I understand why the CI tests are still failing due to some kind of error with not finding jwst.datamodels.outlierifuoutput.py, even though that module is now included in this PR and I _think_ it also has the appropriate change to jwst/setup.cfg to use stdatamodels/master. Everything seems to run fine when using this PR branch in regression tests, so maybe it's just something unique in what the CI tests are trying to test. @zacharyburnett @braingram any ideas?

I believe the CI is throwing errors because the file `jwst/datamodels/outlierifuoutput.py` added in the PR conflicts with the module shadowing added in https://github.com/spacetelescope/jwst/pull/7605 `jwst/datamodels/outlierifuoutput.py` can be removed from this PR which should fix the error.

The 2 errors are because pytest is inspecting `jwst/datamodels/outlierifuoutput.py` to look for tests. After it imports the file, the module appears at a different path (because of the changes in #7605) and pytest gives up with an error.

I believe the regression tests aren't showing the same error because they don't search the `jwst/datamodels` directory for tests (assuming it's using the `pytest --bigdata jwst/regtest` in the readme.

Let me know if it's helpful to document any of this. There is an open issue about improving the documentation on adding a new datamodel which is likely related: https://github.com/spacetelescope/jwst/issues/4436

hbushouse:
@braingram OK, I'll try removing the outlierifuoutput.py module from this PR and see what happens.

hbushouse:
Yay! CI tests now succeed. Thanks @braingram

nden (approved):

docs/jwst/outlier_detection/outlier_detection_ifu.rst
```diff
@@ -5,39 +5,25 @@ Outlier Detection for IFU Data
 
 This module serves as the interface for applying ``outlier_detection`` to IFU
 observations, like those taken with NIRSpec and MIRI.  The code implements the
```
hbushouse, line 7:
```suggestion
observations, like those taken with NIRSpec and MIRI.
```

docs/jwst/outlier_detection/outlier_detection_ifu.rst
```diff
@@ -5,39 +5,25 @@ Outlier Detection for IFU Data
...
+
+An algorithm was developed to flag pixels that are  outliers when compared to their neighbors for a set of
+input files contained in an association. The neighbor pixel differences are the neighbors in spatial direction. For MIRI data ,the neighbor differences are found to the left and right of every
+science pixel. While for NIRSpec data neighbor differences are
+found  between the pixels above and below every science pixel. The pixel differences for each input model 
+in the association is determined and is stored in a stack of pixel differences. 
+For each pixel the minimum difference
+through this stack is determined and normalized. The normalization uses a local median of the difference array
+(set by the kernel size). A pixel is flagged as an outlier if  this normalized minimum difference
+is greater than the input threshold percentage.  Pixels that are found to be outliers are flaged in in the DQ array.
```
hbushouse, line 25:
```suggestion
The algorithm searches for pixels in calibrated 2D IFU images created by the
:ref:`calwebb_spec2 <calwebb_spec2>` pipeline that are consistent outliers
across multiple exposures.
A complication with this approach is the fact that bad pixels on the MIRI detectors vary with time.
The pixels change state from usable to unusable and, at times,
back to usable on a time frame that is too short (sometimes as short as 2 days)  to fold into the bad pixel mask applied in the :ref:`calwebb_detector1 <calwebb_detector1>` pipeline. 
At this time it is believed that NIRSpec IFU data also have bad pixels that vary with time, though the time variation is
still under study.

The IFU algorithm detects and flags pixels that are  outliers when compared to their neighbors for a set of
input files contained in an association. The neighbors used in the pixel difference calculations are the neighbors
in the spatial direction along the IFU slices. For MIRI data, the neighbor differences are found to the left and right
of every science pixel (dispersion axis is vertical), while for NIRSpec data neighbor differences are
found  between the pixels above and below every science pixel (dispersion axis is horizontal).
The pixel differences for each image 
in the association are determined and stored in a stack of pixel differences. 
For each pixel the minimum difference
through this stack is determined and normalized. The normalization uses a local median of the difference array
(set by the kernel size). A pixel is flagged as an outlier if  this normalized minimum difference
is greater than the input threshold percentage.  Pixels that are found to be outliers are flaged in in the DQ array.
```

hbushouse:
Woot! CI success. As soon as the latest regtest run at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/751/console finishes, I'll be merging this.

PR #7598, opened 2023-06-08
JP-3119: Add special refpix treatment for NIRSpec "edgeless" subarrays 

tapastro:
Resolves [JP-3119](https://jira.stsci.edu/browse/JP-3119)

This PR addresses the lack of reference pixel corrections for NIRSpec subarrays that do not share an edge with the full frame readout. This assigns the first and last four columns (i.e. in dispersion direction) that are claimed to be un-illuminated for filter/grating combinations that utilize these subarrays.

**Checklist for maintainers**
- [x] added relevant label(s)


CHANGES.rst
```diff
@@ -96,6 +96,12 @@ ramp_fitting
   of groups in a segment are computed when applying optimal weighting to line
   fit segments. [#7560, spacetelescope/stcal#163]
 
+refpix
+------
+
+- Assign reference pixel flag to first and last four columns for
+  NIRSpec subarrays that do not share an edge with full frame. [#7598]
```
hbushouse, line 103:
```suggestion
  NIRSpec subarrays that do not share an edge with full frame,
  so that corrections can be computed from those unilluminated pixels. [#7598]
```

jwst/refpix/reference_pixels.py
```diff
@@ -269,6 +277,12 @@ def get_pixeldq(self):
             pixeldq[4:fullrows - 4, 0:4] = refpixdq_dontuse
             pixeldq[4:fullrows - 4, fullcols - 4:fullcols] = refpixdq_dontuse
             pixeldq[self.rowstart:self.rowstop, self.colstart:self.colstop] = self.input_model.pixeldq.copy()
+            if self.subarray in NRS_edgeless_subarrays:
+                # Log assignment as rows (in DMS plane) despite assigning columns (in detector plane)
+                log.info(f"Subarray {self.subarray} has no reference pixels: "
+                         f"assigning top and bottom four rows as reference pixels.")
+                pixeldq[self.rowstart:self.rowstop, self.colstart:self.colstart+4] = dqflags.pixel['REFERENCE_PIXEL']
```
hbushouse, line 284: Is it OK to just assign them the REFERENCE_PIXEL value or should it be OR'd with any existing flags already set for those pixels in the pixeldq array? For example you don't want to wipe out any existing DO_NOT_USE flags, because it looks like there's a check for that later on.

tapastro: Good call; this change doesn't appear to significantly alter the output for my test data, but it should propagate the flags from dq_init if present.

docs/jwst/refpix/description.rst
```diff
@@ -136,6 +136,12 @@ the reference pixels in each group and subtracted from each pixel.
   are populated when the ``dq_init`` step is run, so it is important to run that
   step before running the ``refpix`` step on subarray data.
 
+  Additionally, certain NIRSpec subarrays (SUB32, SUB512 and SUB512S) do not read out
+  any detector edges, leading to a lack of default reference pixels in the subarray.
+  For these subarrays, we assign the first and last four columns the "REFERENCE_PIXEL"
+  DQ flag, as they should not receive any incoming light with the filter/grating combinations
+  for which they are approved for use.
```
hbushouse, line 143:
```suggestion
  Additionally, certain NIRSpec subarrays (SUB32, SUB512 and SUB512S) do not include
  any physical reference pixels in their readouts.
  For these subarrays, the first and last four image columns should not receive
  any incoming light with the filter+grating combinations for which they are
  approved for use, hence they can be used in place of actual reference pixels.
  The step assigns the "REFERENCE_PIXEL" DQ flag to these image columns,
  which then causes them to be used to perform the reference pixel correction.
```

PR #166, opened 2016-08-17
Add NIRSpec bad shutter ref file keyword to core schema

hbushouse:
Added the reference file keyword "R_BADSHR", which will be used to record the name of the NIRSpec bad shutter reference file that was used during calibration.

PR #7600, opened 2023-06-09
JP-3224 MRS time dependent Photom correction 

jemorrison:
Resolves [JP-3224](https://jira.stsci.edu/browse/JP-3224)

Closes #

This PR adds a correction for a time-variable photometric response that is particularly significant at long wavelengths. An update to stdatamodels was needed to read in the new format of the reference files. The PR for stdatamodels for this change is PR 166 it has been merged.

DO NOT MERGE THIS PR. THIS PR IS A STILL BEING WORKED.
I pushed this to GitHub for David Law to review the basic set up. 

**Checklist for maintainers**
- [x] added relevant label(s)


jemorrison:
@drlaw1558 It is getting closer to what you what. I am going to ask someone about a better way to test if the time coefficients are in the reference file. But this is close.

jemorrison:
Not ready for review - need to add some comments

jemorrison:
@nden  or @hbushouse I am not sure the best way to check if a photom reference file contains the new time coefficients. I want the  step to apply the time correction IF the time coefficients are in the reference file. I have added the new time coefficients to the photom data model (#166 already merged into stdatamodels).  Since the reading of the photom reference files is done through datamodels, I check if the timecoeffs exist in photom.py by  checking if there are values in the first array: np.any(ftab.timecoeff_ch1['binwave']. Is there a better way to do this

jemorrison:
@drlaw1558 I realize I need to update the documentation on MRS photom. Is it your plan to include the timecoefficients for ALL the MRS photom reference files ? Or just the standard operational ones  (ie. not cross dichroics - LONG-MEDIUM, LONG-SHORT, ....)

drlaw1558:
@jemorrison I'm not touching cross-dichroic right now; depending on what the root cause for the loss is (e.g., if due to optical surfaces) the cross-dichroic behavior could be quite different from the normal modes.  Since cross dichroic is currently engineering-only those files won't deal with it.

jwst/photom/photom.py
```diff
@@ -96,6 +98,9 @@ def __init__(self, model, inverse=False, source_type=None, correction_pars=None)
         source_type : str or None
             Force processing using the specified source type.
 
+        mrs_time_correction: boolean
```
nden, line 101:
```suggestion
        mrs_time_correction : bool
```

jwst/photom/photom.py
```diff
@@ -497,6 +503,17 @@ def calc_miri(self, ftab):
             # Update the science dq
             self.input.dq = np.bitwise_or(self.input.dq, ftab.dq)
 
+            # Check if reference file contains time dependent correction
+            if np.any(ftab.timecoeff_ch1['binwave']) and self.mrs_time_correction:
```
nden, line 507: The check as written assumes `ftab.timecoeff_ch1` exists. If that's not always the case it should include also a check like `if hasattr(ftab, 'timecoeff_ch1`)`

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+def expfunc_bounded(x, a, b, c, x0):
+    """ Function using time dependent coefficients """
+
+    temp = a * np.exp(-b * (x - x0) / 100) + c
+    temp[temp > 1] = 1
+    return temp
+
+
+def get_correction_function(side, timecoeff, mid_time):
+    """ Find the time and wavelength dependent function """
```
nden, line 18: needs definition of parameters

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+    # We want to correct the channels seperately
+    # If we have MIRIFUSHORT data then channel 1 is on the left and channel 2
+    # is on the right. If we have MIRIFULONG  data then channel 4 is on the left
+    # and channel 3 is on right.
+
+    timecoeff = {}
+    timecoeff['left'] = {}
+    timecoeff['right'] = {}
+    left = 'ch1'
+    right = 'ch2'
```
nden, line 76: Because `left` and `right` are not conditionally set they will be assigned to the cross-dichroic case as well. Is this the intention?

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+    timecoeff['right']['acoeff'] = table_ch[right]['acoeff']
+    timecoeff['right']['bcoeff'] = table_ch[right]['bcoeff']
+    timecoeff['right']['ccoeff'] = table_ch[right]['ccoeff']
+    timecoeff['right']['x0'] = table_ch[right]['x0']
+
+    ysize, xsize = input.data.shape
+    y, x = np.mgrid[:ysize, :xsize]
+    # the correction is time and wavelength dependent. Pull out the
+    # wavelength of the data
+    _, _, wave = input.meta.wcs(x, y)
```
nden, line 104: This doesn't take into account the bounding box. The preferred method is
```

x,y=grid_from_boundign_box(input.meta.wcs.bounding_box)
```
In addition it takes care of using the center of the pixels.

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Find the function based on time and wavelength
+    func = get_correction_function(side, timecoeff, mid_time)
+    waveim = wave[:, l_xstart:l_xend]
+
+    # Determine the correction based input wavelength
+    result1_1d = func(waveim.ravel())
+    result1 = np.reshape(result1_1d, waveim.shape)
+    indx = np.where(waveim == 0)
+    result1[indx] = 0.0
```
nden, line 119: Suggestion:
```
result1[waveim==0] = 0
```

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # correction of the right side: set up the pixels to extract this region
```
nden, line 121: It looks like the code to do the region extraction is a good candidate for a function, so the code for `left` and `right` is not repeated. Decreases the chance of bugs twice. :-)

nden (changes requested):
I did a superficial review. I can't review in earnest unless there's either documentation or tests (best both) because I'm not sure of the details of what it is supposed to do.

hbushouse:
> @nden or @hbushouse I am not sure the best way to check if a photom reference file contains the new time coefficients. I want the step to apply the time correction IF the time coefficients are in the reference file. I have added the new time coefficients to the photom data model (#166 already merged into stdatamodels). Since the reading of the photom reference files is done through datamodels, I check if the timecoeffs exist in photom.py by checking if there are values in the first array: np.any(ftab.timecoeff_ch1['binwave']. Is there a better way to do this

Given that the datamodel schema for the ref file has been modified to include the additional coefficient extensions, I'm not sure it's possible to ever load an old ref file that does NOT have them and hence no check may be necessary. The schema validator may crash if those extensions aren't present. But I'm not entirely sure. Have you experimented with loading an old ref file with the new data model?  If it works, then you might be able to use the `getarray_noinit()` method that's available in the data model class (see e.g. https://github.com/spacetelescope/stdatamodels/blob/master/src/stdatamodels/jwst/datamodels/ramp.py#L49). If calling that method succeeds, then you know the array is there. If it fails, it isn't.

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+from scipy.interpolate import RegularGridInterpolator
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+
+def expfunc_bounded(x, a, b, c, x0):
+    """ Function using time dependent coefficients """
+
+    temp = a * np.exp(-b * (x - x0) / 100) + c
```
hbushouse, line 12: Curious what the units of `x` and `x0` are and hence the need for the division by 100. Perhaps add a list of parameter definitions giving that kind of info.

jwst/photom/ifu_mrs.py
hbushouse, line 18: Minor nit, but I don't think I'd describe this as "finding" the time and wavelength function, but rather "building" or "constructing" the function based on retrieved coefficients.

hbushouse: I agree that having a definition for params like `side` would help a lot.

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+    return func
+
+
+def time_correction(input, detector, ftab, mid_time):
+    """
+    Find the time and wavelength dependent photom correction for MRS data.
+
+    Parameters
+    ----------
+    input : JWST data model
```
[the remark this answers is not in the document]
hbushouse: Presumably it's an `IFUImageModel`?

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+        Input science data model to be corrected.
+
+    detector : string
+        MRS detector working on
+
+    ftab : stdatamodels.jwst.datamodels.MirMrsPhotomModel
+        MRS PHotom reference file
+
+    mid_time: float
```
hbushouse, line 49:
```suggestion
    mid_time : float
```

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+    # and channel 3 is on right.
+
+    timecoeff = {}
+    timecoeff['left'] = {}
+    timecoeff['right'] = {}
+    left = 'ch1'
+    right = 'ch2'
+    timecoeff['left']['xstart'] = 0
+    timecoeff['left']['xend'] = 512
+    timecoeff['right']['xstart'] = 513
```
hbushouse, line 79: If these are python zero-indexed array indexes, wouldn't the right half start with column 512, instead of 513?

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+    timecoeff['right']['xend'] = 1031
+    if detector == 'MIRIFULONG':
+        left = 'ch4'
+        right = 'ch3'
+        # Check the reference file has the time dependent coefficients
+        # check that table 1 wavelength bin is an array with values
```
hbushouse, line 85: Try using the `getarray_noinit()` method on the input model to see if the coeff arrays exist. What happens if they don't exist? Do you set all the coeffs to suitable defaults, such as zero or one, so that the correction factors are all 1?

jwst/photom/ifu_mrs.py
hbushouse: with all of the appropriate whitespace between elements, in order to make codestyle happy!  ;-)

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+    # correction for left side: set up the pixels to extract this region
+    side = 'left'
+    l_xstart = timecoeff[side]['xstart']
+    l_xend = timecoeff[side]['xend']
+
+    # Find the function based on time and wavelength
+    func = get_correction_function(side, timecoeff, mid_time)
+    waveim = wave[:, l_xstart:l_xend]
+
+    # Determine the correction based input wavelength
```
hbushouse, line 115:
```suggestion
    # Determine the correction based on input wavelength
```

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+    result2 = np.reshape(result2_1d, waveim.shape)
+    indx = np.where(waveim == 0)
+    result2[indx] = 0.0
+
+    # Combine the bands to have 1 correction for the entire detector
+    result = np.zeros((ysize, xsize))
+    result[:, l_xstart: l_xend] = result1
+    result[:, r_xstart: r_xend] = result2
+    indx = np.where(result == 0)
+    result[indx] = 1.0
```
hbushouse, line 141:
```suggestion
    result[result==0] = 1.0
```

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
...
+    indx = np.where(np.isnan(result))
+    result[indx] = 1.0
```
hbushouse, line 143:
```suggestion
    result[np.isnan(result)] = 1.0
```

jwst/photom/ifu_mrs.py
```diff
@@ -0,0 +1,145 @@
+import logging
```
hbushouse, line 1: Not a big deal, but given that this new module is exclusive to MIRI MRS, and not for all IFU modes, I would suggest renaming to `miri_mrs.py`.

nden:
> @nden or @hbushouse I am not sure the best way to check if a photom reference file contains the new time coefficients. 

I added a suggestion in the code.

jwst/photom/photom.py
```diff
@@ -9,6 +9,7 @@
 from stdatamodels.jwst.datamodels import dqflags
 
 from .. lib.wcs_utils import get_wavelengths
+from . import ifu_mrs
```
hbushouse, line 12: If you follow the suggestion above to rename `ifu_mrs` to `miri_mrs`, this line will need to change.

jwst/photom/photom.py
hbushouse: Not sure the `hasattr()` approach is correct. Because it's defined in the datamodel schema, `hasattr` is likely to always return true, even if the array/extension doesn't exist in the input file that was loaded. Suggest using the `getarray_noinit` method in the model class, within a try/except block. If the array isn't in the file, it'll fail.

jwst/photom/photom.py
```diff
@@ -497,6 +503,17 @@ def calc_miri(self, ftab):
             # Update the science dq
             self.input.dq = np.bitwise_or(self.input.dq, ftab.dq)
 
+            # Check if reference file contains time dependent correction
+            if np.any(ftab.timecoeff_ch1['binwave']) and self.mrs_time_correction:
+                log.info("Applying MRS IFU time dependent correction.")
+                mid_time = self.input.meta.exposure.mid_time
+                correction = ifu_mrs.time_correction(self.input, self.detector,
+                                                     ftab, mid_time)
+                self.input.data /= correction
```
hbushouse, line 512: Just so I understand the overall approach here, is the regular flux calibration correction applied first, and then the time-dependent correction is applied separately here as an additional correction?

hbushouse:
I've confirmed that when loading an old style MIRI MRS photom ref file (e.g. jwst_miri_photom_0079.fits) into the new version of `MirMrsPhotomModel`, which contains definitions for `timecoeff_ch1` through `timecoeff_ch4`, the result of `hasattr(ref_model, "timecoeff_ch1")` is always True and hence can't be used to tell if the new arrays exist.

So using a try/except block with `getarray_noinit` is the way to go, e.g.
```
try:
    ref_model.getarray_noinit("timecoeff_ch1")
except AttributeError:
    # Old style ref file; do whatever necessary to skip the new correction
```

jwst/photom/photom.py
jemorrison: yes.  Normal photom correction and if the time coefficients for band are present then make the time_correction. Unless the user set the mrs_time_correction to False

jwst/photom/photom.py
jemorrison: the hasattr does not work - it is always true because it is in the data models. I tried that first

jwst/photom/ifu_mrs.py
jemorrison: At this time no plans are to include the time coefficient in the cross dichroic case. That is an engineering mode and they do not have the data to derive a time correction. If this changes in the future we can come back to this

jwst/photom/photom.py
```diff
@@ -96,6 +98,9 @@ def __init__(self, model, inverse=False, source_type=None, correction_pars=None)
         source_type : str or None
             Force processing using the specified source type.
 
+        mrs_time_correction: bool
```
hbushouse, line 101: Need a whitespace before the colon

hbushouse (commented):
Looks like it still needs photom step doc updates to explain this new time-dependent correction for MRS.

jemorrison:
@finish updates from review. Going the work on the docs for this addition and figure out a unit test(s) to add.

jwst/photom/photom.py
hbushouse, line 101:
```suggestion
        mrs_time_correction : bool
```

hbushouse:
Started a regtest run at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/746

hbushouse:
regtest run 746 came back clean. Of course this was run using all the existing (old) reference files and hence the time-dependent correction was not applied. But at least it shows that running the new code with the old ref files causes no harm.

drlaw1558:
Hm, a quick check with the new ref files I created suggests it also isn't using the new corrections when they ARE there.  Testing quickly to see if I'm doing something wrong.

hbushouse:
> Hm, a quick check with the new ref files I created suggests it also isn't using the new corrections when they ARE there. Testing quickly to see if I'm doing something wrong.

You need to be using either stdatamodels/master or stdatamodels==1.6.0 in your environment, in order to be using the updated data model for the photom ref file that recognizes and loads the new timecoeff extensions.

drlaw1558:
Yep, that got it.  I'd expected a pip install -e . to get the required data model update, but I guess that doesn't work until everything is merged.  Updating by hand looks like it fixes the issue.  Result is as expected, i.e. the loss is fixed.

hbushouse:
So ... no loss is our gain.

docs/jwst/photom/arguments.rst
```diff
@@ -10,3 +10,8 @@ The ``photom`` step has the following optional arguments.
 ``--source_type`` (string, default=None)
   Force the processing to use the given source type (POINT, EXTENDED),
   instead of using the information contained in the input data.
+
+``--mrs_time_correction`` (boolean, default=False)
```
hbushouse, line 14: The default value in the actual code module is True.

nden:
The documentation build is failing because it reference a non-existing image file `Model_summary.png`.

nden (approved):

docs/jwst/photom/main.rst
```diff
@@ -146,3 +146,24 @@ sensitivity and pixel size data are used to compute a 2-D sensitivity map
...
+Figure 1:
+Time-dependent decrease in the observed MRS count rate from internal flat calibration observations.
+Solid points illustrate measurements at the central wavelength of each of the 12 MRS bands;
+curves represent the best fit models used for correction by the pipeline.
+
+	   
+
+If the time-dependent throughput correction has been
+determined for a MRS band the photom reference file will contain
+additional extension of binary tables containing the correction parameters.
```
hbushouse, line 169:
```suggestion
MIRI MRS data have a time-variable photometric response that is significant at
long wavelengths. A correction has been derived from observations of calibration standard stars.
The form of the correction uses an exponential function that asymptotically approaches a
constant value in each wavelength band. A plot of the count rate loss in each MRS
band, as a function of time, is shown in Figure 1.

.. figure:: Model_summary.png
   :scale: 50%
   :align: center
	   
Figure 1:
Time-dependent decrease in the observed MRS count rate as measured from internal flat-field exposures.
Solid points illustrate measurements at the central wavelength of each of the 12 MRS bands;
curves represent the best fit models used for correction in the pipeline.

The MRS photom reference file contains a table of correction coefficients
for each band in which a correction has been determined. If the time-dependent
coefficients are present in the reference file for a given band, the photom step will
apply the correction to the exposure being processed.
```

docs/jwst/references_general/photom_reffile.inc
```diff
@@ -294,6 +302,10 @@ Scalar PHOTMJSR and PHOTUJA2 values are stored in primary header keywords
 in the MIRI MRS PHOTOM reference files and are copied into the science
 product header by the photom step.
 
+The TIMECOEFF_CH tables contain the parameters to correct the MRS a time-dependent
+throughput loss. If these tables do not exist in the reference file, then the MIRI
+MRS time dependent correction is skipped.
```
hbushouse, line 307:
```suggestion
The TIMECOEFF_CH tables contain the parameters to correct the MRS time-dependent
throughput loss. If these tables do not exist in the reference file, then the MIRI
MRS time-dependent correction is skipped. 
```

hbushouse (approved):

hbushouse:
Looks like this is ready to go. Merging.

PR #7604, opened 2023-06-13
Revert "JP-3229: <Fix a bug> added the user-supplied flat reference file to the log"

hbushouse:
Reverts spacetelescope/jwst#7592

Rats! I *knew* we should've run regression tests before merging this, but I assumed that simple changes to logging would be safe. Shame on me!

Line 109 in flat_field_step.py needs to be changed to include a format specifier, i.e.
`self.log.info('Using flat field reference file: %s', flat_ref_file)`
otherwise it results in a crash.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,8 +54,6 @@ flat_field
 
 - Refactored NIRSpec 1D flat interpolation for improved performance. [#7550]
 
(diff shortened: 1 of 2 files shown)


penaguerrero:
done

hbushouse:
@penaguerrero Has a new PR been submitted with the updated code? I don't see one.

PR #7584, opened 2023-05-09
JP 3231: address warnings in tests

zacharyburnett:
Resolves [JP-3231](https://jira.stsci.edu/browse/JP-3231)

This PR addresses several warnings in the unit and regression tests

**Checklist for maintainers**
- [ ] ~added entry in `CHANGES.rst` within the relevant release section~
- [ ] ~updated or added relevant tests~
- [ ] ~updated relevant documentation~
- [ ] added relevant milestone
- [ ] added relevant label(s)
- [ ] ran regression tests, post a link to the Jenkins job below.
      [How to run regression tests on a PR](https://github.com/spacetelescope/jwst/wiki/Running-Regression-Tests-Against-PR-Branches)
- [ ] Make sure the JIRA ticket is [resolved properly](https://github.com/spacetelescope/jwst/wiki/How-to-resolve-JIRA-issues)

PR #7605, opened 2023-06-13
inspect and shadow stdatamodels.jwst.datamodels

braingram:
Currently adding a new model requires adding the model to stdatamodels and adding stub files and updating `jwst.datamodels.__init__.py` in jwst. This coordination can cause difficulties in testing as each package uses the other in it's test suite.

For example:
https://github.com/spacetelescope/stdatamodels/pull/164
adds `stdatamodels.jwst.datamodels.outlierifuoutput` and `OutlierIFUOutputModel`
This causes the jwst downstream tests to fail because this model and submodule do not yet exist in `jwst.datamodels`.
The corresponding jwst PR https://github.com/spacetelescope/jwst/pull/7590 will need to a `outlierifuoutput` stub file to `jwst.datamodels` and add `OutlierIFUOutputModel` to `jwst.datamodels.__all__` to fix this issue. The jwst CI for this example PR will fail until a version of stdatamodels is released that contains the PR adding the model.

To hopefully avoid needing to change both jwst and stdatamodels when a new model is added this PR introduces dynamic inspection of `stdatamodels.jwst.datamodels` at the time of import to allow `jwst.datamodels` to:
- contain all models included in `stdatamodels.jwst.datamodels.__all__`
- contain all submodules imported in `stdatamodels.jwst.datamodels`
- expose all submodules from `stadatamodels.jwst.datamodels` at `jwst.datamodels` (by adding entries to `sys.modules`)

These changes should allow updates to stdatamodels.jwst.datamodels without needing to also add 'stub' files in jwst.datamodels while maintaining the previous api.

Even though this PR removes the stub files (for example `jwst.datamodels.apcorre`) it preserves importing submodules from `jwst.datamodels` as in the following example:
```python
from jwst.datamodels.apcorr import MirImgApcorrModel
print(MirImgApcorrModel)  # stdatamodels.jwst.datamodels.apcorr.MirImgApcorrModel
```
by adding the stdatamodels submodules (like `stdatamodels.jwst.datamodels.apcorr`) to `sys.modules` using the jwst submodule name (`jwst.datamodels.apcorr`).

~Note that the dynamic inspection of stdatamodels.jwst.datamodels introduced in this PR re-adds the deprecated model classes removed from jwst in (https://github.com/spacetelescope/jwst/pull/7584) because these models still exist in stdatamodels. The PR mentions that this was to clean up warnings in the CI which should be inspected on this PR to see if they are reintroduced. One option to fix this partial reversion would be to remove the models from stdatamodels. Another option would be to specifically ignore those deprecated classes.~ see https://github.com/spacetelescope/jwst/pull/7605#issuecomment-1589940193

**Checklist for maintainers**
- [ ] added relevant label(s)
      https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/741/


braingram:
This PR needs to have the regression tests run. I will try to start these when I see the server is free.

braingram:
I converted this back to draft to see if I can get rid of the warnings for the deprecated models/modules.

nden (approved):
lgtm

PR #7660, opened 2023-06-28
remove uses of stdatamodels.jwst.datamodels.schema

braingram:
[stdatamodels.jwst.datamodels.schema](https://github.com/spacetelescope/stdatamodels/blob/master/src/stdatamodels/jwst/datamodels/schema.py) is an out-of-date duplicate of [stdatamodels.schema](https://github.com/spacetelescope/stdatamodels/blob/master/src/stdatamodels/schema.py). Uses of the stdatamodels.jwst.datamodels.schema were converted to stdatamodels.schema and the automatic shadowing of stdatamodels.jwst.datamodels.schema to jwst.datamodels.schema was removed.

**This removes jwst.datamodels.schema from the api.**

Regression tests run at: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/794/

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,11 +1,17 @@
 1.11.4 (unreleased)
 ===================
 
+datamodels
+----------
+
+- Remove ``jwst.datamodels.schema`` in favor of ``stdatamodels.schema`` [#7660]
+
 flat_field
 ----------
 
 - Modify test_flatfield_step_interface test to prevent it from causing
   other tests to fail [#7752]
+
 
 1.11.3 (2023-07-17)
 ===================
@@ -64,8 +70,6 @@ datamodels
 
 - Added two new header keywords to track the rate of cosmic rays and snowball/showers
   [#7609, spacetelescope/stdatamodels [spacetelescope/stdatamodels#173]
-
-- Remove ``jwst.datamodels.schema`` in favor of ``stdatamodels.schema`` [#7660]
 
 jump
 ----
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -61,6 +61,8 @@ datamodels
 - Added two new header keywords to track the rate of cosmic rays and snowball/showers
   [#7609, spacetelescope/stdatamodels [spacetelescope/stdatamodels#173]
 
+- Remove ``jwst.datamodels.schema`` in favor of ``stdatamodels.schema`` [#7660]
+
 jump
 ----
 
--- a/jwst/datamodels/__init__.py
+++ b/jwst/datamodels/__init__.py
@@ -29,7 +29,7 @@ _jwst_modules = ["container", "source_container"]
 _jwst_models = ["ModelContainer", "SourceModelContainer"]
 
 # Deprecated modules in stdatamodels
-_deprecated_modules = ['drizproduct', 'multiprod']
+_deprecated_modules = ['drizproduct', 'multiprod', 'schema']
 
 # Deprecated models in stdatamodels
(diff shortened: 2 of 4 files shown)


braingram:
devdeps failures should be fixed by either/both:
https://github.com/spacetelescope/stdatamodels/pull/184
https://github.com/asdf-format/asdf/pull/1594

braingram:
Regression tests show only usual F/D/SFLAT keyword errors. I'm puzzled by these :-/

braingram:
Closing and reopening to trigger CI to pick up stdatamodels changes which should fix devdeps failures.

hbushouse:
> Regression tests show only usual F/D/SFLAT keyword errors. I'm puzzled by these :-/

Aren't we all! Not only are the keywords for those ref files not getting put into the output files, it's obvious that they weren't even applied during the flat-field step, because the data show differences too. Sure would like to figure out what's causing this. It usually comes from the same hand-full of tests, but it's never exactly the same ones on any given regtest run.

braingram:
It's good to know I'm not the only one! :)

I've had numerous fruitless suspicions today including:
- maybe it's something with the test suite execution but...
  - running test in parallel vs serial doesn't seem to matter locally
  - running tests with coverage doesn't seem to matter
  - forcing the order of some of the failing tests doesn't seem to matter (I was combining them with a test that runs the `test_nirspec_verify` `run_image2` fixture which [overrides the f/s/dflats](https://github.com/spacetelescope/jwst/blob/d1a60ee11df7bbeb3b7ea7075ce933b5756a3faa/jwst/regtest/test_nirspec_verify.py#L48C1-L48C1) as 'N/A' (whereas the default value for the argument is None)
- maybe it's something with crds and/or the context (the truth files are 1097 the new files are 1100)

Is there a way to see and/or capture the log output for a test/run on plwishmaster? The inability to replicate the issue locally is making this a hard nut to crack.

hbushouse:
Yes, you can capture the processing log by using the "-s" pytest argument in the test parameters. I've tried this in the past and whenever I do, it all runs just fine (because it knows someone's watching ...). I've also tried the single process vs. multi-processing comparisons by using the "-n 1" pytest option to limit the Jenkins run to only 1 core (instead of many cores). It makes the result WORSE!  Yet more tests then end up with this error. But of course it never happens when running the tests locally on one of our personal systems. So it's highly mind-boggling.

braingram:
Thanks! That is all super useful. I will give some of those options a try.

braingram:
I have hope this will fix the regression test failures: https://github.com/spacetelescope/jwst/pull/7752

nden (approved):
LGTM

CHANGES.rst
```diff
@@ -65,6 +65,8 @@ datamodels
 - Added two new header keywords to track the rate of cosmic rays and snowball/showers
   [#7609, spacetelescope/stdatamodels [spacetelescope/stdatamodels#173]
 
+- Remove ``jwst.datamodels.schema`` in favor of ``stdatamodels.schema`` [#7660]
```
hbushouse, line 68: This entry will need to be moved to the top of the log, now that it's not part of the recent releases that've happened.

braingram: Thanks! I updated the changelog moving this entry to the top, 'unreleased' version.

braingram:
Regtests running at: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/806/

hbushouse (approved):
Run 806 looks good.

PR #7709, opened 2023-07-07
JP-3300: Update spec2 docs on order of operations for SOSS

### LINKED ISSUE #7707: Change to spec2 processing flow for NIRISS SOSS not documented
_Issue [JP-3300](https://jira.stsci.edu/browse/JP-3300) was created on JIRA by [Tyler Pauly](https://jira.stsci.edu/secure/ViewProfile.jspa?name=tpauly):_

User encountered unexpected behavior because <https://jira.stsci.edu/browse/JP-2511> did not include a documentation update on the modified spec2 processing flow. Additionally, there is duplicated documentation between the pipeline RTD and JDox, and the JDox page is also out of date - link: <https://jwst-docs.stsci.edu/jwst-science-calibration-pipeline-overview/stages-of-jwst-data-processing/calwebb_spec2>

Re: INC0191219

tapastro:
Resolves [JP-3300](https://jira.stsci.edu/browse/JP-3300)

Closes #7707 

This PR addresses a change in step processing order for NIRISS SOSS data; photom is run after 1D spectral extraction for this mode.

**Checklist for maintainers**
- [x] added relevant label(s)


hbushouse:
The CI doc build test failed because the updates to the table have left the two modified rows out of alignment with all the other rows.

hbushouse:
The table now builds correctly, but when looking at the new version I noticed that this new footnote is noted on the check marks in the SOSS column, while a similar note about different ordering of steps is done via footnote superscripts added to the names of the steps themselves (in column 1 of the table). I wonder if for consistency we should move these new marks to the step names, like the others. That would also prevent inflation of the width of the SOSS column check marks.

tapastro:
You can override if you like, but it was a conscious decision - and I'd argue that the first footnote should be marked in the columns as well, as they're easy to miss otherwise. The second footnote makes sense in the step names, but I'd argue those two steps should maybe not have check marks as-is, due to them not running unless explicitly "un-skipped".

Do you prefer I move it to the step name column?

hbushouse:
> You can override if you like, but it was a conscious decision - and I'd argue that the first footnote should be marked in the columns as well, as they're easy to miss otherwise. The second footnote makes sense in the step names, but I'd argue those two steps should maybe not have check marks as-is, due to them not running unless explicitly "un-skipped".
> 
> Do you prefer I move it to the step name column?

OK, you've convinced me that it's OK to leave in the checkmark columns

hbushouse (approved):
Looks good.

PR #7711, opened 2023-07-07
JP-3250: Exclude entire var if any NaN present

### LINKED ISSUE #7647: NaNs in the flat variance array contaminate the x1d combined error estimate
_Issue [JP-3250](https://jira.stsci.edu/browse/JP-3250) was created on JIRA by [Alicia Canipe](https://jira.stsci.edu/secure/ViewProfile.jspa?name=acanipe):_

From Help Desk ticket INC0189669 (link at the bottom for reference): 

The user noticed the FLUX_ERROR column in an x1d product was an array of NaNs. Howard pointed out that the cal files have all zero values in the FLAT_VARIANCE 2d array, which then propagates through to the x1d files created by the extract_1d step. It appears that the flat variance array is all zero because all of the ERR arrays in the current flat-field reference files for the NIRSpec fixed-slit mode (using the prism) are also zero-filled. 

So two things would need to be updated to address the user's issue:
 # Find and fix the fact that the NaN values in the flat variance array are contaminating the computation of the combined error estimate, which is composed of Poisson, read noise, and flat variances, so we could then at least get error estimates that represent the Poisson and read noise only. 
 # The NIRSpec instrument team would need to compute estimated variance data for the flats and store them in the flat-field reference files used by the calibration pipeline. This should probably be captured separately in a CRDS ticket. 

 

Help desk link: <https://stsci.service-now.com/nav_to.do?uri=%2Fincident.do%3Fsys_id%3D5d12d758477ea190289c4f32736d4364%26sysparm_stack%3Dincident_list.do%253fsysparm_query%253dactive%253dtrue%26sysparm_view%3DJWST%26sysparm_view_forced%3Dtrue>

tapastro:
Resolves [JP-3250](https://jira.stsci.edu/browse/JP-3250)
Closes #7647 

This PR addresses an issue found when a variance array contains NaN values - the sum with the other variance arrays leads to NaN values in the error arrays. The current behavior will propagate any NaNs in a variance array to the final error array.

The approach in this PR is to fully exclude a variance array if at least one NaN is present - this is to prevent the error array from having a variable number of contributions. Alternatively, one could let the error array exclude the NaN values from summation for individual variance entries, leading to the error having 0-3 variance contributions depending on the number of NaNs present.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,12 @@ documentation
 
 - Fix minor formatting typos in associations docs. [#7694]
 
+extract_1d
+----------
+
+- Prevent propagation of NaNs in variance arrays to error array by excluding entire
+  variance array if at least one NaN is present. [#7711]
+
 tweakreg
 --------
 
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -3766,9 +3766,22 @@ def create_extraction(extract_ref_dict,
 
         del temp_flux
 
-        error = np.sqrt(f_var_poisson + f_var_rnoise + f_var_flat)
-        sb_error = np.sqrt(sb_var_poisson + sb_var_rnoise + sb_var_flat)
-        berror = np.sqrt(b_var_poisson + b_var_rnoise + b_var_flat)
+        # Run through each variance array and check for NaNs - if present, exclude from quadrature sum.
+        f_var_list = []
+        sb_var_list = []
+        b_var_list = []
+        for var_list in [[f_var_poisson, f_var_rnoise, f_var_flat, f_var_list],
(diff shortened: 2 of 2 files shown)


izkgao:
Would `np.nansum` be a better option?

```python
import numpy as np
np.nansum([np.nan, 1.0, 2.0])
# 3.0
np.nansum([np.nan, np.nan, np.nan])
# 0.0
```

hbushouse:
Handling of NaNs in sci, err, and variance arrays has had lots of updates and fixes since this PR was originally created and the original ticket, JP-3250, has been declared to be done, given all the other changes. So I'm closing this PR, as I don't believe it's needed anymore.

PR #7738, opened 2023-07-18
JP-2928: Updating tests with place holders for new files.

kmacdonald-stsci:
Resolves [JP-2928](https://jira.stsci.edu/browse/JP-2928)

This PR addresses some regression tests needed changing according to JP-2928.

**Checklist for maintainers**
- [ ] added relevant label(s)

### DIFF
--- a/jwst/regtest/test_nircam_tsgrism.py
+++ b/jwst/regtest/test_nircam_tsgrism.py
@@ -12,6 +12,8 @@ def run_pipelines(jail, rtdata_module):
     rtdata = rtdata_module
 
     # Run tso-spec2 pipeline on the _rateints file, saving intermediate products
+
+    # rtdata.get_data("nircam/tsgrism/jw01366002001_04103_00001-seg001_nrcalong_rateints.fits")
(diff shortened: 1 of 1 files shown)


melanieclarke:
@kmacdonald-stsci - do you remember what was going on with this old PR?  It looks like you proposed a new data set to replace pre-flight data, but we never followed through with it.  Was there a reason why not, or did we just not get around to it?

kmacdonald-stsci:
> @kmacdonald-stsci - do you remember what was going on with this old PR? It looks like you proposed a new data set to replace pre-flight data, but we never followed through with it. Was there a reason why not, or did we just not get around to it?

I believe this was part of the effort to change regression tests to using actual data instead of simulated data.  I was asked to change some regression tests to update file names reflecting that, but I didn't have authority to update the data itself, so it never got approved or merged.  I'm certain this can be closed.

melanieclarke:
Okay, thanks, I will close this one and use the suggested file as a starting point for updating the nircam/tsgrism data later.

PR #7484, opened 2023-03-09
(JP-3029) Update ramp fitting to calculate separate readnoise variance

dmggh:
Resolves [JP-3029](https://jira.stsci.edu/browse/JP-3029)

Closes #

This PR calculates a readnoise variance that can be used by the resampling step, using UNDERSAMP-groups that have been flagged by the undersampling_correction step.   This PR supersedes PR #7443, which will be closed.

**Checklist for maintainers**
- [x] added entry in `CHANGES.rst` within the relevant release section
- [X] updated or added relevant tests
- [X] updated relevant documentation
- [X] added relevant milestone
- [X] added relevant label(s)
- [X] ran regression tests, post a link to the Jenkins job below.
      [How to run regression tests on a PR](https://github.com/spacetelescope/jwst/wiki/Running-Regression-Tests-Against-PR-Branches)
- [ ] Make sure the JIRA ticket is [resolved properly](https://github.com/spacetelescope/jwst/wiki/How-to-resolve-JIRA-issues)

PR #7740, opened 2023-07-19
(JP-3299) Enhancements for undersampling_correction step

### LINKED ISSUE #7703: Enhancements for undersampling_correction step
_Issue [JP-3299](https://jira.stsci.edu/browse/JP-3299) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

The NIRISS team is requested changes and enhancements to the baseline version of the undersampling_correction that was added to the Detector1 pipeline (see JP-3015 and JP-3029). The detailed requirements for the step are listed at <https://outerspace.stsci.edu/display/JWSTCC/Description+of+Undersampling+Correction+step> , which includes descriptions of the requested updates relative to the baseline version that's already in place.

dmggh:
Resolves [JP-3299](https://jira.stsci.edu/browse/JP-3299)

Closes #7703 

This PR addresses these enhancements requested for the undersampling_correction step:
1. Change the default signal threshold to 25000 ADU
2. Pixels adjacent to pixels flagged in the first step shall be flagged with the same flag value in the same group(s) as those of the pixels flagged in the first step.
3. In the calwebb_detector1 pipeline, switch the step execution order so that the undersampling correction step shall be executed just prior to the jump detection step. 
4. The jump detection step shall be adjusted in order to account for the UNDERSAMP flag, by ignoring the groups flagged with UNDERSAMP in the determination of jumps within the integration. This functionality is already followed in the code merged earlier.
5. The ramp fitting step shall be adjusted in order to account for the UNDERSAMP flag for the slope fitting, and in the calculation of the VAR_NOISE parameter.  Both of these functionalities are already followed in the code merged earlier.

**Checklist for maintainers**
- [x] added relevant label(s)


jwst/undersampling_correction/undersampling_correction.py
```diff
@@ -87,18 +91,50 @@ def flag_pixels(data, gdq, signal_threshold):
...
-                np.bitwise_or(gdq[ii_int, ii_grp:, yy, xx], dqflags.group['UNDERSAMP']
-                              | dqflags.group['DO_NOT_USE'])
+            if len(wh_set_flag) > 0 and len(xx) > 0:
+
+                gdq[ii_int, ii_grp:, yy, xx] = \
+                    np.bitwise_or(gdq[ii_int, ii_grp:, yy, xx],
+                                  dqflags.group['UNDERSAMP']
+                                  | dqflags.group['DO_NOT_USE'])
+
+                for ii in range(0, len(xx)):
```
nden, line 112: For each group in each integration this is iterating over every pixel above the threshold.
This is very inefficient. Can it be rewritten using arrays?

jwst/undersampling_correction/undersampling_correction.py
```diff
@@ -87,18 +91,50 @@ def flag_pixels(data, gdq, signal_threshold):
...
-                              | dqflags.group['DO_NOT_USE'])
+            if len(wh_set_flag) > 0 and len(xx) > 0:  # there are exceedances
+                gdq[ii_int, ii_grp:, yy, xx] = \
+                    np.bitwise_or(gdq[ii_int, ii_grp:, yy, xx],
+                                  dqflags.group['UNDERSAMP']
+                                  | dqflags.group['DO_NOT_USE'])
+
+                # Set the same flags for the 4 nearest neighbors
+                wh_west = np.where(xx > 0)  # pixels whose western neighbors to be flagged
+                if len(wh_west[0]) > 0:
```
hbushouse, line 113: Given that a bitwise_or is being used in all of these operations and hence there's no danger of flagging a pixel twice with the same flag, I wonder if these four operations (one for each neighboring pixel) could be simplified to be done in only 2 operations - one for horizontal ("east", "west") neighbors and one for vertical ("north", "south") neighbors. Something like:

np.bitwise_or(gdq[ii_int, ii_grp, yy, xx-1:xx+1], dqflags.group ...)    # for the 3 horizontal pixels
np.bitwise_or(gdq[ii_int, ii_grp, yy-1:yy+1, xx], dqflags.group ...)    # for the 3 vertical pixels

Yes, the center (original) pixel gets operated on multiple times, but it doesn't hurt anything (because of the "or") and it greatly simplifies the code.

dmggh:
This latest commit ( Thu Aug 10 23:16 ) satisfies these updates to the docs :
1  Change the default signal threshold to 25000 ADU .

2  Pixels adjacent to pixels flagged in the first step shall be flagged with the same flag value in the same group(s) as those of the pixels flagged in the first step.

3  In the calwebb_detector1 pipeline, switch the step execution order so that the undersampling correction step shall be executed just prior to the jump detection step. 

These are the updates for all 3 the initial enhancements given and still pending when the ticket was created on 7/6/23. There were 2 other enhancements requested regarding modifications to the jump step and the ramp fitting steps; those enhancements had been previously implemented JP-3029  (GH #7484 ).

dmggh:
Converted from 'draft' to for review. Code committed completes the enhancements initially given in the ticket. The step name change will be handled in 2 separate PRS.

jwst/undersampling_correction/undersampling_correction.py
dmggh: I couldn't refactor this way, but I did improve the efficiency by replacing calls of np.where() using arrays.

jwst/undersampling_correction/undersampling_correction.py
```diff
@@ -69,6 +69,10 @@ def flag_pixels(data, gdq, signal_threshold):
 
     lowest_exc_1d = np.zeros(num_pix) + n_grps
 
+    ncols = data.shape[3]
+    nrows = data.shape[2]
+
+    # Loop over all groups to locate exceedances
     for ii_int in range(n_ints):
```
hbushouse, line 76: I know this isn't part of the current updates in this PR, but I'm wondering if there's a way to avoid all of these loops over integrations and groups (which are slow in python). Why not just take advantage of the natural looping/broadcasting ability of python and numpy to do it all in one fell swoop. For example, can't all of the initial pixels over the limit (in all integrations and all groups) just be flagged all at once using something simple like:

gdq[data > signal_threshold] = np.bitwise_or(gdq[data > signal_threshold], dqflags.group['UNDERSAMP'] | dqflags.group['DO_NOT_USE')

There's no need to check for and exclude pixels that already have UNDERSAMP set, because the logical OR is not going to double up any of the flags. The only additional check that I see as possibly necessary is to exclude pixels that already have DO_NOT_USE set from a previous step, so that the later steps like jump and ramp_fit can sort out which ones are "normal" DO_NOT_USE and which are DO_NOT_USE because of the undersampling correction (which requires special handling in ramp_fit).

So some playing around suggests that the whole thing could be accomplished using something like:

undersamp_pix = (data > signal_threshold) == (gdq != dqflags.group['DO_NOT_USE')
gdq[undersamp_pix] = np.bitwise_or(gdq[undersamp_pix], dqflags.group['UNDERSAMP'])

The first statement sets up an array that's True only where a pixel is above the threshold AND not already set to DO_NOT_USE, and then that array is used to set the new flag only for those pixels, over all 4 dimensions of the data/dq arrays at once. No looping anywhere.

dmggh: I will take a look at this. However, also to consider is that once a pixel in an integration has been  flagged due to being an exceedance in a particular group, a requirement is that all later groups for that pixel and integration must ALWAYS  also be flagged, even if they are not exceedances ( so if even data>signal_threshold is not true), which would not hold for your undersamp_pix expression.

dmggh: Ok,  assuming that the signal will always increase in later groups will simplify the code, as all groups can be considered independent of each other.

nden: If that was not the case we have bigger issues to solve :-)

dmggh:
Update unit tests coming soon. The existing unit tests will fail until then.

dmggh:
The unit tests are now updated, and pass using the new code updates.

docs/jwst/undersampling_correction/description.rst
```diff
@@ -12,11 +12,15 @@ a pixel. When the peak pixels of such stars approach the saturation level, they
 significant :ref:`charge migration <charge_migration>`:
 the spilling of charge from a saturated pixel into its neighboring pixels. This charge migration 
 causes group-to-group differences to decrease significantly once the signal level is greater than 
-~30,000 ADU. As a result, the last several groups of these ramps get flagged by the ``jump`` step. 
-The smaller number of groups used for these pixels in the ``ramp_fitting`` step results in them having 
+~25,000 ADU.  As a result, the last several groups of these ramp will be flagged by the
```
hbushouse, line 15:
```suggestion
~25,000 ADU.  As a result, the last several groups of these ramps will be flagged by the
```

jwst/undersampling_correction/undersampling_correction.py
```diff
@@ -46,7 +51,7 @@ def undersampling_correction(input_model, signal_threshold):
 def flag_pixels(data, gdq, signal_threshold):
     """
     Flag first group in each ramp that exceeds signal_threshold as UNDERSAMP and DO_NOT_USE,
```
hbushouse, line 53:
```suggestion
    Flag each group in each ramp that exceeds signal_threshold as UNDERSAMP and DO_NOT_USE,
```

jwst/undersampling_correction/undersampling_correction.py
```diff
@@ -61,44 +66,79 @@ def flag_pixels(data, gdq, signal_threshold):
...
+    nrows = data.shape[2]
+
+    new_gdq = gdq.copy()   # Updated gdq
+
+    # Flag all exceedances with UNDERSAMP and NO_NOT_USE
+    undersamp_pix = (data > signal_threshold) == (gdq != DNU)
+    new_gdq[undersamp_pix] = np.bitwise_or(new_gdq[undersamp_pix], UNSA)
+    new_gdq[undersamp_pix] = np.bitwise_or(new_gdq[undersamp_pix], DNU)
+
+    # Reset groups previously flagged as DNU
```
hbushouse, line 83: Not sure I understand why this needs to be done. `new_gdq` and `gdq_orig` are both copies of `gdq`, so they should be identical. Then `new_gdq` has additional pixels flagged as DNU+UNSA, but only if they were not already flagged as DNU. So any pixels originally flagged (on input) as DNU should still be there in `new_gdq`, won't they? So why do they need to be copied over again from `gdq_orig`, which is equal to `gdq`?  Maybe I'm missing something.

jwst/undersampling_correction/undersampling_correction.py
```diff
@@ -61,44 +66,79 @@ def flag_pixels(data, gdq, signal_threshold):
...
-    return gdq
+    ncols = data.shape[3]
+    nrows = data.shape[2]
+
+    new_gdq = gdq.copy()   # Updated gdq
+
+    # Flag all exceedances with UNDERSAMP and NO_NOT_USE
+    undersamp_pix = (data > signal_threshold) == (gdq != DNU)
+    new_gdq[undersamp_pix] = np.bitwise_or(new_gdq[undersamp_pix], UNSA)
+    new_gdq[undersamp_pix] = np.bitwise_or(new_gdq[undersamp_pix], DNU)
```
hbushouse, line 81: Could these 2 operations be combined in one statement by using `UNSA | DNU` as a combined last argument to `bitwise_or` (as is done below in some other places)?

hbushouse (commented):
I like that all the loops are now gone in the core part of the flagging algorithm. Just a few questions about other things.

hbushouse:
I don't recall whether this step is currently set to be run when processing any NIRISS exposures, e.g. by a param ref file in CRDS. Hence I'm not sure if running a regression test would have any effect, because it might not get executed at all. @dmggh do you recall whether it's turned on by any param ref files?

docs/jwst/undersampling_correction/description.rst
dmggh: Done

jwst/undersampling_correction/undersampling_correction.py
dmggh: Done

jwst/undersampling_correction/undersampling_correction.py
dmggh: Yes, done

jwst/undersampling_correction/undersampling_correction.py
hbushouse: Ah right, I understand about the neighbors and you're right that we don't want to flag a neighbor if it was already set to DNU. But isn't this operation taking place *before* the attempt to flag neighbors? I would think it would need to be done *after* the neighbors get flagged, i.e. to undo any damage caused by the neighbor flagging.

dmggh: As the code is now, it us done both before any neighbors are flagged (which I agree can be deleted) and again after neighbors are flagged (which should remain).

dmggh:
@hbushouse - Currently the step is not turned on by param reference files.

hbushouse:
CI tests are failing because of mismatches in the unit test test_undersampling_correction.py::test_flag_neighbors

hbushouse (approved):
OK, I think this all looks good now and all the CI tests are passing. No point in running regtests, because the step isn't run by any of them.

PR #7744, opened 2023-07-19
Use current date to fetch current reference files during unit testing

tapastro:
This PR addresses some bespoke header generation during unit testing - fixed dates were being provided, which could prevent testing against current reference files due to CRDS USEAFTER dates.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,11 @@
 1.11.4 (unreleased)
 ===================
 
--
+general
+-------
+
+- Use current date instead of a fixed date while creating headers for
+  unit testing to ensure testing against current reference files [#7744]
 
 1.11.3 (2023-07-17)
 ===================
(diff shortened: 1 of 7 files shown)


nden:
I don't expect these to work. Nirspec tests use the original instrument model and compare to the results of the IDT team pipeline using the same model, to prove we got identical results. A similar test can/should be done if we have results for the new instrument model. These have been requested in JP-3249. 
The situation with MRS tests is similar. Perhaps these tests are not needed any more.

hbushouse:
I'm not sure how I feel about, in terms of the general philosophy. On the one hand, it's useful to have tests use fixed inputs and outputs, so that you can verify that new versions of the code are still giving the same answers as the old versions. At the same time, it would be useful to test the latest ref files to make sure they work properly and that our code works properly with them. The problem with the latter, as @nden already mentioned, is that we often don't have a good way of knowing whether the results from new ref files are correct or not. When it comes to WCS-related ref files and WCS-related code results, we often need to rely on the instrument teams to verify the new results, which historically is difficult to obtain.

melanieclarke:
@tapastro - it sounds like this change is not desirable.  Does this need more discussion or can we close this PR?

tapastro:
We can close it - I think updating our test data should cover most of the issues this was trying to address.

PR #7761, opened 2023-07-25
JP-3317: Use same logic for resample limits for SCI, ERR, and VAR* images

mcara:
Resolves [JP-3317](https://jira.stsci.edu/browse/JP-3317)

Resample step has two branches: one for resampling of the science image (SCI extension in the FITS file) and one for resampling error and variance arrays (ERR, VARIANCE_* extensions). The logic in the branch for processing the science image ignores the bounding box from the input images' WCS while the logic for processing ERR, VAR* arrays does take it into account.

Both science and error images should be processed (resampled) using the same limits for input image pixels.

This PR solves this issue by using the same algorithm for computing resample limits (in input image coordinates) to science, error, and variance arrays from input image shape and input WCS' bounding box.

**Checklist for maintainers**
- [x] added relevant label(s)

Regression test is running here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/810


mcara:
In addition, the logic for computing resample limits was different (by 1 pixel) depending whether bounding box was used or not. For example, if input WCS did not have a bounding box then the algorithm would use image's shape to infer `xmax` and `ymax`:

```python
ymax, xmax = image.shape
```

For an image of size `nx=100` and `ny=120`, this would give `xmax=100` and `ymax=120`. However, if a bounding box for such an image were to be computed, it would normally be (assuming all input pixels have valid/defined WCS transformations) computed as:

```python
bounding_box = ((-0.5, image.shape[1] - 0.5), (-0.5, image.shape[0] - 0.5))
```
see, for example, https://github.com/spacetelescope/jwst/blob/7556988f110205042334d00c1c9d351c4bfb7773/jwst/assign_wcs/util.py#L916-L918

Attempting to compute `xmax` and `ymax` using current logic from the `bounding_box` would result in `xmax=99` and `ymax=119`.

This PR also addresses this issue.

mcara:
Regression test https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/816/ is being run to demonstrate that commit 675cf4a9 effectively restores the old (incorrect) behavior (regression test should pass),

jwst/resample/resample_utils.py
```diff
@@ -303,3 +303,28 @@ def decode_context(context, x, y):
         )
 
     return idx
+
+
+def _xy_range_from_bbox(xmin, xmax, ymin, ymax, shape, bbox=None):
+
+    if (xmin is xmax is ymin is ymax) and xmin in (0, None):
```
nden, line 310: The default value of `xmin`, `xmax`, `ymin`, `ymax` is 0. Why is this check for `None`?
Shouldn't the default be `None`? 
Or why not use an existing function to be consistent?
 https://github.com/spacetelescope/jwst/blob/675cf4a9c677c472072cfa2ea45bc28f50636a2a/jwst/assign_wcs/util.py#L902

jwst/resample/resample_utils.py
```diff
@@ -303,3 +303,28 @@ def decode_context(context, x, y):
...
+        xmin = ymin = 0
+        ymax, xmax = np.subtract(shape, 1, dtype=int)
+
+        if bbox:
+            ((x1, x2), (y1, y2)) = bbox
+            if x1 > x2:
+                x1, x2 = x2, x1
+            if y1 > y2:
+                y1, y2 = y2, y1
```
nden, line 319: Is there any evidence this is possible, i.e. the starting value is larger than the ending value?
astropy.modeling raises a warning if this is the case: https://github.com/astropy/astropy/blob/main/astropy/modeling/bounding_box.py#L84

jwst/resample/resample_utils.py
```diff
@@ -303,3 +303,28 @@ def decode_context(context, x, y):
...
+
+            # use 0.5+/-eps to avoid having edge pixels with flux coming from
+            # a tiny part of input pixels (with low weight). eps can be set
+            # safely to 0.
+            eps = 0.000001
+            xmin = max(0, int(x1 + (0.5 + eps)))
+            ymin = max(0, int(y1 + (0.5 + eps)))
```
nden, line 326: Is the goal of the above two lines to compute the center of the pixel? If so you may want to use https://github.com/spacetelescope/gwcs/blob/master/gwcs/utils.py#L54

In general there's no guarantee that the bounding_box values use +/- 0.5 of the center of the pixel.

I am curious - does drizzle require the center of the pixels to be used?

jwst/resample/resample_utils.py
```diff
@@ -303,3 +303,28 @@ def decode_context(context, x, y):
...
+            xmax = min(xmax, int(x2 + (0.5 - eps)))
+            ymax = min(ymax, int(y2 + (0.5 - eps)))
+
+    return xmin, xmax + 1, ymin, ymax + 1
```
nden, line 330: Also the logic in this function seems inverted to me. I would think `bbox` should be used by default and overwritten by user values if supplied but it's the opposite - if `bbox` is available it's always used. Is this what is desired?

jwst/resample/resample_utils.py
mcara: > Or why not use an existing function to be consistent?

Because this is the "inverse" function of `wcs_bbox_from_shape()`.

> Why is this check for None. Shouldn't the default be None?

I don't know. I didn't write this code. I do not think there is need to allow `None` but I also do not see harm in it. `cdrizz` uses the entire input image when `xmin=xmax=ymin=ymax=0` (both in `drizzle` and in `drizzlepac`). It does not accept `None` even though, IMO, it would be a more natural way to say "no limits - use the entire image". However, the way this is done here is inconsistent (again) with `dodrizzle()` in `gwcs_drizzle` module - that one uses 0 exclusively.

jwst/resample/resample_utils.py
mcara: I have no evidence of this. I preserved the existing functionality: https://github.com/spacetelescope/jwst/pull/6041/files#diff-8726cc4bf91b91dd490c35ce61e58c0c840b249b1868fa50cd25a012b13f3d98R383-R386

I do not know why it was implemented this way. Given that `astropy` does not support "inverted" bounding boxes, I agree this should be removed.

jwst/resample/resample_utils.py
mcara: > I am curious - does drizzle require the center of the pixels to be used?

Used for what? Not as far as I know.

> In general there's no guarantee that the bounding_box values use +/- 0.5 of the center of the pixel.

That's OK.

> Is the goal of the above two lines to compute the center of the pixel? If so you may want to use https://github.com/spacetelescope/gwcs/blob/master/gwcs/utils.py#L54

No. The goal is to compute the first and last row and column that includes the bounding box.

jwst/resample/resample_utils.py
mcara: Both `dodrizzle()` in `gwcs_drizzle` and `ResampleData.drizzle_arrays()` in the `resample` module accept `xmin`, `xmax`, `ymin`, `ymax`. This is the only way to indicate what part of the input image should be used. When all `x/y min/max` are 0 => resample the entire input image. That's all. Unfortunately, all these functions are called with these default parameters (`dodrizzle` ultimately is called from `add_image()` using defaults) => resample the entire image.

Looking at https://github.com/spacetelescope/jwst/pull/6041/files#diff-8726cc4bf91b91dd490c35ce61e58c0c840b249b1868fa50cd25a012b13f3d98R248 I see that default values have been changed (but only when resampling variance arrays!) from `0` to `None` without documenting this. Looking at the code, if `x/y min/max` are all `0` then they would just get passed to `cdriz` "as is" (meaning use the entire image) but when `x/y min/max` are all `None`, then the code as implemented in https://github.com/spacetelescope/jwst/pull/6041 would use the bounding box.

IMO this is a bad choice of default parameters (both for `cdriz` and `dodrizzle`, `drizzle_arrays`). I would have preferred `None` to mean "no limits" (use entire image). If there are limits - they should be computed at the top level by the caller, i.e., `add_image`.

The problem that I was trying to fix was to have the same behavior both for resample variance and resample SCI image. Now I realize that because `dodrizzle` has defaults set to 0 (for x and y limits), it is still not using the WCS bounding box as it does when resampling variance arrays.

But before I fix this, can we decide whether it is OK to modify API for these functions (`dodrizzle` and `drizzle_arrays`) to have the same defaults (undo changes to defaults made in  https://github.com/spacetelescope/jwst/pull/6041) and set the limits in the caller? CC: @hbushouse

mcara:
I will open another PR to overhaul `add_image` and `drizzle_arrays` to compute bbox at the top level and pass it to lower level routines and do this in a consistent way.

hbushouse:
Has this been superseded by #7774 ? If so, can this PR just be closed?

nden:
Yes, this is superseded by #7774

PR #7764, opened 2023-07-26
use scipy directly instead of astropy modeling for residual fringe fit

braingram:
While looking at the regression tests. I noticed that `test_residual_fringe_cal` is one of the slower tests.

Running just that test took ~2566 seconds:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/817

Profiling the test locally revealed that the majority of the time is spent in 2 functions:
- [new_fit_1d_fringes_bayes_evidence](https://github.com/spacetelescope/jwst/blob/7556988f110205042334d00c1c9d351c4bfb7773/jwst/residual_fringe/utils.py#L572)
- [fit_1d_background_complex](https://github.com/spacetelescope/jwst/blob/7556988f110205042334d00c1c9d351c4bfb7773/jwst/residual_fringe/utils.py#L391C5-L391C30)
Screen Shot 2023-07-26 at 8 46 27 AM

This PR aims to improve the second function `fit_1d_background_complex` by:
- replacing the [ChiSqOutlierRejectionFitter](https://github.com/spacetelescope/jwst/blob/7556988f110205042334d00c1c9d351c4bfb7773/jwst/residual_fringe/fitter.py#L6) class (defined in fitter.py shown in the above plot) with an equivalent function
- replacing the use of splines from astropy modeling with direct use of scipy.interpolate.BSpline

As mentioned in the [astropy Spline1D docs](https://docs.astropy.org/en/stable/api/astropy.modeling.spline.Spline1D.html#astropy.modeling.spline.Spline1D) much of the functionality of `Spline1D` is handled by `scipy.interpolate.BSpline`. The spline fitting here boiled down to:
- a call to `LSQUnivariateSpline` (which astropy wraps with `SplineExactKnotsFitter`)
- repeated fits with outlier rejection using the `ChiSqOutlierRejectionFitter`

The repeated fits are particularly costly with `Spline1D` due in large part to it's calls to `inspect` which are avoided by using `BSpline` directly. Here's a zoomed in portion of the plot showing what occurs during `Spline1D.__init__`:
Screen Shot 2023-07-26 at 9 04 40 AM

Running `test_residual_fringe_cal` with the changes in this PR took 1562 seconds (~60% the previous time):
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/818

[...]

PR #7772, opened 2023-07-26
Run pixel_replace before finalizing cal output in calwebb_spec2

tapastro:
Resolves [JP-3320](https://jira.stsci.edu/browse/JP-3320)

This PR addresses a bug reported by users setting pixel_replace to run and save results - this caused the default file_cal.fits output to go missing, as a result of incorrect ordering inside calwebb_spec2. This fixes the issue by running pixel_replace before setting metadata and suffix name for the datamodel that the pipeline returns.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 1.11.4 (unreleased)
 ===================
 
+calwebb_spec2
+-------------
+
+- Run ``pixel_replace`` before setting metadata and suffix of datamodel
+  that is returned by the pipeline to ensure a file is created with the
+  expected ``_cal`` suffix. [#7772]
+
 datamodels
 ----------
 
--- a/jwst/pipeline/calwebb_spec2.py
+++ b/jwst/pipeline/calwebb_spec2.py
@@ -271,13 +271,13 @@ class Spec2Pipeline(Pipeline):
         else:
             calibrated = self._process_common(calibrated)
 
+        # Replace bad pixels before rectification
+        calibrated = self.pixel_replace(calibrated)
+
         # Record ASN pool and table names in output
         calibrated.meta.asn.pool_name = pool_name
         calibrated.meta.asn.table_name = op.basename(asn_file)
         calibrated.meta.filename = self.make_output_path(suffix=suffix)
-
-        # Replace bad pixels before rectification
-        calibrated = self.pixel_replace(calibrated)
 
         # Produce a resampled product, either via resample_spec for
         # "regular" spectra or cube_build for IFU data. No resampled
--- a/jwst/pipeline/calwebb_spec2.py
+++ b/jwst/pipeline/calwebb_spec2.py
@@ -271,9 +271,6 @@ class Spec2Pipeline(Pipeline):
         else:
             calibrated = self._process_common(calibrated)
 
-        # Replace bad pixels before rectification
-        calibrated = self.pixel_replace(calibrated)
-
         # Record ASN pool and table names in output
         calibrated.meta.asn.pool_name = pool_name
         calibrated.meta.asn.table_name = op.basename(asn_file)
@@ -480,6 +477,7 @@ class Spec2Pipeline(Pipeline):
         calibrated = self.barshadow(calibrated)
         calibrated = self.wfss_contam(calibrated)
         calibrated = self.photom(calibrated)
+        calibrated = self.pixel_replace(calibrated)
 
         return calibrated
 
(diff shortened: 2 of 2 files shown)


tapastro:
Don't know of any changes this would cause in regtests, but here's a run: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/

braingram:
FYI devdeps will almost certainly fail due to a recent numpy 2.0 change breaking asdf numpy scalar encoding. See: https://github.com/spacetelescope/jwst/pull/7764#issuecomment-1652138463

hbushouse (commented):
Works for me. I'm wondering if it would be worth considering moving it into the various _process_xxxx functions, in order to avoid this kind of thing.

tapastro:
It might be - when I first submitted the pixel_replace code, I had hopes of it being used on more than just LRS data, which would require adding it to a few of the process_... flows, though I wasn't sure which. Should I add it to all of them minus the SOSS method, or just the common method?

hbushouse:
> It might be - when I first submitted the pixel_replace code, I had hopes of it being used on more than just LRS data, which would require adding it to a few of the process_... flows, though I wasn't sure which. Should I add it to all of them or just the common method?

I think I would go ahead and add it to all of them, unless you think there's no way the algorithm would ever work right for a given mode (e.g. wfss, soss). Speaking of SOSS, does the new, fancy ATOCA extraction already handle things like this and hence the replacement step would not be needed or would be redundant?

hbushouse:
Oh, and you should go ahead and file a JP ticket to track this, because if we're possibly going to ask DMS for a patch release because of this, it'll need traceability and visibility.

hbushouse:
Also, is there a reason this wasn't noticed in our regtests? Do we not have a test that runs this step, or does it naturally create a cal file when running the whole pipeline from the command line but not when running each step from python, or what? I know we have calwebb_spec2 tests of MIRI LRS exposures, which is the only mode for which this is turned on right now, so it should've run, unless the regtests don't use the param ref file that turns the step on.

tapastro:
It wasn't noticed because it requires the user to to set `pixel_replace.save_results=True`, which doesn't happen anywhere in testing (or OPS). To be honest, I still don't fully understand why the cal isn't created in that case (or at least, why it IS created when pixel_replace is run but save_results is not True), but this fix generates expected results under all permutations.

hbushouse:
Hmmm, well I think I now understand why it didn't show up in regtests, which is because it appears the step is never being run. The default in the step code is `skip=True`, but I *thought* that the MIRI LRS folks had delivered param ref files to turn it on. However, there aren't any "pars-pixelreplace" param files in CRDS yet, and the existing "pars-spec2pipeline" ref file that gets selected for MIR_LRS-FIXEDSLIT mode is still "https://jwst-crds.stsci.edu/browse/jwst_miri_pars-spec2pipeline_0003.asdf", which makes no mention of the pixel_replace step at all. Hence when you just run the calwebb_spec2 pipeline in a normal fashion, the pixel_replace step still gets skipped, and hence the cal file gets created normally.

Has the LRS team intentionally not turned this on yet, because they want more testing, or are the updated param ref files only in CRDS-TEST or what?

tapastro:
They appear to only be in TEST: https://jwst-crds-test.stsci.edu/browse/jwst_miri_pars-spec2pipeline_0006.asdf and https://jwst-crds-test.stsci.edu/browse/jwst_miri_pars-spec2pipeline_0007.asdf both include it.

hbushouse:
OK, when running the full calwebb_spec2 with all default options and pointing towards CRDS-TEST, such that it makes use of jwst_miri_pars-spec2pipeline_0006.asdf, it finally does turn on the pixel_replace step and the cal file is created normally at the very end of the process and has the results of pixel_replace applied to it (including having S_PXREPL='COMPLETE'). So perhaps this isn't an issue after all for use in Operations and hence an urgent patch is NOT required for B9.3. As long as we have the fix available in the development/master version on github, off-line users can pull that version for personal use.

hbushouse:
I did a full regtest run using CRDS-TEST (https://plwishmaster.stsci.edu:8081/job/RT/job/JWST/2496) which shows lots of differences in results for tests involving the new NIRCam specwcs ref files (wfss and tsgrism tests) and differences in LRS processing thanks to the pixel_replace step being applied. But the results show no actual errors and prove that the cal file still does get created properly for the LRS mode when pixel_replace is applied, so things look good for running B9.3 in Ops, even without this PR.

hbushouse (approved):
Looks good.

Issue #7762, opened 2023-07-25
Resample for SCI image ignores WCS's bounding box


stscijgbot-jp:
_Issue [JP-3317](https://jira.stsci.edu/browse/JP-3317) was created on JIRA by [Mihai Cara](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mcara):_

Resample step has two branches: one for resampling of the science image (SCI extension in the FITS file) and one for resampling error and variance arrays (ERR, VARIANCE_* extensions). The logic in the branch for processing the science image ignores the bounding box from the input images' WCS while the logic for processing ERR, VAR* arrays does take it into account.

Both science and error images should be processed (resampled) using the same limits for input image pixels.

stscijgbot-jp:
_Comment by [Nadia Dencheva](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dencheva) on [JIRA](https://jira.stsci.edu/browse/JP-3317?focusedCommentId=654509#comment-654509):_

#7774

PR #7774, opened 2023-07-27
JP-3317: Use same logic for resample limits for SCI, ERR, and VAR* images

mcara:
Resolves [JP-3317](https://jira.stsci.edu/browse/JP-3317)
Closes #7761

Resample step has two branches: one for resampling of the science image (SCI extension in the FITS file) and one for resampling error and variance arrays (ERR, VARIANCE_* extensions). The logic in the branch for processing the science image ignores the bounding box from the input images' WCS while the logic for processing ERR, VAR* arrays does take it into account.

Both science and error images should be processed (resampled) using the same limits for input image pixels.

This PR solves this issue by using the same algorithm for computing resample limits (in input image coordinates) to science, error, and variance arrays from input image shape and input WCS' bounding box.

**Checklist for maintainers**
- [x] added relevant label(s)
      [Regression test 824](https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/824)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,13 @@ master_background
 - Allow the user to write the 2D expanded version of the user-provided 1D background for each
   file in the assocation. [#7714]
 
+resample
+--------
+
+- Use the same logic for computing input range for resample step from input
+  image shape and the bounding box both for ``SCI`` image as well as for the
+  ``ERR`` and ``VARIANCE_*`` images. [#7774]
+
 
 1.11.3 (2023-07-17)
 ===================
--- a/jwst/resample/gwcs_drizzle.py
+++ b/jwst/resample/gwcs_drizzle.py
@@ -144,29 +144,29 @@ class GWCSDrizzle:
             Must have the same dimensions as insci. If none is supplied,
             the weighting is set to one.
 
-        xmin : float, optional
+        xmin : int, optional
             This and the following three parameters set a bounding rectangle
-            on the output image. Only pixels on the output image inside this
-            rectangle will have their flux updated. Xmin sets the minimum value
-            of the x dimension. The x dimension is the dimension that varies
-            quickest on the image. If the value is zero or less, no minimum will
-            be set in the x dimension. All four parameters are zero based,
-            counting starts at zero.
+            on the input image. Only pixels on the input image inside this
+            rectangle will have their flux added to the output image. Xmin
+            sets the minimum value of the x dimension. The x dimension is the
+            dimension that varies quickest on the image. All four parameters
+            are zero based, counting starts at zero.
 
-        xmax : float, optional
+        xmax : int, optional
             Sets the maximum value of the x dimension on the bounding box
-            of the output image. If the value is zero or less, no maximum will
-            be set in the x dimension.
+            of the input image. If ``xmax = 0``, no maximum will
+            be set in the x dimension (all pixels in a row of the input image
+            will be resampled).
 
-        ymin : float, optional
+        ymin : int, optional
             Sets the minimum value in the y dimension on the bounding box. The
             y dimension varies less rapidly than the x and represents the line
-            index on the output image. If the value is zero or less, no minimum
-            will be set in the y dimension.
+            index on the input image.
 
-        ymax : float, optional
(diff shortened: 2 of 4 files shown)


mcara:
Regression test: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/823

nden:
LGTM but needs a regression test run.

mcara:
Regression tests are running here: [Regression test 824](https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/824)

I expect differences, especially in error/variance arrays, and especially in the top row or right column. The results will need to be OKified if they look mostly OK.

nden (approved):
LGTM

hbushouse (approved):
Looks good to me. 

Just one question for clarification: the description makes it sound like the way the SCI array is resampled is changing to match the ERR and VAR arrays, yet many of the regtest results show small changes in the ERR and VAR arrays, but no change in SCI. Meanwhile some results show small changes in all of them. Can this be explained?

mcara:
> Looks good to me.
> 
> Just one question for clarification: the description makes it sound like the way the SCI array is resampled is changing to match the ERR and VAR arrays, yet many of the regtest results show small changes in the ERR and VAR arrays, but no change in SCI. Meanwhile some results show small changes in all of them. Can this be explained?

Yes, most can be explained. SCI extension was resampled without regard to the bounding box using all pixels (of course, where WCS was out of box and returning `nan` - those pixels were ignored. Now it will use the bounding box and it is possible that in most cases you will see no effect since it is effectively the same region of valid pixels (in the sense that their coordinates are valid; they are within the bounding box).

You could see a difference in the SCI extension *IF* the WCS would return valid coordinates outside of the bounding box and now we are imposing cutoff at the bounding box. Otherwise, you will not see anything.

There should be differences at the top row and rightmost column (large X coord) since the resample limits for VAR/ERR was computed as `int(bounding_box[0][1])` (e.g., for the X coordinate) and now it is computed as `int(bounding_box[0][1]) + 0.5` which is the correct computation. So, you could see some extra pixels to the right of the rightmost column or you could see a change of values in the last column (because more input pixels can contribute to the flux computations for those output pixels).

> Meanwhile some results show small changes in all of them

Small changes in **arbitrary** places - no, that should not happen. But you could see some changes in the resampled images for stage 3 if there are multiple input images because one extra column from those input images will contribute to output pixels. So, small changes where "stitching" between input images happens in the output frame are OK.

mcara:
In the discussion about `int(bounding_box[0][1] + 0.5)` I wanted to say/add that the old formula was resulting in lower `xmax` (or `ymax`) comutation than it should. Now we are expanding input range by 1 column/row (this assumes that the bounding box for a `100x100` image is `((-0.5, 99.5), (-0.5, 99.5))`. Old computation would result in `xmax=99` and now it is `xmax=100`. However, I think is this not ideal because the coordinate of the last raw is 99 and not 100. That is why in https://github.com/spacetelescope/jwst/issues/7762 I wanted to introduce `eps` - a small value by which to reduce `+0.5` so that we do not encounter such situations (but this would result in situations when a column of input image is ignored if bounding box says that only, let's say 0.000001 of that column's width is inside the bounding box).

I can make a quick PR to restore the use of `eps` if that is desired. From the `drizzle` point of view this - the use of an extra column) should work (not result in a crash) since it is checking for image boundaries and valid pixmap (coordinate transformations).

CC: @nden

hbushouse:
So it won't create an extra column or row in the output image that ends up with no data in it?

mcara:
Basically, the any location with `x` coordinate between `98.500000...1` and `99.5` should belong to 100th column with x coordinate 99.

mcara:
It should not.

hbushouse:
In that case I don't see an obvious need to implement the use of `eps`.

mcara:
But I would prefer not to have the coordinate 100 (hence the use of `eps`). However, the old formula was incorrect because a location of `98.99` would result in column coordinate 98 which is not the right coordinate (pixels are centered on the integer coordinates - not in the corners).

PR #7793, opened 2023-08-04
JP-3237 Compute correct exposure times for combined images

### LINKED ISSUE #7737: Compute correct exposure times for combined images
_Issue [JP-3237](https://jira.stsci.edu/browse/JP-3237) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

When the resample step creates combined mosaics of images it computes the correct total exposure time for the mosaic, leaving out sums over multiple detectors, but then only stores that correct total time in the EFFEXPTM keyword in the output image. There are several other exposure time related keywords, like XPOSURE (the recommended FITS time standard keyword) and DURATION that end up with incorrect total times when the mosaic involves multiple detectors.

The values for those keywords are based on the "fits_blender" rules defined for each keyword in the core schema. Those keywords currently use the simple "sum" rule, which means the time from all images that went into the combined mosaic will be summed, which results in a value that is too high by a factor equal to the number of detectors (e.g. 4 or 8 for NIRCam SW and 2 for NIRCam LW).

The resample step function "update_exposure_times" should be updated to store the correct total exposure time in more of those keywords, instead of just EFFEXPTM.

Note: This issue has been raised in both Help Desk tickets and Slack inquiries.

penaguerrero:
Resolves [JP-3237](https://jira.stsci.edu/browse/JP-3237)

Closes #7737 

This PR sets the exposure time related keywords EFFEXPTM, XPOSURE, DURATION, and TELAPSE to their appropriate combined times in the resample/resample_spec output products, reflecting the proper sum of times across all input exposures (but not across multiple detectors).

**Checklist for maintainers**
- [x] added relevant label(s)


jwst/resample/resample.py
```diff
@@ -374,6 +374,29 @@ def update_exposure_times(self, output_model):
         output_model.meta.exposure.end_time = max(exposure_times['end'])
         output_model.meta.resample.product_exposure_time = total_exposure_time
 
+        # Update other exposure time keywords:
+        # XPOSURE (identical to EFFEXPTM)
+        xposure = output_model.meta.exposure.frame_time * (output_model.meta.exposure.ngroups *\
```
hbushouse, line 379: Given that the `XPOSURE` keyword contains identical values as `EFFEXPTM` and the combined value for the latter is already computed as `total_exposure_time`, then why not just set `xposure = total_exposure_time`, instead of trying to recompute it all from scratch here. Also, there's apparently something in this recomputation that doesn't work right, because it's killing all the CI tests.

jwst/resample/resample.py
```diff
@@ -374,6 +374,29 @@ def update_exposure_times(self, output_model):
...
+            output_model.meta.exposure.nframes + (output_model.meta.exposure.ngroups - 1) *\
+            output_model.meta.exposure.groupgap + output_model.meta.exposure.drop_frames1) *\
+            output_model.meta.exposure.nints
+        output_model.meta.exposure.effective_exposure_time = xposure
+        # EFFINTTM (effective integration time)
+        effinttm = (output_model.meta.exposure.ngroups - 1) * output_model.meta.exposure.group_time \
```
hbushouse, line 385: It looks like this is just computing the value for one model, i.e. one exposure's values of ngroups, group_time, etc. We want the total over all input exposures, so it could be simply computed as the sum of all the `exposure[n].meta.exposure.integration_time` values, as is done in the loop starting at line 366 for the EFFEXPTM values. All of these could/should be computed by just doing that kind of sum over the values from all the input exposures, rather than trying to recompute from scratch.

hbushouse (changes requested):

jwst/resample/resample.py
penaguerrero: I addressed these changes

CHANGES.rst
```diff
@@ -53,6 +53,9 @@ resample
   image shape and the bounding box both for ``SCI`` image as well as for the
   ``ERR`` and ``VARIANCE_*`` images. [#7774]
 
+- Update the following exposure time keywords: XPOSURE (EFFEXPTM), EFFINTTM,
```
hbushouse, line 56: EFFINTTM should not be included in the updates. It's integration time within a single exposure and doesn't change for combined products.

jwst/resample/resample.py
```diff
@@ -363,17 +363,30 @@ def update_exposure_times(self, output_model):
         """Modify exposure time metadata in-place"""
         total_exposure_time = 0.
         exposure_times = {'start': [], 'end': []}
+        effinttm, duration = 0.0, 0.0
         for exposure in self.input_models.models_grouped:
-            total_exposure_time += exposure[0].meta.exposure.exposure_time
-            exposure_times['start'].append(exposure[0].meta.exposure.start_time)
-            exposure_times['end'].append(exposure[0].meta.exposure.end_time)
+            for n, _ in enumerate(exposure):
```
hbushouse, line 368: You need to keep the original design in the loop here, which only adds up values from `exposure[0]` in each of the grouped models. Each group contains models from multiple detectors in instruments like NIRCam and NIRSpec. But we don't want the times from multiple detectors included in the sums. So just use the first entry in each group.

jwst/resample/resample.py
```diff
@@ -363,17 +363,30 @@ def update_exposure_times(self, output_model):
...
+                total_exposure_time += exposure[n].meta.exposure.exposure_time
+                exposure_times['start'].append(exposure[n].meta.exposure.start_time)
+                exposure_times['end'].append(exposure[n].meta.exposure.end_time)
+                effinttm += exposure[n].meta.exposure.integration_time
```
hbushouse, line 372: don't include `integration_time`. That's an individual exposure quantity. Only include EFFEXPTM=XPOSURE, and DURATION=TELAPSE.

EFFEXPTM=meta.exposure.exposure_time
XPOSURE=meta.exposure.effective_exposure_time
DURATION=meta.exposure.duration
TELAPSE=meta.exposure.elapsed_exposure_time

So just create sums for `effexptm` and `duration`. Then below ...

jwst/resample/resample.py
```diff
@@ -363,17 +363,30 @@ def update_exposure_times(self, output_model):
...
         output_model.meta.resample.product_exposure_time = total_exposure_time
 
+        # Update other exposure time keywords:
+        # XPOSURE (identical to the total effective exposure time, EFFEXPTM)
+        xposure = total_exposure_time
+        output_model.meta.exposure.effective_exposure_time = xposure
+        # EFFINTTM (effective integration time) and DURATION (identical to TELAPSE, elapsed time)
+        output_model.meta.exposure.integration_time = effinttm
+        output_model.meta.exposure.duration = duration
+        output_model.meta.exposure.elapsed_exposure_time = duration
```
hbushouse, line 388:
```suggestion
        # XPOSURE (identical to the total effective exposure time, EFFEXPTM)
        output_model.meta.exposure.effective_exposure_time = total_exposure_time
        # DURATION (identical to TELAPSE, elapsed time)
        output_model.meta.exposure.duration = duration
        output_model.meta.exposure.elapsed_exposure_time = duration
```

hbushouse (changes requested):

jwst/resample/resample.py
penaguerrero: done

jwst/resample/resample.py
penaguerrero: done

hbushouse (approved):
Looks good.

hbushouse:
Regression test started at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/859

PR #7795, opened 2023-08-07
JP-3233: Changing Associations for FXD_SLIT

### LINKED ISSUE #7701: Combine S200A1 and S200A2 spectra in level 3 products
_Issue [JP-3233](https://jira.stsci.edu/browse/JP-3233) was created on JIRA by [Jeff Valenti](https://jira.stsci.edu/secure/ViewProfile.jspa?name=valenti):_

{*}Request{*}: For fixed slit spectroscopy with NIRSpec H gratings, enhance the association generator so that spec3 will combine exposures with the target in the S200A1 and S200A2 slits. From a data quality perspective, this will double the number of exposures being combined, substantially improving outlier rejection. From a usability perspective, the user will get one product with complete wavelength coverage, rather than two products each with a detector gap in a different wavelength range. Depending on the implementation, spec2 may be changed to encode in SOURCEID the slit containing the target and the slit being extracted. This would help users identify which level 3 products contain the target and may simplify association logic at level 3.

{*}Context{*}: For fixed slit spectroscopy with NIRSpec H gratings, [JDox](https://jwst-docs.stsci.edu/jwst-near-infrared-spectrograph/nirspec-observing-strategies/nirspec-dithering-recommended-strategies#NIRSpecDitheringRecommendedStrategies-NIRSpecdecisiontreeinfixedslitsmode) recommends taking exposures with the target in S200A1 and S200A2, if the observer wants complete wavelength coverage. [APT](https://jwst-docs.stsci.edu/jppom/nirspec-templates/nirspec-fixed-slit-spectroscopy-template-parameters#NIRSpecFixedSlitSpectroscopyTemplateParameters-Slit) provides an "S200A1 and S200A2" option for this purpose.

kmacdonald-stsci:
Resolves [JP-3233](https://jira.stsci.edu/browse/JP-3233)

Closes #7701 

This PR addresses how associations are created for `FXD_SLIT`.  For candidates that have values for this field, the association will include the values `S200A1` and `S200A2`, instead of separating those two values into separate associations.

**Checklist for maintainers**
- [x] added relevant label(s)


jwst/associations/lib/rules_level3_base.py
```diff
@@ -756,11 +756,25 @@ def __init__(self):
...
+                        sources=['fxd_slit'],
+                        value=['s200a1|s200a2'],
+                        force_unique=False,
+                        required=False,
+                    ),
+                    DMSAttrConstraint(
+                        name='opt_elem4',
+                        sources=['fxd_slit'],
+                        required=False,
+                    ),
```
hbushouse, line 773: So let me make sure I understand the result of the new logic. The 2nd new `DMSAttrConstraint` will continue to create a level-3 asn for a set of exposures that contain any value of `FXD_SLIT`, while the 1st new constraint will make sure that sets of exposures that have `FXD_SLIT` values of both S200A1 and S200A2 end up in the same level-3 ASN. Right?

CHANGES.rst
```diff
@@ -6,6 +6,13 @@ assign_wcs
 
 - Use isinstance instead of comparison with a type for lamp_mode inspection [#7801]
 
+associations
+------------
+
+- Changes the way associations are create for field  ``FXD_SLIT``.  Assocations
+  for this field will now include values for both ``S200A1`` and ``S200A2``,
+  instead of separating them. [#7795]
```
hbushouse, line 14:
```suggestion
- Update the Level 3 rules for NIRSpec fixed-slit observations so that sets of
  exposures using ``FXD_SLIT=S200A1`` and ``S200A2`` will be placed into
  the same ASN.  [#7795]
```

jwst/associations/lib/rules_level3_base.py
kmacdonald-stsci: Yes.

hbushouse:
The discussion in [JP-3233](https://jira.stsci.edu/browse/JP-3233) indicates that the exposures using both S200A1 and S200A2 should only be combined into a single Level 3 ASN *if* one of the high-dispersion gratings is in use, i.e. grating = g140h, g235h, or g395h. So this may need to be added as an additional constraint.

kmacdonald-stsci:
> The discussion in [JP-3233](https://jira.stsci.edu/browse/JP-3233) indicates that the exposures using both S200A1 and S200A2 should only be combined into a single Level 3 ASN _if_ one of the high-dispersion gratings is in use, i.e. grating = g140h, g235h, or g395h. So this may need to be added as an additional constraint.

The grating constraint was already present in that constraint class.

stscieisenhamer (approved):
Besides adding the new pool to artifactory so that `test_associations_sdp_pools` will pick it up, LGTM

hbushouse:
@kmacdonald-stsci  Has a regtest run been done for this? I don't see any indication of such. Also, @stscieisenhamer suggested adding the test pool for the program that contains some example observations to the existing artifactory suite of pools, so that it'll become part of the regtests. Has that been done?

hbushouse:
Regtests started at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/911

hbushouse:
Regtest run showed only unrelated differences (due to recent ref file updates).

hbushouse:
Closed in favor of #7879

PR #7796, opened 2023-08-07
JP-3244: Use source_{x,y}pos for NRS_FIXEDSLIT and NRS_MSASPEC

### LINKED ISSUE #7677: Extraction aperture centering for NIRSpec FS data needs a fix
_Issue [JP-3244](https://jira.stsci.edu/browse/JP-3244) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

The extract_1d step currently uses an automated centering routine to place the extraction aperture, based on the source coordinates. For reasons that aren't completely clear but likely related to the accuracy of the pointing and/or target coordinate keywords, the centering rarely if ever works correctly for FS data. If these keywords can't be fixed, then a new algorithm is needed.

Because FS observations are always done using target acquisition, the position of the source in the slit should always be known to very high accuracy (~5-10 mas given the typical TA uncertainties). So, it would make sense for the extraction aperture to be centered at the known dither position, independent of the actual source sky coordinates.

There are several potential options for implementing this in stage 2:
 # The pipeline could base the trace location on the dither keywords XOFFSET & YOFFSET, by shifting from the center of the s2d spectrum by that amount. This assumes that the center can be determined consistently.
 # Use dither-specific reference files to specify the extraction aperture locations. This would require separate files for each dither point in each dither pattern, along with new CRDS selection criteria, which would be cumbersome.

For stage 3, if the current algorithm can't be made to work, the only other option would be to use reference files. However, this is complicated by the fact that the spatial extent of the combined 2D spectrum depends on the number of dithers, so as with stage 2, multiple reference files would be required.

### LINKED ISSUE #7681: Extraction aperture centering for MOS needs a fix
_Issue [JP-3245](https://jira.stsci.edu/browse/JP-3245) was created on JIRA by [James Muzerolle](https://jira.stsci.edu/secure/ViewProfile.jspa?name=muzerol):_

The extract_1d step currently uses an automated centering routine to place the extraction aperture, based on the source coordinates. For reasons that aren't completely clear but likely related to the accuracy of the pointing keywords, the centering rarely if ever works correctly for MOS data.

The main difficulty for MOS is that sources can appear in any part of the shutter, and so unlike for FS and IFU, the positions are not the same in all cases for a given dither. The step reference file cannot be used for all sources because it only specifies a single aperture location.

The planned shutter positions (in relative shutter coordinates) could be used instead; given the excellent performance of MSATA, the source should be very close to that location in the vast majority of cases. That info is available from the MSA metafile (ESTIMATED_SOURCE_IN_SHUTTER_X/Y); however, these can't currently be used to determine the location of the extraction aperture because the resampled spectrum no longer contains any information on that reference frame.

One solution would be to convert these relative positions to a spatial offset from the center of the resampled 2D spectrum in the spatial direction; this would require knowing the location of the center on a consistent basis. This may be more complicated at stage 3, particularly if the exposure ensemble includes cases where the source was at very different shutter locations (e.g., not the typical slitlet nodding case).

Another option would be to change the resample_spec algorithm to do the spatial resampling in relative slit coordinates instead of sky coordinates. In that case, it should be more straightforward to then find the planned slit position in the s2d data. However, it's not clear how this can be handled at stage 3 processing, as dithered exposures will have different source positions within the slitlet.

Some discussion between the instrument team and the developers will be needed before we can decide on a path forward.

tapastro:
Resolves [JP-3244](https://jira.stsci.edu/browse/JP-3244)
Resolves [JP-3245](https://jira.stsci.edu/browse/JP-3245)
Closes #7677
Closes #7681 

This PR addresses extraction region issues for NIRSpec modes by utilizing source_xpos/ypos metadata to center the extraction region.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -90,6 +90,12 @@ engdb_tools
 - Check alternative host is alive before attempting to run test for
   access to avoid waiting the full timeout during test runs [#7780]
 
+extract_1d
+----------
+
+- Use ``source_{x/y}pos`` metadata to center extraction region for NIRSpec
+  (non-IFU) data. [#7796]
+
 extract_2d
 ----------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,9 @@ assign_wcs
   slits use a two-digit source_id value that reflects both the primary target in
   use and the secondary slit from which the data are extracted. [#7879]
 
+- Compute sky position of dithered slit center for MIRI LRS fixed slit data, and
+  store in dither metadata under ``dithered_ra`` and ``dithered_dec``. [#7796]
+
 associations
 ------------
 
@@ -94,7 +97,7 @@ extract_1d
 ----------
 
 - Use ``source_{x/y}pos`` metadata to center extraction region for NIRSpec
-  (non-IFU) data. [#7796]
+  (non-IFU) data; use dithered pointing info for MIRI LRS fixed slit data. [#7796]
 
 extract_2d
 ----------
--- a/docs/jwst/extract_1d/arguments.rst
+++ b/docs/jwst/extract_1d/arguments.rst
@@ -79,14 +79,13 @@ The ``extract_1d`` step has the following step-specific arguments.
   off-center due to things like nodding or dithering. If turned on, the position
   of the source is used in conjunction with the World Coordinate System (WCS) to
   compute the x/y source location. For NIRSpec non-IFU modes, the source position
-  is determined from the ``source_xpos/source_ypos`` parameters, while other modes
-  use ``targ_ra/targ_dec``. For long-slit type modes (e.g. MIRI LRS and NIRSpec
-  fixed-slit and MOS), only the position in the cross-dispersion direction is used
-  to potentially offset the extraction regions in that direction. If this parameter
-  is specified in the :ref:`EXTRACT1D <extract1d_reffile>` reference file, the
-  reference file value will override any automatic settings based on exposure and
-  source type. As always, a value given by the user as an argument to the step
-  overrides all settings in the reference file or within the step code.
+  is determined from the ``source_xpos/source_ypos`` parameters. For MIRI LRS fixed slit,
+  the dither offset is applied to the sky pointing location to determine source position.
+  All other modes use ``targ_ra/targ_dec``. If this parameter is specified in the
+  :ref:`EXTRACT1D <extract1d_reffile>` reference file, the reference file value will
+  override any automatic settings based on exposure and source type. As always, a value
+  given by the user as an argument to the step overrides all settings in the reference
+  file or within the step code.
 
 ``--center_xy``
   A list of two integer values giving the desired x/y location for the center
--- a/docs/jwst/extract_1d/arguments.rst
+++ b/docs/jwst/extract_1d/arguments.rst
@@ -76,16 +76,17 @@ The ``extract_1d`` step has the following step-specific arguments.
   the step will make the decision of whether to use the source position based
   on the observing mode and the source type. The source position will only be
   used for point sources and for modes where the source could be located
-  off-center due to things like nodding or dithering. If turned on, the sky
-  (RA/Dec) position of the source is used in conjunction with the World
-  Coordinate System (WCS) to compute the x/y source location. For long-slit
-  type modes (e.g. MIRI LRS and NIRSpec fixed-slit and MOS), only the position
-  in the cross-dispersion direction is used to potentially offset the
-  extraction regions in that direction. If this parameter is specified in the
-  :ref:`EXTRACT1D <extract1d_reffile>` reference file, the reference file value
-  will override any automatic settings based on exposure and source type.
-  As always, a value given by the user as an argument to the step overrides all
-  settings in the reference file or within the step code.
+  off-center due to things like nodding or dithering. If turned on, the position
+  of the source is used in conjunction with the World Coordinate System (WCS) to
+  compute the x/y source location. For NIRSpec non-IFU modes, the source position
+  is determined from the ``source_xpos/source_ypos`` parameters, while other modes
+  use ``targ_ra/targ_dec``. For long-slit type modes (e.g. MIRI LRS and NIRSpec
+  fixed-slit and MOS), only the position in the cross-dispersion direction is used
+  to potentially offset the extraction regions in that direction. If this parameter
+  is specified in the :ref:`EXTRACT1D <extract1d_reffile>` reference file, the
+  reference file value will override any automatic settings based on exposure and
+  source type. As always, a value given by the user as an argument to the step
+  overrides all settings in the reference file or within the step code.
 
 ``--center_xy``
   A list of two integer values giving the desired x/y location for the center
--- a/jwst/assign_wcs/assign_wcs.py
+++ b/jwst/assign_wcs/assign_wcs.py
@@ -6,6 +6,7 @@ from .util import (update_s_region_spectral, update_s_region_imaging,
 from ..lib.exposure_types import IMAGING_TYPES, SPEC_TYPES, NRS_LAMP_MODE_SPEC_TYPES
 from ..lib.dispaxis import get_dispersion_direction
 from ..lib.wcs_utils import get_wavelengths
+from .miri import store_dithered_position
 
 log = logging.getLogger(__name__)
 log.setLevel(logging.DEBUG)
@@ -93,6 +94,10 @@ def load_wcs(input_model, reference_files={}, nrs_slit_y_range=None):
                 except Exception as exc:
                     log.info("Unable to update S_REGION for type {}: {}".format(
                         output_model.meta.exposure.type, exc))
+
+        # Store position of dithered pointing location in metadata for later spectral extraction
+        if output_model.meta.exposure.type.lower() == 'mir_lrs-fixedslit':
+            output_model = store_dithered_position(output_model)
 
     log.info("COMPLETED assign_wcs")
     return output_model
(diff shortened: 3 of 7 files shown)


tapastro:
Regtests running here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/839/

nden:
For `NRS_BRIGTOBJ` the attributes are set in [extract_2d.](https://github.com/spacetelescope/jwst/blob/master/jwst/extract_2d/nirspec.py#L158)

tapastro:
brightobj added, regtests #2 here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/842

jwst/extract_1d/extract.py
```diff
@@ -1439,11 +1439,24 @@ def locn_from_wcs(
...
+                                              'NRS_BRIGHTOBJ']:
+            if slit is None:
+                xpos = input_model.source_xpos
+                ypos = input_model.source_ypos
+            else:
+                xpos = slit.source_xpos
+                ypos = slit.source_ypos
+
+            slit2det = self.wcs.get_transform('slit_frame', 'detector')
+            x_y = slit2det(xpos, ypos, middle_wl)
```
hbushouse, line 1452: I wonder if we should have a log message in this new branch saying that the source x/y pos from the meta data is being used for centering, instead of the target RA/Dec. A user could run this with `use_source_posn` turned on, assuming it would compute the position from the RA/Dec and then not get that result.

hbushouse (commented):
We should have updates to the docs to describe this alternate approach for the relevant NIRSpec modes.

hbushouse:
Do we not have any equivalent source x/y pos info for the MIRI LRS mode, such that we could use this for that mode too?

hbushouse:
Wayne posted in JP-3245 a list of MOS exposures for which he thinks the basic pointing info should be good, which could make good additional tests for this PR. Have you tried any of those?

jwst/extract_1d/extract.py
```diff
@@ -1439,11 +1439,26 @@ def locn_from_wcs(
...
-            log.warning("Inverse wcs is not implemented, so can't use target coordinates to get location of spectrum.")
-            return
+        if input_model.meta.exposure.type in ['NRS_FIXEDSLIT', 'NRS_MSASPEC',
+                                              'NRS_BRIGHTOBJ']:
+            if slit is None:
+                xpos = input_model.source_xpos
+                ypos = input_model.source_ypos
+            else:
+                xpos = slit.source_xpos
+                ypos = slit.source_ypos
```
hbushouse, line 1449: Someone mentioned in the JP ticket discussing MOS mode that the "estimated_source_in_slit_x/y" meta values were no longer available in the resampled (s2d) images used as input to extract_1d. I assume these source_x/ypos values originate from estimated_source_in_slit_x/y and hence the meta data are actually still there, just under a different name?

nden:
Note that for NRS FIXEDSLIT the `source_x/ypos` are computed only for the primary slit. For the other slits they default to 0. This should be clarified in the documentation.

tapastro:
Note: This PR now requires merge of https://github.com/spacetelescope/stdatamodels/pull/203 and subsequent release of `stdatamodels` prior to merging, so setting to draft.

jwst/assign_wcs/miri.py
```diff
@@ -603,3 +603,32 @@ def get_wavelength_range(input_model, path=None):
...
     return dict([(ch + band, wr[ch + band]) for ch in channel])
+
+
+def store_dithered_position(input_model):
+    """Store the location of the dithered pointing
+    location in the wcsinfo
+
+    Parameters
+    ----------
+    input_model : jwst.datamodels.ImageModel
```
hbushouse, line 614: minor nit, but we usually enclose the datamodels name in back tics (e.g. see line 585 above)

jwst/assign_wcs/miri.py
```diff
@@ -603,3 +603,32 @@ def get_wavelength_range(input_model, path=None):
...
+
+    dithered_v2, dithered_v3 = idltov23(input_model.meta.dither.x_offset, input_model.meta.dither.y_offset)
+
+    v23toworld = input_model.meta.wcs.get_transform('v2v3', 'world')
+    dithered_ra, dithered_dec, _ = v23toworld(dithered_v2, dithered_v3, 0.0)  # needs a wavelength,
+                                                                              # but does not affect return
+
+    input_model.meta.dither.dithered_ra = dithered_ra
+    input_model.meta.dither.dithered_dec = dithered_dec
+    return input_model
```
nden, line 634: Is it necessary to return the model? It only adds attributes and makes no copies.

jwst/extract_1d/extract.py
nden: That is correct.

nden (approved):
I have only one minor comment about returning the model in the new function, which is not wrong but unnecessary.

jwst/extract_1d/extract.py
```diff
@@ -1439,11 +1439,32 @@ def locn_from_wcs(
...
+
+            slit2det = self.wcs.get_transform('slit_frame', 'detector')
+            x_y = slit2det(xpos, ypos, middle_wl)
+            log.info("Using source_xpos and source_ypos to center extraction.")
+
+        elif input_model.meta.exposure.type == 'MIR_LRS-FIXEDSLIT':
+            try:
+                if slit is None:
+                    dithra = input_model.meta.dither.dithered_ra
```
hbushouse, line 1458: So, is "dithra" a relative of "mothra"?   ;-)

hbushouse (commented):
Looks good. Just a few comments/questions.

jwst/assign_wcs/miri.py
```diff
@@ -603,3 +603,32 @@ def get_wavelength_range(input_model, path=None):
     band = input_model.meta.instrument.band
 
     return dict([(ch + band, wr[ch + band]) for ch in channel])
+
+
+def store_dithered_position(input_model):
+    """Store the location of the dithered pointing
+    location in the wcsinfo
```
nden, line 610: Ah, and it's stored in the `dither` section now, not `wcsinfo`.

jwst/assign_wcs/miri.py
tapastro: Good call, return statement has been removed.

CHANGES.rst
```diff
@@ -90,6 +90,12 @@ engdb_tools
 - Check alternative host is alive before attempting to run test for
   access to avoid waiting the full timeout during test runs [#7780]
 
+extract_1d
+----------
+
+- Use ``source_{x/y}pos`` metadata to center extraction region for NIRSpec
```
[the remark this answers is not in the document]
tapastro: ✅

docs/jwst/extract_1d/arguments.rst
```diff
@@ -76,16 +76,17 @@ The ``extract_1d`` step has the following step-specific arguments.
...
-  in the cross-dispersion direction is used to potentially offset the
-  extraction regions in that direction. If this parameter is specified in the
-  :ref:`EXTRACT1D <extract1d_reffile>` reference file, the reference file value
-  will override any automatic settings based on exposure and source type.
-  As always, a value given by the user as an argument to the step overrides all
-  settings in the reference file or within the step code.
+  off-center due to things like nodding or dithering. If turned on, the position
+  of the source is used in conjunction with the World Coordinate System (WCS) to
+  compute the x/y source location. For NIRSpec non-IFU modes, the source position
+  is determined from the ``source_xpos/source_ypos`` parameters, while other modes
```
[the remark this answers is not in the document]
tapastro: ✅

jwst/assign_wcs/miri.py
tapastro: ✅

jwst/assign_wcs/assign_wcs.py
```diff
@@ -94,5 +95,10 @@ def load_wcs(input_model, reference_files={}, nrs_slit_y_range=None):
                     log.info("Unable to update S_REGION for type {}: {}".format(
                         output_model.meta.exposure.type, exc))
 
+        # Store position of dithered pointing location in metadata for later spectral extraction
+        if output_model.meta.exposure.type.lower() == 'mir_lrs-fixedslit':
+            output_model = store_dithered_position(output_model)
```
nden, line 100: Since the function doesn't have a return value this should be changed too, otherwise `output_model` is  None.

tapastro: 🤦

hbushouse (approved):
LGTM

hbushouse:
Any luck getting a regtest to run?

tapastro:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/926/ started now.

nden:
The current error is due to a missing v3yangle n a test fixture [here](https://github.com/spacetelescope/jwst/blob/master/jwst/master_background/tests/test_master_background.py#L33).

tapastro:
Looks like I was able to get a run to stick by changing the conda python version in the JenkinsfileRT - link: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/935/

tapastro:
After a long wait, it looks like the regression tests all went as expected. The MIRI LRS differences appear to show that we don't test any LRS point-like extractions, unless by chance we exactly align with prior extraction windows - the diffs show the new metadata for dither sky position but no flux differences. A slew of flux changes for NRS FS/MOS, nearly all of which show an increase in reported flux. Also some diffs for NIRCam imaging, but this is due to photom ref file updates.

jwst/master_background/tests/test_master_background.py
```diff
@@ -43,6 +43,8 @@ def science_image():
     image.meta.subarray.ystart = 1
     image.meta.wcsinfo.v2_ref = 0
     image.meta.wcsinfo.v3_ref = 0
+    image.meta.wcsinfo.v3yangle = 0
+    image.meta.wcsinfo.vparity = 0
```
nden, line 47: I believe this should be 1 though it's probably not used in the test

nden (commented):
I thought I posted these but apparently not.
Also I'm wondering why there are no data differences in the extract1d products for miri?

PR #7801, opened 2023-08-08
CI fixes: Replace type comparison with isinstance and increase lower pin for asdf

braingram:
I noticed the CI fail on the main branch due to likely a ruff update (which caused a new E721 rule violation due to a comparison using a type instead of use of `isinstance`) and the oldest deps failing (due to the asdf dependency, jsonschema, updating).

This PR switches the type comparisons to calls to `isinstance` and increases the lower asdf version to 2.14.4.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,12 @@
 1.11.4 (unreleased)
 ===================
 
+assign_wcs
+----------
+
+- Use isinstance instead of comparison with a type for lamp_mode
+  inspection [#7801]
+
 calwebb_spec2
 -------------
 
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -19,6 +19,11 @@ flat_field
 - Modify the test_flatfield_step_interface unit test to prevent it from causing
   other tests to fail [#7752]
 
+general
+-------
+
+- Require minimum asdf version 2.14.4 [#7801]
```
hbushouse, line 25: Should also include an entry for `assign_wcs`, indicating the change that was made.

braingram: Thanks! I added an entry to that section.

braingram:
Regression run started:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/853/

PR #7808, opened 2023-08-11
Fix NIRCam imaging regtest

hbushouse:
This PR fixes a problem in the way the Detector1 and Image2 pipelines were being called back-to-back in the NIRCam imaging mode regtest. The Image2 test was doing a call to `get_data()` on the rate file created by the Detector1 test, so instead of using the output rate file created by Detector1, it was pulling an old version of the rate file from Artifactory. D'oh. The modification no longer call `get_data()` for that file, but just uses the file already sitting in the working directory.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/regtest/test_nircam_image.py
+++ b/jwst/regtest/test_nircam_image.py
@@ -50,7 +50,6 @@ def run_image3pipeline(run_image2pipeline, rtdata_module, jail):
     # produce fresh _cal files for the image3 pipeline.  We won't check these
     # or look at intermediate products, including the resampled i2d
     rate_files = [
-        "nircam/image/jw01538046001_03105_00001_nrcalong_rate.fits",
         "nircam/image/jw01538046001_03105_00001_nrcblong_rate.fits",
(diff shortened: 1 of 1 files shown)


hbushouse:
regtest running at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/854

tapastro:
Huh, and the other four rates called in image3 aren't causing any issues? I suppose if it passes, it's good enough for me!

tapastro (approved):

hbushouse:
> Huh, and the other four rates called in image3 aren't causing any issues? I suppose if it passes, it's good enough for me!

No. The real problem was that the 1 rate file used as input to Image2 was getting pulled into the working directory *after* Detector1 had created a new rate for that exposure, so the output of Detector1 was getting overwritten. Hence when you run "okify", it was using that (old) rate file to overwrite the truth file and so the truth file was never actually getting updated with the new result from Detector1.

PR #7809, opened 2023-08-11
JP-3343: Add bounding_box to imaging WCS objects without one

### LINKED ISSUE #7820: NIRSpec imaging products have no bounding box assigned to WCS
_Issue [JP-3343](https://jira.stsci.edu/browse/JP-3343) was created on JIRA by [Tyler Pauly](https://jira.stsci.edu/secure/ViewProfile.jspa?name=tpauly):_

From a Help Desk ticket (from [Charles Proffitt](https://jira.stsci.edu/secure/ViewProfile.jspa?name=proffitt)), it was noted that NIRSpec imaging products have no FITS SIP parameters assigned. This was tracked down to the WCS not having a bounding_box assigned. Simple fix is available.

tapastro:
Resolves [JP-3343](https://jira.stsci.edu/browse/JP-3343)

Closes #7820 

This PR addresses a help desk inquiry regarding missing FITS SIP approximations for a NIRSpec TAcq image. These values were not filled during assign_wcs because the bounding_box was never defined. This PR is an experiment to see if saving a bounding box to imaging WCS objects of size {data.shape}, when one does not already exist, will break things.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -958,6 +958,7 @@ def update_s_region_imaging(model):
 
     if bbox is None:
         bbox = wcs_bbox_from_shape(model.data.shape)
+        model.meta.wcs.bounding_box = bbox
 
     # footprint is an array of shape (2, 4) as we
     # are interested only in the footprint on the sky
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -956,10 +956,6 @@ def update_s_region_imaging(model):
 
     bbox = model.meta.wcs.bounding_box
 
-    if bbox is None:
-        bbox = wcs_bbox_from_shape(model.data.shape)
-        model.meta.wcs.bounding_box = bbox
-
     # footprint is an array of shape (2, 4) as we
     # are interested only in the footprint on the sky
     footprint = model.meta.wcs.footprint(bbox, center=True, axis_type="spatial").T
@@ -986,8 +982,6 @@ def compute_footprint_spectral(model):
     """
     swcs = model.meta.wcs
     bbox = swcs.bounding_box
-    if bbox is None:
-        bbox = wcs_bbox_from_shape(model.data.shape)
 
     x, y = grid_from_bounding_box(bbox)
     ra, dec, lam = swcs(x, y)
--- a/jwst/assign_wcs/util.py
+++ b/jwst/assign_wcs/util.py
@@ -956,6 +956,10 @@ def update_s_region_imaging(model):
 
     bbox = model.meta.wcs.bounding_box
 
+    if bbox is None:
+        bbox = wcs_bbox_from_shape(model.data.shape)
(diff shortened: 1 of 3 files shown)


tapastro:
Checking regtests here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/

hbushouse:
Since this is in response to a Help Desk query, it would be good to retroactively create a JP ticket and link it here. Also helps with the bean counting of how many tickets/bugs we fix!  ;-)

hbushouse (approved):
Seems simple and straightforward enough to me. But really need the blessing of the WCS czar (@nden ).

hbushouse:
Regtest results show mainly the differences due to a recent update to the refpix step for NRS IRS2 data. One test, for an NRS2 image, does show that the FITS SIP keywords are now included in the cal product header and fully populated. So that's a good sign.

nden:
I suggest we add the code in `assign_wcs_step`, just before the SIP computation. It doesn't feel right to mix it with the computation of `S_REGION`.

tapastro:
New regtest run here: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/867/

Strange failure has disappeared from regtests.

I moved the bbox generation to right before the S_REGION calculations, because most of them check for bbox presence and mock one if it didn't exist. This isn't quite where @nden wanted it - I can move it to `assign_wcs_step` if that's still preferred.

jwst/assign_wcs/util.py
```diff
@@ -985,8 +982,6 @@ def compute_footprint_spectral(model):
     """
     swcs = model.meta.wcs
     bbox = swcs.bounding_box
-    if bbox is None:
-        bbox = wcs_bbox_from_shape(model.data.shape)
```
nden, line 989: I think these lines need to stay. For spectral exposure which go through extract_2d, it doesn't make sense to have a bounding_box assigned but it is needed to compute the footprint. Here the shape of the array is used as a bounding box only to compute `s_region` and is not assigned to the WCS.

jwst/assign_wcs/util.py
```diff
@@ -956,9 +956,6 @@ def update_s_region_imaging(model):
 
     bbox = model.meta.wcs.bounding_box
 
-    if bbox is None:
-        bbox = wcs_bbox_from_shape(model.data.shape)
```
nden, line 960: This needs to be here as a guard for L3 products without a bounding box.

nden (commented):
Initially I was suggesting to assign a bounding_box to all imaging modes in assign_wcs_step. Looking at the current changes I realized this will not cover resampled products. Now I think your first solution is best. Sorry for the flip flop.

nden (approved):

hbushouse:
Is everyone agreed that this is all OK now?  @nden

PR #7811, opened 2023-08-11
JP-3336: Fix WCS bugs for internal_cal cubes

### LINKED ISSUE #7810: Internal cal cube has incorrect WCS
_Issue [JP-3336](https://jira.stsci.edu/browse/JP-3336) was created on JIRA by [Melanie Clarke](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mclarke):_

The output WCS for cubes made in the 'internal_cal' coordinates (slicer x/y, in meters) is incorrect.  The lower left corner has the correct coordinates (-0.011, -0.007), but all other pixels have negative values when they should range up to (0.011, 0.007), with 0.0 at the center.

It looks like the issue is that there are a couple places in the cube_build step where the output WCS is assumed to be celestial, when it is not, in this case.  The output units are incorrectly divided by 3600 to convert from arcsec to degrees, and the x axis is reversed for RA conventions.

In addition, the output scale for the cube is forced to be square (CDELT2 = CDELT1), when internal slicer coordinates are rectangular.  Leaving the pixel units as is makes a more natural product that is easier to analyze: point sources in the output cube appear circular instead of elliptical.

This fix is needed for NIRSpec calibration support, but the WCS corrections do also affect MIRI internal_cal cubes.

melanieclarke:
Resolves [JP-3336](https://jira.stsci.edu/browse/JP-3336)

Closes #7810 

This PR addresses an inconsistency in handling the output WCS for IFU cubes built in internal_cal coordinates.  

**Checklist for maintainers**
- [x] added relevant label(s)


jemorrison:
@melanieclarke I am curious about these changes. I had a few comments. These types of cube are really for NIRSpec calibrations. Is the NIRSpec team making these types of cubes now. Just curious.
Also you are going to need an update to the CHANGE LOG

melanieclarke:
@jemorrison - Yes, this came up because these cubes were needed for NIRSpec calibration, to confirm recent changes to the instrument model for the IFU SIAF position.  We'd like to get the fix in the pipeline because more change/analysis rounds might be necessary.

I'll update the change notes.  Thanks!

jemorrison:
What about trying to make the units of the output cube in arc seconds. Is that correct  since I thought the units for NIRSpec with internal_cal are meters ?
Though it has been a while since I thought about NIRSpec internal_cal type cubes so maybe I have that wrong

jemorrison:
Have you run the regression tests on this branch ?

melanieclarke:
@jemorrison - I answered in the JP ticket, but for completeness here too:

> The final units for the internal_cal cube should be meters. 
> 
> In ifu_cube.py, around line 2214, the output cdelt values are divided by 3600 to convert from an assumed arcsec value to degrees.  This correction is not appropriate to internal_cal cubes, because the units are meters, not arcsec.
> 
> In my fix, I left this line as is for most IFU cubes, but uncorrected it specifically for internal_cal cubes by multiplying 3600 back in, around line 2295.

I have not run regression tests myself.  I'm not a pipeline developer -- I don't think I have access.

jemorrison:
Let me see how to run the regression tests for you.

jemorrison (commented):
Looks good.  I am trying to run the regression tests for this PR.

jemorrison:
@melanieclarke  
Just to confirm you want cdelt2 defined this way: self.cdelt2 = self.cdelt1 / 2.0 (line 1391 in ifu_cube.py)

I was able to run the regression tests on your branch. Can you access this site:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/864/ and look at the regression tests for the
test_cube_build_nirspec_internal_cal. The differences in what the truth is in the regression database and the output from this test seem reasonable. I just wanted to confirm that with you.

jemorrison:
@melanieclarke we recently made some changes to cube_build you need to pull those changes into this branch. So you don't get the out-of-date warning.

melanieclarke:
@jemorrison 
Thanks for running the regression tests.  I can see them, and the changes do look right, except that PC1_1 was changed from a float to an integer.  I pushed a small change to just switch the sign on PC1_1, instead of setting directly to 1.

The self.cdelt2 = self.cdelt1 / 2.0 line is correct: the y-axis scale on the slicer is half the scale of the x-axis.  The NIRSpec IFU SIAF expert confirms that the WCS for cubes built this way look correct.

jemorrison (approved):
Looks good.

CHANGES.rst
```diff
@@ -33,6 +33,8 @@ cube_build
 
 - Replace scale1 and scale2 arguments with scalexy, add debug option debug_spaxel,
   and add more details to docs. [#7783]
+- Fix output WCS for IFU cubes built in internal_cal coordinates, for NIRSpec
+  calibration analysis. [#7811]
```
hbushouse, line 37:
```suggestion

- Fix output WCS for IFU cubes built in internal_cal coordinates, for NIRSpec
  calibration analysis. [#7811]
```

hbushouse: Can you elaborate a bit on exactly how the output WCS has been fixed? what was wrong with it before?

jwst/cube_build/ifu_cube.py
```diff
@@ -395,7 +395,7 @@ def set_geometryAB(self, corner_a, corner_b, lambda_min, lambda_max):
             across_cdelt = self.cdelt2
 
         if self.instrument == 'NIRSPEC':
-            self.cdelt2 = self.cdelt1  # make cubes square
+            # self.cdelt2 = self.cdelt1  # make cubes square
```
hbushouse, line 398: Probably should just remove this line, instead of leaving it comment out.

jwst/cube_build/ifu_cube.py
```diff
@@ -2291,6 +2291,13 @@ def setup_final_ifucube_model(self, model_ref):
             ifucube_model.meta.ifu.roi_wave = 0
             ifucube_model.meta.ifu.roi_spatial = 0
 
+            # uncorrect cdelt for degree conversion
+            ifucube_model.meta.wcsinfo.cdelt1 *= 3600.0
+            ifucube_model.meta.wcsinfo.cdelt2 *= 3600.0
+
+            # correct "RA" axis orientation
+            ifucube_model.meta.wcsinfo.pc1_1 *= -1.0
```
hbushouse, line 2299: Good catch. A lot of folks forget about the fact that RA runs right to left on the sky!

jwst/cube_build/ifu_cube.py
melanieclarke: Done.

CHANGES.rst
melanieclarke: I just pushed a clarification: the corrections were to the output scale and orientation for slicer coordinates.

CHANGES.rst
```diff
@@ -33,6 +33,8 @@ cube_build
 
 - Replace scale1 and scale2 arguments with scalexy, add debug option debug_spaxel,
   and add more details to docs. [#7783]
+- Correct slicer scale and orientation in output WCS for IFU cubes built in internal_cal
+  coordinates, for NIRSpec calibration analysis. [#7811]
```
hbushouse, line 37:
```suggestion

- Correct slicer scale and orientation in output WCS for IFU cubes built in internal_cal
  coordinates, for NIRSpec calibration analysis. [#7811]
```

hbushouse (approved):
Looks good now. I'm just making 1 minor update to the change log (to add a blank line between entries).

PR #7823, opened 2023-08-15
JP-3345: Implement mingrad method for MIRI MRS pixel replacement

### LINKED ISSUE #7822: MRS pixel replacement modifications
_Issue [JP-3345](https://jira.stsci.edu/browse/JP-3345) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

<https://jira.stsci.edu/browse/JP-3005> implemented pixel replacement for slit spectroscopy, along with a draft implementation that could be used for MRS as well.  While it performed fairly well, there were still clear failure cases in which the replaced values were significantly incorrect.  At the same time, pixel replacement is less critical for the MRS since the cube building algorithm natively does a degree of replacement using neighboring pixels as part of the voxel overlap computation, and the decision was made to keep this step off for MRS for the time being.

More recently, analysis of data from PID 1268 has shown that NaNs in the cal files are the source of small (generally negative) spikes in the extracted spectra.  The issue is far less pronounced than for the LRS.  The root cause seems to be undersampling; when NaNs occur in the middle of a bright point source trace cube build fills in the corresponding cube voxel using adjacent pixels that are significantly lower in flux.  This is mitigated somewhat by additional dither positions, but not fully.

As a result, I've looked back at the pixel replacement approach and am testing a revised implementation to see if it can do a better job.  This minimum gradient approach is super simple: For every NaN in the cal array, calculate the absolute gradient along the spatial and spectral axes (X and Y) using immediately adjacent pixels.  Pick whichever dimension has the minimum absolute gradient and replace the missing pixel with the average of the two adjacent pixels along that dimension.

This aims to make the process extremely local as global approaches will not generally be reliable; near point sources it will do the replacement along the spectral axis avoiding sampling issues, while near bright extended emission line the replacement will be along the spatial axis.  It may still be suboptimal near bright emission lines from unresolved point sources.  There is no attempt at any replacement if a NaN value is bordered by another NaN value along a given axis.

Test results to follow.

drlaw1558:
Resolves [JP-3345](https://jira.stsci.edu/browse/JP-3345) by implementing a local pixel replacement alternative for the MIRI MRS.

Closes #7822 

**Checklist for maintainers**
- [x] added relevant label(s)


tapastro (approved):
Looks good! Thanks for contributing an upgrade to the step 🙂  I will leave merging to the bosses.

jwst/pixel_replace/pixel_replace.py
```diff
@@ -326,6 +333,104 @@ def fit_profile(self, model):
...
+            case of DataModels containing only one 2D spectrum,
+            or a single 2D spectrum from the input DataModel
+            containting multiple spectra (i.e. MultiSlitModel).
+            Requires data and dq attributes.
+
+        Returns
+        -------
+        model_replaced : DataModel
+            DataModel with flagged bad pixels now flagged with
+            TO-BE-DETERMINED and holding a flux value estimated
```
hbushouse, line 364: Don't we now have a DQ name that's been appropriated for this? If so, the comment could be updated.

jwst/pixel_replace/pixel_replace.py
```diff
@@ -326,6 +333,104 @@ def fit_profile(self, model):
...
+                    newdq[yindx[ii], xindx[ii]] = self.FLUX_ESTIMATED
+                # If original pixel was in non-science region set DQ accordingly
+                else:
+                    newdq[yindx[ii], xindx[ii]] = self.FLUX_ESTIMATED + self.NON_SCIENCE + self.DO_NOT_USE
+
+                nreplaced += 1
+            except Exception:
+                pass
+
+        log.info(f"Number of pixels replaced: {nreplaced}")
```
hbushouse, line 426: I recall seeing similar log statements in the upper-level routine. Are we getting double reporting (or at least more than necessary)?

hbushouse:
There's one minor code style error that needs fixing:
jwst/pixel_replace/pixel_replace.py:392:49: E712 [*] Comparison to `False` should be `cond is False` or `if not cond:`

hbushouse:
Regression test run started at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/875

drlaw1558:
@hbushouse I wasn't sure entirely what to do about that style error on line 392.  When I tried replacing == with 'is' it busted the entire routine and it never flagged anything, returning a zero-length set from the np.where.  'is' doesn't seem to be an appropriate replacement for == in this case.  Is there some different syntax that should be used?

jwst/pixel_replace/pixel_replace.py
drlaw1558: Yes, FLUX_ESTIMATED.  I'll make this change here and in the original code for the default method.

hbushouse:
> @hbushouse I wasn't sure entirely what to do about that style error on line 392. When I tried replacing == with 'is' it busted the entire routine and it never flagged anything, returning a zero-length set from the np.where. 'is' doesn't seem to be an appropriate replacement for == in this case. Is there some different syntax that should be used?

In that case we'll "call shenanigans" (to quote J. Eisenhamer) on the code style routine and override it to accept the as is code.

drlaw1558:
Ok, cleaned up some outdated doc stuff and moved the higher-level logging statement out of the else clause so that it applies regardless of which method is used instead of having dedicated log statements in the subroutine.

hbushouse (approved):
LGTM

PR #7827, opened 2023-08-16
Allow background accumulation for combinations of full and subarray observations

### LINKED ISSUE #7807: "background" fails when combining full and subarray data
Not sure if this is intended behaviour, but when trying to combine subarray science observations with full chip background observations the background subtraction step fails because the arrays are different sizes. An example for this are two of the MIRI observations for M33 (NGC0598) in program 2130, where SUB256 is used for two tiles in a mosaic, but the background observations use the FULL MIRI array.

thomaswilliamsastro:
Closes #7807

This PR changes the way background accumulation is handled, to properly extract regions of the chip if combining e.g. main chip and subarray data. Tested on imaging data (works in these edge cases, and is identical to the old behaviour otherwise). Spectra should work but haven't explicitly been tested. I'm not at STScI so can't run the full regression suite.

**Checklist for maintainers**
- [ ] added relevant label(s)


jwst/background/background_sub.py
```diff
@@ -57,14 +164,17 @@ def background_sub(input_model, bkg_list, sigma, maxiters):
...
-def average_background(bkg_list, sigma, maxiters):
+def average_background(input_model, bkg_list, sigma, maxiters):
     """
     Average multiple background exposures into a combined data model.
     Processes backgrounds from various DataModel types, including those
     having 2D (rate) or 3D (rateints) backgrounds.
 
     Parameters:
     -----------
+    input_model: JWST data model
```
hbushouse, line 175: Not the fault of this PR, but all of the param names in the list here should have whitespace on *both* sides of the colon, i.e. "input_model : JWST data model", to keep with doc string conventions.

jwst/background/background_sub.py
```diff
@@ -97,32 +210,37 @@ def average_background(bkg_list, sigma, maxiters):
     # Loop over the images to be used as background
     for i, bkg_file in enumerate(bkg_list):
         log.info(f'Accumulate bkg from {bkg_file}')
-        bkg_model = datamodels.open(bkg_file)
+
+        bkg_array = ImageSubsetArray(bkg_file)
+        if not bkg_array.overlaps(im_array):
```
hbushouse, line 215: I'm thinking it might be useful to report non-overlap cases in the log, at least at the debug level, if not info. Just so user has an understanding of what's being included in the average bkg.

jwst/background/background_sub.py
```diff
@@ -81,10 +191,13 @@ def average_background(bkg_list, sigma, maxiters):
         The averaged background exposure
     """
 
-    # Determine the dimensionality of the background files
-    bkg_model = datamodels.open(bkg_list[0])
-    bkg_dim = len(bkg_model.data.shape)
-    image_shape = bkg_model.data.shape[-2:]
+    # Determine the dimensionality of the input file
+    in_model = datamodels.open(input_model)
```
hbushouse, line 195: If `input_model` is an already open data model, I know it doesn't hurt to call `datamodels.open()` on it again, because that method is smart enough to do nothing when presented with an already open model, but then is it really necessary to be calling `datamodels.open` at all here? Can't `input_model` just be used directly, instead of the new `in_model`?

jwst/background/background_sub.py
```diff
@@ -81,10 +191,13 @@ def average_background(bkg_list, sigma, maxiters):
...
+    bkg_dim = len(in_model.data.shape)
+    image_shape = in_model.data.shape[-2:]
+    in_model.close()
```
hbushouse, line 198: I hope this doesn't have the indirect effect of closing the original `input_model` too.

hbushouse (commented):
Looks good overall. Just a few comments and questions.

hbushouse:
All the CI failures are due to a mismatch between jwst/master and the latest stdatamodels release. A new stdatamodels release is being created, which should fix this.

jwst/background/background_sub.py
thomaswilliamsastro: Corrected throughout background_sub.py, but might still be lurking elsewhere in the other step files

jwst/background/background_sub.py
thomaswilliamsastro: I'd missed this in just translating over the old logic, you're right that input_model is already a data model so this is unnecessary and has been changed

jwst/background/background_sub.py
thomaswilliamsastro: changed as above

thomaswilliamsastro:
Comments addressed, and I've rebased and pushed

hbushouse:
Thanks for the updates. Looks good. Wouldn't hurt to add a short blurb to https://github.com/spacetelescope/jwst/blob/master/docs/jwst/background_step/description.rst to mention that the step now has the capability to create an average bkg image from inputs of different (overlapping) sizes.

Looks like the failing CI tests are due to numpy deprecation warnings from other code modules, but still need to check them all to confirm.

hbushouse:
Started a regression test run at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/895

hbushouse:
Oh wait, regarding the failing CI tests. It does look like it's encountering a problem in the new code in this PR when running one of the unit tests:
`FAILED jwst/background/tests/test_background.py::test_nirspec_gwa - TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'`

which is due to:

`        # Account for the fact these things are 1-indexed
        self.jmin = im.meta.subarray.ystart - 1
>       self.jmax = im.meta.subarray.ysize + self.jmin
E       TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'`

Suggests that `ysize` is None for whatever data is being used in the unit test. Perhaps the YSIZE keyword is just missing from that test data file.

thomaswilliamsastro:
Could be, there's some code in there to catch if starts are missing but not sizes. Just made a quick edit on that which should hopefully catch it (unless subarray data, in which case it should likely error out anyway). Added a quick note in the description as well that this will use the overlapping regions

hbushouse:
In addition to adding a trap in the code for missing xsize/ysize values, it looks like the unit test should also be updated so that it sets those to begin with in the fake images it's creating. It sets xstart/ystart, but not xsize/ysize. For example, lines could be added here https://github.com/spacetelescope/jwst/blob/master/jwst/background/tests/test_background.py#L46 and https://github.com/spacetelescope/jwst/blob/master/jwst/background/tests/test_background.py#L72 to set `meta.subarray.xsize` and `ysize` by either hardwiring to the same 10x10 dimensions used to create the datamodel or set to `image.data.shape[-1]` and `image.data.shape[-2]` just to leave the code more general.

thomaswilliamsastro:
Done! I used the image shape in case these things change in the future

jwst/background/tests/test_background.py
```diff
@@ -45,6 +45,9 @@ def background(tmpdir_factory):
         image.meta.subarray.xstart = 1
         image.meta.subarray.ystart = 1
 
+        image.meta.subarray.xsize = image.data.shape[-1]
```
hbushouse, line 48: Need to also add these lines in the `science_image` function, down around line 75 in this module.

thomaswilliamsastro: whoops! sorted

hbushouse:
CI failures are unrelated.

PR #7845, opened 2023-08-22
JP-3311: OutlierDetectionStep producing extraneous outlier_i2d files

### LINKED ISSUE #7747: OutlierDetectionStep producing extraneous outlier_i2d files
_Issue [JP-3311](https://jira.stsci.edu/browse/JP-3311) was created on JIRA by [Jonathan Eisenhamer](https://jira.stsci.edu/secure/ViewProfile.jspa?name=eisenham):_

Through a user support call, it has been discovered that OutlierDetectionStep is creating files with the suffix and type of `_outlier_i2d`. This output does not appear to be documented. The creation of such files does not honor either the `save_results` or `save_intermediate_results` parameters. Even when both are `False`, the files are still created.

Suggestion that, as a minimum, this output be documented. Ideally, the creation should be regulated by a step parameter.

penaguerrero:
Resolves [JP-3311](https://jira.stsci.edu/browse/JP-3311)

Closes #7747 

This PR addresses the outlier_i2d file showing up regardless of save_intermediary_results.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/outlier_detection/outlier_detection_step.py
+++ b/jwst/outlier_detection/outlier_detection_step.py
@@ -71,7 +71,7 @@ class OutlierDetectionStep(Step):
         scale_detection = boolean(default=False)
         search_output_file = boolean(default=False)
         allowed_memory = float(default=None)  # Fraction of memory to use for the combined image
-        in_memory = boolean(default=True)
+        in_memory = boolean(default=False)
     """
 
     def process(self, input_data):
@@ -173,7 +173,7 @@ class OutlierDetectionStep(Step):
             state = 'COMPLETE'
             if self.input_container:
                 if not self.save_intermediate_results:
-                    self.log.info("The following files will be deleted since save_intermediate_results=False:")
+                    self.log.debug("The following files will be deleted since save_intermediate_results=False:")
                 for model in self.input_models:
                     model.meta.cal_step.outlier_detection = state
                     if not self.save_intermediate_results:
@@ -184,11 +184,11 @@ class OutlierDetectionStep(Step):
                         outlr_path = crf_file.replace(outlr_basename, outlr_file)
                         if os.path.isfile(outlr_path):
                             os.remove(outlr_path)
-                            self.log.info(f"    {outlr_path}")
+                            self.log.debug(f"    {outlr_path}")
                         blot_file = crf_file.replace('crf', 'blot')
                         if os.path.isfile(blot_file):
                             os.remove(blot_file)
-                            self.log.info(f"    {blot_file}")
+                            self.log.debug(f"    {blot_file}")
             else:
                 self.input_models.meta.cal_step.outlier_detection = state
             return self.input_models
--- a/jwst/outlier_detection/outlier_detection_step.py
+++ b/jwst/outlier_detection/outlier_detection_step.py
@@ -176,7 +176,7 @@ class OutlierDetectionStep(Step):
                     self.log.info("The following files will be deleted since save_intermediate_results=False:")
                 for model in self.input_models:
                     model.meta.cal_step.outlier_detection = state
-                    if not self.save_intermediate_results:
+                    if not self.in_memory and not self.save_intermediate_results:
                         # remove unwanted files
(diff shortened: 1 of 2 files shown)


jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -69,13 +71,16 @@ class OutlierDetectionStep(Step):
         scale_detection = boolean(default=False)
         search_output_file = boolean(default=False)
         allowed_memory = float(default=None)  # Fraction of memory to use for the combined image.
-        in_memory = boolean(default=False)
```
hbushouse, line 72: I'm not sure I like the idea of removing the `in_memory` param from the step. What if, for example, a user wants to avoid memory use (`in_memory = False`), but also doesn't want or need any of the intermediate files to persist after the step is done (`save_intermediate_results = False`). With this change they have no way to do that. If they don't want the intermediate files to persist, they are forced to have the step run with all the data in memory. If they want to conserve memory (or there isn't enough memory to run it that way), it's OK to have the intermediate files appear temporarily, you just want to clean them up at the end.

I would rather see both params remain in use and have the new code (to delete the files at the end) get activated when `in_memory = False` and `save_intermediate_files = False`.

penaguerrero: done

penaguerrero: I removed the if in memory requirement for deleting the files. This is only controlled by save_intermediate_results, though the user can choose to set in_memory = Files, and this will generate the files

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -68,14 +70,13 @@ class OutlierDetectionStep(Step):
         good_bits = string(default="~DO_NOT_USE")  # DQ flags to allow
         scale_detection = boolean(default=False)
         search_output_file = boolean(default=False)
-        allowed_memory = float(default=None)  # Fraction of memory to use for the combined image.
-        in_memory = boolean(default=False)
+        allowed_memory = float(default=None)  # Fraction of memory to use for the combined image
+        in_memory = boolean(default=True)
```
hbushouse, line 74: For use in the operations environment we want `in_memory` to default to `False`, in order to keep processing runs from throwing out-of-memory errors. So this should be set back to the original value. With `save_intermediate_results` also defaulting to `False`, those files should not be left behind in the operations environment.

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -171,8 +172,23 @@ def process(self, input_data):
 
             state = 'COMPLETE'
             if self.input_container:
+                if not self.save_intermediate_results:
+                    self.log.info("The following files will be deleted since save_intermediate_results=False:")
```
hbushouse, line 176: You could perhaps change these log statements (here and below in lines 187 and 191) to be debug level, instead of info. It's not really necessary to see these messages during routine processing.

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -171,8 +172,23 @@ def process(self, input_data):
...
                 for model in self.input_models:
                     model.meta.cal_step.outlier_detection = state
+                    if not self.save_intermediate_results:
+                        # remove unwanted files
+                        crf_file = self.make_output_path(basepath=model.meta.filename)
+                        outlr_basename = os.path.basename(crf_file)
+                        outlr_file = model.meta.filename.replace('cal', 'outlier_i2d')
+                        outlr_path = crf_file.replace(outlr_basename, outlr_file)
+                        if os.path.isfile(outlr_path):
```
hbushouse, line 185: Does this statement basically test to see if the file exists before trying to delete it?

jwst/outlier_detection/outlier_detection_step.py
penaguerrero: makes sense. Done

hbushouse (approved):
Looks good. Kicking off a regtest run to make sure nothing breaks.

hbushouse:
Regtest run was clean. Looks good.

PR #7858, opened 2023-08-29
Update MIRI MRS TSO regtest

hbushouse:
Related to [JP-3359](https://jira.stsci.edu/browse/JP-3359)

This PR updates the MIRI MRS TSO regression test to temporarily skip the MRS time-dependent correction in the photom step of the spec2 pipeline, because that algorithm can't handle 3D MRS data right now. Previously, the test had been working OK, because the DATE-OBS in the input file was pre-flight, such that an old photom ref file was used, which did not contain the time-dependent corrections, so no error occurred. This, however, allowed that error to get through our B9.3 testing and into operations. So I've updated the DATE-OBS in the test input file to use the most recent ref files, which will trigger the error, until it gets fixed by JP-3359.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/regtest/test_miri_mrs_tso.py
+++ b/jwst/regtest/test_miri_mrs_tso.py
@@ -25,6 +25,7 @@ def run_spec2(jail, rtdata_module):
             '--steps.srctype.save_results=true',
             '--steps.fringe.save_results=true',
             '--steps.photom.save_results=true',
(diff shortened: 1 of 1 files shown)


tapastro:
Should we have a set of regtests than can be run with a datetime override, such that the obstime is set to the current time? I assume this won't work for all tests, but it seems like it might be useful for test coverage of CRDS updates.

tapastro (approved):

PR #7863, opened 2023-08-31
JP-3233: Update level-3 product naming for NRS fixed-slit

### LINKED ISSUE #7701: Combine S200A1 and S200A2 spectra in level 3 products

hbushouse:
Resolves [JP-3233](https://jira.stsci.edu/browse/JP-3233)

Closes #7701 

PR #7795 changed the way level-3 ASN's are built for NIRSpec FS observations that put the target in both the S200A1 and S200A2 slits, such that the data from both slits is included in a single ASN to be combined during calwebb_spec3 processing. In addition to combining the data from both slits, the NIRSpec FS team also requested changes to the way the level-3 product names are constructed. Instead of having all source-based product names reference the primary slit, even for sources that are in the other slits, each product name should contain the name of the slit from it the data were extracted and always use source_id=1 for the primary target, regardless of which slit it's in.

So this PR is a companion to #7795 that makes the needed changes in the Cal code to implement the changes in source_id assignments for each slit and the construction of the level-3 product names. The changes to the source_id assignments for each slit are done in the assign_wcs/nirspec/get_open_fixed_slits function, which defines the meta attributes for each slit. The changes to the level-3 product names are done in the calwebb_spec3 pipeline module.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,9 @@ assign_wcs
 - Save bounding box to imaging WCS matching the shape of the data, for datamodels
   without a defined bounding box. [#7809]
 
+- Update NIRSpec fixed-slit `source_id` assignments for each `SlitModel`, so that
+  the slit with the primary target always has source_id=1. [#7863]
+
 associations
 ------------
 
@@ -34,6 +37,13 @@ calwebb_spec2
 - Run ``pixel_replace`` before setting metadata and suffix of datamodel
   that is returned by the pipeline to ensure a file is created with the
   expected ``_cal`` suffix. [#7772]
+
+calwebb_spec3
+-------------
+
+- Update the construction of output product names for NIRSpec fixed-slit observations,
+  so that each source-based product contains the name of the slit from the data were
+  extracted. [#7863]
 
 charge_migration
 ----------------
--- a/jwst/assign_wcs/nirspec.py
+++ b/jwst/assign_wcs/nirspec.py
@@ -413,7 +413,7 @@ def get_open_slits(input_model, reference_files=None, slit_y_range=[-.55, .55]):
     elif exp_type == "nrs_fixedslit":
         slits = get_open_fixed_slits(input_model, slit_y_range)
     elif exp_type == "nrs_brightobj":
-        slits = [Slit('S1600A1', 3, 0, 0, 0, slit_y_range[0], slit_y_range[1], 5, 4)]
+        slits = [Slit('S1600A1', 3, 0, 0, 0, slit_y_range[0], slit_y_range[1], 5, 1)]
     elif exp_type in ["nrs_lamp", "nrs_autowave"]:
         if lamp_mode in ['fixedslit', 'brightobj']:
             slits = get_open_fixed_slits(input_model, slit_y_range)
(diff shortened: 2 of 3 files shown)


hbushouse:
@stscieisenhamer I toyed with the idea of trying to change the `format_product` routine in associations to get it to somehow make the proper product name to begin with, e.g. by feeding it the `slit_name` corresponding to each product, thinking that would be the more elegant solution (as opposed to changing the name later). But it was not obvious to me how this could be done in a seamless way, given the fact that the product name listed in the input ASN already has a `slit_name` field hardwired into it (hence that existing name would have to be removed/replaced). I suppose we could change the product name created for the ASN such that it contains a {slit_name} field, like {source_id}, which would then be filled in when `format_product` is called. But that would require figuring out how to make the compound "s200a1-s200a2" name for these observations where the target is put into both slits. It would also need be made exclusive to NRS FS mode, cuz we don't want anything mucking around with, for example, the NRS MOS product names.

Any clever ideas from you would be most welcome. The change presented here in calwebb_spec3 is a bit of a hack just to make it work.

stscieisenhamer:
Doesn't [PR#7795](https://github.com/spacetelescope/jwst/pull/7795) handle the naming? Or is this another aspect of this?

hbushouse:
> Doesn't [PR#7795](https://github.com/spacetelescope/jwst/pull/7795) handle the naming? Or is this another aspect of this?

The changes in #7795 make a product name that comes out looking like:
jw02288-o003_{source_id}_nirspec_f100lp-g140h-s200a1-s200a2
in the spec3 ASN file. This naming, with the combined "s200a1-s200a2" slits is fine for data that actually comes from those 2 slits. But we're also extracting data from the other 3 slits, which right now just end up with different `source_id` values, but still have the hardwired "s200a1-s200a2" slit names in them, even though the data don't come from those slits. So this is an additional update to force the product names for each source/slit to have a slit name that makes more sense.

stscieisenhamer:
Yep, if the slit name is even partially up to the pipeline, the whole thing will need to be just left up to pipeline/step, like source_id.

So, because part of the aforementioned PR is the combining of the two slits, I would say that should get merged. Then, as part of this PR, fix the name creation portion so that the rule doesn't put the name in, and just add the {slit_name} field.

Question then: should the rule even care about separation along slit name? I.e. all exposures, regardless of slit, be put into the same L3 association?

hbushouse:
Closed in favor of #7879

Issue #7815, opened 2023-08-14
Allow users to scale MRS aperture extraction radii


stscijgbot-jp:
_Issue [JP-3340](https://jira.stsci.edu/browse/JP-3340) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

At present, MRS 1d spectral extraction for point sources is always performed using a conical aperture in which the radius is 2x the PSF FWHM at all wavelengths (based on a linear model of the FWHM outlined by Law et al. 2023).  However, it can be very useful to be able to use different radii, particularly at longer wavelengths where the high thermal background can easily dominate the counts within large apertures by factors of 50 or more.  While it's inadvisable to make the extraction aperture too small due to issues with resampling noise, some flexibility is important and the aperture correction reference files already provide scaling factors from 0.5x to 3x the PSF FWHM.  (Although see <https://jira.stsci.edu/browse/JP-3339)>

At the moment using different apertures is extremely clunky; the extract1d asdf reference file must be modified locally to specify the new aperture radii, and then the pipeline overridden to use this locally edited reference file.

This ticket is to include an easy command line option in the arguments to extract_1d to allow users to scale the default aperture extraction radius by a fixed amount.  E.g., ifu_rscale might define the number of PSF FWHM to use for the extraction radius, accept values in the range of 0.5 to 3.0, and have a default value of 2.0.

stscijgbot-jp:
_Comment by [Jane Morrison](https://jira.stsci.edu/secure/ViewProfile.jspa?name=morrison) on [JIRA](https://jira.stsci.edu/browse/JP-3340?focusedCommentId=661512#comment-661512):_

[David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw)  I am assuming only the extraction radius is scaled. The inner_bkg and outer_bkg are not scaled or changed at all - CORRECT ?

stscijgbot-jp:
_Comment by [Jane Morrison](https://jira.stsci.edu/secure/ViewProfile.jspa?name=morrison) on [JIRA](https://jira.stsci.edu/browse/JP-3340?focusedCommentId=661513#comment-661513):_

[David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw) 

ifu_rscale can have values of 0.5 to 3. These are the psf FWHM.

The default radius was created with a value of 2.

So really the scaling factor is ifu_rscale/2 - Correct ?
[...]

PR #7883, opened 2023-09-11
JP-3355 added parameter to extract 1d to override srctype and  use point source extraction 

### LINKED ISSUE #7846: Allow extract_1d to override POINT vs EXTENDED
_Issue [JP-3355](https://jira.stsci.edu/browse/JP-3355) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

<https://jira.stsci.edu/browse/JP-1750> identified the ability to override POINT vs EXTENDED directly in Extract1D as a desirable functionality, but it looks like that part of the ticket got lost in the final implementation of other things.

This ticket is to renew this requested functionality; it is often useful to process nominally-extended source data through the point-source pipeline if the central source is sufficiently point like.  At present, the only workaround is to reprocess the data from the srctype step in spec2, or to hand-hack the headers of the cubes, which are rather more complicated than simply passing a keyword override.

jemorrison:
Resolves [JP-3355](https://jira.stsci.edu/browse/JP-3355) and Resolves [JP-3340](https://jira.stsci.edu/browse/JP-3340)

Closes #7846 and #7815

This PR allows the user to override the srctype and perform POINT source extraction for IFU data. 

**Checklist for maintainers**
- [x] added relevant label(s)
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/962/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,12 @@
 1.12.1 (unreleased)
 ===================
 
-- 
+extract_1d
+----------
+
+- For IFU data allow the user to set the src_type. For MIRI MRS data allow
+  the user to scale the extraction radius between 0.5 to 3.0 FWHM [#7883]
+ 
 
 1.12.0 (2023-09-18)
 ===================
@@ -111,6 +116,7 @@ extract_1d
 
 - Use ``source_{x/y}pos`` metadata to center extraction region for NIRSpec
   (non-IFU) data; use dithered pointing info for MIRI LRS fixed slit data. [#7796]
+
 
 extract_2d
 ----------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@
 extract_1d
 ----------
 
-- For IFU data allow the user to set the src_type. For MIRI MRS data allow
+- For MIRS MRS IFU data allow the user to set the src_type and allow 
   the user to scale the extraction radius between 0.5 to 3.0 FWHM [#7883]
  
 
--- a/docs/jwst/extract_1d/arguments.rst
+++ b/docs/jwst/extract_1d/arguments.rst
@@ -115,9 +115,9 @@ The ``extract_1d`` step has the following step-specific arguments.
   not changed, instead the extraction method used is based on this parameter setting. This is
   only allowed for IFU data. 
 
-``ifu_rscale``
+``--ifu_rscale``
    A float designating the number of PSF FWHMs to use for the extraction radius. This
-   is  MIRI MRS only paramenter. Values accepted are between 0.5 to 3.0. The default extraction
+   is a MIRI MRS only paramenter. Values accepted are between 0.5 to 3.0. The default extraction
    size is set to 2 * FWHM. Values below 2 will result in a smaller
    radius, a value of 2 results in no change to radius and a value above 2 results in a larger
    extraction radius.  
--- a/docs/jwst/extract_1d/arguments.rst
+++ b/docs/jwst/extract_1d/arguments.rst
@@ -109,6 +109,19 @@ The ``extract_1d`` step has the following step-specific arguments.
   Switch to select whether or not to run 1d residual fringe correction on the
   extracted 1d spectrum (MIRI MRS only). Default is ``False``.
 
+``--ifu_set_srctype``
+  A string that can be used to override the extraction method for the source_type
+  given by the SRCTYPE keyword. The allowed values are POINT and EXTENDED. The SRCTYPE keyword is
+  not changed, instead the extraction method used is based on this parameter setting. This is
+  only allowed for IFU data. 
+
+``ifu_rscale``
+   A float designating the number of PSF FWHMs to use for the extraction radius. This
+   is  MIRI MRS only paramenter. Values accepted are between 0.5 to 3.0. The default extraction
+   size is set to 2 * FWHM. Values below 2 will result in a smaller
+   radius, a value of 2 results in no change to radius and a value above 2 results in a larger
+   extraction radius.  
+   
 ``--soss_atoca``
   This is a NIRISS-SOSS algorithm-specific parameter; if True, use the ATOCA
   algorithm to treat order contamination. Default is ``True``.
--- a/docs/jwst/extract_1d/arguments.rst
+++ b/docs/jwst/extract_1d/arguments.rst
@@ -113,7 +113,7 @@ The ``extract_1d`` step has the following step-specific arguments.
   A string that can be used to override the extraction method for the source_type
   given by the SRCTYPE keyword. The allowed values are POINT and EXTENDED. The SRCTYPE keyword is
   not changed, instead the extraction method used is based on this parameter setting. This is
-  only allowed for IFU data. 
+  only allowed for MIRI MRS IFU data. 
 
 ``--ifu_rscale``
    A float designating the number of PSF FWHMs to use for the extraction radius. This
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -2804,7 +2804,7 @@ def do_extract1d(
         for MIRI MRS IFU spectra.  Default is False.
 
     ifu_set_srctype : str
-        For IFU data override srctype and set it to either POINT or EXTENDED.
+        For MIRI MRS IFU data override srctype and set it to either POINT or EXTENDED.
 
     ifu_rscale: float
         For MRS IFU data a value for changing the extraction radius. The value provided is the number of PSF
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -2547,7 +2547,7 @@ def run_extract1d(
         center_xy: Union[float, None],
         ifu_autocen: Union[bool, None],
         ifu_rfcorr: Union[bool, None],
-        ifu_set_point_source: Union[bool, None],
+        ifu_set_srctype: str,
         ifu_rscale: float,
         was_source_model: bool = False,
 ) -> DataModel:
@@ -2609,9 +2609,8 @@ def run_extract1d(
         Switch to select whether or not to apply a 1d residual fringe correction
         for MIRI MRS IFU spectra.  Default is False.
 
-    ifu_set_point_source : bool
-        Override srctype and set POINT source extraction for IFU spectra.
-        Default is False.
+    ifu_set_srctype : str
+        For IFU data override srctype and set it to either POINT or EXTENDED.
 
     ifu_rscale: float
         For MRS IFU data a value for changing the extraction radi. The value provided is the number of PSF
@@ -2668,7 +2667,7 @@ def run_extract1d(
         center_xy,
         ifu_autocen,
         ifu_rfcorr,
-        ifu_set_point_source,
+        ifu_set_srctype,
         ifu_rscale,
         was_source_model,
     )
@@ -2732,7 +2731,7 @@ def do_extract1d(
         center_xy: Union[int, None] = None,
         ifu_autocen: Union[bool, None] = None,
         ifu_rfcorr: Union[bool, None] = None,
-        ifu_set_point_source: Union[bool, None] = None,
+        ifu_set_srctype: str = None,
         ifu_rscale: float = None,
         was_source_model: bool = False
 ) -> DataModel:
@@ -2804,9 +2803,8 @@ def do_extract1d(
         Switch to select whether or not to apply a 1d residual fringe correction
         for MIRI MRS IFU spectra.  Default is False.
 
-    ifu_set_point_source : bool
-        Override srctype and perform POINT source extraction for IFU spectra.
-        Default is False.
+    ifu_set_srctype : src
+        For IFU data override srctype and set it to either POINT or EXTENDED.
 
     ifu_rscale: float
         For MRS IFU data a value for changing the extraction radi. The value provided is the number of PSF
@@ -3022,8 +3020,8 @@ def do_extract1d(
             if source_type is None:
                 source_type = "UNKNOWN"
 
-            if ifu_set_point_source:
-                source_type = "POINT"
+            if ifu_set_srctype is not None:
+                source_type = ifu_set_srctype
                 log.info('Overriding source type and setting it to POINT')
             output_model = ifu.ifu_extract1d(
                 input_model, extract_ref_dict, source_type, subtract_background,
--- a/jwst/extract_1d/extract.py
+++ b/jwst/extract_1d/extract.py
@@ -2610,7 +2610,7 @@ def run_extract1d(
         for MIRI MRS IFU spectra.  Default is False.
 
     ifu_set_srctype : str
-        For IFU data override srctype and set it to either POINT or EXTENDED.
+        For MIRI MRS IFU data override srctype and set it to either POINT or EXTENDED.
 
     ifu_rscale: float
         For MRS IFU data a value for changing the extraction radius. The value provided is the number of PSF
@@ -2803,7 +2803,7 @@ def do_extract1d(
         Switch to select whether or not to apply a 1d residual fringe correction
         for MIRI MRS IFU spectra.  Default is False.
 
-    ifu_set_srctype : src
+    ifu_set_srctype : str
         For IFU data override srctype and set it to either POINT or EXTENDED.
 
     ifu_rscale: float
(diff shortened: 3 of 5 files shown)


jemorrison:
Will update documentation next

jemorrison:
I am going to add the work needed for JP-3340 since it is so similar to is needed for this ticket.

jwst/extract_1d/extract.py
```diff
@@ -2586,6 +2588,17 @@ def run_extract1d(
         Switch to select whether or not to apply a 1d residual fringe correction
         for MIRI MRS IFU spectra.  Default is False.
 
+    ifu_set_point_source : bool
```
hbushouse, line 2591: Instead of having a param that just hardwires the srctype to POINT (even if it's already set to POINT upstream), would it be useful and more flexible to just have the param allow the user to choose which srctype they want, i.e. either POINT or EXTENDED? Hence instead of a boolean that just hardwires to POINT, make it a string with two allowed options of POINT and EXTENDED.

tapastro:
Is this worth including over recommending reprocessing spec2 with a srctype step parameter, or having a user modify the header value prior to running extract_1d? I feel like this is an overreach for extract_1d parameters, which is already a bloated list.

drlaw1558:
This modification is based on a number of user interactions and their frustration with it being so hard to change the spectral extraction behaviour; it's also helpful from a testing perspective.

jwst/extract_1d/extract.py
jemorrison: I thought about that but I can not think when anyone would want the EXTENDED code and extracting the entire field of view. But I can put it in

jemorrison:
I have had to modify the SRCTYPE to POINT for all the data I have worked with and then set up a specialized extract 1d reference file with a scaled radius. IT IS A PAIN! Having 2 options to extract1d would be so much easier

jwst/extract_1d/extract.py
hbushouse: You could make it a generic param for setting the srctype, that would then get passed to and used by all modes.

jemorrison:
@jmuzerolle @melanieclarke @leonardoubeda I am making these changes to extract1d  for MIRI MRS data(option to change to POINT Source, option to scale extraction radius). The NIRSpec aperture correction has multiple radii to use so it could also handle having a scaling radius.  But this code change is focused on MIRI because of how the default extraction radii is set. Looking at the NIRSpec App corr files they are still marked as DUMMY and there are 3 fix radius extractions (0.3, 0.5, 1.0). I don't know if the team has done enough study to want to scale the extraction radii. Unless I hear otherwise I will keep these options as only MIRI MRS options

jwst/extract_1d/extract.py
tapastro: Is that not what the [srctype step parameter](https://jwst-pipeline.readthedocs.io/en/latest/jwst/srctype/arguments.html) is already? If a user sets this, will it not propagate to all future steps?

jemorrison:
This is really an issue with how the data is extracted. If we change it in SRC type step then if there are point source specific steps applied that would happened and we may not want that. I am thinking of NIRSpec pathloss mostly. At least that is my take on it. We just want to have an extraction radius and that is only possible in extract 1d with a POINT source type of data.

hbushouse:
So given all the commentary regarding how *everyone* has to rerun extract_1d on their MRS cubes to have the extraction done as a point source, why are we ever setting SRCTYPE='EXTENDED' to begin with? How much would it affect upstream processing if the srctype step were just hardwired to always set it to point source for every MRS exposure? Alternatively, if the extended source version of IFU extraction in extract_1d is totally worthless, then we should consider removing it, or at least changing the logic in the step so that it just always uses point source extraction for MRS, period.

drlaw1558:
I realize I should clarify; it's not that everyone has to rerun extract_1d.  Many programs can use the extended source extraction just fine, or at least it's as good as possible given their scene geometry of fuzzy stuff.  It's just that in a number of cases there is a complex extended-type scene with multiple point sources within it, or a single source that turns out to be only slightly extended and can for most purposes be treated as a point source.  In those cases it's helpful to be able to use the pipeline to run point-source like extractions at specific locations in the scene rather than users rebuilding the wheel trying to figure out how to use the aperture correction file outside the pipeline.

jemorrison:
@hbushouse @drlaw1558 I made all the suggested changes. I tested it some. David you might want to test it more.
I will see about getting this run through the regression tests today

docs/jwst/extract_1d/arguments.rst
```diff
@@ -109,6 +109,19 @@ The ``extract_1d`` step has the following step-specific arguments.
   Switch to select whether or not to run 1d residual fringe correction on the
   extracted 1d spectrum (MIRI MRS only). Default is ``False``.
 
+``--ifu_set_srctype``
+  A string that can be used to override the extraction method for the source_type
+  given by the SRCTYPE keyword. The allowed values are POINT and EXTENDED. The SRCTYPE keyword is
+  not changed, instead the extraction method used is based on this parameter setting. This is
+  only allowed for IFU data. 
+
+``ifu_rscale``
```
hbushouse, line 118: Needs a dash-dash out front, for consistency with others in the list

docs/jwst/extract_1d/arguments.rst
```diff
@@ -109,6 +109,19 @@ The ``extract_1d`` step has the following step-specific arguments.
...
+   A float designating the number of PSF FWHMs to use for the extraction radius. This
+   is  MIRI MRS only paramenter. Values accepted are between 0.5 to 3.0. The default extraction
```
hbushouse, line 120: needs an "a" between "is" and "MIRI"

jwst/extract_1d/extract.py
```diff
@@ -2973,9 +2999,12 @@ def do_extract1d(
             if source_type is None:
                 source_type = "UNKNOWN"
 
+            if ifu_set_srctype is not None:
+                source_type = ifu_set_srctype
+                log.info('Overriding source type and setting it to POINT')
```
hbushouse, line 3004: the log message should not be hardwired to say "POINT", because the user can set it to either POINT or EXTENDED

jwst/extract_1d/extract_1d_step.py
```diff
@@ -81,6 +81,16 @@ class Extract1dStep(Step):
         Switch to select whether or not to apply a 1d residual fringe correction
         for MIRI MRS IFU spectra.  Default is False.
 
+    ifu_set_srctype: src
```
hbushouse, line 84: "src" should be "str" for the data type

hbushouse (commented):
Still needs a change log entry

CHANGES.rst
```diff
@@ -90,6 +90,11 @@ engdb_tools
 - Check alternative host is alive before attempting to run test for
   access to avoid waiting the full timeout during test runs [#7780]
 
+extract_1d
+----------
+- For IFU data allow the user to set the src_type. For MIRI MRS data allow
```
hbushouse, line 95:
```suggestion

- For IFU data allow the user to set the src_type. For MIRI MRS data allow
```

jwst/extract_1d/extract.py
```diff
@@ -2766,6 +2782,16 @@ def do_extract1d(
         Switch to select whether or not to apply a 1d residual fringe correction
         for MIRI MRS IFU spectra.  Default is False.
 
+    ifu_set_srctype : src
```
hbushouse, line 2785:
```suggestion
    ifu_set_srctype : str
```

jwst/extract_1d/extract_1d_step.py
```diff
@@ -149,6 +159,8 @@ class Extract1dStep(Step):
     apply_apcorr = boolean(default=True)  # apply aperture corrections?
     ifu_autocen = boolean(default=False) # Auto source centering for IFU point source data.
     ifu_rfcorr = boolean(default=False) # Apply 1d residual fringe correction
+    ifu_set_srctype = option("POINT", "EXTENDED", None, default=None) # user-supplied source type
+    ifu_rscale = float(default=None, min=0.5, max=3) Radius in terms of PSF FWHM to scale extraction radii
```
hbushouse, line 163:
```suggestion
    ifu_rscale = float(default=None, min=0.5, max=3) # Radius in terms of PSF FWHM to scale extraction radii
```

hbushouse: I think it would be useful to add a little if/then block down around line 230 in this top-level step module (after all the checking is done on input datamodel types) so that a log warning message is issued if either `ifu_set_srctype` or `ifu_rscale` have been set by the user and `meta.exposure.type != "MIR_MRS"` to let them know that the values they supplied don't apply to this exp_type and will be ignored (rather than just ignoring them silently).

jemorrison: good suggestion. I will do that

jemorrison:
@leonardoubeda @jmuzerolle @bethanjames 
I have made a few changes to extract1d to support MIRI MRS extraction changes.

This pr allows the user for IFU to change from EXTENDED to POINT source (or vice versa). Is this a feature NIRSpec IFU should allow ?

leonardoubeda:
@jemorrison With regards to the comment:
"This pr allows the user for IFU to change from EXTENDED to POINT source (or vice versa). Is this a feature NIRSpec IFU should allow ?"
I was thinking about the fact that NIRSpec IFU treats EXTENDED and POINT differently when calculating fluxes (https://github.com/spacetelescope/jwst/pull/7569) and may be allowing users to select the source type might result in wrong fluxes without intention to do so.

jemorrison:
@leonardoubeda Yeah I was concerned about that as well. This option to switch to POINT to EXTENDed or vice versa is probably just something MIRI MRS people would want to try. But it is only changing the fact that the extraction can be done using a radius and not the entire FOV of the s3d cube. But since an aperture correction would then also be applied which is
more for POINT like data (unless that is turned off), it might be less confusing to people working with NIRSpec data to just go back and change the SRCTYPE in the calspec2 pipeline. So I will leave this PR as it is and only allowed for MIRI MRS data

jemorrison:
@hbushouse The regression test are failing for this PR for NIRISS data and I don't understand why. It is related to the new parameters I added. 
Here is a link to the failures:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/957/testReport/junit/jwst.regtest/test_niriss_wfss/_stable_deps__test_nis_wfss_spec3_s00015_cal_/

drlaw1558 (approved):
LGTM.  Tested the pt/extended swap and also that various choices of scaling radius give results that match what I get when overriding the relevant reference file to implement the change by hand.  Seems to all be working as expected.

jemorrison:
@hbushouse Here are the results of the regression tests. The two failures do not seem related to this PR- NIRSpec MSA tests that can not find the msa meta file. https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/958/

hbushouse:
@jemorrison There are EIGHT failures in your regtest run, not just 2. The first 2 are unrelated, as you say, but the remaining 6 are coming from the new extract_1d code trying to work on WFSS data:
```
                # --------------------------------------------------------------
                # Data is WFSS
                if input_model[0].meta.exposure.type in extract.WFSS_EXPTYPES:
    
                    # For WFSS level-3, the input is a single entry of a
                    # SourceContainer, which contains a list of multiple
                    # SlitModels for a single source. Send the whole list
                    # into extract1d and put all results in a single product.
                    apcorr_ref = (
                        self.get_reference_file(input_model[0], 'apcorr') if self.apply_apcorr is True else 'N/A'
                    )
    
                    if apcorr_ref == 'N/A':
                        self.log.info('APCORR reference file name is "N/A"')
                        self.log.info('APCORR will NOT be applied')
                    else:
                        self.log.info(f'Using APCORR file {apcorr_ref}')
    
                    extract_ref = 'N/A'
                    self.log.info('No EXTRACT1D reference file will be used')
    
                    result = extract.run_extract1d(
                        input_model,
                        extract_ref,
                        apcorr_ref,
                        self.smoothing_length,
                        self.bkg_fit,
                        self.bkg_order,
                        self.bkg_sigma_clip,
                        self.log_increment,
                        self.subtract_background,
                        self.use_source_posn,
                        self.center_xy,
                        self.ifu_autocen,
                        self.ifu_rfcorr,
                        self.ifu_set_srctype,
>                       self.ifu_rscsale,
                        was_source_model=was_source_model
                    )
E                   AttributeError: 'Extract1dStep' object has no attribute 'ifu_rscsale'
```
I guess the new `ifu_rscale` attribute  hasn't been added to the step object properly?

jemorrison:
@hbushouse Ok now  there are only 2 failures:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/962/

And those failures are because the MSA file can not be found.

hbushouse (approved):
All looks good now. So I guess the final consensus on whether or not to make the srctype option available for NRS_IFU was no?

drlaw1558:
I'm getting some major crashes after this merge that I hadn't seen in testing.  Extract1d runs in a standalone manner, but not as part of the spec3 pipeline; crashing at
232 if self.ifu_rfcorr:
--> 233     if input_model.meta.exposure.type != "MIR_MRS":
with 
AttributeError: No attribute 'exposure'

Was this exposure type test introduced for the first time here?

hbushouse:
> I'm getting some major crashes after this merge that I hadn't seen in testing. Extract1d runs in a standalone manner, but not as part of the spec3 pipeline; crashing at 232 if self.ifu_rfcorr: --> 233 if input_model.meta.exposure.type != "MIR_MRS": with AttributeError: No attribute 'exposure'
> 
> Was this exposure type test introduced for the first time here?

It might be due to the fact that the combined s3d file created during level-3 processing doesn't have the `exposure` attribute carried over into the meta of the data model? It should have all original meta copied over, but that's my guess.

tapastro:
If the input is a ModelContainer, you may need to access a model in the container to check meta properties. Wouldn't this be seen in regression testing?

hbushouse:
> If the input is a ModelContainer, you may need to access a model in the container to check meta properties. Wouldn't this be seen in regression testing?

Those new params are set to None by default, so the if block wouldn't be executed in normal tests. Obviously need a test that exercises the new params.

jemorrison:
Dang. Ok I will add a test and make figure out what is needed when the s3d is created from multiple channels.

jemorrison:
Working on this now to understand the error. In all cases whether calspec2 or calspec3 the output from cube_build  is an IFU model. But it could pass a Container of IFU models if the association contains multiple Channel cubes.

jemorrison:
How does this seem. I will put in a PR for it

if isinstance(input_model, datamodels.IFUImageModel):
            exp_type = input_model.meta.exposure.type
 elif isinstance(input_model, ModelContainer):
            exp_type = input_model[0].meta.exposure.type
 else:
            exp_type = None

 if self.ifu_rfcorr:
            if exp_type != "MIR_MRS":
                self.log.warning("The option to apply a residual refringe correction is"
                                 f" not supported for {input_model.meta.exposure.type} data.")

 if self.ifu_rscale is not None:
            if exp_type != "MIR_MRS":
                self.log.warning("The option to change the extraction radius is"
                                 f" not supported for {input_model.meta.exposure.type} data.")

 if self.ifu_set_srctype is not None:
            if exp_type != "MIR_MRS":
                self.log.warning("The option to change the source type is"
                                 f" not supported for {input_model.meta.exposure.type} data.")

From: Howard Bushouse ***@***.***>
Sent: Thursday, September 21, 2023 9:10 AM
To: spacetelescope/jwst ***@***.***>
Cc: Jane Morrison ***@***.***>; Mention ***@***.***>
Subject: [EXT]Re: [spacetelescope/jwst] JP-3355 added parameter to extract 1d to override srctype and use point source extraction (PR #7883)


External Email

External Email

But by the time the combined s3d cube gets fed to extract_1d during calwebb_spec3 processing, shouldn't it be just a single model, as opposed to a model container? The original input to calwebb_spec3 will be a model container, loaded from the list of exposures in the ASN file. But I would assume that the output of cube_build is a single model, because the exposures have been combined. Or does it create multiple cubes, one for each channel/band combination? If the input to extract_1d during level-3 is a container, then you just need something like input_model[0].meta.exposure.type, i.e. just pick the first one. They'd better all be the same EXP_TYPE!

PR #7905, opened 2023-09-15
JP-3373: Add pixelarea warnings to photom and source_catalog

### LINKED ISSUE #7903: Photom needs to warn when PIXAR_xx keywords are missing.
_Issue [JP-3373](https://jira.stsci.edu/browse/JP-3373) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

It's come to our attention that if someone creates and delivers a PHOTOM ref file to CRDS that's missing the PIXAR_SR and PIXAR_A2 keywords, the photom step just quietly goes on its merry way without saying a word. The photom step itself doesn't need these values for anything it's doing, but downstream steps, like resample and source_catalog, rely on them. So if they aren't populated by photom, processing crashes later on in the pipeline for no obvious reason. The photom step needs to at least throw a big red warning when those keywords are missing from the ref file.

hbushouse:
Resolves [JP-3373](https://jira.stsci.edu/browse/JP-3373)

Closes #7903 

This PR adds log warnings to the photom and source_catalog steps when the pixelarea meta data is not available. In the case of photom, this means the keywords were not included in the PHOTOM ref file. For source_catalog, it means they did not get passed down from the photom step.

Note that the photom step code already had a try/except block around the retrieval of the keyword value, but that doesn't work, because there's always an attribute in the data model even when the keyword is absent. In that case the meta attribute is None. So the check has been changed to look for None.

**Checklist for maintainers**
- [x] added relevant label(s)


hbushouse:
Started regtest at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/950

hbushouse:
Regression test results are fine. All failures are due to a new MIRI GAIN ref file.

braingram (approved):
LGTM.

braingram (approved):
Looks like I do not have the power?

PR #7910, opened 2023-09-18
reorganize tests and prepare for future inclusion of data caching

zacharyburnett:
Resolves [SCSB-79](https://jira.stsci.edu/browse/SCSB-79)

This PR reorganizes CI tests and moves devdeps tests to a new optional workflow that can be started with a label

**Checklist for maintainers**
- [x] ~added entry in `CHANGES.rst` within the relevant release section~
- [x] ~updated relevant documentation~
- [x] added relevant label(s)
- [x] ~ran regression tests, post a link to the Jenkins job below.
      [How to run regression tests on a PR](https://github.com/spacetelescope/jwst/wiki/Running-Regression-Tests-Against-PR-Branches)~


.github/workflows/tests_devdeps.yml
```diff
@@ -0,0 +1,66 @@
+name: tests with development versions of dependencies
```
zacharyburnett, line 1:
```suggestion
name: test with development versions
```

.github/workflows/ci_cron.yml
```diff
@@ -1,4 +1,4 @@
-name: weekly tests
+name: scheduled tests
```
zacharyburnett, line 1:
```suggestion
name: test on schedule
```

hbushouse:
Why all the CI failures? Looks like something to do with our pinning of the python version <3.12.

zacharyburnett:
in the regular tests it looks like an issue with `timecoeff`? 

this one is from `py311-xdist` on Ubuntu:
https://github.com/spacetelescope/jwst/actions/runs/7062856370/job/19227602452#step:10:410
```
  =========================== short test summary info ============================
  FAILED jwst/photom/tests/test_photom.py::test_miri_image - AttributeError: No attribute 'timecoeff'
  = 1 failed, 2705 passed, 532 skipped, 4 xfailed, 110 warnings in 952.53s (0:15:52) =
  ```
  
Then in the devdeps tests, there's the issue you said for the `py3` tests and also the error with using Numpy 1.x at build time and Numpy 2.x at run time; I'll look into how to resolve that

zacharyburnett:
If the regular tests pass here, we can likely merge this and then deal with the Numpy 2.0 devdeps issues in another PR

hbushouse:
So the remaining CI errors all seem to be due to numpy.core.multiarray import errors. Is that what you expect and hence is this ready to merge as-is?

zacharyburnett:
> So the remaining CI errors all seem to be due to numpy.core.multiarray import errors. Is that what you expect and hence is this ready to merge as-is?

yes, I think that's expected. This is ready to merge, and then we can fix the devdeps

PR #6847, opened 2022-05-16
Consolidate build config into `pyproject.toml` using `poetry-core`

zacharyburnett:
**Description**

These changes test building `jwst` with Poetry using the `poetry-core` backend. This is an experimental change that is not yet ready for review.

Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Change log
- [ ] Milestone
- [ ] Label(s)

PR #7912, opened 2023-09-19
use PyPI upload workflow from OpenAstronomy

### LINKED ISSUE #8054: Moar release wheels
Pipeline release should provide wheels for more architectures (maybe at least Mac M1/M2 if not Windows given your user base) and also for Python 3.11+. See https://pypi.org/project/jwst/#files

This would be much easier to accomplish if you switch from this:

https://github.com/spacetelescope/jwst/blob/5f01af445b365390122c44fe53b7dc886a981743/.github/workflows/publish-to-pypi.yml#L17

To this:

https://github.com/OpenAstronomy/github-actions-workflows#build-and-publish-a-python-package

Example implementation in `astropy`:

https://github.com/astropy/astropy/blob/be40cc47ded9727b75fff268a41b74c78a48df75/.github/workflows/publish.yml#L30

zacharyburnett:
Resolves [SCSB-108](https://jira.stsci.edu/browse/SCSB-108)

closes #8054 

This PR changes the PyPI upload process to use the OpenAstronomy workflow instead of the STScI one. The workflow in this PR is configured to test the build with each pull request push, and then publish upon an official release.

This will facilitate catching build issues early on (ideally within the PR) before the release date. The reusable workflow uses `cibuildwheel` to build wheels for all supported Pythons on Linux, macOS X64, and macOS ARM64 (we don't currently have automated testing for ARM64, but we should soon with the on-prem regtesting effort)

~We can't use this workflow until we make a PyPI token (with write / upload permission to packages) for the STScI maintainer account. Currently, we use a username and password to PyPI; however, this method will be deprecated at some point in the future in favor of tokens.~

**Checklist for maintainers**
- [x] ~added entry in `CHANGES.rst` within the relevant release section~
- [x] ~updated or added relevant tests~
- [x] ~updated relevant documentation~
- [x] added relevant label(s)
- [x] ~ran regression tests, post a link to the Jenkins job below.
      [How to run regression tests on a PR](https://github.com/spacetelescope/jwst/wiki/Running-Regression-Tests-Against-PR-Branches)~
- [x] ~Make sure the JIRA ticket is [resolved properly](https://github.com/spacetelescope/jwst/wiki/How-to-resolve-JIRA-issues)~


zacharyburnett:
The following error occurred while building:
https://github.com/spacetelescope/jwst/actions/runs/6262235017/job/17003963430#step:7:703
```
                  Optimized (vendor) Blas libraries are not found.
                  Falls back to netlib Blas library which has worse performance.
                  A better performance should be easily gained by switching
                  Blas library.
                if self._calc_info(blas):
              /tmp/pip-install-x3ogdb8y/numpy_88331d5ab39a42d0ade0340a4d4d111e/numpy/distutils/system_info.py:1914: UserWarning:
                  Blas (http://www.netlib.org/blas/) libraries not found.
                  Directories to search for the libraries can be specified in the
                  numpy/distutils/site.cfg file (section [blas]) or by setting
                  the BLAS environment variable.
```

zacharyburnett:
using CPython fixed the issue

.github/workflows/build.yml
```diff
@@ -0,0 +1,31 @@
...
+    types: [ released ]
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  check:
+    name: check that `requirements-sdp.txt` is populated
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
zacharyburnett, line 14:
```suggestion
      - uses: actions/checkout@v4
```

zacharyburnett:
the source distribution issue in the OpenAstronomy workflow is now resolved (see https://github.com/OpenAstronomy/github-actions-workflows/pull/169)

.github/workflows/build.yml
```diff
@@ -0,0 +1,32 @@
...
+  check:
+    if: github.event_name == 'release'
+    name: check that `requirements-sdp.txt` is populated
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: grep -v '^ *#' requirements-sdp.txt
+  build:
+    needs: [ check ]
+    if: always() && (needs.check.result == 'success' || needs.check.result == 'skipped')
```
pllim, line 19: I don't grok this condition but whatever works for you, I guess.

.github/workflows/ci.yml
```diff
@@ -51,7 +51,6 @@ jobs:
         - linux: check-style
         - linux: check-security
         - linux: check-dependencies
-        - linux: build-dist
```
pllim, line 54: Should also remove it from tox.ini since it is no longer used.

pllim (commented):
Still no Windows support? Didn't someone fixed it being broken on Windows at some point? Is it broken again?

.github/workflows/build.yml
```diff
@@ -0,0 +1,32 @@
+name: build
+
+on:
+  release:
+    types: [ released ]
+  pull_request:
+  workflow_dispatch:
+
```
pllim, line 8:
```suggestion

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

```

pllim (commented):
Also if this is a high traffic repo and building the wheels is resource intensive, should consider only building them when a special label is applied for pull requests. Probably also wise to auto-cancel duplicate builds. See `astropy` on how this is done in production.

.github/workflows/build.yml
zacharyburnett: this allows the `requirements-sdp.txt` check job to be skipped (for a PR or manual dispatch) but still trigger the `build` job

pllim: I have no idea what that requirements file is for, so I'll leave that to your good judgement.

pllim (commented):
I left what comments I can. Someone from your team should approve.

I think this fixes #8054 . Thanks!

nden:
Is the dev job fixed in a different PR?
Changes look good. I'm OK with merging as is and fixing the dev tests in another PR if that's what @zacharyburnett prefers.

pllim:
Re: devdeps -- You need pyerfa-dev, see:

* https://github.com/astropy/astropy/issues/15638

But even with that, I don't think you can resolve the numpy ABI failure. You have a lot of dependencies that build their own C-extensions. Anything that builds that need to be *built* with numpy 2.0 to work with numpy 2.0 (drizzle, spherical-geometry, scikit-image, etc). NumPy claims that stuff built with numpy 2.0 would be backward compatible with 1.x (up to a point).

And if you have your own C-extensions, you might also run into compatibility issues in your own build after you resolve the dependencies, so be on the lookout for that.

I know that scikit-image is still struggle to provide a dev wheel that is compatible with numpy 2.0, please follow:

* https://github.com/scikit-image/scikit-image/pull/7251

tl;dr -- You probably can merge this as-is and deal with devdeps later because it is going to be painful.

zacharyburnett:
Ok, I will merge this as-is then, and work on #7910 and #6847

nden (approved):
Approving - given that dev builds are going to be handled in a different PR.

PR #7961, opened 2023-09-20
JP-3342: add median file to outlier_detection cleanup block

### LINKED ISSUE #7959: Remove outlier_detection "median" files at end of processing
_Issue [JP-3422](https://jira.stsci.edu/browse/JP-3422) was created on JIRA by [Howard Bushouse](https://jira.stsci.edu/secure/ViewProfile.jspa?name=bushouse):_

SDP reports that in initial testing of the Cal 1.12.0 delivery they're seeing "_median.fits" files left behind by the outlier_detection step when doing level-3 processing of coronagraphic observations. An example file is "jw00628032001_03106_00001_nrcalong_median.fits". This causes problems for the schemes used in SDP to try to figure out which products coming from Cal need to be archived. The median files are strictly intermediate in nature and not to be considered archive products.

So similar to the way outlier_detection was updated for JP-3311 to remove the blot and outlier_i2d intermediate files, it also needs to be updated to remove the median files at the end of processing. But this should only be done if the "save_intermediate_results" parameter is set to False. It is set to False by default in operations processing. It is only turned on by users during off-line processing.

tapastro:
Resolves [JP-3422](https://jira.stsci.edu/browse/JP-3422)

Closes #7959 

This PR addresses another intermediate file not properly cleaned up by `outlier_detection` if `save_intermediate_results` is set to False. The median file has a variable output name, depending on if the step resamples the data or not. If resampling occurs, the name is not constructed from an input model filename but a drizzled model. 

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/outlier_detection/outlier_detection_step.py
+++ b/jwst/outlier_detection/outlier_detection_step.py
@@ -177,18 +177,20 @@ class OutlierDetectionStep(Step):
                 for model in self.input_models:
                     model.meta.cal_step.outlier_detection = state
                     if not self.save_intermediate_results:
-                        # remove unwanted files
-                        crf_file = self.make_output_path(basepath=model.meta.filename)
-                        outlr_basename = os.path.basename(crf_file)
-                        outlr_file = model.meta.filename.replace('cal', 'outlier_i2d')
-                        outlr_path = crf_file.replace(outlr_basename, outlr_file)
-                        if os.path.isfile(outlr_path):
-                            os.remove(outlr_path)
-                            self.log.debug(f"    {outlr_path}")
-                        blot_file = crf_file.replace('crf', 'blot')
-                        if os.path.isfile(blot_file):
-                            os.remove(blot_file)
-                            self.log.debug(f"    {blot_file}")
+                        # Remove unwanted files
+                        crf_path = self.make_output_path(basepath=model.meta.filename)
+                        crf_file = os.path.basename(crf_path)
+                        outlr_file = crf_file.replace('crf', 'outlier_i2d')
+                        outlr_path = crf_path.replace(crf_file, outlr_file)
+                        blot_path = outlr_path.replace('outlier_i2d', 'blot')
+                        # Need to insert outlier into filename, as median filename is generated from drizzled model
+                        undersplit = outlr_file.split('_')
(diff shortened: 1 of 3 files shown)


tapastro:
Realizing it's probably better to clean up the median filename in situ, rather than work around the wonky behavior as is. Though perhaps this will cause issues for MAST? 

Will push an updated commit soon that alters the default filename of the median output, such that the name is consistent whether the step resamples or not. The branching behavior is due to an interaction with suffix removal and the default suffix containing an underscore, leading to part of the original suffix sticking around.

tapastro:
There was some wonky behavior with the first go - use of `output_dir` moves the blot and median files to the correct location, but the `outlier_i2d` intermediate files don't appear to interact properly with `output_dir`, such that this PR has to look in the working dir rather than the proper path. This needs to be cleaned up in a future PR (unless I can get the outlier files to save to the correct directory quickly).

jwst/outlier_detection/outlier_detection_step.py
```diff
@@ -177,18 +177,19 @@ def process(self, input_data):
                 for model in self.input_models:
                     model.meta.cal_step.outlier_detection = state
                     if not self.save_intermediate_results:
-                        # remove unwanted files
-                        crf_file = self.make_output_path(basepath=model.meta.filename)
-                        outlr_basename = os.path.basename(crf_file)
+                        # Remove unwanted files
+                        crf_path = self.make_output_path(basepath=model.meta.filename)
```
hbushouse, line 181: So all of this file name building is being done just so that it can find the right files to delete, right? Presumably it matches the equivalent code (wherever it is) that creates/saves the files in the first place?

tapastro: Yes, and ideally the behavior would match across all files generated. Unfortunately, it looks as though the outlier_i2d files behave differently - possibly as a result of the datamodels.open(save_open) code? I'm trying to understand exactly where these files come from. The blot and median files are dumped using the typical model.save() method, and are easy to replicate.

tapastro:
Remaining issue from JIRA comment lies in bad intermediate filenames when flag is True - the unwanted files now get cleaned up as desired.

hbushouse (approved):
Looks good. No point (I don't think) in running regtests on the PR, since none of the regtests involve any of the intermediate files.

PR #7442, opened 2023-01-29
JP-3078: Add support custom reference WCS for the resample step

mcara:
Resolves [JP-3078](https://jira.stsci.edu/browse/JP-3078)

This PR adds``output_wcs`` argument to the ``resample_step`` and ``resample_spec_step`` that allows users to provide their own reference GWCS to the resample step. 

**Checklist for maintainers**
- [x] added entry in `CHANGES.rst` within the relevant release section
- [x] updated or added relevant tests
- [x] updated relevant documentation
- [x] added relevant milestone
- [x] added relevant label(s)
- [x] ran regression tests, post a link to the Jenkins job below.
      [How to run regression tests on a PR](https://github.com/spacetelescope/jwst/wiki/Running-Regression-Tests-Against-PR-Branches)
- [ ] Make sure the JIRA ticket is [resolved properly](https://github.com/spacetelescope/jwst/wiki/How-to-resolve-JIRA-issues)

#### Regression test(s):

https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/554/

PR #7971, opened 2023-09-25
Simplify ResampleStep output_wcs interface

jdavies-st:
This PR simplifies the interface to match a resampled output grid to an existing image in `ResampleStep`.  Currently, one needs to do something like the following to resample data onto the same output grid as an existing reference mosaic:

```python
import asdf
from jwst.steps import ResampleStep


with asdf.open("reference_mosaic_i2d.fits") as af:
    wcs = af.tree["meta"]["wcs"]
    new_af = asdf.AsdfFile(tree=dict(wcs=wcs))
    new_af.write_to("reference_mosaic_wcs.asdf")
    ysize, xsize = af.tree["data"].shape

result = ResampleStep.call("my_asn.json", output_wcs="reference_mosaic_wcs.asdf", output_shape=(xsize, ysize))
```

This PR improves the interface by not needing one to extract the wcs from the reference mosaic or figure out its data shape.  Instead, just pass the reference mosaic itself as the output wcs:

```python
from jwst.steps import ResampleStep


result = ResampleStep.call("my_asn.json", output_wcs="reference_mosaic_i2d.fits")
```

I have also updated the `ResampleStep.spec` comments as these are not comments for use by developers, but are the documentation to the `strun` command line interface.

**Checklist for maintainers**
- [x] added relevant label(s)


braingram:
Thanks for providing the example in the description of this PR. It has brought to my attention the usage of `asdf.open` to open a fits file. This is deprecated in asdf and will be removed in asdf 3.0 (which will be released soon). Please see these docs for more information:
https://asdf.readthedocs.io/en/latest/asdf/deprecations.html#asdf-in-fits-deprecation

It appears that to retain the functionality in your example the resample step will need to be updated to use stdatamodels to open the file (which now supports AsdfInFits).

I suspect this was not caught until now as the existing test uses an asdf file (instead of an AsdfInFits file) to supply the wcs. The changes in this PR to handle `ImageModel` and `SlitModel` instances may also require changes to accommodate the use of stdatamodels (instead of asdf) to open fits files (like in the test modification included in this PR).

jdavies-st:
Ah, thanks @braingram.  I will open with `datamodels.open()` then.  That will actually simplify and also allow one to pass the datamodel in memory.

Any idea what sort of exception will be raised in the future in asdf if I try to open a FITS file with `asdf.open()`?

braingram:
Using `datamodels.open` sounds like a great idea!

The error raised when a fits file is passed to `asdf.open` can be seen in the devdeps testing run (asdf main has already dropped AsdfInFits support):
https://github.com/spacetelescope/jwst/actions/runs/6301749854/job/17107497466?pr=7971#step:10:710
The error from the log is:
```
E           ValueError: Does not appear to be a ASDF file.
```

jdavies-st:
> The changes in this PR to handle `ImageModel` and `SlitModel` instances may also require changes to accommodate the use of stdatamodels (instead of asdf) to open fits files (like in the test modification included in this PR).

Now fixed in [2e8afbd](https://github.com/spacetelescope/jwst/pull/7971/commits/2e8afbda99389c8db5e7395ed39348a5a5747c09)

Oh, and I see the `devdeps` job caught this.  And its a ValueError.  🚀

braingram:
Thanks! Regression tests running at:
https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST-Developers-Pull-Requests/detail/JWST-Developers-Pull-Requests/971/pipeline/201

mcara:
Actually, originally [when I implemented this parameter](https://github.com/spacetelescope/jwst/pull/7442), `output_wcs`, I was thinking of allowing this usage. But I do not remember why I decided not to do this. I think this looks good.

One thing that worries me is that it allows users to easily provide _any_ data model as `output_wcs`, even a `cal` file. With ASDF one has to make a conscious effort to create that file (hopefully from a resampled image). For the very least we should add additional note to the docs that this WCS is expected to be without distortions and that there is no check for this.

docs/jwst/resample/arguments.rst
```diff
@@ -55,14 +55,17 @@ image.
...
-    bounding box of the WCS (if any). Argument ``output_shape`` overrides
-    computed image size and it is required when output WCS does not have
-    ``bounding_box`` property set.
+    File name of either a JWST i2d or s2d FITS file with a GWCS stored under
+    ``meta.wcs``, or an ``ASDF`` file with a GWCS stored under the ``"wcs"`` key
+    under the root of the file. The output image size is determined from the i2d
+    or s2d data size in the case of a FITS file, or from the bounding box of the
+    WCS in the enclosed ``ASDF`` file. Argument ``output_shape`` overrides
+    computed image size and it is required when output WCS is ASDF and does not
+    have ``bounding_box`` property set.
```
mcara, line 68:
```suggestion
    computed image size and it is required when output WCS in ASDF does not
    have its ``bounding_box`` property set.
```

mcara: OR
```suggestion
    computed image size and it is required when output WCS is provided as an ASDF file and does not
    have ``bounding_box`` property set.
```

jwst/resample/resample_step.py
```diff
@@ -170,12 +171,20 @@ def _check_list_pars(vals, name, min_vals=None):
...
             return None
 
-        with asdf.open(asdf_wcs_file) as af:
-            wcs = deepcopy(af.tree["wcs"])
+        try:
+            # File is ASDF file with a top-level wcs in the tree
+            with asdf.open(input_wcs_file) as af:
+                wcs = deepcopy(af.tree["wcs"])
+        except (KeyError, ValueError):
+            # File is ImageModel or SlitModel with the wcs under meta
```
mcara, line 183: Can this be a `SlitModel` in a `ResampleStep`?

mcara:
Oh, maybe it was because this code was supposed to be used by Roman and they will use ASDF files

> Thanks! Regression tests run with no errors:
> https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST-Developers-Pull-Requests/detail/JWST-Developers-Pull-Requests/971/pipeline/201

I do not think there is a test that actually uses and external reference WCS.

mcara:
> Thanks! Regression tests run with no errors:
> https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST-Developers-Pull-Requests/detail/JWST-Developers-Pull-Requests/971/pipeline/201

I do not think there is a test that actually uses and external reference WCS.

jwst/resample/resample_step.py
jdavies-st: Theoretically yes, but practically I think it will always be `ImageModel`.

jdavies-st:
> I do not think there is a test that actually uses an external reference WCS.

Right, just the unit test does, which I updated (perhaps not optimally).

jwst/resample/resample_step.py
```diff
@@ -170,12 +171,20 @@ def _check_list_pars(vals, name, min_vals=None):
...
+    def _load_custom_wcs(input_wcs_file, output_shape):
+        if not input_wcs_file:
             return None
 
-        with asdf.open(asdf_wcs_file) as af:
-            wcs = deepcopy(af.tree["wcs"])
+        try:
+            # File is ASDF file with a top-level wcs in the tree
+            with asdf.open(input_wcs_file) as af:
+                wcs = deepcopy(af.tree["wcs"])
```
nden, line 181: Would it be too much burden to require the `wcs` attribute in an asdf file to be under `meta` as well, to make it uniform? An argument for this is that this pipeline is supposed to work equivalently with asdf files, so it's conceivable some time in the future the reference file will be an `i2d.asdf` file with the wcs under `model.meta.wcs`. 
I think the uniformity of the file structure may help users, though if it's considered too much of a burden, I'd be OK with the current proposal.

jwst/resample/resample_step.py
nden: This function is also called by `ResampleSpecStep`, right?
Does passing an external WCS to spectral observations work?

jwst/resample/resample_step.py
```diff
@@ -170,12 +171,20 @@ def _check_list_pars(vals, name, min_vals=None):
...
+        except (KeyError, ValueError):
+            # File is ImageModel or SlitModel with the wcs under meta
+            with datamodels.open(input_wcs_file) as model:
+                wcs = deepcopy(model.meta.wcs)
+                if not output_shape:
+                    output_shape = model.data.shape[::-1]
```
nden, line 187: If I'm reading this correctly, it means that if an external reference file is provided the data shape takes precedence to the bounding box of the WCS (assuming it exists too). Is this the correct decision? It seems to me the bounding_box should take precedence.

nden: Not part of this PR but the line below (line 189) seems wrong, especially when `wcs is None`

jwst/resample/resample_step.py
jdavies-st: I think that would be an API change, as up to this point it's been required to be at the top level.  I.e. this PR doesn't change that.  I do know there's existing analysis code in the wild (ERO CEERS code on github) that specifically uses this functionality.

Agree it would be good to have consistency.  Maybe better to move `wcs` out of `meta` instead, something we've talked about off and on before.

jwst/resample/resample_step.py
mcara: > This function is also called by `ResampleSpecStep`, right? Does passing an external WCS to spectral observations work?

It should.

mcara:
> Actually, originally [when I implemented this parameter](https://github.com/spacetelescope/jwst/pull/7442), `output_wcs`, I was thinking of allowing this usage. But I do not remember why I decided not to do this. I think this looks good.
> 
> One thing that worries me is that it allows users to easily provide _any_ data model as `output_wcs`, even a `cal` file. With ASDF one has to make a conscious effort to create that file (hopefully from a resampled image). For the very least we should add additional note to the docs that this WCS is expected to be without distortions and that there is no check for this.

I think another reason for not allowing FITS as reference files is that users may try to use the same reference files from `drizzlepac` (used to process HST data) to resample JWST to the same grid as some of their HST observations.

I think this PR has a potential to open a lot of issues later unless everything is well documented and there are checks in the code to provide a meaningful error messages for unsupported situations.

Saving a WCS to a file is not very difficult:
```python
from jwst.datamodels import ImageModel
import asdf
im = ImageModel('resampled_i2d.fits')
af = asdf.AsdfFile({'wcs': im.meta.wcs})
af.write_to('ref_wcs.asdf')
```

melanieclarke:
It sounds like the conclusion from this conversation is that the user convenience from this change is not worth the additional complications it would introduce.

I will close this PR.  @jdavies-st, please feel free to file an issue or reopen if this needs more discussion.

PR #7972, opened 2023-09-25
JP-3304: use error table from f-flat reference files

braingram:
Partially resolves [JP-3304](https://jira.stsci.edu/browse/JP-3304)
Fully resolving this requires reference files to test the code modified in this PR.

https://github.com/spacetelescope/stdatamodels/pull/183 added support for `error` columns in `NirspecFlatModel` and `NirspecQuadFlatModel` files. This PR modifies `combine_fast_slow` based on suggestions in the above ticket to use the new error columns.

Regression tests running at: https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST-Developers-Pull-Requests/detail/JWST-Developers-Pull-Requests/972/pipeline/201

Some failures are expected as `f_flat` reference files do not contain `error` columns. https://github.com/spacetelescope/stdatamodels/pull/183 fills in the missing `error` column with `nan` values and the resulting interpolated `error` (with this PR) is `nan`. It appears that on the current main branch (and in the truth files) these `error` values are instead `0`. This should result in many `FITSDiff` errors where the `ERR` hdu contains `nan` instead of `0`.

**Checklist for maintainers**
- [ ] added relevant label(s)

### DIFF
--- a/jwst/flatfield/tests/test_combine_fast_slow.py
+++ b/jwst/flatfield/tests/test_combine_fast_slow.py
@@ -34,6 +34,7 @@ def test_combine_fast_slow():
              'c5': 2.0e-3}
     poly = polynomial.Polynomial1D(degree=5, **coeff)
     tab_flat = poly(tab_wl)
+    tab_flat_err = np.full(tab_flat.shape, np.nan, dtype=tab_flat.dtype)
 
     wl0 = 7.37
     dwl0 = 2.99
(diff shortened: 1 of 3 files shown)


jwst/flatfield/tests/test_combine_fast_slow.py
```diff
@@ -34,6 +34,7 @@ def test_combine_fast_slow():
              'c5': 2.0e-3}
     poly = polynomial.Polynomial1D(degree=5, **coeff)
     tab_flat = poly(tab_wl)
+    tab_flat_err = np.full(tab_flat.shape, np.nan, dtype=tab_flat.dtype)
```
braingram, line 37: I wasn't quite sure what values would be suitable for testing the error interpolation. I picked an array of `nan` as that is the expected "default" value that the datamodel will generate if a reference file is opened that does not contain an `error` column.

PR #7976, opened 2023-09-26
JP-3365: Increase MOS cutout margin

### LINKED ISSUE #7877: Resampling edge effects for off-center MOS targets
_Issue [JP-3365](https://jira.stsci.edu/browse/JP-3365) was created on JIRA by [Melanie Clarke](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mclarke):_

In performing calibration analysis for MOS mode, we discovered that there are discontinuities in the extracted level 2 spectra and s2d images at the upper and/or lower nod positions. These may be as large as ~ 40%.  The cause appears to be that there is missing flux along partial rows of the upper and lower edges in the resampled s2d file.  These edges correspond to areas in the cal file set to NaN due to missing WCS information.

The effects in level 3 products appear to be much smaller and/or absent, since combining the nods results in at least some coverage for the full spatial extent.  However, flux calibration and/or data quality may still be impacted in ways that could be improved if the wings of the upper and lower spectra were included in the final combination.

We can currently work around this issue by editing the MSA metadata FITS files to add an extra "open" shutter to each defined slitlet.  That is, for each 3-shutter slitlet, we pretend that 5 shutters were open, where the actual nod pattern uses only the central 3.  This allows the wings of the spectra to be included in the cal file and resampled properly in the s2d file, prior to extraction.  However, this workaround may also impact calibration for the source, including the pathloss correction (JP-3357).

The NIRSpec team will continue analyzing the issue to determine an appropriate fix.  We may want to extend the WCS coverage to all pixels currently included in the cal file.  We may also want add an extra pixel or two to the cutout boundaries to further mitigate the edge effects.

melanieclarke:
Resolves [JP-3365](https://jira.stsci.edu/browse/JP-3365)

Closes #7877 

This PR addresses edge effects in resampling MOS slit cutouts by adding a couple extra pixels of WCS coverage at the top and bottom of each slit.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,10 @@
 1.12.2 (unreleased)
 ===================
 
-- 
+assign_wcs
+----------
+
+- Increase margin at edges of NIRSpec MOS slits to reduce edge effects in resampling. [#7976]
 
 1.12.1 (2023-09-26)
 ===================
--- a/jwst/assign_wcs/nirspec.py
+++ b/jwst/assign_wcs/nirspec.py
@@ -635,7 +635,7 @@ def get_open_msa_slits(msa_file, msa_metadata_id, dither_position,
         slitlet_ids_unique.remove(-1)
 
     # add a margin to the slit y limits
-    margin = 0.05
+    margin = 0.5
 
     # Now lets look at each unique slitlet id
     for slitlet_id in slitlet_ids_unique:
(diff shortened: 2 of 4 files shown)


hbushouse:
Started a regression test run against the PR branch at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/977  Expecting small differences in results for any MOS processing.

hbushouse (commented):
Code updates are simple enough. Will wait for regtests to finish before final approval.

hbushouse:
Regression test results show a few unrelated failures (cause is known) and many expected differences in all MOS products beyond the stage where extract_2d has been run to extract individual slitlets. All cal, s2d, and other intermediate level-2b products show differences in the extraction region size, almost always an extra 4 pixels in the cross-dispersion direction (2 on top and bottom?), all of which seems reasonable. 1D extractions also show small differences due to inclusion of new edge pixel values. So I think this looks OK.

hbushouse (approved):

PR #7978, opened 2023-09-26
JP-3429: Flat error propagation

### LINKED ISSUE #7973: Propagate 1D flat errors
_Issue [JP-3429](https://jira.stsci.edu/browse/JP-3429) was created on JIRA by [Melanie Clarke](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mclarke):_

NIRSpec is updating its flat files to include a 1D error array, associated with the fast variation component in the F-flats (JP-3304).  The flat field step needs to be updated to propagate this error from the reference files to the output error arrays.

In particular, the combine_fast_slow function needs to be updated to take an error array as input and interpolate it onto the wavelengths for the observation.  The interpolated error should then be added in quadrature to any error from other components.

Two implementation notes:

1. Both zero and NaN are possible values for all flat error arrays, 2D and 1D.  The S and D flats will not have error arrays for their 1D components, so these values will default to NaN; their 2D error components will be set to zero, since all flat errors are accounted for in the F-flat instead. When combining flat error values, NaN values should be ignored and zero values should be treated as valid numbers.  

2. The 1D flat vector in the combine_fast_slow algorithm is interpolated with a 3-point Gaussian integration.  A rigorous error propagation through that step should take into account cross-correlations between the pixels that are weighted and summed together, but we don't recommend that.  The flat errors are not likely to be known to enough precision to justify the complexity.  Instead, we recommend interpolating the error vector onto the wavelengths with a simple linear interpolation, e.g.:
  error_value = np.interp(wavelengths, tab_wl, tab_flat_error, left=np.nan, right=np.nan)

melanieclarke:
Resolves [JP-3429](https://jira.stsci.edu/browse/JP-3429)

Closes #7973 

This PR updates flat propagation methods to account for 1D errors from fast variation components. This is needed to support upcoming NIRSpec flats, which will include errors on the 1D F-flats.  Missing errors are handled as zero-error components, so these changes should be backwards-compatible.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,12 @@ charge_migration
 - Added tests to see if the data array is changed after runnung the step and
   setting signal_threshold to 25000. [#7895]
 
+flat_field
+----------
+
+- Update ``combine_fast_slow`` to use 1D error values provided by f-flat
+  reference files. [#7978]
+
 1.12.1 (2023-09-26)
 ===================
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,8 +10,8 @@ charge_migration
 flat_field
 ----------
 
-- Update ``combine_fast_slow`` to use 1D error values provided by f-flat
-  reference files. [#7978]
+- Update the ``combine_fast_slow`` function for NIRSpec spectroscopic flats 
+   to use 1D error values provided by F-flat reference files. [#7978]
 
 1.12.1 (2023-09-26)
 ===================
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -1378,12 +1378,11 @@ def combine_fast_slow(wl, flat_2d, flat_dq, flat_err, tab_wl, tab_flat, tab_flat
     combined_dq[missing] |= dqflags.pixel['NO_FLAT_FIELD']
     combined_dq[missing] |= dqflags.pixel['DO_NOT_USE']
 
-    # Handle NaNs in errors
-    combined_err[np.isnan(combined_err)] = 0.0
-    error_value[np.isnan(error_value)] = 0.0
-
-    # Add new 1D errors and input 2D errors in quadrature
-    combined_err = np.sqrt(combined_err**2 + error_value**2)
+    # Add new 1D errors and input 2D errors in quadrature,
+    # treating NaNs as zeros
+    v1 = error_value**2 * flat_2d**2
+    v2 = combined_err**2 * values**2
+    combined_err = np.sqrt(np.nansum([v1, v2], axis=0))
 
     return flat_2d * values, combined_dq, combined_err
 
--- a/jwst/flatfield/flat_field.py
+++ b/jwst/flatfield/flat_field.py
@@ -434,7 +434,7 @@ def nirspec_fs_msa(output_model, f_flat_model, s_flat_model, d_flat_model, dispa
         flat_data_squared = slit_flat.data ** 2
         slit.var_poisson /= flat_data_squared
         slit.var_rnoise /= flat_data_squared
-        slit.var_flat = slit.data ** 2 / flat_data_squared * slit_flat.err ** 2
+        slit.var_flat = (slit.data / slit_flat.data * slit_flat.err) ** 2
         slit.err = np.sqrt(slit.var_poisson + slit.var_rnoise + slit.var_flat)
 
         # Combine the science and flat DQ arrays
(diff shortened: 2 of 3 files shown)


braingram:
Thanks! I started regression tests for this PR at:
https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/975/

CHANGES.rst
```diff
@@ -7,6 +7,12 @@ charge_migration
 - Added tests to see if the data array is changed after runnung the step and
   setting signal_threshold to 25000. [#7895]
 
+flat_field
+----------
+
+- Update ``combine_fast_slow`` to use 1D error values provided by f-flat
+  reference files. [#7978]
```
hbushouse, line 14:
```suggestion
- Update the ``combine_fast_slow`` function for NIRSpec spectroscopic flats 
   to use 1D error values provided by F-flat reference files. [#7978]
```

jwst/flatfield/flat_field.py
```diff
@@ -434,7 +434,7 @@ def nirspec_fs_msa(output_model, f_flat_model, s_flat_model, d_flat_model, dispa
         flat_data_squared = slit_flat.data ** 2
         slit.var_poisson /= flat_data_squared
         slit.var_rnoise /= flat_data_squared
-        slit.var_flat = slit.data ** 2 / flat_data_squared * slit_flat.err ** 2
```
hbushouse, line 437: If `flat_data_squared` is no longer used anywhere, the computation of it could be removed.

jwst/flatfield/flat_field.py
```diff
@@ -510,7 +510,7 @@ def nirspec_brightobj(output_model, f_flat_model, s_flat_model, d_flat_model, di
     flat_data_squared = interpolated_flat.data ** 2
     output_model.var_poisson /= flat_data_squared
     output_model.var_rnoise /= flat_data_squared
-    output_model.var_flat = output_model.data ** 2 / flat_data_squared * interpolated_flat.err ** 2
```
hbushouse, line 513: Same comment here regarding `flat_data_squared`

jwst/flatfield/flat_field.py
```diff
@@ -576,7 +576,7 @@ def nirspec_ifu(output_model, f_flat_model, s_flat_model, d_flat_model, dispaxis
         flat_data_squared = flat ** 2
         output_model.var_poisson /= flat_data_squared
         output_model.var_rnoise /= flat_data_squared
-        output_model.var_flat = output_model.data ** 2 / flat_data_squared * flat_err ** 2
```
hbushouse, line 579: ... and one more time

hbushouse (commented):
Looks good. Just a few minor comments. And I assume this is all backwards compatible with old F-flats that don't have the error column in the table?

hbushouse:
Let me know when you're ready for another regression test run and I'll fire one off.

jwst/flatfield/flat_field.py
melanieclarke: It's still used in the two lines just above.

melanieclarke:
Thanks @hbushouse - I'm done with my local testing, so I think this is ready for another regression test run.

melanieclarke:
> And I assume this is all backwards compatible with old F-flats that don't have the error column in the table?

Yes, this should work for the old F-flats.  The error values are all assigned to NaN when they're read in, which is handled as zero error in the code.  I've tested with the existing CRDS flats and see no change in the output, compared to products reduced with v1.12.1.

jwst/flatfield/flat_field.py
hbushouse: D'oh!  Of course. Never mind ...

hbushouse:
I'll start another regtest run as soon as the queue opens up.

hbushouse:
Regression tests started at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/980

hbushouse:
Regression test results are at https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/980/testReport/  @melanieclarke can you see those results (not sure if you have permission)?
They show differences in the VAR_FLAT and ERR array values for many NIRSpec fixed-slit and IFU test results, but none for any MOS mode tests (not sure why). The differences are small numerical differences, or differences showing a value where there used to be a zero or NaN, all of which seems expected due to the changes here in the way the various error estimates are propagated (even without using new ref files).

melanieclarke:
@hbushouse - I have permission enough to see the test results.  I'm pretty sure those changes are as expected, but I'll take a closer look to confirm.

melanieclarke:
For the values where there used to be NaN: this is due to the change in NaN handling for input flat values.  All NaNs are now set to zero so that NaNs in flat variance don't contaminate later error estimates.

For the numerical differences: from the filenames, it looks like the inputs are all pre-flight data, for which there are still some associated flats that do not have all 2D error values set to 0.  The difference comes from a change in assumption about the meaning of the 2D error. Before, the code assumed that the 2D error was the only source of error for each flat type, and it was normalized by the composed slow * fast components before being added in quadrature to the other components.  Now, it is assumed that 2D errors apply only to the 2D component, since 1D errors are provided separately. This changes the normalization for the non-zero errors.

This change is moot for all flight data: 2D errors are currently set to zero in CRDS flats, and we do not expect to provide non-zero 2D errors going forward.  However, this does resolve a long-standing difference in the NIRSpec validation tests, which also use pre-flight data (JP-3103).  Error calculations from this change now match the validation code.

hbushouse (approved):

PR #7980, opened 2023-09-27
MRS adjustments to straylight and 1d residual fringe routines

### LINKED ISSUE #7967: MRS 1d residual fringe doesn't apply to backgrounds
_Issue [JP-3427](https://jira.stsci.edu/browse/JP-3427) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

The MRS 1d residual fringe correction called from extract_1d is currently being applied only to the returned 'FLUX' array, and should also apply to the 'SURF_BRIGHT' and 'BACKGROUND' arrays.  Likewise, the computation is happening in the wrong place (prior to division by the aperture area as a function of wavelength, which could make it more likely for the routine to fail if there are appreciable differences in the aperture area for some wavelengths.

The routine should be moved and applied to the extra arrays.  Draft code for this exists, PR to follow once more testing has been done.

### LINKED ISSUE #7968: MRS straylight calculations need to be revised
_Issue [JP-3428](https://jira.stsci.edu/browse/JP-3428) was created on JIRA by [David Law](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dlaw):_

The MRS straylight routine does two things; it removes the cross-artifact and the pedestal dark.  However, recent investigation suggests that these two artifacts interact with one another at the 0.02 DN/s level, which is sufficient to produce appreciable artifacts in short wavelength data cubes for which the background is less than the pedestal dark.

Specifically, the pedestal should be subtracted first, so that the cross artifact correction isn't applied to light coming from detector glow rather than through the MRS optics.

PR to follow after additional testing of draft correction.

drlaw1558:
Resolves [JP-3427](https://jira.stsci.edu/browse/JP-3427)
Resolves [JP-3428](https://jira.stsci.edu/browse/JP-3428)

Closes #7967 
Closes #7968 

This PR addresses both JP-3427 and JP-3428, addressing minor issues with MRS 1d residual fringe correction and MRS straylight.  MRS residual fringe correction was not being applied to anything other than the FLUX array, and was taking place at a suboptimal location in the code.  MRS straylight was being applied prior to pedestal dark subtraction, which could cause small artifacts at the 0.01 DN/s level that were noticeable in some data.

**Checklist for maintainers**
- [x] added relevant label(s)


jemorrison (approved):
Looks good to me.

jemorrison:
@David did you run the regression tests on this PR?

drlaw1558:
@jemorrison Nope, I'm not actually sure how to do that.

CHANGES.rst
```diff
@@ -7,6 +7,18 @@ charge_migration
 - Added tests to see if the data array is changed after runnung the step and
   setting signal_threshold to 25000. [#7895]
 
+extract_1d
+----------
+
+- Move MIRI MRS residual fringe correction later in the code, and apply to surface
```
hbushouse, line 13: Perhaps include a reason *why* it's getting moved, i.e. what effect does this change have?

jwst/straylight/straylight.py
```diff
@@ -184,25 +204,28 @@ def correct_xartifact(input_model, modelpars):
...
+    usedata = usedata - model
 
-    # Remove any remaining pedestal stray light values based on inter-channel region
-    # chosen based on analysis of multiple bands of data from various programs
-    if (channel == '12'):
-        pedestal = np.nanmedian(output.data[280:740, 503:516])
-    else:
-        pedestal = np.nanmedian(output.data[280:740, 474:507])
+    # Now measure and remove the pedestal dark count rate measured between the channels
+    # Embed in a try/except loop to catch unusual failures
```
hbushouse, line 214:
```suggestion
    # Embed in a try/except block to catch unusual failures
```

jemorrison:
@drlaw1558 @hbushouse I set up a regression test using David's branch. It seems to be running. It is run #979

CHANGES.rst
drlaw1558: Added a more detailed note.  In brief, the residual fringe was getting applied in the stage where the spectrum is based on the aperture sum results; given that the area going into this sum can change for each wavelength plane the background or surf brightness spectra aren't necessarily robust until later when the normalization by the pixel area in the aperture has been done.

jwst/straylight/straylight.py
drlaw1558: Done, here and in a similar statement

jemorrison:
Regression test run: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/979/
@drlaw1558 I don't know if you have access to the regression test results. There are expected  small differences since you changed the straylight step. So I think they are ok. But see if you can view the results

drlaw1558:
Differences in the regtest look ok to me.

hbushouse (approved):
Looks good.

PR #7997, opened 2023-10-06
JP-3438: Allow custom grouping of models

### LINKED ISSUE #7996: Allow users to specify custom grouping of models for tweakreg and skymatch
_Issue [JP-3438](https://jira.stsci.edu/browse/JP-3438) was created on JIRA by [Mihai Cara](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mcara):_

In <https://jira.stsci.edu/browse/JP-3435> I described a problem with aligning NIRCAM images. A common approach used by the users is to manipulate various observation attributes in data models so that grouping algorithm would assign unique group ID to each input data model. This, however, is not very convenient. An alternative is to run pipeline on each input image independently. This has its own limitations.

We need to create a simpler way for users to specify a custom grouping for images in an ASN.

mcara:
Resolves [JP-3438](https://jira.stsci.edu/browse/JP-3438)
Closes #7996

This PR adds support for custom group ID so that users can conveniently modify how input models are grouped for the `tweakreg` and `skymatch` steps.

**Checklist for maintainers**
- [x] added relevant label(s)


nden (commented):
Can you add a test? 
If I understand this correctly I should be able to edit an association file and add a custom `group_id`. This does not work. Is `group_id` even saved to an association_file?

jwst/datamodels/container.py
```diff
@@ -402,11 +423,13 @@ def path(filename, idx=None):
     def models_grouped(self):
         """
         Returns a list of a list of datamodels grouped by exposure.
-        Assign an ID grouping by exposure.
+        Assign an ID grouping by exposure if not already assigned.
```
hbushouse, line 426:
```suggestion
        Assign a grouping ID by exposure, if not already assigned.
```

jwst/datamodels/container.py
```diff
@@ -402,11 +423,13 @@ def path(filename, idx=None):
...
 
-        Data from different detectors of the same exposure will have the
-        same group id, which allows grouping by exposure.  The following
-        metadata is used for grouping:
+        If ``model.meta.group_id`` does not exist or it is `None` then data
+        from different detectors of the same exposure will be assigned
+        same group ID, which allows grouping by exposure in the ``tweakreg`` and
+        in the ``skymatch`` steps. The following metadata is used when
+        determining grouping:
```
hbushouse, line 432:
```suggestion
        If ``model.meta.group_id`` does not exist or it is `None`, then data
        from different detectors of the same exposure will be assigned the
        same group ID, which allows grouping by exposure in the ``tweakreg`` and
        ``skymatch`` steps. The following metadata is used when
        determining grouping:
```

hbushouse (commented):
Will need updates to RTD somewhere (wherever appropriate) to explain this new capability.

mcara:
> Can you add a test? If I understand this correctly I should be able to edit an association file and add a custom `group_id`. This does not work. Is `group_id` even saved to an association_file?

It does work. Can you explain what is not working as expected?

I added a unit test. I had to modify the `headers.fits` file to add the following `meta` info:
```
meta.observation.sequence_id = '2'
meta.observation.visit_group = '02'
meta.observation.activity_id = '01'
```
Without `sequence_id `, `visit_group`, and `activity_id` defined, `ModelContainer.models_grouped` would not work.

nden:
> Can you explain what is not working as expected?

I was trying to generate an association file with a custom `group_id` . As described somewhere above I opened the files, assigned `model.meta.goup_id` and saved them. Then I ran `asn_from_list` to generate the association file. The new association is missing `goup_id`.  This may be a user error or an association enhancement.
However, it looks like the only way to use a custom group_id or not group observations is to edit an association. Does it make sense to have a parameter to calweb_image3, tweakreg or sky_match which disables grouping?

mcara:
> I was trying to generate an association file with a custom `group_id` . As described somewhere above I opened the files, assigned `model.meta.goup_id` and saved them. Then I ran `asn_from_list` to generate the association file.

`asn_from_list` would need to be updated in order to be able to do this. Currently, this is not supported and I am not sure there is need for this: If you edited the files and set `model.group_id` (NOT `model.meta.goup_id`) then `ModelContainer` will pick this up. This PR is similar to https://github.com/spacetelescope/jwst/pull/7022 in how it handles stuff. In fact, most machinery to support reading extra members from ASN and setting corresponding model attributes was done in that PR.

> However, it looks like the only way to use a custom group_id or not group observations is to edit an association.

If you set `model.group_id` and save the model, it should work.

> Does it make sense to have a parameter to calweb_image3, tweakreg or sky_match which disables grouping?

In principle yes, but it is harder to achieve that, I think... And it makes it easier to have confusing situations like what takes precedence. I need to investigate this a little.

nden:
This looks good to me. The other enhancements can be done in a different PR. Need to rebase this one.

docs/jwst/tweakreg/README.rst
```diff
@@ -124,6 +124,35 @@ in the catalog is the ``'weight'`` column, which when present, will be used
...
+of images may fail (or result in poor alignment). In this case, it may be
+desirable to align each image independently. This can be achieved either by
+setting ``image_model.meta.group_id`` attribute to a unique string or integer
+value or by adding ``group_id`` attribute to the ``members`` of the input ASN
+table - see `~jwst.datamodels.ModelContainer` for more details.
+
+.. note::
+    Group ID (``group_id``) is used by both ``tweakreg`` and ``skymatch`` steps
+    and so modifying it for one step will affect the results in another step.
+    If it is desirable to apply different grouping strategies to ``tweakreg``
```
hbushouse, line 152:
```suggestion
    If it is desirable to apply different grouping strategies to the ``tweakreg``
```

nden (approved):

PR #7998, opened 2023-10-06
keep cube build input models open to allow reading wcs

braingram:
asdf previously had a bug where contents from a closed file could be accessed. See:
https://github.com/asdf-format/asdf/issues/1539

The cube build step was relying on this bug, closing models before the wcs was fully read (this is lazy loaded).

This commit prevents the early model closure. A unit test (using fake `nirspec` data) is added that replicates the issue and passes with the changes in this PR.

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/cube_build/data_types.py
+++ b/jwst/cube_build/data_types.py
@@ -94,9 +94,9 @@ class DataTypes():
             # close files opened above
             self.close()
             raise TypeError("Failed to process file type {}".format(type(input_models)))
-        # if the user has set the output name - strip out *.fits
-        # later suffixes will be added to this name to designate the
-        # channel, subchannel or grating,filter the data is covers.
+        # If the user has set the output name, strip off *.fits.
+        # Suffixes will be added to this name later, to designate the
+        # channel+subchannel (MIRI MRS) or grating+filter (NRS IFU) the output cube covers.
 
 
         if output_file is not None:
--- a/jwst/cube_build/data_types.py
+++ b/jwst/cube_build/data_types.py
@@ -69,30 +69,34 @@ class DataTypes():
 
         self.input_models = []
         self.output_name = None
-        
+
         # open the input with datamodels
         # if input is filename or model when it is opened it is a model
         # if input if an association name or ModelContainer then it is opened as a container
 
-        with datamodels.open(input) as input_models:
+        input_models = datamodels.open(input)
+        # if input is a filename, we will need to close the opened file 
+        self._opened = [input_models]
 
-            if isinstance(input_models, datamodels.IFUImageModel):
-                # It's a single image that's been passed in as a model
-                # input is a model
-                filename = input_models.meta.filename
-                self.input_models.append(input_models)
-                self.output_name = self.build_product_name(filename)
+        if isinstance(input_models, datamodels.IFUImageModel):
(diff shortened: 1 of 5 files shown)


jwst/cube_build/data_types.py
```diff
@@ -69,30 +69,34 @@ def __init__(self, input, single, output_file, output_dir):
...
+            self.input_models = input_models
+            if not single:  # find the name of the output file from the association
+                self.output_name = input_models.meta.asn_table.products[0].name
+        else:
+            # close files opened above
+            self.close()
+            raise TypeError("Failed to process file type {}".format(type(input_models)))
+        # if the user has set the output name - strip out *.fits
+        # later suffixes will be added to this name to designate the
+        # channel, subchannel or grating,filter the data is covers.
```
hbushouse, line 99:
```suggestion
        # If the user has set the output name, strip off *.fits.
        # Suffixes will be added to this name later, to designate the
        # channel+subchannel (MIRI MRS) or grating+filter (NRS IFU) the output cube covers.
```

hbushouse (commented):
Looks OK to me, other than the one suggestion. Really need input from @jemorrison on this.

hbushouse:
Needs a change log entry.

jemorrison (approved):
Looks good to me. Sorry for the delay in reviewing.

braingram:
Oh no... this wasn't quite ready for review. It will be very soon. I was able to sort out a unit test that runs (without error) the cube build step with fake nirspec data.

The new unit test `test_call_cube_build_nirspec` fails when run with a filename as input and without the changes in this PR when run with asdf main. This replicates the issue seen in the failing regression test (without having to run the regression test). The new unit test passes with the fixes applied in this PR.

Note the new test makes it through the cube build step and each test takes ~1.5 minutes on my machine. The new test is parametrized to run twice, once with a filename and once with a model as an input. Some of the 1.5 minutes is in the fixture for the test so overall this PR may slow the test suite down by about 2 minutes (depending on the system running the tests).

braingram:
Running the `test_cube_build_nirspec_internal_cal` regression tests locally now no longer fails with the error about attempting to read from a closed file. This local testing was done on a mac and the test does still fail (as it has been failing on the mac regression tests) with a `FitsDiff` error:
```
E         Extension HDU 1 (SCI, 1):
E
E            Data contains differences:
E              Data differs at [25, 34, 573]:
E                 a> 0.041410204
E                  ?       - ^^
E                 b> 0.04140949
E                  ?        ^ +
E              Data differs at [1, 34, 1647]:
E                 a> -0.045893423
E                  ?          ^^^
E                 b> -0.045893997
E                  ?          ^^^
E              2 different pixels found (0.00% different).
```
When the server is free I will start a new regression test run with the added parameter to install asdf from main (to get what will be asdf 3.0).

braingram:
Regression tests running (with installed asdf main/3.0) at: https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FJWST-Developers-Pull-Requests/detail/JWST-Developers-Pull-Requests/989/pipeline/198

nden (approved):
Regression tests show the same two failures in in a Nircam catalog tests as the nightly.

PR #8004, opened 2023-10-11
enable dependabot

braingram:
This PR adds a dependabot configuration file which should also enable dependabot to run on this repo.

The configuration file contains sections to keep both github actions versions up-to-date and upper pins for python dependencies (via pip) up-to-date.

I don't have access to the repository settings and I'm not sure if dependabot is enabled there as I found one old PR for jwst from dependabot:
https://github.com/spacetelescope/jwst/pull/6092

However it appears that at least the github actions are not being checked as an older version (3) of the checkout action is in use:
https://github.com/spacetelescope/jwst/blob/5c5b968e3c192b051c431ee8d42702e1c22a80ed/.github/workflows/publish-to-pypi.yml#L12

For an example of what to expect from dependabot PRs, here is the PR for asdf opened by dependabot to update the checkout action in that repository:
https://github.com/asdf-format/asdf/pull/1639

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    target-branch: "master"
+    schedule:
+      interval: "weekly"
+
+  # Maintain dependencies for pip
+  - package-ecosystem: "pip"
+    directory: "/"
+    target-branch: "master"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,8 @@ updates:
     target-branch: "master"
     schedule:
       interval: "weekly"
+    reviewers:
+      - "zacharyburnett"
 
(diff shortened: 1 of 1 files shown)


zacharyburnett (changes requested):
this looks great, thanks! I'm assuming the dependabot PRs ping the codeowners by default, but to keep @hbushouse from being spammed by PRs all the time, is there a way to make it ping me in addition / instead?

braingram:
Thanks for taking a look and good idea!

I think this should do it: ce3748f9227d3c75bc0d40ebd05538ce69b2fa4f

I added you as one of the [reviewers](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#reviewers). There is another option for [assignees](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#assignees) but reviewers seemed like a better fit. I've never used either option so we may just have to wait and see what the first PR looks like and adjust as needed.

zacharyburnett (approved):

hbushouse (approved):

PR #8013, opened 2023-10-19
allow for slow github mac runners

braingram:
As seen in the CI run for a recent PR (and for others at seemingly random times):
https://github.com/spacetelescope/jwst/actions/runs/6562032419/job/17823126182?pr=8010#step:10:182

the `test_exec_time_0_crs` test fails due to a seemingly slow github mac runner:
```
>       assert t_elapsed < MAX_TIME
E       assert 18.11580204963684 < 10
```

This PR updates the test to:
- use [time.monotonic](https://docs.python.org/3/library/time.html#time.monotonic)
- increase the threshold to 60 seconds for mac osx runners on github (which sets the `CI` environment variable)

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/jump/tests/test_detect_jumps.py
+++ b/jwst/jump/tests/test_detect_jumps.py
@@ -7,6 +7,8 @@ from stdatamodels.jwst.datamodels import GainModel, ReadnoiseModel, RampModel, d
 from jwst.jump.jump import run_detect_jumps
 import multiprocessing
 
+import os
+import platform
 import time
 import random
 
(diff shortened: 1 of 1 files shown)


jwst/jump/tests/test_detect_jumps.py
```diff
@@ -28,13 +30,17 @@ def test_exec_time_0_crs(setup_inputs):
...
     # using dummy variable in next to prevent "F841-variable is assigned to but never used"
     _ = run_detect_jumps(model, gain, rnoise, 4.0, 5.0, 6.0, 'none', 200, 4, True)
-    tstop = time.time()
+    tstop = time.monotonic()
 
     t_elapsed = tstop - tstart
-    MAX_TIME = 10  # takes 1.6 sec on my Mac
+    # sometimes this takes a long time on github mac osx runners
+    if platform.system() == 'Darwin' and os.environ.get('CI', False):
+        MAX_TIME = 60
```
hbushouse, line 41: Given that the original code mentions a time of 1.6 sec on a personal mac, a new limit of 60 sec seems unusually high to me. That's a factor of nearly 40x and just the kind of increase this test is supposed to be looking for. Could we maybe at least downgrade the new limit to something like 20?  Even that is a large increase relative to the original 1.6.

braingram: Thanks for looking at this. I switched the `MAX_TIME` to 20 for the mac github runners. The linux runners still use the same 10 seconds as before (as do any local tests or jenkins linux runs). I'm not certain the jenkins mac runs set the CI variable or not but can look into that if there's a concern for relaxing the threshold for those runs.

I also updated the comment to point to a closed (but still active) issue for slow/inconsistent mac runners which illustrates that this is a general issue and nothing specific to jwst:
https://github.com/actions/runner-images/issues/1336

hbushouse (approved):

PR #8014, opened 2023-10-19
update SourceCatalog meta data handling for new photutils changes

braingram:
The development version of photutils updated what is contained in the `SourceCatalog` meta data. One specific change that is impacting jwst is changing `version` to `versions` in the `meta`.
https://github.com/astropy/photutils/pull/1640

This is causing devdeps failures:
https://github.com/spacetelescope/jwst/actions/runs/6562032419/job/17823126866?pr=8010#step:10:339

This PR updates `set_segment_properties` to reproduce `version` from the new `versions` key in the `SourceCatalog` `meta`.

Regression tests (with released version of photutils) running at: https://plwishmaster.stsci.edu:8081/job/RT/job/JWST-Developers-Pull-Requests/1048/

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/jwst/source_catalog/source_catalog.py
+++ b/jwst/source_catalog/source_catalog.py
@@ -239,8 +239,7 @@ class JWSTSourceCatalog:
         # so copy over the new 'versions' into 'version' and remove 'versions'
         # to stay consistent with the old behavior that this code expects
         if 'versions' in segm_cat.meta:
-            segm_cat.meta['version'] = segm_cat.meta['versions']
-            del segm_cat.meta['versions']
+            segm_cat.meta['version'] = segm_cat.meta.pop('versions')
 
         self.meta.update(segm_cat.meta)
         for key in ('sklearn', 'matplotlib'):
--- a/jwst/source_catalog/source_catalog.py
+++ b/jwst/source_catalog/source_catalog.py
@@ -235,6 +235,13 @@ class JWSTSourceCatalog:
         self._xpeak = segm_cat.maxval_xindex
         self._ypeak = segm_cat.maxval_yindex
 
+        # newer versions of photutils use 'versions' instead of 'version'
+        # so copy over the new 'versions' into 'version' and remove 'versions'
+        # to stay consistent with the old behavior that this code expects
+        if 'versions' in segm_cat.meta:
+            segm_cat.meta['version'] = segm_cat.meta['versions']
+            del segm_cat.meta['versions']
+
         self.meta.update(segm_cat.meta)
         for key in ('sklearn', 'matplotlib'):
             self.meta['version'].pop(key)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,11 @@ resample
   results in modified values in the resampled images. New computations
   significantly reduce photometric errors. [#7894]
 
+source_catalog
+--------------
+
+- Update meta data for compatibility with photutils changes [#8014]
+
 tweakreg
 --------
 


jwst/source_catalog/source_catalog.py
```diff
@@ -235,6 +235,13 @@ def set_segment_properties(self):
         self._xpeak = segm_cat.maxval_xindex
         self._ypeak = segm_cat.maxval_yindex
 
+        # newer versions of photutils use 'versions' instead of 'version'
+        # so copy over the new 'versions' into 'version' and remove 'versions'
+        # to stay consistent with the old behavior that this code expects
+        if 'versions' in segm_cat.meta:
+            segm_cat.meta['version'] = segm_cat.meta['versions']
+            del segm_cat.meta['versions']
+
```
hbushouse, line 244: Is there a more elegant way to do this by somehow storing either `version` or `versions` (whichever appears in the meta data) in a local variable (or something like that) that then gets used in subsequent code?  Not sure I like the idea of forcing the code to continue using the old entry forever.

braingram: If the changed output is acceptable I can look at updating this pr to use either `version` or `versions` and produce a file that for all released versions of photutils would produce a file with `version` and for all new versions produce a file with `versions. @hbushouse is it preferable to match the new output of photutils or match the previous output?

jwst/source_catalog/source_catalog.py
```diff
@@ -235,6 +235,13 @@ def set_segment_properties(self):
         self._xpeak = segm_cat.maxval_xindex
         self._ypeak = segm_cat.maxval_yindex
 
+        # newer versions of photutils use 'versions' instead of 'version'
+        # so copy over the new 'versions' into 'version' and remove 'versions'
+        # to stay consistent with the old behavior that this code expects
+        if 'versions' in segm_cat.meta:
+            segm_cat.meta['version'] = segm_cat.meta['versions']
+            del segm_cat.meta['versions']
```
mcara, line 243:
```suggestion
            segm_cat.meta['version'] = segm_cat.meta.pop('versions')
```

braingram: Thanks!

larrybradley:
Why is this necessary?  The meta dict in the output catalog is informational only.  While technically this is a breaking change, no one should be accessing that dict in downstream code.

braingram:
> Why is this necessary? The meta dict in the output catalog is informational only. While technically this is a breaking change, no one should be accessing that dict in downstream code.

Thanks for taking a look. The meta dictionary is both copied into the source catalog meta data and modified here in the jwst source catalog step:
https://github.com/spacetelescope/jwst/blob/68b8c200b20130ebe33e01643211507a021f9ad5/jwst/source_catalog/source_catalog.py#L238-L240
The change to use `versions` (instead of `version`) causes the modifications to fail with a `KeyError` (as seen in the [devdeps testing on main](https://github.com/spacetelescope/jwst/actions/runs/6788681962/job/18454223297#step:10:812).

larrybradley:
Instead, that block of code (L238-240) should be removed or updated.  The intent was to not include optional dependencies not used in the pipeline.  However, since I originally wrote the block, `photutils` has added new optional dependencies which are *not* currently being removed.  Either we should simply remove that code block, or it should be updated to only *include* (instead of *exclude*) certain packages for which we want to log the versions.

larrybradley:
I'm going to revert this change in `photutils` and postpone it for a future release.  I was planning a release soon, and I don't want to break the pipeline.  I didn't expect that change to break anything, but of course it was something that I wrote in the pipeline long ago.  😅    In the meantime, I'll update the code here to prepare for that change in a future photutils release.

braingram:
Let me know if there's anything I can do to help. I'll convert this to draft if that works for you.

braingram:
Closing in favor of https://github.com/spacetelescope/jwst/pull/8066

